Issue #305, opened 2018-04-03
[miral-shell] Hiding "minimized" windows makes it impossible to restore them


AlanGriffiths:
While applications can change the state (e.g. to restored) there's no way for a user to interact with them to make this happen.

[edit]
Actually, this seems different that first appears:

Using `miral-app -launcher gedit`

1. Minimize the window from the titlebar
2. Press F11
3. Press Alt-Tab
4. Press F11

The edit window appears maximized after step 4, implying that it gained focus with step 3 (as the application processed the F11 key).

AlanGriffiths:
Actually, most of this logic exists in miral::BasicWindowManager, that's where a "sensible default" should be implemented.

AlanGriffiths:
Minimized windows should be in the Alt-Tab list. Hidden ones should remain hidden.

PR #741, opened 2019-02-27
Restore minimized windows when they receive input focus

AlanGriffiths:
Restore minimized windows when they receive input focus. (Fixes: #305)


AlanGriffiths:
Hmm, this doesn't (yet) work with multiple apps running.

PR #751, opened 2019-03-01
Check for zero size Windows in WindowWlSurfaceRole

### LINKED ISSUE #674: Non-critical error logged as critical
On 18.10 with Mir1.1.0-rc12 I ran:

    $ miral-desktop -kiosk

then ran some programs and exited. The following appeared in the log.
```
[2018-12-17 10:14:46.872497] <CRITICAL> frontend:Wayland: Exception processing Surface::commit() request: /build/mir-fN5xrV/mir-1.1.0~rc12-gd4574168b8/src/miral/basic_window_manager.cpp(1949): Throw in function void miral::BasicWindowManager::validate_modification_request(const miral::WindowSpecification&, const miral::WindowInfo&) const
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::runtime_error> >
std::exception::what: width must be positive
```

wmww:
Fixes #674. Check to make sure the window has a non-zero size before modifying the surface.

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -248,9 +248,12 @@ void mf::WindowWlSurfaceRole::set_state_now(MirWindowState state)
     }
 }
 
-geom::Size mf::WindowWlSurfaceRole::window_size()
+std::experimental::optional<geom::Size> mf::WindowWlSurfaceRole::window_size()
 {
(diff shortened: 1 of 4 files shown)


wmww:
@RAOF Good catch. I wrote another solution, which IMO improves overall code quality. If someone can please test resizing clients, that would be great. I'm having some weird stuff happen even on master, so until I get that sorted I can't test this very well. It passes WLCS.

PR #755, opened 2019-03-06
Add an undocumented option to miral-app to run mir_demo_server

AlanGriffiths:
This is useful when testing features that only mir_demo_server exposes


wmww:
What is the reason for keeping it undocumented? If it's not useful to anyone else, why is it useful to you?

AlanGriffiths:
On Wed, 6 Mar 2019 09:25:38 -0800, William Wold wrote:
> What is the reason for keeping it undocumented? If it's not 
> useful to anyone else, why is it useful to you? 
>

Because it only works if both mir-demos and mir-test-tools are installed.

-- 

wmww:
Ah, that makes sense

bors r+

PR #756, opened 2019-03-07
Don't "GPL-required" check protocols

AlanGriffiths:
By "chance", the current protocols pass the check, but they shouldn't be checked.

Other protocols (e.g. KWin's "server-decoration" protocol) would fail.

(As a byproduct, this also cleans up some wlcs submodule legacy.)


AlanGriffiths:
> Could you please clean up the commits? You've got a stray “Add an undocumented option to miral-app to run mir_demo_server” commit in there which does nothing of the sort grinning

Fixed.

Sorry, wasn't paying attention when cherry-picking.

AlanGriffiths:
bors r=wmww

PR #758, opened 2019-03-08
Remove unused internal #include from generated Wayland wrappers.

AlanGriffiths:


AlanGriffiths:
Let's see if rawhide is clean now...

bors r=wmww

AlanGriffiths:
> Let's see if rawhide is clean now...

No, dropped it for now.

bors r=wmww

PR #762, opened 2019-03-12
XWayland: Improve listen/x-client sockets

mariogrip:
This improves the listen sockets

X11 should now work without need for lazy override
I also moved lazy to use the listen sockets mir creates

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -87,6 +87,8 @@ void mf::XWaylandServer::spawn()
     int status;
     std::string fd_str, abs_fd_str, wm_fd_str;
 
+    xserver_status = STARTING;
+
     // Try 5 times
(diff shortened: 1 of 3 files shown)


mariogrip:
It is indeed very C80, I plan to clean it up with modern c++. Reason why i left it right now is that i want to get the functional part of it working before starting ripping it apart again. Also try to understand the timings between when Xwayland expects connections etc. It does borrow lots from weston and whats where the old C comes from :) but I will clean that up.

AlanGriffiths:
> It is indeed very C80, I plan to clean it up with modern c++. Reason why i left it right now is that i want to get the functional part of it working before starting ripping it apart again. Also try to understand the timings between when Xwayland expects connections etc. It does borrow lots from weston and whats where the old C comes from :) but I will clean that up.

I understand. You don't have to rush, I simply didn't have time to convince myself that this doesn't break anything that isn't already broken.

PR #767, opened 2019-03-21
Remove `BufferStream::resize()`

### LINKED ISSUE #604: Window resizing isn't smooth
It stutters (this is much more obvious on a slower device).

(Originally reported: https://github.com/AlanGriffiths/egmde/issues/13 by @punhetonildo)

wmww:
A `BufferStream`'s size is now just the size of the most recent buffer (or the initial size sent to the constructor, if there is no buffer yet). This simplifies the code, fixes #604, and the non-atomic resize problem noted in a TODO in wl_surface.cpp.

### DIFF
--- a/include/server/mir/frontend/buffer_stream.h
+++ b/include/server/mir/frontend/buffer_stream.h
@@ -42,7 +42,6 @@ public:
     virtual ~BufferStream() = default;
     
     virtual void submit_buffer(std::shared_ptr<graphics::Buffer> const& buffer) = 0;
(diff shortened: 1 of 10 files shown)


wmww:
It should be noted that window resize still isn't perfect. When resizing from the top left there is a slight jitter in lower and right edges. This is due Mir moving, clipping and rendering the buffer slightly faster than the client can respond with a resized buffer. The fix would probably be something like having the surface ask the client to resize without actually performing the resize on itself.

PR #768, opened 2019-03-22
wmww's suggestions for bespoke extensions

wmww:
I don't have time now, but I also imagine filtering would be changed a bit. Instead of a single filter function that gets the interface by name, I'd suggest a filter function only gets the app as an argument and is attached to an extension.


AlanGriffiths:
I know we started to discussed this idea on IRC, but I'll respond for the record.

I think there are two "problems" with the initial proposal that you are trying to solve:
1. to associate the name of the extension with the builder for the extension; and,
2. to provide something safer to use than std::function<>?

# associate the name of the extension with the builder for the extension

I agree, that the association of name and builder could be made stronger. If it were likely that the user wants to pass name+builder around, then combining them would be a good design.

However, passing both name and builder to a function is simpler than using them to create an object only when passing that object to a function.

# something safer to use than std::function<>

While your approach could be argued to be safer than std::function<>, it is less familiar, needs more code to use and is less flexible.

Let me take these points in turn:

* "less familiar" users are far more likely to have encountered std::function<> than any class we write (and  `miral::WaylandExtension` in particular,
* "needs more code" instead of passing a functor it becomes necessary to implement an interface and make a shared_ptr<> to an instance of the interface.
* "less flexible" the original code allowed a functor that could be a plain function, a lambda, or a user written class. This approach forces a user written class.

_Entia non sunt multiplicanda praeter necessitatem._

include/miral/miral/wayland_extensions.h
```diff
@@ -31,6 +31,43 @@ namespace mir { class Server; }
...
+    {
+    };
+
+    /// The name of the implemented Wayland protocol interface
+    virtual auto interface_name() const -> std::string = 0;
+
+    /// Create an instance for the given context
+    virtual auto instantiate_for(Context const& context) const -> std::shared_ptr<Instance> = 0;
+
+    virtual ~WaylandExtension() = default;
```
AlanGriffiths: We ought to disable CopyAssign for an interface.

include/miral/miral/wayland_extensions.h
```diff
@@ -31,6 +31,43 @@ namespace mir { class Server; }
...
+        friend WaylandExtensions;
+    };
+
+    /// An instance of this extension, connected to a specific Wayland display
+    class Instance
+    {
+    };
+
+    /// The name of the implemented Wayland protocol interface
+    virtual auto interface_name() const -> std::string = 0;
```
AlanGriffiths: A virtual function seems overkill for what amounts to a constant property.

AlanGriffiths:
For reviewers: Another point that came up on IRC interfaces that user code implement are part of the ABI and can only change when ABI breaks.

AlanGriffiths:
@wmww there is one thing I do like here, and will adopt regardless of any other votes, is passing a Context object into the builder. That is extensible without breaking ABI in a way that my original proposal to pass specific parameters isn't.

wmww:
With passing in a `Context` object, the `std::function` approach is cleaner (one of my objections was messy, confusing type signatures).

I'd argue it is important to connect the name and the builder, if only to keep them in the same file. Say, for example, you upgraded from one version of an unstable protocol to another. You should be able to make all the required changes in one place, without hunting through the code for somewhere a name is attached (keeping in mind that the same name *always* goes with a given protocol). This could also be accomplished with a `static const struct` that holds the builder function and the name string.

wmww:
@AlanGriffiths It may be neither here nor there, but perhaps an extension builder could call a `Context::register_extension(std::shared_ptr<void>)` method, instead of returning a `std::shared_ptr<void>`/`std::shared_ptr<Context::Instance>`. This could at first just accept a `std::shared_ptr<void>`, but could later be extended in an ABI stable way.

For example if we wanted to be able to toggle extensions dynamically, we could create an `ExtensionInstance` class with `.stop()` and `.start()` methods. Then we could provide the `Context::register_extension(std::shared_ptr<ExtensionInstance>)` overload without breaking ABI.

AlanGriffiths:
> I'd argue it is important to connect the name and the builder, if only to keep them in the same file. Say, for example, you upgraded from one version of an unstable protocol to another. You should be able to make all the required changes in one place, without hunting through the code for somewhere a name is attached (keeping in mind that the same name _always_ goes with a given protocol). This could also be accomplished with a `static const struct` that holds the builder function and the name string.

Sure, the client is free to do that.

PR #772, opened 2019-03-25
Configure Wayland extensions

AlanGriffiths:
Make it easier for servers to set defaults for configuring Wayland extensions

### DIFF
--- a/src/miral/wayland_extensions.cpp
+++ b/src/miral/wayland_extensions.cpp
@@ -100,12 +100,7 @@ auto miral::WaylandExtensions::supported_extensions() const -> std::string
     return self->available_extensions.substr(0, self->available_extensions.size()-1);
 }
 
-auto miral::WaylandExtensions::default_extensions() const -> std::string
-{
-    return self->default_value;
-}
-
-auto miral::WaylandExtensions::standard_extensions() -> std::string
+auto miral::WaylandExtensions::recommended_extensions() -> std::string
 {
     return mo::wayland_extensions_value;
 }
@@ -113,7 +108,10 @@ auto miral::WaylandExtensions::standard_extensions() -> std::string
 void miral::WaylandExtensions::operator()(mir::Server& server) const
 {
     self->validate(self->default_value);
-    server.add_configuration_option(mo::wayland_extensions_opt, "Wayland extensions to enable", self->default_value);
+    server.add_configuration_option(
+        mo::wayland_extensions_opt,
+        ("Wayland extensions to enable. [" + supported_extensions() + "]"),
+        self->default_value);
 
     server.add_pre_init_callback([self=self, &server]
         {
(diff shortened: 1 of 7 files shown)


src/miral/wayland_extensions.cpp
```diff
@@ -98,6 +96,16 @@ miral::WaylandExtensions::WaylandExtensions(std::string const& default_value) :
 }
 
 auto miral::WaylandExtensions::supported_extensions() const -> std::string
+{
+    return self->available_extensions.substr(0, self->available_extensions.size()-1);
```
wmww: It appears this relies on `self->available_extensions` ending with a trailing colon. As this doesn't seem to be enforced (as far as I can tell), we should probably check before we chop it off.

Alternatively, we could *not* use colon separated strings internally, and not have this problem.

AlanGriffiths:
> ## Naming
> 
> By my understanding, there are no bespoke extensions by default. The user can add bespoke extensions and also override which standard extensions are on and which are off. This leaves us with five non-exclusive groups of extensions
> 
>     * Those supported by Mir internally, and turned on by default
> 
>     * Those supported by Mir internally, but turned off by default
> 
>     * Those supported by Mir internally, and turned on currently
> 
>     * Those supported by Mir internally, but turned off currently
> 
>     * Those implemented by 3rd parties using the bespoke API (always on if added?)
> 
> 
> It is non obvious from the descriptions what these groups are, or which functions return which combinations of above groups.

I think of it differently:

1. There is a set of supported extensions `supported_extensions()`
1.1. This includes the Wayland extensions supported by Mir
1.2. There is a suggested "standard" subset of 1.1: `WaylandExtensions::standard()`
1.3. Additional Wayland extensions can be included by the server `with_extension()`

2. When the server is started, the enabled extensions are selected from the supported extensions.
2.1. If not supplied by runtime configuration, the default set of extensions are enabled. `default_extensions()`
2.2. If the server doesn't set the default extensions, the `standard()` set is the default.

Maybe this would be clearer if I drop `default_extensions()` or add a `set_default_extensions(...)`?

wmww:
Ohhh... I think I get it. The term "default" implies that whatever it is, it is possible to override it with a non-default value. I now understand that `default_extensions()` is not a default provided by Mir for shells to override (that's `standard()`), but instead a default provided by shells for the runtime configuration to override. If you communicate in the comments that available and standard are things provided by Mir, default is what the shell dev controls, and how a non-default value might be used, that may clear things up.

AlanGriffiths:
> ## Colon separated string
> 
> I believe we should be using a `std::vec<std::string>` or `std::set<std::string>` internally, instead of the colon separated string. We would still expose the old strings for backwards compatibility, but they would be deprecated and new Miral functions would only expose the new data structure. This would have the following advantages:

This class is largely about supporting a configuration option that is supplied as a colon separated string. The only usecases I can envisage for server code is setting alternative defaults, and that will usually be adding extensions:
```
        miral::WaylandExtensions{miral::WaylandExtensions::recommended_extensions() + ":zwlr_layer_shell_v1"},
```
I don't think that gains from using a set:
```
        auto default_extensions = miral::WaylandExtensions::recommended_extensions();
        default_extensions.insert(":zwlr_layer_shell_v1");
        miral::WaylandExtensions{default_extensions},
```

wmww:
bors cancel

PR #773, opened 2019-03-25
Use function 'std::abs' (not '::abs') for double parameter.

### LINKED ISSUE #770: Compilation error with clang 7
```
/home/pmos/build/src/mir-1.1.2/src/miral/static_display_config.cpp:306:37: error: using integer absolute value function 'abs' when argument is of floating point type [-Werror,-Wabsolute-value]
                                if (abs(conf.refresh.value() - mode->vrefresh_hz) < 1.0)
                                    ^
/home/pmos/build/src/mir-1.1.2/src/miral/static_display_config.cpp:306:37: note: use function 'std::abs' instead
                                if (abs(conf.refresh.value() - mode->vrefresh_hz) < 1.0)
                                    ^~~
                                    std::abs
1 error generated.
```

AlanGriffiths:
Use function 'std::abs' (not '::abs') for double parameter. (Fixes #770)

### DIFF
--- a/src/miral/static_display_config.cpp
+++ b/src/miral/static_display_config.cpp
@@ -303,7 +303,7 @@ void miral::StaticDisplayConfig::apply_to(mg::DisplayConfiguration& conf)
                         {
                             if (conf.refresh.is_set())
                             {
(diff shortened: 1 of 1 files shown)


wmww:
Different error
bors r+

PR #784, opened 2019-04-04
For comment: prototyping more support for bespoke Wayland extensions

AlanGriffiths:
Pushed this so that people can comment on the new APIs:
```c++
miral::application_for(wl_client*);
miral::window_for(wl_resource*);
```
I've added some (rather pointless) client code that uses them. I think this is a case where writing tests will be significantly more code than this implementation, so I'd appreciate feedback before investing time in that.


include/miral/miral/wayland_extensions.h
```diff
@@ -119,6 +123,20 @@ class WaylandExtensions
...
+auto application_for(wl_client* client) -> Application;
+
+/// Get the MirAL Window for a Wayland Surface, XdgSurface, etc.
+/// Note that there may not be a corresponding miral::Window (e.g. the
+/// surface is created and assigned properties before 'commit' creates the
+/// miral::Window).
+///
+/// \return The window (null if no window is found)
+/// \remark Since MirAL 2.5
+auto window_for(wl_client* client, wl_resource* surface) -> Window;
```
wmww: 1. Is some sort of optional a thing we can return? Would do a better job signalling it can be null, and forcing those using this API into dealing with that case

2. `wl_resource_get_client()` is a thing that exists, so why require the client get sent in separately (assuming the client must always be the one the surface resource is on)

examples/mir_demo_server/generated/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+set(PROTOCOL "${CMAKE_CURRENT_SOURCE_DIR}/../protocol/server-decoration.xml")
+
+add_custom_command(
+    OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/server-decoration_wrapper.cpp
+    OUTPUT ${CMAKE_CURRENT_SOURCE_DIR}/server-decoration_wrapper.h
+    DEPENDS mir_wayland_generator
+    DEPENDS ${PROTOCOL}
+    COMMAND "sh" "-c" "${CMAKE_BINARY_DIR}/bin/mir_wayland_generator org_kde_kwin_ ${PROTOCOL} header >${CMAKE_CURRENT_SOURCE_DIR}/server-decoration_wrapper.h"
+    COMMAND "sh" "-c" "${CMAKE_BINARY_DIR}/bin/mir_wayland_generator org_kde_kwin_ ${PROTOCOL} source >${CMAKE_CURRENT_SOURCE_DIR}/server-decoration_wrapper.cpp"
+)
```
wmww: I think I wrote a macro in the other generated code CMakeLists that makes all this less repetitive and error prone. Might want to use something like that if possible.

examples/mir_demo_server/server_example.cpp
```diff
@@ -143,14 +145,22 @@ try
     InputFilters input_filters;
     me::TestClientRunner test_runner;
 
+    miral::WaylandExtensions extensions{
+        miral::WaylandExtensions::recommended_extensions()
+            + ":" + mir::examples::server_decoration_extension.name
+            + ":" + mir::examples::xdg_decoration_extension.name};
+    extensions.add_extension(mir::examples::server_decoration_extension);
+    extensions.add_extension(mir::examples::xdg_decoration_extension);
+    extensions.set_filter([&](auto, char const* protocol) { puts(protocol); return true; });
```
wmww: I *still* don't like manually constructing a colon separated string, this be better:
```
miral::WaylandExtensions extensions;
extensions.add(mir::examples::server_decoration_extension);
extensions.add(mir::examples::xdg_decoration_extension);
extensions.set_filter([&](auto, char const* protocol) { puts(protocol); return true; });
```
(with `extensions.enable(name)` and `extensions.disable(name)` if you wanted to turn on or off default or added ones)

include/miral/miral/wayland_extensions.h
RAOF: I wonder whether this wouldn't be better as
`miral::dynamic_cast<Target>(wl_resource *) -> std::experimental::optional<Target>` or moral equivalent.

As @wmww says, we don't need the `wl_client*`, as each `wl_resource` is already associated with its client.

We'll have a bunch of these accessors, and they look quite a lot like downcasts (“I know this `wl_resource` is really a `Window`, please give me the `Window`). It seems like it might be convenient to consolidate them into a downcast-like template?

RAOF: (This doesn't apply to `miral::application_for(wl_client*)`, as we'll only map a single MirAL object to a `wl_client` (?))

examples/mir_demo_server/server_example_decoration.cpp
```diff
@@ -0,0 +1,103 @@
...
+        if (mode != decoration_mode)
+            decoration_mode = mode;
+
+        send_mode_event(decoration_mode);
+    }
+
+    std::thread test{[this]
+        {
+            // We put this on a thread and sleep because we need to wait for the client
+            // to 'commit' before the Window object exists
```
RAOF: This seems like it'll be a bit of a sharp edge - particularly if extensions want to participate in window setup, like `xdg_decoration`, where the client will want to render their first committed buffer differently depending on what the shell sends.

Do we have any plan of how to address this? Will we allow `Window`s that aren't visible (due to not having an associated buffer yet)?

examples/mir_demo_server/server_example.cpp
RAOF: I'm not sure if I prefer @wmww's suggestion here, but as an API edge I think “pass in a colon-delimited string” is something we could do better than.

For my oar, maybe something like:
```
miral::WaylandExtensions extensions{
    miral::WaylandExtensions::recommended_extensions() +
    mir::examples::server_decoration_extension +
    miral::WaylandExtensions::Disabled(mir::examples::xdg_decoration_extension)};
```

include/miral/miral/wayland_extensions.h
AlanGriffiths: 1. `Window` already has `operator bool()` - doesn't that make adding an optional/fallible return type redundant?

2. That makes sense. I was unconsciously reluctant to introduce a direct dependency of libmiral on libwayland.

3. I'm not convinced of the need for a wrapper template. (The logic cannot be implemented directly in the template without exposing unstable ABIs.) But we _could_ add it later - simply add a specialization that forwards to `window_for()`.

examples/mir_demo_server/server_example.cpp
AlanGriffiths: What @wmww suggests is consistent with the approach we chose in an earlier PR (#775). This code predates that and probably shouldn't be supported.

In any case, this is isn't what I plan to land from this prototype, just experimenting with how it might be used.

examples/mir_demo_server/server_example_decoration.cpp
AlanGriffiths: This is one of the things I've been trying resolve, I don't have a perfect, general solution.

At the root is the different initialization paradigm adopted by the mirclient API and Wayland protocols. The former has a single state initialization and sends all the setup information to the server at once, the latter has a multi-stage initialization: create an object, populate it, and only then send `commit` for the server to act on it.

We do allow `Window`s that don't have a buffer, but we don't create them from the Wayland frontend as we don't otherwise know when the "populate it" stage of initialization completes. (When we added Wayland support we had bugs caused by code that didn't expect new Windows to already have a buffer.) 

It seems better to expose what we can do simply than hope for a swift rework of this logic.

AlanGriffiths:
OK, thanks for the feedback, will propose a "real thing".

PR #786, opened 2019-04-08
LTTNG: Generate USDT tracepoints from LTTNG tracepoints.

RAOF:
Userspace Statically Defined Tracepoints are approximately zero-cost (the introduce a single
noop into the codestream + some ELF metadata in the DSO) and can be consumed with various
tracing frameworks, most notably ftrace & bpftrace.

LTTNG has support for generating a USDT for each LTTNG tracepoint, but only if such support is enabled.

### DIFF
--- a/src/server/report/lttng/CMakeLists.txt
+++ b/src/server/report/lttng/CMakeLists.txt
@@ -9,6 +9,16 @@ add_compile_options(
   -Wno-error=unused-parameter
 )
 
+# Add USDT hooks to all LTTNG tracepoints
+#
+# This might be gnarly; this tweaks an LTTNG header to enable its
+# USDT integration. The integration only involves #include-ing a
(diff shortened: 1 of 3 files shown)


src/server/report/lttng/CMakeLists.txt
```diff
@@ -9,6 +9,16 @@ add_compile_options(
   -Wno-error=unused-parameter
 )
 
+# Add USDT hooks to all LTTNG tracepoints
+#
+# This might be gnarly; this tweaks an LTTNG header to enable its
+# USDT integration. The integration only involves #include-ing a
+# SystemTap header and calling a macro from it in the LTTNG macro,
+# but there aren't guarantees this won't change in future.
+add_compile_definitions(
```
AlanGriffiths: That needs CMake 3.12 or later. For 3.5 (Xenial) you need `add_definitions()`.

PR #787, opened 2019-04-08
Add utility functions to map from Wayland types to MirAL types

AlanGriffiths:


wmww:
bors try

tests/miral/org_kde_kwin_server_decoration.c
```diff
@@ -0,0 +1,71 @@
+/* Generated by wayland-scanner 1.16.0 */
```
wmww: I get this is a test, but even so why is there a generated C file?

AlanGriffiths: Simpler than ensuring `wayland-scanner` is available and generating it every time. There's a `.h` too.

wmww: Sounds good

PR #788, opened 2019-04-09
LTTNG: Void tracepoints don't need a dummy field.

RAOF:
We have a bunch of no-context tracepoints - things like compositor::started(), which
have no useful context to add to their bare emission.

LTTNG is quite happy to have tracepoints with no fields, but for some reason we were
creating tracepoints with a dummy integer field and always setting it to 0.

Just use a tracepoint with no fields instead.


RAOF:
Ah. This obviously changed between LTTNG 2.7 (in 16.04) and LTTNG 2.10 (in 18.04).

Close it for now, and re-visit once 16.04 is no more.

Issue #796, opened 2019-04-17
Experimental X11 support: Mir blocks on exit if an X11 application is running


AlanGriffiths:
```plain
mir_demo_server --x11-display 4 --startup xeyes
```
Press `Ctrl+Alt+BkSp`

Expect: mir_demo_server (and xeyes) exit
Actual: mir_demo_server closes its "wallpaper" and xeyes stops reacting to cursor 

_Note: this is probably better reproduced on trunk_

PR #797, opened 2019-04-17
Make sure Xwayland exits to ensure a clean shutdown

AlanGriffiths:
Make sure Xwayland exits to ensure a clean shutdown. (Fixes: #796)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -64,7 +64,12 @@ mf::XWaylandServer::~XWaylandServer()
       if (xserver_status == RUNNING)
         wm->destroy();
 
-      kill(pid, SIGTERM);
+      if (kill(pid, SIGTERM) == 0)
+      {
+          usleep(100000);           // After 100ms...
+          if (kill(pid, 0) == 0)    // ...if Xwayland is still running...
+            kill(pid, SIGKILL);     // ...then kill it!
+      }
       spawn_thread->join();
     }
 


src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -64,7 +64,12 @@ mf::XWaylandServer::~XWaylandServer()
       if (xserver_status == RUNNING)
         wm->destroy();
 
-      kill(pid, SIGTERM);
+      if (kill(pid, SIGTERM) == 0)
+      {
+          usleep(100000);           // After 100ms...
+          if (kill(pid, 0) == 0)    // ...if Xwayland is still running...
+            kill(pid, SIGKILL);     // ...then kill it!
```
wmww: Is there a known reason SIGTERM isn't working?

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -64,7 +64,12 @@ mf::XWaylandServer::~XWaylandServer()
...
+      }
```
wmww: Upon failure of `kill()`, shouldn't we at least log an error?

AlanGriffiths: If the process has already exited, we'll get an "error", but why would we need to log it?

wmww: I see, never mind

src/server/frontend_xwayland/xwayland_server.cpp
AlanGriffiths: Empirically, Xwayland doesn't exit while there's an active X11 client.

wmww: I question if a 0.1s delay to shutdown in all cases is an ideal solution. If SIGTERM is expected to often fail, and SIGKILL is harmless, why not just do that in the first place?

RAOF: Of course, there's no guarantee that `pid` still refers to the xwayland process… :grin: 

Unfortunately, this is wandering into the parts of the system where the API is so bad it is *impossible* to do the right thing. The *right* thing to do is to monitor the process we spawn and wait for it to die, but that requires signal handling that we can't rely on working because we're a library.

SIGTERM→SIGKILL seems like a reasonable kludge for now.

RAOF: `systemd-run` is pretty close to useful, but not quite.

AlanGriffiths: It is indeed possible that the pid has been reused, and even possible we're allowed to send signals to the new process.

But while racy it isn't as bad as it might be: The thread we're about to join _does_ use signal handling (via `waitpid()`) to monitor the Xwayland process and sets `xserver_status` (which we've just read) accordingly.

The code in this component needs further tidy up, which is why we default to disabling it behind an "experimental" configuration option. The plan is to incrementally improve it to the point where things reliably work as expected. Hopefully, we will be able to identify X11 clients properly to ask them politely to shut down before SIGKILLing anything.

wmww:
It works, and looks like this is the best we're going to get for now

bors r+

PR #800, opened 2019-04-22
Support multi-output in Mir-onX

wmww:
Enhance the `--x11-output` option to support a colon separated list of output sizes. Outputs are lined up horizontally from Mir's perspective, and are each shown in a different X11 window. Useful if you need to test multi-output code paths and do not have auxiliary physical display handy.

### DIFF
--- a/src/platforms/mesa/server/x11/X11_resources.cpp
+++ b/src/platforms/mesa/server/x11/X11_resources.cpp
@@ -21,6 +21,7 @@
 
 #include "X11_resources.h"
 
+namespace mg=mir::graphics;
 namespace mx = mir::X;
 
 //Force synchronous Xlib operation - for debugging
@@ -56,4 +57,23 @@ std::shared_ptr<::Display> mx::X11Resources::get_conn()
 #endif
     connection = new_conn;
     return new_conn;
+}
+
+void mx::X11Resources::set_output_config_for_win(Window win, mg::DisplayConfigurationOutput* configuration)
+{
+    output_configs[win] = configuration;
+}
+
+void mx::X11Resources::clear_output_config_for_win(Window win)
+{
+    output_configs.erase(win);
+}
+
+std::experimental::optional<mg::DisplayConfigurationOutput*> mx::X11Resources::get_output_config_for_win(Window win)
+{
+    auto configuration = output_configs.find(win);
+    if (configuration != output_configs.end() && configuration->second != nullptr)
+        return {configuration->second};
+    else
+        return std::experimental::nullopt;
 }
--- a/src/platforms/mesa/server/x11/X11_resources.cpp
+++ b/src/platforms/mesa/server/x11/X11_resources.cpp
@@ -38,7 +38,7 @@ int mx::mir_x11_error_handler(Display* dpy, XErrorEvent* eev)
     return 0;
 }
 
-std::shared_ptr<::Display> mx::X11Resources::get_conn()
+auto mx::X11Resources::get_conn() -> std::shared_ptr<::Display>
 {
     if (auto conn = connection.lock())
         return conn;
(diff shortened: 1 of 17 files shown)


src/platforms/mesa/server/x11/X11_resources.cpp
```diff
@@ -55,3 +58,22 @@ std::shared_ptr<::Display> mx::X11Resources::get_conn()
...
+{
+    output_configs[win] = configuration;
+}
+
+void mx::X11Resources::clear_output_config_for_win(Window win)
+{
+    output_configs.erase(win);
+}
+
+std::experimental::optional<mg::DisplayConfigurationOutput*> mx::X11Resources::get_output_config_for_win(Window win)
```
RAOF: If you're going to return a `std::experimental::optional`, I'd mildly prefer `std::experimental::optional<mg::DisplayConfigurationOutput&>` to the pointer version.

src/platforms/mesa/server/x11/graphics/display.cpp
```diff
@@ -231,39 +234,50 @@ unsigned long mgx::X11Window::red_mask() const
...
-                         *gl_config);
+    for (auto const& requested_size : requested_sizes)
+    {
+        auto actual_size = clip_to_display(x_dpy, requested_size);
+        auto window = std::make_unique<X11Window>(x_dpy, shared_egl.display(), actual_size, shared_egl.config());
+        auto red_mask = window->red_mask();
+        auto pf = (red_mask == 0xFF0000 ?
+            mir_pixel_format_argb_8888 :
+            mir_pixel_format_abgr_8888);
+        auto configuration = static_cast<std::unique_ptr<graphics::DisplayConfigurationOutput>>(
```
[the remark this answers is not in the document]
wmww: Oops. I ended up with an unnecessary `graphics::X::DisplayConfigurationOutput` subclass that needed to be cast from. Refactored that into a static builder method `graphics::X::DisplayConfiguration`, which cleans this up.

src/platforms/mesa/server/x11/X11_resources.cpp
wmww: I don't think that's allowed. When I try, I get
```
/home/wmww/code/mir/src/platforms/mesa/server/x11/X11_resources.cpp:72:122:   required from here
/usr/include/c++/8/experimental/optional:461:15: error: non-static data member ‘std::experimental::fundamentals_v1::_Optional_base<const mir::graphics::DisplayConfigurationOutput&, false>::<unnamed union>::_M_payload’ in a union may not have reference type ‘std::experimental::fundamentals_v1::_Optional_base<const mir::graphics::DisplayConfigurationOutput&, false>::_Stored_type’ {aka ‘const mir::graphics::DisplayConfigurationOutput&’}
```
I also found [this](https://foonathan.net/blog/2018/07/12/optional-reference.html) on the topic.

wmww:
Added tests and cleared up the ownership issues (hopefully). The reason I use a global map is because events get a `Window`, and I didn't find a way to attach use data onto that. Now the values windows map to are const weak pointers, so maybe better?

AlanGriffiths:
The layout of the X windows is a bit odd:

- The first window is placed top left of the *current* output.
- Each subsequent window is placed to the right of the previous window **unless** it would extend outside the *first* output, in which case it is placed on top of the previous window.

So, if I'm working on my right monitor, all the windows pile up on top of each other.
If I'm working on my left monitor, the first couple of windows tile nicely and the rest pile up on top of each other.

This isn't a blocker, but it would make more sense to use the entire display area.

wmww:
@AlanGriffiths are you talking about how the "output" windows are placed within the parent desktop environment? If so, I don't believe we're exerting any control of that client-side (I certainly didn't touch any window moving code). X obviously gives us enough control to place windows how we like, but that would be an entirely new feature (which I imagine we'd agree isn't worth it).

AlanGriffiths:
> @AlanGriffiths are you talking about how the "output" windows are placed within the parent desktop environment? If so, I don't believe we're exerting any control of that client-side (I certainly didn't touch any window moving code). X obviously gives us enough control to place windows how we like, but that would be an entirely new feature (which I imagine we'd agree isn't worth it).

Yeah, I didn't notice any code for it, but the different behaviour depending on different outputs made me wonder. (As I said, "non-blocking".)

RAOF:
More obvious ownership, *and* tests? Be still my beating heart!

I don't *think* we're doing anything that would make it less likely a decent window manager would position our windows well. GNOME Shell should be doing a better job, here.

bors r+

PR #802, opened 2019-04-26
Use generated protocol names

wmww:
Use the protocol names from generated code instead of hardcoding another copy here


wmww:
Oh, this is wrong.

PR #803, opened 2019-04-29
Fix clang 9 -Wdefaulted-function-deleted

RAOF:


RAOF:
clang-9 has started warning about a couple of pieces of Mir code, for example:
```
../include/miral/miral/window_management_options.h:56:5: error: explicitly defaulted default constructor is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
    WindowManagerOptions() = default;
    ^
../include/miral/miral/window_management_options.h:61:44: note: default constructor of 'WindowManagerOptions' is implicitly deleted because field 'policies' of const-qualified type 'const std::vector<WindowManagerOption>' would not be initialized
    std::vector<WindowManagerOption> const policies;
```

It's simple to fix them, so do.

include/miral/miral/window_management_options.h
```diff
@@ -53,7 +53,7 @@ inline auto add_window_manager_policy(std::string const& name, Args&... args) ->
 class WindowManagerOptions
 {
 public:
-    WindowManagerOptions() = default;
+    WindowManagerOptions();
```
AlanGriffiths: This looks like the wrong fix. I suggest:
```c++
    WindowManagerOptions() = delete;
```
as that does not change the meaning of the code.

RAOF: Hah! The warning proves valuable already. I thought the intent of `WindowManagerOptions() = default` was to make a no-parameter constructor available :grin:

AlanGriffiths: That is a natural reading. But as clang told you that's not what it does, and if you look at the class semantics it doesn't make sense either.

PR #812, opened 2019-05-03
Harmless formatting

wmww:
Just some formatting and the removal of an unused function. I've had these things sitting around for a while so I might as well PR them now.

### DIFF
--- a/src/server/compositor/multi_monitor_arbiter.cpp
+++ b/src/server/compositor/multi_monitor_arbiter.cpp
@@ -85,12 +85,6 @@ bool mc::MultiMonitorArbiter::buffer_ready_for(mc::CompositorID id)
        ((current_buffer_users.find(id) == current_buffer_users.end()) && current_buffer);
 }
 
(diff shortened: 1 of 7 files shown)


src/server/compositor/multi_monitor_arbiter.cpp
```diff
@@ -48,8 +48,10 @@ std::shared_ptr<mg::Buffer> mc::MultiMonitorArbiter::compositor_acquire(composit
     if (current_buffer_users.find(id) != current_buffer_users.end() || !current_buffer)
     {
         if (schedule->num_scheduled())
+        {
             current_buffer = schedule->next_buffer();
-        current_buffer_users.clear();
+            current_buffer_users.clear();
+        }
```
AlanGriffiths: This isn't just formating.

wmww: Correct you are. Like I said, this branch was old and these lines slipped my review. I believe the change to be correct, so I've moved the commit to another branch which I will PR separately later.

PR #814, opened 2019-05-06
Use drmIsMaster() from libdrm when it is provided

RAOF:
Closes: #813

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -282,6 +282,7 @@ pkg_check_modules(XCB_COMPOSITE REQUIRED xcb-composite)
 pkg_check_modules(XCB_XFIXES REQUIRED xcb-xfixes)
 pkg_check_modules(XCB_RENDER REQUIRED xcb-render)
 pkg_check_modules(X11_XCURSOR REQUIRED xcursor)
+pkg_check_modules(DRM REQUIRED libdrm)
 
 include_directories (SYSTEM ${GLESv2_INCLUDE_DIRS})
 include_directories (SYSTEM ${EGL_INCLUDE_DIRS})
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -282,11 +282,18 @@ pkg_check_modules(XCB_COMPOSITE REQUIRED xcb-composite)
 pkg_check_modules(XCB_XFIXES REQUIRED xcb-xfixes)
 pkg_check_modules(XCB_RENDER REQUIRED xcb-render)
 pkg_check_modules(X11_XCURSOR REQUIRED xcursor)
+pkg_check_modules(DRM REQUIRED libdrm)
```
AlanGriffiths: Will requiring drm cause problems for UBports builds targetting libhybris/android? @mariogrip

RAOF: As I mentioned in the first commit message, we already [unconditionally require libdrm](https://github.com/MirServer/mir/blob/34f92104efea90aaa79f90d9acd5b9f638f3d21f/src/server/console/minimal_console_services.cpp#L32
); we just don't unconditionally check for it.

AlanGriffiths: Ack. Makes sense.

src/server/console/minimal_console_services.cpp
```diff
@@ -37,7 +37,7 @@
 
 // TODO: once libdrm provides this symbol, we need a way to set MIR_LIBDRM_HAS_IS_MASTER
```
AlanGriffiths: Comment needs updating to reflect these changes.

wmww:
bors r=AlanGriffiths

PR #815, opened 2019-05-07
Add Z-index property to Mir surfaces

wmww:
Surfaces with higher Z-indices will always appear above surfaces with lower ones. This is useful in implementing Layer Shell, but may have other applications as well. All the interesting logic changes are in `surface_stack.cpp`.

### DIFF
--- a/include/server/mir/scene/null_surface_observer.h
+++ b/include/server/mir/scene/null_surface_observer.h
@@ -54,6 +54,7 @@ public:
     void placed_relative(Surface const* surf, geometry::Rectangle const& placement) override;
     void input_consumed(Surface const* surf, MirEvent const* event) override;
     void start_drag_and_drop(Surface const* surf, std::vector<uint8_t> const& handle) override;
+    void z_index_set_to(Surface const* surf, int z_index) override;
 
 protected:
     NullSurfaceObserver(NullSurfaceObserver const&) = delete;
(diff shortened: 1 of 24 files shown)


RAOF:
Hm. I think I'm going to strengthen my “z-order should be pre-constructed layer” opinion. One of the things I want to do in the mid-term future is support dropping in a new `Renderer`; this will pretty much require that the `Renderer` has a scenegraph, and *ideally* this scenegraph would be the definitive one - that is, `SurfaceStack` and the `Renderer`'s scenegraph would both implement the relevant interface, and when the `Renderer` is replaced Mir would use the `Renderer`'s implementation.

It's easy to support explicit layers in a general scenegraph. It's rather more difficult to support arbitrary `z_order` as in this PR - you basically need to reconstruct a set of layers out of the z-order.

So, in addition to being semantically preferable, I think that making the layers explicit and up-front would make future implementations easier, too.

AlanGriffiths:
I concur with @RAOF - this is both the wrong name and the wrong abstraction.

There's also history to this thinking: We [removed a broken "depth ID"](https://github.com/MirServer/mir/commit/74ee479a40be7c8d3fa343acfd439dad245ec2ef) from the scenegraph some years ago.

The lesson is that we want to get the abstraction right this time. And give it a good name.

PR #821, opened 2019-05-09
Disable rawhide builds in CI (until we have a fix)

AlanGriffiths:
Rawhide builds are failing: "Package 'glesv2', required by 'virtual:world', not found".

Disable rawhide builds until we have a fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,11 @@ env:
   - SYSTEM=google:ubuntu-18.10-64 VARIANT=amd64
   - SYSTEM=google:fedora-29-64 VARIANT=amd64
 
+matrix:
+  allow_failures:
+    - env: SYSTEM=google:fedora-rawhide-64 VARIANT=amd64
+  fast_finish: true
+
 before_install:
 - mkdir -p ${SPREAD_PATH}
 - pushd "${SPREAD_PATH}"


spread.yaml
```diff
@@ -25,7 +25,7 @@ backends:
             - ubuntu-devel-64:
                 image: ubuntu-os-cloud/ubuntu-1810
             - fedora-29-64
-            - fedora-rawhide-64
+#            - fedora-rawhide-64
```
Saviq: No need for this, since it won't be run in CI.

.travis.yml
```diff
@@ -44,7 +44,7 @@ env:
   - SPREAD_PATH=/tmp/spread
   matrix:
   - SYSTEM=google:ubuntu-devel-64 VARIANT=amd64
-  - SYSTEM=google:fedora-rawhide-64 VARIANT=amd64
+#  - SYSTEM=google:fedora-rawhide-64 VARIANT=amd64
```
Saviq: How about allowing it to fail instead? https://github.com/MirServer/mir/commit/77c4ef82a7c997d76dd1259b4d2c1c0fc8f54e58

AlanGriffiths: How's this?

Issue #816, opened 2019-05-07
[miral] We only handle move and resize requests based on pointer events


AlanGriffiths:
This is only the "example" code, but we should handle touch also.

PR #824, opened 2019-05-09
Provide a new MinimalWindowManager strategy

### LINKED ISSUE #761: Incomplete window manager used for WLCS
At https://github.com/MirServer/mir/blob/ab2f9c149371a7dc1f03033b731ab137be4a5a52/tests/miral/test_server.cpp#L87 we use a `TestWindowManagerPolicy` for WLCS. This window manager doesn't handle moving or resizing properly, which inhibits testing of those features (why [`XdgToplevelStableTest.interactive_move`](https://github.com/MirServer/wlcs/blob/d4d99e63482fed379c790cb5696f2466b34c81f6/tests/xdg_toplevel_stable.cpp#L166) is disabled)

AlanGriffiths:
Provide a new MinimalWindowManager strategy. (Fixes #761, #816)

With this the MinimalWindowManager is only used by `mir_demo_server` and tests. Will follow up by integrating with the example shells.

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -564,6 +564,56 @@ void miral::BasicWindowManager::focus_next_application()
     select_active_window(focussed_surface ? info_for(focussed_surface).window() : Window{});
 }
 
+void miral::BasicWindowManager::focus_prev_application()
+{
+    if (auto const prev = active_window())
+    {
+        auto const workspaces_containing_window = workspaces_containing(prev);
+
+        if (!workspaces_containing_window.empty())
+        {
+            do
+            {
+                focus_controller->focus_prev_session();
+
+                if (can_activate_window_for_session_in_workspace(
+                    focus_controller->focused_session(),
+                    workspaces_containing_window))
+                {
+                    return;
+                }
+            }
+            while (focus_controller->focused_session() != prev.application());
+        }
+        else
+        {
+            do
+            {
+                focus_controller->focus_prev_session();
+
+                if (can_activate_window_for_session(focus_controller->focused_session()))
+                {
+                    return;
+                }
+            }
+            while (focus_controller->focused_session() != prev.application());
+        }
+    }
+    else
+    {
+        focus_controller->focus_prev_session();
+
+        if (can_activate_window_for_session(focus_controller->focused_session()))
+        {
+            return;
+        }
+    }
+
+    // Last resort: accept wherever focus_controller places focus
+    auto const focussed_surface = focus_controller->focused_surface();
+    select_active_window(focussed_surface ? info_for(focussed_surface).window() : Window{});
+}
+
 auto miral::BasicWindowManager::workspaces_containing(Window const& window) const
 -> std::vector<std::shared_ptr<Workspace>>
 {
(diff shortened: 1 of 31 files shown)


AlanGriffiths:
Nice! With this PR:
```
miral-test --gtest_filter=DragAndDrop.after_drag_finishes_pointer_events_no_longer_contain_handle
```
Fails _when built with clang_.

I can reproduce, but how does that work?

src/miral/basic_window_manager.cpp
```diff
@@ -564,6 +564,56 @@ void miral::BasicWindowManager::focus_next_application()
...
+        if (can_activate_window_for_session(focus_controller->focused_session()))
+        {
+            return;
+        }
+    }
+
+    // Last resort: accept wherever focus_controller places focus
+    auto const focussed_surface = focus_controller->focused_surface();
+    select_active_window(focussed_surface ? info_for(focussed_surface).window() : Window{});
+}
```
wmww: Looks like there's some code duplication in this function with `focus_next_application()`. Would it make sense to merge the two? (perhaps keep the names for ABI compat but implement them each as just a call to a new function: `switch_focus_application(bool)`)

src/server/scene/session_container.cpp
```diff
@@ -82,3 +83,27 @@ std::shared_ptr<ms::Session> ms::SessionContainer::successor_of(std::shared_ptr<
...
+        return std::shared_ptr<Session>();
+
+    for (auto it = apps.rbegin(); it != apps.rend(); it++)
+    {
+        if (*it == session)
+        {
+            auto successor = ++it;
+            if (successor == apps.rend())
+                return *apps.rbegin();
+            else return *successor;
```
wmww: Like `focus_prev_application`, *maybe* this needs a whole separate implementation from `successor_of`, but I'd like to hear why.

src/miral/basic_window_manager.cpp
AlanGriffiths: Well, we _could_ pass a `void FocusController::(*focus_some_session)()` pointer into a shared implementation function, but I don't think we'd gain much clarity. (Or use a template I suppose.)

src/server/scene/session_container.cpp
AlanGriffiths: There are ways to factor out the difference between `begin()` and `rbegin()` etc. but I don't see that the code would be more reliable or easier to follow.

src/server/scene/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ ADD_LIBRARY(
   basic_surface.cpp
   broadcasting_session_event_sink.cpp
   default_configuration.cpp
-  default_session_container.cpp
+        session_container.cpp
```
RAOF: Nit: indentation.

AlanGriffiths:
It has been long enough for anyone to comment:

bors r=wmww

src/miral/basic_window_manager.cpp
RAOF: There seems to be a *reasonable* amount of logic in these, and it's trivially duplicated; it would be *nice* to consolidate that.

I won't block on that, but maybe it's something to explore?

src/server/scene/session_container.cpp
RAOF: In contrast to `focus_{next,previous}_application()` there's basically no logic here. I don't think there's anything to be gained trying to factor this out.

AlanGriffiths:
If anyone still cares, they can PR a consolidation of the `focus_{next,previous}_application()` implementations.

bors r=wmww

PR #828, opened 2019-05-15
Add base objects for Wayland objects and globals

wmww:
Adds new non-generated classes that generated code inherits from


src/wayland/base/wayland_base.h
```diff
@@ -0,0 +1,59 @@
...
+namespace wayland
+{
+
+class Object
+{
+public:
+    Object(wl_resource* resource);
+    virtual ~Object() = default;
+
+    void destroy_wayland_object() const;
```
RAOF: Does destroying the `wl_resource` *really* not change the state of the `mw::Object`? :wink:

src/wayland/base/wayland_base.h
```diff
@@ -0,0 +1,59 @@
...
+
+    wl_client* const client;
```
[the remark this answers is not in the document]
wmww: I noticed this, but preexisting. This PR isn't supposed to change the Wayland objects, just move some of the bits into a base class.

tests/miral/generated/server-decoration_wrapper.h
```diff
@@ -13,29 +13,24 @@
...
 namespace mir
 {
 namespace wayland
 {
 
-class ServerDecorationManager
+class ServerDecorationManager : public Object
 {
 public:
     static char const constexpr* interface_name = "org_kde_kwin_server_decoration_manager";
```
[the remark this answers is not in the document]
wmww: Dammit C++. Will address in a later PR

src/wayland/base/wayland_base.h
```diff
@@ -0,0 +1,59 @@
...
+
+namespace mir
+{
+namespace wayland
+{
+
+class Object
+{
+public:
+    Object(wl_resource* resource);
```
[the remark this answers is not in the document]
wmww: Done here, but I didn't change in all the generated code. I'll do that later.

src/wayland/base/wayland_base.h
wmww: Preexisting, but removing const didn't break anything so I did.

wmww:
There may be cases where we want to pass around `mw::Object`s or `mw::Global`s (the latter may be especially useful for out protocol enabling/disabling system), but this PR is mainly about moving shared functionality out of generated code. If there's something that's the same for every type, keeping it in a single non-generated class has lots of advantages. Not only can it be used generically, but it is also easier to modify, easier to review, faster to compile and easier to understand. As far as I can tell, there are no real advantages to generating 50 copies of the same code.

As for why inheritance was used, it is both logical for the situation (all the Wayland types *are* wayland objects, with additional functionality) and the most straightforward way of ripping common data and functions out of the generated classes with minimal code churn to the rest of the project.

AlanGriffiths:
> I'm uncertain about the purpose of this. When would code want to operate on a variable of type `mw::Object&`? Under what circumstances would you write a function that takes a `mw::Global&` parameter?
> 
> Inheritance is generally not the best mechanism for code reuse, and this isn't really reusing much code, anyway.

I don't have a particular objection to a base class: I think there's code where we have a collection of `shared_ptr<void>` when, with a common base, we could use unique_ptr<Global>.. (Not entirely convinced by the name "Object".)

But using it to share (constant) state is less convincing unless it supports the LSP.,

RAOF:
So, I'm not *averse* to a base class; I just don't know what this one does/is :grin:.

For contrast, when I have a `mw::Global` I know that I've got a handle to a valid Wayland protocol global. But as far as I can tell, a `mw::Object` is just a slightly obfuscated `wl_resource*`; if I've got a `mw::Object` I've got access to a `wl_resource*`, but I've got no idea whether it's a valid Wayland resource. Notably, `mw::Object` does *not* factor out any code; `destroy_wayland_resource()` is just a rename of `wl_resource_destroy()`, something that anyone who has a `mw::Object` could call directly.

Furthermore, this doesn't really seem like an intermediate step to having an `mw::Object` which *is* more than a `wl_resource*`.

I think if we're going to introduce a `mw::Object` it should be something to resolve the Wayland object lifetime issues we have; either a `mw::ResourceRef` smart-pointerish thing, or  `mw::ResourceRef::do_with_resource(std::function<void(wl_resource*)>)`

wmww:
There seems to be some differences of understanding on what the *point* of this abstraction is. Here are the benefits of merging this PR as I see them:

* It is the *right* abstraction
  * Conceptually, Wayland objects are all different but in some ways similar. This relationship maps perfectly to single inheritance (this alone is not that big a reason, but it's relevant)
  * It is a wrapper around a `wl_resource*` because libmirwayland is a wrapper around libwayland, and we choose the same abstraction
* We get to pass around pointers and refs to `mw::Object` and `mw::Global`
  * The usefulness of this has been debated here
  * I'm not sure about it, that was never the main purpose
* The only code that should be generated is code that needs to be generated, not many copies of the same things
  * Reading the wrappers is more time consuming and intimidating the more code is there. You eventually pick up on patterns of what is unique to each interface, but trying to guess that can also lead to missing important details
  * You can read the base class once and know exactly what is common to every Wayland object
  * This applies to us implementing interfaces, us reviewing changes to libmirwayland and anyone who's trying to understand the code
* More code will belong in the base class in the future, and this lays the foundation for those improvements
  * Lifetime management is the biggest one that comes to mind (I disagree, and think this is very much an intermediate step to that)

wmww:
I synced with @RAOF last night. We decided we would do this, but leave some of the more problematic bits in the generated code until we're actually ready to deal with them. I moved `resource`, `client` and `destroy_wayland_resource()` back to the generated code, and renamed `wayland::Object` to `wayland::Resource`.

src/wayland/base/wayland_base.h
```diff
@@ -0,0 +1,55 @@
...
+namespace mir
+{
+namespace wayland
+{
+
+class Resource
+{
+public:
+    Resource();
+    virtual ~Resource() = default;
```
AlanGriffiths: For consistency I think we should `delete` CopyAssign semantics.

src/wayland/generated/wayland_wrapper.h
```diff
@@ -13,16 +13,17 @@
 #include "mir/fd.h"
 #include <wayland-server-core.h>
 
+#include "wayland_base.h"
```
AlanGriffiths: Having generated code dependent on a header we don't publish (src/wayland/base/wayland_base.h) is going to make things difficult for bespoke Wayland protocol writers.

src/wayland/CMakeLists.txt
```diff
@@ -19,6 +24,7 @@ target_link_libraries(mirwayland
 
 target_include_directories(mirwayland
   PUBLIC
+    ${CMAKE_CURRENT_SOURCE_DIR}/base
```
AlanGriffiths: Simpler to place files directly in `src/wayland`

Although the header should probably be in include/wayland/mir/wayland (along with adding this directory to install and packaging).

src/wayland/base/wayland_base.h
wmww: Done

src/wayland/generated/wayland_wrapper.h
wmww: Done

src/wayland/CMakeLists.txt
wmww: Done

src/wayland/CMakeLists.txt
```diff
@@ -19,6 +24,7 @@ target_link_libraries(mirwayland
 
 target_include_directories(mirwayland
   PUBLIC
+    ${PROJECT_SOURCE_DIR}/include/wayland
     ${CMAKE_CURRENT_SOURCE_DIR}/generated
```
AlanGriffiths: Anything under ${CMAKE_CURRENT_SOURCE_DIR} should be PRIVATE

wmww:
> Should Global be a type of Resource?
No. They are different concepts, and share no functionality

> I'm no longer sure of the purpose of mir::wayland::Resource
I just want to get the architecture in place

> Should this wait for a PR that gives it a function?
Pros of merging now:
* (Slight) net reduction in LoC
* Easier to review now than when there's more stuff
* We can discuss things like file placement, CMake changes, class names, etc here and get it decided instead of bogging down the conversation in a more difficult PR to land
* Easier to resolve merge conflicts or no merge conflicts when adding multiple features separately that live in `Resource`
Cons:
* ???

include/wayland/mir/wayland/wayland_base.h
```diff
@@ -0,0 +1,62 @@
...
+struct wl_resource;
+struct wl_global;
+struct wl_client;
+
+namespace mir
+{
+namespace wayland
+{
+
+class Resource
```
AlanGriffiths: Note that `Resource` and `Global` are part of a published API and ABI that will be implemented by client code.

This means we cannot add data members or virtual member functions without breaking ABI. Are we confident that this will be a sufficiently rare occurrence to be supportable?

src/wayland/CMakeLists.txt
```diff
@@ -36,4 +42,6 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/mirwayland.pc.in
 
 install(TARGETS     mirwayland                           LIBRARY DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}")
 install(FILES       ${CMAKE_CURRENT_BINARY_DIR}/mirwayland.pc    DESTINATION "${CMAKE_INSTALL_FULL_LIBDIR}/pkgconfig")
+install(DIRECTORY   ${CMAKE_SOURCE_DIR}/include/wayland/mir      DESTINATION "include/mirwayland")
 
+set(INCLUDEDIR "${CMAKE_INSTALL_PREFIX}/include/mirwayland")
```
AlanGriffiths: We need to add to the include file(s) to debian/libmirwayland-dev.install, or packaging will fail.

src/wayland/CMakeLists.txt
wmww: Preexisting, and making is private it breaks everything. Not the place to fix it.

AlanGriffiths: On reflection this is OK. But means we could accidentally write examples that depend on `${CMAKE_CURRENT_SOURCE_DIR}/generated`.

include/wayland/mir/wayland/wayland_base.h
wmww: In the immediate future: no

Hopefully once we get the life-cycle situation sorted out it will be stable enough

src/wayland/CMakeLists.txt
wmww: Done

include/wayland/mir/wayland/wayland_base.h
AlanGriffiths: > Hopefully we can do that before next release.

This rework is the only thing stopping me rolling out the release. 8^)

wmww: As said in IRC, we should probably just release while this is still unstable, and get it good by next release.

AlanGriffiths: Having asked around, we'll do as you suggest. But I'd like to leave this header out of the release. So to make things easy I'll mark this "No Merge" until I've branched,

PR #832, opened 2019-05-17
Force support for latest Wayland interface versions

wmww:
This will cause a compile-time error if XML/generated classes are updated to a new version without also updating the child classes.

### DIFF
--- a/src/wayland/base/wayland_base.h
+++ b/src/wayland/base/wayland_base.h
@@ -28,6 +28,11 @@ namespace mir
 namespace wayland
 {
 
+template<int V>
+struct Version
+{
+};
+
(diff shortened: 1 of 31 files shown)


src/wayland/base/wayland_base.h
```diff
@@ -28,6 +28,11 @@ namespace mir
 namespace wayland
 {
 
+template<int V>
+struct Version
+{
+};
+
```
AlanGriffiths: I'd make this template a member of `Resource` (and `Global`) - it doesn't really have a use in its own right, and it simplifies the user code: `s/mw::Version/Version`.

Issue #799, opened 2019-04-21
Shells inconsistantly crash when output width is much larget than height


wmww:
`miral-shell --x11-output 800x100` usually crashes. Less extreme aspect ratios crash less often, and by the time it gets to 2:1 they go away. There does not seem to be a problem with tall skinny outputs. `mir_demo_server` does the same thing.

AlanGriffiths:
Is this only on shells with text on wallpaper? If `--window-manager tiling` and `miral-kiosk` are unaffected then check the wallpaper text sizing logic.

wmww:
Yes, both are unaffected and gdb says the crash happens at `/home/wmww/code/mir/examples/example-server-lib/decoration_provider.cpp:209`. I'll take care of this next week

PR #833, opened 2019-05-20
Don't let help text get too high for the output

AlanGriffiths:
Don't let help text get too high for the output. (Fixes: #799)

### DIFF
--- a/examples/example-server-lib/decoration_provider.cpp
+++ b/examples/example-server-lib/decoration_provider.cpp
@@ -159,7 +159,7 @@ void Printer::printhelp(BackgroundInfo const& region)
 
         auto const line = converter.from_bytes(rawline);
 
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
> After looking through this terrible, terrible function for several minutes I don't feel any closer to understanding why this stops the crash, much less verifying it will eradicate this memory error in all cases. Instead of adding more magic numbers, lets just check where we write to the buffer.

Restricting the width will reduce the character size including the height. Admittedly, someone could still select a typeface with absurdly high characters and cause a crash, but anything "normal" should remain within 80% of the screen.

If you have a less terrible, terrible way to implement this function feel welcome to PR. I'm not impressed by the freetype2 API, but maybe I, and the author of the code I based this on, are missing something.

AlanGriffiths:
Retiring in favor of #834

PR #834, opened 2019-05-20
Improve miral shell help text rendering (Fixes #799)

### LINKED ISSUE #799: Shells inconsistantly crash when output width is much larget than height

wmww:
Improve miral shell help text rendering. (Fixes #799)

### DIFF
--- a/examples/example-server-lib/decoration_provider.cpp
+++ b/examples/example-server-lib/decoration_provider.cpp
@@ -38,7 +38,7 @@
 #include <system_error>
 #include <iostream>
 
-namespace geom = mir::geometry;
+using namespace mir::geometry;
 
 namespace
 {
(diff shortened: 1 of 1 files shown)


wmww:
Fixes #799

wmww:
See also #833

examples/example-server-lib/decoration_provider.cpp
```diff
@@ -38,6 +38,8 @@
 #include <system_error>
 #include <iostream>
 
+namespace geom = mir::geometry;
```
AlanGriffiths: My current preference is for `using namespace mir::geometry`, but that's non-blocking.

wmww:
Thanks for the improvements!

PR #843, opened 2019-05-28
WaylandExecutor: Eagerly execute work when on Wayland thread.

RAOF:
This would appear to be a fairly minor optimisation, as the Wayland thread
was going to dispatch the Executor (and hence perform this work) before
going to sleep anyway.

However, because Wayland events are processed in-order, getting them
in the stream *now*, rather than after the rest of the events are processed/
sent, can make a reasonable difference in client behaviour.

### DIFF
--- a/src/server/frontend_wayland/wayland_executor.cpp
+++ b/src/server/frontend_wayland/wayland_executor.cpp
@@ -59,10 +59,21 @@ public:
     explicit State(wl_event_loop* loop)
         : loop{loop}
     {
+        enqueue(
+            []()
+            {
+                on_wayland_thread = true;
+            });
     }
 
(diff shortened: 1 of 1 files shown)


RAOF:
> I can't help a nagging feeling branching on a TLS flag isn't the best way to do this. I guess what I _want_ to do is dynamically change the implementation of enque() - but there's no thread-owned object to do that on (and C++ doesn't really support a dynamic object model).

I actually have most of a `WaylandMainLoop` implementation which more completely encapsulates this which I did before I realised I could much more easily just drop in a single TLS boolean. I could clean that up and propose it if you like.

(CI failure is Travis failing to check out the code, retrying)
bors r=AlanGriffiths

PR #845, opened 2019-05-28
Force latest protocol version

wmww:
This will cause a compile-time error if XML/generated classes are updated to a new version without also updating the child classes.

(Reopened from #832)

### DIFF
--- a/include/wayland/mir/wayland/wayland_base.h
+++ b/include/wayland/mir/wayland/wayland_base.h
@@ -36,6 +36,9 @@ struct Version
 class Resource
 {
 public:
+    template<int V>
+    using Version = Version<V>;
+
     Resource();
     virtual ~Resource() = default;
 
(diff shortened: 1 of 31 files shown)


RAOF:
> I can't think of a scenario where we'd generate a newer protocol wrapper without wanting to update the implementation, but we are losing that capability. I'll let @RAOF express an opinion before landing.

Hm.

So, for the simple, single-issue protocols - layer shell, xdg-wm-base, zlinux-dma-buf, etc - you'd not want to update the XML without also implementing the new XML.

But the core `wayland.xml` contains a whole bunch of protocols which don't get updated in lock-step, and we may well want to, say, support `wl_seat` v7 without also supporting `wl_data_offer` v4, or whatever.

Can we easily split this into core/non-core?

wmww:
Updates to stable protocols:

1. Are somewhat infrequent
2. Often require no compositor changes (just add optional events)
3. Rarely require significant changes

Furthermore, we should strive to support the latest protocol versions (even when it does take some effort). I don't think we should put too much work into making it easy to be outdated, especially when the case where this is desirable is purely theoretical.

RAOF:
I'm persuaded by that reasoning.

bors r+

PR #846, opened 2019-05-29
New and improved surface depth layer

wmww:
Based on feedback from #815, I've changed the property name from `z_index` to `depth_layer` and changed the type from an int to an enum. I've also documented things a bit better and added more tests.

The `MirDepthLayer` does not map to ABI-stable ints. Instead it is left to the doc comments to explain the order of the layers, and up to `SurfaceStack` to properly implement that (of course, for now they are all in order, and hopefully we won't need to add any layers for a while).

### DIFF
--- a/include/core/mir_toolkit/common.h
+++ b/include/core/mir_toolkit/common.h
@@ -483,7 +483,10 @@ typedef enum MirOutputGammaSupported
 
 /**
  * Depth layer controls Z ordering of surfaces.
+ *
  * A surface will always appear on top of surfaces with a lower depth layer, and below those with a higher one.
+ * A depth layer can be converted to a number with mir::mir_depth_layer_get_index().
+ * This is useful for creating a list indexed by depth layer, or comparing the height of two layers.
  */
 typedef enum MirDepthLayer
 {
--- a/include/core/mir_toolkit/common.h
+++ b/include/core/mir_toolkit/common.h
@@ -481,6 +481,21 @@ typedef enum MirOutputGammaSupported
     mir_output_gamma_supported
 } MirOutputGammaSupported;
 
+/**
+ * Depth layer controls Z ordering of surfaces.
+ * A surface will always appear on top of surfaces with a lower depth layer, and below those with a higher one.
+ */
+typedef enum MirDepthLayer
+{
+    mir_depth_layer_background,         /**< For desktop backgrounds and alike (lowest layer) */
+    mir_depth_layer_below,              /**< For panels or other controls/decorations below normal windows */
+    mir_depth_layer_application,        /**< For normal application windows */
+    mir_depth_layer_always_on_top,      /**< For always-on-top application windows */
+    mir_depth_layer_above,              /**< For panels or notifications that want to be above normal windows */
+    mir_depth_layer_overlay,            /**< For overlays such as lock screens (heighest layer) */
+} MirDepthLayer;
+
+
 /**@}*/
 
 #endif
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -114,7 +114,12 @@ auto miral::BasicWindowManager::add_surface(
 
     auto& session_info = info_for(session);
 
-    WindowSpecification const& spec = policy->place_new_window(session_info, place_new_surface(session_info, params));
+    WindowSpecification spec = policy->place_new_window(session_info, place_new_surface(session_info, params));
+
+    if (!spec.depth_layer().is_set() && spec.parent().is_set())
+        if (auto parent_surface = spec.parent().value().lock())
+            spec.depth_layer() = parent_surface->depth_layer();
+
     scene::SurfaceCreationParameters parameters;
     spec.update(parameters);
     auto const surface_id = build(session, parameters);
@@ -854,6 +859,20 @@ void miral::BasicWindowManager::move_tree(miral::WindowInfo& root, mir::geometry
     }
 }
 
+void miral::BasicWindowManager::set_tree_depth_layer(miral::WindowInfo& root, MirDepthLayer new_layer)
+{
+    std::shared_ptr<scene::Surface> surface = root.window();
+    if (surface->depth_layer() != new_layer || root.depth_layer() != new_layer)
+    {
+        surface->set_depth_layer(new_layer);
+        root.depth_layer(new_layer);
+        for (auto& window : root.children())
+        {
+            set_tree_depth_layer(info_for(window), new_layer);
+        }
+    }
+}
+
 void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpecification const& modifications)
 {
     WindowInfo window_info_tmp{window_info};
@@ -922,7 +941,7 @@ void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpe
     auto& window = window_info.window();
 
     if (modifications.depth_layer().is_set())
-        std::shared_ptr<scene::Surface>(window)->set_depth_layer(modifications.depth_layer().value());
+        set_tree_depth_layer(window_info, modifications.depth_layer().value());
 
     if (window_info.type() != window_info_tmp.type())
         std::shared_ptr<scene::Surface>(window)->configure(mir_window_attrib_type, window_info.type());
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -878,6 +878,7 @@ void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpe
     COPY_IF_SET(confine_pointer);
     COPY_IF_SET(userdata);
     COPY_IF_SET(shell_chrome);
+    COPY_IF_SET(depth_layer);
 
 #undef COPY_IF_SET
 
@@ -919,6 +920,9 @@ void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpe
     std::swap(window_info_tmp, window_info);
 
     auto& window = window_info.window();
+
+    if (modifications.depth_layer().is_set())
+        std::shared_ptr<scene::Surface>(window)->set_depth_layer(modifications.depth_layer().value());
 
     if (window_info.type() != window_info_tmp.type())
         std::shared_ptr<scene::Surface>(window)->configure(mir_window_attrib_type, window_info.type());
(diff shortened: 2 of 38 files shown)


include/core/mir_toolkit/common.h
```diff
@@ -481,6 +481,21 @@ typedef enum MirOutputGammaSupported
...
+/**
+ * Depth layer controls Z ordering of surfaces.
+ * A surface will always appear on top of surfaces with a lower depth layer, and below those with a higher one.
+ */
+typedef enum MirDepthLayer
+{
+    mir_depth_layer_background,         /**< For desktop backgrounds and alike (lowest layer) */
+    mir_depth_layer_below,              /**< For panels or other controls/decorations below normal windows */
+    mir_depth_layer_application,        /**< For normal application windows */
+    mir_depth_layer_always_on_top,      /**< For always-on-top application windows */
```
RAOF: I'm unconvinced that this is best modeled as a layer in this scheme. For example: it's entirely possible that one would want an always-on-top window on `mir_depth_layer_background` - for example, you want your Conky desktop widget showing your system stats *on top* of any other background, but below everything else.

src/miral/basic_window_manager.cpp
```diff
@@ -116,13 +116,19 @@ auto miral::BasicWindowManager::add_surface(
...
     auto const surface_id = build(session, parameters);
-    Window const window{session, session->surface(surface_id)};
+    auto surface = session->surface(surface_id);
+    Window const window{session, surface};
     auto& window_info = this->window_info.emplace(window, WindowInfo{window, spec}).first->second;
 
     if (spec.parent().is_set() && spec.parent().value().lock())
+    {
         window_info.parent(info_for(spec.parent().value()).window());
+        surface->set_depth_layer(spec.parent().value().lock()->depth_layer());
```
RAOF: I'm uncomfortable with this

    if (... && foo.lock())
    {
        bar(..., foo.lock()->dereference_here())
    }

setup. We've locked the WindowManager, so we know that the `weak_ptr<Surface>→ WindowInfo&` lookup should succeed, but do we know that the `weak_ptr<Surface>` is still valid here? It would be simple to capture the `shared_ptr<Surface>` from the conditional to ensure that the pointer is still valid.

src/server/scene/surface_stack.h
```diff
@@ -82,43 +83,57 @@ class SurfaceStack : public compositor::Scene, public input::Scene, public shell
...
     SurfaceStack(const SurfaceStack&) = delete;
     SurfaceStack& operator=(const SurfaceStack&) = delete;
     void create_rendering_tracker_for(std::shared_ptr<Surface> const&);
     void update_rendering_tracker_compositors();
+    void insert_surface_at_top_of_depth_layer(std::shared_ptr<Surface> const& surface);
+
+    /**
+     * Converts a MirDepthLayer into an index for a list of layers (0 is bottom-most layer)
+     */
+    static auto mir_depth_layer_to_index(MirDepthLayer depth_layer) -> unsigned int;
```
RAOF: I believe you could mark this as constexpr?

src/server/scene/surface_stack.cpp
```diff
@@ -363,15 +443,40 @@ void ms::SurfaceStack::update_rendering_tracker_compositors()
...
+    switch (depth_layer)
+    {
+    case mir_depth_layer_background:        return 0;
+    case mir_depth_layer_below:             return 1;
+    case mir_depth_layer_application:       return 2;
+    case mir_depth_layer_always_on_top:     return 3;
+    case mir_depth_layer_above:             return 4;
+    case mir_depth_layer_overlay:           return 5;
+    }
+    assert(false);
```
RAOF: Clang will assume that the `depth_layer` parameter contains a valid enum value, will happily accept this code without the `assert()`, and *without* the `assert()` will check that the switch above is exhaustive.

[This](https://github.com/MirServer/mir/blob/954d759c552f53730d1a725fd29a815ea45c43b8/src/common/posix_rw_mutex.cpp#L81) provides an example of how to do this.

AlanGriffiths:
> (This is implicitly arguing that `depth_layer` should be a construct-only property, and immutable after that, but if there's a requirement that surfaces _can_ change layers then we need to consider what happens when they do)

I don't think it can be immutable: shells will want to support, for example, "always on top" options.

include/core/mir_toolkit/common.h
wmww: That is exactly what `mir_depth_layer_below` is for. If you add an "always on top" boolean (as you seem to be advocating for), I don't see how that's different than just doubling the number of layers.

src/miral/basic_window_manager.cpp
wmww: I got a funny feeling when I wrote this, now I see why.

src/server/scene/surface_stack.cpp
wmww: With a default case I don't get a warning/error if I'm missing an enum case (which I do with GCC now). I would hope Clang checks that the switch is exhaustive  with or without the assert.

src/server/scene/surface_stack.h
wmww: Not sure why, as we never use it as `constexpr`, but ok

wmww: Never mind. Due to a GCC bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67371), there isn't an easy way to raise an error if we fall through the switch. Since it's a private function and we don't use it `constexpr`, I'm not going to bother working around that.

src/server/scene/surface_stack.cpp
wmww: I added the `#ifndef __clang__`, but not in a default case.

wmww:
I've been trying to get those tests up, but not having much luck. The obvious place to put high level tests of how depth layer effects actual behavior would be a new file in tests/miral. unfortunately `TestWindowManagerTools` uses stubbed out scene and surfaces, and I haven't been able to get it to test anything useful. It might be easier once I add the depth layer property to Miral.

RAOF:
> > (This is implicitly arguing that `depth_layer` should be a construct-only property, and immutable after that, but if there's a requirement that surfaces _can_ change layers then we need to consider what happens when they do)
> 
> I don't think it can be immutable: shells will want to support, for example, "always on top" options.

Right. This is one reason why I don't think “always on top” maps to a layer; surfaces on *each* layer may wish to be always-on-top (of that layer).

include/core/mir_toolkit/common.h
RAOF: I think that, on any given layer, it's perfectly reasonable for a surface to want to be on top of all other surfaces on that layer. This can't be expressed as doubling the number of layers, for fear of infinities :grin:

wmww:
> surfaces on each layer may wish to be always-on-top (of that layer).

The concept you're describing seems identical to the concept of "the next layer". `mir_depth_layer_below` is the "always on top" of `mir_depth_layer_background`. `mir_depth_layer_overlay` is the "always on top" of `mir_depth_layer_above`. etc.

wmww:
Btw child surfaces staying on their parent's layer by default is on the way (with tests). Let me know if there is a way you would prefer that to work. Are child surfaces currently *always* pinned above their parent?

RAOF:
> > surfaces on each layer may wish to be always-on-top (of that layer).
> 
> The concept you're describing seems identical to the concept of "the next layer". `mir_depth_layer_below` is the "always on top" of `mir_depth_layer_background`. `mir_depth_layer_overlay` is the "always on top" of `mir_depth_layer_above`. etc.

Not quite - always on top of that layer *also* means always *below* the above layer.

RAOF:
> Btw child surfaces staying on their parent's layer by default is on the way (with tests). Let me know if there is a way you would prefer that to work. Are child surfaces currently _always_ pinned above their parent?

I think that an ancestor-tree of surfaces should not span layers; ie: all children of a surface should be on the same layer.

This, of course, is another thing that's broken by “always-on-top is a layer” :grin:

wmww:
> Not quite - always on top of that layer also means always below the above layer. You can split that by introducing `mir_depth_layer_above` < `mir_depth_layer_above_always_on_top` < `mir_depth_layer_overlay` < `mir_depth_layer_always_on_top`, but that way lies infinite regression :)

This isn't a problem because we don't need `mir_depth_layer_above_always_on_top`. We have enough layers. If we had an always-on-top boolean property we would reduce the number of layers. If we did for some reason need always-on-top for for every layer we've got, we could double the number of layers once and be done.

> This, of course, is another thing that's broken by “always-on-top is a layer” grin

Or perhaps if it should be possible for child windows to be always-on-top, that is a reason to not restrict the whole tree to a single layer.

src/server/scene/surface_stack.cpp
RAOF: Aha! This must be a new thing for GCC. Cool!

wmww:
Currently, I have it so that when a parent's depth layer is set, the whole tree is updated, but new layers can be set for children afterward. There are theoretical use cases in which this wouldn't be ideal, however both modifying depth layer and having children on different layers are edge cases, so I don't see the need in getting their interaction perfect until it becomes a problem.

If we do need to "get it right", we can just add a `lock_depth_layer` boolean property to `miral::WindowSpecification`, `miral::WindowInfo` and `shell::SurfaceSpecification`. `scene::Surface` will not need to change.

The current behavior is documented and tested.

wmww:
bors try

RAOF:
So, I continue to think that “layer” and “always on top” are orthogonal concepts. I don't mind if always-on-top gets *implemented* in Mir as an extra layer associated with each API-visible layer or an extra bool on the surface we filter on when sorting or whatever, but I think that the API we present should differentiate the two.

wmww:
My take is that the concept of always-on-top is a subset of the concept of layers. `mir_depth_layer_always_on_top` exists because toolkits have always-on-top (and generally don't support more layers than that). If the client *can* specify one of several layers it wants to be on (such as a layer shell surface), it no longer needs always-on-top.

I fail to see why:
```
bottom (normal)
bottom (always-on-top)
application (normal)
application (always-on-top)
top (normal)
top (always-on-top)
```
is any better than
```
background
below
application
always_on_top
above
overlay
```

RAOF:
So, I've made my peace with this; my underlying concerns are basically “I don't think layer-shell is well designed”, but (a) the API proposed here has (some) more semantics implied than layer-shell and (b) we want to implement layer-shell.

bors r+

wmww:
bors r=RAOF

PR #847, opened 2019-05-31
Explicit handling of extensions in MirAL and Mir

wmww:
This eliminates the colon separated extension list from everywhere except the command line option. It adds new functions to get standard and supported extensions from the frontend, and telling the fontend which to enable. It also adds new `enable()` and `disable()` functions to `miral::WaylandExtensions`, which should make correct usage of the API more obvious.


include/miral/miral/wayland_extensions.h
```diff
@@ -33,6 +33,14 @@ namespace mir { class Server; }
 
 namespace miral
 {
+
+/// Enables shell components such as panels, notifications and lock screens. Malicious clients could potentially
+/// use this protocol to steal input focus or otherwise bother the user
+static char constexpr zwlr_layer_shell_v1[] = "zwlr_layer_shell_v1";
```
AlanGriffiths: Injecting a character array into every translation unit that uses this header is an interesting way to avoid ABI breaks if we ever drop a protocol extension.

I'm not sure I like it (but open to persuasion). Only dropping extension when we break ABI doesn't seem an undue burden.

include/miral/miral/wayland_extensions.h
```diff
@@ -53,6 +61,8 @@ class WaylandExtensions
     /// Initialize "enabled by default" to a custom set of extensions (colon
     /// separated list).
     /// \note This can only be a subset of supported_extensions()
+    /// \deprecated A better option is to use the default constructor, enable()
+    /// and disable()
```
AlanGriffiths: Having thought about it, I don't think we should deprecate this. It covers scenarios that are not addressed otherwise: where the compositor needs to support a specific set of extensions (i.e. not additions & deletions from default.)

tests/miral/wayland_extensions.cpp
```diff
@@ -180,12 +180,16 @@ struct ClientDecorationCreator
         wl_display_roundtrip(display);
         wl_display_roundtrip(display);
 
+        ASSERT_THAT(compositor, NotNull());
+        ASSERT_THAT(decoration_manager, NotNull());
+        ASSERT_THAT(shell, NotNull());
```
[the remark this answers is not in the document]
wmww: My thought was that other tests that don't use this should be allowed to run even if something is seriously broken with what wayland interfaces are exposed.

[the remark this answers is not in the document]
AlanGriffiths: If these fail, then the problem is not with the code being tested here and should not be reported as such.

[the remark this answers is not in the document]
wmww: It's obvious what the failures mean though. There will be a big line of identical failures, and they lead here upon any investigation. It wont waste a significant amount of development time, unlike aborting the program which makes the tests increasingly heard to work with.

Imagine you're porting to a new platform or something, a lot of tests are failing but half of them are aborting the program instead of reporting failures. You'd have to do a bunch of --gtest_filter trial-and-error just to see the full list of what's failing.

include/miral/miral/wayland_extensions.h
wmww: In that case, perhaps we should add another constructor that takes a `std::vector<std::string>` so we can deprecate the colon list being used anywhere but command line options.

include/miral/miral/wayland_extensions.h
wmww: Seems reasonable to only drop extensions on an ABI break, and then we can throw if an unsupported extension is requested. I'll put the strings in the cpp file and this in symbols.map.

include/miral/miral/wayland_extensions.h
```diff
@@ -33,13 +34,21 @@ namespace mir { class Server; }
 
 namespace miral
 {
+/**
+ * Supported wayland extensions that are not enabled by default.
+ * These can be passed into WaylandExtensions::enable() to turn them on.
+ * @{ */
```
AlanGriffiths: These should be namespaced in WaylandExtensions (just like supported()).

wmww:
bors r=wmww

PR #848, opened 2019-06-03
Allow unfocused windows to receive mouse down events

wmww:
When the user clicks on an unfocused window, the window gets focus. We were not also sending that click event through to the client, and with this PR we are. I believe this is a better UX, and it fixes a layer-shell bug where unforeseeable windows could not be clicked. Most importantly, it is the standard behavior in other desktops (tested on Gnome Wayland and MacOS).

EDIT: to test:
* Open two windows next to each other
* Click in the unfocused one
* Note that it becomes focused, and also the click action happens (menu opens, icon is selected, whatever)

### DIFF
--- a/src/miral/minimal_window_manager.cpp
+++ b/src/miral/minimal_window_manager.cpp
@@ -434,8 +434,7 @@ bool miral::MinimalWindowManager::Impl::handle_touch_event(MirTouchEvent const* 
     {
         if (auto const window = tools.window_at(new_touch))
         {
-            if (tools.active_window() != tools.select_active_window(window))
-                consumes_event = true;
+            tools.select_active_window(window);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/miral/minimal_window_manager.cpp
```diff
@@ -355,8 +355,7 @@ bool miral::MinimalWindowManager::Impl::handle_pointer_event(MirPointerEvent con
     {
         if (auto const window = tools.window_at(new_cursor))
         {
-            if (tools.active_window() != tools.select_active_window(window))
-                consumes_event = true;
+            tools.select_active_window(window);
```
AlanGriffiths: I'm not sure (beyond "history") why the logic was this way, but we probably need to change the corresponding case in `handle_touch_event()` to match. I need to try that out...

wmww: `handle_touch_event()` updated

Issue #850, opened 2019-06-05
LP builds failing because there are new libmiral3 symbols


AlanGriffiths:
```
--- debian/libmiral3.symbols (libmiral3_2.5.0_amd64)
+++ dpkg-gensymbols7iPqU8	2019-06-04 16:46:40.884983419 +0000
@@ -5,6 +5,11 @@
  MIRAL_2.3@MIRAL_2.3 2.3.0
  MIRAL_2.4@MIRAL_2.4 2.4.0
  MIRAL_2.5@MIRAL_2.5 2.5.0
+ MIRAL_2.6@MIRAL_2.6 2.5.0
+ _ZN5miral10WindowInfo11depth_layerE13MirDepthLayer@MIRAL_2.6 2.5.0
+ _ZN5miral19WindowSpecification11depth_layerEv@MIRAL_2.6 2.5.0
+ _ZNK5miral10WindowInfo11depth_layerEv@MIRAL_2.6 2.5.0
+ _ZNK5miral19WindowSpecification11depth_layerEv@MIRAL_2.6 2.5.0
  (c++)"miral::AddInitCallback::AddInitCallback(std::function<void ()> const&)@MIRAL_2.0" 2.0.0
  (c++)"miral::AddInitCallback::operator()(mir::Server&) const@MIRAL_2.0" 2.0.0
  (c++)"miral::AddInitCallback::~AddInitCallback()@MIRAL_2.0" 2.0.0
src/miral/CMakeFiles/check-miral-symbols.dir/build.make:61: recipe for target 'src/miral/CMakeFiles/check-miral-symbols' failed
```

PR #851, opened 2019-06-05
Update MirAL minor version, regenerate-miral-symbols-map script and debian .symbols

AlanGriffiths:
Update MirAL minor version, regenerate-miral-symbols-map script and debian .symbols. (Fixes: #850)


Saviq:
Aha!

bors r+

Issue #852, opened 2019-06-05
ClientCredsTestFixture.session_authorizer_receives_pid_of_connecting_clients fails in TearDown()


AlanGriffiths:
This is the most common LP test failure recorded in #841 

Looking at the console log it seems we;re timing out in waiting for the server to exit:
* `Process::wait_for_termination()` is called by
* `InterprocessClientServerTest::stop_server()`  is called by
* `InterprocessClientServerTest::TearDown()`

PR #854, opened 2019-06-06
Avoid hanging server used by ClientCredsTestFixture.session_authorizer_receives_pid_of_connecting_clients during TearDown()

AlanGriffiths:
Avoid hanging server used by ClientCredsTestFixture.session_authorizer_receives_pid_of_connecting_clients during TearDown(). (Fixes: #852)

### DIFF
--- a/tests/acceptance-tests/test_client_authorization.cpp
+++ b/tests/acceptance-tests/test_client_authorization.cpp
@@ -32,6 +32,8 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include <atomic>
+
 namespace mf = mir::frontend;
 namespace mt = mir::test;
 namespace mtf = mir_test_framework;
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
That's a valid concern. But pre-existing: the timed_wait was not on the inc thread and served only to stop the server exiting.

RAOF:
Is there any particular reason *not* to just make `SharedRegion` contain `std::atomic<pid_t> client_pid`, etc?

I mean, that could be a separate PR, but while you're fixing the test you might as well *fix* it :grin:

AlanGriffiths:
> Is there any particular reason _not_ to just make `SharedRegion` contain `std::atomic<pid_t> client_pid`, etc?
> 
> I mean, that could be a separate PR, but while you're fixing the test you might as well _fix_ it grin

I'm OK with that and planned to do so. But, caution, AFAIK the C++ memory model doesn't formally address shared memory.

AlanGriffiths:
bors r=RAOF

PR #855, opened 2019-06-07
Refactor TestWindowManagerTools and add output tests

wmww:
Forgot to PR this earlier. This moves the implementation of `TestWindowManagerTools` to a new `test_window_manager_tools.cpp` file. It also moves `TestWindowManagerTools` into the `mir::test` namespace and adds tests for output updates. I needed similar tests for application zones, so I thought I'd add outputs as well.

### DIFF
--- a/tests/miral/output_updates.cpp
+++ b/tests/miral/output_updates.cpp
@@ -1,0 +1,125 @@
+/*
+ * Copyright © 2017 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Alan Griffiths <alan@octopull.co.uk>
+ */
+
+#include "test_window_manager_tools.h"
+#include "miral/output.h"
+
+#include <experimental/optional>
+
+using namespace miral;
+using namespace testing;
+namespace mt = mir::test;
+
+namespace
+{
+Rectangle const display_area_a{{20, 30}, {600, 400}};
+Rectangle const display_area_b{{620, 0}, {800, 500}};
+
+struct OutputUpdates : mt::TestWindowManagerTools
+{
+    void SetUp() override
+    {
+        basic_window_manager.add_session(session);
+    }
+};
+}
+
+TEST_F(OutputUpdates, policy_notified_of_output_creation)
+{
+    std::experimental::optional<Output> output_a;
+
+    EXPECT_CALL(*window_manager_policy, advise_output_create(_))
+        .WillOnce(Invoke([&](Output const& output){ output_a = output; }));
+
+    set_outputs({display_area_a});
+
+    Mock::VerifyAndClearExpectations(window_manager_policy);
+    EXPECT_THAT(output_a.value().extents(), Eq(display_area_a));
+}
+
+TEST_F(OutputUpdates, policy_notified_of_multiple_outputs)
+{
+    std::experimental::optional<Output> output_a;
+    std::experimental::optional<Output> output_b;
+
+    EXPECT_CALL(*window_manager_policy, advise_output_create(_))
+        .WillOnce(Invoke([&](Output const& output){ output_a = output; }))
+        .WillOnce(Invoke([&](Output const& output){ output_b = output; }));
+
+    set_outputs({display_area_a, display_area_b});
+
+    Mock::VerifyAndClearExpectations(window_manager_policy);
+    EXPECT_THAT(output_a.value().extents(), Eq(display_area_a));
+    EXPECT_THAT(output_b.value().extents(), Eq(display_area_b));
+    EXPECT_FALSE(output_a.value().is_same_output(output_b.value()));
+}
+
+TEST_F(OutputUpdates, policy_notified_of_output_update)
+{
+    std::experimental::optional<Output> output_initial;
+    std::experimental::optional<Output> output_original;
+    std::experimental::optional<Output> output_updated;
+
+    EXPECT_CALL(*window_manager_policy, advise_output_create(_))
+        .WillOnce(Invoke([&](Output const& output){ output_initial = output; }));
+
+    set_outputs({display_area_a});
+
+    Mock::VerifyAndClearExpectations(window_manager_policy);
+
+    EXPECT_CALL(*window_manager_policy, advise_output_update(_, _))
+        .WillOnce(Invoke([&](Output const& updated, Output const& original){
+            output_original = original;
+            output_updated = updated;
+        }));
+
+    set_outputs({display_area_b});
+
+    Mock::VerifyAndClearExpectations(window_manager_policy);
+    EXPECT_TRUE(output_initial.value().is_same_output(output_original.value()));
+    EXPECT_TRUE(output_original.value().is_same_output(output_updated.value()));
+
+    EXPECT_THAT(output_initial.value().extents(), Eq(display_area_a));
+    EXPECT_THAT(output_original.value().extents(), Eq(display_area_a));
+    EXPECT_THAT(output_updated.value().extents(), Eq(display_area_b));
+}
+
+TEST_F(OutputUpdates, policy_notified_of_output_delete)
+{
+    std::experimental::optional<Output> output_a;
+    std::experimental::optional<Output> output_b;
+    std::experimental::optional<Output> output_b_deleted;
+
+    EXPECT_CALL(*window_manager_policy, advise_output_create(_))
+        .WillOnce(Invoke([&](Output const& output){ output_a = output; }))
+        .WillOnce(Invoke([&](Output const& output){ output_b = output; }));
+
+    set_outputs({display_area_a, display_area_b});
+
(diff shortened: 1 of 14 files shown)


wmww:
Sorry there's added code and a refactor in the same PR. The only functionality added to `TestWindowManagerTools` is `mt::TestWindowManagerTools::set_outputs()`.

tests/miral/output_updates.cpp
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright © 2017 Canonical Ltd.
```
AlanGriffiths: Forgot to edit after cut & paste?

tests/miral/output_updates.cpp
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Alan Griffiths <alan@octopull.co.uk>
```
AlanGriffiths: Forgot to edit after cut & paste?

tests/miral/output_updates.cpp
```diff
@@ -0,0 +1,125 @@
...
+    set_outputs({display_area_a, display_area_b});
+
+    Mock::VerifyAndClearExpectations(window_manager_policy);
+
+    EXPECT_CALL(*window_manager_policy, advise_output_delete(_))
+        .WillOnce(Invoke([&](Output const& output){ output_b_deleted = output; }));
+
+    set_outputs({display_area_a});
+
+    Mock::VerifyAndClearExpectations(window_manager_policy);
```
AlanGriffiths: Calling Mock::VerifyAndClearExpectations() multiple times "smells" of testing multiple things in one test. At the very least there should be a comment on why it is called at all.

tests/miral/test_window_manager_tools.cpp
```diff
@@ -0,0 +1,290 @@
...
+    void handle_request_drag_and_drop(miral::WindowInfo& /*window_info*/) {}
+    void handle_request_move(miral::WindowInfo& /*window_info*/, MirInputEvent const* /*input_event*/) {}
+    void handle_request_resize(miral::WindowInfo& /*window_info*/, MirInputEvent const* /*input_event*/, MirResizeEdge /*edge*/) {}
+    mir::geometry::Rectangle confirm_placement_on_display(const miral::WindowInfo&, MirWindowState, mir::geometry::Rectangle const& new_placement)
+    {
+        return new_placement;
+    }
+};
+
+struct TestDisplayConfigurationObserver : mir::ObserverRegistrar<mir::graphics::DisplayConfigurationObserver>
```
AlanGriffiths: I'd call TestDisplayConfigurationObserver FakeDisplayConfigurationObserverRegistrar as that better describes its use here.

tests/miral/test_window_manager_tools.cpp
```diff
@@ -0,0 +1,290 @@
...
+auto mt::TestWindowManagerTools::create_surface(
+    std::shared_ptr<mir::scene::Session> const& session,
+    mir::scene::SurfaceCreationParameters const& params) -> mir::frontend::SurfaceId
+{
+    // This type is Mir-internal, I hope we don't need to create it here
+    std::shared_ptr<mir::frontend::EventSink> const sink;
+    return session->create_surface(params, sink);
+}
+
+void mt::TestWindowManagerTools::set_outputs(std::vector<miral::Rectangle> outputs)
```
AlanGriffiths: Why is set_outputs() doing so much? There seem to be three distinct elements:
1. Creating a DisplayConfiguration double (it happens to be a Mock, but I'm not convinced that is always right);
2. Checking the state of display_configuration_observer; and
3. Sending a configuration_applied notification.

I think it would be more flexible to have:
1. create_mock_display_configuration(outputs) -> shared<DisplayConfiguration>();
2. notify_configuration_applied(shared<DisplayConfiguration>()); and, 
3. possibly now or in the future, create_fake_display_configuration(outputs) -> shared<DisplayConfiguration>() (which does not check that for_each_output() is called exactly once).

(And notify_configuration_applied() can call a notify_configuration_applied() on FakeDisplayConfigurationObserverRegistrar which can worry about its own state?)

tests/miral/output_updates.cpp
wmww: Before the rest of the test can happen, we need to make sure the expected functions have been called or else variables we expect to hold values will be unset. If Mir was broken in such a way the calls were not made, I think it would be more helpful to see the error that those functions were not called, than see an error that the values they set are not what is expected. I have dropped some comments in.

tests/miral/output_updates.cpp
wmww: Fixed

tests/miral/output_updates.cpp
wmww: Fixed

tests/miral/test_window_manager_tools.cpp
wmww: Done

tests/miral/test_window_manager_tools.cpp
wmww: Done as you suggest. I also changed the expected call to `display_config->for_each_output()` from `WillOnce` to `WillRepeatedly`, as that happening exactly once (or ever) isn't a thing we need to test for.

tests/miral/test_window_manager_tools.cpp
```diff
@@ -255,7 +255,7 @@ auto mt::TestWindowManagerTools::create_mock_display_configuration(std::vector<m
 {
     auto const display_config = std::make_shared<const mir::test::doubles::MockDisplayConfiguration>();
     EXPECT_CALL(*display_config, for_each_output(testing::_))
-        .WillOnce(testing::Invoke([outputs](std::function<void(mir::graphics::DisplayConfigurationOutput const&)> func){
+        .WillRepeatedly(testing::Invoke([outputs](std::function<void(mir::graphics::DisplayConfigurationOutput const&)> func){
```
AlanGriffiths: If we "Don't test that display_config->for_each_output() is called exactly once" then we should probably have `.Times(AtLeast(1))` instead of an implied `.Times(1)`.

tests/miral/test_window_manager_tools.cpp
```diff
@@ -0,0 +1,297 @@
...
+        }));
+    return display_config;
+}
+
+void mt::TestWindowManagerTools::notify_configuration_applied(
+    std::shared_ptr<graphics::DisplayConfiguration const> display_config)
+{
+    auto config_observer = self->display_configuration_observer.observer.lock();
+    ASSERT_THAT(config_observer, testing::NotNull());
+    config_observer->configuration_applied(display_config);
```
AlanGriffiths: I'm not convinced this logic belongs here and not on FakeDisplayConfigurationObserverRegistrar, but won't block on it.

tests/miral/test_window_manager_tools.cpp
wmww: As is with `WillRepeatedly`, it can be called any number of times (including zero). I believe that's what we want, because just getting a config doesn't imply if/how many times a test is going to use it.

AlanGriffiths: But the second and subsequent calls will needlessly spam the logs with "unexpected call".

tests/miral/test_window_manager_tools.cpp
wmww: That makes some sense, done.

tests/miral/test_window_manager_tools.cpp
wmww: I created a `FakeDisplayConfiguration` as we discussed

AlanGriffiths:
Unrelated fail:

```
[ RUN      ] ClientSurfaceEventsTest.surface_resizes_under_pointer
[2019-06-12 08:20:39.278527] <information> mirserver: Starting
[2019-06-12 08:20:39.278558] < - debug - > mirserver: No session management supported
[2019-06-12 08:20:39.278597] <information> VT switch key handler: No VT switching support available: MinimalConsoleServices does not support VT switching
[2019-06-12 08:20:39.278675] <information> mirserver: Selected driver: mir:stub-graphics (version 1.2.0)
[2019-06-12 08:20:39.285560] <information> mirserver: Using software cursor
[2019-06-12 08:20:39.285948] <information> mirserver: Selected input driver: mir:stub-input (version: 1.2.0)
unable to lock lockfile /run/user/0/wlcs-tests.lock, maybe another compositor is running
[2019-06-12 08:20:39.286137] <information> mirserver: Mir version 1.2.0
[2019-06-12 08:20:39.286572] <information> mirserver: Initial display configuration:
[2019-06-12 08:20:39.286602] <information> mirserver: * Output 1: VGA connected, used
[2019-06-12 08:20:39.286614] <information> mirserver: . |_ Physical size 0.0" 0x0mm
[2019-06-12 08:20:39.286621] <information> mirserver: . |_ Power is on
[2019-06-12 08:20:39.286629] <information> mirserver: . |_ Current mode 1600x1600 60.00Hz
[2019-06-12 08:20:39.286635] <information> mirserver: . |_ Preferred mode 1600x1600 60.00Hz
[2019-06-12 08:20:39.286641] <information> mirserver: . |_ Orientation normal
[2019-06-12 08:20:39.286648] <information> mirserver: . |_ Logical size 1600x1600
[2019-06-12 08:20:39.286654] <information> mirserver: . |_ Logical position +0+0
[2019-06-12 08:20:39.286661] <information> mirserver: . |_ Scaling factor: 1.00


99% tests passed, 1 tests failed out of 427

Total Test time (real) = 600.12 sec

The following tests FAILED:
	  8 - wlcs.ClientSurfaceEventsTest.* (Timeout)
```

bors r+

PR #857, opened 2019-06-07
Restore Fedora rawhide to being a required CI target

AlanGriffiths:
Rawhide is now building.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,11 +53,11 @@ env:
   - SYSTEM=google:fedora-29-64 VARIANT=amd64
   - SYSTEM=google:fedora-30-64 VARIANT=amd64
 
-## Commented out, not deleted, as a reminder of the syntax
-#matrix:
+matrix:
+# Commented out, not deleted, as a reminder of the syntax
 #  allow_failures:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -53,10 +53,11 @@ env:
   - SYSTEM=google:fedora-29-64 VARIANT=amd64
   - SYSTEM=google:fedora-30-64 VARIANT=amd64
 
-matrix:
-  allow_failures:
-    - env: SYSTEM=google:fedora-rawhide-64 VARIANT=amd64
```
Saviq: Maybe just comment this out, then?

AlanGriffiths: I don't think it is a syntax error "All checks have passed"

Saviq: I meant this would be:
```yaml
allow_failures:
#  - env: SYSTEM=google:fedora-rawhide-64 VARIANT=amd64
fast_finish: true
```

That's apparently valid YAML, not sure if Travis will like it, though.

Still, worth a try - commenting out just the rawhide item. I'd drop the `# Commented out…` line, too.

Issue #858, opened 2019-06-07
Mir with eglstreams crashes on shutdown


gerboland:
I have mir-test-tools deb and snap installed on a Bionic system with nvidia 430 driver. Running

    SNAP=/snap/mir-test-tools/current/ gdb --arg mir_demo_server --test-client /snap/mir-test-tools/current/qml/spinning-rect.sh

Mir shuts down with a crash. 

```
[2019-06-07 15:42:39.948940] <information> mirserver: Starting
[2019-06-07 15:42:39.951208] < - debug - > mirserver: Not using logind for session management: Seat has no active session
[2019-06-07 15:42:39.951461] < - debug - > mirserver: Using Linux VT subsystem for session management
[2019-06-07 15:42:39.951512] <information> mircommon: Loading modules from: /usr/lib/x86_64-linux-gnu/mir/server-platform
[2019-06-07 15:42:39.951546] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-mesa-kms.so.16
[2019-06-07 15:42:39.951557] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/server-mesa-x11.so.16
[2019-06-07 15:42:39.951564] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-eglstream-kms.so.16
[2019-06-07 15:42:39.951572] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/input-evdev.so.7
[2019-06-07 15:42:39.952241] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/input-stub.so
[2019-06-07 15:42:39.952386] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-dummy.so
[2019-06-07 15:42:39.954744] <information> mesa-kms: EGL platform does not support EGL_KHR_platform_gbm extension
[2019-06-07 15:42:39.991979] <information> mesa-kms: Detected software renderer: llvmpipe (LLVM 7.0, 256 bits)
[2019-06-07 15:42:40.004991] <information> mirplatform: Found graphics driver: mir:mesa-kms (version 1.2.0) Support priority: 128
[2019-06-07 15:42:40.005014] <information> mirplatform: Found graphics driver: mir:mesa-x11 (version 1.2.0) Support priority: 0
[...]

PR #860, opened 2019-06-07
Make mgek::EGLOutput::~EGLOutput() noexcept

AlanGriffiths:
We use mgek::EGLOutput with std::shared_ptr and std::vector.

We may not throw from the destructor when used with the standard library.

Therefore: Make mgek::EGLOutput::~EGLOutput() noexcept. (Might "fix" #858, or expose an underlying issue.)


wmww:
```
/spread/mir/src/platforms/eglstream-kms/server/egl_output.cpp:161:13: error: â€˜loggingâ€™ has not been declared
         log(logging::Severity::error,
             ^~~~~~~
```

PR #861, opened 2019-06-07
Application zone

wmww:
Adds a new `Zone` class to miral and updates about application zones. These are the areas we talked about which maximized windows will fill. Right now there is one application zone per output and each covers the entire area of its output, but they will shrink to accommodate exclusive zones when layer shell is implemented. On top of #855

### DIFF
--- a/examples/example-server-lib/tiling_window_manager.cpp
+++ b/examples/example-server-lib/tiling_window_manager.cpp
@@ -21,7 +21,7 @@
 #include <miral/application_info.h>
 #include <miral/window_info.h>
 #include <miral/window_manager_tools.h>
-#include <miral/output.h>
+#include <miral/zone.h>
 
 #include <linux/input.h>
 #include <algorithm>
@@ -655,15 +655,15 @@ void TilingWindowManagerPolicy::advise_end()
     dirty_tiles = false;
 }
 
-void TilingWindowManagerPolicy::advise_output_create(const Output& output)
+void TilingWindowManagerPolicy::advise_application_zone_create(miral::Zone const& zone)
 {
-    displays.add(output.extents());
+    displays.add(zone.extents());
     dirty_tiles = true;
 }
 
-void TilingWindowManagerPolicy::advise_output_update(const Output& updated, const Output& original)
+void TilingWindowManagerPolicy::advise_application_zone_update(miral::Zone const& updated, miral::Zone const& original)
 {
-    if (!equivalent_display_area(updated, original))
+    if (original.extents() != updated.extents())
     {
         displays.remove(original.extents());
         displays.add(updated.extents());
@@ -672,9 +672,9 @@ void TilingWindowManagerPolicy::advise_output_update(const Output& updated, cons
     }
 }
 
-void TilingWindowManagerPolicy::advise_output_delete(Output const& output)
+void TilingWindowManagerPolicy::advise_application_zone_delete(miral::Zone const& zone)
 {
-    displays.remove(output.extents());
+    displays.remove(zone.extents());
     dirty_tiles = true;
 }
 
--- a/include/miral/miral/window_management_policy.h
+++ b/include/miral/miral/window_management_policy.h
@@ -288,6 +288,12 @@ public:
         virtual ~ApplicationZoneAddendum() = default;
         ApplicationZoneAddendum(ApplicationZoneAddendum const&) = delete;
         ApplicationZoneAddendum& operator=(ApplicationZoneAddendum const&) = delete;
+        /**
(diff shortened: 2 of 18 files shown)


include/miral/miral/window_management_policy.h
```diff
@@ -268,6 +269,18 @@ class WindowManagementPolicy
...
 
+/** @name notification of changes to the current application zones
+ * An application zone is the area a maximized application will fill.
+ * There is often (but not necessarily) one zone per output.
+ * The areas normal applications windows should avoid (such as the areas covered by panels)
+ * will not be part of an application zone
+ *  @{ */
+    virtual void advise_application_zone_create(Zone const& application_zone);
+    virtual void advise_application_zone_update(Zone const& updated, Zone const& original);
+    virtual void advise_application_zone_delete(Zone const& application_zone);
```
AlanGriffiths: This is an ABI break (and this is not yet the time for one). Please put these functions in a different interface.

If you look back to Mir 0.29 you'll see some examples of this.

wmww: Ah yes, I forgot to do that

wmww:
bors try

AlanGriffiths:
Failures look like they are same underlying cause #865 (which is fixed now).

bors try

examples/example-server-lib/tiling_window_manager.cpp
```diff
@@ -21,7 +21,7 @@
 #include <miral/application_info.h>
 #include <miral/window_info.h>
 #include <miral/window_manager_tools.h>
-#include <miral/output.h>
+#include <miral/zone.h>
```
AlanGriffiths: The changes to TilingWindowManagerPolicy are simplistic.

The policy takes the bounding rectangle of "displays" and lays out the "tiles" within that. That means it will tile over any gaps between "application zones" which is unlikely to be the intent.

In part this is pre-existing (as if the outputs are not arranged in a rectangle then areas outside them will be used). So this isn't a blocker, just a note.

wmww: At least this will still make it work with exclusive zones if there is only one monitor attached. It may not be perfect, but I'd call that preexisting and this an improvement.

wmww:
Done with the trace stuff. This rebase is only to solve conflicts.

wmww:
Rebased to solve conflicts, but it's the last two commits (`Move logic to get ApplicationZoneAddendum...` and `Add application zones to WindowManagementTrace` that are new)

AlanGriffiths:
> Rebased to solve conflicts, but it's the last two commits (`Move logic to get ApplicationZoneAddendum...` and `Add application zones to WindowManagementTrace`) that are new

Unfortunately git doesn't grok rename: create_mock_display_configuration => create_fake_display_configuration

src/miral/basic_window_manager.cpp
```diff
@@ -2270,40 +2371,54 @@ void miral::BasicWindowManager::update_windows_for_outputs()
...
+                update_window = true;
                 break;
 
             default:
                 break;
             }
+
+            // TODO: Maybe remove update_window and update only if the rect has changed?
+            // This would probably be more correct, but also may or may not break existing users
+            // if (updated_window_rect != original_window_rect)
```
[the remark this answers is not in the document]
wmww: 1. The state of the windows is not changing. The state of the outputs/application zones is changing, but the policy is notified of that separately.
2. Nowhere is it documented that `confirm_placement_on_display()` will be called on output/application zone changes that don't effect the window
3. Policies can (and should) update window positions in `advise_end()` instead of `confirm_placement_on_display()`, else they may not have enough information to do it properly. For example, the order of `advise_application_zone_delete()`, `advise_output_delete()` and `confirm_placement_on_display()` is not defined, so no particular order should be assumed. The only correct way to remember all useful calls and process them in `advise_end()`

src/miral/basic_window_manager.h
```diff
@@ -197,6 +196,28 @@ class BasicWindowManager : public virtual mir::shell::WindowManager,
     void invoke_under_lock(std::function<void()> const& callback) override;
 
 private:
+    /// An area for windows to be placed in
+    struct WindowArea
```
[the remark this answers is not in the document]
wmww: I have no problem with `DisplayRegion`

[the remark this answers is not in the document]
wmww: Upon further thought, I don't see the advantage of "region" over "area". I do get the that it is non-obvious that a `WindowArea` contains many windows, so I'll go with `DisplayArea`.

PR #862, opened 2019-06-07
Comment, refactor and fix minor logic error in MultiMonitorArbiter

wmww:
This doesn't fix any bug we've seen in the wild, it's just to make the code more readable and correct. The code was clearly wrong before (clearing current_buffer_users without advancing the buffer) but the case where that would manifest as a bug is obscure. I captured such a situation in the 2nd test I add, which works as follows:

    Compositor A acquires a buffer
    Compositor B acquires the same buffer
    Compositor A acquires the buffer again
    A new buffer is added to the schedule
    Compositor B tries to acquire

What should happen (and does with this PR) is the new buffer is returned. With the old code current_buffer_users had been cleared in step 3 and only A had been added back, so the schedule would not advance and the old buffer would be returned.

### DIFF
--- a/src/server/compositor/multi_monitor_arbiter.cpp
+++ b/src/server/compositor/multi_monitor_arbiter.cpp
@@ -48,8 +48,10 @@ std::shared_ptr<mg::Buffer> mc::MultiMonitorArbiter::compositor_acquire(composit
     if (current_buffer_users.find(id) != current_buffer_users.end() || !current_buffer)
     {
         if (schedule->num_scheduled())
+        {
             current_buffer = schedule->next_buffer();
-        current_buffer_users.clear();
+            current_buffer_users.clear();
+        }
     }
     current_buffer_users.insert(id);
 
--- a/src/server/compositor/multi_monitor_arbiter.cpp
+++ b/src/server/compositor/multi_monitor_arbiter.cpp
@@ -42,19 +42,26 @@ std::shared_ptr<mg::Buffer> mc::MultiMonitorArbiter::compositor_acquire(composit
 {
     std::lock_guard<decltype(mutex)> lk(mutex);
 
-    if (!current_buffer && !schedule->num_scheduled())
-        BOOST_THROW_EXCEPTION(std::logic_error("no buffer to give to compositor"));
-
-    if (current_buffer_users.find(id) != current_buffer_users.end() || !current_buffer)
+    // If there is no current buffer or there is, but this compositor is already using it...
+    if (!current_buffer || current_buffer_users.find(id) != current_buffer_users.end())
     {
-        if (schedule->num_scheduled())
+        // And if there is a scheduled buffer
+        if (schedule->num_scheduled() > 0)
         {
+            // Advance the current buffer
             current_buffer = schedule->next_buffer();
             current_buffer_users.clear();
         }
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
Could we have a better name and explanatory comment?

"MultiMonitorArbiter improvements" doesn't tell us anything - would you propose "MultiMonitorArbiter worsening"? Tell us how it is improved. (support for some scenario perhaps?)

"A slight change in logic where current_buffer_users.clear() is only called if the current buffer is updated." The code already says that: the comment should explain why we want that.

wmww:
This doesn't fix any bug we've seen in the wild, it's just to make the code more readable and correct. The code was clearly wrong before (clearing `current_buffer_users` without advancing the buffer) but the case where that would manifest as a bug is obscure. I captured such a situation in the 2nd test I add, which works as follows:

1. Compositor A acquires a buffer
2. Compositor B acquires the same buffer
3. Compositor A acquires the buffer again
4. A new buffer is added to the schedule
5. Compositor B tries to acquire

What should happen (and does with this PR) is the new buffer is returned. With the old code `current_buffer_users` had been cleared in step 3 and only A had been added back, so the schedule would not advance and the old buffer would be returned.

AlanGriffiths:
> This doesn't fix any bug we've seen in the wild, it's just to make the code more readable and correct. The code was clearly wrong before (clearing `current_buffer_users` without advancing the buffer) but the case where that would manifest as a bug is obscure. I captured such a situation in the 2nd test I add, which works as follows:
> 
>     1. Compositor A acquires a buffer
> 
>     2. Compositor B acquires the same buffer
> 
>     3. Compositor A acquires the buffer again
> 
>     4. A new buffer is added to the schedule
> 
>     5. Compositor B tries to acquire
> 
> 
> What should happen (and does with this PR) is the new buffer is returned. With the old code `current_buffer_users` had been cleared in step 3 and only A had been added back, so the schedule would not advance and the old buffer would be returned.

Could we have a better name and explanatory comment so that this is recorded for posterity?

wmww:
It's hard to write a name that means much at all without it being >100 words long. I added a comment description.

wmww:
bors try

AlanGriffiths:
> It's hard to write a name that means much at all without it being >100 words long. I added a comment description.

/o\

I'm talking about the PR name and top comment. These end up in the commit log.

> "MultiMonitorArbiter improvements" doesn't tell us anything - would you propose "MultiMonitorArbiter worsening"? Tell us how it is improved. (support for some scenario perhaps?)
> 
> "A slight change in logic where current_buffer_users.clear() is only called if the current buffer is updated." The code already says that: the comment should explain why we want that.

For a recent example: https://github.com/MirServer/mir/pull/860

Ends up with the log:

860: Make mgek::EGLOutput::~EGLOutput() noexcept r=wmww a=AlanGriffiths

We use mgek::EGLOutput with std::shared_ptr and std::vector.

We may not throw from the destructor when used with the standard library.

Therefore: Make mgek::EGLOutput::~EGLOutput() noexcept. (Might "fix" #858, or expose an underlying issue.)

wmww:
@RAOF poke

You don't need to read the previous discussion too closely. It's just Alan telling me to improve the name and comment on the PR, which I did.

PR #867, opened 2019-06-12
Auto update debian symbols

wmww:
Before CMake would call `dpkg-gensymbols` and spit out a warning if the symbols were wrong. Now, it calls a new python script that checks, transforms them into C++ symbols and automatically adds them to the debian symbols file.


AlanGriffiths:
I've fixed that (if @wmww agrees my change):

bors delegate=wmww

wmww:
Looks good, thanks

bors r=AlanGriffiths

PR #870, opened 2019-06-14
Better data structure for MultiMonitorArbiter::current_buffer_users.

RAOF:
A `std::set<>` is highly likely to be backed by some form of tree structure. While that's got better
asymptotic behaviour than a linear search through a vector, we're extremely unlikely to have more
than 10 outputs here, and the common case is *1* output.

There are other microöptimisations here - a SmallVec would be an obvious choice - but just do
the simplest thing because I haven't done any benchmarking ☺

### DIFF
--- a/src/server/compositor/multi_monitor_arbiter.cpp
+++ b/src/server/compositor/multi_monitor_arbiter.cpp
@@ -32,6 +32,8 @@ mc::MultiMonitorArbiter::MultiMonitorArbiter(
     std::shared_ptr<Schedule> const& schedule) :
     schedule(schedule)
 {
+    // We're highly unlikely to have more than 6 outputs
+    current_buffer_users.reserve(6);
 }
 
 mc::MultiMonitorArbiter::~MultiMonitorArbiter()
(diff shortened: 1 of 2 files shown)


RAOF:
I just felt the need to do this after seeing it in the other `MultiMonitorArbiter` review :speak_no_evil:

RAOF:
> I agree that vector<> is a better choice, but I think we'd do even better micro-optimisation using nullptr to denote empty elements than std::experimental::optional<>{}.
> 
> bors delegate=RAOF

I considered that, but the APIs using `mc::CompositorID` document it as an [arbitrary ID](https://github.com/MirServer/mir/blob/b52bb47fd2fef68ae2c7bfeef538c861ed95709c/include/server/mir/compositor/scene.h#L47), so I think it'd *technically* be an API break to treat `nullptr` as an invalid ID.

I can *do* that, but is it something that we want to do?

AlanGriffiths:
It isn't worth debating a "technical API break".

bors r+

Issue #871, opened 2019-06-14
[wayland] creating wrapper objects when the client has not supplied a wl_resource requires needs symbols not declared in the header


AlanGriffiths:
These are needed to construct protocol objects in the server: For example:

src/server/frontend_wayland/data_device.cpp:32

```c++
namespace mir
{
namespace wayland
{
extern struct wl_interface const wl_data_offer_interface_data;
}
}
```

AlanGriffiths:
Or, maybe we simply need to generate a wrapper constructor that doesn't need a resource. Because the use is:
```c++
        wl_resource* new_resource = wl_resource_create(
            client,
            &mw::wl_data_offer_interface_data,
            wl_resource_get_version(resource),
            0);
        current_offer = new DataOffer{new_resource, source, this};
```

PR #877, opened 2019-06-17
[Wayland] Wrapper objects can easily be constructed in the server

AlanGriffiths:
[Wayland] Wrapper objects can easily be constructed in the server. (Fixes: #871)

### DIFF
--- a/src/server/frontend_wayland/data_device.cpp
+++ b/src/server/frontend_wayland/data_device.cpp
@@ -25,14 +25,6 @@
 namespace mf = mir::frontend;
 namespace mw = mir::wayland;
 
-namespace mir
-{
-namespace wayland
-{
-extern struct wl_interface const wl_data_offer_interface_data;
-}
-}
-
 namespace
 {
 class DataDeviceManager;
@@ -41,7 +33,7 @@ struct DataDevice;
 
 struct DataOffer : mw::DataOffer
 {
-    DataOffer(wl_resource* new_resource, DataSource* source, DataDevice* device);
+    DataOffer(wl_client* client, int protocol_version, DataSource* source, DataDevice* device);
 
     void accept(uint32_t serial, std::experimental::optional<std::string> const& mime_type) override
     {
@@ -288,12 +280,7 @@ void DataDevice::notify_new(DataSource* source)
 
     if (has_focus)
     {
-        wl_resource* new_resource = wl_resource_create(
-            client,
-            &mw::wl_data_offer_interface_data,
-            wl_resource_get_version(resource),
-            0);
-        current_offer = new DataOffer{new_resource, source, this};
+        current_offer = new DataOffer{client, wl_resource_get_version(resource), source, this};
     }
 }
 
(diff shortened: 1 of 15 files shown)


AlanGriffiths:
> The idea works, but I don't like sending around a raw client and runtime version. Would it work to pass in a `mw::Resource *` and pull the client and version out of that in generated code?

That sounds very easy to mistake for the existing signature:
```c++
    DataOffer(struct wl_resource* resource, Version<3>);
    DataOffer(struct wl_client* client, int runtime_version, Version<3> static_version);
```


> The idea works, but I don't like sending around a raw client and runtime version. Would it work to pass in a `mw::Resource *` and pull the client and version out of that in generated code?

wmww:
Perhaps `mw::Resource::make_resource<mw::ResourceType>` that returns a `wl_resource`, which then you can pass into the normal constructor.

AlanGriffiths:
> Perhaps `mw::Resource::make_resource<mw::ResourceType>` that returns a `wl_resource`, which then you can pass into the normal constructor.

Having a template function would need a way to map from ResourceType to '"&", wl_name, "_interface_data"' (and having that declaration). If we have the declaration we can do what we do now:

```c++
        wl_resource* new_resource = wl_resource_create(
            client,
            &mw::wl_data_offer_interface_data,
            wl_resource_get_version(resource),
            0);
        current_offer = new DataOffer{new_resource, source, this};
```

I thought it more elegant to have a suitable constructor.

wmww:
Passing the raw version seems error-prone (easy to forget the version is supposed to come from another resource). We could also offer a generated static function `auto make_resource(wl_resource const* source) -> wl_resource *` and then use the normal constructor.

AlanGriffiths:
> Passing the raw version seems error-prone (easy to forget the version is supposed to come from another resource). We could also offer a generated static function `auto make_resource(wl_resource const* derive_from) -> wl_resource *` and then use the normal constructor.


How about we pull the `resource` (and `client`?) members up to `Resource` and generate a constructor with the signature: `DataOffer(Resource* parent, Version<3> static_version);`?

AlanGriffiths:
@wmww how do you like this iteration?

src/server/frontend_wayland/data_device.cpp
```diff
@@ -320,17 +307,12 @@ void DataDevice::focus_on(wl_client* focus)
...
-            0);
-        current_offer = new DataOffer{new_resource, current_source, this};
+        current_offer = new DataOffer{client, resource, current_source, this};
     }
 }
 
-DataOffer::DataOffer(wl_resource* new_resource, DataSource* source, DataDevice* device) :
-    mw::DataOffer(new_resource, Version<3>()),
+DataOffer::DataOffer(wl_client* client, wl_resource* parent, DataSource* source, DataDevice* device) :
+    mw::DataOffer(client, parent, Version<3>()),
```
wmww: Why pass in the parent resource? Why not just use the one on `device`?

src/wayland/generated/wayland_wrapper.cpp
```diff
@@ -191,6 +194,9 @@ mw::Compositor::Compositor(struct wl_resource* resource, Version<4>)
     wl_resource_set_implementation(resource, Thunks::request_vtable, this, &Thunks::resource_destroyed_thunk);
 }
 
+mw::Compositor::Compositor(struct wl_client* client, struct wl_resource* parent, Version<4> static_version)
+    : Compositor{wl_resource_create(client, &wl_compositor_interface_data , wl_resource_get_version(parent), 0), static_version} {}
```
wmww: Why pass in a resource and a client? Why not use `wl_resource_get_client()`?

AlanGriffiths: There is an existing signature for passing the resource from the client request:
```c++
    Compositor(struct wl_resource* resource, Version<3>);
```

src/server/frontend_wayland/data_device.cpp
AlanGriffiths: OK, I _could_ use `device->client` & `device->resource`

src/wayland/generated/wayland_wrapper.cpp
RAOF: So, I think that:
 * For globals, this constructor doesn't make sense; instances of a global can only be constructed from the `bind(wl_resource* new_resource)` method, which already has a resource
 * For other resources (eg: `mw::Surface`), I *think* this can quite easily be reworked as `Surface(Compositor* parent)`. And then we'd (maybe) remove the raw `wl_resource*` constructor(?)

AlanGriffiths:
Dropped in favour of #882

PR #880, opened 2019-06-18
Implement Ctrl-Alt-T handling for miral-shell

AlanGriffiths:


AlanGriffiths:
Urgh! That interesting `/usr/bin/gnome-terminal` script has defeated me.

It all works fine with `weston-terminal`.

With `qterminal`(and `"gnome-terminal", "--disable-factory"`) we have the zombie window after exit.

I suspect there is a bug in the shutdown code.

PR #881, opened 2019-06-18
Add make_resource() to generated classes

### LINKED ISSUE #871: [wayland] creating wrapper objects when the client has not supplied a wl_resource requires needs symbols not declared in the header

wmww:
Allows the server to create new Wayland resources instead of them coming from the client. The new resource can then be past into the constructor of a Wayland object, just like new resources from clients. Calling `make_resource()` explicitly shows that a new `wl_resource` is created, and the given resource is the parent.

Fixes #871, alternative to #877

### DIFF
--- a/src/server/frontend_wayland/data_device.cpp
+++ b/src/server/frontend_wayland/data_device.cpp
@@ -41,7 +41,7 @@ struct DataDevice;
 
 struct DataOffer : mw::DataOffer
 {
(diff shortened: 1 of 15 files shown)


AlanGriffiths:
Dropped in favour of #882

PR #887, opened 2019-06-20
[nested] Make sure we destroy eglImage before class gets destructed

mariogrip:
We're not allowed to destroy the `EGLClientBuffer` until all of the associated images are destroy by `eglDestroyImageKHR()`.

On arm/android the buffer was destroyed before all the images were destroyed, it isn't obvious why but simplifying the code as proposed seems to fix this.

### DIFF
--- a/src/server/graphics/nested/buffer.cpp
+++ b/src/server/graphics/nested/buffer.cpp
@@ -55,6 +55,14 @@ public:
     {
     }
 
+    ~TextureAccess() {
+    	for(auto& it : egl_image_map)
+    	{
+            EGLDisplay disp = it.first.first;
+            extensions.eglDestroyImageKHR(disp, it.second);
+    	}
+    }
+
     void bind() override
     {
         using namespace std::chrono;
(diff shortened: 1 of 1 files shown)


RAOF:
Hm. Does this fix anything? Because by my understanding of C++ this code is (almost) exactly equivalent to the code it's replacing (the only difference is that it's missing a pointless heap allocation).

When a C++ object is destroyed, the defined destructor implementation (if any) is called, then members are destroyed in reverse order of declaration, and any not-yet-destroyed members are fair game for access.

So this moves the image destruction from first thing (after the non-existent explicit destructor), to first thing (now the explicit destructor)

AlanGriffiths:
I (mostly) agree with Chris:

The `egl_image_map` is the first member (or base class) to be destroyed.

Version| Sequence
--|--
Before| each `map::value_type` is destroyed in turn: first the `EGLImageUPtr` (which calls `eglDestroyImageKHR()` and deletes the `EGLImageKHR`) and then the `ImageResources`.
After| `eglDestroyImageKHR()` is called on all the `EGLImageKHR` members first. Then each `map::value_type` is destroyed in turn: first the `EGLImageKHR` and then the `ImageResources`.

If that change of order actually makes a difference, then it isn't clear why and the top comment is misleading about the mechanism.

We need a better comment for the commit. (At worst the change is harmless.)

mariogrip:
Not sure if this is a specific issue on arm or not, but it somehow destroys the image and display before eglDestroy gets called. I don't know how and why this only seems to happen on arm/android. But in my debugging this is what seems to happens.

It does indeed not seem to be an issue on amd64 after testing with a vm.

AlanGriffiths:
> Hm. I wonder if the Android `EGLDisplay` or `EGLContext` are not void*, and so actually have a destructor?

I'd expect them to be C-compatible types: if not a pointer then an int.

> I still can't see the problem (although if the Android types _do_ have a destructor, then this does significantly change the destruction order).

Nor I. I have a suspicion that this is simply masking a problem by changing memory layout. But that's not enough to block on.

PR #888, opened 2019-06-21
server: Allow mirclient connector construction to fail.

RAOF:
We've deprecated mirclient. There are a bunch of platform interfaces which are needed
*only* by the mirclient interfaces, so allow platforms to throw instead of providing
a stub implementation (that then won't work anyway) of these interfaces.

Platforms which currently support mirclient are unaffected; only platforms which don't
support mirclient anyway will change behaviour.

### DIFF
--- a/src/server/display_server.cpp
+++ b/src/server/display_server.cpp
@@ -45,6 +45,65 @@ namespace mg = mir::graphics;
 namespace mi = mir::input;
 namespace msh = mir::shell;
 
+namespace
+{
+class StubConnector : public mir::frontend::Connector
+{
+public:
+    void start() override
+    {
+    }
+
+    void stop() override
+    {
+    }
+
+    auto client_socket_fd() const -> int override
+    {
+        BOOST_THROW_EXCEPTION((std::runtime_error{"mirclient support unavailable"}));
+    }
+
+    auto client_socket_fd(std::function<void(std::shared_ptr<mir::frontend::Session> const&)> const&) const
+        -> int override
+    {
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
I have concerns about using exceptions for "normal" control flow, but I guess it isn't trivial to deal with `RenderingPlatform::make_ipc_operations()` otherwise.

I think what I'd prefer is for `the_connector()` to default to a stub implementation (with a new configuration option to enable it). Then there would be no need to handle the exception (here at least) as enabling the option for an unsupported platform is a user/configuration error.

RAOF:
That would be an option. I did it this way because it's minimally disruptive for existing users - anything that currently works continues to work, and systems which used to cause mirclients to fail on connect instead don't expose a mirclient socket.

I guess we *do* plan to deprecate mirclient in the near-term future, so I guess it's not *so* disruptive for users who currently use mirclient-clients to add an extra command-line option...

AlanGriffiths:
> I guess we _do_ plan to deprecate mirclient in the near-term future, so I guess it's not _so_ disruptive for users who currently use mirclient-clients to add an extra command-line option...

I guess how we do that is something to discuss that goes beyond the scope of this PR.

There's a range of options: from a `miral::MirClientSupport` that allows the shell to "opt in" to a configuration option to, not having a visible configuration option and just testing for an environment variable "MIR_ENABLE_MIRCLIENT_LEGACY".

Something to discuss at tomorrow's sync?

RAOF:
Sure!

PR #889, opened 2019-06-21
[miral] Handle exceptions propagating from InternalClient code

### LINKED ISSUE #886: Repeatedly rotating the output crashes the server
1. Run `miral-shell`
2. Start a terminal (Ctrl-Alt-Shift-T)
3. Run: `while mirout output 1 rotate left; do sleep 1; mirout output 1 rotate normal; sleep 1; done`

After a while `miral-shell` crashes.

Also seen with `egmde`, but not (so far) reproduced with `mir_demo_server` - which implies it may be in the "background" internal client code.

AlanGriffiths:
[miral] Handle exceptions propagating from InternalClient code.

### DIFF
--- a/src/miral/internal_client.cpp
+++ b/src/miral/internal_client.cpp
@@ -218,11 +218,19 @@ void WlInternalClientRunner<Base>::run(mir::Server& server)
 
     thread = std::thread{[this]
         {
-            if (auto const display = wl_display_connect_to_fd(fd))
+            try
             {
-                auto const deleter = mir::raii::deleter_for(display, &wl_display_disconnect);
-                client_code(display);
-                wl_display_roundtrip(display);
+                if (auto const display = wl_display_connect_to_fd(fd))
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
Handling the exception is clearly sensible.

But #886 was a crash seen the exception throwing code (which no longer happens). I'm unsure why adding code to handle an exception prevents the problem that causes the exception.

AlanGriffiths:
Oh! Despite initial appearances this isn't a fix: we now "eat" the exception:
```plain
[2019-06-21 14:45:36.957280] <information> miral::Internal Client: Failed to read Wayland events: Value too large for defined data type: Dynamic exception type: std::exception
std::exception::what: std::exception
```
(But because the background fails to disappear when the client "exits" then things look sort-of OK.

The code still makes sense, but isn't really a fix. Comment adjusted accordingly.

PR #890, opened 2019-06-21
Wayland: process output update notifications on the correct thread

AlanGriffiths:
Wayland: process output update notifications on the correct thread. (Fixes: #886)


AlanGriffiths:
Bors r=wmww

PR #893, opened 2019-06-24
Add mir_window_state_attached

wmww:
Add the `mir_surface_type_attached` window state, and two new properties to windows: `attached_edges` and `exclusive_rect`. These will be important for implementing layer surfaces. I still need to add tests, but I want to throw this up now so I can start getting feedback on it.

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -153,6 +153,10 @@ auto miral::BasicWindowManager::add_surface(
         break;
     }
 
+    case mir_window_state_attached:
+        place_attached(window_info);
+        break;
+
     default:
         break;
     }
@@ -1091,6 +1095,12 @@ void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpe
         set_state(window_info, modifications.state().value());
     }
 
+    if (window_info.state() == mir_window_state_attached)
+    {
+        if (modifications.state().is_set() || modifications.attached_edges().is_set())
+            place_attached(window_info);
+    }
+
     if (modifications.confine_pointer().is_set())
         std::shared_ptr<scene::Surface>(window)->set_confine_pointer_state(modifications.confine_pointer().value());
 }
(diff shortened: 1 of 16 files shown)


wmww:
Now it's tested and working

src/miral/basic_window_manager.cpp
```diff
@@ -1127,6 +1137,54 @@ void miral::BasicWindowManager::place_and_size(WindowInfo& root, Point const& ne
     move_tree(root, new_pos - root.window().top_left());
 }
 
+void miral::BasicWindowManager::place_attached(WindowInfo& root)
+{
+    // We can't place the attached window if there aren't any outputs
+    if (root.type() != mir_window_type_attached || outputs.size() == 0)
```
RAOF: Likewise, `root.type() != mir_window_type_attached` here is a precondition failure.

src/miral/window_info.cpp
```diff
@@ -174,10 +180,12 @@ bool miral::WindowInfo::can_be_active() const
     case mir_window_type_gloss:
     case mir_window_type_tip:          /**< AKA "tooltip"                       */
     case mir_window_type_decoration:
-    default:
+    case mir_window_type_attached:     /**< AKA "panel"/"notification"/"etc"    */
```
RAOF: Can panel/notification windows *really* not have input focus? That seems odd!

AlanGriffiths:
I've not read the code, but "attached" sounds more like a window state than a window type.

I can imagine "attached" being used for "normal", "gloss" or "inputmethod" windows types.

I can't imagine "attached" taking "fullscreen", "vertmaximized" or "minimized" window states.

wmww:
Switched attached from a window type to a window state as per my discussion this morning with Alan. How's it looks now? If there are any concerstatens from before I've forgot to address, just reraise them.

wmww:
Added another test and fixed the unused function warning in Clang

src/miral/window_info.cpp
```diff
@@ -114,11 +116,15 @@ miral::WindowInfo::Self::Self(Window window, WindowSpecification const& params)
     min_aspect(optional_value_or_default(params.min_aspect(), default_min_aspect_ratio)),
     max_aspect(optional_value_or_default(params.max_aspect(), default_max_aspect_ratio)),
     shell_chrome(optional_value_or_default(params.shell_chrome(), mir_shell_chrome_normal)),
-    depth_layer(optional_value_or_default(params.depth_layer(), mir_depth_layer_application))
+    depth_layer(optional_value_or_default(params.depth_layer(), mir_depth_layer_application)),
+    attached_edges(optional_value_or_default(params.attached_edges(), mir_placement_gravity_center))
```
AlanGriffiths: It isn't obvious why this should default to "center". Are we using that as a synonym for "none"?

wmww: Attaching to the center of the output seems like as good a default as any. I don't think it's terribly important what the default is, as the value will generally be set at the same time as the state is changed to attached.

wmww:
No changes, just fixed the conflict

src/miral/window_info.cpp
RAOF: Is it *mandatory* that the attachment should be set when switching to attached state? That sounds reasonable?

wmww: No. If it was then that would imply it's mandatory to set the edges every time you switch into an attached state, even if you had already set the edges the previous time. Also, that would also imply it's impossible to change the attached edges when the window is not in an attached state, which in some cases may be convenient. I don't see the potential for problems with just having them be independent properties.

AlanGriffiths: As `attached_edges` is always set to _something_ then not requiring it to be set when switching to "attached" is consistent. If that's what we elect to do then any arbitrary default value works and "center" is the canonical "middle of the road".

src/miral/symbols.map
```diff
@@ -438,15 +438,19 @@ global:
     miral::WaylandExtensions::supported[*;
     miral::WaylandExtensions::zwlr_layer_shell_v1*;
     miral::WaylandExtensions::zxdg_output_manager_v1*;
+    miral::WindowInfo::attached_edges*;
```
AlanGriffiths: Unless we plan on cherry-picking into 1.3.0 (I think that unnecessary) these need moving to 2.7

wmww:
Symbols should be good to go now

wmww:
bors r=AlanGriffiths

PR #894, opened 2019-06-25
clients: Make the mirclient socket optional.

RAOF:
We're deprecating libmirclient. A lot of platform interfaces are used only for
mirclient support, so it would be convenient for writing new platforms if
those interfaces could simply throw rather than returning a useless stub (and
having clients fail after connecting to the socket).

This adds a new, disabled by default, option `--enable-mirclient`.

When specified, Mir will listen on a mirclient socket as normal. If the server
uses a platform that throws rather than providing working mirclient interfaces
the server will fail to start up.

When not specified, the server will not provide a mirclient socket to listen
on, and platforms will not need to provide the mirclient interfaces.


RAOF:
Alternative to #888

src/platform/symbols.map
```diff
@@ -234,3 +234,10 @@ MIR_PLATFORM_1.1.1 {
     mir::graphics::EGLExtensions::PlatformBaseEXT*;
   };
 } MIR_PLATFORM_1.1.0;
+
+MIR_PLATFORM_1.2.0 {
```
AlanGriffiths: We already released Mir 1.2.0, this is probably wants to be 1.3?

src/platform/options/default_configuration.cpp
```diff
@@ -213,6 +214,8 @@ mo::DefaultConfiguration::DefaultConfiguration(
             "in unexpected ways] throw an exception (instead of a core dump)")
         (debug_opt, "Enable extra development debugging. "
             "This is only interesting for people doing Mir server or client development.")
+        (enable_mirclient_opt, po::value<bool>()->default_value(false),
```
AlanGriffiths: Elsewhere you test to see if `enable-mirclient` is supplied (not its value): by providing a default value this guarantees it will always be supplied.

src/platform/symbols.map
AlanGriffiths: We've now also released 1.3...

AlanGriffiths:
> We're deprecating libmirclient. A lot of platform interfaces are used only for
> mirclient support, so it would be convenient for writing new platforms if
> those interfaces could simply throw rather than returning a useless stub (and
> having clients fail after connecting to the socket).

I don't think they even need to throw: I'd be happy if they `abort()` with a meaningful message.

PR #896, opened 2019-06-25
Wlcs integration doesn't need a MockGL

AlanGriffiths:
Wlcs integration doesn't need a MockGL.


AlanGriffiths:
As this "conflicts" with my other branch I'll merge it there. (As I really only pushed it to prove it works in CI)

Issue #878, opened 2019-06-18
[Wayland] there's no support for downstream servers testing with wlcs


AlanGriffiths:
We need to publish a TestDisplayServer/miral integration based API/library (or something equivalent).

PR #898, opened 2019-06-26
Add miral::TestWlcsDisplayServer to mirtest-dev

AlanGriffiths:
Add miral::TestWlcsDisplayServer to mirtest-dev. (Fixes: #878)

This is the second and final part of what a downstream would need to test their own server.

(Proposed against the first part for the convenience of reviewers, if the first part lands it will need retargetting.)


AlanGriffiths:
bors try

PR #905, opened 2019-07-03
Bump wl_output version 2 -> 3

wmww:
All that's added is a release request, which we don't have to respond to.


wmww:
Yes, I agree it's due for a rewrite. I don't thinks its been touched much since before I started here.

PR #907, opened 2019-07-03
[wlcs test fixture] Configure wayland extensions

### LINKED ISSUE #906: Command line options are ignored by the wlcs test fixture
`miral::TestWlcsDisplayServer` helpfully provides a constructor that accepts the command line passed from, for example, wlcs.

This _tries_ to pass this to the server (via a lambda), but this happens after command line parsing.

The command line used comes from the default constructor of `miral::TestDisplayServer` which supplies a dummy. There is no facility to pass command-line options to the runner.

AlanGriffiths:
[wlcs test fixture] Configure wayland extensions. (Fixes #906)

1. Ensure all supported Wayland extensions are switched on by default.
2. Fix command-line processing so that `--wayland-extensions` can be used.

### DIFF
--- a/include/test/miral/test_display_server.h
+++ b/include/test/miral/test_display_server.h
@@ -48,6 +48,7 @@ private:
 struct TestDisplayServer : private TestRuntimeEnvironment
 {
     TestDisplayServer();
+    TestDisplayServer(int argc, char const** argv);
 
     virtual ~TestDisplayServer();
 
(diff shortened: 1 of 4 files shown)


AlanGriffiths:
A nice follow-up would be to configure the `WlcsExtensionDescriptor` according to the extensions that are enabled. But that wasn't trivial and I decided to leave that for the "interested reader".

PR #908, opened 2019-07-03
Add missing geometry overloads and converters

wmww:
Add a functions to make our ~over-engineered~ expressive and typesafe geometry system more usable. Hopefully this will get rid of randomly subtracting `Width{}` or `X{}` from everything in the expression until it compiles.

### DIFF
--- a/examples/example-server-lib/floating_window_manager.cpp
+++ b/examples/example-server-lib/floating_window_manager.cpp
@@ -374,8 +374,8 @@ bool FloatingWindowManagerPolicy::handle_keyboard_event(MirKeyboardEvent const* 
                     (modifications.size().is_set() ? modifications.size().value() : active_window.size()).width;
 
                 modifications.top_left() = window_info.needs_titlebar(window_info.type()) ?
-                                           active_output.top_right() - as_displacement({new_width, 0}) + title_bar_height :
-                                           active_output.top_right() - as_displacement({new_width, 0});
+                                           active_output.top_right() - Displacement{as_delta(new_width), 0} + title_bar_height :
+                                           active_output.top_right() - Displacement{as_delta(new_width), 0};
                 break;
             }
 
--- a/include/core/mir/geometry/dimensions.h
+++ b/include/core/mir/geometry/dimensions.h
@@ -169,6 +169,20 @@ inline constexpr DeltaX operator*(DeltaX const& dx, Scalar scale) { return scale
 template<typename Scalar>
 inline constexpr DeltaY operator*(DeltaY const& dy, Scalar scale) { return scale*dy; }
 
+// Converting between types is fine, as long as they are along the same axis
+inline constexpr Width as_width(DeltaX const& dx) { return Width{dx.as_int()}; }
+inline constexpr Height as_height(DeltaY const& dy) { return Height{dy.as_int()}; }
+inline constexpr X as_x(DeltaX const& dx) { return X{dx.as_int()}; }
+inline constexpr Y as_y(DeltaY const& dy) { return Y{dy.as_int()}; }
+inline constexpr DeltaX as_delta(X const& x) { return DeltaX{x.as_int()}; }
+inline constexpr DeltaY as_delta(Y const& y) { return DeltaY{y.as_int()}; }
+inline constexpr X as_x(Width const& w) { return X{w.as_int()}; }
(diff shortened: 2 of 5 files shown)


include/core/mir/geometry/dimensions.h
```diff
@@ -118,6 +118,10 @@ inline constexpr DeltaX operator+(DeltaX lhs, DeltaX rhs) { return DeltaX(lhs.as
 inline constexpr DeltaY operator+(DeltaY lhs, DeltaY rhs) { return DeltaY(lhs.as_int() + rhs.as_int()); }
 inline constexpr DeltaX operator-(DeltaX lhs, DeltaX rhs) { return DeltaX(lhs.as_int() - rhs.as_int()); }
 inline constexpr DeltaY operator-(DeltaY lhs, DeltaY rhs) { return DeltaY(lhs.as_int() - rhs.as_int()); }
+inline DeltaX& operator+=(DeltaX& lhs, DeltaX rhs) { return lhs = DeltaX(lhs.as_int() + rhs.as_int()); }
+inline DeltaY& operator+=(DeltaY& lhs, DeltaY rhs) { return lhs = DeltaY(lhs.as_int() + rhs.as_int()); }
+inline DeltaX& operator-=(DeltaX& lhs, DeltaX rhs) { return lhs = DeltaX(lhs.as_int() - rhs.as_int()); }
+inline DeltaY& operator-=(DeltaY& lhs, DeltaY rhs) { return lhs = DeltaY(lhs.as_int() - rhs.as_int()); }
```
RAOF: Non-blocking, but couldn't you just implement these as `return lhs = lhs + rhs;`, since you've already got the above operators?

include/core/mir/geometry/dimensions.h
```diff
@@ -161,6 +169,20 @@ inline constexpr DeltaX operator*(DeltaX const& dx, Scalar scale) { return scale
...
+inline constexpr X as_x(DeltaX const& dx) { return X{dx.as_int()}; }
+inline constexpr Y as_y(DeltaY const& dy) { return Y{dy.as_int()}; }
+inline constexpr DeltaX as_delta(X const& x) { return DeltaX{x.as_int()}; }
+inline constexpr DeltaY as_delta(Y const& y) { return DeltaY{y.as_int()}; }
+inline constexpr X as_x(Width const& w) { return X{w.as_int()}; }
+inline constexpr Y as_y(Height const& h) { return Y{h.as_int()}; }
+inline constexpr Width as_width(X const& x) { return Width{x.as_int()}; }
+inline constexpr Height as_height(Y const& y) { return Height{y.as_int()}; }
+inline constexpr DeltaX as_delta(Width const& w) { return DeltaX{w.as_int()}; }
+inline constexpr DeltaY as_delta(Height const& h) { return DeltaY{h.as_int()}; }
```
RAOF: These ones are... ok?

I'd kinda prefer things using these to be eg: `{0} + delta;`, but C++ won't infer the right things there…

wmww: The problem isn't time inference, the problem is that `auto left_margin = {0} + ((rect.size.width - {0}) - exclusive.left() - {0}))` is not easy code to write or to read.

RAOF:
Sure.

bors r+

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -374,8 +374,8 @@ bool FloatingWindowManagerPolicy::handle_keyboard_event(MirKeyboardEvent const*
                     (modifications.size().is_set() ? modifications.size().value() : active_window.size()).width;
 
                 modifications.top_left() = window_info.needs_titlebar(window_info.type()) ?
-                                           active_output.top_right() - as_displacement({new_width, 0}) + title_bar_height :
-                                           active_output.top_right() - as_displacement({new_width, 0});
+                                           active_output.top_right() - as_displacement(Size{new_width, 0}) + title_bar_height :
+                                           active_output.top_right() - as_displacement(Size{new_width, 0});
```
RAOF: Wait. Why are we first creating a `Size` and then running `as_displacement` on it, rather than just creating a `Displacement`?

wmww: Because `new_width` is of type `Width`, although now that I think about it you're right. One sec...

PR #909, opened 2019-07-04
[MirAL]: Exclusive zones

wmww:
This adds support for exclusive zones to `BasicWindowManager`. We already have `exclusive_rect` property floating around. This PR uses it to shrink the application zones and the windows inside them. To accomplish this I had to reshuffle some calls, and move responsibility of placing attached surfaces into `update_windows_for_outputs()`.


AlanGriffiths:
`update_windows_for_outputs()` is now even more of a misnamed function, but I won't block on that...

PR #910, opened 2019-07-04
Layer shell popups

wmww:
Allow layer shell surfaces to make XDG popups.

If you desire to test you can do so with gtk-layer-demo from [GTK Layer Shell](https://github.com/wmww/gtk-layer-shell), though It's probably sufficient to just convince yourself it doesn't break XDG shell. WLCS tests will be ready at some point.

### DIFF
--- a/src/server/frontend_wayland/layer_shell_v1.cpp
+++ b/src/server/frontend_wayland/layer_shell_v1.cpp
@@ -160,11 +160,7 @@ void mf::LayerSurfaceV1::set_keyboard_interactivity(uint32_t keyboard_interactiv
 
 void mf::LayerSurfaceV1::get_popup(struct wl_resource* popup)
 {
-    auto popup_window_role =
-        static_cast<WindowWlSurfaceRole*>(
-            static_cast<XdgPopupStable*>(
-                static_cast<wayland::XdgPopup*>(
-                    wl_resource_get_user_data(popup))));
+    auto* const popup_window_role = XdgPopupStable::from(popup);
 
     mir::shell::SurfaceSpecification specification;
     specification.parent_id = surface_id();
--- a/src/server/frontend_wayland/layer_shell_v1.cpp
+++ b/src/server/frontend_wayland/layer_shell_v1.cpp
@@ -20,6 +20,7 @@
 
 #include "wl_surface.h"
 #include "window_wl_surface_role.h"
+#include "xdg_shell_stable.h"
 
 #include "mir/shell/surface_specification.h"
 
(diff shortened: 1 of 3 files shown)


src/server/frontend_wayland/layer_shell_v1.cpp
```diff
@@ -159,7 +160,16 @@ void mf::LayerSurfaceV1::set_keyboard_interactivity(uint32_t keyboard_interactiv
 
 void mf::LayerSurfaceV1::get_popup(struct wl_resource* popup)
 {
-    (void)popup;
+    auto popup_window_role =
+        static_cast<WindowWlSurfaceRole*>(
+            static_cast<XdgPopupStable*>(
+                static_cast<wayland::XdgPopup*>(
+                    wl_resource_get_user_data(popup))));
```
AlanGriffiths: That looks horrid. Would this work?
```C++
auto const popup_window_role = dynamic_cast<WindowWlSurfaceRole*>(wayland::XdgPopup::from(popup);
```

wmww: Because `XdgPopupStable` inherits from both `wayland::XdgPopup` and `WindowWlSurfaceRole`, so you can't cast directly between the two unrelated base classes. Stand by for a `XdgPopupStable::from` implementation to clear this up.

AlanGriffiths: But casting directly between two unrelated base classes is exactly what `dynamic_cast` is good for!?

src/server/frontend_wayland/layer_shell_v1.cpp
```diff
@@ -160,11 +160,7 @@ void mf::LayerSurfaceV1::set_keyboard_interactivity(uint32_t keyboard_interactiv
 
 void mf::LayerSurfaceV1::get_popup(struct wl_resource* popup)
 {
-    auto popup_window_role =
-        static_cast<WindowWlSurfaceRole*>(
-            static_cast<XdgPopupStable*>(
-                static_cast<wayland::XdgPopup*>(
-                    wl_resource_get_user_data(popup))));
+    auto popup_window_role = static_cast<WindowWlSurfaceRole*>(XdgPopupStable::from(popup));
```
AlanGriffiths: Do we need `static_cast` to reach an accessible base class?

wmww: I just did that so I could use `auto`.

AlanGriffiths:
```C++
    auto* const popup_window_role = XdgPopupStable::from(popup);
```
or, if the exact type is really important:
```c++
    WindowWlSurfaceRole* const popup_window_role = XdgPopupStable::from(popup);
```

src/server/frontend_wayland/layer_shell_v1.cpp
wmww: Oh, cool. I assumed `dynamic_cast` could only handle down casts, but now I see it somehow has enough runtime info to pull that off. Anyway, the new system with `XdgPopupStable::from()` is still better IMO.

wmww:
Oh, true. The type isn't important, not sure why I thought it was.

wmww:
bors r=AlanGriffiths

Saviq:
Seems like a fedora issue…

bors r=AlanGriffiths

PR #916, opened 2019-07-09
Fix google::protobuf::closure related errors (protobuf 3.8.0)

### LINKED ISSUE #913: build fails with protobuf 3.8.0
Alpine Linux edge has just updated their protobuf from 3.6.1 to 3.8.0. postmarketOS builds on top of Alpine edge and it has mir 1.1.2 packaged. Due to the soname change from protobuf, mir needs to be rebuilt. But it does not build against protobuf 3.8.0:

```
[ 20%] Building CXX object src/server/scene/CMakeFiles/mirscene.dir/broadcasting_session_event_sink.cpp.o
In file included from /home/pmos/build/src/mir-1.1.2/src/server/frontend/session_mediator.cpp:19:
In file included from /home/pmos/build/src/mir-1.1.2/src/server/frontend/session_mediator.h:22:
In file included from /home/pmos/build/src/mir-1.1.2/src/server/frontend/display_server.h:22:
/home/pmos/build/src/mir-1.1.2/src/include/common/mir/protobuf/display_server.h:37:27: error: no type named 'Closure' in namespace 'google::protobuf'
        google::protobuf::Closure* done) = 0;
        ~~~~~~~~~~~~~~~~~~^
/home/pmos/build/src/mir-1.1.2/src/include/common/mir/protobuf/display_server.h:41:27: error: no type named 'Closure' in namespace 'google::protobuf'
        google::protobuf::Closure* done) = 0;
        ~~~~~~~~~~~~~~~~~~^
...
```

Mir's `display_server.h` has not been changed since the last two years, so I expect the current version of mir to behave the same way.

I've been reading through the protobuf source, and it seems that `google::protobuf::Closure` was removed between versions. Unfortunately the git log of protobuf is [unreadable](https://github.com/protocolbuffers/protobuf/commit/d61aede89cf188367766b971f59cf57d7835d8e8) and I couldn't find anything useful online. Closure used to be defined in `src/google/protobuf/stubs/common.h`.

(CC: @z3ntu)

ollieparanoid:
I am not familiar with the codebase, this is just a makeshift patch that made it compile with protobuf 3.8.0. Creating the PR as requested in #913.

To do things properly, we'll need some kind of check for the protobuf version or maybe an entirely different approach of doing things. Feedback welcome :wink: 

Fixes #913

### DIFF
--- a/src/client/rpc/mir_basic_rpc_channel.h
+++ b/src/client/rpc/mir_basic_rpc_channel.h
@@ -27,6 +27,8 @@
 #include <condition_variable>
 #include <functional>
 
+#include <google/protobuf/stubs/callback.h>
+
 namespace google
 {
 namespace protobuf
--- a/src/include/common/mir/protobuf/display_server.h
+++ b/src/include/common/mir/protobuf/display_server.h
@@ -19,6 +19,7 @@
 #ifndef MIR_PROTOBUF_DISPLAY_SERVER_H_
 #define MIR_PROTOBUF_DISPLAY_SERVER_H_
 
+#include <google/protobuf/stubs/callback.h>
 #include "mir_protobuf.pb.h"
 
 namespace mir
--- a/src/include/common/mir/protobuf/display_server_debug.h
+++ b/src/include/common/mir/protobuf/display_server_debug.h
@@ -19,6 +19,7 @@
 #ifndef MIR_PROTOBUF_DISPLAY_SERVER_DEBUG_H_
 #define MIR_PROTOBUF_DISPLAY_SERVER_DEBUG_H_
 
+#include <google/protobuf/stubs/callback.h>
 #include "mir_protobuf.pb.h"
 
 namespace mir


wmww:
Thanks! As you say we'll need some sort of check to make it compile on on older distros and pass the tests. I vaguely remember checking protobuf versions being a PITA, but also I remember coming up with a solution. Don't worry about it, and when I get a chance I'll dig through Mir's CMake files to see how I did that.

wmww:
Turns out I was thinking of gtest. Checking protobuf version should be as easy as `#if GOOGLE_PROTOBUF_VERSION > 3006000` (see https://github.com/protocolbuffers/protobuf/blob/de57caa1fd813e131f1ef52e04fe9722249ed988/src/google/protobuf/stubs/common.h#L84). I'd prefer if you put the includes in the required .cpp files, as the .h don't currently include any protobuf headers, so probably don't need to. I can't do it because I don't have distro with a new enough protobuf at the moment.

ollieparanoid:
Cool, I'll take a shot at it some time soon and then update the patch.

AlanGriffiths:
> ...the .h don't currently include any protobuf headers, so probably don't need to. I can't do it because I don't have distro with a new enough protobuf at the moment.

I think that's wrong for the headers that already include `mir_protobuf.pb.h`:
```c++
#include <google/protobuf/stubs/common.h>
```
...and the relevant section of `google/protobuf/stubs/common.h` is:
```c++
// TODO(liujisi): Remove the following includes after the include clean-up.
#include <google/protobuf/stubs/logging.h>
#include <google/protobuf/stubs/scoped_ptr.h>
#include <google/protobuf/stubs/mutex.h>
#include <google/protobuf/stubs/callback.h>
```
@ollieparanoid: Like wmww, I don't have a protobuf 3.8.0 to hand, but I think that in the headers that use `mir_protobuf.pb.h` we just need:
```c++
#include "mir_protobuf.pb.h"
#if GOOGLE_PROTOBUF_VERSION >= 3008000
#include <google/protobuf/stubs/callback.h>
#endif
```
And we might need that in `src/client/rpc/mir_basic_rpc_channel.cpp` too.

AlanGriffiths:
> Cool, I'll take a shot at it some time soon and then update the patch.

I took a shot - could you test with 3.8? https://github.com/MirServer/mir/pull/920/

PR #918, opened 2019-07-12
Wayland frontend: update window size more correctly

wmww:
As far as I know, the previous behavior wasn't causing any bugs. It was, however causing `modify_window()` to get called unnecessarily each frame in some cases (tested with wlroots/examples/simple). The reason was that `pending_window_size` wasn't getting set, so `committed_window_size` wasn't getting updated so it thought the size had changed each time. Now it correctly managers pending and current size, and keeps track of if the size has been set explicitly.

This entire system probably will need a rewrite at some point, but now is not the time.

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.h
+++ b/src/server/frontend_wayland/window_wl_surface_role.h
@@ -87,8 +87,15 @@ public:
 protected:
     std::shared_ptr<bool> const destroyed;
 
-    std::experimental::optional<geometry::Size> window_size() const;
-    std::experimental::optional<geometry::Size> requested_window_size(); // Window size requested by Mir
+    /// The size the window will be after the next commit
+    auto pending_size() const -> geometry::Size;
+
+    /// The size the window currently is (the committed size, or a reasonable default if it has never committed)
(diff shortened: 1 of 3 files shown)


src/server/frontend_wayland/window_wl_surface_role.h
```diff
@@ -102,8 +109,16 @@ class WindowWlSurfaceRole : public WlSurfaceRole
...
     std::shared_ptr<WlSurfaceEventSink> const sink;
     std::unique_ptr<scene::SurfaceCreationParameters> const params;
-    std::experimental::optional<geometry::Size> pending_window_size;
-    std::experimental::optional<geometry::Size> committed_window_size;
+
+    /// The explicitly set (not taken from the surface buffer size) uncommitted window size
+    std::experimental::optional<geometry::Size> pending_explicit_size;
+
+    /// If the committed window size was set explicitly, rather than being taken from the buffer size
+    bool committed_size_set_explicitly{false};
```
AlanGriffiths: Having both optional `committed_size` _and_ a boolean to describe how to interpret it is a bit jarring. It smells of a missing abstraction but I don't have a better idea right now.

PR #919, opened 2019-07-13
Printf common WLCS integration error

wmww:
Every time a new shell surface protocol is added, I forget to add it to the `is_window` list. That's all good and well, except the error gets eaten by Mir and I end up digging through GDB to solve the strange bug. This clarifies the error, and more importantly, actually shows it.


AlanGriffiths:
> It also draws my attention to the near duplication between `mir::frontend::get_window()` and `bool const is_window = ...`.

On reflection all this logic happens while creating the Mir window, so `get_window()` has nothing to return yet.

bors r+

PR #926, opened 2019-07-16
LayerShellSurface: configure correctly

wmww:
Correct layer shell behavior so that it passes the new WLCS tests. On top of #925.

### DIFF
--- a/src/server/frontend_wayland/layer_shell_v1.cpp
+++ b/src/server/frontend_wayland/layer_shell_v1.cpp
@@ -111,6 +111,7 @@ private:
     void destroy() override;
 
     // from WindowWlSurfaceRole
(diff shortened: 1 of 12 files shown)


wmww:
bors try

src/server/frontend_wayland/layer_shell_v1.cpp
```diff
@@ -282,16 +284,42 @@ auto mf::LayerSurfaceV1::get_exclusive_rect() const -> std::experimental::option
     }
 }
 
+void mf::LayerSurfaceV1::configure()
+{
+    // TODO: Error if told to configure on an axis the window is not streatched along
```
AlanGriffiths: s/streatched/stretched/

PR #933, opened 2019-07-19
Delete setup-partial-armhf-chroot - it's of no use now

AlanGriffiths:
Delete setup-partial-armhf-chroot - it's of no use now.


wmww:
What was this for, and why did it become non-useful?

AlanGriffiths:
> What was this for, and why did it become non-useful?

It was used for setting up a cross-build environment on a development machine to target the phone stack. Mostly for working on the libhybris/android based platform.

It became non-useful because we're not focussing on phones, not maintaining a libhybris/android based platform and, for the armhf hardware that interested in, there are other, more reliable, ways to build.

PR #934, opened 2019-07-19
Track registering miral::WaylandExtensions objects with mir::Servers and error if there is more than one

### LINKED ISSUE #875: Creating and passing two WaylandExtensions objects to MirRunner::run_with() causes inconsistent behaviour
We should be able to detect and diagnose this. C.f. #872

AlanGriffiths:
Track registering miral::WaylandExtensions objects with mir::Servers and error if there is more than one. (Fixes #875)

### DIFF
--- a/src/miral/wayland_extensions.cpp
+++ b/src/miral/wayland_extensions.cpp
@@ -27,7 +27,9 @@
 #include <mir/options/option.h>
 #include <mir/options/configuration.h>
 
+#include <mutex>
 #include <set>
+#include <vector>
 
 namespace mo = mir::options;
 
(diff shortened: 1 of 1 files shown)


src/miral/wayland_extensions.cpp
```diff
@@ -40,6 +42,52 @@ auto vec2set(std::vector<std::string> vec) -> std::set<std::string>
...
+
+    static void remove_extension(void* extension)
+    {
+        std::lock_guard<decltype(mutex)> lock{mutex};
+
+        for (auto x = begin(extensions); x != end(extensions); ++x)
+        {
+            if (x->extension == extension)
+            {
+                extensions.erase(x);
```
wmww: I was unaware this was safe. I assume you know what you're doing, but perhaps a `std::remove_if` would be more obviously safe?

src/miral/wayland_extensions.cpp
```diff
@@ -40,6 +42,52 @@ auto vec2set(std::vector<std::string> vec) -> std::set<std::string>
 {
     return std::set<std::string>{vec.begin(), vec.end()};
 }
+
+struct StaticExtensionTracker
```
wmww: Can we get some documentation on this as to what it is and why it's needed? `StaticExtensionTracker` does not make it's purpose immediately obvious.

src/miral/wayland_extensions.cpp
AlanGriffiths: Why wouldn't it be safe? There is no iterator being invalidated or incremented past `end(extensions)`.

AlanGriffiths:
bors r=wmww

PR #937, opened 2019-07-25
Allow frontend to add scene observer

wmww:
This gives a path for the Wayland frontend to add a `scene::Observer` so it can get notified about surfaces that are created or destroyed.

I also made some classes inherit from `scene::NullObserver` instead of `scene::Observer` so they don't have to override methods they don't use.


wmww:
This is going to need updating

PR #939, opened 2019-07-25
Switch from /** */ comments to /// comments

wmww:
I prefer `///`. I thought `/** */` was standard in Mir, but I've been seeing more `///` comments and I don't want my relatively recent code to be part of the problem.

### DIFF
--- a/include/miral/miral/window_specification.h
+++ b/include/miral/miral/window_specification.h
@@ -125,35 +125,29 @@ public:
     auto confine_pointer() -> mir::optional_value<MirPointerConfinementState>&;
     auto userdata() -> mir::optional_value<std::shared_ptr<void>>&;
 
-    /**
-     * The depth layer of a child window is updated with the depth layer of its parent, but can be overridden
-     *  @{
-     */
+    /// The depth layer of a child window is updated with the depth layer of its parent, but can be overridden
+    ///@{
     auto depth_layer() const -> mir::optional_value<MirDepthLayer> const&;
     auto depth_layer() -> mir::optional_value<MirDepthLayer>&;
-    /** @}*/
+    ///@}
 
-    /**
-     * The set of window eges that are attched to edges of the output
-     * If attached to perpendicular edges, it is attached to the corner where the two edges intersect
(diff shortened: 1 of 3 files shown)


include/miral/miral/window_specification.h
```diff
@@ -125,35 +125,29 @@ class WindowSpecification
...
-     *  @{
-     */
+    ///@}
+
+    /// The area over which the window should not be occluded
+    /// Only meaningful for windows attached to an edge
+    /// Setting to optional_value{} will not change the rect when this specification is applied
+    /// Setting to optional_value{optional_value{}} will clear the exclusive rect
+    ///@{
     auto exclusive_rect() const -> mir::optional_value<mir::optional_value<mir::geometry::Rectangle>> const&;
```
RAOF: Pre-existing, but... what? Why is this Optional<Optional<Rectangle>>? What are the semantics here? What does Some(None) mean here, and how does it differ from None(None)?

wmww: I'll make this a little more obvious

wmww:
Hopefully that's clearer. If not I'm open to suggestions on how to word it. I'd love to use rusty `Some(None)` syntax, but I'm afraid that wont be helpful to those not familiar with Rust. Would rather use "set" and "unset", as that's the language Mir uses.

AlanGriffiths:
I don't have a strong preference, but existing usage:
```bash
$ find -name \*.h | xargs grep -F '/**' | wc -l
1149
$ find -name \*.h | xargs grep -F '///' | wc -l
556
```
As `///` is repeated in multi-line comment I think it is in a substantial minority. Certainly not worth spending time changing from the majority style to it.

wmww:
I'm not planning to spend any more time changing them, I just did this at some point and didn't want to throw away the commit.

include/miral/miral/window_specification.h
RAOF: That's clearer, thanks.

AlanGriffiths:
bors r=RAOF

PR #940, opened 2019-07-26
Use scene::NullObserver and scene::NullSurfaceObserver where it makes sense

wmww:
This makes it easier to add surface properties without changing a bunch of observers that don't care


wmww:
That is correct, and intentional. My reasoning is that processing new properties is rarely needed aside from the places they are created for. For example since there was not previously any session property, all observers work correctly without being notified of the surface session. If we add that property, the only observer that needs it is the one we are creating the property for. Others may find it useful in the future, but it's not essential that all observers are updated to explicitly ignore it immediately.

PR #941, opened 2019-07-26
Remove ubuntu-1810 from spread.yaml

wmww:
`ubuntu-os-cloud/ubuntu-1910` [doesn't seem to be a thing](https://travis-ci.org/MirServer/mir/jobs/564109665)

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -23,7 +23,7 @@ backends:
             - ubuntu-19.04-64:
                 image: ubuntu-os-cloud/ubuntu-1904
             - ubuntu-devel-64:
-                image: ubuntu-os-cloud/ubuntu-1810
+                image: ubuntu-os-cloud-devel/ubuntu-1910
             - fedora-29-64
             - fedora-30-64
             - fedora-rawhide-64


spread.yaml
```diff
@@ -22,8 +22,6 @@ backends:
                 image: ubuntu-os-cloud/ubuntu-1804-lts
             - ubuntu-19.04-64:
                 image: ubuntu-os-cloud/ubuntu-1904
-            - ubuntu-devel-64:
-                image: ubuntu-os-cloud/ubuntu-1810
```
Saviq: This is actually a typo. This should say `1910` instead.

.travis.yml
```diff
@@ -43,7 +43,6 @@ env:
   - DEBEMAIL="mir-ci-bot@canonical.com"
   - SPREAD_PATH=/tmp/spread
   matrix:
-  - SYSTEM=google:ubuntu-devel-64 VARIANT=amd64
```
Saviq: This should remain.

spread.yaml
wmww: I tried that and it didn't work: https://travis-ci.org/MirServer/mir/jobs/564109665

```
2019-07-26 17:10:09 Allocating google:ubuntu-devel-64...
2019-07-26 17:10:09 Cannot allocate google:ubuntu-devel-64: cannot find any Google image matching "ubuntu-os-cloud/ubuntu-1910" on project "ubuntu-os-cloud"
2019-07-26 17:10:09 Successful tasks: 0
2019-07-26 17:10:09 Aborted tasks: 1
error: unsuccessful run
```

Saviq: Make this `1904` then, it will get upgraded to devel in CI.

PR #942, opened 2019-07-29
Let frontend look up a surface with the SurfaceId

wmww:
This adds `frontend::Session::get_surface_id(Surface*)`, which is implemented efficiently using a new id-to-shared_ptr map. The new map also allows for a more efficient implementation of `ApplicationSession::destroy_surface()`. Previously a single surface destroy call was O(N) based on the current number of surfaces. Now it's the complexity of a `map::find()` (O(log n)).

### DIFF
--- a/include/server/mir/frontend/session.h
+++ b/include/server/mir/frontend/session.h
@@ -51,7 +51,9 @@ class Session
 public:
     virtual ~Session() = default;
 
+    /// Throws std::out_of_range if the surface is invalid
     virtual SurfaceId get_surface_id(Surface* surface) const = 0;
+    /// Throws std::out_of_range if the id is invalid
     virtual std::shared_ptr<Surface> get_surface(SurfaceId surface) const = 0;
 
     virtual std::shared_ptr<BufferStream> get_buffer_stream(BufferStreamId stream) const = 0;
--- a/include/server/mir/frontend/session.h
+++ b/include/server/mir/frontend/session.h
@@ -51,6 +51,7 @@ class Session
 public:
     virtual ~Session() = default;
 
+    virtual SurfaceId get_surface_id(Surface* surface) const = 0;
     virtual std::shared_ptr<Surface> get_surface(SurfaceId surface) const = 0;
 
     virtual std::shared_ptr<BufferStream> get_buffer_stream(BufferStreamId stream) const = 0;
--- a/include/test/mir/test/doubles/stub_session.h
+++ b/include/test/mir/test/doubles/stub_session.h
@@ -32,8 +32,9 @@ struct StubSession : scene::Session
 {
     StubSession(pid_t pid = -1);
 
-    std::shared_ptr<frontend::Surface> get_surface(
-        frontend::SurfaceId surface) const override;
+    frontend::SurfaceId get_surface_id(frontend::Surface* surface) const override;
+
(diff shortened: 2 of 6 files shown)


include/server/mir/frontend/session.h
```diff
@@ -51,6 +51,9 @@ class Session
 public:
     virtual ~Session() = default;
 
+    /// Throws std::out_of_range if the surface is invalid
+    virtual SurfaceId get_surface_id(Surface* surface) const = 0;
```
RAOF: I think we'd typically spell this `virtual SurfaceId get_surface_id(std::weak_ptr<Surface> surface) const = 0`.

Generally I try to avoid raw pointers (at least in interfaces); either one of the various smart pointers, or a reference, tell you more about how thing will be used, what valid values are, and so on.

wmww:
The point of this PR is to assist with the implementation of wlr-foreign-toplevel-management. I have a `scene::Observer` hooked up, which alerts the frontend code when new surfaces appear. `scene::Observer::surface_added()` is sent a raw `scene::Surface*`, and I need to be able to manipulate surfaces. I've experimented with a few possible architectures, and the cleanest seemed to be to map the raw pointer to a `SurfaceId`, then use that to call `modify_surface()`.

include/server/mir/frontend/session.h
AlanGriffiths: I think raw pointers are fine. They (should) imply no transfer of ownership and therefore make no lifetime guarantees beyond the duration of the call.

AlanGriffiths:
> The new map also allows for a more efficient implementation of `ApplicationSession::destroy_surface()`. Previously a single surface destroy call was O(N) based on the current number of surfaces. Now it's the complexity of a `map::find()` (O(log n)).

The efficiency argument is suspect. A session would need to be hold an awful lot of surfaces before we amortize the cost of using a map instead of a vector.

AlanGriffiths:
> I've experimented with a few possible architectures, and the cleanest seemed to be to map the raw pointer to a `SurfaceId`, then use that to call `modify_surface()`.

As you know from conversation last week I have the same instincts about this as @RAOF, at best this reduces the code written now at the cost of increasing technical debt. We should be implementing the things wayland_frontend needs directly, not building on frontend infrastructure we plan to deprecate.

wmww:
I'm now not convinced `frontend::Shell` and `frontend::SurfaceId` are on the path to deprecation. I don't have a problem with getting rid of them, but they appear all through the frontend, session and shell. They serve the purpose of providing a handle to a surface owned by a session. This could be done by a `std::weak_ptr` I suppose, but actually making the switch may have unintended consequences and will take a lot of work.

I can implement foreign toplevel management without them, all I need is to change `scene::Observer` to provide a `std::weak_ptr<scene::Surface>` instead of a raw pointer.

AlanGriffiths:
The reason `frontend::SurfaceId` exists is to have a serializable identifier for a surface in IPC to the mirclient library. That is all legacy that we plan to deprecate.

It is possible that `frontend::Shell` could continue to provide service, but much of it[*] exists to translate `SurfaceId` to `Surface`.

* technically, `FrontendShell`.

wmww:
Well.. how *should* we hold non-owning references to surfaces? `std::shared_ptr<scene::Surface>`? `std::weak_ptr<scene::Surface>`? Which of those should `scene::Observer` provide?

wmww:
After some discussion it appears this is not the rout we're taking.

PR #943, opened 2019-07-29
Add scene::Shell::add_observer/remove_observer

wmww:
Allows anywhere with access to `scene::Scene` to add a `scene::Observer`.


AlanGriffiths:
Why? Even looking at the follow-up PR the motivation is unclear.

If, say, `WaylandConnector` or `configure_wayland_extensions()`, needs to add and remove observers, why not pass `the_scene()` as an argument and use `add_observer()` directly?

wmww:
Going to do this another way

PR #944, opened 2019-07-29
Add surface properties

wmww:
Adds or partially adds the following 3 properties to surfaces. I'm putting them all in one PR to reduce merge conflicts and for simplicity. If you'd rather review them separately, say the word and I'll do them one at a time.

* Focus state
  * Needed because it appears to be missing from `scene::Surface`
  * Already stored on `BasicSurface`, and accessible with the deprecated attribute system
  * Needed to add a `scene::Surface` setter and getter
  * Did not need to get added to the surface observer, as it is already handled in `attrib_changed()`
  * No changes to the frontend or MirAL needed
* Application ID
  * Added to `scene::Surface` and `scene::SurfaceObserver`
  * Added to `SurfaceSpecification` and `SurfaceCreationParameters`
  * Set from the Wayland frontend
  * Exposed in MirAL, as it might be useful to the shell
* Session
  * Added `scene::Surface` setter and getter
  * Added to the observer
  * No changes to the frontend or MirAL needed
  * Set from `ApplicationSession`
  * Application session needed to support `std::shared_from_this<>` to be able to set the property
  * Needed to be a weak or shared session pointer so it could be used in calls to `modify_surface()`

### DIFF
--- a/include/miral/miral/window_specification.h
+++ b/include/miral/miral/window_specification.h
@@ -149,6 +149,13 @@ public:
     auto exclusive_rect() -> mir::optional_value<mir::optional_value<mir::geometry::Rectangle>>&;
     ///@}
 
+    /// The D-bus service name and basename of the app's .desktop file
+    /// See http://standards.freedesktop.org/desktop-entry-spec/
+    ///@{
+    auto application_id() const -> mir::optional_value<std::string> const&;
+    auto application_id() -> mir::optional_value<std::string>&;
+    ///@}
+
 private:
     struct Self;
     std::unique_ptr<Self> self;
--- a/debian/libmiral3.symbols
+++ b/debian/libmiral3.symbols
@@ -392,10 +392,14 @@ libmiral.so.3 libmiral3 #MINVER#
  (c++)"vtable for miral::WindowManagementPolicy::ApplicationZoneAddendum@MIRAL_2.6" 2.6.0
  (c++)"miral::WindowManagementPolicy::ApplicationZoneAddendum::from(miral::WindowManagementPolicy*)@MIRAL_2.6" 2.6.0
  MIRAL_2.7@MIRAL_2.7 2.7.0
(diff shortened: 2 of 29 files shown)


wmww:
A concern I have with the session property is that there's no way to lock the session or the shell from the outside. This means that the following could happen with no way to prevent it:
```c++
auto session = surface->session();
// The surface is removed from the session on another thread
assume_surface_blongs_to_session(session, surface)
```

include/miral/miral/window_specification.h
```diff
@@ -149,6 +149,13 @@ class WindowSpecification
     auto exclusive_rect() -> mir::optional_value<mir::optional_value<mir::geometry::Rectangle>>&;
     ///@}
 
+    /// The D-bus service name and basename of the app's .desktop file
+    /// See http://standards.freedesktop.org/desktop-entry-spec/
```
AlanGriffiths: I don't follow the intention here. The equivalence of "D-bus service name" and "the basename" is only required for "Applications that support being launched by D-Bus". And is this optional attribute a good value to chose for `application_id()`?

AlanGriffiths:
>     * Focus state
>       
>       * Needed because it appears to be missing from `scene::Surface`

Why would it be needed? In my head the Surface with focus is a property of the scene, not of the surface.

AlanGriffiths:
> * Session
>   
>   * Added `scene::Surface` setter and getter

As already discussed, I fail to imagine where this would be needed.

include/miral/miral/window_specification.h
wmww: I'm just describing what it will be set to by XDG shell. Better if I change it to "Generally the D-bus service name..."?

wmww:
Getting a surface's session is needed when you have only a `std::shared_ptr<scene::Surface>` (as is obtained by the scene observer), but not the surface's session. Going forward, I'm considering changing `shell::modify_surface()` to obtain a session from the given surface instead of requiring all callers to have one. Since a surface belongs to only one session, I don't see why it should be necessary to pass around both.

wmww:
Will split this into smaller PRs

PR #948, opened 2019-07-30
Send shared_ptr instead of raw surface pointer from scene::Observer

wmww:
We discussed earlier using `weak_ptr` instead of `shared_ptr`, but I decided that would require otherwise unnecessary locks and checks. It would always be safe to assume the given weak pointer would be valid during the notification calls, but then why not indicate that with a shared_ptr? I think its fine to pass share_ptrs around as long as we only store weak pointers.

### DIFF
--- a/include/server/mir/scene/observer.h
+++ b/include/server/mir/scene/observer.h
@@ -33,18 +33,19 @@ class Surface;
 class Observer
 {
 public:
-    virtual void surface_added(Surface* surface) = 0;
-    virtual void surface_removed(Surface* surface) = 0;
+    virtual void surface_added(std::shared_ptr<Surface> surface) = 0;
+    virtual void surface_removed(std::shared_ptr<Surface> surface) = 0;
     virtual void surfaces_reordered() = 0;
-    
-    // Used to indicate the scene has changed in some way beyond the present surfaces
-    // and will require full recomposition.
+
+    /// Used to indicate the scene has changed in some way beyond the present surfaces
+    /// and will require full recomposition.
     virtual void scene_changed() = 0;
 
-    // Called at observer registration to notify of already existing surfaces.
-    virtual void surface_exists(Surface* surface) = 0;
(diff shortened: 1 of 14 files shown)


RAOF:
Is this meant to be stacked on the `NullSurfaceObserver` changes?

wmww:
Yes, and there's a bunch of conflicts when I try to rebase it out. Can we make a decision over there first? I don't mind dealing with the conflicts, but no need to do it twice.

include/server/mir/scene/observer.h
```diff
@@ -33,18 +33,19 @@ class Surface;
 class Observer
 {
 public:
-    virtual void surface_added(Surface* surface) = 0;
-    virtual void surface_removed(Surface* surface) = 0;
+    virtual void surface_added(std::shared_ptr<Surface> surface) = 0;
```
AlanGriffiths: Please use `std::shared_ptr<Surface>const&` to avoid unnecessary copies and the barriers they need.

wmww: Ack, will change when I get back to this

wmww: Done

src/server/input/surface_input_dispatcher.cpp
```diff
@@ -150,7 +152,10 @@ mi::SurfaceInputDispatcher::SurfaceInputDispatcher(std::shared_ptr<mi::Scene> co
       started(false)
 {
     scene_observer = std::make_shared<InputDispatcherSceneObserver>(
-        [this](ms::Surface* s){surface_removed(s);},
+        std::bind(
+            std::mem_fn(&SurfaceInputDispatcher::surface_removed),
+            this,
+            std::placeholders::_1),
```
AlanGriffiths: Lambdas should be preferred to std::bind (that's so very C++11). Something like:
```c++
        [this](auto const& s) { surface_removed(s); },
```

PR #952, opened 2019-08-02
Send wl_buffer.release on the Wayland thread

### LINKED ISSUE #951: Mir sends corrupt Wayland messages
There's an intermittent crash that manifests as either the client or Mir complaining of invalid data. GTK surfaces this as "Value is too large for defined data type", and Mir shows "error in client communication" (only one of the two happens, not both). Spoiler alert: its a race condition caused by `~WlShmBuffer()` sending `wl_buffer.release` on the wrong thread. I already have a fix, am just posting this so the issue is searchable and whatnot.

wmww:
Appears to fix a longstanding crash. The Wayland debug feature I built to detect thread errors tells me this was the only instance of such a problem in Mir (at least in the standard features I've tested with). Fixes #951

### DIFF
--- a/src/server/frontend_wayland/wl_surface.cpp
+++ b/src/server/frontend_wayland/wl_surface.cpp
@@ -337,6 +337,7 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
             {
                 mir_buffer = WlShmBuffer::mir_buffer_from_wl_buffer(
                     buffer,
+                    executor,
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
Running on 18.04 (in case it matters).

1. miral-app
2. Move cursor over the QTerminal window
3. Segmentation fault

wmww:
Reproduced and fixed the issue with QTerminal

PR #954, opened 2019-08-02
Exclusive zone fixes

### LINKED ISSUE #953: Full screen windows should ignore exclusive zones
This might not apply to master, but in the current MATE branch applications which are fullscreen still do not overlap the panels. They should.

wmww:
Fixes the following issues:
* Fixes #953: Makes new fullscreen windows ignore exclusive zones
* Places windows correctly when they are fullscreened
* Adjusts maximized windows when an exclusive window becomes not attached
* Tests everything mentioned


AlanGriffiths:
This all looks reasonable. And there are even tests!

Is there a handy test client we can use to create exclusive zones and try visually?

PR #966, opened 2019-08-16
Move wayland event unpacking to event sink

wmww:
This probably wasn't really required, but I started fiddling with things and one thing lead to another. This moves all the code to unpack Mir events into `WlSurfaceEventSink` and out of the Wayland objects. It also generally breaks up the code into more logical chunks with better names and more type safety (ints are replaced by geometry types and chrono types where possible). I've tested the keyboard and mouse portions, but not touch. The WLCS touch tests pass even when I know it's broken (I should probably work on them at some point), so manual testing is needed. I'll comment when I've got it done building and tested on my under-powered touch machine.

My next PR will be changing the surface event sink to a surface observer, which cuts out large pieces of the Mir event system from the Wayland code paths. After that *should* be the final PR that ties it all together.

### DIFF
--- a/src/server/frontend_wayland/wayland_connector.cpp
+++ b/src/server/frontend_wayland/wayland_connector.cpp
@@ -231,11 +231,6 @@ void setup_new_client_handler(wl_display* display, std::shared_ptr<mf::Shell> co
 }
 }
 
-int64_t mir_input_event_get_event_time_ms(const MirInputEvent* event)
-{
-    return mir_input_event_get_event_time(event) / 1000000;
-}
-
 class WlCompositor : public wayland::Compositor::Global
 {
 public:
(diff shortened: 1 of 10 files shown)


wmww:
There's a bug in which the position of the initial "down" event is offset from that of the subsequent "move" events. Presumably its to do with window geometry. Not sure if it's preexisting or not (it's hardly noticeable in practice without wayland-debug), but either way, it should be fixed here.

wmww:
Issue was preexisting and is now fixed. I also tested it and related behavior: https://github.com/MirServer/wlcs/pull/133. I'm a lot more confident now that WLCS (with the PR) passing indicates touch is actually implemented correctly. Will also manually test shortly.

wmww:
All looks good in my manual testing. The offset issue was the root cause of a few minor glitches I saw before.

PR #970, opened 2019-08-19
Allow Fedora rawhides builds to fail

AlanGriffiths:
We're seeing "GPG check FAILED" errors in CI.

Temporarily ignore the failing builds.


Saviq:
Eesh.

bors r+

PR #977, opened 2019-08-29
Add ImmediateExecutor

wmww:
Adds an executor that just runs the function synchronously, in the current thread. In a branch I'm working on I pass one of these to `registrar->register_interest()`.

The default executor throws the call on the main loop, which allows it to get called after `unregister_interest()` has been called, and the observer has been destroyed. This race condition foot-gun results in an intermittent crash that is difficult to diagnose. Perhaps we should rethink how the observer system works, and how we can prevent this sort of thing. As far as I can tell, the only way to safely register an observer with the default executor is to make sure that it stays alive for as long as messages are being sent, or it stays alive for at lase one main loop iteration after it is unregistered.

### DIFF
--- a/include/platform/mir/executor.h
+++ b/include/platform/mir/executor.h
@@ -48,7 +48,6 @@ public:
      */
     virtual void spawn(std::function<void()>&& work) = 0;
 
-protected:
     virtual ~Executor() = default;
 };
 
--- a/include/platform/mir/immediate_executor.h
+++ b/include/platform/mir/immediate_executor.h
@@ -1,0 +1,40 @@
+/*
+ * Copyright © 2019 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


RAOF:
> The default executor throws the call on the main loop, which allows it to get called after unregister_interest() has been called, and the observer has been destroyed

Hm? How do you manage to do this? `ObserverRegistrar::register_interest` takes a `weak_ptr` to the observer, and only ever calls into live observers.

There *is* a problem with `ObserverRegistrar` in that `register_interest` takes a reference to an `Executor`; this was following the C++ drafts where Executors were expected to be semi-global and definitely statically-outlive anything that was spawned upon them.

wmww:
> ObserverRegistrar::register_interest takes a weak_ptr to the observer, and only ever calls into live observers.

Except it does the check on the incoming thread, then spawns the function with a raw pointer to the observer (see https://github.com/MirServer/mir/blob/beaaa1655994afb952fc478df2676501e6201fcb/src/include/server/mir/observer_multiplexer.h#L117). Ideally though it wouldn't check if the weak pointer is alive, but rather if the interest has been unregistered.

RAOF:
Ah, quite true.

We could easily solve that by passing the `shared_ptr` through to `bind`, though (or the `weak_ptr` through a lambda).

It *won't* call observers that have been unregistered. Although unregistration can race with the `spawn` it's surprisingly tricky to make `unregister_interest` block until it is guaranteed that no thread is within the observer.

PR #978, opened 2019-08-30
Split arch-specific part of libmirwayland-dev into a -bin package 

RAOF:
This makes the PPA packaging match the Ubuntu archive packaging, and is necessary should anyone actually want to install `libmirwayland-dev` for multiple architectures (eg: using a cross-compiler)

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -586,8 +586,23 @@ Pre-Depends: ${misc:Pre-Depends}
 Depends: libmirwayland0 (= ${binary:Version}),
          libmircore-dev (= ${binary:Version}),
          ${misc:Depends},
+         libmirwayland-bin (= ${binary:Version})
 Description: Display server for Ubuntu - generated wrappers for Wayland
  protocol extensions.
  .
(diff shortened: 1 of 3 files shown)


debian/control
```diff
@@ -586,8 +586,25 @@ Pre-Depends: ${misc:Pre-Depends}
...
          libmircore-dev (= ${binary:Version}),
          ${misc:Depends},
+         libmirwayland-bin (= ${binary:Version})
 Description: Display server for Ubuntu - generated wrappers for Wayland
  protocol extensions.
  .
  Contains the developer files for using the shared library containing generated
  wrappers for Wayland protocol extensions
+
+Package: libmirwayland-bin
```
Saviq: Should this be `…-dev-tools` rather, to signify better what it is?

RAOF:
I think there's a convention of -bin packages; see also libglib2.0-bin

Saviq:
Then there's `qt*-dev-tools`, but _fine_ ;)

bors r+

PR #981, opened 2019-09-02
ObserverMultiplexer: Make unregister_interest deterministic

RAOF:
It would be very convenient when reasoning about code if `unregister_interest()`
guaranteed that once it returned no further calls to the observer would be
made.

This comes at the cost of preventing multiple observations to be made in parallel. This shouldn't be too much of a problem, as our existing `Executor`s - `GLibMainLoop` and `WaylandExecutor` - are single-threaded and so can't have multiple observations executing at once anyway.

### DIFF
--- a/include/server/mir/observer_registrar.h
+++ b/include/server/mir/observer_registrar.h
@@ -73,8 +73,12 @@ public:
      * Remove an observer from the set notified of all observations.
      *
      * This is threadsafe and can be called in any context.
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
"Could not parse store response: b'<html><body><h1>503 Service Unavailable</h1>\nNo server is available to handle this request.\n</body></html>\n'"

bors r+

Issue #980, opened 2019-09-01
Mir does not render layer_shell clients on the correct monitor


Pajn:
This can be reproduced most easily with [swaybg](https://github.com/swaywm/swaybg) and [mirco](https://github.com/wmww/mirco/) and two virtual monitors with `--x11-output 1024x768:1024x768`.

After starting swaybg, notice in its how it correctly creates two windows but both of them ends up on the first monitor making it seem like only one was rendered.

I have also confirmed that `output_id()` of the `WindowSpecification` passed to `place_new_window` is always `0` and `top_left` are equal as well.

I assume that this is the problem:
https://github.com/MirServer/mir/blob/master/src/server/frontend_wayland/layer_shell_v1.cpp#L202

wmww:
Yes, this is still and incomplete part of our Layer Shell implementation. I'll try to get to this within the next two weeks.

Pajn:
With both #982 and #984 merged I can no longer find any issues with specifying output.

PR #982, opened 2019-09-05
Set output_id for LayerShell surfaces

Pajn:
This is a shot at #980 

While I know way to little to know if a lot of things are missing, it at least fixes the case described in #980

### DIFF
--- a/src/miral/window_specification.cpp
+++ b/src/miral/window_specification.cpp
@@ -71,7 +71,7 @@ miral::WindowSpecification::Self::Self(mir::shell::SurfaceSpecification const& s
     pixel_format(spec.pixel_format),
     buffer_usage(),
     name(spec.name),
-    output_id(spec.output_id),
+    output_id(),
     type(spec.type),
     state(spec.state),
     preferred_orientation(spec.preferred_orientation),
--- a/src/miral/window_specification.cpp
+++ b/src/miral/window_specification.cpp
@@ -71,7 +71,7 @@ miral::WindowSpecification::Self::Self(mir::shell::SurfaceSpecification const& s
     pixel_format(spec.pixel_format),
     buffer_usage(),
     name(spec.name),
-    output_id(),
+    output_id(spec.output_id),
     type(spec.type),
     state(spec.state),
     preferred_orientation(spec.preferred_orientation),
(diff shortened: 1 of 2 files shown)


Saviq:
Hi @Pajn, thanks for taking the time :)

We'd love to accept your contribution, and Mir is covered with the [Canonical Contributor License Agreement](https://ubuntu.com/legal/contributors). I couldn't find you in the current contributor list, did you sign the CLA before?

Pajn:
No, I haven't done that yet. I started filling out the form as an individual contributor but need help with the "Canonical Project Manager or contact" field. What should I write there?
Also while I ask, is "launchpad id" the launchpad username? I can not see a field named id on the launchpad profile page.

Saviq:
> No, I haven't done that yet. I started filling out the form as an individual contributor but need help with the "Canonical Project Manager or contact" field. What should I write there?

Drop me in there - "Michał Sawicz <michal.sawicz@canonical.com>".

> Also while I ask, is "launchpad id" the launchpad username? I can not see a field named id on the launchpad profile page.

Yes, that's it.

Pajn:
Now I have signed the CLA

Saviq:
Great :+1: 

I'll leave the Mir devs to look at this PR :)

RAOF:
The code looks good to me, but I haven't tested it yet.

Testing this properly might require `wlcs` improvements (particularly, the ability to create multiple test outputs), so I think this could land without a test for now.

If William doesn't get to it sooner I'll manually test and land this on Monday.

src/miral/window_specification.cpp
```diff
@@ -71,7 +71,7 @@ miral::WindowSpecification::Self::Self(mir::shell::SurfaceSpecification const& s
     pixel_format(spec.pixel_format),
     buffer_usage(),
     name(spec.name),
-    output_id(),
+    output_id(spec.output_id),
```
AlanGriffiths: This is unnecessary (see line 133) and wrong.

I was initially surprised it complies as it constructs a `optional_value<int>` with a `optional_value<graphics::DisplayConfigurationOutputId>`, but instead of taking the value of `spec.output_id.value().as_value()` it appears to take the value of `spec.output_id.operator bool()` and sets `output_id` to `0` or `1` accordingly.

There's clearly a gap in our test coverage as nothing breaks.

Pajn: Interesting :) The initializer syntax really confuses me about C++ but I can see what you mean now when you describe what happens.

What I could see from Rust (via the generated bindings so it might not be Mirs fault), output id was always set to 0 instead of not having a value so some gap in the test coverage around output_id seems very possible.

PR #983, opened 2019-09-06
Allow clipping windows to a specified area

Pajn:
I'm opening this to get some feedback and ask some questions around the concept.

Use case:
I have separate workspaces on each monitor and do not want windows near the edge to span multiple monitors and thus show up on another workspace.

The interesting bits are the new API introduced on Surface with `exclusive_display` and `set_exclusive_display` and the usage of that API in `SurfaceStack::scene_elements_for`. The rest is plumbing. 
On IRC I initially talked with @RAOF about filtering surfaces to a specific `CompositorID`, I have implemented this with filtering on a Rectangle instead as I did not see a good way to get the `CompositorID` in `miral` (and with it being a pointer to the DisplayBufferCompositor I don't think that is a good idea anyway). I used the `view_area` Rectangle because it was easy but I don't think that's actually a good idea either and is instead interested in adding a display id to `DisplayBuffer`. Is that reasonable or should I use another approach?
@RAOF also raised the question if miral already had enough hooks to implement the wanted functionality without this, though I can not see how it would. At least not with this approach. It can hide and show a surface but can not do that per display.

@AlanGriffiths envisioned a more powerful API:
>If it's relevant, I think clipping to a rectangle is probably right in the compositor. The policy should decide to set the rectangle to a zone, output, or something else.

Which is not possible with this approach. I can only leave this as an open question and let you decide what's best here.

The methods on `miral::Window` is only there to test out the functionality and a miral API will have to be decided so I'm not currently asking for feedback there but if you have ideas or pointers I'm of course interested in them.

### DIFF
--- a/include/miral/miral/window_info.h
+++ b/include/miral/miral/window_info.h
@@ -134,6 +134,10 @@ struct WindowInfo
     /// (only meaningful when the window is attached to an edge)
     void exclusive_rect(mir::optional_value<mir::geometry::Rectangle> const& rect);
 
+    /// Mir will not render anything outside this rectangle
+    auto clip_area() const -> mir::optional_value<mir::geometry::Rectangle>;
+    void set_clip_area(mir::optional_value<mir::geometry::Rectangle> zone);
+
 private:
     struct Self;
     std::unique_ptr<Self> self;
--- a/include/miral/miral/window_info.h
+++ b/include/miral/miral/window_info.h
@@ -136,7 +136,7 @@ struct WindowInfo
 
     /// Mir will not render anything outside this rectangle
     auto clip_area() const -> mir::optional_value<mir::geometry::Rectangle>;
-    void set_clip_area(mir::optional_value<mir::geometry::Rectangle> zone);
+    void set_clip_area(mir::optional_value<mir::geometry::Rectangle> const& area);
 
 private:
     struct Self;
--- a/include/miral/miral/window_info.h
+++ b/include/miral/miral/window_info.h
@@ -136,7 +136,7 @@ struct WindowInfo
 
     /// Mir will not render anything outside this rectangle
     auto clip_area() const -> mir::optional_value<mir::geometry::Rectangle>;
-    void set_clip_area(mir::optional_value<mir::geometry::Rectangle> const& area);
+    void clip_area(mir::optional_value<mir::geometry::Rectangle> const& area);
 
 private:
     struct Self;
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -474,10 +474,10 @@ MIRAL_2.7 {
 global:
   extern "C++" {
     miral::Output::id*;
-    miral::Window::exclusive_display*;
-    miral::Window::set_exclusive_display*;
     miral::WindowInfo::attached_edges*;
+    miral::WindowInfo::clip_area*;
     miral::WindowInfo::exclusive_rect*;
+    miral::WindowInfo::set_clip_area*;
     miral::WindowSpecification::attached_edges*;
     miral::WindowSpecification::exclusive_rect*;
   };
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -474,6 +474,8 @@ MIRAL_2.7 {
 global:
   extern "C++" {
     miral::Output::id*;
+    miral::Window::exclusive_display*;
+    miral::Window::set_exclusive_display*;
     miral::WindowInfo::attached_edges*;
     miral::WindowInfo::exclusive_rect*;
     miral::WindowSpecification::attached_edges*;
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -475,10 +475,15 @@ global:
   extern "C++" {
     miral::Output::id*;
     miral::WindowInfo::attached_edges*;
-    miral::WindowInfo::clip_area*;
     miral::WindowInfo::exclusive_rect*;
-    miral::WindowInfo::set_clip_area*;
     miral::WindowSpecification::attached_edges*;
     miral::WindowSpecification::exclusive_rect*;
   };
 } MIRAL_2.6;
+
+MIRAL_2.8 {
+global:
+  extern "C++" {
+    miral::WindowInfo::clip_area*;
+  };
+} MIRAL_2.7;
--- a/src/miral/window_info.cpp
+++ b/src/miral/window_info.cpp
@@ -659,9 +659,9 @@ auto miral::WindowInfo::clip_area() const -> mir::optional_value<mir::geometry::
 
     return surface->clip_area();
 }
-void miral::WindowInfo::set_clip_area(mir::optional_value<mir::geometry::Rectangle> zone)
+void miral::WindowInfo::set_clip_area(mir::optional_value<mir::geometry::Rectangle> area)
 {
     std::shared_ptr<mir::scene::Surface> const surface = self->window;
 
-    surface->set_clip_area(zone);
+    surface->set_clip_area(area);
 }
--- a/src/miral/window_info.cpp
+++ b/src/miral/window_info.cpp
@@ -659,7 +659,7 @@ auto miral::WindowInfo::clip_area() const -> mir::optional_value<mir::geometry::
 
     return surface->clip_area();
 }
-void miral::WindowInfo::set_clip_area(mir::optional_value<mir::geometry::Rectangle> area)
+void miral::WindowInfo::set_clip_area(mir::optional_value<mir::geometry::Rectangle> const& area)
 {
     std::shared_ptr<mir::scene::Surface> const surface = self->window;
 
--- a/src/miral/window_info.cpp
+++ b/src/miral/window_info.cpp
@@ -659,6 +659,7 @@ auto miral::WindowInfo::clip_area() const -> mir::optional_value<mir::geometry::
 
     return surface->clip_area();
 }
+
 void miral::WindowInfo::clip_area(mir::optional_value<mir::geometry::Rectangle> const& area)
 {
     std::shared_ptr<mir::scene::Surface> const surface = self->window;
--- a/src/miral/window_info.cpp
+++ b/src/miral/window_info.cpp
@@ -659,7 +659,7 @@ auto miral::WindowInfo::clip_area() const -> mir::optional_value<mir::geometry::
 
     return surface->clip_area();
 }
-void miral::WindowInfo::set_clip_area(mir::optional_value<mir::geometry::Rectangle> const& area)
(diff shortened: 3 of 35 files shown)


wmww:
It may make sense to use `miral::Zone` for this sort of thing. It's basically just a shared pointer to a rectangle.It *should* also guarantee that the wrapped rectangle is not mutable, but we might have some cleanup to do before that is the case. You don't want a shared pointer to a mutable rectangle, because there's no way to subscribe to changes.

As for naming, might I suggest `clip_zone` or `clip_area` (depending on if a `miral::Zone` is involved)? As Alan says, anything to do with displays appears to be inaccurate in the current iteration and `view_area` sounds a little bit vague to me.

Pajn:
I started discussing this in IRC but thought it might be best to move the discussion here.

I discovered that DiplayBuffers (and therefore) DisplayBufferCompositors can be shared for overlapping outputs with the kms backend. AlanG confirmed on IRC that these does not only apply to complete overlap (mirrored outputs) but can also be only partially overlapping monitors. Thus we can not filter on a DisplayBufferCompositors level, and thus we need clipping so I'm thinking about pivoting to having a general containment area instead of an exclusive display.

The APIs I'm thinking about:

```
optional_value<Rectangle> Surface::containment_area()
void Surface::set_containment_area(optional_value<Rectangle> area)

// Returns a rectangle from top_left() and size() but clipped to the containment_area()
Rectangle Surface::clipped_area() 
```

Filtering in `SurfaceStack::scene_elements_for` can still be done for surfaces having a `render_area` that does not intersect with the `DisplayBuffer::view_area` but when surfaces do intersect, the renderer will have to clip the surfaces and for giving information about that I'm thinking about adding `Rectangle clipped_area()` to Renderable. Depending on if we want the logic in Renderable or the Renderer, this might instead return `optional_value<Rectangle>` to avoid clipping when not needed.

Does that seem reasonable?

The above was written before I saw wmwws post, clip_area is probably a better name as this does not actually try to contain the window. As per the miral api I think it might make sense to only provide an api to set `clip_area` on the `Surface` to a Zone. A more general API can always be added later if a use case for that is seen.

wmww:
That sounds good. If there aren't technical limitations, I think clipping to a rectangle instead of an output is preferable.

> I think it might make sense to only provide an api to set clip_area on the Surface to a Zone. A more general API can always be added later if a use case for that is seen.

Users of MirAL can make new zones with arbitrary extents, so using zones shouldn't be in any way limiting.

AlanGriffiths:
> It may make sense to use `miral::Zone` for this sort of thing. It's basically just a shared pointer to a rectangle.It _should_ also guarantee that the wrapped rectangle is not mutable, but we might have some cleanup to do before that is the case. You don't want a shared pointer to a mutable rectangle, because there's no way to subscribe to changes.

I disagree regarding mutability. So long as access to the clipping rectangle is atomic (it isn't for the existing miral::Zone) I think having a shared pointer (or equivalent handle class) to a mutable abstraction is right. There should be no need to change the clipping rectangle for every associated window, just update the rectangle. Yes, that may need to trigger re-compositing, but that just implies clipping rectangles (and their changes) are known to the scene and compositor.

Pajn:
Now I have the basic concept working. I'm currently using a Rectangle instead of a Zone in the miral API due to some beginner problem with understanding the compiler errors. I will look into that tomorrow.

The C++ references still confuses me so I don't actually now how the mutability situation is and if I clone the rectangle anywhere on it's path to the renderer. Again more research for tomorrow.

I assume that you want unit tests on both the filtering in surface_stack.cpp and the enablement of clipping in renderer.cpp so I'll look into that afterwords.

wmww:
After some thought, I don't think using zones or sharing pointers to rectangles is helpful for this API. Passing a rectangle by value should be fine.

A common pattern is to pass around const references, and store values. That means you can pass a structure though several layers of functions calls and only make a copy when it needs to be stored after the function returns. Returned types should generally not be references. With this pattern, your function signatures would look like this:
```c++
    /// Mir will not render anything outside this rectangle
    auto clip_area() const -> mir::optional_value<mir::geometry::Rectangle>;
    void set_clip_area(mir::optional_value<mir::geometry::Rectangle> const& zone);
```
This is functionally the same as pass-by-value, just a little more efficient.

I don't know if there will be a trivial way to test that the windows are actually visually clipped. What will be easier is testing that the input region is clipper properly.

Pajn:
Thank you, I will make sure to update the signatures to follow that pattern. Input regions are not clipped at all currently, so that I will have to fix as well.

Pajn:
Now I've updated signature according to said pattern, added input region clipping and some tests (which caught an error in the argument to glScissor).

I also cleaned up the filtering of completely clipped surfaces, though to do that I had to move the filtering from `SurfaceStack::scene_elements_for` to `BasicSurface::generate_renderables` which is against the initial request by @RAOF. Though with the change in direction and how much cleaner it got I'm not sure that still applies but I'm open to changing it if you still prefer.

include/miral/miral/window_info.h
```diff
@@ -134,6 +134,10 @@ struct WindowInfo
     /// (only meaningful when the window is attached to an edge)
     void exclusive_rect(mir::optional_value<mir::geometry::Rectangle> const& rect);
 
+    /// Mir will not render anything outside this rectangle
+    auto clip_area() const -> mir::optional_value<mir::geometry::Rectangle>;
+    void set_clip_area(mir::optional_value<mir::geometry::Rectangle> zone);
```
wmww: This should also probably be a `const&`.

src/miral/symbols.map
```diff
@@ -475,7 +475,9 @@ global:
   extern "C++" {
     miral::Output::id*;
     miral::WindowInfo::attached_edges*;
+    miral::WindowInfo::clip_area*;
```
AlanGriffiths: These should be added to a new 2.8 stanza. (This is not your fault, we have failed to update src/miral/regenerate-miral-symbols-map.py since the last release - see #995.)

Once that is fixed they are also needed in debian/libmiral3.symbols

wmww: You have to commit the debian symbols to git, but the changes should be done either automatically or by `make regenerate-miral-symbols-map`

Pajn: I rebased and regenerated the symbols to get them into 2.8. But I can't figure out how to regenerate the symbols in `debian/libmiral3.symbols`. It doesn't do it automatically and `make regenerate-miral-symbols-map` only regenerates `src/miral/symbols.map`.

AlanGriffiths: I had a poke around, the regeneration only happens if you're on a debian based distro and the program `dpkg-gensymbols` is installed. (See `src/miral/CMakeLists.txt` for the gory details.)

Pajn: Hmm, I'm on Ubuntu and do have `dpkg-gensymbols` installed. Manually running `dpkg-gensymbols -plibmiral3` did include the new symbols either for some reason. However your tip about the cmake file led me to the python script `src/miral/check-and-update-debian-symbols.py` and manually running `python3 src/miral/check-and-update-debian-symbols.py build/lib miral 2.8.0 3` did generate the symbols.

If it's due to some bad setup on my computer then I guess it doesn't matter as the symbols are there now but it might be worth investigating if there's some buildchain problem (maybe with the update to 2.8?). I don't know anything about about gensymbols so it's very possible that I have screwed something up though.

wmww: IIRC you also have to be using GCC. If you are then idk

src/miral/window_info.cpp
```diff
@@ -652,3 +652,16 @@ void miral::WindowInfo::exclusive_rect(mir::optional_value<mir::geometry::Rectan
 {
     self->exclusive_rect = rect;
 }
+
+auto miral::WindowInfo::clip_area() const -> mir::optional_value<mir::geometry::Rectangle>
+{
+    std::shared_ptr<mir::scene::Surface> const surface = self->window;
+
+    return surface->clip_area();
+}
```
AlanGriffiths: Nit: should have a blank line between functions.

tests/unit-tests/input/test_cursor_controller.cpp
```diff
@@ -178,6 +178,14 @@ struct StubInputSurface : public mtd::StubSceneSurface
         cursor_image_ = image;
     }
 
+    mir::optional_value<mir::geometry::Rectangle> clip_area() const override
+    {
+        return mir::optional_value<mir::geometry::Rectangle>();
+    }
```
AlanGriffiths: There should be a blank line between function bodies.

src/server/scene/basic_surface.h
```diff
@@ -145,6 +145,9 @@ class BasicSurface : public Surface
     auto depth_layer() const -> MirDepthLayer override;
     void set_depth_layer(MirDepthLayer depth_layer) override;
 
+    optional_value<geometry::Rectangle> clip_area() const override;
```
RAOF: I think we want to move off `mir::optional_value` in favour of `std::experimental::optional` (and then `std::optional` when we can require C++17)

The miral bits are OK because consistency in API is more important, but we don't currently have any `mir::optional_value` uses here.

tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -140,6 +140,7 @@ struct SurfaceStack : public ::testing::Test
     std::shared_ptr<ms::SceneReport> const report = mr::null_scene_report();
     ms::SurfaceStack stack{report};
     void const* compositor_id{&default_params};
+    geom::Rectangle view_area;
```
RAOF: Is this some left-over code? I don't see what it connects to.

src/server/scene/basic_surface.h
Pajn: Done. It was mostly a search and replace job but I had to do conversion between `mir::optional_value` and `std::experimental::optional` in `window_info.cpp`

tests/unit-tests/scene/test_surface_stack.cpp
Pajn: Jup, left over from the first attempt that required the DisplayBuffers view_are to be passed in to `scene_elements_for`

Pajn:
>I think this API is more general than you actually need

Absolutely. I would only need support for clipping to a display, not any arbitrary area. And I assume the same would apply for about every other use case as well (if they need clipping at all). But I would argue that the API is motivated anyways because 1. Mir needs the support internally as DisplayBufferCompositors can handle multiple partially overlapping monitors and 2. The API isn't harder to use (well, a bit as you need to update the clipping area on `advise_output_update` but any use case needing clipping would probably need to handle that event anyway).

RAOF:
Thanks for your work!
bors r+

PR #987, opened 2019-09-09
Move MockDisplayConfigurationObserver to its own file

wmww:
Removes `MockDisplayConfigurationReport` from `tests/acceptance-tests/test_nested_mir.cpp`  and uses a new file containing `MockDisplayConfigurationObserver`


tests/acceptance-tests/test_nested_mir.cpp
```diff
@@ -52,6 +52,7 @@
 #include "mir/test/doubles/stub_cursor.h"
 #include "mir/test/doubles/stub_display_configuration.h"
 #include "mir/test/signal_actions.h"
+#include "../include/mir/test/doubles/mock_display_configuration_observer.h"
```
AlanGriffiths: That should be #include "mir/test/doubles/mock_display_configuration_observer.h"

AlanGriffiths:
Why?

Nothing specifically against this change, but introduce another file? Mocks are typically specific to the tests that use them and can happily live in the same file.

tests/acceptance-tests/test_nested_mir.cpp
wmww: oops

wmww:
> Why?

In a later PR I need `MockDisplayConfigurationObserver` in test_mediating_display_changer.cpp

> Mocks are typically specific to the tests that use them and can happily live in the same file

I find this policy odd. It adds overhead to creating a new test with the same mock, and leads to creating multiple implementations of the same mock that have to be maintained (just do a search for "class MockSurfaceObserver" to see what I mean). Since the mock's user isn't a unit test for `DisplayConfigurationObserver`, it would have have been reasonable to assume that the mock may be needed elsewhere even before it actually was.

tests/acceptance-tests/test_nested_mir.cpp
wmww: Actually that doesn't work here, and I'm not sure why

PR #988, opened 2019-09-09
Remove frontend types from Wayland

wmww:
Moves the Wayland frontend away from using mirclient specific types
`frontend::Shell` -> `shell::Shell`
`frontend::MirClientSession` -> `scene::Session`
`frontend::SurfaceId` -> `scene::Surface`

### DIFF
--- a/src/server/frontend_wayland/wayland_connector.h
+++ b/src/server/frontend_wayland/wayland_connector.h
@@ -47,6 +47,14 @@ namespace geometry
 {
 struct Size;
 }
+namespace shell
+{
+class Shell;
+}
+namespace scene
+{
+class Surface;
+}
 namespace frontend
 {
 class WlCompositor;
@@ -54,9 +62,6 @@ class WlSubcompositor;
 class WlApplication;
 class WlSeat;
 class OutputManager;
-class MirClientSession;
-class Shell;
-class Surface;
 class MirDisplay;
 class SessionAuthorizer;
 class DataDeviceManager;
@@ -71,14 +76,14 @@ public:
 
     virtual void run_builders(wl_display* display, std::function<void(std::function<void()>&& work)> const& run_on_wayland_mainloop);
 
-    void init(wl_display* display, std::shared_ptr<Shell> const& shell, WlSeat* seat, OutputManager* const output_manager);
+    void init(wl_display* display, std::shared_ptr<shell::Shell> const& shell, WlSeat* seat, OutputManager* const output_manager);
 
     auto get_extension(std::string const& name) const -> std::shared_ptr<void>;
 
 protected:
 
     void add_extension(std::string const name, std::shared_ptr<void> implementation);
-    virtual void custom_extensions(wl_display* display, std::shared_ptr<Shell> const& shell, WlSeat* seat, OutputManager* const output_manager);
+    virtual void custom_extensions(wl_display* display, std::shared_ptr<shell::Shell> const& shell, WlSeat* seat, OutputManager* const output_manager);
 
 private:
     std::unordered_map<std::string, std::shared_ptr<void>> extension_protocols;
@@ -91,7 +96,7 @@ public:
 
     WaylandConnector(
         optional_value<std::string> const& display_name,
-        std::shared_ptr<Shell> const& shell,
+        std::shared_ptr<shell::Shell> const& shell,
         std::shared_ptr<MirDisplay> const& display_config,
         std::shared_ptr<input::InputDeviceHub> const& input_hub,
         std::shared_ptr<input::Seat> const& seat,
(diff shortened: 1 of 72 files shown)


src/server/scene/application_session.cpp
```diff
@@ -89,9 +89,9 @@ mf::SurfaceId ms::ApplicationSession::next_id()
     return mf::SurfaceId(next_surface_id.fetch_add(1));
 }
 
-mf::SurfaceId ms::ApplicationSession::create_surface(
+auto ms::ApplicationSession::create_surface(
     SurfaceCreationParameters const& the_params,
-    std::shared_ptr<mf::EventSink> const& surface_sink)
+    std::shared_ptr<mf::EventSink> const& surface_sink) -> std::shared_ptr<scene::Surface>
```
AlanGriffiths: One of the advantages of post-fixing the return type is that the namespace is known and doesn't need restating.

src/server/frontend_wayland/wayland_connector.h
```diff
@@ -131,7 +136,7 @@ class WaylandConnector : public Connector
     std::unique_ptr<DataDeviceManager> data_device_manager_global;
     std::shared_ptr<Executor> const executor;
     std::shared_ptr<graphics::WaylandAllocator> const allocator;
-    std::weak_ptr<frontend::Shell> const weak_shell;
+    std::weak_ptr<shell::Shell> const weak_shell;
```
AlanGriffiths: This is arguably pre-existing, but why would the frontend have a weak pointer to the shell?

The shell should have no dependency on the frontend(s), so having a strong pointer seems safe and correct.

src/server/frontend_wayland/window_wl_surface_role.h
```diff
@@ -76,7 +80,7 @@ class WindowWlSurfaceRole : public WlSurfaceRole
     void set_title(std::string const& title);
     void initiate_interactive_move();
     void initiate_interactive_resize(MirResizeEdge edge);
-    void set_parent(optional_value<SurfaceId> parent_id);
+    void set_parent(std::experimental::optional<std::weak_ptr<scene::Surface>> const& parent);
```
AlanGriffiths: Not really related to this review but...

Seeing the spread of the `std::experimental::optional<std::weak_ptr<Type>>` idiom around the code suggests we're missing an abstraction (and a good name for it).

src/server/frontend_wayland/window_wl_surface_role.h
```diff
@@ -111,13 +115,23 @@ class WindowWlSurfaceRole : public WlSurfaceRole
 
     void commit(WlSurfaceState const& state) override;
 
+    /// Locks and returns weak_shell
+    /// Throws std::logic_error if the shell is null
+    auto shell_checked() -> std::shared_ptr<shell::Shell>;
+
+    /// Locks and returns weak_session
+    /// Throws std::logic_error if the session is null
+    auto session_checked() -> std::shared_ptr<scene::Session>;
```
AlanGriffiths: These "logic_errors" are sound like precondition failures, and exceptions should be reserved for post-condition failures. Throwing an exception when noticing a the program state is wrong is problematic as the situation is likely not recoverable.

Should the problem be reported by calling `mir::fatal_error()`?

tests/unit-tests/scene/test_application_session.cpp
```diff
@@ -1013,7 +1004,7 @@ TEST_F(ApplicationSessionSurfaceOutput, sends_correct_surface_details_to_surface
 
     for (int i = 0; i < 2 ; ++i)
     {
-        EXPECT_THAT(event[i].to_window_output()->surface_id(), Eq(ids[i].as_value()));
+        EXPECT_THAT(app_session.surface(mf::SurfaceId{event[i].to_window_output()->surface_id()}), Eq(surfaces[i]));
```
AlanGriffiths: Maybe worth a utility function in the fixture?

`auto surface_from_event()`?

src/server/frontend_wayland/window_wl_surface_role.h
wmww: Perhaps the surface role should simply hold a strong pointer to the session?

src/server/frontend_wayland/window_wl_surface_role.h
wmww: I'm not sure quite what abstraction you're thinking of, but in this particular case I determined it would be better to pass around an optional shared pointer.

src/server/frontend_wayland/window_wl_surface_role.h
AlanGriffiths: That depends. The prior code doesn't force the session to remain, just expects that the session exists during certain operations.

src/server/frontend_wayland/window_wl_surface_role.h
AlanGriffiths: The abstraction implemented as "an optional shared pointer" but with a meaningful name: `std::experimental::optional<std::weak_ptr<scene::Surface>> const&` is a bit of a mouthful.

PR #989, opened 2019-09-09
ApplicationSession: use a display config observer

wmww:
Currently, the session has a `send_display_config()` function. This PR makes the session use a `DisplayConfigurationObserver` to get that notification. In a coming PR, this will allow `BasicMirClientSession` to get that notification without having to pass the `MirClientSession` around. Wayland already uses a similar method. Stacked on top of #987

### DIFF
--- a/include/server/mir/graphics/display_configuration_observer.h
+++ b/include/server/mir/graphics/display_configuration_observer.h
@@ -115,9 +115,9 @@ public:
     /**
      * Notification that the session should send the given display configuration
      *
-     * It may or may not be a session configuration
+     * It may or may not be a session configuration. Used only to implement the mirclient API
      */
-    virtual void session_should_send_display_configuration(
+    virtual void configuration_updated_for_session(
         std::shared_ptr<scene::Session> const& session,
         std::shared_ptr<DisplayConfiguration const> const& config) = 0;
 
--- a/include/server/mir/graphics/display_configuration_observer.h
+++ b/include/server/mir/graphics/display_configuration_observer.h
@@ -112,6 +112,15 @@ public:
         std::shared_ptr<DisplayConfiguration const> const& failed_fallback,
         std::exception const& error) = 0;
 
+    /**
+     * Notification that the session should send the given display configuration
+     *
+     * It may or may not be a session configuration
+     */
+    virtual void session_should_send_display_configuration(
+        std::shared_ptr<scene::Session> const& session,
+        std::shared_ptr<DisplayConfiguration const> const& config) = 0;
+
 protected:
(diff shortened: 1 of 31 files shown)


wmww:
bors try

tests/unit-tests/scene/test_abstract_shell.cpp
```diff
@@ -93,6 +93,28 @@ struct MockSessionManager : ms::SessionManager
 {
     using ms::SessionManager::SessionManager;
 
+    // Must wrap the constructor as the NiceMock<> constructor can only take 10 arguments on older platforms
```
AlanGriffiths: It is amazing that a mock would require so many arguments.

wmww: Indeed. The mock inherits from `scene::SessionManager`, and previously used its constructor directly. Unfortunately I added an 11th argument to the `scene::SessionManager` constructor, which blew up the mock on older gtest versions. This is the simplest fix. The ideal solution is probably to make the mock inherit from `scene::SessionCoordinator` (the interface) instead of `scene::SessionManager` (the implementation), but such a change is out of scope of this PR.

include/server/mir/graphics/display_configuration_observer.h
```diff
@@ -112,6 +112,15 @@ class DisplayConfigurationObserver
         std::shared_ptr<DisplayConfiguration const> const& failed_fallback,
         std::exception const& error) = 0;
 
+    /**
+     * Notification that the session should send the given display configuration
+     *
+     * It may or may not be a session configuration
+     */
+    virtual void session_should_send_display_configuration(
```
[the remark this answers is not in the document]
wmww: It replaces `scene::Session::send_display_config()`. It's called in `mediating_display_changer.cpp` and It's implementation in `ApplicationSession::DisplayConfigurationObserver` calls `session->send_display_config()`. It's name is a bit odd. Open to suggestions.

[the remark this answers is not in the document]
AlanGriffiths: Oops! Searching the diff on the webpage doesn't always get the right answer.

wmww:
It's probably not the best way to do it. I set it up so I didn't have to change any logic, but I think I can get the desired behavior with the existing observer functions.

include/server/mir/graphics/display_configuration_observer.h
AlanGriffiths: I know this is a weird artefact of a transitional state in refactoring. But can we clarify the name slightly? I've not got a great name, but maybe `display_configuration_updated_for_this_session()`?

PR #991, opened 2019-09-11
Surface ids mirclient only

wmww:
My last major refactoring PR. Based on top of #989. Will rebase on master once that lands.


src/server/frontend_wayland/wayland_surface_observer.h
```diff
@@ -106,7 +106,6 @@ class WaylandSurfaceObserver
     /// Handle user input events
     ///@{
     void handle_input_event(MirInputEvent const* event);
-    void handle_keymap_event(MirKeymapEvent const* event);
```
AlanGriffiths: I'm not sure what this "dead code" has to do with "Surface ids mirclient only"?

Is the fact that this code is unused the cause of #997?

wmww: Moved to #998

wmww:
Not ready for this yet

Issue #997, opened 2019-09-14
miral support for changing keymap of running applications


Pajn:
Setting a new keymap with `miral::Keymap::set_keymap` only applies to new applications, existing applications still continues to use the old keymap. Surface contains a method `set_keymap` keymap which I assume can be used for changing the keymap of a running application, but it requires arguments I'm not sure are that easily accessible from outside mir. 

Would it make sense to add the method `miral::WindowInfo::set_keymap(miral::Keymap keymap)` or something like that for changing keymap of running applications?

AlanGriffiths:
my first thought is that failing to update the keymap of running applications is a a bug. @RAOF do you concur?

RAOF:
I forget whether or not we wanted to support different windows having different keymaps, in which case we would *not* want `miral::Keymap::set_keymap` to update the keymaps of all windows.

I *think* we do want to allow different windows to have different keymaps, so I think that `miral::WindowInfo::set_keymap()` is a sensible addition.

AlanGriffiths:
I think *the keymap* controlled by `miral::Keymap::set_keymap` is the session wide default.

I agree that windows do not have to respect that default, but is that a shell function?

RAOF:
How else can the shell tell the window which (non-default) keymap it should use?

AlanGriffiths:
After having a poke around, setting "the" keymap applies to `mir::common::input`, and the seat registers a `ConfigObserver`. `WlKeyboard` takes the current keymap from the seat, but subsequent changes are not notified.

AlanGriffiths:
> After having a poke around, setting "the" keymap applies to `mir::common::input`, and the seat registers a `ConfigObserver`. `WlKeyboard` takes the current keymap from the seat, but subsequent changes are not notified.

To clarify: `WlSeat::ConfigObserver` updates the current keymap on the seat, but does not notify existing `WlKeyboard`s.

It doesn't sound unreasonable to propagate the change to any apps. It isn't clear to me how apps should respond as there are sensible usecases for different apps to use different keymaps.

PR #998, opened 2019-09-16
Remove dead Wayland keymap code

wmww:
I believe this is not related to #997, but it is a possibility


wmww:
Still waiting for review

AlanGriffiths:
> Still waiting for review

As discussed elsewhere I think the problem is that this is not called, not that it should not exist.

wmww:
I think it should not exist because it's a redundant code path. The keymap *should* be changed via `WaylandSurfaceObserver::keymap_changed()` -> `WlKeyboard::set_keymap(Keymap const&)` which is left untouched.

wmww:
bors r=AlanGriffiths

AlanGriffiths:
Oh, this hung after the switch to travis.com? Merged

PR #1001, opened 2019-09-18
[miral] Make it possible for servers to set up the environment for launching clients

AlanGriffiths:
[miral] Make it possible for servers to set up the environment for launching clients

Add MirRunner properties to retrieve the WAYLAND_DISPLAY and DISPLAY supported by the server.

### DIFF
--- a/include/test/miral/test_display_server.h
+++ b/include/test/miral/test_display_server.h
@@ -94,6 +94,7 @@ struct TestDisplayServer : private TestRuntimeEnvironment
     build_window_manager_policy(WindowManagerTools const& tools) -> std::unique_ptr<WindowManagementPolicy>;
 
     /// Wrapper to gain access to the MirRunner
+    /// \note call after start_server()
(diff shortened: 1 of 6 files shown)


AlanGriffiths:
For example, these will need to be passed to userd when launching from a Snap confined environment.

include/test/miral/test_display_server.h
```diff
@@ -93,6 +93,9 @@ struct TestDisplayServer : private TestRuntimeEnvironment
     virtual auto
     build_window_manager_policy(WindowManagerTools const& tools) -> std::unique_ptr<WindowManagementPolicy>;
 
+    /// Wrapper to gain access to the MirRunner
```
RAOF: I note that `invoke_tools` says that this must be called after `start_server()`; presumably this does as well?

AlanGriffiths: Fixing

AlanGriffiths:
bors r=RAOF

Issue #990, opened 2019-09-09
Build fails on Alpine Linux


z3ntu:
Latest version (1.4.0) fails with 
```
[ 68%] Linking CXX shared module ../../../lib/miral_wlcs_integration.so
/usr/lib/gcc/x86_64-alpine-linux-musl/8.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: CMakeFiles/miral_wlcs_integration.dir/miral_integration.cpp.o: in function `TestWlcsDisplayServer':
/home/pmos/build/src/mir-1.4.0/tests/acceptance-tests/wayland/miral_integration.cpp:56: undefined reference to `miral::WaylandExtensions::supported[abi:cxx11]()'
collect2: error: ld returned 1 exit status
make[2]: *** [tests/acceptance-tests/wayland/CMakeFiles/miral_wlcs_integration.dir/build.make:109: lib/miral_wlcs_integration.so] Error 1
make[1]: *** [CMakeFiles/Makefile2:5646: tests/acceptance-tests/wayland/CMakeFiles/miral_wlcs_integration.dir/all] Error 2
```

AlanGriffiths:
I can only guess: maybe your linker doesn't understand src/miral/symbols.map:438?

z3ntu:
I'm using this cmake command (so the ld linker)
```
	cmake \
		-DCMAKE_INSTALL_PREFIX:PATH=/usr \
		-DCMAKE_INSTALL_LIBDIR=lib \
		-DMIR_USE_LD=ld \
		-DMIR_FATAL_COMPILE_WARNINGS=OFF
```
because there were some other issues with the default linker (ld.gold I think?)

RAOF:
So, this is probably a disagreement between how linkers interpret the symbols stanza. It breaks lld, too, which I use, so I'll fix it.

z3ntu:
Fails with both `GNU ld (GNU Binutils) 2.32` and `GNU gold (GNU Binutils 2.32) 1.16` on Alpine it seems

PR #1006, opened 2019-09-19
miral: Fix glob syntax in symbols.map

RAOF:
The glob syntax treats '[' as the start of a character range, so lld is
pretty reasonable in rejecting a single '[' in the symbol name.

'[[]' (a character range containing only '[') is the correct way to match
a '[' in glob syntax.

Closes: #990


AlanGriffiths:
@z3ntu is this the same issue you see?

z3ntu:
Yes, with that patch Mir 1.4.0 compiles on Alpine! Thanks @RAOF !

PR #1010, opened 2019-09-24
Don't try to draw cursor buffers that are not ready

### LINKED ISSUE #1009: Segfault when cursor moves after monitor plugged in
(There may be simpler, but this works for me.)

Run up `miral-app` with two screens L and R. Terminal appears on R.

Use mouse to move the terminal window to L, leaving mouse over titlebar.

Unplug L, wait for screen to update (terminal appears on R as it's now at 0,0).

Replug L, wait for screen to update (terminal appears on L as it's now at 0,0).

Move cursor off window.

Expect: system continues to work.
Actual: segfault at `src/platforms/mesa/server/kms/cursor.cpp:296`

AlanGriffiths:
Don't try to draw cursor buffers that are not ready. (Fixes #1009)

This results from the MultiMonitorArbiter state machine being in a condition that was not anticipated for streams passed to `set_cursor_stream()`. (I don't think it was possible with the mirclient API.)

Specifically, although the stream has buffers posted, they have been consumed and the `current_buffer` is no longer ready.

### DIFF
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -200,7 +200,8 @@ struct ms::CursorStreamImageAdapter
         }
 
         hotspot = new_hotspot;
-        post_cursor_image_from_current_buffer();
+        if (stream->buffers_ready_for_compositor(this))
+            post_cursor_image_from_current_buffer();
     }
 
 private:
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -200,8 +200,7 @@ struct ms::CursorStreamImageAdapter
         }
 
         hotspot = new_hotspot;
-        if (stream->buffers_ready_for_compositor(this))
-            post_cursor_image_from_current_buffer();
+        post_cursor_image_from_current_buffer();
     }
 
 private:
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
Oops!

Note to self: check you are running the tests from the right build.

RAOF:
Hey, wait.

> Specifically, although the stream has buffers posted, they have been consumed and the current_buffer is no longer ready.

How does it get into this state? Also, what was the actual crash? Even if there's not a *new* buffer available, the `MultiMonitorArbiter` is expected to *always* be able to return a valid buffer as long as at least one buffer has been submitted.

AlanGriffiths:
The linked bug shows how to get there.

This was the crash:
```plain
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __memmove_sse2_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:423
423     ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S: No such file or directory.
[Current thread is 1 (Thread 0x7f214bfff700 (LWP 9622))]
(gdb) bt
#0  0x00007f218f29d6c4 in __memmove_sse2_unaligned_erms ()
    at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:423
#1  0x00007f2186dfb0d8 in mir::graphics::mesa::Cursor::show(mir::graphics::CursorImage const&) (this=0x55fa5ec58010, cursor_image=...) at /home/alan/display_server/mir/src/platforms/mesa/server/kms/cursor.cpp:296
```
Tracing back, the stream `has_submitted_buffer()` is true, but `buffers_ready_for_compositor(this)` is 0.

RAOF:
I think `buffers_ready_for_compositor(this) == 0` is a red-herring here? That's used for triggering composition, not whether or not a valid buffer exists; `lock_compositor_buffer()` *should* always return a valid buffer (or throw, if it's called before at least one buffer is submitted). Something's not upholding invariants.

PR #1014, opened 2019-09-30
Revert "Merge #1010"

RAOF:
This reverts commit 391da119d1e0965f1b7c59b1a3cfbb64701332fa, reversing
changes made to d965d6af1afed7251ec3ef77b6a4d9aa52819dba.

The logic here is wrong - we want to draw the cursor even if this
particular cursor image has already been displayed on the current output
(ie: buffers_ready_for_compositor(this) == 0)

### DIFF
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -208,10 +208,7 @@ private:
     {
         if (stream->has_submitted_buffer())
         {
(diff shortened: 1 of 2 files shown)


RAOF:
This merge request is somewhat hampered by the fact that I can't actually reproduce the crash myself, so this would re-open #1009 

I'm pretty confident the change this reverts is wrong, though.

PR #1015, opened 2019-09-30
Fix crashes in wayland::OutputManager

RAOF:


AlanGriffiths:
This all looks sensible, but are you sharing how to reproduce the crashes that it claims to fix?

wmww:
I just wrote a WLCS test, stand by

AlanGriffiths:
bors r=wmww

PR #1016, opened 2019-09-30
PixelSource::read() should execute the supplied block

AlanGriffiths:
PixelSource::read() should execute the supplied block. (Fixes: #1009)

The WlShmBuffer::read() implementation failed to execute the do_with_pixels block when the underlying wl_buffer has been destroyed.

As the most recent buffer remains in the bufferstream until another buffer is submitted client code (e.g. CursorImageFromBuffer) cannot know that the PixelSource is no longer supported by an underlying wl_buffer.

I've removed the log that this is a logic_error as I don't see how the calling code can know.


wmww:
bors r=RAOF

PR #1018, opened 2019-10-01
Drop the last uses of deprecated alloc_buffer

RAOF:
Platforms which don't support mirclient won't support the
`alloc_buffer(BufferProperties const&)` method; switch the last two
remaining uses to `alloc_software_buffer()`.


AlanGriffiths:
> bors r+

That's premature: first it has to work!

wmww:
True. I missed the test results

PR #1020, opened 2019-10-01
tests: Add basic manual platform test harness

RAOF:
This is approximately the simplest possible thing, with plenty of scope for cleanup and augmentation, but it's useful enough!

### DIFF
--- a/tests/platform_test_harness/graphics_platform_test_harness.cpp
+++ b/tests/platform_test_harness/graphics_platform_test_harness.cpp
@@ -194,15 +194,15 @@ bool test_probe(mir::SharedLibrary const& dso, MinimalServerEnvironment& config)
     }
 }
 
-auto test_display_platform_construction(mir::SharedLibrary const& dso, MinimalServerEnvironment& env)
--> std::shared_ptr<mir::graphics::DisplayPlatform>
+auto test_platform_construction(mir::SharedLibrary const& dso, MinimalServerEnvironment& env)
+-> std::shared_ptr<mir::graphics::Platform>
 {
     try
     {
-        auto create_display_platform = dso.load_function<mg::CreateDisplayPlatform>(
-            "create_display_platform", MIR_SERVER_GRAPHICS_PLATFORM_VERSION);
+        auto create_host_platform = dso.load_function<mg::CreateHostPlatform>(
+            "create_host_platform", MIR_SERVER_GRAPHICS_PLATFORM_VERSION);
 
-        auto display = create_display_platform(
+        auto display = create_host_platform(
             env.options(),
             env.emergency_cleanup_registry(),
             env.console_services(),
@@ -216,26 +216,6 @@ auto test_display_platform_construction(mir::SharedLibrary const& dso, MinimalSe
     catch (...)
     {
         std::throw_with_nested(std::runtime_error{"Failure constructing platform"});
-    }
-}
-
-auto test_render_platform_construction(mir::SharedLibrary const& dso, MinimalServerEnvironment& env)
-    -> std::shared_ptr<mg::RenderingPlatform>
-{
-    try
-    {
-        auto create_render_platform = dso.load_function<mg::CreateRenderingPlatform>(
-            "create_rendering_platform", MIR_SERVER_GRAPHICS_PLATFORM_VERSION);
-
-        auto render = create_render_platform(env.options(), env.null_platform_authentication());
-
-        std::cout << "Successfully constructed RenderingPlatform" << std::endl;
-
-        return render;
-    }
-    catch (...)
-    {
-        std::throw_with_nested(std::runtime_error{"Failure constructing RenderingPlatform"});
     }
 }
 
@@ -634,9 +614,9 @@ int main(int argc, char const** argv)
         success &= test_probe(platform_dso, config);
         if (success)
         {
-            if (auto display_platform = test_display_platform_construction(platform_dso, config))
+            if (auto platform = test_platform_construction(platform_dso, config))
             {
-                if (auto display = test_display_construction(*display_platform, config))
+                if (auto display = test_display_construction(*platform, config))
                 {
                     success &= test_display_supports_gl(*display);
                     success &= dump_egl_config(*display);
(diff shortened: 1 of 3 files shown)


wmww:
What does this... um.. do? What's the point?

RAOF:
This is what I've been developing the vc4 (dispmanx) platform against. It's somewhat between a set of unit-tests and integration-tests for a platform - you can run `mir_platform_graphics_test_harness $YOUR_NEW_PLATFORM.so` on an incomplete platform and get some level of feedback as to *how* incomplete it is.

I think it's worth having in the Mir repository in case we want to bring up further platforms, and it could be cleaned up (mostly by giving better logging and such) and extended into something approximating a platform conformance test - something the OEM team can run on devices and go “this is expected to work” or “these are the parts that are *not* expected to work”.

AlanGriffiths:
Looks as though the code needs to be synced:
```plain
/usr/include/c++/9/ext/new_allocator.h:145:20: error: invalid new-expression of abstract class type â€˜{anonymous}::basic_software_buffer_drawing(mir::graphics::Display&, mir::graphics::GraphicBufferAllocator&, mir::renderer::RendererFactory&)::DummyRenderableâ€™
  145 |  noexcept(noexcept(::new((void *)__p)
      |                    ^~~~~~~~~~~~~~~~~~
  146 |        _Up(std::forward<_Args>(__args)...)))
      |        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/spread/mir/tests/platform_test_harness/graphics_platform_test_harness.cpp:501:11: note:   because the following virtual functions are pure within â€˜{anonymous}::basic_software_buffer_drawing(mir::graphics::Display&, mir::graphics::GraphicBufferAllocator&, mir::renderer::RendererFactory&)::DummyRenderableâ€™:
  501 |     class DummyRenderable : public mg::Renderable
      |           ^~~~~~~~~~~~~~~
In file included from /spread/mir/include/platform/mir/graphics/display_buffer.h:23,
                 from /spread/mir/tests/platform_test_harness/graphics_platform_test_harness.cpp:21:
/spread/mir/include/platform/mir/graphics/renderable.h:53:62: note: 	â€˜virtual std::experimental::fundamentals_v1::optional<mir::geometry::Rectangle> mir::graphics::Renderable::clip_area() constâ€™
   53 |     virtual std::experimental::optional<geometry::Rectangle> clip_area() const = 0;
      |                                                              ^~~~~~~~~
```

AlanGriffiths:
g++ still not happy.

tests/platform_test_harness/graphics_platform_test_harness.cpp
```diff
@@ -0,0 +1,676 @@
...
+    }
+}
+
+auto test_display_platform_construction(mir::SharedLibrary const& dso, MinimalServerEnvironment& env)
+-> std::shared_ptr<mir::graphics::DisplayPlatform>
+{
+    try
+    {
+        auto create_display_platform = dso.load_function<mg::CreateDisplayPlatform>(
+            "create_display_platform", MIR_SERVER_GRAPHICS_PLATFORM_VERSION);
```
AlanGriffiths: "create_display_platform" is unused by Mir. This should be "create_host_platform" as used in src/server/graphics/default_configuration.cpp.

tests/platform_test_harness/graphics_platform_test_harness.cpp
```diff
@@ -0,0 +1,676 @@
...
+            env.logger());
+
+        std::cout << "Successfully constructed DisplayPlatform" << std::endl;
+
+        return display;
+    }
+    catch (std::exception const& err)
+    {
+        std::cout << "Display construction failed: " << boost::diagnostic_information(err) << std::endl;
+        throw;
```
AlanGriffiths: Exceptions propagating out of this function are never caught. Although `return {};` would be handled as an error, we should probably also have a top-level catch handler.

tests/platform_test_harness/graphics_platform_test_harness.cpp
```diff
@@ -0,0 +1,676 @@
...
+    }
+}
+
+auto test_render_platform_construction(mir::SharedLibrary const& dso, MinimalServerEnvironment& env)
+    -> std::shared_ptr<mg::RenderingPlatform>
+{
+    try
+    {
+        auto create_render_platform = dso.load_function<mg::CreateRenderingPlatform>(
+            "create_rendering_platform", MIR_SERVER_GRAPHICS_PLATFORM_VERSION);
```
AlanGriffiths: "create_rendering_platform" is obsolete. It is only needed for `mgn::NestedBufferPlatform`.

Failing to find it should be at most an "information" message, not crash of the test harness because of an unhandled exception.

tests/platform_test_harness/graphics_platform_test_harness.cpp
```diff
@@ -0,0 +1,676 @@
...
+            {
+                success = false;
+            }
+        }
+        else
+        {
+            success = false;
+        }
+    }
+    return success ? 0 : -1;
```
AlanGriffiths: For the sake of clarity when running from the command-line let's `puts(success ? "SUCCESS", "FAILURE")`

AlanGriffiths:
@RAOF have you forgotten this PR?

tests/platform_test_harness/graphics_platform_test_harness.cpp
RAOF: Done

tests/platform_test_harness/graphics_platform_test_harness.cpp
RAOF: Now with top-level exception handler.

tests/platform_test_harness/graphics_platform_test_harness.cpp
RAOF: I'm not sure about this (and, likewise, the rendering platform one).

I was testing the Platform API we *want* to have, rather than the platform API we currently have, and in practice at the moment `create_display_platform` just returns an aggregate `DisplayPlatform` and `RenderingPlatform`…

RAOF:
> @RAOF have you forgotten this PR?

Oops, yes! (Ish)

tests/platform_test_harness/graphics_platform_test_harness.cpp
AlanGriffiths: I think you misread? I just wanted something shown on the console to indicate the result.

tests/platform_test_harness/graphics_platform_test_harness.cpp
AlanGriffiths: While I get the motivation for testing what it should be, that isn't an indicator of what is needed for Mir to work. Shouldn't the test reflect the current requirement and change as that requirement is refined?

tests/platform_test_harness/graphics_platform_test_harness.cpp
RAOF: Hah! I did indeed.

I forgot that not all users have terminals that visually differentiate between “last command returned success” and “last command returned failure”.

tests/platform_test_harness/graphics_platform_test_harness.cpp
RAOF: I've just plain removed this.

AlanGriffiths:
Unrelated failure...

bors r+

PR #1021, opened 2019-10-01
Drop some mirclient support infrastructure

AlanGriffiths:
Drop some mirclient support infrastructure.

We're not evolving or using mirclient and we aim to drop it soon. By dropping unit tests of the internals and debug APIs around it we clear the path.


tests/unit-tests/input/test_seat_input_device_tracker.cpp
```diff
@@ -1,298 +0,0 @@
-/*
```
RAOF: Why are these tests removed? The `SeatInputDeviceTracker` is an implementation detail of the `BasicSeat`, which *is* used in the Wayland frontend, and these tests don't *seem* to be using any of the removed libmirclient-debug functionality?

AlanGriffiths: No, they use `mir::input::receiver::XKBMapper` from the `mirclient-static` library. (I was removing dependencies on mirclient internals.)

On the face of it, this should be in `mircommon`, but it has (via `mir::input::KeyMapper` dependencies on "mir_toolkit/client_types.h" and "mir_toolkit/event.h")

Exporting another symbol from `mirclient` works for now.

PR #1022, opened 2019-10-01
renderer: Use only one of Texture or TextureSource

RAOF:
As it was, the Renderer would try and acquire *both* a Texture and
a TextureSource, and then preferentially use one in some parts of
the code and the other in other parts of the code.

Fortunately, for platforms where it matters we only expose one
or the other interface, not both, and for platforms where we
*do* expose both they're accidentally interchangable.

### DIFF
--- a/src/renderers/gl/renderer.cpp
+++ b/src/renderers/gl/renderer.cpp
@@ -449,16 +449,20 @@ void mrg::Renderer::draw(mg::Renderable const& renderable) const
 
     auto const texture = std::dynamic_pointer_cast<mg::gl::Texture>(renderable.buffer());
     auto const surface_tex =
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
> Why is a lambda created and immediately run? Could you not simply make `surface_tex` mutable and use normal control flow? IMO that would make more sense, unless there's something I'm missing.

Non-mutable is far easier to reason about, so I'm fine using a lambda. It does look like we're missing an abstraction though - it really looks like initialization should be something like:

```plain
    (tex, prog) = [](...){...}(...);
```
But we're messing with two subtly different `Texture` types and that ripples down the rest of the (rather too long)  function.

PR #1025, opened 2019-10-02
Fixes for new Protobuf on Arch

wmww:
We had a version check, but it appears it wasn't checking the right version. Also, some protobuf types have become `final`, so we can't inherit from them. `MirDisplayConfig` becomes a wrapper struct instead of a child struct, and `MirOutput` becomes an opaque struct that's typedef and casted to but never declared.

### DIFF
--- a/src/client/display_configuration.h
+++ b/src/client/display_configuration.h
@@ -27,10 +27,53 @@
 #include <vector>
 #include <memory>
 
-struct MirDisplayConfig : mir::protobuf::DisplayConfiguration
+struct MirDisplayConfig
 {
-    explicit MirDisplayConfig(mir::protobuf::DisplayConfiguration& pconfig) :
-        mir::protobuf::DisplayConfiguration::DisplayConfiguration{std::move(pconfig)} {}
+    MirDisplayConfig()
+    {
+    }
+
+    explicit MirDisplayConfig(mir::protobuf::DisplayConfiguration& wrapped) :
+        wrapped{std::move(wrapped)}
+    {
+    }
+
+    operator mir::protobuf::DisplayConfiguration const&() const
+    {
+        return wrapped;
+    }
+
+    int ByteSize() const
+    {
+        return wrapped.ByteSize();
+    }
+
+    uint8_t* SerializeWithCachedSizesToArray(uint8_t* target) const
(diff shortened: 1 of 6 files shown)


src/client/display_configuration.h
```diff
@@ -27,10 +27,33 @@
 #include <vector>
 #include <memory>
 
-struct MirDisplayConfig : mir::protobuf::DisplayConfiguration
+struct MirDisplayConfig
 {
-    explicit MirDisplayConfig(mir::protobuf::DisplayConfiguration& pconfig) :
-        mir::protobuf::DisplayConfiguration::DisplayConfiguration{std::move(pconfig)} {}
```
AlanGriffiths: \<rant>
While I accept that there are some good use for "final" classes, the default choice should be "don't" (one of the few defaults C++ has right). Forcing your users into designs like this isn't nice.
\</rant>

src/client/display_configuration.h
```diff
@@ -27,10 +27,33 @@
...
+
+    mir::protobuf::DisplayConfiguration* operator->()
+    {
+        return &wrapped;
+    }
+
+    mir::protobuf::DisplayConfiguration const* operator->() const
+    {
+        return &wrapped;
+    }
```
AlanGriffiths: These operators don't help anything: they are not transparent to code that used a `MirDisplayConfig*`.

I can see from the diff there are only a few call sites, and these only use a few functions. I'd prefer to provide forwarding methods and leave the call sites without the weird `(*config)->foo()` incantation this requires.
```c++
    int ByteSize() const { return wrapped.ByteSize(); }
    uint8_t* SerializeWithCachedSizesToArray(uint8_t* target) const { return wrapped.SerializeWithCachedSizesToArray(target); }
    bool ParseFromArray(const void* data, int size) { return wrapped.ParseFromArray(data, size); }
    int display_output_size() const { return wrapped.display_output_size(); }
    mir::protobuf::DisplayOutput const& display_output(int index) const { return wrapped.display_output(index); }
```

AlanGriffiths:
> bors r=wmww

Should be:

bors delegate=wmww

wmww:
bors r=AlanGriffiths

Issue #1013, opened 2019-09-26
XWayland fails on Eoan


wmww:
On Ubuntu Eoan, current master fails to start the XWayland server. When Mir and xeyes are run with the following commands:
```
$ ./bin/mir_demo_server --x11-display-experimental 2
```
```
$ DISPLAY=:2 xeyes
```
Mir produces this result:
```
[2019-09-26 15:20:23.254950] <information> xwaylandserver: Starting Xwayland
[2019-09-26 15:20:23.255782] <information> xwaylandserver: set ok
[2019-09-26 15:20:23.255824] <information> xwaylandserver: set ok
(EE) 
Fatal server error:
(EE) listener function for opcode 3 of zxdg_output_v1 is NULL
(EE) 
[2019-09-26 15:20:23.265811] <information> xwaylandserver: Stalled start of Xserver, trying to start again!
[2019-09-26 15:20:23.265865] <information> xwaylandserver: Starting Xwayland
[2019-09-26 15:20:23.266784] <information> xwaylandserver: set ok
[2019-09-26 15:20:23.266827] <information> xwaylandserver: set ok
(EE) 
Fatal server error:
(EE) listener function for opcode 3 of zxdg_output_v1 is NULL
(EE) 
```
And then repeats this several several more times. When run in GDB it crashes. This issue does not appear to effect Bionic. I suspect it is related to a new version of XWayland or other package that uses a newer Wayland protocol version.

RAOF:
This *looks* like an artefact of Mir sending an unrecognised event on the `zxdg_output_v1` interface; this is the same sort of crash we were seeing when we were sending v2 events on an interface bound as v1.

But that *shouldn't* be possible for `zxdg_output_v1`, because there's only one version?

RAOF:
[Hello!](https://github.com/MirServer/mir/blob/391da119d1e0965f1b7c59b1a3cfbb64701332fa/src/server/frontend_wayland/xdg_output_v1.cpp#L139)

RAOF:
https://github.com/MirServer/mir/blob/391da119d1e0965f1b7c59b1a3cfbb64701332fa/src/server/frontend_wayland/xdg_output_v1.cpp#L139

RAOF:
Alternatively, depending on how Wayland counts its opcodes, this *may* be an XWayland bug, where it assumes that we're exposing `zxdg_output_v1` v3 and so won't send the `zxdg_output_v1.done` event.

[...]

PR #1026, opened 2019-10-03
Bump xdg output to v3

wmww:
Note that this is not a breaking protocol change. I'm bumping the backwards compatible version from 2 to 3. The protocol itself is still `xdg_output_unstable_v1`. No events or requests were added. The only change is that now the `wl_output.done` event is used to commit a change instead of `xdg_output_v1.done`

### DIFF
--- a/src/server/frontend_wayland/xdg_output_v1.cpp
+++ b/src/server/frontend_wayland/xdg_output_v1.cpp
@@ -63,7 +63,8 @@ class XdgOutputV1 : public wayland::XdgOutputV1
 public:
     XdgOutputV1(
         wl_resource* new_resource,
-        graphics::DisplayConfigurationOutput const& config);
+        graphics::DisplayConfigurationOutput const& config,
+        wl_resource* wl_output_resource);
 
 private:
     void destroy();
@@ -79,7 +80,7 @@ auto mf::create_xdg_output_manager_v1(struct wl_display* display, OutputManager*
 }
 
 mf::XdgOutputManagerV1::XdgOutputManagerV1(struct wl_display* display, mf::OutputManager* const output_manager)
-    : Global(display, Version<2>()),
+    : Global(display, Version<3>()),
       output_manager{output_manager}
 {
 }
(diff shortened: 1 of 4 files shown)


wmww:
We don't have XDG output tests in WLCS yet. It's on my list of things to do, but that's preexisting.

wmww:
Fixes #1013

src/server/frontend_wayland/xdg_output_v1.cpp
```diff
@@ -128,18 +129,48 @@ void mf::XdgOutputManagerV1::Instance::get_xdg_output(wl_resource* new_output, w
...
+    /* xdg-output-unstable-v1.xml:
+     * For objects version 3 onwards, after all xdg_output properties have been
+     * sent (when the object is created and when properties are updated), a
+     * wl_output.done event is sent. This allows changes to the output
+     * properties to be seen as atomic, even if they happen via multiple events.
+     */
+    if (wl_resource_get_version(resource) >= 3)
+    {
+        // TODO: Use wrapper methods once wl_output is is uing a wrapper
+        if (wl_resource_get_version(wl_output_resource) >= WL_OUTPUT_DONE_SINCE_VERSION)
```
RAOF: This raises the interesting question of what happens when a client requests `xdg-output-unstable-v1` v3 but `wl_output` v1.

While that's *technically* allowed by the protocols-as-written, it's such clearly incorrect behaviour that I think we should unconditionally send `wl_output.done`.

RAOF: At most we should log, saying a client is doing something stupid, and *then* crash it by sending `wl_output.done`.

wmww: I added a warning, but I don't think we should disconnect or crash the client. The situation is weird, but IMO not a protocol error. The way I'm interpreting the protocol is "`wl_output` v1 is non-atomic, and `xdg_output_v1` v3+ uses `wl_output`'s commit system, so therefor it is also non-atomic if `wl_output` is v1". This reading is a bit dubious, but no more so that interpreting it as a protocol error.

wmww:
Tested by https://github.com/MirServer/wlcs/pull/144

PR #1028, opened 2019-10-04
Initial rpi-dispmanx platform

RAOF:


wmww:
Can I test this on Raspbian on a Pi 4?

RAOF:
Yes! As long as you have the userland parts installed on the pi and in the build environment.

wmww:
I'll try to get to this early next week

AlanGriffiths:
> This is unfortunately difficult to build and test; you'll need the [RPi userland](https://github.com/raspberrypi/userland) built and installed before you can build, and you _might_ need to replace Mesa's libEGL and libGLES with the ones from rpi-userland to build properly.

This complexity (of provisioning the build environment) adds to the case for properly supporting "platforms" out of tree. With the range of odd graphics stacks on arm devices this is something we'll have to address.

src/platforms/rpi-vc4/display.cpp
```diff
@@ -0,0 +1,357 @@
...
+    {
+        switch (format)
+        {
+        case DISPLAY_INPUT_FORMAT_RGB888:
+            return mir_pixel_format_rgb_888;
+        case DISPLAY_INPUT_FORMAT_RGB565:
+            return mir_pixel_format_rgb_565;
+        case DISPLAY_INPUT_FORMAT_INVALID:
+            return mir_pixel_format_rgb_888;
+//            BOOST_THROW_EXCEPTION((std::runtime_error{"Display has invalid pixel format?"}));
```
AlanGriffiths: This comment doesn't add anything.

src/platforms/rpi-vc4/display.cpp
```diff
@@ -0,0 +1,357 @@
...
+    }
+
+    return egl_config;
+}
+
+EGLContext create_context(EGLDisplay display, EGLConfig config)
+{
+    eglBindAPI(MIR_SERVER_EGL_OPENGL_API);
+
+    EGLint const context_attr[] = {
```
AlanGriffiths: I see this definition repeated in the next function, could it be moved to namespace scope and shared?

src/platforms/rpi-vc4/display.cpp
```diff
@@ -0,0 +1,357 @@
...
+{
+}
+
+auto mg::rpi::Display::create_hardware_cursor() -> std::shared_ptr<Cursor>
+{
+    // TODO: Should we implement this, or should we fix the server so that HW cursor is just
+    // an automatically-enabled optimisation when there's a HW plane that can contain it?
+
+    return {};
+//    BOOST_THROW_EXCEPTION((std::runtime_error{"HW cursor not implemented"}));
```
AlanGriffiths: Useless comment

wmww:
I built Mir on a RPI3. I haven't built or installed any userland stuff, so I suspect it didn't get built properly. When I enabled GL/KMS in raspi-config I was able to start Mir (presumably with the normal KMS platform). Apps ran and I could use the mouse, but the keyboard didn't work in Mir for some reason (apps didn't get input, Ctrl+Alt+Backspace didn't quit and I couldn't even TTY switch). Mir log [here](https://pastebin.com/hGNz8GMc). Again, since I haven't set up the userland stuff that may be completely irrelevant.

RAOF:
Ah, yes.

This does not automatically enable the rpi-vc4 platform in the presence of the necessary libs; you'll need to explicitly set the `MIR_PLATFORM=rpi-vc4` (possibly adding other platforms if you want).

wmww:
I did the fallowing on Raspbian:
```
sudo apt install libraspberrypi-dev # was already installed for me
PKG_CONFIG_PATH=/opt/vc/lib/pkgconfig cmake .. -DMIR_PLATFORM=rpi-dispmanx
make
```
This gave me these errors: https://pastebin.com/apmRiRzq

RAOF:
OK. This should be easier to build now.

`PKG_CONFIG_PATH=/usr/lib/arm-linux-gnueabihf/pkgconfig:/opt/vc/lib/pkgconfig cmake .. -DMIR_PLATFORM=rpi-dispmanx` *should* now build it for you; you need to explicitly add the system pkgconfig path to `PKG_CONFIG_PATH` so that it will pick up the existing mesa EGL headers for the things which don't use `brcm_host`, and the broadcom bits for things which do.

wmww:
Progress: https://pastebin.com/RXi3V9Kj

wmww:
Don't spend time debugging that just yet. A clean build seems to be working. Will update by EOD

wmww:
The clean build built, but doesn't start on a TTY:
```
Failed to initialize EGL: EGL_NOT_INITIALIZED (0x3001)
```
Or on X:
```
Failed to create EGL surface for display: EGL_BAD_NATIVE_WINDOW (0x300b)
```

RAOF:
Hm. I've re-tested this on Ubuntu core (in a classic chroot, upgraded to 18.04) with https://github.com/raspberrypi/userland commit `977d9dc`, and it works. I'm not sure what's happening for you (at least on the tty - it's not expected to work under X)

wmww:
I probably won't have access to a Raspberry Pi for the next week or two

wmww:
Ooops

AlanGriffiths:
I still see no harm in landing this.

bors r+

PR #1030, opened 2019-10-04
Add application_id surface property

wmww:
This will be something like `org.gnome.gedit`. Generally the same for all toplevels of a given application, but XDG shell allows setting it on a per-surface bases so so do we. It will be needed for foreign toplevel control, but also I though would be useful to MirAL.

### DIFF
--- a/debian/libmiral3.symbols
+++ b/debian/libmiral3.symbols
@@ -397,6 +397,8 @@ libmiral.so.3 libmiral3 #MINVER#
  (c++)"miral::WindowInfo::attached_edges(MirPlacementGravity)@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowInfo::exclusive_rect() const@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowInfo::exclusive_rect(mir::optional_value<mir::geometry::Rectangle> const&)@MIRAL_2.7" 2.7.0
+ (c++)"miral::WindowSpecification::application_id[abi:cxx11]() const@MIRAL_2.7" 2.7.0
+ (c++)"miral::WindowSpecification::application_id[abi:cxx11]()@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowSpecification::attached_edges() const@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowSpecification::attached_edges()@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowSpecification::exclusive_rect() const@MIRAL_2.7" 2.7.0
--- a/debian/libmiral3.symbols
+++ b/debian/libmiral3.symbols
@@ -397,13 +397,15 @@ libmiral.so.3 libmiral3 #MINVER#
  (c++)"miral::WindowInfo::attached_edges(MirPlacementGravity)@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowInfo::exclusive_rect() const@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowInfo::exclusive_rect(mir::optional_value<mir::geometry::Rectangle> const&)@MIRAL_2.7" 2.7.0
- (c++)"miral::WindowSpecification::application_id[abi:cxx11]() const@MIRAL_2.7" 2.7.0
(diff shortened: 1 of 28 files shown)


debian/libmiral3.symbols
```diff
@@ -397,6 +397,8 @@ libmiral.so.3 libmiral3 #MINVER#
  (c++)"miral::WindowInfo::attached_edges(MirPlacementGravity)@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowInfo::exclusive_rect() const@MIRAL_2.7" 2.7.0
  (c++)"miral::WindowInfo::exclusive_rect(mir::optional_value<mir::geometry::Rectangle> const&)@MIRAL_2.7" 2.7.0
+ (c++)"miral::WindowSpecification::application_id[abi:cxx11]() const@MIRAL_2.7" 2.7.0
+ (c++)"miral::WindowSpecification::application_id[abi:cxx11]()@MIRAL_2.7" 2.7.0
```
AlanGriffiths: These ought to be in the 2.8.0 namespace. (It looks as though they are correct in symbols.map, so what's going wrong?)

src/server/symbols.map
```diff
@@ -941,3 +941,10 @@ MIR_SERVER_0.1.5 {
     mir::frontend::MirClientSession::*
   };
 } MIR_SERVER_0.1.4;
+
+MIR_SERVER_0.1.6 {
```
AlanGriffiths: We haven't released 1.5 yet, no need for a 1.6 stanza.

include/server/mir/scene/surface.h
```diff
@@ -128,6 +128,12 @@ class Surface :
 
     virtual auto focus_state() const -> MirWindowFocusState = 0;
     virtual void set_focus_state(MirWindowFocusState focus_state) = 0;
+
+    /// Often the same as the session name, but on Wayland can be set on a per-window basis
+    /// See xdg_toplevel.set_app_id and http://standards.freedesktop.org/desktop-entry-spec/ for more details
+    /// Empty string if not set
+    virtual auto application_id() const -> std::string = 0;
+    virtual void set_application_id(std::string const& application_id) = 0;
```
AlanGriffiths: Is this comment intended to apply to both methods? That's not how doxygen behaves.

As far as "Empty string if not set" is concerned, it suggests a non-existent "not set" state. If we need to say anything, then "defaults to empty string" is probably clearer. (Or should there actually be a "not set" state?)

include/miral/miral/window_info.h
```diff
@@ -138,6 +138,13 @@ struct WindowInfo
     auto clip_area() const -> mir::optional_value<mir::geometry::Rectangle>;
     void clip_area(mir::optional_value<mir::geometry::Rectangle> const& area);
 
+    /// The D-bus service name and basename of the app's .desktop file
+    /// See http://standards.freedesktop.org/desktop-entry-spec/
```
AlanGriffiths: Maybe add
```c++
    /// \remark Since MirAL 2.8
```

debian/libmiral3.symbols
wmww: I had link time optimization off, so the script wasn't generating this file

include/server/mir/scene/surface.h
wmww: Fixed (see full diff if it doesn't show up in the mini view for some reason)

include/miral/miral/window_info.h
wmww: Done

src/server/scene/surface_creation_parameters.cpp
```diff
@@ -250,7 +252,11 @@ bool ms::operator==(
         lhs.confine_pointer == rhs.confine_pointer &&
         lhs.depth_layer == rhs.depth_layer &&
         lhs.attached_edges == rhs.attached_edges &&
-        lhs.exclusive_rect == rhs.exclusive_rect;
+        lhs.exclusive_rect == rhs.exclusive_rect &&
+        lhs.depth_layer == rhs.depth_layer &&
+        lhs.attached_edges == rhs.attached_edges &&
+        lhs.exclusive_rect == rhs.exclusive_rect &&
```
AlanGriffiths: Looks like duplication of depth_layer, attached_edges and exclusive_rect checks.

PR #1033, opened 2019-10-07
Convert buffer stream to cursor images in wl_pointer

### LINKED ISSUE #1000: When a client sets the cursor on two windows Mir shows wrong cursor
In some cases, Mir shows the previous cursor supplied by the client, instead of the current one. This seems to only happen when the are multiple windows (EDIT: multiple windows of the same app). The hotspot is always correct, only the image is wrong (when it should show the I cursor, it will show the arrow but too high).

wmww:
Fixes #1000

Unfortunately, the old paths are still used by mirclient so I can't get rid of them, but once we drop that `basic_surface` should no longer deal with stream cursors.

### DIFF
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -33,6 +33,7 @@
 
 #include <linux/input-event-codes.h>
 #include <boost/throw_exception.hpp>
+#include <string.h> // memcpy
 
 namespace mf = mir::frontend;
 namespace ms = mir::scene;
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -260,6 +260,7 @@ void mf::WlPointer::set_cursor(
         if (!compositor_stream)
             BOOST_THROW_EXCEPTION(std::logic_error("Surface does not have a compositor buffer stream"));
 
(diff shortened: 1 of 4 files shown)


AlanGriffiths:
Also, it doesn't fix #1000 - using the test scenario I still see the problem.

AlanGriffiths:
> This PR copies the image _currently_ contained by the pointer surface but ignores any subsequent changes to the surface content.

Oops. Misread the code.

AlanGriffiths:
Closing in favour of #1034

wmww:
This PR was falling for a similar pitfall as the code it replaces (failing to maintain single ownership of the frame posted callback). I was already planning a follow up PR that would have made this error obvious It will attach a surface role to the cursor surface, and throw if multiple roles are attached. Until then, I believe this to be correct, and a cleaner solution than #1034.

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -212,18 +281,53 @@ void mf::WlPointer::release()
...
+    }
+}
+
+void WlStreamCursor::apply_latest_buffer()
+{
+    if (auto const surface = surface_under_cursor.lock())
+    {
+        if (stream->has_submitted_buffer())
+        {
+            auto const cursor_image = std::make_shared<BufferCursorImage>(
```
AlanGriffiths: This is an additional memory allocation per frame than #1034, but this path is far from optimal anyway (and we can fix this if it ever shows up on a performance metric).

PR #1034, opened 2019-10-08
Remove false optimization that short-cut assigning a cursor surface for the second and subsequent times

AlanGriffiths:
Remove false optimization that short-cut assigning a cursor surface for the second and subsequent times. (Fixes: #1000)

(This didn't work if it has been assigned to another surface meanwhile.)

### DIFF
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -94,11 +94,16 @@ void mf::WlPointer::leave()
 {
     if (!surface_under_cursor)
         return;
-    surface_under_cursor.value()->remove_destroy_listener(this);
+    auto const surface = surface_under_cursor.value();
+    surface->remove_destroy_listener(this);
+    if (auto mir_surface = surface->scene_surface())
+    {
+        // TODO [purge-mirclient] We should clean up once we've got no mirclient to support:
+        // Called for side-effect set_cursor_image() implicitly detatches the cursor stream
+        mir_surface.value()->set_cursor_image({});
+    }
     auto const serial = wl_display_next_serial(display);
-    send_leave_event(
-        serial,
-        surface_under_cursor.value()->raw_resource());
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
The fundamental reason it didn't work was that `set_frame_posted_callback()` needs to be called as we cannot assume the frame callback is the one we set the first time.

wmww:
Nope, there's an issue with this. Will give a full explanation in a moment, but need to cancel the build.

bors r-

wmww:
The problem I was thinking of was that if a surface is closed while it doesn't have pointer focus, `~CursorStreamImageAdapter()` will reset another surface's frame posted callback. I believe this does happen, but as long as `set_cursor_stream()` is called along with every cursor buffer update it's not an issue. If a client *did* try to update a cursor's buffer without calling `wl_pointer.set_cursor`, **and** it had another window that recently closed, I suspect the cursor wouldn't update. To me this is an indication that `WlPointer` should own this logic, and then distribute cursor images as needed.

AlanGriffiths:
> The problem I was thinking of was that if a surface is closed while it doesn't have pointer focus, `~CursorStreamImageAdapter()` will reset another surface's frame posted callback. I believe this does happen, but as long as `set_cursor_stream()` is called along with every cursor buffer update it's not an issue. If a client _did_ try to update a cursor's buffer without calling `wl_pointer.set_cursor`, **and** it had another window that recently closed, I suspect the cursor wouldn't update.

I think you are right, and while this is pre-existing we should fix it while here.

Within the existing architecture there a conflict: for Wayland we would want to call `cursor_stream_adapter->reset()` when the pointer "leaves", but we don't want to do that for mirclient.

> To me this is an indication that `WlPointer` should own this logic, and then distribute cursor images as needed. There should only be one object at a time controlling a stream's frame posted callback.

I didn't move the logic because there was no need. Maybe the above issue provides a reason to duplicate the code.

AlanGriffiths:
Dropped in favour of #1033

Issue #999, opened 2019-09-16
Modifier state broken after TTY switch


wmww:
When Mir is running on a TTY and a user switches to another TTY with Ctrl+Alt+#, the Ctrl and Alt modifiers are sent to the client with current keyboard focus. When switching back to the Mir TTY, they are not removed (even after pressing and releasing the modifiers). This generally results in the appearance that input is broken. I have a recollection of this getting reported before, but I can't find it.

PR #1035, opened 2019-10-08
Resync keyboard state after server is paused

AlanGriffiths:
Resync keyboard state after server is paused. (Fixes: #999)


wmww:
It appears to fix the problem. Haven't looked at the code yet

PR #1036, opened 2019-10-08
More wl_pointer improvements

wmww:
Stacked on #1033

Makes `WlSurface` throw if a surface role is set without the previous one being cleared (not sure why it didn't already). This allows a surface to loose one role and gain a different one, which isn't *technically* allowed by Wayland, but it's an improvement.

 Makes the wl_pointer cursor set a null surface role, so the surface can't be given a different role or be used by a different wl_pointer (this would have exposed the bug in #1033).

### DIFF
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -299,14 +299,11 @@ void mf::WlPointer::release()
 WlSurfaceCursor::WlSurfaceCursor(mf::WlSurface* surface, geom::Displacement hotspot)
     : surface{surface},
       surface_destroyed{surface->destroyed_flag()},
-      stream{std::dynamic_pointer_cast<mc::BufferStream>(surface->stream)},
+      stream{surface->stream},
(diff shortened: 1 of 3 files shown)


src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -280,25 +296,33 @@ void mf::WlPointer::release()
...
 
-WlStreamCursor::WlStreamCursor(
-    std::shared_ptr<mc::BufferStream> const& stream,
-    geom::Displacement hotspot) :
-    stream{stream},
-    hotspot{hotspot}
+WlSurfaceCursor::WlSurfaceCursor(mf::WlSurface* surface, geom::Displacement hotspot)
+    : surface{surface},
+      surface_destroyed{surface->destroyed_flag()},
+      stream{std::dynamic_pointer_cast<mc::BufferStream>(surface->stream)},
```
AlanGriffiths: We can and should just change `WlSurface::stream` to `std::shared_ptr<mir::compositor::BufferStream>`.

PR #1041, opened 2019-10-09
Add session property to surfaces

wmww:
Adds and tests a new session property to surfaces. Allows modifying a surface without also storing it's session (will allow the implementation of foreign toplevel control). This could also allow us to reduce the number of places a surface is passed around with its session. Since a surface only ever belongs to a single session, it makes more sense to just pass around the surface and get the session when we need it.

### DIFF
--- a/include/server/mir/scene/null_surface_observer.h
+++ b/include/server/mir/scene/null_surface_observer.h
@@ -56,6 +56,8 @@ public:
     void start_drag_and_drop(Surface const* surf, std::vector<uint8_t> const& handle) override;
     void depth_layer_set_to(Surface const* surf, MirDepthLayer depth_layer) override;
     void application_id_set_to(Surface const* surf, std::string const& application_id) override;
+    void session_set_to(Surface const* surf, std::shared_ptr<Session> const& session) override;
+    void session_cleared(Surface const* surf) override;
 
 protected:
     NullSurfaceObserver(NullSurfaceObserver const&) = delete;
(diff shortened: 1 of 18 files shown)


wmww:
I do agree, but your claim

> This could be quite a bit simpler if the `Session` was an immutable property of the `Surface`.

is proving to be incorrect. Changing the `BasicSurface` constructor necessitates modifying a great many tests. Furthermore, passing a `shared_ptr<Session>` into `SurfaceFactory::create_surface()` means that `ApplicationSession` must use `std::enable_shared_from_this<>`.  Unfortunately that causes a number of tests to fail for unclear reasons (no, the session isn't trying to use `shared_from_this()` in its constructor. Yes, it is *actually* living in a shared pointer). I suppose I will eventually get to the bottom of this, but if you have any suggestions in the mean time I would be interested.

RAOF:
Hm.

Is there a particular reason that `Surface::session()` has to return a `std::shared_ptr<Session>`, rather than a raw pointer or reference? I'm not entirely clear on where this will be used.

Hm. No, because threading means that the `Session` could be invalidated between calling `Surface::session()` and actually using it.

Could `Surface::session()` return a `MutexGuard<Session>` (you'd need to move `src/platforms/mesa/server/kms/mutex.h` into a common include path)? Is that likely to result in deadlocks?

RAOF:
Hm, or something morally equivalent; you could return a `WrapperSession : public Session` that acts as a locking wrapper around the session.

wmww:
The goal is to be able to call `shell->modify_surface()` without a session. Currently, it takes shared pointers to a surface and a session. I haven't dug deep enough to determine *why* it needs a session, or if it could be modified to work with some sort of raw pointer to the session.

wmww:
Superseded by #1046

PR #1046, opened 2019-10-14
Add Immutable session property to surfaces

wmww:
Adds and tests an immutable session property to surfaces. Alternative to #1041

### DIFF
--- a/include/server/mir/scene/surface.h
+++ b/include/server/mir/scene/surface.h
@@ -44,6 +44,7 @@ struct StreamInfo
 };
 
 class SurfaceObserver;
+class Session;
 
 class Surface :
     public input::Surface,
@@ -136,6 +137,9 @@ public:
     virtual auto application_id() const -> std::string = 0;
     virtual void set_application_id(std::string const& application_id) = 0;
     ///@}
+
+    /// The session this surface was created by
+    virtual auto session() const -> std::weak_ptr<Session> = 0;
 };
 }
 }
--- a/src/server/scene/application_session.cpp
+++ b/src/server/scene/application_session.cpp
@@ -86,9 +86,7 @@ auto ms::ApplicationSession::create_surface(
     std::shared_ptr<ms::SurfaceObserver> const& observer) -> std::shared_ptr<Surface>
 {
     if (session && session.get() != this)
-    {
-        BOOST_THROW_EXCEPTION(std::logic_error("Incorrect session"));
-    }
+        fatal_error("Incorrect session");
 
     //TODO: we take either the content or the first stream's content for now.
     //      Once the surface factory interface takes more than one stream,
--- a/src/server/scene/application_session.cpp
+++ b/src/server/scene/application_session.cpp
@@ -81,9 +81,15 @@ ms::ApplicationSession::~ApplicationSession()
 }
 
 auto ms::ApplicationSession::create_surface(
+    std::shared_ptr<Session> const& session,
     SurfaceCreationParameters const& the_params,
     std::shared_ptr<ms::SurfaceObserver> const& observer) -> std::shared_ptr<Surface>
 {
+    if (session && session.get() != this)
+    {
+        BOOST_THROW_EXCEPTION(std::logic_error("Incorrect session"));
+    }
+
     //TODO: we take either the content or the first stream's content for now.
     //      Once the surface factory interface takes more than one stream,
     //      we can take all the streams as content.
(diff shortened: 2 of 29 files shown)


src/server/scene/application_session.cpp
```diff
@@ -81,9 +81,15 @@ ms::ApplicationSession::~ApplicationSession()
 }
 
 auto ms::ApplicationSession::create_surface(
+    std::shared_ptr<Session> const& session,
     SurfaceCreationParameters const& the_params,
     std::shared_ptr<ms::SurfaceObserver> const& observer) -> std::shared_ptr<Surface>
 {
+    if (session && session.get() != this)
+    {
+        BOOST_THROW_EXCEPTION(std::logic_error("Incorrect session"));
```
AlanGriffiths: I realize this is a precondition failure, but I don't think we should expect execution to continue. I would prefer `fatal_error()`

include/server/mir/scene/surface.h
```diff
@@ -136,6 +137,9 @@ class Surface :
     virtual auto application_id() const -> std::string = 0;
     virtual void set_application_id(std::string const& application_id) = 0;
     ///@}
+
+    /// The session this surface was created by
+    virtual auto session() const -> std::weak_ptr<Session> = 0;
```
AlanGriffiths: I'm concerned how users of this API can be proven to handle cases where the session is EOL but the surface owned by the session remains.

AFAICS that mapping of "surface" to "owning session" should only be needed in a small part of the mirserver implementation. Making it part of the mirserver API seems excessive.

src/server/scene/application_session.cpp
wmww: I disagree. If we have a bug that only triggers in specific conditions (perhaps it only happens on clients using a specific protocol), we should __not__ crash the whole server over that. Disconnecting one client (as I assume throwing an exception will) should be disruptive enough to get our attention.

include/server/mir/scene/surface.h
wmww: I think it's unclear how many places can benefit from this API because it doesn't exist yet. There are loads of places that store and pass around both a surface and a session (such as the entire `modify_window()` pipeline). In a future PR, I hope to switch them to only passing around a surface, and retrieving the session where necessary.

There's no single way to "prove" all users of the API correctly handle when the session dies before the surface, but individual cases can be tested.

wmww:
Closing and reopening to get travis to try again

include/server/mir/scene/surface.h
RAOF: The thought occurs that another way to more directly skin this cat would be to remove the `Session` parameter from the `modify_window()` pipeline rather than try to make the `Session` accessible where people want to call `modify_window()`.

From my *very brief* skim this doesn't seem like it'd be too much churn. Particularly, `AbstractShell::modify_surface()` only uses the `session` to pass through to `WindowManager::modify_surface()`, and our primary user of this (`miral::WindowManager`) ignores the `session` parameter already.

wmww: My plan *is* to remove session from `modify_surface()`, but I still think we should put this property here. For one thing, `AbstractShell::modify_surface()` calls `session->configure_streams()`. Furthermore, there are other places that currently hold both surfaces and sessions (such as `miral::Window::Self`). With this property those could also be simplified. Of could one could hold maps from surface -> session all over the place, but keeping the property on the surface makes more sense to me.

wmww:
Still waiting for further review

include/server/mir/scene/surface.h
AlanGriffiths: I don't think anyone disputes that change is necessary, but it isn't clear that *this* change is the way to go.

To take your example, I see nothing in `ApplicationSession::configure_streams()` to explain why it needs [to be implemented on] session.

wmww: Let's chat about this in our sync

src/server/scene/application_session.cpp
AlanGriffiths: AFAICS the only reason for passing `session` to `create_surface()` is that we can't create a `shared_ptr` from `this`?

I've spent too much time trying to figure out the original cause of problems in systems that "carried on regardless" (or relied on someone spotting a log message to have confidence that anything less than a core will be "enough to get our attention". We ought to be covering any invocations of this function in our testing: the very few call points for this function and they should all be under our control.

Your scenario that some client using a specific protocol finds a codepath we haven't covered in our tests is unlikely but, if it does happen, we'll likely never hear about it unless we do more than hope we disconnect the client.

wmww: Fair enough

include/server/mir/scene/surface.h
wmww: I believe the result of that sync was that while we may slowly move to architecture that doesn't need this property, adding it is the best step forward right now.

Issue #1049, opened 2019-10-15
Unnecessary modify_surface calls


wmww:
If you run `miral-shell --window-management-trace` and Gedit inside it you'll notice `handle_modify_window` gets called far too often. Somewhere between the Wayland frontend and the window manager it should be detect that our surface spec structs don't actually hold any modifications (or they are resetting values that are identical on the surface). This is a case where generating those structs would be helpful.

wmww:
Fixed by #1056

CatalinTermure:
Hello, I checked out the project recently and noticed that this issue is happening again. Far too many identical calls to `handle_modify_window` are issued for window creation and window movement.

It would be difficult for me to pinpoint the commit or a range where the issue started happening again since I could not build old commits easily. Let me know if you need any more help with reproduction.

Excerpt from opening kitty in a miral-shell session with `--window-management-trace`:
```
[2025-03-29 16:34:58.043880] <information> miral::Window Management: handle_modify_window window_info={name=kitty, type=freestyle, state=restored, top_left=154, 184, size=(989, 947), children={}, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0}, modifications={name=bash}
[2025-03-29 16:34:58.043902] <information> miral::Window Management: modify_window window_info={name=kitty, type=freestyle, state=restored, top_left=154, 184, size=(989, 947), children={}, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0}, modifications={name=bash}
[2025-03-29 16:34:58.043926] <information> miral::Window Management: ====
[2025-03-29 16:34:58.082632] <information> miral::Window Management: handle_modify_window window_info={name=, type=freestyle, state=fullscreen, top_left=0, 0, size=(949, 1031), restore_rect=(320, 256, (640, 512)), children={}, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0, output_id=1}, modifications={}
[2025-03-29 16:34:58.082657] <information> miral::Window Management: modify_window window_info={name=, type=freestyle, state=fullscreen, top_left=0, 0, size=(949, 1031), restore_rect=(320, 256, (640, 512)), children={}, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0, output_id=1}, modifications={}
[2025-03-29 16:34:58.082665] <information> miral::Window Management: ====
[...]

PR #1050, opened 2019-10-15
WlSurface: don't set pending values if nothing changed

wmww:
Addressed part of #1049, but not the min/max size stuff


src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -289,14 +292,53 @@ void mf::WlSurface::set_input_region(std::experimental::optional<wl_resource*> c
...
+            // The same input shape could be represented by the same rectangles in a different order, or event
+            // different rectangles. We don't check for that, however, because it would only cause an unnecessary
+            // update and not do any real harm. Checking for identical vectors should cover most cases.
+            shape_changed = false;
+            for (unsigned i = 0; i < shape.size(); i++)
+            {
+                if (input_shape.value()[i] != shape[i])
+                {
+                    shape_changed = true;
+                    break;
```
AlanGriffiths: Is the complexity of short-circuiting this loop really so much better than the simpler `shape_changed = input_shape.value() != shape`?

wmww: For some reason I had it in my mind that `std::vector` didn't implement `==`. Fixed.

wmww:
> Wouldn't it be simpler to leave the set_foo() methods unchanged and put all the equality checks in one place? I.e. WlSurface::commit()?

Yes

> And do we really need to optimize this anyway?

I would say so, since it was resulting in (in some cases) empty `modify_window` calls all the way through to the policy every frame.

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -392,6 +392,15 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
 
 void mf::WlSurface::commit()
 {
+    if (pending.offset && *pending.offset == offset_)
+        pending.offset = std::experimental::nullopt;
+
+    // The same input shape could be represented by the same rectangles in a different order, or event
```
AlanGriffiths: s/event/even/

PR #1055, opened 2019-10-17
AbstractShell: don't alert window manager whenever cursor changes

wmww:
Since the window manager doesn't care about cursor images changes, don't call `window_manager->modify_cursor()` every time it happens


wmww:
This isn't even used by current code. The cursor change stuff can probably be removed from the surface spec API along with mirclient. Not worth the trouble of testing

PR #1056, opened 2019-10-17
Don't call modify_window if size bounds have been set to the same value

wmww:
GTK loves to set max and min win size. Previously we were calling all the way through to the window management policy every time this was set (even if there was no change). This PR prevents that

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -49,14 +49,14 @@ geom::Size const max_possible_size{
 /// Clears pending if it holds a value different than cache
 /// sets cache to pending and leaves pending alone if it holds a different value
 template<typename T>
-inline void clear_pending_if_unchanged(mir::optional_value<T>* pending, T* cache)
+inline void clear_pending_if_unchanged(mir::optional_value<T>& pending, T& cache)
 {
-    if (*pending)
+    if (pending)
     {
-        if (pending->value() == *cache)
-            pending->consume();
+        if (pending.value() == cache)
+            pending.consume();
         else
-            *cache = pending->value();
+            cache = pending.value();
     }
 }
 }
@@ -360,10 +360,10 @@ void mf::WindowWlSurfaceRole::commit(WlSurfaceState const& state)
 
         if (pending_changes)
         {
-            clear_pending_if_unchanged(&pending_changes->min_width,  &committed_min_size.width);
-            clear_pending_if_unchanged(&pending_changes->min_height, &committed_min_size.height);
-            clear_pending_if_unchanged(&pending_changes->max_width,  &committed_max_size.width);
-            clear_pending_if_unchanged(&pending_changes->max_height, &committed_max_size.height);
+            clear_pending_if_unchanged(pending_changes->min_width,  committed_min_size.width);
+            clear_pending_if_unchanged(pending_changes->min_height, committed_min_size.height);
+            clear_pending_if_unchanged(pending_changes->max_width,  committed_max_size.width);
+            clear_pending_if_unchanged(pending_changes->max_height, committed_max_size.height);
         }
 
         if (pending_changes && !pending_changes->is_empty())
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -335,7 +335,7 @@ void mf::WindowWlSurfaceRole::commit(WlSurfaceState const& state)
             populate_spec_with_surface_data(spec());
         }
 
-        if (pending_changes)
+        if (pending_changes && !pending_changes->is_empty())
             shell->modify_surface(session, scene_surface, *pending_changes);
 
         pending_changes.reset();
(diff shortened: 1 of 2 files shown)


wmww:
Fixes #1049

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -40,6 +40,13 @@ namespace ms = mir::scene;
 namespace msh = mir::shell;
 namespace geom = mir::geometry;
 
+namespace
+{
+static geom::Size const max_possible_size{
```
RAOF: You need the `static` here; anonymous namespace implicitly specifies static duration and internal linkage.

That said, it's entirely harmless.

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -383,12 +394,47 @@ mir::shell::SurfaceSpecification& mf::WindowWlSurfaceRole::spec()
...
+        else                                \
+            committed = pending.value();    \
+    }
+
+            SET_IF_CHANGED(pending_changes->min_width,  committed_min_size.width);
+            SET_IF_CHANGED(pending_changes->min_height, committed_min_size.height);
+            SET_IF_CHANGED(pending_changes->max_width,  committed_max_size.width);
+            SET_IF_CHANGED(pending_changes->max_height, committed_max_size.height);
+
+#undef set_if_changed
```
RAOF: I'm pretty sure this won't do what you want, as `set_if_changed` ≠ `SET_IF_CHANGED` :grin:

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -383,12 +394,47 @@ mir::shell::SurfaceSpecification& mf::WindowWlSurfaceRole::spec()
...
+        if (pending_changes)
+        {
+
+#define SET_IF_CHANGED(pending, committed)  \
+    if (pending)                            \
+    {                                       \
+        if (pending.value() == committed)   \
+            pending.consume();              \
+        else                                \
+            committed = pending.value();    \
```
RAOF: Should this also be `committed = pending.consume()`?

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -383,12 +394,47 @@ mir::shell::SurfaceSpecification& mf::WindowWlSurfaceRole::spec()
     return *pending_changes;
 }
 
+void mf::WindowWlSurfaceRole::apply_pending_size_bounds()
+{
+    if (weak_scene_surface.lock())
```
RAOF: Is there any particular reason to do this check? You obviously don't *use* the `scene::Surface` here, so this must be here purely for optimisation (ie: we don't need to do the following if the surface has died)?

It's not entirely obvious to me that this will *be* an optimisation, as it introduces a bunch of atomic operations.

src/server/frontend_wayland/window_wl_surface_role.cpp
AlanGriffiths: C++ allows this to be written as an inline function template. No need for a macro. (I'm not sure that it can be done as a lambda for 16.04, but internal linkage at namespace scope would work for me.)

@RAOF I think you misread the intent: we don't reset on a change as we later rely on `pending_changes->is_empty()` to see if we applied changes. (The names don't help with following the code.)

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -335,7 +344,9 @@ void mf::WindowWlSurfaceRole::commit(WlSurfaceState const& state)
             populate_spec_with_surface_data(spec());
         }
 
-        if (pending_changes)
+        apply_pending_size_bounds();
```
AlanGriffiths: This function name is misleading. It does two significant things:

1. consuming idempotent size bound changes
2. remembering the new size bounds so as to implement 1

The name does not reflect _either_ of these but the first is most important. The casual reader will not spot the side effect of changing `pending_changes->is_empty()`.

Maybe it should be called `consume_idempotent_size_bound_changes()` as 2 is an implementation detail?

src/server/frontend_wayland/window_wl_surface_role.cpp
wmww: It's not an optimization. It's so we use `params` if the surface hasn't been created yet. It is common throughout this file to take alternative paths before and after the surface has been created.

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -40,6 +40,27 @@ namespace ms = mir::scene;
...
+namespace
+{
+geom::Size const max_possible_size{
+    std::numeric_limits<int>::max(),
+    std::numeric_limits<int>::max()};
+
+/// Clears pending if it holds a value different than cache
+/// sets cache to pending and leaves pending alone if it holds a different value
+template<typename T>
+inline void clear_pending_if_unchanged(mir::optional_value<T>* pending, T* cache)
```
AlanGriffiths: In C++ I'd use references, not pointers.

wmww: I wanted to make it obvious from the call site that the function modifies the parameters

AlanGriffiths: That's a C convention, not C++. You don't write, for example, `std::swap(&left, &right)`

wmww: Alright. I don't like it, but won't argue with language conventions

PR #1057, opened 2019-10-18
Rename surface->size() to surface->window_size()

wmww:
This makes things more explicit as we prepare for server side decorations

### DIFF
--- a/include/server/mir/scene/surface.h
+++ b/include/server/mir/scene/surface.h
@@ -61,7 +61,7 @@ public:
     /// Top-left corner (of the window frame if present)
     virtual geometry::Point top_left() const = 0;
     /// Size of the surface including window frame (if any)
-    virtual geometry::Size size() const = 0;
+    virtual geometry::Size window_size() const = 0;
 
     virtual graphics::RenderableList generate_renderables(compositor::CompositorID id) const = 0; 
     virtual int buffers_ready_for_compositor(void const* compositor_id) const = 0;
@@ -71,6 +71,7 @@ public:
     virtual void hide() = 0;
     virtual void show() = 0;
     virtual bool visible() const = 0;
+    /// Move the top-left of the frame to the given point
     virtual void move_to(geometry::Point const& top_left) = 0;
 
     /**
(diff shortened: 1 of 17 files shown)


wmww:
The existing confusion is *why* I want a name that isn't mixed up in it.

`frame_size()` makes it obvious that there is a different size that may be smaller but not larger. If I saw `window_size()`, I might assume it's the same size by a different name. `window_management_size()` and `window_frame_size()` are both usable, but a bit wordy for my taste.

I'd also like to point out that existing comments reference an unimplemented "window frame" that appears to be exactly what I'm building. That doesn't necessarily mean the name is ideal, but it may be relevant.

@RAOF?

RAOF:
> `frame_size()` makes it obvious that there is a different size that may be smaller but not larger.

Except this is incorrect - `xdg_surface.set_window_geometry()` will make the “window” size smaller than the surface size.

The fundamental question here is: what is the *purpose* of this size?

As far as I can tell from the usages (and its presence in `scene::Surface`), this is the (end-)user-visible size of the window. And is the *only* sensible size for `scene::Surface` (aka: `scene::Window`?). What would you want any other size out of `scene::Surface` - the renderer will need actual buffer size, but the renderer doesn't deal with `scene::Surface`.

I'm not sure that `scene::Surface::size()` *needs* disambiguating, but if it does, I'd paint the bikeshead `scene::Surface::logical_size()`.

RAOF:
Incidentally, `scene::Surface::size()` should probably be normalised to scaled coördinates: if the surface moves from a display of scale-factor 2 to a display of scale-factor 1 the window management shouldn't see the window double (or halve) in size.

wmww:
> the renderer will need actual buffer size, but the renderer doesn't deal with `scene::Surface`
What would you want any other size out of `scene::Surface` - the renderer will need actual buffer size, but the renderer doesn't deal with `scene::Surface`

Buffer size will not be exposed via `scene::Surface`. What is exposed is the client size. When there is a window frame, this is the size of the client area inside that. For that reason, I don't think `logical_size` is unambiguous, because the client size is also "logical" in the sense that it's not the "real" size of the buffer. I don't see how`frame_size` or `client_size` could be confused.

> `xdg_surface.set_window_geometry()` will make the “window” size smaller than the surface size.

Because I don't understand exactly what you mean by window size or surface size, I can't determine if this is correct or not. I can describe what's happening with less ambiguous terms as fallows:

`xdg_surface.set_window_geometry()` can set a surface's client size to something smaller than the size of its attached buffers. If server side decorations are not enabled, the frame size will be the same as the client size (also less than the size of the buffer). If SSD *is* enabled the frame size will be larger than the client size (and may be smaller or larger than the buffers). The frame size will *never* be smaller than the client size.

I have not yet incorporated window scaling into my mental model. What you say about that sounds correct.

wmww:
My take on my meeting with Alan: `window_size` might be preferable to `frame_size`, due to the multiple meanings of "frame" in various contexts. I think we're on the same page in terms of high level plan. Not going to make that change until I see what @RAOF has to say

wmww:
I think changing `resize` to `set_window_size` might be reasonable, but I don't see a pressing need to do it in this PR.

PR #1060, opened 2019-10-22
SurfaceObserver: refactor resized_to() to supply both a frame and client size

wmww:
Companion PR to #1057


RAOF:
This is, I think, superseded by #1061?

wmww:
Correct, I think we've pretty well decided this isn't the way we want to do it

PR #1061, opened 2019-10-23
SurfaceObserver: split resized_to() into window_resized_to() and content_resized_to()

wmww:
Alternative to #1060

### DIFF
--- a/src/server/symbols.map
+++ b/src/server/symbols.map
@@ -1,4 +1,4 @@
-MIR_SERVER_0.1.5 {
+MIR_SERVER_0.1.6 {
  global:
   extern "C++" {
     mir::check_for_termination_exception*;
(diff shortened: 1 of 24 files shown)


src/server/symbols.map
```diff
@@ -1,4 +1,4 @@
-MIR_SERVER_0.1.5 {
+MIR_SERVER_0.1.6 {
```
wmww: I'm not sure this is what I was supposed to do

AlanGriffiths: https://mir-server.io/doc/dso_versioning_guide.html

wmww: So... this is the correct way to change a symbol and break backwards compatibility?

AlanGriffiths: Part of it. The soname needs bumping too.

wmww: Oh, what does it mean to bump the soname? Doc just says when to do it.

AlanGriffiths: 1. bump MIRSERVER_ABI
2. tools/update_package_abis.sh

Maybe you can improve the docs?

wmww: When you say MIRSERVER_ABI, you mean change `MIR_SERVER_0.1.5` to `MIR_SERVER_0.1.6` in src/server/symbols.map, or is MIRSERVER_ABI a different thing?

AlanGriffiths: src/server/CMakeLists.txt:166

wmww: Ack, will get it passing the tests and then push version bump

RAOF: Hm. We would traditionally call this `MIR_SERVER_1.5.0`, as that's the version this would be released in.

AlanGriffiths: Let's be consistent with ourselves and use MIR_SERVER_1.6.0 as that's the version this would be released in.

wmww:
Like that?

PR #1063, opened 2019-10-24
BasicSurface: add ProofOfMutexLock

wmww:
It just so happens that the private version of `BasicSurface::visible()` only needs to be called from functions that hold a `std::lock_guard` and not a `std::unique_lock`. However, another private function will need to be called when both types of locks are held, so I'm introducing `BasicSurface::ProofOfMutexLock` which can be created from either. The system isn't full proof, but it does give a clear indication that functions that take it are intended to be called when the mutex is locked, and gives some level of compile-time checking.

### DIFF
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -168,7 +168,7 @@ void ms::SurfaceObservers::application_id_set_to(Surface const* surf, std::strin
 ms::BasicSurface::ProofOfMutexLock::ProofOfMutexLock(std::unique_lock<std::mutex> const& lock)
 {
     if (!lock.owns_lock())
-        BOOST_THROW_EXCEPTION(std::logic_error("ProofOfMutexLock created with unlocked unique_lock"));
+        fatal_error("ProofOfMutexLock created with unlocked unique_lock");
 }
 
 struct ms::CursorStreamImageAdapter
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -165,6 +165,12 @@ void ms::SurfaceObservers::application_id_set_to(Surface const* surf, std::strin
                  { observer->application_id_set_to(surf, application_id); });
 }
 
+ms::BasicSurface::ProofOfMutexLock::ProofOfMutexLock(std::unique_lock<std::mutex> const& lock)
+{
+    if (!lock.owns_lock())
+        BOOST_THROW_EXCEPTION(std::logic_error("ProofOfMutexLock created with unlocked unique_lock"));
+}
+
 struct ms::CursorStreamImageAdapter
 {
     CursorStreamImageAdapter(ms::BasicSurface &surface)
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -427,7 +427,7 @@ bool ms::BasicSurface::visible() const
     return visible(lock);
 }
 
-bool ms::BasicSurface::visible(std::lock_guard<std::mutex> const&) const
+bool ms::BasicSurface::visible(ProofOfMutexLock const&) const
 {
     bool visible{false};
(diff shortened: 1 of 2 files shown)


src/server/scene/basic_surface.h
```diff
@@ -156,7 +156,15 @@ class BasicSurface : public Surface
     void set_application_id(std::string const& application_id) override;
 
 private:
-    bool visible(std::lock_guard<std::mutex> const&) const;
+    struct ProofOfMutexLock
+    {
+        ProofOfMutexLock(std::lock_guard<std::mutex> const&) {}
+        ProofOfMutexLock(std::unique_lock<std::mutex> const&) {}
```
RAOF: Note that this doesn't actually prove the lock is held; 'std::unique_lock`s can be unlocked.

RAOF: The way to fix this would be something along the lines of `if (!lock) { lock.lock() }`

wmww: `if (!lock) { lock.lock(); }` would not be safe because on the rare occasion the current thread did not hold the lock but another thread did, it would assume the lock was held by the current thread and not obtain it. Perhaps we could throw in the `unique_lock` constructor if the lock is not held.

RAOF: The only way that could happen is if the `std::unique_lock` instance was shared between threads, which is (a) possible, but (b) very much the wrong thing to do, and is *also* possible with `std::lock_guard`.

That said, probably what you *actually* want is a `MutexGuard<State>`. That's a rather more expansive change, though.

wmww: Ah, I see what you mean. You *could* safely lock the lock, but I think trying to create a proof of a lock without already having a lock should be a `std::logic_error`.

src/server/scene/basic_surface.cpp
```diff
@@ -165,6 +165,12 @@ void ms::SurfaceObservers::application_id_set_to(Surface const* surf, std::strin
                  { observer->application_id_set_to(surf, application_id); });
 }
 
+ms::BasicSurface::ProofOfMutexLock::ProofOfMutexLock(std::unique_lock<std::mutex> const& lock)
+{
+    if (!lock.owns_lock())
+        BOOST_THROW_EXCEPTION(std::logic_error("ProofOfMutexLock created with unlocked unique_lock"));
```
AlanGriffiths: I think this ought to be a `fatal_error`, because what we are checking is the internal logic of `BasicSurface`.

wmww: I'm open to alternative ways of reporting this error, but not a fan of crashing production builds of Mir unless all hope is lost. If we hit this in production we may need to destroy a surface or disconnect a client, but IMO that shouldn't bring down the whole server along with unrelated windows.

AlanGriffiths: We're not checking for unexpected external behaviour. We're double checking our own internal logic (which ought to right). If it is wrong, then we need to fix it long before release.

An occasional lost window or client could be overlooked in testing, especially if intermittent. A nice core will not.

wmww: Ok. My feeling is that perhaps we should have a mechanism that errors catastrophically in dev/testing and less so in production (though that might be controversial and is *way* out-of-scope for this PR)

AlanGriffiths: > Ok. My feeling is that perhaps we should have a mechanism that errors catastrophically in dev/testing and less so in production (though that might be controversial and is _way_ out-of-scope for this PR)

A bit like `fatal_error`? That defaults to `fatal_error_abort` but can be changed to `fatal_error_except` when deployed to systems that want handle exceptions instead?

wmww: If it doesn't always bring down mir I'm not sure if `fatal_error` is correctly named. I'll ask you how that works next sync.

Issue #1064, opened 2019-10-25
Displacement/DeltaX/DeltaY should be replaced by Size/Width/Height


wmww:
I often end up blindly converting between deltas and sizes, and rarely if ever has this distinction caught a bug for me. I don't completely understand what the conceptual difference is. In the current implementation `Width - Width -> DeltaX`, but `Width - Width -> Width` would also make perfect sense (for example, `screen_width - panel_width -> maximized_window_width`). Also, I propose `Size + Size -> Size`, `Point + Size -> Point` and `Point - Size -> Point`. Basically, I think it makes sense to use sizes in every context we currently use displacements, and removing displacement will make code easier to write, easier to read, faster to compile, and actually *less* prone to bugs (because it's easier to spot them without `as_size()`, `as_displacement()` scattered everywhere).

AlanGriffiths:
There are issues with these types, but I don't think the answer is _clearly_ conflating "displacement" and "size", that may be a meaningful distinction.

Consider an alternative resolution: "size" _IsA_ "displacement" and make "size" explicitly constructable from "displacement". To cover your examples:

Expression|Meaning
--|--
`Width - Width`|`operator-(DeltaX, DeltaX) -> DeltaX`
`Size + Size -> Size`|`operator+(Displacement, Displacement) -> Displacement`
`Point + Size -> Point`|`operator+(Point, Displacement) -> Point`
`Point - Size -> Point`|`operator-(Point, Displacement) -> Point`

There should be no need for `as_displacement()` - that ought to be an implicit conversion.
We'd keep `as_size()`, or equivalently have an `explicit Size(Displacement)` constructor.

wmww:
Adding an implicit conversion from `Size` to `Displacement` may be a good idea, but what you propose has issues. For example it wouldn't allow `Size + Size + Size` or `Size - Size - Size` (since the first expression would return a `Displacement`, which can't be added or subtracted from `Size`). I think it would be enough to have implicit `Size -> Displacement` conversion __and__ add operators for `Size + Size -> Size` and `Size - Size -> Size`.

wmww:
Oh, true. I think `Width + Width -> Width`, `Width - Width -> Width` (and same for height) makes sense, but I didn't consider how that fails to transfer sensibly to `Size`.

AlanGriffiths:
Shrug

PR #1065, opened 2019-10-25
Add width and height operators

wmww:
More missing geometry operators. More needs to be done (I'm hoping to completely remove displacements, see #1064), but this helps.

### DIFF
--- a/include/core/mir/geometry/dimensions.h
+++ b/include/core/mir/geometry/dimensions.h
@@ -143,6 +143,12 @@ inline Height& operator+=(Height& lhs, DeltaY rhs) { return lhs = lhs + rhs; }
 inline Width& operator-=(Width& lhs, DeltaX rhs) { return lhs = lhs - rhs; }
 inline Height& operator-=(Height& lhs, DeltaY rhs) { return lhs = lhs - rhs; }
 
+// Adding Widths and Heights is fine
+inline constexpr Width operator+(Width lhs, Width rhs) { return Width(lhs.as_int() + rhs.as_int()); }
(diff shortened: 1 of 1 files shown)


include/core/mir/geometry/size.h
```diff
@@ -65,6 +65,11 @@ inline constexpr Size operator*(Size const& size, Scalar scale)
     return scale*size;
 }
 
+inline constexpr Size operator+(Size const& lhs, Size const& rhs)
+{
+    return Size{lhs.width + rhs.width, lhs.height + rhs.height};
```
AlanGriffiths: The other operators introduced here are arguable (while adding widths has a clear semantic, I'm not sure it is useful).

Howeverm _this_ one is wrong. (As discussed in #1064)

wmww: Removed

include/core/mir/geometry/displacement.h
```diff
@@ -91,6 +91,11 @@ inline constexpr Displacement operator-(Point const& lhs, Point const& rhs)
     return Displacement{lhs.x - rhs.x, lhs.y - rhs.y};
 }
 
+inline constexpr Displacement operator-(Size const& lhs, Size const& rhs)
+{
+    return Displacement{lhs.width - rhs.width, lhs.height - rhs.height};
+}
```
AlanGriffiths: Sorry, should have commented on this one too: it is wrong. (As discussed in #1064)

wmww: Good catch

AlanGriffiths:
The title and commit comment probably need to be tidied up before landing.

And @RAOF deserves a chance to say he doesn't like these operators. ;)

wmww:
bors r=RAOF

PR #1066, opened 2019-10-28
Drop workaround for broken std::uncaught_exception()

RAOF:
This was fixed in GCC/libstdc++ v5.3. Ubuntu 16.04 has v5.4 in -updates,
so none of our currently supported platforms suffer from this bug.

### DIFF
--- a/src/include/common/mir/unwind_helpers.h
+++ b/src/include/common/mir/unwind_helpers.h
@@ -30,25 +30,28 @@ class RevertIfUnwinding
 public:
     template<typename Apply>
     RevertIfUnwinding(Apply && apply, Unwind&& unwind)
-        : unwind{std::move(unwind)}
+        : unwind{std::move(unwind)},
+          initial_exception_count{std::uncaught_exceptions()}
     {
         apply();
     }
 
     RevertIfUnwinding(Unwind&& unwind)
-        : unwind{std::move(unwind)}
+        : unwind{std::move(unwind)},
+          initial_exception_count{std::uncaught_exceptions()}
     {
     }
 
     RevertIfUnwinding(RevertIfUnwinding<Unwind>&& rhs)
-        : unwind{std::move(rhs.unwind)}
+        : unwind{std::move(rhs.unwind)},
+          initial_exception_count{std::uncaught_exceptions()}
     {
         rhs.unwind = nullptr;
     }
 
     ~RevertIfUnwinding()
     {
-        if (std::uncaught_exceptions())
+        if (std::uncaught_exceptions() > initial_exception_count)
             unwind();
     }
 
@@ -57,6 +60,7 @@ private:
     RevertIfUnwinding& operator=(RevertIfUnwinding const&) = delete;
 
     Unwind unwind;
+    int const initial_exception_count;
 };
 
 template<typename Apply, typename Revert>
--- a/src/include/common/mir/unwind_helpers.h
+++ b/src/include/common/mir/unwind_helpers.h
@@ -40,9 +40,10 @@ public:
     {
     }
 
-    RevertIfUnwinding(RevertIfUnwinding<Unwind> && rhs)
+    RevertIfUnwinding(RevertIfUnwinding<Unwind>&& rhs)
         : unwind{std::move(rhs.unwind)}
     {
+        rhs.unwind = nullptr;
     }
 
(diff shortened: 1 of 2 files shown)


RAOF:
Seems like there's a consistent behavioural difference between Ubuntu and Fedora here. I'll take a look.

RAOF:
I still don't know why there is a difference in behaviour between Fedora and Ubuntu, but I now believe that the Fedora behaviour is correct and I don't know why Ubuntu passes!

The analysis is: `on_unwind()` is broken. Or, rather, `on_unwind()` triggers if *any* unwinding is happening, *not* whether code in `start()` or `stop()` threw an exception.

Furthermore, the code in `on_unwind()` does not actually restore state; notably, it doesn't remove an added `observer` or re-add a removed `observer`.

This gets triggered by exception handling in `MediatingDisplayChanger::apply_config()` - we successfully `stop()`, then `display->configure()` throws an exception, and the scoped `compositor->start()` gets invoked while unwinding is happening, resulting in the compositor being successfully started and *then* having the unwind cleanup destroy the compositor threads and reset the state to `stopped` but, importantly, *not* removing the `observer`.

Then, when Mir is tearing down, `~MediatingDisplayChanger` calls `stop()`, which finds that the compositor is *already* in the stopped state and so does nothing, leaving the `observer` registered. Subsequently, the destructor of the software cursor triggers a scene recomposition, which notifies `MediatingDisplayChanger`'s not-unregistered observer, which results in sadness.

I'm not entirely sure what to do here. Abandoning this PR will leave the (accidentally) working status-quo, at least until we switch to C++17 and `on_unwind` can use `std::uncaught_exceptions()` to *correctly* implement it's desired behaviour.

But looking at it, I'm not entirely sure that trying to recover from exceptions in `start()` and `stop()` is a good idea at all (particularly since we've *never* done so correctly ;)).

AlanGriffiths:
> I'm not entirely sure what to do here. Abandoning this PR will leave the (accidentally) working status-quo, at least until we switch to C++17 and `on_unwind` can use `std::uncaught_exceptions()` to _correctly_ implement it's desired behaviour.

We have switched to C++17 and `on_unwind` uses `std::uncaught_exceptions()`. But I'm not sure about the issue described with `MediatingDisplayChanger`. Is it not respecting the strong guarantee? (That sounds like a simple bug, but I've not followed closely.)

I'm going to rebase this PR and see what happens now.

AlanGriffiths:
@RAOF I assume the above analysis is no longer a problem? I think it was fixed by 1a1ed1359f7297a60346bd6f01d53db3e16f035e, but can you confirm?

bors delegate=@RAOF

RAOF (commented):
I've lost the context that generated my previous analysis, but I'm pretty sure that `MultiThreadedCompositor` isn't providing the strong guarantee, and we're not testing correctly. This is no worse than the current state, though, so +1.

I've fixed up a couple of things in `RevertIfUnwinding` I noticed, so passing the review baton back to you.

bors delegate=@AlanGriffiths

AlanGriffiths:
> I've fixed up a couple of things in `RevertIfUnwinding` I noticed, so passing the review baton back to you.

You really think it worth adding complexity to deal with starting the compositor during unwinding? Surely this would be enough:

```C++
    RevertIfUnwinding(Apply && apply, Unwind&& unwind)
        : unwind{std::move(unwind)}
    {
        if (std::uncaught_exceptions())
            fatal_error("This is crazy! Using unwind helpers during unwinding");
        ...
    }
```

AlanGriffiths:
Oh well, we don't need to polish further.

bors r+

RAOF:
> > I've fixed up a couple of things in `RevertIfUnwinding` I noticed, so passing the review baton back to you.
> 
> You really think it worth adding complexity to deal with starting the compositor during unwinding? Surely this would be enough:
> 
> ```c++
>     RevertIfUnwinding(Apply && apply, Unwind&& unwind)
>         : unwind{std::move(unwind)}
>     {
>         if (std::uncaught_exceptions())
>             fatal_error("This is crazy! Using unwind helpers during unwinding");
>         ...
>     }
> ```

We can't do that, as we *do* use the unwind helpers during unwinding. Specifically: `MediatingDisplayChanger` uses the unwind helpers to call `compositor->start()` and `compositor->stop()`, which itself uses the unwind helpers.

PR #1068, opened 2019-11-01
Wayland platform

AlanGriffiths:
Assuming there's an existing Wayland server at `wayland-0`, enable connecting to it with:

`--wayland-host wayland-0`

This works well enough as a first cut. The main issue I know of is that the modifier key state gets out of sync when switching between this window and others on a desktop. It's still usable though as toggling the key resynchronises.

Still no tests, but there's not a lot to meaningfully test for platforms beyond "Mir lives". Will look at that after the sprint.

### DIFF
--- a/src/platforms/wayland/cursor.cpp
+++ b/src/platforms/wayland/cursor.cpp
@@ -1,0 +1,49 @@
+/*
+ * Copyright © 2019 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Alan Griffiths <alan@octopull.co.uk>
+ */
+
+#include "cursor.h"
+
+namespace mpw = mir::platform::wayland;
+
+mpw::Cursor::Cursor(wl_display* display, std::shared_ptr<graphics::CursorImage> const& default_image) :
+    display(display),
+    default_image(default_image)
+{
+    puts(__PRETTY_FUNCTION__);
+}
+
+mpw::Cursor::~Cursor() = default;
+
+void mpw::Cursor::move_to(geometry::Point)
+{
+}
+
+void mpw::Cursor::show(graphics::CursorImage const& /*cursor_image*/)
+{
+    puts(__PRETTY_FUNCTION__);
+}
+
+void mpw::Cursor::show()
+{
+    puts(__PRETTY_FUNCTION__);
+}
+
+void mpw::Cursor::hide()
+{
+    puts(__PRETTY_FUNCTION__);
+}
--- a/src/platforms/wayland/cursor.cpp
+++ b/src/platforms/wayland/cursor.cpp
@@ -31,6 +31,7 @@ mpw::Cursor::~Cursor() = default;
 
 void mpw::Cursor::move_to(geometry::Point)
 {
+    puts(__PRETTY_FUNCTION__);
 }
 
 void mpw::Cursor::show(graphics::CursorImage const& /*cursor_image*/)
--- a/src/platforms/wayland/cursor.cpp
+++ b/src/platforms/wayland/cursor.cpp
@@ -17,36 +17,133 @@
  */
 
 #include "cursor.h"
+#include "displayclient.h"
+
+#include <mir/fd.h>
 
 #include <wayland-client.h>
 
+#include <boost/throw_exception.hpp>
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/mman.h>
+
+#include <cstring>
+#include <system_error>
+
 namespace mpw = mir::platform::wayland;
 
-mpw::Cursor::Cursor(wl_display* display, std::shared_ptr<graphics::CursorImage> const& default_image) :
-    display(display),
-    default_image(default_image)
+namespace
 {
-//    puts(__PRETTY_FUNCTION__);
+struct wl_shm_pool* make_shm_pool(struct wl_shm* shm, int size, void **data)
+{
+    static auto (*open_shm_file)() -> mir::Fd = []
+        {
+            open_shm_file = []
+                {
+                    // As we're a Wayland client, create the shm file like Wayland clients.
+                    // While using O_TMPFILE would be more elegant, this works with Snap-confined servers.
+                    static auto const template_filename =
+                        std::string{getenv("XDG_RUNTIME_DIR")} + "/wayland-cursor-shared-XXXXXX";
+
+                    auto const filename = strdup(template_filename.c_str());
+                    mir::Fd fd{mkostemp(filename, O_CLOEXEC)};
+                    if (fd != -1)
+                    {
+                        if (unlink(filename) < 0)
+                        {
+                            free(filename);
+                            return mir::Fd{};
+                        }
+                    }
+                    free(filename);
+                    return fd;
+                };
+
+            return open_shm_file();
+        };
+
+    auto fd = open_shm_file();
+
+    if (fd < 0) {
+        BOOST_THROW_EXCEPTION((std::system_error{errno, std::system_category(), "Failed to open shm buffer"}));
+    }
+
+    if (auto error = posix_fallocate(fd, 0, size))
+    {
+        BOOST_THROW_EXCEPTION((std::system_error{error, std::system_category(), "Failed to allocate shm buffer"}));
+    }
+
+    if ((*data = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0)) == MAP_FAILED)
+    {
+        BOOST_THROW_EXCEPTION((std::system_error{errno, std::system_category(), "Failed to mmap buffer"}));
+    }
+
+    return wl_shm_create_pool(shm, fd, size);
+}
 }
 
-mpw::Cursor::~Cursor() = default;
+mpw::Cursor::Cursor(wl_display* display, wl_compositor* compositor, wl_shm* shm, std::shared_ptr<graphics::CursorImage> const& default_image) :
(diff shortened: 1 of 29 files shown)


wmww:
Why have a `--wayland-host` option instead of picking up the `WAYLAND_DISPLAY` environment variable and defaulting to `wayland-0` if it's unset?

AlanGriffiths:
Mir also has to behaves as any other Wayland compositor and provide the endpoint specified by `WAYLAND_DISPLAY` if it is set.

Behaving as a client isn't typical, so a different way of configuring and enabling this is called for. It is also wrong to default to using `wayland-0`: consider switching from a Wayland based desktop to another VT and trying to run Mir...

src/platforms/wayland/displayclient.cpp
```diff
@@ -0,0 +1,856 @@
...
+mgw::DisplayClient::Output::Output(
+    wl_output* output,
+    DisplayClient* owner,
+    std::function<void(Output const&)> on_constructed,
+    std::function<void(Output const&)> on_change) :
+    output{output},
+    owner{owner},
+    surface{wl_compositor_create_surface(owner->compositor)},
+    on_done{[this, on_constructed = std::move(on_constructed), on_change=std::move(on_change)]
+        (Output const& o) mutable { on_constructed(o), on_done = std::move(on_change); }}
```
RAOF: What happens here when `on_done()` is called the third time?

src/platforms/wayland/displayclient.cpp
```diff
@@ -0,0 +1,856 @@
...
+    return this;
+}
+
+void mgw::DisplayClient::Output::make_current()
+{
+    if (eglctx == EGL_NO_CONTEXT)
+        eglctx = eglCreateContext(owner->egldisplay, owner->eglconfig, owner->eglctx, ctxattribs);
+
+    if (!eglMakeCurrent(owner->egldisplay, eglsurface, eglsurface, eglctx))
+        throw std::runtime_error("Can't eglMakeCurrent");
```
RAOF: Could you kindly use `mg::egl_error` here? The EGL error can sometimes be illuminating.

You probably also mean `BOOST_THROW_EXCEPTION`, right?

src/platforms/wayland/displayclient.cpp
```diff
@@ -0,0 +1,856 @@
...
+
+void mgw::DisplayClient::Output::release_current()
+{
+    eglMakeCurrent(owner->egldisplay, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
+}
+
+void mgw::DisplayClient::Output::swap_buffers()
+{
+    if (eglSwapBuffers(owner->egldisplay, eglsurface) != EGL_TRUE)
+        fatal_error("Failed to perform buffer swap");
```
RAOF: There doesn't seem to be a particular reason not to be throwing an `mg::egl_error` here, either?

src/platforms/wayland/displayclient.cpp
```diff
@@ -0,0 +1,856 @@
...
+            EGL_STENCIL_SIZE, gl_config->stencil_buffer_bits(),
+            EGL_RENDERABLE_TYPE, MIR_SERVER_EGL_OPENGL_BIT,
+            EGL_NONE
+        };
+
+    eglBindAPI(MIR_SERVER_EGL_OPENGL_API);
+
+    egldisplay = eglGetDisplay((EGLNativeDisplayType)(display));
+    if (egldisplay == EGL_NO_DISPLAY)
+        throw std::runtime_error("Can't eglGetDisplay");
```
RAOF: s/`std::runtime_error`/`mg::egl_error`/g

src/platforms/wayland/displayclient.cpp
```diff
@@ -0,0 +1,856 @@
...
+}
+
+void mgw::DisplayClient::touch_orientation(
+    wl_touch* /*touch*/,
+    int32_t /*id*/,
+    wl_fixed_t /*orientation*/)
+{
+}
+
+void mgw::DisplayClient::seat_capabilities(wl_seat* seat, uint32_t capabilities)
```
RAOF: This fails to handle being called multiple times; it will result in events being multiplied if it is called multiple times.

I don't know if that should be blocking (seat capabilities changing should be rare), but it should definitely have at least a TODO.

You should ideally be saving the `wl_pointer` etc across calls, and releasing them if the new capabilities do not contain the relevant capability.

src/platforms/wayland/displayclient.h
```diff
@@ -0,0 +1,176 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EGMDE_EGFULLSCREENCLIENT_H
```
RAOF: *Might* want to rename the include guards? :)

src/server/input/input_probe.cpp
```diff
@@ -74,7 +74,9 @@ mir::UniqueModulePtr<mi::Platform> mi::probe_input_platforms(
                 auto const probe = module->load_function<mi::ProbePlatform>(
                     "probe_input_platform", MIR_SERVER_INPUT_PLATFORM_VERSION);
 
-                if (probe(options, *console) > reject_platform_priority)
+                auto const priority = probe(options, *console);
+                printf("************ priority=%d, reject_platform_priority=%d\n", (int)priority, (int)reject_platform_priority);
```
RAOF: You've left some debugging in here.

(It might be nice to clean it up and leave it, though)

src/platforms/wayland/displayclient.cpp
RAOF: (And in a bunch more places here, and `BOOST_THROW_EXCEPTION`)

src/server/input/input_probe.cpp
AlanGriffiths: Ack. I think the right way is to explicitly kill Mir in the forgotten exception handler.

src/platforms/wayland/displayclient.cpp
AlanGriffiths: Adding TODO comment

src/platforms/wayland/displayclient.cpp
AlanGriffiths: > What happens here when `on_done()` is called the third time?

Unless I'm missing something, the supplied `on_change` handler is called for the second time.

src/platforms/wayland/display.cpp
```diff
@@ -0,0 +1,540 @@
...
+            {shutdown_signal, POLLIN, 0},
+        };
+
+    while (!(fds[shutdown].revents & (POLLIN | POLLERR)))
+    {
+        while (wl_display_prepare_read(display) != 0)
+        {
+            if (wl_display_dispatch_pending(display) == -1)
+            {
+                BOOST_THROW_EXCEPTION((std::system_error{errno, std::system_category(), "Failed to dispatch Wayland events"}));
```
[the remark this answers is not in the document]
AlanGriffiths: Added the forgotten handler and killed Mir there properly.

src/platforms/wayland/displayclient.cpp
RAOF: ...oh, yes!

Is it safe to overwrite the `std::function` from its `operator()`?

AlanGriffiths: Same rules as `delete this`: Do not access its state afterwards.

src/platforms/wayland/display.h
```diff
@@ -0,0 +1,155 @@
...
+    Display(
+        wl_display* const wl_display,
+        std::shared_ptr<GLConfig> const& gl_config,
+        std::shared_ptr<DisplayReport> const& report);
+
+    ~Display();
+
+    static void set_keyboard_sink(std::shared_ptr<input::wayland::KeyboardInput> const& keyboard_sink);
+    static void set_pointer_sink(std::shared_ptr<input::wayland::PointerInput> const& pointer_sink);
+    static void set_touch_sink(std::shared_ptr<input::wayland::TouchInput> const& touch_sink);
```
wmww: These should also be optional or documented as being nullable.

src/platforms/wayland/displayclient.cpp
```diff
@@ -0,0 +1,860 @@
...
+    auto const output = self->bound_outputs.find(id);
+    if (output != self->bound_outputs.end())
+    {
+        self->on_output_gone(output->second.get());
+        self->bound_outputs.erase(output);
+    }
+    // TODO: We should probably also delete any other globals we've bound to that disappear.
+}
+
+void mgw::DisplayClient::keyboard_keymap(wl_keyboard* /*keyboard*/, uint32_t /*format*/, int32_t fd, uint32_t size)
```
wmww: Perhaps we should check to make sure `format` is `WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1` (if that is what we are assuming)

src/platforms/wayland/cursor.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+namespace mpw = mir::platform::wayland;
+
+namespace
+{
+struct wl_shm_pool* make_shm_pool(struct wl_shm* shm, int size, void **data)
+{
+    static auto (*open_shm_file)() -> mir::Fd = []
+        {
+            open_shm_file = []
```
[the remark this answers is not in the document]
AlanGriffiths: No, cut & paste problem. (The code I copied set `open_shm_file` to one of several strategies the first time it ran, but that's not needed.)

src/platforms/wayland/cursor.cpp
```diff
@@ -0,0 +1,149 @@
...
+{
+    if (default_image) show(*default_image);
+}
+
+mpw::Cursor::~Cursor()
+{
+    wl_surface_destroy(surface);
+
+    std::lock_guard<decltype(mutex)> lock{mutex};
+    if (buffer) wl_buffer_destroy(buffer);
```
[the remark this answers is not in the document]
AlanGriffiths: Wayland *clients* are not required to make all API calls on single thread.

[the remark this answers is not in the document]
RAOF: Correct. There are *some* things which are not threadsafe (`add_listener`, for example, races with dispatch), but the core “send messages on the object” bits *are* threadsafe.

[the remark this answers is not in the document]
wmww: Good to know

PR #1069, opened 2019-11-08
Fix build against GMock/GTest ≥ 1.9

RAOF:
There's a 1.9 snapshot in focal-proposed, and Mir fails to build against it.

This fixes the build.


RAOF:
I think this might be missing some necessary changes

RAOF:
False alarm. *This* doesn't miss any of the necessary changes, the package I *built from* this was missing (all the) changes :speak_no_evil:

Saviq:
@RAOF can you rebase on/merge #1070 to see the improvement in CI?

RAOF:
This got rebased on #1070, so shouldn't land before that.

Let's get a travis run in before then!
bors try

RAOF:
Oops. I didn't actually base on #1070. Anyway, since that's been merged it's rather a moot point.

bors r=AlanGriffiths

PR #1070, opened 2019-11-08
[travis,spread] refresh releases and add devel-proposed testing

Saviq:


RAOF:
Hm.
`Mir fatal error: Could not find surface to remove in miral::BasicWindowManager::remove_surface()`
That's odd.

Seems to be not entirely consistent, and enabling 19.10 is the right thing to do *anyway*. Let's give this another try...

bors r+

RAOF:
Different failure, on Fedora 30 this time.

bors r+

PR #1071, opened 2019-11-11
Drop the --wayland-socket-name configuration option

AlanGriffiths:
Drop the --wayland-socket-name configuration option: It duplicates what WAYLAND_DISPLAY means to a server, so use that instead.


AlanGriffiths:
> LGTM bors r+

You need bors on its own line:

bors r=wmww

Issue #1073, opened 2019-11-14
Broken mir-platform-graphics-wayland packaging


AlanGriffiths:
```plain
dh_install: mir-platform-graphics-wayland16 missing files: usr/lib/*/mir/server-platform/graphics-wayland.so.16
dh_install: missing files, aborting
```
https://launchpadlibrarian.net/451543116/buildlog_ubuntu-bionic-arm64.mir_1.5.0+dev82-gde6bc80ea4-0ubuntu18.04_BUILDING.txt.gz

PR #1074, opened 2019-11-14
Add wayland platform to debian/rules

AlanGriffiths:
Add wayland platform to debian/rules. (Fixes: #1073)


AlanGriffiths:
CI won't tell us anything, landing it right now.

PR #1076, opened 2019-11-14
fix-wayland-platform-test

AlanGriffiths:
Make the platform test and the wayland platform play nice

### DIFF
--- a/src/platforms/wayland/platform_symbols.cpp
+++ b/src/platforms/wayland/platform_symbols.cpp
@@ -64,7 +64,7 @@ mg::PlatformPriority probe_graphics_platform(
 {
     mir::assert_entry_point_signature<mg::PlatformProbe>(&probe_graphics_platform);
 
-    return mpw::connection(options) ?
+    return mpw::supported(options) ?
         mg::PlatformPriority::best :
         mg::PlatformPriority::unsupported;
 }
--- a/src/platforms/wayland/platform_symbols.cpp
+++ b/src/platforms/wayland/platform_symbols.cpp
@@ -64,9 +64,9 @@ mg::PlatformPriority probe_graphics_platform(
 {
     mir::assert_entry_point_signature<mg::PlatformProbe>(&probe_graphics_platform);
 
-    return mpw::supported(options) ?
-        mg::PlatformPriority::best :
-        mg::PlatformPriority::unsupported;
+    return mpw::supported(options)  ? mg::PlatformPriority::best :
+           mpw::connection(options) ? mg::PlatformPriority::supported :
+           mg::PlatformPriority::unsupported;
 }
 
 mir::ModuleProperties const* describe_graphics_module()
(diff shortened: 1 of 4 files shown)


src/platforms/wayland/platform_symbols.cpp
```diff
@@ -64,7 +64,7 @@ mg::PlatformPriority probe_graphics_platform(
 {
     mir::assert_entry_point_signature<mg::PlatformProbe>(&probe_graphics_platform);
 
-    return mpw::connection(options) ?
+    return mpw::supported(options) ?
         mg::PlatformPriority::best :
         mg::PlatformPriority::unsupported;
```
RAOF: The thought occurs that you *could* call `wl_display_connect(NULL)` and check for successful connection and return `::supported`.

RAOF: It's not necessary to make that change now, though.

tests/platform_test_harness/graphics_platform_test_harness.cpp
```diff
@@ -611,37 +611,34 @@ int main(int argc, char const** argv)
     bool success = true;
     try
     {
-        success &= test_probe(platform_dso, config);
-        if (success)
+        test_probe(platform_dso, config);
```
RAOF: I think this should still be `success &= test_probe()`; failure to return a supported probe value will result in the platform failing to work.

AlanGriffiths: Done

src/platforms/wayland/platform_symbols.cpp
AlanGriffiths: Done

src/platforms/wayland/wayland_display.cpp
```diff
@@ -48,11 +48,13 @@ void mpw::add_connection_options(boost::program_options::options_description& co
...
-
-    static auto const wayland_display =
-        std::make_unique<WaylandDisplay>(options.get<std::string>(wayland_host_option_name).c_str());
+    static auto const wayland_display = std::make_unique<WaylandDisplay>(options.is_set(wayland_host_option_name) ?
+        options.get<std::string>(wayland_host_option_name).c_str() : nullptr);
 
     return wayland_display->wl_display;
 }
+
+auto mir::platform::wayland::supported(mir::options::Option const& options) -> bool
```
RAOF: This is a weirdly-named function now; it's not returning whether mpw is *supported*, but whether it *should* be used. Given it's only used in one place, I think this could be just inlined into `probe_graphics_platform` with a suitable comment.

AlanGriffiths: Inlining would require access to (or duplication) of the option name. Simpler to use a better name.

AlanGriffiths:
bors r=RAOF

PR #1079, opened 2019-11-15
Add mir-spinner to mir-test-tools

RAOF:
This is the last bit of uninstalled stuff; packages now build locally, so *should* also build in the PPA!

### DIFF
--- a/debian/mir-test-tools.install
+++ b/debian/mir-test-tools.install
@@ -2,6 +2,7 @@ usr/bin/mir_stress
 usr/bin/mir_performance_tests
 usr/bin/mir-smoke-test-runner
 usr/bin/mir_platform_graphics_test_harness
+usr/bin/mir-spinner
 usr/lib/*/mir/tools/libmirclientlttng.so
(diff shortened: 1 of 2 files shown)


debian/mir-test-tools.install
```diff
@@ -1,6 +1,8 @@
 usr/bin/mir_stress
 usr/bin/mir_performance_tests
 usr/bin/mir-smoke-test-runner
+usr/bin/mir_platform_graphics_test_harness
```
AlanGriffiths: Isn't this line already on master?

RAOF: Yes, but that commit should be on master, too. I wonder why it's showing here again?

RAOF:
I *think* the answer as to why that line shows up again is that GitHub is confused. If that's not the case, then bors should cleanly fail to merge, so…

bors r=AlanGriffiths

RAOF:
Huh. Apparently GitHub mangled the git history in some way, so that previous commit wasn't picked up.

Anyway,
bors r=AlanGriffiths

PR #1081, opened 2019-11-15
Trying to blindly fix 1080

AlanGriffiths:
I *think* egl_wayland_allocator.cpp instantiates Texture::Texture() and toolchain bugs on 16.04/armhf that leads to LTO (incorrectly) complaining about a duplicate vtable.

This ought to stop that, but I don't have armhf kit to test locally.


Saviq:
What can go wrong…

bors r+

PR #1083, opened 2019-11-17
Adjust regex to fix build with googletest 1.10.0

z3ntu:
The regex in tests/CMakeLists.txt didn't account for two digits and failed for 1.10.0.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -43,7 +43,7 @@ else()
 endif()
 
 string(REGEX MATCH
-  "([0-9])\\.([0-9])\\.([0-9])" GTEST_VERSION_PARSED
+  "([0-9])\\.([0-9]+)\\.([0-9])" GTEST_VERSION_PARSED
   ${GTEST_VERSION_STR})
 if (GTEST_VERSION_PARSED)
   set(GTEST_VERSION_MAJOR ${CMAKE_MATCH_1})


tools/tsan-suppressions
```diff
@@ -1,6 +1,5 @@
 race:libglib*
 race:libgio*
-race:testing::internal::FunctionMockerBase
```
z3ntu: what about this? Leave it in? Remove it? I'm not sure what it's used for but I noticed it also references FunctionMockerBase so I didn't just want to ignore it

RAOF: You can remove it; this is just in the TSan suppression lists, and we're (sadly) not TSan clean, nor can we reasonably become so until TSan can understand GLib's locking primitives.

tests/CMakeLists.txt
```diff
@@ -43,7 +43,7 @@ else()
 endif()
 
 string(REGEX MATCH
-  "([0-9])\\.([0-9])\\.([0-9])" GTEST_VERSION_PARSED
+  "([0-9])\\.([0-9]+)\\.([0-9])" GTEST_VERSION_PARSED
```
RAOF: Ah, good catch! Thanks.

This should be the only necessary component of this PR; the code in `gmock_fixes.h` is only built for GMock < 1.8.1, so once the version detection is fixed everything should work!

tests/include/mir/test/gmock_fixes.h
```diff
@@ -77,9 +77,9 @@ class ActionResultHolder<std::unique_ptr<T, D>>
   // result in a new-ed ActionResultHolder.
   template <typename F>
   static ActionResultHolder* PerformDefaultAction(
-      const FunctionMockerBase<F>* func_mocker,
+      const FunctionMocker<F>* func_mocker,
```
RAOF: This breaks the build on old GoogleTest.

Thankfully, it's not actually necessary. Please drop it :grin:

z3ntu: I got this error when compiling without this PR: https://pastebin.com/raw/BJUaEcV3 . I then made this change and it got further until I hit something where I noticed that the GTEST_AT_LEAST macro was broken and discovered that the regex was broken. After I fixed that, the build went thorugh

tests/CMakeLists.txt
wmww: Any reason to add multi-digit support only for the minor version and not major or patch?

z3ntu: not really

PR #1085, opened 2019-11-21
rpi-dispmanx: Initial dumb accelerated GL client support

RAOF:
This requires a version of the rpi/userland with Wayland platform
support.

It currently has *terrible* performance - as far as I can tell there's
no way to import a dispmanx handle into the GPU to texture from, so this
does a copy from dispmanx handle (in GPU space) to CPU buffer and then
*uploads* that into a GL texture.

This is, you will not be terribly shocked to discover, not high
performance.

Fortunately, the dispmanx 2D composition engine is quite flexible, so we
should be able to eliminate this overhead by directly adding the
client's dispmanx buffers to the overlay planes, entirely avoiding the
CPU bouncing necessary here.


RAOF:
This is stacked-on #1028. It's *even more* difficult to build than that, because you need to use [my branch of the rpi/userland](https://github.com/MirServer/rpi-userland) repository.

It's also a bit broken, still. Rendering appears correct, but clients deadlock after an non-constant time; that might be a race in the client EGL, as I see the it receive the relevant `wl_buffer.release()`.

PR #1087, opened 2019-11-22
Use EGL_NO_X11 as well as MESA_EGL_NO_X11_HEADERS

### LINKED ISSUE #1086: Use EGL_NO_X11 instead of MESA_EGL_NO_X11_HEADERS
Building on Arch I'm getting errors like
```
In file included from /usr/include/EGL/eglplatform.h:128,
                 from /usr/include/EGL/egl.h:39,
                 from /home/wmww/code/mir/include/platform/mir/client/client_buffer.h:32,
                 from /home/wmww/code/mir/src/client/presentation_chain.cpp:20:
/usr/include/google/protobuf/stubs/logging.h:68:7: error: expected identifier before ‘int’
   68 | class Status;
      |       ^~~~~~
```
and
```
In file included from /usr/include/X11/Xlib.h:44,
                 from /usr/include/EGL/eglplatform.h:128,
                 from /home/wmww/code/mir/include/platform/mir/client/client_platform.h:26,
                 from /home/wmww/code/mir/tests/include/mir_test_framework/stub_client_platform_factory.h:23,
                 from /home/wmww/code/mir/tests/mir_test_framework/stub_client_platform_module.cpp:22:
/usr/include/gtest/internal/gtest-type-util.h:125:8: error: expected identifier before numeric constant
  125 | struct None {};
      |        ^~~~
```
It appears this is caused by names like `Status` and `None` being #defined in X11 headers (in Xlib.h and X.h respectively) and EGL headers including them. I'm not sure why this problem is only showing up now, as these sections of Mir and the X headers have been unchanged for a long time. My only guess is that EGL changed to include the X headers for some reason.

The way Xlib's `Status` conflicts with protobuf has been reported here: https://github.com/protocolbuffers/protobuf/issues/4186

Related to #1084

wmww:
Only tested on Arch so far. I don't know how long `EGL_NO_X11` has been around. If this fails on older distros, I'll just use both defines.

Fixes #1086

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,6 +214,7 @@ add_definitions(
 endif(MIR_DISABLE_EPOLL_REACTOR)
 
 add_definitions(-DEGL_NO_X11)
+add_definitions(-DMESA_EGL_NO_X11_HEADERS) # Can be removed when all platforms support EGL_NO_X11
 
 pkg_check_modules(WAYLAND_EGLSTREAM wayland-eglstream)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,7 +213,7 @@ add_definitions(
 )
 endif(MIR_DISABLE_EPOLL_REACTOR)
 
-add_definitions(-DMESA_EGL_NO_X11_HEADERS)
+add_definitions(-DEGL_NO_X11)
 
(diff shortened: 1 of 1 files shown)


mariogrip:
This seems to break everything else lol :P can't you just define both?

wmww:
I was hoping `EGL_NO_X11` had been around for a while, but apparently not. Yes, I can define both

PR #1092, opened 2019-11-25
Report wl_display_add_socket() errors

AlanGriffiths:
If wl_display_add_socket() fails, report an error instead of silently carrying on.

This happens, for example, under GNOME/Wayland as WAYLAND_DISPLAY points to the GNOME socket.

### DIFF
--- a/tests/mir_test_framework/test_wlcs_display_server.cpp
+++ b/tests/mir_test_framework/test_wlcs_display_server.cpp
@@ -819,7 +819,9 @@ miral::TestWlcsDisplayServer::TestWlcsDisplayServer(int argc, char const** argv)
     WlcsDisplayServer::create_touch = &wlcs_server_create_touch;
 
     add_to_environment("MIR_SERVER_ENABLE_KEY_REPEAT", "false");
-    add_to_environment("WAYLAND_DISPLAY", "wlcs-tests");
+    char buffer[32];
+    sprintf(buffer, "wlcs-tests-%d", getpid());
+    add_to_environment("WAYLAND_DISPLAY", buffer);
 
     add_server_init([this](mir::Server& server)
         {
--- a/tests/mir_test_framework/test_wlcs_display_server.cpp
+++ b/tests/mir_test_framework/test_wlcs_display_server.cpp
@@ -820,7 +820,7 @@ miral::TestWlcsDisplayServer::TestWlcsDisplayServer(int argc, char const** argv)
 
     add_to_environment("MIR_SERVER_ENABLE_KEY_REPEAT", "false");
     char buffer[32];
-    sprintf(buffer, "wlcs-tests-%d", getpid());
+    snprintf(buffer, sizeof buffer, "wlcs-tests-%d", getpid());
     add_to_environment("WAYLAND_DISPLAY", buffer);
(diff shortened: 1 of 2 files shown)


tests/mir_test_framework/test_wlcs_display_server.cpp
```diff
@@ -819,7 +819,9 @@ miral::TestWlcsDisplayServer::TestWlcsDisplayServer(int argc, char const** argv)
     WlcsDisplayServer::create_touch = &wlcs_server_create_touch;
 
     add_to_environment("MIR_SERVER_ENABLE_KEY_REPEAT", "false");
-    add_to_environment("WAYLAND_DISPLAY", "wlcs-tests");
+    char buffer[32];
+    sprintf(buffer, "wlcs-tests-%d", getpid());
+    add_to_environment("WAYLAND_DISPLAY", buffer);
```
wmww: Perhaps this would be cleaner and more C++ish:
```
auto wlcs_display = "wlcs-tests-" + std::to_string(getpid());
add_to_environment("WAYLAND_DISPLAY", wlcs_display.c_str());
```

AlanGriffiths: I guess by "cleaner" you means direct initialization? But dynamically allocating two buffers in the initialization expression and using RAII to release them doesn't seem cleaner to me. That's the sort of hidden cost that gives C++ a bad name. (Not that the cost actually matters here.)

@RAOF WDYT?

AlanGriffiths: `add_to_environment("WAYLAND_DISPLAY", ("wlcs-tests-" + std::to_string(getpid())).c_str());`

wmww: I'd like to point out that code likely to lead to buffer overflows when modified in the name of saving a single allocation is what gives C a bad name. I'm fine with the one-line version as well. It's the hard-coded buffer size that gives me the hebegebees.

AlanGriffiths: OK, fixed.

wmww: Is it really though? I wasn't worried about a very large PID as much as someone lengthening the socket name without increasing the buffer size. I suppose incorrect behavior *is* an improvement over undefined behavior. @RAOF can bors or critique as he likes.

RAOF: I'd probably mildly prefer @wmww 's first suggestion (although I'd add a `const` :wink:). (If you wanted to get correct output, no buffer overflows, *and* just a single allocation you could use `asprintf`, but I wouldn't bother)

I wouldn't block the current `snprintf`, though. That's fine.

RAOF:
bors r=AlanGriffiths

RAOF:
Whoops.

bors delegate=AlanGriffiths

AlanGriffiths:
I'm interested to see if this helps with some of the intermittent wlcs failures we've seen in CI: https://github.com/MirServer/wlcs/issues/104

bors r=wmww,RAOF

Issue #1091, opened 2019-11-25
No pointer or touch events using `--window-manager system-compositor` with Wayland clients


AlanGriffiths:
(On master, this isn't yet released)

Switch to a VT and start a Mir "host":
```plain
miral-shell --window-manager system-compositor
```

From another machine ssh in and start a nested Mir:
```plain
WAYLAND_DEBUG=client miral-app --wayland-host wayland-1 2>&1 | grep wl_touch
```
(Adjust wayland-1 to the socket exposed by the host.)

Try touching the screen.

Expect: touch events logged
Actual: no touch events logged

A bit more: using `miral-shell` without the `--window-manager system-compositor` the touch events are seen.

AlanGriffiths:
There's the same problem with pointer events

AlanGriffiths:
Oddly, keyboard events travel much the same codepath but successfully "arrive".

One for tomorrow.

AlanGriffiths:
After tracing this through, there's a custom input rectangle: `{0, 0, 0, 0}` that filters out the pointer and touch events. (With the other `--window-manager` options this rectangle is the size of the window.)

At least that explains what is happening, now for why...

PR #1093, opened 2019-11-26
Don't set input regions when they are not set

AlanGriffiths:
Don't set input regions when they are not set. (Fixes: #1091)

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -418,8 +418,10 @@ void mf::WindowWlSurfaceRole::create_mir_window()
 {
     params->size = pending_size();
     params->streams = std::vector<shell::StreamSpecification>{};
-    params->input_shape = std::vector<geom::Rectangle>{};
-    surface->populate_surface_data(params->streams.value(), params->input_shape.value(), {});
+
+    std::vector<geom::Rectangle> input_shape;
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
This looks a bit ugly but as `WlSurface::populate_surface_data()` is used recursively it isn't just a matter of passing in an `optional<>`.

AlanGriffiths:
> If a toplevel has no input region set but a subsurface does, wouldn't that cause the surface to only get input in the subsurface's region? 

Ouch! I didn't run the tests before proposing. Yes, there are scenarios that this breaks (as well as the one this fixes).

> I think the real problem is that `populate_surface_data()` isn't being called or isn't correctly setting the input region. If a client hasn't set an input shape, `populate_surface_data()` should be being called every time a differently sized buffer is committed (`WlSurface::commit()` calls `state.invalidate_surface_data()`, which then triggers the update).

Well, the real problem is a mismatch between the historical model of input regions in Mir and the the model implicit in the Wayland protocols. Both make the input region(s) optional, but in Mir they are associated with windows, not (as they are in Wayland) surfaces. When a window has multiple surfaces, only some of which have input regions set it gets messy to synchronize these approaches.

I think the real solution is to change the Mir model of input regions to match Wayland. That's a bit invasive and should probably go on the list for after dropping the mirclient code.

I have an alternative quick fix for #1091 that I'll PR in a few minutes.

Issue #1094, opened 2019-11-27
Input regions are not completely Wayland compatible


AlanGriffiths:
There is a mismatch between the historical model of input regions in Mir and the the model implicit in the Wayland protocols. Both make the input region(s) optional, but in Mir they are associated with windows, not (as they are in Wayland) surfaces. When a window has multiple surfaces, only some of which have input regions set it gets messy to synchronize these approaches.

The solution is to change the Mir model of input regions to match Wayland. That's a bit invasive and should be addressed after dropping the mirclient code.

PR #1095, opened 2019-11-27
Clear input regions when they should not be set

### LINKED ISSUE #1091: No pointer or touch events using `--window-manager system-compositor` with Wayland clients

AlanGriffiths:
Clear input regions when they should not be set. (Fixes #1091)

This is a workaround for dubious logic documented in #1094 and discussed in comments to #1093.

### DIFF
--- a/src/server/shell/system_compositor_window_manager.cpp
+++ b/src/server/shell/system_compositor_window_manager.cpp
@@ -73,6 +73,13 @@ auto msh::SystemCompositorWindowManager::add_surface(
     placed_parameters.top_left = rect.top_left;
     placed_parameters.size = rect.size;
 
+    if (placed_parameters.input_shape.is_set())
+    {
+        // Shouldn't be set anyway as it doesn't make sense for a hosted session.
+        // It is currently set in frontend_wayland, even if the client doesn't want it, as a workaround for #1094
+        placed_parameters.input_shape.consume();
+    }
+
     auto const surface = build(session, placed_parameters);
 
     auto const session_ready_observer = std::make_shared<SurfaceReadyObserver>(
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
> The Wayland frontend _should_ be setting and updating the input region as needed. The fact that the input region is being set incorrectly would indicate there is a bug in the frontend.

I disagree, setting the input region is only the responsibility of the frontend because of a mismatch between the way Wayland and mirclient work. See #1094.

I don't object to "fixing" the frontend to handle the mismatch better (if that also fixes the problem), but there's no need to spend time on it right now.

AlanGriffiths:
@wmww I've a different fix that will be more to your liking...

wmww:
I think the sensible implementation may be to have a separate `BasicInputSurface` that implements `input::Surface` and have `BasicSurface` manage a list of them. Each subsurface would have both a `BufferStream` **and** an `InputSurface`. That, however, is an entirely separate problem from the system compositor window manager failing to respect the requested input region.

AlanGriffiths:
> I think the sensible implementation may be to have a separate `BasicInputSurface` that implements `input::Surface` and have `BasicSurface` manage a list of them. Each subsurface would have both a `BufferStream` **and** an `InputSurface`. That, however, is an entirely separate problem from the system compositor window manager failing to respect the requested input region.

Yes, as discussed [elsewhere](#1094), we should revisit once we no longer have to support the mirclient related codepaths.

bors r=wmww

PR #1102, opened 2019-11-28
[spread] add sbuild builds (Fixes #1005)

### LINKED ISSUE #1005: Mir CI does not accurately test Ubuntu packaging
At the cost of an increase in CI time we could build packages in a more LP-buildd-like environment, which would catch a large proportion of the cases where our CI passes but PPA builds fail.

Saviq:


RAOF:
I'm not entirely sure if it's relevant, but you can use LXD as a backend for sbuild; that *might* be faster, depending on the relative speeds to the LXD image server vs `archive.ubuntu.com` + package unpack.

Issue #1096, opened 2019-11-27
FTBFS in LP


AlanGriffiths:
The wlcs tests are failing as `XDG_RUNTIME_DIR` is not set (and that causes `wl_display_add_socket()` to fail in a way we used to ignore prior to #1092):
```plain
error: XDG_RUNTIME_DIR not set in the environment
ERROR: /<<BUILDDIR>>/mir-1.5.0+dev150-gc82f4ed32b/src/server/frontend_wayland/wayland_connector.cpp(691): Throw in function mir::frontend::WaylandConnector::WaylandConnector(const std::shared_ptr<mir::shell::Shell>&, const std::shared_ptr<mir::frontend::MirDisplay>&, const std::shared_ptr<mir::input::InputDeviceHub>&, const std::shared_ptr<mir::input::Seat>&, const std::shared_ptr<mir::graphics::GraphicBufferAllocator>&, const std::shared_ptr<mir::frontend::SessionAuthorizer>&, bool, std::unique_ptr<mir::frontend::WaylandExtensions>, const WaylandProtocolExtensionFilter&)
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::system_error> >
std::exception::what: Failed to add Wayland socket: No such file or directory
```

PR #1103, opened 2019-11-28
Report wl_display_add_socket() errors

AlanGriffiths:
If wl_display_add_socket() fails, report an error instead of silently carrying on.

This happens, for example, under GNOME/Wayland as WAYLAND_DISPLAY points to the GNOME socket.

(It also causes wlcs tests to fail if XDB_RUNTIME_DIR is not set, as on LP builders, so work around that too.)


AlanGriffiths:
> LGTM, but how is this different than the previous PR? If it's just the last commit, can you rebase on master so GitHub detects that?

It is just the last commit. The previous PR was reverted because it [FTBFS in LP](#1096).

bors r=wmww

Issue #1106, opened 2019-11-28
Crashes & lockups


AlanGriffiths:
1. run `miral-shell`
2. start `wayland-terminal` with Ctrl-Alt-Shift-T
3. start glmark with `glmark2-wayland --fullscreen`
4. Send "close window" to glmark with `Alt-F4`

Expect: glmark closes and `miral-shell` continues to run.
Actual: `miral-shell` crashes or locks up, sometimes stealing keyboard input.

AlanGriffiths:
Some dump info from one of the crashes:
```plain
(gdb) bt
#0  0x00007fdf5679efa0 in __GI___pthread_mutex_lock (mutex=0xd0) at ../nptl/pthread_mutex_lock.c:65
#1  0x00007fdf55a73dd7 in __gthread_mutex_lock(__gthread_mutex_t*) (__mutex=0xd0)
    at /usr/include/x86_64-linux-gnu/c++/7/bits/gthr-default.h:748
#2  0x00007fdf55a7468a in std::mutex::lock() (this=0xd0) at /usr/include/c++/7/bits/std_mutex.h:103
#3  0x00007fdf55a75563 in std::unique_lock<std::mutex>::lock() (this=0x7fdf3bffe020)
    at /usr/include/c++/7/bits/std_mutex.h:267
#4  0x00007fdf55a74d9c in std::unique_lock<std::mutex>::unique_lock(std::mutex&) (this=0x7fdf3bffe020, __m=...)
    at /usr/include/c++/7/bits/std_mutex.h:197
#5  0x00007fdf55a97898 in mir::RecursiveReadWriteMutex::read_lock() (this=0xd0)
    at /home/alan/display_server/mir/src/common/thread/recursive_read_write_mutex.cpp:27
#6  0x00007fdf55b9143f in mir::RecursiveReadLock::RecursiveReadLock(mir::RecursiveReadWriteMutex&) (this=0x7fdf3bffe0b0, mutex=...) at /home/alan/display_server/mir/src/include/common/mir/recursive_read_write_mutex.h:59
#7  0x00007fdf55c5cd48 in mir::ThreadSafeList<std::shared_ptr<mir::scene::SurfaceObserver> >::for_each(std::function<void (std::shared_ptr<mir::scene::SurfaceObserver> const&)> const&) (this=0x7fdf3000d9f8, f=...)
    at /home/alan/display_server/mir/src/include/common/mir/thread_safe_list.h:78
[...]

PR #1108, opened 2019-11-29
Fix thread-unsafe code

AlanGriffiths:
Fix thread-unsafe code. (Fixes: #1106)

The "frame posted" callback set in BasicSurface::set_streams() can be invoked (e.g. by WlSurface::commit()) after BasicSurface::observers had been destroyed (with the surface). Switch to a shared pointer to ensures that the SurfaceObservers are not destroyed while they are in use.


AlanGriffiths:
No test as that's flaky for thread safety concerns and the design change should be easy enough to "prove" with a little thought.

PR #1111, opened 2019-12-02
mg::common::ShmBuffer: Drop legacy interfaces.

RAOF:
We no longer need to support `renderer::gl::TextureSource`, as our renderer can use the `mg::gl::Texture` interface (as can QtMir, our only other known renderer implementation).

`renderer::gl::TextureSource` is only used by the Screencast interface, which is mirclient only.

This lets us drop the funky disambiguating classes required to let ShmBuffer implement both `mg::gl::Texture` and `renderer::gl::TextureSource`.

### DIFF
--- a/src/platforms/common/server/shm_buffer.cpp
+++ b/src/platforms/common/server/shm_buffer.cpp
@@ -127,27 +127,6 @@ MirPixelFormat mgc::ShmBuffer::pixel_format() const
     return pixel_format_;
 }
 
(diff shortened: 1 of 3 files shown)


RAOF:
Note that this probably won't affect UBports too much, as the Android platform does *not* use `mg::common::ShmBuffer`; it provides its own ShmBuffer implementation, which still exposes all these interfaces.

AlanGriffiths:
bors r=wmww

PR #1113, opened 2019-12-02
Add Surface::set_window_margins() for SSD

wmww:
This allows setting a surface's "window margins" (the padding around a surface that makes room for decorations). There are alternative ways we could have done this, each with its advantages and disadvantages. We settled on this approach several weeks ago because it's easy to implement with minimum code churn, and appears to be similar to what whoever was here last intended to do. In the future (especially if we split up the `scene::Surface` interface) we may want to refactor this, but it has proven to be sufficient for usable SSDs.

### DIFF
--- a/include/server/mir/scene/surface.h
+++ b/include/server/mir/scene/surface.h
@@ -142,6 +142,15 @@ public:
 
     /// The session this surface was created by
     virtual auto session() const -> std::weak_ptr<Session> = 0;
+
+    /// Sets the geometry of the margins around a surface
+    /// Margins make room for server-side-decorations, which are attached as a child surface
+    /// This call will trigger the content_size to be adjusted and the window_size will remain unchanged
+    virtual void set_window_margins(
+        geometry::DeltaY top,
+        geometry::DeltaX left,
+        geometry::DeltaY bottom,
+        geometry::DeltaX right) = 0;
 };
 }
 }
--- a/include/test/mir/test/doubles/stub_surface.h
+++ b/include/test/mir/test/doubles/stub_surface.h
@@ -33,6 +33,7 @@ struct StubSurface : scene::Surface
     std::string name() const override;
     void move_to(geometry::Point const& top_left) override;
     geometry::Size window_size() const override;
+    geometry::Displacement content_offset() const override;
     geometry::Size content_size() const override;
     std::shared_ptr<frontend::BufferStream> primary_buffer_stream() const override;
     void set_streams(std::list<scene::StreamInfo> const& streams) override;
--- a/include/test/mir/test/doubles/stub_surface.h
+++ b/include/test/mir/test/doubles/stub_surface.h
@@ -78,7 +78,12 @@ struct StubSurface : scene::Surface
     void set_focus_state(MirWindowFocusState new_state) override;
     std::string application_id() const override;
     void set_application_id(std::string const& application_id) override;
-    std::weak_ptr<scene::Session> session() const  override;
+    std::weak_ptr<scene::Session> session() const override;
+    void set_window_margins(
+        geometry::DeltaY top,
+        geometry::DeltaX left,
+        geometry::DeltaY bottom,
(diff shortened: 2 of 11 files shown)


include/test/mir/test/doubles/stub_surface.h
```diff
@@ -78,7 +78,12 @@ struct StubSurface : scene::Surface
     void set_focus_state(MirWindowFocusState new_state) override;
     std::string application_id() const override;
     void set_application_id(std::string const& application_id) override;
-    std::weak_ptr<scene::Session> session() const  override;
+    std::weak_ptr<scene::Session> session() const override;
```
wmww: This line is modified because I removed an excess space before `override`

include/server/mir/scene/surface.h
```diff
@@ -142,6 +142,16 @@ class Surface :
 
     /// The session this surface was created by
     virtual auto session() const -> std::weak_ptr<Session> = 0;
+
+    /// Sets the dimensions of the window frame
+    /// It is assumed that any decorations are provided by a child window
+    /// Frame size is not directly accessable, but can be determined by comparing client_size() and frame_size()
```
AlanGriffiths: What is `frame_size()`?

If it's a misnomer for `window_size()` then maybe this function should be `set_window_margins()`?

include/server/mir/scene/surface.h
```diff
@@ -142,6 +142,16 @@ class Surface :
...
+    /// This will trigger a resize in which the client size is adjusted and the frame size remains unchanged
+    virtual void set_frame_geometry(
+        geometry::Height top,
+        geometry::Width left,
+        geometry::Height bottom,
+        geometry::Width right) = 0;
```
AlanGriffiths: Unless I misunderstand the intent, these should be `DeltaY` & `DeltaX`, not `Width` & `Height`?

src/server/scene/basic_surface.cpp
```diff
@@ -1052,3 +1056,47 @@ auto mir::scene::BasicSurface::session() const -> std::weak_ptr<Session>
...
+        observers->content_resized_to(this, size);
+    }
+}
+
+auto mir::scene::BasicSurface::content_size(ProofOfMutexLock const&) const -> geometry::Size
+{
+    return geom::Size{
+        surface_rect.size.width  - as_delta(frame.left) - as_delta(frame.right),
+        surface_rect.size.height - as_delta(frame.top)  - as_delta(frame.bottom)};
+}
```
AlanGriffiths: We should be storing the margins as deltas not converting them wherever used.

include/server/mir/scene/surface.h
```diff
@@ -142,6 +142,15 @@ class Surface :
...
     virtual auto session() const -> std::weak_ptr<Session> = 0;
+
+    /// Sets the geometry of the margins around a surface
+    /// Margins make room for server-side-decorations, which are attached as a child surface
+    /// This call will trigger the content_size to be adjusted and the window_size will remain unchanged
+    virtual void set_window_margins(
+        geometry::DeltaY top,
+        geometry::DeltaX left,
+        geometry::DeltaY bottom,
+        geometry::DeltaX right) = 0;
```
AlanGriffiths: I'm tempted by:
```c++
    virtual void set_window_geometry(
        geometry::DeltaY top_margin,
        geometry::DeltaX left_margin,
        geometry::DeltaY bottom_margin,
        geometry::DeltaX right_margin) = 0;
```
but I'm not going to block on it. (Just mention it and see what happens.)

wmww: No, on account of there being a confusingly similar but distinct Wayland message called `set_window_geometry`.

wmww:
Added since review: `Point += Displacement` and `Point -= Displacement` geometry operators, a new `content_offset()` surface property and a test that `AbstractShell` adjusts the aux rect provided by the frontend.

PR #1119, opened 2019-12-04
SSD surface creation param

wmww:
This lets the XWayland frontend mark its windows as server-side-decorated and causes the shell to decorate them.

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -30,6 +30,7 @@
 #include "mir/frontend/wayland.h"
 #include "null_event_sink.h"
 
+#include "mir/log.h"
 #include "mir/scene/surface.h"
 #include "mir/scene/surface_creation_parameters.h"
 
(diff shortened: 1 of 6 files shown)


src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -275,6 +276,18 @@ void mf::WindowWlSurfaceRole::set_fullscreen(std::experimental::optional<struct
     }
 }
 
+void mf::WindowWlSurfaceRole::set_server_side_decorated(bool server_side_decorated)
+{
+    if (weak_scene_surface.lock())
+    {
+        log_warning("server_side_decorated decorated property can not be changed after surface created");
```
AlanGriffiths: Why not?

AFAICS `zxdg_toplevel_decoration_v1.set_mode/unset_mode` can be called at any time.

wmww: I haven't added the property to `SurfaceSpecification`, only `SurfaceCreationParamaters`. Such a property could be added in the future, but there isn't a need yet.

AlanGriffiths: In that case, can reword the log message as it is more a matter of "not implemented" than "not possible". (E.g. adding "yet".)

mariogrip:
Is there a way to disable SSD drawn by mir? as unity8 draws it's own

wmww:
There is currently not a way to do this. I have made an issue about it: #1417

PR #1121, opened 2019-12-04
SSD MVP

wmww:
Not based on other PRs, but won't take effect until combined with #1119. After that decorations should appear on all X11 windows. Adds a working title bar with resize edges and minimize/maximize/close buttons.

### DIFF
--- a/src/server/shell/decoration/basic_decoration.cpp
+++ b/src/server/shell/decoration/basic_decoration.cpp
@@ -1,0 +1,392 @@
+/*
+ * Copyright © 2019 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored By: William Wold <william.wold@canonical.com>
+ */
+
+#include "basic_decoration.h"
+#include "window.h"
+#include "input.h"
+#include "renderer.h"
+#include "threadsafe_access.h"
+
+#include "mir/executor.h"
+#include "mir/shell/shell.h"
+#include "mir/shell/surface_specification.h"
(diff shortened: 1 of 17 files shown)


AlanGriffiths:
As well as the compiler errors, resizing needs to be constrained so that the client window has a sane size.

src/server/shell/decoration/basic_decoration.cpp
```diff
@@ -158,6 +158,8 @@ msd::BasicDecoration::BasicDecoration(
     std::shared_ptr<ms::Surface> const& window_surface)
     : threadsafe_self{std::make_shared<ThreadsafeAccess<BasicDecoration>>(this, executor)},
       static_geometry{std::make_unique<StaticGeometry>(default_geometry)},
+      window_state{nullptr},
+      input_state{nullptr},
```
AlanGriffiths: Why? The default initialization does the right thing.

wmww: I just want to make it explicit that null initialization isn't a mistake.

AlanGriffiths: :man_shrugging:

AlanGriffiths:
I think what's missing is mapping from the size constraints set for the client_size to constraints for the window_size seen by Window Management. That hasn't been a distinction before, but becomes one with decorations.

From a quick examination that needs a chunk of infrastructure that could be landed separately (and first).

wmww:
Size constraints don't work at all with XWayland surfaces, and I don't know why. Doesn't matter for now. Crashes seemed to be caused by logic bugs in the decoration, which have now been fixed.

AlanGriffiths:
The latest changes do indeed prevent a crash, but they don't fix the logic that leads to the crash.

Take a simple X11 application (I'm using xeyes) and drag the bottom margin up: you can drag it into the titlebar. The resize ought to be constrained before that happens (try it with a GTK or Qt app.)

The same is true of dragging the right margin into the left etc. (This also leads to weird visuals: the horizontal margins are not as wide as the vertical ones.)

wmww:
> they don't fix the logic that leads to the crash

Lead to the crash (past tense). The crash was caused by a decoration drawing bug, which is fixed.

As I said, size constraints seem to have no effect on XWayland windows (with or without decorations). This just wasn't apparent before because there was no way to resize them. This is an issue for another day.

wmww:
Update: it appears max/min size are ignored for all surfaces except when they are initially set. Wayland surfaces appear to have a min size only because the client refuses to draw smaller buffers, and we don't clip.

wmww:
Update: I fixed max/min size constraints on Friday and will PR that once I write tests on Monday or Tuesday

PR #1122, opened 2019-12-05
Split mgc::ShmBuffer into a storageless-base class and memory-backed implementation

RAOF:
`mgc::ShmBuffer` is now a usable base for a `gl::Texture` supporting view on a CPU-mapped buffer.

`mgc::MemoryBackedShmBuffer` is a full implementation (including a `sw::PixelSource`) that backed by a self-allocated memory buffer. It drops the `mmap`ed file implementation, as this is only required to support deprecated mirclient APIs.

This is then usable as-is for platforms, so the existing platforms are switched over.


RAOF:
bors try

PR #1129, opened 2019-12-09
Correctly constrain window size

wmww:
Max and min window size were only being respected when first set. This PR constrains the window size whenever the window is resized. Tests included.

EDIT: note that this PR alone will not fix SSD size constraints, but it will fix the visible problem of windows moving weirdly when you resize them too small.


AlanGriffiths:
Looks good. But some broken tests.

bors delegate=wmww

Issue #1125, opened 2019-12-06
ShmBuffer leaks GL resources


RAOF:
Urgh.

If a `ShmBuffer` has been used for GL texturing (and so has a non-zero `tex_id`) `mgc::ShmBuffer::~ShmBuffer()` will call `glDeleteTextures()` to delete that texture.

Unfortunately, `mg::Buffer`s are not guaranteed to be destroyed on a thread with a current EGL context, and in that case `glDeleteTextures` will silently fail, leaving the texture allocated.

This happens for (at least) cursor updates, which occur from the Wayland mainloop which does not have an active EGL context.

The buffers from `mg::wayland::buffer_from_resource` don't leak, as they carry a `renderer::gl::Context` around with them (instead, they may suffer from #1124)

PR #1136, opened 2019-12-11
ShmBuffer: Ensure GL resources are released

RAOF:
This sets up a dedicated thread and EGL context for running ShmBuffer code on.

At this point only the GL texture deletion is run there, fixing #1125.

### DIFF
--- a/src/platforms/common/server/egl_context_delegate.h
+++ b/src/platforms/common/server/egl_context_delegate.h
@@ -24,6 +24,7 @@
 #include <thread>
 #include <condition_variable>
 #include <mutex>
+#include <vector>
 
 namespace mir
 {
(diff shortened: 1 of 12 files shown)


RAOF:
(This is stacked on the other ShmBuffer branch, so will need to be rescued once that's merged)

src/platforms/common/server/egl_context_delegate.h
```diff
@@ -0,0 +1,69 @@
...
+{
+class Context;
+}
+}
+
+namespace graphics
+{
+namespace common
+{
+class EGLContextDelegate
```
RAOF: Suggestions for a better name are appreciated :)

AlanGriffiths: Yes, naming is hard.

`SurrogateEGLContext`?

RAOF: It's kinda an `EGLContextExecutor`, really…

PR #1137, opened 2019-12-11
x11 tweaks

AlanGriffiths:
Drop some code used in early testing.

Add a configuration option to specify the Xwayland executable.


AlanGriffiths:
@wmww the `--xwayland-path` option here is what egmde-snap needs to package its own Xwayland binary without polluting the app environment.

PR #1138, opened 2019-12-11
Refactor decoration

wmww:
* Add a few properties to the `StaticGeometry` struct that were previously magic numbers strewn about the code
* Make `static_geometry` a shared pointer and give all decoration components that need it a copy (instead of having to request it from a `WindowState`
* Give `Renderer` a new `Theme` struct that can be changed when switching between the focused and unfocused look

A PR adding title text rendering is to come after this is merged.


wmww:
All instances of `StaticGeometry` are already const, but sure why not

PR #1139, opened 2019-12-11
ShmBuffer: Ensure GL resources are released

RAOF:
This sets up a dedicated thread and EGL context for running ShmBuffer code on.

At this point only the GL texture deletion is run there, fixing #1125.

### DIFF
--- a/src/platforms/common/server/egl_context_delegate.h
+++ b/src/platforms/common/server/egl_context_delegate.h
@@ -24,6 +24,7 @@
 #include <thread>
 #include <condition_variable>
 #include <mutex>
+#include <vector>
 
 namespace mir
 {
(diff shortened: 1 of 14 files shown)


src/platforms/common/server/egl_context_delegate.h
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Run a run a function on a thread with a current EGL context
+     */
+    void defer_to_egl_context(std::function<void()>&& functor);
+private:
+    static void process_loop(EGLContextDelegate* const me);
+
+    std::unique_ptr<renderer::gl::Context> const ctx;
+    std::mutex mutex;
+    std::condition_variable new_work;
```
wmww: `std::condition_variable` is a good thing to know exists

wmww:
The ought to pass first
bors r-
bors delegate+

src/platforms/common/server/egl_context_executor.h
```diff
@@ -0,0 +1,71 @@
...
+{
+class Context;
+}
+}
+
+namespace graphics
+{
+namespace common
+{
+class EGLContextExecutor : public Executor
```
RAOF: Bikeshed resolution: call it an `EGLContextExecutor`, that's roughly what it is. Because it is, implement mir::Executor on it.

PR #1140, opened 2019-12-12
WaylandAllocator: Add buffer_from_shm

RAOF:
This allows the graphics platform to provide special handling for `wl_shm` buffers, and provides a default handler for platforms which don't care to do anything special.


RAOF:
This is stacked-on #1139

PR #1144, opened 2019-12-13
Draw window titles on SSDs

wmww:
Uses FreeType to render window titles on server side decorations. It appears the current XWayland implementation does not update the window title after it is initially set, but this will handle title updates when it happens.

### DIFF
--- a/src/server/shell/decoration/renderer.cpp
+++ b/src/server/shell/decoration/renderer.cpp
@@ -96,6 +96,15 @@ public:
         Pixel color) override;
 
 private:
+    struct FontPath
+    {
+        std::string const filename;
+        std::vector<std::string> prefixes;
+    };
+
+    static std::vector<FontPath> const font_paths;
+    static std::vector<std::string> const font_path_search_paths;
+
     std::mutex mutex;
     FT_Library library;
     FT_Face face;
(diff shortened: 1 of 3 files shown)


src/server/shell/decoration/renderer.cpp
```diff
@@ -74,6 +80,300 @@ inline void render_row(
...
+        "dejavu",               // Ubuntu (others?)
+        "",                     // Arch
+    }},
+};
+
+std::vector<std::string> const msd::Renderer::Text::Impl::font_path_search_paths{
+    "/usr/share/fonts/truetype",    // Ubuntu/Debian
+    "/usr/share/fonts/TTF",         // Arch
+    "/usr/share/fonts",             // Fedora/Arch
+};
```
AlanGriffiths: I know there are plans to make vector and string constexpr, but do we need to construct user-defined types here?

The code is simpler (no calls to `c_str()`) with a `char const* const ...font_path_search_paths[]`.

src/server/shell/decoration/renderer.cpp
```diff
@@ -74,6 +80,300 @@ inline void render_row(
...
+            }
+        }
+    }
+}
+
+auto msd::Renderer::Text::Impl::font_path() -> std::string
+{
+    // Similar to default_font() in examples/example-server-lib/wallpaper_config.cpp
+
+    std::vector<std::string> usable_search_paths;
```
AlanGriffiths: Shouldn't we use `boost::filesystem`? We already have dependencies on it.

AlanGriffiths:
While testing this I got a weird state:

Command|comment
--|--
`miral-app -demo-server --x11-display 3`|Run test binary
`export DISPLAY=:3`<br/>`xeyes`|In the nested terminal start xeyes 
&nbsp;|Move and resize xeyes

Unexpectedly there are two xeyes windows, but resizing one affects the client of the other.

image

wmww:
Hmm.. I see various strange thing, but I can not reproduce this one. Almost certainly not related to this PR. same in master?

AlanGriffiths:
> Hmm.. I see various strange thing, but I can not reproduce this one. Almost certainly not related to this PR. same in master?

Been testing various things, not seen this until now but I can't reproduce (with the same binary) either.

src/server/shell/decoration/renderer.cpp
```diff
@@ -49,6 +56,33 @@ uint32_t const default_unfocused_background = color(0x80, 0x80, 0x80);
 uint32_t const default_focused_text         = color(0xFF, 0xFF, 0xFF);
 uint32_t const default_unfocused_text       = color(0xA0, 0xA0, 0xA0);
 
+struct FontPath
+{
+    char const* filename;
+    std::vector<char const*> prefixes;
+};
+
+FontPath const font_paths[]{
```
RAOF: Could we make this a CMake variable to let distros handle this as per their own policies/locations?

AlanGriffiths: Would that be able to work by default? I don't know of a standard way to query the build system for the distro policy.

Or would anyone building from source have to deal with specifying the font location as an extra step? (Doing the same sort of polling in CMake is a bad idea as the build system need not have fonts installed to be detected.)

There's also the consideration of snaps running on a variety of host distros. "Classic" snaps *could*, if we look around for them, pick up fonts from the host system.

AlanGriffiths: Maybe we should use [`FcFontMatch()`](https://www.freedesktop.org/software/fontconfig/fontconfig-devel/fcfontmatch.html)?

wmww:
Follow up PRs can address issues and make improvements if desired

bors r+

PR #1151, opened 2019-12-16
XWayland refactor and fix

### LINKED ISSUE #1147: XWayland connection created on wrong thread
`XWaylandServer::spawn()` calls `wl_client_create()` from the `"Mir/X11 Reader"` thread, **NOT** the `"Mir/Wayland"` where wayland calls are supposed to be made.

wmww:
Previously, `XWaylandServer::spawn()` forked, and handled both forks in a large switch statement. This splits the two branches into `XWaylandServer::execl_xwayland()` and `XWaylandServer::connect_to_xwayland()`. Also runs `wl_client_create()` on the Wayland thread to fix #1147. It may be helpful to look over the commits individually.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -194,7 +194,19 @@ void mf::XWaylandServer::execl_xwayland(int wl_client_client_fd, int wm_client_f
 
 void mf::XWaylandServer::connect_to_xwayland(int wl_client_server_fd, int wm_server_fd)
 {
-    auto wlclient = wl_client_create(wlc->get_wl_display(), wl_client_server_fd);
+    wl_client* client = nullptr;
+    std::mutex client_mutex;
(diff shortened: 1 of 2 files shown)


wmww:
This doesn't fix all the XWayland threading woes. I don't fully understand the threading model behind the XWayland fronted, and it wouldn't surprise me if no one does. I'll bring this up in the Mir sync tomorrow.

AlanGriffiths:
> This doesn't fix all the XWayland threading woes. I don't fully understand the threading model behind the XWayland fronted, and it wouldn't surprise me if no one does. I'll bring this up in the Mir sync tomorrow.

I'll mention this before the sync so that others can join in:

The xwayland_frontend code was provided by @mariogrip. We knew it had issues when we landed it. (That's why it is flagged as an "experimental" feature.) Merging it has allowed us to keep the code in sync, and incrementally improve it.

The time has come to pay off this technical debt.

PR #1152, opened 2019-12-17
Tidy up x11 frontend

AlanGriffiths:


AlanGriffiths:
I'd been testing all morning without seeing the "strange duplicate window" issue.

Then I pushed the commit. And right away I see it again.

:sigh:

wmww:
It's time "strange duplicate window" had a number: #1153

src/server/frontend_xwayland/xwayland_connector.cpp
```diff
@@ -17,49 +18,40 @@
...
 
 namespace mf = mir::frontend;
 
 mf::XWaylandConnector::XWaylandConnector(
     const int xdisplay,
-    std::shared_ptr<mf::WaylandConnector> wc,
+    std::shared_ptr<WaylandConnector> const& wc,
     std::string const& xwayland_path) :
-    enabled(!!wc->get_extension("x11-support"))
+    start_xwayland{!!wc->get_extension("x11-support") ?
```
wmww: `!!` is unnecessary, as the `?` will force it into a bool anyway

src/server/frontend_xwayland/xwayland_connector.h
```diff
@@ -44,10 +41,10 @@ class XWaylandConnector : public Connector
         std::function<void(std::shared_ptr<scene::Session> const& session)> const& connect_handler) const override;
 
     auto socket_name() const -> optional_value<std::string> override;
+
 private:
-    bool enabled;
-    std::shared_ptr<XWaylandServer> xwayland_server;
-    std::unique_ptr<dispatch::ThreadedDispatcher> xserver_thread;
+    std::function<std::unique_ptr<XWaylandServer>()> const start_xwayland;
+    std::unique_ptr<XWaylandServer> xwayland_server;
```
wmww: Perhaps this ought to be a `std::experimental::optional`?

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -86,52 +98,109 @@ mf::XWaylandServer::~XWaylandServer()
...
+}
+
 void mf::XWaylandServer::spawn()
 {
+    set_thread_name("XWaylandServer::spawn");
+
     enum { server, client, size };
     int wl_client_fd[size], wm_fd[size];
 
+retry_all:
```
wmww: Alan: mumble mumble bad C code
Also Alan: *introduces gotos into an already complex function*

Surely there's a better way?

src/server/frontend_xwayland/xwayland_connector.cpp
AlanGriffiths: OK

src/server/frontend_xwayland/xwayland_connector.h
AlanGriffiths: Why? That would require a definition of a copyable `XWaylandServer`.

src/server/frontend_xwayland/xwayland_server.cpp
AlanGriffiths: You don't agree this is less bad than the recursive calls to `spawn()` that were scattered around? (And didn't work reliably.)

At least the funky logic is all in one place and most of the routine logic separated out. (Thanks for starting that off.)

wmww: Quite possibly a net improvement, but introduces too much new complexity for me to comfortably call all issues pre-existing. Would it be a burden to abstract a single "attempt" into a function and call that in a loop?

AlanGriffiths: How about I split out https://github.com/MirServer/mir/pull/1152/commits/78a9f8f340a18af33ab2ff973606ccf8f051bcd9 and we land that separately?

src/server/frontend_xwayland/xwayland_connector.h
wmww: I mean an optional unique_ptr to indicate that it can be null

src/server/frontend_xwayland/xwayland_server.cpp
wmww: Sounds good

src/server/frontend_xwayland/xwayland_connector.h
wmww: I won't block on this, we can always do it later

Issue #1153, opened 2019-12-18
XWayland "strange duplicate window"


wmww:
Occasionally opening a single XWayland window will result in two identical Mir surfaces. They can be moved independently, but resize as one.

PR #1154, opened 2019-12-18
[X11 frontend] Make Wayland calls on the correct thread

AlanGriffiths:
Make Wayland calls on the correct thread. (Fixes: #1153)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm.cpp
@@ -492,7 +492,7 @@ void mf::XWaylandWM::handle_surface_id(std::shared_ptr<XWaylandWMSurface> surfac
     uint32_t id = event->data.data32[0];
     surface->set_surface_id(id);
 
-    wlc->run_on_wayland_display([=, wlclient=wlclient](auto)
+    wlc->run_on_wayland_display([wlclient=wlclient, id, surface](auto)
         {
             wl_resource* resource = wl_client_get_object(wlclient, id);
             auto* wlsurface = resource ? WlSurface::from(resource) : nullptr;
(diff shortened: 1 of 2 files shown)


src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -340,6 +340,32 @@ void mf::XWaylandWM::handle_events()
...
+    std::condition_variable cv;
+
+    wlc->run_on_wayland_display([&](auto)
+                                    {
+                                        functor();
+
+                                        std::lock_guard<std::mutex> lock{mutex};
+                                        ready = true;
+                                        cv.notify_all();
+                                    });
```
wmww: Formatting nit: excessive indentation

src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -492,10 +518,13 @@ void mf::XWaylandWM::handle_surface_id(std::shared_ptr<XWaylandWMSurface> surfac
...
-    auto *wlsurface = resource ? WlSurface::from(resource) : nullptr;
-    if (wlsurface)
-        surface->set_surface(wlsurface);
+    run_on_wayland_dispatch([&, this]()
+        {
+            wl_resource* resource = wl_client_get_object(wlclient, id);
+            auto* wlsurface = resource ? WlSurface::from(resource) : nullptr;
+            if (wlsurface)
+                surface->set_surface(wlsurface);
+        });
```
wmww: This could lead to excessive blocking of the X11 thread if many windows need to be looked up at once. Is there a reason we can't capture everything we need, jump into the Wayland thread with `wlc->run_on_wayland_display()` and finish up there without blocking X11?

RAOF: Hm. Is there anything part of the system that assumes that the surface_id message is handled before something else happens? If not, it would be perfectly sensible to just post this to the Wayland thread rather than waiting for it to be handled.

PR #1155, opened 2019-12-18
Allow adding a scene observer to the surface stack

wmww:
The primary implementation of `shell::SurfaceStack` (`scene::SurfaceStack`) already implements add/remove observer for `compositor::Scene` and `input::Scene`. This just exposes that through the `SurfaceStack` interface.

### DIFF
--- a/include/server/mir/shell/surface_stack.h
+++ b/include/server/mir/shell/surface_stack.h
@@ -33,6 +33,7 @@ class Surface;
 struct SurfaceCreationParameters;
 class SurfaceObserver;
 class Session;
+class Observer;
 }
 
 namespace shell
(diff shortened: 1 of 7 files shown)


AlanGriffiths:
Why is this useful?

In our MVC architecture the "shell" takes the Controller role, so why does it need to add Observers (which I associate with the View role).

wmww:
I'm implementing `wlr-foreign-toplevel-management`. It lives in the frontend and wants to observe what surfaces are created and destroyed. It could observer the compositor scene or the input scene for the same effect, but that seems conceptually wrong to me. The surface stack is the logically correct thing for it to observe.

AlanGriffiths:
> t could observer the compositor scene or the input scene for the same effect, but that seems conceptually wrong to me. The surface stack is the logically correct thing for it to observe.

I don't see why it is more logical for the frontend to use the scene's interface to the SurfaceStack than the compositor's interface or the input's interface.

It sounds as though the frontend ought to have its own interface?

wmww:
Closing in favor of #1161

PR #1156, opened 2019-12-18
BasicWindowManager requests on unknown windows should noop

wmww:
It is impossible for one subsystem in Mir to know if another subsystem on another thread is in the process of destroying a surface. It is also impossible to externally lock the window manager. Therefore, a surface may be destroyed twice in quick succession, or the window manager may get a `modify_surface` after the surface has been destroyed.

The current behavior is to `fatal_error()` if a surface is destroyed twice, and throw `std::out_of_range` or do nothing in other cases.

This PR implements, documents and tests: Requests on unknown windows warn and noop.

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -100,7 +100,10 @@ void miral::BasicWindowManager::remove_session(std::shared_ptr<scene::Session> c
     auto info = app_info.find(session);
     if (info == app_info.end())
     {
-        log_warning("Removed unknwon session");
+        log_debug(
+            "BasicWindowManager::remove_session() called with unknown or already removed session %s (PID: %d)",
+            session->name().c_str(),
+            session->process_id());
         return;
     }
     policy->advise_delete_app(info->second);
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -180,9 +180,8 @@ void miral::BasicWindowManager::modify_surface(
     shell::SurfaceSpecification const& modifications)
 {
     Locker lock{this};
-    if (!surface_known(surface))
+    if (!surface_known(surface, "modify"))
     {
-        log_warning("Unknown surface modified");
         return;
     }
     auto& info = info_for(surface);
@@ -199,7 +198,10 @@ void miral::BasicWindowManager::remove_surface(
     Locker lock{this};
     if (app_info.find(session) == app_info.end())
     {
-        log_warning("Removed surface with unknwon session");
+        log_debug(
+            "BasicWindowManager::remove_surface() called with unknown or already removed session %s (PID: %d)",
+            session->name().c_str(),
+            session->process_id());
         return;
     }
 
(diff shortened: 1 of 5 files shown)


AlanGriffiths:
What is the problem you're trying to solve?

The only subsystem that should be destroying a surface is the shell (in the guise of the window manager). That ought to be internally locked by `BasicWindowManager`.

wmww:
I'm implementing `wlr-foreign-toplevel-management`. It wants to sent `modify_surface()` and other requests to the shell, but has no way of verifying that the window it's operating on will still be valid when those requests are processed.

wmww:
> as the "controller", the shell could keep track of "nuked" surfaces and consume any associated updates

I don't believe this is technically possible at present time, as the required locking system is not in place.

wmww:
Requests on unknown windows now warn and noop. The warnings will probably generally be ignored, but are good to have so if something starts causing a lot of invalid requests it will get noticed.

src/miral/basic_window_manager.cpp
```diff
@@ -96,7 +97,13 @@ void miral::BasicWindowManager::add_session(std::shared_ptr<scene::Session> cons
 void miral::BasicWindowManager::remove_session(std::shared_ptr<scene::Session> const& session)
 {
     Locker lock{this};
-    policy->advise_delete_app(app_info[session]);
+    auto info = app_info.find(session);
+    if (info == app_info.end())
+    {
+        log_warning("Removed unknwon session");
```
AlanGriffiths: s/unknwon/unknown/

src/miral/basic_window_manager.cpp
```diff
@@ -182,14 +197,16 @@ void miral::BasicWindowManager::remove_surface(
...
     Locker lock{this};
-    try
-    {
-        remove_window(session, info_for(surface));
-    }
-    catch (std::out_of_range const& ex)
+    if (app_info.find(session) == app_info.end())
     {
-        fatal_error("Could not find surface to remove in miral::BasicWindowManager::remove_surface()");
+        log_warning("Removed surface with unknwon session");
```
AlanGriffiths: "unknwon"?

tests/miral/invalid_requests.cpp
```diff
@@ -0,0 +1,175 @@
...
+    std::string name;
+    std::function<Window(InvalidRequests&)> window;
+};
+
+std::ostream& operator<<(std::ostream& ostream, InvalidParam const& param)
+{
+    return ostream << param.name;
+}
+
+struct InvalidRequests : mt::TestWindowManagerTools, WithParamInterface<InvalidParam>
```
AlanGriffiths: Maybe "Invalid" is overly strong wording for requests that are valid, but will be ignored. "Obsolete"?

AlanGriffiths:
@RAOF feel free to land this if it looks OK to you.

src/miral/basic_window_manager.cpp
wmww: I will not mistype unknown
I will not mistype unknown
I will not mistype unknown

AlanGriffiths:
FWIW this fixes the crash mentioned in #1169

PR #1157, opened 2019-12-19
rpi-dispmanx: Update for ShmBuffer changes

RAOF:


src/platforms/rpi-dispmanx/display.cpp
```diff
@@ -18,6 +18,14 @@
  */
 
 #include <bcm_host.h>
+/*
+ * vcos_types.h “helpfully” #defines countof
+ * Sadly, glm also implements glm::countof, with hilarious consequences.
+ *
+ * No Mir code cares about finding the count of a statically sized C-style array,
```
AlanGriffiths: The comment is wrong: grep for `length_of`.

PR #1161, opened 2019-12-19
Add frontend::SurfaceStack

wmww:
Replaces #1155

### DIFF
--- a/include/server/mir/frontend/surface_stack.h
+++ b/include/server/mir/frontend/surface_stack.h
@@ -1,0 +1,48 @@
+/*
+ * Copyright © 2019 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


include/server/mir/frontend/surface_stack.h
```diff
@@ -0,0 +1,48 @@
...
+namespace frontend
+{
+class SurfaceStack
+{
+public:
+    SurfaceStack() = default;
+    virtual ~SurfaceStack() = default;
+
+    virtual void add_observer(std::shared_ptr<scene::Observer> const& observer) = 0;
+    virtual void remove_observer(std::weak_ptr<scene::Observer> const& observer) = 0;
```
AlanGriffiths: I don't think we'll make enough use of this to be worth suggesting that `scene::Observer` is too wide an interface for the frontend requirements.

\<rant>
C++ really doesn't have good support for designs where `frontend::SceneObserver` only offers some of the `scene::Observer` methods.

E.g. `frontend::SceneObserver` could no-op & `final` a bunch of unwanted stuff. But there's no contravariant  parameter passing for `std::shared_ptr<>` that allows it to be passed directly. (That would need an addition forwarding function here to force the conversion.)
\</rant>

PR #1164, opened 2019-12-20
fix mir_performance_tests

AlanGriffiths:
fix mir_performance_tests.

The CompositorPerformance.regression_test_1563287 is failing on master. We don't care what broke it, delete the test.


AlanGriffiths:
> While these tests do rely on the old mirclient, it appears to me we might be better off porting them to Wayland rather than loosing all this code and having to recreate it at some point in the future. I would not be opposed to simply disabling these tests in the mean time.

The code will still be in git. Do we need to keep it compiling?

AlanGriffiths:
Curiosity got the better of me and I bisected:
```plain
e14dd13555737af3291104e72702e248b4230ca8 is the first bad commit
commit e14dd13555737af3291104e72702e248b4230ca8
Author: William Wold <wm@wmww.sh>
Date:   Fri Dec 6 17:20:40 2019 -0500

    BasicWindowManager::modify_window(): respect size constraints
```

AlanGriffiths:
Hmm. `git bisect` is not returning useful results. Probably because the `CompositorPerformance.regression_test_1563287` test sometimes claims to work when it clearly doesn't.

AlanGriffiths:
> We could port them to Wayland, or we could drop in external Wayland demos.

Done

PR #1165, opened 2020-01-06
[packaging] Add libwayland-dev to Build-Depends

RAOF:
Why did we get away with not having this before?


RAOF:
Apparently whatever transitive dependencies we were relying upon have changed in `focal`, so now we need to actually depend on the Wayland development headers!

PR #1170, opened 2020-01-07
Fix clang found clangers

RAOF:
Clang 10 picks up a couple of extra warnings, which have picked up some incorrect Mir code.


tests/miral/resize_window.cpp
```diff
@@ -103,10 +103,10 @@ struct ResizeWindow : mt::TestWindowManagerTools, WithParamInterface<ResizeParam
         Size old_size = window.size();
         Point result = window.top_left();
 
-        if (edge | mir_resize_edge_west)
+        if (edge & mir_resize_edge_west)
```
AlanGriffiths: Interestingly, the test results are unchanged because the same code is used to generate the requested change and the expected change.

PR #1173, opened 2020-01-08
XWayland: create scene surface only when needed

### LINKED ISSUE #1172: XWayland: windows can not choose their initial size
When you open Gedit in Wayland or actual X11, it sets its initial size to whatever its size was last time. When running in XWayland in Mir, it is always the same (I believe 640x480).

wmww:
XWayland: create scene surface only when needed. (Fixes: #1172)

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -429,6 +429,9 @@ mir::shell::SurfaceSpecification& mf::WindowWlSurfaceRole::spec()
 
 void mf::WindowWlSurfaceRole::create_scene_surface()
 {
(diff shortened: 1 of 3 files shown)


wmww:
Hmm... what app(s) did you test with?

AlanGriffiths:
> Hmm... what app(s) did you test with?

xeyes and xterm

wmww:
I can reproduce. I'll figure it out

wmww:
A different approach, but still fixes #1172. If a surface has not had its initial commit when creating an XWayland window from it, wait for that commit before creating the Mir scene surface. If it has already committed, create the Mir scene surface immediately.

wmww:
bors r=AlanGriffiths

PR #1176, opened 2020-01-08
Some initial XWayland cleanup

### LINKED ISSUE #1172: XWayland: windows can not choose their initial size

wmww:
Removes a lot of dead code that we wont need, moves some stuff around, and patches initial window properties closer to where they will be used. This shouldn't have any noticeable effects on behavior ~except for the first commit, which is the same as #1173 and fixes X #1172~


wmww:
#1173 had problems, so this is no longer stacked on that PR

Issue #1180, opened 2020-01-09
master/X11: Segfault when cursor moves off xterm


AlanGriffiths:
1. Start mir_demo_server with X11 enabled: `mir_demo_server --x11-display 3`
2. Start xterm: `DISPLAY=:3 xterm`
3. Put focus on the Mir-on-X window and position cursor over xterm
4. Type a command, e.g. `ls`
5. Move cursor off xterm

Expect: Mir continues to run
Actual: Mir segfaults

AlanGriffiths:
Ah! This is a `PixelSource::read()` that fails to execute the `do_with_pixels` functor.

```c++
`WlShmBuffer::read_internal(...)
...
        else
        {
            mir::log_debug("Wayland buffer destroyed before use; rendering will be incomplete");
        }
```

PR #1182, opened 2020-01-09
Wayland: don't assume PixelSource::read() will execute the do_with_pixels functor

AlanGriffiths:
Wayland: (cursor code) don't assume PixelSource::read() will execute the do_with_pixels functor. (Fixes: #1180)

### DIFF
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -55,13 +55,12 @@ public:
         auto pixel_source = dynamic_cast<mrs::PixelSource*>(buffer.native_buffer_base());
         if (pixel_source)
         {
-            pixel_source->read([&](unsigned char const* buffer_pixels)
+            size_t const buffer_size_bytes = buffer_size.width.as_int() * buffer_size.height.as_int()
+                * MIR_BYTES_PER_PIXEL(buffer.pixel_format());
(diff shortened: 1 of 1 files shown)


src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -357,7 +356,12 @@ void WlSurfaceCursor::apply_latest_buffer()
             auto const cursor_image = std::make_shared<BufferCursorImage>(
                 *stream->lock_compositor_buffer(this),
                 hotspot);
-            surface->set_cursor_image(cursor_image);
+            if (cursor_image->as_argb_8888())
```
wmww: If we're trying to do the thing only if the pixels were copied, perhaps it would be more obvious if we set a bool in the lambda, rather than relying on the logic here to be the same as the logic in `pixel_source->read()`. If there's some other reason for the if, document it.

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -357,7 +356,12 @@ void WlSurfaceCursor::apply_latest_buffer()
...
+                surface->set_cursor_image(cursor_image);
+            else
+            {
+
+            }
```
wmww: This looks awkward. Can you either drop the else block or put something in it (and make braces/lack thereof consistent)?

PR #1185, opened 2020-01-10
RFC Drop cursor code (that achieves nothing)

AlanGriffiths:
Drop cursor code (that achieves nothing)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm.cpp
@@ -115,9 +115,6 @@ void mf::XWaylandWM::start()
 
     xcb_flush(xcb_connection);
 
-//    create_wm_cursor();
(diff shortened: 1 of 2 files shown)


wmww:
Why is this not needed? Is XWayland doing this for us, and thus sending updated cursors along the normal Wayland path?

AlanGriffiths:
> Why is this not needed? Is XWayland doing this for us, and thus sending updated cursors along the normal Wayland path?

I was wondering why we get the wrong cursor in corners and looked at this code. It loads a set of cursors and "sets" one of them once (`CursorLeftPointer`) but none are actually used for anything.

I concluded, as you suggest, that Xwayland is supplying (sometimes incorrect) cursors for X11 windows.

AlanGriffiths:
OK, but it seems overkill to load 9 cursors to change a default I didn't see in testing.

PR #1186, opened 2020-01-10
Make XWaylandWMSurface thread safe

wmww:
Make Wayland calls on the Wayland thread and lock everything else behind a mutext

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm_surface.cpp
@@ -57,7 +57,8 @@ mf::XWaylandWMSurface::XWaylandWMSurface(XWaylandWM *wm, xcb_create_notify_event
           event->parent,
           {event->x, event->y},
           {event->width, event->height},
-          (bool)event->override_redirect}
+          (bool)event->override_redirect},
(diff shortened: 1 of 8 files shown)


src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -266,17 +282,23 @@ void mf::XWaylandWMSurface::read_properties()
...
-        mir::log_warning("XWaylandWMSurface::move_resize() called with unknown detail %d", detail);
+    aquire_shell_surface([detail](auto shell_surface)
+        {
+            if (detail == _NET_WM_MOVERESIZE_MOVE)
+                shell_surface->initiate_interactive_move();
+            else if (auto const edge = wm_resize_edge_to_mir_resize_edge(detail))
+                shell_surface->initiate_interactive_resize(edge.value());
+            else
+                mir::log_warning("XWaylandWMSurface::move_resize() called with unknown detail %d", detail);
+        });
```
AlanGriffiths: Should only the `shell_surface->...` calls be in the `aquire_shell_surface(` functor?

wmww: Yes, done

PR #1187, opened 2020-01-10
[spread] bring RPi into the `ubuntu` task

Saviq:


Saviq:
@RAOF here's my proposal - alternative would be to split out the different bits of the task into scripts and put them together in more granular tasks?

Issue #1188, opened 2020-01-10
XWayland dies when the user VT switches away from Mir


wmww:
For some reason, Mir shuts down XWayland when running on a VT and the user switches away, and restarts it when the user switches back. This is bad because X clients can't connect while the user is not on the Mir VT, and any X clients running are killed.

PR #1191, opened 2020-01-13
X11: Don't stop Xwayland when pausing the Mir server

AlanGriffiths:
We don't need to stop Xwayland when pausing the Mir server. (Fixes: #1188)


AlanGriffiths:
bors r=wmww

PR #1192, opened 2020-01-13
Add top_left to shell::SurfaceSpecification

wmww:
It appears this was never added because no mirclient or Wayland protocols needed it. XWayland will.

### DIFF
--- a/include/server/mir/shell/surface_specification.h
+++ b/include/server/mir/shell/surface_specification.h
@@ -65,6 +65,7 @@ struct SurfaceSpecification
     bool is_empty() const;
     void update_from(SurfaceSpecification const& that);
 
+    optional_value<geometry::Point> top_left;
     optional_value<geometry::Width> width;
     optional_value<geometry::Height> height;
     optional_value<MirPixelFormat> pixel_format;
@@ -96,7 +97,6 @@ struct SurfaceSpecification
     optional_value<std::vector<geometry::Rectangle>> input_shape;
 
     // TODO scene::SurfaceCreationParameters overlaps this content but has additional fields:
-    //    geometry::Point top_left;
     //    input::InputReceptionMode input_mode;
     //
     //    it also has size instead of width + height
--- a/src/miral/window_specification.cpp
+++ b/src/miral/window_specification.cpp
@@ -67,7 +67,7 @@ struct miral::WindowSpecification::Self
 };
 
 miral::WindowSpecification::Self::Self(mir::shell::SurfaceSpecification const& spec) :
-    top_left(),
+    top_left(spec.top_left),
     size(),
     pixel_format(spec.pixel_format),
     buffer_usage(),
--- a/src/server/scene/surface_creation_parameters.cpp
+++ b/src/server/scene/surface_creation_parameters.cpp
@@ -140,6 +140,8 @@ ms::SurfaceCreationParameters& ms::SurfaceCreationParameters::with_buffer_stream
 
 void ms::SurfaceCreationParameters::update_from(msh::SurfaceSpecification const& that)
 {
+    if (that.top_left.is_set())
+        top_left = that.top_left.value();
     if (that.width.is_set() && that.height.is_set())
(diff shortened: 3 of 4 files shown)


RAOF:
Why does XWayland need this? Isn't this something that's properly dealt with in the X window manager layer?

wmww:
X allows windows to access their location and move themselves around. We don't have to support this, but I don't see why we shouldn't. At the time of creating this PR, I was also unsure if it would be needed for popups, but now I know it won't be. It also brings us one property closer to merging `SurfaceSpecification` with `SurfaceCreationParams`.

RAOF:
Oh, we absolutely can and should support this for XWayland, but the point of having the X11 WM built-in to Mir is that we *don't* have to expose this outside the X11 WM bits.

wmww:
To update after IRC convo, I choose this method so Mir's window manager and policy can process the requested move.

btw, the XWayland frontend already has access to the `scene::Surface`

RAOF:
Right.

I think what is left to assuage my concerns is some way of differentiating “the client will break if you ignore the positioning” and “the client has requested a position”. Basically: how are we going to differentiate `override-redirect` windows setting their position from normal windows?

AlanGriffiths:
I'm pretty sure adding and setting this will not achieve anything without other changes. It will simply be overwritten in, for example, `CanonicalWindowManagerPolicy::handle_place_new_surface()`.

That said, I agree that the X11 window management does need a way to set window positions, but I need convincing that adding a "top_left" option is the clearest way to do it.

> To update after IRC convo, I choose this method so Mir's window manager and policy can process the requested move.

Being precise, this is an initial placement, not a move.

> btw, the XWayland frontend already has access to the `scene::Surface`

True, that is an existing, minor, issue. The X11 frontend is an MVC "View", so it should be treating Surface as read-only (ideally it would be using an interface or proxy without modifier methods to prevent accidents).

wmww:
> Being precise, this is an initial placement, not a move.

`SurfaceCreationParamaters` is used for initial placement (it already has a `top_left`, which is seemingly ignored). `SurfaceSpecification` (what this PR modifies) is used for surface modification.

AlanGriffiths:
> > Being precise, this is an initial placement, not a move.
> 
> `SurfaceCreationParamaters` is used for initial placement (it already has a `top_left`, which is seemingly ignored). `SurfaceSpecification` (what this PR modifies) is used for surface modification.

You are correct. I misread.

wmww:
> Naturally, any existing window management code is likely to ignore this if it is set, so we may find problems once we start setting it.

I haven't tested it yet, but...
* `WindowSpecification(mir::shell::SurfaceSpecification const&)` (well technically `WindowSpecification::Self::Self`) is updated to accept `top_left`.
* `BasicWindowManager::modify_surface()` uses `WindowSpecification(mir::shell::SurfaceSpecification const&)` to create the `WindowSpecification` sent to the policy
* Unless the policy explicitly clears it, the same `top_left` value is sent back in the `WindowSpecification` that is a parameter to `BasicWindowManager::modify_window()`
* `BasicWindowManager::modify_window()` already respects `top_left`.
Therefore, I would not be suprised if it just worked™.

RAOF:
> What is not clear is, as RAOF says, it is currently unclear whether there's a difference between our X11WM code wanting to position things and requests from other hypothetical sources. But we don't need to solve that yet.

Specifically, there are *two* different X11 sources here - there are windows setting their position via WM interaction, and windows with the `override-redirect` flag (ie: “don't mess with me, WM”) setting their position. It's correct for the former to go through the Mir WM machinery and possibly modified, but modifying window positions/sizes from `override-redirect` windows is likely to result in incorrect behaviour.

wmww:
In the case of popups when a "transient for" window is set, Mir's popup mechanism will be used and Mir's window manager *shouldn't* mess with those windows. I'm not sure that there will be other cases in which an app depends on being able to position its windows.

wmww:
bors r=RAOF

PR #1193, opened 2020-01-13
Mir release 1.7.0

AlanGriffiths:
Mir release 1.7.0


AlanGriffiths:
@wmww @RAOF please indicate whether you're happy with this collection of cherry-picks.

AlanGriffiths:
> Everything looks good to me. Unless someone is waiting on this release, I suggest waiting another week or so. A bunch of XWayland improvements are still on the way, which will have a meaningful impact on users.

You said the same last week, you will likely say the same next week and the week after. ;)

As a general principle, I don't like to hold up releases for features (just blocking bugs): there will always be a new release for those in a few more weeks and we don't need to hurry them into this release.

The X11 support is still "experimental", and this release doesn't introduce X11 support bugs that were not in 1.6. So I don't think fixing those is a "blocker". The only thing that would prevent X11 bugfixes subsequently being released as 1.7.1 would be an ABI change (like #1192).

But there's no immediate hurry to tag the release, there's at least one more commit I want to cherry-pick. We can wait until after our call to agree the way forward.

wmww:
> You said the same last week, you will likely say the same next week and the week after. ;)

Fair enough. I don't think this should be the last release with mirclient support (since features are on the way that don't require removing it). If there will be another, then releasing this is fine.

AlanGriffiths:
## Test plan: https://wiki.ubuntu.com/Process/Merges/TestPlans/Mir

Target|Tested| Status| Comments
------------ | -------------|---|--
18.04LTS|||
16.04LTS|||
20.04LTS|||
19.10|||
Nvidia|||

### 1. Qt app menus appear in weird places.

This seems to be a Qt issue as it can be reproduced with `QT_QPA_PLATFORM=wayland <app>` on Gnome DE.

## mir-test-tools

Target|Tested| Status| Comments
------------ | -------------|---|--
18.04LTS|||
16.04LTS|||
19.10|||
Core18 (AMD64)|||
Core18 (armhf)|||

## mir-kiosk
Target|Tested| Status| Comments
------------ | -------------|---|--
18.04LTS|||
16.04LTS|||
19.10|||
Core18 (AMD64)|||
Core18 (armhf)|||

AlanGriffiths:
bors r=wmww,RAOF

AlanGriffiths:
bors r=wmww,RAOF

PR #1195, opened 2020-01-14
X11 is no longer "experimental"

AlanGriffiths:
X11 is no longer "experimental"


AlanGriffiths:
We're not quite ready to land this yet.

src/platform/options/default_configuration.cpp
```diff
@@ -50,7 +50,7 @@ char const* const mo::fatal_except_opt            = "on-fatal-error-except";
 char const* const mo::debug_opt                   = "debug";
 char const* const mo::composite_delay_opt         = "composite-delay";
 char const* const mo::enable_key_repeat_opt       = "enable-key-repeat";
-char const* const mo::x11_display_opt             = "x11-display-experimental";
+char const* const mo::x11_display_opt             = "x11-display";
```
wmww: This has the potential to be confused with the X11 display the X backend is running on. Perhaps `--xwayland-display` would be less error-prone? I'm aware other things are already expecting it to change to `--x11-display`, but if we're ever going to change the option it should be now.

AlanGriffiths: Maybe I'm just conditioned to not seeing this as a problem, but...

1. I don't see that `--xwayland-display` would be less confusing. There are plenty of desktops where there's already an Xwayland instance (or two!) running.

2. Xwayland is simply a detail of how we provide an X11 `DISPLAY`, and not the abstraction being configured. (We could, hypothetically, use Xmir to provide X11 support.)

wmww: Both your points are correct from a technical standpoint, but I still believe `--x11-display` is less obvious (and thus a worse user interface).

1. In practice, an `--xwayland-display` option related to the Mir-on-X backend wouldn't make sense. I *think* users will generally realize this, and not be confused. On the other hand, an `--x11-display` option for the X11 backend would absolutely be reasonable.

2. XWayland may be an implementation detail, but it **is** the only thing we will use for the foreseeable future and it's something many people are familiar with.

Open to other options, but I think we can do better than `--x11-display`.

AlanGriffiths: On IRC @RAOF points out that Xwayland should pick the display itself.

That would make the option `--enable-x11` (which is consistent with other `--enable-*` options we have).

I need to test this will work in snap confinement, but this PR will be back.

AlanGriffiths:
OK, this is enough to resubmit.

A follow-up will provide a way to write to supply an FD to receive the display. (Like Xwayland's -displayfd.)

Maybe we should provide a way to select the socket from the command-line, but that's less urgent.

PR #1198, opened 2020-01-14
XWayland: improve verbose logging

wmww:
Improvements to the verbose logs generated by the XWayland frontend. Probably not interesting to anyone but me, but merging will make maintaining my branches simpler.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm.cpp
@@ -79,10 +79,10 @@ namespace mf = mir::frontend;
 namespace
 {
 template<typename T>
-auto c_array_to_string(T* data, size_t size) -> std::string
+auto data_buffer_to_debug_string(T* data, size_t elements) -> std::string
 {
     std::string result = "[";
-    for (unsigned i = 0; i < size; i++)
+    for (unsigned i = 0; i < elements; i++)
     {
         if (i)
             result += ", ";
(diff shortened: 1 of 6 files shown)


src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -76,6 +76,23 @@ static const struct cursor_alternatives cursors[] = {
...
+auto c_array_to_string(T* data, size_t size) -> std::string
+{
+    std::string result = "[";
+    for (unsigned i = 0; i < size; i++)
+    {
+        if (i)
+            result += ", ";
+        result += std::to_string(*(data + i));
+    }
+    result += "]";
```
AlanGriffiths: If you're going to index into the "array", use `data[i]`.

If you're going to dereference pointers, use pointers! (`for (T* p = data; p != data+size; ++p)`)

src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -76,6 +76,23 @@ static const struct cursor_alternatives cursors[] = {
 
 namespace mf = mir::frontend;
 
+namespace
+{
+template<typename T>
+auto c_array_to_string(T* data, size_t size) -> std::string
```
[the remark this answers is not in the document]
wmww: Fixed by renaming function to reflect what it actually is

src/server/frontend_xwayland/xwayland_wm.cpp
wmww: Done

src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -76,6 +76,23 @@ static const struct cursor_alternatives cursors[] = {
 
 namespace mf = mir::frontend;
 
+namespace
+{
+template<typename T>
+auto data_buffer_to_debug_string(T* data, size_t elements) -> std::string
+{
+    std::string result = "[";
```
RAOF: I know this is only debug code, but this would be better as a `std::stringstream` rather than successive allocation / strcat.

PR #1199, opened 2020-01-14
Miral: add parent info to window management trace

wmww:
Does what it says. This info is missing and useful.

### DIFF
--- a/src/miral/window_management_trace.cpp
+++ b/src/miral/window_management_trace.cpp
@@ -93,6 +93,11 @@ struct BracedItemStream
         out.fill(fill);
         return *this;
     }
+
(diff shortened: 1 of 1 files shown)


src/miral/window_management_trace.cpp
```diff
@@ -136,6 +141,7 @@ auto dump_of(miral::WindowInfo const& info) -> std::string
         APPEND(name);
         APPEND(type);
         APPEND(state);
+        APPEND(parent);
```
AlanGriffiths: I'd be inclined to only output this if not null to reduce spew.

wmww:
Upon further investigation I found that parent info was already being output. I just didn't see it because it was further down the list, and nothing showed up if the window didn't have a parent. I may idly think on how window management trace could be made to be more readable, but nothing need be done now.

PR #1200, opened 2020-01-15
XWayland: replace C-style code with safer C++ alternatives

wmww:
Mostly replaces C arrays with vectors and adds const. Should improve readability and reduce future errors.


src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -134,21 +136,28 @@ void mf::XWaylandWM::start()
     wm_get_resources();
     setup_visual_and_colormap();
 
-    values[0] =
-        XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT | XCB_EVENT_MASK_PROPERTY_CHANGE;
-    xcb_change_window_attributes(xcb_connection, xcb_screen->root, XCB_CW_EVENT_MASK, values);
+    std::vector<uint32_t> const attrib_values{
```
RAOF: Is there a particular reason that you're using a single-element `Vec<u32>` here, rather than just `uint32_t const attrib_value = XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT | XCB_EVENT_MASK_PROPERTY_CHANGE` (and then passing `&attrib_value` rather than `attrib_values.data()`)?

src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -134,21 +136,28 @@ void mf::XWaylandWM::start()
...
-    supported[0] = xcb_atom.net_wm_moveresize;
-    supported[1] = xcb_atom.net_wm_state;
-    supported[2] = xcb_atom.net_wm_state_fullscreen;
-    supported[3] = xcb_atom.net_wm_state_maximized_vert;
-    supported[4] = xcb_atom.net_wm_state_maximized_horz;
-    supported[5] = xcb_atom.net_active_window;
-    xcb_change_property(xcb_connection, XCB_PROP_MODE_REPLACE, xcb_screen->root, xcb_atom.net_supported, XCB_ATOM_ATOM,
-                        32, /* format */
-                        ARRAY_LENGTH(supported), supported);
+    std::vector<xcb_atom_t> const supported{
```
RAOF: When we can switch on C++17 this should be `std::array const supported{…};`.

AlanGriffiths: Surely std::array is C++11? We use it lots.

src/server/frontend_xwayland/xwayland_wm.cpp
AlanGriffiths: Personally, I think "construction == initialization" is sufficient here:
```c++
    uint32_t const values[] = {
        XCB_EVENT_MASK_SUBSTRUCTURE_NOTIFY | XCB_EVENT_MASK_SUBSTRUCTURE_REDIRECT | XCB_EVENT_MASK_PROPERTY_CHANGE
        };
```

src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -214,8 +236,6 @@ void mf::XWaylandWM::create_wm_cursor()
 
 void mf::XWaylandWM::create_wm_window()
 {
-    static const char name[] = "Mir XWM";
```
AlanGriffiths: I see no need to increase the scope of this variable beyond the function.

src/server/frontend_xwayland/xwayland_wm.cpp
wmww: There are probably lighter alternatives to `std::vector`, but I want to make it obvious to anyone reading that more elements could be added

wmww:
> Real arrays (initialized with `...[] = { ...};`) or `std::array` (initialized with `...{ ...};`) should be preferred.

@AlanGriffiths both have problems.

As far as I can tell, extracting the size out of a "real" array requires `sizeof(foo)/sizeof(foo[0]);` or a template function. The former is ugly and error prone and the latter requires too much code overhead for just a few uses.

`std::array` doesn't deduce size out of initialization list. If specified size is too small it errors, but if too large *it still compiles*.

IMO a few possibly avoidable allocations in code that isn't especially performance critical is a small price to pay for a straightforward solution that makes correctness obvious.

AlanGriffiths:
> As far as I can tell, extracting the size out of a "real" array requires `sizeof(foo)/sizeof(foo[0]);` or a template function. The former is ugly and error prone and the latter requires too much code overhead for just a few uses.

1. the `length_of()` function is already used in this file.
2. `std::distance(std::begin(array), std::end(array));`

> `std::array` doesn't deduce size out of initialization list. If specified size is too small it errors, but if too large _it still compiles_.

:(

wmww:
There's also an instance of sending a fixed length list in `xwayland_wm_surface.cpp`. Shall I move `length_of` to a header? Also, if it's so useful, shouldn't `length_of` be in a standard Mir header?

src/server/frontend_xwayland/xwayland_wm.cpp
RAOF: `std::array` is C++11. The template deduction guide which lets you go `std::array const supported{…}` is C++17.

We *could* do `std::array<xcb_atom_t, 6> const supported{…}` right now.

AlanGriffiths:
> Shall I move `length_of` to a header? Also, if it's so useful, shouldn't `length_of` be in a standard Mir header?

I guess we can move it to a header, but I don't think it is a function Mir should provide for downstream. Its use is really for interacting with C-style APIs.

----

It is sorely tempting to start introducing stuff like:
```C++
template<typename T, uint32_t  data_len>
xcb_void_cookie_t
xcb_change_property(xcb_connection_t *c,
                     uint8_t          mode,
                     xcb_window_t     window,
                     xcb_atom_t       property,
                     xcb_atom_t       type,
                     uint8_t          format,
                     T(data&)[data_len])
{
    return xcb_change_property (c, mode, window, property, type, format, data_len, data);
}
```

wmww:
Oh, turns out I wasn't using `.size()` with a const vector in another file after all.

wmww:
@RAOF replaced unnecessary vectors with arrays, but still not using using single values so as not to imply that that's all the functions would accept.

src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -588,38 +605,27 @@ void mf::XWaylandWM::handle_configure_request(xcb_configure_request_event_t *eve
...
-
-    uint32_t values[6];
-    int i = -1;
+    std::vector<uint32_t> values;
 
     if (event->value_mask & XCB_CONFIG_WINDOW_X)
-    {
-        values[++i] = event->x;
-    }
+        values.push_back(event->x);
```
RAOF: I mildly prefer always having `{ }` around the bodies of `if`, even if one-line.

I note that we don't have anything about this in the style guide. It's not a huge problem, though (gcc will warn if you do it wrong, mostly).

wmww:
bors r=RAOF

PR #1205, opened 2020-01-17
Better execl arguments when executing Xwayland

AlanGriffiths:
Better execl arguments when executing Xwayland.

Fixes the results from ps etc.


AlanGriffiths:
bors r=RAOF

PR #1207, opened 2020-01-21
More minor XWayland improvements

wmww:
Notably moves processing of a single event into its own function and catches exceptions while processing X11 events. Wrote code last Thursday, but required PRs hadn't landed yet

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm.cpp
@@ -303,9 +303,13 @@ void mf::XWaylandWM::handle_events()
         {
             handle_event(event);
         }
-        catch (std::exception const& e)
+        catch (...)
         {
-            log_error(e.what());
+            log(
+                logging::Severity::warning,
+                MIR_LOG_COMPONENT,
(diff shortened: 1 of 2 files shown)


src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -298,7 +298,14 @@ void mf::XWaylandWM::handle_events()
...
     while (xcb_generic_event_t* const event = xcb_poll_for_event(xcb_connection))
     {
-        handle_event(event);
+        try
+        {
+            handle_event(event);
+        }
+        catch (std::exception const& e)
+        {
+            log_error(e.what());
```
AlanGriffiths: It seems wrong to log an "error" when we're going to carry on by ignoring it.

How about:
```c++
        catch (...)
        {
            log(logging::Severity::warning,
                MIR_LOG_COMPONENT,
                std::current_exception(),
                "Failed to handle xcb event.");
        }
```

wmww: Sure, done

PR #1208, opened 2020-01-21
XWaylandWM::get_wm_surface()

wmww:
More code left over from Thursday. Adds a new `XWaylandWM::get_wm_surface()` method. This makes converting from XCB windows to wm surfaces more convenient and publicly accessible (the latter will be important when shell surfaces want to look up their parent windows). This also fixes the issue of accidentally creating null shell surfaces in the surfaces map via careless usage of the '[]` map operator.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm.cpp
@@ -286,6 +286,15 @@ auto mf::XWaylandWM::build_shell_surface(
     return shell->build_shell_surface(wm_surface, wayland_client, wayland_surface);
 }
 
+auto mf::XWaylandWM::get_wm_surface(xcb_window_t xcb_window) -> std::experimental::optional<std::shared_ptr<XWaylandWMSurface>>
+{
+    auto const surface = surfaces.find(xcb_window);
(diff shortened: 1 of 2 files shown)


src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -536,18 +535,17 @@ void mf::XWaylandWM::handle_client_message(xcb_client_message_event_t *event)
...
-    auto const surface = surface_iter->second;
+    auto const surface = get_wm_surface(event->window);
 
-    if (event->type == xcb_atom.net_wm_moveresize)
-        handle_move_resize(surface, event);
-    else if (event->type == xcb_atom.net_wm_state)
-        handle_state(surface, event);
-    else if (event->type == xcb_atom.wl_surface_id)
-        handle_surface_id(surface, event);
+    if (surface)
```
AlanGriffiths: Can't we reduce the scope of surface to the `if`? Vis:
```c++
    if (auto const surface = get_wm_surface(event->window))
```

wmww: Good idea. Done

PR #1211, opened 2020-01-21
XWayland: manage window state correctly

### LINKED ISSUE #1209: Restored windows do not get resized properly
Repro:
1. Open Gedit in XWayland
2. Maximize 
3. Unmaximize

Expected: window becomes restored and the size it was before the maximize

Actual: window becomes restored, but keeps the maximized size an position

### LINKED ISSUE #1210: XWayland: windows can not request a change in state
The maximize/minimize controls on Gedit don't work on XWayland (but SSDs do)

wmww:
Fixes #1210. Also alerts windows to their current state (for example, Gedit draws slightly differently when it knows it's maximized).

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm_surface.cpp
@@ -29,11 +29,20 @@ namespace mf = mir::frontend;
 
 namespace
 {
+/// See ICCCM 4.1.3.1 (https://tronche.com/gui/x/icccm/sec-4.html)
 enum class WmState: uint32_t
 {
     WITHDRAWN = 0,
     NORMAL = 1,
     ICONIC = 3,
+};
+
+/// See https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#sourceindication
+enum class SourceIndication: uint32_t
+{
+    UNKNOWN = 0,
+    APPLICATION = 1,
+    PAGER = 2,
 };
 
 auto wm_resize_edge_to_mir_resize_edge(uint32_t wm_resize_edge) -> std::experimental::optional<MirResizeEdge>
@@ -83,6 +92,60 @@ mf::XWaylandWMSurface::~XWaylandWMSurface()
         {
             delete shell_surface;
         });
+}
+
+void mf::XWaylandWMSurface::net_wm_state_client_message(uint32_t (&data)[5])
+{
+    // The client is requesting a change in state
+    // see https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html#idm45390969565536
+
+    enum class NetWmStateClientMessageAction: uint32_t
+    {
+        REMOVE = 0,
+        ADD = 1,
+        TOGGLE = 2,
+    };
+
+    struct NetWmStateClientMessage
+    {
+        NetWmStateClientMessageAction const action;
+        xcb_atom_t const properties[2];
+        SourceIndication const source_indication;
+    };
+
+    auto const message = (NetWmStateClientMessage*)data;
+
+    WindowState new_window_state;
+
+    {
+        std::lock_guard<std::mutex> lock{mutex};
+
+        new_window_state = window_state;
+
+        for (size_t i = 0; i < length_of(message->properties); i++)
+        {
+            if (xcb_atom_t const property = message->properties[i]) // if there is only one property, the 2nd is 0
+            {
+                bool nil{false}, *prop_ptr = &nil;
+
+                if (property == xwm->xcb_atom.net_wm_state_hidden)
+                    prop_ptr = &new_window_state.minimized;
+                else if (property == xwm->xcb_atom.net_wm_state_maximized_horz) // assume vert is also set
+                    prop_ptr = &new_window_state.maximized;
+                else if (property == xwm->xcb_atom.net_wm_state_fullscreen)
+                    prop_ptr = &new_window_state.fullscreen;
+
+                switch (message->action)
+                {
+                case NetWmStateClientMessageAction::REMOVE: *prop_ptr = false; break;
+                case NetWmStateClientMessageAction::ADD: *prop_ptr = true; break;
+                case NetWmStateClientMessageAction::TOGGLE: *prop_ptr = !*prop_ptr; break;
+                }
+            }
+        }
+    }
+
+    set_window_state(new_window_state);
 }
 
 void mf::XWaylandWMSurface::dirty_properties()
--- a/src/server/frontend_xwayland/xwayland_wm_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm_surface.cpp
@@ -176,7 +176,6 @@ void mf::XWaylandWMSurface::read_properties()
     props[XCB_ATOM_WM_TRANSIENT_FOR] = XCB_ATOM_WINDOW;
     props[xwm->xcb_atom.wm_protocols] = TYPE_WM_PROTOCOLS;
     props[xwm->xcb_atom.wm_normal_hints] = TYPE_WM_NORMAL_HINTS;
-    props[xwm->xcb_atom.net_wm_state] = TYPE_NET_WM_STATE;
     props[xwm->xcb_atom.net_wm_window_type] = XCB_ATOM_ATOM;
     props[xwm->xcb_atom.net_wm_name] = XCB_ATOM_STRING;
     props[xwm->xcb_atom.motif_wm_hints] = TYPE_MOTIF_WM_HINTS;
@@ -244,27 +243,6 @@ void mf::XWaylandWMSurface::read_properties()
         }
         case TYPE_WM_NORMAL_HINTS:
         {
-            break;
-        }
-        case TYPE_NET_WM_STATE:
-        {
-            xcb_atom_t *value = reinterpret_cast<xcb_atom_t *>(xcb_get_property_value(reply));
-            uint32_t i;
-            for (i = 0; i < reply->value_len; i++)
-            {
-                if (value[i] == xwm->xcb_atom.net_wm_state_fullscreen && !fullscreen)
-                {
-                    fullscreen = true;
-                }
-            }
-            if (value[i] == xwm->xcb_atom.net_wm_state_maximized_horz && !maximized)
-            {
-                maximized = true;
-            }
-            if (value[i] == xwm->xcb_atom.net_wm_state_maximized_vert && !maximized)
-            {
-                maximized = true;
-            }
             break;
         }
         case TYPE_MOTIF_WM_HINTS:
--- a/src/server/frontend_xwayland/xwayland_wm_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm_surface.cpp
@@ -106,14 +106,10 @@ void mf::XWaylandWMSurface::net_wm_state_client_message(uint32_t const (&data)[5
         TOGGLE = 2,
     };
 
-    int i = 0;
-    auto const action = static_cast<Action>(data[i]);
-    i++;
-    auto const prop_a = static_cast<xcb_atom_t>(data[i]);
-    i++;
-    auto const prop_b = static_cast<xcb_atom_t>(data[i]);
-    i++;
-    auto const source_indication = static_cast<SourceIndication>(data[i]);
+    auto const* pdata = data;
+    auto const action = static_cast<Action>(*pdata++);
+    xcb_atom_t const properties[2] = { static_cast<xcb_atom_t>(*pdata++),  static_cast<xcb_atom_t>(*pdata++) };
+    auto const source_indication = static_cast<SourceIndication>(*pdata++);
 
     (void)source_indication;
 
@@ -124,7 +120,7 @@ void mf::XWaylandWMSurface::net_wm_state_client_message(uint32_t const (&data)[5
 
         new_window_state = window_state;
 
-        for (xcb_atom_t const property : {prop_a, prop_b})
+        for (xcb_atom_t const property : properties)
         {
             if (property) // if there is only one property, the 2nd is 0
             {
--- a/src/server/frontend_xwayland/xwayland_wm_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm_surface.cpp
@@ -221,12 +221,13 @@ void mf::XWaylandWMSurface::set_workspace(int workspace)
 
 void mf::XWaylandWMSurface::apply_mir_state_to_window(MirWindowState new_state)
 {
-    std::vector<xcb_atom_t> net_wm_states;
-
     WindowState new_window_state;
 
     {
         std::lock_guard<std::mutex> lock{mutex};
+
+        if (new_state == cached_mir_window_state)
+            return;
 
         cached_mir_window_state = new_state;
         new_window_state = window_state;
--- a/src/server/frontend_xwayland/xwayland_wm_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm_surface.cpp
@@ -305,3 +305,75 @@ void mf::XWaylandWMSurface::aquire_shell_surface(std::function<void(XWaylandWMSh
                 work(shell_surface);
         });
 }
+
+void mf::XWaylandWMSurface::set_window_state(WindowState const& new_window_state)
+{
+    WmState const wm_state{
+        new_window_state.minimized ?
+        WmState::ICONIC :
+        WmState::NORMAL};
+
+    uint32_t const wm_state_properties[]{
+        static_cast<uint32_t>(wm_state),
+        XCB_WINDOW_NONE // Icon window
+    };
+
+    xcb_change_property(
+        xwm->get_xcb_connection(),
+        XCB_PROP_MODE_REPLACE,
+        window, xwm->xcb_atom.wm_state,
+        xwm->xcb_atom.wm_state, 32,
+        length_of(wm_state_properties), wm_state_properties);
+
+    std::vector<xcb_atom_t> net_wm_states;
+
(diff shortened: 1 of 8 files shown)


src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -118,46 +219,76 @@ void mf::XWaylandWMSurface::set_workspace(int workspace)
...
+            // don't change new_window_state.maximized
+            new_window_state.fullscreen = true;
+            break;
+
+        case mir_window_state_maximized:
+        case mir_window_state_vertmaximized:
+        case mir_window_state_horizmaximized:
+        case mir_window_state_attached:
+            new_window_state.minimized = false;
+            new_window_state.maximized = true;
```
RAOF: I think you might need to split this into vert & horiz maximised state; there's a reasonable amount of stuff that does one but not the other?

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -335,3 +436,75 @@ void mf::XWaylandWMSurface::aquire_shell_surface(std::function<void(XWaylandWMSh
...
+    xcb_change_property(
+        xwm->get_xcb_connection(),
+        XCB_PROP_MODE_REPLACE,
+        window, xwm->xcb_atom.wm_state,
+        xwm->xcb_atom.wm_state, 32,
+        length_of(wm_state_properties), wm_state_properties);
+
+    std::vector<xcb_atom_t> net_wm_states;
+
+    if (new_window_state.maximized)
```
RAOF: Do you mean `new_window_state.minimized` here? :confused:

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -78,6 +94,91 @@ mf::XWaylandWMSurface::~XWaylandWMSurface()
...
+    };
+
+    struct NetWmStateClientMessage
+    {
+        NetWmStateClientMessageAction const action;
+        xcb_atom_t const properties[2];
+        SourceIndication const source_indication;
+    };
+
+    auto const message = (NetWmStateClientMessage*)data;
```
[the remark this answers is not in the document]
AlanGriffiths: As all of the members are effectively `uint32_t` I think we can argue away the alignment and padding concerns. But I don't think there's standardese to support "defined behaviour", and it is certainly fragile code.

Would it be clearer to use named indexes into `data`? Vis:
```C++
        for (size_t i = wmstateclientmessage_properties_begin; i != wmstateclientmessage_properties_end; ++i)
        {
            if (xcb_atom_t const property = data[i]) // if there is only one property, the 2nd is 0
...
                switch (data[wmstateclientmessage_action])
```

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -78,6 +94,91 @@ mf::XWaylandWMSurface::~XWaylandWMSurface()
...
+    }
+
+    set_window_state(new_window_state);
+}
+
+void mf::XWaylandWMSurface::wm_change_state_client_message(uint32_t (&data)[5])
+{
+    // See ICCCM 4.1.4 (https://tronche.com/gui/x/icccm/sec-4.html)
+
+    WmState requested_state = (WmState)data[0];
```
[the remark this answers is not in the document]
AlanGriffiths: It can. Also there's no need to make `requested_state` mutable...
```c++
    auto const requested_state = static_cast<WmState>(data[0]);
```

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -78,6 +94,91 @@ mf::XWaylandWMSurface::~XWaylandWMSurface()
...
+
+        for (size_t i = 0; i < length_of(message->properties); i++)
+        {
+            if (xcb_atom_t const property = message->properties[i]) // if there is only one property, the 2nd is 0
+            {
+                bool nil{false}, *prop_ptr = &nil;
+
+                if (property == xwm->xcb_atom.net_wm_state_hidden)
+                    prop_ptr = &new_window_state.minimized;
+                else if (property == xwm->xcb_atom.net_wm_state_maximized_horz) // assume vert is also set
```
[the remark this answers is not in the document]
wmww: No application I've tested with requests vertical-only or horizontal-only maximization, and supporting it makes this section of code significantly more complicated. If there is an example of a prominent app that makes such a request I will reconsider.

src/server/frontend_xwayland/xwayland_wm_surface.cpp
wmww: yes

src/server/frontend_xwayland/xwayland_wm_surface.cpp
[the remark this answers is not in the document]
wmww: Done

src/server/frontend_xwayland/xwayland_wm_surface.cpp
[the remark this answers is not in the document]
AlanGriffiths: I don't think the assumption holds, but it is usually the shell that imposes semi-maximized state. OTOH I would be willing to land the PR without addressing this as this is clearly an improvement over the existing logic.

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -111,9 +111,12 @@ void mf::XWaylandWMSurface::net_wm_state_client_message(uint32_t (&data)[5])
         NetWmStateClientMessageAction const action;
         xcb_atom_t const properties[2];
         SourceIndication const source_indication;
+        uint32_t const pad;
     };
 
-    auto const message = (NetWmStateClientMessage*)data;
+    static_assert(sizeof(NetWmStateClientMessage) == sizeof(data), "Structure size incorrect");
+
+    auto const message = reinterpret_cast<NetWmStateClientMessage const*>(data);
```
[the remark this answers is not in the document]
wmww: Fine, I'll do it good

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -29,6 +29,27 @@ namespace mf = mir::frontend;
...
+        if (i >= 5)
+            mir::fatal_error("ClientMessageData::unpack() called too many times");
+        auto prev = i;
+        i++;
+        return static_cast<T>(data[prev]);
+    }
+
+    uint32_t const* const data;
+    size_t i{0};
+};
```
AlanGriffiths: Seems a needless abstraction that makes the code harder to parse.

wmww: My requirements are:
- Layout and types for a message defined in one place: all the code that maps the protocol to strongly typed data should be on a few lines, in contrast to spread all over the function.
- No hard-coded indices: these are extremely easy to screw up, and the resulting errors may be difficult to detect
- Doesn't rely on using the correct variant of `i++` vs `++i`: IMO this is almost never good code
- No UB

Do you have a better alternative that satisfies all 4?

RAOF:
Before we get into the weeds on specific code, have we answered the top level question? Why does this code need to exist? Why doesn't it *already* work?

AlanGriffiths:
> Before we get into the weeds on specific code, have we answered the top level question? Why does this code need to exist?

It doesn't surprise me that some code is needed to manage window states. With Xmir we embedded an X11 WM into the Xmir executable and that translated window states, but AIUI Xwayland doesn't attempt that, and the X11 WM code is in our frontend.

> Why doesn't it _already_ work?

One of the issues this purports to fix (#1209) does already work. :wink:

src/server/frontend_xwayland/xwayland_wm_surface.h
```diff
@@ -122,38 +123,53 @@ class XWaylandWMShellSurface;
...
     void send_resize(const geometry::Size& new_size);
     void send_close_request();
 
 private:
     /// Runs work on the Wayland thread if the shell surface hasn't been destroyed
     void aquire_shell_surface(std::function<void(XWaylandWMShellSurface* shell_surface)>&& work);
 
+    /// contains more information than just a MirWindowState
+    /// (for example if a minimized window would otherwise be maximized)
+    struct WindowState
```
RAOF: Why does this exist? This information *should* be stored in Mir's internal representation, because this is a core WM concern, not something specific to X11.

RAOF:
> > Before we get into the weeds on specific code, have we answered the top level question? Why does this code need to exist?
> 
> It doesn't surprise me that some code is needed to manage window states. With Xmir we embedded an X11 WM into the Xmir executable and that translated window states, but AIUI Xwayland doesn't attempt that, and the X11 WM code is in our frontend.

Oh, yes. Quite true. My actual question is the more narrow question of why the `WindowState` mechanism exists here. There absolutely does need to be code in the X11 WM to translate the between X11 and Mir window states.

src/server/frontend_xwayland/xwayland_wm_surface.h
AlanGriffiths: > This information _should_ be stored in Mir's internal representation, because this is a core WM concern, not something specific to X11.

Agreed. But maybe, pending more extensive changes to the "model", storing it in the "view" works well enough?

(We've been holding off reworking the "model" until we clean out mirclient dependencies.)

src/server/frontend_xwayland/xwayland_wm_surface.cpp
AlanGriffiths: Well, I don't necessarily agree your requirements.

E.g. I'm not keen on `unpack<>()` having side effects so that the order of calls matters, nor having to look outside the function to see WTF `values.unpack<Action>()` does. A `static_cast<>` is a well known thing with understood risks.

It's just as easy to screw up the the order of unpacking as it is indexes. So I'll just throw in:

```c++
    struct NetWmStateClientMessage
    {
        NetWmStateClientMessage(uint32_t const (&data)[5]) :
            action{static_cast<NetWmStateClientMessageAction>(data[0])},
            properties{static_cast<xcb_atom_t>(               data[1]), 
                       static_cast<xcb_atom_t>(               data[2])},
            source_indication{static_cast<SourceIndication>(  data[3])},
            pad{static_cast<decltype(pad)>(                   data[4])}
        {
            static_assert(sizeof(NetWmStateClientMessage) == sizeof(data), "Structure size incorrect");
        }

        NetWmStateClientMessageAction const action;
        std::array<xcb_atom_t, 2> const properties;
        SourceIndication const source_indication;
        uint32_t const pad;
    } const message{data};
```

wmww:
Here's an alternative angle: Mir's window manager only cares about the state windows *are currently in*, not what state windows *would hypothetically be in if they changed out of their current state but not into any other state in particular*. Wayland windows also have the concept of "unsetting fullscreen" (which implementing correctly is currently a TODO), but I'm not convinced changing the core Mir representation is the right move. The frontend should set the state it wants its windows to be in. If knowing what that state will be requires caching some data on previous states, so be it. That is a frontend concern.

src/server/frontend_xwayland/xwayland_wm_surface.cpp
wmww: > It's just as easy to screw up the the order of unpacking as it is indexes

Absolutely not. In the current way you can visually see the order, and so if you know the order properties should be the error is obvious. With indices, you could declare properties in the right order, but unpack the wrong ones or (more likely) copy/paste code but leave duplicate indices. Furthermore, noticing this sort of error is quite hard, even if you're looking at the spec.

Your concern about introducing a non-obvious abstraction are valid. Changed to a more basic method that still satisfies my requirements.

src/server/frontend_xwayland/xwayland_wm_surface.h
wmww: > This information should be stored in Mir's internal representation

I would need to be convinced of this. See longer comment below.

src/server/frontend_xwayland/xwayland_wm_surface.cpp
wmww: > there's a reasonable amount of stuff

Can't implement w/o an example I can test with

RAOF:
I think that's exactly backward; *Mir's WM* should set the state of the windows - possibly, but not necessarily, taking into account client requests as translated by the frontend. If we want the restored →maximised → restored transition to leave everything in the original state we should *also* want the restored → maximised → restored transition for Wayland windows to leave everything in the original state.

That speaks to making it a part of the core Mir WM (or, rather, making CanonicalWindowManager do it by default).

RAOF:
To put it another way: the point of the frontend abstraction is to translate as faithfully as possible Mir's core view into the specific client protocol. There should not be logic in the frontend which is not necessary for translation between the internal representation and the protocol-specific representation.

RAOF:
To put it another way: the point of the frontend abstraction is to translate as faithfully as possible Mir's core view into the specific client protocol. There should not be logic in the frontend which is not necessary for translation between the internal representation and the protocol-specific representation.

AlanGriffiths:
I'm in agreement with RAOF that Mir's window manager should "know" that a minimized window would otherwise "like to be maximized" in order to unminimize correctly. That's something that couldn't be expressed via mirclient and has yet to make it into Mir window management.

Part of the implied work this cycle is cleaning up the window management, and this could be deferred until then.

Meanwhile I'm OK with landing a frig in the X11 frontend that we strip out later. (IMO it isn't the biggest usability issue and could also be left for later.)

RAOF:
> Meanwhile I'm OK with landing a frig in the X11 frontend that we strip out later. (IMO it isn't the biggest usability issue and could also be left for later.)

+1

wmww:
Had sync with @RAOF in which I came to agree that Mir *should* know more about window state, however actually implementing that shall be left until later.

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -127,11 +106,14 @@ void mf::XWaylandWMSurface::net_wm_state_client_message(uint32_t const (&data)[5
...
-    auto const prop_b = values.unpack<xcb_atom_t>();
-    auto const source_indication = values.unpack<SourceIndication>();
+    int i = 0;
+    auto const action = static_cast<Action>(data[i]);
+    i++;
+    auto const prop_a = static_cast<xcb_atom_t>(data[i]);
+    i++;
+    auto const prop_b = static_cast<xcb_atom_t>(data[i]);
+    i++;
+    auto const source_indication = static_cast<SourceIndication>(data[i]);
```
AlanGriffiths: Seems a bit verbose, but OK. 

----

Digressing, I'd prefer:
```c++
    auto const* pdata = data;
    auto const action = static_cast<Action>(*pdata++);
    auto const properties[] = { static_cast<xcb_atom_t>(*pdata++),  static_cast<xcb_atom_t>(*pdata++) };
    auto const source_indication = static_cast<SourceIndication>(*pdata++);
```
Using a `properties` array makes the following loop a little cleaner.

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -125,9 +142,9 @@ void mf::XWaylandWMSurface::net_wm_state_client_message(uint32_t const (&data)[5
 
         new_window_state = window_state;
 
-        for (size_t i = 0; i < length_of(message->properties); i++)
+        for (xcb_atom_t const property : {prop_a, prop_b})
         {
-            if (xcb_atom_t const property = message->properties[i]) // if there is only one property, the 2nd is 0
+            if (property) // if there is only one property, the 2nd is 0
```
wmww: I fail to see how the properties array (red) is cleaner than initializer list (greed)

wmww: I fail to see how the properties array (red) is cleaner than initializer list (green)

wmww: I fail to see how the properties array (red) is cleaner than initializer list (green)

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -125,9 +142,9 @@ void mf::XWaylandWMSurface::net_wm_state_client_message(uint32_t const (&data)[5
 
         new_window_state = window_state;
 
-        for (size_t i = 0; i < length_of(message->properties); i++)
+        for (xcb_atom_t const property : {prop_a, prop_b})
```
wmww: I fail to see how the properties array (red) is cleaner than initializer list (green)

wmww: I fail to see how the properties array (red) is cleaner than initializer list (green)

AlanGriffiths:
@wmww compare:

```c++
    for (xcb_atom_t const property : {prop_a, prop_b})
```
with
```c++
    for (auto property : properties)
```

src/server/frontend_xwayland/xwayland_wm_surface.cpp
```diff
@@ -118,46 +221,76 @@ void mf::XWaylandWMSurface::set_workspace(int workspace)
...
+        case mir_window_state_fullscreen:
+            new_window_state.minimized = false;
+            // don't change new_window_state.maximized
+            new_window_state.fullscreen = true;
+            break;
+
+        case mir_window_state_maximized:
+        case mir_window_state_vertmaximized:
+        case mir_window_state_horizmaximized:
+        case mir_window_state_attached:
```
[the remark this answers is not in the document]
wmww: Done

RAOF:
And now we can test whether bors will pick this up:
bors r+

PR #1212, opened 2020-01-22
XWaylandWM: make threadsafe

wmww:
Introduces and uses a `mutex` and inlines `create()` and `destroy()` into the constructor and destructor (they were only being called from there anyway).


wmww:
`get_wm_surface()` needs to be called from the shell surface which is on the Wayland thread in some cases. I'm not sure if any code doing this has landed yet.

RAOF:
bors r=AlanGriffiths

PR #1213, opened 2020-01-23
Drop "nested" Mir-on-Mir

AlanGriffiths:
Drop "nested" Mir-on-Mir.

Delete more code that uses the mirclient API (and no longer works).


AlanGriffiths:
Build is failing because 19.04 has gone. See #1214.

RAOF:
Yeah, it only works if you don't bors in a review comment?
bors r+

PR #1215, opened 2020-01-23
Split WaylandInputDispatcher out from WaylandSurfaceObserver

wmww:
Splits input event dispatching out from the rest of the Wayland surface observer. This will allow more easily writing an `XWaylandSurfaceObserver`.


AlanGriffiths:
I said

bors r+

Issue #1218, opened 2020-01-24
[Mir-on-X11?] Occasionally input locks when launching apps


AlanGriffiths:
I've sometimes seen this with egmde locking up on Ctrl-Alt-A, I'm now seeing it with miral-shell locking up on Ctrl-Alt-T.

Keyboard input remains "grabbed" in the X11 desktop, but Mir is not responsive to it. However, Mir continues to accept new connections and display new clients.

AFAIK this only happens on the X11 platform and only shortly after launching the server.

AlanGriffiths:
Ohh, I have a reproducer.

With master as it exists prior to #1219 this happens consistently on 18.04.

Interestingly, there's a zombie process:
```plain
$ ps -eaf | grep 21300
alan     21300 21294  2 11:25 pts/1    00:00:05 cmake-build-debug/bin/miral-shell.bin --input-report log
alan     21324 21300  0 11:25 pts/1    00:00:00 [Mir/Input Reade] <defunct>
alan     21362  3663  0 11:28 tty4     00:00:00 grep 21300
```

PR #1220, opened 2020-01-24
Exec errors after forking need to be fatal

AlanGriffiths:
Exec errors after forking need to be fatal. (Fixes: #1218)

### DIFF
--- a/src/miral/launch_app.cpp
+++ b/src/miral/launch_app.cpp
@@ -17,12 +17,12 @@
  */
 
 #include "launch_app.h"
+#include <mir/log.h>
 
 #include <unistd.h>
 
 #include <stdexcept>
 #include <cstring>
-
 
 auto miral::launch_app(
     std::vector<std::string> const& app,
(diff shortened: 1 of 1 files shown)


src/miral/launch_app.cpp
```diff
@@ -80,7 +80,8 @@ auto miral::launch_app(
 
         execvp(exec_args[0], const_cast<char*const*>(exec_args.data()));
 
-        throw std::logic_error(std::string("Failed to execute client (") + exec_args[0] + ") error: " + strerror(errno));
+        mir::log_warning("Failed to execute client (\"%s\") error: %s", exec_args[0], strerror(errno));
+        exit(EXIT_FAILURE);
```
RAOF: Should this be `mir::fatal_error`?

AlanGriffiths: No.

1 That _might_ be replaced by the shell to throw an exception; and, 
2. Logging it as "fatal" would be confusing as the original  Mir process continues perfectly happily.

PR #1221, opened 2020-01-24
Don't advertise mirclient support to GDK

AlanGriffiths:
Don't advertise mirclient support to GDK


RAOF:
Yup!

Also, that's a reasonable test of the 
bors r+
only doesn't work in review comments hypothesis :)

PR #1223, opened 2020-01-24
XWayland: use new role and surface observer

wmww:
Previously there was a class, `XWaylandWMShellSurface`, which used the normal Wayland surface role and observer. These classes are highly specific to normal Wayland protocols, and less useful for XWayland.

This PR introduces and uses two new classes: `XWaylandSurfaceRole` and `XWaylandSurfaceObserver`. `XWaylandWMSurface` is now responsible for managing instances of these classes, as well as handling the `scene::Surface`. The new system is (to the best of my knowledge) threadsafe and more efficient than the old system (on account of not having to run many things on the Wayland thread unnecessarily).


AlanGriffiths:
I like the general thrust of this, but something is wrong in the detail:

1. `miral-app -demo-server --x11-display 3`
2. `DISPLAY=:3 xterm`

Expect: xterm appears as a decorated window with a reasonable size
Actual: xterm is a tiny (one pixel?) window

wmww:
Hmm... confirmed. I'm on it

wmww:
This will get fixed better in the future when we respect `CONFIGURE_NOTIFY` events, but for now we can just pick up the Wayland buffer's size like the old system did.

PR #1232, opened 2020-01-29
Unpublish the mirserver headers

AlanGriffiths:
Unpublish the mirserver headers.

This doesn't drop the libmirserver-dev package (which still has mir/version.h), that can come soon.


RAOF:
Does miral now contain everything required for qtmir? How will this affect UBports?

AlanGriffiths:
> Does miral now contain everything required for qtmir? How will this affect UBports?

No. UBports are on 1.x for the foreseeable future because they rely on mirclient (and I don't intend to backport this to 1.x).

We do need a way forward for something like qtmir (ideally also supporting other toolkits such as gtk), but I expect to resolve that before it blocks ubports.

We could also defer doing this until we have a solution to that.

wmww:
I don't think we should merge this until we have qtmir working

AlanGriffiths:
> I don't think we should merge this until we have qtmir working

I disagree. QtMir will work against Mir 1.x, so we're not breaking anything.

Beyond Mir 1.x we want to make significant changes to the APIs it uses (we're dropping mirclient and want to rework mirserver) which implies reworking or replacing QtMir. Continuing to maintain an API that nobody can use is simply going to add friction.

We can seek feedback from UBports: https://forums.ubports.com/topic/3750/mir-1-x-and-beyond

AlanGriffiths:
> bors r+

You can do this in "approve" comments again: bors-ng/bors-ng#855 is fixed.

mariogrip:
qtmir already rely mostly on miral, but it still have some amount of hooks into mirserver. One progress on this is https://github.com/ubports/qtmir/pull/37 that moves the display events into miral.  but there is still some override hooks that does not have an alternative with miral.

For mirclient qtmir does not use that and has been deprecated on our side and will be replaced with wayland in the near feature.

AlanGriffiths:
> there is still some override hooks that does not have an alternative with miral.

Hi @mariogrip,

yes, when UBports is ready to move to Mir 2.x we need a conversation. These  changes will **not** be cherry-picked to 1.x.

PR #1234, opened 2020-01-30
Repurpose XCBAtoms into XCBConnection

wmww:
The new `XCBConnection` class holds atoms as the `XCBAtoms` class used to. It also now manages the lifetime and gives access to the `xcb_connection_t`. It is passed to each surface, which reduces dependence of the surfaces on the window manager. In the future, XCB helper methods such as setting properties on windows could be implemented on the `XCBConnection`.

### DIFF
--- a/src/server/frontend_xwayland/CMakeLists.txt
+++ b/src/server/frontend_xwayland/CMakeLists.txt
@@ -4,7 +4,7 @@ set(
   xwayland_default_configuration.cpp
   xwayland_connector.cpp  xwayland_connector.h
   xwayland_server.cpp     xwayland_server.h
-  xcb_atoms.cpp           xcb_atoms.h
+  xcb_connection.cpp      xcb_connection.h
   xwayland_wm.cpp         xwayland_wm.h
   xwayland_surface.cpp    xwayland_surface.h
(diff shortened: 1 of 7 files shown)


wmww:
> Would providing a `xcb_atom_t XCBConnection::atom(char const* name)` accessor, plus an external set of constant names, be more obvious and flexible?

IMO this would make the code that uses atoms slightly uglier.

We could achieve the same thing without changing the interface. The `Atom` object could save its own name but not call `xcb_intern_atom()` until it's atom is actually needed. A round-trip would then be required the first time we need each atom (as with the system you propose).

I don't have any actual data on which is more performant. I was under the impression in-flight requests was the whole point of XCB, so having a lot of them should work. I would be unwilling to change to a round-trip heavy approach unless there was actual evidence that was better in some way.

wmww:
Another option is to have each atom add itself to a list. A function could could then be called at the end of the wm constructor that completes them all at once. This wouldn't have any disadvantages AFAIK. Ask if you want it PRed.

AlanGriffiths:
If at first bors don't succeed...

bors r=AlanGriffiths,RAOF

PR #1240, opened 2020-01-31
Fix GTest deprecation

RAOF:
GTest conveniently renamed `INSTANTIATE_TEST_CASE_P` to `INSTANTATE_TEST_SUITE_P` and deprecated the old name. This breaks the build on Focal.

Switch to the new name, providing the old name on older GTests via a compatibility `#define`.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -88,7 +88,7 @@ if (HAS_W_INCONSISTENT_MISSING_OVERRIDE)
   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-inconsistent-missing-override")
 endif()
 if (HAS_W_GNU_ZERO_VARIADIC_MACRO_ARGUMENTS)
-  # INSTANTIATE_TEST_CASE_P hits this like a slice of lemon wrapped around a large gold brick.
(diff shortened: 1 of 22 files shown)


AlanGriffiths:
I'm not (yet) seeing the failure on Focal - this is in proposed?

include/test/mir/test/gtest_fixes.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#ifndef MIR_TEST_GTEST_FIXES_H_
+#define MIR_TEST_GTEST_FIXES_H_
+
+#include "check_gtest_version.h"
+
+#if !GTEST_AT_LEAST(1,9,0)
+#define INSTANTIATE_TEST_SUITE_P(...) INSTANTIATE_TEST_CASE_P(__VA_ARGS__)
+#endif
```
AlanGriffiths: I'm not sure this is the best of possible solutions. Alternatives:
1. `#ifndef INSTANTIATE_TEST_SUITE_P`...
2. Add the define in CMake and avoid introducing another header

RAOF: Turns out I can do this in CMake easily enough.

RAOF:
Fixed to use CMake.

bors r+

RAOF:
Ooops.

bors cancel

RAOF:
bors r=wmww,AlanGriffiths

PR #1245, opened 2020-02-04
XWayland: more refactoring

wmww:
In addition to general cleanup, this changes the flow of configure requests to go though the Mir window manager, so they should no longer be ignored.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -395,6 +395,72 @@ void mf::XWaylandSurface::move_resize(uint32_t detail)
     }
 }
 
+auto mf::XWaylandSurface::WindowState::operator==(WindowState const& that) const -> bool
+{
+    return
+        withdrawn == that.withdrawn &&
+        minimized == that.minimized &&
+        maximized == that.maximized &&
+        fullscreen == that.fullscreen;
+}
+
+auto mf::XWaylandSurface::WindowState::mir_window_state() const -> MirWindowState
+{
+    // withdrawn is ignored
+    if (minimized)
+        return mir_window_state_minimized;
+    else if (fullscreen)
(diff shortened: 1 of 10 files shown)


src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -234,42 +319,32 @@ void mf::XWaylandSurface::dirty_properties()
     props_dirty = true;
 }
 
-void mf::XWaylandSurface::set_surface(WlSurface* wl_surface)
+void mf::XWaylandSurface::set_wl_surface(WlSurface* wl_surface)
```
AlanGriffiths: It seems rather strange to have a `set_wl_surface()` method that doesn't set a `wl_surface` property.

I suspect we need to cache the surface for use in `create_scene_surface_if_needed()`, but the naming seems strange anyway as "setting" isn't the primary functionality here.

wmww:
Note to self: testing with xeyes is not the same as testing with xterm

src/server/frontend_xwayland/xwayland_surface.cpp
wmww: Conceptually, this function does set the `WlSurface` that this `XWaylandSurface` will use. The oddity is that the `wl_surface` is not needed directly outside this function, so we use it to initialize some objects and then forget it.

The motivation for doing it this way is that a `WlSurface` can only be safely used on the Wayland thread, so we do everything we need with it in one place. `create_scene_surface_if_needed()` can later be run on any thread.

Open to changes in the name, but any potential name should indicate that after the function completes the `XWaylandSurface` is in a state such that it will use the given `WlSurface` when required, and to me "setting" is the obvious terminology for that.

wmww: Update: we settled on `attach_wl_surface()`

PR #1248, opened 2020-02-05
XWayland: popups!

wmww:
Improve property reading system and place popups correctly. Popups tested with PCManFM (GTK2), gedit (GTK3) and Kate (Qt, see known issues). XTerm and XEyes also tested.

Known issues:
* Tooltips are only sometimes placed correctly on all toolkits
* Specific Kate menus aren't place correctly

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -134,7 +134,7 @@ mf::XWaylandSurface::XWaylandSurface(
               [this](auto value)
               {
                   std::lock_guard<std::mutex> lock{mutex};
-                  pending_spec(lock).application_id = value;
+                  this->pending_spec(lock).application_id = value;
               }),
           property_handler<std::string const&>(
               connection,
@@ -143,7 +143,7 @@ mf::XWaylandSurface::XWaylandSurface(
               [this](auto value)
               {
                   std::lock_guard<std::mutex> lock{mutex};
-                  pending_spec(lock).name = value;
+                  this->pending_spec(lock).name = value;
               }),
           property_handler<std::string const&>(
               connection,
@@ -152,7 +152,7 @@ mf::XWaylandSurface::XWaylandSurface(
               [this](auto value)
               {
                   std::lock_guard<std::mutex> lock{mutex};
-                  pending_spec(lock).name = value;
+                  this->pending_spec(lock).name = value;
               }),
           property_handler<xcb_window_t>(
               connection,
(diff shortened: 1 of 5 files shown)


AlanGriffiths:
Non blocking notes:

Testing with gnome-terminal:

1. F10 does not bring up the file menu (it does when running the wayland backend).

2. Menus that extend below the main window cannot be navigated with the mouse cursor. In the image, the mouse cannot select below "C".

image

wmww:
Noted.

F10 menu works on Sway's XWayland, but not Mir's. Perhaps a focus issue? (we're still not focusing XWayland clients correctly)

The input off the main surface issue will likely take some deep digging to figure out what's going on.

I don't believe either have a quick fix worth jamming into this PR

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -86,12 +126,77 @@ mf::XWaylandSurface::XWaylandSurface(
...
+                  auto const parent_surface = this->xwm->get_wm_surface(value);
+                  if (parent_surface)
+                  {
+                      std::lock_guard<std::mutex> parent_lock{parent_surface.value()->mutex};
+                      parent_scene_surface = parent_surface.value()->weak_scene_surface.lock();
+                  }
+
+                  {
+                      std::lock_guard<std::mutex> lock{this->mutex};
+                      auto const parent = parent_surface.value()->weak_scene_surface.lock(); // Can be nullptr
```
AlanGriffiths: This looks odd: We've a test `if (parent_surface)` followed (regardless of the outcome) by using `parent_surface.value()`. If `parent_surface` needs testing in the first case, how can we use it in the second?

wmww: Yikes, good catch

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -86,12 +126,76 @@ mf::XWaylandSurface::XWaylandSurface(
...
+          property_handler<xcb_window_t>(
+              connection,
+              window,
+              connection->wm_transient_for,
+              [this](xcb_window_t value)
+              {
+                  std::shared_ptr<scene::Surface> parent_scene_surface; // May remain nullptr
+
+                  auto const parent_surface = this->xwm->get_wm_surface(value);
+                  if (parent_surface)
```
AlanGriffiths: Looks like we could now inline the `parent_surface` definition in the `if`. (And avoid the temptation to use it later.)

wmww: Good idea

PR #1252, opened 2020-02-06
X11 not experimental

AlanGriffiths:
X11 is no longer "experimental".

A follow-up will provide a way to write to supply an FD to receive the display. (Like Xwayland's -displayfd.)

Maybe we should provide a way to select the socket from the command-line, but that's less urgent.


AlanGriffiths:
Marked "no-merge" as we need to update egmde-snap for the new option name.

AlanGriffiths:
(A new PR as I can't re-open #1195)

examples/miral-shell/miral-app.sh
```diff
@@ -76,10 +78,10 @@ unset QT_QPA_PLATFORMTHEME
 if [ "${miral_server}" == "miral-shell" ]
 then
   # miral-shell can launch it's own terminal with Ctrl-Alt-T
-  MIR_SERVER_FILE=${socket} WAYLAND_DISPLAY=${wayland_display} MIR_SERVER_SHELL_TERMINAL_EMULATOR=${terminal} NO_AT_BRIDGE=1 ${hostsocket} exec dbus-run-session -- ${bindir}${miral_server} ${enable_mirclient} $*
+  MIR_SERVER_FILE=${socket} WAYLAND_DISPLAY=${wayland_display} MIR_SERVER_ENABLE_X11=1 MIR_SERVER_SHELL_TERMINAL_EMULATOR=${terminal} NO_AT_BRIDGE=1 ${hostsocket} exec dbus-run-session -- ${bindir}${miral_server} ${enable_mirclient} $*
 else
   # miral-kiosk (and mir_demo_server) need a terminal launched
-  sh -c "MIR_SERVER_FILE=${socket} WAYLAND_DISPLAY=${wayland_display} ${hostsocket} ${bindir}${miral_server} ${enable_mirclient} $*"&
+    MIR_SERVER_FILE=${socket} WAYLAND_DISPLAY=${wayland_display} ${hostsocket} ${bindir}${miral_server} ${enable_mirclient} $*&
```
wmww: Looks like an extra two spaces here

src/miral/x11_support.cpp
```diff
@@ -36,7 +36,7 @@ void miral::X11Support::operator()(mir::Server& server) const
 {
     server.add_configuration_option(
         mo::x11_display_opt,
-        "DISPLAY socket to use for experimental X11 support (default: none).", mir::OptionType::integer);
+        "Enable DISPLAY socket to use for X11 support", mir::OptionType::null);
```
wmww: The wording here is unclear. If the meaning is what I think it is, perhaps "Enable X11 support" would be better.

AlanGriffiths:
bors r=wmww

PR #1254, opened 2020-02-06
[X11] add --x11-displayfd option

AlanGriffiths:
[X11] add --x11-displayfd option.

File descriptor to write X11 DISPLAY number to when ready to connect.

### DIFF
--- a/examples/mir_demo_server/server_example.cpp
+++ b/examples/mir_demo_server/server_example.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2012-2020 Canonical Ltd.
+ * Copyright © 2012-2019 Canonical Ltd.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2 or 3 as
@@ -31,6 +31,7 @@
 #include <miral/runner.h>
 #include <miral/set_window_management_policy.h>
 #include <miral/wayland_extensions.h>
+#include <miral/x11_support.h>
 
 #include "mir/abnormal_exit.h"
 #include "mir/server.h"
@@ -135,6 +136,7 @@ try
         miral::display_configuration_options,
         me::add_log_host_lifecycle_option_to,
         me::add_glog_options_to,
+        miral::X11Support{},
         miral::WaylandExtensions{}
             .enable(miral::WaylandExtensions::zwlr_layer_shell_v1)
             .enable(miral::WaylandExtensions::zxdg_output_manager_v1),
--- a/examples/miral-shell/miral-app.sh
+++ b/examples/miral-shell/miral-app.sh
@@ -80,8 +80,14 @@ then
   # miral-shell can launch it's own terminal with Ctrl-Alt-T
   MIR_SERVER_FILE=${socket} WAYLAND_DISPLAY=${wayland_display} MIR_SERVER_ENABLE_X11=1 MIR_SERVER_SHELL_TERMINAL_EMULATOR=${terminal} NO_AT_BRIDGE=1 ${hostsocket} exec dbus-run-session -- ${bindir}${miral_server} ${enable_mirclient} $*
 else
+  x11_display_file=$(tempfile)
   # miral-kiosk (and mir_demo_server) need a terminal launched
-  MIR_SERVER_FILE=${socket} WAYLAND_DISPLAY=${wayland_display} ${hostsocket} ${bindir}${miral_server} ${enable_mirclient} $*&
+  if [ "${miral_server}" == "mir_demo_server" ]
+  then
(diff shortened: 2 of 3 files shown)


examples/mir_demo_server/server_example.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2012-2020 Canonical Ltd.
+ * Copyright © 2012-2019 Canonical Ltd.
```
wmww: whoops

src/miral/x11_support.cpp
```diff
@@ -34,13 +36,42 @@ miral::X11Support::X11Support() : self{std::make_shared<Self>()}
...
+
+            if (options->is_set(x11_displayfd_opt))
+            {
+                mir::Fd const fd{options->get<int>(x11_displayfd_opt)};
+
+                server.the_main_loop()->enqueue(this, [fd, &server]
+                    {
+                        if (auto const x11_display = server.x11_display())
+                        {
+                            auto const display = x11_display.value().substr(1) + "\n";
```
wmww: Can you please document why you are chopping off the first character or, better yet, assert the first character is indeed a `:`

examples/miral-shell/miral-app.sh
```diff
@@ -80,8 +80,14 @@ then
   # miral-shell can launch it's own terminal with Ctrl-Alt-T
   MIR_SERVER_FILE=${socket} WAYLAND_DISPLAY=${wayland_display} MIR_SERVER_ENABLE_X11=1 MIR_SERVER_SHELL_TERMINAL_EMULATOR=${terminal} NO_AT_BRIDGE=1 ${hostsocket} exec dbus-run-session -- ${bindir}${miral_server} ${enable_mirclient} $*
 else
+  x11_display_file=$(tempfile)
   # miral-kiosk (and mir_demo_server) need a terminal launched
-  MIR_SERVER_FILE=${socket} WAYLAND_DISPLAY=${wayland_display} ${hostsocket} ${bindir}${miral_server} ${enable_mirclient} $*&
+  if [ "${miral_server}" == "mir_demo_server" ]
+  then
+    MIR_SERVER_FILE=${socket} MIR_SERVER_ENABLE_X11=1 WAYLAND_DISPLAY=${wayland_display} ${hostsocket} ${bindir}${miral_server} ${enable_mirclient} $* --x11-displayfd 5 5>${x11_display_file}&
```
wmww: Any reason the rest of the options are set as environment variables, but display fd is not?

examples/miral-shell/miral-app.sh
```diff
@@ -80,8 +80,14 @@ then
...
+  else # miral-kiosk
+    MIR_SERVER_FILE=${socket}                                       WAYLAND_DISPLAY=${wayland_display} ${hostsocket} ${bindir}${miral_server} ${enable_mirclient} $*&
```
wmww: Excessive whitespace

examples/miral-shell/miral-app.sh
wmww: It may be non-obvious to some what's going on, and that `x11_display_file` is a text file not a socket. A comment here might help.

examples/miral-shell/miral-app.sh
```diff
@@ -91,6 +97,14 @@ else
 
   while [ ! -e "${XDG_RUNTIME_DIR}/${wayland_display}" ]; do echo "waiting for ${wayland_display}"; sleep 1 ;done
 
+  if [ -e "${x11_display_file}" ]
+  then
+    export DISPLAY=$(cat "${x11_display_file}")
```
wmww: Perhaps add
```
# x11_display_file should now contain the number of the X11 display Mir created
```

src/miral/x11_support.cpp
```diff
@@ -34,13 +36,42 @@ miral::X11Support::X11Support() : self{std::make_shared<Self>()}
 
 void miral::X11Support::operator()(mir::Server& server) const
 {
+    static auto const x11_displayfd_opt = "x11-displayfd";
```
wmww: Seems arbitrary to mash display and fd into one word. Any reason to prefer that over `x11-display-fd`?

examples/mir_demo_server/server_example.cpp
AlanGriffiths: Nope. This comes with the revert.

examples/miral-shell/miral-app.sh
AlanGriffiths: I thought it more readable with the definition and use of "5" adjacent to each other.

src/miral/x11_support.cpp
AlanGriffiths: Vague consistency with Xmir, Xwayland (which use `-displayfd`).

examples/miral-shell/miral-app.sh
wmww: Ah, fair enough

src/miral/x11_support.cpp
wmww: Good enough for me

PR #1255, opened 2020-02-06
Drop unused atoms

wmww:
These are atoms we're not using and won't need in the immediate future. We can always add them back as we need them.


wmww:
bors r=AlanGriffiths

PR #1256, opened 2020-02-07
Move debugging methods from wm to connection class

wmww:
Makes it more convenient to surface useful debugging messages from the surface class


wmww:
bors r=AlanGriffiths

wmww:
bors r=AlanGriffiths

PR #1262, opened 2020-02-11
Tell Java toolkits that we're a non-reparenting window manager

AlanGriffiths:
Tell Java toolkits that we're a non-reparenting window manager.

This, for example, gets CLion painting correctly under X11.


AlanGriffiths:
Thanks to @RAOF for finding https://github.com/UprootLabs/swing-htabs/issues/22

Issue #1265, opened 2020-02-11
[X11] When mir servers have --enable-x11 they abort on first start


AlanGriffiths:
What is happening (on Xenal/Unity and Focal/GNOME) is that `$DISPLAY` is `:0` but there is no `/tmp/.X0-lock` file.

Mir successfully creates `/tmp/.X0-lock` and then fails to bind `@/tmp/.X11-unix/X0` at which point it "fatal errors" leaving a stray `/tmp/.X0-lock` file.

Subsequent attempts to start succeed as Mir sees `/tmp/X0-lock` and moves on to `/tmp/.X1-lock` etc.

PR #1266, opened 2020-02-11
[X11] Once we've got the lockfile, check for an existing socket

AlanGriffiths:
[X11] Once we've got the lockfile, check for an existing socket before trying to use it. (Fixes: #1265)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -369,13 +369,13 @@ int create_lockfile(int xdisplay)
     if (fd < 0)
         return EEXIST;
 
-    int const size = dprintf(fd, "%10d\n", getpid());
+    bool const success_writing_to_lock_file = dprintf(fd, "%10d\n", getpid()) == 11;
 
     // Check if anyone else has created the socket (even though we have the lockfile)
     char x11_socket[256];
     snprintf(x11_socket, sizeof x11_socket, x11_socket_fmt, xdisplay);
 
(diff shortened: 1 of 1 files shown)


src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -355,19 +355,27 @@ int create_socket(struct sockaddr_un *addr, size_t path_size) {
...
         return EEXIST;
 
     int const size = dprintf(fd, "%10d\n", getpid());
-    if (size != 11)
+
+    // Check if anyone else has created the socket (even though we have the lockfile)
+    char x11_socket[256];
+    snprintf(x11_socket, sizeof x11_socket, x11_socket_fmt, xdisplay);
+
+    if (size != 11 || access(x11_socket, F_OK) == 0)
```
wmww: Where did the 11 come from? I think the answer is `"%10d\n"` will end up being 11 bytes, which is non-obvious. It is especially non-obvious that this magic number would have to be updated whenever that format string changes. I don't see any need to verify a specific number of bytes were written, just that the write was successful. Perhaps this would be better:
```
bool const success_writing_to_lock_file = dprintf(fd, "%10d\n", getpid()) > 0;

...

if (success_writing_to_lock_file && access(x11_socket, F_OK) == 0)
```

AlanGriffiths: As we have no tests of this area of the system I was aiming to fix the logic in the least invasive way.

But sure, I can rework.

PR #1270, opened 2020-02-12
Wait for wl_surface to appear before attaching it to XWayland surface

### LINKED ISSUE #1269: Race condition creating surface
When running in GDB+wayland-debug (but not normal GDB), XWayland GTK3 surfaces do not appear. I'm not particularly concerned about this symptom, but it indicates we have a race condition (as timing is generally the only thing wayland-debug interferes with.

wmww:
Fixes #1269. See that issue for more info on the problem. This allows XWayland to register a callback when a surface with a particular ID is created. It is admittedly a lot of new code, and may not be worth it. It's probably not worth the complexity just to fix Mir under wayland-debug, but hypothetically we could hit the same race condition on a very slow machine, or in some other situation where the Wayland thread was bogged down. I'm open to alternative ideas on how to solve this.

### DIFF
--- a/src/server/frontend_wayland/wayland_connector.cpp
+++ b/src/server/frontend_wayland/wayland_connector.cpp
@@ -246,9 +246,12 @@ public:
     {
     }
 
+    void on_surface_created(wl_client* client, uint32_t id, std::function<void(WlSurface*)> const& callback);
+
 private:
     std::shared_ptr<mg::WaylandAllocator> const allocator;
     std::shared_ptr<mir::Executor> const executor;
+    std::map<std::pair<wl_client*, uint32_t>, std::vector<std::function<void(WlSurface*)>>> surface_callbacks;
 
     class Instance : wayland::Compositor
     {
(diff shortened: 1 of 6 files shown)


src/server/frontend_wayland/wayland_connector.cpp
```diff
@@ -271,9 +274,34 @@ class WlCompositor : public wayland::Compositor::Global
...
+        surface_callbacks[std::make_pair(client, id)].push_back(callback);
+    }
+}
+
 void WlCompositor::Instance::create_surface(wl_resource* new_surface)
 {
-    new WlSurface{new_surface, compositor->executor, compositor->allocator};
+    auto const surface = new WlSurface{new_surface, compositor->executor, compositor->allocator};
+    auto const key = std::make_pair(wl_resource_get_client(new_surface), wl_resource_get_id(new_surface));
+    auto const callbacks = compositor->surface_callbacks.find(key);
```
AlanGriffiths: Shouldn't this be `callbacks = compositor->surface_callbacks.**equal_range**(key);` etc.?

AlanGriffiths: Actually, as surface_callbacks is a map that doesn't matter. (For some reason I was thinking it was a multimap, not a map with vector values)

wmww:
Feels like overkill, but it's not *bad* code and correctness trumps brevity.
bors r+

PR #1273, opened 2020-02-13
BasicSurface: allow frame to be posted after surface destroyed

### LINKED ISSUE #1272: Racy closing of windows kills Xwayland
Testing with ppa:mir-team/dev & egmde (running Fritz on Wine on X11) we crash the Xwayland server...
```plain
[2020-02-13 14:25:47.590090] < -warning- > frontend:Wayland: Exception processing Surface::commit() request: /build/mir-CBxqNk/mir-1.7.0+dev449-g39cf79a94b/src/server/scene/basic_surface.cpp(291): Throw in function mir::scene::BasicSurface::BasicSurface(const std::shared_ptr<mir::scene::Session>&, const string&, mir::geometry::Rectangle, const std::weak_ptr<mir::scene::Surface>&, MirPointerConfinementState, const std::__cxx11::list<mir::scene::StreamInfo>&, const std::shared_ptr<mir::graphics::CursorImage>&, const std::shared_ptr<mir::scene::SceneReport>&)::<lambda(const auto:4&)> [with auto:4 = mir::geometry::Size]
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::runtime_error> >
std::exception::what: Frame posted to dead surface

error in client communication (pid 28439)
(EE) 
Fatal server error:
(EE) failed to write to XWayland fd: Broken pipe
(EE) 
XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":3"
      after 17 requests (17 known processed) with 0 events remaining.
[2020-02-13 14:25:48.047875] <information> xwayland: Xserver stopped
[2020-02-13 14:25:48.048046] <information> xwayland: Cleaning cursors
[2020-02-13 14:25:48.048265] <information> xwayland: Starting Xwayland
```

wmww:
Reading through the code, it looks like setting all buffer stream callbacks to `[](){}` is allowed (that's what we already do in `BasicSurface::set_streams()`, and why shouldn't it be?). If an noop callback is allowed, a conditionally noop callback should also be allowed. I see no reason to throw this error, as it is perfectly reasonable for a surface to be destroyed before its buffer streams.

Fixes #1272

### DIFF
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -287,8 +287,6 @@ ms::BasicSurface::BasicSurface(
         {
             if (auto const o = observers.lock())
                 o->frame_posted(this, 1, size);
-            else
-                BOOST_THROW_EXCEPTION(std::runtime_error("Frame posted to dead surface"));
         };
 
     for (auto& layer : layers)
@@ -925,8 +923,6 @@ void ms::BasicSurface::set_streams(std::list<scene::StreamInfo> const& s)
                 {
                     if (auto const o = observers.lock())
                         o->frame_posted(this, 1, size);
-                    else
-                        BOOST_THROW_EXCEPTION(std::runtime_error("Frame posted to dead surface"));
                 });
         surface_top_left = surface_rect.top_left;
     }
--- a/tests/unit-tests/scene/test_basic_surface.cpp
+++ b/tests/unit-tests/scene/test_basic_surface.cpp
@@ -1319,8 +1319,10 @@ TEST_F(BasicSurfaceTest, registers_frame_callbacks_on_construction)
         { buffer_stream1, {0,0}, {} }
     };
 
-    EXPECT_CALL(*buffer_stream0, set_frame_posted_callback(_));
-    EXPECT_CALL(*buffer_stream1, set_frame_posted_callback(_));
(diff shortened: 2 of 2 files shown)


wmww:
On second thought you should probably wait for me to write some tests

AlanGriffiths:
It is only two months since I introduced that code, there was a reason for it.

(I think I encountered a zombie client that spammed the server. Let me dig...)

wmww:
Oh, I didn't notice the usage of `this`. However, I don't understand how it was any safer before this PR. Did throwing the error prevent a possible future crash?

AlanGriffiths:
> Oh, I didn't notice the usage of `this`. However, I don't understand how it was any safer before this PR. Did throwing the error prevent a possible future crash?

Throwing disconnected glmark and that stopped it spamming us with unwanted buffers.

wmww:
Hmm.. I believe a fully safe solution would be to add
```
for(auto& layer : layers)
        layer.stream->set_frame_posted_callback([](auto){});
```
to the destructor. The only issue I can imagine would be it would dead lock if the surface is destroyed from a callback. I will push this along with some tests in a bit.

wmww:
This is now ready to merge. I'll let someone else bors after they check the tests are sensible.

PR #1279, opened 2020-02-14
Improve XCBConnection atom lookup

wmww:
Fixes a potential thread safety issue in the resolving of `XCBConnection::Atom` and starts caching atom names for faster lookup.


wmww:
On second thought, there's no reason for these to be in the same PR

wmww:
I may have been able to reproduce once, but I can't do so again. Do you have strong reason to believe this PR in particular introduces a new problem?

AlanGriffiths:
> Do you have strong reason to believe this PR in particular introduces a new problem?

Not seen before applying; seen consistently with this PR (both on master, and in combination with other PRs under review), not seen after reverting.

wmww:
After some testing it looks like the atom atomic was being garbage initialized after all, and my last commit fixes it.

wmww:
(I say after all because that was my theory, then I thought I class struct members are initialized to zero by default so I thought making it explicit wouldn't change anything)

PR #1281, opened 2020-02-14
Fix XWaylandSurface::property_notify()

wmww:
We need to fully process properties before consuming the pending spec, and should be reading properties even when we do not have a scene surface.


wmww:
What changes in practice is that changes to the pending spec made in `completion()` get picked up and applied to the surface, which can't happen if the spec is consumed before `completion()` is called.

bors r+

PR #1286, opened 2020-02-17
Fix unnecessary copies in range-for loops.

RAOF:


RAOF:
Clang-11 will warn for this, so why not fix it now? :grin:

PR #1290, opened 2020-02-17
Fix `miral-app -demo-server`

AlanGriffiths:
Fix `miral-app -demo-server`.

Export DISPLAY correctly when starting mir_demo_server.


AlanGriffiths:
> Shhh I'm technically not working today

That's what I thought. Thanks for landing these anyway!

PR #1293, opened 2020-02-19
platform: New CPU-accessible buffer API.

RAOF:
The two fundamental types of CPU-accessble buffers are those that can be
mapped into CPU accessible memory and those that require explicit
transfers to/from CPU accessible memory.

Model these as {Read,Write}MappableBuffer and
{Read,Write}TransferableBuffer respectively.


RAOF:
(No-merge as I expect to push more cleanup, but this can usefully be reviewed from a high-level perspective)

RAOF:
I think this is probably enough of a chunk to merge now. Once the tests pass!

include/platform/mir/renderer/sw/pixel_source.h
```diff
@@ -30,6 +35,74 @@ namespace renderer
 namespace software
 {
 
+class BufferDescriptor
```
RAOF: Candidates for a better name for `BufferDescriptor` are most welcome!

include/platform/mir/renderer/sw/pixel_source.h
```diff
@@ -30,6 +35,74 @@ namespace renderer
...
+
+/**
+ * A Buffer that can be mapped into CPU-accessible memory and directly written to.
+ */
+class WriteMappableBuffer
+{
+public:
+    virtual ~WriteMappableBuffer() = default;
+
+    virtual std::unique_ptr<Mapping<unsigned char>> map_writeable() = 0;
```
RAOF: Aaargh.

As soon as I try to document this I realise we need a different API. This is close, but you can't express performance-useful code like “only write to the bits of the buffer you care about”, or “write only to the damaged areas”.

On the other hand, how much do we care about performance here? :grimacing:

include/platform/mir/renderer/sw/pixel_source.h
AlanGriffiths: Well, this appears to simply be some properties a has Buffer. So "Buffer"? Or "GenericBuffer"?

include/platform/mir/renderer/sw/pixel_source.h
AlanGriffiths: Is the different API incompatible? Could it be added if we find we need it?

Having something that works now seems better than something perfect at some indefinite future moment.

include/platform/mir/renderer/sw/pixel_source.h
```diff
@@ -30,6 +35,74 @@ namespace renderer
...
+    virtual ~WriteTransferableBuffer() = default;
+
+    virtual void transfer_into_buffer(unsigned char const* source) = 0;
+};
+
+auto as_read_mappable_buffer(
+    std::shared_ptr<graphics::Buffer> buffer) -> std::shared_ptr<ReadMappableBuffer>;
+auto as_write_mappable_buffer(
+    std::shared_ptr<graphics::Buffer> buffer) -> std::shared_ptr<WriteMappableBuffer>;
+
```
AlanGriffiths: Do these need to take `buffer` by value?

RAOF: They don't *have* to, but they always consume `buffer` so there's no advantage to passing in a `shared_ptr<> const&`.

clang-tidy started to suggest passing-by-value in such cases, and I don't think it's wrong.

include/platform/mir/renderer/sw/pixel_source.h
RAOF: The different API would be incompatible, yes. Or, rather, it wouldn't change any symbol names, but it would render existing usages incorrect - with incorrect usages resulting in rendering from uninitialised memory.

It turns out that none of our *existing* code cares about this at all, and could live with a simpler “alloc-and-fill-a-buffer” API that doesn't have these questions. The decoration code particularly would benefit from an extended API, but this'll do for now.

PR #1294, opened 2020-02-19
Add wayland "Recommends" to mir-demos dependencies

AlanGriffiths:


debian/control
```diff
@@ -233,7 +233,9 @@ Provides: miral-examples (=1.5)
 Depends: ${misc:Depends},
          ${shlibs:Depends},
 Recommends: dmz-cursor-theme,
-            fonts-liberation
+            fonts-liberation,
+            qtwayland5,
+            xwayland
```
Saviq:
```suggestion
            xwayland,
```
(This way the next diff won't have a `-`)
Otherwise, good.

AlanGriffiths: Ack

Saviq:
bors d=AlanGriffiths

Issue #1295, opened 2020-02-19
Frontend is not told what surface has input focus


wmww:
The surface observer is informed of if each surface is focused at any given time, but that is different than if a surface is the single surface with "Input focus". It currently works because the last surface to become focused is always assumed to have input focus, and we have some hacks in place to make that assumption hold. Fixing this is probably required for #827.

wmww:
The hack mentioned is here: https://github.com/MirServer/mir/blob/f64659b9e44676976cc00f979fc20779d80300d2/src/server/shell/abstract_shell.cpp#L419-L423

We need to check that any potential fix does not regress #823.

Issue #1274, opened 2020-02-13
[X11] When using CLion the keyboard focus gets screwed up


AlanGriffiths:
When editing code in CLion (under our X11 support on master).

When typing "tooltips" are shown, e.g. for completion. These appear and disappear as you type.

Sometimes, while this is happening, the input cursor disappears and typing has no effect. But up/down arrows change the highlighted line.

A mouse click restores the cursor and normal function.

(Not sure if this is CLion specific or if it could affect other Java based UIs.)

AlanGriffiths:
or a simple reproducer:

1. Create a new C++ project (C++ Executable)
2. Move cursor to body of main
2. Type "for (auto "

By this time we should be in the weird state described.

wmww:
I love how Java sets `WM_NAME`, `_NET_WM_NAME`, `WM_ICON_NAME` and `_NET_WM_ICON_NAME` to the same very long value after every keystroke. And of course we don't cache that so we send a `modify_window()` on every keystroke.

wmww:
I think I know what's going on. In some situations, CLion/IntelliJ tears down a popup then immediately creates a new one. As the first one gets destroyed, we set input focus to the parent. Somehow that's interfering with the creation of the new popup, and we're left in a bad state. If I stop setting input focus to the parent it fixes the problem but breaks other apps. Looking into a better solution.

wmww:
Ah! removing the hack in abstract_shell.cpp mentioned in #1295 also fixes it

wmww:
Feel free to release once current PRs are merged. A few of the XWayland issues still pose a serious threat to usability, but those aren't getting fixed this week.

PR #1296, opened 2020-02-19
Fix XWaylandWM::set_focus()

### LINKED ISSUE #1278: XWaylandWM::set_focus() is very broken
Specifically, this is implicitly casting a bool to an XCB window:
https://github.com/MirServer/mir/blob/ca0188c1326edfb3083033dd680d731ac86a5ce4/src/server/frontend_xwayland/xwayland_wm.cpp#L288
I have a fix, but it causes some popups to immediately disappear so I want to investigate further before I merge it.

wmww:
Fixes #1278. Input focus should now be correctly switched between apps. Due to the workaround for #1295, when a Wayland app is focused the last X app is not told it is unfocused, but it **does** stop receiving any input until it is refocused.


AlanGriffiths:
I was hoping this might help with #1274, but it has no effect on that. :(

PR #1299, opened 2020-02-19
Give windows human readable names

wmww:
Makes remembering window names when debugging much easier. I threw up a generic form of this function here: https://github.com/wmww/name-of/.


wmww:
bors try

wmww:
bors r=AlanGriffiths

PR #1302, opened 2020-02-20
X11 platform: scale option for "fake" output

wmww:
We already have an option that allows creating multiple X11 windows that act as outputs of different sizes. This adds the additional functionality of optionally specifying a scale for each output with the syntax `^SCALE`.

For example, `--x11-output 640x480:1020x720^2.5` would result in two "outputs", a 640x480 one with a scale of 1 and a 1020x720 one with a scale of 2.5.

This aims to make testing HiDPI much easier.

### DIFF
--- a/src/platforms/mesa/server/x11/graphics/platform.cpp
+++ b/src/platforms/mesa/server/x11/graphics/platform.cpp
@@ -49,40 +49,39 @@ auto parse_size_dimension(std::string const& str) -> int
     }
 }
 
-auto parse_size(std::string const& str) -> geom::Size
+auto parse_size(std::string const& str) -> mgx::X11OutputConfig
 {
     size_t x = str.find('x');
     if (x == std::string::npos || x <= 0 || x >= str.size() - 1)
         BOOST_THROW_EXCEPTION(std::runtime_error("Output size \"" + str + "\" does not have two dimensions"));
-    return geom::Size{
+    return mgx::X11OutputConfig{geom::Size{
         parse_size_dimension(str.substr(0, x)),
-        parse_size_dimension(str.substr(x + 1, std::string::npos))};
-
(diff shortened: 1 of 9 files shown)


src/platforms/mesa/server/x11/graphics/platform.cpp
```diff
@@ -49,40 +49,67 @@ auto parse_size_dimension(std::string const& str) -> int
...
+    return mgx::X11OutputConfig{
+        geom::Size{
+            parse_size_dimension(str.substr(0, x)),
+            parse_size_dimension(str.substr(x + 1, scale_start - x - 1))},
+        scale};
 }
 }
 
-std::vector<geom::Size> mgx::Platform::parse_output_sizes(std::string output_sizes)
+auto mgx::Platform::parse_output_sizes(std::string output_sizes) -> std::unique_ptr<std::vector<mgx::X11OutputConfig>>
```
AlanGriffiths: Why the change from vector<> to unique_ptr<vector<>>?

wmww: So that `mgx::X11OutputConfig` can be forward declared

AlanGriffiths: I don't understand that answer. How does using `unique_ptr<>` change the ability to say `class X11OutputConfig`?

wmww: I was incorrect in two ways:
1. `X11OutputConfig` is declared in platform.h (not forward declared like I originally planned it to be)
2. I incorrectly believed you couldn't have a vector of a forward declared type.

Will push changes

wmww:
I doubt if that is new (although it may expose an existing bug). The flicker appears to be #1317.

AlanGriffiths:
Is bors getting webhook notifications yet?

bors r+

PR #1303, opened 2020-02-20
Strip out any MIR_* environment variables when launching apps

AlanGriffiths:
Strip out any MIR_* environment variables when launching apps

These can cause annoyance when launching Mir compositors (such as tests) from a Mir Shell.


AlanGriffiths:
bors r=wmww

PR #1304, opened 2020-02-20
Only modify surface if something changed

wmww:
*Some* toolkits (\*cough\* Java \*cough\*) like to spam changes to window properties. No need to propagate that all the way through Mir.


wmww:
bors r=AlanGriffiths

PR #1315, opened 2020-02-24
Don't interleave iterating over environ with unsetting environment variables

AlanGriffiths:
Don't interleave iterating over environ with unsetting environment variables

### DIFF
--- a/src/miral/launch_app.cpp
+++ b/src/miral/launch_app.cpp
@@ -23,12 +23,15 @@
 
 #include <stdexcept>
 #include <cstring>
+#include <vector>
 
 namespace
 {
 void strip_mir_env_variables()
 {
     static char const mir_prefix[] = "MIR_";
+
+    std::vector<std::string> vars_to_remove;
 
     for (auto var = environ; *var; ++var)
     {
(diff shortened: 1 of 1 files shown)


src/miral/launch_app.cpp
```diff
@@ -23,28 +23,36 @@
...
             if (auto var_end = strchr(var_begin, '='))
             {
-                unsetenv(std::string(var_begin, var_end).c_str());
+                vars_to_remove.emplace_back(var_begin, var_end);
             }
             else
             {
-                unsetenv(var_begin);
+                vars_to_remove.emplace_back(var_begin);
             }
```
Saviq: For future's sake:
```
<Saviq> alan_g: for my sake, IIUC some elements of `environ` have equal signs, some don't? it's because some of them will have values, other will just be defined?
<alan_g> Saviq, exactly
<alan_g> Or rather, they all seem to have "=" but the docs don't seem to guarantee it
<Saviq> hmm, "by convention"
<alan_g> Actually, I guess "=" is required. I was just paranoid.
<alan_g> https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html
<alan_g> (But that's pre-existing.)
```

+1 on this fixing the problem of removing from a moving target.

bors r+

Issue #1321, opened 2020-02-25
[Mir Shell] running GTK apps makes at-spi2-registr hog CPU


AlanGriffiths:
Probably 20.04 only

AlanGriffiths:
Workaround:
```plain
$ cat ~/.config/miral-shell.config 
app-env=GDK_BACKEND=wayland:QT_QPA_PLATFORM=wayland:SDL_VIDEODRIVER=wayland:NO_AT_BRIDGE=-3:_JAVA_AWT_WM_NONREPARENTING=1:-GTK_MODULES
```

AlanGriffiths:
So the problem is that we don't want to "force" accessibility when we have no support in miral-shell.

>One can force enabling accessibility (but ideally it should be automatic thanks to the property mentioned above):
>
> export GTK_MODULES=gail:atk-bridge
> export OOO_FORCE_DESKTOP=gnome
> export GNOME_ACCESSIBILITY=1
> export QT_ACCESSIBILITY=1
> export QT_LINUX_ACCESSIBILITY_ALWAYS_ON=1

https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/

PR #1322, opened 2020-02-25
[miral] Unset any app environment variables that might force enabling accessibility

AlanGriffiths:
[miral] Unset any app environment variables that might force enabling accessibility. (Fixes: #1321)

See https://www.freedesktop.org/wiki/Accessibility/AT-SPI2/ for a list of such variables.


AlanGriffiths:
bors r=wmww

AlanGriffiths:
bors r=wmww

PR #1323, opened 2020-02-25
Use correct physical output size in mgm::Display::configure_locked()

wmww:
Previous approach seems convoluted and error-prone to me, and was causing CRTC related failures when scale set to anything other than 1.

### DIFF
--- a/src/platforms/mesa/server/kms/display.cpp
+++ b/src/platforms/mesa/server/kms/display.cpp
@@ -488,6 +488,7 @@ void mgm::Display::configure_locked(
             // Each vector<KMSOutput> is a single GPU memory domain
             std::vector<std::vector<std::shared_ptr<KMSOutput>>> kms_output_groups;
             glm::mat2 transformation;
+            geom::Size physical_size{};
 
             group.for_each_output(
                 [&](DisplayConfigurationOutput const& conf_output)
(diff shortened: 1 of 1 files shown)


src/platforms/mesa/server/kms/display.cpp
```diff
@@ -488,6 +488,7 @@ void mgm::Display::configure_locked(
             // Each vector<KMSOutput> is a single GPU memory domain
             std::vector<std::vector<std::shared_ptr<KMSOutput>>> kms_output_groups;
             glm::mat2 transformation;
+            geom::Size physical_size{};
```
RAOF: I'm not entirely wild about the name `physical_size` here, because we *also* have a field for the actual physical of the display :grin: 

Maybe `resolution`? That's got a long history of meaning the correct thing.

src/platforms/mesa/server/kms/display.cpp
```diff
@@ -518,13 +521,8 @@ void mgm::Display::configure_locked(
             }
             else
             {
-                glm::vec2 const logical_size{
-                    bounding_rect.size.width.as_uint32_t(),
-                    bounding_rect.size.height.as_uint32_t()};
-
-                auto const physical_size = transformation * logical_size;
```
RAOF: Whoops! This is *exactly* wrong - `logical_size = physical_size * transformation`! No wonder it blows up with non-trivial transformations.

AlanGriffiths: Well, it is accidentally right some of the time: when `transformation` is a rotation of n*90 degrees. (Which is probably the unstated assumption here.)

wmww:
I believe the transformation is applied before a thing is rendered to the `GBMOutputSurface`, so its size should always match the screens size.

AlanGriffiths:
bors r=RAOF,AlanGriffiths

PR #1325, opened 2020-02-26
Fix launching gnome-terminal from "Mir Shell"

AlanGriffiths:
Fix launching gnome-terminal from "Mir Shell".

We should not supply both --disable-factory and --app-id, as that spawns multiple terminal servers. (Which eventually leads to madness.).


AlanGriffiths:
bors r=wmww

PR #1328, opened 2020-02-27
Mesa/KMS: position cursor correctly when output scaled

wmww:
This correctly applies output transform and scale when positioning the hardware cursor

### DIFF
--- a/src/platforms/mesa/server/kms/cursor.cpp
+++ b/src/platforms/mesa/server/kms/cursor.cpp
@@ -333,7 +333,7 @@ void mgm::Cursor::hide()
 }
 
 void mgm::Cursor::for_each_used_output(
-    std::function<void(KMSOutput&, geom::Rectangle const&, MirOrientation orientation)> const& f)
+    std::function<void(KMSOutput& output, DisplayConfigurationOutput const& conf)> const& f)
 {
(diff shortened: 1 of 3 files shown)


src/platforms/mesa/server/kms/cursor.cpp
```diff
@@ -382,18 +381,35 @@ void mgm::Cursor::place_cursor_at_locked(
...
+                2.0f - glm::vec2{1};
+
+            auto const transformed_vec = normalized_vec * conf.transformation();
+
+            auto const output_space_vec = (transformed_vec + glm::vec2{1}) / 2.0f *
+                glm::vec2{output.size().width.as_int(), output.size().height.as_int()};
+
+            auto const position_on_output = geom::Point{roundf(output_space_vec.x), roundf(output_space_vec.y)};
+
+            //geom::Point const hotspot_position_on_output = transform(output_rect, position - output_rect.top_left, orientation);
```
AlanGriffiths: This comment contradicts the code. How can we trust either?

AlanGriffiths:
@wmww the "failed" builds are Fedora images that we're allowing to fail because of issues outside our code. Maybe rebasing on master will clarify things for Travis.

Issue #1327, opened 2020-02-26
[miral-shell] Hangs on exit when run with multiple X11 windows


AlanGriffiths:
```plain
miral-app --x11-output 640x480:640x480
```
Hit `Ctrl-Alt-BkSp` to exit.

Expect: both windows close
Actual: Both windows go black and keyboard input remains grabbed.

AlanGriffiths:
This is a problem with the "spinner". I have the following stacktrace:

```plain
#3  0x00005555555d4a05 in MirEglSurface::MirEglSurface(std::shared_ptr<MirEglApp> const&, wl_output*, int, int) (this=0x7fffbc2b0ce0, mir
_egl_app=std::shared_ptr<MirEglApp> (use count 2, weak count 0) = {...}, wl_output=0x7fffbc005dc0, width_=640, height_=480)
    at /home/alan/display_server/mir/examples/miral-shell/spinner/miregl.cpp:245
#4  0x00005555555d7fb0 in __gnu_cxx::new_allocator<MirEglSurface>::construct<MirEglSurface, std::shared_ptr<MirEglApp> const&, wl_output*
 const&, int const&, int const&>(MirEglSurface*, std::shared_ptr<MirEglApp> const&, wl_output* const&, int const&, int const&) (this=0x7f
ffd97f9677, __p=0x7fffbc2b0ce0, __args#0=std::shared_ptr<MirEglApp> (use count 2, weak count 0) = {...}, __args#1=@0x7fffbc005ec0: 0x7fff
bc005dc0, __args#2=@0x7fffbc005ed0: 640, __args#3=@0x7fffbc005ed4: 480) at /usr/include/c++/7/ext/new_allocator.h:136
#5  0x00005555555d7c12 in std::allocator_traits<std::allocator<MirEglSurface> >::construct<MirEglSurface, std::shared_ptr<MirEglApp> cons
t&, wl_output* const&, int const&, int const&>(std::allocator<MirEglSurface>&, MirEglSurface*, std::shared_ptr<MirEglApp> const&, wl_outp
ut* const&, int const&, int const&) (__a=..., __p=0x7fffbc2b0ce0, __args#0=std::shared_ptr<MirEglApp> (use count 2, weak count 0) = {...}
, __args#1=@0x7fffbc005ec0: 0x7fffbc005dc0, __args#2=@0x7fffbc005ed0: 640, __args#3=@0x7fffbc005ed4: 480)
[...]

PR #1329, opened 2020-02-27
[miral-shell] Avoid unnecessary blocking call in Spinner setup

AlanGriffiths:
[miral-shell] Avoid unnecessary blocking call in Spinner setup. (Fixes: #1327)


AlanGriffiths:
bors r=wmww

PR #1331, opened 2020-02-27
miral-shell background support scaling

wmww:
Because Wayland only supports integer scaling the miral-shell background is too small on outputs with fractional scaling. In the future this could be fixed by using the logical size from `xdg_output`.


AlanGriffiths:
With a screen scale of 2 I now see this:

Screenshot from 2020-02-27 18-03-39

That seems wrong.

wmww:
I need to look more into it, but I suspect I forgot to include a required piece into this PR

wmww:
There is a missing piece, but also this is not doing things quite right. Closing for now

Issue #1332, opened 2020-02-28
[master, 18.04] Segfault in renderer


AlanGriffiths:
1. Start multiscreen Mir-on-X: `miral-app -demo-server --x11-output 640x480:800x600`.
2. Position windows side-by-side to make it easy to move cursor across windows.
3. Drag the terminal window back and forth between windows.

Expect: everything works smoothly.
Actual: lots of flickering and odd display effects, followed by a core dump.

```plain
(gdb) bt
#0  0x00007fdba09fda9a in  () at /usr/lib/x86_64-linux-gnu/dri/i965_dri.so
#1  0x00007fdba09fdf4a in  () at /usr/lib/x86_64-linux-gnu/dri/i965_dri.so
#2  0x00007fdba0a068ca in  () at /usr/lib/x86_64-linux-gnu/dri/i965_dri.so
#3  0x00007fdba0a06b84 in  () at /usr/lib/x86_64-linux-gnu/dri/i965_dri.so
#4  0x00007fdba09d6def in  () at /usr/lib/x86_64-linux-gnu/dri/i965_dri.so
#5  0x00007fdba0c64c28 in  () at /usr/lib/x86_64-linux-gnu/dri/i965_dri.so
#6  0x00007fdba0c64cc5 in  () at /usr/lib/x86_64-linux-gnu/dri/i965_dri.so
#7  0x00007fdbadb6c0d4 in mir::renderer::gl::Renderer::draw(mir::graphics::Renderable const&) const (this=0x7fdb8c000b20, renderable=...)
    at /home/alan/display_server/mir/src/renderers/gl/renderer.cpp:615
#8  0x00007fdbadb6b45d in mir::renderer::gl::Renderer::render(std::vector<std::shared_ptr<mir::graphics::Renderable>, std::allocator<std::shared_ptr<mir::graphics::Renderable> > > const&) const (this=0x7fdb8c000b20, renderables=std::vector of length 1, capacity 1 = {...})
    at /home/alan/display_server/mir/src/renderers/gl/renderer.cpp:421
#9  0x00007fdbad930728 in mir::compositor::DefaultDisplayBufferCompositor::composite(std::vector<std::shared_ptr<mir::compositor::SceneElement>, std::allocator<std::shared_ptr<mir::compositor::SceneElement> > >&&) (this=0x7fdb8c02c7e0, scene_elements=...)
    at /home/alan/display_server/mir/src/server/compositor/default_display_buffer_compositor.cpp:84
[...]

Issue #1317, opened 2020-02-25
Flicker when multiple outputs are connected


wmww:
Observed on my XPS running 20.04. When an external monitor is connected and windows are on both outputs, both tend to flicker and sometimes get the content of each other. It is especially bad when a window is being dragged between the two outputs.

wmww:
When the window has a subsurface (such as the gedit open dropdown), the main window and the subsurface sometimes end up separated (one on the correct output, the other in the correct position but on the wrong output). Since subsurfaces are all part of the same Mir `BasicSurface`, this indicates to me the bug happens after  renderables have been created by the surface.

wmww:
Observed on the X11 platform as well

AlanGriffiths:
@wmww on #1332 I discovered similar problems that exist only on master. Is that also the case here?

AlanGriffiths:
Not reproduced with egmde-snap/edge on 18.04 or 16.04. (Likely fixed by #1343)

PR #1148, opened 2019-12-16
graphics/wayland: Handle wl_shm buffers through the WaylandAllocator

RAOF:
This lets platforms do platform-specific optimisations for SHM buffers. All extant platforms currently use the newly-added `mg::wayland::buffer_from_wl_shm` helper, but can now opportunistically transition to platform-specific code.

PR #1333, opened 2020-02-28
Revert merge #1148

AlanGriffiths:
Revert merge #1148 (Fixes: #1332)

I know we want most of this back again. But this gets us back to a working state.


AlanGriffiths:
@RAOF what I *think* is happening is that buffers are being released after being consumed by the first compositor. That begins a race between the second compositor using the buffer and the client destroying it. (But I've not spent time proving that theory - we can talk next week.)

wmww:
Since this is a lot of code and most of it is probably good, I'm in favor of fixing it without reverting and then having to re-review the whole thing

AlanGriffiths:
> Since this is a lot of code and most of it is probably good, I'm in favor of fixing it without reverting and then having to re-review the whole thing

It isn't hard to review a revert of a revert commit. Fixes can come on top of that.

AlanGriffiths:
@wmww I'm curious does this also fix #1317? (If my theory is right it should, but I've reproduced.)

wmww:
Don't have access to a screen right now. Will try to remember to test when I do

wmww:
This does fix #1317

AlanGriffiths:
Closing: We have a fix under review.

PR #1341, opened 2020-03-02
[MirAL] Implement a "no active window" state

peat-psuwit:
When setting the active window to a null window, mru_active_windows
won't be updated. This commit add the `allow_active_window` flag which
will be false when a null window is set, disabling the active window.

This is used by Lomiri during spreads and other effects and animations.

Co-authored-by: Alan Griffiths <alan@octopull.co.uk>

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -213,7 +213,7 @@ void miral::BasicWindowManager::remove_surface(
 
 void miral::BasicWindowManager::remove_window(Application const& application, miral::WindowInfo const& info)
 {
-    bool const is_active_window{mru_active_windows.top() == info.window()};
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
Sorry! I intended to PR against this branch, not push to it.

But I think this fixes the broken behavior. Hope that's OK

AlanGriffiths:
If we agree the updated code, we need to update the title and merge comment to reflect the new behaviour, not the implementation initially proposed.

peat-psuwit:
I'll squash the commit, test it on Ubuntu Touch, and then report back to you.

peat-psuwit:
Ok. I confirmed that your updated code fixes the issue in Unity8/Lomiri.

PR #1343, opened 2020-03-03
platform: Ensure each rendering thread has a separate gl::Program.

RAOF:
The GL uniforms (such as, for example, screen position) are per Program state, *not*
per context state, so when there were multiple rendering threads they would race.

Fix this by generating a `Program` per `ProgramFactory`; we already had a `ProgramFactory`
per thread.

Fixes: #1332, #1317

### DIFF
--- a/include/platform/mir/graphics/program_factory.h
+++ b/include/platform/mir/graphics/program_factory.h
@@ -41,10 +41,12 @@ public:
     /**
      * Compile and link a fragment-shader fragment into a full shader Program
      *
-     * \note    As the id needs to be globally unique, we suggest using the address of a local
-     *          static variable as the id.
+     * \note    The ID is used to identify this shader program for caching purposes. Accidentally
+     *          using an ID which is already used by another shader will result in incorrect rendering.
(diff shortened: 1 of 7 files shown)


include/platform/mir/graphics/program_factory.h
```diff
@@ -41,16 +41,21 @@ class ProgramFactory
     /**
      * Compile and link a fragment-shader fragment into a full shader Program
      *
-     * \param extension_fragment [in]    Fragment to include necessary extensions
+     * \note    As the id needs to be globally unique, we suggest using the address of a local
+     *          static variable as the id.
```
AlanGriffiths: Does that make it globally unique? A local static variable will have the same address in multiple invocations.

RAOF: Hm. I've clearly communicated badly.

The `id` needs to be the same for each invocation of `compile_fragment_shader` for the same shader, and different for each different shader.

It doesn't need to (and, in fact, needs *not* to be) be unique for each call.

PR #1349, opened 2020-03-05
Allow division of geometry types by a scalar

wmww:


include/core/mir/geometry/dimensions.h
```diff
@@ -175,6 +175,16 @@ inline constexpr DeltaX operator*(DeltaX const& dx, Scalar scale) { return scale
 template<typename Scalar>
 inline constexpr DeltaY operator*(DeltaY const& dy, Scalar scale) { return scale*dy; }
 
+// Dividing by a scaler value is fine
+template<typename Scalar>
+inline constexpr Width operator/(Width const& w, Scalar scale) { return w / scale; }
```
AlanGriffiths: Doesn't this recursively invoke `Width operator/(Width& w, Scalar)`?

That would not be useful!

PR #1350, opened 2020-03-05
DisplayConfiguration respect scale

wmww:
Make `DisplayConfiguration::extents_of()` return the scaled logical size, instead of the pixel size of the mode

### DIFF
--- a/tests/unit-tests/graphics/test_display_configuration.cpp
+++ b/tests/unit-tests/graphics/test_display_configuration.cpp
@@ -391,3 +391,37 @@ TEST(DisplayConfiguration, output_extents_empty_when_there_are_no_modes)
     geom::Rectangle empty{};
     EXPECT_EQ(empty, out.extents());
 }
+
+TEST(DisplayConfiguration, output_extents_are_scaled)
+{
+    mg::DisplayConfigurationOutput out = tmpl_output;
+    out.scale = 2.0f;
+
+    EXPECT_EQ(out.modes[out.current_mode_index].size * 0.5, out.extents().size);
+}
+
+TEST(DisplayConfiguration, output_extents_are_scaled_fractionally)
+{
+    mg::DisplayConfigurationOutput out = tmpl_output;
+    out.scale = 0.8f;
+
+    EXPECT_EQ(out.modes[out.current_mode_index].size * 1.25, out.extents().size);
+}
+
+TEST(DisplayConfiguration, user_display_configuration_output_extents_are_scaled)
+{
+    mg::DisplayConfigurationOutput out = tmpl_output;
+    mg::UserDisplayConfigurationOutput user{out};
+    user.scale = 2.0f;
+
+    EXPECT_EQ(user.modes[user.current_mode_index].size * 0.5, user.extents().size);
+}
+
+TEST(DisplayConfiguration, user_display_configuration_output_extents_are_scaled_fractionally)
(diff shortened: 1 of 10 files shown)


tests/unit-tests/graphics/test_display_configuration.cpp
```diff
@@ -391,3 +391,37 @@ TEST(DisplayConfiguration, output_extents_empty_when_there_are_no_modes)
     geom::Rectangle empty{};
     EXPECT_EQ(empty, out.extents());
 }
+
+TEST(DisplayConfiguration, output_extents_are_scaled)
+{
+    mg::DisplayConfigurationOutput out = tmpl_output;
+    out.scale = 2.0f;
+
+    EXPECT_EQ(out.modes[out.current_mode_index].size * 0.5, out.extents().size);
```
AlanGriffiths: I've a mild preference for:
```c++
     EXPECT_THAT(out.modes[out.current_mode_index].size * 0.5, Eq(out.extents().size));
```
it reads more fluently.

wmww:
I don't know why that particular case isn't working. There's still quite a few things to patch through (some I have working in other branches, some are still in progress). I'll make sure that works in the end, but I don't think we should block, else this PR becomes the entire HiDPI feature.

There's probably nothing wrong with this PR except that it "fixes" one component without also applying fixes to other components. Since output scaling is a new feature anyway (not to to be used in production yet), I don't think there should be a problem with temporarily breaking use cases as long as we're progressing towards a fully working state).

AlanGriffiths:
>  output scaling is a new feature anyway (not to to be used in production yet), I don't think there should be a problem with temporarily breaking use cases as long as we're progressing towards a fully working state).

That's not quite true. Other Wayland "hosts" (e.g. GNOME) already have scaled outputs.

Currently, Mir-on-Wayland works with the host being a scaled GNOME output. With this PR it fails in the same way as the reproducer.

But if you want to land this, I'll take a look at Mir-on-Wayland.

bors delegate+

wmww:
I'm under the impression the problem is in the host, not Mir-on-Wayland (since, as you say, that works on other compositors).

bors r+

AlanGriffiths:
> I'm under the impression the problem is in the host, not Mir-on-Wayland (since, as you say, that works on other compositors).

This PR introduces a problem on other compositors. (As I say, "With this PR it fails")

But I'll look into the problem

wmww:
Oh, I misinterpreted what you meant. In that case I don't think this should land, at least without better understanding what's going on.

bors r-

wmww:
Somehow my HiDPI branch ends up fixing it. I don't entirely understand why. On different commits it's broken in different ways, then at the end **bam!** It works.

AlanGriffiths:
I wouldn't be surprised if there's an issue in Mir-on-Wayland. That did need some rescaling logic, which may have been one of [two wrongs that made a right](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch86.html).

AlanGriffiths:
> I wouldn't be surprised if there's an issue in Mir-on-Wayland. That did need some rescaling logic, which may have been one of [two wrongs that made a right](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch86.html).

Specifically, commit 42fb1856461e8eda37f2f57f6b7e1a784bbb4f58 likely needs to be reverted.

PR #1351, opened 2020-03-05
Fix hybrid config chosing

RAOF:
`mgmh::EGLHelper` got a fix for when mesa started supporting `rgba_1010102` visuals, resulting in `eglChooseConfig` picking a config incompatible with our `rgba_8888` GBM buffers.

The hybrid bounce-buffer code did not get updated at the same time, so suffers the same problem.

Fix this by using the `EGLHelper` code in `EGLBufferCopier`. Convenenly, this also eliminates quite a lot of code from `EGLBufferCopier`. Score!"


AlanGriffiths:
bors r=wmww

PR #1354, opened 2020-03-06
Fix WaylandTexBuffer

AlanGriffiths:
Fix WaylandTexBuffer.

Buffers may be consumed by multiple compositors. We must protect mutable state.


AlanGriffiths:
Oops! Random changes were included. Need to fix that.

PR #1360, opened 2020-03-12
Example clients: use configured size instead of output size

wmww:
__problem:__ When rendering example clients (such as the spinner, splash and background), we previously used the size from the current output mode for each output. This is not a robust system, as raw output size is not necessarily the size the surface should be to fill the screen (for example, raw output size does not reflect transforms including scaling).

Some of these transforms can be detected and compensated for, but this results in a complicated and fragile system. Others, such as fractional scaling, can not be compensated for in this way as the protocol does not supply enough information.

__Alternative Solution:__ The `xdg_output_unstable_v1` protocol (which we implement) has a `.logical_size` event. This would work, but would have added the complication of generating/updating a non-core protocol, plus the code to actually use it. I found this to be unnecessary, but worth mentioning here in case my current approach turns out to have problems I'm unaware of.

__Selected Solution:__ Since we're already making the client windows full screen, they're already getting configured with the size the compositor wants them to be. Here, I add the logic to use the sizes from those `wl_surface.configure` events. This should make these components work reliably without modification as the way Mir handles transformations and scaling change.

### DIFF
--- a/examples/example-server-lib/decoration_provider.cpp
+++ b/examples/example-server-lib/decoration_provider.cpp
@@ -94,10 +94,15 @@ Printer::~Printer()
 
 struct BackgroundInfo
 {
-    BackgroundInfo(Output const& output) :
-        output{output}
+    BackgroundInfo(
+        wl_display* display,
+        Output const* output,
+        std::function<void(BackgroundInfo&)> redraw_func)
+        : output{output},
+          display{display},
+          redraw_func{redraw_func}
     {}
-    
+
     ~BackgroundInfo()
     {
         if (buffer)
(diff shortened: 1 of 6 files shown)


AlanGriffiths:
I need to think about how this _should_ work.

For example should things like the internal background should be scale aware, and submit prescaled buffers?

`DecorationProviderClient::draw_background()` already respects WL_OUTPUT_TRANSFORM_XXX, so I don't see why it wouldn't also respect scale. (Except that we don't store the output scale currently.)

AlanGriffiths:
> For example should things like the internal background should be scale aware, and submit prescaled buffers?

Of course, it would help if we supported clients that do that!

```C++
void mf::WlSurface::set_buffer_scale(int32_t scale)
{
    (void)scale;
    // TODO
}
```

wmww:
Yes, they should be aware of scale. How it will work is they will get configured, and then render at a multiple of the size they are configured to be depending on scale. This should be relatively easy to implement.

wmww:
> `DecorationProviderClient::draw_background()` already respects WL_OUTPUT_TRANSFORM_XXX, so I don't see why it wouldn't also respect scale.

It didn't *really* respect the output's transform (it didn't render transformed buffers), it just used the transform in conjunction with the current mode's size to guess how big the surface should be. this logic is made obsolete by respecting the surface's configured size.

It never needs to repect the transform, unless we care a lot about optimizing out rendering on rotated displays. It *does* need to actually respect scale, and making it do so will not be difficult.

Issue #1361, opened 2020-03-12
Handle client buffers with a transform


wmww:
We need to correctly render client buffers that have a transform set. I believe this will require changes all the way through to the renderer.

PR #1362, opened 2020-03-12
mf::Output: don't send output transform

wmww:
We can not yet deal with client supplied buffers (see #1361). Sending output transform is currently not a problem because surfaces do not enter outputs, but as soon as they do some clients will be attaching buffers with transforms and everything will fall apart. If we don't claim any outputs have a transform, no sane clients should send transformed buffers.

This will break example backgrounds and splashes on transformed outputs until #1360 lands.


wmww:
I'd like to note that the implementation we had wasn't complete. When we add it back, we'll presumably also want to support inverted transforms.

PR #1364, opened 2020-03-12
Report the correct output scale to Wayland clients

AlanGriffiths:
Report the correct output scale to Wayland clients


wmww:
Included in #1366

PR #1365, opened 2020-03-12
X11 platform output layout

wmww:
Make X11 platform output layout correctly use logical size instead of physical size


wmww:
bors r=AlanGriffiths

PR #1366, opened 2020-03-13
Support scaled buffers

wmww:
Renders scaled buffers at correct size, and sends correct output scales so clients know to render scaled buffers. Clients will not start sending scaled buffers until #1363 has also landed, as they do not know they're on a scaled output.

### DIFF
--- a/src/gl/tessellation_helpers.cpp
+++ b/src/gl/tessellation_helpers.cpp
@@ -26,7 +26,6 @@ namespace geom = mir::geometry;
 mgl::Primitive mgl::tessellate_renderable_into_rectangle(
     mg::Renderable const& renderable, geom::Displacement const& offset)
 {
-    auto const& buf_size = renderable.buffer()->size();
     auto rect = renderable.screen_position();
     rect.top_left = rect.top_left - offset;
     GLfloat left = rect.top_left.x.as_int();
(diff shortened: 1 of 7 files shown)


src/gl/tessellation_helpers.cpp
```diff
@@ -37,10 +36,8 @@ mgl::Primitive mgl::tessellate_renderable_into_rectangle(
     mgl::Primitive rectangle;
     rectangle.type = GL_TRIANGLE_STRIP;
 
-    GLfloat tex_right = static_cast<GLfloat>(rect.size.width.as_int()) /
-                        buf_size.width.as_int();
-    GLfloat tex_bottom = static_cast<GLfloat>(rect.size.height.as_int()) /
-                         buf_size.height.as_int();
```
wmww: The old logic attempts to keep buffer size the same regardless of the size of the rectangle the buffer is being rendered to. The new logic stretches the buffer to the full are of the rectangle. Scaling is achieved by having a rectangle a different size than the buffer.

wmww:
Failure seemed non-deterministic to me. To be clear, this won't have much effect before #1363, but shouldn't cause problems if it lands before that one.

AlanGriffiths:
We can fix any corner cases separately.

bors r+

PR #1372, opened 2020-03-18
Fix DisplayConfiguration tests

wmww:
`user_display_configuration_output_extents_are_scaled_fractionally` was failing due to a floating point error

### DIFF
--- a/src/platform/graphics/display_configuration.cpp
+++ b/src/platform/graphics/display_configuration.cpp
@@ -242,7 +242,9 @@ mir::geometry::Rectangle extents_of(
     if (current_mode_index >= modes.size())
         return mir::geometry::Rectangle();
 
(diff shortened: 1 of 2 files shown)


src/platform/graphics/display_configuration.cpp
```diff
@@ -242,7 +242,9 @@ mir::geometry::Rectangle extents_of(
     if (current_mode_index >= modes.size())
         return mir::geometry::Rectangle();
 
-    auto const& size = modes[current_mode_index].size * (1.0f / scale);
+    mir::geometry::Size const size{
+        roundf(modes[current_mode_index].size.width.as_int() / scale),
+        roundf(modes[current_mode_index].size.height.as_int() / scale)};
```
AlanGriffiths: Hmm. Maybe we should make rounding a feature of `operator/()` and `operator*()`

PR #1386, opened 2020-03-23
Add a font & location for the "wallpaper" that works on Fedora "rawhide"

AlanGriffiths:
Add a font & location for the "wallpaper" that works on Fedora "rawhide".

We really need a mor flexible approach to choosing a typeface. But this works for now.


AlanGriffiths:
> I would much rather find a way to share code between this and the SSD font path searching

That sounds like a good idea. But let's fix the immediate problem first.

bors r=wmww

PR #1389, opened 2020-03-24
Add missing gl/ subdirectory to unit tests CMake

wmww:
I sorted them alphabetically so I could easily check to make sure there weren't any other missing subdirectories, which there weren't


AlanGriffiths:
Given the test failures, are you sure the tests are useful?

wmww:
@RAOF said the interface the broken tests were testing is deprecated. Only approve after confirming that we no longer need the dropped tests

AlanGriffiths:
> @RAOF said the interface the broken tests were testing is deprecated. Only approve after confirming that we no longer need the dropped tests

Hmm. Deprecated, doesn't mean "unsupported" nor "unused". In particular, it is used by `mrg::Renderer`.

RAOF:
> Hmm. Deprecated, doesn't mean "unsupported" nor "unused". In particular, it is used by mrg::Renderer.

It's only used for buffers which support `gl:: TextureSource` but not `gl::Texture`. Only the `WlShmBuffer` implementation used when a platform doesn't provide a `WaylandAllocator` matches that description in our tree (I don't know about Ubports' Android platform, though)

AlanGriffiths:
> I don't know about Ubports' Android platform, though

@mariogrip, @APokorny does your Android platform use `mrg::Renderer`?

PR #1397, opened 2020-03-30
Fix Fedora rawhide build in CI

AlanGriffiths:
Partial fix for Fedora rawhide build in CI. (Fixes: #1400, fixes: #1399)

1. Drop dependency that's unsatisfiable on Fedora rawhide (I think it's redundant)
2. Add a missing dependency
3. Clear WAYLAND_DISPLAY from the test environment
4. Disable a couple of tests that try to frig with /dev/random

That leaves us with a build, but hanging tests (#1405) but that can be fixed later.

### DIFF
--- a/tests/miral/external_client.cpp
+++ b/tests/miral/external_client.cpp
@@ -93,7 +93,7 @@ TEST_F(ExternalClient, default_app_env_is_as_expected)
 
 TEST_F(ExternalClient, default_app_env_x11_is_as_expected)
 {
-    if (getenv("XDG_RUNTIME_DIR") == nullptr)
+    if (access("/tmp/.X11-unix/", W_OK) == 0)
         return; // Starting an X server on LP builder doesn't work - skip the test
 
     add_server_init(x11);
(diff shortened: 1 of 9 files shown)


AlanGriffiths:
Using this to experiment with Fedora builds

AlanGriffiths:
We can land this now. The rawhide image needs a refresh before that platform passes, but that's underway.

AlanGriffiths:
bors r=wmww

AlanGriffiths:
I've encountered some more test issues that manifest on rawhide. Let's fix those too.

tests/miral/external_client.cpp
```diff
@@ -93,7 +93,7 @@ TEST_F(ExternalClient, default_app_env_is_as_expected)
 
 TEST_F(ExternalClient, default_app_env_x11_is_as_expected)
 {
-    if (getenv("XDG_RUNTIME_DIR") == nullptr)
+    if (access("/tmp/.X11-unix/", W_OK) == 0)
```
RAOF, line 96: This is probably ok, but it would be more principled to have a `MIR_DISABLE_X11_TESTS` CMake option in the same way that we have `MIR_DISABLE_EPOLL_REACTOR`.

AlanGriffiths: Possibly more principled, but it assumes (probably correctly) that the build environment is a predictor for the runtime environment.

PR #1402, opened 2020-04-01
HiDPI related tests

wmww:
Tests renderable tessellation and stream scaling

### DIFF
--- a/tests/unit-tests/gl/test_tessellation_helpers.cpp
+++ b/tests/unit-tests/gl/test_tessellation_helpers.cpp
@@ -27,6 +27,34 @@ namespace mt = mir::test;
 namespace mtd = mir::test::doubles;
 namespace mgl = mir::gl;
 
+struct BoundingBox
+{
+    float left, right, top, bottom;
+
+    static auto from(geom::Rectangle const& rect) -> BoundingBox
+    {
+        return BoundingBox{
+            static_cast<float>(rect.left().as_int()),
+            static_cast<float>(rect.right().as_int()),
+            static_cast<float>(rect.top().as_int()),
+            static_cast<float>(rect.bottom().as_int())};
+    }
+
+    auto operator==(BoundingBox const& other) const -> bool
+    {
+        return left == other.left &&
+               right == other.right &&
+               top == other.top &&
+               bottom == other.bottom;
+    }
+};
+
+auto operator<<(std::ostream& os, const BoundingBox& bb) -> std::ostream&
+{
+    os << "{ left=" << bb.left << ", right=" << bb.right << ", top=" << bb.top << ", bottom=" << bb.bottom << " }";
+    return os;
+}
+
 class Tessellation : public testing::Test
 {
 public:
@@ -35,11 +63,6 @@ public:
         ON_CALL(renderable, screen_position())
             .WillByDefault(Return(rect));
     }
-
-    struct BoundingBox
-    {
-        float left, right, top, bottom;
-    };
 
     static auto bounding_box(mgl::Primitive const& primitive) -> BoundingBox
     {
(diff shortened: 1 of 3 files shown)


wmww:
Let's try that again

tests/unit-tests/gl/test_tessellation_helpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    static void expect_tex_coords_1_or_0(mgl::Primitive const& primitive)
+    {
+        for (int i = 0; i < primitive.nvertices; i++)
+        {
+            for (int axis = 0; axis < 2; axis++)
+            {
+                float const tex_coord = primitive.vertices[i].texcoord[axis];
+                EXPECT_THAT(tex_coord == 0.0 || tex_coord == 1.0, Eq(true));
```
AlanGriffiths, line 70: If this fails, the report will be:
```plain
Value of: tex_coord == 0.0 || tex_coord == 1.0
Expected: is equal to true
  Actual: false (of type bool)
```

It might be helpful to report what `tex_coord` actually is (and axis). Vis:
```c++
        EXPECT_THAT(tex_coord, AnyOf(Eq(1.0), Eq(0.0))) << "..axis=" << 1;
```
```
Value of: tex_coord
Expected: (is equal to 1) or (is equal to 0)
  Actual: 0.5 (of type float)
..axis=1
```

tests/unit-tests/gl/test_tessellation_helpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+    EXPECT_THAT(box.bottom, Eq(rect.bottom().as_int() - y));
+}
+
+TEST_F(Tessellation, vertex_zs_are_1)
+{
+    int const x{3}, y{8};
+    mgl::Primitive const primitive = mgl::tessellate_renderable_into_rectangle(renderable, {x, y});
+    for (int i = 0; i < primitive.nvertices; i++)
+    {
+        EXPECT_THAT(primitive.vertices[i].position[2], Eq(0));
```
AlanGriffiths, line 113: Again, if there's an error, it might be of interest what `i` is.

tests/unit-tests/gl/test_tessellation_helpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+TEST_F(Tessellation, bounding_box_gets_offset)
+{
+    int const x{3}, y{8};
+    mgl::Primitive const primitive = mgl::tessellate_renderable_into_rectangle(renderable, {x, y});
+    auto const box = bounding_box(primitive);
+    EXPECT_THAT(box.left, Eq(rect.top_left.x.as_int() - x));
+    EXPECT_THAT(box.top, Eq(rect.top_left.y.as_int() - y));
+    EXPECT_THAT(box.right, Eq(rect.right().as_int() - x));
+    EXPECT_THAT(box.bottom, Eq(rect.bottom().as_int() - y));
```
AlanGriffiths, line 104: I guess this is OK, but the test would likely read better with a custom matcher for `BoundingBox`. Vis:
```c++
EXPECT_THAT(bounding_box(primitive), BoundingBoxEq(rect.top_left.x.as_int() - x, rect.top_left.y.as_int() - y, rect.right().as_int() - x, rect.bottom().as_int() - y));

tests/unit-tests/gl/test_tessellation_helpers.cpp
wmww: Good idea! I noticed that this wasn't ideal, but I didn't think of the fancier gtest matchers.

wmww:
That should resolve everything. I decided giving bounding box stream and equality operators and comparing with `Eq()` was preferable to a custom matcher.

Issue #892, opened 2019-06-24
[miral] Process launched by ExternalClientLauncher::launch are left <defunct> on exit


AlanGriffiths:
Apart from stray entries in the process table this is unimportant.

It should be avoided by having a `SIGCHLD` handler call `wait()` (is there any value in making the exit status available through the API?)

RAOF:
Yes, probably worth having the exit status available through the API.

PR #1406, opened 2020-04-06
Avoid child processes becoming zombies when they exit

AlanGriffiths:
Avoid child processes becoming zombies when they exit. (Fixes: #892)


src/server/run_mir.cpp
```diff
@@ -102,6 +102,19 @@ void mir::run_mir(
 
     static std::atomic<unsigned int> concurrent_calls{0};
 
+    if (!concurrent_calls)
```
RAOF, line 105: Why is this guarded by `concurrent_calls`? It doesn't guarantee that it's only called once, because it happens before the increment of `concurrent_calls`. Is there any reason not to just do this unconditionally?

PR #1407, opened 2020-04-07
Release/1.7

wmww:
https://wiki.ubuntu.com/Process/Merges/TestPlans/Mir

I will test on at least 20.04 and Arch


wmww:
Once I got everything installed correctly everything seems fine on 20.04. There is an issue when a window is on multiple outputs where popups always appear on the output that more than half of the window is on,but I believe that is preexisting and it isn't a major problem so I'm making a new issue for that.

include/platform/mir/graphics/program_factory.h
```diff
@@ -41,16 +41,23 @@ class ProgramFactory
...
+     *                                  fragment defining a function:
+     *                                  vec4 sample_to_rgba(in vec2 texcoord)
+     *                                  returning a vector containing the RGBA value at texcoord.
+     *                                  The elements of the uniform array tex[n] will be bound to
+     *                                  GL_TEXTURE0, GL_TEXTURE1, …, GL_TEXTURE(n-1). (n <= 8)
+     * \return  A reference to the fully compiled and linked Program
      */
-    virtual std::unique_ptr<Program> compile_fragment_shader(
+    virtual Program& compile_fragment_shader(
+        void* id,
```
wmww, line 60: @AlanGriffiths could this be the thing that causes the ABI break?

AlanGriffiths: That sure looks like an ABI break to me.

wmww:
Closing and switching to 1.8 due to ABI break

PR #1412, opened 2020-04-08
Do not send events to windows after they are released

AlanGriffiths:
[mirclient] Do not send events to windows after they are released. (And update tests to match.)


AlanGriffiths:
> When applied over the Ubuntu Touch fork of Mir 1.2.0, this fixes issues with poorly-behaved apps (and a poorly-behaved DE) receiving window events after the window is destroyed, causing a client segmentation fault.

@wmww could you cherry-pick for 1.8?

wmww:
Will do

PR #1418, opened 2020-04-10
Allow MirAL shells to enable/disable SSD

### LINKED ISSUE #1417: Give shell control over which windows get server-side decorated
The shell should be able to set if a window is server-side decorated. Specifically, Unity8 would like to disable all internal SSD because it already provides them. In the future we will want a new way for shells to easily draw and control decorations, but for now just allowing them to enable/disable Mir's decorations would be an improvement.

wmww:
Allows shells to enable or disable server-side decorations during surface construction. Our tiling window manager now disables all SSDs. Fixes #1417. This isn't necessarily what we want to do or how we want to do it. @mariogrip does it fix your problem?


src/miral/symbols.map
```diff
@@ -497,3 +497,10 @@ global:
     miral::ExternalClientLauncher::launch_using_x11*;
   };
 } MIRAL_2.8;
+
+MIRAL_2.10 {
+global:
+  extern "C++" {
+    miral::WindowSpecification::server_side_decorated*;
+  };
+} MIRAL_2.9;
```
AlanGriffiths, line 506: symbols.map should be updated via src/miral/regenerate-miral-symbols-map.py (or will be overwrritten).

wmww:
bors r=AlanGriffiths

PR #1419, opened 2020-04-14
Fix FTBFS in PPA

AlanGriffiths:
#1397 broke PPA builds, this should fix it


RAOF:
Should we now be making this a CMake option?

AlanGriffiths:
> Should we now be making this a CMake option?

I agree this is ugly, but I don't see that a CMake option is cleaner.

What do you want the CMake option to do? Predict at build time what the runtime environment will be?

Meantime, this seems like the simplest way to fix ppa-mir-team/dev

PR #1426, opened 2020-04-17
rpi-dispmanx: Minimal update to fix CPU-accessible buffers

RAOF:


RAOF:
This apparently got half-done in the CPU-accessible-buffers work. I will (hopefully) have the kinks in the much improved overlay branch worked out soon, but there's no reason for master to be broken in the mean time.

Issue #1405, opened 2020-04-03
On Fedora rawhide tests have a tendency to hang for no obvious reason


AlanGriffiths:
Happens to miral-test, mir_acceptance_tests, mir_integration_tests and probably others.

If you attach gdb to the process, there's nothing obviously wrong, and the test continues without problems if you "continue" or detach (or quit).

I suspect this is the reason for timeouts in CI.

To reproduce:

```plain
while bin/miral-test; do :; done
```
May take a minute or two before it hangs.

AlanGriffiths:
I've found a reason:
```plain
[alan@localhost mir]$ ps -eaf | grep miral
alan       15035   12756  3 10:22 pts/0    00:00:00 cmake-build-debug/bin/miral-test.bin
alan       15278   15228  0 10:23 pts/1    00:00:00 grep --color=auto miral
[alan@localhost mir]$ ps -o state= -p 15035
T
```
            T    stopped by job control signal

But WHY?!

AlanGriffiths:
Well, `kill -SIGCONT <pid>` doesn't restart the test, but closing and opening the lid does.

AlanGriffiths:
> Well, `kill -SIGCONT <pid>` doesn't restart the test, but closing and opening the lid does.

Also working:
```plain
killall -SIGSTOP cmake-build-debug/bin/miral-test.bin; killall -SIGCONT cmake-build-debug/bin/miral-test.bin
```

So, stopping and starting the process also "unhangs" it.

AlanGriffiths:
Something is inconsistent or crazy! Maybe it is me!?
```plain
$ ps -o state=,cmd=,pid= -a | grep miral | grep -v grep
S cmake-build-debug/bin/miral   31508
```
>               S    interruptible sleep (waiting for an event to complete)

I'm glad I pasted the "T" result above. I'd think I dreamt it otherwise.

AlanGriffiths:
Now a simple reproducer:
```plain
while cmake-build-debug/bin/miral-test --gtest_filter=TestServer.connect_client_works; do :; done
```
This (eventually) hangs.
```plain
$ ps -o state= -o command:37= -o pid= -C miral-test.bin
S cmake-build-debug/bin/miral-test.bin    93315
```
There's nothing entirely obvious from attaching `gdb`.
Attaching `strace` and then STOP/START:
```plain
[...]

PR #1427, opened 2020-04-17
The Fedora 30 build is working, so keep it that way

AlanGriffiths:
The Fedora 30 build is working, so keep it that way


AlanGriffiths:
Typical! Now Fedora-30 fails. (The log looks like #1405 - which is intermittent)

PR #1428, opened 2020-04-17
Use const where possible

AlanGriffiths:
Use const where possible


AlanGriffiths:
bors r=wmww

AlanGriffiths:
bors r=wmww

PR #1429, opened 2020-04-17
[mir-shell] Get the user keymap from gsetting

AlanGriffiths:
[mir-shell] Get the user keymap from gsetting.

### DIFF
--- a/examples/miral-shell/desktop/mir-shell.sh
+++ b/examples/miral-shell/desktop/mir-shell.sh
@@ -39,8 +39,14 @@ fi
 if which gsettings > /dev/null
 then
   keymap_index=$(gsettings get org.gnome.desktop.input-sources current | cut -d\  -f 2)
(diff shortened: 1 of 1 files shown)


wmww:
On my Arch install without GNOME shell, `gsettings get org.gnome.desktop.input-sources sources` produces just `@a(ss) []` and so `MIR_SERVER_KEYMAP` will end up an empty string. Will that break anything?

AlanGriffiths:
> On my Arch install without GNOME shell, `gsettings get org.gnome.desktop.input-sources sources` produces just `@a(ss) []` and so `MIR_SERVER_KEYMAP` will end up an empty string. Will that break anything? (it doesn't _seem_ to break anything)

AFAICS using "" for the keymap will just default (likely to the equivalent of "us").

I wonder where Sway, KDE etc. keeps the corresponding setting?

PR #1436, opened 2020-04-22
rpi-dispmanx: Use DispmanX overlays when possible

RAOF:
This provides an immense performance improvement, so Mir is actually useable.

### DIFF
--- a/src/platforms/rpi-dispmanx/CMakeLists.txt
+++ b/src/platforms/rpi-dispmanx/CMakeLists.txt
@@ -12,7 +12,7 @@ add_library(mirplatformgraphicsrpidispmanxobjects OBJECT
   display.h
   display_buffer.cpp
   display_buffer.h
-)
+  helpers.h helpers.cpp)
 
 target_include_directories(
   mirplatformgraphicsrpidispmanxobjects
--- a/src/platforms/rpi-dispmanx/buffer_allocator.cpp
+++ b/src/platforms/rpi-dispmanx/buffer_allocator.cpp
@@ -124,7 +124,7 @@ public:
         std::shared_ptr<mg::common::EGLContextExecutor> egl_executor)
         : ShmBuffer(size, format, std::move(egl_executor)),
           stride_{stride},
-          handle{mg::rpi::dispmanx_resource_for(size, format)}
+          handle{mg::rpi::dispmanx_resource_for(size, stride_, format)}
     {
     }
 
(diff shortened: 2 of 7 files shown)


RAOF:
This should be testable on a Pi by installing the `rpi-userland` snap and building with:
```
PKG_CONFIG_PATH=/usr/lib/arm-linux-gnueabihf/pkgconfig:/snap/rpi-userland/current/opt/vc/lib/pkgconfig cmake $PATH_TO_MIR_SOURCE -DMIR_PLATFORM=rpi-dispmanx
```
Running a server will require setting the `LD_LIBRARY_PATH=/snap/rpi-userland/current/opt/vc/lib/` environment variable. You might run into #1437, among others, while testing.

wmww:
I'm having even more problems then before now. miral-shell doesn't start up at all. mir_demo_server seems to start. The cursor appears and can be moved but the text of the TTY doesn't go away and surfaces don't show up. I'll be online later and we can troubleshoot.

wmww:
I got this to work with the following steps (I'm not saying all of these are optimal or essential, this is just precisely what I did):
- `sudo apt install vim snapd devscripts equivs git`
- Put `gpu_mem=386` at the bottom of `/boot/config.txt` (otherwise you get "allocation failed" and "no memory" type errors)
- Reboot
- `sudo snap install rpi-userland --edge`
- Clone Mir and check out this branch (or master if this has merged)
- Remove "wlcs" from `debian/control`
  - We will thus need to disable tests in cmake (see below)
  - Cloning, building and installing WLCS would presumably also work
- `mk-build-deps -i -s sudo` in Mir repo root
- Make and enter a build directory
- `export PKG_CONFIG_PATH=/usr/lib/arm-linux-gnueabihf/pkgconfig:/snap/rpi-userland/current/opt/vc/lib/pkgconfig` to pick up the stuff from the userland snap
- `cmake .. -DMIR_PLATFORM=rpi-dispmanx -DMIR_ENABLE_TESTS=OFF`
- `make`
- Start Mir by invoking the following script from the pi home directory **with sudo**:
```bash
#MIR_SHELL=mir_demo_server
MIR_SHELL=miral-shell
export XDG_RUNTIME_DIR=/run/user/1000
export LD_LIBRARY_PATH=/snap/rpi-userland/current/opt/vc/lib/:$LD_LIBRARY_PATH
(
    sleep 3
    echo
    echo chowning the display
    chown pi /run/user/1000/wayland-0
    echo
) &
./mir/build/bin/$MIR_SHELL --console-provider=vt
```

PR #1439, opened 2020-04-22
Clear wl_resource implementation on destroy

### LINKED ISSUE #1438: Use-after-free in Wayland frontend
Specifically, if an exception is thrown in the constructor of a Wayland object after the base class has been initialized, The base class is immediately freed but the `wl_resource` has already been created and set up. When the resource is destroyed (likely because the client is being terminated due to the exception that was just caught), it tries to free the C++ Wayland object again. This is the cause of the problem tested in https://github.com/MirServer/wlcs/pull/158#pullrequestreview-398203895.

wmww:
Fixes #1438 (see that for description of the problem). The fix is to unlink the `wl_resource` from the object when the object is destroyed. This should eliminate this and potential future bugs involving wayland objects getting destroyed early.

I've looked at libwayland to verify that:
-  `wl_resource_set_implementation()` is  extremely low cost (just 3 assignments)
- libwayland doesn't assume these values are non-null

### DIFF
--- a/src/wayland/generated/wayland_wrapper.cpp
+++ b/src/wayland/generated/wayland_wrapper.cpp
@@ -80,6 +80,11 @@ mw::Callback::Callback(struct wl_resource* resource, Version<1>)
     }
 }
 
+mw::Callback::~Callback()
+{
+    wl_resource_set_implementation(resource, nullptr, nullptr, nullptr);
+}
+
 void mw::Callback::send_done_event(uint32_t callback_data) const
 {
     wl_resource_post_event(resource, Opcode::done, callback_data);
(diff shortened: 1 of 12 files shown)


RAOF:
bors r=AlanGriffiths,RAOF

wmww:
> Resource doesn't actually own a wl_resource. (Why not?)

I'm under the impression it is you who did not want `Resource` to own a `wl_resouce*`. I don't remember exactly what the reasoning was, and it may have been based on outdated factors. I'm happy to move resource and version into the `Resource` class if that's acceptable.

RAOF:
> > Resource doesn't actually own a wl_resource. (Why not?)
> 
> I'm under the impression it is you who did not want `Resource` to own a `wl_resouce*`. I don't remember exactly what the reasoning was, and it may have been based on outdated factors. I'm happy to move resource and version into the `Resource` class if that's acceptable.

Hm. I wonder what past-me was thinking?

Ah. Looking at #828, I think my expectation was that `Resource` *would* end up owning a `wl_resource*`, or at least we'd end up with *something* that does.

PR #1441, opened 2020-04-23
[miral-shell] Use a better default terminal emulator

AlanGriffiths:
[miral-shell] Use a better default for --shell-terminal-emulator

(It is *possible* that miral-terminal isn't installed, but weston-terminal is even less likely to be found!)


AlanGriffiths:
Was bors asleep?

bors r=wmww

PR #1443, opened 2020-04-24
Allow tests to coexist with existing Wayland compositor

wmww:
Running `miral-test` on sway I get a lot of `unable to lock lockfile /run/user/1000/wayland-0.lock, maybe another compositor is running`, and setting `WAYLAND_DISPLAY` on the command line doesn't help (because the variable is unset, as you can see).


AlanGriffiths:
> Running `miral-test` on sway I get a lot of `unable to lock lockfile /run/user/1000/wayland-0.lock, maybe another compositor is running`, and setting `WAYLAND_DISPLAY` on the command line doesn't help (because the variable is unset, as you can see).

It's unfortunate we don't have a simple way to prevent libwayland spamming, but the message is otherwise harmless.

wmww:
We could set the display based on PID and/or thread ID, but I don't care enough to do that if the message is harmless

PR #1444, opened 2020-04-24
Unhiding window takes focus if active window is same app

wmww:
As discussed with @AlanGriffiths this morning. This fixes a new WLCS test that needs to depend on unmapping/mapping a surface not changing it's stacking order. It also just makes sense.

### DIFF
--- a/tests/miral/active_window.cpp
+++ b/tests/miral/active_window.cpp
@@ -233,7 +233,7 @@ TEST_F(ActiveWindow, a_second_window_hiding_makes_first_active)
     assert_active_window_is(test_name);
 }
 
-TEST_F(ActiveWindow, a_second_window_unhiding_leaves_first_active)
+TEST_F(ActiveWindow, a_second_window_of_same_app_becomes_active_when_unhiding)
 {
     char const* const test_name = __PRETTY_FUNCTION__;
     auto const connection = connect_client(test_name);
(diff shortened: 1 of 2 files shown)


tests/miral/active_window.cpp
```diff
@@ -243,10 +243,26 @@ TEST_F(ActiveWindow, a_second_window_unhiding_leaves_first_active)
...
+    auto const connection_a = connect_client(test_name);
+    auto const connection_b = connect_client(another_name);
+
+    auto const window_a = create_window(connection_a, test_name, sync1);
+    auto const window_b = create_window(connection_b, another_name, sync2);
+
+    sync1.exec([&]{ mir_window_set_state(window_b, mir_window_state_hidden); });
+
+    // Expect this to timeout
+    sync2.exec([&]{ mir_window_set_state(window_b, mir_window_state_restored); });
```
RAOF, line 264: Why would you execute this in `sync2` context if you expect it to timeout? Wouldn't it be clearer just to set the window state directly and then wait?

AlanGriffiths: I think this is fine. It times out if there's no state change (equivalent to waiting) and matches the other tests.

AlanGriffiths:
> This fixes a new WLCS test that needs to depend on unmapping/mapping a surface not changing it's stacking order.

That's a weird constraint. How would that work on \<random Wayland compositor>?

wmww:
Yes, I agree. Merging this may still be a good idea, but the tests should not depend on it. What we need is a way to raise a window from WLCS.

RAOF:
>     2. "A window appearing on the screen gets focus" is reasonable behaviour whether or not the window is new or was previously hidden. (Even more so if the app interacting with the user shows it.)

I disagree here; often a new window appearing should *not* get focus - this is the whole purpose of the `MirCookie` logic that we'll need to port to a Wayland extension sometime.

But that's really just agreeing with Alan on the broader point - wlcs tests cannot in general assume that a newly mapped window will have focus.

wmww:
Clearly this is more complicated than just fixing Mir to be correct, so I've removed any dependence on hide/unhide stacking order from the tests I was writing (means I can't test one thing without adding to the WLCS interface, but that's fine). Leaving PR up in case this change is desirable anyway.

AlanGriffiths:
I've nothing against this PR, but don't see any motivation to make the change just now.

Issue #1447, opened 2020-04-24
WlSurface::frame_callbacks are awkwardly managed shared pointers


wmww:
Our general current approach to Wayland object construction is to `new` a class that inherits from a generated class. Our deletion path is to call `destroy_wayland_object()`, which is a generated function that destroys the `wl_resource` which calls the resource destroyed thunk (also generated) which `delete`s the C++ object.

The Way we handle frame callbacks are different. We create a shared pointer for them. We call `destroy_wayland_object()` on each instance, and then drop the shared pointer. One might think that would result in a double free, but because frame callbacks don't have requests, a destroyed thunk also isn't generated or set, so it seems to work. This was the cause of #1445 

This whole situation feels like a mess. I will think on how it can be improved.

PR #1446, opened 2020-04-24
Revert "Merge #1439"

wmww:
This reverts commit d7d9ffc65e0d9c8f342f6b027e45f4f0a4266299, reversing
changes made to cc924a3b12e443504ce682d8062824f588df245a.

Fixes #1445

Issue #1445, opened 2020-04-24
Clearing resoruce implementations broke master


wmww:
#1439 doesn't appear to be as harmless as I hoped, and it looks like I didn't do enough testing.

To reproduce:
1. Open miral-shell
2. Ctrl+Alt+T a few times (opening any app with do it)
3. Mir crashes with `free(): corrupted unsorted chunks`

No need to investigate. I'm on it.

PR #1448, opened 2020-04-24
Lets try again: clear wl_resource implementation on destroy

### LINKED ISSUE #1438: Use-after-free in Wayland frontend

wmww:
Fixes #1438. The sequel to #1439 because it was broken (#1445) and had to be reverted (#1446). The problem was callbacks were not getting handled correctly, which I have now fixed. The root cause is tied up with messiness around our Wayland object destruction paths as better explained in #1447.

### DIFF
--- a/src/wayland/generated/wayland_wrapper.cpp
+++ b/src/wayland/generated/wayland_wrapper.cpp
@@ -80,6 +80,10 @@ mw::Callback::Callback(struct wl_resource* resource, Version<1>)
     }
 }
 
+mw::Callback::~Callback()
+{
+}
+
 void mw::Callback::send_done_event(uint32_t callback_data) const
 {
     wl_resource_post_event(resource, Opcode::done, callback_data);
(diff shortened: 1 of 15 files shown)


src/wayland/generated/wayland_wrapper.cpp
```diff
@@ -80,6 +80,10 @@ mw::Callback::Callback(struct wl_resource* resource, Version<1>)
     }
 }
 
+mw::Callback::~Callback()
+{
+}
```
wmww, line 85: Specifically note that the generated `~Callback()` is is empty, where as in the previous PR it tried to clear the resource implementation. This is because the deletion path for objects without requests is different.

AlanGriffiths: A shame it's not the more modern C++ `= default`. :)

wmww: I considered that, but this makes the generator implementation moderately more obvious. Since this whole area is convoluted and needs to get refactored, I decided it was good enough.

AlanGriffiths:
Looks OK, but I don't have time to test (more thoroughly than last time) before EOW.

wmww:
That's fine. No rush.

PR #1453, opened 2020-04-30
Split out and refactor xcursor loading

wmww:
We were loading a number of cursors with xcursor but only using the default pointer. We need to load the pointer and set it as the cursor for the root window or else an ugly X will appear as the cursor in some apps (such as GTK apps). This splits the cursor loading logic into its own file, and only loads the default cursor that we need. If we need additional cursors in the future for whatever reason extending it to do more would be easy.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_cursors.cpp
+++ b/src/server/frontend_xwayland/xwayland_cursors.cpp
@@ -25,72 +25,92 @@
 #include <cstring>
 #include <boost/throw_exception.hpp>
 
-namespace
-{
-template<typename T, size_t length>
-constexpr size_t length_of(T(&)[length])
-{
-    return length;
-}
-}
-
-#ifndef ARRAY_LENGTH
-#define ARRAY_LENGTH(a) length_of(a)
-#endif
-
-#define CURSOR_ENTRY(x)        \
-    {                          \
-        (x), ARRAY_LENGTH((x)) \
-    }
-
-static const char *bottom_left_corners[] = {"bottom_left_corner", "sw-resize", "size_bdiag"};
-
-static const char *bottom_right_corners[] = {"bottom_right_corner", "se-resize", "size_fdiag"};
-
-static const char *bottom_sides[] = {"bottom_side", "s-resize", "size_ver"};
-
-static const char *left_ptrs[] = {"left_ptr", "default", "top_left_arrow", "left-arrow"};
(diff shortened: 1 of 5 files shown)


src/server/frontend_xwayland/xwayland_cursors.cpp
```diff
@@ -0,0 +1,192 @@
...
+        {
+            if (formats[i].direct.red_mask != 0xff && formats[i].direct.red_shift != 16)
+            {
+                continue;
+            }
+
+            // if (formats[i].type == XCB_RENDER_PICT_TYPE_DIRECT && formats[i].depth == 24)
+            // {
+            //    result.rgb = formats[i];
+            // }
```
AlanGriffiths, line 107: This comment doesn't add to the code

src/server/frontend_xwayland/xwayland_wm.h
```diff
@@ -113,25 +89,18 @@ class XWaylandWM
...
-    // Cursor
-    xcb_cursor_t xcb_cursor_image_load_cursor(const XcursorImage *img);
-    xcb_cursor_t xcb_cursor_images_load_cursor(const XcursorImages *images);
-    xcb_cursor_t xcb_cursor_library_load_cursor(const char *file);
-
     std::shared_ptr<WaylandConnector> const wayland_connector;
     std::shared_ptr<dispatch::MultiplexingDispatchable> const dispatcher;
     wl_client* const wayland_client;
     std::shared_ptr<XWaylandWMShell> const wm_shell;
+    std::unique_ptr<XWaylandCursors> cursors; ///< Only non-const so it can be reset in the destructor
```
AlanGriffiths, line 96: (Non blocking) "non-const so it can be reset in the destructor" sounds as though we have our life-time management wrong. (Also, on a cursory examination of current `master` I think this is now unnecessary.)

src/server/frontend_xwayland/xwayland_cursors.cpp
wmww: Removed

src/server/frontend_xwayland/xwayland_wm.h
wmww: Correct on both counts. This was working around a life-time management issue, which has now been fixed on master. I rebased this branch on that and made cursors const.

PR #1455, opened 2020-05-01
New sbuild test

RAOF:
We've just had a failure in the PPA due to incomplete packaging landing, let's give this another try…

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ before_install:
 - popd
 
 script:
-- ${SPREAD_PATH}/spread -v ${TASK}
+- travis_wait 30 ${SPREAD_PATH}/spread -v ${TASK}
 
 jobs:
   include:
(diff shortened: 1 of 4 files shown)


AlanGriffiths:
> We've just had a failure in the PPA due to incomplete packaging landing, let's give this another try…

If only we didn't need to maintain packaging in the source tree...

RAOF:
Oh, yeah. This would be a better test if it didn't have merge conflicts!

.travis.yml
```diff
@@ -65,7 +66,7 @@ before_install:
 - popd
 
 script:
-- ${SPREAD_PATH}/spread -v ${TASK}
+- travis_wait 30 ${SPREAD_PATH}/spread -v ${TASK}
```
Saviq, line 69: This shouldn't be necessary? Travis jobs are normally limited to 50 mins, and spread does print the "task running late…" to avoid Travis thinking it got stuck?

RAOF: Yeah, it shouldn't, but for some reason the build got killed. This didn't fix it, though. 😐

RAOF:
This is now only for historical interest; we're now building using sbuild in CI.

PR #1457, opened 2020-05-01
frontend::SurfaceStack: add stacking_order_of()

wmww:
This allows a user of the fronted surface to query the stacking order of a specific set of surfaces. This API will be used by the X11 window manager to inform the X server of the correct window stacking order (which is necessary for correct input dispatch).

### DIFF
--- a/src/include/server/mir/frontend/surface_stack.h
+++ b/src/include/server/mir/frontend/surface_stack.h
@@ -20,12 +20,17 @@
 #define MIR_FRONTEND_SURFACE_STACK_H_
 
 #include <memory>
+#include <set>
+#include <vector>
 
 namespace mir
 {
 namespace scene
 {
 class Observer;
+class Surface;
+using SurfaceSet = std::set<std::weak_ptr<scene::Surface>, std::owner_less<std::weak_ptr<scene::Surface>>>;
+using SurfaceList = std::vector<std::weak_ptr<scene::Surface>>;
 }
 namespace frontend
 {
@@ -37,6 +42,10 @@ public:
 
     virtual void add_observer(std::shared_ptr<scene::Observer> const& observer) = 0;
     virtual void remove_observer(std::weak_ptr<scene::Observer> const& observer) = 0;
+
+    /// Returns the stacking order of the given surfaces, bottom to top
+    /// If a surface is not known to the surface stack it will not be in the returned list
+    virtual auto stacking_order_of(scene::SurfaceSet const& surfaces) const -> scene::SurfaceList = 0;
 
 private:
     SurfaceStack(SurfaceStack const&) = delete;
--- a/src/include/server/mir/shell/surface_stack.h
+++ b/src/include/server/mir/shell/surface_stack.h
@@ -33,6 +33,7 @@ class Surface;
 struct SurfaceCreationParameters;
 class SurfaceObserver;
 class Session;
+using SurfaceSet = std::set<std::weak_ptr<scene::Surface>, std::owner_less<std::weak_ptr<scene::Surface>>>;
 }
 
 namespace shell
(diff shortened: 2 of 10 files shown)


src/include/server/mir/frontend/surface_stack.h
```diff
@@ -38,6 +43,10 @@ class SurfaceStack
     virtual void add_observer(std::shared_ptr<scene::Observer> const& observer) = 0;
     virtual void remove_observer(std::weak_ptr<scene::Observer> const& observer) = 0;
 
+    /// Returns the stacking order of the given surfaces, bottom to top
+    /// If a surface is not known to the surface stack it will not be in the returned list
+    virtual auto stacking_order_of(scene::SurfaceSet const& surfaces) const -> scene::SurfaceList = 0;
```
RAOF, line 48: Performance nit: it seems that there are likely to be very few surfaces in the `SurfaceSet`, which might make the set higher-overhead than just a plain vector?

I'm not sure how this is going to be called, but maybe it should take a `scene::SurfaceList` by value and return a `scene::SurfaceList`?

wmww: I still think a set is a better idea. Here's why:
1. I'm not sure how large a set must be before lookup is faster than linearly searching a vector. I expect the X11 fronted to regularly call this with all X11 surfaces, which could be quite a few.
2. This function should only be called once or twice by different subsystems when surfaces are re-ordered. Not many times each frame. This means if the common case is slightly less performant than optimal, it wont significantly hurt Mir.
3. If we ever *do* need to call this on a large number of surfaces, the cost of incorrectly choosing a vector could be *very* high. The cost of incorrectly choosing a set will always be low. Regardless of what data structure we use, this function will be negligibly fast on small data sizes.
4. The argument is conceptually a set, so giving it that type makes it's nature more obvious to the programmer. This may be worth giving up if the performance gains were significant, but see 1-3.

wmww:
> do we have an invariant that a surface will never appear more than once in the `SurfaceStack`

It doesn't seem to be checked or explicitly stated, but it wouldn't make much sense. A surface can only have on position, so having it at multiple paces in the stacking order would be...weird. If it did happen the current implementation would make it appear multiple times in the returned list, which I think is reasonable. If we ever expect it the behavior of all involved components should be documented.

src/include/server/mir/frontend/surface_stack.h
RAOF: If this *does* get called with lots of surfaces, then we can sort the vector and do a binary search for the same cost as constructing the `std::map` and with the same number of operations for `::find()` and better memory-locality. Depending on how this gets (does the caller maintain a `std::set` of surfaces, and pass that in, or is one constructed for the purposes of this call?) called we can also optimize away one of the memory allocations.

But also this is somewhat of a pedantic sidebar :wink:

wmww: I'm of the opinion this is a language/library level problem. Wanting to collect a large or small number of items and quickly check if an item is among them is literally what `std::set` is built for. If it really is so bad at it's job that we can't use it for that purpose than we shouldn't muddy window stacking logic with workarounds. We should build a new data structure that does the same thing but doesn't suck.

wmww: If we did that it would be practical to do performance tests and determine in what situations our solution is working. As it is, doing actual profiling for every tree set/hash set/vector/sorted vector question is impractical, so we end up debating based on hunches. While your hunches are probably better than mine due to experience, neither are hard data.

wmww: Hmm... my data structure knowledge is somewhat incomplete, but isn't a b-tree based set what we really need? It seems it's incompatible with the STL because iterator invalidation details, but it would have the performance characteristics we want, no? Cache-friendly and as fast as a sorted vector for a small number of elements, and as fast as a binary tree set for a large number of elements.

wmww: Like most problems, I believe "rewrite Mir in Rust" is the optimal solution

RAOF: Yup, `BTreeSet` would be a pretty good candidate. We'll only ever be dealing with many fewer than 1,000 windows, so the asymptotic behaviour isn't terribly interesting for us, but this does indeed give us a better off-the-shelf option.

The `std::set` API makes various promises that we don't care about, so can't perform optimally for us.

Oh, hai! Google searching for “C++ small set” leads one to [Boost.Containers' flat_set](https://www.boost.org/doc/libs/1_73_0/doc/html/container/non_standard_containers.html#container.non_standard_containers.flat_xxx).

We could, if we wanted, drop that in.

PR #1460, opened 2020-05-04
Handle XCB startup errors better

wmww:
We're supposed to check if the XCB connection has errors immediately after creating it, but instead we were initializing the rest of the wm before making the check. If the check failed, the wm was left in-tact with an invalid connection.

With this PR the connection checks itself before doing anything, and throws if there's an error. The exception is caught by `XWaylandServer::connect_wm_to_xwayland()` and treated like other XWayland errors. The logic for retrying a few times and then giving up is already implemented (change `connect_to_fd()` to always throw if you want to test this out).

### DIFF
--- a/src/server/frontend_xwayland/xcb_connection.cpp
+++ b/src/server/frontend_xwayland/xcb_connection.cpp
@@ -117,7 +117,8 @@ mf::XCBConnection::Atom::operator xcb_atom_t() const
 }
 
 mf::XCBConnection::XCBConnection(int fd)
-    : xcb_connection{connect_to_fd(fd)},
+    : fd{fd},
+      xcb_connection{connect_to_fd(fd)},
       xcb_screen{xcb_setup_roots_iterator(xcb_get_setup(xcb_connection)).data},
       atom_name_cache{{XCB_ATOM_NONE, "None/Any"}},
       wm_protocols{"WM_PROTOCOLS", this},
@@ -167,6 +168,7 @@ mf::XCBConnection::XCBConnection(int fd)
 mf::XCBConnection::~XCBConnection()
 {
     xcb_disconnect(xcb_connection);
+    close(fd);
 }
 
 mf::XCBConnection::operator xcb_connection_t*() const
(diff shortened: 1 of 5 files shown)


src/server/frontend_xwayland/xcb_connection.cpp
```diff
@@ -69,6 +69,21 @@ auto number_to_readable_name(uint32_t number) -> std::string
...
 }
+
+auto connect_to_fd(int fd) -> xcb_connection_t*
+{
+    xcb_connection_t* connection = xcb_connect_to_fd(fd, nullptr);
+    if (xcb_connection_has_error(connection))
+    {
+        xcb_disconnect(connection);
+        close(fd);
+        BOOST_THROW_EXCEPTION(std::runtime_error("xcb_connect_to_fd() failed"));
```
RAOF, line 80: It would probably be helpful for future debugging to distinguish between the various forms of error [this can return](https://xcb.freedesktop.org/manual/group__XCB__Core__API.html#ga70a6bade94bd2824db552abcf5fbdbe3).

Non-blocking, though.

wmww:
Improved error reporting, and also moved and clarified ownership of the FD

src/server/frontend_xwayland/xcb_connection.h
```diff
@@ -89,6 +90,7 @@ class XCBConnection
         std::atomic<xcb_atom_t> mutable atom{XCB_ATOM_NONE};
     };
 
+    /// Takes ownership of the given FD
```
AlanGriffiths, line 93: I know this is pre-existing, but we _ought_ to have a way to document this in code. (A "unique" variant of the "shared" `mir::Fd`, say.)

AlanGriffiths:
Sure.

bors r+

src/server/frontend_xwayland/xcb_connection.h
wmww: I believe `mir::Fd` is sufficient. The only reason we need to document ownership is so we know who should close it. I believe in another branch I'm already refactoring this to use a Mir FD, which you'll see at some point.

AlanGriffiths: Maybe shared ownership it is sufficient here. But since C++ acquired proper move semantics representing unique, transferable ownership is also meaningful.

PR #1473, opened 2020-05-12
More XWaylandWM cleanup

wmww:
Make things const and all that good stuff. Stacked on #1453


AlanGriffiths:
Can we rebase on master to make the diff clear for reviewing?

Issue #1472, opened 2020-05-11
XWaylandWM::handle_events() repeatedly called with no events


wmww:
Reproducing this one is tricky. Best I can say is open a bunch of different XWayland apps and hope for the best. Usually when trying to open a window it will fail to open, and the XWayland event loop will start consuming a whole CPU core. `XWaylandWM::handle_events()` gets called constantly with no events. Observed everywhere I've tried (Arch, 20.04, Mir-on-X and Intel/KMS)

wmww:
This goes at least as far back as 72635321eb782807d7a22107779a71e29e0fc20e, indicating it wasn't introduced in an of my recent XWayland shuffling.

wmww:
Ah! I'm guessing `connection->has_error` is not supposed to be 1. That seems like a good place to start.

wmww:
Digging in with GDB turned up nothing except that `_xcb_in_read()` was shutting down the connection with `XCB_CONN_ERROR` for some reason. Bisecting took a while, but was slightly more useful. Looks like the problem was introduced with the abstract socket split: b4fe4784fdd70fe02a5f98ac93cc58f60e96a4df. Still no idea *why*.

PR #1474, opened 2020-05-13
Stack X11 surfaces correctly on the XWayland server

wmww:
If the XWayland server is not told the correct surface stacking order, the desktop will look correct but input will go to the wrong surfaces. Our current solution is to raise a surface when it gains focus, but this is  a bug waiting to happen when someone uses X apps on a desktop that, for example, has focus follow pointer.

To get the correct stacking order we need to plumb the `frontend::SurfaceStack` all the way through to the XWayland window manager, so it can add a scene observer. The wm also has to maintain a map of scene surfaces to XCB windows, and methods are added to allow XWayland surfaces to update that mapping when scene surfaces are created or destroyed. A set of scene surfaces is also added, which is not strictly necessary but prevents us from having to build and destroy it every time we want to call `surface_stack->stacking_order_of()`

### DIFF
--- a/src/server/frontend_wayland/wayland_connector.cpp
+++ b/src/server/frontend_wayland/wayland_connector.cpp
@@ -659,6 +659,7 @@ mf::WaylandConnector::WaylandConnector(
     std::shared_ptr<mi::Seat> const& seat,
     std::shared_ptr<mg::GraphicBufferAllocator> const& allocator,
     std::shared_ptr<mf::SessionAuthorizer> const& session_authorizer,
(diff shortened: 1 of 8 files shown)


AlanGriffiths:
Looks sane, but after opening a few xterm windows to experiment something weird happened to the X subsystem (starting new X11 apps didn't work, closing the existing ones didn't work, ...

Not seen that before applying this, so taking a deeper look.

wmww:
Possibly #1472. Since I observed it on master as well I assumed there wasn't a problem with this PR, but this might make it worse. I'm currently working on it, but I don't have any good leads.

PR #1483, opened 2020-05-20
Fall back to capnproto if capnproto:native is unavailable

AlanGriffiths:
Fall back to capnproto if capnproto:native is unavailable.

Maybe this will work in both CI and PPA.


Saviq:
¯\_(ツ)_/¯

bors merge

PR #1485, opened 2020-05-20
mesa-kms: Handle EOPNOTSUPP from drmModeGetResources.

RAOF:
DRM will return `EOPNOTSUPP` when attempting to access the KMS APIs on a device
node which does not support KMS.

We can hit this if `drmCheckModesettingSupported` is inconclusve (because, for example,
we're on an ARM device where the GPU has no PCI address).

This is the case on the Raspberry Pi 4. `drmCheckModesettingSupported` is inconclusive,
because of course it is, but the V3D DRM node does not support modesetting.

### DIFF
--- a/src/platforms/mesa/server/kms/real_kms_output_container.cpp
+++ b/src/platforms/mesa/server/kms/real_kms_output_container.cpp
@@ -41,12 +41,23 @@ void mgm::RealKMSOutputContainer::update_from_hardware_state()
 {
     decltype(outputs) new_outputs;
 
+    // TODO: Accumulate errors and present them all.
+    std::exception_ptr last_error;
+
     for (auto drm_fd : drm_fds)
     {
-        kms::DRMModeResources resources{drm_fd};
+        std::unique_ptr<kms::DRMModeResources> resources;
+        try
+        {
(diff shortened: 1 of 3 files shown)


src/platforms/mesa/server/kms/real_kms_output_container.cpp
```diff
@@ -41,12 +41,23 @@ void mgm::RealKMSOutputContainer::update_from_hardware_state()
...
-        kms::DRMModeResources resources{drm_fd};
-
+        std::unique_ptr<kms::DRMModeResources> resources;
+        try
+        {
+            resources = std::make_unique<kms::DRMModeResources>(drm_fd);
+        }
+        catch (std::exception const&)
+        {
+            last_error = std::current_exception();
```
AlanGriffiths, line 56: Do we need to "accumulate" errors? Can't we just log failures? Vis:
```c++
            log(Severity::informational, MIR_LOG_COMPONENT, last_error, "Node does not support KMS");
```

RAOF: This is called relatively frequently, so I think it would be unnecessarily noisy to log each time? Maybe?

RAOF:
bors r=AlanGriffiths

PR #1486, opened 2020-05-26
mesa-kms: Remove client platform plugin.

RAOF:
This is only interesting for accelerated libmirclient users, and we haven't had
the necessary Mesa support for accelerated libmirclient users for *some time*!


AlanGriffiths:
> we haven't had
> the necessary Mesa support for accelerated libmirclient users for _some time_!

Actually, we've supported them on 16.04 for 1.x. But I agree we don't need them for 2.x.

PR #1487, opened 2020-05-27
More XWayland cleanup

wmww:
Please note that I remove a `MultiplexingDispatchable` that we don't appear to need. If there is any purpose for the indirection, let me know.


AlanGriffiths:
bors r=RAOF

PR #1488, opened 2020-05-28
mir_demo_client_wayland_egl_spinner: Return a value from main()

RAOF:
For some reason nothing complained that we didn't have a `return` statement
in a function declared as returning an int?!


AlanGriffiths:
I think frustration has made you momentarily forget about `main()`. But this is harmless.

bors r+

RAOF:
Huh, not actually frustration. I didn't know that C++ special-cased `main()` in that way!

PR #1496, opened 2020-06-01
Tests: Make LTTNG fork()-safe

### LINKED ISSUE #1405: On Fedora rawhide tests have a tendency to hang for no obvious reason

RAOF:
[LTTNG documentation](https://lttng.org/man/3/lttng-ust/v2.12/#doc-_using_lttng_ust_with_daemons) makes it clear that using `fork()` without `exec()` in applications linked to `liblttng-ust` is unsafe unless you preload `liblttng-ust-fork.so`. Do so! (Fixes #1405)

This fixes a test failure where running `ThreadedDispatcherSignalTest` immediately after the `Logind` tests would fail.


RAOF:
(Stacked on #1495, because I need that to fix the build)

AlanGriffiths:
/me sees a possible straw: I wonder if this fixes  #1405...

PR #1498, opened 2020-06-02
cmake: Handle tests which also set LD_PRELOAD in their environment.

RAOF:
The umockdev tests work by LD_PRELOADing libumockdev-preload.so.0. Unfortunately, the
LD_PRELOAD of liblttng-ust-fork.so overrides this, causing the tests to fail.

Instead coalesce all the LD_PRELOADs in the environment into a single call, so
both libumockdev-preload.so.0 and liblttng-ust-fork.so get loaded.

### DIFF
--- a/cmake/MirCommon.cmake
+++ b/cmake/MirCommon.cmake
@@ -125,7 +125,7 @@ function (mir_discover_tests_internal EXECUTABLE TEST_ENV_OPTIONS DETECT_FD_LEAK
     endforeach()
 
     # Now construct a single LD_PRELOAD=first:second:…:last string
-    list(JOIN preload_values ":" coalesced_values)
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
Whoops! This appears to break `ptest`

AlanGriffiths:
`ctest` (as documented) only processed the first `ENVIRONMENT` property, so re-ordering LD_PRELOAD to the end broke stuff we noticed. Now we actually set all the `--env`settings.

AlanGriffiths:
Now we just have 18.04 (CMake-3.10) to deal with...
```plain
CMake Error at cmake/MirCommon.cmake:128 (list):
  list does not recognize sub-command JOIN
```

Saviq:
And good to me :)

bors merge

PR #1508, opened 2020-06-08
XWaylandWM::handle_events(): make sure there is no XCB error

wmww:
Handling fatal XCB errors is trickier because we don't have the infrastructure to stop the X server. Both the current code and this PR will result in fatal Mir errors, but at least here the problem is made obvious.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm.cpp
@@ -362,6 +362,8 @@ void mf::XWaylandWM::handle_events()
 {
     bool got_events = false;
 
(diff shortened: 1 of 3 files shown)


src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -362,6 +362,8 @@ void mf::XWaylandWM::handle_events()
 {
     bool got_events = false;
 
+    connection->verify_not_in_error_state();
```
RAOF, line 365: Not blocking, but I'd generally find it more obvious to just not *call* `handle_events` when the connection has errored.

PR #1513, opened 2020-06-09
Add wayland::Handle (.with() interface)

wmww:
Stacked on #1512 

Alternative to #1514 

Makes the destroyed flag part of `wayland::Resource` and adds a `wayland::Handle` class which uses it to allow objects to hold a weak handle to a Wayland object.

`std::shared_ptr<bool>`s are kept around so that `wayland::Resource` can manage them. If instead they were replaced by some sort of strongly typed pointer, the implementation would be messier and every derived class would need to be modified. With this system, new code should only use the destroyed flag indirectly via `wayland::Handle`.

This PR has one of two possible interfaces for `wayland::Handle`. This one has `.with()` and `.with_opt()` methods to access the underlying class. The other has an optional-like interface. I think the other is cleaner than this one, but I'm not going to put up a fight either way.

[this commit](https://github.com/MirServer/mir/pull/1514/commits/55527a84cf031a79c5dcab22c41fb804e2d5e85f) shows the difference between the two interfaces.

### DIFF
--- a/include/wayland/mir/wayland/wayland_base.h
+++ b/include/wayland/mir/wayland/wayland_base.h
@@ -19,6 +19,8 @@
 #ifndef MIR_WAYLAND_OBJECT_H_
 #define MIR_WAYLAND_OBJECT_H_
 
+#include <memory>
+
 struct wl_resource;
 struct wl_global;
 struct wl_client;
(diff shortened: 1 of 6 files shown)


AlanGriffiths:
As discussed f2f I'd prefer something along the lines of:
```c++
    with(possible_foo1, 
        [](auto foo1)
        {
            // code */
        });
```
with the possible extension to...
```c++
    with(possible_foo1, possible_foo2,
        [](auto foo1, auto foo2)
        {
            // code
        });
```
Although that is simple to add after #1514

PS The same naming comment made in #1514 applies to `Handle`: it's a bad name as it doesn't communicate intent.

wmww:
Alternative merged

PR #1514, opened 2020-06-09
Add wayland::Weak (optional-like interface)

wmww:
Stacked on #1512 

Alternative to #1513 

Makes the destroyed flag part of `wayland::Resource` and adds a `wayland::Weak` class which uses it to allow objects to hold a weak reference to a Wayland object.

`std::shared_ptr<bool>`s are kept around so that `wayland::Resource` can manage them. If instead they were replaced by some sort of strongly typed pointer, the implementation would be messier and every derived class would need to be modified. With this system, new code should only use the destroyed flag indirectly via `wayland::Weak`.

This PR has one of two possible interfaces for `wayland::Weak`. This one has an optional-like interface. The other one has `.with()` and `.with_opt()` methods to access the underlying class. I think this one is cleaner, but I'm not going to put up a fight either way.

[this commit](https://github.com/MirServer/mir/pull/1514/commits/55527a84cf031a79c5dcab22c41fb804e2d5e85f) shows the difference between the two interfaces.

### DIFF
--- a/include/wayland/mir/wayland/wayland_base.h
+++ b/include/wayland/mir/wayland/wayland_base.h
@@ -22,6 +22,7 @@
 #include <boost/throw_exception.hpp>
 
 #include <memory>
+#include <stdexcept>
 
 struct wl_resource;
 struct wl_global;
--- a/include/wayland/mir/wayland/wayland_base.h
+++ b/include/wayland/mir/wayland/wayland_base.h
@@ -19,6 +19,8 @@
 #ifndef MIR_WAYLAND_OBJECT_H_
 #define MIR_WAYLAND_OBJECT_H_
 
+#include <boost/throw_exception.hpp>
+
 #include <memory>
 
 struct wl_resource;
@@ -49,6 +51,51 @@ public:
 private:
     std::shared_ptr<bool> mutable destroyed;
 };
+
+/// A weak handle to a Wayland resource
+/// May only be safely used from the Wayland thread
+template<typename T>
+class Weak
+{
+public:
+    Weak(T* resource)
+        : resource{resource},
+          destroyed_flag{resource->destroyed_flag()}
+    {
+    }
+
+    Weak(Weak<T> const&) = default;
+    auto operator=(Weak<T> const&) -> Weak<T>& = default;
+
+    auto operator==(Weak<T> const& other) const -> bool
+    {
+        return resource == other->resource;
+    }
+
+    operator bool() const
+    {
+        return !*destroyed_flag;
+    }
+
+    auto value() const -> T&
+    {
+        if (*destroyed_flag)
+        {
+            BOOST_THROW_EXCEPTION(std::runtime_error("Attempted access of destroyed Wayland resource"));
+        }
+        return *resource;
+    }
+
+private:
+    T* resource;
+    std::shared_ptr<bool> destroyed_flag;
+};
+
+template<typename T>
+auto make_weak(T* resource) -> Weak<T>
+{
+    return Weak<T>{resource};
+}
 
 class Global
 {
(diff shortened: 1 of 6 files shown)


RAOF:
So, there's already a similar (but different) almost implementation in `buffer_from_wl_shm.cpp`, in the form of `SharedWlBuffer`/`SharedWlBuffer::LockedHandle`/`WlResource`.

`wayland::Handle` is not sufficient for the use-case that `SharedWlBuffer` is used for - that needs to work across threads, not just on the Wayland main loop.

`SharedWlBuffer` is obviously not quite generic enough for this task - it's specialised for sending `BUFFER_RELEASE` events at the appropriate time - but could be *made* generic reasonably easily. I'm not sure if it should be, or whether we should have a specialised `SharedWlBuffer` interface as well as a generic `wayland::Handle` interface.

RAOF:
Oh, yeah. As I said, `SharedWlBuffer` cannot be implemented in terms of `wayland::Handle`.

But the core of `SharedWlBuffer` *does* solve the problem that `wayland::Handle` solves; `SharedWlBuffer::lock()` provides the guarantees that `wayland::Handle::operator bool`/`wayland::Handle::value()` do (just a bit more, as well).

(Let's call a de-specialised `SharedWlBuffer` a `SharedWlResource` for clarity)

In some ways `wayland::Handle` is just an optimisation of `SharedWlResource`; is there anywhere where `wayland::Handle` could be used that `SharedWlResource` could not?

As for the overhead, the cost of an uncontended mutex lock/unlock is approximately the same as the cost of an atomic increment/decrement, which is not obviously going to be worse than the memory dereference required for the `destroyed` flag.

I think the relevant question is whether it's *clearer* to have both `wayland::Handle` for non-threadsafe access and `SharedWlResource` for threadsafe access, or just have one.

wmww:
I it's nice to be very clear when you are definitely only working with objects from a single thread. The complexity of a threadsafe handle can't be entirely contained within the object itself. You need locked handles. You have to think about deadlocks. I've come to the conclusion `wayland::Handle` is easier to work with than a threadsafe interface could be.

include/wayland/mir/wayland/wayland_base.h
```diff
@@ -37,12 +41,62 @@ class Resource
...
+    operator bool() const
+    {
+        return !*destroyed_flag;
+    }
+
+    auto value() const -> T&
+    {
+        if (*destroyed_flag)
+        {
+            fatal_error("Attempted to access destroyed Wayland resource");
```
RAOF, line 84: Does this need to be a `fatal_error`? Throwing an exception should result in it being caught and the associated client disconnected with an implementation error.

I'm not blocking on this, just flagging the question.

wmww: I was on the fence about which to make it. I think an exception might be better.

AlanGriffiths:
>  I'd like @AlanGriffiths to weigh i

Sorry, yesterday I was distracted by snapd/desktop-launch. Today I'm distracted by the (lack of) technology set up for https://www.eventbrite.co.uk/e/linuxing-in-london-linux-what-use-is-a-confined-user-shell-tickets-108047398678.

I will get to it soon

include/wayland/mir/wayland/wayland_base.h
```diff
@@ -37,12 +41,62 @@ class Resource
...
+    auto destroyed_flag() const -> std::shared_ptr<bool>;
+
+private:
+    std::shared_ptr<bool> mutable destroyed;
+};
+
+/// A weak handle to a Wayland resource
+/// May only be safely used from the Wayland thread
+template<typename T>
+class Handle
```
AlanGriffiths, line 58: Umm. When I said "use a handle class" I didn't mean "use a class called "Handle"".

The interface and use is closer to the optional idiom than to a smart pointer or handle/body. But as this indirection is about lifetime, maybe "Ephemeral<>"?

PR #1521, opened 2020-06-12
(when verbose logging) warn on X11 errors

wmww:
Error logging is hidden behind the verbose X11 flag because there are a number of errors in normal use and showing them isn't particularly useful. Still something we should have for debugging purposes.


wmww:
It appears my inexplicable X11 errors were actually a result of a missing `break` that Clang failed to warn about (but GCC in CI did). After I fixed that, I'm getting a lot less errors.

PR #1522, opened 2020-06-15
Modernise `GraphicsBufferAllocator`

RAOF:
The server no longer needs a way to allocate “hardware” buffers for its own use,
and we always want *clients* to be able to allocate hardware buffers so it no
longer makes sense for `WaylandAllocator` to be an optional interface.

Fold `WaylandAllocator` into `GraphicsBufferAllocator`, and clean up after it.


RAOF:
Huh. So I rebased this onto master and now the tests pass locally, so :woman_shrugging:

wmww:
Still nope

bors delegate+

AlanGriffiths:
Progress: Just the rpi build now.

RAOF:
…and there's RPi done.

bors r=wmww

PR #1524, opened 2020-06-17
Delete copy constructors and asignment operators for XWayland classes

wmww:
This wasn't causing any problems, but I noticed we weren't doing it. I also removed the explicit delete of some move constructors, as I believe they are deleted by default when the copy constructor is.

### DIFF
--- a/src/server/frontend_xwayland/xcb_connection.h
+++ b/src/server/frontend_xwayland/xcb_connection.h
@@ -78,7 +78,6 @@ public:
 
     private:
         Atom(Atom&) = delete;
-        Atom(Atom&&) = delete;
         Atom& operator=(Atom&) = delete;
 
         XCBConnection* const connection;
@@ -249,7 +248,6 @@ public:
 
 private:
     XCBConnection(XCBConnection&) = delete;
-    XCBConnection(XCBConnection&&) = delete;
     XCBConnection& operator=(XCBConnection&) = delete;
 
     auto xcb_type_atom(XCBType type) const -> xcb_atom_t;
--- a/src/server/frontend_xwayland/xwayland_cursors.h
+++ b/src/server/frontend_xwayland/xwayland_cursors.h
@@ -39,6 +39,9 @@ public:
     void apply_default_to(xcb_window_t window) const;
 
 private:
+    XWaylandCursors(XWaylandCursors const&) = delete;
+    XWaylandCursors& operator=(XWaylandCursors const&) = delete;
+
     struct Cursor
     {
         Cursor(std::shared_ptr<XCBConnection> const& connection, xcb_cursor_t xcb_cursor);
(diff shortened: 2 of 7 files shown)


src/server/frontend_xwayland/xcb_connection.h
```diff
@@ -78,7 +78,6 @@ class XCBConnection
 
     private:
         Atom(Atom&) = delete;
-        Atom(Atom&&) = delete;
```
RAOF, line 81: Any particular reason you've added (or at least, un-removed) the move constructor here?

wmww: According to [this SO answer](https://stackoverflow.com/a/23771245) an explicitly deleted copy constructor counts as a user-declared copy constructor. This is supported by [this Compiler Explorer code](https://gcc.godbolt.org/z/GU8cSd) (note how it starts compiling when you comment out the deleted copy constructor).

AlanGriffiths:
I agree with the deletions. It is almost never necessary to delete move constructors, the language rules are quite conservative about generating them.

But what is the purpose of the additions? You're creating deleted, user declared CopyAssign methods where none previously existed.

wmww:
> But what is the purpose of the additions? You're creating deleted, user declared CopyAssign methods where none previously existed.

Err, yeah? Keep from accidentally copying the classes. We do this all over the place, and it's in the style guild.

AlanGriffiths:
> > 
> 
> Err, yeah? Keep from accidentally copying the classes. We do this all over the place, and it's in the style guild.

I can see the style guilde is not very clear on this. The principle place we explicitly delete CopyAssign methods is in interface classes (which normally should not have copy semantics).

The classes you're changing are not interfaces, and don't have copy semantics.

Before: there are no CopyAssign methods => the classes can't be copied.
After: there are deleted CopyAssign methods => the classes can't be copied.

wmww:
I moved the deletes to interfaces where possible. Since all of these classes do and always will have `const` members, I removed the assignment op delete. I'm not confident an erroneous implicit copy constructor will never be possible (in fact, it definitly will be possible in a future version of `XWaylandServer`), so I'm keeping those. IMO the user shouldn't have to pay attention to when the last `unique_ptr` is removed from a class and delete the copy constructor only then.

AlanGriffiths:
So the reason for this is to ensure future changes don't accidentally enable unwanted copy semantics? That's not unreasonable, I just didn't know what you wanted to achieve here.

I'm now confused because you're apparently fine with the same hypothetical changes enabling unwanted assign semantics? It is hard to construct a case where assigning makes sense while copying doesn't.

As a rule of thumb, copy and assign semantics should be consistent (as should move construction and move assignment).

wmww:
Deleting assignment made explicit

PR #1526, opened 2020-06-17
Don't try to start XWayland multiple times if it fails

wmww:
This auto-restart functionality is complicated and mostly useless. We can do without it.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -107,62 +107,45 @@ void mf::XWaylandServer::spawn()
     enum { server, client, size };
     int wl_client_fd[size], wm_fd[size];
 
-    int xserver_spawn_tries = 0;
     std::unique_lock<decltype(spawn_thread_mutex)> lock{spawn_thread_mutex};
 
-    do
+    spawn_thread_xserver_status = STARTING;
+
+    if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, wl_client_fd) < 0)
     {
-        // Try 5 times
-        if (xserver_spawn_tries >= 5) {
-          mir::log_error("Xwayland failed to start 5 times in a row, disabling Xserver");
-          return;
-        }
-        spawn_thread_xserver_status = STARTING;
-        xserver_spawn_tries++;
-
-        if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, wl_client_fd) < 0)
-        {
-            // "Shouldn't happen" but continuing is weird.
-            mir::fatal_error("wl connection socketpair failed");
-            return; // doesn't reach here
-        }
-
-        if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, wm_fd) < 0)
-        {
-            // "Shouldn't happen" but continuing is weird.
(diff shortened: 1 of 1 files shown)


src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -107,62 +107,45 @@ void mf::XWaylandServer::spawn()
...
-        {
-            // "Shouldn't happen" but continuing is weird.
-            mir::fatal_error("wm fd socketpair failed");
-            return; // doesn't reach here
-        }
+    if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, wl_client_fd) < 0)
+    {
+        // "Shouldn't happen" but continuing is weird.
+        mir::fatal_error("wl connection socketpair failed");
+        return; // doesn't reach here
```
RAOF, line 118: We should mark `mir::fatal_error` as `[[noreturn]]`!

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -107,62 +107,45 @@ void mf::XWaylandServer::spawn()
...
-            }
+    mir::log_info("Starting Xwayland");
+    spawn_thread_pid = fork();
 
-            break;
-        }
+    switch (spawn_thread_pid)
+    {
+    case -1:
+        mir::fatal_error("Failed to fork");
```
RAOF, line 134: Maybe we could include errno?

Or, in fact, just throw a `std::system_error`?

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -107,62 +107,45 @@ void mf::XWaylandServer::spawn()
...
+        return; // Keep compiler happy (doesn't reach here)
+
+    case 0:
+        close(wl_client_fd[server]);
+        close(wm_fd[server]);
+        execl_xwayland(wl_client_fd[client], wm_fd[client]);
```
RAOF, line 140: Should mark `execl_xwayland` as `[[noreturn]]` :grin:

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -107,62 +107,45 @@ void mf::XWaylandServer::spawn()
...
-        if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, wm_fd) < 0)
-        {
-            // "Shouldn't happen" but continuing is weird.
-            mir::fatal_error("wm fd socketpair failed");
-            return; // doesn't reach here
-        }
+    if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, wl_client_fd) < 0)
+    {
+        // "Shouldn't happen" but continuing is weird.
+        mir::fatal_error("wl connection socketpair failed");
```
RAOF, line 117: General question. Should this be a `mir::fatal_error`? When should something be a `mir::fatal_error`, and when should something just result in an exception being thrown?

*My* understanding is that the existence of `mir::fatal_error` is to improve debugging - we want to hit the `abort()` and get the backtrace without any other cleanup. I don't think that applies here; if `socketpair` fails then it's obvious from `errno` (which we're throwing away here :stuck_out_tongue_winking_eye: ).

So I have a weak preference for just throwing a `std::system_error` here.

RAOF:
Oops, I mean `bors delegate`. Oh, well.

PR #1528, opened 2020-06-18
Forget Xenial

AlanGriffiths:
We're no longer trying to build for Xenial, so drop a bunch of workarounds.


AlanGriffiths:
bors r=RAOF

Issue #1525, opened 2020-06-17
FTBFS on armhf & ppc64el (ppa:mir-team/dev)


AlanGriffiths:
## PPA build is failing on groovy, focal and eoan

```plain
cd /<<BUILDDIR>>/mir-1.8.0+dev939-g3769ccdf7d/build-ppc64el/benchmarks/frame-uniformity && /usr/bin/cmake -E cmake_link_script CMakeFiles/frame_uniformity_test_client.dir/link.txt --verbose=1
[...]

PR #1529, opened 2020-06-18
Default to ld linker

AlanGriffiths:
Default to ld linker. (Fixes: #1525)

There's no longer a big speed advantage over ld to gold, and the latter is failing on several targets.

Let's see if this works in the PPA


AlanGriffiths:
Landed without waiting for CI to recover from Fedora issues to see how PPAs respond.

PR #1532, opened 2020-06-19
Make the X11 platform its own top-level thing

RAOF:
With the new `GraphicsBufferAllocator` interface there's no longer any need to tie the X11 platform to be tied to Mesa (or, really, GBM).


debian/control
```diff
@@ -266,19 +266,19 @@ Description: Display server for Ubuntu - shared library
  Contains the shared libraries required for the Mir server and client.
 
 # Longer-term these drivers should move out-of-tree
-Package: mir-platform-graphics-mesa-x16
+Package: mir-platform-graphics-x16
```
AlanGriffiths, line 269: We probably want to Break and/or Replace mir-platform-graphics-mesa-x16

(Although should we really still be on 16? There ought to be incompatibilities with 1.8)

RAOF:
> Looks sane, and works for me. (For a flexible value of "works")
> 
> We still need to work on the automated platform selection, my laptop preferred both `eglstream-kms` and `mesa-kms` to `x11`, only when I forced the X11 option did miral-shell start up. (This is still an improvement, so I won't block landing for this.)

Huh! This correctly autoselects X11 on my systems. I wonder what's different in your setup.

debian/control
RAOF: There's no good reason to automatically remove `mir-platform-graphics-mesa-x16` - it will become eligible for autoremoval.

It should, indeed, be x17. The BufferAllocator PR bumped mirplatform ABI, but must have missed this.

RAOF:
CI failure is Fedora failing at networking.

bors r=AlanGriffiths

debian/control
AlanGriffiths: It was less about autoremoval than about probing both platforms if they are both installed. (Which can happen.) But if we bump the ABI we can only pick the right one.

PR #1536, opened 2020-06-19
Fix ExternalClientLauncher's API

AlanGriffiths:
Fix ExternalClientLauncher's API


AlanGriffiths:
> Hmm... it would seem to me there's no way to use this API safely (such that it's not vulnerable to intermittent issues related to recycled PIDs). Surely there's a better way.

Maybe drop the warning? It's exactly the same with `fork()`.

wmww:
If it's never going to fail then removing the warning would be fine. It looks like Linux takes quite a while before reusing PIDs so I guess this is safe in practice

include/miral/miral/external_client.h
```diff
@@ -37,18 +37,16 @@ class ExternalClientLauncher
 
     void operator()(mir::Server& server);
 
-    void launch(std::vector<std::string> const& command_line) const;
+    /// Launch Wayland and X11 support (if enabled).
+    /// \return The pid of the latest process that was launched or -1.
```
RAOF, line 41: Under what circumstances will this return `-1`? `launch_app_env` throws if `fork()` returns an error (although it should be throwing a `std::system_error` rather than a bare `std::runtime_error`), and that's the only way this can error?

RAOF:
Continues to LGTM.

bors r+

Issue #1518, opened 2020-06-12
Upgrading from Mir 1.x to master isn't smooth


AlanGriffiths:
Given, for example, `libmiral-dev` installed (from the archive or the release PPA)...

Adding ppa:mir-team/dev and `sudo apt upgrade --assume-yes` doesn't work, we get:
```plain
The following packages have been kept back:
  libmiral-dev libmiral3 libmirclient9 libmircommon-dev libmircommon7 libmircookie-dev libmircookie2 libmircore-dev
  libmirplatform18 libmirserver53 libmirwayland-bin libmirwayland-dev libmirwayland0 mir-client-platform-mesa5 mir-demos
  mir-platform-graphics-mesa-kms16 mir-platform-graphics-mesa-x16 mir-platform-graphics-wayland16
  mir-platform-input-evdev7 mir-test-tools
0 to upgrade, 0 to newly install, 0 to remove and 20 not to upgrade.
```
The simple workaround is `sudo apt dist-upgrade`, but that shouldn't be necessary.

@wmww, I think this is also why your graphics platform failed to update yesterday when you added the dev ppa.

I *suspect* this is because libmirclient-dev ceases to exist and will be uninstalled. @RAOF could you apply your debian foo?

PR #1540, opened 2020-06-19
Ensure that libmirclient-dev gets removed

AlanGriffiths:
Ensure that libmirclient-dev gets removed. (Fixes: #1518)


AlanGriffiths:
While it would be better to remove libmirclient9 entirely and avoid installation conflicts that way, there are too many dependencies to untangle. (I spent a few hours trying that before giving up.)

The alternative would be an artificial bump to libmirclient10. Opinions will differ.

PR #1542, opened 2020-06-22
Bump the graphics ABI

AlanGriffiths:
Bump the graphics ABI.

We have a new ABI, so bump the so name, package, & stanza version.


wmww:
bors delegate+

AlanGriffiths:
> bors delegate+

What are you trying to achieve? You've not raised any non-blocking discussion point to be considered before merging.

wmww:
It looks like a merge conflict needs to be fixed

Issue #1530, opened 2020-06-18
[MirAL] No way for WM strategy to query zones


AlanGriffiths:
In egmde, implementing "dock window to the left" there's the snippet:
```c++
                modifications.top_left().value().x = active_output.top_left.x;
```
This is technically wrong, as the active zone may have a margin to the left compared to `active_output`. But there's no simple, obvious way to access the zone to which the window should be docked.

Querying the active zone, the one corresponding to a window, or (possibly) to an output should be supported.

(IIRC there's similar code in miral-shell)

PR #1547, opened 2020-06-22
Add WindowManagerTools::active_application_zone()

AlanGriffiths:
Add WindowManagerTools::active_application_zone(). (Fixes: #1530)

Also uses it in the miral-shell "floating" window manager.

### DIFF
--- a/include/miral/miral/window_manager_tools.h
+++ b/include/miral/miral/window_manager_tools.h
@@ -38,6 +38,7 @@ class Window;
 struct WindowInfo;
 struct ApplicationInfo;
 class WindowSpecification;
+class Zone;
 
 /**
  * Workspace is intentionally opaque in the miral API. Its only purpose is to
@@ -165,7 +166,7 @@ public:
 
     /// Find the active zone area
     /// \remark Since MirAL 3.0
-    auto active_zone() -> mir::geometry::Rectangle const;
+    auto active_application_zone() const -> Zone;
 
     /// Raise window and all its children
     void raise_tree(Window const& root);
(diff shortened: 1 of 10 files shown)


include/miral/miral/window_manager_tools.h
```diff
@@ -163,6 +163,10 @@ class WindowManagerTools
     /// Find the active output area
     auto active_output() -> mir::geometry::Rectangle const;
 
+    /// Find the active zone area
+    /// \remark Since MirAL 3.0
+    auto active_zone() -> mir::geometry::Rectangle const;
```
wmww, line 168: Wouldn't it make more sense to return a `miral::Zone`? The policy may want to match the returned zone to an existing application zone it knows about.

Also, can we change the name to `active_application_zone()`? That would clearly indicate that the returned value will be one of the application zones the policy knows about.

AlanGriffiths:
On Mon, 22 Jun 2020 13:44:34 -0700, William Wold wrote:
> @wmww requested changes on this pull request.
>
> When we get to break the API I want to rethink how the 
> `miral::Zone` class works, but until then 

We are breaking ABI.

> we should use it for 
> what it was built for
>
>> @@ -163,6 +163,10 @@ class WindowManagerTools
>      /// Find the active output area
>      auto active_output() -> mir::geometry::Rectangle const;
>  
> +    /// Find the active zone area
> +    /// \remark Since MirAL 3.0
> +    auto active_zone() -> mir::geometry::Rectangle const;
>
> Wouldn't it make more sense to return a `miral::Zone`? The 
> policy may want to match the returned zone to an existing 
> application zone it knows about.
>
> Also, can we change the name to `active_application_zone()`? 
> That would clearly indicate that the returned value will be one 
> of the application zones the policy knows about.
>

-- 

wmww:
In that case I would like to talk about the `Zone` class, as I think it has some issues

AlanGriffiths:
> In that case I would like to talk about the `Zone` class, as I think it has some issues

Here? On the forum? ...?

PR #1549, opened 2020-06-23
XWayland connector own spawner

wmww:
This moves ownership of the `XWaylandSpawner` into the the connector.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_connector.cpp
+++ b/src/server/frontend_xwayland/xwayland_connector.cpp
@@ -22,6 +22,7 @@
 
 #include "wayland_connector.h"
 #include "xwayland_server.h"
+#include "xwayland_spawner.h"
 
 namespace mf = mir::frontend;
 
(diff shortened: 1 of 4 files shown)


src/server/frontend_xwayland/xwayland_connector.cpp
```diff
@@ -37,6 +38,14 @@ void mf::XWaylandConnector::start()
 {
     if (wayland_connector->get_extension("x11-support"))
     {
+        /// TODO: make threadsafe
+        xwayland_spawner = std::make_unique<XWaylandSpawner>([this]()
+            {
+                if (xwayland_server && xwayland_spawner)
```
AlanGriffiths, line 44: 1. Why add a TODO like this? Just do it.
2. Why are we setting `xwayland_spawner` and then testing it?

wmww:
Oops, this PR should not have been left open. Replaced by #1550

Issue #1552, opened 2020-06-23
[X11] We incorrectly query WM_NAME using string_from()


AlanGriffiths:
I got intrigued by the error messages, and added some debug code...
```plain
[2020-06-23 11:46:45.750617] < -warning- > xwayland: Failed to process property reply: WM_NAME: /home/alan/CLionProjects/mir/src/server/frontend_xwayland/xcb_connection.cpp(236): Throw in function std::string mir::frontend::XCBConnection::string_from(const xcb_get_property_reply_t*) const
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: Supplied reply is of type COMPOUND_TEXT and does not hold string data
```

AlanGriffiths:
A naive check (commenting out `reply_contains_text_data()`) seems to suggest the main problem is that we don't recognise COMPOUND_TEXT. Not sure what we should add to that test though.

PR #1553, opened 2020-06-23
[X11] Treat COMPOUND_TEXT as a string type

AlanGriffiths:
[X11] Treat COMPOUND_TEXT as a string type. (Fixes: #1552)

### DIFF
--- a/src/server/frontend_xwayland/xcb_connection.h
+++ b/src/server/frontend_xwayland/xcb_connection.h
@@ -244,7 +244,7 @@ public:
     Atom const clipboard;
     Atom const clipboard_manager;
     Atom const utf8_string;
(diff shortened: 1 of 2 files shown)


src/server/frontend_xwayland/xcb_connection.h
```diff
@@ -244,6 +244,7 @@ class XCBConnection
     Atom const clipboard;
     Atom const clipboard_manager;
     Atom const utf8_string;
+    Atom const compound_text{"COMPOUND_TEXT", this};
```
wmww, line 247: Can you move the initialization to the `XCBConnection` constructor for consistency with the rest of the atoms? I thought there was a reason they had to be constructed there. If not, we might want to move all the constructors to the header for easier maintenance, but we should be consistent about it.

PR #1560, opened 2020-06-25
Rename XCBConnection atoms to match X11 atom names

wmww:
Open for discussion.

Pros:
- No possibility of accidentally initializing an atom with the wrong name
- IMO all caps distinguish atoms in the code better and match the atom constants provided by XCB
- Obvious exactly which atoms are being used/you can copy paste from the code into google to look up an atom

Cons:
- Technically a style guide violation
- Not the most important thing for me to be doing right now (but it's already done)

(NOTE: I'm not currently working and won't respond to reviews until Friday, Posting now because I just noticed the per-rec PR merged)


wmww:
I prefer this way because it looks clean and doesn't allow for as many things to go wrong at runtime.

wmww:
Fixed the conflicts. If there's no objection, can we please merge this? It makes ctrl+Fing atoms a lot easier when you don't have to guess what they're called in the code.

RAOF (approved):
I am happy with this.

bors r+

Issue #1543, opened 2020-06-22
[Mir 2.0] Mir-on-X not selected on Nvidia graphics


AlanGriffiths:
We've just landed support for running Mir-on-X on Nvidia, but the new "x11" platform is not selected automatically on my kit. Here's a log:
```plain
$ cmake-build-debug/bin/miral-app
MIR_CLIENT_PLATFORM_PATH=cmake-build-debug/bin/../lib/client-modules/
MIR_SERVER_PLATFORM_PATH=cmake-build-debug/bin/../lib/server-modules/
LD_LIBRARY_PATH=cmake-build-debug/bin/../lib
exec=cmake-build-debug/bin/miral-shell.bin
[2020-06-22 11:19:22.244236] <information> mirserver: Starting
[2020-06-22 11:19:22.244440] < - debug - > mirserver: Not trying logind: "DISPLAY" is set and X need not have claimed the VT
[2020-06-22 11:19:22.244603] < - debug - > mirserver: Not using Linux VT subsystem for session management: Failed to find the current VT
[2020-06-22 11:19:22.244642] < - debug - > mirserver: No session management supported
[2020-06-22 11:19:22.244669] <information> VT switch key handler: No VT switching support available: MinimalConsoleServices does not support VT switching
[2020-06-22 11:19:22.244939] <information> mircommon: Loading modules from: cmake-build-debug/bin/../lib/server-modules/
[2020-06-22 11:19:22.245045] <information> mircommon: Loading module: cmake-build-debug/bin/../lib/server-modules/server-mesa-x11.so.17
[2020-06-22 11:19:22.245444] <information> mircommon: Loading module: cmake-build-debug/bin/../lib/server-modules/graphics-wayland.so.16
[2020-06-22 11:19:22.245482] <information> mircommon: Loading module: cmake-build-debug/bin/../lib/server-modules/graphics-eglstream-kms.so.16
[2020-06-22 11:19:22.245516] <information> mircommon: Loading module: cmake-build-debug/bin/../lib/server-modules/graphics-mesa-kms.so.16
[2020-06-22 11:19:22.245529] <information> mircommon: Loading module: cmake-build-debug/bin/../lib/server-modules/server-x11.so.16
[2020-06-22 11:19:22.245558] <information> mircommon: Loading module: cmake-build-debug/bin/../lib/server-modules/input-evdev.so.7
[...]

PR #1561, opened 2020-06-26
Minor platform probing improvements

RAOF:
Add a `PlatformPriority::hosted` probe value, so we can select a hardware-accelerated X11 platform over a software-rasterised `mesa-kms` platform, and check whether or not there is any output hardware before calling the `eglstream-kms` platform appropriate.

Partially fixes: #1543. The only thing now left is that it can select `eglstream-kms`, which will work until trying to load the *input* platform, which will fail. Fixing that requires a more significant rework of platform probing, tying input & graphics platforms together.

### DIFF
--- a/include/platform/mir/graphics/platform.h
+++ b/include/platform/mir/graphics/platform.h
@@ -164,6 +164,10 @@ enum PlatformPriority : uint32_t
     supported = 128,    /**< Capable of providing a functioning Platform on this device,
                          *    possibly with degraded performance or features.
                          */
+    hosted = 192,       /**< Capable of providing a fully-featured Platform on this device,
+                         *   running nested under some other display server rather than with
+                         *   exclusive hardware access.
+                         */
     best = 256          /**< Capable of providing a Platform with the best features and
                          *   performance this device is capable of.
                          */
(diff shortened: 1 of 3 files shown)


include/platform/mir/graphics/platform.h
```diff
@@ -164,6 +164,10 @@ enum PlatformPriority : uint32_t
     supported = 128,    /**< Capable of providing a functioning Platform on this device,
                          *    possibly with degraded performance or features.
                          */
+    hosted = 192,       /**< Capable of providing a fully-featured Platform on this device,
+                         *   running nested under some other display server rather than with
+                         *   exclusive hardware access.
+                         */
```
AlanGriffiths, line 170: Should the Wayland platform also return `hosted`? Currently we have:
```c++
    return mpw::connection_options_supplied(options) ? mg::PlatformPriority::best :
           mg::PlatformPriority::unsupported;
```

AlanGriffiths:
bors delegate=RAOF

include/platform/mir/graphics/platform.h
RAOF: I think in this case Wayland should return `best` - this isn't an auto-probe, this is “the user has explicitly set the command-line option to use the Wayland platform”.

Maybe what should happen here is the Wayland platform should return “the user has set a Wayland platform option, so Wayland is the only sensible option” value? Or do we want to be able to handle “*if* the Wayland platform is autoselected, have it use this option” cases?

Maybe something for a followup discussion?

RAOF:
bors r=AlanGriffiths

PR #1564, opened 2020-06-26
[MirAL] Add ExternalClientLauncher::snapcraft_launch()

AlanGriffiths:
[MirAL] Add ExternalClientLauncher::snapcraft_launch().

This introduces support for launching other snaps via the proposed desktop-launch interface. While that interface has yet to land in snapd the client-side code needed is unlikely to change.

### DIFF
--- a/src/miral/external_client.cpp
+++ b/src/miral/external_client.cpp
@@ -101,7 +101,9 @@ void miral::ExternalClientLauncher::operator()(mir::Server& server)
 auto miral::ExternalClientLauncher::launch(std::vector<std::string> const& command_line) const -> pid_t
 {
     if (!self->server)
-        throw std::logic_error("Cannot launch apps when server has not started");
+    {
+        BOOST_THROW_EXCEPTION(std::logic_error("Cannot launch apps when server has not started"));
+    }
 
     auto const wayland_display = self->server->wayland_display();
     auto const x11_display = self->server->x11_display();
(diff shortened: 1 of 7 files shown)


src/miral/external_client.cpp
```diff
@@ -125,4 +126,26 @@ auto  miral::ExternalClientLauncher::launch_using_x11(std::vector<std::string> c
     return -1;
 }
 
+void miral::ExternalClientLauncher::snapcraft_launch(const std::string& desktop_file) const
+{
+    if (!self->server)
+        throw std::logic_error("Cannot launch apps when server has not started");
```
wmww, line 132: Any reason why this doesn't use `BOOST_THROW_EXCEPTION()` like we do most other places? Also, I believe we're leaning toward braces around single line ifs (as you have just below)

src/miral/open_desktop_entry.cpp
```diff
@@ -0,0 +1,116 @@
...
+                                                        dest,
+                                                        object_path,
+                                                        interface_name,
+                                                        method_name,
+                                                        g_variant_new("(sas)", id.c_str(), builder),
+                                                        nullptr,
+                                                        G_DBUS_CALL_FLAGS_NONE,
+                                                        G_MAXINT,
+                                                        nullptr,
+                                                        &error))
```
wmww, line 103: Excessive indentation

src/miral/open_desktop_entry.cpp
```diff
@@ -0,0 +1,116 @@
...
+    {
+        g_variant_unref(result);
+    }
+
+    if (error)
+    {
+        puts(error->message);
+        printf("dest=%s, object_path=%s, interface_name=%s, method_name=%s, id=%s\n", dest, object_path, interface_name, method_name, id.c_str());
```
wmww, line 111: Shouldn't this be some sort of Mir log call instead of `printf()`?

AlanGriffiths:
> Do we want to merge this when it's ready, or wait for the needed changes to land in snapd?

We can land it. The snapd changes may take some time, but the client side seems settled.

AlanGriffiths:
Github is flaky today, I've pushed changes, they're on the branch, but they're not seen by the PR.

PR #1566, opened 2020-06-26
XWaylandWM: close all clients when shutting down

wmww:
This allows the X server to exit gracefully (no more "did not close, killing it" messages)


wmww:
> But note this doesn't prevent Xwayland being killed. E.g. After start miral-app, start xterm, close miral-shell; we now kill xterm, but Xwayland still gets killed for not exiting.

Hmm, perhaps something else in my branch is also required. Regardless, doesn't hurt

PR #1567, opened 2020-06-29
eglstream-kms: Actually clean up context in probe

RAOF:
We need to capture the `EGLContext` by reference if we want it to be valid when we come to tear it down!


RAOF:
Should have been part of the probe improvements, but got left behind. Whoops!

AlanGriffiths:
(Github is flaky - doesn't let me approve today.)

PR #1568, opened 2020-06-29
More cleanup of XWaylandServer

wmww:
Previously we made `XWaylandServer` fork on construction and removed the window manager logic from it. This cleans up the mess that was left over. It transitions from methods that set mutable class members to standalone initialization functions that are used to initialize const members. If the diff is confusing going through commit by commit might help.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -101,14 +101,10 @@ void exec_xwayland(mf::XWaylandSpawner const& spawner, std::string const& xwayla
 
     execvp(xwayland_path.c_str(), const_cast<char* const*>(args.data()));
 }
-}
 
-mf::XWaylandServer::XWaylandServer(
-    std::shared_ptr<WaylandConnector> const& wayland_connector,
-    XWaylandSpawner const& spawner,
-    std::string const& xwayland_path)
-    : wayland_connector{wayland_connector},
-      xwayland_path{xwayland_path}
+auto fork_xwayland_process(
+    mf::XWaylandSpawner const& spawner,
+    std::string const& xwayland_path) -> mf::XWaylandServer::XWaylandProcess
 {
     enum { MIR, XWAYLAND, SIZE };
     int wayland_fds[SIZE], x11_fds[SIZE];
@@ -128,7 +124,7 @@ mf::XWaylandServer::XWaylandServer(
     }
 
     mir::log_info("Starting XWayland");
-    xwayland_pid = fork();
+    pid_t const xwayland_pid = fork();
 
     switch (xwayland_pid)
     {
@@ -138,42 +134,23 @@ mf::XWaylandServer::XWaylandServer(
         close(x11_fds[MIR]);
         close(x11_fds[XWAYLAND]);
         BOOST_THROW_EXCEPTION(std::runtime_error("Failed to fork XWayland process"));
-        return; // Keep compiler happy (doesn't reach here)
 
     case 0:
         close(wayland_fds[MIR]);
         close(x11_fds[MIR]);
         exec_xwayland(spawner, xwayland_path, wayland_fds[XWAYLAND], x11_fds[XWAYLAND]);
-        return; // Keep compiler happy (doesn't reach here)
+        BOOST_THROW_EXCEPTION(std::logic_error("Should be unreachable"));
 
     default:
         close(wayland_fds[XWAYLAND]);
         close(x11_fds[XWAYLAND]);
-        x11_fd = Fd{x11_fds[MIR]};
-        wayland_fd = Fd{wayland_fds[MIR]};
-        connect_wm_to_xwayland();
-        break;
+        return mf::XWaylandServer::XWaylandProcess{
+            xwayland_pid,
+            mir::Fd{x11_fds[MIR]},
+            mir::Fd{wayland_fds[MIR]}};
     }
 }
 
-mf::XWaylandServer::~XWaylandServer()
-{
-    mir::log_info("Deiniting xwayland server");
-
-    // Terminate any running xservers
-    if (kill(xwayland_pid, SIGTERM) == 0)
-    {
-        std::this_thread::sleep_for(100ms);// After 100ms...
-        if (kill(xwayland_pid, 0) == 0)    // ...if Xwayland is still running...
-        {
-            mir::log_info("Xwayland didn't close, killing it");
-            kill(xwayland_pid, SIGKILL);     // ...then kill it!
-        }
-    }
-}
-
-namespace
-{
 bool spin_wait_for(sig_atomic_t& xserver_ready)
 {
     auto const end = std::chrono::steady_clock::now() + 5s;
@@ -185,9 +162,10 @@ bool spin_wait_for(sig_atomic_t& xserver_ready)
 
     return xserver_ready;
 }
-}
 
-void mf::XWaylandServer::connect_wm_to_xwayland()
+auto connect_xwayland_wl_client(
+    std::shared_ptr<mf::WaylandConnector> const& wayland_connector,
+    mir::Fd const& wayland_fd) -> wl_client*
 {
     // We need to set up the signal handling before connecting wl_client_server_fd
     static sig_atomic_t xserver_ready{ false };
@@ -204,23 +182,25 @@ void mf::XWaylandServer::connect_wm_to_xwayland()
     action.sa_flags = 0;
     sigaction(SIGUSR1, &action, &old_action);
 
+    wl_client* client{nullptr};
+
     {
         std::mutex client_mutex;
         std::condition_variable client_ready;
 
         wayland_connector->run_on_wayland_display(
-            [this, &client_mutex, &client_ready](wl_display* display)
+            [&](wl_display* display)
             {
                 std::lock_guard<std::mutex> lock{client_mutex};
-                wayland_client = wl_client_create(display, wayland_fd);
+                client = wl_client_create(display, wayland_fd);
                 client_ready.notify_all();
             });
 
         std::unique_lock<std::mutex> lock{client_mutex};
-        if (!client_ready.wait_for(lock, 10s, [&]{ return wayland_client; }))
+        if (!client_ready.wait_for(lock, 10s, [&]{ return client; }))
         {
             // "Shouldn't happen" but this is better than hanging.
-            mir::fatal_error("Failed to create wl_client for Xwayland");
+            BOOST_THROW_EXCEPTION(std::runtime_error("Failed to create wl_client for XWayland"));
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -139,25 +77,72 @@ void mf::XWaylandServer::execl_xwayland(XWaylandSpawner const& spawner, int wl_c
...
     execvp(xwayland_path.c_str(), const_cast<char* const*>(args.data()));
 }
 
-namespace
+auto fork_xwayland_process(
+    mf::XWaylandSpawner const& spawner,
+    std::string const& xwayland_path) -> mf::XWaylandServer::XWaylandProcess
 {
+    enum { MIR, XWAYLAND, SIZE };
+    int wayland_fds[SIZE], x11_fds[SIZE];
```
AlanGriffiths, line 102: 1. Let's leave uppercase for preprocessor MACROS.
2. These names are confusing:
2.1. `wayland_fds[MIR]` and `wayland_fds[XWAYLAND]` make some sense once you know how they are used as the former is passed to the Mir Wayland server, while the latter to the Xwayland client, but I don't find the names tell me what they are for.
2.2. Maybe, `wayland_pipe[server]`/`wayland_pipe[client]' would be clearer? And maybe `xcb_pipe[server]`/`xcb_pipe[client]'?

(The names in the original code not great either.)

wmww: I think `server` and `client` are entirely unclear. If you think about it real hard, Mir is a Wayland server/X11 client and XWayland is a X11 server/Wayland client. Trying to use that relationship for naming only complicates matters. Better to name them based on the process that will end up using them.

`*_pipe` naming is good, but I'm not going to change `x11` to `xcb`, since XCB is only used on our side and isn't relevant here.

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -139,25 +77,72 @@ void mf::XWaylandServer::execl_xwayland(XWaylandSpawner const& spawner, int wl_c
...
+    pid_t const xwayland_pid = fork();
+
+    switch (xwayland_pid)
+    {
+    case -1:
+        close(wayland_pipe[mir_side]);
+        close(wayland_pipe[xwayland_side]);
+        close(x11_pipe[mir_side]);
+        close(x11_pipe[xwayland_side]);
+        BOOST_THROW_EXCEPTION(std::runtime_error("Failed to fork XWayland process"));
```
RAOF, line 128: `std::system_error`, and pass in `errno` so we can have some idea why it failed?

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -139,25 +77,72 @@ void mf::XWaylandServer::execl_xwayland(XWaylandSpawner const& spawner, int wl_c
...
+
+    case 0:
+        close(wayland_pipe[mir_side]);
+        close(x11_pipe[mir_side]);
+        exec_xwayland(spawner, xwayland_path, wayland_pipe[xwayland_side], x11_pipe[xwayland_side]);
+        BOOST_THROW_EXCEPTION(std::logic_error("Should be unreachable"));
```
RAOF, line 134: This is not actually going to be helpful, is it? This should probably be an `abort()` to tear everything down faster, because we're in a forked process?

Does the main process have a way of noticing this?

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -139,25 +77,72 @@ void mf::XWaylandServer::execl_xwayland(XWaylandSpawner const& spawner, int wl_c
...
+    case -1:
+        close(wayland_pipe[mir_side]);
+        close(wayland_pipe[xwayland_side]);
+        close(x11_pipe[mir_side]);
+        close(x11_pipe[xwayland_side]);
+        BOOST_THROW_EXCEPTION(std::runtime_error("Failed to fork XWayland process"));
+
+    case 0:
+        close(wayland_pipe[mir_side]);
+        close(x11_pipe[mir_side]);
```
RAOF, line 132: We don't technically need to do this, do we, as these are `CLOEXEC`? It's probably nice to do this explicitly, though.

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -188,23 +174,25 @@ void mf::XWaylandServer::connect_wm_to_xwayland()
...
 
+    wl_client* client{nullptr};
+
     {
         std::mutex client_mutex;
         std::condition_variable client_ready;
 
         wayland_connector->run_on_wayland_display(
-            [this, &client_mutex, &client_ready](wl_display* display)
+            [&](wl_display* display)
```
RAOF, line 184: It's pretty unlikely, but this is a potential use-after-free, isn't it? In the unlikely event that `run_on_wayland_display` takes more than the 10s to execute the functor this will try to lock a destroyed mutex and then write to a stale stack frame.

I'm not sure if this needs fixing or not.

Let's say yes, it should. Potential use-after-frees are potential security bugs.

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -214,6 +202,34 @@ void mf::XWaylandServer::connect_wm_to_xwayland()
...
+}
+
+mf::XWaylandServer::~XWaylandServer()
+{
+    mir::log_info("Deiniting xwayland server");
+
+    // Terminate any running xservers
+    if (kill(xwayland_process.pid, SIGTERM) == 0)
+    {
+        std::this_thread::sleep_for(100ms);// After 100ms...
```
RAOF, line 228: We could do better than sleeping by waiting for the child to exit, but maybe that can wait for a `Process` abstraction…

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -38,92 +32,36 @@
...
-    std::string const& xwayland_path)
-    : wayland_connector{wayland_connector},
-      xwayland_path{xwayland_path}
+namespace
+{
+void exec_xwayland(mf::XWaylandSpawner const& spawner, std::string const& xwayland_path, int wayland_fd, int x11_fd)
 {
-    enum { server, client, size };
-    int wl_client_fd[size], wm_fd[size];
+    setenv("EGL_PLATFORM", "DRM", 1);
```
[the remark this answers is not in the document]
wmww: Absolutely no idea

src/server/frontend_xwayland/xwayland_server.cpp
AlanGriffiths: Well, each pipe has a client and a server end, that's no more unclear than having a Mir end and an Xwayland end. Actually, I think it more clear: for example, the server end of the Wayland pipe being connected to Mir isn't intrinsic to the pipe, just how it happens to be used. (I admit the socket pairs are not intrinsically directional, but the pipes can be treated that way.)

wmww: Changed to client/server naming hopefully in a way a little clearer to both of us

src/server/frontend_xwayland/xwayland_server.cpp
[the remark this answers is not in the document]
wmww: It's not this PR's place to change this, but I can drop it in a future PR if you want

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -139,25 +77,72 @@ void mf::XWaylandServer::execl_xwayland(XWaylandSpawner const& spawner, int wl_c
...
             "-rootless",
-            "-wm", wm_fd_str.c_str(),
+            "-wm", x11_fd_str.c_str(),
             "-terminate",
         };
 
     for (auto const& fd : spawner.socket_fds())
     {
-        XWaylandSpawner::set_cloexec(fd, false);
+        mf::XWaylandSpawner::set_cloexec(fd, false);
```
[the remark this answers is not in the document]
wmww: Removed the `dup()`s.

src/server/frontend_xwayland/xwayland_server.cpp
wmww: Done

src/server/frontend_xwayland/xwayland_server.cpp
wmww: Done

It seems to just lock up for a while as it waits for xwayland to start. Might look at improving the error path.

src/server/frontend_xwayland/xwayland_server.cpp
wmww: Switched to relying on `mir::Fd`'s destructor and `CLOEXEC` as needed

src/server/frontend_xwayland/xwayland_server.cpp
wmww: Hopefully fixed

src/server/frontend_xwayland/xwayland_server.cpp
wmww: I don't know that we need a new abstraction, but that's at least a task for another PR

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -38,93 +32,44 @@
...
+    int pipe[2];
+    if (socketpair(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0, pipe) < 0)
     {
-        std::this_thread::sleep_for(100ms);// After 100ms...
-        if (kill(xwayland_pid, 0) == 0)    // ...if Xwayland is still running...
-        {
-            mir::log_info("Xwayland didn't close, killing it");
-            kill(xwayland_pid, SIGKILL);     // ...then kill it!
-        }
+        BOOST_THROW_EXCEPTION(std::system_error(errno, std::generic_category(), "Creating socket pair failed"));
```
RAOF, line 55: Extremely niche nit: this should probably be `std::system_category` (as it's a OS-generated error). It's functionally the same, though, as we don't need it's “map code to POSIX errno” functionality, given we *are* throwing in the POSIX errno :)

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -139,25 +84,48 @@ void mf::XWaylandServer::execl_xwayland(XWaylandSpawner const& spawner, int wl_c
...
+    pid_t const xwayland_pid = fork();
+
+    switch (xwayland_pid)
+    {
+    case -1:
+        BOOST_THROW_EXCEPTION(std::system_error(errno, std::generic_category(), "Failed to fork XWayland process"));
+
+    case 0:
+        exec_xwayland(spawner, xwayland_path, wayland_pipe.client, x11_wm_pipe.server);
+        fprintf(stderr, "Failed to start XWayland, should be unreachable");
```
RAOF, line 121: If you used `perror("Failed to exec XWayland")` here we'd get the error message from `execvpe` thrown in for free.

RAOF: (It might be better to make that call in `exec_xwayland` immediately after the exec call, though)

PR #1570, opened 2020-06-29
When the focus moves to a new client ensure the keyboard state is (re)synchronized

### LINKED ISSUE #1515: keyboard state of modifier keys gets out of sync
I've been using egmde for development today. Every hour or so, with no pattern I've yet spotted the keyboard state gets out of sync. For example, typing will be in capitals (and the shift keys get inverted), or it will be as though the Alt key is held.

It is possible to workaround this by VT switching away and back (as that resyncs the keyboard courtesy of merge #1035).

I've occasionally seen the problem using miral-shell on Mir-on-X as well as the Mir-on-logind case described above. In that case, press and release of the corresponding key resyncs the state.

AlanGriffiths:
When the focus moves to a new client ensure the keyboard state is (re)synchronized.

Improves, maybe fixes #1515

### DIFF
--- a/src/server/frontend_wayland/wl_seat.cpp
+++ b/src/server/frontend_wayland/wl_seat.cpp
@@ -211,6 +211,8 @@ void mf::WlSeat::notify_focus(wl_client *focus)
         focused_client = focus;
         for (auto const listener : focus_listeners)
             listener->focus_on(focus);
+
+        executor->spawn([this] { resync_keyboard_state(); });
     }
 }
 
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
Hah! So the way to make it go wrong is to submit a PR.

src/server/frontend_wayland/wl_seat.cpp
```diff
@@ -211,6 +211,8 @@ void mf::WlSeat::notify_focus(wl_client *focus)
         focused_client = focus;
         for (auto const listener : focus_listeners)
             listener->focus_on(focus);
+
+        executor->spawn([this] { resync_keyboard_state(); });
```
wmww, line 215: It looks to me like this function is already on the Wayland thread. Why is the executor used? Is there a better way?

PR #1571, opened 2020-06-29
XWaylandConnector: error if destroyed without being stopped

wmww:
There's thread safety issues if this ever happens. Lets make sure it doesn't.


wmww:
Oops, this won't compile until #1569 lands

PR #1574, opened 2020-06-29
wayland::Weak: add .nullable_ptr() method

wmww:
This is useful to easily compare the contained resource with another, or in if assignments:
```c++
if (auto const surface = wl_surface.or_null())
{
    use(surface);
}
```
EDIT: changed example slightly to reflect how things are actually commonly named in the code

### DIFF
--- a/include/wayland/mir/wayland/wayland_base.h
+++ b/include/wayland/mir/wayland/wayland_base.h
@@ -93,6 +93,18 @@ public:
         return *resource;
     }
 
+    auto as_nullable_ptr() const -> T*
+    {
+        if (*this)
+        {
+            return resource;
+        }
+        else
+        {
+            return nullptr;
+        }
+    }
+
 private:
     T* resource;
     /// Is null if and only if resource is null


wmww:
In actual usage it looks like `surface.or_null()`, which reads better than `surface.value_or_null()`. The alternative to null is obvious. `.value()` is always basically just noise in the code telling you to pay attention to nullability. `.or_null()` does a fine job of that on it's own.

wmww:
Alan likes `.value_or_null()` because it's more consistent with optional, so changed

wmww:
Actually that doesn't quite work either. `.value()` returns a reference but `.value_or_null()` returns a pointer. Perhaps `.as_ptr()`? `.nullable_ptr()`? I'm changing the other PRs I have planned to not be blocked by this.

RAOF:
Hm. That actually raises the question: should `value()` return a reference, or a pointer?

This is most frequently going to be used with opaque types we can *only* access through pointers, right? `wl_foo*`-types? Do we *want* to be returning a `wl_foo&`?

We also don't want calling code to swap `T` out from under the `Weak<T>`, which returning a mutable reference to `*resource` allows.

AlanGriffiths:
> Hm. That actually raises the question: should `value()` return a reference, or a pointer?
> 
> This is most frequently going to be used with opaque types we can _only_ access through pointers, right? `wl_foo*`-types? Do we _want_ to be returning a `wl_foo&`?

I concur, it is odd to return a reference: A pointer would make more sense.

wmww:
I don't have any real objection to returning a pointer, but I don't understand your arguments for it.

> We also don't want calling code to swap T out from under the Weak<T>, which returning a mutable reference to *resource allows.

How is this different from a pointer? Can't you get a reference simply by dereferencing a pointer?

wmww:
I don't have any objection to returning a pointer, but I don't understand your arguments for it.

> This is most frequently going to be used with opaque types we can only access through pointers, right?

Um.. no? Or maybe I misunderstand? It's mostly used for things like `WlSurface`, which is not opaque or even currently an interface.

> We also don't want calling code to swap T out from under the Weak, which returning a mutable reference to *resource allows.

How is this different from a pointer? Can't you get a reference simply by dereferencing a pointer?

AlanGriffiths:
The technical difference between returning a raw pointer and a reference is mostly syntactic: "." or "->". (A pointer can be reseated.)

But if `.._or_null()` makes sense, do we even need to make the distinction?

We can always write:
```c++
if (weak.value_or_null())
{
   weak.value_or_null()->foo();
}
```
Where we have written:
```c++
if (weak)
{
   weak.value().foo();
}
```
So why support both usages? We could just make `value()` have the same semantics as proposed for `value_or_null()`.

I think we ought to be clear about what the expected usage of `Weak` is before trying to make it many things.

AlanGriffiths:
I'm perfectly happy with, for example, `value()` returning a (not-null) pointer and `value_or_null()` returning a (possibly null) pointer. But as you say, with the existing interface you can also write:
```c++
template<class Type> auto value_or_null(Weak const& weak) -> Type* {...}
```

wmww:
> I'm perfectly happy with, for example, `value()` returning a (not-null) pointer

I see no reason to change the interface to make the not-nullness of `value()` explained by the documentation when we currently have it checked by the compiler.

AlanGriffiths:
As I've said above, we don't need a member function for this capability. With my "do as the standard does" hat on, I suggest we emulate `std::to_string()` rather than `std::string::c_str()`.

``C++
template<class Type> auto as_nullable_ptr(Weak<Type> const& weak) -> Type* {...}
```

wmww:
I don't see any advantage of a standalone function, nor do I see it particularly more std-compatible. As you point out, while std uses standalones to convert between unrelated types, the more similar operation of returning a pointer to internal data is done with a method. The other vaguely similar operation of locking a weak pointer is also a method.

AlanGriffiths:
Having some members return `Foo&` and some return `Foo*` makes for a jarring consumer experience. (But I think any downside is outweighed by the time this conversation is taking.)

PR #1575, opened 2020-06-29
Allow wayland::Weak to be created null

wmww:
Allow a `wayland::Weak` to be explicitly created null, and handle the case where it is created with a null pointer to a resource


wmww:
bors r=AlanGriffiths

RAOF:
Wow. Google cloud broken?

bors r=AlanGriffiths

PR #1576, opened 2020-06-30
Small performance test improvements

RAOF:


AlanGriffiths:
> From an abundance of caution I am going to dig a bit before borsing.

OK, a (somewhat overdue) reboot seems to have fixed the segfault.

bors r+

AlanGriffiths:
Leaving it looping over the performance tests for 10min eventually got the same segfault. (I doubt this change is to blame.)

PR #1573, opened 2020-06-29
wayland::LifetimeTracker

wmww:
Break out management of the destoryed flag into a new `LifetimeTracker` class so non-resource objects can inherit from it as well. Future PR will make `WindowWlSurfaceRole` use it.

PR #1581, opened 2020-07-01
wayland::LifetimeTracker (as member)

wmww:
Break the `destroyed_flag()` functionality out of `wayland::Resource` so it can be used by other classes. Alternative to #1573 (this variant makes `LifetimeTracker` a member of the classes that use it, and that one uses inheritance). I prefer #1573 because it doesn't require changing everything that uses `->destroyed_flag()` and, in my opinion, inheritance makes it easier to understand what's going on. Regardless, this is perfectly acceptable as well.

### DIFF
--- a/include/wayland/mir/wayland/wayland_base.h
+++ b/include/wayland/mir/wayland/wayland_base.h
@@ -32,6 +32,22 @@ namespace mir
 {
 namespace wayland
 {
+/// The base class of any object that wants to provide a destroyed flag
+/// The destroyed flag is only created when needed and automatically set to true on destruction
+/// This pattern is only safe in a single-threaded context
+class LifetimeTracker
+{
+public:
+    LifetimeTracker() = default;
+    LifetimeTracker(LifetimeTracker const&) = delete;
+    LifetimeTracker& operator=(LifetimeTracker const&) = delete;
+
+    virtual ~LifetimeTracker();
+    auto destroyed_flag() const -> std::shared_ptr<bool>;
+
+private:
+    std::shared_ptr<bool> mutable destroyed{nullptr};
+};
 
 class Resource
 {
@@ -42,18 +58,11 @@ public:
     };
 
     Resource();
-    virtual ~Resource();
 
-    Resource(Resource const&) = delete;
-    Resource& operator=(Resource const&) = delete;
-
-    auto destroyed_flag() const -> std::shared_ptr<bool>;
-
-private:
-    std::shared_ptr<bool> mutable destroyed;
+    LifetimeTracker const lifetime_tracker;
 };
 
-/// A weak handle to a Wayland resource
+/// A weak handle to a Wayland resource (or any Destroyable)
 /// May only be safely used from the Wayland thread
 template<typename T>
 class Weak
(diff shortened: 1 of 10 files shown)


AlanGriffiths:
>  I prefer #1573 because it doesn't require changing everything that uses `->destroyed_flag()`

There's no reason to change everything that uses `->destroyed_flag()`. Why not use the approach I suggested?

include/wayland/mir/wayland/wayland_base.h
```diff
@@ -32,6 +32,22 @@ namespace mir
 {
 namespace wayland
 {
+/// The base class of any object that wants to provide a destroyed flag
```
AlanGriffiths, line 35: Not a base class now?

wmww:
> There's no reason to change everything that uses ->destroyed_flag(). Why not use the approach I suggested?

Oh, right. That does indeed prevent the code changes, but also seems rather ugly. `destroyed_flag` doesn't really describe it anymore and a call operator *only* makes any sort of sense for the purpose of maintaining compatibility. Better to just make the code changes.

src/wayland/symbols.map
```diff
@@ -1,4 +1,4 @@
-MIRWAYLAND_1.2 {
+MIRWAYLAND_1.3 {
```
AlanGriffiths, line 1: I guess we're breaking ABI? If so, we ought to bump the soname too.

The stanza name for Mir 2.0 should be MIRWAYLAND_2.0.

wmww:
> I guess my comment would is: if we're updating users of the code (and there don't appear to be very many), shouldn't we just update them all to use Weak<T>?

There are only a handful of places resource destroyed flags are accessed, but changing every *use* of those flags to `Weak<T>` is a significant logic change. I have that change sitting in another  branch, but it depends on #1574 so it has also been blocked for almost a week. Regardless, I think it makes sense to switch everything over to using `LifetimeTracker` (including classes like `WindowWlSurfaceRole`), *then* switch users over to using `Weak<T>`, *then* change `LifetimeTracker` once the only effected user is `Weak<T>`.

> And anything else which wants to be convertable to Weak<T> just needs to provide an is_live member.

Having a `std::shared_ptr<void> const is_live{nullptr, [](auto){}};` sitting around a bunch of classes makes the code less readable, and will only cause trouble once we inevitably decide there's a better way to keep track of lifetimes. Abstractions are good. Abstractions are free. `LifetimeTracker` would be a useful abstraction even if it was just a type alias of `std::shared_ptr<void>`.

AlanGriffiths:
> ```c++
> 
>     std::shared_ptr<void> const is_live{nullptr, [](auto){}};
> ```

That makes it far too easy to subvert.

```C++
auto const accidental_evil = foo->is_alive;
```

wmww:
That's true, but I don't want to get side-tracked on the best internal implementation of `LifetimeTracker`. For now, it uses `std::shared_ptr<bool>` because that's what's already there. In the future (when it's not blocking other stuff), we can switch that to whatever we want.

AlanGriffiths:
We landed the alternative

PR #1586, opened 2020-07-08
(Unfinished) Rename to gbm platform

RAOF:


AlanGriffiths:
Incorporated in #1588

PR #1593, opened 2020-07-10
Use wayland::Weak in XDG shell implementations

wmww:
Replace the complicated and error prone direct usage of destroyed flags with `wayland::Weak`

### DIFF
--- a/src/server/frontend_wayland/xdg_shell_stable.cpp
+++ b/src/server/frontend_wayland/xdg_shell_stable.cpp
@@ -58,7 +58,7 @@ public:
 
     void send_configure();
 
-    std::experimental::optional<WindowWlSurfaceRole*> const& window_role();
+    wayland::Weak<WindowWlSurfaceRole> const& window_role();
 
     using wayland::XdgSurface::client;
     using wayland::XdgSurface::resource;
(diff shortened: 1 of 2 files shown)


src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -58,16 +58,15 @@ class XdgSurfaceStable : wayland::XdgSurface
 
     void send_configure();
 
-    std::experimental::optional<WindowWlSurfaceRole*> const& window_role();
+    wayland::Weak<WindowWlSurfaceRole> const& window_role();
```
AlanGriffiths, line 61: The `wayland::` prefix looks redundant.

src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -258,24 +261,19 @@ void mf::XdgSurfaceStable::send_configure()
...
-
     return window_role_;
 }
 
 void mf::XdgSurfaceStable::set_window_role(WindowWlSurfaceRole* role)
 {
-    if (window_role())
+    if (window_role_)
+    {
         log_warning("XdgSurfaceStable::window_role set multiple times");
```
AlanGriffiths, line 273: Pre-existing, but we ought to provide some information about which client & window is misbehaving.

wmww: The correct behavior here is to send a Wayland error and kill the client, but knowing what error to send will take a closer look. I believe there may already be a WLCS test for this (at least there are tests for similar things). As you said, pre-existing. I'll look at that another day.

PR #1596, opened 2020-07-10
Use wayland::Weak in the input system

wmww:


src/server/frontend_wayland/wayland_input_dispatcher.cpp
```diff
@@ -58,26 +60,33 @@ void mf::WaylandInputDispatcher::set_keymap(mi::Keymap const& keymap)
...
         return;
+    }
 
     if (has_focus)
         seat->notify_focus(client);
 
     seat->for_each_listener(client, [&](WlKeyboard* keyboard)
         {
-            keyboard->focussed(wl_surface, has_focus);
+            if (wl_surface)
```
AlanGriffiths, line 73: Didn't we check `wl_surface` already? (line 63)

wmww:
> this usage reinforces my opinion that Weak::value() should return T*, not T&

What's the argument for this again? All I remember is that it would make `.value()` return the same type as `.as_nullable_ptr()`. IMO references are *slightly* better because they show in the type that the value is not null, and they're consistent with `std::optional::value`. I don't think the fact that some users have to manually take the address of the value is a sign it should be a pointer. In fact, I'd be more inclined to change these functions to take a reference.

AlanGriffiths:
> > this usage reinforces my opinion that Weak::value() should return T*, not T&
> 
> What's the argument for this again? All I remember is that it would make `.value()` return the same type as `.as_nullable_ptr()`. IMO references are _slightly_ better because they show in the type that the value is not null, and they're consistent with `std::optional::value`. I don't think the fact that some users have to manually take the address of the value is a sign it should be a pointer. In fact, I'd be more inclined to change these functions to take a reference.

I guess it all comes down to what can be encoded in a type.

Really, what we both want is for `value()` to be semantically equivalent to `not_null<T*>` (using the Guidelines Support Library [libmsgsl-dev]). The syntactic difference between `.` and `->` is what decides me, while the local understanding that `value()` refers to an object is what decides you.

If we had:
```c++
template<typename T>
class Weak {
public:
    auto value() const -> not_null<T*>;
    auto maybe_null() const -> T*;
};
```
We'd likely both be happy?

wmww:
I guess? But I still don't see why `value()` can't just return a reference like `optional.value()` does. Why are `->` semantics better/more appropriate than `.` ones?

Btw, I'm planning on removing `.as_nullable_ptr()` and adding some equality operators, as I realized that was the main reason I wanted it.

AlanGriffiths:
> Btw, I'm planning on removing `.as_nullable_ptr()` and adding some equality operators, as I realized that was the main good reason I wanted it.

Let's try to get Weak to its final form before 2.0.

wmww:
I removed the `wl_surface` checks that were scattered everywhere `wl_surface` was used. Now (as before) public methods check `wl_surface` and gracefully return if it's invalidated. Private methods `fatal_error()`. This is to make it __obviously__ safe to use unchecked later in the same function, and to catch issues issues that may come up after some future refactoring.

Issue #1598, opened 2020-07-13
[input] Key repeat generates spurious events


AlanGriffiths:
I instrumented the `clone_event` code in `KeyRepeatDispatcher` while tracking #1515.

Pressing and holding Ctrl+Alt generates a load of events with alternating meta key state. Because each meta key is being repeated separately.

Should we even be repeating meta keys?

PR #1599, opened 2020-07-13
[input] Don't auto repeat meta keys

AlanGriffiths:
[input] Don't auto repeat meta keys. (Fixes: #1598)


AlanGriffiths:
This might also help with #1515 (but I've not finished that investigation)

AlanGriffiths:
> This might also help with #1515 (but I've not finished that investigation)

No such luck. Just seen it again.

PR #1602, opened 2020-07-14
More key repeat tweaking

AlanGriffiths:
More key repeat tweaking.

Reduce the key repeat logic to the simplest thing: don't repeat meta keys, only repeat one key, and don't repeat after modifier state changes.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface_observer.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface_observer.cpp
@@ -56,14 +56,7 @@ void mf::XWaylandSurfaceObserver::attrib_changed(ms::Surface const*, MirWindowAt
     switch (attrib)
     {
     case mir_window_attrib_focus:
-    {
-        auto has_focus = static_cast<bool>(value);
-        wm_surface->scene_surface_focus_set(has_focus);
-        aquire_input_dispatcher(
-            [has_focus](auto input_dispatcher)
-            {
-                input_dispatcher->set_focus(has_focus);
-            });
+    {        focussed(value);
     }   break;
 
     case mir_window_attrib_state:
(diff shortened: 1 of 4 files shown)


src/server/frontend_xwayland/xwayland_surface_observer.cpp
```diff
@@ -113,6 +120,22 @@ void mf::XWaylandSurfaceObserver::input_consumed(ms::Surface const*, MirEvent co
...
+            if (mir_keyboard_event_action(mir_input_event_get_keyboard_event(input_event)) ==
+                mir_keyboard_action_repeat)
+                return;
+
+            if (!has_focus)
+            {
+                log_warning("**** Sending key to surface even though it was not explicitly given keyboard focus");
+            }
+            // If we're sending keys we implicitly have input focus
+            focussed(true);
```
wmww, line 136: I don't like all this. It duplicates logic in `wl_keyboard.cpp`, and we never should be hitting this at all.

src/server/frontend_xwayland/xwayland_surface_observer.h
```diff
@@ -85,6 +86,9 @@ class XWaylandSurfaceObserver
     XWaylandSurfaceObserverSurface* const wm_surface;
     std::shared_ptr<ThreadsafeInputDispatcher> const input_dispatcher;
 
+    std::atomic<bool> has_focus{false};
+    void focussed(bool has_focus);
```
wmww, line 90: `has_focus`, `focussed()` and `include <atomic>` should all be dropped

src/server/frontend_xwayland/xwayland_surface_observer.cpp
```diff
@@ -113,6 +113,12 @@ void mf::XWaylandSurfaceObserver::input_consumed(ms::Surface const*, MirEvent co
 {
     if (mir_event_get_type(event) == mir_event_type_input)
     {
+        // Ignore key repeat events
+        auto const input_event = mir_event_get_input_event(event);
+        if ((mir_input_event_get_type(input_event) == mir_input_event_type_key) &&
+            (mir_keyboard_event_action(mir_input_event_get_keyboard_event(input_event)) == mir_keyboard_action_repeat))
+                return;
```
wmww, line 120: I don't believe this is necessary. If it doesn't return here all it does is call `input_dispatcher->handle_event()` from the Wayland thread, `WaylandInputDispatcher::handle_event()` calls `WaylandInputDispatcher::handle_keyboard_event()` and that does nothing unless the action is `mir_keyboard_action_down` or `mir_keyboard_action_up`. Aside from negligibly improve performance, this logic would seem to have no effect. If it *did* do anything important, we would need to implement that in the `WaylandSurfaceObserver` as well.

AlanGriffiths:
Removed the stuff causing discussion. Landing the rest:

bors r=wmww

Issue #1604, opened 2020-07-14
Mir does not gracefully handle a missing XWayland


wmww:
## To reproduce:
Run `miral-shell` with `--enable-x11` and `--xwayland-path` set to anything invalid, then open any X11 app

## Expected
Polite, concise and informative error message

## Actual
It repeatedly prints the following, and stalls on shutdown for a number of seconds
```
[2020-07-14 17:36:44.245647] <information> xwayland: Starting XWayland
Failed to start XWayland, should be unreachable: No such file or directory
[2020-07-14 17:36:49.259371] < - ERROR - > xwayland: Spawning XWayland failed: /home/wmww/code/mir/src/server/frontend_xwayland/xwayland_server.cpp(197): Throw in function wl_client* {anonymous}::connect_xwayland_wl_client(const std::shared_ptr<mir::frontend::WaylandConnector>&, const mir::Fd&)
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: XWayland server failed to start
```

AlanGriffiths:
One part of the problem is that Xwayland is exec'd in a forked process, so the original process doesn't see the error. I'll PR a simple fix that handles most cases.

PR #1605, opened 2020-07-15
Better diagnostics when Xwayland can't be found or executed

AlanGriffiths:
Better diagnostics when Xwayland can't be found or executed. (Fixes: #1604)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -98,6 +98,7 @@ void exec_xwayland(
     args.push_back(nullptr);
 
     execvp(xwayland_path.c_str(), const_cast<char* const*>(args.data()));
+    perror(("Failed to execute Xwayland binary: xwayland_path='" + xwayland_path + "'").c_str());
 }
 
 auto fork_xwayland_process(
@@ -117,7 +118,7 @@ auto fork_xwayland_process(
 
     case 0:
         exec_xwayland(spawner, xwayland_path, wayland_pipe.client, x11_wm_pipe.server);
-        mir::log_error("Cannot execute Xwayland: --xwayland-path %s", xwayland_path.c_str());
+        // Only reached if Xwayland was not executed
         abort();
 
     default:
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
> It would be far better for X apps to not work than to refuse to let them log in.

Really? Wouldn't it be better to know about a problem straight away instead of carrying with the effect that random apps do not work and having to guess which log file has a useful error message?

wmww:
With Mir running in production, users wont necessarily see the text. What we *should* do is throw up an error message on-screen, but failing that IMO we should make do the best we can. Completely refusing to start Mir just because Xwayland is missing will lead to frustrated users, especially on systems where Mir is the only display server installed.

Saviq:
`enable-x11` is off by default, right? In that case I'm with @AlanGriffiths, if it was explicitly enabled on our command line, then it makes sense to bail. Dependencies should ensure that XWayland is around, and if it's not, we're not really in a position to be debugging that.

If the session fails altogether, it's going to be a clear message that they need to look in the session logs for what's gone wrong.

The two asks that I have is maybe having a clearer error message and a specific exit code? Especially in the second case, where the `exec` fails, is there an error message from the OS we could forward. And a session wrapper could deal with restarting without `--enable-x11` if they see the exit code telling it that's what's failed.

src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -117,7 +117,7 @@ auto fork_xwayland_process(
 
     case 0:
         exec_xwayland(spawner, xwayland_path, wayland_pipe.client, x11_wm_pipe.server);
-        perror("Failed to start XWayland, should be unreachable");
+        mir::log_error("Cannot execute Xwayland: --xwayland-path %s", xwayland_path.c_str());
```
wmww, line 120: Is using `log_error()` in another process a good idea? Even if so, I still want the errno message to be printed. Alternative here: fcf5af1a6329d9d460cbc71360a66e64fcf8a2b5

PR #1611, opened 2020-07-17
Add XWaylandServer::is_running()

wmww:
I've found this to be a more reliable way to check if the process is running than `kill(pid, 0) == 0`. It also results in the process being reaped, so we don't end up with a bunch of zombies. It's not safe to call `waitpid()` after it has returned that the process exited (the PID may be reused) so we only call it if the process is still marked as `running`.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -224,6 +224,12 @@ mf::XWaylandServer::~XWaylandServer()
             kill(xwayland_process.pid, SIGKILL);     // ...then kill it!
         }
     }
+
+    // Calling is_running() one more time will ensure the process is reaped
+    if (is_running())
+    {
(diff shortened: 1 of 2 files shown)


RAOF:
Hm. Could we just pull `mtf::Process` (and maybe some of the associated `mtf::fork_and_run_in_separate_process` machinery) out of the tests and into mircore rather than piecemeal reimplementing it?

AlanGriffiths:
> Hm. Could we just pull `mtf::Process` (and maybe some of the associated `mtf::fork_and_run_in_separate_process` machinery) out of the tests and into mircore rather than piecemeal reimplementing it?

That seems like a fair suggestion.

wmww:
I'll keep in mind that `mtf::Process` is the path forward for better process handling, but I don't see why that needs to block this PR. The amount of additional complexity added here is minimal, and I have changes in the pipeline that need this to land.

PR #1614, opened 2020-07-17
Improve and test wayland::Weak

wmww:
- Add tests for `wayland::Weak`
- Makes`LifetimeTracker::mark_destroyed()` protected
- Fix (and tests) `Weak == Weak` in the case of an old and new resource with the same address
- Add `Weak == Resource` operator (unlike previously discussed, it compares with a reference to a resource, not a nullable pointer)
- Add a `!=` operator (without it, != seems to implicitly cast both sides to bools?)
- Move `as_nullable_ptr()` to be a standalone function as Alan previously suggested

I'm not planning any more changes to `wayland::Weak` in the immediate future (but you never know what will come up)

### DIFF
--- a/tests/unit-tests/wayland/test_wayland_weak.cpp
+++ b/tests/unit-tests/wayland/test_wayland_weak.cpp
@@ -1,0 +1,202 @@
+/*
+ * Copyright © 2020 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: William Wold <william.wold@canonical.com>
+ */
+
+#include "mir/wayland/wayland_base.h"
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <experimental/optional>
+
+namespace mw = mir::wayland;
+
+using namespace testing;
+
+class MockResource : public virtual mw::LifetimeTracker
+{
+public:
+    // Expose protected method
+    void mark_destroyed() const
+    {
+        LifetimeTracker::mark_destroyed();
+    }
+};
+
+template<typename T>
+auto operator<<(std::ostream& out, mw::Weak<T> const& weak) -> std::ostream&
+{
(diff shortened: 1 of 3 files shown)


tests/unit-tests/wayland/test_wayland_weak.cpp
```diff
@@ -0,0 +1,202 @@
...
+public:
+    std::experimental::optional<MockResource> resource{{}};
+    std::experimental::optional<MockResource> resource_a{{}};
+    std::experimental::optional<MockResource> resource_b{{}};
+};
+
+TEST_F(WaylandWeakTest, returns_resource)
+{
+    mw::Weak<MockResource> const weak{&resource.value()};
+    ASSERT_THAT(&weak.value(), Eq(&resource.value()));
```
RAOF, line 56: We'd usually use `EXPECT_THAT`, right? Not that there's much difference in this case.

tests/unit-tests/wayland/test_wayland_weak.cpp
```diff
@@ -0,0 +1,202 @@
...
+}
+
+TEST_F(WaylandWeakTest, false_after_resource_destroyed)
+{
+    mw::Weak<MockResource> const weak{&resource.value()};
+    ASSERT_THAT(weak, Eq(true));
+    resource = std::experimental::nullopt;
+    ASSERT_THAT(weak, Eq(false));
```
RAOF, line 64: `EXPECT_THAT`…

tests/unit-tests/wayland/test_wayland_weak.cpp
```diff
@@ -0,0 +1,202 @@
...
+}
+
+TEST_F(WaylandWeakTest, can_be_created_with_nullptr)
+{
+    mw::Weak<MockResource> const weak{nullptr};
+    ASSERT_THAT(weak, Eq(false));
```
RAOF, line 70: `EXPECT_THAT`

tests/unit-tests/wayland/test_wayland_weak.cpp
```diff
@@ -0,0 +1,202 @@
...
+    mw::Weak<MockResource> const weak;
+    EXPECT_THROW(weak.value(), std::logic_error);
+}
+
+TEST_F(WaylandWeakTest, copies_are_equal)
+{
+    mw::Weak<MockResource> const weak_a{&resource.value()};
+    mw::Weak<MockResource> const weak_b{weak_a};
+    EXPECT_THAT(weak_a, Eq(weak_b));
+    EXPECT_THAT(weak_b, Eq(weak_a));
```
RAOF, line 91: Maybe for all of these we should be testing the full matrix? ie:
a == b : True
b == a : True
a != b: False
b != a : False
?

tests/unit-tests/wayland/test_wayland_weak.cpp
wmww: Oops

wmww:
Eh, on second thought

bors r+

Let me know if you want the tests tweaked differently and I'll make another PR

Issue #1617, opened 2020-07-20
KeyRepeatDispatcher related deadlock


wmww:
I've hit this twice on Arch but I haven't got a reliable reproducer yet. It looks like `device_state.repeat_alarm.reset();` (`key_repeat_dispatcher.cpp:170`) is causing a `TimerGSource` to lock. This is a problem if in another thread the `TimerGSource` is locked and that thread is waiting for the `KeyRepeatDispatcher` to lock. Here's the stack traces of the two relevant threads:
```
Thread 1 "miral-shell.bin" received signal SIGTSTP, Stopped (user).
0x00007f11719f3620 in __lll_lock_wait () from /usr/lib/libpthread.so.0
(gdb) bt
#0  0x00007f11719f3620 in __lll_lock_wait () at /usr/lib/libpthread.so.0
#1  0x00007f11719ebdf3 in pthread_mutex_lock () at /usr/lib/libpthread.so.0
#2  0x00007f1170d60843 in __gthread_mutex_lock(__gthread_mutex_t*) (__mutex=0x564bbb0f08e8) at /usr/include/c++/10.1.0/x86_64-pc-linux-gnu/bits/gthr-default.h:749
#3  0x00007f1170d60894 in std::mutex::lock() (this=0x564bbb0f08e8) at /usr/include/c++/10.1.0/bits/std_mutex.h:100
#4  0x00007f1170d608f0 in std::lock_guard<std::mutex>::lock_guard(std::mutex&) (this=0x7ffda8993e30, __m=...) at /usr/include/c++/10.1.0/bits/std_mutex.h:159
#5  0x00007f1170e132d0 in operator()() const (__closure=0x7f113c023790) at /home/wmww/code/mir/src/server/input/key_repeat_dispatcher.cpp:208
#6  0x00007f1170e13ce4 in std::__invoke_impl<void, mir::input::KeyRepeatDispatcher::handle_key_input(MirInputDeviceId, const MirKeyboardEvent*)::<lambda()>&>(std::__invoke_other, struct {...} &) (__f=...)
    at /usr/include/c++/10.1.0/bits/invoke.h:60
#7  0x00007f1170e13b7b in std::__invoke_r<void, mir::input::KeyRepeatDispatcher::handle_key_input(MirInputDeviceId, const MirKeyboardEvent*)::<lambda()>&>(struct {...} &) (__fn=...)
    at /usr/include/c++/10.1.0/bits/invoke.h:153
[...]

Issue #1607, opened 2020-07-16
[wayland] keyboard state of modifier keys gets out of sync 


AlanGriffiths:
When the focus switches between windows, we can end up sending events to a window that saw a key down event but has missed a corresponding up event.

Commonly affected are Shift and Alt (but that may just reflect my use of the keyboard).

This is easy to see with CLion (via X11) as that uses a lot of pop-ups while editing code but can also be seen with Firefox directly on Wayland.

This may to be associated with the "Sending key to wl_surface@24 even though it was not explicitly given keyboard focus" warning message appearing in the logs.

This is distinct from #1515 which affects keyboard events throughout the system (including internal clients).

AlanGriffiths:
I encountered this again (with a bit of debug information).

Monitoring the state of modifiers in WlKeyboard, the state here is wrong.

1.  Monitoring showed the Ctrl key pressed and tapping the key has no effect.
2. After reinstating the debug code in `XkbMappingState`, I encountered the same with LeftShift. here `XkbMappingState` reports the tap and release correctly, WlKeyboard doesn't see it.

PR #1619, opened 2020-07-21
[input] Don't lock the repeat dispatcher while repeating events

AlanGriffiths:
[input] Don't lock the repeat dispatcher while repeating events. (Fixes: #1617,  #1607)


AlanGriffiths:
To be fair, I don't have a reliable enough reproducer to be certain this fixes the linked bugs. But it does address the deadlock for which @wmww posted the backtrace. Also I've not seen either bug so far while trying to reproduce.

I will continue using this patch on my shell today and see what happens.

AlanGriffiths:
> I don't like the fact that we're `.reset()`ing the timer under lock. That was part of the combination of issues that caused the deadlock.

Yes, I thought about reducing the lock scope, but that goes beyond the fix we need. (And `device_state` needs some protection.)

PR #1620, opened 2020-07-21
Minimize locking in KeyRepeatDispatcher

wmww:
Prevents future issues like #1617

A bit unrelated but also adds a test for key repeats not getting fired for meta keys.

### DIFF
--- a/src/server/input/key_repeat_dispatcher.cpp
+++ b/src/server/input/key_repeat_dispatcher.cpp
@@ -125,10 +125,19 @@ void mi::KeyRepeatDispatcher::remove_device(MirInputDeviceId id)
         touch_button_device.consume();
 }
 
-mi::KeyRepeatDispatcher::KeyboardState& mi::KeyRepeatDispatcher::ensure_state_for_device_locked(std::lock_guard<std::mutex> const&, MirInputDeviceId id)
+void mi::KeyRepeatDispatcher::set_alarm_for_device(
+    MirInputDeviceId id,
+    std::shared_ptr<mir::time::Alarm> repeat_alarm)
 {
-    repeat_state_by_device.insert(std::make_pair(id, KeyboardState()));
-    return repeat_state_by_device[id];
+    {
+        std::lock_guard<std::mutex> lg(repeat_state_mutex);
+        // Get the current KeyboardState or insert a new one
+        auto const& iter = repeat_state_by_device.insert(std::make_pair(id, KeyboardState()));
+        // Set the new alarm
+        std::swap(repeat_alarm, iter.first->second.repeat_alarm);
+    }
+    // Reset the old alarm outside of the mutex lock
+    repeat_alarm.reset();
 }
 
 bool mi::KeyRepeatDispatcher::dispatch(std::shared_ptr<MirEvent const> const& event)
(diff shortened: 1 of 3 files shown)


src/server/input/key_repeat_dispatcher.cpp
```diff
@@ -158,16 +158,19 @@ bool mi::KeyRepeatDispatcher::dispatch(std::shared_ptr<MirEvent const> const& ev
...
     case mir_keyboard_action_up:
     {
-        device_state.repeat_alarm.reset();
+        std::shared_ptr<time::Alarm> alarm;
+        {
+            std::lock_guard<std::mutex> lg(repeat_state_mutex);
+            auto& device_state = ensure_state_for_device_locked(lg, id);
+            alarm = std::move(device_state.repeat_alarm);
+        }
+        alarm.reset();
```
AlanGriffiths, line 173: Looks like there's a repeating pattern here. And one that could be further generalized to:
```c++
auto const exchange_alarm_for_device = [this](MirInputDeviceId id, std::shared_ptr<time::Alarm>& alarm)
    {
            std::lock_guard<std::mutex> lg(repeat_state_mutex);
            std::swap(alarm, ensure_state_for_device_locked(lg, id).repeat_alarm);
    };
```
Which could also cover the setting case.

tests/unit-tests/input/test_key_repeat_dispatcher.cpp
```diff
@@ -243,3 +251,14 @@ TEST_F(KeyRepeatDispatcherOnArale, repeat_for_regular_keys)
     dispatcher.dispatch(a_key_down_event());
     alarm_function();
 }
+
+TEST_F(KeyRepeatDispatcherOnArale, no_repeat_for_meta_keys)
```
AlanGriffiths, line 255: FWIW there's already `TestClientInputKeyRepeat.meta_keys_are_not_repeated_to_clients`

wmww: Ah, so there is. I'll drop it if you want, but I don't imagine it does any harm to have a unit test as well.

src/server/input/key_repeat_dispatcher.cpp
wmww: I came to the conclusion `ensure_state_for_device_locked()` is the wrong abstraction. Since we only use it in the context of setting the repeat alarm, I've changed it to `set_repeat_alarm()`.

src/server/input/key_repeat_dispatcher.cpp
```diff
@@ -158,16 +167,13 @@ bool mi::KeyRepeatDispatcher::dispatch(std::shared_ptr<MirEvent const> const& ev
...
-    auto& device_state = ensure_state_for_device_locked(lg, id);
-
     auto scan_code = mir_keyboard_event_scan_code(kev);
 
     switch (mir_keyboard_event_action(kev))
     {
     case mir_keyboard_action_up:
     {
-        device_state.repeat_alarm.reset();
+        set_alarm_for_device(id, nullptr);
```
AlanGriffiths, line 176: I've a mild preference for `{}` over `nullptr` but that's fine.

PR #1623, opened 2020-07-22
[wayland] don't send/apply input events after rebuilding state

AlanGriffiths:
[wayland] don't send/apply input events after rebuilding state. (Fixes: #1607)

### DIFF
--- a/src/server/frontend_wayland/wl_keyboard.cpp
+++ b/src/server/frontend_wayland/wl_keyboard.cpp
@@ -71,25 +71,29 @@ mf::WlKeyboard::~WlKeyboard()
 
 void mf::WlKeyboard::key(std::chrono::milliseconds const& ms, WlSurface* surface, int scancode, bool down)
 {
+    auto const serial = wl_display_next_serial(wl_client_get_display(client));
+    auto const wayland_state = down ? KeyState::pressed : KeyState::released;
+    send_key_event(serial, ms.count(), scancode, wayland_state);
+
     if (!focused_surface || &focused_surface.value() != surface)
     {
         log_warning(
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
Note: to unblock 2.0 I've added this in rc5.

I know this doesn't solve all our focus issues (e.g. #1625) but it works better for me than the status quo.

wmww:
So sorry. I meant to review this right away yesterday, but I lost the tab somehow and forgot about it.

wmww:
I don't understand *why* any of this is a good idea. Why is Mir sending events to the wrong surface? Why is the correct response to send the currently focused surface (which Mir didn't request), but still change the focus? Why is modifier state only updated if the requested surface is currently focused?

AlanGriffiths:
> I don't understand why any of this is a good idea.

Because it improves the behaviour?

> Why is Mir sending events to the wrong surface? 

I've yet to figure that out. I know we have a focus tracking issue, and all this was, and is, a workaround.

> Why is the correct response to send the currently focused surface (which Mir didn't request), but still change the focus? 

If we don't then (empirically) we end up with the client autorepeating keys.

> Why is modifier state only updated if the requested surface is currently focused?

The modifier state tracks the state of the hardware and needs to be updated with input events. Unless it is resync'd with the hardware.

If we call `focussed()` then we read the *current* hardware status which already reflects current event. The previous workaround got this wrong and processed, for example, a key down for an already pressed key. Multiple key down events is "legal" according to XKB, but the accounting then requires multiple key releases to unpress the key (which never happens and leads to the "locked" keys we sometimes saw).

PR #1633, opened 2020-07-27
[mir_performance_tests] Do not recommend glmark2-es2-mir

AlanGriffiths:
[mir_performance_tests] Do not recommend glmark2-es2-mir.

We don't use it, drop it from depends.


RAOF:
Yup.

Also,
bors r+

Issue #1466, opened 2020-05-06
eglstream: Failed to restore saved crtc


Saviq:
I've noticed these errors on the Nvidia backend:

```logs
< - ERROR - > eglstream: Failed to restore saved crtc: Dynamic exception type: std::exception
```

While it's not fatal, it does look scary.

```irc
<RAOF> Yeah, that's not a huge problem. I wonder why we can't restore, though?
-*- RAOF wonders if we're using the wrong API there…
<RAOF> Oh, or maybe it's because that's just not hooked up to anything.
```

PR #1640, opened 2020-07-29
eglstream-kms: Actually save CRTC on startup

RAOF:
Since we try to restore it on shutdown we should probably save it on startup!

Fixes: #1466


AlanGriffiths:
@Saviq as you raised the original bug, are you set up to try this?

RAOF:
FWIW I tested this on *my* nvidia system and it no longer complains. But my system is somewhat atypical, in that the nvidia card doesn't normally have a framebuffer *at all*.

AlanGriffiths:
bors merge

Saviq:
> @Saviq as you raised the original bug, are you set up to try this?

Also FWIW, I only saw this in the CI lab, don't have Nvidia here.

Will run it there again to see if it's gone.

Issue #1642, opened 2020-07-29
[mir_performance_tests] segfaults on PROX15


AlanGriffiths:
This is master, on a hybrid laptop running 20.04 and happens when running on logind console services. (Doesn't happen on other kit, or using minimal or LVT.)

```plain
$ gdb --args cmake-build-debug/bin/mir_performance_tests --logging on --gtest_filter=GLMark2Wayland.fullscreen
GNU gdb (Ubuntu 9.1-0ubuntu1) 9.1
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from cmake-build-debug/bin/mir_performance_tests...
(gdb) r
Starting program: /home/alan/CLionProjects/mir/cmake-build-debug/bin/mir_performance_tests --logging on --gtest_filter=GLMark2Wayland.fullscreen
MIR_CLIENT_PLATFORM_PATH=/home/alan/CLionProjects/mir/cmake-build-debug/bin/../lib/client-modules/
MIR_SERVER_PLATFORM_PATH=/home/alan/CLionProjects/mir/cmake-build-debug/bin/../lib/server-modules/
LD_LIBRARY_PATH=/home/alan/CLionProjects/mir/cmake-build-debug/bin/../lib
exec=/home/alan/CLionProjects/mir/cmake-build-debug/bin/mir_performance_tests.bin
process 6161 is executing new program: /home/alan/CLionProjects/mir/cmake-build-debug/bin/mir_performance_tests.bin
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff5cc2700 (LWP 6165)]
[New Thread 0x7ffff54c1700 (LWP 6166)]
Running main() from main.cpp
Note: Google Test filter = GLMark2Wayland.fullscreen
[...]

PR #1645, opened 2020-07-30
Don't frig with the shutdown sequence in AsyncServerRunner

AlanGriffiths:
Don't frig with the shutdown sequence in AsyncServerRunner. (Fixes: #1642)

Moves a workaround for a problem with stubbed graphics that can cause issues with real graphics.

### DIFF
--- a/tests/mir_test_framework/async_server_runner.cpp
+++ b/tests/mir_test_framework/async_server_runner.cpp
@@ -94,12 +94,6 @@ void mtf::AsyncServerRunner::add_to_environment(char const* key, char const* val
     env.emplace_back(key, value);
 }
 
-namespace
-{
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
[LP:1728931](https://bugs.launchpad.net/ubuntu/+source/mir/+bug/1728931) says the following fails "after a few seconds", I've run it for 10 minutes.

```plain
while bin/mir_acceptance_tests --gtest_filter=ServerDisconnect.is_detected_by_client; do :; done
```

AlanGriffiths:
Oh! Trying that test case on slower hardware reproduce LP:1728931. :(

AlanGriffiths:
@RAOF NB this fixes a problem @Saviq reported from the lab.

Issue #1625, opened 2020-07-23
[Wayland] Menus do not work from keyboard on some apps


AlanGriffiths:
Two examples: Firefox and Thunderbird

This is likely a focus assignment issue (and I see the "not explicitly given focus" warning while this is happening).

wmww:
Likely a duplicate of #827

AlanGriffiths:
Here's a log of events when pulling up a thunderbird menu:
```plain
[3511394.204] wl_compositor@4.create_surface(new id wl_surface@70)
[3511394.317] xdg_wm_base@13.create_positioner(new id xdg_positioner@77)
[3511394.338] xdg_positioner@77.set_size(213, 408)
[3511394.352] xdg_positioner@77.set_anchor_rect(0, 63, 213, 408)
[3511394.375] xdg_positioner@77.set_offset(0, 0)
[3511394.389] xdg_positioner@77.set_anchor(5)
[3511394.398] xdg_positioner@77.set_gravity(8)
[3511394.409] xdg_positioner@77.set_constraint_adjustment(63)
[3511394.422] xdg_wm_base@13.get_xdg_surface(new id xdg_surface@65, wl_surface@70)
[3511394.446] xdg_surface@65.get_popup(new id xdg_popup@79, xdg_surface@32, xdg_positioner@77)
[3511394.526] xdg_positioner@77.destroy()
[3511394.558]  -> wl_display@1.delete_id(77)
[3511394.567] wl_surface@70.commit()
[2020-07-30 14:40:37.803429] <information> miral::Window Management: place_new_window app_info={application=, windows={Inbox - alan.griffiths@canonical.com - Mozilla Thunderbird}}, requested_specification={name=, type=freestyle, top_left=(0, 63), size=(213, 408), output_id=0, state=restored, aux_rect=((0, 63), (213, 408)), placement_hints=12, window_placement_gravity=5, aux_rect_placement_gravity=5, aux_rect_placement_offset=(0, 0), parent=Inbox - alan.griffiths@canonical.com - Mozilla Thunderbird, exclusive_rect=0} -> {name=, type=freestyle, top_left=(0, 63), size=(213, 408), output_id=0, state=restored, aux_rect=((0, 63), (213, 408)), placement_hints=12, window_placement_gravity=5, aux_rect_placement_gravity=5, aux_rect_placement_offset=(0, 0), parent=Inbox - alan.griffiths@canonical.com - Mozilla Thunderbird, exclusive_rect=0}
[...]

PR #1646, opened 2020-07-30
[Wayland] Fix focus issues with popups

AlanGriffiths:
[Wayland] Fix focus issues with popups. (Fixes: #1625)

A popup ought to be handled as either a `mir_window_type_gloss` or a `mir_window_type_menu` depending upon whether it is configured with a `grab`.

This avoids unexpectedly shifting the focus to an ungrabbed popup.


wmww:
Oops
bors r+

AlanGriffiths:
> We're still not dismissing grabbing popups when the user clicks on another window though.

Raise an issue or PR a fix (lest we forget)

PR #1650, opened 2020-08-03
platform: Drop the NativeDisplay interface

RAOF:
This is only used to get a `gl::ContextSource` out of a `Display`, which we can just do a direct `dynamic_cast<>` for now.

In the near future this will be moved to the `RenderingPlatform` instead, as a part of allowing multiple display platforms to load, but for now leave it where it is.

### DIFF
--- a/include/platform/mir/graphics/display.h
+++ b/include/platform/mir/graphics/display.h
@@ -20,6 +20,7 @@
 #define MIR_GRAPHICS_DISPLAY_H_
 
 #include "mir/graphics/frame.h"
+#include "mir/renderer/gl/context_source.h"
 #include <memory>
 #include <functional>
 #include <chrono>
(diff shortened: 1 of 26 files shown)


AlanGriffiths:
A shame about dispmanx-rpi:
```plain
In file included from /spread/mir/src/platforms/rpi-dispmanx/display.cpp:30:0:
/spread/mir/src/platforms/rpi-dispmanx/display.h:39:7: error: direct base â€˜mir::renderer::gl::ContextSourceâ€™ inaccessible in â€˜mir::graphics::rpi::Displayâ€™ due to ambiguity [-Werror]
 class Display
       ^~~~~~~
cc1plus: all warnings being treated as errors
```

RAOF:
Bah! I did half of that - removing the `NativeDisplay` implementation, but forgot that bit!

PR #1651, opened 2020-08-03
Disable acceptance_tests on arm64

AlanGriffiths:
Disable acceptance_tests on arm64 PPA builds. (Fixes: #1505)

Some of these tests appear to fail only because arm64 builders are too slow for hardcoded realtime waits.


AlanGriffiths:
I don't think it worth fixing these test to avoid realtime dependencies as they are based on the obsolete mirclient API. (And they are passing other architectures.)

Appropriate tests of the functionality should be via miral-tests and wlcs (if/when the functionality is exposed).

AlanGriffiths:
> I'm a bit surprised the arm64 builders are the ones which expose this - I'd expect armhf to be at least as slow.

We would need to *enable* the acceptance and integration tests on armhf to find out.

PR #1654, opened 2020-08-04
platform/frontend: Drop MirClientExtension infrastructure 

RAOF:
This is, as its name suggests, only used for mirclient. This removes all the acceptance tests which use the extension mechanism, as they're *also* only testing mirclient functionality, and then removes the `Platform` interfaces used to support it.


RAOF:
(Stacked-on #1650)

PR #1655, opened 2020-08-04
Extend the timeout for TestClientInput setup

AlanGriffiths:
Extend the timeout for TestClientInput setup. (This was one of the failures on arm64 and appears intermittently on amd64)

This is just a timeout that prevents the test hanging, so a large value has no effect on normal execution.


debian/source/format
```diff
@@ -1 +0,0 @@
-1.0
```
RAOF, line 1: I'm *reasonably* sure you didn't mean to delete `debian/source/format`

AlanGriffiths:
bors r=RAOF

debian/source/format
AlanGriffiths: Thank you for being so understanding

PR #1658, opened 2020-08-04
The AdorningDisplayBufferCompositor doesn't work... drop it

AlanGriffiths:


AlanGriffiths:
"Doesn't work" means:
```plain
$ env -u WAYLAND_DISPLAY mir_demo_server --custom-compositor adorning
...
ERROR: /build/mir-v0n9SW/mir-2.0.0.0+dev52-gb7d31e49f4/examples/mir_demo_server/server_example_adorning_compositor.cpp(239): Throw in function virtual void mir::examples::AdorningDisplayBufferCompositor::composite(mir::compositor::SceneElementSequence&&)
Dynamic exception type: boost::wrapexcept<std::logic_error>
std::exception::what: Buffer does not support GL rendering
```

RAOF:
It would be relatively simple to make this work - probably an afternoon to a day's work - but I'm not sure if we want to invest even that much effort in it.

AlanGriffiths:
> It would be relatively simple to make this work - probably an afternoon to a day's work - but I'm not sure if we want to invest even that much effort in it.

Exactly. This isn't how we plan to support this functionality.

RAOF:
Still approve.

bors r+

Issue #1632, opened 2020-07-27
`mir_performance_tests` fail in the lab on Nvidia gfx


Saviq:
Given [cert-focal-mir-201405-15065.txt](https://github.com/MirServer/mir/files/4982311/cert-focal-mir-201405-15065.txt) as the submitted job I'm getting failures in the lab on Nvidia graphics:

```
Waiting for mir tests to complete...
Running main() from main.cpp
[==========] Running 5 tests from 3 test suites.
[----------] Global test environment set-up.
[----------] 2 tests from GLMark2Wayland
[ RUN      ] GLMark2Wayland.fullscreen
[2020-07-27 12:30:47.560168] <information> mircommon: Loading modules from: /usr/lib/x86_64-linux-gnu/mir/server-platform
[2020-07-27 12:30:47.560324] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/server-x11.so.17
[2020-07-27 12:30:47.560360] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-eglstream-kms.so.17
[2020-07-27 12:30:47.560380] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/input-evdev.so.7

[2020-07-27 12:30:47.562087] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-dummy.so
[2020-07-27 12:30:47.562130] <information> mircommon: Loading module: /usr/lib/x86_64-linux-gnu/mir/server-platform/input-stub.so
[2020-07-27 12:30:47.577241] < - debug - > mirserver: Using logind for session management
[2020-07-27 12:30:47.577610] <information> mirserver: Found graphics driver: mir:x11 (version 2.0.0) Support priority: 0
[2020-07-27 12:30:47.586812] < - debug - > eglstream: Found EGLDeviceEXT with device extensions: EGL_NV_device_cuda EGL_EXT_device_drm
[2020-07-27 12:30:48.077397] <information> mirserver: Found graphics driver: mir:eglstream-kms (version 2.0.0) Support priority: 256
[2020-07-27 12:30:48.111440] <information> mirserver: Found graphics driver: mir:stub-graphics (version 2.0.0) Support priority: 1
[2020-07-27 12:30:48.111916] <information> mirserver: Selected driver: mir:eglstream-kms (version 2.0.0)
[...]

Issue #1653, opened 2020-08-04
CompositorPerformance test uses its own idiosyncratic server runner


RAOF:
`performance_tests/test_compositor.cpp` uses its own, weird server runner involving parsing the output of Mir's `stdout`. Mostly this is harmless, albeit weird, but it means that it errors out when the `MIR_SERVER_LOGGING` is set in the environment.

With a few small and reasonably sensible changes to the CompositorReport infrastructure this could just be a normal `AsyncTestRunner` test.

PR #1659, opened 2020-08-05
performance-tests: Hacky fix for MIR_SERVER_LOGGING

RAOF:
(Really) Fixes: #1632, #1631


RAOF:
Yeah, I might also look at #1653, but it doesn't really seem urgent, given this low-cost hack.
5 Aug 2020 18:19:45 Alan Griffiths <notifications@github.com>:

Issue #1665, opened 2020-08-10
[X11] Menus don't work in chromium


AlanGriffiths:
(To get chromium snap to work on 2.0 you need [the workaround described on the forum](https://discourse.ubuntu.com/t/gtk-backend-selection-or-why-gtk-cannot-open-display-0/17657/).

Activating menus using the keyboard, or by using the mouse to click on the dot menu...

Expect: a menu is displayed;
Actual: a surface is briefly displayed and then disappears.

AlanGriffiths:
Here's the sequence of events:
```plain
[1323500.003] wl_keyboard@14.key(34, 31784585, 33, 1)
[1323523.164]  -> wl_compositor@5.create_surface(new id wl_surface@28)
[1323523.483]  -> wl_pointer@13.set_cursor(4, wl_surface@12, 10, 5)
[1323523.564]  -> wl_surface@12.attach(wl_buffer@16, 0, 0)
[1323523.614]  -> wl_surface@12.damage(0, 0, 32, 32)
[1323523.675]  -> wl_surface@12.frame(new id wl_callback@27)
[1323523.706]  -> wl_surface@12.commit()
[1323523.778]  -> wl_drm@4.create_prime_buffer(new id wl_buffer@29, fd 14, 305, 404, 875713112, 0, 1280, 0, 0, 0, 0)
[1323523.944]  -> wl_surface@28.attach(wl_buffer@29, 0, 0)
[1323523.993]  -> wl_surface@28.damage(0, 0, 305, 404)
[1323524.054]  -> wl_surface@28.frame(new id wl_callback@31)
[1323525.131]  -> wl_surface@28.commit()
[2020-08-10 14:55:32.673978] <information> miral::Window Management: place_new_window app_info={application=, windows={Linux Hardware Reviews, Open-Source Benchmarks & Linux Performance - Phoronix - Chromium, }}, requested_specification={name=, type=freestyle, top_left=(487, 176), size=(305, 404), output_id=0, state=restored, exclusive_rect=0} -> {name=, type=freestyle, top_left=(487, 176), size=(305, 404), output_id=0, state=restored, exclusive_rect=0}
[2020-08-10 14:55:32.674493] <information> miral::Window Management: advise_new_window window_info={name=, type=freestyle, state=restored, top_left=(487, 176), size=(305, 404), children={}, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0, output_id=0}
[...]

PR #1671, opened 2020-08-12
[X11] Better handling of popups and menus

AlanGriffiths:
Process some of the X11 window types and rework the logic to identify parents. (Fixes: #1665)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -129,6 +129,12 @@ auto property_handler(
             return connection->read_property(window, property, handler, on_error);
         });
 }
+
+// We don't use all the Mir types and we do try to parent menus
+bool needs_parent(MirWindowType mir_window_type)
+{
+    return mir_window_type == mir_window_type_gloss || mir_window_type == mir_window_type_menu;
+}
 }
 
 mf::XWaylandSurface::XWaylandSurface(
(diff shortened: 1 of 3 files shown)


src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -860,93 +849,23 @@ void mf::XWaylandSurface::is_transient_for(xcb_window_t transient_for)
...
-                "Failed to find a window for %s to be transient for",
-                connection->window_debug_string(window).c_str());
-        }
+        set_parent(transient_for, std::lock_guard<std::mutex>{mutex});
     }
     else if (verbose_xwayland_logging_enabled())
     {
         log_debug(
             "%s is not transient",
             connection->window_debug_string(window).c_str());
```
wmww, line 867: Parent is not cleared when set to `XCB_WINDOW_NONE`. Is this intentional?

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -860,93 +849,23 @@ void mf::XWaylandSurface::is_transient_for(xcb_window_t transient_for)
 {
     std::shared_ptr<scene::Surface> parent_scene_surface; // May remain nullptr
```
wmww, line 850: `parent_scene_surface` is now unused (not sure why the compiler doesn't complain)

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -1047,3 +966,144 @@ auto mf::XWaylandSurface::latest_input_timestamp(std::lock_guard<std::mutex> con
...
+        {
+            log_debug("Ignoring type (%s) of %s",
+                      this->connection->query_name(wm_type).c_str(),
+                      this->connection->window_debug_string(window).c_str());
+        }
+    }
+}
+
+void mf::XWaylandSurface::set_parent(xcb_window_t xcb_window, std::lock_guard<std::mutex> const& lock)
+{
```
wmww, line 1036: I think the error conditions should explicitly clear the parent (or the parent should be cleared at the start of the function, and stay cleared unless it is set)

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -1047,3 +966,144 @@ auto mf::XWaylandSurface::latest_input_timestamp(std::lock_guard<std::mutex> con
...
+
+void mf::XWaylandSurface::window_type(xcb_atom_t wm_type)
+{
+    auto set_type = [this](MirWindowType type)
+        {
+            std::lock_guard<std::mutex> lock{mutex};
+            pending_spec(lock).type = type;
+        };
+
+    if (this->connection->net_wm_window_type_normal == wm_type)
```
wmww, line 1010: Any reason for the explicit `this->`?

PR #1675, opened 2020-08-14
New mir-wlcs-integration package

AlanGriffiths:
Split the wlcs related libraries out of mir-test-tools and into a new mir-wlcs-integration package

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -232,8 +232,8 @@ Description: Display Server for Ubuntu - test tools
 Package: mir-wlcs-integration
 Architecture: linux-any
 Pre-Depends: ${misc:Pre-Depends}
-Breaks: mir-test-tools (<< 2.1)
-Replaces: mir-test-tools (<< 2.1)
+Breaks: mir-test-tools (<< 2.0.0.0+dev148~)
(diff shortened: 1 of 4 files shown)


debian/control
```diff
@@ -224,11 +224,24 @@ Recommends: mir-demos,
...
  Mir is a display server running on linux systems, with a focus on efficiency,
  robust operation and a well-defined driver model.
  .
- Contains a tool for stress testing the Mir display server
+ Contains tools for smoke and performance testing the Mir display server
+
+Package: mir-wlcs-integration
+Architecture: linux-any
+Pre-Depends: ${misc:Pre-Depends}
+Breaks: mir-test-tools (<< 2.1)
```
RAOF, line 236: Hm. I'm reasonably certain that this is not going to be the right version, because this will matches the *new* package, too (so it'll break itself, and never resolve).

Because it's going to be difficult to get the version numbering right as it's dependent on when it's borsed, I'll fix this up and bors it.

RAOF:
Updated to resolve conflicts

RAOF (approved):

PR #1681, opened 2020-08-19
c++17

AlanGriffiths:


CMakeLists.txt
```diff
@@ -62,12 +62,12 @@ set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "${build_types}" FORCE)
 set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "${build_types}")
 
 # Standard must be set both this way and in the flags below, to cover all cmake versions
```
RAOF, line 64: Hm. What CMake versions *don't* support `CMAKE_CXX_STANDARD`?

RAOF: Woot! [This](https://crascit.com/2015/03/28/enabling-cxx11-in-cmake/) indicates that CMake 3.8 introduced `CMAKE_CXX_STANDARD 17` support, and 18.04 has CMake 3.10. Looks like this comment is outdated.

CMakeLists.txt
```diff
@@ -62,12 +62,12 @@ set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "${build_types}" FORCE)
...
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread -g -Wall -pedantic -Wextra -fPIC")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -g -std=c++14 -Wall -fno-strict-aliasing -pedantic -Wnon-virtual-dtor -Wextra -fPIC")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -g -std=c++17 -Wall -fno-strict-aliasing -pedantic -Wnon-virtual-dtor -Wextra -fPIC")
```
RAOF, line 70: Let's drop the `-std=c++17` here, as we no longer need it!

RAOF (approved):
LGTM, once the workaround for `CMake`s we no longer need to support is removed :+1:

CMakeLists.txt
AlanGriffiths: > Let's drop the `-std=c++17` here, as we no longer need it!

I've not (yet) dug into why, but that breaks linking on 18.04.

AlanGriffiths:
bors r=wmww,RAOF

PR #1684, opened 2020-08-20
Implement wlr_foreign_toplevel_management_unstable_v1

wmww:
Implement `wlr_foreign_toplevel_management_unstable_v1`, which allows taskbars and application switchers. There is a proof-of-concept MATE Panel build that uses this protocol, and Sway implements it so we can probably expect to see more users in the future.

Requires #1683 before it can pass WLCS (its tests have landed but are not in the version of WLCS that's run in Mir's CI).


AlanGriffiths:
Clang missing some symbol exports?
```plain
/usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/foreign_toplevel_manager_v1.cpp.o:(.data.rel.ro+0x328): undefined reference to `virtual thunk to mir::wayland::ForeignToplevelManagerV1::~ForeignToplevelManagerV1()'
/usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/foreign_toplevel_manager_v1.cpp.o:(.data.rel.ro+0x330): undefined reference to `virtual thunk to mir::wayland::ForeignToplevelManagerV1::~ForeignToplevelManagerV1()'
/usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/foreign_toplevel_manager_v1.cpp.o:(.data.rel.ro+0x430): undefined reference to `virtual thunk to mir::wayland::ForeignToplevelHandleV1::~ForeignToplevelHandleV1()'
/usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/foreign_toplevel_manager_v1.cpp.o:(.data.rel.ro+0x438): undefined reference to `virtual thunk to mir::wayland::ForeignToplevelHandleV1::~ForeignToplevelHandleV1()'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[3]: *** [src/server/CMakeFiles/mirserver.dir/build.make:497: lib/libmirserver.so.54] Error 1
```

wmww:
Ready for another look

src/server/frontend_wayland/foreign_toplevel_manager_v1.h
```diff
@@ -0,0 +1,58 @@
...
+
+    std::shared_ptr<shell::Shell> const shell;
+    std::shared_ptr<Executor> const wayland_executor;
+    std::shared_ptr<SurfaceStack> const surface_stack;
+
+private:
+    void bind(wl_resource* new_resource) override;
+};
+}
+}
```
AlanGriffiths, line 56: Do we need all this in a header? I think all we need is a function like `create_xdg_output_manager_v1()`. (Actually, even that could be simplified to return `shared_ptr<void>`!)

AlanGriffiths (changes requested):
More nitpicking (sorry).

We should also bump MIRAL_VERSION_MINOR

Otherwise, looks good

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -0,0 +1,685 @@
...
+    auto const session = surface ? surface->session().lock() : nullptr;
+    if (shell && session && surface)
+    {
+        try
+        {
+            operation(shell, session, surface);
+        }
+        catch (std::out_of_range const&)
+        {
+            log_warning("Attempted operation on invalid surface");
```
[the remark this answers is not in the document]
wmww: This catch is no longer necessary. Instead of throwing `std::out_of_range`, `BasicWindowManager` now uses `BasicWindowManager::surface_known()` to print a more useful error and return. We can completely ignore the error case here.

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -0,0 +1,685 @@
...
+                }
+            });
+    }
+}
+
+void mf::ForeignSurfaceObserver::create_or_close_toplevel_handle_as_needed(std::lock_guard<std::mutex>& lock)
+{
+    bool should_have_handle = true;
+
+    auto const surface = this->weak_surface.lock();
```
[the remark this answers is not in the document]
wmww: Nope

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -0,0 +1,696 @@
...
+    bool should_have_handle = true;
+
+    auto const surface = weak_surface.lock();
+    if (surface)
+    {
+        switch(surface->state())
+        {
+        case mir_window_state_attached:
+        case mir_window_state_hidden:
+            should_have_handle = false;
```
RAOF, line 352: This does raise the interesting question: should windows on other workspaces show up in the switcher, and how can that be accomodated? (I *think* the answer is “with a protocol revision”?)

RAOF (changes requested):
The format string needs to be fixed, you might as well fix the spelling at the same time, and the question is just a question :)

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
wmww: Seeing as the protocol makes no mention of workspaces, I think they should show up for now. If you wanted to filter based on workspaces then that would indeed require updates to the protocol.

AlanGriffiths: The way that miral-shell and egmde implement workspaces is to hide the windows of inactive workspaces. (But the Mir implementation of "workspaces" is just keeping lists of windows, the policy is what gives those lists meaning.)

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -0,0 +1,696 @@
...
+    create_surface_observer(surface);
+}
+
+void mf::ForeignSceneObserver::surface_removed(std::shared_ptr<scene::Surface> const& surface)
+{
+    std::lock_guard<std::mutex> lock{mutex};
+    auto const iter = surface_observers.find(surface);
+    if (iter == surface_observers.end())
+    {
+        log_error("Can not remove ForeignSurfaceObserver: surface %p not in observers map", surface.get());
```
AlanGriffiths, line 252: Non blocking, but is logging `surface.get()` useful? (Maybe we need some standard approach to identifying surfaces in the log: Name (if set), client session etc.)

AlanGriffiths (approved):
LGTM

bors r+

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
AlanGriffiths, line 252: This (`%p`) needs an explicit cast to `void*`, and after #1686 the compiler tells us so.

AlanGriffiths (changes requested):

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
wmww: Done

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -0,0 +1,698 @@
...
+
+void mf::ForeignSceneObserver::create_surface_observer(std::shared_ptr<scene::Surface> const& surface)
+{
+    std::lock_guard<std::mutex> lock{mutex};
+    auto observer = std::make_shared<ForeignSurfaceObserver>(wayland_executor, manager, surface);
+    surface->add_observer(observer);
+    auto insert_result = surface_observers.insert(std::make_pair(surface, observer));
+    if (!insert_result.second)
+    {
+        log_error("Can not add ForeignSurfaceObserver: surface %p already in the observers map");
```
AlanGriffiths, line 281: Another unsatisfied %p

AlanGriffiths (approved):
bors r+

PR #1686, opened 2020-08-25
log: Annotate log functions with [[gnu::format]] attributes

RAOF:
And then fix all the places where we were accidentally invoking nasal daemons.

### DIFF
--- a/include/common/mir/log.h
+++ b/include/common/mir/log.h
@@ -27,9 +27,10 @@
 
 namespace mir
 {
-
+[[gnu::format (printf, 3, 0)]]
 void logv(logging::Severity sev, const char *component,
           char const* fmt, va_list va);
+[[gnu::format (printf, 3, 4)]]
 void log(logging::Severity sev, const char *component,
          char const* fmt, ...);
 void log(logging::Severity sev, const char *component,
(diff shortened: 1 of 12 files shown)


RAOF:
Turns out we actually have a bunch of Undefined Behaviour™ in our logging. Notably, there were some log calls which have a mismatching number of format specifiers and actual arguments (eep!), and it turns out that passing a non `void*` to the `%p` format specifier is *technically* UB.

RAOF:
I thought about adding some “this fails to compile” tests, but decided it was a little too much effort. I could be persuaded otherwise…

AlanGriffiths:
Ah! The joy of format types vs typedef'ed types.
```c++
/spread/mir/src/server/frontend_xwayland/xwayland_wm.cpp:181:78: error: format â€˜%luâ€™ expects argument of type â€˜long unsigned intâ€™, but argument 2 has type â€˜std::map<unsigned int, std::shared_ptr<mir::frontend::XWaylandSurface> >::size_type {aka unsigned int}â€™ [-Werror=format=]
         log_debug("Closing %lu XWayland surface(s)...", local_surfaces.size());
```

AlanGriffiths (approved):
bors r+

PR #1687, opened 2020-08-25
Implement zwlr_layer_shell_v1 version 3

wmww:
Note that the major unstable version of layer shell is still 1, so all the changes are backwards compatible. The important change is the addition of a `.set_layer()` request. Tested in https://github.com/MirServer/wlcs/pull/173. Fixes https://github.com/MirServer/mir/issues/1589.


AlanGriffiths (approved):
It occurs to me that we don't police getting V2 or V3 requests after negotiating V1. But in this case, that seems harmless within the server.

bors r+

wmww:
I *think* libayland itself will complain in that case

PR #1688, opened 2020-08-26
Handle windows that specify multiple window types

### LINKED ISSUE #1250: Some Kate menus are not children on XWayland
When running Kate under XWayland, some menus are correctly placed, but some are not. For example, `View` -> `Split View` are both good, but `Tool Views` just below it is not and the toplevel menu `Bookmarks` is also wrong. The bad menus appear to not be getting a parent, which would imply `WM_TRANSIENT_FOR` is not being set. Further investigation is required.

wmww:
X11 window type [can be a list of types](https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm46035372549728). We were assuming there was only one type. When there was more we were gracefully handling the error, but the type was not getting set. This looks through the list of types and uses the first we know of (they are in order of preference, so this should be correct).

Fixes #1250, based on #1560

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -164,7 +164,7 @@ mf::XWaylandSurface::XWaylandSurface(
           property_handler<std::string const&>(
               connection,
               window,
-              connection->net_wm_name,
+              connection->_NET_WM_NAME,
               [this](auto value)
               {
                   std::lock_guard<std::mutex> lock{mutex};
@@ -185,13 +185,13 @@ mf::XWaylandSurface::XWaylandSurface(
           property_handler<xcb_atom_t>(
               connection,
               window,
-              connection->net_wm_window_type,
+              connection->_NET_WM_WINDOW_TYPE,
               [this](xcb_atom_t wm_type) { window_type(wm_type); },
               []{}),
           property_handler<std::vector<xcb_atom_t> const&>(
               connection,
               window,
-              connection->wm_protocols,
+              connection->WM_PROTOCOLS,
               [this](std::vector<xcb_atom_t> const& value)
               {
                   std::lock_guard<std::mutex> lock{mutex};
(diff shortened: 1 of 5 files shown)


src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -997,37 +997,40 @@ void mf::XWaylandSurface::apply_any_mods_to_scene_surface()
...
+            set_type(mir_window_type_menu);
+            return;
+        }
+        else if (verbose_xwayland_logging_enabled())
         {
             log_debug("Ignoring type (%s) of %s",
-                      connection->query_name(wm_type).c_str(),
-                      connection->window_debug_string(window).c_str());
+                        connection->query_name(wm_type).c_str(),
+                        connection->window_debug_string(window).c_str());
```
AlanGriffiths, line 1029: Doesn't seem to be an improvement

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -182,11 +182,11 @@ mf::XWaylandSurface::XWaylandSurface(
               {
                   is_transient_for(XCB_WINDOW_NONE);
               }),
-          property_handler<xcb_atom_t>(
+          property_handler<std::vector<xcb_atom_t>>(
               connection,
               window,
               connection->_NET_WM_WINDOW_TYPE,
-              [this](xcb_atom_t wm_type) { window_type(wm_type); },
+              [this](std::vector<xcb_atom_t> wm_types) { window_type(wm_types); },
```
AlanGriffiths, line 189: If we're going to pass a vector by value, then let's `move` it rather than copying.

AlanGriffiths (approved):
A couple of minor corrections needed, but looks good.

src/server/frontend_xwayland/xwayland_surface.cpp
wmww: No reason to pass around by value at all

src/server/frontend_xwayland/xwayland_surface.cpp
wmww: Fixed

RAOF (approved):
LGTM.

bors r+

PR #1691, opened 2020-08-31
gbm-kms: Enable more GL debugging for --debug

RAOF:
This sets up a debug GL context, which drivers will generally use to do more checks and output more verbose messages, and installs a maximally-verbose GL debug logging callback to catch those messages.


AlanGriffiths (approved):
Appears to work.

bors r+

PR #1693, opened 2020-09-01
Log SIGHUP and other signals

AlanGriffiths:
Some deployments are sending Mir a SIGHUP that was causing us to exit untidily.

Explicitly handle SIGHUP and log all signal.

### DIFF
--- a/src/include/server/mir/glib_main_loop_sources.h
+++ b/src/include/server/mir/glib_main_loop_sources.h
@@ -114,10 +114,11 @@ private:
         std::vector<int> sigs;
         std::function<void(int)> handler;
     };
+    struct SignalPid { int sig; pid_t pid; };
 
     void dispatch_pending_signal();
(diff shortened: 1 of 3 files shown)


Saviq (changes requested):
I don't think we should be ignoring `SIGHUP` here. Log where it came from, sure, but still process it and shut down as we would on `SIGTERM`.

AlanGriffiths:
> I don't think we should be ignoring `SIGHUP` here. Log where it came from, sure, but still process it and shut down as we would on `SIGTERM`.

OK, that requires a bit more rework as we don't currently get `si_pid` in the terminate handlers.

Saviq (approved):
```
$ pkill -SIGHUP miral-shell &
[1] 859356
# ...
[2020-09-01 21:07:06.626840] < - debug - > mirserver: Handling Hangup from pid=859356
```

bors merge

Issue #1100, opened 2019-11-27
[wayland-platform] poor graphics performance


AlanGriffiths:
There's a marked difference running `glmark2-es2-wayland --fullscreen` in egmde-confined-desktop between Mir-on-X11 and Mir-on-Wayland.

The latter reports 60fps or worse.

AlanGriffiths:
Just as a data point this happens with both Mir and GNOME as the host environment.

AlanGriffiths:
@RAOF you had a long discussion with @mariogrip recently (after he tried the new wayland-platform). I lost track: is there anything there that applies? Or other ideas what's up?

AlanGriffiths:
I think this may be relevant: https://emersion.fr/blog/2018/wayland-rendering-loop/

> So instead of using this default blocking behaviour, maybe you’ll want not to block. This can be done by manually managing frame callbacks and calling eglSwapInterval(egl_display, 0).

PR #1694, opened 2020-09-02
[Wayland-platform] Compositing shouldn't block in eglSwapBuffers()

AlanGriffiths:
[Wayland-platform] Compositing shouldn't block in eglSwapBuffers(). (Fixes: #1100)

### DIFF
--- a/src/platforms/wayland/displayclient.cpp
+++ b/src/platforms/wayland/displayclient.cpp
@@ -28,10 +28,11 @@
 
 #include <boost/throw_exception.hpp>
 
+#include <algorithm>
+#include <condition_variable>
 #include <cstring>
 #include <stdlib.h>
 #include <system_error>
-#include <algorithm>
 
 namespace mgw = mir::graphics::wayland;
 
@@ -93,6 +94,12 @@ public:
     EGLSurface eglsurface{EGL_NO_SURFACE};
 
     std::function<void(Output const&)> on_done;
+
+    std::mutex frame_mutex;
+    bool frame_posted = false;
+    std::condition_variable frame_cv;
+
+    void frame_done(struct wl_callback* callback, uint32_t time);
 
     // DisplaySyncGroup implementation
     void for_each_display_buffer(std::function<void(DisplayBuffer&)> const& /*f*/) override;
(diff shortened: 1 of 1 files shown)


mariogrip (approved):
LGTM :+1: 

This works really well on pinephone and pinetab :)

AlanGriffiths:
Don't land this until we've collected some performance benchmarks for the hosted configuration.

mariogrip:
Tested with glmark2, i have a over 100% improvements

from 200FPS (without this patch) to almost native (without nested) 600FPS

Same story on arm devices.

AlanGriffiths:
> I _still_ don't know why this makes a difference. It looks like it _should_ just be moving from blocking in `swap_buffers` to blocking in `post`, and when I trace the callstack the `post` call gets made immediately after the call to `swap_buffers` anyway.

I was thinking that `scene_elements_for()` locked the client BufferStreams and that these are released before the `post()` call. I'll look more closely.

> Something else is happening. Notably, the X11 platform doesn't suffer from the “locked at 60FPS” problem and _it_ blocks in `eglSwapBuffers` in `swap_buffers`.

I think that is down to the X11 server accepting the buffer far more quickly than does a Mir Wayland server (which waits for the host to consume the buffer). I did try a change to that effect that (rightly) got rejected when I first discovered this issue.

RAOF:
FWIW I've been testing against GNOME Shell, not Mir, as the host compositor.

eglSwapBuffers *should* block until vblank in both the Wayland and X11 cases.

AlanGriffiths:
I'll get back to this discussion after I've dealt with #1700 and #1696

AlanGriffiths:
> I _still_ don't know why this makes a difference. It looks like it _should_ just be moving from blocking in `swap_buffers` to blocking in `post`, and when I trace the callstack the `post` call gets made immediately after the call to `swap_buffers` anyway.
> 
> Something else is happening. Notably, the X11 platform doesn't suffer from the “locked at 60FPS” problem and _it_ blocks in `eglSwapBuffers` in `swap_buffers`.

My earlier understanding of why this works was definitely wrong.

I moved the `cv.wait()` from `post()` to `swap_buffers()` and still see the performance fix.

AlanGriffiths:
@RAOF I don't understand why this works. The original code has `eglSwapBuffers()` implicitly synchronized to the frame, the new code explicitly synchronizes `swap_buffers()` to the frame.

The difference between theory and practice. Again.

RAOF:
I'm ok with landing this, but before landing this I think we should manually check (maybe with wayland protocol debugging?) that this isn't doing something silly, like rendering unthethered to vsync.

:woman_shrugging:

AlanGriffiths:
> I'm ok with landing this, but before landing this I think we should manually check (maybe with wayland protocol debugging?) that this isn't doing something silly, like rendering unthethered to vsync.

I *think* we're OK: I don't see anything that concerns me.

But we're basically dependent on the host compositor to handle vsync, so what does a successful check of this look like to you?

RAOF:
Ok. I think we can land this and then see if anything weird shakes out.

bors r+

PR #1697, opened 2020-09-04
Activate window on unhide

wmww:
I'm not quite sure why I originally wrote this, but I just found it lying around in an old branch and it seemed like good code. The idea is a window shouldn't be able to steal focus from another app by unhiding an existing surface, but it should be able to steal focus from another of its own windows.


wmww:
What the hell? That exact test (`a_second_window_of_different_app_leaves_first_active_when_unhiding`) passes on my machine.

RAOF (approved):
This seems like a reasonable behaviour, one the tests pass ;)

wmww:
The tests that were broken seem to have been dropped from master, so we might be good to go.

CI seems to have ignored my push, hmm
bors try

src/miral/basic_window_manager.cpp
```diff
@@ -1503,11 +1503,13 @@ void miral::BasicWindowManager::set_state(miral::WindowInfo& window_info, MirWin
         break;
 
     default:
-        auto const none_active = !active_window();
+        bool const can_become_active =
+            !active_window() ||
+            active_window().application() == window_info.window().application();
```
AlanGriffiths, line 1508: Nit: `window_info.window()` is already cached as `window`

AlanGriffiths (approved):
bors delegate=wmww

src/miral/basic_window_manager.cpp
wmww: So it is

PR #1698, opened 2020-09-05
[Explicit destruction version] Give each XWayland app it's own session

### LINKED ISSUE #479: All X11 windows appear as part of the same application
This affects, for example window switching (Alt-Tab/Alt-`) and (especially for the tiling window manager) layout.

wmww:
Fixes #479. Allows for better switching between different windows on the same apps, and I believe may help Unity8 (ping @mariogrip).

### DIFF
--- a/src/server/frontend_xwayland/xwayland_client_manager.cpp
+++ b/src/server/frontend_xwayland/xwayland_client_manager.cpp
@@ -41,7 +41,7 @@ mf::XWaylandClientManager::~XWaylandClientManager()
     {
         log_warning(
             "XWaylandClientManager destroyed without all sessions being released "
-            "(sessions_by_pid: %lu, sessions_by_owner: %lu)",
+            "(sessions_by_pid: %zu, sessions_by_owner: %zu)",
             sessions_by_pid.size(),
             sessions_by_owner.size());
 
--- a/src/server/frontend_xwayland/xwayland_client_manager.cpp
+++ b/src/server/frontend_xwayland/xwayland_client_manager.cpp
@@ -60,7 +60,7 @@ mf::XWaylandClientManager::~XWaylandClientManager()
     }
 }
 
-auto mf::XWaylandClientManager::get_session_for_client(
+auto mf::XWaylandClientManager::register_owner_for_client(
     void* owner_key, pid_t client_pid) -> std::shared_ptr<ms::Session>
 {
     std::lock_guard<std::mutex> lock{mutex};
--- a/src/server/frontend_xwayland/xwayland_client_manager.cpp
+++ b/src/server/frontend_xwayland/xwayland_client_manager.cpp
@@ -17,6 +17,7 @@
  */
 
 #include "xwayland_client_manager.h"
+#include "xwayland_log.h"
 #include "null_event_sink.h"
 #include "mir/shell/shell.h"
 #include "mir/scene/session.h"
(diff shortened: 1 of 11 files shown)


src/server/frontend_xwayland/xwayland_client_manager.cpp
```diff
@@ -0,0 +1,148 @@
...
+mf::XWaylandClientManager::~XWaylandClientManager()
+{
+    // release() should have been called for every owner before destroying
+    if (!sessions_by_pid.empty() || !sessions_by_owner.empty())
+    {
+        log_warning(
+            "XWaylandClientManager destroyed without all sessions being released "
+            "(sessions_by_pid: %lu, sessions_by_owner: %lu)",
+            sessions_by_pid.size(),
+            sessions_by_owner.size());
```
AlanGriffiths, line 46: As the compiler mentions, sizes should be formatted with %z.

wmww: Fixed

src/server/frontend_xwayland/xwayland_client_manager.h
```diff
@@ -0,0 +1,72 @@
...
+
+class XWaylandClientManager
+{
+public:
+    XWaylandClientManager(std::shared_ptr<shell::Shell> const& shell);
+    ~XWaylandClientManager();
+
+    /// Returns the session associated with the given client_pid, or creates a new one if needed
+    /// Stores a strong reference to the session associated with the given owner_key
+    auto get_session_for_client(void* owner_key, pid_t client_pid) -> std::shared_ptr<scene::Session>;
```
AlanGriffiths, line 47: 1. The comment appears to describe how the function is implemented but does not agree with the function name on what the function is used for.
2. I don't find `owner_key` a particularly helpful name. The only production-code use is with an `XWaylandSurface*`, so `client_surface` would seem to be a clearer name. And is there any advantage to `void*` as opposed to something more specific?
I think the function does two things: it tracks the surfaces associated with a client and returns a session. That would be better named:
```c++
auto register_surface_for_client(pid_t client_pid, XWaylandSurface* client_surface) -> std::shared_ptr<scene::Session>;
```
Or as two functions:
```c++
void register_surface_for_client(pid_t client_pid, XWaylandSurface* client_surface);
auto session_for_client(pid_t client_pid) -> std::shared_ptr<scene::Session>;
```

AlanGriffiths (changes requested):
This looks like a good start, but I think we could do much better if we changed the `XWaylandClientManager` abstraction and used `shared_ptr<>` to track the windows associated with a pid.

Suppose `XWaylandClientManager::register_surface_for_client()` returns a `shared_ptr<XWaylandClientSession>` and the `XWaylandSurface` keeps that.

`XWaylandClientSession` provides access to the `scene::Session`.

`XWaylandClientManager` keeps a map of `pid` to `weak_ptr`.

`~XWaylandClientSession` and/or a custom deleter can clean up the `scene::Session` and clean out the  `XWaylandClientManager` entry.

WDYT?

src/server/frontend_xwayland/xwayland_client_manager.h
wmww: I use a `void*` simply because I can. The `XWaylandClientManager` doesn't do anything with the surface or care that it's a surface, so why should it get a strongly typed pointer?

wmww:
I opted against the `shared_ptr` version originally because of the disadvantages I listed on the PR that's up now, but I do think it's a close call. Up to you.

src/server/frontend_xwayland/xwayland_client_manager.h
wmww: The two function version doesn't make sense, since the caller doesn't know or care if it's the first surface with that PID. Determining that is one of the main jobs of this abstraction.

I would be willing to change the names, but I don't see why we should use a strongly typed pointer when we're not using that pointer as anything but an ID. Weakest coupling possible is generally a best practice.

AlanGriffiths: > The two function version doesn't make sense, since the caller doesn't know or care if it's the first surface with that PID. Determining that is one of the main jobs of this abstraction.

The two function version doesn't require the caller to know or care if it's the first surface with that PID, it just separates out the "tell" from the "ask". (https://martinfowler.com/bliki/TellDontAsk.html)

> I don't see why we should use a strongly typed pointer when we're not using that pointer as anything but an ID. Weakest coupling possible is generally a best practice.

Stop thinking about how it is implemented, and think about how it is expected to be used. "Owner" is not part of the domain: anyone reading code using this will have to stop and figure out what an "owner" is. There is a meaningful name already in use, why not use it?

AlanGriffiths:
#1709 has been approved

Issue #830, opened 2018-11-27
Short “Ubuntu”-colored flash when mir-kiosk snap starts


tobias-grasse:
First of all: Thank you for your work on this software! I use the mir-kiosk snap with an Ubuntu Core system on a Raspberry Pi 3. 

When the mir-kiosk snap starts up, I see a short colored flash tinted in “Ubuntu” orange.
1) Can I disable this in a way that does not involve forking the snap?
2) If that is not possible, can someone guide me where to start with a MR to make this configurable?

tobias-grasse:
Now I'm the one to reply late 😊 Thanks for clearing this up and outlining the required work. It's currently not a high priority, but once I find the time to get a bit comfortable with C++ I'll give it a try.

knitzsche:
Hi: We have a customer project that is making a device that normal folks in their houses turn on, and the orange flash is not very clean in their opinion. Can you add a way to disable the orange flash via snap config? Thanks.

knitzsche:
I submitted a PR for this: https://github.com/MirServer/mir/pull/1699

AlanGriffiths:
Closed by #1699

PR #1699, opened 2020-09-06
Add --show-splash command line option to control whether splash displays

knitzsche:
Default is to display splash. --show-splash=false suppresses it. (Fixes ~ #830)

### DIFF
--- a/examples/example-server-lib/sw_splash.cpp
+++ b/examples/example-server-lib/sw_splash.cpp
@@ -142,7 +142,7 @@ struct SwSplash::Self : SplashSession
 
     Globals globals;
 
-    bool splash;
+    bool show_splash=true;
 
     DrawContext ctx;
 
@@ -162,8 +162,8 @@ SwSplash::SwSplash() : self{std::make_shared<Self>()} {}
 
 SwSplash::~SwSplash() = default;
 
-void SwSplash::operator()(bool splash_opt){
-    self->splash = splash_opt;
+void SwSplash::enable (bool show_splash_opt){
+    self->show_splash = show_splash_opt;
 }
 
 void SwSplash::operator()(std::weak_ptr<mir::scene::Session> const& session)
@@ -179,7 +179,7 @@ SwSplash::operator std::shared_ptr<SplashSession>() const
 
 void SwSplash::Self::operator()(struct wl_display* display)
 {
-    if (!splash)
+    if (!show_splash)
         return;
     globals.init(display);
 
--- a/examples/example-server-lib/sw_splash.cpp
+++ b/examples/example-server-lib/sw_splash.cpp
@@ -18,7 +18,6 @@
 
 #include "sw_splash.h"
 #include "wayland_helpers.h"
-
 #include <wayland-client.h>
 
 #include <sys/types.h>
@@ -143,6 +142,8 @@ struct SwSplash::Self : SplashSession
 
     Globals globals;
 
+    bool splash;
+
     DrawContext ctx;
 
     std::mutex mutable mutex;
@@ -161,6 +162,10 @@ SwSplash::SwSplash() : self{std::make_shared<Self>()} {}
 
 SwSplash::~SwSplash() = default;
 
+void SwSplash::operator()(bool splash_opt){
+    self->splash = splash_opt;
+}
+
 void SwSplash::operator()(std::weak_ptr<mir::scene::Session> const& session)
 {
     std::lock_guard<decltype(self->mutex)> lock{self->mutex};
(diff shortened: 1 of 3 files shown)


examples/example-server-lib/sw_splash.cpp
```diff
@@ -143,6 +142,8 @@ struct SwSplash::Self : SplashSession
 
     Globals globals;
 
+    bool splash;
+
```
AlanGriffiths, line 146: Probably better to avoid default initialize for a bool. Vis:
```C++
    bool splash = true;
```
And maybe better to call it `show_splash`.

examples/example-server-lib/sw_splash.cpp
```diff
@@ -174,6 +179,8 @@ SwSplash::operator std::shared_ptr<SplashSession>() const
 
 void SwSplash::Self::operator()(struct wl_display* display)
 {
+    if (!splash)
+        return;
```
AlanGriffiths, line 183: This would be clearer as:
```C++
    if (!show_splash)
        return;
```

examples/example-server-lib/sw_splash.h
```diff
@@ -28,6 +28,7 @@ class SwSplash
     SwSplash();
     ~SwSplash();
 
+    void operator()(bool splash_opt);
```
AlanGriffiths, line 31: I know other functions on this are done as `operator()`s, but this is probably better as a normal member function. E.g. `void enable(bool show_splash)`

examples/miral-kiosk/kiosk_main.cpp
```diff
@@ -108,6 +109,12 @@ int main(int argc, char const* argv[])
     MirRunner runner{argc, argv};
 
     DisplayConfiguration display_config{runner};
+ 
+    CommandLineOption show_splash{
+        [&](bool show_splash) {splash(show_splash); },
+        "splash",
+        "Control display splash on startup",
```
AlanGriffiths, line 116: The option name (and explanation) leaves a little to be desired, I would prefer "show-splash".

examples/miral-kiosk/kiosk_main.cpp
```diff
@@ -79,6 +79,7 @@ struct KioskAuthorizer : miral::ApplicationAuthorizer
 };
 
 std::atomic<bool> KioskAuthorizer::startup_only{false};
+
```
AlanGriffiths, line 82: Unrelated?

AlanGriffiths (changes requested):
Thanks for this, it does what it says. However,  few renames would make it clearer.

RAOF (commented):
It's a *little* weird that this disables the splash client by making the client start up and then not do anything rather than just not starting up the splash client, but I can see how this is an easier change to make.

AlanGriffiths:
> It's a _little_ weird that this disables the splash client by making the client start up and then not do anything rather than just not starting up the splash client, but I can see how this is an easier change to make.

That wouldn't be too hard. It just needs something similar to egmde's `run_apps` lambda (but using `InternalClientLauncher`). See: https://github.com/AlanGriffiths/egmde/blob/master/egmde.cpp#L52

examples/example-server-lib/sw_splash.cpp
knitzsche: fixed in pr update

examples/example-server-lib/sw_splash.cpp
knitzsche: fixed in pr update

examples/example-server-lib/sw_splash.h
knitzsche: fixed in pr update

examples/miral-kiosk/kiosk_main.cpp
knitzsche: fixed in pr update

examples/miral-kiosk/kiosk_main.cpp
knitzsche: removed accidental extra line in pr update

knitzsche:
Alan, I made the changes you requested and updated the PR branch, but the PR still shows "changes requested". Do I need to do something to update this?

AlanGriffiths:
> Do I need to do something to update this?

No, I just have to get around to re-reviewing. (It's on my list.)

AlanGriffiths (approved):
LGTM. I'll leave @RAOF have a final say.

knitzsche:
Hi folks. Just checking in since this has an impact on a project. Cheers.

wmww (approved):
LGTM
bors r+

PR #1711, opened 2020-09-11
Pervasive libepoxy

RAOF:
Here's a look at what using libepoxy pervasively would be; this passes all the tests (although needs rebasing to revert the previous revert before it merges cleanly).

Having done it, I don't *think* that it's worth it (hence the no-merge and RFC tags). I'd like a different way of handling E{,GL} extensions, but I think particularly the extra, hidden, setup required for tests with mock EGL and GLES will be an ongoing problem if we choose this.


AlanGriffiths:
We don't need this. (Yet?)

PR #1712, opened 2020-09-11
CMake: Only gcc understands -ffat-lto-objects

RAOF:
Fixes the build with clang and `MIR_LINK_TIME_OPTIMIZATION=ON`.


AlanGriffiths (approved):
Sure

Issue #1713, opened 2020-09-14
eglstream-kms platform only built for amd64


RAOF:
We only build the eglstream-kms platform for amd64, but there exist at least ARM boards with NVIDIA graphics. We should probably just build it everywhere?

Saviq:
FWIW it's unclear to me that the ARM Nvidia hardware (Tegra?) uses the same API. That said, if it ~~blends~~builds, let's.

RAOF:
This is the result of a discussion in #mir-server by someone who wants to use Mir on a Jenson, and has demo code using EGLStream/EGLOutput working. There might be weird pitfalls lurking, but the base EGL & KMS APIs apparently exist there.

PR #1714, opened 2020-09-14
debian: Enable eglstream-kms platform everywhere

RAOF:
You can stick an NVIDIA card in a Power9 desktop, and NVIDIA will sell you
ARM development boards with NVIDIA GPUs. The dependencies exist everywhere,
so we might as well build the `eglstream-kms` platform everywhere, too.

Closes: #1713


Saviq (changes requested):
```
Broken mir-cross-build-deps:arm64 Depends on libnvidia-egl-wayland-dev:arm64 < none @un H >
Removing mir-cross-build-deps:arm64 because I can't find libnvidia-egl-wayland-dev:arm64
```

https://packages.ubuntu.com/search?keywords=libnvidia-egl-wayland-dev

Only there since focal…

RAOF:
Oh, yeah. Bionic is older than focal!

RAOF:
Wait. How does this now fail at resolving `python3`?

RAOF:
How is that bringing in an incompatible python3?!

AlanGriffiths:
> How is that bringing in an incompatible python3?!

It does seem odd, but we had something like this before: https://bugs.launchpad.net/mir/+bug/1494317. That led to the somewhat odd: 
https://github.com/MirServer/mir/blob/727d467ba0a1b97464be3ec8c199109cd7d22f69/debian/control#L43

Saviq:
> > How is that bringing in an incompatible python3?!
> 
> It does seem odd, but we had something like this before: https://bugs.launchpad.net/mir/+bug/1494317. That led to the somewhat odd:
> 
> https://github.com/MirServer/mir/blob/727d467ba0a1b97464be3ec8c199109cd7d22f69/debian/control#L43

Hopefully this could be `:native` by now?

Saviq:
This still:
```
Broken mir-cross-build-deps:arm64 Depends on eglexternalplatform-dev:arm64 < none @un H >
Removing mir-cross-build-deps:arm64 because I can't find eglexternalplatform-dev:arm64
```

I'd bet these don't have multiarch considerations in them…

Saviq:
Down to:

```
unsat-dependency: eglexternalplatform-dev:arm64
```

And that sounds like it's missing the appropriate MultiArch markings?

Conan-Kudo:
Heh, this change brings the Ubuntu mir package in line with Fedora, actually. I've been shipping eglstream-kms support enabled on all arches from the beginning. 😜

AlanGriffiths:
Not sure exactly what's going wrong here. There's a bunch of `ActiveWindow.*` wlcs tests failing on 20.04/arm64. I've tried re-running that build a couple of times but see the same failures.

RAOF:
@AlanGriffiths Yeah, we used not to run the tests on the cross-built arm64. If they're not trivial to fix we should just disable them again.

Saviq:
> @AlanGriffiths Yeah, we used not to run the tests on the cross-built arm64. If they're not trivial to fix we should just disable them again.

How can we run arm64 tests on amd64 in the first place‽

RAOF:
> > @AlanGriffiths Yeah, we used not to run the tests on the cross-built arm64. If they're not trivial to fix we should just disable them again.
> 
> How can we run arm64 tests on amd64 in the first place‽

`qemu-user-static` is installed, so we can run arm64 binaries!

Saviq:
> `qemu-user-static` is installed, so we can run arm64 binaries!

Right, realized that after looking at the diff ;)

Saviq:
@AlanGriffiths now it's this: ¿?
```plain
dpkg-gensymbols: error: some new symbols appeared in the symbols file: see diff output below
dpkg-gensymbols: warning: debian/libmiral4/DEBIAN/symbols doesn't match completely debian/libmiral4.symbols
--- debian/libmiral4.symbols (libmiral4_3.1.0_arm64)
+++ dpkg-gensymbolsgXhEhN	2020-09-29 17:05:08.420975715 +0000
@@ -1,6 +1,12 @@
 libmiral.so.4 libmiral4 #MINVER#
  MIRAL_3.0@MIRAL_3.0 3.0.0
  MIRAL_3.1@MIRAL_3.1 3.1.0
+ _ZN5miral17WaylandExtensions7ContextC1Ev@MIRAL_3.0 3.1.0
+ _ZN5miral17WaylandExtensions7ContextC2Ev@MIRAL_3.0 3.1.0
+ _ZN5miral17WaylandExtensions7ContextD0Ev@MIRAL_3.0 3.1.0
+ _ZN5miral17WaylandExtensions7ContextD1Ev@MIRAL_3.0 3.1.0
+ _ZN5miral17WaylandExtensions7ContextD2Ev@MIRAL_3.0 3.1.0
+ _ZTVN5miral17WaylandExtensions7ContextE@MIRAL_3.0 3.1.0
  (c++)"miral::AddInitCallback::AddInitCallback(std::function<void ()> const&)@MIRAL_3.0" 3.0.0
  (c++)"miral::AddInitCallback::operator()(mir::Server&) const@MIRAL_3.0" 3.0.0
  (c++)"miral::AddInitCallback::~AddInitCallback()@MIRAL_3.0" 3.0.0
```

AlanGriffiths:
> + _ZN5miral17WaylandExtensions7ContextC1Ev@MIRAL_3.0 3.1.0
> + _ZN5miral17WaylandExtensions7ContextC2Ev@MIRAL_3.0 3.1.0
> + _ZN5miral17WaylandExtensions7ContextD0Ev@MIRAL_3.0 3.1.0
> + _ZN5miral17WaylandExtensions7ContextD1Ev@MIRAL_3.0 3.1.0
> + _ZN5miral17WaylandExtensions7ContextD2Ev@MIRAL_3.0 3.1.0
> + _ZTVN5miral17WaylandExtensions7ContextE@MIRAL_3.0 3.1.0

Oh the delights of different code generation on different architectures!

I don't think we need to export the ctor/dtor entry points at all, let me experiment...

AlanGriffiths:
Missed one:

```plain
+ _ZTVN5miral17WaylandExtensions7ContextE@MIRAL_3.0 3.1.0
```

spread.yaml
```diff
@@ -75,3 +75,4 @@ path:
 
 exclude:
     - .git
+    - cmake-build-*
```
Saviq, line 78: Unsure why this…

spread/build/ubuntu-sbuild/task.yaml
```diff
@@ -0,0 +1,34 @@
+environment:
+    DEB_BUILD_OPTIONS: "parallel=$( nproc ) noopt ${NOCHECK}"
```
Saviq, line 2: `parallel` is overridden by `-j` below, no?

spread.yaml
RAOF: Oh, whoops. That's just CLion's default build directories; excluding them makes spread take about a minute less time compressing and transferring the source :wink:

I didn't mean to push that change, but maybe we want it anyway? Other people use CLion, too.

spread/build/ubuntu-sbuild/task.yaml
RAOF: Indeed it is.

Saviq:
There's conflicts here now, FWIW.

PR #1716, opened 2020-09-16
Add an exception for Wayland protocol errors

RAOF:
This adds `mir::wayland::ProtocolError`, an exception that request handlers can throw which will result in a Wayland protocol error being sent to the client.

This is more convenient than bubbling early-returns up the call stack.


AlanGriffiths (approved):
Yes!

bors r+

wmww:
This is good

PR #1720, opened 2020-09-19
Support Layer Shell margin

### LINKED ISSUE #1719: Layer shell implementation does not support margin
We currently ignore the margin, which makes things like notifications look bad.

wmww:
The immediate reason to support margin is so MATE notifications look right. This implementation creates a large `scene::Surface` uses `WindowWlSurfaceRole::set_pending_offset()` to place the `WlSurface` in the correct position. Input dispatching just works (because that's already implemented for offset surfaces), but popups need further adjustment. It might be preferable to do this a different way (such as adding a margin surface property specifically for this use case?) or we might want to merge this and see what design makes sense after the surface/window abstraction split.

Extensive refactoring of the layer shell implementation was required. I also added some protocol errors that we weren't catching before (this will probably make it fail any version of WLCS before https://github.com/MirServer/wlcs/pull/181 lands).

FIxes #1719, tested by https://github.com/MirServer/wlcs/pull/181. I also manually tested with gtk-layer-demo.


wmww:
Merge #1718 first

wmww:
For the record, I don't expect a reviewer to verify the logic is all correct. The only reason I believe it to be mostly correct is because it passes both manual and WLCS testing.

AlanGriffiths (approved):
Sure

bors r+

Issue #1722, opened 2020-07-27
When confined (in mir-test-tools) GLMark2Xwayland.* "succeeds" without displaying anything


AlanGriffiths:
[Mir 2.0] We just get the black `mir_demo_server` window while each of the tests run...
```plain
$ mir-test-tools.performance-test --gtest_filter=GLMark2Xwayland.* --logging on
/snap/mir-test-tools/3467/bin/desktop-launch: line 51: /home/alan/.config/user-dirs.dirs: Permission denied
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.a11y.applications.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.a11y.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.a11y.keyboard.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.a11y.magnifier.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.a11y.mouse.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.app-folders.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.background.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.calendar.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.datetime.gschema.xml': File exists
ln: failed to create symbolic link '/home/alan/snap/mir-test-tools/3467/.local/share/glib-2.0/schemas/org.gnome.desktop.default-applications.gschema.xml': File exists
[...]

PR #1723, opened 2020-09-21
Don't wait for Xwayland to init before starting WM

AlanGriffiths:
Don't wait for Xwayland to init before starting WM. (Fixes: #1722)

There is a race between the X11 client connecting (which triggers launching Xwayland) and the WM connecting. Something about the confined environment makes it likely the client wins the race.

If the client wins the race, our WM never processes the client window and nothing gets displayed.

There's no need to wait on SIGUSR1 before starting the WM, it will block awaiting events anyway. And, empirically, not waiting for SIGUSR1 means the WM wins the race even in the confined environment.

(This PR doesn't actually resolve the race, just fixes the outcome.)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_server.cpp
+++ b/src/server/frontend_xwayland/xwayland_server.cpp
@@ -129,33 +129,26 @@ auto fork_xwayland_process(
     }
 }
 
-bool spin_wait_for(sig_atomic_t& xserver_ready)
-{
-    auto const end = std::chrono::steady_clock::now() + 5s;
-
-    while (std::chrono::steady_clock::now() < end && !xserver_ready)
-    {
-        std::this_thread::sleep_for(100ms);
-    }
-
-    return xserver_ready;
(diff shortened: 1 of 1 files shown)


wmww:
I want to know why the SIGUSR1 stuff was there in the first place. It must have served a purpose (if not in Mir, at least in the Weston code it originated from).

AlanGriffiths:
> I want to know why the SIGUSR1 stuff was there in the first place. It must have served a purpose (if not in Mir, at least in the Weston code it originated from).

I can see a potential point in informing an X-WM that has already been set up that the X-server has come up. But our code isn't structured  for that to be useful. Either way there's no point in not starting to set up the X-WM until the X-server signals "init".

wmww (approved):
Sync has convinced me that, without further evidence, the SIGUSR1 stuff can be assumed to be non-useful in Mir's multi threaded architecture.

bors r+

PR #1724, opened 2020-09-21
Minor XWayland cleanup

wmww:
Including not creating a spawner if it already exists and constructing `XWaylandSurface` with actual properties instead of an event pointer. Nothing super important, but cleans the boring commits out of my xwayland-master branch.


AlanGriffiths (approved):
bors r+

PR #1725, opened 2020-09-22
Manage X11 windows that existed before window manager started

wmww:
Related to #1722 and #1723. It's a lot of code because XCB is messy, but all it's doing is going through the list of windows, getting their properties and setting up a `XWaylandSurface` for them the same way we do on newly created windows. This alone is not enough to fix the race condition because we don't get the `WL_SURFACE_ID` client message. See next PR for fixing that.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_wm.cpp
+++ b/src/server/frontend_xwayland/xwayland_wm.cpp
@@ -170,13 +170,14 @@ mf::XWaylandWM::XWaylandWM(std::shared_ptr<WaylandConnector> wayland_connector, 
 
     // Detect and manage any windows that already exist
     auto const query_tree_cookie = xcb_query_tree(*connection, connection->root_window());
-    auto const query_tree_reply = xcb_query_tree_reply(*connection, query_tree_cookie, nullptr);
+    std::unique_ptr<xcb_query_tree_reply_t> query_tree_reply{
+        xcb_query_tree_reply(*connection, query_tree_cookie, nullptr)};
     if (query_tree_reply)
     {
         std::vector<std::function<void()>> window_setup_funcs;
-        for (int i = 0; i < xcb_query_tree_children_length(query_tree_reply); ++i)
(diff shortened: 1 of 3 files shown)


src/server/frontend_xwayland/xwayland_wm.cpp
```diff
@@ -163,9 +162,67 @@ mf::XWaylandWM::XWaylandWM(std::shared_ptr<WaylandConnector> wayland_connector,
...
 
+    connection->flush();
+
     cursors->apply_default_to(connection->root_window());
 
     wm_shell->surface_stack->add_observer(scene_observer);
+
+    // Detect and manage any windows that already exist
+    auto const query_tree_cookie = xcb_query_tree(*connection, connection->root_window());
+    auto const query_tree_reply = xcb_query_tree_reply(*connection, query_tree_cookie, nullptr);
```
AlanGriffiths, line 173: Can we use unique_ptr<> to ensure the free? We're calling out to C++ code that doesn't obviously guarantee noexecept.

AlanGriffiths (changes requested):
Thanks for looking into this, it was next on my list for cleaning up this startup race.

BTW You can mark it "Fixes: #1722" - I've tested it.

src/server/frontend_xwayland/xwayland_wm.cpp
wmww: Doing it manually is consistent with the rest of the code. I agree some improvements are needed. Can we land this first and then I'll make another refactoring PR?

wmww: Actually, big fancy solution seems to be more trouble than it's worth (at least at this time). Switched to using `unique_ptr` as you suggested.

Saviq (approved):
I just looked at @AlanGriffiths's latest commit, LGTM.

AlanGriffiths (approved):
> I just looked at @AlanGriffiths's latest commit, LGTM.

Rest is OK by me

bors r+

PR #1726, opened 2020-09-22
Create event dispatcher before running server

wmww:
Related to #1722 and #1723. In combination with #1725, I think this fixes the race condition (though I haven't tested in a confined snap yet). This moves socket pair creation out of the `XWaylandServer` constructor, and creates the event dispatcher before creating the window manager or spawning the server. Because `XWaylandConnector`'s mutex is locked, any events will be blocked on the window manager being created and wont be dropped.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_connector.cpp
+++ b/src/server/frontend_xwayland/xwayland_connector.cpp
@@ -133,15 +133,19 @@ void mf::XWaylandConnector::spawn()
 
     try
     {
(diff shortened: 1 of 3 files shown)


wmww:
I experimented with launching the window manager before the server, but it looks like creating an XCB connection requires a round-trip so that deadlocks.

AlanGriffiths (approved):
LGTM

bors r+

PR #1727, opened 2020-09-22
Better synchronization when starting X-server and X-WM

AlanGriffiths:


AlanGriffiths:
An alternative to ripping this code out. (C.f #1723)

AlanGriffiths:
Closing: We're landing the alternative

PR #1728, opened 2020-09-22
Refactor XCBConnection::read_property()

wmww:
Use `unique_ptr` for memory management where possible and simplify error handling into one try/catch block

### DIFF
--- a/src/server/frontend_xwayland/xcb_connection.cpp
+++ b/src/server/frontend_xwayland/xcb_connection.cpp
@@ -227,40 +227,28 @@ auto mf::XCBConnection::read_property(
         0, // no offset
         2048); // big buffer
 
-    return [xcb_connection = xcb_connection, cookie, action, on_error, prop, this]()
+    return [this, cookie, action, on_error, window, prop]()
         {
-            xcb_get_property_reply_t *reply = xcb_get_property_reply(xcb_connection, cookie, nullptr);
-            if (reply && reply->type != XCB_ATOM_NONE)
+            try
             {
-                try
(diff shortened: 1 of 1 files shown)


src/server/frontend_xwayland/xcb_connection.cpp
```diff
@@ -119,11 +119,10 @@ mf::XCBConnection::Atom::operator xcb_atom_t() const
         // The initial atomic check is faster, but we need to check again once we've got the lock
         if (atom == XCB_ATOM_NONE)
         {
-            auto const reply = xcb_intern_atom_reply(*connection, cookie, nullptr);
+            std::unique_ptr<xcb_intern_atom_reply_t> const reply{xcb_intern_atom_reply(*connection, cookie, nullptr)};
```
RAOF, line 122: Is this actually going to do the right thing? This is going to call `delete reply;`, rather than `free(reply);`, right?

You probably need `std::unique_ptr<xcb_intern_atom_reply_t, &free> const reply`?

RAOF (dismissed):
I think this is going to result in memory corruption from pairing `malloc()` with `delete`? Otherwise seems sane.

src/server/frontend_xwayland/xcb_connection.cpp
AlanGriffiths: Something like that.

See https://github.com/MirServer/mir/pull/1725/commits/132134e77501ec894890393e8efe2096fd5ae5a0 for fir the solution I propose on another branch.

AlanGriffiths (approved):
LGTM

PR #1730, opened 2020-09-23
Safer wayland::Foo::from(wl_resource*)

RAOF:
A `static_cast` from `void*` obviously has no safety rails. But we can do better!
`wl_resource_instance_of` will tell us whether `resource` has the same implementation
as our wrapper class - basically a discount `dynamic_cast`.

We can only use it where the `wl_resource` *has* an implementation (so objects without
requests can't use this safety rail), but that covers most objects.

This is needed for the linux-dmabuf branch, as that deals with `wl_resource* buffer` which
*may* be instances of `mw::Buffer`, or may not, and the `static_cast` results in a subsequent
invalid memory access.

### DIFF
--- a/src/wayland/generated/wayland_wrapper.cpp
+++ b/src/wayland/generated/wayland_wrapper.cpp
@@ -56,11 +56,6 @@ struct wl_interface const* all_null_types [] {
 
 // Callback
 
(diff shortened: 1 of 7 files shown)


AlanGriffiths (approved):
Nice!

bors r+

Issue #1663, opened 2020-08-07
Rendering is incorrect on RPi 4


RAOF:
It looks like ![this](https://user-images.githubusercontent.com/804339/89596118-ba92cc00-d899-11ea-948c-8e02aae3be0a.jpeg)

Probably a stride issue.

Update: this is a Mesa issue. Munging Weston to disable the `linux_dmabuf` support, glmark2 under Weston displays the sam eproblem.

Aury88:
This issue also affect lomiri [[1](https://gitlab.com/ubports/community-ports/pinephone/-/issues/12)][[2](https://gitlab.com/ubports/core/rootfs-builder-debos/-/issues/2)] (aka unity8) in UT on RPi4.
 let us know if there is any useful information you need to help solve the problem.

PR #1731, opened 2020-09-23
gbm-kms: Support zwp_linux_dmabuf_unstable_v1

RAOF:
This is the new, improved way to do client buffer submission. Notably,
it features DRM modifier support, allowing clients to allocate fancy
driver-specific high-performance buffer formats.

Closes: #1663

### DIFF
--- a/src/platforms/gbm-kms/server/linux_dmabuf.cpp
+++ b/src/platforms/gbm-kms/server/linux_dmabuf.cpp
@@ -1,0 +1,846 @@
+/*
+ * Copyright © 2020 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Christopher James Halse Rogers <christopher.halse.rogers@canonical.com>
+ */
+
+
+#include "linux_dmabuf.h"
+
+#include "wayland_wrapper.h"
+#include "mir/graphics/egl_extensions.h"
+#include "mir/graphics/egl_error.h"
+#include "mir/renderer/gl/context.h"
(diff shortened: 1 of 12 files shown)


src/platforms/gbm-kms/server/linux_dmabuf.cpp
```diff
@@ -0,0 +1,846 @@
...
+    }
+
+    mir::geometry::Size size() const override
+    {
+        return size_;
+    }
+
+    MirPixelFormat pixel_format() const override
+    {
+        // There's no way to implement this corectly…
```
AlanGriffiths, line 578: Is there a discussion to be had?

src/platforms/gbm-kms/server/linux_dmabuf.cpp
```diff
@@ -0,0 +1,846 @@
...
+    void destroy() override
+    {
+        destroy_wayland_object();
+    }
+
+    EGLDisplay const dpy;
+    std::shared_ptr<mg::EGLExtensions> const egl_extensions;
+    int32_t const width, height;
+    uint32_t const format_;
+    [[maybe_unused]]
```
AlanGriffiths, line 196: That's not going to compile on g++ 9.3.0 and earlier.

AlanGriffiths:
18.04 still not happy! :disappointed:

src/platforms/gbm-kms/server/linux_dmabuf.cpp
RAOF: Yes. We should lift something like [weston's pixel format descriptors](https://gitlab.freedesktop.org/wayland/weston/-/blob/master/libweston/pixel-formats.h) and replace the MirPixelFormat enum.

That's a change touching lots of code, though.

RAOF:
Oh, arse. 18.04 has a too-old cmake to handle library dependencies on object libraries.

RAOF:
Aaargh. And I can't get around it by sneakily installing the cmake snap in the PPA :cry:

AlanGriffiths (approved):
Looks sensible, and doesn't break stuff. (I'd love to be able to verify that it actually works, but don't have a way to do that.)

bors r+

RAOF:
While you can't check if if fixes the Pi, you *can* check that it works on hardware; mesa will preferentially use the newly exported interface on Intel.

24 Sep 2020 19:03:51 Alan Griffiths <notifications@github.com>:

AlanGriffiths:
> While you can't check if if fixes the Pi, you *can* check that it works on hardware; mesa will preferentially use the newly exported interface on Intel.

Well, Mir still works on Intel drivers. :)

mariogrip:
Would this be possible to backport to 1.8? This would fix some (if not many) of our issues on arm devices.

PR #1733, opened 2020-09-23
Add XCBConnection::Handler

wmww:
Stacked on #1728. This simplifies passing around `on_success`/`on_error` callbacks. Starting to look similar to a future but I think it's still fine to have an XCB-specific concept here.

### DIFF
--- a/src/server/frontend_xwayland/xcb_connection.cpp
+++ b/src/server/frontend_xwayland/xcb_connection.cpp
@@ -227,40 +227,28 @@ auto mf::XCBConnection::read_property(
         0, // no offset
         2048); // big buffer
 
-    return [xcb_connection = xcb_connection, cookie, action, on_error, prop, this]()
+    return [this, cookie, action, on_error, window, prop]()
         {
-            xcb_get_property_reply_t *reply = xcb_get_property_reply(xcb_connection, cookie, nullptr);
-            if (reply && reply->type != XCB_ATOM_NONE)
+            try
             {
-                try
+                std::unique_ptr<xcb_get_property_reply_t> reply{xcb_get_property_reply(xcb_connection, cookie, nullptr)};
+                if (reply && reply->type != XCB_ATOM_NONE)
                 {
-                    action(reply);
+                    action(reply.get());
                 }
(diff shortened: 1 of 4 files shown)


src/server/frontend_xwayland/xcb_connection.h
```diff
@@ -90,6 +91,46 @@ class XCBConnection
...
+            if (error)
+            {
+                free(error);
+            }
+        }
+
+        operator xcb_generic_error_t**() { return &error; }
+
+        xcb_generic_error_t* error{nullptr};
+    };
```
AlanGriffiths, line 107: Looks as though ErrorPtr has incorrect CopyAssign semantics. (Is this a case for `UniqueCPtr`?)

src/server/frontend_xwayland/xcb_connection.cpp
```diff
@@ -231,18 +230,23 @@ auto mf::XCBConnection::read_property(
         0, // no offset
         2048); // big buffer
 
-    return [this, cookie, action, on_error, window, prop]()
+    return [this, cookie, handler=std::move(handler), window, prop]()
         {
             try
             {
+                ErrorPtr error;
```
AlanGriffiths, line 237: Looks like this is default initialized and not written before it's read. Can that be right?

AlanGriffiths (changes requested):
There's something odd about `ErrorPtr` and the way it is used. Otherwise, this looks good.

src/server/frontend_xwayland/xcb_connection.h
wmww: I don't think you can get mutable access to the underlying pointer of a `unique_ptr`, right? The error has to be initialized as null, and then optionally set by the XCB function.

src/server/frontend_xwayland/xcb_connection.cpp
wmww: Oops, actual usage got dropped in a rebase somewhere along the line.

AlanGriffiths (approved):
OK now

bors r+

PR #1734, opened 2020-09-24
Respect `nohup` handing of SIGHUP

AlanGriffiths:
If Mir is started via `nohup` do not install a SIGHUP handler.


wmww (approved):
Works for me
bors r+

Issue #1736, opened 2020-09-24
mir_demo_server SIGSEGV in mir::wayland::Global::~Global()


Saviq:
Not a whole lot of detail, but:

```gdb
Thread 1 (Thread 0x7fa013846940 (LWP 151218)):
#0  0x00007fa01430af6d in wl_resource_post_event () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libwayland-server.so.0
#1  0x00007fa01430ba62 in wl_global_remove () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libwayland-server.so.0
#2  0x00007fa01430bac5 in wl_global_destroy () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libwayland-server.so.0
#3  0x00007fa0146b4fcb in mir::wayland::Global::~Global() () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libmirwayland.so.1
#4  0x00007fa013670e41 in ?? () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-gbm-kms.so.18
#5  0x00007fa01366db0f in ?? () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-gbm-kms.so.18
#6  0x00007fa0150b1b98 in ?? () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libmirserver.so.54
#7  0x00007fa01505e6ee in ?? () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libmirserver.so.54
#8  0x00007fa0150b1b98 in ?? () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libmirserver.so.54
#9  0x00007fa014fc1246 in mir::shell::AbstractShell::~AbstractShell() () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libmirserver.so.54
#10 0x00007fa0150b1b98 in ?? () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libmirserver.so.54
#11 0x00007fa0150409bf in ?? () from target:/snap/mir-test-tools/3893/usr/lib/x86_64-linux-gnu/libmirserver.so.54
[...]

Issue #1689, opened 2020-08-26
Wayland globals removed on wrong thread


wmww:
On server shutdown, it appears we're removing globals from the object's destructors, which are being called as part of shared pointer destruction. This is happening in a thread other than the Wayland thread. Noticed this when running Mir in wayland-debug for an unrelated reason.

RAOF:
Not actually closed by 91868fd

wmww:
We do tear down globals on a different thread than the main Wayland one, but we do that after the Wayland thread has been joined. Thus I assume it is safe.

PR #1738, opened 2020-09-25
gbm-kms: Destroy linux-dmabuf global from Wayland mainloop.

RAOF:
Closes: #1736


Saviq:
LGTM, confirmed the crash from #1736 is gone.

bors merge

RAOF:
This definitely doesn't close #1689, as this only fixes the linux-dmabuf global deletion, and that didn't exist when #1689 was filed.

Saviq:
> This definitely doesn't close #1689, as this only fixes the linux-dmabuf global deletion, and that didn't exist when #1689 was filed.

Oups. Sorry.

PR #1748, opened 2020-10-01
Better diagnostic when glGetShaderInfoLog() fails

AlanGriffiths:


src/renderers/gl/renderer.cpp
```diff
@@ -290,9 +290,8 @@ class mrg::Renderer::ProgramFactory : public mir::graphics::gl::ProgramFactory
         glGetShaderiv(id, GL_COMPILE_STATUS, &ok);
         if (!ok)
         {
-            GLchar log[1024];
-            glGetShaderInfoLog(id, sizeof log - 1, NULL, log);
-            log[sizeof log - 1] = '\0';
+            GLchar log[1024] = "(No log info)";
+            glGetShaderInfoLog(id, sizeof log, NULL, log);
```
Saviq, line 294: Don't think there's value in getting the returned length and informing that the operation failed?

AlanGriffiths: I assume the returned length will be 0, and the error from `glGetShaderInfoLog()` will be `GL_INVALID_VALUE` (or maybe `GL_INVALID_OPERATION`), none of which is helpful to log.

Note also that the error flag or flags have likely been set earlier, so we'd also have to loop on `glGetError()` to clear them  before calling `glGetShaderInfoLog()`.

Possible, but seems like over-engineering when all we want is to say "we didn't get a log message".

Saviq (approved):
Yup.

bors merge

PR #1749, opened 2020-10-02
advise_output_delete(): better utilize STL

wmww:
This is slightly more efficient and removes duplicated logic. Let me know if there's a better way to do this (my STL knowledge is hit-and-miss).

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -2517,21 +2517,21 @@ void miral::BasicWindowManager::advise_output_delete(miral::Output const& output
 {
     Locker lock{this};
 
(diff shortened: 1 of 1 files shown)


AlanGriffiths (approved):
I'm not entirely convinced that added cost of stable partitioning (O(N log N)) is outweighed by being able to move instead of copy to `removed_areas` (the original code is O(N)).

But the logic is clearer and the likely size of the data set makes worse asymptotic behaviour a small concern.

bors delegate

wmww:
Cleaning up the duplicated `area->output && area->output.value().is_same_output(output)` was the primary motivator.

bors r+

PR #1755, opened 2020-10-04
Fixes for building with LTO-by-default in Fedora

Conan-Kudo:
This fixes the build for Mir in Fedora with our [LTO-by-default setup for package builds](https://fedoraproject.org/wiki/LTOByDefault).


AlanGriffiths (approved):
This is ugly and some of it is ancient, so let's see what happens with our current toolchains & platforms...

bors r+

Issue #1752, opened 2020-10-02
Logical output groups/display wall


wmww:
Users should be able to configure multiple physical outputs to behave as a single logical display.

PR #1751, opened 2020-10-02
Logical output groups

wmww:
This is the first step towards display walls. Window manager support and tests will come next.

PR #1756, opened 2020-10-05
BasicWindowManager: test and implement logical display groups

wmww:
A little refactoring was required.

`update_application_zones_and_attached_windows()` (formerly `update_windows_for_outputs()`) is now only called in `advise_output_end()` it is also responsible for updating the policy about new and removed application zones. These changes consolidate application zone updates. Without them, the policy might get notified first that a new application zones was created the size of output A and then again that it's expended to encompass the extents of output B. Now, application zone updates are only made once after all related output updates.


wmww:
Related to #1751

wmww:
Related to #1752 (last time didn't take for some reason)

AlanGriffiths (changes requested):
Looks sensible and seems (mostly) to work. Only blocking on the libmiral4.symbols update.

There are issues with the overall feature though - for example `waybar` creates a bar for each physical output and these get stacked at the top of the display area.

image20201008_120819581

wmww:
Looks like the bar issue can be fixed with Waybar output configuration: https://github.com/Alexays/Waybar/wiki/Configuration. I'm not sure what to do with the script. I tried updating the symbols in debian/, did I do it right?

AlanGriffiths (approved):
bors r+

PR #1760, opened 2020-10-07
Pick up min/max size

wmww:
It appears that this isn't supported in the window manager at all and that's a bigger job than I want to do right now, but at least the correct information is getting fed in. I moved the value handler to a template function so property handlers can be easily implemented for more int types.


AlanGriffiths:
> It appears that this isn't supported in the window manager at all

I'm not sure how you reach this conclusion. There has been support and I don't recall it being removed. See, for example, `ClientLibrary.min_size_respected_when_placing_surface`

wmww:
It might be supported initially, but it is not respected when interactively resizing windows (the most obvious way to test). I came to this conclusion by sticking a print statement in `BasicSurface::resize()` and watching the window manager happy try to resize windows to a negative size.

AlanGriffiths (approved):
LGTM

bors r+

wmww:
bors r=AlanGriffiths

PR #1761, opened 2020-10-07
Respect motif decoration hint

### LINKED ISSUE #1758: glxgears -fullscreen has decorations
`glxgears` is an X11 app that Mir provides window borders for, but this shouldn't happen for fullscreen windows.

Run `miral-app` and from a terminal in that window run `glxgears -fullscreen`.

Expect: a fullscreen window without borders
Actual: a fullscreen window **with** borders

_Note, this may be an issue with `glxgears`, other X11 apps do fullscreen correctly._

wmww:
This allows a window to request to be undecorated by setting the decorations element of the `_MOTIF_WM_HINTS` property to 0. This fixes #1758, as well as well as removing the double-decorations seen on gedit.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -72,6 +72,32 @@ enum class NetWmMoveresize: uint32_t
     CANCEL = 11,        /* cancel operation */
 };
 
+// Any standard for the motif hints seems to be lost to time, but Weston has a reasonable definition:
+// https://github.com/wayland-project/weston/blob/f7f8f5f1a87dd697ad6de74a885493bcca920cde/xwayland/window-manager.c#L78
+namespace MotifWmHintsIndices
+{
+enum MotifWmHintsIndices: unsigned
+{
+    FLAGS,
+    FUNCTIONS,
+    DECORATIONS,
+    INPUT_MODE,
+    STATUS,
+    END,
+};
+}
+
+namespace MotifWmHintsFlags
+{
+enum MotifWmHintsFlags: uint32_t
+{
+    FUNCTIONS = (1L << 0),
+    DECORATIONS = (1L << 1),
+    INPUT_MODE = (1L << 2),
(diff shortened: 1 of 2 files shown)


src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -72,6 +72,32 @@ enum class NetWmMoveresize: uint32_t
     CANCEL = 11,        /* cancel operation */
 };
 
+// Any standard for the motif hints seems to be lost to time, but Weston has a reasonable definition:
+// https://github.com/wayland-project/weston/blob/f7f8f5f1a87dd697ad6de74a885493bcca920cde/xwayland/window-manager.c#L78
+namespace MotifWmHintsIndices
+{
+enum MotifWmHintsIndices: unsigned
```
wmww, line 79: Not using enum class because the whole point of this enum is to index into a vector, for which I need explicit int conversion. Static casts would add a lot more code noise and yield no significant safety benefit.

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -72,6 +72,32 @@ enum class NetWmMoveresize: uint32_t
...
+    DECORATIONS,
+    INPUT_MODE,
+    STATUS,
+    END,
+};
+}
+
+namespace MotifWmHintsFlags
+{
+enum MotifWmHintsFlags: uint32_t
```
wmww, line 92: Same rational, except I need to do bitwise operators. I could override the operator, but why?

AlanGriffiths (commented):
LGTM but...

1. #1758 was mostly a note, so I didn't forget to investigate
2. "removing the double-decorations seen on gedit"  - not sure what this means. gedit defaults to using Wayland and, if forced to use X11 by `GDK_BACKEND=x11`, still shows double borders (on 20.4).
3. This does help with a lot of X11 apps that have splash screens and the like.

bors r+

AlanGriffiths (approved):

PR #1770, opened 2020-10-19
Add 'group' option to display config yaml

wmww:
This allows mir-kiosk and other servers that use the YAML display config system to put outputs in logical groups (aka display walls). I can change the option name to `wall`, `display-wall` or something else if preferable, but IMO `group` is the most obvious.


AlanGriffiths (approved):
LGTM

bors r+

PR #1771, opened 2020-10-20
Fix missing #includes for GCC 11

Conan-Kudo:
This is from an effort in Fedora to prepare for the GCC 11 rebase for Fedora 34.


RAOF (approved):
Seems sensible. Thanks!

bors r+

PR #1772, opened 2020-10-20
[mir-kiosk] `run_with` the `ExternalClientLauncher`

AlanGriffiths:
Answers the question "Why does `WAYLAND_DISPLAY=wayland-mir ./bin/miral-kiosk --startup-apps weston-terminal` (built from master) fail with "Cannot launch apps when server has not started?""


wmww (approved):
Tested and working
bors r+

PR #1773, opened 2020-10-20
Allow adding and dropping Wayland extensions

### LINKED ISSUE #1757: Options to add/drop Wayland extensions
To customize the list of supported extensions, it shouldn't be required to pass in the whole list. There should be a way to add or drop specific extensions, while otherwise using the default set.

wmww:
This adds `--add-extensions` and `--drop-extension` options, which allow controlling enabled extensions without specifying an exhaustive list. Fixes #1757

### DIFF
--- a/src/platform/options/default_configuration.cpp
+++ b/src/platform/options/default_configuration.cpp
@@ -51,6 +51,8 @@ char const* const mo::composite_delay_opt         = "composite-delay";
 char const* const mo::enable_key_repeat_opt       = "enable-key-repeat";
 char const* const mo::x11_display_opt             = "enable-x11";
 char const* const mo::wayland_extensions_opt      = "wayland-extensions";
+char const* const mo::add_extensions_opt          = "add-extensions";
+char const* const mo::drop_extensions_opt         = "drop-extensions";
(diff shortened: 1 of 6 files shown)


src/platform/options/default_configuration.cpp
```diff
@@ -51,6 +51,8 @@ char const* const mo::composite_delay_opt         = "composite-delay";
 char const* const mo::enable_key_repeat_opt       = "enable-key-repeat";
 char const* const mo::x11_display_opt             = "enable-x11";
 char const* const mo::wayland_extensions_opt      = "wayland-extensions";
+char const* const mo::add_extensions_opt          = "add-extensions";
+char const* const mo::drop_extensions_opt         = "drop-extensions";
```
AlanGriffiths, line 55: Can we keep "wayland" in the configuration option? Not every configuration in Mir is Wayland related.

AlanGriffiths (changes requested):
Apart from the option names, this looks fine

src/platform/options/default_configuration.cpp
wmww: Done

AlanGriffiths (approved):
LGTM

bors r+

PR #1778, opened 2020-10-27
gbm-kms: Implement composite-bypass for Wayland clients

RAOF:
Client buffers submitted with the `zwp_linux_dmabuf_unstable_v1` are now possible candidates for composite-bypass.

This comes with a number of significant caveats.

Most significantly: there's no way for the compositor to suggest to the client that it should allocate buffers that are compatible with scanout (this requires [dmabuf-hints](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/8)), so whether or not a client *can* be bypassed is a function of whether or not the display hardware can cope with whatever buffer format the EGL driver chooses.

Furthermore, until we switch to the Atomic KMS API there's no way for us to usefully *query* the set of buffer formats the display hardware can scan out of, either.

### DIFF
--- a/include/platform/mir/graphics/dmabuf_buffer.h
+++ b/include/platform/mir/graphics/dmabuf_buffer.h
@@ -32,7 +32,7 @@ namespace graphics
 /**
  * A logical buffer backed by one-or-more dmabuf buffers
  */
-class DMABufBuffer
+class DMABufBuffer : public NativeBufferBase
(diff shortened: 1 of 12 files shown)


RAOF:
Marking this as no-merge as I'd like to get performance numbers with the no-alpha GLMark2 performance test (from #1780) first.

PR #1779, opened 2020-10-27
linux-dmabuf: More correct handling of DRM_FORMAT_MOD_INVALID

RAOF:
It is never correct to pass `DRM_FORMAT_MOD_INVALID` to any external API, so we don't need to be able to distinguish between “no modifier set” and “modifier set to `DRM_FORMAT_MOD_INVALID`”.

The exception here is that if the EGL driver claims support for `EGL_EXT_image_dma_buf_import_modifiers` but *also* does not list any acceptable modifiers for a format then that implies that the driver will perform dark magics on our behalf, and we should publish that format to the client with a `DRM_FORMAT_MOD_INVALID` modifier.

Allows `zwp_linux_dmabuf_unstable_v1` support to be enabled on drivers which don't support modifiers, like amdgpu.

### DIFF
--- a/src/platforms/gbm-kms/server/linux_dmabuf.cpp
+++ b/src/platforms/gbm-kms/server/linux_dmabuf.cpp
@@ -184,7 +184,7 @@ public:
         return image;
     }
 
-    private:
+private:
     void destroy() override
     {
         destroy_wayland_object();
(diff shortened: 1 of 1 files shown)


src/platforms/gbm-kms/server/linux_dmabuf.cpp
```diff
@@ -183,7 +183,8 @@ class DmaBufBuffer : public mir::wayland::Buffer
 
         return image;
     }
-private:
+
+    private:
```
AlanGriffiths, line 187: Stray whitespace change?

AlanGriffiths (approved):
OK

bors r+

PR #1780, opened 2020-10-27
performance-tests: Use a EGL visual with no alpha channel.

RAOF:
This makes more sense as a performance test, as very few windows
are going to want to blend their EGL rendering with what's behind them.

Also, this makes GLMark2 eligible for composite-bypass, when that is implemented,
and that's a very interesting optimisation to be able to see when it
works (and doesn't).


AlanGriffiths (approved):
Sure.

PR #1781, opened 2020-10-27
gbm-kms: Implement composite-bypass for Wayland clients

RAOF:
Client buffers submitted with the `zwp_linux_dmabuf_unstable_v1` are now possible candidates for composite-bypass.

This comes with a number of significant caveats.

Most significantly: there's no way for the compositor to suggest to the client that it should allocate buffers that are compatible with scanout (this requires [dmabuf-hints](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/8)), so whether or not a client *can* be bypassed is a function of whether or not the display hardware can cope with whatever buffer format the EGL driver chooses.

Furthermore, until we switch to the Atomic KMS API there's no way for us to usefully *query* the set of buffer formats the display hardware can scan out of, either.

### DIFF
--- a/include/platform/mir/graphics/dmabuf_buffer.h
+++ b/include/platform/mir/graphics/dmabuf_buffer.h
@@ -1,0 +1,67 @@
+/*
+ * Copyright © 2020 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Christopher James Halse Rogers <christopher.halse.rogers@canonical.com>
+ */
+
+#ifndef MIR_GRAPHICS_DMABUF_BUFFER_H_
+#define MIR_GRAPHICS_DMABUF_BUFFER_H_
+
+#include <cstdint>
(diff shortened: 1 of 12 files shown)


RAOF:
Again, adding no-merge so that we can get a new set of baseline performance runs from #1780 before this lands.

include/platform/mir/graphics/dmabuf_buffer.h
```diff
@@ -0,0 +1,67 @@
...
+class DMABufBuffer : public NativeBufferBase
+{
+public:
+    struct PlaneDescriptor
+    {
+        mir::Fd dma_buf;
+        uint32_t stride;
+        uint32_t offset;
+    };
+    virtual ~DMABufBuffer() = default;
```
AlanGriffiths, line 44: Unnecessary

AlanGriffiths (approved):

AlanGriffiths:
Possibly due to other factors, but over a few runs against "before" and "after"

```plain
glmark2-wayland --visual-config alpha=0 --benchmark jellyfish --fullscreen
```

Before|After
--|--
449|446

Results seem pretty stable and in the wrong direction.

AlanGriffiths:
With a different,more relevant, scenario I see better effects:
```plain
WAYLAND_DISPLAY=wayland-99 miral-kiosk --bypass on|off &
MIR_SERVER_WAYLAND_HOST=wayland-99 mir_demo_server --test-client glmark2-es2-wayland  --bypass on|off &
```
Before|After + `--bypass on`|After + `--bypass off`
--|--|--
glmark2 Score: 2903|glmark2 Score: 2932|glmark2 Score: 2905
glmark2 Score: 2904| glmark2 Score: 2912| 
. |glmark2 Score: 2934|

bors r+

AlanGriffiths:
With bypass just on the "host"
```plain
WAYLAND_DISPLAY=wayland-99 miral-kiosk --bypass on &
MIR_SERVER_WAYLAND_HOST=wayland-99 mir_demo_server --test-client glmark2-es2-wayland  --bypass off &
```
glmark2 Score: 2925
glmark2 Score: 2920

Nicely in the middle!

PR #1784, opened 2020-10-29
[miral-app] more robust launcher script

AlanGriffiths:
Avoid a racy startup, give better diagnostics on failure, tidy up help options

### DIFF
--- a/examples/miral-shell/miral-app.sh
+++ b/examples/miral-shell/miral-app.sh
@@ -30,11 +30,6 @@ if [ "${bindir}" != "" ]; then bindir="${bindir}/"; fi
 
 if [ "${miral_server}" == "miral-shell" ]
 then
-  if [ "$(lsb_release -c -s)" == "xenial" ]
-  then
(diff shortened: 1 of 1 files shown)


examples/miral-shell/miral-app.sh
```diff
@@ -61,6 +61,7 @@ else
     # Start mir_demo_server with the chosen WAYLAND_DISPLAY
     MIR_SERVER_ENABLE_X11=1 WAYLAND_DISPLAY=${wayland_display} ${bindir}${miral_server} $* --x11-displayfd 5 5>${x11_display_file}&
 
+    inotifywait --event close_write "${x11_display_file}"
```
RAOF, line 64: Should this have a timeout? By default this will wait forever, and if Mir for some reason doesn't close the fd the event will never come.

AlanGriffiths: I guess that would lead to a slightly less obscure failure mode, but we don't have an existing check for mir failing to start/crashing and this isn't any worse.

RAOF (approved):
LGTM

bors r+

PR #1785, opened 2020-10-29
linux-dmabuf: Fix sending invalid modifiers.

RAOF:
When iteration over nested containers it's important to use the correct
iteration variable. Otherwise you might accidentally read uninitialised
memory and send that as a modifier to the client!

Fixes the weird “invalid modifier” client errors we've recently seen.


AlanGriffiths (approved):
Amazing how obvious errors can be when you look at the right thing!

bors r+

Issue #1750, opened 2020-10-02
[Packaging] there should be meta packages for graphics platforms


AlanGriffiths:
For example, to get the "wayland" platform without specifying `...wayland18`

PR #1788, opened 2020-10-30
[debian] Add metapackages for each of the graphics platforms

AlanGriffiths:
Makes it possible, for example, to install just the "wayland" or "X" platform without hard coding the current version.

Fixes: #1750


RAOF (approved):
This looks reasonable.

bors r+

PR #1777, opened 2020-10-23
wl_display_roundtrip() must be called on the Wayland thread

AlanGriffiths:
Fixes: #1775

PR #1789, opened 2020-11-02
wl_display_roundtrip() should be called before starting the Wayland thread

AlanGriffiths:
Same reasoning as #1777 applied correctly


Saviq (changes requested):
I don't think you wanted the valgrind suppresions here?

Saviq (approved):
Been running the performance tests in a loop for a while, seems to fix!

bors merge

Saviq (approved):

PR #1803, opened 2020-11-09
Unbind wayland EGL display

### LINKED ISSUE #1799: [mir_performance_tests] intermittent failure running on X graphics platform
```plain
while mir_performance_tests; do :; done
```
Expect: runs indefinitely
Actual: Intermittently fails (various tests) with...
```plain
ERROR: Dynamic exception type: std::_Nested_exception<std::runtime_error>
std::exception::what: Failed to bind Wayland EGL display

unknown file: Failure
C++ exception with description "Failed to start server thread" thrown in SetUp().
```

wmww:
Fixes #1799. I'm adding a pure virtual method to `GraphicBufferAllocator`, which doesn't require changes to any symbols.map files but I *believe* breaks the platform ABI (correct me if I'm wrong). For that reason I'm marking this no-merge until I get confirmation one way or the other and edit the symbols.map appropriately. Also, if this is indeed a platform ABI break I'd like to also get the `EGL_WL_bind_wayland_display` check in there. I should have that ready by EOD.


AlanGriffiths:
On Mon, 9 Nov 2020 11:19:28 -0800, William Wold wrote:
> I&#39;m adding a pure virtual method to 
> `GraphicBufferAllocator`, which doesn&#39;t require changes to 
> any symbols.map files but I *believe* breaks the platform ABI 
> (correct me if I&#39;m wrong). For that reason I&#39;m marking 
> this no-merge until I get confirmation one way or the other

Yes, adding a virtual fcuntion to a class user code derives from breaks ABI

-- 

wmww:
This right? Do I need to bump mirplatform ABI in addition to mirplatformgraphics?

RAOF (commented):
If you wanted to do this in a way that *doesn't* break platform API you could do this cleanup in the various `~GraphicsBufferAllocator` destructors. It might be difficult to arrange the lifetimes of the various objects, though.

wmww:
> do this cleanup in the various `~GraphicsBufferAllocator` destructors

Tried that didn't work. Specifically I think the display needs to be unbound before other bits of the Wayland display. *maybe* we could shuffle around object order in `WaylandConnector` and make it work, but first attempt didn't.

RAOF:
> This right? Is it correct to go 2.1 -> 3.0? Do I need to bump mirplatform ABI in addition to mirplatformgraphics?

Actually, you probably want to go 2.1 -> 2.2 (the version stanza isn't the same as the SONAME). You do indeed need to bump the mirplatform ABI, because this has broken the mirplatform ABI ;).

(Bumps to mirplatform ABI don't necessarily require bumps to the graphics driver ABI, but most bumps to the graphics driver ABI will be accompanied by mirplatform ABI bumps)

wmww:
I think this is now good to go

RAOF:
Hah! For once, RPi isn't the failure :grin: 

Now that you've bumped ABIs you need to *also* update the packaging. The

AlanGriffiths (approved):
bors r+

Issue #914, opened 2019-07-07
Mir fails to start in a VM


wmww:
I have an Ubuntu 18.04 QEMU VM running Ubuntu 18.04. I've installed Mir from the release PPA, and it fails to find an input platform. Other people have encountered problems running the mate-wayland snap in a VM: https://github.com/mate-desktop/mate-panel/pull/991#issuecomment-508987183

wmww:
I have reproduced with latest master compiled from source in the VM. Log:
```
wmww@wmww-Standard-PC-Q35-ICH9-2009:~/mir/build$ ./bin/miral-shell
MIR_CLIENT_PLATFORM_PATH=./bin/../lib/client-modules/
MIR_SERVER_PLATFORM_PATH=./bin/../lib/server-modules/
LD_LIBRARY_PATH=./bin/../lib
exec=./bin/miral-shell.bin
[2019-07-07 09:46:26.002786] <information> mirserver: Starting
[2019-07-07 09:46:26.002945] < - debug - > mirserver: Not trying logind: "DISPLAY" is set and X need not have claimed the VT
[2019-07-07 09:46:26.003055] < - debug - > mirserver: Not using Linux VT subsystem for session management: Failed to find the current VT
[2019-07-07 09:46:26.003080] < - debug - > mirserver: No session management supported
[2019-07-07 09:46:26.003113] <information> VT switch key handler: No VT switching support available: MinimalConsoleServices does not support VT switching
[2019-07-07 09:46:26.003291] <information> mircommon: Loading modules from: ./bin/../lib/server-modules/
[2019-07-07 09:46:26.003345] <information> mircommon: Loading module: ./bin/../lib/server-modules/server-mesa-x11.so.16
[2019-07-07 09:46:26.003363] <information> mircommon: Loading module: ./bin/../lib/server-modules/graphics-mesa-kms.so.16
[2019-07-07 09:46:26.003377] <information> mircommon: Loading module: ./bin/../lib/server-modules/input-evdev.so.7
[2019-07-07 09:46:26.005728] <information> mircommon: Loading module: ./bin/../lib/server-modules/graphics-throw.so
[2019-07-07 09:46:26.005755] <information> mircommon: Loading module: ./bin/../lib/server-modules/input-stub.so
[2019-07-07 09:46:26.006496] <information> mircommon: Loading module: ./bin/../lib/server-modules/graphics-dummy.so
[...]

PR #1804, opened 2020-11-10
Correctly check EGL extensions

wmww:
`eglGetProcAddress()` may return a non-null function pointer even if an extension isn't supported, and some extensions (like the Wayland ones) require an EGL display to check properly. This refactors the code to make the required checks. Does not entirely address #914, but fixes the crash. Stacked on #1803.


src/platform/graphics/egl_extensions.cpp
```diff
@@ -69,7 +58,6 @@ mg::EGLExtensions::EGLExtensions() :
      */
     glEGLImageTargetTexture2DOES{
```
RAOF, line 59: This (and the `egl{Create,Destroy}ImageKHR` extensions above) also should really be checked properly. Only EGL client extensions (of which `EGL_EXT_platform_base` is the only one we use here) can be checked without already having an EGL Display.

That would make for a larger change, but also might make the rest of the changes a bit nicer - by making the big change you could pass in fully-functional extension objects, rather than having the two-step initialisation `optional<foo>` dance?

wmww: I meant to ask about that. https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_image_base.txt doesn't mention if it's a client or display extension. Which string do we check?

RAOF: Everything that's not explicitly a client extension is a display extension. Client extensions didn't exist at all until [EGL_EXT_client_extensions](https://www.khronos.org/registry/EGL/extensions/EXT/EGL_EXT_client_extensions.txt), and that was a fair way into EGL 1.4's lifetime.

wmww: Is `GL_OES_EGL_image_external` the correct name to test if `glEGLImageTargetTexture2DOES` is available? Or is that also `EGL_KHR_image_base`?

wmww:
Didn't finish the rework today

src/platform/graphics/egl_extensions.cpp
RAOF: You're looking for [GL_OES_EGL_image](https://www.khronos.org/registry/OpenGL/extensions/OES/OES_EGL_image.txt), yes. Note that this is actually a GL extension, not an EGL extension, so it's not entirely clear to me why it's in `EGLExtensions`.

AlanGriffiths:
We landed #1806

PR #1806, opened 2020-11-11
Check EGL extensions before using them

wmww:
I've taken a number of stabs at this. Unfortunately the variadic template function pointer variant didn't work out so well so we're left with this. It solves the problem reasonably well in a relatively small diff, so good enough. Related to #914 and improved alternative to #1804.

### DIFF
--- a/include/platform/mir/graphics/egl_extensions.h
+++ b/include/platform/mir/graphics/egl_extensions.h
@@ -247,7 +247,7 @@ struct EGLExtensions
     private:
         std::atomic<bool> mutable has_initialized{false};
         std::mutex mutable mutex;
-        EGLDisplay mutable cached_display{0};
+        EGLDisplay mutable cached_display{EGL_NO_DISPLAY};
         std::optional<Ext> mutable cached_ext;
     };
 
--- a/include/platform/mir/graphics/egl_extensions.h
+++ b/include/platform/mir/graphics/egl_extensions.h
@@ -45,6 +45,10 @@
 #define MIR_GRAPHICS_EGL_EXTENSIONS_H_
 
 #include <optional>
+#include <mutex>
+#include <atomic>
+
+#include <boost/throw_exception.hpp>
 
 #define EGL_EGLEXT_PROTOTYPES
 #include <EGL/egl.h>
@@ -221,22 +225,29 @@ struct EGLExtensions
     {
         auto operator()(EGLDisplay dpy) const -> Ext const&
         {
-            if (cached_display && cached_display.value() != dpy)
+            if (!has_initialized)
             {
-                cached_ext = std::nullopt;
+                std::lock_guard<std::mutex> lock{mutex};
+                if (!has_initialized)
+                {
+                    cached_display = dpy;
+                    cached_ext.emplace(dpy);
+                    has_initialized = true;
+                }
             }
 
-            if (!cached_ext)
+            if (dpy != cached_display)
             {
-                cached_display = dpy;
-                cached_ext.emplace(dpy);
+                BOOST_THROW_EXCEPTION(std::logic_error("Multiple EGL displays used with the same extension object"));
             }
 
             return cached_ext.value();
         }
 
     private:
-        std::optional<EGLDisplay> mutable cached_display;
+        std::atomic<bool> mutable has_initialized{false};
+        std::mutex mutable mutex;
+        EGLDisplay mutable cached_display{0};
         std::optional<Ext> mutable cached_ext;
     };
 
--- a/include/platform/mir/graphics/egl_extensions.h
+++ b/include/platform/mir/graphics/egl_extensions.h
@@ -44,7 +44,7 @@
 #ifndef MIR_GRAPHICS_EGL_EXTENSIONS_H_
 #define MIR_GRAPHICS_EGL_EXTENSIONS_H_
 
-#include <experimental/optional>
+#include <optional>
 
 #define EGL_EGLEXT_PROTOTYPES
 #include <EGL/egl.h>
@@ -223,7 +223,7 @@ struct EGLExtensions
         {
             if (cached_display && cached_display.value() != dpy)
             {
-                cached_ext = std::experimental::nullopt;
+                cached_ext = std::nullopt;
             }
 
             if (!cached_ext)
@@ -236,8 +236,8 @@ struct EGLExtensions
         }
 
     private:
-        std::experimental::optional<EGLDisplay> mutable cached_display;
-        std::experimental::optional<Ext> mutable cached_ext;
+        std::optional<EGLDisplay> mutable cached_display;
+        std::optional<Ext> mutable cached_ext;
     };
 
     EGLExtensions();
@@ -277,7 +277,7 @@ struct EGLExtensions
         PFNEGLGETPLATFORMDISPLAYEXTPROC const eglGetPlatformDisplay;
         PFNEGLCREATEPLATFORMWINDOWSURFACEEXTPROC const eglCreatePlatformWindowSurface;
     };
-    std::experimental::optional<PlatformBaseEXT> const platform_base;
+    std::optional<PlatformBaseEXT> const platform_base;
 
     class DebugKHR
     {
@@ -285,7 +285,7 @@ struct EGLExtensions
         DebugKHR();
 
         static DebugKHR extension_or_null_object();
-        static std::experimental::optional<DebugKHR> maybe_debug_khr();
+        static std::optional<DebugKHR> maybe_debug_khr();
 
         PFNEGLDEBUGMESSAGECONTROLKHRPROC const eglDebugMessageControlKHR;
         PFNEGLLABELOBJECTKHRPROC const eglLabelObjectKHR;
--- a/include/platform/mir/graphics/egl_extensions.h
+++ b/include/platform/mir/graphics/egl_extensions.h
@@ -216,28 +216,60 @@ namespace graphics
 {
 struct EGLExtensions
 {
+    template<typename Ext>
+    struct LazyDisplayExtensions
+    {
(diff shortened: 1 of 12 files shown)


src/platform/graphics/egl_extensions.cpp
```diff
@@ -56,30 +45,37 @@ std::experimental::optional<mg::EGLExtensions::PlatformBaseEXT> maybe_platform_b
...
-    if (!eglCreateImageKHR || !eglDestroyImageKHR)
-        BOOST_THROW_EXCEPTION(std::runtime_error("EGL implementation doesn't support EGLImage"));
+    auto const egl_extensions = eglQueryString(dpy, EGL_EXTENSIONS);
+    if (!egl_extensions || !strstr(egl_extensions, "EGL_KHR_image_base") || !eglCreateImageKHR || !eglDestroyImageKHR)
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error("EGL display doesn't support EGL_KHR_image_base"));
+    }
 
-    if (!glEGLImageTargetTexture2DOES)
+    if (!egl_extensions || !strstr(egl_extensions, "GL_OES_EGL_image") || !glEGLImageTargetTexture2DOES)
```
RAOF, line 72: Have you tested this? Because I'm pretty sure this will fail, as `GL_OES_EGL_image` is a *GL* extension, not an EGL one.

RAOF (commented):
Hm. It's not the approach I would have taken, but it's not wrong (apart from the checking for `GL_OES_EGL_image`, which *is* wrong :wink:)

I was thinking that we'd just refactor things to pass an `EGLDisplay` into everywhere that needs to construct an `EGLExtensions`, but that might be hard.

src/platform/graphics/egl_extensions.cpp
```diff
@@ -56,30 +45,37 @@ std::experimental::optional<mg::EGLExtensions::PlatformBaseEXT> maybe_platform_b
...
-    if (!eglCreateImageKHR || !eglDestroyImageKHR)
-        BOOST_THROW_EXCEPTION(std::runtime_error("EGL implementation doesn't support EGLImage"));
+    auto const egl_extensions = eglQueryString(dpy, EGL_EXTENSIONS);
+    if (!egl_extensions || !strstr(egl_extensions, "EGL_KHR_image_base") || !eglCreateImageKHR || !eglDestroyImageKHR)
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error("EGL display doesn't support EGL_KHR_image_base"));
+    }
 
     if (!glEGLImageTargetTexture2DOES)
+    {
```
RAOF, line 73: This is still wrong; we're using the extension without checking for its presence. That said, this is a strict improvement on the status quo.

src/platform/graphics/egl_extensions.cpp
wmww: Fixed

include/platform/mir/graphics/egl_extensions.h
```diff
@@ -216,28 +216,60 @@ namespace graphics
...
+            {
+                cached_display = dpy;
+                cached_ext.emplace(dpy);
+            }
+
+            return cached_ext.value();
+        }
+
+    private:
+        std::experimental::optional<EGLDisplay> mutable cached_display;
```
RAOF, line 239: Nit: We're on C++17 now. `std::optional<foo>`, baby!

RAOF (commented):
This is not threadsafe, and we access these things on multiple threads. Empirically, TSan doesn't find any problems with the code, but that's a fairly weak test.

For the sake of paranoia, maybe we should do some proper locking?

src/platform/graphics/egl_extensions.cpp
wmww: Can you fix it then? I do not know what I'm supposed to do here.

RAOF: Fair enough. We need to move `glEGLImageTargetTexture2DOES` entirely out of `EGLExtensions`, as it's not an EGL extension, and instead have a `GLExtensions` object which gets constructed with an active GL context (as GL extensions are per *context*).

wmww: I created #1808 so we wont loose it if we don't deal with it now.

RAOF: Ta

include/platform/mir/graphics/egl_extensions.h
```diff
@@ -216,44 +220,83 @@ namespace graphics
...
+                BOOST_THROW_EXCEPTION(std::logic_error("Multiple EGL displays used with the same extension object"));
+            }
+
+            return cached_ext.value();
+        }
+
+    private:
+        std::atomic<bool> mutable has_initialized{false};
+        std::mutex mutable mutex;
+        EGLDisplay mutable cached_display{0};
```
RAOF, line 250: `EGL_NO_DISPLAY` might be better?

include/platform/mir/graphics/egl_extensions.h
```diff
@@ -216,44 +220,83 @@ namespace graphics
...
+                    has_initialized = true;
+                }
+            }
+
+            if (dpy != cached_display)
+            {
+                BOOST_THROW_EXCEPTION(std::logic_error("Multiple EGL displays used with the same extension object"));
+            }
+
+            return cached_ext.value();
```
RAOF, line 244: I *think* this is safe, but I'd appreciate @AlanGriffiths' review of this.

RAOF (approved):
I'm now good with this, but I'd like @AlanGriffiths to check the threading bit.

include/platform/mir/graphics/egl_extensions.h
wmww: Btw according to this totally reliable and guaranteed-to-be-correct [StackOverflow thread](https://stackoverflow.com/questions/40320254/reordering-atomic-operations-in-c) this sort of thing is safe

include/platform/mir/graphics/egl_extensions.h
wmww: Done

include/platform/mir/graphics/egl_extensions.h
AlanGriffiths: Yes, the default on `atomic<>` is global sequential consistency. Hard thinking is only required of we loosen that requirement.

AlanGriffiths:
Now the traditional problem with platform changes: RPi...
```plain
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp: In member function â€˜virtual void mir::graphics::rpi::BufferAllocator::bind_display(wl_display*, std::shared_ptr<mir::Executor>)â€™:
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp:383:9: error: no match for â€˜operator!â€™ (operand type is â€˜const mir::graphics::EGLExtensions::LazyDisplayExtensions<mir::graphics::EGLExtensions::WaylandExtensions>â€™)
  383 |     if (!egl_extensions->wayland)
      |         ^~~~~~~~~~~~~~~~~~~~~~~~
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp:383:9: note: candidate: â€˜operator!(bool)â€™ <built-in>
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp:383:9: note:   no known conversion for argument 1 from â€˜const mir::graphics::EGLExtensions::LazyDisplayExtensions<mir::graphics::EGLExtensions::WaylandExtensions>â€™ to â€˜boolâ€™
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp:388:32: error: base operand of â€˜->ï¿½ï¿½ï¿½ has non-pointer type â€˜const mir::graphics::EGLExtensions::LazyDisplayExtensions<mir::graphics::EGLExtensions::WaylandExtensions>â€™
  388 |     if (egl_extensions->wayland->eglBindWaylandDisplayWL(dpy, display) == EGL_FALSE)
      |                                ^~
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp: In member function â€˜virtual void mir::graphics::rpi::BufferAllocator::unbind_display(wl_display*)â€™:
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp:406:9: error: no match for â€˜operator!â€™ (operand type is â€˜const mir::graphics::EGLExtensions::LazyDisplayExtensions<mir::graphics::EGLExtensions::WaylandExtensions>â€™)
  406 |     if (!egl_extensions->wayland)
      |         ^~~~~~~~~~~~~~~~~~~~~~~~
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp:406:9: note: candidate: â€˜operator!(bool)â€™ <built-in>
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp:406:9: note:   no known conversion for argument 1 from â€˜const mir::graphics::EGLExtensions::LazyDisplayExtensions<mir::graphics::EGLExtensions::WaylandExtensions>â€™ to â€˜boolâ€™
/spread/mir/src/platforms/rpi-dispmanx/buffer_allocator.cpp:411:32: error: base operand of â€˜->â€™ has non-pointer type â€˜const mir::graphics::EGLExtensions::LazyDisplayExtensions<mir::graphics::EGLExtensions::WaylandExtensions>â€™
  411 |     if (egl_extensions->wayland->eglUnbindWaylandDisplayWL(dpy, display) == EGL_FALSE)
      |                                ^~
[ 14%] Linking CXX executable ../../../bin/mir_wayland_generator
```

AlanGriffiths (approved):
I think we can _land_ without testing dispmanx, but we ought to test before tagging the release. @RAOF is there a protocol for this testing documented somewhere?

bors r+

Issue #1801, opened 2020-11-09
glmark2-es-wayland crashes under GLMark2Wayland_fullscreen with bypass on Pi3


Saviq:
With this testflinger host:

```yaml
job_queue: rpi3-001
provision_data:
    url: http://cdimage.ubuntu.com/releases/focal/release/ubuntu-20.04.1-preinstalled-server-armhf+raspi.img.xz
reserve_data:
    ssh_keys:
    - lp:${YOU}
```

Running this (with either `/dev` or `/rc` PPAs):
```bash
$ while env MIR_SERVER_BYPASS=off XDG_RUNTIME_DIR=/tmp mir_performance_tests --logging on --gtest_filter=GLMark2Wayland.fullscreen; do true; done
```

Will run forever. But take `MIR_SERVER_BYPASS` out and you'll see `glmark2-es-wayland` crashing real soon:
```gdb
#0  dri2_query_image_common (value=0xbef74250, attrib=<optimized out>, image=0x0) at ../src/gallium/state_trackers/dri/dri2.c:1048
#1  dri2_query_image (image=0x0, attrib=<optimized out>, value=0xbef74250) at ../src/gallium/state_trackers/dri/dri2.c:1236
#2  0xb69673be in create_wl_buffer (dri2_dpy=dri2_dpy@entry=0x15e5000, dri2_surf=dri2_surf@entry=0x160afd0, image=0x0) at ../src/egl/drivers/dri2/platform_wayland.c:901
#3  0xb6967b4a in dri2_wl_swap_buffers_with_damage (drv=<optimized out>, disp=0x15e47c8, draw=0x160afd0, rects=rects@entry=0x0, n_rects=n_rects@entry=0) at ../src/egl/drivers/dri2/platform_wayland.c:1094
#4  0xb6967bb8 in dri2_wl_swap_buffers (drv=<optimized out>, disp=<optimized out>, draw=<optimized out>) at ../src/egl/drivers/dri2/platform_wayland.c:1169
#5  0xb6960e42 in dri2_swap_buffers (drv=0x15e4f40, disp=0x15e47c8, surf=0x160afd0) at ../src/egl/drivers/dri2/egl_dri2.c:1932
#6  0xb695b928 in eglSwapBuffers (dpy=0x15e47c8, surface=<optimized out>) at ../src/egl/main/eglapi.c:1340
#7  0x004b9892 in CanvasGeneric::update (this=0xbef74470) at ../src/canvas-generic.cpp:114
[...]

PR #1809, opened 2020-11-12
gbm-kms: Disable bypass by default.

RAOF:
This triggers a GPU OOM (issue #1801) on the Pi3 when used with default settings.
There may be due to us hanging on to buffers longer than they need to be, but
it's not clear yet.

Users who want the performance boost and are running on known hardware can
set `--bypass=true`, but by default use the safe option of `false`.


AlanGriffiths (approved):
LGTM

bors r+

PR #1812, opened 2020-11-13
Use standard key repeat values

wmww:
This solves a minor annoyance I've had for a while. Both Weston and Sway use these values by default. GNOME-on-Wayland, on the other hand uses 33 and 500, which I would also be open to. Mostly, the delay just needs to go up.

Of course this will eventually need be configurable for accessability reasons and general preference, but building an input configuration system is a job for another day,


AlanGriffiths (approved):
bors r+

PR #1815, opened 2020-11-15
Fix build with MIR_ENABLE_WLCS_TESTS=ON and wlcs in non-standard prefix

amezin:
When trying to build Mir with wlcs installed into a prefix in my /home, I'm
getting the following error:

    In file included from /home/amezin/mir-egmde/mir/tests/mir_test_framework/test_wlcs_display_server.cpp:20:
    /home/amezin/mir-egmde/mir/include/test/miral/test_wlcs_display_server.h:24:10: fatal error: wlcs/display_server.h: No such file or directory
       24 | #include <wlcs/display_server.h>
          |          ^~~~~~~~~~~~~~~~~~~~~~~
    compilation terminated.

It seems that wlcs include directory is missing from include paths (because
it was never added).

Signed-off-by: Aleksandr Mezin <mezin.alexander@gmail.com>


AlanGriffiths:
@amezin thanks for highlighting this issue. While your solution works, I think there's a simpler answer: #1816

AlanGriffiths:
Alternative fix landed: #1816

PR #1816, opened 2020-11-16
Fix wlcs integration

AlanGriffiths:
As highlighted by #1815 this is broken


wmww (approved):
I've confirmed this doesn't break anything when building without WLCS installed
bors r+

PR #1823, opened 2020-11-24
Clamp window resize to min/max

### LINKED ISSUE #1821: Window manager does not respect min/max size during interactive resize
It seems that the window manager respects min/max size when they are initially set, but not when the window is resized thereafter.

wmww:
Fixes #1821. Since interactive resize and programmatic policy resize looks the same to the window manager, both are clamped. The policy can, however, change the min/max size to whatever it wants. It's documented that that is required and the tiling window manager has been updated to do that.

There was previously logic to prevent maximized windows from being moved or resized. This was untested, and it seems to be unnecessary. It has been dropped because it was breaking the tiling window manager after I made my changes. Manually moving maximized windows is still impossible in both window managers.

### DIFF
--- a/examples/example-server-lib/tiling_window_manager.cpp
+++ b/examples/example-server-lib/tiling_window_manager.cpp
@@ -178,6 +178,19 @@ void reset(mir::optional_value<ValueType>& option)
 {
     if (option.is_set()) option.consume();
 }
+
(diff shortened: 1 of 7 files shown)


AlanGriffiths (approved):
Works for me

bors merge

PR #1825, opened 2020-11-25
Only call update_attached_and_fullscreen_sets() on state change

wmww:
Minor change to prevent an extra call. `update_attached_and_fullscreen_sets()` is slightly expensive and no need to call it if nothing has changed.


AlanGriffiths (approved):
Sure

bors merge

PR #1826, opened 2020-11-25
Shrink initially maximized windows on restore

### LINKED ISSUE #1817: Windows don't change size when restoring if they started maximized
I'm not exactly sure how it gets into this state, but for me I'll often have a maximized window that unmaximizes to about the same size and position it was before. If I don't try to move it or notice the subtle change in decorations, I can't even tell it's been unmaximized. In my opinion, it's a better user experience for windows to always get smaller when they are unmaximized.

I plan to implement this at some point. Please leave any feedback, as it's always easier here than in code review.

wmww:
Fixes #1817 and adds tests for it and other maximization behavior. Calculating the restore rect takes into account the window's size and min size, and it's done lazily rather than on construction to take into account the latest values of those properties.


tests/miral/window_placement_maximized.cpp
```diff
@@ -0,0 +1,201 @@
+/*
+ * Copyright © 2019 Canonical Ltd.
```
AlanGriffiths, line 2: I suspect the year is inaccurate?

AlanGriffiths (approved):
LGTM

bors r+

AlanGriffiths (approved):
bors r+

Issue #1829, opened 2020-12-01
Geometry classes should support floating-point types


wmww:
Rather than re-writing floating-point variants of everything, I propose we change the current types to templates and make specializations of them. `PointI` would be an alias for `Point<int>` and an equivalent of the current `Point`, `DisplacementF` would be an alias of `Displacement<float>`, etc. This would give our input handling code the same level of type safety as the rest of Mir.

wmww:
I have an idea for the *right* way to do this. I started working on it but it's a lot of typing so I don't want to continue until I've gotten a green-light for the idea.

1. Replace `IntWrapper<Tag>` with a more generic `Tagged<Tag, T>` with `T` being the value type.
2. Use `using ... = ...` to define generic versions of dimensions: `WidthT`, `HeightT`, `XT`, `YT`, `DeltaXT` and `DeltaYT` (`T` stands for template, open to other suggestions. We can't use the names directly because they will still mean the int versions for backwards compat)
3. Define `Width`, `Height`, `X`, etc to be the int versions
4. Define `WidthF`, `HeightF`, `XF`, etc to be the float versions
5. Change all the operators to templates
6. Add conversions that change type but don't change tag
7. Change `Point` to more generic `PointT<XType, YType>`,
8. Define `Point` to be the `<Width, Height>` int version, and `PointF` to be the float version
9. Add safe casting to `PointT` as well
10. Do 7 - 9 for `Displacement`, `Size` and `Rectangle`

Compared to making entirely new classes for floating points, this would:
- Lower cost of maintenance, implementation is shared
- Allow allow for arbitrary types (sometimes we need a `SizeT<std::optional<Width>, std::optional<Height>>`)
- Allow safely converting between any types, (It should be easier to go `Width` -> `WidthF` than `Width` -> `HeightF`). Implementing this for just ints and floats might be possible but it would not scale well to any other types.

Unknowns:
- Would this be ABI stable? We'd be changing the type signatures but under the hood it should always just compile down to raw numbers.

AlanGriffiths:
>     * Would this be ABI stable? We'd be changing the type signatures but under the hood it should always just compile down to raw numbers.

No. For one thing type names are encoded in function signatures

wmww:
After discussion, we decided we might be able to make generic types work if we change our existing types to inherit from them. I will attempt this and put up a draft PR

AlanGriffiths:
I think this was addressed by #1858

Issue #1732, opened 2020-09-23
Clean up Window/Surface/BufferStream abstractions


wmww:
We need to clean up Mir's abstractions and make them more closely match the Wayland concepts. This will reduce duplicated logic in the fronted and fix a number of bugs.

PR #1830, opened 2020-12-01
Send sub-pixel input events to Wayland clients

### LINKED ISSUE #1814: Support precision scrolling
Precision scrolling for touchpad users is an absolute must for a modern desktop. It looks like `MirPointerEvent` uses floats for all axis, so there might not be too much that needs to be plumbed up. Precision mouse movements might be nice as well.

### LINKED ISSUE #1828: Send floating point pointer/touch position to Wayland clients
Mir events support floating point pointer and touch position, but it's collapsed down to ints in the Wayland frontend. Shouldn't be.

wmww:
Fixes #1814, fixes #1828. This area of code will be a little nicer when #1829 is done, but I don't think we should block on that. As you can see we use the ugly and rather unsafe `std::pair<float, float>` types to represent sub-pixel positions, but the amount of code that has to handle this is not large.

Another option would be to send the event all the way through until we hit the Wayland wrapper call. This would be a good idea, but it seems pointer events aren't mutable and generic events aren't queryable, so we'd have to do a bunch of messy casting and IMO it's not worth it.

In the long run all this code will disappear once we address #1732 and make subsurfaces real input surfaces.


AlanGriffiths (approved):
bors r+

wmww:
bors r=AlanGriffiths

PR #1833, opened 2020-12-03
CMake: Modernise mirplatform target.

RAOF:
Modern CMake allows us to add dependencies to OBJECT libraries. They then
get their library dependencies *and* their header dependencies added to
whatever you link them to.

With a little helper to combine a bunch of OBJECT libraries into a
SHARED library, this allows us to make all of the OBJECT libraries that
make up `mirplatform` independent and automatically usable by simply
`target_link_libraries`ing.

Ideally we'd also be able to aggregate this, so we had a `mirplatform-obj`
for the unit-tests that got built into a `mirplatform` shared library.
Unfortunately I've not found a way to make that work properly, so
the `MIR_PLATFORM_OBJECTS` variable remains to list the OBJECT libraries
a unit-test testing `mirplatform` will need to link to.

This removes the necessity for the `MIR_PLATFORM_REFERENCES` variable,
and for any `include_directories()` calls required by linking to
platform objects.


RAOF:
The final goal here is to make all the dependencies local to the CMakeLists.txt that builds the library, and then fix the various bits where the build creates files in the source tree rather than the build tree.

AlanGriffiths:
Fedora 32 failure look like something went wrong setting up the udev mocks. This could relate to platform setup, but re-running in case it is transient (it worked on rawhide).

AlanGriffiths:
OK, this clearly breaks on Fedora (and looking at rawhide that is a fake pass).

RAOF:
Hm. This seems to result in a change of linker behaviour, resulting in (at least):
* We can't `dlopen("mir_unit_tests.bin")` anymore, and
* Typeinfo isn't quite right

AlanGriffiths:
This experiment failed. Sad, I'd like to modernise the cmake scripts too.

PR #1840, opened 2020-12-04
clang-format

AlanGriffiths:
Here is a refresh of the .clang-format options along with the result of running on the MirAL code.

Most of the changes I can live with. But whatever options I tweak I find some that are rather painful.

PR'd so that people can form their own opinions.


wmww:
My take on formatting:

Pros:
1. Easier to write code (personally don't find this significant)
2. Easier to read code (again, personally don't have a problem with slightly inconsistently formatted code)
3. Eliminate formatting discussions in PRs (doesn't seem to be a big problem for us)

Cons:
1. clang-format never does quite what you want it to
2. Hand-formatting can produce more readable code (and much more readable diffs) than even the ideal formatter
3. Less things to go wrong (forgot to run formatter, can't set up in IDE, CI has different version than local, etc etc)

Probably clear which side I fall on, but I don't have a super strong preference either way.

include/miral/miral/canonical_window_manager.h
```diff
@@ -55,10 +52,9 @@ class CanonicalWindowManagerPolicy  : public WindowManagementPolicy
     /// Move the child window with the parent
     auto confirm_inherited_move(WindowInfo const& window_info, Displacement movement) -> Rectangle override;
 
-    auto confirm_placement_on_display(
-        WindowInfo const& window_info,
-        MirWindowState new_state,
-        Rectangle const& new_placement) -> Rectangle override;
+    auto confirm_placement_on_display(WindowInfo const& window_info,
+                                      MirWindowState new_state,
+                                      Rectangle const& new_placement) -> Rectangle override;
```
RAOF, line 57: Hard pass.

If we can't get clang-format to format these as continuations rather than box-align them then I think we should give up on this; box-align will lead to stupendously long lines.

RAOF (changes requested):
I'm vaguely in favour of this. All the formatting changes look sensible to me, with the glaring exception of my inline comment.

That said, I don't think this is a huge issue for us, so if it's hard to make clang-format do something sensible with parameter lists, give up quickly and drop this :wink:

include/miral/miral/canonical_window_manager.h
AlanGriffiths: The two options relevant to this block are:
> BinPackParameters: boolean
> If `false`, a function declaration’s or function definition’s parameters will either all be on the same line or will have one line each. 
```
true:
void f(int aaaaaaaaaaaaaaaaaaaa, int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa) {}

false:
void f(int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa) {}
```
and

> AllowAllParametersOfDeclarationOnNextLine: boolean
> If the function declaration doesn’t fit on a line, allow putting all parameters of a function declaration onto the next line even if `BinPackParameters` is `false`. 
```
true:
void myFunction(
    int a, int b, int c, int d, int e);

false:
void myFunction(int a,
                int b,
                int c,
                int d,
                int e);
```
Neither matches our preferences.

src/miral/external_client.cpp
```diff
@@ -71,31 +73,25 @@ void miral::ExternalClientLauncher::operator()(mir::Server& server)
...
-        app_x11_env,
-        "X11 changes to --app-env for launched apps",
-        default_x11_env);
+    server.add_configuration_option(app_x11_env, "X11 changes to --app-env for launched apps", default_x11_env);
 
-    server.add_init_callback([self=self, &server]
-         {
-             auto const options = server.get_options();
+    server.add_init_callback([self = self, &server] {
+        auto const options = server.get_options();
```
AlanGriffiths, line 88: There seems to be no option to place a newline _before_ the opening brace of a lambda to be consistent with the rest of our brace placement.

src/miral/window_management_trace.cpp
```diff
@@ -138,28 +140,42 @@ auto dump_of(miral::WindowInfo const& info) -> std::string
...
+        if (info.min_aspect() != miral::default_min_aspect_ratio)
+            APPEND(min_aspect);
+        if (info.max_aspect() != miral::default_max_aspect_ratio)
+            APPEND(max_aspect);
+        if (info.depth_layer() != mir_depth_layer_application)
+            APPEND(depth_layer);
+        if (info.attached_edges() != 0)
+            APPEND(attached_edges);
+        if (info.exclusive_rect() != mir::geometry::Rectangle{{}, {}})
+            APPEND(exclusive_rect);
```
AlanGriffiths, line 169: I know the "pre" of this block is an exception to our usual style, but I think it is easier to maintain than the "post".

OTOTH it is possible to disable clang-format for a block so exceptions can be made:
```c++
// clang-format off
    void    unformatted_code  ;
// clang-format on
```

AlanGriffiths:
There's no interest in pursuing this further.

PR #1844, opened 2020-12-07
CMake: Support arch-indep-only builds.

sunweaver:


sunweaver:
This PR should go together with various changes on debian/rules.

See: https://salsa.debian.org/mir-server-team/mir/-/commits/master/debian/rules (and scan the logs for arch-only and arch-indep changes).

sunweaver:
For testing, try dpkg-buildpackage -A builds and dpkg-buildpackage -b builds.

AlanGriffiths:
> This PR should go together with various changes on debian/rules.

That's what is needed to pass sbuild in CI?

RAOF:
Hm. Why are you splitting up the arch-dependent and arch-independent build components in the first place? I've done that before, with colord, but that was because the arch-indep step was more than half the build time and used significantly more memory, neither of which is true here?

To the best of my knowledge the doxygen build is a small fraction of the build time and resources. I'd like to know what the benefit is before making the install more cumbersome (as now `install` isn't actually going to install everything; you additionally need to `install-doc`, yes?)

AlanGriffiths:
Lost interest?

PR #1849, opened 2020-12-08
[gha] use org-wide GPG secret

Saviq:
---

This simulates a master push: https://github.com/MirServer/mir/actions/runs/408292731


AlanGriffiths (approved):
If it works...

bors r+

PR #1853, opened 2020-12-08
Disable natural scrolling

### LINKED ISSUE #1813: Scrolling is inverted
Mir seems to only supported inverted "natural" scrolling on touchpads. It appears this flip is hard-coded into the input platforms. I'm not sure if we want to flip the Mir-internal representation or not, but we should probably default to the more common normal scrolling and at some point make it configurable

wmww:
Fixes #1813. This stops flipping vertical scroll direction on the X11 and libinput platforms. It was already unflipped on the Wayland platform.

On X11 I think this is an easy decision. I believe it's currently *always* the inverse of the host X11 environment. If that's correct, I think that's obviously wrong. IMO even when we have input settings they should not effect Mir running nested. Instead, we should respect the configuration of the host system.

For libinput, This just changes the default behavior. Normal (as opposed to natural) scrolling is the default on most Linux systems, so I think that is reasonable.

### DIFF
--- a/tests/mir_test_doubles/mock_libinput.cpp
+++ b/tests/mir_test_doubles/mock_libinput.cpp
@@ -823,6 +823,10 @@ libinput_event* mtd::MockLibInput::setup_axis_event(libinput_device* dev, uint64
         .WillByDefault(Return(vertical));
     ON_CALL(*this, libinput_event_pointer_get_axis_value_discrete(pointer_event, LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL))
         .WillByDefault(Return(horizontal));
+    ON_CALL(*this, libinput_event_pointer_get_axis_value(pointer_event, LIBINPUT_POINTER_AXIS_SCROLL_VERTICAL))
+        .WillByDefault(Return(vertical));
+    ON_CALL(*this, libinput_event_pointer_get_axis_value(pointer_event, LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL))
+        .WillByDefault(Return(horizontal));
     return event;
 }
 
--- a/spread/build/fedora/task.yaml
+++ b/spread/build/fedora/task.yaml
@@ -46,7 +46,7 @@ execute: |
         libdrm-devel
 
     BUILD_DIR=$(mktemp --directory)
-    cmake -H$SPREAD_PATH -B$BUILD_DIR -DCMAKE_BUILD_TYPE=Debug -DMIR_USE_LD_GOLD=ON -DMIR_ENABLE_WLCS_TESTS=OFF
+    cmake -H$SPREAD_PATH -B$BUILD_DIR -DCMAKE_BUILD_TYPE=Debug -DMIR_USE_LD=gold -DMIR_ENABLE_WLCS_TESTS=OFF
     # Run cmake again to pick up wlcs?!?!?!?!
     cmake $BUILD_DIR
     export VERBOSE=1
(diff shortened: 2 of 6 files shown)


wmww:
This now also changes the scale of scroll events.

Previously `1.0` represented a wheel tick. This is fine, except it doesn't match the Wayland units which has scroll being a 2D vector in surface coordinates. Libinput continuous scroll uses the Wayland units, which means we were dividing by `15` in the libinput platform and multiplying by `10` (eyeballed) in the Wayland frontend. On the Wayland platform this wasn't noticed, so we don't adjust it. That means scroll was getting 10x faster for each level of nesting on a Wayland compositor.

Simpler solution is to use scroll-as-a-2D-vector all the way through Mir, so that's what this does. The only time we need to multiply anything is in the X11 platform, where we hard-code that each tick is 10 units (to match the previous behavior).

I also phased out the use of discrete libinput scroll. We will add that back at some point, but only when the events actually support it. Continuous scroll *should* always report meaningful values, so converting scroll ticks to the new scale shouldn't be required. As the `TODO` notes, at some point we should actually support discrete scroll as a separate concept.

AlanGriffiths (approved):
LGTM

bors r+

AlanGriffiths:
F33 terminated again, retrying...

bors r+

wmww:
```
/usr/bin/ld: DWARF error: could not find variable specification at offset 1ebb7
```
wat?

AlanGriffiths:
F33 terminated again, retrying...

bors r+

AlanGriffiths:
> wat?

Probably a side effect of the earlier: `2020-12-11T20:53:53.8271044Z c++: fatal error: Killed signal terminated program cc1plus`

AlanGriffiths:
We seem to be having a bad time on F33, but not sure how to investigate.

AlanGriffiths:
Really!!!

bors r+

AlanGriffiths:
Ah! I see now!!!

Likely fix coming up...

AlanGriffiths:
I guess a change from three years ago can bite when someone adds another build to CI.

AlanGriffiths:
@wmww I fixed the persistent build failures on Fedora 33, but then cherry-picked to #1864 as then I saw tests failing:

```plain
2020-12-16T17:55:10.5571225Z The following tests FAILED:
2020-12-16T17:55:10.5571980Z 	227 - mir_unit_tests.LibInputDeviceOnMouse.* (Failed)
2020-12-16T17:55:10.5572978Z 	230 - mir_unit_tests.LibInputDeviceOnTouchpad.* (Failed)
```

tests/mir_test_doubles/mock_libinput.cpp
```diff
@@ -823,6 +823,10 @@ libinput_event* mtd::MockLibInput::setup_axis_event(libinput_device* dev, uint64
         .WillByDefault(Return(vertical));
     ON_CALL(*this, libinput_event_pointer_get_axis_value_discrete(pointer_event, LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL))
         .WillByDefault(Return(horizontal));
+    ON_CALL(*this, libinput_event_pointer_get_axis_value(pointer_event, LIBINPUT_POINTER_AXIS_SCROLL_VERTICAL))
+        .WillByDefault(Return(vertical));
+    ON_CALL(*this, libinput_event_pointer_get_axis_value(pointer_event, LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL))
+        .WillByDefault(Return(horizontal));
```
wmww, line 829: Note that even on discrete events, the non-discrete functions still return values. This matches the libinput implementation

wmww:
I've now fixed the tests and tested manually (all seems to be working well)

PR #1854, opened 2020-12-08
Precise pointer events for Wayland platform

### LINKED ISSUE #1852: High precision input events for the Wayland platform
When running nested under Wayland, we loose high precision input events

wmww:
This adds support for floating point pointer motion and scrolling to the Wayland platform. It appears touches were already precise. Fixes #1852. There's still a notable problem, where scrolling events don't *appear* to be precise because they are arbitrarily multiplied by 10 by the Wayland frontend. This will be fixed in a different PR.

### DIFF
--- a/src/platforms/wayland/display_input.h
+++ b/src/platforms/wayland/display_input.h
@@ -19,7 +19,8 @@
 
 #include <mir/fd.h>
 #include <mir/events/contact_state.h>
-#include <mir/geometry/displacement.h>
+#include <mir/geometry/point_f.h>
+#include <mir/geometry/displacement_f.h>
 
 #include <xkbcommon/xkbcommon.h>
 
@@ -50,17 +51,17 @@ public:
     virtual void pointer_press(
         std::chrono::nanoseconds event_time,
         int button,
-        std::pair<float, float> const& pos,
-        std::pair<float, float> const& scroll) = 0;
+        geometry::PointF const& pos,
+        geometry::DisplacementF const& scroll) = 0;
     virtual void pointer_release(
         std::chrono::nanoseconds event_time,
         int button,
-        std::pair<float, float> const& pos,
-        std::pair<float, float> const& scroll) = 0;
+        geometry::PointF const& pos,
+        geometry::DisplacementF const& scroll) = 0;
     virtual void pointer_motion(
         std::chrono::nanoseconds event_time,
-        std::pair<float, float> const& pos,
-        std::pair<float, float> const& scroll) = 0;
+        geometry::PointF const& pos,
+        geometry::DisplacementF const& scroll) = 0;
 
     PointerInput() = default;
     virtual ~PointerInput() = default;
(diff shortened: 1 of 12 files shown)


src/platforms/wayland/display_input.h
```diff
@@ -47,9 +47,20 @@ class KeyboardInput
...
 {
 public:
-    virtual void pointer_press(std::chrono::nanoseconds event_time, int button, geometry::Point const& pos, geometry::Displacement scroll) = 0;
-    virtual void pointer_release(std::chrono::nanoseconds event_time, int button, geometry::Point const& pos, geometry::Displacement scroll) = 0;
-    virtual void pointer_motion(std::chrono::nanoseconds event_time, geometry::Point const& pos, geometry::Displacement scroll) = 0;
+    virtual void pointer_press(
+        std::chrono::nanoseconds event_time,
+        int button,
+        std::pair<float, float> const& pos,
+        std::pair<float, float> const& scroll) = 0;
```
AlanGriffiths, line 54: We really ought to have strong types for `pos` and `scroll`. Lists of identical types are easy to get in the wrong order and this continues down that "slippery slope".

I know you've mentioned floating point geometry before (#1829) and this seems like an opportunity to do something towards it.

(I'm not convinced that generalizing the current "geometry" classes is the best way forward, but we could think about that.)

AlanGriffiths: Indeed, following the style here, why isn't this just `tuple<float, float, float, float> const& axes` ;)

wmww: > why isn't this just `tuple<float, float, float, float> const& axes` ;)

Because that has substantially more ways to induce bugs? As it currently is, the main way to get is wrong is passing parameters out of order (which you can check because they have meaningful names) or mixing up `.first` and `.second` which again are at least obviously wrong once you notice. A 4-part tuple has all of those problems, plus making it ambiguous (unless explained in comments) what order things are supposed to be in. A more convincing alternative to me would be `float pos_x, float pos_y, float scroll_horz, float scroll_vert`, but I'm not sure if that has any particular advantages over `std::pair`.

Of course you're right that we need a real solution to #1829, but I'm not convinced this needs to be blocked on that. The in-practice maintenance burden here is low.

AlanGriffiths: In part I agree: this doesn't touch any API accessible from downstream.

But if we're passing a generic 'tuple around several interfaces we really ought to have a type. And that internal type could be replaced with a "real solution to #1829".

AlanGriffiths: > a generic 'tuple

In this case pair is just a 2-tuple

wmww:
This now contains a commit from #1961

AlanGriffiths (approved):
Looks sensible, doesn't break anything obvious. Let's have it on trunk.

bors r+

PR #1856, opened 2020-12-10
[gha] let RPi build fail for now

Saviq:


Saviq:
We crossed commits, it's already in master.

PR #1858, opened 2020-12-10
Floating point geometry

wmww:
Will address #1829 when complete. Adds a new `geometry::generic::Wrapper` class and changes the ABI-stable `IntWrapper` to inherit from it.

Only minor issue is that to be generic, operations need to return `generic::Wrapper<Tag, int>` instead of `IntWrapper<Tag>`. The two can be implicitly converted between, but this can cause a problem in some cases. For example, if `std::max()` is called on an `IntWrapper` and a `generic::Wrapper`, it needs an explicit type to cast to.

When we break ABI, I would like to drop `IntWrapper` and make `Width`, `Height`, `X`, etc typedefs of the `generic::Type<int>`. This would fix those problems.


include/core/mir/geometry/dimensions.h
```diff
@@ -29,180 +30,43 @@ namespace mir
 /// and the operations that they support.
 namespace geometry
 {
-
 namespace detail
 {
+// TODO: On next ABI break, switch to just using Wrapper
 template<typename Tag>
-class IntWrapper
+class IntWrapper : public generic::detail::Wrapper<Tag, int>
```
AlanGriffiths, line 37: Why "public"? Do we really want enable conversions to base?

include/core/mir/geometry/dimensions.h
```diff
@@ -29,180 +30,43 @@ namespace mir
...
-        return (uint32_t)value;
     }
 
-    constexpr int as_int() const
+    template<typename U>
+    explicit constexpr IntWrapper(U const& value)
+        : generic::detail::Wrapper<Tag, int>{value}
     {
-        return value;
     }
```
AlanGriffiths, line 52:
```C++
    using generic::detail::Wrapper<Tag, int>::generic::detail::Wrapper<Tag, int>;
```

include/core/mir/geometry/dimensions.h
```diff
@@ -29,180 +30,43 @@ namespace mir
...
-
-// Dividing by a scaler value is fine
-template<typename Scalar>
-inline constexpr Width operator/(Width const& w, Scalar scale) { return Width{w.as_int() / scale}; }
-template<typename Scalar>
-inline constexpr Height operator/(Height const& h, Scalar scale) { return Height{h.as_int() / scale}; }
-template<typename Scalar>
-inline constexpr DeltaX operator/(DeltaX const& dx, Scalar scale) { return DeltaX{dx.as_int() / scale}; }
-template<typename Scalar>
-inline constexpr DeltaY operator/(DeltaY const& dy, Scalar scale) { return DeltaY{dy.as_int() / scale}; }
```
AlanGriffiths, line 188: The generic operators proposed do not replace these (they return the generic type not the derived type). Either the generic operators need to return the correct type, or these need to be reimplemented in terms of the generic operators.

AlanGriffiths (commented):
I know this is just draft, but I took a look anyway.

include/core/mir/geometry/dimensions.h
wmww: Yes, but they are implicitly and convertible to the the derived types at zero runtime cost, so the results turn into the derived types whenever we need them to. If we assign them to an `auto` variable they don't, but that's fine because the generic types can be used the same way. The only exception is when trying to use `std::min`/`max` with both generic and derived, but that only showed up a few places in the code base.

include/core/mir/geometry/dimensions.h
wmww: Yes

include/core/mir/geometry/dimensions.h
wmww: I'm not sure what this means :/

wmww:
This has now been significantly updated. The correct types are now returned. Will go through and make the rest of the types generic once I get a LGTM on what's already there

wmww:
Maybe we should pull the concept of a tagged value out of geometry? I'm coming across places in the Wayland frontend where I need it for other things (like making sure serials and touch IDs don't get mixed up). I know we have a non-geometry `IntWrapper` somewhere, but these things are `uint32_t`s, not `int`s.

Proposal: outside of geometry somewhere we have `Value<T>::Wrapper<Tag>` and then inside of geometry we define the geometry tags and the behavior of wrappers that use them. We can then make any non-geometry `IntWrapper` classes inherit from `Value<int>::Wrapper<Tag>` and use that directly in places that don't need to be ABI-stable.

include/core/mir/geometry/point_generic.h
```diff
@@ -0,0 +1,94 @@
...
+
+template<typename P, typename std::enable_if<std::is_base_of<detail::PointBase, P>::value, bool>::type = true>
+inline P& operator+=(P& lhs, typename P::template WrapperType<DeltaXTag> rhs) { lhs.x += rhs; return lhs; }
+template<typename P, typename std::enable_if<std::is_base_of<detail::PointBase, P>::value, bool>::type = true>
+inline P& operator+=(P& lhs, typename P::template WrapperType<DeltaYTag> rhs) { lhs.y += rhs; return lhs; }
+
+template<typename P, typename std::enable_if<std::is_base_of<detail::PointBase, P>::value, bool>::type = true>
+inline P& operator-=(P& lhs, typename P::template WrapperType<DeltaXTag> rhs) { lhs.x -= rhs; return lhs; }
+template<typename P, typename std::enable_if<std::is_base_of<detail::PointBase, P>::value, bool>::type = true>
+inline P& operator-=(P& lhs, typename P::template WrapperType<DeltaYTag> rhs) { lhs.y -= rhs; return lhs; }
```
AlanGriffiths, line 81: This use of `typename P::template WrapperType<Foo>` is an "enable_if" alternative?

AlanGriffiths (changes requested):
Not finished reviewing, but there are some obvious simplifications to be had.

include/core/mir/geometry/point_generic.h
wmww: I guess?? I'm not completely sure how this would be replaced by `enable_if`, but there's probably a way.

include/core/mir/geometry/dimensions.h
```diff
@@ -19,190 +19,55 @@
...
-    template<typename AnyInteger>
-    explicit constexpr IntWrapper(AnyInteger value) : value(static_cast<ValueType>(value)) {}
-
-    constexpr uint32_t as_uint32_t() const  // TODO: Deprecate this later
+    template<typename U>
+    constexpr IntWrapper(typename generic::Value<U>::template Wrapper<Tag> const& value)
+        : generic::Value<int>::Wrapper<Tag>{value}
     {
-        return (uint32_t)value;
     }
```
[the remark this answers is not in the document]
wmww: Nice!

include/core/mir/geometry/point.h
```diff
@@ -20,50 +20,25 @@
...
     constexpr Point() = default;
     constexpr Point(Point const&) = default;
     Point& operator=(Point const&) = default;
 
     template<typename XType, typename YType>
-    constexpr Point(XType&& x, YType&& y) : x(x), y(y) {}
-
-    X x;
-    Y y;
+    constexpr Point(XType&& x, YType&& y) : generic::Point<detail::IntWrapper>{x, y} {}
```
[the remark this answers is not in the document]
wmww: Nice!

wmww:
I've now switched all the classes over to be generic and have float versions (except for the multi-rectangle container, which isn't as widely used and I don't think is needed at this time).

include/core/mir/geometry/point_generic.h
AlanGriffiths: NM On looking closer I see what's going on.

include/core/mir/geometry/dimensions_generic.h
```diff
@@ -0,0 +1,290 @@
...
+    /// these restrictions (see the as_*() functions). int values (which are most values) should use the
+    /// derived IntWrapper class, and other types should use this directly.
+    template<typename Tag>
+    struct Wrapper: detail::ValueWrapperBase
+    {
+        using ValueType = T;
+        using TagType = Tag;
+        template<typename OtherTag>
+        using WrapperType = Wrapper<OtherTag>;
+        using IsGemetryWrapper = void; ///< Used to check if this is a geometry wrapper type
```
AlanGriffiths, line 65: I don't think this is used, and it is definitely wrongly spelt.

include/core/mir/geometry/dimensions_generic.h
```diff
@@ -0,0 +1,290 @@
...
+    /// width to a height or adding two x positions together). Of course, explicit casts are possible to get around
+    /// these restrictions (see the as_*() functions). int values (which are most values) should use the
+    /// derived IntWrapper class, and other types should use this directly.
+    template<typename Tag>
+    struct Wrapper: detail::ValueWrapperBase
+    {
+        using ValueType = T;
+        using TagType = Tag;
+        template<typename OtherTag>
+        using WrapperType = Wrapper<OtherTag>;
```
AlanGriffiths, line 64: On reflection, I think the biggest problem I had following this was the name and usage of `WrapperType`. It is a metafunction for creating a _WholeValue_ type based on the same underlying value type and applying a supplied tag.

I would probably have written something using the TMP style, like:
```c++
template<typename OtherTag>
struct Corresponding { using type = Wrapper<OtherTag>; };
```
With the usage:
```c++
template<typename P, typename std::enable_if<std::is_base_of<detail::PointBase, P>::value, bool>::type = true>
inline constexpr P operator+(P lhs, typename P::template Corresponding<DeltaXTag>::type rhs) { return{lhs.x + rhs, lhs.y}; }
```
I'm not sure that my preference is enough for a "request changes", but `Corresponding<DeltaXTag>::type` makes more sense to me than "WrapperType<DeltaXTag>".

AlanGriffiths (changes requested):
Does what it should, but at a minimum `IsGemetryWrapper` needs fixing.

include/core/mir/geometry/dimensions_generic.h
wmww: lol

include/core/mir/geometry/dimensions_generic.h
wmww: That sounds like a better idea for the naming. I'll keep that in mind when I make another pass over this.

wmww:
Ok, finally updated

include/core/mir/geometry/dimensions_generic.h
wmww: Dropped

include/core/mir/geometry/point_generic.h
```diff
@@ -70,24 +70,24 @@ inline constexpr bool operator != (P const& lhs, P const& rhs)
 }
 
 template<typename P, typename std::enable_if<std::is_base_of<detail::PointBase, P>::value, bool>::type = true>
-inline constexpr P operator+(P lhs, typename P::template WrapperType<DeltaXTag> rhs) { return{lhs.x + rhs, lhs.y}; }
+inline constexpr P operator+(P lhs, typename P::template Corresponding<DeltaXTag> rhs) { return{lhs.x + rhs, lhs.y}; }
```
AlanGriffiths, line 73: This repeated "`typename P::template Corresponding<DeltaXTag>`" incantation is annoying. It could be factored out with something like:

```C++
template<class GeometricType, typename Tag>
using corresponding = typename GeometricType::template Corresponding<Tag>;

inline constexpr P operator+(P lhs, corresponding<P, DeltaXTag> rhs) { return{lhs.x + rhs, lhs.y}; }
```

AlanGriffiths: Or even

```C++
template<class GeometricType, typename Tag, typename std::enable_if<std::is_base_of<detail::PointBase, GeometricType>::value, bool>::type = true>
using Convert= typename GeometricType::template Corresponding<Tag>;

template<typename Tag>
using ToCorresponding = Tag;

template<typename P>
inline constexpr P operator+(P lhs, Convert<P, ToCorresponding<DeltaXTag>> rhs) { return{lhs.x + rhs, lhs.y}; }
```

AlanGriffiths (approved):
This is an improvement, so happy to see it land. But there's still a bit that feels awkward and I've made a couple of alternative suggestions.

bors delegate=wmww

wmww:
Since it's a type, I assume it should be capitalized?

AlanGriffiths (approved):
bors r+

PR #1860, opened 2020-12-14
Print help text to stdout instead of stderr

### LINKED ISSUE #1043: Mir help text printed to stderr
When `-h` or `--help` is used, the help text is printed to stderr instead of stdout. This is a problem from a usability standpoint (it makes it harder to grep for arguments) and a compatibility standpoint (everything else prints help text to stdout) and a theoretical standpoint (the help text is the output of the command, not a side effect).

wmww:
Fixes #1043

### DIFF
--- a/include/core/mir/abnormal_exit.h
+++ b/include/core/mir/abnormal_exit.h
@@ -24,11 +24,22 @@
 
 namespace mir
 {
+/// Indicates Mir should exit with an error message printed to stderr
 class AbnormalExit : public std::runtime_error
 {
 public:
     AbnormalExit(std::string const& what) :
         std::runtime_error(what)
+    {
+    }
+};
+
+/// Indicates Mir should exit with the given output (such as help text) printed to stdout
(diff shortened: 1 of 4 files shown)


AlanGriffiths (approved):
Reluctant approve. Reasons I don't like this:

1. Any actions related to exceptions ought to be taken in the "handler" not the "raiser"; and,
2. An empty string is not what the "handler" expects (nor should expect). That leads to a spurious blank line on stderr.

OTOH wiring through a `HelpExit` exception, or extending `AbnormalExit` to indicate the type of information carried is more work.

So, I won't bors yet and will allow someone else to have that honor.

wmww:
You're right of course, and it turned out it wasn't all that much trouble. I made an `ExitWithOutput` subclass of `AbnormalExit`. This way handlers that don't care about printing it to the right stream need only catch the one exception type.

AlanGriffiths (approved):
WFM

bors r+

PR #1861, opened 2020-12-15
Replace `exit()` with a signal-safe function.

AlanGriffiths:
Replace `exit()` with a signal-safe function.

See, for example, https://man7.org/linux/man-pages/man7/signal-safety.7.html


Saviq (approved):
LGTM

bors merge

PR #1863, opened 2020-12-16
Fix FTBFS on s390x

AlanGriffiths:
Fix FTBFS on s390x. ("Borrowed" code gets a benign compiler warning)

https://bugs.launchpad.net/ubuntu/+source/mir/+bug/1889166


AlanGriffiths:
Don' t land until we confirm this is the direction taken with the archive package. (See linked bug)

RAOF (approved):
Yup. This is what I'm doing in the archive packages.

(or, at least, is morally equivalent)
bors r+

PR #1864, opened 2020-12-16
Fix spread options on Fedora

AlanGriffiths:
Looks like CMake option `MIR_USE_LD_GOLD` has been replaced, but the usage wasn't updated.


AlanGriffiths:
Fix has landed

PR #1865, opened 2020-12-16
Move destroy listener functionality to LifetimeTracker

wmww:
Previously we had a mix of ways to be notified when a Wayland object was destroyed including a destroy listener system on `WlSurface` and a callback argument to the constructor of input device objects. This standardizes the process of adding and removing destroy listeners and makes it available to any wayland object or other `LifetimeTracker` subclass. This also fixes a dirty hack where we were generating bad pointers to use as IDs in `WlTouch`.

### DIFF
--- a/src/server/frontend_wayland/wl_seat.cpp
+++ b/src/server/frontend_wayland/wl_seat.cpp
@@ -287,14 +287,13 @@ void mf::WlSeat::Instance::get_keyboard(wl_resource* new_keyboard)
 
 void mf::WlSeat::Instance::get_touch(wl_resource* new_touch)
 {
-    seat->touch_listeners->register_listener(
-        client,
-        new WlTouch{
-            new_touch,
-            [listeners = seat->touch_listeners, client = client](WlTouch* listener)
-            {
-                listeners->unregister_listener(client, listener);
-            }});
+    auto const touch = new WlTouch{new_touch};
+    seat->touch_listeners->register_listener(client, touch);
+    touch->add_destroy_listener(
+        [listeners = seat->touch_listeners, listener = touch, client = client]()
+        {
+            listeners->unregister_listener(client, listener);
+        });
 }
 
 void mf::WlSeat::Instance::release()
--- a/src/server/frontend_wayland/wl_seat.cpp
+++ b/src/server/frontend_wayland/wl_seat.cpp
@@ -236,14 +236,13 @@ mf::WlSeat::Instance::Instance(wl_resource* new_resource, mf::WlSeat* seat)
 
 void mf::WlSeat::Instance::get_pointer(wl_resource* new_pointer)
 {
-    seat->pointer_listeners->register_listener(
-        client,
-        new WlPointer{
-            new_pointer,
-            [listeners = seat->pointer_listeners, client = client](WlPointer* listener)
-            {
-                listeners->unregister_listener(client, listener);
-            }});
+    auto const pointer = new WlPointer{new_pointer};
+    seat->pointer_listeners->register_listener(client, pointer);
+    pointer->add_destroy_listener(
+        [listeners = seat->pointer_listeners, listener = pointer, client = client]()
+        {
+            listeners->unregister_listener(client, listener);
+        });
 }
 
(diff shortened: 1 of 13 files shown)


src/server/frontend_wayland/wl_seat.cpp
```diff
@@ -236,65 +236,64 @@ mf::WlSeat::Instance::Instance(wl_resource* new_resource, mf::WlSeat* seat)
...
+        [listeners = seat->pointer_listeners, listener = pointer, client = client]()
+        {
+            listeners->unregister_listener(client, listener);
+        });
 }
 
 void mf::WlSeat::Instance::get_keyboard(wl_resource* new_keyboard)
 {
-    seat->keyboard_listeners->register_listener(
-        client,
```
wmww, line 252: This looks like a mess, but all that's changing is I'm moving object creation and the destroy listener out of the function call (like for pointer and touch). Diff is messy because big lambda is changing indentation.

tests/unit-tests/wayland/test_lifetime_tracker.cpp
```diff
@@ -0,0 +1,117 @@
...
+}
+
+TEST(LifetimeTrackerTest, destroy_listener_is_called)
+{
+    int call_count = 0;
+    {
+        MockTracker tracker;
+        tracker.add_destroy_listener([&](){ call_count++; });
+    }
+    EXPECT_THAT(call_count, Eq(1));
```
AlanGriffiths, line 52: Maybe it would be more expressive to use a mock for a listener and have something like:

```c++
    EXPECT_CALL(mocklistener, callback()).Times(1);
```

AlanGriffiths (changes requested):
I'm open to persuasion, but I think the tests could be written in a more fluent style. (Using `EXPECT_CALL...Invoke()`)

tests/unit-tests/wayland/test_lifetime_tracker.cpp
wmww: Done

tests/unit-tests/wayland/test_lifetime_tracker.cpp
```diff
@@ -0,0 +1,112 @@
...
+    tracker.add_destroy_listener([&](){ listener.callback(); });
+}
+
+TEST(LifetimeTrackerTest, destroy_listener_only_called_once_when_marked_as_destroyed)
+{
+    MockListener listener;
+    EXPECT_CALL(listener, callback()).Times(1);
+    MockTracker tracker;
+    tracker.add_destroy_listener([&](){ listener.callback(); });
+    tracker.mark_destroyed();
```
AlanGriffiths, line 64: Well, I'd avoid conflating the setup and expectations:
```c++
class MockTracker ...
{
    MockListener listener;
    shared_ptr<MockTracker> tracker = make_shared<MockTracker>;

    void SetUp()
    {
        tracker->add_destroy_listener([&](){ listener.callback(); });
    }

    void TearDown()
    {
        tracker.reset();
        Mock::VerifyAndClearExpectations(&listener);
    }
...

TEST(LifetimeTrackerTest, destroy_listener_only_called_once_when_marked_as_destroyed)
{
    EXPECT_CALL(listener, callback()).Times(1);

    tracker.mark_destroyed();
}

TEST(LifetimeTrackerTest, is_still_alive_when_destroy_listener_called)
{
    mw::Weak<MockTracker> weak{&tracker};

    EXPECT_CALL(listener, callback()).Times(1)
        .WillOnce(InvokeWithoutArgs([&]{ EXPECT_THAT(weak, IsTrue()); }));
}
```

wmww: Moved tracker and listener declarations to a test class, but I don't consider adding the listener to be part of the setup (some tests require multiple listeners to be added, some require holding on to the returned ID, etc. More obvious to just have each test add it itself however it needs.

AlanGriffiths (approved):
Good!

bors r+

wmww:
bors r=AlanGriffiths

PR #1866, opened 2020-12-17
DisplayConfigurationTest: Standardise on 60s timeouts.

RAOF:
The Launchpad builders inconsistently, but fairly regularly, fail these
tests. When this happens it looks like the tests have taken a long time,
so I suspect that they're timing out due to the builders being slow.

Just wait for up to 60s *everywhere*, to give the poor overworked builders
time to hit the deadlines!


AlanGriffiths (approved):
Harmless when things are working.

bors r+

PR #1867, opened 2020-12-17
Add (incomplete) pointer confinement

AlanGriffiths:
Add (incomplete) pointer confinement. Notably, there is currently no support for confinement regions.

As this is incomplete, pointer confinement is disabled by default, to enable:

    `--add-wayland-extensions zwp_pointer_constraints_v1:zwp_relative_pointer_manager_v1`


AlanGriffiths:
This branch has been "parked" for a while, and it looks like it won't be touched for a further while. After discussion the team agreed it is better to land it behind a feature flag than let it suffer "code rot".

The current implementation is sufficient for various SDL2 based games: in particular, mir-kiosk-neverputt works far better (without the invisible cursor "hitting the screen edged").

src/server/frontend_wayland/wayland_input_dispatcher.cpp
```diff
@@ -222,28 +222,34 @@ void mf::WaylandInputDispatcher::handle_pointer_motion_event(
...
+        mir_pointer_event_axis_value(event, mir_pointer_axis_relative_x),
+        mir_pointer_event_axis_value(event, mir_pointer_axis_relative_y)};
+
+    bool const send_position = (!last_pointer_position || position != last_pointer_position.value());
     bool const send_axis = (axis_motion.first || axis_motion.second);
+    bool const send_motion = (motion != geom::Displacement{});
 
     last_pointer_position = position;
 
-    if (send_motion || send_axis)
```
wmww, line 230: No need to drop this (probably it was added since you forked off). Obviously it will need to be updated to include `send_position`.

wmww (changes requested):
As expected, it's not ideal for merging but cleaning it up later might be easier than letting it rot out of tree. I'm good with landing once the formatting and floating point input issues are fixed.

wmww (approved):
Good enough
bors r+

PR #1869, opened 2020-12-18
Drop BasicWaylandSurfaceObserver

### LINKED ISSUE #1868: BasicWaylandSurfaceObserver is not threadsafe
`destroyed` is set on not-the-wayland thread, and there's no locking. When `run_on_wayland_thread_unless_destroyed()` is running the distructor should block until it's finished. Or maybe the whole thing needs a redesign.

wmww:
Fixes #1868. `BasicWaylandSurfaceObserver` (and previously `WaylandSurfaceObserver`, which it was broken off from) were not threadsafe. With this PR instead of trying to determine if the observer is still alive (which would require blocking destruction to do properly) we instead capture whatever we need. `wayland::Weak` is used to access Wayland objects. `BasicWaylandSurfaceObserver` is dropped, as there is no obvious clean way to abstract the requirements of all three observers.


AlanGriffiths (approved):
LGTM

bors r+

PR #1872, opened 2021-01-03
Fix more missing #includes for GCC 11

Conan-Kudo:
This is from an effort in Fedora to prepare for the GCC 11 rebase for Fedora 34.

Followup to #1771.

### DIFF
--- a/src/server/input/default_input_manager.cpp
+++ b/src/server/input/default_input_manager.cpp
@@ -29,6 +29,7 @@
 #include "mir/terminate_with_current_exception.h"
 
 #include <future>
+#include <memory>
 
 namespace mi = mir::input;
 
--- a/tests/unit-tests/input/test_default_input_manager.cpp
+++ b/tests/unit-tests/input/test_default_input_manager.cpp
@@ -33,6 +33,7 @@
 #include <gmock/gmock.h>
 #include <chrono>
 #include <list>
+#include <memory>
 
 namespace mt = mir::test;
 namespace md = mir::dispatch;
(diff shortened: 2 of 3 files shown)


src/server/input/default_input_manager.cpp
```diff
@@ -16,6 +16,8 @@
  * Authored by: Andreas Pokorny <andreas.pokorny@canonical.com>
  */
 
+#include <memory>
+
 #include "default_input_manager.h"
```
AlanGriffiths, line 21: We have a policy of including the "corresponding" header first (in this case `#include "default_input_manager.h"` should be the first include).

Please put `#include <memory>` after `#include <future>`

tests/unit-tests/input/test_default_input_manager.cpp
```diff
@@ -16,6 +16,7 @@
  * Authored by: Andreas Pokorny <andreas.pokorny@canonical.com>
  */
 
+#include <memory>
 #include "src/server/input/default_input_manager.h"
```
AlanGriffiths, line 20: Please put `#include <memory>` after `#include <list>`, not before the project headers.

AlanGriffiths (changes requested):
Thanks for this. Could you please re-order these includes to be more consistent with the project style.

Conan-Kudo:
@AlanGriffiths I believe I've addressed your feedback now.

AlanGriffiths (approved):
LGTM. Thanks!

bors r+

Issue #1875, opened 2021-01-04
Log more information about outputs


Saviq:
When there's an output connected, we log something along those lines:
```plain
<information> mirserver: Selected driver: mir:gbm-kms (version 2.2.0)
< -warning- > gbm-kms: Failed to detect whether device /dev/dri/card0 supports KMS, but continuing anyway
<information> gbm-kms: Using DRM device /dev/dri/card0
< - debug - > gbm-kms: No EDID data available on connector 54
< -warning- > gbm-kms: Unable to determine the current display mode.
<information> gbm-kms: DRM device details:
<information> gbm-kms: /dev/dri/card0: using driver vc4 [Broadcom VC4 graphics] (version: 0.0.0 driver date: 20140616)
<information> gbm-kms: 	Output: DSI-1 (connected)
<information> gbm-kms: 		Mode: 800×480@60.04
< -warning- > gbm-kms: drmModeCrtcSetGamma failed: Function not implemented
<information> gbm-kms: Detected single-GPU DisplayBuffer. Rendering will be sent directly to output
```

But when outputs are missing (I believe), we're just getting this:
```plain
<information> mirserver: Selected driver: mir:gbm-kms (version 2.2.0)
< -warning- > gbm-kms: Failed to detect whether device /dev/dri/card0 supports KMS, but continuing anyway
<information> gbm-kms: Using DRM device /dev/dri/card0
< -warning- > gbm-kms: Unable to determine the current display mode.
```

Could use more details about _why_.

RAOF:
Is that all of the log? I'd expect *at least* a
```
<information> gbm-kms: DRM device details:
```
line after that. The “Unable to determine the current display mode” message is only output if there's an actual DRM connector we've failed to find the mode for.

Saviq:
> Is that all of the log? I'd expect _at least_ a

It was the last message, yes:

https://github.com/MirServer/mir/issues/1875#issuecomment-754275546

RAOF:
Ah! I *think* that must mean that it's *crashed* shortly thereafter, rather than a matter of insufficient logging!

AlanGriffiths:
I don't think this is a current problem

PR #1876, opened 2021-01-05
gbm-kms: Don't warn about being unable to find the current mode of disconnected outputs

RAOF:


RAOF:
Identified from brief investigation of #1875, but not in any way a fix.

wmww (approved):
Sure
bors r+

PR #1879, opened 2021-01-06
[miral-kiosk] Allow wayland extensions to be configured in kiosks

AlanGriffiths:
[miral-kiosk] Allow wayland extensions to be configured in kiosks (and enable pointer confinement by default).

Allowing Wayland extensions to be configured is not very important for kiosks but "nice to have".

However, the incomplete "pointer confinement" extension works well for games in kiosk mode and is worth enabling by default.

Try, for example, mir-kiosk-neverputt.


wmww (approved):
LGTM
bors r+

Issue #1880, opened 2021-01-07
ObserverMultiplexer: deadlocks if observer removed from callback when executor runs immediately


wmww:
`ObserverRegistrar::register_interest()` is documented as such:

> If the thread calling is currently in an Observer call that call will complete, no further ones will be dispatched.

And that is indeed true and tested if the executor in use delegates the callback to another thread or later time. However, if the executor runs `work` immediately, the multiplexer deadlocks. I'm running into this problem because the Wayland executor runs work immediately if already on the Wayland thread. It's possible for me to work around this issue if fixing it would be too difficult or introduce significant tradeoffs.

I've written a test to demonstrate the problem, which I will PR shortly.

EDIT: in the current thing I'm working on not unregistering from a callback actually simplifies the code, so this is no longer a pressing issue.

RAOF:
Yeah, I was a little worried when doing the immediate-execute Wayland executor bit, because it breaks the expectations of `mir::Executor`.

We should perhaps adopt the more precise vocabulary of one of the later Executor proposals; specifically, separating `spawn()` into `dispatch()` (which runs as soon as possible) and `post()` (which definitely does not run in the current stack frame).

wmww:
There's other places where multiplexers use an immediate executor out of convenience. It would be nice to just remove the constraint. I've updated my PR to attempt to do that.

RAOF:
Hm. Where else do multiplexers use immediate executors? Maybe we should have a `mir::system_executor` available in global scope to make this more convenient?

wmww:
`ms::SessionManager::SessionObservers` in `session_manager.cpp` implements `Executor` using immediate dispatch, and uses itself as the multiplexer's default executor. Perhaps we should make the multiplexer do this and provide a default constructor that uses itself?

wmww:
@RAOF isn't this fixed?

wmww:
I'm fairly sure this has been fixed for some time

PR #1881, opened 2021-01-07
Allow eager execution in observer executor

wmww:
Failing tests for #1880. No attempted fix yet. Might be of interest to @RAOF. Not blocking anything at this time.


wmww:
This PR now has a proposed fix. The added costs are an observer vector allocation for each observation dispatched and a 2nd locking of each observer (to guarantee the executor is still valid).

src/include/server/mir/observer_multiplexer.h
```diff
@@ -199,17 +211,24 @@ void ObserverMultiplexer<Observer>::for_each_observer(MemberFn f, Args&&... args
...
-        observer_pair.first.spawn(
-            [invokable_mem_fn, weak_observer = observer_pair.second, args...]() mutable
-            {
-                if (auto observer = weak_observer->lock())
+        std::lock_guard<decltype(observer_mutex)> lock{observer_mutex};
+        local_observers = observers;
+    }
+    for (auto& weak_observer: local_observers)
+    {
+        if (auto observer = weak_observer->lock())
```
RAOF, line 221: I've not written a test (yet) but I'm *pretty* sure that this no longer upholds the “once `unregister_interest()` returns there is guaranteed to be no thread in or entering this Observer”.

If thread *A* is currently in `for_each_observer()` and the dispatch calls `unregister_interest(foo)` while thread *B* is also in `for_each_observer()` in the `for(... : local_observers)` loop, then thread *B* can hit `if (auto observer = weak_observer->lock())` with `*weak_observer == foo`.

At this point, thread *B* will block waiting to acquire the lock until thread *A* finishes the `unregister_interest(foo)` call.

Now *B* will *successfully* acquire a lock of `weak_observer`, because the `lock()` call has kept a strong reference to the observer around, and `observer.spawn()` will (if the executor is an immediate executor) now call into the Observer `foo`.

I *think* this could be fixed by switching the locking around in `WeakObserver::lock()`, but this is an annoyingly subtle area of code.

RAOF (changes requested):
I've not verified that the flaw I've found is the *only* flaw, but I'm pretty sure it is one.

It would probably be simpler to split `spawn()` into `dispatch()` and `post()`, and then use `post()` here :)

src/include/server/mir/observer_multiplexer.h
wmww: Yes, there were a few functions where `WeakObserver`'s observer wasn't being protected by the mutex. In the updated PR I lock the mutex earlier in `WeakObserver::lock()` (as you suggested), and I also do the reset process in `WeakObserver::maybe_reset()` (so it can keep the mutex locked for the duration).

AlanGriffiths:
> It would probably be simpler to split `spawn()` into `dispatch()` and `post()`, and then use `post()` here :)

I think WG21 has been down this road too. They also came to this conclusion. "Copying their homework" is probably a good approach.

AlanGriffiths:
Do we need this? I.e. does this solve a problem? Does the approach suggested by both reviewers also solve it?

wmww:
It fixes #1880, which I think would be a good thing to fix. We can work around this constraint, but removing it may allow code elsewhere to be simpler and easier to write correctly. Of course this does have some performance considerations, including copying the observers vector. I suspect this isn't significant but I haven't profiled or anything.

I'd need to dive back into this to understand what splitting `spawn()` into `dispatch()` and `post()` means and how I would do that. I'm not sure what problem that would solve though, since this PR in it's current state (AFAIK) passes the tests and is correct and safe. The only specific problem that was pointed out in review is fixed.

wmww:
Hm `SurfaceInputCombinations` and other WLCS tests seem to be failing intermittently. I wonder if there is a problem with this PR, or if it's unrelated. Perhaps I'll look into it tomorrow.

AlanGriffiths:
I find the wlcs tests to be failing consistently. This PR breaks something (or at least causes some reordering the tests detect), possibly in the "event faking".

Looking at the code I didn't see an obvious cause but, unfortunately, I also found that the code is not "obviously correct".

wmww:
I will figure out what's going on here at some point, but it's not top of my priority list right now

wmww:
It looks like some nasty interaction was going on between the dispatcher and WLCS's Wayland executor. Hopefully #2009 sorts that out. This PR now includes those commits.

AlanGriffiths (approved):
I still have the feeling this is closer to "no obvious problems" than "obviously no problems", but there are no specific blockers.

@RAOF I'll leave it to you to take a fresh look and bors. (Or not.)

RAOF (approved):
I still think we should make immediate execution an opt-in, but that (and other improvements) can be made later. This is an improvement over the status quo.

bors r+

PR #1882, opened 2021-01-07
Clipboard singleton

wmww:
This adds a new `Clipboard` object to the scene, and ports our implementation of `DataDevice` (and related Wayland objects) to use it. This makes the code more modular, and more importantly will enable implementing copy/paste between Wayland and XWayland.


AlanGriffiths:
A couple of terminology issues: you are neither "adding a _Singleton_" nor "adding a _global_", both of those terms describe something different than what happens here.

This `Clipboard` may be unique to a _ContextObject_ (in this case a `ServerConfiguration`), but multiple instances can exist in a process (i.e. it is not a _Singleton_) and it is not a _global_ because is not accessed through a variable at namespace scope.

You might be aware of this, but others might interpret your words literally.

https://wiki.c2.com/?SingletonPattern
https://wiki.c2.com/?GlobalVariable
https://wiki.c2.com/?ContextObject

AlanGriffiths (approved):
WFM

bors r+

PR #1883, opened 2021-01-07
Delete old workaround

AlanGriffiths:
The LTTNG pkg-config file has been fixed since at least 16.04. I.E. we don't need a replacement any more.


wmww (approved):
LGTM
bors r+

PR #1886, opened 2021-01-13
Use the Wayland executor directly in more places

wmww:
This removes `WlSeat::spawn()` and switches all of its users over to using the Wayland executor directly. There was never any reason for running on the Wayland thread to be tied to the seat. In fact, that's the only thing the seat even had an executor for. It also switches around the XWayland frontend a bit to use executors directly where it makes sense.


AlanGriffiths (approved):
Some of these parameter lists are looking alarmingly long (like we're missing an abstraction somewhere).

But this looks sensible, and we can invent new abstractions in another pass.

bors r+

PR #1887, opened 2021-01-13
gbm-kms: Correctly handle failures when resolving acquire_device future

RAOF:
`ConsoleServices::acquire_device` returns a future that will resolve to a `Device` handle. Should opening the device fail this will resolve to an exception instead, so error handling needs to use `try {} catch()`.


wmww (approved):
LGTM
bors r+

Issue #1792, opened 2020-11-04
BasicWindowManager does not always use correct input event for move/resize


wmww:
While investigating #760, I noticed that when initiating a move/resize gesture we were blindly using the most recent input event without checking it against the serial from the request. This means that if the user drags the titlebar then touches the window and the `move` request from the titlebar doesn't come in until after the touch, the user will end up moving the window with the touch instead of the click.

wmww:
All the functions in `BasicWindowManager` that take timestamps probably need to be changed to take an input event and all the instances of
```
if (timestamp >= last_input_event_timestamp)
```
probably need to be dropped. Consider the following sequence:

| Server | Client |
|--:|:--|
| pointer pressed `serial=1` ⬎ | |
| pointer motion ⬎ | ↳ pointer pressed `serial=1` |
| | ↳ pointer motion |
| touch down `serial=2` ⬎ | ⬐ window move `serial=1` |
| window move `serial=1` ↲ | ↳ touch down `serial=2` |

I believe this currently initiates an interactive move from the touch, which is obviously wrong (the touch may not have even been over the titlebar). I think the correct behavior is for the interactive pointer move to precede uninterrupted. The correct timestamp of the move request would be that of the first pointer press (before both the touch and the pointer motion), which would indicate that requests with the non-latest timestamp should not be discarded.

AlanGriffiths:
In part this is a legacy of bolting on the Wayland frontend: the timestamps were (are?) correctly with respect to the mirclient API: receiving a request tagged with anything other than the latest input event should result in the request being ignored.

The Wayland frontend supplies the latest input timestamp into this without any checking of serial numbers. Maybe this can be improved there, or maybe we should rework the whole thing and rip out any affected mirclient code.

wmww:
In the case of a window move request, the "latest timestamp" is almost always going to be wrong because the request wont be send until there is pointer motion after the button press, but the serial will be for the press because motion doesn't have a serial. It could maybe work if we made it latest-event-with-a-serial, but IMO that's not the *right* way to do it, and leads to undesirable behavior in edge cases (like described in the chart).

wmww:
[...]

PR #1895, opened 2021-01-21
Process events in device classes

wmww:
Lot of code churn, but the logic stays roughly the same. This sends the Mir events into `WlKeyboard`/`WlPointer`/`WlTouch` instead of extracting information from them. This is step in the direction of addressing #1792. We want to attach serials to input events, and this moves the events to where we'll need them to do that.

### DIFF
--- a/src/server/frontend_wayland/wayland_input_dispatcher.h
+++ b/src/server/frontend_wayland/wayland_input_dispatcher.h
@@ -66,15 +66,10 @@ private:
     wayland::Weak<WlSurface> const wl_surface;
 
     std::chrono::nanoseconds timestamp{0};
-    MirPointerButtons last_pointer_buttons{0};
-    std::experimental::optional<std::pair<float, float>> last_pointer_position;
 
     /// Handle user input events
     ///@{
     void handle_keyboard_event(std::chrono::milliseconds const& ms, MirKeyboardEvent const* event);
-    void handle_pointer_event(std::chrono::milliseconds const& ms, MirPointerEvent const* event);
-    void handle_pointer_button_event(std::chrono::milliseconds const& ms, MirPointerEvent const* event);
-    void handle_pointer_motion_event(std::chrono::milliseconds const& ms, MirPointerEvent const* event);
     void handle_touch_event(std::chrono::milliseconds const& ms, MirTouchEvent const* event);
     ///@}
 };
--- a/src/server/frontend_wayland/wayland_input_dispatcher.h
+++ b/src/server/frontend_wayland/wayland_input_dispatcher.h
@@ -70,7 +70,6 @@ private:
     /// Handle user input events
     ///@{
     void handle_keyboard_event(std::chrono::milliseconds const& ms, MirKeyboardEvent const* event);
-    void handle_touch_event(std::chrono::milliseconds const& ms, MirTouchEvent const* event);
     ///@}
 };
 }
--- a/src/server/frontend_wayland/wayland_input_dispatcher.h
+++ b/src/server/frontend_wayland/wayland_input_dispatcher.h
@@ -66,11 +66,6 @@ private:
     wayland::Weak<WlSurface> const wl_surface;
 
     std::chrono::nanoseconds timestamp{0};
-
-    /// Handle user input events
-    ///@{
-    void handle_keyboard_event(std::chrono::milliseconds const& ms, MirKeyboardEvent const* event);
-    ///@}
 };
 }
 }
--- a/src/server/frontend_wayland/wl_keyboard.cpp
+++ b/src/server/frontend_wayland/wl_keyboard.cpp
@@ -92,7 +92,7 @@ void mf::WlKeyboard::event(MirKeyboardEvent const* event, WlSurface& surface)
     }
 
     auto const serial = wl_display_next_serial(wl_client_get_display(client));
-    auto const timestamp = WlSeat::timestamp_of(event);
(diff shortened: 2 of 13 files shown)


src/server/frontend_wayland/wayland_input_dispatcher.h
```diff
@@ -66,16 +66,10 @@ class WaylandInputDispatcher
     wayland::Weak<WlSurface> const wl_surface;
 
     std::chrono::nanoseconds timestamp{0};
-    MirPointerButtons last_pointer_buttons{0};
-    std::experimental::optional<std::pair<float, float>> last_pointer_position;
 
     /// Handle user input events
     ///@{
     void handle_keyboard_event(std::chrono::milliseconds const& ms, MirKeyboardEvent const* event);
```
AlanGriffiths, line 72: I think this can go too.

src/server/frontend_wayland/wl_keyboard.cpp
```diff
@@ -20,10 +20,12 @@
 
 #include "wayland_utils.h"
 #include "wl_surface.h"
+#include "wl_seat.h"
```
AlanGriffiths, line 23: This is just to access `WlSeat::timestamp_of()`?

src/server/frontend_wayland/wl_keyboard.cpp
```diff
@@ -93,20 +112,20 @@ void mf::WlKeyboard::key(std::chrono::milliseconds const& ms, WlSurface* surface
...
             "Sending key 0x%2.2x %s to wl_surface@%u even though it was not explicitly given keyboard focus",
-            scancode, down ? "press" : "release", wl_resource_get_id(surface->resource));
+            scancode, down ? "press" : "release", wl_resource_get_id(surface.resource));
 
         focussed(surface, true);
     }
 }
 
-void mf::WlKeyboard::focussed(WlSurface* surface, bool should_be_focused)
+void mf::WlKeyboard::focussed(WlSurface& surface, bool should_be_focused)
```
AlanGriffiths, line 121: Why this change of signature? It seems to lead to a lot of needless diff.

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -20,6 +20,7 @@
 
 #include "wayland_utils.h"
 #include "wl_surface.h"
+#include "wl_seat.h"
```
AlanGriffiths, line 23: This is just to access WlSeat::timestamp_of()?

src/server/frontend_wayland/wl_seat.h
```diff
@@ -60,6 +63,14 @@ class WlSeat : public wayland::Seat::Global
...
     static auto from(struct wl_resource* seat) -> WlSeat*;
 
+    /// Returns the timestamp of the event in milliseconds, suitable for Wayland events
+    /// @{
+    static auto timestamp_of(MirInputEvent const* event) -> uint32_t;
+    static auto timestamp_of(MirKeyboardEvent const* event) -> uint32_t;
+    static auto timestamp_of(MirPointerEvent const* event) -> uint32_t;
+    static auto timestamp_of(MirTouchEvent const* event) -> uint32_t;
+    /// @}
+
```
AlanGriffiths, line 73: That's a set `static` methods with no obvious connection to `WlSeat`.

1. shouldn't they be at namespace scope?
2. is this the right header? `wl_timestamp.h`?

AlanGriffiths (changes requested):
I don't think `WlSeat` is the right place for `timestamp_of()`.

src/server/frontend_wayland/wl_seat.h
wmww: Should have explained this better, 2 reasons:
1. In the next phase the seat will manage mapping events to serials, which means it will deal with events and everything that uses events will already be using it
2. The timestamp conversion logic is a little messy and I'd rather not scatter it around the code base else it might get out of sync

I supposed I could make `timestamp_of(MirInputEvent)` on the seat and then `namespace {}` functions for calling it with the various specific types, or I could make `timestamp_of()` a standalone function in `wayland_utils.h` or something.

src/server/frontend_wayland/wl_pointer.cpp
wmww: Yes, but it will get included by the next PR regardless (see other comment)

src/server/frontend_wayland/wl_keyboard.cpp
wmww: I guess I didn't need to do that. Did it for consistency with the new code.

src/server/frontend_wayland/wl_seat.h
AlanGriffiths: I'm not sure I follow your response.

1. I still don't see why pure functions that take a `Mir*Event` and return a timestamp are members of `WlSeat`, nor why they are declared in the header that defines it. They are perfectly usable without any of `WlSeat`.
2. Why would you think I'm concerned about reducing the number of overloads? They all have the same semantics (I would be worried if they didn't).

wmww: The issue is that I need *somewhere* to put it, though I suppose that isn't the best argument. Is making them standalone functions declared and implemented in the seat files better? Where do you think they should go?

wmww: Perhaps the correct place is the event system.

AlanGriffiths: OK

AlanGriffiths (approved):
bors r+

PR #1896, opened 2021-01-22
Improve HiDPI support for XWayland

### LINKED ISSUE #1367: HiDPI support for X11 apps
When HiDPI support lands X11 apps will get scaled up, but will be low resolution. From sway I know it's possible to scale them correctly. Not a priority at the moment.

wmww:
NOTE: there is still a rather serious problem where the window frame is too big and input comes in in the wrong place, thus the draftness of the PR.

Until the X11 PRs noted in #1367 get merged, the whole stack is inherently a little jank. There is no standard way to ask X11 apps to scale, and there is no way to know which apps are scaling. The only way to make sure everything's the right size is to turn scaling off for everything and let Mir stretch apps, but that's obviously not preferable.

Most modern apps and toolkits do support scaling one way or another (the [Arch Wiki page](https://wiki.archlinux.org/index.php/HiDPI#IntelliJ_IDEA) is the best resource for getting them all working), so it makes sense to allow the user to set up app scaling for themselves and then tell Mir to assume all apps are scaled.

With this PR, the user can explicitly specify a scale with the `--x11-scale` option. If that's unset, Mir will pull the scale from `GDK_SCALE`, or default to 1 if that's also unset. `GDK_SCALE` is as close to a standard as we've got. GTK3 apps, Steam, and Various Java apps including Intellij respect it,

Fixes #1367 (as good as it can be currently).

### DIFF
--- a/src/miral/x11_support.cpp
+++ b/src/miral/x11_support.cpp
@@ -43,6 +43,11 @@ void miral::X11Support::operator()(mir::Server& server) const
         "Enable X11 support", mir::OptionType::null);
 
     server.add_configuration_option(
+        mo::x11_scale_opt,
+        "The scale to assume X11 apps use. If unset uses the value of GDK_SCALE or 1. Can be fractional",
+        mir::OptionType::string);
+
+    server.add_configuration_option(
         "xwayland-path",
         "Path to Xwayland executable", "/usr/bin/Xwayland");
 
--- a/src/platform/symbols.map
+++ b/src/platform/symbols.map
@@ -194,3 +194,10 @@ MIRPLATFORM_2.3 {
     mir::graphics::LinuxDmaBufUnstable::buffer_from_resource*;
   };
 } MIRPLATFORM_2.2;
+
+MIRPLATFORM_2.4 {
+ global:
+  extern "C++" {
+    mir::options::x11_scale_opt;
+  };
+} MIRPLATFORM_2.3;
--- a/src/platform/symbols.map
+++ b/src/platform/symbols.map
@@ -192,12 +192,6 @@ MIRPLATFORM_2.3 {
     mir::graphics::LinuxDmaBufUnstable::LinuxDmaBufUnstable*;
     mir::graphics::LinuxDmaBufUnstable::?LinuxDmaBufUnstable*;
     mir::graphics::LinuxDmaBufUnstable::buffer_from_resource*;
-  };
-} MIRPLATFORM_2.2;
-
-MIRPLATFORM_2.4 {
- global:
-  extern "C++" {
     mir::options::x11_scale_opt;
   };
-} MIRPLATFORM_2.3;
+} MIRPLATFORM_2.2;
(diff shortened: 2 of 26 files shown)


RAOF:
🎉

wmww:
Oh boy that was a long tail of work to get everything right (and if I know anything about long tails, it will continue to be) but most things seem to be working now. It's blocked on #1898 and #1899. Once both those merge, I think whats left (most of it) should all be reviewed at once. I can split it up if someone wants me to.

wmww:
Serious issues are fixed (AFAIK) and prereqs are merged. This is no longer draft.

src/platform/symbols.map
```diff
@@ -193,4 +193,11 @@ MIRPLATFORM_2.3 {
...
-} MIRPLATFORM_2.2;
\ No newline at end of file
+} MIRPLATFORM_2.2;
+
+MIRPLATFORM_2.4 {
+ global:
+  extern "C++" {
+    mir::options::x11_scale_opt;
+  };
+} MIRPLATFORM_2.3;
```
AlanGriffiths, line 203: I don't think we need a new stanza as Mir2.3 has yet to be branched.

wmww: Ack

AlanGriffiths:
> With this PR, the user can explicitly specify a scale with the `--x11-scale` option. If that's unset, Mir will pull the scale from `GDK_SCALE`, or default to 1 if that's also unset. `GDK_SCALE` is as close to a standard as we've got. GTK3 apps, Steam, and Various Java apps including Intellij respect it,

Why would we not simply take the scale from the output(s)?

AlanGriffiths:
Do we want to consider setting `GDK_SCALE` in `--app-env-x11` if `--x11-scale` is set?

wmww:
> Do we want to consider setting GDK_SCALE in --app-env-x11 if --x11-scale is set?

Hm, maybe, but I think that's out of scope of this PR (and may or may not be considered out of scope of core Mir in general). This PR is focused on the technical side of making X11 HiDPI possible. If we try to do too many things automatically and get it wrong we might just end up being annoying. Also see below for an example of why this might be sometimes wrong.

> Why would we not simply take the scale from the output(s)?

Ideally the output(s) scales, the X11 scale and the app scales would all match perfectly, but they're kept independent because there are cases when tweaking them independently is preferable.
- If you would rather all apps be blury than some be small, you can not set any X11 scaling related environment variables and use the default `--x11-scale 1` (despite the output having a higher scale)
- If you wanted a crisp but slightly big or small app on a fractionally scaled output, you might set `--x11-scale` to match the output but `GDK_SCALE` to be a whole number.

src/miral/x11_support.cpp
```diff
@@ -42,6 +42,11 @@ void miral::X11Support::operator()(mir::Server& server) const
         mo::x11_display_opt,
         "Enable X11 support", mir::OptionType::null);
 
+    server.add_configuration_option(
+        mo::x11_scale_opt,
+        "The scale to assume X11 apps use. If unset uses the value of GDK_SCALE or 1. Can be fractional",
+        mir::OptionType::string);
```
AlanGriffiths, line 48: (Out of scope for this PR)

Maybe we should plumb through a `float` configuration type.

AlanGriffiths (approved):
Looks sensible, and my questions are answered. I've left some suggestions, but haven't experimented enough to be confident of how it works in practice.

Having said that, it should be mostly harmless unless someone tries to use it.

bors r=wmww

AlanGriffiths:
> bors r=wmww

Oops! That's not what I thought I typed.

Never mind, I'll PR my suggestions and we can take it from here.

PR #1898, opened 2021-01-22
Enable xdg_output_unstable_v1 by default

### LINKED ISSUE #1897: X11 app see wrong output size on scaled outputs
To observer this problem:
- run Mir with XWayland enabled and a scaled output (with Mir-on-X that's `miral-shell --enable-x11 --x11-output 2000x1200^2`)
- run `xdpyinfo | grep -B 2 resolution`
- It shows the unscaled resolution, which is incorrect

To demonstrate the resolution is incorrect:
- run `intellij-idea-community`
- if you have a project open, close it
- the welcome screen is now centered on the bottom-right corner (because it thinks the output is larger than it is, and that's the center)

This could be a very serious problem on 4k displays, where windows will appear completely off the screen. This is related to HiDPI support (#1367), but not exactly the same thing because it should have more correct behavior even when being run with blurry windows.

wmww:
In looking at XWayland HiDPI I was surprised to discover we don't enable the XDG output protocol by default. XDG Output helps XWayland handle scaled outputs, and fixes who knows what other usecases.

I see no potential drawbacks or security concerns to for enabling it. It doesn't allow clients to *do* anything or give any information about other clients, it just [gives additional information about outputs](https://github.com/MirServer/mir/blob/master/src/wayland/protocol/xdg-output-unstable-v1.xml).

The only reason I can think of a compositor wouldn't want to use it is to work around some client specific bug, or if the compositor didn't lay out outputs in a global coordinate space. Since Mir always has a global coordinate space and I know of no client bugs to work around, defaulting it on seems like an obvious decision.

Fixes #1897

### DIFF
--- a/src/server/frontend_wayland/wayland_default_configuration.cpp
+++ b/src/server/frontend_wayland/wayland_default_configuration.cpp
@@ -181,7 +181,8 @@ auto mf::get_standard_extensions() -> std::vector<std::string>
     return std::vector<std::string>{
         mw::Shell::interface_name,
         mw::XdgWmBase::interface_name,
-        mw::XdgShellV6::interface_name};
+        mw::XdgShellV6::interface_name,
+        mw::XdgOutputManagerV1::interface_name};
 }
 
 auto mf::get_supported_extensions() -> std::vector<std::string>


AlanGriffiths (approved):
I think thje principle reason it isn't enabled by default is we didn't need the additional features over `wl_output` yet.

But while having this on by default makes sense "on desktop oriented systems" it isn't clear this is right for all uses of mir-kiosk. To quote from the protocol summary:

> Some information are more specific to the concept of an output for
> a desktop oriented system and may not make sense in other applications,
> such as IVI systems for example.

Having said that, it is easy to configure it off programmatically, or via a .config for deployments where that is appropriate. (And I don't think the WM policy any of our examples or downstreams require it to be off.)

bors r+

PR #1899, opened 2021-01-22
Add WlClient class

wmww:
This breaks `ClientPrivate` in `wayland_connector.cpp` into its own files. The immediate benefits are simplifying `wayland_connector.cpp` and improving and documenting the logic around creating clients. 

The main motivation for doing this now is to help with #1896. Outputs for the XWayland client will sometimes need to report different dimensions than outputs for Wayland clients. I could try to pipe the output manager and XDG output manager through to XWayland, add a map of clients, etc etc but the probably easier and IMO cleaner solution is to add a property to clients which XWayland sets the the outputs use. Creating a client class is the prerequisite for adding a property to a client.

### DIFF
--- a/src/server/frontend_wayland/wl_client.h
+++ b/src/server/frontend_wayland/wl_client.h
@@ -44,9 +44,6 @@ class SessionAuthorizer;
 class WlClient
 {
 public:
-    struct ConstructionCtx;
-    struct ClientCtx;
-
     /// Initializes a ConstructionCtx that will create a WlClient for each wl_client created on the display. Should only
     /// be called once per display. Destruction of the ConstructionCtx is handled automatically.
     static void setup_new_client_handler(
@@ -56,6 +53,9 @@ public:
         std::function<void(WlClient&)>&& client_created_callback);
 
     static auto from(wl_client* client) -> WlClient*;
+
+    WlClient(wl_client* raw_client, std::shared_ptr<scene::Session> const& client_session, shell::Shell* shell);
+    ~WlClient();
 
     /// The underlying Wayland client
     wl_client* const raw_client;
@@ -68,11 +68,7 @@ public:
     /// individual apps.
     std::shared_ptr<scene::Session> const client_session;
 
-    ~WlClient();
-
 private:
-    WlClient(wl_client* raw_client, std::shared_ptr<scene::Session> const& client_session, shell::Shell* shell);
-
     /// This shell is owned by the ClientSessionConstructor, which outlives all clients.
     shell::Shell* const shell;
 };
--- a/src/server/frontend_wayland/wl_client.h
+++ b/src/server/frontend_wayland/wl_client.h
@@ -54,11 +54,11 @@ public:
 
     static auto from(wl_client* client) -> WlClient*;
 
-    WlClient(wl_client* raw_client, std::shared_ptr<scene::Session> const& client_session, shell::Shell* shell);
+    WlClient(wl_client* client, std::shared_ptr<scene::Session> const& session, shell::Shell* shell);
     ~WlClient();
 
     /// The underlying Wayland client
-    wl_client* const raw_client;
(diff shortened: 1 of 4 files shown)


src/server/frontend_wayland/wl_client.h
```diff
@@ -0,0 +1,82 @@
...
+    /// The underlying Wayland client
+    wl_client* const raw_client;
+
+    /// The Mir session associated with this client. Be careful when using this that it's actually the session you want.
+    /// All clients have a session but the surfaces they create may get associated with additional sessions.
+    ///
+    /// For example all surfaces from a single XWayland server are attached to a single WlClient with a single cleint
+    /// session, but their scene::Surfaces are associated with multiple sessions created in the XWayland frontend for
+    /// individual apps.
+    std::shared_ptr<scene::Session> const client_session;
```
AlanGriffiths, line 69: This public data, particularly with the caveats, looks odd in a `class`.

src/server/frontend_wayland/wl_client.h
```diff
@@ -0,0 +1,82 @@
...
+
+namespace frontend
+{
+class SessionAuthorizer;
+
+class WlClient
+{
+public:
+    struct ConstructionCtx;
+    struct ClientCtx;
```
AlanGriffiths, line 48: Is there an advantage to declaring these structs in class scope as opposed to the anonymous namespace of the implementation file? They are not mentioned elsewhere in the header and don't need access to private parts.

AlanGriffiths (changes requested):
I don't object to refactoring the code in this direction, but the WlClient abstraction needs some tidying up. (The specific issues commented on look like symptoms of a poor abstraction, not nits to be fixed up.)

src/server/frontend_wayland/wl_client.h
wmww: I have no issue with private const data with getters

src/server/frontend_wayland/wl_client.h
wmww: `ConstructionCtx` didn't need private access at all. `ClientCtx` was using the private constructor, but I'm not sure that was right.

wmww: Fixed by constructing `WlClient` in `handle_client_created` (which IMO is a better place) and making that a private static function.

src/server/frontend_wayland/wl_client.h
AlanGriffiths: I have no issue with `struct`s. :grinning:

src/server/frontend_wayland/wl_client.h
```diff
@@ -0,0 +1,83 @@
...
+
+private:
+    WlClient(wl_client* client, std::shared_ptr<scene::Session> const& session, shell::Shell* shell);
+
+    static void handle_client_created(wl_listener* listener, void* data);
+
+    /// This shell is owned by the ClientSessionConstructor, which outlives all clients.
+    shell::Shell* const shell;
+    wl_client* const client;
+    std::shared_ptr<scene::Session> const session;
```
AlanGriffiths, line 78: Sorry, I have to come back to this and think about the lifetime management. I.e. can `client` or `session` outlive their usefulness, or should they be `Weak` and `weak_ptr`?

wmww: When the `wl_client` is closed, `cleanup_client_ctx()` deletes the `ClientCtx`, which then drops the `WlClient`. The `WlClient` closes the session in it's destructor (so it makes sense for it to hold a strong ref to the session until then. If we want to make `wayland::Weak`s out of clients making that possible would be easy, but I haven't done that here as there is currently no need.

AlanGriffiths (approved):
OK. `WlClient` stills seems more of an implementation detail of wayland_connector.cpp than a meaningful abstraction, but I'll await the follow-up.

bors r+

PR #1901, opened 2021-01-26
Refactor X11Resources to prepare for X11 platform improvements

wmww:
The X11 input platform (which gets events, like window resize) needs talk to the graphics platform (which can apply output configuration changes). The previous method for doing this only shared display configs and only from the graphics platform to the input platform. This PR adds an interface to add flexibility.

### DIFF
--- a/src/platforms/x11/X11_resources.h
+++ b/src/platforms/x11/X11_resources.h
@@ -20,8 +20,9 @@
 #define MIR_X11_RESOURCES_H_
 
 #include <X11/Xlib.h>
-#include <experimental/optional>
+#include <optional>
 #include <unordered_map>
+#include <functional>
 #include <mutex>
 
 namespace mir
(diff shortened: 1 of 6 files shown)


src/platforms/x11/X11_resources.h
```diff
@@ -38,19 +40,31 @@ int mir_x11_error_handler(Display* dpy, XErrorEvent* eev);
...
+    void set_set_output_for_window(Window win, VirtualOutput* output);
+    void clear_output_for_window(Window win);
+
+    /// X11Resources's mutex stays locked while the provided function runs
+    void with_output_for_window(Window win, std::function<void(std::optional<VirtualOutput const*> output)> fn);
 
     static X11Resources instance;
 
 private:
+    std::mutex mutex;
```
AlanGriffiths, line 65: This is for the future? It is not needed at present as all the access is during the single-threaded initialization phase of execution.

src/platforms/x11/graphics/display.h
```diff
@@ -104,17 +106,19 @@ class Display : public graphics::Display
     Frame last_frame_on(unsigned output_id) const override;
 
 private:
-    struct OutputInfo
+    struct OutputInfo : ::mir::X::X11Resources::VirtualOutput
```
AlanGriffiths, line 109: It seems unintuitive for a `struct` to implement an interface. OTOK I guess this is "private" to `Display` and the only reason the struct is declared in the header is to preced the use with `unique_ptr`.

AlanGriffiths (approved):
OK

bors r+

PR #1904, opened 2021-01-29
Update the `--x11-scale` option

AlanGriffiths:
Update the `--x11-scale` option to:

1. Suggest configuring GDK_SCALE in `--app-args-x11`
2. Use GDK_SCALE as the default setting
3. Give an error if `--x11-scale` value cannot be parsed

### DIFF
--- a/src/miral/x11_support.cpp
+++ b/src/miral/x11_support.cpp
@@ -42,10 +42,23 @@ void miral::X11Support::operator()(mir::Server& server) const
         mo::x11_display_opt,
         "Enable X11 support", mir::OptionType::null);
 
+
+    auto const x11_scale_default = []() -> char const* {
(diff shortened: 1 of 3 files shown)


wmww:
Looks mostly good, except just for fun I passed `--x11-scale -1` to see what happens. Started fine, but when I launched an X11 app, Sway crashed. Not Mir. The outer Sway Mir was running in. I'll need to get to the bottom of this.

wmww:
I'm not sure what was happening there, but as long as we prohibit invalid scales it doesn't happen.

wmww (approved):
bors delegate+

AlanGriffiths (commented):
Should possibly be `range_error` but OK

bors merge

PR #1905, opened 2021-01-30
Async XWayland copy paste

wmww:
Pipes the X11 event dispatcher through to the clipboard provider and source, and adds pending operations to it so they do not block.  Clipboard provider's connection needed to be changed from a ref to a `shared_ptr` so `SelectionSender` wouldn't outlive it, and this adds a bit of noise to the diff.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_clipboard_provider.cpp
+++ b/src/server/frontend_xwayland/xwayland_clipboard_provider.cpp
@@ -21,6 +21,7 @@
 
 #include "xwayland_log.h"
 #include "mir/scene/clipboard.h"
+#include "mir/dispatch/multiplexing_dispatchable.h"
 
 #include <xcb/xfixes.h>
 #include <string.h>
@@ -29,6 +30,7 @@
 
 namespace mf = mir::frontend;
 namespace ms = mir::scene;
+namespace md = mir::dispatch;
 
 namespace
 {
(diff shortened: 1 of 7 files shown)


src/server/frontend_xwayland/xwayland_clipboard_provider.cpp
```diff
@@ -52,38 +54,139 @@ auto create_selection_window(mf::XCBConnection const& connection) -> xcb_window_
...
 }
 
-class ReadError : public std::runtime_error
+void send_selection_notify(
+    mf::XCBConnection& connection,
+    xcb_timestamp_t time,
+    xcb_window_t requestor,
+    xcb_atom_t property,
+    xcb_atom_t selection,
+    xcb_atom_t target)
```
wmww, line 63: Needed to be broken out of `XWaylandClipboardProvider` because both that and `SelectionSender` need to call it.

src/server/frontend_xwayland/xwayland_clipboard_provider.cpp
```diff
@@ -240,84 +345,28 @@ void mf::XWaylandClipboardProvider::send_data(
...
-        send_selection_notify(time, requester, property, connection.CLIPBOARD, target);
-    }
-    else
-    {
-        log_warning("failed to send clipboard data to X11 client: incremental transfers not yet implemented");
-        send_selection_notify(time, requester, XCB_ATOM_NONE, connection.CLIPBOARD, target);
-        // TODO: remember to flush the connection before reading
-        return;
+        in_fd = {}; // release ownership of the fd here
     }
```
AlanGriffiths, line 353: I realize this is pre-existing, but why does `in_fd` have a scope that outlasts its use? Fixing that would avoid this ugly reset. (OTOH why does it even exist?)

AlanGriffiths (approved):
LGTM

bors r+

PR #1909, opened 2021-02-02
X11 platform: make things const

wmww:
This changes `EGLHelper` from two stage to one stage initialization, which allows it to be const. Also makes some other things const.


AlanGriffiths (approved):
Thanks for this. The two-step init has bugged me for years but I've never got around to fixing it.

bors r+

PR #1910, opened 2021-02-03
Refactor example clients

wmww:
This cleans up and removes code duplication in the example clients. It consolidates the code for acquiring Wayland globals and creating shell surfaces which was previously duplicated between the software clients and EGL clients. It also fixes a number of leaks and crashes that showed up when I enabled X11 platform resizing.


AlanGriffiths (approved):
LGTM

bors r+

Issue #1314, opened 2020-02-24
[Mir-on-X] Cursor movement mapped incorrectly


AlanGriffiths:
Likely not all of this setup is required, but this is how I got there...

0. Install mir-demos, mir-graphics-drivers-desktop from ppa:mir-team/rc
1. create a `.config/miral-shell.config` with the line `window-manager=tiling`
2. Log in to "Mir Shell"
3. Open a terminal `Ctrl-Alt-T`
4. Install mir-kiosk: `sudo snap install mir-kiosk`
5. Run mir-kiosk `env -u WAYLAND_DISPLAY mir-kiosk`
6. Install mir-kiosk-neverputt: `sudo snap install mir-kiosk-neverputt`
7. Check which WAYLAND_SOCKET mir-kiosk provides and run `WAYLAND_DISPLAY=wayland-0 mir-kiosk-neverputt`

Expect: the mouse drawn by neverputt moves with the mouse drawn by miral-shell
Actual: the mouse drawn by neverputt moves to reflect the actual screen position (when the cursor is on the Mir-on-X window)

AlanGriffiths:
OK, this seems to be an issue with the display size reporting from the Mir-on-X11 platform. We're forcing the Mir-on-X11 window to resize into a tile, but that isn't being reported through mir-kiosk to the neverputt client.

I don't think it is being reported by the Mir-on-X11 platform at all, as that doesn't expect the window to be resized.

A workaround is to specify a `--x11-output` that is smaller than the tile.

wmww:
I can confirm that this has always been a problem

AlanGriffiths:
There is a comment in #1914 saying this is fixed, but while it has improved a lot, **it still misplaces the cursor in the scenario described**.

Initially, the panel containing neverputt mismatches the size of the content drawn by neverputt (which is vertically centered), The mismatch is by the size of the upper margin. Creating additional windows and switching between them leads to a bunch of weird displacements.

wmww:
I think this is the same thing I've been looking into recently because I see it in Sway and i3. I will take care of it (but perhaps not until after the XCB port lands).

wmww:
`mir-kiosk-neverputt` doesn't seem to be working for me. When run normally it fails with
```
Failure to initialize SDL (Could not initialize UDEV)
```
when run with sudo it seems to start, but it just freeses and it or Mir never show up.

Can you check if this is still an issue?

AlanGriffiths:
[...]

Issue #1900, opened 2021-01-25
[Mir-on-X] Locking up GNOME shell


AlanGriffiths:
I've only started seeing this recently, and only on one system where I've installed DisplayLink drivers and switched to GNOME/Wayland. As I've not seen this elsewhere, this may be a specific problem with the Wayland/DisplayLink stack, or this may be exposing a race that could appear elsewhere.

To reproduce:

1. Start a Mir-on-X session. (Could be any of the example servers, could be local or using X-forwarding.)
2. Move the cursor fairly swiftly off the Mir-on-X window. (Doesn't seem to happen when moving the cursor onto the window, or when deliberately moving slowly, but that might just be luck.)

Expect: Everything works fine.
Actual: Occasionally input locks up (VT switching doesn't work, even when ssh'ing from another system)

There's also a lot of CPU being used here:
```plain
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                           
  76111 root      20   0 3689452  79500  10564 S 164.5   0.2  31:28.06 DisplayLinkMana                                                   
   7680 alan      20   0   10.5g 401384 141400 R 100.0   1.2  31:56.10 gnome-shell                                                       
```
If I stop the Mir process by attaching a debugger, then things revert to normal.

AlanGriffiths:
> If I stop the Mir process by attaching a debugger, then things revert to normal.

Correction. The CPU settles down, and input is handled by the shell (e.g. cursor is drawn, workspace and app switching works). BUT apps do not receive cursor or keyboard input.

AlanGriffiths:
Probably related, and suggestive that we have a race that manifests differently on different stacks:

When hosting a Mir-on-X session on another machine (using `ssh -XC`) moving the cursor off the Mir-on-X window sometimes has the cursor disappear and clicking on the window it should be over has no effect. (I.e. input is probably still grabbed.)

Waving the cursor back and forth over the edge of the Mir-on-X window until it appears over other windows restores normal operation.

AlanGriffiths:
Pretty sure now this is an interaction with the DisplayLink driver. I've not reproduced on other stacks, and the above problems (hang and disappearing cursor) only happens when both the GNOME session and the Mir session are running on that hardware. (Including Mir in a VM on that hardware.)

That doesn't prove that Mir isn't doing something wrong, but limits the scope of the problem.

AlanGriffiths:
[...]

PR #1914, opened 2021-02-04
Allow resizing Mir-on-X windows

wmww:
This should work. Mostly. You can still make the background fail by resizing constantly for too long and there's some glitches when there are multiple windows, but I think good enough to land.


AlanGriffiths:
This seems to work on GNOME/Wayland, but not on GNOME/Xorg.

wmww:
What do you mean by "work"? Is it glitchy? Are you unable to resize at all? If the latter, can you maximize?

AlanGriffiths:
> What do you mean by "work"? Is it glitchy? Are you unable to resize at all? If the latter, can you maximize?

I get a resize cursor, but dragging doesn't change window size.

Maximize/unmaximize works

AlanGriffiths:
Oh! I got something like #1900 on GNOME/Wayland without DisplayLink.

wmww:
Fixes #1314

wmww:
There's some flickering and there seems to be some buffering (with lots of size changes it gets behind and then needs to "catch up", but for me it's usable and doesn't crash. I haven't tried it on GNOME yet.

AlanGriffiths:
OK, the problems I was seeing are my separate issues with Xwayland on Displaylink. Could we avoid queueing the repaints by not triggering a redraw until the mouse is released? This doesn't look good:

https://user-images.githubusercontent.com/9048879/114004365-7e63a100-9856-11eb-87ac-ae2d73a21e39.mp4

wmww:
Hm, I'll think about that. I'd rather do it by figuring out where updates are being buffered and short-circuit that (so it doesn't have to process older states when newer ones are available). If that doesn't end up being practical not sending updates until mouse up is a reasonable fallback.

wmww:
Perhaps that's a little better? It still doesn't apply the changes instantly so there's some jank, but it now skips updates when they come faster than they're processed.

src/server/scene/mediating_display_changer.h
```diff
@@ -122,6 +122,12 @@ class MediatingDisplayChanger : public frontend::DisplayChanger,
     std::weak_ptr<scene::Session> currently_previewing_session;
     struct SessionObserver;
     std::unique_ptr<SessionObserver> const session_observer;
+
+    /// Mutex protecting pending_configuration
+    std::mutex pending_configuration_mutex;
+    /// See configure_for_hardware_change(). The config that will be applied by a currently in-flight server action.
+    /// If null, there is no config to apply or hardware config server action queued.
+    std::shared_ptr<graphics::DisplayConfiguration> pending_configuration;
```
AlanGriffiths, line 130: I was about to suggest `atomic<shared_ptr<...>>`, but that's C++20. One for another year...

AlanGriffiths (approved):
Works well enough to land

bors r+

PR #1915, opened 2021-02-04
Example WaylandShm: don't allow incorrectly sized buffers

wmww:
In practice this doesn't currently matter because with only one buffer, buffers are never re-used (which is I guess ok for a static background, where updates are infrequent). However, it was technically incorrect to allow buffers with different geometry than requested to be returned. This PR guarantees a new buffer is created if different geometry is requested.

EDIT: this PR now also ports `SwSplash` to the new classes, which fixes some sketchy buffer management it was doing.


AlanGriffiths (approved):
bors r+

PR #1916, opened 2021-02-04
Decoration provider: reduce redraws

wmww:
The idea here is that if anything is updated, only redraw once after all events are dispatched. This keeps updates from stacking up if they come in faster then we can software render the background. In practice this didn't seem to fix any of the issues I'm seeing, but it's still a good idea.

### DIFF
--- a/examples/example-server-lib/decoration_provider.cpp
+++ b/examples/example-server-lib/decoration_provider.cpp
@@ -284,14 +284,19 @@ DecorationProviderClient::DecorationProviderClient(wl_display* display)
 
 void DecorationProviderClient::events_dispatched()
 {
+    bool needs_flush = false;
     for (auto const& output : outputs)
     {
         if (output.second->needs_redraw)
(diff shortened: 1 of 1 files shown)


examples/example-server-lib/decoration_provider.cpp
```diff
@@ -281,25 +282,31 @@ DecorationProviderClient::DecorationProviderClient(wl_display* display)
...
 
+void DecorationProviderClient::events_dispatched()
+{
+    for (auto const& output : outputs)
+    {
+        if (output.second->needs_redraw)
+        {
+            draw_background(*output.second);
+            output.second->needs_redraw = false;
+            wl_display_flush(display());
```
AlanGriffiths, line 293: Probably minor, but do we need to flush after every output? Or once after all?

AlanGriffiths (approved):
Looks sensible

bors delegate=wmww

PR #1920, opened 2021-02-04
Snap beta (not candidate) from mir-team/rc

AlanGriffiths:
This is to support a change to our workflow that allows an extra verification step.

* edge: automatically built from /dev. (Same as now)
* beta: automatically built from /rc. (As candidate is now)
* candidate: promoted by us from beta. (Typically when we copy /rc to /release)
* stable: promoted from candidate. (After CfT etc.)

### DIFF
--- a/tools/process_snaps.py
+++ b/tools/process_snaps.py
@@ -31,7 +31,7 @@ SOURCE_NAME = "mir"
 SNAPS = {
     "mir-kiosk": {
         "edge": {"ppa": "dev", "recipe": "mir-kiosk-edge"},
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
"No merge" while we discuss

yphus:
LGTM, thanks for the use of the beta channel.

AlanGriffiths:
@Saviq I think we're agreed that we want at least these changes. Can we land these while we discuss others?

Saviq:
Yup.

bors r+

AlanGriffiths:
I don't see why that failed. Does it happen again?

bors r=Saviq

PR #1922, opened 2021-02-08
Update mir-shell script to avoid system-dependent `gnome-terminal` weirdness.

AlanGriffiths:
We've got a `miral-terminal` script to handle variations in the terminal implementations across systems. Use it for the "Mir shell" desktop.


wmww (approved):
This seems to fix the terminal coming up as an X11 app
bors r+

PR #1928, opened 2021-02-09
Bump input platform ABI version to 8

wmww:
We didn't change any symbols, but we did change the meaning of scroll to match the Wayland values. It is now (from wayland.xml):

> the length of a vector along the specified axis in a coordinate space identical to those of motion events, representing a relative movement along the specified axis.

On platforms that haven't made this change scrolling with come out flipped and very slow.


AlanGriffiths (approved):
LGTM

bors r+

PR #1930, opened 2021-02-12
XWayland incremental data transfers

wmww:
This enables large copy-pastes between X11 and Wayland.

The X11 protocol this implements can be found [here](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#incr_properties). Basically, first the property is set to type `INCR`, once the receiver deletes that chunks of data start being sent (a new one after the previous is deleted). The last chunk has a length of zero, and then the receiver knows it's done.

How large a data transfer is needed to make this PR relevant? Unclear. For Wayland -> X11 we have it hard-coded to 64kb (same as Weston). Somewhere I read the real cutoff was 256kb. X11 Kate seems to allow copying out much bigger buffers without incremental transfers. In any case, There is no longer a limit with this PR.


AlanGriffiths (approved):
Looks sensible. (But we clearly don't have a good test case.)

bors r+

Saviq:
bors r=AlanGriffiths

PR #1932, opened 2021-02-12
[snaps] use PPA as defined for the snap/channel

Saviq:
Rather than depending on the recipe's configuration.

`channels` isn't required, so can skip altogether.


AlanGriffiths (approved):
If it works, sure.

bors merge

PR #1933, opened 2021-02-13
Adding DisplayConfigurationPolicy back as a parameter

erlend-g:
Adding DisplayConfigurationPolicy back as a parameter to DisplayConfigurationPolicyWrapper (The qtmir version had this).

This is need by MirDisplayConfigurationPolicy in 
qtmir/src/platforms/mirserver/mirdisplayconfigurationpolicy.cpp

### DIFF
--- a/debian/libmiroil1.symbols
+++ b/debian/libmiroil1.symbols
@@ -21,6 +21,5 @@ libmiroil.so.1 libmiroil1 #MINVER#
  (c++)"miroil::PersistDisplayConfig::~PersistDisplayConfig()@MIROIL_1.0" 2.2.0
  (c++)"typeinfo for miroil::DisplayConfigurationPolicy@MIROIL_1.0" 2.2.0
  (c++)"vtable for miroil::DisplayConfigurationPolicy@MIROIL_1.0" 2.2.0
- MIROIL_2.2@MIROIL_2.2 2.2.0
  (c++)"miroil::PersistDisplayConfig::PersistDisplayConfig(std::shared_ptr<miroil::DisplayConfigurationStorage> const&, std::function<std::shared_ptr<miroil::DisplayConfigurationPolicy> (std::shared_ptr<mir::graphics::DisplayConfigurationPolicy> const&)> const&)@MIROIL_1.0" 2.2.0
  (c++)"miroil::dispatch_input_event(miral::Window const&, MirInputEvent const*)@MIROIL_1.0" 2.2.0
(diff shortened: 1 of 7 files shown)


AlanGriffiths:
Which repo and branch of QtMir are you referencing?

erlend-g:
https://github.com/ubports/qtmir
Branch: xenial_-_edge_-_wayland_-_x11

erlend-g:
Sorry, it made a mistake... I didn't plan to have two commits, in this PR... I was planning to split up into two PRs.

New at this I'm afraid....

So the second commit. Moves the file qtmir/src/platforms/mirserver/eventdispatch.cpp/h into miroil.
Because it uses now internal class mir::scene::Surface

include/miroil/miroil/eventdispatch.h
```diff
@@ -0,0 +1,29 @@
...
+#ifndef MIROIL_EVENTDISPATCH_H
+#define MIROIL_EVENTDISPATCH_H
+
+namespace miral { class Window; }
+
+struct MirInputEvent;
+
+namespace miroil
+{
+void dispatchInputEvent(const miral::Window& window, const MirInputEvent* event);
```
AlanGriffiths, line 26: Mir has an idiosyncratic naming style, functions should be named in snake_case.

AlanGriffiths (changes requested):
The reason the build is failing is that you're changing the miroil ABI - that requires an update to `debian/libmiroil1.symbols`. (This is only checked for release builds with g++ and Debian packaging tools available, so is easy to miss when building locally.)

AlanGriffiths (changes requested):
The reason the build is failing is that you're changing the miroil ABI - that requires an update to debian/libmiroil1.symbols.

erlend-g:
I know, I just had some problems fixing it... But I think it's fixed now.

debian/libmiroil1.symbols
```diff
@@ -21,3 +21,6 @@ libmiroil.so.1 libmiroil1 #MINVER#
  (c++)"miroil::PersistDisplayConfig::~PersistDisplayConfig()@MIROIL_1.0" 2.2.0
  (c++)"typeinfo for miroil::DisplayConfigurationPolicy@MIROIL_1.0" 2.2.0
  (c++)"vtable for miroil::DisplayConfigurationPolicy@MIROIL_1.0" 2.2.0
+ MIROIL_2.2@MIROIL_2.2 2.2.0
```
AlanGriffiths, line 24: You should't need this

erlend-g: Okey, removed the line manually, and "make check-miroil-symbols" doesn't recreate it when I ran it.

It appear after first time I ran the package build.

AlanGriffiths (approved):
Thanks!

Issue #1941, opened 2021-02-19
Qt popup menus sometimes disappear when navigating submenus


wmww:
To reproduce:
- Open Kate
- Open menu with lots of sub-menus (like `tools` at the top)
- Move up and down, opening submenus

Expected: 
- You can keep doing that as long as you like

Observed:
- After a couple submenus are opened and closed, the toplevel menu is closed and you have to open it again

Issue does not exist on Sway or Weston

wmww:
On Sway and Weston I'm seeing a bunch of the following message:
> qt.qpa.wayland: setGrabPopup called with a parent, QtWaylandClient::QWaylandXdgSurface(0x55afb03ce4f0) which does not match the current topmost grabbing popup, QtWaylandClient::QWaylandXdgSurface(0x55afb0609ad0) According to the xdg-shell protocol, this is not allowed. The wayland QPA plugin is currently handling it by setting the parent to the topmost grabbing popup. Note, however, that this may cause positioning errors and popups closing unxpectedly because xdg-shell mandate that child popups close before parents

But interestingly, I'm not seeing that on Mir (where unexpected popup closing is actually happening). Not sure what that means.

wmww:
Reading through the wayland-debug logs of kate on both Mir and sway, I've noticed some things:
- The above warning seems to be the first relevant divergence between the two. It's unclear what Sway and Mir are doing differently to cause it to only appear on Sway
- Later, it seems that Sway is deleting old popups when a new popup is created and Mir is not
- On Mir Kate is creating a toplevel for one of the later submenus instead of a popup. This causes the root popup to close, which presumably causes Qt to close the submenu as well, resulting in the vanishing popups observed.

wmww:
I'm fairly convinced the warning has something to do with the bug. Testing with KDevelop, it seems that both the warning and bug are intermittent, with the popups generally vanishing when the warning is *not* shown.

wmww:
Figured it out. Kate is using the current keyboard-focused window as the popup parent. On other compositors that's the toplevel window. This isn't, but Qt notices this and fixes it (thus the warning). On Mir we're `wl_keyboard.enter`ing the popups. This means when switching from one popup to another (like moving down a list of sub-menus), it tries to use the previous popup as the new one's parent. The previous popup's shell surface has already been destroyed, so this fails in a way Qt can't fix. It instead tries to make a new toplevel. This causes the parent popup to become unmaped, which Qt notices and responds to by destroying the child popup it just created.

[...]

PR #1942, opened 2021-02-20
Don't send input events after dismissing popup menus

### LINKED ISSUE #1818: Popups are not dismissed when clicking again in the same place
To reproduce:
- Open any app with popup menus (tested: gnome-terminal, gedit, kate)
- Click on a menu bar item to open a pupup menu
- Without moving the pointer, click again on the menu bar item

Expected (as seen on Sway, not tested on GNOME but expected to work):
- Menu is dismissed

Actual:
- Menu briefly blinks away and then reappears

wmww:
Fixes #1818, but I'm not convinced this is the best way to do it. Instead of blocking input events and sending `.popup_done` (as this PR does), other compositors seem to send the input events and *not* send `.popup_done` in this case (they still send `.popup_done` in other cases, like the parent loosing focus). Also this is not a fix for #1941, which seems like it might be related.

### DIFF
--- a/src/miral/minimal_window_manager.cpp
+++ b/src/miral/minimal_window_manager.cpp
@@ -56,6 +56,41 @@ auto touch_center(MirTouchEvent const* event) -> mir::geometry::Point
 
     return {total_x/count, total_y/count};
 }
+
+auto is_ancestor(
+    miral::WindowManagerTools const& tools,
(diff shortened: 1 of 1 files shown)


wmww:
Closed in favor of #1948 (which includes this)

PR #1943, opened 2021-02-23
Miroil add event builder

erlend-g:
In Qtmir you have the class: qtmir:EventBuilder which contains both links to Qt and internal Mir stuff. See:
https://github.com/ubports/qtmir/blob/xenial_-_edge_-_wayland_-_x11/src/platforms/mirserver/eventbuilder.cpp

So what I have done is to split this class in two:
1. A base class in miroil that create MirEvents and deals with cookies.
2. A derived class that handles all the Qt stuff in (QtMir).

### DIFF
--- a/include/miroil/miroil/event_builder.h
+++ b/include/miroil/miroil/event_builder.h
@@ -1,0 +1,110 @@
+/*
+ * Copyright (C) 2016-2021 Canonical, Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License version 3, as published by
+ * the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+ * SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIROIL_EVENT_BUILDER_H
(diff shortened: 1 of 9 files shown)


wmww (commented):
This looks reasonable to me

AlanGriffiths:
@erlend-g not directly relevant to this PR but, apparently, the reference version of QtMir should be https://gitlab.com/ubports/core/qtmir.

See: https://github.com/MirServer/mir/issues/1744#issuecomment-700336754

(I've only just noticed this while reviewing Mir issues)

include/miroil/miroil/event_builder.h
```diff
@@ -0,0 +1,110 @@
...
+        float pressure_value, float touch_major_value, float touch_minor_value, float size_value);
+    
+    // Key event
+    mir::EventUPtr make_event(MirInputDeviceId device_id, std::chrono::nanoseconds timestamp,
+        std::vector<uint8_t> const& cookie, MirKeyboardAction action, xkb_keysym_t key_code,
+        int scan_code, MirInputEventModifiers modifiers);
+
+    // Touch event
+    mir::EventUPtr make_event(MirInputDeviceId device_id, std::chrono::nanoseconds timestamp,
+        std::vector<uint8_t> const& mac, MirInputEventModifiers modifiers);
```
AlanGriffiths, line 75: I know the mirclient API has this horrible abuse of overloading `make_event()`, but I don't think it right to repeat the mistake in miroil. Would it be painful in the QtMir code to have `make_touch_event()`, `make_key_event()`, etc?

erlend-g: Yes, that sound like a good idea and is quick to do. So I'll change the names.

erlend-g: Fixed that, and talk to Marius Gripsgard about branches... So https://github.com/ubports/qtmir/tree/xenial_-_edge_-_wayland_-_x11it the correct one to use for now.

AlanGriffiths (approved):

Issue #1616, opened 2020-07-20
When run under GNOME, `miral-app -kiosk` exits immediately


AlanGriffiths:
The script launches `miral-kiosk` and then the terminal, waits for the terminal to exit and closes `miral-kiosk`.

But `gnome-terminal` is a script that launches the terminal and then exits immediately, which means `miral-app` also exits immediately.

PR #1945, opened 2021-02-24
[miral-terminal] wait for gnome-terminal-server to exit

AlanGriffiths:
When we start gnome-terminal-server, we also need to wait for it to exit.

(Fixes: #1616)


wmww (approved):
LGTM
bors r+

PR #1939, opened 2021-02-18
cmake: Fix running WLCS tests under ptest

RAOF:
This fixes running the WLCS tests under ptest, rather than silently(!) failing to run them and reporting success.

It also implements the ptest support for checking that the XFAIL tests actually fail.

PR #1947, opened 2021-02-25
cmake: Fix running WLCS tests under ptest

RAOF:
This fixes running the WLCS tests under ptest, rather than silently(!) failing to run them and reporting success.

It also implements the ptest support for checking that the XFAIL tests actually fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,6 +247,10 @@ CMAKE_DEPENDENT_OPTION(
   MIR_ENABLE_WLCS_TESTS "Also Build WLCS tests" ON
   "MIR_ENABLE_TESTS" OFF
 )
+CMAKE_DEPENDENT_OPTION(
+  MIR_RUN_WLCS_TESTS "Run WLCS tests as part of testsuite" ON
+  "MIR_ENABLE_WLCS_TESTS" OFF
+)
 
 foreach(platform IN LISTS MIR_PLATFORM)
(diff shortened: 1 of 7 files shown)


RAOF:
See also #1939, which is orphaned since its dependent branch got merged.

RAOF:
Once WLCS has built in the mir-team/dev PPA this should be landable.

RAOF:
bors try

RAOF:
Bah! Why does `BadBufferTest` fail in the CI environment?!

RAOF:
So, two problems:
 * something is strange in the non-sbuild CI environment around shared memory (maybe the place we allocate shm files in is a real fs, rather than a tmpfs there?)
 * The BadBuffer test which tests that a SIGSEGV is caught fails on sbuild:arm64 - this would seem to be a qemu-user-static bug, as the test passes on a real arm64 RPi 4.

We solve this by adding a configure option to mark that test as XFAIL, and setting it in the non-sbuild Ubuntu tasks and on the arm64 sbuild task.

AlanGriffiths (approved):
bors r+

RAOF:
Damnit! We've broken some tests between my last push and bors' attempted merge :(

RAOF:
#1988 is the PR that broke the tests.

AlanGriffiths:
Now #1993 has landed, this should be unblocked

bors r+

PR #1948, opened 2021-02-25
Don't give keyboard focus to popups

### LINKED ISSUE #1818: Popups are not dismissed when clicking again in the same place

### LINKED ISSUE #1941: Qt popup menus sometimes disappear when navigating submenus

wmww:
The spec is not super clear to me on where keyboard focus should go with regards to popups, but other compositors including Weston do **not** give keyboard focus to popups. GTK handles both fine, but Qt has issues in edge cases when popups get input focus (#1941). Draft because it does not yet play nice with the proposed popup WLCS in https://github.com/MirServer/wlcs/pull/190 (specifically, it doesn't dismiss popups when new surfaces are created). I'm also not sure if this is the correct place to put this code.

Fixes #1941, Fixes #1818. Better alternative to #1942

### DIFF
--- a/src/include/server/mir/shell/abstract_shell.h
+++ b/src/include/server/mir/shell/abstract_shell.h
@@ -170,6 +170,7 @@ private:
     std::weak_ptr<scene::Session> focus_session;
     std::weak_ptr<scene::Surface> notified_focus_surface;
     std::shared_ptr<scene::SurfaceObserver> focus_surface_observer;
+    std::vector<std::weak_ptr<scene::Surface>> popups_of_focused_surface;
 
     void notify_focus_locked(
         std::unique_lock<std::mutex> const& lock,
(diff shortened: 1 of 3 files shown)


wmww:
This now passes all the tests in https://github.com/MirServer/wlcs/pull/190. Still might not be the right direction.

AlanGriffiths:
> birs r+

bors r+

PR #1951, opened 2021-02-26
Miroil add input device observer

erlend-g:
In Qtmir we have the class: qtmir:MirServerHooks which contains an input device observer. This class is no long available in qtmir and in addition the class MirInputDeviceObserver needs a lot of functions that are now internal to mir (Used to inquire about mir::Input::Device capabilities & appling keymap )

So I have made a new class in miroil.. That function as a input device observer(abstract class) and contains the "missing" mir functions. 

Similar to what I did for the event builder.


include/miroil/miroil/input_device_observer.h
```diff
@@ -0,0 +1,43 @@
...
+ * ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+ * SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef MIROIL_INPUT_DEVICE_OBSERVER_H
+#define MIROIL_INPUT_DEVICE_OBSERVER_H
+#include <mir_toolkit/mir_input_device_types.h>
```
AlanGriffiths, line 18: This header isn't available downstream: we shouldn't be reintroducing the problem Miroil is here to solve.

AlanGriffiths: Sorry, my confusion. This is fine

include/miroil/miroil/input_device_observer.h
```diff
@@ -0,0 +1,43 @@
...
+    virtual ~InputDeviceObserver();
+    
+    void             apply_keymap(const std::shared_ptr<mir::input::Device> &device, const std::string & layout, const std::string & variant);
+    MirInputDeviceId get_device_id(const std::shared_ptr<mir::input::Device> &device);
+    std::string      get_device_name(const std::shared_ptr<mir::input::Device> &device);
+    bool             is_keyboard(const std::shared_ptr<mir::input::Device> &device);
+    bool             is_alpha_numeric(const std::shared_ptr<mir::input::Device> &device);
+    
+    virtual void device_added(const std::shared_ptr<mir::input::Device> &device) = 0;
+    virtual void device_removed(const std::shared_ptr<mir::input::Device> &device) = 0;
```
AlanGriffiths, line 38: This combination of non-virtual and virtual member functions give an impression of a class fulfilling too many roles.

I'd prefer a wrapper for `std::shared_ptr<mir::input::Device>` that has these property methods:

```c++
class InputDevice
{
public:
  ...
    void apply_keymap(std::string const& layout, std::string const& variant);
    auto device_id() const -> MirInputDeviceId;
    auto get_device_name() const -> std::string;
    bool is_keyboard() const;
    bool is_alpha_numeric() const;

private:
    std::shared_ptr<mir::input::Device> const device;
}; 
```

include/miroil/miroil/input_device_observer.h
```diff
@@ -0,0 +1,43 @@
...
+namespace mir { namespace input { class Device; } }
+
+namespace miroil
+{
+class InputDeviceObserver
+{
+public:
+    virtual ~InputDeviceObserver();
+    
+    void             apply_keymap(const std::shared_ptr<mir::input::Device> &device, const std::string & layout, const std::string & variant);
```
AlanGriffiths, line 31: Mir style has `const` "on the right" and "&" or "*" with the type. Vis:
```c++
    void             apply_keymap(std::shared_ptr<mir::input::Device> const& device, std::string const& layout, std::string const& variant);
```

AlanGriffiths (changes requested):
Nice, but I think it can be improved.

Thanks

include/miroil/miroil/input_device_observer.h
erlend-g: Okey, I'll split the class in two and move the const.

include/miroil/miroil/input_device.h
```diff
@@ -0,0 +1,44 @@
...
+#include <string>
+
+namespace mir { namespace input { class Device; } }
+
+namespace miroil
+{
+class InputDevice
+{
+public:
+    InputDevice(std::shared_ptr<mir::input::Device> const & device);
```
AlanGriffiths, line 29: The "&" should be with the type:
```c++
... const& device
```

include/miroil/miroil/input_device.h
```diff
@@ -0,0 +1,44 @@
...
+    ~InputDevice();
+    
+    void             apply_keymap(std::string const & layout, std::string const & variant);
+    MirInputDeviceId get_device_id();
+    std::string      get_device_name();
+    bool             is_keyboard();
+    bool             is_alpha_numeric();
+
+private:
+    std::shared_ptr<mir::input::Device> const & device;
```
AlanGriffiths, line 39: This should not be a reference.

include/miroil/miroil/input_device_observer.h
```diff
@@ -0,0 +1,35 @@
...
+#include "input_device.h"
+#include <memory>
+#include <string>
+
+namespace miroil
+{
+class InputDeviceObserver
+{
+public:
+    virtual ~InputDeviceObserver() {};
```
AlanGriffiths, line 27: With, what is now an interface class it better to `delete` copy/assign too. Vis:
```c++
InputDeviceObserver(InputDeviceObserver const&) = delete;
InputDeviceObserver& operator=(InputDeviceObserver const&) = dellete;
```

include/miroil/miroil/input_device_observer.h
```diff
@@ -0,0 +1,35 @@
...
+    
+    virtual void device_added(std::shared_ptr<miroil::InputDevice> const & device) = 0;
```
AlanGriffiths, line 29: Can we just pass InputDevice by value?

AlanGriffiths (changes requested):
Reference data members are usually wrong, and is wrong in this case. The rest is minor.

include/miroil/miroil/input_device.h
erlend-g: Okey, I was not sure how "shared_ptr const" worked... I though it might not increase the use_count() since it is a const... 

But it does (so i can remove the reference).

include/miroil/miroil/input_device_observer.h
erlend-g: Yes

include/miroil/miroil/input_device_observer.h
```diff
@@ -0,0 +1,38 @@
...
+class InputDeviceObserver
+{
+public:
+    InputDeviceObserver() {};
+    InputDeviceObserver(InputDeviceObserver const&) = delete;
+    InputDeviceObserver& operator=(InputDeviceObserver const&) = delete;    
+    virtual ~InputDeviceObserver() {};
+    
+    virtual void device_added(std::shared_ptr<miroil::InputDevice> device) = 0;
+    virtual void device_removed(std::shared_ptr<miroil::InputDevice> device) = 0;
```
AlanGriffiths, line 33: Sorry, I meant...
```c++
    virtual void device_added(InputDevice device) = 0;
    virtual void device_removed(InputDevice device) = 0;
```

erlend-g: Okey, I wasn't sure what you meant.

AlanGriffiths (approved):
bors merge

PR #1952, opened 2021-03-07
adding PromptSessionListener & PromptSessionManager

erlend-g:
Qtmir uses mir::scene::PromptSessionManager which is no long available, so I've created a wrapper for this class. This is used many places in QtMir.

In addition qtmir listens to mir::scene::PromptSessionListener (also not available)... so I've created an abstract class for that.

This is used in:
https://github.com/ubports/qtmir/blob/xenial/src/platforms/mirserver/mirserverhooks.cpp

class PromptSessionListenerImpl

Which can't be solved in this way because it contains both internal mir stuff and qt stuff. So it has to be split into a mir part(miroil) and Qt part(Qtmir).

### DIFF
--- a/include/miroil/miroil/prompt_session_listener.h
+++ b/include/miroil/miroil/prompt_session_listener.h
@@ -1,0 +1,49 @@
+/*
+ * Copyright © 2016-2021 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIROIL_PROMPT_SESSION_LISTENER_H
+#define MIROIL_PROMPT_SESSION_LISTENER_H
+#include <memory>
+
+namespace mir { namespace scene { class PromptSession; } }
+namespace mir { namespace scene { class Session; } }
+
+namespace miroil {
+    
+class PromptSessionListener
+{
+public:
+    virtual ~PromptSessionListener();
+    
+    PromptSessionListener& operator=(PromptSessionListener const&) = delete;    
+
+    virtual void starting(std::shared_ptr<mir::scene::PromptSession> const& prompt_session) = 0;
+    virtual void stopping(std::shared_ptr<mir::scene::PromptSession> const& prompt_session) = 0;
+    virtual void suspending(std::shared_ptr<mir::scene::PromptSession> const& prompt_session) = 0;
+    virtual void resuming(std::shared_ptr<mir::scene::PromptSession> const& prompt_session) = 0;
+    virtual void prompt_provider_added(mir::scene::PromptSession const& prompt_session,
+                               std::shared_ptr<mir::scene::Session> const& prompt_provider) = 0;
+    virtual void prompt_provider_removed(mir::scene::PromptSession const& prompt_session,
+                                 std::shared_ptr<mir::scene::Session> const& prompt_provider) = 0;
+                                 
+protected:
+    PromptSessionListener() = default;
+    PromptSessionListener(PromptSessionListener const&) = delete;
+};
+
+}
+
+#endif // MIROIL_PROMPT_SESSION_LISTENER_H
--- a/include/miroil/miroil/prompt_session_manager.h
+++ b/include/miroil/miroil/prompt_session_manager.h
@@ -38,13 +38,13 @@ public:
 
     bool operator==(PromptSessionManager const& other);
 
-    auto applicationFor(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const -> miral::Application;
-    void resumePromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
-    void stopPromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
-    void suspendPromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;    
+    auto application_for(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const -> miral::Application;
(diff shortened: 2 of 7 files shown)


AlanGriffiths:
@erlend-g thanks for plugging away at this. I will review the PR later.

In the meantime, it would be easier to track this work and identify problems if you share the corresponding changes to QtMir.

I've set up a project for both sides of this work:

https://forums.ubports.com/topic/5822/miroil-for-lomiri/1

include/miroil/miroil/prompt_session_manager.h
```diff
@@ -0,0 +1,51 @@
...
+
+    auto operator=(PromptSessionManager const& src) -> PromptSessionManager&;
+    auto operator=(PromptSessionManager&& src) -> PromptSessionManager&;
+
+    bool operator==(PromptSessionManager const& other);
+
+    auto applicationFor(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const -> miral::Application;
+    void resumePromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
+    void stopPromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
+    void suspendPromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
```
AlanGriffiths, line 44: The parameter names do not match Mir's style. They should be `snake_case`.

include/miroil/miroil/prompt_session_manager.h
```diff
@@ -0,0 +1,51 @@
...
+
+    bool operator==(PromptSessionManager const& other);
+
+    auto applicationFor(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const -> miral::Application;
+    void resumePromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
+    void stopPromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
+    void suspendPromptSession(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;    
+
+private:
+    std::shared_ptr<mir::scene::PromptSessionManager> m_promptSessionManager;
```
AlanGriffiths, line 47: The member variable name does not match Mir's style. It should be `snake_case` without the `m_` wart.

AlanGriffiths (changes requested):

include/miroil/miroil/prompt_session_listener.h
```diff
@@ -0,0 +1,49 @@
...
+    PromptSessionListener& operator=(PromptSessionListener const&) = delete;    
+
+    virtual void starting(std::shared_ptr<mir::scene::PromptSession> const& prompt_session) = 0;
+    virtual void stopping(std::shared_ptr<mir::scene::PromptSession> const& prompt_session) = 0;
+    virtual void suspending(std::shared_ptr<mir::scene::PromptSession> const& prompt_session) = 0;
+    virtual void resuming(std::shared_ptr<mir::scene::PromptSession> const& prompt_session) = 0;
+    virtual void prompt_provider_added(mir::scene::PromptSession const& prompt_session,
+                               std::shared_ptr<mir::scene::Session> const& prompt_provider) = 0;
+    virtual void prompt_provider_removed(mir::scene::PromptSession const& prompt_session,
+                                 std::shared_ptr<mir::scene::Session> const& prompt_provider) = 0;
```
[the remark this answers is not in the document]
erlend-g: Well, actually nobody is looking inside the mir::scene::PromptSession object. So therefore I concluded that it is not necessary with a wrapper for that object... To keep miroil as lean...

But there is a class qtmir::PromptSession... So I can move that into miroil if you like? Or we can live it as is.

include/miroil/miroil/prompt_session_manager.h
erlend-g: It's a little easy to be blinded by qtmir which uses CamelCase... Fixed now.

include/miroil/miroil/prompt_session_listener.h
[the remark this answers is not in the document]
AlanGriffiths: OK. I might still have used a wrapper (of `std::shared_ptr<void>` to avoid introducing `mir::scene::Session`), but this works.

include/miroil/miroil/prompt_session_manager.h
```diff
@@ -0,0 +1,51 @@
...
+
+    auto operator=(PromptSessionManager const& src) -> PromptSessionManager&;
+    auto operator=(PromptSessionManager&& src) -> PromptSessionManager&;
+
+    bool operator==(PromptSessionManager const& other);
+
+    auto application_for(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const -> miral::Application;
+    void resume_prompt_session(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
+    void stop_prompt_session(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
+    void suspend_prompt_session(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
```
AlanGriffiths, line 44: snake_case for parameters

include/miroil/miroil/prompt_session_manager.h
```diff
@@ -0,0 +1,51 @@
...
+
+    bool operator==(PromptSessionManager const& other);
+
+    auto application_for(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const -> miral::Application;
+    void resume_prompt_session(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
+    void stop_prompt_session(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;
+    void suspend_prompt_session(std::shared_ptr<mir::scene::PromptSession> const& promptSession) const;    
+
+private:
+    std::shared_ptr<mir::scene::PromptSessionManager> promptSessionManager;
```
AlanGriffiths, line 47: snake_case for member variables

AlanGriffiths (approved):
More style nits

AlanGriffiths (changes requested):
Oops!

erlend-g:
Fixed the snake_case's

AlanGriffiths (approved):
Good enough.

bors r+

include/miroil/miroil/prompt_session_manager.h
AlanGriffiths: > I can look into checking in the qtmir stuff as well... But it's a little complicated since nothing compiles.

That's why I set up a fork based on Mir 1.x - everything compiles.

See this forum post:

https://forums.ubports.com/topic/5822/miroil-for-lomiri/1

PR #1953, opened 2021-03-09
Workaround spurious valgrind error

AlanGriffiths:
Valgrind reports a spurious "Invalid write of size 8" beyond the end of `dummy_destination`, extending the test buffer avoids this.


AlanGriffiths:
Test case:
```
valgrind $(pkg-config --variable test_runner wlcs) cmake-build-debug/lib/miral_wlcs_integration.so --gtest_filter=BadBufferTest.test_truncated_shm_file
```

wmww (approved):
I'd normally be somewhat curious what the root cause is, but it's just a test stub so who cares

bors+

RAOF (approved):
Yup!

This doesn't fix the malloc corruption I see when running this test repeatedly, but avoiding a valgrind splat is sensible.

bors r+

PR #1955, opened 2021-03-12
Kiosk: don't allow window to move

### LINKED ISSUE #1954: Kiosk moves window when user attempts to unmaximize it
To reproduce:
1. Open window (such as LXTerminal) in kiosk
2. Double click on decorations to unmaximize

Expected:
- Nothing

Actual:
- Window does not become unmaximized (it's not movable by the user), but it does move partly off the screen.

wmww:
Fixes #1954


AlanGriffiths (approved):
Weird that this is necessary. How is a position being requested?

bors r+

wmww:
My guess is `BasicWindowManager` is requesting the window be unmaximized, be smaller and be moved, and we're wiping out all aspects of that request except the move.

PR #1956, opened 2021-03-12
Kiosk: fullscreen apps

wmww:
We probably don't want to merge this as-is (if for backwards compat if nothing else). Do we want to make this an option though?

### DIFF
--- a/examples/miral-kiosk/kiosk_window_manager.cpp
+++ b/examples/miral-kiosk/kiosk_window_manager.cpp
@@ -138,7 +138,7 @@ auto KioskWindowManagerPolicy::place_new_window(ApplicationInfo const& app_info,
     if ((specification.type() == mir_window_type_normal || specification.type() == mir_window_type_freestyle) &&
         (!specification.parent().is_set() || !specification.parent().value().lock()))
     {
-        specification.state() = mir_window_state_maximized;
+        specification.state() = mir_window_state_fullscreen;
         specification.top_left() = mir::optional_value<Point>{}; // Ignore requested position (if any) when we maximize
         specification.size() = mir::optional_value<Size>{}; // Ignore requested size (if any) when we maximize
         tools.place_and_size_for_state(specification, WindowInfo{});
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
There are two independent changes here.

1. Dropping position requests. This ought to be a no-op as clients don't have a way to request positions. Fine to land it, but why is it necessary?
2. Changing the window state to fullscreen. I've thought about doing this before, but not wanted to break compatibility. We ought to test with "many" clients, and possibly provide a backward compatibility option.

Saviq:
> Changing the window state to fullscreen. I've thought about doing this before, but not wanted to break compatibility. We ought to test with "many" clients, and possibly provide a backward compatibility option.

I'm also unsure about this. I kinda liked the approach we had with i3 where you could configure the window manager to keep a surface fullscreen by matching some properties.

AlanGriffiths:
> Like, it could be good to support multiple apps connecting and placing them on preconfigured outputs.

That is easy to implement, but is `miral-kiosk` the right program to have that functionality?

wmww:
Matching app IDs would be a bit of logic and added complexity. Might be difficult for users to use if they don't know about app IDs or don't know what the app ID of their app is. I think the more straightforward option would be to make it a command line argument, and apply to all apps. This should cover most use-cases, and be relatively simpler. Is that worth implementing?

AlanGriffiths:
I strongly suspect that most "kiosk" apps that care already request fullscreen from the client side and will be unaffected by the proposed change. But we don't (and can't) know all our users, so having a backward-compatibility option is probably the way forward.

I think i3 type layout rules is beyond what mir-kiosk is about, it is more akin to the tiling window manager in miral-shell.

Saviq:
> I strongly suspect that most "kiosk" apps that care already request fullscreen from the client side and will be unaffected by the proposed change. But we don't (and can't) know all our users, so having a backward-compatibility option is probably the way forward.
> 
> I think i3 type layout rules is beyond what mir-kiosk is about, it is more akin to the tiling window manager in miral-shell.

There is one use case I had in mind and wondered where does it fit - multiple display( group)s, and a single (preconfigured) app on each.

AlanGriffiths:
> > 
> 
> There is one use case I had in mind and wondered where does it fit - multiple display( group)s, and a single (preconfigured) app on each.

Stick it in the draft roadmap?

Saviq:
> Stick it in the draft roadmap?

Sure, just wanted to mention what could still fit in mir-kiosk after all.

AlanGriffiths:
I think this has been superceeded by ubuntu-frame?

PR #1957, opened 2021-03-13
Add Alpine Linux CI

z3ntu:
Based on #1949 , thanks for the "template"!

Commit f4d0dd5979cc36c64c76ae7f8ac7eebb7c6a7c53 can be upstreamed as-is imo.

A couple of problems:
* https://github.com/MirServer/mir/commit/e3d9b53512fdd73d81254b219af5cea71e471431
* The tests fail
* Alpine 3.13 won't really be possible (at the moment) because some of the dependencies are only in the testing repo and that repo is only available on the edge version and not on the stable version. I'll see that all dependencies will be included in the Alpine 3.14 release

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -313,9 +313,13 @@ list(APPEND CMAKE_REQUIRED_LIBRARIES ${DRM_LIBRARIES})
 # USDT integration. The integration only involves #include-ing a
 # SystemTap header and calling a macro from it in the LTTNG macro,
 # but there aren't guarantees this won't change in future.
-add_definitions(  # Drop in favour of add_compile_definitions when we drop 16.04
-  -DLTTNG_UST_HAVE_SDT_INTEGRATION
-)
+include(CheckIncludeFile)
+CHECK_INCLUDE_FILE(sys/sdt.h HAVE_SYS_SDT_H)
+if (HAVE_SYS_SDT_H)
+  add_definitions(  # Drop in favour of add_compile_definitions when we drop 16.04
+    -DLTTNG_UST_HAVE_SDT_INTEGRATION
+  )
+endif()
 
 pkg_check_modules(GL REQUIRED glesv2)
 
--- a/spread/build/alpine/task.yaml
+++ b/spread/build/alpine/task.yaml
@@ -1,0 +1,43 @@
+summary: Build (on Alpine)
+systems: [alpine-*]
+variants: [amd64]
+
+execute: |
+    apk add \
+        boost-dev \
+        capnproto-dev \
+        cmake \
+        coreutils \
+        elogind \
+        eudev-dev \
+        freetype-dev \
+        g++ \
+        glib-dev \
+        glm-dev \
+        glog-dev \
+        grep \
+        gtest-dev \
+        libepoxy-dev \
+        libevdev-dev \
+        libinput-dev \
+        libxcursor-dev \
+        libxkbcommon-dev \
+        libxml++-2.6-dev \
+        lttng-ust-dev \
+        lttng-ust-tools \
+        make \
+        mesa-dev \
+        nettle-dev \
+        py3-dbusmock \
+        py3-pillow \
+        umockdev-dev \
+        wayland-dev \
+        yaml-cpp-dev
+
+    BUILD_DIR=$(mktemp -d)
+    cmake -H$SPREAD_PATH -B$BUILD_DIR -DCMAKE_BUILD_TYPE=Debug -DMIR_USE_LD=ld -DMIR_ENABLE_WLCS_TESTS=OFF
+    export VERBOSE=1
+    cmake --build $BUILD_DIR -- -j`nproc`
+
+    export XDG_RUNTIME_DIR=/tmp
+    cmake --build $BUILD_DIR -- ptest
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -19,60 +19,42 @@ jobs:
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
 
-    # Only run if we have access to secrets.
-    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
-
     steps:
     - id: set-matrix
       name: Determine BuildAndTest matrix
       run: |
         set -euo pipefail
 
-        if [ "${{ github.event_name }}" == "push" -a "${{ github.ref }}" == "refs/heads/master" ]; then
-          cat <<EOF | jq -c . | awk '{ print "::set-output name=matrix::" $0 }'
-        {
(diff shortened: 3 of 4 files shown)


AlanGriffiths:
> Commit [f4d0dd5](https://github.com/MirServer/mir/commit/f4d0dd5979cc36c64c76ae7f8ac7eebb7c6a7c53) can be upstreamed as-is imo.

Done: #1958

Newbytee:
> Alpine 3.13 won't really be possible (at the moment) because some of the dependencies are only in the testing repo and that repo is only available on the edge version and not on the stable version.

I asked the maintainers about this a while ago and it is okay to add new packages to stable releases apparently, just something that's not often done. So it shouldn't be necessary to wait for Alpine 3.14 if you're up to adding them.

Saviq:
@z3ntu we can start without `3.13`, looks like `edge` worked OK now?

Once `3.14` gets released, we can add that in again.

z3ntu:
I need some advice what to do about commit e3d9b53 and what to do about the tests. A compile-only check is definitely better than nothing but of course having the tests running would be even better.

Alpine stable can be added later once the deps are in stable (or I'll try to get them into the current stable).

AlanGriffiths:
> I need some advice what to do about commit [e3d9b53](https://github.com/MirServer/mir/commit/e3d9b53512fdd73d81254b219af5cea71e471431)

That commit looks like it could be updated to check for `sys/sdt.h` and conditionally set the define. (We've also dropped 16.04, so it could switch to `add_compile_definitions()` as suggested in the comment.) That could then be upstreamed.

>  and what to do about the tests. A compile-only check is definitely better than nothing but of course having the tests running would be even better.

I've not had time to look into the tests. I think we should make the first goal successful compilation and linking without the tests and only then start working through the test failures.

Is it just the test code failing on Alpine? The production code actually work so far as you can tell?

Saviq:
Hi @z3ntu, any update on this?

z3ntu:
Not really directly but I've opened MRs in Alpine for moving the missing dependencies from testing repository to community so they will be included in the next stable release
* https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/21105
* https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/21106
* https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/21107
* https://gitlab.alpinelinux.org/alpine/aports/-/commit/7538494eb06dbe3c14b7a10f43665129e23b36e7

z3ntu:
@Saviq Actually one question, what should I do with the spread changes from this PR, because it conflicts with https://github.com/MirServer/mir/commit/1cc3754a4df7dbada1a5de8a6d35100bfe5eefcf and I have no clue how and where this spread stuff comes from and if that version includes the Alpine things

Saviq:
bors try

Saviq:
@z3ntu I tweaked the `Spread` workflow so it runs the stuff it can from a fork.

z3ntu:
Thanks, I was about to ask :ok_hand:

Saviq:
@z3ntu we're going in the right direction, down to some failing tests on 3.13 :)

The issue seems to be `XDG_RUNTIME_DIR`, it was unset and setting it to `/tmp` results in:
```plain
unable to lock lockfile /tmp/wayland-0.lock, maybe another compositor is running
```

Maybe there's a better dir to use on Alpine? On Ubuntu it's usually `/run/user/$UID`, for example.

Saviq:
> Btw why does spread not display some of the output?
> 
> Would be quite useful to have the full log

It does at the end, when a task fails.

z3ntu:
I think it's best to merge this without running the tests on Alpine so we at least have the compile-time verification active which is already quite useful. If you agree, then I'll clean up this PR to get it into a nice shape.
The Alpine CI is also now running on both Alpine edge and Alpine stable v3.14 (released 2 days ago)

AlanGriffiths:
@z3ntu I'm in favour of landing anything that helps, but can we pause for a day or two?

It looks like the test problem would be significantly reduced following #2068 and I'd like to see what's left before merging with tests disabled. (I can see there will still be failures, but many of the failing tests are obsolete.)

Saviq:
Let's see how it would fare:

bors try

Saviq:
@z3ntu my bad, have fixed to stick with `allow-fail`.

bors try

Saviq:
I think we need a better `XDG_RUNTIME_DIR`:

```
dbus[56137]: Unable to set up transient service directory: XDG_RUNTIME_DIR "/tmp" can be written by others (mode 041777)
```

z3ntu:
Hey, the build incl. tests passed now I think (although the log in the CI stops at 94%/95%...?)

z3ntu:
I think this could be merged now, even before #2068

z3ntu:
Ok basing this on #2068 makes it succeed again (I think, not 100% because log output is horribly unusable), so let's wait until that's merged

Saviq:
@z3ntu I changed the base so that the diff reflects the changes in this PR (rather than also include #2068).

Saviq:
bors try

spread/build/alpine/task.yaml
```diff
@@ -0,0 +1,44 @@
...
+        libinput-dev \
+        libxcursor-dev \
+        libxkbcommon-dev \
+        libxml++-2.6-dev \
+        lttng-ust-dev \
+        lttng-ust-tools \
+        make \
+        mesa-dev \
+        nettle-dev \
+        protobuf-dev \
```
AlanGriffiths, line 31: #2068 removes the need for protobuf, that can be dropped.

CMakeLists.txt
```diff
@@ -313,9 +313,13 @@ list(APPEND CMAKE_REQUIRED_LIBRARIES ${DRM_LIBRARIES})
...
 # but there aren't guarantees this won't change in future.
-add_definitions(  # Drop in favour of add_compile_definitions when we drop 16.04
-  -DLTTNG_UST_HAVE_SDT_INTEGRATION
-)
+include(CheckIncludeFile)
+CHECK_INCLUDE_FILE(sys/sdt.h HAVE_SYS_SDT_H)
+if (HAVE_SYS_SDT_H)
+  add_definitions(  # Drop in favour of add_compile_definitions when we drop 16.04
+    -DLTTNG_UST_HAVE_SDT_INTEGRATION
+  )
```
AlanGriffiths, line 321: Pre-existing (and not a blocker) but like the comment says `add_compile_definitions()` would be better:
```cmake
add_compile_definitions(LTTNG_UST_HAVE_SDT_INTEGRATION)
```

AlanGriffiths (approved):
Looks sane and appears to work fine.

CMakeLists.txt
z3ntu: I'll do in a different PR

Issue #1959, opened 2021-03-22
Cursor movement is not propagating to clients correctly when zwp_relative_pointer_manager_v1 is enabled


AlanGriffiths:
Cursor movement is not propagating to X clients correctly when `zwp_relative_pointer_manager_v1` is enabled. This manifests in a number of ways, e.g. using menus. But the most obvious is to open x-eyes:

Expect: eyes follow cursor when it is over any X client window
Actual: eyes follow cursor only at the points it moves onto any X client window

AlanGriffiths:
This is weird. I see this consistently using egmde (both snap and local, native build). But I don't see it with miral-shell.

AlanGriffiths:
OK, so the difference is `zwp_relative_pointer_manager_v1`, if I disable this in egmde then pointer with X applications behave correctly. (I haven't dug into what Xwayland is doing with this protocol, or if we're handling it correctly - I suspect not.)

AlanGriffiths:
OK, this is wrong:

```c++
void mf::WlPointer::enter_or_motion(MirPointerEvent const* event, WlSurface& root_surface)
...
    else if (!relative_pointer && position_on_target != current_position)
    {
        send_motion_event(...
```
We shouldn't be supressing motion events just because there's a `relative_pointer`.

PR #1960, opened 2021-03-22
Do not suppress motion events when relative motion is used

AlanGriffiths:
Fixes: #1959


wmww (approved):
LGTM, and I've confirmed it fixes the problem with xeyes

PR #1961, opened 2021-03-22
Geometry value type conversions

wmww:
This allows explicitly constructing the 2D geometry types (point, displacement and size) from the same sorts of vectors with different value types (ie you can make a `Size` from a `SizeF`). It also makes it impossible to assign wrapped values with different value types, so an explicit conversion is required.


wmww:
Hmm... builds on my machine

RAOF (approved):
Seems to do what it says on the tin, and seems a reasonable set of conversions.

PR #1965, opened 2021-03-23
Check for locked cursor before sending pointer movement

AlanGriffiths:
The changes in #1960 led to clients getting spurious pointer movement events (e.g. mir-kiosk-neverputt ends up with jitters). We shouldn't be sending those if the pointer is locked.


wmww (approved):
LGTM

bors r+

PR #1970, opened 2021-03-26
Gracefully handle XWayland errors

### LINKED ISSUE #1969: Mir hangs on shutdown if XWayland failed to start
If XWayland fails at startup, Mir does not print an error and hangs on shutdown. It seems to be trying to connect xcb to the socket, which hangs indefinitely because the server isn't coming. We should detect that there's a problem, print an error and close the fd so the window manager also fails.

I don't have a great way to reproduce this. I noticed it in the mate-wayland snap, and I'm testing by adding
```
setenv("LD_LIBRARY_PATH", "/snap/mate-wayland/current/usr/lib/x86_64-linux-gnu:/snap/mate-wayland/current/lib/x86_64-linux-gnu", 1);
```
just above `exec_xwayland()` in `xwayland_server.cpp`

wmww:
Our logic for launching XWayland has long been good but not great. #1969 is a symptom of this, and while looking into it I found other symptoms (notably when XWayland fails to start apps *also* hang for a similar but unrelated reason). This PR attempts to clean it up, show errors when needed and prevent deadlocks. Fixes #1969.

Notable changes:
- Pull in the main loop so `ThreadedDispatcher`s can be destroyed inside their own calls. The previous method of creating and joining a new thread was heavy, looked weird and didn't seem to even work in all cases.
- Use `std::enable_shared_from_this` with `XWaylandConnector`--this is because we need a `weak_ptr` when punting to the main loop in case the object dies.
- Move creation of socket pairs to `XWaylandServer`. This prevents them from being held in `XWaylandConnector::spawn()` after the server was forked, which was causing #1969
- Make error messages more correct (for example it previously said "killing XWayland" when it was really restarting it)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_connector.cpp
+++ b/src/server/frontend_xwayland/xwayland_connector.cpp
@@ -170,15 +170,18 @@ void mf::XWaylandConnector::spawn()
                 // NOTE: we do not stop the spawner, so the server/wm will be recreated when new clients connect
                 auto local_wm{std::move(wm)};
                 auto local_server{std::move(server)};
-                auto local_wm_event_thread{std::move(wm_event_thread)};
+                // This lambda is called by the window manager event dispatcher. It is a runtime error to destroy a
+                // ThreadedDispatcher from it's own call, so we destroy it on the main loop thread.  We turn it into a
+                // shared_ptr because unique_ptrs can not be captured in lambdas.
+                main_loop->spawn(
+                    [wm_event_thread=std::shared_ptr<md::ThreadedDispatcher>{wm_event_thread.release()}]()
+                    {
+                    });
 
                 lock.unlock();
 
                 local_wm.reset();
                 local_server.reset();
-
-                // We can't destroy a ThreadedDispatcher from inside a call it made, so do it from another thread
-                std::thread{[&](){ local_wm_event_thread.reset(); }}.join();
             });
         server = std::make_unique<XWaylandServer>(
             wayland_connector,
--- a/src/server/frontend_xwayland/xwayland_connector.cpp
+++ b/src/server/frontend_xwayland/xwayland_connector.cpp
@@ -141,17 +141,13 @@ void mf::XWaylandConnector::spawn()
 
     try
     {
-        auto const wayland_socket_pair = XWaylandServer::make_socket_pair();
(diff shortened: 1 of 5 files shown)


src/server/frontend_xwayland/xwayland_connector.cpp
```diff
@@ -126,9 +115,43 @@ auto mf::XWaylandConnector::socket_name() const -> optional_value<std::string>
...
+    auto local_wm_event_thread{std::move(wm_event_thread)};
+    auto local_wm{std::move(wm)};
+    auto local_server{std::move(server)};
+
+    lock.unlock();
+
+    local_spawner.reset();
+    local_wm_event_thread.reset();
+    local_wm.reset();
+    local_server.reset();
```
AlanGriffiths, line 149: These area about to go out of scope, so resetting is redundant. (And *if*, for some dubious reason, the order matters, they can be defined in a suitable order.)

AlanGriffiths (changes requested):
Seems to work and looks more reasonable than the existing code.

However, there's something a bit odd about the way `maybe_create_spawner()` and `clean_up()` handle locks.

src/server/frontend_xwayland/xwayland_connector.cpp
wmww: I don't think the order matters, but I moved them around to keep it the same just in case. (current order has at least been tested)

src/server/frontend_xwayland/xwayland_connector.cpp
```diff
@@ -126,9 +115,43 @@ auto mf::XWaylandConnector::socket_name() const -> optional_value<std::string>
...
+    }
+
+    if (is_started && !spawner)
+    {
+        // If we should be running but a spawner does not exist, create one
+        spawner = std::make_unique<XWaylandSpawner>([this]() { spawn(); });
+    }
+}
+
+void mf::XWaylandConnector::clean_up(std::unique_lock<std::mutex>& lock)
```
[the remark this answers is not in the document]
wmww: Done

AlanGriffiths (approved):
LGTM

bors r_

AlanGriffiths:
> bors r_

bors r+

PR #1973, opened 2021-03-31
Use a dark gray background instead of black by default

### LINKED ISSUE #1971: Non-black default screen when no apps open
It's currently difficult to tell when Mir kiosk (or any Mir) is running because the default screen is black. Is there any reason not to set a glClearColor to something more distinctive, so it's clear when a Mir server is running with no windows open?

wmww:
Anything bright or canonical-colored might be annoying, but I see no reason a simple dark gray background would be worse than a black one. This makes it much easier to tell when a Mir server has started without the overhead of rendering a background surface.

Fixes #1971

### DIFF
--- a/src/renderers/gl/renderer.cpp
+++ b/src/renderers/gl/renderer.cpp
@@ -351,7 +351,7 @@ mrg::Renderer::Program::Program(GLuint program_id)
 
 mrg::Renderer::Renderer(graphics::DisplayBuffer& display_buffer)
     : render_target(&display_buffer),
-      clear_color{0.0f, 0.0f, 0.0f, 0.0f},
+      clear_color{0.2f, 0.2f, 0.2f, 1.0f},
       default_program(family.add_program(vshader, default_fshader)),
       alpha_program(family.add_program(vshader, alpha_fshader)),
       program_factory{std::make_unique<ProgramFactory>()},
--- a/tests/unit-tests/renderers/gl/test_gl_renderer.cpp
+++ b/tests/unit-tests/renderers/gl/test_gl_renderer.cpp
@@ -220,10 +220,10 @@ TEST_F(GLRenderer, avoids_src_alpha_for_rgbx_blending)  // LP: #1423462
     renderer.render(renderable_list);
 }
 
-TEST_F(GLRenderer, clears_all_channels_zero)
+TEST_F(GLRenderer, clears_all_channels)
 {
     InSequence seq;
-    EXPECT_CALL(mock_gl, glClearColor(0.0f, 0.0f, 0.0f, 0.0f));
+    EXPECT_CALL(mock_gl, glClearColor(_, _, _, _));
     EXPECT_CALL(mock_gl, glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE));
     EXPECT_CALL(mock_gl, glClear(_));
 


RAOF (approved):
This makes sense to me (but is a missed opportunity to use Xorg's old crosshatch background :wink: )

You'll need to update the tests which check for `glClearColor(0, 0, 0, 0)`, though!

AlanGriffiths:
I can imagine changing this in `mir-kiosk` could be undesirable: it will change the aesthetic during startup. Would a configurable RGB value solve your problem?

wmww:
Hmm, will anyone really care though? I would be inclined to first make the change, then make it configurable if that's requested.

AlanGriffiths:
> Hmm, will anyone really care though? I would be inclined to first change the default, then make it configurable if that's requested.

Quite possibly a paying customer, yes. (E.g. one that needed the orange splash switched off.)

wmww:
I see. Are you ok with changing the default color to dark gray as long as it's configurable? Or is changing the default behavior the issue?

Saviq:
> I see. Are you ok with changing the default color to dark gray as long as it's configurable? Or is changing the default behavior the issue?

We can probably deal with setting the default back to black in the `mir-kiosk` snap, if unset. So yeah, IMO just make it configurable here, and a grey default is OK.

wmww:
> We can probably deal with setting the default back to black in the `mir-kiosk` snap

Mir Kiosk is the main user of Mir that doesn't already have a background. If it is going to be black by default still than IMO making this feature probably isn't worth it. Specifically, I was hoping to help new users who install Mir Kiosk on Ubuntu Core and aren't sure if it's working.

Saviq:
> > We can probably deal with setting the default back to black in the `mir-kiosk` snap
> 
> Mir Kiosk is the main user of Mir that doesn't already have a background. If it is going to be black by default still than IMO making this feature probably isn't worth it. Specifically, I was hoping to help new users who install Mir Kiosk on Ubuntu Core and aren't sure if it's working.

I meant that we can make it black in existing installs (i.e. on refresh, rather than install) to keep the current behaviour.

wmww:
> I meant that we can make it black in existing installs (i.e. on refresh, rather than install) to keep the current behaviour.

I considered that, but it seems to me that would be not be good. Then instead of the change being predictable and easy to notice and test, it happens whenever a user switches machines or otherwise resets their system. This could cause confusion, and could cause them to not notice the change in testing but hit it when they roll out to customers.

AlanGriffiths:
> I meant that we can make it black in existing installs (i.e. on refresh, rather than install) to keep the current behaviour.

That's the worst of all possible behaviours. New installations should behave the same as existing ones.

Saviq:
> That's the worst of all possible behaviours. New installations should behave the same as existing ones.

Well, we're allowed to change defaults, IMO. Another option would be to infer from `show-splash` - if that's `false`, set the background to black.

AlanGriffiths:
> Well, we're allowed to change defaults, IMO. Another option would be to infer from `show-splash` - if that's `false`, set the background to black.

This is beginning to sound like we're talking past each other.

What William and I object to is having a different default on existing installations to new ones. And having `show-splash` change the _background_ is a horrid idea. (Except on April 1st)

If a client tests updating to mir-kiosk version X and sees no problems they may reasonably expect no problems from installing mir-kiosk version X directly.

AlanGriffiths:
I think we'd be good if we provide a configuration option for the fill colour.

That option can be set in the gadget snap for any client that cares about the specific behaviour.

We could also consider not changing the default - that couldn't break anything. And anyone that wants "dark grey" could set the configuration accordingly.

wmww:
The value of this PR is mostly from it being a default. It's particularly useful for people first setting up to notice that something is working, and anyone who particularly cares about the background should already be providing one somehow. I'm not convinced it's worth implementing the option unless we're going to make it something non-black by default.

AlanGriffiths:
The conversation died. Let me summarise where I think we are at:

1. Changing the default background is, in principle, possible. But...
2. It should be possible for mir-kiosk users to retain the black background.
3. The default for mir-kiosk install should not depend on whether it is a new installation or a refresh.

As currently formulated the PR fails 2. If there's further interest we can revisit it.

Issue #1967, opened 2021-03-24
Firefox does not take input in Mir


wmww:
When I open Firefox in Mir, it does not load a page or accept input. Interestingly, it does respond to being resized and properly redraws.

Tested on a number of Mir based shells (including miral-shell and egmde), on both KMS and X11, and with a number of versions of Mir (current master to v2.3.2). I've mostly tested with Firefox running in Wayland. FF in X11 on Mir seems to work at first, but may have additional problems.

The only messages printed to FF's terminal are:
```
wmww@wmwwXPS:~$ firefox 
[GFX1-]: No GPUs detected via PCI
[GFX1-]: glxtest: process failed (received signal 11)
```

AlanGriffiths:
Can you provide clear steps to reproduce?

I've used Firefox on Wayland on (mostly) egmde across 20.04, 18.04, and rawhide. It needs configuring (as documented by https://discourse.ubuntu.com/t/using-firefox-wayland-on-mir-servers/19704) but has been fine for me.

wmww:
To reproduce:
1. Open Firefox 86.0 in Mir on Ubuntu 21.04
2. Observe

wmww:
More interesting notes:
- I somehow had rebrender off, but turning it on didn't seem to help much
- Firefox now eventually starts, after ~30 seconds
- It uses a lot of CPU on all cores (both while in the weird state and once it's running apparently normally)
- Still some strangeness (for example in the right-click menu on a webpage, the hovered-over item doesn't show as selected)

AlanGriffiths:
> * It uses a lot of CPU on all cores (both while in the weird state and once it's running apparently normally)

FWIW I see that on 18.04 (firefox 86.0.1, Wayland on egmde on Mir2.3.3~rc12)

Is this a separate bug? There's no problem with input.

wmww:
Hm, quite possibly, good to know

wmww:
Looks like FF is getting stuck in an endless loop of Wayland roundtrips. In wayland-debug, seing a constant stream of this:
```
86.7569 → wl_surface@31.0 set_buffer_scale(scale=1)
86.7577 → wl_surface@31.0 frame(callback=new wl_callback@51.13799)
86.7587 → wl_surface@31.0 commit()
86.7595 wl_display@1.0 delete_id(id=51) -- destroyed wl_callback@51.13799 after 0.0018s ↲
86.7605 wl_callback@51.13799 done(callback_data=0) ↲
[...]

PR #1974, opened 2021-03-31
Send timestamps with frame done event

wmww:
Noticed this looking into #1967. This doesn't appear to help, but it is more correct then before.


RAOF (approved):
Yup, that's correct. The frame callback is specified to return the “current time” in ms, with an unspecified base, and `chrono::steady_clock` satisfies that.

bors r+

PR #1976, opened 2021-04-01
Do not move on_consumed callback in buffer allocators

wmww:
My understanding of rvalue references is still a little shaky so maybe I'm wrong, but this looks like a bug to me. In each case we're moving an rvalue reference, only to then potentially use it later. I have not observed any symptoms.


RAOF (approved):
You are correct; this should have changed when we added the extra branch.

bors r+

PR #1977, opened 2021-04-01
Wait until surface rendered before sending frame events

wmww:
In the case that there is no buffer, Mir was immediately sending frame events. This was causing excessive CPU usage in Firefox (as noted in #1967). This delays the notifications until the surface is rendered by adding a new callback to the surface. I didn't want to add it to the surface observer because on most surfaces it's not needed most of the time.

### DIFF
--- a/src/server/frontend_wayland/wl_surface.cpp
+++ b/src/server/frontend_wayland/wl_surface.cpp
@@ -332,6 +332,16 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
     // callbacks in wl_surface because if a client commits multiple times before the first buffer is handled, all the
     // callbacks should be sent at once.
     frame_callbacks.insert(end(frame_callbacks), begin(state.frame_callbacks), end(state.frame_callbacks));
+    auto const executor_send_frame_callbacks = [executor = executor, weak_self = mw::make_weak(this)]()
+        {
+            executor->spawn([weak_self]()
+                {
+                    if (weak_self)
+                    {
+                        weak_self.value().send_frame_callbacks();
+                    }
+                });
+        };
 
     if (state.offset)
         offset_ = state.offset.value();
@@ -354,17 +364,6 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
         }
         else
         {
-            auto const executor_send_frame_callbacks = [executor = executor, weak_self = mw::make_weak(this)]()
-                {
-                    executor->spawn([weak_self]()
-                        {
-                            if (weak_self)
-                            {
-                                weak_self.value().send_frame_callbacks();
-                            }
-                        });
-                };
-
             std::shared_ptr<graphics::Buffer> mir_buffer;
 
             if (auto const shm_buffer = wl_shm_buffer_get(buffer))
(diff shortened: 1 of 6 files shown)


wmww:
This now fixes #1967, but not in a production-ready way. If we wanted to to this method but better, I guess we'd pull in the main loop somehow and set up some sort of timer using it (which would be more efficient than spawning a new thread). Or there might be some better way to fix it, but I'm not sure what that would be.

RAOF:
This seems like something that we should be working out what the *actual* behaviour should be is. Requesting a frame callback at the same time as submitting a null buffer - and so *unmapping the surface* - seems like a Firefox bug.

The spec is not particularly prescriptive around exactly when frame callbacks are signalled, but it *does* say that compositors are expected to not fire a frame callback for a non-visible surface; since a surface with a null buffer is non-visible, it seems we'd be perfectly within our rights to immediately discard and *never* fire any frame callbacks requested in that commit.

AlanGriffiths:
The more I read the protocol wording the more I feel that the the requirement is unclear.

The frame notification is a requested _hint_.

AFAICS the only time it is clearly useful to send a frame event is shortly before compositing for an output the window is on. (Unless the compositor knows how long the client is going to need to render it is hard to know when "shortly before" is.)

Ignoring what events we currently signal within Mir, the earliest we know that we are going to composite, is when there's a change in the scene (a buffer sent, a window moved or deleted) for that output. Compositing should be throttled by output refresh rate, so there may be a gap between this notification and reading the buffer for compositing during which the client could render. That would be the logical time to send a frame event. We could approximate that by sending outstanding frame notifications whenever the scene changes.

As @RAOF implies, do we ever need to send frame events? Isn't it just a hint?

wmww:
@RAOF FF is not sending a null buffer, it is sending no buffer-related request at all (meaning the surface continues to be mapped with the old buffer).

@AlanGriffiths as we discussed on the call, sending frames when the scene changes is exactly what a previous version of this PR did, but it doesn't work because FF stalls without a constant stream of frame events.

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -427,7 +429,11 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
     }
     else if (auto const ss = scene_surface())
     {
-        ss.value()->on_next_render(std::move(executor_send_frame_callbacks));
+        std::thread{[callback=std::move(executor_send_frame_callbacks)]()
+            {
+                usleep(10000);
+                callback();
+            }}.detach();
```
AlanGriffiths, line 435: I guess this is the essence of the proposal. My concerns are twofold:

1. We ought to be scheduling callbacks WRT vsync, not when the frame is requested; and,
2. Spinning up a thread every frame just so it can sleep seems an expensive way to do things.

I realise the vsync information isn't available locally, so I might let that go for a "workaround". But we do have a way (`AlarmFactory`) to schedule callbacks at a future time.

Maybe we can have an executor that in the first iteration runs pending work every interval, and later improve that to sync with vsync?

wmww: The new thread approach is just for testing, def not intended to be merged. Good idea to abstract the timing code. I will poke around Mir and maybe also look at what wlroots does to see if we can do the right thing.

wmww:
Superseded by #2006 (which is merged)

Issue #1408, opened 2020-04-07
Popups always appear on output with majority of window


wmww:
As you can see, popups select the output that holds the majority of their parent window, even if they would entirely fit on another output.
![a](https://user-images.githubusercontent.com/9331264/78680025-f6e8f800-78b8-11ea-8751-6a8d9d4c2455.png)

AlanGriffiths:
That's weird. None of the miral popup placement code should be "output aware".

wmww:
Um.. yes it should? That's a major consideration in placing popups.

AlanGriffiths:
Cannot reproduce this year

AlanGriffiths:
> Cannot reproduce this year

Correction, reproduced using egmde-snap/edge as the shell on 18.04.

AlanGriffiths:
Right, this placement logic stems from the Unity8 days when it was enforced that windows could not span outputs.

But that's not what we should have for the "floating" window manager. I suspect we should be taking the display area that contains the anchor rect, not the one that contains the parent.

PR #1982, opened 2021-04-07
Place popups on output their anchor is on

AlanGriffiths:
Only when the parent window spans outputs does this change the behaviour.

In these cases, child window is constrained to the output that contains the anchor rectangle, not the one that contains the parent rectangle. This seems right to me.

Fixes: #1408

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -2053,7 +2053,6 @@ auto miral::BasicWindowManager::place_relative(mir::geometry::Rectangle const& p
 -> mir::optional_value<Rectangle>
 {
     auto const hints = parameters.placement_hints().value();
-    auto const active_output_area = active_output();
     auto const win_gravity = parameters.window_placement_gravity().value();
 
     if (parameters.size().is_set())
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
@wmww (re recent conversation) I suspect this behaviour should be open to customization, but I think this works for now.

src/miral/basic_window_manager.cpp
```diff
@@ -763,23 +773,19 @@ auto miral::BasicWindowManager::display_area_for(WindowInfo const& info) const -
...
                 best_area = area;
                 min_distance = distance;
             }
         }
     }
-
-    if (best_area)
-        return best_area.value();
-    else
-        return std::make_shared<DisplayArea>(window_rect);
```
wmww, line 782: Perhaps `display_area_for(Rectangle)` should take a fallback rectangle argument and create a new `DisplayArea` with it instead of returning an optional. That would slightly simplify the code in both places that call it. Just a thought, feel free to disregard.

wmww (approved):
Just nits, looks sensible

bors delegate+

src/miral/basic_window_manager.cpp
AlanGriffiths: I feel that makes for a weird signature. Maybe:
```c++
    auto area_of_display_area_for(Rectangle const& rect) const -> std::optional<Rectangle>;
...
    auto const placement_bounds = area_of_display_area_for(aux_rect).value_or(parent);
```

AlanGriffiths:
bors r=wmww

PR #1987, opened 2021-04-09
Don't clamp Mir-on-X window size

AlanGriffiths:


wmww (approved):
LGTM, I never liked that "feature"

bors r+

PR #1988, opened 2021-04-09
Fix resize of layer shell surfaces

AlanGriffiths:
Fixes: #1986

PR #1989, opened 2021-04-13
Unbreak wlcs tests

AlanGriffiths:
As described in #1947 some wlcs tests were silently broken by #1988.

This corrects the code to pass those tests. (Which will unblock #1947.)

### DIFF
--- a/src/server/frontend_wayland/layer_shell_v1.cpp
+++ b/src/server/frontend_wayland/layer_shell_v1.cpp
@@ -129,8 +129,8 @@ private:
 
     struct OptionalSize
     {
-        std::experimental::optional<geom::Width> width;
-        std::experimental::optional<geom::Height> height;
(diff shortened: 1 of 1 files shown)


wmww (changes requested):
This isn't quite how I would do it. Most obviously `width_requested` and `height_requested` duplicate the pending size (I think?) which can be retrieved with `opt_size.pending()`. I'm also not convinced `send_configure()` is right. I will make a counter-proposal.

AlanGriffiths:
> Most obviously `width_requested` and `height_requested` duplicate the pending size (I think?) which can be retrieved with `opt_size.pending()`.

They don't. That retrieves the current size regardless of whether it was requested by the client or set by the window manager.

Will look at your alternative.

AlanGriffiths:
Dropped in favour of #1993

PR #1990, opened 2021-04-13
Fix fatal error in wm due to attached window inconsistency

wmww:
In some cases in shutdown (observed writing a WLCS test), an attached window would become restored, which would trigger a fatal error. There is probably an underlying logic bug in the code that arbitrarily makes windows restored, BUT for now I'm just fixing the obvious problem. This calls `update_attached_and_fullscreen_sets()` to remove the inconsistency and changes the fatal error to a non-fatal error, because it's not reporting anything catastrophic. There's no reason the window manger needs to abort.

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -1280,7 +1280,7 @@ void miral::BasicWindowManager::place_attached_to_zone(
     }
 
     default:
-        fatal_error("BasicWindowManager::place_maximized() called for window not in a maximized or attached state");
+        log_error("BasicWindowManager::place_maximized() called for window not in a maximized or attached state");
     }
 
     // TODO: Maybe remove update_window and update only if the rect has changed?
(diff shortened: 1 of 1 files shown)


src/miral/basic_window_manager.cpp
```diff
@@ -1280,7 +1280,7 @@ void miral::BasicWindowManager::place_attached_to_zone(
     }
 
     default:
-        fatal_error("BasicWindowManager::place_maximized() called for window not in a maximized or attached state");
+        log_error("BasicWindowManager::place_maximized() called for window not in a maximized or attached state");
```
AlanGriffiths, line 1283: Preexisting, but the error message is clearly incorrect: the function is `BasicWindowManager::place_attached_to_zone()`

src/miral/basic_window_manager.cpp
```diff
@@ -1592,6 +1592,7 @@ auto miral::BasicWindowManager::select_active_window(Window const& hint) -> mira
     {
         if (!info_for_hint.is_visible())
         {
+            update_attached_and_fullscreen_sets(info_for_hint, mir_window_state_restored);
```
AlanGriffiths, line 1595: This simply looks wrong: we're navigating a tree to find the window to focus on, so why start updating the attachment sets? Based on a faked state??

AlanGriffiths: Oh, looking at the context I see what is happening.

The logic around there is wrong. Just because an invisible window can be activated and isn't _hidden_ doesn't mean it can be restored!

AlanGriffiths:
This is inconsistent with stuff that has landed/is landing. Closing. (The only thing left is the error reporting which can be proposed separately.)

PR #1991, opened 2021-04-13
Make attached surfaces respect exclusive zones

wmww:
Before they were respecting exclusive zones when they were attached to one side, but when they were attached to opposite sides they anchored to the screen edges instead. This makes the behavior consistent and fixes an upcoming WLCS test.


AlanGriffiths:
If you change a test called `stretched_attached_window_ignores_exclusive_zone` to test that exclusive zones are respected, then it is probably good to rename the test.

tests/miral/window_placement_attached.cpp
```diff
@@ -426,7 +426,7 @@ TEST_P(WindowPlacementAttached, window_ignores_exclusive_zone_when_fullscreened)
     EXPECT_THAT(normal.size(), Eq(display_area.size));
 }
 
-TEST_P(WindowPlacementAttached, stretched_attached_window_ignores_exclusive_zone)
+TEST_P(WindowPlacementAttached, stretched_attached_window_respects_exclusive_zone)
```
wmww, line 429: I... did rename the test?

AlanGriffiths (approved):
bors r+

tests/miral/window_placement_attached.cpp
AlanGriffiths: Weird. I see it now.

:shrug:

PR #1993, opened 2021-04-14
Fix layer shell resizing behavior

wmww:
Alternative to #1989. When combined with #1990 and #1991 we pass relevant WLCS tests including those proposed in https://github.com/MirServer/wlcs/pull/194.


AlanGriffiths (approved):
WFM

bors r+

PR #1995, opened 2021-04-14
Only minimized windows should be restored to get focus

AlanGriffiths:
Long ago I wrote 10761d2 which doesn't play nice with the more recent logic for attached windows.

Alternative to #1990


wmww (approved):
LGTM
bors r+

PR #1996, opened 2021-04-14
Tidy BasicWindowManager::update_application_zones_and_attached_windows()

AlanGriffiths:
With clearer logic for placing windows into 1st and 2nd passes some duplicated logic can be removed.

(We could also drop explicit `first_pass` and `second_pass` vectors, but that probably doesn't improve readability.)


wmww (approved):
LGTM
bors r+

PR #1999, opened 2021-04-14
XWayland: drop human-readable window names in debug output

wmww:
This was cute and fairly useful in the early stages of developing XWayland support, but I don't think it's needed any more.


RAOF (approved):
Yeah. Cute, but probably unnecessary at this point.

bors r+

wmww:
If anyone ever needs this for anything else (it's especially useful when you need to print and understand lots of pointers), it lives at https://github.com/wmww/name-of/blob/master/name_of.cpp

PR #2000, opened 2021-04-14
Add error message for window manager inconsistancy

wmww:
This way the `place_attached_to_zone()` can `fatal_error()` if it's preconditions aren't met (meaning there's an obvious logic error), but we get a non-fatal error when the state is wrong (indicating there is an error, but an easier one to slip into the code).


AlanGriffiths (approved):
If we detect an inconsistency in our own logic we should do more to fix it than write a log message. (Ideally make it impossible, otherwise fail noisily.)

But I guess this is arguably an improvement. (But we ought to have something monitoring the log for errors/critical issues. Maybe a "testing logger" for our use is the answer.)

Issue #1889, opened 2021-01-14
[Mir-on-X] Improve titlebar


AlanGriffiths:
The "Mir-on-X" titlebar is not ideal.

1. It should be the name of the current compositor (or configurable by it)
2. When running on a remote X server it should append " (on $(hostname)"

![image](https://user-images.githubusercontent.com/9048879/104575079-af18ad80-564e-11eb-8f89-a658ef8b508d.png)

PR #2005, opened 2021-04-20
Port X11 platform to XCB

### LINKED ISSUE #1471: Keys repeated
After a number of XWayland windows are open, quickly typed keys sometimes come out repeated a bunch of times. It only seems to happen when a bunch of XWayland windows are open, but both XWayland and Wayland windows are effected. I suspect this is due to a combination of causes:
- Either an XWayland thread or the Wayland thread is bogged down (it shouldn't take that long to get events out the door)
- Popper timestamps not being used (I haven't looked into this, but if we're keeping timestamps attached to the events it shouldn't matter if there's a delay going out)
- Short time before keys start repeated (Not the real issue, but our hard-coded value is quite short)

wmww:
XCB is generally cleaner and better than XLib.


AlanGriffiths:
I'm still seeing spurious key repeating (#1471) when using this over ssh -X  forwarding.

AlanGriffiths:
> I'm still seeing spurious key repeating (#1471) when using this over ssh -X forwarding.

Actually, this does make a difference: The spurious repeat stops after a second or so.

AlanGriffiths:
That's more like it:

```
The following tests FAILED:
	155 - mir_unit_tests_x11.X11GraphicsPlatformTest.* (Failed)
	156 - mir_unit_tests_x11.GraphicsPlatform.* (Failed)
	157 - mir_unit_tests_x11.X11DisplayTest.* (Failed)
	158 - mir_unit_tests_x11.DisplayTestGeneric.* (Failed)
	305 - mir_umock_unit_tests.InputPlatformProbe.* (Failed)
	306 - mir_umock_unit_tests.X11PlatformTest.* (Failed)
	307 - mir_umock_unit_tests.X11Platform.* (Failed)
```

AlanGriffiths:
This also segfaulted on me a couple of times:
```
Core was generated by `cmake-build-debug/bin/miral-shell.bin'.
Program terminated with signal SIGSEGV, Segmentation fault.
[Current thread is 1 (Thread 0x7f8115ffb700 (LWP 3028762))]
(gdb) bt
#0  0x00007f81946a5743 in ?? () from /lib/x86_64-linux-gnu/libEGL_mesa.so.0
#1  0x00007f81946a5a75 in ?? () from /lib/x86_64-linux-gnu/libEGL_mesa.so.0
#2  0x00007f8192de5c9a in ?? () from /usr/lib/x86_64-linux-gnu/dri/swrast_dri.so
#3  0x00007f8192de5e27 in ?? () from /usr/lib/x86_64-linux-gnu/dri/swrast_dri.so
#4  0x00007f81946a31ea in ?? () from /lib/x86_64-linux-gnu/libEGL_mesa.so.0
#5  0x00007f819469a72a in ?? () from /lib/x86_64-linux-gnu/libEGL_mesa.so.0
#6  0x00007f8194692f03 in ?? () from /lib/x86_64-linux-gnu/libEGL_mesa.so.0
#7  0x000055823f968a68 in MirEglApp::swap_buffers (this=0x7f80e8004d10, eglsurface=0x7f80e8253830) at /home/alan/CLionProjects/mir/examples/miral-shell/spinner/miregl.cpp:201
```

wmww:
Did it segfault in the tests or in normal use? Either way it's strange. How is the platform causing an example client to crash?

AlanGriffiths:
Sorry, it segfaulted when I ran miral-app over `ssh -XC` approx 3 times out of 6.

AlanGriffiths:
> Sorry, it segfaulted when I ran miral-app over `ssh -XC` approx 3 times out of 6. It is the spinner that crashes.

I've now seen this segfault with the xlib based platform. It seems that the spinner doesn't "play nice" with either Mir-on-X version. It definitely happens more now, and doesn't happen on other platforms.

AlanGriffiths:
@wmww I have tracked down two problems:

1. with ssh -XC over the LAN there's a timing issue. (This doesn't happen on the same machine, even with `ssh -X localhost`.)

2. Separately, if the Mir-on-X window loses focus when a key is pressed (press a key and click on another window) autorepeat happens. And continues after focus is regained. We should probably react to loss of focus by clearing the keyboard state.

Problem 1 goes away with any "printf debugging" and also by adding a short sleep:

```c++
            if (!consumed)
            {
                process_input_event(event.get());
            }
            using namespace std::chrono_literals;
            std::this_thread::sleep_for(1ns);
        }
        else
        {
            mir::log_error("input event received with no sink to handle it");
```
That makes me think we occasionally don't see both events of a key repeat because of network lag. (If we need to yield in production code, there are probably better places to do it.)

wmww:
One thing to note is that if a key up event is delayed on the wire, there's not much we can do. By the time we get it a client will have already gotten an extended press and have interpreted that as a key repeat. We can use event timestamps to detect what happened, but short of synthesizing a backspace and sending it to the client there's no untyping the key repeats.

If this was a serious problem we could maybe always immediately send key-ups and then relay key repeats from the X11 server as we receive them. This would work, at least when the user is typing text normally. It would completely break games or anything that actually cares if the keys are up or down though.

I was hoping this wouldn't be the real problem though, and still maybe something else is going on which we can fix.

AlanGriffiths:
> One thing to note is that if a key up event is delayed on the wire, there's not much we can do.

I suspect what is happening is that autorepeat event pair is being split: the key down is arriving after the key up is processed.

Which leads to a question why are we ignoring key repeats and then synthesizing then for ourselves? (My hack effectively passed all events straight through.)

wmww:
In Wayland *clients* generate key repeats (compositors just give them the timing parameters). We may also be generating and subsequently dropping key repeats internally in Mir, because mir protocol used to care about them.

AlanGriffiths:
That makes sense. I can try to prove my theory tomorrow now I have something concrete to test. Then there's the more interesting question of the right thing to do about it.

Let me know if you make progress by your EOD.

wmww:
If repeat pair was being split, the current XCB platform should
1. See the key up
2. Peek at the next pending event
3. Next pending event is null, because we haven't gotten the key down yet
4. Assume it's not a key repeat
5. Send the key up on to the client
6. When the key down comes in, send it to the client as well

It would technically be incorrect because the client would receive an up/down pair when there was none really. However the worst that could happen is a key repeat with the X11 timing instead of the Wayland timing (if they're different), or odd behavior in a game/other non-typing situation. Key repeat pairs being split should *not* result in key repeats for normally typed letters (there aren't even any key repeats sent by the X server in the normal situation).

AlanGriffiths:
Yeah, I know it can't be the whole problem. But the sleep "fixing" it is very suggestive of something racy going on. And the need for a slow connection to provoke it in the first place...

AlanGriffiths:
I've managed to prove this is a false trail. Running Mir on a slower machine I've proved that:

1. the sleep doesn't offer a way to fix the problem (even if extended to 1ms)
2. I can add some printf debugging without hiding the issue

With the debugging information I can prove this is unrelated to the "drop key repeat" logic which does seem to work. Although it could be a lot simpler: https://github.com/MirServer/mir/commit/dbe96284461b84391a1397199126bbc6d61a115c

There is clearly still something about the X11 platforms that is messing up keyboard input, but it isn't a blocker for this PR.

AlanGriffiths:
> One thing to note is that if a key up event is delayed on the wire, there's not much we can do. By the time we get it a client will have already gotten an extended press and have interpreted that as a key repeat. We can use event timestamps to detect what happened, but short of synthesizing a backspace and sending it to the client there's no untyping the key repeats.

Yeah, I haven't proved this is actually "on the wire" but this is, in effect, what's happening. With the right debugging:
```plain
[2021-04-23 15:58:43.483781] < - debug - > x11-input: ====> XInputDevice::key_press(71003340000000, o, ...)
[2021-04-23 15:58:48.662458] < - debug - > x11-input: ==> XInputDevice::key_release(71003444000000, o, ...) // 104ms since press
```
The event timestamps are 104ms apart, but the times the events are processed are 5s apart.

wmww:
Getting distracted by more important things and lots of confusion around the best way to make the tests run has delayed this PR, but it's still making progress. I will interoperate your simplification of the `with_next_pending_event()` (or something like it) as I agree that's over-engineered currently.

wmww:
Looks like CI failures are either triggered by the added dependencies, or not this PRs fault at all

AlanGriffiths:
The failure in `spread/build/ubuntu:rpi` is indeed a dependency issue. (It isn't immediately obvious to me what it is complaining about though.)

AlanGriffiths:
I just noticed you don't set the window title. Maybe you can fix #1889 at the same time?

wmww:
Do we have a way of knowing the name of the current shell? Or would that require adding an API? If the latter, that would seem out of scope here.

AlanGriffiths:
The current Xlib platform sets the title to "Mir on X". This PR doesn't set it at all.

Setting the actual shell name would be nicer, but I don't think we have that anywhere. (We do extract the name of the executable in a few places, which could be an option.)

wmww:
Just setting to `Mir on X` for now because it's easy. Future PR can do something fancier.

wmww:
Two problems I've noticed:
1. This seems to introduce a new problem with regards to initial window size
2. This seems to break the `--x11-output` option??

wmww:
Initial window size bug was fixed by `XSetEventQueueOwner()` (thanks RAOF for finding that)

AlanGriffiths (approved):
I don't think this fixes anything, but it is easier to follow.

bors r+

PR #2006, opened 2021-04-21
Send bufferless frame callbacks after timeout

### LINKED ISSUE #1967: Firefox does not take input in Mir

wmww:
Fixes #1967, replaces #1977. Adds a frame callback executor that sends frame callbacks on a delay. Approach is similar to #1977, but uses the main loop instead of spinning up new threads (more efficient).

In the future we might want to consider synchronizing with V-sync, although in practice I'm not sure that's important. This is only used when clients are intentionally skipping frames, so I don't think the timing matters much in this case.

Another potential improvement would be to stop sending callbacks when surfaces are not being drawn by Mir. This would reduce chatter on the wayland socket, and might have additional performance benefits in clients. I think there might be a way to do this. I will look into it, but it doesn't need to block this PR.

### DIFF
--- a/src/server/frontend_wayland/frame_callback_executor.h
+++ b/src/server/frontend_wayland/frame_callback_executor.h
@@ -1,0 +1,64 @@
+/*
+ * Copyright © 2021 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: William Wold <william.wold@canonical.com>
+ */
+
+#ifndef MIR_FRONTEND_FRAME_CALLBACK_EXECUTOR_H
+#define MIR_FRONTEND_FRAME_CALLBACK_EXECUTOR_H
+
+#include <mir/executor.h>
+
+#include <vector>
+#include <mutex>
+#include <memory>
+
+namespace mir
+{
+namespace time
+{
+class Alarm;
+class AlarmFactory;
+}
+
+namespace frontend
+{
+
+/// Runs frame callbacks that do not have a buffer to be attached to. Is threadsafe. Given callbacks are run on the
+/// main loop thread, the wayland executor is NOT automatically used.
+class FrameCallbackExecutor : public Executor
+{
+public:
+    FrameCallbackExecutor(time::AlarmFactory& alarm_factory);
+
+    void spawn(std::function<void()>&& work) override;
+
+private:
+    struct Callbacks
+    {
+        std::mutex mutex;
+        std::vector<std::function<void()>> queued;
+    };
+
+    std::shared_ptr<Callbacks> const callbacks; // shared_ptr so it can potentially outlive this object
+    std::unique_ptr<time::Alarm> const alarm;
+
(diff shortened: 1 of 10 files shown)


src/server/frontend_wayland/frame_callback_executor.h
```diff
@@ -0,0 +1,64 @@
...
+class Alarm;
+class AlarmFactory;
+}
+
+namespace frontend
+{
+
+/// Runs frame callbacks that do not have a buffer to be attached to. Is threadsafe. Given callbacks are run on the
+/// main loop thread, the wayland executor is NOT automatically used.
+class FrameCallbackExecutor : public Executor
```
AlanGriffiths, line 41: I think something is wrong here, probably the name.

If we really wanted to constrain the type of work to frame callbacks, then we should not derive from `Executor` (which takes any type of work) but have a `spawn` function that only accepts `WlSurface` objects on which it calls `send_frame_callbacks()`.

Instead, it should be named by the service it provides, not the expected usage. `VsyncExecutor` is clearly a lie, but how about `CompositionSyncExecutor`, or if "frame" is interpreted more broadly `FrameExecutor`?

AlanGriffiths (changes requested):
Mostly nits.

I do wonder about a scenario where a client submits a buffer while there's a frame callback outstanding. As I understand it, this could then send both frame callbacks when the alarm fires before composition consumes the buffer. Could that lead to "bad stuff"?

wmww:
> I do wonder about a scenario where a client submits a buffer while there's a frame callback outstanding. As I understand it, this could then send both frame callbacks when the alarm fires before composition consumes the buffer. Could that lead to "bad stuff"?

I don't think so? There's no strict requirement that the frame callback can't be sent too soon. We've observed FF causing trouble, but that's only if a `roundtrip` after a frame *consistently* results in the frame event being sent (it gets stuck in an endless loop). I wouldn't worry about this case unless we have more reason to believe it's a problem.

src/server/frontend_wayland/frame_callback_executor.h
```diff
@@ -0,0 +1,64 @@
...
+{
+class Alarm;
+class AlarmFactory;
+}
+
+namespace frontend
+{
+
+/// Runs frame callbacks that do not have a buffer to be attached to. Is threadsafe. Given callbacks are run on the
+/// main loop thread, the wayland executor is NOT automatically used.
```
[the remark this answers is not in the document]
wmww: I was previously considering that, but RAOF wants to keep callbacks connected to buffers. Solving the problem well for multiple outputs seems like a hard problem. We can change the docs here when we do that. For now I want to make it clear haw this is intended to be used, so people who find this class aren't confused when it doesn't process all frame callbacks.

src/server/frontend_wayland/frame_callback_executor.h
wmww: I did struggle a bit with the name. I'm going with `FrameExecutor`, since I think that is currently the most accurate.

src/server/frontend_wayland/frame_executor.h
```diff
@@ -0,0 +1,59 @@
...
+namespace time
+{
+class Alarm;
+class AlarmFactory;
+}
+
+namespace frontend
+{
+
+/// Runs frame callbacks that do not have a buffer to be attached to.
```
AlanGriffiths, line 37: I still don't like this comment. Documentation (and naming) should be about the feature an abstraction provides, not the initial usecase. In this case, the feature is delayed and throttled execution.

AlanGriffiths (approved):
Good enough.

bors r+

PR #2008, opened 2021-04-22
[Mir-on-X] Fix key repeat

AlanGriffiths:
Don't drop KeyRelease when the host X server has key repeat events queued. (Fixes: #1471)

### DIFF
--- a/src/platforms/x11/input/input_platform.cpp
+++ b/src/platforms/x11/input/input_platform.cpp
@@ -193,28 +193,29 @@ void mix::XInputPlatform::process_input_event()
                 }
 
             case KeyPress:
+                // Ignore key repeats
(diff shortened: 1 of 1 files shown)


wmww:
I'm confused. How did this ever work? Did it get merged without ever being tested to work? If it really does fix #1471, why doesn't #2005 also fully work since the key repeat behavior there *is* tested to work.

AlanGriffiths:
> I'm confused. How did this ever work? Did it get merged without ever being tested to work? If it really does fix #1471, why doesn't #2005 also fully work since the key repeat behavior there _is_ tested to work.

I've confused myself. This _does_ seem to stop the symptom. But it fails tests and the logic looks wrong. (The test looks wrong too.)

AlanGriffiths:
Further investigation proved this wrong

PR #2009, opened 2021-04-22
WLCS WaylandExecutor: don't do work under lock

wmww:
Using a recursive mutex and doing `work()` under lock was causing the deadlock observed in #1881. I'm unsure of if that exposed a bug that was already there, or if changes there created a new bug, but either way it's dependent on the wayland executor jobs being run under lock.

I don't think there's any reason to do that or to use a recursive mutex. The "real" Wayland executor doesn't execute arbitrary work under lock (as far as I can tell), so the WLCS one shouldn't need to either.

### DIFF
--- a/tests/acceptance-tests/wayland/test_wlcs_display_server.cpp
+++ b/tests/acceptance-tests/wayland/test_wlcs_display_server.cpp
@@ -199,7 +199,7 @@ public:
     void spawn (std::function<void()>&& work) override
     {
         {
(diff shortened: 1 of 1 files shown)


AlanGriffiths (approved):
LGTM

It does look as though there's nothing preventing `on_display_destruction()` deleting `shim` while work is executing, but that is pre-existing, probably doesn't happen in practice and doesn't guarantee a problem.

bors r+

PR #2015, opened 2021-04-30
Keep track of past surface states

### LINKED ISSUE #1979: Unfullscreening app also unmaximizes it
To reproduce:
1. maximize app
2. fullscreen it
3. unfullscreen it

Expected: app is maximized

Actual: app is restored

wmww:
Fixes #1979 and similar bugs that haven't been reported. With this PR, the surface stores info about past states. When the frontend or other part of Mir wants to "remove" a state (for example unfullscreen), it can then query that info to get the new state it should set.


src/server/symbols.map
```diff
@@ -169,6 +169,7 @@ MIR_SERVER_1.6.0 {
     mir::scene::SurfaceObserver::operator*;
     mir::scene::SurfaceObserver::?SurfaceObserver*;
     mir::scene::SurfaceObserver::SurfaceObserver*;
+    mir::scene::SurfaceStateStack::*;
```
AlanGriffiths, line 172: We've going to be releasing this with 2.4, so 1.6 is the wrong stanza for this to be added to.

src/include/server/mir/scene/surface_state_stack.h
```diff
@@ -0,0 +1,81 @@
...
+{
+namespace scene
+{
+/// Represents a surface's current state, as well as the states it will fall back to as states are removed. For example
+/// A minimized surface may also be fullscreen (which will become relevant when it's unminimized) but not maximized
+/// (which will become relevant when it's unfullscreened). This class keeps track of all the active and inactive states.
+/// States have this order of precedence (earlier ones will be the active state if present):
+/// - hidden, minimized, fullscreen, attached, maximized/vertmaximized/horizmaximized, restored
+/// This struct can NOT be safely accessed by multiple threads at once.
+struct SurfaceStateStack
```
AlanGriffiths, line 34: I think "stack" sets algorithmic expectations that are not what this does. In particular, there are no "push/pop" operations.

How does `SurfaceStateTracker` sound?

src/include/server/mir/scene/surface_state_stack.h
```diff
@@ -0,0 +1,81 @@
...
+namespace mir
+{
+namespace scene
+{
+/// Represents a surface's current state, as well as the states it will fall back to as states are removed. For example
+/// A minimized surface may also be fullscreen (which will become relevant when it's unminimized) but not maximized
+/// (which will become relevant when it's unfullscreened). This class keeps track of all the active and inactive states.
+/// States have this order of precedence (earlier ones will be the active state if present):
+/// - hidden, minimized, fullscreen, attached, maximized/vertmaximized/horizmaximized, restored
+/// This struct can NOT be safely accessed by multiple threads at once.
```
AlanGriffiths, line 33: Nitpicking: this class _can_ be accessed across multiple threads. What isn't safe is mutating when other threads may be accessing or mutating.

It is tempting to make the class immutable. Vis:
1. `set_active_state()`, `with()` and `without()` return new instances; and,
2. assignment is deleted.

That would require a mutable "holder" that can be assigned or reseated with any necessary locking. (`unique_ptr` would do, but seems a bit heavy.) Maybe make `operator=()` protected and have a mutable derived class that makes assignment public?

src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -573,22 +573,24 @@ void mf::XdgToplevelStable::send_toplevel_configure()
...
-    case mir_window_state_maximized:
-    case mir_window_state_horizmaximized:
-    case mir_window_state_vertmaximized:
-        if (uint32_t *state = static_cast<decltype(state)>(wl_array_add(&states, sizeof *state)))
-            *state = State::maximized;
-        break;
+        auto const state = surface.value()->state_stack();
+        if (state.has(mir_window_state_maximized) ||
+            state.has(mir_window_state_horizmaximized) ||
+            state.has(mir_window_state_vertmaximized))
```
AlanGriffiths, line 581: Seeing this usage, maybe `has()` should take an `initializer_list<MirWindowState>`?

AlanGriffiths (changes requested):
I like it. Apart from the name.

src/include/server/mir/scene/surface_state_stack.h
wmww: Understandable, as I worked on this I was noticing the API drifted further and further from the original vision of a stack. `SurfaceStateTracker` is as good as anything I can think of.

src/include/server/mir/scene/surface_state_stack.h
wmww: I agree with making the class immutable _except_ for assignment. This would give it an interface similar to a primitive data type, which I think is reasonable.

src/server/symbols.map
wmww: Done (I think?)

src/server/frontend_wayland/xdg_shell_stable.cpp
wmww: Done

AlanGriffiths (approved):
LGTM

bors r+

PR #2024, opened 2021-05-04
Add mechanism to skip "broken" wlcs tests

AlanGriffiths:
We already dropped a test upstream: https://github.com/MirServer/wlcs/pull/196

But until wlcs is released we need a way to skip it


wmww (approved):
LGTM
bors r+

PR #2031, opened 2021-05-08
X11 platform: handle key state changes when window is not focused

### LINKED ISSUE #2029: Mir-on-X modifier state not cleared
I don't *think* this is the same as #2025, as that would seem to be in the frontend logic while this is in specifically the Mir-on-X platform. To reproduce
1. Open Mir-on-X
2. Switch to a different window with the modifier down
3. Switch back with the modifier up
4. Modifier is seen as down in Mir until you press it again

wmww:
Fixes #2029. There are a couple pieces to this.
1. Use the XKB extension to track the modifier state. This gives us more correct symcodes (which we send on to Mir but don't seem to care much about) as well as the ability to know the state of the modifiers when the window is focused. 
2. Cache pressed keys. For a number of reasons key ups and key downs may not be perfectly paired up (XKB events duplicating normal events in some cases, keys that are pressed when the window is unfocused and released when the window is focused, etc). The `pressed_keys` set allows us to make sure the events we send on to Mir are more coherent.
3. Synthesize key up events for all non-modifiers when the window looses focus. This keeps us from getting endlessly repeating keys. Modifiers are any key that an XKB event has related to in the past (because these are the keys for which we *will* see changes made when the window isn't focused).

This ended up being probably more code than was worth it, but it's done now.

### DIFF
--- a/src/platforms/x11/input/input_platform.cpp
+++ b/src/platforms/x11/input/input_platform.cpp
@@ -71,7 +71,7 @@ auto const scroll_factor = 10;
 
 auto init_xkb_extension(mir::X::X11Resources* x11_resources) -> xcb_query_extension_reply_t const*
 {
-    auto const xkb_extension = xcb_get_extension_data(x11_resources->conn->connection(), &xcb_xkb_id);
+    auto const xkb_extension = x11_resources->conn->get_extension_data(&xcb_xkb_id);
     if (!xkb_extension || !xkb_extension->present)
     {
         mir::log_warning("XKB X11 extension not available, keyboard will not work");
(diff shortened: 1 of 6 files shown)


src/platforms/x11/input/input_platform.cpp
```diff
@@ -39,6 +39,18 @@
...
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wkeyword-macro"
+#endif
+#define explicit explicit_
+#include <xcb/xkb.h>
+#undef explicit
+#ifdef __clang__
+#pragma clang diagnostic pop
+#endif
```
AlanGriffiths, line 52: I'm a little concerned that this incantation is repeated. But I'll let it go until there's a third occurance.

AlanGriffiths (approved):
WFM

bors r+

PR #2033, opened 2021-05-11
platform/graphics: Renderables don't have a swap_interval.

RAOF:
It doesn't make sense to ask what the swap interval of a Renderable is;
a Renderable is something that you draw *from*, not draw to.

And indeed, the fact that the method is unused suggests its lack of utility.


wmww (approved):
LGTM

bors r+

PR #2035, opened 2021-05-11
X11 platform: fix cursor position on scaled outputs

### LINKED ISSUE #2034: X11 platform: cursor movement incorrect on scaled output
When output is scaled, cursor moves twice as fast as it should, resulting in the actual cursor becoming misaligned with the Mir cursor

wmww:
Fixes #2034. Also introduces floating point geometry into the X11 platform, and removes the usage of `as_int()` from generic geometry calculations.


AlanGriffiths (approved):
Yes

bors merge

PR #2038, opened 2021-05-13
XWayland: improve popup window types and positioning

### LINKED ISSUE #2037: XWayland popups sometimes get input focus
Depending on what window types X apps set, XWayland popups can get a Mir window type that gets input focus. This seems to be incorrect, and causes problems in KDevelop for example.

wmww:
Fixes #2037. I hope this refactor moderately improves some of the convoluted logic. Now window parent and type are set in one place, so we always have the information to set them correctly. Draft PR for now because I've seen occasional segfaults, so I think this introduces a memory issue somewhere.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -1316,14 +1316,6 @@ void mf::XWaylandSurface::apply_cached_transient_for_and_type(std::lock_guard<st
             }
         }
     }
-    else
-    {
-        // Type should not have parent, if it does we should use a different type
-        if (parent)
-        {
-            type = mir_window_type_gloss;
-        }
-    }
 
     effective_parent = parent;
 
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -230,11 +230,11 @@ auto property_handler(
 }
 
 template<typename T>
-auto optional_from(bool return_value, T&& value) -> std::optional<T>
+auto make_optional_if(bool condition, T&& value) -> std::optional<T>
 {
-    if (return_value)
+    if (condition)
     {
-        return std::optional<T>(value);
+        return std::optional<T>(std::move(value));
     }
     else
     {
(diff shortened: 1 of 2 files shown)


wmww:
This now fixes most of the problems I've observed thus far with Qt popups on X11

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -183,6 +228,19 @@ auto property_handler(
 {
     return property_handler<T>(connection, window, property, mf::XCBConnection::Handler<T>{std::move(handler)});
 }
+
+template<typename T>
+auto optional_from(bool return_value, T&& value) -> std::optional<T>
```
AlanGriffiths, line 233: I see the utility, but I'm not entirely convinced by this naming. The "optional" part of the name could be referring to the internal logic or what is built, the "from" part doesn't really suggest constructing a new object.

What I'd like is `make_optional(value, if(condition))`. But making that compile would be abusing the language.

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -183,6 +228,19 @@ auto property_handler(
...
+{
+    if (return_value)
+    {
+        return std::optional<T>(value);
```
AlanGriffiths, line 237: Is there any point in taking an rvalue reference if we don't `std::move()` from it?

AlanGriffiths (approved):
I'll approve, but suggest renaming `optional_from` to "something better" before landing.

bors delegate+

src/server/frontend_xwayland/xwayland_surface.cpp
AlanGriffiths: How about:
```c++
auto make_optional_when(bool return_value, T&& value) -> std::optional<T>
...
    make_optional_when(event->value_mask & XCB_CONFIG_WINDOW_X, geom::X{event->x}),
    make_optional_when(event->value_mask & XCB_CONFIG_WINDOW_Y, geom::Y{event->y}),
    make_optional_when(event->value_mask & XCB_CONFIG_WINDOW_WIDTH, geom::Width{event->width}),
    make_optional_when(event->value_mask & XCB_CONFIG_WINDOW_HEIGHT, geom::Height{event->height}));
```

src/server/frontend_xwayland/xwayland_surface.cpp
wmww: Fixed

AlanGriffiths (approved):
We should get some tests around this code. But "pre-existing".

bors r+

PR #2041, opened 2021-05-18
renderer/gl: Drop no-longer-used mir::gl::Texture infrastructure.

RAOF:
This is replaced by mir::graphics::Texture from the platform interface


wmww (approved):
Needs work on the tests, but looks good to me
bors delegate+

RAOF:
Well, that sure was A Journey.

bors r=wmww

PR #2042, opened 2021-05-20
scene: Drop surface snapshot support and associated infrastructure

RAOF:
This is currently broken - with any graphics platform attempting to
snapshot a surface will currently throw an unhandled `std::logic_error`.

It was also not very useful even *before* platform changes broke it -
it only captures the single “default” BufferStream of a surface
and we were in the process of *deprecating* the default BufferStream.

A more useful surface snapshot infrastructure might be interesting in
the future, but it'll need to be based on rendering a subgraph of
the scenegraph, and so this code won't be relevant even then.


RAOF:
I've also reached out to UBPorts; as far as they can tell they don't use this interface.

AlanGriffiths (approved):
Yes, put it out if it's misery

bors merge

PR #2049, opened 2021-05-25
Improve Wayland object destruction path

### LINKED ISSUE #1447: WlSurface::frame_callbacks are awkwardly managed shared pointers

wmww:
This changes the generated code to automatically destroy Wayland resources. This cleans up a ton of user functions, almost all of which were just calling `destroy_wayland_object()`. For the few that had additional logic, it's been moved to the destructor.

`destroy_wayland_object()` is now completely gone, and replaced with `destroy_and_delete()` for objects that don't have a destructor.

Fixes #1447

### DIFF
--- a/src/platform/graphics/linux_dmabuf.cpp
+++ b/src/platform/graphics/linux_dmabuf.cpp
@@ -514,11 +514,6 @@ public:
         return planes_;
     }
 private:
-    void destroy() override
-    {
-        destroy_wayland_object();
-    }
-
     EGLDisplay const dpy;
     std::shared_ptr<mg::EGLExtensions> const egl_extensions;
     BufferGLDescription const& desc;
@@ -595,11 +590,6 @@ private:
     EGLDisplay dpy;
     std::shared_ptr<mg::EGLExtensions> egl_extensions;
     std::shared_ptr<mg::DmaBufFormatDescriptors const> const formats;
-
-    void destroy() override
-    {
-        destroy_wayland_object();
-    }
 
     void add(
         mir::Fd fd,
@@ -1065,11 +1055,6 @@ public:
         }
     }
 private:
-    void destroy() override
-    {
-        destroy_wayland_object();
-    }
-
     void create_params(struct wl_resource* params_id) override
     {
         new LinuxDmaBufParams{params_id, dpy, egl_extensions, formats};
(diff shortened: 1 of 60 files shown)


wmww:
This also makes the destruction order for role objects more strictly complaint with the spec. This results in some tests failing because we aren't correctly destroying subsurfaces in WLCS. Fixed by https://github.com/MirServer/wlcs/pull/199.

AlanGriffiths:
> This also makes the destruction order for role objects more strictly complaint with the spec. This results in some tests failing because we aren't correctly destroying subsurfaces in WLCS. Fixed by [MirServer/wlcs#199](https://github.com/MirServer/wlcs/pull/199).

Well, we need to something about failing tests before landing. Like marking them "broken"? Or first releasing an updated wlcs?

wmww:
Indeed, I wasn't quite sure. I guess I'll figure out exactly which tests are failing and mark those as bad, with a note to remove that designation when we release a new WLCS.

RAOF:
We don't have to *release* a new WLCS to use it in CI. We only need merge any fixes, and then copy the package from the wlcs-dev PPA

AlanGriffiths:
> We don't have to _release_ a new WLCS to use it in CI. We only need merge any fixes, and then copy the package from the wlcs-dev PPA.

Technically true, but the more we integrate management of the two projects the less point there is to having them in separate repositories.

wmww:
Whatever we're going to do, can we hurry it along? This PR does touch quire a lot of things so getting it merged quickly would be preferable.

AlanGriffiths (approved):
Looks like a sensible cleanup to me.

bors r+

PR #2054, opened 2021-06-02
Release 2.4.0

wmww:


debian/changelog
```diff
@@ -1,3 +1,39 @@
...
+      . mircommon ABI unchanged at 7
+      . mirplatform ABI bumped to 22
+      . mirprotobuf ABI unchanged at 3
+      . mirplatformgraphics ABI unchanged at 18
+      . mirinputplatform ABI unchanged at 8
+      . mircore ABI unchanged at 1
+      . mircookie ABI unchanged at 2
+      . mirwayland ABI unchanged at 2
+    - Enhancements:
+      . [gbm-kms] Also respect --driver-quirks in the RenderingPlatform
```
RAOF, line 19: This is a fix/follow-up for the initial `--driver-quirks` work, which we haven't mentioned.

This should probably be something like:
[gbm-kms] Add --driver-quirks option to avoid probing devices where probing crashes

RAOF (changes requested):

AlanGriffiths:
# rc4

## Debs

Test plan: https://wiki.ubuntu.com/Process/Merges/TestPlans/Mir

### Installed from from ppa:mir-team/rc

Platform|21.10|21.04|20.04LTS
--|--|--|--
gbm-kms|||
gbm-x11|||@AlanGriffiths
eglstream-kms|||
dispmanx|||

Console|21.10|21.04|20.04LTS
--|--|--|--
minimal|||@AlanGriffiths
linux-VT|||
logind|||

WM|21.10|21.04|20.04LTS
--|--|--|--
floating|||@AlanGriffiths
tiling|||
kiosk|||

AlanGriffiths:
Also quickly tried the egmde & ubuntu-core-kiosk snaps from `--beta`, these seem to work as expected. (Except for some weird stuff on my hybrid laptop I need to investigate further - probably pre-existing.)

AlanGriffiths:
> (Except for some weird stuff on my hybrid laptop I need to investigate further - probably pre-existing.)

The problem I see is that with a client that submits a lot of buffers, such as mir-kiosk-kodi, it works for a short while and then the system locks up.

This isn't pre-existing and appears to relate to hybrid graphics. I have a workaround: Adding `driver-quirks=skip:driver:nouveau` to the config.

It looks as though Mir is trying to run in a hybrid mode and failing, I've attached a log:

[ubuntu-core-kiosk.log](https://github.com/MirServer/mir/files/6586000/ubuntu-core-kiosk.log)

@RAOF is this a blocker? Or something to fix in the next release?

RAOF:
> The problem I see is that with a client that submits a lot of buffers, such as mir-kiosk-kodi, it works for a short while and then the system locks up.
> 
> This isn't pre-existing and appears to relate to hybrid graphics. I have a workaround: Adding driver-quirks=skip:driver:nouveau to the config.

What? That's *not* pre-existing? I didn't think any of the changes we've made would have had any functional change yet?

Is this maybe a change in Mesa? Is there a new mesa in the new snap? Is it easy to test a rebuild of the 2.3.2 snap with any new mesa?

AlanGriffiths:
> What? That's _not_ pre-existing? I didn't think any of the changes we've made would have had any functional change yet?
> 
> Is this maybe a change in Mesa? Is there a new mesa in the new snap? Is it easy to test a rebuild of the 2.3.2 snap with any new mesa?

Your suspicion is confirmed. Rebuilding with Mir 2.3.3 (from ppa:mir-team/release) I see the same problem. (But I have no workaround.)

**So this is not a blocker.**

As to what (other than Mir) has changed, there's a lot:

```plain
diff /snap/egmde/7346/snap/manifest.yaml /snap/egmde/7813/snap/manifest.yaml | grep -v -e libmir -e mir-platform
```

[egmde-manifest.txt](https://github.com/MirServer/mir/files/6590268/egmde-manifest.txt)

RAOF (approved):

AlanGriffiths:
@RAOF in our sync I mentioned problems with displaylink. It amounts to:
```plain
MESA-LOADER: failed to open evdi ...
```
This is, apparently, "known" and fixed in the latest versions of Mesa and evdi. But on 20.04 (where I was testing) these are not in sync. As I've a usable workaround this is also not a blocker.

wmww:
We may want to block the release on #2055. If there are clients out there destroying things in the wrong order that we don't know about, that could get very annoying very fast.

AlanGriffiths:
@RAOF was the eglstream-kms testing really on rc4 or (as I suspect) rc5?

AlanGriffiths:
# rc5

Both egmde and confined-desktop snaps work when build with rc5

RAOF:
> @RAOF was the eglstream-kms testing really on rc4 or (as I suspect) rc5?

You are correct; rc5.

RAOF:
OK. I've done some more testing of NVIDIA now that [#egl-wayland/33](https://github.com/NVIDIA/egl-wayland/issues/33) is fixed, and, while native Wayland clients work fine, XWayland clients appear to quickly violate our EGL expectations, resulting in XWayland being killed with a protocol violation at best or crashing Mir at worst.

I don't think this is a regression, though; I just haven't extensively tested XWayland before.

AlanGriffiths:
bors r=@RAOF

PR #2056, opened 2021-06-03
Do not throw exceptions in surface_destroyed()

### LINKED ISSUE #2055: Mir crashes when surface and role destroyed in wrong order
#2049 made Mir less tolerant of clients that destroy their surface's before destroying the associated role objects. It throws an exception in this case which is supposed to be caught and processed as a protocol error that disconnects the client. Unfortunately the exception is being thrown inside a destructor, which causes Mir to segfault

So far I've only observed this with Squeekboard on my branch with protocol stubs added, but the error is obvious from the code.

wmww:
Fixes #2055. Adds a new function to mirwayland (not an ABI break), so check to make sure I did the symbols right.

### DIFF
--- a/include/wayland/mir/wayland/wayland_base.h
+++ b/include/wayland/mir/wayland/wayland_base.h
@@ -215,6 +215,10 @@ public:
     wl_global* const global;
 };
 
+/// Calls wl_client_post_implementation_error() if available
+[[gnu::format(printf, 2, 3)]]
+void post_implementation_error(wl_client* client, char const* fmt, ...);
+
 void internal_error_processing_request(wl_client* client, char const* method_name);
 
 }
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -447,9 +447,14 @@ void mf::WindowWlSurfaceRole::surface_destroyed()
     {
         // "When a client wants to destroy a wl_surface, they must destroy this 'role object' wl_surface"
         // NOTE: the wl_shell_surface specification seems contradictory, so this method is overridden in it's implementation
-        BOOST_THROW_EXCEPTION(std::runtime_error{
-            "wl_surface@" + std::to_string(wl_resource_get_id(surface->resource)) +
-            " destroyed before associated role"});
+        log_warning(
+            "wl_surface@%d destroyed before associated role",
+            wl_resource_get_id(surface->resource));
+        // Post error manually instead of throwing exception because this method is called by the surface destructor
+        wayland::post_implementation_error(
+            weak_client.value().raw_client(),
+            "wl_surface@%d destroyed before associated role",
+            wl_resource_get_id(surface->resource));
     }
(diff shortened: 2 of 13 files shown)


src/wayland/wayland_base.cpp
```diff
@@ -143,13 +143,23 @@ mw::Global::~Global()
     wl_global_destroy(global);
 }
 
-void mw::internal_error_processing_request(wl_client* client, char const* method_name)
+void mw::post_implementation_error(wl_client* client, char const* fmt, ...)
 {
+    va_list va;
+    va_start(va, fmt);
 #if (WAYLAND_VERSION_MAJOR > 1 || (WAYLAND_VERSION_MAJOR == 1 && WAYLAND_VERSION_MINOR > 16))
```
AlanGriffiths, line 150: Maybe not in this PR, but I think we can drop this backwards-compatibility hack. We no longer target 16.04/core16

include/wayland/mir/wayland/wayland_base.h
```diff
@@ -215,6 +215,10 @@ class Global
     wl_global* const global;
 };
 
+/// Calls wl_client_post_implementation_error() if available
+[[gnu::format(printf, 2, 3)]]
+void post_implementation_error(wl_client* client, char const* fmt, ...);
+
 void internal_error_processing_request(wl_client* client, char const* method_name);
```
AlanGriffiths, line 222: The naming and functionality of these functions seems a bit unclear:

1. the "internal" function calls the not-"internal" function, it is usually the other way around.
2. everywhere we call `post_implementation_error()`, we also log the error (which is the difference in functionality)

It looks to me that the _only_ reason for `post_implementation_error()` to exist is the absence of an active exception when called from `WindowWlSurfaceRole::surface_destroyed()`.

AlanGriffiths (changes requested):
I'll PR an alternative approach.

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -441,15 +441,20 @@ void mf::WindowWlSurfaceRole::commit(WlSurfaceState const& state)
...
         // "When a client wants to destroy a wl_surface, they must destroy this 'role object' wl_surface"
         // NOTE: the wl_shell_surface specification seems contradictory, so this method is overridden in it's implementation
-        BOOST_THROW_EXCEPTION(std::runtime_error{
-            "wl_surface@" + std::to_string(wl_resource_get_id(surface->resource)) +
-            " destroyed before associated role"});
+        log_warning(
+            "wl_surface@%d destroyed before associated role",
+            wl_resource_get_id(surface->resource));
+        // Post error manually instead of throwing exception because this method is called by the surface destructor
+        wayland::post_implementation_error(
```
RAOF, line 454: This is not an implementation error; this is a client protocol violation, right?

wmww:
Superseded by #2057

AlanGriffiths:
> Superseded by #2057

PR #2057, opened 2021-06-04
Don't propagate exceptions from `~WlSurface()`

AlanGriffiths:
Alternative to  #2056. (Fixes: #2055)

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -33,7 +33,6 @@
 #include "mir/log.h"
 #include "mir/scene/surface.h"
 #include "mir/scene/surface_creation_parameters.h"
-#include "mir/wayland/wayland_base.h"
 
 #include <boost/throw_exception.hpp>
 
@@ -448,10 +447,9 @@ void mf::WindowWlSurfaceRole::surface_destroyed()
     {
         // "When a client wants to destroy a wl_surface, they must destroy this 'role object' wl_surface"
         // NOTE: the wl_shell_surface specification seems contradictory, so this method is overridden in it's implementation
-        BOOST_THROW_EXCEPTION((wayland::ProtocolError{
-            surface->resource,
-            std::numeric_limits<int32_t>::max(),    // There's no good error code: use a bad one
-            "wl_surface@%d destroyed before associated role", wl_resource_get_id(surface->resource)}));
+        BOOST_THROW_EXCEPTION(std::runtime_error{
+            "wl_surface@" + std::to_string(wl_resource_get_id(surface->resource)) +
+            " destroyed before associated role"});
     }
     else
     {
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -33,6 +33,7 @@
 #include "mir/log.h"
 #include "mir/scene/surface.h"
 #include "mir/scene/surface_creation_parameters.h"
+#include "mir/wayland/wayland_base.h"
 
 #include <boost/throw_exception.hpp>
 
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
Thanks clang! I learnt something today:
```
2021-06-04T09:39:55.8163591Z /spread/mir/src/server/frontend_wayland/wl_surface.cpp:113:43: error: cannot refer to a non-static member from the handler of a destructor function try block [-Werror,-Wexceptions]
2021-06-04T09:39:55.8165268Z     mw::internal_error_processing_request(client, "WlSurface::~WlSurface()");
```
> Before any catch clauses of a function-try-block on a destructor are entered, all bases and non-variant members have already been destroyed. 

I will fix.

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -102,8 +102,19 @@ mf::WlSurface::WlSurface(
...
+    // all bases and non-variant members have already been destroyed."
+    try
+    {
+        // Destroy the buffer stream first, as surface_destroyed() may throw
+        session->destroy_buffer_stream(stream);
+        role->surface_destroyed();
+    }
+    catch (...)
+    {
+        mw::internal_error_processing_request(client, "WlSurface::~WlSurface()");
```
RAOF, line 116: I think this (and the equivalent in William's PR) is wrong. This isn't (necessarily) a Mir implementation error - the client could be destroying surface roles in incorrect order, which is a client protocol violation not a Mir error.

If we ensure that `surface_destroyed` is throwing an appropriate subclass of `ProtocolException` we can just add a catch handler for that before this catch-all.

RAOF (changes requested):

src/server/frontend_wayland/wl_surface.cpp
AlanGriffiths: I agree this is a client protocol violation, but to throw a `ProtocolError` we need a suitable error code. `wl_surface` has "invalid_scale" and "invalid_transform" neither of which seems appropriate.

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -447,9 +448,10 @@ void mf::WindowWlSurfaceRole::surface_destroyed()
     {
         // "When a client wants to destroy a wl_surface, they must destroy this 'role object' wl_surface"
         // NOTE: the wl_shell_surface specification seems contradictory, so this method is overridden in it's implementation
-        BOOST_THROW_EXCEPTION(std::runtime_error{
-            "wl_surface@" + std::to_string(wl_resource_get_id(surface->resource)) +
-            " destroyed before associated role"});
+        BOOST_THROW_EXCEPTION((wayland::ProtocolError{
+            surface->resource,
+            std::numeric_limits<int32_t>::max(),    // There's no good error code: use a bad one
```
wmww, line 453: I'm not sure how I feel about this, but if we're going to do it this way `wl_subcompositor.cpp` should be changed to throw a protocol error as well.

AlanGriffiths: OK

AlanGriffiths:
Oh boy, how this has grown with each review.

AlanGriffiths:
No. I decided that sending `wl_surface` a protocol error in response to `wl_surface_destroy()` is a bad idea.

I think the original `wl_client_post_implementation_error()` is the least-bad option.

src/server/frontend_wayland/wl_surface.cpp
AlanGriffiths: On consideration, I don't think there's much point sending a protocol error to a deleted object. This is likely the least bad way to notify the problem.

wmww:
Considering clients can not gracefully handle protocols errors no matter what they are, it's not terribly important. I think this is the less confusing method.

wmww (approved):
LGTM, and I tested that it works with Squeekboard

bors r+

PR #2058, opened 2021-06-04
Remove backward compatibility for stuff that predates focal

AlanGriffiths:
The oldest series we need to support is focal (which has 1.18)

bionic-security & bionic-updates have 1.16. (For info pure bionic has 1.14)


RAOF (approved):
LGTM.

bors r+

PR #2068, opened 2021-06-10
Remove mirclient support

AlanGriffiths:
This removes all the traces of mirclient support that I could find. (And removes the protobuf dependency)

It does remove some tests around the remaining code, but I think it introduces a less expensive technical debt than what we had.


debian/libmirclient11.install
```diff
@@ -0,0 +1 @@
+usr/lib/*/libmirclient.so.11
```
wmww, line 1: Why is there still a libmirclient library?

AlanGriffiths: Because parts of the event and input code are there for historical reasons. I didn't want to mix moving it with what is mostly pure deletion. Also, we need to agree somewhere for it to go: mirplatform and mircommon are both possibilities.

src/client/mir_screencast.cpp
```diff
@@ -1,291 +0,0 @@
...
-    
-    std::unique_lock<decltype(mutex)> lk(mutex);
-
-    requests.emplace_back(std::make_unique<ScreencastRequest>(buffer, cb, context));
-
-    mir::protobuf::ScreencastRequest request;
-    request.set_buffer_id(buffer->rpc_id());
-    request.mutable_id()->set_value(protobuf_screencast->screencast_id().value());
-
-    server->screencast_to_buffer(
```
wmww, line 287: This is fine to drop because there isn't any actual screencast logic going on. Poking around did lead me to discover that the bulk of the implementation was dropped in 35ff6b619d8, which may be useful to look at when I start trying to implement it for Wayland again.

tests/unit-tests/shared_library_test.cpp
```diff
@@ -36,10 +36,10 @@ class SharedLibrary : public testing::Test
 public:
     SharedLibrary()
         : nonexistent_library{"imma_totally_not_a_library"},
-          existing_library{mtf::client_platform("dummy.so")},
+          existing_library{mtf::library_path() + "/libmircore.so"},
           nonexistent_function{"yo_dawg"},
-          existing_function{"create_client_platform"},
-          existent_version{MIR_CLIENT_PLATFORM_VERSION},
+          existing_function{"_ZN3mir17fatal_error_abortEPKcz"},
+          existent_version{"MIR_CORE_0.25"},
```
wmww, line 42: Is this change correct? Will It need to be updated for each mircore version bump?

wmww (approved):
Aside from the odd library loading test I commented on above, all looks very good! I'm fine with losing some test coverage for now if it means dropping all this code.

tests/unit-tests/shared_library_test.cpp
AlanGriffiths: Agreed, it would need changing if mircore ever broke ABI. But mircore is designated to be ABI stable. (And, so far, that has held. It is the only library we have that is "ABI unchanged at 1".)

RAOF (changes requested):
That's a gratifying amount of red. We loose some test coverage, but it's worth it.

Only blocking on the `wayland_connector->start()`/`->stop()` thing - either fixing it, or convincing me that it's correct. Otherwise, there's just even more you can remove!

src/server/display_server.cpp
```diff
@@ -176,15 +166,9 @@ struct mir::DisplayServer::Private
...
                 [&, this] { input_dispatcher->stop(); });
 
-            auto prompt = try_but_revert_if_unwinding(
-                [this] { prompt_connector->start(); },
-                [&, this] { prompt_connector->stop(); });
-
             auto wayland = try_but_revert_if_unwinding(
                 [this] { wayland_connector->start(); },
                 [&, this] { wayland_connector->stop(); });
-
```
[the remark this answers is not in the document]
AlanGriffiths: I agree it is misleading, but not wrong: there is no way that `wayland_connector->stop();` can be called (as there is no more code in this scope). Specifically, `[this] { wayland_connector->start(); }` is called in the constructor, so the destructor won't run if it throws (it would be wrong everywhere else if that were not the case).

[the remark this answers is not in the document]
Saviq: +1 from me.

Saviq (approved):
Approving just the last change, previous were ACKed already.

bors r=wmww,RAOF,Saviq

PR #2071, opened 2021-06-17
Add stream operator for miral::Window

wmww:
This makes window's correctly appear in test failures, and may have other uses. https://github.com/google/googletest/issues/1149 explains why the stream operators really need to be declared in the same place as the class. I tried a number of other ways, and this is the only one I found to be successful.

### DIFF
--- a/include/miral/miral/window.h
+++ b/include/miral/miral/window.h
@@ -79,7 +79,7 @@ bool operator==(Window const& lhs, Window const& rhs);
 bool operator==(std::shared_ptr<mir::scene::Surface> const& lhs, Window const& rhs);
 bool operator==(Window const& lhs, std::shared_ptr<mir::scene::Surface> const& rhs);
 bool operator<(Window const& lhs, Window const& rhs);
-auto operator<<(std::ostream& stream, const miral::Window& window) -> std::ostream&;
+auto operator<<(std::ostream& stream, miral::Window const& window) -> std::ostream&;
 
 inline bool operator!=(Window const& lhs, Window const& rhs) { return !(lhs == rhs); }
 inline bool operator!=(std::shared_ptr<mir::scene::Surface> const& lhs, Window const& rhs) { return !(lhs == rhs); }
--- a/include/miral/miral/window.h
+++ b/include/miral/miral/window.h
@@ -25,6 +25,7 @@
 #include <mir/geometry/size.h>
 
 #include <memory>
+#include <iosfwd>
 
 namespace mir
 {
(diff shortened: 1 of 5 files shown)


include/miral/miral/window.h
```diff
@@ -78,6 +79,7 @@ bool operator==(Window const& lhs, Window const& rhs);
 bool operator==(std::shared_ptr<mir::scene::Surface> const& lhs, Window const& rhs);
 bool operator==(Window const& lhs, std::shared_ptr<mir::scene::Surface> const& rhs);
 bool operator<(Window const& lhs, Window const& rhs);
+auto operator<<(std::ostream& stream, const miral::Window& window) -> std::ostream&;
```
AlanGriffiths, line 82: We use right const around here!

src/miral/window.cpp
```diff
@@ -125,3 +125,16 @@ bool miral::operator<(Window const& lhs, Window const& rhs)
 {
     return lhs.self.owner_before(rhs.self);
 }
+
+auto miral::operator<<(std::ostream& stream, const miral::Window& window) -> std::ostream&
```
AlanGriffiths, line 129: We use right const around here!

AlanGriffiths (approved):
Sure

AlanGriffiths:
CI say: "You need to add `miral::operator<<(std::basic_ostream<char, std::char_traits<char> >&, miral::Window const&)` to the symbols"

include/miral/miral/window.h
wmww: Oop

wmww:
Hmm looks like GitHub is broken

wmww:
I'm going to interpret "Sure" + approval as a bors delegate

bors r=AlanGriffiths

AlanGriffiths:
> I'm going to interpret "Sure" + approval as a bors delegate

Fair enough. (I'm not really convinced this is a widely useful streaming operator, nor that it has to be declared outside the test framework, but it is relatively harmless.)

wmww:
> I'm not really convinced this is a widely useful streaming operator, nor that it has to be declared outside the test framework

I can't say it has to be, but I put a bit of work into trying different things and this is the only one I found to actually work.

bors r=AlanGriffiths

AlanGriffiths:
OK, so the reason this is problematic is that in the `MIRAL_3.0` stanza of the version script (symbols.map) there's a wildcard that matches the new operator:
```plain
    miral::operator*;
```
Currently, that corresponds to:
```
miral::operator<(miral::Window const&, miral::Window const&)
miral::operator==(miral::Output::PhysicalSizeMM const&, miral::Output::PhysicalSizeMM const&)
miral::operator==(miral::Window const&, miral::Window const&)
miral::operator==(miral::Window const&, std::shared_ptr<mir::scene::Surface> const&)
miral::operator==(std::shared_ptr<mir::scene::Surface> const&, miral::Window const&)
```
So these need to be pulled in unambiguously (e.g. by using the mangled names).

wmww:
@RAOF Alan mentioned to may know some secret of escaping characters in symbols.map?

AlanGriffiths (changes requested):
Having poked around a bit I can't find a simple way to find this operator at the point of instantiation.

OTOH I think it would be better to provide a `PrintTo()` function for gtest, not a streaming operator that might be confused with something useful. That is, in `miral/window.h`:

```c++
// For Google test
void PrintTo(Window const& window, std::ostream* out);
```

AlanGriffiths:
Closing in favour of #2081

PR #2074, opened 2021-06-19
Add focus_mode surface property for Layer Shell

wmww:
This PR:
- Adds a new enum `MirFocusMode`
- Adds a `focus_mode` property to surface and surface specification types (in Mir and MirAL)
- Implements logic to respect the focus mode in the window manager
- Tests the window manager logic
- Uses the focus mode to implement keyboard interactivity in Layer Shell

Layer shell keyboard interactivity is tested by https://github.com/MirServer/wlcs/pull/203. This PR will probably fail against versions of WLCS prior to that, because they do not set keyboard interactivity correctly for the popup tests. A couple tests will probably need to be temporarily disabled until the WLCS PR gets merged and released, but I haven't done that yet.


wmww:
I have no idea how to fix this

AlanGriffiths (changes requested):
I think (pre-existing) the bad `parent` parameter name for `miral::BasicWindowManager::refocus()` the comment above the change I object to are confusing.

s/parent/hint/
s/Try to make the parent active/If there's a hint try using that first/

src/miral/basic_window_manager.cpp
```diff
@@ -274,11 +274,11 @@ void miral::BasicWindowManager::remove_window(Application const& application, mi
...
 
 void miral::BasicWindowManager::refocus(
-    miral::Application const& application, miral::Window const& parent,
+    miral::Application const& application, miral::Window const& hint,
     std::vector<std::shared_ptr<Workspace>> const& workspaces_containing_window)
 {
-    // Try to make the parent active
-    if (parent && parent == select_active_window(parent))
+    // If there's a hint, try using that first
+    if (hint && hint == select_active_window(hint))
```
AlanGriffiths, line 281:
```c++
if (hint && select_active_window(hint))
```

wmww: When I made this change non-interactive layer shell surfaces were being given focus after popups were destroyed, so I had to do some more things to make it work.

wmww:
What gives? I thought `libmiral4.symbols` was only supposed to cause trouble when an ambiguous symbol like an operator was added.

AlanGriffiths (approved):
LGTM

bors r+

PR #2078, opened 2021-06-23
Make GTK's approach to hiding cursors work on Mir

### LINKED ISSUE #2073: GTK on Wayland: cursor moves when it should be hidden
To reproduce:
- Open GTK app that hides cursor while typing (tested with gedit and lxterminal)
- Start typing with cursor over text area

Expected:
- Cursor disappears (as it does in Sway)

Actual:
- Cursor moves down and to the right by about one cursor size, but stays visible 

I *think* we would have noticed before if this never worked, but I don't know what broke it.

AlanGriffiths:
There was a logic error in `WlPointer::set_cursor()` that compared a pointer with an optional (both of which, unfortunately, convert to `bool`).

We also needed to hook into the "commit" on the corresponding surface to detect an attempt to unmap by sending a null buffer. It looks a bit messy, but fixes #2073.

### DIFF
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -444,7 +444,8 @@ WlSurfaceCursor::WlSurfaceCursor(mf::WlSurface* surface, geom::Displacement hots
       surface_role{surface, std::move(on_commit)},
       hotspot{hotspot}
 {
-    surface->set_role(&surface_role);
+    if (surface)
+        surface->set_role(&surface_role);
 
     stream->set_frame_posted_callback(
         [this](auto)
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -354,7 +354,7 @@ private:
 
     mw::Weak<mf::WlSurface> const surface;
     std::shared_ptr<mc::BufferStream> const stream;
-    CursorSurfaceRole surface_role; // Used only to assert unique ownership
+    CursorSurfaceRole surface_role;
 
     std::weak_ptr<ms::Surface> surface_under_cursor;
     geom::Displacement hotspot;
@@ -370,7 +370,7 @@ struct WlHiddenCursor : mf::WlPointer::Cursor
     void set_hotspot(geom::Displacement const&) override {};
     auto cursor_surface() const -> std::experimental::optional<mf::WlSurface*> override { return {}; };
 private:
-    CursorSurfaceRole surface_role; // Used only to assert unique ownership
+    CursorSurfaceRole surface_role;
 };
 }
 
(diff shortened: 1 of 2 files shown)


src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -337,17 +354,23 @@ struct WlSurfaceCursor : mf::WlPointer::Cursor
 
     mw::Weak<mf::WlSurface> const surface;
     std::shared_ptr<mc::BufferStream> const stream;
-    mf::NullWlSurfaceRole surface_role; // Used only to assert unique ownership
+    CursorSurfaceRole surface_role; // Used only to assert unique ownership
```
wmww, line 357: No longer `// Used only to assert unique ownership`

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -337,17 +354,23 @@ struct WlSurfaceCursor : mf::WlPointer::Cursor
...
 {
+    WlHiddenCursor(
+        mf::WlSurface* surface,
+        std::function<void(mf::WlSurface*)>&& on_commit);
+
     void apply_to(mf::WlSurface* surface) override;
     void set_hotspot(geom::Displacement const&) override {};
     auto cursor_surface() const -> std::experimental::optional<mf::WlSurface*> override { return {}; };
+private:
+    CursorSurfaceRole surface_role; // Used only to assert unique ownership
```
wmww, line 373: Same

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -359,36 +382,70 @@ void mf::WlPointer::set_cursor(
...
+        cursor = std::make_unique<WlHiddenCursor>(nullptr, [this](WlSurface* surface){ on_commit(surface); });
         if (surface_under_cursor)
             cursor->apply_to(&surface_under_cursor.value());
     }
 
     (void)serial;
 }
 
-WlSurfaceCursor::WlSurfaceCursor(mf::WlSurface* surface, geom::Displacement hotspot)
+
```
wmww, line 411: Extra newline

wmww (commented):
Does `on_commit` need to be a lambda? Seems like the only thing it's set to is `WlPointer::on_commit()`, so perhaps it would be simpler if it was given a pointer and called into that itself. If it's a public/private issue than perhaps `CursorSurfaceRole` should be nested inside `WlPointer`.

I'll need to take a closer look before I'm confident there's no chance of use-after-free's with buggy clients. I don't see much usage of `wayland::Weak` or other protections, which makes me suspicious there might be.

AlanGriffiths:
> Does `on_commit` need to be a lambda? Seems like the only thing it's set to is `WlPointer::on_commit()`, so perhaps it would be simpler if it was given a pointer and called into that itself. If it's a public/private issue than perhaps `CursorSurfaceRole` should be nested inside `WlPointer`.

It doesn't: See #2079 for an alternative. But I'm not sure it reads any better.

AlanGriffiths:
The alternate landed

PR #2079, opened 2021-06-24
Make GTK's approach to hiding cursors work on Mir

### LINKED ISSUE #2073: GTK on Wayland: cursor moves when it should be hidden

AlanGriffiths:
There was a logic error in WlPointer::set_cursor() that compared a pointer with an optional (both of which, unfortunately, convert to bool).

We also needed to hook into the "commit" on the corresponding surface to detect an attempt to unmap by sending a null buffer. It looks a bit messy, but fixes #2073.

(Alternative to #2078)

### DIFF
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -395,17 +395,14 @@ void mf::WlPointer::set_cursor(
     {
         auto const wl_surface = WlSurface::from(*surface);
         cursor_hotspot = {hotspot_x, hotspot_y};
-        if (cursor->cursor_surface() && wl_surface == *cursor->cursor_surface())
-        {
-            cursor->set_hotspot(cursor_hotspot);
-        }
-        else
+        if (!cursor->cursor_surface() || wl_surface != *cursor->cursor_surface())
         {
             cursor.reset(); // clean up old cursor before creating new one
             cursor = std::make_unique<WlSurfaceCursor>(wl_surface, cursor_hotspot, commit_handler);
             if (surface_under_cursor)
                 cursor->apply_to(&surface_under_cursor.value());
         }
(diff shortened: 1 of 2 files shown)


src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -321,11 +321,28 @@ void mf::WlPointer::maybe_frame()
...
+struct CursorSurfaceRole : mf::NullWlSurfaceRole
+{
+    mf::WlSurface* const surface;
+    mf::CommitHandler* const commit_handler;
+    explicit CursorSurfaceRole(mf::WlSurface* surface, mf::CommitHandler* commit_handler) :
+        NullWlSurfaceRole(surface),
+        surface{surface},
+        commit_handler{commit_handler} {}
+
+    void commit(mir::frontend::WlSurfaceState const& state) override
```
wmww, line 333: If `this` is destroyed before the surface, the role is not cleared and `commit()` may be called on a freed object.

wmww: Resolved by my commit

wmww (approved):
LGTM
bors delegate+

wmww:
One thing I noticed in manual testing: the cursor now flickers to the wrong position before hiding

wmww:
I think if the cursor is unchanged, we should wait until the next commit to apply the hotspot so it's done atomically.

AlanGriffiths:
Tweaks make sense

bors r=wmww

PR #2080, opened 2021-06-25
Standardize on keysym/key_code names

wmww:
Previously `key_code` was a badly overloaded name, referring to keysyms in some contexts and scan codes in other contexts. This PR standardizes the names across Mir to completely avoid the use of `key_code` (with the exception of the MirAL API, which keeps it in it's "keysym" meaning in order to remain stable).

### DIFF
--- a/src/common/symbols.map
+++ b/src/common/symbols.map
@@ -454,3 +454,11 @@ MIR_COMMON_2.5 {
   };
 
 } MIR_COMMON_0.26;
+
+MIR_COMMON_2.6 {
+ global:
+  extern "C++" {
+    MirKeyboardEvent::keysym*;
+    MirKeyboardEvent::set_keysym*;
+  };
+} MIR_COMMON_2.5;
--- a/src/common/symbols.map
+++ b/src/common/symbols.map
@@ -451,14 +451,7 @@ MIR_COMMON_2.5 {
     vtable?for?MirTouchEvent;
     vtable?for?MirPointerEvent;
     vtable?for?MirCloseSurfaceEvent;
-  };
-
-} MIR_COMMON_0.26;
-
-MIR_COMMON_2.6 {
- global:
-  extern "C++" {
     MirKeyboardEvent::keysym*;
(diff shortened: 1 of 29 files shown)


src/common/symbols.map
```diff
@@ -453,4 +453,12 @@ MIR_COMMON_2.5 {
     vtable?for?MirCloseSurfaceEvent;
   };
 
-} MIR_COMMON_0.26;
\ No newline at end of file
+} MIR_COMMON_0.26;
+
+MIR_COMMON_2.6 {
```
AlanGriffiths, line 458: No, current stuff will be in 2.5, we don't need 2.6. (And the rename isn't breaking ABI because we already did that this cycle)

src/platforms/x11/input/input_platform.cpp
```diff
@@ -389,19 +394,19 @@ void mix::XInputPlatform::process_input_event(xcb_generic_event_t* event)
         // Key repeats look like a release and an immediate press with the same timestamp. The only way to detect and
         // discard them is by peaking at the next event.
 
-        next_pending_event_callback = [this, keycode = release_ev->detail, time = release_ev->time](auto next_event)
+        next_pending_event_callback = [this, x11_keycode = release_ev->detail, time = release_ev->time](auto next_event)
```
AlanGriffiths, line 397: `xcb_keycode`?

AlanGriffiths (changes requested):
Non-blocking, but I'd be tempted to also introduce `miral::toolkit::mir_keyboard_event_keysym()` and deprecate `miral::toolkit::mir_keyboard_event_key_code()`.

wmww:
> I'd be tempted to also introduce miral::toolkit::mir_keyboard_event_keysym()

I started to do that, but decided it could go in a follow-up PR

src/platforms/x11/input/input_platform.cpp
wmww: Sure

AlanGriffiths (approved):
bors r+

PR #2081, opened 2021-06-25
Add gtest customization to print surface names in error messages

AlanGriffiths:
Better solution than #2071


debian/libmiral4.symbols
```diff
@@ -393,5 +393,6 @@ libmiral.so.4 libmiral4 #MINVER#
  MIRAL_3.1@MIRAL_3.1 3.1.0
  (c++)"miral::WaylandExtensions::zwlr_foreign_toplevel_manager_v1@MIRAL_3.1" 3.1.0
  MIRAL_3.2@MIRAL_3.2 3.2.0
- (c++)"miral::Output::logical_group_id()@MIRAL_3.2" 3.2.0
```
wmww, line 396: What happened here, why is this dropped?

wmww (approved):
Assuming the dropped symbol isn't a problem, looks good and works

bors delegate+

debian/libmiral4.symbols
AlanGriffiths: Because it caused FTBFS on my system.

I'm not sure why it was there (I don't think we ever had that symbol) and I don't know why builds ever worked with it listed as required.

AlanGriffiths:
bors r=wmww

PR #2085, opened 2021-06-28
Use less out-of-date cmake

AlanGriffiths:
There's a lot of archaic cmake around the project, and this chips away a piece.

Rather than setting public include directories in dependent targets they can be automated using `target_include_directories(... PUBLIC ...)`.

This makes a start with mircore and mircookie.

### DIFF
--- a/src/common/events/CMakeLists.txt
+++ b/src/common/events/CMakeLists.txt
@@ -36,6 +36,10 @@ add_library(
   ${EVENT_SOURCES}
 )
 
+target_link_libraries(mirevents
+    mircore
+)
+
 include_directories(
     ${PROJECT_SOURCE_DIR}/include/client
     ${PROJECT_SOURCE_DIR}/src/include/cookie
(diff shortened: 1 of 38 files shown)


AlanGriffiths:
For the curious, this "chip of archaic cmake" got the way of my efficiently moving most of the remaining mirclient code to mircommon.

AlanGriffiths:
> For the curious, this "chip of archaic cmake" got the way of my efficiently moving most of the remaining mirclient code to mircommon.

Plus it has bugged me for years.

wmww (approved):
LGTM
bors r+

src/common/events/CMakeLists.txt
```diff
@@ -36,6 +36,11 @@ add_library(
   ${EVENT_SOURCES}
 )
 
+target_link_libraries(mirevents
+  LINK_PUBLIC
```
RAOF, line 40: Is there a reason we're using `LINK_PUBLIC` rather than just `PUBLIC`? [`LINK_PUBLIC`](https://cmake.org/cmake/help/v3.16/command/target_link_libraries.html#libraries-for-a-target-and-or-its-dependents-legacy) is described as legacy.

PR #2086, opened 2021-06-29
Modernise CMake harder

RAOF:
A follow-on from #2085.

Use the non-legacy spelling of `LINK_PUBLIC`, and tell CMake we require 3.16


AlanGriffiths (approved):
bors merge

AlanGriffiths:
Why would `wlcs.MultiRectEdges/RegionSurfaceInputCombinations.*` timeout with clang? Just in case this is a one-off:

bors merge

PR #2087, opened 2021-06-29
Drop mirclient shared library completely

RAOF:
Now that we're not using protobuf we no longer have a process-wide
singleton that'll `abort()` if we accidentally include it twice!

This means that we can just build mirclient as a static library,
have the relevant bits linked in directly to the consumers, and
not bother installing it.

The last (user-visible) vestiges of the Old Republic have been
swept away!


AlanGriffiths:
This isn't the way I want to do it. It will end up breaking the ODR by pulling the same code into multiple consumers (I know this is more a theoretical problem than a practical one as I don't think there's much writing to things with static storage duration).

I'll put up my version for review today and we can discuss.

AlanGriffiths:
My alternative is #2088

AlanGriffiths:
The alternative landed

PR #2089, opened 2021-06-29
udev: Handle the first entry of an udev iterator being invalid

RAOF:
If a udev device is removed between the time we scan and the time
we try and iterate through it then the `device_from_syspath` call
will fail with a `std::runtime_error`.

We catch that on `++iterator` and `iterator++` but *not* if the
very first udev device has disappeared; in that case we crash
with an unhandled exception.

Extend the `increment()` handling to also handle the inital case,
and then use it to handle the case where the first udev device
has disappeared.

### DIFF
--- a/src/include/platform/mir/udev/wrapper.h
+++ b/src/include/platform/mir/udev/wrapper.h
@@ -113,7 +113,7 @@ public:
         iterator ();
         iterator (std::shared_ptr<Context> const& ctx, udev_list_entry* entry);
 
-        void increment();
+        void increment(udev_list_entry* start_from);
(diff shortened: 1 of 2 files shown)


RAOF:
This might fix https://github.com/MirServer/mir-kiosk/issues/60 ?

AlanGriffiths (approved):
LGTM

bors r+

AlanGriffiths:
What is this with the clang build? `wlcs.SurfaceInputRegions/SurfaceInputCombinations.*...Child aborted`

bors merge

PR #2092, opened 2021-06-29
Change input::Keymap to an interface

wmww:
As discussed this makes keymaps an abstract interface, and passes them around with shared_ptrs where it makes sense. Right now `ParameterKeymap` is the only real implementation, but for OSK support I will write an implementation that holds keymap text data (the way it comes in from Wayland clients).


wmww:
I also took the opportunity to drop the completely unused `Surface::set_keymap()` and the functions exclusively called by it..

include/common/mir/input/keymap.h
```diff
@@ -20,58 +20,38 @@
...
-        : model{model}, layout{layout}, variant{variant}, options{options}
-    {
-    }
+    virtual ~Keymap() = default;
 
-    std::string model{"pc105+inet"};
-    std::string layout{"us"};
-    std::string variant;
-    std::string options;
+    virtual auto operator==(Keymap const& other) const -> bool = 0;
```
AlanGriffiths, line 41: On principle, I don't like `operator==()` on an interface. The meaning is never clear as **interfaces do not have value semantics**. What does it mean semantically for it to be `virtual`?

I would prefer a good function name.

include/common/mir/input/mir_keyboard_config.h
```diff
@@ -34,14 +34,15 @@ struct MirKeyboardConfig
 {
     MirKeyboardConfig();
     ~MirKeyboardConfig();
-    MirKeyboardConfig(mir::input::Keymap&& keymap);
+    MirKeyboardConfig(std::shared_ptr<mir::input::Keymap> && keymap);
```
AlanGriffiths, line 37: Do we need an rvalue reference? Surely `MirKeyboardConfig(std::shared_ptr<mir::input::Keymap> keymap)` is at least as efficient.

src/common/input/parameter_keymap.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <stdexcept>
+#include <boost/throw_exception.hpp>
+#include <xkbcommon/xkbcommon.h>
+
+namespace mi = mir::input;
+
+auto mi::ParameterKeymap::operator==(Keymap const& other) const -> bool
+{
+    auto const params = dynamic_cast<ParameterKeymap const*>(&other);
+    return params &&
```
AlanGriffiths, line 30: OK, so this is implementing "equal iff rhs is a similar type (the same type, or derived from the same type) and has the same data values"? That isn't an equivalence relationship as it isn't symmetric.

Using `operator==()` is best kept for equivalence relationships.

AlanGriffiths (changes requested):
Looks mostly good. Apart from `operator==()` which is bad.

src/common/input/parameter_keymap.cpp
AlanGriffiths: For example, if we rename the existing implementation as `same(Keymap const& other)` you could have:
```C++
inline auto operator==(Keymap const& lhs, Keymap const& rhs)
{
    return lhs.same(rhs) && rhs.same(lhs);
}
```

include/common/mir/input/mir_keyboard_config.h
wmww: Done

include/common/mir/input/keymap.h
wmww: Done

src/common/input/parameter_keymap.cpp
wmww: Done

include/common/mir/input/keymap.h
```diff
@@ -20,58 +20,45 @@
...
-    {
-    }
+    virtual ~Keymap() = default;
 
-    std::string model{"pc105+inet"};
-    std::string layout{"us"};
-    std::string variant;
-    std::string options;
+    /// If other's type is the same as or derived from this's type, and it's values are equal
+    virtual auto matches(Keymap const& other) const -> bool = 0;
```
AlanGriffiths, line 42: OK, I know this is how I described the initial implementation. But, still on the "what does it actually mean" theme I don't think this is what is actually needed by the code that uses it (via `operator==()`).

There are three uses:
1. `mi::DefaultDevice::set_keyboard_configuration()` actually wants to know if the `XkbMappingState` has changed. It really doesn't care about the type.
2. `mf::WlSeat::ConfigObserver::device_added()/device_changed()/changes_complete()` wants to know about the equivalence of a potential call to `make_unique_xkb_keymap()`.
3. `MirKeyboardConfig::operator==()` is used by `MirInputDevice::operator==()` where it might be valid. I don't think the "different objects with the same dynamic type and value" is a significant case. Just "same object" looks like enough.

I think trying to make all these uses fit an equivalence relationship is confusing.

src/server/frontend_wayland/wl_seat.cpp
```diff
@@ -117,7 +117,7 @@ void mf::WlSeat::ConfigObserver::device_changed(std::shared_ptr<input::Device> c
 {
     if (auto keyboard_config = device->keyboard_configuration())
     {
-        if (*current_keymap != keyboard_config.value().device_keymap())
+        if (!current_keymap->matches(keyboard_config.value().device_keymap()))
         {
             pending_keymap = keyboard_config.value().device_keymap_shared();
         }
```
AlanGriffiths, line 123: I would be inclined to drop the conditionals here and put the test in `changes_complete()`.

One reason being, that if there are two calls to `device_added()/device_changed()` and the second matches `current_keymap` it gets dropped on the floor.

wmww: Good idea

AlanGriffiths (approved):
LGTM

bors r+

PR #2095, opened 2021-07-02
Fix a log error message

AlanGriffiths:
We were slicing the exception being reported by copying to the runtime type.

Before:
```
...Dynamic exception type: std::exception
std::exception::what: std::exception
```

After:
```
... /home/alan/CLionProjects/mir/src/platform/udev/udev_wrapper.cpp(60): Throw in function {anonymous}::DeviceImpl::DeviceImpl(udev_device*)
std::exception::what: Udev device does not exist
```


wmww (approved):
LGTM
bors r+

PR #2097, opened 2021-07-02
Fix InputDeviceHub threadsafety

wmww:
Reading through this code, I saw what appeared to be a number of threadsafety issues. I'm not sure if these were mistakes or it was determined that nothing would be using these classes in ways that cause race conditions. Either way, doing stuff from the Wayland thread for OSK support may introduce problems, so better get everything straight first.

A potential downside is that now some observer methods are called while under lock, which is generally not advisable. I don't *think* there's any case in which we need to change input devices from an observer call. Not calling the observer under lock would be nontrivial (and slightly less performant).

### DIFF
--- a/src/server/input/default_input_device_hub.cpp
+++ b/src/server/input/default_input_device_hub.cpp
@@ -53,6 +53,8 @@ struct mi::ExternalInputDeviceHub::Internal : InputDeviceObserver
 
     std::shared_ptr<ServerActionQueue> const observer_queue;
     ThreadSafeList<std::shared_ptr<InputDeviceObserver>> observers;
+
+    std::mutex mutex;
     std::vector<std::shared_ptr<Device>> devices_added;
     std::vector<std::shared_ptr<Device>> devices_changed;
     std::vector<std::shared_ptr<Device>> devices_removed;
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
> A potential downside is that now some observer methods are called while under lock, which is generally not advisable.

We have a design solution for that in the code already. Written up here:

https://accu.org/journals/overload/22/124/griffiths_2040/

(Our code evolved since, so some class names changed.

wmww:
It seems to me that just covers adding and removing observers. The core problem here is that the observer needs to be notified of existing input devices. If it's not all under lock, new input devices can be added that aren't notified about (the fall in the gap between the initial batch of notifications and normal notifications). I've attempted to solve this with a recursive mutex. Let me know more precisely if there's a better way.

src/server/input/default_input_device_hub.cpp
```diff
@@ -95,10 +101,10 @@ void mi::ExternalInputDeviceHub::remove_observer(std::weak_ptr<InputDeviceObserv
         data->observer_queue->enqueue_with_guaranteed_execution(
             [&,this]
             {
-                // Unless remove() is on the same thread as add() external synchronization would be needed
+                std::lock_guard<std::recursive_mutex> lock{data->mutex};
                 data->observers.remove(observer);
 
-                std::lock_guard<decltype(mutex)> lock{mutex};
+                std::lock_guard<decltype(mutex)> vc_lock{mutex};
```
AlanGriffiths, line 107: OK, I get there's a name clash with `lock` but where does `vc_` come from?

AlanGriffiths (approved):
There's a bit of weird naming and a confused comment, but otherwise looks OK.

bors delegate

AlanGriffiths:
bors merge

PR #2099, opened 2021-07-02
Catch exception setting keymap in MirAL

### LINKED ISSUE #2098: Deadlocks when invalid keymap set from miral
To reproduce:
- Add `Keymap("invalid"),` to the `runner.run_with(...)` call in `examples/miral-shell/shell_main.cpp`

Expected:
- Error is printed or even shell aborts

Actual:
- Deadlock

wmww:
Fixes #2098. I'm not sure exactly how the deadlock was happening, but this prevents it.

### DIFF
--- a/src/miral/keymap.cpp
+++ b/src/miral/keymap.cpp
@@ -156,7 +156,18 @@ struct miral::Keymap::Self : mir::input::InputDeviceObserver
             model = keyboard_config.value().device_keymap().model();
         }
         std::shared_ptr<mi::Keymap> keymap{std::make_shared<mi::ParameterKeymap>(model, layout, variant, options)};
-        keyboard->apply_keyboard_configuration(std::move(keymap));
+        try
+        {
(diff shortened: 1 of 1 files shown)


AlanGriffiths (changes requested):
I think this is in the wrong place doing the wrong thing.

What I would expect, is that the exception /1/ propagates to `run_with()` and /2/ the program exits with an error code.

AlanGriffiths:
I understand why this isn't behaving as I expect: the keymap doesn't fail when set during initialization, it fails when the keyboard device is added.

We should be detecting the problem earlier.

AlanGriffiths (changes requested):
I think the least bad place to handle the exception is in `device_added()` and have proposed that as #2103.

But that's only dealing with part of the problem. We should detect and report the bad configuration option before a device is plugged in and trying to apply it.

wmww:
Superseded by #2103

PR #2100, opened 2021-07-02
Drop legacy deprecation macros

AlanGriffiths:


include/core/mir_toolkit/events/enums.h
```diff
@@ -28,24 +28,20 @@ extern "C" {
...
-    mir_event_type_window = mir_event_type_surface,
+    mir_event_type_key __attribute__((deprecated("UNUSED since Mir 0.26"))),
+    mir_event_type_motion __attribute__((deprecated("UNUSED since Mir 0.26"))),
+    mir_event_type_window,
     mir_event_type_resize,
     mir_event_type_prompt_session_state_change,
     mir_event_type_orientation,
-    MIR_DEPRECATED_ENUM(mir_event_type_close_surface, "mir_event_type_close_window"),
-    mir_event_type_close_window = mir_event_type_close_surface,
+    mir_event_type_close_window,
```
wmww, line 37: It's unclear to me if this enum needs to remain stable or not. If so, doesn't dropping `mir_event_type_close_surface` brak it? If not, why not also drop the deprecated values like `mir_event_type_key`?

AlanGriffiths: The enum needs to remain stable to maintain API.

But dropping `mir_event_type_close_surface` shouldn't change the value of `mir_event_type_close_window` (nor what follows) as it is no longer being initialized.

wmww: Oh, rightrightright

wmww (approved):
bors r+

PR #2102, opened 2021-07-05
Add more keymap properties


AlanGriffiths:
In merging recent changes into the miroil branch I noticed that we can only retrieve the `model` attribute of the keymap. I don't see why this is correct or useful in isolation, so have added `layout`, `variant` and `options` (which are mentioned in that code).

If there is a better way for Lomiri to change the keymap let's discuss that.

### DIFF
--- a/include/common/mir_toolkit/mir_input_device.h
+++ b/include/common/mir_toolkit/mir_input_device.h
@@ -237,10 +237,10 @@ void mir_keyboard_config_set_keymap_options(
     MirKeyboardConfig* conf, char const* options);
 
 /**
- * Set the variant of the keymap as a null terminated string.
+ * Set the variant_ of the keymap as a null terminated string.
  *
  * \param [in] conf     The keyboard config
- * \param [in] variant  The keymap variant
+ * \param [in] variant  The keymap variant_
  */
 void mir_keyboard_config_set_keymap_variant(
     MirKeyboardConfig* conf, char const* variant);
--- a/src/common/input/parameter_keymap.cpp
+++ b/src/common/input/parameter_keymap.cpp
@@ -29,9 +29,9 @@ auto mi::ParameterKeymap::matches(Keymap const& other) const -> bool
     auto const params = dynamic_cast<ParameterKeymap const*>(&other);
     return params &&
         model_ == params->model_ &&
-        layout == params->layout &&
-        variant == params->variant &&
-        options == params->options;
+           layout_ == params->layout_ &&
+           variant_ == params->variant_ &&
+        options_ == params->options_;
 }
 
 auto mi::ParameterKeymap::model() const -> std::string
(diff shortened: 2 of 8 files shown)


include/common/mir_toolkit/mir_input_device.h
```diff
@@ -237,10 +237,10 @@ void mir_keyboard_config_set_keymap_options(
     MirKeyboardConfig* conf, char const* options);
 
 /**
- * Set the variant of the keymap as a null terminated string.
+ * Set the variant_ of the keymap as a null terminated string.
```
wmww, line 240: Global seach-replace of "variant" was perhaps a little overzealous

src/common/input/parameter_keymap.cpp
```diff
@@ -29,9 +29,9 @@ auto mi::ParameterKeymap::matches(Keymap const& other) const -> bool
     auto const params = dynamic_cast<ParameterKeymap const*>(&other);
     return params &&
         model_ == params->model_ &&
-        layout == params->layout &&
-        variant == params->variant &&
-        options == params->options;
+           layout_ == params->layout_ &&
+           variant_ == params->variant_ &&
```
wmww, line 33: Indentation issues

wmww (changes requested):
Couple minor issues, but I also don't think this PR is a good idea at large. Model was added because it was needed by miral, and because "virtual-keyboard" will not be an unreasonable value to return for it. I don't know that there is such reasonable behavior for the other parameters on keymaps that do not have real values for them.

When all properties are available it implies to me that it's always possible to pull the properties out of one keymap and construct an equivalent parameter keymap using them. For virtual keyboard keymaps there is no way for Mir to do this (because the parameters used to create the keymap will not be known to our process). Unless there's a way to extract parameters from a compiled keymap, we need to rethink any code that does this sort of thing.

EDIT: now that I think about it, there's a decent chance miral keymap setting will be broken for virtual keyboards because model is invalid. In any case it shouldn't be possible to set the keymap for virtual keybaords. `model()` will probably be removed eventually.

wmww:
I'm not sure exactly what sort of functionality this is going to be used for in miroil, but I suspect there's no way to implement it correctly for virtual keyboard keymaps no matter what the interface is. Perhaps a better solution would be to add these getters only to `ParameterKeymap` and then `dynamic_cast` keymaps to that to use them (with of course a fallback path for other types of keymaps)?

AlanGriffiths:
> I'm not sure exactly what sort of functionality this is going to be used for in miroil, but I suspect there's no way to implement it correctly for virtual keyboard keymaps no matter what the interface is. Perhaps a better solution would be to add these getters only to `ParameterKeymap` and then `dynamic_cast` keymaps to that to use them (with of course a fallback path for other types of keymaps)?

The specific code in miroil is: 
```c++
void miroil::InputDevice::apply_keymap(std::string const& layout, std::string const& variant)
{
    MirKeyboardConfig oldConfig;
    
    mir::input::Keymap keymap;
    if (device->keyboard_configuration().is_set()) { // preserve the model and options
        oldConfig = device->keyboard_configuration().value();
        keymap.model = oldConfig.device_keymap().model;
        keymap.options = oldConfig.device_keymap().options;
    }
    keymap.layout  = layout;
    keymap.variant = variant;

    device->apply_keyboard_configuration(std::move(keymap));
}
```
https://github.com/MirServer/mir/blob/84fe324bf95399e00230509e40c2a78bd27a78de/src/miroil/input_device.cpp#L39

Obviously, there are alternative solutions, but this got some feedback quickly.

AlanGriffiths:
I'm going to try applying "tell, don't ask" to this problem...

PR #2103, opened 2021-07-05
Prevent exceptions propagating where they are not expected

AlanGriffiths:
This mitigates #2098, but a real fix would also detect and report setting an invalid keymap.

### DIFF
--- a/src/miral/keymap.cpp
+++ b/src/miral/keymap.cpp
@@ -176,7 +176,7 @@ struct miral::Keymap::Self : mir::input::InputDeviceObserver
     catch (...)
     {
         mir::log(mir::logging::Severity::warning, MIR_LOG_COMPONENT, std::current_exception(),
-            "problem remove device (" + device->name() + ")");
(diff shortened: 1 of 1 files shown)


src/miral/keymap.cpp
```diff
@@ -160,12 +166,18 @@ struct miral::Keymap::Self : mir::input::InputDeviceObserver
...
     {
         std::lock_guard<decltype(mutex)> lock{mutex};
 
         if (mir::contains(device->capabilities(), mir::input::DeviceCapability::keyboard))
             keyboards.erase(std::find(begin(keyboards), end(keyboards), device));
     }
+    catch (...)
+    {
+        mir::log(mir::logging::Severity::warning, MIR_LOG_COMPONENT, std::current_exception(),
+            "problem remove device (" + device->name() + ")");
```
wmww, line 179: s/remove/removing

wmww (approved):
LGTM
bors r+

PR #2104, opened 2021-07-06
Revert #2092

AlanGriffiths:
Reverts recent commit that causes a segfault when a client connects


AlanGriffiths:
Easy way to test:

Run miral-app and press Ctrl-Alt-T. Mir then segfaults. (Dereferences a null pointer to provide `initial_keymap` to `WlKeyboard::WlKeyboard`).

AlanGriffiths:
We got a fix

wmww:
There is currently an intermittent deadlock on master and this seems to make it less common, but does not fix it

PR #2106, opened 2021-07-06
Do not use keymap after move

wmww:
Fixes same issue as #2104 without reverting


AlanGriffiths (approved):
WFM

bors merge

wmww:
bors cancel

wmww:
Cancelled because there is a deadlock on master this doesn't fix. However it appears that is unrelated, so merging

bors r=AlanGriffiths

PR #2109, opened 2021-07-07
Enable changing the keyboard layout

AlanGriffiths:
Lomir expects to be able to set the layout of a keyboard. The code in miroil (that worked with the old Keymap struct is:

```c++
void miroil::InputDevice::apply_keymap(std::string const& layout, std::string const& variant)
{
    MirKeyboardConfig oldConfig;
    
    mir::input::Keymap keymap;
    if (device->keyboard_configuration().is_set()) { // preserve the model and options
        oldConfig = device->keyboard_configuration().value();
        keymap.model = oldConfig.device_keymap().model;
        keymap.options = oldConfig.device_keymap().options;
    }
    keymap.layout  = layout;
    keymap.variant = variant;

    device->apply_keyboard_configuration(std::move(keymap));
}
```

### DIFF
--- a/include/common/mir/input/keymap.h
+++ b/include/common/mir/input/keymap.h
@@ -44,9 +44,6 @@ public:
     virtual auto model() const -> std::string = 0;
     virtual auto make_unique_xkb_keymap(xkb_context* context) const -> XKBKeymapPtr = 0;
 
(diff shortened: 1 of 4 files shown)


wmww (approved):
Sure
bors r+

PR #2111, opened 2021-07-08
Virtual keyboard protocol

### LINKED ISSUE #2063: Support on-screen keyboards
## Rationale
With the exception of Lomiri, Mir-based compositors can not currently use an on-screen keyboard (OSK). To make Mir usable in more contexts we should support this.

## Specification
To save development costs and harmonize with the rest of the Wayland ecosystem, we should not build an OSK specifically for Mir. Instead, we should implement existing protocols so current and future OSKs are compatible.

The most fully featured Wayland OSK currently appears to be [Squeekboard](https://source.puri.sm/Librem5/squeekboard), which was developed by Purism for the Librem 5. It communicates with the display server and other apps using Wayland protocol extensions and D-Bus. At least one extension new will need to be implemented in Mir to make it work (see below)

## Further Information
Squeekboard can be easily installed from the AUR on Arch. I haven’t tried on Ubuntu or manually built from source, but it doesn't seem it would be difficult. It works on Sway, but not Mir because:

```
** (squeekboard:670444): ERROR **: 17:21:48.254: No virtual keyboard manager Wayland global available.
```

On Sway the keyboard doesn’t automatically pop up and disappear (as it does on phones running Phosh). Instead, I have to make it appear with:

```
busctl call --user sm.puri.OSK0 /sm/puri/OSK0 sm.puri.OSK0 SetVisible b true
```

It appears there are four Wayland protocol extensions used by Squeekboard:
- `wlr_layer_shell_unstable_v1`: used for positioning the keyboard surface, already implemented in Mir.
- `virtual_keyboard_unstable_v1`: used to send keymaps and key presses from the keyboard to the server, not implemented and required.
- `text_input_unstable_v3`: I think this gives app clients (ie not the OSK) additional context about the user’s input. I don’t completely understand how this protocol works or what it’s for. More investigation is needed. May not be required for functional proof-of-concept.
- `input_method_unstable_v2`: my current understanding is that the input method client (in this case an OSK) talks to the server with this protocol, and the server then relays those messages to the client using text_input_unstable_v3. If that’s correct, it will make sense to implement both at the same time, as they are not useful without each other. May not be required for functional proof-of-concept.

wmww:
Closes #2063. Includes a bunch of stuff that will land in individual PRs before this one is marked ready.

To test:
- Run mir something like this:
```
WAYLAND_DISPLAY=mir ./bin/miral-shell --add-wayland-extensions='zwlr_layer_shell_v1:zwp_virtual_keyboard_manager_v1'
```
- Run squeekboard on Mir
```
WAYLAND_DISPLAY=mir squeekboard
```
- Show the OSK using a D-Bus message
```
busctl call --user sm.puri.OSK0 /sm/puri/OSK0 sm.puri.OSK0 SetVisible b true
```
Type into a window!

I still need to look into how to make automatic show/hide work


AlanGriffiths:
> I still need to look into how to make automatic show/hide work

There must be a request from the client when focus goes to text input, and something decides if OSK is needed. But is that something Mir? Or bespoke in the shell? Or a service?

For some usecases, just having the OSK on always is enough. For others a swipe-up gesture.

RAOF:
> > I still need to look into how to make automatic show/hide work
> 
> There must be a request from the client when focus goes to text input, and something decides if OSK is needed. But is that something Mir? Or bespoke in the shell? Or a service?

Hey, aren't you meant to be on holiday 😛.

I'm reasonably sure that's [text-input](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/unstable/text-input/text-input-unstable-v3.xml) + [input-method](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/unstable/input-method/input-method-unstable-v1.xml)

wmww:
> I'm reasonably sure that's text-input + input-method

That's what I'm hoping

AlanGriffiths:
A small note about squeekboard it isn't in 20.04 archives, but is in 21.04 on:

```plain
$ rmadison squeekboard
 squeekboard | 1.12.0-1 | hirsute/universe | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 squeekboard | 1.12.0-1 | impish/universe  | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
```

src/server/frontend_wayland/virtual_keyboard_v1.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+namespace
+{
+auto unique_keyboard_name() -> std::string
+{
+    static std::mutex mutex;
+    static int id = 0;
+    std::lock_guard<std::mutex> lock{mutex};
+    auto const result = "virt-key-" + std::to_string(id);
+    id++;
```
AlanGriffiths, line 53: This could simply be an atomic `next_id` with `fetch_add(1)` with no need for a mutex.

src/wayland/symbols.map
```diff
@@ -401,3 +401,22 @@ global:
     virtual?thunk?to?mir::wayland::RelativePointerV1::?RelativePointerV1*;
   };
 } MIRWAYLAND_2.1;
+
+MIRWAYLAND_2.2.2 {
```
AlanGriffiths, line 405: This should match the Mir release: MIRWAYLAND_2.5

AlanGriffiths (changes requested):
A couple of small things. Seems to work.

And, seemingly, we could do a Frame snap with a "fixed" OSK. (Just not "on demand".)

wmww:
In a possibly controversial change, I've decided we should stop sticking "manager" on the end of protocol names. Can revert last commit if you think it's very important that protocols have the same name as their globals. Seems like code noise to me.

AlanGriffiths (approved):
LGTM

bors r+

AlanGriffiths:
> I've decided we should stop sticking "manager" on the end of protocol names

Yeah, not sure how we started that. Likely an accident.

PR #2112, opened 2021-07-12
InputDeviceRegistry: return device handle when adding device

wmww:
This will allow the thing that created a device to set it's config (specifically, allow for keymaps to be set). Part of #2111.

### DIFF
--- a/include/platform/mir/input/input_device_registry.h
+++ b/include/platform/mir/input/input_device_registry.h
@@ -35,7 +35,7 @@ public:
     InputDeviceRegistry() = default;
     virtual ~InputDeviceRegistry() = default;
 
-    virtual auto add_device(std::shared_ptr<InputDevice> const& device) -> std::shared_ptr<Device> = 0;
+    virtual auto add_device(std::shared_ptr<InputDevice> const& device) -> std::weak_ptr<Device> = 0;
     virtual void remove_device(std::shared_ptr<InputDevice> const& device) = 0;
 protected:
     InputDeviceRegistry(InputDeviceRegistry const&) = delete;
--- a/include/platform/mir/input/input_device_registry.h
+++ b/include/platform/mir/input/input_device_registry.h
@@ -26,6 +26,7 @@ namespace mir
 {
 namespace input
 {
+class Device;
 class InputDevice;
 
 class InputDeviceRegistry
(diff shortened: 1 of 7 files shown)


include/platform/mir/input/input_device_registry.h
```diff
@@ -34,7 +35,7 @@ class InputDeviceRegistry
     InputDeviceRegistry() = default;
     virtual ~InputDeviceRegistry() = default;
 
-    virtual void add_device(std::shared_ptr<InputDevice> const& device) = 0;
+    virtual auto add_device(std::shared_ptr<InputDevice> const& device) -> std::shared_ptr<Device> = 0;
```
AlanGriffiths, line 38: I'm tempted by the alternative design:
```c++
     virtual auto add_device_with_config(std::shared_ptr<InputDevice> const& device, std::function<void(Device& device)> const& configure) = 0;
```
Howeever, I recognise the principle I'm following (of not pointlessly handing out ownership of `Devices`) hasn't been followed in the `InputDeviceObserver` interface, so I won't insist.

AlanGriffiths (approved):
I think we'd be marginally better with the suggested "tell don't ask" interface, but I won't block.

bors delegate=@wmww

include/platform/mir/input/input_device_registry.h
wmww: How would the user of this interface handle the lifecycle of devices? Would they be assumed to keep track of when they destroy a device and not used the `Device` reference after that? Perhaps using a `weak_ptr` (either returned from the function or as an argument to an argument function) would be a better option?

AlanGriffiths: They would be expected to use the device reference during the call (where it is valid) and not to keep a reference/pointer to something they don't own.

AlanGriffiths: Having looked a bit deeper, there's `DeviceChangeTracker` that shares ownership of `Devices`, so my idea that ownerships shouldn't be shared doesn't fit the existing usage pattern.

It still seems odd that `add(foo)` gives ownership of a `bar` (that isn't used as a handle for `remove()`).

In practice, the "configure" code here doesn't keep ownership of the device, so it isn't clear why you want to share ownership this way. If, hypothetically, ownership were retained, you don't get any notification that the device being configured has been removed, so how do you know to release the `Device` returned?

wmww: I've poked around a bit more. #2111 does store the handle and use it later, so returning/having a function called with a reference wont cut it. It shouldn't return ownership though, so I've changed it to returning a `weak_ptr`. I can make it send the device to a provided function if you want, but that adds syntax clutter and (by my understanding) doesn't have any benefits when ownership is handled by smart pointers.

AlanGriffiths (approved):
I don't agree that `weak_ptr` doesn't share ownership. But...

bors r+

PR #2113, opened 2021-07-12
Use shared pointers to move around keymaps

wmww:
In order to property manage and move around keymaps more things will need to hold on to them, so making them available as a shared pointer makes sense. Part of #2111.


AlanGriffiths (approved):
LGTM

bors r+

PR #2114, opened 2021-07-12
MirKeyboardEvent: add keymap property

wmww:
Allow setting the keymap for a key event. In a subsequent PR the `KeyMapper` will set this property and the Wayland frontend will use it to determine if a new keymap needs to be sent.


AlanGriffiths (approved):
Ugly but consistent with the surrounding code.

bors r+

PR #2116, opened 2021-07-12
Send keymaps as needed

wmww:
When a key event comes through with a different keymap, send the keymap event to Wayland clients before sending the key. Based on #2114 and #2113. Part of #2111.


AlanGriffiths:
> Based on #2114 and #2113. Part of #2111.

It is better to set the target branch as the pre-requisite. Bors will sort out retargetting when that lands (instead of github detecting a merge conflict)

wmww:
Ok, I'll do that from now on. Prereqs have merged, so ready for review.

AlanGriffiths (approved):
LGTM

bors r+

PR #2117, opened 2021-07-13
Calibrate event timestamps

wmww:
With this PR, `DefaultEventBuilder` holds a `time::Clock`, which is uses to calibrate incoming events with. This way events coming from multiple sources (each with an undefined timestamp base) and events generated inside Mir will all have correct timestamps relative to each other. The `EventBuilder` interface is not changed, and no changes are needed for users of it.

### DIFF
--- a/include/platform/mir/input/event_builder.h
+++ b/include/platform/mir/input/event_builder.h
@@ -41,6 +41,11 @@ public:
     /// Timestamps in returned events are automatically calibrated to the Mir clock
     using Timestamp = std::chrono::nanoseconds;
 
+    /// True resets timestamp calibration to it's initial state, so the next event with a timestamp will be at the
+    /// "current" time (according to the Mir clock) and subsequent events will be relative to that. False disables
+    /// calibration, so future events will not have their timestamps modified.
+    virtual void calibrate_timestamps(bool enable) = 0;
+
     virtual EventUPtr key_event(Timestamp timestamp, MirKeyboardAction action, xkb_keysym_t keysym, int scan_code) = 0;
 
     virtual EventUPtr pointer_event(Timestamp timestamp, MirPointerAction action, MirPointerButtons buttons_pressed,
--- a/include/platform/mir/input/event_builder.h
+++ b/include/platform/mir/input/event_builder.h
@@ -41,10 +41,8 @@ public:
     /// Timestamps in returned events are automatically calibrated to the Mir clock
     using Timestamp = std::chrono::nanoseconds;
 
-    /// True resets timestamp calibration to it's initial state, so the next event with a timestamp will be at the
-    /// "current" time (according to the Mir clock) and subsequent events will be relative to that. False disables
-    /// calibration, so future events will not have their timestamps modified.
-    virtual void calibrate_timestamps(bool enable) = 0;
+    /// Disables calibration, so future events will not have their timestamps modified
+    virtual void disable_timestamp_calibration() = 0;
 
     virtual EventUPtr key_event(Timestamp timestamp, MirKeyboardAction action, xkb_keysym_t keysym, int scan_code) = 0;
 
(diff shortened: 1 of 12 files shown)


wmww:
#2111 will behave more correctly once this lands (though if there don't seem to be any visible symptoms of timestamp problems)

include/platform/mir/input/event_builder.h
```diff
@@ -38,8 +38,14 @@ class EventBuilder
 public:
     EventBuilder() = default;
     virtual ~EventBuilder() = default;
+    /// Timestamps in returned events are automatically calibrated to the Mir clock
     using Timestamp = std::chrono::nanoseconds;
 
+    /// True resets timestamp calibration to it's initial state, so the next event with a timestamp will be at the
+    /// "current" time (according to the Mir clock) and subsequent events will be relative to that. False disables
+    /// calibration, so future events will not have their timestamps modified.
+    virtual void calibrate_timestamps(bool enable) = 0;
```
AlanGriffiths, line 47: Not sure about this name (too close to `calibrate_timestamp()`), nor the `bool` parameter (which is always `false`).

tests/mir_test_framework/fake_input_device_impl.cpp
```diff
@@ -324,6 +324,9 @@ void mtf::FakeInputDeviceImpl::InputDevice::map_touch_coordinates(float& x, floa
 
 void mtf::FakeInputDeviceImpl::InputDevice::start(mi::InputSink* destination, mi::EventBuilder* event_builder)
 {
+    // We keep track of events based on their timestamps, so we can't have those being modified
+    event_builder->calibrate_timestamps(false);
+
```
AlanGriffiths, line 329: Is this the only use of `calibrate_timestamps()`? Wouldn't it be better to supply a test double as the clock?

wmww: That's a little inconvenient, because the code where it's obvious that we need to do this doesn't create the builder.

include/platform/mir/input/event_builder.h
wmww: Better?

include/platform/mir/input/event_builder.h
```diff
@@ -38,8 +38,12 @@ class EventBuilder
 public:
     EventBuilder() = default;
     virtual ~EventBuilder() = default;
+    /// Timestamps in returned events are automatically calibrated to the Mir clock
     using Timestamp = std::chrono::nanoseconds;
 
+    /// Disables calibration, so future events will not have their timestamps modified
+    virtual void disable_timestamp_calibration() = 0;
```
AlanGriffiths, line 45: Is intended functionality? Or a convenience for some tests?

tests/mir_test_framework/fake_input_device_impl.cpp
AlanGriffiths: It doesn't create the builder passed to `start()`, but it _could_ create the builder used by `FakeInputDeviceImpl`.

I'm not sure it is worth the effort, but I'd like clarity on the purpose of this method.

wmww:
Now `FakeInputDeviceImpl` sets event timestamps itself, so change to the event builder interface is not needed. I considered making WLCS remember post-calibration timestamps as we discussed, but that's rather impractical.

include/platform/mir/input/event_builder.h
```diff
@@ -38,6 +38,7 @@ class EventBuilder
 public:
     EventBuilder() = default;
     virtual ~EventBuilder() = default;
+    /// Timestamps in returned events are automatically calibrated to the Mir clock
     using Timestamp = std::chrono::nanoseconds;
```
AlanGriffiths, line 42: It is rather a shame to introduce another `Timestamp` alias. (C.f. `mir/time/types.h`).

Particularly, as elsewhere this is known as `mir::time::Duration`

AlanGriffiths (approved):
bors delegate=@wmww

include/platform/mir/input/event_builder.h
wmww: Maybe this could use some sorting out. The values being represented are timestamps, not durations, so using that alias would be incorrect. `mir::Timestamp` should probably be used across Mir, but that refactor is out of scope here.

PR #2118, opened 2021-07-13
X11 platform: use synthisized timestamps instead of last timestamp

wmww:
This changes the `EventBuilder` interface to take optional timestamps, and to synthesize a timestamp from current time if none is given. It also uses that in the X11 platform where before we were using the last timestamp we saw. Stacked on top of #2117.

### DIFF
--- a/include/platform/mir/input/event_builder.h
+++ b/include/platform/mir/input/event_builder.h
@@ -25,6 +25,7 @@
 #include <memory>
 #include <chrono>
 #include <vector>
+#include <optional>
 
 namespace mir
 {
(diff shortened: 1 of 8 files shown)


AlanGriffiths:
Needs rebasing on the updated #2117

wmww:
Done

include/platform/mir/input/event_builder.h
```diff
@@ -38,21 +39,34 @@ class EventBuilder
 public:
     EventBuilder() = default;
     virtual ~EventBuilder() = default;
-    /// Timestamps in returned events are automatically calibrated to the Mir clock
+    /// Timestamps in returned events are automatically calibrated to the Mir clock. If nullopt is given current mir
+    /// clock time is used.
     using Timestamp = std::chrono::nanoseconds;
```
AlanGriffiths, line 44: I don't think this documentation belongs to this type.

This `Timestamp` type represent a timestamp (implemented as a duration from an implicit epoch). 

1. "calibration" is changing to a different epoch, but isn't done by `Timestamp`; and,
2. a `Timestamp` cannot be set to `nullopt`.

src/platforms/x11/input/input_device.h
```diff
@@ -39,6 +40,8 @@ namespace X
 class XInputDevice : public input::InputDevice
 {
 public:
+    using EventTime = std::optional<std::chrono::nanoseconds>;
```
AlanGriffiths, line 43: Wouldn't this make more sense shared with `EventBuilder`?

AlanGriffiths (approved):
OK, I'll consider the issue with type names and documentation as pre-existing.

bors r+

PR #2119, opened 2021-07-14
WlTouch: use Clock to synthisize timestamps

wmww:
Instead of the old, hacky system of using `steady_clock` with an offset. Most of the code is getting the clock to the `WlTouch`. May not produce correct behavior until #2117 lands, as it assumes event timestamps have the same base as the Mir clock.


AlanGriffiths (approved):
LGTM

bors r+

PR #2120, opened 2021-07-23
Improve message when snap recipe not found

AlanGriffiths:


Saviq (approved):
Yeah it's because LP itself doesn't call them recipes...

bors merge

PR #2123, opened 2021-07-27
Fix open() interposer

RAOF:
So, `open()` is actually a varargs function, taking an optional `mode_t` third parameter.

We've always interposed this with a `open(char const*, int, mode_t)` function, which is
a) Obviously not varargs, and so any calls to `open()` were being handled by a function with incompatible signature, but
b) We never hit the UB this introduced… until now.

Fixes the build on ppc64el. Tested on ppc64el and x86-64.


AlanGriffiths (approved):
WFM

bors r+

PR #2124, opened 2021-07-27
Groovy is EOL

AlanGriffiths:


Saviq:
We should include `hirsute` instead and flip `devel` to `impish`, then.

Saviq (approved):
Yup.

bors merge

PR #2127, opened 2021-07-28
WlSeat: improve focus listeners

wmww:
This renames and refactors `WlSeat::ListenerTracker` to `WlSeat::FocusListener`. Focus listeners are tied to a single client and only get notifications relevant to that client, and they are told what surface is focused. These things will help implement `zwp_text_input_v3`, as it needs to respond to changing focus.


AlanGriffiths (approved):
Not "obviously no problems" but neither was the previous code. (I can't quite convince myself there are no significant changes in behaviour.) So let's land it and use it for a while.

bors r+

PR #2128, opened 2021-07-28
Remove more vestigial mirclient support

RAOF:
The `PlatformAuthentication` interface was required for supporting mirclient's “give me a DRM fd I can render with” extension API.
That's long gone.
Drop the associated code.


AlanGriffiths (approved):
I see red!

bors r+

Issue #209, opened 2018-03-01
We have DISABLED tests


AlanGriffiths:
These should be reworked and enabled (or obsoleted)

AlanGriffiths:
Most of these come from working around these issues:

- #138 NestedInputWithMouse.mouse_pointer_position_is_in_sync_with_host_server unstable
- #147 [ FAILED ] NestedInput.*
- #148 NestedServer.given_nested_server_set_base_display_configuration_when_monitor_plugged_in_configuration_is_reset

PR #2129, opened 2021-07-28
Clean up our DISABLED tests

AlanGriffiths:
Remove or reinstate DISABLED tests.

There remain some conditionally DISABLED tests, but these are still running in most configurations.

Fixes: #209


RAOF (approved):
Yup, those seem reasonable.

bors r+

PR #2132, opened 2021-07-29
WlKeyboard: if refocus is needed, send key event after it completes

wmww:
If we got a key event to the surface that the keyboard isn't currently on, we used to send the event then refocus the keyboard. This corrects that order.

1. Open LXTerminal
2. Go to Tabs -> Name Tab
3. Click Cancel
4. Type key

__master:__ first letter typed doesn't appear in main window
__this PR:__ first letter does appear

### DIFF
--- a/src/server/frontend_wayland/wl_keyboard.cpp
+++ b/src/server/frontend_wayland/wl_keyboard.cpp
@@ -73,8 +73,6 @@ mf::WlKeyboard::~WlKeyboard()
 
 void mf::WlKeyboard::event(MirKeyboardEvent const* event, WlSurface& surface)
 {
-    set_keymap(event->keymap());
-
     bool down;
     switch (mir_keyboard_event_action(event))
     {
@@ -94,20 +92,8 @@ void mf::WlKeyboard::event(MirKeyboardEvent const* event, WlSurface& surface)
     auto const timestamp = mir_input_event_get_wayland_timestamp(mir_keyboard_event_input_event(event));
     int const scancode = mir_keyboard_event_scan_code(event);
     uint32_t const wayland_state = down ? KeyState::pressed : KeyState::released;
-    send_key_event(serial, timestamp, scancode, wayland_state);
 
-    if (as_nullable_ptr(focused_surface) == &surface)
-    {
-        /*
-         * HACK! Maintain our own XKB state, so we can serialise it for
-         * wl_keyboard_send_modifiers
-         */
-        xkb_key_direction const xkb_state = down ? XKB_KEY_DOWN : XKB_KEY_UP;
-        xkb_state_update_key(state.get(), scancode + 8, xkb_state);
-
-        update_modifier_state();
-    }
-    else
+    if (focused_surface != surface)
     {
         log_warning(
             "Sending key 0x%2.2x %s to wl_surface@%u even though it was not explicitly given keyboard focus",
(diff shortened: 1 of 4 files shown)


src/server/frontend_wayland/wl_keyboard.cpp
```diff
@@ -94,27 +92,25 @@ void mf::WlKeyboard::event(MirKeyboardEvent const* event, WlSurface& surface)
...
-         * HACK! Maintain our own XKB state, so we can serialise it for
-         * wl_keyboard_send_modifiers
-         */
-        xkb_key_direction const xkb_state = down ? XKB_KEY_DOWN : XKB_KEY_UP;
-        xkb_state_update_key(state.get(), scancode + 8, xkb_state);
-
-        update_modifier_state();
-    }
-    else
+    if (focused_surface != surface)
```
RAOF, line 96: Um, how does this work? As far as I can tell `WlSurface` doesn't implement `operator==` or `operator!=`, so what is this comparing?

RAOF (commented):
Seems reasonable, but how does it work?!

src/server/frontend_wayland/wl_keyboard.cpp
AlanGriffiths: `focused_surface` is a `Weak<>`, which has member function: `auto operator==(T const& other) const -> bool`.

Personally, I strongly prefer `operator==()` to be a non-member, and implement an equivalence relationship. (Neither of which is true for this case.)

Maybe the semantics of this `operator==()` are sufficiently confusing we should have objected when this was proposed (#1614).

But that's how this works.

AlanGriffiths:
Looks reasonable, but I cannot reproduce the problem this claims to fix.

> **master:** first letter typed doesn't appear in main window
> **this PR:** first letter does appear

What I see:

**master**: first letter typed appears in main window
**this PR**: first letter typed appears in main window

wmww:
> I cannot reproduce the problem this claims to fix

If you hit ESC instead of clicking cancel it works correctly on master. Other than that I'm not sure why it wouldn't repro unless something in LXTerminal changed. I haven't found an easy way to expose this problem in other apps.

Even if you can't reproduce, I think the new logic is clearly more correct. There's also reasons relating to implementing input method to do this, if it doesn't cause any trouble.

I agree comparing wayland::Weaks is sub-optimal. Will make a PR that improves that.

AlanGriffiths:
> If you hit ESC instead of clicking cancel it works correctly on master. Other than that I'm not sure why it wouldn't repro unless something in LXTerminal changed. I haven't found an easy way to expose this problem in other apps.

I have been pressing ESC

...
> I agree comparing wayland::Weaks is sub-optimal. ~Will make a PR that improves that.~
> 
> Edit I changed the `Weak` API in this PR. Doing an equality operator correctly seemed like more trouble and likely to introduce footguns (ie implicitly converting to bool in weird cases), so I added an `is` method. Equality between two `Weak`s is still implemented as before. what do you think?

Big improvement.

AlanGriffiths (approved):
I would prefer to see this change something. But it seems harmless.

bors r+

wmww:
> > If you hit ESC instead of clicking cancel it works correctly on master

> I have been pressing ESC

Sorry, this might have been confusing

You *should* be hitting cancel, because the bug this PR fixed did not show up when you hit ESC

You can test if you want, but now that it's merged doesn't matter much

PR #2133, opened 2021-07-29
WlSeat: notify focus observers when focused surface is destroyed

wmww:
Currently the only `FocusObserver` is the data device. The effect of this PR is that when the focused surface is destroyed `has_focus` will be set to false for that client and future data offers will not be given to it. In practice this behavior causing problems is unlikely, but the new behavior is more correct for data device and useful for text input.

### DIFF
--- a/src/server/frontend_wayland/wl_seat.cpp
+++ b/src/server/frontend_wayland/wl_seat.cpp
@@ -183,9 +183,10 @@ mf::WlSeat::~WlSeat()
     }
 }
 
-auto mf::WlSeat::from(struct wl_resource* seat) -> WlSeat*
+auto mf::WlSeat::from(struct wl_resource* resource) -> WlSeat*
 {
-    return static_cast<mf::WlSeat::Instance*>(wayland::Seat::from(seat))->seat;
+    auto const instance = static_cast<mf::WlSeat::Instance*>(wayland::Seat::from(resource));
+    return instance ? instance->seat : nullptr;
 }
 
 void mf::WlSeat::for_each_listener(wl_client* client, std::function<void(WlPointer*)> func)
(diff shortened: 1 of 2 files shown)


src/server/frontend_wayland/wl_seat.cpp
```diff
@@ -228,8 +233,23 @@ void mf::WlSeat::set_focus_to(WlSurface* new_surface)
...
     }
+    if (focused_surface)
+    {
+        focused_surface.value().remove_destroy_listener(focused_surface_destroy_listener_id);
+    }
     focused_client = new_client;
     focused_surface = mw::make_weak(new_surface);
+    if (new_surface)
+    {
+        focused_surface_destroy_listener_id = new_surface->add_destroy_listener([this]()
```
RAOF, line 244: Is there any guarantee that the `WlSeat` will outlive `new_surface`? ie: what is stopping the `this` pointer from dangling? I guess that would only happen during teardown, but it looks to me like we destroy the `WlSeat` global before we destroy the `wl_display`. I'm not sure if tearing down the `wl_display` will actually run the `wl_surface` destroy listeners, though?

wmww: The listener is removed in the `WlSeat` destructor, and I added a comment to point that out

AlanGriffiths:
> Currently the only `FocusObserver` is the data device. The effect of this PR is that when the focused surface is destroyed `has_focus` will be set to false for that client and future data offers will not be given to it. In practice this behavior causing problems is unlikely, but the new behavior is more correct for data device and useful for text input.

I'm having trouble parsing this: "In practice this [new] behavior causing problems is unlikely, _but_ the new behavior is more correct"?

Can we make it clearer?

"In practice, this change in behaviour causing problems is unlikely, _and_ the new behavior is more correct"?

AlanGriffiths (approved):
LGTM

bors r+

wmww:
"In practice the previous behavior causing problems for data device is unlikely. That being said, the new behavior of data device is *more* correct and the change helps implement text input, therefor it is worth it to merge."

Sorry that made no sense before

PR #2134, opened 2021-08-03
platforms/X11,wayland: Drop duplicate RenderingPlatform code

RAOF:
The `mg::RenderingPlatform` code for both the X11 and wayland is identical to the gbm-kms code.

Since we now load separate rendering and display platforms, gbm-kms can happily provide that code.


AlanGriffiths (changes requested):
> Since we now load separate rendering and display platforms, gbm-kms can happily provide that code.

AFAICS this assumes that the `gbm-kms` platform is always installed when the `x11` or `wayland` platform is installed. That would not be true, for example, when `mir-graphics-drivers-nvidia` is installed instead of `mir-graphics-drivers-desktop`.

Should we move the code to a new shared rendering platform for these backends?

RAOF:
That's a good point. I think I'll defer creating the split-out platform until a bit later, though.

PR #2137, opened 2021-08-04
Frontend: use std::optional instead of std::experimental::optional

wmww:
We've been on C++17 for quite a while. Inconsistent usage of `std::experimental::optional` was annoying me. Better to pull the band aid off all at once than clutter up the diffs of a bunch of future PRs.

### DIFF
--- a/src/wayland/generated/wayland_wrapper.h
+++ b/src/wayland/generated/wayland_wrapper.h
@@ -8,7 +8,7 @@
 #ifndef MIR_FRONTEND_WAYLAND_WAYLAND_XML_WRAPPER
 #define MIR_FRONTEND_WAYLAND_WAYLAND_XML_WRAPPER
 
-#include <experimental/optional>
+#include <optional>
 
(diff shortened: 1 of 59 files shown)


src/wayland/generated/wayland_wrapper.h
```diff
@@ -305,7 +305,7 @@ class DataOffer : public Resource
     static bool is_instance(wl_resource* resource);
 
 private:
-    virtual void accept(uint32_t serial, std::experimental::optional<std::string> const& mime_type) = 0;
+    virtual void accept(uint32_t serial, std::optional<std::string> const& mime_type) = 0;
```
AlanGriffiths, line 308: This is an ABI break. We should also be bumping MIRWAYLAND_ABI etc.

AlanGriffiths (approved):
It really breaks the flow of reviewing to come across unrelated changes (formatting or rewording). Could you keep them separate in future?

bors r+

PR #2138, opened 2021-08-05
platform: Stop pretending the Logger isn't a global singleton

RAOF:


RAOF:
This is somewhat of an RFC, but the platform code (at least) treats the logger as a global and almost completely ignores the `logger` instance we pass to platform construction.

AlanGriffiths (approved):
LGTM

wmww (approved):
LGTM
bors delegate+

PR #2139, opened 2021-08-06
Support zwp_input_method_v2 and zwp_text_input_v3

wmww:
This allows Squeekboard to auto-show, and show different layouts depending on the type of text field. This PR will be split into a number of pieces and merged separately.

To test:
- Run Mir like so: `WAYLAND_DISPLAY=mir ./bin/miral-shell --add-wayland-extension 'zwlr_layer_shell_v1:zwp_input_method_v2:zwp_virtual_keyboard_v1'`
- Make sure the GNOME OSK setting is enabled: `gsettings set org.gnome.desktop.a11y.applications screen-keyboard-enabled true`
- Run Squeekboard in the Mir session: `WAYLAND_DISPLAY=mir squeekboard`
- Select different applications to see the keyboard auto show/hide and show different content (for example it adds an extra row at the top in terminals)

You can test all the different types of input fields by running `tools/entry.py` in the squeekboard repo.

With the GNOME OSK set to false the keyboard can be manually shown with `busctl call --user sm.puri.OSK0 /sm/puri/OSK0 sm.puri.OSK0 SetVisible b true`, but it will not auto-show.


src/server/frontend_wayland/wayland_default_configuration.cpp
```diff
@@ -223,7 +235,8 @@ auto mf::get_standard_extensions() -> std::vector<std::string>
         mw::Shell::interface_name,
         mw::XdgWmBase::interface_name,
         mw::XdgShellV6::interface_name,
-        mw::XdgOutputManagerV1::interface_name};
+        mw::XdgOutputManagerV1::interface_name,
+        mw::TextInputV3::interface_name};
```
wmww, line 239: The text input protocol (without input method) is simply an alternative way for the server to give clients input. Since there are no security concerns, it's added to the set of standard extensions for simplicity.

include/miral/miral/wayland_extensions.h
```diff
@@ -126,6 +126,12 @@ class WaylandExtensions
     /// so malicious clients could use it to take actions without user input.
     /// \remark Since MirAL 3.3
     static char const* const zwp_virtual_keyboard_v1;
+
+    /// Allows clients (such as on-screen keyboards) to act as a source of text input for other clients.
+    /// Clients are not required to display anything to send input using this extension, so malicious
+    /// clients could use it to take actions without user input.
+    /// \remark Since MirAL 3.3
+    static char const* const zwp_input_method_v2;
```
wmww, line 134: Only `zwp_input_method_v2` is added here, since `zwp_text_input_v3` is in the standard extensions.

wmww:
We have a bit of a problem. I pulled in the input method protocol from [Squeekboard](https://gitlab.gnome.org/World/Phosh/squeekboard/-/blob/master/protocols/input-method-unstable-v2.xml), but wlrrots has a different, incompatible [version](https://github.com/swaywm/wlroots/blob/master/protocol/input-method-unstable-v2.xml) of `input-method-unstable-v2`. Someone screwed up somewhere, and we should sort that out before merging.

wmww:
I think what's going on is Squeekboard ended up with a development version of the text input v2 protocol, but since the only breaking change was to do with the grab keyboard (which it doesn't use), no one noticed. I will update this PR to use the correct one, and PR it to Squeekboard as well.

wmww:
Protocol problem has been addressed, ready for review again.

AlanGriffiths:
> Protocol problem has been addressed, ready for review again. Squeekboard PR is [here](https://gitlab.gnome.org/World/Phosh/squeekboard/-/merge_requests/472)

Thanks for catching and dealing with that!

(Sorry for the review delay - I need to switch dev env to test this sensibly...)

AlanGriffiths:
> To test:
> 
>     * Run Mir like so: `WAYLAND_DISPLAY=mir ./bin/miral-shell --add-wayland-extension 'zwlr_layer_shell_v1:zwp_input_method_v2:zwp_virtual_keyboard_v1'`
> 
>     * Make sure the GNOME OSK setting is enabled: `gsettings set org.gnome.desktop.a11y.applications screen-keyboard-enabled true`
> 
>     * Run Squeekboard in the Mir session: `WAYLAND_DISPLAY=mir squeekboard`
> 
>     * Select different applications to see the keyboard auto show/hide and show different content (for example it adds an extra row at the top in terminals

This didn't work first time for me. I'll take another look on Monday.

AlanGriffiths (approved):
> > ```
> > 
> > ```
> 
> 
> This didn't work first time for me. I'll take another look on Monday.

Today it worked! (Somewhat.)

I'm happy to land this as it is effectively behind a feature switch.

But the `ctrl` shift state doesn't seem to be working on the keyboard.

bors r+

PR #2140, opened 2021-08-06
Layer shell: allow destroying surface before the role object

wmww:
As noted in my old bug report [here](https://gitlab.gnome.org/World/Phosh/squeekboard/-/issues/285), Squeekboard violates the protocol when it closes it's layer surface. Currently on Mir, this causes it to be disconnected. With several days to a week of time investment I could probably solve this, which may or may not get merged upstream in a timely manor. In possibly a little less time I could make a patch for wlroots which would force all clients to do the right thing, which again may or may not be merged.

In the short term we need Squeekboard to work. This PR allows the protocol violation in Mir. In `WindowWlSurfaceRole`, `surface` had to be changed to a `wayland::Weak` to make these changes safe.

Part of #2139

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -403,11 +403,12 @@ auto mf::WindowWlSurfaceRole::latest_timestamp() const -> std::chrono::nanosecon
 
 void mf::WindowWlSurfaceRole::commit(WlSurfaceState const& state)
 {
-    if (surface)
+    if (!surface)
     {
-        surface.value().commit(state);
+        return;
     }
 
+    surface.value().commit(state);
     handle_commit();
 
     auto size = pending_size();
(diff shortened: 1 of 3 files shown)


src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -387,7 +403,10 @@ auto mf::WindowWlSurfaceRole::latest_timestamp() const -> std::chrono::nanosecon
 
 void mf::WindowWlSurfaceRole::commit(WlSurfaceState const& state)
 {
-    surface->commit(state);
+    if (surface)
```
RAOF, line 406: You can early-return here, right? There's no point in continuing here if `!surface`, because the surface can never come back?

RAOF (approved):
Seems reasonable.

I wonder whether it might be more sensible to use the Null Object pattern rather than using `Weak<WlSurface>`, though?

AlanGriffiths:
> I wonder whether it might be more sensible to use the Null Object pattern rather than using `Weak<WlSurface>`, though?

I concur: NullObject would make the code simpler.

wmww:
I agree it would simplify some things in this class, but it would require refactoring `WlSurface` to be an interface among other things. This is probably needs to happen in the long run, but I'm not sure it's the time. The surface interface will change significantly if we ever split up the window and surface concepts in Mir, so it seems to me like doing that refactor first makes sense.

This PR is trying to make as small a possible change to make one thing work. It's not pretty, but it's obvious and safe.

AlanGriffiths (approved):
Sure

bors r+

PR #2141, opened 2021-08-06
Add scene::TextInputHub

wmww:
This adds a new `TextInputHub` interface, basic implementation and server context object. This will be used to tie together the text input/input method protocol implementations. It allows an arbitrary number of input methods, but only one handler (client that receives input) at a time.

This doesn't have tests yet. It will hopefully have tests soon. Blocking merging it until tests are ready is reasonable, but not required by me.

Part of #2139.


src/include/server/mir/scene/text_input_hub.h
```diff
@@ -0,0 +1,169 @@
...
+};
+
+/// Gets notifications of new text entered by the user
+class TextInputChangeHandler
+{
+public:
+    TextInputChangeHandler() = default;
+    virtual ~TextInputChangeHandler() = default;
+
+    virtual void change(TextInputChange const& change) = 0;
```
RAOF, line 139: Nit: this is an observer method, right? The name should be passive rather than active - `changed` rather than `change`?

src/server/scene/basic_text_input_hub.h
```diff
@@ -0,0 +1,99 @@
...
+#include "mir/scene/text_input_hub.h"
+#include "mir/observer_multiplexer.h"
+
+#include <mutex>
+
+namespace mir
+{
+namespace scene
+{
+class TextInputStateObserverMultiplexer: mir::Executor, public ObserverMultiplexer<TextInputStateObserver>
```
RAOF, line 31: Heh. We could really do with a `mir::system_executor`, couldn't we ;)

RAOF (commented):
Hm. I think this would be easier to review with either some tests or some usage - at this point it doesn't actually *do* anything, right?

src/server/scene/basic_text_input_hub.h
wmww: Or have `ObserverMultiplexer` implement `Executor`, so we don't have to implement it in each subclass.

But this is not the time to decide on that.

src/include/server/mir/scene/text_input_hub.h
wmww: Done

wmww:
> Hm. I think this would be easier to review with either some tests or some usage

It now has tests!

(also, it's used in #2139)

AlanGriffiths (approved):
Sure

bors r+

PR #2143, opened 2021-08-13
Delete dead code in tests

AlanGriffiths:

PR #2144, opened 2021-08-16
[mir_performance_tests] Use a window management policy that supports windowing

AlanGriffiths:
Fixes "windowed" tests broken by #2143


Saviq (commented):
LGTM

bors r+

PR #2157, opened 2021-08-26
Test the new tests would have detected the problem

AlanGriffiths:


wmww:
`mir_performance_tests` is working on my machine with this PR. Does it have to run on a RasPi or something?

AlanGriffiths:
> `mir_performance_tests` is working on my machine with this PR. Does it have to run on a RasPi or something?

I see it intermittently on my system (but it doesn't take many iterations to see a failure). I was hoping to see it fail in CI but...

AlanGriffiths:
This is getting too fragile. Going to go a different way.

PR #2159, opened 2021-08-27
test/graphics-dummy: Guard against unwanted symbol export.

RAOF:
GoogleTest *very conveniently* uses at least one global with a non-trivial
destructor. Due to the vagaries of static linking, these globals are
exported in the symbol table unless you explicitly hide them.

This interacts badly with `dlopen()`. The object will get constructed during
the initial loading process, along with the other globals, and then will
be constructed *again* when the ELF-module-constructor machinery runs
to load the `dlopen`ed DSO.

Even better, during process teardown the DSO's ELF-destructor will run
the destructor for the global. *Then* the executable's ELF-destructor will
run the destructor for the global *again*. This is rarely what you want,
and in *this* case it results in GoogleTest hitting an `abort()` when it
tries to destroy a `pthread_key_t` a second time.

Hiding every symbol we don't explicitly want to export solves this. The
main executable has a GoogleTest global, and graphics-dummy.so has a
GoogleTest global, but they're now at different addresses and so we
don't double destruct a single object.

### DIFF
--- a/tests/mir_test_framework/open_wrapper.cpp
+++ b/tests/mir_test_framework/open_wrapper.cpp
@@ -30,6 +30,7 @@
 #include <dlfcn.h>
 #include <cstdarg>
 #include <fcntl.h>
+#include <boost/throw_exception.hpp>
 
 namespace mtf = mir_test_framework;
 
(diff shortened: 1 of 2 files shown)


RAOF:
Please feel free to copyedit the description!

(Should now pass on Alpine)

AlanGriffiths (approved):
bors merge

PR #2162, opened 2021-09-07
Consolidate ProofOfMutexLock

wmww:
We now use `ProofOfMutexLock` in two places, and it stands to reason it will be useful in more. This moves it to it's own header `core/mir/proof_of_mutex_lock.h`.


AlanGriffiths (approved):
bors r+

PR #2163, opened 2021-09-07
Use SurfaceStateTracker to track X11 window state

### LINKED ISSUE #1217: XWayland WM conflates virt_maximised/horiz_maximized/maximised
I don't think this assumption holds? There's plenty of stuff that will vertically maximise but not horizontally maximise.

_Originally posted by @RAOF in https://github.com/MirServer/mir/pull/1211_

wmww:
Not too long ago we introduced `SurfaceStateTracker` to store multiple surface states at once. Previously the XWayland frontend was using it's own class to do this. With this PR, it uses the common one. Fixes #1217.

### DIFF
--- a/src/include/server/mir/scene/surface_state_tracker.h
+++ b/src/include/server/mir/scene/surface_state_tracker.h
@@ -64,9 +64,6 @@ public:
     [[nodiscard]]
     auto without(MirWindowState state) const -> SurfaceStateTracker;
 
-    /// Returns if all states are equal
-    auto operator==(SurfaceStateTracker const& other) const -> bool;
-
 private:
     SurfaceStateTracker()
         : hidden{false},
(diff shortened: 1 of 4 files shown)


src/include/server/mir/scene/surface_state_tracker.h
```diff
@@ -64,6 +64,9 @@ struct SurfaceStateTracker
     [[nodiscard]]
     auto without(MirWindowState state) const -> SurfaceStateTracker;
 
+    /// Returns if all states are equal
+    auto operator==(SurfaceStateTracker const& other) const -> bool;
```
AlanGriffiths, line 68: There are no conversions available for `SurfaceStateTracker` so this is moot, but it is good practice for `operator==` to be a free function. It is always good practice to define both `operator==` and `operator!=`.

AlanGriffiths (changes requested):
Looks sensible (how I wish for tests!).

Just some nitpicking about member `operator==` it should be:
```c++
friend auto operator==(SurfaceStateTracker const& lhs, SurfaceStateTracker const& rhs) ...

auto operator==(SurfaceStateTracker const& lhs, SurfaceStateTracker const& rhs)...
auto operator!=(SurfaceStateTracker const& lhs, SurfaceStateTracker const& rhs)...
```

src/include/server/mir/scene/surface_state_tracker.h
wmww: Done

AlanGriffiths (approved):
bors r+

Issue #2164, opened 2021-09-08
No way for frontend to change non-active window state


wmww:
TLDR: in some cases the frontend can't do quite what it wants, but its not a big deal and we can just ignore it for now.

For example if a window is fullscreen is may or may not also be maximized. This does not have any immediate effect, but will matter when the window is unfullscreened. The frontend can view this state via `surface->state_tracker()`, but there is no mechanism for it to change it.

Possible workarounds (not currently used or advisable, but included for completeness):
- Cache desired state in the frontend. This would add complexity, and only work if a single component was operating on the state (in the case of foreign toplevel manager + XWayland, this might not be the case)
- unfullscreen, unmaximize and refullscreen. Would cause flicker, it's generally better to just ignore the change.

Cause: `SurfaceStateTracker` was designed around the exisiting `SurfaceSpecificiation`/`SurfaceCreationParameters` APIs, miral, and the existing window manager logic. It solved many problems with relatively little code churn. This is the problem it can't solve.

Solution: any solution would probably add more complexity than it's worth. Right now the only visible effect is a WLCS test failing (`ForeignToplevelHandleTest.can_maximize_foreign_while_fullscreen`), so I say we ignore it.

AlanGriffiths:
My recollection is that we first encountered this before we deleted mirclient and the tests associated with the expectations there. We no longer have the cost associated with updating that code, so maybe we should revisit that choice.

wmww:
In my opinion we should wait until it causes noticeably non-ideal behavior in the wild. It might be easier than before, but it will still be a bunch of work and add complexity to get 100% right.

PR #2165, opened 2021-09-08
Use state tracker for foreign toplevel control

wmww:
Makes our foreign toplevel implementation simpler and more robust by not keeping track of surface state in the frontend. This does cause one WLCS test to fail, which is explained in #2164. I do not consider changing non-active state to be particularly important, and if it is doing it by keeping a cache in the frontend is not the way to go about it.


AlanGriffiths (approved):
LGTM

bors r+

PR #2168, opened 2021-09-16
Miral client_facing_state

wmww:
This introduces a 2nd way for shells to set window states. They may now set a `client_facing_state` that is different from the `state` used for window management. If the client facing state is not set, `state` is used by default. 

In the short term this helps https://github.com/MirServer/ubuntu-frame/pull/15, because it allows us to make clients look fullscreen but behave as if they're maximized (critically, this causes them to get out of the way of the OSK). In the long term it may also be useful for tiling window managers, and other shells that want fine-graned control over how clients behave.

Marked as draft until I've written tests.


wmww:
Now with tests

AlanGriffiths:
> make clients look fullscreen but behave as if they're maximized (critically, this causes them to get out of the way of the OSK).

I may have missed something here. What prevents "clients getting out of the way of the OSK"? Surely any resizing/repositioning should be happening in Mir, not the client?

I can imagine that Mir is not doing things right, but the shell should get to decide what "fullscreen" means. Historically, Unity8 had "fullscreen" with and without "chrome" (specifically a notification bar that was only hidden for a few apps).

AlanGriffiths:
After discussion we decided this is not the right approach

PR #2169, opened 2021-09-17
MirAL: move depth layer and focus mode setup to place_new_surface()

wmww:
I noticed this code looked out of place in `BasicWindowManager::add_surface()`. Where it was before it couldn't be overridden by the shell. No pressing reason for this change, but ability to customize these things could be useful later.


AlanGriffiths:
bors try

AlanGriffiths (approved):
bors merge

PR #2170, opened 2021-09-21
WindowWlSurfaceRole use surface spec instead of creation params

wmww:
Simplifies `WindowWlSurfaceRole` by transitioning it to only creating surface creation parameters when needed. Part of the project to transition away from surface creation parameters entirely.


AlanGriffiths (approved):
Looks and behaves right

bors r+

PR #2174, opened 2021-09-23
Drop SurfaceCreationParameters and use SurfaceSpecification in it's place

wmww:
The day is here when we can have one less surface parameters struct!


RAOF (approved):
LGTM. Could probably be made better by just removing (AFAICT never actually used) `input_mode` field entirely?

src/include/server/mir/shell/surface_specification.h
```diff
@@ -95,12 +99,8 @@ struct SurfaceSpecification
     optional_value<std::weak_ptr<scene::Surface>> parent;
 
     optional_value<std::vector<geometry::Rectangle>> input_shape;
+    optional_value<input::InputReceptionMode> input_mode;
```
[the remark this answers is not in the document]
AlanGriffiths: Answer: It is read in `ApplicationSession::create_surface()` and `miral::WindowSpecification::Self::Self()`;  it is set in `AbstractShell.create_surface_allows_window_manager_to_set_create_parameters` etc.

I think it _may_ be possible to remove `input_mode`, but doing so isn't a pre-requisite for landing this rework.

debian/libmiral4.symbols
```diff
@@ -259,7 +259,6 @@ libmiral.so.4 libmiral4 #MINVER#
  (c++)"miral::WindowManagerTools::window_at(mir::geometry::Point) const@MIRAL_3.0" 3.0.0
  (c++)"miral::WindowManagerTools::~WindowManagerTools()@MIRAL_3.0" 3.0.0
  (c++)"miral::WindowSpecification::WindowSpecification()@MIRAL_3.0" 3.0.0
- (c++)"miral::WindowSpecification::WindowSpecification(mir::scene::SurfaceCreationParameters const&)@MIRAL_3.0" 3.0.0
```
AlanGriffiths, line 262: This (and `miral::WindowSpecification::update(mir::scene::SurfaceCreationParameters&)`) is a ABI break.

We can probably get away with this as there is no sane way a downstream could be using these functions. (This would be clearer if they had been made private, or non-member `friend` functions originally.)

src/server/shell/surface_specification.cpp
```diff
@@ -147,6 +158,66 @@ void msh::SurfaceSpecification::update_from(SurfaceSpecification const& that)
...
+    const msh::SurfaceSpecification& rhs)
+{
+    return
+        lhs.name == rhs.name &&
+        lhs.width == rhs.width &&
+        lhs.height == rhs.height &&
+        lhs.top_left == rhs.top_left &&
+        lhs.buffer_usage == rhs.buffer_usage &&
+        lhs.pixel_format == rhs.pixel_format &&
+        lhs.input_mode == rhs.input_mode &&
```
AlanGriffiths, line 178: This is also mentioned on line 204. (It is really annoying there's no easy way to write `return std::tie($all_data_members) == std::tie(rhs.$all_data_members);`

AlanGriffiths (approved):
There one "nit" and @RAOF's suggestion.

bors delegate=@wmww

src/include/server/mir/shell/surface_specification.h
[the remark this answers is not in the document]
wmww: It's in the `miral::WindowSpecification` API, so removing it would mean either breaking ABI or making that feature not work.

debian/libmiral4.symbols
wmww: - Restored the dropped functions and replaced their implementation with a deprecation warning.
- Created internal `WindowSpecification` header and source files
- Replaced the new `apply_to` function with a standalone internal `make_surface_spec()` function.

src/server/shell/surface_specification.cpp
wmww: Fixed

include/miral/miral/window_specification.h
```diff
@@ -56,7 +56,9 @@ class WindowSpecification
     auto operator=(WindowSpecification const& that) -> WindowSpecification&;
 
     WindowSpecification(mir::shell::SurfaceSpecification const& spec);
+    /// \deprecated and rendered inert in MirAL 3.3
     WindowSpecification(mir::scene::SurfaceCreationParameters const& params);
+    /// \deprecated and rendered inert in MirAL 3.3
     void update(mir::scene::SurfaceCreationParameters& params) const;
```
AlanGriffiths, line 62: These cannot be called by client code as it has no access to `mir::scene::SurfaceCreationParameters`, so while dropping them is technically an ABI break I don't think we should go to any effort to "deprecated and rendered inert".

src/miral/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ add_library(miral-internal STATIC
     xcursor.c                           xcursor.h
                                         join_client_threads.h
                                         window_info_defaults.h
+    window_specification_internal.cpp   window_specification_internal.h
```
AlanGriffiths, line 30: Why are we splitting an implementation detail of `miral::WindowSpecification` into a separate file? That will just increase build times (and, arguably, make the code harder to follow as it isn't clear that nothing else uses it).

AlanGriffiths (changes requested):
I preferred the version I reviewed last time: there was just one piece of mostly harmless code duplication.

src/miral/CMakeLists.txt
wmww: I want `make_surface_spec(WindowSpecification const&)` to be a standalone function which isn't in the miral ABI. I get link errors if I put it's implementation in window_specification.cpp, or move window_specification.cpp from the `miral SHARED` library to the `miral-internal STATIC` library in cmake.

AlanGriffiths: Repeating back to ensure I understand what you want:

You want `make_surface_spec()` to:

1. have `visibility=hidden` so it isn't exported by libmiral,
2. have access to the implementation of WindowSpecification,
3. not be a member function of WindowSpecification

Isn't it clearer to put it in `window_specification.cpp` and mark it `__attribute__ ((visibility ("hidden")))`?

AlanGriffiths: Ah. Further discussion adds another key requirement:

4. available when linking tests against `libmiral-internal`.

That is probably best done by moving all this into `libmiral-internal`.

AlanGriffiths (approved):
bors merge

PR #2175, opened 2021-09-24
Fake discrete scrolling

AlanGriffiths:
We don't currently have discrete scrolling information for our (internal) pointer events.

In principle, this should be captured by the input platforms (libinput, X11, Wayland etc.) and sent on to the client.

This is a hack to provide these events anyway as some clients rely on them. C.f.

https://forum.snapcraft.io/t/the-quest-for-building-a-plain-wayland-electron-kiosk/26739/3


wmww:
I haven't tested yet, but this seems likely to me to cause problems. Even if we were faking them, we would want to at least accumulate scroll delta and send one discrete event per amount of delta. However, I think it's probably easier to just do it properly.

Issue #2177, opened 2021-09-27
[linux_dmabuf_unstable_v1] does not allow requested format modifier DRM_FORMAT_MOD_INVALID


AlanGriffiths:
The specification of zwp_linux_buffer_params_v1.add says:

        Warning: It should be an error if the format/modifier pair was not
        advertised with the modifier event. This is not enforced yet because
        some implementations always accept `DRM_FORMAT_MOD_INVALID`. Also
        version 2 of this protocol does not have the modifier event.

Mir, unlike “some implementations” (one of which is Mutter), is treating this as an error.

This breaks electron 15 clients as they request `DRM_FORMAT_MOD_INVALID`. (see https://forum.snapcraft.io/t/the-quest-for-building-a-plain-wayland-electron-kiosk/26739/17)

PR #2178, opened 2021-09-27
Allow zwp_linux_buffer_params_v1.add() with DRM_FORMAT_MOD_INVALID

AlanGriffiths:
Some clients make this technically invalid request.

Fixes: #2177

### DIFF
--- a/src/platform/graphics/linux_dmabuf.cpp
+++ b/src/platform/graphics/linux_dmabuf.cpp
@@ -738,6 +738,21 @@ private:
                 }
             }
         }
+
+        // The specification of zwp_linux_buffer_params_v1.add says:
+        //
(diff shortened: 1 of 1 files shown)


wmww (approved):
LGTM

@RAOF, any thoughts?

bors delegate+

RAOF (approved):
Yup. We'll handle this correctly (or, at least, as correctly as is possible) elsewhere in the code. We can't tell whether this will actually *work*, of course, but if the client gets totally wrong rendering hopefully the user will blame the client :woman_shrugging: .

This *probably* can't be used maliciously? Possibly it could be used to exploit a driver bug, but I think we handle things as correctly as we can.

bors r=wmww,RAOF

PR #2179, opened 2021-09-27
Send axis_source events

wmww:
Alternative to #2175. `.axis_source` is optional according to the protocol, but it appears electron ignores non-discrete scroll events without it set to `finger`. This PR is technically incorrect for non-touchpads, but it's unlikely it will break anything and it fixes scrolling in electron.


RAOF:
I mildly prefer this to #2175?

AlanGriffiths:
> I mildly prefer this to #2175?

I think that we should do as @wmww suggested there: "just do it properly."

PR #2182, opened 2021-09-29
Add discrete scrolling events

AlanGriffiths:


src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -232,9 +232,22 @@ void mf::WlPointer::axis(MirPointerEvent const* event)
         break;
 
     case mir_pointer_axis_source_wheel:
+    {
         send_axis_source_event(AxisSource::wheel);
+
+        if (auto const h_scroll_discrete = mir_pointer_event_axis_value(event, mir_pointer_axis_hscroll_discrete))
+        {
+            send_axis_discrete_event(Axis::horizontal_scroll, h_scroll_discrete);
+        }
```
wmww, line 241: 1. Is it wise to only send discrete events for `axis_source_wheel`? I believe our platform will only add discrete scroll to that source, but the protocol doesn't forbid it for other sources. I see no reason to prevent other sources from sending discrete events here.
2. The protocol is very specific that:
    > This event does not occur on its own, it is coupled with wl_pointer.axis event
  
    in general we will send both, but if for some reason the platform gives us discrete scroll with 0 normal scroll, we will not and thus make a protocol violation.

wmww (changes requested):
Sorry it took me so long to get to, this must have slipped through the cracks before.

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -232,9 +232,22 @@ void mf::WlPointer::axis(MirPointerEvent const* event)
...
+
+        if (auto const v_scroll_discrete = mir_pointer_event_axis_value(event, mir_pointer_axis_vscroll_discrete))
+        {
+            send_axis_discrete_event(Axis::horizontal_scroll, v_scroll_discrete);
```
wmww, line 245: I stared at this three times before I noticed, can you spot the problem?

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -113,6 +113,30 @@ auto timestamp_of(MirPointerEvent const* event) -> uint32_t
...
+
+auto wayland_axis_source(MirPointerAxisSource mir_source) -> std::optional<uint32_t>
+{
+    switch (mir_source)
+    {
+    case mir_pointer_axis_source_none:
+        return std::nullopt;
+
+    case mir_pointer_axis_source_wheel:
+        return static_cast<uint32_t>(mw::Pointer::AxisSource::wheel);
```
wmww, line 125: I have no idea why these `static_cast`s are needed, but both GCC and Clang give
```
undefined symbol: mir::wayland::Pointer::AxisSource::wheel
```
without them.

AlanGriffiths: > I have no idea why these `static_cast`s are needed, but both GCC and Clang give

It is because there's no definition of them. Here's the declaration:
```c++
    struct AxisSource
    {
        static uint32_t const wheel = 0;
        static uint32_t const finger = 1;
        static uint32_t const continuous = 2;
        static uint32_t const wheel_tilt = 3;
    };
```

Somewhere there should be a definition:
```C++
uint32_t const mir::wayland::Pointer::AxisSource::wheel;
```
This appears to be a generator problem.

PR #2185, opened 2021-10-01
Add exclusive_mode property to windows

wmww:
Adds a property that allows shells to control if attached surfaces should push fullscreen surfaces out of the way. With this PR, application zones are refreshed ever time a surfaces becomes fullscreened or unfullscreened. It doesn't have tests yet, or pass existing tests (I'm working on it). Alternative to #2183.

### DIFF
--- a/include/core/mir_toolkit/common.h
+++ b/include/core/mir_toolkit/common.h
@@ -428,4 +428,16 @@ typedef enum MirFocusMode
                                       leaving it as long as it has this mode */
 } MirFocusMode;
 
+/**
+ * Controls this surface's exclusive rect is applied. Only revenant to attached surfaces that have an exclusive rect.
+ */
+typedef enum MirExclusiveMode
+{
+    mir_exclusive_mode_enabled,             /**< If this surface is attached it's exclusive rect will affect the
+                                                 application zone */
+    mir_exclusive_mode_disabled,            /**< This surface's exclusive rect will not affect the application zone */
+    mir_exclusive_mode_affect_fullscreen,   /**< Same as mir_exclusive_mode_enabled, except this surface's exclusive
+                                                 rect will also effect fullscreen surfaces */
+} MirExclusiveMode;
+
 #endif
--- a/include/miral/miral/window_info.h
+++ b/include/miral/miral/window_info.h
@@ -133,6 +133,10 @@ struct WindowInfo
     /// \remark Since MirAL 3.3
     auto focus_mode() const -> MirFocusMode;
 
+    /// If the window's exclusive rect should affect the application zone
+    /// \remark Since MirAL 3.3
+    auto exclusive_mode() const -> MirExclusiveMode;
+
 private:
     friend class BasicWindowManager;
     void name(std::string const& name);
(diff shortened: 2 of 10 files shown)


wmww:
Ah! problem was a previously undetected issue with the tests. Fixed.

AlanGriffiths (changes requested):
Tagging individual attached surfaces does make sense. But the implementation needs a lot more work:

1. This will change the application zones on _all outputs_ if there's a fullscreen window on any of them.
2. This will change the application zones for maximized windows if there's a fullscreen window hidden behind them, a transparent one over them or on another output.
3. When adding to the Miral API you have to get in the habit of running a "Release" build with g++: that will regenerate the .symbols file for you.
4. As you are already aware this needs tests.

include/core/mir_toolkit/common.h
```diff
@@ -428,4 +428,16 @@ typedef enum MirFocusMode
                                       leaving it as long as it has this mode */
 } MirFocusMode;
 
+/**
+ * Controls this surface's exclusive rect is applied. Only revenant to attached surfaces that have an exclusive rect.
```
AlanGriffiths, line 432: I don't think you mean "revenant"· ;)

include/core/mir_toolkit/common.h
```diff
@@ -428,4 +428,16 @@ typedef enum MirFocusMode
...
+ */
+typedef enum MirExclusiveMode
+{
+    mir_exclusive_mode_enabled,             /**< If this surface is attached it's exclusive rect will affect the
```
Saviq, line 436:
```suggestion
    mir_exclusive_mode_enabled,             /**< If this surface is attached its exclusive rect will affect the
```

include/core/mir_toolkit/common.h
```diff
@@ -433,11 +433,11 @@ typedef enum MirFocusMode
...
 {
-    mir_exclusive_mode_enabled,             /**< If this surface is attached its exclusive rect will affect the
-                                                 application zone */
-    mir_exclusive_mode_disabled,            /**< This surface's exclusive rect will not affect the application zone */
-    mir_exclusive_mode_affect_fullscreen,   /**< Same as mir_exclusive_mode_enabled, except this surface's exclusive
-                                                 rect will also effect fullscreen surfaces */
+    mir_exclusive_mode_none,    /**< This surface's exclusive rect will not affect application zones */
+    mir_exclusive_mode_default, /**< If this surface is attached its exclusive rect will affect application
+                                     zones for non-fullscreen windows */
+    mir_exclusive_mode_all,     /**< Same as mir_exclusive_mode_default, except this surface's exclusive rect
```
wmww, line 439: These names seem significantly less obvious than the previous ones. They are shorter, but give no indication what the modes actually do.

wmww:
> When adding to the Miral API you have to get in the habit of running a "Release" build with g++

Ohhh, I've been wondering why my symbols never work.

AlanGriffiths:
> > 
> 
> This will be difficult to fix with the current architecture (and MirAL API, since application zones are shared to MirAL). I'm thinking perhaps #2183 is the better solution for now.

I don't think it is difficult to fix, just needs a different implementation approach: maintaining a set of "fullscreen zones" that track the application zone logic with a few tweaks.

But, as you've shown in https://github.com/MirServer/ubuntu-frame/pull/15, there is no urgency for either solution.

wmww:
> I don't think it is difficult to fix, just needs a different implementation approach: maintaining a set of "fullscreen zones" that track the application zone logic with a few tweaks.

There's some issues:
1. Is MirAL informed of these fullscreen zones? If not customizing the logic can't work, if so that's new complexity for the API
2. What if an attached surface that does affect fullscreen surfaces is pushed around by an attached surface that doesn't? When a window is firescreened the former needs to be moved as the later disappears. This makes things messy.

PR #2187, opened 2021-10-04
Sequence AsyncServerRunner::start_server() setup on one thread

AlanGriffiths:
Release builds on 21.04 segfault in mir_performance_tests

That happens because the callback passed to `add_init_callback()` executes with an invalid context.

I don't see how this could happen as the `std::functor` that contains it is built before the thread that executes it is started. (So that should be  sequenced-before.)

Anyway, moving the call to `add_init_callback()` onto the "AsyncServer" seems to avoid this and is otherwise harmless.

### DIFF
--- a/include/test/mir_test_framework/async_server_runner.h
+++ b/include/test/mir_test_framework/async_server_runner.h
@@ -51,13 +51,6 @@ public:
 
     mir::Server server;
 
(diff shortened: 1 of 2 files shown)


wmww (approved):
LGTM
bors r+

PR #2199, opened 2021-10-25
Error on invalid xdg_surface size

wmww:
I ran into this trying neverball on Arch.

I use `mw::ProtocolError` even though there is not valid error code, because that makes the client show the error. If I threw a `std::runtime_error` it only shows up in the Mir log, which a client dev may not check (I haven't checked if this is true for all clients or just SDL). In any case, I think this is more correct, and it's what wlroots does.


RAOF:
My understanding is that we've fixed the Sid build. Let's give this another try!

bors r+

PR #2200, opened 2021-10-26
Add display-scale option

wmww:
This may be useful in all sorts of contexts. It only allows specifying one scale for all outputs, but in many cases that's all you need.

### DIFF
--- a/src/miral/display_configuration_option.cpp
+++ b/src/miral/display_configuration_option.cpp
@@ -44,6 +44,10 @@ char const* const display_alpha_descr = "Select a display mode with alpha[{on,of
 char const* const display_alpha_off = "off";
 char const* const display_alpha_on = "on";
 
+char const* const display_scale_opt = "display-scale";
+char const* const display_scale_descr = "Scale for all displays";
+char const* const display_scale_default = "1.0";
+
 class PixelFormatSelector : public mg::DisplayConfigurationPolicy
 {
 public:
@@ -97,6 +101,7 @@ void miral::display_configuration_options(mir::Server& server)
     // Add choice of monitor configuration
     server.add_configuration_option(display_config_opt, display_config_descr,   sidebyside_opt_val);
     server.add_configuration_option(display_alpha_opt,  display_alpha_descr,    display_alpha_off);
+    server.add_configuration_option(display_scale_opt,  display_scale_descr,    display_scale_default);
 
     server.wrap_display_configuration_policy(
         [&](std::shared_ptr<mg::DisplayConfigurationPolicy> const& wrapped)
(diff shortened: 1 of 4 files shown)


src/miral/display_configuration_option.cpp
```diff
@@ -105,15 +110,39 @@ void miral::display_configuration_options(mir::Server& server)
...
+            try
+            {
+                scale = std::stod(scale_str);
+            }
+            catch (std::invalid_argument const&)
+            {
+            }
+            if (scale < 0.01 || scale > 100.0)
+            {
+                mir::fatal_error("Invalid scale %s", scale_str.c_str());
```
RAOF, line 125: It's easy enough here to distinguish between “user passed something that's not a double” and “user passed a value we reject”, and I think it would be useful to tell people that we're going to (perfectly sensibly!) reject scales outside the range 0.01 - 100.0.

RAOF (commented):
I think we should change the error message to say something along the lines of “only scales between 0.01 and 100.0 are accepted” (and maybe also have a different message for when we just can't parse the value).

Once that's done,
bors delegate=wmww

src/miral/display_configuration_option.cpp
wmww: Done, I'll leave the borsing to you

src/miral/display_configuration_option.cpp
```diff
@@ -105,15 +110,42 @@ void miral::display_configuration_options(mir::Server& server)
             auto const options = server.get_options();
             auto display_layout = options->get<std::string>(display_config_opt);
             auto with_alpha = options->get<std::string>(display_alpha_opt) == display_alpha_on;
+            auto const scale_str = options->get<std::string>(display_scale_opt);
+
+            double scale{0};
+            static double const scale_min = 0.01;
```
RAOF, line 116: I think we can trust the compiler to optimise these out without explicitly marking them `static`, but I also don't think it'll hurt to do so…

RAOF (approved):
bors r+

AlanGriffiths:
bors retry

Saviq:
Not irrelevant, is it?

```plain
124/230 Test #129: mir_unit_tests.SideBySideDisplayConfigurationPolicyTest.* ......................***Failed    0.64 sec
Running main() from main.cpp
Note: Google Test filter = SideBySideDisplayConfigurationPolicyTest.*:-
[==========] Running 4 tests from 1 test suite.
[----------] 4 tests from SideBySideDisplayConfigurationPolicyTest
[ RUN ] SideBySideDisplayConfigurationPolicyTest.uses_all_connected_valid_outputs
[ OK ] SideBySideDisplayConfigurationPolicyTest.uses_all_connected_valid_outputs (0 ms)
[ RUN ] SideBySideDisplayConfigurationPolicyTest.placement_respects_scale
/spread/mir/tests/unit-tests/graphics/test_default_display_configuration_policy.cpp:349: Failure
Expected equality of these values:
expected_position
Which is: 62, 0
output.top_left
Which is: 123, 0
[ FAILED ] SideBySideDisplayConfigurationPolicyTest.placement_respects_scale (1 ms)
[ RUN ] SideBySideDisplayConfigurationPolicyTest.default_policy_is_power_mode_on
[ OK ] SideBySideDisplayConfigurationPolicyTest.default_policy_is_power_mode_on (0 ms)
[ RUN ] SideBySideDisplayConfigurationPolicyTest.default_orientation_is_normal
[ OK ] SideBySideDisplayConfigurationPolicyTest.default_orientation_is_normal (0 ms)
[----------] 4 tests from SideBySideDisplayConfigurationPolicyTest (1 ms total)

[==========] 4 tests from 1 test suite ran. (1 ms total)
[ PASSED ] 3 tests.
[ FAILED ] 1 test, listed below:
[ FAILED ] SideBySideDisplayConfigurationPolicyTest.placement_respects_scale
```

AlanGriffiths (approved):
bors r+

PR #2204, opened 2021-11-02
Add WaylandExtensions::conditionally_enable()

### LINKED ISSUE #2194: Shells are unable to respect wayland extensions-related options when using a filter
Normally, a user can control the enabled Wayland extensions with the `--add-wayland-extensions` option (and related options). Unfortunately, shell's can't query this option. This makes the following behavior unimplementable:
- Shell wants to disable a given extension by default
- But enable it for some clients
- Or enable it for all clients if the user has specified it in `--add-wayland-extensions`

wmww:
Fixes #2194. It's my first stab at it, so there might be a cleaner way.

This introduces a new per-extension conditional callback, which can be used to replace `set_filter()`. That function takes a struct instead of a list of arguments for future extendability.

No tests yet, I'll write them once we decide if we're going to use this approach.


include/miral/miral/wayland_extensions.h
```diff
@@ -92,9 +92,24 @@ class WaylandExtensions
...
+    struct Info
+    {
+        /// The application that is being given access to this extension
+        Application const& app;
+        /// If this extension is in the recommended set
+        bool recommended;
+        /// If the user has enabled this extension using the wayland-extensions or related option
+        bool manually_enabled;
+    };
+
```
AlanGriffiths, line 106: Structs with data members can become problematic for maintaining ABI. When there are changes the code creating or copying the `struct` need not have been built with the same definition as that using it.

include/miral/miral/wayland_extensions.h
```diff
@@ -92,9 +92,24 @@ class WaylandExtensions
...
     /// \remark Since MirAL 2.5
     using Filter = std::function<bool(Application const& app, char const* protocol)>;
 
+    /// \remark Since MirAL 3.4
+    using Condition = std::function<bool(Info const& info)>;
```
AlanGriffiths, line 111: The names `Condition` and `Info` are rather generic - nothing to suggest they are specific to protocol enablement.

AlanGriffiths:
It is probably worth tabulating the possibilities and desired outcome:

Config|Shell|`Filter`|`Condition`|Result|Comment
--|--|--|--|--|--
(none)|`enable()`|(none)|(none)| **enabled**
(none)|`enable()`|`true`|(none)| **enabled**
(none)|`enable()`|`false`|(none)| disabled
add-wayland-extensions|(none)|(none)|(none)| **enabled**
add-wayland-extensions|(none)|`true`|(none)| **enabled**
add-wayland-extensions|(none)|`false`|(none)| disabled
drop-wayland-extensions|(none)|N/A| ?? | disabled
drop-wayland-extensions|`enable()`|N/A|(none)| disabled | What happens now, but may not be right

Etc

wmww:
Note that we should probably address #2208 before we land something like this.

AlanGriffiths:
> Note that we should probably address #2208 before we land something like this.

Please elaborate: I don't see there being a connection.

wmww:
This function enables based on extension name but filters based on global name. Until those two are the same thing it will not work right.

include/miral/miral/wayland_extensions.h
wmww: Done

include/miral/miral/wayland_extensions.h
wmww: Updated

wmww:
The new logic is (as is documented and tested):
- How `.enable()`, `.disable()` command line options and `.set_filter()` interact is unchanged:
  - command line options override `.enable()` and `.disable()`
  - `.set_filter()` filters everything
- `.conditionally_enable()` throws a `std::logic_error` if used with filter, so there are no questions about how they interact
-  `.conditionally_enable()` can override command line options, but it can also be used to respect them (it's given that info)
- A `.conditionally_enable()`ed extension is neither "enabled" or "disabled", it's in a 3rd state (and calling `.enable()` or `.disable()` on it changes it to the corresponding state)

include/miral/miral/wayland_extensions.h
```diff
@@ -163,16 +188,24 @@ class WaylandExtensions
...
-    /// Throws a std::runtime_error if the extension is not supported
+    /// Disable a Wayland extension by default. The user can still enable it with the add-wayland-extensions or
+    /// wayland-extensions options. The extension can be forced to be disabled regardless of user options with
+    /// conditionally_enable().
     /// \remark Since MirAL 2.6
     auto disable(std::string name) -> WaylandExtensions&;
 
+    /// Enable a Wayland extension only when the callback returns true. The callback can use info.user_preference()
+    /// to respect the extension options the user provided, it is not required. Unlike enable() and disable(),
+    /// conditionally_enable() can override the user options.
```
AlanGriffiths, line 205: [Not blocking, but a note]

What we discussed on Mattermost was:

> Following your suggestion above, for conditionally_enable(), we should probably not callback if the config is used to enable a protocol.

This proposal may well be a better approach, but we should choose explicitly

src/miral/wayland_extensions.cpp
```diff
@@ -339,15 +394,58 @@ void miral::WaylandExtensions::operator()(mir::Server& server) const
...
+                            app,
+                            protocol,
+                            user_pref});
+                    }
+                    else
+                    {
+                        return self->extensions_filter ? self->extensions_filter.value()(app, protocol) : true;
+                    }
+                });
         });
```
AlanGriffiths, line 449: This lambda seems overcomplicated (as does the surrounding one). We can do a lot of the logic before calling `server.set_wayland_extension_filter()`:
```c++
            if (self->extensions_filter)
            {
                server.set_wayland_extension_filter(
                    [self](Application const& app, char const* protocol) -> bool
                    {
                        return self->extensions_filter.value()(app, protocol);
                    });
            }
            else if (!self->conditional_extensions.empty())
            {
                server.set_wayland_extension_filter(
                    [&self = self, manually_enabled_extensions, manually_disabled_extensions]
                    (Application const& app, char const* protocol) -> bool
                    {
                        auto const cond = self->conditional_extensions.find(protocol);

                        if (cond == self->conditional_extensions.end())
                        {
                            return true;
                        }
                        else
                        {
                            std::optional<bool> user_pref;
                            if (manually_enabled_extensions.find(protocol) != manually_enabled_extensions.end())
                            {
                                user_pref = true;
                            }
                            else if (manually_disabled_extensions.find(protocol) != manually_disabled_extensions.end())
                            {
                                user_pref = false;
                            }

                            return cond->second(EnableInfo{ app, protocol, user_pref});
                        }
                    });
            }
```

(Need to restore L369-373 too for that to work)

AlanGriffiths (approved):
OK, there's some follow up, but let's get in on master

bors r+

PR #2210, opened 2021-11-04
Change extension names to always match global names

### LINKED ISSUE #2208: Extension names and global names do not match 1:1
Some extensions have globals with a different name than the protocol (often with `_manager` appended). This is a problem if you try to compare the name given to filter with the name used to enable the extension.

wmww:
Fixes #2208. Extension names are now automatically deduced for built-in extensions, and if a bespoke extension with an invalid name is enabled, a fatal error is raised at startup.

### DIFF
--- a/src/miral/wayland_extensions.cpp
+++ b/src/miral/wayland_extensions.cpp
@@ -37,8 +37,12 @@ namespace mo = mir::options;
 char const* const miral::WaylandExtensions::zwlr_layer_shell_v1{"zwlr_layer_shell_v1"};
 char const* const miral::WaylandExtensions::zxdg_output_manager_v1{"zxdg_output_manager_v1"};
 char const* const miral::WaylandExtensions::zwlr_foreign_toplevel_manager_v1{"zwlr_foreign_toplevel_manager_v1"};
-char const* const miral::WaylandExtensions::zwp_virtual_keyboard_v1{"zwp_virtual_keyboard_v1"};
-char const* const miral::WaylandExtensions::zwp_input_method_v2{"zwp_input_method_v2"};
+/// Not in the header, but keeping around for ABI compat
+char const* const miral::WaylandExtensions::zwp_virtual_keyboard_v1{"zwp_virtual_keyboard_manager_v1"};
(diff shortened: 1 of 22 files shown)


AlanGriffiths (changes requested):
This does add consistency but, unless I miss something, it breaks existing configurations:

Suppose a Mir based system (mir-kiosk say) is deployed and the `.config` says:
```
add-wayland-extension=foo_master
```

With the change of `foo_master` to `foo` that is no longer going to work. There needs to be a way to support configurations with the legacy names.

NB In addition to the real possibilies of mir-kiosk `.config` files and `ubuntu-frame` snap configuration options which we might reasonably migrate with text substitution outside Mir itself there is also the theoretical option of command-line parameters in a runner script.

Hence, I think the best approach is to "translate" any legacy names in `parse_extensions_option()`.

wmww:
That sounds reasonable. Note that this would not effect extensions turned on by shells, but hopefully they were using our helper variables in which case they'll be fine.

AlanGriffiths:
> That sounds reasonable. Note that this would not effect extensions turned on by shells, but hopefully they were using our helper variables in which case they'll be fine.

Yep. I don't think we have enough downstreams to worry about that level of idiot proofing. (Unless you have a neat idea that caters for them too.)

src/miral/wayland_extensions.cpp
```diff
@@ -93,6 +98,30 @@ struct StaticExtensionTracker
 
 decltype(StaticExtensionTracker::mutex)       StaticExtensionTracker::mutex;
 decltype(StaticExtensionTracker::extensions)  StaticExtensionTracker::extensions;
+
+std::map<std::string, std::string> const alternative_extension_names{
+    {"zwp_input_method_v2", "zwp_input_method_manager_v2"},
+    {"zwp_virtual_keyboard_v1", "zwp_virtual_keyboard_manager_v1"},
+};
```
AlanGriffiths, line 105: [Non blocking]

I'm probably old fashioned, but it seems odd to use `map<>` for such small N. A couple of `if ... else if ...` conditions would avoid all the non-constinit setup and teardown.

AlanGriffiths (approved):
bors r+

PR #2211, opened 2021-11-04
Only call extension filter once per app per extension

### LINKED ISSUE #2209: MirAL extension filter invoked multiple times for the same app and same protocol
It is invoked for each protocol once every time the client creates a registry (can be multiple times) and once for every time the global is bound. I see no reason it should be called more than once for each app/protocol combination.

wmww:
Fixes #2209. I was screwing with this code anyway, so I gave it a stab. I don't feel strongly about if this gets merged or not.

### DIFF
--- a/src/server/frontend_wayland/wl_client.cpp
+++ b/src/server/frontend_wayland/wl_client.cpp
@@ -38,10 +38,14 @@ struct ConstructionCtx
     ConstructionCtx(
         std::shared_ptr<msh::Shell> const& shell,
         std::shared_ptr<mf::SessionAuthorizer> const& session_authorizer,
-        std::function<void(mf::WlClient&)>&& client_created_callback)
+        std::function<void(mf::WlClient&)> const& client_created_callback,
+        std::function<bool(std::shared_ptr<ms::Session> const&, char const*)> const& extension_filter)
         : shell{shell},
           session_authorizer{session_authorizer},
-          client_created_callback{std::make_unique<std::function<void(mf::WlClient&)>>(std::move(client_created_callback))}
+          client_created_callback{std::make_unique<std::function<void(mf::WlClient&)>>(
+              client_created_callback)},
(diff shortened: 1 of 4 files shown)


src/server/frontend_wayland/wl_client.cpp
```diff
@@ -38,10 +38,14 @@ struct ConstructionCtx
...
         std::shared_ptr<msh::Shell> const& shell,
         std::shared_ptr<mf::SessionAuthorizer> const& session_authorizer,
-        std::function<void(mf::WlClient&)>&& client_created_callback)
+        std::function<void(mf::WlClient&)> const& client_created_callback,
+        std::function<bool(std::shared_ptr<ms::Session> const&, char const*)> const& extension_filter)
         : shell{shell},
           session_authorizer{session_authorizer},
-          client_created_callback{std::make_unique<std::function<void(mf::WlClient&)>>(std::move(client_created_callback))}
+          client_created_callback{std::make_unique<std::function<void(mf::WlClient&)>>(
+              client_created_callback)},
```
AlanGriffiths, line 46: It's a small pessimisation, but why are we now copying `client_created_callback` instead of moving it? The callback created in `WaylandConnector::WaylandConnector()` is not used elsewhere.

AlanGriffiths (changes requested):
I don't think this was worth writing, but I'm willing to land it if it gets tests for this behaviour.

wmww (commented):
Oops, my comment has been languishing in the "pending" state for all these months, since I forgot to submit.

AlanGriffiths:
This has died of neglect. (If anyone cares to reopen it they should fix the merge conflict)

src/server/frontend_wayland/wl_client.cpp
wmww: Changed to move()

PR #2212, opened 2021-11-05
Improve check-miral-symbols target

wmww:
Renames it to `regenerate-miral-debian-symbols` and makes it always enabled (but off by default on unsupported systems). This allows playing with it everywhere.


AlanGriffiths (approved):
Looks right (hopefully the farm will be less busy)

bors r+

AlanGriffiths:
bors retry

PR #2221, opened 2021-11-10
Use same font search logic in Mir and example server lib

wmww:
As discussed in chat, code duplication is the correct dependency management tool for this situation. This PR fully duplicates the logic, which fixes Miral shell text on Arch (has always been broken)


AlanGriffiths (approved):
bors merge

PR #2223, opened 2021-11-16
Differentiate between active surfaces and surfaces with keyboard input

### LINKED ISSUE #1626: Better handle keyboard input sent to a non-focused surface
There's a mismatch between Mir's keyboard input model and Wayland's. In Mir, keyboard events get sent to surfaces, and in Wayland they get sent to seats. This means that it's theoretically possible for Mir input events to go to surfaces that have not been given keyboard focus, while this is not possible in Wayland. To bridge this gap, we have logic in `WlKeyboard::key` that just-in-time switches focus to the surface that's receiving the events.

This logic was always broken as explained in #1625. That PR seems to improve behavior, but doesn't appear to be completely correct. We should take another look at this whole problem, and make sure everything is doing what it should.

### LINKED ISSUE #2189: OSK disappears when focus returns to window from popup
To reproduce:
- open gedit inside ubuntu-frame with OSK enabled
- ctrl+s
- Save-as window appears, OSK is still there
- esc
- ~~Save-as window (unexpectedly) closes and OSK (unexpectedly) closes~~
- Save-as window (expectedly) closes and OSK (unexpectedly) disappears. Clicking the gedit window does *not* bring it back.

On Sway the OSK does not disappear when save-as window is closed.

wmww:
Fixes #1626, fixes #2189. Adds a new `mir_window_focus_state_active` focus state, which is used for surfaces that are in the active stack but do not have keyboard focus. We did not previously have a way to represent this state properly, which caused us a number of problems.

### DIFF
--- a/include/core/mir_toolkit/common.h
+++ b/include/core/mir_toolkit/common.h
@@ -88,8 +88,9 @@ typedef enum MirWindowState
 
 typedef enum MirWindowFocusState
 {
-    mir_window_focus_state_unfocused = 0,
-    mir_window_focus_state_focused
+    mir_window_focus_state_unfocused = 0,   /**< Inactive and does not have focus           */
+    mir_window_focus_state_focused,         /**< Active and has keybaord focus              */
+    mir_window_focus_state_active           /**< Active but does not have keyboard focus    */
 } MirWindowFocusState;
 
 typedef enum MirWindowVisibility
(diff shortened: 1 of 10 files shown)


AlanGriffiths (commented):
This looks sensible. But the logic is (pre-existing) gnarly - I need to experiment a bit to feel confident we've not introduced unexpected problems.

AlanGriffiths (approved):

AlanGriffiths:
Yes, Seems to be an improvement.

bors r+

PR #2224, opened 2021-11-18
Hack X11 platform to respect power mode

wmww:
While playing with power savings stuff, this is and will be useful to test what we build.


wmww:
An updated version of this was included in #2242

PR #2228, opened 2021-11-23
Allow globals with unknown names

### LINKED ISSUE #2227: Frontend aborts on unknown Wayland global names
The `wl_display_global_filter_func` aborts Mir [if it encounters a global name Mir is unaware of.](https://github.com/MirServer/mir/blob/86ebb35e9f62d33d11cce7b464482f1f03fe57fa/src/server/frontend_wayland/wayland_connector.cpp#L789). `mf::WaylandExtensions::init()` attempts to add known globals which are out of Mir's control, but unfortunately this list can never be complete - Mir graphics platforms can publish Wayland globals of their own. Worse, *EGL* implementations can publish globals, and these are entirely outside Mir's control.

We should not kill the server upon encountering an unknown global. We should work out what else to do.

wmww:
Fixes #2227. We lose a diagnostic, but I don't see a good way of keeping it.

### DIFF
--- a/include/miral/miral/wayland_extensions.h
+++ b/include/miral/miral/wayland_extensions.h
@@ -82,7 +82,7 @@ public:
     /// \remark Since MirAL 2.5
     struct Builder
     {
-        /// Name of the protocol extension
+        /// Name of the protocol extension's Wayland global
         std::string name;
 
         /// Functor that creates and registers an extension protocol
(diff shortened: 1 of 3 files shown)


RAOF (approved):
Yeah, this seems reasonable.

I think we could reuse some of this later, *if* we wanted to prevent shell authors shooting themselves in the foot by eg: filtering out the `wl_compositor` global. We could use the list of extensions this removes to only call the supplied filter on sensible extensions (ie: not `wl_compositor`, or anything added by the EGL implementation, but on any custom extension or something core we expect might be sensibly disabled).

That's something we'd need to decide whether or not is a good idea, anyway. For now,
bors r+

PR #2229, opened 2021-11-23
Do not rescale on every output resize

### LINKED ISSUE #2225: --display-scale 2 results in apps being scaled up every time output is resized
From a visual inspection it looks like apps are being given the correct scale, rendering that and then being blown up more by Mir.

wmww:
Fixes #2225

### DIFF
--- a/src/include/server/mir/graphics/default_display_configuration_policy.h
+++ b/src/include/server/mir/graphics/default_display_configuration_policy.h
@@ -31,48 +31,21 @@ namespace graphics
 class CloneDisplayConfigurationPolicy : public DisplayConfigurationPolicy
 {
 public:
-    CloneDisplayConfigurationPolicy() = default;
-    CloneDisplayConfigurationPolicy(double scale)
-        : scale{scale}
-    {
-    }
-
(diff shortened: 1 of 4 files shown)


RAOF (commented):
I agree that this layered approach of wrapping with `ScaleSetter` is architecturally cleaner. This changes behaviour, though - previously the scale multiplied whatever scale the output had, now it overrides. That seems fine, but is it deliberate?

(It also appears to be missing a test)

AlanGriffiths (approved):
> This changes behaviour, though - previously the scale multiplied whatever scale the output had, now it overrides. That seems fine, but is it deliberate?

It is a correction to unreleased functionality. Seems fine to me.

> (It also appears to be missing a test)

Agreed

bors delegate

wmww:
I don't think testing that a class which sets the scale to 2 actually sets the scale to 2 is particularly useful. I do think testing the whole display scale option system is useful, but I've had a lot of trouble getting those tests working (see https://github.com/mirserver/mir/tree/test-display-config-option). I don't think we should block this on those tests.

PR #2230, opened 2021-11-24
Add IdleHub and support powering down outputs on idle

wmww:
So far:
- Adds new `IdleHub` interface and implementation
- Pokes the idle hub on input events
- Observes the idle state from `` and powers down outputs when needed

Still to do:
- Tests
- Configuration (currently hard-coded in `DefaultServerConfiguration::the_idle_hub()` in src/server/scene/default_configuration.cpp)
- Support for the various relevant Wayland protocols (which will include support for wakelocks)
- Display dimming to warn it will soon be shut off

### DIFF
--- a/src/include/server/mir/scene/idle_hub.h
+++ b/src/include/server/mir/scene/idle_hub.h
@@ -1,0 +1,70 @@
+/*
+ * Copyright © 2021 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored By: William Wold <william.wold@canonical.com>
+ */
+
+#ifndef MIR_SCENE_IDLE_HUB_H_
+#define MIR_SCENE_IDLE_HUB_H_
+
+#include "mir/observer_registrar.h"
+
+#include <memory>
+#include <optional>
+#include <string>
+
+namespace mir
+{
+namespace scene
+{
+enum class IdleState
+{
+    awake,      ///< Outputs should be on
+    dim,        ///< Outputs should be on, but dimmed
+    standby,    ///< Outputs should be in mir_power_mode_standby
+    off,        ///< Outputs should be in mir_power_mode_off
+};
+
+/// Gets notifications about changes in the idle state
+class IdleStateObserver
+{
+public:
+    IdleStateObserver() = default;
+    virtual ~IdleStateObserver() = default;
+
+    virtual void idle_state_changed(IdleState state) = 0;
+
+private:
+    IdleStateObserver(IdleStateObserver const&) = delete;
+    IdleStateObserver& operator=(IdleStateObserver const&) = delete;
+};
+
+/// Interface allowing for input methods to get state from and send text input to clients. When a new observer is
+/// registered, it is immediately given the initial state on the current thread.
+class IdleHub : public ObserverRegistrar<IdleStateObserver>
+{
+public:
+    virtual ~IdleHub() = default;
+
+    /// Current state
+    virtual auto state() -> IdleState = 0;
+
+    /// Wakes Mir if it's not already awake, and starts the timer from now
+    virtual void poke() = 0;
+};
+}
+}
+
+#endif // MIR_SCENE_IDLE_HUB_H_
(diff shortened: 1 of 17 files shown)


wmww:
This is stacked on #2224 so it can be seen to work on X11. It also appears to work on KMS.

src/include/server/mir/scene/idle_hub.h
```diff
@@ -0,0 +1,70 @@
...
+    virtual ~IdleStateObserver() = default;
+
+    virtual void idle_state_changed(IdleState state) = 0;
+
+private:
+    IdleStateObserver(IdleStateObserver const&) = delete;
+    IdleStateObserver& operator=(IdleStateObserver const&) = delete;
+};
+
+/// Interface allowing for input methods to get state from and send text input to clients. When a new observer is
```
RAOF, line 54: Is this comment copied from somewhere? It doesn't appear to correspond to the code.

src/server/scene/basic_idle_hub.cpp
```diff
@@ -0,0 +1,103 @@
...
+        std::vector<StateEntry> const& state_sequence,
+        std::shared_ptr<mt::AlarmFactory> const& alarm_factory)
+    : state_sequence{add_awake_at_start(state_sequence)},
+      next_state_alarm{alarm_factory->create_alarm([this]()
+          {
+              increment_state();
+          })},
+      current_state_index{0}
+{
+    std::unique_lock<std::mutex> lock{mutex};
```
RAOF, line 48: There's no need to lock in the constructor; it's not possible for this object to be shared between threads yet.

RAOF (changes requested):
Apart from the nits in the detailed review, I think this is the wrong interface - simultaneously too general and too specific? Eg: `enum IdleState` maps directly on to output power modes, so it's not clearly appropriate for anything *other* than setting the device power state, but it *also* contains the `dim` value, which is (a) currently unused and (b) doesn't correspond to anything Mir has control over.

Also, this interface cannot be used to build a `kde_idle` implementation, so I think it reasonably likely we'll want a more powerful, less display-coupled interface.

Probably the interface we need is something that lets code register interest in when idle time exceeds some value, and for the idle->active transition?

src/server/scene/basic_idle_hub.cpp
wmww: This is to call `set_state_index()` on the next line, which takes and unlocks a lock. It could be avoided (ie by using an optional lock) if there was any reason to, but since it's guaranteed to be uncontested and thus pretty fast this is simpler.

src/include/server/mir/scene/idle_hub.h
wmww: Oops!

wmww:
First, the specific points and then I'll explain my reasoning for the overall approach:

> dim value, which is (a) currently unused and (b) doesn't correspond to anything Mir has control over.

It is not implemented yet (draft PR), but I imagine something in the graphics subsystem could implement it easily enough. It doesn't have to lower the actual brightness, just perceived brightness (it's a warning that the screen is about to go blank if you don't move the cursor, not a power saving feature itself). This is now explained better in the doc comment.

> this interface cannot be used to build a kde_idle implementation

Correct, it will need to be extended if we want to implement that protocol. I don't think that sort of interface is the bast option for everything though.

One of the main considerations that may not be obvious (because it's not yet implemented) is configuration. If we have a finite list of states, the user can configure which of those states are used and what the timeout on them is in one place. The idle hub can hold and understand that configuration, and all other subsystems just have to wait for the state changes they're interested in.

If, on the other hand, the idle hub only exposes the ability to register timeouts then every subsystem has to manage it's own configuration. The graphics code that will dim the output and the display code that will put it in standby are completely different subsystems. If you want to debug why the screen is only dim for half a second before it blanks, there should be a single place to look for that timing logic.

RAOF:
So, I'm not sure. Those are reasonable points, but I'm still not quite comfortable.

> >this interface cannot be used to build a kde_idle implementation
>
>  Correct, it will need to be extended if we want to implement that protocol. I don't think that sort of interface is the bast option for everything though.

So, I don't think you'd *extend* this to implement this interface; you'd implement *this* interface using the lower-level timer interface, and also implement `kde_idle` using that lower-level interface.

> If, on the other hand, the idle hub only exposes the ability to register timeouts then every subsystem has to manage it's own configuration. The graphics code that will dim the output and the display code that will put it in standby are completely different subsystems. If you want to debug why the screen is only dim for half a second before it blanks, there should be a single place to look for that timing logic.

Aaah. I think *this* might be the root of my misgivings? I think that I *don't* think that subsystems should register to listen to `IdleHub` announcements so they can act when appropriate. Rather, `IdleHub` should make *requests* to subsystems. That way there *is* one place to look for timing logic. It makes the fact that `enum IdleState` almost, but not quite, maps directly onto output power modes more reasonable - other subsystems aren't going to need to make sense of an `IdleState`. And finally, it makes it easier to effect policies that affect more than one subsystem. For example, `IdleState::dim` *could* be handled in the display subsystem for some platforms, but would need to be handled by a change of rendering otherwise, but you wouldn't want *both* to be applied. If `IdleHub` is in charge of the policy, you can easily try the display method and then fall back on graphics.

So, I think a better design here would be to implement a timer-registrar thing and then implement the policy component on top of that as the sole (for now) consumer of the timer-registrar. I don't think that's an unreasonable amount of work from where it is here?

wmww:
That sounds like a reasonable way to do it, I'll whip something up.

wmww:
Closed in favor of #2242

Issue #2226, opened 2021-11-23
gbm-kms platform assumes XRGB8888 is an acceptable framebuffer format


RAOF:
This is incorrect on at least one platform, which supports only the ARGB8888 gbm format.

PR #2232, opened 2021-11-29
gbm-kms: Try harder to pick a working GBM framebuffer format

RAOF:
Some drivers (*cough* Mali) don't support rendering to `RGBX` GBM buffers. We don't care about the difference between `RGBX` and `RGBA` for framebuffers, so check whether `RGBX` will work and try to fall back to `RGBA` if it doesn't.

There are some obvious points of further work here - checking what formats the *display* hardware supports and using that to populate the `MirDisplayConfiguration`, respecting the pixel format specified in the `MirDisplayConfiguration`, and so on - but that should be deferred until we make the switch to Atomic KMS.

Closes: #2226

### DIFF
--- a/include/test/mir/test/doubles/mock_egl.h
+++ b/include/test/mir/test/doubles/mock_egl.h
@@ -89,6 +89,11 @@ MATCHER_P2(EGLConfigContainsAttrib, attrib, value, "")
         ++arg_mut;
     }
 
+    size_t end{0};
+    for (; arg[end] != EGL_NONE; ++end);
+
+    *result_listener << "attribute list is " << testing::PrintToString(std::vector<EGLint>(arg, arg + end));
+
     return false;
 }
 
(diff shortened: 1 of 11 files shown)


RAOF:
Ahem. Now with remembering to run and update the tests!

AlanGriffiths (approved):
This has an unfinished feel to it. Especially with the raft of TODOs it introduces and lack of tests for the behaviour it adds.

But as it reportedly fixes a real world issue (and doesn't look likely to break anything new) I'm OK with landing it.

bors delegate

RAOF:
All but one of the TODOs is an existing limitation I'm pointing out.

You're right, though, it should have a test.

AlanGriffiths (approved):
WFM

bors r+

Saviq:
bors retry

PR #2239, opened 2021-12-06
gbm-kms: Attempt rendering probe with a DRM master fd.

RAOF:
It seems the Mali EGL implementation requires DRM master. In order for
probe() to succeed, try giving it one.

But, since *most* platforms don't need a master fd to bring EGL up,
fall back to plain `::open()`, as before.


RAOF:
Turns out this doesn't actually work - by the time we probe the rendering platform the display platform has already claimed DRM master and so the rendering probe still fails.

PR #2242, opened 2021-12-06
Add IdleHub and IdleHandler

wmww:
Alternative to #2230, using @RAOF's suggestion. After implementing both, I agree this is the superior (albeit somewhat more complex) design. No tests until I get feedback that I should move forward with it. It now dims the display after 5 seconds, and turns it off after 10. Based on top of #2224.


RAOF:
This looks good to me (modulo some naming nits, which I need to actually record); feel free to add the tests ;)

Saviq:
```plain
[ 82%] Building CXX object src/miral/CMakeFiles/miral.dir/append_event_filter.cpp.o
In file included from /usr/include/gtest/gtest-matchers.h:47,
                 from /usr/include/gtest/internal/gtest-death-test-internal.h:39,
                 from /usr/include/gtest/gtest-death-test.h:41,
                 from /usr/include/gtest/gtest.h:64,
                 from /usr/src/googletest/googlemock/include/gmock/internal/gmock-internal-utils.h:47,
                 from /usr/src/googletest/googlemock/include/gmock/gmock-actions.h:51,
                 from /usr/src/googletest/googlemock/include/gmock/gmock.h:59,
                 from /root/parts/mir/build/tests/unit-tests/CMakeFiles/mir_unit_tests.dir/cmake_pch.hxx:5,
                 from <command-line>:
/usr/include/gtest/gtest-printers.h: In instantiation of ‘static void testing::internal2::TypeWithoutFormatter<T, kTypeKind>::PrintValue(const T&, std::ostream*) [with T = mir::Executor; testing::internal2::TypeKind kTypeKind = testing::internal2::kOtherType; std::ostream = std::basic_ostream<char>]’:
/usr/include/gtest/gtest-printers.h:246:66:   required from ‘std::basic_ostream<_CharT, _Traits>& testing::internal2::operator<<(std::basic_ostream<_CharT, _Traits>&, const T&) [with Char = char; CharTraits = std::char_traits<char>; T = mir::Executor]’
/usr/include/gtest/gtest-printers.h:287:7:   required from ‘void testing_internal::DefaultPrintNonContainerTo(const T&, std::ostream*) [with T = mir::Executor; std::ostream = std::basic_ostream<char>]’
/usr/include/gtest/gtest-printers.h:468:49:   required from ‘void testing::internal::DefaultPrintTo(testing::internal::WrapPrinterType<testing::internal::kPrintOther>, const T&, std::ostream*) [with T = mir::Executor; std::ostream = std::basic_ostream<char>]’
/usr/include/gtest/gtest-printers.h:503:17:   required from ‘void testing::internal::PrintTo(const T&, std::ostream*) [with T = mir::Executor; std::ostream = std::basic_ostream<char>]’
/usr/include/gtest/gtest-printers.h:679:12:   required from ‘static void testing::internal::UniversalPrinter<T>::Print(const T&, std::ostream*) [with T = mir::Executor; std::ostream = std::basic_ostream<char>]’
/usr/include/gtest/gtest-printers.h:869:30:   [ skipping 3 instantiation contexts, use -ftemplate-backtrace-limit=0 to disable ]
/usr/include/gtest/gtest-printers.h:629:15:   required from ‘void testing::internal::PrintTupleTo(const T&, std::integral_constant<long unsigned int, _Nm>, std::ostream*) [with T = std::tuple<const std::weak_ptr<mir::scene::IdleStateObserver>&, mir::Executor&, std::chrono::duration<long int, std::ratio<1, 1000> > >; long unsigned int I = 3; std::ostream = std::basic_ostream<char>]’
/usr/include/gtest/gtest-printers.h:642:15:   required from ‘void testing::internal::PrintTo(const std::tuple<_Tps ...>&, std::ostream*) [with Types = {const std::weak_ptr<mir::scene::IdleStateObserver>&, mir::Executor&, std::chrono::duration<long int, std::ratio<1, 1000> >}; std::ostream = std::basic_ostream<char>]’
/usr/include/gtest/gtest-printers.h:679:12:   required from ‘static void testing::internal::UniversalPrinter<T>::Print(const T&, std::ostream*) [with T = std::tuple<const std::weak_ptr<mir::scene::IdleStateObserver>&, mir::Executor&, std::chrono::duration<long int, std::ratio<1, 1000> > >; std::ostream = std::basic_ostream<char>]’
/usr/include/gtest/gtest-printers.h:869:30:   required from ‘void testing::internal::UniversalPrint(const T&, std::ostream*) [with T = std::tuple<const std::weak_ptr<mir::scene::IdleStateObserver>&, mir::Executor&, std::chrono::duration<long int, std::ratio<1, 1000> > >; std::ostream = std::basic_ostream<char>]’
/usr/src/googletest/googlemock/include/gmock/gmock-spec-builders.h:1679:19:   required from ‘void testing::internal::FunctionMocker<R(Args ...)>::UntypedDescribeUninterestingCall(const void*, std::ostream*) const [with R = void; Args = {const std::weak_ptr<mir::scene::IdleStateObserver>&, mir::Executor&, std::chrono::duration<long int, std::ratio<1, 1000> >}; std::ostream = std::basic_ostream<char>]’
/usr/src/googletest/googlemock/include/gmock/gmock-spec-builders.h:1671:8:   required from here
/usr/include/gtest/gtest-printers.h:157:9: error: invalid application of ‘sizeof’ to incomplete type ‘const mir::Executor’
  157 |         sizeof(value), os);
      |         ^~~~~~~~~~~~~
make[2]: *** [tests/unit-tests/CMakeFiles/mir_unit_tests.dir/build.make:1189: tests/unit-tests/CMakeFiles/mir_unit_tests.dir/input/test_idle_poking_dispatcher.cpp.o] Error 1
```

src/server/shell/basic_idle_handler.cpp
```diff
@@ -0,0 +1,181 @@
...
+        std::shared_ptr<mi::Scene> const& input_scene,
+        std::shared_ptr<mg::GraphicBufferAllocator> const& allocator)
+        : input_scene{input_scene},
+          allocator{allocator}
+    {
+    }
+
+    void active() override
+    {
+        std::lock_guard<std::mutex> lock{mutex};
```
RAOF, line 106: I feel like this internal locking *should* be unnecessary, and either `IdleHub` should provide the guarantee that it's not going to call `idle()` and `active()` at the same time on different threads or this code should provide it with an `Executor` making that guarantee.

src/server/shell/basic_idle_handler.cpp
```diff
@@ -0,0 +1,181 @@
...
+    std::shared_ptr<ms::IdleHub> const& idle_hub,
+    std::shared_ptr<input::Scene> const& input_scene,
+    std::shared_ptr<graphics::GraphicBufferAllocator> const& allocator,
+    std::shared_ptr<msh::DisplayConfigurationController> const& display_config_controller)
+    : idle_hub{idle_hub},
+      timeouts{
+          Timeout{5s, std::make_shared<Dimmer>(input_scene, allocator)},
+          Timeout{10s, std::make_shared<PowerModeSetter>(display_config_controller, mir_power_mode_standby)},
+          Timeout{15s, std::make_shared<PowerModeSetter>(display_config_controller, mir_power_mode_suspend)},
+          Timeout{20s, std::make_shared<PowerModeSetter>(display_config_controller, mir_power_mode_off)},
```
RAOF, line 166: These seem like pretty aggressive defaults? Dim the screen after 5 seconds of inactivity, turn it off after 20? That's much more aggressive than my phone :)

RAOF (changes requested):
I'm *pretty* sure the locking in `BasicIdleHub` is incorrect, and can deadlock.
Demonstration:
|Thread A | Thread B |
|---|---|
| <alarm timeout fires> | |
| Alarm takes internal lock | |
| | `register_interest` called |
| | locks `BasicIdleHub::mutex` |
| Alarm calls `alarm_fired()` | |
| Waits to lock `BasicIdleHub::mutex` | |
| | ... |
| | calls `alarm->reschedule_in()` |
| | waits to lock Alarm-internal mutex |

-----
I think it'd be good to have a comment outlining the theory of operation of `BasicIdleHub`. Because we've talked about it, I know that it's "set an alarm for the earliest timeout, then when that timer fires set an alarm for the difference between that timeout and the next timeout", but I don't think that's entirely obvious just from the code.
------
It would be nice to have some tests for `BasicIdleHub`, too (although they wouldn't necessarily have caught the locking snafu)

src/server/shell/basic_idle_handler.cpp
wmww: Ah yes, that was just for testing purposes. This really should be configurable.

src/server/scene/basic_idle_hub.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include "mir/time/alarm.h"
+#include "mir/time/alarm_factory.h"
+#include "mir/time/clock.h"
+
+namespace ms = mir::scene;
+namespace mt = mir::time;
+
+namespace
+{
+struct DirectExecutor: mir::Executor
```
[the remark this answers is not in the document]
wmww: I've wanted something like that for a while. Since you seem to have a specific idea of how to do it, want to make the PR?

wmww:
Ah! This situation seems to be what `AlarmFactory::create_alarm(std::unique_ptr<LockableCallback>)` was made for.

src/server/input/idle_poking_dispatcher.cpp
```diff
@@ -0,0 +1,65 @@
...
+    bool const result = next_dispatcher->dispatch(event);
+    switch (mir_event_get_type(event.get()))
+    {
+    case mir_event_type_input:
+        switch (mir_input_event_get_type(mir_event_get_input_event(event.get())))
+        {
+            case mir_input_event_type_key:
+            case mir_input_event_type_pointer:
+            case mir_input_event_type_touch:
+                idle_hub->poke();
```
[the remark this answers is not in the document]
wmww: Fixed

src/server/shell/basic_idle_handler.cpp
wmww: Does the main loop executor make such a guarantee

RAOF:
Looking good. I'll get you a PR with `mir::system_executor` and `mir::sequential_executor` up this afternoon.

RAOF:
(`mir::system_executor` now available in #2295)

wmww:
I think this may be ready to go now. The timeout of `IdleHandler` can now be dynamically set, and is off by default. Since nothing is setting it, this PR should now not change Mir behavior. A follow up PR will add a configuration option that can be used to set a timeout, but it will stay disabled by default for backwards compatibility.

src/server/scene/basic_idle_hub.cpp
```diff
@@ -0,0 +1,246 @@
...
+    {
+        iter->second->register_interest(observer, executor);
+        is_active = (alarm_timeout && alarm_timeout.value() <= timeout);
+    }
+    lock.unlock();
+    if (auto const shared = observer.lock())
+    {
+        if (is_active)
+        {
+            shared->active();
```
RAOF, line 174: Oh! This isn't dispatched on the `executor`, like other all observations will be. It probably should be?

src/server/scene/basic_idle_hub.cpp
```diff
@@ -0,0 +1,246 @@
...
+        }
+        else
+        {
+            shared->idle();
+        }
+    }
```
RAOF, line 180: I feel like there's a race here, where a timeout is registered just as it would timeout
|Thread A | Timer Thread|
|-|-|
|`alarm->reschedule_in(1)`||
|`is_active = true`||
| `lock.unlock()` ||
| <Thread loses timeslice, and is scheduled out> | |
| | <Timer fires> |
| | <Calls `observer->idle()`> |
| `if (is_active)` branch is taken| |
| `observer->active()` is called | |

This results in the observer believing Mir is currently active, and once `poke()` is called will result in `active()` being called a second time in a row.

On the one hand, this is not actually likely to happen (with the current code), as the observers are registered once at startup. On the other hand, *debugging* this should it occur would be a nightmare; it would manifest as *sometimes* it appearing that an idle timeout is failing to fire.

RAOF: A corollary of this is that code *using* the `IdleHub` almost certainly needs to be using a serialising `Executor` if it wants to act on *state*, rather than just transitions.

RAOF (commented):
Just a comment for now, as I've run out of time to review this today. I'll get back to it tomorrow.

src/server/scene/basic_idle_hub.cpp
AlanGriffiths: Just an observation at this time - I've not understood this code well enough to decide if there are real problems.

This discussion is reminiscent of the early "observer" iterations that led to `BasicObservers`. The story is here:

https://accu.org/journals/overload/22/124/griffiths_2040/

src/include/server/mir/scene/idle_hub.h
```diff
@@ -0,0 +1,78 @@
...
+    /// Timeout is the amount of time Mir will stay idle before triggering the observer
+    virtual void register_interest(
+        std::weak_ptr<IdleStateObserver> const& observer,
+        time::Duration timeout) = 0;
+
+    /// Timeout is the amount of time Mir will stay idle before triggering the observer
+    virtual void register_interest(
+        std::weak_ptr<IdleStateObserver> const& observer,
+        Executor& executor,
+        time::Duration timeout) = 0;
```
AlanGriffiths, line 67: This overload seems to be an implementation detail of `BasicIdleHub` - there's no need (at present) to expose this in the `IdleHub` interface.

src/server/shell/basic_idle_handler.h
```diff
@@ -0,0 +1,76 @@
...
+public:
+    BasicIdleHandler(
+        std::shared_ptr<scene::IdleHub> const& idle_hub,
+        std::shared_ptr<input::Scene> const& input_scene,
+        std::shared_ptr<graphics::GraphicBufferAllocator> const& allocator,
+        std::shared_ptr<shell::DisplayConfigurationController> const& display_config_controller);
+
+    ~BasicIdleHandler();
+
+    void set_display_off_timeout(std::optional<time::Duration> timeout) override;
```
AlanGriffiths, line 58: Where is this used outside of tests? Without calling this the PR is pointless.

AlanGriffiths (changes requested):
Some of the tests are fragile: they test things that are not mentioned in the test name (and are tested elsewhere). I've highlighted one example.

AFAICS `BasicIdleHandler` is the only `IdleHub` user and, assuming that `set_display_off_timeout()` is intended to be called during initialization, has no need for a callback notification when registering interest, only on transitions. Removing that callback would address @RAOF 's concerns about it being racy.

src/include/server/mir/scene/idle_hub.h
wmww: It's not strictly needed, but it makes the interface consistent with `ObserverRegistrar`

src/server/shell/basic_idle_handler.h
wmww: The idea was a followup PR would add a configuration option that could be used to set it, thought now that we're under time crunch I'll add that to this PR if doesn't land before I get the chance.

wmww:
It took a little weirdness to have it make the initial calls on the given executor, maintain the preregister guarantees, and remain safe when the idle hub is destroyed, but hopefully I got that right. Now all it needs is #2308.

src/server/scene/basic_idle_hub.cpp
```diff
@@ -0,0 +1,265 @@
...
+                if (is_active)
+                {
+                    shared_observer->active();
+                }
+                else
+                {
+                    shared_observer->idle();
+                }
+            }
+        });
```
RAOF, line 203: Oh! This code is here to make `unregister_interest` guarantee that no further calls will be made to the observer. Urgh, that's gnarly.

I've not quite convinced myself that it's correct yet. I feel like it should *probably* be replaced by adding a way for `ObserverMultiplexer::register_interest` to send an initial state, but I'm not quite sure how that would look. If it stays, it could do with a comment saying what it's for :)

wmww:
I updated `ObserverMultiplexer` so an observation can be sent to a single observer, maintaining the rest of the ordering/executor guarantees. I have the impression that in addition to sending initial state, this feature might also be useful when porting other Mir observer systems to the new one (thought it's been a while since I looked into that, could be mistaken). The perf characteristics aren't great. It has to lock all existing executors to send a notification to one of them, but I don't think that matters too much. We can always complicate/optimize the code later if it becomes an issue.

wmww:
`--idle-timeout` option can now be set.

src/server/shell/basic_idle_handler.cpp
AlanGriffiths: We now have [`linearising_executor`](https://github.com/MirServer/mir/pull/2308) that makes these guarantees.

src/server/shell/default_configuration.cpp
```diff
@@ -89,6 +92,27 @@ auto mir::DefaultServerConfiguration::wrap_shell(std::shared_ptr<msh::Shell> con
...
+                the_input_scene(),
+                the_buffer_allocator(),
+                the_display_configuration_controller());
+
+            auto options = the_options();
+            int const idle_timeout_seconds = options->get<int>(options::idle_timeout_opt);
+            if (idle_timeout_seconds > 0)
+            {
+                idle_handler->set_display_off_timeout(std::chrono::seconds{idle_timeout_seconds});
+            }
```
AlanGriffiths, line 110: If `idle_timeout_seconds` is negative do we want to quietly ignore it? Or, perhaps, fail noisily?

tests/unit-tests/shell/test_basic_idle_handler.cpp
```diff
@@ -0,0 +1,204 @@
...
+        mt::fake_shared(input_scene),
+        mt::fake_shared(allocator),
+        mt::fake_shared(display)};
+    EXPECT_CALL(idle_hub, register_interest(_, _))
+        .Times(AnyNumber());
+    EXPECT_CALL(idle_hub, register_interest(_, Eq(30s)))
+        .WillOnce(SaveArg<0>(&observer));
+    handler.set_display_off_timeout(30s);
+    auto const locked = observer.lock();
+    ASSERT_THAT(locked, Ne(nullptr));
```
AlanGriffiths, line 124: This is all test setup right? We ought to `Mock::VerifyAndClearExpectations(&idle_hub)` before continuing. That both ensures we're not matching any "setup" expectations during the test and provides a visual aid to knowing what is part of the test proper.

tests/unit-tests/shell/test_basic_idle_handler.cpp
```diff
@@ -0,0 +1,204 @@
...
+    EXPECT_CALL(idle_hub, unregister_interest(_))
+        .Times(AnyNumber());
```
AlanGriffiths, line 126: It takes a careful reading to see that this is not part of the setup, but filtering "uninteresting" events. Maybe this would be clearer:
```suggestion
    EXPECT_CALL(idle_hub, unregister_interest(Not(Ref(*locked))))
        .Times(AnyNumber());
```

src/include/server/mir/scene/idle_hub.h
```diff
@@ -0,0 +1,78 @@
...
+#include <memory>
+#include <optional>
+#include <string>
+
+namespace mir
+{
+class Executor;
+namespace scene
+{
+/// Gets notifications about changes in the idle state
```
AlanGriffiths, line 33: This comment is wrong.
```suggestion
/// Gets notifications about the idle state when registered and subsequent state changes until unregistered
```

AlanGriffiths (changes requested):
The main thing is the tests are hard to follow. I've commented in detail on one example I think is typical of the problems. But even a bit of whitespace (blank line) between "setup" and "test body" would help a lot.

There's also @RAOF comments that presumably led to his "linearising" that can now be resolved.

Finally, there's a lot of small stuff noted that I wouldn't block on, but might be nice to address.

src/server/shell/default_configuration.cpp
```diff
@@ -104,7 +104,15 @@ auto mir::DefaultServerConfiguration::the_idle_handler() -> std::shared_ptr<msh:
 
             auto options = the_options();
             int const idle_timeout_seconds = options->get<int>(options::idle_timeout_opt);
-            if (idle_timeout_seconds > 0)
+            if (idle_timeout_seconds < 0)
+            {
+                fatal_error("Invalid %s value %d, must be > 0", options::idle_timeout_opt, idle_timeout_seconds);
```
AlanGriffiths, line 109: I'd probably use `throw mir::AbnormalExit(...` (but non-blocking as it is a stupid user-error path.

wmww: Easy enough

AlanGriffiths (approved):
OK. A bit of a monster, so I'd prefer a second set of eyes. But low risk as it is dead code by default.

bors merge

PR #2243, opened 2021-12-07
gbm-kms: Be more relaxed in the rendering probe.

RAOF:
The libmali EGL implementation requires that the `gbm_device` passed in to
`eglGetPlatformDisplay` be created from a DRM master fd.

We can't guarantee this in the rendering probe, as the display platform may
have already claimed DRM master.

Instead, assume that an EGL platform which claims to support the GBM platform
will be at least minimally usable with GBM.


AlanGriffiths (approved):
bors merge

AlanGriffiths:
bors retry

PR #2244, opened 2021-12-07
Give keyboard focus to menus (grabbing popups)

### LINKED ISSUE #2241: Keyboard focus given to popup only on key press
Reproduce:
- Open gnome-terminal
- Open right-click menu
- Press a key

Expected: when menu opens, the cursor in the main window becomes empty

Actual: the main window cursor keeps blinking until a key is pressed, and Mir gives a warning message:
```
[2021-12-06 17:09:26.497322] < -warning- > mirserver: Sending keyboard event to wl_surface@34 even though it was not explicitly given keyboard focus
```

This also manifests as a minor OSK bug: the OSK does not go away as it should when the menu appears, and can be used to type into the main surface while the menu is still up (which does not dismiss the menu).

EDIT: this effects ALL apps, not just gnome-terminal

wmww:
According to XDG Shell, grabbing popups should be given keyboard focus and non-grabbing ones should not. This PR brings our keyboard focus logic in line with that.

Fixes #2241. Tested by the recently-updated https://github.com/MirServer/wlcs/pull/216.

### DIFF
--- a/src/server/shell/abstract_shell.cpp
+++ b/src/server/shell/abstract_shell.cpp
@@ -416,14 +416,13 @@ void msh::AbstractShell::notify_focus_locked(
 {
     auto const current_focus = notified_keyboard_focus_surface.lock();
 
-    // Don't give keyboard focus to popups
+    // Don't give keyboard focus to non-grabbing popups
     auto new_keyboard_focus = new_active_surface;
     while (new_keyboard_focus)
     {
         auto const type = new_keyboard_focus->type();
         if (type != mir_window_type_gloss &&
-            type != mir_window_type_tip &&
-            type != mir_window_type_menu)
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
At first (and second) look, this seems to remove the need for the following hack:
```c++
    if (!focused_surface.is(surface))
    {
        log_warning(
            "Sending keyboard event to wl_surface@%u even though it was not explicitly given keyboard focus",
            wl_resource_get_id(surface.resource));

        focussed(surface, true);
    }
```

Maybe I'll add that as a follow-on.

AlanGriffiths (approved):

Saviq (approved):
Looks legit, confirmed the desired behaviour and ACK on the broken tests.

bors merge

PR #2245, opened 2021-12-08
Don't send keyboard events without focus

AlanGriffiths:


src/server/frontend_wayland/wl_keyboard.cpp
```diff
@@ -49,10 +49,10 @@ void mf::WlKeyboard::handle_event(MirInputEvent const* event, WlSurface& surface
     if (!focused_surface.is(surface))
     {
         log_warning(
```
Saviq, line 51: Shouldn't this be a `log_error` now? It happening would be an error in our WM logic, no?

src/server/frontend_wayland/wl_keyboard.cpp
```diff
@@ -49,10 +49,10 @@ void mf::WlKeyboard::handle_event(MirInputEvent const* event, WlSurface& surface
...
-            "Sending keyboard event to wl_surface@%u even though it was not explicitly given keyboard focus",
+            "Attempting to sending keyboard event to wl_surface@%u even though it was not given keyboard focus",
```
Saviq, line 52:
```suggestion
            "Attempt to send keyboard event to wl_surface@%u even though it was not given keyboard focus",
```

Saviq (changes requested):

src/server/frontend_wayland/wl_keyboard.cpp
AlanGriffiths: Yes, but it should be `fatal_error()`

Saviq (approved):
Good!

bors merge

PR #2251, opened 2021-12-10
Add `-gdb` option to miral-app script to run the server under gdb

AlanGriffiths:


wmww:
Perhaps I can also later add an option to run in GDB + wayland-debug? (wayland-debug launches GDB, so it would have to be separate).

Saviq:
bors retry

PR #2254, opened 2021-12-13
XWayland fixes for dialog window type

### LINKED ISSUE #2250: [Xwayland] weird focus problems with CLion
There are a lot of similar scenarios, documenting one for consistency.

In the CLion edit window start typing something (e.g. "ret") that gets a suggestions popup. `TAB` to select a suggestion.

Expect: text cursor after the inserted text and able to type.
Actual: no text cursor and unable to type.

The editor does still respond to the keyboard (e.g. `UP` and `DOWN` arrows move the current line highlight).

We've seen this before and hacked a solution, but it has returned with Mir 2.6.

Note: CLion is using our X11 support (Xwayland) and is Java app with sometimes odd ways of doing things on X11.

wmww:
The first, smaller commit fixes #2250, the 2nd one makes the behavior more complaint with the spec. I've tested this with intellij (which I was able to reproduce #2250 on), Kate and a couple other things. More manual testing is highly advised since correct behavior in this area is entirely non-obvious.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -169,7 +169,9 @@ auto wm_resize_edge_to_mir_resize_edge(NetWmMoveresize wm_resize_edge) -> std::o
 
 auto wm_window_type_to_mir_window_type(
     mf::XCBConnection* connection,
-    std::vector<xcb_atom_t> const& wm_types) -> MirWindowType
+    std::vector<xcb_atom_t> const& wm_types,
+    bool is_transient_for,
+    bool override_redirect) -> MirWindowType
 {
     // See https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm46515148839648
     for (auto const& wm_type : wm_types)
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
This does fix the case documented in #2250, but there's another case that still fails. (Documented there)

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -182,10 +184,6 @@ auto wm_window_type_to_mir_window_type(
             // TODO: perhaps desktop and dock should set the layer as well somehow?
             return mir_window_type_freestyle;
         }
-        if (wm_type == connection->_NET_WM_WINDOW_TYPE_DIALOG)
-        {
-            return mir_window_type_dialog;
-        }
```
AlanGriffiths, line 188: It seems unintuitive that `_NET_WM_WINDOW_TYPE_DIALOG` isn't a dialog and using "gloss" instead doesn't seem justified by a first reading of the spec. (It does work better though!)

AlanGriffiths (approved):
> there's another case that still fails

But that also fails with 2.5, so we can land this and I'll raise it separately.

bors merge

PR #2263, opened 2021-12-28
Package config should not mention defunct mirclient

AlanGriffiths:


AlanGriffiths (commented):
As there's nobody around:

bors r+

PR #2265, opened 2022-01-03
Fix pkg-config mistakes

AlanGriffiths:


RAOF (approved):
Yup, those seem sensible

bors r+

PR #2267, opened 2022-01-04
release/1.8: Work around capnproto incompatibility

z3ntu:
For some reason including capnproto before X11 header makes the compiler
happy.

See https://github.com/capnproto/capnproto/issues/1393

---

Capnproto is gone with Mir 2.x so it's not a problem there fortunately.


RAOF (approved):
Can't¹ hurt!

bors r+

¹: Or, at least, is *extremely unlikely* to hurt :wink:

PR #2270, opened 2022-01-06
Deactivate text input when the Wayland object is deleted

### LINKED ISSUE #2269: OSK does not hide when application closes
Reproduced with ubuntu-frame and egmde (candidate)

1. Start server and OSK: `ubuntu-frame& ubuntu-frame-osk&`  (OSK does not appear - as expected)
2. Start gedit: `gedit` (OSK appears - as expected)
3. Close gedit: E.g. `Ctrl-C`

**Expect:** OSK closes
**Actual:** OSK remains open

Even after closing and restarting `ubuntu-frame-osk` it remains visible suggesting that internal state of Mir is incorrect rather than the OSK misbehaving.

wmww:
That was easy.

Fixes #2269.

Tested by https://github.com/MirServer/wlcs/pull/217.

### DIFF
--- a/src/server/frontend_wayland/text_input_v3.cpp
+++ b/src/server/frontend_wayland/text_input_v3.cpp
@@ -282,6 +282,7 @@ mf::TextInputV3::TextInputV3(
 mf::TextInputV3::~TextInputV3()
 {
     seat.remove_focus_listener(client, this);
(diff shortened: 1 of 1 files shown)


AlanGriffiths (approved):
WFM

bors r+

PR #2273, opened 2022-01-10
WlKeyboard: use WlSeat::FocusListener for focus tracking

### LINKED ISSUE #2272: Mir fatal errors when Firefox gets initial keyboard input
To reproduce:
- Open Mir
- Open Firefox on Wayland in Mir
- Press a key without unfocusing Firefox
```
Mir fatal error: Attempt to send keyboard event to wl_surface@25 even though it was not given keyboard focus
```

wmww:
It seems at some point we made `WlSeat` expose a generic focus tracking API, but never ported `WlKeyboard` to use it. This PR removes that code duplication, and because `WlSeat::FocusListener` gets notified of the focus state when initially added, it fixes #2272 for free.


AlanGriffiths (approved):
LGTM

bors r+

PR #2289, opened 2022-01-20
Fix logic  for dropping X11 key repeats

AlanGriffiths:


wmww (approved):
I didn't notice any differences in behavior on the apps I tested (gedit and blender), but it looks more correct and makes sense
bors r+

AlanGriffiths:
> I didn't notice any differences in behavior on the apps I tested (gedit and blender), but it looks more correct and makes sense

It can be hard to distinguish X11 host based key repeats from client generated ones.

I've actually been thinking it would make more sense not to filter out the key repeats here, but instead to disable them for clients when hosted on X11. (But that's a different PR, this just reinstates some of the logic lost in the last commit.)

PR #2294, opened 2022-01-25
Refactor mtd::AdvanceableClock

wmww:
This rolls `mt::FakeClock` into `mtd::AdvanceableClock`, and makes some improvements. It's included in #2242. `FakeTimer` is currently unused, but is needed by #2242 so I keep it around.


wmww:
I just noticed `mtd::FakeAlarmFactory` exists, and should probably be used in favor of `mtd::FakeTimer`. Tomorrow I'll work out how this should be handled.

wmww:
No longer needed

PR #2295, opened 2022-01-25
Add mir::system_executor

RAOF:
We have a bunch of places where code requires an ambient `Executor`.
Currently, these are handled by ad-hoc implementations.

Add a global `mir::system_executor` `Executor` implementation to mircommon,
so it's available to all Mir code.


src/common/system_executor.cpp
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    std::function<void()> work;
+    std::atomic<bool>* shutdown;
+    std::binary_semaphore work_available;
+    std::thread thread;
+};
+
+class ThreadPool : public mir::SystemExecutor
```
wmww, line 137: Why doesn't this use `mir::thread::BasicThreadPool` (the thread pool used by `MultiThreadedCompositor`)? Their implementations look quite a bit different, but do they have different requirements?

wmww (changes requested):
Mostly looks good. but I don't want to introduce class duplication unless it's justified.

src/common/system_executor.cpp
RAOF: So `BasicThreadPool` lives in `mirserver`, and is tied to symbols there, so it can't go into `mircommon` without changes (which, admittedly, would be pretty easy). It also has a bunch of superfluous functionality that doesn't get used at all - particularly, scheduling work on a specific thread and each thread having a queue of work. It is also not self-limiting - a manual call to `shrink()` is necessary in order to reap threads (although, again, this could be worked around).

It'd be simple to get `MultiThreadedCompositor` to use `mir::system_executor` instead; I'll try that.

RAOF:
`/spread/mir/src/common/system_executor.cpp:133:10: error: 'binary_semaphore' in namespace 'std' does not name a type; did you mean 'binary_search'?`

Alpine doesn't have a C++20 compatible stdc++?

AlanGriffiths:
> ``
> 
> Alpine doesn't have a C++20 compatible stdc++?

Worth trying https://github.com/MirServer/mir/pull/2301?

AlanGriffiths:
> Worth trying #2301?

Actually, not. This header doesn't exist until g++11 (https://en.cppreference.com/w/cpp/compiler_support)

wmww:
I really don't like having separate code paths that run exclusively on platforms it's less-than convenient to manually test on. If it doesn't work we shouldn't use it on alpine, and if it does work why not use it everywhere?

src/server/compositor/multi_threaded_compositor.cpp
```diff
@@ -230,6 +238,15 @@ class CompositingFunctor
         started_future.get();
     }
 
+    void wait_until_stopped()
+    {
+        stop();
+        if (stopped_future.wait_for(10s) != std::future_status::ready)
+            BOOST_THROW_EXCEPTION(std::runtime_error("Compositor thread failed to stop"));
+
+        stopped_future.get();
```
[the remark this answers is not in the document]
RAOF: Mainly this is habit; while `stopped_future` is never going to contain a value, it *might* contain an *exception*, which would be swallowed if we didn't call `.get()`

src/common/system_executor.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+void just_rethrow_exception()
+{
+    std::rethrow_exception(std::current_exception());
+}
+
+/* We use an atomic void(*)() rather than a std::function to avoid needing to take a mutex
+ * in exception context, as taking a mutex can itself throw an exception!
+ */
+std::atomic<void(*)()> exception_handler = &just_rethrow_exception;
```
AlanGriffiths, line 96: I don't see why we give `just_rethrow_exception()` a name. Surely this is better to inline?
```suggestion
/* We use an atomic void(*)() rather than a std::function to avoid needing to take a mutex
 * in exception context, as taking a mutex can itself throw an exception!
 */
std::atomic<void(*)()> exception_handler{[] { std::rethrow_exception(std::current_exception()); }};
```

AlanGriffiths (approved):
> I really don't like having separate code paths that run exclusively on platforms it's less-than convenient to manually test on. If it doesn't work we shouldn't use it on alpine, and if it does work why not use it everywhere?

This is something we should have adequate automated test coverage on. Concentrate on ensuring that is true, as manual testing should only be needed for emergent behaviours when interacting with (possibly flaky) externals (hardware, OS & apps).

Also, this should be temporary: in about 6-9 months we should be able move to a core22 base and drop platforms with earlier versions of libstdc++ (I expect libstdc++-11 to be in the latest version of all our distros by then). After that, Chris has promised to remove the hack.

src/common/system_executor.cpp
RAOF: Aha! I was having trouble getting the lambda to decay into something the compiler would accept. This is indeed better.

wmww (approved):
Ok, I'm convinced. LGTM

bors r+

PR #2299, opened 2022-01-27
Filter 2 distinct bogus touch event scenarios

Flohack74:
This came out of porting Ubuntu Touch to the Oneplus 5T mobile phone. Some touch input passes driver and libinput but makes not much sense. So we suggest filtering this in Mir:

- Some panels send multiple touch-up events when no corresponding touch-down was received. That will be checked when we are updating the event in the last_seen_properties[] array.
- Some panels wake up and generate (for initial state?) fake events. In the case of Oneplus 5T its an event with 1 touch with the last coordinates that were valid plus 9 others with coordinates at 0,0 - That will be checked as requiring that not more than one touch event can be registered at 0,0 in the same event.


wmww:
Are you running Mir 2.X on the OP5T? I was under the impression Ubuntu Touch was still all Mir 1.X, which would mean you wouldn't benefit from this PR landing in main.

wmww:
Could this potentially be fixed in libinput instead of Mir? Nothing against landing it here if this is the right place, but if it gets fixed in libinput all compositors benefit.

AlanGriffiths:
@wmww FYI @Flohack74 has talked to me about this.

> Are you running Mir 2.X on the OP5T? I was under the impression Ubuntu Touch was still all Mir 1.X, which would mean you wouldn't benefit from this PR landing in main.

There is a corresponding PR against the UBports repository: https://github.com/ubports/mir/pull/38

> Could this potentially be fixed in libinput instead of Mir? Nothing against landing it here if this is the right place, but if it gets fixed in libinput all compositors benefit.

I've also asked this. I'm not sure how easy this is in libinput, as it will already have passed on the multiple touch & release events before the frame event where the detection can happen.

It is something that will not impact most libinput users: they likely ignore a touch down and release within a frame without special logic, I think our logic is wrong as it generates a release when no down was notified.

As an alternative approach, I've wondered about detecting a touch and release within a single frame and not notifying an unchanged state.

AlanGriffiths:
@Flohack74 you might find this useful to resolve the problems on Alpine:

https://discourse.ubuntu.com/t/lxc-set-me-up-set-up-a-container-to-test-mir-on-assorted-oss/26368/

Flohack74:
@wmww well probably it could be fixed in libinput (better in kernel driver but since we are dealing with a ton of broken Android kernels I dont think its sustainable), but a) this is way over my head b) we do not want another forked/upstream package where we have to load our patches into until next LTS (Though maybe Ubuntu Touch switches to STS at some point for reasons like this).
Also as Alan mentioned libinput might be too stateless to grasp whats happening.
@AlanGriffiths I wonder how that unit tests compiled on our CI while it failed on yours. Its basically a mock problem, I will have to look into that because I simply copied code from other unit tests where this should work.

AlanGriffiths:
> @AlanGriffiths I wonder how that unit tests compiled on our CI while it failed on yours. Its basically a mock problem, I will have to look into that because I simply copied code from other unit tests where this should work.

I've not looked into it, but probably a stricter compiler or newer gmock.

Flohack74:
Cleanup done. @AlanGriffiths I dont think I need refactor stage :P

Flohack74:
@wmww after downstream review I added another commit, if you are fine with that I can squash this again into 1 commit only.

wmww (approved):
LGTM, we don't require squashing. I'll leave to Alan to bors.

Flohack74:
BTW latest job failed apparently for a remote repo error

AlanGriffiths:
OK, so here's a specific scenario that concerns me:
```
touch@1 down
frame // we pass on the touch down
// My understanding of weird frame...
for i in 1..10: touch@$i down 
for i in 1..10: touch@$i up
frame // we drop the frame and clear out last_seen_properties
```
After this we have emitted `touch@1 down`, but our state is that it is "up" (and we'll only send a "down").

AlanGriffiths (changes requested):
Simplifying a bit (by ignoring moves), we have two state for each touch: up and down.

We _should_ always send a notification on a state transition, I think this is an important guarantee as it ensures the consumer is notified of the current state. The new code introduces a potential transition from "down" to "up" that doesn't send a notification. (See my previous comment)

As noted in this PR, the existing code isn't guilt free either: if we get an down and up notification in the same frame (or just a spurious "up") we'll send a "up" notification without a preceding "down". I think that less worrying as a stuck "up" state is less problematic than a stuck "down" state.

@Flohack74 I think we need to add a little more logic to check whether we've actually sent a "down" notification and check that when cleaning out `contacts`.

There may be better approaches, but adding an `auto down_notified = false;` member to `ContactData` and setting in `convert_touch_frame()` (unless we delete the entry) should provide the additional context needed.

AlanGriffiths:
@Flohack74 are you intending to progress this further?

Flohack74:
Huh good question. Currently I see myself unable to do so, but its for sure worth fixing. The thing is that there is higher prio items on UBports side as we currently got stuck with our OTA cycle and I need to manage unblocking and release. Maybe thereafter.

graysonguarino:
> OK, so here's a specific scenario that concerns me:
> 
> ```
> touch@1 down
> frame // we pass on the touch down
> // My understanding of weird frame...
> for i in 1..10: touch@$i down 
> for i in 1..10: touch@$i up
> frame // we drop the frame and clear out last_seen_properties
> ```
> 
> After this we have emitted `touch@1 down`, but our state is that it is "up" (and we'll only send a "down").

I believe I should have this addressed in the recent commits.

Flohack74:
@graysonguarino Lovely that you help with that!

AlanGriffiths:
> * Some panels wake up and generate (for initial state?) fake events. In the case of Oneplus 5T its an event with 1 touch with the last coordinates that were valid plus 9 others with coordinates at 0,0 - That will be checked as requiring that not more than one touch event can be registered at 0,0 in the same event.

@Flohack74 is it possible to add udev recordings of the misbehaviour on such a device? Having hard data on what actually goes wrong would be useful in deciding what the right approach is.

My recollection of earlier discussions was that the fake events could potentially be identified as touch+release in the same frame. (Without scrapping the whole frame if only two of the touchpoints are 0,0.)

src/platforms/evdev/libinput_device.cpp
```diff
@@ -358,12 +361,26 @@ mir::EventUPtr mie::LibInputDevice::convert_touch_frame(libinput_event_touch* to
         if (data.action == mir_touch_action_down)
             data.action = mir_touch_action_change;
 
-        if (data.action == mir_touch_action_up)
+        if (data.action == mir_touch_action_up || data.down_notified)
             it = last_seen_properties.erase(it);
         else
             ++it;
+
+        data.down_notified = true;
```
AlanGriffiths, line 369: It is a bad idea to modify `data.down_notified` after (on one path) erasing the object referenced by `data`.

src/platforms/evdev/libinput_device.cpp
```diff
@@ -358,12 +361,26 @@ mir::EventUPtr mie::LibInputDevice::convert_touch_frame(libinput_event_touch* to
         if (data.action == mir_touch_action_down)
             data.action = mir_touch_action_change;
 
-        if (data.action == mir_touch_action_up)
+        if (data.action == mir_touch_action_up || data.down_notified)
             it = last_seen_properties.erase(it);
```
AlanGriffiths, line 365: This (`|| data.down_notified`) looks wrong!

If (on a previous frame) we notified a "down" then we drop the contact data regardless of what the current action is. So if the current action is a `LIBINPUT_EVENT_TOUCH_MOTION`, then our internal state changes to "up" and we never pass on the release.

AlanGriffiths (changes requested):
It is getting hard to reason about the code.

One (preexisting) cause of this is that `convert_touch_frame()` conflates generating the contact information to be notified and updating the `last_seen_properties` of active touchpoints.

Other than that, there are some specific issues mentioned with the proposed code.

Flohack74:
@AlanGriffiths Unfortunately I have the Oneplus 5T which where this happened not any longer under my fingers, but I can ask the original owner to help with that.

tests/unit-tests/input/evdev/test_libinput_device.cpp
```diff
@@ -553,6 +553,110 @@ TEST_F(LibInputDeviceOnMouse, process_event_handles_scroll)
...
+    env.mock_libinput.setup_touch_event(fake_device, LIBINPUT_EVENT_TOUCH_DOWN, event_time_2, slot1, x, y, major, minor,
+                                        pressure, orientation);
+    env.mock_libinput.setup_touch_frame(fake_device, event_time_2);
+
+    // Send single up
+    env.mock_libinput.setup_touch_up_event(fake_device, event_time_3, slot1);
+    env.mock_libinput.setup_touch_frame(fake_device, event_time_3);
+    process_events(touch_screen);
+}
+
```
AlanGriffiths, line 659: This test is wrong. We should probably treat the second down as a move and then output the up.

Instead, the code that was committed with it actually generates two down events and then resets the internal state so that the up is ignored.

AlanGriffiths:
@graysonguarino sorry for hacking on your code!

I was trying to understand why code that looked wrong was passing tests that looked wrong in a different way.

I think there are still tests missing (e.g. two downs in the same frame). And there may still be refactoring needed.

As an aside, I think the matchers could do with improvement: they don't give any meaningful output on failure (e.g. if the "action" is wrong it would be nice to know that, similarly the position or slot). That is probably for another PR though.

include/test/mir/test/event_matchers.h
```diff
@@ -35,6 +35,59 @@
 void PrintTo(MirEvent const& event, std::ostream *os);
```
AlanGriffiths, line 35: @graysonguarino could we please do the event matcher improvements as a separate PR?

Flohack74:
Thx everyone for the effort!

PR #2303, opened 2022-02-05
update copyright and replace backslash with 2 spaces in last 2 lines

realKarthikNair:


AlanGriffiths:
`\copyright` is the Doxygen copyright command

PR #2308, opened 2022-02-10
Add a mir::linearising_executor

RAOF:
This can be used to ensure work is executed in the order it is `spawn()`ed.

### DIFF
--- a/include/common/mir/linearising_executor.h
+++ b/include/common/mir/linearising_executor.h
@@ -1,0 +1,39 @@
+/*
+ * Copyright © 2022 Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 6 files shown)


wmww (approved):
LGTM
bors r+

src/common/linearising_executor.cpp
```diff
@@ -0,0 +1,83 @@
...
+    ~LinearisingAdaptor() noexcept
+    {
+        std::unique_lock<decltype(mutex)> lock{mutex};
+        workqueue.clear();
+        while (!idle)
+        {
+            lock.unlock();
+            std::this_thread::sleep_for(std::chrono::milliseconds{1});
+            lock.lock();
+        }
```
AlanGriffiths, line 45: This looks like a job for `std::condition_variable::wait_for` is there an advantage to spinning?

include/common/mir/linearising_executor.h
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * An Executor that makes the following concurrency guarantees:
+ *
+ * 1.   If linearising_executor.spawn(A) happens-before linearising_executor.spawn(B) then
+ *      A() happens-before B(). That is, the entire execution of A is completed before B is started.
+ * 2.   No work is performed concurrently. For any two calls linearising_executor.spawn(A)
+ *      linearising_executor.spawn(B) either A() happens-before B() or B() happens-before A().
+ * 3.   Work is deferred; linearising_executor.spawn(A) will not block on the execution of A
+ */
+extern Executor& linearising_executor;
```
AlanGriffiths, line 35: Maybe it is my background in mathematics where "linearising" is a method of approximation, but the naming doesn't seem clear. I would likely have gone with "serialising" (but that too has unfortunate associations).

(Not blocking as I don't have a clearly better suggestion)

AlanGriffiths (approved):
WFM

bors r+

Issue #2313, opened 2022-02-14
Double free on exit


AlanGriffiths:
```plain
$ mir_demo_server
```
And close with [X] in the corner, or `Ctrl-Alt-BkSp`...
```
...
double free or corruption (fasttop)
Aborted (core dumped)
```

wmww:
Not present in 2.6.0 (so needs to get fixed before release). Bisect points to 94fa9b4a31c96ce4a5316c88cdefa389412529cb. @RAOF might want to take a look.

RAOF:
Huh. I ran the tests under valgrind and ASan to try and shake such things out! Let's see if I can reproduce locally…

PR #2315, opened 2022-02-15
SystemExecutor: Quiesce before exiting `run_mir()`

RAOF:
Global destruction order is not specified (or specifiable), *and* code
we use may register its own global destructors. In an ideal world we
could ensure the SystemExecutor gets destroyed before any other global,
and hence guarantee that none of its threads access state which may
have already been destroyed. Sadly, attempting that is *at best*
non-portable and fragile, at worst impossible.

Do the next best thing by explicitly waiting in `run_mir()` for
work to finish.

Fixes: #2313 - double free on shutdown.

Mesa EGL registers some thread-local storage and frees it during
global teardown. If this teardown is done before all threads which
access that storage have exited then a double-free will occur when
thread termination triggers deallocation of the TLS object.

### DIFF
--- a/include/common/mir/system_executor.h
+++ b/include/common/mir/system_executor.h
@@ -39,6 +39,10 @@ public:
      */
     static void set_unhandled_exception_handler(void (*handler)());
 
+    /**
+     * Wait for all current work to finish and terminate all worker threads
+     */
+    static void quiesce();
 protected:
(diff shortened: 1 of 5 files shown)


AlanGriffiths (approved):
Looks sane, and works.

bors merge

PR #2322, opened 2022-02-16
Don't overwrite build intermediate

AlanGriffiths:


AlanGriffiths:
Test infrastructure failure:

> Cannot allocate google:ubuntu-20.04-64: cannot allocate or deallocate (!) new Google server google:ubuntu-20.04-64 (feb161434-960546): cannot provision google:ubuntu-20.04-64 (feb161434-960546): quota 'CPUS' 

(Will retry later)

RAOF (approved):
This kinda looks like a job for `mkstemp`, but this isn't an attack vector and just using different names for the different files will do.

bors r+

PR #2323, opened 2022-02-16
Fix 22.04 FTBFS

AlanGriffiths:
* Don't use ld, use ldd;
* Don't check .symbols unless using ld; and
* Working incantation for gmock/gtest headers

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -32,6 +32,11 @@ ifeq ($(filter noopt,$(DEB_BUILD_OPTIONS)),noopt)
 	DEB_BUILD_MAINT_OPTIONS += optimize=-lto
 endif
 
+# Use ldd on jammy due to failing to build with ld
+ifneq ($(filter jammy,$(DEB_DISTRIBUTION)),)
+	COMMON_CONFIGURE_OPTIONS += -DMIR_USE_LD=ldd
+endif
+
 # We can't guarantee non-Ubuntu builds will build against WLCS packages we
 # control, so disable it on non-Ubuntu builds
 ifeq ($(filter Ubuntu,$(DEB_VENDOR)),)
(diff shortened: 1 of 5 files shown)


src/miroil/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ set_target_properties(miroil
 
 find_program(MIR_DPKG_GENSYMBOLS dpkg-gensymbols)
 if (CMAKE_COMPILER_IS_GNUCXX   AND  # clang generates slightly different symbols (but we don't care)
+        MIR_USE_LD EQUAL "ld"      AND  # lld on 22.04 generates slightly different symbols (but we don't care)
```
Saviq, line 64:
```suggestion
    MIR_USE_LD EQUAL "ld"      AND  # lld on 22.04 generates slightly different symbols (but we don't care)
```

debian/rules
```diff
@@ -32,10 +32,9 @@ ifeq ($(filter noopt,$(DEB_BUILD_OPTIONS)),noopt)
 	DEB_BUILD_MAINT_OPTIONS += optimize=-lto
 endif
 
-# Disable LTO on jammy due to failing to build
+# Use ldd on jammy due to failing to build with ld
 ifneq ($(filter jammy,$(DEB_DISTRIBUTION)),)
```
Saviq, line 36:
```suggestion
ifeq ($(DEB_DISTRIBUTION),jammy)
```

Saviq (changes requested):
Couple nits.

debian/rules
AlanGriffiths: Makes sense, but I wanted to only touch the minimum.

src/miroil/CMakeLists.txt
AlanGriffiths: Fixed

Saviq (approved):
Yes.

bors merge

AlanGriffiths:
OK, so `debian/compat` can't contain 9 anymore. Just need to see what we can bump it to, and what changes there are...

AlanGriffiths:
OK, toolchain is beyond workarounds

PR #2327, opened 2022-02-17
Don't check miroil symbols on 32bit platforms (like armhf)

AlanGriffiths:


AlanGriffiths:
I'm just going to land this to see what happens in ppa:mir-team/dev

PR #2328, opened 2022-02-17
Don't check miroil symbols on 32bit platforms (like armhf)

AlanGriffiths:


AlanGriffiths:
As this is only a packaging change I'm landing it

PR #2330, opened 2022-02-17
Do not give menus keyboard focus (again)

### LINKED ISSUE #2324: Opening Kate submenus causes menus to disappear
Repro: 
- Open Kate on Wayland
- Open a few submenus
- Eventually the main menu will disappear

Introduced somewhere between v2.5.0 and v2.6.0, I'll update this issue once I bisect down to a specific commit.

Kate running on XWayland is fine.

wmww:
Fixes #2324, and refactors `AbstractShell::notify_focus_locked()` in an attempt to make future changes easier to understand.

### DIFF
--- a/src/server/shell/abstract_shell.cpp
+++ b/src/server/shell/abstract_shell.cpp
@@ -85,12 +85,13 @@ auto get_non_popup_parent(std::shared_ptr<ms::Surface> surface) -> std::shared_p
     return surface;
 }
 
-auto get_active_surfaces(std::shared_ptr<ms::Surface> surface) -> std::vector<std::shared_ptr<ms::Surface>>
+// Returns a vector comprising the supplied surface (if not null), parent, grandparent, etc. in order
+auto get_ancestry(std::shared_ptr<ms::Surface> surface) -> std::vector<std::shared_ptr<ms::Surface>>
 {
     std::vector<std::shared_ptr<ms::Surface>> result;
-    for (auto item = surface; item; item = item->parent())
+    for (auto item = std::move(surface); item; item = item->parent())
     {
-        result.insert(begin(result), item);
+        result.push_back(item);
     }
     return result;
 }
@@ -435,7 +436,7 @@ void msh::AbstractShell::set_focus_to(
     if (last_requested_focus_surface.lock() != focus_surface)
     {
         last_requested_focus_surface = focus_surface;
-        auto new_active_surfaces = get_active_surfaces(focus_surface);
+        auto new_active_surfaces = get_ancestry(focus_surface);
 
         /// HACK: Grabbing popups (menus, in Mir terminology) should be given keyboard focus according to xdg-shell,
         /// however, giving menus keyboard focus breaks Qt submenus. As of February 2022 Weston and other compositors
--- a/src/server/shell/abstract_shell.cpp
+++ b/src/server/shell/abstract_shell.cpp
@@ -468,20 +468,22 @@ void msh::AbstractShell::notify_active_surfaces(
         auto const found = old_active_surfaces.find(new_active);
         if (found == end(old_active_surfaces))
         {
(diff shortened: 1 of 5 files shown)


src/server/shell/abstract_shell.cpp
```diff
@@ -406,102 +432,114 @@ void msh::AbstractShell::set_focus_to(
...
+    std::unique_lock<std::mutex> const&,
+    std::shared_ptr<ms::Surface> const& new_keyboard_focus_surface,
+    std::vector<std::shared_ptr<ms::Surface>> const& new_active_surfaces)
 {
-    auto const current_focus = notified_focus_surface.lock();
+    std::vector<std::shared_ptr<ms::Surface>> prev_active_surfaces;
+    prev_active_surfaces.reserve(notified_active_surfaces.size());
 
-    if (current_focus != new_focus_surface)
+    for (auto const& current_active_weak: notified_active_surfaces)
```
RAOF, line 461: I think this might be faster and clearer if we kept `notified_active_surfaces` and `new_active_surfaces` sorted, computed the set of surfaces that are in `notified_active_surfaces` but not `new_active_surfaces` in one pass rather than doing a bunch of O(n²) passes over them, and then doing the notifications at the end.

src/server/shell/abstract_shell.cpp
```diff
@@ -406,102 +432,114 @@ void msh::AbstractShell::set_focus_to(
...
-                }
-                else if (active == current_focus)
+                current_active->set_focus_state(mir_window_focus_state_unfocused);
+
+                // When a menu loses focus we should close and unmap it
+                if (current_active->type() == mir_window_type_menu || current_active->type() == mir_window_type_gloss)
                 {
-                    active->set_focus_state(mir_window_focus_state_active);
+                    current_active->request_client_surface_close();
+                    current_active->hide();
```
RAOF, line 475: Is this going to do the right thing? We need to not only close the popup, but also any parent popups it has, and in child-first order, whereas I *think* you're iterating over this in parent-first order?

RAOF (changes requested):
Naming nit, but also it's not clear that we're notifying nested menus in the correct (child-first) order.

I *think* my suggestions might make the code a bit clearer, too?

AlanGriffiths:
Empirically this does seem to fix the problem with kate menus.

I agree with @RAOF that some of the code isn't clearly doing the right thing. Will try reworking that and then do some further testing with other client stacks.

AlanGriffiths:
@wmww I've taken the liberty of pushing some updates. I think they cover @RAOF's concerns without breaking stuff. But please try to prove me wrong.

AlanGriffiths (approved):
WFM (now)

src/server/shell/abstract_shell.cpp
```diff
@@ -85,12 +85,13 @@ auto get_non_popup_parent(std::shared_ptr<ms::Surface> surface) -> std::shared_p
...
-auto get_active_surfaces(std::shared_ptr<ms::Surface> surface) -> std::vector<std::shared_ptr<ms::Surface>>
+// Returns a vector comprising the supplied surface (if not null), parent, grandparent, etc. in order
+auto get_ancestry(std::shared_ptr<ms::Surface> surface) -> std::vector<std::shared_ptr<ms::Surface>>
 {
     std::vector<std::shared_ptr<ms::Surface>> result;
-    for (auto item = surface; item; item = item->parent())
+    for (auto item = std::move(surface); item; item = item->parent())
     {
-        result.insert(begin(result), item);
+        result.push_back(item);
```
wmww, line 95: Oops, good catch

AlanGriffiths (approved):
Ok

Saviq (approved):
LGTM!

bors merge

AlanGriffiths:
Failed twice with "Cannot find cached LXD image for lxd:alpine-edge" time for brute force...

PR #2333, opened 2022-02-21
Allow --app-env-amend to be supplied multiple times

AlanGriffiths:
This makes it easier to apply both packaging related options (e.g. variables used by snap) and user options.


wmww (approved):
LGTM
bors r+

PR #2337, opened 2022-02-22
Don't enable zwp_text_input_v2 by default

AlanGriffiths:
The zwp_text_input_v2 implementation needs further work.

Fixes https://github.com/MirServer/mir/issues/2334

### DIFF
--- a/src/server/frontend_wayland/wayland_default_configuration.cpp
+++ b/src/server/frontend_wayland/wayland_default_configuration.cpp
@@ -237,7 +237,6 @@ auto mf::get_standard_extensions() -> std::vector<std::string>
         mw::XdgWmBase::interface_name,
         mw::XdgShellV6::interface_name,
         mw::XdgOutputManagerV1::interface_name,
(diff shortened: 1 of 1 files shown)


wmww:
Does our implementation need work? It looks to me like the bugs are in Qt/QTerminal, and there's not much we can do to work around them besides not exposing this protocol.

RAOF:
Yeah, I wouldn't necessarily blame our implementation and I *definitely* wouldn't blame our `zwp_text_input_v2` implementation. Disabling the protocol will avoid the crash, but only because the crash is qterminal indexing into a null array before making a `text_input` call.

RAOF:
More investigation shows that qterminal crashes in the same way under kwin; I don't think we should disable `text_input` by default to work around this.

AlanGriffiths:
> More investigation shows that qterminal crashes in the same way under kwin; I don't think we should disable `text_input` by default to work around this.

Agreed

PR #2339, opened 2022-02-22
Revert "Merge #2245"

AlanGriffiths:
Fixes #2338

PR #2340, opened 2022-02-22
Further simplify and correct keyboard focus setting

### LINKED ISSUE #2338: [2.7~rc6] fatal error
With ubuntu-frame `beta` (built with [2.7.0~rc60]) and qterminal installed.

From a terminal:

```
export WAYLAND_DISPLAY=wayland-99
ubuntu-frame&
QT_QPA_PLATFORM=wayland qterminal
```

Now, focus on the "Mir-on-X" window and press `Ctrl-Shift-W` and when confirm exit appears press "n".

Finally press a letter.

Expect: the letter appears in the terminal tab
Actual: Mir crashes with:
```
Mir fatal error: Attempt to send keyboard event to wl_surface@8 even though it was not given keyboard focus
```

wmww:
This PR corrects a behavior where Mir would fail to set the requested surface as focused if it was already active. Instead of adding additional error-prone logic I simply gave up on our previous strategy of only notifying surfaces when necessary.

`BasicSurface` already returns immediately if `focus_state` is set to the same value multiple times. The only additional performance cost is that of locking all surfaces that were active both before and after, which will never be never be the majority of the cost of changing focus because we already lock all surfaces in the new focus tree once anyway to get their parent.

Fixes #2338

### DIFF
--- a/tests/unit-tests/scene/test_abstract_shell.cpp
+++ b/tests/unit-tests/scene/test_abstract_shell.cpp
@@ -919,9 +919,20 @@ TEST_F(AbstractShell, makes_parent_active_when_switching_to_child)
 
     msh::FocusController& focus_controller = shell;
     focus_controller.set_focus_to(surface_parent->session().lock(), surface_parent);
+    /* window         | expected focus state
+     * -----------------------------------------------
+     * surface_parent | mir_window_focus_state_focused 
+     * surface_child  | <don't care>
+    */
 
     EXPECT_CALL(mock_surface, set_focus_state(mir_window_focus_state_active));
     focus_controller.set_focus_to(surface_child->session().lock(), surface_child);
+    /* window         | expected focus state
+     * -----------------------------------------------
+     * surface_parent | mir_window_focus_state_active
+     * surface_child  | mir_window_focus_state_focused
+    */
+
 }
 
 TEST_F(AbstractShell, makes_parent_focused_when_switching_back_from_child)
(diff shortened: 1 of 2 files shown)


wmww:
Alternative to #2339

tests/unit-tests/scene/test_abstract_shell.cpp
```diff
@@ -904,10 +904,42 @@ TEST_F(AbstractShell, does_not_deactivate_parent_when_switching_children)
...
+    NiceMock<mtd::MockSurface> mock_surface_child;
+    auto const surface_child = create_surface(mock_surface_child);
+    ON_CALL(mock_surface_child, parent())
+        .WillByDefault(Return(surface_parent));
+
+    msh::FocusController& focus_controller = shell;
+    focus_controller.set_focus_to(surface_parent->session().lock(), surface_parent);
+
+    EXPECT_CALL(mock_surface, set_focus_state(mir_window_focus_state_active));
+    focus_controller.set_focus_to(surface_child->session().lock(), surface_child);
```
RAOF, line 924: This looks like we're testing that the parent surface gets sent a *duplicate* `mir_window_focus_state_active` notification? Am I reading this test correctly?

If so, I don't think that's a useful thing to test; it's testing a quirk of the current implementation rather than a behaviour that we want to preserve.

RAOF (commented):
Can confirm that this fixes #2338 (weirdly, I can *only* reproduce that on ubuntu-frame, not miral-shell or miral-kiosk)

Have a question about the tests, though

RAOF (approved):
I've turned my question about the tests into a series of comments in the test.

I think this is good to go, after checking those comments are sensible.

bors delegate=wmww

wmww:
LGTM
bors r+

PR #2342, opened 2022-02-23
wl_pointer: do not send events when not compatible with version

### LINKED ISSUE #2341: We send pointer events without checking supported version
Repro:
- open glmark-es2-wayland
- scroll

It crashes, complaining about a wl_pointer opcode. Present in 2.6, I'm not sure when it was introduced but I believe it was somewhat recently.

wmww:
Fixes #2341


RAOF (approved):
This is a simple and clearly-correct fix. We can think about systematic fixes later.

bors r+

Issue #2331, opened 2022-02-17
Keyboard focus/keyboard event inconsistancy


wmww:
#2245 we get
```
Attempt to send keyboard event to wl_surface@# even though it was not given keyboard focus
```
If Mir sends the Wayland frontend keyboard events for a surface that does not have keyboard focus. The problem is there is no synchronization between switching keyboard focus and sending input events, so events sent to the wrong surface could intermittently slip through (this has not been observed in the wild AFAIK). We should either restore the old behavior, or somehow ensure events arrive correctly.

wmww:
This is happening when opening weston-terminal. Repro:
- From any terminal, run `weston-terminal` by pressing enter
- Occasionally, Mir will crash

wmww:
Note that the #2330 may have introduced this in practice or made it more common since it rearranged timing and order of the relevant calls, but the underlying race condition was always present.

AlanGriffiths:
> Fixed by #2339

I disagree. Reporting a problem quietly (as a log message) instead of loudly (fatal error) only reduces the severity of the manifestation.

PR #2345, opened 2022-02-24
Restore the "fatal_error" when focus is screwed

AlanGriffiths:
While we're not ready for release with this, it has been useful identifying and fixing genuine problems in development.


wmww (changes requested):
I don't think we should restore behavior that is known to cause Mir to crash. This would restore #2331. I do think we should refactor the input system to make this question irrelevant, which I plan to do once I have the available time.

RAOF:
Wait - did GH eat my comment here? Urgh.

Anyway the summary of it is, (a) I agree that we should not have another release with this `fatal_error` and (b) we should restore the fatal error during development. There are a bunch of bugs I don't think we'd have noticed during the 2.7 release process without this being fatal.

So I'm +1 on this.

wmww (approved):
Ok
bors r+

PR #2349, opened 2022-03-01
[x11-platform] Make window title a configuration option

### LINKED ISSUE #2219: Rename "Mir on X" window header
As we promote the Ubuntu Frame simulation feature for developers to test their graphic applications before embedding them - we need to change the name on the window header;

"Mir on X" is not a good name.
Making it configurable in Mir, so that Ubuntu Frame can set it to something like "Ubuntu Frame simulator", is a better option.

Feature request:
The current Mir-on-Wayland support doesn't support a "windowed mode". That's not useful for the current scenario.

graysonguarino:
Fixes #2219 by making the initialized title an argument in X11Window, and if the argument is an empty string, sets it to the previous default, "Mir on X".

### DIFF
--- a/src/platforms/x11/graphics/display.cpp
+++ b/src/platforms/x11/graphics/display.cpp
@@ -79,11 +79,12 @@ private:
 }
 
 mgx::X11Window::X11Window(mx::X11Resources* x11_resources,
+                          std::string* title,
                           EGLDisplay egl_dpy,
                           geom::Size const size,
-                          EGLConfig const egl_cfg,
-                          std::string title)
-    : x11_resources{x11_resources}
+                          EGLConfig const egl_cfg)
+    : x11_resources{x11_resources},
+      title{title}
 {
     auto const conn = x11_resources->conn.get();
 
@@ -118,12 +119,12 @@ mgx::X11Window::X11Window(mx::X11Resources* x11_resources,
         char buffer[128] = { '\0' };
         if (gethostname(buffer, sizeof buffer - 1) == 0)
         {
-            title += " - ";
-            title += buffer;
+            *title += " - ";
+            *title += buffer;
         }
     }
 
-    conn->change_property(win, x11_resources->_NET_WM_NAME, x11_resources->UTF8_STRING, 8, title.size(), title.c_str());
+    conn->change_property(win, x11_resources->_NET_WM_NAME, x11_resources->UTF8_STRING, 8, title->size(), title->c_str());
 
     conn->map_window(win);
 }
--- a/src/platforms/x11/graphics/display.cpp
+++ b/src/platforms/x11/graphics/display.cpp
@@ -144,6 +144,7 @@ unsigned long mgx::X11Window::red_mask() const
 }
 
 mgx::Display::Display(std::shared_ptr<mir::X::X11Resources> const& x11_resources,
+                      std::shared_ptr<std::string> const& title,
                       std::vector<X11OutputConfig> const& requested_sizes,
                       std::shared_ptr<DisplayConfigurationPolicy> const& initial_conf_policy,
                       std::shared_ptr<GLConfig> const& gl_config,
@@ -162,10 +163,10 @@ mgx::Display::Display(std::shared_ptr<mir::X::X11Resources> const& x11_resources
         auto actual_size = requested_size.size;
         auto window = std::make_unique<X11Window>(
             x11_resources.get(),
+            title.get(),
             shared_egl.display(),
             actual_size,
-            shared_egl.config(),
-            "Mir On X");
+            shared_egl.config());
         auto red_mask = window->red_mask();
         auto pf = (red_mask == 0xFF0000 ?
             mir_pixel_format_argb_8888 :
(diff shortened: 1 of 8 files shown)


AlanGriffiths (changes requested):
> This solves issue https://github.com/MirServer/mir/issues/2219

No it doesn't. (Yet.)

It is a good step along the way, but to solve the issue, there also needs to be a way to configure the title at runtime. We work with feature branches rather than continuously integrating because we're working across timezones and reviewing asynchronously (and not pairing).

Can you:
1. add a further commit to this PR that sets the title based on a configuration option?
2. Change the wording above to "Fixes: 2219" (that way github links the issue to the merge)

src/platforms/x11/graphics/display.cpp
```diff
@@ -111,8 +112,9 @@ mgx::X11Window::X11Window(mx::X11Resources* x11_resources,
     // Enable the WM_DELETE_WINDOW protocol for the window (causes a client message to be sent when window is closed)
     conn->change_property(win, x11_resources->WM_PROTOCOLS, XCB_ATOM_ATOM, 32, 1, &x11_resources->WM_DELETE_WINDOW);
 
-    // Set the window title
-    std::string title{"Mir on X"};
+    // Set the window title if null
+    if (title.empty())
+        title = "Mir on X";
```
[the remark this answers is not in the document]
graysonguarino: An empty title does not break anything. This will be removed in the next pull request.

AlanGriffiths (changes requested):
All the issues I see arise from trying to treat "Title" as reference type (referring to a particular, shared string).

"Title" is a conceptually a value and best treated as such (the code will be simpler, more efficient and avoid the bug noted).

tests/unit-tests/platforms/x11/test_display.cpp
```diff
@@ -103,7 +103,7 @@ class X11DisplayTest : public ::testing::Test
     {
         return std::make_shared<mgx::Display>(
                    mt::fake_shared(x11_resources),
-                   std::make_shared<std::string>("Mir on X"),
+                   std::string("Mir on X"),
```
AlanGriffiths, line 106: Since its introduction in C++11 we tend to use uniform initialization syntax (braces for initialization) everywhere.

But here there is no need as there is a converting constructor.

```suggestion
                   "Mir on X",
```

src/platforms/x11/graphics/display.cpp
```diff
@@ -79,10 +79,12 @@ class XGLContext : public mir::renderer::gl::Context
 }
 
 mgx::X11Window::X11Window(mx::X11Resources* x11_resources,
+                          std::string title,
                           EGLDisplay egl_dpy,
                           geom::Size const size,
                           EGLConfig const egl_cfg)
-    : x11_resources{x11_resources}
+    : x11_resources{x11_resources},
+      title{title}
```
AlanGriffiths, line 87:
```suggestion
    : x11_resources{x11_resources}
```

src/platforms/x11/graphics/platform.cpp
```diff
@@ -108,9 +108,11 @@ auto mgx::Platform::parse_output_sizes(std::string output_sizes) -> std::vector<
 }
 
 mgx::Platform::Platform(std::shared_ptr<mir::X::X11Resources> const& x11_resources,
+                        std::string const title,
                         std::vector<X11OutputConfig> output_sizes,
                         std::shared_ptr<mg::DisplayReport> const& report)
     : x11_resources{x11_resources},
+      title{title},
```
AlanGriffiths, line 115: We can move (not copy) title as the parameter is only used to initialize the data member.

```suggestion
                        std::string title,
                        std::vector<X11OutputConfig> output_sizes,
                        std::shared_ptr<mg::DisplayReport> const& report)
    : x11_resources{x11_resources},
      title{std::move(title)},
```

AlanGriffiths (changes requested):
Some very minor tweaks for good style, it seems to work.

src/platforms/x11/graphics/display.h
```diff
@@ -63,6 +64,7 @@ class X11Window
 
 private:
     mir::X::X11Resources* const x11_resources;
+    std::string const title;
```
AlanGriffiths, line 67: This is unused,

```suggestion
```

tests/unit-tests/platforms/x11/test_platform.cpp
```diff
@@ -69,6 +69,7 @@ class X11GraphicsPlatformTest : public ::testing::Test
     {
         return std::make_shared<mg::X::Platform>(
             std::make_shared<mtd::MockX11Resources>(),
+            std::string("Mir on X"),
```
AlanGriffiths, line 72: Since its introduction in C++11 we tend to use uniform initialization syntax (braces for initialization) everywhere.

But here there is no need as there is a converting constructor.

```suggestion
            "Mir on X",
```

tests/unit-tests/platforms/x11/test_platform.cpp
```diff
@@ -88,6 +89,7 @@ TEST_F(X11GraphicsPlatformTest, failure_to_open_x11_display_results_in_an_error)
         {
             std::make_shared<mg::X::Platform>(
                 nullptr,
+                std::string("Mir on X"),
```
AlanGriffiths, line 92: Since its introduction in C++11 we tend to use uniform initialization syntax (braces for initialization) everywhere.

But here there is no need as there is a converting constructor.

```suggestion
            "Mir on X",
```

AlanGriffiths (changes requested):
Did I forget these last time?

AlanGriffiths (approved):
LGTM

bors merge

PR #2357, opened 2022-03-08
cmake: Fix the build with g++-12

RAOF:
g++-12 is not a fan of the GoogleMock headers:
```
/usr/src/googletest/googlemock/include/gmock/internal/gmock-internal-utils.h:296:10: error: infinite recursion detected [-Werror=infinite-recursion]
  296 | inline T Invalid() {
      |          ^~~~~~~
/usr/src/googletest/googlemock/include/gmock/internal/gmock-internal-utils.h:301:20: note: recursive call
  301 |   return Invalid<T>();
      |          ~~~~~~~~~~^~
```

This is not *actually* infinite-recursion, because `Invalid()` calls `Abort()` before reaching the return. `Abort()` will *eventually* end up calling a `[[noreturn]]` function, but conditionally, and it seems g++-12's program flow analysis doesn't trace down far enough to notice.


AlanGriffiths (approved):
bors merge

PR #2359, opened 2022-03-09
Refactor Wayland keyboard input

### LINKED ISSUE #2331: Keyboard focus/keyboard event inconsistancy

wmww:
#2331, and related issues caused us some trouble last release. This PR refactors the way keyboard input is processed internally to make it better match up with how events need to be delivered to Wayland clients.

A new observer is introduced through which both keyboard focus and keyboard events flow. Assuming the executor used preserves ordering of events, this prevents race conditions from causing input to go to non-focused surfaces. It also removes the need for the slightly complex system we previously had where we watched the state of all surfaces to determine the currently focused surface.

Fixes #2331.

~~Marked as draft initially because I still have tests to write.~~ done

### DIFF
--- a/tests/unit-tests/shell/test_decoration_basic_decoration.cpp
+++ b/tests/unit-tests/shell/test_decoration_basic_decoration.cpp
@@ -79,6 +79,7 @@ struct MockSurface
         : ms::BasicSurface{
               session,
               {},
+              {},
               {{},{}},
               mir_pointer_unconfined,
               { { std::make_shared<testing::NiceMock<mtd::MockBufferStream>>(), {0, 0}, {} } },
--- a/tests/unit-tests/shell/test_decoration_basic_decoration.cpp
+++ b/tests/unit-tests/shell/test_decoration_basic_decoration.cpp
@@ -41,6 +41,8 @@ namespace mc = mir::compositor;
 namespace msh = mir::shell;
 namespace geom = mir::geometry;
 namespace mev = mir::events;
+namespace mf = mir::frontend;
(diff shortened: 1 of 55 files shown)


tests/unit-tests/shell/test_decoration_basic_decoration.cpp
```diff
@@ -133,8 +135,9 @@ struct MockShell
         uint64_t,
         MirResizeEdge));
 
-    MOCK_METHOD3(create_surface, std::shared_ptr<ms::Surface>(
+    MOCK_METHOD4(create_surface, std::shared_ptr<ms::Surface>(
```
AlanGriffiths, line 138: Please, let's use `MOCK_METHOD` for new code.

AlanGriffiths (approved):
Looks sensible, passes automated tests. No problems in exploratory testing.

And early in the release cycle is a good time to land this rework.

bors merge

tests/unit-tests/shell/test_decoration_basic_decoration.cpp
wmww: Oh, cool, I did not know about this.

Issue #2364, opened 2022-03-22
Global signal handling is broken


RAOF:
There are two problems in the signal handling code we do in `src/server/run_mir.cpp`.

Firstly: [here](https://github.com/MirServer/mir/blob/637984bb07ec9b25286630fe70bc037364e586b5/src/server/run_mir.cpp#L107) we have the problem that `signal` has unspecified behaviour in multithreaded programs. Mostly theoretical, as it seems to work, mostly?

Secondly, we have https://github.com/MirServer/wlcs/issues/30 which happens because `libwayland` installs a `SIGBUS` handler the *first* time that `wl_shm_begin_access()` is called in a process. When we have multiple calls to `run_mir()` in a process, this results in the old `SIGBUS` handler being restored at the end of the first `run_mir()`, then replaced by Mir's `SIGBUS` handler (that performs fatal error cleanup, then re-raises), but is *not* then replaced by `libwayland's` convert-SIGBUS-in-wl_shm-into-protocol-error handler, as libwayland thinks it has already installed it.

I've got a branch mostly fixing this, should be reviewable soon.

AlanGriffiths:
> Firstly: [here](https://github.com/MirServer/mir/blob/637984bb07ec9b25286630fe70bc037364e586b5/src/server/run_mir.cpp#L107) we have the problem that `signal` has unspecified behaviour in multithreaded programs.

Mostly the program is single threaded at that point. But yeah.

PR #2366, opened 2022-03-24
mir::Server: Fix signal handling

RAOF:
Fixes: #2364, MirServer/wlcs#30

### DIFF
--- a/src/server/run_mir.cpp
+++ b/src/server/run_mir.cpp
@@ -126,6 +126,25 @@ private:
     std::array<std::atomic<struct sigaction*>, intercepted.size()> handlers = { nullptr };
 } old_handlers;
 
+auto signum_to_string(int sig) -> std::string
+{
+    switch(sig)
+    {
+        case SIGQUIT:
+            return "SIGQUIT";
+        case SIGABRT:
+            return "SIGABRT";
+        case SIGFPE:
+            return "SIGFPE";
+        case SIGSEGV:
+            return "SIGSEGV";
+        case SIGBUS:
+            return "SIGBUS";
+        default:
+            return std::string{"(Unknown signal: "} + std::to_string(sig) + ")";
+    }
+}
+
 extern "C" [[noreturn]] void fatal_signal_cleanup(int sig, siginfo_t* info, void* ucontext)
 {
     perform_emergency_cleanup();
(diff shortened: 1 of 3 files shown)


src/server/run_mir.cpp
```diff
@@ -47,16 +48,126 @@ extern "C" void perform_emergency_cleanup()
...
+         * Re-raising the signal will destroy this context, so call the handler directly.
+         */
+        (*old_handler->sa_sigaction)(sig, info, ucontext);
+        /* We don't *expect* to get here - fatal signal handlers will *generally* re-raise the signal,
+         * and so don't return.
+         *
+         * However, we've just performed emergency cleanup, so if the handler was trying to fix things up
+         * and continue we unfortunately can't let them.
+         */
+        std::abort();
```
AlanGriffiths, line 147:
```suggest
        mir::fatal_error_abort("Unsupported attempt to continue after a fatal signal: "s + signum_to_string(sig) );
```

src/server/run_mir.cpp
```diff
@@ -47,16 +48,126 @@ extern "C" void perform_emergency_cleanup()
...
+    case SIGABRT:
+        return "SIGABRT";
+    case SIGFPE:
+        return "SIGFPE";
+    case SIGSEGV:
+        return "SIGSEGV";
+    case SIGBUS:
+        return "SIGBUS";
+    default:
+        return "(Unknown signal)";
```
AlanGriffiths, line 172: Maybe give the signal number in this case?

src/server/run_mir.cpp
```diff
@@ -110,7 +221,20 @@ void mir::run_mir(
...
+
+                    if (sigaction(sig, &sig_handler_desc, old_handlers.initialiser_for_signal(sig)))
+                    {
+                        using namespace std::string_literals;
+                        BOOST_THROW_EXCEPTION((
+                            std::system_error{
+                                errno,
+                                std::system_category(),
+                                "Failed to install signal handler for "s + signum_to_string(sig)
+                            }));
```
AlanGriffiths, line 239: Is throwing an exception actually useful?

src/server/run_mir.cpp
```diff
@@ -110,7 +221,20 @@ void mir::run_mir(
             {
                 for (auto sig : intercepted)
                 {
-                    old_handler[sig] = signal(sig, fatal_signal_cleanup);
+                    struct sigaction sig_handler_desc;
+                    sig_handler_desc.sa_flags = SA_SIGINFO;
+                    sig_handler_desc.sa_sigaction = &fatal_signal_cleanup;
+
+                    if (sigaction(sig, &sig_handler_desc, old_handlers.initialiser_for_signal(sig)))
```
AlanGriffiths, line 231: This lifetime of temporary trick with `AtomicOutPtr` is quite cool.

AlanGriffiths (approved):
This does what it claims. It is unfortunate we need to special case SIGBUS because of non-local knowledge (of libwayland internals).

I think the error handling could be simplified, but not blocking on that.

bors delegate

src/server/run_mir.cpp
RAOF: Yeah! It was really interesting seeing the summary of P1132R8 scroll across my twitter just before I had a need for such a gadget!

src/server/run_mir.cpp
RAOF: As opposed to simply `fatal_error_abort()`? Unlike the usages above, here we're in a context where there's going to be a catch handler further up the stack, and throwing gets us a nicer error and lets consuming code do something before dying.

If the alternative is not checking the return of `sigaction()` then that's maybe ok; if I understand it correctly that call can only fail in cases where we've made a programmer error. On the other hand, I *like* getting told when I've made a programming error ;)

RAOF:
bors r=AlanGriffiths

AlanGriffiths (approved):
bors r+

AlanGriffiths:
bors retry

PR #2369, opened 2022-03-29
Add NonBlockingExecutor type

### LINKED ISSUE #2310: Nonblocking executor interface
Some APIs may want to invoke an executor under lock, or otherwise need to guarantee the executor will not do the work immediately. We should have a subclass of `Executor`, that is only implemented by executors that never execute immediately.

See also #2309.

wmww:
Closes #2310

### DIFF
--- a/include/common/mir/executor.h
+++ b/include/common/mir/executor.h
@@ -52,6 +52,13 @@ protected:
     virtual ~Executor() = default;
 };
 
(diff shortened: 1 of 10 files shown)


RAOF (approved):
Matches my understanding of what we talked about, and can provide useful context in APIs.

bors r+

PR #2372, opened 2022-03-31
Prevent deadlocks in ObserverMultiplexer

wmww:
This PR adds three multiplexer tests (all of which previously deadlocked) and fixes all of them.

The core issue was that spawning an observation required locking the same mutex that remained locked while observations were being processed. The use of `recursive_mutex` mitigated that if everything was on the same thread, but if the receiving-observation thread was blocked on some other thread, which was blocked on sending an observation there was trouble.

This PR uses separate mutexes for spawning and running observations. Both are locked when an observer is unregistered, but unregister does not lock them simultaneously.

This fixes some of the deadlocks I've encountered in #2371, and assuming it's correct is an overall improvement.

### DIFF
--- a/src/include/server/mir/observer_multiplexer.h
+++ b/src/include/server/mir/observer_multiplexer.h
@@ -83,11 +83,11 @@ private:
     public:
         explicit WeakObserver(std::weak_ptr<Observer> observer, Executor& executor)
             : executor{&executor},
-              run_observer{observer}
+              observer{observer}
         {
         }
 
-        /// An observer with run_mutex and run_observer locked. Used to make run observations.
+        /// A handle that excludes attempts to unregister or destroy the Observer
         class LockedObserver
         {
         public:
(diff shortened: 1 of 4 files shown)


RAOF (approved):
I think the double-locking and duplicate `weak_ptr` was unnecessarily complicated, and I think my modifications perform the same function.

This does means that we'll need @AlanGriffiths or @graysonguarino to review ;)

AlanGriffiths:
... but re-running the job passed. :confused:

AlanGriffiths (approved):
LGTM, but now I've added a commit

AlanGriffiths (approved):
Dropping my wrong-headed commit

bors r+

wmww:
My version had one additional guarantee (that wasn't well documented): The executor wont be used after `unregister_interest()` returns (even if the observer is still alive), By dropping the 2nd lock you lose that guarantee.

RAOF:
Is that a guarantee we need to provide? We've got a set of executors with essentially `'static` lifetime available, and for a custom executor it's not *terribly* difficult to ensure the executor outlives the observer.

AlanGriffiths:
> it's not terribly difficult to ensure the executor outlives the observer.

Any teardown code _doesn't_ control the lifetime of the observer - because ownership is by `shared_ptr<>`.

PR #2374, opened 2022-04-01
ExplicitExecutor: repeat execute until queue is drained

wmww:
Makes some tests work in my better surface observer branch, and seems to be a general improvement to me.

### DIFF
--- a/tests/include/mir/test/doubles/explicit_executor.h
+++ b/tests/include/mir/test/doubles/explicit_executor.h
@@ -54,13 +54,20 @@ public:
 
     void execute()
     {
-        std::unique_lock lock{mutex};
-        auto const items = std::move(work_items);
-        work_items.clear();
-        lock.unlock();
-        for (auto const& work : items)
+        while (true)
         {
-            work();
+            std::unique_lock lock{mutex};
+            auto const items = std::move(work_items);
+            work_items.clear();
+            lock.unlock();
+            if (items.empty())
+            {
+                break;
+            }
+            for (auto const& work : items)
+            {
+                work();
+            }
         }
     }
 
--- a/tests/include/mir/test/doubles/explicit_executor.h
+++ b/tests/include/mir/test/doubles/explicit_executor.h
@@ -54,21 +54,19 @@ public:
 
     void execute()
     {
-        while (true)
+        decltype(work_items) drained_items;
+        do
         {
-            std::unique_lock lock{mutex};
-            auto const items = std::move(work_items);
-            work_items.clear();
-            lock.unlock();
-            if (items.empty())
+            drained_items.clear();
             {
-                break;
+                std::lock_guard lock{mutex};
+                swap(drained_items, work_items);
             }
-            for (auto const& work : items)
+            for (auto const& work : drained_items)
             {
                 work();
             }
-        }
+        } while (!drained_items.empty());
     }
 
 private:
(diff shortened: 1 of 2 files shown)


RAOF (approved):
This could, perhaps, be clarified if rewritten as a `do {} while(!items.empty())` loop, but I'm not going to block on that.

bors delegate=wmww

tests/include/mir/test/doubles/explicit_executor.h
```diff
@@ -54,13 +54,20 @@ class ExplicitExecutor
 
     void execute()
     {
-        std::unique_lock<std::mutex> lock{mutex};
-        auto const items = std::move(work_items);
-        work_items.clear();
-        lock.unlock();
-        for (auto const& work : items)
+        while (true)
```
AlanGriffiths, line 57:
```suggestion
        auto drain_work_items = [this]
        {
            decltype(work_items) drained_items;
            std::lock_guard<std::mutex> lock{mutex};
            swap(drained_items, work_items);
            return drained_items;
        };
```

tests/include/mir/test/doubles/explicit_executor.h
```diff
@@ -54,13 +54,20 @@ class ExplicitExecutor
...
         {
-            work();
+            std::unique_lock<std::mutex> lock{mutex};
+            auto const items = std::move(work_items);
+            work_items.clear();
+            lock.unlock();
+            if (items.empty())
+            {
+                break;
+            }
```
AlanGriffiths, line 66:
```suggestion
        for (auto items = drain_work_items(); !items.empty(); items = drain_work_items())
        {
```

AlanGriffiths (approved):
I guess we all have ideas on the tidiest way to write this.

wmww:
Hope we can all agree on that control flow. Also the software cursor tests where wrong.

tests/unit-tests/graphics/test_software_cursor.cpp
```diff
@@ -194,9 +194,9 @@ TEST_F(SoftwareCursor, tolerates_being_hidden_while_being_shown)
     EXPECT_CALL(mock_input_scene, add_input_visualization(_))
         .WillOnce(Invoke([&](auto)
             {
-                cursor.hide(); // should do nothing
+                cursor.hide();
             }));
-    EXPECT_CALL(mock_input_scene, remove_input_visualization(_)).Times(0);
+    EXPECT_CALL(mock_input_scene, remove_input_visualization(_));
```
AlanGriffiths, line 199: > Also the software cursor tests where wrong.

Agreed, they were wrong. But I think they are still (differently) wrong.

IMO they are testing that "something sensible" happens `hide()` is called during `show()` or _vice versa_.

They used to test that `hide()` has no effect (after `executor.execute()`), they now test that it has effect.

```suggestion
    EXPECT_CALL(mock_input_scene, remove_input_visualization(_)).AnyNumber();
```

I don't think _these_ tests should be mandating whether `hide()` has effect here.

AlanGriffiths (approved):
bors merge

wmww:
`MultiThreadedCompositor.can_schedule_from_display_observer_when_adding_display` failed with:
```
2022-04-20T16:57:50.7745311Z qemu: uncaught target signal 11 (Segmentation fault) - core dumped
2022-04-20T16:57:50.7745592Z /<<PKGBUILDDIR>>/tools/detect_fd_leaks.bash: line 84: 55427 Segmentation fault      (core dumped) $@ 2>&1
2022-04-20T16:57:50.7745768Z      55428 Done                    | detect_fd_leaks
```

hmm

bors r=AlanGriffiths

PR #2376, opened 2022-04-04
ObserverMultiplexer: guarentee executor is not used after unregister_interest()

wmww:
Fixes problem I pointed out in #2372 using a 2nd lock

### DIFF
--- a/src/include/server/mir/observer_multiplexer.h
+++ b/src/include/server/mir/observer_multiplexer.h
@@ -82,8 +82,8 @@ private:
     {
     public:
         explicit WeakObserver(std::weak_ptr<Observer> observer, Executor& executor)
-            : executor{&executor},
-              observer{observer}
+            : observer{observer},
+              executor{&executor}
         {
         }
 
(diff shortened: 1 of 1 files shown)


RAOF (changes requested):
I'm not sure what problem this is actually solving. *Do* we have code where it is not easy to guarantee the executor outlives the observer (such as by holding shared ownership of the executor *in* the observer)?

If we do have such code, then I think this is the wrong solution. Rather than preventing `unregister_interest` from proceeding while code wants to use the executor we should simply drop the “calling code must somehow ensure the executor outlives the observer” API requirement and just take shared (or weak) ownership of the executor.

AlanGriffiths:
> I'm not sure what problem this is actually solving.

I've not been following closely. But the guarantee a user of these APIs needs is that once `unregister_interest()` returns then there will be no further callbacks. The rest is implementation detail. (That is: given a guarantee there is no outstanding nor new work, the lifetimes of executor and observer don't matter.)

wmww:
If a user registers an observer and an executor, then unregisters the observer, I was thinking it should be able to immediately destroy the executor. As the code currently is, a currently running `for_each_observer()` could spawn on the executor after `unregister_interest()` returns. The notification wont go to the observer, but a race and/or crash could arise if the executor is destroyed immediately. This could effect Mir shutdown, or anywhere were executors are dropped.

RAOF:
> I was thinking it should be able to immediately destroy the executor.

It can, as long as it destroys the observer first.

I'm still not sure this is an actual (rather than potential) problem, but if we want to do something about it, my opinion is the same as before: If the problem is an object not living long enough, we should be sharing the ownership rather than adding another set of locks to emulate sharing the ownership.

wmww:
Oh, I forgot that we were holding the observer alive while spawning. In that case, I guess the guarantees are plenty. I'll drop this.

PR #2378, opened 2022-04-07
Platform: New platform probe design

RAOF:
This accomplishes multiple goals:
* Platforms can now “bind” to a device, and the platform which claims to support that device best wins.
* Platforms can list *each* device that they can drive, so later we can construct a `DisplayPlatform` or `RenderingPlatform` on the specific device.
* Platforms can share data between the `probe` and construction phase


src/platforms/gbm-kms/server/kms/platform_symbols.cpp
```diff
@@ -285,13 +285,13 @@ mg::PlatformPriority probe_rendering_platform(
     mir::udev::Enumerator drm_devices{udev};
     drm_devices.match_subsystem("drm");
     drm_devices.match_sysname("card[0-9]*");
-    drm_devices.match_sysname("render[0-9]*");
+    drm_devices.match_sysname("renderD[0-9]*");
```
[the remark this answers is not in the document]
RAOF: Yes. This means that we *actually* probe for rendernodes, which have names like `/dev/dri/renderD128`.

RAOF:
bors try

This failed because of the WLCS update.

include/platform/mir/graphics/platform.h
```diff
@@ -118,6 +119,38 @@ enum PlatformPriority : uint32_t
...
+        virtual ~DriverPrivate() = default;
+    };
+    /**
+     * Platform-private data from probing
+     *
+     * If there is any extra data helpful in creating a Platform on this device, the platform can
+     * stash it here, and it will be passed in to the platform constructor.
+     */
+    std::unique_ptr<DriverPrivate> platform_data;
+};
```
AlanGriffiths, line 152: Maybe this would be better as an interface than as `struct`.

Movable, non-copyable structs are certainly legal, but that means they are also mutable and I'm not sure that's ideal. (I wondered at first why this wasn't immutable)

include/platform/mir/graphics/platform.h
```diff
@@ -118,6 +119,38 @@ enum PlatformPriority : uint32_t
...
+        DriverPrivate() = default;
+        virtual ~DriverPrivate() = default;
+    };
+    /**
+     * Platform-private data from probing
+     *
+     * If there is any extra data helpful in creating a Platform on this device, the platform can
+     * stash it here, and it will be passed in to the platform constructor.
+     */
+    std::unique_ptr<DriverPrivate> platform_data;
```
AlanGriffiths, line 151: I think `std::any` has this covered.

AlanGriffiths (commented):
A few comments about one struct: I've not tried to get my head around the whole PR yet

include/platform/mir/graphics/platform.h
RAOF: Yeah, it's annoying that *movable* and *immutable* are mutually exclusive.

I don't have a particularly strong opinion here (although it's mildly convenient elsewhere in the code that this is *not* immutable).

include/platform/mir/graphics/platform.h
```diff
@@ -131,9 +159,10 @@ typedef mir::UniqueModulePtr<mir::graphics::RenderingPlatform>(*CreateRenderPlat
 typedef void(*AddPlatformOptions)(
     boost::program_options::options_description& config);
 
-typedef mir::graphics::PlatformPriority(*PlatformProbe)(
+typedef std::vector<mir::graphics::SupportedDevice>(*PlatformProbe)(
     std::shared_ptr<mir::ConsoleServices> const&,
-        mir::options::ProgramOption const& options);
+    std::shared_ptr<mir::udev::Context> const&,
+    mir::options::ProgramOption const& options);
```
AlanGriffiths, line 165: I think, inter alia, this is a platform ABI break: `MIR_SERVER_GRAPHICS_PLATFORM_ABI` needs bumping

src/platform/symbols.map
```diff
@@ -202,5 +202,8 @@ MIR_PLATFORM_2.8 {
     mir::graphics::DRMFormat::operator?unsigned?int*; /* Is actually operator uint32_t(), but 🤷 */
 
     mir::graphics::drm_modifier_to_string*;
-  };
+ 
+    mir::udev::Device::clone*;
+    mir::udev::Context::char_device_from_devnum*;
+   };
 } MIRPLATFORM_2.7;
```
AlanGriffiths, line 209: If I'm right about the ABI break, then we should be consolidating a MIRPLATFORM_2.8 stanza.

AlanGriffiths, line 209: If I'm right about the ABI break, then we should be consolidating a MIRPLATFORM_2.8 stanza.

AlanGriffiths (changes requested):
Looks good and works for me.

A bit of nitpicking about ABI management, but otherwise good to go.

bors delegate

include/platform/mir/graphics/platform.h
RAOF: You're right; this *does* break platform ABI (and `MIR_SERVER_GRAPHICS_PLATFORM_ABI` needs bumping), but I don't think this *does* change `libmirplatform` ABI (and hence we don't need to change SONAME).

RAOF:
I *think* this is right, but I'd prefer a second review before accepting it.

include/platform/mir/graphics/platform.h
AlanGriffiths: In further reflection you're right. We don't actually use the changes in that ABI.

AlanGriffiths (approved):
bors merge

PR #2380, opened 2022-04-07
Implement zwp_idle_inhibit_manager_v1

graysonguarino:
Basic implementation of `zwp_idle_inhibit_manager_v1` which does not currently consider the visibility of the window calling for `idle_inhibit()`.


src/include/server/mir/scene/idle_hub.h
```diff
@@ -68,6 +68,12 @@ class IdleHub
 
     virtual void unregister_interest(IdleStateObserver const& observer) = 0;
 
+    /// Inhibits idling on all surfaces until resumed
```
RAOF, line 71: Idling is not really a property of “surfaces”, right?

This *probably* doesn't need to have a comment. :)

src/server/frontend_wayland/idle_inhibit_v1.cpp
```diff
@@ -0,0 +1,173 @@
...
+}
+}
+
+auto mf::create_idle_inhibit_manager_v1(
+        wl_display* display,
+        std::shared_ptr<Executor> const& wayland_executor,
+        std::shared_ptr<scene::IdleHub> const& idle_hub)
+-> std::shared_ptr<mw::IdleInhibitManagerV1::Global>
+{
+    auto ctx = std::shared_ptr<IdleInhibitV1Ctx>{new IdleInhibitV1Ctx{wayland_executor, idle_hub}};
```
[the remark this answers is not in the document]
graysonguarino: Currently, `IdleInhibitV1Ctx` is a struct with no constructor. This seems to change the behavior of `std::make_shared()` and trying to create a `shared_ptr` with `make_shared` only works if I pass in zero arguments. This follows the pattern as seen in `text_input_v2.cpp`.

If there is a performance/security consideration in this approach, a constructor could be added to `IdleInhibitV1Ctx`.

src/server/frontend_wayland/idle_inhibit_v1.cpp
```diff
@@ -0,0 +1,173 @@
...
+}
+
+mf::IdleInhibitorV1::IdleInhibitorV1(wl_resource *resource, std::shared_ptr<IdleInhibitV1Ctx> const& ctx)
+        : wayland::IdleInhibitorV1{resource, Version<1>()},
+          ctx{ctx},
+          state_observer{std::make_shared<StateObserver>(this)}
+{
+    mir::log_info("IdleInhibitorV1 created!");
+    ctx->idle_hub->inhibit_idle();
+    ctx->idle_hub->register_interest(state_observer, time::Duration(0));
```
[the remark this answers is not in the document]
graysonguarino: From the behavior I can gather from Firefox, it seems that the client doesn't call `create_inhibitor()` when the surface is created, but rather when idle inhibiting should take place (for example, when a YouTube video is played, it asks for an inhibitor object to be created, then later destroyed upon pausing or changing the tab.)

The client, however, remains unaware of the surface's focus state, so that is Mir's responsibility to track and handle.

src/server/frontend_wayland/idle_inhibit_v1.h
```diff
@@ -0,0 +1,43 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authored by: Grayson Guarino <grayson.guarino@canonical.com>
+ */
+
+#ifndef MIR_FRONTEND_IDLE_INHIBIT_V1_H
```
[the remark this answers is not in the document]
graysonguarino: Oops, I think I was basing my style off of the generated code and not frontend code. Fixed.

src/server/frontend_wayland/idle_inhibit_v1.cpp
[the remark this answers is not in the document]
AlanGriffiths: > `std::make_shared<IdleInhibitV1Ctx>(wayland_executor, idle_hub)`

As @graysonguarino says, that doesn't work with lexical types. But adding a trivial constructor would fix it. (If adopting that approach adding a default constructor would avoid surprising semantics.)

> If there is a performance/security consideration in this approach

Consider: constructing the control block in the `shared_ptr<>` could throw. At that point, nothing owns the explicitly `newed` `IdleInhibitV1Ctx` and it will leak (along with anything it owns).

The least typing to get working code:

```suggestion
    auto ctx = std::shared_ptr<IdleInhibitV1Ctx>{IdleInhibitV1Ctx{wayland_executor, idle_hub}};
```

src/wayland/symbols.map
```diff
@@ -445,6 +445,20 @@ global:
...
+    vtable?for?mir::wayland::IdleInhibitManagerV1;
+    typeinfo?for?mir::wayland::IdleInhibitManagerV1::Global;
+    vtable?for?mir::wayland::IdleInhibitManagerV1::Global;
+    virtual?thunk?to?mir::wayland::IdleInhibitManagerV1::?IdleInhibitManagerV1*;
+
+    mir::wayland::IdleInhibitorV1::*;
+    non-virtual?thunk?to?mir::wayland::IdleInhibitorV1::*;
+    typeinfo?for?mir::wayland::IdleInhibitorV1;
+    vtable?for?mir::wayland::IdleInhibitorV1;
+    virtual?thunk?to?mir::wayland::IdleInhibitorV1::?IdleInhibitorV1*;
```
AlanGriffiths, line 460: These do not belong in the MIRWAYLAND_2.5 stanza - they were not introduced in Mir 2.5.

src/server/frontend_wayland/idle_inhibit_v1.cpp
```diff
@@ -0,0 +1,173 @@
...
+class IdleInhibitorV1
+    : public wayland::IdleInhibitorV1
+{
+public:
+    IdleInhibitorV1(wl_resource* resource, std::shared_ptr<IdleInhibitV1Ctx> const& ctx);
+    ~IdleInhibitorV1();
+
+private:
+    // Not sure why, but removing this unused StateObserver stops the functionality of the IdleInhibit.
+    // It seems unlikely that it will ever actually be used.
```
AlanGriffiths, line 75: Not sure why, but this comment suggest that this class isn't used when it clearly is.

src/server/frontend_wayland/idle_inhibit_v1.cpp
```diff
@@ -0,0 +1,173 @@
...
+    struct StateObserver : ms::IdleStateObserver
+    {
+        StateObserver(IdleInhibitorV1* idle_inhibitor)
+                : idle_inhibitor{idle_inhibitor}
+        {
+        }
+
+        void idle() override;
+
+        void active() override;
```
AlanGriffiths, line 85: Implementations missing.

AlanGriffiths (changes requested):
There are merge conflicts to resolve: It would be a good idea to rebase on main

src/wayland/symbols.map
graysonguarino: Are you saying that changes to `symbols.map` are no longer needed?

src/server/frontend_wayland/idle_inhibit_v1.cpp
graysonguarino: I never need to use the `StateObserver` for anything here, but when it is removed, Idle Inhibit stops functioning. Is there some behavior that `StateObserver` is doing that I am unaware of?

src/server/frontend_wayland/idle_inhibit_v1.cpp
graysonguarino: Will these need to be implemented? Currently, I don't need any functionality from `StateObserver`. I am currently working on implementing `SurfaceObserver` to check if the `MirWindowFocusState` changes.

src/wayland/symbols.map
AlanGriffiths: No. That, as they will be introduced in Mir 2.8, they should go in a 2.8 stanza.

src/server/frontend_wayland/idle_inhibit_v1.cpp
AlanGriffiths:
```C++
          state_observer{std::make_shared<StateObserver>(this)}
{
//...
    ctx->idle_hub->register_interest(state_observer, time::Duration(0));
```

src/server/frontend_wayland/idle_inhibit_v1.cpp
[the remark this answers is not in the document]
wmww: This pattern was inherited from my code. I didn't think to pass the struct into `make_shared()` by value. I'll have to update my other protocol impls to do that.

src/server/frontend_wayland/idle_inhibit_v1.h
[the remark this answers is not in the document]
wmww: I was never informed of these reasons, and often don't include the trailing `_`. I do now see that's it's in the style guide, so will try to include it for consistency if for nothing else. Perhaps I should update the generated code to also include include it.

src/server/scene/basic_idle_hub.cpp
```diff
@@ -261,3 +264,16 @@ void ms::BasicIdleHub::schedule_alarm(ProofOfMutexLock const&, time::Timestamp c
         alarm_timeout = std::nullopt;
     }
 }
+
+void ms::BasicIdleHub::inhibit_idle()
+{
+    // Only the window in focus should be able to call inhibit_idle()
+    mir::log_info("Calling BasicIdleHub::inhibit_idle()");
+    alarm->cancel();
```
wmww, line 272: First of all, this isn't calling `poke()` nor does it reimplement the part of poke that loops through and activates currently idle observers. This means if Mir has already hit any idle thresholds, those observer aren't woken up by the idle inhibit (which I think we want them to be).

Secondly. the alarm gets rescheduled every time the idle hub is poked (every input event). Therefore this should only function to not start the idle timeout until there is another input event. I think what's happening is the 0-timeout idle observer is preventing `BasicIdleHub` from working as intended (in a very roundabout way). 

My suggestion on how to fix this:
- add some way of checking if we are inhibited or not (such as a boolean, or locking a `weak_ptr` if you take my API suggestion)
- change `poke()` to only schedule the alarm if we are not inhibited
- call `poke()` after cancelling the alarm and marking the hub inhibited

That way any idle observers will be waken, future pokes wont schedule the alarm and you don't need the weird 0-duration timeout observer.

src/server/frontend_wayland/idle_inhibit_v1.cpp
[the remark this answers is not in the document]
wmww: This code is not tracking the surface's focus state, it's tracking Mir's idle state.

Except it's not tracking Mir's idle state because the observer is doing nothing.

Except! it is critical to this PR functioning, because the 0-duration timeout is causing Mir to schedule the idle timeout for the current time, which apparently causes it to never be called, which means Mir never idles after this is registered and the alarm is reset.

This is obviously not the ideal way to get this behavior, and I believe the `state_observer` should be removed entirely.

wmww (changes requested):

src/include/server/mir/scene/idle_hub.h
```diff
@@ -68,6 +68,10 @@ class IdleHub
 
     virtual void unregister_interest(IdleStateObserver const& observer) = 0;
 
+    virtual void inhibit_idle() = 0;
+
+    virtual void resume_idle() = 0;
```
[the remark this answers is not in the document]
graysonguarino: This is a good suggestion. I will be taking this approach instead. I hadn't considered the issues that would occur with two surfaces calling for `resume_idle()`. While I think the way my implementation works would be fine in the common case (if a surface loses focus, call `resume_idle()`, but if another surface comes into focus and wants to inhibit idle, that calls `inhibit_idle()`, thus retaining the state, this could all easily fall apart upon an exception.

PR #2381, opened 2022-04-08
Update WLCS XFAIL list

RAOF:
With WLCS CI now publishing into Mir CI we've got a new WLCS, and that has fixed a couple of tests we XFAILed.


AlanGriffiths (approved):
LGTM

bors merge

PR #2382, opened 2022-04-08
wlcs: Add target to list XFAIL tests which don't exist

RAOF:
You can run the `list_missing_wlcs_xfail_tests` to get a list of the tests that we mark as expected failures that don't exist in the installed version of WLCS.

We don't want to *fail* XFAIL tests which don't exist, because we want to be able to mark a test that is *going* to land in WLCS as XFAIL.

However, where we've added tests to the expected failure list because the *test* is incorrect we should periodically run this target and remove any XFAILs that are now removed in WLCS.


wmww (approved):
LGTM
bors r+

PR #2388, opened 2022-04-12
Fix Clang build on 22.04

RAOF:
I've got tired of not being able to build with Clang on 22.04, so as a birthday
present to me here's a branch fixing the build.


AlanGriffiths (approved):
Happy birthday!

bors merge

PR #2390, opened 2022-04-12
Stop using `signal()` has unspecified behaviour in multithreaded programs

AlanGriffiths:
Use `sigaction() throughout as `signal()` has unspecified behaviour in multithreaded programs.


RAOF (approved):
Looks sensible.

bors r+

PR #2392, opened 2022-04-12
Fix ThreadedDispatcherSignalTest.keeps_dispatching_after_signal_interruption

### LINKED ISSUE #2377: ThreadedDispatcherSignalTest.keeps_dispatching_after_signal_interruption is flaky
This test will *sometimes* fail, with `result.exited_normally()` being false at line 391.

Annoyingly, this only seems to happen when running as a part of the full test suite; just running the `ThreadedDispatcher*` tests in a loop is insufficient to reproduce.

AlanGriffiths:
Don't try to cleanup child process on exit.

There can be "random" cleanup left by other tests that we don't care about.

Fixes #2377


Saviq:
bors retty

Saviq:
bors retry

Saviq (approved):
bors retry

PR #2393, opened 2022-04-13
Don't wake threads that can't run

AlanGriffiths:
1. Do not call `condition_variable::notify_*` under lock: The woken thread cannot get lock the mutex.
2. Call `condition_variable::notify_one()` when only one thread will be able to proceed.


RAOF (approved):
Yup.

I thought we generally did this (well, not the `notify_one` bit), but we've obviously missed some.

bors r+

PR #2395, opened 2022-04-13
Change default MIR_RUN_PERFORMANCE_TESTS option to OFF

AlanGriffiths:
This test more that doubles the time taken by `make ptest`. That's not worth it as it is very unlikely to detect problems during normal development activities.

### DIFF
--- a/tests/performance-tests/CMakeLists.txt
+++ b/tests/performance-tests/CMakeLists.txt
@@ -27,7 +27,7 @@ CMAKE_DEPENDENT_OPTION(
 )
 
 CMAKE_DEPENDENT_OPTION(
-  MIR_RUN_PERFORMANCE_TESTS "Run mir_performance_tests as part of testsuite" ON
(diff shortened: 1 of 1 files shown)


Saviq:
Does it make sense to run the performance tests on the builders? Do they actually ever fail, when other tests would not?

AlanGriffiths:
> Does it make sense to run the performance tests on the builders? Do they actually ever fail, when other tests would not?

It isn't essential. Maybe adding them in the coverage build would be enough?

Saviq:
> It isn't essential. Maybe adding them in the coverage build would be enough?

I can subscribe to that :)

We wouldn't then measure _performance_ still, but we'd measure what's covered by our performance tests.

AlanGriffiths:
> > It isn't essential. Maybe adding them in the coverage build would be enough?
> 
> I can subscribe to that :)

Done

Saviq (approved):
Yup.

bors merge

Saviq:
bors retry

Issue #2375, opened 2022-04-02
Tests failing on focal/riscv64


Saviq:
https://launchpad.net/~mir-team/+archive/ubuntu/dev/+build/23517839

```plain
The following tests FAILED:
	  3 - wlcs.CopyCutPaste.* (Failed)
	 50 - wlcs.TouchInputSubsurfaces/SubsurfaceTest.one_subsurface_to_another_fallthrough/0_fails (Failed)
	 53 - wlcs.TouchInputSubsurfaces/SubsurfaceTest.pointer_input_correctly_offset_for_subsurface/0_fails (Failed)
	 55 - wlcs.TouchInputSubsurfaces/SubsurfaceTest.subsurface_extends_parent_input_region/0_fails (Failed)
	 63 - wlcs.WlShellSubsurfaces/SubsurfaceTest.one_subsurface_to_another_fallthrough/0_fails (Failed)
	 66 - wlcs.WlShellSubsurfaces/SubsurfaceTest.pointer_input_correctly_offset_for_subsurface/0_fails (Failed)
	 68 - wlcs.WlShellSubsurfaces/SubsurfaceTest.subsurface_extends_parent_input_region/0_fails (Failed)
	 73 - wlcs.XdgShellStableSubsurfaces/SubsurfaceTest.one_subsurface_to_another_fallthrough/0_fails (Failed)
	 76 - wlcs.XdgShellStableSubsurfaces/SubsurfaceTest.pointer_input_correctly_offset_for_subsurface/0_fails (Failed)
	 78 - wlcs.XdgShellStableSubsurfaces/SubsurfaceTest.subsurface_extends_parent_input_region/0_fails (Failed)
	 83 - wlcs.XdgShellV6Subsurfaces/SubsurfaceTest.one_subsurface_to_another_fallthrough/0_fails (Failed)
	 86 - wlcs.XdgShellV6Subsurfaces/SubsurfaceTest.pointer_input_correctly_offset_for_subsurface/0_fails (Failed)
	 88 - wlcs.XdgShellV6Subsurfaces/SubsurfaceTest.subsurface_extends_parent_input_region/0_fails (Failed)
	127 - miral-test.WaylandExtensions.* (Timeout)
```

Example failure:
```plain
/build/wlcs-Tgwlc7/wlcs-1.1.0+dfsg/tests/copy_cut_paste.cpp:100: Failure
[...]

PR #2398, opened 2022-04-16
[debian] disable miral tests on Launchpad riscv64 (#2375)

Saviq:


Saviq:
Since we won't know until it ends up in the PPA…

bors merge

PR #2402, opened 2022-04-22
Drop file authors

wmww:
We discussed this previously. It doesn't seem authors are required by the GPL, and their usefulness is dubious. More fine-grained and accurate info is available in git.

The command I used for the first commit (most of the changes) is:
```
find . -name '*.cpp' -o -name '*.c' -o -name '*.h' | xargs perl -i -0777pe 's/\s*\*\s*\n\s*\*\s*(authored by|authors?):[^\n]*(\s*\*\s+[^\n]*)*//sei'
```


AlanGriffiths (approved):
WFM

bors merge

PR #2405, opened 2022-04-25
Implement zwp_idle_inhibit_manager_v1 with WakeLock

graysonguarino:
This is a revised version of the previous draft PR #2380 with the WakeLock approach to inhibiting idle. This also adds a `SurfaceObserver` that will be used when adding the final functionality in later stages. For a full implementation of idle inhibition, the state of the surfaces should be observed and only those active should be allowed to inhibit idle.


src/server/frontend_wayland/idle_inhibit_v1.cpp
```diff
@@ -0,0 +1,198 @@
...
+            if (attrib == mir_window_attrib_focus && value == mir_window_focus_state_unfocused)
+                ; // TODO
+            else if (attrib == mir_window_attrib_focus)
+                ; // TODO
+        };
+    };
+
+    std::shared_ptr<Executor> const wayland_executor;
+    std::shared_ptr<ms::IdleHub> const idle_hub;
+    std::shared_ptr<ms::SurfaceObserver> surface_observer;
```
AlanGriffiths, line 64: I'm not a big fan of shared structs with mutable state: it can become hard to enforce invariants.

Relatedly I also see `ctx->idle_hub->inhibit_idle()` which might better be written `ctx->inhibit_idle()` if we made this a class (with member functions and private data).

AlanGriffiths (changes requested):
I have a suspicion that you would have come to a simpler design faster if you started with a test you needed to pass and wrote just the code needed to pass it.

src/server/frontend_wayland/idle_inhibit_v1.cpp
wmww: The point of the ctx object is to provide access to a set of other things with as little code overhead as possible. In my opinion it's not supposed to encapsulate any functionality.

I agree it's not the right place for mutable state.

src/include/server/mir/scene/idle_hub.h
```diff
@@ -68,6 +76,8 @@ class IdleHub
 
     virtual void unregister_interest(IdleStateObserver const& observer) = 0;
 
+    virtual std::shared_ptr<IdleHub::WakeLock> inhibit_idle() = 0;
```
wmww, line 79: We generally use `auto foo() -> return_type` syntax in new code.

src/server/frontend_wayland/idle_inhibit_v1.cpp
```diff
@@ -0,0 +1,198 @@
...
+public:
+    IdleInhibitManagerV1(
+            wl_resource* resource,
+            std::shared_ptr<IdleInhibitV1Ctx> const& ctx);
+
+
+private:
+    void create_inhibitor(struct wl_resource* id, struct wl_resource* surface) override;
+
+    std::shared_ptr<IdleInhibitV1Ctx> const& ctx;
```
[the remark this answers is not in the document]
wmww: Yes, passing around reference to smart pointers often makes sense, but rarely do you want to store them. If you know for sure the object will outlive the reference than a reference directly to the pointed at object makes more sense, and if you want to hold the object alive or safely cope with it being destroyed a smart or weak pointer respectively (not a reference to those types) is required.

src/server/frontend_wayland/idle_inhibit_v1.cpp
graysonguarino: Upon review, there is no reason that `surface_observer` needs to be mutable. I believe that got held over from my previous implementation.

If the declaration of `SurfaceObserver` exists within `IdleInhibitV1Ctx`, *without* a mutable state of `surface_observer`, should I still make this a class or leave it as a struct.

src/server/scene/basic_idle_hub.cpp
```diff
@@ -114,6 +117,17 @@ struct ms::BasicIdleHub::PendingRegistration
...
 
+ms::BasicIdleHub::WakeLock::WakeLock(std::shared_ptr<BasicIdleHub> idle_hub)
+        : idle_hub{idle_hub}
+{
+}
+
+ms::BasicIdleHub::WakeLock::~WakeLock()
+{
+    idle_hub.lock()->alarm->reschedule_in(std::chrono::duration_cast<std::chrono::milliseconds>(
+            idle_hub.lock()->alarm_timeout.value()));
```
wmww, line 128: You're calling `.lock()` on a `weak_ptr` and directly using that result. `.lock()` returns a `shared_ptr` to the object if it still exists, or a null `shared_ptr` if it doesn't. You can either use a `shared_ptr` to keep the idle hub alive as long as this object or you can:
```cpp
if (auto const shared_hub = idle_hub.lock())
{
    shared_hub->...
}
```
But you generally should not do `.lock()->` on a `weak_ptr`

src/server/scene/basic_idle_hub.cpp
```diff
@@ -261,3 +275,18 @@ void ms::BasicIdleHub::schedule_alarm(ProofOfMutexLock const&, time::Timestamp c
...
+    {
+        auto shared_lock = std::make_shared<BasicIdleHub::WakeLock>(shared_from_this());
+        wake_lock = shared_lock;  // WHERE SHOULD I GO?
+        alarm->cancel(); // TODO - move this logic into state logic
+        return shared_lock;
+    }
+    else // wake_lock is not null
+    {
+        return wake_lock.lock();
+    }
```
wmww, line 291: This is racy. From [cppreference](https://en.cppreference.com/w/cpp/memory/weak_ptr/expired):
> This function is inherently racy if the managed object is shared among threads. In particular, a false result may become stale before it can be used. A true result is reliable. 

The object may not be expired when `.expired()` is checked, but is expired by the time `wake_lock.lock()` returns. This would lead to returning a null shared pointer. In this case the wake lock object isn't called into so it being null would just cause the wake lock not to work, but in many other cases it would lead to a crash.

Instead what you want to do is lock it first and then check if the returned shared pointer is null or not. If you put the lock in the if you can condense the code a bit:
```cpp
if (auto const shared_wake_lock = wake_lock.lock())
{
    return shared_wake_lock;
}
else
{
    ...
}
```

You also probably want to do any changes to the `BasicIdleHub` state (such as assigning the wake_lock and canceling the alarm) under lock.

src/server/scene/basic_idle_hub.h
```diff
@@ -52,6 +51,17 @@ class BasicIdleHub : public IdleHub
...
 
+    struct WakeLock: public IdleHub::WakeLock
+    {
+    public:
+        WakeLock(std::shared_ptr<BasicIdleHub> idle_hub);
+
+        ~WakeLock() override;
+
+    private:
+        std::weak_ptr<BasicIdleHub> idle_hub;
```
wmww, line 62: I believe this can be `const`

src/server/scene/basic_idle_hub.h
```diff
@@ -76,6 +88,7 @@ class BasicIdleHub : public IdleHub
 
     std::shared_ptr<time::Clock> const clock;
     std::unique_ptr<time::Alarm> const alarm;
+    std::weak_ptr<BasicIdleHub::WakeLock> wake_lock;
```
wmww, line 91: Generally we put mutable state below the `mutex` to indicate it should only be changed under lock.

src/wayland/generated/idle-inhibit-unstable-v1_wrapper.h
```diff
@@ -0,0 +1,82 @@
+/*
```
wmww, line 1: Since you started this PR, we stopped committing generated code to git and deleted `src/wayland/generated`. The genereated files can be removed (they're now auto-built in the build directory)

wmww (changes requested):

src/server/frontend_wayland/idle_inhibit_v1.cpp
AlanGriffiths: > Alan may disagree.

In general, I'd prefer to keep state directly in an object rather than in a shared subobject. But my primary concern here was _shared mutable state_.

Specifically, I'd either:

1. Copy the `wayland_executor`, `idle_hub` and `surface_observer` from the `ctx` supplied to the constructor as required and use them directly; or,
2. Make `ctx` into a class that forwards requests to the corresponding subobjects.

src/server/frontend_wayland/idle_inhibit_v1.cpp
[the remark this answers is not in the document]
AlanGriffiths: > Yes, passing around reference to smart pointers often makes sense

MUCH less so since the incorporation of move semantics into the language and library.

The normal idiom should now be copy (or move) into a call or constructor and then move (or swap) as needed.

src/server/scene/basic_idle_hub.cpp
graysonguarino: That *sure does* make more sense.

AlanGriffiths (changes requested):

AlanGriffiths:
And here's your next test case:
```c++
TEST_F(BasicIdleHub, when_a_wake_lock_is_held_calling_poke_does_not_restart_idle_timeout)
{
    auto const observer = std::make_shared<NiceMock<MockObserver>>();
    hub->register_interest(observer, executor, 5s);
    EXPECT_CALL(*observer, active()).Times(AnyNumber());

    {
        auto const wake_lock = hub->inhibit_idle();
        EXPECT_CALL(*observer, idle()).Times(0);

        hub->poke();
        advance_by(6s);
        executor.execute();

        Mock::VerifyAndClearExpectations(observer.get());
    }

    EXPECT_CALL(*observer, idle()).Times(AnyNumber());
    advance_by(6s);
    executor.execute();
}
```

src/include/server/basic_idle_hub.h
```diff
@@ -35,13 +35,12 @@ class AlarmFactory;
 }
 namespace scene
 {
-
```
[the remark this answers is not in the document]
graysonguarino: I see that you said that with the new changes this move is no longer necessary, but do you believe this remains the preferred place for `idle_hub.h`?

AlanGriffiths:
idle_hub.h is already in the right place. No need to move it

src/server/frontend_wayland/idle_inhibit_v1.cpp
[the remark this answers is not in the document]
AlanGriffiths: This is still storing a dangling reference:
```c++
    auto ctx = std::make_shared<IdleInhibitV1Ctx>(IdleInhibitV1Ctx{wayland_executor, idle_hub});
    return std::make_shared<IdleInhibitManagerV1Global>(display, ctx);
```
`ctx` goes out of scope after the `return`.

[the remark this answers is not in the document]
graysonguarino: I want `IdleInhibitManagerV1Global` to own `ctx` so I followed your first suggestion and changed it to a non-reference.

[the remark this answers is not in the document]
wmww: > The normal idiom should now be copy (or move) into a call or constructor and then move (or swap) as needed.

I'm not yet sold on that, we should discuss at some point.

Grayson It looks like you changed the constructor argument, but not the value held by the class. The important thing is that the class member is not a reference, or else as soon as the `shared_ptr` instance it refers to is dropped it will be invalid.

src/server/frontend_wayland/wayland_default_configuration.cpp
```diff
@@ -141,6 +142,13 @@ std::vector<ExtensionBuilder> const internal_extension_builders = {
                 ctx.text_input_hub,
                 ctx.composite_event_filter);
         }),
+    make_extension_builder<mw::IdleInhibitManagerV1>([](auto const& ctx)
+        {
+            return mf::create_idle_inhibit_manager_v1(
+                    ctx.display,
+                    ctx.wayland_executor,
+                    ctx.idle_hub);
```
wmww, line 150: Double indentation

src/server/scene/basic_idle_hub.cpp
```diff
@@ -133,6 +133,9 @@ ms::BasicIdleHub::~BasicIdleHub()
 void ms::BasicIdleHub::poke()
 {
     std::lock_guard lock{mutex};
+    if (!wake_lock.expired())
+        return;
```
wmww, line 137: No braces around if block

src/server/scene/basic_idle_hub.cpp
```diff
@@ -259,3 +262,35 @@ void ms::BasicIdleHub::schedule_alarm(ProofOfMutexLock const&, time::Timestamp c
...
+    struct WakeLock: public IdleHub::WakeLock
+    {
+        WakeLock(std::weak_ptr<IdleHub> idle_hub) : idle_hub{std::move(idle_hub)}
+        {
+        }
+
+        ~WakeLock()
+        {
+            if (auto const shared_hub = idle_hub.lock())
+                shared_hub->poke();
```
wmww, line 277: No braces around if block

tests/unit-tests/scene/test_basic_idle_hub.cpp
```diff
@@ -41,7 +41,8 @@ struct BasicIdleHub: Test
 {
     mtd::AdvanceableClock clock;
     mtd::FakeAlarmFactory alarm_factory{};
-    ms::BasicIdleHub hub{mt::fake_shared(clock), alarm_factory};
+    std::shared_ptr<ms::BasicIdleHub> hub{std::make_shared<ms::BasicIdleHub>(
+        mt::fake_shared(clock), alarm_factory)};
```
wmww, line 45: If we do a line split like this, we generally break each argument onto a different line (either of the following would be preferred):
```cpp
    std::shared_ptr<ms::BasicIdleHub> hub{std::make_shared<ms::BasicIdleHub>(
        mt::fake_shared(clock),
        alarm_factory)};
```
```cpp
    std::shared_ptr<ms::BasicIdleHub> hub{
        std::make_shared<ms::BasicIdleHub>(mt::fake_shared(clock), alarm_factory)};
```

wmww (changes requested):

src/server/scene/basic_idle_hub.cpp
graysonguarino: It may be wise to [update this.](https://mir-server.io/doc/cppguide/?showone=Conditionals#Conditionals)

wmww: Indeed! #2416

src/server/frontend_wayland/idle_inhibit_v1.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+}
+
+auto mf::create_idle_inhibit_manager_v1(
+    wl_display* display,
+    std::shared_ptr<Executor> const& wayland_executor,
+    std::shared_ptr<ms::IdleHub> const& idle_hub)
+-> std::shared_ptr<mw::IdleInhibitManagerV1::Global>
+{
+    auto ctx = std::make_shared<IdleInhibitV1Ctx>(IdleInhibitV1Ctx{wayland_executor, idle_hub});
```
AlanGriffiths, line 100: [non-blocking] Again, it is better to pass smart pointers by value and move them.
```suggestion
    std::shared_ptr<Executor> wayland_executor,
    std::shared_ptr<ms::IdleHub> idle_hub)
-> std::shared_ptr<mw::IdleInhibitManagerV1::Global>
{
    auto ctx = std::make_shared<IdleInhibitV1Ctx>(IdleInhibitV1Ctx{std::move(wayland_executor), std::move(idle_hub)});
```

tests/unit-tests/scene/test_basic_idle_hub.cpp
```diff
@@ -252,17 +252,121 @@ TEST_F(BasicIdleHub, observer_marked_idle_after_shorter_timeout_removed_and_poke
...
     advance_by(2s);
     executor.execute();
 }
+
+TEST_F(BasicIdleHub, inhibit_idle_inhibits_until_wake_lock_released)
+{
+    auto const observer = std::make_shared<NiceMock<MockObserver>>();
+    hub->register_interest(observer, executor, 5s);
+    // Check that it goes idle
+    EXPECT_CALL(*observer, idle());
```
AlanGriffiths, line 275: This (and other) test is testing multiple things. It should focus on testing one thing, and if two thing need testing there should be two tests.

Specifically, `inhibit_idle_inhibits_until_wake_lock_released` should not fail because "Check that it goes idle" fails. AFAICS the only reason this expectation is set is copy/paste.

tests/unit-tests/scene/test_basic_idle_hub.cpp
```diff
@@ -252,17 +252,121 @@ TEST_F(BasicIdleHub, observer_marked_idle_after_shorter_timeout_removed_and_poke
...
+    advance_by(6s);
+    executor.execute();
+    // Check that it stays awake if idle inhibited
+    hub->poke();
+    {
+        auto wake_lock = hub->inhibit_idle();
+        EXPECT_CALL(*observer, active());
```
AlanGriffiths, line 282: I think this is the expectation that _this_ test should be testing.

tests/unit-tests/scene/test_basic_idle_hub.cpp
```diff
@@ -252,17 +252,121 @@ TEST_F(BasicIdleHub, observer_marked_idle_after_shorter_timeout_removed_and_poke
...
+        advance_by(6s);
+        executor.execute();
+    }
+    // Check that it goes idle when wake_lock goes out of scope
+    EXPECT_CALL(*observer, idle());
```
AlanGriffiths, line 287: I think this expectation belongs to another test:

`when_a_wake_lock_is_released_idle_timeout_is_restarted`

src/include/server/mir/scene/idle_hub.h
```diff
@@ -50,6 +50,16 @@ class IdleHub
...
 
+    /// Lock that is held by the surface which is inhibiting idle
+    struct WakeLock
+    {
+    public:
+        WakeLock() = default;
+        virtual ~WakeLock() = default;
+        WakeLock(WakeLock const&) = delete;
+        WakeLock& operator=(WakeLock const&) = delete;
+    };
```
AlanGriffiths, line 61: There's no need for this type to be **defined** in the header. (Only declared)

If the definition is moved to the .cpp, then you can also refactor by "pushing up" the derived class implementation and lose the derived class.

tests/unit-tests/scene/test_basic_idle_hub.cpp
```diff
@@ -252,17 +252,121 @@ TEST_F(BasicIdleHub, observer_marked_idle_after_shorter_timeout_removed_and_poke
...
+        hub->poke();
+        advance_by(6s);
+        executor.execute();
+
+        Mock::VerifyAndClearExpectations(observer.get());
+    }
+
+    EXPECT_CALL(*observer, idle()).Times(AnyNumber());
+    advance_by(6s);
+    executor.execute();
```
AlanGriffiths, line 371: I know I suggested this test, but it also has too many expectations:

```suggestion
    }
```

include/test/mir/test/doubles/mock_idle_hub.h
```diff
@@ -37,6 +37,7 @@ struct MockIdleHub : mir::scene::IdleHub
         mir::NonBlockingExecutor&,
         time::Duration));
     MOCK_METHOD1(unregister_interest, void(mir::scene::IdleStateObserver const&));
+    MOCK_METHOD0(inhibit_idle, std::shared_ptr<IdleHub::WakeLock>());
```
AlanGriffiths, line 40: [Non-blocking]

We try to use `MOCK_METHOD` in new code and clean up `MOCK_METHOD<n>` as we go. E.g.

```suggestion
    MOCK_METHOD(std::shared_ptr<IdleHub::WakeLock>, inhibit_idle, ());
```

AlanGriffiths (changes requested):
No big problems. And flagged some possible improvements I won't block on, but you should be aware of.

The tests are better than some already in the codebase but could be even better if each only tested one thing. (And as that's actually easy starting from here, I'll insist on it.)

src/include/server/mir/scene/idle_hub.h
graysonguarino: I am a bit confused: are you suggesting that WakeLock be defined within `idle_hub.cpp`? If so, that doesn't exist. 

If instead it should be defined within `basic_idle_hub.cpp`, should WakeLock still be declared within `idle_hub.h`? I don't believe non-defined structs can be inherited as they're incomplete types.

AlanGriffiths:
> On 9 May 2022, at 20:29, Grayson Guarino ***@***.***> wrote:
> 
> If instead it should be defined within basic_idle_hub.cpp, should WakeLock still be declared within idle_hub.h? I don't believe non-defined structs can be inherited as they're incomplete types.

Declare the class in idle_hub.h, define the class in basic_idle_hub.cpp. I.e. where a definition is used

src/include/server/mir/scene/idle_hub.h
AlanGriffiths: Yes, declare `WakeLock` in `idle_hub.h`, just remove the definition from there and define in `basic_idle_hub.cpp` - the only place a definition is used or needed.

graysonguarino: I am having an issue where I cannot inherit `IdleHub::WakeLock` when it is undefined in `idle_hub.h` as it is considered an incomplete type.

Maybe you can help clear up some confusion surrounding inheritance that I've been struggling with:
If we are declaring that `IdleHub::inhibit_idle() -> std::shared_ptr<WakeLock>`, but we withhold the implementation details of the WakeLock until `BasicIdleHub::WakeLock`, how can the incomplete type `IdleHub::WakeLock` be held in `BasicIdleHub::wake_lock` which is a `std::weak_ptr<IdleHub::WakeLock>`? 

How can the destructor that we are declaring inside `BasicIdleHub::WakeLock` be called if its type is treated as `IdleHub::WakeLock` and not `BasicIdleHub::WakeLock`?

If `BasicIdleHub::wake_lock` should hold type `std::weak_ptr<BasicIdleHub::WakeLock>`, then `BasicIdleHub::inhibit_idle()` can not match the interface and return an `std::shared_ptr<IdleHub::WakeLock>` as that would require downcasting to an incomplete type.

AlanGriffiths: 1. Why do you want to introduced a `BasicIdleHub::WakeLock`?
2. `std::weak_ptr<>` can hold an incomplete type.
3. If you use `make_shared<>` then it creates a _deleter_ that deletes the creation type.

Here's what I mean:

```
diff --git a/src/include/server/mir/scene/idle_hub.h b/src/include/server/mir/scene/idle_hub.h
index 5f4ff073a7..65a1df5194 100644
--- a/src/include/server/mir/scene/idle_hub.h
+++ b/src/include/server/mir/scene/idle_hub.h
@@ -51,14 +51,7 @@ public:
     virtual ~IdleHub() = default;
 
     /// Lock that is held by the surface which is inhibiting idle
-    struct WakeLock
-    {
-    public:
-        WakeLock() = default;
-        virtual ~WakeLock() = default;
-        WakeLock(WakeLock const&) = delete;
-        WakeLock& operator=(WakeLock const&) = delete;
-    };
+    struct WakeLock;
 
     /// Wakes Mir if it's not already awake, and starts the timer from now
     virtual void poke() = 0;
diff --git a/src/server/scene/basic_idle_hub.cpp b/src/server/scene/basic_idle_hub.cpp
index 2abece0744..bffe8cfb81 100644
--- a/src/server/scene/basic_idle_hub.cpp
+++ b/src/server/scene/basic_idle_hub.cpp
@@ -265,26 +265,27 @@ void ms::BasicIdleHub::schedule_alarm(ProofOfMutexLock const&, time::Timestamp c
     }
 }
 
-auto ms::BasicIdleHub::inhibit_idle() -> std::shared_ptr<WakeLock>
+/// Lock that is held by the surface which is inhibiting idle
+struct ms::IdleHub::WakeLock
 {
-    struct WakeLock: public IdleHub::WakeLock
+    WakeLock(std::weak_ptr<IdleHub> idle_hub) : idle_hub{std::move(idle_hub)}
     {
-        WakeLock(std::weak_ptr<IdleHub> idle_hub) : idle_hub{std::move(idle_hub)}
-        {
-        }
+    }
 
-        ~WakeLock()
+    ~WakeLock()
+    {
+        if (auto const shared_hub = idle_hub.lock())
         {
-            if (auto const shared_hub = idle_hub.lock())
-            {
-                shared_hub->poke();
-            }
+            shared_hub->poke();
         }
+    }
 
-    private:
-        std::weak_ptr<IdleHub> const idle_hub;
-    };
+private:
+    std::weak_ptr<IdleHub> const idle_hub;
+};
 
+auto ms::BasicIdleHub::inhibit_idle() -> std::shared_ptr<WakeLock>
+{
     std::lock_guard<std::mutex> lock(mutex);
     if (auto const shared_wake_lock = wake_lock.lock()) // wake_lock is already held
     {
```

graysonguarino: > 1. Why do you want to introduced a `BasicIdleHub::WakeLock`?

This was my issue. I was introducing a new type of WakeLock that would inherit from `idle_hub.h`.

src/server/frontend_wayland/idle_inhibit_v1.cpp
```diff
@@ -0,0 +1,149 @@
...
+        {
+        }
+
+        void idle() override;
+
+        void active() override;
+    };
+
+    std::shared_ptr<IdleInhibitV1Ctx> const ctx;
+    std::shared_ptr<StateObserver> const state_observer;
```
wmww, line 79: This doesn't do anything, does it? Previously it was inadvertently causing something to work, but I think it can be removed now.

tests/unit-tests/scene/test_basic_idle_hub.cpp
```diff
@@ -252,17 +252,106 @@ TEST_F(BasicIdleHub, observer_marked_idle_after_shorter_timeout_removed_and_poke
...
+        auto wake_lock_2 = hub->inhibit_idle();
+        EXPECT_CALL(*observer, idle()).Times(0);
+        advance_by(6s);
+        executor.execute();
+    }
+
+    EXPECT_CALL(*observer, idle());
+    advance_by(6s);
+    executor.execute();
+}
```
AlanGriffiths, line 318: It is unclear from the test name what the requirement being tested is.

Also the EXPECTS look like they are testing two different requirements.

Is it:

1. `when_two_wake_locks_are_held_idle_timeout_is_suspended` (first EXPECT)
2. `when_two_wake_locks_are_held_and_released_idle_timeout_is_resumed` (second EXPECT)

Maybe it is trying to test both things?

Do we also need.

3. `when_one_of_two_wake_locks_is_released_idle_timeout_is_suspended`?

tests/unit-tests/scene/test_basic_idle_hub.cpp
```diff
@@ -252,17 +252,106 @@ TEST_F(BasicIdleHub, observer_marked_idle_after_shorter_timeout_removed_and_poke
...
+
+TEST_F(BasicIdleHub, inhibit_idle_inhibits_when_a_wake_lock_is_held_then_released_then_held)
+{
+    auto const observer = std::make_shared<NiceMock<MockObserver>>();
+    hub->register_interest(observer, executor, 5s);
+    hub->poke();
+
+    {
+        auto wake_lock = hub->inhibit_idle();
+        EXPECT_CALL(*observer, idle()).Times(0);
```
AlanGriffiths, line 328: This EXPECT is already tested by `inhibit_idle_inhibits_until_wake_lock_released`, no need for it here

src/server/frontend_wayland/idle_inhibit_v1.cpp
graysonguarino: It works without it now! Still don't exactly understand why this was needed for it work under the old implementation, but it passes tests and runs fine without the StateObserver now.

AlanGriffiths (approved):
WFM

wmww (approved):
LGTM
bors r+

PR #2406, opened 2022-04-26
Refactor out mf::MirDisplay

wmww:
This moves `MirDisplay`'s functionality to `OutputManager`, and fixes some of it's problems. Previously we used the base configuration in places where we should be using the currently applied configuration. Now we use the correct configuration everywhere, and the logic is more centralized. This does not fix any bugs we're currently seeing in production, the motivation is code quality and technical correctness.


src/server/frontend_wayland/output_manager.cpp
```diff
@@ -213,19 +215,42 @@ void mf::Output::resource_destructor(wl_resource* resource)
     client_resource_list.erase(erase_from, client_resource_list.end());
 }
 
+struct mf::OutputManager::DisplayConfigObserver: public graphics::NullDisplayConfigurationObserver
```
RAOF, line 218: Probably a `class`?

src/server/frontend_wayland/output_manager.cpp
```diff
@@ -213,19 +215,42 @@ void mf::Output::resource_destructor(wl_resource* resource)
...
-mf::OutputManager::OutputManager(wl_display* display, std::shared_ptr<MirDisplay> const& display_config, std::shared_ptr<Executor> const& executor) :
-    display_config_{display_config},
-    display{display},
-    executor{executor}
+mf::OutputManager::OutputManager(
+    wl_display* display,
+    std::shared_ptr<Executor> const& executor,
+    std::shared_ptr<ObserverRegistrar<graphics::DisplayConfigurationObserver>> const& registrar)
+    : display{display},
+      executor{executor},
```
RAOF, line 244: You should probably store the executor in the `DisplayConfigObserver` rather than the `OutputManager`; then you maintain the invariant `ObserverRegistrar` wants which is that the Executor lives at least as long as the Observer.

RAOF (changes requested):
Seems broadly sensible. Once the Executor lifetime thing is fixed feel free to land.

RAOF (approved):
LGTM now.

bors r+

RAOF:
bors merge

PR #2414, opened 2022-05-02
Kill clients with error instead of sending unsupported protocol messages

### LINKED ISSUE #2343: Too easy to send unsupported events
If a client binds to a version of a protocol that is older than what we support, sending newer events will generally crash the client. We should not do this. As #2341 demonstrates it is quite easy to do this accidentally. We should consider changing the generated code to either make this harder, or make it harmless.

wmww:
Fixes #2343

### DIFF
--- a/src/wayland/wayland_base.cpp
+++ b/src/wayland/wayland_base.cpp
@@ -151,3 +151,23 @@ void mw::internal_error_processing_request(wl_client* client, char const* method
         std::current_exception(),
         std::string() + "Exception processing " + method_name + " request");
 }
+
+void mw::tried_to_send_unsupported_event(wl_client* client, wl_resource* resource, char const* event, int required_version)
+{
+    wl_client_post_implementation_error(
+        client,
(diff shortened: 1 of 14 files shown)


src/wayland/wayland_base.cpp
```diff
@@ -151,3 +151,23 @@ void mw::internal_error_processing_request(wl_client* client, char const* method
...
+    wl_client_post_implementation_error(
+        client,
+        "Mir tried to send %s@%d.%s to object with version %d (requires version %d)",
+        wl_resource_get_class(resource),
+        wl_resource_get_id(resource),
+        event,
+        wl_resource_get_version(resource),
+        required_version);
+
+    log_error(
```
RAOF, line 166: Nit: Maybe `log_critical`? Getting here means that we've hit an application-breaking Mir logic bug.

RAOF (approved):
LGTM.

Feel free to switch to `log_critical`, but I won't need to re-review for that.

bors delegate=wmww

wmww:
bors r=RAOF

Saviq:
bors retry

PR #2415, opened 2022-05-02
XWaylandSurface: don't allow inflight_configures to grow indefinitely

wmww:


wmww:
This is a response to reading the code and thinking "hmm, that seems like something that could go wrong", not an observed issue.

RAOF (approved):
Seems like sensible paranoia.

bors r+

PR #2419, opened 2022-05-09
Simplify Wayland EGL helpers

RAOF:
The various Wayland EGL helpers do not need an `EGLContext` and a Wayland executor; they only need access to a thread with a current `EGLContext` to do cleanup.

At the time they were written, we did not have a good way to provide that, but now we have `EGLContextExecutor` which provides exactly (and only) that.


AlanGriffiths (approved):
Yes, that's a fitter abstraction.

bors merge

AlanGriffiths:
bors retry

AlanGriffiths:
bors merge

Saviq (approved):
Yups for the merge.

bors r=AlanGriffiths

PR #2420, opened 2022-05-09
Fix possible ODR flagged by clang-tidy

RAOF:


tests/include/mir/test/doubles/stub_buffer_allocator.h
```diff
@@ -38,7 +38,7 @@ namespace doubles
 
 namespace
 {
-void memcpy_from_shm_buffer(struct wl_shm_buffer* buffer)
```
AlanGriffiths, line 41: How is that a possible ODR violation? The anon namespace ensures that different translation units have distinct definitions.

AlanGriffiths: Oh! To answer my own question: it is used by an inline function in the header. Different translation units will have different definitions of the `buffer_from_shm()` function as they use different definitions of `buffer_from_shm()`.

But inlining `buffer_from_shm()` doesn't fix that. (Even if it placates clang-tidy)

AlanGriffiths (changes requested):
This is the wrong fix (or only part of the fix).

The problem is caused by _using_ a name from the anon namespace in an inline function.

There are two natural ways to avoid this:

1. Move all the implementation of `StubBufferAllocator` into a .cpp file.
2. Remove the anon namespace from the header and either make `memcpy_from_shm_buffer()` inline or move it to a .cpp.

AlanGriffiths (approved):
bors merge

PR #2428, opened 2022-05-16
udev: Test (and fix) Context::char_device_from_devnum

RAOF:
Fixes: https://github.com/MirServer/mir-test-tools/issues/50

### DIFF
--- a/src/platform/udev/udev_wrapper.cpp
+++ b/src/platform/udev/udev_wrapper.cpp
@@ -300,7 +300,7 @@ auto mu::Context::device_from_syspath(std::string const& syspath) -> std::unique
 
 auto mu::Context::char_device_from_devnum(dev_t devnum) -> std::unique_ptr<Device>
 {
-    return std::make_unique<DeviceImpl>(udev_device_new_from_devnum(context, 'C', devnum));
+    return std::make_unique<DeviceImpl>(udev_device_new_from_devnum(context, 'c', devnum));
 }
 
 udev* mu::Context::ctx() const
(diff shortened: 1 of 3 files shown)


AlanGriffiths (approved):
bors merge

RAOF:
I suspect the test failure to be `umockdev`-version-related. Whatever it is *might* have been fixed in 0.17.0?

Saviq:
I can confirm this gets us further in the lab:

```plain
[2022-05-16 10:53:54.262957] <information> GLRenderer: GL vendor: NVIDIA Corporation                                                                                                                       
[2022-05-16 10:53:54.262981] <information> GLRenderer: GL renderer: NVS 510/PCIe/SSE2                                                                                                                      
[2022-05-16 10:53:54.262991] <information> GLRenderer: GL version: OpenGL ES 3.2 NVIDIA 470.103.01                                                                                                         
[2022-05-16 10:53:54.262999] <information> GLRenderer: GLSL version: OpenGL ES GLSL ES 3.20
```

Saviq:
> I suspect the test failure to be `umockdev`-version-related. Whatever it is _might_ have been fixed in 0.17.0?

Fails on: 0.14.1 and 0.16.3, so that tracks. Any way to work around (I see Alan worked around…)?

Also, we shouldn't be dropping the exception on the floor, we should at least log it? Could meaningfully be a separate PR though.

AlanGriffiths:
@Saviq I agree with @RAOF's suspicion - umockdev 1.14.1 is not doing what I expect. Can you approve the hack I've added?

AlanGriffiths:
> Also, we shouldn't be dropping the exception on the floor, we should at least log it? Could meaningfully be a separate PR though.

The exception was only "dropped" by the test, which logs it. (I.e. "nothing to see here")

Saviq (approved):
> Did you mean with the original issue? I think that was already addressed by #2430

Ah yes, that's what I meant. Didn't catch up on everything that happened yet.

> Can you approve the hack I've added?

bors merge r=AlanGriffiths

Issue #2061, opened 2021-06-08
[eglstream-kms] GLX clients in XWayland crash Mir


RAOF:
Attempting to start `glxgears` from xterm when using the eglstream-kms platform results in Mir crashing, with
```
terminate called after throwing an instance of 'boost::wrapexcept<std::system_error>'
  what():  Failed to create EGLStream from Wayland buffer: EGL_BAD_ACCESS (0x3002)
```

The proximate cause is that the `mge::BufferAllocator` has an implementation of a Wayland interface not built using the wrapper-generator, and so exceptions bubble up into the C code of the Wayland main loop.

It's not entirely clear what the underlying cause is here, though. The relevant [EGL extension](https://www.khronos.org/registry/EGL/extensions/KHR/EGL_KHR_stream.txt) doesn't list `EGL_BAD_ACCESS` as an error the call can generate (at least, not in the way we're calling it), and it's not clear how to gracefully handle a failure, as the relevant protocol has no support for failure events.

PR #2429, opened 2022-05-16
eglstream-kms: Kill clients, not Mir, when they submit bad EGLStreams

RAOF:
This seems to particularly affect XWayland when clients try to use GLX.
It's not great to kill XWayland, but it's better than killing Mir.

We should probably additionally move this code to the mirwayland wrapper
generator, but this particular fix is simple.

Fixes: #2061


AlanGriffiths (approved):
It feels bad not resolving the underlying issue, but that's probably out of our control

bors merge

Issue #2426, opened 2022-05-13
Hybrid AMD/NVIDIA system fails


RAOF:
Something in Mesa has changed, and now Mir fails to start on my AMD/NVIDIA hybrid system.
```
[2022-05-13 17:43:28.603698] <information> mirserver: Found rendering driver: mir:eglstream-kms (version 2.8.0)
libEGL warning: DRI2: failed to create dri screen
libEGL warning: DRI2: failed to create dri screen
[2022-05-13 17:43:28.645889] < - debug - > eglstream: Failed to initialise EGL: EGL_NOT_INITIALIZED (0x3001)
[2022-05-13 17:43:28.658006] <information> eglstream: EGLDevice found but unsuitable. Missing extension EGL_KHR_stream_consumer_gltexture
[2022-05-13 17:43:28.658049] <information> eglstream: EGLDevice found but unsuitable. Missing extension EGL_NV_stream_attrib
ERROR: /home/chris/Canonical/Mir/mir/main/src/server/graphics/default_configuration.cpp(335): Throw in function virtual const std::vector<std::shared_ptr<graphics::RenderingPlatform>> &mir::DefaultServerConfiguration::the_rendering_platforms()
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: Exception while creating rendering platform
ERROR: /home/chris/Canonical/Mir/mir/main/src/platforms/eglstream-kms/server/utils.cpp(35): Throw in function dev_t mir::graphics::eglstream::devnum_for_device(EGLDeviceEXT)
Dynamic exception type: boost::wrapexcept<std::system_error>
std::exception::what: Failed to determine DRM device node path from EGLDevice: EGL_BAD_PARAMETER (0x300c)
```
It looks like Mesa is supporting more of the EGLDevice platform, but is doing so in a way that we don't expect.

PR #2430, opened 2022-05-16
eglstream-kms: Handle EGL errors in `devnum_for_device`

RAOF:
It seems that Mesa is now exposing rendernodes as EGLDevices, but using the EGL query to find the device node of such devices fails.

Handle those exceptions by assuming the device is unsuitable.

Fixes: #2426


AlanGriffiths (approved):
bors merge

PR #2432, opened 2022-05-18
[mir-smoke-test-runner] Enable working in a Wayland only environment

AlanGriffiths:
Inspired by https://github.com/MirServer/mir/pull/2423#issuecomment-1129215985

Wayland support on desktop is becoming common and we ought to be able to run our smoke tests without reliance on X11.

This still retains the existing default of running on the X11 platform, but if that isn't available then we should try working on Wayland. If neither X11 nor Wayland support is available, the "hardware" platforms are still selectable.

This will allow, for example, the following to work:

```
miral-app -kiosk -terminal mir-smoke-test-runner
```

### DIFF
--- a/tools/mir-smoke-test-runner.sh
+++ b/tools/mir-smoke-test-runner.sh
@@ -21,6 +21,12 @@ if [ -z "$XDG_RUNTIME_DIR" ]; then
   export XDG_RUNTIME_DIR=/tmp
 fi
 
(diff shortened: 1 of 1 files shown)


graysonguarino (approved):

Saviq (approved):
Because of #2434, this was tricky to test, but finally worked on Mir-on-X-on-Wayland :D

bors merge

PR #2433, opened 2022-05-23
eglstream-kms: Retain DRM device access for whole probe.

RAOF:
Oops. While fixing up #2426 I changed the scope of `device_holder`, so that
it would almost immediately go out of scope. This broke probing on the VT
console provider, as the VT console provider, and *only* the VT console
provider, drops DRM master when you release the `mir::Device`.

Fixes: #2431.


AlanGriffiths (approved):
bors merge

PR #2435, opened 2022-05-24
Miroil sync

mariogrip:


wmww:
I'm not super familiar with this stuff, but since you're trying to merge this into the main branch of the Mir source tree, it's safe to assume Mir version will always be >=2.8.0. Any ifdef for previous versions of Mir will never be used, so can be removed from the code.

AlanGriffiths:
> Any ifdef for previous versions of Mir will never be used, so can be removed from the code.

They've been in the code historically as it helped sync between 1.x and 2.x branches of miroil. Not sure if that is still useful. @mariogrip WDYT?

src/miroil/mir_prompt_session.cpp
```diff
@@ -1,4 +1,5 @@
 #include <miroil/mir_prompt_session.h>
+#include <mir/uncaught.h>
```
AlanGriffiths, line 2: Why? I see no related change(s) to the body of the file

src/miroil/mir_server_hooks.cpp
```diff
@@ -25,6 +25,8 @@
 #include <mir/input/cursor_images.h>
 #include <mir/version.h>
 
+#include <stdexcept>
+
```
AlanGriffiths, line 29: Why? I see no related change(s) to the body of the file

AlanGriffiths (changes requested):
There are essentially two sets of changes, Surface and GLBuffer

# Surface

The changes to `Surface` and `SurfaceObserverImpl` are uncontentious but largely moot. As @wmww has observed they will be dropped by the preprocessor.

# GLBuffer

The issues arise around the changes to `GLBuffer`: These look like breaking abstractions for no clear reason.

* Mir 2.x no longer supports `gl::TextureSource` making `GLTextureSourceBuffer` irrelevant and the split between `GLBuffer` and `GLTextureBuffer` pointless
* the differences exposed between `GLTextureSourceBuffer` and `GLTextureBuffer` amounts to the functions `GLTextureSourceBuffer::upload_to_texture()` and `GLTextureBuffer::tex_bind()`
  * Couldn't one function handle both uses by switching on the buffer type? (Both functions are testing the type already)
  * Supplying both functions in `GLBuffer` would be ugly, but still cleaner than what we have here
  * There may be a missing abstraction in the 1.x codebase around how textures are bound, but I don't think splitting out `GLTextureSourceBuffer` and `GLTextureBuffer` is the way to fix this.
* If we are to have a heirarchy based on `GLBuffer`, then it ought to suppress Copy/Assign as we don't want slicing

There are preexisting issues with `GLBuffer` that might play into this, so I'll mention them:

* No default constructor - that's usually wrong for a value type
* the overloaded `reset()` methods look odd: what is the advantage over assigning with a newly constructed `GLBuffer`?

AlanGriffiths:
As you will know I did some work earlier in the year refining QtMir code and moving it to Miroil. These changes are things like removing Qt dependencies and fixing the naming style to match Mir's.

This PR reverts some of that cleanup and copies an earlier version of the QtMir code to Miroil unchanged.

My preference, in the absence of any rationale for something different, would be for my QtMir changes to be adopted.

As there has been no response to the review comments questioning the rationale for these changes, I'm closing the PR

src/miroil/mirbuffer.cpp
```diff
@@ -26,58 +27,44 @@
...
-    {
-        glGenTextures(1, &m_textureId);
-        m_isOldTex = true;
-    }
+    // We would like to use gl::Texture, but if we cant, fallback to gl::textureSource
+    if (!dynamic_cast<mir::graphics::gl::Texture*>(buffer->native_buffer_base()))
+        usingTextureSource = true;
 
-    m_inited = true;
+    qDebug() << "Mir buffer is" << (usingTextureSource ? "gl:TextureSource (old)" : "gl:Texture (new)");
```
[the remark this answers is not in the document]
mariogrip: also https://gitlab.com/ubports/development/core/qtmir/-/blob/f04dbcff29e7cb7e7cd84749e30028d00ba1ddbb/src/platforms/mirserver/miral/mirbuffer.cpp#L35

Issue #2434, opened 2022-05-24
Failing to run nested on GNOME/Wayland


Saviq:
I've been trying to test #2432, but failed to run any bit of Mir on 22.04 on GNOME in a Wayland session.

With stable, it segfaults in `wl_proxy_get_version`:
```gdb
$ env -u DISPLAY -u WAYLAND_DISPLAY MIR_SERVER_WAYLAND_HOST=${WAYLAND_DISPLAY} gdb -args mir_demo_server
# ...
#0  0x00007ffff72ade94 in wl_proxy_get_version () at /lib/x86_64-linux-gnu/libwayland-client.so.0
#1  0x00007ffff574861a in wl_shell_get_shell_surface (surface=0x5555555ff010, wl_shell=0x0) at /usr/include/wayland-client-protocol.h:3024
        id = <optimized out>

            shell_surface_listener = {ping = 0x7ffff57483c0 <_FUN(void*, wl_shell_surface*, uint32_t)>, configure = 0x7ffff5748260 <_FUN(void*, wl_shell_surface*, uint32_t, int32_t, int32_t)>, popup_done = 0x7ffff5748270 <_FUN(void*, wl_shell_surface*)>}
        size = <optimized out>
#2  mir::graphics::wayland::DisplayClient::Output::for_each_display_buffer(std::function<void (mir::graphics::DisplayBuffer&)> const&) (this=0x555555636690, f=...)
    at ./src/platforms/wayland/displayclient.cpp:305

            shell_surface_listener = {ping = 0x7ffff57483c0 <_FUN(void*, wl_shell_surface*, uint32_t)>, configure = 0x7ffff5748260 <_FUN(void*, wl_shell_surface*, uint32_t, int32_t, int32_t)>, popup_done = 0x7ffff5748270 <_FUN(void*, wl_shell_surface*)>}
        size = <optimized out>
# ...
```
[gdb_stable.txt](https://github.com/MirServer/mir/files/8764407/gdb_stable.txt)
[mir_stable.txt](https://github.com/MirServer/mir/files/8764552/mir_stable.txt)

With the RC, it fails to compile a shared and then segfaults as above:
```plain
[2022-05-24 18:18:55.197567] < - ERROR - > : failed to create BasicScreenShooter: ./src/renderers/gl/renderer.cpp(232): Throw in function static GLuint mir::renderer::gl::Renderer::ProgramFactory::compile_shader(GLenum, const GLchar*)
[...]

PR #2437, opened 2022-05-24
Port Wayland platform to xdg-shell

wmww:
I opted to manually invoke wayland-scanner because I didn't see any justification for complicating the build when these files may never need to be updated and should definitely never change without manual action. Unfortunately this PR alone not seem to fully address #2434. When I run `WAYLAND_DISPLAY=mir ./bin/miral-shell --platform-display-libs mir:wayland --wayland-host wayland-0` on Sway I get the ScreenShooter error:
```
[2022-05-24 16:27:36.461300] < - ERROR - > : failed to create BasicScreenShooter: /home/wmww/code/mir/src/renderers/gl/renderer.cpp(232): Throw in function static GLuint mir::renderer::gl::Renderer::ProgramFactory::compile_shader(GLenum, const GLchar*)
Dynamic exception type: boost::wrapexcept<std::system_error>
std::exception::what: Failed to create shader: GL_INVALID_OPERATION (0x502)
```
Which should be non-fatal but is a bad sign, and then it deadlocks. I haven't figured out what's going on yet. Mir-on-Wayland still works when run on Mir-with-X.


Saviq:
> I get the ScreenShooter error:

That'd be #2440

RAOF:
This works for me on GNOME Shell, with the minor nit that Shell believes the Mir window to be unresponsive (and so pops up the “this application is unresponsive” dialog) because it's not responding to pings.

AlanGriffiths:
I can't speak for Sway, but I've merged this with trunk and that works as expected with both Mutter and egmde on 22.04:
```
WAYLAND_DISPLAY=wayland-99 cmake-build-debug/bin/miral-app --wayland-host $WAYLAND_DISPLAY
```

AlanGriffiths (changes requested):
Mostly sane. But I think there's some more plumbing needed.

AlanGriffiths:
Another thing that is probably worth investigating while touching this code is [[Mir-on-Wayland] On GNOME/Wayland all windows open on first monitor](https://github.com/MirServer/mir/issues/1903)

wmww:
Hmm, doesn't seem to fix the "not responding" problem. I can't tell in the Wayland logs were GNOME is getting the impression our app isn't responding.

RAOF:
> Hmm, doesn't seem to fix the "not responding" problem. I can't tell in the Wayland logs were GNOME is getting the impression our app isn't responding.

It's not sending [`xdg_wm_base::ping`](https://wayland.app/protocols/xdg-shell#xdg_wm_base:event:ping) and expecting us to `pong`?

wmww:
Actually, I think I may have only been seeing that when MIr was stuck in GDB (which is reasonable), but I also don't seem to see it on old iterations. Not sure if the problem is solved or not.

AlanGriffiths (approved):
I no longer see the "not responding" problem (on GNOME/20.04 ATM). But, if I read the spec correctly, we need to "ack" `xdg_toplevel_listener.configure`:

```
	 * Clients must send an ack_configure in response to this event.
	 * See xdg_surface.configure and xdg_surface.ack_configure for
	 * details.
```

Anyway, approving pending further investigation

src/platforms/wayland/displayclient.cpp
```diff
@@ -294,17 +308,22 @@ void mgw::DisplayClient::Output::done(void* data, struct wl_output* /*wl_output*
...
 {
-    if (!window)
+    if (!shell_surface)
     {
-        static wl_shell_surface_listener const shell_surface_listener{
-            [](void*, wl_shell_surface* shell_surface, uint32_t serial){ wl_shell_surface_pong(shell_surface, serial); },
-            [](void*, wl_shell_surface*, uint32_t, int32_t, int32_t){},
-            [](void*, wl_shell_surface*){}
+        static xdg_surface_listener const shell_surface_listener{
+            [](void*, auto xdg_surface, auto serial) { xdg_surface_ack_configure(xdg_surface, serial); },
```
wmww, line 314: The configure is acked here. The protocol is worded a bit weird, but the `.ack_configure` should be in response to `xdg_surface.configure` (which commits the new state), not the preceding `xdg_toplevel.configure` (which only sets the pending state).

src/platforms/wayland/displayclient.cpp
```diff
@@ -294,17 +308,22 @@ void mgw::DisplayClient::Output::done(void* data, struct wl_output* /*wl_output*
...
         };
+        shell_surface = xdg_wm_base_get_xdg_surface(owner->shell, surface);
+        xdg_surface_add_listener(shell_surface, &shell_surface_listener, this);
 
-        window = wl_shell_get_shell_surface(owner->shell, surface);
-        wl_shell_surface_add_listener(window, &shell_surface_listener, this);
-        wl_shell_surface_set_fullscreen(window, WL_SHELL_SURFACE_FULLSCREEN_METHOD_SCALE, 0, output);
+        static xdg_toplevel_listener const shell_toplevel_listener{
```
graysonguarino, line 319: Should this brace be on its own line? Can't check the style guide because the docs seem to be down right now.

Besides this, LGTM.

wmww: I'm not sure if/what the style guide says about it, but when we're initializing a value we generally put the brace on the same line as the value (unlike declaring functions, classes and structs)

graysonguarino:
bors merge

PR #2439, opened 2022-05-24
Wayland platform: improve failed to connect error

wmww:
This was giving me an "EGL_SUCCESS" error in the common case that the Wayland host is set to a display that does not currently exist. I see no reason why this would ever be an EGL error, or why the exception should be thrown so far from were the error can initially be detected. This PR drops gives the relevant context of the display name it tried to connect to.

### DIFF
--- a/src/platforms/wayland/wayland_display.cpp
+++ b/src/platforms/wayland/wayland_display.cpp
@@ -17,8 +17,8 @@
 #include "wayland_display.h"
 
 #include <mir/options/option.h>
+#include <mir/fatal.h>
 #include <boost/throw_exception.hpp>
-#include <optional>
 
 namespace mpw = mir::platform::wayland;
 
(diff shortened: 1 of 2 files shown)


src/platforms/wayland/wayland_display.cpp
```diff
@@ -46,10 +48,26 @@ void mpw::add_connection_options(boost::program_options::options_description& co
...
+    }
+    else if (wayland_host)
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "Failed to connect to Wayland display '" + wayland_host.value() + "'"));
+    }
+    else
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error("Failed to connect to default Wayland display"));
+    }
```
AlanGriffiths, line 70: I think you're started pulling one of those threads that unravel things.

We can't get here: if `wayland_host` is `nullopt`, then `probe_graphics_platform()` returns `nullopt` (and not an "unsupported" platform that can be manually selected).

As a consequence we never try to connect, even if we manually request the platform.

AlanGriffiths: @RAOF you've been reworking this. In your opinion, should `probe_rendering_platform`/`probe_display_platform` return a `SupportedDevice` with `PlatformPriority::unsupported` that can be manually selected? Or should manual selection be able to call a platform with no `SupportedDevice`(s)?

RAOF: Those are good questions.

I think that part of the answer is going to have to be a change in the manual platform selection - now that we're binding platforms to specific devices we're going to want to be able to manually select platforms for specific devices

For the first part, though, the probes should return a `SupportedDevice` with `PlatformPriority::unsupported`, and the hardware platforms currently do.

The X11 platform returns `::hosted` if and only if it can get an X11 display, so the “what happens if probing doesn't return a device but it's manually specified” isn't an interesting question; it fails, but it was going to fail anyway. If the wayland platform might succeed even though probing fails, it should be returning a `SupportedDevice` with `::unsupported` priority.

AlanGriffiths: Thanks @RAOF, that makes more sense than the code.

I think that the case where we can manually select the Wayland backend and for it to succeed when `--wayland-host` isn't supplied is fragile enough to _intentionally_ NOT support it.

(For reference, it needs `WAYLAND_HOST` to be unset (so that Mir can chose for itself) and the host to be the default of `wayland-0`.

@wmww I think we should `fatal_error` as soon as we detect `wayland-host` is not supplied (and not try to work regardless).

AlanGriffiths (changes requested):

AlanGriffiths (approved):
bors merge

wmww:
bors retry

PR #2443, opened 2022-05-26
Fix OSK on Electron apps

graysonguarino:
Electron applications run with  `--enable-features=UseOzonePlatform,WaylandWindowDecorations --ozone-platform=wayland --enable-wayland-ime` need to bind to `zwp_text_input_manager_v1` to properly use Squeekboard/ubuntu-frame-osk. This pull request contains an implementation of that protocol to enable OSK use. Most code is ported directly from `zwp_text_input_manager_v2`.


wmww:
Does electron really respect `GTK_IM_MODULE=wayland`? If it did, I would expect it to use the GTK implementation and thus not need a different protocol.

graysonguarino:
> Does electron really respect `GTK_IM_MODULE=wayland`? If it did, I would expect it to use the GTK implementation and thus not need a different protocol.

`GTK_IM_MODULE=wayland` is not necessary. I had this in my testing script from an earlier iteration and forgot to remove it.

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -0,0 +1,421 @@
...
+    {
+        send_enter_event(surface->resource);
+    }
+    else
+    {
+        deactivate(nullptr); // TODO - Is this right?
+    }
+}
+
+void TextInputV1::activate(wl_resource *seat, wl_resource *surface)
```
wmww, line 320: I'd call `seat` `seat_resource` or something (or change the name of `TextInputV1::seat` to avoid the explicit `this` below,

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -0,0 +1,421 @@
...
+{
+    (void)surface;
+
+    auto const wl_seat = mf::WlSeat::from(seat);
+    if (!wl_seat)
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error("failed to resolve WlSeat activating TextInputV1"));
+    }
+    this->seat = std::optional<mf::WlSeat*>(wl_seat);
+    this->seat.value()->add_focus_listener(client, this);
```
wmww, line 330: Why not `wl_seat->add_focus_listener(...)`

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -0,0 +1,421 @@
...
+
+void TextInputV1::show_input_panel()
+{
+    commit_count++;
+    pending_state.emplace(); // TODO - This makes sure that pending_state is ALWAYS true, making the next check redundant. Consult with Sophie.
+
+    if (pending_state && current_surface)
+    {
+        auto const new_serial = ctx->text_input_hub->set_handler_state(handler, on_new_input_field, *pending_state);
+        state_serials.push_back({commit_count, new_serial});
```
wmww, line 354: It appears this is incorrect, as is the code you copied it from. v3 uses the commit count as a serial, but both v2 and v1 use an explicit serial given by the client. This might mean you need to move this logic to `commit_state()`? Not sure.

wmww: v2 fixed in #2444. You might be able to move all the logic here to `commit_state()` and leave `show_input_panel()` and `hide_input_panel()` empty (as in v2).

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -0,0 +1,421 @@
...
+{
+    std::shared_ptr<mir::Executor> const wayland_executor;
+    std::shared_ptr<ms::TextInputHub> const text_input_hub;
+};
+
+class TextInputManagerV1Global
+    : public mw::TextInputManagerV1::Global
+{
+public:
+    TextInputManagerV1Global(wl_display* display, std::shared_ptr<TextInputV1Ctx> ctx);
```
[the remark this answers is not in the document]
AlanGriffiths: And I'm recommending we pass by value. But in this case, we should be passing `unique_ptr` anyway.

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -0,0 +1,421 @@
...
+public:
+    TextInputV1(wl_resource* resource, std::shared_ptr<TextInputV1Ctx> const& ctx);
+    ~TextInputV1();
+
+private:
+    struct Handler : ms::TextInputChangeHandler
+    {
+        Handler(TextInputV1* const text_input, std::shared_ptr<mir::Executor> const wayland_executor)
+            : text_input{text_input},
+              wayland_executor{wayland_executor}
```
AlanGriffiths, line 148: Here, we have problematic top-level const, which prevents `wayland_executor{std::move(wayland_executor)}`

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -0,0 +1,421 @@
...
+    void commit_state(uint32_t serial) override;
+    void invoke_action(uint32_t button, uint32_t index) override;
+    /// @}
+};
+
+auto mf::create_text_input_manager_v1(
+    wl_display *display,
+    std::shared_ptr<Executor> wayland_executor,
+    std::shared_ptr<scene::TextInputHub> text_input_hub)
+-> std::shared_ptr<wayland::TextInputManagerV1::Global>
```
AlanGriffiths, line 214: We're not _sharing_ ownership here. So return a `unique_ptr`.

AlanGriffiths (changes requested):
Sorry that @wmww and I are again debating on one of your PRs. (I thought that had been settled)

But see, for example:

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f26-use-a-unique_ptrt-to-transfer-ownership-where-a-pointer-is-needed

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f27-use-a-shared_ptrt-to-share-ownership

src/server/frontend_wayland/text_input_v1.cpp
graysonguarino: It seems that `commit_state()` never gets called by Electron. Additionally, `commit_state()` seems to be misleadingly named.

This is the behavior of `commit_state()` in v1 according to the protocol:
`used to identify the known state`

This is opposed to `update_state()` in v2:
```
Allows to atomically send state updates from client.

This request should follow after a batch of state updating requests
like set_surrounding_text, set_content_type, set_cursor_rectangle and
set_preferred_language.
...
```
It seems their behavior is different. Any advice on how to proceed?

AlanGriffiths: > This is the behavior of `commit_state()` in v1 according to the protocol:
> `used to identify the known state`

"New serials are sent by the text input in the commit_state request and are used by the input method to indicate the known text input state in events like preedit_string, commit_string, and keysym. The text input can then ignore events from the input method which are based on an outdated state (for example after a reset)."

So it sounds like it is yet another way to manage the serial.

src/server/frontend_wayland/text_input_v1.cpp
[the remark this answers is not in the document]
graysonguarino: I have retained pass-by-value but migrated to `unique_ptr`

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -143,7 +143,7 @@ class TextInputV1
 private:
     struct Handler : ms::TextInputChangeHandler
     {
-        Handler(TextInputV1* const text_input, std::shared_ptr<mir::Executor> const wayland_executor)
+        Handler(TextInputV1* const text_input, std::shared_ptr<mir::Executor> wayland_executor)
             : text_input{text_input},
               wayland_executor{wayland_executor}
```
AlanGriffiths, line 148: Unnecessary copy
```suggestion
              wayland_executor{std::move(wayland_executor)}
```

wmww:
Since this extension doesn't introduce any security concerns (it doesn't allow clients to do anything that might interfere with or spy on other clients), this extension can be added to `get_standard_extensions()` in src/server/frontend_wayland/wayland_default_configuration.cpp so it's on by default.

wmww (approved):
Seems to work with Chromium 102 at least

bors r+

PR #2444, opened 2022-05-26
Text Input v2: do not use commit count as serial

wmww:
It appears Qt, our only client I know of that uses this protocol, doesn't care about the serial. Both previous and new version seem to work with it, but better to be correct.


AlanGriffiths (approved):
Seems reasonable

AlanGriffiths:
bors merge

PR #2446, opened 2022-05-31
Clean up debian packaging

RAOF:
We have a bunch of unnecessary divergence between the archive packages and the packaging in GitHub. Drop (most of) it!


AlanGriffiths (approved):
I'm curious about what necessary divergence exists and why. ;)

But _this_ looks sensible!

bors merge

PR #2448, opened 2022-06-01
Wayland platform size from configure

### LINKED ISSUE #2447: Mir-on-Wayland doesn't support fractional scaling
To repro:
- Log in to GNOME shell
- Enable fractional scaling in GNOME display settings
- Set scale to 150%
- Run Mir-on-Wayland (ex `WAYLAND_DISPLAY=mir DISPLAY=no MIR_SERVER_WAYLAND_HOST=wayland-0 miral-shell`)

Expected: Mir takes up the whole screen

Actual: Mir only takes up 75% of the screen in each direction

This could be blamed on a number of things including the Wayland protocol, but in my opinion the ideal fix is to the size from the `.configure` event instead of the output size. This will also make supporting Windowed mode on Wayland easier.

wmww:
Fixes #2447, based on top of #2437 (thus draft PR until that lands). This is also a step towards being able to run the Wayland platform in windowed mode.

### DIFF
--- a/src/platforms/wayland/displayclient.cpp
+++ b/src/platforms/wayland/displayclient.cpp
@@ -68,6 +68,8 @@ public:
     EGLContext eglctx{EGL_NO_CONTEXT};
     EGLSurface eglsurface{EGL_NO_SURFACE};
 
+    std::optional<geometry::Size> pending_toplevel_size;
+    bool surface_has_been_configured{false};
     std::function<void(Output const&)> on_done;
 
     // wl_output events
@@ -89,7 +91,7 @@ public:
     void surface_configure(uint32_t serial);
 
     // DisplaySyncGroup implementation
-    void for_each_display_buffer(std::function<void(DisplayBuffer&)> const& /*f*/) override;
+    void for_each_display_buffer(std::function<void(DisplayBuffer&)> const& f) override;
     void post() override;
     std::chrono::milliseconds recommended_sleep() const override;
 
@@ -150,6 +152,11 @@ mgw::DisplayClient::Output::Output(
     dcout.scale = 1.0;
     dcout.form_factor = MirFormFactor::mir_form_factor_monitor;
     dcout.gamma_supported = MirOutputGammaSupported::mir_output_gamma_unsupported;
+
+    while (!surface_has_been_configured)
+    {
+        wl_display_roundtrip(owner->display);
+    }
 }
 
 mgw::DisplayClient::Output::~Output()
@@ -277,23 +284,6 @@ void mgw::DisplayClient::Output::scale(int32_t factor)
 
 void mgw::DisplayClient::Output::done()
 {
-    on_done(*this);
-}
-
-void mgw::DisplayClient::Output::toplevel_configure(int32_t width, int32_t height, wl_array* states)
-{
-    (void)width;
-    (void)height;
-    (void)states;
-}
-
-void mgw::DisplayClient::Output::surface_configure(uint32_t serial)
-{
-    xdg_surface_ack_configure(shell_surface, serial);
-}
-
-void mgw::DisplayClient::Output::for_each_display_buffer(std::function<void(DisplayBuffer & )> const& f)
-{
     if (!shell_surface)
     {
         static xdg_surface_listener const shell_surface_listener{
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
> [Saviq](https://github.com/Saviq) changed the base branch from main to wayland-platform-xdg-shell 1 hour ago

That doesn't work, because this isn't (so far as GitHub is concerned) "based on top of https://github.com/MirServer/mir/pull/2437":

> Conflicting files
src/platforms/wayland/displayclient.cpp

wmww:
Rebased, should work now

wmww:
Fixes #2447

src/platforms/wayland/displayclient.cpp
```diff
@@ -123,9 +118,84 @@ static EGLint const ctxattribs[] =
...
+    dcout.power_mode = mir_power_mode_on;
+    dcout.orientation = MirOrientation::mir_orientation_normal;
+    dcout.scale = 1.0;
+    dcout.form_factor = MirFormFactor::mir_form_factor_monitor;
+    dcout.gamma_supported = MirOutputGammaSupported::mir_output_gamma_unsupported;
+
+    while (!surface_has_been_configured)
+    {
+        wl_display_roundtrip(owner->display);
+    }
```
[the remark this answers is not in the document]
AlanGriffiths: On third thoughts, would we be better with this in one place? E.g. at the end of `mgw::DisplayClient::DisplayClient`

```c++
wait_for_configure:
    for (auto const& pair: bound_outputs)
    {
        if (!pair.second->surface_has_been_configured)
        {
            wl_display_roundtrip(display);
            goto wait_for_configure;
        }
    }
```

src/platforms/wayland/displayclient.cpp
```diff
@@ -224,117 +283,76 @@ void mgw::DisplayClient::Output::mode(
...
         wl_surface_set_buffer_scale(surface, round(dcout.scale));
-        wl_display_dispatch(owner->display);
+        wl_surface_commit(surface);
 
-        auto const& size = dcout.modes[dcout.current_mode_index].size;
+        // We call on_done() in the surface configure event handler
+        while (!surface_has_been_configured)
+        {
+            wl_display_roundtrip(owner->display);
+        }
```
AlanGriffiths, line 316: Do we need this both here and in the constructor? I would expect `surface_configure()` to be called before `done()`

src/platforms/wayland/displayclient.cpp
```diff
@@ -349,7 +367,7 @@ auto mgw::DisplayClient::Output::recommended_sleep() const -> std::chrono::milli
 
 auto mgw::DisplayClient::Output::view_area() const -> geometry::Rectangle
 {
-    return dcout.extents();
+    return {dcout.extents().top_left, dcout.extents().size};
```
AlanGriffiths, line 370: Why?

src/platforms/wayland/displayclient.cpp
```diff
@@ -123,9 +118,84 @@ static EGLint const ctxattribs[] =
...
+mgw::DisplayClient::Output::Output(
+    wl_output* output,
+    DisplayClient* owner,
+    std::function<void(Output const&)> on_constructed,
+    std::function<void(Output const&)> on_change) :
+    output{output},
+    owner{owner},
+    surface{wl_compositor_create_surface(owner->compositor)},
+    on_done{[this, on_constructed = std::move(on_constructed), on_change=std::move(on_change)]
+        (Output const& o) mutable { on_constructed(o), on_done = std::move(on_change); }}
```
AlanGriffiths, line 130: We're now passing the thing as both `on_constructed` and `on_change`. I think we can simplify this further.

AlanGriffiths (changes requested):
This seems to work sensibly, but could be simpler and more efficient.

src/platforms/wayland/displayclient.cpp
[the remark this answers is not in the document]
AlanGriffiths: On fourth thoughts, we could avoid "goto":

```c++
    while (end(bound_outputs) != find_if(begin(bound_outputs),
                                         end(bound_outputs),
                                         [](auto const& pair) { return !pair.second->surface_has_been_configured; }))
    {
        wl_display_roundtrip(display);
    }
```

src/platforms/wayland/displayclient.cpp
wmww: Oops, remnant of previous iteration

src/platforms/wayland/displayclient.cpp
wmww: Since we're always finishing construction by the end of the `DisplayClient` constructor, we don't need `on_constructed()` at all anymore.

PR #2451, opened 2022-06-05
[autopkgtest] Add file missing from initial autopkgtest commit.

RAOF:
Oops. The autopkgtest calls this CMake snippet, so it should be in the tree!

### DIFF
--- a/tests/acceptance-tests/wayland/capture_command_output.cmake
+++ b/tests/acceptance-tests/wayland/capture_command_output.cmake
@@ -1,0 +1,20 @@
+string(REPLACE ";" ":" TESTS_TO_QUERY_STR "${TESTS_TO_QUERY}")
+  
+execute_process(
+  COMMAND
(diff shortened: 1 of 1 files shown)


AlanGriffiths (commented):
I'm not clear where or how this is invoked. (Nor how we would detect that it broke in future)

Saviq:
> I'm not clear where or how this is invoked. (Nor how we would detect that it broke in future)

The Ubuntu infrastructure runs this:
https://github.com/MirServer/mir/blob/main/debian/tests/wlcs

As defined by:
https://github.com/MirServer/mir/blob/main/debian/tests/control

We'd need to build (we do) && install (we don't) packages in CI and run the tests.

Saviq (commented):
@RAOF I can't see any reference to this file anywhere, and naively running `sh debian/tests/wlcs`, having installed the dependencies, seems to work?

And I've not found this file anywhere else…

RAOF:
Oh, whoops. I mistook this for `PrintExcludedTests.cmake`, which I *did* add, and *would* cause `sh debian/tests/wlcs` to fail if we didn't.

PR #2453, opened 2022-06-06
mtf: Add mmap() interposer

RAOF:
The new Platform API tests need to add the ability of `MockDRM` to intercept `mmap()` calls. Take the existing `open()` interceptor, file off the serial numbers, and use it for `mmap()`, too.


AlanGriffiths:
bors merge

Issue #2457, opened 2022-06-06
Wayland platform display configuration racy


wmww:
Like #2456, we should only be making changes to displays in `Display::configure()`. This one is potentially worse because if an underlying `wl_output` disappears we will destroy the display buffer, but it may continue to be rendered on, resulting in a use-after-free.

PR #2459, opened 2022-06-07
Wayland platform: do not delete outputs until display configure

wmww:
Mitigates #2457. Just a small change that doesn't refactor the architecture, but should prevent the obvious use-after-free.


AlanGriffiths (commented):
I don't think this is any more right than the existing code.

The configuration shouldn't be changed in `~DisplayClient()`, but rather in a `configure()` call resulting from a "hardware" change.

Is the intent to avoid a crash without fixing the code?

AlanGriffiths (approved):
> This is a much smaller change that avoids a potential crash

OK

bors r+

PR #2464, opened 2022-06-10
performance_test: record renderer and mode

Saviq:


include/common/mir/logging/logger.h
```diff
@@ -58,6 +59,7 @@ class Logger
     virtual ~Logger() = default;
     Logger(const Logger&) = delete;
     Logger& operator=(const Logger&) = delete;
+    void format_message(std::ostream&, Severity, const std::string&, const std::string&);
```
AlanGriffiths, line 62: I don't think this interface is the correct place for formatting function that assumes streamed output.

This would not be needed, for example, by a logger that outputs to a database.

I would have no objection to a new `StreamingLogger` class (or a standalone function) to share implementation between `DumbConsoleLogger` and `FileLogger`.

Saviq: I didn't want to add new public API, but ACK. Will fix.

include/test/mir/test/doubles/file_logger.h
```diff
@@ -0,0 +1,47 @@
...
+class FileLogger : public mir::logging::Logger
+{
+public:
+    FileLogger(const std::string& filename, bool std_too);
+
+protected:
+    void log(mir::logging::Severity severity, const std::string& message, const std::string& component) override;
+
+private:
+    std::unique_ptr<std::ofstream> out;
```
AlanGriffiths, line 40: Why use a `unique_ptr`?

include/test/mir/test/doubles/file_logger.h
```diff
@@ -0,0 +1,47 @@
...
+namespace mir
+{
+namespace test
+{
+namespace doubles
+{
+class FileLogger : public mir::logging::Logger
+{
+public:
+    FileLogger(const std::string& filename, bool std_too);
```
AlanGriffiths, line 34: This signature seems somewhat odd and limiting.

1. Should this constructor be responsible for opening the file? Or should an `ofstream` be moved in? (That allows the output stream to be constructed in other ways)
2. The `std_too` flag mostly replicates `DumbConsoleLogger` - that duplication is unnecessary. (And what about `glog_too`?)
3. Shouldn't it be possible to log different severities to different files?

Re 2 I think _Composite_ pattern would be a better way of combining multiple logging sinks.

AlanGriffiths (changes requested):
Mostly issues with the redesign of the Logger hierarchy

include/common/mir/logging/logger.h
AlanGriffiths: This _is_ a new public API. ;)

include/common/mir/logging/logger.h
```diff
@@ -19,6 +19,7 @@
 
 #include <memory>
 #include <string>
+#include <ostream>
```
AlanGriffiths, line 22:
```suggestion
#include <iosfwd>
```

include/test/mir/test/doubles/file_logger.h
Saviq: 2. Resolved.

include/test/mir/test/doubles/multi_logger.h
```diff
@@ -0,0 +1,47 @@
...
+namespace mir
+{
+namespace test
+{
+namespace doubles
+{
+class MultiLogger : public mir::logging::Logger
+{
+public:
+    void add(std::shared_ptr<mir::logging::Logger> logger);
```
AlanGriffiths, line 35: A mutable `Logger` leaves a lot of potential for threading issues.

How about:
```
    MultiLogger(std::initializer_list<std::shared_ptr<logging::Logger>> loggers);
```

(And, maybe, `unique_ptr` is better)

include/test/mir/test/doubles/file_logger.h
```diff
@@ -0,0 +1,46 @@
...
+class FileLogger : public mir::logging::Logger
+{
+public:
+    FileLogger(const std::string& filename);
+
+protected:
+    void log(mir::logging::Severity severity, const std::string& message, const std::string& component) override;
+
+private:
+    std::unique_ptr<std::ofstream> out;
```
AlanGriffiths, line 40: > alan_g as for the unique_ptr<ofstream>, it was for differentiating between the first time streaming failed (when I printed the error) and subsequent ones (when I wanted to stay silent)
But with the composite pattern that will go away since the MultiLogger can drop the failing logger altogether

include/test/mir_test_framework/async_server_runner.h
```diff
@@ -49,6 +49,9 @@ class AsyncServerRunner
 
     mir::Server server;
 
+protected:
+    std::string output_filename;
```
AlanGriffiths, line 53: If you're going to give derived classes direct access, at least make it const.

include/test/mir/test/doubles/multi_logger.h
```diff
@@ -0,0 +1,47 @@
...
+#include <string>
+#include <vector>
+
+namespace mir
+{
+namespace test
+{
+namespace doubles
+{
+class MultiLogger : public mir::logging::Logger
```
AlanGriffiths, line 32: I think `MultiLogger` (and possibly `FileLogger`) could sensibly be added to mircommon

AlanGriffiths (changes requested):
More nitpicking

include/test/mir/test/doubles/file_logger.h
Saviq: 1. Resolved.
3. I think is out of scope (although could mean `DumbConsoleLogger` would be a special case of `StreamLogger`)

tests/mir_test_doubles/file_logger.cpp
```diff
@@ -25,23 +25,23 @@
...
-void mtd::StreamLogger::log(mir::logging::Severity severity,
+void mtd::FileLogger::log(mir::logging::Severity severity,
                           const std::string& message,
                           const std::string& component)
 {
-    if (!out->good())
+    if (!file.good())
     {
-        BOOST_THROW_EXCEPTION(std::runtime_error{"Failed to write to log stream"});
+        BOOST_THROW_EXCEPTION(std::runtime_error{"Failed to write to log file"});
```
AlanGriffiths, line 43: Another thought: It really a good idea to throw exceptions when logging fails?

Maybe this is arguably OK for test code, but not a good precedent.

include/test/mir/test/doubles/file_logger.h
```diff
@@ -14,12 +14,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef MIR_TEST_DOUBLES_STREAM_LOGGER_H_
-#define MIR_TEST_DOUBLES_STREAM_LOGGER_H_
+#ifndef MIR_TEST_DOUBLES_FILE_LOGGER_H_
+#define MIR_TEST_DOUBLES_FILE_LOGGER_H_
```
AlanGriffiths, line 18: Why is this in `mir_test_doubles`? Isn't `mir_test_framework` more appropriate?

AlanGriffiths (changes requested):
And more nits

include/common/mir/logging/logger.h
```diff
@@ -62,6 +63,7 @@ class Logger
 
 void log(Severity severity, const std::string& message, const std::string& component);
 void set_logger(std::shared_ptr<Logger> const& new_logger);
+void format_message(std::ostream&, Severity, const std::string&, const std::string&);
```
RAOF, line 66: The parameters in an interface should probably be named :wink: 

Also, we spell `const std::string&` as `std::string const&`

tests/mir_test_doubles/multi_logger.cpp
```diff
@@ -0,0 +1,51 @@
...
+    for (auto it = loggers.begin(); it != loggers.end();)
+    {
+        try
+        {
+            (*it)->log(severity, message, component);
+        }
+        catch (const std::exception& e)
+        {
+            std::cerr << e.what() << std::endl;
+            it = loggers.erase(it);
```
AlanGriffiths, line 46: It is a **really bad idea** to mutate a `MultiLogger` without first acquiring a lock. (And acquiring a lock for every iteration over `loggers` is a bad idea)

include/test/mir_test_framework/async_server_runner.h
Saviq: Done.

src/common/logging/multi_logger.cpp
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+#include "mir/logging/multi_logger.h"
+
+#include <boost/throw_exception.hpp>
+
+#include <iostream>
+#include <ctime>
+#include <cstdio>
+
```
AlanGriffiths, line 25: None of these should be needed

src/common/logging/multi_logger.cpp
```diff
@@ -0,0 +1,41 @@
...
+ml::MultiLogger::MultiLogger(std::initializer_list<std::shared_ptr<Logger>> loggers)
+: loggers{loggers}
+{
+}
+
+void ml::MultiLogger::log(Severity severity,
+                          std::string const& message,
+                          std::string const& component)
+{
+    for (auto& logger : loggers)
```
AlanGriffiths, line 37:
```suggestion
    for (auto const& logger : loggers)
```

tests/mir_test_framework/async_server_runner.cpp
```diff
@@ -62,24 +62,35 @@ mtf::AsyncServerRunner::AsyncServerRunner()
     server.add_configuration_option(mtd::logging_opt, mtd::logging_descr, false);
     server.override_the_logger([&]()
         {
-            auto logger = std::make_shared<mtd::MultiLogger>();
+            std::shared_ptr<ml::Logger> result{};
```
AlanGriffiths, line 65: The postcondition would be clearer with multiple returns of non-null pointers than with multiple conditionals that "have" to initialization this before returning it.

tests/mir_test_framework/async_server_runner.cpp
```diff
@@ -62,24 +62,35 @@ mtf::AsyncServerRunner::AsyncServerRunner()
...
             }
-            catch(const std::exception& e)
+            else
             {
-                std::cerr << e.what() << std::endl;
+                std::cerr << "Failed to open log file: " << output_filename << std::endl;
+                if (!server.get_options()->get<bool>(mtd::logging_opt))
+                {
+                  result = std::make_shared<mtd::NullLogger>();
+                }
```
AlanGriffiths, line 91: So, if the file doesn't open, and `logging` is "on", then we return a null pointer? Seems wrong

AlanGriffiths (changes requested):

tests/mir_test_framework/async_server_runner.cpp
Saviq: …and logging is __off__, then we return a null logger.

src/common/logging/multi_logger.cpp
Saviq: Leftovers. Gone.

tests/mir_test_framework/async_server_runner.cpp
Saviq: Fixed.

tests/mir_test_framework/async_server_runner.cpp
```diff
@@ -79,18 +77,19 @@ mtf::AsyncServerRunner::AsyncServerRunner()
...
                 std::cerr << "Failed to open log file: " << output_filename << std::endl;
                 if (!server.get_options()->get<bool>(mtd::logging_opt))
                 {
-                  result = std::make_shared<mtd::NullLogger>();
+                    return std::make_shared<mtd::NullLogger>();
                 }
             }
-            return result;
+
+            return nullptr;
```
AlanGriffiths, line 92: I don't think we should ever return a null pointer from `the_logger()`

```suggestion
            return std::make_shared<ml::DumbConsoleLogger>();
```

AlanGriffiths (changes requested):

src/common/symbols.map
```diff
@@ -513,4 +512,5 @@ MIR_COMMON_2.9 {
     vtable?for?mir::logging::FileLogger;
     vtable?for?mir::logging::MultiLogger;
   };
+  local: *;
```
AlanGriffiths, line 515: Why? We don't do that elsewhere

Saviq: https://mir-server.io/doc/dso_versioning_guide.html#autotoc_md15 says:

> Note that the script is read top to bottom; wildcards are greedily bound when first encountered, so to avoid surprises you should only have a wildcard in the final stanza.

AlanGriffiths: Hmm:

* regenerate-miral-symbols-map.py
* src/miral/symbols.map
* src/platform/symbols.map
* src/wayland/symbols.map

Saviq: :man_shrugging: undid, then.

tests/mir_test_framework/async_server_runner.cpp
```diff
@@ -49,14 +60,36 @@ mtf::AsyncServerRunner::AsyncServerRunner()
...
+            {
+                std::cerr << "Failed to open log file: " << output_filename << std::endl;
+                if (!server.get_options()->get<bool>(mtd::logging_opt))
+                {
+                    return std::make_shared<mtd::NullLogger>();
+                }
+            }
 
-            return result;
+            return std::make_shared<ml::DumbConsoleLogger>();
```
AlanGriffiths, line 92: I know I suggested this, but probably better not to change behaviour:

```suggestion
            // If we return null then the default will be used
            return {};
```

AlanGriffiths (approved):
bors delegate

Saviq (commented):
Thanks!

bors r=AlanGriffiths,RAOF

Issue #2465, opened 2022-06-12
ABI breakage of libmircore.so.1 with v2.8.0 vs v1.8.2


JamiKettunen:
Attempting to launch [Lomiri shell](https://gitlab.com/ubports/development/core/lomiri) with `libmircore.so.1` from v2.8.0 (same, conflicting soname with v1.8.2) causes the following error:
```
lomiri: symbol lookup error: /usr/lib64/libmirplatform.so.18: undefined symbol: _ZN3mir8geometrylsERSoRKNS0_5PointE, version MIR_CORE_0.25
```
Simply replacing the shared library file with a known working one from v1.8.2 resolves the issue immediately (with everything else being libraries from v1.8.2).

Here's my x86_64 glibc samples of the libraries if they could be useful: [libmircore-shared-libraries.zip](https://github.com/MirServer/mir/files/8886501/libmircore-shared-libraries.zip)

My suggestion: bump soname of `libmircore.so.1` to `libmircore.so.2` on [main](https://github.com/MirServer/mir/tree/main), but perhaps there's something I'm missing

AlanGriffiths:
```
$ c++filt _ZN3mir8geometrylsERSoRKNS0_5PointE
mir::geometry::operator<<(std::basic_ostream<char, std::char_traits<char> >&, mir::geometry::Point const&)
```

PR #2466, opened 2022-06-13
Bump mircore soname

AlanGriffiths:
Fixes: #2465


Saviq (commented):
Shouldn't it be `debian/libmircore2.symbols` now, rather than `1`?

RAOF (approved):
LGTM.

We can get any other ABI breaking changes @wmww wants between now and release!

bors r+

PR #2468, opened 2022-06-13
Add --add-all-wayland-extensions option

wmww:
Convenient for development and debugging, as well as running in contexts where all clients are fully trusted

### DIFF
--- a/include/platform/mir/options/configuration.h
+++ b/include/platform/mir/options/configuration.h
@@ -44,6 +44,7 @@ extern char const* const x11_display_opt;
 extern char const* const x11_scale_opt;
 extern char const* const wayland_extensions_opt;
 extern char const* const add_wayland_extensions_opt;
+extern char const* const add_all_wayland_extensions_opt;
 extern char const* const drop_wayland_extensions_opt;
 extern char const* const idle_timeout_opt;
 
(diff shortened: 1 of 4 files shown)


AlanGriffiths:
> Hmm, heads up `platform/symbols.map` has `MIRPLATFORM_2.5` and `MIRPLATFORM_2.7` (no underscore), but `MIR_PLATFORM_2.8` (with underscore). Does one of those need to be fixed?

It is inconsistent, but we shouldn't change history.

We are inconsistent with these:

No underscore|Underscore
--|--
&nbsp;|`MIR_CORE...`
|`MIRAL...`
`MIRPLATFORM...`|`MIR_PLATFORM...`
`MIRWAYLAND...`|

AlanGriffiths:
We already have a lot of overlapping configuration options (not all of them useful). I don't think we need this one in the core Mir libraries as the same functionality can be implemented using the `miral::WaylandExtensions` API.

I would have no problem with that implementation being added to `example-shell-lib` and included in the example servers.

(Not blocking, awaiting further opinions)

Saviq:
How about supporting wildcards?

AlanGriffiths:
Superseded by #2498

PR #2469, opened 2022-06-13
wlr-screencopy-v1: send .damage event when required

wmww:
The `.copy_with_damage` request still isn't implemented fully (it does not wait for there to be damage or report the damaged area of the screen), but with this change it works well enough for wayvnc to work (albeit with higher than needed CPU load)


AlanGriffiths (approved):
LGTM (well, good enough)

bors++

AlanGriffiths:
> bors++

Really! What was I typing?

bors r++

PR #2470, opened 2022-06-14
Send changed area to frame posted notification

wmww:
The surface observer's `frame_posted()` notification is current given the size of the posted frame. The only use of this info seems to be in `NonLegacySurfaceChangeNotification`, where it is used in conjunction with the surface's position to determine an area of the screen that has changed. Unfortunately, the way this is done is broken in many ways. Scaled surfaces, streams with offsets from the surface and surfaces with margins will all cause it to report an incorrect damage rectangle.

With this PR, the actual area of the screen which is effected is reported, and all of the above cases are handled correctly and tested.


wmww:
This PR will allow us to correct implement damage over the screencopy protocol, which should significantly improve VNC performance.

tests/unit-tests/scene/test_basic_surface.cpp
```diff
@@ -1038,41 +1039,116 @@ TEST_F(BasicSurfaceTest, notifies_about_cursor_removal_when_empty_stream_set)
     surface.set_cursor_stream(buffer_stream, {});
 }
 
-TEST_F(BasicSurfaceTest, cursor_can_be_set_from_stream_that_started_empty)
```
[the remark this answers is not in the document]
wmww: Oops, no

tests/unit-tests/scene/test_basic_surface.cpp
```diff
@@ -1038,41 +1039,116 @@ TEST_F(BasicSurfaceTest, notifies_about_cursor_removal_when_empty_stream_set)
...
+TEST_F(BasicSurfaceTest, frame_posted_given_stream_size)
```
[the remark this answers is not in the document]
wmww: I took a stab at better names, let me know if they still need tweaking

tests/unit-tests/scene/test_basic_surface.cpp
```diff
@@ -1038,41 +1039,116 @@ TEST_F(BasicSurfaceTest, notifies_about_cursor_removal_when_empty_stream_set)
...
+        .WillByDefault(Return(geom::Size{10, 15}));
 
     surface.add_observer(mt::fake_shared(mock_surface_observer));
-    surface.set_cursor_stream(buffer_stream, {});
-    stub_buffer = std::make_shared<mtd::StubBuffer>();
-    (*frame_posted_callback)({});
+    surface.set_streams({ms::StreamInfo{buffer_stream, {}, {}}});
+
+    geom::Rectangle given_area;
+    EXPECT_CALL(mock_surface_observer, frame_posted(_, _, _)).WillOnce(SaveArg<2>(&given_area));
```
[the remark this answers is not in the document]
wmww: Done

include/test/mir/test/geometry_matchers.h
```diff
@@ -0,0 +1,40 @@
...
+{
+namespace test
+{
+
+MATCHER_P(TopLeftEq, expected_top_left, "")
+{
+    return arg.top_left == expected_top_left;
+}
+
+MATCHER_P(SizeEq, expected_size, "")
```
AlanGriffiths, line 32: These matcher names are probably a bit too generic for something in a public header: They require `arg` to be a `Rectangle` (or similar) but there's no mention of that constraint in the naming (or scope)

tests/unit-tests/scene/test_basic_surface.cpp
```diff
@@ -1075,6 +1077,116 @@ TEST_F(BasicSurfaceTest, cursor_can_be_set_from_stream_that_started_empty)
     (*frame_posted_callback)({});
 }
 
+TEST_F(BasicSurfaceTest, observer_notified_of_frame_with_correct_size_when_frame_is_posted)
+{
+    using namespace testing;
+    geom::Size stream_size{20, 30};
```
AlanGriffiths, line 1083:
```suggestion
    geom::Size const stream_size{20, 30};
```

AlanGriffiths: Also, could be moved into the fixture?

include/core/mir/proof_of_mutex_lock.h
```diff
@@ -36,6 +36,17 @@ class ProofOfMutexLock
...
+
+protected:
+    ProofOfMutexLock() {}
+};
+
+class ProofOfMutexLock::Fake : public ProofOfMutexLock
+{
+public:
+    Fake() {}
 };
```
AlanGriffiths, line 50: This is a test double isn't it? Does it really belong in a non-test header?

src/include/server/mir/scene/surface_observer.h
```diff
@@ -51,7 +51,8 @@ class SurfaceObserver
     virtual void content_resized_to(Surface const* surf, geometry::Size const& content_size) = 0;
     virtual void moved_to(Surface const* surf, geometry::Point const& top_left) = 0;
     virtual void hidden_set_to(Surface const* surf, bool hide) = 0;
-    virtual void frame_posted(Surface const* surf, int frames_available, geometry::Size const& size) = 0;
+    /// area is given in logical, global coordinates
+    virtual void frame_posted(Surface const* surf, int frames_available, geometry::Rectangle const& area) = 0;
```
AlanGriffiths, line 55: Which "area" is area? The "damage area"?

src/server/scene/basic_surface.cpp
```diff
@@ -263,16 +263,7 @@ ms::BasicSurface::BasicSurface(
...
-        {
-            if (auto const o = observers.lock())
-                o->frame_posted(this, 1, size);
-        };
-
-    for (auto& layer : layers)
-    {
-        layer.stream->set_frame_posted_callback(callback);
-    }
+    update_frame_posted_callbacks(ProofOfMutexLock::Fake{});
```
AlanGriffiths, line 266: OK now I'm confused. Why do we require `ProofOfMutexLock` if we fake it in production code?

We need at least a comment explaining or (better) code that makes the conditions clear.

AlanGriffiths (changes requested):
Test names are a lot better, but "when ... then ..." is more widely used than "... when ..." (but not a blocking issue)

And I don't like the introduction of `ProofOfMutexLock::Fake` it will get abused. BTW we don't actually need another class, just a member like:
```c++
class ProofOfMutexLock
{
public:
    static ProofOfMutexLock const trust_me_this_is_uncontended;
```

Which is a bit clearer than `...::Fake`

src/server/scene/basic_surface.cpp
wmww: I will try to improve names and documentation around this

wmww:
Since the addition to `ProofOfMutexLock` proved unexpectedly contentious, I removed it and just locked the lock in the constructor and destructor. I have no objection to refactoring `BasicSurface` to use a data-owning mutex if someone wants to do that, but please make that a separate PR from this. This one has sat in review long enough.

wmww:
Is this what #2478 fixes?
```
2022-06-17T23:37:46.8595402Z ##[group]Run snapcraft upload *.snap --release "edge/mir-pr2470"
2022-06-17T23:37:46.8595815Z [36;1msnapcraft upload *.snap --release "edge/mir-pr2470"[0m
2022-06-17T23:37:46.8650623Z shell: /usr/bin/bash -e {0}
2022-06-17T23:37:46.8650888Z ##[endgroup]
2022-06-17T23:37:47.9789738Z This login method is not longer supported                                      
2022-06-17T23:37:47.9791832Z Cannot parse config: File contains no section headers.
2022-06-17T23:37:47.9795129Z file: '<string>', line: 1
2022-06-17T23:37:48.0253140Z '***           \n'                                           
2022-06-17T23:37:48.0253954Z Full execution log: '/home/runner/.cache/snapcraft/log/snapcraft-20220617-233747.976702.log'                                                                   
2022-06-17T23:37:48.0689403Z ##[error]Process completed with exit code 1.
```

RAOF (approved):
I'm good with this now. I'll propose a PR to do the owning-mutex thing separately.

I think you do need to rebase on the fixed snapcraft bits, yes.

Saviq:
bors r=RAOF

Issue #2427, opened 2022-05-13
Screenshotting and background


AlanGriffiths:
When screenshotting, we end up transparency in the image wherever there is no (or transparent) background.

This can be seen with swaybar as used by egmde, or if there is no background painted over all or part of the screen.

PR #2471, opened 2022-06-14
renderers/gl: Clear framebuffer to opaque black, not fully-transparent.

RAOF:
The fully-transparent `glClearColor` was introduced in 2015 in commit dabbc2bc1f
titled “merge mir” (it is not a merge commit).

Since the practical result of a fully-transparent framebuffer is expensive black
on all hardware platforms (at least until transparent displays become a thing),
switch the behaviour *back* to clearing the framebuffer to black.

Fixes: #2427


RAOF:
If anyone can think of a reason why we might *want* a transparent framebuffer, I'm not particularly attached to this.

AlanGriffiths:
I wonder if this fixes #2427...

AlanGriffiths (approved):
bors merge

PR #2477, opened 2022-06-16
Tidy up eglstream probe

AlanGriffiths:


wmww (approved):
LGTM

(but also looks like code RAOF might should look at, alan can decide if that's the case)

bors delegate+

src/platforms/eglstream-kms/server/platform_symbols.cpp
```diff
@@ -163,7 +163,7 @@ auto probe_display_platform(
     }
 
     std::vector<mg::SupportedDevice> supported_devices;
-    for (auto i = 0; i <= device_count; ++i)
+    for (auto i = 0; i != device_count; ++i)
```
RAOF, line 166: I'd generally spell this as `<` rather than `!=`, but that's just out of boundless paranoia.

Sadly, that paranoia doesn't protect one from off-by-one errors. :woman_facepalming:

RAOF (approved):
's fine by me!

bors r+

PR #2481, opened 2022-06-17
Refactor scene and surface change notification classes

wmww:
For presumably historical reasons, we previously had `LegacySceneChangeNotification`, which used either `LegacySurfaceChangeNotification` or `NonLegacySurfaceChangeNotification` depending on how it was constructed. Production code only used the non-legacy path, and the non-legacy one inherited from the legacy one. This PR consolidates the surface observers into a single class, and drops the legacy prefix from both surface and scene observers. It should not have any impact on behavior.


AlanGriffiths:
bors retry

Saviq:
bors retry

PR #2483, opened 2022-06-21
scene/BasicSurface: Switch to a data-owning Mutex

RAOF:
There's a *lot* of state that's guarded by this one mutex, but that's how the
current code works.

Fixes up a couple of places in the code where these values were accidentally
accessed without the lock held.

### DIFF
--- a/include/core/mir/mutex.h
+++ b/include/core/mir/mutex.h
@@ -91,7 +91,6 @@ public:
     {
         return MutexGuard<Guarded>{std::unique_lock{mutex}, value};
     }
-
 private:
     std::mutex mutex;
     Guarded value;
--- a/include/core/mir/mutex.h
+++ b/include/core/mir/mutex.h
@@ -91,6 +91,16 @@ public:
     {
         return MutexGuard<Guarded>{std::unique_lock{mutex}, value};
     }
+
+    /**
+     * Unlock an acquired mutex
+     *
+     * This is a convenience method for cases when the mutex must be dropped
+     * but cannot easily be made to leave scope.
+     */
+    static void drop(MutexGuard<Guarded> /*to_drop*/)
+    {
+    }
 private:
     std::mutex mutex;
     Guarded value;
--- a/src/server/scene/basic_surface.h
+++ b/src/server/scene/basic_surface.h
@@ -24,6 +24,7 @@
 #include "mir/wayland/wayland_base.h"
 #include "mir/geometry/rectangle.h"
 #include "mir_toolkit/common.h"
+#include "mir/mutex.h"
 
 #include <glm/glm.hpp>
 #include <vector>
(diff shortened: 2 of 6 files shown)


include/core/mir/mutex.h
```diff
@@ -92,6 +99,15 @@ class Mutex
         return MutexGuard<Guarded>{std::unique_lock{mutex}, value};
     }
 
+    /**
+     * Unlock an acquired mutex
+     *
+     * This is a convenience method for cases when the mutex must be dropped
+     * but cannot easily be made to leave scope.
+     */
+    static void drop(MutexGuard<Guarded> /*to_drop*/)
```
wmww, line 108: The syntax for this seems a little verbose.

 Also, it's not clear to me what state the guard is left in after a move. Does it throw an error if you try to access it? Segfault?

wmww (approved):
The code ended up prettier than I was expecting, which makes me feel less ambivalent and more positive towards this change. Not super happy with `drop()` though.

include/core/mir/mutex.h
```diff
@@ -33,11 +33,18 @@ class MutexGuard
...
+    MutexGuard(MutexGuard&& from) noexcept
+        : value{from.value},
+          lock{std::move(lock)}
+    {
+        from.value = nullptr;
+    }
+
     ~MutexGuard() noexcept(false)
     {
         if (lock.owns_lock())
```
wmww, line 50: pre-existing, but why do we need to explicitly unlock the lock on destruction? Doesn't `unique_lock` already do that for us?

wmww:
I proposed #2492 against this PR to address my issues with `drop`.

include/core/mir/mutex.h
```diff
@@ -92,6 +113,15 @@ class Mutex
...
+     * Unlock an acquired mutex
+     *
+     * This is a convenience method for cases when the mutex must be dropped
+     * but cannot easily be made to leave scope.
+     */
+    static void drop(MutexGuard<Guarded> /*to_drop*/)
+    {
+    }
 private:
     std::mutex mutex;
```
AlanGriffiths, line 126: It is usual for a `mutex` to be `mutable`, the only reason that isn't needed here is there's been attempt at const-correctness which means that mutability has been propagated to the instance.

On the surface, it would appear to be possible to provide const overloads that return const `MutexGuard<Guarded const>` and so maintain const correctness.

include/core/mir/mutex.h
```diff
@@ -48,14 +55,28 @@ class MutexGuard
...
+    void drop()
+    {
+        value = nullptr;
+        lock.unlock();
     }
+
 private:
-    Guarded& value;
+    Guarded* value;
     std::unique_lock<std::mutex> lock;
```
AlanGriffiths, line 80: It is unfortunate that we need to use `std::unique_lock` and can't optimise with `std::lock_guard` where that provides the required guarantees.

AlanGriffiths (changes requested):
I really like the idea, but I want to polish it further.

include/core/mir/mutex.h
```diff
@@ -33,11 +33,18 @@ class MutexGuard
 {
 public:
     MutexGuard(std::unique_lock<std::mutex>&& lock, Guarded& value)
```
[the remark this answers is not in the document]
RAOF: Hm. So, the names were stolen from the Rust standard library; this is how `Mutex<T>` works there.

I'm not averse to renaming them, but I don't think `Secure<State>` is a better name. Particularly, it doesn't describe what it's securing the state *from*, and there are various prominent `SecureFoo` things (`SecureString` being I think the most notable) which are information-security related which this might get confused with.

This class ensures that all accesses to the contained data are coherent, in that you never see partially-updated state. Perhaps `Synchronised<State>` and `Synchronised<State>::Accessor` (moving `MutexGuard` into an inner type) for the two types?

It's also kinda important that the *implementation* is mutual-exclusion rather than, for example, atomic CAS? Is `Mutex<T>` is going to be confusing to anyone?

include/core/mir/mutex.h
```diff
@@ -92,6 +113,15 @@ class Mutex
...
 
+    /**
+     * Unlock an acquired mutex
+     *
+     * This is a convenience method for cases when the mutex must be dropped
+     * but cannot easily be made to leave scope.
+     */
+    static void drop(MutexGuard<Guarded> /*to_drop*/)
+    {
+    }
```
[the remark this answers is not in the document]
RAOF: Correct; it is vestigial.

include/core/mir/mutex.h
```diff
@@ -87,13 +102,24 @@ class Mutex
...
     }
 
+    /**
+     * Lock the mutex and return an accessor for the protected data.
+     *
+     * \return A smart-pointer-esque accessor for the contained data.
+     *         While code has access to the MutexGuard it is guaranteed to have exclusive
+     *         access to the contained data.
+     */
+    auto lock() const -> MutexGuard<Guarded const>
```
RAOF, line 117: Hm. I suspect that this will be a compile error if someone asks for `Mutex<Foo const>`, right? Do we care about doing the necessary template shenanigans to make that work? `Mutex<Foo const>` doesn't seem like a useful thing to ask for.

AlanGriffiths: I agree that `Mutex<Foo const>` "doesn't seem like a useful thing to ask for", but would there be a compile error?

include/core/mir/mutex.h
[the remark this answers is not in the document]
AlanGriffiths: Unfortunately, prior art is no guarantee of good names. None of the suggestions so far (including mine) are clearly good. Maybe `Synchronised` and `SyncLock`?

Anyway, in the absence of a clearly better alternative I won't block on these names.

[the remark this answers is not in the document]
wmww: `Synchronized<State>::lock() -> Unique<State>`? `Locked` and `SyncLock` also fine. I'm not hugely particular about the names.

src/server/scene/basic_surface.h
```diff
@@ -167,58 +168,65 @@ class BasicSurface : public Surface
...
+        struct {
+            geometry::DeltaY top;
+            geometry::DeltaX left;
+            geometry::DeltaY bottom;
+            geometry::DeltaX right;
+        } margins{};
+
+        MirFocusMode focus_mode = mir_focus_mode_focusable;
+    };
+    mir::Mutex<State> mutable mutable_state;
```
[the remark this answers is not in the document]
RAOF: I've renamed this to `synchronised_state`.

include/core/mir/mutex.h
RAOF: Drop no longer exists

AlanGriffiths (approved):
Approving, but not borsing: I'm not convinced the [naming discussion](https://github.com/MirServer/mir/pull/2483#discussion_r908587202) is over. Yet.

RAOF:
I concur.

include/core/mir/mutex.h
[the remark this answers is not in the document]
Saviq: I think I'd go for `Synchronized<State>::lock() -> Locked<State>`.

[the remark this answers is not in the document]
RAOF: Good news! It turns out we've actually *already* had this naming discussion, and `mir::Synchronised<T>` already exists! (servicing `LinuxVTProvider`).

I think we've approximate consensus on `Synchonised`/`Synchronised::Locked`, so I've updated the naming to that and folded in the exsiting, near identical  `src/include/server/mir/synchronised.h`.

src/server/console/linux_virtual_terminal.cpp
```diff
@@ -64,11 +65,11 @@ class mir::LinuxVirtualTerminal::DeviceList
         }
 
     private:
-        DeviceListIterable(Synchronised<std::vector<Device*>>::Locked&& devices)
+        DeviceListIterable(Synchronised<std::vector<Device*>>::Locked<std::vector<Device*>>&& devices)
```
RAOF, line 68: Supporting the `Synchronised<T>::lock() const` overload makes code like this marginally less nice. We could probably fix that up with some typedefs, though.

RAOF: Yup, that's easy.

If the typedefs are at all controversial, though, I'm happy to remove that commit.

include/core/mir/mutex.h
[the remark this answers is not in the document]
AlanGriffiths: Wow! The previous naming discussion #384 was remarkably similar. I guess none of us learnt from the experience.

AlanGriffiths (approved):
bors merge

Saviq:
bors retry

PR #2485, opened 2022-06-21
SurfaceChangeNotification: make threadsafe

wmww:
There's still a possibility a surface move will get processed out of order with a buffer submit, but since a surface move generates a full scene change that invalidates everything that should be fine.


AlanGriffiths (approved):
bors merge

PR #2486, opened 2022-06-21
Screencopy: wait for damage before copying frame

wmww:
Waits until the copy area has been damaged in `.copy_with_damage` request.

The protocol isn't designed very well and is not super clear in how this is supposed to be implemented. The naive way would be to start waiting for damage when the request comes in, but that would allow a client to miss frames. Instead the way this is built is to dynamically create `CaptureArea`s for each area the client captures and track damage to those areas. That way a frame can be captured immediately with the correct damage if the damage happened between one frame being captured and the following frame being requested.

It would be nice to have this tested, but we need the screenshoter to work in WLCS for that and it currently does not.

### DIFF
--- a/src/server/frontend_wayland/wlr_screencopy_v1.cpp
+++ b/src/server/frontend_wayland/wlr_screencopy_v1.cpp
@@ -38,9 +38,7 @@ namespace mg = mir::graphics;
 namespace ms = mir::scene;
 namespace geom = mir::geometry;
 
-namespace
-{
-struct FrameParams
+struct mf::WlrScreencopyDamageTracker::FrameParams
 {
     geom::Rectangle area;
     wl_resource* output;
@@ -50,7 +48,30 @@ struct FrameParams
         return area == other.area && output == other.output;
     }
 };
-}
+
+class mf::WlrScreencopyDamageTracker::Area
+{
+public:
+    explicit Area(FrameParams const& params);
+    ~Area();
+
+    /// Adds the given damage (damages everything if null), and captures the frame if needed
+    void apply_damage(std::optional<geometry::Rectangle> const& damage);
+    /// Adds a frame, and immediately captures it if there is already damage
+    void add_frame(Frame* frame);
+
+    FrameParams const params;
+
+private:
+    void capture_frame();
+
+    /// The amount of damage since the last frame was captured (should be none unless pending_frame is null)
+    DamageAmount damage_amount;
+    /// Only has meaning when damage_amount is partial
+    geom::Rectangle damage_rect;
+    /// The frame that will be captured once this capture area takes damage
+    wayland::Weak<Frame> pending_frame;
+};
 
 namespace mir
 {
@@ -83,13 +104,10 @@ class WlrScreencopyManagerV1
 {
 public:
     WlrScreencopyManagerV1(wl_resource* resource, std::shared_ptr<WlrScreencopyV1Ctx> const& ctx);
-    ~WlrScreencopyManagerV1();
 
-    void maybe_wait_for_damage(FrameParams const& params, WlrScreencopyFrameV1* frame);
+    void capture_on_damage(WlrScreencopyDamageTracker::Frame* frame);
 
 private:
-    void create_change_notifier();
-
     /// From wayland::WlrScreencopyManagerV1
     /// @{
     void capture_output(wl_resource* frame, int32_t overlay_cursor, wl_resource* output) override;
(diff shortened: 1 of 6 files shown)


src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -20,18 +20,38 @@
...
+{
+struct FrameParams
+{
+    geom::Rectangle area;
+    wl_resource* output;
+
+    auto operator==(FrameParams const& other) const -> bool
+    {
+        return area == other.area && output == other.output;
+    }
```
AlanGriffiths, line 51: It is a shame we support compilers without
```suggestion
    bool operator==(FrameParams const&) = default;
```
... also and the "spaceship operator"

src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -55,22 +76,72 @@ class WlrScreencopyManagerV1Global
...
+        none,
+        partial,
+        full,
+    };
+
+    /// Used to track damage to an area for a single manager (and thus a single client)
+    class CaptureArea
+    {
+    public:
+        CaptureArea(FrameParams const& params);
```
AlanGriffiths, line 117:
```suggestion
        explicit CaptureArea(FrameParams const& params);
```

src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -154,29 +298,145 @@ void mf::WlrScreencopyManagerV1::capture_output_region(
...
+            damage_rect = intersection;
+        }
+    }
+    else
+    {
+        // If damage amount is already full, or if damage is nullopt (and thus full damage should be applied)
+        damage_amount = DamageAmount::full;
+    }
+
+    if (pending_frame)
```
AlanGriffiths, line 341: This check is repeated in `capture_frame()` (which suggests it is NOT a precondition) and at every call site (which suggests it IS a precondition).

Maybe it should be called `capture_any_pending_frame()` (and omit the checks at call sites)

src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -55,22 +76,72 @@ class WlrScreencopyManagerV1Global
...
+        /// Only has meaning when damage_amount is partial
+        geom::Rectangle damage_rect;
+        /// The frame that will be captured once this capture area takes damage
+        wayland::Weak<WlrScreencopyFrameV1> pending_frame;
+    };
+
+    /// Only accessed from the Wayland thread
+    /// @{
+    /// Created the first time a frame from this manager calls .copy_with_damage
+    std::shared_ptr<scene::SceneChangeNotification> change_notifier;
```
AlanGriffiths, line 140: I'd be tempted to initialize (not register) `change_notifier` on construction and make it `const`.

src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -134,14 +219,73 @@ mf::WlrScreencopyManagerV1::WlrScreencopyManagerV1(
...
+    }
+    else
+    {
+        // We capture the given frame immediately, and also push an empty capture area so that if we get another
+        // capture request with the same params it will wait for damage since this frame.
+        frame->capture(std::nullopt);
+        capture_areas.emplace_back(params);
+        // If an unusually high number of capture areas have been created for some reason, clear the list rather than
+        // getting bogged down (it's ok, worst case scenario waiting for damage doesn't work)
+        if (capture_areas.size() > 100)
```
AlanGriffiths, line 253: Can capture areas ever be cancelled? Is there really a problem with large numbers? Should we simply stop adding them instead of clearing the existing ones?

AlanGriffiths (changes requested):
Mostly nitpicking about naming.

[Aside] It should be _possible_ to introduce tests for `CaptureArea` by having it take a `FrameCapture` interface (implemented by `WlrScreencopyFrameV1`)

src/server/frontend_wayland/wlr_screencopy_v1.cpp
AlanGriffiths: My reading of the protocol is that captures are requested a buffer at a time. If I have that right `capture_areas` is acting both as a list of current captures and as a cache of previous ones? (The cache is obviously sensible as requests for damage only make sense when the same parameters are used for repeated requests)

The right way to invalidate the cache in this scenario would be dropping the least recently used entries.

src/server/frontend_wayland/wlr_screencopy_v1.cpp
wmww: That would be the obvious way, but in many cases a client will will never call `.copy_with_damage`, making the overhead of setting up a change notifier and processing scene changes unnecessary.

src/server/frontend_wayland/wlr_screencopy_v1.cpp
wmww: Indeed making it a deque and popping from the front might be optimal, but needing any sort of cache invalidation is unlikely so I just went with the simplest thing that would keep it from getting bogged down.

src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -103,16 +164,166 @@ auto mf::create_wlr_screencopy_manager_unstable_v1(
...
+        // We capture the given frame immediately, and also push an empty capture area so that if we get another
+        // capture request with the same params it will wait for damage since this frame.
+        frame->capture(std::nullopt);
+        areas.push_back(std::make_unique<Area>(frame_params));
+        // If an unusually high number of capture areas have been created for some reason, clear the list rather than
+        // getting bogged down (it's ok, worst case scenario waiting for damage doesn't work)
+        if (areas.size() > 100)
+        {
+            areas.clear();
+        }
```
AlanGriffiths, line 220: Might be slightly better to have `push_back()` after the `clear()`

tests/unit-tests/frontend_wayland/test_screencopy_v1_damage_tracker.cpp
```diff
@@ -0,0 +1,465 @@
...
+    damage_area(area);
+    executor.execute();
+
+    StrictMock<MockFrame> frame{area};
+    EXPECT_CALL(frame, capture(Eq(std::nullopt)));
+    tracker.capture_on_damage(&frame);
+    executor.execute();
+}
+
+TEST_F(DamageTrackerV1, captured_with_combined_area_when_multiple_damages_happen_before_capture)
```
AlanGriffiths, line 447: The test names seem a bit strange.

`when_multiple_damages_happen_before_capture_then_captures_combined_damage_areas`?

AlanGriffiths (approved):
Some nits but...

bors delegate

Saviq:
bors retry

Saviq:
status.snapcraft.io says it's fine again.

bors retry

AlanGriffiths:
Maybe now?

bors retry

Issue #2474, opened 2022-06-14
wlcs tests flaky in CI


Saviq:
These failures are on arm{hf,64}, which run through qemu, so that might be a factor.

E.g.

https://github.com/MirServer/mir/runs/6879373164?check_suite_focus=true#step:5:9640

```
[ RUN      ] AllSurfaceTypes/TouchTest.sends_touch_up_on_surface_destroy/subsurface_at_x7_y12
[2022-06-14 11:49:25.832214] <information> mirserver: Starting
[2022-06-14 11:49:25.842273] < - debug - > mirserver: No session management supported
[2022-06-14 11:49:25.845956] <information> VT switch key handler: No VT switching support available: MinimalConsoleServices does not support VT switching
[2022-06-14 11:49:25.922347] <information> mircommon: Loading modules from: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/
[2022-06-14 11:49:25.923540] <information> mircommon: Loading module: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/graphics-wayland.so.20
[2022-06-14 11:49:25.923894] <information> mircommon: Loading module: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/graphics-eglstream-kms.so.20
[2022-06-14 11:49:25.924024] <information> mircommon: Loading module: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/graphics-gbm-kms.so.20
[2022-06-14 11:49:25.924116] <information> mircommon: Loading module: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/server-x11.so.20
[2022-06-14 11:49:25.924200] <information> mircommon: Loading module: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/input-evdev.so.8
[2022-06-14 11:49:25.924284] <information> mircommon: Loading module: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/graphics-dummy.so
[2022-06-14 11:49:25.924364] <information> mircommon: Loading module: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/input-stub.so
[2022-06-14 11:49:25.924441] <information> mircommon: Loading module: /<<PKGBUILDDIR>>/build-arm64/lib/server-modules/graphics-throw.so
[2022-06-14 11:49:25.930734] <information> mirserver: Found display driver: mir:stub-graphics (version 2.8.0)
[...]

PR #2488, opened 2022-06-22
Disable WLCS tests on LP/arm* (as they are currently flaky)

AlanGriffiths:
Workaround: #2474

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -41,6 +41,11 @@ endif
 # control, so disable it on non-Ubuntu builds
 ifeq ($(filter Ubuntu,$(DEB_VENDOR)),)
 	COMMON_CONFIGURE_OPTIONS += -DMIR_RUN_WLCS_TESTS=OFF
+else
+  # Disable wlcs tests on Launchpad armhf, arm64 (MirServer/mir#2474)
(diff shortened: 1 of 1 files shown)


wmww (approved):
I think it's a good idea, but not sure if I'm the one to approve it
bors delegate+

RAOF:
We'll experiment with trying qemu from 22.04 and see if that makes things more stable before landing this.

Saviq:
#2490

debian/rules
```diff
@@ -41,6 +41,11 @@ endif
 # control, so disable it on non-Ubuntu builds
 ifeq ($(filter Ubuntu,$(DEB_VENDOR)),)
 	COMMON_CONFIGURE_OPTIONS += -DMIR_RUN_WLCS_TESTS=OFF
+else
+  # Disable wlcs tests on Launchpad armhf, arm64 (MirServer/mir#2474)
+  ifneq ($(filter armhf arm64,$(DEB_HOST_ARCH)),)
+    COMMON_CONFIGURE_OPTIONS += -DMIR_RUN_WLCS_TESTS=OFF
+  endif
```
Saviq, line 48: If at all, we should only disable them when cross-building - which is where we see those failures.

But let's see where does #2490 take us.

PR #2490, opened 2022-06-23
spread: move cross builds to jammy

Saviq:
The newer QEMU may have a better time running wlcs tests.


AlanGriffiths:
That looks like `MIR_BAD_BUFFER_TEST_ENVIRONMENT_BROKEN=ON` is being set where it is no longer needed...

AlanGriffiths (approved):
bors merge

PR #2492, opened 2022-06-23
Add .drop method to mutex

wmww:
Simplifies the syntax for dropping a mutex guard, and makes sure the program crashes if a mutex guard is used after being dropped (we could change this to a fatal error or exception at the cost of a single conditional branch based on a non-atomic bool each time the guard is accessed).

### DIFF
--- a/include/core/mir/mutex.h
+++ b/include/core/mir/mutex.h
@@ -38,13 +38,6 @@ public:
     {
     }
 
-    MutexGuard(MutexGuard&& from) noexcept
-        : value{from.value},
-          lock{std::move(lock)}
-    {
-        from.value = nullptr;
-    }
-
     ~MutexGuard() noexcept(false)
     {
         if (lock.owns_lock())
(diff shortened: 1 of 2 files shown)


include/core/mir/mutex.h
```diff
@@ -38,13 +38,6 @@ class MutexGuard
     {
     }
 
-    MutexGuard(MutexGuard&& from) noexcept
```
RAOF, line 41: Why are you dropping the explicit move-constructor here? This change doesn't stop a `MutexGuard` from being moved, but it *does* mean that you can access the contained data from a moved-from `MutexGuard`.

RAOF (changes requested):
Yeah, this is perhaps better. Initially, `MutexGuard<T>` contained a `T&`, which we couldn't invalidate, but once that changed to a `T*` we can sensibly `drop()` on the guard itself.

I'm good with this once you've either reintroduced the invalidating move-constructor or told me why you removed it :)

include/core/mir/mutex.h
wmww: Whoops, my intention was to disallow moving, as it was no longer necessary with the `.drop` method. However, I supposed it might be useful for other reasons so I'll reinstate it instead of `= delete`ing it.

RAOF (approved):
Yar.

I wonder; does bors understand merging to non-main branches? Let's see!

bors r+

RAOF:
Hm.

bors r+

PR #2493, opened 2022-06-24
AbstractShell: refactor confinement

wmww:
This area is and continues to be a mess. These changes fix some problems with the new surface observer stuff I'm working on.

### DIFF
--- a/src/include/server/mir/shell/abstract_shell.h
+++ b/src/include/server/mir/shell/abstract_shell.h
@@ -149,8 +149,6 @@ public:
 
     bool handle(MirEvent const& event) override;
 
-    void update_focused_surface_confined_region();
-
     void set_drag_and_drop_handle(std::vector<uint8_t> const& handle) override;
     void clear_drag_and_drop_handle() override;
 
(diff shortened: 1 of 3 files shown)


wmww:
Uhh, CI? y u no go?

Saviq:
> Uhh, CI? y u no go?

B/c conflicts:
> Conflicting files
> src/server/shell/abstract_shell.cpp

Saviq:
> B/c conflicts:

Fixed.

AlanGriffiths:
**Needs information**

[Fix surface lifetime in abstract shell tests](https://github.com/MirServer/mir/pull/2493/commits/50a3bf0e460fae5696336f2864e6da942caffb8c) seems like the wrong solution: it prevents crashes by ensuring that surfaces outlive the `shell`, I don't expect that to be necessary.

wmww:
In production the lifetime of the surfaces is tracked with a weak_ptr, but since in the tests we use `mt::fake_shared()`, it assumes the surface is alive as long as the pointer is held. The only alternative I can think of is using real `shared_ptr`s, but that seems like more churn.

AlanGriffiths (approved):
> In production the lifetime of the surfaces is tracked with a weak_ptr, but since in the tests we use `mt::fake_shared()`, it assumes the surface is alive as long as the pointer is held. The only alternative I can think of is using real `shared_ptr`s, but that seems like more churn.

Seems like we could be missing real surface-lifetime related defects, but that's pre-existing

bors merge

PR #2494, opened 2022-06-24
Add ImmediateExecutor

wmww:
There are contexts in which you want to do work immediately, specifically this will be useful in porting old observers to the new observer system, when in some places we want the old behavior to remain.


src/common/symbols.map
```diff
@@ -497,6 +497,7 @@ MIR_COMMON_2.8 {
     typeinfo?for?mir::ThreadPoolExecutor;
     vtable?for?mir::ThreadPoolExecutor;
     mir::thread_pool_executor;
+    mir::immediate_executor;
```
AlanGriffiths, line 500: Mir 2.8 has shipped

AlanGriffiths (changes requested):
There's already an `ImmediateExecutor` in the codebase. Maybe that should be moved, not duplicated.

And do we really need a new header for each new Executor we add?

include/common/mir/immediate_executor.h
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef MIR_IMMEDIATE_EXECUTOR_H_
+#define MIR_IMMEDIATE_EXECUTOR_H_
+
+#include "executor.h"
+
+namespace mir
+{
+/// An Executor that runs work on the current thread within spawn()
+extern Executor& immediate_executor;
```
RAOF, line 25: Should we have a
```
class ImmediateExecutor : public Executor { };

extern ImmediateExecutor& immediate_executor;
```

We already have `NonBlockingExecutor` for using the type to describe the requirement on the executor; since `immediate_executor` is going to be used for cases where the non-deferred-ness is necessary for correctness, maybe we should be able to express that, too?

RAOF (commented):
I'd also be happy for all the global executors to be declared in `mir/executor.h`

include/common/mir/immediate_executor.h
AlanGriffiths: +1

wmww: I fail to see how a visible ImmediateExecutor type would be useful (aside from maybe using dynamic_cast<>ing to check if things are it, which we probably want to avoid). Since there's only one possible implementation of ImmediateExecutor, anything that would take an ImmediateExecutor could simply take nothing and hard-code in the one to use.

RAOF: That's a reasonable argument, but is it actually the case that there's only one possible implementation of `ImmediateExecutor`? I'm not committed either way, but it seems a `MockImmediateExecutor` would be possibly useful, or *maybe* an implementation that caught and logged all exceptions or something?

include/common/mir/executor.h
```diff
@@ -57,6 +57,47 @@ class NonBlockingExecutor: public Executor
 {
 };
 
+class ThreadPoolExecutor : public NonBlockingExecutor
```
RAOF, line 60: Ah! I only meant to move the `extern NonBlockingExecutor& thread_pool_executor;` into `mir/executors.h`.

That said, I don't think I *mind* that this is here.

AlanGriffiths (changes requested):
The symbols.map is still wrong

wmww:
Oops, now fixed

wmww:
A number of observer multiplexers previously had their own implementations of ImmediateExecutor, which are now replaced. There might be an argument with replacing those defaults with thread_pool_executor` or something else, but because that's likely to introduce subtle behavior changes, I recommend doing that in it's own PR if desired so testing, bisecting, etc is easier.

PR #2495, opened 2022-06-25
Attempt to refactor geometry types without breaking MirAL ABI

wmww:
Attempt to do #2475 without breaking the MirAL ABI. Unclear if this is useful. Not yet working but I could probably make it work.


wmww:
Oops, that is not an optimal PR name

wmww:
We broken the MirAL ABI, so this is irrelevant

PR #2496, opened 2022-06-28
Refactor SurfaceObserver types to be compatible with new observer system

wmww:
Because the modern observer system does not always send notifications on the same thread, it needs to be given types it can hold ownership of. This PR changes types used by SurfaceObserver to work. All methods still take a raw surface pointer, which is generally not used and should be removed, but that is out of scope here.


graysonguarino:
LGTM

bors merge

PR #2503, opened 2022-07-04
Fix build on Alpine edge

AlanGriffiths:


RAOF (commented):
Because we bind `wl_seat` with a fixed version we know that libwayland promises not to call any uninitialised function pointers, so this is safe.

But *maybe* we want to handle this like we handled the `wl_touch_listener` case below, by `#ifdef`ing the handlers in/out? That way if someone wants to update the `wl_seat` version for some incidental reason, they won't accidentally cause crashes in the `wl_pointer_listener`?

AlanGriffiths:
> But _maybe_ we want to handle this like we handled the `wl_touch_listener` case below

This is what we do "everywhere" in the Wayland implementation. The `#ifdef` is about handling old versions we do support, not new ones we don't yet.

RAOF:
Yeah, I'm just wondering if that's the best way.

In the other hand, it *should* be easy enough to notice if we break it, so

bors r+

PR #2505, opened 2022-07-07
graphics: Drop vestigial VirtualOutput interface

RAOF:
None of our platforms support this, and a better solution for the same problem
is coming Real Soon Now™


AlanGriffiths (approved):
bors merge

PR #2506, opened 2022-07-08
Add SurfaceReadyRegistrar to replace SurfaceReadyObserver

wmww:
Window managers need to be notified when a surface has it's first frame posted. Currently they register an observer which then unregisters itself when this happens. When we move to the new observer system, this will no longer work because surfaces will hold weak pointers to observers instead of shared pointers.

Making an observer that owns itself until it receives notification of first frame is inadvisable because a surface may be removed and destroyed before getting a frame, and we don't want to leak the observer in that case.

This PR makes a new class to replace the functionality of `SurfaceReadyObserver`. It is notified when surfaces are created and destroyed, and manages setting up and removing observers for them.

A piece of #2371


src/include/server/mir/shell/surface_ready_registrar.h
```diff
@@ -0,0 +1,54 @@
...
+#include <mutex>
+
+namespace mir
+{
+namespace scene { class Session; class Surface; }
+namespace geometry { struct Size; }
+
+namespace shell
+{
+/// Notifies the user when a surface is ready
```
RAOF, line 32: I feel this could be better documented, or maybe could be easier to use? Particularly: how easy would it be for this to *not* have the `register_surface`/`unregister_surface` pair, and just fire the `ActivateFunction` when *any* surface becomes ready?

src/include/server/mir/shell/surface_ready_registrar.h
```diff
@@ -0,0 +1,54 @@
...
+class SurfaceReadyRegistrar
+{
+public:
+    using ActivateFunction = std::function<void(std::shared_ptr<scene::Surface> const& surface)>;
+
+    SurfaceReadyRegistrar(ActivateFunction const& surface_ready);
+    ~SurfaceReadyRegistrar();
+    void register_surface(std::shared_ptr<scene::Surface> const& surface);
```
RAOF, line 40: What are the requirements on callers to ensure that this doesn't race? That is: what do callers need to do to guarantee that `surface_ready` will not miss the first buffer being submitted?

src/server/symbols.map
```diff
@@ -915,6 +909,14 @@ MIR_SERVER_2.8 {
     vtable?for?mir::shell::WindowManagerTools;
     vtable?for?mir::time::Alarm;
     vtable?for?mir::time::AlarmFactory;
+    mir::DefaultServerConfiguration::the_keyboard_observer_registrar*;
+    mir::DefaultServerConfiguration::the_keyboard_observer_registrar*;
```
RAOF, line 913: Huh. Where did this come from?

src/server/symbols.map
```diff
@@ -915,6 +909,14 @@ MIR_SERVER_2.8 {
...
+    mir::shell::SurfaceReadyRegistrar::?SurfaceReadyRegistrar*;
+    mir::shell::SurfaceReadyRegistrar::SurfaceReadyRegistrar*;
+    typeinfo?for?mir::shell::SurfaceReadyRegistrar;
+    vtable?for?mir::shell::SurfaceReadyRegistrar;
+    mir::shell::SurfaceReadyRegistrar::register_surface*;
+    mir::shell::SurfaceReadyRegistrar::unregister_surface*;
+    mir::shell::SurfaceReadyRegistrar::clear*;
   };
- local: *;
```
RAOF, line 919: I'm pretty sure this is a mistake, yes?

src/server/symbols.map
```diff
@@ -915,6 +909,14 @@ MIR_SERVER_2.8 {
     vtable?for?mir::shell::WindowManagerTools;
     vtable?for?mir::time::Alarm;
     vtable?for?mir::time::AlarmFactory;
+    mir::DefaultServerConfiguration::the_keyboard_observer_registrar*;
+    mir::DefaultServerConfiguration::the_keyboard_observer_registrar*;
+    mir::shell::SurfaceReadyRegistrar::?SurfaceReadyRegistrar*;
+    mir::shell::SurfaceReadyRegistrar::SurfaceReadyRegistrar*;
+    typeinfo?for?mir::shell::SurfaceReadyRegistrar;
+    vtable?for?mir::shell::SurfaceReadyRegistrar;
+    mir::shell::SurfaceReadyRegistrar::register_surface*;
```
RAOF, line 918: Mir 2.8 has released; this is an ABI break, so everything should be in a shiny new `MIR_SERVER_2.9` stanza, right?

RAOF (changes requested):
I feel there's a simpler interface hiding in the bushes somewhere, but I'm not sufficiently across this bit of the code to entirely make it out.

You *do* need to fix the `symbols.map`, though :)

AlanGriffiths:
@wmww some love needed here too

wmww:
My suggestion is to do #2526 instead

AlanGriffiths:
> My suggestion is to do #2526 instead

That's landed

PR #2508, opened 2022-07-12
Bump wl_seat to v8 and implement hi-res scrolling

### LINKED ISSUE #2499: Upgrade to wl_seat v7
We currently expose v6, wayvnc wants v7.

graysonguarino:
Fixes #2499 by bumping wl_seat to v8 (instead of the requested v7). v8 introduces high resolution scrolling.


src/common/symbols.map
```diff
@@ -74,6 +74,7 @@ MIR_COMMON_2.8 {
     mir::events::make_key_event*;
     mir::events::make_keymap_event*;
     mir::events::make_pointer_axis_discrete_scroll_event*;
+    mir::events::make_pointer_axis_value120_scroll_event*;
```
AlanGriffiths, line 77: Mir 2.8 has been released, new symbols should go in a `2.9` stanza

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -262,13 +264,47 @@ void mf::WlPointer::axis(MirPointerEvent const* event)
...
+        needs_frame=true;
+    }
+    else if (h_scroll_discrete && version_supports_axis_value120())
+    {
+        auto value120 = h_scroll_discrete * 120;
+        // "The value120 must not be zero"
+        if (value120 == 0) 
+        {
+            value120 = 1;
+        }
```
AlanGriffiths, line 279: Is this the right approach? Without reading the extension, I'd expect the event to only be sent when non-zero.

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -262,13 +264,47 @@ void mf::WlPointer::axis(MirPointerEvent const* event)
...
+        needs_frame=true;
+    }
+    else if (v_scroll_discrete && version_supports_axis_value120())
+    {
+        auto value120 = v_scroll_discrete * 120;
+        // "The value120 must not be zero"
+        if (value120 == 0) 
+        {
+            value120 = 1;
+        }
```
AlanGriffiths, line 302: ditto

AlanGriffiths (changes requested):
Looks sensible, but the linker script needs fixing

AlanGriffiths:
@graysonguarino looks like you need to fix some builds:
```plain
/<<PKGBUILDDIR>>/src/platforms/evdev/libinput_device.cpp:315:13: error: ‘libinput_event_pointer_get_scroll_value_v120’ was not declared in this scope; did you mean ‘libinput_event_pointer_get_axis_value’?
  315 |             libinput_event_pointer_get_scroll_value_v120(pointer, LIBINPUT_POINTER_AXIS_SCROLL_HORIZONTAL) :
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |             libinput_event_pointer_get_axis_value
```

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -262,13 +264,43 @@ void mf::WlPointer::axis(MirPointerEvent const* event)
...
+    if (h_scroll_value120 && version_supports_axis_value120())
+    {
+        send_axis_value120_event(Axis::horizontal_scroll, h_scroll_value120);
+        needs_frame=true;
+    }
+    else if (h_scroll_discrete && version_supports_axis_value120())
+    {
+        auto value120 = h_scroll_discrete * 120;
+        // "The value120 must not be zero"
+        if (value120 != 0)
```
wmww, line 276: If `h_scroll_discrete != 0` (as checked in the else if above) than can't we assume `h_scroll_discrete * 120 != 0`?

RAOF (commented):
I think this would be improved by removing the `#ifdef`ery in the tests; they don't need to be gated on the libinput version.

Otherwise, I agree with Sophie that the whole `mir::events::*` infrastructure is showing its age, but there's nothing to be done about that here.

src/platforms/evdev/CMakeLists.txt
```diff
@@ -6,6 +6,12 @@ else ()
   add_compile_definitions(MIR_LIBINPUT_HAS_ACCEL_PROFILE=1)
 endif ()
 
+if ("${LIBINPUT_VERSION}" VERSION_LESS "1.19")
+  add_compile_definitions(MIR_LIBINPUT_HAS_VALUE120=0)
+else ()
+  add_compile_definitions(MIR_LIBINPUT_HAS_VALUE120=1)
+endif ()
+
```
AlanGriffiths, line 14: I know this is consistent with other `MIR_LIBINPUT...` macros, but IME the usual idiom is:
```
if ("${LIBINPUT_VERSION}" VERSION_GREATER_EQUAL "1.19")
  add_compile_definitions(MIR_LIBINPUT_HAS_VALUE120)
endif ()
```
with
```
#ifdef MIR_LIBINPUT_HAS_VALUE120
...
```
Is there a reason for this approach?

AlanGriffiths:
> I think this would be improved by removing the `#ifdef`ery in the tests; they don't need to be gated on the libinput version.

I concur. The only place the `#if`ery is needed is where we actually call the libinput API. 

> Otherwise, I agree with Sophie that the whole `mir::events::*` infrastructure is showing its age, but there's nothing to be done about that here.

Agreed. (I don't remember ever liking it, but this isn't the PR to address that)

src/platforms/evdev/CMakeLists.txt
graysonguarino: No reason 🤷‍♂️ Just following what was done for `MIR_LIBINPUT_HAS_ACCEL_PROFILE`. Now changing this.

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -558,6 +538,36 @@ void mf::WlPointer::on_commit(WlSurface* surface)
...
+    case Axis::horizontal_scroll:
+        counter = &h_value120_counter;
+        break;
+    case Axis::vertical_scroll:
+        counter = &v_value120_counter;
+        break;
+    default:
+        BOOST_THROW_EXCEPTION(std::runtime_error{
+            "Invalid axis given to update_value120_counter()"});
+    }
```
AlanGriffiths, line 555: Can't we just create a reference?

```c++

   auto const lambda = [](int& counter, float value120) {...};

    switch (axis)
    {
    case Axis::horizontal_scroll:
        return lambda(h_value120_counter, value120);
    case Axis::vertical_scroll:
        return lambda(v_value120_counter, value120);
    default:
        BOOST_THROW_EXCEPTION(std::runtime_error{
            "Invalid axis given to update_value120_counter()"});
    }
```

AlanGriffiths (commented):
I'm confused by the latest changes. Will look in the AM when I have more headspace

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -558,6 +538,36 @@ void mf::WlPointer::on_commit(WlSurface* surface)
...
+    *counter += value120;
+    auto const discrete = *counter % 120;
+    if (*counter >= 120)
+    {
+        *counter -= discrete * 120;
+    }
+    else if (*counter <= -120)
+    {
+        *counter += discrete * 120;
+    }
```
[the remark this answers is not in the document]
graysonguarino: Even without the headspace, you caught my slip-up.

src/server/frontend_wayland/wl_pointer.cpp
graysonguarino: We sure can!

graysonguarino:
Reverting to draft while waiting for #2514 to land.

wmww:
Pointer event refactor is now long complete, time to take another look at this.

As we've discussed, my suggestion has changed to including both discrete and v120 values in the event, as v120 values can not be converted to discrete values without an accumulator, and we need to support context-free conversion to discrete scroll events to maintain backwards compatibility of our event ABI/API.

Since we haven't released since the refactor, you can add v120 info to `ScrollAxisV1` (if we had of released since, we would need to create a `ScrollAxisV2` to maintain ABI stability)

wmww:
Hmm, something went wrong with that merge, commit, lets try again

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -263,33 +263,51 @@ void mf::WlPointer::axis(MirPointerEvent const* event)
...
+        {
+            send_axis_discrete_event(Axis::vertical_scroll, v_scroll_discrete);
+            needs_frame = true;
+        }
     }
 
-    // If we sent discrete scroll events, we're required to also send corresponding non-discrete ones (even if their
-    // values are 0 for some reason)
+    // If we sent value120 scroll events, we're required to also send a non-value120 event on the same axis, evin if the
+    // value is 0.
```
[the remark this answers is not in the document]
RAOF: Also, is it the case that non-zero `discrete` implies non-zero `value120`?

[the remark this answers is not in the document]
wmww: > While the axis_value120 event description states that this replaces the axis_discrete event, it does not say that an axis event is also sent

Since the discrete event requires an axis event, and the value120 replaces discrete with no mention either of requiring or not requiring it, I think it's at least ambiguous. Since there's no harm in sending it, might as well.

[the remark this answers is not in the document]
wmww: The reason is likely because the .axis is what carries the timestamp.

src/server/frontend_wayland/virtual_pointer_v1.cpp
```diff
@@ -346,12 +346,14 @@ void mf::VirtualPointerV1::axis_discrete(uint32_t time, uint32_t axis, double va
...
+        pending.scroll_h.value120 += geom::generic::DeltaX<int>{discrete * 120};
         pending.scroll_h.precise += geom::DeltaXF{value};
         break;
 
     case mw::Pointer::Axis::vertical_scroll:
         // TODO: replace with geom::DeltaX{...} once geom cleanup lands
         pending.scroll_v.discrete += geom::generic::DeltaY<int>{discrete};
+        pending.scroll_v.value120 += geom::generic::DeltaY<int>{discrete * 120};
         pending.scroll_v.precise += geom::DeltaYF{value};
         break;
```
AlanGriffiths, line 358:
```suggestion
    case mw::Pointer::Axis::horizontal_scroll:
        pending.scroll_h.discrete += geom::DeltaX{discrete};
        pending.scroll_h.value120 += geom::DeltaX{discrete * 120};
        pending.scroll_h.precise += geom::DeltaXF{value};
        break;

    case mw::Pointer::Axis::vertical_scroll:
        pending.scroll_v.discrete += geom::DeltaY{discrete};
        pending.scroll_v.value120 += geom::DeltaY{discrete * 120};
        pending.scroll_v.precise += geom::DeltaYF{value};
        break;
```

AlanGriffiths (approved):
A small cleanup is possible (but not essential)

bors delegate

AlanGriffiths:
WFM

bors merge

AlanGriffiths:
> * [BuildAndTest (google:ubuntu-22.04-64:spread/build/sbuild:ubuntu_jammy)](https://github.com/MirServer/mir/runs/7785116980?check_suite_focus=true)

MultiThreadedCompositor again! (Unlikely to be related as we're seeing this "everywhere" recently)

bors retry

PR #2510, opened 2022-07-18
Wl pointer behavior tweaks

wmww:
Some small improvements to the behavior of `wl_pointer`. Originally part of #2504 and needed for the virtual pointer WLCS tests to pass.


AlanGriffiths (approved):
I'm not sure I understand exactly what this affects, but it looks more reasonable than the previous behaviour

AlanGriffiths (changes requested):
Doesn't build with the reworked event interface on main

AlanGriffiths (approved):
bors merge

PR #2511, opened 2022-07-19
Clean up legacy feature checks

AlanGriffiths:
I noticed a bunch of feature checks that are not needed for any supported series and cleaned them up


src/platform/graphics/CMakeLists.txt
```diff
@@ -38,10 +38,6 @@ add_library(mirplatformgraphicscommon OBJECT
 mir_generate_protocol_wrapper(mirplatformgraphicscommon "zwp_" protocol/linux-dmabuf-unstable-v1.xml)
 
 if (DRM_VERSION VERSION_GREATER 2.4.107)
-  target_compile_definitions(
```
RAOF, line 41: Is this correct? We're only checking for `drm >= 2.4.84`, and `drmGetFormatModifierName`  appeared in 2.4.107 as far as I can tell.

RAOF: Hm, and you're not actually removing any of the `#ifdef MIR_HAVE_DRM_GET_MODIFIER_NAME`s?

RAOF (changes requested):
I think the `ModiferName` changes are incorrect; the rest looks fine.

src/platform/graphics/CMakeLists.txt
AlanGriffiths: Oops. Dropped that commit as it is messed up

cmake/Doxygen.cmake
```diff
@@ -1,13 +1,13 @@
 # Check if doxygen is present and add 'make doc' target
-find_package(Doxygen)
+find_package(Doxygen 1.8.0)
 
 option(
   BUILD_DOXYGEN
   "Build Doxygen documentation as part of the default build"
-  OFF
+  OFF`
```
RAOF, line 7: This looks like a spurious '`'?

RAOF (approved):
Other than the typo, looks good to me.

bors delegate=AlanGriffiths

PR #2512, opened 2022-07-19
platform/graphics: Drop `DisplayConfiguration::for_each_card`

RAOF:
Nothing has ever used this information, and in the new platform interface
we have one `DisplayPlatform` per card anyway.


AlanGriffiths (approved):
LGTM

bors r+

PR #2514, opened 2022-07-19
Refactor pointer events to use geometry types

wmww:
Also adds a single pointer event builder method that includes all properties, and deprecates the older event builder methods.


wmww:
I know `mir::geometry::generic::Displacement<mir::geometry::generic::Value<bool>::Wrapper>` is quite ugly, but it will collapse down into `mir::geometry::generic::Displacement<bool>` once we get #2475 landed.

AlanGriffiths:
> once we get #2475 landed

At the current rate of progress that could be some time! :wink:

RAOF (commented):
I'm not sure what I think this should *eventually* look like, but I think this is an improvement on the status quo.

I *think* that scroll would be clearer if it were not handled as a `Displacement`, but I'm willing to hear arguments :wink:

AlanGriffiths:
> I _think_ that scroll would be clearer if it were not handled as a `Displacement`

I concur. scroll events are handled as horizontal/vertical scrolls not like move events

wmww:
> They're logically two separate event-like-things rather than a single logical event like motion

> scroll events are handled as horizontal/vertical scrolls not like move events

Either I'm not understanding or we have a different impression of what scroll is. To me, they're just a different type of 2D motion. The Wayland protocol has this to say:

> For scroll events (vertical and horizontal scroll axes), the value parameter is the length of a vector along the specified axis in a coordinate space identical to those of motion events, representing a relative movement along the specified axis.

I see how `Displacement<bool>` is a little strange. It's also a little strange to allow a non-zero scroll value and also `scroll_stop`, but I don't think particularly harmful. The variant approach would arguably be more elegant, but I'm not sure it would yield any significant practical benefit.

wmww:
After discussions with the team, we've decided to have a `ScrollAxis` struct, and then have horizontal and vertical instances of that.

include/common/mir/events/event_builders.h
```diff
@@ -95,6 +99,20 @@ void add_touch(
...
+    MirInputDeviceId device_id,
+    std::chrono::nanoseconds timestamp,
+    std::vector<uint8_t> const& cookie,
+    MirInputEventModifiers mods,
+    MirPointerAction action,
+    MirPointerButtons buttons,
+    std::optional<mir::geometry::PointF> position,
+    mir::geometry::DisplacementF motion,
+    MirPointerAxisSource axis_source,
+    events::ScrollAxisV1<mir::geometry::DeltaXTag> h_scroll,
```
RAOF, line 113: Some places you use `ScrollAxisV1<Tag>`, some places you use `ScrollAxis<Tag>`, and some places you use `ScrollAxisH`/`ScrollAxisV`. There's presumably some reason behind this, but I can't see it?

RAOF (commented):
Yeah, that's nicer!

I *think* we should be using `ScrollAxisV`/`ScrollAxisH` everywhere, though? I don't *think* there's any reason not to? The underlying type is `ScrollAxisV1<Tag>`, so we can still add a `ScrollAxisV2<Tag>` without breaking ABI, and I *think* that we can even do that more easily if the API is the `ScrollAxisV/H` types, as then you might not need source changes to update to the new stuff?

include/platform/mir/input/event_builder.h
```diff
@@ -46,13 +49,15 @@ class EventBuilder
         xkb_keysym_t keysym,
         int scan_code) = 0;
 
+    [[deprecated("use the pointer_event() that includes all properties instead")]]
     virtual EventUPtr pointer_event(
         std::optional<Timestamp> timestamp,
         MirPointerAction action,
         MirPointerButtons buttons_pressed,
         float hscroll_value, float vscroll_value,
         float relative_x_value, float relative_y_value) = 0;
```
AlanGriffiths, line 58: Is there any point to retaining these APIs?

include/common/mir/events/event_builders.h
wmww: How it should be is `ScrollAxisV1<Tag>` is used in places where we don't want the version auto-updated (for example, in function signatures of ABI-stable functions), `ScrollAxis<Tag>` is used in places that handle the tag generically but should always use the latest version and `ScrollAxisH`/`ScrollAxisV` are used most places where neither of those are the case. As it is now there are residual uses of `ScrollAxis<Tag>`, which I'll fix.

include/platform/mir/input/event_builder.h
wmww: As far as I know we're not breaking platform ABI

src/common/events/event_builders.cpp
```diff
@@ -240,6 +239,33 @@ void mev::add_touch(
...
+    std::chrono::nanoseconds timestamp,
+    std::vector<uint8_t> const& cookie,
+    MirInputEventModifiers mods,
+    MirPointerAction action,
+    MirPointerButtons buttons,
+    std::optional<geom::PointF> position,
+    geom::DisplacementF motion,
+    MirPointerAxisSource axis_source,
+    events::ScrollAxisV1H h_scroll,
+    events::ScrollAxisV1V v_scroll)
```
RAOF, line 253: I think it's important for the next person who's reading this to document that this is *deliberately* using the `V1` types while the header declares these parameters with the unversioned alias so that we get a compile error should someone bump the unversioned alias without remembering to do the rest of the work to maintain ABI.

RAOF (approved):
LGTM, modulo the addition of the comment.

bors delegate=wmww

include/platform/mir/input/event_builder.h
AlanGriffiths: This isn't an ABI we're expecting outside folks to use yet. So breaking it isn't a big deal

PR #2516, opened 2022-07-21
Bump MirAL so to version 5 and MirAL version to 3.6

wmww:
We need to bump MirAL .so for #2475. We can probably take this opportunity to do additional cleanup"


wmww:
That's what I was attempting to do in #2495, but it was proving tricky. It might be made to work if there's significant motivation, but since shells will need to recompile anyway to pick up the new mircore, why not bump miral?

wmww:
CI is failing because it wants me to change all the mircore symbols from `2.9.0` to `2.8.0`. They are 2.9 in current master, and this PR doesn't touch any mircore related stuff. I am at a loss.

AlanGriffiths (approved):
bors merge

PR #2517, opened 2022-07-22
Auto generate symbols map

wmww:
WIP attempt to auto-generate symbols.map and debian/*.symbols files based on doxygen data. Only file it tries to generate so far is miral/symbols.map, and that isn't complete enough to work yet.


include/miral/miral/window_manager_tools.h
```diff
@@ -54,10 +54,11 @@ class Workspace;
 class WindowManagerToolsImplementation;
 
 /// Window management functions for querying and updating MirAL's model
+/// \remark Since MirAL 3.0
 class WindowManagerTools
 {
 public:
-    explicit WindowManagerTools(WindowManagerToolsImplementation* tools);
+    explicit WindowManagerTools(miral::WindowManagerToolsImplementation* tools);
```
RAOF, line 61: What's the failure mode if the (otherwise unnecessary) `miral::` namespace specifier is left off here?

RAOF (commented):
This seems an interesting approach, but it looks like maybe there could be more smarts in the script and less lying in the comments? Particularly: it seems the script could take a “what is the MirAL SOVER” parameter, and use that as the version for all symbols either without “Since” annotations, or with annotations indicating a version less than SOVER?

That'd dramatically cut down the diff, remove the need to lie in comments, and reduce the amount of future diff needed, too?

include/miral/miral/window_manager_tools.h
wmww: Explained in comment

include/core/mir/fatal.h
```diff
@@ -31,6 +31,7 @@ namespace mir
  * Note the reason parameter is a simple char* so its value is clearly visible
  * in stack trace output.
  * \remark There is no attempt to make this thread-safe, if it needs to be changed
+ * \remark Since MirCore 0.25
```
AlanGriffiths, line 34: I don't think we want (nor should we need) "Since" annotations that date from before the current major version/SONAME.

debian/libmiral4.symbols
```diff
@@ -419,4 +418,9 @@ libmiral.so.4 libmiral4 #MINVER#
...
  (c++)"miral::socket_fd_of(std::shared_ptr<mir::scene::Session> const&)@MIRAL_3.4" 3.4.0
  MIRAL_3.5@MIRAL_3.5 3.5.0
- (c++)"miral::WaylandExtensions::zwlr_screencopy_manager_v1@MIRAL_3.5" 3.5.0
\ No newline at end of file
+ (c++)"miral::WaylandExtensions::zwlr_screencopy_manager_v1@MIRAL_3.5" 3.5.0
+ MIRAL_3.6@MIRAL_3.6 3.6.0
+ (c++)"miral::WaylandExtensions::Context::Context()@MIRAL_3.6" 3.6.0
+ (c++)"miral::WaylandExtensions::Context::~Context()@MIRAL_3.6" 3.6.0
+ (c++)"vtable for miral::WaylandExtensions::Context@MIRAL_3.6" 3.6.0
+ (c++)"typeinfo for miral::WaylandExtensions::Context@MIRAL_3.6" 3.6.0
```
AlanGriffiths, line 426: This is wrong. `WaylandExtensions::Context` comes from MIRAL_2.5 - so it belongs in the 3.0 stanza

include/miral/miral/window_manager_tools.h
```diff
@@ -194,30 +197,30 @@ class WindowManagerTools
      * \remark the tools hold only a weak_ptr<> to the workspace - there is no need for an explicit "destroy".
      * @return a shared_ptr owning the workspace
      */
-    auto create_workspace() -> std::shared_ptr<Workspace>;
+    auto create_workspace() -> std::shared_ptr<miral::Workspace>;
```
AlanGriffiths, line 200: Not looked at fixing this, but I'm not a fan of adding noise to code where it isn't necessary (except for deficiencies in this toolchain).

src/miral/symbols.map
```diff
@@ -1,483 +1,479 @@
...
-    vtable?for?miral::WindowInfo;
-    vtable?for?miral::WindowManagementPolicy;
-    vtable?for?miral::WindowManagerOption;
-    vtable?for?miral::WindowManagerOptions;
-    vtable?for?miral::WindowManagerTools;
-    vtable?for?miral::WindowSpecification;
-    vtable?for?miral::WindowSpecification::AspectRatio;
-    vtable?for?miral::X11Support;
-    vtable?for?miral::Zone;
+    "::SetApplicationAuthorizer(typename..., Args const&...)";
```
AlanGriffiths, line 4: This is wrong: `template<typename Policy> class SetApplicationAuthorizer` is in namespace `miral`, and templates don't export symbols unless instantiated.

AlanGriffiths (changes requested):
A bold attempt. But it is getting stuff wrong and the need for "Since" tags and "unnecessary" explicit naming will be high maintenance

include/core/mir/fatal.h
wmww: What exactly are you suggesting? Are you saying you prefer the current system of maintaining a separate file with symbols, or do you just want a different type of in-code annotation?

This PR comes out of discussions with RAOF where we concluded keeping export info alongside the symbols in the code would be nicer than in a separate file (in which case, all exported functions need to be annotated somehow). I suppose the generator could read a symbols.map file and leave previous stanzas as-is, but I like the idea of the generator being a pure function from source code to symbols.

AlanGriffiths: > it seems the script could take a “what is the MirAL SOVER” parameter, and use that as the version for all symbols either without “Since” annotations

wmww: It's less about the version, and more about knowing which symbols to export at all. If the version seems like clutter we could make a different annotation that doesn't include the version but still marks them for export.

AlanGriffiths: > It's less about the version, and more about knowing which symbols to export at all. If the version seems like clutter we could make a different annotation that doesn't include the version but still marks them for export.

It isn't obvious why annotations are required, or even that they are a good, solution to this.

Symbols should be exported if they are:
1. Declared in a public header;
2. Accessible (i.e. namespace scope, public/protected static members, public/protected member functions);
3. Not inline, nor template

RAOF: Oooh, yeah. That's a nice approach!

There's the chance with 3. that they *do* need to be exported, but that's only for weird scenarios (eg: where something takes the address of the function) and I don't *think* we rely on any of that behaviour.

It seems that Doxygen collects that information, how hard would it be to use it?

AlanGriffiths: Not very, that's used in `regenerate-miral-symbols-map.py` already

wmww:
Ah, looks like we're not building the docs in CI. We need to do that for the test to pass.

Saviq:
@wmww should this go on?

wmww:
I think it's worth doing, bug it needs a rework based on feedback. Haven't gotten back to that yet.

Saviq:
Superseded by:
- #3297

PR #2518, opened 2022-07-22
Add tools/scan_symbols.py script

wmww:
Given the path to a .so file, this script will compare the symbols in the library with the symbols in the debian symbols file and show the symbols only present in one or the other. Not perfect, but makes manually editing symbols easier.

### DIFF
--- a/tools/scan_symbols.py
+++ b/tools/scan_symbols.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# Copyright © 2022 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 or 3
+# as published by the Free Software Foundation.
+#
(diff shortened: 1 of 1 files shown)


tools/scan_symbols.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python
```
AlanGriffiths, line 1: This doesn't work for me - Ubuntu doesn't interpret `python` as `python3`

FWIW other scripts are using:

```suggestion
#! /usr/bin/python3
```
And that does work

wmww: Huh, my 21.10 system does. I wonder if I explicitly set that up at some point.

AlanGriffiths (approved):
Probably useful and certainly harmless

bors merge

tools/scan_symbols.py
AlanGriffiths: 21.10 is EOL!

PR #2520, opened 2022-07-27
Deduplicate libraries to link against

AlanGriffiths:
Don't use both _LDFLAGS and _LIBRARIES as the former (usually) contains the latter. (But, to make it interesting _LDFLAGS is sometimes empty)


wmww (approved):
LGTM
bors r+

PR #2521, opened 2022-07-30
Allow clients to get repeated string options

AlanGriffiths:
This capability exists in the mirserver API (as used by `driver-quirks`) but was never exposed through MirAL


AlanGriffiths:
Actually, it would be better to land #2516 first

PR #2523, opened 2022-08-01
Allow servers to get repeated string options

AlanGriffiths:
This capability exists in the mirserver API (as used by `driver-quirks`) but was never exposed through MirAL


src/miral/regenerate-miral-symbols-map.py
```diff
@@ -530,13 +530,11 @@ def _report(publish, symbol):
...
-  };
-local: *;
-};
-global:'''
+    vtable?for?miral::WaylandExtensions::EnableInfo;'''
 
-END_NEW_STANZA = ''
+END_NEW_STANZA = '''  };
+local: *;
+};'''
```
Saviq, line 537: Did the `global:` go away on purpose?

Saviq: Also this doesn't seem like a change related to this PR?

AlanGriffiths: > Also this doesn't seem like a change related to this PR?

True, it fixes the changes made in #2516. (I noticed because I changed it in the earlier version of this PR.) Want it split out?

Saviq (approved):
Yup.

bors merge

PR #2524, opened 2022-08-01
Fix regenerate-miral-symbols-map.py

AlanGriffiths:
...so it regenerates symbols.map correctly. (We accidentally broke it in #2516)


wmww (approved):
bors r+

PR #2525, opened 2022-08-01
Fix CommandLine strings option

AlanGriffiths:


wmww (changes requested):
Formatting nit: braces around single line if

wmww (approved):
LGTM
bors r+

PR #2526, opened 2022-08-01
Add surface_ready() shell/wm method

wmww:
Alternative to #2506. Makes the surface ready notification come from the frontend instead of being based on when a surface gets it's first buffer (in practice it should be the same behavior, but this architecture gives us more flexibility to control in the frontend when a surface is ready). Most importantly, no messy observer logic.


AlanGriffiths (approved):
bors merge

PR #2156, opened 2021-08-25
gha: use LXD in spread tasks

Saviq:

PR #2530, opened 2022-08-04
spread: optimize in sbuild, too

Saviq:
Otherwise symbols may not be as we expect them.

### DIFF
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -76,17 +76,9 @@ jobs:
 
     - name: Set up Spread
       run: |
-        set -e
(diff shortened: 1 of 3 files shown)


AlanGriffiths (changes requested):
You should need to revert 555c3a06eaf613e4d18e81c0b23c6854509c9975 too!

AlanGriffiths (approved):
Not sure if this is now stale?

bors try

Saviq:
Not stale, but slow. And requires a fix in spread.

Saviq:
This got folded into #2156

PR #2531, opened 2022-08-04
Clean out symbols that don't (or may not) exist from mircore symbols.map

AlanGriffiths:
Some of these went with recent ABI changes, some never existed, some only exist without optimization


Saviq (approved):
Would be nice if we didn't have to maintain all that manually…

bors merge

AlanGriffiths:
> Would be nice if we didn't have to maintain all that manually…

Watch this space. We have `regenerate-FOO-symbols-map.py` scripts for miral and miroil that partially automate this. And Sophie is trying to improve on that: #2517

Issue #2535, opened 2022-08-05
Send keyboard modifiers after keyboard enter


wmww:
The newest version of the protocol states that is required, it doesn't appear we do it.

wmww:
Confirmed. If the modifiers change we send them before the enter event, and if they haven't changed we don't send them at all. The protocol states we need to always send modifiers after the enter event.

AlanGriffiths:
This sounds like a likely cause for #2025 - and looking at the code, we don't seem to have any logic to send modifiers to a window gaining focus (whether or not they have changed).

Issue #2025, opened 2021-05-04
Keyboard modifier state gets out of sync


AlanGriffiths:
Occasionally the modifier (Ctrl/Alt/Meta/Shift) state gets locked out of sync.

It corrects after shifting input focus (in egmde `Ctrl-Alt-A`, `Esc` for example), so I think that Mir has the right state internally, but that it has failed to send an update to the client.

PR #2537, opened 2022-08-05
Send keyboard modifiers after keyboard enter

AlanGriffiths:
Possibly a naive implementation (as I've not read the protocol requirements)

Fixes: #2535
Fixes: #2025


wmww (approved):

AlanGriffiths:
Ah, that's the hidden path

bors r=wmww

PR #2538, opened 2022-08-08
ci: disable precompiled headers

Saviq:


Saviq:
More trouble than it's worth.

PR #2539, opened 2022-08-08
Include null terminator in keymap

### LINKED ISSUE #2533: Null-terminate buffer behind keymap fd
The latest protocol states that the `xkb_v1` keymap format is a "null-terminated string".

I my interpretation of our code is that it's currently incorrect, and adding a +1 to the length src/server/frontend_wayland/keyboard_helper.cpp might be enough to fix it, but I don't have time to think or poke at it more now.

wmww:
Fixes #2533. I haven't observed any bugs related to this, but it is technically required by the latest wayland.xml (my impression is that in previous versions it was ambiguous but the new behavior was always generally expected, so changing behavior for all versions should be most correct).


RAOF (approved):
Weird.

I *guess* `xkb_keymap_new_from_buffer` wasn't available in the initial libxkbcommon, but it's strange to send both a `buffer,len` pair *and* specify that the buffer is null-terminated. :woman_shrugging: 

That is what the protocol says, though!

bors r+

PR #2540, opened 2022-08-08
mmap keymap fd instead of reading from it

wmww:
This and #2508 are the final two pieces required to make input work with wayvnc. I don't completely understand why reading from the fd works for squeekboard and not wayvnc, but mmap works for both and it's what wlroots does.

### DIFF
--- a/src/server/frontend_wayland/virtual_keyboard_v1.cpp
+++ b/src/server/frontend_wayland/virtual_keyboard_v1.cpp
@@ -58,7 +58,7 @@ auto load_keymap(uint32_t format, mir::Fd fd, size_t size) -> std::shared_ptr<mi
         BOOST_THROW_EXCEPTION(std::runtime_error("invalid keymap format " + std::to_string(format)));
     }
 
-    void* const data = mmap(NULL, size, PROT_READ, MAP_PRIVATE, fd, 0);
+    void* const data = mmap(NULL, size, PROT_READ, MAP_SHARED, fd, 0);
     if (data == MAP_FAILED)
     {
         BOOST_THROW_EXCEPTION(std::system_error(errno, std::system_category(), "failed to mmap keymap fd"));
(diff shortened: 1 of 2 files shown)


src/server/frontend_wayland/virtual_keyboard_v1.cpp
```diff
@@ -57,31 +58,14 @@ auto load_keymap(uint32_t format, mir::Fd fd, size_t size) -> std::shared_ptr<mi
...
-                "keymap fd hit EOF " +
-                std::to_string(remaining) +
-                " bytes before specified size"));
-        }
-        else
-        {
-            current += result;
-            remaining -= result;
-        }
+        BOOST_THROW_EXCEPTION(std::runtime_error("failed to mmap keymap fd"));
```
RAOF, line 64: Could we please throw a `std::system_error` here? We can capture `errno` and make failure a little more debuggable.

RAOF (approved):
Looks sensible.

Please use a `system_error` where we've got access to a relevant `errno`. With that change,

bors delegate=wmww

wmww:
Marked as draft until I figure out why CI is failing

Saviq:
> Marked as draft until I figure out why CI is failing

FWIW it's tricky to dig out in the logs, but:

```
C++ exception with description "Failed to map file: Invalid argument" thrown in the test body.
```

wmww:
I still haven't narrowed in on *exactly* what went wrong, and my intuition tells me doing so may or may not take orders of magnitude more time than this problem has already consumed.

What we know for sure is that our wrapped mmap implementation is broken (removing it makes the failing tests pass), so I switched it from `dlsym`ing the libc mmap wrapper to calling the mmap syscall directly.

AlanGriffiths (approved):
LGTM

bors merge

PR #2544, opened 2022-08-10
[miral] Improvement to ExternalClientLauncher

AlanGriffiths:
Don't force clients to split command lines themselves


AlanGriffiths:
Note that this will conflict with the "reaper" code in Miriway, but I think that's the only downstream affected.

Once this lands I'll push an update to Miriway.

AlanGriffiths:
Converted to draft until I fix the failures

AlanGriffiths:
I'll abandon 0327028 as it clearly leaves things in a dubious state because there's no way to unregister a signal handler. (And even hacking one doesn't solve all the raciness)

RAOF:
> I'll abandon [0327028](https://github.com/MirServer/mir/commit/0327028443b05efe778f81133102573f303482f8) as it clearly leaves things in a dubious state because there's no way to unregister a signal handler. (And even hacking one doesn't solve all the raciness)

*This* looks like a use-case for pidfd! The file descriptor will become readable when the process terminates, so we can use the fd-watching infrastructure rather than a signal handler.

graysonguarino:
LGTM

bors merge

PR #2549, opened 2022-08-12
mtf/mmap_wrapper: Pick correct mmap symbol on 32bit platforms

RAOF:
`mmap` has two different declarations, depending on whether `_FILE_OFFSET_BITS=64` is defined.
On 64-bit platforms `off_t` and `off64_t` are both 64 bits, but on 32-bit platforms `off_t`
is 32 bits and `off64_t` is 64 bits. Hello, armhf!

Use the `mmap64` symbol when `off_t` is `typedef`ed to `off64_t` (ie: when
`_FILE_OFFSET_BITS=64`).

### DIFF
--- a/tests/mir_test_framework/mmap_wrapper.cpp
+++ b/tests/mir_test_framework/mmap_wrapper.cpp
@@ -24,17 +24,9 @@
 #include "mir_test_framework/interposer_helper.h"
 
 #include <boost/throw_exception.hpp>
-#include <sys/syscall.h>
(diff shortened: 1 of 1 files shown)


RAOF:
Fixes :crossed_fingers: CI for #2540

RAOF:
Clang looks to be an unrelated failure? Retrying that build.

AlanGriffiths (approved):
Yup

PR #2550, opened 2022-08-12
Add a new driver quirk to disable KMS modeset probe

AlanGriffiths:
With an implicit `display-quirks=disable-kms-probe:virtio_gpu` as we need it for virtio

### DIFF
--- a/src/platforms/gbm-kms/server/kms/quirks.cpp
+++ b/src/platforms/gbm-kms/server/kms/quirks.cpp
@@ -135,9 +135,35 @@ public:
         return should_skip_driver || should_skip_devnode;
     }
 
+    auto require_modesetting_support(mir::udev::Device const& device) const -> bool
+    {
+        auto const devnode = value_or(device.devnode(), "");
+        auto const parent_device = device.parent();
+        auto const driver =
+            [&]()
+            {
+                if (parent_device)
+                {
+                    return value_or(parent_device->driver(), "");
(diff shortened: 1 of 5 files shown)


src/platforms/gbm-kms/server/kms/quirks.cpp
```diff
@@ -58,52 +58,37 @@ class mgg::Quirks::Impl
...
-            {
-                mir::log_warning(
-                    "Ignoring unexpected value for %s option: %s (valid types are “devnode” and “driver”)",
-                    quirks_option_name,
-                    type.c_str());
-                continue;
-            }
+            // If we didn't `continue` above, we're ignoring...
+            mir::log_warning(
+                "Ignoring unexpected value for %s option: %s (expects value of the form “skip:<type>:<value>”)",
```
RAOF, line 89: This is no longer true; it expects either “skip:<type>:<value>” or “disable-kms-probe:<value>”.

Next type of quirk we add we should make a more principled “here's what the supported quirks are, and how to specify them” infrastrucutre :wink:

RAOF (approved):
It would be nice to test that we parse malformed `--driver-quirks=` options sensibly, but that's not blocking.

Feel free to merge after rewording the failed-to-parse warning.

bors delegate=AlanGriffiths

AlanGriffiths:
bors r=RAOF

PR #2551, opened 2022-08-12
Drop (accidental?) extern "C++" in miaral symbols

wmww:
Not sure how that slipped through, at least on LLD it prevents Mir from building.


graysonguarino:
bors merge

graysonguarino:
bors merge

PR #2552, opened 2022-08-12
Expose miral::Zone::id()

wmww:
Not having this has been a recurring annoyance in writing shells. With the ID a shell can have a hashmap of zones, store an associated zone with a window, etc. Without it identifying zones is much less convenient and efficient.


AlanGriffiths (approved):
I think there are problems with `Zone` "identity", but this PR doesn't introduce them.

bors merge

PR #2553, opened 2022-08-12
Add virtual pointer protocol name to miral

wmww:
Forgot this in original PR


AlanGriffiths (approved):
Rebased this on a correct fix to the miral symbols issues

bors delegate

PR #2554, opened 2022-08-15
trivial: Use C++17 [[fallthrough]] attribute

RAOF:
It exists, is standardised, so why not. Compilers might not continue to treat a `//fallthrough` or `//falls through` comment equivalently!


AlanGriffiths:
Rerunning failed Fedora job for sanity

AlanGriffiths:
bors retry

PR #2556, opened 2022-08-15
Add more defaults to driver-quirks

AlanGriffiths:
Mir's gbm-kms support isn't (yet) working with nvidia or evdi.

Better to disable drivers here than in Ubuntu Frame configuration: When we have them working we just need update one project. 

See: https://github.com/MirServer/ubuntu-frame/blob/a6ba2e1d84a2d6edd3b7e5c439a3c06d78998f96/snap/hooks/configure#L81

### DIFF
--- a/src/platforms/gbm-kms/server/kms/quirks.cpp
+++ b/src/platforms/gbm-kms/server/kms/quirks.cpp
@@ -146,7 +146,9 @@ public:
     }
 
 private:
-    std::unordered_set<std::string> drivers_to_skip;
+    // Mir's gbm-kms support isn't (yet) working with nvidia
+    // Mir's gbm-kms support isn't (yet) working with evdi
+    std::unordered_set<std::string> drivers_to_skip = { "nvidia", "evdi" };
     std::unordered_set<std::string> devnodes_to_skip;
     // We know this is currently useful for virtio_gpu
     std::unordered_set<std::string> skip_modesetting_support = { "virtio_gpu" };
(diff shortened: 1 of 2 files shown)


RAOF (changes requested):
Hm. So, there's currently no way to *override* these defaults.

For the `skip_modesetting_support` default quirk this is fine, as it just skips a check that would prevent Mir from considering the device.

These new default *prevent* Mir from starting on a device; having them non-overridable means that we'll need to rebuild Mir in order to test these devices.

I think that adding these default quirks is useful, but we shouldn't do it until there's a way of overriding them.

AlanGriffiths:
> These new default _prevent_ Mir from starting on a device; having them non-overridable means that we'll need to rebuild Mir in order to test these devices.

Yeah, I wondered how strongly you'd feel about that.



> I think that adding these default quirks is useful, but we shouldn't do it until there's a way of overriding them.

Coming right up

src/platforms/gbm-kms/server/kms/quirks.cpp
```diff
@@ -146,7 +153,9 @@ class mgg::Quirks::Impl
     }
 
 private:
-    std::unordered_set<std::string> drivers_to_skip;
+    // Mir's gbm-kms support isn't (yet) working with nvidia
+    // Mir's gbm-kms support isn't (yet) working with evdi
+    std::unordered_set<std::string> drivers_to_skip = { "nvidia", "evdi" };
     std::unordered_set<std::string> devnodes_to_skip;
     // We know this is currently useful for virtio_gpu
     std::unordered_set<std::string> skip_modesetting_support = { "virtio_gpu" };
```
Saviq, line 161: Do we want tests ensuring these defaults?

Saviq (approved):
Your call on the testing of the default quirks.

bors delegate+

AlanGriffiths:
I don't think testing the defaults gains anything. It just duplicates some strings.

But maybe there should also be a way to list the defaults? But that can await a future PR...

bors merge

PR #2559, opened 2022-08-16
Use Clang parser for generating docs

AlanGriffiths:
Mostly this makes little difference.

But there are places (like many of the source listings) where the output improves.


Saviq (approved):
ACK

bors merge

PR #2560, opened 2022-08-16
Add a libmiral-bin metapackage

AlanGriffiths:
We have the following in a scattering of downstreams:
```yaml
    stage-packages:
      - try:
          - libmiral5
      - else:
          - libmiral4
```
Having a stable name will avoid that in future

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -484,7 +484,7 @@ Section: libdevel
 Architecture: linux-any
 Multi-Arch: same
 Pre-Depends: ${misc:Pre-Depends}
-Depends: libmiral5 (= ${libmiral5:Version}),
+Depends: libmiral5,
 Description: MirAL binary metapackage
  MirAL provides an ABI-stable abstraction layer for Mir based shells,
  insulating them from mirserver ABI breaks.
--- a/debian/control
+++ b/debian/control
@@ -479,12 +479,25 @@ Description: Display server for Ubuntu - ABI preserving abstraction layer
  .
  Contains the shared library containing MirAL abstraction layer.
 
-Package: libmiral-dev
+Package: libmiral-bin
 Section: libdevel
 Architecture: linux-any
 Multi-Arch: same
 Pre-Depends: ${misc:Pre-Depends}
 Depends: libmiral5 (= ${libmiral5:Version}),
+Description: MirAL binary metapackage
+ MirAL provides an ABI-stable abstraction layer for Mir based shells,
+ insulating them from mirserver ABI breaks.
+ .
+ This package depends on the latest shared library containing MirAL abstraction
+ layer.
+
+Package: libmiral-dev
+Section: libdevel
+Architecture: linux-any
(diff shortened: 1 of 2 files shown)


Saviq:
@RAOF what's your Ubuntu developer hat say about this?

RAOF:
> @RAOF what's your Ubuntu developer hat say about this?

This is not something that we'd have in the archive; there's no reason why anyone would install `libmiral-bin` on a system - binaries automatically get dependencies on the correct `libmiral<SOVER>` package, and to *develop* with miral you install `libmiral-dev`.

There's a (small) cost to each binary package in the archive.

This seems like it *should* be solvable in snapcraft? But perhaps not by us :cry:

Saviq:
> This seems like it _should_ be solvable in snapcraft? But perhaps not by us cry

Right, we'd need a `${shlibs:Depends}` in snapcraft. It does even _monitor_ for missing dependencies and warns about it. Wonder if a `stage-packages: [ "$shared_libraries" ]` of some kind is in the minds of crafters.

AlanGriffiths:
> This is not something that we'd have in the archive

This solves a problem we don't have in the archive, as you go on to say that sorts out dependencies.

The problem happens because we change the package name in the PPA and build the dependency in the snap. If we built the dependency (`frame`, say) as a deb in the PPA that would also avoid the problem.

RAOF:
> > This seems like it _should_ be solvable in snapcraft? But perhaps not by us cry
> 
> Right, we'd need a `${shlibs:Depends}` in snapcraft. It does even _monitor_ for missing dependencies and warns about it. Wonder if a `stage-packages: [ "$shared_libraries" ]` of some kind is in the minds of crafters.

Right. It *is* kinda weird that it'll list off all the libraries missing from -stage that are *right there* in the build tree and not do anything but warn about them :woman_shrugging:

RAOF:
> > This is not something that we'd have in the archive
> 
> This solves a problem we don't have in the archive, as you go on to say that sorts out dependencies.
> 
> The problem happens because we change the package name in the PPA and build the dependency in the snap. If we built the dependency (`frame`, say) as a deb in the PPA that would also avoid the problem.

Yeah, I understand the problem this solves. I just don't think we (as in, Ubuntu archive people) currently accept imposing (individually small!) costs on all users of the Ubuntu archive in order to make it easier to build snaps.

AlanGriffiths:
Don't we have different packaging for PPA and Ubuntu Archive?

Saviq:
As discussed, I'm -1 on this. Maintaining it in the long run would be more work than bumping the package version once every so often. And it's really Snapcraft that should have a solution for this.

Saviq:
Filed a bug with Snapcraft: [LP#1987421](https://launchpad.net/bugs/1987421).

AlanGriffiths:
> As discussed, I'm -1 on this.

OK, I have the feedback I sought

PR #2561, opened 2022-08-17
More default quirks

AlanGriffiths:
We're already skipping the modesetting check in Frame and mir-kiosk, just consolidating these as defaults:

* RPi3b: vc4-drm
* RPi4: vc4-drm & v3d


Saviq (approved):
Sure.

bors merge

PR #2569, opened 2022-08-22
Only call global_mock->XInitThreads() if global mock exists

### LINKED ISSUE #2568: mir_unit_tests crashes in initialization on Arch
It appears `call_init()` in dl-init calls `XInitThreads()`. This attempts to call into the global mock, but no mock exists yet so that crashes.

wmww:
Fixes #2568


AlanGriffiths (approved):
Weird that this is necessary, but it should be harmless

bors merge

AlanGriffiths:
bors retry

PR #2578, opened 2022-08-24
ObserverMultiplexer: allow simultaneous observations/unregisters

### LINKED ISSUE #2573: Deadlock while resizing Mir-on-X
It appears this is what's happening (only important parts shown):
## Wayland thread
```
...
#7  mir::RecursiveReadLock::RecursiveReadLock()
#8  mir::scene::SurfaceStack::for_each()
#9  mir::input::SurfaceInputDispatcher::find_target_surface()
#15 (anonymous namespace)::context_for_event()
#16 mir::input::SurfaceInputDispatcher::surface_moved()
#22 (anonymous namespace)::InputDispatcherSceneObserver::moved_to()
...
#30 std::function<void ()>::operator()() const
#31 (anonymous namespace)::ImmediateExecutor::spawn()
#32 mir::ObserverMultiplexer<mir::scene::SurfaceObserver>::WeakObserver::spawn()
#33 mir::ObserverMultiplexer<mir::scene::SurfaceObserver>::for_each_observer()
#34 mir::scene::BasicSurface::Multiplexer::moved_to()
#35 mir::scene::BasicSurface::set_streams()
#36 mir::scene::ApplicationSession::configure_streams()
#37 mir::shell::AbstractShell::modify_surface()
#39 mir::frontend::WlSurface::commit()
...
```
## Compositor thread
```
,,,
#8  std::unique_lock<std::recursive_mutex>::unique_lock()
#9  mir::ObserverMultiplexer<mir::scene::SurfaceObserver>::WeakObserver::lock()
...
#14 std::function<void ()>::operator()() const
#15 (anonymous namespace)::ImmediateExecutor::spawn()
#16 mir::ObserverMultiplexer<mir::scene::SurfaceObserver>::WeakObserver::spawn()
#17 mir::ObserverMultiplexer<mir::scene::SurfaceObserver>::for_each_observer()
#18 mir::scene::BasicSurface::Multiplexer::attrib_changed()
#19 mir::scene::BasicSurface::set_visibility()
#20 mir::scene::BasicSurface::configure()
#21 mir::scene::RenderingTracker::configure_visibility()
#22 mir::scene::RenderingTracker::active_compositors()
#23 mir::scene::SurfaceStack::update_rendering_tracker_compositors()
#24 mir::scene::SurfaceStack::unregister_compositor(void const*)
#25 mir::compositor::CompositingFunctor::operator()()::{lambda()#7}::operator()() const
#26 mir::raii::PairedCalls::~PairedCalls()
#27 mir::compositor::CompositingFunctor::operator()()
```
As you can see the surface stack sends a surface observation while it is locked, and the surface input dispatcher tries to use the surface stack from within a surface observation. These two actions combine to deadlock, specifically while the Mir-on-X window is resizing.

wmww:
Fixes #2573 

Allows multiple observations to be simultaneously dispatched. This means that invoking an observer doesn't block on other observations completing, even if an immediate executor is used. Also allows unregistering an observer from different threads simultaneously. All unregisteres block until all threads currently making observations either complete or call unregister.


RAOF (changes requested):
I think the use of `std::mutex` in `live_lock` is unsound, and should be replaced.

As mentioned on our call, I think this is sufficiently subtle to require a detailed, high-level description of:
1. The guarantees `ObserverMultiplexer` does and does not provide (eg: it is always safe to call `for_each_observer`, even in the presence of `immediate_executor` and `unregister_observer`, and things like "calls to `for_each_observer` are not synchronised, even in the presence of `linarising_executor`)
2. How it provides those guarantees

wmww:
I updated the PR to use a `condition_variable` instead of a `unique_lock`, which should address the undefined behavior concerns. I believe the new logic with a status enum is also more straightforward to understand.

I added a long comment describing whats going on, but feel free to request more specific documentation if you think that would be helpful.

wmww:
This is ready for re-review

RAOF (approved):
There are a couple of spelling nits, but this looks sound.

With the nits resolved,
bors delegate=wmww

Issue #2570, opened 2022-08-23
Wayland platform uses EGL 1.5 entry points


RAOF:
Some EGL implementations (still!) don't support EGL 1.5. We don't *really* need anything beyond 1.4, but
https://github.com/MirServer/mir/blob/eebf0ff47353b8d4c8f7636144cb4e3248be5afd/src/platforms/wayland/displayclient.cpp#L331
is an EGL 1.5 entrypoint that isn't provided by some EGL implementations. This prevents the Wayland platform from loading on EGL 1.4 implementations (even if they support `eglCreatePlatformWindowSurfaceEXT`, the equivalent EGL 1.4 extension).

This then makes the nested tests fail.

AlanGriffiths:
I'm slightly confused by the [Khronos docs](https://registry.khronos.org/EGL/sdk/docs/man/html/eglCreatePlatformWindowSurface.xhtml)

> Attribute EGL_GL_COLORSPACE is supported only if the EGL version is 1.5 or greater.
>
> Attributes EGL_RENDER_BUFFER, EGL_VG_ALPHA_FORMAT, and EGL_VG_COLORSPACE, and the corresponding attribute values, are supported only if the EGL version is 1.2 or greater. 

Which imply this entry point exists for EGL earlier than 1.5. (Although there is no statement on when it was introduced)

RAOF:
Urgh. The Khronos docs are docs [are buggy](https://github.com/KhronosGroup/EGL-Registry/issues/169) :disappointed:

The documentation for `eglGetPlatformDisplay` correctly describe it as new in EGL 1.5, and this is a prerequisite for `eglCreatePlatformWindowSurface` (since it is the only way of getting an EGL display with an explicitly specified, rather than implementation-defined, platform).

Those entrypoints also don't appear in the EGL 1.4 specification.

PR #2589, opened 2022-08-25
[Wayland platform] Don't use uses EGL 1.5 entry points

AlanGriffiths:
Fixes: #2570


Saviq (approved):
Looks good, and smoke and performance tests (so nested) pass on both gbm-kms and wayland platforms.

bors merge

PR #2590, opened 2022-08-26
Fix timestamp calibration 

### LINKED ISSUE #2577: [2.9] Weird scrolling in, for example Firefox
1. Start firefox/Wayland on a webpage large enough to scroll
2. Use the scroll wheel to scroll up or down

Expect: can scroll to the top and bottom
Actual: scrolling stops before reaching the top/bottom

Observation: on the first scroll event in the opposite direction the scrolling does reach the top/bottom. This was obvious when tracking the events in a terminal window with:

```
WAYLAND_DEBUG=client firefox 2>&1 | grep axis_discrete
```

_This is intermittent, but seems consistent within a run of a Mir server. Restarting the server "rolls the dice" again with around a 65% chance of seeing problems._

wmww:
Fixes #2577

Prevent timestamps in the future ever being sent, and use unmodified timestamps if they're close to Mir's internal timestamps.

### DIFF
--- a/src/server/input/default_event_builder.cpp
+++ b/src/server/input/default_event_builder.cpp
@@ -28,13 +28,11 @@ namespace mi = mir::input;
 mi::DefaultEventBuilder::DefaultEventBuilder(
     MirInputDeviceId device_id,
     std::shared_ptr<time::Clock> const& clock,
-    std::shared_ptr<mir::cookie::Authority> const& cookie_authority,
-    std::shared_ptr<mi::Seat> const& seat)
+    std::shared_ptr<mir::cookie::Authority> const& cookie_authority)
     : device_id(device_id),
       clock(clock),
       timestamp_offset(Timestamp::max()),
-      cookie_authority(cookie_authority),
-      seat(seat)
+      cookie_authority(cookie_authority)
 {
 }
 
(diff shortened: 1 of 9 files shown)


wmww:
An alternative to this would be to remove timestamp calibration entirely, but I think that would be unwise as I said in the issue.

AlanGriffiths:
> An alternative to this would be to remove timestamp calibration entirely

There are at least two ways to do this:

1. Use the supplied timestamp
2. Use `now()` regardless of any supplied timestamp

The more complex logic proposed here is clearly an improvement on the status quo, but continually recalibrating creates a number of "corner cases" we need to enumerate and assess.

AlanGriffiths (approved):
This is an improvement and solves the symptoms. I'll land this and propose a simplification separately

bors merge

RAOF:
I think we should do the simpler thing, and remove calibration (at least in the common code).

Particularly: I think we should standardise all Mir timestamps on `CLOCK_MONOTONIC`:
* libinput guarantees its event timestamps are on `CLOCK_MONOTONIC`
* It's easy to generate `CLOCK_MONOTONIC` timestamps
* DRM reports timestamps in `CLOCK_MONOTONIC`
* It's likely what client input generators are using.

So, I think that we could basically say "Mir timestamps are all `CLOCK_MONOTONIC`" *now*, and only need to delete the calibration code.

If it turns out some timestamp source does not (or cannot) produce timestamps on a `CLOCK_MONOTONIC` base, we can translate them in the code ingesting that source.

AlanGriffiths:
> So, I think that we could basically say "Mir timestamps are all `CLOCK_MONOTONIC`" _now_

Yes. We use `std::chrono::steady_clock`, which I believe is the same thing

PR #2591, opened 2022-08-26
Don't try to calibrate timestamps

AlanGriffiths:
Everyone should be using `steady_clock` (or equivalant)

### DIFF
--- a/src/server/input/default_event_builder.cpp
+++ b/src/server/input/default_event_builder.cpp
@@ -31,7 +31,6 @@ mi::DefaultEventBuilder::DefaultEventBuilder(
     std::shared_ptr<mir::cookie::Authority> const& cookie_authority)
     : device_id(device_id),
       clock(clock),
-      timestamp_offset(Timestamp::max()),
       cookie_authority(cookie_authority)
 {
 }
@@ -42,7 +41,7 @@ mir::EventUPtr mi::DefaultEventBuilder::key_event(
     xkb_keysym_t keysym,
     int scan_code)
 {
-    auto const timestamp = calibrate_timestamp(source_timestamp);
+    auto const timestamp = value_or_now(source_timestamp);
     auto const cookie = cookie_authority->make_cookie(timestamp.count());
     return me::make_key_event(
         device_id, timestamp, cookie->serialize(), action, keysym, scan_code, mir_input_event_modifier_none);
@@ -58,7 +57,7 @@ mir::EventUPtr mi::DefaultEventBuilder::pointer_event(
     const float x_axis_value = 0;
     const float y_axis_value = 0;
     std::vector<uint8_t> vec_cookie{};
-    auto const timestamp = calibrate_timestamp(source_timestamp);
+    auto const timestamp = value_or_now(source_timestamp);
     if (action == mir_pointer_action_button_up || action == mir_pointer_action_button_down)
     {
         auto const cookie = cookie_authority->make_cookie(timestamp.count());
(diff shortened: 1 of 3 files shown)


wmww (changes requested):
According [this discussion](https://github.com/MirServer/mir/pull/2590#issuecomment-1228254151) libinput uses the same time base as `steady_clock`, I'd like a source for that, but empirically it's correct. X11 also seems to use the same timestamps, but again, is this guaranteed anywhere source?

Unfortunately Wayland is not so kind:
> The time argument is a timestamp with millisecond granularity, with an undefined base.

All the timestamps in wayland.xml and also the virtual pointer/virtual keyboard protocols include this. That means even if all known clients use the same timestamps, it is conformant with the protocol to use any timestamp base. Therefore to support the protocol we need calibration for the wayland platform and our implementation of those protocols.

We could implement it separately, but that would involve code duplication. The easiest place to share the code is the event builder, where calibration already lives. I see no reason to manually disable it for libinput and X11, since there's logic in place to not modify timestamps if it's plausibly using the standard timestamp base.

RAOF:
So, [libinput using `CLOCK_MONOTONIC` is (undocumented) ABI](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/25#note_581841). DRM uabi is [documented](https://dri.freedesktop.org/docs/drm/gpu/drm-uapi.html#c.DRM_CAP_TIMESTAMP_MONOTONIC) to use `CLOCK_MONOTONIC` for vblank events.

While ([most](https://gitlab.freedesktop.org/wayland/wayland-protocols/blob/1.18/stable/presentation-time/presentation-time.xml#L100-103)) Wayland timestamps are defined to use an unspecified base, they do all need to use the *same* base (this *does* mean that the `virtual_input` protocol cannot be portably implemented as-is, at least if you ever need to synthesise a timestamp).

I'm not opposed to having timestamp-calibration code, I just don't think that events we *know* have correct timestamps should run through that code. Any timestamps that we ingest from unknown sources can be run through a `TimestampCalibrator` or whatever before being fed into the event machinery, and timestamps we know the base clock for we can feed directly into the event machinery.

wmww:
I suppose refactoring timestamp calibration into it's own class and only using it for some platforms would be ok, but I don't see the motivation since:
- The current code will not mess with timestamps that are close to `CLOCK_MONOTONIC` timestamps (and this is tested)
- There isn't a single source of timestamps that's actually documented to be `CLOCK_MONOTONIC` (AFAIK no timestamps sent over Wayland originate from DRM)
- Doing research on if each specific source definitely uses `CLOCK_MONOTONIC` seems like a waste of effort (is X1 documented to use it? even if the events are coming over a network from another machine? Is this worth caring about when the current code already solves it automatically?)

AlanGriffiths:
> I don't see the motivation

The current code updates the timestamps in a stateful manner, if that introduces problems they will be subtle and hard to reproduce. For example, debugging though this codepath will change timings and affect the behaviour.

So far as I see the problem is theoretical: in practice, if sources were to use different clocks it would cause confusion for clients (with or without the existing "correction").

To take the example of events coming from another machine: fluctuating network delays will lead to unpredictable "corrections" being applied to events.

> The current code will not mess with timestamps that are close to CLOCK_MONOTONIC timestamps

Yes, that's why I landed it, pending this discussion.

> the current code already solves it automatically

That is the point of this discussion, I'm not yet convinced. The current code does something, but it isn't clear that this "something" is a solution or an additional source of problems.

Why is it better than, for example, taking `now()`, or clamping to `(now()-delta, now()]`?

wmww:
I think I see where you're coming from. Using now() unless the timestamp is within an acceptable range is also not perfect, but it might be a simpler, less statefull and equally or more correct solution.

AlanGriffiths:
#2602 is good enough

Issue #1793, opened 2020-11-05
When moving GTK apps, titlebar grab "slips" at start


wmww:
The bug is *much* more pronounced if you run it in [Wayland Debug](https://github.com/wmww/wayland-debug) GDB mode but noticeable when run normally.

### To reproduce
- Open a GTK app (such as gedit)
- Left click down on a specific point on the titlebar (such as the corner of a letter in the window title)
- Drag the pointer in any direction

### Expected
GTK window stays still for a moment, then "snaps" to the pointer with the same relative position as when the mouse button was first pressed.

### Actual
GTK window stays still for a moment, then starts to move but does not snap to the original position relative to the cursor. This creates the feeling the cursor slipped a bit over it.

### Cause
We base the move off of where the cursor is when we process the move request, rather than where it was for the serial that the move request was made with. Related to #1792. This bug is also present in Sway, but not Weston. Have not tested GNOME.

wmww:
This appears to have been fixed by #2593, not quite sure why as that doesn't fully track serials

wmww:
It looks like in `BasicWindowManager::update_event_timestamp()` we ignore motion events, which means when we start a gesture with the last remembered event it's a button event not a motion. #2593 changes the gesture logic to pay attention to the position of the event that starts a gesture, so it fixes this issue.

#1792 is still present.

PR #2593, opened 2022-08-27
MinimalWindowManager: allow moving window from the maximized state

### LINKED ISSUE #2584: Add ability to move maximized windows
It's common that grabbing a maximized window by their title bar unmaximizes it, and allows the user to drag it. Mir doesn't do anything when grabbing a maximized window.

wmww:
Fixes #2584

Pretty quick, can definitely be improved later.

EDIT: I fixed the most pressing problem, which was the window not always staying under the cursor. This branch also now moves closer to fixing #1793 (although we still need to plumb through proper serial handling and input event remembering for that to matter)

### DIFF
--- a/src/miral/minimal_window_manager.cpp
+++ b/src/miral/minimal_window_manager.cpp
@@ -39,6 +39,13 @@ enum class Gesture
     touch_resizing,
 };
 
+auto pointer_position(MirPointerEvent const* event) -> mir::geometry::Point
+{
+    return {
+        mir_pointer_event_axis_value(event, mir_pointer_axis_x),
+        mir_pointer_event_axis_value(event, mir_pointer_axis_y)};
+}
+
 auto touch_center(MirTouchEvent const* event) -> mir::geometry::Point
 {
     auto const count = mir_touch_event_point_count(event);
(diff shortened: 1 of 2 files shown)


Saviq (commented):
Works fine, with one exception: say, every 5th move, I can't maximize again by double-clicking on the title bar until I move the window again (tested with Firefox).

My use case is usually moving maximized windows between two screens, so I grab and drag, and then immediately want to maximize again. Does that ring a bell?

wmww:
I can reproduce, and it seems like this consistently happens when you release the drag and then don't move the pointer at all before trying to maximize. You also can't re-drag from the exact same position. If you move the pointer at all, it works.

wmww:
Ah! if you start the drag from the left edge of the window it works. The problem is that on button release the window isn't informed that the cursor now has a new relative position, so when it gets the next double-click it thinks it's outside of the window.

wmww:
Latest commit should fix the maximize-without-moving problem

Saviq:
> Latest commit should fix the maximize-without-moving problem

@RAOF you had a bad feeling about this approach?

PR #2602, opened 2022-08-30
Simplify timestamp calibration

wmww:
Use the timestamp if it is reasonable, and use the measured current time otherwise. As discussed in #2591.


AlanGriffiths (approved):
bors merge

Issue #2585, opened 2022-08-25
Menus of panels don't close when losing (or don't lose) focus


Saviq:
Open a drop-down menu of mate-panel, then focus a window. You can see the panel menu staying on.
![frame_2022-08-25T11:31+02:00](https://user-images.githubusercontent.com/231077/186629998-fa14f81b-ba70-4169-943f-7b6bc0a2f855.png)

Using this [Miriway](https://github.com/Miriway/Miriway) config:

```ini
x11-window-title=Miriway
idle-timeout=180
ctrl-alt=t:miriway-terminal # Default "terminal emulator finder"

enable-x11=true

shell-component=swaybg -i /usr/share/backgrounds/warty-final-ubuntu.png
shell-component=mate-panel
meta=a:mate-panel --run-dialog
ctrl-alt=t:mate-terminal
```

wmww:
The issue is the the current implementation dismisses popups when they lose focus, however panels generally don't gain or loose focus. This can be tested by running a recent version of gtk-layer-demo. When keyboard is set to none (never focused) or exclusive (always focused), popups don't close when other windows are clicked. When it's set to on-demand they do.

Saviq:
That also means you can't navigate the menus with the keyboard, which would be another issue to fix?

wmww:
The various keyboard interactivity modes are defined in the layer shell protocol and AFAIK we correctly implement them. If a change in behavior there was desirable that would need to be patched in mate-panel.

wmww:
In my investigation I discovered #2604, which will likely be fixed at the same time

Saviq:
Fixed by #2606

PR #2606, opened 2022-09-01
Better grabbing popups

### LINKED ISSUE #2604: Popups dismissed by Mir when parent clicked
#2330 Fixed several previous and reoccurring issues (see #2324 for an upsettingly in-depth analysis), however it appears it's still not doing quite the same thing as other compositors.

The current Mir code roughly gives grabbing popups pointer focus but not keyboard focus. Therefore, when the user clicks on the parent window outside of a popup the click is consumed and the popup is dismissed. In wlroots and weston in this situation the parent gets the click events, and the popup is not automatically dismissed by the compositor.

To see a difference in behavior:
- Open Kate
- Open a menu and hover over an item with a submenu (the submenu will appear)
- Click
- On Mir, the submenu disappears
- On wlroots/weston the submenu does not disappear

In fixing this we must be mindful to not reintroduce #1941 or #2241. Those should be tested in WLCS, but I'll verify that.

wmww:
Fixes #2604 and #2585 by decoupling focus state from dismissing popup grabs. Draft because it doesn't have tests yet, but it should work.

### DIFF
--- a/src/include/server/mir/shell/abstract_shell.h
+++ b/src/include/server/mir/shell/abstract_shell.h
@@ -174,6 +174,8 @@ private:
     std::vector<std::weak_ptr<scene::Surface>> notified_active_surfaces;
     std::weak_ptr<scene::Surface> last_requested_focus_surface;
     std::weak_ptr<scene::Surface> notified_keyboard_focus_surface;
+    std::weak_ptr<scene::Surface> popup_parent;
+    std::vector<std::weak_ptr<scene::Surface>> grabbing_popups;
     std::shared_ptr<SurfaceConfinementUpdater> const surface_confinement_updater;
 
     void notify_active_surfaces(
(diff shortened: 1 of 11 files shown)


Saviq:
```
[ RUN ] AbstractShell.removing_focus_from_menu_closes_it
/spread/mir/tests/unit-tests/scene/test_abstract_shell.cpp:972: Failure
Actual function call count doesn't match EXPECT_CALL(mock_surface1, request_client_surface_close())...
Expected: to be called once
[ FAILED ] AbstractShell.removing_focus_from_menu_closes_it (0 ms)
```

Saviq (commented):
Empirically, this fixes #2585.

AlanGriffiths (approved):
> Draft because it doesn't have tests yet, but it should work.

Looks sensible, appears to work, and even has tests!

Still marked "draft" though

bors delegate

wmww:
> Still marked "draft" though

Huh? On my side it says I marked it ready for review 4 days ago.

bors r+

AlanGriffiths:
> Huh? On my side it says I marked it ready for review 4 days ago.

I see that too. Now. I wonder what I was looking at?

PR #2610, opened 2022-09-01
[MultiMonitorArbiter] track buffer users better

AlanGriffiths:
The compositor ID should not be added repeatedly when there is no new buffer available


wmww (approved):
LGTM
bors r+

Issue #1753, opened 2020-10-02
Incorrect rendering when a surface spans multiple outputs


wmww:
![a](https://user-images.githubusercontent.com/9331264/94972026-4ca00500-04d6-11eb-8007-07c1eff66a62.png)

In the screenshot I'm running Mir-on-X, but I've also observed the problem on mesa+KMS. It happens about 1/50 frames. To reproduce, I open a terminal, position it between the two outputs and press enter until it happens.

AlanGriffiths:
I've not been able to reproduce on my kit.

I doubt we have changes that affect this, but are you using release or master? Can you post a log including details of the graphics stack chosen?

wmww:
Current master running on my XPS. Log:
```
wmww@wmwwXPS:~/code/mir/build$ WAYLAND_DISPLAY=wayland-mir ./bin/miral-shell --x11-output 500x1024:600x1024
MIR_CLIENT_PLATFORM_PATH=./bin/../lib/client-modules/
MIR_SERVER_PLATFORM_PATH=./bin/../lib/server-modules/
LD_LIBRARY_PATH=:/usr/local/lib:/usr/local/lib:./bin/../lib
exec=./bin/miral-shell.bin
[2020-10-05 13:38:32.250200] <information> mirserver: Starting
[2020-10-05 13:38:32.250286] < - debug - > mirserver: Not trying logind: "DISPLAY" is set and X need not have claimed the VT
[2020-10-05 13:38:32.250407] < - debug - > mirserver: Not using Linux VT subsystem for session management: Failed to find the current VT
[2020-10-05 13:38:32.250429] < - debug - > mirserver: No session management supported
[2020-10-05 13:38:32.250455] <information> VT switch key handler: No VT switching support available: MinimalConsoleServices does not support VT switching
[2020-10-05 13:38:32.250677] <information> mircommon: Loading modules from: ./bin/../lib/server-modules/
[2020-10-05 13:38:32.250754] <information> mircommon: Loading module: ./bin/../lib/server-modules/graphics-wayland.so.18
[2020-10-05 13:38:32.250778] <information> mircommon: Loading module: ./bin/../lib/server-modules/graphics-gbm-kms.so.18
[2020-10-05 13:38:32.250790] <information> mircommon: Loading module: ./bin/../lib/server-modules/server-x11.so.18
[2020-10-05 13:38:32.250801] <information> mircommon: Loading module: ./bin/../lib/server-modules/graphics-eglstream-kms.so.18
[...]

PR #2614, opened 2022-09-02
Fix rendering that spans outputs

AlanGriffiths:
Fixes: #1753

### DIFF
--- a/src/platforms/common/server/buffer_from_wl_shm.cpp
+++ b/src/platforms/common/server/buffer_from_wl_shm.cpp
@@ -26,6 +26,8 @@
 #include <boost/throw_exception.hpp>
 #include <mutex>
 #include <atomic>
+#include <thread>
+#include <unordered_set>
 
 #include <GLES2/gl2.h>
 
@@ -319,13 +321,14 @@ public:
 
     void bind() override
     {
+        auto const thread_id = std::this_thread::get_id();
         ShmBuffer::bind();
         std::lock_guard lock{upload_mutex};
-        if (!uploaded)
+        if (threads_uploaded.find(thread_id) == end(threads_uploaded))
         {
             auto const mapping = map_readable();
             upload_to_texture(mapping->data(), mapping->stride());
-            uploaded = true;
+            threads_uploaded.insert(thread_id);
         }
     }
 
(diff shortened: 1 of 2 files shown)


AlanGriffiths:
Note that this may well be the wrong way to solve the problem: someone with a better understanding of GL may well have a correct idea.

wmww (approved):
Looks reasonable to me and empirically fixes the problem (yay!). I'll let @RAOF have a look since he's the one that knows this part of the code.

RAOF (changes requested):
Urgh.

So, this is likely to work in practise (although it involves unnecessary texture uploads), but I *think* that, *technically*, uploading to the texture invalidates it across *all* GL contexts in the absence of other *GL* synchronisation. In practise this will almost certainly work, because we're uploading identical data, but weird GPU memory quirks might stymie that?

I *think* `glFinish()` after the `upload_to_texture()` would work (although I'm *still* not sure if that's mandated to DTRT across EGLContexts), but the thing that's *guaranteed* to work is an EGL fence ([base extension](https://registry.khronos.org/EGL/extensions/KHR/EGL_KHR_fence_sync.txt) and [GPU-side-wait extension](https://registry.khronos.org/EGL/extensions/KHR/EGL_KHR_wait_sync.txt)).

AlanGriffiths:
> I _think_ `glFinish()` after the `upload_to_texture()` would work

Empirically that "works", but should it be in, rather than after, the `upload_to_texture()`? (The other  call site in `MemoryBackedShmBuffer::bind()` could have the same issue.)

AlanGriffiths:
> the thing that's _guaranteed_ to work is an EGL fence ([base extension](https://registry.khronos.org/EGL/extensions/KHR/EGL_KHR_fence_sync.txt) and [GPU-side-wait extension](https://registry.khronos.org/EGL/extensions/KHR/EGL_KHR_wait_sync.txt)).

I think it more likely that `glFinish()` DTRT than me getting this magic right: I've tried reading the docs, but it looks non-trivial to insert `egl*` calls into this context, and I'm not sure how to use them.

RAOF (approved):
I *think* it will be easy enough to use fences here, but I'm happy to poke at that later and I guess we'll need the `glFinish` as a fallback in case the relevant extensions aren't supported.

It is surprisingly hard to find documentation on cross-thread, cross-context synchronisation, but `glFinish` *has* to be a big enough hammer.

bors r+

PR #2616, opened 2022-09-05
Avoid racy access to shm buffers

AlanGriffiths:


RAOF (commented):
This *should* do nothing. This (very slightly) perturbs the timing of when we send the `wl_surface.frame()` callbacks, but this is only an advisory to the client that it should start rendering a new frame; we haven't (or shouldn't!) have sent the `wl_buffer.release()` event which would tell the client that this shm buffer is free to reuse.

PR #2617, opened 2022-09-05
Ignore wl_pointer.set_cursor if incorrect serial is given

### LINKED ISSUE #2532: Verify wl_pointer.set_cursor() serial matches latest wl_pointer.enter() serial
The latest wayland.xml adds
> The serial parameter must match the latest wl_pointer.enter serial number sent to the client. Otherwise the request will be ignored.

Since it's all in the same file, storing this serial and checking it should be trivial.

wmww:
Fixes #2532


AlanGriffiths (approved):
bors merge

Issue #318, opened 2018-04-09
Wayland interfaces fail to verify `serial` parameter


RAOF:
There are a bunch of Wayland interfaces which take an input-event serial; this is similar to our MirCookie approach, but only for within-process (or, specifically, within `wl_display`) attestation.

The current implementations ignore the `serial` parameter. We should verify it in much the same way as we verify the MirCookie contains the most-recent timestamp.

AlanGriffiths:
This is non-trivial to implement but ought to be done before diving into "Drag & Drop"

AlanGriffiths:
There are a bunch of Wayland extensions that do check, is this more-or less resolved?

wmww:
| request | serial verified? |
|-|-|
| `wl_data_offer`.`accept` | ❌ (unimplemented) |
| `wl_data_device`.`start_drag` | ❌ (unimplemented) |
| `wl_data_device`.`set_selection` | ❌ |
| `wl_shell_surface`.`pong` | ❌ (unimplemented) |
| `wl_shell_surface`.`move` | ❌ |
| `wl_shell_surface`.`resize` | ❌ |
| `wl_shell_surface`.`set_popup` | ❌ |
| `wl_pointer`.`set_cursor` | ❌ (#2532) |
| `xdg_wm_base`.`pong` | ❌ (unimplemented) |
| `xdg_surface`.`ack_configure` | ❌ (unimplemented) |
| `xdg_toplevel`.`show_window_menu` | ❌ (unimplemented) |
| `xdg_toplevel`.`move` | ❌ |
| `xdg_toplevel`.`resize` | ❌ |
| `xdg_popup`.`grab` | ❌ |
| `zxdg_shell_v6`.`pong` | ❌ (unimplemented) |
| `zxdg_surface_v6`.`ack_configure` | ❌ (unimplemented) |
| `zxdg_toplevel_v6`.`show_window_menu` | ❌ (unimplemented) |
| `zxdg_toplevel_v6`.`move` | ❌ |
| `zxdg_popup_v6`.`grab` | ❌ |
| `zwp_text_input_v1`.`commit_state` | ✅ |
| `zwp_text_input_v2`.`update_state` | ✅ |
| `zwp_input_method_v2`.`commit` | ✅ |
| `zwlr_layer_surface_v1`.`ack_configure` | ✅ |

As you can see, we do not check any serials in the core Wayland protocol, or either supported version of XDG shell. We do check and use serials in the newer protocols.

[...]

PR #2622, opened 2022-09-06
Verify serials for move/resize requests

wmww:
Instead of blindly using the lastest timestamps when requesting a move/resize, this PR pulls the timestamp from the input event with the matching serial. Infrastructure to map track serials and map them to events is also put into place. This is designed so that in the future we can easily send the events (instead of the timestamps) to the shell, but we're not ready for that yet. Does not fix fully #1792, but is a step in that direction. Fixes some of #318. 

For XWayland, there's no obvious way to match events to a move/resize request, so we simply use the last applicable event.


wmww:
We're getting some weird order-of-destruction WlClient bugs. It seems to all be single-threaded and not racy, so that's nice. Will investigate tomorrow.

wmww:
Ah, brilliant, we get to flip a WLCS test from expect fail to expect success.

src/server/frontend_wayland/wl_client.h
```diff
@@ -53,6 +57,7 @@ class WlClient : public wayland::LifetimeTracker
         std::function<void(WlClient&)>&& client_created_callback);
 
     static auto from(wl_client* client) -> WlClient*;
+    static auto from_or_throw(wl_client* client) -> WlClient&;
```
AlanGriffiths, line 60: The naming and semantics of this seem a bit odd. I'd expect the name to reflect the contract and an exception being thrown to indicate the contract cannot be met.

There are some exceptions to this rule (`std::rethrow()` for example), but I don't see the motivation here.

Looking at the implementation, we throw a `logic_error` which also suggests that an exception indicates a precondition failure.

Looking at the use, we either take the address immediately, or simply use a member function. That is, the requirement is is simply a stronger version of the `WlClient::from()` post-condition.

In fact, why can't we just make "result is not `nullptr`" a post-condition of `WlClient::from()`? AFAICS it would need a `logic_error` (or worse) for this to return `nullptr`.

AlanGriffiths (changes requested):
This looks mostly good.

I think our existing code already uses the idea that `WlClient::from()` cannot sensibly return `nullptr`, so I think it would be better to formalise that as an explicit post-condition than to introduce a new function with that property.

I've tried inserting `fatal_error` into `WlClient::from()` to avoid returning `nullptr`: all tests still pass, and a short experiment with assorted clients exposed no problems.

wmww:
Ok for now. There is a bit of an oddity in the libwayland API where the client destroy listener is called before all of the resources are cleaned up, meaning the client can be non-existent while it still has resources. Right now I think the logic is sound, but a bit fragile. I'm going to work on a solution involving shared ownership of the client class, but that can be a follow up PR.

src/server/frontend_wayland/wl_client.h
```diff
@@ -52,7 +56,7 @@ class WlClient : public wayland::LifetimeTracker
         std::shared_ptr<SessionAuthorizer> const& session_authorizer,
         std::function<void(WlClient&)>&& client_created_callback);
 
-    static auto from(wl_client* client) -> WlClient*;
+    static auto from(wl_client* client) -> WlClient&;
```
AlanGriffiths, line 59: I wouldn't have changed this signature, just document the post-condition.

src/server/frontend_wayland/wl_client.cpp
```diff
@@ -112,25 +114,58 @@ void mf::WlClient::setup_new_client_handler(
...
+    {
+        fatal_error("WlClient::from(): listener is null");
+    }
     auto ctx = ClientCtx::from(listener);
-    return ctx ? ctx->client.get() : nullptr;
+    if (!ctx)
+    {
+        fatal_error("WlClient::from(): ctx is null");
+    }
+    return *ctx->client.get();
```
AlanGriffiths, line 133: How sure are we that `client.get()` will never be `nullptr`?

AlanGriffiths (approved):
WFM. Just a couple of thoughts

bors delegate

src/server/frontend_wayland/wl_client.cpp
wmww: I believe fairly sure

src/server/frontend_wayland/wl_client.h
wmww: Makes the expectation a little more explicit

Saviq:
Ran out of CPUs.

bors retry

PR #2623, opened 2022-09-06
Modernize touch events

wmww:
This renames `ContactState` to `TouchContact`, and versions it similar to how scroll axis is versioned. The old `ContactState` struct is kept around for ABI stability, and is aliased to `TouchContactV1`

The motivation for this is that I want to add local position to touch events (and pointer events) as part of my work on #1792, but to do that without breaking ABI I need more flexibility in the touch contact struct.


AlanGriffiths (approved):
bors merge

Issue #2626, opened 2022-09-08
Implement Modifier+Drag window move gesture


Saviq:
While #2625 is debatable ("I _want_ to move my window under the panel!"), we're missing a common pattern where you can Modifier+Drag windows to recover them from most corners of the world.

AlanGriffiths:
To clarify, this _is_ [implemented in the examples](https://github.com/MirServer/mir/blob/aa0a01e1b9d46cebf47650243dbf2245c5e9025f/examples/example-server-lib/floating_window_manager.cpp#L94), but not in `MinimalWindowManager` nor Miriway.

Moving the logic to `MinimalWindowManager` would fix both. (Do we want to hard-code the "Modifier" as `Alt`?)

Saviq:
> Do we want to hard-code the "Modifier" as `Alt`?

As long as compositors can override? Meta+Drag is another common gesture.

AlanGriffiths:
> As long as compositors can override? Meta+Drag is another common gesture.

The question is _how_ do we enable overriding:

Compositors don't _need_ to use `MinimalWindowManager` so they can always override that way, or they can use `MinimalWindowManager` and override just `handle_pointer_event()` (which is how the examples do it now).

For example, adding a virtual function `is_start_of_drag_gesture(MirPointerEvent const* event) => bool` breaks MirAL ABI...

PR #2627, opened 2022-09-08
[MinimalWindowManager] Move pointer-drag implementation from examples

AlanGriffiths:
Fixes: #2626

### DIFF
--- a/debian/libmiral5.symbols
+++ b/debian/libmiral5.symbols
@@ -126,9 +126,9 @@ libmiral.so.5 libmiral5 #MINVER#
  (c++)"miral::Output::used() const@MIRAL_3.6" 3.6.0
  (c++)"miral::Output::valid() const@MIRAL_3.6" 3.6.0
  (c++)"miral::Output::~Output()@MIRAL_3.6" 3.6.0
+ (c++)"miral::PrintTo(miral::Window const&, std::basic_ostream<char, std::char_traits<char> >*)@MIRAL_3.6" 3.6.0
  (c++)"miral::PrependEventFilter::PrependEventFilter(std::function<bool (MirEvent const*)> const&)@MIRAL_3.6" 3.6.0
  (c++)"miral::PrependEventFilter::operator()(mir::Server&)@MIRAL_3.6" 3.6.0
- (c++)"miral::PrintTo(miral::Window const&, std::basic_ostream<char, std::char_traits<char> >*)@MIRAL_3.6" 3.6.0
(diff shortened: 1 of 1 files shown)


Saviq (approved):
Works for me!

src/miral/minimal_window_manager.cpp
```diff
@@ -91,8 +91,12 @@ struct miral::MinimalWindowManager::Impl
     bool handle_touch_event(MirTouchEvent const* event);
 
     void apply_resize_by(Displacement movement);
+
+    static MirInputEventModifier pointer_drag_modifier;
```
wmww, line 95: Why is this static? Why not a property in the window manager instance?

include/miral/miral/minimal_window_manager.h
```diff
@@ -72,6 +73,12 @@ class MinimalWindowManager : public WindowManagementPolicy
     /// Raises newly focused window
     void advise_focus_gained(WindowInfo const& window_info) override;
 
+    /// Change the modifer used to identify a window drag gesture
+    /// The default is mir_input_event_modifier_alt
+    /// \warning this is only expected to be called during initialization, there is no thread synchronisation
+    /// \remark Since MirAL 3.7
+    static void pointer_drag_modifier(MirInputEventModifier new_modifier);
```
wmww, line 80: I would call this `set_pointer_drag_modifier()` to clarify it's purpose. There should be some way to use this function to disable the behavior but if there is it's not documented.

wmww (changes requested):
It works, and surprisingly it seems to be disabled on the floating window manager (not sure why). I do have a few nits about the approach.

include/miral/miral/minimal_window_manager.h
AlanGriffiths: I don't see a need for this to be a mechanism to disable the behaviour, there are many other ways to achieve that.

src/miral/minimal_window_manager.cpp
AlanGriffiths: Yeah, wasn't the best idea. Have reworked

AlanGriffiths:
> surprisingly it seems to be disabled on the floating window manager

I don't see that, what did you do?

wmww:
Sorry, I think I was being confusing. I meant that alt+drag (correctly) does nothing on the tiling window manager. I suppose that shouldn't be surprising since titlebar drag also doesn't move windows.

wmww (approved):

Saviq (approved):
Still works.

bors merge

Issue #2592, opened 2022-08-26
CapsLock state gets stuck


Saviq:
From time to time my CapsLock gets stuck on. I can still type lower case by keeping Shift down, so it's different to #2025.

Shifting focus _sometimes_ fixes it.

It feels related to idling out to some extent.

Saviq:
Having encountered this again it seems to be triggered by shifting focus in the first place.

Turn CapsLock on, then Alt+Tab to another window. CapsLock is on, and can't turn it off. Going back to the original app, CapsLock sometimes fixes.

AlanGriffiths:
OK, here's a sequence that seems repeatable (on `miral-app`):

1. Open a **terminal**
2. start **gedit**
3. Press CapsLock and type a couple of characters (show lowercase)
4. Press CapsLock and type a couple of characters (show uppercase)
5. `Alt-Tab` to **terminal**
6. Type a couple of characters (show uppercase)
7. Press CapsLock and type a couple of characters (_show uppercase_ - expect lowercase)
8. 4. `Alt-Tab` to **gedit**
9. type a couple of characters (show _lowercase_ - WTF?)
10. _goto 4_

AlanGriffiths:
Urgh, this is another maze of twisty codepaths.

I've established that our logic for not sending modifiers when they haven't changed does not "do the right thing", but not how to fix it correctly, nor whether fixing it fixes actually this bug.

Time for a weekend and revisiting next week!

AlanGriffiths:
Ah! I've found the place this is going wrong. Just need to work out the best way to fix it up.

Saviq:
Fixed by #2631

AlanGriffiths:
> Fixed by https://github.com/MirServer/mir/pull/2631

Not yet

PR #2631, opened 2022-09-12
Track keyboard modifiers better

AlanGriffiths:
Fixes: #2592

### DIFF
--- a/src/server/frontend_wayland/keyboard_helper.cpp
+++ b/src/server/frontend_wayland/keyboard_helper.cpp
@@ -149,6 +149,7 @@ void mf::KeyboardHelper::set_keymap(std::shared_ptr<mi::Keymap> const& new_keyma
 {
     if (!new_keymap || new_keymap == current_keymap)
     {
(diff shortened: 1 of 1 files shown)


Saviq (commented):
Could we have a test for this?

AlanGriffiths:
> Could we have a test for this?

I agree this is desirable, and would reduce our technical debt.

But this isn't simple to achieve: there are no existing tests around this area and the behaviour results from the interaction of several components with ill-defined responsibilities. Identifying a sensible scope for a test fixture that would not be fragile is a significant chunk of work. I don't feel this fix, in isolation, warrants tackling that.

Saviq (approved):
> But this isn't simple to achieve: there are no existing tests around this area and the behaviour results from the interaction of several components with ill-defined responsibilities. Identifying a sensible scope for a test fixture that would not be fragile is a significant chunk of work. I don't feel this fix, in isolation, warrants tackling that.

Hence me asking.

Fixes!

bors merge

PR #2633, opened 2022-09-13
Add quirks for false-positive compiler warnings

RAOF:
We semi-regularly run into toolchain bugs where a warning false-positives on our test code¹. Add the minimal infrastructure to let us disable `-Werror` on the smallest easily-accessible code unit, and quirk off `Werror` on the current set of gcc-12 false-positives.

¹: GTest is apparently a compiler fuzzing suite

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,7 @@ if(MIR_FATAL_COMPILE_WARNINGS)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
 endif()
 
-set(MIR_COMPILER_QUIRKS "" CACHE STRING "List of compiler quirks to enable")
+set(MIR_COMPILER_QUIRKS "" CACHE STRING "Semicolon-separated list of compiler quirks to enable")
 
 include(CheckCXXCompilerFlag)
 check_cxx_compiler_flag(-Wmismatched-tags HAS_W_MISMATCHED_TAGS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,6 +74,8 @@ if(MIR_FATAL_COMPILE_WARNINGS)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
 endif()
 
+set(MIR_COMPILER_QUIRKS "" CACHE STRING "List of compiler quirks to enable")
+
 include(CheckCXXCompilerFlag)
 check_cxx_compiler_flag(-Wmismatched-tags HAS_W_MISMATCHED_TAGS)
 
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -74,6 +74,8 @@ if(MIR_FATAL_COMPILE_WARNINGS)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
 endif()
 
+set(MIR_COMPILER_QUIRKS "" CACHE STRING "List of compiler quirks to enable")
```
Saviq, line 77:
```suggestion
set(MIR_COMPILER_QUIRKS "" CACHE STRING "Semicolon-separated list of compiler quirks to enable")
```

tests/mir_test_framework/CMakeLists.txt
```diff
@@ -61,6 +61,12 @@ set_property(
     SOURCE udev_environment.cpp
     PROPERTY COMPILE_OPTIONS -Wno-variadic-macros)
 
+if ("libinput_environment.cpp:restrict" IN_LIST MIR_COMPILER_QUIRKS)
+  set_property(
+    SOURCE libinput_environment.cpp
+    PROPERTY COMPILE_OPTIONS -Wno-error=restrict)
+endif()
```
Saviq, line 68: This feels like too wide a net, should we take the compiler ID / version into account? This way we get feedback from other toolchains whether this becomes / stops being a problem?

tests/unit-tests/CMakeLists.txt
```diff
@@ -142,6 +142,13 @@ set_property(
   SOURCE graphics/test_platform_prober.cpp
   PROPERTY COMPILE_OPTIONS -Wno-variadic-macros)
 
+if ("test_linearising_executor.cpp:use-after-free" IN_LIST MIR_COMPILER_QUIRKS)
+  set_property(
+    SOURCE test_linearising_executor.cpp
+    PROPERTY COMPILE_OPTIONS -Wno-error=use-after-free  # riscv64 spuriously warns here
+  )
+endif()
```
Saviq, line 150: Ditto.

Saviq (changes requested):

tests/mir_test_framework/CMakeLists.txt
RAOF: My idea here is to punt *when* to enable these quirks to configure time. I can modify `debian/rules` to additionally only enable these on gcc-12.

Saviq: > My idea here is to punt _when_ to enable these quirks to configure time. I can modify `debian/rules` to additionally only enable these on gcc-12.

Sure, that works, too.

Saviq (approved):
OK I went on a rabbit chase to use this for the pre-compiled headers issue ([LP#1983852](https://bugs.launchpad.net/ubuntu/+source/gcc-12/+bug/1983852)), but at 10 files and counting I don't think it's worth it.

bors merge

PR #2634, opened 2022-09-13
CMake: Use non-fat LTO

RAOF:
Fat LTO objects slow down the build. We don't *need* fat LTO objects, as we don't ship any LTO'd static archives.


wmww (approved):
LGTM
bors r+

PR #2640, opened 2022-09-14
Only call extension filter once per app per extension

### LINKED ISSUE #2209: MirAL extension filter invoked multiple times for the same app and same protocol

wmww:
Updated version of #2211. Fixes #2209.

### DIFF
--- a/src/server/frontend_wayland/wayland_connector.cpp
+++ b/src/server/frontend_wayland/wayland_connector.cpp
@@ -239,8 +239,7 @@ mf::WaylandConnector::WaylandConnector(
       executor{std::make_shared<WaylandExecutor>(wl_display_get_event_loop(display.get()))},
       allocator{allocator_for_display(allocator, display.get(), executor)},
       shell{shell},
-      extensions{std::move(extensions_)},
-      extension_filter{extension_filter}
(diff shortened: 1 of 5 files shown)


Saviq:
@wmww there are a couple conflicts here now, if you could fix.

wmww:
I think they'll go away once #2642 lands

wmww:
Ayy conflicts gone

RAOF (commented):
I'm not sure if this is a good idea?

It seems like fair amount of extra complexity? Documenting that the filter function may be called multiple times for the same client/protocol pair seems simpler, and it's not clear to me how usage of the filter API is made more difficult by the requirement that it be callable multiple times.

wmww:
If there's any performance penalty to calling the filter, *someone* should be caching the results. I don't think pushing that complexity to the shell is particularly more attractive.

AlanGriffiths:
As I commented in #2211, I see no need for this and I also don't see why anyone would care about #2209, nor why adding a doc comment wouldn't fix any hypothetical confusion

PR #2642, opened 2022-09-15
Verify serials again

wmww:
This reverts commit 71a40e7, reversing changes made to 387705a. I fixed the obvious uninitialized use in xwayland_surface.cpp that seemed to cause the PPA failure. Not sure why neither local builds nor CI picked up on that but PPA did.

PR #2644, opened 2022-09-15
Improve Client class lifetime management

wmww:
Libwayland notifies us that clients are destroyed before it cleans up all of the client's resources. The `wl_client` isn't freed until that is done. Previously, our WlClient class being destroyed before the resources, and logic was needed to prevent using using it after. This Makes `Client` an interface in `mir/wayland` and had `Resource` hold a shared pointer to it. This way the client isn't destroyed until it's last resource is. It also makes the client class more readily available.


wmww:
Draft for now. I'll fix the merge conflicts once #2642 and #2635 have landed.

include/wayland/mir/wayland/resource.h
```diff
@@ -26,19 +26,25 @@ namespace mir
...
     struct Version
     {
     };
 
     Resource(wl_resource* resource);
+    virtual ~Resource();
 
     wl_resource* const resource;
-    wl_client* const client;
+    Client& client;
```
AlanGriffiths, line 47: I'm not entirely convinced that making `client` look like a member variable is the best (I'd be happier with a pointer). But that's a point we often disagree on and not worth blocking over

AlanGriffiths (approved):
There's a trivial merge conflict...

bors delegate

include/wayland/mir/wayland/resource.h
AlanGriffiths: I also note that almost all the use of client is with `client.raw_client()`: maybe that could be encapsulated as a helper function?

include/wayland/mir/wayland/resource.h
```diff
@@ -26,19 +26,25 @@ namespace mir
 {
 namespace wayland
 {
+class Client;
+
 class Resource
     : public virtual LifetimeTracker
 {
+private:
+    std::shared_ptr<Client> owned_client;
```
AlanGriffiths, line 35:
```suggestion
    std::shared_ptr<Client> const owned_client;
```

include/wayland/mir/wayland/resource.h
wmww: I tend to agree about refs/ptrs, changing.

Another option instead of `raw_client()` is to make a `wl_client*` operator on `mw::Client`, but we can put this off to a future PR.

wmww:
No one seems to be rushing to review and I got a delegate, so

bors r+

Issue #2613, opened 2022-09-02
miral does not expose Alarm functionality from MainLoop


graysonguarino:
MirAL currently has no wrapper for `mir::time::Alarm` and there is no way to create an alarmed callback in a similar vein to `miral::MirRunner::add_start_callback()` and `miral::MirRunner::add_stop_callback()`. This could come in handy for clients that need to delay function calls.

AlanGriffiths:
Quick note that the similarity to `add_start_callback()` and `add_stop_callback()` is limited: these are called before running the server, whereas the primary usage of alarms is whilst the server is running.

What is needed is something along the lines of:

```c++
namespace miral
{
class AlarmFactory
{
...
   auto create_alarm(std::function<void()> const& callback) -> std::unique_ptr<Alarm>;
};
```

Saviq:
Duplicate of #2543

graysonguarino:
Closing this again as it actually may be unnecessary. Similar functionality can be achieved with `timerfd` which can be utilized once #2646 is completed.

PR #2646, opened 2022-09-19
Expose wait features

### LINKED ISSUE #2543: miral does not expose wait features based on MainLoop
We probably don't want to expose the MainLoop to server code, but it may want a way to wait on FDs or signals.

E.g. waiting for clients to exit (SIGCHLD)
E.g. waiting for an inotify file descriptor

graysonguarino:
Fixes #2543 by exposing registering/deregistering of file descriptors on the MainLoop and registering of signals.


wmww (approved):
LGTM, I'll let someone who understood the motivation for this feature better give the final approval.

graysonguarino:
@RAOF mentioned it would be wise to expose a handle for registering file descriptors that deregisters once destroyed. Revering to a draft until this is properly implemented.

graysonguarino:
The rationale for the "Introduce FdManager" change was too large to fit into a commit message.

There was no way to properly test that the destruction of handlers had the correct behavior so @wmww suggested creating an `FdManager` to handle the creation and destruction of all `FdHandles`.

include/miral/miral/fd_manager.h
```diff
@@ -0,0 +1,84 @@
+/*
```
AlanGriffiths, line 1: This file is *not* for use by client code, so it should be in `src/miral`, not `include/miral`

include/miral/miral/runner.h
```diff
@@ -90,6 +105,14 @@ class MirRunner
     MirRunner& operator=(MirRunner const&) = delete;
     struct Self;
     std::unique_ptr<Self> const self;
+
+    struct SignalInfo
+    {
+        std::initializer_list<int> signals;
+        std::function<void(int)> const& handler;
+    };
```
AlanGriffiths, line 113: This struct is not used in the public interface: it should not be declared in the public interface.

src/miral/CMakeLists.txt
```diff
@@ -53,6 +53,7 @@ add_library(miral SHARED
     cursor_theme.cpp                    ${miral_include}/miral/cursor_theme.h
     display_configuration.cpp           ${miral_include}/miral/display_configuration.h
     external_client.cpp                 ${miral_include}/miral/external_client.h
+    fd_manager.cpp                      ${miral_include}/miral/fd_manager.h
```
AlanGriffiths, line 56: `fd_manager.cpp` is not for use by client code, it should be built as part of `miral-internal` (which ensures that no entry points are leaked)

include/miral/miral/runner.h
```diff
@@ -90,6 +105,14 @@ class MirRunner
...
+
+    void unregister_fd_handler(void const* owner);
```
AlanGriffiths, line 115: Why are we declaring a private function in a public header? Users cannot call it and we don't. If we did use it, any functionality would reside on `Self` and should be invoked directly.

AlanGriffiths (changes requested):
Most importantly this needs tests!

As [noted in #2613](https://github.com/MirServer/mir/issues/2613#issuecomment-1235667260) I think a design that doesn't add more cruft to the `MirRunner` is to be preferred, but I won't block on that if we have a working solution.

Other than that, there are some issues noted concerning API & ABI hygiene. I'll look at the functionality after there are some tests to prove it works.

graysonguarino:
> Most importantly this needs tests!
Tests are coming today! :) The need for the ‘FdManager’ was to make the code testable.

AlanGriffiths:
> The need for the ‘FdManager’ was to make the code testable

I've not raised a problem with there being `FdManager`, just with it being part of the API & ABI: There a `miral-test-internal` executable for testing Miral internals!

src/miral/runner.cpp
```diff
@@ -134,8 +144,19 @@ try
         server->set_command_line(argc, argv);
         server->apply_settings();
         apply_env_hacks(*server);
-
+
```
wmww, line 147: Accidental whitespace

src/miral/runner.cpp
```diff
@@ -134,8 +144,19 @@ try
...
+        auto const main_loop = server->the_main_loop();
+
+        for (auto const& signal : signal_backlog)
+        {
+            main_loop->register_signal_handler(signal.signals, signal.handler);
+        }
+        signal_backlog.clear();
+
+        fd_manager->set_weak_main_loop(std::move(main_loop));
```
wmww, line 158: I'm not sure why this is working, it's generally not possible to move a const value (since moving implies changing the origin to null). Perhaps `main_loop` should not be const on line 150?

src/miral/symbols.map
```diff
@@ -466,5 +466,8 @@ MIRAL_3.7 {
 global:
   extern "C++" {
     "miral::MinimalWindowManager::MinimalWindowManager(miral::WindowManagerTools const&, MirInputEventModifier)";
+    miral::MirRunner::register_signal_handler*;
+    miral::MirRunner::register_fd_handler*;
+    miral::MirRunner::unregister_fd_handler*;
```
wmww, line 471: I believe `MirRunner::unregister_fd_handler` no longer exists

src/miral/fd_manager.cpp
```diff
@@ -0,0 +1,89 @@
...
+        main_loop->unregister_fd_handler(owner);
+    }
+    else
+    {
+        // Remove all entries with same owner from backlog
+        backlog.erase(remove_if(begin(backlog), end(backlog), [&](auto& info){ return info.owner == owner; }), end(backlog));
+    }
+}
+
+void FdManager::process_backlog()
```
wmww, line 63: Why not roll `process_backlog()` into `set_weak_main_loop()`, considering they can be thought of as a single operation? This way the function could take a `shared_ptr`, not have to check if it's null, use it within the function body and then store it as a weak pointer.

src/miral/fd_manager.cpp
```diff
@@ -0,0 +1,89 @@
...
+void FdManager::unregister_handler(void const* owner)
+{
+    if (auto const main_loop = weak_main_loop.lock())
+    {
+        main_loop->unregister_fd_handler(owner);
+    }
+    else
+    {
+        // Remove all entries with same owner from backlog
+        backlog.erase(remove_if(begin(backlog), end(backlog), [&](auto& info){ return info.owner == owner; }), end(backlog));
```
wmww, line 59: Nothing is protecting `unregister_handler()` from being called from one thread at the same time as `register_handler()` is being called from a different thread. I don't think your usage of weak_ptr above is threadsafe if the weak pointer may be changed in a different thread and you usage of vector here definitely is not.

src/miral/runner.cpp
```diff
@@ -184,6 +205,31 @@ void miral::MirRunner::add_stop_callback(std::function<void()> const& stop_callb
...
+    {
+        auto const signal_info = Self::SignalInfo{signals, handler};
+        self->signal_backlog.push_back(signal_info);
+    }
+}
+
+auto miral::MirRunner::register_fd_handler(mir::Fd fd, std::function<void(int)> const& handler)
+-> std::unique_ptr<miral::FdHandle>
+{
+    std::lock_guard lock{self->mutex};
```
wmww, line 229: Making calls to FdManager while holding the MirRunner mutex lock is a bad idea. This lock cannot fully protect FdManager (it won't be held when an FdHandle is dropped), and trying to lock multiple mutexes at the same time often leads to deadlocks. Since `fd_manager` ~~is~~ should be const, you don't need to hold a lock when calling it as long as it's threadsafe internally.

src/miral/runner.cpp
```diff
@@ -60,6 +61,15 @@ struct miral::MirRunner::Self
     std::function<void()> stop_callback{[this]{ join_client_threads(weak_server.lock().get()); }};
     std::function<void()> exception_handler{static_cast<void(*)()>(mir::report_exception)};
     std::weak_ptr<mir::Server> weak_server;
+    std::shared_ptr<miral::FdManager> fd_manager = std::make_shared<miral::FdManager>();
```
wmww, line 64: `fd_manager` should be const, no? Does it ever change?

wmww (changes requested):
Mostly looks good, but there's a bit of cleanup to do to make it threadsafe.

tests/miral/fd_manager.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    auto const fd = mir::Fd{42};
+
+    {
+        auto const handle = manager->register_handler(fd, [](int) { std::function<void(int)>(); });
+        EXPECT_CALL(*main_loop.get(), unregister_fd_handler(_));
+    }
+}
+
+TEST(MockMainLoop, dropping_fd_handle_before_main_loop_created_does_not_register_handler)
```
wmww, line 44: This test would be more useful if you set the main loop after the register and unregister happened. Or you can add another test that does this. As it is this doesn't test if the handler is removed from the backlog.

src/miral/fd_manager.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+void FdManager::set_weak_main_loop(std::shared_ptr<mir::MainLoop> main_loop)
```
wmww, line 67: I would now drop `weak` and call it `set_main_loop()`

src/miral/fd_manager.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
```
wmww, line 68: Need to lock the mutex

wmww (changes requested):

src/miral/fd_manager.h
```diff
@@ -0,0 +1,87 @@
...
+    FdHandle(FdHandle const&) = delete;
+    FdHandle& operator=(FdHandle const&) = delete;
+    FdHandle(FdHandle&&) = default;
+
+    std::shared_ptr<FdManager> manager;
+};
+
+/// A struct holding the necessary info to register a file descriptor if FdManager::register_handler()
+/// is called before the Server has started.
+struct FdInfo
```
wmww, line 79: Since `FdInfo` is only used inside of `FdManager, I would declare it in `FdManager`'s private section.

src/miral/fd_manager.h
```diff
@@ -0,0 +1,87 @@
...
+#include <mutex>
+#include <vector>
+#include <functional>
+
+namespace mir { class MainLoop; }
+
+namespace miral
+{
+struct FdHandle;
+struct FdInfo;
```
wmww, line 33: Which also means `FdInfo` does not need to be forward declared.

src/miral/fd_manager.h
```diff
@@ -0,0 +1,87 @@
...
+public:
+    ~FdHandle();
+
+private:
+    friend class FdManager;
+
+    FdHandle(std::shared_ptr<FdManager> manager);
+    FdHandle(FdHandle const&) = delete;
+    FdHandle& operator=(FdHandle const&) = delete;
+    FdHandle(FdHandle&&) = default;
```
wmww, line 72: Move constructors do not work how you expect them to. They do not "consume" the old object, they allow moving data out of it. The destructor is still called on both the moved from and moved to object eventually. Also moving an object changes it's address, so the usage of `this` in the destructor will have unpredicatable results.

As it currently is this isn't causing problems for unclear reasons, perhaps because the only time a handle is moved is when returning and that's covered by [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision)?

Since the move constructor here is private it cannot be used outside of `FdHandle` and it's friend `FdManager`, but this limits how it can be used. If you make it public and move handles in the tests things will be more obviously broken. I reccomend switching back to `std::unique_ptr`

wmww (changes requested):

src/miral/fd_manager.h
graysonguarino: Huh, it seems I accidentally pushed 4ceb1d8 prematurely. I was playing around with some things @AlanGriffiths and I were talking about but wasn’t ready to share. Will get a better look at this in the morning.

graysonguarino:
There is an issue with this implementation, but I am pushing for opinions on how to fix it.

Inside `MirRunner::Self::run_with()`, I have to store a `shared_ptr<mir::MainLoop>` to give to the `FdManager`. The theory is that `FdManager` should *really* hold a weak reference to `MainLoop`, but because of `MainLoop` being stored as a `CachedPtr`, I cannot guarantee the same address will exist for the loop after whatever scope calls `FdManager::set_main_loop()` goes out of scope.

But when I hold onto a strong reference to `MainLoop` to give to `FdManager` (in order to ensure it is still valid when the manager needs it), it seems that the server crashes because it expects that nothing holds a strong reference to `MainLoop` after `server->run()` has completed.

Does anyone have an idea as to how I can maintain the lifetime of a `MainLoop` without causing any unwanted side effects?

You can see that the [Ubuntu Frame Diagnostic delay](https://github.com/MirServer/ubuntu-frame/pull/89) works with this state of this PR, but `miral-test` segfaults immediately.

AlanGriffiths:
> Inside `MirRunner::Self::run_with()`, I have to store a `shared_ptr<mir::MainLoop>` to give to the `FdManager`. The theory is that `FdManager` should _really_ hold a weak reference to `MainLoop`, but because of `MainLoop` being stored as a `CachedPtr`, I cannot guarantee the same address will exist for the loop after whatever scope calls `FdManager::set_main_loop()` goes out of scope.

It sounds as though the problem is that the lifetime of `FdManager` is the problem. There is no point in this existing before (or after) the lifetime of `MainLoop` as it needs the `MainLoop` to be running to work.

If you ensure that it is created after `MainLoop` (e.g. through an "init" callback) and destroyed first (e.g. through a "stop" callback) then could be no problem tracking `MainLoop` - you just need a pointer to the object.

AlanGriffiths (changes requested):
Please provide some tests of the public functions that are introduced

graysonguarino:
It seems like there is some sort of race condition that is getting triggered by the new tests that I still haven't found yet. Pushing as it's nearing EOW and I'm not sure I'll find it within the hour.

include/test/miral/test_display_server.h
```diff
@@ -63,6 +63,16 @@ struct TestDisplayServer : private TestRuntimeEnvironment
     /// \note call before start_server()
     void add_stop_callback(std::function<void()> const& stop_callback);
 
+    /// Add signal handler to the server's main loop.
+    /// \remark Since MirAL 3.7
```
AlanGriffiths, line 67: You should also bump `MIRAL_VERSION_MINOR` in `CMakeLists.txt` - that will allow downstream code to test for the version

tests/miral/runner.cpp
```diff
@@ -35,6 +38,23 @@ struct Runner : miral::TestServer
...
+    }
+
+    void LockingCallback()
+    {
+        {
+            std::lock_guard<std::mutex> lock(mutex);
+            ready_to_tear_down = true;
+        }
+
+        cv.notify_all();
```
AlanGriffiths, line 55:
```suggestion
        cv.notify_one();
```

AlanGriffiths (changes requested):
Some tidy-up required

tests/miral/fd_manager.cpp
```diff
@@ -0,0 +1,104 @@
...
+using namespace testing;
+
+using namespace mir::test::doubles;
+
+TEST(FdManager, dropping_fd_handle_after_main_loop_created_unregisters_handler)
+{
+    auto const manager = std::make_shared<miral::FdManager>();
+    auto const main_loop = std::make_shared<NiceMock<MockMainLoop>>();
+
+    // MirRunner::run_with() triggers the following
```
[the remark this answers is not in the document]
graysonguarino: I put that comment in there to clarify that the tests have to be implemented differently than how the code actually runs due to `MirRunner` not being mocked here. 

Looking back at it, a user would never hold an `FdManager` outside of a `MirRunner` anyways, so the clarification doesn't seem necessary. I agree the comments should be removed.

tests/miral/fd_manager.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <gmock/gmock.h>
+
+using namespace testing;
+
+using namespace mir::test::doubles;
+
+TEST(FdManager, dropping_fd_handle_after_main_loop_created_unregisters_handler)
+{
+    auto const manager = std::make_shared<miral::FdManager>();
+    auto const main_loop = std::make_shared<NiceMock<MockMainLoop>>();
```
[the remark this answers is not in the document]
graysonguarino: I believe `FdManager` will need to remain a `shared_ptr` here because it inherits `std::enable_shared_from_this` so `FdHandleImpl` can keep the `FdManager` alive until all handles are destroyed.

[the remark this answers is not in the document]
AlanGriffiths: OK, although like most use of `enable_shared_from_this` that's ugly and should probably not be combined with an accessible constructor.

I still think a test fixture would unclutter the test code

tests/miral/runner.cpp
```diff
@@ -73,6 +93,73 @@ TEST_F(Runner, x11_socket_is_not_returned_by_default)
...
+    register_signal_handler({signum}, [this](int) { LockingCallback(); });
+
+    miral::TestServer::SetUp();
+
+    EXPECT_CALL(*this, callback())
+        .Times(AtLeast(1));
+    kill(getpid(), signum);
+
+    miral::TestServer::TearDown();
+}
```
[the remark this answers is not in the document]
graysonguarino: Upon reading this, I thought a separate `wait` outside of TearDown() would be unnecessary. As it turns out, it is. 

The `FdHandle` wasn't being held once it went out of scope after a test run (which then calls `TearDown()` automatically), so the Fd was being dropped and the callback usually wasn't triggered.

Calling `TearDown()` directly "fixed" the problem because it was keeping the handle in scope!

tests/miral/runner.cpp
```diff
@@ -35,6 +38,36 @@ struct Runner : miral::TestServer
...
+        std::unique_lock<std::mutex> lock(mutex);
+        cv.wait(lock, [this] { return ready_to_tear_down; });
+        lock.unlock();
+    }
+
+    std::mutex mutex;
+    std::condition_variable cv;
+    bool ready_to_tear_down = true;
+    mt::Pipe pipe;
+    char const data_to_write = 'a';
```
AlanGriffiths, line 70: Given the semantics here, this looks like a helper class:

```suggestion
class ScopeExitSemaphore
{
public:
   ScopeExitSemaphore() = default;

    void ~ScopeExitSemaphore()
    {
        semaphore.acquire();
    }
   
    void operator()()
    {
        semaphore.release();
    }
private:
    std::binary_semaphore semaphore{0};
};
```

Then the tests can be written:

```
TEST_F(Runner, register_signal_handler_before_setup_invokes_callback_after_setup)
{
    int const signum{SIGUSR1};

    ScopeExitSemaphore scope_exit_signal;
    register_signal_handler({signum}, [this, &](int){ callback(); scope_exit_signal(); });

    miral::TestServer::SetUp();

    EXPECT_CALL(*this, callback())
        .Times(AtLeast(1));

    kill(getpid(), signum);
}

**Review by AlanGriffiths — CHANGES_REQUESTED** [review]
One more iteration. Sorry!

File: tests/miral/runner.cpp [hunk 24]
```diff
@@ -73,6 +106,66 @@ TEST_F(Runner, x11_socket_is_not_returned_by_default)
...
+    register_signal_handler({signum}, [this](int){ locking_callback(); });
+
+    miral::TestServer::SetUp();
+
+    EXPECT_CALL(*this, callback())
+        .Times(AtLeast(1));
+
+    kill(getpid(), signum);
+    wait_to_tear_down();
+}
```
AlanGriffiths, line 123: Or... we could use a naked semaphore (a shame that latch isn't available for 20.04):

```suggestion
TEST_F(Runner, register_signal_handler_before_setup_invokes_callback_after_setup)
{
    std::binary_semaphore semaphore{0};
    int const signum{SIGUSR1};

    register_signal_handler({signum}, [this, &](int){ locking_callback(); semaphore.release(); });

    miral::TestServer::SetUp();

    EXPECT_CALL(*this, callback())
        .Times(AtLeast(1));

    kill(getpid(), signum);
    semaphore.try_acquire_until(/* a short wait*/); // Wait for callback before exiting scope
}
```

tests/miral/runner.cpp
AlanGriffiths: It is moot (because I'm happy with either approach) but neither approach allows "FdHandle to go out of scope and unregister the fd even if the semaphore was held".

tests/include/mir/test/doubles/mock_main_loop.h
```diff
@@ -55,6 +55,7 @@ class MockMainLoop : public MainLoop
     MOCK_METHOD1(unregister_fd_handler, void(void const*));
 
     MOCK_METHOD2(enqueue, void(void const*, ServerAction const&));
+    MOCK_METHOD(void, enqueue_with_guaranteed_execution, (ServerAction const&));
```
RAOF, line 58: Can you make this
`MOCK_METHOD(void, enqueue_with_guaranteed_execution, (ServerAction const&), (override));`?

RAOF (approved):
This looks good to me now.

I'm confident @AlanGriffiths will be happy with this, but I'll let him do the final borsing.

tests/miral/runner.cpp
```diff
@@ -80,15 +55,15 @@ TEST_F(Runner, stop_callback_is_called)
 
 TEST_F(Runner, start_callback_is_called)
 {
-    mt::Signal signal;
+    auto signal = std::make_shared<mt::Signal>();
```
AlanGriffiths, line 58: Why introduce a `shared_ptr`?

I do realise that in the failure case the lambda will have a stale reference to `signal`, and that _could_ lead to undefined behaviour (if the reference were actually used).

However, taking the `shared_ptr` by reference doesn't change that.

AlanGriffiths (approved):
> I'm confident @AlanGriffiths will be happy with this, but I'll let him do the final borsing.

I've fixed the things I was unhappy with. ;) But I'll leave someone else to do the final borsing as it includes my changes

bors delegate

Saviq:
Me!

bors r=AlanGriffiths,RAOF

Issue #2639, opened 2022-09-14
Shutdown crash when OSK is running


Saviq:
Steps (handle `WAYLAND_DISPLAY` and `XDG_RUNTIME_DIR` as appropriate):
1. run e.g. `miriway-shell --add-wayland-extensions all`
2. run `ubuntu-frame-osk`
3. interrupt `miriway-shell`

Expected:
1. clean shutdown

Current:
1. segfault in https://github.com/MirServer/mir/blob/e3ef13e0165d5125042163c3689756627bd97549/src/miral/wayland_extensions.cpp#L379

[shutdown_gdb.txt](https://github.com/MirServer/mir/files/9567187/shutdown_gdb.txt)

wmww:
I can't seem to repro with the steps given and I don't have a theory as to what's happening, but #2640 was an overall improvement to extension filtering. Can you give that a spin?

Saviq:
@wmww that does seem to help, yes.

Saviq:
Duplicate of #2209

AlanGriffiths:
I'm surprised this code would be active during shutdown. #2640 might mask the symptoms, but probably doesn't address the root cause

Saviq:
@AlanGriffiths could you reproduce? It's a 100% here if I:

```shell
env \
  -u LD_LIBRARY_PATH \
  MIR_SERVER_WAYLAND_HOST=$WAYLAND_DISPLAY \
  WAYLAND_DISPLAY=wayland-99 \
  miriway-shell --add-wayland-extension all \
& WAYLAND_DISPLAY=wayland-99 ubuntu-frame-osk \
& sleep 2; kill %1
```

AlanGriffiths:
Yes, with a simpler incantation:
```plain
miriway --shell-component ubuntu-frame-osk& sleep 2; kill %1
```

AlanGriffiths:
Annoyingly, as soon as I try with a local build it doesn't crash.

AlanGriffiths:
OK, so I've not traced it all through yet but...

During "teardown" destroying the `DisplayServer` is destroying the `mf::WaylandConnector` and that (indirectly) destroys `WlrScreenCopyManagerV1Global` which destroys `mw::Global::~Global()` destructor calls `wl_global_destroy()`, and that (indirectly) calls `wl_global_is_visible()` and that ends up calling our `mf::WaylandConnector::wl_display_global_filter_func()` which leads to the segfault location.

[...]

PR #2647, opened 2022-09-20
WaylandConnector: Fix use-after-free

RAOF:
The libwayland global filter is called during wl_global_remove (sensibly, as it needs to know whether or not the client knows about the global that's being removed).

This means that the filter will be called from the destructor of any extension object that publishes a global (ie: all of them).

Ensure it lives long enough to avoid UAF.


AlanGriffiths:
Does this fix #2639?

Saviq:
> Does this fix #2639?

It does, yes.

Fixes #2639

wmww (approved):
LGTM
bors r+

PR #2648, opened 2022-09-20
Implement X11 platform pixel format selection correctly

### LINKED ISSUE #2624: mgx::X11Window::r_mask used uninitialized
It is used, but it is never set. Spotted in Valgrind.

wmww:
Fixes #2624.


RAOF (approved):
I had to run it to check that we're returning the right pixel format (the interaction of ARGB and endianness *always* has be double-guessing) :)

bors r+

PR #2649, opened 2022-09-21
CMake: Only run package-abis test if toplevel debian dir exists

### LINKED ISSUE #2596: Tests always fail for non-Debian systems on package-abis due to missing debian/ dir
I'm working on packaging Mir for [Void Linux](https://voidlinux.org/) and starting with 2.9.0 tests **always** fail due to the absence of the `debian/` directory:
```
=> libmir-2.9.0_1: running do_check ...
[0/1] Running tests...
Test project /builddir/mir-2.9.0/build
      Start  1: LGPL-required
 1/81 Test  #1: LGPL-required ............................................................   Passed    0.01 sec
      Start  2: GPL-required
 2/81 Test  #2: GPL-required .............................................................   Passed    0.01 sec
      Start  3: package-abis
 3/81 Test  #3: package-abis .............................................................***Failed    0.00 sec
Failed to find debian/ directory. Make sure you run the script from the source tree root.

```

I propose a configure-time flag like `-DMIR_DEBIAN_ABIS_TEST_BROKEN=ON` or similar, or simply skipping/not running the `package-abis` test if `debian/` cannot be found as is the case in [the release tarball of v2.9.0](https://github.com/MirServer/mir/archive/refs/tags/v2.9.0.tar.gz).

To avoid something like this in the future I think tests should also be ran on at least one non-Debian-based job such as `alpine-edge` (would also cover testing on `musl`) which I can see [passed the 2.9.0 release commit "test and build" job](https://github.com/MirServer/mir/runs/8033782800) without actually even running the built tests [according to the full build log](https://github.com/MirServer/mir/commit/3811e4d26786b342ec784f8326b231c33d3c82be/checks/8033782800/logs).

I also have another test failure (test 71/80 `test_wayland_generator`) currently as well but I think I'll make a separate issue for that as well.

JamiKettunen:
Attempting to run this on non-Debian systems building from the release tarballs breaks running tests due to non-existent debian dir.

Fixes #2596.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -348,8 +348,10 @@ if (MIR_ENABLE_TESTS)
     COMMAND /bin/sh -c "! grep -rl --exclude-dir=protocol --exclude=org_kde* 'GNU Lesser' ${PROJECT_SOURCE_DIR}/src/server ${PROJECT_SOURCE_DIR}/include/server ${PROJECT_SOURCE_DIR}/src/include/server ${PROJECT_SOURCE_DIR}/tests ${PROJECT_SOURCE_DIR}/examples"
   )
 
-  mir_add_test(NAME package-abis
-    COMMAND /bin/sh -c "cd ${PROJECT_SOURCE_DIR} && tools/update_package_abis.sh --check --verbose")
+  if (IS_DIRECTORY "${PROJECT_SOURCE_DIR}/debian")
+    mir_add_test(NAME package-abis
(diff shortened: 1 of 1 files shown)


RAOF (approved):
We *might* be better off having an explicit “extra CI checks” target that's not run by the normal `test` target.

This is an improvement on the status quo, and we can easily move to an extra target later if we decide we want to.

bors r+

PR #2657, opened 2022-09-22
Deactivate text input in destructor for v1 and v2 protocols

### LINKED ISSUE #2651: OSK stays on screen with unfocused app
Steps:
1. run `miriway --shell-component ubuntu-frame-osk`
2. start [`telegram-desktop`](https://snapcraft.io/telegram-desktop)
3. focus a text field
4. unfocus the app

Expected:
1. OSK closes when the app is unfocused

Current:
1. OSK opens and stays on, ~~even when the app goes away (#2657)~~

There probably isn't enough knowledge in the compositor (OSK is like any other layer-shell surface) to hide it, so when the app misbehaves, we're at the mercy of other apps doing the right thing.

wmww:
~~Fixes #2651,~~ should have been included in #2270

### DIFF
--- a/src/server/frontend_wayland/text_input_v1.cpp
+++ b/src/server/frontend_wayland/text_input_v1.cpp
@@ -259,6 +259,7 @@ TextInputV1::~TextInputV1()
 
     on_new_input_field = false;
     pending_state.reset();
+    ctx->text_input_hub->deactivate_handler(handler);
 }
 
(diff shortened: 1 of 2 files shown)


Saviq (approved):
Sure, this helps when the session goes away. But it doesn't help when unfocusing and the app doesn't DTRT then.

bors merge

PR #2660, opened 2022-09-23
Server/run_mir: Fatal signal handling is the worst.

RAOF:
Our fatal signal handler is (necessarily!) signal-unsafe, so we can deadlock while trying to run it.

At least document this, and dump a message using a known-safe method before proceeding into the swamp.


wmww (approved):
LGTM
bors r+

Issue #2663, opened 2022-09-25
FTBFS: unused_result in run_mir.cpp


Saviq:
Introduced in #2660 

https://github.com/MirServer/mir/blob/c308112149638d177b7979e0802becd3a19c9b8e/src/server/run_mir.cpp#L161-L163

```
/<<PKGBUILDDIR>>/src/server/run_mir.cpp: In function ‘void {anonymous}::fatal_signal_cleanup(int, siginfo_t*, void*)’:
/<<PKGBUILDDIR>>/src/server/run_mir.cpp:163:32: error: ignoring return value of ‘ssize_t write(int, const void*, size_t)’ declared with attribute ‘warn_unused_result’ [-Werror=unused-result]
  163 |         static_cast<void>(write(STDERR_FILENO, warning, len));
      |                           ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Saviq:
Urm.

```
/<<PKGBUILDDIR>>/src/server/run_mir.cpp:165:20: error: ignoring return value of ‘ssize_t write(int, const void*, size_t)’ declared with attribute ‘warn_unused_result’ [-Werror=unused-result]
  165 |         (void)write(STDERR_FILENO, warning, len);
      |               ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

PR #2664, opened 2022-09-26
Server/run_mir: Fix build failure.

RAOF:
*Of course* there exist compilers for which `static_cast<void>(foo)` does not disable the “must-use return value” warning for `write()`…

Fixes: #2663


Saviq (approved):
Thanks!

bors merge

Issue #2659, opened 2022-09-23
mock_input_surface.h: Mock method mistakenly fails to override base method


RAOF:
```
In file included from /home/chris/Canonical/Mir/mir/main/tests/unit-tests/scene/test_surface.cpp:25:
/home/chris/Canonical/Mir/mir/main/tests/include/mir/test/doubles/mock_input_surface.h:39:18: warning: 'mir::test::doubles::MockInputSurface::consume' hides overloaded virtual function [-Woverloaded-virtual]
    MOCK_METHOD1(consume, void(MirEvent const*));
                 ^
/home/chris/Canonical/Mir/mir/main/src/include/server/mir/input/surface.h:52:18: note: hidden overloaded virtual function 'mir::input::Surface::consume' declared here: type mismatch at 1st parameter ('const std::shared_ptr<const MirEvent> &' (aka 'const shared_ptr<const MirEvent> &') vs 'typename ::testing::internal::Function< ::testing::internal::identity_t<void (const MirEvent *)>>::template Arg<0>::type' (aka 'const MirEvent *'))
    virtual void consume(std::shared_ptr<MirEvent const> const& event) = 0;
```

PR #2665, opened 2022-09-26
mtd::MockInputSurface: Fix `consume` method signature 

RAOF:
This was apparently missed in the recent Event refactor.

Fixes: #2659


Saviq (approved):
Yup.

bors merge

PR #2668, opened 2022-09-26
Support scaled screenshots

### LINKED ISSUE #2653: Scaling messes up screencopy
There's a bunch of transparency around screenshots when scaling is used:
frame_2022-09-22T17:39+02:00

wmww:
Fixes #2653, a few systems needed to be refactored. With this PR software buffers now expose their properties without being mapped, render targets expose their size so the renderer doesn't have to query the current EGL surface (something that never worked for screenshots), and our wlr-screencopy implementation requests clients use buffers of the full output resolution rather than the logical size.

### DIFF
--- a/src/platforms/common/server/buffer_from_wl_shm.cpp
+++ b/src/platforms/common/server/buffer_from_wl_shm.cpp
@@ -391,7 +391,7 @@ public:
 
             auto size() const -> mir::geometry::Size override
             {
-                return parent->ShmBuffer::size();
+                return parent->size();
             }
 
             auto data() -> T* override
@@ -463,7 +463,7 @@ public:
                 mir::geometry::Stride stride_;
             };
 
-            return std::make_unique<FallbackMapping>(pixel_format(), ShmBuffer::size(), stride_);
+            return std::make_unique<FallbackMapping>(pixel_format(), size(), stride_);
         }
     }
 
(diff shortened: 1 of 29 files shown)


src/platforms/common/server/buffer_from_wl_shm.cpp
```diff
@@ -390,7 +391,7 @@ class WlShmBuffer :
 
             auto size() const -> mir::geometry::Size override
             {
-                return parent->size();
+                return parent->ShmBuffer::size();
```
RAOF, line 394: Explicitly calling the `ShmBuffer` implementation of `size()` here is unnecessary, right? Both `parent->ShmBuffer::size()` and `parent->RWMappableBuffer::size()` call the same (necessarily explicitly implemented, because `RWMappableBuffer` is abstract) `WlShmBuffer::size()` implementation.

src/platforms/common/server/buffer_from_wl_shm.cpp
```diff
@@ -462,10 +463,14 @@ class WlShmBuffer :
                 mir::geometry::Stride stride_;
             };
 
-            return std::make_unique<FallbackMapping>(pixel_format(), size(), stride_);
+            return std::make_unique<FallbackMapping>(pixel_format(), ShmBuffer::size(), stride_);
```
RAOF, line 466: As above; `ShmBuffer::size()` is unnecessary?

RAOF (approved):
This looks good(ish) to me, with the caveat that I'd like to make annoying multiple inheritance as simple as possible by fully-qualifying the method in as few places as possible.

With those changes, good by me.

bors delegate=@wmww

RAOF (approved):
bors r+

PR #2672, opened 2022-09-27
Do not send zxdg_output.name after initialization

### LINKED ISSUE #2666: Bug in implementation of `xdg_output`
In the journal of my Miriway session, I see:
```
xdg-desktop-portal-kde[11860]: qt.qpa.wayland: zxdg_output_v1.name received after output has been initialized, this is most likely a bug in the compositor
```

We should work out if this is a Qt bug or a Mir bug.

wmww:
Fixes #2666

### DIFF
--- a/src/server/frontend_wayland/xdg_output_v1.cpp
+++ b/src/server/frontend_wayland/xdg_output_v1.cpp
@@ -62,11 +62,10 @@ public:
 
 private:
     auto output_config_changed(graphics::DisplayConfigurationOutput const& config) -> bool override;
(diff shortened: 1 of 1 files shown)


RAOF (changes requested):
Hm. I think this would be clearer by instead moving the `send_name_if_supported()` call into the constructor (so that it's only sent once by definition), and having the `cached_position` and `cached_size` members be `std::optional`, with empty being the no-value-yet state.

Since we're using >C++17, `position != cached_position` will be false when `cached_position` is an unoccupied optional.

RAOF (approved):
LGTM.

bors r+

PR #2673, opened 2022-09-27
Server/run_mir: Handle SIGQUIT as a normal termination

RAOF:
We don't *need* to handle SIGQUIT as an immediate crash, there doesn't seem to be any particular reason to, and handling it as an immediate crash goes through the unavoidably broken `fatal_signal_handler`.

### DIFF
--- a/src/server/run_mir.cpp
+++ b/src/server/run_mir.cpp
@@ -34,7 +34,7 @@
 
 namespace
 {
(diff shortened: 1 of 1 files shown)


Saviq (commented):
From what I read SIGQUIT is supposed to also dump core, and that doesn't seem to be happening?

RAOF:
Urgh, yeah. `SIGQUIT` is expected to dump core. That's a good reason for it to go through the `fatal_signal_handler`.

Issue #2678, opened 2022-09-28
Mir doesn't start if any DRM devices do not have an associated DRI driver


RAOF:
If a DRM device on the system has outputs but does *not* have a working DRI driver (and, so, any `gbm_*` calls will fail) then Mir will fail to start up, even if it could drive one of the other GPUs in the system.

An example from a system with an AMD GPU and an AST (basic 2D-only output hardware) on the motherboard follows. You can see that we (correctly) detect that AST is not an appropriate rendering device, as it's software rasterised, but because we detect an output connected to the AST hardware we try and bring up a display on it, and bail because we cannot use gbm or EGL.

```
2022-08-04T20:49:01Z ubuntu-frame.daemon[1893]: + exec /snap/ubuntu-frame/2845/usr/local/bin/frame
2022-08-04T20:49:01Z ubuntu-frame.daemon[1893]: [2022-08-04 20:49:01.152004] <information> mirserver: Starting
2022-08-04T20:49:01Z ubuntu-frame.daemon[1893]: [2022-08-04 20:49:01.156426] < - debug - > mirserver: Using Linux VT subsystem for session management
2022-08-04T20:49:01Z ubuntu-frame.daemon[1893]: [2022-08-04 20:49:01.156526] <information> mircommon: Loading modules from: /snap/ubuntu-frame/2845/usr/lib/x86_64-linux-gnu/mir/server-platform
2022-08-04T20:49:01Z ubuntu-frame.daemon[1893]: [2022-08-04 20:49:01.156548] <information> mircommon: Loading module: /snap/ubuntu-frame/2845/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-gbm-kms.so.20
2022-08-04T20:49:01Z ubuntu-frame.daemon[1893]: [2022-08-04 20:49:01.156555] <information> mircommon: Loading module: /snap/ubuntu-frame/2845/usr/lib/x86_64-linux-gnu/mir/server-platform/server-x11.so.20
2022-08-04T20:49:01Z ubuntu-frame.daemon[1893]: [2022-08-04 20:49:01.156560] <information> mircommon: Loading module: /snap/ubuntu-frame/2845/usr/lib/x86_64-linux-gnu/mir/server-platform/input-evdev.so.8
2022-08-04T20:49:01Z ubuntu-frame.daemon[1893]: [2022-08-04 20:49:01.156957] <information> mirserver: Found display driver: mir:gbm-kms (version 2.8.0)
[...]

PR #2679, opened 2022-09-28
gbm-kms/quirks: Quirk off AST devices

RAOF:
These are basic 2D output devices built into motherboards (particularly server MBs). They have no 3D hardware, and they have no DRI drivers. Unfortunately, bug #2678 means that this causes Mir to fail to come up on a system with AST hardware.

Like evdi, this hardware *should* be supportable in the New Platform API™, but for now quirking it off is the appropriate response.


Saviq (approved):
Yes.

bors merge

Issue #2643, opened 2022-09-15
Cursor only updates once per enter on XWayland


wmww:
When the cursor enters an X11 window it is correctly set. After that as it moves around the window the cursor is not updated until it leaves and re-enters. I suspect this is to do with #2617, but I need to investigate.

wmww:
Bisect reveals the issue was introduced in 7417656e3805b22d02dd298cf43bac7c0d8ee07d, part of #2621. Cause still unclear.

RAOF:
That commit is definitely wrong - it means that frame events are *only* sent when a buffer is used by the GL compositor, but buffers can be used in other ways (notably, as you've found, by the cursor subsystem).

Actually fixing that commit is a bit fiddly. I hope that I can just fix the underlying problem properly and this will fall out of it.

PR #2680, opened 2022-09-28
[shm buffers] notify consumed for all methods of consumption

AlanGriffiths:
(I know Chris wants to improve the code around here, but this fixes the specific problem quickly)

Fixes: #2643


RAOF (approved):
Yeah, that's good.

bors r+

PR #2681, opened 2022-09-29
XWayland support input hint

wmww:
Fixes #2255 by accommodating clients that manage their own input focus.

PR #2682, opened 2022-09-29
XWayland: refactor state management

wmww:
Refactors state management to be more flexible and readable. Adds a new XWayland-specific struct instead of relying fully on `scene::SurfaceStateTracker`. This refactor is stacked on top of #2681 and was originally written as part of that, but was found to be unnecessary. If in the future we want to implement `_NET_WM_STATE_FOCUSED` this refactor will help greatly.


src/server/frontend_xwayland/xwayland_surface.h
```diff
@@ -75,6 +91,52 @@ class XWaylandSurface
...
+        }
+
+        auto active_mir_state() const -> MirWindowState
+        {
+            return withdrawn ? mir_window_state_hidden : mir_state.active_state();
+        }
+        auto with_active_mir_state(MirWindowState state) const
+        {
+            return StateTracker{mir_state.with_active_state(state), false};
+        }
```
AlanGriffiths, line 113: I'm a bit confused by this design.

The idiom `auto with...(...) const -> NewObject` is usually associated with immutable types (and these have obvious benefits for thread safety), but `StateTracker` is not immutable: it supports copy and assignment.

So why not go one way or the other? Either:
1. Accept that `StateTracker` is mutable and have mutator methods; or,
2. Make `StateTracker` immutable and "own" the tracked instance through an owning pointer.

wmww: `StateTracker`, much like `ms::SurfaceStateTracker`, is mutable only by assignment, similar to primitive value types. Making it mutable would lead to it being non-obvious where changes to it were being made. It would also make the class less ergonomic to use. Using a `unique_ptr<StateTracker>` would cost a bunch of allocations and pointer lookups and I don't see any benefit that would yield (unique pointer assignment is not threadsafe without additional locking, so it would leave us in the same place there). A `const StateTracker` can be used if assignment is not desired.

AlanGriffiths (approved):
I'm not sure I like the design, but it works

bors merge

Issue #2674, opened 2022-09-27
Waking screens up after `idle-timeout` often stops or locks Mir up


Saviq:
I would estimate about 1/3 of wakes ends up either with going back to gdm (without an apparent crash) or deadlock (sometimes it's still possible to `chvt …` remotely, sometimes not even that.

I will collect what I can find here.

Most recent example had me completely locked up, I had to reboot. GDB showed: [gdb.txt](https://github.com/MirServer/mir/files/9653069/gdb.txt)

Saviq:
This sounds like #2632, but I've not seen evidence of actually crashing.

Saviq:
Likely related to #2661

AlanGriffiths:
I have yet to reproduce the problem (I've seen it a few times over the weeks, but nothing like 1/3).

I notice that you have a lot of threads inside iris_dri.so but my systems use i915. Maybe there's a behavioural difference with the newer hardware/driver that makes the problem a lot more noticeable.

Saviq:
> I notice that you have a lot of threads inside iris_dri.so but my systems use i915. Maybe there's a behavioural difference with the newer hardware/driver that makes the problem a lot more noticeable.

Could well be, I do have some other issues with this hw (primarily HDMI through USB-C being unstable, to the point of unusable [LP#1988989](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1988989)).

Issue #2661, opened 2022-09-23
Mir sometimes shuts down when moving to a different TTY when logind-driven


Saviq:
Steps
1. start a Miriway session
2. Ctrl+Alt+F* away from it

Expected
1. you move to the requested TTY

Current
1. sometimes Miriway quits

Logs say:
```
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338483] < - debug - > mirserver: Received logind force-pause event for device 13:70
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338622] < - debug - > mirserver: Received logind force-pause event for device 13:90
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338677] < - debug - > mirserver: Received logind force-pause event for device 13:88
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338729] < - debug - > mirserver: Received logind force-pause event for device 13:94
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338786] < - debug - > mirserver: Received logind force-pause event for device 13:69
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338831] < - debug - > mirserver: Received logind force-pause event for device 13:66
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338881] < - debug - > mirserver: Received logind force-pause event for device 226:0
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338928] < - debug - > mirserver: Received logind force-pause event for device 13:89
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.338970] < - debug - > mirserver: Received logind force-pause event for device 13:87
wrz 23 11:46:16 michal-laptop /usr/libexec/gdm-wayland-session[369987]: [2022-09-23 11:46:16.339054] < - debug - > mirserver: Received logind force-pause event for device 13:92
[...]

Issue #2632, opened 2022-09-13
Miriway frequently crashes to GDM after exiting powersave


RAOF:
Frequently when coming back from screen-blank Miriway will crash to GDM after displaying a single non-black frame of output.

I'll get a backtrace for this, just recording here to make sure I don't forget.

RAOF:
Hm. It looks like this might be a gbm-kms crash when my external monitor turns on & off / is unplugged? But I'll need to trigger it again to get the backtrace to with symbols.

Saviq:
I just had this happen despite running #2684.

[miriway.log tail](https://github.com/MirServer/mir/files/9721582/miriway.log)
```plain
paź 06 07:32:28 michal-laptop /usr/libexec/gdm-wayland-session[137660]: [2022-10-06 07:32:28.849412] < -warning- > gbm-kms: drmModeCrtcSetGamma failed: Function not implemented
paź 06 07:32:35 michal-laptop /usr/libexec/gdm-wayland-session[137660]: [2022-10-06 07:32:35.193984] <information> evdev-input: Removed /dev/input/event5
paź 06 07:32:39 michal-laptop /usr/libexec/gdm-wayland-session[137660]: [2022-10-06 07:32:39.205031] < -warning- > gbm-kms: drmModeCrtcSetGamma failed: Function not implemented
paź 06 07:32:39 michal-laptop /usr/libexec/gdm-wayland-session[137660]: !!! Fatal signal received. Attempting cleanup, but deadlock may occur
paź 06 07:32:39 michal-laptop /usr/libexec/gdm-wayland-session[137660]: Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGSEGV
paź 06 07:32:39 michal-laptop /usr/libexec/gdm-wayland-session[137660]: !!! Fatal signal received. Attempting cleanup, but deadlock may occur
paź 06 07:32:39 michal-laptop /usr/libexec/gdm-wayland-session[137660]: Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGABRT
paź 06 07:32:39 michal-laptop /usr/libexec/gdm-wayland-session[138395]: (EE) failed to read Wayland events: Broken pipe
paź 06 07:32:39 michal-laptop /usr/libexec/gdm-wayland-session[137633]: Aborted (core dumped)
```

It may well be a driver problem: [dmesg.log](https://github.com/MirServer/mir/files/9721617/dmesg.log)
```
[...]

PR #2684, opened 2022-09-30
Deal with confused state of KMS outputs better

AlanGriffiths:
Fixes: #2661
Fixes: #2674


Saviq:
Can confirm that this helps with VT-switching / displays going to sleep. With it, I didn't notice any of those problems. Possibly also fixes #2632?

PR #2685, opened 2022-10-04
Disable text input when client is unfocused

### LINKED ISSUE #2651: OSK stays on screen with unfocused app

wmww:
Fixes #2651

### DIFF
--- a/src/server/frontend_wayland/text_input_v1.cpp
+++ b/src/server/frontend_wayland/text_input_v1.cpp
@@ -313,6 +313,7 @@ void TextInputV1::focus_on(mf::WlSurface* surface)
     else
     {
         deactivate(nullptr);
+        ctx->text_input_hub->deactivate_handler(handler);
     }
 }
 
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
Huh? The Fedora 35 build failed with multiple, unrelated:
```plain
2022-10-04T00:58:01.8248432Z error: XDG_RUNTIME_DIR not set in the environment
2022-10-04T00:58:01.8248653Z Mir fatal error: Unable to bind Wayland socket
```
That looks like a change to the build environment. (Raised #2686)

AlanGriffiths (approved):
bors merge

PR #2689, opened 2022-10-07
Add some logging to terminate_with_current_exception()

AlanGriffiths:
We've encountered a few issues recently where we think we're trying to terminate, but don't have clear diagnostic information.

This ensures we log the "current exception" before trying to terminate.


Saviq (approved):
Works: #2695

bors merge

Issue #2690, opened 2022-10-07
`Cannot access memory` segfault in RealKMSOutput on shutdown


Saviq:
```gdb
Thread 95 (Thread 0x7f63d17fa6c0 (LWP 9883) "Mir/Workqueue"):
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007f657069ec46 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007f65706857fc in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007f6570ace803 in mir::fatal_error_abort(char const*, ...) (reason=0x7f65703d5128 "Unsupported attempt to continue after a fatal signal: %s") at /home/michal/dev/MirServer/mir/src/core/fatal.cpp:40
#6  0x00007f656ffd3f23 in (anonymous namespace)::fatal_signal_cleanup(int, siginfo_t*, void*) (sig=11, info=0x7f63d17f7b30, ucontext=0x7f63d17f7a00) at /home/michal/dev/MirServer/mir/src/server/run_mir.cpp:186
#7  0x00007f657069ecf0 in <signal handler called> () at /lib/x86_64-linux-gnu/libc.so.6
#8  0x00007f656dcd543c in mir::geometry::generic::Value<int, mir::geometry::WidthTag>::Value<unsigned short, true>(unsigned short const&) (this=0x7f63d17f87b8, value=<error reading variable: Cannot access memory at address 0x4>) at /home/michal/dev/MirServer/mir/include/core/mir/geometry/dimensions.h:79
#9  0x00007f656dcd4461 in mir::geometry::generic::Size<int>::Size<unsigned short const&, unsigned short const&>(unsigned short const&, unsigned short const&) (this=0x7f63d17f87b8, width=<error reading variable: Cannot access memory at address 0x4>, height=<error reading variable: Cannot access memory at address 0xe>) at /home/michal/dev/MirServer/mir/include/core/mir/geometry/size.h:52
[...]

PR #2691, opened 2022-10-07
Disconnected hardware has no modes: invent a size

AlanGriffiths:
In #2690, we're segfaulting trying to access the size of an output. Maybe this "fixes" it?

Marked as "draft" as it is experimental

### DIFF
--- a/src/platforms/gbm-kms/server/kms/real_kms_output.cpp
+++ b/src/platforms/gbm-kms/server/kms/real_kms_output.cpp
@@ -129,6 +129,10 @@ void mgg::RealKMSOutput::reset()
 
 geom::Size mgg::RealKMSOutput::size() const
 {
+    // Disconnected hardware has no modes: invent a size
+    if (connector->connection == DRM_MODE_DISCONNECTED)
+        return {0, 0};
+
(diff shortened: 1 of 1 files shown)


Saviq:
Could also help with #2695?

AlanGriffiths:
> Could also help with #2695?

No, that looks like yet another failure mode

RAOF (approved):
This is not *wrong*, but getting here is a result of breaking invariants.

It should not be possible to have a connector be both in state `DRM_MODE_DISCONNECTED` and used in the configuration. (The same applies to `max_refresh_rate`, below.)

That said, I can't see how we could uphold that invariant with the current code, and this *is* a strict improvement.

AlanGriffiths:
> It should not be possible to have a connector be both in state `DRM_MODE_DISCONNECTED` and used in the configuration. (The same applies to `max_refresh_rate`, below.)

I totally agree. But the fact this "works" proves this is what is happening (and that the current code is a flaky mess)

AlanGriffiths:
After discussion this improves the state of things, so marking "ready"

AlanGriffiths:
Fixes: #2690

RAOF (approved):
A cheap improvement on the status quo!

bors r+

PR #2692, opened 2022-10-07
Clean up copy/paste messages

wmww:
Before we were sending multiple null sections in a row, and may have been allowing clients to paste after they were unfocused. This PR cleans up that logic so that clients can only paste from the currently active offer and events are sent only as necessary.

### DIFF
--- a/src/server/frontend_wayland/wl_data_device.cpp
+++ b/src/server/frontend_wayland/wl_data_device.cpp
@@ -64,6 +64,7 @@ public:
     }
 
 private:
+    friend mf::WlDataDevice;
     wayland::Weak<WlDataDevice> const device;
     std::shared_ptr<scene::ClipboardSource> const source;
 };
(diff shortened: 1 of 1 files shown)


src/server/frontend_wayland/wl_data_device.cpp
```diff
@@ -78,12 +79,14 @@ mf::WlDataDevice::Offer::Offer(WlDataDevice* device, std::shared_ptr<scene::Clip
...
     }
-    device->send_selection_event(resource);
 }
 
 void mf::WlDataDevice::Offer::receive(std::string const& mime_type, mir::Fd fd)
 {
-    source->initiate_send(mime_type, fd);
+    if (device && device.value().current_offer.is(*this))
+    {
+        source->initiate_send(mime_type, fd);;
```
graysonguarino, line 88: Double semicolon here

graysonguarino (commented):
Besides the one thing, LGTM.

bors delegate

PR #2693, opened 2022-10-09
Fix compilation failure with GCC-12 (missing headers)

gladk:
Adding missing headers, needed for compilation with GCC-12.

### DIFF
--- a/src/client/rpc/mir_protobuf_rpc_channel.cpp
+++ b/src/client/rpc/mir_protobuf_rpc_channel.cpp
@@ -47,6 +47,7 @@
 
 #include <stdexcept>
 #include <cstring>
+#include <array>
 
 namespace mf = mir::frontend;
 namespace mev = mir::events;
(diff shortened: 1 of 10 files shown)


Saviq:
Hi @gladk, could you please give a bit more context as to how you found these missing includes? We're building with GCC 12 in CI and locally, and have not seen this.

gladk:
It is coming from [Debian Bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1012998). Also we could reproduce it in the [CI](https://salsa.debian.org/mir-server-team/mir/-/jobs/3341135).

This patch resolves the problem and the package successfully built on all supported platforms in Debia.

RAOF (approved):
@Saviq this is a merge into the 1.8 branch, which we *don't* routinely build in CI :)

Looks sensible; all of those files use `std::array<>`, so we should be explicitly including it.

bors r+

AlanGriffiths:
bors retry

Saviq:
> @Saviq this is a merge into the 1.8 branch, which we _don't_ routinely build in CI :)

D'oh!

> bors r+

> bors retry

And the above is why these won't work. 1.8 still has travis in it, we'd have to move it to the new world if we really needed to.

bors r-

PR #2696, opened 2022-10-10
ci: fix Fedora environment

### LINKED ISSUE #2686: [spread] Build failure on Fedora 35
We have a build failure on Fedora 35 with multiple test failing because:
```plain
2022-10-04T00:58:01.8248432Z error: XDG_RUNTIME_DIR not set in the environment
2022-10-04T00:58:01.8248653Z Mir fatal error: Unable to bind Wayland socket
```
That looks like a change to the build environment.

In the Alpine and sbuild tasks we set XDG_RUNTIME_DIR, maybe we should do so for Fedora too

https://github.com/MirServer/mir/actions/runs/3178554440/jobs/5180169817

Saviq:
Fixes #2686

### DIFF
--- a/spread/build/fedora/task.yaml
+++ b/spread/build/fedora/task.yaml
@@ -49,5 +49,6 @@ execute: |
     cmake $BUILD_DIR
     export VERBOSE=1
     cmake --build $BUILD_DIR -- -j`nproc`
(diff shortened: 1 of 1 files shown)


AlanGriffiths (approved):
bors merge

PR #2700, opened 2022-10-14
[miral] Respect the output id supplied when modifying a window

AlanGriffiths:


Saviq (approved):
Yes, please!

This allows me to do what I need, also confirmed OSK and swaybg behave as expected still.

bors merge

PR #2702, opened 2022-10-17
ThreadPoolExecutor: Drop std::binary_semaphore polyfill.

RAOF:
All targets we compile for should now support enough C++20 stdlib features to remove the need for us to provide a fallback.


RAOF:
Ha, ha! Joke's on me :rofl: 

This can't land until we drop Ubuntu 20.04 support.

PR #2705, opened 2022-10-17
Change mmap to MAP_PRIVATE in example clients

### LINKED ISSUE #2534: Change from MAP_SHARED to MAP_PRIVATE in our clients
In `wl_keyboard.keymap`, the protocol states:
> From version 7 onwards, the fd must be mapped with MAP_PRIVATE by the recipient, as MAP_SHARED may fail.

Our clients seem to to all use MAP_SHARED.

graysonguarino:
Fixes #2534

### DIFF
--- a/src/platforms/wayland/displayclient.cpp
+++ b/src/platforms/wayland/displayclient.cpp
@@ -659,7 +659,7 @@ void mgw::DisplayClient::keyboard_keymap(wl_keyboard* /*keyboard*/, uint32_t for
         BOOST_THROW_EXCEPTION(std::runtime_error("platform currently requires a keymap"));
     }
 
-    char* keymap_string = static_cast<decltype(keymap_string)>(mmap(NULL, size, PROT_READ, MAP_SHARED, fd, 0));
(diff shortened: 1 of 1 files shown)


examples/client/wayland_client.c
```diff
@@ -117,7 +117,7 @@ make_shm_pool(struct wl_shm* shm, int size, void **data)
 
     posix_fallocate(fd, 0, size);
 
-    *data = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+    *data = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE, fd, 0);
```
RAOF, line 120: This is the wrong `mmap` call to change. Here, we're creating a mapping for submitting a SHM buffer - mapping this as `MAP_PRIVATE` ensures that the server will not be able to read anything we've written to the buffer :grin: 

[This](https://github.com/MirServer/mir/blob/2a06ae488d95b509470331716d55e5d10c54b8f9/src/platforms/wayland/displayclient.cpp#L662) is the site where we `mmap` a keymap sent from the server.

RAOF (changes requested):

RAOF (approved):
Yup! That's it :)

bors r+

PR #2706, opened 2022-10-18
ci: bump actions/checkout version

Saviq:
Because Node.js 12 actions are getting deprecated:

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


AlanGriffiths (approved):
bors merge

Issue #2344, opened 2022-02-24
miral-shell hangs on shutdown if eglspinner internal client errors


RAOF:
I'm testing on my awkward desktop with both NVIDIA and AMD GPUs, and on 22.04 the default behaviour of client EGL is to use the NVIDIA GPU via `linux_dmabuf`. Unfortunately, the NVIDIA driver sends buffers with a (format,modifier) pair unknown to the AMD driver, causing the client to be killed with a Wayland protocol error.

When this happens to the internal eglspinner client, `miral-shell` hangs during shutdown waiting to `join()` the internal client thread.

RAOF:
(The EGL clients failing because they're trying to use the NVIDIA card is not a Mir problem; *all* Wayland compositors I've tested suffer from it)

AlanGriffiths:
I don't know how much you've dug into this, but is the problem that we are mishandling an error from EGL (when we should be terminating the client thread) or is EGL not reporting the error and hanging the thread?

PR #2718, opened 2022-10-25
miral/InternalClient: don't use a dead wl_display.

RAOF:
Once a Wayland error has occurred on a wl_display it can no longer be used. Check whether the display is in an error state before trying to do a roundtrip on internal client cleanup.

Fixes: #2344

### DIFF
--- a/src/miral/internal_client.cpp
+++ b/src/miral/internal_client.cpp
@@ -21,6 +21,7 @@
 #include <mir/server.h>
 #include <mir/scene/session.h>
 #include <mir/raii.h>
+#include <wayland-client-core.h>
 
 #define MIR_LOG_COMPONENT "miral::Internal Client"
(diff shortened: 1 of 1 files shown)


RAOF:
Note that I haven't *tested* that this fixes #2344, because it has turned out to be difficult to reproduce now, but I'm (a) confident that what we're currently doing is wrong, and (b) this would match the symptoms.

AlanGriffiths (approved):
This is clearly an improvement, but I suspect there are more places we should check?

bors merge

AlanGriffiths:
> We might have something more in the wayland platform, but as far as I can tell this is the only in-server site we use the _client_ `wl_display` API in such a potentially broken way.

Yeah, I was more thinking that `client_code(display)` should be throwing, not exiting

Issue #829, opened 2019-05-16
Mir doesn't use libwayland-server's `wl_*_interface` definition


RAOF:
The binding generator appears to provide Mir its own definition of `wl_surface_interface` (and, indeed, all the interface definitions), but this is hidden (as it's not explicitly exported in `symbols.map`).

I'm not sure the extent to which this is a problem, but *I've* written code in the past which checks for reference equality against a Wayland interface symbol; this would fail with current Mir.

wmww:
Indeed Mir provides it's own implementation of all the generated things, as we don't use `wayland-scanner` output at all.

> *I've* written code in the past which checks for reference equality

I'm not sure precisely what you're trying to accomplish. Are you checking if a `wl_resource` is of a specific type? If so, we have `wayland::WlType::is_instance(resource)` for that. If it's something else, I'm sure I can add it easily.

AlanGriffiths:
The only mentions of `wl_surface_interface` I see are in `mir/tests/miral/org_kde_kwin_server_decoration.c`. And these, AFAICS, reference `libwayland-server`'s definition.

Mir does provide `wl_surface_interface_data`, but that is exported.

@RAOF could you clarify?

RAOF:
> Indeed Mir provides it's own implementation of all the generated things, as we don't use `wayland-scanner` output at all.

Is there any particular reason to duplicate the existing definitions in `libwayland-server`? This doesn't require using the `wayland-scanner` output, just what's in the existing `libwayland-server.so`.

> 
> > _I've_ written code in the past which checks for reference equality
> 
> I'm not sure precisely what you're trying to accomplish. Are you checking if a `wl_resource` is of a specific type? If so, we have `wayland::WlType::is_instance(resource)` for that. If it's something else, I'm sure I can add it easily.

Yes. The (rather minor) problem is that there's existing Wayland server code that does reference-equality testing on the interface, which could be copy-and-pasted and unexpectedly break.

There's also the `wl_resource_instance_of()` call, which has a fast-path on reference-equality.

None of these are particularly serious issues, but I didn't think that there was a particularly good reason *to* define our own duplicates of these symbols.

wmww:
> This doesn't require using the wayland-scanner output, just what's in the existing libwayland-server.so

Is libwayland-server built with `wl_surface_interface` included? Is it built with all of the interfaces for wayland.xml types? I suppose we could special-case interfaces from standard Wayland types, then generate the ones from other protocols.

[...]

PR #2721, opened 2022-10-27
Wayland generator: Use core interfaces from libwayland-server

RAOF:
We don't need to provide definitions of the `wl_foo_interface` structs for types in core Wayland; they are provided by libwayland-server.

Fixes: #829


AlanGriffiths:
Maybe we do need to generate our own interface implementations?

On series with an older libwayland-server we get:
```
2022-10-27T05:14:55.6258802Z wl_global_create: implemented version for 'wl_seat' higher than interface version (8 > 7)
2022-10-27T05:14:55.6258997Z ERROR: ../../../src/wayland/global.cpp(29): Throw in function mir::wayland::Global::Global(wl_global*)
```

RAOF:
Hm, yes. I didn't think of old libwayland versions.

In that case, we should generate our own interfaces, and if we're going to do it on older libwayland versions there's not much point using libwayland even if it has the versions we need.

PR #2722, opened 2022-10-27
Move XKB modifier tracking from Wayland frontend to KeyMapper

### LINKED ISSUE #2145: Support virtual keyboard modifiers
As noticed by @AlanGriffiths in #2139, modifiers don't work on the virtual keyboard protocol.

To reproduce:
- Spin up a Squeekboard in Mir as described in #2139
- Focus on a terminal
- Enable `Ctrl` on the virtual keyboard
- Press `C`

Expected: same as Ctrl+C on a normal keyboard 

Actual: a `c` is typed into the terminal

### LINKED ISSUE #2720: Multi-keyboard modifier state tracking is slightly cursed
Repro:
- Have two physical keyboards connected
- Keyboard A press and hold Shift
- Keyboard B tab shift Shift
- Keyboard A press a letter
- Letter is incorrectly lower case
- With keyboard A Shift still held, press the same letter again 
- Letter is now correctly upper case

Note: if you repeat the steps but the pressed letter is on the other keyboard it is always correctly upper case.

wmww:
Instead of the long-standing hacky solution of maintaining an XKB state in the Wayland frontend, this PR uses the state already maintained in the KeyMapper. XKB modifiers are now passes as part of key events so they are available for the frontend to send to clients. With these changes in place it was possible to add support for virtual keyboard modifiers.

Fixes #2720
Fixes #2145

### DIFF
--- a/include/core/mir_toolkit/events/enums.h
+++ b/include/core/mir_toolkit/events/enums.h
@@ -87,6 +87,8 @@ typedef enum {
     /* System policy has triggered a key repeat on a key
        which was already down */
     mir_keyboard_action_repeat,
+    /* Modifiers have been updated without a key event. keysym, scan_code and text are not set. */
+    mir_keyboard_action_modifiers,
 
     mir_keyboard_actions
 } MirKeyboardAction;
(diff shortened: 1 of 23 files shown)


AlanGriffiths:
I like the idea, a shame the compiler doesn't

RAOF (approved):
Looks good. Apparently older libstdc++s lack `operator==(std::optional<T>, T)`, so you'll need to fix those comparisons.

Once that's done,
bors delegate=wmww

Issue #640, opened 2018-11-13
The cursor does not support attaching an icon (e.g. for Drag-n-Drop)


AlanGriffiths:
Wayland `wl_data_device.start_drag` "provides an icon to be moved around with the cursor", we should be able to use it.

PR #2734, opened 2022-11-30
Begin visual aspects of start_drag implementation 

graysonguarino:
This is only the beginning of the implementation of `start_drag`. Currently, the icon is dragged around but the pasting of the data the icon represents is not implemented. Touch dragging is also not implemented at this point in time.

Closes: #640


graysonguarino:
It seems like the protocol for `start_drag` never explicitly defines drag behavior over touch. This seems to mean that only cursor events should trigger drags.

src/server/frontend_wayland/wl_surface.h
```diff
@@ -101,6 +101,20 @@ class NullWlSurfaceRole : public WlSurfaceRole
...
+public:
+    DragWlSurfaceRole(WlSurface* surface);
+    ~DragWlSurfaceRole();
+    auto scene_surface() const -> std::optional<std::shared_ptr<scene::Surface>> override;
+    void commit(WlSurfaceState const& state) override;
+    void create_scene_surface();
+
+private:
+    wayland::Weak<WlSurface> const surface;
+    std::weak_ptr<scene::Surface> weak_scene_surface;
```
RAOF, line 115: It's not clear to me what value having this be a `weak_ptr` gives; it's added to the scene immediately, which holds a strong reference until something removes it from the scene, and the way *that* happens is either in `~DragWlSurfaceRole()` or on client destruction.

In any case, this seems like it'd be happer as a `shared_ptr`?

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -499,3 +500,77 @@ auto mf::NullWlSurfaceRole::scene_surface() const -> std::optional<std::shared_p
...
+mf::DragWlSurfaceRole::~DragWlSurfaceRole()
+{
+    if (surface)
+    {
+        surface.value().clear_role();
+
+        if (shared_scene_surface)
+        {
+            auto const& session = surface.value().session;
+            session->destroy_surface(shared_scene_surface);
```
RAOF, line 521: This is the wrong place to remove the scene surface; you should be implementing `surface_destroyed()` (which is called when the underlying `wl_surface` is destroyed) and removing the scene surface then.

This destructor is not called when the underlying `WlSurface` is destroyed.

RAOF (changes requested):
I think there's still a bunch of simplification that can be done here.

Also, this should have the TODOs (particularly: *actually* implementing the drag and drop bits, so we don't give the impression that D&D might work when it won't) implemented before we land this in a release.

It might be worth landing this as a “drag-and-drop” branch, then PRing future work into that branch, *then* merging the drag-and-drop branch into main. That should make it possible to do smallish, iterative reviews and but only land the feature when it works.

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -499,3 +500,77 @@ auto mf::NullWlSurfaceRole::scene_surface() const -> std::optional<std::shared_p
...
+auto mf::DragWlSurfaceRole::scene_surface() const -> std::optional<std::shared_ptr<scene::Surface>>
+{
+    if (shared_scene_surface)
+    {
+        return shared_scene_surface;
+    }
+    else
+    {
+        return std::nullopt;
+    }
```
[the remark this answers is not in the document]
graysonguarino: You're referring to `surface` and not `shared_scene_surface`, right? Because this seems right for the former but the latter isn't created until `DragWlSurfaceRole::create_scene_surface()`.

src/server/frontend_wayland/wl_surface.cpp
RAOF: Hm. On second thoughts, this is *one of* the places that the scene surface needs to be removed, so that the scene surface is removed when the drag ends.

The *other* one is in a `surface_destroyed()` implementation, so that the scene surface is removed if the underlying `wl_surface` is destroyed.

AlanGriffiths:
> 
2023-03-24 10:52:38.905 :: 2023-03-24 10:52:38.585 subprocess.CalledProcessError: Command '['snap', 'download', 'mir-libs', '--channel', 'latest/edge/pr2734/22/edge']' returned non-zero exit status 1.

That looks familiar (and I just rebased, so probably not an artefact of rework in progress)

AlanGriffiths:
Closing this as I think there's too much logic introduced into `frontend_wayland` that belongs in `scene`.

PR #2747, opened 2022-12-12
cmake: Fix ptest target

RAOF:
We've been accidentally misusing `tools/discover-gtests.sh` by trying to pass `--gtest_filter` through. Due to the way GTest parses `--gtest_filter` (specifically, that it only respects the *last* `--gtest_filter` on the command line, ignoring any others), this results in any filter we were *trying* to pass in being ignored.

Instead, add a `--gtest-includes` option to `tools/discover-gtests.sh`, error out if the command contains `--gtest_filter`, use this in test discovery infrastructure, and explicitly pass tests we don't expect to run as `--gtest-excludes`.

### DIFF
--- a/cmake/MirCommon.cmake
+++ b/cmake/MirCommon.cmake
@@ -70,7 +70,7 @@ function (mir_discover_tests_internal EXECUTABLE TEST_ENV_OPTIONS DETECT_FD_LEAK
 
   if(ENABLE_MEMCHECK_OPTION)
     set(test_cmd ${VALGRIND_CMD} ${test_cmd_no_memcheck})
-    set(test_no_memcheck_filter "*DeathTest.*:ClientLatency.*")
+    set(test_no_memcheck_filter "*DeathTest.*" "ClientLatency.*")
   endif()
 
   if(cmake_build_type_lower MATCHES "threadsanitizer")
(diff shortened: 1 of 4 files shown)


cmake/MirCommon.cmake
```diff
@@ -70,7 +70,7 @@ function (mir_discover_tests_internal EXECUTABLE TEST_ENV_OPTIONS DETECT_FD_LEAK
 
   if(ENABLE_MEMCHECK_OPTION)
     set(test_cmd ${VALGRIND_CMD} ${test_cmd_no_memcheck})
-    set(test_no_memcheck_filter "*DeathTest.*:ClientLatency.*")
+    set(test_no_memcheck_filter "*DeathTest.*" "ClientLatency.*")
```
Saviq, line 73: Slightly more explicit maybe?

```suggestion
    list(APPEND test_no_memcheck_filter "*DeathTest.*" "ClientLatency.*")
```

cmake/MirCommon.cmake
```diff
@@ -126,13 +126,34 @@ function (mir_discover_tests_internal EXECUTABLE TEST_ENV_OPTIONS DETECT_FD_LEAK
...
+    file(APPEND ${CMAKE_BINARY_DIR}/discover_all_tests.sh
+      "sh ${CMAKE_SOURCE_DIR}/tools/discover_gtests.sh ${discover_env} --test-name ${test_name}_no_memcheck --gtest-exclude ${test_exclusion_str} --gtest-include ${memcheck_exclusion_str} -- ${discover_cmd_no_memcheck}\n")
+  else()
+    file(APPEND ${CMAKE_BINARY_DIR}/discover_all_tests.sh
+      "sh ${CMAKE_SOURCE_DIR}/tools/discover_gtests.sh ${discover_env} --test-name ${test_name} --gtest-exclude ${test_exclusion_str} -- ${discover_cmd}\n")
+  endif()
+
+  # Final commands
+  set(test_cmd "${test_cmd}" "--gtest_filter=-${memcheck_exclusion_str}:${test_exclusion_str}")
+  set(test_cmd_no_memcheck "${test_cmd_no_memcheck}" "--gtest_death_test_style=threadsafe" "--gtest_filter=${memcheck_exclusion_str}:-${test_exclusion_str}")
```
Saviq, line 155: Since you key by `test_no_memcheck_filter` above, could move those up there to avoid memcheck exclusions if they're empty? Not a strong voice, mind you, since if they're empty it'll just mean an extra leading `:`, but…

tools/discover_gtests.sh
```diff
@@ -78,7 +91,14 @@ then
     testname=$(basename $test_binary)
 fi
 
-tests=$($test_binary --gtest_list_tests --gtest_filter=-$excludes | grep -v '^ ' | cut -d' ' -f1 | grep '\.' | sed 's/$/*/')
+if [ -z "$includes" ];
+then
+    discover_filter="-$excludes"
+else
+    discover_filter="$includes:-$excludes"
+fi
```
Saviq, line 99: If you decide against "avoiding" colons where I commented above, then this would be in the same vein?

```suggestion
discover_filter="$includes:-$excludes"
```

Saviq (approved):
Tiny nitpicks, feel free to ignore.

bors delegate

tools/discover_gtests.sh
AlanGriffiths: Can't we have the best of both?

```suggestion
    discover_filter="${includes:+$includes:}-$excludes"
```

Saviq: We can!

Saviq:
bors merge

Saviq:
bors retry

Saviq:
One last time :P

bors retry

AlanGriffiths (approved):
Doesn't look any uglier than what went before

bors merge

Saviq:
Gah. Thanks.

bors t=Saviq,AlanGriffiths

Saviq:
bors r=Saviq,AlanGriffiths

Issue #2733, opened 2022-11-25
Mir fatal error: surface observer should have been deleted, but was not (use count 2) when running Audacity in egmde


Saviq:
Steps:
1. install the audacity snap
2. run egmde from edge
3. run audacity in it

Expected:
- things work

Current:
- every couple launches Mir crashes:
  ```plain
  [2022-11-25 10:52:28.203988] < -warning- > mirserver: Window title Audacity uruchamia si... is not valid UTF-8
  [2022-11-25 10:52:29.925243] < - ERROR - > xwayland: XCB error: no reply data
  Mir fatal error: surface observer should have been deleted, but was not (use count 2)
  !!! Fatal signal received. Attempting cleanup, but deadlock may occur
  Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGABRT
  (EE) failed to read Wayland events: Connection reset by peer
  [1]    2055644 IOT instruction (core dumped)  egmde
  ```

Saviq:
NB: Audacity may continue running after Mir crashes, which may be an issue in its own right (but I can't see Xwayland around, when it does).

PR #2749, opened 2022-12-12
Fix race

AlanGriffiths:
mf::XWaylandSurface::close() and attach_wl_surface() are not sufficiently sychronised when managing observer and surface. Fix that by adding a manager class to own the association.

Fixes: #2733

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -444,21 +444,15 @@ void mf::XWaylandSurface::close()
 {
     std::shared_ptr<XWaylandClientManager::Session> local_client_session;
     std::shared_ptr<scene::Surface> scene_surface;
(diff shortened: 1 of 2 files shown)


wmww (changes requested):
The reason we fatal error if we detect the observer is still alive is because as long as the observer is alive and attached to a `ms::Surface` it may receive observations and call into it's `XWaylandSurface`. This could cause unexpected behavior or a segfault if it happens after `close()` completes.

Previously we created the observer, set `XWaylandSurface::surface_observer` and then created a surface that used that observer. With the new changes we don't set `XWaylandSurface::surface_observer` until after we've created a surface that uses the observer, meaning that it opens us up to getting events after `close()`.

I will look into alternative ways of fixing this, and possibly documenting it better.

wmww (approved):
LGTM
bors r+

AlanGriffiths:
bors retry

Saviq:
bors retry

PR #2750, opened 2022-12-13
platform/DRMFormat: Add as_mir_format converter

RAOF:
Ideally we would migrate to a world where `DRMFormat` is the canonical type for pixel formats; there are more of them, and the descriptor contains more useful information that code using `MirPixelFormat` has had to ad-hoc.

Make it possible to start doing this by adding a conversion from `DRMFormat` to `MirPixelFormat`. This is necessarily lossy, as there are a lot of `DRMFormat`s that have no corresponding `MirPixelFormat`, but that implies that *current* code couldn't have properly handled them, so it's OK.


AlanGriffiths (approved):
I guess we want the dependency first...

Saviq:
This can now go in.

bors r=AlanGriffiths

PR #2752, opened 2022-12-13
GitHub: Try to upload core files on failure

RAOF:
Sometimes, tests fail with a crash. If we could get these tests to dump a core we could inspect later, that would be nice!

### DIFF
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -114,6 +114,8 @@ jobs:
         # Dump cores in a known directory, for later upload
         mkdir coredumps
         echo "$PWD/coredumps/corefile-%e-%p-%t" | sudo tee /proc/sys/kernel/core_pattern
+        # schroot runs suid, so we need to relax core dump security policies
+        sudo sysctl fs.suid_dumpable=2
 
         snap run spread-mir-ci.spread -v ${{ matrix.spread-task }}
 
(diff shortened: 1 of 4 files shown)


RAOF:
CI is being *pretty* consistent at failing that armhf test, so hopefully this will get tested :)

RAOF:
Oh goddamnit!

Saviq:
Is the core useful at all without being retraced in situ? We may need to artifact the binaries produced?

Saviq:
So this _should_ give us dumps from native binaries… but I can't see the target arch / qemu dumps anywhere when driven from sbuild… If I run things by hand, I get `qemu_*.core` files in `$PWD`… but I couldn't see them anywhere in the chroot :/.

RAOF:
Oh! I wonder if apport is running in the LXD container, and resetting `core_pattern`?

Saviq:
> Oh! I wonder if apport is running in the LXD container, and resetting `core_pattern`?

Testing locally the dumps ended up in `core_pattern`, but _inside_ the container + chroot, which is why I went the mount way.

The qemu — the more interesting ARM ones? — I only ever saw in `$PWD`, which I suppose makes sense since it's QEMU dumping those, and not the kernel?

AlanGriffiths:
This has been inactive too long. At a minimum it needs rebasing on main, and the discussion probably needs to be resolved too

PR #2756, opened 2022-12-16
Spike dynamic display configuration

AlanGriffiths:
This isn't necessarily something we want to land in this form.

But it automatically applies changes to the `.display` config file. (At least for the modified miral-kiosk))

I won't be working on it over the holidays, but shared as a draft in case there's interest


AlanGriffiths:
Superseded by #2798

Issue #2742, opened 2022-12-07
[clang] we should rework unqualified call to std::move and std::forward


AlanGriffiths:
Newer clang is warning for such calls as `foo(move(some_shared_ptr))` and suggests `foo(std::move(some_shared_ptr))`.

Having researched the motivation for this warning, we should rework the code

AlanGriffiths:
Hmm, does this also affect `*pointer_cast<>`?

AlanGriffiths:
OK, so having consulted some experts, I'll quote Jonathan Wakely:

> For std::unique_ptr<foo::bar> the associated namespaces are std and foo, so it really is possible that ADL could find some other std::move.
> 
> I do think it's good practice to always, *always* qualify std::move and std::forward, to be very clear you're using the right ones.

So, despite our habit of avoiding names that clash with namespace std, I tend to agree that we should keep the warning and rework the code.

PR #2764, opened 2023-01-09
Fixup unqualified `std::move` warnings

RAOF:
Clang warns about calling `std::move` as an unqualified call to `move`, and is reasonable to do so because Argument Dependent Lookup *could* cause the compiler to resolve `bar::T baz; move(baz)` to some other `bar::move(baz)` if there happens to be a `bar::move(T)` implementation in the same namespace as `T` is defined.

Closes: #2742


Saviq (approved):
Yas.

bors merge

PR #2765, opened 2023-01-09
spread/sbuild: Don't purge the schroot session, more for local runs

RAOF:
This is both a (tiny) CI optimisation - we're running sbuild inside an ephemeral LXD container, so there's no point in spending time cleaning up the schroot - and also makes for easier debugging, as you can run `spread -debug` to stop after sbuild fails and still have the build available in the preserved schroot session.

### DIFF
--- a/spread/build/sbuild/task.yaml
+++ b/spread/build/sbuild/task.yaml
@@ -49,6 +49,7 @@ execute: |
       --dist=${RELEASE}
       --build=${DEB_BUILD_ARCH}
       --host=${DEB_HOST_ARCH}
+      --purge=never
     )
 
     SCHROOT_NAME="${RELEASE}-${DEB_BUILD_ARCH}"
--- a/spread/build/sbuild/task.yaml
+++ b/spread/build/sbuild/task.yaml
@@ -68,7 +68,7 @@ execute: |
       MKSBUILD_OPTS+=("--arch=${DEB_BUILD_ARCH}" "--target=${DEB_HOST_ARCH}")
 
       # Quirk for the eglexternalplatform-dev build dependency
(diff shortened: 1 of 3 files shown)


spread/build/sbuild/task.yaml
```diff
@@ -49,6 +49,7 @@ execute: |
       --dist=${RELEASE}
       --build=${DEB_BUILD_ARCH}
       --host=${DEB_HOST_ARCH}
+      --purge=never
```
Saviq, line 52: In my recent experience this still removes all the packages from the chroot, doesn't it? Do you know a way to avoid that as well?

RAOF: Hm. I thought that would suppress that as well.

`--purge-build-deps` is probably the option needed, but (a) I thought that `--purge` implied that, and at least *some* of the `--purge-*` options are mutually exclusive 🤷‍♀️

Saviq: Maybe I used `--purge-session=never`…

Saviq (changes requested):
I would apply a handful more improvements, then: #2766

On my setup I also needed:
```sh
sed -i 's/aufs/none/' /etc/schroot/chroot.d/sbuild-*
```
Any idea how to avoid that?

RAOF:
Hm. I'm not sure why your mk-sbuild is setting up aufs mounts (it should have been using overlayfs for *many* years), nor why you need to fiddle with them; I don't need to do that locally.

Saviq:
bors merge

Saviq:
bors retry

AlanGriffiths:
bors retry

PR #2770, opened 2023-01-09
examples/miral-kiosk/fake-mir-kiosk.sh: Contains bashisms, thus use /…

sunweaver:
…bin/bash in shebang.

 This is the quick and easy solution. Better approach might be
 de-bash'ing the fake-mir-kiosk.sh shell script.


AlanGriffiths (approved):
I think we're better dropping the script - it has been superseded by the frame-it snap (and Ubuntu Frame)

bors delegate

AlanGriffiths:
We're better just dropping the script: #2787

PR #2776, opened 2023-01-10
debian: use newer -dev packages

Saviq:
Those were highlighted by Debian's linter.


RAOF (approved):
These are all sensible changes.

bors r+

PR #2779, opened 2023-01-12
Some test code clean up  (before I make changes)

AlanGriffiths:
There's some legacy style that can be cleaned up


Saviq (approved):
ACK.

bors merge

Issue #2790, opened 2023-01-16
[Xwayland] "surface observer should have been deleted, but was not"


AlanGriffiths:
Repeatedly running `mir_performance_tests` on Miriway eventually crashes Miriway:

```
#5  0x00007fabbfb437d5 in mir::fatal_error_abort(char const*, ...)
    (reason=0x7fabbf0df590 "surface observer should have been deleted, but was not (use count %d)")
    at /home/alan/display_server/mir1/src/core/fatal.cpp:40
#6  0x00007fabbefcee40 in mir::frontend::XWaylandSurface::XWaylandSurfaceObserverManager::~XWaylandSurfaceObserverManager() (this=0x7fab1c7f8dc0, __in_chrg=<optimised out>)
    at /home/alan/display_server/mir1/src/server/frontend_xwayland/xwayland_surface.cpp:1423
#7  0x00007fabbefc91ca in mir::frontend::XWaylandSurface::close() (this=0x7fab48002310)
    at /home/alan/display_server/mir1/src/server/frontend_xwayland/xwayland_surface.cpp:478
#8  0x00007fabbefa733a in mir::frontend::XWaylandWM::handle_destroy_notify(xcb_destroy_notify_event_t*)
    (this=0x7fab34001740, event=0x7fab40001d70)
    at /home/alan/display_server/mir1/src/server/frontend_xwayland/xwayland_wm.cpp:742
#9  0x00007fabbefa64f9 in mir::frontend::XWaylandWM::handle_event(xcb_generic_event_t*)
    (this=0x7fab34001740, event=0x7fab40001d70)
    at /home/alan/display_server/mir1/src/server/frontend_xwayland/xwayland_wm.cpp:578
#10 0x00007fabbefa4d9a in mir::frontend::XWaylandWM::handle_events() (this=0x7fab34001740)
    at /home/alan/display_server/mir1/src/server/frontend_xwayland/xwayland_wm.cpp:285
#11 0x00007fabbef84e84 in operator()() const (__closure=0x7fab34001168)
    at /home/alan/display_server/mir1/src/server/frontend_xwayland/xwayland_connector.cpp:174
#12 0x00007fabbef873fc in std::__invoke_impl<void, mir::frontend::XWaylandConnector::spawn()::<lambda()>&>(std::__invoke_other, struct {...} &) (__f=...) at /usr/include/c++/12/bits/invoke.h:61
[...]

PR #2791, opened 2023-01-17
Do not abort when nothing is wrong

AlanGriffiths:
We need to ensure that no further callbacks occur after `~XWaylandSurfaceObserverManager()` executes, and this is guaranteed by `ObserverRegistrar::unregister_interest()`:

     * It is guaranteed that once unregister_interest() returns that no
     * other thread is in, or will receive, observations from this
     * ObserverRegistrar. If the thread calling unregister_interest()
     * is currently in an Observer call that call will complete, but
     * no further ones will be dispatched.

What is not guaranteed though is that there will be no remaining `shared_ptr` to `observer`.

In practice, that is not so as `WeakObserver::spawn()` may still hold `live_observer` after `WeakObserver::invoke()` has returned.

As the check is both redundant and wrong remove it.

Fixes: #2790

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -1412,19 +1412,6 @@ mf::XWaylandSurface::XWaylandSurfaceObserverManager::~XWaylandSurfaceObserverMan
     {
         scene_surface->unregister_interest(*surface_observer);
     }
-
-    if (surface_observer)
-    {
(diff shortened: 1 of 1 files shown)


wmww (approved):
Fair enough, with `XWaylandSurfaceObserverManager` hopefully the logic is simple enough it wont get screwed up again.
bors r+

AlanGriffiths:
bors retry

PR #2795, opened 2023-01-19
Document that extension filter callbacks are called multiple times

wmww:
Instead of #2640


RAOF (approved):
I approve.

I've added a note about the performance implications of the callback; please also give that a review.

AlanGriffiths (approved):
bors merge

PR #2798, opened 2023-01-20
Automate reloading display configuration

AlanGriffiths:
The display configuration can be supplied as a yaml file in two ways:

1. `miral::display_configuration_options()` provides a `--display-config=static=<filename>`; and,
2. `miral::DisplayConfiguration` searches for `<program>.display`

These options are used by Miriway and Frame respectively.

This PR sets up inotify monitoring for changes to the file and attempts to reload and apply the configuration.


AlanGriffiths:
Alpine has a stupid `basename()`!
```
2023-01-20T10:05:43.3879602Z /spread/mir/src/miral/static_display_config.cpp: In constructor 'miral::StaticDisplayConfig::StaticDisplayConfig(const std::string&)':
2023-01-20T10:05:43.3880102Z /spread/mir/src/miral/static_display_config.cpp:113:61: error: invalid conversion from 'const char*' to 'char*' [-fpermissive]
2023-01-20T10:05:43.3880352Z   113 |     ReloadingYamlFileDisplayConfig(::basename(filename.c_str()))
2023-01-20T10:05:43.3880495Z       |                                               ~~~~~~~~~~~~~~^~
2023-01-20T10:05:43.3880635Z       |                                                             |
2023-01-20T10:05:43.3880787Z       |                                                             const char*
2023-01-20T10:05:43.3880998Z In file included from /spread/mir/src/miral/static_display_config.cpp:37:
2023-01-20T10:05:43.3881328Z /usr/include/libgen.h:9:16: note:   initializing argument 1 of 'char* basename(char*)'
2023-01-20T10:05:43.3881455Z     9 | char *basename(char *);
```

AlanGriffiths:
I've found there are cases where the display isn't doesn't update until something causes recompositing. Will update soon

RAOF (commented):
I shall finish reviewing this, but just as a first pass:
* it looks broadly sane
* should we be using `std::filesystem::path`?

src/miral/static_display_config.cpp
```diff
@@ -459,3 +475,133 @@ auto miral::StaticDisplayConfig::list_layouts() const -> std::vector<std::string
...
+        auto const filename = config_path_.value() + "/" + basename;
+
+        if (access(filename.c_str(), F_OK))
+        {
+            std::ofstream out{filename};
+            print_template_config(out);
+
+            mir::log_debug(
+                "%s display configuration template: %s",
+                out ? "Written" : "Failed writing",
```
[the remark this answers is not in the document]
AlanGriffiths: > * should we be using `std::filesystem::path`?

Not until we move to core22

src/miral/static_display_config.cpp
```diff
@@ -459,3 +475,133 @@ auto miral::StaticDisplayConfig::list_layouts() const -> std::vector<std::string
...
+        BOOST_THROW_EXCEPTION((std::system_error{errno, std::system_category(), "Failed to initialize inotify_fd"}));
+
+    config_path_wd = inotify_add_watch(inotify_fd, config_path_.value().c_str(), IN_CLOSE_WRITE | IN_CREATE | IN_DELETE);
+
+    return inotify_fd;
+}
+
+void miral::ReloadingYamlFileDisplayConfig::auto_reload()
+{
+    if (auto const icf = inotify_config_path())
```
Saviq, line 563: Should we not unregister any existing handlers here? I know in the current use it won't happen, but if anyone would call `init_auto_reload()` more than once, we'd end up handling notify's multiple times?

Saviq (changes requested):
Overall looking good!

It doesn't notice if you move the config file in from another location / name — it will, if you then modify it.

There's a bit of a confusing behaviour with `miral::DisplayConfiguration` in that it will only look at the different roots once, and not load the configuration from a location higher in the lookup list if it appeared there. Something to be documented, at least - but we may need to deal with it in Frame when ran as a user.

We may to be a bit more verbose, log which file we're reading in / monitoring for changes?

src/miral/display_configuration.cpp
```diff
@@ -20,77 +20,59 @@
...
         {
-            mir::log_debug("Nowhere to write display configuration template: Neither XDG_CONFIG_HOME or HOME is set");
+            (config_roots = home) += "/.config:";
+            config_path(home + "/.config"s);
         }
+
+        if (auto config_dirs = getenv("XDG_CONFIG_DIRS"))
+            config_roots += config_dirs;
         else
+            config_roots += "/etc/xdg";
```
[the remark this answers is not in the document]
AlanGriffiths: > There's a bit of a confusing behaviour with `miral::DisplayConfiguration` in that it will only look at the different roots once, and not load the configuration from a location higher in the lookup list if it appeared there. Something to be documented, at least - but we may need to deal with it in Frame when ran as a user.

Good suggestion. At present we just watch the file we found.

Watching locations where we didn't find a file is possible, but then do we also want to watch for those locations appearing? E.g. if `~/.config' doesn't exist do we first monitor `~` for the creation of `.config` and then monitor that?

AlanGriffiths:
> It doesn't notice if you move the config file in from another location / name — it will, if you then modify it.

So, add `IN_MOVED_TO`?

src/miral/static_display_config.cpp
AlanGriffiths: Maybe just abort for a logic error?

src/miral/display_configuration.cpp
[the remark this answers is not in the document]
Saviq: > Watching locations where we didn't find a file is possible,

I'm not _sure_ we want it… IMO documenting that behaviour would be enough.

> do we also want to watch for those locations appearing

But if you think worth it - I don't think we need to go that far… If a root doesn't exist, log/warn about it and skip it.

src/miral/static_display_config.cpp
Saviq: Sure.

Saviq (commented):
> > It doesn't notice if you move the config file in from another location / name — it will, if you then modify it.
> 
> So, add `IN_MOVED_TO`?

Sounds like it, yeah.

include/miral/miral/display_configuration.h
```diff
@@ -33,6 +33,8 @@ class ConfigurationOption;
 /// the XDG Base Directory Specification. Vis:
 ///($XDG_CONFIG_HOME or $HOME/.config followed by $XDG_CONFIG_DIRS)
 /// \remark Since MirAL 2.4
+/// \note From MirAL 3.8 will monitor the configuration file or, if none found,
+/// for the creation of a file in $XDG_CONFIG_HOME or $HOME/.config
```
Saviq, line 37: Something went wrong here?

AlanGriffiths: WDYM?

Saviq: Hmm. Just needed to read it multiple times to parse properly. Sorry for the noise.

AlanGriffiths: Maybe I need to write simpler sentences

Saviq (approved):
All good!

bors merge

Saviq:
bors retry

PR #2802, opened 2023-01-20
Add missing #include statements

musicinmybrain:
- `<stdexcept>` for `std::logic_error`
- `<cstdint>` for `uint*_t`

Fixes failure to compile with GCC 13.


wmww (approved):
LGTM
bors r+

Issue #2774, opened 2023-01-10
A couple ShmBacking.* tests failing on PowerPC


Saviq:
Seen e.g. [in PPA](https://launchpad.net/~mir-team/+archive/ubuntu/dev/+build/25467048):

```
[ RUN ] ShmBacking.access_fault_is_true_after_invaild_read                                                                                                                                                  
/root/mir-2.10.0+dev147-g0593df6a8f/tests/unit-tests/test_shm_backing.cpp:310: Failure      
Value of: map->access_fault()                                                                                                                                                                               
Actual: false                                                                                         
Expected: true                                                                                                                                                                                              
[ FAILED ] ShmBacking.access_fault_is_true_after_invaild_read (1 ms)
[ RUN ] ShmBacking.access_fault_is_true_after_invaild_write                                                                                                                                                 
/root/mir-2.10.0+dev147-g0593df6a8f/tests/unit-tests/test_shm_backing.cpp:328: Failure
Value of: map->access_fault()                                                                                                                                                                               
Actual: false                                                                                         
Expected: true                                                                                                                                                                                              
[ FAILED ] ShmBacking.access_fault_is_true_after_invaild_write (0 ms)
```

These are new from #2744.

Saviq:
https://github.com/MirServer/wlcs/issues/262

Saviq:
I thought this could've been caused by `-O3` (see MirServer/wlcs#263), but setting `-O2` unfortunately didn't help.

PR #2806, opened 2023-01-23
tests/ShmBacking: Improve invalid access tests.

RAOF:
So, `mmap` will provide a mapping *of pages*, which means its going to be a multiple of `PAGE_SIZE`, even if the backing file is not a-multiple-of-`PAGE_SIZE` big. It zero-pads the end up to the next page, *and accesses to this zero-initialised region do not fault*.

Now, it turns out I picked `4000` as the backing size, which is close enough to the `PAGE_SIZE` of `4096` on *most* kernels Ubuntu builds that doubling it puts you in a separate page, so these tests pass on all systems with a 4K `PAGE_SIZE`.

ppc64el has a 64K `PAGE_SIZE` 😠

Fix this by querying `sysconf(_SC_PAGE_SIZE)` and making our backing file *exactly* one page in size. Combined with `mmap` being specified to map on a page-aligned address (which is really the only thing it could do), that guarantees that *any* access outside the valid range is going to hit a different page, and hence fault.

Fixes: #2774

### DIFF
--- a/tests/unit-tests/test_shm_backing.cpp
+++ b/tests/unit-tests/test_shm_backing.cpp
@@ -274,8 +274,16 @@ TEST(ShmBacking, read_from_invalid_memory_returns_0)
 {
     using namespace testing;
 
-    constexpr size_t const shm_size = 4000;
-    constexpr size_t const claimed_size = shm_size * 2;    // Lie about our backing size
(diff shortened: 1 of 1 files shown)


Saviq (approved):
Looks legit, nice digging!

bors merge

Saviq:
Let's try again in hope that lxd is back in business.

bors retry

Issue #2807, opened 2023-01-23
Updates to the display configuration need to schedule compositing for orientation and scale changes


AlanGriffiths:
We check if we to reallocate display buffers and "ignore difference in orientation, scale factor, form factor, subpixel arrangement" and that is sensible.

Our current logic restarts the compositor to reallocate buffers which implicitly schedules compositing. And that works.

However, when we don't reallocate buffers and orientation and/or scale factor changes, we should also schedule compositing. (But we don't)

PR #2808, opened 2023-01-23
Recomposite when display configuration changes need it

AlanGriffiths:
When we don't reallocate buffers but orientation and/or scale factor changes, we should schedule compositing

Fixes: #2807

### DIFF
--- a/src/server/scene/mediating_display_changer.cpp
+++ b/src/server/scene/mediating_display_changer.cpp
@@ -389,6 +389,50 @@ bool configuration_has_new_outputs_enabled(
         });
     return has_new_output;
 }
+
+bool configuration_changes_require_recompositing(
+        mg::DisplayConfiguration const& existing,
+        mg::DisplayConfiguration const& updated)
+{
+    struct O_S
+    {
+        MirOrientation orientation;
+        float scale;
+#ifdef __cpp_impl_three_way_comparison
+        bool operator<=>(O_S const&) const = default;
+#else
+        bool operator!=(O_S const& that) const
+        {
+            return std::tie(orientation, scale) != std::tie(that.orientation, that.scale);
+        }
+#endif
+    };
+
+    std::unordered_map<mg::DisplayConfigurationOutputId, O_S> configs;
+
+    existing.for_each_output([&configs](auto const& output)
+    {
+        if (output.used)
+        {
+            configs.emplace(output.id, O_S{output.orientation, output.scale});
+        }
+    });
+
+    bool result = false;
+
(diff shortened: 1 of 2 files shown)


graysonguarino (approved):
LGTM, but this is a bit out of my wheelhouse. I'll delegate, but if anyone else wants to merge before Alan, I think that would be okay.

bors delegate

src/server/scene/mediating_display_changer.cpp
```diff
@@ -405,6 +449,13 @@ void ms::MediatingDisplayChanger::apply_config(
                 [this] { compositor->start(); }};
             display->configure(*conf);
         }
+        else if (configuration_changes_require_recompositing(*existing_configuration, *conf))
+        {
+            // Workaround for #2807: we could do better if Compositor had a `schedule_compositing(damage)` method
```
RAOF, line 454: Is there any particular reason *not* to introduce `Compositor::schedule_compositing` here? Tearing down and rebuilding the compositor is reasonably heavy, and if adding a method to `Compositor` isn't hard maybe we should do that instead?

AlanGriffiths: That would be an optimisation and, obviously, I considered it. But that opens a whole new rabbit hole that needs implementing across platforms. How about landing this and I'll raise an issue?

Saviq: :+1:, this will only run every so often, so shouldn't be a real performance issue?

AlanGriffiths:
bors merge

AlanGriffiths:
Shouldn't that have worked?

bors r=graysonguarino

Saviq:
@AlanGriffiths something's weird:

AlanGriffiths:
That's even more weird. I merged main (with an empty diff) and pushed.

Then I see that I closed it! But I can't reopen

Saviq:
> Then I see that I closed it! But I can't reopen

B/C you pushed to `main`, so there's nothing to commit? Or did bors?

https://github.com/MirServer/mir/commits/main

AlanGriffiths:
OK, the commits are on main with my name on. So I screwed up

Saviq:
> OK, the commits are on main with my name on. So I screwed up

Not necessarily, it's ([the merge](https://github.com/MirServer/mir/commit/6876d248511e380129e8b7e643df0e0db2c871d4)) what bors pushed to `staging`, just _someone_ pushed that to `main` - it may well have been bors, just got confused. And GitHub, too - since it didn't close-merged this one.

:shrug:

PR #2809, opened 2023-01-23
Add wl_output version 4

### LINKED ISSUE #2792: [23.04] swaybg requires a newer version of wl-output than we support
On 23.04 swaybg doesn't start:
```
$ swaybg
wl_registry@2: error 0: invalid version for global wl_output (21): have 3, wanted 4
```

graysonguarino:
Fixes #2792 by bumping wl_output to version 4. ~~Also adds in done events which the protocol specifies to send after name, description, and geometry events.~~

To test:
```
miral-shell --add-wayland-extensions all
sway-bg -i [path-to-wallpaper-image]
```

### DIFF
--- a/include/platform/mir/graphics/display_configuration.h
+++ b/include/platform/mir/graphics/display_configuration.h
@@ -189,6 +189,7 @@ struct UserDisplayConfigurationOutput
     MirOutputGammaSupported const& gamma_supported;
     std::vector<uint8_t const> const& edid;
     mir::optional_value<geometry::Size>& custom_logical_size;
+    std::string const name = "OUT-" + std::to_string(id.as_value());
 
     UserDisplayConfigurationOutput(DisplayConfigurationOutput& main);
     geometry::Rectangle extents() const;
--- a/include/platform/mir/graphics/display_configuration.h
+++ b/include/platform/mir/graphics/display_configuration.h
@@ -96,6 +96,8 @@ struct DisplayConfigurationOutput
 {
     /** The output's id. */
     DisplayConfigurationOutputId id;
+    /** The output's displayable name. */
+    std::string name;
     /** The id of the card the output is connected to. */
     DisplayConfigurationCardId card_id;
     /** The id of the logical output group (aka display wall) this output belongs to, or 0 for none. */
--- a/include/platform/mir/graphics/display_configuration.h
+++ b/include/platform/mir/graphics/display_configuration.h
@@ -189,7 +189,7 @@ struct UserDisplayConfigurationOutput
     MirOutputGammaSupported const& gamma_supported;
     std::vector<uint8_t const> const& edid;
     mir::optional_value<geometry::Size>& custom_logical_size;
-    std::string const name = "OUT-" + std::to_string(id.as_value());
+    std::string const& name;
 
     UserDisplayConfigurationOutput(DisplayConfigurationOutput& main);
     geometry::Rectangle extents() const;
--- a/include/platform/mir/graphics/display_configuration.h
+++ b/include/platform/mir/graphics/display_configuration.h
@@ -189,7 +189,7 @@ struct UserDisplayConfigurationOutput
     MirOutputGammaSupported const& gamma_supported;
     std::vector<uint8_t const> const& edid;
     mir::optional_value<geometry::Size>& custom_logical_size;
-    std::string const& name = "OUT-" + std::to_string(id.as_value());
+    std::string const name = "OUT-" + std::to_string(id.as_value());
 
     UserDisplayConfigurationOutput(DisplayConfigurationOutput& main);
     geometry::Rectangle extents() const;
--- a/include/platform/mir/graphics/display_configuration.h
+++ b/include/platform/mir/graphics/display_configuration.h
@@ -96,8 +96,6 @@ struct DisplayConfigurationOutput
 {
     /** The output's id. */
     DisplayConfigurationOutputId id;
-    /** The output's displayable name. */
-    std::string name;
     /** The id of the card the output is connected to. */
     DisplayConfigurationCardId card_id;
     /** The id of the logical output group (aka display wall) this output belongs to, or 0 for none. */
@@ -142,6 +140,10 @@ struct DisplayConfigurationOutput
     std::vector<uint8_t> edid;
 
     mir::optional_value<geometry::Size> custom_logical_size;
+
+    /** The output's displayable name. */
+    // TODO: Better output names that are consistant between sessions
+    std::string name = "OUT-" + std::to_string(id.as_value());
 
     /** The logical rectangle occupied by the output, based on its position,
         current mode and orientation (rotation) */
--- a/include/platform/mir/graphics/display_configuration.h
+++ b/include/platform/mir/graphics/display_configuration.h
@@ -189,7 +189,7 @@ struct UserDisplayConfigurationOutput
     MirOutputGammaSupported const& gamma_supported;
     std::vector<uint8_t const> const& edid;
     mir::optional_value<geometry::Size>& custom_logical_size;
-    std::string const name = "OUT-" + std::to_string(id.as_value());
+    std::string const& name = "OUT-" + std::to_string(id.as_value());
 
     UserDisplayConfigurationOutput(DisplayConfigurationOutput& main);
     geometry::Rectangle extents() const;
--- a/src/server/frontend_wayland/output_manager.cpp
+++ b/src/server/frontend_wayland/output_manager.cpp
@@ -106,8 +106,6 @@ auto mf::OutputInstance::output_config_changed(mg::DisplayConfigurationOutput co
         "Fake model",
         mw::Output::Transform::normal);
 
-    send_done();
-
     for (size_t i = 0; i < config.modes.size(); ++i)
     {
         auto const& mode = config.modes[i];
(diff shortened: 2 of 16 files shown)


graysonguarino:
I believe this is failing to build because of a CI issue. `build/bin/update-wayland-generator-test.sh` needs to be run in order for `build/bin/test-wayland-generator.sh` to pass. This is run in build target `wayland_generator_test_generated_files`.

Saviq:
@graysonguarino you need to check in the result of `${CMAKE_BINARY_DIR}/bin/update_wayland_generator_test.sh` - that's what the test compares the output from the generator to so it verifies the generator is DTRT.

graysonguarino:
> @graysonguarino you need to check in the result of `${CMAKE_BINARY_DIR}/bin/update_wayland_generator_test.sh` - that's what the test compares the output from the generator to so it verifies the generator is DTRT.

I could be misunderstanding what you're saying, but after running `update_wayland_generator_test.sh` manually on my machine, `test_wayland_generator.sh` passes.

I think CI is not building the target which runs `update_wayland_generator_test.sh`.

Saviq:
> I could be misunderstanding what you're saying, but after running `update_wayland_generator_test.sh` manually on my machine, `test_wayland_generator.sh` passes.

Yes, and then you need to _check in_ the result. See `git status` after you ran it :)

graysonguarino:
> > I could be misunderstanding what you're saying, but after running `update_wayland_generator_test.sh` manually on my machine, `test_wayland_generator.sh` passes.
> 
> Yes, and then you need to _check in_ the result. See `git status` after you ran it :)
> 
> Check in, as in commit the differences, having reviewed they're correct.

Oops! I thought this was supposed to generate in CI like we do with the Wayland generated code.

Saviq:
> Oops! I thought this was supposed to generate in CI like we do with the Wayland generated code.

This bit is actually testing that what we generate in CI will be like what we expect it to be.

src/server/frontend_wayland/xdg_output_v1.cpp
```diff
@@ -112,8 +112,11 @@ mf::XdgOutputV1::XdgOutputV1(
     // TODO: Better output names that are consistant between sessions
     auto output_name = "OUT-" + std::to_string(output_global.current_config().id.as_value());
     send_name_event_if_supported(output_name);
+    send_done_event();
```
RAOF, line 115: This is incorrect - `xdg_output_v1` is not the same interface as `wl_output`, and [as per the protocol](https://wayland.app/protocols/xdg-output-unstable-v1#zxdg_output_v1:event:done) the `done` event on `xdg_output_v1` is deprecated and should not be sent on ≥v3 objects.

src/server/frontend_wayland/output_manager.cpp
```diff
@@ -105,6 +105,9 @@ auto mf::OutputInstance::output_config_changed(mg::DisplayConfigurationOutput co
         "Fake manufacturer",
         "Fake model",
         mw::Output::Transform::normal);
+
+    send_done();
+
```
RAOF, line 110: This is the wrong spot to send the `done` event; it should be sent after *all* the events relevant to the output change have been sent.

The documentation in output_manager.h for `OutputConfigListener` says that `output_config_changed` should return `true` in if a `done` event should be sent, so this will result in *two* done events being sent.

RAOF (changes requested):
This isn't actually sending `wl_output.name()` at all (it's *not great* that `xdg_output_v1.name()` exists and is a totally different thing, but :woman_shrugging: )

Since `wl_output.name()` is documented to be sent exactly once at bind time, you should send this in the `OutputInstance` constructor.

(There's maybe a little refactoring that could be done there to make things clearer - specifically, the `OutputInstance` constructor could be responsible for sending the initial output configuration, rather than `OutputGlobal::bind` doing that manually)

tests/mir_test_doubles/stub_display_configuration.cpp
```diff
@@ -75,6 +76,7 @@ mtd::StubDisplayConfigurationOutput::StubDisplayConfigurationOutput(
     MirSubpixelArrangement subpixel_arrangement) :
         DisplayConfigurationOutput{
             id,
+            "OUT-" + std::to_string(id.as_value()),
```
AlanGriffiths, line 79: This code snippet appears in too many places.

Maybe an answer is to move the `name` member to the end of `DisplayConfigurationOutput` and provide an in-class default member initializer?

AlanGriffiths (changes requested):
Not yet looked at the protocol changes, but there's some code tidy-up

include/platform/mir/graphics/display_configuration.h
```diff
@@ -185,6 +189,7 @@ struct UserDisplayConfigurationOutput
     MirOutputGammaSupported const& gamma_supported;
     std::vector<uint8_t const> const& edid;
     mir::optional_value<geometry::Size>& custom_logical_size;
+    std::string const name = "OUT-" + std::to_string(id.as_value());
```
AlanGriffiths, line 192: This should be a const reference to the underlying `DisplayConfigurationOutput` (like the other members)

tests/mir_test_doubles/stub_display_configuration.cpp
```diff
@@ -209,8 +209,9 @@ mtd::StubDisplayConfig::StubDisplayConfig(unsigned int num_displays, std::vector
 
         geometry::Size physical_size{};
         geometry::Point top_left{};
+        int const id = i + 1;
         graphics::DisplayConfigurationOutput output{
-            graphics::DisplayConfigurationOutputId{static_cast<int>(i + 1)},
+            graphics::DisplayConfigurationOutputId{id},
```
AlanGriffiths, line 214: Unnecessary code churn

tests/miral/test_window_manager_tools.cpp
```diff
@@ -336,8 +336,9 @@ auto mt::TestWindowManagerTools::create_fake_display_configuration(
     for (auto i = 0u; i < outputs.size(); i++)
     {
         auto const& rect = outputs[i].second;
+        int const id = i + 1;
         config_outputs.push_back(mir::graphics::DisplayConfigurationOutput{
-            mir::graphics::DisplayConfigurationOutputId{(int)(i + 1)}, // id
+            mir::graphics::DisplayConfigurationOutputId{id}, // id
```
AlanGriffiths, line 341: Unnecessary code churn

tests/acceptance-tests/wayland-generator/protocol.xml
```diff
@@ -2414,7 +2414,7 @@
     </event>
   </interface>
 
-  <interface name="wl_output" version="3">
+  <interface name="wl_output" version="4">
```
AlanGriffiths, line 2417: It isn't strictly necessary to update the acceptance test input and reference output. But :shrug:

AlanGriffiths (changes requested):
Mostly OK, just a few tweaks needed

include/platform/mir/graphics/display_configuration.h
graysonguarino: I think it may be necessary to not have this be a const reference since it is generating the string upon construction. If this is a reference to something temporary, it cannot compile.

AlanGriffiths: It shouldn't be a reference to a temporary. It should be a reference to `DisplayConfigurationOutput::name`

include/platform/mir/graphics/display_configuration.h
```diff
@@ -141,6 +141,10 @@ struct DisplayConfigurationOutput
 
     mir::optional_value<geometry::Size> custom_logical_size;
 
+    /** The output's displayable name. */
+    // TODO: Better output names that are consistant between sessions
+    std::string name = "OUT-" + std::to_string(id.as_value());
```
AlanGriffiths, line 146: This may not be the best approach to naming. The spec gives examples like:

> Examples of names include 'HDMI-A-1', 'WL-1', 'X11-1'

And, elsewhere, we use similar names:
```yaml
layouts:
# keys here are layout labels (used for atomically switching between them)
# when enabling displays, surfaces should be matched in reverse recency order
  default:                         # the default layout
    cards:
    # a list of cards (currently matched by card-id)
    - card-id: 0
      eDP-1:
        # This output supports the following modes: 2560x1440@60.0
        #
        # Uncomment the following to enforce the selected configuration.
        # Or amend as desired.
        #
        # state: enabled	# {enabled, disabled}, defaults to enabled
        # mode: 2560x1440@60.0	# Defaults to preferred mode
        # position: [0, 0]	# Defaults to [0, 0]
        # orientation: normal	# {normal, left, right, inverted}, defaults to normal
        # scale: 1
        # group: 0	# Outputs with the same non-zero value are treated as a single display
      DisplayPort-1:
        # (disconnected)
      HDMI-A-1:
        # (disconnected)
      DisplayPort-2:
        # (disconnected)
      HDMI-A-2:
        # This output supports the following modes: 1024x600@60.0, 1920x1080@60.0,
        # 1600x900@60.0, 1366x768@59.9, 1280x720@60.0, 1280x720@59.9
        #
        # Uncomment the following to enforce the selected configuration.
        # Or amend as desired.
        #
        # state: enabled	# {enabled, disabled}, defaults to enabled
        # mode: 1024x600@60.0	# Defaults to preferred mode
        # position: [0, 0]	# Defaults to [0, 0]
        # orientation: normal	# {normal, left, right, inverted}, defaults to normal
        # scale: 1
        # group: 0	# Outputs with the same non-zero value are treated as a single display
```

graysonguarino: Agreed. I've opened #2815 to address this.

AlanGriffiths (approved):
LGTM

bors merge

PR #2810, opened 2023-01-24
platform/Display: Remove register_pause_resume_handlers

RAOF:
The `ConsoleServices` is a singleton, and which the `DisplayPlatform` only needs
to interact with during construction (in order to `acquire_device`).

There's no need to force the `Display` to keep a reference to the `ConsoleServices`
purely so that it can forward `register_pause_resume_handlers` to it.

### DIFF
--- a/include/platform/mir/graphics/display.h
+++ b/include/platform/mir/graphics/display.h
@@ -140,17 +140,6 @@ public:
         DisplayConfigurationChangeHandler const& conf_change_handler) = 0;
 
     /**
-     * Registers handlers for pausing and resuming the display subsystem.
-     *
-     * The implementation should use the functionality provided by the EventHandlerRegister
-     * to register the handlers in a way appropriate for the platform.
-     */
-    virtual void register_pause_resume_handlers(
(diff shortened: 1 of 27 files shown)


AlanGriffiths (approved):
WFM but this does need a mirserver soname bump as ServerConfiguration vtab has changed

bors merge

Saviq:
bors retry

Saviq:
This one needs fixing:

```plain
[ FAILED ] DisplayServerMainLoopEvents.postpones_configuration_when_paused (87 ms)
```

RAOF:
I still can't see why the initial code changes resulted in the arm64 tests becoming flaky, but this has now looped 10,000 times locally without a failure, so is probably now not significantly flaky.

AlanGriffiths (approved):
Re the tests changes: I now don't see why the tests ever worked as the functor runs "after constructing the server, but before invoking DisplayServer::run()", but the tests rely on the server running to process events.

We could probably add better support for such tests, but this works.

bors merge

PR #2812, opened 2023-01-24
Enable bypass by default

AlanGriffiths:
It has been over two years since #1801

Maybe this works now. (If not, maybe we should do some quirk magic)


AlanGriffiths:
After discussion, this isn't good in this form

Issue #2800, opened 2023-01-20
Building docs fails when trying to not remove doc/html/cppguide: Is a directory


OPNA2608:
On Mir 2.12.0 / since https://github.com/MirServer/mir/pull/2728, the `doc` target runs
```sh
rm -f  ${CMAKE_BINARY_DIR}/doc/html/{*,search}
```
in an effort to not delete the just-generated `cppguide` directory. Unfortunately this seems to error out on my machine because `rm -f` only handles missing paths gracefully, not directories (or i.e. lack of permissions):
![image](https://user-images.githubusercontent.com/23431373/213681257-8b5adabf-8a3b-4c35-a00f-87ba8cc9c0f4.png)

So this stops the build:
```
[  0%] Generating ../doc/html/cppguide/styleguide.css
[100%] Generating ../doc/html/cppguide/index.html
[100%] Built target guides
rm: cannot remove '/build/source/build/doc/html/cppguide': Is a directory
make[3]: *** [CMakeFiles/doc.dir/build.make:70: CMakeFiles/doc] Error 1
make[2]: *** [CMakeFiles/Makefile2:2048: CMakeFiles/doc.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2055: CMakeFiles/doc.dir/rule] Error 2
make: *** [Makefile:179: doc] Error 2
error: builder for '/nix/store/sjm09f0xf5pci3b5cdmv7yw7ivc6qv4f-mir-2.12.0.drv' failed with exit code 2;
```

OPNA2608:
```sh
rm -f  ${CMAKE_BINARY_DIR}/doc/html/{*,search} || true
```

This seems to work better.

AlanGriffiths:
This appears to be a change in `rm` behaviour. I'll come up with a more elegant solution

PR #2817, opened 2023-01-31
doc: better removal of existing docs

Saviq:
`rm -f dir` doesn't do.

Fixes: #2800


AlanGriffiths (approved):
This feels more elaborate than is inherent in the problem solved. But it is an improvement, and time is better spent elsewhere.

bors merge

PR #2819, opened 2023-01-31
Replace input::Scene::for_each() with input_surface_at()

wmww:
`for_each()` was only being used to implement finding the topmost input surface. Our `input::Scene` implementation already has a more performant function to find the topmost surface.

### DIFF
--- a/tests/unit-tests/scene/test_surface_stack.cpp
+++ b/tests/unit-tests/scene/test_surface_stack.cpp
@@ -531,9 +531,7 @@ TEST_F(SurfaceStack, scene_observer_can_query_scene_within_surface_exists_notifi
     MockSceneObserver observer;
 
     auto const scene_query = [&]{
-        stack.for_each([&](std::shared_ptr<mi::Surface> const& surface){
-            EXPECT_THAT(surface.get(), Eq(stub_surface1.get()));
-        });
+        EXPECT_THAT(stack.input_surface_at({}).get(), Eq(stub_surface1.get()));
     };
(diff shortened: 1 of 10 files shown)


tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -531,9 +531,7 @@ TEST_F(SurfaceStack, scene_observer_can_query_scene_within_surface_exists_notifi
     MockSceneObserver observer;
 
     auto const scene_query = [&]{
-        stack.for_each([&](std::shared_ptr<mi::Surface> const& surface){
-            EXPECT_THAT(surface.get(), Eq(stub_surface1.get()));
-        });
+        EXPECT_THAT(stack.input_surface_at({}).get(), Eq(stub_surface1.get()));
```
RAOF, line 534: Although it's a bit more effort, I feel we should be actually passing a non-default `Point` here (and in the other tests) rather than relying on `{}` to somehow be inside the `{}` size, `{}` position stub surface.

As it is, it looks to me like these tests erroneously pass? All the surfaces are 0-sized, and `Rectangle::contains` *should* return false for zero-sized rectangles?

RAOF (changes requested):
The core of this looks good, but I think the tests need work. Basically, I think you should duplicate the existing tests for `surface_at` for `input_surface_at`

Then `for_each_enumerates_all_input_surfaces` can just be removed, and `overlays_do_not_appear_in_input_enumeration` can be fixed to check that `surface_at` *and* `input_surface_at` ignore overlays.

AlanGriffiths (approved):
LGTM

bors merge

PR #2820, opened 2023-01-31
Drop some legacy defines

AlanGriffiths:
We no longer care about the older versions that need these


graysonguarino (approved):
LGTM

bors merge

PR #2821, opened 2023-02-01
wayland: Add missing cstdint includes

Conan-Kudo:
This fixes building for Fedora Linux 38, which ships GCC 13.


AlanGriffiths (approved):
bors merge

PR #2829, opened 2023-02-09
More tweaking of the display configuration code

AlanGriffiths:
This makes it easier to add display configuration strategies and applies these strategies when a `display-layout` option is not found in the configuration file (or the file isn't found).


RAOF (approved):
Looks like a sensible approach.

bors r+

PR #2831, opened 2023-02-10
More tidy-up of the display configuration

AlanGriffiths:
* Roll the "default" layout into the `layout_strategies` table
* Write the .display configuration file before applying the selected layout to preserve the default settings
* If an output "appears" after the .display configuration is written apply the named strategy (if found) and use those values by default


graysonguarino (approved):
LGTM!

bors merge

PR #2832, opened 2023-02-13
Add a .display-layout configuration file

AlanGriffiths:
In cases where a .display configuration fle is used add an additional .display-layout configuration file.

Creating or updating this file will update the active display layout.

### DIFF
--- a/src/miral/static_display_config.cpp
+++ b/src/miral/static_display_config.cpp
@@ -108,6 +108,7 @@ miral::StaticDisplayConfig::StaticDisplayConfig(std::string const& filename) :
     ReloadingYamlFileDisplayConfig(::basename(filename.c_str()))
 {
     config_path(filename.substr(0, filename.size() - basename.size()));
+    check_for_layout_override();
 
     std::ifstream config_file{filename};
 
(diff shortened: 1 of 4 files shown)


src/miral/static_display_config.cpp
```diff
@@ -638,3 +655,15 @@ void miral::ReloadingYamlFileDisplayConfig::auto_reload()
     }
 }
 
+void miral::ReloadingYamlFileDisplayConfig::check_for_layout_override()
+{
+    std::lock_guard lock{mutex};
+    if (!config_path_) return;
+
+    if (std::ifstream layout_file{config_path_.value() + "/" + basename + "-layout"})
```
wmww, line 663: The `"-layout"` string literal only appears twice, but still seems like it would be a better idea to not repeat it, either by putting it in a const global, a `layout_override_for_basename(basename)` standalone helper or something else.

wmww (changes requested):
Other than the one thing looks good to me

RAOF (approved):
LGTM

wmww (approved):
bors r+

PR #2835, opened 2023-02-14
Make sure to call connection_is_allowed on xwayland sessions

mariogrip:
As xwayland/x11 has no way of telling us gid uid we assume it's the same as mir itself.

Note i return a nullptr if session is not allowed and then closes the xwayland session, but im not sure if this is the best way of doing this.


src/server/frontend_xwayland/xwayland_client_manager.cpp
```diff
@@ -80,6 +84,11 @@ auto mf::XWaylandClientManager::session_for_client(pid_t client_pid) -> std::sha
     }
     else
     {
+        if (!session_authorizer->connection_is_allowed({client_pid, getuid(), getgid()}))
```
AlanGriffiths, line 87: Assuming `client_pid` is sensible, there ought to be a way to get gid/uid from that. This seems clunky but...

```c++
struct stat proc_stat;
stat(std::format("/proc/{}", client_pid).c_str(), &proc_stat);
auto const uid = proc_stat.st_uid;
auto const gid = proc_stat.st_gid;
```

mariogrip: I was considering the exact same thing, so i'll try to add this.

mariogrip: Implemented this, but did not use std::format as that require gcc 13++

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -688,12 +688,17 @@ void mf::XWaylandSurface::attach_wl_surface(WlSurface* wl_surface)
...
     reply_functions.push_back(connection->read_property(
         window, connection->_NET_WM_PID,
         XCBConnection::Handler<uint32_t>{
             [&](uint32_t pid)
             {
                 local_client_session = client_manager->session_for_client(pid);
+                if (!local_client_session) {
+                    rejected = true;
+                    return;
+                }
```
AlanGriffiths, line 701:
```suggestion
                if (!local_client_session)
                {
                    rejected = true;
                    return;
                }
```

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -709,6 +714,12 @@ void mf::XWaylandSurface::attach_wl_surface(WlSurface* wl_surface)
         reply_function();
     }
 
+    if (rejected) {
+        scene_surface_close_requested();
+        close();
+        return;
+    }
```
AlanGriffiths, line 721:
```suggestion
    if (rejected)
    {
        scene_surface_close_requested();
        close();
        return;
    }
```

PR #2840, opened 2023-02-16
[Wayland] Don't kill bad clients

AlanGriffiths:
Some toolkits & clients are not deleting the surface role before deleting the surface. 

Killing the connection when they do this is too harsh: Wayland servers are not conformance test suites, and can be flexible in what they expect.

There is text in wl-shell that directs that, despite the client not being allowed to do this, the server should delete the role object. And we already handled that case this way. Mutter does the same with xdg-shell, so there is prior art for what is proposed here.

We have a different workaround in layer-shell, as squeekboard doesn't delete the surface role before the surface, but also crashes if the role object is destroyed by the server.

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -418,6 +418,10 @@ void mf::WindowWlSurfaceRole::surface_destroyed()
         log_warning("wl_surface@%s destroyed before associated role",
                     (surface ? std::to_string(wl_resource_get_id(surface.value().resource)) : "?").c_str());
 
+        // This isn't strictly correct (as it only applies to wl-shell) but is commonly assumed (e.g. by SDL2) and
+        // implemented (e.g. Mutter for xdg-shell)
+        // wl_shell_surface: "On the server side the object is automatically destroyed when the related wl_surface is
+        // destroyed"
         destroy_role();
     }
     else
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -410,13 +410,15 @@ void mf::WindowWlSurfaceRole::surface_destroyed()
 {
     if (!client->is_being_destroyed())
     {
+        mark_destroyed();
+
         // "When a client wants to destroy a wl_surface, they must destroy this 'role object' before the wl_surface"
         // NOTE: the wl_shell_surface specification seems contradictory, so this method is overridden in its implementation
         // NOTE: it's also overridden in layer shell, for reasons explained there
-        BOOST_THROW_EXCEPTION(std::runtime_error{
-            "wl_surface@" +
-            (surface ? std::to_string(wl_resource_get_id(surface.value().resource)) : "?") +
-            " destroyed before associated role"});
+        log_warning("wl_surface@%s destroyed before associated role",
+                    (surface ? std::to_string(wl_resource_get_id(surface.value().resource)) : "?").c_str());
+
+        delete this;
     }
     else
     {
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -418,7 +418,7 @@ void mf::WindowWlSurfaceRole::surface_destroyed()
         log_warning("wl_surface@%s destroyed before associated role",
                     (surface ? std::to_string(wl_resource_get_id(surface.value().resource)) : "?").c_str());
 
-        delete this;
+        destroy_role();
     }
     else
     {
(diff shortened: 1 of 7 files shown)


wmww:
What toolkits? Using what protocols? Are there upstream bugs for this behavior? Are they doing it for any particular reason?

AlanGriffiths:
> What toolkits?

The current example I have is SDL2 on 23.04 (I tested with neverputt), but I think this has come up before. I'm simply wondering if Mir should be enforcing the standard here, or just accepting the practice.

> Are there upstream bugs for this behavior?

I've not yet looked.

wmww (approved):
If other compositors aren't enforcing this error I'm fine with us not enforcing it either. Note though that we should confirm with `wayland-debug` clients are *actually* doing the incorrect thing on other compositors, in case the initial error is actually in Mir somehow.

AlanGriffiths:
Well here are the corresponding sequences:

Mir
```plain
[3991439.738]  -> wl_buffer@60.destroy()
[3991439.765]  -> wl_buffer@48.destroy()
[3991439.778]  -> wl_buffer@47.destroy()
[3991439.795]  -> wl_buffer@45.destroy()
[3991439.854]  -> zwp_idle_inhibitor_v1@37.destroy()
[3991439.879]  -> wl_surface@29.destroy()
[3991442.516]  -> zwp_linux_dmabuf_v1@32.destroy() 
Creating a window (640x480, windowed)
[3991444.462]  -> wl_display@1.get_registry(new id wl_registry@35)
[3991444.527]  -> wl_display@1.sync(new id wl_callback@54)
[3991444.561] wl_display@1.delete_id(60)
[3991444.575] wl_display@1.delete_id(48)
[3991444.582] wl_display@1.delete_id(47)
[3991444.589] wl_display@1.delete_id(45)
[3991444.595] wl_display@1.delete_id(37)
[3991444.603] wl_display@1.error(wl_display@1, 3, "Mir internal error processing WlSurface::~WlSurface() request")
wl_display@1: error 3: Mir internal error processing WlSurface::~WlSurface() request
```

Mutter:
```plain
4031110.116]  -> wl_buffer@51.destroy()
[4031110.133]  -> wl_buffer@40.destroy()
[4031110.145]  -> wl_buffer@55.destroy()
[4031110.158]  -> zwp_linux_dmabuf_feedback_v1@42.destroy()
[4031110.207]  -> wl_surface@38.destroy()
[4031112.114]  -> zwp_linux_dmabuf_v1@37.destroy()
Creating a window (640x480, windowed)
[4031114.542]  -> wl_display@1.get_registry(new id wl_registry@49)
[4031114.611]  -> wl_display@1.sync(new id wl_callback@52)
[4031127.048] wl_display@1.delete_id(51)
[4031127.097] wl_display@1.delete_id(40)
[4031127.114] wl_display@1.delete_id(55)
[4031127.127] wl_display@1.delete_id(42)
[4031127.138] wl_display@1.delete_id(41)
[4031127.148] wl_display@1.delete_id(45)
[4031127.158] wl_display@1.delete_id(39)
[4031127.169] wl_display@1.delete_id(43)
[4031127.180] wl_display@1.delete_id(38)
[4031127.190] wl_display@1.delete_id(37)
[4031127.200] wl_display@1.delete_id(52)
```

In both cases there's an `xdg_toplevel@39` around that is simply leaked by the client. (In the Mutter case, we see `wl_display@1.delete_id(39)`)

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -410,13 +410,15 @@ void mf::WindowWlSurfaceRole::surface_destroyed()
...
         // NOTE: the wl_shell_surface specification seems contradictory, so this method is overridden in its implementation
         // NOTE: it's also overridden in layer shell, for reasons explained there
-        BOOST_THROW_EXCEPTION(std::runtime_error{
-            "wl_surface@" +
-            (surface ? std::to_string(wl_resource_get_id(surface.value().resource)) : "?") +
-            " destroyed before associated role"});
+        log_warning("wl_surface@%s destroyed before associated role",
+                    (surface ? std::to_string(wl_resource_get_id(surface.value().resource)) : "?").c_str());
+
+        delete this;
```
wmww, line 421: I'm not sure deleting the role is the best way to handle this situation, perhaps we should just leave it until the client does get around to cleaning it up (or until the client closes if it never does)

AlanGriffiths: There may be a better way to clean up, but I don't think we should be leaking the object.

A client that subsequently tries to delete, or use, the role would have problems with Mutter (which, as mentioned above, has deleted the id)

AlanGriffiths: Specifically, we should probably delete via `wl_resource_destroy(resource)` - except getting to `resource` isn't trivial: typically it is in an inaccessible base class of the derived class.

AlanGriffiths: Oh!

So Squeekboard with layer-surface **does** delete it later

Whereas SDL2 with xdg-shell doesn't delete it at all (and couldn't "later" as Mutter has deleted it anyway)

Messy!

wmww (approved):
This is the least bad thing I can think of
bors r+

PR #2843, opened 2023-02-18
Fix BasicIdleHub behavior

wmww:
I've noticed a number of edge cases that cause `BasicIdleHub` to misbehave. Tests are written, implementation to come.

### DIFF
--- a/tests/unit-tests/scene/test_basic_idle_hub.cpp
+++ b/tests/unit-tests/scene/test_basic_idle_hub.cpp
@@ -386,3 +386,45 @@ TEST_F(BasicIdleHub, when_a_wake_lock_is_held_calling_poke_does_not_restart_idle
         executor.execute();
     }
 }
+
+TEST_F(BasicIdleHub, when_observer_is_idle_and_a_wake_lock_is_acquired_observer_becomes_active)
+{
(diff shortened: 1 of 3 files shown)


tests/unit-tests/scene/test_basic_idle_hub.cpp
```diff
@@ -386,3 +386,45 @@ TEST_F(BasicIdleHub, when_a_wake_lock_is_held_calling_poke_does_not_restart_idle
...
+    EXPECT_CALL(*observer, active()).Times(AnyNumber());
+    hub->register_interest(observer, executor, 5s);
+    EXPECT_CALL(*observer, idle()).Times(0);
+    executor.execute();
+
+    advance_by(6s);
+    executor.execute();
+}
+
+TEST_F(BasicIdleHub, when_weaklock_is_held_new_observer_marked_initially_active)
```
AlanGriffiths, line 419:
```suggestion
TEST_F(BasicIdleHub, when_wakelock_is_held_new_observer_marked_initially_active)
```

RAOF (approved):
LGTM.

I was concerned that the `ms::BasicIdleHub::Multiplexer` changes weren't threadsafe, but the multiplexers are a part of `Synchronised` state, so they don't need to be threadsafe.

bors r+

PR #2844, opened 2023-02-18
Migrate BasicIdleHub to use Synchronised<T>

wmww:
Makes maintaining thread safety easier. Should not have an impact on behavior.


AlanGriffiths (approved):
WFM

bors merge

PR #2848, opened 2023-02-23
[miral] Provide output names

AlanGriffiths:
These are useful for compositors (e.g. Frame) that want to do lexical matching (e.g. HDMI-1: "Kodi")


wmww (approved):
LGTM
bors r+

PR #2852, opened 2023-02-28
Use wayland::Weak in place of raw pointers in XDG surfaces

wmww:
While poking around in WLCS I noticed that we have some potential for use-after-free in XDG surfaces. This clears those issues up.

### DIFF
--- a/src/server/frontend_wayland/xdg_shell_stable.cpp
+++ b/src/server/frontend_wayland/xdg_shell_stable.cpp
@@ -40,7 +40,7 @@ namespace mir
 namespace frontend
 {
 
-class XdgSurfaceStable : mw::XdgSurface
+class XdgSurfaceStable : public mw::XdgSurface
 {
 public:
     static XdgSurfaceStable* from(wl_resource* surface);
@@ -67,7 +67,7 @@ private:
     void set_window_role(WindowWlSurfaceRole* role);
 
     mw::Weak<WindowWlSurfaceRole> window_role_;
-    WlSurface* const surface;
+    mw::Weak<WlSurface> const surface;
 
 public:
     XdgShellStable const& xdg_shell;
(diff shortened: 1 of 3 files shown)


src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -40,7 +40,7 @@ namespace mir
 namespace frontend
 {
 
-class XdgSurfaceStable : mw::XdgSurface
+class XdgSurfaceStable : public mw::XdgSurface
```
RAOF, line 43: I don't think we ever require that `XdgSurfaceStable` is-a `mw::XdgSurface`, do we?

I know it's somewhat weird to use private inheritance, but is it a good idea to make this change? (Actual question)

src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -200,7 +200,14 @@ mf::XdgSurfaceStable::XdgSurfaceStable(wl_resource* new_resource, WlSurface* sur
 
 void mf::XdgSurfaceStable::get_toplevel(wl_resource* new_toplevel)
 {
-    auto toplevel = new XdgToplevelStable{new_toplevel, this, surface};
+    if (!surface)
+    {
+        BOOST_THROW_EXCEPTION(mw::ProtocolError(
+            resource,
+            mw::generic_error_code,
```
RAOF, line 207: Maybe this (and all the other ones) are [defunct_role_object](https://wayland.app/protocols/xdg-shell#xdg_surface:enum:error:entry:defunct_role_object)?

RAOF (commented):
Yay helpful infrastructure!

I'm not sure about the inheritance change, and maybe the error code sent, but I'm not particularly fussed about those; fix or not :woman_shrugging: 

bors delegate=wmww

src/server/frontend_wayland/xdg_shell_stable.cpp
wmww: Creating a `wayland::Weak` out of a surface requires a public inheritence chain from `LifetimeTracker`.

src/server/frontend_wayland/xdg_shell_stable.cpp
wmww: We can't use that yet because we haven't bumped to XDG shell stable v5. Event when we do I don't think that's right since that's an XDG surface error and we're sending the error on the XDG toplevel (because the XDG surface doesn't exist).

PR #2855, opened 2023-03-02
Prevent freetype from linking in an older glib

n3rdopolis:
This should fix https://github.com/MirServer/mir/issues/2854


Saviq:
Hi @n3rdopolis thanks for digging! I made a stab at modernizing our CMake use (some of it is could drink alcohol by now in some countries I imagine ;P).

Can you have a look if #2856 also fixes this for you?

I'll still merge this if it goes through our CI, but just wanted to know that my approach also helps.

Saviq (approved):
Thanks, this is fine!

bors merge

PR #2856, opened 2023-03-02
cmake: modernize find_package and use imported targets

Saviq:


CMakeLists.txt
```diff
@@ -259,36 +256,37 @@ foreach(platform IN LISTS MIR_PLATFORM)
...
-pkg_check_modules(XCB_RENDER REQUIRED xcb-render)
-pkg_check_modules(X11_XCURSOR REQUIRED xcursor)
-pkg_check_modules(DRM REQUIRED libdrm)
-
-include_directories (SYSTEM ${GLESv2_INCLUDE_DIRS})
-include_directories (SYSTEM ${EGL_INCLUDE_DIRS})
-include_directories (SYSTEM ${GLM_INCLUDE_DIRS})
+pkg_check_modules(GLM glm)
+if(NOT GLM_FOUND)
+  find_package(glm REQUIRED)  # There's no glm.pc on Fedora, but find_package(glm) fails on Ubuntu...
```
RAOF, line 265: Urgh :woman_facepalming:

tests/acceptance-tests/wayland/CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
-pkg_check_modules(WLCS REQUIRED wlcs)
-include_directories(${WLCS_INCLUDE_DIRS})
+pkg_check_modules(WLCS REQUIRED IMPORTED_TARGET wlcs)
```
RAOF, line 1: Huh. We don't actually *use* the `PkgConfig::WLCS` anywhere? This seems like it might be a latent bug in our existing CMake?

CMakeLists.txt
```diff
@@ -259,36 +256,37 @@ foreach(platform IN LISTS MIR_PLATFORM)
...
+pkg_check_modules(X11_XCURSOR REQUIRED IMPORTED_TARGET xcursor)
+pkg_check_modules(XCB REQUIRED IMPORTED_TARGET xcb)
+pkg_check_modules(XCB_COMPOSITE REQUIRED IMPORTED_TARGET xcb-composite)
+pkg_check_modules(XCB_RENDER REQUIRED IMPORTED_TARGET xcb-render)
+pkg_check_modules(XCB_XFIXES REQUIRED IMPORTED_TARGET xcb-xfixes)
 
 include(CheckCXXSymbolExists)
 list(APPEND CMAKE_REQUIRED_INCLUDES ${DRM_INCLUDE_DIRS})
 list(APPEND CMAKE_REQUIRED_LIBRARIES ${DRM_LIBRARIES})
+include_directories(SYSTEM ${DRM_INCLUDE_DIRS})
```
RAOF, line 289: Is the reason why we're adding this global include path because there's so much stuff that needs the DRM headers? It might be better if we could avoid this?

tests/acceptance-tests/wayland/CMakeLists.txt
Saviq: I suppose this is so that CMake fails if WLCS is unavailable. This whole subdir is conditional on `MIR_ENABLE_WLCS_TESTS`, so it will only barf when that's `ON`.

CMakeLists.txt
Saviq: And because in a lot of places we only need the headers, and don't actually link (so no `target_link_libraries` to populate `-I`). I could instead go and do `target_include_directories` everywhere that's needed.

tests/acceptance-tests/wayland/CMakeLists.txt
RAOF: acceptance-tests/wayland/test_wlcs_display_server.{cpp,h} both #include WLCS headers, so we should be at least pulling *those* header locations from somewhere!

CMakeLists.txt
RAOF: I guess we don't need to do that right now, but maybe it would be sensible to do eventually.

tests/acceptance-tests/wayland/CMakeLists.txt
RAOF: I've pushed a commit fixing this

RAOF (approved):
I've resolved *my* WLCS quibble; please review that and then we're good to go!

bors delegate=Saviq

Saviq:
Sure, thanks! :)

bors r=RAOF

PR #2858, opened 2023-03-07
Implement XDG poup constraint adjustment support

### LINKED ISSUE #2857: Popup positioning does not seem to respect constraint adjustment
As tested by https://github.com/MirServer/wlcs/pull/276, Mir seems to always use slide constraint adjustment regardless of what's specified.

wmww:
Passes the test, but still doesn't work perfectly with gedit. Might be an issue in precedence order. Will take out of draft when I sort that out. 

Fixes #2857


wmww:
I think I was incorrect about there being issues with gedit. No longer seeing them, and the implementation logic in BasicWindowManager all looks consistent with the protocol.

AlanGriffiths (approved):
LGTM

(retrying failed builds before borsing - looks like infrastructure)

PR #2867, opened 2023-03-16
Bump mirplatform soname

AlanGriffiths:
We've introduced members to mir::graphics::Display, DisplayConfigurationOutput and UserDisplayConfigurationOutput that break ABI


Saviq (approved):
Looks legit!

bors merge

PR #2874, opened 2023-03-20
ci: cancel previous PR runs

Saviq:
Ref.:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

This will cancel any pending runs for a PR, avoiding wasted work and speeding up results when we get throttled down.

### DIFF
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -12,6 +12,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number && format('pr{0}', github.event.number) || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   GetMatrix:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/spread.yml
```diff
@@ -12,6 +12,10 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number && format('pr{0}', github.event.number) || github.run_id }}
```
AlanGriffiths, line 16: I'm not sure I know how to interpret this incantation, but what harm can come of it?

AlanGriffiths (approved):
bors merge

.github/workflows/spread.yml
Saviq: If `github.event.number` is not empty (only in pull requests), use `pr###`, otherwise use `github.run_id` - which will be unique for each run. So outside of pull requests, no cancellations will happen.

PR #2879, opened 2023-03-26
Allow "attached" windows to be "restored" by drag gesture

AlanGriffiths:
We introduced "attached" windows to support layer-shell. But the window manager can use them to support docked applications. But docked applications should be dragged into a restored state. (See https://github.com/Miriway/Miriway/pull/28 for an example of this)

I don't think this affects the layer-shell support as these windows won't initiate a drag gesture.


wmww (approved):
Sure
bors r+

PR #2880, opened 2023-03-27
Drop the dispmanx platform

AlanGriffiths:
Just admitting it: this is dead


RAOF (approved):
Yeah, fair enough. Gone, but not forgotten if it turns out someone wants it.

I'll leave @Saviq to do the final borsing, if he agrees.

Saviq:
> Yeah, fair enough. Gone, but not forgotten if it turns out someone wants it.

Exactly.

> I'll leave @Saviq to do the final borsing, if he agrees.

bors r=RAOF

PR #2881, opened 2023-03-28
Drive-by cleanup

AlanGriffiths:
There's no longer any point in keeping miroil source-compatibile between 1.x and 2.x. It has been introduced and used in both places


wmww (approved):
bors r+

PR #2882, opened 2023-03-28
[mir-libs] bump timeout

AlanGriffiths:
I've seen this "cancelled after 30 minutes" a few times.

Bumping the timeout


Saviq (approved):
"Total duration: 29m 56s" yeah that's close enough ;)

bors merge

PR #2883, opened 2023-03-29
Drop legacy workaround

AlanGriffiths:
I remember having an old laptop (circa 2006) that needed the `MIR_DRM_CURSOR_64x64` workaround. But I don't think it we've encountered a need for this for a long time


wmww (approved):
LGTM
bors r+

PR #2893, opened 2023-04-13
Rename ClipboardSource to DataExchangeSource

AlanGriffiths:
Extract and rename data exchange interface that is common between cut & paste, select & paste, and drag and drop.


Saviq (approved):
Good!

bors merge

PR #2894, opened 2023-04-13
Drop legacy drag & drop code

AlanGriffiths:
This is not functional without the mirclient API and content-hub, and not useful as a framework for further development.

There are miral and miroil APIs that mention drag and drop. For ABI stability these are retained and deprecated, but do nothing.


wmww (approved):
Looks good
bors r+

PR #2895, opened 2023-04-14
server/ShmBacking: Fix leak in `install_sigbus_handler`

RAOF:
We `new` a `struct sigaction` each time this gets called, so we need to either stash it somewhere (as `previous_handler`) or `delete` it.


AlanGriffiths (approved):
bors merge

Issue #2903, opened 2023-04-26
Wayland platform should handle failures in wl_display_flush


RAOF:
As per [the documentation](https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1a8463b6e5f4cf9a2a3ad2d543aedcf429), `wl_display_flush` can fail with EAGAIN in the (highly unlikely) event that there's more to write than can be non-blockingly written.

We should, as per the documentation, handle that by polling the display fd for writableness and re-calling `wl_display_flush` until it fully succeeds.

(We should probably also bail with an error if `wl_display_flush` fails with `errno != EAGAIN`)

cf: #2901

PR #2901, opened 2023-04-26
Wayland platform: flush connection after sending pong

wmww:
Seems to fix #2876, at least on the Wayland platform.

PR #2904, opened 2023-04-27
Wayland platform: flush display before polling

### LINKED ISSUE #2902: Wayland platform should call wl_display_flush before blocking for events
As per the documentation of [`wl_display_flush`](https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1a8463b6e5f4cf9a2a3ad2d543aedcf429).

We block waiting for display events here:
https://github.com/MirServer/mir/blob/7100296bea4de6f5c8e57733d7706e530abc59f2/src/platforms/wayland/display.cpp#L260

(and have a sometimes-flush later in the loop, but this should be unconditional and before blocking)

wmww:
Fixes #2902, alternative to #2901, I've looked into #2903 but that seems a little more complicated (and tricky to test)

### DIFF
--- a/src/platforms/wayland/display.cpp
+++ b/src/platforms/wayland/display.cpp
@@ -257,6 +257,8 @@ try
             }
         }
 
+        wl_display_flush(display);
+
         if (poll(fds, indices, -1) == -1)
         {
             wl_display_cancel_read(display);
@@ -289,7 +291,6 @@ try
                 lock.lock();
             }
             lock.unlock();
-            wl_display_flush(display);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


RAOF (approved):
Looks good to me.

I went ahead and fixed #2903 (or, at least, think I've fixed it) because it seemed easy enough to do while we're here.

That means I can't bors it, though.

AlanGriffiths (approved):
> I went ahead and fixed #2903 (or, at least, think I've fixed it)

I think so too. I agree with @wmww too:

> tricky to test

Yes, without mocking the API (or having a mock server). But the logic added is clearly an improvement.

bors merge

PR #2920, opened 2023-05-08
coverage: split coverage by suite

Saviq:


Saviq:
This is not doing what I'd like it to do… You can't drill into the individual flags' coverage (e.g. to see that a path isn't covered in performance testing).

Issue #2837, opened 2023-02-14
Mir apparently missing the first evdev input device


OPNA2608:
With Mir 2.12.0 launched from a tty / display manager on aarch64 - hardware & Qemu virtual machine - the keyboard stops responding and needs to be disconnected & reconnected. Happens on NixOS unstable (miral-app & Miriway) and Ubuntu 22.10 (source-built Mir, ubuntu-frame snap complains "No appropriate input platform module found" when launched from tty), but seemingly not on x86_64?

<details>
  <summary>Startup log</summary>

```
[2023-02-15 00:16:47.914277] <information> mirserver: Starting
[2023-02-15 00:16:47.927907] < - debug - > mirserver: Using logind for session management
[2023-02-15 00:16:47.929120] <information> mircommon: Loading modules from: /home/puna/devel/Mir/install/lib/mir/server-platform
[2023-02-15 00:16:47.929530] <information> mircommon: Loading module: /home/puna/devel/Mir/install/lib/mir/server-platform/graphics-gbm-kms.so.20
[2023-02-15 00:16:47.929733] <information> mircommon: Loading module: /home/puna/devel/Mir/install/lib/mir/server-platform/graphics-wayland.so.20
[2023-02-15 00:16:47.929799] <information> mircommon: Loading module: /home/puna/devel/Mir/install/lib/mir/server-platform/server-x11.so.20
[2023-02-15 00:16:47.929842] <information> mircommon: Loading module: /home/puna/devel/Mir/install/lib/mir/server-platform/input-evdev.so.8
[2023-02-15 00:16:47.930033] <information> mirserver: Found display driver: mir:gbm-kms (version 2.12.0)
[2023-02-15 00:16:47.939257] < - debug - > gbm-kms: Quirks: checking device with devnode: /dev/dri/card1, driver vc4-drm
[2023-02-15 00:16:48.523143] < -warning- > gbm-kms: Failed to detect whether device /dev/dri/card1 supports KMS, continuing with lower confidence
[2023-02-15 00:16:48.529209] < - debug - > gbm-kms: Quirks: checking device with devnode: (null), driver 
[2023-02-15 00:16:48.530132] < - debug - > gbm-kms: Quirks: checking device with devnode: (null), driver 
[2023-02-15 00:16:48.530760] < - debug - > gbm-kms: Quirks: checking device with devnode: (null), driver 
[...]

PR #508, opened 2018-08-03
LogindConsoleServices: Support VT switching

RAOF:
Add a `VTSwitcher` interface that can be provided by `ConsoleServices`, implement it for `LogindConsoleServices` and `LinuxVirtualTerminal`, and then use it in the `input::VTFilter` to switch VTs, rather than directly opening `/dev/tty0` and frobbing it.

Fixes #459.

PR #2924, opened 2023-05-12
LogindConsoleServices: Un-work-around puzzling logind behaviour.

RAOF:
When logind processes a `ReleaseDevice` request, it sends a final `PauseDevice("gone")` after completing the DBus call. When *we* receive a `PauseDevice("gone")` event for a device we have an active `Device` handle for we call `ReleaseDevice`.

During startup, we call `AcquireDevice`, use that device in probing, then `ReleaseDevice` multiple times before the selected platform ends up calling `AcquireDevice` to get the fd that will be used for the life of the compositor.

This means we see a whole bunch of `PauseDevice("gone")` events during startup. We work around them for DRM devices by just ignoring this type of event entirely, but we don't do that for input devices.

Since input platform probing generally only needs to check the *first* candidate device, this means that we have been `ReleaseDevice`ing the first input device before we actually try to give it to libinput, preventing the first enumerated input device from working at startup.

There's also something of a race here, between the `PauseDevice` signal and the platform trying to `AcquireDevice`. But! The device probing and initial platform setup is done *before* our mainloop is running, so DBus events are only processed when something is pumping the glib mainloop, and the synchronous DBus calls we make here pump the loop. So the first time the `PauseDevice` DBus signal can be handled is during or after the platform's `AcquireDevice` call, by which point there's an entry in the `acquired_devices` list and the `PauseDevice` handler will then helpfully remove it and call `ReleaseDevice`. Urgh.

So, don't do any of this.

Stop ignoring `PauseDevice("gone")` for DRM devices - DRM devices *can* legitimately go away - and stop calling `ReleaseDevice` from the gone handler.

Fixes: #2837

### DIFF
--- a/src/server/console/logind_console_services.cpp
+++ b/src/server/console/logind_console_services.cpp
@@ -692,38 +692,10 @@ void mir::LogindConsoleServices::on_pause_device(
         }
         else if ("gone"s == suspend_type)
         {
-            if (major == 226)
-            {
-                /* This is a DRM device.
-                 *
-                 * During startup, we TakeDevice during probe() to check that we can
-                 * actually use the DRM device; we ReleaseDevice it before returning from probe().
-                 *
-                 * Then, during Platform construction we TakeDevice again.
-                 * For some reason, logind feels the need to send a “gone” signal after the
-                 * second TakeDevice call, which would result in us dropping the device and
-                 * everything breaking.
-                 *
-                 * A DRM device is quite unlikely to *actually* be gone, so just ignore
-                 * PauseDevice("gone") signals for DRM devices.
-                 */
-                mir::log_debug(
-                    "Ignoring logind PauseDevice(\"gone\") event for DRM device %i:%i",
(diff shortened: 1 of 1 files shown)


fredldotme:
Applied the patch on top of Mir from lunar, tested on my M1 MacBook Air, works! :smile:

Saviq:
@RAOF you'll need to convince the author of the workaround (#508) that it's indeed OK to go ;)

OPNA2608:
Also confirming that this fixes the problem for me, both on hardware and test VMs.

RAOF:
> @RAOF you'll need to convince the author of the workaround (#508) that it's indeed OK to go ;)

Oh, no! It *was* working, but further investigation has made it break? Specifically, VT switching to miral-shell now reliably fails for me :fire:.

RAOF:
Closed in favour of #2928; if that needs to be revisited, the linked bugs contain enough detail to implement (one of several)a solution(s).

PR #2925, opened 2023-05-12
codecov: don't complain about coverage going down

Saviq:


AlanGriffiths:
Are we seeing a problem this fixes?

Saviq:
> Are we seeing a problem this fixes?

Codecov will put a red ✗ against the PR, so it can deter reviewers.

AlanGriffiths:
> @AlanGriffiths actually, maybe it's better to disable the statuses altogether:

I think that is clearer on the intent.

Saviq:
> I think that is clearer on the intent.

Done.

AlanGriffiths (approved):
OK, let's try it this way

PR #2926, opened 2023-05-12
ci: get sanitizer results

Saviq:


RAOF:
Ok. So, UBSan looks like it should be good to add to the block-CI set; we don't invoke any behaviour it can detect as UB.

We *should* be able to turn on ASan and disable leak checking, but
* The gcc asan build is baroquen.
* The clang asan build picks up some legitimate bugs - ODR violations, something maybe broken in `ShmMapping::CurrentMapping`, and maybe a UAF in `ObserverMultiplexer` - and then *some*thing in the `UDevWrapperDeathTest`, which may well just be a false positive.

So, current status: We should add UBsan to the list off builds required to pass in CI, and fix the bugs that asan detects.

Saviq:
Parts of this are now in #3022, #3028 #3026.

PR #2928, opened 2023-05-15
Assume evdev platform is supported

AlanGriffiths:
If we're trying the evdev platform, then there is no fallback. Testing that it can work work in practice causes problems.

And this testing doesn't gain much: a marginally better error message when things don't work. ("No appropriate input platform module found")

Fixes: #2837

### DIFF
--- a/src/platforms/evdev/platform_factory.cpp
+++ b/src/platforms/evdev/platform_factory.cpp
@@ -37,63 +37,6 @@ mir::ModuleProperties const description = {
     MIR_VERSION_MICRO,
     mir::libname()
 };
-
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
Note, I've not confirmed a reproducer of #2837 but this should work if I've understood the discussion correctly.

Saviq (approved):
FWIW, this makes sense to me. Especially since the first input device could well come after Mir starts up?

wmww (approved):
Seems sensible to me. I've tested and it does appear to fix the problem on my XPS.

PR #2929, opened 2023-05-15
ci: cancel previous merge groups

Saviq:
If we're triggered by a new `merge_group`, any previous groups are invalid already.


Saviq:
Hmm no. This is handled here:

image

Issue #2943, opened 2023-05-18
Broken X11 window controls (titlebar, drag, resize)


Saviq:
The server-side X11 window decoration controls don't work. You can't drag or close X11 apps.

AlanGriffiths:
I don't see this with Mir 2.13 (ppa:mir-team/release), only with main (ppa:mir-team/dev)

AlanGriffiths:
And.. this is the commit that broke it: 8c2f31bf7598e4dd4f7e9e87dbd59b7d4549f2ef

PR #2945, opened 2023-05-18
Revert #2785

AlanGriffiths:
I couldn't spot why #2785 (specifically 8c2f31bf7598e4dd4f7e9e87dbd59b7d4549f2ef) causes #2943.

As we're not (yet) relying on these changes, just back them out.

Fixes: #2943


Saviq (approved):
Thanks!

PR #2951, opened 2023-05-27
Pass event instead of timestamp to move/resize requests

### LINKED ISSUE #1792: BasicWindowManager does not always use correct input event for move/resize

wmww:
Fixes #1792 the way I've planned to for some time


RAOF (approved):
Yeah, this seems sensible.

Issue #2830, opened 2023-02-09
Not all "clients" trigger a SessionAuthorizer::connection_is_allowed


mariogrip:
Seems like some applications/clients do not trigger a connection_is_allowed call, for example firefox running on xwayland.  Seems might be similar as https://github.com/MirServer/mir/pull/117

mariogrip:
So this seems to happen due to https://github.com/MirServer/mir/pull/1709 where it creates a session without asking SessionAuthorizer.

@wmww what to do here? I wonder if we should not rely on SessionAuthorizer to create session *scaffolding* in qtmir

mariogrip:
Here is a (unfinished/crude) example that works for me. but get{gid/uid} is not correct and it will not reject applications.

``` 
diff --git a/src/server/frontend_wayland/wayland_connector.cpp b/src/server/frontend_wayland/wayland_connector.cpp
index ceeb2b355c..2cf4d819f7 100644
--- a/src/server/frontend_wayland/wayland_connector.cpp
+++ b/src/server/frontend_wayland/wayland_connector.cpp
@@ -298,6 +298,7 @@ mf::WaylandConnector::WaylandConnector(
         display.get(),
         executor,
         shell,
+        session_authorizer,
         main_clipboard,
         primary_selection_clipboard,
         text_input_hub,
diff --git a/src/server/frontend_wayland/wayland_connector.h b/src/server/frontend_wayland/wayland_connector.h
index b688280213..84868a795e 100644
--- a/src/server/frontend_wayland/wayland_connector.h
+++ b/src/server/frontend_wayland/wayland_connector.h
@@ -82,6 +82,7 @@ public:
         wl_display* display;
         std::shared_ptr<Executor> wayland_executor;
         std::shared_ptr<shell::Shell> shell;
+        std::shared_ptr<SessionAuthorizer> session_authorizer;
         std::shared_ptr<scene::Clipboard> main_clipboard;
         std::shared_ptr<scene::Clipboard> primary_selection_clipboard;
         std::shared_ptr<scene::TextInputHub> text_input_hub;
diff --git a/src/server/frontend_wayland/wayland_default_configuration.cpp b/src/server/frontend_wayland/wayland_default_configuration.cpp
index c04188c5ee..6d2df23e8f 100644
--- a/src/server/frontend_wayland/wayland_default_configuration.cpp
+++ b/src/server/frontend_wayland/wayland_default_configuration.cpp
[...]

PR #2953, opened 2023-05-31
Make sure to call connection_is_allowed on xwayland sessions

mariogrip:
As xwayland/x11 has no way of telling us gid and uid so we try to get this from running stat on /proc/(pid)

Same as https://github.com/MirServer/mir/pull/2835 but fixed issues pointed out by @AlanGriffiths 

Part of fix for: https://github.com/MirServer/mir/issues/2830

### DIFF
--- a/src/server/frontend_xwayland/xwayland_client_manager.cpp
+++ b/src/server/frontend_xwayland/xwayland_client_manager.cpp
@@ -86,20 +86,19 @@ auto mf::XWaylandClientManager::session_for_client(pid_t client_pid) -> std::sha
     }
     else
     {
-        struct stat proc_stat;
-        std::string proc = "/proc/";
-        proc += std::to_string(client_pid);
+        auto const proc = "/proc/" + std::to_string(client_pid);
 
-        // Fallback to get(uid,gid) if we can't stat
-        auto uid = getuid();
-        auto gid = getgid();
-        if (stat(proc.c_str(), &proc_stat) != -1)
+        struct stat proc_stat;
+        if (stat(proc.c_str(), &proc_stat) == -1)
         {
-            uid = proc_stat.st_uid;
-            gid = proc_stat.st_gid;
+            log_debug("Failed to get uid & gid for PID %d using stat(%s, ...), falling back to get(uid,gid)", 
+                      client_pid, proc.c_str());
+
+            proc_stat.st_uid = getuid();
+            proc_stat.st_gid = getgid();
         }
 
-        if (!session_authorizer->connection_is_allowed({client_pid, uid, gid}))
+        if (!session_authorizer->connection_is_allowed({client_pid, proc_stat.st_uid, proc_stat.st_gid}))
         {
             log_error("X11 session not authorized for PID %d, rejecting!", client_pid);
             return nullptr;
--- a/src/server/frontend_xwayland/xwayland_client_manager.cpp
+++ b/src/server/frontend_xwayland/xwayland_client_manager.cpp
@@ -20,6 +20,10 @@
 #include "mir/shell/shell.h"
 #include "mir/scene/session.h"
 #include "mir/log.h"
+#include "mir/frontend/session_authorizer.h"
+#include "mir/frontend/session_credentials.h"
+
+#include <sys/stat.h>
 
 namespace mf = mir::frontend;
 namespace msh = mir::shell;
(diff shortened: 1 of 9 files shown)


AlanGriffiths:
> fixed issues pointed out by @AlanGriffiths

Not the build errors though? :disappointed:

src/server/frontend_xwayland/xwayland_client_manager.cpp
```diff
@@ -80,6 +86,24 @@ auto mf::XWaylandClientManager::session_for_client(pid_t client_pid) -> std::sha
...
+        // Fallback to get(uid,gid) if we can't stat
+        auto uid = getuid();
+        auto gid = getgid();
+        if (stat(proc.c_str(), &proc_stat) != -1)
+        {
+            uid = proc_stat.st_uid;
+            gid = proc_stat.st_gid;
+        }
+
+        if (!session_authorizer->connection_is_allowed({client_pid, uid, gid}))
```
AlanGriffiths, line 102:
```suggestion
        auto const proc = "/proc/" + std::to_string(client_pid);

        struct stat proc_stat;
        if (stat(proc.c_str(), &proc_stat) == -1)
        {
            log_debug("Failed to get uid & gid for PID %d using stat(%s, ...), falling back to get(uid,gid)", 
                      client_pid, proc.c_str());

            proc_stat.st_uid = getuid();
            proc_stat.st_gid = getgid();
        }

        if (!session_authorizer->connection_is_allowed({client_pid, proc_stat.st_uid, proc_stat.st_gid}))
```

AlanGriffiths (approved):
I've suggested a slight rewrite for clarity, but this seems good

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -709,6 +715,13 @@ void mf::XWaylandSurface::attach_wl_surface(WlSurface* wl_surface)
         reply_function();
     }
 
+    if (rejected)
+    {
+        scene_surface_close_requested();
```
wmww, line 720: The main job of `scene_surface_close_requested()` is to detect and use the `WM_DELETE_WINDOW` protocol, which allows clients to receive close requests (for example so they can show a "save your work?" popup instead of just being killed). Since the window is not being opened and closing is not optional it seems to me this shouldn't be used.

Perhaps this call can be removed entirely? Or perhaps replaced with a call to `xcb_kill_client()` (as `scene_surface_close_requested()` does when `WM_DELETE_WINDOW` isn't supported)?

I haven't tested it, but my guess would be the most correct thing to do would be to call `xcb_kill_client()` and then `connection->flush()` after the call to `close()`

wmww (dismissed):

src/server/frontend_xwayland/xwayland_surface.cpp
AlanGriffiths: Agree this needs testing to confirm, but does sound like the right approach

AlanGriffiths (changes requested):
I've "suggested" @wmww's improvement, which seems fine in my testing.

Note this change doesn't affect apps that _don't_ set `_NET_WM_PID` (xeyes for instance): these are connected regardless. I think the SessionAuthorizer should also have a say on these? (Even if it is simply passed the Xwayland ids.)

Also of significance is that all the X11 applications that I found that set the `_NET_WM_PID` atom also support a Wayland backend. (Which limits the usefulness of this change.)

src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -688,12 +688,18 @@ void mf::XWaylandSurface::attach_wl_surface(WlSurface* wl_surface)
 
     std::shared_ptr<XWaylandClientManager::Session> local_client_session;
     std::shared_ptr<ms::Session> session;
+    bool rejected = false;
     reply_functions.push_back(connection->read_property(
         window, connection->_NET_WM_PID,
```
RAOF, line 693: We can actually get this generically, rather than relying on the client to set `_NET_WM_PID`, via the XResources extension. Mutter has [some code](https://gitlab.gnome.org/GNOME/mutter/-/commit/e5347af0005a68ea9804690658abd876b2186ab6#abaf68a20e31f3ff44fa4596691ecc717f319bd5_129_194) for this.

RAOF (commented):
I agree with Alan's suggested change.

To make it more useful we should use something like the linked Mutter code to grab client PID without requiring `_NET_WM_PID` support. I don't object to landing this first and adding the full support later, though.

AlanGriffiths:
> I don't object to landing this first and adding the full support later, though.

Agreed. With the suggested change this can land. But there should be a follow-up to close the "no `_NET_WM_PID`" loophole. (And `xcb_res_query_client_ids()` seems like the best idea so far.)

RAOF (approved):

AlanGriffiths (approved):

AlanGriffiths:
Note that this doesn't really fix #2830 as it relies on the client setting `_NET_WM_PID`

PR #2954, opened 2023-05-31
Clean out legacy of mirclient drag-and-drop support

AlanGriffiths:
It is now clear we don't need this at all for implementing drag and drop in Wayland.

So delete the implementation and explicitly deprecate the APIs

### DIFF
--- a/include/common/mir/events/event_builders.h
+++ b/include/common/mir/events/event_builders.h
@@ -200,8 +200,8 @@ EventUPtr make_touch_event(
 EventUPtr clone_event(MirEvent const& event);
 void set_window_id(MirEvent& event, int window_id);
 
+[[deprecated("Not meaningful: legacy of mirclient API")]]
 EventUPtr make_start_drag_and_drop_event(frontend::SurfaceId const& surface_id, std::vector<uint8_t> const& handle);
-void set_drag_and_drop_handle(MirEvent& event, std::vector<uint8_t> const& handle);
 
 [[deprecated("Internally functions from event_helpers.h should be used, externally this should not be needed")]]
 void transform_positions(MirEvent& event, mir::geometry::Displacement const& movement);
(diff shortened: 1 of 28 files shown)


include/common/mir/events/event_builders.h
```diff
@@ -200,8 +200,8 @@ EventUPtr make_touch_event(
 EventUPtr clone_event(MirEvent const& event);
 void set_window_id(MirEvent& event, int window_id);
 
+[[deprecated("Not meaningful: legacy of mirclient API")]]
 EventUPtr make_start_drag_and_drop_event(frontend::SurfaceId const& surface_id, std::vector<uint8_t> const& handle);
-void set_drag_and_drop_handle(MirEvent& event, std::vector<uint8_t> const& handle);
```
wmww, line 204: I'm confused. This is in a public header and you mark the above function deprecated, so why are you allowed to delete this one?

wmww (commented):
Mostly looks good, just the one question

include/common/mir/events/event_builders.h
AlanGriffiths: I'm not. Good  catch

wmww (approved):
LGTM
bors r+

Saviq:
This caused a test to timeout:

```
356/356 Test #344: mir_umock_unit_tests.UdevWrapperDeathTest.* ......................................................................................***Timeout 600.03 sec
Running main() from main.cpp
Note: Google Test filter = UdevWrapperDeathTest.*:-:--
[==========] Running 2 tests from 1 test suite.
[----------] 2 tests from UdevWrapperDeathTest
[ RUN ] UdevWrapperDeathTest.DereferencingEndReturnsInvalidObject


99% tests passed, 1 tests failed out of 356

Total Test time (real) = 665.28 sec

The following tests FAILED:
	344 - mir_umock_unit_tests.UdevWrapperDeathTest.* (Timeout)
```

AlanGriffiths:
> This caused a test to timeout:

I don't understand how these changes can impact that test.

Should we revert to prove it?

Saviq:
> I don't understand how these changes can impact that test.

Weird. It failed like above across all Ubuntu releases in the PPA on amd64… But subsequent builds (post #2955) went fine, except for one.

As you were.

Issue #2900, opened 2023-04-19
22: performance tests dying with "Bus error" on arm*


Saviq:
```
$ sudo mir-test-tools.performance-test
Running main() from main.cpp
[==========] Running 7 tests from 4 test suites.
[----------] Global test environment set-up.
[----------] 2 tests from GLMark2Wayland
[ RUN      ] GLMark2Wayland.fullscreen
Saving server logs to: /tmp/GLMark2Wayland_fullscreen_server.log
Bus error
```

<details>
<summary>server.log</summary>

```
touch: cannot touch '/tmp/.X11-unix/X0': Permission denied
touch: cannot touch '/tmp/.X11-unix/X1': Permission denied
touch: cannot touch '/tmp/.X11-unix/X2': Permission denied
touch: cannot touch '/tmp/.X11-unix/X3': Permission denied
touch: cannot touch '/tmp/.X11-unix/X4': Permission denied
touch: cannot touch '/tmp/.X11-unix/X5': Permission denied
[2023-04-19 15:27:02.529370] <information> mirserver: Starting
[2023-04-19 15:27:02.538717] < - debug - > mirserver: Not using logind for session management: Seat has no active session
[2023-04-19 15:27:02.539972] < - debug - > mirserver: Using Linux VT subsystem for session management
[2023-04-19 15:27:02.540358] <information> mircommon: Loading modules from: /snap/mir-test-tools/11306/usr/lib/aarch64-linux-gnu/mir/server-platform
[2023-04-19 15:27:02.540656] <information> mircommon: Loading module: /snap/mir-test-tools/11306/usr/lib/aarch64-linux-gnu/mir/server-platform/graphics-gbm-kms.so.20
[2023-04-19 15:27:02.540724] <information> mircommon: Loading module: /snap/mir-test-tools/11306/usr/lib/aarch64-linux-gnu/mir/server-platform/graphics-wayland.so.20
[2023-04-19 15:27:02.540775] <information> mircommon: Loading module: /snap/mir-test-tools/11306/usr/lib/aarch64-linux-gnu/mir/server-platform/server-x11.so.20
[2023-04-19 15:27:02.540823] <information> mircommon: Loading module: /snap/mir-test-tools/11306/usr/lib/aarch64-linux-gnu/mir/server-platform/input-evdev.so.8
[2023-04-19 15:27:02.544302] <information> mirserver: Found display driver: mir:gbm-kms (version 2.13.0)
[...]

PR #2956, opened 2023-06-02
Ensure that the SetWindowManagementPolicy lives long enough

AlanGriffiths:
Having SetWindowManagementPolicy as a temporary didn't guarantee it lived long enough to be used by the server startup. Making it a member variable does (as it lasts much longer - until server shutdown).

Fixes: #2900


Saviq (approved):
Looks legit. Will have to wait for lab results to confirm that indeed fixes.

AlanGriffiths:
> Will have to wait for lab results to confirm that indeed fixes

It works on my machine. :wink:

PR #2959, opened 2023-06-08
Support screen lockers

wmww:
Adds support for https://wayland.app/protocols/ext-session-lock-v1


wmww:
The latest release of Swaylock works with this, but current master doesn't due to https://github.com/swaywm/swaylock/issues/299. I have proposed https://github.com/swaywm/swaylock/pull/300 to fix the problem on their side.

It's also possible for us to add logic into `WlSurface` that lets it grab the SHM buffer data on attach, but it's unclear if this is desirable (according to @RAOF Weston does the same thing Mir currently does, and the protocol doesn't specify anything about being allowed to destroy buffers before commit.

wmww:
Includes #2961

include/miral/miral/window_info.h
```diff
@@ -132,7 +132,7 @@ struct WindowInfo
     auto focus_mode() const -> MirFocusMode;
 
     /// If this surface should be shown while the compositor is locked
-    /// \remark Since MirAL 3.3
+    /// \remark Since MirAL 3.9
```
wmww, line 135: Oops

AlanGriffiths:
> Includes #2961

Rebased as that has landed, and wasn't set as the target

AlanGriffiths:
> The latest release of Swaylock works with this

Note that on 22.04 (swaylock 1.5) we see:

```bash
$ swaylock -f
2023-06-12 10:08:11 - [main.c:1186] Compositor does not support the input inhibitor protocol, refusing to run insecurely
```

OTOH with 23.04 this works

AlanGriffiths (approved):
This all looks uncontentious.

The question is: do we want to delay landing until we have wlcs tests?

Personally, I'm fine with landing now. But need another opinion (especially as I've committed some "tidy up" to the branch).

Saviq (approved):
> The question is: do we want to delay landing until we have wlcs tests?

No point letting it rot.

wmww:
FYI https://github.com/swaywm/swaylock/pull/300 landed so swaylock shouldn't break for us in future versions.

PR #2961, opened 2023-06-09
Refactor deleted_for_resource into ResourceLifetimeTracker

### LINKED ISSUE #2960: Use-after-free if buffer is destroyed before used
In `WlSurface::commit` we use `deleted_flag_for_resource()` to make sure we only release buffers if they still exist, however this assumes the buffer exists when the surface is committed. If a buffer is attached and then destroyed before the surface is committed than Mir can crash. I encountered this with swaylock.

wmww:
Fixes #2960


wmww:
Note to self (or whoever takes over this PR): `run_unless()` can also be dropped now

AlanGriffiths (approved):
LGTM

PR #2963, opened 2023-06-15
Simplify SurfaceInputDispatcher

AlanGriffiths:
This removes some broken (and untested) support for multiple pointers.

The code removed incorrectly treated each input device as driving a separate pointer. That isn't true in general (a touchpad and associated buttons drive a single pointer) or specifically (multiple pointer input devices drive a single pointer in Mir).

The problem this code caused was that using a different device causes spurious leave/enter transitions.

After clicking on surface A...

1. Using the trackpad to move the pointer from surface A to surface B causes A.leave/B.enter (that's correct).

2. Subsequently pressing a button causes A.leave/B.enter (that's *incorrect*).

3. Subsequently using a newly attached mouse to move the pointer within surface B causes yet another B.enter (that's *incorrect*).

These spurious leave/enter events have typically not caused problem, so this hasn't been noticed. But these spurious transitions will be problematic for drag-n-drop.

### DIFF
--- a/src/server/input/surface_input_dispatcher.cpp
+++ b/src/server/input/surface_input_dispatcher.cpp
@@ -215,7 +215,7 @@ struct SceneChangeContext
 
 SceneChangeContext context_for_event(
     MirEvent const* last_pointer_event,
-    std::function<std::shared_ptr<mi::Surface>*(MirInputDeviceId)> const& get_current_target,
+    std::shared_ptr<mi::Surface>& current_target,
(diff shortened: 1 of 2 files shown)


RAOF (approved):
An alternative here might be to coopt the `MirInputDeviceId` to be seat-based. But… no, let's not do that.

LGTM.

AlanGriffiths:
> An alternative here might be to coopt the `MirInputDeviceId` to be seat-based. But… no, let's not do that.

Yeah, we'll have to revisit if/when we want to support multiple pointers (multi-seat comes to mind). But this is simpler for now.

PR #2965, opened 2023-06-16
Drag and drop

AlanGriffiths:
This is a cleaned up version of the "spike" implementation.

I need to have a think about how automate testing. The problem for testing is that there's a lot of state spread between the source, the target and parts of this implementation.

While I think about that (that may take some time), please feel free to  do some exploratory testing and to review the code.

Easy way to test:

    snap install --classic --channel edge/mir-pr2965 miriway

### DIFF
--- a/src/include/server/mir/scene/data_exchange.h
+++ b/src/include/server/mir/scene/data_exchange.h
@@ -19,8 +19,9 @@
 
 #include "mir/fd.h"
 
-#include <vector>
+#include <optional>
 #include <string>
+#include <vector>
 
 namespace mir
 {
@@ -40,6 +41,24 @@ public:
     /// Instructs the source to start sending it's data in the given mime type to the given fd.
     virtual void initiate_send(std::string const& mime_type, Fd const& target_fd) = 0;
 
+    /// The destination has cancelled the (DnD) transfer
+    virtual void cancelled() {}
+
+    /// The user has dropped the (DnD) transfer
+    virtual void dnd_drop_performed() {}
+
+    /// The DnD actions supported
+    virtual auto actions() -> uint32_t { return 0; }
+
+    /// target accepted a mime type
+    virtual void offer_accepted(std::optional<std::string> const& mime_type) {(void)mime_type; }
+
+    /// target indicated an action
+    virtual uint32_t offer_set_actions(uint32_t dnd_actions, uint32_t preferred_action)
+        { (void)dnd_actions; (void)preferred_action; return 0; }
+
+    /// The client has finished the (DnD) transfer
+    virtual void dnd_finished() {}
 private:
     DataExchangeSource(DataExchangeSource const&) = delete;
     DataExchangeSource& operator=(DataExchangeSource const&) = delete;
--- a/src/include/server/mir/scene/data_exchange.h
+++ b/src/include/server/mir/scene/data_exchange.h
@@ -42,23 +42,22 @@ public:
     virtual void initiate_send(std::string const& mime_type, Fd const& target_fd) = 0;
 
     /// The destination has cancelled the (DnD) transfer
-    virtual void cancelled() {}
+    virtual void cancelled() = 0;
 
     /// The user has dropped the (DnD) transfer
-    virtual void dnd_drop_performed() {}
+    virtual void dnd_drop_performed() = 0;
 
     /// The DnD actions supported
-    virtual auto actions() -> uint32_t { return 0; }
+    virtual auto actions() -> uint32_t = 0;
 
     /// target accepted a mime type
-    virtual void offer_accepted(std::optional<std::string> const& mime_type) {(void)mime_type; }
+    virtual void offer_accepted(std::optional<std::string> const& mime_type) = 0;
 
     /// target indicated an action
-    virtual uint32_t offer_set_actions(uint32_t dnd_actions, uint32_t preferred_action)
-        { (void)dnd_actions; (void)preferred_action; return 0; }
+    virtual uint32_t offer_set_actions(uint32_t dnd_actions, uint32_t preferred_action) = 0;
 
     /// The client has finished the (DnD) transfer
-    virtual void dnd_finished() {}
+    virtual void dnd_finished() = 0;
 private:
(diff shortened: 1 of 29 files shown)


src/include/server/mir/scene/data_exchange.h
```diff
@@ -40,6 +41,24 @@ class DataExchangeSource
...
+
+    /// target accepted a mime type
+    virtual void offer_accepted(std::optional<std::string> const& mime_type) {(void)mime_type; }
+
+    /// target indicated an action
+    virtual uint32_t offer_set_actions(uint32_t dnd_actions, uint32_t preferred_action)
+        { (void)dnd_actions; (void)preferred_action; return 0; }
+
+    /// The client has finished the (DnD) transfer
+    virtual void dnd_finished() {}
```
RAOF, line 61: Are these deliberately concrete implementations in this otherwise abstract interface? It's not a leftover from the spike?

src/server/frontend_wayland/wl_data_device.cpp
```diff
@@ -49,24 +73,28 @@ class mf::WlDataDevice::Offer : public wayland::DataOffer
 
     void accept(uint32_t serial, std::optional<std::string> const& mime_type) override
     {
-        (void)serial, (void)mime_type;
+        (void)serial;
```
RAOF, line 76: I think I have a weak preference for `[[maybe_unused]]` annotations instead?

src/server/frontend_wayland/wl_data_device.cpp
```diff
@@ -125,8 +153,46 @@ void mf::WlDataDevice::set_selection(std::optional<wl_resource*> const& source,
...
 }
 
+void mf::WlDataDevice::start_drag(
+    std::optional<wl_resource*> const& source,
+    wl_resource* origin,
+    std::optional<wl_resource*> const& icon,
+    uint32_t serial)
+{
+    // "The [origin surface] and client must have an active implicit grab that matches the serial"
+    if (!origin)
```
RAOF, line 163: I don't *think* we can get here? The protocol doesn't specify this as `allow-none`, so I *think* this would be filtered out before here?

src/server/frontend_wayland/wl_data_device.cpp
```diff
@@ -125,8 +153,46 @@ void mf::WlDataDevice::set_selection(std::optional<wl_resource*> const& source,
...
+    {
+        BOOST_THROW_EXCEPTION(
+            mw::ProtocolError(resource, Error::role, "Origin surface does not exist."));
+    }
+
+    validate_pointer_event(client->event_for(serial));
+
```
RAOF, line 170: We're missing a check that the serial corresponds to the initiation of an implicit-grab *on the origin surface*.

src/include/server/mir/scene/data_exchange.h
AlanGriffiths: A bit of both. It avoids churn in the test code, which was convenient for the spike. As I've not decided on a test strategy I've not addressed it. Yet

src/server/frontend_wayland/wl_data_device.cpp
AlanGriffiths: Pre-existing. :grin:

There's also `uint32_t /*serial*/`

src/server/frontend_wayland/wl_data_device.cpp
AlanGriffiths: Good catch

src/server/frontend_wayland/wl_data_device.cpp
AlanGriffiths: Agreed, missed this when cleaning up the code

AlanGriffiths:
Rebased on main

AlanGriffiths:
OK, I'm removing the `no-merge` tag as I'm working to get testing in place in the mir-ci repo. That is, landing without tests would not be an unmanaged "technical debt".

Saviq:
By messing with this at https://www.javascripttutorial.net/sample/webapis/drag-n-drop-basics/, (`frame-it firefox`) I was able to get the app stuck frantically dragging it around. Something that didn't happen under GNOME.

Similarly, with `frame-it nautilus` I got Nautilus stuck playing with a couple tabs and some folders between them.

AlanGriffiths:
> Oh and the dragged surface doesn't naturally follow where it got dragged from, it always gets repositioned to "cursor at 0,0" at start of drag.

Yes, it is currently a limitation on placing the image. I wasn't aiming to sort that in this PR (and @wmww tells me that Sway has the same limitation)

AlanGriffiths:
> I was able to get the app stuck frantically dragging around

I've not been able to reproduce: Could you be specific about what you did and what you saw?

Saviq:
> > I was able to get the app stuck frantically dragging around
> 
> I've not been able to reproduce: Could you be specific about what you did and what you saw?

When dragging the yellow square all around, the green boxes light up with a red dotted line on hover. When doing that in Firefox on GNOME, it's all working fine.

When doing the same under Miriway, the highlights are sometimes lagging behind, sometimes they just stop working. Maybe it's just shipping to many move events…?

AlanGriffiths:
> When doing the same under Miriway, the highlights are sometimes lagging behind, sometimes they just stop working. Maybe it's just shipping to many move events…?

Thanks! Knowing what to look for helps

AlanGriffiths:
> When dragging the yellow square all around, the green boxes light up with a red dotted line on hover. When doing that in Firefox on GNOME, it's all working fine.

The last couple of commits (especially https://github.com/MirServer/mir/pull/2965/commits/8e052b2589c79ef9c263740829f399407e9f1439) hopefully address that.

AlanGriffiths:
After playing with this a while I found a couple of (probably related) issues:

* Ending the drag gesture when there's no drop area under the mouse doesn't work (the drag cursor remains, the source is not notified); and,
* I Hung the DE by starting the terminal having done the above. (But not been able to reproduce, yet)

RAOF (approved):
I will undoubtedly have missed things, but this matches my understanding of how DnD is meant to work and looks good.

Issue #2968, opened 2023-06-28
eglstreams platform is not naming outputs correctly


AlanGriffiths:
Seen here: https://github.com/MirServer/ubuntu-frame/issues/143#issuecomment-1611187059

PR #2969, opened 2023-06-28
Fix naming of outputs on Nvidia cards

AlanGriffiths:
Fixes: #2968

### DIFF
--- a/src/platforms/eglstream-kms/server/kms_display_configuration.cpp
+++ b/src/platforms/eglstream-kms/server/kms_display_configuration.cpp
@@ -20,13 +20,16 @@
 #include "kms-utils/drm_mode_resources.h"
 #include "mir/graphics/pixel_format_utils.h"
 #include "mir/graphics/egl_error.h"
+#include "mir/graphics/pixel_format_utils.h"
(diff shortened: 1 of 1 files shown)


src/platforms/eglstream-kms/server/kms_display_configuration.cpp
```diff
@@ -72,6 +75,26 @@ kms_connector_type_to_output_type(uint32_t connector_type)
...
 }
 
+void name_outputs(std::vector<std::shared_ptr<mge::kms::EGLOutput>>& outputs)
+{
+    std::map<mg::DisplayConfigurationCardId, std::map<mg::DisplayConfigurationOutputType, int>> card_map;
+
+    for (auto& output : outputs)
+    {
+        auto const type = output->type;
+        auto const index_by_type = ++card_map[output->card_id][type];
```
RAOF, line 85: This might be because I'm still sick, but I had to stare at this a bit to work out what it's actually doing.

But what it's doing is sensible.

RAOF (approved):
Maybe slightly too fancy for my current state, but yeah, this is correct.

PR #2970, opened 2023-06-30
Don't dispatch pointer events on the Wayland thread

AlanGriffiths:


RAOF (approved):
The rest of our input stuff is on different threads, so I'm not *astounded* that it doesn't work on the Wayland thread, but I can't see why this fixes anything?

It's not-incorrect, and obviously fixes something for you, but I can't see how the existing code deadlocks.

AlanGriffiths:
> can't see why this fixes anything?

That's because you've not looked at the drag-n-drop code. That needs to reconfigure the routing of input events (they go to different objects and clients during a drag). That doesn't work so well when the input is dispatched on the Wayland thread.

PR #2972, opened 2023-07-03
Drop dead code

AlanGriffiths:


mattkae (approved):
I have confirmed that `MyMirBlob`, `MockConnector`, and `arbitrary_timestamp` are all unused. Looks good to me !

Issue #2148, opened 2021-08-20
WindowManager::add_display()/remove_display()


AlanGriffiths:
In one of the implementations of this interface SystemCompositorWindowManager we have:
```c++
    void add_display(mir::geometry::Rectangle const& area) override
    __attribute__((deprecated("Mir doesn't reliably call this: it is ignored. Use add_display_for_testing() instead")));
```
And, in this implementation, the outputs are tracked by the `ActiveOutputsListener` interface.

But the other implementation `SystemCompositorWindowManager` uses this function to track the outputs.

There's clearly something wrong with the diagnostic, one of the implementations, or several of these.

AlanGriffiths:
The implementation of `BasicWindowManager` (which ignores this function) is clearly more widely used than `SystemCompositorWindowManager` which does.

As we don't need two ways to do the same thing, we should update `SystemCompositorWindowManager` to use the `ActiveOutputsListener` interface and remove this function.

mattkae:
## Investigation
1. `BasicWindowManager` inherits from `ActiveOutputsListener`.
2. `ActiveOutputsListener` appears to be responsible for managing the state of connected outputs. These methods only appear to be called during `DisplayConfigurationListeners::configuration_applied`. According to `DisplayConfigurationObserver::configuration_applied`, this is applied after the `DisplayConfiguration` has been applied to the hardware.

## Problem

- Whenever a new display is added, `SystemCompositorWindowManager` (which also inherits from `ActiveOutputsListener`) will iterate the map OutputMap. For each surface/output pair, we resize the surface to fit it inside of the Output (so it seems).

## Solution
- Like Alan said, it would be more appropriate to handle the coming and goings of outputs via the `ActiveOutputsListener` interface

AlanGriffiths:
Reopening as there has been no change (so far) to the mess described in this issue.

As [mentioned elsewhere](https://github.com/MirServer/mir/pull/2975#pullrequestreview-1521758940) we've now removed any dependency on the suspect logic and could 
> Drop `shell::WindowManager::add_display()` et alia (which would resolve #2975)

mattkae:
@AlanGriffiths going to resolve the remainder of this one this  morning!

PR #2975, opened 2023-07-06
(#2148) Implementing a mechanism whereby we can override the default Window Manager

mattkae:
# https://github.com/MirServer/mir/issues/2148

## What's new?
- I built a mechanism to override the "default" WindowManager behavior in the mir server via the mirAL runner. My intention here was to bring the default WindowManager creator fully in control of mirAL instead of core mir
- `DefaultWindowManager` will now be the default for all mirAL based compositors. As it stands, it is an almost direct clone of the `SystemCompositorWindowManager`

## Other possible solutions
- I investigated bringing the `ActiveOutputsListener` into core mir, but I ran into a myriad of issues. `ActiveOutputsListener` indirectly relies on the `Output` class which is a core mirAL concept. The refactor would have been large and it would have led to some tricky interface breakages that I don't think are worthwhile for this change
- I also investigated moving `DefaultWindowManager` into mirAL but this would mean that `DefaultServerConfiguration` would have to move with it _or_ I would have to pull off some tricks that I wasn't to keen on for readability's sake

## What's next?
- In the next PR, I will refactor `DefaultWindowManager` to use `ActiveOutputsListener`
- We could deprecate the old `DefaultWindowManager` as well

## Screenshots
[Screencast from 07-06-2023 05:34:45 PM.webm](https://github.com/MirServer/mir/assets/25062299/caafa9f6-8834-45b9-ba6e-b565d099d7f2)
[Screencast from 07-07-2023 03:52:26 PM.webm](https://github.com/MirServer/mir/assets/25062299/eda5f27b-f3df-4c41-8f6c-0034864e8ae5)


AlanGriffiths:
Hmm, the ubuntu-22.04:spread/build/sbuild:ubuntu_arm64 failure has 
```
 [       OK ] ClientSurfaceEventsTest.frame_timestamp_increases (553 ms)
[----------] 1 test from ClientSurfaceEventsTest (554 ms total)

[==========] 1 tests from 1 test cases run. (562ms total elapsed)
[  PASSED  ] 1 test
[XPASS]: Test unexpectedly passes
```

@RAOF  just updated `frame_timestamp_increases`, but it still expectedly fails on my machine (and in most builds).

src/miral/runner.cpp
```diff
@@ -142,6 +143,17 @@ try
 
         // Provide the command line and run the server
         server->set_command_line(argc, argv);
+
+        // TODO: This is a band-aid solution, and perhaps not the best one. The mir core libraries provide a default
+        // window manager, but mirAL would like to provide a different default window manager.
+        server->override_the_default_window_manager_builder([&server](mir::shell::FocusController* focus_controller) -> std::shared_ptr<mir::shell::WindowManager>
```
AlanGriffiths, line 149: Can't we just call `override_the_window_manager_builder()` before `for (auto& option : options)` and drop the new function?

src/server/symbols.map
```diff
@@ -144,6 +144,7 @@ MIR_SERVER_2.9 {
     mir::Server::override_the_session_listener*;
     mir::Server::override_the_shell*;
     mir::Server::override_the_window_manager_builder*;
+    mir::Server::override_the_default_window_manager_builder*;
```
AlanGriffiths, line 147: Wrong stanza. If we need this function (I don't think we do) then it belongs in `MIR_SERVER_2.14`. (The stanza is part of our ABI management and indicates the Mir release the function is introduced in.)

AlanGriffiths (changes requested):
You've a tendency to use eight spaces where the project uses 4. I've flagged some of these, but there are others.

I've made a suggestion that I think reduces the complexity and avoids the need to introduce `Server::override_the_default_window_manager_builder()`.

And we should finish the implementation before landing.

AlanGriffiths:
@mattkae I don't think the build failures are related to your PR, you should ignore them for now and progress the implementation. (I don't see how the wlcs changes would cause them either, but that seems more likely)

src/include/server/mir/server.h
```diff
@@ -290,9 +290,20 @@ class Server
...
+    /// \see DefaultWindowManager src/miral/default_window_manager.h
     void override_the_window_manager_builder(shell::WindowManagerBuilder const wmb);
 
+    /// Sets an override for the default window manager creation behavior.
+    /// If no window manager override is specified via override_the_window_manager_builder,
+    /// the system will fallback to a creator function that returns DefaultWindowManager.
+    /// Implementors may override the DefaultWindowManager via this method.
+    /// \param wmb The default window manager creation function.
+    /// \see Server::override_the_window_manager_builder
+    void override_the_default_window_manager_builder(shell::WindowManagerBuilder const wmb);
```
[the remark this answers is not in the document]
mattkae: Aha. That's a much better solution! I will do that instead.

src/server/symbols.map
mattkae: I don't think we need this function any longer, but I am curious to learn a little bit more about this `symbols.map` file during our chat today. I noticed that I couldn't link with out implementing this line

src/miral/runner.cpp
mattkae: Yes!

src/miral/default_window_manager.cpp
```diff
@@ -0,0 +1,349 @@
...
+void miral::DefaultWindowManager::advise_output_create(miral::Output const& output)
+{
+    std::cout << &output << std::endl;
+//    Locker lock{this};
+//
+//    outputs.add(output.extents());
+//
+//    add_output_to_display_areas(lock, output);
+//    application_zones_need_update = true;
+//    policy->advise_output_create(output);
```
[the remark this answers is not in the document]
mattkae: I will remove these commented out blocks. It looks like I forgot to commit that

src/miral/default_window_manager.cpp
```diff
@@ -0,0 +1,349 @@
...
+namespace msh = mir::shell;
+namespace geom = mir::geometry;
+
+miral::DefaultWindowManager::DefaultWindowManager(
+        msh::FocusController* focus_controller,
+        std::shared_ptr<msh::DisplayLayout> const& display_layout,
+        std::shared_ptr<ms::SessionCoordinator> const& session_coordinator) :
+        focus_controller{focus_controller},
+        display_layout{display_layout},
+        session_coordinator{session_coordinator}
```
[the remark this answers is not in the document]
mattkae: Changed my editor settings to reflect this :+1:

src/server/symbols.map
AlanGriffiths: Look at the end of the file for functions we're adding to the next release

mattkae: Understood! And this file is updated by hand? Or is it generated?

mattkae: Looks like I found the answer [here](https://mir-server.io/doc/dso_versioning_guide.html)

mattkae:
> You've a tendency to use eight spaces where the project uses 4. I've flagged some of these, but there are others.
> 
> I've made a suggestion that I think reduces the complexity and avoids the need to introduce `Server::override_the_default_window_manager_builder()`.
> 
> And we should finish the implementation before landing.

I will finish the implementation now. All initial comments have been addressed!

src/miral/system_compositor_window_manager.cpp
```diff
@@ -0,0 +1,312 @@
...
+{
+    std::lock_guard lock{mutex};
+
+    // When a display gets updated, we reposition all surfaces across their outputs, most likely to fit them to the screen.
+    for (auto const& surface_output_pair : output_map)
+    {
+        if (auto surface = surface_output_pair.first.lock())
+        {
+            auto const output_id = surface_output_pair.second;
+            reposition_surface_in_output(surface, output_id);
```
mattkae, line 292: @AlanGriffiths I made `reposition_surface_in_output` a separate function because I would expect us to only have to update the output that changed in this scenario instead of all the outputs. However, I am still updating all of the outputs in the current implementation. Would updating a single output be the correct thing to do?

AlanGriffiths: Yes, I think that would be a sane optimization

AlanGriffiths (approved):
AFAICS this successfully replicates the behaviour of `shell::DefaultWindowManager` in libmiral.

That leaves the way forward to do a couple of things:

1. Publish a `miral::SystemCompositorWindowManager` API
2. Drop `shell::WindowManager::add_display()` _et alia_ (which would resolve #2975)

Both of these likely need some conversation before progressing.

Meanwhile, there are a couple of nits that would be good to address before landing

src/miral/system_compositor_window_manager.h
```diff
@@ -0,0 +1,152 @@
...
+
+    /// Called the first time each surface owned by the session posts its first buffer
+    virtual void on_session_ready(std::shared_ptr<mir::scene::Session> const& session) const;
+/** @} */
+
+protected:
+    mir::shell::FocusController* const focus_controller;
+    std::shared_ptr<mir::shell::DisplayLayout> const display_layout;
+    std::shared_ptr<mir::scene::SessionCoordinator> const session_coordinator;
+
```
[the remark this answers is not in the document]
mattkae: Agreed !

RAOF:
So, the test failures are a combined result of frame_timestamp_increases *incorrectly* being marked XFAIL (Mir does the right thing here, and has for ages), and the newly-fixed frame_timestamp_increases test being flaky.

https://github.com/MirServer/wlcs/pull/298 fixes the latter problem.

PR #2976, opened 2023-07-11
tests/wlcs: Drop frame_timestamp_increases from expected failures

RAOF:
Mir has actually been sending correct (or, at least, not wildly incorrect) timestamps on frame events for quite some time, but the test itself was broken.

Now that the test is not broken (and was then fixed *again*), remove it from the XFAIL list so CI will pass again.


RAOF:
Before https://github.com/MirServer/wlcs/pull/298 lands this will have *some* test failures ;)

mattkae (approved):
Beautiful :+1:

RAOF:
Now that the WLCS fix has landed, this should be landable. Retriggering the failed builds.

Issue #2629, opened 2022-09-08
Alt+Tab should (optionally?) reorder apps it's switching between


Saviq:
Traditionally a double Alt+Tab toggles between two apps. In Mir, there's a list of clients (by connection order?) Alt+Tab cycles through, so you always have to cycle through the whole set to switch and then switch back.

Not sure if the current behaviour is desirable (as an option?), haven't seen it in other environments.

AlanGriffiths:
Yes, the current code is simplistic: it treats each `Alt+Tab` event in isolation. Consider getting to a third app. Traditionally, you'd do:

1. Alt (down)
2. Tab
3. Tab
4. Alt (up)

`MinimalWindowManager` handles this as:

1. Alt+Tab
2. Alt+Tab

And if the second `Alt-Tab` returned to the first window navigating to a third window wouldn't be possible. Some years ago this was the simplest thing that could possibly work, but we should do something better.

Saviq:
Fixed by #2981

PR #2981, opened 2023-07-18
(#2629) Alt + Tab behavior for the MinimalWindowManager

mattkae:
# #2629 

## Implementation
- When a user clicks Alt + Tab, the MinimalWindowManager calls into the `ApplicationSelector` who raises the next application after the currently selected one.
- When a user clicks Tab again while still holding Alt, the application after the currently raised one will be raised, and so on. (Note: A "raised" application is _not_ focused)
- When a user releases Alt, the currently raised application will become focused
- When an application becomes focused, the `SessionManager` moves it to the front of the app list via `SessionCoordinator::move_to_front_of_list`.

## What's new?

- Added new methods to `FocusController` API:
  - `FocusController:get_next_session(std::shared_ptr<scene::Session>) -> std::shared_ptr<scene::Session>`
  - `FocusController::get-prev_session(std::shared_ptr<scene::Session>) -> std::shared_ptr<scene::Session>` 

- Added a few methods to `WindowManagerTools`:
  - `WindowManagerTools::focus_this_application(Application) -> void`
  - `WindowManagerTools::can_focus_application(Application) -> bool`
  - `WindowManagerTools::get_next_application_info(ApplicationInfo) -> ApplicationInfo`
  - `WindowManagerTools::get_previous_application_info(ApplicationInfo) -> ApplicationInfo`

-  Added a method to `SessionContainer` called `move_to_front_of_list(std::shared_ptr<Session> const& session)`. Whenever focus is gained on a session, it moves to the front of the list.
- Created an `ApplicationSelector` class which handles Alt + Tab behavior in miral 
- Used the `ApplicationSelector` API in `MinimalWindowManager` to handle all Alt + Tab behavior

## Demo

https://github.com/MirServer/mir/assets/25062299/08870924-5617-4106-84a1-c7e78c21c3b8


src/miral/application_selector.cpp
```diff
@@ -0,0 +1,71 @@
+//
+// Created by matthew on 7/18/23.
+//
```
AlanGriffiths, line 3: Not our template

src/miral/application_selector.cpp
```diff
@@ -0,0 +1,71 @@
...
+{}
+
+void ApplicationSelector::start(bool in_reverse)
+{
+    if (is_active())
+    {
+        return;
+    }
+
+    printf("Started\n");
```
AlanGriffiths, line 21: Debug code I presume - it should be removed. (Note you probably want to `flush(stdout)` too, as otherwise it can get confusing)

AlanGriffiths:
Reading the code I don't think this does what I think the user expects. I think we should be swapping with depth N, not rotating from depth N.

Suppose we start with the Z-order A, B, C. And then tap Tab twice while holding Alt.

I think the expectation is we get C, A, B (that is, moving second successor to the top)

 but I think you're proposing we get C, B, A (that is, moving B, then C to the top)

Maybe I'm thinking wrong, and I've not tried the code yet (because the tests don't build and the snap doesn't get built)

mattkae:
> Reading the code I don't think this does what I think the user expects. I think we should be swapping with depth N, not rotating from depth N.
> 
> Suppose we start with the Z-order A, B, C. And then tap Tab twice while holding Alt.
> 
> I think the expectation is we get C, A, B (that is, moving second successor to the top)
> 
> but I think you're proposing we get C, B, A (that is, moving B, then C to the top)
> 
> Maybe I'm thinking wrong, and I've not tried the code yet (because the tests don't build and the snap doesn't get built)

Hi Alan. Sorry this was in a pseudo-complete state while I reviewed the draft. It should work appropriately now. I attached a video to this code review. I believe I have implemented exactly the behavior as you described, but let me know!

AlanGriffiths:
@mattkae as the Snap builds are working, so you can test with:

    snap refresh miriway --channel edge/mir-pr2981

include/miral/miral/window_manager_tools.h
```diff
@@ -18,6 +18,7 @@
 #define MIRAL_WINDOW_MANAGER_TOOLS_H
 
 #include "miral/application.h"
+#include "miral/application_info.h"
```
AlanGriffiths, line 21: Why? There should be no need for this change

include/miral/miral/window_manager_tools.h
```diff
@@ -143,6 +143,10 @@ class WindowManagerTools
     /// move the window
     void drag_window(Window const& window, mir::geometry::Displacement movement);
 
+    /// make the provided application active
+    /// \param application To focus
+    void focus_this_application(Application application);
```
AlanGriffiths, line 148: We annotate with `/// \remark Since MirAL 3.10`

src/miral/application_selector.cpp
```diff
@@ -0,0 +1,130 @@
...
+    auto surface = application ? application->default_surface() : nullptr;
+    if (surface)
+    {
+        auto window = tools.info_for(surface).window();
+        tools.raise_tree(window);
+    }
+    else
+    {
+        mir::log(mir::logging::Severity::warning, MIR_LOG_COMPONENT,
+                 "ApplicationSelector::raise_next: Failed to raise the newly selected window.");
```
AlanGriffiths, line 127: There's a `log_warning()` convenience wrapper...

```suggestion
        mir::log_warning("ApplicationSelector::raise_next: Failed to raise the newly selected window.");
```

src/miral/application_selector.h
```diff
@@ -0,0 +1,72 @@
...
+    /// \return true if it is running, otherwise false
+    auto is_active() -> bool;
+
+private:
+    WindowManagerTools tools;
+    bool reverse = false;
+    bool is_started = false;
+
+    /// The raised application
+    ApplicationInfo selected;
```
AlanGriffiths, line 63: Why are we storing `ApplicationInfo` here? We mostly use `selected.application()`.

I think the right abstraction is `std::weak_ptr<mir::scene::Session>`

src/miral/application_selector.h
```diff
@@ -0,0 +1,72 @@
...
+    /// that immediately follows the currently selected application
+    /// in the SessionContainer's list.
+    /// \param reverse If true, selection will happen in reverse, otherwise forward.
+    /// \returns The raised application, or nullptr in the case of a failure
+    auto start(bool reverse) -> Application;
+
+    /// Lowers the currently raised application and raises the application
+    /// that follows it.
+    /// \returns The raised application, or nullptr in the case of a failure
+    auto next() -> Application ;
```
AlanGriffiths, line 47:
```suggestion
    auto next() -> Application;
```

src/miral/basic_window_manager.h
```diff
@@ -165,12 +165,18 @@ class BasicWindowManager : public virtual mir::shell::WindowManager,
...
     void focus_next_application() override;
     void focus_prev_application() override;
 
     void focus_next_within_application() override;
     void focus_prev_within_application() override;
 
+    auto can_focus_application(Application application) const -> bool override;
+
+    auto get_next_application_info(ApplicationInfo) const -> ApplicationInfo override;
+    auto get_previous_application_info(ApplicationInfo) const -> ApplicationInfo override;
```
AlanGriffiths, line 178:
```suggestion
    auto get_next_application(Application application) const -> Application override;
    auto get_previous_application(Application application) const -> Application override;
```

src/miral/minimal_window_manager.cpp
```diff
@@ -65,8 +67,11 @@ auto touch_center(MirTouchEvent const* event) -> mir::geometry::Point
 
 struct miral::MinimalWindowManager::Impl
 {
-    Impl(WindowManagerTools const& tools, MirInputEventModifier pointer_drag_modifier) :
-        tools{tools}, pointer_drag_modifier{pointer_drag_modifier} {}
+    Impl(WindowManagerTools const& tools,
+         MirInputEventModifier pointer_drag_modifier) :
+        tools{tools},
+        application_selector(tools),
+        pointer_drag_modifier{pointer_drag_modifier} {}
```
AlanGriffiths, line 74: Spurious change?

src/miral/minimal_window_manager.cpp
```diff
@@ -103,7 +111,7 @@ miral::MinimalWindowManager::MinimalWindowManager(WindowManagerTools const& tool
 
 miral::MinimalWindowManager::MinimalWindowManager(WindowManagerTools const& tools, MirInputEventModifier pointer_drag_modifier):
     tools{tools},
-    self{new Impl{tools, pointer_drag_modifier}}
+    self{new Impl{tools,pointer_drag_modifier}}
```
AlanGriffiths, line 114: Spurious change?

src/miral/minimal_window_manager.cpp
```diff
@@ -159,7 +167,14 @@ bool miral::MinimalWindowManager::handle_keyboard_event(MirKeyboardEvent const*
...
         case KEY_TAB:
-            tools.focus_next_application();
+            if (self->application_selector.is_active())
+            {
+                self->application_selector.next();
+            }
+            else
+            {
+                self->application_selector.start(false);
+            }
```
AlanGriffiths, line 177: Asking the `application_selector` for its state to decide which `application_selector` method to call just looks weird.

src/miral/minimal_window_manager.cpp
```diff
@@ -187,6 +209,19 @@ bool miral::MinimalWindowManager::handle_keyboard_event(MirKeyboardEvent const*
...
     }
 
+    if (action == mir_keyboard_action_up)
+    {
+        switch (mir_keyboard_event_scan_code(event))
+        {
+            case KEY_LEFTALT:
+                if (self->application_selector.is_active()) {
+                    self->application_selector.complete();
+                }
```
AlanGriffiths, line 219: Asking the `application_selector` for its state to decide whether to  call an `application_selector` method just looks weird. (Also braces layout)

AlanGriffiths (changes requested):
I've fixed up the symbols and done a pass first over the code. I still have to convince myself the functionality is actually what I think is expected, but  it looks promising.

AlanGriffiths:
> snap refresh miriway --channel edge/mir-pr2981

I've tried this and I think the visual feedback is lacking:

1. Arrange several applications so they do not overlap.
2. Hold Alt and press tab a few times - there is no feedback
3. Release Alt - focus visibly changes

I think we'd be better with actually changing focus at step 2 so that apps redecorate.

@Saviq opinion please?

Saviq:
> I think we'd be better with actually changing focus at step 2 so that apps redecorate.

Agreed, without the shell dealing with visual hints for which app will be left in front, this is the best. It will also reduce visual noise when letting go of Alt.

Another thing I think needs tweaking, is if I have (front→back) ABC, then switch to C by Alt+Tabbing twice, I end up with CBA visually, but CAB (as expected) focus-wise. I.e. A is raised to top on a single Alt+Tab, despite it being at the back, visually.

OTOH the only other option is to temporarily raise the to-be-focused window, with the previous one going back to its original spot in the stack - which may be a lot of movement on screen.

mattkae:
@AlanGriffiths @Saviq I agree that we should focus the applications instead of "raise"-ing them too. I will implement that this morning, but it might make for a little tricker of an API

AlanGriffiths:
> Out of curiosity, how difficult would it be to implement a visual system that does not involve focus?

We already provide decorations for X-based apps.

So, I don't think it terribly difficult, but would definitely take more time to agree and implement visuals. (One for the foreseeable future)

src/miral/minimal_window_manager.cpp
mattkae: Agreed, this felt a bit wonky. In reality, you can call `next()` in any context and it would be valid.

include/miral/miral/window_manager_tools.h
```diff
@@ -156,6 +156,10 @@ class WindowManagerTools
     /// make the prev surface active within the active application
     void focus_prev_within_application();
 
+    /// Returns true if the application is focusable, otherwise false.
+    /// \remark Since MirAL 3.10
+    auto can_focus_application(Application application) const -> bool;
```
AlanGriffiths, line 161: I'm not sure we need this function: it is only used in one place, and that place probably belongs in `WindowManagerTools`

tests/miral/application_selector.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    ApplicationSelectorTest()
+        : mt::TestWindowManagerTools(),
+          application_selector(window_manager_tools)
+    {}
+};
+
+/// Testing if we can cycle through 3 windows and then select one
+/// with a "forward" selection strategy.
+TEST_F(ApplicationSelectorTest, run_forward)
```
AlanGriffiths, line 37: I understand why you've written it this way, but tests that check multiple state transitions are both hard to name and have complicated failure modes that can be hard to interpret.

AlanGriffiths (changes requested):
More tweaking.

Also, I still think that after a double tab, that ABC should become CAB (not CBA)

include/miral/miral/minimal_window_manager.h
```diff
@@ -76,9 +76,15 @@ class MinimalWindowManager : public WindowManagementPolicy
...
-    /// Raises newly focused window
+    /// Raises newly focused window and notifies the application selector
     void advise_focus_gained(WindowInfo const& window_info) override;
 
+    /// Notifies the application selector about the new app
+    void advise_new_app(miral::ApplicationInfo &app_info) override;
+
+    /// Notifies the application selector about the deleted app
+    void advise_delete_app(miral::ApplicationInfo const &app_info) override;
+
```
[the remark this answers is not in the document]
mattkae: Agreed!

src/miral/application_selector.cpp
```diff
@@ -0,0 +1,171 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "application_selector.h"
+#include <miral/application_info.h>
+#include <mir/scene/session.h>
```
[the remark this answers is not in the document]
mattkae: Aha. That makes sense !

tests/miral/application_selector.cpp
mattkae: Agreed! I will make them more precise

AlanGriffiths (changes requested):
FYI I fixed up libliral5.symbols again

I still think (I've said it every review) that, after a double tab, that ABC should become CAB (not CBA)

(I mean that as both Z order and focus order)

The current iteration seems to be:

Z-order: ABC => CBA
Focus: ABC=> CAB

Which is neither what a user expects nor is it consistent with itself

As we've discussed, this is achievable by, on the Nth tab, swapping the position of the current focus/top with the Nth in the stack.

AlanGriffiths:
> Which is neither what a user expects nor is it consistent with itself

This inconsistency is only possible because the Z order and the tab order are maintained separately

mattkae:
> > Which is neither what a user expects nor is it consistent with itself
> 
> This inconsistency is only possible because the Z order and the tab order are maintained separately

This should do it @AlanGriffiths 

https://github.com/MirServer/mir/assets/25062299/c29a47c7-f1d1-4f64-8953-7d52d14a70df

AlanGriffiths:
@mattkae please check that it builds. You can fix the debian/libmiral5.symbols with (depending on your toolchain):

    ninja regenerate-miral-debian-symbols

mattkae:
> @mattkae please check that it builds. You can fix the debian/libmiral5.symbols with (depending on your toolchain):
> 
> ```
> ninja regenerate-miral-debian-symbols
> ```

Done!

debian/libmiral5.symbols
```diff
@@ -426,3 +426,10 @@ libmiral.so.5 libmiral5 #MINVER#
  (c++)"miral::WindowInfo::visible_on_lock_screen() const@MIRAL_3.9" 3.9.0
  (c++)"miral::WindowSpecification::visible_on_lock_screen() const@MIRAL_3.9" 3.9.0
  (c++)"miral::WindowSpecification::visible_on_lock_screen()@MIRAL_3.9" 3.9.0
+ MIRAL_3.10@MIRAL_3.10 3.10.0
+ (c++)"miral::MinimalWindowManager::advise_delete_app(miral::ApplicationInfo const&)@MIRAL_3.10" 3.10.0
+ (c++)"miral::MinimalWindowManager::advise_focus_lost(miral::WindowInfo const&)@MIRAL_3.10" 3.10.0
+ (c++)"miral::MinimalWindowManager::advise_new_app(miral::ApplicationInfo&)@MIRAL_3.10" 3.10.0
+ (c++)"miral::WindowManagerTools::can_select_application(std::shared_ptr<mir::scene::Session>, miral::Window&) const@MIRAL_3.10" 3.10.0
+ (c++)"miral::WindowManagerTools::can_select_application(std::shared_ptr<mir::scene::Session>, miral::Window&)@MIRAL_3.10" 3.10.0
+ (c++)"miral::WindowManagerTools::send_tree_to_back(miral::Window const&)@MIRAL_3.10" 3.10.0 (c++)"miral::WindowManagerTools::swap_tree_order(miral::Window const&, miral::Window const&)@MIRAL_3.10" 3.10.0
```
AlanGriffiths, line 435:
```suggestion
 (c++)"miral::WindowManagerTools::send_tree_to_back(miral::Window const&)@MIRAL_3.10" 3.10.0
 (c++)"miral::WindowManagerTools::swap_tree_order(miral::Window const&, miral::Window const&)@MIRAL_3.10" 3.10.0
```

AlanGriffiths (changes requested):
Just checking the functionality...

This isn't aware of workspaces.

E.g.

Running miral-app from a VT
1. open three apps (terminal, firefox, gedit)
2. Press `Meta-Ctrl-F2` to take gedit to the second workspace
3. Press tab

Expect: focus remains on gedit
Actual: Focus moves from gedit, and no amount of tabbing gets it back

mattkae:
> Just checking the functionality...
> 
> This isn't aware of workspaces.
> 
> E.g.
> 
> Running miral-app from a VT
> 
> 1. open three apps (terminal, firefox, gedit)
> 2. Press `Meta-Ctrl-F2` to take gedit to the second workspace
> 3. Press tab
> 
> Expect: focus remains on gedit Actual: Focus moves from gedit, and no amount of tabbing gets it back

Fixed! It was a bug :+1:

include/miral/miral/minimal_window_manager.h
```diff
@@ -79,6 +79,12 @@ class MinimalWindowManager : public WindowManagementPolicy
     /// Raises newly focused window
     void advise_focus_gained(WindowInfo const& window_info) override;
 
+    void advise_focus_lost(WindowInfo const& window_info) override;
+
+    void advise_new_app(miral::ApplicationInfo &app_info) override;
+
+    void advise_delete_app(miral::ApplicationInfo const &app_info) override;
```
AlanGriffiths, line 86:
```suggestion
    void advise_new_app(miral::ApplicationInfo& app_info) override;

    void advise_delete_app(miral::ApplicationInfo const& app_info) override;
```

src/miral/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ add_library(miral-internal STATIC
                                          join_client_threads.h
                                          window_info_defaults.h
     window_specification_internal.cpp    window_specification_internal.h
+    application_selector.cpp
```
AlanGriffiths, line 32: This belongs in `miral`, not `miral-internal` - we need to publish the associated entry points

src/miral/application_selector.cpp
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    return *it;
+}
+
+auto ApplicationSelector::complete() -> Application
+{
+    if (!is_active())
+    {
+        mir::log_warning("Cannot call ApplicationSelector::stop when the ApplicationSelector is not active.");
```
AlanGriffiths, line 185: The logged text is wrong: the function is called `complete()`. Also, this gets spammed to the log when launching apps and is probably not something logging and certainly not a warning the user would care about.

src/miral/basic_window_manager.cpp
```diff
@@ -881,6 +881,49 @@ void miral::BasicWindowManager::focus_prev_within_application()
...
+    miral::Window new_focus;
+    auto const prev_window = active_window();
+
+    // We return "false" in the enumeration when we set "can_focus" to true
+    // because we only stop looking when we return false from the enumerator.
+    mru_active_windows.enumerate([&](miral::Window& window)
+    {
+        // We don't want to select a different application
+        if (window.application() != application) {
+             return true;
```
AlanGriffiths, line 896: Braces

src/server/scene/surface_stack.cpp
```diff
@@ -425,6 +425,105 @@ void ms::SurfaceStack::raise(SurfaceSet const& ss)
...
+
+void ms::SurfaceStack::send_to_back(const mir::scene::SurfaceSet &ss)
+{
+    bool surfaces_reordered{false};
+    {
+        RecursiveWriteLock ul(guard);
+        for (auto& layer : surface_layers)
+        {
+            auto const old_layer = layer;
+
```
AlanGriffiths, line 497: Copying the whole container seems an expensive way to determine if changes were made

src/server/scene/surface_stack.cpp
```diff
@@ -425,6 +425,105 @@ void ms::SurfaceStack::raise(SurfaceSet const& ss)
...
+            auto to_send_to_back = std::vector<std::shared_ptr<Surface>>{split, layer.end()};
+
+            // Chop off the surfaces we are moving from the old vector (they are now only in to_send_to_back)
+            layer.erase(split, layer.end());
+
+            // One by one insert to_send_to_back surfaces at the front of the vector
+            for (auto const& surface : to_send_to_back)
+            {
+                layer.insert(layer.begin(), surface);
+            }
```
AlanGriffiths, line 513: This just looks weird!

1. use std::stable_partition() to move "stuff" to the end
2. copy "stuff"
3. erase "stuff" from the end
4. insert "stuff" at the beginning

Isn't that simply "use std::stable_partition() to move "stuff" to the beginning"?

tests/miral/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ mir_add_wrapped_executable(miral-test-internal NOINSTALL
     ignored_requests.cpp
     focus_mode.cpp
     fd_manager.cpp
+    application_selector.cpp
```
AlanGriffiths, line 53: Shouldn't be "internal" now we're publishing `ApplicationSelector`

AlanGriffiths (changes requested):
Some weird looking logic and some nitpicking

tests/miral/CMakeLists.txt
mattkae: As a note, `application_selector.cpp` relies on `test_window_manager_tools.h`. `test_window_Manager_tools.h` relies on `basic_window_manager.h`, which is internal to mir.

As a result, I will have to be moving `test_window_manager.h` to `/sr/tests/miral` (unless there exists a better option, but the dependency seems very important)

AlanGriffiths: Leave it then

mattkae: Okay

src/server/scene/surface_stack.cpp
mattkae: Wow I got to in the weeds here! Much easier :+1:

src/server/scene/surface_stack.cpp
mattkae: Yeah I agree. This is how we were doing it in `SurfaceStack::raise` as well. Would there be an easier way of checking if `std::stable_partition` resulted in a change (other than rolling a custom std::stable_partition)?

RAOF:
> > I think we'd be better with actually changing focus at step 2 so that apps redecorate.
> 
> Agreed, without the shell dealing with visual hints for which app will be left in front, this is the best. It will also reduce visual noise when letting go of Alt.

I'm slightly concerned about having transient client-visible changes during alt-tab; some clients *do* things on focus change (a particular example I can think of is some fullscreen games will minimise their window when loosing focus).

A quick survey of other compositors' behaviour:
* GNOME (and I *think* KDE Plasma, although the default alt-tab there *only* shows the currently selected window) remove focus from all windows when entering alt-tab, and doesn't change focus during alt-tab
* Weston doesn't change focus at all (the currently focused window retains focus througout alt-tab)
* Sway doesn't have alt-tab (?!)

So switching focus during alt-tab is not a common behaviour, but alt-tab temporarily changing focus *is* a common behaviour, so it's *probably* fine?

src/server/scene/surface_stack.cpp
```diff
@@ -425,61 +425,75 @@ void ms::SurfaceStack::raise(SurfaceSet const& ss)
...
+            std::stable_sort(swap_begin, swap_end, [to_front, to_back](std::weak_ptr<Surface> s1, std::weak_ptr<Surface> s2)
             {
-                auto found_first = std::find(first_move.begin(), first_move.end(), *it);
-                auto found_second = std::find(second_move.begin(), second_move.end(), *it);
-                if (found_first != first_move.end() || found_second != second_move.end())
-                     layer.erase(it);
+                if (to_front.count(s1))
+                {
+                    return -1;
+                }
```
AlanGriffiths, line 479: Note that the result of the compare functor is contextually converted to bool: returning `-1`, `1` and `0` is confusing.

Also, for a strict weak ordering, if both `s1` and `s2` are in the same equivalence class (e.g. `to_front.count(s1)` and `to_front.count(s2)`) then the result should be `false`.

AlanGriffiths (changes requested):
Dropping the dubious pointer arithmetic seems to have fixed the SIGSEG, but there's still dubious logic and spurious copies about.

AlanGriffiths (changes requested):
some quick nits

Saviq (approved):
This seems fine to me from a UX perspective!

:tada:

src/server/scene/surface_stack.cpp
```diff
@@ -425,6 +425,108 @@ void ms::SurfaceStack::raise(SurfaceSet const& ss)
...
+                        first_to_front = true;
+                        swap_begin = it;
+                    }
+
+                    num_second_found++;
+                }
+            }
+
+            // Finally, move the to_front items to the front of the group and the to_back to the back of the group
+            auto to_front = first_to_front ? first : second;
```
AlanGriffiths, line 472:
```suggestion
            auto const& to_front = first_to_front ? first : second;
```

src/server/scene/surface_stack.cpp
```diff
@@ -425,6 +425,108 @@ void ms::SurfaceStack::raise(SurfaceSet const& ss)
...
+            auto const& to_back = first_to_front ? second : first;
+            std::stable_sort(swap_begin, swap_end, [&](std::weak_ptr<Surface> const& s1, std::weak_ptr<Surface> const& s2)
+            {
+                if (to_front.count(s1))
+                    return to_front.count(s2) == 0;
+                else if (to_front.count(s2))
+                    return false;
+                else if (to_back.count(s1))
+                    return to_back.count(s2) == 0;
```
AlanGriffiths, line 481: Something is wrong either in the code or in my understanding: If `s1` is in the "to back" partition we should never return true (because `s1` is never strictly before `s2`). But `to_back.count(s2) == 0` can be true.

AlanGriffiths (changes requested):
Some tests of `SurfaceStack::swap_z_order() and `SurfaceStack::send_to_back()` would clarify whether they are doing what is expected of them

AlanGriffiths (changes requested):
> Some tests of `SurfaceStack::swap_z_order() and `SurfaceStack::send_to_back()` would clarify whether they are doing what is expected of them

I've fixed what looked like a bug, but we really needs tests for the above non-trivial functions.

With that the functionality and UX seems good to me.

I've left a bunch of comments for "polishing" the implementation.

----

I think we need another opinion about the remaining API design issues (ping @Saviq, @RAOF): 

1. how we want to evolve `ApplicationSelector`:

* Can it be more useful to downstreams with some tweaking?
* Should it be a pluggable policy for `MinimalWindowManager` that can be overridden?
* Is it so specific to this usage it should be rolled into the `MinimalWindowManager` implementation?

2. How we want to treat the changed vtab entries in `MinimalWindowManager` (we already discussed that with Saviq, but there's been time to think since)

src/miral/application_selector.cpp
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <miral/application_selector.h>
+#include <miral/application_info.h>
+#include <miral/application.h>
+#include <mir/log.h>
+#include <iostream>
```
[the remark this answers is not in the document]
AlanGriffiths: That was meant to just highlight  iostream

src/server/scene/surface_stack.cpp
mattkae: From what I understand, if both s1 and s2 are in `to_back`, then they are in the the correct order (so we return true). This will maintain. The relative order of s1 and s2. If s2 is not in `to_back`, s1 and s2 should be swapped (so we return false)

include/miral/miral/window_manager_tools.h
```diff
@@ -156,6 +156,10 @@ class WindowManagerTools
     /// make the prev surface active within the active application
     void focus_prev_within_application();
 
+    /// Returns true if the application can be selected, otherwise false.
+    /// \remark Since MirAL 3.10
+    auto can_select_application(const Application, Window&) const -> bool;
```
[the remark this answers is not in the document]
mattkae: Agreed!

include/miral/miral/application_selector.h
```diff
@@ -0,0 +1,92 @@
...
+    void advise_focus_lost(WindowInfo const&);
+
+    /// Called when an application is deleted.
+    void advise_delete_app(Application const&);
+
+    /// Raises the next selectable application in the list for focus selection.
+    /// \param reverse If true, the selector will move in reverse through the list of applications
+    /// otherwise it will move forward through the list of applications.
+    /// \returns The raised application, or nullptr in the case of a failure
+    auto next(bool reverse) -> Application;
```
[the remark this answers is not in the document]
mattkae: Agreed :+1:

Saviq:
> I think we need another opinion about the remaining API design issues (ping @Saviq, @RAOF):
> 
>     1. how we want to evolve `ApplicationSelector`:
>     * Is it so specific to this usage it should be rolled into the `MinimalWindowManager` implementation?

It feels to me like it should be folded into MinimalWindowManager; I don't think there's much room for changing its behaviour — Alt+` would be an extension, and cross-workspace handling I don't think can happen without a visual switcher.

In reality downstreams are likely to replace it with a visual switcher that would just raise the selected window on completion.

>     2. How we want to treat the changed vtab entries in `MinimalWindowManager` (we already discussed that with Saviq, but there's been time to think since)

I vote to be a good citizen and bump SONAME.

include/miral/miral/application_selector.h
```diff
@@ -0,0 +1,92 @@
...
+    /// Represents the current order of focus by application. Most recently focused
+    /// applications are at the beginning, while least recently focused are at the end.
+    std::vector<Application> focus_list;
+
+    /// The application that was selected when next was first called
+    Application originally_selected;
+
+    /// The application that is currently selected.
+    Application selected;
+    Window active_window;
```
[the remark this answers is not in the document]
mattkae: Agreed!

AlanGriffiths:
> > 3\. How we want to treat the changed vtab entries in `MinimalWindowManager` (we already discussed that with Saviq, but there's been time to think since)
> 
> I vote to be a good citizen and bump SONAME.

We could also put the logic into a new `class MinimalWindowManagerImprovedAltTab: public `MinimalWindowManager ...` and not break ABI

src/miral/basic_window_manager.h
```diff
@@ -320,6 +324,9 @@ class BasicWindowManager : public virtual mir::shell::WindowManager,
     void advise_output_end() override;
     /// Updates the application zones of all display areas and moves attached windows as needed
     void update_application_zones_and_attached_windows();
+
+    /// Collects all children windows provided a WindowInfo
+    auto collect_windows(WindowInfo const& info) -> std::vector<Window>;
```
AlanGriffiths, line 329: Everywhere it is used the vector returned by `collect_windows()` is copied into a `SurfaceSet`, why doesn't it return a `SurfaceSet` and avoid all that copying?

src/miral/symbols.map
```diff
@@ -493,3 +493,28 @@ global:
...
+    miral::ApplicationSelector::advise_new_app*;
+    miral::ApplicationSelector::advise_focus_gained*;
+    miral::ApplicationSelector::advise_focus_lost*;
+    miral::ApplicationSelector::advise_delete_app*;
+    miral::ApplicationSelector::next*;
+    miral::ApplicationSelector::prev*;
+    miral::ApplicationSelector::complete*;
+    miral::ApplicationSelector::cancel*;
+    miral::ApplicationSelector::is_active*;
+    miral::ApplicationSelector::get_focused*;
```
AlanGriffiths, line 518: These should be gone

src/miral/window_management_trace.cpp
```diff
@@ -576,6 +576,15 @@ try {
 }
 MIRAL_TRACE_EXCEPTION
 
+auto miral::WindowManagementTrace::window_to_select_application(const Application app) const -> std::optional<Window>
```
AlanGriffiths, line 579:
```suggestion
auto miral::WindowManagementTrace::window_to_select_application(Application const app) const -> std::optional<Window>
```

src/miral/window_management_trace.cpp
```diff
@@ -576,6 +576,15 @@ try {
...
+try {
+    log_input();
+    mir::log_info("%s", __func__);
```
AlanGriffiths, line 582:
```suggestion
    mir::log_info("%s app=%s", __func__, dump_of(app).c_str());
```

src/server/scene/surface_stack.cpp
```diff
@@ -425,6 +425,110 @@ void ms::SurfaceStack::raise(SurfaceSet const& ss)
...
+                    break;
+                }
+
+                // Find the start position and count how many we've found
+                bool in_first = first.count(*it) > 0;
+                bool in_second = second.contains(*it) > 0;
+                if (in_first && in_second)
+                {
+                    BOOST_THROW_EXCEPTION(std::logic_error("Swapped sets must not have any overlap"));
+                }
```
AlanGriffiths, line 457: Mixing precondition checks with logic makes for hard to read code.

If we need to validate preconditions (which is debatable here), then it is best at the start of the function:

```c++
    for (auto const& s : first)
    {
        if (second.count(s))
            fatal_error("Swapped sets must not have any overlap");
    }
```

And, if anywhere, this check belongs in `BasicWindowManager` as that is where we can first detect a problem. (In that context, it is clear that `first` and `second` are the same `SurfaceSet`, so a no-op would also be sensible behaviour)

tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -1374,3 +1374,99 @@ TEST_F(SurfaceStack, when_screen_is_locked_surface_ignores_surface)
     EXPECT_THAT(stack.surface_at(cursor_position).get(), IsNull());
     handle->allow_to_be_dropped();
 }
+
+TEST_F(SurfaceStack, surfaces_tht_are_sent_to_back_appear_at_the_front_of_the_list)
```
AlanGriffiths, line 1378:
```suggestion
TEST_F(SurfaceStack, surfaces_that_are_sent_to_back_appear_at_the_front_of_the_list)
```

tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -1374,3 +1374,99 @@ TEST_F(SurfaceStack, when_screen_is_locked_surface_ignores_surface)
...
+TEST_F(SurfaceStack, swapping_overlapping_surface_sets_causes_an_exception)
+{
+    stack.add_surface(stub_surface1, mi::InputReceptionMode::normal);
+    stack.add_surface(stub_surface2, mi::InputReceptionMode::normal);
+    stack.add_surface(stub_surface3, mi::InputReceptionMode::normal);
+
+    EXPECT_THAT([&]() { stack.swap_z_order({stub_surface1, stub_surface2}, {stub_surface2, stub_surface3}); },
+                Throws<std::logic_error>(Property(&std::logic_error::what,
+                                                    HasSubstr("Swapped sets must not have any overlap"))));
+}
```
AlanGriffiths, line 1472: Throwing exceptions is not normally part of the contract. Particularly for precondition violations. (Preconditions are the responsibility of caller not the callee.)

Exceptions should be used to report that the post-condition cannot be met.

So, I wouldn't have this test because, assuming the precondition is met, then it doesn't happen.

But if I did, then I'd also want this to pass:

```c++
TEST_F(SurfaceStack, swapping_overlapping_surface_sets_causes_an_exception2)
{
    stack.add_surface(stub_surface1, mi::InputReceptionMode::normal);
    stack.add_surface(stub_surface3, mi::InputReceptionMode::normal);

    EXPECT_THAT([&]() { stack.swap_z_order({stub_surface1, stub_surface2}, {stub_surface2, stub_surface3}); },
                Throws<std::logic_error>(Property(&std::logic_error::what,
                                                  HasSubstr("Swapped sets must not have any overlap"))));
}
```

AlanGriffiths (changes requested):
Mostly tidy-up after earlier iterations

AlanGriffiths:
Oh, and a behavioural problem this introduces:

Run gimp (only) with "single window mode" unchecked (so there are a couple of satellite windows)
Press Alt-Tab

**Expect:** nothing happens
**Actual:** the satellite windows move southeast with each tab

This doesn't happen with Miriway/edge or with GNOME

tests/unit-tests/scene/test_surface_stack.cpp
mattkae: I have moved the check to BasicWindowManager instead :+1:

AlanGriffiths (commented):
Nitpicking

AlanGriffiths (changes requested):

AlanGriffiths (approved):
Looks OK to me.

But I've looked at this too many times: I'll let fresh eyes take a look before we land it.

AlanGriffiths (approved):
Still looks OK after a break of a few days. Let's do it!!

PR #2983, opened 2023-07-19
Fix FTBFS we see on Debian/sid and Ubuntu/Mantic

AlanGriffiths:


debian/rules
```diff
@@ -69,8 +69,8 @@ ifeq ($(USER) $(DEB_HOST_ARCH),buildd riscv64)
   endif
 endif
 
-# Disable pre-compiled headers on GCC-12
-ifneq ($(shell gcc --version | grep '12.[[:digit:]]\+.[[:digit:]]\+$$'),)
+# Disable pre-compiled headers on GCC-12 & 13
+ifneq ($(shell gcc --version | grep -e '12.[[:digit:]]\+.[[:digit:]]\+$$' -e '13.[[:digit:]]\+.[[:digit:]]\+$$'),)
```
Saviq, line 73: How about simpler?

```suggestion
ifneq ($(shell gcc --version | grep '1[23].[[:digit:]]\+.[[:digit:]]\+$$'),)
```
or 
```suggestion
ifneq ($(shell gcc --version | grep '\(12\|13\).[[:digit:]]\+.[[:digit:]]\+$$'),)
```

Saviq (approved):
Good!

Issue #2985, opened 2023-07-21
Keyboard meta key state gets out of step


AlanGriffiths:
Originally reported [on discourse](https://discourse.ubuntu.com/t/mir-release-2-14-0/37082/11):

> Can’t write text on Firefox when switching window
> 
> 1. Open Firefox (not full screen, just windowed)
> 2. Open Konsole or Chromium (not full screen)
> 3. Focus on a text box of Firefox (e.g. the discourse message compositor I’m using for this message) and try to write. It works move to Konsole with ALT+TAB
> 4. go back to Firefox (mouse-click on Firefox window, do not use Alt+Tab again)
> 5. press any letter on the keyboard, you will see the Firefox menu appears. It behaves like if you are keeping the ALT key pressed.
>
> I can “release” the fake ALT key pressing the Meta/Super/Windows key on the keyboard and I’m be able to write again.

mattkae:
## Reproducing
- Firefox opens the "Alt" menu upon the alt key going "up"
- The bug no longer happens when we remove the Alt + Tab behavior from `MinimalWindowManager`, which leads me to believe that this is a focus + input issue.

mattkae:
## Researching
- I have confirmed that when the compositor handles an event, it does not allow the `wl_seat` to take that event
- I have confirmed that the Alt + Up event is appropriately sent to the tabbed-to process instead of firefox

mattkae:
## Conclusion (so far)
- My best guess at the moment is that Firefox continues to think that alt is held down internally as it has received no information to cancel that state

AlanGriffiths:
>     * My best guess at the moment is that Firefox continues to think that alt is held down internally as it has received no information to cancel that state

From the discussion above (and experience of seeing this but being unable to reproduce) I think this is handled correctly when focus changes using the keyboard. But not when using the mouse: I suspect there is code to resync the keyboard state on one path, but not the other. (Try looking in the Wayland frontend)

mattkae:
Another way to experience this bug:
1. Open firefox and Konsole
2. Type some stuff in the address bar
3. Hold alt and then click on Konsole
4. Release alt once Konsole is focused
5. Click back to Firefox and note that you are still stuck in the same bugged state

I am not so certain that it is handled correctly when we use the keyboard either. That _might_ be a fluke as we end up sending firefox an "alt modifier released" event when we inevitable alt + tab back to it. But I will investigate further

mattkae:
[...]

PR #2988, opened 2023-07-24
(#2985) Fixed modifiers not being sent when a window is focused

mattkae:
# https://github.com/MirServer/mir/issues/2985

## What was the problem?
- We wanted to refresh the modifiers when an application gained keyboard focus, so we were calling `helpers->refresh_modifiers` which - eventually - was calling `send_modifiers`.
- However, `focused_surface` ended up being `nullptr` at this point, so we never got the message on the client side

## Solution
- Call `refresh_modifiers` after we assign the pointer

### DIFF
--- a/src/server/frontend_wayland/wl_keyboard.cpp
+++ b/src/server/frontend_wayland/wl_keyboard.cpp
@@ -21,7 +21,6 @@
 #include "mir/log.h"
 #include "mir/events/keyboard_event.h"
 #include "mir/wayland/client.h"
-#include <iostream>
 
 #include <xkbcommon/xkbcommon.h>
 #include <cstring> // memcpy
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
In the interests of understanding why the code is the way it is, this logic dates from #2537 which "fixed" #2535 and #2025. (The latter sounds remarkably similar to #2985!)

Looking that leads to the question: why is there a test for `focused_surface` in `WlKeyboard::send_modifiers()`? That seems unnecessary at best and, according to the above analysis, wrong.

I think removing that check is the simpler fix: #2989

AlanGriffiths:
I've tried the reported scenario with both this and my rework. Neither appears to fix #2985.

mattkae:
Closed in favor of: https://github.com/MirServer/mir/pull/2989

PR #2989, opened 2023-07-25
 Send modifiers after keyboard enter

AlanGriffiths:
We need to re-sync modifier state after `wl_keyboard.enter()`, so we need `WlKeyboard::send_modifiers()` to send the modifiers unconditionally.

Fixes: #2985

Note that this condition originates in #2537 which "fixed" #2535 and #2025.

Alternative to #2988


mattkae (approved):
This event is not dependent on the `focused_surface`. Looks like a good change :1st_place_medal:

PR #2990, opened 2023-07-25
ci: don't fail-fast on snaps

Saviq:
Ref.:

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures


AlanGriffiths (approved):
Yes

PR #2991, opened 2023-07-25
kinetic (a.k.a. 22.10) is obsolete

AlanGriffiths:


Saviq (changes requested):
There's also this:

https://github.com/MirServer/mir/blob/1d825e909814c156241de746b2f4d8bce88fc39b/.github/workflows/spread.yml#L45

Also, one of the commit messages looks alien.

Saviq (approved):
Yup, thanks!

Will you do wlcs, too?

PR #2997, opened 2023-08-02
Allow evdi

AlanGriffiths:
Empirically, the platform API branch works with evdi (modulo some display config tracking that should be fixed).

So this is a reminder to change the defaults


AlanGriffiths:
> **decrease** coverage by `0.01%`

How can deleting a comment in uncovered code **decrease** the coverage?

RAOF (approved):
There's no reason not to actually merge this in now.

PR #2999, opened 2023-08-08
Fix MultiplexingDisplay names

RAOF:
This ensures that output names in a `MultiplexingDisplay` configuration match the existing names.

### DIFF
--- a/src/server/graphics/multiplexing_display.cpp
+++ b/src/server/graphics/multiplexing_display.cpp
@@ -184,10 +184,19 @@ private:
 
         static auto name_for_output(
             mg::DisplayConfigurationOutput const& output,
-            std::map<mg::DisplayConfigurationOutputType, unsigned>& output_counts) -> std::string
+            std::map<mg::DisplayConfigurationCardId, std::map<mg::DisplayConfigurationOutputType, unsigned>>& output_counts,
+            bool multiple_cards) -> std::string
         {
-            output_counts[output.type]++;    // The map will default-initialise any as-yet-unseen types to 0
-            return std::string{mir::output_type_name(static_cast<unsigned>(output.type))} + "-" + std::to_string(output_counts[output.type]);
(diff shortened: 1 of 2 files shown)


Saviq:
Not sure what's with GH, but here's the error that makes things red:

```plain
 ../../../../src/server/graphics/multiplexing_display.cpp:192:31: error: aggregate ‘std::stringstream name’ has incomplete type and cannot be defined
  192 |             std::stringstream name;
      |                               ^~~~
```

src/server/graphics/multiplexing_display.cpp
```diff
@@ -181,25 +184,47 @@ class CompositeDisplayConfiguration : public mg::DisplayConfiguration
...
+                                                             // and any as-yet-unseen output types to 0.
+            std::stringstream name;
+            name << mir::output_type_name(static_cast<unsigned>(output.type));
+            if (multiple_cards)
+            {
+                name << "-" << output.card_id.as_value() + 1;
+            }
+            name << "-" << output_counts[output.card_id][output.type];
+            return name.str();
+        }
```
AlanGriffiths, line 202: This doesn't quite match the behaviour of the existing `name_outputs()` implementations. I think it would avoid problems to match the current behaviour (and would be simple to implement).

The existing behaviour:

Card 0: names are not decorated with a card number
Card n, [n>0] : names are decorated with n

This PR:

Card 0: names are either not decorated with a card number or, if there are multiple cards, decorated with 1
Card n, [n>0] : names are decorated with n+1

I don't see the motivation for making names on card 0 change when another card is added, nor for changing the names on subsequent cards.

Saviq: Agreed, let's try and stick to the names we have.

AlanGriffiths (approved):
With my amendment I'm happy with this. But someone else should check before landing.

Saviq (approved):
J'approve of your changes, @AlanGriffiths.

RAOF:
Huh, whoops. I misread the existing algorithm! Ta.

Issue #2777, opened 2023-01-10
Support for input-method-unstable-v1


dobey:
Currently, only `input-method-unstable-v1` is in the upstream wayland-protocols, and `input-method-unstable-v2`  which Purism seemingly created for Phosh, is unfortunately incompatible. As [maliit](https://github.com/maliit/framework) implements v1 which is used by kwin and Qt compositor, it would be good to have this in Mir as well, so that maliit based keyboards can work under Mir based compositors such as Lomiri, using the Wayland protocols instead of the old input context plugins.

Thanks.

wmww:
Seems like a good idea to me

mattkae:
Hi @dobey . I am working on supporting this ticket in Mir right now, and I was wondering if you had a wayland environment in which this currently functions? I tried out weston and gnome to no avail. (gnome simply doesn't support input-method-v1, but weston claims to, but segfaults when it gets there)

dobey:
Hi @mattkae 

It works under kwin_wayland. Make sure `GTK_IM_MODULE` and `QT_IM_MODULE` are unset in the shell you're running from, and with `maliit-keyboard` and `kwin-wayland` installed, running `KWIN_IM_SHOW_ALWAYS=1 kwin_wayland --width 1280 --height 720 --inputmethod /usr/bin/maliit-keyboard gedit`  should open a window in which gedit is focused and with the maliit keyboard showing and able to type in gedit with the OSK.

This is how I often test locally when working on the OSK code, and it works normally well for me on Debian testing.

Saviq:
Hi @dobey o/!

@mattkae I was able to get it to work in a 23.10 VM on the console like Rodney pointed out:

```plain
sudo apt-get install kwin-wayland maliit-keyboard qtwayland5 gedit --no-install-recommends
KWIN_IM_SHOW_ALWAYS=1 kwin_wayland --inputmethod /usr/bin/maliit-keyboard gedit
```

I didn't manage to get things to nest on a GNOME desktop, though.

mattkae:
> Hi @mattkae
> 
[...]

PR #3000, opened 2023-08-08
(#2777) Implementation of input-method-v1 for maliit-keyboard support

mattkae:
# #2777 input-method-v1
- **Tests**: https://github.com/MirServer/wlcs/pull/302
- **Protocol**: https://wayland.app/protocols/input-method-unstable-v1

## What's new?
- Implemented `InputMethodV1` which allows the input method to modify the state of a text input, most likely a text input v1 or v2, as v3 has a considerably newer API
- Implemented `InputPanelV1` which allows a user to specify a surface to open at a particular position
- Moved shared information between input method v1 and v2 to an "input_method_common" file
- Fixed a bug in `wl_surface.cpp` where sometimes we would get an `input_shape` _before_ we get a new buffer size, which was leading to the new buffer size never being set
- Implemented show/hide input panel for text input v1 and text input v2
- **NOTE**: I have not seen an instance where `set_overlay_panel` is used, even when using a pinyin keyboard (which I would expect to be the typical use case, but perhaps there is another that I am not thinking of). As such, it is unimplemented until I have a solid demo case for it

## Demo
[Screencast from 08-11-2023 10:48:05 AM.webm](https://github.com/MirServer/mir/assets/25062299/1151fa74-e907-4dbe-8e67-1f291ecdb74a)

## Demo in Miriway
[Screencast from 08-11-2023 04:08:03 PM.webm](https://github.com/MirServer/mir/assets/25062299/0662e1f5-458b-42cb-9873-82dd6be0056f)

### DIFF
--- a/src/server/frontend_wayland/input_method_v1.cpp
+++ b/src/server/frontend_wayland/input_method_v1.cpp
@@ -197,7 +197,7 @@ private:
         {
             none,
             commit_string,
-            preedit_string
+            preedit_string,
         };
 
         struct InputMethodV1Change
@@ -328,6 +328,7 @@ private:
         void modifiers_map(struct wl_array *map) override
         {
             change.pending_change.modifier_map = map;
+            change.waiting_status = InputMethodV1ChangeWaitingStatus::none;
         }
 
         void keysym(uint32_t serial, uint32_t time, uint32_t sym, uint32_t state, uint32_t modifiers) override


src/include/server/mir/scene/text_input_hub.h
```diff
@@ -158,8 +182,14 @@ class TextInputHub : public ObserverRegistrar<TextInputStateObserver>
     /// NOTE: a handler may be kept alive and changes may be sent to the handler after this returns.
     virtual void deactivate_handler(std::shared_ptr<TextInputChangeHandler> const& handler) = 0;
 
-    /// Used by the input method to dispatch entered text to the handler
+    /// Used by commit-statethe input method to dispatch entered text to the handler
```
AlanGriffiths, line 185: "commit-statethe"?

src/wayland/symbols.map
```diff
@@ -610,5 +610,19 @@ global:
...
+    vtable?for?mir::wayland::InputMethodV1;
+    typeinfo?for?mir::wayland::InputMethodV1::Global;
+    vtable?for?mir::wayland::InputMethodV1::Global;
+
+    mir::wayland::InputPanelV1::*;
+    non-virtual?thunk?to?mir::wayland::InputPanelV1::*;
+    typeinfo?for?mir::wayland::InputPanelV1;
+    vtable?for?mir::wayland::InputPanelV1;
+    typeinfo?for?mir::wayland::InputPanelV1::Global;
+    vtable?for?mir::wayland::InputPanelV1::Global;
```
AlanGriffiths, line 626: These were not introduced in Mir 2.14: they belong in a new 2.15 stanza

src/server/frontend_wayland/input_method_v1.cpp
```diff
@@ -0,0 +1,583 @@
...
+#include "mir/scene/text_input_hub.h"
+#include "mir/shell/surface_specification.h"
+#include "mir/shell/shell.h"
+#include "mir/scene/session.h"
+#include "mir/log.h"
+#include "wl_surface.h"
+#include "output_manager.h"
+#include "window_wl_surface_role.h"
+#include "input_method_common.h"
+#include "input-method-unstable-v1_wrapper.cpp" // TODO: Super temporary. Can't figure out why link is broken.
```
AlanGriffiths, line 27: You're missing some symbols from wayland/symbols.map

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -397,7 +397,7 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
             stream->submit_buffer(mir_buffer);
             auto const new_buffer_size = stream->stream_size();
 
-            if (!input_shape && std::make_optional(new_buffer_size) != buffer_size_)
```
AlanGriffiths, line 400: This looks wrong: if the client has set an input shape, then why would we recalculate it?

mattkae: In my case, I was getting an `input_shape` first and _then_ a `new_buffer_size` afterward. Because pointer events are dependent on the intersection between these two, it made sense to me that we update whenever the `new_buffer_size` comes through

AlanGriffiths:
The Clang build is failing ("undefined reference to" looks like some symbols are not being exported):
```
2023-08-16T13:39:54.1448352Z /usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x23
2023-08-16T13:39:54.1449576Z frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/input_method_v1.cpp.o:(.data.rel.ro._ZTCN3mir8frontend13InputMethodV18InstanceE0_NS_7wayland13InputMethodV1E[_ZTCN3mir8frontend13InputMethodV18InstanceE0_NS_7wayland13InputMethodV1E]+0x40): undefined reference to `virtual thunk to mir::wayland::InputMethodV1::~InputMethodV1()'
2023-08-16T13:39:54.1450691Z /usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/input_method_v1.cpp.o:(.data.rel.ro._ZTCN3mir8frontend13InputMethodV18InstanceE0_NS_7wayland13InputMethodV1E[_ZTCN3mir8frontend13InputMethodV18InstanceE0_NS_7wayland13InputMethodV1E]+0x48): undefined reference to `virtual thunk to mir::wayland::InputMethodV1::~InputMethodV1()'
2023-08-16T13:39:54.1452229Z /usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/input_method_v1.cpp.o:(.data.rel.ro._ZTCN3mir8frontend13InputMethodV18Instance20InputMethodContextV1E0_NS_7wayland20InputMethodContextV1E[_ZTCN3mir8frontend13InputMethodV18Instance20InputMethodContextV1E0_NS_7wayland20InputMethodContextV1E]+0xa8): undefined reference to `virtual thunk to mir::wayland::InputMethodContextV1::~InputMethodContextV1()'
2023-08-16T13:39:54.1453756Z /usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/input_method_v1.cpp.o:(.data.rel.ro._ZTCN3mir8frontend13InputMethodV18Instance20InputMethodContextV1E0_NS_7wayland20InputMethodContextV1E[_ZTCN3mir8frontend13InputMethodV18Instance20InputMethodContextV1E0_NS_7wayland20InputMethodContextV1E]+0xb0): undefined reference to `virtual thunk to mir::wayland::InputMethodContextV1::~InputMethodContextV1()'
2023-08-16T13:39:54.1454854Z /usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/input_method_v1.cpp.o:(.data.rel.ro._ZTCN3mir8frontend12InputPanelV18InstanceE0_NS_7wayland12InputPanelV1E[_ZTCN3mir8frontend12InputPanelV18InstanceE0_NS_7wayland12InputPanelV1E]+0x48): undefined reference to `virtual thunk to mir::wayland::InputPanelV1::~InputPanelV1()'
2023-08-16T13:39:54.1455944Z /usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/input_method_v1.cpp.o:(.data.rel.ro._ZTCN3mir8frontend12InputPanelV18InstanceE0_NS_7wayland12InputPanelV1E[_ZTCN3mir8frontend12InputPanelV18InstanceE0_NS_7wayland12InputPanelV1E]+0x50): undefined reference to `virtual thunk to mir::wayland::InputPanelV1::~InputPanelV1()'
2023-08-16T13:39:54.1457365Z /usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/input_method_v1.cpp.o:(.data.rel.ro._ZTCN3mir8frontend12InputPanelV18Instance19InputPanelSurfaceV1E0_NS_7wayland19InputPanelSurfaceV1E[_ZTCN3mir8frontend12InputPanelV18Instance19InputPanelSurfaceV1E0_NS_7wayland19InputPanelSurfaceV1E]+0x50): undefined reference to `virtual thunk to mir::wayland::InputPanelSurfaceV1::~InputPanelSurfaceV1()'
2023-08-16T13:39:54.1458784Z /usr/bin/ld: frontend_wayland/CMakeFiles/mirfrontend-wayland.dir/input_method_v1.cpp.o:(.data.rel.ro._ZTCN3mir8frontend12InputPanelV18Instance19InputPanelSurfaceV1E0_NS_7wayland19InputPanelSurfaceV1E[_ZTCN3mir8frontend12InputPanelV18Instance19InputPanelSurfaceV1E0_NS_7wayland19InputPanelSurfaceV1E]+0x58): undefined reference to `virtual thunk to mir::wayland::InputPanelSurfaceV1::~InputPanelSurfaceV1()'
```

AlanGriffiths (changes requested):
A few nits, but mostly I'd appreciate confirmation of a configuration for experimenting with this.

I'm envisaging something like:

```
snap install miriway --channel edge/mir-pr3000
apt install maliit-keyboard
```
Add this to miriway.config
```
shell-add-wayland-extension zwp_input_method_v1
shell-add-wayland-extension zwp_input_panel_v1
shell-component mallit-keyboard
```
And run `miriway`

AlanGriffiths:
> I'm envisaging something like:
> 
> ```
> snap install miriway --channel edge/mir-pr3000
> apt install maliit-keyboard
> ```
> 
> Add this to miriway.config
> 
> ```
> shell-add-wayland-extension zwp_input_method_v1
> shell-add-wayland-extension zwp_input_panel_v1
> shell-component mallit-keyboard
> ```
> 
> And run `miriway`

I tried this and the `miriway-shell.config` seems to be:

```
shell-add-wayland-extension=zwp_input_method_v1
shell-add-wayland-extension=zwp_input_panel_v1
shell-component=miriway-unsnap maliit-keyboard
```

After which, I can see `maliit-keyboard` is running, but it doesn't activate when focus is on a text input (e.g. the terminal, or this input box in Firefox)

mattkae:
> > I'm envisaging something like:
> > ```
> > snap install miriway --channel edge/mir-pr3000
> > apt install maliit-keyboard
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Add this to miriway.config
> > ```
> > shell-add-wayland-extension zwp_input_method_v1
> > shell-add-wayland-extension zwp_input_panel_v1
> > shell-component mallit-keyboard
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And run `miriway`
> 
> I tried this and the `miriway-shell.config` seems to be:
> 
> ```
> shell-add-wayland-extension=zwp_input_method_v1
> shell-add-wayland-extension=zwp_input_panel_v1
> shell-component=miriway-unsnap maliit-keyboard
> ```
> 
> After which, I can see `maliit-keyboard` is running, but it doesn't activate when focus is on a text input (e.g. the terminal, or this input box in Firefox)

I had just taken the default `miriway-shell.config` and added:
```
add-wayland-extensions=zwp_input_method_v1:zwp_input_panel_v1
```

Then I logged in and ran `maliit-keyboard` from the terminal within the shell. For some reason, my miriway doesn't like it when I specify a "shell-component" in the config. Is that the way to set a startup application?

Here is the result in my case (running in VM):
image

src/server/frontend_wayland/input_method_v1.cpp
```diff
@@ -0,0 +1,569 @@
...
+
+        void language(uint32_t /*serial*/, const std::string &/*language*/) override
+        {
+            // TODO
+        }
+
+        void text_direction(uint32_t /*serial*/, uint32_t /*direction*/) override
+        {
+            // TODO
+        }
```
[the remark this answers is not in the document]
mattkae: 1. I implemented text direction
2. I left a good comment for language
3. I did not implement `grab_keyboard`, `key`, and `modifiers` as I have no use case for them yet. `key` and `modifiers` rely on `grab_keyboard`

mattkae:
@AlanGriffiths For the `input_shape` business, I can reproduce it with `maliit-keyboard` by I triggering the virtual keyboard to open. Basically;
1. Run `miral-shell` (on this PR)
2. Run `maliit-keyboard` from within the shell
3. `unset GTK_IM_MODULE`
4. Run `gedit`

If you have the `input_shape` check there, you won't be able to click anything (rectangle is zero), but if you remove it, the rectangle will be appropriately set

AlanGriffiths:
> I had just taken the default `miriway-shell.config` and added:
> 
> ```
> add-wayland-extensions=zwp_input_method_v1:zwp_input_panel_v1
> ```
> 
> Then I logged in and ran `maliit-keyboard` from the terminal within the shell.

When I try that I see the following, but no keyboard appears (and no additional output when running ` env -u GTK_IM_MODULE gedit`):

```plain
$ maliit-keyboard 
wordengine.cpp plugin "/usr/lib/x86_64-linux-gnu/maliit/keyboard2/languages/en/libenplugin.so" loaded
inputmethod_p.h registerActiveLanguage(): activeLanguage is: "en"
in inputMethod.cpp setActiveLanguage() activeLanguage is: "en"
DB path: /usr/lib/x86_64-linux-gnu/maliit/keyboard2/languages/en/database_en.db
spellechecker.cpp in setLanguage() lang= "en" dictPath= "/usr/share/hunspell"
spellechecker.cpp in setLanguage() aff_file= "/usr/share/hunspell/en_AU.aff" dic_file= "/usr/share/hunspell/en_AU.dic" user dictionary= "/home/alan/.local/share/maliit-keyboard/en_userDictionary.dic"
inputMethod::reset()
in clear preedit.. clearing word engine
QSoundEffect(pulseaudio): Error decoding source file:///usr/share/maliit/keyboard2/sounds/key_tick2_quiet.wav
```

> For some reason, my miriway doesn't like it when I specify a "shell-component" in the config. Is that the way to set a startup application?

What do you mean "doesn't like it"? There are already some "shell-components" in the config: the background and top bar, are stated this way.

AlanGriffiths:
OK, some progress:

using `miriway --add-wayland-extensions=all` I am able to start the keyboard from the terminal, so a part of the problem is identifying the extensions it needs.

However, the keyboard is not always appearing when focus is on a text field. E.g. after starting the terminal I pressed `Ctrl-Shift-T` to get a tab to start gedit, and the keyboard disappeared. Will list some more scenarios after identifying the Wayland extensions being used by Maliit.

AlanGriffiths:
The other extension that maliit needs is `zwp_primary_selection_device_manager_v1`. So the following works:

miriway-shell.config:
```
add-wayland-extensions=zwp_primary_selection_device_manager_v1:zwp_input_method_v1:zwp_input_panel_v1
shell-component=miriway-unsnap env WAYLAND_DEBUG=client maliit-keyboar
```

However, `shell-add-wayland-extension` doesn't work (probably unrelated to this PR). I have to investigate why  as we don't want to enable these extensions globally (just for trusted components like the keyboard).

Saviq:
> The other extension that maliit needs is `zwp_primary_selection_device_manager_v1`.

That doesn't sound like something an OSK inherently needs? Maybe can be made optional in Maliit. @dobey do you happen to know?

AlanGriffiths:
The reason that `shell-component` doesn't work with mallit-keyboard is that the process isn't the one launched by miral-shell. Some debugging:

```
y$ ps -eaf | grep keyboard
gdm        37769   37490  0 Aug16 tty1     00:00:00 /usr/libexec/gsd-keyboard
alan       80563   80434  1 11:56 tty4     00:00:00 maliit-keyboard
alan       80740   80672  0 11:56 pts/0    00:00:00 grep --color=auto keyboard
$ grep component miriway.log 
[2023-08-17 11:54:15.396568] < - debug - > miriway: shell-component: dbus-update-activation-environment, pid=80040
[2023-08-17 11:54:15.399034] < - debug - > miriway: shell-component: swaybg, pid=80041
[2023-08-17 11:54:15.402276] < - debug - > miriway: shell-component: yambar, pid=80042
[2023-08-17 11:54:15.404770] < - debug - > miriway: shell-component: miriway-unsnap, pid=80044
[2023-08-17 11:54:15.407475] < - debug - > miriway: shell-component: miriway-unsnap, pid=80046
[2023-08-17 11:54:15.410292] < - debug - > miriway: shell-component: miriway-unsnap, pid=80047
[2023-08-17 11:54:15.413103] < - debug - > miriway: shell-component: miriway-unsnap, pid=80048
```
It isn't obvious why maliit-keyboard would fork()/exec(), but apparently it does.

AlanGriffiths:
Another problem here, is that while `squeekboard` (used by Ubuntu Frame OSK) sets up an exclusion zone around the keyboard (via the layer-shell extension) we're not doing anything equivalent here.

That means that it is very easy to have the keyboard obscuring the corresponding input field.

AlanGriffiths (changes requested):
With the above "working" configuration:

* there's a problem with the keyboard disappearing when opening a terminal tab;
* we need to discuss whether we can set up an exclusion zone so that the keyboard doesn't hide the input; and,
* "enter" isn't working with any of the apps I tried (gnome-console, gedit, Firefox)..

There are additional mallit-keyboard issues, that shouldn't block this PR:

* mallit-keyboard doesn't work without zwp_primary_selection_device_manager_v1
* mallit-keyboard runs as a different pid than the process we launch

@dohbee any insight on these?

mattkae:
> Given
> 
> ```
> snap install miriway --channel edge/mir-pr3000
> apt install maliit-keyboard
> ```
> 
> and with the following `miriway-shell.config` configuration:
> 
> ```
> add-wayland-extensions=zwp_primary_selection_device_manager_v1:zwp_input_method_v1:zwp_input_panel_v1
> shell-component=miriway-unsnap maliit-keyboard
> ```
> 
> * there's a problem with the keyboard disappearing when opening a terminal tab;
> * we need to discuss whether we can set up an exclusion zone so that the keyboard doesn't hide the input; and,
> * "enter" and "bksp" are not working with any of the apps I tried (gnome-console, gedit, Firefox)..
> 
> There are additional mallit-keyboard issues, that shouldn't block this PR:
> 
> * mallit-keyboard doesn't work without zwp_primary_selection_device_manager_v1
> * mallit-keyboard runs as a different pid than the process we launch
> 
> @dobey any insight on these?

As we spoke about earlier, enter and backspace will not work  easily with any program that is using text-input-v3. If the problem is encountered so frequently though, then I am tempted to implement that as part of this work instead of leaving it for a later ticket. What are you thoughts on this @AlanGriffiths ?

mattkae:
> * we need to discuss whether we can set up an exclusion zone so that the keyboard doesn't hide the input; and,

This appears to be what `kwin_wayland` does, so I imagine that's what we should do :smile: Does there exist a mechanism in Mir to accomplish this?

AlanGriffiths:
> > * we need to discuss whether we can set up an exclusion zone so that the keyboard doesn't hide the input; and,
> 
> This appears to be what `kwin_wayland` does, so I imagine that's what we should do 😄 Does there exist a mechanism in Mir to accomplish this?

Yes, but I'm not sure of the plumbing needed.

If, instead of using maliit-keyboard, you try ubuntu-frame-osk you can see it in action:

```
snap install ubuntu-frame-osk
snap connect ubuntu-frame-osk:wayland
```

miriway-shell.config
```
shell-component=miriway-unsnap ubuntu-frame-osk
```

At the window management end, the application "Zone" is adjusted to reflect the exclusion zone set via layer-shell. (Let me know if you need help tracing how it works - I've not looked at it recently)

dobey:
> * mallit-keyboard doesn't work without zwp_primary_selection_device_manager_v1

I am not sure what that is, and afaik it is not used by maliit, so it is likely causing some side effect initialization in Mir, which input-method-v1 or input-panel-v1 needs to do but is not?

> mallit-keyboard runs as a different pid than the process we launch

Is something perhaps causing it to crash and be restarted to cause this? I don't see how else this could happen.

AlanGriffiths:
> I am not sure what that is, and afaik it is not used by maliit, so it is likely causing some side effect initialization in Mir, which input-method-v1 or input-panel-v1 needs to do but is not?

No, this is maliit: if we don't publish the `zwp_primary_selection_device_manager_v1` global, then maliit doesn't bind to either `zwp_input_method_v1` or `zwp_input_panel_v1`. As a result of which it doesn't work. (And it does bind `zwp_primary_selection_device_manager_v1`, so it likely does use it.)

dobey:
> > I am not sure what that is, and afaik it is not used by maliit, so it is likely causing some side effect initialization in Mir, which input-method-v1 or input-panel-v1 needs to do but is not?
> 
> No, this is maliit: if we don't publish the `zwp_primary_selection_device_manager_v1` global, then maliit doesn't bind to either `zwp_input_method_v1` or `zwp_input_panel_v1`. As a result of which it doesn't work. (And it does bind `zwp_primary_selection_device_manager_v1`, so it likely does use it.)

A `grep -ri zwp_prim` in `maliit-framework` (where the wayland support is implemented) returns nothing, so maliit itself is definitely not using it. Only thing I can think of then is that maybe QtWayland binds it internally. Otherwise, I have no idea what could cause that.

AlanGriffiths:
> > * there's a problem with the keyboard disappearing when opening a terminal tab;
> >   This sounds like a normal occurrence with any OSK, if the tab is taking focus, rather than the terminal widget.

That would be surprising. With squeekboard, for example, it is possible to use the keyboard with the new terminal tab.

> > * we need to discuss whether we can set up an exclusion zone so that the keyboard doesn't hide the input; and,
> >   Somehow, the rectangle for the keyboard is known by Qt, and such apps can access the data, to resize internally for avoiding having input areas blocked by the keyboard. However, I'm not entirely sure of the specifics of how that works, though it is expected for apps to use this on Ubuntu Touch for example, and I believe it also works on Android for Qt apps. Though, kwin_wayland tries to be more clever and resizes app windows as the keyboard animates in and out, I suppose it should be up to WM devs who might be using Mir underneath to decide this policy for their WM, rather than Mir enforcing a policy for all WMs using it. Also, there will always be a case where some input will either be off screen or underneath the keyboard rectangle, if app devs do not build convergent applications with consideration for an OSK or other IME being used. So I'd suggest this shouldn't be a blocker for merging the support in.

This is a bit of missed integration into Mir's window management. @mattkae is well on the way to fixing it.

> > * "enter" and "bksp" are not working with any of the apps I tried (gnome-console, gedit, Firefox)..
> >   That is quite odd, and I'm pretty sure they work fine with these apps under kwin_wayland. Though, I do not use snaps, so if those apps are from snaps for you, there may be some issue there instead of with this change, or the keyboard itself.

According to @mattkae this is an issue with text-input-v3 integration, and text-input-v{1,2} works. He's looked at how kwin gets this to work and intends to apply the same approach to Mir

AlanGriffiths:
> > mallit-keyboard runs as a different pid than the process we launch
> 
> Is something perhaps causing it to crash and be restarted to cause this? I don't see how else this could happen.

No, I think I reached the wrong conclusion. Something weird is happening in Miriway's handling of extensions and I need to debug there

mattkae:
> However, the keyboard is not always appearing when focus is on a text field. E.g. after starting the terminal I pressed `Ctrl-Shift-T` to get a tab to start gedit, and the keyboard disappeared. Will list some more scenarios after identifying the Wayland extensions being used by Maliit.

This wasn't a problem for me. Was there another step to the repo?
[Screencast from 08-17-2023 04:06:58 PM.webm](https://github.com/MirServer/mir/assets/25062299/76eca9b8-f128-476f-8579-0a431b43912e)

RAOF:
> Somehow, the rectangle for the keyboard is known by Qt, and such apps can access the data, to resize internally for avoiding having input areas blocked by the keyboard. However, I'm not entirely sure of the specifics of how that works, though it is expected for apps to use this on Ubuntu Touch for example, and I believe it also works on Android for Qt apps.

This would be provided by the `input_panel_state` event on `text-input-unstable-v2`; I note that we never send that event, so clients are unable to move their content out of the way.

src/server/frontend_wayland/text_input_v3.cpp
```diff
@@ -292,6 +322,26 @@ void mf::TextInputV3::send_text_change(ms::TextInputChange const& change)
...
+        auto nano = std::chrono::milliseconds{change.keysym->time};
+        keyboard_device->if_started_then([&](input::InputSink* sink, input::EventBuilder* builder)
+            {
+                auto state = change.keysym->state;
+                auto scan_code = change.keysym->sym;
+                switch (scan_code)
+                {
+                    case XKB_KEY_Return:
+                        // TODO: kwin_wayland maps this to 28, but I am unsure how they get there
+                        scan_code = 28;
```
RAOF, line 336: So, (one possible) scancode for Enter is `0x1C`, which *is* 28. Perhaps kwin is attempting to canonicalise the scancode for some reason? There's probably some libxkb function to do whatever this is. :woman_shrugging:

RAOF (commented):
I am not a huge fan of implementing a protocol by reverse engineering what a single client does, but that's the world we live in.

We might want to implement sending `input_panel_state` in order to get clients to move their content out of the way of the OSK, but that might be followup.

src/server/frontend_wayland/text_input_v3.cpp
AlanGriffiths: Well, to avoid a magic number, there's `KEY_ENTER` from `linux/input-event-codes.h`. (That is, as Chris says, `0x1c`.)

It isn't obvious why this mapping is necessary, but if it works...

I thought I'd seen something similar in the X11 input platform, but all I found was `xcb_keycode_get_scan_code()` which is different "magic".

I don't see an `xkb_*` function to do this mapping.

AlanGriffiths:
> > However, the keyboard is not always appearing when focus is on a text field. E.g. after starting the terminal I pressed `Ctrl-Shift-T` to get a tab to start gedit, and the keyboard disappeared. Will list some more scenarios after identifying the Wayland extensions being used by Maliit.
> 
> This wasn't a problem for me. Was there another step to the repo? [Screencast from 08-17-2023 04:06:58 PM.webm](https://github.com/MirServer/mir/assets/25062299/76eca9b8-f128-476f-8579-0a431b43912e)

Are you using the default gnome-terminal? That doesn't show the problem. I was using gnome-console. (Sorry, I didn't make that clear and hadn't tested other terminals.)

```
apt install gnome-console
kgx
```

No, I don't know why the gnome-console binary is called kgx.

----

qterminal and xfce4-terminal also don't show a problem. And weston-terminal doesn't support input methods.

While it is possible that gnome-console is in error here, we should understand what happens (as it doesn't happen with squeekboard)

src/server/frontend_wayland/text_input_v3.cpp
RAOF: Oh! I wonder if kwin feeds these scancodes into it's xkb state to get a keysym and then does the reverse mapping with `keycode_get_scancode`. It's not really obvious what the protocol is expecting to happen here 🤷‍♀️

AlanGriffiths: I assume something similar is required for BkSp?

I don't see Tab, Home, End, Insert, Del, Pg{Up,Down}, arrow keys, meta keys or function keys on the default maliit-keyboard so I suspect there are other mappings required here that we're yet to discover.

AlanGriffiths:
@mattkae I have found and fixed the miriway bug that prevented this `miriway-shell.config` working:

```
shell-add-wayland-extension=zwp_input_method_v1
shell-add-wayland-extension=zwp_input_panel_v1
shell-add-wayland-extension=zwp_primary_selection_device_manager_v1 # it isn't obvious why maliit-keyboard needs this
shell-component=miriway-unsnap maliit-keyboard
```

You may need to refresh the snap to get the latest mir-pr build:

```
snap refresh miriway --channel edge/mir-pr3000
```

@dobey I was wrong about maliit-keyboard doing something weird and forking, I guess I looked at debug logs from different runs, but the unexplained use of `zwp_primary_selection_device_manager_v1` is still a thing. But it obviously isn't a blocker for this PR

src/server/frontend_wayland/text_input_v3.cpp
mattkae: Funny enough, backspace works without any additional mapping

mattkae: It was time to read `kwin`'s code I suppose :smile: Looks like they do the following in `https://github.com/KDE/kwin/blob/master/src/inputmethod.cpp`:
```cpp
static quint32 keysymToKeycode(quint32 sym)
{
    switch (sym) {
    case XKB_KEY_BackSpace:
        return KEY_BACKSPACE;
    case XKB_KEY_Return:
        return KEY_ENTER;
    case XKB_KEY_Left:
        return KEY_LEFT;
    case XKB_KEY_Right:
        return KEY_RIGHT;
    case XKB_KEY_Up:
        return KEY_UP;
    case XKB_KEY_Down:
        return KEY_DOWN;
    default:
        return KEY_UNKNOWN;
    }
}
```

mattkae: If it's a text input v3, they have no other choice. If it works, it works :shrug:

AlanGriffiths:
Ugh, more weirdness!

The [above configuration](https://github.com/MirServer/mir/pull/3000#issuecomment-1683717721) works when starting Miriway from a VT, but not when started from GDM. (With the latter, `maliit-keyboard` seems to start fine, but never appears.)

src/server/frontend_wayland/text_input_v3.cpp
```diff
@@ -292,6 +323,41 @@ void mf::TextInputV3::send_text_change(ms::TextInputChange const& change)
...
+                    case XKB_KEY_Up:
+                        scan_code = KEY_UP;
+                        break;
+                    case XKB_KEY_Down:
+                        scan_code = KEY_DOWN;
+                        break;
+                    default:
+                        scan_code = KEY_UNKNOWN;
+                        break;
+                }
```
AlanGriffiths, line 356: Maybe extract this as a utility function?

```suggestion
                auto const scan_code = sym_to_scan_code(change.keysym->sym);
```

src/server/frontend_wayland/text_input_v3.cpp
```diff
@@ -331,11 +332,26 @@ void mf::TextInputV3::send_text_change(ms::TextInputChange const& change)
...
+                        scan_code = KEY_RIGHT;
+                        break;
+                    case XKB_KEY_Up:
+                        scan_code = KEY_UP;
+                        break;
+                    case XKB_KEY_Down:
+                        scan_code = KEY_DOWN;
                         break;
                     default:
+                        scan_code = KEY_UNKNOWN;
```
AlanGriffiths, line 354: That default is different from previous versions. Intentional?

mattkae: Yes it appears to be the case. `kwin` thinks that the defaults in the cases are the _only_ keysym events that should be sent via input method

mattkae:
Confirmed that the `kgx` bug is present in `kwin` as well, but still going to investigate what happens there, in case its a fixable thing
[Screencast from 08-18-2023 10:24:47 AM.webm](https://github.com/MirServer/mir/assets/25062299/461e991a-9cfe-4db4-9d5e-be5d869c8c80)

mattkae:
## What's happening with `kgx`?
- When you open up `kgx`, you initial send a `zwp_text_input_v3::enable`
- When you do a "Ctrl + Alt + T", we disable the previous text input with `zwp_text_input_v3::disable` _BUT_ `kgx` never re-enables the new text input, so we do not see the keyboard

AlanGriffiths:
>     * When you do a "Ctrl + Alt + T", we disable the previous text input with `zwp_text_input_v3::disable` _BUT_ `kgx` never re-enables the new text input, so we do not see the keyboard

I assume you mean `Ctrl-Shift-T`, OK that's likely a gnome-console bug. Definitely not a blocker

AlanGriffiths (changes requested):

AlanGriffiths (approved):
Looks good now. (But we need to check the libmiral versioning - we already have a release branched with 3.10)

PR #3009, opened 2023-08-16
Fix the Alpine/edge FTBFS

AlanGriffiths:
Alpine/edge has dropped the mmap64 symbol that previously existed "everywhere"

### DIFF
--- a/tests/mir_test_framework/mmap_wrapper.cpp
+++ b/tests/mir_test_framework/mmap_wrapper.cpp
@@ -61,6 +61,14 @@ void* mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset)
     void* (*real_mmap)(void *addr, size_t length, int prot, int flags, int fd, off_t offset);
     *(void **)(&real_mmap) = dlsym(RTLD_NEXT, real_mmap_symbol_name());
 
+#if _FILE_OFFSET_BITS == 64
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
Beyond the FTBFS, I am a bit puzzled by this code (maybe I once knew, but have forgotten):

1. do we need to resolve the forwarding on every call? Or could we cache the result?
2. What is the advantage of forwarding to mmap64?

RAOF (approved):
I'd mildly prefer to fix `real_mmap_symbol_name()` to return the real mmap symbol name, but a half-hour's browse of POSIX standards has left me no closer to understanding *when* `mmap64` is expected to exist.

I've tried a different approach in #3011 that I'd be happier with, but apparently my happiness correlates with failure on armhf, so is not a good indicator :woman_facepalming: 

> do we need to resolve the forwarding on every call? Or could we cache the result?

We do not; we *could* cache the result. This code doesn't, because this is not performance-critical and caching adds an extra thing that might (although, almost certainly won't) go wrong.

> What is the advantage of forwarding to mmap64?

It *used* to be the case that `mmap64` was the *only* symbol that existed on 64-bit platforms; `mmap` was implemented as `#define mmap mmap64`, which `dlsym` obviously can't do anything useful with.

I'm still a *little* concerned that this will come back to bite us - the failure mode for picking the wrong `mmap` symbol is that *sometimes* things die in apparently-impossible ways on *some* builds, maybe - but I can't think of a better option. :woman_shrugging:

PR #3010, opened 2023-08-16
Release 2.15.0

Saviq:


AlanGriffiths:
You also need to bump MIR_VERSION_MINOR

AlanGriffiths:
Also, we talked about bumping the libmiral ABI (which could also be an opportunity to drop some deprecated symbols)

Saviq (commented):
This is good to go.

The actual merge into main will look like this:

https://github.com/MirServer/mir/compare/main-with-2.15?expand=1

AlanGriffiths (approved):

PR #3011, opened 2023-08-22
tests: Fix detection of real mmap symbol

RAOF:
For hysterical raisins glibc `#define mmap mmap64`s, which means attempting to `dlsym(..., "mmap")` fails, because "mmap" isn't a real function.

Conversely, musl libc does not define `mmap64` *at all* anymore, so we can't just `dlsym(..., "mmap64")`.

Instead of trying to guess which symbol really exists, just get the preprocessor to expand that out for us and use *that*.

Fixes FTBFS on Alpine edge (alternative to #3009)


AlanGriffiths (approved):
Looks good, but I see mir_umock_unit_tests.UdevWrapperDeathTest.* failing unexpectedly. Looks like Chris re-ran that, so will discuss before landing

Issue #3014, opened 2023-08-22
ubsan: "‘new’ of ‘initializer_list’ does not extend the lifetime"


Saviq:
This popped up a couple weeks ago:

https://github.com/MirServer/mir/actions/workflows/spread.yml?query=branch%3Amain

```
[ 73%] Building CXX object tests/unit-tests/CMakeFiles/mir_unit_tests.dir/input/test_touchspot_controller.cpp.o
[...]

PR #3015, opened 2023-08-22
Don't pass temporaries via initializer_list

AlanGriffiths:
Fixes: #3014


AlanGriffiths:
This code originated with a "Fix FTBFS with g++-13", so let's first see if it builds everywhere

Saviq (approved):
Yup!

PR #3016, opened 2023-08-23
snap: fix multi arch matrix

Saviq:
And fix indentation.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -49,6 +49,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        snap:
+        - MirServer/mir-test-tools
+        - MirServer/confined-shell-wip
+        - MirServer/ubuntu-frame
+        - Miriway/Miriway
         architecture:
         - amd64
         - armhf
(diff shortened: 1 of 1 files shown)


.github/workflows/snap.yml
```diff
@@ -49,10 +49,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        snap:
+        - MirServer/mir-test-tools
+        - MirServer/confined-shell-wip
+        - MirServer/ubuntu-frame
+        - Miriway/Miriway
```
AlanGriffiths, line 56: I'm not sure of the significance of this. These snaps are already listed under `include` (with additional `track` and `review-opts` fields

Saviq: That's [the correct way](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations) to actually have a product of the two. The `include` ones end up as annotations, in a way, matched by the common keys.

So it first goes and makes a product of `architectures` × `snaps` and then annotates the jobs with values from `include`, where `snap` matches.

Compare the jobs ran now vs. previously:

https://github.com/MirServer/mir/actions/runs/5944659782

AlanGriffiths (approved):
OK, that makes sense

PR #3020, opened 2023-08-25
CI/snaps: Bump timeout for building mir-libs.

RAOF:
armhf has consistently been *slightly* too slow to hit the 4hr(!) timeout we currently have.

Since this doesn't gate merging, bump it up to 6hrs to ensure people can test PR builds. Eventually.

### DIFF
--- a/.github/workflows/snap.yml
+++ b/.github/workflows/snap.yml
@@ -12,7 +12,7 @@ jobs:
   mir-libs:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


Saviq (approved):
And just as you did that ;)

> succeeded … in 3h 30m 36s

But yeah, let's. I'll think of speeding this up somehow. May also want `fail-fast: false`, but that's a bit tricky with `needs: mir-libs` below anyway.

PR #3021, opened 2023-08-25
snap: try `remote-build` instead

Saviq:


RAOF (approved):
That is a *nice* increase in speed!

PR #3023, opened 2023-08-29
platform: Drop `NativeBuffer` detritus

RAOF:
Turns out there's *still* a bunch of only-used-by-mirclient detritus in the platform code. Remove this bit.


AlanGriffiths (approved):
I see red

PR #3026, opened 2023-08-29
ci: enable AddressSanitizer

Saviq:


mattkae:
## `Runner.register_signal_handler_before_setup_invokes_callback_after_setup`
I am 50% sure about this one, but I was able to fix it...

- We are passing a `std::initializer_list<int>` to `register_signal_handler` by rvalue
- We are most likely passing the values in this initializer list by reference, depending on the implementation
- _However_, this is only a problem if we use it before we call `miral::TestServer::SetUp()`. This is because `SetUp` _also_ registers a signal in `mir::run_mir`, which somehow results in a copy

I don't know why this all happens, but I will investigate deeper later

mattkae:
## `UdevWrapperDeathTest.DereferencingEndReturnsInvalidObject` && `UdevWrapperDeathTest.MemberDereferenceOfEndDies`

In both cases, these are because of "The signal is caused by a READ memory access". I am 90% sure that the point of these tests is to have a bad memory access. @Saviq Is there a way to silence asan in these instances?

mattkae:
## `FullProbeStack.select_display_modules_loads_all_available_hardware_when_no_nested`
- This one is a simple case of loading `stub_input_platform.cpp` twice, leading the static members on `StubInputPlatform` to be defined twice

mattkae:
## `TextInputV2WithInputMethodV1Test.input_method_can_set_modifiers_map`
That was a bug! The modifiers were being used after they were freed, since that resource isn't trivialyl copyable. It got released, and then we tried to use it later on in another thread

mattkae:
@AlanGriffiths Requesting you as a reviewer here since I modified some C++ stuff in a way that you  might be interested in

mattkae:
Ah it's breaking again :thinking:

mattkae:
## FullProbeStack.* and InputPlatformProbe.*
These are failing (surprisingly only on clang) because `mirplatforminputevdevobjects` is getting loaded twice (once it is linked to at build time, the other time it is getting loaded dynamically).

tests/include/mir_test_framework/stub_input_platform.h
```diff
@@ -49,18 +60,16 @@ class StubInputPlatform : public mir::input::Platform
...
+    void unregister_dispatchable(std::shared_ptr<mir::dispatch::Dispatchable> const& queue);
 
 private:
     std::shared_ptr<mir::dispatch::MultiplexingDispatchable> const platform_dispatchable;
     std::shared_ptr<mir::dispatch::ActionQueue> const platform_queue;
     std::shared_ptr<mir::input::InputDeviceRegistry> const registry;
-    static std::atomic<StubInputPlatform*> stub_input_platform;
-    static std::vector<std::weak_ptr<mir::input::InputDevice>> device_store;
-    static std::mutex device_store_guard;
+    std::shared_ptr<DeviceStore> const device_store;
```
AlanGriffiths, line 72: Would an alternative design be to move these static data members to namespace scope in the implementation file?

AlanGriffiths (changes requested):
Mostly, I don't like the input hub rework.

Classes that own resources (`wl_array*` in this case) should only manage that resource & data members should handle their own copy/assign semantics

src/include/server/mir/scene/text_input_hub.h
```diff
@@ -105,6 +106,53 @@ struct TextInputChange
...
+            wl_array_release(modifier_map.value());
+            modifier_map = std::nullopt;
+        }
+
+        auto copy = new wl_array();
+        wl_array_init(copy);
+        wl_array_copy(copy, map);
+        modifier_map = copy;
+    }
+
```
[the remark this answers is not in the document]
mattkae: That is the better idea that I was looking for! I felt the same way, but did not think of this

tests/include/mir_test_framework/stub_input_platform.h
mattkae: Good question. Let me know if I'm right in my understanding as to why that won't work. 

The problem there would be that we'd have two symbols included from different sources. The test code still needs to access the `StubInputPlatform` symbol _and_ it needs to load the dynamic library at runtime. Hence, if we included a `static` variable in the `StubInputPlatform`, we would be declaring it twice.

Does that check out for you too?

AlanGriffiths: Which `StubInputPlatform` symbol?

Moving the statics means there is, for example, no `StubInputPlatform::stub_input_platform` symbol. Just a local symbol in the implementation file that can't be accessed from elsewhere

tests/unit-tests/test_udev_wrapper.cpp
```diff
@@ -318,6 +324,10 @@ typedef UdevWrapperTest UdevWrapperDeathTest;
 
 TEST_F(UdevWrapperDeathTest, DereferencingEndReturnsInvalidObject)
 {
+#if defined(__SANITIZE_ADDRESS__)
+    (void)udev_environment;
+    (void)KilledByInvalidMemoryAccess;
+#else
```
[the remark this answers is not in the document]
mattkae: Yeah I think so too. I wonder if I could disable them in a smarter way globally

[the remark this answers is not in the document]
AlanGriffiths: 1. Have a `gtest_filter` list to exclude from the run
2. Move them to a separate file (or files) and don't build them

tests/include/mir_test_framework/stub_input_platform.h
mattkae: Hmm I guess you're right there! It should be okay, I will try that

src/include/server/mir/scene/text_input_hub.h
```diff
@@ -97,6 +98,48 @@ struct TextInputState
...
+    CopyableWlArray(CopyableWlArray const& other)
+        : data_{copy(other.data_)}
+    {
+    }
+
+    CopyableWlArray& operator=(CopyableWlArray const& other)
+    {
+        data_ = copy(other.data_);
+        return *this;
+    }
```
AlanGriffiths, line 118: This leaks `data_`. The canonical form of assignment is:
```suggestion
    CopyableWlArray& operator=(CopyableWlArray other)
    {
        std::swap(data, other.data);
        return *this;
    }
```

tests/include/mir_test_framework/stub_input_platform.h
mattkae: Ah nevermind. What about the static methods:

```c++
static void add(std::shared_ptr<mir::input::InputDevice> const& dev);
    static void remove(std::shared_ptr<mir::input::InputDevice> const& dev);
    static void register_dispatchable(std::shared_ptr<mir::dispatch::Dispatchable> const& queue);
    static void unregister_dispatchable(std::shared_ptr<mir::dispatch::Dispatchable> const& queue);
```

Those are public. I needed to move them into the `StubInputPlatformAccessor` to allow those who need it to include those symbols.

src/include/server/mir/scene/text_input_hub.h
```diff
@@ -97,6 +98,48 @@ struct TextInputState
...
+public:
+    explicit CopyableWlArray(wl_array* in_data)
+        : data_{copy(in_data)}
+    {
+    }
+
+    CopyableWlArray(CopyableWlArray const& other)
+        : data_{copy(other.data_)}
+    {
+    }
```
AlanGriffiths, line 112: This could be a forwarding constructor
```suggestion
    CopyableWlArray(CopyableWlArray const& other)
        : CopyableWlArray{other.data_}
    {
    }
```

src/include/server/mir/scene/text_input_hub.h
```diff
@@ -97,6 +98,48 @@ struct TextInputState
...
+    CopyableWlArray& operator=(CopyableWlArray const& other)
+    {
+        data_ = copy(other.data_);
+        return *this;
+    }
+
+    ~CopyableWlArray()
+    {
+        if (data_)
+            wl_array_release(data_);
```
AlanGriffiths, line 123: AFAICS, `data_ != nullptr` is an invariant. (There is no default constructor, and the converting constructor will crash if called with `nullptr)`

AlanGriffiths (changes requested):
Just simplifying CopyableWlArray

tests/include/mir_test_framework/stub_input_platform.h
AlanGriffiths: Remember, you don't need to change this, I was simply curious why this way

mattkae:
Trying to figure out why the asan_clang build works locally but fails in Github actions right now

tests/unit-tests/test_udev_wrapper.cpp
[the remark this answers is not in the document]
AlanGriffiths: For prior art see: https://github.com/canonical/mir/blob/605712f075587d06b3808cee98a350e9e4f98244/cmake/MirCommon.cmake#L76-L90

mattkae:
Got the build error:

```
[  7%] Building CXX object src/common/events/CMakeFiles/mirevents.dir/window_placement_event.cpp.o
cd /spread/mir/build-amd64/src/common/events && ccache /usr/bin/clang++ -DEGL_NO_X11 -DLTTNG_UST_HAVE_SDT_INTEGRATION -DMIR_LOG_COMPONENT_FALLBACK=\"mircommon\" -DMIR_USE_APPARMOR -DMIR_VERSION_MAJOR=2 -DMIR_VER
SION_MICRO=0 -DMIR_VERSION_MINOR=18 -D_FILE_OFFSET_BITS=64 -D_GNU_SOURCE -I/spread/mir/src/include/common -I/spread/mir/include/common -I/spread/mir/include/core -isystem /usr/include/libdrm -fpch-validate-input
-files-content -pthread -g -Wall -fno-strict-aliasing -pedantic -Wnon-virtual-dtor -Wextra -fPIC -Werror -Wno-mismatched-tags -Wno-psabi -std=c++23 -fsanitize=address -fno-omit-frame-pointer -MD -MT src/common/e
vents/CMakeFiles/mirevents.dir/window_placement_event.cpp.o -MF CMakeFiles/mirevents.dir/window_placement_event.cpp.o.d -o CMakeFiles/mirevents.dir/window_placement_event.cpp.o -c /spread/mir/src/common/events/w
indow_placement_event.cpp
make[2]: Leaving directory '/spread/mir/build-amd64'
`.text.asan.module_ctor.149' referenced in section `.init_array.1[asan.module_ctor.149]' of /tmp/lto-llvm-8df671.o: defined in discarded section `.text.asan.module_ctor.149[asan.module_ctor]' of /tmp/lto-llvm-8d
f671.o
`.text.asan.module_ctor.158' referenced in section `.init_array.1[asan.module_ctor.158]' of /tmp/lto-llvm-8df671.o: defined in discarded section `.text.asan.module_ctor.158[asan.module_ctor]' of /tmp/lto-llvm-8d
f671.o
`.text.asan.module_dtor.150' referenced in section `.fini_array.1[asan.module_dtor.150]' of /tmp/lto-llvm-8df671.o: defined in discarded section `.text.asan.module_dtor.150[asan.module_dtor]' of /tmp/lto-llvm-8d
f671.o
`.text.asan.module_dtor.159' referenced in section `.fini_array.1[asan.module_dtor.159]' of /tmp/lto-llvm-8df671.o: defined in discarded section `.text.asan.module_dtor.159[asan.module_dtor]' of /tmp/lto-llvm-8d
f671.o
[  7%] Built target wayland_generator_test_generated_files
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [examples/client/CMakeFiles/mir_demo_client_mir_shell.dir/build.make:118: bin/mir_demo_client_mir_shell.bin] Error 1
make[2]: Leaving directory '/spread/mir/build-amd64'
make[1]: *** [CMakeFiles/Makefile2:5044: examples/client/CMakeFiles/mir_demo_client_mir_shell.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/spread/mir/build-amd64'
[  7%] Built target mirevents
make[1]: Leaving directory '/spread/mir/build-amd64'
make: *** [Makefile:149: all] Error 2

```

mattkae:
We get many spurious failures in the asan build due to some threading issues, which isn't great...

AlanGriffiths:
> We get many spurious failures in the asan build due to some threading issues, which isn't great...

If we identify threading issues in our code we should be fixing them

src/include/server/mir/scene/text_input_hub.h
```diff
@@ -97,6 +99,47 @@ struct TextInputState
...
+    static wl_array* copy(wl_array* in_data)
+    {
+        auto data = new wl_array();
+        wl_array_init(data);
+        wl_array_copy(data, in_data);
+        return data;
+    }
+
+    wl_array* data_;
+};
```
AlanGriffiths, line 141: This fails to delete `data_`, but the whole thing can be further simplfied...
```suggestion
class CopyableWlArray
{
public:
    explicit CopyableWlArray(wl_array const* in_data)
    {
        wl_array_init(&data_);
        wl_array_copy(&data_, const_cast<wl_array*>(in_data)); // wl_array_copy() doesn't mark source as const
    }

    CopyableWlArray(CopyableWlArray const& other)
        : CopyableWlArray{&other.data_}
    {
    }

    CopyableWlArray& operator=(CopyableWlArray other)
    {
        std::swap(data_, other.data_);
        return *this;
    }

    ~CopyableWlArray()
    {
        wl_array_release(&data_);
    }

    [[nodiscard]] wl_array* data() const
    {
        // Our generated `send_modifiers_map_event()` requires a non-const wl_array, so
        // this workaround may be the wrong thing
        return const_cast<wl_array*>(&data_);
    }

private:

    wl_array data_;
};
```

tests/unit-tests/test_udev_wrapper.cpp
```diff
@@ -26,11 +26,17 @@
 #include <poll.h>
 #include <sys/sysmacros.h>
 
+#if defined(__has_feature)
+#   if __has_feature(address_sanitizer) // for clang
+#       define __SANITIZE_ADDRESS__ // GCC already sets this
+#   endif
+#endif
+
```
AlanGriffiths, line 34: Rather than relying on a compiler specific macro with a reserved name and faking it on a different compiler, I'd prefer to deal with this in CMake. Either by supplying a `MIR_...` definition we control, or by having the affected code in a separate file that we can exclude from the build.

AlanGriffiths (changes requested):

AlanGriffiths (approved):
I'm not entirely sure about the StubInputPlatform rework.

The runner.cpp change is weird: the same incantation occurs in two other places and they are not changed.

But nothing I'll block on.

AlanGriffiths:
@mattkae I assume you are still investigating the ubuntu:asan_clang failure?

mattkae:
> @mattkae I assume you are still investigating the ubuntu:asan_clang failure?

Yup. Doesn't repro on my machine, testing some things out now

mattkae:
Fixed & Cleaned up the commits. Just waiting on one last check :crossed_fingers:

tests/include/mir/test/barrier.h
```diff
@@ -38,7 +38,7 @@ class Barrier
     void ready()
     {
         std::unique_lock lock(mutex);
-        if (--wait_threads)
+        if ((--wait_threads) > 0)
```
AlanGriffiths, line 41: You realise that's an implicit conversion of 0 to unsigned? That just makes the code confusing

mattkae: Oh whoops you are write!

PR #3033, opened 2023-09-01
(EXPERIMENTAL) Provides a "transitional" boot from plymouth to mir

mattkae:
# Experimental WIP

## What's new?

- Introducing `SmoothBootSupport` option for miral applications that, when set, will provide a smooth transition from the previous image on screen (e.g. `plymouth`) to the compositor
- `Display`s can now provide an `InitialRender` that will be shown by the compositor for "some time"
- The `InitialRenderManager` manages how `InitialRender`s are shown. For instance, the InitialRenderManager will fade the renders out over time
- When the `smooth_boot` option is true, then e do NOT set an initial framebuffer in `display_buffer.cpp`, thus preserving the previous screen


mattkae:
Shelves for the time being

PR #3034, opened 2023-09-06
Disable the initial set of the CRTC when modes match

mattkae:
## How to test :test_tube: 
1. Provide the `run_with` command for your compositor with the new `SmoothBoothSupport` option. An example of adding this to `shell_main.cpp` would be: 

```cpp
...
return runner.run_with(
        {
            WaylandExtensions{},
            X11Support{},
            SmoothBootSupport{}, // Add me!
            window_managers,
            display_configuration_options,
            external_client_launcher,
            launcher,
            config_keymap,
            AppendEventFilter{quit_on_ctrl_alt_bksp},
            StartupInternalClient{spinner},
            ConfigurationOption{run_startup_apps, "startup-apps", "Colon separated list of startup apps", ""},
            pre_init(ConfigurationOption{[&](std::string const& typeface) { ::wallpaper::font_file(typeface); },
                                         "shell-wallpaper-font", "font file to use for wallpaper", ::wallpaper::font_file()}),
            ConfigurationOption{[&](std::string const& cmd) { terminal_cmd = cmd; },
                                "shell-terminal-emulator", "terminal emulator to use", terminal_cmd}
        });
```

2. Start your computer in a tty (this means that `gbm` cannot start. see [this link](systemctl set-default multi-user.target) on how to do that)
3. Once started, `ssh` into your computer from another computer as root
4. Run:
```sh
plymouthd
plymouth show-splash
plymouth deactivate
```
5. Then run your shell, e.g. `miral-shell`
6. Note that you should no longer see a black screen in-between plymouth and the compositor

## What's new :new: 
- Providing `create_display` with the options to that it can read whether or not it should do a smooth boot
- If it should do a smooth boot, the `Displaybuffer` will NOT initialize the initial visible frame buffer AND it will NOT set the CRTC for each output. This ensures that plymouth's previous contents remain on the screen

## Some caveats :hole: 
- We _may_ see a flicker if the next `set_crtc` changes the resolution or something

## Up next :next_track_button: 
I have another branch ready to go that will read the mapping into a Mir buffer and offer it back to the `Scene` to be displayed. However, this will call on us to transition the buffer out of existence, which is a bigger story


Saviq:
@mattkae would you please add the option to the demo server?

src/platforms/gbm-kms/server/kms/display_buffer.cpp
```diff
@@ -63,22 +64,24 @@ mgg::DisplayBuffer::DisplayBuffer(
...
-        false,
-        DRMFormat{DRM_FORMAT_XRGB8888},
-        area.size);
-
-    auto mapping = initial_fb->map_writeable();
-    ::memset(mapping->data(), 24, mapping->len());
-
-    visible_fb = std::move(initial_fb);
-    for (auto& output : outputs)
+    if (!smooth_transition)
```
RAOF, line 67: So, this is going to break if someone sets `smooth_transition` and the previous resolution was not the same as the resolution our policy has set - we'll have a `view_area()` that doesn't match the actual view area; I'm not sure what else (if anything!) will notice the discrepancy between the display configuration requested and the actual hardware configuration.

It now occurs to me that we could *always* skip setting the CRTC if the current mode is the same as the mode we want to set. The `smooth_boot` option could then be either removed and you get flicker-free boot if and only if the requested resolution is the same as the pre-existing mode, or plumbed through here and a *warning* generated if `smooth_boot` is asked for but cannot be provided because we need to modeset, or handled higher up (there'd need to be some way to pipe the *actual* display configuration up to the display configuration policy? Maybe?).

mattkae: Yeah that sounds like the bug that I would expect! I wonder if there exists some way to query the current resolution and then just set our default view area to that instead?

That's a good point... We would still want the flag for the follow up pull request where we provide a neat transition from the previous screen to the compositor's first frame, but it also makes sense that it would be a sane default.

At any rate, I think we have two options:

1. When we start, we can read the current CRTC and set mir's internals to this resolution. This would avoid _any_ `set_crtc` from happening, as we would always adopt the previous resolution
2. When we start, we can read the current CRTC and compare it to the CRTC that we want. If they mismatch, we would have to `set_crtc` and cause a flicker. Otherwise, we would have a smooth boot.

@Saviq @RAOF thoughts?

mattkae: Note: `drmModeGetCrtc` does what you think it does :smile:

mattkae: Another note: Option 2 is really easy to pull off

Saviq: +1 on doing it by default and doing away with the option altogether :)

src/platforms/gbm-kms/server/kms/display_buffer.cpp
```diff
@@ -63,22 +63,44 @@ mgg::DisplayBuffer::DisplayBuffer(
...
-
-    visible_fb = std::move(initial_fb);
+    bool needs_crtc_set = false;
     for (auto& output : outputs)
     {
-        output->set_crtc(*visible_fb);
+        try
+        {
+            geometry::Rectangle rectangle = output->get_rectangle();
+            if (rectangle != area)
```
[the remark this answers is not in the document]
mattkae: Some questions:
- Do we even need to compare the entire `drmModeModeInfo` here? We have `current_mode_index` and the requested `mode_index`. Can we just compare the indices? Or is there some benefit to doing a deep compare there?
- At startup, it _seems_ to me like we're already setting the requested `mode_index` to the `current_mode_index` (which is read from the `current_crtc`) [See `mgg::Display::configure_locked`]. We can still do the check, but 99% of the time it appears that we are going to try and set the mode to the same mode that is currently set anyway (unless there is an error).

I think that moving the logic here to `KMSOutput` makes a lot of sense though!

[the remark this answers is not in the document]
mattkae: Eh, we can trivially compare them, why just compare the indices :shrug:

RAOF (approved):
Yup!

We'll probably extend this later into removing `schedule_set_crtc` and `needs_set_crtc` entirely, but that can probably wait for Atomic KMS.

Issue #3036, opened 2023-09-09
display scaling breaks gl renderer clipping


seanpringle:
A display scaling factor appears to break Mir GL renderer window clipping:

* Using main at 7d67a3378cd22f5d8cf2b3442f9d64999d231bfe 
* In a Gnome3 shell Wayland session, set display scaling to 200%
* Run miral-shell directly pointing to the host Wayland, using the example tiling window manager:
```bash
WAYLAND_DISPLAY=mir DISPLAY=no MIR_SERVER_WAYLAND_HOST=wayland-0 build/bin/miral-shell --window-manager tiling
```
* Start a single instance of the Ctrl-Alt-T terminal window, or any app, or just watch the splash screen
* Window geometry looks full-screen but only the lower left quarter is rendered

```
[2023-09-08 17:24:17.272855] <information> mirserver: Initial display configuration:
[2023-09-08 17:24:17.272876] <information> mirserver: * Output 0: unknown connected, used
[2023-09-08 17:24:17.272891] <information> mirserver: . |_ Physical size 27.7" 610x350mm
[2023-09-08 17:24:17.272899] <information> mirserver: . |_ Power is on
[2023-09-08 17:24:17.272906] <information> mirserver: . |_ Current mode 3840x2160 60.00Hz
[2023-09-08 17:24:17.272913] <information> mirserver: . |_ Preferred mode 3840x2160 60.00Hz
[2023-09-08 17:24:17.272919] <information> mirserver: . |_ Orientation normal
[2023-09-08 17:24:17.272927] <information> mirserver: . |_ Logical size 1920x1080
[2023-09-08 17:24:17.272933] <information> mirserver: . |_ Logical position +0+0
[2023-09-08 17:24:17.272939] <information> mirserver: . |_ Scaling factor: 2.00
```

Full log in case: [mir-bug.txt](https://github.com/MirServer/mir/files/12565711/mir-bug.txt)

![mir-clipping](https://github.com/MirServer/mir/assets/120825/5bbb28ee-1ee2-4b5f-8a7d-ce2028f684b3)

I am currently compensating for this in `mir::renderer::gl::Draw()` by scaling the renderable's clip_area and the logical viewport to match the render_target size, before calculating the screen-space rectangle to pass to `glScissor()`, like this: 319466f75f27c46848bcc2d9ab92f2e404486429

[...]

PR #3037, opened 2023-09-11
[wayland platform] Less broken display on scaled hosts

AlanGriffiths:
Don't confuse scaling of host environment with internal scaling of outputs.

Fixes: #3036

Note: this fixes the display and cursor positioning and so is generally an improvement. However, there are remaining oddities with the cursor image. These are arguably pre-existing, and don't make things unusable

### DIFF
--- a/src/platforms/wayland/displayclient.cpp
+++ b/src/platforms/wayland/displayclient.cpp
@@ -58,6 +58,7 @@ public:
 
     DisplayConfigurationOutput dcout;
     geom::Size output_size;
+    float host_scale{1.0f};
 
     wl_output* const output;
     DisplayClient* const owner;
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
The alpine-edge failure is fixed by #3038

src/platforms/wayland/displayclient.cpp
```diff
@@ -718,7 +719,8 @@ void mgw::DisplayClient::pointer_enter(
     {
         if (surface == out.second->surface)
         {
-            pointer_displacement = out.second->dcout.top_left - geometry::Point{};
+            pointer_displacement = geom::DisplacementF{out.second->dcout.top_left - geometry::Point{}};
+            host_scale = out.second->host_scale;
```
mattkae, line 723: Why are we setting the `host_scale` during `pointer_enter`?

mattkae (changes requested):
One question that maybe warrants a comment in the code if it is non-obvious

PR #3038, opened 2023-09-11
Drop support for ancient gtest versions

AlanGriffiths:


Saviq (approved):
If this blends, it blends!

Issue #3040, opened 2023-09-14
Window resize not always sent to clients


AlanGriffiths:
Failing to notify the client of a window resize is the root cause of https://github.com/MirServer/ubuntu-frame-osk/issues/58

Tis can be reproduced (intermittently - around 1 in 4) with:

    snap set frame-it osk=enable
    MIR_SERVER_WINDOW_MANAGEMENT_TRACE=1 frame-it ./wayland-debug-wrapper wpe-webkit-mir-kiosk.cog

Here's an extract from a log of the messages received by the client and the Mir log with `--window-management-trace`.

Note that the only `xdg_toplevel@17.configure()` is to the smaller size of (1268, 741), and occurs before the `advise_application_zone_update updated=(0, 0, (1268, 994)), original=(0, 0, (1268, 741))`.

Following the `advise_application_zone_update()` the Frame window manager correctly resizes the window to `new_size=(1268, 994)` but there is no corresponding `xdg_toplevel@17.configure()` event. As a result the client continues to submit (1268, 741) buffers (which is interpreted as a resize request which is overruled by the WM)

```
[545592.862] xdg_toplevel@17.configure(1268, 741, array)
[545592.873] xdg_surface@16.configure(9)
[545592.877]  -> xdg_surface@16.ack_configure(9)
[545592.881] wl_keyboard@20.enter(11, wl_surface@13, array)
[545592.888] zwp_text_input_v3@18.enter(wl_surface@13)
[545601.569] wl_display@1.delete_id(24)
[545601.618] wl_callback@24.done(290107331)
[545601.765] wl_display@1.delete_id(15)
[545601.779] wl_display@1.delete_id(13)
[545601.784] wl_display@1.delete_id(14)
[545601.788] wl_callback@13.done(0)
Cog-Core-Message: 16:14:06.299: <https://mir-server.io/> Loading...
[2023-09-14 16:14:06.362411] < -warning- > mirserver: Ignoring layer shell protocol violation: wl_surface destroyed before associated zwlr_layer_surface_v1@28
[2023-09-14 16:14:06.362501] <information> miral::Window Management: advise_delete_window window_info={name=, type=freestyle, state=attached, top_left=0, 741, size=(1268, 253), restore_rect=(634, 310, (0, 126)), children={}, depth_layer=4, attached_edges=11, exclusive_rect=1, preferred_orientation=0xf, confine_pointer=0, output_id=1}
[...]

PR #3042, opened 2023-09-15
Ensure we send a configure when we resize

AlanGriffiths:
Checking against the size of the most recent buffer (`impl->window_size`) is mostly unnecessary and may be incorrect.

Unnecessary, as we already have logic in the scene to avoid sending idempotent resize notifications. (But when first sizing surfaces, we still check to avoid a redundant `configure()` that breaks some tests.)

It is incorrect when we've sent a `configure()` but not received a buffer with configured size.

Fixes: #3040


Saviq (approved):
What's more, this seems to avoid an extra osk/resize/redraw on startup (visible with `snap set wpe-webkit-mir-kiosk url=https://google.com/`).

PR #3044, opened 2023-09-18
Reduce scope for confusion between Tips and Gloss

AlanGriffiths:
We've failed to distinguished "Gloss" and "Tip" in our window management implementation, and that's led us to labelling "Tips" `mir_window_type_gloss`.

Label "Tips" `mir_window_type_tip` correctly. (Issue about the incorrect "Gloss" implementation: #3045.)


mattkae (approved):
Makes sense to me!

PR #3048, opened 2023-09-20
Avoid spurious configure events

AlanGriffiths:
When determing whether the client has sized the window, we were inadvertantly applying a fallback size of 640x480.

That led to spurious configure events.

Fixes: #3046


AlanGriffiths:
This also fixes the tests that were broken by #3042 - the spurious configure events also were what caused that trouble

Saviq (approved):
WFM!

Issue #2509, opened 2022-07-18
Resizing Mattermost sends it crazy


AlanGriffiths:
1. Open the `mattermost-desktop` snap on a Mir compositor (without hacking it to use Wayland)
2. Try to resize it's window

Expect: window resizes
Actual: window disappears and mattermost hogs a CPU

(This doesn't occur using Wayland, so is likely to do with our Xwayland interaction)

mattkae:
Note that this even happens when we fullscreen the mattermost application

mattkae:
Fun fact: It does not _disappear_: it just gets a teeny tiny width and height. The _very tiny_ black bar on the left hand side of this image is mattermost:
![Screenshot from 2023-09-21 15-02-36](https://github.com/MirServer/mir/assets/25062299/14774765-5c42-46cf-b1fd-dedb322129db)

mattkae:
The problem is that the `WindowInfo::max_width/max_height` are both 1

mattkae:
Fixed by #3050

PR #3049, opened 2023-09-21
(#2509) Handling integer overflows for dimensions for XWaylandSurface

mattkae:
# https://github.com/MirServer/mir/issues/2509

You can find some more info on the solution to this one in the ticket itself :point_up: 

## How to test
- Run `miral-app` or something
- Open up mattermost
- Click the fullscreen button _or_  resize it using the edges
- Note that this now works!

## Solution

- I am going to assume that the width and height are never naturally negative. Hence, if we overflow during the multiplication, we can expect the int to go negative.
- This might just be an unsafe multiplication all-around, especially if we have a negative scale (e.g. a scale of 0.25 would result in an `inv_scale` of 8)

## Screen recording
[Screencast from 09-21-2023 05:02:28 PM.webm](https://github.com/MirServer/mir/assets/25062299/c29bab31-daf9-4755-88e8-1bd94b4fbf63)

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -1223,10 +1223,10 @@ void mf::XWaylandSurface::prep_surface_spec(ProofOfMutexLock const&, msh::Surfac
 
 #define SCALE_SIZE(type, prop) \
     if (mods.prop) \
-    {                          \
+    { \
         auto scaled_value = mods.prop.value() * inv_scale; \
-        if (scaled_value.as_int() >= 0)            \
-        {                      \
+        if (scaled_value.as_int() >= 0) \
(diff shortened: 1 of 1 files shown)


RAOF (commented):
Hm. I *think* this might be better handled in the converting `Value` constructors? (`geometry/dimensions.h` lines 72 and 78)

Rather than just YOLOing `static_cast<>` there, do a saturating cast? (ie: check `value` against `std::numeric_limits<T>` and return `::max`/`::min` as appropriate).

There's *never* a case where we want UB there (which is the current situation) and I don't think there's ever a case where we'd want anything other than saturation?

AlanGriffiths:
Thanks for tracking this down, but I don't like this solution either. And I'm not sure the converting constructor is the place either. See also #3050

PR #3053, opened 2023-09-26
Architecture diagram and description of the project's folders

mattkae:
# How to test
- Open up read the docs at the bottom of this pull request

# What's new?
- Implemented an architecture diagram for Mir
- Implement an architecture document for Mir


doc/sphinx/architecture.md
```diff
@@ -0,0 +1,131 @@
+# Overview
+This document describes the high-level architecture of *Mir*, including a diagram and a brief introduction of the directory structure.
+
+# Concepts
+## MirAL
+- The "Mir Abstraction Layer" (aka `mirAL` or `miral`) is a layer that makes the core Mir implementation easier for compositor-implementers to work with.
```
AlanGriffiths, line 6: I don't think this clarifies anything
```suggestion
- The "Mir Abstraction Layer" (aka `MirAL` or `miral`) is an API that makes Mir easy for compositor-implementers to work with. It provides the core window management concepts and an interface that is more ABI stable than Mir's internal mirserver API.
```

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,131 @@
...
+- If one wishes to implement a compositor themselves, they will spend most of their time interacting with the `mirAL` API and only really touch the `mir` API when they require core concepts like a `mir::geometry::Rectangle`.
```
AlanGriffiths, line 7: `Rectangle` is from mircore, and that's very different from the mirserver APIs (that compositor writers should never touch.

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,131 @@
...
+  - *X11*: a useful platform for testing locally
+  - *Wayland*: another useful platform for testing locally
+  - *EGLStreams/KMS*: like the *GBM/KMS* platform, but for Nvidia 
+- For the input platforms, we have:
+  - evdev
+  - X11
+  - Wayland
+
+## `/src/platform` (no "s")
+- Graphics and input code that is shared between platforms
```
AlanGriffiths, line 103: Comment: In principle, we aim to make this a public interface for third party "platforms". As Chris has been reworking these APIs, they are currently private

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+- This directory *strictly* includes the interfaces that are useful to a compositor-implementer.
+- If one is only interested in implementing their own *Mir*-based compositor, they need not look any further than this directory.
+
+## `/src`
+- The `src` folder at the root of the project comprises the core implementation of *Mir*.
+
+## `/examples`
+- A collection of demo *Mir* compositors that are used as examples of how one might build a true compositor. Examples include:
+  - `miral-app`
+  - `miral-shel`
```
AlanGriffiths, line 22:
```suggestion
  - `miral-shell`
```

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+  - `mir_demo_server`
+
+## `/tests`
+- Unit tests for the entire project.
```
AlanGriffiths, line 26: Not just _unit_ tests

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+    ├── mirshelldecoration
+    ├── mirscene
+    ├── mirgraphics
+    ├── mirreport
+    ├── mirlttng
+    ├── mirlogging
+    ├── mirnullreport
+    ├── mirconsole
+    ├── mirgl
+    └── mirrenderergl
```
AlanGriffiths, line 52: _I_ wouldn't expand mirserver here. I think that introduces too many concepts.

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+- Data structures and utility functions like Mir-event building, logging, keymaps, etc.
+- Differs from `mircore` in that the concepts map strictly onto *Mir*'s concepts instead of being general concepts like a Rectangle
+- Public-facing
+- Directories:
+  - `include/common`
+  - `src/common`
+  - `src/include/common`
+
+## mircookie
+- Provides event timestamps for inputs events that are difficult to spoof.
```
AlanGriffiths, line 94: @RAOF I suspect that mircookie isn't actually useful now we're no longer shipping Mir's input events to clients. WDYT?

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+- Provides the core *Mir* runtime, with the entry-point being the `mir::Server` class.
+- The `Server` class has a `DisplayServer` which provides access to the interfaces that that support the server's runtime.
+  - The `ServerConfiguration` provides a bunch of methods called `the_XYZ()` to access these various instances, where "XYZ" can be replaced with the interface of interest.
+- The classes that the `Server` relies on for its runtime can be found organized in separate folders in this directory.
+- Directories:
+  - `include/server`
+  - `src/server`
+  - `src/include/server`
+  
+## mircompositor
```
AlanGriffiths, line 111: I don't like this naming (probably comes from trying to treat it at the same level of abstraction as libraries).

It is "just" a namespace within mirserver.

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+- Included in `mirserver.so`
+- Deals with everything input (e.g. keyboard presses, mouse clicks)
+- The `InputManager` provides access to the specific input platform that is running (e.g. X, Wayland, or libinput).
+- The `InputDispatcher` is responsible for taking events bubbled up from the platform and sending them off to the appropriate listeners
+- Directories
+  - `src/include/server/mir/input`
+  - `src/server/input`
+  
+## mirfrontend
+- TODO: How do I describe this one?
```
AlanGriffiths, line 128: `src/server/frontend` is vestigial - we should probably merge it with `src/server/frontend_wayland`

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+## mirshell
+- The `Shell` is responsible for managing the surfaces that the compositor wants to show.
+- It provides an interface to create, update, and remove these surfaces from the display.
+- As an example, the `WaylandConnector` might ask the `Shell` to create a new surface for it.
+- Directories:
+  - `src/include/server/mir/shell`
+  - `src/server/shell`
+  
+## mirshelldecoration
+- Manages the decoration of *Mir*'s windows
```
AlanGriffiths, line 155:
```suggestion
- Manages *Mir*'s server-side decoration of windows. Currently only used for X11 clients.
```

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+- Directories:
+  - `src/include/mir/server/graphics`
+  - `src/server/graphics`
+  
+## mirreport
+- Default logging and reporting facilities
+- Directories:
+  - `src/server/report`
+
+## mirlttng
```
AlanGriffiths, line 179: I like this naming even less. `lttng` is a directory/namespace within `report` - it shouldn't be named at the same level as `mirwayland`. Similarly, `logging`, `null`.

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,252 @@
...
+## graphics-x11
+- An X11-based display/input platform
+- Primary platform used for testing
+- Directories:
+  - `src/platforms/x11`
+  
+## graphics-wayland
+- A Wayland-based display/input platform
+- Directories:
+  - `src/platforms/wayland`
```
AlanGriffiths, line 242: Suddenly we're not prefixing everything with "mir"! :grinning: 

I think it works better to explain what a platform is, and then list examples.

AlanGriffiths (changes requested):
There is too much "flattening" here. There is meaningful hierarchy:

1. mirserver is the "god" library.

2. miral (and miroil) are the support for building compositors with Mir.

3. Platforms are adaptors that allow mirserver to work across  different graphics stacks

4. There are support libraries that provide the underpinnings for the above: mircore, mircommon, mircookie, mirplatform & mirwayland.

If you try and present parts of parts of mirserver as top-level concepts you end up with too many things to explain at the same time

doc/sphinx/architecture.md
mattkae: It could be `mirserver::mircompositor`?

doc/sphinx/architecture.md
mattkae: I believe I only did this because the final `.so` names aren't prefaced with mir here :joy: But yes, agreed

doc/sphinx/architecture.md
AlanGriffiths: Sure, they need explaining but we're well past the number of concepts that should be introduced together. (Even Agatha Christie stops at a dozen suspects and she was aiming to confuse!)

First introduce [the high level concepts](https://github.com/MirServer/mir/pull/3053#pullrequestreview-1650985226). Then, break down each one in turn: (e.g. "These are the platforms... this one is for X11...", "This is the mirserver... it has these components...").

That gives the reader some structure to organise their understanding.

doc/sphinx/architecture.md
AlanGriffiths: That's worse! How is a reader to associate something in the code (`mir::compositor::Foo`, say) with `mirserver::mircompositor`?

doc/sphinx/architecture.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Mirwayland
+Compositor authors may want to define their own wayland protocol extensions in addition to the ones that the core *Mir* implementation defines. The `mirwayland` library satisfies this requirement. This library may be used in conjunction with either `miral` or `miroil`.
+
+## The Mir engine
+The `mirserver` library is the core implementation of *Mir*. It serves as the engine for both `miral` and `miroil`. This library does the heavy-lifting for the compositor and, as a result, it is the largest piece of *Mir*. This section will explain the primary concepts that drive the engine.
+
+### Core Concepts
+At the heart of `mirserver` are two interfaces: the **Shell** and the **Scene**. The `Shell` is responsible for fielding requests from the rest system. The `Shell` then modifies the state of the `Scene` to reflect the requested changes. For example, part of the system would ask the `Shell` to move a window. The `Shell` would then decide what position to move that window to and update the state of the `Scene` to reflect that change.
```
AlanGriffiths, line 34:
```suggestion
At the heart of `mirserver` are two interfaces: the **Shell** and the **Scene**. The `Shell` is responsible for fielding requests from the rest system. The `Shell` then modifies the state of the `Scene` to reflect the requested changes. For example, the `Frontend` would ask the `Shell` to initiate dragging a window. The `Shell` would then decide how to move that window to and update the state of the `Scene` to reflect that change.
```

AlanGriffiths (commented):
The architecture.md reads much better now. Please press on updating the rest to match

AlanGriffiths:
>     * Open up read the docs at the bottom of this pull request

Should there be a link somewhere? I don't see it.

Saviq:
> Should there be a link somewhere? I don't see it.

Click "Show all checks" and:

obraz
`                                                                                                         `:point_up:

AlanGriffiths (changes requested):
Content works for me.

A couple of tweaks though:

1. Could we incorporate the diagrams into the "Architecture" page instead of a separate page?
2. The "Closeup" would look tidier with the "Shell" tree and the "Compositor" tree swapped (or the "Scene" and "miral" trees). Is there a way to tweak the layout?

mattkae:
> 2. The "Closeup" would look tidier with the "Shell" tree and the "Compositor" tree swapped (or the "Scene" and "miral" trees). Is there a way to tweak the layout?

The order that things appear in the document is the order that they will appear in the diagram (loosely). Should be possible!

Saviq (approved):
Yup, good start! :)

PR #3054, opened 2023-09-28
platforms/wayland: Fix a lock in DisplayClient

RAOF:
It's important to actually *name* your RAII lock object, so it doesn't immediately go out of scope and unlock the mutex!


Saviq (approved):
Yas.

PR #3055, opened 2023-09-29
Avoid undefined behavior

AlanGriffiths:
We don't guarantee that `XKBMapper::map_event()` sets `last_device_id` before we use it in `XKBMapper::update_modifier()` and valgrind complains.


Saviq (approved):
Yups.

PR #3057, opened 2023-09-30
Add .mailmap file to map old commits to Sophie's new name and email

wmww:
Hi!

My old name is all over Mir git history. I don't care that much so I never proposed filter-branching it out, but I recently found out with a simple .mailmap you can get a lot of the same benefits without changing history. If this is merged my new name will be displayed on my old commits and on git blame, both locally and on github.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,4 @@
+Sophie Winter <git@phie.me> <9331264+wmww@users.noreply.github.com>
+Sophie Winter <git@phie.me> <wilcoze@gmail.com>
+Sophie Winter <git@phie.me> <william.wold@canonical.com>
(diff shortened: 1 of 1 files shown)


wmww:
If this isn't objectionable I might make similar PRs to a bunch of the other Mir team repos I contributed to.

Saviq:
Hey o/, can you point at documentation on what respects this?

AlanGriffiths (approved):
This looks fine to me, but @Saviq has a question before we land

wmww:
Sorry, off GH for a while.

Git docs here: https://git-scm.com/docs/gitmailmap

I can't find any sort of official github docs but this stack overflow says it works: https://stackoverflow.com/questions/53629125/does-github-consider-mailmap-for-contribution-graph

PR #3060, opened 2023-10-04
Rework the demo client to use xdg-shell

AlanGriffiths:
Also:

  * give better errors when the compositor doesn't support required extensions; and,
  * allows compositor to resize windows


examples/client/wayland_client.c
```diff
@@ -47,30 +79,30 @@ static void new_global(
...
     uint32_t version)
 {
-    (void)version;
-    struct globals* globals = data;
+    (void)data;
 
     if (strcmp(interface, "wl_compositor") == 0)
     {
-        globals->compositor = wl_registry_bind(registry, id, &wl_compositor_interface, 3);
+        globals.compositor = wl_registry_bind(registry, id, &wl_compositor_interface, min(version, 3));
```
mattkae, line 86: Why are we accepting any version below here? Same question for other globals. Is it possible that a version isn't backwards compatible?

mattkae (changes requested):
One question about the meaning of `min` versions there. However, I tested it and the rest looks good :+1:

examples/client/wayland_client.c
AlanGriffiths: We can't bind to a version higher than the version supported by the server. In theory, we need to avoid any requests that don't exist in the agreed version, but we are pretty conservative in what we use in the demo,

mattkae (approved):
My one question is answered :+1: Works well on my end

PR #3061, opened 2023-10-04
Tidy up "frontend"

AlanGriffiths:
If we were to need a generic frontend we would need to rework stuff anyway. So no need to keep the last bit separate


mattkae (approved):
Not sure if the CI break is on your end, but the project builds and makes sense to me :+1:

PR #3062, opened 2023-10-04
Fix for satellite windows being moved when the parent moves

mattkae:
## What's new?
This is a suggestion, but I think it is correct. Satellite windows should not move with their parent, right?

## To test
- Fire up gimp with satellite windows
- Move the main window and note that the satellites don't  move as well

### DIFF
--- a/src/miral/minimal_window_manager.cpp
+++ b/src/miral/minimal_window_manager.cpp
@@ -272,6 +272,11 @@ bool miral::MinimalWindowManager::begin_pointer_resize(
 auto miral::MinimalWindowManager::confirm_inherited_move(WindowInfo const& window_info, Displacement movement)
 -> Rectangle
 {
+    // Satellite windows don't need to move with their parent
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
1. Gimp's satellites are not being identified as `mir_window_type_satellite`
2. Satellites are supposed to [move with the parent](https://mir-server.io/docs/window-positions-under-wayland)

AlanGriffiths:
> 1. Gimp's satellites are not being identified as `mir_window_type_satellite`

Oh, that's wrong

mattkae:
> 1. Gimp's satellites are not being identified as `mir_window_type_satellite`
> 
>     2. Satellites are supposed to [move with the parent](https://mir-server.io/docs/window-positions-under-wayland)

Interesting that satellites are supposed to move with the parent. That doesn't seem to be the behavior that other compositors (e.g. mutter) gives to gimp's windows here. Would this mean that those windows are a different type of window (not-satellite)?

PR #3065, opened 2023-10-06
Add display-autoscale option to miral::display_configuration_options()

AlanGriffiths:
This adds an autoscaling option to miral's display configuration options.

So to scale to emulate, for example, 1080p use `--display-autoscale=1080`

The scaling is quantized to the nearest quarter pixel

### DIFF
--- a/src/miral/display_configuration_option.cpp
+++ b/src/miral/display_configuration_option.cpp
@@ -25,6 +25,7 @@
 #include <algorithm>
 
 namespace mg = mir::graphics;
+using namespace mir::geometry;
 
 namespace
 {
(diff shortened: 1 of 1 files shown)


src/miral/display_configuration_option.cpp
```diff
@@ -119,12 +123,47 @@ void ScaleSetter::apply_to(mg::DisplayConfiguration& conf)
...
+void AutoscaleSetter::apply_to(mg::UserDisplayConfigurationOutput& output)
+{
+    auto const output_height =
+        (output.orientation == mir_orientation_normal || output.orientation == mir_orientation_inverted) ?
+        output.modes[output.current_mode_index].size.height.as_int() :
+        output.modes[output.current_mode_index].size.height.as_int();
+
+    static auto const steps = 4.0;
+    output.scale = roundf((steps * output_height) / target) / steps;
+}
```
Saviq, line 158: So this only takes resolution into account… Shouldn't we try and determine whether it's a high pixel density display (from physical size) instead?

AlanGriffiths: There are lots of things we *could* take into account, but we don't know what the right choices are.

For example , we could consider screen width (this approach may not be great on tall, thin outputs).

With pixel density we don't know whether the right choice is "Foo fits on screen" or "Foo has readable text".

Saviq (approved):
Works :)

PR #3067, opened 2023-10-06
Sharing CpuAdressableDisplayProvider between eglstreams and gbm platforms

mattkae:
## What's new?
- The EglStreams platform now has a `CPUAddressableDisplayProvider` so that it works in this specific hybrid situation
- Moved some classes to be shared between GBM and EGLStreams

## How to test
```
sudo XDG_RUNTIME_DIR=/run ./bin/mir_demo_server --platform-rendering-libs mir:gbm-kms
```

It still SIGABRTs after, but it gets through the selection part.

### DIFF
--- a/src/platforms/common/server/cpu_addressable_fb.cpp
+++ b/src/platforms/common/server/cpu_addressable_fb.cpp
@@ -16,7 +16,6 @@
 
 #include "cpu_addressable_fb.h"
 
-#include "mir/geometry/forward.h"
 #include "mir/log.h"
 #include "mir_toolkit/common.h"
 
@@ -26,9 +25,8 @@
 #include <drm_fourcc.h>
 
 namespace mg = mir::graphics;
-namespace mgg = mg::gbm;
 
-class mgg::CPUAddressableFB::Buffer : public mir::renderer::software::RWMappableBuffer
+class mg::CPUAddressableFB::Buffer : public mir::renderer::software::RWMappableBuffer
 {
     template<typename T>
     class Mapping : public mir::renderer::software::Mapping<T>
@@ -53,7 +51,7 @@ class mgg::CPUAddressableFB::Buffer : public mir::renderer::software::RWMappable
             if (::munmap(const_cast<typename std::remove_const<T>::type *>(data_), len_) == -1)
             {
                 // It's unclear how this could happen, but tell *someone* about it if it does!
-                mir::log_error("Failed to unmap CPU buffer: %s (%i)", strerror(errno), errno);
+                mir::log(mir::logging::Severity::error, "Failed to unmap CPU buffer: %s (%i)", strerror(errno), errno);
             }
         }
 
@@ -102,7 +100,7 @@ public:
 
         if (auto const err = drmIoctl(drm_fd, DRM_IOCTL_MODE_DESTROY_DUMB, &params))
         {
-            mir::log_error("Failed destroy CPU-accessible buffer: %s (%i)", strerror(-err), -err);
+            mir::log(mir::logging::Severity::error, "Failed destroy CPU-accessible buffer: %s (%i)", strerror(-err), -err);
         }
     }
 
@@ -248,7 +246,7 @@ private:
     size_t const size_;
 };
 
-mgg::CPUAddressableFB::CPUAddressableFB(
+mg::CPUAddressableFB::CPUAddressableFB(
     mir::Fd const& drm_fd,
     bool supports_modifiers,
     DRMFormat format,
@@ -257,12 +255,12 @@ mgg::CPUAddressableFB::CPUAddressableFB(
 {
 }
 
-mgg::CPUAddressableFB::~CPUAddressableFB()
+mg::CPUAddressableFB::~CPUAddressableFB()
 {
     drmModeRmFB(drm_fd, fb_id);
 }
 
-mgg::CPUAddressableFB::CPUAddressableFB(
+mg::CPUAddressableFB::CPUAddressableFB(
     mir::Fd drm_fd,
     bool supports_modifiers,
     DRMFormat format,
@@ -273,32 +271,32 @@ mgg::CPUAddressableFB::CPUAddressableFB(
 {
 }
 
-auto mgg::CPUAddressableFB::map_writeable() -> std::unique_ptr<mir::renderer::software::Mapping<unsigned char>>
+auto mg::CPUAddressableFB::map_writeable() -> std::unique_ptr<mir::renderer::software::Mapping<unsigned char>>
 {
     return buffer->map_writeable();
 }
 
-auto mgg::CPUAddressableFB::format() const -> MirPixelFormat
+auto mg::CPUAddressableFB::format() const -> MirPixelFormat
 {
     return buffer->format();
 }
 
-auto mgg::CPUAddressableFB::stride() const -> geometry::Stride
+auto mg::CPUAddressableFB::stride() const -> geometry::Stride
 {
     return buffer->stride();
 }
 
-auto mgg::CPUAddressableFB::size() const -> geometry::Size
+auto mg::CPUAddressableFB::size() const -> geometry::Size
 {
     return buffer->size();
 }
 
-mgg::CPUAddressableFB::operator uint32_t() const
+mg::CPUAddressableFB::operator uint32_t() const
 {
     return fb_id;
 }
 
-auto mgg::CPUAddressableFB::fb_id_for_buffer(
+auto mg::CPUAddressableFB::fb_id_for_buffer(
     mir::Fd const &drm_fd,
     bool supports_modifiers,
     DRMFormat format,
--- a/src/platforms/common/server/cpu_addressable_fb.cpp
+++ b/src/platforms/common/server/cpu_addressable_fb.cpp
@@ -51,7 +51,7 @@ class mg::CPUAddressableFB::Buffer : public mir::renderer::software::RWMappableB
             if (::munmap(const_cast<typename std::remove_const<T>::type *>(data_), len_) == -1)
             {
                 // It's unclear how this could happen, but tell *someone* about it if it does!
-                mir::log(mir::logging::Severity::error, "Failed to unmap CPU buffer: %s (%i)", strerror(errno), errno);
+                log_error("Failed to unmap CPU buffer: %s (%i)", strerror(errno), errno);
             }
         }
 
@@ -100,7 +100,7 @@ public:
 
         if (auto const err = drmIoctl(drm_fd, DRM_IOCTL_MODE_DESTROY_DUMB, &params))
         {
-            mir::log(mir::logging::Severity::error, "Failed destroy CPU-accessible buffer: %s (%i)", strerror(-err), -err);
+            log_error("Failed destroy CPU-accessible buffer: %s (%i)", strerror(-err), -err);
         }
     }
 
--- a/src/platforms/common/server/kms_cpu_addressable_display_provider.cpp
+++ b/src/platforms/common/server/kms_cpu_addressable_display_provider.cpp
@@ -14,7 +14,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "basic_cpu_addressable_display_provider.h"
+#include "kms_cpu_addressable_display_provider.h"
 #include "cpu_addressable_fb.h"
 #include <drm_fourcc.h>
 #include <xf86drm.h>
@@ -39,20 +39,20 @@ auto drm_get_cap_checked(mir::Fd const& drm_fd, uint64_t cap) -> uint64_t
 namespace mg = mir::graphics;
 namespace geom = mir::geometry;
 
-mg::BasicCPUAddressableDisplayProvider::BasicCPUAddressableDisplayProvider(mir::Fd drm_fd)
+mg::kms::CPUAddressableDisplayProvider::CPUAddressableDisplayProvider(mir::Fd drm_fd)
     : drm_fd{std::move(drm_fd)},
       supports_modifiers{drm_get_cap_checked(this->drm_fd, DRM_CAP_ADDFB2_MODIFIERS) == 1}
 {
 }
 
-auto mg::BasicCPUAddressableDisplayProvider::supported_formats() const
+auto mg::kms::CPUAddressableDisplayProvider::supported_formats() const
 -> std::vector<mg::DRMFormat>
 {
     // TODO: Pull out of DRM info
     return {mg::DRMFormat{DRM_FORMAT_XRGB8888}, mg::DRMFormat{DRM_FORMAT_ARGB8888}};
 }
 
-auto mg::BasicCPUAddressableDisplayProvider::alloc_fb(
+auto mg::kms::CPUAddressableDisplayProvider::alloc_fb(
     geom::Size size, DRMFormat format) -> std::unique_ptr<MappableFB>
 {
     return std::make_unique<mg::CPUAddressableFB>(drm_fd, supports_modifiers, format, size);
--- a/src/platforms/common/server/kms_cpu_addressable_display_provider.cpp
+++ b/src/platforms/common/server/kms_cpu_addressable_display_provider.cpp
@@ -43,6 +43,10 @@ mg::kms::CPUAddressableDisplayProvider::CPUAddressableDisplayProvider(mir::Fd dr
     : drm_fd{std::move(drm_fd)},
       supports_modifiers{drm_get_cap_checked(this->drm_fd, DRM_CAP_ADDFB2_MODIFIERS) == 1}
 {
+    if (!drm_get_cap_checked(this->drm_fd, DRM_CAP_DUMB_BUFFER))
+    {
+        BOOST_THROW_EXCEPTION((std::runtime_error{"DRM device doesn't support CPU buffers"}));
+    }
 }
 
 auto mg::kms::CPUAddressableDisplayProvider::supported_formats() const
--- a/src/platforms/common/server/kms_cpu_addressable_display_provider.cpp
+++ b/src/platforms/common/server/kms_cpu_addressable_display_provider.cpp
@@ -43,10 +43,6 @@ mg::kms::CPUAddressableDisplayProvider::CPUAddressableDisplayProvider(mir::Fd dr
     : drm_fd{std::move(drm_fd)},
       supports_modifiers{drm_get_cap_checked(this->drm_fd, DRM_CAP_ADDFB2_MODIFIERS) == 1}
 {
-    if (!drm_get_cap_checked(this->drm_fd, DRM_CAP_DUMB_BUFFER))
-    {
-        BOOST_THROW_EXCEPTION((std::runtime_error{"DRM device doesn't support CPU buffers"}));
-    }
 }
 
 auto mg::kms::CPUAddressableDisplayProvider::supported_formats() const
@@ -60,4 +56,17 @@ auto mg::kms::CPUAddressableDisplayProvider::alloc_fb(
     geom::Size size, DRMFormat format) -> std::unique_ptr<MappableFB>
 {
     return std::make_unique<mg::CPUAddressableFB>(drm_fd, supports_modifiers, format, size);
+}
+
+auto mir::graphics::kms::CPUAddressableDisplayProvider::create_if_supported(mir::Fd const& drm_fd)
+-> std::shared_ptr<CPUAddressableDisplayProvider>
+{
+    if  (drm_get_cap_checked(drm_fd, DRM_CAP_DUMB_BUFFER))
+    {
+        return std::shared_ptr<CPUAddressableDisplayProvider>(new CPUAddressableDisplayProvider{drm_fd});
+    }
+    else
+    {
+        return {};
+    }
 }
--- a/src/platforms/eglstream-kms/server/display.cpp
+++ b/src/platforms/eglstream-kms/server/display.cpp
@@ -231,7 +231,11 @@ public:
 
             scheduled_fb = std::move(next_swap);
             next_swap = nullptr;
-            output->queue_atomic_flip(*scheduled_fb, event_handler->drm_event_data());
+            if (!output->queue_atomic_flip(*scheduled_fb, event_handler->drm_event_data()))
+            {
+                // If we failed to submit the flip then we're not going to get the event!
+                event_handler->cancel_flip_events(output->crtc_id());
+            }
             return;
         }
 
--- a/src/platforms/eglstream-kms/server/display.cpp
+++ b/src/platforms/eglstream-kms/server/display.cpp
@@ -18,6 +18,7 @@
 
 #include "display.h"
 #include "egl_output.h"
+#include "kms_framebuffer.h"
 
 #include "kms-utils/drm_mode_resources.h"
 #include "mir/graphics/platform.h"
@@ -130,16 +131,12 @@ public:
         EGLContext ctx,
         EGLConfig config,
         std::shared_ptr<mge::DRMEventHandler> event_handler,
-        mge::kms::EGLOutput const& output,
+        std::shared_ptr<mge::kms::EGLOutput> output,
         std::shared_ptr<mg::DisplayReport> display_report)
         : owner{std::move(owner)},
           dpy{dpy},
           ctx{create_context(dpy, config, ctx)},
-          layer{output.output_layer()},
-          crtc_id{output.crtc_id()},
-          view_area_{output.extents()},
-          output_size{output.size()},
-          transform{output.transformation()},
+          output{output},
           drm_node{std::move(drm_node)},
           event_handler{std::move(event_handler)},
           display_report{std::move(display_report)}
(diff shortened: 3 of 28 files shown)


RAOF:
This'll get Mir starting, but it won't actually *display* anything. To get the CPU buffers on the screen, you'll need to additionally touch [`DisplayBuffer::overlay`](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/display.cpp#L285) to store the `Framebuffer` being passed in, and [`DisplayBuffer::post`](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/display.cpp#L199) to actually *use* the `Framebuffer` in a KMS call.

[Here](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/egl_output.cpp#L208) is where that is currently done, and specifically

mattkae:
> This'll get Mir starting, but it won't actually _display_ anything. To get the CPU buffers on the screen, you'll need to additionally touch [`DisplayBuffer::overlay`](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/display.cpp#L285) to store the `Framebuffer` being passed in, and [`DisplayBuffer::post`](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/display.cpp#L199) to actually _use_ the `Framebuffer` in a KMS call.
> 
> [Here](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/egl_output.cpp#L208) is where that is currently done, and specifically [here](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/egl_output.cpp#L268) is where the KMS FB ID is used.

Oh that's the `next_swap` business, huh?

mattkae:
> This'll get Mir starting, but it won't actually _display_ anything. To get the CPU buffers on the screen, you'll need to additionally touch [`DisplayBuffer::overlay`](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/display.cpp#L285) to store the `Framebuffer` being passed in, and [`DisplayBuffer::post`](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/display.cpp#L199) to actually _use_ the `Framebuffer` in a KMS call.
> 
> [Here](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/egl_output.cpp#L208) is where that is currently done, and specifically [here](https://github.com/MirServer/mir/blob/6b3b3bb202cbfdcf886b1f28ab8b82e3a8568355/src/platforms/eglstream-kms/server/egl_output.cpp#L268) is where the KMS FB ID is used.

And does that mean that we _sometimes_ use the `next_swap` and sometimes use the existing path?

RAOF:
> And does that mean that we _sometimes_ use the `next_swap` and sometimes use the existing path?

Yes. When there's a non-null `std::unique_ptr<Framebuffer>` in there it should take the direct KMS path, otherwise the EGL path.

As a practical matter I *think* any given instance is only ever going to use one of these paths - either always direct KMS, or always EGLStream.

mattkae:
I'm getting a permission denied error on `set_crtc` for the gbm-kms platform. I am unsure of the reason at the moment

Saviq:
I've not seen this one elsewhere:
```plain
 [ RUN ] RealKMSOutputTest.drm_set_gamma
/spread/mir/tests/unit-tests/platforms/gbm-kms/kms/test_real_kms_output.cpp:516: Failure
Actual function call count doesn't match EXPECT_CALL(mock_drm, drmModeCrtcSetGamma(drm_fd, crtc_ids[0], gamma.red.size(), const_cast<uint16_t*>(gamma.red.data()), const_cast<uint16_t*>(gamma.green.data()), const_cast<uint16_t*>(gamma.blue.data())))...
Expected: to be called once

[ FAILED ] RealKMSOutputTest.drm_set_gamma (0 ms)
[ RUN ] RealKMSOutputTest.drm_set_gamma_failure_does_not_throw
/spread/mir/tests/unit-tests/platforms/gbm-kms/kms/test_real_kms_output.cpp:545: Failure
Actual function call count doesn't match EXPECT_CALL(mock_drm, drmModeCrtcSetGamma(drm_fd, crtc_ids[0], gamma.red.size(), const_cast<uint16_t*>(gamma.red.data()), const_cast<uint16_t*>(gamma.green.data()), const_cast<uint16_t*>(gamma.blue.data())))...
Expected: to be called once

[ FAILED ] RealKMSOutputTest.drm_set_gamma_failure_does_not_throw (3 ms)
```

src/platforms/eglstream-kms/server/display.cpp
```diff
@@ -198,21 +195,37 @@ class DisplayBuffer
...
                     mg::Frame {
                         frame_count,
                         mir::time::PosixTimestamp(CLOCK_MONOTONIC, frame_time)});
             });
 
+        if (next_swap) // This is the KMS route
+        {
+            auto const predicted_render_time = 50ms; // Predicted worst case render time for the next frame...
+
+            output->set_crtc(*next_swap);
```
RAOF, line 218: This will do an immediate, tearing CRTC update (ie: it will not wait for vsync). I don't think this will trigger the flip event handler, either?

RAOF:
> I'm getting a permission denied error on `set_crtc` for the gbm-kms platform. I am unsure of the reason at the moment

Hm. It's not obvious why that would be the case. I *was* wondering whether we'd need to not do the EGLStream setup if we're using the CPUAllocated path. That is, defer the `eglStreamConsumerOutputEXT` call until an EGLStream consumer requests it.

But I wouldn't expect that to manifest as permission denied. I'll test this and see what I can immediately dig up tomorrow.

mattkae:
> > I'm getting a permission denied error on `set_crtc` for the gbm-kms platform. I am unsure of the reason at the moment
> 
> Hm. It's not obvious why that would be the case. I _was_ wondering whether we'd need to not do the EGLStream setup if we're using the CPUAllocated path. That is, defer the `eglStreamConsumerOutputEXT` call until an EGLStream consumer requests it.
> 
> But I wouldn't expect that to manifest as permission denied. I'll test this and see what I can immediately dig up tomorrow.

Sounds good let me know :+1:

RAOF:
With `--platform-display-libs=mir:eglstream-kms` and `--platform-rendering-libs=mir:gbm-kms` I do not see the permission denied. I *do* see an `abort()` in libepoxy, though.

Wait - I could have *sworn* that a previous version of this set `next_swap` in `overlay`, but the current version does not. That's not going to work!

With *that* fixed, it then doesn't crash, but it *does* hang indefinitely in `pending_flip.get()`, presumably because we haven't actually hooked up the flip event handling.

RAOF:
OK. The last commit I pushed *almost* works (and is not suitable for merging). Everything *seems* to work - post() doesn't block, no exceptions are raised, the frame events get signalled - except that nothing appears on the screen. :woman_facepalming:

RAOF:
https://gitlab.freedesktop.org/daniels/kms-quads/-/blob/master/kms.c is a pretty good reference for Atomic KMS shenanigans.

mattkae:
> OK. The last commit I pushed _almost_ works (and is not suitable for merging). Everything _seems_ to work - post() doesn't block, no exceptions are raised, the frame events get signalled - except that nothing appears on the screen. 🤦‍♀️

Reproduced this on my local setup at least :smile: Although I'm getting a mysterious error:

```
error in client communication (pid 1387)
[destroyed object]: error 4: Client requested unsupported format/modifier combination DRM_FORMAT_ARGB8888/NVIDIA:BLOCK_LINEAR_2D,HEIGHT=4,KIND=254,GEN=0,SEC

```

src/platforms/eglstream-kms/server/display.cpp
```diff
@@ -198,21 +195,40 @@ class DisplayBuffer
...
             });
 
+        if (next_swap) // This is the KMS route
+        {
+            visible_fb = std::move(scheduled_fb);
+            scheduled_fb = nullptr;
+
+            scheduled_fb = std::move(next_swap);
+            next_swap = nullptr;
+            auto const predicted_render_time = 50ms; // Predicted worst case render time for the next frame...
```
RAOF, line 221: We don't actually need to hook up the recommended sleep stuff here; it's not really hooked up anywhere else and it's not really useful.

src/platforms/eglstream-kms/server/egl_output.cpp
```diff
@@ -316,6 +320,42 @@ uint32_t mgek::EGLOutput::crtc_id() const
...
+    /* Set a surface for the plane */
+    drmModeAtomicAddProperty(request.get(), plane_id, plane_props->id_for("CRTC_ID"), crtc_id());
+    drmModeAtomicAddProperty(request.get(), plane_id, plane_props->id_for("FB_ID"), fb);
+
+    /* We don't monitor the DRM events (yet), so have no userdata */
+    auto ret = drmModeAtomicCommit(drm_fd, request.get(), DRM_MODE_ATOMIC_NONBLOCK | DRM_MODE_PAGE_FLIP_EVENT, const_cast<void*>(drm_event_userdata));
+
+    if (ret != 0)
+    {
+        BOOST_THROW_EXCEPTION((std::system_error{-ret, std::system_category(), "Failed to commit Atomic KMS state"}));
```
RAOF, line 352: We should *at least* fall back to trying again with `ALLOW_MODESET` (that would require sharing some more state with `configure`, notably the “mode blob”).

I don't *think* we need to have a fallback to drmModeSetCrtc, as we've guaranteed we have the atomic API available

src/platforms/eglstream-kms/server/egl_output.cpp
```diff
@@ -316,6 +320,42 @@ uint32_t mgek::EGLOutput::crtc_id() const
     return static_cast<uint32_t>(crtc_id);
 }
 
+bool mgek::EGLOutput::queue_atomic_flip(FBHandle const& fb, void const* drm_event_userdata)
```
RAOF, line 323: This should probably be shared with `configure`; it's largely the same code. Probably `configure` can be changed to call this.

RAOF (changes requested):

mattkae:
@RAOF I'm finding the following error when running `./bin/miral-shell.bin --platform-display-libs=mir:eglstream-kms --platform-rendering-libs=mir:gbm-kms` btw:

```
error in client communication (pid 2859)
[destroyed object]: error 4: Client requested unsupported format/modifier combination DRM_FORMAT_ARGB8888/NVIDIA:BLOCK_LINEAR_2D,HEIGHT=4,KIND=254,GEN=0,SEC
```

src/platforms/eglstream-kms/server/egl_output.cpp
mattkae: Well... actually it seems like configure _should_ do the ALLOW_MODESET for us huh?

src/platforms/eglstream-kms/server/egl_output.cpp
```diff
@@ -316,6 +275,67 @@ uint32_t mgek::EGLOutput::crtc_id() const
...
+    using_saved_crtc = false;
+    return 0;
+}
+
+bool mgek::EGLOutput::queue_atomic_flip(FBHandle const& fb, void const* drm_event_userdata)
+{
+    int ret = atomic_commit(fb, drm_event_userdata, DRM_MODE_ATOMIC_NONBLOCK | DRM_MODE_PAGE_FLIP_EVENT);
+    if (ret != 0)
+    {
+        BOOST_THROW_EXCEPTION((std::system_error{-ret, std::system_category(), "Failed to commit Atomic KMS state"}));
```
RAOF, line 333: So, there are a bunch of ways `atomic_commit` can fail, but transiently. It can require a modeset after switching back from VT switching. It can also fail with `EBUSY` if there's already a pending commit that hasn't finished (although we shouldn't hit that, having waited for previous page flips.)

Maybe we don't have to handle `EBUSY` here, and just assume that our page-filp handling will cover it.

We *should* handle the resume-from-VT-switch, which probably means twiddling a flag in `mge::Display::resume()` that causes `queue_atomic_flip` to additionally pass `ALLOW_MODESET`?

RAOF (changes requested):
Looks good, works¹ for me, both with just `eglstream-kms` driving the display + `gbm-kms` rendering, and with `MIR_EXPERIMENTAL_HYBRID` so that both amdgpu and nvidia displays are active.

Just a couple of nits, plus some error handling needed.

¹: There's a malloc-corruption crash/EGL_BAD_ALLOC I see with gbm-kms, but I don't think this is related and I'll continue to investigate.

src/platforms/eglstream-kms/server/egl_output.cpp
mattkae: I implemented something that makes a lot of sense, but I get a fatal error whenever I VT switch :thinking:

RAOF (changes requested):
Ok, so this *does* look plausible, but we die with `EPERM` in atomic-commit when VT switching away.

What's happening here is that we've lost DRM master, but are still in the process of scheduling a flip. We should catch `EPERM` in this case or, possibly, not throw an exception *at all* here and just log any failure - that's what gbm-kms does.

mattkae:
@RAOF Dropping this here while I figure it out. We have broken the default path it seems. Running `./bin/miral-shell --platform-display-libs=mir:eglstream-kms --platform-rendering-libs=mir:eglstream-kms` causes a flicker _and_ it prevents us from ever stopping

mattkae:
> Ok, so this _does_ look plausible, but we die with `EPERM` in atomic-commit when VT switching away.
> 
> What's happening here is that we've lost DRM master, but are still in the process of scheduling a flip. We should catch `EPERM` in this case or, possibly, not throw an exception _at all_ here and just log any failure - that's what gbm-kms does.

VT switching is working now :+1:

RAOF:
Ok, a debugging log:
* VT switching away works, but VT switching *back* dies
  * Oh, it's in `WaylandConnector::start`!?
  * *Oh*, it's because the VT switch *away* didn't properly *stop* the `WaylandConnector`, and so there's an already running Wayland dispatch thread?!
  * *Ooooh*, #3087 
  * Ah, if you stop it in gdb before the compositor-shut-down timeout, you find that it's blocking in `DisplayBuffer::post()`.
  * *Aaaah*, because it's waiting for a flip event that will never come, because we failed to actually submit the flip, because we've lost access to the DRM node due to VT switch
  * [Fix that](https://github.com/MirServer/mir/pull/3067/commits/5f43f9086fa095ca3f0648cd5ec695c99dc1abf3) and now… it still fails when VT switching back? But now it does so with `EBUSY`.
  * Make modesetting commits blocking, to work around `EBUSY`?
  * Ok, that gets things mostly working
  * Oh, but *repeatedly* VT switching it looks like sometimes we hit

```
[2023-10-23 17:41:47.476460] < - debug - > mirserver: Received logind force-pause event for device 13:64                                                                                                                                    │
│[2023-10-23 17:41:47.476480] < - debug - > mirserver: Received logind force-pause event for device 13:92                                                                                                                                    │
│[2023-10-23 17:41:47.512323] < CRITICAL! > mirserver: terminate_with_current_exception(): /home/chris/Canonical/Mir/mir/main/src/platforms/eglstream-kms/server/egl_output.cpp(355): Throw in function bool mir::graphics::eglstream::kms::E│
│GLOutput::queue_atomic_flip(const FBHandle &, const void *)                                                                                                                                                                                 │
│Dynamic exception type: boost::wrapexcept<std::system_error>                                                                                                                                                                                │
│std::exception::what: Failed to commit Atomic KMS state: Invalid argument                                                                                                                                                                   │
│                                                                                                                                                                                                                                            │
│[2023-10-23 17:41:47.512411] < - debug - > mirserver: Handling Terminated from pid=632670                                                                                                                                                   │
│[2023-10-23 17:41:47.512695] < -warning- > mirserver: wl_surface@9 destroyed before associated role                                                                                                                                         │
│[2023-10-23 17:41:47.516028] <information> evdev-input: Removed /dev/input/event7: Ducky One 3 TKL RGB                                                                                                                                      │
│ERROR: /home/chris/Canonical/Mir/mir/main/src/platforms/eglstream-kms/server/egl_output.cpp(355): Throw in function bool mir::graphics::eglstream::kms::EGLOutput::queue_atomic_flip(const FBHandle &, const void *)                        │
│Dynamic exception type: boost::wrapexcept<std::system_error>                                                                                                                                                                                │
│std::exception::what: Failed to commit Atomic KMS state: Invalid argument
```

RAOF:
So next steps here:
* Rather than making `ALLOW_MODESET` commits blocking, we should instead return `EBUSY` up to `post()`, and if we hit that, sleep a couple of milliseconds and try again.
* See if the `EINVAL` that *sometimes* shows up still does with a retry-on-`EBUSY` approach. If it does, enable some DRM kernel debugging and see if that gives any clues as to *what* is invalid.

KMS kernel debugging: `echo 0x14 | sudo tee /sys/module/drm/parameters/debug` will (should) enable KMS debugging (`0x4`) and atomic debugging (`0x10`). After enabling that, modesetting should be *very* verbose in `dmesg`.

RAOF:
Alright! Apart from #3089 (which happens on 2.15, too :woman_facepalming:), this appears to entirely work for me now. :tada: 

VT switching with the `CPUAddressableDisplayProvider` works reliably, things work with `MIR_EXPERIMENTAL_HYBRID_GRAPHICS=1`, and the `EGLStreamOutput` only crashes in ways that the current release does!

RAOF (approved):
This is now good to go as far as I'm concerned. The only vaguely concerning thing is that `set_flags_for_next_flip` is thread-unsafe, but this doesn't matter because it's only called from `resume()`, which is called from `DisplayServer::Private::resume()`, which calls it before the compositor thread(s) are started.

I can't really approve this now, having authored a bunch of it, so @AlanGriffiths ?

Saviq:
And I can confirm our smoke tests are fine, too :)

```plain
  job passed   : Run Mir smoke tests
  job passed   : Run Mir smoke tests as root                                                                                       
  job passed   : Run hybrid Mir smoke tests rendering on CometLake-H GT2 [UHD Graphics]
  job passed   : Run hybrid Mir smoke tests rendering on TU117GLM [Quadro T2000 Mobile / Max-Q]
```

src/platforms/common/server/cpu_addressable_fb.cpp
```diff
@@ -53,7 +51,7 @@ class mgg::CPUAddressableFB::Buffer : public mir::renderer::software::RWMappable
             if (::munmap(const_cast<typename std::remove_const<T>::type *>(data_), len_) == -1)
             {
                 // It's unclear how this could happen, but tell *someone* about it if it does!
-                mir::log_error("Failed to unmap CPU buffer: %s (%i)", strerror(errno), errno);
+                mir::log(mir::logging::Severity::error, "Failed to unmap CPU buffer: %s (%i)", strerror(errno), errno);
```
AlanGriffiths, line 54: Why? This doesn't make the code clearer

src/platforms/common/server/cpu_addressable_fb.cpp
```diff
@@ -102,7 +100,7 @@ class mgg::CPUAddressableFB::Buffer : public mir::renderer::software::RWMappable
 
         if (auto const err = drmIoctl(drm_fd, DRM_IOCTL_MODE_DESTROY_DUMB, &params))
         {
-            mir::log_error("Failed destroy CPU-accessible buffer: %s (%i)", strerror(-err), -err);
+            mir::log(mir::logging::Severity::error, "Failed destroy CPU-accessible buffer: %s (%i)", strerror(-err), -err);
```
AlanGriffiths, line 103: Why? This doesn't make the code clearer

src/platforms/common/server/kms_cpu_addressable_display_provider.cpp
```diff
@@ -0,0 +1,63 @@
...
+namespace geom = mir::geometry;
+
+mg::kms::CPUAddressableDisplayProvider::CPUAddressableDisplayProvider(mir::Fd drm_fd)
+    : drm_fd{std::move(drm_fd)},
+      supports_modifiers{drm_get_cap_checked(this->drm_fd, DRM_CAP_ADDFB2_MODIFIERS) == 1}
+{
+    if (!drm_get_cap_checked(this->drm_fd, DRM_CAP_DUMB_BUFFER))
+    {
+        BOOST_THROW_EXCEPTION((std::runtime_error{"DRM device doesn't support CPU buffers"}));
+    }
```
AlanGriffiths, line 49: I think we are checking in the wrong place and reporting by the wrong mechanism.

If the "DRM device doesn't support CPU buffers", then surely we should be returning a `nullptr` from `maybe_create_interface()` not propagating an exception. And to achieve that, we should move the `DRM_CAP_DUMB_BUFFER` check out of the constructor into `maybe_create_interface()`.

src/platforms/eglstream-kms/server/display.cpp
```diff
@@ -130,16 +133,12 @@ class DisplayBuffer
         EGLContext ctx,
         EGLConfig config,
         std::shared_ptr<mge::DRMEventHandler> event_handler,
-        mge::kms::EGLOutput const& output,
+        std::shared_ptr<mge::kms::EGLOutput> output,
```
AlanGriffiths, line 136: Are we intentionally dropping `const` here?

src/platforms/eglstream-kms/server/display.cpp
```diff
@@ -403,11 +486,11 @@ void mge::Display::configure(DisplayConfiguration const& conf)
 {
     auto kms_conf = dynamic_cast<KMSDisplayConfiguration const&>(conf);
     active_sync_groups.clear();
-    kms_conf.for_each_output([this](kms::EGLOutput const& output)
+    kms_conf.for_each_output([this](std::shared_ptr<kms::EGLOutput> const& output)
```
AlanGriffiths, line 489: Shouldn't `for_each_output()` keep the same signature and dereference the `shared_ptr<>`? This is exposing an implementation detail

src/platforms/eglstream-kms/server/egl_output.cpp
```diff
@@ -208,21 +212,15 @@ int mgek::EGLOutput::max_refresh_rate() const
...
+    refresh_connector(drm_fd, connector);
 
     mgk::DRMModeCrtcUPtr crtc;
     mgk::DRMModePlaneUPtr plane;
-
-    refresh_connector(drm_fd, connector);
-
     std::tie(crtc, plane) = mgk::find_crtc_with_primary_plane(drm_fd, connector);
-    auto const crtc_id = crtc->crtc_id;
+    _crtc_id = crtc->crtc_id;
```
AlanGriffiths, line 220: We don't use a `_` prefix. See style guide

src/platforms/eglstream-kms/server/kms_display_configuration.h
```diff
@@ -45,7 +45,7 @@ class KMSDisplayConfiguration : public DisplayConfiguration
     size_t get_kms_mode_index(DisplayConfigurationOutputId id, size_t conf_mode_index) const;
     void update();
 
-    void for_each_output(std::function<void(kms::EGLOutput const&)> const& f) const;
+    void for_each_output(std::function<void(std::shared_ptr<kms::EGLOutput> const&)> const& f) const;
```
AlanGriffiths, line 48: I think we should change the implementation, not the signature and all calling code

AlanGriffiths (changes requested):
I can't really test this, but it looks reasonable.

Some nitpicking to address

src/platforms/eglstream-kms/server/display.cpp
AlanGriffiths: On closer examination, ownership **is** being passed this way

src/platforms/eglstream-kms/server/kms_display_configuration.h
AlanGriffiths: On closer examination, ownership **is** being passed this way

src/platforms/eglstream-kms/server/display.cpp
AlanGriffiths: Yes

AlanGriffiths (approved):
Looks OK now

Not merging as @mattkae is planning on some (hopefully final) testing

Saviq:
Got a test to fix on clang:

```
	183 - mir_unit_tests_gbm-kms.MesaDisplayMultiMonitorTest.* (Failed)
```

AlanGriffiths:
> 183 - mir_unit_tests_gbm-kms.MesaDisplayMultiMonitorTest.* (Failed)

OK, I know _what_ is different, just not (yet) worked out why...

In the clang build `drmModeAddFB2()` is intercepted once, and so only stores `fb_id` once. After which, the `drmModePageFlip()` calls that check the `fb_id` parameter fail (because they are passed `0`).

In the gcc build, `drmModeAddFB2()` is intercepted and stores `fb_id` three times. And the `drmModePageFlip()` calls that get the expected `fb_id` parameter.

Now to figure out how this needs fixing...

mattkae:
@RAOF and @AlanGriffiths: I found these two bugs in my final testing, but feel free to tackle them at a later date if you think they shouldn't block this work:
- https://github.com/MirServer/mir/issues/3098
- https://github.com/MirServer/mir/issues/3097

AlanGriffiths (approved):

PR #3069, opened 2023-10-10
linux_dmabuf: use eglDestroyImageKHR extension

Saviq:
Rather than naked eglDestroyImage.

https://registry.khronos.org/EGL/sdk/docs/man/html/eglDestroyImage.xhtml


AlanGriffiths (approved):
A lot of `importing_provider->dpy` repetition, but sure

PR #3072, opened 2023-10-11
Add a feature flag to hide hybrid capability

AlanGriffiths:
The intent is to allows us to merge without exposing functionality we are still polishing.

This uses an environment variable (`MIR_EXPERIMENTAL_HYBRID_GRAPHICS`), not a configuration option, as that is easy to drop seamlessly later.


AlanGriffiths:
```
::   147 |         platform_modules.resize(std::min(1ul, platform_modules.size()));
::       |                                 ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Rats! I wrote "1uz" and the compiler complained that I was using c++23.

Saviq (approved):
Awesomes.

Issue #3070, opened 2023-10-10
Unable to close miral-shell using keybind via VNC due to `DecorationProvider`


mattkae:
## How to reproduce
1. Enable the following extensions in `shell_main.cpp`:
```c++
WaylandExtensions{}
                .enable(WaylandExtensions::zwlr_screencopy_manager_v1)
                .enable(WaylandExtensions::zwlr_virtual_pointer_manager_v1)
                .enable(WaylandExtensions::zwp_virtual_keyboard_manager_v1),
```
2. Run `miral-shell` with `--platform-display-libs=mir:x11` (although I suspect this happens on all platforms)
3. Open up wayvnc and use `gvncviewer` to interact with your shell
4. Use the key command that you have bound to exit the shell (I used `tab` for easy testing purposes)
5. Note that the process doesn't stop :open_mouth: 

## What's happening?
Mir attempts to join the `InternalClient` thread when the user clicks the right keybinds. This triggers the `DecorationProvider` to deconstruct at some point, which calls `wl_display_roundtrip` in both `BackgroundInfo` and `WaylandApp`. From what I can tell, this is a problem because the wayland thread cannot process any more wayland request because it is blocked waiting for `InternalClient` to join its threads. This leaves `wl_display_roundtrip` waiting around forever.

That is my best guess though!

AlanGriffiths:
> ```c++
> WaylandExtensions{}
>                 .enable(WaylandExtensions::zwlr_screencopy_manager_v1)
>                 .enable(WaylandExtensions::zwlr_virtual_pointer_manager_v1)
>                 .enable(WaylandExtensions::zwp_virtual_keyboard_manager_v1),
> ```

Why not enable on the command line (or in the config):

```
miral-app --add-wayland-extensions=zwlr_screencopy_manager_v1:zwlr_virtual_pointer_manager_v1:zwp_virtual_keyboard_manager_v1
```

> This triggers the `DecorationProvider` to deconstruct at some point, which calls `wl_display_roundtrip` in both `BackgroundInfo` and `WaylandApp`. From what I can tell, this is a problem because the wayland thread


This sounds like we are calling libwayland functions that process the request queue on the wrong thread. (There should be a worker thread for these)

There was a [similar issue with Frame](https://github.com/MirServer/ubuntu-frame/pull/152) recently

AlanGriffiths:
[...]

PR #3074, opened 2023-10-12
Fix MirRunner::stop()

AlanGriffiths:
We were trying to shut down the server on whatever thread was calling MirRunner::stop(). That worked fine for the input thread (for example), but if it were the Wayland thread the server became unresponsive and internal clients couldn't close.

Fixes: #3070


mattkae (approved):
Works swimmingly :dolphin:

PR #3077, opened 2023-10-13
coverage: actually run performance tests

Saviq:


RAOF (approved):
Looks good to me.

We should keep an eye on whether the performance tests fail due to resource starvation (they're not *super* intensive, but who knows how much real CPU the CI runners get?), but they haven't been failing here.

PR #3079, opened 2023-10-13
Better naming for rendering classes, functions and variables

AlanGriffiths:


RAOF (approved):
Yeah, I agree that's better.

PR #3082, opened 2023-10-16
Decoupling writing the display configuration from processing it

mattkae:
## What's new?
- Added a `confirm` to the `DisplayConfigurationPolicy` to handle writing the config to file

### DIFF
--- a/include/platform/mir/graphics/display_configuration_policy.h
+++ b/include/platform/mir/graphics/display_configuration_policy.h
@@ -30,6 +30,7 @@ public:
     virtual ~DisplayConfigurationPolicy() = default;
 
     virtual void apply_to(DisplayConfiguration& conf) = 0;
+    virtual void confirm(DisplayConfiguration const& conf) = 0;
 
 protected:
     DisplayConfigurationPolicy() = default;
--- a/src/include/server/mir/graphics/default_display_configuration_policy.h
+++ b/src/include/server/mir/graphics/default_display_configuration_policy.h
@@ -29,21 +29,24 @@ namespace graphics
 class CloneDisplayConfigurationPolicy : public DisplayConfigurationPolicy
 {
 public:
-    void apply_to(DisplayConfiguration& conf);
+    void apply_to(DisplayConfiguration& conf) override;
+    void confirm(DisplayConfiguration const& conf) override;
 };
 
 /// Each screen placed to the right of the previous one
 class SideBySideDisplayConfigurationPolicy : public DisplayConfigurationPolicy
 {
 public:
-    void apply_to(graphics::DisplayConfiguration& conf);
+    void apply_to(graphics::DisplayConfiguration& conf) override;
+    void confirm(DisplayConfiguration const& conf) override;
 };
 
 /// Just use the first screen
 class SingleDisplayConfigurationPolicy : public DisplayConfigurationPolicy
 {
 public:
-    void apply_to(graphics::DisplayConfiguration& conf);
+    void apply_to(graphics::DisplayConfiguration& conf) override;
+    void confirm(DisplayConfiguration const& conf) override;
 };
 /** @} */
 }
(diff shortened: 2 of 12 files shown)


AlanGriffiths (changes requested):
I don't think this works well.

Given: the user has supplied a .display configuration with multiple layouts
When: the user overwrites the .display-layout configuration file with a layout name
Then:  Mir switches to the chosen layout

But with this the  .display configuration is overwritten with the current layout. The other layouts are dropped and cannot be selected

mattkae:
> I don't think this works well.
> 
> Given: the user has supplied a .display configuration with multiple layouts When: the user overwrites the .display-layout configuration file with a layout name Then: Mir switches to the chosen layout
> 
> But with this the .display configuration is overwritten with the current layout. The other layouts are dropped and cannot be selected

Confirmed, I see. :+1: I will rework

AlanGriffiths:
It doesn't look related, but we have two platforms failing:

```
[ RUN ] RealKMSOutputTest.drm_set_gamma
/spread/mir/tests/unit-tests/platforms/gbm-kms/kms/test_real_kms_output.cpp:516: Failure
Actual function call count doesn't match EXPECT_CALL(mock_drm, drmModeCrtcSetGamma(drm_fd, crtc_ids[0], gamma.red.size(), const_cast<uint16_t*>(gamma.red.data()), const_cast<uint16_t*>(gamma.green.data()), const_cast<uint16_t*>(gamma.blue.data())))...
Expected: to be called once

[ FAILED ] RealKMSOutputTest.drm_set_gamma (1 ms)
```

src/server/symbols.map
```diff
@@ -912,6 +912,8 @@ MIR_SERVER_2.15 {
     mir::shell::SurfaceStackWrapper::send_to_back*;
     mir::shell::ShellWrapper::swap_z_order*;
     mir::shell::ShellWrapper::send_to_back*;
+    mir::Server::the_display_configuration_author*;
+    mir::Server::override_the_display_configuration_author*;
```
AlanGriffiths, line 916: This is the wrong stanza: these functions were **not** introduced in 2.15

Saviq:
> It doesn't look related, but we have two platforms failing:

That's because we're not merging `main` in CI here, b/c it targets `platform-API-merge`, which is a ways behind `main`.

AlanGriffiths (changes requested):
In effect, all that's been done here is to separate out the serialization logic from `ReloadingYamlFileDisplayConfig::apply_to()` and call it directly from `MultiplexingDisplay::MultiplexingDisplay()`. That is clearer, and more correct now we have multiple platforms.

But I suspect that adding a `write()` method to `DisplayConfigurationPolicy` would be simpler and less fragile.

include/platform/mir/graphics/display_configuration_author.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace mir
+{
+namespace graphics
+{
+
+class DisplayConfiguration;
+
+/// Responsible for writing new display configuration files to the filesystem.
+class DisplayConfigurationAuthor
```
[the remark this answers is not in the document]
mattkae: Scribe is nice :)

include/platform/mir/graphics/display_configuration_policy.h
```diff
@@ -30,6 +30,7 @@ class DisplayConfigurationPolicy
     virtual ~DisplayConfigurationPolicy() = default;
 
     virtual void apply_to(DisplayConfiguration& conf) = 0;
+    virtual void confirm(DisplayConfiguration& conf) = 0;
```
AlanGriffiths, line 33: I think `conf` should be `const`

src/include/server/mir/server.h
AlanGriffiths, line 1: We don't actually need to touch this file

AlanGriffiths (changes requested):
I think we can clean up the commit history

AlanGriffiths (approved):
OK

PR #3083, opened 2023-10-16
Better (IMO) names for DisplayInterfaceBase and associates

AlanGriffiths:


RAOF:
Ok. `DisplayInterfaceBase` -> `DisplayProvider` is just unambigously better. I'll use that name from here on in.

I'm not happy with `DisplayTarget`, perhaps because the `DisplayBuffer` is already more of a display target? What `DisplayTarget` provides is a way of allocating something that can be submitted to the `DisplayBuffer`.

Actually, maybe that's the solution? Most of the `DisplayProvider`s are kinda providing *two* services - some sort of global state hookup (`GBMDisplayProvider::gbm_device()`, `EGL*Provider::get_egl_display()`), and some sort of allocator (`GBMDisplayProvider::make_surface()`, `EGLStreamDisplayProvider::claim_stream()` etc).

If we separate those concerns out, maybe we can have `DisplayBuffer::get_allocator<AllocatorType>()` and `DisplayPlatform::get_provider<ProviderType>()` and entirely remove `DisplayTarget`.

RAOF:
A quick spike suggests this is feasible, if that seems like better naming to you.

AlanGriffiths:
> A quick spike suggests this is feasible, if that seems like better naming to you.

Sure, I struggled with a good name for `DisplayTarget` and just went with the fact that the existing code called `DisplayInterfaceProvider` "target" more than any other name. There being no good name is suggestive that there's no clear concept that it implements.

AlanGriffiths:
Have landed the uncontentious bits

PR #3085, opened 2023-10-19
Hold the rendering platforms in description->name order

AlanGriffiths:
Make the choice of buffer allocator determanistic and prefer "eglstream" over "gbm".

I think this resolves the problem as I understand it, but I don't have easy access to hardware to verify my understanding.


Saviq (approved):
Works.

PR #3086, opened 2023-10-20
Support for not compositing until at least one renderable has been sent to the compositor

mattkae:
## To test
1. Checkout https://github.com/MirServer/mir/pull/3086 and run `sudo make install`
2. Checkout https://github.com/MirServer/ubuntu-frame/pull/157 and build it
3. From a VT, run plymouth
```sh
sudo plymouthd
sudo plymouth show-splash
sudo plymouth deactivate
```
The splash should still be on the screen
4. Run frame:
```
./frame --wallpaper=false
```
5. Wait however long you like and note that you still see the splash
6. Run some other app (e.g. gedit) and note that the splash disappears

## What's new?
- `DefaultDisplayBufferCompositor::composite` now returns a boolean telling the caller if any compositing happened or not
- If the `MultithreadedCompositor` needs compositing, a `post` will happen, otherwise nothing will happen

### DIFF
--- a/src/include/server/mir/compositor/display_buffer_compositor.h
+++ b/src/include/server/mir/compositor/display_buffer_compositor.h
@@ -29,7 +29,8 @@ class DisplayBufferCompositor
 public:
     virtual ~DisplayBufferCompositor() = default;
 
-    virtual void composite(SceneElementSequence&& scene_sequence) = 0;
+    /// Returns true if any compositing happened, otherwise false.
+    virtual bool composite(SceneElementSequence&& scene_sequence) = 0;
 
 protected:
     DisplayBufferCompositor() = default;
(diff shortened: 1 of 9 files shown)


Saviq:
Empirically, this behaves fine, the only difference seems to be that e.g. `mir_demo_server` will not draw until there's a renderable, and may seem stuck?

Couple tests to fix:
```
The following tests FAILED:
	 55 - mir_integration_tests.SurfaceStackCompositor.* (Failed)
	110 - mir_unit_tests.DefaultDisplayBufferCompositor.* (Failed)
```

mattkae:
> Couple tests to fix:
> 
> ```
> The following tests FAILED:
> 	 55 - mir_integration_tests.SurfaceStackCompositor.* (Failed)
> 	110 - mir_unit_tests.DefaultDisplayBufferCompositor.* (Failed)
> ```

There are quite a few tests here so I'll fix them once we know that this is the right solution. One question I have is that do we need to `Renderer::suspend` in this case as well?

AlanGriffiths:
This approach looks sensible.

The failing tests are partly because we've changed the expected behaviour, but more because they are bad tests (testing the implementation works the way it is implemented, not the requirements, and testing more than one behaviour in a test).

It may be worth reworking the tests as a pre-requisite

AlanGriffiths (approved):
Sure

Issue #3088, opened 2023-10-23
`--platform-*-libs` chokes on comma-separated list


Saviq:
Debugging #3056 I found that we broke support for comma-separated list of platforms:

```
$ env -u WAYLAND_DISPLAY mir-test-tools.demo-server --platform-display-libs help
# …
Detected modules are: 
        mir:eglstream-kms
        mir:gbm-kms
        mir:wayland
        mir:egl-generic
        mir:x11
Failed to find:
        help

$ env -u WAYLAND_DISPLAY mir-test-tools.demo-server --platform-display-libs mir:x11,mir:wayland
[2023-10-23 19:47:26.716049] <information> mirserver: Starting
[2023-10-23 19:47:26.716112] < - debug - > mirserver: Not trying logind: "DISPLAY" is set and X need not have claimed the VT
[2023-10-23 19:47:26.716152] < - debug - > mirserver: Not using Linux VT subsystem for session management: Failed to find the current VT
[2023-10-23 19:47:26.716160] < - debug - > mirserver: No session management supported
[2023-10-23 19:47:26.716167] <information> VT switch key handler: No VT switching support available: MinimalConsoleServices does not support VT switching
[2023-10-23 19:47:26.716225] <information> mircommon: Loading modules from: /snap/mir-test-tools/12609/usr/lib/x86_64-linux-gnu/mir/server-platform
[2023-10-23 19:47:26.716249] <information> mircommon: Loading module: /snap/mir-test-tools/12609/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-eglstream-kms.so.20
[2023-10-23 19:47:26.716258] <information> mircommon: Loading module: /snap/mir-test-tools/12609/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-gbm-kms.so.20
[2023-10-23 19:47:26.716262] <information> mircommon: Loading module: /snap/mir-test-tools/12609/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-wayland.so.20
[2023-10-23 19:47:26.716265] <information> mircommon: Loading module: /snap/mir-test-tools/12609/usr/lib/x86_64-linux-gnu/mir/server-platform/renderer-egl-generic.so.20
[2023-10-23 19:47:26.716270] <information> mircommon: Loading module: /snap/mir-test-tools/12609/usr/lib/x86_64-linux-gnu/mir/server-platform/server-x11.so.20
[...]

PR #3090, opened 2023-10-24
Fix processing of `--platform-*-libs`

AlanGriffiths:
Fixes: #3088


Saviq (approved):
Yups.

I was tempted to talk about ordering, but that would be papering over things we need to fix.

PR #3094, opened 2023-10-24
Continue building debs in ppa for mantic

herdwick:


Saviq:
Thanks @herdwick, I've gone ahead and added some CI-related changes as well.

PR #3099, opened 2023-10-26
graphics: drop MIR_EXPERIMENTAL_HYBRID_GRAPHICS and use the virtual platform in CI

Saviq:


Saviq:
Needed to fix:
- #3102
- #3101

AlanGriffiths:
>     * [The compositor report doesn't work on the virtual platform #3101](https://github.com/MirServer/mir/issues/3101)

I think we can drop this test, it isn't relevant anymore

Saviq:
> I think we can drop this test, it isn't relevant anymore

Well, it's as relevant as it was before? Just not on the virtual platform? :)

Would still be good to run it to include its coverage, but yeah that's not a priority.

AlanGriffiths:
> Well, it's as relevant as it was before? Just not on the virtual platform? :)

It checks that we don't exhaust the frontend thread pool by starting "many" clients. That hasn't been since we dropped mirclient (and not actually possible for a long time before that).

AlanGriffiths (approved):
No known blockers now

Issue #3103, opened 2023-10-26
Display scaling is broken post platform refactor


Saviq:
It doesn't crash on Qemu, but still doesn't work correctly:

![imagen](https://github.com/MirServer/mir/assets/105285003/21bbc5b9-b50f-46a3-9491-770a976c21e0)

_Originally posted by @sergio-costas in https://github.com/MirServer/mir/issues/3095#issuecomment-1781180159_

Saviq:
Another failure mode is cropping:

![obraz](https://github.com/MirServer/mir/assets/231077/77dfefee-c73d-4967-85ee-f96fad3faddf)
![obraz](https://github.com/MirServer/mir/assets/231077/ba63e508-56df-4335-9c32-c6da0cc8a71c)

Saviq:
A simple `miral-shell --display-scale=`… makes it clear it's not an auto-scale problem.

Also using a `--display-config=static=` exhibits the same.

Using a scale > 1 doesn't render at all, and with < 1 things are rendered wrong.

mattkae:
**Live Thoughts**
I am focusing on the gbm-kms platform for testing
- The `area` sent to `DisplayBuffer` matches the size provider. So if we're on a 1980x1080 resolution with a scale of 0.75, we'll get 2560x1440.

AlanGriffiths:
@mattkae your comment sounds confused.

1. The number of pixels on the screen always matches the display buffer because physics. So on a 1920x1080 output that must be the size of the display buffer
2. Wayland only supports integer scaling. So for any scaling not greater than 1 the client will see a scale of 1 (and the compositor handles the rest. Similarly between 1 and 2, the scale will be reported as 2

The only way any scale comes into the size of the screen is that if we are hosted on Wayland and an output is scaled by the host, we need to know the scale to work out the number of real pixels and to submit buffers of the correct scale. (But that host scale has nothing to do with our scaling,)

mattkae:
> @mattkae your comment sounds confused.
> 
>     1. The number of pixels on the screen always matches the display buffer because physics. So on a 1920x1080 output that must be the size of the display buffer
> 
>     2. Wayland only supports integer scaling. So for any scaling not greater than 1 the client will see a scale of 1 (and the compositor handles the rest. Similarly between 1 and 2, the scale will be reported as 2
> 
> 
[...]

PR #3104, opened 2023-10-27
platform/DisplayBuffer: Add pixel_size() accessor.

RAOF:
We need this when allocating `Framebuffer`s.

Fixes: #3103

### DIFF
--- a/include/platform/mir/graphics/display_buffer.h
+++ b/include/platform/mir/graphics/display_buffer.h
@@ -58,6 +58,9 @@ public:
     /** The area the DisplayBuffer occupies in the virtual screen space. */
     virtual geometry::Rectangle view_area() const = 0;
 
+    /** The size in pixels of the underlying display */
+    virtual auto pixel_size() const -> geometry::Size = 0;
+
     /** This will render renderlist to the screen and post the result to the 
      *  screen if there is a hardware optimization that can be done.
      *  \param [in] renderlist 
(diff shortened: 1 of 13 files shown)


RAOF:
I think this can be better done with some wider changes to the interfaces; the better-naming branch will explore that.

But this can (a) validate the hypothesis that I know what the problem is, and (b) serve as a fixed baseline.

Saviq (approved):
Empirically, this fixes \o/

src/platforms/gbm-kms/server/kms/display_buffer.cpp
```diff
@@ -101,6 +101,12 @@ geom::Rectangle mgg::DisplayBuffer::view_area() const
     return area;
 }
 
+auto mgg::DisplayBuffer::pixel_size() const -> geom::Size
+{
+    // All the outputs (are meant to) have the same size; KMSOutput::size() gives the size in pixels
```
AlanGriffiths, line 106: [non-blocking] This is a consequence of only sharing a display buffer between compatible outputs. Sounds like it can be asserted somewhere?

include/platform/mir/graphics/display_buffer.h
```diff
@@ -58,6 +58,9 @@ class DisplayBuffer
     /** The area the DisplayBuffer occupies in the virtual screen space. */
     virtual geometry::Rectangle view_area() const = 0;
 
+    /** The size in pixels of the underlying display */
+    virtual auto pixel_size() const -> geometry::Size = 0;
```
AlanGriffiths, line 62: I'm not totally sold on the name, it is ambiguous, "size of pixel" vs "size in pixels".

AlanGriffiths (approved):

PR #3105, opened 2023-10-27
Refactored wayland_client.c to support the virtual platform

mattkae:
## What's new?
- Fixes https://github.com/MirServer/mir/issues/3102
- Rendering in the same manner as ubuntu-frame does it so that everything works with the virtual platform

## To test:
Run this and see that we don't block any longer:
```sh
WAYLAND_DISPLAY=wayland-99 MIR_SERVER_PLATFORM_DISPLAY_LIBS=mir:virtual MIR_SERVER_VIRTUAL_OUTPUT=1280x1024 ./bin/mir_demo_server --test-timeout=3 --test-client ./bin/mir_demo_client_wayland
```

### DIFF
--- a/examples/client/wayland_client.c
+++ b/examples/client/wayland_client.c
@@ -27,6 +27,8 @@
 #include <stdio.h>
 #include <stdbool.h>
 #include <signal.h>
+#include <poll.h>
+#include <sys/eventfd.h>
 
 #include <wayland-client.h>
 #include <wayland-client-core.h>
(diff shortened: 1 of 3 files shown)


examples/client/wayland_client.c
```diff
@@ -482,6 +488,70 @@ static struct xdg_toplevel_listener const shell_toplevel_listener =
     handle_xdg_toplevel__capabilities,
 };
 
+void run(struct wl_display* display, int* shutdown_signal, int* draw_signal, draw_context* ctx)
```
AlanGriffiths, line 491: As [g_]shutdown_signal and [g_]draw_signal are global, why pass them as parameters?

AlanGriffiths (approved):
Yes.

Feel free to deal with the nits or not before landing

examples/client/wayland_client.c
mattkae: This was my minor effort to make it less global-ly. But that might be a losing battle, I'm okay with using globals

examples/client/wayland_client.c
```diff
@@ -218,11 +220,18 @@ find_free_buffer(draw_context* ctx)
     return NULL;
 }
 
-static void draw_new_stuff(void* data, struct wl_callback* callback, uint32_t time);
+static int g_draw_signal;
```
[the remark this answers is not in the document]
mattkae: Eh I didn't want to shadow the locals, but if I don't need them, then I'll get rid of them

PR #3106, opened 2023-10-27
Fixing the platform-input-lib parameter so that it behaves how you would expect it to

mattkae:
fixes https://github.com/MirServer/mir/issues/3073

## What's new?
- You can now specify the input platform using the name instead of an absolute path to the `.so`
- The input platform. with graphics attached to it can now be overridden by a manually specified input platform. For example, you were unable to make the X11 graphics platform use an evdev input platform before, but now you can
- Simplified the logic a bit, as we had some duplicate code in there

## To test
You can run `miral-shell` with any of the following values for `--platform-input-lib` and they will work (or try to work) in all scenarios:
- mir:wayland
- mir:x11-input
- mir:evdev-input
- mir:stub-input

This is nice because you can - for example - test evdev input on an X11 platform (it's weird, but it works)


src/server/input/input_probe.cpp
```diff
@@ -49,6 +49,46 @@ mir::UniqueModulePtr<mi::Platform> create_input_platform(
...
+
+auto select_platforms_from_list(
+    std::string const& selection,
+    std::vector<std::shared_ptr<mir::SharedLibrary>> const& modules
+) -> std::shared_ptr<mir::SharedLibrary>
+{
+    std::vector<std::shared_ptr<mir::SharedLibrary>> selected_modules;
+    std::vector<std::string> found_module_names;
+
+    // Our platform modules are a comma-delimited list.
```
AlanGriffiths, line 62: No it isn't

AlanGriffiths (approved):
@Saviq I've added a couple of commits here. Could you sanity check?

src/server/input/input_probe.cpp
```diff
@@ -49,6 +49,44 @@ mir::UniqueModulePtr<mi::Platform> create_input_platform(
 
     return result;
 }
+
+
+auto select_platforms_from_list(
```
Saviq, line 54: Could we refactor this together with the `default_configuration.cpp` part to avoid the copy-paste?

src/server/input/input_probe.cpp
```diff
@@ -88,32 +126,49 @@ mir::UniqueModulePtr<mi::Platform> mi::probe_input_platforms(
             return Selection::persist;
         };
 
+    // First, check if any already-loaded graphics modules double as an input platform
```
Saviq, line 129: Shouldn't this go after the manually-selected ones instead?

mattkae: We first search to see if our graphics platform doubles as an input platform _and_ it is manually selected. If that's the case, we can just return the already loaded platform (as opposed to loading it again when we try and get it manually).

Otherwise, we check for a manually loaded platform. And, if not that, we get the best possible platform.

src/server/input/input_probe.cpp
Saviq: Most of the body here is the same?

https://github.com/MirServer/mir/blob/7f057531eed603203b925cf9af201c58ff46c163/src/server/graphics/default_configuration.cpp#L81-L131

src/server/input/input_probe.cpp
Saviq: In that case, the comment below is wrong?

>     // Otherwise, try and load a manually specified platform, as that takes precedence over the loaded platform

It doesn't take precedence, then? But shouldn't it?

src/server/input/input_probe.cpp
mattkae: @Saviq There is some _slightly_ different logic when it comes to graphics. On the input side, we always only select one platform, while for graphics we select multiple potentially

mattkae: We could share it, but I feel it would just complicate things more than it needs to

Saviq (approved):
LGTM!

AlanGriffiths (approved):
LGTM

PR #3114, opened 2023-11-01
Platform: Better names and responsibilities for platform types

RAOF:
Currently, most of the `DisplayProvider`s have two responsibilities: providing some global, platform-wide, state, and some sort of allocate-a-display-surface function. That is part of the reason why `DisplayProvider` has been difficult to name.

This splits these concerns out into `DisplayProvider`, providing the global state, and `DisplayAllocator` providing the `DisplayBuffer`-specific allocation services.

This split also allows us to hang the `DisplayAllocator` directly off the `DisplayBuffer`, rather than having a shared intermediate `DisplayInterfaceProvider`.

The new `DisplayBuffer::acquire_allocator<>` continues our tradition of hanging object lifetimes off that of the `DisplayBuffer`, simplifying the interface.


RAOF:
There's some natural follow-up work to this:
* Now that `DisplayAllocator`s are more directly connected to the `DisplayBuffer`, we can eliminate the `pixel_size()` accessor *and* the `size` parameter to `GLRenderingPlatform::surface_for_output`, as the `DisplayAllocator` can know the only correct value to use there
* It's possible we want to rename `DisplayBuffer`, possibly to `DisplayTarget`? `DisplayBuffer` is no longer anything like a buffer.

AlanGriffiths (changes requested):
On a skim read looks a sensible approach. Flagged one comment I think is confusing.

Will take a deeper look when I have time

Saviq:
> Flagged one comment I think is confusing.

Nuh uh! Must've been in a packet that was lost.

include/platform/mir/graphics/display_buffer.h
```diff
@@ -98,8 +99,56 @@ class DisplayBuffer
...
+     *                      The lifetime of this Allocator implementation is bound
+     *                      to that of the parent DisplayBuffer.
+     *          On failure: nullptr
+     */
+    template<typename Allocator>
+    auto acquire_allocator() -> Allocator*
+    {
+        static_assert(
+            std::is_convertible_v<Allocator*, DisplayAllocator*>,
+            "Can only acquire a DisplayProvider; Provider must implement DisplayAllocator");
```
AlanGriffiths, line 122: `DisplayProvider`?

AlanGriffiths (approved):
Let's talk about the other renames when you wake up

PR #3115, opened 2023-11-04
Packaging: Add new -dev dependencies to the relevant packages

RAOF:
We use these packages in the headers, so include them in the `-dev` dependencies.


AlanGriffiths (approved):
LGTM

AlanGriffiths (approved):
Ack

PR #3124, opened 2023-11-10
Make egl-generic the rendering platform of last resort

AlanGriffiths:
Alternative to #3122


mattkae (approved):
I like this one better :+1:

PR #3130, opened 2023-11-16
Add --mir-version command line option. solves MirServer/mir#3058

pillowtrucker:
Tested with miral-shell built from within this repo and with my own compositor linking against this version.
I also looked in unit tests to see if I could add it there, but the existing test for options uses test data rather than real responses from built artifacts. Please let me know if there's another one that I've missed that would be suitable for this.

### DIFF
--- a/src/platform/options/default_configuration.cpp
+++ b/src/platform/options/default_configuration.cpp
@@ -295,7 +295,7 @@ void mo::DefaultConfiguration::parse_arguments(
     {
         desc.add_options()
             ("help,h", "this help text");
-		desc.add_options()
+        desc.add_options()
             ("version,V", "display Mir version and exit");
 
         options.parse_arguments(desc, argc, argv);
@@ -306,11 +306,11 @@ void mo::DefaultConfiguration::parse_arguments(
             tokens.push_back(token.c_str());
         if (!tokens.empty()) unparsed_arguments_handler(tokens.size(), tokens.data());
         if (options.is_set("version"))
-		{
-			std::ostringstream mir_version;
-			mir_version << MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << std::endl;
-			BOOST_THROW_EXCEPTION(mir::ExitWithOutput(the_version_of_mir.str()));
-		}
+        {
+            std::ostringstream mir_version;
+            mir_version << MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << std::endl;
+            BOOST_THROW_EXCEPTION(mir::ExitWithOutput(mir_version.str()));
+        }
         if (options.is_set("help"))
         {
             std::ostringstream help_text;
--- a/src/platform/options/default_configuration.cpp
+++ b/src/platform/options/default_configuration.cpp
@@ -295,7 +295,7 @@ void mo::DefaultConfiguration::parse_arguments(
     {
         desc.add_options()
             ("help,h", "this help text");
-		desc.add_options()
+        desc.add_options()
             ("version,V", "display Mir version and exit");
 
         options.parse_arguments(desc, argc, argv);
@@ -305,17 +305,19 @@ void mo::DefaultConfiguration::parse_arguments(
         for (auto const& token : unparsed_arguments)
             tokens.push_back(token.c_str());
         if (!tokens.empty()) unparsed_arguments_handler(tokens.size(), tokens.data());
-        if (options.is_set("version"))
-		{
-			std::ostringstream mir_version;
-			mir_version << MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << std::endl;
-			BOOST_THROW_EXCEPTION(mir::ExitWithOutput(the_version_of_mir.str()));
-		}
-        if (options.is_set("help"))
-        {
-            std::ostringstream help_text;
-            help_text << desc;
-            BOOST_THROW_EXCEPTION(mir::ExitWithOutput(help_text.str()));
+        
+        // Handle one-off options such as `help` or `version`
+        std::map<const char *,std::function<std::ostringstream & (std::ostringstream &)>> oneoff_options {
+            {"version",[&](std::ostringstream & oneoff_output) -> std::ostringstream & {oneoff_output <<MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << std::endl; return oneoff_output;}},
+            {"help",[&](std::ostringstream & oneoff_output) -> std::ostringstream & {oneoff_output << desc; return oneoff_output;}}
+        };
+        for(auto oopt : oneoff_options) {
+            if(options.is_set(oopt.first))
+            {
+                std::ostringstream oneoff_output;
+                // Intentionally throw exception for non-error condition as per ExitWithOutput documentation.
+                BOOST_THROW_EXCEPTION(mir::ExitWithOutput(oopt.second(oneoff_output).str()));
+            }
         }
     }
     catch (po::error const& error)
--- a/src/platform/options/default_configuration.cpp
+++ b/src/platform/options/default_configuration.cpp
@@ -296,7 +296,7 @@ void mo::DefaultConfiguration::parse_arguments(
         desc.add_options()
             ("help,h", "this help text");
 		desc.add_options()
-            ("mir-version", "display MIR platform version");
+            ("version,V", "display Mir version and exit");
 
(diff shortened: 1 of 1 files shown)


src/platform/options/default_configuration.cpp
```diff
@@ -295,6 +295,8 @@ void mo::DefaultConfiguration::parse_arguments(
     {
         desc.add_options()
             ("help,h", "this help text");
+		desc.add_options()
+            ("mir-version", "display MIR platform version");
```
Saviq, line 299:
```suggestion
            ("version,V", "display Mir version and exit");
```

To be frank, off the top of my head I don't recall whether case is differentiated on. `-v` is usually used for `--verbose`, so this will be better, if we can't do `-V`:

```suggestion
            ("version", "display Mir version and exit");
```

src/platform/options/default_configuration.cpp
```diff
@@ -303,7 +305,12 @@ void mo::DefaultConfiguration::parse_arguments(
         for (auto const& token : unparsed_arguments)
             tokens.push_back(token.c_str());
         if (!tokens.empty()) unparsed_arguments_handler(tokens.size(), tokens.data());
-
+        if (options.is_set("mir-version"))
+		{
+			std::ostringstream the_version_of_mir;
+			the_version_of_mir << "MIR Platform version " << MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << "." << std::endl;
```
Saviq, line 311:
```suggestion
        if (options.is_set("version"))
		{
			std::ostringstream mir_version;
			mir_version << MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << std::endl;
```

src/platform/options/default_configuration.cpp
```diff
@@ -303,7 +305,12 @@ void mo::DefaultConfiguration::parse_arguments(
...
+			BOOST_THROW_EXCEPTION(mir::ExitWithOutput(the_version_of_mir.str()));
```
Saviq, line 312: This shouldn't be throwing an exception, a clean exit is appropriate when the ask is to show the version string.

Saviq (changes requested):
Hi @pillowtrucker thanks for this!

Some suggestions inline.

src/platform/options/default_configuration.cpp
pillowtrucker: Updated to just print to stdout and exit cleanly. It seemed too simple to be true, but I can't think of a better way and it seems to shut down cleanly.

pillowtrucker:
Replied in the remaining thread. I'm also not sure if there was a good reason why `--help` was also throwing an exception, so I've flattened it in the same way. The 0 return code makes my shell happier.

src/platform/options/default_configuration.cpp
```diff
@@ -303,19 +305,27 @@ void mo::DefaultConfiguration::parse_arguments(
...
+			std::cout << help_text.str();
+			exit(0);
         }
     }
     catch (po::error const& error)
     {
         std::ostringstream help_text;
         help_text << "Failed to parse command line options: " << error.what() << "." << std::endl << desc;
         BOOST_THROW_EXCEPTION(mir::AbnormalExit(help_text.str()));
+
```
Saviq, line 328: Please fix indentation here and drop the extra newline.

src/platform/options/default_configuration.cpp
```diff
@@ -303,19 +305,27 @@ void mo::DefaultConfiguration::parse_arguments(
...
             tokens.push_back(token.c_str());
         if (!tokens.empty()) unparsed_arguments_handler(tokens.size(), tokens.data());
-
+        if (options.is_set("version"))
+		{
+			std::ostringstream mir_version;
+			mir_version << MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << std::endl;
+			std::cout << mir_version.str();
+			exit(0);
+		}
```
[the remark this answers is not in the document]
Saviq: And we can probably get rid of `mir_version` altogether now?

src/platform/options/default_configuration.cpp
```diff
@@ -303,19 +305,26 @@ void mo::DefaultConfiguration::parse_arguments(
...
+            std::cout << mir_version.str();
+            exit(0);
+        }
         if (options.is_set("help"))
         {
             std::ostringstream help_text;
             help_text << desc;
-            BOOST_THROW_EXCEPTION(mir::ExitWithOutput(help_text.str()));
+            std::cout << help_text.str();
+            exit(0);
```
AlanGriffiths, line 320: Obviously, most things ought to be cleaned up on exit, but I don't think there are enough guarantees about the state of the program when these options are processed.

It is cleaner to unwind the stack using the `mir::ExitWithOutput` exception that is intended for this purpose.

AlanGriffiths (changes requested):

src/platform/options/default_configuration.cpp
pillowtrucker: Also tried to remove most duplicate logic that I could without making it entirely incomprehensible

pillowtrucker:
pushed new rebased version ignoring the stray commits from before, added a comment about the thrown exceptions and tried to DRY it a little bit

src/platform/options/default_configuration.cpp
```diff
@@ -303,12 +305,19 @@ void mo::DefaultConfiguration::parse_arguments(
...
+            {"version",[&](std::ostringstream & oneoff_output) -> std::ostringstream & {oneoff_output <<MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << std::endl; return oneoff_output;}},
+            {"help",[&](std::ostringstream & oneoff_output) -> std::ostringstream & {oneoff_output << desc; return oneoff_output;}}
+        };
+        for(auto oopt : oneoff_options) {
+            if(options.is_set(oopt.first))
+            {
+                std::ostringstream oneoff_output;
+                // Intentionally throw exception for non-error condition as per ExitWithOutput documentation.
+                BOOST_THROW_EXCEPTION(mir::ExitWithOutput(oopt.second(oneoff_output).str()));
+            }
```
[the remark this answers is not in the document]
pillowtrucker: Fair enough. I felt like I was having a bit too much fun there ;)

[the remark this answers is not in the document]
AlanGriffiths: I was having fun reviewing that! I think their may be value anyway. So I'll add here...

I think the map setup is more verbose and harder to follow than the twofold repetition that is the alternative.

Also, the use of whitespace isn't as we do it in Mir.

1. "&" (and "*") go with the type
2. braces (except single line blocks) go on their own line
3. braced blocks have a space after/before the brace
4. We don't like lines longer than 120 characters

I tried simplifying the loop a bit, but it still looks unwieldy:

```c++
        // Handle one-off options such as `help` or `version`
        std::map<const char*, std::function<void(std::ostream&)>> oneoff_options
        {
            {
                "version",
                [&](auto& out) { out << MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO; }
            },
            {
                "help",
                [&](auto& out) { out << desc; }
            }
        };
        for(auto oopt : oneoff_options) 
        {
            if(options.is_set(oopt.first))
            {
                std::ostringstream oneoff_output;
                oopt.second(oneoff_output);
                // Intentionally throw exception for non-error condition as per ExitWithOutput documentation.
                BOOST_THROW_EXCEPTION(mir::ExitWithOutput(oneoff_output.str()));
            }
        }
```

But I still think unrolling the loop would be easier to follow

AlanGriffiths (approved):
Personally, I'd leave blank lines around the `if` block to give the reader clearer code "blocks". But not everyone on the project feels the same

src/platform/options/default_configuration.cpp
```diff
@@ -303,7 +305,12 @@ void mo::DefaultConfiguration::parse_arguments(
         for (auto const& token : unparsed_arguments)
             tokens.push_back(token.c_str());
         if (!tokens.empty()) unparsed_arguments_handler(tokens.size(), tokens.data());
-
+        if (options.is_set("version"))
+        {
+            std::ostringstream mir_version;
+            mir_version << MIR_VERSION_MAJOR << "." << MIR_VERSION_MINOR << "." << MIR_VERSION_MICRO << std::endl;
+            BOOST_THROW_EXCEPTION(mir::ExitWithOutput(mir_version.str()));
+        }
```
Saviq, line 313: As discussed, please move it down below `help` and that'll be it :)

Saviq (approved):
Thank you!

Issue #3116, opened 2023-11-04
Libreoffice Impress slideshow doesn't fullscreen correctly 


AlanGriffiths:

mattkae:
From the logs, libreoffice impress seems to be doing `xdg_toplevel.set_max_size` to a width and height that are slightly less than what it should be (e.g. the width and height are 800x600, but it is setting the max size to 748x548). I have no clue why just yet :smile:

AlanGriffiths:
The immediate problem isn't having a max size set, but that the client is committing buffers of the smaller size. This affects both the "presentation" window and the "presenter's view" window. Here's the first:

```plain
[2023-11-15 17:53:53.427080] <information> miral::Window Management: place_new_window app_info={application=, windows={Miriway - bringing Wayland to your desktop.odp - LibreOffice Impress}}, requested_specification={name=Presenting: Miriway - bringing Wayland to your desktop, type=freestyle, top_left=0, 0, size=(1920, 1600), output_id=2, state=fullscreen} -> {name=Presenting: Miriway - bringing Wayland to your desktop, type=freestyle, top_left=0, 0, size=(1920, 1600), output_id=2, state=fullscreen}
[2023-11-15 17:53:53.427187] <information> miral::Window Management: advise_new_window window_info={name=Presenting: Miriway - bringing Wayland to your desktop, type=freestyle, state=fullscreen, top_left=0, 0, size=(1920, 1600), restore_rect=(480, 400, (960, 800)), children={}, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0, output_id=2}
[2023-11-15 17:53:53.427282] <information> miral::Window Management: handle_modify_window window_info={name=Presenting: Miriway - bringing Wayland to your desktop, type=freestyle, state=fullscreen, top_left=0, 0, size=(1920, 1600), restore_rect=(480, 400, (960, 800)), children={}, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0, output_id=2}, modifications={}
[2023-11-15 17:53:53.427299] <information> miral::Window Management: modify_window window_info={name=Presenting: Miriway - bringing Wayland to your desktop, type=freestyle, state=fullscreen, top_left=0, 0, size=(1920, 1600), restore_rect=(480, 400, (960, 800)), children={}, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0, output_id=2}, modifications={}
[...]

PR #3132, opened 2023-11-16
Don't allow fullscreen surfaces to be resized

AlanGriffiths:
Fixes: #3116

See also #3040, as the fix for that introduced this problem.


AlanGriffiths:
> From the analysis of #3040, I think this also works there. But will leave as "draft" pending more testing.

Unfortunately, this re-introduces  #3040, more thinking required...

AlanGriffiths:
OK, now a different approach

mattkae (approved):
Worked for me!

PR #3143, opened 2023-11-19
Remove TODO from 2015 to detect host refresh rate/max fps for Mir-on-X client.

pillowtrucker:
There's a few sticky things here:
- First of all, I'm adding a new dependency. However, xcb-randr should already exist anywhere xcb is installed, as far as I know. On my distribution this is not even a separate package. I've added the relevant stanza for debian/ubuntu in the control file (commented out for now), but, if I remember debian packaging correctly, this won't apply until the version is bumped in the changelog. 
- I'm following the general trend of ignoring xcb errors. They're usually non-fatal.
- If the randr extension version of the host X11 is <1.2, this will probably either not work or return nonsense. OTOH, if someone is doing development or testing of Mir-on-X on an Xorg that doesn't support 1.2+, they're probably doing it deliberately.
- More importantly, should I also follow [this guide](https://github.com/MirServer/mir/blob/c1d267021c9c13798d161d7ca73aad9bd4b93bbe/doc/sphinx/dso_versioning_guide.md?plain=1#L106) to make a versioned `mgx::DisplayConfiguration::build_output` ? Would I also have to add it to the relevant symbols.map, first ?

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -45,7 +45,6 @@ Build-Depends: cmake,
                python3-venv:native,
                libxcb-composite0-dev,
                libxcb-xfixes0-dev,
-# iirc debian/changelog also needs to be updated for the package to get rebuilt/upgraded with xcb-randr support
                libxcb-randr0-dev,
                libxcb-render0-dev,
                libxcb-composite0-dev,
--- a/debian/control
+++ b/debian/control
@@ -46,7 +46,7 @@ Build-Depends: cmake,
                libxcb-composite0-dev,
                libxcb-xfixes0-dev,
 # iirc debian/changelog also needs to be updated for the package to get rebuilt/upgraded with xcb-randr support
-#               libxcb-randr0-dev,
+               libxcb-randr0-dev,
                libxcb-render0-dev,
                libxcb-composite0-dev,
                libx11-xcb-dev,
--- a/debian/control
+++ b/debian/control
@@ -45,6 +45,8 @@ Build-Depends: cmake,
                python3-venv:native,
                libxcb-composite0-dev,
                libxcb-xfixes0-dev,
+# iirc debian/changelog also needs to be updated for the package to get rebuilt/upgraded with xcb-randr support
+#               libxcb-randr0-dev,
                libxcb-render0-dev,
                libxcb-composite0-dev,
                libx11-xcb-dev,
--- a/src/platforms/x11/x11_resources.cpp
+++ b/src/platforms/x11/x11_resources.cpp
@@ -24,6 +24,10 @@
 #include <xcb/randr.h>
 #include <boost/throw_exception.hpp>
 
+#include <stdexcept>
+#include <memory>
+#include <string>
+
 namespace mg=mir::graphics;
 namespace mx = mir::X;
 
@@ -103,24 +107,15 @@ public:
         return atom;
     }
 
-    auto get_output_refresh_rate() const -> uint16_t override
+    auto get_output_refresh_rate() const -> double override
     {
         // I'm assuming we handle xcb errors somewhere with events.
         auto ver_cookie = xcb_randr_query_version_unchecked(conn, 1, 2);
         xcb_randr_query_version_reply(conn, ver_cookie,nullptr);
         auto screen_cookie = xcb_randr_get_screen_info_unchecked(conn,screen_->root);
         auto screen_reply = xcb_randr_get_screen_info_reply(conn, screen_cookie, nullptr);
-        #ifdef __cpp_lib_format
-        auto log_msg = std::format("Detected {}Hz host output refresh rate.",screen_reply->rate);
-        #else
-        auto log_msg = [&]()
-        {
-          std::ostringstream x;
-          x << "Detected " << screen_reply->rate << "Hz host output refresh rate.";
-          return x.str();
-        }();
-        #endif
-        mir::logging::log(mir::logging::Severity::debug, log_msg, "mir:x11");
+        auto refresh_rate = static_cast<double>(screen_reply->rate);
+        mir::log_debug("Detected %.2fHz host output refresh rate.", refresh_rate);
         return screen_reply->rate;
     }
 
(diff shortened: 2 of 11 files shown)


pillowtrucker:
The new build errors are actually interesting - it seems that many targets aren't capable of building c++20, let alone c++23

pillowtrucker:
xcb would not, in fact, be the most arcane part of this PR.

AlanGriffiths:
> it seems that many targets aren't capable of building c++20, let alone c++23, which I assumed is the standard we should aim to conform to, based on the cmake files..
> Should I just replace std::format with something more verbose ?

The "oldest" target we support is the one corresponding to the base in our various snaps (which is currently `core22` - corresponding to 22.04LTS). The compiler support for c++23 is mostly there but, as you encountered, the library support is lacking in places.

We aim to transition to `core24`/24.04LTS once these are released and supported.

AlanGriffiths:
> * If the randr extension version of the host X11 is <1.2, this will probably either not work or return nonsense. OTOH, if someone is doing development or testing of Mir-on-X on an Xorg that doesn't support 1.2+, they're probably doing it deliberately.

We don't make any effort to support development on anything earlier than our current snap base (currently `core22`) - which is Ubuntu 22.04LTS. We do however, expect snaps (such as Ubuntu Frame) to run on supported versions of Ubuntu (so, that currently includes 14.04LTS which is still on ESM).

But even 14.04 (trusty) should be fine:

```
$ rmadison libxcb-randr0-dev
 libxcb-randr0-dev | 1.10-2ubuntu1      | trusty         | amd64, arm64, armhf, i386, powerpc, ppc64el
 libxcb-randr0-dev | 1.11.1-1ubuntu1    | xenial         | amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 libxcb-randr0-dev | 1.13-1             | bionic         | amd64, arm64, armhf, i386, ppc64el, s390x
 libxcb-randr0-dev | 1.13-2~ubuntu18.04 | bionic-updates | amd64, arm64, armhf, i386, ppc64el, s390x
 libxcb-randr0-dev | 1.14-2             | focal          | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
 libxcb-randr0-dev | 1.14-3ubuntu3      | jammy          | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
 libxcb-randr0-dev | 1.15-1             | lunar          | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
 libxcb-randr0-dev | 1.15-1             | mantic         | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
 libxcb-randr0-dev | 1.15-1             | noble          | amd64, arm64, armhf, i386, ppc64el, riscv64, s390x
```

debian/control
```diff
@@ -45,6 +45,8 @@ Build-Depends: cmake,
                python3-venv:native,
                libxcb-composite0-dev,
                libxcb-xfixes0-dev,
+# iirc debian/changelog also needs to be updated for the package to get rebuilt/upgraded with xcb-randr support
```
AlanGriffiths, line 48: There's no need for this. And the comment would date rapidly

src/platforms/x11/x11_resources.h
```diff
@@ -20,12 +20,19 @@
...
 #include <optional>
 #include <unordered_map>
 #include <functional>
 #include <mutex>
-
+#if __has_include(<format>)
+#include <format>
+#else
+#include <sstream>
+#endif
```
AlanGriffiths, line 35: I don't think any of these changes are needed in _this header_. Those additional headers needed by the implementation should be added to the implementation file.

src/platforms/x11/x11_resources.cpp
```diff
@@ -101,6 +103,27 @@ class BasicXCBConnection : public mx::XCBConnection
...
+        auto log_msg = std::format("Detected {}Hz host output refresh rate.",screen_reply->rate);
+        #else
+        auto log_msg = [&]()
+        {
+          std::ostringstream x;
+          x << "Detected " << screen_reply->rate << "Hz host output refresh rate.";
+          return x.str();
+        }();
+        #endif
+        mir::logging::log(mir::logging::Severity::debug, log_msg, "mir:x11");
```
AlanGriffiths, line 123:
```suggestion
        mir::log_debug("%s", log_msg.c_str());
```

AlanGriffiths (changes requested):
Looks mostly good with a few tweaks.

The snaps action failing should be addressed by #3137, so I recommend rebasing on main once that lands.

It does seem inelegant to use two sources (`if __has_include(<format>)` and `#ifdef __cpp_lib_format`) for the one truth, but I' wouldn't block on it.

src/platforms/x11/x11_resources.cpp
pillowtrucker: The existence of log_debug makes the whole thing  much much simpler.

pillowtrucker:
My apologies. The `#ifdef` inconsistency and header misplacement was purely due to blindly following clangd's preferences - it tries to be clever and take the tested expression into consideration while syntax-checking, but doesn't really store enough context to keep it consistent. 
The existence of the mir loggers with the standard format specifier significantly simplifies the logging code, so I've cleaned all of that up and rebased onto main.
It looks like some of the CI jobs are a little bit confused now, though..

src/platforms/x11/x11_resources.cpp
```diff
@@ -14,14 +14,20 @@
...
 
 #include "x11_resources.h"
 
 #include <X11/Xlib-xcb.h>
+#include <xcb/randr.h>
 #include <boost/throw_exception.hpp>
 
+#include <stdexcept>
+#include <memory>
+#include <string>
```
AlanGriffiths, line 29: AFAICS you've not added a requirement for any of these?

src/platforms/x11/x11_resources.cpp
```diff
@@ -14,14 +14,20 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+
```
AlanGriffiths, line 17: Spurious new whitespace

AlanGriffiths (changes requested):
Much cleaner, thanks. A couple of nits

pillowtrucker:
All fixed. I think the CI failures are still transient and unrelated to the code. (hopefully...)

AlanGriffiths (approved):

PR #3146, opened 2023-11-22
Bump platformgraphics ABI

AlanGriffiths:
We should bump before release: I'm sure there have been changes. ;^)


RAOF (approved):
Yes, that is correct :)

RAOF:
And, particularly, we've already bumped `libmirplatform` ABI.

PR #3147, opened 2023-11-22
Wlr foreign toplevel manager will return the desktop id as the app_id if it can

mattkae:
## What's new?
- I am proposing a new protocol called to `mir-foreign-toplevel-management-unstable-v1.xml` that will provide everything that the WLR version provides in addition to a mapping from the surface to a `.desktop` file
- The solution is convoluted to say the least
- **Why?**: some shell elements (specifically launchers) need to know which app is open so that they can make an informed decision about what to do when a desktop file is interacted with (e.g. clicking the icon of an already-opened app in the launcher should focus it instead of opening a new instance).

This is super convoluted, as there is no standardized way of doing this AFAIK. There is `WM_CLASS`, but that is hardly used and maybe deprecated. 

## Questions for the reviewer
- :heavy_check_mark:  Should this be a new protocol, e.g. `mir-foreign-toplevel-management-unstable-v1.xml`?
- :heavy_check_mark: How often do we refresh the app cache? Should that even be cached?

### DIFF
--- a/src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
+++ b/src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
@@ -24,13 +24,19 @@
 #include "mir/scene/null_observer.h"
 #include "mir/scene/null_surface_observer.h"
 #include "mir/scene/surface.h"
+#include "mir/scene/session.h"
 #include "mir/log.h"
 #include "mir/executor.h"
+#include "mir/main_loop.h"
 
 #include <algorithm>
 #include <mutex>
 #include <map>
 #include <boost/throw_exception.hpp>
+#include <fstream>
+#include <gio/gdesktopappinfo.h>
+#include <filesystem>
+#include <sys/inotify.h>
 
 namespace mf = mir::frontend;
 namespace ms = mir::scene;
@@ -46,6 +52,28 @@ class ForeignSurfaceObserver;
 class ForeignToplevelManagerV1;
 class ForeignToplevelHandleV1;
 
+class DesktopFileManager
+{
+public:
+    DesktopFileManager(std::shared_ptr<MainLoop> const& main_loop);
+    ~DesktopFileManager();
+
+    void refresh_app_cache();
+    std::string resolve_app_id(const scene::Surface*);
+private:
+    std::shared_ptr<MainLoop> const main_loop;
+    mir::Fd inotify_fd;
+    std::vector<int> config_path_wd_list;
+    GAppInfo* resolve_from_wayland_app_id(std::string& app_id);
+    GAppInfo* lookup_basename(std::string& name);
+    GAppInfo* resolve_if_snap(int pid);
+    GAppInfo* resolve_if_flatpack(int pid);
+    GAppInfo* resolve_if_executable_matches(int pid);
+    std::map<std::string, GAppInfo*> id_to_app;
+    std::map<std::string, std::string> wm_class_to_app_info_id;
+    GList* app_infos = NULL;
+};
+
 /// Informs a client about toplevels from itself and other clients
 /// The Wayland objects it creates for each toplevel can be used to aquire information and control that toplevel
 /// Useful for task bars and app switchers
@@ -57,11 +85,13 @@ public:
         wl_display* display,
         std::shared_ptr<shell::Shell> const& shell,
         std::shared_ptr<Executor> const& wayland_executor,
-        std::shared_ptr<SurfaceStack> const& surface_stack);
+        std::shared_ptr<SurfaceStack> const& surface_stack,
+        std::shared_ptr<MainLoop> const& main_loop);
 
     std::shared_ptr<shell::Shell> const shell;
     std::shared_ptr<Executor> const wayland_executor;
     std::shared_ptr<SurfaceStack> const surface_stack;
+    std::shared_ptr<DesktopFileManager> const desktop_file_manager;
 
 private:
     void bind(wl_resource* new_resource) override;
@@ -136,6 +166,7 @@ public:
     ~ForeignToplevelManagerV1();
 
     std::shared_ptr<shell::Shell> const shell;
+    std::shared_ptr<DesktopFileManager> const desktop_file_manager;
 
 private:
     /// Wayland requests
@@ -188,10 +219,11 @@ auto mf::create_foreign_toplevel_manager_v1(
     wl_display* display,
     std::shared_ptr<shell::Shell> const& shell,
     std::shared_ptr<Executor> const& wayland_executor,
-    std::shared_ptr<SurfaceStack> const& surface_stack)
+    std::shared_ptr<SurfaceStack> const& surface_stack,
+    std::shared_ptr<MainLoop> const& main_loop)
 -> std::shared_ptr<mw::ForeignToplevelManagerV1::Global>
 {
-    return std::make_shared<ForeignToplevelManagerV1Global>(display, shell, wayland_executor, surface_stack);
+    return std::make_shared<ForeignToplevelManagerV1Global>(display, shell, wayland_executor, surface_stack, main_loop);
 }
 
 // ForeignToplevelManagerV1Global
@@ -200,11 +232,13 @@ mf::ForeignToplevelManagerV1Global::ForeignToplevelManagerV1Global(
     wl_display* display,
     std::shared_ptr<shell::Shell> const& shell,
     std::shared_ptr<Executor> const& wayland_executor,
-    std::shared_ptr<SurfaceStack> const& surface_stack)
+    std::shared_ptr<SurfaceStack> const& surface_stack,
+    std::shared_ptr<MainLoop> const& main_loop)
     : Global{display, Version<2>()},
       shell{shell},
       wayland_executor{wayland_executor},
-      surface_stack{surface_stack}
+      surface_stack{surface_stack},
+      desktop_file_manager{std::make_shared<DesktopFileManager>(main_loop)}
 {
 }
 
@@ -369,7 +403,7 @@ void mf::ForeignSurfaceObserver::create_or_close_toplevel_handle_as_needed(std::
             handle = std::make_shared<mw::Weak<ForeignToplevelHandleV1>>();
 
             std::string name = surface->name();
-            std::string app_id = surface->application_id();
+            std::string app_id = manager.value().desktop_file_manager->resolve_app_id(surface.get());
             auto const focused = surface->focus_state();
             auto const state = surface->state_tracker();
 
@@ -383,8 +417,7 @@ void mf::ForeignSurfaceObserver::create_or_close_toplevel_handle_as_needed(std::
 
             if (!name.empty())
                 handle->value().send_title_event(name);
-            if (!app_id.empty())
-                handle->value().send_app_id_event(app_id);
+            handle->value().send_app_id_event(app_id);
             handle->value().send_state(focused, state);
             handle->value().send_done_event();
         }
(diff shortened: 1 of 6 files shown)


RAOF:
> * How often do we refresh the app cache? Should that even be cached?

We put an inotify watch on the directories, and update when there's changes. :wink:

wayland-protocols/wlr-foreign-toplevel-management-unstable-v1.xml
```diff
@@ -255,5 +255,25 @@
...
+        This event is emitted whenever the parent of the toplevel changes.
+
+        No event is emitted when the parent handle is destroyed by the client.
+      </description>
+      <arg name="parent" type="object" interface="zwlr_foreign_toplevel_handle_v1" allow-null="true"/>
+    </event>
+
+    <!-- Version 4 additions -->
+
+    <event name="desktop_app_id">
```
RAOF, line 272: This is kinda the same as the existing `app_id` event, right? The difference being that this will correspond to a *usable* `app_id` more often, because we do some matching in Mir?

Do you know how often applications set a bogus `app_id` on their `xdg_toplevel`s? That is: would it be sensible to pass on the `app_id` of applications that set it, and synthesise an `app_id` for the others, and then use the existing `app_id` event here?

mattkae: They seem to set the `app_id` pretty regularly and when they set it, it's _pretty_ hard to get back to the `.desktop` file from there. For example, opening up `gnome-terminal` reports as `gnome-terminal-server` which in _theory_  I can arrive at `org.gnome.Terminal.desktop` for... but it would be weird given the contents of `org.gnome.Terminal.desktop`

mattkae:
@RAOF Do you think I should make this a `mir`-prefaced protocol too?

wayland-protocols/wlr-foreign-toplevel-management-unstable-v1.xml
mattkae: And the catch with all of this is that the client would never be able to do the complicated work to resolve that `app_id` into anything usable anyway (especially not without access to a PID for snaps, flatpaks, etc.)

AlanGriffiths:
>     * Should this be a new protocol, e.g. `mir-foreign-toplevel-management-unstable-v1.xml`?

Yes. Introducing a spurious new version of a protocol owned by someone else will definitely lead to confusion

mattkae:
@AlanGriffiths I have made it to a new protocol. I am not sharing code between the two protocols although they are very similar, as I think I'd have to get pretty spooky with the generics for it to work (each protocol deals with its own set of classes). If it is okay, I'll keep it copy-and-pasted for clarity's sake

RAOF:
Do we actually need a new protocol, here? (serious question)

The existing protocol includes the `app_id` event, and the `app_id` is *meant* to be the XDG desktop file ID. Would things break is Mir sent the resolved ID as `app_id` instead of what the client sent in the `xdg_toplevel` request?

AlanGriffiths:
> Do we actually need a new protocol, here? (serious question)
> 
> The existing protocol includes the `app_id` event, and the `app_id` is _meant_ to be the XDG desktop file ID. Would things break if Mir sent the resolved ID as `app_id` instead of what the client sent in the `xdg_toplevel` request?

There seem to be a limited number of cases where the `app_id ` does follow the best practice suggestion in `xdg_toplevel.set_app_id` and not match the basename of the application's .desktop file. The following are all fine:

```
WAYLAND_DEBUG=1 xfce4-terminal 2>&1 | grep set_app_id
[ 395583.125]  -> xdg_toplevel@24.set_app_id("xfce4-terminal")
$ find /usr/share/applications -name xfce4-terminal.desktop
/usr/share/applications/xfce4-terminal.desktop

$ WAYLAND_DEBUG=1 gnome-chess 2>&1 | grep set_app_id
[  29910.863]  -> xdg_toplevel@27.set_app_id("org.gnome.Chess")
$ find /usr/share/applications -name org.gnome.Chess.desktop
/usr/share/applications/org.gnome.Chess.desktop

$ WAYLAND_DEBUG=1 kgx 2>&1 | grep set_app_id
[ 744864.441]  -> xdg_toplevel@25.set_app_id("org.gnome.Console")
$ find /usr/share/applications/ -name org.gnome.Console.desktop
/usr/share/applications/org.gnome.Console.desktop
```

As mentioned above, `gnome-terminal` is weird (and not just in this way). How common are problems really?

RAOF:
> As mentioned above, `gnome-terminal` is weird (and not just in this way). How common are problems really?

Well, no XWayland application will call `set_app_id`, so we need something for that.

I think I'm firming behind the opinion that this does *not* require extra protocol, and we should just back the existing `app_id` event on the new desktop file lookup code.

mattkae:
@AlanGriffiths and @RAOF : In _Mir_, we're going to resolve the `app_id` to a desktop file ID if it is empty. In this case, the `app_id` event in the protocol will return either of two things: 1) the plain app_id, or 2) the desktop file ID. To prevent the user from having to resolve that `app_id` to a desktop file themselves on their end, I recommend that we _always_ return the desktop file ID, for consistency.

Would this be "breaking" the protocol? Would it be bad if this field behaved slightly differently for Mir?

Saviq:
Poking CI.

Saviq:
Something here causes errors starting apps under Mir:

```
< -warning- > frontend:Wayland: Exception processing Surface::commit() request: Dynamic exception type: std::logic_error
std::exception::what: basic_string::_M_construct null not valid                           
error in client communication (pid 11036)
Error 71 (Błąd protokołu) dispatching to Wayland display.
gnome-terminal-server.service: Main process exited, code=exited, status=1/FAILURE
gnome-terminal-server.service: Failed with result 'exit-code'.
# Error: Message recipient disconnected from message bus without replying        
```

mattkae:
> Something here causes errors starting apps under Mir:
> 
> ```
> < -warning- > frontend:Wayland: Exception processing Surface::commit() request: Dynamic exception type: std::logic_error
> std::exception::what: basic_string::_M_construct null not valid                           
> error in client communication (pid 11036)
> Error 71 (Błąd protokołu) dispatching to Wayland display.
> gnome-terminal-server.service: Main process exited, code=exited, status=1/FAILURE
> gnome-terminal-server.service: Failed with result 'exit-code'.
> # Error: Message recipient disconnected from message bus without replying        
> ```

This wasn't happening for me. Is there something different about your setup? I did:
```
sudo snap install miriway --classic --channel=latest/edge/mir-pr3147
```

Then I ran `gnome-terminal` with no issue

Saviq:
> fix: xwayland does not provide an executable

Cool. It does help. But this is a lot of untested logic. I'd like to see some unit tests here.

mattkae:
@RAOF Turns out: we do resolve X apps properly: https://github.com/MirServer/mir/blob/main/src/server/frontend_xwayland/xwayland_surface.cpp#L311

Conan-Kudo:
> I am proposing a new protocol called to mir-foreign-toplevel-management-unstable-v1.xml that will provide everything that the WLR version provides in addition to a mapping from the surface to a .desktop file

Have you considered proposing an `ext-` protocol that both wlr and mir could use? Then also kwin could consider implementing it too.

mattkae:
> Have you considered proposing an `ext-` protocol that both wlr and mir could use? Then also kwin could consider implementing it too.

I considered a protocol extension at first, but it ended up being the case that we only wanted to modify the `app_id` event to be more informative for the end-user, so we decided against that.

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -458,26 +489,148 @@ void mf::ForeignSurfaceObserver::renamed(ms::Surface const*, std::string const&
...
 {
     std::lock_guard lock{mutex};
 
     std::string id = application_id;
-    with_toplevel_handle(lock, [id](ForeignToplevelHandleV1& handle)
+    with_toplevel_handle(lock, [&](ForeignToplevelHandleV1& handle)
         {
-            handle.send_app_id_event(id);
+            auto app_id = manager.value().desktop_file_manager->resolve_app_id(surface);
+            handle.send_app_id_event(app_id);
```
RAOF, line 501: Likewise, need to handle `app_id == nullptr`.

tests/unit-tests/frontend_wayland/test_desktop_file_manager.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+TEST_F(DesktopFileManager, can_find_app_when_app_id_matches_with_gnome_prefix)
+{
+    const std::string GNOME_PREFIX = "gnome-";
+    const std::string gnome_prefixed_desktop_file_app_id = GNOME_PREFIX + DESKTOP_FILE_APP_ID;
+    auto new_file = std::make_shared<mf::DesktopFile>(gnome_prefixed_desktop_file_app_id.c_str(), nullptr, nullptr);
+    cache->files.push_back(new_file);
+    auto app_id = file_manager->resolve_app_id(&surface);
+    EXPECT_THAT(app_id, Eq(new_file->id));
+}
```
RAOF, line 149: Isn't this duplicated by the parametrised test below? Do I understand that correctly?

tests/unit-tests/frontend_wayland/test_desktop_file_manager.cpp
```diff
@@ -0,0 +1,282 @@
...
+    EXPECT_THAT(desktop_file, Eq(new_file));
+}
+
+TEST_F(DesktopFileManager, can_find_desktop_file_if_exec_name_matches_despite_arguments_given)
+{
+    auto new_file = std::make_shared<mf::DesktopFile>(nullptr, nullptr, "gnome-calendar");
+    cache->files.push_back(new_file);
+    auto desktop_file = file_manager->parse_exec_call_to_desktop_file("gnome-calendar --arg1 --arg2");
+    EXPECT_THAT(desktop_file, Eq(new_file));
+}
```
RAOF, line 212: Pretty sure this is unnecessary, because your input data doesn't look like "gnome-calendar --arg1 --arg2".

RAOF (commented):
Getting closer! Sorry I didn't pick up on these things the first time around.

It might be nice to have tests of `GDesktopFileCache`? Since it respects `XDG_DATA_DIRS` you could set that to test directories and copy in .desktop files, delete them from there, and so on. That'd make it easier to check you handle weird .desktop files (as you could just write some :wink: )

Not required now, but it looks like this'll start causing Mir to do a fair chunk of work whenever apps change their `app_id`s; it might be worth following GNOME Shell and caching the *results* of the `app_id` lookup, too. Again, not required now, but we should track that follow-up.

RAOF (commented):
GDesktopFileCache tests look good! Sorry about the other architectural change suggestion :grimacing:

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -458,26 +471,167 @@ void mf::ForeignSurfaceObserver::renamed(ms::Surface const*, std::string const&
...
+            // If the manager is not available for whatever reason, we should still try to
+            // send the application_id specified by the app itself.
+            if (!manager)
+            {
+                handle.send_app_id_event(surface->application_id());
+                handle.send_done_event();
+                return;
+            }
+
+            auto app_id = manager.value().desktop_file_manager->resolve_app_id(surface);
```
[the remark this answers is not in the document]
mattkae: Yup agreed! So more of a singleton instance. On it :vulcan_salute:

src/server/frontend_wayland/foreign_toplevel_manager_v1.h
```diff
@@ -36,8 +39,30 @@ auto create_foreign_toplevel_manager_v1(
...
+public:
+    GDesktopFileCache(std::shared_ptr<MainLoop> const& main_loop);
+    ~GDesktopFileCache() override = default;
+    std::shared_ptr<DesktopFile> lookup_by_app_id(std::string const&) const override;
+    std::shared_ptr<DesktopFile> lookup_by_wm_class(std::string const&) const override;
+    std::vector<std::shared_ptr<DesktopFile>> const& get_desktop_files() const override;
+    void refresh_app_cache();
+
+private:
+    mutable std::recursive_mutex update_mutex;
```
RAOF, line 57: Hm. `recursive_mutex` is somewhat of a code smell. It's also not clear (from here) what it's protecting.

You could, perhaps, use `mir::Synchronised` (or two different `Synchronised`s; uncontended lock acqisition is *super* cheap) here, or at least move the mutex down to the fields it's protecting.

RAOF (commented):
Hm. I must have been unclear in my last review, sorry; this doesn't address it?

Or, rather, it addresses half of it (not lots of `desktop_file_manager`s) but not the trigger for that the suggestion - that the toplevel observer should share ownership of them, rather than requiring that `foreign_manager` stick around.

I'll ping you in my morning to see if we can't get this landed!

mattkae:
> Or, rather, it addresses half of it (not lots of `desktop_file_manager`s) but not the trigger for that the suggestion - that the toplevel observer should share ownership of them, rather than requiring that `foreign_manager` stick around.

Oh I understand you more clearly now. We can make the observer share ownership instead of guessing if our parent has it. That makes sense to me too.

src/server/frontend_wayland/foreign_toplevel_manager_v1.h
mattkae: It's protecting access to the cached files, maps, etc. Basically, the update of those variable can happen in another thread (when we are notified that the files changed), so I'm trying to avoid the situation where another thread queries for the filed and receives an half-built list.

mattkae: And recursive_mutex was used for bad reasons. We can chat about this in your morning ;)

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -458,19 +482,156 @@ void mf::ForeignSurfaceObserver::renamed(ms::Surface const*, std::string const&
...
+    std::unique_ptr<GList, decltype(free_app_infos)> app_infos(g_app_info_get_all(), free_app_infos);
+    std::map<std::string, std::shared_ptr<DesktopFile>> new_id_to_app;
+    std::map<std::string, std::shared_ptr<DesktopFile>> new_wm_class_to_app_info;
+    std::map<std::string, std::shared_ptr<DesktopFile>> new_exec_to_app;
+
+    GList* info;
+    for (info = app_infos.get(); info != NULL; info = info->next)
+    {
+        GAppInfo *app_info = static_cast<GAppInfo*>(info->data);
+        if (!g_app_info_should_show(app_info))
```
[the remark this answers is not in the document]
RAOF: And the answer is: no, because there are cases where both a `NoDisplay` and should-display-in .desktop file will match, and we want to ensure that the normal desktop file wins.

cf: https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2721

RAOF (approved):
LGTM now.

RAOF:
While there is some fiddling that could be done to improve detection, there's *always* going to be fiddling to be done to improve detection, so I think this is sufficiently useful to merge.

RAOF:
Of course, the segfault in the test on Fedora 38 needs fixing.

Issue #3149, opened 2023-11-24
[VT console provider] Input device not recognised after disconnect/reconnect cycle


RAOF:
When using `--console=vt`, my USB keyboard works both if it is plugged in before Mir is started or if it is plugged in after Mir has started.

However, when it is unplugged, Mir (correctly) notices it has disappeared, but never notices when it is plugged *back* in.

This is specific to the `LinuxVirtualTerminal` console provider; when using logind I can plug and unplug the device multiple times and it works as expected.

RAOF:
Aha! This is because in `evdev/platform.cpp` we rely on the `ConsoleProvider` to call the `removed()` method on the `InputDeviceObserver` when a device is removed, but only the `LogindConsoleProvider` actually does this.

For `LinuxVirtualTerminal`, failure to call `removed()` means that on hot-unplug the input device is removed, but the corresponding `mir::Device` is *not* removed from `device_watchers`, and so when plugged back in our udev ADDED event handler finds it in the `device_watchers` array, assumes we're already using it, and bails.

AlanGriffiths:
Thanks for tracking this down. I can now reproduce

PR #3150, opened 2023-11-24
platforms/evdev: Clean up mir::Device handle on device remove

RAOF:
Not all `ConsoleProvider`s will send a `removed` event on the `DeviceObserver` (in fact, only logind will), so we can't rely on that to clean up the handle in `device_watchers`.

Furthermore, libinput itself will notice a device going away (by monitoring the fd) and generate `LIBINPUT_EVENT_DEVICE_REMOVED`, which we might process *before* the udev `REMOVED` event, and in *that* case, `device_removed` will have already deleted the `LibInputDevice` and so the udev event handler will *also* not clean up `device_watchers`.

Resolve this by cleaning up `device_watchers` *only* in `device_removed`, and rely on the `DeviceObserver::removed` codepath triggering a libinput removal event.

Fixes: #3149

### DIFF
--- a/src/platforms/evdev/platform.cpp
+++ b/src/platforms/evdev/platform.cpp
@@ -212,9 +212,7 @@ public:
         action_queue->enqueue(
             [
                 &devices = devices,
-                syspath = syspath,
-                &device_watchers = device_watchers,
-                devnum = devnum
+                syspath = syspath
             ]()
             {
                 for (auto const& input_device : devices)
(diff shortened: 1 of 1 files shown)


RAOF:
Please note that *this* code has not been fully tested; a version of this with more debugging successfully resolves the plug/unplug/replug scenario, though.

To test this you need to run Mir at a console and use `--console-provider=vt`; the (default) logind provider does not hit this bug.

AlanGriffiths (approved):
Confirmed this fixes the linked bug, and it looks like a sane application of "single source of truth".

Am going to test further before merging (and cherry-picking into 2.16)

PR #3153, opened 2023-11-27
Make server-side decorations scale aware

### LINKED ISSUE #1369: Server-side decorations not scale aware
They are low resolution when on a scaled output. They should know when they are on a scaled output and render a buffer at the appropriate scale

hbatagelo:
Fixes #1369.

The decoration builder now selects the maximum scale factor among all display outputs, passing it to `msd::BasicDecoration` and `msd::Renderer` to create buffers with a resolution that ensures sharp-looking decorations on all outputs.

### DIFF
--- a/src/server/shell/decoration/basic_decoration.cpp
+++ b/src/server/shell/decoration/basic_decoration.cpp
@@ -266,15 +266,15 @@ void msd::BasicDecoration::set_cursor(std::string const& cursor_image_name)
     shell->modify_surface(session, decoration_surface, spec);
 }
 
-void msd::BasicDecoration::set_scale(float scale)
+void msd::BasicDecoration::set_scale(float new_scale)
 {
-    scale_ = scale;
+    scale = new_scale;
     window_state_updated();
 }
 
 auto msd::BasicDecoration::new_window_state() const -> std::unique_ptr<WindowState>
 {
-    return std::make_unique<WindowState>(static_geometry, window_surface, scale_);
+    return std::make_unique<WindowState>(static_geometry, window_surface, scale);
 }
 
 auto msd::BasicDecoration::create_surface() const -> std::shared_ptr<scene::Surface>
--- a/src/server/shell/decoration/basic_decoration.cpp
+++ b/src/server/shell/decoration/basic_decoration.cpp
@@ -158,8 +158,7 @@ msd::BasicDecoration::BasicDecoration(
     std::shared_ptr<mg::GraphicBufferAllocator> const& buffer_allocator,
     std::shared_ptr<Executor> const& executor,
     std::shared_ptr<input::CursorImages> const& cursor_images,
-    std::shared_ptr<ms::Surface> const& window_surface,
-    float scale)
+    std::shared_ptr<ms::Surface> const& window_surface)
     : threadsafe_self{std::make_shared<ThreadsafeAccess<BasicDecoration>>(executor)},
       static_geometry{std::make_shared<StaticGeometry>(default_geometry)},
       shell{shell},
@@ -167,7 +166,7 @@ msd::BasicDecoration::BasicDecoration(
       cursor_images{cursor_images},
       session{window_surface->session().lock()},
       buffer_streams{std::make_unique<BufferStreams>(session)},
-      renderer{std::make_unique<Renderer>(buffer_allocator, static_geometry, scale)},
+      renderer{std::make_unique<Renderer>(buffer_allocator, static_geometry)},
       window_surface{window_surface},
       decoration_surface{create_surface()},
       window_state{new_window_state()},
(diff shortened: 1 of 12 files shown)


AlanGriffiths (changes requested):
This works and looks fine for the case when outputs remain as configured on startup. But the scale can be reconfigured dynamically (e.g. by updating a "static" `.display` configuration).

The `msd::BasicManager` (or setup) likely needs to register a `DisplayConfigurationObserver` to track scale changes.

hbatagelo:
`msd::BasicManager` now monitors display changes, updating existing decorations when the scale changes by invoking `Decoration::set_scale` for each decoration.

While this solution seems to work well, it differs from my initial attempt. Initially, I tried rebuilding all decorations when the scale changed, but the behavior was not as I expected. When a window surface is undecorated, the new decoration appears detached from the window, as if the decoration surface is no longer a child of the window surface. I'd like to continue investigating this behavior in order to understand what is causing it. In summary, the current implementation is functional, but if I can fix this new issue I can propose an alternative approach that avoids requiring classes derived from `Decoration` to implement a `set_scale` function.

src/server/shell/decoration/basic_decoration.cpp
```diff
@@ -266,9 +266,15 @@ void msd::BasicDecoration::set_cursor(std::string const& cursor_image_name)
     shell->modify_surface(session, decoration_surface, spec);
 }
 
+void msd::BasicDecoration::set_scale(float scale)
+{
+    scale_ = scale;
+    window_state_updated();
+}
```
AlanGriffiths, line 273: Is access to `scale_` threadsafe? AFAICS it isn't atomic, guarded by a mutex, or guaranteed to be accessed on a single thread

src/server/shell/decoration/basic_manager.cpp
```diff
@@ -85,3 +139,30 @@ void msd::BasicManager::undecorate_all()
...
+        {
+            if (!output.used || !output.connected) return;
+            if (!output.valid() || (output.current_mode_index >= output.modes.size())) return;
+
+            max_output_scale = std::max(max_output_scale, output.scale);
+        });
+
+    if (max_output_scale == scale) return;
+
+    scale = max_output_scale;
```
AlanGriffiths, line 158: `scale` is being read and mutated without locking the mutex

src/server/shell/decoration/basic_manager.cpp
```diff
@@ -85,3 +139,30 @@ void msd::BasicManager::undecorate_all()
...
+
+    std::lock_guard lock{mutex};
+    for (auto& it : decorations)
+    {
+        if (it.second)
```
AlanGriffiths, line 163: I don't think this condition can fail: an empty decoration should not be possible

src/server/shell/decoration/basic_manager.h
```diff
@@ -18,34 +18,56 @@
...
+{
+public:
+    virtual void advise_output_update(mir::graphics::DisplayConfiguration const& config) = 0;
+
+protected:
+    OutputListener() = default;
+    virtual ~OutputListener() = default;
+    OutputListener(OutputListener const&) = delete;
+    OutputListener operator=(OutputListener const&) = delete;
+};
```
AlanGriffiths, line 56: I see no reason to make `OutputListener` part of the `basic_manager.h` interface. Or even why we need that interface.

`OutputListener` is only used by `msd::DisplayConfigurationListener` which is implemented entirely in `basic_manager.cpp` and that just needs something to call (e.g. `std::function<void(float new_scale)>`)

AlanGriffiths (changes requested):
> I tried rebuilding all decorations when the scale changed, but the behavior was not as I expected

That sounds like a wrong approach: rescaling ought to be less work than tearing down and rebuilding

hbatagelo:
- `msd::DisplayConfigurationListener` now inherits from `mg::NullDisplayConfigurationObserver`.
- Usage of `BasicManager::scale` is synchronized under a locked mutex.
- `BasicDecoration::set_scale` is also called under the manager's locked mutex. Since the decorations are uniquely owned by the manager, this should be sufficient to guarantee thread safety.
- Unnecessary `nullptr` checks have been removed. Additionally, test cases for the decoration manager fixture have been updated to ensure the mocked decoration builder returns a valid pointer (the reason the checks were added originally).
- `OutputListener` has been eliminated.
- For consistency, `BasicDecoration::scale_` has been renamed to `BasicDecoration::scale`.

I've also changed the initialization of `decoration_builder` in the constructor of `msd::BasicManager`. The rvalue reference parameter is now moved instead of copied to the member variable.

AlanGriffiths:
@hbatagelo sorry for the delay reviewing - I ran into a problem (#3165) while testing

hbatagelo:
No problem. Feel free to postpone the review until #3165 is resolved.

AlanGriffiths (approved):
Thanks! Looks good now

Saviq (approved):
Nice and crisp!

obraz

PR #3155, opened 2023-11-30
tests: updated the desktop file manager so that we can more easily write tests for it

mattkae:
## What's new?
- Fixed a bug where `gnome-disks` wasn't showing up (we needed a `NoDisplay` check)
- Wrote tests for everything except fileystem-specific stuff

### DIFF
--- a/src/server/frontend_wayland/desktop_file_manager.cpp
+++ b/src/server/frontend_wayland/desktop_file_manager.cpp
@@ -25,9 +25,9 @@
 namespace mf = mir::frontend;
 
 mf::DesktopFile::DesktopFile(const char* id, const char* wm_class, const char* exec)
-    : id{id},
-      wm_class{wm_class},
-      exec{exec}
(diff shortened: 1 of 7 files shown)


src/server/frontend_wayland/desktop_file_manager.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // We are reading the security profile here, which comes to us in the form:
+    //      snap.name-space.binary-name (current).
+    std::string attr_file = "/proc/" + std::to_string(pid) + "/attr/current";
+    if (!std::filesystem::exists(attr_file))
+        return nullptr;
+
+    std::string contents;
+    std::getline(std::ifstream(attr_file), contents, '\0');
+
+    const char* SNAP_SECURITY_LABEL_PREFIX = "snap.";
```
AlanGriffiths, line 127:
```suggestion
    char const* const snap_security_label_prefix = "snap.";
```

PR #3156, opened 2023-12-01
Fix dragging and resizing satellite windows

AlanGriffiths:
I'm not sure why we were checking that refocusing on the window selected for the gesture activated it. It seems both pointless and wrong. So dropped that code.

There are not a lot of apps to illustrate the issues, but gimp (with floating toolboxes) is one: without this it isn't possible to move the toolboxes.


mattkae (approved):
Working for me :) Odd that we were doing that, and I also see no reason to do so

PR #3157, opened 2023-12-04
mir::GLibMainLoop: Fix deprecated `std::result_of_t` usage.

RAOF:
`result_of_t` was deprecated in C++17, and removed in C++20, which we claim to use. Replace it with the equivalent `invoke_result_t` templates.

Fixes the Ubuntu 24.04 build with modern Clang.


Saviq (approved):
ACK

PR #3159, opened 2023-12-05
More X11/Xwayland debug logging

AlanGriffiths:
This is a version of some tracing I found useful fixing some of the CLion menu/gloss/tip weirdness.

I still occasionally see, for example, the right click popup on the git log panel being transparant to mouse input. So having this additional logging could help understand what is happening.


Saviq (approved):
ACK

Issue #3128, opened 2023-11-16
Layout of layer-shell surfaces non-deterministic


Saviq:
Say you're building a shell with a couple components - top panel and a dock on the left. Both share the same layer, both have exclusive zones. Depending on ~~the order in which they start~~luck, they will be laid out differently.

[The protocol](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:set_exclusive_zone) explicitly states that:

> The compositor's use of this information is implementation-dependent

We need a better story, either by defining a "precedence" order of the screen edges, or making it configurable.

Saviq:
Here's a reproducer. Use two waybars with the following configs:

```json
{
    "modules-center": ["clock"],
    "modules-right": ["cpu", "memory"],
    "tray": {
        "spacing": 10
    },
    "clock": {
        "format": "{:%y-%m-%d %H:%M}",
        "tooltip-format": "{:%y-%m-%d | %H:%M}",
        "format-alt": "{:%H:%M}"
    },
    "cpu": {
        "format": "CPU: {usage}%",
        "tooltip": false
    },
    "memory": {
        "format": "Mem: {}%"
    },
}
```
```json
{
    "position": "left",
    "modules-center": ["clock"],
    "modules-right": ["cpu", "memory"],
    "tray": {
        "spacing": 10
    },
    "clock": {
        "format": "{:%y-%m-%d %H:%M}",
        "tooltip-format": "{:%y-%m-%d | %H:%M}",
        "format-alt": "{:%H:%M}"
    },
    "cpu": {
        "format": "CPU: {usage}%",
        "tooltip": false
    },
    "memory": {
        "format": "Mem: {}%"
    },
}
```

Start one, then start/kill the other, you'll see that sometimes the top bar pushes the side bar off, sometimes the other way around:
<img src="https://github.com/MirServer/mir/assets/231077/d1631578-f1c0-444c-b44b-846c6f674294" width="30%" /> <img src="https://github.com/MirServer/mir/assets/231077/022e5ffb-34fb-49f7-8322-d636e51d9f8d" width="30%" />

Issue #3125, opened 2023-11-15
[miral] Applications Zones need a better representation of exclusion zones


AlanGriffiths:
As currently implemented application zones unconditionally respect exclusion zones. However, applications with access to layer-shell can opt out of exclusion zones and should use an unadjusted application zone (e.g a background)

AlanGriffiths:
See also #1992

mattkae:
Maybe related: https://github.com/canonical/mir/issues/3274

mattkae:
Also, I made [this ticket](https://github.com/canonical/wlcs/issues/332) in WLCS to implement a test for `-1` exclusion zones

Issue #1992, opened 2021-04-13
Layer shell: respect -1 exclusive zone


wmww:
If an exclusive zone of -1 is given over the layer shell protocol, it means...

> the surface indicates that it would not like to be moved to accommodate for other surfaces, and the compositor should extend it all the way to the edges it is anchored to.

We do not currently implement this behavior. Our current behavior is inconsistent. After #1991 is merged we will consistently work correctly for exclusive zones >=0.

PR #3160, opened 2023-12-07
(#3128) bugfix: east/west attached surfaces always push north/south attached out of the way

mattkae:
fixes https://github.com/MirServer/mir/issues/3128

## What's new?
- Before we place a surface in its zone, we sort the list such that surfaces which span east-to-west (e.g. top and bottom panels) are at the end and therefore "push" all other surfaces out of their zones.

## Questions
- Should we demand that any other types of surfaces take precedence over any other type?

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -2848,11 +2848,15 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
     {
         Rectangle zone_rect = area->area;
 
-        /// The first pass will modify the application zone as it goes
-        /// The second pass will use the final application zone
+        // The placement algorithm operates on windows spanning the width of the screen first,
+        // followed by any other "attached" window (e.g. a window attached to the side of the screen
+        // vertically), and finally maximized windows. This ensures that windows attached to the
+        // top and bottom of the screen (in the "full_width_attached_windows" list) push windows
+        // attached to the sides of the screen or elsewhere (in the "other_attached_windows") out
+        // of the way. Maximized windows are then guaranteed to be on top, as they appear last.
         std::vector<WindowInfo*> full_width_attached_windows;
         std::vector<WindowInfo*> other_attached_windows;
-        std::vector<WindowInfo*> second_pass;
+        std::vector<WindowInfo*> maximized_windows;
 
         for (auto const& window : area->attached_windows)
         {
@@ -2869,8 +2873,8 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
                         // force everyone else out of the way when they're placed. By "east-to-west", we are
                         // referring to "top" and "bottom" panels, as opposed to "left" and "right" panels.
                         MirPlacementGravity  edges = info.attached_edges();
-                        bool isFullWidth = (edges & mir_placement_gravity_east) && (edges & mir_placement_gravity_west);
-                        if (isFullWidth)
+                        bool is_full_width = (edges & mir_placement_gravity_east) && (edges & mir_placement_gravity_west);
+                        if (is_full_width)
                             full_width_attached_windows.push_back(&info);
                         else
                             other_attached_windows.push_back(&info);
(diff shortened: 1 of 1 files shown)


Saviq:
> * Before we place a surface in its zone, we sort the list such that east/west surfaces are at the end and therefore "push" all other surfaces out of their zones.

I thought we're doing the opposite - e.g. top panel always spanning the screen :shrug:

AlanGriffiths:
>     * Should we demand that any other types of surfaces take precedence over any other type?

See also #1992 & #3125

src/miral/basic_window_manager.cpp
```diff
@@ -2881,6 +2881,23 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
             }
         }
 
+        // We are sorting east/west connected edges to the back of the list so that they
+        // force everyone else out of the way when they're placed.
+        std::sort(first_pass.begin(), first_pass.end(), [](WindowInfo* first, WindowInfo* second) {
```
AlanGriffiths, line 2886: 1. Wouldn't `partition` (or `remove_if`!) be more efficient than `sort`?
2. Maybe we should split `first_pass` into a `horizontal` and `vertical` pass? instead of partitioning?

AlanGriffiths (changes requested):
While we're in this code, could we also deal with [exclusive zone of -1](#1992) - and, maybe, that allows more flexible shell building?

mattkae:
> I thought we're doing the opposite - e.g. top panel always spanning the screen 🤷

Oh I meant surfaces spanning east to west (e.g. top and bottom panels). So same thing

src/miral/basic_window_manager.cpp
mattkae: 2 sounds good!

src/miral/basic_window_manager.cpp
```diff
@@ -2864,7 +2865,15 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
                 case mir_window_state_attached:
                     if (info.exclusive_rect().is_set())
                     {
-                        first_pass.push_back(&info);
+                        // We are sorting east-to-west connected edges to the back of the list so that they
+                        // force everyone else out of the way when they're placed. By "east-to-west", we are
+                        // referring to "top" and "bottom" panels, as opposed to "left" and "right" panels.
+                        MirPlacementGravity  edges = info.attached_edges();
+                        bool isFullWidth = (edges & mir_placement_gravity_east) && (edges & mir_placement_gravity_west);
```
AlanGriffiths, line 2872:
```suggestion
                        bool is_full_width = (edges & mir_placement_gravity_east) && (edges & mir_placement_gravity_west);
```

src/miral/basic_window_manager.cpp
```diff
@@ -2850,7 +2850,8 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
 
         /// The first pass will modify the application zone as it goes
         /// The second pass will use the final application zone
```
AlanGriffiths, line 2852: Comments that disagree with the code are worse than useless

AlanGriffiths (changes requested):
Better, but needs tweaking

src/miral/basic_window_manager.cpp
```diff
@@ -2864,14 +2869,22 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
                 case mir_window_state_attached:
                     if (info.exclusive_rect().is_set())
                     {
-                        first_pass.push_back(&info);
+                        // We are sorting east-to-west connected edges to the back of the list so that they
+                        // force everyone else out of the way when they're placed. By "east-to-west", we are
+                        // referring to "top" and "bottom" panels, as opposed to "left" and "right" panels.
```
AlanGriffiths, line 2874: We have a comment above that agrees with the code, we don't need another here that disagrees

Saviq (approved):
Yups.

PR #3162, opened 2023-12-11
various: Drop nonexistent symbols from symbols.map

RAOF:
`mold` likes to complain about these; newer `lld`s make this a hard error


AlanGriffiths (approved):
+1

PR #3166, opened 2023-12-13
[RFC] mir-shell

AlanGriffiths:
Prototype implementation of mir-shell prototcol extension.

This includes a demo client: `mir_demo_client_mir_shell`.

The client can windows from the "normal", "satellite", "dialog" and "utility" architypes.

* normal: created when running the app, closed by `Ctrl-Q`
* satellite: created by `Ctrl-T` (for "toolbar"), press again to create on another edge, closed by `esc`
* dialog: created by `Ctrl-D`, closed by `esc`
* utility: created by `Ctrl-U`, closed by `esc`

Note that while "freestyle" is mentioned in the protocol, it is underspecified and the implemention is only a placeholder.

### DIFF
--- a/wayland-protocols/mir-shell-unstable-v1.xml
+++ b/wayland-protocols/mir-shell-unstable-v1.xml
@@ -1,0 +1,532 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="mir_shell_unstable_v1">
+  <copyright>
+    Copyright © 2023 Canonical Limited
+
+    Permission to use, copy, modify, distribute, and sell this
+    software and its documentation for any purpose is hereby granted
+    without fee, provided that the above copyright notice appear in
+    all copies and that both that copyright notice and this permission
+    notice appear in supporting documentation, and that the name of
+    the copyright holders not be used in advertising or publicity
+    pertaining to distribution of the software without specific,
+    written prior permission.  The copyright holders make no
+    representations about the suitability of this software for any
+    purpose.  It is provided "as is" without express or implied
+    warranty.
+
+    THE COPYRIGHT HOLDERS DISCLAIM ALL WARRANTIES WITH REGARD TO THIS
+    SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND
+    FITNESS, IN NO EVENT SHALL THE COPYRIGHT HOLDERS BE LIABLE FOR ANY
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="mir_shell_v1" version="1">
+    <description summary="create surface archetypes that are recognised by the desktop">
+      Clients can use this interface to assign an archetype to wl_surfaces.
+    </description>
+
+    <request name="get_normal_surface">
+      <description summary="create a normal surface from a surface">
+        Create a normal surface archetype for an existing surface. This assigns the archetype
+        of normal_surface, or raises a protocol error if another archetype is already
+        assigned and the transisition is disallowed.
+
+        A normal surface exists in the "bottom" layer (c.f. wlr_layer_shell_unstable_v1)
+      </description>
+      <arg name="id" type="new_id" interface="mir_normal_surface_v1"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
+    </request>
+
+    <request name="get_utility_surface">
+      <description summary="create a utility surface from a surface">
+        Create a utility surface for an existing surface. This assigns the archetype
+        of utility_surface, or raises a protocol error if another archetype is already
+        assigned and the transisition is disallowed.
+
+        A normal surface exists in the "top" layer (c.f. wlr_layer_shell_unstable_v1)
+        and will not be docked.
+      </description>
+      <arg name="id" type="new_id" interface="mir_utility_surface_v1"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
+    </request>
+
+    <request name="get_dialog_surface">
+      <description summary="create a dialog surface from a surface">
+        Create a dialog surface for an existing surface. This assigns the archetype
+        of dialog_surface, or raises a protocol error if another archetype is already
+        assigned and the transisition is disallowed.
+
+        TODO Text describing dialog behaviour
+      </description>
+      <arg name="id" type="new_id" interface="mir_dialog_surface_v1"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
+    </request>
+
+    <request name="get_satellite_surface">
+      <description summary="create a dialog surface from a surface">
+        Create a satellite surface for an existing surface. This assigns the archetype
+        of satellite_surface, or raises a protocol error if another archetype is already
+        assigned and the transisition is disallowed.
+
+        TODO Text describing satellite behaviour
+      </description>
+      <arg name="id" type="new_id" interface="mir_satellite_surface_v1"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
+      <arg name="positioner" type="object" interface="mir_positioner_v1"/>
+    </request>
+
+    <request name="get_freestyle_surface">
+      <description summary="create a freestyle surface from a surface">
+        Create a freestyle surface for an existing surface. This assigns the archetype
+        of freestyle_surface, or raises a protocol error if another archetype is already
+        assigned and the transisition is disallowed.
+
+        TODO Text describing freestyle behaviour
+      </description>
+      <arg name="id" type="new_id" interface="mir_freestyle_surface_v1"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
+      <arg name="positioner" type="object" interface="mir_positioner_v1" allow-null="true"/>
+      <!-- TODO optional attributes describing freestyle behaviour -->
+    </request>
+
+    <request name="create_positioner">
+      <description summary="create a positioner object">
+        Create a positioner object. A positioner object is used to position
+        surfaces relative to some parent surface. See the interface description
+        and xdg_surface.get_popup for details.
+      </description>
+      <arg name="id" type="new_id" interface="mir_positioner_v1"/>
+    </request>
+
+    <enum name="error">
+      <entry name="archetype" value="0" summary="wl_surface has another archetype that prevents this assignment"/>
+    </enum>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the mir_shell object">
+        This request indicates that the client will not use the mir_shell
+        object any more. Objects that have been created through this instance
+        are not affected.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="mir_normal_surface_v1" version="1">
+    <description summary="Mir metadata interface">
+      An interface that may be implemented by a wl_surface, for surfaces that
+      are designed to be rendered in a desktop-like environment.
+    </description>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the normal surface">
+        This request destroys the mir surface.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="mir_utility_surface_v1" version="1">
+    <description summary="Mir metadata interface">
+      An interface that may be implemented by a wl_surface, for surfaces that
+      are designed to be rendered in a desktop-like environment.
+    </description>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the utility_surface">
+        This request destroys the mir surface.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="mir_dialog_surface_v1" version="1">
+    <description summary="Mir metadata interface">
+      An interface that may be implemented by a wl_surface, for surfaces that
+      are designed to be rendered in a desktop-like environment.
+    </description>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the dialog_surface">
+        This request destroys the mir surface.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="mir_satellite_surface_v1" version="1">
+    <description summary="Mir metadata interface">
+      An interface that may be implemented by a wl_surface, for surfaces that
+      are designed to be rendered in a desktop-like environment.
+    </description>
+
+    <request name="reposition">
+      <description summary="recalculate the satellite's location">
+        Reposition an already-mapped satellite. The satellite will be placed given the
+        details in the passed mir_positioner object, and a
+        mir_satellite_surface_v1.repositioned followed by mir_satellite_surface_v1.configure and
+        wl_surface.configure will be emitted in response. Any parameters set
+        by the previous positioner will be discarded.
+
+        The passed token will be sent in the corresponding
+        xdg_satellite.repositioned event. The new satellite position will not take
+        effect until the corresponding configure event is acknowledged by the
+        client. See xdg_satellite.repositioned for details. The token itself is
+        opaque, and has no other special meaning.
+
+        If multiple reposition requests are sent, the compositor may skip all
+        but the last one.
+
+        If the satellite is repositioned in response to a configure event for its
+        parent, the client should send an mir_positioner.set_parent_configure
+        and possibly an mir_positioner.set_parent_size request to allow the
+        compositor to properly constrain the satellite.
+
+        If the satellite is repositioned together with a parent that is being
+        resized, but not in response to a configure event, the client should
+        send an mir_positioner.set_parent_size request.
+      </description>
+      <arg name="positioner" type="object" interface="mir_positioner_v1"/>
+      <arg name="token" type="uint" summary="reposition request token"/>
+    </request>
+
+    <event name="repositioned">
+      <description summary="signal the completion of a repositioned request">
+        The repositioned event is sent as part of a satellite configuration
+        sequence, together with mir_satellite_surface_v1.configure and lastly
+        wl_surface.configure to notify the completion of a reposition request.
+
+        The repositioned event is to notify about the completion of a
+        mir_satellite_surface_v1.reposition request. The token argument is the token passed
+        in the xdg_satellite.reposition request.
+
+        Immediately after this event is emitted, mir_satellite_surface_v1.configure and
+        wl_surface.configure will be sent with the updated size and position,
+        as well as a new configure serial.
+
+        The client should optionally update the content of the satellite, but must
+        acknowledge the new satellite configuration for the new position to take
+        effect. See mir_satellite_surface_v1.ack_configure for details.
+      </description>
+      <arg name="token" type="uint" summary="reposition request token"/>
+    </event>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the satellite_surface">
+        This request destroys the mir surface.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="mir_freestyle_surface_v1" version="1">
+    <description summary="Mir metadata interface">
+      An interface that may be implemented by a wl_surface, for surfaces that
+      are designed to be rendered in a desktop-like environment.
+    </description>
+
+    <request name="reposition">
+      <description summary="recalculate the freestyle's location">
+        Reposition an already-mapped freestyle. The freestyle will be placed given the
+        details in the passed mir_positioner object, and a
+        mir_freestyle_surface_v1.repositioned followed by mir_freestyle_surface_v1.configure and
+        wl_surface.configure will be emitted in response. Any parameters set
+        by the previous positioner will be discarded.
+
+        The passed token will be sent in the corresponding
+        xdg_freestyle.repositioned event. The new freestyle position will not take
+        effect until the corresponding configure event is acknowledged by the
+        client. See xdg_freestyle.repositioned for details. The token itself is
+        opaque, and has no other special meaning.
+
+        If multiple reposition requests are sent, the compositor may skip all
+        but the last one.
+
+        If the freestyle is repositioned in response to a configure event for its
+        parent, the client should send an mir_positioner.set_parent_configure
+        and possibly an mir_positioner.set_parent_size request to allow the
+        compositor to properly constrain the freestyle.
+
+        If the freestyle is repositioned together with a parent that is being
+        resized, but not in response to a configure event, the client should
+        send an mir_positioner.set_parent_size request.
+      </description>
+      <arg name="positioner" type="object" interface="mir_positioner_v1"/>
+      <arg name="token" type="uint" summary="reposition request token"/>
+    </request>
+
+    <event name="repositioned">
+      <description summary="signal the completion of a repositioned request">
+        The repositioned event is sent as part of a freestyle configuration
+        sequence, together with mir_freestyle_surface_v1.configure and lastly
+        wl_surface.configure to notify the completion of a reposition request.
+
+        The repositioned event is to notify about the completion of a
+        xdg_freestyle.reposition request. The token argument is the token passed
+        in the xdg_freestyle.reposition request.
+
+        Immediately after this event is emitted, mir_freestyle_surface_v1.configure and
+        wl_surface.configure will be sent with the updated size and position,
+        as well as a new configure serial.
+
+        The client should optionally update the content of the freestyle, but must
+        acknowledge the new freestyle configuration for the new position to take
+        effect. See wl_surface.ack_configure for details.
+      </description>
+      <arg name="token" type="uint" summary="reposition request token"/>
+    </event>
+
+    <!-- TODO requests/events describing freestyle behaviour -->
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the freestyle_surface">
+        This request destroys the mir surface.
+      </description>
+    </request>
+  </interface>
+
+  <interface name="mir_positioner_v1" version="1">
+    <description summary="child surface positioner">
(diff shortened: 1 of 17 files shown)


wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+    </request>
+
+    <request name="get_utility_surface">
+      <description summary="create a utility surface from a surface">
+        Create a utility surface for an existing surface. This assigns the archetype
+        of utility_surface, or raises a protocol error if another archetype is already
+        assigned and the transisition is disallowed.
+
+        A normal surface exists in the "top" layer (c.f. wlr_layer_shell_unstable_v1)
+        and will not be docked.
```
RAOF, line 52: What does this mean? Do you mean a *utility* surface exists in the "top" layer?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+        positioned relative to. If a corner anchor is set (e.g. 'top_left' or
+        'bottom_right'), the anchor point will be at the specified corner;
+        otherwise, the derived anchor point will be centered on the specified
+        edge, or in the center of the anchor rectangle if no edge is specified.
+      </description>
+      <arg name="anchor" type="uint" enum="anchor"
+           summary="anchor"/>
+    </request>
+
+    <enum name="gravity">
```
Saviq, line 377: I believe we discussed using north/south/east/west could be a better choice for gravity?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+    </enum>
+
+    <request name="set_anchor">
+      <description summary="set anchor rectangle anchor">
+        Defines the anchor point for the anchor rectangle. The specified anchor
+        is used derive an anchor point that the child surface will be
+        positioned relative to. If a corner anchor is set (e.g. 'top_left' or
+        'bottom_right'), the anchor point will be at the specified corner;
+        otherwise, the derived anchor point will be centered on the specified
+        edge, or in the center of the anchor rectangle if no edge is specified.
```
Saviq, line 371:
```suggestion
        edge, or in the center of the anchor rectangle if no anchor is specified.
```

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+
+    <request name="set_gravity">
+      <description summary="set child surface gravity">
+        Defines in what direction a surface should be positioned, relative to
+        the anchor point of the parent surface. If a corner gravity is
+        specified (e.g. 'bottom_right' or 'top_left'), then the child surface
+        will be placed towards the specified gravity; otherwise, the child
+        surface will be centered over the anchor point on any axis that had no
+        gravity specified. If the gravity is not in the ‘gravity’ enum, an
+        invalid_input error is raised.
```
Saviq, line 397: Do we need to spell out the "out of enum" case? If so, it's missing above.

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+      </description>
+      <arg name="gravity" type="uint" enum="gravity"
+           summary="gravity direction"/>
+    </request>
+
+    <enum name="constraint_adjustment" bitfield="true">
```
Saviq, line 403: Not sure I agree with the use of "constraint" / "constrained"… I mean, it's a constraint the positioner is applying on the adjustment, not the result of a constraint… gah.

What cases are we thinking of this taking effect?
- occlusion by other windows (?)
- extending beyond the work area
- ?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+
+          First try to slide towards the direction of the gravity on the x axis
+          until either the edge in the opposite direction of the gravity is
+          unconstrained or the edge in the direction of the gravity is
+          constrained.
+
+          Then try to slide towards the opposite direction of the gravity on the
+          x axis until either the edge in the direction of the gravity is
+          unconstrained or the edge in the opposite direction of the gravity is
+          constrained.
```
Saviq, line 435:
```suggestion
          Then try to slide against the direction of the gravity on the
          x axis until either the edge in the direction of the gravity is
          unconstrained or the edge against the direction of the gravity is
          constrained.
```

Though, since it's the same as above, just reversed, maybe we can say "Next, it does this in the opposite direction"?

I think we also miss what happens when sliding doesn't resolve things?

Same below for `y`.

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+      <entry name="flip_y" value="8">
+        <description summary="invert the anchor and gravity on the y axis">
+          Invert the anchor and gravity on the y axis if the surface is
+          constrained on the y axis. For example, if the bottom edge of the
+          surface is constrained, the gravity is 'bottom' and the anchor is
+          'bottom', change the gravity to 'top' and the anchor to 'top'.
+
+          The adjusted position is calculated given the original anchor
+          rectangle and offset, but with the new flipped anchor and gravity
+          values.
```
Saviq, line 474: This is missing above.

But also, DRY and refer to the above, "but on the Y axis"?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+        <description summary="horizontally resize the surface">
+          Resize the surface horizontally so that it is completely
+          unconstrained.
+        </description>
+      </entry>
+      <entry name="resize_y" value="32">
+        <description summary="vertically resize the surface">
+          Resize the surface vertically so that it is completely unconstrained.
+        </description>
+      </entry>
```
Saviq, line 491: Will min/max width/height and increments be taken into account?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+        the position of the surface, if the unadjusted position would result
+        in the surface being partly constrained.
+
+        Whether a surface is considered 'constrained' is left to the compositor
+        to determine. For example, the surface may be partly outside the
+        compositor's defined 'work area', thus necessitating the child surface's
+        position be adjusted until it is entirely inside the work area.
+
+        The adjustments can be combined, according to a defined precedence: 1)
+        Flip, 2) Slide, 3) Resize.
```
Saviq, line 415: Should this be left to the compositor, though? I suppose I'd need better examples to imagine how could resizing help, but sliding, not (being bigger than the work area, I suppose?).

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+      <arg name="constraint_adjustment" type="uint"
+           summary="bit mask of constraint adjustments"/>
+    </request>
+
+    <request name="set_offset">
+      <description summary="set surface position offset">
+        Specify the surface position offset relative to the position of the
+        anchor on the anchor rectangle and the anchor on the surface. For
+        example if the anchor of the anchor rectangle is at (x, y), the surface
+        has the gravity bottom|right, and the offset is (ox, oy), the calculated
```
Saviq, line 519:
```suggestion
        has the gravity bottom_right, and the offset is (ox, oy), the calculated
```

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+        surface position will be (x + ox, y + oy). The offset position of the
+        surface is the one used for constraint testing. See
+        set_constraint_adjustment.
+
+        An example use case is placing a popup menu on top of a user interface
+        element, while aligning the user interface element of the parent surface
+        with some user interface element placed somewhere in the popup surface.
+      </description>
```
Saviq, line 527: Above we say that you can't have a child window positioned "away" from the parent, but this has no such limitation, or discusses an error condition in that case?

Saviq (changes requested):
I've focused on the protocol definition here, a chunk of questions / tweaks inline.

FWIW I don't think we should be referring layer_shell, even if they share concepts. We don't want to introduce a dependency.

Or are we saying that these could replace layer shell? E.g. since Freestyle is documented with "It should be rarely used and require unusual steps to enable" - should it be a separate interface, or mediated somehow?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="mir_shell_v1" version="1">
+    <description summary="create surface archetypes that are recognised by the desktop">
+      Clients can use this interface to assign an archetype to wl_surfaces.
+    </description>
+
+    <request name="get_normal_surface">
```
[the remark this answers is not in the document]
AlanGriffiths: I guess I had read `mir_window_type_normal` just before writing this. It should change

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+      <description summary="create a normal surface from a surface">
+        Create a normal surface archetype for an existing surface. This assigns the archetype
```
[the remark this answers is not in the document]
AlanGriffiths: True. But why are you asking? (FWIW **xdg-shell** doesn't use "role" in its `get_` methods either.)

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,532 @@
...
+    </request>
+
+    <request name="get_utility_surface">
+      <description summary="create a utility surface from a surface">
+        Create a utility surface for an existing surface. This assigns the archetype
+        of utility_surface, or raises a protocol error if another archetype is already
+        assigned and the transisition is disallowed.
+
+        A utility surface exists in the "top" layer (c.f. wlr_layer_shell_unstable_v1)
+        and will not be docked.
```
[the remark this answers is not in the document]
AlanGriffiths: Yes, I need to clean up this terminology. For reference:

```c++
typedef enum MirDepthLayer
{
    mir_depth_layer_background,         /**< For desktop backgrounds and alike (lowest layer) */
    mir_depth_layer_below,              /**< For panels or other controls/decorations below normal windows */
    mir_depth_layer_application,        /**< For normal application windows */
    mir_depth_layer_always_on_top,      /**< For always-on-top application windows */
    mir_depth_layer_above,              /**< For panels or notifications that want to be above normal windows */
    mir_depth_layer_overlay,            /**< For overlays such as lock screens (heighest layer) */
} MirDepthLayer;
```

"heighest"!

wayland-protocols/mir-shell-unstable-v1.xml
AlanGriffiths: Ah, missed that this text needs rework after paste.

Regarding a possible "error", the compositor fixes it (applying the `constraint_adjustment`, if applicable).

AlanGriffiths:
> FWIW I don't think we should be referring layer_shell, even if they share concepts. We don't want to introduce a dependency.

Agreed.

> Or are we saying that these could replace layer shell? E.g. since Freestyle is documented with "It should be rarely used and require unusual steps to enable" - should it be a separate interface, or mediated somehow?

I'm happy to drop freestyle for the time being. It is (even in the old design docs) under-specified, it was here as a placeholder in case review identified a reason to fill that design out.

wayland-protocols/mir-shell-unstable-v1.xml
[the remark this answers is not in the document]
Saviq: I just mean that it's inconsistent - here we create the archetype, below we create surfaces? Just needs consistent wording, is all.

Saviq (commented):
> I'm happy to drop freestyle for the time being. It is (even in the old design docs) under-specified, it was here as a placeholder in case review identified a reason to fill that design out.

IMO that makes sense, yeah.

wayland-protocols/mir-shell-unstable-v1.xml
AlanGriffiths: It is left to the compositor (both in our design and in the xdg-shell design). The application simply requests

Saviq: I mean even the "defined precedence", should that be part of the protocol definition? I can be convinced both ways, really.

AlanGriffiths: It is part of the protocol for xdg-shell, we don't have a strong reason to do things differently

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,452 @@
...
+    SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="mir_shell_v1" version="1">
+    <description summary="create surface archetypes that are recognised by the desktop">
+      Clients can use this interface to assign an archetype to wl_surfaces.
```
RAOF, line 30:
```suggestion
      Clients can use this interface to assign an archetype to wl_surfaces.
      
      Archetypes 
```

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,452 @@
...
+    </description>
+
+    <request name="get_regular_surface">
+      <description summary="assign the mir_regular_surface_v1 archetype">
+        Create a mir_regular_surface_v1 archetype for the given surface. This assigns
+        the archetype of regular_surface to the wl_surface, or raises a protocol error
+        if another archetype is already assigned and the transition is disallowed.
```
RAOF, line 37: What should happen to the previous archetype's protocol object in this case?

```suggestion
        if another archetype is already assigned and the transition is disallowed.
        The previous archetype's role object becomes inert; the client should destroy
        the previous archetype role object. Any further calls to the previous archetype
        role object are a protocol error.
```

RAOF, line 37: How does a client know what transitions are allowed?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,452 @@
...
+      <arg name="surface" type="object" interface="wl_surface"/>
+    </request>
+
+    <request name="get_floating_regular_surface">
+      <description summary="assign the mir_floating_regular_surface_v1 archetype">
+        Create a mir_floating_regular_surface_v1 archetype for the given surface. This assigns
+        the archetype of floating_regular to the wl_surface, or raises a protocol error
+        if another archetype is already assigned and the transition is disallowed.
+
+        A floating regular surface exists in the mir_depth_layer_always_on_top layer and will
```
RAOF, line 51: Again, this seems like an implementation detail leakage.

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,452 @@
...
+      <description summary="signal the completion of a repositioned request">
+        The repositioned event is sent as part of a satellite configuration
+        sequence, together with mir_satellite_surface_v1.configure and lastly
+        wl_surface.configure to notify the completion of a reposition request.
+
+        The repositioned event is to notify about the completion of a
+        mir_satellite_surface_v1.reposition request. The token argument is the token passed
+        in the xdg_satellite.reposition request.
+
+        Immediately after this event is emitted, mir_satellite_surface_v1.configure and
```
RAOF, line 190: `mir_satellite_surface_v1.configure` doesn't appear to exist?

RAOF (commented):
A global question: the archetype presumably wants to be double-buffered state, so that both client and server can have a consistent view of things. Should `get_*_surface` require the `wl_surface` to be committed before taking effect?

Since the surfaces we're talking about here are almost certainly also `xdg_toplevel`s, how does this integrate with the `xdg_surface` mapping/configure dance?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,452 @@
...
+      Clients can use this interface to assign an archetype to wl_surfaces.
+    </description>
+
+    <request name="get_regular_surface">
+      <description summary="assign the mir_regular_surface_v1 archetype">
+        Create a mir_regular_surface_v1 archetype for the given surface. This assigns
+        the archetype of regular_surface to the wl_surface, or raises a protocol error
+        if another archetype is already assigned and the transition is disallowed.
+
+        A normal surface exists in the mir_depth_layer_application layer
```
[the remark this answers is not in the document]
AlanGriffiths: I'm open to suggestions as to what we should be saying about layers. It seems excessive to define the layers here simply for documentation purposes. But maybe that is the best way to make things clear?

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,452 @@
...
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="mir_shell_v1" version="1">
+    <description summary="create surface archetypes that are recognised by the desktop">
+      Clients can use this interface to assign an archetype to wl_surfaces.
+    </description>
+
+    <request name="get_regular_surface">
```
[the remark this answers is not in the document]
AlanGriffiths: We do: An `xdg_toplevel` is "freestyle" by default and the window management for that is distinct from "regular".

wayland-protocols/mir-shell-unstable-v1.xml
AlanGriffiths: I need to document a lot more of this section of "Mir and Unity: Surfaces, input, and displays" somewhere. I put off doing that because I felt it more urgent to get this discussion going.

wayland-protocols/mir-shell-unstable-v1.xml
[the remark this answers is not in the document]
RAOF: I think we can actually just go the other way, and not mention layers *at all*; just that "regular floating" means "always on top".

wayland-protocols/mir-shell-unstable-v1.xml
RAOF: Ah, good. I wasn't sure if the plan was to dump a bunch of that in here.

RAOF (commented):
Top level question: we need to document how this interacts with `xdg_wm_base`. Particularly: The archetype of a window is going to potentially change client-visible state - we'll want to send different initial `configure` events for a regular window vs a satellite positioned near the edge of the screen. It will *definitely* influence the `xdg_wm_base` events we send out - for example, a dialog is not mimimisable, so we'll want to *not* send `minimized` in the `wm_capabilities` event for such a window, but a normal window *is* minimisable, so we need to include `minimized` in `wm_capabilities` for such a window.

Likewise, changing archetype is likely to generate client-visible `xdg_surface` or `xdg_toplevel` events.

This suggests that these state changes need to double-buffered state bound to `wl_surface.commit()` and be tied in to the `xdg_surface` `configure`/`ack_configure` dance.

I'd suggest that we specify that an archetype SHOULD be set in the initial buffer-less `configure` of the `xdg_surface` construction dance, and that it behaves as `xdg_surface`-role state.

Because we need to specify in such detail how this interacts with `xdg_wm_base`, I think that there's little point in having this act on `wl_surface`s rather than explicitly on `xdg_toplevel`s. `wl_shell` is dead to the point that major compositors (*at least* GNOME Shell) no longer advertise it *at all*, and should any other shell interface come along we would likewise need to specify how `mir_shell_v1` interacts with the new shell system, which would need a protocol revision.

AlanGriffiths:
> I'd suggest that we specify that an archetype SHOULD be set in the initial buffer-less `configure` of the `xdg_surface` construction dance, and that it behaves as `xdg_surface`-role state.

That ignores the possibility of changing archetypes (that's more text I need to incorporate)

RAOF:
> > I'd suggest that we specify that an archetype SHOULD be set in the initial buffer-less `configure` of the `xdg_surface` construction dance, and that it behaves as `xdg_surface`-role state.
> 
> That ignores the possibility of changing archetypes (that's more text I need to incorporate)

No, you'd still be able to change archetypes (although we should also mention what dance should be done to do that). My suggestion is that the *initial* archetype should be set in the initial `xdg-surface` `configure` dance.

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,509 @@
...
+      An archetype is like a `wl_surface` role: a surface may have at most one
+      archetype, but unlike a `wl_surface` role a client may assign a new archetype
+      to a surface which already has one. This will remove the old archetype and
+      apply the new one atomically.
+
+      The archetype of a `wl_surface` influences the window-management policies
+      applied to it.
+
+      The client must call wl_surface.commit on the corresponding wl_surface
+      for the archetype to take effect.
```
RAOF, line 41: I think this is about what I mean with the integration with `xdg_wm_base`. 

```suggestion
    All archetype state is double-buffered; changing a surface archetype or updates to any archetype state are not applied until the `wl_surface` is `commit`ted.
    
    Archetype state augments `xdg_toplevel` state. If a client intends to use an archetype, it SHOULD assign the surface an initial archetype during the initial commit without a buffer attached (see `xdg_surface`).
    
    Any events sent to an archetype object are latched to and extend the `xdg_surface.configure` event. Any such events should be treated as a part of an atomic set of configuration changes (including any `xdg_toplevel` events) with the `xdg_surface.configure` event committing the accumulated state and requiring a `xdg_surface.ack_configure` call, as normal.
    
    Changing archetype follows a similar sequence to the initial `xdg_surface` commit. First, a new archetype must be committed. This must be the only state committed in the `wl_surface.commit` request. The compositor will respond with any changes to the surface state prompted by the change of archetype followed by an `xdg_surface.configure` event. The client must ack the configure event (as normal) and the subsequent `wl_surface.commit` will cause the new archetype to be fully applied.
```

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,509 @@
...
+    AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION,
+    ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF
+    THIS SOFTWARE.
+  </copyright>
+
+  <interface name="mir_shell_v1" version="1">
+    <description summary="create surface archetypes that are recognised by the desktop">
+      Clients can use this interface to assign an archetype to wl_surfaces.
+
+      An archetype is like a `wl_surface` role: a surface may have at most one
```
RAOF, line 32: Oh! New question: what happens if you try and apply an archetype to a `wl_surface` that has the `wl_subsurface` role, or `wl_cursor` role, or `xdg_popup` role?

AlanGriffiths: Or setting a role on a window that has one of these archetypes? A protocol error, I guess.

wayland-protocols/mir-shell-unstable-v1.xml
```diff
@@ -0,0 +1,533 @@
...
+      committed in the `wl_surface.commit` request. The compositor will respond with
+      any changes to the surface state prompted by the change of archetype followed
+      by an `xdg_surface.configure` event. The client must ack the configure event
+      (as normal) and the subsequent `wl_surface.commit` will cause the new
+      archetype to be fully applied.
+
+      There are some `wl_surface` roles that conflict with the archetypes described
+      here. Attempting to set both an archetypes and such roles on a surface is a
+      protocol error. In particular, `wl_subsurface`, `wl_cursor` or `xdg_popup` may
+      not be combined with an archetype.
```
RAOF, line 65: Since the behaviour of “you tried a bad combination” is “your application crashes” and there's no other way than reading the protocol of discovering what combinations are bad I don't think we can be anything but exhaustive in enumerating the valid combinations here.

What `wl_surface` roles *should* be supported? `xdg_toplevel`, and…? *Maybe* layer shell surfaces, but they're already pretty special and I don't think any of this window management applies to them. `wl_shell` is dead dead dead and no client is going to use it in new code.

This does further suggest that the correct object type for this extension to act on is `xdg_toplevel` rather than `wl_surface`.

RAOF:
The last comment aside, I think this is pretty good to go. We can refine the protocol over time, but I think the foundations are solid enough here to usefully start on a spike.

wayland-protocols/mir-shell-unstable-v1.xml
AlanGriffiths: > This does further suggest that the correct object type for this extension to act on is `xdg_toplevel` rather than `wl_surface`.

I don't, at this stage, want to tie `mir-shell` to `xdg-shell`. As specified here it is compatible with `wl-shell`. I know `wl-shell` is deprecated, but I want to be open to a hypothetical alternative (for AR perhaps?) and it still exists.

In future we will likely want to propose something based on experience to wayland-protocols, that is the time to think more closely about integration with (or extension of) `xdg-shell`.

Relying on `xdg_toplevel` and `xdg_positioner` is clearly a plausible change if we choose that.

AlanGriffiths: > What `wl_surface` roles _should_ be supported? `xdg_toplevel`, and…? _Maybe_ layer shell surfaces, but they're already pretty special and I don't think any of this window management applies to them. `wl_shell` is dead dead dead and no client is going to use it in new code.

An alternative is to include the other archetypes Mir currently deduces from the `xdg-shell` role & attributes. And document how that happens. Then we need only talk about archetypes?

Layer-shell, is something orthogonal

Saviq:
Flipping for CI verification.

Issue #3165, opened 2023-12-12
Changing scale factor at runtime leads to misrendering


RAOF:
When using `gbm-kms` as the display platform and Mir is started with `--display-config static=$FILENAME`, changing the scale for an output in `$FILENAME` and saving it will result in Mir mis-rendering the scene on that output.

Specifically, the Renderer does not take the scaling into account so if an output started with `scale: 1` and was modified to `scale: 2`, only the top-left-hand quarter of the display will contain output (although if you move a window so some of it is outside the logical screen space, it will be rendered into the otherwise black section of the output).

This appears to be due to the reconfiguration process assigning the wrong logical size to the relevant `DisplaySink`; in `OverlappingOutputGroup` it's taken from the existing logical size of the output.

AlanGriffiths:
Synchronicity! I just ran into this too.

AlanGriffiths:
And, after checking out miriway --revision 2114 (Mir 2.15), this is new with 2.16

AlanGriffiths:
Another weird thing to note: if the `.display` is rewritten without further changes another reconfiguration happens (but subsequent rewrites don't cause further reconfigurations)

PR #3169, opened 2023-12-15
mg::MultiplexingDisplay: Copy *all* mutable fields on user configuration update

RAOF:
We missed copying the `scale` field from the copy of the configuration handed out to be mutated during configuration changes.

We also missed a couple of other fields. They're less interesting, but also copy them while we're here.

Fixes: #3165


RAOF:
A good demonstration of the limitations of coverage tracking 🤪

15 Dec 2023 18:08:46 codecov[bot] ***@***.***>:

AlanGriffiths (approved):
Looks right, and cures much of the rescaling problem.

(Testing with Miriway/mir-pr3169, I don't see the swaybg background resize until I save the file a second time - but that was happening on 2.15 too, so probably a different cause)

PR #3174, opened 2024-01-04
Simplify code

AlanGriffiths:
Many years ago we tried to get some `libinput_event_touch_` functions upstreamed to libinput, but they were rejected. Ubuntu no longer carries the patch that implements them.

There are simpler ways to do what this code did (e.g. resolve the possible library functions at compile time or make the function pointers global) but, as the functions are never available, it doesn't need to do it anyway.

The new version will give a compile error if the functions are introduced into libinput and we can fix the code appropriately if that happens.


RAOF (approved):
Huh, I was sure I'd reviewed this before!

Yeah, this code was weird when I looked at it recently, too :grin:

PR #3176, opened 2024-01-05
feature: remove mir cookie

mattkae:
## What's new?
- mircookie and its friends are removed
- This breaks ABI


RAOF (approved):
This looks good to me.

We might want *something* similar for [xdg-activation](https://wayland.app/protocols/xdg-activation-v1), but since that has a client-requests-token -> server-generates-token loop we shouldn't need any of the cookie-for-every-event infrastructure this removes.

If some of the code turns out to be useful there, we can easily fish it out of history.

include/core/mir_toolkit/events/event_type_to_string.h
AlanGriffiths, line 1: I think this header belongs in `include/common`, not `include/core` - see where the function is implemented.

src/miral/regenerate-miral-symbols-map.py
```diff
@@ -411,7 +411,6 @@ def _report(publish, symbol):
     miral::toolkit::mir_input_event_get_pointer_event*;
     miral::toolkit::mir_input_event_get_touch_event*;
     miral::toolkit::mir_input_event_get_type*;
-    miral::toolkit::mir_input_event_has_cookie*;
```
AlanGriffiths, line 414: We should be bumping ABI, not dropping symbols without changing soname

AlanGriffiths (changes requested):
> * This breaks ABI

We need to do some ABI bumping then. I don't mind if that is a separate PR, but needs to follow shortly.

As most of our libraries use events, I imagine they all need bumping. (mircore is probably OK)

src/miral/regenerate-miral-symbols-map.py
mattkae: Yeah this might be one that I don't understand. Should I leave all of the `.map` files unchanged, or am I supposed to remove the symbols from them now?

Also, is there any simple mechanism for bumping the so names that need to be bumped?

AlanGriffiths (approved):
This looks good except for the symbol/soname versioning. That is addressed in a dependent PR: #3178.

I think the best approach is to merge that into this PR before landing. But marking my approval

mattkae:
@AlanGriffiths FYI, it appears that the failing Fedora tests are failing on main too.

AlanGriffiths (approved):

PR #3178, opened 2024-01-09
Bumping mircommon, miral, and mirserver sonames for mircookie work

mattkae:


src/miral/regenerate-miral-symbols-map.py
```diff
@@ -555,11 +555,11 @@ def _report(publish, symbol):
   local: *;
 };
 
-MIRAL_4.1 {
+MIRAL_4.2 {
 global:'''
 
 END_NEW_STANZA = '''
-} MIRAL_4.0;
+} MIRAL_4.1;
```
AlanGriffiths, line 562: We need to clear these symbols from the script and let it fill in the current stanza. See 8e3130d7f30204b6d2545694218dab935f68e8a2 for an example

src/server/symbols.map
```diff
@@ -1,4 +1,4 @@
-MIR_SERVER_2.15 {
+MIR_SERVER_2.16 {
```
AlanGriffiths, line 1: `MIR_SERVER_2.17` - 2.16 has already been released

src/common/symbols.map
```diff
@@ -1,4 +1,4 @@
-MIR_COMMON_2.8 {
+MIR_COMMON_2.15 {
```
AlanGriffiths, line 1: `MIR_SERVER_2.17` - 2.15 has already been released

src/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 set(MIRPLATFORM_ABI 27)
 
 set(MIRAL_VERSION_MAJOR 4)
-set(MIRAL_VERSION_MINOR 1)
+set(MIRAL_VERSION_MINOR 2)
```
AlanGriffiths, line 5: 5.0, not 4.2

AlanGriffiths (changes requested):
Thanks for trying, but you're not done yet!

For the libs you've touched, there need to be changes to libmiral6.symbols

And there are more libs to touch: I *think* that mircore is the only ABI that isn't being broken.

For example, `miroil` depends on `mircommon`. So a process needs to be able to load a consistent versions of both libraries. That is only possible with a bump to the miroil .soname.

The same applies to, for example, platforms

src/server/symbols.map
mattkae: Cool! Why did it say 2.15? Because we didn't introduce any new symbols on the prev release?

mattkae:
> For the libs you've touched, there need to be changes to libmiral6.symbols

Do I do this part by hand?

> For example, `miroil` depends on `mircommon`. So a process needs to be able to load a consistent versions of both libraries. That is only possible with a bump to the miroil .soname.
> 
> The same applies to, for example, platforms

Kk makes sense!

mattkae:
@AlanGriffiths does that include mirwayland?

src/server/symbols.map
AlanGriffiths: Exactly - the stanza is the release the symbol was introduce to the current .soname

AlanGriffiths:
> @AlanGriffiths does that include mirwayland?

```plain
target_link_libraries(mirwayland
  PUBLIC
    mircore
    PkgConfig::WAYLAND_SERVER
  PRIVATE
    mircommon
)
```

AlanGriffiths:
> > For the libs you've touched, there need to be changes to libmiral6.symbols
> 
> Do I do this part by hand?

The check-and-update-debian-symbols.py (runs as part of a build) adds missing symbols. So just strip all the symbols and `cmake -B ...`.

src/miroil/symbols.map
AlanGriffiths, line 1: There's a regenerate-miroil-symbols-map.py (and build target) too

AlanGriffiths (changes requested):
We also need to bump the platform ABIs

- MIR_SERVER_INPUT_PLATFORM_ABI
- MIR_SERVER_GRAPHICS_PLATFORM_ABI

RAOF:
I am confused by the failures here. It looks like miroil symbols *used* to be just ignored (`libmiroil3.symbols` when the library was `libmiroil.so.4`), and the miroil symbols diff is removing and then adding back the same symbols.

I'll have another look on Monday if this still fails. Hopefully I'll be fresher then!

AlanGriffiths (changes requested):
I think part of the problem is #3184

AlanGriffiths:
@mattkae leave this a bit. I'm debugging the helper stuff

AlanGriffiths:
@mattkae I have to dive into #3185 now, so good luck. (I've fixed some things, but the proof will be in CI. )

Some useful incantations for running the symbol (re)generation scripts:

    cmake --build cmake-build-release --target regenerate-miral-symbols-map
    cmake --build cmake-build-release --target regenerate-miral-debian-symbols

    cmake --build cmake-build-release --target regenerate-miroil-symbols-map
    cmake --build cmake-build-release --target check-miroil-symbols

src/miroil/symbols.map
```diff
@@ -1,4 +1,4 @@
-MIROIL_2.0 {
+MIROIL_5.0 {
```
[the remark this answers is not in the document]
mattkae: Why should this be 2.17? Is this supposed to be the mir version, not the miroil version?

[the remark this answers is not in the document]
AlanGriffiths: When I wrote that I was thinking the mir version is more logical. But the check-miroil-symbols script says otherwise.

mattkae:
@AlanGriffiths Ok i think this is good to go again. The symbols that were bumped are:

- miral
- miroil
- mircommon
- mirserver
- mirwayland
- mir-platform-graphics-*
- mir-platform-rendering
- mir-platform-input-*

Am I missing anything?

AlanGriffiths (changes requested):
> Am I missing anything?

```
target_link_libraries(mirplatform
  PRIVATE
    mircommon
    ${MIR_PLATFORM_REFERENCES}
  PUBLIC
    PkgConfig::EPOXY
)
```

So, `libmirplatform` also needs an soname bump

PR #3186, opened 2024-01-12
Change placement of new floating windows

### LINKED ISSUE #3175: floating: place new windows offset from previous placement
Steps:
- run miral-app
- run e.g. gnome-calculator twice or more

Expected:
- you can see multiple windows of gnome-calculator

Current:
- you can only see the newest window of gnome-calculator, since it's placed exactly on top of the previous one

hbatagelo:
Fixes #3175 using a combination of cascading and random positioning.

Rather than replacing the original algorithm, it has been extended. If the new optically centered window shares the same top-left position as an existing window within a margin of tolerance, the new window is cascaded both vertically and horizontally. Following cascading, if any part of the window extends beyond the display area, it is randomly placed within a "feasible placement region" defined as the morphological erosion of the display region by the window region. If this region is empty, the window is resized horizontally and/or vertically to fit the display area and is also moved to the top and/or left of the display area.

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -29,6 +29,8 @@
 #include <boost/throw_exception.hpp>
 
 #include <algorithm>
+#include <random>
+#include <unordered_set>
 
 using namespace mir;
 using namespace mir::geometry;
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
@hbatagelo no need to change this PR, but you're now able to create branches in MirServer/mir. That has the advantage that snaps will be built from the branch which facilitates testing and review. Using another repo means the [snap builds are skipped](https://github.com/MirServer/mir/actions/runs/7508036981/job/20443094904?pr=3186).

PR #3187, opened 2024-01-15
Change placement of new floating windows

### LINKED ISSUE #3175: floating: place new windows offset from previous placement

hbatagelo:
Fixes #3175 using a combination of cascading and random positioning.

Rather than replacing the original algorithm, it has been extended. If the new optically centered window shares the same top-left position as an existing window within a margin of tolerance, the new window is cascaded both vertically and horizontally. Following cascading, if any part of the window extends beyond the display area, it is randomly placed within a "feasible placement region" defined as the morphological erosion of the display region by the window region. If this region is empty, the window is resized horizontally and/or vertically to fit the display area and is also moved to the top and/or left of the display area.

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -29,8 +29,6 @@
 #include <boost/throw_exception.hpp>
 
 #include <algorithm>
-#include <random>
-#include <unordered_set>
 
 using namespace mir;
 using namespace mir::geometry;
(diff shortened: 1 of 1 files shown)


hbatagelo:
Okay, I'm good with a simpler algorithm, checking the center and corners instead of cascading. Should I keep random placement as a fallback, and the resizing when the window is bigger than the display area?

AlanGriffiths:
If you're only checking positions against visible windows then it is unlikely that all of the five positions are occupied. (Because, for example, one at NW probably occludes ones at centre, SW, SE and NE.) So fallback the centre is probably good enough, and less confusing than "random".

src/miral/basic_window_manager.cpp
```diff
@@ -2016,6 +2016,36 @@ auto miral::BasicWindowManager::place_new_surface(WindowSpecification parameters
...
+                auto const window{window_at(position)};
+                auto const window_type{info_for(window).type()};
+                auto const window_state{info_for(window).state()};
+
+                if (!window ||
+                    window_state != mir_window_state_restored ||
+                    (window_type != mir_window_type_normal &&
+                     window_type != mir_window_type_decoration &&
+                     window_type != mir_window_type_freestyle) ||
+                     window.top_left() != position)
```
AlanGriffiths, line 2041: The logic here seems wrong.

1. `!window` cannot happen: a logic error would have been thrown by `info_for`
2. We don't want to exclude restored windows
3. We don't want to exclude normal and freestyle windows

src/miral/basic_window_manager.cpp
```diff
@@ -2016,6 +2016,36 @@ auto miral::BasicWindowManager::place_new_surface(WindowSpecification parameters
...
+            std::array const positions {
+                parameters.top_left().value(),
+                parameters.top_left().value() + Displacement( offset,  offset),
+                parameters.top_left().value() + Displacement(-offset,  offset),
+                parameters.top_left().value() + Displacement( offset, -offset),
+                parameters.top_left().value() + Displacement(-offset, -offset)};
+
+            for (auto const& position : positions)
+            {
+                auto const window{window_at(position)};
```
AlanGriffiths, line 2032:
```suggestion
                if (auto const window{window_at(position)})
                {
                    ...
```

AlanGriffiths (changes requested):
There are a lot of tests failing

AlanGriffiths:
> There are a lot of tests failing

To run the tests locally:

    cmake --build <build directory> --target ptest

src/miral/basic_window_manager.cpp
AlanGriffiths: Sorry, I misread the condition (which itself suggests it could be clearer).

AlanGriffiths:
I tried installing and running this:

```plain
snap install mir-test-tools_pr3187 --channel=22/edge/mir-pr3187
WAYLAND_DISPLAY=wayland-99 mir-test-tools_pr3187.demo-server&
WAYLAND_DISPLAY=wayland-99 gnome-chess&
```

But see the error:

```
[2024-01-16 15:02:02.461418] < -warning- > frontend:Wayland: Exception processing Surface::commit() request: Dynamic exception type: std::out_of_range
std::exception::what: map::at

error in client communication (pid 25499)
Gdk-Message: 15:02:02.461: Error 71 (Protocol error) dispatching to Wayland display.
```

AlanGriffiths:
> I tried installing and running this:

Building locally doesn't show this effect (and the code looks like it shouldn't). Possibly, the snap store has an old version...

src/miral/basic_window_manager.cpp
```diff
@@ -2016,6 +2016,32 @@ auto miral::BasicWindowManager::place_new_surface(WindowSpecification parameters
...
+                    !window ||
+                    info_for(window).state() != mir_window_state_restored ||
+                    (info_for(window).type() != mir_window_type_normal &&
+                     info_for(window).type() != mir_window_type_decoration &&
+                     info_for(window).type() != mir_window_type_freestyle) ||
+                    window.top_left() != position)
+                {
+                    parameters.top_left().value() = position;
+                    break;
+                }
```
AlanGriffiths, line 2042:
```suggestion
                auto const window{window_at(position)};

                static auto const ignored_state_or_type = [](WindowInfo const& info)
                    {
                        switch (info.type())
                        {
                        case mir_window_type_normal:
                        case mir_window_type_decoration:
                        case mir_window_type_freestyle:
                            return info.state() != mir_window_state_restored;

                        default:
                            return true;
                        }
                    };

                if (!window || ignored_state_or_type(info_for(window)) || window.top_left() != position)
                {
                    parameters.top_left().value() = position;
                    break;
                }
```

AlanGriffiths (approved):
The offset used is (obviously) arbitrary and may seem big or small depending on the screen dimensions. (Maybe a fraction of the screen width/height would be better?)

I think the conditional is a bit hard to read, but isn't wrong. I've suggested a rework that might be easier to read (but I am probably biased).

Anyway, does what was asked. So I'll approve and see if anyone else has an opinion.

(Feel free to land with or without my suggestion)

mattkae:
I found a bug that is _maybe_ unrelated to this. It looks like during the startup of miral-shell, all of the windows end up in the same position. In the attached video, I am opening up multiple windows in a row. WDYT @AlanGriffiths , is this blocking?


[Screencast from 01-16-2024 12:15:06 PM.webm](https://github.com/MirServer/mir/assets/25062299/28d8fb84-6882-4bf7-be52-c819ccbacf6a)

hbatagelo:
I think these first windows are placed at the same position because the splash screen is being displayed. Maybe the splash screen is picked by `window_at` instead of the terminal windows. I'll check.

mattkae:
> I think these first windows are placed at the same position because the splash screen is being displayed. Maybe the splash screen is picked by `window_at` instead of the terminal windows. I'll check.

Oh good call, that might be it. This could be a separate bug for the splash screen though. If so, we can do it separately :)

AlanGriffiths:
> WDYT @AlanGriffiths , is this blocking?

I to assume `window_at` is returning the splash screen (because it is on top). Which is a special case I wouldn't block on (you might have a different opinion).

AlanGriffiths:
> Oh good call, that might be it. This could be a separate bug for the splash screen though. If so, we can do it separately :)

Well, if it is a bug, it is in `window_at()` not the splash screen.

mattkae:
> I to assume `window_at` is returning the splash screen (because it is on top). Which is a special case I wouldn't block on (you might have a different opinion).

I think it's non-blocking too. As you said, probably a bug in `window_at` that we can record and fix later

hbatagelo:
I tested locally and can confirm that `window_at` (wrongly) selects the splash screen. Since it is a top-level, freestyle, fullscreen window, placement is allowed at the initial position.

AlanGriffiths:
> Well, if it is a bug, it is in `window_at()` not the splash screen.

I change my opinion: it is the splash window. This _should_ be transparent to input, and that would ensure it isn't returned by `surface_at`.

But this is pre-existing (e.g. you can't drag a window until the splash animation ends), and not very important.

PR #3191, opened 2024-01-18
Add missing header for std::count_if

jwakely:


AlanGriffiths (approved):
Thanks, Jon!

jwakely:
Oh sorry, I didn't see you'd set this to merge. I just force-pushed to the same branch to fix another missing `<algorithm>`.

jwakely:
... because `rendering_tracker.cpp` uses `std::set_intersection`

FWIW we've documented the changes in transitive includes for GCC 14 here:
https://gcc.gnu.org/gcc-14/porting_to.html#header-dep-changes

jwakely:
And there are more. I'll change this to draft status until I've found them all ...

jwakely:
OK, I think it's ready now ... at least based on the fixes needed for the 2.15.0 release that we have in Fedora. It's possible `main` has others I didn't find.

jwakely:
If the problems you see are in rawhide, this PR might have fixed 'em! 2.15.0 builds for me now (although I'm destabilising things by pushing a new boost to rawhide, so let the dust settle).

PR #3192, opened 2024-01-18
Fix rawhide FTBFS

AlanGriffiths:
```
 /spread/mir/examples/client/wayland_client.c:533:39: error: ‘calloc’ sizes specified with ‘sizeof’ in the earlier argument and not in the later argument [-Werror=calloc-transposed-args]
  533 |     draw_context* ctx = calloc(sizeof *ctx, 1);
      |                                       ^
/spread/mir/examples/client/wayland_client.c:533:39: note: earlier argument should specify number of elements, later size of each element
cc1: all warnings being treated as errors
```

### DIFF
--- a/src/platform/graphics/egl_buffer_copy.cpp
+++ b/src/platform/graphics/egl_buffer_copy.cpp
@@ -66,13 +66,13 @@ class GLBulkHandle
 public:
     GLBulkHandle()
     {
-        (*allocator)(1, &id);
+        allocator(1, &id);
     }
 
     ~GLBulkHandle()
     {
         if (id)
-            (*deleter)(1, &id);
+            deleter(1, &id);
     }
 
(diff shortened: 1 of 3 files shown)


hbatagelo (approved):
LGTM

AlanGriffiths:
There are more rawhide failures to fix. But I won't have time before my EOD

src/platform/graphics/egl_buffer_copy.cpp
hbatagelo, line 1: Why not simply use `allocator(1, &id)`, `deleter(id)`, etc? A function call [should work](https://eel.is/c++draft/expr.call#1) with a function pointer as the expression.

AlanGriffiths: True. I confess I only thought about fixing the compile error, not fixing the code.

PR #3197, opened 2024-01-22
Drive-by tidy up of libinput use

AlanGriffiths:
Tidying input code that should be clearer. Mostly simplifying `mie::LibInputDevice` which is confuses about whether it represents l`ibinput_device*` or `libinput_device_group*`.

Before:

`mie::LibInputDevice` has a data member `devices` that is populated via `add_device_of_group()` that would appear to be intended to reflect the `libinput_device_group`.

However `add_device_of_group()` will only be called with the same `libinput_device` (and no code makes reference to `libinput_device_group`).

Further, most of the class is implemented based on `mie::LibInputDevice::device()` which returns the first element of `devices`.

After:

`mie::LibInputDevice` represents a single device.

Some tests of the "multiple devices in one `mie::LibInputDevice`" code dropped

### DIFF
--- a/src/platforms/evdev/libinput_device.cpp
+++ b/src/platforms/evdev/libinput_device.cpp
@@ -108,14 +108,8 @@ auto get_axis_source(libinput_event_pointer* pointer) -> MirPointerAxisSource
 }
 
 mie::LibInputDevice::LibInputDevice(std::shared_ptr<mi::InputReport> const& report, LibInputDevicePtr dev)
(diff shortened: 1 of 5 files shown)


AlanGriffiths:
> `add_device_of_group()` will only be called with the same `libinput_device`

FWIW I don't see how that can happen

mattkae (approved):
Makes sense to me :+1:

Issue #3201, opened 2024-01-23
`clip_area` produces a clip that is opposite in the Y axis when using nouveau drivers 


mattkae:
This bug has been driving me a little wacky in https://github.com/mattkae/miracle.

## What's wrong
On all other platforms, if I set the `clip_area` of a window to a rectangle like `(x: 0, y: 0, w: 100, h :100)`, the absolute X, Y coordinate appears to be the top left of the screen. 

However with nouveau drivers, the absolute X, Y coordinate resolved to the bottom left of the screen.

For my tiling window manager, this makes windows "disappear" on my Nvidia setup, as their clip does _not_ correspond to the intended window position.

PR #3204, opened 2024-01-25
(#3201) When the layout of the texture disagrees with the layout of the output surface, then the clip area needs to be modified

mattkae:
When the layout of the texture disagrees with the layout of the output surface, then the clip area needs to be modified

Fixes: #3201

### DIFF
--- a/src/renderers/gl/renderer.cpp
+++ b/src/renderers/gl/renderer.cpp
@@ -377,23 +377,38 @@ auto mrg::Renderer::render(mg::RenderableList const& renderables) const -> std::
 
 void mrg::Renderer::draw(mg::Renderable const& renderable) const
 {
+    auto const texture = gl_interface->as_texture(renderable.buffer());
     auto const clip_area = renderable.clip_area();
     if (clip_area)
     {
         glEnable(GL_SCISSOR_TEST);
-        glScissor(
-            clip_area.value().top_left.x.as_int() -
-                viewport.top_left.x.as_int(),
+
+        // NOTE: We are handling the following two cases here:
+        // 1. The layout of the output surface matches that of the texture.
+        //    The clip_area provided has the origin in the top-left corner of the screen, while OpenGL
+        //    has the origin in the bottom-left corner of the screen. Hence, when the layout of the
+        //    output surface matches that of the texture, we need to invert the provided clip_area.
+        // 2. The layout of the output surface DOES NOT match that of the texture.
+        //    Note that the call to glViewport in update_gl_viewport will modify the behavior of glScissor.
+        //    Hence, if the layouts do NOT match, then we do NOT need to invert the clip area. This
+        //    is because we WANT the inverted layout which clip_area provides to us by default.
+        auto const invert_clip_area = (output_surface->layout() == graphics::gl::OutputSurface::Layout::TopRowFirst) ==
+            (texture->layout() == mg::gl::Texture::Layout::TopRowFirst);
+        auto const scissor_y = invert_clip_area ?
             viewport.top_left.y.as_int() +
                 viewport.size.height.as_int() -
                 clip_area.value().top_left.y.as_int() -
-                clip_area.value().size.height.as_int(),
-            clip_area.value().size.width.as_int(), 
+                clip_area.value().size.height.as_int() :
+            clip_area.value().top_left.y.as_int();
+
+        glScissor(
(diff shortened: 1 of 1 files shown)


src/renderers/gl/renderer.cpp
```diff
@@ -377,24 +377,35 @@ auto mrg::Renderer::render(mg::RenderableList const& renderables) const -> std::
...
     {
         glEnable(GL_SCISSOR_TEST);
-        glScissor(
-            clip_area.value().top_left.x.as_int() -
-                viewport.top_left.x.as_int(),
-            viewport.top_left.y.as_int() +
+
+        int y;
+        if (output_surface->layout() == graphics::gl::OutputSurface::Layout::TopRowFirst
+            && texture->layout() != mg::gl::Texture::Layout::TopRowFirst)
```
AlanGriffiths, line 388: Is this actually the correct test? I know it covers the case you've hit but wouldn't this be more general?
```suggestion
        if ((output_surface->layout() == graphics::gl::OutputSurface::Layout::TopRowFirst) !=
            (texture->layout() == mg::gl::Texture::Layout::TopRowFirst))
```

src/renderers/gl/renderer.cpp
```diff
@@ -377,24 +377,35 @@ auto mrg::Renderer::render(mg::RenderableList const& renderables) const -> std::
...
     auto const clip_area = renderable.clip_area();
     if (clip_area)
     {
         glEnable(GL_SCISSOR_TEST);
-        glScissor(
-            clip_area.value().top_left.x.as_int() -
-                viewport.top_left.x.as_int(),
-            viewport.top_left.y.as_int() +
+
+        int y;
```
AlanGriffiths, line 386: It would be nice to initialise y directly.
```suggestion
        auto const invert_clip_area = ....;
        auto const scissor_y = invert_clip_area ?
            (viewport.top_left.y + viewport.size.height).as_int() - (clip_area.value().top_left.y - clip_area.value().size.height).as_int() :
            clip_area.value().top_left.y.as_int();
```

AlanGriffiths (approved):
Looks sensible, just one question and a style suggestion

mattkae:
@AlanGriffiths I believe you are right. I fixed it so that the following is true:

| Output  Surface | Texture | Inverted |
| ----------- | ----------- |----------- |
| Top      | Top       | Y |
| Bottom   | Top        | N |
| Top   | Bottom        | N |
| Bottom   | Bottom        | Y |

src/renderers/gl/renderer.cpp
```diff
@@ -377,24 +377,30 @@ auto mrg::Renderer::render(mg::RenderableList const& renderables) const -> std::
...
     auto const clip_area = renderable.clip_area();
     if (clip_area)
     {
         glEnable(GL_SCISSOR_TEST);
-        glScissor(
-            clip_area.value().top_left.x.as_int() -
-                viewport.top_left.x.as_int(),
+
+        auto const invert_clip_area = (output_surface->layout() == graphics::gl::OutputSurface::Layout::TopRowFirst) ==
+            (texture->layout() == mg::gl::Texture::Layout::TopRowFirst);
```
AlanGriffiths, line 387: Is this correct? I expected:
```suggestion
        auto const invert_clip_area = (output_surface->layout() == graphics::gl::OutputSurface::Layout::TopRowFirst) !=
            (texture->layout() == mg::gl::Texture::Layout::TopRowFirst);
```

AlanGriffiths: That is definitely confusing when the PR is titled "fixing clip area when an output is inverted but the surface is not" and the comment is "When the layout of the texture disagrees with the layout of the output surface, then the clip area needs to be modified".

Maybe clarify that text and add a comment explaining this condition?

mattkae: Done! Let me know if it sounds good enough

AlanGriffiths: I see no changes to the PR text, just a comment explaining the condition.

mattkae: Updated :+1:

mattkae: I will fix up the commits too

mattkae: Done

AlanGriffiths (approved):

Issue #3198, opened 2024-01-22
Cursor from GDM Login remains on screen after logging into Miriway


mattkae:
## Repro
- In GDM, put in your login credentials
- Move the cursor out of the text input if it is in there such that it is the default pointer
- Login
- Note that the cursor remains as a ghost on the desktop _forever_

## Environment
- Nvidia with Nouveau Drivers

Saviq:
Wonder if related to #3206

AlanGriffiths:
> Wonder if related to #3206

I don't see this with Miriway `edge/mir-pr3206`

AlanGriffiths:
> (Now I have to work out why I hadn't noticed this before...)

Ah, so the the cursor isn't there _forever_ on Intel. It disappears after around 10s

Issue #3205, opened 2024-01-26
The mouse cursor is a lot slower on 2.16.2 compared to 2.15.0


AlanGriffiths:
This has been seen by @freefos on an ARM Cortex-A55 based SOC (Rockchip RK3568B2) and might be restricted to ARM systems.

> The mouse cursor is a lot slower on 123-mir2.16.2 compared to 121-mir2.15.0.
> 
> On 121-mir2.15.0 the mouse cursor moves very smooth.
> On 123-mir2.16.2 the mouse cursor moves very slowly and stutters.

[Original report](https://github.com/MirServer/mir/issues/3185#issuecomment-1912080208)

Saviq:
@RAOF would you be able to look at this in your am?

RAOF:
Hm. So on careful inspection the only thing that stands out between the two logs is the framebuffer config: 2.15 uses an RGBA 8880 config, 2.16 uses an RGBA 8888 config.

I suspect that the stuttery mouse is an artifact of a general performance difference - @freefos, is that correct?

#3206 re-adds hardware cursor support, which will likely make the *cursor* more responsive.

I think I need to add some extra information to our logging in order to track down what is actually happening here. We don't have enough information about how the framebuffer is being allocated to say anything more concrete.

Saviq:
@freefos could you please try `snap refresh ubuntu-frame --channel edge/mir-pr3206`? It has the hardware accelerated cursor again.

freefos:
@Saviq I've tested the new version.
The performance is the same and there is no cursor anymore.

There seems to be an issue with setting the cursor.
A reboot didn't help.
```
2024-01-29T14:11:00+01:00 ubuntu-frame.daemon[14180]: [2024-01-29 14:11:00.697377] < -warning- > gbm-kms: set_cursor: drmModeSetCursor failed (No such device or address)
2024-01-29T14:11:00+01:00 ubuntu-frame.daemon[14180]: [2024-01-29 14:11:00.705170] < -warning- > gbm-kms: move_cursor: drmModeMoveCursor failed (Bad address)
```
[log_129-mir2.16.2+dev68.txt](https://github.com/MirServer/mir/files/14084703/log_129-mir2.16.2%2Bdev68.txt)

freefos:
@Saviq 
Currently just by scrolling on a webpage. 
So not very scientific...

How do you guys test this?

Saviq:
[...]

PR #3206, opened 2024-01-29
graphics/MultiplexingDisplay: Add HW cursor support.

RAOF:
Create a HW cursor on each `Display`; if any `Display` cannot support a HW cursor, bail.

Our existing HW cursor implementations already handle setting the position outside of their display area, so we can just directly delegate to all multiplexed `Cursor` implementations without needing to track which platform(s) should actually be handling the event.

Fixes: #3198

### DIFF
--- a/src/server/graphics/multiplexing_display.cpp
+++ b/src/server/graphics/multiplexing_display.cpp
@@ -15,11 +15,14 @@
  */
 
 #include "multiplexing_display.h"
+#include "multiplexing_hw_cursor.h"
 #include "mir/graphics/display_configuration.h"
 #include "mir/renderer/gl/context.h"
 #include "mir/output_type_names.h"
+#include "mir/log.h"
 
 #include <boost/throw_exception.hpp>
+#include <exception>
 #include <stdexcept>
 #include <sstream>
 #include <functional>
@@ -319,5 +322,21 @@ void mg::MultiplexingDisplay::resume()
 
 auto mg::MultiplexingDisplay::create_hardware_cursor() -> std::shared_ptr<Cursor>
 {
-    return {};
+    std::vector<Display*> platform_displays;
+    for (auto& display : displays)
+    {
+        platform_displays.push_back(display.get());
+    }
+    try
+    {
+        return std::make_shared<MultiplexingCursor>(platform_displays);
+    }
+    catch (std::exception const&)
+    {
+        mir::log(
+            mir::logging::Severity::informational,
+            "display",
+            "Failed to create hardware cursor");
+        return nullptr;
+    }
 }
--- a/src/server/graphics/multiplexing_hw_cursor.cpp
+++ b/src/server/graphics/multiplexing_hw_cursor.cpp
@@ -1,0 +1,68 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "multiplexing_hw_cursor.h"
+
+#include "mir/graphics/display.h"
+#include <boost/throw_exception.hpp>
+
+namespace mg = mir::graphics;
+
+namespace
+{
+auto construct_platform_cursors(std::span<mg::Display*> platform_displays) -> std::vector<std::shared_ptr<mg::Cursor>>
+{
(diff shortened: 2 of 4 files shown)


src/server/graphics/multiplexing_hw_cursor.cpp
```diff
@@ -0,0 +1,68 @@
...
+auto construct_platform_cursors(std::span<mg::Display*> platform_displays) -> std::vector<std::shared_ptr<mg::Cursor>>
+{
+    std::vector<std::shared_ptr<mg::Cursor>> cursors;
+    for (auto display : platform_displays)
+    {
+        cursors.push_back(display->create_hardware_cursor());
+        if (cursors.back() == nullptr)
+        {
+            BOOST_THROW_EXCEPTION((std::runtime_error{"Platform failed to create hardware cursor"}));
+        }
```
AlanGriffiths, line 35: Probably not a problem here, but thinking forward: this means that when adding a virtual display we offer different behaviour. (Obviously, fixing that would need something more complex than "hardware cursor" xor "software cursor")

src/server/graphics/multiplexing_display.cpp
```diff
@@ -319,5 +322,21 @@ void mg::MultiplexingDisplay::resume()
...
+    {
+        return std::make_shared<MultiplexingCursor>(platform_displays);
+    }
+    catch (std::exception const&)
+    {
+        mir::log(
+            mir::logging::Severity::informational,
+            "display",
+            "Failed to create hardware cursor");
+        return nullptr;
```
AlanGriffiths, line 340: Not sure why, but I prefer default constructing a `shared_ptr<...>` (original code) to implicitly constructing it with `nullptr`.

src/server/graphics/multiplexing_hw_cursor.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+namespace
+{
+auto construct_platform_cursors(std::span<mg::Display*> platform_displays) -> std::vector<std::shared_ptr<mg::Cursor>>
+{
+    std::vector<std::shared_ptr<mg::Cursor>> cursors;
+    for (auto display : platform_displays)
+    {
+        cursors.push_back(display->create_hardware_cursor());
+        if (cursors.back() == nullptr)
```
AlanGriffiths, line 32: `operator!()` seems more idiomatic `operator==(T*)`

AlanGriffiths (approved):
A couple of thoughts on style, but I haven't come up with an argument for change.

Approving, but won't queue yet (in case someone wants to chime in)

Saviq:
Something's not right here:

https://github.com/MirServer/mir/issues/3205#issuecomment-1914686561

RAOF:
> > Something's not right here:
> > [#3205 (comment)](https://github.com/MirServer/mir/issues/3205#issuecomment-1914686561)
> 
> @RAOF if I understand your theory in that issue correctly, then PR is "merely" exposing an underlying bug, not itself causing problems?

My theory is that the HW cursor (specifically, `drmMode*Cursor`) is broken on that device/driver, and this PR causes Mir to *use* the HW cursor API, so the cursor is broken.

I *think* the last time we tried to use this API was in 2.13, so testing a 2.13-based frame and checking in the logs that it enables the hardware cursor would be a way to verify this.

AlanGriffiths:
> My theory is that the HW cursor (specifically, `drmMode*Cursor`) is broken on that device/driver

That brings back memories...
```bash
  # Workaround for https://github.com/MirServer/mir-kiosk/issues/23
  # "Mouse offset from image on Raspberry Pi 3 with Ubuntu Core 18"
  if grep -q snap_core=core18 /proc/cmdline && grep -q snap_kernel=pi-kernel /proc/cmdline
  then cursor=software
  fi
```
(I know it's a different bug and doesn't help here)

AlanGriffiths:
> Something's not right here:

@Saviq that seems to have been resolved. We should land unless you have thoughts on my "nits"?

Saviq:
> @Saviq that seems to have been resolved. We should land unless you have thoughts on my "nits"?

I think we need confirmation on what goes on with the hw cursor on a Pi - did it ever work? I'll have a look with 2.13 as Chris suggests.

Saviq:
> My theory is that the HW cursor (specifically, `drmMode*Cursor`) is broken on that device/driver, and this PR causes Mir to _use_ the HW cursor API, so the cursor is broken.
> 
> I _think_ the last time we tried to use this API was in 2.13, so testing a 2.13-based frame and checking in the logs that it enables the hardware cursor would be a way to verify this.

Like this?
```
<information> mirserver: Using hardware cursor
# …
<information> mirserver: Mir version 2.13.1
# …
<information> GLRenderer: GL vendor: Broadcom
<information> GLRenderer: GL renderer: V3D 4.2
```

So yeah, _should_ work?

Saviq (commented):
Ah, but with `22/edge` I get:

```
<information> mirserver: Using software cursor
```

And on this PR:
```
<information> mirserver: Using hardware cursor
```

Still, we should give @RAOF the chance to reply to your nits. ACK otherwise.

Saviq (approved):

AlanGriffiths:
> Still, we should give @RAOF the chance to reply to your nits. ACK otherwise.

Three days is enough chance

PR #3210, opened 2024-01-30
debian: package input-stub

Saviq:
It may be useful to disable input altogether with this.


AlanGriffiths:
When would this be a better solution than `--enable-input`?

Saviq:
It wouldn't.

PR #3211, opened 2024-01-31
feature: abstract away application selector behavior  in preparation for alt-tick behavior

mattkae:
## What's new?
- Abstracted away the window cycling algorithm so that we can instead its use to be not only between applications but also _within_ applications

A follow up PR will integrate the window selection within apps.

### DIFF
--- a/src/miral/application_selector.cpp
+++ b/src/miral/application_selector.cpp
@@ -27,22 +27,22 @@ struct WeakApplicationSelector
 {
     using value_t = std::weak_ptr<mir::scene::Session>;
 
-    WeakApplicationSelector(WindowManagerTools tools)
+    explicit WeakApplicationSelector(WindowManagerTools const& tools)
         : tools{tools}
     {}
 
-    auto equals(value_t const &first, value_t const &second) -> bool
+    static auto equals(value_t const& first, value_t const& second) -> bool
     {
         return !first.owner_before(second) && !second.owner_before(first);;
     }
 
-    auto selectable(value_t const &session) -> bool
+    auto selectable(value_t const& session) const -> bool
     {
         return tools.window_to_select_application(session.lock()) != std::nullopt;
     }
 
     void
-    on_selected(value_t const &previously_selected, value_t const &newly_selected, value_t const &originally_selected)
+    on_selected(value_t const& previously_selected, value_t const& newly_selected, value_t const& originally_selected)
     {
         auto previous_session = previously_selected.lock();
         auto new_session = newly_selected.lock();
@@ -58,7 +58,7 @@ struct WeakApplicationSelector
             // Edge case: if we have gone full circle around the list back to the original app
             // then we will wind up in a situation where the original app - now in the second z-order
             // position - will be swapped with the final app, putting the final app in the second position.
-            for (auto const &window: tools.info_for(new_session).windows())
+            for (auto const& window: tools.info_for(new_session).windows())
                 tools.send_tree_to_back(window);
         } else
         {
@@ -68,7 +68,7 @@ struct WeakApplicationSelector
         tools.select_active_window(new_window.value());
     }
 
-    void on_cancelled(value_t const &session)
+    void on_cancelled(value_t const& session)
     {
         auto window_to_select = tools.window_to_select_application(session.lock());
         if (window_to_select != std::nullopt)
@@ -77,7 +77,7 @@ struct WeakApplicationSelector
         }
     }
 
-    void reset(value_t &session)
+    static void reset(value_t& session)
     {
         session.reset();
     }
@@ -90,21 +90,18 @@ class TabOrderSelector : public SelectorBase
 {
 public:
     using T = typename SelectorBase::value_t;
-
-    template<typename ... Args>
-    TabOrderSelector(Args &&... args): SelectorBase(std::forward<Args>(args) ...)
-    {}
+    using SelectorBase::SelectorBase;
 
     TabOrderSelector(TabOrderSelector const &) = delete;
 
     auto operator=(TabOrderSelector const &) -> TabOrderSelector & = delete;
 
-    void add(T const &t)
+    void add(T const& t)
     {
         focus_list.push_back(t);
     }
 
-    void remove(T const &t)
+    void remove(T const& t)
     {
         auto it = find(t);
         if (it == focus_list.end())
@@ -136,7 +133,7 @@ public:
         focus_list.erase(it);
     }
 
-    void focus(T const &t)
+    void focus(T const& t)
     {
         auto it = find(t);
         if (it == focus_list.end())
@@ -155,7 +152,7 @@ public:
         selected = t;
     }
 
-    void unfocus(T const &t)
+    void unfocus(T const& t)
     {
         if (SelectorBase::equals(t, selected))
         {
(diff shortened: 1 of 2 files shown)


src/miral/application_selector.cpp
```diff
@@ -21,243 +21,439 @@
...
 using namespace miral;
 
-ApplicationSelector::ApplicationSelector(miral::WindowManagerTools const& in_tools) :
-    tools{in_tools}
-{}
+namespace miral
+{
+
+template <typename T>
+class WindowCyclingBehavior
```
mattkae, line 28: I don't particularly like this name, if you can think of something better

src/miral/application_selector.cpp
```diff
@@ -21,243 +21,439 @@
...
+{
+public:
+    WindowCyclingBehavior(
+        std::function<bool(T const&, T const&)> comparison_func,
+        std::function<bool(T const&)> can_select,
+        std::function<void(T const&, T const&, T const&)> on_raised,
+        std::function<void(T const&)> on_cancelled,
+        std::function<void(T&)> reset
```
AlanGriffiths, line 36: OK, so this is what you were asking about? Declaring these as virtual functions vs providing them as functors?

I think what you actually want here is a traits class:

```c++
template<typename T> WindowCyclingTraits {};

struct WindowCyclingTraits<WeakApplication>
{
    static auto equals(WeakApplication const&, WeakApplication const&) => bool;
    static auto selectable(WeakApplication const&) => bool;
    ...
};
```

Then `WindowCyclingBehavior` can `using Traits = WindowCyclingTraits<T>` and `Traits::equals(...)`.

src/miral/application_selector.cpp
mattkae: The class handles  ordering a stack of `T` items such that they are always in the correct Alt + Tab order that we want

src/miral/application_selector.cpp
mattkae: That makes sense to me, but do the traits need to be `static`? The problem being that functions like `can_select` depend on `WindowManagerTools`, which is passed via a constructor. Would a non-static class defeat the purpose of the traits pattern?

AlanGriffiths: TPM traits are normally types or static functions. And if you're passing in `WindowManagerTools` to construct a "trait" instance, you might just as well pass it to the static method.

OTOH another design option is to template on a base class (which provides these functions), not the stored type.

src/miral/application_selector.cpp
```diff
@@ -21,243 +21,439 @@
...
+    ) : comparison_func{comparison_func},
+        can_select{can_select},
+        on_raised{on_raised},
+        on_cancelled{on_cancelled},
+        reset{reset}
+    {
+    }
```
AlanGriffiths, line 43:
```suggestion
struct ApplicationSelector::WeakApplicationSelector
{
    using value_t = std::weak_ptr<mir::scene::Session>;

    WeakApplicationSelector(ApplicationSelector* selector, WindowManagerTools tools);
    auto equals(value_t const&, value_t const&) -> bool;
    auto selectable(value_t const&) -> bool;
    void on_selected(value_t const& previously_selected, value_t const& newly_selected, value_t const& originally_selected);
    void on_cancelled(value_t const&);
    void reset(value_t&);

    WindowManagerTools tools;
};

template <typename SelectorBase>
class TabOrderSelector : SelectorBase
{
public:

    using SelectorBase::SelectorBase;
    using T = typename SelectorBase::value_t;
```

src/miral/application_selector.cpp
```diff
@@ -21,243 +21,439 @@
...
+        get_can_select_func(tools),
+        get_on_raised_func(tools),
+        get_on_cancelled_func(tools),
+        get_reset_func(),
+        other.window_cycler->get_focus_list(),
+        other.window_cycler->get_originally_focused(),
+        other.window_cycler->get_focused(),
+        other.window_cycler->get_is_active()
+    );
 }
```
AlanGriffiths, line 295: 1. Do we actually copy `ApplicationSelector` instances?
2. `window_cycler` should be initialised, not initialised and then assigned
3. It looks like `WindowCyclingBehavior` needs a copy constructor, not `getters` for all its state

src/miral/application_selector.cpp
```diff
@@ -21,243 +21,439 @@
...
+        get_on_raised_func(tools),
+        get_on_cancelled_func(tools),
+        get_reset_func(),
+        other.window_cycler->get_focus_list(),
+        other.window_cycler->get_originally_focused(),
+        other.window_cycler->get_focused(),
+        other.window_cycler->get_is_active()
+    );
     return *this;
 }
```
AlanGriffiths, line 312: I guess this is partly pre-existing, but the canonical, exception safe, way to write assignment is:

```suggestion
auto ApplicationSelector::operator=(ApplicationSelector const other) -> ApplicationSelector&
{
    std::swap(tools, other.tools);
    std::swap(window_cycler, other.window_cycler);
    return *this;
}
```

src/miral/application_selector.cpp
```diff
@@ -21,243 +21,439 @@
...
-    focus_list.push_back(application);
+    window_cycler->add(application);
 }
 
 void ApplicationSelector::advise_focus_gained(WindowInfo const& window_info)
 {
     auto window = window_info.window();
     if (!window)
     {
+        mir::log_error("ApplicationSelector::advise_focus_gained: Could not find the window");
```
AlanGriffiths, line 324: Why is this an _error_? A WindowInfo with no window sounds like invariants have been broken. In which case, should we really be ignoring the problem and carrying on?

We ought to either:

1. Assume the invariant holds; or,
2. Report the program is in a bad state: `fatal_error()`

src/miral/application_selector.cpp
AlanGriffiths: `TabOrderSelector`?

AlanGriffiths (changes requested):

src/miral/application_selector.cpp
AlanGriffiths: Also, things line `equals`/`comparison_func` could just be found by overloading.

src/miral/application_selector.cpp
mattkae: I believe `MinimalWindowManager` doesn't have copies deleted, so I needed to define this. But do we actually copy the `MinimalWindowManager`...?

AlanGriffiths (changes requested):

src/miral/application_selector.cpp
```diff
@@ -21,243 +21,360 @@
...
+class TabOrderSelector : public SelectorBase
 {
-    auto window = window_info.window();
-    if (!window)
+public:
+    using T = typename SelectorBase::value_t;
+
+    template<typename ... Args>
+    TabOrderSelector(Args &&... args): SelectorBase(std::forward<Args>(args) ...)
+    {}
```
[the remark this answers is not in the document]
mattkae: Oh neat. I didn't know about this

mattkae:
Well, I just realized that this could be much easier :sweat_smile: A better implementation is to come!

mattkae:
This was unnecessary

PR #3212, opened 2024-02-01
Log platform selection

RAOF:
Some logging, so I can get a snap on the Pi.

### DIFF
--- a/src/platforms/common/server/owning_c_str.h
+++ b/src/platforms/common/server/owning_c_str.h
@@ -14,25 +14,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#ifndef MIR_OWNING_C_STR_
+#define MIR_OWNING_C_STR_
 #include <cstdlib>
 #include <memory>
-
-namespace
-{
-class CStringFree
-{
-public:
-    void operator()(char* str)
-    {
-        if(str)
-        {
-            free(str);
-        }
-    }
-};
-}
-
 namespace mir
 {
-using CStr = std::unique_ptr<char[], CStringFree>;
+using CStr = std::unique_ptr<char[], decltype([](char const* p) { ::free(const_cast<char *>(p)); })>;
 }
+#endif
--- a/src/platforms/common/server/owning_c_str.h
+++ b/src/platforms/common/server/owning_c_str.h
@@ -1,0 +1,38 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <cstdlib>
+#include <memory>
+
+namespace
+{
+class CStringFree
+{
+public:
+    void operator()(char* str)
+    {
+        if(str)
+        {
+            free(str);
+        }
+    }
+};
+}
+
+namespace mir
+{
+using CStr = std::unique_ptr<char[], CStringFree>;
+}
--- a/src/platforms/gbm-kms/server/kms/display_buffer.cpp
+++ b/src/platforms/gbm-kms/server/kms/display_buffer.cpp
@@ -43,6 +43,7 @@
 #include <stdexcept>
 #include <chrono>
 #include <algorithm>
+#include <xf86drm.h>
 
 namespace mg = mir::graphics;
 namespace mgg = mir::graphics::gbm;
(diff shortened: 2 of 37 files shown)


src/platforms/common/server/owning_c_str.h
```diff
@@ -0,0 +1,38 @@
...
+            free(str);
+        }
+    }
+};
+}
+
+namespace mir
+{
+using CStr = std::unique_ptr<char[], CStringFree>;
+}
```
AlanGriffiths, line 38: That looks like a recipe for an ODR violation - thanks to the anon namespace `CStringFree` is a different type in each translation unit and that makes the `mir::CStr` definition different in each TU.

AlanGriffiths, line 38: 1. what? No header guards?
2. The null check is redundant
3. The `operator()` can be `const`
4. This doesn't handle `const` c_strings

```suggestion
#ifndef MIR_OWNING_C_STR_
#define MIR_OWNING_C_STR_

#include <cstdlib>
#include <memory>

namespace mir
{
using CStr = std::unique_ptr<char[], decltype([](char const* p) { ::free(const_cast<char *>(p)); })>;
}

#endif
```

AlanGriffiths (commented):
There's also a legacy `CStrFree` hidden in existing code

src/server/graphics/default_configuration.cpp
```diff
@@ -370,30 +369,6 @@ auto mir::DefaultServerConfiguration::the_rendering_platforms() ->
...
-                egl_generic = rp.second;
-            }
-        }
-
-        // If we skipped egl-generic, add it to the end
-        if (egl_generic)
-        {
-            rendering_platforms.push_back(egl_generic);
-        }
-
```
AlanGriffiths, line 396: I know this code is ugly, but why are we dropping it?

The commit comment "FIXME: Ooops! Don't throw away most of our hardware platforms!" doesn't reflect the change. (It just makes the order non-deterministic.)

src/platforms/gbm-kms/server/kms/display_buffer.cpp
```diff
@@ -311,6 +312,13 @@ auto mgg::DisplaySink::gbm_device() const -> std::shared_ptr<struct gbm_device>
     return gbm;
 }
 
+auto mgg::DisplaySink::describe_output() const -> std::string
+{
+    std::unique_ptr<char[], void(*)(char*)> drm_node{drmGetDeviceNameFromFd2(drm_fd()), [](char* to_free) { free(to_free); }};
```
AlanGriffiths, line 317: Another incantation for `CStr`

Saviq:
Can confirm this brings Pi's performance back to what's expected.

@freefos this is work-in-progress, but could you confirm with `snap refresh ubuntu-frame --channel 22/edge/mir-pr3212` please?

AlanGriffiths:
> Can confirm this brings Pi's performance back to what's expected.

I'm struggling to know what changed to make that happen.

Apart from logging, the change to `src/server/graphics/default_configuration.cpp` looks like the only thing. But that avoids sorting `rendering_platforms` to "mostly alphabetical, with mir:egl-generic at the end" and leaves it in "order encountered". That probably just makes the first viable platform non-deterministic, which isn't necessarily a good thing.

Saviq:
> I'm struggling to know what changed to make that happen.

Here's @RAOF's explanation from Matrix:

> In the “load egl-generic last” patch we made it so that we only keep the last device's platform for any given platform type.
So all the “make sure to match the renderer to the display when possible” code is selecting from a set that doesn't include the renderer that matches the display!

Yes, it needs a better solution (and bringing back the determinism), but this PR, so far, just confirmed the issue.

With both 2.15 and this PR, MotionMark in WPE detects 60fps with Frame at ~15% CPU, compared to 2.16.2 detecting 15fps at 50-60% CPU usage of Frame.

The logging from this branch provides:
> Platform selection: Output gbm-kms (/dev/dri/card0) - KMS connector id: 89: Selected renderer: GBM surface on
 /dev/dri/card0

AlanGriffiths:
> > In the “load egl-generic last” patch we made it so that we only keep the last device's platform for any given platform type.

Oh! We load multiple instances of each platform? I hadn't understood that. Easy to fix. (Incoming...)

src/platforms/common/server/owning_c_str.h
```diff
@@ -0,0 +1,25 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_OWNING_C_STR_
+#define MIR_OWNING_C_STR_
+#include <cstdlib>
+#include <memory>
+namespace mir
+{
+using CStr = std::unique_ptr<char[], decltype([](char const* p) { ::free(const_cast<char *>(p)); })>;
```
RAOF, line 23: Huh. You learn something new every day. `decltype(<lambda literal>)` is a neat way of getting a type with a specific `operator()`!

freefos:
@Saviq With this release wpe-webkit-mir-kiosk restarts once in a while.
[log_129-mir2.16.2+dev81_wpe-kiosk.txt](https://github.com/MirServer/mir/files/14136634/log_129-mir2.16.2%2Bdev81_wpe-kiosk.txt)

freefos:
@Saviq Its again the issue where the memory usage increases until it crashes.
As described here: [https://github.com/MirServer/ubuntu-frame/issues/159](https://github.com/MirServer/ubuntu-frame/issues/159)
This hapens only when there is a vnc connection open.
But now it wasn't our web application but the motionmark website that is active.
image

Saviq:
Thanks @freefos. #3213 is what we're going to backport to 2.16 soon.

Saviq:
@RAOF should we close this? Revive?

RAOF:
Close this for now.

PR #3213, opened 2024-02-02
platform/graphics: Allow for multiple RenderingPlatforms of each type 

RAOF:
We load one `RenderingPlatform` for each hardware device detected, and then select the best one to drive each display. If we don't keep all the `RenderingPlatform`s then we might not keep the one capable of direct-output, requiring a CPU copy to the display.

Fixes: #3205

### DIFF
--- a/src/server/graphics/default_configuration.cpp
+++ b/src/server/graphics/default_configuration.cpp
@@ -41,7 +41,6 @@
 
 #include <boost/throw_exception.hpp>
 
-#include <ranges>
 #include <sstream>
 
 namespace mg = mir::graphics;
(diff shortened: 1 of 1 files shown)


RAOF:
This is Alan's fix pulled out of #3212. Although we should land more logging to help diagnose problems in future, we can land this quickly and cleanly.

AlanGriffiths (approved):
Obviously, I'm OK with this and, by implication, @RAOF is too. So we're probably good to land. @Saviq WDYT?

Saviq:
Let me just confirm on the Pi and will merge, indeed.

Saviq (approved):
Yes. Can confirm this helps on a Pi.

PR #3216, opened 2024-02-05
feature: alt + grave behavior now implemented alongside alt + tick

mattkae:
## What's new?
- I realized that the last PR was going to be messy, as we would have to maintain a tab order for every application _and_ every window.
- However, I realized that this list was one in the same. Basically, we just keep the order of Windows, and we choose whether or not we are jumping from App to App or from Window to Window within an app
- So much simpler, less data to keep track of, and easier to read :magic_wand:

### DIFF
--- a/debian/libmiral7.symbols
+++ b/debian/libmiral7.symbols
@@ -79,11 +79,11 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::MinimalWindowManager::MinimalWindowManager(miral::WindowManagerTools const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::MinimalWindowManager(miral::WindowManagerTools const&, MirInputEventModifier)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::advise_delete_app(miral::ApplicationInfo const&)@MIRAL_5.0" 5.0.0
- (c++)"miral::MinimalWindowManager::advise_delete_window(miral::WindowInfo const&)@MIRAL_5.1" 5.0.0
+ (c++)"miral::MinimalWindowManager::advise_delete_window(miral::WindowInfo const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::advise_focus_gained(miral::WindowInfo const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::advise_focus_lost(miral::WindowInfo const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::advise_new_app(miral::ApplicationInfo&)@MIRAL_5.0" 5.0.0
- (c++)"miral::MinimalWindowManager::advise_new_window(miral::WindowInfo const&)@MIRAL_5.1" 5.0.0
+ (c++)"miral::MinimalWindowManager::advise_new_window(miral::WindowInfo const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::begin_pointer_move(miral::WindowInfo const&, MirInputEvent const*)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::begin_pointer_resize(miral::WindowInfo const&, MirInputEvent const*, MirResizeEdge const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::begin_touch_move(miral::WindowInfo const&, MirInputEvent const*)@MIRAL_5.0" 5.0.0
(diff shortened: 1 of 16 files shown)


include/miral/miral/minimal_window_manager.h
```diff
@@ -82,6 +82,10 @@ class MinimalWindowManager : public WindowManagementPolicy
 
     void advise_delete_app(miral::ApplicationInfo const& app_info) override;
 
+    void advise_new_window(WindowInfo const& app_info) override;
+
+    void advise_delete_window(WindowInfo const& app_info) override;
+
```
AlanGriffiths, line 88: We should document when these are introduced. 
```suggestion
    /// \remark Since MirAL 5.0
    void advise_new_window(WindowInfo const& app_info) override;

    /// \remark Since MirAL 5.0
    void advise_delete_window(WindowInfo const& app_info) override;

```

debian/libmiral7.symbols
```diff
@@ -79,9 +79,11 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::MinimalWindowManager::MinimalWindowManager(miral::WindowManagerTools const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::MinimalWindowManager(miral::WindowManagerTools const&, MirInputEventModifier)@MIRAL_5.0" 5.0.0
  (c++)"miral::MinimalWindowManager::advise_delete_app(miral::ApplicationInfo const&)@MIRAL_5.0" 5.0.0
+ (c++)"miral::MinimalWindowManager::advise_delete_window(miral::WindowInfo const&)@MIRAL_5.1" 5.0.0
```
AlanGriffiths, line 82: These symbols are Miral 5.0, not 5.1

include/miral/miral/window_manager_tools.h
```diff
@@ -161,6 +161,9 @@ class WindowManagerTools
     /// \remark Since MirAL 3.10
     auto window_to_select_application(const Application) const -> std::optional<Window>;
 
+    /// Check if the provided window can be selected
+    auto can_select_window(Window const&) const -> bool;
```
AlanGriffiths, line 165:
```suggestion
    /// Check if the provided window can be selected
    /// \remark Since MirAL 5.0
    auto can_select_window(Window const&) const -> bool;
```

src/miral/application_selector.cpp
```diff
@@ -76,42 +69,34 @@ void ApplicationSelector::advise_focus_gained(WindowInfo const& window_info)
...
+    auto it = find(window_info.window());
     if (it == focus_list.end())
     {
         mir::log_warning("ApplicationSelector::advise_delete_app could not delete the app.");
         return;
     }
 
     // If we delete the selected application, we will try to select the next available one.
-    if (application == selected.lock())
+    if (*it == selected)
```
AlanGriffiths, line 97: The code no longer agrees wit hthe comment

src/miral/symbols.map
```diff
@@ -470,3 +470,14 @@ global:
...
 local: *;
 };
+
+MIRAL_5.1 {
+global:
+  extern "C++" {
+    miral::MinimalWindowManager::advise_new_window*;
+    non-virtual?thunk?to?miral::MinimalWindowManager::advise_new_window*;
+    miral::MinimalWindowManager::advise_delete_window*;
+    non-virtual?thunk?to?miral::MinimalWindowManager::advise_delete_window*;
```
AlanGriffiths, line 480: We don't need a new stanza - we haven't released Miral 5.0 yet. (Just run the regenerate script)

AlanGriffiths (changes requested):
I've not yet reviewed the functionality, but there's a bunch of "housekeeping" that can be cleared away already.

AlanGriffiths (changes requested):
This looks like a sensible implementation, and works sensibly in the scenarios I've tried.

However, there don't seem to be enough tests. I'd expect a similar number of new tests for within-app switching as we have existing tests for between-app switching. Am I missing something?

debian/libmiral7.symbols
```diff
@@ -149,6 +151,8 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::StartupInternalClient::StartupInternalClient(std::function<void (wl_display*)>, std::function<void (std::weak_ptr<mir::scene::Session>)>)@MIRAL_5.0" 5.0.0
  (c++)"miral::StartupInternalClient::operator()(mir::Server&)@MIRAL_5.0" 5.0.0
  (c++)"miral::StartupInternalClient::~StartupInternalClient()@MIRAL_5.0" 5.0.0
+ (c++)"miral::WaylandExtensions::Context::Context()@MIRAL_5.0" 5.0.0
+ (c++)"miral::WaylandExtensions::Context::~Context()@MIRAL_5.0" 5.0.0
```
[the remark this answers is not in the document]
mattkae: When I regenerated the symbols, it got generated too

AlanGriffiths (approved):

Saviq:
@mattkae had to revert this, as it made Frame fail to start up:

```plain
Mir fatal error: advise_focus_gained: window was not found in the list
[1]    36090 IOT instruction (core dumped)  ubuntu-frame
```

mattkae:
> @mattkae had to [revert this](https://github.com/canonical/mir/commit/74d847ea2f2221ad567ec3a133a03659fb6b9393), as it made Frame fail to start up:
> 
> ```
> Mir fatal error: advise_focus_gained: window was not found in the list
> !!! Fatal signal received. Attempting cleanup, but deadlock may occur
> Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGABRT
> [1]    36090 IOT instruction (core dumped)  ubuntu-frame
> ```
> 
> EDIT: `--wallpaper=false` works around, so whatever the wallpaper client does isn't handled here.
> 
> EDIT2: Ah no, starting a client without the wallpaper results in the same. So it must be Frame's special window management.

Ah that is no good! Anything I can do to help on frame's side of things? Spooky that the window was not found in the list... probably not calling a super method?

AlanGriffiths:
> Ah that is no good! Anything I can do to help on frame's side of things? Spooky that the window was not found in the list... probably not calling a super method?

Maybe you can track that down?

mattkae:
> > Ah that is no good! Anything I can do to help on frame's side of things? Spooky that the window was not found in the list... probably not calling a super method?
> 
> Maybe you can track that down?

Yup! I'll look into it now

PR #3218, opened 2024-02-06
first draft: lockscreen protocol for mir

mattkae:
# Draft 2 (Latest)
proposed

# Draft 1
This is how a logind interaction would look:
proposed

### DIFF
--- a/wayland-protocols/ext-lockscreen_v1.xml
+++ b/wayland-protocols/ext-lockscreen_v1.xml
@@ -1,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="ext-lockscreen_v1">
+  <copyright>    
+    Copyright © Canonical Ltd.
+
+    This program is free software: you can redistribute it and/or modify it
+    under the terms of the GNU General Public License version 2 or 3,
+    as published by the Free Software Foundation.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+    
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  </copyright>
+  <description summary="compositor authoritative screen-locking mechanism">
+    This protocol allows for a priveleged Wayland client to display arbitrary graphics
+    on the screen when the compositor informs the client that the session has been
+    locked.
+
+    The compositor may choose to restrict this protocol to a special client
+    launched by the compositor itself or expose it to all privileged clients.
+    This is up to the compositor policy.
+
+    The compositor is responsible for deciding when the session should be locked and unlocked.
+    If the client dies while the session is locked, the client may simply resubmit
+    a buffer for display at a later time. During this time, the compositor itself
+    may have been unlocked via some other mechanism (e.g. loginctl). Exactly how and
+    when the compositor locks and unlocks is the policy of the compositor.
+
+    A typical use case is for the compositor to lock or unlock when it receives a
+    signal from Logind to do so.
+
+    The key words "must", "must not", "required", "shall", "shall not",
+    "should", "should not", "recommended",  "may", and "optional" in this
+    document are to be interpreted as described in IETF RFC 2119.
+  </description>
+
+  <interface name="ext_lockscreen_manager_v1" version="1">
+    <description summary="used to create lockscreen surfaces">
+      
+    </description>
+    <request name="associate">
+      This request creates a lockscreen surface and asks the compositor to show this
+      surface on the provided output when the session is next locked. If the session is
+      already locked, the compositor will display this surface right away.
+
+      Surfaces may display different content on different outputs.
+      
+      This surface will be shown immediately after the session is locked. It is expected
+      that the compositor does not render intermediary content if a lockscreen surface
+      is provided for an output. If a surface is NOT provided for an output, then it is up to the
+      policy of the compositor to render what it wants.
+
+      The client is expected to destroy lockscreen surfaces as outputs are removed.
+      The client is expected to associate lockscreen surfaces as outputs are added.
+
+      Associating more than one surface for an output is a duplicate_surface error.
+      Associating a surface with a null buffer is a null_buffer error.
+      
+      <arg name="id" type="new_id" interface="ext_lockscreen_surface_v1"/>
+      <arg name="surface" type="object" interface="wl_surface"/>
+      <arg name="output" type="object" interface="wl_output"/>
+    </request>
+
+    <enum name="error">
+      <entry name="duplicate_surface" value="0"
+             summary="attempted to associate more than once surface as the lockscreen surface for an output"/>
+      <entry name="null_buffer" value="1"
+        summary="surface committed with a null buffer"/>
+    </enum>
+  </interface>
+
+  <interface name="ext_lockscreen_surface_v1" version="1">
+    <request name="destroy" type="destructor">
+      This informs the compositor that the lockscreen surface object will no longer be
+      used.
+
+      If the session is locked when this happens, the compositor will display a default
+      render of its choosing until the session is unlocked.
+    </request>
+    
+    <event name="locked">
+      This event is sent when the compositor decides to lock the session.
+      At this time, the compositor will show the surface.
+    </event>
+    <event name="unlocked">
+      This event is sent when the compositor decides to unlock the session.
+      At this time, the compositor will hide the surface.
+    </event>
+    <event name="configure">
+      This event is sent when the lockscreen surface is first created and anytime that the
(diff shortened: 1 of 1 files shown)


wayland-protocols/ext-lockscreen_v1.xml
```diff
@@ -0,0 +1,104 @@
...
+      This surface will be shown immediately after the session is locked. It is expected
+      that the compositor does not render intermediary content if a lockscreen surface
+      is provided for an output. If a surface is NOT provided for an output, then it is up to the
+      policy of the compositor to render what it wants.
+
+      The client is expected to destroy lockscreen surfaces as outputs are removed.
+      The client is expected to associate lockscreen surfaces as outputs are added.
+
+      Associating more than one surface for an output is a duplicate_surface error.
+      Associating a surface with a null buffer is a null_buffer error.
```
RAOF, line 61: What happens if a client waits until receiving the `locked` event to commit its first buffer?

RAOF (commented):
Seems reasonable. I think we could tweak things a little if we were ok with having the client submit the lockscreen buffer on-demand, but this is fine.

wayland-protocols/ext-lockscreen_v1.xml
```diff
@@ -0,0 +1,104 @@
...
+      This event is sent when the compositor decides to unlock the session.
+      At this time, the compositor will hide the surface.
+    </event>
+    <event name="configure">
+      This event is sent when the lockscreen surface is first created and anytime that the
+      dimensions of the output changes after that. The client may use the width and height
+      to update their buffer to fit the output dimensions.
+
+      The width and height are in surface-local coordinates. The client is allowed to ignore
+      the dimensions.
```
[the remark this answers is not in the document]
mattkae: Then they have a poorly sized buffer (either one that is too large or too small), and the default render is shown in the places that are not filled by the lockscreen. This _could_ be a protocol error, but it felt a little brash to do so.

wayland-protocols/ext-lockscreen_v1.xml
```diff
@@ -0,0 +1,104 @@
...
+    </request>
+    
+    <event name="locked">
+      This event is sent when the compositor decides to lock the session.
+      At this time, the compositor will show the surface.
+    </event>
+    <event name="unlocked">
+      This event is sent when the compositor decides to unlock the session.
+      At this time, the compositor will hide the surface.
+    </event>
```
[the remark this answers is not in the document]
mattkae: Yeah it's useful to show _something_ interesting when a lock first happens, like an animation.

Although maybe "unlock" is less useful, because the surface would have already been hidden at that point. Or maybe we can wait on the lockscreen client's confirmation so that it can first show an unlocking animation and _then_ unlock the session? Maybe a `confirm_unlock` request?

wayland-protocols/ext-lockscreen_v1.xml
mattkae: Good question. I am assuming a protocol error

wayland-protocols/ext-lockscreen_v1.xml
```diff
@@ -0,0 +1,104 @@
...
+  </description>
+
+  <interface name="ext_lockscreen_manager_v1" version="1">
+    <description summary="used to create lockscreen surfaces">
+      
+    </description>
+    <request name="associate">
+      This request creates a lockscreen surface and asks the compositor to show this
+      surface on the provided output when the session is next locked. If the session is
+      already locked, the compositor will display this surface right away.
```
AlanGriffiths, line 48: Just checking I've understood correctly: the client needs to track outputs being added, removed and reconfigured in order to create and associate the corresponding surfaces and buffers _regardless of whether the session is locked_.

Is there a reason not to create lockscreen content "on demand"?

Saviq: > Is there a reason not to create lockscreen content "on demand"?

The reason we're thinking of is visual feedback - if the buffers are "preloaded", the compositor can slot them in immediately, rather than waiting for the lockscreen to "warm up".

Yes, the trade off is higher memory use, so we should support both modes.

Saviq: Where do you (not) see it? Or you mean you don't expect it to be noticable?

I mean, I suppose the app can have the surface pre-cooked without actually pushing a buffer through - in which case, yeah I agree that shouldn't be noticable.

mattkae:
@Saviq or @RAOF : Another question that I have is: **do we even need the concept of a lockscreen surface**? I'm thinking about it, and I suspect the answer is "no". If we have a registered lockscreen client, then this client may render whatever it likes to as many outputs as it likes, not just a single surface. This is a bit more flexible and removes the complication of another interface. WDYT?

mattkae:
> @Saviq or @RAOF : Another question that I have is: **do we even need the concept of a lockscreen surface**? I'm thinking about it, and I suspect the answer is "no". If we have a registered lockscreen client, then this client may render whatever it likes to as many outputs as it likes while the session is locked. After the session is locked, we can enforce that the client must remove al of its surfaces. This is a bit more flexible and removes the complication of another interface. WDYT?

Although, maybe it's useful to keep track of which buffers are actually lockscreen buffers... I stand corrected ;)

Saviq:
> **do we even need the concept of a lockscreen surface**? I'm thinking about it, and I suspect the answer is "no". If we have a registered lockscreen client, then this client may render whatever it likes to as many outputs as it likes while the session is locked.

How will the lockscreen client draw above everything else?

The client may well not be just a lockscreen (say it's an integrated shell with docks and such), so yes I think it needs to be explicit which surfaces are the lockscreen ones.

mattkae:
> How will the lockscreen client draw above everything else?
> 
> The client may well not be just a lockscreen (say it's an integrated shell with docks and such), so yes I think it needs to be explicit which surfaces are the lockscreen ones.

Yup agreed. We should at least mention which surfaces are the lockscreen surfaces.

mattkae:
I have come up with a scheme that is much better, in my opinion. Here's how it goes down:

1. A client registers as the lockscreen client. Being the lockscreen client means that only this client can submit surfaces that will be seen while the session is locked
2. When the session is locked, the compositor immediately obfuscates its outputs and displays all previously submitted lockscreen surfaces, if any
3. The lockscreen client may continue to submit more surfaces while the screen is locked
4. When the screen is unlocked, the compositor will hide all lockscreen surfaces

Where this differs a lot from the previous draft is that the client may submit has many surfaces as it likes, instead of just one per output. This gives us a lot more flexibility to support a lockscreen with any level of complication, which I think is a good thing.

This new draft also gives us the best of both worlds in that a client can choose to submit the buffer beforehand if they so desire, but they could not do so if the deem the memory footprint too high for them (e.g. submit on demand).

Also, like the previous draft, it fixes the VT switch bug.

Let me know what you think! I

RAOF (commented):
There's no longer any way for a client to present a surface on a specific output, or to know what size to make a surface?

RAOF:
Ok. I think we need to work out what we user-visible behaviour we *actually* want, and work backwards from there.

So, in order from absolutely most bare bones to fanciest shiny behaviour: 

____________________________________________
We lock the screen by just not rendering anything and having `glClear()` black our outputs for us.

What we absolutely need:
* A way to blank the screen
  * This can be very easily handled in Mir; just don't pass any surfaces to the compositor.
* A way to *unlock* the session
  * This means we need at minimum to be able to put up a password entry box and then communicate with logind to unlock the session
  * Handling text rendering/layout, input, and such means that we'll want to use a toolkit, which means we'll need a Wayland client that can display when the screen is locked. For robustness this client should be out of process.
  * Because this is only needed to *unlock*, it's not time-sensitive, so we can start it whenever
  * This client *does not* necessarily need any special protocol - because it only needs to be presented for unlock, we can spawn the unlock helper on a special socket to treat it specially.

____________________________________________

We lock the screen by immediately displaying a static image

What we absolutely need:
* Some way to draw the image
  * This *could* be done easily enough in the compositor, although we now need an image loading library
  * This could be an Wayland client with some kind of lock screen protocol. That protocol could:
    * Have the client supply the images ahead of time, as lockscreen surfaces. In this case, the compositor can easily and instantly switch to those static images on lock. Or
    * Have a “please supply some lockscreen images now” event, and have the compositor *stop compositing* until those lockscreen surfaces have arrived or a timeout elapses.
* As above, some way to unlock the session. That could be the same client that provides the lockscreen images, or it could be some other client.

____________________________________________

We lock the screen by displaying a static image, with a transition

Two options here:
* The compositor animates the transition. In this case, all the same mechanisms as above work
* The lockscreen client animates the transition. In this case, only the “please supply some lockscreen images now” event works, as the client needs to know when to *start* animating the transition.

____________________________________________

We lock the screen by heavily blurring the existing content (or something like that)

This cannot be reasonably done by an external client; this would require `Renderer` participation (and as such couldn't easily be done in a Mir consumer, it would require we do it in Mir).
This also requires an unlock client.

____________________________________________

I think we need to work out what point on this spectrum we're aiming for, as it has significant influence on what any protocol will need to do - or even if we need *any* extra protocol at all.

PR #3219, opened 2024-02-06
miroil: Pass through list of GLRenderingProviders when making the Compositor`

RAOF:
The `Compositor` is going to need to use one-or-more rendering providers, so there needs to be some way for the consumer to get them.


RAOF:
This is draft because the qtmir side of this is not quite finished, and also because I *think* that qtmir will have use for a slightly less debug-log-only version of https://github.com/MirServer/mir/pull/3212

RAOF:
Closing for now; I'll open something like this again when QtMir is further along.

PR #3220, opened 2024-02-09
Better platform selection

RAOF:
Implement, and test, the new probing algorithm:

* If any nested platform claims support, select only the single nested platform with highest support level.
  * If multiple nested platforms claim equal support, the first platform in reverse alphabetical order of DSO filename is chosen
* Otherwise, for each hardware device select the hardware platform that claims the highest support level.
  * If multiple hardware platforms claim the same support level for a device, the first platform in reverse alphabetical order of DSO filename is chosen


AlanGriffiths (changes requested):
TODO: fix the code:

```plain
 ../../../../src/server/graphics/platform_probe.cpp: In function ‘std::vector<std::pair<mir::graphics::SupportedDevice, std::shared_ptr<mir::SharedLibrary> > > mir::graphics::modules_for_device(const std::function<std::vector<mir::graphics::SupportedDevice>(const mir::SharedLibrary&)>&, const std::vector<std::shared_ptr<mir::SharedLibrary> >&, mir::graphics::TypePreference)’:
../../../../src/server/graphics/platform_probe.cpp:254:1: error: control reaches end of non-void function [-Werror=return-type]
  254 | }
      | ^
cc1plus: all warnings being treated as errors
```

RAOF:
This does *not* quirk the `gbm-kms` platform to return `supported` (rather than `best`) for NVIDIA, as there's already a `gbm-kms` quirk skipping nvidia-driven DRM nodes entirely.

RAOF:
The test failure here is some difference in environment between `make -C build-amd64 ptest` and `debian/rules binary` :confused:

Saviq:
> The test failure here is some difference in environment between `make -C build-amd64 ptest` and `debian/rules binary` 😕

If you can't reproduce locally, remember there's this (restart with debug logging and ssh into the environment on failure):

https://github.com/canonical/mir/blob/117a4df4815995950c413be967ea009bda050a86/.github/workflows/spread.yml#L121-L125

src/server/graphics/platform_probe.cpp
```diff
@@ -168,22 +179,32 @@ auto modules_for_device(
                         best_modules_so_far.emplace_back(std::move(device), module);
                     }
                 }
-                else if (device.support_level > mg::probe::unsupported)
+                else
                 {
-                    // Devices with null associated udev device are not combined with any others
-                    best_modules_so_far.emplace_back(std::move(device), module);
+                    /* Platforms with no associated `mir::udev::Device` are *nested* platforms
+                     * We want at most one of these
```
AlanGriffiths, line 185: We have `probe::hosted`: is using both "hosted" and "nested" confusing?

Saviq:
> Implement, and test, the new probing algorithm:

What about the virtual platform?

Looks like something got broke with it:
```
ERROR: /home/runner/work/mir/mir/src/server/graphics/default_configuration.cpp(198): Throw in function virtual const std::vector<std::shared_ptr<mir::graphics::DisplayPlatform> >& mir::DefaultServerConfiguration::the_display_platforms()
std::exception::what: Exception while creating graphics platform
ERROR: /home/runner/work/mir/mir/src/server/graphics/platform_probe.cpp(369): Throw in function std::vector<std::shared_ptr<mir::SharedLibrary> > {anonymous}::select_platforms_from_list(const string&, const std::vector<std::shared_ptr<mir::SharedLibrary> >&)
std::exception::what: Failed to find all requested platform modules.
Detected modules are: 
	mir:x11
	mir:egl-generic
	mir:wayland
	throw-on-creation
	mir:gbm-kms
	mir:eglstream-kms
	mir:stub-graphics
Failed to find:
	mir:virtual
```

src/server/graphics/platform_probe.cpp
RAOF: Hm. There's no longer really any need for `probe::hosted` *at all*; we they're on a nested/hardware is now a separate axis to unsupported/best.

Should we *remove* `probe::hosted`?

Saviq: Or just stick to `probe::hosted`, rather than `::nested`?

AlanGriffiths: Well, _if_ the intention of this PR is to change the terminology from "hosted" to "nested" it makes sense to remove (or rename) `probe::hosted`.

But I've memories of long ago replacing "nested" with "hosted" and wonder if we want revisit that again?

tests/unit-tests/graphics/test_platform_prober.cpp
```diff
@@ -281,3 +294,742 @@ TEST(ServerPlatformProbe, IgnoresNonPlatformModules)
...
+{
+public:
+    using PlatformHandle = mir::SharedLibrary*;
+
+    auto add_platform() -> PlatformHandle
+    {
+        /* SharedLibrary ends up calling dlopen; dlopen(nullptr) is valid, and returns a handle to
+         * the current binary. We don't care about that, we won't be dlsymming anything, we just want
+         * a valid SharedLibrary.
+         */
```
AlanGriffiths, line 308: :hear_no_evil:

AlanGriffiths (approved):
We should use either "nested, or "hosted" consistently. I don't have strong feelings about which.

Approving. (But not queueing, so that the bike-shed can look its best)

AlanGriffiths (approved):
OK

@Saviq want to take a last look?

src/server/graphics/platform_probe.cpp
```diff
@@ -203,8 +224,55 @@ auto modules_for_device(
...
+     * GCC does not use this particular UB to provide a better diagnostic,
+     * so we need to provide a default case.
+     *
+     * As normal, leave it guarded by !Clang, so our clang builds will
+     * error out at compile-time if we add a new enum variant.
+     */
+    default:
+        mir::fatal_error("Impossible TypePreference");
+        // We can't get here, either, but sadly we can't mark mir::fatal_error as [[noreturn]]
+        // to get the compiler to enforce that 🤦‍♀️
```
Saviq, line 253: Do we want to introduce emojis (or non-ASCII, for that matter)? :)

src/server/graphics/platform_probe.cpp
```diff
@@ -231,5 +300,238 @@ auto mir::graphics::rendering_modules_for_device(
...
+            if (auto const i = std::find(requested_modules.begin(), requested_modules.end(), description->name);
+                i != requested_modules.end())
+            {
+                selected_modules.push_back(module);
+                requested_modules.erase(i);
+            }
+        }
+        catch (std::exception const&)
+        {
+            // Should we log anything here?
```
Saviq, line 351: Should probably resolve that question before merging…

Saviq (approved):
Just two nits about comments inline, feel free to merge :)

src/server/graphics/platform_probe.cpp
RAOF: 🤷‍♀️ 😛

src/server/graphics/platform_probe.cpp
RAOF: Hm. Maybe the resolution should be to remove the `try` entirely, and make “we're passed a list of graphics platform modules” a precondition (one that already hold).

AlanGriffiths (approved):

PR #3221, opened 2024-02-12
jira: add sync config

Saviq:
Ref.: https://github.com/canonical/gh-jira-sync-bot

### DIFF
--- a/.github/.jira_sync_config.yaml
+++ b/.github/.jira_sync_config.yaml
@@ -1,12 +1,13 @@
 settings:
   jira_project_key: MIRENG
+  # Only act on issues labeled "triaged"
+  labels:
+  - triaged
   status_mapping:
     opened: Triaged
     closed: Done
   components:
   - Mir
-  labels:
-  - triaged
   label_mapping:
     enhancement: Story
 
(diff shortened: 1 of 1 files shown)


.github/.jira_sync_config.yaml
```diff
@@ -0,0 +1,12 @@
+settings:
+  jira_project_key: MIRENG
+  status_mapping:
+    opened: Triaged
```
AlanGriffiths, line 4: Do we really want this? I'd expect "Untriaged" as per the linked README

Saviq: Since this will only act on issues that we put a label on (I said "triaged"), I thought maybe we don't need this extra step.

Can be convinced otherwise, but we'd need a different label to key on.

AlanGriffiths: OK, I thought it would "triage" everything that was open

Saviq: I'll add some comments on the config to clarify

Saviq: Done.

AlanGriffiths: OTOH, we have many Github labels I wouldn't consider "Triaged", E.g. "wontfix".

Saviq: Right, but do we want/need to add them to JIRA? I had this as "we want to work on this soon, add it to JIRA so we can plan it in".

AlanGriffiths (approved):

PR #3232, opened 2024-02-13
feature: logind control over a locked session in mir

mattkae:
## What's new?
- You can do a `loginctl lock-session <ID>` to lock the session
- You can do a `loginctl unlock-session <ID>` to unlock the session
- In MirAL, you can receive a notification about locks and unlocks, where you can do whatever you like (e.g. spawn a lockscreen process)

Note that you need to be in an environment where `LogindConsoleServices` is being used for this to function properly.

## How to use
In your main  runner function, you can do this:

```c++
return runner.run_with(
        {
           ....
            SessionLockListener(
                [&]() { printf("Locked\n"); },
                [&]() { printf("unlocked\n"); }
            )
       });
```

Then, when you lock the screen, you will see the `printf`s happening in console :tada:


include/platform/mir/console_services.h
```diff
@@ -152,6 +152,10 @@ class ConsoleServices
         int major, int minor,
         std::unique_ptr<Device::Observer> observer) = 0;
 
+    virtual void register_lock_handler(
+        std::function<void()> const& lock,
+        std::function<void()> const& unlock) = 0;
```
AlanGriffiths, line 157: This is two-stage initialization (and only needed by one of multiple subclasses).

A cleaner design is to have a `ConsoleLocker` interface (implemented by `SessionLocker`) and pass that to the `LogindConsoleServices` constructor.

src/server/display_server.cpp
```diff
@@ -185,6 +191,16 @@ struct mir::DisplayServer::Private
...
+    void lock()
+    {
+        session_locker->request_lock();
+    }
+
+    void unlock()
+    {
+        session_locker->request_unlock();
+    }
+
```
AlanGriffiths, line 203: I don't think we need involve the `DisplayServer::Private` in forwarding these calls: `the_session_locker()` should be a dependency resolved by `the_console_services()`. (I.e. have it implement an interface used by `LogindConsoleServices` and supply it in `the_console_services()`.

AlanGriffiths (changes requested):

AlanGriffiths:
> ## Questions
> 
>     1. If the client unlocks the screen but logind has _not yet_ unlocked the screen, do we unlock the screen immediately? Or do we wait for logind? Also, do we quit the client immediately (and show the black screen), or should the client wait for logind as well before exiting?

AIUI we are supporting two scenarios:

1. A client invokes locking (e.g. swaylock) and holds the lock until released; and,
2. Logind  invokes locking and we start a client (that also locks?) and then both the client and logind unlock (with a possible delay)

The simplest thing that might possibly work is to have a lock count and resume normal operation when that hits zero.

@Saviq WDYT?

include/platform/mir/console_services.h
AlanGriffiths: You've acked this but not changed it.

src/server/scene/session_locker.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace ms = mir::scene;
+
+ms::SessionLocker::SessionLocker(
+    std::shared_ptr<mir::ConsoleServices> const& console_services,
+    std::shared_ptr<mf::SurfaceStack> const& surface_stack)
+    : surface_stack{surface_stack}
+{
+    console_services->register_lock_handler(
+        [&] { request_lock(); },
+        [&] { request_unlock(); });
```
AlanGriffiths, line 32: Not only is this two-phase initialisation still, it may be initialising an object that's about to be dropped on the floor. (Nothing locally ensures that `console_services` isn't released when the constructor returns.

AlanGriffiths (changes requested):
I don't think you understood my earlier review

include/platform/mir/console_services.h
mattkae: Oh I see now

RAOF:
> > ## Questions
> > 
> > 1. If the client unlocks the screen but logind has _not yet_ unlocked the screen, do we unlock the screen immediately? Or do we wait for logind? Also, do we quit the client immediately (and show the black screen), or should the client wait for logind as well before exiting?
> 
> AIUI we are supporting two scenarios:
> 
>     1. A client invokes locking (e.g. swaylock) and holds the lock until released; and,
> 
>     2. Logind  invokes locking and we start a client (that also locks?) and then both the client and logind unlock (with a possible delay)
> 
> 
> The simplest thing that might possibly work is to have a lock count and resume normal operation when that hits zero.
> 
> @Saviq WDYT?

I think we want the opposite behaviour, actually. Locking via starting the lockscreen (eg swaylock) *or* receiving a `Locked()` signal from logind should lock the session, and unlocking the lock screen *or* receiving an `Unlocked()` signal from logind should unlock the session.

The important properties are that however the user locks the screen, the screen should be locked, and however the user tries to unlock the screen the screen should unlock.

Logind's `Lock()` and `Unlock()` methods are conveniently idempotent, so we don't need to worry too much about races there.

Saviq:
> The important properties are that however the user locks the screen, the screen should be locked, and however the user tries to unlock the screen the screen should unlock.

You're not wrong… I'm getting convinced that'd be DTRT. So then, it's either the client releasing the lock, and Mir unlocking itself after, or Mir receiving the `Unlock()` signal from Logind and telling the lockscreen to go away… feels like we've gone full circle (if not many…) on this.

src/server/scene/session_locker.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+    if (!surface_stack->screen_is_locked())
+    {
+        if (lock_count != 0)
+            mir::fatal_error("Lock count should be zero");
+
+        lock_count++;
+        screen_lock_handle = surface_stack->lock_screen();
+        for (auto const& observer : observers)
+            observer->on_lock();
```
RAOF, line 52: This is unsafe if `on_lock` calls either `add_observer` or `remove_observer` (and is UB if other threads call `add_observer` or `remove_observer`, too).

`mir::ObserverMultiplexer<SessionLockObserver>` handles all this.

AlanGriffiths (changes requested):
Our discussion yesterday was based on "logind (if used) is the source of truth". But...

Today @RAOF makes a good argument that "Mir is the source of truth" and that we take lock and unlock requests from both logind and lockscreen client and tell logind what is happening. I'm good with that.

mattkae:
> I think we want the opposite behaviour, actually. Locking via starting the lockscreen (eg swaylock) _or_ receiving a `Locked()` signal from logind should lock the session, and unlocking the lock screen _or_ receiving an `Unlocked()` signal from logind should unlock the session.
> 

Wouldn't this mean that the following scenario can happen:
1. Logind sends a `Lock()` signal
2. Swaylock starts up
3. Swaylock unlocks
4. Logind is left in a "locked" state, since no one has called an authenticated `Unlock()` on it (e.g. `LockedHint` will still be true, and anyone else dependent on this signal will have never been notified about it)

mattkae:
> Today @RAOF makes a good argument that "Mir is the source of truth" and that we take lock and unlock requests from both logind and lockscreen client and tell logind what is happening. 

We can't tell logind what is happening from the compositor, specifically for unlocks. To call the `Unlock()` method, a client needs to be authenticated via PAM (or otherwise). Hence, if the lockscreen client authenticates with PAM but the compositor does not, the compositor will not have the ability to call that method.

Saviq:
> 4. Logind is left in a "locked" state, since no one has called an authenticated `Unlock()` on it (e.g. `LockedHint` will still be true, and anyone else dependent on this signal will have never been notified about it)

> We can't tell logind what is happening from the compositor, specifically for unlocks. To call the `Unlock()` method, a client needs to be authenticated via PAM (or otherwise). Hence, if the lockscreen client authenticates with PAM but the compositor does not, the compositor will not have the ability to call that method.

As discussed, yeah it can. The compositor alone is what interacts with Logind - either by reacting to the `Lock()` signal, starting the ext-session-lock-v1 client on its own - or from an externally ran ext-session-lock-v1 client and then going `setLockedHint(true)`. There's no need to call `Lock()` then, the compositor is meant to listen to the signals and setting the hint. Something else (shell components, unlock client) is responsible to call the `(Un)Lock()` methods.

mattkae:
Listing out the scenarios that we need to cover:

1. Logind `Lock()`s
  - Session is locked
  - Client starts
  - Client calls lock
  - Logind `Lock()` is called
  - Lock is a noop since we're already locked
2. Client requests a lock
  - Logind `Lock()` is called
  - Logind `Lock()`s
  - Session is locked

Saviq:
>     1. Logind `Lock()`s
>     * Logind `Lock()` is called
>     * Lock is a noop since we're already locked

Who would call `Lock()` here?

mattkae:
> Who would call `Lock()` here?

The compositor. Clients no longer speak with logind.

Saviq:
> > Who would call `Lock()` here?
> 
> The compositor. Clients no longer speak with logind.

No need to call `Lock()` when it's already locked, only `setLockedHint(true)`.

These are the two lock scenarios:
1. session calls logind `Lock()`
   1. Mir locks and launches the lockscreen in response to the `Lock` signal
   2. Proceed from 2.

2. lockscreen does the ext-session-lock-v1 dance
   1. Mir locks, if not locked already
   2. Mir calls `setLockedHint(true)`

And two unlock scenarios:
1. something calls logind `Unlock()`
   1. Mir does the ext-session-lock-v1 `finished` dance
   2. Mir unlocks
   3. Mir calls `setLockedHint(false)`

1. ext-session-lock-v1 `unlock_and_destroy`
   1. Mir unlocks, if locked
   2. Mir calls `setLockedHint(false)`

src/server/console/logind_console_services.cpp
```diff
@@ -43,21 +44,23 @@ class GErrorPtr
...
+    GErrorPtr &operator=(GErrorPtr const &) = delete;
 
     ~GErrorPtr()
     {
         if (raw_error)
             g_error_free(raw_error);
     }
 
-    GError** operator&()
+    GError **operator&()
```
AlanGriffiths, line 58: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -43,21 +44,23 @@ class GErrorPtr
...
     {
         return &raw_error;
     }
 
-    GError* operator->()
+    GError *operator->()
```
AlanGriffiths, line 63: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -68,7 +71,7 @@ class GErrorPtr
     }
 
 private:
-    GError* raw_error;
+    GError *raw_error;
```
AlanGriffiths, line 74: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -78,9 +81,9 @@ class GErrorPtr
     ::boost::throw_line((int)__LINE__))
 
 std::unique_ptr<LogindSession, decltype(&g_object_unref)> simple_proxy_on_system_bus(
-    mir::GLibMainLoop& ml,
-    GDBusConnection* connection,
-    char const* object_path)
+    mir::GLibMainLoop &ml,
+    GDBusConnection *connection,
+    char const *object_path)
```
AlanGriffiths, line 86: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -104,9 +107,9 @@ std::unique_ptr<LogindSession, decltype(&g_object_unref)> simple_proxy_on_system
 
                 auto error_msg = error ? error->message : "unknown error";
                 BOOST_THROW_EXCEPTION((
-                    std::runtime_error{
-                        "Failed to connect to DBus interface at "s +
-                        object_path + ": " + error_msg}));
+                                          std::runtime_error{
+                                              "Failed to connect to DBus interface at "s +
+                                              object_path + ": " + error_msg}));
```
AlanGriffiths, line 112: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -119,9 +122,9 @@ std::unique_ptr<LogindSession, decltype(&g_object_unref)> simple_proxy_on_system
 }
 
 std::unique_ptr<LogindSeat, decltype(&g_object_unref)> simple_seat_proxy_on_system_bus(
-    mir::GLibMainLoop& ml,
-    GDBusConnection* connection,
-    char const* object_path)
+    mir::GLibMainLoop &ml,
+    GDBusConnection *connection,
+    char const *object_path)
```
AlanGriffiths, line 127: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -145,9 +148,9 @@ std::unique_ptr<LogindSeat, decltype(&g_object_unref)> simple_seat_proxy_on_syst
                 using namespace std::literals::string_literals;
                 auto error_msg = error ? error->message : "unknown error";
                 BOOST_THROW_EXCEPTION((
-                    std::runtime_error{
-                        "Failed to connect to DBus interface at "s +
-                        object_path + ": " + error_msg}));
+                                          std::runtime_error{
+                                              "Failed to connect to DBus interface at "s +
+                                              object_path + ": " + error_msg}));
```
AlanGriffiths, line 153: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -159,7 +162,7 @@ std::unique_ptr<LogindSeat, decltype(&g_object_unref)> simple_seat_proxy_on_syst
         }).get();
 }
 
-std::string object_path_for_current_session(LogindSeat* seat_proxy)
+std::string object_path_for_current_session(LogindSeat *seat_proxy)
```
AlanGriffiths, line 165: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -189,7 +192,7 @@ std::string object_path_for_current_session(LogindSeat* seat_proxy)
 }
 
 std::unique_ptr<GDBusConnection, decltype(&g_object_unref)>
-connect_to_system_bus(mir::GLibMainLoop& ml)
+connect_to_system_bus(mir::GLibMainLoop &ml)
```
AlanGriffiths, line 195: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -205,8 +208,8 @@ connect_to_system_bus(mir::GLibMainLoop& ml)
     {
         auto error_msg = error ? error->message : "unknown error";
         BOOST_THROW_EXCEPTION((
-            std::runtime_error{
-                "Failed to find address of DBus system bus: "s + error_msg}));
+                                  std::runtime_error{
+                                      "Failed to find address of DBus system bus: "s + error_msg}));
```
AlanGriffiths, line 212: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -229,17 +232,45 @@ connect_to_system_bus(mir::GLibMainLoop& ml)
             {
                 auto error_msg = error ? error->message : "unknown error";
                 BOOST_THROW_EXCEPTION((
-                    std::runtime_error{
-                        "Failed to connect to DBus system bus: "s + error_msg}));
+                                          std::runtime_error{
+                                              "Failed to connect to DBus system bus: "s + error_msg}));
```
AlanGriffiths, line 236: There are a lot of spurious, whitespace changes to this file. A lot of them away from project style

src/server/console/logind_console_services.cpp
```diff
@@ -229,17 +232,45 @@ connect_to_system_bus(mir::GLibMainLoop& ml)
...
 }
+
+class LogindSessionLockObserver : public mir::frontend::SessionLockObserver
+{
+public:
+    LogindSessionLockObserver(
+        std::function<void()> const& on_lock,
+        std::function<void()> const& on_unlock)
+        : on_lock_{on_lock},
+          on_unlock_{on_unlock}
```
AlanGriffiths, line 250: As `LogindSessionLockObserver` is only used in this file, is there any need for an indirection through `std::function`? Can't `logind_session_set_locked_hint()`  be called directly?

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -157,9 +211,22 @@ void mf::SessionLockV1::get_lock_surface(wl_resource* id, wl_resource* surface_r
...
+mf::SessionLockV1::SessionLockV1Observer::SessionLockV1Observer(mf::SessionLockV1& lock)
+    : lock{lock}
+{
+}
+
+void mf::SessionLockV1::SessionLockV1Observer::on_lock() {}
+
+void mf::SessionLockV1::SessionLockV1Observer::on_unlock()
+{
+    lock.send_finished_event();
```
AlanGriffiths, line 227: I think this needs to be executed on the wayland thread

AlanGriffiths (changes requested):

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -127,23 +172,32 @@ void mf::SessionLockManagerV1::Instance::lock(wl_resource* lock)
...
+                         "The compositor may be stuck in a locked state forever.");
+        manager.try_relinquish_locking_privilege(this);
     }
     else
     {
-        lock_handle->allow_to_be_dropped();
+        if (!manager.try_unlock(this))
+            mir::log_warning("Duplicate SessionLockV1 will not cause unlock");
     }
 }
```
RAOF, line 202: Oh! I was wondering about this, and it turns out this logic here is because we have a bug in our wrapper generator!

There are *two* requests that destroy `SessionLockV1`: `destroy()` and `unlock_and_destroy()`, but our wrapper generator doesn't expose that and instead just destroys the object in the implementation for both. This would be easier (and we'd be better able to be protocol-compliant) if the generator [did something different](https://github.com/canonical/mir/issues/3241)

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -127,23 +172,32 @@ void mf::SessionLockManagerV1::Instance::lock(wl_resource* lock)
...
     : mw::SessionLockV1(new_resource, Version<1>()),
       manager{manager},
-      lock_handle{manager.surface_stack->lock_screen()}
+      observer{std::make_shared<SessionLockV1Observer>(*this)}
 {
-    send_locked_event();
+    if (manager.try_lock(this))
+    {
+        send_locked_event();
+        manager.session_locker->register_interest(observer, manager.wayland_executor);
```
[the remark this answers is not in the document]
AlanGriffiths: No, a comment here that wouldn't have helped

src/include/server/mir/frontend/session_locker.h
```diff
@@ -0,0 +1,52 @@
...
+
+/// Responsible for triggering session locks and notifying observers
+class SessionLocker : public ObserverMultiplexer<SessionLockObserver>
+{
+protected:
+    explicit SessionLocker(Executor& default_executor)
+        : ObserverMultiplexer{default_executor}
+    {
+    }
+};
```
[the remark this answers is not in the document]
AlanGriffiths: In the current code the only reason I see for `SessionLocker` to exist is to allow declarations. (There's are a bunch of places that write `class SessionLocker;`.)

However, I don't think that all users of `SessionLocker` should automatically have access to the public interface of `ObserverMultiplexer` - that should really be an implementation detail and suggests the abstraction is wrong.

Talking of wrong abstractions: what _should_ the interface of `SessionLocker` be? Well, `lock()` and `unlock()` seem to be missing...

src/server/console/logind_console_services.cpp
```diff
@@ -43,21 +44,23 @@ class GErrorPtr
         : raw_error{nullptr}
     {
     }
-    GErrorPtr(GErrorPtr const&) = delete;
-    GErrorPtr& operator=(GErrorPtr const&) = delete;
+
+    GErrorPtr(GErrorPtr const &) = delete;
+
+    GErrorPtr &operator=(GErrorPtr const &) = delete;
```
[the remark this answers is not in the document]
mattkae: this surprises me too! I must have accidentally formatted (maybe some auto formatting...)

src/include/server/mir/frontend/session_locker.h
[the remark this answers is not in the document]
mattkae: Yeah I _unfortunately_ ended up calling them `on_lock()` and `on_unlock()`, which does feel wrong.

I think my understanding of `ObserverRegistrar` vs `ObserverMultiplexer` is flawed. I suspect in this case that I want an `ObserverRegistrar`, as the `SessionLocker` itself is not an observer.

Also, I agree that the abstract `SessionLocker` class is a useless indirection.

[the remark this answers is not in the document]
mattkae: Fixed :ok_hand:

src/include/server/mir/frontend/session_locker.h
```diff
@@ -0,0 +1,58 @@
...
+public:
+    virtual ~SessionLocker() = default;
+    virtual void lock() = 0;
+    virtual void unlock() = 0;
+    virtual void register_interest(std::weak_ptr<SessionLockObserver> const& observer) = 0;
+    virtual void register_interest(
+        std::weak_ptr<SessionLockObserver> const& observer,
+        Executor& executor) = 0;
+    virtual void unregister_interest(SessionLockObserver const& observer) = 0;
+};
```
AlanGriffiths, line 53: Interfaces ought to delete CopyAssign

src/server/console/logind_console_services.cpp
AlanGriffiths: This^

src/server/frontend_wayland/session_lock_v1.cpp
AlanGriffiths: Probably worth fixing the generator and then this

AlanGriffiths (changes requested):

src/server/frontend_wayland/session_lock_v1.cpp
mattkae: I can work on that today :+1:

mattkae:
## Known bug with swaylock as the lock screen
https://github.com/swaywm/swaylock/issues/346

Because of this, we get into a _tricky_ situation where:
1. The lockscreen should close
2. The lockscreen exits with `2`
3. Miriway sees a failure in a shell component and restarts it
4. This triggers an `on_lock` event which tries to start the lockscreen
5. The lockscreen repeatedly fails to start because there is already a lockscreen running :facepalm: 

I believe the spec says otherwise

mattkae:
AAAAnd now I need to fix https://github.com/canonical/mir/pull/3251 before this can go in :upside_down_face:

mattkae:
@AlanGriffiths and @Saviq : This should be GTG again :+1:

tests/unit-tests/console/test_logind_console_services.cpp
RAOF, line 1: Non-blocking: it would be nice to have a test or two for the interaction between session locking and `LogindConsoleServices`.

RAOF (changes requested):

src/include/server/mir/frontend/session_locker.h
```diff
@@ -0,0 +1,62 @@
...
+public:
+    SessionLockObserver() = default;
+    virtual ~SessionLockObserver() = default;
+    SessionLockObserver (const SessionLockObserver&) = delete;
+    SessionLockObserver& operator= (const SessionLockObserver&) = delete;
+    virtual void on_lock() = 0;
+    virtual void on_unlock() = 0;
+};
+
+/// Responsible for triggering session locks and notifying observers
```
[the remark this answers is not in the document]
AlanGriffiths: I think the name and description are wrong and that adds dissonance to the "and".

This interface doesn't "trigger session locks" - the trigger is what causes the call to `lock()`, this interface "locks and unlocks the session while informing observers".

Calling it `SessionLock` would better reflect its behaviour and be consistent with the `SessionLockObserver` naming.

(With that naming scheme I don't have a problem with there being functions to attach & detach observers as part of the interface)

src/server/scene/session_locker.cpp
```diff
@@ -0,0 +1,82 @@
...
+{
+}
+
+void ms::SessionLocker::lock()
+{
+    if (!surface_stack->screen_is_locked())
+    {
+        screen_lock_handle = surface_stack->lock_screen();
+        multiplexer.on_lock();
+    }
```
AlanGriffiths, line 54: This looks racy. What happens if another thread calls `SessionLocker::lock()` between the calls to `screen_is_locked()` and `lock_screen()`?

src/server/scene/session_locker.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+void ms::SessionLocker::unlock()
+{
+    if (surface_stack->screen_is_locked())
+    {
+        screen_lock_handle->allow_to_be_dropped();
+        screen_lock_handle.reset();
+        multiplexer.on_unlock();
+    }
```
AlanGriffiths, line 64: This looks racy

src/server/scene/session_locker.h
```diff
@@ -0,0 +1,75 @@
...
+
+class SessionLockerObserverMultiplexer : public ObserverMultiplexer<mf::SessionLockObserver>
+{
+public:
+    explicit SessionLockerObserverMultiplexer(std::shared_ptr<Executor> const& executor);
+    void on_lock() override;
+    void on_unlock() override;
+};
+
+class SessionLocker : public mf::SessionLocker
```
AlanGriffiths, line 49: I'm not convinced this is a class in it's own right. Certainly, it has a very intimate relationship with the `SurfaceStack`.

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -127,24 +177,18 @@ void mf::SessionLockManagerV1::Instance::lock(wl_resource* lock)
 
 // SessionLockV1
 
-mf::SessionLockV1::SessionLockV1(wl_resource* new_resource, mf::SessionLockManagerV1& manager)
+mf::SessionLockV1::SessionLockV1(wl_resource* new_resource, SessionLockManagerV1& manager)
     : mw::SessionLockV1(new_resource, Version<1>()),
       manager{manager},
-      lock_handle{manager.surface_stack->lock_screen()}
```
AlanGriffiths, line 133: With the `SessionLocker` we supersede and no longer need `mf::SurfaceStack::lock_screen()` (and mf::SurfaceStack::ScreenLockHandle`)?

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -157,9 +201,64 @@ void mf::SessionLockV1::get_lock_surface(wl_resource* id, wl_resource* surface_r
...
 
+void mf::SessionLockV1::destroy()
+{
+    // Note that this is a deviation from the spec. The spec reads:
+    // "It is a protocol error to make this request if the locked event was sent,
+    // the unlock_and_destroy request must be used instead."
+    //
+    // In our case, it is ONLY a protocol error if the session is still locked,
+    // as the session may have been unlocked via logind independent of the
+    // wayland protocol.
```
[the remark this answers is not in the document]
AlanGriffiths: I don't see how "the session may have been unlocked via logind" affects whether the locked event was sent.

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -157,9 +201,64 @@ void mf::SessionLockV1::get_lock_surface(wl_resource* id, wl_resource* surface_r
...
+
+    if (manager.surface_stack->screen_is_locked())
```
AlanGriffiths, line 218: In the new era `SessionLocker` should we be asking the `surface_stack` about locks?

AlanGriffiths (changes requested):

src/server/console/logind_console_services.cpp
```diff
@@ -238,8 +238,11 @@ connect_to_system_bus(mir::GLibMainLoop& ml)
 }
 }
 
-mir::LogindConsoleServices::LogindConsoleServices(std::shared_ptr<mir::GLibMainLoop> const& ml)
+mir::LogindConsoleServices::LogindConsoleServices(
+    std::shared_ptr<mir::GLibMainLoop> const& ml,
+    std::shared_ptr<frontend::SessionLocker> const& session_locker)
```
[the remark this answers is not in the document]
mattkae: I did what Alan said and made the `SessionLocker` be a `SessionLock`, thus keeping the register/unregister functions on the `SessionLock`

tests/unit-tests/scene/test_session_locker.cpp
```diff
@@ -0,0 +1,100 @@
...
+};
+
+class StubExecutor : public mir::Executor
+{
+public:
+    void spawn(std::function<void()>&& work) override
+    {
+        work();
+    }
+};
```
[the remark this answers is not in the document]
mattkae: Unfortunately, I can't get access to it as a shared_ptr, which is what I want. That's locked inside a cpp file. I could refactor, but it feels a bit overkill to facilitate a test suite IMO (or I can include the cpp file?)

src/server/console/logind_console_services.h
```diff
@@ -19,6 +19,7 @@
 #include <future>
 #include <unordered_map>
 #include "mir/console_services.h"
+#include "mir/frontend/session_locker.h"
```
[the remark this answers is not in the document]
mattkae: I will move it to `scene`, it sounds right :+1:

src/server/scene/session_locker.h
mattkae: I feel that the `SurfaceStack` can also be a `SessionLock`

src/server/scene/session_locker.cpp
mattkae: I think that the new way of doing this (via SurfaceStack) should circumvent these issues

tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -1458,4 +1438,21 @@ TEST_F(SurfaceStack, multiple_surfaces_can_be_swapped)
...
+}
+
+TEST_F(SurfaceStack, observer_is_notified_on_unlock)
+{
+    auto observer = std::make_shared<MockSessionLockObserver>();
+    EXPECT_CALL(*observer, on_unlock()).Times(1);
+    stack.lock();
+    stack.register_interest(observer);
+    stack.unlock();
+}
```
AlanGriffiths, line 1458: I'm sure we've discussed more scenarios to test: multiple `lock()s` for instance.

src/server/scene/surface_stack.cpp
```diff
@@ -599,78 +602,53 @@ auto ms::SurfaceStack::stacking_order_of(SurfaceSet const& surfaces) const -> Su
...
+{
+    if (!screen_is_locked())
     {
-        if (auto const shared = surface_stack.lock())
         {
-            // shared->screen_is_locked() should already be false
-            shared->emit_scene_changed();
+            RecursiveReadLock lk(guard);
+            is_locked = true;
         }
```
AlanGriffiths, line 618: Racy

src/server/scene/surface_stack.cpp
```diff
@@ -599,78 +602,53 @@ auto ms::SurfaceStack::stacking_order_of(SurfaceSet const& surfaces) const -> Su
...
-    }
-
-    ~BasicScreenLockHandle()
-    {
-        if (allowed_to_be_dropped)
         {
-            delete lock;
+            RecursiveReadLock lk(guard);
+            is_locked = false;
         }
```
AlanGriffiths, line 631: Racy

AlanGriffiths (changes requested):
That seems a better architecture. Lots of nits for picking

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -157,9 +212,53 @@ void mf::SessionLockV1::get_lock_surface(wl_resource* id, wl_resource* surface_r
...
 
+void mf::SessionLockV1::destroy()
+{
+    if (manager.is_active_lock(this))
+    {
+        manager.try_relinquish_locking_privilege(this);
+    }
+    else
+    {
+        mir::log_error("SessionLockV1 destroy failed");
```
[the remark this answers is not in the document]
mattkae: Nope, the throw to the user is enough :+1:

tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -1448,11 +1456,36 @@ TEST_F(SurfaceStack, observer_is_notified_on_lock)
...
-TEST_F(SurfaceStack, observer_is_notified_on_unlock)
+TEST_F(SurfaceStack, observer_is_notified_only_on_initial_lock)
+{
+    auto observer = std::make_shared<MockSessionLockObserver>();
+    EXPECT_CALL(*observer, on_lock()).Times(1);
+    stack.register_interest(observer);
+    stack.lock();
+    stack.lock();
+    stack.lock();
+}
```
AlanGriffiths, line 1467: The test body doesn't agree with the test name (you likely want to add a strategic `VerifyAndClearExpectations` after the first lock and set new expectations)

tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -1448,11 +1456,36 @@ TEST_F(SurfaceStack, observer_is_notified_on_lock)
...
 {
     auto observer = std::make_shared<MockSessionLockObserver>();
+    EXPECT_CALL(*observer, on_lock()).Times(1);
     EXPECT_CALL(*observer, on_unlock()).Times(1);
+    stack.register_interest(observer);
     stack.lock();
+    stack.unlock();
+    stack.unlock();
+    stack.unlock();
+}
```
AlanGriffiths, line 1479: The test body doesn't agree with the test name

src/server/console/default_configuration.cpp
```diff
@@ -105,8 +106,11 @@ std::shared_ptr<mir::ConsoleServices> mir::DefaultServerConfiguration::the_conso
                 {
                     try
                     {
+                        auto session_lock_instance = the_session_lock();
                         auto const vt_services = std::make_shared<mir::LogindConsoleServices>(
-                            std::dynamic_pointer_cast<mir::GLibMainLoop>(the_main_loop()));
+                            std::dynamic_pointer_cast<mir::GLibMainLoop>(the_main_loop()),
+                            session_lock_instance);
+                        session_lock_instance->register_interest(vt_services);
```
AlanGriffiths, line 113: It looks a bit odd to `register_interest()` here, and not in the constructor. I'd probably prefer the "named constructor idiom" - make the actual constructor private and add a static method to `LogindConsoleServices`:

```c++
auto LogindConsoleServices::create(std::shared_ptr<MainLoop> ml, std::shared_ptr<scene::SessionLock> const& session_lock)
-> std::shared_ptr<mir::LogindConsoleServices>
{
    auto cs = std::make_shared<mir::LogindConsoleServices>(std::move(the_main_loop()), session_lock);
    session_lock->register_interest(vt_services);
    return cs;
}
```

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -157,9 +212,53 @@ void mf::SessionLockV1::get_lock_surface(wl_resource* id, wl_resource* surface_r
...
+    }
+
+}
+
+void mf::SessionLockV1::unlock_and_destroy()
+{
+    if (!manager.try_unlock(this))
+    {
+        mir::log_error("SessionLockV1 unlock_and_destroy failed");
+        BOOST_THROW_EXCEPTION(mw::ProtocolError(
```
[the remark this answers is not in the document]
AlanGriffiths: Logging and throwing STILL seems like overkill

src/server/console/default_configuration.cpp
AlanGriffiths: Oh! You actually need `GLibMainLoop` - I'd move the necessary casts into the constructor/named constructor

mattkae: Also, i'd like to leave the cast in `the_console_services()`, as you technically can't know if the MainLoop coming in has the right type

tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -1458,4 +1438,53 @@ TEST_F(SurfaceStack, multiple_surfaces_can_be_swapped)
...
     Mock::VerifyAndClearExpectations(&observer);
-}
\ No newline at end of file
+}
+
+TEST_F(SurfaceStack, observer_is_notified_on_lock)
+{
+    auto observer = std::make_shared<MockSessionLockObserver>();
+    EXPECT_CALL(*observer, on_lock()).Times(1);
+    stack.register_interest(observer);
```
RAOF, line 1447: If we remove the `Executor` parameter from `SurfaceStack()` then we'll need to explicitly register with the immediate executor here (and the subsequent places):
```suggestion
    stack.register_interest(observer, mir::immediate_executor);
```

RAOF (changes requested):
A couple of nits.

Design wise, I think that `SessionLock` should ` : public ObserverRegistrar<SessionLockObserver>` rather than implement all the methods of `ObserverRegistrar<>` but not BeAn `ObserverRegistrar<>`.

src/server/console/default_configuration.cpp
```diff
@@ -105,8 +106,11 @@ std::shared_ptr<mir::ConsoleServices> mir::DefaultServerConfiguration::the_conso
                 {
                     try
                     {
-                        auto const vt_services = std::make_shared<mir::LogindConsoleServices>(
-                            std::dynamic_pointer_cast<mir::GLibMainLoop>(the_main_loop()));
+                        auto session_lock_instance = the_session_lock();
+                        auto const vt_services = mir::LogindConsoleServices::create(
+                            std::dynamic_pointer_cast<mir::GLibMainLoop>(the_main_loop()),
+                            session_lock_instance);
+                        session_lock_instance->register_interest(vt_services);
```
AlanGriffiths, line 113: The point of introducing `LogindConsoleServices::create()` is that it does this initialization
```suggestion
```

src/server/console/logind_console_services.cpp
```diff
@@ -238,8 +238,21 @@ connect_to_system_bus(mir::GLibMainLoop& ml)
 }
 }
 
-mir::LogindConsoleServices::LogindConsoleServices(std::shared_ptr<mir::GLibMainLoop> const& ml)
+std::shared_ptr<mir::LogindConsoleServices> mir::LogindConsoleServices::create(
+    std::shared_ptr<GLibMainLoop> const& main_loop,
+    std::shared_ptr<scene::SessionLock> session_lock_instance)
+{
+    std::shared_ptr<mir::LogindConsoleServices>cs(new LogindConsoleServices(
+        main_loop, session_lock_instance));
```
AlanGriffiths, line 246:
```suggestion
std::shared_ptr<mir::LogindConsoleServices> mir::LogindConsoleServices::create(
    std::shared_ptr<GLibMainLoop> main_loop,
    std::shared_ptr<scene::SessionLock> session_lock_instance)
{
    std::shared_ptr<mir::LogindConsoleServices>cs(new LogindConsoleServices(
        std::move(main_loop), session_lock_instance));
```

AlanGriffiths (changes requested):
More nits, but I think we're about there

src/include/server/mir/scene/session_lock.h
```diff
@@ -0,0 +1,61 @@
...
+    SessionLock (SessionLock const&) = delete;
+    SessionLock& operator= (SessionLock const&) = delete;
+
+    virtual void lock() = 0;
+    virtual void unlock() = 0;
+    virtual void register_interest(std::weak_ptr<SessionLockObserver> const& observer) = 0;
+    virtual void register_interest(
+        std::weak_ptr<SessionLockObserver> const& observer,
+        Executor& executor) = 0;
+    virtual void unregister_interest(SessionLockObserver const& observer) = 0;
```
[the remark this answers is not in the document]
mattkae: I find the template makes the code a bit messier. Would you be opposed to leaving out the template but making `SessionLock : public ObserverRegistrar<SessionLockObserver>` still?

src/server/console/logind_console_services.cpp
```diff
@@ -238,8 +238,21 @@ connect_to_system_bus(mir::GLibMainLoop& ml)
...
+    session_lock_instance->register_interest(cs);
+    return cs;
+}
+
+mir::LogindConsoleServices::LogindConsoleServices(
+    std::shared_ptr<mir::GLibMainLoop> const& ml,
+    std::shared_ptr<scene::SessionLock> const& session_lock)
     : ml{ml},
+      session_lock{session_lock},
```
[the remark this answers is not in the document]
mattkae: Oh I see, these aren't const ref'd either!

src/include/server/mir/scene/session_lock.h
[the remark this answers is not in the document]
AlanGriffiths: There's no need for the helper template in this PR. (Not sure why it would make anything messy though)

src/include/server/mir/default_server_configuration.h
```diff
@@ -424,6 +425,7 @@ class DefaultServerConfiguration : public virtual ServerConfiguration
     CachedPtr<cookie::Authority> cookie_authority;
     CachedPtr<input::KeyMapper> key_mapper;
     std::shared_ptr<ConsoleServices> console_services;
+    CachedPtr<scene::SessionLock> session_lock;
```
AlanGriffiths, line 428:
```suggestion
```

src/server/scene/surface_stack.cpp
```diff
@@ -599,78 +599,46 @@ auto ms::SurfaceStack::stacking_order_of(SurfaceSet const& surfaces) const -> Su
...
+    return is_locked;
+}
 
-struct ms::SurfaceStack::BasicScreenLockHandle : mf::ScreenLockHandle
+void ms::SurfaceStack::lock()
 {
-    BasicScreenLockHandle(std::shared_ptr<SharedScreenLock> lock)
-        : lock{new std::shared_ptr<SharedScreenLock>{std::move(lock)}}
+    bool expected = false;
+    if (is_locked.compare_exchange_weak(expected, true))
```
AlanGriffiths, line 610:
```suggestion
    bool temp = false;
    if (is_locked.compare_exchange_strong(temp, true))
```

src/server/scene/surface_stack.cpp
```diff
@@ -599,78 +599,46 @@ auto ms::SurfaceStack::stacking_order_of(SurfaceSet const& surfaces) const -> Su
...
-        {
-            fatal_error("The screen was not properly unlocked");
-        }
-    }
-
-    void allow_to_be_dropped() override
+void ms::SurfaceStack::unlock()
+{
+    bool expected = true;
+    if (is_locked.compare_exchange_weak(expected, false))
```
AlanGriffiths, line 620:
```suggestion
    bool temp = true;
    if (is_locked.compare_exchange_strong(temp, false))
```

tests/include/mir/test/doubles/stub_session_lock.h
```diff
@@ -0,0 +1,47 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_TEST_DOUBLES_STUB_session_lock_H
+#define MIR_TEST_DOUBLES_STUB_session_lock_H
```
AlanGriffiths, line 18:
```suggestion
#ifndef MIR_TEST_DOUBLES_STUB_SESSION_LOCK_H
#define MIR_TEST_DOUBLES_STUB_SESSION_LOCK_H
```

AlanGriffiths (changes requested):
More and the same nits

src/include/server/mir/scene/session_lock.h
```diff
@@ -0,0 +1,57 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_FRONTEND_SESSION_LOCK_H
+#define MIR_FRONTEND_SESSION_LOCK_H
```
AlanGriffiths, line 18:
```suggestion
#ifndef MIR_SCENE_SESSION_LOCK_H
#define MIR_SCENE_SESSION_LOCK_H
```

src/include/server/mir/scene/session_lock.h
```diff
@@ -0,0 +1,57 @@
...
+    SessionLock& operator= (SessionLock const&) = delete;
+
+    virtual void lock() = 0;
+    virtual void unlock() = 0;
+};
+
+}
+}
+
+#endif //MIR_FRONTEND_SESSION_LOCK_H
```
AlanGriffiths, line 57:
```suggestion
#endif //MIR_SCENE_SESSION_LOCK_H
```

AlanGriffiths (changes requested):

AlanGriffiths (approved):
Looks OK to me. I'll let the other reviewers have a go

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -59,10 +59,23 @@ class SessionLockV1 : wayland::SessionLockV1
...
+        SessionLockV1& lock;
+    };
+
     void get_lock_surface(wl_resource* id, wl_resource* surface, wl_resource* output) override;
+    void destroy() override;
+    void unlock_and_destroy() override;
 
     mf::SessionLockManagerV1& manager;
-    std::unique_ptr<frontend::ScreenLockHandle> const lock_handle;
+    std::shared_ptr<SessionLockV1Observer> const observer;
```
RAOF, line 78:
```suggestion
    /* Lifetime annotation:
     * This is safe even though `SessionLockV1Observer` keeps a reference to `this`
     * because the `shared_ptr` is only only shared with the `ObserverRegistrar<>`
     * (which uses a `weak_ptr` internally) and observations are delegated to the
     * `WaylandExecutor` so all interactions are single-threaded
     */
    std::shared_ptr<SessionLockV1Observer> const observer;
```

RAOF (approved):
A couple of comment nits/improvements, but good to go.

include/miral/miral/session_lock_listener.h
```diff
@@ -0,0 +1,44 @@
...
+#define MIRAL_SESSION_LOCK_LISTENER_H
+
+#include <functional>
+#include <memory>
+
+namespace mir { class Server; }
+
+namespace miral
+{
+/// Add callbacks to
```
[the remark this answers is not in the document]
RAOF:
```suggestion
/// Add callbacks notifying when the session has been locked and unlocked
```

Issue #3241, opened 2024-02-20
Wrapper generator ignores requests with `destructor` attribute


RAOF:
Lots of Wayland interfaces have a `destroy()` request, with `type="destructor"`, to… destroy the associated object.

The Wayland wrapper generator eats all these requests, generating a thunk that (eventually) calls the class destructor.

This is *mostly* what you want, but the `ext_session_lock_v1` has *two* requests marked `type="destructor"`; the `destroy()` request as normal, and an `unlock_and_destroy()` request. The `destroy` request should be sent if the client failed to lock the screen (and protocol error otherwise), and the `unlock_and_destroy` request should be sent when the client wants to *un*lock the screen (and is a protocol error if the screen is *not* locked).

We should fix up the wrapper generator to support this. I think I favour having the generator generate something like

```
class Foo
{
public:
…
    virtual ~Foo();
…
private:
    virtual void destroy() { };  // Default implementation is a no-op but can be overridden
}
```

mattkae:
> The destroy request should be sent if the client failed to lock the screen (and protocol error otherwise), and the unlock_and_destroy request should be sent when the client wants to unlock the screen (and is a protocol error if the screen is not locked).

To elaborate, there are two scenarios:

1. The client unlocks itself via `unlock_and_destroy` (most likely because someone entered a password in the client's GUI)
2. The server decides that the session should be unlocked. It sends a `finished` event to the client who, in return, replies with a `destroy` request (most likely because logind said that it's time to unlock)

mattkae:
So that `destroy` method should be `protected`, right?

RAOF:
> To elaborate, there are two scenarios:
> 
>     1. The client unlocks itself via `unlock_and_destroy` (most likely because someone entered a password in the client's GUI)
> 
>     2. The server decides that the session should be unlocked. It sends a `finished` event to the client who, in return, replies with a `destroy` request (most likely because logind said that it's time to unlock)

Not quite, for (2). If the client has *ever* received a `locked` event on the `ext_session_lock_v1` object, then it *must* use `unlock_and_destroy` to destroy the lock object, even in response to a `finished` event; otherwise, it *must* use `destroy` to destroy the lock object.

RAOF:
@mattkae - why was this reopened? The wrapper generator changes landed, right?

mattkae:
@RAOF You are correct! I am not sure what button I accidentally clicked on Feb 23

PR #3243, opened 2024-02-20
(#3241) Wayland wrapper generator now generates a virtual stub function for destructors

mattkae:
fixes https://github.com/canonical/mir/issues/3241

## What's new?
- We now generate a virtual stub method with destructors so that you can do something like this:

```c++
void mf::SessionLockV1::unlock_and_destroy()
{
    printf("Being unlocked and destroyed\n");
}
```

This is called right before destruction happens. By default, the method is a stub.

### DIFF
--- a/src/wayland/generator/request.cpp
+++ b/src/wayland/generator/request.cpp
@@ -26,7 +26,7 @@ Emitter Request::virtual_mir_prototype() const
 {
     if (is_destroy())
     {
-        return nullptr;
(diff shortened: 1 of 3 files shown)


RAOF (changes requested):
This looks correct, but you'll need to update the tests, too!

mattkae:
This should be ready now @RAOF  :+1:

RAOF (approved):
Yup! It would be nice to add doxygen comments to these, so that we could get docs in IDEs when implementing these :wink:

PR #3244, opened 2024-02-21
Implements wl_surface enter/leave events

### LINKED ISSUE #342: Missing surface enter and leave events
When a surface is mapped to a display, `wl_surface.enter` should be called, and `wl_surface.leave` should be called when it is no longer visible. There are currently no known bugs caused by this.

hbatagelo:
Fixes #342.

How it works:

* Output tracking is performed in `BasicSurface::track_outputs()`. This function is called within `BasicSurface` whenever the surface is moved, resized, or a new display configuration is applied. Tracking begins for an output when it is _valid_, _used_, and begins to overlap the surface rectangle. An `entered_output` notification is sent when this occurs, and the output configuration ID is retained for tracking purposes. When the tracked output no longer overlaps the surface rectangle or is no longer valid or in use, a `left_output` notification is sent.
* `WaylandSurfaceObserver` listens for `entered_output`/`left_output` notifications and invokes `handle_enter_output(id)`/`handle_leave_output(id)` in `WindowWlSurfaceRole`.
* The handlers in `WindowSurfaceRole` access the `OutputGlobal` for the given IDs and iterate over the currently bound `OutputInstance` objects to send `wl_surface.enter/leave` events.

Special care is taken when the output configuration changes dynamically:

* When an output is disabled at runtime, `handle_leave_output(id)` may be called too late, after the associated `OutputInstance` objects have already been destroyed by the output manager. To prevent this, `ObserverRegistrar` has been modified to allow an observer to be registered as an "early" observer, i.e., one that is always notified before all other "non-early" observers. This approach is used for the display configuration observer registered in `BasicSurface`. Now, the `configuration_applied` notification is always handled before it is processed by the output manager, ensuring that the output is still valid when the leave notification is handled.
* When an output is enabled at runtime, `handle_enter_output(id)` may be called too soon, before the associated `OutputGlobal` or `OutputInstance` bound to the client have been updated. To address this, the `OutputManager` has been modified to allow someone to listen for the creation of an `OutputGlobal` (`output_global_created` notification). `WaylandSurfaceRole` acts as a listener for this notification. Now, if `handler_enter_output(id)` fails to send the "enter" event, the event is saved as pending. Afterwards, when `output_global_created` is received, `WaylandSurfaceRole` registers itself as a listener for `output_config_changed`, which is notified when the `OutputInstance` is finally bound to the client and the pending event can be sent.

Another case that required special consideration is when `track_outputs()` is called from multiple threads, with one of them being the `Mir/Wayland` thread. Because the `WaylandExecutor` eagerly executes works in the Wayland thread, an event in the Wayland thread can be executed before one that has already been enqueued in another thread. This leads to duplicate "enter" and "leave" events, as observed when resizing a window across side-by-side outputs on a QEMU machine. The problem is resolved by invoking `track_outputs()` through `linearising_executor` when it is called in the Wayland thread and waiting for `track_outputs()` to be executed in the other thread before continuing.


src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -468,24 +471,59 @@ void mf::WindowWlSurfaceRole::create_scene_surface()
...
+
+    output_manager->current_config().for_each_output([&](mg::DisplayConfigurationOutput const& conf)
         {
-            auto const output = output_manager->output_for(conf.id);
-            if (output)
+            if (auto const output{output_manager->output_for(conf.id)}; output)
             {
                 output.value()->for_each_output_bound_by(
                     client,
                     [&](OutputInstance* output)
```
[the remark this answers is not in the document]
hbatagelo: Done.

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -468,24 +471,59 @@ void mf::WindowWlSurfaceRole::create_scene_surface()
...
-                        surface.value().send_enter_event(output->resource);
+                        auto const& current_config{output->global.value().current_config()};
+                        auto const tracked_it{std::ranges::find_if(tracked_outputs,
+                            [&](auto const& tracked_output)
+                            {
+                                return tracked_output.config == current_config;
+                            })};
+
+                        auto const output_rect{current_config.extents()};
+                        if (output_rect.overlaps({scene_surface->top_left(), scene_surface->window_size()}))
```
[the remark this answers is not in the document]
hbatagelo: Okay, done.

src/server/frontend_wayland/window_wl_surface_role.h
```diff
@@ -157,6 +159,13 @@ class WindowWlSurfaceRole
 
     std::unique_ptr<shell::SurfaceSpecification> pending_changes;
 
+    struct TrackedOutput
+    {
+        mir::graphics::DisplayConfigurationOutput config;
+        bool bound{true};
```
AlanGriffiths, line 165: Somehow `bound` seems misleading as it is more an "in use", sometimes reflecting whether we have "entered" at some time in the past and sometimes is a working variable.

AlanGriffiths (changes requested):
There are no tests to clarify the intended functionality. This is needed especially when display configurations change.

Also, I'm not sure that the Wayland frontend is the best place for this logic. Calling `track_overlapping()` when the frontend is notified of changes is fragile. For example, if an output is added or reconfigured so that it overlaps the window I would expect an `enter` event immediately not the next time some change of the window happens to call `track_overlapping()`.

I think the `scene` which manages both surfaces and displays would be a better place to `track_overlapping()` and it should notify the frontend of enter and leave events.

include/miroil/miroil/surface_observer.h
```diff
@@ -58,6 +59,8 @@ class SurfaceObserver
     virtual void start_drag_and_drop(mir::scene::Surface const* surf, std::vector<uint8_t> const& handle) = 0;
     virtual void depth_layer_set_to(mir::scene::Surface const* surf, MirDepthLayer depth_layer) = 0;
     virtual void application_id_set_to(mir::scene::Surface const* surf, std::string const& application_id) = 0;
+    virtual void entered_output(mir::scene::Surface const* surf, mir::graphics::DisplayConfigurationOutputId id) = 0;
+    virtual void left_output(mir::scene::Surface const* surf, mir::graphics::DisplayConfigurationOutputId id) = 0;
```
AlanGriffiths, line 63: miroil does not need these notifications

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -468,26 +471,44 @@ void mf::WindowWlSurfaceRole::create_scene_surface()
...
+        output_manager->for_each_output_bound_by(
+            client,
+            [&](OutputInstance* output)
+            {
+                auto const& config{output->global.value().current_config()};
+                if (config.valid() && config.id == id)
+                {
+                    surface.value().send_enter_event(output->resource);
+                }
+            });
```
AlanGriffiths, line 491:
```suggestion
        if (auto const output = output_manager->output_for(id))
        {
            surface.value().send_enter_event((*output)->resource);
        }
```

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -468,26 +471,44 @@ void mf::WindowWlSurfaceRole::create_scene_surface()
...
+        output_manager->for_each_output_bound_by(
+            client,
+            [&](OutputInstance* output)
+            {
+                auto const& config{output->global.value().current_config()};
+                if (config.valid() && config.id == id)
+                {
+                    surface.value().send_leave_event(output->resource);
+                }
+            });
```
AlanGriffiths, line 508:
```suggestion
        if (auto const output = output_manager->output_for(id))
        {
            surface.value().send_leave_event((*output)->resource);
        }
```

src/server/frontend_wayland/output_manager.h
```diff
@@ -104,6 +104,7 @@ class OutputManager
 
     auto output_for(graphics::DisplayConfigurationOutputId id) -> std::optional<OutputGlobal*>;
     auto current_config() -> graphics::DisplayConfiguration const& { return *display_config; }
+    void for_each_output_bound_by(wayland::Client* client, std::function<void(OutputInstance*)> const& functor);
```
AlanGriffiths, line 107: I think this is obsolete

src/server/scene/application_session.h
AlanGriffiths, line 1: Why is the `ApplicationSession` getting involved? So far as I understand the problem only outputs and surfaces are involved.

src/server/scene/application_session.cpp
```diff
@@ -44,6 +46,41 @@ namespace mg = mir::graphics;
...
+    }
+    void placed_relative(Surface const* surf, geometry::Rectangle const& /*placement*/) override
+    {
+        session->track_overlapping_outputs(surf);
+    }
+    ///@}
+private:
+    ApplicationSession* session;
+};
+
```
AlanGriffiths, line 83: We are observing the surface so that we can tell the surface that something changed?

That's weird - the surface already knows something changed, it just needs to decide what notifications are needed.

AlanGriffiths (changes requested):
You already said you know the tests are wrong

include/miroil/miroil/surface_observer.h
hbatagelo: Okay. Notifications removed.

src/server/frontend_wayland/window_wl_surface_role.cpp
hbatagelo: We need `OutputInstance` to access the resource.

src/server/frontend_wayland/window_wl_surface_role.cpp
hbatagelo: We need `OutputInstance` to access the resource.

src/server/frontend_wayland/output_manager.h
hbatagelo: Yes, removed.

src/server/scene/application_session.h
hbatagelo: Fixed with [17dd081](https://github.com/canonical/mir/pull/3244/commits/17dd081983813c68e316aac08f7d93810790d093). The tracking logic is now done in `BasicSurface`.

src/server/scene/application_session.cpp
hbatagelo: Fixed with [17dd081](https://github.com/canonical/mir/pull/3244/commits/17dd081983813c68e316aac08f7d93810790d093).

include/core/mir/int_wrapper.h
```diff
@@ -70,6 +70,12 @@ inline constexpr bool operator < (IntWrapper<Tag,ValueType> const& lhs, IntWrapp
 {
     return lhs.as_value() < rhs.as_value();
 }
+
+template<typename Tag, typename ValueType>
+inline constexpr bool operator > (IntWrapper<Tag,ValueType> const& lhs, IntWrapper<Tag,ValueType> const& rhs)
+{
+    return lhs.as_value() > rhs.as_value();
+}
```
AlanGriffiths, line 78: That does fix an omission, but I prefer to fix by modernising the code (#3266)

hbatagelo: Sure. Much better.

AlanGriffiths (changes requested):
This is progress, but it doesn't yet correctly handle disabling and enabling outputs.

I've set up sidebyside outputs via a `.display` file and can edit the configuration in that file.

Then I open a test app and place it spanning the outputs monitoring the Wayland debug log when I disable and enable one of the displays:

```
$ WAYLAND_DEBUG=client gedit 2>&1 | grep -e "wl_surface.*enter" -e "wl_surface.*leave"
[ 214434.926] wl_surface@22.enter(wl_output@10)
[ 217807.257] wl_surface@22.enter(wl_output@13)
```

Then when I move the window off of and back onto the second output:
```
[ 436896.230] wl_surface@22.leave(wl_output@35)
[ 438376.629] wl_surface@22.enter(wl_output@35)
```
We've not seen the `leave` event for `wl_output@13` when it was disabled nor the `enter` event for `wl_output@35` when it was enabled. (It is the same display, but different `wl_output` objects because they get destroyed and recreated.)

AlanGriffiths:
Also note this needs rebasing on main

src/include/server/mir/observer_multiplexer.h
```diff
@@ -264,28 +268,36 @@ void ObserverMultiplexer<Observer>::register_interest(
...
 }
 
+template<class Observer>
+void ObserverMultiplexer<Observer>::register_early_observer(
+    std::weak_ptr<Observer> const& observer,
+    Executor& executor)
+{
+    std::lock_guard lock{observer_mutex};
+
+    early_observers.emplace_back(std::make_shared<WeakObserver>(observer, executor));
```
AlanGriffiths, line 278: I think there could be a significant simplification (eliminating `early_observers`) by using:

```suggestion
    observers.push_front(std::make_shared<WeakObserver>(observer, executor));
```

(With a note on `observers` that partitioning early observers before other observers is an invariant.)

src/include/server/mir/scene/surface_observer.h
```diff
@@ -62,8 +62,14 @@ class SurfaceObserver
...
-    virtual void entered_output(Surface const* surf, graphics::DisplayConfigurationOutputId const& id) = 0;
-    virtual void left_output(Surface const* surf, graphics::DisplayConfigurationOutputId const& id) = 0;
+    virtual void entered_output(
+        Surface const* surf,
+        graphics::DisplayConfigurationOutputId const& id,
+        std::function<bool(graphics::DisplayConfigurationOutputId const&)> const& hook) = 0;
+    virtual void left_output(
+        Surface const* surf,
+        graphics::DisplayConfigurationOutputId const& id,
+        std::function<bool(graphics::DisplayConfigurationOutputId const&)> const& hook) = 0;
```
AlanGriffiths, line 72: Adding this parameter to this interface looks very odd. Why would a generic Surface observer want to be passed a boolean functor that takes one of the other parameters? And what are the semantics of it?

src/server/scene/basic_surface.cpp
```diff
@@ -988,27 +1004,40 @@ auto mir::scene::BasicSurface::content_top_left(State const& state) const -> geo
...
     display_config->for_each_output(
         [&](mg::DisplayConfigurationOutput const& output)
         {
-            if (output.valid() && output.extents().overlaps(state->surface_rect))
+            if (output.valid() && output.used && output.extents().overlaps(state->surface_rect))
             {
                 if (!tracked_outputs.contains(output.id))
                 {
-                    observers->entered_output(this, output.id);
+                    observers->entered_output(this, output.id, on_enter);
```
AlanGriffiths, line 1035: Here we are passing a lambda to a observer that may run on another thread. But that lambda does nothing to synchronise access `tracked_outputs`.

It also seems odd that we are testing `!tracked_outputs.contains(output.id)` before the call and then, in the lambda repeating the same test before updating `tracked_outputs`.

I feel there must be a cleaner solution to whatever problem this is intended to solve. (What does it solve?)

tests/unit-tests/scene/test_basic_surface.cpp
```diff
@@ -1167,77 +1183,77 @@ TEST_F(BasicSurfaceTest, notifies_of_entered_output_on_move)
...
-    surface.resize({50, 50});
-    surface.move_to({75, 0});
-    display_config_registrar->disconnect_output(1);
-}
+//     surface.register_interest(mock_surface_observer, executor);
+//     surface.show();
+//     surface.resize({50, 50});
+//     surface.move_to({75, 0});
+//     display_config_registrar->disconnect_output(1);
+// }
```
AlanGriffiths: !

AlanGriffiths (changes requested):

src/include/server/mir/observer_multiplexer.h
hbatagelo: Yeah, I've just done that. Will include the note about the invariant.

src/include/server/mir/scene/surface_observer.h
hbatagelo: Additionally, the hook specifically addresses cases where an output is enabled at runtime. Without the hook, the scene surface might assume it has entered an output, but the frontend never emitted the event because the output instance wasn't yet ready when the frontend handler processed it. When an output is dynamically enabled, it may take some time for the new `wl_resource` to bind to the output instance. Only late `entered_output` notifications will catch up with the new output instance.

src/server/scene/basic_surface.cpp
hbatagelo: I think they are all run in the wayland thread, but will check.

The double checks are needed because we may have multiple `entered_output` and `left_output` notifications. For instance, it may happen that, between the time an `entered_output` notification is emitted and handled, other `entered_output` notifications were handled.

tests/unit-tests/scene/test_basic_surface.cpp
hbatagelo: I need to refactor these tests to take into account the hooks, but will leave for later when we decide what is an adequate approach for the tracking problem.

src/server/scene/basic_surface.cpp
hbatagelo: Well, actually, only the checks in the hooks are needed for properly functioning. The checks in the `tracked_outputs` could be removed. However, by doing so, most notifications would be spurious.

hbatagelo: Hmm, I think I'll remove these checks in `tracked_outputs`. They will simplify syncing access to `tracked_outputs` across threads.

src/include/server/mir/observer_multiplexer.h
hbatagelo: Done.

src/server/scene/basic_surface.cpp
hbatagelo: Okay, done: removed the double checks, and added a mutex for tracked_outputs.

src/include/server/mir/scene/surface_observer.h
AlanGriffiths: All of which discussion confirms that responsibilities are not being encapsulated. This interaction is not "Observer" pattern: What happens if there are multiple observers, how do they co-ordinate updating the scene through "hook"?

The scene should notify observers when an enter occurs, it is not be responsible for tracking whether the frontend has actually sent a notification. Likewise the frontend should not be responsible for calling a callback to update the scene.

I take the point about the frontend being unable to send `enter` before the output is bound, but wouldn't that be better handled by recording "pending enter" in the frontend? And sending that if/when the output is bound.

hbatagelo: I agree. I'll revert to the previous tracking approach and start recording the pending enter events in the frontend. Addressing the missing enter events was the only reason to start using hooks.

src/server/scene/basic_surface.cpp
AlanGriffiths: This is all moot - we shouldn't be storing the state of frontend state in scene

src/include/server/mir/scene/surface_observer.h
hbatagelo: Done.

tests/unit-tests/scene/test_basic_surface.cpp
hbatagelo: The tests are back.

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -79,10 +81,23 @@ mf::WindowWlSurfaceRole::WindowWlSurfaceRole(
...
 }
 
 mf::WindowWlSurfaceRole::~WindowWlSurfaceRole()
 {
+    output_manager->current_config().for_each_output(
+        [&](graphics::DisplayConfigurationOutput const& config)
+        {
+            if (auto* global{output_manager->output_for(config.id).value_or(nullptr)})
+            {
+                mir::log_info("\033[1;36mRemoved listener (output %d)\033[0m", config.id.as_value());
```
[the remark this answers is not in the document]
hbatagelo: Yeah, forgot to remove from debugging.

src/server/frontend_wayland/window_wl_surface_role.h
```diff
@@ -97,6 +100,9 @@ class WindowWlSurfaceRole
     void remove_state_now(MirWindowState state);
     void create_scene_surface();
 
+    void handle_enter_output(graphics::DisplayConfigurationOutputId id);
+    void handle_leave_output(graphics::DisplayConfigurationOutputId id) const;
```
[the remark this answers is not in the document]
hbatagelo: Only the "enter" handler needs to update state about pending events. We don't need to track pending leave events.

AlanGriffiths:
> _Not sure why the enter/leave/enter sequence_

That may have been a twitch of the mouse, repeating the move I see a single `enter`

AlanGriffiths:
> **I expect an `enter` here**

And that just might be https://github.com/canonical/mir/pull/3244/commits/64ae007b2a2fba8a67ed19ad8be73b99b2d1589f (which hasn't yet been uploaded to the store)

hbatagelo:
Meanwhile, here's an up-to-date "how it works":

* Output tracking is performed in `BasicSurface::track_outputs()`. This function is called within `BasicSurface` whenever the surface is moved, resized, or a new display configuration is applied. Tracking begins for an output when it is _valid_, _used_, and begins to overlap the surface rectangle. An `entered_output` notification is sent when this occurs, and the output configuration ID is retained for tracking purposes. When the tracked output no longer overlaps the surface rectangle or is no longer valid or in use, a `left_output` notification is sent.
* `WaylandSurfaceObserver` listens for `entered_output`/`left_output` notifications and invokes `handle_enter_output(id)`/`handle_leave_output(id)` in `WindowWlSurfaceRole`.
* The handlers in `WindowSurfaceRole` access the `OutputGlobal` for the given IDs and iterate over the currently bound `OutputInstance` objects to send `wl_surface.enter/leave` events.

Special care is taken when the output configuration changes dynamically:

* When an output is disabled at runtime, `handle_leave_output(id)` may be called too late, after the associated `OutputInstance` objects have already been destroyed by the output manager. To prevent this, `ObserverRegistrar` has been modified to allow an observer to be registered as an "early" observer, i.e., one that is always notified before all other "non-early" observers. This approach is used for the display configuration observer registered in `BasicSurface`. Now, the `configuration_applied` notification is always handled before it is processed by the output manager, ensuring that the output is still valid when the leave notification is handled.
* When an output is enabled at runtime, `handle_enter_output(id)` may be called too soon, before the associated `OutputGlobal` or `OutputInstance` bound to the client have been updated. To address this, the `OutputManager` has been modified to allow someone to listen for the creation of an `OutputGlobal` (`output_global_created` notification). `WaylandSurfaceRole` acts as a listener for this notification. Now, if `handler_enter_output(id)` fails to send the "enter" event, the event is saved as pending. Afterwards, when `output_global_created` is received, `WaylandSurfaceRole` registers itself as a listener for `output_config_changed`, which is notified when the `OutputInstance` is finally bound to the client and the pending event can be sent.

Another case that required special consideration is when `track_outputs()` is called from multiple threads, with one of them being the `Mir/Wayland` thread. Because the `WaylandExecutor` eagerly executes works in the Wayland thread, an event in the Wayland thread can be executed before one that has already been enqueued in another thread. This leads to duplicate "enter" and "leave" events, as observed when resizing a window across side-by-side outputs on a QEMU machine. The problem is resolved by invoking `track_outputs()` through `linearising_executor` when it is called in the Wayland thread and waiting for `track_outputs()` to be executed in the other thread before continuing.

AlanGriffiths:
> Meanwhile, here's an up-to-date "how it works":

Please update the top comment. If anyone looks back at this PR that's what they will read

AlanGriffiths (approved):
Nice!

Issue #3236, opened 2024-02-17
`grimshot save output` is using the exact opposite clip area on my machine


mattkae:
## What I see
![Screenshot from 2024-02-17 14-36-37](https://github.com/canonical/mir/assets/25062299/4c908ca3-0f8b-4dc2-b972-0205700196c2)

## What is outputted from the screenshot tool:
![2024-02-17T14:36:10,794486618-05:00](https://github.com/canonical/mir/assets/25062299/fe1c4a1b-2846-4a4d-8d19-a937d7ea82d0)

PR #3188, opened 2024-01-15
Fix transformation for Layout::TopRowFirst

AlanGriffiths:
Fixes: #3185

PR #3245, opened 2024-02-21
(#3236) `CPUCopyOutputSurface` now used a GL output layout

mattkae:
fixes https://github.com/canonical/mir/issues/3236

Am I right to think this @RAOF ? It looks like `CPUCopyOutputSurface`  is trying to have a `GL` layout. I could be wrong, but this makes my screenshots work.

## Before
2024-02-21T17:01:19,606591722-05:00

## After
2024-02-21T17:06:29,424484886-05:00

### DIFF
--- a/src/renderers/gl/renderer.cpp
+++ b/src/renderers/gl/renderer.cpp
@@ -379,32 +379,30 @@ void mrg::Renderer::draw(mg::Renderable const& renderable) const
 {
     auto const texture = gl_interface->as_texture(renderable.buffer());
     auto const clip_area = renderable.clip_area();
+    glm::mat4 transform = renderable.transformation();
+    if (texture->layout() == mg::gl::Texture::Layout::TopRowFirst)
+    {
+        // GL textures have (0,0) at bottom-left rather than top-left
+        // We have to invert this texture to get it the way up GL expects.
+        transform *= glm::mat4{
+            1.0, 0.0, 0.0, 0.0,
+            0.0, -1.0, 0.0, 0.0,
+            0.0, 0.0, 1.0, 0.0,
+            0.0, 0.0, 0.0, 1.0
+        };
+    }
+
     if (clip_area)
     {
         glEnable(GL_SCISSOR_TEST);
 
-        // NOTE: We are handling the following two cases here:
-        // 1. The layout of the output surface matches that of the texture.
-        //    The clip_area provided has the origin in the top-left corner of the screen, while OpenGL
-        //    has the origin in the bottom-left corner of the screen. Hence, when the layout of the
-        //    output surface matches that of the texture, we need to invert the provided clip_area.
-        // 2. The layout of the output surface DOES NOT match that of the texture.
-        //    Note that the call to glViewport in update_gl_viewport will modify the behavior of glScissor.
-        //    Hence, if the layouts do NOT match, then we do NOT need to invert the clip area. This
-        //    is because we WANT the inverted layout which clip_area provides to us by default.
-        auto const invert_clip_area = (output_surface->layout() == graphics::gl::OutputSurface::Layout::TopRowFirst) ==
-            (texture->layout() == mg::gl::Texture::Layout::TopRowFirst);
-        auto const scissor_y = invert_clip_area ?
-            viewport.top_left.y.as_int() +
-                viewport.size.height.as_int() -
-                clip_area.value().top_left.y.as_int() -
-                clip_area.value().size.height.as_int() :
-            clip_area.value().top_left.y.as_int();
+        glm::vec4 transformed_clip_area =
+            transform * glm::vec4(clip_area.value().top_left.x.as_int(), clip_area.value().top_left.y.as_int(), 0, 0);
 
         glScissor(
-            clip_area.value().top_left.x.as_int() -
+            (int)transformed_clip_area.x -
             viewport.top_left.x.as_int(),
-            scissor_y,
+            (int)transformed_clip_area.y,
             clip_area.value().size.width.as_int(),
             clip_area.value().size.height.as_int()
         );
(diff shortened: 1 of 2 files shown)


RAOF:
Wait, _*what*_?

RAOF:
Hm. I think this might be correct; [glReadPixels says](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glReadPixels.xhtml): 
> return pixel data from the frame buffer, starting with the pixel whose lower left corner is at location (x, y)

However, this breaks CPU output on the gbm-kms platform (as in: flips it), so it's possible that somewhere else has the wrong orientation, too!

mattkae:
> Hm. I think this might be correct; [glReadPixels says](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glReadPixels.xhtml):
> 
> > return pixel data from the frame buffer, starting with the pixel whose lower left corner is at location (x, y)
> 
> However, this breaks CPU output on the gbm-kms platform (as in: flips it), so it's possible that somewhere else has the wrong orientation, too!

You're right! So it's probably wrong for the GL case, but right for the gbm-kms case...

What would be a good way to go about fixing that? I suppose the `BasicScreenShooter` could request a buffer that _always_ has a particular layout? Or maybe it could have a "passthrough" buffer that overrides everything _except_ for the layout? WDYT @RAOF ?

RAOF:
Actually, I suspect that the layout of the `gl::OutputSurface` is not the underlying problem - what's happening is that the clip geometry for (some of!) the surfaces is going wrong - the *rest* of the screenshot (your panel, and background) is fine regardless, yes?

RAOF:
Right, but the fact that the *image as a whole* is the right way up suggests that the `OutputSurface` is writing pixels to the underlying buffer in the right layout? So the Renderer is writing pixels to the buffer in the right layout, it's just that the clip handling is broken?

mattkae:
@RAOF Doing this seems to leave every getting clipped when it is uneven in the Y-axis
```
if (clip_area)
    {
        glEnable(GL_SCISSOR_TEST);

        glm::vec4 transformed_clip_area =
            transform * glm::vec4(clip_area.value().top_left.x.as_int(), clip_area.value().top_left.y.as_int(), 0, 0);

        glScissor(
            (int)transformed_clip_area.x -
            viewport.top_left.x.as_int(),
            (int)transformed_clip_area.y,
            clip_area.value().size.width.as_int(),
            clip_area.value().size.height.as_int()
        );
    }
```

src/renderers/gl/renderer.cpp
```diff
@@ -379,32 +379,30 @@ void mrg::Renderer::draw(mg::Renderable const& renderable) const
...
-        auto const invert_clip_area = (output_surface->layout() == graphics::gl::OutputSurface::Layout::TopRowFirst) ==
-            (texture->layout() == mg::gl::Texture::Layout::TopRowFirst);
-        auto const scissor_y = invert_clip_area ?
-            viewport.top_left.y.as_int() +
-                viewport.size.height.as_int() -
-                clip_area.value().top_left.y.as_int() -
-                clip_area.value().size.height.as_int() :
-            clip_area.value().top_left.y.as_int();
+        glm::vec4 transformed_clip_area =
+            transform * glm::vec4(clip_area.value().top_left.x.as_int(), clip_area.value().top_left.y.as_int(), 0, 0);
```
mattkae, line 400: This is wrong, but _why_ it is wrong, I do not know at the moment

RAOF: My *guess* is that this is wrong because you're not using [homogeneous coördinates](https://en.wikipedia.org/wiki/Homogeneous_coordinates) (the `w` component should be `1.0` rather than `0.0`, as this represents a point rather than a direction)

mattkae: True :wink: That didn't fix it though, going to mess around with it a bit more

src/renderers/gl/renderer.cpp
```diff
@@ -379,32 +379,30 @@ void mrg::Renderer::draw(mg::Renderable const& renderable) const
...
+    if (texture->layout() == mg::gl::Texture::Layout::TopRowFirst)
+    {
+        // GL textures have (0,0) at bottom-left rather than top-left
+        // We have to invert this texture to get it the way up GL expects.
+        transform *= glm::mat4{
+            1.0, 0.0, 0.0, 0.0,
+            0.0, -1.0, 0.0, 0.0,
+            0.0, 0.0, 1.0, 0.0,
+            0.0, 0.0, 0.0, 1.0
+        };
```
AlanGriffiths, line 392: Drive by question: do we really want to invert before the transform? And not after?

I've not read the surrounding code to check, just been triggered because the snippet looks like #3188

mattkae:
@RAOF I have made sense of all things using my brain. Here is what we need to do:

1. Make it so that the Y-coordinate is relative to the bottom left of the screen (the coordinate comes to us relative to the top left of the screen)
2. Transform the clip area in the same way that we're transforming the display, since this is the fate of that buffer's pixels on the screen as well

We don't need to take into account the buffer's transform (which should be renamed to `model` IMO) since this is a local transform on the model.

src/renderers/gl/renderer.cpp
mattkae: I fixed this PR so that we no longer use it, so this would be unrelated. Buuuut, I wonder as well! This `transform` is actually the `model` matrix for the buffer. I _imagine_ that we would want to invert _and then_ apply this transformation.

For example: If we rotate the buffer by 45 degrees, we will want to rotate the _inverted buffer_ instead of the pre-inverted buffer. In the latter case, we would have a resulting buffer that is rotated 135 degrees, which would be incorrect.

mattkae: This is probably a fix for another pull request though. And I would at least have to confirm my thoughts before throwing it in...

mattkae:
And voila, a perfect screenshot!
meow

AlanGriffiths:
> (I wish I understood what led to the previous logic - that would give me more confidence I'm not missing something)

Having revised the discussion on #3204, I now see where we went wrong. (Essentially, by hand coded special cases for the display transformation.)

Issue #3238, opened 2024-02-19
Crash on wake when outputs have the same geometry


AlanGriffiths:
On my X1C1 laptop, with external screen attached (which, by chance, has the same geometry)

1. login to Miriway (tried edge, beta and edge/mir-pr3237)
5. Wait until outputs are switched off
6. Use touchpad, keyboard or mouse to wake system

Expect: desktop session wakes
Actual: Momentary display of desktop followed by return to greeter

_I do know that I've not seen this crash in many other scenarios_

AlanGriffiths:
OK, so here's the failure mode:
```
[2024-02-19 14:28:09.038892] < CRITICAL! > mirserver: terminate_with_current_exception(): ./src/platforms/gbm-kms/server/kms/kms_page_flipper.cpp(69): Throw in function virtual bool mir::graphics::gbm::KMSPageFlipper::schedule_flip(uint32_t, uint32_t, uint32_t)
Dynamic exception type: boost::wrapexcept<std::logic_error>
std::exception::what: Page flip for crtc_id is already scheduled

[2024-02-19 14:28:09.039049] < - debug - > mirserver: Handling Terminated from pid=1848353
```

AlanGriffiths:
So, it looks like the necessary condition is for both outputs to be sharing a frame buffer. (On the same card and of the same geometry?)

And, to confirm, with `display-config=sidebyside` there's no problem.

AlanGriffiths:
Further clarification: the display configuration was "static" but placing both outputs with the same geometry

AlanGriffiths:
A bit more information: the two outputs are (correctly) in a group, and the first is displayed correctly. It is when a page flip is requested for the second output that a problem is detected.

AlanGriffiths:
...and that is because, after the suspend, both outputs have the same `crtc_id`.

Before the suspend they are:
```
[2024-02-20 16:28:56.745592] < - debug - > gbm-kms: **** mgg::KMSPageFlipper::schedule_flip(60, 100, 83)
[2024-02-20 16:28:56.745816] < - debug - > gbm-kms: **** mgg::KMSPageFlipper::schedule_flip(45, 100, 76)
```

After the suspend:
```
[2024-02-20 16:30:49.359230] < - debug - > gbm-kms: **** mgg::KMSPageFlipper::schedule_flip(45, 104, 83)
[...]

PR #3247, opened 2024-02-22
Don't use the same crtc for multiple outputs in a group

AlanGriffiths:
Fixes: #3238

### DIFF
--- a/src/platforms/common/server/kms-utils/kms_connector.cpp
+++ b/src/platforms/common/server/kms-utils/kms_connector.cpp
@@ -19,6 +19,7 @@
 
 #include <boost/throw_exception.hpp>
 #include <algorithm>
+#include <list>
 #include <stdexcept>
 
 namespace mgk = mir::graphics::kms;
@@ -98,7 +99,6 @@ std::string mgk::connector_name(mgk::DRMModeConnectorUPtr const& connector)
     name += std::to_string(connector->connector_type_id);
     return name;
 }
-
 namespace
 {
 std::tuple<mgk::DRMModeCrtcUPtr, int> find_crtc_and_index_for_connector(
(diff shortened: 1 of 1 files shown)


src/platforms/common/server/kms-utils/kms_connector.cpp
```diff
@@ -117,13 +130,34 @@ std::tuple<mgk::DRMModeCrtcUPtr, int> find_crtc_and_index_for_connector(
...
 
+    // We couldn't find a crtc we haven't used, so return the least recently returned one
+    for (auto pcrtc_id = recent_crtc_ids.begin(); pcrtc_id != recent_crtc_ids.end(); ++pcrtc_id)
+    {
+        for (auto& skipped_crtc : skipped_crtcs)
+        {
+            if (*pcrtc_id == get<0>(skipped_crtc)->crtc_id)
+            {
+                recent_crtc_ids.erase(pcrtc_id);
+                recent_crtc_ids.push_back(get<0>(skipped_crtc)->crtc_id);
```
mattkae, line 155: Should we also remove this from `skipped_crtcs`?

src/platforms/common/server/kms-utils/kms_connector.cpp
```diff
@@ -109,6 +109,19 @@ std::tuple<mgk::DRMModeCrtcUPtr, int> find_crtc_and_index_for_connector(
...
+    // We don't "seize" crtcs when we find them which means they remain "available". So there is
+    // an opportunity for the client logic to request several crtcs before using the first makes it
+    // "unavailable" and get the same crtc for multiple requests.
+    //
+    // By ensuring we don't return the same crtc until we've exhausted all alternatives we prevent
+    // the client logic using it twice.
+    //
+    // Using "thread_local" as each group of outputs is accessed on its own thread.
+    thread_local std::list<u_int32_t> recent_crtc_ids;
+    std::list<std::tuple<mgk::DRMModeCrtcUPtr, int>> skipped_crtcs;
```
mattkae, line 123: Why isn't `skipped_crtcs` also `thread_local`?

mattkae (changes requested):
A few questions!

src/platforms/common/server/kms-utils/kms_connector.cpp
AlanGriffiths: Because it doesn't have static duration

src/platforms/common/server/kms-utils/kms_connector.cpp
AlanGriffiths: No

AlanGriffiths:
Not needed (yet). C.f. #3248

PR #3248, opened 2024-02-22
Ignore zero length gamma curves

AlanGriffiths:
After a wake we might see an invalid Gamma curve.  Ignore it.

Fixes: #3238

It fixes the problem because `ensure_crtc()` might allocate a new `current_crtc` but the `drmModeCrtcSetGamma()` will fail leaving it unused - but still associated with the output.

### DIFF
--- a/src/platforms/gbm-kms/server/kms/real_kms_output.cpp
+++ b/src/platforms/gbm-kms/server/kms/real_kms_output.cpp
@@ -360,6 +360,12 @@ void mgg::RealKMSOutput::set_power_mode(MirPowerMode mode)
 
 void mgg::RealKMSOutput::set_gamma(mg::GammaCurves const& gamma)
 {
+    if (!gamma.red.size())
+    {
+        mir::log_warning("Ignoring attempt to set zero length gamma");
+        return;
+    }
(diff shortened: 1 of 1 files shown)


RAOF (approved):
This change certainly isn't *wrong*, but why are *we* attempting to set a zero-sized gamma curve? What's trying to set this further up the stack?

AlanGriffiths:
> Will have a further look, but IMO this mitigation is good anyway

The **only** place gamma gets set is `mgg::RealKMSOutput::update_from_hardware_state()` and there it is set by this dubious path:

```c++
    GammaCurves gamma;

    /* Get information about the current mode */
    if (current_crtc) {
        current_mode_info = current_crtc->mode;

        if (current_crtc->gamma_size > 0)
            gamma = mg::LinearGammaLUTs(current_crtc->gamma_size);
    }
...
    output.gamma = gamma;
```

The upshot of which is, that if there no `current_crtc` then it is set to the default constructed `GammaCurves gamma`. (Even if it had been set previously - which I doubt.)

And, empirically, the usual way to reach this code is with `current_crtc` false.

PR #3251, opened 2024-02-23
Revert "Merge pull request #3243 from canonical/feature/3241"

RAOF:
This reverts commit 8346ca9ac88341ef81fb0294f205504dd434368f, reversing changes made to 120d2ccfb744d6f43fd32df806936e6b92c4b3a8.

It turns out that sometimes the userdata is set to nullptr, leading `me->destroy()` to attempt to jump to null.

### DIFF
--- a/src/wayland/generator/request.cpp
+++ b/src/wayland/generator/request.cpp
@@ -26,7 +26,7 @@ Emitter Request::virtual_mir_prototype() const
 {
     if (is_destroy())
     {
-        return {"virtual void ", name, "(", mir_args(), ") {}"};
+        return nullptr;
     }
     else
     {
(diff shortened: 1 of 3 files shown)


RAOF:
It's not clear to me how this can happen, but `wl_output.release()` can be called with the userdata for `resource` set to `nullptr`. This makes for a quick crash on startup, I think when the splash screen exits.

AlanGriffiths (approved):
Sure, I trust you! (I didn't review the original, but you did.)

mattkae:
@RAOF Spooky! I'll check it out now

mattkae:
It _must_ be that `resource_destroyed_thunk` was run somehow beforehand...

mattkae:
I cannot reproduce this error at all, but I _guess_ it could validly happen? Here's how I can see it happening (although this is a guess):

1. The compositor starts running
2. The configuration is processed
3. A new configuration happens, so:
4. The Output gets deleted
5. In the meantime, a client tried to `release()` the resource, but it doesn't exist

mattkae:
I have no reliable repro for this. I'll chat with RAOF on Monday to try and dig into it what he's seeing on his end. I tried everything with `miral-app` and nothing makes it happen for me

RAOF:
I cannot seem to reproduce this today?! It *was* 100% reproducible; just starting up `miral-shell` either under X11 or on a VT would cause it to crash shortly after startup and was reproducible under gdb with a backtrace with the second-to-top stack frame being in `wayland::Output::release()` with `me == nullptr`.

I couldn't see *how* that could happen, but it definitely was happening on Friday.

Did I just have a bad build?

mattkae:
I haven't seen this once, so If I have the blessing to revert the revert, then I will merge it. I can always add a null-check on `me` if we think that we might hit it again...

PR #3254, opened 2024-02-23
Revert "Revert "Merge pull request #3243 from canonical/feature/3241""

mattkae:
This reverts commit e351efa6117d22f476d5652a2d943ccd834ad460.

### DIFF
--- a/src/wayland/generator/request.cpp
+++ b/src/wayland/generator/request.cpp
@@ -26,7 +26,7 @@ Emitter Request::virtual_mir_prototype() const
 {
     if (is_destroy())
     {
-        return nullptr;
+        return {"virtual void ", name, "(", mir_args(), ") {}"};
     }
     else
     {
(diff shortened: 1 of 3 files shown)


mattkae:
Working to fix the issues found here: https://github.com/canonical/mir/pull/3251

RAOF:
As I commented [here](https://github.com/canonical/mir/pull/3251#issuecomment-1963184714) I can't reproduce the crash anymore, so I'm not sure what's happening here.

mattkae:
@RAOF Would it be a good idea to null-check `me` in this case, if it's somewhat of a leprechaun? I can't see _why_ it would happen, but if it _could_ happen (somehow) we could at least not blow up. Maybe log an error?

Saviq (approved):
> @RAOF Would it be a good idea to null-check `me` in this case, if it's somewhat of a leprechaun? I can't see _why_ it would happen, but if it _could_ happen (somehow) we could at least not blow up. Maybe log an error?

Or better to blow up, if something as unexpected as this happens? A WLCS test?

Will merge, we'll get feedback from a bunch of runs in the lab.

PR #3255, opened 2024-02-26
spread: deal with usrmerge issues (LP#2054925)

Saviq:


Saviq:
> Weird. But seems to be what is recommended

Was hoping it will help with `debian_sid`, too - as that seems failing in a related way… alas, no.

PR #3258, opened 2024-02-29
spread: sid needs merged

Saviq:


RAOF:
I think we need to wait for Sid to complete the 64bit `time_t` transition before we can expect anything at all to work there.

(This seems reasonable to me in isolation, but we're not going to be able to test Sid for a little while still)

Issue #3261, opened 2024-03-04
Segfault in `_dl_lookup_symbol_x` for `describe_graphics_module`


Saviq:
```
$ snap refresh mir-test-tools --channel 22/edge
$ env -u WAYLAND_DISPLAY mir-test-tools.demo-server
[2024-03-04 13:57:38.238983] <information> mirserver: Starting
[2024-03-04 13:57:38.256499] < - debug - > mirserver: Not trying logind: "DISPLAY" is set and X need not have claimed the VT
[2024-03-04 13:57:38.261864] < - debug - > mirserver: Not using Linux VT subsystem for session management: Failed to find the current VT
[2024-03-04 13:57:38.261904] < - debug - > mirserver: No session management supported
[2024-03-04 13:57:38.261928] <information> VT switch key handler: No VT switching support available: MinimalConsoleServices does not support VT switching
[2024-03-04 13:57:38.262075] <information> mircommon: Loading modules from: /snap/mir-test-tools/14320/usr/lib/x86_64-linux-gnu/mir/server-platform
[2024-03-04 13:57:38.262133] <information> mircommon: Loading module: /snap/mir-test-tools/14320/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-eglstream-kms.so.22
[2024-03-04 13:57:38.262159] <information> mircommon: Loading module: /snap/mir-test-tools/14320/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-gbm-kms.so.22
[2024-03-04 13:57:38.262172] <information> mircommon: Loading module: /snap/mir-test-tools/14320/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-wayland.so.22
[2024-03-04 13:57:38.262184] <information> mircommon: Loading module: /snap/mir-test-tools/14320/usr/lib/x86_64-linux-gnu/mir/server-platform/renderer-egl-generic.so.22
[2024-03-04 13:57:38.262194] <information> mircommon: Loading module: /snap/mir-test-tools/14320/usr/lib/x86_64-linux-gnu/mir/server-platform/server-x11.so.22
[2024-03-04 13:57:38.262204] <information> mircommon: Loading module: /snap/mir-test-tools/14320/usr/lib/x86_64-linux-gnu/mir/server-platform/input-evdev.so.9
[2024-03-04 13:57:38.263421] <information> mirserver: Found display driver: mir:x11 (version 2.16.3)
[...]

PR #3263, opened 2024-03-04
Do not assume there is a "mir:virtual" platform

AlanGriffiths:
Fixes: #3261

Note: this is a "quick and dirty" fix to unbreak main.

A fix with tests needs a rework of `mg::select_display_modules()` so that it doesn't both load the available platform modules and select amongst them. That will need a little more time.


Saviq (approved):
Thanks!

PR #3264, opened 2024-03-04
bugfix: adding a quirk to the DesktopFileManager's resolution of gnome-terminal-server

mattkae:
## What's new?
- If we encounter `gnome-terminal-server`, then we always return `org.gnome.Terminal.desktop`


src/server/frontend_wayland/desktop_file_manager.cpp
```diff
@@ -55,6 +59,9 @@ std::string mf::DesktopFileManager::resolve_app_id(const scene::Surface* surface
     if (app)
         return app->id;
 
+    if (app_id_to_desktop_file_quirks.contains(app_id))
+        return app_id_to_desktop_file_quirks.at(app_id);
+
```
AlanGriffiths, line 64: Doing two lookups seems spuriously inefficient.

```suggestion
    for (auto const& q: app_id_to_desktop_file_quirks)
        if (q.first == app_id) return q.second;
```

src/server/frontend_wayland/desktop_file_manager.cpp
```diff
@@ -27,6 +27,10 @@ namespace mf = mir::frontend;
 namespace
 {
 const char* DESKTOP_FILE_POSTFIX = ".desktop";
+
+const std::map<std::string, std::string> app_id_to_desktop_file_quirks = {
```
AlanGriffiths, line 31: We use east-const around here. (I note the line above doesn't, nor does it follow our naming style - but that's pre-existing.)

Also, I doubt that we will ever have enough quirks to justify a map. (Also a shame it can't be `constexpr`!)

I think
```suggestion
std::pair<std::string, std::string> const app_id_to_desktop_file_quirks[] = {
```
would be cleaner and more efficient.

AlanGriffiths (changes requested):
Good idea, but implementation pulls in needless complexity

AlanGriffiths (approved):
WFM, but I'll let @Saviq vote.

> It _could also be_ that the shell who is responsible? But I find it nice if the resolution is consistent across components

Agreed: this is a sensible place. This workaround sanitises "garbage in" at the point it enters our control.

Saviq (approved):
**FINE**.

PR #3275, opened 2024-03-07
Rewrite some MOCK_METHOD[0-9]\( -> MOCK_METHOD(

AlanGriffiths:
(Not the most interesting hobby, but filled some gaps.)


RAOF (approved):
Ah, neat! Some of these I already have pending in a local branch, but good to get them all(?) out of the way in one go!

AlanGriffiths:
> but good to get them all(?) out of the way in one go!

This is not all of them, it is less than half. (700 ==> 400, not counting 80 `MOCK_CONST_METHOD[0-9]\(`)

PR #3276, opened 2024-03-07
feature: improving logging for lockscreen related code

mattkae:


src/server/console/logind_console_services.cpp
```diff
@@ -844,6 +845,7 @@ void mir::LogindConsoleServices::request_lock(
     GObject*,
     gpointer ctx) noexcept
 {
+    mir::log_info("Received Lock() signal for this session from logind");
```
RAOF, line 848: Here's an example of one that I think is a good `log_info`

src/server/console/logind_console_services.cpp
```diff
@@ -934,10 +937,12 @@ mir::LogindConsoleServices::~LogindConsoleServices()
 
 void mir::LogindConsoleServices::on_lock()
 {
+    mir::log_info("Setting the locked hint for the session to true");
```
RAOF, line 940: Also a good `log_info`, although:
```suggestion
    mir::log_info("Notifying logind of session lock");
```

src/server/console/logind_console_services.cpp
```diff
@@ -934,10 +937,12 @@ mir::LogindConsoleServices::~LogindConsoleServices()
...
     logind_session_set_locked_hint(session_proxy.get(), true);
 }
 
 void mir::LogindConsoleServices::on_unlock()
 {
+    mir::log_info("Setting the locked hint for the session to false");
```
RAOF, line 946: ...and
```suggestion
    mir::log_info("Notifying logind of session unlock");
```

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -227,9 +234,11 @@ void mf::SessionLockV1::destroy()
     if (manager.is_active_lock(this))
     {
         manager.try_relinquish_locking_privilege(this);
+        mir::log_info("SessionLockV1 successfully destroyed itself");
     }
     else
     {
+        mir::log_info("SessionLockV1 attempted to destroy itself when it was not the lock");
```
RAOF, line 241: Hm. I wonder if this would be better served by logging *from the wrapper generator* when a protocol error is caught and sent?

src/server/scene/surface_stack.cpp
```diff
@@ -609,19 +610,29 @@ void ms::SurfaceStack::lock()
     bool temp = false;
     if (is_locked.compare_exchange_weak(temp, true))
     {
+        mir::log_info("Session is now locked and listeners are being notified");
```
RAOF, line 613: This `log_info` is good...

src/server/scene/surface_stack.cpp
```diff
@@ -609,19 +610,29 @@ void ms::SurfaceStack::lock()
...
         emit_scene_changed();
         multiplexer.on_lock();
     }
+    else
+    {
+        mir::log_info("SurfaceStack received lock() but could not compare_exchange_weak");
```
RAOF, line 619: ...this `log_info` is bad - although that's partially because this should be `compare_exchange_strong`. This *should* be something like:
```suggestion
        mir::log_debug("Session received duplicate lock request");
```

RAOF (changes requested):
A lot of these `mir::log_info`s seem to me like they should be `mir::log_debug`s.

I feel that `log_info` should be at least vaguely user-understandable. Some of these are, some are deep in our-specific-implementation weeds.

src/server/scene/surface_stack.cpp
```diff
@@ -609,19 +610,29 @@ void ms::SurfaceStack::lock()
...
+    }
 }
 
 void ms::SurfaceStack::unlock()
 {
     bool temp = true;
     if (is_locked.compare_exchange_weak(temp, false))
```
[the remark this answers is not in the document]
AlanGriffiths: +1

src/server/frontend_wayland/session_lock_v1.cpp
mattkae: Eh I kinda like it being optional? Maybe it's not interesting in some contexts?

AlanGriffiths: In what contexts do you imagine a protocol error would be uninteresting AND reported by an exception?

It is useful to log a protocol error when it is ignored, but not when it is also reported via an exception.

AlanGriffiths (changes requested):
"Info" messages should be helpful to users of Mir trying to understand the interaction with other software. E.g. 

* "SessionLockV1 ..." is internal terminology that someone running Mir is unlikely to understand. That's not "info".

* "Received Unlock() signal for this session from logind" is interaction with things a user could be interested in. That's "info" (not "debug")

"Debug" messages are for a different audience: developers of Mir, they can be expected to know what SessionLockV1 is.

src/server/frontend_wayland/session_lock_v1.cpp
mattkae: That's fair to me if we think that's useful info. Should this be a separate ticket though to modify the generator to include logging?

mattkae: So basically we don't need the logs here, since we'll get the exception info anyway. Right?

src/server/frontend_wayland/session_lock_v1.cpp
```diff
@@ -241,6 +250,7 @@ void mf::SessionLockV1::unlock_and_destroy()
 {
     if (!manager.try_unlock(this))
     {
+        mir::log_info("SessionLockV1 attempted to unlock_and_destroy but it is not allowed to");
         BOOST_THROW_EXCEPTION(mw::ProtocolError(
             resource,
             Error::invalid_unlock,
```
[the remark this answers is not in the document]
mattkae: This would happen in the generator, right?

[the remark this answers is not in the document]
AlanGriffiths: I've not looked. But, in general terms, if we throw an exception there should somewhere we catch it. And the catch site has more context about what was intended to happen and how to handle the failure (including what needs logging)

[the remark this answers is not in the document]
mattkae: Yeah we `catch(ProtocolError const& err)` everywhere. I can either make a ticket to add a log to these statements in the generator code OR `WAYLAND_DEBUG=1` would just give you that info anyway, since we forward all of those errors with `wl_resource_post_error(...)`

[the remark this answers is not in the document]
AlanGriffiths: As you imply, reporting the error with both "post_error" and logging sounds redundant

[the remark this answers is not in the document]
mattkae: Cool, then I'll leave it for now :+1:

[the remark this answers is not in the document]
RAOF: > As you imply, reporting the error with both "post_error" and logging sounds redundant

The reason we may want to log something as well as `wl_resource_post_error` is that `post_error` is otherwise silent in the *Mir* log (it's noisy on the cilent end; the client will be killed and will by default print the error we sent it).

“We deliberately killed a client because it behaved incorrectly” seems likely to be both infrequent and of interest to users when it happens.

src/server/console/logind_console_services.cpp
```diff
@@ -844,6 +845,7 @@ void mir::LogindConsoleServices::request_lock(
     GObject*,
     gpointer ctx) noexcept
 {
+    mir::log_debug("Received Lock() signal for this session from logind");
```
RAOF, line 848: This probably wants to be `log_info`; this is the interaction of Mir with another system component, something that a user is plausibly going to know about and be interested in.
```suggestion
    mir::log_info("Received Lock() signal for this session from logind");
```

src/server/console/logind_console_services.cpp
```diff
@@ -852,6 +854,7 @@ void mir::LogindConsoleServices::request_unlock(
     GObject*,
     gpointer ctx) noexcept
 {
+    mir::log_debug("Received Unlock() signal for this session from logind");
```
RAOF, line 857: Also `log_info`
```suggestion
    mir::log_info("Received Unlock() signal for this session from logind");
```

src/server/console/logind_console_services.cpp
```diff
@@ -934,10 +937,12 @@ mir::LogindConsoleServices::~LogindConsoleServices()
 
 void mir::LogindConsoleServices::on_lock()
 {
+    mir::log_debug("Notifying logind of session lock");
```
RAOF, line 940: Also `log_info`
```suggestion
    mir::log_info("Notifying logind of session lock");
```

src/server/console/logind_console_services.cpp
```diff
@@ -934,10 +937,12 @@ mir::LogindConsoleServices::~LogindConsoleServices()
...
     logind_session_set_locked_hint(session_proxy.get(), true);
 }
 
 void mir::LogindConsoleServices::on_unlock()
 {
+    mir::log_debug("Notifying logind of session unlock");
```
RAOF, line 946: Also `log_info`
```suggestion
    mir::log_info("Notifying logind of session unlock");
```

RAOF (changes requested):
Final(!) nitpicks!

AlanGriffiths (approved):

PR #3277, opened 2024-03-08
platform/linux-dmabuf: Be more lenient with EGL implementations

RAOF:
Some EGL implementations (*cough* NVIDIA) return formats from `eglQueryDmaBufFormatsEXT` that generate
`EGL_BAD_PARAMETER` errors when querying for the associated modifiers with `eglQueryDmaBufModifiersEXT`.

This doesn't need to be fatal, as long as there are *some* formats for which we can successfully query
the modifiers, so make it non-fatal.

Fixes: #3278


RAOF:
Closed in favour of #3279 which should have the magical branch name necessary to automatically close Jira issues?

PR #3279, opened 2024-03-08
platform/linux-dmabuf: Be more lenient with EGL implementations

RAOF:
Some EGL implementations (*cough* NVIDIA) return formats from `eglQueryDmaBufFormatsEXT` that generate
`EGL_BAD_PARAMETER` errors when querying for the associated modifiers with `eglQueryDmaBufModifiersEXT`.

This doesn't need to be fatal, as long as there are *some* formats for which we can successfully query
the modifiers, so make it non-fatal.

Fixes: #3278

### DIFF
--- a/src/platform/graphics/linux_dmabuf.cpp
+++ b/src/platform/graphics/linux_dmabuf.cpp
@@ -101,7 +101,18 @@ public:
                     nullptr,
                     &num_modifiers) != EGL_TRUE)
             {
-                BOOST_THROW_EXCEPTION((mg::egl_error("Failed to query number of modifiers")));
+                mir::log_warning("eglQueryDmaBufModifiers failed for format %s: %s",
+                    mg::drm_format_to_string(static_cast<uint32_t>(format)),
+                    mg::egl_category().message(eglGetError()).c_str());
+
+                // Remove that format and its modifiers from our list
+                formats.erase(formats.begin() + i);
+                modifiers_for_format.erase(modifiers_for_format.begin() + i);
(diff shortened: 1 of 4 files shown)


src/platform/graphics/linux_dmabuf.cpp
```diff
@@ -101,7 +101,18 @@ class mg::DmaBufFormatDescriptors
...
+                    mg::drm_format_to_string(static_cast<uint32_t>(format)),
+                    mg::egl_category().message(eglGetError()).c_str());
+
+                // Remove that format and its modifiers from our list
+                formats.erase(formats.begin() + i);
+                modifiers_for_format.erase(modifiers_for_format.begin() + i);
+                external_only_for_format.erase(external_only_for_format.begin() + i);
+                // formats[i] is now the format *after* the one we've just removed, so go back one
+                // and continue the loop.
+                --i;
```
AlanGriffiths, line 114: It is a bit fragile and ugly relying on underflow being defined for unsigned types. I think I'd go for an empty _iteration-expression_ and an increment at the end of the loop.

AlanGriffiths (approved):
I don't _like_ for loops that mess with the iteration variable, so I'll approve while allowing discussion before merging

src/platform/graphics/linux_dmabuf.cpp
RAOF: Yeah, fair enough. Done!

AlanGriffiths (approved):

PR #3280, opened 2024-03-08
compare_exchange_strong is needed to avoid spurious failures

AlanGriffiths:


mattkae (approved):
That makes sense to me... This is probably breaking something. Not sure why I thought to use weak there without a loop.

PR #3282, opened 2024-03-08
feature: support for `zwlr_layer_surface_v1::set_exclusive_zone(-1)`

### LINKED ISSUE #1992: Layer shell: respect -1 exclusive zone

mattkae:
fixes #1992 (and maybe #3125?)

## What's new
- If we have an attached window with `ignore_exclusion_zones` set to `true`, then the surface will attach to its edges without regard for other exclusion zones
- When we're locked, we shouldn't calculate exclusion zones that are _not_ visible on the lockscreen

## To Test
- Run `waybar`
- Run `swaybg -i /path/to/jpg`
- Note that `swaybg` will appear _underneath_ `waybar`, as it should!

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -2883,16 +2883,16 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
         }
     }
 
-    for (auto& area : display_areas)
+    for (auto const& area : display_areas)
     {
         Rectangle zone_rect = area->area;
 
-        // The placement algorithm operates on windows spanning the width of the screen first,
-        // followed by any other "attached" window (e.g. a window attached to the side of the screen
-        // vertically), and finally maximized windows. This ensures that windows attached to the
-        // top and bottom of the screen (in the "full_width_attached_windows" list) push windows
-        // attached to the sides of the screen or elsewhere (in the "other_attached_windows") out
-        // of the way. Maximized windows are then guaranteed to be on top, as they appear last.
+        // The placement order is as follows:
+        //  1. First, attached windows that ignore exclusion zones
+        //  2. Next, windows spanning the screen horizontally
+        //  3. Then, windows spanning the screen vertically
+        //  4. Finally, other maximized windows since they respect all exclusion zones
+        std::vector<WindowInfo*> attached_windows_that_ignore_exclusion_zones;
         std::vector<WindowInfo*> full_width_attached_windows;
         std::vector<WindowInfo*> other_attached_windows;
         std::vector<WindowInfo*> maximized_windows;
@@ -2906,7 +2906,12 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
                 switch (info.state())
                 {
                 case mir_window_state_attached:
-                    if (info.exclusive_rect().is_set())
+                    if (info.ignore_exclusion_zones())
+                    {
+                        attached_windows_that_ignore_exclusion_zones.push_back(&info);
+                        break;
+                    }
+                    else if (info.exclusive_rect().is_set())
                     {
                         MirPlacementGravity  edges = info.attached_edges();
                         bool is_full_width = (edges & mir_placement_gravity_east) && (edges & mir_placement_gravity_west);
@@ -2928,6 +2933,11 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
                     break;
                 }
             }
+        }
+
+        for (auto info_ptr : attached_windows_that_ignore_exclusion_zones)
+        {
+            place_attached_to_zone(*info_ptr, zone_rect);
         }
 
         for (auto info_ptr : full_width_attached_windows)
--- a/src/server/frontend_wayland/layer_shell_v1.cpp
+++ b/src/server/frontend_wayland/layer_shell_v1.cpp
@@ -196,7 +196,7 @@ private:
         wl_resource_destroy(resource);
     }
 
-    DoubleBuffered<uint32_t> exclusive_zone{0};
+    DoubleBuffered<int32_t> exclusive_zone{0};
     DoubleBuffered<Anchors> anchors;
     DoubleBuffered<Margin> margin;
     bool width_set_by_client{false}; ///< If the client gave a width on the last .set_size() request
@@ -434,6 +434,13 @@ auto mf::LayerSurfaceV1::inform_window_role_of_pending_placement()
     spec.exclusive_rect = exclusive_rect ?
         mir::optional_value<geom::Rectangle>(exclusive_rect.value()) :
         mir::optional_value<geom::Rectangle>();
+
+    // Per the spec: https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:set_exclusive_zone
+    // If the zone is -1, it indicates that the surface will not be moved to respect the exclusion zones of other
+    // surfaces.
+    auto zone = exclusive_zone.pending();
+    if (zone == -1)
+        spec.ignore_exclusion_zones = true;
 
     apply_spec(spec);
 }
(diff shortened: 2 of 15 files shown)


mattkae:
Keep in mind that this WLCS ticket (https://github.com/canonical/wlcs/issues/332) may be a good followup, but I will keep them as distinct tasks for now

src/miral/basic_window_manager.cpp
```diff
@@ -89,25 +91,53 @@ miral::BasicWindowManager::Locker::Locker(BasicWindowManager* self) :
...
+    std::shared_ptr<mir::frontend::SurfaceStack> const& surface_stack,
+    std::shared_ptr<mir::scene::SessionLock> const& session_lock) :
     focus_controller(focus_controller),
     display_layout(display_layout),
     persistent_surface_store{persistent_surface_store},
+    surface_stack{surface_stack},
+    session_lock{session_lock},
+    lock_listener{std::make_shared<LockListener>(
+        [this]() { update_application_zones_and_attached_windows(); },
+        [this]() { update_application_zones_and_attached_windows(); }
```
AlanGriffiths, line 127: Maybe I'm missing something, but why does (un)locking the session require an update?

src/miral/basic_window_manager.cpp
```diff
@@ -89,25 +91,53 @@ miral::BasicWindowManager::Locker::Locker(BasicWindowManager* self) :
...
+}
+
 miral::BasicWindowManager::BasicWindowManager(
     shell::FocusController* focus_controller,
     std::shared_ptr<shell::DisplayLayout> const& display_layout,
     std::shared_ptr<mir::shell::PersistentSurfaceStore> const& persistent_surface_store,
     mir::ObserverRegistrar<mir::graphics::DisplayConfigurationObserver>& display_configuration_observers,
-    WindowManagementPolicyBuilder const& build) :
+    WindowManagementPolicyBuilder const& build,
+    std::shared_ptr<mir::frontend::SurfaceStack> const& surface_stack,
```
AlanGriffiths, line 118: `miral` code shouldn't be referencing types from `frontend`, that leads to unnecessary coupling

src/miral/basic_window_manager.cpp
```diff
@@ -2902,15 +2932,25 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
             if (window)
             {
                 auto& info = info_for(window);
+                if (surface_stack->screen_is_locked() && !info.visible_on_lock_screen())
+                    continue;
```
AlanGriffiths, line 2936: This looks wrong. BWM doesn't seem the right place for logic choosing which windows are visible in the scene.

src/miral/basic_window_manager.cpp
```diff
@@ -2902,15 +2932,25 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
...
 
                 switch (info.state())
                 {
                 case mir_window_state_attached:
-                    if (info.exclusive_rect().is_set())
+                    if (info.ignore_exclusion_zones())
+                    {
+                        full_screen_attached_windows.push_back(&info);
```
AlanGriffiths, line 2943: `ignore_exclusion_zones()` does not imply fullscreen. Just that to ignores exclusion zones for attached edges.

src/server/frontend_wayland/layer_shell_v1.cpp
```diff
@@ -359,9 +359,8 @@ auto mf::LayerSurfaceV1::get_anchored_edge() const -> MirPlacementGravity
         return mir_placement_gravity_center;
 }
 
-auto mf::LayerSurfaceV1::get_exclusive_rect() const -> std::optional<geom::Rectangle>
+auto mf::LayerSurfaceV1::get_exclusive_rect(int zone) const -> std::optional<geom::Rectangle>
 {
-    auto zone = exclusive_zone.pending();
```
AlanGriffiths, line 364: It is not meaningful to pass `zone` to `LayerSurfaceV1::get_exclusive_rect()` as a surface only has one, optional, exclusive rect to get! Essentially, this introduces the unnecessary precondition that the one and only valid value is passed. (Either the change is wrong, or the function name is wrong (c.f. `compute_exclusive_rect_for(zone)`)

AlanGriffiths (changes requested):

src/miral/basic_window_manager.cpp
mattkae: Unlocking requires an update because the exclusion zones for existing windows don't get recalculated automatically after an unlock, meaning that if you "fullscreen" a window, it will appear behind the panels

src/miral/basic_window_manager.cpp
mattkae: To be honest, a reasonable way around this is to set `ignore_exclusion_zones` to `true` for all ext-session-lock-v1 surfaces. This _would_ make sense WRT to the spec, and then we wouldn't need to update when exclusion zones change at all. This might be a better thing to do. WDYT

src/miral/basic_window_manager.cpp
mattkae: _Maybe_ we don't need the locking idea in the `BasicWindowManager` at all. See below

src/miral/basic_window_manager.cpp
mattkae: @AlanGriffiths I left a question for you here, which is why I'm requesting a rereview

src/miral/basic_window_manager.cpp
AlanGriffiths: Either way, this needs addressing

AlanGriffiths (changes requested):

mattkae:
@AlanGriffiths  
> From your other comment:
> 
>     Unlocking requires an update because the exclusion zones for existing windows don't get recalculated automatically after an unlock, meaning that if you "fullscreen" a window, it will appear behind the panels
> 
> Wouldn't that be addressed by not skipping windows here?

But we need to skip windows in the first place so that non-lockscreen windows don't weight in on the exclusive zones when the session is locked.

AlanGriffiths:
> > Wouldn't that be addressed by not skipping windows here?
> 
> But we need to skip windows in the first place so that non-lockscreen windows don't weight in on the exclusive zones when the session is locked.

Not really, I expect exclusive zones to affect windows in `mir_depth_layer_overlay`, which is where we should be putting lockscreen windows

mattkae:
@AlanGriffiths Updated to not care about the lockscreen: much cleaner!

RAOF (approved):
Yup, LGTM.

RAOF:
> > > Wouldn't that be addressed by not skipping windows here?
> > 
> > 
> > But we need to skip windows in the first place so that non-lockscreen windows don't weight in on the exclusive zones when the session is locked.
> 
> Not really, I do not expect exclusive zones to affect windows in `mir_depth_layer_overlay`, which is where we should be putting lockscreen windows

There's nothing specific in the protocol about how exclusive zones interact with layers, so there are use-cases that are unfulfillable by the protocol as is.

A (non-exhaustive!) list of things clients might conceivably want to do:
* Ignore all exclusive zones
* Be constrained by exclusive zones only on my layer
* Be constrained by exclusive zones on my layer or below
* Be constrained by exclusive zones on my layer or above
* Be constrained by *all* exclusive zones

Since we have to pick which ones, I think the useful ones to support “are ignore all exclusive zones” (`-1`) and “respect all exclusive zones”

AlanGriffiths (approved):
This looks fine, but can we drop the rambling commit history before merging?

mattkae:
> This looks fine, but can we drop the rambling commit history before merging?

On it!

PR #3287, opened 2024-03-15
Drop dead code

AlanGriffiths:
We don't use it anymore


mattkae (approved):
If no one expects it, then I don't either :+1:

PR #3291, opened 2024-03-18
Add unit test for early observers

hbatagelo:
This PR suggests a unit test for the "early observers" feature added to `ObserverRegistrar` in #3244.


AlanGriffiths (approved):
OK

PR #3292, opened 2024-03-18
(#3259) buffer size should be the possibly-inverted extents size so that the final image is correct

### LINKED ISSUE #3259: Screenshot outputs weirdly shifted image.
Hi! 
I am using ubuntu frame for a kiosk. Everything went smoothly but now I am trying to implement screenshot-based sanity checks. When I call `ubuntu-frame.screenshot` the output image is shifted.

I am using a 1920x1080@60.0 screen that is rotated so orientation is set to left.
The program running is wpe-webkit-mir-kiosk.

Thanks a lot in advance

mattkae:
fixes #3259 

# Screenshots
## Left
20240318_15h54m23s_grim

## Right
20240318_15h54m44s_grim

## Inverted (Scaled)
20240318_15h55m27s_grim

## Normal
20240318_15h56m07s_grim

### DIFF
--- a/src/server/frontend_wayland/wlr_screencopy_v1.cpp
+++ b/src/server/frontend_wayland/wlr_screencopy_v1.cpp
@@ -378,8 +378,7 @@ void mf::WlrScreencopyManagerV1::capture_output(
     (void)overlay_cursor;
     auto const& output_config = OutputGlobal::from_or_throw(output).current_config();
     auto const extents = output_config.extents();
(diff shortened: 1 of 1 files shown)


AlanGriffiths (approved):
This looks reasonable, but could you also look at the next function in the same file? That also appears to conflate the output mode size with the configured extents.

mattkae:
> This looks reasonable, but could you also look at the next function in the same file? That also appears to conflate the output mode size with the configured extents.

I have fixed this now too. Re-requesting you :+1:

PR #3295, opened 2024-03-21
Build fixes for Clang & Ubuntu 24.04

RAOF:
I don't know why these are now tripping up clang on 24.04,
(or why the built *anywhere* before), but they now do, and
they're correct changes, so :woman_shrugging:


Saviq (approved):
Yups.

PR #3305, opened 2024-03-29
feature: generating miral and miroil symbols using a new script

mattkae:
## What's new?
- Created a new project in `tools/symbols_map_generator/` which will read the header files of a library and output a corresponding `symbols.map` file
- Used this script to update miral and mirserver `symbols.map` files

## Usage
Please note that you _may_ have to install clang and set the following environment variables, depending on your default version of libclang, e.g. mine is version 19 (I recommend either 18 or 19):
```sh
export MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH=/usr/lib/llvm-19/lib/libclang.so.1
export MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib
```

Then, you can run like this:
```sh
cmake --build ./build --target generate-miral-symbols-map -- -j 10
cmake --build ./build --target generate-miroil-symbols-map -- -j 10
```


tools/symbols_map_generator.py
```diff
@@ -0,0 +1,284 @@
...
+        "external_output_file": "src/miral/symbols_external.map",
+        "internal_output_file": "src/miral/symbols_internal.map"
+    }
+}
+
+
+def get_absolute_path_form_project_path(project_path: str) -> str:
+    script_path = os.path.dirname(os.path.realpath(__file__))
+    path = Path(script_path).parent.joinpath(project_path)
+    return path.absolute().as_posix()
```
Saviq, line 57: *from

But also, stick to `Path`:
```python
return Path(__file__).parent / project_path
```

You can then [`.glob()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) that, no need for the `glob` module.

mattkae: It could be an arg if you want? Although it's not like it will be ran from anywhere else IMO

mattkae:
@Saviq : What do you think the natural use of this is? I see a couple avenues:

1. We mention it in documentation, and then people can use it as-needed
2. We run it in CI (but this would mean we need to be more sophisticated about it IMO, as we'd want to generate new stanzas intelligently)

At any rate, it generates the right thing now. Although maybe the mirserver's `symbols.map` needs a new stanza instead of a modified one.

Saviq:
> @Saviq : What do you think the natural use of this is?

Ideally we would fail CI if the generator comes up with different symbols than are checked in, disregarding separated stanzas. We only need to let the author / reviewer know that there's a difference.

But for now, documenting its use would be a step in the right direction.

AlanGriffiths:
> > This was born from giving symbols from the "internal" headers some distinctive linkage. I think that implies the `MIR_SERVER_2.17` stanza being changed.
> 
> Do you mean that we should change it to `MIR_SERVER_2.18`? Changed how?

No, `2.17` is right for the next minor release. But we want to distinguish "internal" symbols. There was some discussion of prior art (Qt) that I can't find just now. But something like `MIR_SERVER_INTERNAL_2.17` is appropriate for `src/include` symbols

AlanGriffiths:
> > > This was born from giving symbols from the "internal" headers some distinctive linkage. I think that implies the `MIR_SERVER_2.17` stanza being changed.
> > 
> > 
> > Do you mean that we should change it to `MIR_SERVER_2.18`? Changed how?
> 
> No, `2.17` is right for the next minor release. But we want to distinguish "internal" symbols. There was some discussion of prior art (Qt) that I can't find just now. But something like `MIR_SERVER_UNSTABLE_2.17` or `MIR_SERVER_INTERNAL_2.17` is appropriate for `src/include` symbols

tools/symbols_map_generator.py
```diff
@@ -0,0 +1,343 @@
...
+
+def output_symbols_to_file(symbols: list[str], library: str, version: str, output_file: str):
+    joint_symbols = "\n    ".join(symbols)
+    output_str = f'''{library.upper()}_{version} {"{"}
+global:
+  extern "C++" {"{"}
+    {joint_symbols}
+  {'};'}
+local: *;
+{'}'};'''
```
AlanGriffiths, line 278: This looks like it is outputting a single stanza, which is wrong in the general case. Except when ABI breaks symbols should be in a stanza that matches the minor release in which they were introduced. (Have a look at `src/common/symbols.map` for an example.)

That means we need to reproduce "old" stanzas verbatum and strip any symbols those provide from the outputt to the "current" stanza. (See `src/miral/regenerate-miral-symbols-map.py:566` for the corresponding logic)

src/miral/symbols.map
```diff
@@ -32,6 +32,7 @@ global:
     miral::CanonicalWindowManagerPolicy::handle_raise_window*;
     miral::CanonicalWindowManagerPolicy::handle_window_ready*;
     miral::CanonicalWindowManagerPolicy::place_new_window*;
+    miral::CanonicalWindowManagerPolicy::tools;
```
AlanGriffiths, line 35: The filtering is wrong - instance data members are not exported as symbols.

src/miral/symbols.map
```diff
@@ -82,8 +85,7 @@ global:
     miral::MinimalWindowManager::handle_touch_event*;
     miral::MinimalWindowManager::handle_window_ready*;
     miral::MinimalWindowManager::place_new_window*;
-    miral::MinimalWindowManager::advise_new_window*;
-    miral::MinimalWindowManager::advise_delete_window*;
+    miral::MinimalWindowManager::tools;
```
AlanGriffiths, line 88: Instance data members are not exported as symbols

src/miral/symbols.map
```diff
@@ -99,6 +101,9 @@ global:
     miral::MirRunner::x11_display*;
     miral::Output::?Output*;
     miral::Output::Output*;
+    miral::Output::PhysicalSizeMM::height;
+    miral::Output::PhysicalSizeMM::width;
+    miral::Output::Type*;
```
AlanGriffiths, line 106: Instance data members and enum names are not exported as symbols

src/miral/symbols.map
```diff
@@ -137,6 +142,8 @@ global:
     miral::StartupInternalClient::StartupInternalClient*;
     miral::StartupInternalClient::operator*;
     miral::WaylandExtensions::?WaylandExtensions*;
+    miral::WaylandExtensions::Builder::Context;
+    miral::WaylandExtensions::Builder::name;
```
AlanGriffiths, line 146: Instance data members are not exported as symbols

AlanGriffiths (changes requested):
We need to support multiple stanzas (from previous releases) as we typically maintain ABI over multiple releases. Support for this is the biggest issue here.

It would be nice if the clang based parsing allowed us to fully specify symbols: `"mir::Foo::bar(Bas*) const"` instead of `mir::Foo::bar*` as we sometimes have issues with new symbols matching the glob in an old stanza. (The existing script needs manual "hacking" of the "old" stanza to accommodate this.)

There are also some spurious symbols added to the map, that is comparitively benign except that the globs could accidentally match a symbol in future.

tools/symbols_map_generator.py
```diff
@@ -0,0 +1,343 @@
...
+        return node.displayname
+    
+    if node.kind == clang.cindex.CursorKind.DESTRUCTOR:
+        return f"?{node.spelling[1:]}*"
+    
+    if ((node.kind == clang.cindex.CursorKind.FUNCTION_DECL 
+        or node.kind == clang.cindex.CursorKind.CXX_METHOD
+        or node.kind == clang.cindex.CursorKind.CONVERSION_FUNCTION)
+        and is_operator_overload(node.spelling)):
+        return "operator*"
```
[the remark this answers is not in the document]
mattkae: So I got non-globbing working _somewhat_, but it seems like more of a rabbit hole than I'd like. For example, clang will parse `void do_x(int const& y)` to `void do_x(const int & y)`. These are the same thing, but if I output the symbol as the latter, then the former won't resolve. It seems to be an exact match, rather than a semantic match? I suppose I could go and find the exact line in the header file and copy and paste that somehow though... I will check that out. It' super finicky though

tools/symbols_map_generator.py
```diff
@@ -0,0 +1,278 @@
...
+        "headers_dir": [
+            "include/server/mir",
+            "src/include/server/mir"
+        ],
+        "map_file": "src/server/symbols.map",
+        "include_headers": [
+            "src/include/server"
+        ]
+    }
+}
```
RAOF, line 54: I feel like these should be command line parameters?
`--headers-dir` / `--internal-headers-dir` / `--symbol-map`?

tools/symbols_map_generator.py
RAOF: > That means we need to reproduce "old" stanzas verbatum and strip any symbols those provide from the outputt to the "current" stanza. (See src/miral/regenerate-miral-symbols-map.py:566 for the corresponding logic)

An (more invasive) alternative we *could* choose here would be to tag everything with `[[mir:since("$VERSION")]]`, pull that out in the generator, and generate the set of stanzas based on the since-version.

That's probably more work than we want to do now, though.

tools/symbols_map_generator.py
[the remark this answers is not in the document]
AlanGriffiths: I don't get what you have tried and what didn't work.

The actual symbol that gets matched by the linker is the mangled form (something like `_ZN4do_xERKi` in your example). I don't see how the left const or right const spelling can affect that. All we need to do is be consistent in the representation used in the symbol file.

tools/symbols_map_generator.py
AlanGriffiths: > That's probably more work than we want to do now, though.

It also adds the complexity of needing to group any pre-ABI break symbols together.

AlanGriffiths (changes requested):
Just to document discussion between @Saviq, @mattkae and myself:

The latest approach suggested is to simply generate a list of symbols and compare that list with those in the symbols.map file and manually resolve discrepancies. For this to work we need to list "missing" and "unrecognised" symbols.

This will work best if we avoid glob matching and use the full function signatures, and to do that means a one-off manual exercise to update the symbols.map files to use that format. (This is a good time, as we we broke ABI on main and only have a single stanza to manage).

tools/symbols_map_generator.py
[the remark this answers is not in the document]
mattkae: Somehow, the west vs east const spelling does affect it. I find this odd too, but when the `const` matches that found in the file, then the symbols resolves. Otherwise, it fails to resolve. I will confirm this with an example, but it is odd to me as well.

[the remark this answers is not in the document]
mattkae: Maybe I'm doing something wrong? As an example:

```diff
- miral::ExternalClientLauncher::launch*;
+ "miral::ExternalClientLauncher::launch(std::vector<std::string> const& command_line)";
```

This doesn't work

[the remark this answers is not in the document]
mattkae: Also, I _can_ output the mangled name here. Could we just do that?

tools/symbols_map_generator.py
mattkae: I also think that `--internal-headers-dir` is useless to be honest. Those are two different libraries being created by two different commands IMO

tools/symbols_map_generator.py
mattkae: Removed that :)

tools/symbols_map_generator.py
```diff
@@ -0,0 +1,343 @@
...
+            if parent.kind == clang.cindex.CursorKind.TRANSLATION_UNIT:
+                break
+
+            namespace_str = f"{parent.spelling}::{namespace_str}"
+            parent = parent.semantic_parent
+
+        # Classes and structs have a specific outpu
+        if (node.kind == clang.cindex.CursorKind.CLASS_DECL
+            or node.kind == clang.cindex.CursorKind.STRUCT_DECL):
+            result.add(f"vtable?for?{namespace_str};")
```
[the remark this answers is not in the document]
mattkae: Yeah it seemed like we outputted it all the time before. We could filter to _just_ classes that have virtual methods though

tools/symbols_map_generator.py
mattkae: How do we manage dependencies here actually? We need to tell clang which includes it should resolve. I'm not sure how to do this without hardcoded paths ATM

RAOF: > I also think that --internal-headers-dir is useless to be honest. Those are two different libraries being created by two different commands IMO

Wasn't the original point of this exercise to be able to spit out a `MIR_SERVER_INTERNAL` stanza for `libmirserver.so`? That'll be one library and one `symbols.map` file but with two disjoint sets of headers.

AlanGriffiths: > Wasn't the original point of this exercise to be able to spit out a `MIR_SERVER_INTERNAL` stanza

It was. Although _at present_ all the symbols from libmirserver should be `MIR_SERVER_INTERNAL`

tools/symbols_map_generator.py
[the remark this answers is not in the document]
AlanGriffiths: "Before" was limited by the doxygen XML output - we couldn't tell whether there were virtual members.

AlanGriffiths (dismissed):
I tried running this and the first thing I see is:

```
/snap/clion/274/bin/cmake/linux/x64/bin/cmake --build /home/alan/CLionProjects/mir/cmake-build-debug --target regenerate-miral-debian-symbols -- -j 10
Traceback (most recent call last):
  File "/home/alan/CLionProjects/mir/tools/symbols_map_generator.py", line 21, in <module>
    import clang.cindex
```
We should ensure dependencies are met. Probably by setting up a venv?

I also found this has a dependency on libclang-dev, which is also not addressed.

I guess these are not strictly "build-deps" as this in not running in the build, but they ought to be addressed somehow. At the very least a "please install Foo & Bar" rather than crashing.

---

In any case, to try something more interesting than the single stanza we have currently, I tried reverting miral/symbols.map to 06f516cb6b8ef268b228a8a8ddf6be6925afa31f and building `regenerate-miral-debian-symbols`:

```
cmake --build /home/alan/CLionProjects/mir/cmake-build-debug --target regenerate-miral-debian-symbols -- -j 10
INFO:__main__:Symbols map generation is beginning for library=miral with version=5.0
DEBUG:__main__:Processing external directory: /home/alan/CLionProjects/mir/include/miroil
Traceback (most recent call last):
  File "/home/alan/CLionProjects/mir/tools/symbols_map_generator.py", line 376, in <module>
    main()
  File "/home/alan/CLionProjects/mir/tools/symbols_map_generator.py", line 304, in main
    parsed_symbols = parsed_symbols.union(process_directory(header_dir, search_dirs))
  File "/home/alan/CLionProjects/mir/tools/symbols_map_generator.py", line 226, in process_directory
    tu = idx.parse(
  File "/home/alan/.local/lib/python3.10/site-packages/clang/cindex.py", line 2722, in parse
    return TranslationUnit.from_source(path, args, unsaved_files, options,
  File "/home/alan/.local/lib/python3.10/site-packages/clang/cindex.py", line 2837, in from_source
    raise TranslationUnitLoadError("Error parsing translation unit.")
clang.cindex.TranslationUnitLoadError: Error parsing translation unit.
gmake[3]: *** [src/miral/CMakeFiles/regenerate-miral-debian-symbols.dir/build.make:70: src/miral/CMakeFiles/regenerate-miral-debian-symbols] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:3939: src/miral/CMakeFiles/regenerate-miral-debian-symbols.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:3946: src/miral/CMakeFiles/regenerate-miral-debian-symbols.dir/rule] Error 2
gmake: *** [Makefile:998: regenerate-miral-debian-symbols] Error 2
```

So, while the symbols.map you get looks sane, I cannot reproduce. Yet.

mattkae:
> I guess these are not strictly "build-deps" as this in not running in the build, but they ought to be addressed somehow. At the very least a "please install Foo & Bar" rather than crashing.

Would the "help" text of the script be a good place to put this? Perhaps I could write up some documentation? I want to modify the `DSO versioning guide` file at some point. Would this be the time you think?

> So, while the symbols.map you get looks sane, I cannot reproduce. Yet.

Whoop.s I made an 11th hour change and miral process miroil headers. The error should now be fixed!

mattkae:
> > > 
> > 
> > 
> > Whoops! I made an 11th hour change and miral process miroil headers. The error should now be fixed!
> 
> Sorry, but...
> 
> ```
> cmake --build /home/alan/CLionProjects/mir/cmake-build-debug --target regenerate-miral-debian-symbols -- -j 10
> INFO:__main__:Symbols map generation is beginning for library=miral with version=5.0
> DEBUG:__main__:Processing external directory: /home/alan/CLionProjects/mir/include/miral
> Traceback (most recent call last):
>   File "/home/alan/CLionProjects/mir/tools/symbols_map_generator.py", line 376, in <module>
>     main()
>   File "/home/alan/CLionProjects/mir/tools/symbols_map_generator.py", line 304, in main
>     parsed_symbols = parsed_symbols.union(process_directory(header_dir, search_dirs))
>   File "/home/alan/CLionProjects/mir/tools/symbols_map_generator.py", line 226, in process_directory
>     tu = idx.parse(
>   File "/home/alan/.local/lib/python3.10/site-packages/clang/cindex.py", line 2722, in parse
>     return TranslationUnit.from_source(path, args, unsaved_files, options,
>   File "/home/alan/.local/lib/python3.10/site-packages/clang/cindex.py", line 2837, in from_source
>     raise TranslationUnitLoadError("Error parsing translation unit.")
> clang.cindex.TranslationUnitLoadError: Error parsing translation unit.
> gmake[3]: *** [src/miral/CMakeFiles/regenerate-miral-debian-symbols.dir/build.make:70: src/miral/CMakeFiles/regenerate-miral-debian-symbols] Error 1
> gmake[2]: *** [CMakeFiles/Makefile2:3939: src/miral/CMakeFiles/regenerate-miral-debian-symbols.dir/all] Error 2
> gmake[1]: *** [CMakeFiles/Makefile2:3946: src/miral/CMakeFiles/regenerate-miral-debian-symbols.dir/rule] Error 2
> gmake: *** [Makefile:998: regenerate-miral-debian-symbols] Error 2
> ```

That's odd. Works on my machine :wink: You're using `libclang 18`?

mattkae:
I refactored the script to automatically enter a venv. I also added a `README` to instruct the user to download `libclang-dev` themseleces

AlanGriffiths:
> That's odd. Works on my machine 😉 You're using `libclang 18`?

No...

```
$ apt policy libclang-dev
libclang-dev:
  Installed: 1:14.0-55~exp2
  Candidate: 1:14.0-55~exp2
  Version table:
 *** 1:14.0-55~exp2 500
        500 http://gb.archive.ubuntu.com/ubuntu jammy/universe amd64 Packages
        100 /var/lib/dpkg/status
```

tools/symbols_map_generator/main.py
```diff
@@ -348,19 +348,24 @@ def main():
                 symbols = list(stanza[1])
                 symbols.sort()
 
+                # Only the final stanza should contain the local symbols
+                if i == len(previous_symbols) - 1:
+                    closing_line = "local: *;\n"
+                else:
+                    closing_line = ""
```
AlanGriffiths, line 355: I think the "local: *;" incantation belongs in the primary (first) stanza

mattkae:
We need to handle C symbols as well :upside_down_face:

mattkae:
> > That's odd. Works on my machine 😉 You're using `libclang 18`?
> 
> No...
> 
> ```
> $ apt policy libclang-dev
> libclang-dev:
>   Installed: 1:14.0-55~exp2
>   Candidate: 1:14.0-55~exp2
>   Version table:
>  *** 1:14.0-55~exp2 500
>         500 http://gb.archive.ubuntu.com/ubuntu jammy/universe amd64 Packages
>         100 /var/lib/dpkg/status
> ```
> 
> In fact 18 isn't available in any supported release:
> version 	release
> 1:3.4-0ubuntu1 	trusty/universe
> 1:3.8-33ubuntu3 	xenial/universe
> 1:3.8-33ubuntu3.1 	xenial-updates/universe
> 1:6.0-41~exp4 	bionic/universe
> 1:6.0-41~exp5~ubuntu1 	bionic-updates/universe
> 1:10.0-50~exp1 	focal/universe
> 1:14.0-55~exp2 	jammy/universe
> 1:16.0-57 	mantic/universe
> 1:18.0-59~exp2 	noble/universe

Maybe clang 18 isn't the problem. I'm also on 16. What ubuntu version are you on? I'll try to repro on my end

AlanGriffiths:
> Maybe clang 18 isn't the problem. I'm also on 16. What ubuntu version are you on? I'll try to repro on my end

I'm using clang14 on jammy (as shown above):

> > ```
> >  *** 1:14.0-55~exp2 500
> >         500 http://gb.archive.ubuntu.com/ubuntu jammy/universe amd64 Packages
> > ```

src/miral/CMakeLists.txt
```diff
@@ -137,8 +137,7 @@ if (CMAKE_COMPILER_IS_GNUCXX   AND  # clang generates slightly different symbols
 endif()
 
 add_custom_target(regenerate-miral-debian-symbols ${MIR_CHECK_MIRAL_SYMBOLS_DEFAULT}
-    DEPENDS miral ${PROJECT_SOURCE_DIR}/debian/libmiral${MIRAL_ABI}.symbols
-    COMMAND python3 ${CMAKE_CURRENT_SOURCE_DIR}/check-and-update-debian-symbols.py ${CMAKE_LIBRARY_OUTPUT_DIRECTORY} miral ${MIRAL_VERSION} ${MIRAL_ABI}
+    COMMAND ${PROJECT_SOURCE_DIR}/tools/symbols_map_generator/run.sh --library miral --version ${MIRAL_VERSION} --output_symbols
     WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
     VERBATIM)
```
AlanGriffiths, line 142: This is the wrong target to be changing. This is to generate `debian/libmiral${MIRAL_ABI}.symbols`

src/miral/CMakeLists.txt
```diff
@@ -153,12 +152,6 @@ configure_file(
     ${miral_include}/miral/version.h
 )
 
-if(TARGET doc)
-    add_custom_target(regenerate-miral-symbols-map
-            ${CMAKE_CURRENT_SOURCE_DIR}/regenerate-miral-symbols-map.py ${CMAKE_BINARY_DIR}/doc/sphinx/xml/*.xml > ${symbol_map}
-            DEPENDS doc)
-endif()
-
```
AlanGriffiths, line 161: This is the target that should be changed

AlanGriffiths:
I tried this (getting a symbols.map with many stanzas):

```
git checkout 06f516cb6b8ef268b228a8a8ddf6be6925afa31f src/miral/symbols.map
cmake --build cmake-build-debug/ --target regenerate-miral-symbols-map
```

But the result isn't what I expect (earlier stanzas unchanged, new stuff in new stanza)

mattkae:
> I tried this (getting a symbols.map with many stanzas):
> 
> ```
> git checkout 06f516cb6b8ef268b228a8a8ddf6be6925afa31f src/miral/symbols.map
> cmake --build cmake-build-debug/ --target regenerate-miral-symbols-map
> ```
> 
> But the result isn't what I expect (earlier stanzas unchanged, new stuff in new stanza)

What do you mean that your `symbols.map` has many stanzas? (FYI: I just changed the name of the target to `generate-miral-symbols-map`)

mattkae:
@AlanGriffiths I ran this on Ubuntu 22 and it ran just fine. Not sure what could be up on your system

AlanGriffiths:
> What do you mean that your `symbols.map` has many stanzas? (FYI: I just changed the name of the target to `generate-miral-symbols-map`)

I mean I checked out an old version which had many stanzas:

```
git checkout 06f516cb6b8ef268b228a8a8ddf6be6925afa31f src/miral/symbols.map
```

mattkae:
> > What do you mean that your `symbols.map` has many stanzas? (FYI: I just changed the name of the target to `generate-miral-symbols-map`)
> 
> I mean I checked out an old version which had many stanzas:
> 
> ```
> git checkout 06f516cb6b8ef268b228a8a8ddf6be6925afa31f src/miral/symbols.map
> ```

I just tried this, and I got the right result, namely:
1. New symbols appeared in the 5.0 stanza
2. Previous stanzas had some _minor_ changes due to the different sorting algorithm (I sort by string), but they remained unchanged in content

mattkae:
I am unsure why, but using C++23 as the standard library leaves the `std` namespace open such that mir symbols remain in the std namespace. I switched to C++17 and it all worked, but this is obviously wrong... We could ignore items in the `std::` namespace for now, but that is an odd thing to do.

tools/symbols_map_generator/main.py
```diff
@@ -0,0 +1,459 @@
...
+                    dest="external_headers")
+    parser.add_argument('--internal-headers-directory', type=str,
+                    help=f'absolute path to the directory containing private headers of this library',
+                    dest="internal_headers")
+    parser.add_argument('--version', type=str,
+                    help='current version of the library',
+                    required=True)
+    parser.add_argument('--diff', action='store_true',
+                    help='if true a diff should be output to the console')
+    parser.add_argument('--output_symbols', action='store_true',
```
RAOF, line 315: Nit: Should probably be spelt `--output-symbols` to match other options?

tools/symbols_map_generator/main.py
```diff
@@ -0,0 +1,459 @@
...
+                        help='if true, the symbols.map file will be updated with the new new symbols')
+    parser.add_argument('--include_dirs', type=str, help="colon separated list of directories to search for symbols", required=True)
```
RAOF, line 317: Likewise, `--include-dirs`?

RAOF (commented):
This looks pretty good, but doing the `git checkout 06f516cb6b8ef268b228a8a8ddf6be6925afa31f -- src/miral/symbols.map` followed by `ninja --verbose -C build generate-miral-symbols-map` I get the following diff:
```diff
diff --git a/src/miral/symbols.map b/src/miral/symbols.map
index d1d0d69bc9..b1f6b53233 100644
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -1,10 +1,4 @@
 MIRAL_4.0 {
-# WaylandExtensions::Context ctor and dtor symbols are inline
-# but can be exported by unoptimized builds.
-# This makes the "regenerate" script hide them:
-#    miral::WaylandExtensions::Context::?Context*;
-#    miral::WaylandExtensions::Context::Context*;
-#    vtable?for?miral::WaylandExtensions::Context;
 global:
   extern "C++" {
     miral::AddInitCallback::?AddInitCallback*;
@@ -411,25 +405,25 @@ global:
     typeinfo?for?miral::Keymap;
     typeinfo?for?miral::MinimalWindowManager;
     typeinfo?for?miral::MirRunner;
-    typeinfo?for?miral::Output;
     typeinfo?for?miral::Output::PhysicalSizeMM;
+    typeinfo?for?miral::Output;
     typeinfo?for?miral::PrependEventFilter;
     typeinfo?for?miral::SetCommandLineHandler;
     typeinfo?for?miral::SetTerminator;
     typeinfo?for?miral::SetWindowManagementPolicy;
     typeinfo?for?miral::StartupInternalClient;
-    typeinfo?for?miral::WaylandExtensions;
     typeinfo?for?miral::WaylandExtensions::Builder;
     typeinfo?for?miral::WaylandExtensions::Context;
     typeinfo?for?miral::WaylandExtensions::EnableInfo;
+    typeinfo?for?miral::WaylandExtensions;
     typeinfo?for?miral::Window;
     typeinfo?for?miral::WindowInfo;
     typeinfo?for?miral::WindowManagementPolicy;
     typeinfo?for?miral::WindowManagerOption;
     typeinfo?for?miral::WindowManagerOptions;
     typeinfo?for?miral::WindowManagerTools;
-    typeinfo?for?miral::WindowSpecification;
     typeinfo?for?miral::WindowSpecification::AspectRatio;
+    typeinfo?for?miral::WindowSpecification;
     typeinfo?for?miral::X11Support;
     typeinfo?for?miral::Zone;
     vtable?for?miral::AddInitCallback;
@@ -448,42 +442,40 @@ global:
     vtable?for?miral::Keymap;
     vtable?for?miral::MinimalWindowManager;
     vtable?for?miral::MirRunner;
-    vtable?for?miral::Output;
     vtable?for?miral::Output::PhysicalSizeMM;
+    vtable?for?miral::Output;
     vtable?for?miral::PrependEventFilter;
     vtable?for?miral::SetCommandLineHandler;
     vtable?for?miral::SetTerminator;
     vtable?for?miral::SetWindowManagementPolicy;
     vtable?for?miral::StartupInternalClient;
-    vtable?for?miral::WaylandExtensions;
     vtable?for?miral::WaylandExtensions::Builder;
     vtable?for?miral::WaylandExtensions::EnableInfo;
+    vtable?for?miral::WaylandExtensions;
     vtable?for?miral::Window;
     vtable?for?miral::WindowInfo;
     vtable?for?miral::WindowManagementPolicy;
     vtable?for?miral::WindowManagerOption;
     vtable?for?miral::WindowManagerOptions;
     vtable?for?miral::WindowManagerTools;
-    vtable?for?miral::WindowSpecification;
     vtable?for?miral::WindowSpecification::AspectRatio;
+    vtable?for?miral::WindowSpecification;
     vtable?for?miral::X11Support;
     vtable?for?miral::Zone;
   };
-
-  local: *;
+local: *;
 };
 
 MIRAL_4.1 {
 global:
   extern "C++" {
-    miral::WaylandExtensions::zwp_input_method_v1*;
-    miral::WaylandExtensions::zwp_input_panel_v1*;
     miral::SmoothBootSupport::?SmoothBootSupport*;
     miral::SmoothBootSupport::SmoothBootSupport*;
     miral::SmoothBootSupport::operator*;
+    miral::WaylandExtensions::zwp_input_method_v1*;
+    miral::WaylandExtensions::zwp_input_panel_v1*;
     typeinfo?for?miral::SmoothBootSupport;
     vtable?for?miral::SmoothBootSupport;
   };
-
 } MIRAL_4.0;
 ```
(that's the diff against 06f516cb6b8ef268b228a8a8ddf6be6925afa31f).

The diff against main is:
```diff
diff --git a/src/miral/symbols.map b/src/miral/symbols.map
index 549ac28fd0..b1f6b53233 100644
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -1,4 +1,4 @@
-MIRAL_5.0 {
+MIRAL_4.0 {
 global:
   extern "C++" {
     miral::AddInitCallback::?AddInitCallback*;
@@ -64,11 +64,9 @@ global:
     miral::MinimalWindowManager::?MinimalWindowManager*;
     miral::MinimalWindowManager::MinimalWindowManager*;
     miral::MinimalWindowManager::advise_delete_app*;
-    miral::MinimalWindowManager::advise_delete_window*;
     miral::MinimalWindowManager::advise_focus_gained*;
     miral::MinimalWindowManager::advise_focus_lost*;
     miral::MinimalWindowManager::advise_new_app*;
-    miral::MinimalWindowManager::advise_new_window*;
     miral::MinimalWindowManager::begin_pointer_move*;
     miral::MinimalWindowManager::begin_pointer_resize*;
     miral::MinimalWindowManager::begin_touch_move*;
@@ -99,7 +97,6 @@ global:
     miral::MirRunner::x11_display*;
     miral::Output::?Output*;
     miral::Output::Output*;
-    miral::Output::Type*;
     miral::Output::attribute*;
     miral::Output::attributes_map*;
     miral::Output::connected*;
@@ -122,9 +119,6 @@ global:
     miral::PrependEventFilter::PrependEventFilter*;
     miral::PrependEventFilter::operator*;
     miral::PrintTo*;
-    miral::SessionLockListener::?SessionLockListener*;
-    miral::SessionLockListener::SessionLockListener*;
-    miral::SessionLockListener::operator*;
     miral::SetCommandLineHandler::?SetCommandLineHandler*;
     miral::SetCommandLineHandler::SetCommandLineHandler*;
     miral::SetCommandLineHandler::operator*;
@@ -138,8 +132,6 @@ global:
     miral::StartupInternalClient::StartupInternalClient*;
     miral::StartupInternalClient::operator*;
     miral::WaylandExtensions::?WaylandExtensions*;
-    miral::WaylandExtensions::Context::?Context*;
-    miral::WaylandExtensions::Context::Context*;
     miral::WaylandExtensions::Context::operator*;
     miral::WaylandExtensions::EnableInfo::app*;
     miral::WaylandExtensions::EnableInfo::name*;
@@ -160,8 +152,6 @@ global:
     miral::WaylandExtensions::zwlr_screencopy_manager_v1*;
     miral::WaylandExtensions::zwlr_virtual_pointer_manager_v1*;
     miral::WaylandExtensions::zwp_input_method_manager_v2*;
-    miral::WaylandExtensions::zwp_input_method_v1*;
-    miral::WaylandExtensions::zwp_input_panel_v1*;
     miral::WaylandExtensions::zwp_virtual_keyboard_manager_v1*;
     miral::WaylandExtensions::zxdg_output_manager_v1*;
     miral::Window::?Window*;
@@ -187,7 +177,6 @@ global:
     miral::WindowInfo::focus_mode*;
     miral::WindowInfo::has_output_id*;
     miral::WindowInfo::height_inc*;
-    miral::WindowInfo::ignore_exclusion_zones*;
     miral::WindowInfo::is_visible*;
     miral::WindowInfo::max_aspect*;
     miral::WindowInfo::max_height*;
@@ -206,7 +195,6 @@ global:
     miral::WindowInfo::restore_rect*;
     miral::WindowInfo::shell_chrome*;
     miral::WindowInfo::state*;
-    miral::WindowInfo::swap*;
     miral::WindowInfo::type*;
     miral::WindowInfo::userdata*;
     miral::WindowInfo::visible_on_lock_screen*;
@@ -244,7 +232,6 @@ global:
     miral::WindowManagerTools::active_window*;
     miral::WindowManagerTools::add_tree_to_workspace*;
     miral::WindowManagerTools::ask_client_to_close*;
-    miral::WindowManagerTools::can_select_window*;
     miral::WindowManagerTools::count_applications*;
     miral::WindowManagerTools::create_workspace*;
     miral::WindowManagerTools::drag_active_window*;
@@ -273,7 +260,6 @@ global:
     miral::WindowManagerTools::window_at*;
     miral::WindowManagerTools::window_to_select_application*;
     miral::WindowSpecification::?WindowSpecification*;
-    miral::WindowSpecification::InputReceptionMode*;
     miral::WindowSpecification::WindowSpecification*;
     miral::WindowSpecification::application_id*;
     miral::WindowSpecification::attached_edges*;
@@ -285,7 +271,6 @@ global:
     miral::WindowSpecification::exclusive_rect*;
     miral::WindowSpecification::focus_mode*;
     miral::WindowSpecification::height_inc*;
-    miral::WindowSpecification::ignore_exclusion_zones*;
     miral::WindowSpecification::input_mode*;
     miral::WindowSpecification::input_shape*;
     miral::WindowSpecification::max_aspect*;
@@ -312,7 +297,6 @@ global:
     miral::WindowSpecification::window_placement_gravity*;
     miral::X11Support::?X11Support*;
     miral::X11Support::X11Support*;
-    miral::X11Support::default_to_enabled*;
     miral::X11Support::operator*;
     miral::Zone::?Zone*;
     miral::Zone::Zone*;
@@ -338,6 +322,7 @@ global:
     miral::toolkit::mir_input_event_get_pointer_event*;
     miral::toolkit::mir_input_event_get_touch_event*;
     miral::toolkit::mir_input_event_get_type*;
+    miral::toolkit::mir_input_event_has_cookie*;
     miral::toolkit::mir_keyboard_event_action*;
     miral::toolkit::mir_keyboard_event_input_event*;
     miral::toolkit::mir_keyboard_event_key_text*;
@@ -367,13 +352,10 @@ global:
     non-virtual?thunk?to?miral::CanonicalWindowManagerPolicy::handle_window_ready*;
     non-virtual?thunk?to?miral::CanonicalWindowManagerPolicy::place_new_window*;
     non-virtual?thunk?to?miral::FdHandle::?FdHandle*;
-    non-virtual?thunk?to?miral::MinimalWindowManager::?MinimalWindowManager*;
     non-virtual?thunk?to?miral::MinimalWindowManager::advise_delete_app*;
-    non-virtual?thunk?to?miral::MinimalWindowManager::advise_delete_window*;
     non-virtual?thunk?to?miral::MinimalWindowManager::advise_focus_gained*;
     non-virtual?thunk?to?miral::MinimalWindowManager::advise_focus_lost*;
     non-virtual?thunk?to?miral::MinimalWindowManager::advise_new_app*;
-    non-virtual?thunk?to?miral::MinimalWindowManager::advise_new_window*;
     non-virtual?thunk?to?miral::MinimalWindowManager::confirm_inherited_move*;
     non-virtual?thunk?to?miral::MinimalWindowManager::confirm_placement_on_display*;
     non-virtual?thunk?to?miral::MinimalWindowManager::handle_keyboard_event*;
@@ -407,7 +389,6 @@ global:
     non-virtual?thunk?to?miral::WindowManagementPolicy::advise_removing_from_workspace*;
     non-virtual?thunk?to?miral::WindowManagementPolicy::advise_resize*;
     non-virtual?thunk?to?miral::WindowManagementPolicy::advise_state_change*;
-    std::swap*;
     typeinfo?for?miral::AddInitCallback;
     typeinfo?for?miral::AppendEventFilter;
     typeinfo?for?miral::ApplicationAuthorizer;
@@ -427,7 +408,6 @@ global:
     typeinfo?for?miral::Output::PhysicalSizeMM;
     typeinfo?for?miral::Output;
     typeinfo?for?miral::PrependEventFilter;
-    typeinfo?for?miral::SessionLockListener;
     typeinfo?for?miral::SetCommandLineHandler;
     typeinfo?for?miral::SetTerminator;
     typeinfo?for?miral::SetWindowManagementPolicy;
@@ -465,13 +445,11 @@ global:
     vtable?for?miral::Output::PhysicalSizeMM;
     vtable?for?miral::Output;
     vtable?for?miral::PrependEventFilter;
-    vtable?for?miral::SessionLockListener;
     vtable?for?miral::SetCommandLineHandler;
     vtable?for?miral::SetTerminator;
     vtable?for?miral::SetWindowManagementPolicy;
     vtable?for?miral::StartupInternalClient;
     vtable?for?miral::WaylandExtensions::Builder;
-    vtable?for?miral::WaylandExtensions::Context;
     vtable?for?miral::WaylandExtensions::EnableInfo;
     vtable?for?miral::WaylandExtensions;
     vtable?for?miral::Window;
@@ -488,3 +466,16 @@ global:
 local: *;
 };
 
+MIRAL_4.1 {
+global:
+  extern "C++" {
+    miral::SmoothBootSupport::?SmoothBootSupport*;
+    miral::SmoothBootSupport::SmoothBootSupport*;
+    miral::SmoothBootSupport::operator*;
+    miral::WaylandExtensions::zwp_input_method_v1*;
+    miral::WaylandExtensions::zwp_input_panel_v1*;
+    typeinfo?for?miral::SmoothBootSupport;
+    vtable?for?miral::SmoothBootSupport;
+  };
+} MIRAL_4.0;
+
```

As you can see, it's missing the whole MIRAL_5.0 stanza? Unless I'm misunderstanding the expected use, this doesn't seem to be generating the expected symbols?

src/miral/symbols.map
```diff
@@ -403,6 +407,7 @@ global:
     non-virtual?thunk?to?miral::WindowManagementPolicy::advise_removing_from_workspace*;
     non-virtual?thunk?to?miral::WindowManagementPolicy::advise_resize*;
     non-virtual?thunk?to?miral::WindowManagementPolicy::advise_state_change*;
+    std::swap*;
```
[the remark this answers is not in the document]
mattkae: We are :smile:

tools/symbols_map_generator/main.py
```diff
@@ -0,0 +1,459 @@
...
+    if (node.access_specifier == clang.cindex.AccessSpecifier.PRIVATE):
+        return result
+    
+    # Check if we need to output a symbol
+    if ((node.kind == clang.cindex.CursorKind.FUNCTION_DECL 
+          or node.kind == clang.cindex.CursorKind.CXX_METHOD
+          or node.kind == clang.cindex.CursorKind.VAR_DECL
+          or node.kind == clang.cindex.CursorKind.CONSTRUCTOR
+          or node.kind == clang.cindex.CursorKind.DESTRUCTOR
+          or node.kind == clang.cindex.CursorKind.ENUM_DECL
```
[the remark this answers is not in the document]
mattkae: Yup, you're probably right. I can remove em!

mattkae:
@RAOF I am definitely getting different results compared to you. For your example, I do get a 5.0 stanza. After running the tool, I do:
```
[10:06:31] ~/Github/mir (feature/symbols_generator) $ git diff origin/main -- src/miral/symbols.map
```

And I get:
```diff
diff --git a/src/miral/symbols.map b/src/miral/symbols.map
index f588c74cb2..256c09ff65 100644
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -1,4 +1,4 @@
-MIRAL_5.0 {
+MIRAL_4.0 {
 global:
   extern "C++" {
     miral::AddInitCallback::?AddInitCallback*;
@@ -82,8 +82,6 @@ global:
     miral::MinimalWindowManager::handle_touch_event*;
     miral::MinimalWindowManager::handle_window_ready*;
     miral::MinimalWindowManager::place_new_window*;
-    miral::MinimalWindowManager::advise_new_window*;
-    miral::MinimalWindowManager::advise_delete_window*;
     miral::MirRunner::?MirRunner*;
     miral::MirRunner::MirRunner*;
     miral::MirRunner::add_start_callback*;
@@ -121,9 +119,6 @@ global:
     miral::PrependEventFilter::PrependEventFilter*;
     miral::PrependEventFilter::operator*;
     miral::PrintTo*;
-    miral::SessionLockListener::?SessionLockListener*;
-    miral::SessionLockListener::SessionLockListener*;
-    miral::SessionLockListener::operator*;
     miral::SetCommandLineHandler::?SetCommandLineHandler*;
     miral::SetCommandLineHandler::SetCommandLineHandler*;
     miral::SetCommandLineHandler::operator*;
@@ -137,8 +132,6 @@ global:
     miral::StartupInternalClient::StartupInternalClient*;
     miral::StartupInternalClient::operator*;
     miral::WaylandExtensions::?WaylandExtensions*;
-    miral::WaylandExtensions::Context::?Context*;
-    miral::WaylandExtensions::Context::Context*;
     miral::WaylandExtensions::Context::operator*;
     miral::WaylandExtensions::EnableInfo::app*;
     miral::WaylandExtensions::EnableInfo::name*;
@@ -159,8 +152,6 @@ global:
     miral::WaylandExtensions::zwlr_screencopy_manager_v1*;
     miral::WaylandExtensions::zwlr_virtual_pointer_manager_v1*;
     miral::WaylandExtensions::zwp_input_method_manager_v2*;
-    miral::WaylandExtensions::zwp_input_method_v1*;
-    miral::WaylandExtensions::zwp_input_panel_v1*;
     miral::WaylandExtensions::zwp_virtual_keyboard_manager_v1*;
     miral::WaylandExtensions::zxdg_output_manager_v1*;
     miral::Window::?Window*;
@@ -183,7 +174,6 @@ global:
     miral::WindowInfo::constrain_resize*;
     miral::WindowInfo::depth_layer*;
     miral::WindowInfo::exclusive_rect*;
-    miral::WindowInfo::ignore_exclusion_zones*;
     miral::WindowInfo::focus_mode*;
     miral::WindowInfo::has_output_id*;
     miral::WindowInfo::height_inc*;
@@ -269,7 +259,6 @@ global:
     miral::WindowManagerTools::swap_tree_order*;
     miral::WindowManagerTools::window_at*;
     miral::WindowManagerTools::window_to_select_application*;
-    miral::WindowManagerTools::can_select_window*;
     miral::WindowSpecification::?WindowSpecification*;
     miral::WindowSpecification::WindowSpecification*;
     miral::WindowSpecification::application_id*;
@@ -280,7 +269,6 @@ global:
     miral::WindowSpecification::confine_pointer*;
     miral::WindowSpecification::depth_layer*;
     miral::WindowSpecification::exclusive_rect*;
-    miral::WindowSpecification::ignore_exclusion_zones*;
     miral::WindowSpecification::focus_mode*;
     miral::WindowSpecification::height_inc*;
     miral::WindowSpecification::input_mode*;
@@ -309,7 +297,6 @@ global:
     miral::WindowSpecification::window_placement_gravity*;
     miral::X11Support::?X11Support*;
     miral::X11Support::X11Support*;
-    miral::X11Support::default_to_enabled*;
     miral::X11Support::operator*;
     miral::Zone::?Zone*;
     miral::Zone::Zone*;
@@ -335,6 +322,7 @@ global:
     miral::toolkit::mir_input_event_get_pointer_event*;
     miral::toolkit::mir_input_event_get_touch_event*;
     miral::toolkit::mir_input_event_get_type*;
+    miral::toolkit::mir_input_event_has_cookie*;
     miral::toolkit::mir_keyboard_event_action*;
     miral::toolkit::mir_keyboard_event_input_event*;
     miral::toolkit::mir_keyboard_event_key_text*;
@@ -379,8 +367,6 @@ global:
     non-virtual?thunk?to?miral::MinimalWindowManager::handle_touch_event*;
     non-virtual?thunk?to?miral::MinimalWindowManager::handle_window_ready*;
     non-virtual?thunk?to?miral::MinimalWindowManager::place_new_window*;
-    non-virtual?thunk?to?miral::MinimalWindowManager::advise_new_window*;
-    non-virtual?thunk?to?miral::MinimalWindowManager::advise_delete_window*;
     non-virtual?thunk?to?miral::WaylandExtensions::Context::?Context*;
     non-virtual?thunk?to?miral::WindowManagementPolicy::?WindowManagementPolicy*;
     non-virtual?thunk?to?miral::WindowManagementPolicy::advise_adding_to_workspace*;
@@ -419,26 +405,25 @@ global:
     typeinfo?for?miral::Keymap;
     typeinfo?for?miral::MinimalWindowManager;
     typeinfo?for?miral::MirRunner;
-    typeinfo?for?miral::Output;
     typeinfo?for?miral::Output::PhysicalSizeMM;
+    typeinfo?for?miral::Output;
     typeinfo?for?miral::PrependEventFilter;
-    typeinfo?for?miral::SessionLockListener;
     typeinfo?for?miral::SetCommandLineHandler;
     typeinfo?for?miral::SetTerminator;
     typeinfo?for?miral::SetWindowManagementPolicy;
     typeinfo?for?miral::StartupInternalClient;
-    typeinfo?for?miral::WaylandExtensions;
     typeinfo?for?miral::WaylandExtensions::Builder;
     typeinfo?for?miral::WaylandExtensions::Context;
     typeinfo?for?miral::WaylandExtensions::EnableInfo;
+    typeinfo?for?miral::WaylandExtensions;
     typeinfo?for?miral::Window;
     typeinfo?for?miral::WindowInfo;
     typeinfo?for?miral::WindowManagementPolicy;
     typeinfo?for?miral::WindowManagerOption;
     typeinfo?for?miral::WindowManagerOptions;
     typeinfo?for?miral::WindowManagerTools;
-    typeinfo?for?miral::WindowSpecification;
     typeinfo?for?miral::WindowSpecification::AspectRatio;
+    typeinfo?for?miral::WindowSpecification;
     typeinfo?for?miral::X11Support;
     typeinfo?for?miral::Zone;
     vtable?for?miral::AddInitCallback;
@@ -457,28 +442,65 @@ global:
     vtable?for?miral::Keymap;
     vtable?for?miral::MinimalWindowManager;
     vtable?for?miral::MirRunner;
-    vtable?for?miral::Output;
     vtable?for?miral::Output::PhysicalSizeMM;
+    vtable?for?miral::Output;
     vtable?for?miral::PrependEventFilter;
-    vtable?for?miral::SessionLockListener;
     vtable?for?miral::SetCommandLineHandler;
     vtable?for?miral::SetTerminator;
     vtable?for?miral::SetWindowManagementPolicy;
     vtable?for?miral::StartupInternalClient;
-    vtable?for?miral::WaylandExtensions;
     vtable?for?miral::WaylandExtensions::Builder;
-    vtable?for?miral::WaylandExtensions::Context;
     vtable?for?miral::WaylandExtensions::EnableInfo;
+    vtable?for?miral::WaylandExtensions;
     vtable?for?miral::Window;
     vtable?for?miral::WindowInfo;
     vtable?for?miral::WindowManagementPolicy;
     vtable?for?miral::WindowManagerOption;
     vtable?for?miral::WindowManagerOptions;
     vtable?for?miral::WindowManagerTools;
-    vtable?for?miral::WindowSpecification;
     vtable?for?miral::WindowSpecification::AspectRatio;
+    vtable?for?miral::WindowSpecification;
     vtable?for?miral::X11Support;
     vtable?for?miral::Zone;
   };
 local: *;
-};
\ No newline at end of file
+};
+
+MIRAL_4.1 {
+global:
+  extern "C++" {
+    miral::SmoothBootSupport::?SmoothBootSupport*;
+    miral::SmoothBootSupport::SmoothBootSupport*;
+    miral::SmoothBootSupport::operator*;
+    miral::WaylandExtensions::zwp_input_method_v1*;
+    miral::WaylandExtensions::zwp_input_panel_v1*;
+    typeinfo?for?miral::SmoothBootSupport;
+    vtable?for?miral::SmoothBootSupport;
+  };
+} MIRAL_4.0;
+
+MIRAL_5.0 {
+global:
+  extern "C++" {
+    miral::MinimalWindowManager::advise_delete_window*;
+    miral::MinimalWindowManager::advise_new_window*;
+    miral::SessionLockListener::?SessionLockListener*;
+    miral::SessionLockListener::SessionLockListener*;
+    miral::SessionLockListener::operator*;
+    miral::WaylandExtensions::Context::?Context*;
+    miral::WaylandExtensions::Context::Context*;
+    miral::WindowInfo::ignore_exclusion_zones*;
+    miral::WindowInfo::swap*;
+    miral::WindowManagerTools::can_select_window*;
+    miral::WindowSpecification::ignore_exclusion_zones*;
+    miral::X11Support::default_to_enabled*;
+    non-virtual?thunk?to?miral::MinimalWindowManager::?MinimalWindowManager*;
+    non-virtual?thunk?to?miral::MinimalWindowManager::advise_delete_window*;
+    non-virtual?thunk?to?miral::MinimalWindowManager::advise_new_window*;
+    std::swap*;
+    typeinfo?for?miral::SessionLockListener;
+    vtable?for?miral::SessionLockListener;
+    vtable?for?miral::WaylandExtensions::Context;
+  };
+} MIRAL_4.1;
+
```

mattkae:
@RAOF Are you setting the right environment variables to point to a proper version of clang (see the description)? Are you pointing at a clang that supports c++23?

Saviq:
We found that clang-18 from Ubuntu is not enough. Installing it from here makes it work:

https://github.com/opencollab/llvm-jenkins.debian.net/blob/master/llvm.sh

mattkae:
https://github.com/canonical/mir/pull/3336 This pr needs to be merged first :) It adds CI for all of this.

Also @RAOF, it turns out that the clang-18 version in the Ubuntu repo is (somehow) different from the upstream version. If you install using `llvm.sh` (like CI does), then you will find success

RAOF (approved):
Aha! With the extra context of the `MIR_SYMBOLS_MAP_GENERATOR_*` environment variables I can verify this works with 24.04's clang-19. Excellent!

mattkae:
@Saviq BTW, I will clean up the commit history, so once you approve, I'll clean it and _then_ merge

tools/symbols_map_generator/README.md
```diff
@@ -0,0 +1,26 @@
+# Symbols Map Generator
+This tool parses the header files of a library in the Mir project to extract symbols.
+These symbols can then either be outputted to `stdout` to display a diff or to the
+corresponding `symbols.map` file.
```
Saviq, line 4: That's out of date now, below too.

tools/symbols_map_generator/run.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+cd "$(dirname "$0")"
+rm -rf venv
+python3 -m venv venv
+source venv/bin/activate
+pip install -r ./requirements.txt
+./main.py "$@"
```
Saviq, line 7: Think we can get rid of this script, let the user decide how they want to supply it?

Saviq (changes requested):
Not an in-depth review from me, but here goes.

I'd rely on `python3-clang` coming from the repository, drop the run.sh. We can leave guarding the CMake targets on the existence of the Pyhton `clang` module.

argparse can deal with Paths and files natively:

https://docs.python.org/3/library/argparse.html
```python
parser.add_argument('dest_file', type=argparse.FileType('w', encoding='latin-1'))
parser.add_argument('datapath', type=pathlib.Path)
```

I think there's a couple unsolved requested changes (like filtering out `std:`.

mattkae:
> I think there's a couple unsolved requested changes (like filtering out `std:`.

Why would we want to filter out `std::`? We don't output `std` symbols. In the case mentioned, we _actually_ add a symbol to `std` (cause we are wild people).

src/miral/symbols.map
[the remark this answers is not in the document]
Saviq: @mattkae not according to @RAOF 👆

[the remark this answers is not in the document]
mattkae: This doesn't count then? https://github.com/canonical/mir/blob/f6296889ba81e8067d2f7b6d057f362daea36b73/include/miral/miral/window_info.h#L181

tools/symbols_map_generator.py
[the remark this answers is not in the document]
mattkae: Should I remove classes that we know contain no virtual members then?

RAOF:
Oooh! We should run CI with one of our different linker options, as they expose different warnings! 

I think it's `lld` that will warn if your linker script lists a symbol that isn't actually provided by the DSO. (I shall check tomorrow)

src/miral/symbols.map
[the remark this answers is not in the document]
AlanGriffiths: An inline function (or, as in this case, an inline explicit function template specialization) is in a grey area. 

Because the C++ standard doesn't talk about shared libraries there's nothing to guarantee where the implementation(s) live, it would be conforming to export the implementation and use it. I have seen cases where clang does this (with default destructors IIRC)

OTOH, not publishing the symbol can only cause problems if the address of the implementation is used. (Which would be weird for `std::swap<>`)

AlanGriffiths:
> it turns out that the clang-18 version in the Ubuntu repo is (somehow) different from the upstream version. If you install using `llvm.sh` (like CI does), then you will find success

Should we be using stuff from outside Ubuntu (i.e. `http://apt.llvm.org`) in CI?

mattkae:
> > it turns out that the clang-18 version in the Ubuntu repo is (somehow) different from the upstream version. If you install using `llvm.sh` (like CI does), then you will find success
> 
> Should we be using stuff from outside Ubuntu (i.e. `http://apt.llvm.org`) in CI?

After talking with Saviq, we deemed that this is the best that we can do for now. This is slightly unsafe though. I don't have a better solution.

src/miral/symbols.map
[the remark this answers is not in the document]
mattkae: So would it be fair to publish the symbol to stay on the safe side in this instance?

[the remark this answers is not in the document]
mattkae: Fun fact: there is no way via `libclang` to know if we have an explicitly inlined function :) But we can check if a function has a definition so that _should_ do the trick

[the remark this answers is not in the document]
mattkae: Done :+1:

tools/symbols_map_generator/main.py
```diff
@@ -75,6 +75,35 @@ def create_node_symbol_name(node: clang.cindex.Cursor) -> str:
         return node.spelling
 
 
+def is_virtual_class_or_struct(node: clang.cindex.Cursor):
```
AlanGriffiths, line 78: This name seems dubious.: It doesn't refer to virtual [base] classes, but to classes that have (or inherit) virtual methods.

Straw man: `has_virtual_member_functions()`

AlanGriffiths: Or even `has_vtable()` - since that's what the calling code wants to know

mattkae: `has_vtable` works for me

mattkae: Done :+1:

AlanGriffiths (approved):
OK

AlanGriffiths:
> I think it's `lld` that will warn if your linker script lists a symbol that isn't actually provided by the DSO. (I shall check tomorrow)

@RAOF did you want to come back on this before merging?

mattkae:
Going to fix up the commit history here too

mattkae:
Done cleaning up commit history

RAOF:
> > I think it's `lld` that will warn if your linker script lists a symbol that isn't actually provided by the DSO. (I shall check tomorrow)
> 
> @RAOF did you want to come back on this before merging?

So, the resolution here:
`lld` will warn if the version script lists a symbol that doesn't exist, *but* this doesn't apply to wildcards¹, so until we generate fully-qualified symbols for our `symbols.map` this is not going to be useful to us.

¹: Unless this would make an entire version stanza empty

RAOF (approved):
With that, I think everything's ready to hit the button!

PR #3311, opened 2024-04-02
docs: move Mir documentation from Discourse

Saviq:
The Mir documentation from https://mir-server.io/docs needs moving into readthedocs.

Once this is released we'll create redirects and edit the Discourse topics to include links to the new locations.


AlanGriffiths (approved):
I remain to be convinced that moving this content from "wiki" pages on discourse to the Mir git repository makes them better. It certainly introduces a barrier to maintaining them.

PR #3313, opened 2024-04-03
Fix positioning of satellite windows

hbatagelo:
When a satellite window is positioned, an extra vertical offset is added to its top-left position, making the vertical anchoring incorrect. The offset is added because `WindowInfo::needs_titlebar()` returns `true` when called from the floating window policy manager. This PR fixes that.

### DIFF
--- a/include/miral/miral/window_info.h
+++ b/include/miral/miral/window_info.h
@@ -47,10 +47,9 @@ struct WindowInfo
 
     bool is_visible() const;
 
-    /// \remark Since MirAL 5.0
-    /// \deprecated Use Surface::content_size()/Surface::window_size() instead
+    /// \deprecated Obsolete: Window::size() includes decorations
     /// @{
-    [[deprecated("use Surface::content_size()/Surface::window_size() instead")]]
+    [[deprecated("Obsolete: Window::size() includes decorations")]]
     static bool needs_titlebar(MirWindowType type);
     /// @}
 
--- a/include/miral/miral/window_info.h
+++ b/include/miral/miral/window_info.h
@@ -47,7 +47,12 @@ struct WindowInfo
 
     bool is_visible() const;
 
+    /// \remark Since MirAL 5.0
+    /// \deprecated Use Surface::content_size()/Surface::window_size() instead
+    /// @{
+    [[deprecated("use Surface::content_size()/Surface::window_size() instead")]]
     static bool needs_titlebar(MirWindowType type);
+    /// @}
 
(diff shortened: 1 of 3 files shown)


RAOF (dismissed):
I *think* the right change here is to make `needs_titlebar` unconditionally `return false`, but I'm not entirely sure why it *currently* doesn't make normal windows position incorrectly.

AlanGriffiths:
**I** think the correct change is to deprecate `needs_titlebar()` and strip out any use of it in our code.

 The `Surface::content_size()`/ `Surface::window_size()` distinction supersedes the earlier approach that used `needs_titlebar()` and [xdg-decoration](https://wayland.app/protocols/xdg-decoration-unstable-v1) support should leverage the newer approach.

hbatagelo:
Okay, I'll deprecate it then. What is the recommended approach to do that, besides removing the code that uses it? Add a comment, mark as `[[deprecated]]`, something else?

AlanGriffiths:
> Okay, I'll deprecate it then. What is the recommended approach to do that, besides removing the code that uses it? Add a comment, mark as `[[deprecated]]`, something else?

See 374cd363140f10adaeead7fba5d0a58ba0141270 for an example

include/miral/miral/window_info.h
```diff
@@ -47,7 +47,12 @@ struct WindowInfo
 
     bool is_visible() const;
 
+    /// \remark Since MirAL 5.0
+    /// \deprecated Use Surface::content_size()/Surface::window_size() instead
+    /// @{
+    [[deprecated("use Surface::content_size()/Surface::window_size() instead")]]
```
AlanGriffiths, line 53: Misleading and unhelpful: This far predates MirAL 5.0; and, `Surface::content_size()`/`Surface::window_size()` are not public APIs

```suggestion
    /// \deprecated Obsolete: Window::size() includes decorations
    /// @{
    [[deprecated("Obsolete: Window::size() includes decorations")]]
```

AlanGriffiths (approved):
LGTM. @RAOF?

include/miral/miral/window_info.h
hbatagelo: Much better. Thanks.

AlanGriffiths (approved):
OK

AlanGriffiths:
> I _think_ the right change here is to make `needs_titlebar` unconditionally `return false`, but I'm not entirely sure why it _currently_ doesn't make normal windows position incorrectly.

Probably because this only affects the miral-shell example - and that isn't widely used

PR #3316, opened 2024-04-04
Compositor & Scene: Simplify buffering

RAOF:
In various places we notify observers of how many buffers have changed when events happen.

Since we don't have any queues anymore, we only ever supply `1` as the number of buffers.

Remove some of the buffer counts and simplify `MultiThreadedCompositor` appropriately.

### DIFF
--- a/src/server/compositor/multi_threaded_compositor.cpp
+++ b/src/server/compositor/multi_threaded_compositor.cpp
@@ -33,7 +33,6 @@
 #include <thread>
 #include <chrono>
 #include <future>
-#include <semaphore>
 #include <boost/throw_exception.hpp>
 
 using namespace std::literals::chrono_literals;
@@ -60,7 +59,7 @@ public:
         compositor_factory{db_compositor_factory},
         group(group),
         scene(scene),
-        wakeup{0},
+        needs_wakeup{false},
         running{true},
         force_sleep{fixed_composite_delay},
         display_listener{display_listener},
@@ -121,7 +120,9 @@ public:
             while (running)
             {
                 /* Wait until compositing has been scheduled or we are stopped */
-                wakeup.acquire();
+                needs_wakeup.wait(false);
+                // We've been awoken; reset the trigger
+                needs_wakeup = false;
 
                 /*
                  * Check if we are running before compositing, since we may have
@@ -166,8 +167,8 @@ public:
 
     void schedule_compositing()
     {
-        wakeup.try_acquire();
-        wakeup.release();
+        needs_wakeup = true;
+        needs_wakeup.notify_one();
     }
 
     void schedule_compositing(geometry::Rectangle const& damage)
@@ -178,16 +179,16 @@ public:
 
         if (took_damage)
         {
-            wakeup.try_acquire();
-            wakeup.release();
+            needs_wakeup = true;
+            needs_wakeup.notify_one();
         }
     }
 
     void stop()
     {
         running = false;
-        wakeup.try_acquire();
-        wakeup.release();
+        needs_wakeup = true;
+        needs_wakeup.notify_one();
     }
 
     void wait_until_started()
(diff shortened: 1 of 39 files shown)


src/server/compositor/multi_threaded_compositor.cpp
```diff
@@ -254,18 +210,15 @@ class CompositingFunctor
     std::shared_ptr<mc::DisplayBufferCompositorFactory> const compositor_factory;
     mg::DisplaySyncGroup& group;
     std::shared_ptr<mc::Scene> const scene;
+    std::binary_semaphore wakeup;
     bool running;
```
AlanGriffiths, line 214: We're no longer guarding `running` with `run_mutex`. So we should make it `atomic<>`

AlanGriffiths (changes requested):
I can't help feeling that we will want this back at some point. But I agree, "not yet".

However, I'm not yet convinced that the logic replacing the `running` bool/mutex/condition_variable is right

src/server/compositor/multi_threaded_compositor.cpp
RAOF: I'm *moderately* certain that the semaphore provides a sufficient barrier, but just using an `atomic<bool>` will definitely make it *obviously* correct.

RAOF:
> I can't help feeling that we will want this back at some point. But I agree, "not yet".

Yeah, I've had a look at the proposed `wp-transaction`/`wp-commit-timing`/`wp-fifo` protocols. We will absolutely need *some* sort of queuing in the future, but I don't think what I'm removing here will be the right abstraction when we want to implement those anyway.

RAOF:
@AlanGriffiths I've switched to `std::atomic<bool> running`. Any other concerns?

AlanGriffiths:
> @AlanGriffiths I've switched to `std::atomic<bool> running`. Any other concerns?

Yeah, sorry I didn't get back to it yesterday - too many distractions. I need to get my head around the synchronization and convince myself it really works

src/server/compositor/multi_threaded_compositor.cpp
AlanGriffiths: I don't see that there was a sufficient barrier:

Thread A: `CompositingFunctor::operator()()` - acquires
Thread B: `CompositingFunctor::stop()` - running = false
Thread A: `CompositingFunctor::operator()()` - if (running) // undefined behavour: no guarantee running is one of true or false (even though it is on Intel, for example)

AlanGriffiths (changes requested):

src/server/compositor/multi_threaded_compositor.cpp
```diff
@@ -196,41 +164,30 @@ class CompositingFunctor
...
-        std::unique_lock lock{run_mutex};
-
-        if (num_frames > frames_scheduled)
-        {
-            frames_scheduled = num_frames;
-            lock.unlock();
-            run_cv.notify_one();
-        }
+        wakeup.try_acquire();
+        wakeup.release();
```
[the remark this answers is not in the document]
AlanGriffiths: **Preconditions**

Both `update >= 0` and `update <= max() - counter` are true, where counter is the value of the internal counter. 

https://en.cppreference.com/w/cpp/thread/counting_semaphore/release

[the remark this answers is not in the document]
RAOF: Oh, urgh, you're quite right. And it's worse; even *if* `try_acquire` doesn't spuriously fail, there's a race between the `try_acquire` and `release` that could result in two threads `release`ing at the same time.

I *know* there's an OS primitive for “sleep until this value changes”, damnit. Why doesn't C++ expose that‽

AlanGriffiths (approved):
OK. Cool! I had missed `atomic<>` growing `wait()`/`notify...()`. I like it!

PR #3317, opened 2024-04-05
ThreadPoolExecutor: Drop std::binary_semaphore polyfill.

RAOF:
All targets we compile for should now support enough C++20 stdlib features to remove the need for us to provide a fallback.

### DIFF
--- a/src/common/thread_pool_executor.cpp
+++ b/src/common/thread_pool_executor.cpp
@@ -24,62 +24,6 @@
 #include <list>
 #include <future>
 
-#include <version>
-
-#ifdef __cpp_lib_semaphore
-#include <semaphore>
-#else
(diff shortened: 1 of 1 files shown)


src/common/thread_pool_executor.cpp
```diff
@@ -24,62 +24,6 @@
...
-        return false;
-    }
-private:
-    std::mutex mutex;
-    std::condition_variable cv;
-    bool raised;
-};
-}
-#endif
-
```
AlanGriffiths, line 82: I can see this compiles without the `<semaphore>` header, but should we be relying on that?

RAOF: Oops! No, we shouldn't.

AlanGriffiths (approved):
Cool!

Issue #3318, opened 2024-04-05
FTBFS since google-glog v0.7.0


JamiKettunen:
Reproducible by building [v2.16.4](https://github.com/canonical/mir/releases/tag/v2.16.4) on e.g. Arch Linux:
```
[ 98%] Building CXX object examples/mir_demo_server/CMakeFiles/mir_demo_server.dir/server_example_test_client.cpp.o    
In file included from .../mir/src/mir-2.16.4/examples/mir_demo_server/glog_logger.cpp:19:    
/usr/include/glog/logging.h:60:4: error: #error <glog/logging.h> was not included correctly. See the documention for how to consume the library.    
   60 | #  error <glog/logging.h> was not included correctly. See the documention for how to consume the library.    
      |    ^~~~~    
In file included from /usr/include/glog/logging.h:63:    
/usr/include/glog/flags.h:45:4: error: #error <glog/flags.h> was not included correctly. See the documention for how to consume the library.    
   45 | #  error <glog/flags.h> was not included correctly. See the documention for how to consume the library.    
      |    ^~~~~    
In file included from /usr/include/glog/logging.h:77:    
/usr/include/glog/log_severity.h:38:4: error: #error <glog/log_severity.h> was not included correctly. See the documention for how to consume the library.    
   38 | #  error <glog/log_severity.h> was not included correctly. See the documention for how to consume the library.    
      |    ^~~~~    
In file included from /usr/include/glog/logging.h:78:    
/usr/include/glog/vlog_is_on.h:71:4: error: #error <glog/vlog_is_on.h> was not included correctly. See the documention for how to consume the library.    
   71 | #  error <glog/vlog_is_on.h> was not included correctly. See the documention for how to consume the library.    
      |    ^~~~~
```

Saviq:
Hi @JamiKettunen, can you confirm #3320 fixes this for you?

Saviq:
Could you maybe contribute CI steps similar to what we have for Alpine and Fedora?
Ref.
https://github.com/canonical/mir/blob/main/spread.yaml
https://github.com/canonical/mir/tree/main/spread/build

PR #3320, opened 2024-04-05
FTBFS since google-glog v0.7.0

AlanGriffiths:
Fixes: #3318


RAOF (approved):
I assume this works (I don't have glog 0.7.0 handy), but checking [upstream](https://github.com/google/glog/releases/tag/v0.7.0) they describe the pkg-config files as “undocumented and unsupported”, so maybe we want to do #3321 instead?

AlanGriffiths:
Closing in favour of #3330

PR #3321, opened 2024-04-08
CMake: Use GLog's CMake integration

RAOF:
Upstream describe their pkg-config support as “Undocumented and untested”, and it doesn't work for GLog 0.7.0.

Fixes: #3318


AlanGriffiths:
Looking at the build failures this approach isn't better supported across distros than pkg-config. :disappointed: 

(I'm beginning to think that maintaining the `glog_logger.cpp` example is too much work for the value it offers)

AlanGriffiths:
Closing in favour of #3330

PR #3322, opened 2024-04-08
compositor/Stream: Remove `buffers_ready_for_compositor`

RAOF:
The only internal user of this has been removed. QtMir is an *external* user of this, but:
* Doesn't actually *need* this behaviour, and
* Needs significant work in this area anyway to update for 2.16's Platform changes


RAOF:
NB: This is a miroil API & ABI change; that has not been done in this PR.

AlanGriffiths:
> NB: This is a miroil API & ABI change; that has not been done in this PR.

Good! (It was already done when we dropped mircookie)

AlanGriffiths (approved):
We said "A", "B" follows

PR #3324, opened 2024-04-09
Add floating point options

AlanGriffiths:
We already have some floating point "input configuration" options in our examples. This provides a little more support internally that I used in hacking that code about.


mattkae (approved):
Cool :+1: Is there anywhere that this is used yet?

PR #3329, opened 2024-04-12
msd::Renderer: Use freetype error return value

RAOF:
Clang has started (correctly) identifying that we're not using the error variable we're declaring. We might as well!


AlanGriffiths (approved):
Might be useful, and can't hurt

PR #3330, opened 2024-04-12
Drop the glog example - nobody uses it

AlanGriffiths:
Fixes: #3318


Saviq (approved):
Ugh? How was this in platform's symbols.map…?

PR #3331, opened 2024-04-15
common/Signal: New worker-thread synchronisaton primitive.

RAOF:
This is *very* much like `std::binary_semaphore`, but doesn't have the preconditon that the signal must not be raised before attempting to raise it.

Use it to fix the UB (on shutdown) in `mir::ThreadPoolExecutor`

### DIFF
--- a/include/common/mir/signal.h
+++ b/include/common/mir/signal.h
@@ -68,7 +68,7 @@ public:
      */
     void wait();
 private:
-    std::atomic<bool> flag;
+    std::atomic<bool> raised;
 };
 }
 
--- a/include/common/mir/signal.h
+++ b/include/common/mir/signal.h
@@ -63,7 +63,7 @@ public:
     /**
      * Wait for the signal to be raised
      *
-     * This synchronises-wtih a previous call to `raise()`, and
+     * This synchronises-with a previous call to `raise()`, and
      * then resets the signal.
      */
     void wait();
--- a/include/common/mir/signal.h
+++ b/include/common/mir/signal.h
@@ -1,0 +1,75 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_SIGNAL_H_
+#define MIR_SIGNAL_H_
+
+#include <atomic>
+
+namespace mir
+{
+/**
+ * A thread-synchronisation barrier.
+ *
+ * The expected usage is that one thread sits in `wait()` until
+ * signalled from another thread with `raise()`.
+ *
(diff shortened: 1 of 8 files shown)


RAOF:
(This should also be used in MultiThreadedCompositor)

include/common/mir/signal.h
```diff
@@ -0,0 +1,75 @@
...
+ *    happens-after the most recent return from `wait()`.
+ *
+ * The primary use-case for such a barrier is to signal a
+ * worker thread to run the next iteration of work.
+ *
+ * This is very similar to a `std::binary_semaphore` but
+ * without the precondition that `raise()` is not called
+ * if the Signal is already raised.
+ */
+class Signal
```
mattkae, line 45: Could we write some unit tests for this class?

include/common/mir/signal.h
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * Two unsynchronised calls to `raise()` may result in either one
+     * or two calls to `wait()` being unblocked, depending on timing.
+     */
+    void raise();
+
+    /**
+     * Wait for the signal to be raised
+     *
+     * This synchronises-wtih a previous call to `raise()`, and
```
mattkae, line 66:
```suggestion
     * This synchronises-with a previous call to `raise()`, and
```

mattkae (changes requested):
A small request for tests, but overall it makes sense. It is unfortunate that 1 raise != 1 wait, but so long as its explicit, I am a happy person

mattkae (approved):
Good by me!

include/common/mir/signal.h
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Wait for the signal to be raised
+     *
+     * This synchronises-with a previous call to `raise()`, and
+     * then resets the signal.
+     */
+    void wait();
+private:
+    std::atomic<bool> flag;
```
AlanGriffiths, line 71: Isn't "raised" a better name?

src/common/signal.cpp
```diff
@@ -0,0 +1,34 @@
...
+void mir::Signal::raise()
+{
+    flag = true;
+    flag.notify_all();
+}
+
+void mir::Signal::wait()
+{
+    flag.wait(false);
+    flag = false;
```
AlanGriffiths, line 33: Isn't this racy? `raise()` can be called between these lines.

AlanGriffiths (changes requested):
There seems to be something inherently hard in reusing signal types. Using a lock, cv and flag is heavier, but easy to do correctly.

src/common/signal.cpp
RAOF: It *is* racy, but the race is benign for our purposes, and is documented by 
> ```* Two unsynchronised calls to `raise()` may result in either one```
> ```* or two calls to `wait()` being unblocked, depending on timing.```

This is also not a race that's fixed by using a mutex + condition variable; two unsynchronised `raise()`s will still result in either one or two `wait()`s being unblocked, depending on whether both `raise()`s grab the mutex before the `wait()` does.

include/common/mir/signal.h
RAOF: Yes; fixed.

RAOF:
I'm ok with flipping this to a mutex/condition variable if that's more clearly correct.

(I'm a little bit irrationally annoyed that C++ doesn't expose the common “wait on this memory address” OS functionality, but I can cope :wink:)

src/common/signal.cpp
AlanGriffiths: Ack

AlanGriffiths (approved):
I think this design could lead to unexpected (albeit documented) behaviour with multiple threads calling `wait()`. But it the cases it is used for here don't go there

PR #3336, opened 2024-04-18
actions: add step for checking miral and miroil symbols map

mattkae:


.github/workflows/symbols-check.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          sudo apt-get build-dep ./
+
+      - name: Configure
+        run: >
+          cmake -B build ${{ github.workspace }}
+
+      - name: Install clang
+        run: |
+          wget https://apt.llvm.org/llvm.sh
```
RAOF, line 36: Hm. What are the security concerns with this `curl | sudo sh -`?

It's in the GitHub runners infrastructure, so infrastructure escape shouldn't be a problem, but what does it have access to *here*?

I think the answer is that it's got access to the GitHub secrets? Which means it has access to the release signing key, and the signing key to upload to the PPA?

I don't think this should be a blocker here, but maybe we should record this somewhere?

RAOF (approved):
LGTM. Approving, but not merging, to give the security question a little more airing.

tools/symbols_map_generator/main.py
```diff
@@ -466,8 +479,10 @@ def main():
 
                 f.write(output_str)
             
-
-    exit(0)
+    if has_changed_symbols:
+        exit(1)
+    else:
+        exit(0)
```
Saviq, line 485:
```suggestion
    if args.diff and has_changed_symbols:
        exit(1)
```

.github/workflows/symbols-check.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: >
+          cmake -B build ${{ github.workspace }}
+
+      - name: Install clang
+        run: |
+          wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
+          sudo add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy main"
+          sudo apt install libclang1-19
+          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH=/usr/lib/llvm-19/lib/libclang-19.so.1" >> $GITHUB_ENV
+          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib" >> $GITHUB_ENV
```
Saviq, line 40: Maybe move this up above `dependencies`, as this is a cheaper step?

.github/workflows/symbols-check.yml
```diff
@@ -0,0 +1,48 @@
...
+  Run:
+    runs-on: ubuntu-latest
+
+    timeout-minutes: 10
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          # So we can determine the merge base
+          fetch-depth: 0
```
Saviq, line 24: Don't need this.

```suggestion
```

.github/workflows/symbols-check.yml
```diff
@@ -0,0 +1,48 @@
...
+          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH=/usr/lib/llvm-19/lib/libclang-19.so.1" >> $GITHUB_ENV
+          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib" >> $GITHUB_ENV
+
+      - name: Check miral symbols
+        run: |
+          cmake --build build --target check-miral-symbols-map
+
+      - name: Check miroil symbols
+        run: |
+          cmake --build build --target check-miroil-symbols-map
```
Saviq, line 48: Let's exit late.

```suggestion
      - name: Check symbols
        run: |
          RET=0
          cmake --build build --target check-miral-symbols-map || RET=$?
          cmake --build build --target check-miroil-symbols-map || RET=$?
          exit $RET
```

Saviq (changes requested):
Some nits inline

.github/workflows/symbols-check.yml
```diff
@@ -0,0 +1,41 @@
...
+    runs-on: ubuntu-latest
+
+    timeout-minutes: 10
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get build-dep ./
```
Saviq, line 25: This is the expensive step :)
```suggestion
```

.github/workflows/symbols-check.yml
```diff
@@ -0,0 +1,41 @@
...
+          wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
+          sudo add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy main"
+          sudo apt install libclang1-19
+          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH=/usr/lib/llvm-19/lib/libclang-19.so.1" >> $GITHUB_ENV
+          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib" >> $GITHUB_ENV
```
Saviq, line 30:
```suggestion
          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib" >> $GITHUB_ENV
          sudo apt-get build-dep ./
```

Saviq (changes requested):

.github/workflows/symbols-check.yml
mattkae: I want the build deps so that 3rd party libraries resolve in clang, right?

Saviq (approved):
Sure.

PR #3338, opened 2024-04-19
experiment: providing a gateway for writing your own renderer through MirAL

mattkae:
**WARNING**: Highly experimental, but it works

## What's new?
- Users can now provide `CustomRenderer` to `miral` which gives them the rites to entirely replace the existing renderer
- Although this might not be 100% percent what we want, **I would be VERY interested in seeing something like this go into an "internal" headers somewhere** as it would enable a lot of interesting things


include/miral/miral/custom_renderer.h
```diff
@@ -0,0 +1,61 @@
...
+namespace miral
+{
+typedef std::function<std::unique_ptr<mir::renderer::Renderer>
+    (std::unique_ptr<mir::graphics::gl::OutputSurface>,
+     std::shared_ptr<mir::graphics::GLRenderingProvider>)> CreateRenderer;
+
+class CustomRenderer
+{
+public:
+    explicit CustomRenderer(CreateRenderer const& renderer);
```
AlanGriffiths, line 54:
```suggestion
    using Builder = std::function<
        std::unique_ptr<mir::renderer::Renderer>(
            std::unique_ptr<mir::graphics::gl::OutputSurface>, std::shared_ptr<mir::graphics::GLRenderingProvider>)
    >;

    explicit CustomRenderer(Builder&& renderer);
```

include/miral/miral/custom_renderer.h
```diff
@@ -0,0 +1,61 @@
...
+    void operator()(mir::Server& server) const;
+private:
+    std::shared_ptr<mir::renderer::RendererFactory> factory;
```
AlanGriffiths, line 57:
```suggestion
    struct Self;
    std::shared_ptr<Self> self;
```

AlanGriffiths (changes requested):
I'd tweak the MirAL API slightly to be more consistent with other MirAL APIs, but I don't think there's anything problematic about the semantics.

However, my question is whether it is possible to implement the `CustomRender::Builder` with the APIs we make available? (I suspect that might need "internal" APIs)

mattkae:
> I'd tweak the MirAL API slightly to be more consistent with other MirAL APIs, but I don't think there's anything problematic about the semantics.
> 
> However, my question is whether it is possible to implement the `CustomRender::Builder` with the APIs we make available? (I suspect that might need "internal" APIs)

I was able to copy over the entire GL renderer into `miracle-wm` with a lot of success: https://github.com/mattkae/miracle-wm/pull/103

It is a really rough copy, but I only required `tesselation_helpers.h` and `primitive.h` to be copied over. The rest was already exposed to me in `MIRRENDERER`. Those two headers aren't very fundamental in my opinion, so I'd say we're good with what we have now.

src/miral/custom_renderer.cpp
```diff
@@ -0,0 +1,71 @@
...
+    [[nodiscard]] auto create_renderer_for(
+        std::unique_ptr<mir::graphics::gl::OutputSurface> output_surface,
+        std::shared_ptr<mir::graphics::GLRenderingProvider> gl_provider) const
+        -> std::unique_ptr<mir::renderer::Renderer> override
+    {
+        return renderer(std::move(output_surface), std::move(gl_provider));
+    }
+
+private:
+    miral::CustomRenderer::Builder const& renderer;
```
AlanGriffiths, line 42: Should this really be a reference?

src/miral/custom_renderer.cpp
```diff
@@ -0,0 +1,71 @@
...
+};
+}
+
+struct miral::CustomRenderer::Self
+{
+    Self(Builder const& renderer, std::shared_ptr<mir::graphics::GLConfig> const& config)
```
AlanGriffiths, line 48:
```suggestion
    Self(Builder&& renderer, std::shared_ptr<mir::graphics::GLConfig> const& config)
```

src/miral/custom_renderer.cpp
```diff
@@ -0,0 +1,71 @@
...
+    {
+    }
+
+    std::shared_ptr<mir::renderer::RendererFactory> factory;
+    std::shared_ptr<mir::graphics::GLConfig> config;
+};
+
+miral::CustomRenderer::CustomRenderer(
+    Builder&& renderer, std::shared_ptr<mir::graphics::GLConfig> const& config)
+    : self{std::make_shared<Self>(renderer, config)}
```
AlanGriffiths, line 60:
```suggestion
    : self{std::make_shared<Self>(std::move(renderer), config)}
```

AlanGriffiths (changes requested):
I think `miral::CustomRenderer` should do one thing, not two

src/miral/custom_renderer.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+struct miral::CustomRenderer::Self
+{
+    explicit Self(Builder&& renderer)
+        : factory(std::make_shared<RendererFactory>(std::move(renderer)))
+    {
+    }
+
+    std::shared_ptr<mir::renderer::RendererFactory> factory;
+};
```
AlanGriffiths, line 54: This would be simpler (fewer classes, less forwarding code) if `miral::CustomRenderer::Self` IsA `mir::renderer::RendererFactory`, not HasA. Vis:

```suggestion
struct miral::CustomRenderer::Self : public mir::renderer::RendererFactory
{
    explicit Self(Builder&& renderer)
        : factory(std::move(renderer))
    {
    }

    [[nodiscard]] auto create_renderer_for(
        std::unique_ptr<mir::graphics::gl::OutputSurface> output_surface,
        std::shared_ptr<mir::graphics::GLRenderingProvider> gl_provider) const
        -> std::unique_ptr<mir::renderer::Renderer> override
    {
        return renderer(std::move(output_surface), std::move(gl_provider));
    }

private:
    miral::CustomRenderer::Builder const renderer;
};
```

AlanGriffiths: Alternatively, `Self::factory` should be const.

AlanGriffiths (approved):
IMO the code could be a little clearer. But otherwise fine.

(Approving, but not merging to allow tweak)

src/miral/custom_renderer.cpp
mattkae: I agree with that change!

src/miral/custom_renderer.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+miral::CustomRenderer::CustomRenderer(Builder&& renderer)
+    : self{std::make_shared<Self>(std::move(renderer))}
+{
+}
+
+void miral::CustomRenderer::operator()(mir::Server &server) const
+{
+    std::function<std::shared_ptr<mir::renderer::RendererFactory>()> builder = [&]() { return self; };
```
AlanGriffiths, line 47: It is probably safer to take self by value.
```suggestion
    std::function<std::shared_ptr<mir::renderer::RendererFactory>()> builder = [self=self] { return self; };
```

AlanGriffiths (approved):
More nits

PR #3344, opened 2024-04-24
Platforms do not need to depend on /src/include/common

AlanGriffiths:
This breaks the dependency our platform implementations had on internal headers


mattkae (approved):
Makes sense to me!

PR #3349, opened 2024-04-25
experimental: it would be useful if the `Renderable` and the `Surface` include the same userdata as the `WindowInfo`

mattkae:
## What's new?
- When I'm rendering, I want a bit  more context as to _what_ exactly I am rendering
- This PR introduces the idea of `Renderable::userdata` which one can access and cast in the `Renderer` to gain some more context about what's being rendered

Let me know if you believe there to be a better solution than this.

## Other Possible Solutions
- Now that I think of it, I _could_ also just generate a list of external `Renderable`s that I want rendered, and then I wouldn't need this PR. Let me explore how viable that is...


mattkae:
Rethinking this one..

mattkae:
No I think I like this one. It makes since you'll want to know the context of a `Renderable` so that you can inject some other `Renderable` at the right position.

AlanGriffiths:
I'm not convinced this is a good approach.

I'm guessing it is modelled on the `userdata()` property of `miral::WindowInfo`? Which allows compositor specific information to be associated with Windows. For example, it is used to manage workspaces.

The situation here is different, but you are plumbing the _same_ `userdata()` through to the surface and then to any associated renderables. I really don't think that the renderer cares about the workspace management data.

I think there may be some specific concerns that the renderer has, but I don't think this answers them. You mentioned, for instance, knowing whether there was an associated, window with focus. Adding a `has_focus()` property probably addresses that better.

mattkae:
> I think there may be some specific concerns that the renderer has, but I don't think this answers them. You mentioned, for instance, knowing whether there was an associated, window with focus. Adding a `has_focus()` property probably addresses that better.

I think that my real wish would be for their to be some connection from the `Renderable` to the `Window`, meaning I would need access to the surface via the `Renderable`. My purpose for doing this is if we need more data than just `has_focus()`, e.g. "Is this a tiled window AND does it have focus?" or something like that.

For that reason, I think it would be a good idea for the `Renderable` to provide access back to the whole `Surface`, if possible. WDYT?

AlanGriffiths:
> For that reason, I think it would be a good idea for the `Renderable` to provide access back to the whole `Surface`, if possible. WDYT?

I think that `auto Renderable::surface_if_any() const -> std::optional<std::shared_ptr<Surface>>` would give you what you want. But that is also a leaky abstraction: it exposes a lot of additional functionality that is no business of a renderer. 

It seems better for the `Renderable` interface to support the information needed by the renderer, even if that means the `Renderable` implementation has to keep a backpointer to the `Surface`. Do you have an example of information isn't available with that approach?

mattkae:
> It seems better for the `Renderable` interface to support the information needed by the renderer, even if that means the `Renderable` implementation has to keep a backpointer to the `Surface`. Do you have an example of information isn't available with that approach?

I can't think of anything that I wouldn't be able to resolve given the `Surface`. From the `Surface`, I can resolve the `Window`, which means that I can get `WindowInfo` and `userdata()` if I want. At that point, I know exactly the context of the surface that I'm dealing with. This sounds like a plausible solution to me.

AlanGriffiths:
> > It seems better for the `Renderable` interface to support the information needed by the renderer, even if that means the `Renderable` implementation has to keep a backpointer to the `Surface`. Do you have an example of information isn't available with that approach?
> 
> I can't think of anything that I wouldn't be able to resolve given the `Surface`. From the `Surface`, I can resolve the `Window`, which means that I can get `WindowInfo` and `userdata()` if I want. At that point, I know exactly the context of the surface that I'm dealing with. This sounds like a plausible solution to me.

Sorry, I didn't realise I was unclear but you not answering the paragraph you quote.

I said that `surface_if_any()` is a bad idea, and asked if there were queries that extending the `Renderable` couldn't address

mattkae:
> > > It seems better for the `Renderable` interface to support the information needed by the renderer, even if that means the `Renderable` implementation has to keep a backpointer to the `Surface`. Do you have an example of information isn't available with that approach?
> > 
> > 
> > I can't think of anything that I wouldn't be able to resolve given the `Surface`. From the `Surface`, I can resolve the `Window`, which means that I can get `WindowInfo` and `userdata()` if I want. At that point, I know exactly the context of the surface that I'm dealing with. This sounds like a plausible solution to me.
> 
> Sorry, I didn't realise I was unclear but you are not answering the paragraph you quote.
> 
> I said that `surface_if_any()` is a bad idea, and asked if there were queries that extending the `Renderable` couldn't address

Oh, I misunderstood you. An example query would be a window manager specific one, e.g.: is this surface a tile in my window manager or a floating window?

PR #3351, opened 2024-04-25
Improved debug logging around modifier sends

mattkae:
## What's new?
- This will  give us some logs for when modifiers are sent so that we can compare when we expect them to be sent

### DIFF
--- a/src/server/frontend_wayland/keyboard_helper.cpp
+++ b/src/server/frontend_wayland/keyboard_helper.cpp
@@ -21,6 +21,7 @@
 #include "mir/events/keyboard_event.h"
 #include "mir/input/seat.h"
 #include "mir/fatal.h"
+#include "mir/log.h"
 
 #include <cstring> // memcpy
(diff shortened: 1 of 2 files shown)


src/server/frontend_wayland/keyboard_helper.cpp
```diff
@@ -148,6 +150,20 @@ void mf::KeyboardHelper::set_keymap(std::shared_ptr<mi::Keymap> const& new_keyma
 
 void mf::KeyboardHelper::set_modifiers(MirXkbModifiers const& new_modifiers)
 {
+    mir::log_debug(
+    "Modifiers are being set from: depressed=%u, latched=%u, locked=%u, effective_layout=%u",
```
AlanGriffiths, line 154: In my experience it is better to log bitmasks as hex values

AlanGriffiths (changes requested):

RAOF (approved):
LGTM

PR #3355, opened 2024-04-26
feature: provide surface access via the Renderable

mattkae:
Alternative to https://github.com/canonical/mir/pull/3349

## What's new?
- Now you can access the `Surface` via the `Renderable`

### DIFF
--- a/include/platform/mir/graphics/renderable.h
+++ b/include/platform/mir/graphics/renderable.h
@@ -25,6 +25,9 @@
 
 namespace mir
 {
+
+namespace scene { class Surface; }
+
 namespace graphics
 {
 
@@ -69,6 +72,9 @@ public:
     virtual glm::mat4 transformation() const = 0;
 
     virtual bool shaped() const = 0;  // meaning the pixel format has alpha
+
+    virtual auto surface_if_any() const
+        -> std::optional<mir::scene::Surface const*> = 0;
 protected:
     Renderable() = default;
     Renderable(Renderable const&) = delete;
--- a/src/server/graphics/software_cursor.cpp
+++ b/src/server/graphics/software_cursor.cpp
@@ -100,6 +100,12 @@ public:
         return true;
     }
 
+    auto surface_if_any() const
+        -> std::optional<mir::scene::Surface const*> override
+    {
+        return std::nullopt;
+    }
+
     void move_to(geom::Point new_position)
     {
         std::lock_guard lock{position_mutex};
--- a/src/server/input/touchspot_controller.cpp
+++ b/src/server/input/touchspot_controller.cpp
@@ -83,7 +83,13 @@ public:
         return true;
     }
 
-// TouchspotRenderable    
+    auto surface_if_any() const
+        -> std::optional<mir::scene::Surface const*> override
+    {
+        return std::nullopt;
+    }
+
+    // TouchspotRenderable
     void move_center_to(geom::Point pos)
     {
         std::lock_guard lg(guard);
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -697,14 +697,16 @@ public:
         std::optional<geom::Rectangle> const& clip_area,
         glm::mat4 const& transform,
         float alpha,
-        mg::Renderable::ID id)
+        mg::Renderable::ID id,
+        ms::Surface const* surface)
     : underlying_buffer_stream{stream},
       compositor_id{compositor_id},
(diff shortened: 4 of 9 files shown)


RAOF:
So, I'm torn on this whole enterprise.

On one hand, this is the wrong spot to do this. Particularly, there's no guarantee that the compositor is actually going to *call* your `Renderer`; the `MultiThreadedCompositor` only calls into the renderer when it hasn't been able to place everything on a hardware plane¹.

On the other hand, even the `Compositor` only has access to a list of `SceneElements`, which doesn't include what you're after anyway.

On the third hand, the eternal scenegraph discussion. :woman_shrugging: 

¹: Which, to be fair, it won't be able to *at the moment*, as that's not properly implemented.

include/platform/mir/graphics/renderable.h
```diff
@@ -69,6 +72,9 @@ class Renderable
     virtual glm::mat4 transformation() const = 0;
 
     virtual bool shaped() const = 0;  // meaning the pixel format has alpha
+
+    virtual auto surface_if_any() const
+        -> std::optional<mir::scene::Surface const*> = 0;
```
RAOF, line 77: `std::optional<pointer const*>` is not *super* different to plain `pointer const*`; I could go either way on this; we should probably have an actual style guide decision on this.

RAOF:
To be clear: this seems like a pragmatic solution to get *something* working; we've not prioritised getting a *nice* API for this so waiting for that is unreasonable.

include/platform/mir/graphics/renderable.h
AlanGriffiths: So long as we adhere to the Core Guidelines:

> R.2: In interfaces, use raw pointers to denote individual objects (only)
> R.3: A raw pointer (a T*) is non-owning

I'm also fine with a named `Foo cont*` provided that there's a clear indication (like `_if_any`) when `nullptr` is possible.

For reference, Sophie was keen on `std::optional<>`, so there's some of that around the codebase.

src/server/graphics/software_cursor.cpp
```diff
@@ -100,6 +100,12 @@ class mg::detail::CursorRenderable : public mg::Renderable
         return true;
     }
 
+    auto surface_if_any() const
+        -> std::optional<mir::scene::Surface const*> override
+    {
+        return {};
```
AlanGriffiths, line 106:
```suggestion
        return std::nullopt;
```

AlanGriffiths (approved):
There are not problems in this code, but thinking on the usecase you mention...

Keeping pointer values in any sort of lookup needs care: a pointer is only valid until the memory it references is deallocated (typically part of deleting the referenced object).  After that just _accessing_ (not dereferencing) the pointer is undefined behaviour.

(Not merging pending discussion on `std::optional<>`)

mattkae:
> So, I'm torn on this whole enterprise.
> 
> On one hand, this is the wrong spot to do this. Particularly, there's no guarantee that the compositor is actually going to _call_ your `Renderer`; the `MultiThreadedCompositor` only calls into the renderer when it hasn't been able to place everything on a hardware plane¹.

I was not aware that we wanted to do this! Would this mean that instead of going through the GL renderer, surfaces (somehow) just get blipped onto a hardware plane?

> On the other hand, even the `Compositor` only has access to a list of `SceneElements`, which doesn't include what you're after anyway.

:+1: 

> On the third hand, the eternal scenegraph discussion. 🤷‍♀️

Yeah this _might_ be the better solution, but I am afraid I would wait around a while for it, huh?

include/platform/mir/graphics/renderable.h
mattkae: I like `std::optional` personally because it shows intent. We don't `null`-check every `shared_ptr` in the repo, so I would assume that when we're accessing _any_ pointer, we can assume that it is probably set.

mattkae:
> There are not problems in this code, but thinking on the usecase you mention...
> 
> Keeping pointer values in any sort of lookup needs care: a pointer is only valid until the memory it references is deallocated (typically part of deleting the referenced object). After that just _accessing_ (not dereferencing) the pointer is undefined behaviour.
> 
> (Not merging pending discussion on `std::optional<>`)

For my use case, I won't be storing these references, but rather using them when they come to me to resolve back to the `Window` and its userdata. The pointer won't be invalidated by the time the `render` happens, so I should be okay on that front

RAOF (approved):
Yup. As I say, a pragmatic hack!

PR #3359, opened 2024-04-30
feature: moving WindowManagerToolsImplementation to a public-facing miral folder

mattkae:
## Why?
- People might want to implement their own version of the Window Manager, especially for testing
- We could alternatively publish this in a `MIRAL_INTERNAL` API, but it feels fairly stable. I could be wrong though!


AlanGriffiths:
tl;dr: I'm not convinced this is the right solution to the right problem.

This class was clearly not intended as part of the API, and the intention was that window management is customised by the window management policy (and that the Window Manager suggests provides defaults). If you feel that isn't working then there may be better solutions.

If testing is the unfulfilled requirement, then I think providing a test double from Mir is better than expecting every downstream to roll their own.

> * it feels fairly stable. I could be wrong though!

Stable? **You** have changed it 15 times in the last year!

AlanGriffiths:
Closing as this feels dead. @mattkae do you still intend to pursue this approach?

mattkae:
> Closing as this feels dead. @mattkae do you still intend to pursue this approach?

Yup, I'm okay with that. I think a test harness is the real solution

PR #3362, opened 2024-05-01
Revert spurious bump to mirserver soname

AlanGriffiths:
In 2.16 mirserver was 59, in 2.17 it should be 60


mattkae (approved):
Whoops! Good catch.

PR #3366, opened 2024-05-01
bugfix: mircommon-internal-dev was erroneously including some libs

mattkae:
My local version of `pkgconfig` doesn't complain about this, but it is most definitely an error!


AlanGriffiths (approved):
Looks good!

Issue #3300, opened 2024-03-28
`std::exception::what: Event of unknown device received` when driving with virtual input


Saviq:
We seem to be racing somewhere, got this when running the drag_drop tests on a Pi4:
```
│ [2024-03-28 00:41:28.178911] < CRITICAL! > mirserver: terminate_with_current_exception(): ./src/server/input/seat_input_device_tracker.cpp(135): Throw in function void mir::input::SeatInputDeviceTracker::update_seat_properties(const MirInputEvent*)
│ Dynamic exception type: boost::wrapexcept<std::logic_error>
│ std::exception::what: Event of unknown device received
│ 
│ [2024-03-28 00:41:28.179448] < - debug - > mirserver: Handling Terminated from pid=19872
│ ERROR: ./src/server/input/seat_input_device_tracker.cpp(135): Throw in function void mir::input::SeatInputDeviceTracker::update_seat_properties(const MirInputEvent*)
│ Dynamic exception type: boost::wrapexcept<std::logic_error>
│ std::exception::what: Event of unknown device received
```

mattkae:
My guess is the following happens:
- Device gets added
- Keyboard event starts
- Device gets removed
- Keyboard event gets processed
- :bomb:

PR #3367, opened 2024-05-01
(#3300) bugfix for a race condition in VirtualPointerV1

mattkae:
fixes https://github.com/canonical/mir/issues/3300

See [this comment](https://github.com/canonical/mir/issues/3300#issuecomment-2088897189) to understand my rationale here. I think this should fix it, but I don't have a solid repo on my end (and I don't see this in CI anywhere). All other code paths seemed to be fine.

### DIFF
--- a/src/server/frontend_wayland/virtual_pointer_v1.cpp
+++ b/src/server/frontend_wayland/virtual_pointer_v1.cpp
@@ -202,7 +202,12 @@ mf::VirtualPointerV1::VirtualPointerV1(
 
 mf::VirtualPointerV1::~VirtualPointerV1()
 {
-    ctx->device_registry->remove_device(pointer_device);
+    // Destruction must happen on the linearising_executor to prevent races with create_and_dispatch_event
+    mir::linearising_executor.spawn(
+        [pointer_device=pointer_device, device_registry=ctx->device_registry]()
+    {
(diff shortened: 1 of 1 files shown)


src/server/frontend_wayland/virtual_pointer_v1.cpp
```diff
@@ -202,7 +202,12 @@ mf::VirtualPointerV1::VirtualPointerV1(
 
 mf::VirtualPointerV1::~VirtualPointerV1()
 {
-    ctx->device_registry->remove_device(pointer_device);
+    // Destruction must happen on the linearising_executor to prevent races with create_and_dispatch_event
```
RAOF, line 205: Ah! This is actually racing with `frame()` (which is what dispatches something to the `linearising_executor`)!

RAOF (approved):
I'm not sure that the rationale *entirely* holds up, but this will *definitely* stop the virtual input device being removed while a sequence of events are being emitted on it.

src/server/frontend_wayland/virtual_pointer_v1.cpp
mattkae: Yup! That was the only race that I saw here from `VirtualPointer`'s perspective. I am also unsure if the rationale holds up, but I'll be curious to see if this prevents further breaking

PR #3368, opened 2024-05-02
Drop the unnecessary indirection of `mc::BufferStreamFactory`

RAOF:
We only create a `mc::Stream` in like 1 place, the implementation of `BufferStreamFactory` is trivial, and I'm about to remove *all* the parameters from the `mc::Stream` constructor anyway.

There's no need to keep this indirection (and the tests that check that we correctly call through that indirection)


src/server/compositor/default_configuration.cpp
```diff
@@ -33,16 +32,6 @@ namespace mc = mir::compositor;
 namespace ms = mir::scene;
 namespace mg = mir::graphics;
 
-std::shared_ptr<ms::BufferStreamFactory>
-mir::DefaultServerConfiguration::the_buffer_stream_factory()
```
mattkae, line 37: Do we need to do anything in the `symbols.map` file to remove this?

mattkae (changes requested):
A single question!

src/server/compositor/default_configuration.cpp
mattkae: https://github.com/canonical/mir/actions/runs/8920919391/job/24499927929?pr=3368 :)

RAOF: :tada: 

Hurray for extra CI checks!

mattkae:
Eh, it still seems to want `the_buffer_stream_factory`? I can check that out today

RAOF:
Huh. Thanks, symbol checks, for picking that up.

AlanGriffiths (commented):
I'm not against this, but with `the_buffer_stream_factory()` we had a customisation point.

I have no recollection of why that existed (apparently it was part of a refactoring I did 11 years ago), but are we sure hypothetical downstreams (or ourselves) we don't need it?

RAOF:
> I'm not against this, but with `the_buffer_stream_factory()` we had a customisation point.
> 
> I have no recollection of why that existed (apparently it was part of a refactoring I did 11 years ago), but are we sure hypothetical downstreams (or ourselves) we don't need it?

I *think* it was a configuration point for testing purposes? Our `BufferStream`s *used* to have actual logic in them. I'm not aware of any downstream that currently overrides `the_buffer_stream_factory()` (QtMir definitely doesn't), nor can I think offhand of any time a downstream would *want* a way to wrap all the `BufferStream`s?

AlanGriffiths:
Poking, because we've waited too long for statuses to be reported

RAOF:
Hm. What needs to be done here? #3418 includes this.

AlanGriffiths:
> Hm. What needs to be done here? #3418 includes this

Well, yesterday it looked like you'd addressed @mattkae's concerns, but there were a bunch of checks "pending status". I've fixed that. If @mattkae is happy now, then this can land.

mattkae (approved):
I am good with this now that the symbols are all happy!

PR #3369, opened 2024-05-03
Revert the frig

AlanGriffiths:
Test allowing remote-builds


AlanGriffiths:
@Saviq I think this is good to go: the test tools and frame failures look like they need fixing in those projects?

Saviq (approved):
Yups.

Issue #3357, opened 2024-04-30
Provide a mechanism in miral to move the cursor to an (X, Y) coordinate


mattkae:
## Behavior
- There should be a `WindowManagerTools::move_cursor(int x, int y)`
- This method will move the cursor to the provided position and also clamp it within the bounds of the outputs

PR #3370, opened 2024-05-03
(#3357) provide a mechanism to move the cursor to a position immediately

mattkae:
## What's new?
Compositor implementers can now control the cursor like so:

```c++
bool FloatingWindowManagerPolicy::handle_keyboard_event(MirKeyboardEvent const* event)
{
    // ...

    auto const action = mir_keyboard_event_action(event);
    auto const scan_code = mir_keyboard_event_scan_code(event);
    auto const modifiers = mir_keyboard_event_modifiers(event) & modifier_mask;

    if (action == mir_keyboard_action_down && modifiers == mir_input_event_modifier_alt && scan_code == KEY_0)
    {
        tools.move_cursor_to(mir::geometry::PointF(0, 0)); // Move the cursor to top left corner
    }

    // ...
}
```

All bounds checking is handled by the internal input code, so no need for bounds checking at the interface layer :)


src/miral/window_management_trace.cpp
```diff
@@ -938,3 +938,12 @@ try {
     return policy->advise_application_zone_delete(application_zone);
 }
 MIRAL_TRACE_EXCEPTION
+
+void miral::WindowManagementTrace::move_cursor_to(mir::geometry::PointF point)
+try {
+    std::stringstream out;
+    out << point;
+    mir::log_info("%s cursor=%s", __func__, out.str().c_str());
```
AlanGriffiths, line 946:
```suggestion
    mir::log_info("%s point=%s", __func__, out.str().c_str());
```

include/miral/miral/window_manager_tools.h
```diff
@@ -256,6 +256,13 @@ class WindowManagerTools
      */
     void invoke_under_lock(std::function<void()> const& callback);
 
+    /**
+     * Move the cursor to the provided point. If the point is outside of the range of the outputs,
+     * the point is clamped.
+     * @param point
+     */
+    void move_cursor_to(mir::geometry::PointF point);
```
AlanGriffiths, line 264: What is the rationale for using `PointF` over `Point`?

src/miral/symbols.map
```diff
@@ -264,6 +264,7 @@ global:
     miral::WindowManagerTools::info_for_window_id*;
     miral::WindowManagerTools::invoke_under_lock*;
     miral::WindowManagerTools::modify_window*;
+    miral::WindowManagerTools::move_cursor_to*;
```
AlanGriffiths, line 267: Wrong stanza

src/miral/basic_window_manager.cpp
```diff
@@ -2981,3 +2993,28 @@ void miral::BasicWindowManager::update_application_zones_and_attached_windows()
         area->zone_policy_knows_about = area->application_zone;
     }
 }
+
+void miral::BasicWindowManager::move_cursor_to(mir::geometry::PointF point)
+{
+    action_queue->enqueue(this, [
```
AlanGriffiths, line 2999: I don't think the `ServerActionQueue` is the right executor - this could end up executing on the current thread: `linearising_executor` would be better.

AlanGriffiths (changes requested):
There are conflicts to resolve

src/server/server.cpp
```diff
@@ -115,7 +115,7 @@ struct TemporaryCompositeEventFilter : public mi::CompositeEventFilter
     MACRO(the_rendering_platforms)  \
     MACRO(the_input_targeter)  \
     MACRO(the_logger)\
-    MACRO(the_main_loop)\
+    MACRO(the_main_loop)       \
```
AlanGriffiths, line 118:
```suggestion
    MACRO(the_main_loop)\
```

AlanGriffiths (approved):
Okay

PR #3374, opened 2024-05-06
all: move to 24.04 as the development focus

Saviq:
[`release/2.16`](https://github.com/canonical/mir/tree/release/2.16) now in maintenance mode.


AlanGriffiths (approved):
\o/

.github/workflows/snap.yml
```diff
@@ -35,7 +35,6 @@ jobs:
       fail-fast: false
       matrix:
         architecture: ${{ fromJSON(needs.arches.outputs.arches) }}
-        base: ["22", "24"]
```
AlanGriffiths, line 38: If you drop this, you need to stop using it:
```
        publish-channel: ${{ matrix.track }}/edge/mir-pr${{ github.event.number }}
```

AlanGriffiths (changes requested):

.github/workflows/snap.yml
Saviq: Yeah yeah. I know. Just looking at fixing the spread runs now, since [`lsb-invalid-mta`](https://packages.ubuntu.com/jammy/lsb-invalid-mta) is gone in noble.

debian/control
```diff
@@ -53,7 +53,7 @@ Build-Depends: cmake,
                libyaml-cpp-dev,
                libwayland-dev,
                libnvidia-egl-wayland-dev,
-               eglexternalplatform-dev,
+               eglexternalplatform-dev:all,
```
Saviq, line 56: @RAOF unsure why this would be needed all of a sudden? ¯\\\_(ツ)_/¯

Saviq: Also that doesn't help sbuild :/

Maybe we should actually fix that package for real? :D

AlanGriffiths (approved):
Go for it!

PR #3377, opened 2024-05-07
platforms/eglstream-kms: Don't invoke callback under `configuration_mutex`

RAOF:
The `conf_change_handler()` callback comes from external code, and could do anything. In practice, it *now* ends up (eventually) calling `Display::configuration`, which *also* tries to take the configuration mutex, with hilarious consequences!

Drop the lock before invoking the callback.

Fixes: #3375


AlanGriffiths (approved):
I guess it **is** reasonable for changes to be applied whilst the `conf_change_handler` is running. (And, if not, preventing calls from one display is insufficient.)

PR #3380, opened 2024-05-07
ForeignToplevelManagerV1 now reports hidden window handles instead of purposefully hiding them

mattkae:
When we switch workspaces, foreign top level was reporting different windows, which doesn't seem right. The spec doesn't specify whether or not hidden windows should be reported.


RAOF:
Urgh. This is another “I'm not against it as a practical improvement, but it's the wrong thing”. I think this wants a discussion about our window model next week.

mattkae:
> Urgh. This is another “I'm not against it as a practical improvement, but it's the wrong thing”. I think this wants a discussion about our window model next week.

I still think that this is a worthwhile bugfix until we actually refactor our entire window type system. It usefully fixes a strange bug

RAOF (approved):
Yeah, fair enough. This isn't going to impede doing a better job later!

Issue #3379, opened 2024-05-07
Alt+` is confusing Firefox


AlanGriffiths:
When I updated my Miriway snap to Mir trunk I discovered weirdness...

1. Open a couple of Firefox windows.
2. Alt-` between them
3. Move over a scrollable area and drag two fingers up/down

Expect: area scolls
Actual: Firefox revisits history

AlanGriffiths:
Probably same as #3307, but happens consistently now

mattkae:
Aha! you know what... I've seen this before in miracle-wm, but I couldn't quite put my finger on it: https://github.com/mattkae/miracle-wm/issues/42

AlanGriffiths:
Fixed by #3382

Issue #3307, opened 2024-03-31
Alt key sometimes getting stuck after Alt+Tab


Saviq:
We fixed this before:
- #2946

But it's biting me again from time to time. Of course no reproducer, but we may need to add some debug logging or something so we can track this down.

Saviq:
Data point: it gets cleared after the first event.

mattkae:
## Notes
Last time this was a problem, we were failing on `WlKeyboard::send_modifiers`. In this case, we were failing to send the modifiers due to a faulty condition. My suspicion here is that we're somehow sending incorrect modifiers, or not sending them at all.

mattkae:
Let's get some logging first: https://github.com/canonical/mir/pull/3351

AlanGriffiths:
I have a reproducer for what is likely the same bug: I have a script to swap my outputs (because they come up in an indeterminate order):
```
$ cat `which miriway-swap`
#!/bin/sh
exec sed -i "s/position: \[1920, 0\]/@@@/;s/position: \[0, 0\]/position: \[1920, 0\]/;s/@@@/position: \[0, 0\]/" ~/.config/miriway-shell.display
```
And have that bound:
```
$ grep swap ~/.config/miriway-shell.config
ctrl-alt=s:miriway-swap
```
I have found that after using `Ctrl-Alt-S`, I need to press and release `Alt` before it works again.

(Sorry, not running a build with your debug info yet, maybe later)

mattkae:
@AlanGriffiths Is that reproducer consistent for you, or does it still happen only sometimes? I am unable to reproduce it on my end when I follow your steps

AlanGriffiths:
> Is that reproducer consistent for you, or does it still happen only sometimes?

When I posted, it seemed consistent. Now, when I retry, not only it is intermittent, but pressing and releasing `Alt` doesn't fix it reliably either. Not sure what is different, but will monitor in the hope of enlightenment.

mattkae:
> > Is that reproducer consistent for you, or does it still happen only sometimes?
> 
> When I posted, it seemed consistent. Now, when I retry, not only it is intermittent, but pressing and releasing `Alt` doesn't fix it reliably either. Not sure what is different, but will monitor in the hope of enlightenment.

Ok let me know! Remember that you can use this pull request to get some better info: https://github.com/canonical/mir/pull/3351

AlanGriffiths:
Fixed by #3382

PR #3382, opened 2024-05-07
 (#3379 #3307) Fix accidentally consuming input events on alt key up 

mattkae:
## What's new?
- We were erroneously eating the event on complete, which made the final "up" on an alt some times not complete :facepalm:

### DIFF
--- a/src/miral/minimal_window_manager.cpp
+++ b/src/miral/minimal_window_manager.cpp
@@ -197,10 +197,8 @@ bool miral::MinimalWindowManager::handle_keyboard_event(MirKeyboardEvent const* 
         switch (mir_keyboard_event_scan_code(event))
         {
             case KEY_LEFTALT:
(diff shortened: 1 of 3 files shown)


src/miral/minimal_window_manager.cpp
```diff
@@ -197,10 +197,8 @@ bool miral::MinimalWindowManager::handle_keyboard_event(MirKeyboardEvent const*
         switch (mir_keyboard_event_scan_code(event))
         {
             case KEY_LEFTALT:
-                if (self->application_selector.complete() != nullptr)
-                    return true;
-                break;
-            default:;
+                self->application_selector.complete();
+                default:;
```
AlanGriffiths, line 201:
```suggestion
                self->application_selector.complete();
                [[fallthrough]];
            default:;
```

RAOF (approved):
Aha!

PR #3385, opened 2024-05-09
(#3309) selecting a correct next window after the focused one is closed or disabled

### LINKED ISSUE #3309: Closing a window brings one of the remaining windows of the same client to the front
Steps:
- have two apps of App A open
- have App B open
- Alt+Tab to App B, then App A
- close foreground App A window

Expected:
- focus goes back to App B

Current:
- focus goes to the other window of App A

mattkae:
fixed #3309

## What's new?
- When a focused window closes, we first try to select the next window given the previous scheme. If that doesn't work, we choose the other scheme. If that doesn't work, we select no window.
- Moved the `BasicWindowManager` check for `is_active_window` below `policy->advise_delete_window`. In this way, we only call `BasicWindowManager::refocus` if the active window remains unchaged after the `advise`


RAOF (commented):
If I understand this correctly, what this change will do is:
when the currently focused window is closed, do the equivalent of Alt-Tab or Alt-` depending on which one the user used last?

That doesn't seem like reasonable behaviour to me, so maybe I'm misunderstanding the change?

mattkae:
> If I understand this correctly, what this change will do is: when the currently focused window is closed, do the equivalent of Alt-Tab or Alt-` depending on which one the user used last?
> 
> That doesn't seem like reasonable behaviour to me, so maybe I'm misunderstanding the change?

Your understanding is correct. The reason for this is that you want to select the next app/window _in the focus order_. Hence, the next available one could be found by Alt + Tab/Tick depending on their last action.

AlanGriffiths:
> Your understanding is correct. The reason for this is that you want to select the next app/window _in the focus order_. Hence, the next available one could be found by Alt + Tab/Tick depending on their last action.

I too don't find this reasonable.

When I close a window I am not thinking about whether I last used `Alt-Tab` or `Alt-Tick` and don't expect the result to depend on it. I expect the next window in _the_ focus order, not "the next window of another app" nor "the next window of the same app".

AlanGriffiths:
I do agree though, that the current behaviour needs fixing

mattkae:
I think I should revisit this later. I have a plan to do a full refactor of the focus system here, and this bug would be easily fixed by such a refactor

PR #3387, opened 2024-05-10
miniscule fix for building on systems where the hardcoded path to libdrm headers doesn't exist

pillowtrucker:
, e.g. NixOS.
I hope this still works on other systems.

### DIFF
--- a/src/platform/graphics/CMakeLists.txt
+++ b/src/platform/graphics/CMakeLists.txt
@@ -49,13 +49,15 @@ if (DRM_VERSION VERSION_GREATER 2.4.107)
       PkgConfig::DRM)
 endif()
 
+find_path(DRM_FOURCC_INCLUDE_DIR NAMES "drm_fourcc.h" PATH_SUFFIXES "libdrm" "" HINTS ${DRM_INCLUDE_DIRS})
+
 add_custom_command(
   OUTPUT
     drm-formats
   VERBATIM
   COMMAND
     "sh" "-c"
-      "grep \"#define DRM_FORMAT\" /usr/include/libdrm/drm_fourcc.h | \
+      "grep \"#define DRM_FORMAT\" ${DRM_FOURCC_INCLUDE_DIR}/drm_fourcc.h | \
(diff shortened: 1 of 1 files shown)


RAOF (dismissed):
Neat!

This information should be available in pkgconfig, right? I think it would be a better idea to pull it from there (as that's where the compiler will pull it from) - I'm not sure if CMake's `find_path` will match the pkgconfig behaviour, and things could get weird if those don't match.

pillowtrucker:
Oh, I totally missed that there is a similar variable already set by pkg_check_modules.
The problem with that though is that it can (and does) return more than one directory, so it would still have to be something like
```cmake
find_path(DRM_FOURCC_INCLUDE_DIR NAMES "drm_fourcc.h" PATH_SUFFIXES "libdrm" "" HINTS ${DRM_INCLUDE_DIRS})
```
which is kind of redundant, but I guess it doesn't hurt either. `find_path` always returns the first match, so it's guaranteed to be at most one entry. Should I make that change ?

pillowtrucker:
Fixed up and rebuilt and briefly tested. There was a third occurrence of the hardcoded path there that I had missed which apparently didn't cause any build or runtime failure on my computer, but left the DRM_MODIFIERS_FILE incomplete, that's also been taken into account now.

RAOF:
Thanks for the ping! This fell off my plate when travelling to the recent sprint, the week-long sprint, and then the travel back :wink: 

I've tweaked the `find_path` invocation slightly, to make it fail if it doesn't find `drm_fourcc.h` and to make sure it *only* looks in places pkgconf lists. Please check that this still works for you, and then I think this is good!

pillowtrucker:
Yep that still builds and runs on nix. Good call, because otherwise the find_* functions first look in all of the paths in one of the CMAKE_* environment variables set by nix.

AlanGriffiths (approved):
Ignoring paranoid thoughts like "what if `DRM_FOURCC_INCLUDE_DIR` includes spaces?" this looks reasonable.

Issue #3383, opened 2024-05-08
Strange libc++ display_configuration.h std::map FTBFS


JamiKettunen:
While packaging Mir for https://chimera-linux.org (musl/clang/libc++/compiler-rt/libunwind environment) I got this error:
```
FAILED: src/platform/graphics/CMakeFiles/mirplatformgraphicscommon.dir/overlapping_output_grouping.cpp.o
[...]

PR #3388, opened 2024-05-12
Fix building on LLVM libc++

JamiKettunen:
This should complete the first part of my journey of running Mir on [Chimera Linux](https://chimera-linux.org) initially reported on #3383.

Now I'm figuring out this runtime failure of `mir_demo_server` (https://paste.c-net.org/umtbetmnnvsn) which I cannot reproduce on Arch with (https://paste.c-net.org/a3azgse7x8en) or without these patches, further opinions welcome :)

### DIFF
--- a/src/miral/static_display_config.h
+++ b/src/miral/static_display_config.h
@@ -69,7 +69,7 @@ private:
         mir::optional_value<float>  scale;
         mir::optional_value<MirOrientation>  orientation;
         mir::optional_value<int> group_id;
-        std::map<std::string const, std::optional<std::string>> custom_attribute;
+        std::map<std::string, std::optional<std::string>> custom_attribute;
     };
 
     using Port2Config = std::map<std::string, Config>;
--- a/src/platforms/virtual/display.h
+++ b/src/platforms/virtual/display.h
@@ -19,6 +19,7 @@
 
 #include "platform.h"
 #include <mir/graphics/display.h>
+#include <mutex>
 
 namespace mir
 {
--- a/src/server/compositor/multi_monitor_arbiter.cpp
+++ b/src/server/compositor/multi_monitor_arbiter.cpp
@@ -18,6 +18,7 @@
 #include "mir/graphics/buffer.h"
 #include "mir/frontend/event_sink.h"
 #include <boost/throw_exception.hpp>
+#include <algorithm>
 
 namespace mg = mir::graphics;
 namespace mc = mir::compositor;
--- a/src/server/console/linux_virtual_terminal.cpp
+++ b/src/server/console/linux_virtual_terminal.cpp
@@ -655,7 +655,7 @@ std::future<std::unique_ptr<mir::Device>> mir::LinuxVirtualTerminal::acquire_dev
         {
             using namespace boost::iostreams;
             char line_buffer[1024];
-            stream<file_descriptor_source> uevent{fd, file_descriptor_flags::never_close_handle};
+            stream<file_descriptor_source> uevent{static_cast<int>(fd), file_descriptor_flags::never_close_handle};
 
             while (uevent.getline(line_buffer, sizeof(line_buffer)))
             {
--- a/src/server/console/minimal_console_services.cpp
+++ b/src/server/console/minimal_console_services.cpp
@@ -100,7 +100,7 @@ std::future<std::unique_ptr<mir::Device>> mir::MinimalConsoleServices::acquire_d
     auto const devnode = [](auto const& fd) {
         using namespace boost::iostreams;
         char line_buffer[1024];
-        stream<file_descriptor_source> uevent{fd, file_descriptor_flags::never_close_handle};
+        stream<file_descriptor_source> uevent{static_cast<int>(fd), file_descriptor_flags::never_close_handle};
 
         while (uevent.getline(line_buffer, sizeof(line_buffer)))
         {
--- a/include/common/mir/log.h
+++ b/include/common/mir/log.h
@@ -22,6 +22,7 @@
 #include "mir/logging/logger.h"  // for Severity
 #include <string>
 #include <cstdarg>
+#include <exception>
 
 namespace mir
 {
--- a/include/miral/miral/application.h
+++ b/include/miral/miral/application.h
@@ -21,6 +21,7 @@
 
 #include <memory>
 #include <string>
+#include <unistd.h>
 
 namespace mir
 {
--- a/include/platform/mir/graphics/display_configuration.h
+++ b/include/platform/mir/graphics/display_configuration.h
@@ -148,7 +148,7 @@ struct DisplayConfigurationOutput
     std::string name = "OUT-" + std::to_string(id.as_value());
 
     /// Custom attributes (typically set via the .display configuration file
-    std::map<std::string const, std::optional<std::string>> custom_attribute = {};
+    std::map<std::string, std::optional<std::string>> custom_attribute = {};
 
     /** The logical rectangle occupied by the output, based on its position,
         current mode and orientation (rotation) */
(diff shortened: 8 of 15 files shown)


src/server/console/minimal_console_services.cpp
```diff
@@ -100,7 +100,7 @@ std::future<std::unique_ptr<mir::Device>> mir::MinimalConsoleServices::acquire_d
     auto const devnode = [](auto const& fd) {
         using namespace boost::iostreams;
         char line_buffer[1024];
-        stream<file_descriptor_source> uevent{fd, file_descriptor_flags::never_close_handle};
+        stream<file_descriptor_source> uevent{(int)fd, file_descriptor_flags::never_close_handle};
```
AlanGriffiths, line 103: `Fd` has an implicit conversion to `int`, so I'm surprised this is needed. But it should not be a c-style cast.

include/miral/miral/output.h
```diff
@@ -108,7 +108,7 @@ class Output
 
     /// A custom attribute map
     /// \remark Since MirAL 3.8
-    auto attributes_map() const -> std::map<std::string const, std::optional<std::string>>;
+    auto attributes_map() const -> std::map<std::string, std::optional<std::string>>;
```
AlanGriffiths, line 111: Removing the old signature is an ABI break. We should retain that.

src/miral/static_display_config.h
```diff
@@ -69,7 +69,7 @@ class YamlFileDisplayConfig : public mir::graphics::DisplayConfigurationPolicy
         mir::optional_value<float>  scale;
         mir::optional_value<MirOrientation>  orientation;
         mir::optional_value<int> group_id;
-        std::map<std::string const, std::optional<std::string>> custom_attribute;
+        std::map<std::string, std::optional<std::string>> custom_attribute;
```
AlanGriffiths, line 72: Changing this type is fine

src/server/console/linux_virtual_terminal.cpp
```diff
@@ -655,7 +655,7 @@ std::future<std::unique_ptr<mir::Device>> mir::LinuxVirtualTerminal::acquire_dev
         {
             using namespace boost::iostreams;
             char line_buffer[1024];
-            stream<file_descriptor_source> uevent{fd, file_descriptor_flags::never_close_handle};
+            stream<file_descriptor_source> uevent{(int)fd, file_descriptor_flags::never_close_handle};
```
AlanGriffiths, line 658: `Fd` has an implicit conversion to `int`, so I'm surprised this is needed. But it should not be a c-style cast.

src/platforms/virtual/display.h
```diff
@@ -17,6 +17,7 @@
 #ifndef MIR_GRAPHICS_VIRT_DISPLAY_H_
 #define MIR_GRAPHICS_VIRT_DISPLAY_H_
 
+#include <mutex>
```
AlanGriffiths, line 20: We keep standard headers after project ones

src/server/compositor/multi_monitor_arbiter.cpp
```diff
@@ -17,6 +17,7 @@
 #include "multi_monitor_arbiter.h"
 #include "mir/graphics/buffer.h"
 #include "mir/frontend/event_sink.h"
+#include <algorithm>
```
AlanGriffiths, line 20: We keep standard headers after project ones

AlanGriffiths (changes requested):
Thanks for this, there's a little nitpicking but it is basically fine

include/miral/miral/output.h
JamiKettunen: How do you propose that is done? If I simply add the new one right below it FTBFS with:
```
In file included from ../src/miral/active_outputs.cpp:18:
In file included from ../src/miral/display_configuration_listeners.h:20:
../include/miral/miral/output.h:112:10: error: functions that differ only in their return type cannot be overloaded
  112 |     auto attributes_map() const -> std::map<std::string, std::optional<std::string>>;
      |          ^
../include/miral/miral/output.h:111:10: note: previous declaration is here
  111 |     auto attributes_map() const -> std::map<std::string const, std::optional<std::string>>;
      |          ^
```

AlanGriffiths: What problem did you see that motivated you to change the return type?

JamiKettunen: #3383 basically, if I don't do each hunk of the changes in 517b4b13a1bc033fd4fd03f453157d8f88ef7389 I see a similar error as I originally reported there

JamiKettunen: To be specific, without the `attributes_map` (`include/miral/miral/output.h` and `src/miral/output.cpp`) changes I get:
```
../src/miral/output.cpp:127:12: error: no viable conversion from returned value of type 'map<basic_string<...>, [...]>' to function return type 'map<const basic_string<...>, [...]>'
  127 |     return self->custom_attribute;
      |            ^~~~~~~~~~~~~~~~~~~~~~
/usr/bin/../include/c++/v1/map:1078:25: note: candidate constructor not viable: no known conversion from 'std::map<std::string, std::optional<std::string>>' (aka 'map<basic_string<char>, optional<basic_string<char>>>') to 'const map<const string, optional<string>> &' for 1st argument
 1078 |   _LIBCPP_HIDE_FROM_ABI map(const map& __m) : __tree_(__m.__tree_) { insert(__m.begin(), __m.end()); }
      |                         ^   ~~~~~~~~~~~~~~
/usr/bin/../include/c++/v1/map:1096:25: note: candidate constructor not viable: no known conversion from 'std::map<std::string, std::optional<std::string>>' (aka 'map<basic_string<char>, optional<basic_string<char>>>') to 'map<const string, optional<string>> &&' for 1st argument
 1096 |   _LIBCPP_HIDE_FROM_ABI map(map&& __m) _NOEXCEPT_(is_nothrow_move_constructible<__base>::value)
      |                         ^   ~~~~~~~~~
/usr/bin/../include/c++/v1/map:1106:25: note: candidate constructor not viable: no known conversion from 'std::map<std::string, std::optional<std::string>>' (aka 'map<basic_string<char>, optional<basic_string<char>>>') to 'initializer_list<value_type>' (aka 'initializer_list<pair<const std::string, std::optional<std::string>>>') for 1st argument
 1106 |   _LIBCPP_HIDE_FROM_ABI map(initializer_list<value_type> __il, const key_compare& __comp = key_compare())
      |                         ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/bin/../include/c++/v1/map:1034:34: note: explicit constructor is not a candidate
 1034 |   _LIBCPP_HIDE_FROM_ABI explicit map(const key_compare& __comp) _NOEXCEPT_(
      |                                  ^
/usr/bin/../include/c++/v1/map:1128:34: note: explicit constructor is not a candidate
 1128 |   _LIBCPP_HIDE_FROM_ABI explicit map(const allocator_type& __a) : __tree_(typename __base::allocator_type(__a)) {}
      |                                  ^

```

AlanGriffiths: It should be enough to change:
```c++
  return self->custom_attribute;
```
to 
```c++
   return {self->custom_attribute.begin(), self->custom_attribute.end()};
```

src/server/console/minimal_console_services.cpp
JamiKettunen: Switched to `static_cast<int>(fd)`

include/miral/miral/output.h
JamiKettunen: Indeed that appears to do the trick

AlanGriffiths (approved):
Looks good now, thanks!

PR #3394, opened 2024-05-23
platforms/gbm-kms: Debug printing for GBM surface acquire/release

RAOF:
Just to check on some systems that we're doing the right thing / gain extra insight into what's going wrong.


RAOF:
This debugging has run its course.

PR #3401, opened 2024-05-27
linux-dmabuf: Fix cross-GPU imports

RAOF:
One-half of fixing hybrid on my laptop is to allocate
the bounce buffers necessary to get the client buffer
somewhere the compositing GPU can read it in a *format*
that the compositing GPU can read.

The code *was* trying to do that by matching on supported
modifiers. Unfortunately, due to other constraints, the
code accepts (and must accept, elsewhere) `DRM_FORMAT_MOD_INVALID`.

`MOD_INVALID` means “use a driver-specific side-band to know
what the format is”, and is unlikely to work across GPUs.


AlanGriffiths (approved):
Sure

PR #3402, opened 2024-05-28
graphics/PlatformProbe: Implement a more principled `GraphicBufferAllocator` selector

RAOF:
This implements an actual selection process for the GraphicBufferAllocator:
* First, find the `RenderingPlatform`(s) with *fewest* displays they *can't* drive, then
* Select the `RenderingPlatform` with the highest number of displays probed 'best' for

In the case of ties, the algorithm picks the first `RenderingPlatform` in the list;
due to the selections elsewhere, that will be the first platform in alphabetical order

Fixes: #3400

### DIFF
--- a/tests/include/mir/test/doubles/mock_gl_rendering_provider.h
+++ b/tests/include/mir/test/doubles/mock_gl_rendering_provider.h
@@ -21,6 +21,7 @@
 
 #include "mir/graphics/gl_config.h"
 #include "mir/graphics/platform.h"
+#include "mir/renderer/gl/gl_surface.h"
 
 namespace mir::test::doubles
 {
--- a/src/include/server/mir/default_server_configuration.h
+++ b/src/include/server/mir/default_server_configuration.h
@@ -22,7 +22,6 @@
 
 #include <memory>
 #include <string>
-#include <mutex>
 
 namespace mir
 {
--- a/src/server/graphics/default_configuration.cpp
+++ b/src/server/graphics/default_configuration.cpp
@@ -39,7 +39,6 @@
 #include "mir/main_loop.h"
 #include "mir/udev/wrapper.h"
 
-#include <algorithm>
 #include <boost/throw_exception.hpp>
 
 #include <sstream>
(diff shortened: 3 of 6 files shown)


RAOF:
This needs further testing on hardware I have (laptop, RPi4) before merging.

AlanGriffiths (approved):
This looks sensible, and obviously does what is described.

But could there be tests?

RAOF:
Yes, there could (and should!) be tests!

RAOF:
Ok! glmark on the Pi is noisy, but this does not appear to have a significant performance effect there (and it still renders correctly :wink:).

It *works* on my laptop - rendering is correct, both from i915 and amdgpu, and performance is better when rendering on i915. Performance seems to be a lot worse on amdgpu. Here we run into “how useful a benchmark is glmark?”. I'm not sure, and I'm still investigating exactly where the difference is.

RAOF:
OK. So, the difference between trunk and this:
* On trunk, amdgpu uses `linux_dmabuf`. The client driver selects a format/modifier pair that's (apparently) much faster for amdgpu to render to, but is incompatible with i915.
* On this (and, it turns out, 2.16.2) the client driver uses `wl_drm` and `create_prime_buffer`; this will (presumably, there's no easy way to introspect this) allocate a linear buffer that is slower to render to.

So this is *not* a performance regression vs 2.16.2 on AMD on my Intel/AMD hybrid laptop, and renders correctly.

It might be interesting in future to see how rendering to the optimal buffer layout, and then going through the bounce-buffer to be sampleable from the Intel GPU, but that's definitely future work.

This should be good to go.

tests/include/mir/test/doubles/mock_gl_rendering_provider.h
```diff
@@ -21,6 +21,7 @@
 
 #include "mir/graphics/gl_config.h"
 #include "mir/graphics/platform.h"
+#include "mir/renderer/gl/gl_surface.h"
```
AlanGriffiths, line 24: Spurious addition?

AlanGriffiths (approved):
Yeah, that looks better

tests/include/mir/test/doubles/mock_gl_rendering_provider.h
AlanGriffiths: Apparently not

RAOF: Yeah, need the definition so the destructor can be instantiated!

PR #3403, opened 2024-05-29
drive-by simplification of the smoke test script

AlanGriffiths:


mattkae (approved):
Working well for me even when `eglinfo` returns with -1

PR #3405, opened 2024-05-30
(#3301) feature: providing a listener for dim/screen off notifications and wiring it through miral

### LINKED ISSUE #3301: MirAL: provide a notification of screen dim & off
Shells built on Mir need a way to react when an idle-timeout happens.

C.f. https://github.com/Miriway/Miriway/issues/95

mattkae:
fixes #3301 

## What's new?
- Created `mir::shell::IdleHandlerObserver`
- `mir::shell::IdleHandler` now implements `ObserverRegistrar`
- `mir::shell::BasicIdleHandler` uses `ObserverMultiplexer` to notify interested parties about screen dims and offs
- `miral::IdleListener` exposes a way for shell authors to register notifications with the `IdleHandler`
- Regenerated symbols for `mirserver` and `miral`

## To Test
```c++
// shell_main.cpp
#include <miral/idle_listener.h>

...

runner.run_with({
   ....
   IdleListener{}
                .on_dim([]{ printf("Dim!\n"); })
                .on_off([]{ printf("Off!\n"); })
                .on_wake([]{ printf("Wake!\n"); })
});
```

Then run `miral-app --idle-timeout=20`. "Dimmed!" will print after 10 seconds and "Off!" will print after 20 seconds.


include/miral/miral/idle_listener.h
```diff
@@ -0,0 +1,43 @@
...
+namespace mir { class Server; }
+
+namespace miral
+{
+class IdleListener
+{
+public:
+    using Callback = std::function<void()>;
+
+    IdleListener(Callback const& on_dm, Callback const& on_off);
```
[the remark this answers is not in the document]
AlanGriffiths: And, if so, there comes a point when multiple parameters with identical type are better represented as a struct or (in this case, maybe) an interface.

AlanGriffiths (changes requested):
Just some API tweaking

include/miral/miral/idle_listener.h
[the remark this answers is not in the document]
mattkae: I was wondering the same thing. We could add `on_wake` fairly easily. Do you think it's worth adding now?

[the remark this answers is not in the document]
mattkae: Let's add `wake` too!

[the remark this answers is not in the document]
mattkae: Fixed :+1:

AlanGriffiths (changes requested):
Symbols need fixing

include/miral/miral/idle_listener.h
```diff
@@ -0,0 +1,51 @@
...
+namespace miral
+{
+
+struct IdleListenerParams
+{
+    using Callback = std::function<void()>;
+    Callback on_dim;
+    Callback on_off;
+    Callback on_wake;
+};
```
AlanGriffiths, line 34: Given the added complexity of checking these members are set, maybe it would be better to have an interface:
```suggestion
class IdleObserver
{
public:
    IdleObserver() = default;
    virtual ~IdleObserver() = default;
    virtual void on_dim();
    virtual void on_off();
    virtual void on_wake();

private:
    IdleObserver(IdleObserver const&) = delete;
    IdleObserver& operator=(IdleObserver const&) = delete;
};
```
Or at least provide default initialisation (making the contract "don't reset this"):
```suggestion
struct IdleListenerParams
{
    using Callback = std::function<void()>;
    Callback on_dim = []{};
    Callback on_off = []{};
    Callback on_wake = []{};
};
```

src/miral/idle_listener.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <mir/server.h>
+#include <mir/shell/idle_handler.h>
+
+class miral::IdleListener::Impl : public mir::shell::IdleHandlerObserver
+{
+public:
+    explicit Impl(IdleListenerParams const& params)
+        : on_dim{params.on_dim},
+          on_off{params.on_off},
+          on_wake{params.on_wake}
```
AlanGriffiths, line 27: I think I'd move the tests for empty `function<>`s here:
```suggestion
        : on_dim{params ? params.on_dim : []{}},
          on_off{params ? params.params.on_off : []{}},
          on_wake{params ? params.params.on_wake : []{}}
```

(Why is there no `std::function<>::else(...)`?)

AlanGriffiths (changes requested):
Some nitpicking.

I mention several options, I think my favourite is "provide initialisers and don't check for empty".

src/miral/idle_listener.cpp
mattkae: `IdleListenerParams` now has default empty callbacks

include/miral/miral/idle_listener.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <functional>
+#include <memory>
+
+namespace mir { class Server; }
+
+namespace miral
+{
+
+struct IdleListenerParams
```
AlanGriffiths, line 28: I don't think "Params" is good naming. And if it were, then `IdleListener::Params` would probably be better scoping. Maybe `IdleListener::Callbacks`?

Saviq: It does sound like there were more `Params` before, now they're just callbacks?

Also `IdleListener::Callbacks::Callback` would probably need moving up?

mattkae: Updated to:
- `IdleListener::Callbacks`
- `IdleListener::Callback`

include/miral/miral/idle_listener.h
```diff
@@ -0,0 +1,51 @@
...
+public:
+    using Callback = std::function<void()>;
+    struct Callbacks
+    {
+        Callback on_dim = []{};
+        Callback on_off = []{};
+        Callback on_wake = []{};
+    };
+
+    explicit IdleListener(Callbacks const&);
```
AlanGriffiths, line 40: I have been hasty or imprecise in my comments a few iterations back. I see no improvement from:

```c++
IdleListener foo{[&]{stuff}, [&]{more-stuff}, [&]{even-more-stuff}};
```
to what we have now:

```c++
IdleListener foo{{[&]{stuff}, [&]{more-stuff}, [&]{even-more-stuff}}};
```
I.e. multiple parameters of the same type in some order.

I think this would be better:

```c++
auto foo = IdleListener{}
    .on_dim([&]{stuff})
    .on_off([&]{more-stuff}
    .on_wake([&]{even-more-stuff});
```

include/miral/miral/idle_listener.h
```diff
@@ -0,0 +1,54 @@
...
+class IdleListener
+{
+public:
+    using Callback = std::function<void()>;
+    struct Callbacks
+    {
+        Callback on_dim = []{};
+        Callback on_off = []{};
+        Callback on_wake = []{};
+    };
```
AlanGriffiths, line 38: This is not part of the API

debian/libmiral7.symbols
```diff
@@ -67,6 +67,13 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::ExternalClientLauncher::snapcraft_launch(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const@MIRAL_5.0" 5.0.0
  (c++)"miral::ExternalClientLauncher::split_command(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::ExternalClientLauncher::~ExternalClientLauncher()@MIRAL_5.0" 5.0.0
+ (c++)"miral::IdleListener::IdleListener()@MIRAL_5.0" 5.0.0
+ (c++)"miral::IdleListener::IdleListener(miral::IdleListenerParams const&)@MIRAL_5.0" 5.0.0
+ (c++)"miral::IdleListener::IdleListener(std::function<void ()> const&, std::function<void ()> const&)@MIRAL_5.0" 5.0.0
+ (c++)"miral::IdleListener::on_dim(std::function<void ()> const&)@MIRAL_5.0" 5.0.0
+ (c++)"miral::IdleListener::on_off(std::function<void ()> const&)@MIRAL_5.0" 5.0.0
+ (c++)"miral::IdleListener::on_wake(std::function<void ()> const&)@MIRAL_5.0" 5.0.0
+ (c++)"miral::IdleListener::operator()(mir::Server&) const@MIRAL_5.0" 5.0.0
```
AlanGriffiths, line 76: Miral 5.0 was Mir 2.17, These need to go in 5.1 (Except the ones that don't exist)

src/server/symbols.map
```diff
@@ -593,6 +594,8 @@ global:
     mir::shell::FocusController::operator*;
     mir::shell::IdleHandler::?IdleHandler*;
     mir::shell::IdleHandler::IdleHandler*;
+    mir::shell::IdleHandlerObserver::?IdleHandlerObserver*;
+    mir::shell::IdleHandlerObserver::IdleHandlerObserver*;
```
AlanGriffiths, line 598: These, and other new symbols, belong in a 2.18 stanza

src/miral/symbols.map
```diff
@@ -56,6 +56,12 @@ global:
     miral::ExternalClientLauncher::snapcraft_launch*;
     miral::ExternalClientLauncher::split_command*;
     miral::FdHandle::?FdHandle*;
+    miral::IdleListener::?IdleListener*;
+    miral::IdleListener::IdleListener*;
+    miral::IdleListener::on_dim*;
+    miral::IdleListener::on_off*;
+    miral::IdleListener::on_wake*;
+    miral::IdleListener::operator*;
```
AlanGriffiths, line 64: These belong in Miral 5.1

AlanGriffiths (changes requested):
Introducing new APIsdoesn't break ABI, but they belong in new stanzas as they were not in the 2.17 release

debian/libmiral7.symbols
```diff
@@ -143,7 +143,6 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::PrintTo(miral::Window const&, std::basic_ostream<char, std::char_traits<char> >*)@MIRAL_5.0" 5.0.0
  (c++)"miral::SessionLockListener::SessionLockListener(std::function<void ()> const&, std::function<void ()> const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::SessionLockListener::operator()(mir::Server&) const@MIRAL_5.0" 5.0.0
- (c++)"miral::SessionLockListener::~SessionLockListener()@MIRAL_5.0" 5.0.0
```
AlanGriffiths, line 146: As discussed on Matrix, it is problematic to remove symbols that are part of a released ABI. (It is unclear what change led to these being reported missing)

mattkae:
Symbols issues will be fixed by https://github.com/canonical/mir/pull/3423

src/server/shell/basic_idle_handler.h
```diff
@@ -42,6 +44,30 @@ namespace shell
 {
 class DisplayConfigurationController;
 
+class BasicIdleHandlerObserverMultiplexer: public ObserverMultiplexer<IdleHandlerObserver>
```
AlanGriffiths, line 47: I started wondering why this was at namespace scope, and then realised that it is used by `Dimmer` and `PowerModeSetter` (which actually only need the base interface `IdleHandlerObserver`). Perhaps not worth changing but...

mattkae: Worth changing IMO :+1:

src/server/shell/basic_idle_handler.cpp
```diff
@@ -100,7 +100,7 @@ struct Dimmer : ms::IdleStateObserver
     Dimmer(
         std::shared_ptr<mi::Scene> const& input_scene,
         std::shared_ptr<mg::GraphicBufferAllocator> const& allocator,
-        msh::BasicIdleHandlerObserverMultiplexer& multiplexer)
+        msh::IdleHandlerObserver& multiplexer)
```
AlanGriffiths, line 103: Probably worth renaming the parameter (and member)

AlanGriffiths (approved):
OK

PR #3406, opened 2024-05-30
Refresh the Alpine builds

AlanGriffiths:
Alpine 3.20 is out and Alpine 3.18 is old (e.g. doesn't have __cpp_lib_format (std::format))


mattkae (approved):
Good by me :saluting_face:

Issue #3408, opened 2024-05-31
Closing a window causes minimized windows to be focused


mattkae:
## Repro
1. Open up a terminal
2. Open up gedit
3. Minimize the terminal
4. Close gedit
5. Watch as the terminal is restored and focused

## Expected
The terminal stays minimized

mattkae:
I think this one isn't fixed from #3409. It is actually somehow unrelated :sweat_smile:

tarek-y-ismail:
Maybe this is related to how we manage surfaces/windows with a stack? Whenever you pop the second window, the first one becomes the "active window". I'm not sure if we handle the case where we pop and don't have an active window or not.

AlanGriffiths:
Fix reverted

PR #3409, opened 2024-05-31
bugfix: alt+tab no longer restores windows if they don't end up focused

### LINKED ISSUE #3323: Alt+Tab restores minimized windows, even if they don't end up focused
Steps:
- have at least three non-obscuring windows open
- minimize one of them
- cycle through Alt+Tab to reach the other un-minimized window

Expected:
- the minimized window remains minimized

Current:
- the minimized window is restored

Also: should minimized windows be moved to the bottom of the stack? That would certainly alleviate this issue.

mattkae:
fixes #3323 

Note that this does **NOT** fix https://github.com/canonical/mir/issues/3408

## What's new?
- We now remember the restore state of the selected window as we're cycling through it

### DIFF
--- a/src/miral/application_selector.cpp
+++ b/src/miral/application_selector.cpp
@@ -159,7 +159,7 @@ auto ApplicationSelector::complete() -> Window
 
     originally_selected_it = focus_list.end();
     is_active_ = false;
-    restore_state = mir_window_state_unknown;
+    restore_state.reset();
     return selected;
 }
 
--- a/src/miral/application_selector.cpp
+++ b/src/miral/application_selector.cpp
@@ -50,6 +50,25 @@ void ApplicationSelector::advise_new_window(WindowInfo const& window_info)
     focus_list.push_back(window_info.window());
 }
 
+void ApplicationSelector::select(miral::Window const& window)
+{
+    if (selected)
+    {
+        // Restore the previously selected window to its previous state when another is selected
+        auto const& info = tools.info_for(selected);
+        if (restore_state && restore_state != info.state())
+        {
+            WindowSpecification spec;
+            spec.state() = restore_state.value();
+            tools.modify_window(selected, spec);
+            restore_state.reset();
+        }
+    }
+
+    // Update the current selection
+    selected = window;
+}
+
 void ApplicationSelector::advise_focus_gained(WindowInfo const& window_info)
 {
     auto window = window_info.window();
@@ -68,8 +87,7 @@ void ApplicationSelector::advise_focus_gained(WindowInfo const& window_info)
             mir::fatal_error("advise_focus_gained: window was not found in the list");
     }
 
-    // Update the current selection
-    selected = window;
+    select(window);
 }
 
 void ApplicationSelector::advise_focus_lost(miral::WindowInfo const& window_info)
@@ -81,7 +99,7 @@ void ApplicationSelector::advise_focus_lost(miral::WindowInfo const& window_info
         return;
     }
 
-    selected = Window();
+    select(Window());
 }
 
 void ApplicationSelector::advise_delete_window(WindowInfo const& window_info)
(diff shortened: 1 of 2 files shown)


src/miral/application_selector.cpp
```diff
@@ -250,7 +270,9 @@ auto ApplicationSelector::advance(bool reverse, bool within_app) -> Window
     else
         tools.swap_tree_order(next_window.value(), selected);
 
+    auto next_state_to_preserve =  tools.info_for(next_window.value()).state();
     tools.select_active_window(next_window.value());
+    restore_state = next_state_to_preserve;
```
RAOF, line 275: It's not clear to me why `next_state_to_preserve` exists? Why don't we immediately save it to `restore_state`?

This makes the code smell like there's spooky non-local state things happening here.

src/miral/application_selector.h
```diff
@@ -90,6 +91,7 @@ class ApplicationSelector
 
     /// The application that is currently selected.
     Window selected;
+    MirWindowState restore_state = mir_window_state_unknown;
```
RAOF, line 94: `restore_state` seems to be a good candidate for `std::optional<MirWindowState>` rather than an in-band sentinal?

RAOF (changes requested):
I wonder if the state machine here is getting sufficiently complex that it should be deliberately modelled?

A more invasive, but more principled approach, would be to save the entire window manager state when initiating the switcher, let the switcher fiddle around as it pleased, and then on completion take the saved state, apply *only* the finally-selected change, and switch to that.

I'm not blocking on doing that, just floating that it might need to be done eventually.

(The only *blocking* thing here is removing the `printf` debugging; I think turning `restore_state` into an `optional` would be good, but not blocking)

mattkae:
> I wonder if the state machine here is getting sufficiently complex that it should be deliberately modelled?
> 
> A more invasive, but more principled approach, would be to save the entire window manager state when initiating the switcher, let the switcher fiddle around as it pleased, and then on completion take the saved state, apply _only_ the finally-selected change, and switch to that.
> 
> I'm not blocking on doing that, just floating that it might need to be done eventually.
> 
> (The only _blocking_ thing here is removing the `printf` debugging; I think turning `restore_state` into an `optional` would be good, but not blocking)

I more or less agree with this too, but I think it should wait to move this into the `FocusController`. The `ApplicationSelector` is actually a poor idea all around, and I think this would be much much better if it were the job of the `FocusController` to do this sort of focus-related stuff. I began prototyping that last week, but it will have to wait for me to have free cycles to make it better.

src/miral/application_selector.cpp
mattkae: When we select the active window, the current state of the window _might_ change state if it was previously minimized. It will go from minimized to restored, but we want minimized to be saved.

On top of that, when we focus a new window, we restore the state of the previously-selected window using `restore_state`, so we don't want to override it before we call `select_active_window`

src/miral/application_selector.h
mattkae: Agreed

src/miral/application_selector.cpp
```diff
@@ -141,6 +159,7 @@ auto ApplicationSelector::complete() -> Window
 
     originally_selected_it = focus_list.end();
     is_active_ = false;
+    restore_state = mir_window_state_unknown;
```
AlanGriffiths, line 162:
```suggestion
    restore_state.reset();
```

AlanGriffiths (approved):
This looks like another special case handler for our growing pile of window management special cases.

It clearly fixes one usecase and doesn't obviously break others.

It would be good if it obviously didn't break usecases, but I don't see how we can achieve that without:

1. A test framework for window management behaviours
2. A lot of tests of window management behaviours
3. Refactoring and then reworking the window management code

Currently we're afraid of 3 because we lack 1 & 2

PR #3415, opened 2024-06-05
Add input configuration options

AlanGriffiths:
--mouse-acceleration arg              Select acceleration profile for mice 
                                         and trackballs [none, adaptive]
   --mouse-cursor-acceleration-bias arg  Constant factor (+1) to velocity or 
                                         bias to the acceleration curve within 
                                         the range [-1.0, 1.0] for mice
   --mouse-scroll-speed-scale arg        Scales mice scroll events, use negative
                                         values for natural scrolling
   --touchpad-disable-while-typing arg   Disable touchpad while typing on 
                                         keyboard configuration [true, false]
   --touchpad-tap-to-click arg           Enable or disable tap-to-click on this 
                                         device, with a default mapping of 1, 2,
                                         3 finger tap mapping to left, right, 
                                         middle click, respectively [true, 
                                         false]
   --touchpad-cursor-acceleration-bias arg
                                         Constant factor (+1) to velocity or 
                                         bias to the acceleration curve within 
                                         the range [-1.0, 1.0] for touchpads
   --touchpad-scroll-speed-scale arg     Scales touchpad scroll events, use 
                                         negative values for natural scrolling
   --touchpad-scroll-mode arg            Select scroll mode for touchpads: 
                                         [edge, two-finger, button-down]
   --touchpad-click-mode arg             Select click mode for touchpads: 
                                         [{area, finger-count}]


AlanGriffiths:
> Could we tell what the defaults are?

Not until the input system starts and libinput can report what the devices and their defaults are.

PR #3416, opened 2024-06-06
feature: a test harness for window managers + tests for the MinimalWindowManager

mattkae:
This PR establishes an initial test harness that is suitable for `MinimalWindowManager`. We might need more harnessing for `FloatingWindowManager`, but that should come as a follow up if we need it. 

## What's new?
- Built out the `WindowManagementTestHarness`
- Wrote some initial integration tests for the `MinimalWindowManager`

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -77,6 +77,7 @@ include_directories(
 
 option(MIR_BUILD_ACCEPTANCE_TESTS "Build acceptance tests" ON)
 option(MIR_BUILD_INTEGRATION_TESTS "Build integration tests" ON)
+option(MIR_BUILD_WINDOW_MANAGEMENT_TESTS "Build window management tests" ON)
 option(MIR_BUILD_PERFORMANCE_TESTS "Build performance tests" ON)
 option(MIR_BUILD_MIRAL_TESTS "Build miral tests" ON)
 option(MIR_BUILD_UNIT_TESTS "Build unit tests" ON)
@@ -108,6 +109,10 @@ include_directories(
 if (MIR_BUILD_INTEGRATION_TESTS)
   add_subdirectory(integration-tests/)
 endif (MIR_BUILD_INTEGRATION_TESTS)
+
+if (MIR_BUILD_WINDOW_MANAGEMENT_TESTS)
+  add_subdirectory(window_management_tests/)
+endif(MIR_BUILD_WINDOW_MANAGEMENT_TESTS)
 
 if (MIR_BUILD_UNIT_TESTS)
   add_subdirectory(unit-tests/)
--- a/tests/include/mir_test_framework/window_management_test_harness.h
+++ b/tests/include/mir_test_framework/window_management_test_harness.h
@@ -1,0 +1,71 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_TEST_FRAMEWORK_WINDOW_MANAGEMENT_TEST_HARNESS_H
+#define MIR_TEST_FRAMEWORK_WINDOW_MANAGEMENT_TEST_HARNESS_H
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <miral/window_management_policy.h>
+#include <miral/application.h>
+#include <miral/window_manager_tools.h>
+#include <mir/shell/surface_specification.h>
+#include "mir/test/doubles/fake_display_configuration_observer_registrar.h"
+#include "mir_test_framework/headless_in_process_server.h"
+
+namespace mir::scene
+{
+class Surface;
+class Session;
+}
+
+namespace mir_test_framework
+{
+using WindowManagementPolicyBuilder =
+    std::function<std::unique_ptr<miral::WindowManagementPolicy>(miral::WindowManagerTools const& tools)>;
+
+/// A harness for window management testing. To use, extend this class and provide the
+/// necessary virtual methods.
+class WindowManagementTestHarness : public mir_test_framework::HeadlessInProcessServer
+{
+public:
+    WindowManagementTestHarness();
+    void SetUp() override;
+    void TearDown() override;
+
+    auto open_application(std::string const& name) -> miral::Application;
+
+    /// Create a window with the provided spec
+    auto create_window(
+        miral::Application const&,
+        mir::shell::SurfaceSpecification spec) -> miral::Window;
+
+    void publish_event(MirEvent const& event);
+    void request_resize(miral::Window const&, MirInputEvent const*, MirResizeEdge);
+    void request_move(miral::Window const&, MirInputEvent const*);
+
+    auto focused_surface() -> std::shared_ptr<mir::scene::Surface>;
+    auto tools() -> miral::WindowManagerTools const&;
+    virtual auto get_builder() -> WindowManagementPolicyBuilder = 0;
+    virtual auto get_output_rectangles() -> std::vector<mir::geometry::Rectangle> = 0;
+
+    struct Self;
+    std::shared_ptr<Self> self;
+};
+
+}
+
+#endif //MIR_TEST_FRAMEWORK_WINDOW_MANAGEMENT_TEST_HARNESS_H
--- a/tests/mir_test_framework/CMakeLists.txt
+++ b/tests/mir_test_framework/CMakeLists.txt
@@ -5,6 +5,7 @@ include_directories(
   ${PROJECT_SOURCE_DIR}/src/include/server
   ${PROJECT_SOURCE_DIR}/src/include/client
   ${PROJECT_SOURCE_DIR}/include/renderers/sw
+  ${PROJECT_SOURCE_DIR}/include/miral
   ${CMAKE_SOURCE_DIR}
 )
 
@@ -45,6 +46,7 @@ add_library(mir-public-test-framework OBJECT
   test_display_server.cpp  ${PROJECT_SOURCE_DIR}/include/test/miral/test_display_server.h
   ${PROJECT_SOURCE_DIR}/tests/include/mir_test_framework/mmap_wrapper.h
   mmap_wrapper.cpp
+  window_management_test_harness.cpp ${PROJECT_SOURCE_DIR}/tests/include/mir_test_framework/window_management_test_harness.h
 )
 
 target_link_libraries(mir-public-test-framework
--- a/tests/mir_test_framework/window_management_test_harness.cpp
+++ b/tests/mir_test_framework/window_management_test_harness.cpp
@@ -1,0 +1,191 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "mir/test/doubles/fake_display.h"
+#include "mir/test/doubles/stub_buffer_stream.h"
+#include "mir_test_framework/window_management_test_harness.h"
+#include <mir/executor.h>
+#include <mir/scene/surface.h>
+#include <mir/scene/surface_observer.h>
+#include <mir/shell/surface_specification.h>
+#include <mir/shell/shell.h>
+#include <mir/wayland/weak.h>
+#include <miral/window_manager_tools.h>
+#include <miral/set_window_management_policy.h>
+
+#include <queue>
+
+namespace ms = mir::scene;
+namespace msh = mir::shell;
+namespace mf = mir::frontend;
+namespace mi = mir::input;
+namespace mtd = mir::test::doubles;
+namespace mg = mir::graphics;
+namespace mc = mir::compositor;
+namespace geom = mir::geometry;
+using namespace testing;
+
+struct mir_test_framework::WindowManagementTestHarness::Self : public ms::SurfaceObserver
+{
+    explicit Self() : tools{nullptr}
+    {
+    }
+
+    void attrib_changed(ms::Surface const*, MirWindowAttrib, int) override {}
+    void window_resized_to(ms::Surface const*, geom::Size const&) override {}
+    void content_resized_to(ms::Surface const*, geom::Size const&) override {}
+    void moved_to(ms::Surface const*, geom::Point const&) override {}
+    void hidden_set_to(ms::Surface const*, bool) override {}
+    void frame_posted(ms::Surface const*, geom::Rectangle const&) override {}
+    void alpha_set_to(ms::Surface const*, float) override {}
+    void orientation_set_to(ms::Surface const*, MirOrientation) override {}
+    void transformation_set_to(ms::Surface const*, glm::mat4 const&) override {}
+    void reception_mode_set_to(ms::Surface const*, mir::input::InputReceptionMode) override {}
+
+    void cursor_image_set_to(
+        ms::Surface const*,
+        std::weak_ptr<mg::CursorImage> const&) override {}
+
+    void client_surface_close_requested(ms::Surface const* surf) override
+    {
+        auto it = std::find_if(known_surfaces.begin(), known_surfaces.end(), [surf](std::shared_ptr<ms::Surface> const& s)
+        {
+            return s.get() == surf;
+        });
+
+        if (it != known_surfaces.end())
+        {
+            surfaces_to_destroy.push_back(*it);
+            known_surfaces.erase(it);
+        }
+    }
+
+    void renamed(ms::Surface const*, std::string const&) override {}
+
+    void cursor_image_removed(ms::Surface const*) override {}
+    void placed_relative(ms::Surface const*, geom::Rectangle const&) override {}
+    void input_consumed(ms::Surface const*, std::shared_ptr<MirEvent const> const&) override {}
+    void application_id_set_to(ms::Surface const*, std::string const&) override {}
+    void depth_layer_set_to(ms::Surface const*, MirDepthLayer) override {}
+    void entered_output(ms::Surface const*, mg::DisplayConfigurationOutputId const&) override {}
+    void left_output(ms::Surface const*, mg::DisplayConfigurationOutputId const&) override {}
+
+    std::vector<std::shared_ptr<mc::BufferStream>> streams;
+    std::vector<std::shared_ptr<ms::Surface>> known_surfaces;
+    std::vector<std::shared_ptr<ms::Surface>> surfaces_to_destroy;
+    mtd::FakeDisplay* display = nullptr;
+    miral::WindowManagerTools tools;
+};
+
+mir_test_framework::WindowManagementTestHarness::WindowManagementTestHarness()
+    : self{std::make_shared<Self>()}
+{
+}
+
+void mir_test_framework::WindowManagementTestHarness::SetUp()
+{
+    miral::SetWindowManagementPolicy policy([&](miral::WindowManagerTools const& tools)
+    {
+        self->tools = tools;
+        return get_builder()(tools);
+    });
+    policy(server);
+
+
+    auto fake_display = std::make_unique<mtd::FakeDisplay>(get_output_rectangles());
+    self->display = fake_display.get();
+    preset_display(std::move(fake_display));
+
+    mir_test_framework::HeadlessInProcessServer::SetUp();
+}
+
+void mir_test_framework::WindowManagementTestHarness::TearDown()
+{
+    mir_test_framework::HeadlessInProcessServer::TearDown();
+}
+
+auto mir_test_framework::WindowManagementTestHarness::open_application(
+    std::string const& name) -> miral::Application
+{
+    return server.the_shell()->open_session(
+        __LINE__,
+        mir::Fd{mir::Fd::invalid},
+        name,
+        std::shared_ptr<mir::frontend::EventSink>());;
+}
+
+auto mir_test_framework::WindowManagementTestHarness::create_window(
+    miral::Application const& session,
+    mir::shell::SurfaceSpecification spec) -> miral::Window
+{
+    auto stream = std::make_shared<mir::test::doubles::StubBufferStream>();
+    self->streams.push_back(stream);
+    spec.streams = std::vector<msh::StreamSpecification>();
+    spec.streams.value().push_back(msh::StreamSpecification{
+        stream,
+        geom::Displacement{},
+    });
+
+    self->display->configuration()->for_each_output([&](mg::DisplayConfigurationOutput const& output)
+    {
+        spec.output_id = output.id;
+    });
+    auto surface = server.the_shell()->create_surface(
+        session,
+        {},
+        spec,
+        self,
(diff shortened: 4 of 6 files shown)


AlanGriffiths:
@mattkae any idea why "Coverage" failed with:

```
[ FAILED ] 2 tests, listed below:
[ FAILED ] Process.a_main_fn_is_executed
[ FAILED ] Process.a_successful_exit_function_succeeds
```

It isn't like it runs different tests?

tests/include/mir/test/doubles/stub_shell_report.h
```diff
@@ -0,0 +1,53 @@
...
+    MirWindowAttrib, int) override {}
+    void destroying_surface(mir::scene::Session const&, mir::scene::Surface const&) override {}
+    void started_prompt_session(mir::scene::PromptSession const&, mir::scene::Session const&) override {}
+    void added_prompt_provider(mir::scene::PromptSession const&, mir::scene::Session const&) override {}
+    void stopping_prompt_session(mir::scene::PromptSession const&) override {}
+    void adding_display(mir::geometry::Rectangle const&) override {}
+    void removing_display(mir::geometry::Rectangle const&) override {}
+    void input_focus_set_to(mir::scene::Session const*, mir::scene::Surface const*) override {}
+    void surfaces_raised(mir::shell::SurfaceSet const&) override {}
+};
```
AlanGriffiths, line 48: This looks very similar to the existing `mir::report::null::ShellReport` implementation. Do we need both?

tests/integration-tests/CMakeLists.txt
```diff
@@ -55,10 +56,13 @@ target_link_libraries(
   mir-test-doubles-static
 
   mircommon
+  miral
+  miral-internal
```
AlanGriffiths, line 60: I wonder if this would be better as a separate, more focused, test executable? (Instead of bundling a lot of dependencies into an existing one)

AlanGriffiths (commented):
Overall, this is the sort of thing I imagined. Not finished looking through the detail yet. But mentioning some nits before my EOD

tests/integration-tests/CMakeLists.txt
AlanGriffiths: Yes

AlanGriffiths:
OK, my problems here are not so much with the test harness itself, but with the restructuring needed to support it.

And that actually comes from the legacy structure of the project.

1. There's an "undefined behaviour" problem with including both `miral` and `miral-internal`, as `miral` include `miral-internal` (internally) which gives two instances of any code from `miral-internal` (breaks the ODR). (I am aware that, as these copies are identical and not exported, it is unlikely we're doing anything that breaks in practice.)

2. Publishing even more mirserver internals "just" to _use_ them in a test harness seems wrong. We should have a better way. (I did a slightly better job with miral, as there is a separate `miral-internal` library that could be included in test binaries, but as evidenced by /1/ it isn't perfect.)

I need to think about what the options are and either propose a solution or reconcile myself to the "ugliness".

mattkae:
Symbols issue will be fixed by https://github.com/canonical/mir/pull/3423

mattkae:
> OK, my problems here are not so much with the test harness itself, but with the restructuring needed to support it.
> 
> And that actually comes from the legacy structure of the project.
> 
>     1. There's an "undefined behaviour" problem with including both `miral` and `miral-internal`, as `miral` include `miral-internal` (internally) which gives two instances of any code from `miral-internal` (breaks the ODR). (I am aware that, as these copies are identical and not exported, it is unlikely we're doing anything that breaks in practice.)
> 
>     2. Publishing even more mirserver internals "just" to _use_ them in a test harness seems wrong. We should have a better way. (I did a slightly better job with miral, as there is a separate `miral-internal` library that could be included in test binaries, but as evidenced by /1/ it isn't perfect.)
> 
> 
> I need to think about what the options are and either propose a solution or reconcile myself to the "ugliness".

I have moved the tests to their own suite, so let me know what you want to do in this domain.

tests/window_management_tests/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+target_link_libraries(
+  mir_window_management_tests
+
+  mir-test-static
+  mir-test-framework-static
+  mir-test-doubles-static
+
+  mircommon
+  miral
+  miral-internal
```
AlanGriffiths, line 50: Linking both `miral` (which includes `miral-internal` internally) and `miral-internal` risk multiple definitions. (Which is at best non-portable, and at worst causes hard to track down bugs.)

tests/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ include_directories(
 
 option(MIR_BUILD_ACCEPTANCE_TESTS "Build acceptance tests" ON)
 option(MIR_BUILD_INTEGRATION_TESTS "Build integration tests" ON)
+option(MIR_BUILD_WINDOW_MANAGEMENT_TESTS "Build integration tests" ON)
```
AlanGriffiths, line 80:
```suggestion
option(MIR_BUILD_WINDOW_MANAGEMENT_TESTS "Build window management tests" ON)
```

tests/integration-tests/CMakeLists.txt
```diff
@@ -55,10 +55,13 @@ target_link_libraries(
...
```
AlanGriffiths, line 59: Linking both `miral` (which includes `miral-internal` internally) and `miral-internal` risk multiple definitions. (Which is at best non-portable, and at worst causes hard to track down bugs.)

tests/include/mir_test_framework/window_management_test_harness.h
```diff
@@ -0,0 +1,116 @@
...
+        mir::geometry::Width const&,
+        mir::geometry::Height const&);
+    void publish_event(MirEvent const& event);
+    void request_resize(miral::Window const&, MirInputEvent const*, MirResizeEdge);
+
+    /// Returns the underlying abstract shell that the harness interacts with.
+    /// WARNING: It is not advised to send commands directly to the shell, as
+    /// that circumvents the harness. Please only use this to confirm state
+    /// about the shell (e.g. the focused surface).
+    [[nodiscard]] std::shared_ptr<mir::shell::AbstractShell> const& get_shell() const;
```
AlanGriffiths, line 97: Would returning a `shared_ptr<AbstractShell const>` help police this policy?

tests/include/mir_test_framework/window_management_test_harness.h
```diff
@@ -0,0 +1,116 @@
...
+
+    /// Returns the underlying surface stack that the harness interacts with.
+    /// WARNING: It is not advised to send commands directly to the surface stack, as
+    /// that circumvents the harness. Please only use this to confirm state
+    /// about the stack (e.g. the focus order).
+    [[nodiscard]] std::shared_ptr<mir::scene::SurfaceStack> const& get_surface_stack() const;
```
AlanGriffiths, line 103: Would returning a `shared_ptr<SurfaceStack const>` help police this policy?

tests/include/mir_test_framework/window_management_test_harness.h
AlanGriffiths: Also, the only use is `EXPECT_EQ(get_shell()->focused_surface()...)` which could be handled with a forwarding function. (We also prefer the "fluent" test assertions. I.e. `EXPECT_THAT(focused_surface(), Eq(...))`

AlanGriffiths (changes requested):
I'm still experimenting with ideas to clean up using our library internals for testing

RAOF (commented):
This looks like a good start (I've some code-style nits with not using our `auto foo() -> bar` default in places).

This is not *quite* what I was expecting - rather than validating that the expected WM hooks are called at the appropriate time, I was expecting that the WM *state* would be validated at the end. Something like
```
<create windows>
EXPECT_THAT(stacking_order(), Eq({window1, window2, window3, window4});
<emit alt-tab sequence>
EXPECT_THAT(stacking_order(), Eq({window2, window1, window3, window4})
```
and suchlike.

I think that would be both simpler to implement and would be more stable over code changes (we're not testing the implementation details of *how* WM happens, we're testing *what* window management happens).

This PR also *basically* implements that, right? You'd turn `get_shell` et al into a set of `focused_surface()`, `stacking_order()`, etc methods, as Alan suggests, and then maybe (re)move a bunch of code (like the whole `WindowManagementVerifier` infrastructure) into a separate set of tests?

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -0,0 +1,366 @@
...
+    auto const app = open_application("test");
+}
+
+TEST_F(MinimalWindowManagerTest, advise_new_window_is_called_when_a_window_is_created)
+{
+    EXPECT_CALL(*verifier, advise_new_app(_));
+    auto const app = open_application("test");
+    EXPECT_CALL(*verifier, advise_new_window(_));
+    create_window(app, geom::Width{100}, geom::Height{100});
+}
```
RAOF, line 94: This is not actually testing `MinimalWindowManager`, is it? As far as I can tell, this is testing `VerifyableWindowManagementPolicy`? (as in: as far as I can tell, the only change to `MinimalWindowManager` that would cause this test to fail is if `MinimalWindowManager::advice_new_app` threw an exception?)

tests/window_management_tests/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+  ${PROJECT_SOURCE_DIR}/src/include/common
+  ${PROJECT_SOURCE_DIR}/src/include/server
+  ${PROJECT_SOURCE_DIR}/src/include/client
+  ${PROJECT_SOURCE_DIR}/src/include/gl
+  ${PROJECT_SOURCE_DIR}/include/renderers/sw
+)
+
+set(WINDOW_MANAGEMENT_TESTS_SRCS
+  test_minimal_window_manager.cpp
+)
```
AlanGriffiths, line 17: There doesn't seem any point to this indirection

tests/window_management_tests/CMakeLists.txt
```diff
@@ -0,0 +1,74 @@
...
+
+link_directories(${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+mir_add_wrapped_executable(mir_window_management_tests NOINSTALL
+  ${WINDOW_MANAGEMENT_TESTS_SRCS}
+  ${MIR_SERVER_OBJECTS}
+  ${MIR_PLATFORM_OBJECTS}
```
AlanGriffiths, line 23: Are these useful? The corresponding shared libraries are already linked implicitly which risks having two definitions at runtime

tests/window_management_tests/CMakeLists.txt
AlanGriffiths: See #3424

tests/include/mir_test_framework/window_management_test_harness.h
mattkae: True, we can expose passthrough methods here instead of the whole object :+1:

tests/include/mir_test_framework/window_management_test_harness.h
mattkae: We only use the one method, so I think it is best to proxy and expose _just_ what we need

mattkae:
> This looks like a good start (I've some code-style nits with not using our `auto foo() -> bar` default in places).
> 
> This is not _quite_ what I was expecting - rather than validating that the expected WM hooks are called at the appropriate time, I was expecting that the WM _state_ would be validated at the end. Something like
> 
> ```
> <create windows>
> EXPECT_THAT(stacking_order(), Eq({window1, window2, window3, window4});
> <emit alt-tab sequence>
> EXPECT_THAT(stacking_order(), Eq({window2, window1, window3, window4})
> ```
> 
> and suchlike.
> 
> I think that would be both simpler to implement and would be more stable over code changes (we're not testing the implementation details of _how_ WM happens, we're testing _what_ window management happens).
> 
> This PR also _basically_ implements that, right? You'd turn `get_shell` et al into a set of `focused_surface()`, `stacking_order()`, etc methods, as Alan suggests, and then maybe (re)move a bunch of code (like the whole `WindowManagementVerifier` infrastructure) into a separate set of tests?

Yeah I agree. The `WindowManagementVerifier` got a little out of hand. I am going to remove it, but the rest is still valid :smile:

RAOF (commented):
Hm. I *think* this could be dramatically simplified by using `mtf::HeadlessTest` and `mtf::add_fake_input_device`?

What the `WindowManagementTestHarness` requires is the ability to inject input events (provided by `mtf::FakeInputDevice`) and access to `the_shell()` and `the_surface_stack()`, both of which are already provided by the `mir::Server` you can get from `mtf::HeadlessTest`. Oh, and to set the actual window manager under test, which you can do with `mir::Server::override_the_window_manager_builder`?

That way you wouldn't need to expose a bunch of extra stuff, or do dummy implementations of some of the stuff.

mattkae:
> Hm. I _think_ this could be dramatically simplified by using `mtf::HeadlessTest` and `mtf::add_fake_input_device`?
> 
> What the `WindowManagementTestHarness` requires is the ability to inject input events (provided by `mtf::FakeInputDevice`) and access to `the_shell()` and `the_surface_stack()`, both of which are already provided by the `mir::Server` you can get from `mtf::HeadlessTest`. Oh, and to set the actual window manager under test, which you can do with `mir::Server::override_the_window_manager_builder`?
> 
> That way you wouldn't need to expose a bunch of extra stuff, or do dummy implementations of some of the stuff.

Hm, that is a really good point! We could indeed just wrap `HeadlessTest` and have nearly the same functionality (albeit a bit heaver, bordering on end-to-end test).  I can't think of any push back off of the top of my head, but maybe @AlanGriffiths can.

AlanGriffiths:
> Hm, that is a really good point! We could indeed just wrap `HeadlessTest` and have nearly the same functionality (albeit a bit heaver, bordering on end-to-end test). I can't think of any push back off of the top of my head, but maybe @AlanGriffiths can.

It is a good point: I hadn't recognised the commonality

mattkae:
@RAOF and @AlanGriffiths : Yup, using the `HeadlessInProcessServer` was indeed the answer. We still need a `SurfaceObserver` to get the timings right (usually this would be a Wayland-specific `SurfaceObserver`), but this has simplified everything else pretty dramatically.

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    void run()
+    {
+        while (next()) {}
+    }
+
+private:
+    std::queue<std::function<void()>> work_queue;
+};
```
RAOF, line 67: I don't think you need the `WorkQueue` anymore? There's no longer recursion into multiple observers from one observation.

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -0,0 +1,213 @@
...
+            return s.get() == surf;
+        });
+
+        if (it != known_surfaces.end())
+        {
+            work_queue.enqueue([shell=harness->server.the_shell(), surface=*it]()
+            {
+                shell->destroy_surface(surface->session().lock(), surface);
+            });
+        }
```
RAOF, line 106: It's a pity this can't be just
```
shell->destroy_surface(surf->session().lock(), surf)
```
but
1. That would require changing the `destroy_surface` signature, which is out-of-scope here, and
2. I'm not sure that dropping the `shared_ptr` from the `destroy_surface` signature would be safe anyway.

RAOF (changes requested):
This is a great simplification. I *think* this can be simplified *even further* (and could be slightly nicer stylistically by using the `ms::` etc namespace aliases consistently throughout).

The only blocking thing here is the comment in `::SetUp()`.

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -0,0 +1,213 @@
...
+{
+    miral::SetWindowManagementPolicy policy(get_builder());
+    policy(server);
+
+    mir_test_framework::HeadlessInProcessServer::SetUp();
+    self = std::make_shared<Self>(this);
+
+    auto fake_display = std::make_unique<mtd::FakeDisplay>(get_output_rectangles());
+    self->display = fake_display.get();
+    preset_display(std::move(fake_display));
```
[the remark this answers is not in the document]
mattkae: `tests/miral/active_outputs.cpp` has the same order for some reason? I was following the example, but I _would_ imagine that you are correct!

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -0,0 +1,213 @@
...
+        std::shared_ptr<mir::frontend::EventSink>());;
+}
+
+auto mir_test_framework::WindowManagementTestHarness::create_window(
+    miral::Application const& session,
+    mir::geometry::Width const& width,
+    mir::geometry::Height const& height) -> miral::Window
+{
+    auto stream = std::make_shared<mir::test::doubles::StubBufferStream>();
+    self->streams.push_back(stream);
```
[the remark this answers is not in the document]
mattkae: The reason that I kept it around is to keep a valid reference alive. `StreamSpecification` takes a `weak_ptr`, so I need to keep it somewhere

tests/mir_test_framework/window_management_test_harness.cpp
mattkae: We still need the `WorkQueue` because there will be recursion with event handling, for example.

tests/include/mir_test_framework/window_management_test_harness.h
```diff
@@ -0,0 +1,76 @@
...
+    void SetUp() override;
+    void TearDown() override;
+
+    auto open_application(std::string const& name) -> miral::Application;
+
+    /// Create a regular application window with the provided width and height
+    auto create_window(
+        miral::Application const&,
+        mir::geometry::Width const&,
+        mir::geometry::Height const&) -> miral::Window;
```
AlanGriffiths, line 54: Do we need this alongside the more general overload? And, if we do, wouldn't `mg::Size` be clearer?

mattkae: Eh you're probably right. This was a QOL thing, but it's probably too opaque

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -0,0 +1,487 @@
...
+    MirInputEventModifiers const modifiers{mir_input_event_modifier_alt};
+    auto const event = mir::events::make_key_event(
+        mir_input_event_type_key,
+        event_timestamp,
+        action,
+        keysym,
+        scan_code,
+        modifiers);
+    publish_event(*event);
+}
```
RAOF, line 97: Where is the check that the window has been closed?

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    auto next() -> bool
+    {
+        if (work_queue.empty())
+            return false;
+
+        auto const& f = work_queue.front();
+        f();
+        work_queue.pop();
```
RAOF, line 58: This can result in infinite recursion, right? Almost all usages of the workqueue are of the form `queue.enqueue(...); queue.run()`, and if that gets run *from* the workqueue it'll push something to the back of the queue and then run itself again?

RAOF (changes requested):
You appear to have accidentally a test :wink: 

The workqueue still appears weird to me; the only use is to have `client_surface_close_requested` call `destroy_surface` in a deferred fashion? So the only time it's actually interesting is for `publish_event`? And *that* matters because the WM takes a lock when processing the input event, and `shell->destroy_surface` would *also* try to take that lock.

Right?

mattkae:
> You appear to have accidentally a test 😉
> 
> The workqueue still appears weird to me; the only use is to have `client_surface_close_requested` call `destroy_surface` in a deferred fashion? So the only time it's actually interesting is for `publish_event`? And _that_ matters because the WM takes a lock when processing the input event, and `shell->destroy_surface` would _also_ try to take that lock.
> 
> Right?

Yes exactly.

tests/window_management_tests/test_minimal_window_manager.cpp
mattkae: True :wink:

tests/mir_test_framework/window_management_test_harness.cpp
mattkae: Yeah I agree that it's kind of nasty... I am trying to find a good way to avoid both the nastiness of the queue and the nastiness of asynchronicity

mattkae: After thinking about this deeper, it _can_ result in recursion, but this isn't necessarily a bad thing. The `WorkQueue` will just continually be run recursively, which will work.

If you end up in a situation where the recursion is _infinite_, then that suggests that your WM is doing something wrong. e.g. if every time you resize a window, you also move it, and every time you move a window, you also resize it, then that is on you

mattkae:
@RAOF: I see no clever way around the `WorkQueue` at the moment. This queue at least avoids some complexities with threading, which I'm a fan of. I think that users can get themselves in sticky situations only when their WM is doing something wrong.

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -0,0 +1,232 @@
...
+    miral::Window const& window,
+    MirInputEvent const* event,
+    MirResizeEdge edge)
+{
+    self->work_queue.enqueue([&]()
+    {
+        auto surface = window.operator std::shared_ptr<ms::Surface>();
+        server.the_shell()->request_resize(surface->session().lock(), surface, event, edge);
+    });
+    self->work_queue.run();
```
RAOF, line 210:
```suggestion
    auto surface = window.operator std::shared_ptr<ms::Surface>();
    server.the_shell()->request_resize(surface->session().lock(), surface, event,     
```

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+void mir_test_framework::WindowManagementTestHarness::request_move(
+    miral::Window const& window, MirInputEvent const* event)
+{
+    self->work_queue.enqueue([&]()
+    {
+        auto surface = window.operator std::shared_ptr<ms::Surface>();
+        server.the_shell()->request_move(surface->session().lock(), surface, event);
+    });
+    self->work_queue.run();
```
RAOF, line 221: ...and likewise, here
```suggestion
    auto surface = window.operator std::shared_ptr<ms::Surface>();
    server.the_shell()->request_move(surface->session().lock(), surface, event);
```

tests/mir_test_framework/window_management_test_harness.cpp
RAOF: So, the problem I'm talking about is running `f()` *before* running `work_queue.pop()`. If `f()` happens to run `work_queue.run()`, then `work_queue.run()` will call `work_queue.front()`, receive *the same* `f` again, and recurse.

If you popped before calling `f()` (and `std::move()`d out of `front()`), that wouldn't happen.

RAOF (approved):
So, the `WorkQueue` *seems* like it could change from a work queue into just a `std::vector<std::shared_ptr<ms::Surface>> pending_surfaces_to_destroy;` plus a loop at the end of `publish_event` destroying those surfaces.

Neither `request_resize` nor `request_move` actually use the `work_queue`, right? They push a single functor into the queue and then immediately pop that functor off the front and run it - there's no way for `client_surface_close_requested` to be called from either of those codepaths?

I won't block on this, but I feel like that would be simpler?

mattkae:
> So, the `WorkQueue` _seems_ like it could change from a work queue into just a `std::vector<std::shared_ptr<ms::Surface>> pending_surfaces_to_destroy;` plus a loop at the end of `publish_event` destroying those surfaces.
> 
> Neither `request_resize` nor `request_move` actually use the `work_queue`, right? They push a single functor into the queue and then immediately pop that functor off the front and run it - there's no way for `client_surface_close_requested` to be called from either of those codepaths?
> 
> I won't block on this, but I feel like that would be simpler?

I can agree with that :+1: Really only the destroy code needs it for now

mattkae:
@RAOF I also better understand the issue now, misunderstood before. I will pop _before_ calling _f_

mattkae:
@RAOF: Now that I fully grasp your point, you were right all along :smile:

tests/mir_test_framework/CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ include_directories(
   ${PROJECT_SOURCE_DIR}/src/include/server
   ${PROJECT_SOURCE_DIR}/src/include/client
   ${PROJECT_SOURCE_DIR}/include/renderers/sw
+  ${PROJECT_SOURCE_DIR}/include/miral
+  ${PROJECT_SOURCE_DIR}/src/miral
```
AlanGriffiths, line 9: `${PROJECT_SOURCE_DIR}/src/miral` is not needed

AlanGriffiths (approved):
One nit, but otherwise good to go

PR #3418, opened 2024-06-07
Associate an explicit source rect/destination size with each submitted buffer

RAOF:
This associates two pieces of information with each buffer submitted to a `BufferStream`:
* The logical output size the buffer should have (rather than calculating this from the physical buffer size and the surface scale), and
* The subrectangle of the physical buffer that should be used.

This is infrastructure for supporting `wp_viewporter`, which allows clients to explicitly set these two properties.


AlanGriffiths:
Apart from the conflicts (because it includes stuff that landed) this looks sane. Will give it a final look once conflicts are resolved

src/server/symbols.map
```diff
@@ -234,6 +234,8 @@ global:
     mir::check_for_termination_exception*;
     mir::clear_termination_exception*;
     mir::compositor::BufferStream::?BufferStream*;
+    mir::compositor::BufferStream::Submission::?Submission*;
+    mir::compositor::BufferStream::Submission::Submission*;
```
AlanGriffiths, line 238: New symbols after 2.17 don't belong in MIR_SERVER_INTERNAL_2.17

AlanGriffiths (approved):
Looks sensible, and I don't see it breaking anything

PR #3423, opened 2024-06-11
bugfix: ignoring certain miral symbols purposefully due to inlining

mattkae:
## What's new?
- The symbols generator script now has a "quirk" check to filter out symbols that we do not want to generate
- Removed a few miral symbols purposefully because they cause us headaches

### DIFF
--- a/tools/symbols_map_generator/main.py
+++ b/tools/symbols_map_generator/main.py
@@ -25,6 +25,12 @@ import os
 
 _logger = logging.getLogger(__name__)
 
+HIDDEN_SYMBOLS = {
+    "miral::SessionLockListener::?SessionLockListener*;",
+    "miral::WaylandExtensions::Context::?Context*;",
+    "miral::WaylandExtensions::Context::Context*;",
+    "vtable?for?miral::WaylandExtensions::Context;",
+}
 
 class HeaderDirectory(TypedDict):
     path: str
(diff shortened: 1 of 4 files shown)


tools/symbols_map_generator/main.py
```diff
@@ -147,8 +147,23 @@ def get_namespace_str(node: clang.cindex.Cursor) -> list[str]:
     
     return get_namespace_str(node.semantic_parent) + [spelling]
 
+def symbol_passes_quirk_check(library_name: str, symbol: str):
+    """
+    Checks if the provided
```
Saviq, line 152: Fell asleep? ;D

tools/symbols_map_generator/main.py
```diff
@@ -147,8 +147,23 @@ def get_namespace_str(node: clang.cindex.Cursor) -> list[str]:
...
+    """
+
+    # WaylandExtensions::Context ctor and dtor symbols are inline
+    # but can be exported by unoptimized builds.
+    # This makes the "regenerate" script hide them:
+    if library_name == "miral":
+        return (symbol != "miral::SessionLockListener::?SessionLockListener*;"
+            and symbol != "miral::WaylandExtensions::Context::?Context*;"
+            and symbol != "miral::WaylandExtensions::Context::Context*;"
+            and symbol != "vtable?for?miral::WaylandExtensions::Context;")
```
Saviq, line 162: Couldn't this be a constant `set()` at the top, and then this would be a simple `symbol in HIDDEN_SYMBOLS` or such?

tools/symbols_map_generator/main.py
```diff
@@ -168,17 +183,21 @@ def traverse_ast(node: clang.cindex.Cursor, filename: str, result: set[str]) ->
         namespace_str = "::".join(get_namespace_str(node))
         _logger.debug(f"Emitting node {namespace_str} in file {node.location.file.name}")
 
+        def add_symbol_str(s: str):
+            if symbol_passes_quirk_check(library_name, s):
+                result.add(s)
+
```
Saviq, line 189: Or even simpler, this could be like this?

```
if s in HIDDEN_SYMBOLS:
    return result
```

Saviq (changes requested):
IMO this could be simpler — what does the `library_name` check give us, just not checking the strings if it's not `miral`? All the symbols we're hiding are in miral, and that name is in the strings checked… don't think it's worth the added complexity.

mattkae:
> IMO this could be simpler — what does the `library_name` check give us, just not checking the strings if it's not `miral`? All the symbols we're hiding are in miral, and that name is in the strings checked… is it worth the added complexity?
> 
> If we want it still, I'd still go for a structure like this:
> 
> ```python
> HIDDEN_SYMBOLS = {
>     "miral": (
>         "miral::SessionLockListener::?SessionLockListener*;",
>         "miral::WaylandExtensions::Context::?Context*;",
>         "miral::WaylandExtensions::Context::Context*;",
>         "miral::WaylandExtensions::Context;",
>     )
> }
> ```
> 
> And then a simple:
> 
> ```python
> if s in HIDDEN_SYMBOLS.get(library_name, ()):
>     return result
> ```

Agreed & fixed :+1:

Saviq (approved):
Sures.

PR #3424, opened 2024-06-13
Provide an intermediate OBJECT library for MirAL's external API

AlanGriffiths:
This makes it possible to use all miral objects in tests (without risking duplicate definitions)


mattkae (approved):
Look good to me :+1:

PR #3425, opened 2024-06-14
Wayland server side decorations

### LINKED ISSUE #3371: [Wayland] XDG decoration protocol extension
Implement the XDG decoration protocol extension so that applications wanting SSD use the SSD we have for X11 applications

tarek-y-ismail:
Closes #3371

### DIFF
--- a/src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
+++ b/src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
@@ -35,10 +35,9 @@ namespace frontend
 class ToplevelsWithDecorations
 {
 public:
-    ToplevelsWithDecorations() :
-        toplevels_with_decorations{std::unordered_set<wl_resource*>()}
-    {
-    }
+    ToplevelsWithDecorations() = default;
+    ToplevelsWithDecorations(ToplevelsWithDecorations const&) = delete;
+    ToplevelsWithDecorations& operator=(ToplevelsWithDecorations const&) = delete;
 
     /// \return true if no duplicates existed before insertion, false otherwise.
     bool register_toplevel(wl_resource* toplevel)
--- a/src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
+++ b/src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
@@ -23,6 +23,7 @@
 #include "xdg_output_v1.h"
 #include "xdg_shell_stable.h"
 
+#include <memory>
 #include <unordered_set>
 
 namespace mir
@@ -45,7 +46,7 @@ public:
 
 private:
     void get_toplevel_decoration(wl_resource* id, wl_resource* toplevel) override;
-    std::unordered_set<wl_resource*> toplevels_with_decorations;
+	std::shared_ptr<std::unordered_set<wl_resource*>> toplevels_with_decorations;
 };
 
 class XdgToplevelDecorationV1 : public wayland::XdgToplevelDecorationV1
@@ -84,7 +85,8 @@ void mir::frontend::XdgDecorationManagerV1::Global::bind(wl_resource* new_zxdg_d
 }
 
 mir::frontend::XdgDecorationManagerV1::XdgDecorationManagerV1(wl_resource* resource) :
-    mir::wayland::XdgDecorationManagerV1{resource, Version<1>{}}
+    mir::wayland::XdgDecorationManagerV1{resource, Version<1>{}},
+	toplevels_with_decorations(std::make_shared<std::unordered_set<wl_resource*>>())
 {
 }
 
@@ -92,7 +94,7 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
 {
     using Error = mir::frontend::XdgToplevelDecorationV1::Error;
 
-    if (toplevels_with_decorations.contains(toplevel))
+    if (toplevels_with_decorations->contains(toplevel))
     {
         BOOST_THROW_EXCEPTION(mir::wayland::ProtocolError(
             resource, Error::already_constructed, "Decoration already constructed for this toplevel"));
@@ -104,9 +106,22 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
         BOOST_THROW_EXCEPTION(std::runtime_error("Invalid toplevel pointer"));
     }
 
+    auto decoration = new XdgToplevelDecorationV1{id, tl};
+    toplevels_with_decorations->insert(toplevel);
 
-    new XdgToplevelDecorationV1{id, tl};
-    toplevels_with_decorations.insert(toplevel);
+    decoration->add_destroy_listener([toplevels_with_decorations = this->toplevels_with_decorations, toplevel]()
+                                     { toplevels_with_decorations->erase(toplevel); });
+
+    tl->add_destroy_listener(
+        [toplevels_with_decorations = this->toplevels_with_decorations, toplevel]()
+        {
+            if (toplevels_with_decorations->contains(toplevel))
+            {
+                toplevels_with_decorations->erase(toplevel);
+                BOOST_THROW_EXCEPTION(mir::wayland::ProtocolError(
+                    toplevel, Error::orphaned, "Toplevel destroyed before its attached decoration"));
+            }
+        });
 }
 
 mir::frontend::XdgToplevelDecorationV1::XdgToplevelDecorationV1(
--- a/src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
+++ b/src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
@@ -59,7 +59,7 @@ public:
 private:
     void update_mode(uint32_t new_mode);
 
-    static const uint32_t default_mode = Mode::client_side;
+    static uint32_t const default_mode = Mode::client_side;
 
     mir::frontend::XdgToplevelStable* toplevel;
     uint32_t mode;
@@ -73,8 +73,8 @@ auto mir::frontend::create_xdg_decoration_unstable_v1(wl_display* display)
     return std::make_shared<XdgDecorationManagerV1::Global>(display);
 }
 
-mir::frontend::XdgDecorationManagerV1::Global::Global(wl_display* display)
-    : wayland::XdgDecorationManagerV1::Global::Global{display, Version<1>{}}
+mir::frontend::XdgDecorationManagerV1::Global::Global(wl_display* display) :
+    wayland::XdgDecorationManagerV1::Global::Global{display, Version<1>{}}
 {
 }
 
@@ -83,8 +83,8 @@ void mir::frontend::XdgDecorationManagerV1::Global::bind(wl_resource* new_zxdg_d
     new XdgDecorationManagerV1{new_zxdg_decoration_manager_v1};
 }
 
-mir::frontend::XdgDecorationManagerV1::XdgDecorationManagerV1(wl_resource* resource)
-    : mir::wayland::XdgDecorationManagerV1{resource, Version<1>{}}
+mir::frontend::XdgDecorationManagerV1::XdgDecorationManagerV1(wl_resource* resource) :
+    mir::wayland::XdgDecorationManagerV1{resource, Version<1>{}}
 {
 }
 
@@ -94,8 +94,8 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
 
     if (toplevels_with_decorations.contains(toplevel))
     {
-        BOOST_THROW_EXCEPTION(mir::wayland::ProtocolError(resource, Error::already_constructed,
-                                                          "Decoration already constructed for this toplevel"));
+        BOOST_THROW_EXCEPTION(mir::wayland::ProtocolError(
+            resource, Error::already_constructed, "Decoration already constructed for this toplevel"));
     }
 
     auto* tl = mir::frontend::XdgToplevelStable::from(toplevel);
@@ -110,9 +110,10 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
     toplevels_with_decorations.insert(toplevel);
 }
 
-mir::frontend::XdgToplevelDecorationV1::XdgToplevelDecorationV1(wl_resource* id,
-                                                                mir::frontend::XdgToplevelStable* toplevel)
-    : wayland::XdgToplevelDecorationV1{id, Version<1>{}}, toplevel{toplevel}
+mir::frontend::XdgToplevelDecorationV1::XdgToplevelDecorationV1(
+    wl_resource* id, mir::frontend::XdgToplevelStable* toplevel) :
+    wayland::XdgToplevelDecorationV1{id, Version<1>{}},
+    toplevel{toplevel}
 {
 }
 
@@ -135,12 +136,6 @@ void mir::frontend::XdgToplevelDecorationV1::update_mode(uint32_t new_mode)
     send_configure_event(mode);
 }
 
-void mir::frontend::XdgToplevelDecorationV1::set_mode(uint32_t mode)
-{
-    update_mode(mode);
-}
+void mir::frontend::XdgToplevelDecorationV1::set_mode(uint32_t mode) { update_mode(mode); }
 
-void mir::frontend::XdgToplevelDecorationV1::unset_mode()
-{
-    update_mode(default_mode);
-}
+void mir::frontend::XdgToplevelDecorationV1::unset_mode() { update_mode(default_mode); }
--- a/src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
+++ b/src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
@@ -1,68 +1,69 @@
 #include "xdg_decoration_unstable_v1.h"
 
-#include "mir/shell/surface_specification.h"
 #include "mir/log.h"
+#include "mir/shell/surface_specification.h"
 
 #include "xdg-decoration-unstable-v1_wrapper.h"
-#include "xdg-shell_wrapper.h"
 #include "xdg_output_v1.h"
 #include "xdg_shell_stable.h"
 
 #include <optional>
 
-namespace mir {
-namespace frontend {
-class XdgDecorationManagerV1 : public wayland::XdgDecorationManagerV1 
+namespace mir
+{
+namespace frontend
+{
+class XdgDecorationManagerV1 : public wayland::XdgDecorationManagerV1
 {
 public:
     XdgDecorationManagerV1(wl_resource *resource);
 
-    class Global : public wayland::XdgDecorationManagerV1::Global 
+    class Global : public wayland::XdgDecorationManagerV1::Global
     {
-        public:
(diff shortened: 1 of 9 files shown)


src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,131 @@
...
+        mir::frontend::XdgToplevelStable* toplevel;
+        std::optional<uint32_t> mode;
+};
+} // namespace frontend
+} // namespace mir
+  
+auto mir::frontend::create_xdg_decoration_unstable_v1(wl_display* display) 
+-> std::shared_ptr<mir::wayland::XdgDecorationManagerV1::Global> 
+{
+    log_info("Creating XdgDecorationManagerV1::Global");
```
AlanGriffiths, line 54: "Printf debugging" should be stripped

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,131 @@
...
+                spec.server_side_decorated = false;
+                break;
+            case Mode::server_side:
+                spec.server_side_decorated = true;
+                break;
+        }
+    } else {
+        // Client side by default
+        spec.server_side_decorated = false; 
+    }
```
AlanGriffiths, line 105:
```suggestion
    if (mode) 
    {
        switch (*mode)
        {
        case Mode::client_side:
            spec.server_side_decorated = false;
            break;
        case Mode::server_side:
            spec.server_side_decorated = true;
            break;
        }
    } 
    else
    {
        // Client side by default
        spec.server_side_decorated = false; 
    }
```

src/server/frontend_wayland/xdg_decoration_unstable_v1.h
```diff
@@ -0,0 +1,32 @@
...
+#define MIR_FRONTEND_XDG_DECORATION_UNSTABLE_V1_H
+
+#include "xdg-decoration-unstable-v1_wrapper.h"
+
+namespace mir
+{
+namespace frontend
+{
+auto create_xdg_decoration_unstable_v1(
+    wl_display* display)
```
AlanGriffiths, line 27: No need to split this line

src/server/frontend_wayland/xdg_shell_stable.h
```diff
@@ -20,6 +20,8 @@
 #include "xdg-shell_wrapper.h"
 #include "window_wl_surface_role.h"
 
+namespace mw = mir::wayland;
+
```
AlanGriffiths, line 24: Please don't introduce this alias to global namespace. In headers it is better to spell things in full

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "xdg_shell_stable.h"
+
+#include <optional>
+
+namespace mir {
+namespace frontend {
+class XdgDecorationManagerV1 : public wayland::XdgDecorationManagerV1 
+{
+public:
+    XdgDecorationManagerV1(wl_resource *resource);
```
AlanGriffiths, line 18: We put the `*` with the type

AlanGriffiths (changes requested):
There's a style guide at https://mir-server.io/doc/cppguide/

AlanGriffiths (changes requested):
Checking the WAYLAND_DEBUG log, we're never sending a `toplevel_decoration_v1.configure()` event. The client needs this to know whether or not to draw decorations

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,131 @@
...
+void mir::frontend::XdgToplevelDecorationV1::set_mode(uint32_t mode) 
+{
+    log_info(__PRETTY_FUNCTION__); 
+    update_mode(mode);
+}
+
+void mir::frontend::XdgToplevelDecorationV1::unset_mode() 
+{
+    log_info(__PRETTY_FUNCTION__); 
+    update_mode(Mode::server_side);
```
AlanGriffiths, line 119: As our decorations are ugly, I think we should treat use `client_side` here

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,131 @@
...
+            case Mode::server_side:
+                spec.server_side_decorated = true;
+                break;
+        }
+    } else {
+        // Client side by default
+        spec.server_side_decorated = false; 
+    }
+
+    this->toplevel->apply_spec(spec);
```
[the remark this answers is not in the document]
tarek-y-ismail: Yep, that's what was missing. bomber / vlc now seem to work properly. Thanks for pointing it out!

AlanGriffiths (commented):
This looks like the basics are working. Once you're satisfied with "error handling" etc please mark the PR "ready"

As discussed, there should be some corresponding tests in wlcs.

AlanGriffiths:
**Not needed for this PR**

We should discuss a follow on PR that allows the decoration decision strategy to be customised by shells built with Mir.

There are eight possible strategies: the client has three requests (ssd, csd and unset) to which we can respond in two ways. But half of the "legal" strategies are perverse. What remains are:

1. always-csd - whatever the client requests, send "csd"
2. prefer-csd - if the client requests ssd then send "ssd", otherwise send "csd"
3. prefer-ssd - if the client requests csd then send "csd", otherwise send "ssd"
4. always-ssd - whatever the client requests, send "ssd"

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -79,23 +83,36 @@ void mir::frontend::XdgDecorationManagerV1::Global::bind(wl_resource* new_zxdg_d
...
-    if (tl)
+    if (!tl)
     {
-        new XdgToplevelDecorationV1{id, tl};
+        BOOST_THROW_EXCEPTION(std::runtime_error("Invalid toplevel pointer"));
     }
+
+
+    new XdgToplevelDecorationV1{id, tl};
+    toplevels_with_decorations.insert(toplevel);
```
AlanGriffiths, line 109: We need to clean entries from `toplevels_with_decorations` otherwise it will contain stale entries.

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -118,12 +135,6 @@ void mir::frontend::XdgToplevelDecorationV1::update_mode(uint32_t new_mode)
     send_configure_event(mode);
 }
 
-void mir::frontend::XdgToplevelDecorationV1::set_mode(uint32_t mode)
-{
-    update_mode(mode);
-}
+void mir::frontend::XdgToplevelDecorationV1::set_mode(uint32_t mode) { update_mode(mode); }
```
AlanGriffiths, line 138: I prefer the four line version as it makes it obvious this is a definition when scanning the file. (However I don't think the guide disallows this, so not a blocker)

AlanGriffiths (changes requested):
We need to manage `toplevels_with_decorations` better. Stale entries are problematic for two reasons:

1. decoration objects can be destroyed, leaving the corresponding toplevel incorrectly classified
2. toplevel objects can be destroyed leaving an invalid pointer

Comparisons (during lookup) with invalid pointers are undefined behaviour

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
tarek-y-ismail: You're right, they only solution I have in mind at the moment that would fix the two issues you mentioned is modfying toplevels to hold their decorations and removing `toplevels_with_decorations`.

This would rectify point 1 by checking if the given toplevel already has a decoration assigned to it or not. Point 2 would be rectified by us not having to store pointers to toplevels at all. 

What do you think?

AlanGriffiths: I don't think that solves anything, it just moves the problem around. E.g. when a decoration is destroyed the toplevel needs to be notified (instead of the `toplevels_with_decorations` collection).

We need to handle both decorations and toplevels being destroyed. At the base of out Wayland object hierarchy there's `LifetimeTracker` which allows us to track the life of objects and process their destruction. That should allow you to do the needed checks

AlanGriffiths: "When a toplevel is destroyed, we delete the decoration object" - the decoration object is "owned" by the wayland protocol code. If we delete/destroy it then things will break. I *think* you mean: we remove the toplevel from `toplevels_with_decorations` and, if appropriate, raise a protocol error as we then have an orphaned decoration

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,155 @@
...
+    public:
+        Global(wl_display* display);
+
+    private:
+        void bind(wl_resource* new_zxdg_decoration_manager_v1) override;
+    };
+
+private:
+    void get_toplevel_decoration(wl_resource* id, wl_resource* toplevel) override;
+	std::shared_ptr<std::unordered_set<wl_resource*>> toplevels_with_decorations;
```
AlanGriffiths, line 49: I think that rather than sharing a "naked" set, you should wrap it in a class with `register/unregister` methods. That makes the invariants a lot clearer.

AlanGriffiths, line 49:
```suggestion
    std::shared_ptr<std::unordered_set<wl_resource*>> toplevels_with_decorations;
```

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    decoration->add_destroy_listener([toplevels_with_decorations = this->toplevels_with_decorations, toplevel]()
+                                     { toplevels_with_decorations->erase(toplevel); });
+
+    tl->add_destroy_listener(
+        [toplevels_with_decorations = this->toplevels_with_decorations, toplevel]()
+        {
+            if (toplevels_with_decorations->contains(toplevel))
+            {
+                toplevels_with_decorations->erase(toplevel);
```
AlanGriffiths, line 120: No need to search the container twice: `erase()` returns the count of deleted items.

AlanGriffiths (changes requested):
Some nits pending side-channel discussion

tests/acceptance-tests/wayland/CMakeLists.txt
```diff
@@ -113,6 +113,10 @@ set(EXPECTED_FAILURES
 
   # Not yet implemented, see https://github.com/MirServer/mir/issues/2861
   XdgPopupTest.when_parent_surface_is_moved_a_reactive_popup_is_moved
+
+  # Should throw an exception, but we log instead because of
```
AlanGriffiths, line 117: The exception is an implementation detail. We should report an "orphaned" protocol violation

AlanGriffiths (changes requested):
There are also unresolved comments from an earlier review

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
tarek-y-ismail: Done. I'm not so sure about the naming, but I don't think it's a big deal since it's something internal.

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -32,6 +32,31 @@ namespace mir
...
+
+    /// \return true if only one element was erased, false otherwise.
+    bool unregisterToplevel(wl_resource* toplevel)
+    {
+        return toplevels_with_decorations->erase(toplevel) == 1;
+    }
+
+private:
+    std::shared_ptr<std::unordered_set<wl_resource*>> toplevels_with_decorations;
+};
```
AlanGriffiths, line 58: As you later capture `toplevels_with_decorations` by reference, nothing ever copies a `ToplevelsWithDecorations`, so there's no point to using a shared pointer here.

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -48,7 +73,7 @@ class XdgDecorationManagerV1 : public wayland::XdgDecorationManagerV1
 
 private:
     void get_toplevel_decoration(wl_resource* id, wl_resource* toplevel) override;
-	std::shared_ptr<std::unordered_set<wl_resource*>> toplevels_with_decorations;
+    ToplevelsWithDecorations toplevels_with_decorations;
```
AlanGriffiths, line 76: It is unclear that `XdgDecorationManagerV1` is guaranteed to outlive all the references you capture.

It is clearer to hold `ToplevelsWithDecorations` via a shared pointer and capture that by copy.

```suggestion
    std::shared_ptr<ToplevelsWithDecorations> const toplevels_with_decorations{std::make_shared<ToplevelsWithDecorations>()};
```

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -32,6 +32,31 @@ namespace mir
...
+class ToplevelsWithDecorations
+{
+public:
+    ToplevelsWithDecorations() :
+        toplevels_with_decorations{std::make_shared<std::unordered_set<wl_resource*>>()}
+    {
+    }
+
+    /// \return true if no duplicates existed before insertion, false otherwise.
+    bool registerToplevel(wl_resource* toplevel)
```
AlanGriffiths, line 44: Wrong naming style for this project

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -88,36 +113,37 @@ void mir::frontend::XdgDecorationManagerV1::Global::bind(wl_resource* new_zxdg_d
...
+    if (!toplevels_with_decorations.registerToplevel(toplevel))
+    {
+        BOOST_THROW_EXCEPTION(mir::wayland::ProtocolError(
+            resource, Error::already_constructed, "Decoration already constructed for this toplevel"));
+    }
 
-    decoration->add_destroy_listener([toplevels_with_decorations = this->toplevels_with_decorations, toplevel]()
-                                     { toplevels_with_decorations->erase(toplevel); });
+    decoration->add_destroy_listener(
+        [&toplevels_with_decorations = this->toplevels_with_decorations, toplevel]()
```
AlanGriffiths, line 138: Capturing by reference risks the reference becoming invalid

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -88,36 +113,37 @@ void mir::frontend::XdgDecorationManagerV1::Global::bind(wl_resource* new_zxdg_d
...
+        {
+            toplevels_with_decorations.unregisterToplevel(toplevel);
+        });
 
     tl->add_destroy_listener(
-        [toplevels_with_decorations = this->toplevels_with_decorations, client = this->client, toplevel]()
+        [&toplevels_with_decorations = this->toplevels_with_decorations, client = this->client, toplevel]()
```
AlanGriffiths, line 144: Capturing by reference risks the reference becoming invalid

AlanGriffiths (changes requested):

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,190 @@
...
+{
+namespace frontend
+{
+class ToplevelsWithDecorations
+{
+public:
+    ToplevelsWithDecorations() :
+        toplevels_with_decorations{std::unordered_set<wl_resource*>()}
+    {
+    }
```
AlanGriffiths, line 41: There's no need to copy-construct from a temporary. The only reason to write a default constructor would be disable copying:
```suggestion
    ToplevelsWithDecorations() = default;
    ToplevelsWithDecorations(ToplevelsWithDecorations const&) = delete;
    ToplevelsWithDecorations& operator=(ToplevelsWithDecorations const&) = delete;
```

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,190 @@
...
+    decoration->add_destroy_listener(
+        [toplevels_with_decorations = this->toplevels_with_decorations, toplevel]()
+        {
+            toplevels_with_decorations->unregister_toplevel(toplevel);
+        });
+
+    tl->add_destroy_listener(
+        [toplevels_with_decorations = this->toplevels_with_decorations, client = this->client, toplevel]()
+        {
+            if (!client->is_being_destroyed() && !toplevels_with_decorations->unregister_toplevel(toplevel))
```
AlanGriffiths, line 146: We probably want to unregister regardless of `is_being_destroyed()`? (`&&` short-circuits)

AlanGriffiths (changes requested):

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -47,10 +46,13 @@ class ToplevelsWithDecorations
         return inserted;
     }
 
-    /// \return true if only one element was erased, false otherwise.
+    /// \return true if the toplevel didn't exist in the set, false otherwise.
+    /// The return value is used in the destroy callback of the toplevel to
+    /// check if the toplevel is destroyed before the decoration (orphaned
+    /// decoration).
```
AlanGriffiths, line 52: 1. This is a very weird semantic: returning `false` on success.
2. It is normal to document the preconditions, action, and postconditions of a function. Not where its result happens to be used

AlanGriffiths (changes requested):
Getting there...

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -0,0 +1,193 @@
...
+        [toplevels_with_decorations = this->toplevels_with_decorations, toplevel]()
+        {
+            toplevels_with_decorations->unregister_toplevel(toplevel);
+        });
+
+    tl->add_destroy_listener(
+        [toplevels_with_decorations = this->toplevels_with_decorations, client = this->client, toplevel]()
+        {
+            const auto orphaned_decoration = !toplevels_with_decorations->unregister_toplevel(toplevel);
+            if (!client->is_being_destroyed() && orphaned_decoration)
```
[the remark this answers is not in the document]
tarek-y-ismail: I can't see what is achieved here other than saving a variable declaration (which the compiler likely optimizes anyway). I think the original might be a bit more readable.

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
tarek-y-ismail: I agree. I myself was a bit confused while using this method. 

I've moved the explanation to where the mentioned case is actually handled. I might've detailed things too much, but I think this will save the next person (likely me :)) that works on this code some time.

AlanGriffiths (approved):
LGTM

PR #3430, opened 2024-06-18
(#3414) Don't re-send wl_output.geometry if unchanged

### LINKED ISSUE #3414: Don't re-send wl_output.geometry if unchanged
> Correction: we would be correct to resend `geometry` if any of the properties changed:
> > The event is sent when binding to the output object _and whenever any of the properties change_.

_Originally posted by @AlanGriffiths in https://github.com/canonical/mir/issues/3393#issuecomment-2124532188_

hbatagelo:
Fixes #3414

Removes a duplicate call to `OutputConfigListener::output_config_changed` during the binding of the output object.


AlanGriffiths (approved):
Okay

PR #3441, opened 2024-06-24
(#620) Enabling keyboard LEDs when the modifier state changes

### LINKED ISSUE #620: Numlock LED does not change
I’m creating an electron kiosk app following this tutorial: https://tutorials.ubuntu.com/tutorial/electron-kiosk. 1

All works great, but I’m having a problem with the lock of the numeric pad. By default the numeric pad is disabled but I need to change this. Is there any way to change the default status?

If I press the num lock key, the numeric keyboard enables, but the LED not change. So is difficult for the user to know if it is enabled or not.

mattkae:
fixes #620 

## What's new?
- `XkbMapper` takes a callback for when the leds need to be updated for a device 
- Whenever a modifier is changed for a device, `XkbMapper` notifies the corresponding device of a possible change to its LEDs using `Device::set_leds`
- In the case if a libinput device,`LibInputDevice::set_leds` calls `libinput_device_led_update` to update the LEDs

### DIFF
--- a/src/include/server/mir/input/xkb_mapper_registrar.h
+++ b/src/include/server/mir/input/xkb_mapper_registrar.h
@@ -1,0 +1,147 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_INPUT_RECEIVER_XKB_MAPPER_H_
+#define MIR_INPUT_RECEIVER_XKB_MAPPER_H_
+
+#include "mir/input/led_observer_registrar.h"
+#include "mir/input/key_mapper.h"
+#include "mir/input/keymap.h"
+#include "mir/input/keyboard_leds.h"
+#include "mir/optional_value.h"
+#include "mir/events/xkb_modifiers.h"
+#include "mir/observer_multiplexer.h"
+
+#include <xkbcommon/xkbcommon.h>
+#include <xkbcommon/xkbcommon-compose.h>
+#include <mutex>
+#include <optional>
+#include <unordered_map>
+#include <unordered_set>
+
+namespace mir
+{
+namespace input
+{
+
+using XKBContextPtr = std::unique_ptr<xkb_context, void(*)(xkb_context*)>;
+XKBContextPtr make_unique_context();
+
+using XKBStatePtr = std::unique_ptr<xkb_state, void(*)(xkb_state*)>;
+using XKBComposeTablePtr = std::unique_ptr<xkb_compose_table, void(*)(xkb_compose_table*)>;
+using XKBComposeStatePtr = std::unique_ptr<xkb_compose_state, void(*)(xkb_compose_state*)>;
+
+namespace receiver
+{
+
+class XKBMapperRegistrar : public KeyMapper, public LedObserverRegistrar
+{
+public:
+    explicit XKBMapperRegistrar(Executor&);
+
+    void set_key_state(MirInputDeviceId id, std::vector<uint32_t> const& key_state) override;
+    void set_keymap_for_device(MirInputDeviceId id, std::shared_ptr<Keymap> map) override;
+    void set_keymap_for_all_devices(std::shared_ptr<Keymap> map) override;
+    void clear_keymap_for_device(MirInputDeviceId id) override;
+    void clear_all_keymaps() override;
+    void map_event(MirEvent& event) override;
+    MirInputEventModifiers modifiers() const override;
+    MirInputEventModifiers device_modifiers(MirInputDeviceId di) const override;
+    auto xkb_modifiers() const -> MirXkbModifiers override;
+    void register_interest(std::weak_ptr<LedObserver> const& observer, MirInputDeviceId id) override;
+    void unregister_interest(LedObserver const& observer, MirInputDeviceId id) override;
+
+protected:
+    XKBMapperRegistrar(XKBMapperRegistrar const&) = delete;
+    XKBMapperRegistrar& operator=(XKBMapperRegistrar const&) = delete;
+
+private:
+    void set_keymap(MirInputDeviceId id, std::shared_ptr<Keymap> new_keymap);
+    void set_keymap(std::shared_ptr<Keymap> new_keymap);
+    void update_modifier();
+
+    std::mutex mutable guard;
+
+    struct ComposeState
+    {
+        ComposeState(XKBComposeTablePtr const& table);
+        xkb_keysym_t update_state(xkb_keysym_t mapped_key, MirKeyboardAction action, std::string& text);
+    private:
+        XKBComposeStatePtr state;
+        std::unordered_set<xkb_keysym_t> consumed_keys;
+        mir::optional_value<std::tuple<xkb_keysym_t,xkb_keysym_t>> last_composed_key;
+    };
+
+    struct XkbMappingState
+    {
+        class XkbMappingStateLedRegistrar : public ObserverMultiplexer<LedObserver>
+        {
+        public:
+            explicit XkbMappingStateLedRegistrar(Executor&);
+            void leds_set(KeyboardLeds leds) override;
+        };
+
+        explicit XkbMappingState(
+            std::shared_ptr<Keymap> keymap,
+            std::shared_ptr<xkb_keymap> compiled_keymap,
+            Executor& executor);
+        void set_key_state(std::vector<uint32_t> const& key_state);
+
+        bool update_and_map(MirEvent& event, ComposeState* compose_state);
+        MirInputEventModifiers modifiers() const;
+        auto xkb_modifiers() const -> MirXkbModifiers;        
+        void notify_leds_changed();
+        XkbMappingStateLedRegistrar& get_registrar();
+    private:
+        /// Returns a pair containing the keysym for the given scancode and if any XKB modifiers have been changed
+        auto update_state(
+            uint32_t scan_code,
+            MirKeyboardAction direction,
+            ComposeState* compose_state,
+            std::string& text) -> std::pair<xkb_keysym_t, bool>;
+        void press_modifier(MirInputEventModifiers mod);
+        void release_modifier(MirInputEventModifiers mod);
+
+        std::shared_ptr<Keymap> const keymap;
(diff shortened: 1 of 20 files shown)


include/platform/mir/input/input_device.h
```diff
@@ -66,6 +67,8 @@ class InputDevice
 
     virtual optional_value<TouchscreenSettings> get_touchscreen_settings() const = 0;
     virtual void apply_settings(TouchscreenSettings const&) = 0;
+
+    virtual void set_leds(KeyboardLeds leds) = 0;
```
AlanGriffiths, line 71: Just noting that this is a platform ABI break

src/platforms/evdev/libinput_device.cpp
```diff
@@ -694,3 +694,16 @@ void mie::LibInputDevice::apply_settings(mi::TouchscreenSettings const& settings
 {
     touchscreen = settings;
 }
+
+void mie::LibInputDevice::set_leds(mir::input::KeyboardLeds leds)
+{
+    int led = 0;
```
AlanGriffiths, line 700: Can't we avoid the `static_cast` later with:

```suggestion
    libinput_led led = 0;
```

src/server/input/default_configuration.cpp
```diff
@@ -326,9 +327,16 @@ std::shared_ptr<mi::DefaultInputDeviceHub> mir::DefaultServerConfiguration::the_
...
        {
-           return std::make_shared<mi::receiver::XKBMapper>();
+           return std::make_shared<mi::receiver::XKBMapper>([this](auto device_id, auto leds)
+           {
+               the_default_input_device_hub()->for_each_mutable_input_device([&](mir::input::Device& device)
+               {
+                    if (device.id() == device_id)
+                        device.set_leds(leds);
+               });
+           });
```
AlanGriffiths, line 339: We should probably capture `the_default_input_device_hub()`, not call it every time the mapping state changes.

And, surely, there could be a more direct way from `XKBMapper` to the corresponding device than scanning everything registered on the device hub? (E.g. a map `MirInputDeviceId` -> callback)

AlanGriffiths (commented):
My first thought is that this should be more explicit.

* Keyboard devices register for modifier state updates
* Modifier updates are passed through that registrar

The only question then is whether the register is a new object or just a new interface supported by XKBMapper

(Maybe you can shoot this idea down though?)

src/server/input/default_configuration.cpp
mattkae: The problem is that ``the_default_input_device_hub()` also relies on the `the_key_mapper()`, so it's a circular dependency. I will see if there is some other way around this, perhaps with the "registration" idea that you said either.

src/platforms/evdev/libinput_device.cpp
mattkae: The compiler gets mad, saying:

```
Cannot initialize a variable of type 'libinput_led' with an rvalue of type 'int'
```

And then all of the `|=` don't work too:

```
Assigning to 'libinput_led' from incompatible type 'int'
```

src/common/input/xkb_mapper.cpp
```diff
@@ -14,7 +14,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "mir/input/xkb_mapper.h"
+#include "mir/input/xkb_mapper_deprecated.h"
```
AlanGriffiths, line 17: Why would we rename the file and break downstreams that use it?

src/include/common/mir/input/key_mapper_deprecated.h
AlanGriffiths, line 1: I don't see why KeyMapper is being changed. Especially when there are now two different definitions

include/platform/mir/input/input_device.h
AlanGriffiths: I think this change, and the ABI break are now unnecessary

src/include/server/mir/input/key_mapper.h
```diff
@@ -30,14 +32,21 @@ namespace input
 {
 class Keymap;
 
+class KeyMapperObserver
+{
+public:
+    virtual ~KeyMapperObserver() = default;
+    virtual void leds_set(MirInputDeviceId id, KeyboardLeds leds) = 0;
+};
+
```
AlanGriffiths, line 41: This isn't a KeyMapper observer. It is a LED state observer.

We should be keeping the KeyMapper interface unchanged. The ObserverRegister can be part of another interface implemented by XkbMapper.

Also, wouldn't it be better to register with the device `id` and only send notifications to the corresponding device. (And not notify all devices most of which drop the notification.

src/server/input/default_input_device_hub.cpp
```diff
@@ -39,6 +40,26 @@
 
 namespace mi = mir::input;
 
+namespace
+{
+class DefaultInputDeviceHubKeyMapperObserver : public mi::KeyMapperObserver
```
AlanGriffiths, line 45: I don't think the DefaultInputDeviceHub should be registering. Only devices that support LEDs

AlanGriffiths (changes requested):
After we spoke I thought you were going to propose a solution that preserved ABI. This isn't that

src/common/input/xkb_mapper.cpp
mattkae: So the tricky part is that the import will be the same when we include `mir/input/xkb_mapper.h`. So one of them has to be named differently, and I chose the old one.

Would it be a better idea to rename the new one to something slightly different? Or is there a better solution?

AlanGriffiths: So, I thought "we don't change old stuff, just deprecate it". New stuff can be called by new names, old stuff cannot

src/include/server/mir/input/key_mapper.h
mattkae: Yup I agree :+1:

src/include/server/mir/input/led_observer_registrar.h
```diff
@@ -0,0 +1,43 @@
...
+namespace mir
+{
+namespace input
+{
+class LedObserver
+{
+public:
+    virtual ~LedObserver() = default;
+    virtual void leds_set(MirInputDeviceId id, KeyboardLeds leds) = 0;
+};
```
AlanGriffiths, line 33: It is our practice to disable CopyAssign for interfaces

src/include/server/mir/input/led_observer_registrar.h
```diff
@@ -0,0 +1,43 @@
...
+
+class LedObserverRegistrar : public ObserverRegistrar<LedObserver>
+{
+public:
+    virtual ~LedObserverRegistrar() = default;
+};
+}
```
AlanGriffiths, line 40: As previously discussed I think we should be registering with a device id, which isn't supported by `ObserverRegistrar<>`

include/platform/mir/input/input_device.h
AlanGriffiths: I still think this is unnecessary here (as is the ABI break)

src/include/server/mir/input/xkb_mapper_registrar.h
```diff
@@ -0,0 +1,148 @@
...
+    void clear_keymap_for_device(MirInputDeviceId id) override;
+    void clear_all_keymaps() override;
+    void map_event(MirEvent& event) override;
+    MirInputEventModifiers modifiers() const override;
+    MirInputEventModifiers device_modifiers(MirInputDeviceId di) const override;
+    auto xkb_modifiers() const -> MirXkbModifiers override;
+    void register_interest(std::weak_ptr<LedObserver> const& observer) override;
+    void register_interest(
+        std::weak_ptr<LedObserver> const& observer,
+        Executor& executor) override;
```
AlanGriffiths, line 67: I don't see this implemented or used?

AlanGriffiths (changes requested):
This seems unfinished

mattkae:
> This seems unfinished

I have updated it :+1:

src/platforms/evdev/libinput_device.cpp
```diff
@@ -694,3 +724,21 @@ void mie::LibInputDevice::apply_settings(mi::TouchscreenSettings const& settings
...
     touchscreen = settings;
 }
+
+void mie::LibInputDevice::associate_to_id(MirInputDeviceId id)
+{
+    try_stop_observing_leds();
+    led_observer = std::make_shared<LibInputDeviceLedObserver>(this);
+    led_observer_registrar->register_interest(led_observer, id);
+    device_id = id;
+}
```
AlanGriffiths, line 734: This code looks like initialization, but is not in a constructor. Is there a good reason it is here? (And not, say, in `mi::DefaultDevice`?)

src/platforms/evdev/platform.cpp
```diff
@@ -427,9 +430,14 @@ void mie::Platform::device_added(libinput_device* dev)
...
-        devices.emplace_back(std::make_shared<mie::LibInputDevice>(report, std::move(device_ptr)));
+        auto new_device = std::make_shared<mie::LibInputDevice>(
+            report, led_observer_registrar, std::move(device_ptr));
+        devices.emplace_back(new_device);
 
-        input_device_registry->add_device(devices.back());
+        auto weak_device = input_device_registry->add_device(devices.back());
+
+        if (auto device = weak_device.lock())
+            new_device->associate_to_id(device->id());
```
AlanGriffiths, line 440: Oh!

So we're making `mi::Device` visible and passing `LedObserverRegistrar` around so that we can do two-phase initialization here?

Can we simplify things by registering the callback from `DefaultInputDeviceHub`/`DefaultDevice`?

AlanGriffiths (changes requested):
I think we could simplify further and reduce change

src/platforms/evdev/platform.cpp
mattkae: That would require `InputDevice` to have `set_leds` defined on it, which would break the input ABI again :smile: That's how I had it before.

src/platforms/evdev/libinput_device.cpp
mattkae: The reason it is here is to not break the input ABI, but perhaps I'm not seeing something ;)

src/platforms/evdev/platform.cpp
mattkae: But I guess we broke it anyway my change the `create_platform` signature :joy:

mattkae:
@AlanGriffiths For your AM tomorrow, here's a summary what I have right now:

- We have a new interface called `LedObserverRegistrar` that gets passed to the input `Platform`
- Every platform except evdev ignores that argument. The evdev platform uses it to register on Led events in `LibinputDevice`
- `XkbMapperRegistrar` will then notify `LibInputDevice` when an LED changes

In this way, only the evdev platform needs to concern itself with the LEDs, because it is the only one that cares about it.

AlanGriffiths:
@mattkae I've had a quick pass at the approach I prefer with https://github.com/canonical/mir/pull/3471 - it is still a bit rough, but you are allowed to look

AlanGriffiths (approved):
I don't have the hardware to test, but this all looks sensible

AlanGriffiths:

Looks like a rebase is in order

Issue #2599, opened 2022-08-30
Support wp_viewporter


dos1:
Several clients such as SDL or Waydroid are now using [viewporter](https://wayland.app/protocols/viewporter) protocol in order to crop and/or scale the attached buffer. It's also the recommended way for clients to deal with fractional output scaling.

Mir appears to be missing support for this protocol entirely.

Weston comes with a simple client to verify the implementation: `weston-scaler`.

RAOF:
`wp_viewporter` is a sensible thing to support - a lot of overlay hardware has capabilities that map very directly onto the interface exposed by `wp_viewporter`, so this provides optimisation opportunities that would otherwise not be available.

That you can use it to kinda implement fractional scaling is somewhat of a side effect; ideally fractional scaling would be negotiated directly, with the compositor's explicit participation, rather than worked around with a protocol that can be made to sorta get the right effect.

JamiKettunen:
Mir v2.11's missing `wp_viewporter` support is now visible on https://wayland.app/protocols/viewporter#compositor-support, when support is added (and new version tagged) https://github.com/vially/wayland-explorer should get a follow-up PR by running [`wlprobe`](https://github.com/PolyMeilex/wlprobe) against `mir_demo_server` and replacing the existing `src/data/compositors/mir.json` ^^

AlanGriffiths:
Closed by #3445

PR #3445, opened 2024-06-26
frontend/wayland: `wp_viewporter` support

RAOF:
Accidentally all squashed together into a single mega-commit, here's `wp_viewporter`.

You can test with the `weston-scaler` client in... weston, and also with https://github.com/canonical/wlcs/pull/340

### DIFF
--- a/include/platform/mir/graphics/renderable.h
+++ b/include/platform/mir/graphics/renderable.h
@@ -51,6 +51,13 @@ public:
     virtual std::shared_ptr<Buffer> buffer() const = 0;
 
     virtual geometry::Rectangle screen_position() const = 0;
+    /**
+     * The region of \ref buffer that should be sampled from.
+     *
+     * This is in buffer coordinates, so {{0, 0}, {buffer->size().width, buffer->size().height} means
+     * “use the whole buffer”.
+     */
+    virtual geometry::RectangleD src_bounds() const = 0;
     virtual std::optional<geometry::Rectangle> clip_area() const = 0;
 
     // These are from the old CompositingCriteria. There is a little bit
--- a/src/gl/tessellation_helpers.cpp
+++ b/src/gl/tessellation_helpers.cpp
@@ -14,32 +14,58 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 #include "mir/gl/tessellation_helpers.h"
+#include "mir/geometry/forward.h"
 #include "mir/graphics/renderable.h"
 #include "mir/graphics/buffer.h"
 
 namespace mg = mir::graphics;
 namespace mgl = mir::gl;
 namespace geom = mir::geometry;
+
+namespace
+{
+struct SrcTexCoords
+{
+    GLfloat top;
+    GLfloat bottom;
+    GLfloat left;
+    GLfloat right;
+};
+
+auto tex_coords_from_rect(geom::Size buffer_size, geom::RectangleD sample_rect) -> SrcTexCoords
+{
+    /* GL Texture coordinates are normalised to the size of the buffer, so (0.0, 0.0) is the top-left
+     * and (1.0, 1.0) is the bottom-right
+     */
+    SrcTexCoords coords;
+    coords.top = sample_rect.top().as_value() / buffer_size.height.as_uint32_t();
+    coords.bottom = sample_rect.bottom().as_value() / buffer_size.height.as_uint32_t();
+    coords.left = sample_rect.left().as_value() / buffer_size.width.as_uint32_t();
+    coords.right = sample_rect.right().as_value() / buffer_size.width.as_uint32_t();
+    return coords;
+}
+}
+
 mgl::Primitive mgl::tessellate_renderable_into_rectangle(
     mg::Renderable const& renderable, geom::Displacement const& offset)
 {
     auto rect = renderable.screen_position();
     rect.top_left = rect.top_left - offset;
-    GLfloat left = rect.top_left.x.as_int();
-    GLfloat right = left + rect.size.width.as_int();
-    GLfloat top = rect.top_left.y.as_int();
-    GLfloat bottom = top + rect.size.height.as_int();
+    GLfloat const left = rect.top_left.x.as_int();
+    GLfloat const right = left + rect.size.width.as_int();
+    GLfloat const top = rect.top_left.y.as_int();
+    GLfloat const bottom = top + rect.size.height.as_int();
 
     mgl::Primitive rectangle;
     rectangle.type = GL_TRIANGLE_STRIP;
 
-    GLfloat const tex_right = 1.0f;
-    GLfloat const tex_bottom = 1.0f;
+    auto const [tex_top, tex_bottom, tex_left, tex_right] =
+        tex_coords_from_rect(renderable.buffer()->size(), renderable.src_bounds());
 
     auto& vertices = rectangle.vertices;
-    vertices[0] = {{left,  top,    0.0f}, {0.0f,      0.0f}};
-    vertices[1] = {{left,  bottom, 0.0f}, {0.0f,      tex_bottom}};
-    vertices[2] = {{right, top,    0.0f}, {tex_right, 0.0f}};
+    vertices[0] = {{left,  top,    0.0f}, {tex_left, tex_top}};
+    vertices[1] = {{left,  bottom, 0.0f}, {tex_left, tex_bottom}};
+    vertices[2] = {{right, top,    0.0f}, {tex_right, tex_top}};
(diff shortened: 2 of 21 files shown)


src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -366,49 +408,107 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
...
+                if (as_int_size_if_exact(src_sample.size))
+                {
+                    return as_int_size_if_exact(src_sample.size).value() * inv_scale;
+                }
+                else
+                {
+                    throw wayland::ProtocolError{
+                        viewport.value().resource,
+                        wayland::Viewport::Error::bad_size,
+                        "No wp_viewport destination set, and source size (%f × %f) is not integral", src_sample.size.width.as_value(), src_sample.size.height.as_value()};
```
mattkae, line 498: Same here. I would expect this to be thrown in the Viewporter protocol

include/platform/mir/graphics/renderable.h
```diff
@@ -51,6 +51,13 @@ class Renderable
     virtual std::shared_ptr<Buffer> buffer() const = 0;
 
     virtual geometry::Rectangle screen_position() const = 0;
+    /**
+     * The region of \ref buffer that should be sampled from.
+     *
+     * This is in buffer coordinates, so {{0, 0}, {buffer->size().width, buffer->size().height} means
+     * “use the whole buffer”.
+     */
+    virtual geometry::RectangleD src_bounds() const = 0;
```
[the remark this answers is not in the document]
RAOF: Maybe? That pushes complexity from the producer of `src_bounds` to the consumer (which now needs to be able to synthesise the right bounds).

I'll see how it goes.

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -366,49 +408,107 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
...
+                     */
+                    geom::RectangleD source_in_buffer_coords{
+                        {raw_source.left().as_value() / inv_scale, raw_source.top().as_value() / inv_scale},
+                        raw_source.size / inv_scale
+                    };
+                    if (!entire_buffer.contains(source_in_buffer_coords))
+                    {
+                        throw wayland::ProtocolError{
+                            viewport.value().resource,
+                            wayland::Viewport::Error::out_of_buffer,
```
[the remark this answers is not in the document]
RAOF: This error actually cannot be checked before `Surface::commit()` time! We cannot know what buffer the viewport is going to be applied to before the client calls `wl_surface_commit()` - one obvious case is the client can set the viewport before there is *any* buffer, but also the client could set the viewport to something that would be invalid with the current buffer and then attach a new buffer that the viewport is valid with.

src/server/frontend_wayland/wl_surface.cpp
RAOF: Likewise! We can only check this at `commit` time because we need to know the *`scale`* before we know what the source viewport actually is, and we only know the `scale` at `commit` time.

src/server/frontend_wayland/wl_surface.cpp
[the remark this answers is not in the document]
RAOF: Hm. *Or*, these checks could be in the `Viewport` implementation, which might make `commit` simpler, too!

include/platform/mir/graphics/renderable.h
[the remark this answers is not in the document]
RAOF: I think this ends up being a bit worse than the alternative. How strongly do you feel about this? :wink:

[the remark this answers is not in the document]
AlanGriffiths: I think this is the better structure - no need for the consumer to deal with a side-channel

src/gl/tessellation_helpers.cpp
```diff
@@ -14,32 +14,58 @@
...
+auto tex_coords_from_rect(geom::Size buffer_size, geom::RectangleD sample_rect) -> SrcTexCoords
+{
+    /* GL Texture coordinates are normalised to the size of the buffer, so (0.0, 0.0) is the top-left
+     * and (1.0, 1.0) is the bottom-right
+     */
+    SrcTexCoords coords;
+    coords.top = sample_rect.top().as_value() / buffer_size.height.as_uint32_t();
+    coords.bottom = sample_rect.bottom().as_value() / buffer_size.height.as_uint32_t();
+    coords.left = sample_rect.left().as_value() / buffer_size.width.as_uint32_t();
+    coords.right = sample_rect.right().as_value() / buffer_size.width.as_uint32_t();
```
AlanGriffiths, line 44: Hmm. All this `.as_value()` ... `.as_uint32_t()` makes me wonder if there's an operator missing in our geometry. This would look safer:

```c++
coords.top = sample_rect.top() / buffer_size;
```

src/gl/tessellation_helpers.cpp
```diff
@@ -14,32 +14,58 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  */
 #include "mir/gl/tessellation_helpers.h"
+#include "mir/geometry/forward.h"
```
AlanGriffiths, line 17: It seems odd that we'd want `forward.h` in an implementation file

src/server/frontend_wayland/wp_viewporter.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+namespace mf = mir::frontend;
+namespace geom = mir::geometry;
+
+namespace
+{
+class ViewporterInstance : public mir::wayland::Viewporter
+{
+public:
+    ViewporterInstance(wl_resource* resource)
```
AlanGriffiths, line 34:
```suggestion
    explicit ViewporterInstance(wl_resource* resource)
```

src/server/frontend_wayland/wp_viewporter.h
```diff
@@ -0,0 +1,78 @@
...
+#include <optional>
+
+namespace mir::frontend
+{
+class WlSurface;
+
+class WpViewporter : public wayland::Viewporter::Global
+{
+public:
+    WpViewporter(wl_display* display);
```
AlanGriffiths, line 36:
```suggestion
    explicit WpViewporter(wl_display* display);
```

src/server/frontend_wayland/wp_viewporter.h
```diff
@@ -0,0 +1,78 @@
...
+class Viewport : public wayland::Viewport
+{
+public:
+    Viewport(wl_resource* new_viewport, WlSurface* surface);
+    ~Viewport() override;
+
+    /**
+     * Have the source or destination values changed since the last call to dirty()?
+     */
+    auto dirty() -> bool;
```
AlanGriffiths, line 56: Not sure "dirty" is the best name here. Obvious interpretations are "is dirty" and "make dirty", but this is neither. And the actual meaning won't be obvious reading the code that calls it. Even the comment isn't clear about what the first call does.

Maybe `unchanged_since_last_call()`?

Or, `unreported_changes()`?

AlanGriffiths (changes requested):
Some nits

src/gl/tessellation_helpers.cpp
RAOF: Oops, we don't, but `clangd` always wants to pull it in. I missed this one!

mattkae (approved):
This looks to be a good implementation to me! I will let Alan get it over the line

src/gl/tessellation_helpers.cpp
RAOF: I've ended up with `sample_rect.top() / buffer_size.width` (as `sample_rect.top() / buffer_size` is ambiguous).

src/server/frontend_wayland/wp_viewporter.h
RAOF: I've gone for `changed_since_last_resolve()`, and updated the implementation slightly to make that true.

AlanGriffiths:
"This branch has conflicts that must be resolved"

src/include/gl/mir/gl/tessellation_helpers.h
```diff
@@ -18,6 +18,7 @@
 #define MIR_GL_TESSELLATION_HELPERS_H_
 #include "mir/gl/primitive.h"
 #include "mir/geometry/displacement.h"
+#include "mir/geometry/rectangle.h"
```
AlanGriffiths, line 21: Looks spurious

AlanGriffiths (approved):
OK now

Fuseteam:
This closes #2599 i think

PR #3448, opened 2024-06-27
debian: new broken binutils upload

Saviq:


AlanGriffiths (approved):
Do we need to be so precise?

Saviq:
> Do we need to be so precise?

If we're not, we'll miss it if/when it gets fixed…

Saviq:
> If we're not, we'll miss it if/when it gets fixed…

Or maybe more importantly - _doesn't_ get fixed, when a new binutils is uploaded.

Issue #3449, opened 2024-06-27
gbm-kms: Abrupt cursor transition between outputs.


ehopperdietzel:
Hi, with the gbm-kms backend, I've noticed that when moving the cursor from one output to another, the transition looks a bit abrupt. This is because the hw cursor visibility for each output is toggled based on the cursor position. Instead, I believe employing the cursor 'rectangle' would be more appropriate so that it remains visible on all outputs it intersects.

It seems this can be addressed by replacing this line [src/platforms/gbm-kms/server/kms/cursor.cpp:355](https://github.com/canonical/mir/blob/0980da33355a85e45ace121e21bce9ccbad8a64a/src/platforms/gbm-kms/server/kms/cursor.cpp#L355) with:

```cpp
if (output_rect.overlaps(geometry::Rectangle(position - hotspot, size)))
```

It works for outputs using a scale factor of 1 and a normal transform, but I haven't tested other scenarios. Is there an easy way to configure outputs without modifying the example compositors code?

Also, I assume that the cursor position, hotspot, size, and the output_rect are in logical coordinates. Please correct me if I'm wrong.

AlanGriffiths:
> Is there an easy way to configure outputs without modifying the example compositors code?

Yes (depending on the compositor you are using)...

With Miriway I have the following configuration option: `display-config=static=/home/alan/.config/miriway-shell.display`

Which means I can update the display configuration by editing `.config/miriway-shell.display`. (I have also saved "docked" and "undocked" configurations I just copy over the top when moving the laptop)

ehopperdietzel:
Okay, thanks. I tested with different scale factors, and it seems to only affect surfaces, the default cursor size doesn't change (at least with the gbm-kms backend). When looking at [display_configuration_option.cpp](https://github.com/canonical/mir/blob/main/src/miral/display_configuration_option.cpp), I couldn't find an option to modify the transform.

I also tested different cursors with various hotspots, and in all cases, the transitions appear smooth. Should I create a pull request?

Saviq:
> I also tested different cursors with various hotspots, and in all cases, the transitions appear smooth. Should I create a pull request?

Yes, that's the best way to get feedback :)

AlanGriffiths:
> Hi, with the gbm-kms backend, I've noticed that when moving the cursor from one output to another, the transition looks a bit abrupt. This is because the hw cursor visibility for each output is toggled based on the cursor position.

[...]

PR #3451, opened 2024-06-27
gbm-kms: Make cursor transitions across outputs smoother

ehopperdietzel:
Keep hardware cursor planes visible on all outputs intersected by the cursor geometry instead of only its position.

Issue: [#3449](https://github.com/canonical/mir/issues/3449)


ehopperdietzel:
> presumably because the behaviour is now that the cursor is on both outputs, because it spans both outputs

Exactly, the test outputs are too small hahaha, but now it should be fixed.

PR #3453, opened 2024-06-28
Add command line option to control the behavior of `xdg_decoration_unstable_v1`

tarek-y-ismail:
Depends on #3425.

Stuff that needs to be improved:
1. The usage with `run_with` is quite un-ergonomic
2. ??


tarek-y-ismail:
One solution I tried to fix the ugliness in `run_with` is to add a method to `XdgDecorationInterface` that allows it to be converted to a `std::function<void(Server&)>`. It makes the usage something like:
```cpp
return runner.run_with(  
    {
        ...,
        *XdgDecorations::always_ssd(),
        ...
    }
);
```

AlanGriffiths:
> Depends on #3425.

You should change the target branch to match. That way, the diff will only include the changes on this branch (and the target will change automatically when the other PR lands)

tarek-y-ismail:
I actually didn't know that, thank you for telling me!

tarek-y-ismail:
Another, more elegant solution I came up with while playing with the code was returning the function object from `always_ssd` and its siblings, this way, you can plop it among other options in `run_with` and it fits right in.

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -128,17 +141,33 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
...
+        mode = Mode::server_side;
+        break;
+    case ServerMode::always_csd:
+        mode = Mode::client_side;
+        break;
+    case ServerMode::prefer_ssd:
+    case ServerMode::prefer_csd:
+        mode = new_mode;
+        break;
+    }
```
AlanGriffiths, line 167: There is no need for a switch anywhere. Your `XdgDecorationInterface` is doing the wrong job. What is needed is a Strategy interface (which can be implemented by `Decorations::Self`:

```c++
enum class DecorationsType { ssd, csd };

class DecorationStrategy
{
public:
    virtual auto default_style() const -> DecorationsType = 0;
    virtual auto request_style(DecorationsType requested) const -> DecorationsType = 0;

   ...
};
```

There can be a subclass of Decoration::Self in each of the Decoration named constructors that hard codes the correct responses

include/core/mir/xdg_decorations_interface.h
AlanGriffiths, line 1: We don't need this interface in this form, and not in mircore (a `DecorationStrategy` is needed in mirserver. (See discussion elsewhere)

AlanGriffiths (changes requested):
I'm sorry, there was much of the brief that was not clear to you and I failed to pick up on that.

1. We explicitly don't want a _command line option_ to control the decoration strategy. There only needs to be a way for the author of a compositor to override the default.
2. The xdg-decoration protocol is an implementation detail that should be hidden (as would "kwin-server-decoration" if we chose to implement that).

What we intended was a miral::Decoration class with named constructors for the supported strategies.

PR #3454, opened 2024-07-01
Add Wayland decorations strategy

tarek-y-ismail:
Fixes the misunderstandings of the doomed #3453.


src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -154,29 +164,63 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
...
 {
-    auto spec = shell::SurfaceSpecification{};
+    switch (type)
+    {
+    case DecorationStrategy::DecorationsType::ssd:
+        return Mode::server_side;
+    case DecorationStrategy::DecorationsType::csd:
+        return Mode::client_side;
+    default:
+        mir::log_warning("%s: Got invalid protocol decorations type, defaulting to client side", __FUNCTION__);
```
AlanGriffiths, line 184: This should be unreachable. No need for a `default` case

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -60,60 +62,68 @@ class ToplevelsWithDecorations
...
-    XdgToplevelDecorationV1(wl_resource* id, mir::frontend::XdgToplevelStable* toplevel);
+    XdgToplevelDecorationV1(
+        wl_resource* id, mir::frontend::XdgToplevelStable* toplevel, std::shared_ptr<DecorationStrategy> strategy);
 
     void set_mode(uint32_t mode) override;
     void unset_mode() override;
 
 private:
+    auto decorations_type_to_protocol_mode(DecorationStrategy::DecorationsType) -> uint32_t;
+    auto protocol_mode_to_decorations_type(uint32_t) -> DecorationStrategy::DecorationsType;
```
AlanGriffiths, line 95: These could be class methods, they don't need to be instance methods. Also, the names seem verbose
```suggestion
    static auto to_mode(DecorationStrategy::DecorationsType) -> uint32_t;
    static auto to_decorations_type(uint32_t) -> DecorationStrategy::DecorationsType;
```

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -60,60 +62,68 @@ class ToplevelsWithDecorations
...
     void update_mode(uint32_t new_mode);
 
-    static uint32_t const default_mode = Mode::client_side;
-
     mir::frontend::XdgToplevelStable* toplevel;
-    uint32_t mode;
+    std::shared_ptr<DecorationStrategy> decoration_strategy;
```
AlanGriffiths, line 99: Can be `const`

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -60,60 +62,68 @@ class ToplevelsWithDecorations
...
 };
 } // namespace frontend
 } // namespace mir
 
-auto mir::frontend::create_xdg_decoration_unstable_v1(wl_display* display)
+auto mir::frontend::create_xdg_decoration_unstable_v1(wl_display* display, std::shared_ptr<DecorationStrategy> strategy)
     -> std::shared_ptr<mir::wayland::XdgDecorationManagerV1::Global>
 {
-    return std::make_shared<XdgDecorationManagerV1::Global>(display);
+    return std::make_shared<XdgDecorationManagerV1::Global>(display, strategy);
```
AlanGriffiths, line 107: Move avoids unnecessary atomic operations
```suggestion
    return std::make_shared<XdgDecorationManagerV1::Global>(display, std::move(strategy));
```

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -60,60 +62,68 @@ class ToplevelsWithDecorations
...
+    new XdgDecorationManagerV1{new_zxdg_decoration_manager_v1, decoration_strategy};
 }
 
-mir::frontend::XdgDecorationManagerV1::XdgDecorationManagerV1(wl_resource* resource) :
+mir::frontend::XdgDecorationManagerV1::XdgDecorationManagerV1(
+    wl_resource* resource, std::shared_ptr<DecorationStrategy> strategy) :
     mir::wayland::XdgDecorationManagerV1{resource, Version<1>{}},
-    toplevels_with_decorations{}
+    toplevels_with_decorations{},
+    decoration_strategy{strategy}
```
AlanGriffiths, line 126: Move avoids unnecessary atomic operations

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -154,29 +164,63 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
 }
 
 mir::frontend::XdgToplevelDecorationV1::XdgToplevelDecorationV1(
-    wl_resource* id, mir::frontend::XdgToplevelStable* toplevel) :
+    wl_resource* id, mir::frontend::XdgToplevelStable* toplevel, std::shared_ptr<DecorationStrategy> strategy) :
     wayland::XdgToplevelDecorationV1{id, Version<1>{}},
-    toplevel{toplevel}
+    toplevel{toplevel},
+    decoration_strategy{strategy}
```
AlanGriffiths, line 170: Move avoids unnecessary atomic operations

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -154,29 +164,63 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
...
     switch (mode)
     {
     case Mode::client_side:
-        spec.server_side_decorated = false;
-        break;
+        return DecorationStrategy::DecorationsType::csd;
     case Mode::server_side:
+        return DecorationStrategy::DecorationsType::ssd;
+    default:
+        mir::log_warning("%s: Got invalid protocol mode (%d), defaulting to client side", __FUNCTION__, mode);
```
AlanGriffiths, line 199: A warning log message should explain what the client did wrong (e.g. what the `new_mode` was and, ideally, which client). I don't think logging the function name adds to that explanation.

src/server/symbols.map
AlanGriffiths, line 1: This isn't strictly correct as the 2.17 ABI is being broken. But we'll deal with that separately

src/include/server/mir/decoration_strategy.h
```diff
@@ -0,0 +1,33 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <cstdint>
+namespace mir
+{
+class DecorationStrategy
+{
+public:
+    enum class DecorationsType : uint32_t
```
AlanGriffiths, line 23: There's no reason for an _enum-base_, you can drop this (and the corresponding `#include`)

src/core/symbols.map
AlanGriffiths, line 1: There's no need to change this file

src/include/server/mir/decoration_strategy.h
```diff
@@ -0,0 +1,33 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <cstdint>
+namespace mir
+{
+class DecorationStrategy
```
AlanGriffiths, line 20: We disable copy and assign for interfaces. Vis:

```C++
    DecorationStrategy() = default;
    DecorationStrategy(DecorationStrategy const&) = delete;
    DecorationStrategy& operator=(DecorationStrategy const&) = delete;
```

AlanGriffiths (changes requested):
A lot of tweaks, but this is the right idea

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -127,7 +138,7 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
         BOOST_THROW_EXCEPTION(std::runtime_error("Invalid toplevel pointer"));
     }
 
-    auto decoration = new XdgToplevelDecorationV1{id, tl};
+    auto decoration = new XdgToplevelDecorationV1{id, tl, std::move(decoration_strategy)};
```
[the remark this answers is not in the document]
tarek-y-ismail: I had this exact same thought, but I tried confirming my doubts by running the code with multiple windows and nothing crashed or misbehaved. I think I should've asked before accidentally introducing a subtle bug into our code. 

I've got two questions:
1. Why didn't the code crash? Could it be that my crude testing method was incorrect, or perhaps something else?
2. I'm not sure I understand what  XdgDecorationManagerV1::decoration_strategy being const has to do with calling `std::move`? Is it because moving modifies the shared pointer, which shouldn't happen with a const object?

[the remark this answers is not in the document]
tarek-y-ismail: So I just ran the code under GDB. The use count increases normally and the stored pointer is not nulled. Digging a bit further, it's seems that the copy constructor is being called under the hood...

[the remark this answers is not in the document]
AlanGriffiths: Sorry, I wasn't clear. The behaviour of `std::move()` on a `const` source is surprising (to me at least): It returns a `const ...&&`. So the copy constructor is called as you observe.

But the reader of the code may well be misled into thinking that move semantics change the source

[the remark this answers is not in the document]
tarek-y-ismail: Indeed, removing the `const` causes the move constructor to be called and Mir to crash. Sorry! Should be fixed in the latest commit.

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
tarek-y-ismail: Note that `to_decorations_type` had to be reverted to be a class method since I needed to access `client` from inside it. Would it be better to add a parameter in make it into a static function?

src/miral/symbols.map
```diff
@@ -41,6 +41,12 @@ global:
     miral::CustomRenderer::?CustomRenderer*;
     miral::CustomRenderer::CustomRenderer*;
     miral::CustomRenderer::operator*;
+    miral::Decorations::Decorations*;
+    miral::Decorations::always_csd*;
+    miral::Decorations::always_ssd*;
+    miral::Decorations::operator*;
+    miral::Decorations::prefer_csd*;
+    miral::Decorations::prefer_ssd*;
```
AlanGriffiths, line 49: New symbols (after 2.17) belong in the MIRAL_5.1 stanza

src/miral/symbols.map
```diff
@@ -312,6 +318,13 @@ global:
     miral::X11Support::X11Support*;
     miral::X11Support::default_to_enabled*;
     miral::X11Support::operator*;
+    miral::XdgDecorations::?XdgDecorations*;
+    miral::XdgDecorations::XdgDecorations*;
+    miral::XdgDecorations::always_csd*;
+    miral::XdgDecorations::always_ssd*;
+    miral::XdgDecorations::operator*;
+    miral::XdgDecorations::prefer_csd*;
+    miral::XdgDecorations::prefer_ssd*;
```
AlanGriffiths, line 327: I don't think these exist

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -162,29 +173,64 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
...
-        spec.server_side_decorated = false;
-        break;
+        return DecorationStrategy::DecorationsType::csd;
     case Mode::server_side:
+        return DecorationStrategy::DecorationsType::ssd;
+    default:
+        pid_t pid;
+        wl_client_get_credentials(client->raw_client(), &pid, nullptr, nullptr); // null pointers are allowed
+
+        mir::log_warning("Client PID: %d, got invalid protocol mode (%d), defaulting to client side.", pid, mode);
```
AlanGriffiths, line 208:
```suggestion
        mir::log_warning("Client PID: %d, attempted to set invalid zxdg_toplevel_decoration_v1 mode (%d), defaulting to client side.", pid, mode);
```

include/core/mir/decoration_strategy.h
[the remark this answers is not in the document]
AlanGriffiths: Correction: there's already a DecorationStrategy there, this header doesn't need to be at all.

src/server/symbols.map
```diff
@@ -10,10 +10,14 @@ global:
...
     mir::BasicCallback::operator*;
     mir::BasicCallback::unlock*;
+    mir::DecorationStrategy::?DecorationStrategy*;
+    mir::DecorationStrategy::DecorationStrategy*;
+    mir::DecorationStrategy::operator*;
     mir::DefaultServerConfiguration::DefaultServerConfiguration*;
     mir::DefaultServerConfiguration::add_wayland_extension*;
     mir::DefaultServerConfiguration::default_reports*;
     mir::DefaultServerConfiguration::set_enabled_wayland_extensions*;
+    mir::DefaultServerConfiguration::set_the_decoration_strategy*;
```
AlanGriffiths, line 20: New symbols should be in the MIR_SERVER_INTERNAL_2.18 stanza

AlanGriffiths (changes requested):
Some tidying up needed, but looking good

AlanGriffiths:

Probably easiest to rebase on `main`

src/server/symbols.map
tarek-y-ismail: I tried moving symbol declarations to the `MIR_SERVER_INTERNAL_2.18` section while merging with main. Does this fix the ABI?

include/miral/miral/decorations.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#ifndef MIRAL_DECORATIONS_H
+#define MIRAL_DECORATIONS_H
+
+
+#include <memory>
+
+namespace mir { class Server; class DecorationStrategy; }
+
```
AlanGriffiths, line 24: DecorationStrategy is not needed by this header

src/include/server/mir/decoration_strategy.h
AlanGriffiths, line 23: There's no reason for an enum-base, you can drop this (and the corresponding #include)

src/server/symbols.map
AlanGriffiths: In theory, it doesn't fix the ABI (because we have changed the DefaultServerConfiguration vtable), but in practice we've not cared about that as an "internal" breakage.

Don't worry about it for this PR: it is pre-existing and only a concern because of recent changes.  We'll deal with that separately.

src/include/server/mir/decoration_strategy.h
tarek-y-ismail: I thought I did so in an earlier commit. I guess I missed something while resolving conflicts?

AlanGriffiths:
There's a bunch of build failures: I think you need to run `cmake --build build --target regenerate-miral-debian-symbols`

include/miral/miral/decorations.h
```diff
@@ -0,0 +1,60 @@
...
+#define MIRAL_DECORATIONS_H
+
+
+#include <memory>
+
+namespace mir { class Server; }
+
+namespace miral
+{
+class Decorations
```
AlanGriffiths, line 27:
```suggestion
/// Configures the window decoration strategy.
/// \note The strategy can only be applied to clients that are able to negotiate the decoration style with the server.
/// \remark Since MirAL 5.1
class Decorations
```

AlanGriffiths (approved):
I am suggesting a little more documentation of `miral::Decorations`, but looks good

src/server/frontend_wayland/xdg_decoration_unstable_v1.cpp
```diff
@@ -162,29 +173,64 @@ void mir::frontend::XdgDecorationManagerV1::get_toplevel_decoration(wl_resource*
...
     case Mode::server_side:
+        return DecorationStrategy::DecorationsType::ssd;
+    default:
+        pid_t pid;
+        wl_client_get_credentials(client->raw_client(), &pid, nullptr, nullptr); // null pointers are allowed
+
+        mir::log_warning("Client PID: %d, attempted to set invalid zxdg_toplevel_decoration_v1 mode (%d), defaulting to client side.", pid, mode);
+
+        return DecorationStrategy::DecorationsType::csd;
+    }
```
mattkae, line 211:
```suggestion
    default:
    {
        pid_t pid;
        wl_client_get_credentials(client->raw_client(), &pid, nullptr, nullptr); // null pointers are allowed

        mir::log_warning("Client PID: %d, attempted to set invalid zxdg_toplevel_decoration_v1 mode (%d), defaulting to client side.", pid, mode);

        return DecorationStrategy::DecorationsType::csd;
    }
    }
```

mattkae (changes requested):
One teeny suggestion about leaking in a `switch` (technically, although it hardly matters)

include/miral/miral/decorations.h
tarek-y-ismail: I don't really have any more info to document on my mind right now. Any suggestions?

mattkae (approved):
I am happy with this :smile:

Issue #3431, opened 2024-06-19
regression: crash in `mir::renderer::gl::Renderer::draw` when screenshotting on hybrid with Nvidia and `--display-config=single`


Saviq:
This regressed in:
- https://github.com/canonical/mir/pull/3402

On a hybrid Nvidia / Intel system with two displays connected (one per GPU):

```
$ mir_demo_server --add-wayland-extensions=all --display-config=single &
$ glmark2-es2-wayland &
$ grim
```

```gdb
#0  mir::renderer::gl::Renderer::draw (this=0x555555629ba0, renderable=...) at /usr/src/mir-2.17.0+dev113-g601aa2ba5c-0ubuntu24.04/src/renderers/gl/renderer.cpp:413
        texture = std::shared_ptr<mir::graphics::gl::Texture> (empty) = {get() = 0x0}
        clip_area = std::optional [no contained value]
        prog = <optimized out>
        rect = <optimized out>
        centrex = <optimized out>
        centrey = <optimized out>
        transform = {value = {{{x = 3.00976562, r = 3.00976562, s = 3.00976562}, {y = 0, g = 0, t = 0}, {z = 2.2958874e-41, b = 2.2958874e-41, p = 2.2958874e-41}, {w = 0, a = 0, q = 0}}, {{x = 9.96890384e+26, r = 9.96890384e+26, s = 9.96890384e+26}, {y = 4.59163468e-41, g = 4.59163468e-41, t = 4.59163468e-41}, {z = -2.53689518e+28, b = -2.53689518e+28, p = -2.53689518e+28}, {w = 4.59163468e-41, a = 4.59163468e-41, q = 4.59163468e-41}}, {{x = 1.55663708e+13, r = 1.55663708e+13, s = 1.55663708e+13}, {y = 3.0611365e-41, g = 3.0611365e-41, t = 3.0611365e-41}, {z = -2.53694217e+28, b = -2.53694217e+28, p = -2.53694217e+28}, {w = 4.59163468e-41, a = 4.59163468e-41, q = 4.59163468e-41}}, {{x = 2.2958874e-41, r = 2.2958874e-41, s = 2.2958874e-41}, {y = 0, g = 0, t = 0}, {z = -3.47259381e+23, b = -3.47259381e+23, p = -3.47259381e+23}, {w = 4.59163468e-41, a = 4.59163468e-41, q = 4.59163468e-41}}}}
[...]

PR #3456, opened 2024-07-02
bugfix: BasicScreenShooter now selects a GLRenderingProvider based off of its compatibility with the BufferAllocator

mattkae:
fixes: #3431


RAOF (approved):
I cannot seem to reproduce the bug that this is should fix, so I can't test that it *does* fix it, but this *looks* right to me.

Issue #3462, opened 2024-07-04
Nautilus open-with dialog crashes with Mir internal error (Missing `DRM_FORMAT_ABGR16161616F`)


RAOF:
When opening the “open-with” dialog in Nautilus, it crashes with a protocol error. In the Mir log, we have:
```
[2024-07-04 16:51:47.171822] < -warning- > frontend:Wayland: Exception processing LinuxBufferParamsV1::create_immed() request: /home/chris/Canonical/Mir/mir/main/src/platform/graphics/drm_formats.cpp(456): Throw in function const mg::DRMFormat::FormatInfo &(anonymous namespace)::info_for_format(uint32_t)
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: Unsupported DRM format: DRM_FORMAT_ABGR16161616F
```

Looks like we're missing an interesting DRM format descriptor.

PR #3463, opened 2024-07-05
platform/DRMFormat: Don't require a `FormatInfo` for each format.

RAOF:
We're now using `DRMFormat` in a situation (linux-dmabuf) where:
* We don't actually need to care about many the properties of the format, and
* We're processing formats submitted by cilents.

This means that the existing “throw an exception if we haven't got a `FormatInfo`” approach works badly; we will disconnect clients with a Wayland internal error if they submit buffers in a format we haven't explicitly listed, even though it *would* work if we just treated the format as an opaque identifier.

So, don't do that anymore. Instead, store the fourcc code directly in `DRMFormat`, only look up the `FormatInfo` (now `DRMFormat::Info`) when code asks for it.

Additionally, we log the first time code asks for the `DRMFormat::Info` for a format we haven't got info for; this way we can fill out the needed formats on an as-noticed basis.

Fixes: #3462

### DIFF
--- a/src/platform/graphics/drm_formats.cpp
+++ b/src/platform/graphics/drm_formats.cpp
@@ -493,7 +493,7 @@ auto mg::DRMFormat::info() const -> std::optional<Info const>
         mir::log_warning("Detailed info for format %s missing; please report so this can be added", name());
         unknown_formats->insert(fourcc);
     }
-    return {};
+    return std::nullopt;
 }
 
 mg::DRMFormat::operator uint32_t() const
--- a/src/platform/graphics/drm_formats.cpp
+++ b/src/platform/graphics/drm_formats.cpp
@@ -490,7 +490,8 @@ auto mg::DRMFormat::info() const -> std::optional<Info const>
     auto unknown_formats = unknown_formats_guard.lock();
     if (!unknown_formats->contains(fourcc))
     {
-        mir::log_warning("Detailed info for format %s missing; please report so this can be added", name());
+        mir::log_warning(
+            "Detailed info for format %s missing; please report this to https://github.com/canonical/mir/issues/new so this can be added", name());
         unknown_formats->insert(fourcc);
     }
     return std::nullopt;
--- a/src/platform/graphics/drm_formats.cpp
+++ b/src/platform/graphics/drm_formats.cpp
@@ -69,7 +69,7 @@ struct mg::DRMFormat::FormatInfo
     bool has_alpha;
     std::optional<uint32_t> opaque_equivalent;
     std::optional<uint32_t> alpha_equivalent;
-    std::optional<mg::DRMFormat::RGBComponentInfo> components;
+    std::optional<Info::RGBComponentInfo> components;
 };
 
 namespace
@@ -80,7 +80,7 @@ constexpr std::array const formats = {
         false,
         DRM_FORMAT_XRGB4444,
         DRM_FORMAT_ARGB4444,
-        mg::DRMFormat::RGBComponentInfo{
+        mg::DRMFormat::Info::RGBComponentInfo{
             4, 4, 4, {}
         },
     },
@@ -89,7 +89,7 @@ constexpr std::array const formats = {
         false,
         DRM_FORMAT_XBGR4444,
         DRM_FORMAT_ABGR4444,
-        mg::DRMFormat::RGBComponentInfo{
+        mg::DRMFormat::Info::RGBComponentInfo{
             4, 4, 4, {}
         },
     },
@@ -98,7 +98,7 @@ constexpr std::array const formats = {
         false,
         DRM_FORMAT_RGBX4444,
         DRM_FORMAT_RGBA4444,
-        mg::DRMFormat::RGBComponentInfo{
+        mg::DRMFormat::Info::RGBComponentInfo{
             4, 4, 4, {}
         },
     },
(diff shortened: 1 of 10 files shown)


RAOF:
I'm not sure if I like this; particularly because we now don't have any signal for “it would be useful to have information on this format, but we don't have it”.

It might be better to separate `DRMFormat` and `DRMFormat::FormatDetails`, and have the latter be an `optional<>` that may contain the current details. Then, only code that cares about the details can try to access them, and we can error (or, at least, log a warning/error) if those details are unavailable.

RAOF:
> It might be better to separate `DRMFormat` and `DRMFormat::FormatDetails`, and have the latter be an `optional<>` that may contain the current details. Then, only code that cares about the details can try to access them, and we can error (or, at least, log a warning/error) if those details are unavailable.

I have now Done This.

src/platform/symbols.map
```diff
@@ -10,14 +10,15 @@ MIR_PLATFORM_2.17 {
...
+    mir::graphics::DRMFormat::Info::alpha_equivalent*;
     mir::graphics::DRMFormat::as_mir_format*;
-    mir::graphics::DRMFormat::components*;
+    mir::graphics::DRMFormat::Info::components*;
     mir::graphics::DRMFormat::from_mir_format*;
-    mir::graphics::DRMFormat::has_alpha*;
+    mir::graphics::DRMFormat::Info::has_alpha*;
     mir::graphics::DRMFormat::name*;
-    mir::graphics::DRMFormat::opaque_equivalent*;
+    mir::graphics::DRMFormat::Info::opaque_equivalent*;
```
mattkae, line 19: Is this a breaking change for the symbols since we're deleting some symbols? Would we need to bump the major version in this case?

mattkae (changes requested):
A minor thing about symbols, but this makes sense to me overall

src/platform/symbols.map
RAOF: Yes, thanks for the catch; I have now done so!

src/platform/symbols.map
```diff
@@ -1,4 +1,4 @@
-MIR_PLATFORM_2.17 {
+MIR_PLATFORM_2.18 {
```
mattkae, line 1: Would this be a major version bump since we're breaking ABI :thinking:

mattkae (approved):
A single question that I might be wrong on, but looks good to me :smile:

src/platform/symbols.map
AlanGriffiths: No

src/platform/graphics/drm_formats.cpp
```diff
@@ -442,68 +444,60 @@ constexpr auto maybe_info_for_format(uint32_t fourcc_format) -> mg::DRMFormat::F
...
-    return info->components;
+    if (auto info = maybe_info_for_format(fourcc))
+    {
+        return Info(info);
+    }
+    /* TODO: This could fire quite frequently.
+     * Ideally we would have a rate-limited (or once-per-format) logger, but for now, just
+     * throw it in the debug stream.
+     */
+    mir::log_debug("Detailed info for format %s missing; please report so this can be added", name());
```
AlanGriffiths, line 489: There's a lot of TODOs being added. In `linux_dmabuf.cpp` we say something similar, but in that context we don't log. I think we need a better approach to reporting this. BTW "rate-limiting" doesn't sound right either.

With this approach, I think "debug" logging requesting reporting is addressing the wrong audience. "debug" and "trace" are for those familiar with the code. "Warning" might be better.

src/platform/graphics/drm_formats.cpp
```diff
@@ -442,68 +444,60 @@ constexpr auto maybe_info_for_format(uint32_t fourcc_format) -> mg::DRMFormat::F
...
+    return {};
```
AlanGriffiths, line 490: `std::nullopt`?

AlanGriffiths (changes requested):
Looks mostly sensible. But we can do better with reporting missing format info

RAOF:
There we go! I have JFDI and resolved the TODOs rather than adding them.

src/platform/graphics/drm_formats.cpp
```diff
@@ -482,12 +484,16 @@ auto mg::DRMFormat::info() const -> std::optional<Info const>
...
-    mir::log_debug("Detailed info for format %s missing; please report so this can be added", name());
-    return {};
+
+    // Actually, we probably want `std::flat_set`, but no libstdc++ support yet.
+    static Synchronised<std::unordered_set<decltype(fourcc)>> unknown_formats_guard;
+    auto unknown_formats = unknown_formats_guard.lock();
+    if (!unknown_formats->contains(fourcc))
+    {
+        mir::log_warning("Detailed info for format %s missing; please report so this can be added", name());
+        unknown_formats->insert(fourcc);
```
AlanGriffiths, line 494: I know this is a micro optimisation but...
```suggestion
    if (unknown_formats->insert(fourcc).second)
    {
        mir::log_warning("Detailed info for format %s missing; please report so this can be added", name());
```

Also, maybe some guidance on where to report?

AlanGriffiths (approved):
Some nitpicking. Feel free to land anyway

src/platform/graphics/drm_formats.cpp
RAOF: I've put a link to GitHub issues in the message text. I *haven't* taken the micro-optimisation, because I think it makes the intent slightly less obvious and it is the micro-est of micro-optimisations.

PR #3471, opened 2024-07-10
Simpler "plumbing"

AlanGriffiths:


src/server/input/default_input_device_hub.cpp
```diff
@@ -492,6 +496,11 @@ auto mi::DefaultInputDeviceHub::add_device(std::shared_ptr<InputDevice> const& d
         seat->add_device(*handle);
         dev->start(seat, input_dispatchable);
 
+        if (auto const observer = std::dynamic_pointer_cast<LedObserver>(device))
+        {
+            led_observer_registrar->register_interest(observer, handle->id());
+        }
```
mattkae, line 502: We will need to `unregister_interest` in `remove_device`, right?

mattkae (changes requested):
Makes sense with a lot less plumbing required! I think we just need to unregister_interest, which I will do. Testing this now

mattkae (approved):
Looks good to me!

src/server/input/default_input_device_hub.cpp
AlanGriffiths: > We will need to `unregister_interest` in `remove_device`, right?

Oh, I was misled by "register" taking a weak_ptr. That made me think that EOL was an implicit deregister

PR #3474, opened 2024-07-11
(#3461) bugfix: screencopy works at scale for fullscreen screenshots

### LINKED ISSUE #3461: Screenshots no longer respect output scale
When screenshotting with grim to see if I could reproduce #3431 to see if I could *test* #3456, I instead discovered that
* I could indeed screenshot on my AMD/NVIDIA system, but
* Those screenshots only took up 1/4 of the captured output (presumably the screenshot buffer's size-in-pixels is based on the scaled size, and the screenshot didn't take that into account)

For example

(If you have a look there there's a perfectly normal screenshot in the bottom quarter, and three fully-transparent rectangles around it.)

mattkae:
fixes #3461


AlanGriffiths (approved):
Once you find the right line of code... obvious

PR #3478, opened 2024-07-12
feature: provide a way to specify an additional data source for configuration data

mattkae:
## Why?
In miracle (and probably other compositors), people may want to specify configuration options in a different format (e.g. a json file, a yaml file, etc.). To facilitate this, users should be able to specify a simple map that contains the key-value pairs that were parsed from another source.

## What's new?
- Added a new class to miral called `ConfigurationDataSource`
- Added `DefaultConfiguration::set_options_map` (maybe this should be in the constructor, but the constructors were a bit large)
- Added `DefaultConfiguration::parse_custom`
- Added `ProgramOption::parse_map`
- Fixed the symbols map generator

## To test
In your main runner function, do something like:

```c++
return runner.run_with(
        {
            ...
            ConfigurationDataSource([]()
            {
                return std::map<std::string, std::string>{
                    {"startup-apps", "gedit"}
                };
            }),
            ...
         });
```

### DIFF
--- a/include/miral/miral/configuration_data_source.h
+++ b/include/miral/miral/configuration_data_source.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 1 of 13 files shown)


AlanGriffiths:
> In miracle (and probably other compositors), people may want to specify configuration options in a different format (e.g. a json file, a yaml file, etc.). To facilitate this, users should be able to specify a simple map that contains the key-value pairs that were parsed from another source.

I'm not convinced this is useful in this form.

This adds complexity to the current "configuration fixed at startup" options. As you know, I'm working to replace a lot of what  "configuration fixed at startup" is currently used for with a way to dynamic configuration changes  whilst running. I suspect a lot of what you want to source from "a json file, a yaml file, etc." is going to fall into the "dynamic configuration" category and is better built on that work.

mattkae:
> I'm not convinced this is useful in this form.
> 
> This adds complexity to the current "configuration fixed at startup" options. As you know, I'm working to replace a lot of what "configuration fixed at startup" is currently used for with a way to dynamic configuration changes whilst running. I suspect a lot of what you want to source from "a json file, a yaml file, etc." is going to fall into the "dynamic configuration" category and is better built on that work.

So long as I can set a Mir-defined configuration option from some external source (e.g. somehow set "--display-config" from a JSON file), I'd be happy. If that work will enable it, then I will close this one in favor :+1:

PR #3479, opened 2024-07-13
Check for required C++ features during initial build (std::format, etc)

mikekasprzak:
Per the discussion in #3472, this is an alternative approach of informing users that their C++ compiler and standard library doesn't support certain required features. It was born out of my challenges getting mainline _Mir_ built and running on *Pop!_OS* (_Ubuntu 22.04 LTS_) instead of _Ubuntu 24.04 LTS_.

Admittedly it's a "big ask" of any project to track used language features, but C++ is an unusual case, for how closely your compilers are tied to your Linux distro.

To make managing the language feature list serve a dual purpose, and in lieu of a coding style guide, I proposing treating these additions to the `CMakeLists.txt` file as a master list of **recommended** C++ features for all contributors. That doesn't mean adding every _C++20_ feature to the list, just the ones you want contributors to use.

Under the hood, I wrote some macros that wrap `CMAKE_CXX_COMPILE_FEATURES` and `check_cxx_symbol_exists` in a way that should be both forward and backward compatible with different versions of _cmake_. It relies on C++20 feature testing, using the exact names from the specification. You can find a handy list of C++ feature macro names here:

<https://en.cppreference.com/w/cpp/utility/feature_test>

NOTE: _cmake_ refers to C++ as `cxx`, but C++ feature testing refers to it as `cpp`.

Usage looks something like this.

## CMakeLists.txt
```cmake
...

set(CMAKE_CXX_STANDARD 20)
set(CMAKE_CXX_EXTENSIONS OFF)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

include (cmake/RequireCXXFeature.cmake)
# Reference: https://en.cppreference.com/w/cpp/utility/feature_test

# Required C++ features
require_cxx_feature(__cpp_constexpr)

# Required C++ standard library features
require_cxx_feature(__cpp_lib_format)

...
```

If your tools meets the requirements, nothing out of the ordinary will occur. If you don't meet them, you will get an output similar to this:

```console
mike@DevL14:~/Work/mir$ cmake -B build
-- The C compiler identification is GNU 11.4.0
-- The CXX compiler identification is GNU 11.4.0
...
-- Looking for __cpp_constexpr
-- Looking for __cpp_constexpr - found
-- Looking for __cpp_lib_format
-- Looking for __cpp_lib_format - not found
CMake Error at cmake/RequireCXXFeature.cmake:28 (message):
  Required C++ feature __cpp_lib_format not found!

  See the feature test documentation, or
  https://en.cppreference.com/w/cpp/feature_test for details.

Call Stack (most recent call first):
  CMakeLists.txt:73 (require_cxx_feature)


-- Configuring incomplete, errors occurred!
mike@DevL14:~/Work/mir$ 
```

This is a proof-of-concept, but it should be fully functional for the use case described. The next step would be to dig into the source code and populate the list of recommended C++ features.

## Polyfilling features
For completeness, I included cmake macros for both `require_cxx_feature` and `check_cxx_feature`. In practice `require_cxx_feature` is probably all that's needed, but in case you wanted to polyfill a C++ feature or standard library, `check_cxx_feature` can be used instead.

```cmake
...
include (cmake/CheckCXXFeature.cmake)
include (cmake/RequireCXXFeature.cmake)
# Reference: https://en.cppreference.com/w/cpp/utility/feature_test

# Required C++ features
require_cxx_feature(__cpp_constexpr)

# Required C++ standard library features
check_cxx_feature(__cpp_lib_format HAS_LIB_FORMAT)
if (NOT DEFINED "${HAS_LIB_FORMAT}")
  # polyfill config goes here
endif()
...
```

## Nitty gritty
`require_cxx_feature` and `check_cxx_feature` are a simpler interface to the standard cmake module `CheckCXXSymbolExists`. The compilers define the language feature symbols (i.e. `__cpp_FEATURENAME`), and C++20 requires that the standard library provide a header called `version` defining all library features (i.e. `__cpp_lib_FEATURENAME`). `GCC 11` for example supports many C++20 features, but the standard library shipped alongside it was incomplete. The `version` header exists, but per the examples above, the `format` header and symbol `__cpp_lib_format` do not.

[`CheckCXXSymbolExists`](https://cmake.org/cmake/help/v3.8/module/CheckCXXSymbolExists.html) stores the variable given in the final argument in the _cmake_ cache. Once cached, the check will not be done again until the cache is purged. `require_cxx_feature` purges that cache entry on failure, so the test will be run again when _cmake_ is run again. Once it succeeds, like `CheckCXXSmybolExists` it will not be checked again until the _cmake_ cache is purged.


RAOF (commented):
I'm fairly happy with this approach.

I don't think we'll keep the required feature list up to date (and this is missing a bunch of features we use, like `operator <=>` and `std::binary_semaphore`), but when people run into problems and pop up in the bugtracker we can reactively encode that knowledge.

AlanGriffiths:
Closed for because of lack of activity for ~8 months

PR #3485, opened 2024-07-15
ci: replace codecov with TICS

Saviq:


mattkae (changes requested):
This looks like it maybe timed out in CI?

mattkae:
This brings up a good question: Do we want an action that takes over an hour on every pull request?

Saviq:
> This brings up a good question: Do we want an action that takes over an hour on every pull request?

We can probably limit what gets calculated:

https://ticsdocumentation.tiobe.com/latest/docs/index.html#doc=user/clientoptions.html%23MetricAliases

But let's start without pull requests, then, so we get the reference runs going on `main`.

PR #3486, opened 2024-07-15
Add dependencies to mir-x11-kiosk

AlanGriffiths:
These are needed to make it work in a snap environment, so make life easy by adding them here


mattkae (approved):
Cool by me :+1:

PR #3487, opened 2024-07-16
Placate some of TIOBE's MISRA warnings

AlanGriffiths:


mattkae (approved):
Reasonable enough change to me!

src/common/event.cpp
```diff
@@ -80,10 +78,15 @@ std::string mir::event_type_to_string(MirEventType t)
     case mir_event_type_window_output:
         return "mir_event_type_window_output";
     default:
-        abort();
+        std::unreachable();
```
RAOF, line 81: Should we do our typical
```
#ifndef CLANG
default:
    <things>
#endif
```
dance here to make hitting this `std::unreachable()` a compile error on clang?

src/common/input/input_event.cpp
```diff
@@ -43,8 +45,8 @@ std::string input_event_type_to_string(MirInputEventType input_event_type)
             return "mir_input_event_type_touch";
         case mir_input_event_type_pointer:
             return "mir_input_event_type_pointer";
-        default:
-            abort();
+    default:
+            std::unreachable();
```
RAOF, line 49: Also here. Make it a compile error on clang?

Also, inconsistent indentation (although what you've added is more correct according to the style guide)

src/core/depth_layer.cpp
```diff
@@ -16,24 +16,21 @@
...
+    case mir_depth_layer_background:
+    case mir_depth_layer_below:
+    case mir_depth_layer_application:
+    case mir_depth_layer_always_on_top:
+    case mir_depth_layer_above:
+    case mir_depth_layer_overlay:
+        return depth_layer;
+
+    default:
+        std::unreachable();
```
RAOF, line 34: Oh. Do you know if clang will fail the build if this `std::unreachable()` is actually reachable? (ie: if we add an extra enumeration value?)

RAOF (commented):
I'd *like* to keep the “build fails if we miss adding an enumeration” property that we currently have. Do you know if clang treats potentially hitting `std::unreachable()` the same way as potentially not hitting a `case`?

src/core/geometry/rectangles.cpp
```diff
@@ -17,6 +17,7 @@
 #include "mir/geometry/rectangles.h"
 #include "mir/geometry/displacement.h"
 #include <algorithm>
+#include <compare>
```
mattkae, line 20: Spurious or necessary?

mattkae (approved):
One question, but this is good to me other than Chris' comments

src/common/event.cpp
AlanGriffiths: MISRA doesn't like that...
```
Rule:	COV_MISRA_CPP_2008_Rule_16_2_1
Level:	4
Category:	Preprocessor Directives
Synopsis:	The pre-processor shall only be used for file inclusion and include guards.
Ruleset:	C++ Coding Standard
Tool:	Coverity 2023.12.0
```

src/core/geometry/rectangles.cpp
AlanGriffiths: I'm hoping this fixes:
```
[ERROR 5073] The following error(s) occurred while 'Coverity' was running: 'RW.BAD_ORDERING_TYPE: (CID:NA-7663bcf08764a1c9a897744e961cdd7f) no valid std::partial_ordering type found ( must be included)'.
```

RAOF:
Ok. I think the `std::unreachable()` is a mistake; we shouldn't be deliberately injecting UB if we fail to update our switch statement when adding a new enum variant.

Factoring that pattern out into a `MIR_ASSERT_EXHAUSTIVE` macro should reduce the MISRA warning down to the single file containing that definition, which we could then ignore because that's a silly warning.

RAOF (changes requested):
Ok. I think the `std::unreachable()` is a mistake; we shouldn't be deliberately injecting UB if we fail to update our switch statement when adding a new enum variant. That is making our code *less* robust to satisfy the static analyser.

Factoring that pattern out into a `MIR_ASSERT_EXHAUSTIVE` macro should reduce the MISRA warning down to the single file containing that definition, which we could then ignore because that's a silly warning.

AlanGriffiths:
I'll split out the unconscientious changes in my AM and rework the PR

RAOF (approved):
Yeah, this looks good to me now.

PR #3491, opened 2024-07-17
Xwayland complains that `-listen` is deprecated, use `-listenfd` instead

AlanGriffiths:


src/server/frontend_xwayland/xwayland_server.cpp
```diff
@@ -83,7 +83,7 @@ void exec_xwayland(
     for (auto const& fd : spawner.socket_fds())
     {
         mf::XWaylandSpawner::set_cloexec(fd, false);
-        args.push_back("-listen");
+        args.push_back("-listenfd");
         // strdup() may look like a leak, but execvp() will trash all resource management
         args.push_back(strdup(std::to_string(fd).c_str()));
```
RAOF, line 88: Oh, no! Pre-existing, but:

> After a fork() in a multithreaded program, the child can safely call only async-signal-safe functions (see signal-safety(7)) until such time as it calls execve(2).

RAOF (approved):
Apart from the pre-existing nit, yup.

AlanGriffiths:
Merging: You've raised the "nit" as #3494, and it is more work than I feel justified for this PR

PR #3495, opened 2024-07-18
debian: drop custom LTO handling

Saviq:


RAOF (approved):
Yup, reasonable

RAOF:
We probably want to follow up to make sure that the mir-libs snap (and everything else) is doing the right thing.

Saviq:
> We probably want to follow up to make sure that the mir-libs snap (and everything else) is doing the right thing.

I dropped the only other mentions of `MIR_LINK_TIME_OPTIMIZATION`, so we should be fine. Nothing in `mir-libs`, or everything else that should be affected.

RAOF:
> > We probably want to follow up to make sure that the mir-libs snap (and everything else) is doing the right thing.
> 
> I dropped the only other mentions of `MIR_LINK_TIME_OPTIMIZATION`, so we should be fine. Nothing in `mir-libs`, or everything else that should be affected.

Oh, I meant to make sure that these places retain the benefit of LTO.

Saviq:
> > > We probably want to follow up to make sure that the mir-libs snap (and everything else) is doing the right thing.
> 
> Oh, I meant to make sure that these places retain the benefit of LTO.

*cough*. They `mir-libs` does not have LTO enabled... but it being used in CI only, I wonder if it's appropriate anyway? I think we want to optimize more for build time than runtime performance here.

What "everything else" did you have in mind?

RAOF:
> What "everything else" did you have in mind?

I… didn't, really. I guess: ensure the PPA builds have LTO? Is there anywhere else that we pull artifacts from, that *should* have LTO enabled?

Saviq:
> ensure the PPA builds have LTO?

Wasn't the purpose to let the distro admins to decide? :shrug:
But yes, Ubuntu has it enabled by default - Debian, not so.

> Is there anywhere else that we pull artifacts from, that _should_ have LTO enabled?

Not that I know of.

This is good to go from my PoV.

RAOF:
Good to go by me then, too.

PR #3497, opened 2024-07-18
tests: replace `tmpnam` use

### LINKED ISSUE #2675: warning: the use of `tmpnam' is dangerous, better use `mkstemp'
```
/usr/bin/ld: tests/miral/CMakeFiles/miral-test.dir/external_client.cpp.o: in function `(anonymous namespace)::ExternalClient::ExternalClient()':                                                                                              
/home/michal/dev/MirServer/mir/tests/miral/external_client.cpp:32: warning: the use of `tmpnam' is dangerous, better use `mkstemp'                                                                                                            
/usr/bin/ld: tests/miral/CMakeFiles/miral-test-internal.dir/static_display_config.cpp.o: in function `StaticDisplayConfig_nonexistent_config_file_is_no_error_Test::TestBody()':                                                              
/home/michal/dev/MirServer/mir/tests/miral/static_display_config.cpp:151: warning: the use of `tmpnam' is dangerous, better use `mkstemp' 
```

Saviq:
Fixes #2675

### DIFF
--- a/tests/mir_test/file_utils.cpp
+++ b/tests/mir_test/file_utils.cpp
@@ -1,0 +1,44 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <system_error>
+#include <unistd.h>
+
+#include "mir/test/file_utils.h"
+
+std::string mir::test::create_temp_file()
+{
+    char temp_file[] = "/tmp/temp_file_XXXXXX";
+    if (int fd = mkstemp(temp_file) == -1)
+    {
+        throw std::system_error(errno, std::system_category(), "Failed to create temp file");
+    }
+    else
+    {
+        close(fd);
+    }
+
+    return temp_file;
+}
+
+std::string mir::test::create_temp_dir()
+{
+    char temp_dir[] = "/tmp/temp_dir_XXXXXX";
+    if (mkdtemp(temp_dir) == nullptr)
+        throw std::system_error(errno, std::system_category(), "Failed to create temp dir");
+
+    return temp_dir;
+}
--- a/tests/miral/external_client.cpp
+++ b/tests/miral/external_client.cpp
@@ -18,6 +18,8 @@
 #include <miral/external_client.h>
 #include <miral/x11_support.h>
 
+#include "mir/test/file_utils.h"
+
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
 #include <fstream>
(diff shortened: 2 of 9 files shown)


tests/mir_test/file_utils.cpp
```diff
@@ -0,0 +1,44 @@
...
+    {
+        throw std::system_error(errno, std::system_category(), "Failed to create temp file");
+    }
+    else
+    {
+        close(fd);
+    }
+
+    return temp_file;
+}
```
RAOF, line 35: So, this suffers basically all the problems that `tmpnam` has.

*But* it's only used in test code where the problems of `tmpnam` aren't really problems, so :woman_shrugging:

*But* it would be reasonably easy to replace the current uses of `tmpname` with `mkstemp` directly. So even more :woman_shrugging:

tests/miral/external_client.cpp
```diff
@@ -43,7 +45,7 @@ struct ExternalClient : miral::TestServer
     miral::X11Support x11_disabled_by_default{};
     miral::X11Support x11_enabled_by_default{miral::X11Support{}.default_to_enabled()};
 
-    std::string const output = tmpnam(nullptr);
+    std::string const output = mir::test::create_temp_file();
```
RAOF, line 48: This could be a `mir::Fd` from `mkstmp`; the `external_client.launch` lines would then redirect to the fd using `>&$FD`.

tests/miral/static_display_config.cpp
```diff
@@ -151,7 +152,9 @@ struct StaticDisplayConfig : Test
 
 TEST_F(StaticDisplayConfig, nonexistent_config_file_is_no_error)
 {
-    miral::StaticDisplayConfig{tmpnam(nullptr)};
+    auto filename = mir::test::create_temp_file();
+    unlink(filename.c_str());
```
RAOF, line 156: This could be `create_temp_dir` plus an arbitrary filename.

tests/unit-tests/frontend_wayland/test_desktop_file_manager.cpp
```diff
@@ -194,7 +195,7 @@ TEST_F(DesktopFileManager, can_resolve_from_valid_flatpak_info)
     std::stringstream flatpak_info_path;
     flatpak_info_path << "/proc/" << PID << "/root/.flatpak-info";
     auto const flatpak_info = flatpak_info_path.str();
-    auto tmp_file_name = std::tmpnam(NULL);
+    auto tmp_file_name = mir::test::create_temp_file().c_str();
```
RAOF, line 198: This could be `mkstmp`; nothing really uses the filename here.

tests/unit-tests/frontend_wayland/test_desktop_file_manager.cpp
```diff
@@ -228,7 +229,7 @@ TEST_F(DesktopFileManager, app_id_will_not_resolve_from_flatpak_info_when_name_i
...
```
RAOF, line 232: As above

RAOF (commented):
I don't know? `create_temp_file()` has basically all the same reasons to not use as `tmpnam` except that it won't trigger the linker warning, but might be an attractive nuisance?

On the other hand, I don't really think that it matters? :woman_shrugging:

tests/mir_test/file_utils.cpp
Saviq: > So, this suffers basically all the problems that `tmpnam` has.

Yeah this is just about silencing the linker warning. But you're right, we can just use `mkstemp` directly - thanks for the reality check.

AlanGriffiths: > > So, this suffers basically all the problems that `tmpnam` has.
> 
> Yeah this is just about silencing the linker warning. But you're right, we can just use `mkstemp` directly - thanks for the reality check.

I concur

PR #3498, opened 2024-07-18
bugfix: fixing how previous versions are parsed for the symbols map generator

mattkae:


tools/symbols_map_generator/main.py
```diff
@@ -41,8 +41,19 @@ class LibraryInfo(TypedDict):
     header_directories: list[HeaderDirectory]
     map_file: str
 
+def get_version_from_library_version_str(version: str) -> str:
+    """
+    Given a string like MIR_SERVER_INTERNAL_5.0.0, returns the version
+    string (e.g. 5.0.0)
+    """
+    s = version.split("_")
+    return s[len(s) - 1]
```
Saviq, line 50: This should do?

```suggestion
    return s[-1]
```

tools/symbols_map_generator/main.py
```diff
@@ -41,8 +41,19 @@ class LibraryInfo(TypedDict):
     header_directories: list[HeaderDirectory]
     map_file: str
 
+def get_version_from_library_version_str(version: str) -> str:
+    """
+    Given a string like MIR_SERVER_INTERNAL_5.0.0, returns the version
+    string (e.g. 5.0.0)
+    """
+    s = version.split("_")
+    return s[-1]
```
Saviq, line 50:
```suggestion
    return version.split("_")[-1]
```

Saviq (approved):
Even simpler :)

PR #3500, opened 2024-07-22
platforms/common: Move DRMEventHandler in from eglstream-kms.

RAOF:
This should be useful when transitioning to the Atomic KMS API in gbm-kms.


RAOF:
I'm not sure if anything else will want to be migrated to common code; I'll see as implementation continues. This will, though.

AlanGriffiths (approved):
Sure

PR #3504, opened 2024-07-23
ci: add documentation checks

Saviq:


tarek-y-ismail:
Looks like this has some issues with the Doxyfile inclusion thing we discussed before. It assumes that `doc/sphinx` will have a Doxyfile but in reality it's at `<build dir>/doc/sphinx`.

Saviq:
It's worse, there isn't a `Doxyfile` at all, just a `Doxyfile.in` - I need to run `cmake` on it before.

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,16 @@
+name: Main Documentation Checks
+
+on:
+  - push
+  - pull_request
+  - workflow_dispatch
```
Saviq, line 6: As discussed, I think we should just go for manual runs for now.
```suggestion
  - workflow_dispatch
```

doc/sphinx/Makefile
```diff
@@ -66,13 +70,16 @@ pa11y-install:
 		}
 
 install: $(VENVDIR)
+	sudo apt install cmake doxygen libboost-dev libboost-iostreams-dev libboost-filesystem-dev libboost-program-options-dev libboost-system-dev libdrm-dev libegl-dev libepoxy-dev libfreetype-dev libglib2.0-dev libgles2-mesa-dev libglm-dev libinput-dev liblttng-ust-dev libwayland-dev libxcb-composite0-dev libxcb-randr0-dev libxcursor-dev libxkbcommon-dev libxml++2.6-dev libyaml-cpp-dev xsltproc -y
```
Saviq, line 73: Let's split that up into package-per-line so diffs are nicer later.

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,16 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true  
+
+jobs:
+  documentation-checks:
+    uses: canonical/documentation-workflows/.github/workflows/documentation-checks.yaml@main
+    with:
+      working-directory: doc/sphinx
```
Saviq, line 16: Couldn't this be `build_doc/doc/sphinx`, and then you wouldn't need to change the individual check targets?

Saviq (commented):
For all these:
```
cd ../build_doc/doc/sphinx/
```
Couldn't we actually change the `working-directory`?

.github/workflows/automatic-doc-checks.yml
tarek-y-ismail: `build_doc/doc/sphinx` doesn't exist before the job is  run. For that to happen, we'd need to run a configure job first, upload `build_docs`,  then download it again _as a step of the documentation-checks job_. If we're going to edit it either way, might as well add a step to configure in `documentation-checks` itself.

Saviq: Right, because the install step's _in_ that `Makefile`… yeah, we need a step in between. Anyway, fine for now.

doc/sphinx/.sphinx/spellingcheck.yaml
```diff
@@ -9,7 +9,7 @@ matrix:
     - .custom_wordlist.txt
     output: .sphinx/.wordlist.dic
   sources:
-  - _build/**/*.html
+  - _build/**/*.html|!_build/**/api/**/*.html
```
Saviq, line 12: The api is directly under `_build`, no?

```suggestion
  - _build/**/*.html|!_build/api/**/*.html
```

.github/workflows/automatic-doc-checks.yml
Saviq, line 6: I think just manual for now?

```suggestion
  - workflow_dispatch
```

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,16 @@
+name: Main Documentation Checks
+
+on:
+  - push
+  - pull_request
+  - workflow_dispatch
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
Saviq, line 9: May as well go for the version ready for pull requests:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.number && format('pr{0}', github.event.number) || github.run_id }}
```

doc/sphinx/.sphinx/pa11y.json
```diff
@@ -1,9 +1,11 @@
...
-  "chromeLaunchConfig": {
-    "args": [
-      "--no-sandbox"
-    ]
+  "defaults": {
+    "chromeLaunchConfig": {
+      "args": [
+	"--no-sandbox"
+      ]
+    }
```
Saviq, line 7: This something that should be contributed [upstream](https://github.com/canonical/sphinx-docs-starter-pack/blob/main/.sphinx/pa11y.json)?

doc/sphinx/Makefile
```diff
@@ -9,8 +9,19 @@ SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
 VENVDIR       = $(SPHINXDIR)/venv
-PA11Y         = $(SPHINXDIR)/node_modules/pa11y/bin/pa11y.js --config $(SPHINXDIR)/pa11y.json
+PA11Y         = $(SPHINXDIR)/node_modules/pa11y-ci/bin/pa11y-ci.js --config $(SPHINXDIR)/pa11y.json
```
Saviq, line 12: Does `-ci` give us significant boost now? Should we propose to the starter pack first?

doc/sphinx/Makefile
```diff
@@ -60,22 +75,27 @@ woke-install:
...
-run: install
-	. $(VENV); sphinx-autobuild -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS)
+run: install configure
+	. $(VENV); cd $(WORKSPHINX); sphinx-autobuild -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS)
 
 # Doesn't depend on $(BUILDDIR) to rebuild properly at every run.
-html: install
-	. $(VENV); $(SPHINXBUILD) -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" -w $(SPHINXDIR)/warnings.txt $(SPHINXOPTS)
+html: install configure
+	# Don't rebuild html if already built
```
Saviq, line 91: Hmm… shouldn't this be a Make dependency, then?

Similar to `configure`, if it was a proper target that depended on another, we wouldn't configure again?

Saviq (commented):
Like it how the custom wordlist shows us what can be easily improved.

Still uneasy about all the `cd ${WORKSPHINX}` bits… but I suppose we'd need to modify the upstream workflow to improve that (e.g. have a separate `${CONFIGDIR}` that the `configure`, or, say `${PREPAREDIR}`/`prepare` target would generate and then the individual checks targets would depend on that. If it happened to be equal (the default), it would be NOOP.

doc/sphinx/.sphinx/pa11y.json
tarek-y-ismail: This is specifically for `pa11y-ci`, so it depends on whether or not we decide to upstream `pa11y-ci` (see discussion below)

doc/sphinx/Makefile
tarek-y-ismail: I haven't yet compared performance, but the readability improvements are worth it alone

Saviq: Sure, I just mean diverging from the starter pack is going to make it harder to refresh it, so if we think this an improvement, we should propose it upstream sooner than later.

doc/sphinx/Makefile
tarek-y-ismail: Check the latest commit, does it match what you had in mind?

doc/sphinx/Makefile
tarek-y-ismail: Yeah, I'll check real quick if there's a reason the docs team didn't use it (license or similar), if not, I'll open a PR

doc/sphinx/Makefile
Saviq: Yeah, and then `configure` can be `$(WORKSPHINX)/CMakeCache.txt`

tarek-y-ismail: It won't offer much speed improvements, just a few less lines of output

Saviq: Yeah, it's just The Way™ in Makefiles ;)

OTOH, unless you enumerate the sources/dependencies, I suppose that might just bit you in the foot since it won't update :)

doc/sphinx/.sphinx/spellingcheck.yaml
```diff
@@ -27,3 +27,6 @@ matrix:
       - div.visually-hidden
       - img
       - a.p-navigation__link
+      - li:matches(.toctree-l3, .toctree-l2) # ignore table of content tags
+      - div:matches(.mermaid)                # ignore mermaid.js diagrams
```
Saviq, line 31: Not sure we should be ignoring those in the long run. They're actually meant to be somewhat human-readable even in text form.

But it's probably fine to start with.

There's some terms that got added to the wordlist, then, that can be removed.

tarek-y-ismail: The problem is that these tags are included in every html file, so they fill the output with duplicate lines.

doc/sphinx/Makefile
```diff
@@ -60,22 +75,27 @@ woke-install:
...
+$(WORKSPHINX)/_build/index.html:
+	cd $(WORKSPHINX); \
+	. $(VENV); \
+	$(SPHINXBUILD) -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" -w $(SPHINXDIR)/warnings.txt $(SPHINXOPTS)
 
-# Doesn't depend on $(BUILDDIR) to rebuild properly at every run.
-html: install
-	. $(VENV); $(SPHINXBUILD) -b dirhtml "$(SOURCEDIR)" "$(BUILDDIR)" -w $(SPHINXDIR)/warnings.txt $(SPHINXOPTS)
+# Make sure that HTML docs are built at least once
+html: install configure $(WORKSPHINX)/_build/index.html
```
Saviq, line 95: Reading the comment above, it looks to be rebuilt on purpose (otherwise a manual `make html` will not work on subsequent runs). While I get it that this costs us in CI, I'm not sure it's worth the potential confusion when it doesn't work?

Saviq (commented):
One last thing inline.

doc/sphinx/Makefile
tarek-y-ismail: HTML docs need to only be built once per run, no need to rebuild them for different targets. So why would we need to rebuild them each time?

Saviq:
Looks like we have just a few `broken` links to get this in a happy place? Most are "missing anchor", which should be ingored, but a couple seem legit issues.

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,16 @@
+name: Main Documentation Checks
+
+on:
+  - push
+  - workflow_dispatch
```
Saviq, line 5: I think we're close enough that we should just bite the bullet :)

```suggestion
  - pull_request
```

.github/workflows/automatic-doc-checks.yml
Saviq: This got lost somewhere?

doc/sphinx/Makefile
```diff
@@ -9,11 +9,33 @@ SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
 BUILDDIR      = _build
 VENVDIR       = $(SPHINXDIR)/venv
-PA11Y         = $(SPHINXDIR)/node_modules/pa11y/bin/pa11y.js --config $(SPHINXDIR)/pa11y.json
+PA11YCI       = $(SPHINXDIR)/node_modules/pa11y-ci/bin/pa11y-ci.js --config $(SPHINXDIR)/pa11y-ci.json
 VENV          = $(VENVDIR)/bin/activate
-
-.PHONY: help full-help woke-install pa11y-install install run html epub serve \
-        clean clean-doc spelling linkcheck woke pa11y Makefile
```
Saviq, line 16: This I'd leave as-is, that it's using `pa11y-ci` is an implementation detail.

doc/sphinx/Makefile
```diff
@@ -27,7 +49,7 @@ help:
         "* check links: make linkcheck \n" \
         "* check spelling: make spelling \n" \
         "* check inclusive language: make woke \n" \
-        "* check accessibility: make pa11y \n" \
+        "* check accessibility: make pa11y-ci \n" \
```
Saviq, line 52: And this, no need to leak the `-ci` detail.

doc/sphinx/Makefile
```diff
@@ -9,11 +9,33 @@ SPHINXBUILD   ?= sphinx-build
...
+# don't build multiple times (multiple invocations = multiple complete builds
+# even if files didn't change)
+ifeq ($(GITHUB_WORKFLOW), "")
+HTMLINDEX     = ""
+else
+HTMLINDEX     = $(WORKSPHINX)/_build/index.html
+endif
+
+.PHONY: help full-help woke-install pa11y-ci-install install run html epub serve \
+        clean clean-doc spelling linkcheck woke pa11y-ci Makefile
```
Saviq, line 38:
```suggestion
```

Saviq (commented):
Just some nitpicks.

tarek-y-ismail:
> Looks like we have just a few broken links to get this in a happy place? Most are "missing anchor", which should be ingored, but a couple seem legit issues.

I think fixing the issues that the checks exposed should be in a separate PR as this is bloated enough :)

Saviq:
> I think fixing the issues that the checks exposed should be in a separate PR as this is bloated enough :)

Fine :)

doc/sphinx/Makefile
tarek-y-ismail: Why remove the phony target?

Saviq: Because if you don't add `-ci` to `pa11y` above, it's there already

tarek-y-ismail: Should be handled now

.github/workflows/automatic-doc-checks.yml
```diff
@@ -0,0 +1,15 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number && format('pr{0}', github.event.number) || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  documentation-checks:
+    uses: canonical/documentation-workflows/.github/workflows/documentation-checks.yaml@main
+    with:
+      working-directory: doc/sphinx
+      pa11y-target: pa11y-ci
```
Saviq, line 15:
```suggestion
```

Saviq (commented):
Last leftover and let's get it in :)

tarek-y-ismail (approved):
image

PR #3507, opened 2024-07-24
Make lambda_as_function work for more cases

AlanGriffiths:


tarek-y-ismail (approved):
Can't say I can understand the template magic (an explanation would be greatly appreciated, if possible). But I can confirm that it fixes the compilation problem

AlanGriffiths:
> an explanation would be greatly appreciated, if possible

The clang error is:
```
/usr/include/miral/miral/lambda_as_function.h:36:6: note: candidate template ignored: substitution failure [with Lambda = (lambda at /home/alan/CLionProjects/mir-example/main.cpp:16:29) &]: type '(lambda at /home/alan/CLionProjects/mir-example/main.cpp:16:29) &' cannot be used prior to '::' because it has no members
   36 | auto lambda_as_function(Lambda&& lambda) -> typename detail::FunctionType<decltype(&Lambda::operator())>::type
```

Quite correctly, it says that the type `...&` has no members: We need to drop the `&`. That's what `std::remove_reference` is for. (And `std::remove_reference`  is a no-op for non-reference types, so the change won't break existing code)

PR #3509, opened 2024-07-25
Add support for `wp_fractional_scale_v1`

tarek-y-ismail:


src/server/frontend_wayland/fractional_scale_v1.cpp
```diff
@@ -0,0 +1,105 @@
...
+    auto fractional_scale = new FractionalScaleV1{id};
+    fractional_scale->send_preferred_scale_event(120 * scale);
+
+    auto surf = WlSurface::from(surface);
+    auto on_destroy = [this, surface]()
+    {
+        surfaces_with_fractional_scale->unregister_resource_owner(surface);
+    };
+    surf->add_destroy_listener(on_destroy);
+    fractional_scale->add_destroy_listener(on_destroy);
```
RAOF, line 99: Hm. I suspect this is the wrong approach. I think it's likely that a full implementation of this will want to store the `FractionalScaleV1` in the surface - we'll need to send out a new `prefered_scale` whenever the surface moves to an output with different scale.

(The [`wp_viewporter` merge](https://github.com/canonical/mir/pull/3445/files) might give some pointers as to how that would go, although because the fractional-scale object isn't double-buffered state you wouldn't need the pending/current split)

RAOF (changes requested):

src/server/frontend_wayland/fractional_scale_v1.cpp
tarek-y-ismail: Should be good now! Thanks for all the help today :)

src/server/frontend_wayland/window_wl_surface_role.cpp
```diff
@@ -528,6 +528,9 @@ void mf::WindowWlSurfaceRole::handle_enter_output(graphics::DisplayConfiguration
                     client,
                     [&](OutputInstance* instance)
                     {
+                        if(auto fractional_scale = surface.value().get_fractional_scale())
+                            fractional_scale->output_entered(instance);
+
```
RAOF, line 533: I think this is the wrong spot for the `entered` call? We don't want to call `output_entered` once per global bound by the client, we just want to call it once, with the output that the surface has entered.

That would be up at line 526 - within the `if (config.id == id) { ... }` conditional, not within the `for_each_output_bound_by` functor.

This would probably mean changing the signature of `output_{entered,left}` to take a `DisplayConfigurationOutput const&`, and the `FractionalScaleV1` object would then want to have a set of `{id, scale}` instead of `OutputInstance*`.

src/server/frontend_wayland/unique_resource_tracker.h
```diff
@@ -0,0 +1,47 @@
+/*
```
RAOF, line 1: You don't actually need the `UniqueResourceTracker` changes at all now, do you?

RAOF (changes requested):

src/server/frontend_wayland/window_wl_surface_role.cpp
tarek-y-ismail: You are right. I incorrectly understood your explanation and thought that `handle_enter_output` was called once and `for_each_output_bound_by` would handle looping over the outputs that interest us (In retrospect, I have no idea why I thought of this). I just tried your suggestion and the code still works. Thanks!

src/server/frontend_wayland/unique_resource_tracker.h
tarek-y-ismail: I thought of leaving it separated just in case. But  I agree that it would be easier to review without these changes (and they're easy to replicate). Should be fixed in the latest commit.

src/server/frontend_wayland/wl_surface.h
```diff
@@ -183,6 +187,7 @@ class WlSurface : public wayland::Surface
     std::optional<std::vector<mir::geometry::Rectangle>> input_shape;
     std::vector<SceneSurfaceCreatedCallback> scene_surface_created_callbacks;
     wayland::Weak<Viewport> viewport;
+    FractionalScaleV1* fractional_scale{nullptr};
```
RAOF, line 190: Hm. This should be a `wayland::Weak<FractionalScaleV1>` - the lifetime of Wayland objects is outside of our control, and the client might destroy the `fractional_scale` without destroying the surface, which would leave this a dangling pointer.

RAOF (changes requested):
This needs changes to handle the Wayland protocol object lifetimes. Particularly, we should be storing a `wayland::Weak<FractionalScaleV1>`, and checking that it's not expired before use.

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -525,6 +528,17 @@ auto mf::NullWlSurfaceRole::scene_surface() const -> std::optional<std::shared_p
 {
     return std::nullopt;
 }
+
+void mf::WlSurface::set_fractional_scale(mir::frontend::FractionalScaleV1* fractional_scale)
+{
+    this->fractional_scale = wayland::Weak(fractional_scale);
```
RAOF, line 534: We would typically spell this
```suggestion
    this->fractional_scale = wayland::Weak{fractional_scale};
```

RAOF (approved):
Yup! Looks good now.

PR #3515, opened 2024-07-29
sbuild: don't override gcc

Saviq:
This was ill-advised.


mattkae (dismissed):
Cool :+1:

Saviq:
> Cool 👍

Yeah, it wasn't ;)

RAOF (approved):
Yeah, this seems like a thing we probably don't want to be doing. We want to make sure that we're buildable with the distro default toolchain!

PR #3516, opened 2024-07-30
Drop overengineered, leaky and broken error reporting on VT switch

AlanGriffiths:
This was broken as, despite passing an exception around, the exception is never thrown and so `current_exception()` is null and we get a segfault.

It was leaky as we called `new std::function<void(std::exception const&)>{error_handler};` and never delete.

It was over-engineered as there is no need to pass an exception around to log a runtime failure.


mattkae (approved):
Makes sense to me!

Issue #3499, opened 2024-07-18
Crash when changing display configurations


AlanGriffiths:
I recently switched to Miriway/edge and since that have had several compositor crashes when switching between "docked" and "undocked" display configurations.

The only info I have from the logs (so far):

```
2024-07-18T14:40:07.876575+01:00 Octopull-Framework16 /usr/libexec/gdm-wayland-session[12790]: Segmentation fault (core dumped)
```

AlanGriffiths:
Maybe this is also related to #3517 - the dock is a USB device. Have to check that...

AlanGriffiths:
...seems to be.

PR #3517, opened 2024-07-30
Unregister callback before item is nullptr

AlanGriffiths:
The previous ordering led to a segfault using `sudo chvt 1` - as `*pos` was null 

Also:
Fixes: #3518
Fixes: #3499


RAOF (approved):
Ah, right. Yes. `remove_if` leaves `[pos, end)` in valid-but-unspecified state :woman_facepalming:

PR #3521, opened 2024-07-31
Drop in a mutex to lock around data access

AlanGriffiths:
I can't convince myself that all access to `mir_test_framework::WindowManagementTestHarness::Self` data members is thread safe. So dropping in a mutex to see if it fixes the weird failures we see (mostly on riscv64).

### DIFF
--- a/tests/mir_test_framework/window_management_test_harness.cpp
+++ b/tests/mir_test_framework/window_management_test_harness.cpp
@@ -109,6 +109,7 @@ void mir_test_framework::WindowManagementTestHarness::SetUp()
 
     auto fake_display = std::make_unique<mtd::FakeDisplay>(get_output_rectangles());
     {
+        std::lock_guard lock{self->mutex};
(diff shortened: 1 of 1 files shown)


tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -106,7 +108,9 @@ void mir_test_framework::WindowManagementTestHarness::SetUp()
 
 
     auto fake_display = std::make_unique<mtd::FakeDisplay>(get_output_rectangles());
-    self->display = fake_display.get();
+    {
+        self->display = fake_display.get();
```
RAOF, line 112: Did you mean to lock a mutex here?

RAOF (commented):
Looks not unreasonable, but one of the changes looks like maybe you've forgotten to add a lock?

mattkae (approved):
Good catch. I suppose it should be harmless at least

Issue #885, opened 2019-06-20
If multiple server-mesa-x11 platforms are present, all add an "x11-output" option


AlanGriffiths:
For example, if both `server-mesa-x11.so.15` and `server-mesa-x11.so.16` are present then it is impossible to use:

```bash
miral-shell --x11-output 800x600
```

as "option --x11-output is ambiguous"

RAOF:
This needs a rework of the configuration system in order to be solved. Urgh.

RAOF:
This got resolved as a part of handling the fact that the `atomic-kms` and `gbm-kms` platforms share option names.

PR #3525, opened 2024-08-05
Initial Atomic KMS platform

RAOF:
This is a minimal-changes copy of the display half of the `gbm-kms` platform, switched over to all Atomic KMS API.

It's parallel installable with the existing `gbm-kms` platform, which will still be preferentially used on devices with atomic support.

Currently, over the existing `gbm-kms` display platform it only has improved gamma support.
Immediate missing features:
* Hardware cursor support
* Querying list of supported pixel formats


src/platforms/atomic-kms/server/kms/quirks.cpp
```diff
@@ -0,0 +1,204 @@
...
+{
+    return impl->should_skip(device);
+}
+
+void mga::Quirks::add_quirks_option(boost::program_options::options_description& config)
+{
+    config.add_options()
+        (quirks_option_name,
+         boost::program_options::value<std::vector<std::string>>(),
+         "[platform-specific] Driver quirks to apply (may be specified multiple times; multiple quirks are combined)");
```
AlanGriffiths, line 185: This same option is added by `gbm-kms` - and that spells trouble if both platforms are available. (Similar to #885)

Saviq:
Hmm with dual outputs I get:

    < - ERROR - > atomic-kms: Failed to set CRTC: No space left on device (28)

And an occasional SIGABRT after.

Yes. I. Have. Disk. Space.

Saviq:
@tarek-y-ismail's ploughed through drm debug logs and found:

```
[drm:drm_atomic_plane_check] [PLANE:32:plane 1A] invalid source coordinates 2256.000000x1504.000000+0.000000+0.000000 (fb 1920x1200)
[drm:drm_atomic_check_only] [PLANE:32:plane 1A] atomic core check failed
```

This suggests there's a problem with cloning, as it tries to map the smaller buffer onto a bigger display.

tarek-y-ismail:
> Hmm with dual outputs I get:

By "dual outputs", do you mean buitlin-monitor + external monitor, or two external monitors?
I don't get this error, but I also don't get external output (no parameters passed except `--platform-display-libs=mir:atomic-kms --platform-rendering-libs=mir:gbm-kms`)

Saviq:
> By "dual outputs", do you mean builtin-monitor + external monitor, or two external monitors?

Bulting + external.

> I don't get this error, but I also don't get external output.

As discussed, to get output through your Nvidia GPU, you'll need `,mir:eglstream-kms` added to your display plugins.

tarek-y-ismail:
Just tried it, both monitors work fine. I assume this is because the second output is not using atomic-kms and thus not triggering the bug

Saviq:
> Just tried it, both monitors work fine. I assume this is because the second output is not using atomic-kms and thus not triggering the bug

Yup, that could well be it - they're not sharing the platform like in my case.

src/platforms/atomic-kms/server/kms/quirks.cpp
tarek-y-ismail: Is this resolved by #3590?

Saviq:
Trying heterogeneous refresh rates I got:

```
[2024-10-09 18:19:28.744101] <information> mirserver: Initial display configuration:
[2024-10-09 18:19:28.744134] <information> mirserver: * Output 1: eDP connected, used
[2024-10-09 18:19:28.744156] <information> mirserver: . |_ EDID manufacturer: BOE
[2024-10-09 18:19:28.744172] <information> mirserver: . |_ EDID product code: 2399
[2024-10-09 18:19:28.744190] <information> mirserver: . |_ Physical size 13.3" 280x190mm
[2024-10-09 18:19:28.744203] <information> mirserver: . |_ Power is on
[2024-10-09 18:19:28.744216] <information> mirserver: . |_ Current mode 2256x1504 59.99Hz
[2024-10-09 18:19:28.744228] <information> mirserver: . |_ Preferred mode 2256x1504 59.99Hz
[2024-10-09 18:19:28.744240] <information> mirserver: . |_ Orientation normal
[2024-10-09 18:19:28.744255] <information> mirserver: . |_ Logical size 2256x1504
[2024-10-09 18:19:28.744268] <information> mirserver: . |_ Logical position +0+0
[2024-10-09 18:19:28.744280] <information> mirserver: . |_ Scaling factor: 1.00
[2024-10-09 18:19:28.744292] <information> mirserver: * Output 2: DisplayPort connected, used
[2024-10-09 18:19:28.744305] <information> mirserver: . |_ EDID monitor name: DELL U2413
[2024-10-09 18:19:28.744326] <information> mirserver: . |_ EDID manufacturer: DEL
[2024-10-09 18:19:28.744338] <information> mirserver: . |_ EDID product code: 61512
[2024-10-09 18:19:28.744351] <information> mirserver: . |_ Physical size 24.0" 520x320mm
[2024-10-09 18:19:28.744364] <information> mirserver: . |_ Power is on
[2024-10-09 18:19:28.744377] <information> mirserver: . |_ Current mode 1920x1080 23.97Hz
[2024-10-09 18:19:28.744403] <information> mirserver: . |_ Preferred mode 1920x1200 59.95Hz
[2024-10-09 18:19:28.744416] <information> mirserver: . |_ Orientation normal
[2024-10-09 18:19:28.744433] <information> mirserver: . |_ Logical size 1920x1080
[2024-10-09 18:19:28.744445] <information> mirserver: . |_ Logical position +2256+0
[2024-10-09 18:19:28.744457] <information> mirserver: . |_ Scaling factor: 1.00
[2024-10-09 18:19:28.744474] <information> mirserver: * Output 3: DisplayPort disconnected
[2024-10-09 18:19:28.744487] <information> mirserver: * Output 4: DisplayPort disconnected
[2024-10-09 18:19:28.744503] <information> mirserver: * Output 5: DisplayPort disconnected
[2024-10-09 18:19:28.744747] < - debug - > miral: Configuring pointer: 'basic-window-manager'
[2024-10-09 18:19:28.744814] <information> input-hub: Device configuration: basic-window-manager, capabilities={pointer}
[2024-10-09 18:19:28.761780] < - ERROR - > atomic-kms: Failed to set CRTC: Invalid argument (22)
Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGSEGV
```
[drm.log](https://github.com/user-attachments/files/17311521/drm.log)

RAOF:
Oh! It's not heterogeneous *refresh* rates, it's heterogeneous *resolutions*!

Saviq:
So we're clear on the issue I'm seeing still:

https://github.com/user-attachments/assets/42f0560b-0a25-4acf-ad60-e5c59b1eed30

RAOF:
AHA! It took writing a script to try and reproduce this, but I now can.

What's happening here is that the final frame submitted by the terminal is racing the modeset sequence.

When it stops updating, what's happening is the terminal is waiting on a frame event from its last-buffer-before-modeset before submitting any new frames. Mir is not sending a frame event until *something else* triggers a composition.

So this is a problem somewhere in our buffer queuing, I think.

src/platforms/atomic-kms/server/display_helpers.h
```diff
@@ -0,0 +1,89 @@
...
+#include "mir/fd.h"
+
+#include <cstddef>
+#include <memory>
+#include <vector>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic warning "-Wall"
+#include <gbm.h>
+#pragma GCC diagnostic pop
```
AlanGriffiths, line 30: I'm curious: most places we include `<gbm.h>` we don't disable warnings. Why is this occasion different?

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -0,0 +1,214 @@
...
+using LockedFrontBuffer = std::unique_ptr<gbm_bo, std::function<void(gbm_bo*)>>;
+
+class GBMBoFramebuffer : public mg::FBHandle
+{
+public:
+    static auto framebuffer_for_frontbuffer(mir::Fd const& drm_fd, LockedFrontBuffer bo) -> std::unique_ptr<GBMBoFramebuffer>
+    {
+        if (auto cached_fb = static_cast<std::shared_ptr<uint32_t const>*>(gbm_bo_get_user_data(bo.get())))
+        {
+            return std::unique_ptr<GBMBoFramebuffer>{new GBMBoFramebuffer{std::move(bo), *cached_fb}};
```
AlanGriffiths, line 58: AFAICS, the only reason for using raw `new` is that the constructor is private. But, as the class is local to this translation unit, and the member data is deep const, that seems a bad reason.

```suggestion
            return std::make_unique<GBMBoFramebuffer>(std::move(bo), *cached_fb);
```

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -0,0 +1,214 @@
...
+        auto fb_id = new std::shared_ptr<uint32_t>{
+            new uint32_t{0},
+            [drm_fd](uint32_t* fb_id)
+            {
+                if (*fb_id)
+                {
+                    drmModeRmFB(drm_fd, *fb_id);
+                }
+                delete fb_id;
+            }};
```
AlanGriffiths, line 70: Do we _really_ want to allocate a `shared_ptr<>` on the stack? And ultimately leak it (preventing the deleter ever being used)?

```suggestion
        auto fb_id = std::shared_ptr<uint32_t>{
            new uint32_t{0},
            [drm_fd](uint32_t* fb_id)
            {
                if (*fb_id)
                {
                    drmModeRmFB(drm_fd, *fb_id);
                }
                delete fb_id;
            }};
```

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+        /* Create a KMS FB object with the gbm_bo attached to it. */
+        auto ret = drmModeAddFB2(drm_fd, width, height, format,
+                                 handles, strides, offsets, fb_id->get(), 0);
+        if (ret)
+            return nullptr;
+
+        gbm_bo_set_user_data(bo.get(), fb_id, [](gbm_bo*, void* fb_ptr) { delete static_cast<std::shared_ptr<uint32_t const>*>(fb_ptr); });
+
+        return std::unique_ptr<GBMBoFramebuffer>{new GBMBoFramebuffer{std::move(bo), *fb_id}};
```
AlanGriffiths, line 88:
```suggestion
        return std::make_unique<GBMBoFramebuffer>(std::move(bo), std::move(fb_id));
```

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -0,0 +1,214 @@
...
+auto create_gbm_surface(gbm_device* gbm, geom::Size size, mg::DRMFormat format, std::span<uint64_t> modifiers)
+    -> std::shared_ptr<gbm_surface>
+{
+    auto const surface =
+        [&]()
+        {
+            if (modifiers.empty())
+            {
+                // If we have no no modifiers don't use the with-modifiers creation path.
+                auto foo = GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT;
```
AlanGriffiths, line 125:
```suggestion
                auto constexpr foo = GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT;
```

src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -0,0 +1,831 @@
...
+        info1->clock == info2->clock &&
+        info1->hdisplay == info2->hdisplay &&
+        info1->hsync_start == info2->hsync_start &&
+        info1->hsync_end == info2->hsync_end &&
+        info1->htotal == info2->htotal &&
+        info1->hskew == info2->hskew &&
+        info1->vdisplay == info2->vdisplay &&
+        info1->vsync_start == info2->vsync_start &&
+        info1->vsync_end == info2->vsync_end &&
+        info1->vtotal == info2->vtotal;
```
AlanGriffiths, line 58:
```suggestion
        std::tie(info1->clock, info1->hdisplay, ...) ==
        std::tie(info2->clock, info2->hdisplay, ...);
```

src/platforms/atomic-kms/server/kms/display.h
```diff
@@ -0,0 +1,131 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_GRAPHICS_GBM_DISPLAY_H_
+#define MIR_GRAPHICS_GBM_DISPLAY_H_
```
AlanGriffiths, line 18:
```suggestion
#ifndef MIR_GRAPHICS_ATOMIC_GBM_DISPLAY_H_
#define MIR_GRAPHICS_ATOMIC_GBM_DISPLAY_H_
```

src/platforms/atomic-kms/server/kms/egl_helper.cpp
```diff
@@ -0,0 +1,286 @@
...
+ */
+
+#include "egl_helper.h"
+#include "mir/graphics/gl_config.h"
+#include "mir/graphics/egl_error.h"
+#include <boost/exception/errinfo_errno.hpp>
+#include <boost/throw_exception.hpp>
+#include <gbm.h>
+
+#define MIR_LOG_COMPONENT "EGL"
```
AlanGriffiths, line 24: This doesn't distinguish from gbm

src/platforms/atomic-kms/server/kms/egl_helper.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+    if (eglInitialize(dpy, &major, &minor) == EGL_FALSE)
+        BOOST_THROW_EXCEPTION(mg::egl_error("Failed to initialize EGL display"));
+
+    if ((major < minimum_major_version) ||
+        (major == minimum_major_version && minor < minimum_minor_version))
+    {
+        BOOST_THROW_EXCEPTION(
+            boost::enable_error_info(std::runtime_error("Incompatible EGL version")));
+        // TODO: Insert egl version major and minor into exception
```
AlanGriffiths, line 79: Let's resolve the TODO before landing

src/platforms/atomic-kms/server/kms/egl_helper.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+void mgmh::EGLHelper::report_egl_configuration(std::function<void(EGLDisplay, EGLConfig)> f)
+{
+    f(egl_display, egl_config);
+}
+
+mgmh::EGLHelper::NoMatchingEGLConfig::NoMatchingEGLConfig(uint32_t /*format*/)
+    : std::runtime_error("Failed to find matching EGL config")
+{
+    // TODO: Include the format string; need to extract from linux_dmabuf.cpp
```
AlanGriffiths, line 285: Let's resolve the TODO

src/platforms/atomic-kms/server/kms/egl_helper.h
```diff
@@ -0,0 +1,87 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_GRAPHICS_GBM_EGL_HELPER_H_
+#define MIR_GRAPHICS_GBM_EGL_HELPER_H_
```
AlanGriffiths, line 18:
```suggestion
#ifndef MIR_GRAPHICS_ATOMIC_EGL_HELPER_H_
#define MIR_GRAPHICS_ATOMIC_EGL_HELPER_H_
```

AlanGriffiths (changes requested):
A bunch of surface cruft that needs cleaning away. At least one more review pass needed

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
AlanGriffiths: On closer examination the code was copied from [elsewhere](https://github.com/canonical/mir/pull/3644) and we only sometimes leak the shared_ptr.

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+auto mga::GBMDisplayAllocator::supported_formats() const -> std::vector<DRMFormat>
+{
+    // TODO: Pull out of KMS plane info
+    return { DRMFormat{DRM_FORMAT_XRGB8888}, DRMFormat{DRM_FORMAT_ARGB8888}};
+}
+
+auto mga::GBMDisplayAllocator::modifiers_for_format(DRMFormat /*format*/) const -> std::vector<uint64_t>
+{
+    // TODO: Pull out off KMS plane info
```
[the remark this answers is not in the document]
AlanGriffiths: No worse than existing platforms

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -0,0 +1,214 @@
...
+mga::GBMDisplayAllocator::GBMDisplayAllocator(mir::Fd drm_fd, std::shared_ptr<struct gbm_device> gbm, geom::Size size)
+    : fd{std::move(drm_fd)},
+      gbm{std::move(gbm)},
+      size{size}
+{
+}
+
+auto mga::GBMDisplayAllocator::supported_formats() const -> std::vector<DRMFormat>
+{
+    // TODO: Pull out of KMS plane info
```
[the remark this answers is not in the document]
AlanGriffiths: No worse than existing platforms

AlanGriffiths (changes requested):
This appears to work with one or two monitors (I'm using Miriway/mir-pr3525)

However, when I plug or unplug an external monitor I get:

```
<- ERROR - > atomic-kms: Failed to schedule page flip: Invalid arguement (22)
Fatal signal received
...
``` 

[update]

Built Mir locally, with the "cleanup" code disabled, it didn't crash as consistently, but did after a couple of cycles.

The core wasn't as useful as I hoped: it died in an "out of order" shutdown (which is secondary to the problem causing the shutdown, but also worthy of investigation).

AlanGriffiths:
> This appears to work with one or two monitors (I'm using Miriway/mir-pr3525)
> 
> However, when I plug or unplug an external monitor I get:
> 
> ```
> <- ERROR - > atomic-kms: Failed to schedule page flip: Invalid arguement (22)
> Fatal signal received
> ...
> ```

Rats! This is still happening. (And the warning I introduced above isn't in the log). It must be a different failure mode than the one I fixed above.

AlanGriffiths (approved):
With the last fix, I'm content. Someone else should cast a vote too (especially about my last commit)

Saviq:
> Someone else should cast a vote too (especially about my last commit)

Looks legit to me, @RAOF feel free to merge if you agree with where the exception is caught.

RAOF:
> AHA! It took writing a script to try and reproduce this, but I now can. (The relevant magic is that you need to trigger the modeset _from the terminal that will be hung_)
> 
> What's happening here is that the final frame submitted by the terminal is racing the modeset sequence.
> 
> When it stops updating, what's happening is the terminal is waiting on a frame event from its last-buffer-before-modeset before submitting any new frames. Mir is not sending a frame event until _something else_ triggers a composition.
> 
> So this is a problem somewhere in our buffer queuing, I think.

Just to loop back to this; I can trivially reproduce this on `gbm-kms`, too.

RAOF:
Here are a couple of extra fixes - [one trivial](https://github.com/canonical/mir/pull/3525/commits/a816adbb49357ece4d5c71ce40d12dac08461507), and one thread race use-after-free.

Finally, I've identified a problem when disconnecting and re-connecting an external display to a different port on my laptop - we're not disassociating the CRTC from the old connector, so we run into resource conflicts. The fix *there* will be to update `clear_connected_unused_outputs` to also clear outputs which are *not* connected, but which have resources set up.

RAOF:
Bah, that doesn't fix it for some reason; I still see a CRTC connected to my disconnected output.

src/platforms/atomic-kms/server/kms/display.cpp
```diff
@@ -262,6 +262,8 @@ void mga::Display::clear_connected_unused_outputs()
 {
     current_display_configuration.for_each_output([&](DisplayConfigurationOutput const& conf_output)
     {
+        auto kms_output = current_display_configuration.get_output_for(conf_output.id);
```
AlanGriffiths, line 265:
```suggestion
        auto const kms_output = current_display_configuration.get_output_for(conf_output.id);
```

AlanGriffiths (approved):
Works well enough to land

AlanGriffiths:
> Works well enough to land

OTOH, I've tried using this a while and found that docking element to the right of my rightmost monitor crashes Miriway.

RAOF:
Now properly supports unplugging a monitor and plugging it in to a different port!

src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -0,0 +1,911 @@
...
+    catch (...)
+    {
+        /*
+         * In order to actually clear the output, we need to have a crtc
+         * connected to the output/connector so that we can disconnect
+         * it. However, not being able to get a crtc is OK, since it means
+         * that the output cannot be displaying anything anyway.
+         */
+        return;
+    }
```
AlanGriffiths, line 385: A try block around a single line of code followed by a catch block that silently eats everything suggests we are using the wrong reporting mechanism.

AlanGriffiths:
> and will be preferentially used on devices with atomic support

This isn't my experience. I have to explicitly choose `platform-display-libs=mir:atomic-kms` (defaulting to the well tested `mir:gbm-kms` isn't bad, just not what is advertised above)

AlanGriffiths:
> This isn't my experience. I have to explicitly choose `platform-display-libs=mir:atomic-kms` (defaulting to the well tested `mir:gbm-kms` isn't bad, just not what is advertised above)

As that's been confirmed by @Saviq, I'll fix the text

AlanGriffiths (approved):

PR #3531, opened 2024-08-07
live-input-configuation

AlanGriffiths:
Allow servers to dynamically update input configuration.

There's a very simple "mir_demo_server.input" addition to `mir_demo_server` to exercise part of this API, but that is not intended prototype.


AlanGriffiths:
This is still not complete:

- [ ] There's no code/tests to exercise this API
- [ ] There's no facility to recover the default or configured configuration

AlanGriffiths:
> * input configuration should remain in the main `.config` file

I don't think that a (hypothetical) input configuration utility should be updating the main `.config` file

AlanGriffiths:
> > I don't think that a (hypothetical) input configuration utility should be updating the main `.config` file
> 
> Agree, that will have to come from the shell implementation, e.g. integrating gsettings.

Right, so the input configuration needs to be exposed (like this WIP), not "remain in the main `.config` file"

Saviq:
> Right, so the input configuration needs to be exposed (like this WIP), not "remain in the main `.config` file"

Sorry, different trains of thought. …as opposed to a dedicated file.

examples/mir_demo_server/server_example.cpp
```diff
@@ -116,6 +119,38 @@ try
 {
     miral::MirRunner runner{argc, argv, "mir/mir_demo_server.config"};
 
+    miral::InputConfiguration input_configuration;
+
+    miral::ConfigFile test{runner, "mir_demo_server.input", miral::ConfigFile::Mode::reload_on_change,
+        [&input_configuration](auto& in, auto path)
+    {
+        std::cout << "** Reloading: " << path << std::endl;
```
mattkae, line 127: Should we use `mir::log_info(...)` in the example instead?

examples/mir_demo_server/server_example.cpp
```diff
@@ -116,6 +119,38 @@ try
...
+
+        auto mouse = input_configuration.mouse();
+        auto touchpad = input_configuration.touchpad();
+
+        for (std::string line; std::getline(in, line);)
+        {
+            std::cout << line << std::endl;
```
mattkae, line 134: Same here?

src/miral/input_config.h
```diff
@@ -0,0 +1,82 @@
...
+    std::optional<double> acceleration_bias;
+    std::optional<double> vscroll_speed;
+    std::optional<double> hscroll_speed;
+    std::optional<MirTouchpadClickMode> click_mode;
+    std::optional<MirTouchpadScrollMode> scroll_mode;
+    std::optional<bool> tap_to_click;
+    std::optional<bool> middle_button_emulation;
+};
+
+class InputDeviceConfig : public mir::input::InputDeviceObserver
```
mattkae, line 57: Should this file be renamed to `input_device_config.h`? It is a bit ambiguous at the moment next to `input_configuration.h`

src/miral/input_configuration.cpp
```diff
@@ -0,0 +1,323 @@
...
+    auto touchpad() -> Touchpad;
+    void touchpad(Touchpad const& val);
+
+    void apply(Mouse const& m)
+    {
+        if (input_device_hub)
+        {
+            input_device_hub->for_each_mutable_input_device([&m](auto& input_device)
+            {
+                m.self->apply_to(input_device);
```
mattkae, line 55: Question: is there a world where we have multiple mice and we want a different configuration for each? It seems unlikely

mattkae (changes requested):
Some minor things, but seems like a good change overall! Now we just need to get the keyboard stuff in there too :)

examples/mir_demo_server/server_example.cpp
AlanGriffiths: Yes

src/miral/input_configuration.cpp
AlanGriffiths: Agreed

examples/mir_demo_server/server_example.cpp
AlanGriffiths: There's a lot more `cout` than anything else in the examples: No `log_XXX` one instance of `mir::log()`.

So this is consistent with the example code. But maybe we want to change?

mattkae: I'm okay with `cout` if it is consistent. I didn't know whether or not we're pushing compositor authors to use mir's logging facilities (I am using them in miracle, for example)

AlanGriffiths: Maybe we should be. (I also do it in Miriway and Frame)

But that's another PR...

src/miral/input_config.h
AlanGriffiths: Done

mattkae (approved):
This appears to be a sensible change form the options that I tested!

mattkae:
@AlanGriffiths The symbols check is still upset though it seems!

AlanGriffiths:
Seems I'm plagued with #3581 today

PR #3539, opened 2024-08-08
ci: symbols on LLVM 18

Saviq:
Alan's reporting this is fine on clang-18

### DIFF
--- a/.github/workflows/symbols-check.yml
+++ b/.github/workflows/symbols-check.yml
@@ -22,12 +22,9 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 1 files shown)


mattkae (approved):
Yup! I suppose clang-18 supports c++23, so it shouldn't be a problem

Saviq:
@mattkae didn't you have trouble with 18? Or was it just that we didn't have 18, and 19 was available in the LLVM archives?

AlanGriffiths (approved):
Cool

mattkae:
> @mattkae didn't you have trouble with 18? Or was it just that we didn't have 18, and 19 was available in the LLVM archives?

19 was latest so I was just doing that. 18 is minimum :)

PR #3542, opened 2024-08-08
Attempt to fix XWayland app scaling

tarek-y-ismail:
Probably caused by my changes:
- [x] While apps and their decorations render correctly on XWayland now, input events are only received in the original area (non-scaled). So for a 2.0 scale factor, only the top left quarter works correctly.
- [ ] Wayland apps with server side decorations are broken: I probably broke something while focusing on fixing XWayland, but that's fine. I'll return to this once I fixed the XWayland problems and have a global view on what was changes.
- [ ] (?) Dragging to resize is not 1:1 with the mouse. Still works though

Already exists on main:
- [ ] Maximized windows are scaled outside the bounds of the screen, which shouldn't happen.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -1267,7 +1267,7 @@ void mf::XWaylandSurface::prep_surface_spec(ProofOfMutexLock const&, msh::Surfac
         }
     }
 
-    auto scale_size_clamped = [inv_scale](auto& optional_prop)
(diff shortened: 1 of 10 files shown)


tarek-y-ismail:
> - [ ] Wayland apps with server side decorations are broken: I probably broke something while focusing on fixing XWayland, but that's fine. I'll return to this once I fixed the XWayland problems and have a global view on what was changes.

This is most likely because I apply the X11 scale to all decorations regardless of whether they're for Wayland or XWayland applications.

PR #3544, opened 2024-08-09
miral::ConfigFile

AlanGriffiths:
A utility class to locate and (re)load configuration files when they change. The actual `Loader` processing of the file is one customization point, the reloading strategy another.

### DIFF
--- a/src/miral/reloading_config_file.cpp
+++ b/src/miral/reloading_config_file.cpp
@@ -136,43 +136,45 @@ void miral::ReloadingConfigFile::Self::register_handler(MirRunner& runner)
 {
     if (directory_watch_descriptor)
     {
-        fd_handle = runner.register_fd_handler(inotify_fd, [icf=inotify_fd, this] (int)
+        fd_handle = runner.register_fd_handler(inotify_fd, [this] (int)
         {
+            static size_t const sizeof_inotify_event = sizeof(inotify_event);
+
+            // A union ensures buffer is aligned for inotify_event
             union
             {
-                inotify_event event;
-                char buffer[sizeof(inotify_event) + NAME_MAX + 1];
-            } inotify_buffer;
+                char buffer[sizeof_inotify_event + NAME_MAX + 1];
+                inotify_event unused [[maybe_unused]];
+            } inotify_magic;
 
-            ssize_t readsize = read(icf, &inotify_buffer, sizeof(inotify_buffer));
-            if (readsize < static_cast<ssize_t>(sizeof(inotify_event)))
+            auto const readsize = read(inotify_fd, &inotify_magic, sizeof(inotify_magic));
+            if (readsize < static_cast<ssize_t>(sizeof_inotify_event))
             {
                 return;
             }
 
-            auto raw_buffer = inotify_buffer.buffer;
-            while (raw_buffer != inotify_buffer.buffer + readsize)
+            auto raw_buffer = inotify_magic.buffer;
+            while (raw_buffer != inotify_magic.buffer + readsize)
             {
+                // This is safe because inotify_magic.buffer is aligned and event.len includes padding for alignment
                 auto& event = reinterpret_cast<inotify_event&>(*raw_buffer);
-                if (event.mask & (IN_CLOSE_WRITE | IN_MOVED_TO))
+                if (event.mask & (IN_CLOSE_WRITE | IN_MOVED_TO) && event.wd == directory_watch_descriptor.value())
                 try
                 {
-                    if (event.name == filename)
-                    {
-                        auto const& file = directory.value() / filename;
+                    auto const& file = directory.value() / filename;
 
-                        if (std::ifstream config_file{file})
-                        {
-                            load_config(config_file, file);
-                        }
+                    if (std::ifstream config_file{file})
+                    {
+                        load_config(config_file, file);
                     }
                 }
                 catch (...)
                 {
-                    mir::log(mir::logging::Severity::warning, "ReloadingConfigFile", std::current_exception(), "Failed to reload configuration");
+                    mir::log(mir::logging::Severity::warning, "ReloadingConfigFile", std::current_exception(),
+                        "Failed to reload configuration");
                 }
 
-                raw_buffer += sizeof(struct inotify_event)+event.len;
+                raw_buffer += sizeof_inotify_event+event.len;
             }
         });
     }
--- a/src/miral/reloading_config_file.cpp
+++ b/src/miral/reloading_config_file.cpp
@@ -18,6 +18,7 @@
 
 #include <miral/runner.h>
 
+#define MIR_LOG_COMPONENT "ReloadingConfigFile"
 #include <mir/log.h>
 
 #include <boost/throw_exception.hpp>
@@ -91,6 +92,11 @@ miral::ReloadingConfigFile::Self::Self(MirRunner& runner, path file, Loader load
 {
     register_handler(runner);
 
+    if (directory_watch_descriptor.has_value())
+    {
+        mir::log_debug("Monitoring %s for configuration changes", (directory.value()/filename).c_str());
+    }
+
     // With C++26 we should be able to use the optional directory as a range to
     // initialize config_roots.  Until then, we'll just do it the long way...
     std::vector<path> config_roots;
@@ -120,6 +126,7 @@ miral::ReloadingConfigFile::Self::Self(MirRunner& runner, path file, Loader load
         if (std::ifstream config_file{filepath})
         {
             load_config(config_file, filepath);
+            mir::log_debug("Loaded %s", filepath.c_str());
             break;
         }
     }
@@ -168,13 +175,18 @@ void miral::ReloadingConfigFile::Self::register_handler(MirRunner& runner)
                         if (std::ifstream config_file{file})
                         {
                             load_config(config_file, file);
+                            mir::log_debug("(Re)loaded %s", file.c_str());
+                        }
+                        else
+                        {
+                            mir::log_debug("Failed to open %s", file.c_str());
                         }
                     }
                 }
                 catch (...)
                 {
-                    mir::log(mir::logging::Severity::warning, "ReloadingConfigFile", std::current_exception(),
-                        "Failed to reload configuration");
+                    mir::log(mir::logging::Severity::warning, MIR_LOG_COMPONENT, std::current_exception(),
+                             "Failed to reload configuration");
                 }
 
                 raw_buffer += sizeof_inotify_event+event.len;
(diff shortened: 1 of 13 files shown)


tarek-y-ismail (approved):

AlanGriffiths:
Split this out of #3531 as it is generally useful

tarek-y-ismail:
Might also handle #3478

Saviq (changes requested):
There's one thing that we could consider here: monitoring _all_ the applicable folders for new files. I mean, if a file shows isn't in `$XDG_CONFIG_HOME` on startup, but is added later, we won't consider it.

Also the `XDG_CONFIG_DIRS` handling is static - we only load from it once.

---

Could we have tests, please?

Saviq:
Should we consider `$PWD`, too?

It's surprising that this: `--display-config=static=foo.display` doesn't look in `$PWD` for the config file.

AlanGriffiths:
> Should we consider `$PWD`, too?

I vote not. The XDG Base Directory spec. doesn't mention `$PWD` (which can also change during the life of the application). I would be more inclined to drop support for absolute paths and treat relative paths as relative to the base directory list. So that servers could organise configuration files in a subdirectory. (I adopted what we currently do for .display, which may not be ideal.)

> It's surprising that this: `--display-config=static=foo.display` doesn't look in `$PWD` for the config file.

For arbitrary files I would agree. But _config_ files typically don't load from the current directory.

> There's one thing that we could consider here: monitoring _all_ the applicable folders for new files. I mean, if a file shows isn't in `$XDG_CONFIG_HOME` on startup, but is added later, we won't consider it.

Yes we do.

> Also the `XDG_CONFIG_DIRS` handling is static - we only load from it once.

We could, but is there any benefit to monitoring for system-wide configuration? Session wide configuration changes should be in `$XDG_CONFIG_HOME`.

> Could we have tests, please?

We could. Just seemed a bit much for glue code that hits the filesystem

AlanGriffiths:
> Adding some paint to the bikeshed :) But `--display-config=static=foo.display` is kind of odd to me

Before we paint the wrong bikeshed, this is about configuration files used by a compositor, not files specified on the command-line (or in the configuration).

src/miral/reloading_config_file.cpp
```diff
@@ -0,0 +1,169 @@
...
+    std::filesystem::path const filename;
+    std::optional<std::filesystem::path> const directory;
+    std::optional<mir::Fd> const directory_watch_fd;
+
+    void register_handler(MirRunner& runner);
+    std::unique_ptr<miral::FdHandle> fd_handle;
+};
+
+miral::ReloadingConfigFile::Self::Self(MirRunner& runner, path file, Loader load_config) :
+    inotify_fd{inotify_init()},
```
RAOF, line 86: We probably want to be using `inotify_init1(IN_CLOEXEC)` here, no? Make sure the watch isn't leaked to child processes.

RAOF (commented):
> We could, but is there any benefit to monitoring for system-wide configuration? Session wide configuration changes should be in `$XDG_CONFIG_HOME`.

I think we should be consistent between system-wide and user-local configuration behaviour.

Also, for the `ubuntu-frame` configuration do we try to stick it in the user-local or system-wide location?

AlanGriffiths:
OK, the following questions raised above have not been resolved:

1. How should we treat relative paths? They could be relative to the XDG base locations. (This would make monitoring more complex as we need to watch for subdirectory creation and for files within that.)
2. How do we treat absolute paths?
3. Should we monitor the `XDG_CONFIG_DIRS` paths? (This would make monitoring more complex as we need to prioritise one path over another.)

Also, not mentioned (yet):

4. Should we detect file deletion and load a config from a later base location?

The proposed code answers as follows:

1. How should we treat relative paths? Relative to `$PWD`, and monitor for changes
2. How do we treat absolute paths? As absolute paths, and monitor for changes
3. Should we monitor the `XDG_CONFIG_DIRS` paths? No
4. Should we detect file deletion and load a config from a later base location? No

Also, mentioned above, but I think resolved:

5. Should we look in `$PWD`? No

I think this is useful and understandable

src/miral/reloading_config_file.cpp
```diff
@@ -0,0 +1,184 @@
...
+                try
+                {
+                    if (event.name == filename)
+                    {
+                        auto const& file = directory.value() / filename;
+
+                        if (std::ifstream config_file{file})
+                        {
+                            load_config(config_file, file);
+                        }
```
Saviq, line 171: Else?

tests/miral/reloading_config_file.cpp
```diff
@@ -0,0 +1,357 @@
...
+char const* const home = "/tmp/test_reloading_config_file/home";
+char const* const home_config = "/tmp/test_reloading_config_file/home/.config";
+char const* const xdg_conf_home = "/tmp/test_reloading_config_file/xdg_conf_dir_home";
+char const* const xdg_conf_dir0 = "/tmp/test_reloading_config_file/xdg_conf_dir_zero";
+char const* const xdg_conf_dir1 = "/tmp/test_reloading_config_file/xdg_conf_dir_one";
+char const* const xdg_conf_dir2 = "/tmp/test_reloading_config_file/xdg_conf_dir_two";
+
+TestReloadingConfigFile::TestReloadingConfigFile()
+{
+    std::filesystem::remove_all("/tmp/test_reloading_config_file/");
```
Saviq, line 106: I'd rather this be a generated temp directory, but I've been down this rabbit hole a bit recently…

We are leaking it, though?

Saviq (approved):
>     1. How should we treat relative paths? Relative to `$PWD`, and monitor for changes
>     2. How do we treat absolute paths? As absolute paths, and monitor for changes

I think I'm OK with this. But maybe we want to be more verbose about it? Logging which file we've loaded and/or are monitoring?

>     3. Should we monitor the `XDG_CONFIG_DIRS` paths? No

Sounds like a potential extension later.

It does relate to:

> > There's one thing that we could consider here: monitoring _all_ the applicable folders for new files. I mean, if a file shows isn't in `$XDG_CONFIG_HOME` on startup, but is added later, we won't consider it.
> 
> Yes we do.

OK I miswrote. `XDG_CONFIG_HOME` (or rather, `config_directory()`) is the only folder we're monitoring. If things happen within "earlier" `XDG_CONFIG_DIRS`, we won't notice. But that may be extreme.

>     4. Should we detect file deletion and load a config from a later base location? No

My only problem with that is that we're behaving differently live than on startup.

>     5. Should we look in `$PWD`? No

:+1:

---

Thanks for the tests! Happy they helped :)

tests/miral/reloading_config_file.cpp
AlanGriffiths: We are leaking, yes. That is intentional: when tests fail it is far easier to examine what happened. (And it is small, on a temporary filesystem)

AlanGriffiths:
> OK I miswrote. `XDG_CONFIG_HOME` (or rather, `config_directory()`) is the only folder we're monitoring. If things happen within "earlier" `XDG_CONFIG_DIRS`, we won't notice. But that may be extreme.

There are no "earlier" `XDG_CONFIG_DIRS`: `$XDG_CONFIG_HOME`/`$HOME/.config` is the first place considered

src/miral/reloading_config_file.cpp
AlanGriffiths: Else we can't open the file, so no point trying to load a config from it

Saviq:
> There are no "earlier" `XDG_CONFIG_DIRS`: `$XDG_CONFIG_HOME`/`$HOME/.config` is the first place considered

There are, if there's more than one in `XDG_CONFIG_DIRS`, and only the last one is where you find the file. Either way, we're not monitoring _any_ `XDG_CONFIG_DIRS`, only looking at them at startup. That's something we should at least spell out. Unless I'm reading the code completely wrong, there's only ever one folder we're monitoring. And that's never any of `XDG_CONFIG_DIRS`.

AlanGriffiths:
> > There are no "earlier" `XDG_CONFIG_DIRS`: `$XDG_CONFIG_HOME`/`$HOME/.config` is the first place considered
> 
> There are, if there's more than one in `XDG_CONFIG_DIRS`, and only the last one is where you find the file. Either way, we're not monitoring _any_ `XDG_CONFIG_DIRS`, only looking at them at startup. That's something we should at least spell out. Unless I'm reading the code completely wrong, there's only ever one folder we're monitoring. And that's never any of `XDG_CONFIG_DIRS`.

We're in agreement on what happens, just not clear on the referent to "earlier"

Saviq:
> We're in agreement on what happens, just not clear on the referent to "earlier"

Anyway! I don't think it's a blocker, but we probably should spell it out that `XDG_CONFIG_DIRS` are only considered on startup.

Saviq (approved):

src/miral/reloading_config_file.cpp
Saviq: Let's log it, then?

Saviq (approved):
I would just like to see more logging here. But you have my ACK.

"Monitoring $path/$file for configuration changes"
"Loaded configuration from $path/$file"
"Failed to open $path/$file"

RAOF:
Oh, no. While pondering
> Should we monitor the XDG_CONFIG_DIRS paths? (This would make monitoring more complex as we need to prioritise one path over another.)
I was *going* to answer “Yes, we should monitor `XDG_CONFIG_DIRS`”. But then I considered again, and this is kinda unusual behaviour, and might trip people up (eg: saving partial config changes and having things break).

I *do* think that we should support changing the system-wide config at runtime, and the standard way of doing that is something like “reload config on `SIGHUP`”. And if we implemented *that* for system-wide config, would we still want to monitor the user-local file for changes?

Saviq:
> Oh, no. While pondering

To recap what we discussed:
- indeed maybe SIGHUP would be more appropriate to load the main `.config` file
  - it would be more in line with other long-running services
  - would allow for `systemctl --user reload …` when in a systemd session
  - avoids accidental / in-progress / incomplete saves
  - can be easily wrapped to cause live reloads
- display config we can keep reloading live, it's mostly used by Frame anyway - and so can use this util

AlanGriffiths:
A (different, in addition) summary of discussion:

* This PR was never intended as the foundation of future dynamic configuration support
  - It just generalises what we do for .display (and could also support Frame's diagnostic) and makes it available
* We might generalise this to offer options: e.g. load-on-startup, auto-reload and reload-on SIGHUP

Maybe I should rename the class `ConfigFile` and add a strategy enum with `load-on-startup and auto-reload` options?

tests/miral/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ target_link_libraries(miral-test-internal
 
 mir_add_wrapped_executable(miral-test NOINSTALL
     external_client.cpp
-    reloading_config_file.cpp
+        config_file.cpp
```
Saviq, line 72:
```suggestion
    config_file.cpp
```

Saviq (approved):
Whitespace nit, otherwise good with me.

PR #3548, opened 2024-08-13
Fix XWayland scaling attempt 2

### LINKED ISSUE #3508: Scaling of XWayland apps broken
It seems that we're a bit confused when `x11-scale=2.0`. Client *input* is scaled to the window border you see there, but the rendering is not :grin: 
Screenshot of the problem

tarek-y-ismail:
Fixes: #3554, #3508

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -441,6 +441,7 @@ mf::XWaylandSurface::XWaylandSurface(
               })}
 {
     cached.geometry = geometry;
+    cached.geometry.size = cached.geometry.size * scale;
     cached.override_redirect = override_redirect;
 
     uint32_t const value = XCB_EVENT_MASK_PROPERTY_CHANGE | XCB_EVENT_MASK_FOCUS_CHANGE;
--- a/src/server/frontend_xwayland/xwayland_surface_role.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface_role.cpp
@@ -125,7 +125,13 @@ void mf::XWaylandSurfaceRole::commit(WlSurfaceState const& state)
         BOOST_THROW_EXCEPTION(std::runtime_error("Got XWaylandSurfaceRole::commit() when the role had no surface"));
     }
 
-    wl_surface->commit(state);
+    // HACK: Some applications might scale up while users don't want them to
+    // `--x11-scale` helps with this by scaling down the application according
+    // to user preference.
+    auto state_copy{state};
+    state_copy.scale = scale;
+
+    wl_surface->commit(state_copy);
 
     auto const surface = this->scene_surface();
     auto const session = surface ? surface.value()->session().lock() : nullptr;
@@ -145,7 +151,7 @@ void mf::XWaylandSurfaceRole::commit(WlSurfaceState const& state)
         }
 
         std::vector<std::shared_ptr<void>> keep_alive_until_spec_is_used;
-        if (state.surface_data_needs_refresh())
+        if (state_copy.surface_data_needs_refresh())
         {
             populate_surface_data_scaled(wl_surface, scale, spec, keep_alive_until_spec_is_used);
         }
--- a/src/server/frontend_xwayland/xwayland_surface_role.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface_role.cpp
@@ -82,12 +82,6 @@ void mf::XWaylandSurfaceRole::populate_surface_data_scaled(
             }
             stream.displacement = stream.displacement * inv_scale;
         }
-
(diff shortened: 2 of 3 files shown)


src/server/frontend_xwayland/xwayland_surface.cpp
```diff
@@ -1267,26 +1267,6 @@ void mf::XWaylandSurface::prep_surface_spec(ProofOfMutexLock const&, msh::Surfac
...
-            }
-        };
-
-    scale_size_clamped(mods.width);
-    scale_size_clamped(mods.height);
-    scale_size_clamped(mods.min_width);
-    scale_size_clamped(mods.min_height);
-    scale_size_clamped(mods.max_width);
-    scale_size_clamped(mods.max_height);
-
```
AlanGriffiths, line 1289: We should not be dropping the clamping. (Rewriting using `std::clamp` could be a thing though)

AlanGriffiths (changes requested):
Apart from dropping the clamp logic, this looks sensible

src/server/frontend_xwayland/xwayland_surface.cpp
AlanGriffiths: That makes sense now.

AlanGriffiths (approved):
Looks sensible to me. Will leave @RAOF to have the final say

RAOF:
Hm. This is a partial fix, I think. The (some?) initial buffer(s?) is still wrong, but after that we get input/rendering alignment.

I'm not sure if it's scaled correctly (it *seems* like maybe my test clients are double the size I'd expect, but that might be an artifact of my test environment).

tarek-y-ismail:
> I'm not sure if it's scaled correctly (it _seems_ like maybe my test clients are double the size I'd expect, but that might be an artifact of my test environment).

You are correct. Should be fixed now

RAOF (changes requested):
Right. That has fixed the first frame being wrong.

However, further testing has identified that the content of X11 windows isn't getting scaled at all. Setting the `x11-scale` option changes the size that Mir will configure the window to, but the content remains unscaled.

You should be able to test this locally, even without a HiDPI monitor, by setting `x11-scale=0.5`. This *should* result in the content of X11 windows being half-size; what currently happens is the window is configured to be half-size, but the content is rendered at 1.0 scale.

Saviq:
> However, further testing has identified that the content of X11 windows isn't getting scaled at all. Setting the `x11-scale` option changes the size that Mir will configure the window to, but the content remains unscaled.
> 
> You should be able to test this locally, even without a HiDPI monitor, by setting `x11-scale=0.5`. This _should_ result in the content of X11 windows being half-size; what currently happens is the window is configured to be half-size, but the content is rendered at 1.0 scale.

Looks like you (too?) have `--x11-scale` in reverse:

> Our x11-scale option doesn't affect the client:
> ```
> $ miral-shell --help | grep --after 4 x11-scale
>   --x11-scale arg (=1.0)                The scale to assume X11 apps use. 
>                                         Defaults to the value of GDK_SCALE or 
>                                         1. Can be fractional. (Consider also 
>                                         setting GDK_SCALE in app-env-x11 when 
>                                         using this)
> ```

Saviq (approved):
Far I can tell, this looks good to me.

src/server/frontend_xwayland/xwayland_surface_role.cpp
```diff
@@ -125,7 +119,13 @@ void mf::XWaylandSurfaceRole::commit(WlSurfaceState const& state)
         BOOST_THROW_EXCEPTION(std::runtime_error("Got XWaylandSurfaceRole::commit() when the role had no surface"));
     }
 
-    wl_surface->commit(state);
+    // HACK: Some applications might scale up while users don't want them to
+    // `--x11-scale` helps with this by scaling down the application according
+    // to user preference.
+    auto state_copy{state};
+    state_copy.scale = scale;
```
RAOF, line 126: Oops! C++ will silently convert from `float` to `optional<int>`, meaning this only works for integer scale. (If you try `--x11-scale=0.5` Mir will immediately crash when the first X11 app starts, because this truncates `0.5f32` to `0i32`, and we later divide by `scale`)

RAOF (changes requested):
This works for `--x11-scale=2.0`, but still isn't quite right. :thinking: 

I'll sync up with you this afternoon and see if we can't work out what right *is*!

RAOF (approved):
This now works.

I continue to suspect that this is fixing the problem by adding a new incorrect transformation that counteracts an existing incorrect transformation, but this generates the correct behaviour so is a strict improvement over the status-quo!

I think we need an overall review of WM coordinate handling, and that will be easier to act upon if the current state of the code generates the right behaviour!

PR #3550, opened 2024-08-14
inotify_init1(IN_CLOEXEC)

AlanGriffiths:
As [mentioned](https://github.com/canonical/mir/pull/3544#discussion_r1716168734) by @RAOF, we don't want the FD leaking to child processes


Saviq (approved):
Looks legit!

PR #3561, opened 2024-08-20
platform/graphics: Use interface Option type in interfaces.

RAOF:
There's no particular reason these interfaces need the concrete `ProgramOption`, and it's simpler if we don't need to wildly `dynamic_cast<ProgramOption>` everywhere.

### DIFF
--- a/include/platform/mir/graphics/platform.h
+++ b/include/platform/mir/graphics/platform.h
@@ -41,7 +41,6 @@ class Surface;
 namespace options
 {
 class Option;
-class ProgramOption;
 }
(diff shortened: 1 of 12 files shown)


mattkae (approved):
That seems very reasonable to me. Why were we every casting to begin with?

AlanGriffiths:
> There's no particular reason these interfaces need the concrete `ProgramOption`, and it's simpler if we don't need to wildly `dynamic_cast<ProgramOption>` everywhere.

I remember someone claiming the platform ABI was going to be stable

PR #3564, opened 2024-08-21
Fix decorations eating into application area

tarek-y-ismail:
Now `xlogo` and other X11 applications are sized correctly!

### DIFF
--- a/src/include/server/mir/shell/decoration.h
+++ b/src/include/server/mir/shell/decoration.h
@@ -1,0 +1,28 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_SHELL_DECORATION_H_
+#define MIR_SHELL_DECORATION_H_
+
+#include "mir/geometry/forward.h"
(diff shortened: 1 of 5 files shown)


src/server/shell/decoration/basic_manager.cpp
```diff
@@ -95,6 +95,7 @@ void msd::BasicManager::decorate(std::shared_ptr<ms::Surface> const& surface)
         auto decoration = decoration_builder(locked_shell, surface);
         lock.lock();
         decoration->set_scale(scale);
+
```
tarek-y-ismail, line 98: Will remove in future rebase

tarek-y-ismail, line 98: Will remove in future rebase

src/include/server/mir/shell/decoration.h
```diff
@@ -0,0 +1,7 @@
+#include "mir/geometry/forward.h"
+#include "mir_toolkit/common.h"
+
+namespace mir::shell::decoration
+{
+    auto compute_size_with_decorations(geometry::Size content_size, MirWindowType type, MirWindowState state) -> geometry::Size;
+}
```
tarek-y-ismail, line 7: Might not be the best way to expose `compute_size_with_decorations` to xwayland code, suggestions welcome

src/include/server/mir/shell/decoration.h
AlanGriffiths, line 1: Lacks header comment and include guards

src/include/server/mir/shell/decoration.h
AlanGriffiths: This will cause issues when we come to expose a configuration point for decorations. But good for now.

AlanGriffiths (changes requested):
Needs some tiding up, but the idea looks sensible

src/include/server/mir/shell/decoration.h
tarek-y-ismail: What's another better approach I could try?

AlanGriffiths: This is a good approach (for now). Don't make things complicated to solve problems we haven't encountered yet.

AlanGriffiths:
I'm going to rework this and resubmit to tackle the problems with Wayland clients discussed in #3565

Issue #3559, opened 2024-08-19
SSDs not shown as negotiated


AlanGriffiths:
Running Miriway/edge...

neverputt requests SSD and is configured with SSD...

```
[2708099.380]  -> zxdg_decoration_manager_v1@15.get_toplevel_decoration(new id zxdg_toplevel_decoration_v1@30, xdg_toplevel@35)
[2708099.383]  -> zxdg_toplevel_decoration_v1@30.set_mode(2)
[2708099.421] zxdg_toplevel_decoration_v1@30.configure(2)
```
But no decorations are shown

PR #3565, opened 2024-08-22
Show/hide modifications on surface modification

AlanGriffiths:
Fixes: #3559

### DIFF
--- a/src/server/shell/abstract_shell.cpp
+++ b/src/server/shell/abstract_shell.cpp
@@ -302,6 +302,18 @@ void msh::AbstractShell::modify_surface(std::shared_ptr<scene::Session> const& s
             update_confinement_for(surface);
         }
     }
+
+    if (modifications.server_side_decorated.is_set())
+    {
+        if (modifications.server_side_decorated.value())
+        {
+            decoration_manager->decorate(surface);
+        }
+        else
+        {
+            decoration_manager->undecorate(surface);
+        }
+    }
 }
 
 void msh::AbstractShell::destroy_surface(


tarek-y-ismail (approved):
Curiously, this revealed two other bugs which didn't occur with QT apps. 
1. The app's content slightly overflows the decorations (see screenshot). I think this might be related to neverputt not being resizable/not responding to resize events?
2. Our decorations don't respect that an app is not resizable.

AlanGriffiths:
> 1. The app's content slightly overflows the decorations (see screenshot). I think this might be related to neverputt not being resizable/not responding to resize events? Might  also be related to https://github.com/canonical/mir/pull/3564

I'm investigating this - it isn't related to #3564 (that is X11 specific). But is likely the same bug that addresses - we're not resizing the window to reflect the need for decorations.

The fact that this affects Wayland apps as well as X11 ones suggests that we need a more generic approach than #3564.

I'll share results of my investigations soon.

> 2\. Our decorations don't respect that an app is not resizable.

Feel free to have a look at that one

tarek-y-ismail:
> > 2. Our decorations don't respect that an app is not resizable.
> 
> Feel free to have a look at that one

Out of curiosity I tried compiling and running SDL3's examples. All of the ones I tried are not resizable and they worked fine (decorations draw correctly and don't resize) on Mir. Additionally, I tried another SDL2 game (cdogs-sdl) and it seemed to start with the wrong decoration size then  it fixes itself after it finishes loading, so this makes me more suspicious of SDL2.

Another observation is when starting neverputt with the X11 backend, something similar also happens. The game starts with the content slightly overflowing, but as the first frame renders things get fixed.

AlanGriffiths:
> I'll share results of my investigations soon.

To confirm our suspicions: applying the same `compute_size_with_decorations()` transform when adding decorations causes the borders to appear at the correct size for neverputt. I think we should land this and rework #3564 to fix both X11 and Wayland (I've already started during my investigation, so happy to continue)

AlanGriffiths:
> Out of curiosity I tried compiling and running SDL3's examples. All of the ones I tried are not resizable and they worked fine (decorations draw correctly and don't resize) on Mir. Additionally, I tried another SDL2 game (cdogs-sdl) and it seemed to start with the wrong decoration size then it fixes itself after it finishes loading, so this makes me more suspicious of SDL2.
> 
> Another observation is when starting neverputt with the X11 backend, something similar also happens. The game starts with the content slightly overflowing, but as the first frame renders things get fixed.

I suspect SDL3 is doing things in the more reasonable order of negotiating decorations before creating the surface. Adding decorations to an existing window means we either have to resize the window or resize the client area. (Both options have  a downside, but resizing the window gives us more control)

Issue #3562, opened 2024-08-20
[2.19] MIR_SERVER_GRAPHICS_PLATFORM_VERSION needs bumping


AlanGriffiths:
#3561 changed the ABI

PR #3574, opened 2024-08-23
abi-breaks

AlanGriffiths:
We've changed some ABIs since 2.17

Fixes: #3562


src/server/symbols.map
```diff
@@ -392,7 +401,21 @@ global:
     mir::input::VirtualInputDevice::if_started_then*;
     mir::input::default_cursor_size*;
     mir::input::input_platform_from_graphics_module*;
+    mir::input::make_unique_context*;
     mir::input::probe_input_platforms*;
+    mir::input::receiver::XKBMapperRegistrar::XKBMapperRegistrar*;
```
mattkae, line 406: Huh? Why are these in the 2.17 stanza now?

mattkae: Is this the fault of the tool, or did you do this on  purpose?

mattkae (changes requested):

src/server/symbols.map
AlanGriffiths: Huh!? Everything should have been in the 2.18 stanza. Mo...

AlanGriffiths: Good catch. Fixed

mattkae (approved):
All symbols appear to be 2.18 now, so I am happy :)

PR #3575, opened 2024-08-23
Fix toplevel window resize pingpong

### LINKED ISSUE #3573: Apps constantly resizing
I've seen an intermittent (maybe once a week) "constantly resizing" when dragging edges to resize chromium-browser windows and some electron apps. Probably the same underlying cause as #3450. (A ping-pong of compositor requested sizes and buffers of a different size being submitted.)

@tarek-y-ismail says "I don't really have issues causing the bug. I just launch bomber and resize a couple of times"

And he's right!

tarek-y-ismail:
closes #3573.


src/server/frontend_wayland/wayland_surface_observer.cpp
```diff
@@ -77,7 +77,11 @@ void mf::WaylandSurfaceObserver::content_resized_to(ms::Surface const*, geom::Si
     run_on_wayland_thread_unless_window_destroyed(
         [content_size](Impl* impl, WindowWlSurfaceRole* window)
         {
-            if (impl->requested_size.value_or(impl->window_size) != content_size)
+            auto const should_resize =
+                (!impl->requested_size && impl->window_size != content_size) ||                                      // No requested resize, current size is different
+                (impl->requested_size && impl->requested_size != content_size && impl->window_size != content_size); // both requested and current sizes are different
```
AlanGriffiths, line 82: I think this approach is wrong. In essence, it is guessing whether the client has reacted to a configure event by looking at the current size. That will never work for all cases.

As we have sent a configure, we should be tracking the ack_configure (similar to the layer-shell logic you just fixed)

AlanGriffiths: (And, yes making the test more complex is tempting. I fell for it when I looked at https://github.com/canonical/mir/issues/3040 a year ago)

tarek-y-ismail:
So I think the "loop" was intentional, but the guard clause doesn't work because `requested_size` isn't cleared after the client size is set, thus checking against the OLD value if requested_size on the next iteration (which continuously flip-flops) instead of checking against the window size. Instead, since `requested_size` is used as a sort of double buffering, it should be cleared at some point (when `window_size` is updated for example`)

image

Always Sunny Reaction GIF

AlanGriffiths (approved):
As previously discussed, I don't think this is right, but it does improve things and doesn't seem to cause additional problems. (In particular, https://github.com/canonical/ubuntu-frame-osk/issues/58 remains fixed.)

(I have a suspicion that there may be more conflation of `window_size` and  `content_size` around that needs resolving for SSD, but that's out of scope for this PR)

PR #3577, opened 2024-08-27
tics: don't run it in parallel

Saviq:
TICS can't do parallel runs.

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -5,6 +5,9 @@ on:
     branches:
     - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
 jobs:
   Run:
     runs-on: ubuntu-24.04


.github/workflows/tics.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
     - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
mattkae, line 9: Would you also want `cancel-in-progress: true`?

mattkae (changes requested):
One question!

.github/workflows/tics.yml
Saviq: I thought no, because we want results for each merge into `main`, and then I'm not sure how well does TICS behave if interrupted.

Saviq: Or rather, we have evidence that it doesn't behave well if interrupted…

https://github.com/canonical/mir/actions/runs/10559902271

PR #3584, opened 2024-09-03
bugfix: disable window management tests on riscv

mattkae:


AlanGriffiths (approved):
I continue to hope that one day we'll discover these failures are caused by an actual bug that can be fixed

But this is a good practical workaround

PR #3585, opened 2024-09-04
Fix missing `wl_surface.enter` events for outputs enabled at runtime.

hbatagelo:
In my [previous fix](https://github.com/canonical/mir/pull/3430) for #3414, I accidentally disabled sending `wl_surface.enter` events for surfaces on outputs enabled at runtime. This PR resolves that issue while still addressing #3414.

### DIFF
--- a/src/server/frontend_wayland/output_manager.cpp
+++ b/src/server/frontend_wayland/output_manager.cpp
@@ -224,7 +224,10 @@ void mf::OutputGlobal::bind(wl_resource* resource)
 {
     auto const instance = new OutputInstance(resource, this);
     instances[instance->client].push_back(instance);
-    instance->output_config_changed(output_config);
+    for (auto const& listener : listeners)
+    {
+        listener->output_config_changed(output_config);
+    }
     instance->send_done();
 }
 


tarek-y-ismail (approved):

AlanGriffiths:
Trying this (miriway --channel=edge/mir-pr3585) I find that _disconnecting_ an external monitor gives:
```
[1204127.872] wl_surface@23.leave(nil)
```

When the display adapts (not sure "nil" is the best we can do but...)

When _reconnecting_ again:

**expect**: an `enter` event when the display adapts
**actual**: there's no `enter`

Curiously, when `disabling` & `enabling` the same output I do see the `enter` event as expected.

AlanGriffiths (approved):
Yep

tarek-y-ismail:
> [This may be spurious - I just noticed the amd64 snap build failed and I was testing edge]
> 
> Trying this (miriway --channel=edge/mir-pr3585) I find that _disconnecting_ an external monitor gives:
> 
> ```
> [1204127.872] wl_surface@23.leave(nil)
> ```
> 
> When the display adapts (not sure "nil" is the best we can do but...)
> 
> When _reconnecting_ again:
> 
> **expect**: an `enter` event when the display adapts **actual**: there's no `enter`
> 
> Curiously, when `disabling` & `enabling` the same output I do see the `enter` event as expected.

So disconnecting and reconnecting the monitor physically doesn't call `enter`, but disabling/re-enabling it through software does?

AlanGriffiths:
> So disconnecting and reconnecting the monitor _physically_ doesn't call `enter`, but disabling/re-enabling it through _software_ does?

Weird huh? (But I was testing main, not this PR)

tarek-y-ismail:
Well, when I use i3 and physically disconnect the HDMI on my external monitor, it doesn't call xrandr to update the output state, maybe this is a similar situation?

AlanGriffiths:
> Hmm, `miriway --channel=edge/mir-pr3585` works fine for me:

See "This may be spurious - I just noticed the amd64 snap build failed and I was testing edge"

PR #3586, opened 2024-09-05
OutputManager: When listeners are registered, they need to be notified of existing globals

AlanGriffiths:
In particular, `WindowWlSurfaceRole::output_global_created()` uses the notification to register as an observer on the global. Without this, it cannot track configuration changes

### DIFF
--- a/src/server/frontend_wayland/output_manager.cpp
+++ b/src/server/frontend_wayland/output_manager.cpp
@@ -310,6 +310,11 @@ auto mf::OutputManager::output_for(graphics::DisplayConfigurationOutputId id) ->
 void mf::OutputManager::add_listener(OutputManagerListener* listener)
 {
     listeners.push_back(listener);
+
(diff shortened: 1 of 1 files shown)


mattkae (approved):
This change makes sense to me.

However, this _does_ seem like a minor paradigm shift with our listeners. For example, `XdgOutputV1` is an `OutputConfigListener` who listens on `output_config_changed`. When we add the listener (in `XdgOutputV1`'s constructor), we don't automatically trigger the event for the listener, but rather allow the listener to manually trigger its own event.

It might be surprising that some listeners trigger events on listen, while others do not. But that's my 2c :)

AlanGriffiths:
> However, this _does_ seem like a minor paradigm shift with our listeners.

I'm sure there are both other registrations that do this and other registrations that don't. I guess it depends on what is being notified - in this case, I don't see a downside.

hbatagelo (approved):
Looks good to me.

PR #3587, opened 2024-09-05
Fix rendering of resized XWayland applications with client side decorations

### LINKED ISSUE #3570: IntelliJ/CLion using X11 backend renders incorrectly on resize
## How to reproduce
1. Open up IntelliJ/CLion
2. Try resizing the window, you'll see two different behaviors:
  If you enlarge or shrink the window, you'll see the UI update, but the rendering will be still confined to the old boundaries
  
  

https://github.com/user-attachments/assets/fa0c5b7a-f12e-46a1-b132-f1745396ae22

tarek-y-ismail:
Previously, resizing didn't work at all unless you did a vertical resize followed by a horizontal one (or vice versa). Now, resizes work but only if you shrink the application. Any resizing outside application bounds stops after a couple of pixels.

### DIFF
--- a/src/server/frontend_xwayland/xwayland_surface.cpp
+++ b/src/server/frontend_xwayland/xwayland_surface.cpp
@@ -1228,13 +1228,12 @@ void mf::XWaylandSurface::apply_any_mods_to_scene_surface()
             spec.value()->top_left.value() == scene_surface->top_left())
             spec.value()->top_left.consume();
 
-        if (spec.value()->width.is_set() &&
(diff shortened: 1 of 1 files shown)


tarek-y-ismail:
> The reason that moving the cursor outside the window stops the resizing is probably that target window loses focus

This will be fixed in another PR. I already had  an attempt at fixing it, thinking that the issue was related to input event handling. I'll have another go at it soon™ going for the focus angle.

AlanGriffiths (approved):
Not necessarily the right thing, but an improvement

Issue #3494, opened 2024-07-17
Our `fork`/`exec` spawning is unsafe


RAOF:
`exec_wayland()` is called after `fork()`:
https://github.com/canonical/mir/blob/953de04e07444a8f4410dd2f044ad9d24fbec2ef/src/server/frontend_xwayland/xwayland_server.cpp#L53-L98

Unfortunately, `man 2 fork` says:

> After a fork() in a multithreaded program, the child can safely call only async-signal-safe functions (see signal-safety(7)) until such time as it calls execve(2).

We should instead do as much setup as possible in the parent, and *then* call `fork()`. (Amusingly, even `execvp` is not async-signal-safe :woman_facepalming: )

AlanGriffiths:
> Amusingly, even `execvp` is not async-signal-safe

That probably doesn't matter: we should use `execve()` directly (we default `xwayland-path` to "/usr/bin/Xwayland", which doesn't need PATH handling)

RAOF:
Oops! When reviewing https://github.com/canonical/mir/pull/3497 I also see that `miral::launch_app_env` suffers from the same problems.

PR #3591, opened 2024-09-09
Rework `miral::launch_app_env()` to avoid calling functions that are not async-signal-safe after fork()

AlanGriffiths:
Fixes: #3494

### DIFF
--- a/src/miral/launch_app.cpp
+++ b/src/miral/launch_app.cpp
@@ -121,7 +121,7 @@ auto miral::launch_app_env(
     }
     else
     {
-        unsetenv("WAYLAND_DISPLAY");
+        application_environment.unsetenv("WAYLAND_DISPLAY");
     }
 
     if (x11_display)
@@ -130,7 +130,7 @@ auto miral::launch_app_env(
     }
     else
     {
-        unsetenv("DISPLAY");
+        application_environment.unsetenv("DISPLAY");
     }
 
     auto const exec_env = application_environment.exec_env();
--- a/src/miral/launch_app.cpp
+++ b/src/miral/launch_app.cpp
@@ -29,29 +29,69 @@
 
 namespace
 {
-auto mir_env_variables() -> std::vector<std::string>
+class Environment
+{
+public:
+    Environment();
+
+    void setenv(std::string const& name, std::string const& value);
+    void unsetenv(std::string const& name);
+    auto exec_env() const & -> std::vector<char const*>;
+private:
+    std::vector<std::string> env_strings;
+};
+
+Environment::Environment()
 {
     static char const mir_prefix[] = "MIR_";
-
-    std::vector<std::string> vars_to_remove;
 
     for (auto var = environ; *var; ++var)
     {
         auto const var_begin = *var;
-        if (strncmp(var_begin, mir_prefix, sizeof(mir_prefix) - 1) == 0)
+        if (strncmp(var_begin, mir_prefix, sizeof(mir_prefix) - 1) != 0)
         {
-            if (auto var_end = strchr(var_begin, '='))
-            {
-                vars_to_remove.emplace_back(var_begin, var_end);
-            }
-            else
-            {
-                vars_to_remove.emplace_back(var_begin);
-            }
+            env_strings.emplace_back(var_begin);
+        }
+    }
+}
+
+void Environment::setenv(std::string const& name, std::string const& value)
+{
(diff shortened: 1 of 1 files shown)


src/miral/launch_app.cpp
```diff
@@ -61,65 +101,62 @@ auto miral::launch_app_env(
...
-        {
-            unsetenv("WAYLAND_DISPLAY");
-        }
+    if (wayland_display)
+    {
+        application_environment.setenv("WAYLAND_DISPLAY", wayland_display.value()); // configure Wayland socket
+    }
+    else
+    {
+        unsetenv("WAYLAND_DISPLAY");
```
mattkae, line 124: Why not `application_environment.unsetenv(key);`?

mattkae (approved):
Is the problem with this that when you `fork`, the environment that you established isn't necessarily maintained?

src/miral/launch_app.cpp
```diff
@@ -29,29 +29,69 @@
...
+            env_strings.emplace_back(var_begin);
+        }
+    }
+}
+
+void Environment::setenv(std::string const& name, std::string const& value)
+{
+    auto const entry = name + "=" + value;
+
+    for (auto it = env_strings.begin(); it != env_strings.end(); ++it)
```
RAOF, line 62: This could be
```
for (auto& it : env_strings)
```
right? (plus the associated pointer->reference changes)

I'm not missing some C++ subtlety?

src/miral/launch_app.cpp
```diff
@@ -61,65 +101,62 @@ auto miral::launch_app_env(
...
+
+    if (pid == 0)
+    {
         mir::log_debug("Restoring sigmask");
         sigset_t all_signals;
         sigfillset(&all_signals);
         pthread_sigmask(SIG_UNBLOCK, &all_signals, nullptr);
 
-        execvp(exec_args[0], const_cast<char*const*>(exec_args.data()));
+        execvpe(exec_args[0], const_cast<char*const*>(exec_args.data()), const_cast<char*const*>(exec_env.data()));
```
RAOF, line 159: `execvpe`,sadly, isn't async-signal-safe (presumably because the obvious implementation requires memory allocation?).

I believe the correct approach here is to do `$PATH` handling ourselves; pre-construct all the possible paths that would be searched through, and then in the child calling `execve` on each in turn, annoyingly needing to keep track of error returns.

What a faff!

RAOF: For added bonus giggles, if `$PATH` is unset, we should be looking in [`_CS_PATH`](https://www.gnu.org/software/libc/manual/html_node/String-Parameters.html). Dear lord!

RAOF (approved):
So, I think this is still wrong, but I'm approving anyway because it's a strict improvement and the *correct* thing is *wild*.

Feel free to do the correct thing, or merge this as an improvement.

RAOF:
> Is the problem with this that when you `fork`, the environment that you established isn't necessarily maintained?

The problem is that when you call `fork`, other threads are in an unknown state¹, in much the same way as in a signal handler, so you can't call anything that isn't [async-signal-safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html).

¹: Notably, they may have taken any number of internal libc locks, which will *never* be released in the child because the while the *state* of everything is duplicated only one thread of execution is forked by `fork()`.

src/miral/launch_app.cpp
RAOF: (For the record, *none* of the `exec*p*` functions are async-signal-safe)

src/miral/launch_app.cpp
AlanGriffiths: Right

src/miral/launch_app.cpp
AlanGriffiths: > What a faff!

Indeed. This is something that library implementers could (and should) get right for the rest of us

AlanGriffiths: I don't think there's a need for memory allocation in "the obvious implementation": remember _C99_ has VLAs. (See https://codebrowser.dev/glibc/glibc/posix/execvpe.c.html for an example implementation.)

So requiring `execvpe()` to be async-signal-safe is plausible (but outside our authority).

mattkae (approved):
I am still okay with this, but I will leave it up to @RAOF to do the final merge, since he had the one concern about execvpe that he _might_ want to check out.

Issue #3252, opened 2024-02-23
Tracking bug for issues with gbm-kms platform on NVIDIA


Saviq:
NVIDIA is (slowly!) becoming more usable on gbm-kms, which is interesting to us for superior XWayland support.

There are still a few bugs (as of the NIDIA 545 drivers - currently the most recent drivers available in 24.04) that we either need to wait for fixes for or work around. Those are:

- [ ] `gbm_surface_create` fails with `flags != 0` (need driver quirk, as other drivers (correctly) require `GBM_BO_USE_SCANOUT` for scanout buffers)
- [ ] #3296
- [x] #3278
- [ ] Our platform probing does not know that `/dev/dri/card0` and `/dev/dri/renderD128` refer to the same physical device. This results in both `gbm-kms` (bound to `/dev/dri/renderD128`) and `eglstream-kms` (bound to `/dev/dri/card0`) being loaded as rendering platforms

Original bug:
```
[2024-02-23 08:59:11.472038] <information> xwayland: Starting XWayland                                                                                                        
(WW) Option "-listen" for file descriptors is deprecated                                                                                                                      
Please use "-listenfd" instead.                                                                                                                                               
(WW) Option "-listen" for file descriptors is deprecated                                                                                                                      
Please use "-listenfd" instead.                                                                                                                                               
eglInitialize() failed                                                                                                                                                        
EGL setup failed, disabling glamor                                                                                                                                            
Failed to initialize glamor, falling back to sw
```

> Saviq: Have a try with MIR_XWAYLAND_OPTION=-eglstream

> morphis: the above solves the `eglInitialize` part, `Failed to initialize glamor, falling back to sw` is still in there

Ref.:

https://github.com/canonical/mir/blob/d0a68ae3de28ce59923c13e2e89ddea9a9dfdc8b/src/platforms/eglstream-kms/server/platform.cpp#L176-L183

- #3182

Saviq:
> RAOF: One of the reasons this was never a priority is that passing -eglstream there will (might) get you accelerated rendering for the X server, it won't get you acceleration for X11 clients.
> Saviq: Wonder if GBM is the right direction here, then?
> RAOF: Yes; that has 3D accel plumbed through to clients.
> RAOF: And might even work!

RAOF:
Problems with NVIDIA on gbm-kms so far:
[...]

PR #3595, opened 2024-09-11
Atomic KMS: enable bypass

tarek-y-ismail:
Apologies for the general suckiness of the code, will improve that once it actually works.

To test:
```sh
./build/bin/miral-kiosk --show-splash=off --platform-display-libs=mir:atomic-kms --platform-rendering-libs=mir:gbm-kms --cursor=null &
sleep 2
glmark2-wayland
```

Note that until hardware cursor support is merged, the cursor must be disabled when testing bypass (current implementation works only with one renderable)

### DIFF
--- a/include/platform/mir/graphics/platform.h
+++ b/include/platform/mir/graphics/platform.h
@@ -403,7 +403,7 @@ public:
     virtual auto make_surface(DRMFormat format, std::span<uint64_t> modifiers) -> std::unique_ptr<GBMSurface> = 0;
 };
 
-class DmaBufBuffer;
+class DMABufBuffer;
 
 class DmaBufDisplayAllocator : public DisplayAllocator
 {
@@ -412,7 +412,7 @@ public:
     {
     };
 
-    virtual auto framebuffer_for(std::shared_ptr<DmaBufBuffer> buffer)
+    virtual auto framebuffer_for(std::shared_ptr<DMABufBuffer> buffer)
         -> std::unique_ptr<Framebuffer> = 0;
 };
 
--- a/include/platform/mir/graphics/platform.h
+++ b/include/platform/mir/graphics/platform.h
@@ -412,7 +412,7 @@ public:
     {
     };
 
-    virtual auto framebuffer_for(std::shared_ptr<DMABufBuffer> buffer)
+    virtual auto framebuffer_for(mir::graphics::DisplaySink& sink, std::shared_ptr<DMABufBuffer> buffer)
         -> std::unique_ptr<Framebuffer> = 0;
 };
 
--- a/include/platform/mir/graphics/platform.h
+++ b/include/platform/mir/graphics/platform.h
@@ -412,7 +412,7 @@ public:
     {
     };
 
-    virtual auto framebuffer_for(mir::graphics::DisplaySink& sink, std::shared_ptr<DMABufBuffer> buffer)
+    virtual auto framebuffer_for(std::shared_ptr<DMABufBuffer> buffer)
         -> std::unique_ptr<Framebuffer> = 0;
 };
 
--- a/src/platforms/atomic-kms/server/kms/display_buffer.cpp
+++ b/src/platforms/atomic-kms/server/kms/display_buffer.cpp
@@ -272,7 +272,7 @@ auto import_gbm_bo(
     std::shared_ptr<mg::DMABufBuffer> buffer,
     std::array<int, 4> dmabuf_fds,
     std::array<int, 4> pitches,
-    std::array<int, 4> offsets) -> struct gbm_bo*
+    std::array<int, 4> offsets) -> std::shared_ptr<struct gbm_bo>
 {
     auto const plane_descriptors = buffer->planes();
 
@@ -287,7 +287,8 @@ auto import_gbm_bo(
         .modifier = buffer->modifier().value_or(DRM_FORMAT_MOD_NONE),
     };
 
-    return gbm_bo_import(gbm.get(), GBM_BO_IMPORT_FD_MODIFIER, (void*)&import_data, GBM_BO_USE_SCANOUT);
+    auto* gbm_bo = gbm_bo_import(gbm.get(), GBM_BO_IMPORT_FD_MODIFIER, (void*)&import_data, GBM_BO_USE_SCANOUT);
+    return std::shared_ptr<struct gbm_bo>(gbm_bo, &gbm_bo_destroy);
 }
 
 auto drm_fb_id_from_dma_buffer(
@@ -295,7 +296,7 @@ auto drm_fb_id_from_dma_buffer(
     -> std::shared_ptr<uint32_t>
 {
     auto [dmabuf_fds, pitches, offsets, modifiers] = get_import_buffers(buffer);
-    auto* gbm_bo = import_gbm_bo(gbm, buffer, dmabuf_fds, pitches, offsets);
+    auto gbm_bo = import_gbm_bo(gbm, buffer, dmabuf_fds, pitches, offsets);
     if (!gbm_bo)
     {
         mir::log_warning("Failed to import buffer");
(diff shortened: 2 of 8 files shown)


src/platforms/atomic-kms/server/kms/display_buffer.cpp
```diff
@@ -335,6 +339,78 @@ void mga::DisplaySink::set_next_image(std::unique_ptr<Framebuffer> content)
...
+    auto height = buffer->size().height;
+    auto pixel_format = buffer->format();
+
+    uint32_t bo_handles[4] = {0};
+    uint32_t pitches[4] = {0};
+    uint32_t offsets[4] = {0};
+
+    for (std::size_t i = 0; i < std::min(4zu, plane_descriptors.size()); i++)
+    {
+        bo_handles[i] = plane_descriptors[i].dma_buf;
```
RAOF, line 358: This is where the problem lies; `drmModeAddFB2` expects GEM handles, and you're giving it dmabuf file descriptors.

[Here](https://gitlab.freedesktop.org/mesa/drm/-/blob/main/xf86drm.h?ref_type=heads#L816) is the necessary DRM API for converting the two, along with an ominous warning about the sharp edges of that API.

Since I think we share the underlying FD with the `RenderingPlatform` (and so with EGL etc), I think we'll need to do the “everything goes through gbm” route. *That* means going through [`gbm_bo_import`](https://gitlab.freedesktop.org/mesa/mesa/-/blob/a33bd78a543ca6de3b54fa78f2eb4f2403553876/src/gbm/main/gbm.h#L406) (with `type = GBM_BO_IMPORT_FD_MODIFIER` and `flags = GBM_BO_USE_SCANOUT`) and then pulling all the GEM handles out with [`gbm_bo_get_handle_for_plane`](https://gitlab.freedesktop.org/mesa/mesa/-/blob/a33bd78a543ca6de3b54fa78f2eb4f2403553876/src/gbm/main/gbm.h#L481) (and then handling the lifetime of the `gbm_bo`, keeping a track of it, and destroying it when necessary)

include/platform/mir/graphics/platform.h
```diff
@@ -396,7 +396,7 @@ class DmaBufDisplayAllocator : public DisplayAllocator
     {
     };
 
-    virtual auto framebuffer_for(std::shared_ptr<DmaBufBuffer> buffer)
+    virtual auto framebuffer_for(mir::graphics::DisplaySink& sink, std::shared_ptr<DMABufBuffer> buffer)
```
RAOF, line 399: You don't need this, as the `DisplayAllocator` is already bound to the `DisplaySink` (you get a `DisplayAllocator` *from* a `DisplaySink`, it's associated with that sink, and the lifetime is bound to that sink)

src/platforms/atomic-kms/server/kms/display_buffer.cpp
```diff
@@ -354,5 +498,13 @@ auto mga::DisplaySink::maybe_create_allocator(DisplayAllocator::Tag const& type_
         }
         return gbm_allocator.get();
     }
+    if(dynamic_cast<DmaBufDisplayAllocator::Tag const*>(&type_tag))
+    {
+        if (!bypass_allocator)
+        {
+           bypass_allocator = std::make_shared<DmaBufDisplayAllocator>(drm_fd());
```
RAOF, line 505: Just pass `gbm` to the `DmaBufDisplayAllocator` constructor here.

tarek-y-ismail:
Could reuse a rebase after we figure out why artifacts appear when bypass is active.

Saviq:
> Could reuse a rebase after we figure out why artifacts appear when bypass is active.

Feel free to rebase the chain of PRs already.

src/platforms/atomic-kms/server/kms/display_buffer.cpp
```diff
@@ -335,6 +344,136 @@ void mga::DisplaySink::set_next_image(std::unique_ptr<Framebuffer> content)
...
+            mir::log_warning("Failed to import buffer");
+            return {};
+        }
+
+        auto const plane_descriptors = buffer->planes();
+        uint32_t gem_handles[MAX_PLANES] = {0};
+        for (std::size_t i = 0; i < std::min(MAX_PLANES, plane_descriptors.size()); i++)
+            gem_handles[i] = gbm_bo_get_handle_for_plane(gbm_bo, i).u32;
+
+        auto fb_id = std::shared_ptr<uint32_t>{
```
RAOF, line 413: As it stands, I think this code leaks a `gbm_bo` each time it's called? It calls `gbm_bo_import`, which will allocate a `gbm_bo`, but nothing every frees it.

Maybe `import_gbm_bo` should return a `shared_ptr<gbm_bo>` with cleanup functor, and that should in turn be captured by *this* `shared_ptr`?

RAOF (changes requested):

src/platforms/atomic-kms/server/kms/display_buffer.cpp
tarek-y-ismail: Hmm, I didn't know that, I thought we only got a (cached) handle to the buffer object (i.e. created once on the first import and reused later).

tarek-y-ismail: Also, I've noticed that we import and destroy a buffer object every frame, this doesn't feel correct. I think it might also be related to the artifacts showing in glmark2

AlanGriffiths:
@tarek-y-ismail what is the status of this PR? Clearly it needs conflicts resolved, but does it need more?

tarek-y-ismail:
> what is the status of this PR? Clearly it needs conflicts resolved, but does it need more?

@RAOF is investigating the artifacts.

tarek-y-ismail:
Small update:
To help pinpoint the cause of artifacts and figure out whether it's a driver thing, a mesa thing, or a kernel thing (both AMD and Intel GPUs show the bug, we still don't know about nvidia since  it's naughty with atomic KMS), I tried running with `--driver-quirks=allow:driver:nvidia`, which caused a crash in `gbm_surface_create` as mentioned in #3252. Hardcoding `0` for flags caused another crash somewhere in [`drm_fb_id_from_dma_buffer`](https://github.com/canonical/mir/blob/82cdf9c0cbd6ec8ffd0abb37f40a555fe5ae1d7c/src/platforms/atomic-kms/server/kms/display_buffer.cpp#L390), related to `gbm_bo`. I tried debugging it for a bit, but couldn't  pinpoint further.

RAOF:
This works correctly (as in: GL composition is bypassed, rendering is correct) on all devices I can test at the moment - both my (amdgpu) desktop and my (i915) laptop bypass without rendering artefacts.

I can't test my (amdgpu) Framework 16, because `vkgears` now submits buffers in a non-scanout capable format and we don't implement [`zwp_linux_dmabuf_feedback_v1`](https://wayland.app/protocols/linux-dmabuf-v1#zwp_linux_dmabuf_feedback_v1) which would be necessary to ask client apps to re-allocate with a supported buffer format.

RAOF:
With a quick hack to our linux_dmabuf implementation (to *only* send modifiers that can be scannout-ed), it seems that this renders correctly on my Framework 16 as well; I am unable to reproduce the (extremely obvious) corruption that used to be visible.

Either a mesa or kernel update has fixed whatever problem I was seeing, or the explicit-sync work *has* fixed this for me.

Saviq:
Having tried things out here, we found that it's only `glmark2-wayland` that's triggering those (now?) on my `Intel Meteorlake (Gen12)`.

`vkgears` behaves fine.

tarek-y-ismail:
> vkgears behaves fine.

Interesting, I've been using glmark2-wayland so far thinking that either could work... I'll try testing vkgears ASAP

tarek-y-ismail:
Ok, so I just re-tested things. `vkgears` bypasses and I can't see any artifacts,  but `glmark2-wayland` still exhibits artifacts. Out of curiosity, I tested `bypass-for-fw16` as well. Same behavior with `vkgears`/`glmark2-wayland`. The only difference is that the artifacts there are different.

So now we know that on nvidia and intel Xe GPUs, `glmark2-wayland` misbehaves. While on AMD and older intel GPUs it works fine.

RAOF:
@tarek-y-ismail I'm *pretty* sure the most recent commit fixes all the artefacts. (It may not be the *best* solution, but it is one).

tarek-y-ismail:
> @tarek-y-ismail I'm pretty sure the most recent commit fixes all the artefacts. (It may not be the best solution, but it is one).

Sure does fix it on my end. Both `glmark2-wayland` and `vkgears` show no artefacts on my end

Saviq (approved):
Yess!

And from some quick tests I'm seeing a 10% improvement in the `GLMark2Wayland.fullscreen` score :+1:

PR #3596, opened 2024-09-12
platforms/atomic-kms: Populate supported pixel formats correctly

RAOF:
`DisplayConfigurationOutput.pixel_formats` now contains the list of accepted pixel formats (or, at least, those pixel formats that are representable in the `MirPixelFormats` enum; many aren't).

### DIFF
--- a/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
+++ b/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
@@ -667,8 +667,7 @@ auto formats_for_output(mir::Fd const& drm_fd, mgk::DRMModeConnectorUPtr const& 
     }
 
     PropertyBlobData format_blob{drm_fd, static_cast<uint32_t>(plane_props["IN_FORMATS"])};
-    drmModeFormatModifierIterator iter;
-    memset(&iter, 0, sizeof(iter));
+    drmModeFormatModifierIterator iter{};
 
     std::vector<mg::DRMFormat> supported_formats;
     while (drmModeFormatModifierBlobIterNext(format_blob.raw(), &iter))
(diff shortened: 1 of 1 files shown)


src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -647,6 +654,43 @@ std::vector<uint8_t> edid_for_connector(mir::Fd const& drm_fd, uint32_t connecto
...
+    mgk::ObjectProperties plane_props{drm_fd, plane->plane_id, DRM_MODE_OBJECT_PLANE};
+
+    if (!plane_props.has_property("IN_FORMATS"))
+    {
+        return {mg::DRMFormat{DRM_FORMAT_ARGB8888}, mg::DRMFormat{DRM_FORMAT_XRGB8888} };
+    }
+
+    PropertyBlobData format_blob{drm_fd, static_cast<uint32_t>(plane_props["IN_FORMATS"])};
+    drmModeFormatModifierIterator iter;
+    memset(&iter, 0, sizeof(iter));
```
AlanGriffiths, line 671: Using `memset()` makes assumptions of about the representation of data members that are not guaranteed by the language.  value-initialization should do the job:

```suggestion
    drmModeFormatModifierIterator iter{};
```

src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -647,6 +654,43 @@ std::vector<uint8_t> edid_for_connector(mir::Fd const& drm_fd, uint32_t connecto
...
+         * {fmt1, mod3}
+         * {fmt2, mod2}
+         * {fmt2, mod4}
+         * ...
+         *
+         * We only care about the format, so we only add when we see a new format
+         */
+        if (supported_formats.empty() || supported_formats.back() != iter.fmt)
+        {
+            supported_formats.push_back(mg::DRMFormat{iter.fmt});
```
AlanGriffiths, line 689:
```suggestion
            supported_formats.emplace_back(iter.fmt);
```

AlanGriffiths (changes requested):
A couple of nits picked

Issue #3592, opened 2024-09-10
Chromium unmaximises when focus is lost


AlanGriffiths:
Under a Mir server (`miral-app` or `miriway`)

Run the chromium snap on Wayland

    chromium --enable-features=UseOzonePlatform --ozone-platform=wayland

Maximise it (e.g. double click titlebar)

Start another all (`Ctrl-Alt-T`)

**Expect:** app (terminal) appears, chromium remains maximised
**Actual:** app (terminal) appears, chromium is unmaximised

AlanGriffiths:
Hmm. Reverting chromium (to 127.0.6533.72) fixes the behaviour

PR #3599, opened 2024-09-18
Revert "Fix toplevel window resize pingpong (#3575)" and fix better

AlanGriffiths:
Fixes: #3600 
Fixes: #3592 
Fixes: #3573

### DIFF
--- a/src/server/frontend_wayland/wayland_surface_observer.h
+++ b/src/server/frontend_wayland/wayland_surface_observer.h
@@ -64,7 +64,6 @@ public:
     void latest_client_size(geometry::Size window_size)
     {
         impl->window_size = window_size;
-        impl->requested_size = {};
     }
 
     /// Should only be called from the Wayland thread
--- a/src/server/frontend_wayland/wayland_surface_observer.cpp
+++ b/src/server/frontend_wayland/wayland_surface_observer.cpp
@@ -77,11 +77,21 @@ void mf::WaylandSurfaceObserver::content_resized_to(ms::Surface const*, geom::Si
     run_on_wayland_thread_unless_window_destroyed(
         [content_size](Impl* impl, WindowWlSurfaceRole* window)
         {
-            if (impl->requested_size.value_or(impl->window_size) != content_size)
+            if (impl->requested_size && impl->requested_size == content_size)
             {
-                impl->requested_size = content_size;
-                window->handle_resize(std::nullopt, content_size);
+                // We've already requested this size, no action required
+                return;
             }
+
+            if (content_size == impl->window_size)
+            {
+                // The window manager is accepting the client's window size, no action required
+                return;
+            }
+
+            // The window manager is changing the client's window size, tell the client
+            impl->requested_size = content_size;
+            window->handle_resize(std::nullopt, content_size);
         });
 }
 


src/server/frontend_wayland/wayland_surface_observer.h
```diff
@@ -64,7 +64,10 @@ class WaylandSurfaceObserver : public scene::NullSurfaceObserver
     void latest_client_size(geometry::Size window_size)
     {
         impl->window_size = window_size;
-        impl->requested_size = {};
+        if (auto& r = impl->requested_size; r && r.value() == window_size)
+        {
+            r.reset();
+        }
```
Saviq, line 70: Any chance for a regression test?

AlanGriffiths:
> don't know how to build ubuntu-frame with a local Mir build

You don't need to: `snap refresh --channel=edge/mir-pr3599 ubuntu-frame`

> `mir-kiosk`, then `wpe-webkit-mir-kiosk.cog`

Are you running the local build?

`$BUILD_DIR/miral-app -kiosk -terminal wpe-webkit-mir-kiosk.cog`

tarek-y-ismail:
> You don't need to: `snap refresh --channel=edge/mir-pr3599 ubuntu-frame`

> Are you running the local build?
Indeed I am, doubled checked that I'm on the correct branch and that the build is up to date. Running the command you supplied results in the same output. 

Both result in the same output (window management-wise, WPE seems a bit broken here...)
image

AlanGriffiths:
> Running the command you supplied results in the same output

Just double checking: you did replace `$BUILD_DIR` didn't you?

Saviq:
Bug looks fixed, to me.

```
$ snap list frame-it ubuntu-frame  
Name          Version         Rev    Tracking       Publisher       Notes
frame-it      1.2             18     latest/stable  alangriffiths✪  classic
ubuntu-frame  198-mirv2.18.0  10866  24/edge/…      canonical✓      -

$ frame-it wpe-webkit-mir-kiosk.cog
```

image

AlanGriffiths:
I do, however, find that `kgx` still isn't handled correctly:

    frame-it kgx

NOT fullscreened. (Although it doesn't have rounded corners, so it "thinks" it is)

tarek-y-ismail:
> Bug looks fixed, to me.

Very interesting, I have the same versions but the output is similar to the image I posted earlier. What could be the cause? I assume I'm the outlier here?

AlanGriffiths:
> I assume I'm the outlier here?

Yes. It is likely a timing issue, not something you are doing wrong. (The WAYLAND_DEBUG log might clarify things (see [this command](https://github.com/canonical/mir/issues/3600#issuecomment-2358551152), but isn't essential)

Anyway, I think this approach (resetting `requested_size` _here_) is fundamentally flawed as we're potentially sending incorrect configure events as a result.

Simply reverting #3600 (to get the behaviour of 2.17) is also unattractive as that reintroduces different problems. I'm still thinking...

tarek-y-ismail:
> Yes. It is likely a timing issue

Hmm, that would mean that it should be inconsistent on both sides, right?

AlanGriffiths:
> Hmm, that would mean that it should be inconsistent on both sides, right?

No. The timing could be different between your kit and ours

src/server/frontend_wayland/wayland_surface_observer.h
AlanGriffiths: We need tests around this, and @tarek-y-ismail tried to write one of them. But we're not there yet

AlanGriffiths:
OK, while I rethink "fix better" I'm going to drop the first attempt. That means everyone has build of the "Revert" to experiment with.

AlanGriffiths:
Still got to assess test results.

AlanGriffiths:
OK, as noted above, there's a bunch of weird stuff. This does not fix that, just the cased noted in the linked bugs.

I'll follow up with a PR that addresses "weird stuff", but if manual testing (yes, I know!) verifies my findings this should be good to cherry-pick into 2.18.1

tarek-y-ismail:
Can confirm that `bomber` behaves well on `miral-shell`. And `wpe-webkit-mir-kiosk.cog`, `kgx`, and `bomber` are all sized correctly inside `mir-kiosk`.

What the last commit essentially do is "ignore the resize request if the newly requested size is the same as the last requested size, or the window size". I think we stumbled on this before but I'm not sure why we didn't pursue it?

tarek-y-ismail (approved):

PR #3604, opened 2024-09-20
Add APIs to override decoration managers and customize decorations

tarek-y-ismail:


src/include/server/mir/server_configuration.h
```diff
@@ -114,6 +115,8 @@ class ServerConfiguration
 
     virtual auto the_decoration_strategy() -> std::shared_ptr<DecorationStrategy> = 0;
     virtual void set_the_decoration_strategy(std::shared_ptr<DecorationStrategy> strategy) = 0;
+
+    virtual void set_the_decoration_manager(std::shared_ptr<shell::decoration::Manager> manager) = 0;
```
AlanGriffiths, line 119: This is a pointless API. There's no way for a caller to pass a `shell::decoration::Manager` as that interface isn't available to implement

tarek-y-ismail: This is more of a leftover from when I was experimenting with decorations. I'll push a commit to "expose" that interface (apologies in advance if I do it wrong)

tarek-y-ismail: Should be handled now

src/include/server/mir/shell/decoration.h
```diff
@@ -20,6 +20,122 @@
...
+    }
+
+    virtual ~Decoration()
+    {
+        if(decoration_surface)
+            decoration_surface->unregister_interest(*input_observer);
+
+        if(window_surface)
+            window_surface->unregister_interest(*window_observer);
+    }
```
AlanGriffiths, line 111: Far too much implementation exposed in a header file. That is: "Wrapping" the decoration implementation with observers should be factored out into a wrapper class that integrates the supplied decoration into the rest of the system.

We probably want `Decoration` to be an interface. Vis:

```suggestion
    Decoration() = default;
    virtual ~Decoration() = default;
```

(And with no data members)

src/include/server/mir/shell/decoration.h
```diff
@@ -20,6 +20,122 @@
...
+#include <mir/events/event.h>
+
+#include <string>
+
+namespace mir
+{
+namespace shell
+{
+namespace decoration
+{
```
AlanGriffiths, line 37: There's no need for two identical namespace blocks in this header

src/include/server/mir/shell/decoration.h
```diff
@@ -20,6 +20,122 @@
...
+        if(decoration_surface)
+            decoration_surface->unregister_interest(*input_observer);
+
+        if(window_surface)
+            window_surface->unregister_interest(*window_observer);
+    }
+
+    // Called whenever the visible aspect of the window changes.
+    // Examples: Scale, title, focus, visibility, size, etc...
+    virtual void redraw() = 0;
```
AlanGriffiths, line 115: That sounds like it doesn't belong in the interface. Surely it is up to the implementation whether it uses a common `redraw()` method to process all these changes?

Alternatively, if this is the only correct way to implement this, then there should not be separate `set_scale()`, etc methods as well.

AlanGriffiths (changes requested):

src/include/server/mir/shell/decoration.h
tarek-y-ismail: There is no "right" way, really. Just different ways of doing the same thing. This is the most "easy to follow" interface I was able to extract from the code we currently have implemented. Additionally, I had in mind that a decoration author might want to redraw on every frame (for animations and such), so having a common redraw method is not a bad idea in my opionion.

`set_scale` and friends exist to allow decoration authors to be able to handle state changes, they could invoke `redraw` immediately inside of them, or just update the state and redraw at some other point.

AlanGriffiths (changes requested):

src/include/server/mir/shell/decoration.h
tarek-y-ismail: You're right, I originally meant to leave it to allow for us to maybe have fancier decoratios that update independently of input, but that's too far from now.

src/server/shell/decoration/basic_decoration.cpp
```diff
@@ -269,35 +293,48 @@ void msd::BasicDecoration::set_cursor(std::string const& cursor_image_name)
 void msd::BasicDecoration::set_scale(float new_scale)
 {
     scale = new_scale;
-    window_state_updated();
+    redraw();
```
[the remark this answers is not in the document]
tarek-y-ismail: I was following the existing code. I'm not sure if `set_scale` can be called from multiple threads simultaneously. Doesn't that only happen when we modify the display config or move the window between monitors with different scales?

src/server/symbols.map
```diff
@@ -1403,3 +1403,25 @@ global:
...
+    non-virtual?thunk?to?mir::shell::decoration::Decoration::?Decoration*;
+    non-virtual?thunk?to?mir::shell::decoration::Manager::?Manager*;
+    typeinfo?for?mir::shell::decoration::Decoration;
+    typeinfo?for?mir::shell::decoration::DecorationNotifier;
+    typeinfo?for?mir::shell::decoration::Manager;
+    virtual?thunk?to?mir::DefaultServerConfiguration::set_the_decoration_manager*;
+    vtable?for?mir::shell::decoration::Decoration;
+    vtable?for?mir::shell::decoration::Manager;
+  };
+} MIR_SERVER_INTERNAL_2.18;
```
AlanGriffiths, line 1427: Only if ABI is preserved do we keep the old stanza.

The mirserver ABI is being broken. Therefore the soname needs to be bumped and _all_ the symbols go into the `MIR_SERVER_INTERNAL_2.19` stanza.

src/server/shell/decoration/basic_decoration.cpp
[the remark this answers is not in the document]
AlanGriffiths: It doesn't matter how many threads call `set_scale()` at the same time. What matters is if multiple threads access the same memory.

E.g. `scale` is also accessed by `new_window_state()`, so can you guarantee that all calls to `set_scale()` and `new_window_state()` occur either:

1. on the _same_ thread; OR,
2.  guarded by the same synchronization primitive (e.g. mutex)?

The same applies to `redraw()` as that also contains no synchronization - are all calls to `redraw()` on the same thread?

src/server/shell/decoration/basic_decoration.h
```diff
@@ -75,25 +74,27 @@ class BasicDecoration
         std::shared_ptr<scene::Surface> const& window_surface);
     ~BasicDecoration();
 
-    void window_state_updated();
-    void input_state_updated();
+    void redraw();
```
AlanGriffiths, line 77: `redraw()` can probably be made private - it is only accessed by member functions.

src/server/shell/decoration/basic_decoration.h
```diff
@@ -75,25 +74,27 @@ class BasicDecoration
...
     void request_resize(MirInputEvent const* event, MirResizeEdge edge);
     void request_toggle_maximize();
     void request_minimize();
     void request_close();
-    void set_cursor(std::string const& cursor_image_name);
 
-    void set_scale(float scale) override;
+    void set_cursor(std::string const& cursor_image_name);
 
 protected:
```
AlanGriffiths, line 94: There's no reason for `protected` here (there's rarely a good reason), this should be `private`.

AlanGriffiths (changes requested):

src/server/shell/decoration/basic_decoration.cpp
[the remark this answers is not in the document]
tarek-y-ismail: So here's what I managed to figure out

Decoration::set_scale:
  - Called inside BasicManager::decorate (synchronized)
  - Called inside BasicManager::set_scale (synchronized)

BasicDecoration::redraw:
  - Depends on the thread safety of set_scale (safe)
  - Remainder of calls are through `threadsafe_self`

BasicDecoration::new_window_state:
  - Depends on the thread safety of BasicDecoration::redraw (safe)
  - Called inside BasicDecoration::BasicDecoration and BasicDecoration::redraw
    - Can't redraw before the constructor finishes

So these methods seem safe, yes. But I have to say that things could be a lot clearer. I propose (in another PR of course) that we try to bring down all locking/synchronization (if possible) to the decoration itself.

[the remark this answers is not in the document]
AlanGriffiths: > * Depends on the thread safety of set_scale (safe)
> 
> * Remainder of calls are through `threadsafe_self`

In other words this is unsafe.

From what you say above, `set_scale()` is synchronised on `BasicManager::mutex`. Whereas "Remainder of calls" are synchronised on different mutex (a `ThreadsafeSelf::mutex`).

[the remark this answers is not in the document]
tarek-y-ismail: Ok, the multiple locks thing went right over my head. `redraw` inside `set_scale` is now accessed through `threadsafe_self`. I gave all other `BasicDecoration` methods a quick skim to make sure they're safe and they seem so.

src/server/shell/decoration/basic_manager.h
```diff
@@ -47,27 +48,31 @@ class BasicManager :
...
     ~BasicManager();
 
     void init(std::weak_ptr<shell::Shell> const& shell) override;
     void decorate(std::shared_ptr<scene::Surface> const& surface) override;
     void undecorate(std::shared_ptr<scene::Surface> const& surface) override;
     void undecorate_all() override;
 
+protected:
+    std::unordered_map<scene::Surface*, std::unique_ptr<Decoration>> decorations;
+
```
AlanGriffiths, line 66: I'm not enthusiastic about making member data public (`protected` is public in disguise).

In this case the only use is in tests, which manipulate `decorations` directly. That seems fragile and suggests there's an abstraction missing.

I'll try to come up with a concrete suggestion.

AlanGriffiths (changes requested):
This looks mostly sensible, but I'd like to see an example using this API. Can we add something to miral-shell?

tarek-y-ismail:
> This looks mostly sensible, but I'd like to see an example using this API. Can we add something to miral-shell?

Sure thing. Anything quick in mind? The first thing that comes to mind for me is a variant of `BasicDecoration` that assigns a random color to each new decoration.

AlanGriffiths:
> > This looks mostly sensible, but I'd like to see an example using this API. Can we add something to miral-shell?
> 
> Sure thing. Anything quick in mind? The first thing that comes to mind for me is a variant of `BasicDecoration` that assigns a random color to each new decoration.

That sounds bizarre. How about "light mode"?

tarek-y-ismail:
> That sounds bizarre. How about "light mode"?

Hmm, that should do too. But first  things first, let's make `decorations` private. Would `add_decoration`, `override_decoration`, and `num_decorations` be a good interface? I think `BasicManager` is a good base for authors to build atop so I don't want to close it up entirely

AlanGriffiths:
> I think `BasicManager` is a good base for authors to build atop so I don't want to close it up entirely

Here's your first problem then: `basic_manager.h` is internal to mirserver. Authors cannot use it

tarek-y-ismail:
> `basic_manager.h` is internal to mirserver.

Ok, should be easy to fix. I propose `include/common/mir/decoration/` as a location, any better fitting locations?

AlanGriffiths:
> Ok, should be easy to fix. I propose `include/common/mir/decoration/` as a location, what do you think?

I think that's for APIs supported by mircommon

tarek-y-ismail:
What about `include/server/mir/decoration`?

AlanGriffiths:
> What about `include/server/mir/decoration`?

`src/include/server/mir/decoration`

tarek-y-ismail:
I went around and implemented "light mode" in the laziest/easiest way possible (i.e. just paramterizing `BasicDecoration` on the different needed colors) . Nitpicks welcome!

AlanGriffiths (changes requested):
We are trying to provide a way for users to customise decorations themselves.

Instead, `miral::LigtMode` provides an option to select a customisation we wrote, not the capability to write it as a user.

AlanGriffiths (changes requested):
This needs cleaning up:

1. Examples the user can adapt (`UserDecorationManagerExample` _et alia_) do not belong in the `miral` namespace nor library.
2. The example files `user_decoration_example/...` belong in `examples` not `include/miral` and `src/miral`
3. The example code should be used in an example shell
4. There is no need for `miral::decoration::LightMode`, nor for the changes introduced to support it.

tarek-y-ismail:
- [x] Examples the user can adapt (UserDecorationManagerExample et alia) do not belong in the miral namespace nor library.
- [x] The example files user_decoration_example/... belong in examples not include/miral and src/miral
- [x] The example code should be used in an example shell
- [ ] (half done) There is no need for miral::decoration::LightMode, nor for the changes introduced to support it.
  Still need to go through the commits and figure out what needs to be reverted.

tarek-y-ismail:
> Still need to go through the commits and figure out what needs to be reverted.

I think I went through most of the changes and undid them. The only thing I don't think is a good idea to undo is the parameterization of `BasicDecoration` as I see no harm in leaving it as is.

include/core/mir_toolkit/events/enums.h
```diff
@@ -126,6 +126,10 @@ typedef enum {
 /* Axis representing the diameter of a circle centered on the touch
    point */
     mir_touch_axis_size = 5,
+/* Axis representing the local x coordinate for the touch */
+    mir_touch_axis_local_x = 6,
+/* Axis representing the local y coordinate for the touch */
+    mir_touch_axis_local_y = 7,
```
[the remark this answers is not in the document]
AlanGriffiths: Breaking ABI is moot as we don't need these changes.

The "boilerplate" in `InputManager::handle_input_event()` that uses it should be part of the API implementation. The user code need only supply the functions (such as `pointer_event()` & `pointer_leave()`) that can meaningfully differ between implementations.

AlanGriffiths (changes requested):
This MR has accumulated a lot of history and you've learnt a lot. But I think you should start over [Mikado Method](https://understandlegacycode.com/blog/a-process-to-do-safe-changes-in-a-complex-codebase/) style.

I think the first goal (starting from `main`) should be to isolate the policy elements of the decoration code from the generic. As you've seen, pulling all the code out of `libmirserver` introduces difficulties with dependencies on internal APIs. A slimmed down "policy" would help avoid that, and I think you've learnt enough to make that separation work.

src/server/shell/decoration/decoration_notifier.cpp
```diff
@@ -0,0 +1,78 @@
...
+    input_observer{std::make_shared<InputObserver>(decoration)},
+    window_observer{std::make_shared<WindowSurfaceObserver>(decoration)}
+{
+    decoration_surface->register_interest(input_observer);
+    window_surface->register_interest(window_observer);
+}
+
+msd::DecorationNotifier::~DecorationNotifier()
+{
+    if (decoration_surface)
```
tarek-y-ismail, line 73: Superfluous

src/miral/decoration_light_mode.cpp
tarek-y-ismail, line 1: Should be removed

tarek-y-ismail (commented):
Could be squashed into 7dce0d3ccc7a7bc0129bb6677dc456f495cadcae

tarek-y-ismail:
600a57a62f93397b7d0fb0ba81737671733083ba and 519ce5d1180f899be919866664c1392a46bdf2ba could be squashed

src/server/shell/decoration/renderer.cpp
```diff
@@ -464,36 +449,43 @@ auto msd::Renderer::Text::Impl::utf8_to_utf32(std::string const& text) -> std::u
...
+            {ButtonFunction::Close, render_close_icon},
+            {ButtonFunction::Maximize, render_maximize_icon},
+            {ButtonFunction::Minimize, render_minimize_icon},
+        };
+
+        buttons.emplace(button_function, std::make_pair(icon, drawing_functions.at(button_function)));
+    }
+
+    return buttons;
+}
```
tarek-y-ismail, line 472: I see no reason to have the icon, button function, and icon drawing function split. A list of tuples should do?

include/miral/miral/basic_decoration/basic_decoration.h
tarek-y-ismail, line 1: This and its friends should probably be split into their own commit

examples/miral-custom-decorations/user_decoration_manager_example.cpp
```diff
@@ -0,0 +1,210 @@
...
+#include "input.h"
+#include "user_decoration_example.h"
+#include "user_decoration_manager_example.h"
+#include "miral/decoration/decoration_adapter.h"
+
+#include <mir/shell/decoration.h>
+#include <mir/graphics/display_configuration.h>
+#include <mir/graphics/null_display_configuration_observer.h>
+/* #include <bffoost/throw_exception.hpp> */
+#include "mirserver-internal/mir/observer_registrar.h"
```
tarek-y-ismail, line 26: Is this okay, or should observer_registrar.h be exposed / I should access it some other way?

include/miral/miral/decoration/decoration_adapter.h
```diff
@@ -0,0 +1,79 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIRAL_DECORATION_DECORATION_ADAPTER_H
+#define MIRAL_DECORATION_DECORATION_ADAPTER_H
+
+#include "mir/shell/decoration_notifier.h"
```
AlanGriffiths, line 20: This should not be used by a public libmiral header.

include/miral/miral/decoration/decoration_adapter.h
```diff
@@ -0,0 +1,79 @@
...
+
+#include <functional>
+#include <memory>
+
+
+namespace miral::decoration
```
AlanGriffiths, line 26: I doubt we need a nested `decoration` namespace

include/miral/miral/decoration/decoration_input_resolver_adapter.h
```diff
@@ -0,0 +1,70 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIRAL_DECORATION_DECORATION_INPUT_MANAGER_ADAPTER_H
+#define MIRAL_DECORATION_DECORATION_INPUT_MANAGER_ADAPTER_H
+
+#include "mir/shell/input_resolver.h"
```
AlanGriffiths, line 20: This should not be used by a public libmiral header.

include/miral/miral/decoration/decoration_input_resolver_adapter.h
```diff
@@ -0,0 +1,70 @@
...
+    friend InputResolverBuilder;
+    InputResolverAdapter();
+
+    std::function<void(DeviceEvent& device)> on_process_enter;
+    std::function<void()> on_process_leave;
+    std::function<void()> on_process_down;
+    std::function<void()> on_process_up;
+    std::function<void(DeviceEvent& device)> on_process_move;
+    std::function<void(DeviceEvent& device)> on_process_drag;
+};
```
AlanGriffiths, line 48: The Builder exists because this class doesn't belong in a public header. Users of libmiral should be isolated from such details.

include/miral/miral/decoration/decoration_adapter.h
```diff
@@ -0,0 +1,79 @@
...
+
+    std::function<void(std::shared_ptr<MirEvent const> const&)> on_handle_input_event;
+    std::function<void(float new_scale)> on_set_scale;
+    std::function<void(mir::scene::Surface const* window_surface, MirWindowAttrib attrib, int value)> on_attrib_changed;
+    std::function<void(mir::scene::Surface const* window_surface, mir::geometry::Size const& window_size)>
+        on_window_resized_to;
+    std::function<void(mir::scene::Surface const* window_surface, std::string const& name)> on_window_renamed;
+
+    mir::shell::decoration::DecorationNotifier decoration_notifier;
+};
```
AlanGriffiths, line 53: The Builder exists because this class doesn't belong in a public header. Users of libmiral should be isolated from such details.

examples/CMakeLists.txt
```diff
@@ -5,5 +5,6 @@ add_subdirectory(miral-kiosk)
 add_subdirectory(miral-system-compositor)
 add_subdirectory(mir_demo_server)
 add_subdirectory(mir-x11-kiosk)
+add_subdirectory(miral-custom-decorations)
```
AlanGriffiths, line 8: I don't think custom decorations need their own example. Adding them as a miral-shell option would be fine

examples/miral-custom-decorations/miral_custom_decorations.cpp
```diff
@@ -0,0 +1,195 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+#include "mir/server.h"
+#include "mir/main_loop.h" // Needed for the conversion of `the_main_loop` to an executor
```
AlanGriffiths, line 19: If this is needed, then something is wrong with the proposed API

examples/miral-custom-decorations/miral_custom_decorations.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+};
+
+struct CustomDecorations
+{
+    CustomDecorations() = default;
+
+    void operator()(mir::Server& s) const
+    {
+        s.add_pre_init_callback(
```
AlanGriffiths, line 71: User code should not be accessing the Server in this way (it isn't a public header)

examples/miral-custom-decorations/miral_custom_decorations.cpp
```diff
@@ -0,0 +1,195 @@
...
+            "terminal emulator to use",
+            terminal_cmd},
+        /* [&external_client_launcher](mir::Server& s) */
+        /* { */
+        /*     s.add_init_callback( */
+        /*         [&external_client_launcher] */
+        /*         { */
+        /*             external_client_launcher.launch("xeyes"); */
+        /*         }); */
+        /* }, */
```
AlanGriffiths, line 193:
```suggestion
```

AlanGriffiths (changes requested):
This iteration of the API is leaking implementation details

AlanGriffiths:
> I think the first goal (starting from `main`) should be to isolate the policy elements of the decoration code from the generic. As you've seen, pulling all the code out of `libmirserver` introduces difficulties with dependencies on internal APIs. A slimmed down "policy" would help avoid that, and I think you've learnt enough to make that separation work.

include/miral/miral/decoration/decoration_adapter.h
tarek-y-ismail: So, like, create a `miral::*` version of the class, with the header file completely isolating the internal symbols from the user, and the source file using/including the header? 

Does this also apply to similar cases in this review? (`input_resolver.h`, adapters, etc...)?

AlanGriffiths: No, it should just be internal and unavailable to the shell author

examples/miral-custom-decorations/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ include_directories(
   ${PROJECT_SOURCE_DIR}/include/miral
   ${PROJECT_SOURCE_DIR}/src/include/server
   ${PROJECT_SOURCE_DIR}/include/platform
+  ${PROJECT_SOURCE_DIR}/include/wayland
```
AlanGriffiths, line 13: Example code should not depend on any of these include paths

PR #3607, opened 2024-09-23
debian: package atomic-kms

Saviq:


debian/control
```diff
@@ -404,6 +418,7 @@ Architecture: linux-any
 Multi-Arch: same
 Pre-Depends: ${misc:Pre-Depends}
 Depends: ${misc:Depends},
+         mir-platform-graphics-atomic-kms,
```
RAOF, line 421: Hm. But you haven't actually *added* a `mir-platform-graphics-atomic-kms` package, just a `mir-platform-graphics-atomic-kms22` package.

Maybe what we should do is add `mir-platform-graphics-atomic-kms22` to the dependencies of the `mir-platform-graphics-gbm-kms` metapackage?

RAOF (changes requested):

debian/control
Saviq: And then we'll have a `Provides:` / `Breaks:` / `Replaces:`, when we want to move over?

We should probably add `-atomic-kms` already anyway.

Saviq: > We should probably add `-atomic-kms` already anyway.

Done.

RAOF (changes requested):
>  Unknown package: debian/control contains versioned package mir-platform-graphics-atomic-kms but it is unknown to this script

We'll need to fix this before it can land :wink:

debian/control
```diff
@@ -416,6 +431,22 @@ Description: Display server for Ubuntu - desktop driver metapackage
  This package depends on a full set of graphics and input drivers for traditional desktop
  systems.
 
+Package: mir-platform-graphics-atomic-kms
+Section: libs
+Architecture: linux-any
+Multi-Arch: same
+Pre-Depends: ${misc:Pre-Depends}
+Depends: ${misc:Depends},
```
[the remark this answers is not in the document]
Saviq: Or split the rendering platform out?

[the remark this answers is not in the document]
Saviq: I actually added a `Depends` between in `atomic-kms22` on `gbm-kms22`, going through the metapackage felt wrong.

AlanGriffiths (approved):
Yeah, looks fine now

PR #3609, opened 2024-09-24
Revert "Fix random leak"

AlanGriffiths:
Fixes x11-kiosk crash when running on `mir:wayland`

(It looks like any use if the `mir:wayland` platform would fail)


AlanGriffiths:
@Saviq this is one to cherry-pick into 2.18.1

Issue #3610, opened 2024-09-25
Smoke tests are failing


AlanGriffiths:
```
$ mir-smoke-test-runner
...
I: The following clients failed to execute successfully:
I:     eglinfo
I:     /usr/bin/mir_demo_client_wayland
I:     /usr/bin/mir_demo_client_wayland_egl_spinner
I:     /usr/bin/mir_demo_client_mir_shell
I: Smoke testing complete with returncode -1
```

Reproduced with 2.18.0 and 2.18.1 (release and rc PPAs), but not 2.14 (24.04 archive)

AlanGriffiths:
This is a shutdown problem in mir_demo_server. (Apparently in `TouchspotController::~TouchspotController()`, but that may be spurious.)

PR #3611, opened 2024-09-25
Do not extend the `InputDeviceHub` lifetime beyond teardown

AlanGriffiths:
Fixes: #3610

### DIFF
--- a/src/miral/input_configuration.cpp
+++ b/src/miral/input_configuration.cpp
@@ -38,7 +38,7 @@ class miral::InputConfiguration::Touchpad::Self : public TouchpadInputConfigurat
 class miral::InputConfiguration::Self
 {
 public:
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
@Saviq another cherry-pick for 2.18

It only affects shells that use `miral::InputConfiguration` (which will probably remain rare, but they will crash on exit)

tarek-y-ismail:
Can confirm that this fixes the smoke tests

tarek-y-ismail (approved):

AlanGriffiths:
> Can confirm that this fixes the smoke tests

You should also confirm it doesn't break the demo of input configuration changes. (I.e. changes written to `~/.config/mir_demo_server.input` are applied up by `mir_demo_server`)

(Works on my machine, but...

PR #3614, opened 2024-09-25
feature: users can specify an app id for the wayland surface when running on the wayland platform

mattkae:
## What's new?
- Added the `--wayland-surface-app-id=<APP_ID>` option for the wayland platform that, when set, will call [xdg_top_level::set_app_id](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id) with the value `APP_ID`
- Added the `--wayland-surface-title=<TITLE>` option for the wayland platform that, when set, will call [xdg_top_level::set_title](https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_title) with the value `TITLE`

### DIFF
--- a/src/platforms/wayland/platform_symbols.cpp
+++ b/src/platforms/wayland/platform_symbols.cpp
@@ -60,7 +60,7 @@ mir::UniqueModulePtr<mg::DisplayPlatform> create_display_platform(
 
     std::optional<std::string> title;
     if (options->is_set(wayland_surface_title_option))
-        app_id = options->get<std::string>(wayland_surface_title_option);
+        title = options->get<std::string>(wayland_surface_title_option);
 
     mir::assert_entry_point_signature<mg::CreateDisplayPlatform>(&create_display_platform);
     return mir::make_module_ptr<mgw::Platform>(mpw::connection(*options), report, app_id, title);
--- a/src/platforms/wayland/platform_symbols.cpp
+++ b/src/platforms/wayland/platform_symbols.cpp
@@ -40,8 +40,11 @@ mir::ModuleProperties const description = {
     mir::libname()
 };
 
-char const* wayland_surface_app_id_option{"wayland-surface-app-id"};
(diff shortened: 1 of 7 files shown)


mattkae:
> Should this be `--wayland-surface-app-id` since that's the Wayland nomenclature?
> 
> Should we go for `--wayland-surface-title` as well?

We don't have a `wayland-window-title` currently

src/platforms/wayland/platform_symbols.cpp
```diff
@@ -39,6 +39,9 @@ mir::ModuleProperties const description = {
     MIR_VERSION_MICRO,
     mir::libname()
 };
+
+char const* wayland_surface_app_id_option{"wayland-surface-app-id"};
+char const* wayland_surface_app_id_option_description{"Defines the XdgToplevel app id on the surface created by the wayland platform"};
```
AlanGriffiths, line 44: These can be const

Saviq:
> We don't have a `wayland-window-title` currently

That's what I mean - should we add it, in one go?

src/platforms/wayland/platform_symbols.cpp
```diff
@@ -48,14 +54,30 @@ mir::UniqueModulePtr<mg::DisplayPlatform> create_display_platform(
     std::shared_ptr<mir::ConsoleServices> const& /*console*/,
     std::shared_ptr<mg::DisplayReport> const& report)
 {
+    std::optional<std::string> app_id;
+    if (options->is_set(wayland_surface_app_id_option))
+        app_id = options->get<std::string>(wayland_surface_app_id_option);
+
+    std::optional<std::string> title;
+    if (options->is_set(wayland_surface_title_option))
+        app_id = options->get<std::string>(wayland_surface_title_option);
```
Saviq, line 63:
```suggestion
        title = options->get<std::string>(wayland_surface_title_option);
```

Saviq (approved):
Legit!

PR #3615, opened 2024-09-25
Atomic kms hardware cursor support

tarek-y-ismail:
Now with 100% more atomic API!

Still a bit broken:
1. If we don't commit on cursor move, the cursor becomes very choppy
2. If we attempt a blocking commit on cursor move, but use , cursor and application window movement become sluggish when dragging a window around.
3. If we attempt a nonblocking commit on cursor move, the kernel/hardware/?? can't keep up and complains a bunch (works best though!).
4. Crashes sometimes (might be related to the input thread calling `set_cursor`, attempting a commit, and the main thread attempting a commit at the same time?)
5. Cursor sometimes disappears when entering or leaving application windows (actual content, drop shadow doesn't affect it)

How to test:
1. Log out of your graphical session
2. Switch to a VT
3. Make sure hardware cursors initialize correctly, this can be done by  running any Mir example with the argument `--platform-display-libs=mir:atomic-kms`  and piping the output to `grep -i 'hardware cursor'`
4. If the hardware cursor initializes correctly, you can go on using your compositor as normal.

### DIFF
--- a/src/platforms/atomic-kms/server/kms/CMakeLists.txt
+++ b/src/platforms/atomic-kms/server/kms/CMakeLists.txt
@@ -22,6 +22,7 @@ add_library(
   atomic_kms_output.cpp
   atomic_kms_output.h
   bypass.cpp
+  cursor.cpp
   display.cpp
   display_buffer.cpp
   platform.cpp
--- a/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
+++ b/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
@@ -483,9 +483,9 @@ void mga::AtomicKMSOutput::wait_for_page_flip()
     pending_page_flip.get();
 }
 
-std::shared_ptr<uint32_t> mga::AtomicKMSOutput::cursor_gbm_bo_to_drm_fb_id(gbm_bo* gbm_bo)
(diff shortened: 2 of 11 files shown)


tarek-y-ismail:
First three points are now addressed by attempting a non-blocking commit and not logging if the hardware is busy. 
Fourth point fixed sometime while I was fixing the first three, not sure exactly what fixed it.
Fifth point addressed by just not clearing the cursor. This is probably wrong but I don't quite understand why it behaves weirdly inside application surfaces. It seems that applications only set/clear it based on their own updates regardless of the cursor moving.

tarek-y-ismail:
> Fifth point addressed by just not clearing the cursor. This is probably wrong but I don't quite understand why it behaves weirdly inside application surfaces. It seems that applications only set/clear it based on their own updates regardless of the cursor moving.

Turns out that this accidentally disabled hardware cursors, hence why things seemed to work correctly. Adjusting `clear_cursor` to just return `true` still results in the cursor being hidden when entering/leaving application surfaces.

tarek-y-ismail:
So the crashes were potentially caused by multiple threads modifying `current_crtc`. Synchronizing render/input threads with a mutex seems to alleviate the issue.

AlanGriffiths:
Converting to "draft" until discussions and conflicts are resolved

PR #3616, opened 2024-09-26
examples/miral-shell: Use generic logos for example applications

Conan-Kudo:
These are some self-created simple crafted logos for miral-shell to replace the usage of the Ubuntu logo to make Mir examples easier to ship.


Saviq:
Replaced by #3617

PR #3617, opened 2024-09-27
examples/miral-shell: Use generic logos for example applications

Saviq:
These are some self-created simple crafted logos for miral-shell to replace the usage of the Ubuntu logo to make Mir examples easier to ship.

Thanks @Conan-Kudo!


mattkae (approved):
My only feedback is that maybe the glow is too subtle, but this works as advertised for me!

PR #3619, opened 2024-10-03
Ignore X11 attempts to place window

AlanGriffiths:


Saviq (approved):
Yup, can confirm the fix!

PR #3622, opened 2024-10-07
[xwayland] Don't try to place X11 windows

AlanGriffiths:
X11 CONFIGURE events may specify a position, but we will ignore it


mattkae (approved):
Nice !

PR #3624, opened 2024-10-07
Release 2.18.3

Saviq:
## Enhancements:
- Ensure windows are placed and repainted on X11 "CONFIGURE" #3619


AlanGriffiths (approved):
Yup. Don't think we're missing anything else

PR #3634, opened 2024-10-14
Performance

AlanGriffiths:


doc/sphinx/explanation/performance.md
```diff
@@ -0,0 +1,59 @@
...
+possible to defeat the available performance in such products.
+
+## Memory usage
+
+Memory usage varies a lot depending on: 
+
+* the GPU characteristics of the system
+* the number, size and configuration of the displays
+* the number of application windows being composited
+* the rate at which buffers are submitted to the application windows
```
tarek-y-ismail, line 17: This block can probably be moved to the top of the document.

tarek-y-ismail:
Some things that I had in mind that could be added (just a rough draft), feel free to use/ignore/edit them as you see  fit.

Mir splits into multiple threads at some point. One for rendering/compositing, one for input, one for XWayland events, one for Wayland events, etc... (as far as I understand). Focusing on the compositing and input threads this improves performance (response time) by letting the two threads work independently. For example, the compositing thread might sleep until an application or another thing triggers an update. The input thread on the other hand, sleeps until an event arrives on the socket it listens to, and immediately calls the appropriate handle. This results in updates being delivered as soon as they are requested, and inputs being handled immediately. 

In the case where two threads might overlap, critical sections are kept as small as possible with attention given to reduce locking overhead and time.

---

Almost all drawing/compositing (except for built-in decorations) is done via GPU if available to reduce CPU load and take advantage of GPUs' better efficiency and speed.

doc/sphinx/explanation/performance.md
AlanGriffiths: Only the last three lines of the block are common

AlanGriffiths:
> Mir splits into multiple threads at some point

The performance trade-offs of threading is a vast, contentious, subject and a level of detail I don't think appropriate.

doc/sphinx/explanation/performance.md
Saviq: I agree it would read better if the common lines here were abstracted up.

doc/sphinx/explanation/performance.md
```diff
@@ -0,0 +1,59 @@
...
+
+The range being from around 200Mb for a simple compositor (Frame) running on a
+single display with a single application to 7Gb for a "desktop capable"
```
Saviq, line 20:
```suggestion
The range being from around 200MB for a simple compositor (Frame) running on a
single display with a single application to 7GB for a "desktop capable" 
```

doc/sphinx/explanation/performance.md
```diff
@@ -0,0 +1,59 @@
...
+
+Mir normally has no difficulty keeping up with applications but frame rate can
+vary a lot depending on:
+
+* the CPU, GPU and memory characteristics of the system
+* the number, size and configuration of the displays
+* the number of application windows being composited
+* the rate at which buffers are submitted to the application windows
+
+Part of our test automation is to some test applications (e.g.
```
Saviq, line 56: is to… test?

Saviq (changes requested):
Please add it to explanation.md and to the TOC.

I would also like to see a blurb about using GPU acceleration, where possible.

We mentioned display planes / overlays as another theme that could be discussed?

Saviq:
I wouldn't be averse to merging Performance and Energy efficiency into one page, as there's only so much that is different between them…

AlanGriffiths:
> Please add it to explanation.md and to the TOC.

I presume that's `explanation/index.md`?

AlanGriffiths:
> I wouldn't be averse to merging Performance and Energy efficiency into one page, as there's only so much that is different between them…

Strange, I don't see there being much in common between them :man_shrugging:

Saviq (approved):
LGTM!

PR #3635, opened 2024-10-16
Ignore invalid display config files and boot normally.

### LINKED ISSUE #3628: StaticDisplayConfig: `unrecognized content` prevents Mir starting up
When the display file is not a valid yaml, or does not have the right contents, we should still start up with the default configs.

Not rewrite the file, like we do when empty - but still start up.

tarek-y-ismail:
Closes #3628 

Adds a new exception type `StaticDsiplayConfig::InvalidConfig`, which is now used inside `YamlFileDisplayConfig::load_config` to signal a failure in parsing. When this exception is caught, we skip the rest of config parsing and the remaining code that depends on it and run as if `--display-config=static=path/to/invalid/file.display` was not provided.

### DIFF
--- a/src/miral/display_configuration_option.cpp
+++ b/src/miral/display_configuration_option.cpp
@@ -225,9 +225,21 @@ void miral::display_configuration_options(mir::Server& server)
                 {
                     mir::fatal_error("Display scale option can't be used with static display configuration");
                 }
-                auto sdc = std::make_shared<StaticDisplayConfig>(display_layout.substr(strlen(static_opt_val)));
-                server.add_init_callback([sdc, &server]{ sdc->init_auto_reload(server); });
-                layout_selector = std::move(sdc);
+
+                try
+                {
+                    auto sdc = std::make_shared<StaticDisplayConfig>(display_layout.substr(strlen(static_opt_val)));
+                    server.add_init_callback(
+                        [sdc, &server]
+                        {
+                            sdc->init_auto_reload(server);
+                        });
+                    layout_selector = std::move(sdc);
+                }
+                catch (miral::StaticDisplayConfig::InvalidConfig& e)
+                {
+                    mir::log_error("%s. Using defaults", e.what());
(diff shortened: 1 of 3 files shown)


Saviq (changes requested):
Let's make the new behaviour optional (off by default)?

Keep the current constructor so we don't break ABI, but add a new one with a `bool exit_on_error=true`?

This way, Frame we can make to start up regardless of display config issues.

Also, in `StaticDisplayConfig` we could then log an error for the less critical bits and ignore / use defaults for that particular element rather than bail completely.

src/miral/display_configuration_option.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "miral/display_configuration_option.h"
+#include "mir/abnormal_exit.h"
```
AlanGriffiths, line 18: Why is this added when use of `AbnormalExit` is being removed?

AlanGriffiths (changes requested):
Just a note that I'm against the resolution suggested in the issue. (This implements that resolution: I'm not against the implementation)

Saviq:
As discussed on the bug, maybe having a special case (and that could be default) for the file being _empty_ would also do.

AlanGriffiths:
> Let's make the new behaviour optional (off by default)?

+1

> Keep the current constructor so we don't break ABI, but add a new one with a `bool exit_on_error=true`?

Or use the "named constructor" idiom: `StaticDisplayConfig:on_errors_use_default()` is clearer than `StaticDisplayConfig{true}`/`StaticDisplayConfig{false}`

> This way, Frame we can make to start up regardless of display config issues.

I'm still not convinced this is desirable

AlanGriffiths:
Also, we should have tests

tarek-y-ismail:
> Also, we should have tests

Lucky for me, we already do (the ones failing right now :)). I'll see if I can add any.

Issue #2595, opened 2022-08-28
Switch focus on URL open


Saviq:
This is probably a contentious topic while the compositor doesn't know about an app's request to open another (e.g. calling `xdg-open`), as it's working against #2586. But today you click a link in one app, it opens in another, and you have no idea.

I know we had designs for this in mirclient, and these likely need to migrate into the new world.

RAOF:
This is actually a request to implement the [xdg-activation](https://wayland.app/protocols/xdg-activation-v1) protocol that we contributed input to.

RAOF:
(Plus waiting for clients to actually *use* that protocol)

Saviq:
Fixed by #3639

Issue #2586, opened 2022-08-25
Implement focus stealing prevention


Saviq:
I'm running Thunderbird under Miriway (same issue with notifications from websites in Firefox). The notification surfaces (part of the problem is lack of a notification manager) show in random locations on the screen, incorrectly cropped and steal focus.

There may be nothing that Mir can do visually, if those are simply top-level surfaces - but focus stealing prevention is something we need to do. May need a separate issue?

Using this [Miriway](https://github.com/Miriway/Miriway) config:

```ini
x11-window-title=Miriway
idle-timeout=180
ctrl-alt=t:miriway-terminal # Default "terminal emulator finder"

enable-x11=true

shell-component=swaybg -i /usr/share/backgrounds/warty-final-ubuntu.png
shell-component=mate-panel
meta=a:mate-panel --run-dialog
ctrl-alt=t:mate-terminal
```

wmww:
We *might* be able to do something to prevent focus stealing, but fwiw the problem exists in sway as well. Easiest solution is probably to run mate-notification-daemon, which I ported to layer shell a while back.

Saviq:
> Easiest solution is probably to run mate-notification-daemon, which I ported to layer shell a while back.

Looks like it wants to be DBus-activated, as it silently exits a few seconds after starting...

Saviq:
> Looks like it wants to be DBus-activated, as it silently exits a few seconds after starting...

Hmm, it even _is_ DBus-activated I think. Still shows up in the middle of the screen, though.
![frame_2022-08-26T10:51+02:00](https://user-images.githubusercontent.com/231077/186867741-c87dc6bf-0072-4da2-a788-967a841fde76.png)

AlanGriffiths:
> > Looks like it wants to be DBus-activated, as it silently exits a few seconds after starting...
> 
> Hmm, it even _is_ DBus-activated I think. Still shows up in the middle of the screen, though. ![frame_2022-08-26T10:51+02:00](https://user-images.githubusercontent.com/231077/186867741-c87dc6bf-0072-4da2-a788-967a841fde76.png)

It won't have access to `zwlr_layer_shell` unless started from `miriway-shell`, so try enabling `add-wayland-extensions=all` and see if that works

Saviq:
> It won't have access to `zwlr_layer_shell` unless started from `miriway-shell`, so try enabling `add-wayland-extensions=all` and see if that works

Bingo!

I've modified the bug to talk about focus stealing prevention only.

[...]

PR #3639, opened 2024-10-16
feature: implementation of xdg_activation_v1 because I want my auto focuses

mattkae:
[Screencast from 2024-10-16 16-21-05.webm](https://github.com/user-attachments/assets/86f9b0bf-181c-4aab-acfd-4ceb1e4e2c38)

## What's new?
- Added the xdg_activation_v1 protocol
- Implemented the xdg_activation_v1 protocol

## To test
- Run `miral-shell`
- Open `gnome-terminal`
- Run `gedit`
- Focus `gnome-terminal` again
- Run `gedit` again
- Note that `gedit` gets focus :magic_wand: 

Similarly:
- Open up `emacs`
- Open up `firefox`
- In emacs, write `https://google.com`
- Ctrl + Click the link
- Watch as firefox gets focus!

## Open questions
- The "timeout" is not configurable at the moment, but perhaps it should be. However, I don't think we do per-protocol options AFAICT

## Todos
- I should probably write some WLCS tests!

### DIFF
--- a/src/server/frontend_wayland/xdg_activation_v1.cpp
+++ b/src/server/frontend_wayland/xdg_activation_v1.cpp
@@ -1,0 +1,312 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "xdg_activation_v1.h"
+#include "wl_surface.h"
+#include "wl_seat.h"
+#include "mir/main_loop.h"
+#include "mir/scene/surface.h"
+#include "mir/shell/shell.h"
+#include "mir/shell/focus_controller.h"
+#include "mir/log.h"
+#include <random>
+#include <chrono>
+
+namespace mf = mir::frontend;
+namespace mw = mir::wayland;
+namespace msh = mir::shell;
+
+namespace mir::frontend
+{
+/// Time in milliseconds that the compositor will wait before invalidating a token
+auto constexpr TIMEOUT_MS = std::chrono::milliseconds(3000);
+
+std::string generate_token()
+{
+    std::string str("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz");
+    std::random_device rd;
+    std::mt19937 generator(rd());
+    std::shuffle(str.begin(), str.end(), generator);
+    return str.substr(0, 32);
+}
+
+class XdgActivationV1 : public wayland::XdgActivationV1::Global
+{
+public:
+    XdgActivationV1(
+        struct wl_display* display,
+        std::shared_ptr<msh::Shell> const& shell,
+        std::shared_ptr<msh::FocusController> const& focus_controller,
+        std::shared_ptr<MainLoop> const& main_loop);
+
+    std::string const& create_token();
+    bool try_consume_token(std::string const& token);
+
+private:
+    class Instance : public wayland::XdgActivationV1
+    {
+    public:
+        Instance(
+            struct wl_resource* resource,
+            mf::XdgActivationV1* xdg_activation_v1,
+            std::shared_ptr<msh::Shell> const& shell,
+            std::shared_ptr<msh::FocusController> const& focus_controller,
+            std::shared_ptr<MainLoop> const& main_loop);
+
+    private:
+        void get_activation_token(struct wl_resource* id) override;
+        void activate(std::string const& token, struct wl_resource* surface) override;
+
+        mf::XdgActivationV1* xdg_activation_v1;
+        std::shared_ptr<msh::Shell> shell;
+        std::shared_ptr<msh::FocusController> const focus_controller;
+        std::shared_ptr<MainLoop> main_loop;
+    };
+
+    struct Token
+    {
+        std::string token;
+        std::unique_ptr<time::Alarm> alarm;
+    };
+
+    void bind(wl_resource* resource) override;
+
+    std::shared_ptr<msh::Shell> shell;
+    std::shared_ptr<msh::FocusController> const focus_controller;
+    std::shared_ptr<MainLoop> main_loop;
+    std::vector<Token> pending_tokens;
+    std::mutex lock;
+};
+
+class XdgActivationTokenV1 : public wayland::XdgActivationTokenV1
+{
+public:
+    XdgActivationTokenV1(
+        struct wl_resource* resource,
+        std::string const& token,
+        std::shared_ptr<msh::FocusController> const& focus_controller);
+
+private:
+    void set_serial(uint32_t serial, struct wl_resource* seat) override;
+    void set_app_id(std::string const& app_id) override;
+    void set_surface(struct wl_resource* surface) override;
+    void commit() override;
+
+    std::string token;
+    std::shared_ptr<msh::FocusController> const focus_controller;
+    std::optional<uint32_t> serial;
+    std::optional<struct wl_resource*> seat;
+    std::optional<std::string> app_id;
+    std::optional<struct wl_resource*> requesting_surface;
+};
+}
+
+auto mf::create_xdg_activation_v1(
+    struct wl_display* display,
+    std::shared_ptr<msh::Shell> const& shell,
+    std::shared_ptr<msh::FocusController> const& focus_controller,
+    std::shared_ptr<MainLoop> const& main_loop)
+    -> std::shared_ptr<mw::XdgActivationV1::Global>
+{
+    return std::make_shared<mf::XdgActivationV1>(display, shell, focus_controller, main_loop);
+}
+
+mf::XdgActivationV1::XdgActivationV1(
+    struct wl_display* display,
+    std::shared_ptr<msh::Shell> const& shell,
+    std::shared_ptr<msh::FocusController> const& focus_controller,
+    std::shared_ptr<MainLoop> const& main_loop)
+    : Global(display, Version<1>()),
+      shell{shell},
+      focus_controller{focus_controller},
+      main_loop{main_loop}
+{
+
+}
+
+std::string const& mf::XdgActivationV1::create_token()
+{
+    auto generated = generate_token();
+    Token token{generated, main_loop->create_alarm([this, generated]()
+    {
+        try_consume_token(generated);
+    })};
+    token.alarm->reschedule_in(TIMEOUT_MS);
+
+    std::lock_guard guard(lock);
+    pending_tokens.emplace_back(std::move(token));
+    return pending_tokens[pending_tokens.size() - 1].token;
+}
+
+bool mf::XdgActivationV1::try_consume_token(std::string const& token)
+{
+    std::lock_guard guard(lock);
+    for (auto it = pending_tokens.begin(); it != pending_tokens.end(); it++)
+    {
+        if (it->token == token)
+        {
+            pending_tokens.erase(it);
+            return true;
+        }
+    }
+
+    return false;
+}
+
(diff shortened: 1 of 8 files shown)


AlanGriffiths:
> * I don't currently have a way to check if a `serial` is valid, so that is left as a TODO. If anyone knows, please direct me in the right spot!

If you look around `WindowWlSurfaceRole::input_event_for()` you'll see existing code checks

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -0,0 +1,312 @@
...
+#include "mir/log.h"
+#include <random>
+#include <chrono>
+
+namespace mf = mir::frontend;
+namespace mw = mir::wayland;
+namespace msh = mir::shell;
+
+namespace mir::frontend
+{
```
AlanGriffiths, line 33: Not the project style

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -0,0 +1,312 @@
...
+auto constexpr TIMEOUT_MS = std::chrono::milliseconds(3000);
+
+std::string generate_token()
+{
+    std::string str("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz");
+    std::random_device rd;
+    std::mt19937 generator(rd());
+    std::shuffle(str.begin(), str.end(), generator);
+    return str.substr(0, 32);
+}
```
AlanGriffiths, line 44: This looks somewhat suspect (limited range of characters, each character can only occur once).

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -0,0 +1,312 @@
...
+std::string const& mf::XdgActivationV1::create_token()
+{
+    auto generated = generate_token();
+    Token token{generated, main_loop->create_alarm([this, generated]()
+    {
+        try_consume_token(generated);
+    })};
+    token.alarm->reschedule_in(TIMEOUT_MS);
+
+    std::lock_guard guard(lock);
```
AlanGriffiths, line 149: It is strange seeing a lock here - shouldn't all this code run on the Wayland thread?

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -0,0 +1,312 @@
...
+    pending_tokens.emplace_back(std::move(token));
+    return pending_tokens[pending_tokens.size() - 1].token;
```
AlanGriffiths, line 151:
```suggestion
    return pending_tokens.back().token;
```

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+void mf::XdgActivationTokenV1::set_surface(struct wl_resource* surface)
+{
+    requesting_surface = surface;
+}
+
+void mf::XdgActivationTokenV1::commit()
+{
+    // In the event of a failure, we send 'done' with a new, invalid token
+    // which cannot be used later.
```
AlanGriffiths, line 253: That doesn't accord with my reading of the protocol:

```xml
        <event name="done">
            <description summary="the exported activation token">
                The 'done' event contains the unique token of this activation request
                and notifies that the provider is done.
            </description>
            <arg name="token" type="string" summary="the exported activation token"/>
        </event>
```
Is there prior art for sending invalid tokens?

AlanGriffiths (changes requested):
Interesting

src/server/frontend_wayland/xdg_activation_v1.cpp
mattkae: I adapted this from wlroots, who may or may not be doing a good thing. https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_xdg_activation_v1.c?ref_type=heads#L91 The error case always sends an invalid token string

src/server/frontend_wayland/xdg_activation_v1.cpp
mattkae: The lock is because we're also setting an `Alarm` for timeout on the token. That would be on the main thread, no?

src/server/frontend_wayland/xdg_activation_v1.cpp
mattkae: Yeah this is a good candidate to use `uuid`

src/server/frontend_wayland/xdg_activation_v1.cpp
AlanGriffiths: I guess it is a case of the protocol not having "nack" event, so there's no obvious way to "reject" the request. (Making token nullable would be enough)

AlanGriffiths (changes requested):
More nits. Main thing left is the serial validation TODO

Saviq:
Ref.:
- https://github.com/canonical/mir/issues/2595
- https://github.com/canonical/mir/issues/2586

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -273,8 +285,13 @@ void mf::XdgActivationTokenV1::commit()
             return;
         }
 
-        // TODO:
-        // Next, verify that the serial belongs to the seat
+        auto event = client->event_for(serial.value());
+        if (!event)
+        {
+            mir::log_error("XdgActivationTokenV1::commit serial not found");
```
AlanGriffiths, line 296: I think this is at most a "warning" and possibly an "info": this doesn't impact the running of the server, just indicates the client is misbehaving. (If possible, it might also be helpful to indicate which client)

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -0,0 +1,343 @@
...
+        std::unique_ptr<time::Alarm> alarm;
+    };
+
+    void bind(wl_resource* resource) override;
+
+    std::shared_ptr<msh::Shell> shell;
+    std::shared_ptr<msh::FocusController> const focus_controller;
+    std::shared_ptr<MainLoop> main_loop;
+    std::vector<Token> pending_tokens;
+    std::mutex mutex;
```
AlanGriffiths, line 99: Perhaps things could be clearer:
```suggestion
    // pending_tokens is accessed by an alarm callback, not only the wayland thread
    std::mutex pending_tokens_mutex;
    std::vector<Token> pending_tokens;
```

AlanGriffiths (changes requested):
A couple more tweaks, but looking good

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -309,10 +326,9 @@ void mf::XdgActivationTokenV1::commit()
         }
 
         auto const& scene_surface = scene_surface_opt.value();
-
         if (focus_controller->focused_surface() != scene_surface)
         {
-            mir::log_error("XdgActivationTokenV1::commit the focused surface is not the surface that requesteda ctivation");
+            mir::log_error("XdgActivationTokenV1::commit the focused surface is not the surface that requested activation");
```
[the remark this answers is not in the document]
mattkae: The more I think about this, the more I wonder if a correct behavior would be to check this when we receive the `activate` request. E.g.: Surface A commits its info and gets a token, then it passes the token to Surface B to activate. If Surface A is no longer focused (because the user selected a different surface in the time that it took for Surface B to come up), then we don't activate Surface B because it could be jarring for the user. WDYT?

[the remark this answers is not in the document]
mattkae: Although maybe they should be warnings instead of errors. But that would be a more appropriate time to them IMO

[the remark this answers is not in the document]
mattkae: Yeah unfortunately there doesn't exist an error code for that, which I think is a bit silly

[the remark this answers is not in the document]
Saviq: > If Surface A is no longer focused

It's not about not being focused (because, in theory, you could've switched to C and back to A) - it's about whether there was a focus _change_ between issuing the token and receiving the activation request.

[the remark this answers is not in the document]
Saviq: I believe it's really that simple:

1. issue the token
1. invalidate the token on key/button down or focus change (say A went away)
1. respect the activation request if token is valid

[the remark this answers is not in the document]
mattkae: I think instead of disrespecting when _any_ other surface is focused is a bad idea. The protocol allows users to specify _any_ surface or even an app_id for a number of surfaces. Because of this, it may be valid for clients to:
1. Set the "requesting surface" to Surface B
2. Request the token on Surface A
3. Focus Surface B
4. Create and Activate Surface C (expecting Surface B to be focused)

It is contrived, but by sending up the surface over the protocol, I would only expect that particular surface to matter.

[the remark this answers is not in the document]
mattkae: I agree about the key event thing though :)

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -0,0 +1,391 @@
...
+            {
+                mir::log_error("XdgActivationTokenV1::commit the focused surface is not the surface that requested activation");
+                return;
+            }
+        }
+
+        auto const& scene_surface = scene_surface_opt.value();
+        auto const now = std::chrono::steady_clock::now().time_since_epoch();
+        auto ns = std::chrono::duration_cast<std::chrono::nanoseconds >(now).count();
+        shell->raise_surface(wl_surface->session, scene_surface, ns);
```
AlanGriffiths, line 349: I think this should be a request _via_ the window manager, not bypassing the window manager.

In particular, the WM may want to switch workspaces or have a reason (screen lock?) to ignore the request.

AlanGriffiths (changes requested):

src/server/frontend_wayland/xdg_activation_v1.cpp
mattkae: As we've discussed, what I've done is the right thing :+1:

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -21,16 +21,21 @@
...
 #include "mir/shell/shell.h"
-#include "mir/shell/focus_controller.h"
 #include "mir/wayland/protocol_error.h"
 #include "mir/log.h"
 #include <random>
 #include <chrono>
 #include <uuid.h>
+#include <mir/events/keyboard_event.h>
+
+#include "../../../examples/client/wayland_runner.h"
```
AlanGriffiths, line 34: I don't think we should be depending on example code

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -364,14 +399,24 @@ void mf::XdgActivationTokenV1::set_serial(uint32_t serial_, struct wl_resource*
...
+    //  Most likely this is supposed to be used that the application who
+    //  is using the application token is the application that we intend
+    //  to use that token.
+    (void)app_id;
 }
 
 void mf::XdgActivationTokenV1::set_surface(struct wl_resource* surface)
 {
-    token->requesting_surface = surface;
+    // TODO: This is the application id of the requesting surface.
```
AlanGriffiths, line 414: This doesn't look like an app id

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -250,22 +289,48 @@ mf::XdgActivationV1::KeyboardObserver::KeyboardObserver(mir::frontend::XdgActiva
...
+        return;
+
+    auto input_event = event->to_input();
+    if (input_event->input_type() != mir_input_event_type_key)
+        return;
+
+    auto keyboard_event = input_event->to_keyboard();
+    if (keyboard_event->action() == mir_keyboard_action_down)
+        xdg_activation_v1->invalidate_all();
 }
```
AlanGriffiths, line 305: I forget the discussion: Why do we care about keyboard events?

AlanGriffiths (changes requested):
Nearly there now

AlanGriffiths (approved):
LGTM

AlanGriffiths:
@Saviq you OK with this version?

Saviq (approved):
Seems to work!

Just need to be default :)

AlanGriffiths (approved):
Yes, even better now

PR #3661, opened 2024-11-01
Use PkgConfig to find development headers

sfalken:
openSUSE puts it's xkbcommon and libinput development headers in a non-standard location, which isn't Mir's problem, but it also doesn't implement PkgConfig properly to find those headers.

This is what I did to get it to package for openSUSE, but this should benefit any distribution that uses PkgConfig to find development files.

### DIFF
--- a/src/common/events/CMakeLists.txt
+++ b/src/common/events/CMakeLists.txt
@@ -39,4 +39,5 @@ add_library(
 target_link_libraries(mirevents
   PUBLIC
     mircore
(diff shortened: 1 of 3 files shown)


sfalken:
> Hey! sfalken has not signed the Canonical CLA which is required to get this contribution merged on this project.
> 
> Please head over to https://ubuntu.com/legal/contributors to read more about it.

Went and signed the thing, my launchpad ID is sfalken-u

PR #3662, opened 2024-11-06
Fix X11 dropdowns persisting for too long on unfocus

tarek-y-ismail:
For some reason, our code ignored synthetic unmap requests, even though [ICCCM 4.1.4](https://tronche.com/gui/x/icccm/sec-4.html) doesn't mention that it should be ignored. I tried commenting out the couple of lines responsible for ignoring the request and tested with gimp, qtcreator, and imagemagick. All seem to work better, especially gimp and qtcreator where this bug is most apparent.


AlanGriffiths (approved):
This looks sensible and, at worst, harmless.

But how to reproduce the problem it purports to solve? (qtcreator uses Wayland, when I switch focus from gimp menus are unmapped, ...)

Issue #3663, opened 2024-11-06
Divide by zero in `WindowInfo::constrain_resize()`


AlanGriffiths:
The methods `min_aspect()` and `max_aspect()` can return `AspectRatio{0U, std::numeric_limits<unsigned>::max()}` and `AspectRatio{std::numeric_limits<unsigned>::max(), 0U}` respectively. But `WindowInfo::constrain_resize()` divides by both `ar.height` and `ar.width` without checking for 0:

```
Thread 1 "miral-shell.bin" received signal SIGFPE, Arithmetic exception.
0x00007b8821332135 in miral::WindowInfo::constrain_resize (this=0x7b87a808a9c0, requested_pos=..., requested_size=...) at /home/alan/CLionProjects/mir/src/miral/window_info.cpp:229
229	            auto const width_correction  = (error+(ar.height-1))/ar.height;
```

PR #3664, opened 2024-11-06
Aspect ratios shouldn't contain zero

AlanGriffiths:
Fixes: #3663


robert-ancell (approved):
As long as we don't have to support a new compositor with zero width windows :)

PR #3665, opened 2024-11-08
Use old DRM cursor API as a placeholder for the atomic KMS hardware cursor

tarek-y-ismail:
This code will act as a placeholder temporarily until we implement damage, at which point, the real atomic KMS cursor  code (#3615) can be updated and merged.

### DIFF
--- a/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
+++ b/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
@@ -491,11 +491,12 @@ void mga::AtomicKMSOutput::wait_for_page_flip()
 bool mga::AtomicKMSOutput::set_cursor(gbm_bo* buffer)
 {
     int result = 0;
-    if (configuration.lock()->current_crtc)
+    auto conf = configuration.lock();
+    if (conf->current_crtc)
     {
         result = drmModeSetCursor(
             drm_fd_,
-            configuration.lock()->current_crtc->crtc_id,
+            conf->current_crtc->crtc_id,
             gbm_bo_get_handle(buffer).u32,
             gbm_bo_get_width(buffer),
             gbm_bo_get_height(buffer));
@@ -509,10 +510,11 @@ bool mga::AtomicKMSOutput::set_cursor(gbm_bo* buffer)
 
 void mga::AtomicKMSOutput::move_cursor(geometry::Point destination)
 {
-    if (configuration.lock()->current_crtc)
+    auto conf = configuration.lock();
+    if (conf->current_crtc)
     {
         if (auto result =
-                drmModeMoveCursor(drm_fd_, configuration.lock()->current_crtc->crtc_id, destination.x.as_int(), destination.y.as_int()))
+                drmModeMoveCursor(drm_fd_, conf->current_crtc->crtc_id, destination.x.as_int(), destination.y.as_int()))
         {
             mir::log_warning("move_cursor: drmModeMoveCursor failed (%s)", strerror(-result));
         }
@@ -522,9 +524,10 @@ void mga::AtomicKMSOutput::move_cursor(geometry::Point destination)
 bool mga::AtomicKMSOutput::clear_cursor()
 {
     int result = 0;
-    if (configuration.lock()->current_crtc)
+    auto conf = configuration.lock()
+    if (conf->current_crtc)
     {
-        result = drmModeSetCursor(drm_fd_, configuration.lock()->current_crtc->crtc_id, 0, 0, 0);
+        result = drmModeSetCursor(drm_fd_, conf->current_crtc->crtc_id, 0, 0, 0);
 
         if (result)
             mir::log_warning("clear_cursor: drmModeSetCursor failed (%s)",
--- a/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
+++ b/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
@@ -487,9 +487,8 @@ void mga::AtomicKMSOutput::wait_for_page_flip()
     pending_page_flip.get();
 }
 
-bool mga::AtomicKMSOutput::set_cursor(gbm_bo* buffer)
+void mga::AtomicKMSOutput::set_cursor(gbm_bo* buffer)
 {
-    int result = 0;
     if (auto conf = configuration.lock(); conf->current_crtc)
     {
         if (auto result = drmModeSetCursor(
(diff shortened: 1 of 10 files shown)


src/platforms/common/server/kms-utils/kms_connector.cpp
```diff
@@ -186,20 +189,40 @@ auto mgk::find_crtc_with_primary_plane(
...
+            ObjectProperties plane_props{drm_fd, plane->plane_id, DRM_MODE_OBJECT_PLANE};
+            switch(plane_props["type"])
             {
-                return std::make_pair(std::move(crtc), std::move(plane));
+                case DRM_PLANE_TYPE_PRIMARY:
+                    primary_plane = std::move(plane);
+                    break;
+                case DRM_PLANE_TYPE_CURSOR:
+                    cursor_plane = std::move(plane);
+                    break;
```
tarek-y-ismail, line 217: Not sure if this and other cursor plane related code fit in this PR.

src/platforms/atomic-kms/server/kms/cursor.cpp
```diff
@@ -0,0 +1,438 @@
...
+    case mir_orientation_normal:
+        return vector;
+    }
+}
+
+int get_drm_cursor_height(int fd)
+{
+    uint64_t height;
+    drmGetCap(fd, DRM_CAP_CURSOR_HEIGHT, &height);
+    return int(height);
```
RAOF, line 59: I know this is pre-existing, but this is an unusual way of spelling
```suggestion
    return static_cast<int>(height);
```

src/platforms/atomic-kms/server/kms/cursor.cpp
```diff
@@ -0,0 +1,438 @@
...
+}
+
+int get_drm_cursor_width(int fd)
+{
+    uint64_t width;
+    drmGetCap(fd, DRM_CAP_CURSOR_WIDTH, &width);
+    return int(width);
```
RAOF, line 66:
```suggestion
    return static_cast<int>(width);
```

src/platforms/common/server/kms-utils/kms_connector.cpp
RAOF: Yeah, I don't think that we need any of the cursor plane code in this one. Just assume a hardware cursor exists until a failure (which should happen in the `hide()` call from the `Cursor` constructor) tells you otherwise.

RAOF (commented):
Yup, that looks about right.

You can remove all the cursor plane stuff, and it needs conflicts resolving (there's locking around the current configuration, now).

src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -488,23 +488,50 @@ void mga::AtomicKMSOutput::wait_for_page_flip()
...
 }
 
 bool mga::AtomicKMSOutput::clear_cursor()
 {
-    return false;
-}
+    int result = 0;
+    if (configuration.lock()->current_crtc)
+    {
+        result = drmModeSetCursor(drm_fd_, configuration.lock()->current_crtc->crtc_id, 0, 0, 0);
```
RAOF, line 527: This is a TOCTOU¹ bug - because you're not *holding* the configuration lock, it's possible that another thread will null out `current_crtc` between your check and use.
```suggestion
    auto conf = configuration.lock()
    if (conf->current_crtc)
    {
        result = drmModeSetCursor(drm_fd_, conf->current_crtc->crtc_id, 0, 0, 0);
```

¹: Time Of Check to Time Of Use

src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -488,23 +488,50 @@ void mga::AtomicKMSOutput::wait_for_page_flip()
...
+    return !result;
 }
 
-void mga::AtomicKMSOutput::move_cursor(geometry::Point)
+void mga::AtomicKMSOutput::move_cursor(geometry::Point destination)
 {
+    if (configuration.lock()->current_crtc)
+    {
+        if (auto result =
+                drmModeMoveCursor(drm_fd_, configuration.lock()->current_crtc->crtc_id, destination.x.as_int(), destination.y.as_int()))
```
RAOF, line 515:
```suggestion
    auto conf = configuration.lock();
    if (conf->current_crtc)
    {
        if (auto result =
                drmModeMoveCursor(drm_fd_, conf->current_crtc->crtc_id, destination.x.as_int(), destination.y.as_int()))
```

RAOF (changes requested):
Apart from the mis-locking, this looks good. (Notably, in case anyone else wants to review, the major diff in `cursor.cpp` and `cursor.h` are a straight copy + namespace change of the same files from gbm-kms).

AlanGriffiths (approved):
Just nitpicking

src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -488,23 +488,53 @@ void mga::AtomicKMSOutput::wait_for_page_flip()
     pending_page_flip.get();
 }
 
-bool mga::AtomicKMSOutput::set_cursor(gbm_bo*)
+bool mga::AtomicKMSOutput::set_cursor(gbm_bo* buffer)
 {
-    return false;
+    int result = 0;
+    auto conf = configuration.lock();
+    if (conf->current_crtc)
```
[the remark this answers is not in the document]
RAOF: Huh. You live and learn. A bit of C++17 I was unaware of!

src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -488,23 +488,53 @@ void mga::AtomicKMSOutput::wait_for_page_flip()
...
+            conf->current_crtc->crtc_id,
+            gbm_bo_get_handle(buffer).u32,
+            gbm_bo_get_width(buffer),
+            gbm_bo_get_height(buffer));
+        if (result)
+        {
+            mir::log_warning("set_cursor: drmModeSetCursor failed (%s)", strerror(-result));
+        }
+    }
+    return !result;
```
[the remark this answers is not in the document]
tarek-y-ismail: Chris and I discussed this and came to the conclusion that failure/success in `set_cursor` doesn't really matter (at the moment).

RAOF (approved):
Yup, that's resolved the comments.

PR #3666, opened 2024-11-12
Add APIs to override decoration managers and customize decorations (Take 2)

tarek-y-ismail:
Attempt 1: Naive. Trying to expose an internal (an unstable) mir API to users.
Attempt 2: Bottom up. Trying to bring out the API from `msd::BasicDecoration` and expose it via adapters.
Attempt 3 (we are here): Top down. Trying to extract the minimal API needed for users to implement at least something similar to `msd::BasicDecoration` (The code is nasty on purpose)


src/miral/custom_decorations.cpp
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+#include "miral/custom_decorations.h"
+#include "decoration_manager_adapter.h"
+
+#include "miral/decoration_manager_builder.h"
+
+#include "mir/options/option.h"
+#include "mir/server.h"
+#define MIR_LOG_COMPONENT "CustomDecorations"
```
AlanGriffiths, line 24: I don't see any need to override the default (`MIR_LOG_COMPONENT_FALLBACK="miral"`)

src/miral/custom_decorations.cpp
```diff
@@ -0,0 +1,64 @@
...
+            if (options->is_set(custom_light_decorations_opt))
+            {
+                mir::log_info("Using custom decorations!");
+                server.set_the_decoration_manager_init(
+                    [adapter = self->adapter]
+                    {
+                        return adapter;
+                    });
+            }
+        });
```
AlanGriffiths, line 63: This is hard coding a choice that the user code should be making. (There may be different options to on|off, and there may not be any option.)

src/miral/symbols.map
```diff
@@ -458,6 +458,10 @@ global:
   extern "C++" {
     miral::ConfigFile::?ConfigFile*;
     miral::ConfigFile::ConfigFile*;
+    miral::CustomDecorations::CustomDecorations*;
+    miral::CustomDecorations::operator*;
+    miral::DecorationManagerBuilder::build*;
+    miral::DecorationManagerBuilder::done*;
```
AlanGriffiths, line 464: New symbol need to be in a 5.2 stanza, 5.1 shipped in Mir 2.18

src/server/shell/default_configuration.cpp
```diff
@@ -67,23 +65,27 @@ auto mir::DefaultServerConfiguration::the_window_manager_builder() -> shell::Win
...
+        [this]
         {
-            return std::make_shared<msd::BasicManager>(
-                *the_display_configuration_observer_registrar(),
-                [buffer_allocator = the_buffer_allocator(),
-                 executor = the_main_loop(),
-                 cursor_images = the_cursor_images()](
-                    std::shared_ptr<shell::Shell> const& shell,
-                    std::shared_ptr<scene::Surface> const& surface) -> std::unique_ptr<msd::Decoration>
+            if (!decoration_manager_init)
```
AlanGriffiths, line 70: There doesn't need to be an `if` here - just a working default implementation. But you'll probably be reworking the logic following the "wrap_..." model.

src/server/symbols.map
```diff
@@ -17,6 +17,7 @@ global:
     mir::DefaultServerConfiguration::add_wayland_extension*;
     mir::DefaultServerConfiguration::default_reports*;
     mir::DefaultServerConfiguration::set_enabled_wayland_extensions*;
+    mir::DefaultServerConfiguration::set_the_decoration_manager_init*;
```
AlanGriffiths, line 20: New symbols need a new stanza

examples/miral-shell/shell_main.cpp
```diff
@@ -150,6 +154,6 @@ int main(int argc, char const* argv[])
             pre_init(ConfigurationOption{[&](std::string const& typeface) { ::wallpaper::font_file(typeface); },
                                          "shell-wallpaper-font", "font file to use for wallpaper", ::wallpaper::font_file()}),
             ConfigurationOption{[&](std::string const& cmd) { terminal_cmd = cmd; },
-                                "shell-terminal-emulator", "terminal emulator to use", terminal_cmd}
+                                "shell-terminal-emulator", "terminal emulator to use", terminal_cmd},
```
AlanGriffiths, line 157: Spurious change?

examples/miral-shell/shell_main.cpp
```diff
@@ -135,6 +138,7 @@ int main(int argc, char const* argv[])
 
     return runner.run_with(
         {
+            CustomDecorations{UserDecorationManagerExample::get_adapter()},
```
AlanGriffiths, line 141: Is the adaptor created once on program startup and used for all decorations? Or might it need to maintain state associated with each window (and therefore be created once per window?)

include/miral/miral/decoration_manager_builder.h
```diff
@@ -0,0 +1,56 @@
...
+        std::function<void(std::shared_ptr<mir::scene::Surface> const& surface)> on_undecorate,
+        std::function<void()> on_undecorate_all) -> DecorationManagerBuilder;
+
+    auto done() -> std::shared_ptr<DecorationManagerAdapter>;
+
+private:
+    DecorationManagerBuilder();
+    std::shared_ptr<DecorationManagerAdapter> adapter;
+};
+}
```
AlanGriffiths, line 54: 1. I think the implementation type should be opaque (e.g. `DecorationManagerBuilder::Self`) for future proofing. (It could currently be a trivial subclass of `DecorationManagerAdapter`)
2. There needs to be a way to attach functors for input handling in addition to the decorate handling.

AlanGriffiths (changes requested):
Proposing the interface for review without "dragging implementation code along" is  good.

However, I would rather have focused first on separating out the "policy" part of decorations as otherwise the `miral::*Decoration*` interface is somewhat speculative. (In particular, there's no way to "hook up" input in the current draft.)

src/miral/custom_decorations.cpp
tarek-y-ismail: This is a remnant from when I had this file as an external example class. Thanks for pointing it out.

src/miral/custom_decorations.cpp
tarek-y-ismail: In my haste to experiment and get the code compiling, I forgot about checking the actual value the user passed. The decoration manager is now overridden only if they explicitly specify `--enable-custom-decorations=true`. 

I'm just using this for experimentation, so expect it to change soon.

examples/miral-shell/shell_main.cpp
tarek-y-ismail: Yep.

examples/miral-shell/shell_main.cpp
tarek-y-ismail: > Is the adaptor created once on program startup and used for all decorations? 

Yep.

> Or might it need to maintain state associated with each window (and therefore be created once per window?)
As it stands with `BasicDecoration`, the manager keeps track of decorations, and each decoration has it's own window-specific state.

include/miral/miral/decoration_manager_builder.h
```diff
@@ -0,0 +1,56 @@
...
+
+namespace miral
+{
+
+class DecorationManagerAdapter;
+class DecorationManagerBuilder
+{
+public:
+    static auto build(
+        std::function<void(std::weak_ptr<mir::shell::Shell> const& shell)> on_init,
```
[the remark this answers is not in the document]
tarek-y-ismail: Looking at `BasicDecoration`,  it uses  the shell for a bunch of things: resizing itself, moving, maximizing, minimizing, etc..

Is there another way to access this functionality?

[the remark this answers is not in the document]
AlanGriffiths: Maybe it would be a good thing to work on isolating the decoration "strategy" code before designing the API to customise it?

examples/miral-shell/decoration/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+include_directories(
+  ${CMAKE_CURRENT_BINARY_DIR}
+)
```
AlanGriffiths, line 19: Why?

examples/miral-shell/decoration/user_decoration_manager_example.cpp
AlanGriffiths, line 1: Unused?

examples/miral-shell/decoration/user_decoration_manager_example.h
AlanGriffiths, line 1: Unused?

examples/miral-shell/shell_main.cpp
```diff
@@ -14,6 +14,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "decoration/decoration.h"
+#include "miral/decoration.h"
```
AlanGriffiths, line 18:
```suggestion
```

examples/miral-shell/shell_main.cpp
```diff
@@ -33,9 +35,13 @@
 #include <miral/x11_support.h>
 #include <miral/wayland_extensions.h>
 
+#include "miral/custom_decorations.h"
+
 #include <xkbcommon/xkbcommon-keysyms.h>
 
 #include <cstring>
+#include <string>
+#include <vector>
```
AlanGriffiths, line 44:
```suggestion
```

examples/miral-shell/shell_main.cpp
```diff
@@ -132,24 +139,35 @@ int main(int argc, char const* argv[])
...
+            ::wallpaper::font_file()}),
+        ConfigurationOption{
+            [&](std::string const& cmd)
+            {
+                terminal_cmd = cmd;
+            },
+            "shell-terminal-emulator",
+            "terminal emulator to use",
+            terminal_cmd},
+    });
```
AlanGriffiths, line 172: I *think* the only substantive change is adding `CustomDecorations{UserDecoration::create_manager},`. The rest is unnecessary, and distracting, whitespace adjustment

include/miral/miral/decoration_manager_builder.h
AlanGriffiths, line 1: Not used by the example: shouldn't be part of the API

include/miral/miral/decoration_basic_manager.h
```diff
@@ -14,30 +14,42 @@
...
+public:
+    DecorationBasicManager(mir::Server&, Decoration::DecorationBuilder&&);
 
+    auto to_adapter() -> std::shared_ptr<DecorationManagerAdapter>;
 private:
-    Decoration(Decoration const&) = delete;
-    Decoration& operator=(Decoration const&) = delete;
+    struct Self;
+    std::shared_ptr<Self> self;
 };
```
AlanGriffiths, line 52: There doesn't seem to be much purpose to this class. It takes "stuff" in its constructor and returns "other stuff" from `to_adapter()` - that's pretty much a function in disguise.

AlanGriffiths (changes requested):
OK, you said this was "ready to review", but I now see it change whilst reviewing. I'll leave it for now and just post what I've seen so far

include/miral/miral/decoration_basic_manager.h
tarek-y-ismail: It now does a bunch of stuff before decoration creation (creating the window), and a bunch of stuff after to hook up the decoration adapter to some dependencies from the server.

examples/miral-shell/decoration/user_decoration_manager_example.h
tarek-y-ismail: Removed. Superseded by `miral::DecorationBasicManager`

examples/miral-shell/decoration/CMakeLists.txt
tarek-y-ismail: Copy-pasted from `examples/miral-shell/spinner/CMakeLitsts.txt`. Should be removed now

include/miral/miral/decoration_manager_builder.h
tarek-y-ismail: Now used in `miral::DecorationBasicManager`. It can be internal to `miral`, but I left it public in case any decoration authors need to build their own decoration manager from scratch (I don't really have usecase for this so we can try discussing it)

include/miral/miral/decoration_manager_builder.h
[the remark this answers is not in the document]
tarek-y-ismail: Ok, I understood your point when I dug into what `init` does. I decided that for the current use case, `DecorationManagerBuilder` is not really needed and `DecorationBasicManager` can directly use `DecorationManagerAdapter`.

src/server/symbols.map
tarek-y-ismail, line 20: Symbols for a non-existent function. Should be removed.

examples/miral-shell/decoration/decoration.cpp
```diff
@@ -0,0 +1,474 @@
...
+auto UserDecoration::InputManager::buttons() const -> std::vector<std::shared_ptr<Widget const>>
+{
+    std::vector<std::shared_ptr<Widget const>> button_rects;
+    for (auto const& widget : widgets)
+    {
+        if (widget->button)
+            button_rects.push_back(widget);
+    }
+
+    return button_rects;
```
tarek-y-ismail, line 46: Would've been nice to do this using the ranges API but it seems not ready yet...

AlanGriffiths: "not ready yet"? There's not much code using ranges in our codebase, but we have used it

include/miral/miral/decoration_adapter.h
```diff
@@ -0,0 +1,256 @@
...
+class Renderer
+{
+public:
+    static auto area(geometry::Size const size) -> size_t;
+
+    using Pixel = uint32_t;
+
+    // Can we use even more keywords?
+    static inline auto const buffer_format = mir_pixel_format_argb_8888;
+    static inline auto const bytes_per_pixel = 4;
```
AlanGriffiths, line 133: This looks like we are only supporting argb-8888 s/w rendering? Might the user want more choice?

include/miral/miral/decoration_adapter.h
```diff
@@ -0,0 +1,256 @@
...
+namespace ms = mir::scene;
+namespace mc = mir::compositor;
+namespace mg = mir::graphics;
+namespace geometry = mir::geometry;
+
+struct MirEvent;
+
+namespace miral
+{
+namespace decoration
```
AlanGriffiths, line 73: I'm still not convinced a namespace is appropriate. (I would probably nest auxiliary types in `CustomDecoration`.

include/miral/miral/decoration.h
```diff
@@ -0,0 +1,85 @@
...
+
+// Maybe it'd be best to move Device event to somewhere common between mir and miral?
+class DeviceEvent
+{
+public:
+    DeviceEvent(mir::shell::decoration::DeviceEvent);
+    operator mir::shell::decoration::DeviceEvent() const;
+
+    auto location() const -> mir::geometry::Point;
+    auto pressed() const -> bool;
```
AlanGriffiths, line 52: I think shell authors may want more information:

* what modifier keys are pressed
* touch count, and individual touches
* which mouse buttons are pressed

include/miral/miral/decoration.h
```diff
@@ -0,0 +1,85 @@
...
+struct DecorationRedrawNotifier
+{
+    void notify() const;
+
+    void register_listener(std::function<void()> on_redraw);
+    void clear_listener();
+
+private:
+    std::function<void()> on_redraw;
+};
```
AlanGriffiths, line 68: Not used, why does it exist?

include/miral/miral/decoration.h
```diff
@@ -0,0 +1,85 @@
...
+
+class Decoration // Placeholder names
```
AlanGriffiths, line 70: "Placeholder names"?

include/miral/miral/decoration_adapter.h
```diff
@@ -0,0 +1,256 @@
...
+class Buffer;
+}
+}
+
+namespace msh = mir::shell;
+namespace msd = msh::decoration;
+namespace ms = mir::scene;
+namespace mc = mir::compositor;
+namespace mg = mir::graphics;
+namespace geometry = mir::geometry;
```
AlanGriffiths, line 67: Pulling far too much implementation detail into the API

include/miral/miral/decoration_adapter.h
```diff
@@ -0,0 +1,256 @@
...
+
+    auto to_decoration() -> std::unique_ptr<msd::Decoration>;
+
+    void init(
+        std::shared_ptr<ms::Surface> const& window_surface,
+        std::shared_ptr<ms::Surface> const& decoration_surface,
+        std::shared_ptr<mg::GraphicBufferAllocator> const& buffer_allocator,
+        std::shared_ptr<msh::Shell> const& shell,
+        std::shared_ptr<mir::input::CursorImages> const& cursor_images
+    );
```
AlanGriffiths, line 247: This is not used by the example code, and looks like an implementation detail being leaked.

include/miral/miral/decoration_adapter.h
```diff
@@ -0,0 +1,256 @@
...
+
+        // Called by Mir to inform user code that the display scale changed
+        OnScaleChanged scale_changed,
+
+        // Called whenever the internal WindowState is updated
+        OnWindowStateUpdated update_decoration_window_state);
+
+    void set_custom_geometry(std::shared_ptr<miral::decoration::StaticGeometry> geometry);
+
+    auto to_decoration() -> std::unique_ptr<msd::Decoration>;
```
AlanGriffiths, line 239: This is not used by the example code, and looks like an implementation detail being leaked

include/miral/miral/decoration_adapter.h
```diff
@@ -0,0 +1,256 @@
...
+        // Called by Mir to inform user code of changes to the window being decorated
+        OnWindowAttribChanged attrib_changed,
+        OnWindowResized window_resized_to,
+        OnWindowRenamed window_renamed,
+
+        // Called by Mir to inform user code that the display scale changed
+        OnScaleChanged scale_changed,
+
+        // Called whenever the internal WindowState is updated
+        OnWindowStateUpdated update_decoration_window_state);
```
AlanGriffiths, line 235: A long list of parameters like this is not maintainable: C.f. discussion about using builders to help

AlanGriffiths (changes requested):
I still think that "expose everything so that a consumer-side implementation works, then try to hide the stuff that shouldn't be exposed" is a long route to the destination.

First separating the stuff that should be exposed and then exposing it would be easier

examples/miral-shell/decoration/decoration.cpp
tarek-y-ismail: Looks like I'm who's not ready. I spent about 20 minutes trying to filter and collect but couldn't figure it out.

include/miral/miral/decoration.h
tarek-y-ismail: It is used. This is the main way for the decoration to signal to Mir that it needs redrawing.

include/miral/miral/decoration.h
tarek-y-ismail: This is more of a note to myself. When I was getting the code to work at the start, I wrote down the first name that came to mind without giving it much thought. So I think giving that some time once everything settles might be good

include/miral/miral/decoration_adapter.h
tarek-y-ismail: `set_custom_decoration` is, `to_decoration` isn't.

include/miral/miral/decoration_adapter.h
AlanGriffiths: If these restrictions were part of the example, I would be fine with it. If we're telling shell developers "this is the only way" then I'm not. Our objective is an API we are willing to support indefinitely, that's very different to "while I'm experimenting around".

include/miral/miral/decoration_adapter.h
AlanGriffiths: I thought that `to_decoration()` was what I was commenting on

include/miral/miral/decoration.h
AlanGriffiths: This PR is trying to cover a number of issues:

1. separation of customisation points from generic logic
2. An API that exposes these customisation points
3. An implementation of that API
4. Demonstrating the API is sufficient and maintainable

This class is _not used by (nor useful for) the consumer of the API_: it should not exist in the API.

It is also a trivial wrapper around a functor: `using DecorationRedrawNotifier = std::function<void()>` is an alternative design

PR #3668, opened 2024-11-14
Small atomic-kms simplifications

RAOF:


AlanGriffiths:
As I read this, we're changing `schedule_page_flip()` to synchronously `page_flip()`. Apart from the misleading name, could that have performance implications for display groups?

RAOF:
The evolution here was:
1. Asynchronous page flip, with an optimisation for immediately waiting for it if there's only one output connected (ie: no clone mode/display group)
2. atomic-kms dropped display grouping; each output gets driven by its own compositor. Asynchronous page flip, but we then unconditionally immediately wait for it (this is the prior state)
3. Just synchronously page flip (you are here)

So this (shouldn't be) a visible behaviour change; we're just synchronously waiting fractionally earlier in the code.

(I have changed the name to make it hopefully less misleading)

tarek-y-ismail (approved):

tarek-y-ismail:
Just need #3665 to merge first, then we can fix whatever conflicts come up in this PR

PR #3673, opened 2024-11-18
Fix stuck-frame-after-mode-switch bug

RAOF:


tarek-y-ismail (commented):
Code and new docs look good. @Saviq does this fix the issue on your end?

Edit: Tried to see if I can test on my set up using the built-in display. Mir is convinced it only supports 1920x1080@144 Hz :/

src/server/compositor/multi_threaded_compositor.cpp
```diff
@@ -117,6 +118,23 @@ class CompositingFunctor
...
+            for (auto const& [_, compositor] : compositors)
+            {
+                auto elements = scene->scene_elements_for(compositor.get());
+                for (auto const& element : elements)
+                {
+                    // We need to actually access the buffer in order for it to be marked in use.
+                    element->renderable()->buffer();
+                }
+            }
+
```
AlanGriffiths, line 137: Ugly but I don't have a cleaner alternative to offer

PR #3674, opened 2024-11-18
feature: throwing an invalid_stride error when the size provided to wl_shm::create_pool is less than or equal to zero

mattkae:
I encountered this while running `cosmic-bg` on Mir. It still doesn't fix the problems there, but it avoids us blowing up in mmap


AlanGriffiths (approved):
Looks legit

PR #3675, opened 2024-11-18
bugfix: wlr layer shell surface are guaranteed to have a nonzero size

mattkae:
## What's new?
- Layer shell surfaces are guaranteed to have a requested size of 640x480 in the event that no size is found. This requests clients from making the mistake of allocating a zero-sized buffer
- `WindowWlSurfaceRole::current_size()` now returns std::nullopt when it cannot reasonably resolve a size

## To test
- Install cosmic-bg
- Run it
- Note that you do not crash


AlanGriffiths:
As discussed on chat, I think this is the wrong place for the fix: See #3676 for my alternative.

mattkae:
> As discussed on chat, I think this is the wrong place for the fix: See #3676 for my alternative.

Unfortunately did not fix my particular issue

mattkae:
Closing in favor of https://github.com/canonical/mir/pull/3676

PR #3679, opened 2024-11-26
bugfix: when placing a surface that ignores exclusion zones, the application_zone should be the full display_area

mattkae:
If the surface ignores exclusion zones, then we should use the display area instead of the application zone


AlanGriffiths (approved):
LGTM

PR #3681, opened 2024-11-27
Fix OSK issues on electron

### LINKED ISSUE #3580: OSK hides a half second after popping up on some input fields with `zwp_text_input_v1` (Electron)
Steps:
- tweak https://github.com/canonical/iot-example-graphical-snap/tree/22/Electron-quick-start to point at:
  https://demo.inductiveautomation.com/data/perspective/client/OnlineDemo/feature/reporting/examples
- build and install it
- start Mir, ubuntu-frame-osk and `iot-example-graphical-snap`
- find and focus the "Inteval" field

Expected:
- OSK pops up and stays up

Current:
- OSK pops up and quickly hides again

Even though the there's a lot of traffic, nothing in the client wayland requests suggests it should be hiding:
```
[4063403.674]  -> zwp_text_input_v1@23.reset()
[4063414.767]  -> zwp_text_input_v1@23.reset()
[4063418.886]  -> zwp_text_input_v1@23.activate(wl_seat@10, wl_surface@21)
[4063418.907]  -> zwp_text_input_v1@23.set_content_type(4, 3)
[4063419.005]  -> zwp_text_input_v1@23.show_input_panel()
[4063419.051]  -> zwp_text_input_v1@23.set_cursor_rectangle(1022, 354, 246, 236)
[4063419.086]  -> zwp_text_input_v1@23.set_surrounding_text("1", 1, 1)
[4063419.253]  -> zwp_text_input_v1@23.set_cursor_rectangle(1022, 354, 246, 236)
[4063419.343]  -> zwp_text_input_v1@23.set_cursor_rectangle(1010, 513, 0, 20)
[4063419.378]  -> zwp_text_input_v1@23.reset()
[4063419.461]  -> zwp_text_input_v1@23.show_input_panel()
[4063419.471]  -> zwp_text_input_v1@23.set_cursor_rectangle(1010, 513, 0, 20)
[4063419.497]  -> zwp_text_input_v1@23.set_cursor_rectangle(1010, 513, 0, 20)
[4063419.510]  -> zwp_text_input_v1@23.set_surrounding_text("1", 0, 1)
[4063419.542]  -> zwp_text_input_v1@23.set_cursor_rectangle(1010, 513, 0, 20)
[4063419.603] zwp_text_input_v1@23.leave()
[4063419.615] zwp_text_input_v1@23.enter(wl_surface@21)
[4063419.633]  -> zwp_text_input_v1@23.set_cursor_rectangle(1003, 513, 7, 20)
[4063421.949]  -> zwp_text_input_v1@23.set_cursor_rectangle(1003, 513, 7, 20)
[4063422.271]  -> zwp_text_input_v1@23.set_cursor_rectangle(1003, 513, 7, 20)
[4063519.588]  -> zwp_text_input_v1@23.show_input_panel()
[4063519.610]  -> zwp_text_input_v1@23.set_cursor_rectangle(1003, 513, 7, 20)
[4063622.079]  -> zwp_text_input_v1@23.set_cursor_rectangle(1003, 513, 7, 20)
[4063622.377]  -> zwp_text_input_v1@23.set_cursor_rectangle(1003, 513, 7, 20)
```

tarek-y-ismail:
Closes #3580. 

Fixes a few issues:
1. The "Interval" field showing the OSK for a split second

2. The "Interval" field (and similar fields) didn't get input changes as you used the OSK

3. Tabbing from the interval field to the minutes/hours/days field to its right opened the OSK in number mode.
This was due to `reset` being called between the content type being set in the state and the state being actually committed, causing Mir to lose the state and not commit it to the handler, leading to the previous mode (numbers) carrying over to the new field

### DIFF
--- a/src/server/frontend_wayland/text_input_v1.cpp
+++ b/src/server/frontend_wayland/text_input_v1.cpp
@@ -371,7 +371,7 @@ void TextInputV1::deactivate(wl_resource *seat)
     pending_state.reset();
 }
 
-/// Electron appears to call show_input_panel() when commit_state() should be called. 
+/// Electron appears to call show_input_panel() when commit_state() should be called.
 void TextInputV1::show_input_panel()
 {
     commit_state(0);
@@ -386,7 +386,7 @@ void TextInputV1::hide_input_panel()
 
 void TextInputV1::reset()
 {
-    pending_state.reset();
+    pending_state.emplace();
 }
 
 void TextInputV1::set_surrounding_text(const std::string &text, uint32_t cursor, uint32_t anchor)
--- a/src/server/frontend_wayland/text_input_v1.cpp
+++ b/src/server/frontend_wayland/text_input_v1.cpp
@@ -21,9 +21,6 @@
 #include "mir/executor.h"
 #include "mir/scene/text_input_hub.h"
 
-#define MIR_LOG_COMPONENT "TextInputV1"
-#include "mir/log.h"
-
 #include <memory>
 #include <boost/throw_exception.hpp>
 #include <deque>
@@ -369,6 +366,7 @@ void TextInputV1::activate(wl_resource *seat_resource, wl_resource *surface)
     {
         on_new_input_field = true;
         state = State::active;
+        pending_state = {};
     }
 }
 
@@ -394,11 +392,6 @@ void TextInputV1::hide_input_panel()
 
 void TextInputV1::reset()
 {
-    if(state == State::inactive)
-    {
-        mir::log_warning("Attempted to reset state while inactive");
-        return;
-    }
     pending_state = {};
 }
 
--- a/src/server/frontend_wayland/text_input_v1.cpp
+++ b/src/server/frontend_wayland/text_input_v1.cpp
@@ -20,7 +20,9 @@
 #include "wl_surface.h"
 #include "mir/executor.h"
 #include "mir/scene/text_input_hub.h"
-#include "mir/wayland/client.h"
+
+#define MIR_LOG_COMPONENT "TextInputV1"
+#include "mir/log.h"
 
 #include <memory>
 #include <boost/throw_exception.hpp>
@@ -171,8 +173,13 @@ private:
     mw::Weak<mf::WlSurface> current_surface;
     /// Set to true if and only if the text input has been enabled since the last commit
     bool on_new_input_field{false};
-    /// nullopt if the state is inactive, otherwise holds the pending and/or committed state
-    std::optional<ms::TextInputState> pending_state;
+
+    enum class State
+    {
+        active,
+        inactive,
+    } state;
+    ms::TextInputState pending_state;
 
     struct SerialPair
     {
@@ -247,7 +254,8 @@ TextInputV1::TextInputV1(
     std::shared_ptr<TextInputV1Ctx> const& ctx)
     : mw::TextInputV1{resource, Version<1>()},
       ctx{ctx},
-      handler{std::make_shared<Handler>(this, ctx->wayland_executor)}
+      handler{std::make_shared<Handler>(this, ctx->wayland_executor)},
+      state{State::inactive}
 {
 }
 
@@ -258,15 +266,15 @@ TextInputV1::~TextInputV1()
         seat.value()->remove_focus_listener(client, this);
     }
 
+    state = State::inactive;
     on_new_input_field = false;
-    pending_state.reset();
     ctx->text_input_hub->deactivate_handler(handler);
 }
 
 void TextInputV1::send_text_change(ms::TextInputChange const& change)
 {
     auto const client_serial = find_client_serial(change.serial);
-    if (!pending_state || !current_surface || !client_serial)
+    if (state == State::inactive || !current_surface || !client_serial)
     {
         // We are no longer enabled, or we don't have a valid serial
         return;
@@ -360,7 +368,7 @@ void TextInputV1::activate(wl_resource *seat_resource, wl_resource *surface)
     if (current_surface)
     {
         on_new_input_field = true;
-        pending_state.emplace();
+        state = State::active;
     }
 }
 
@@ -368,7 +376,7 @@ void TextInputV1::deactivate(wl_resource *seat)
 {
     (void)seat;
     on_new_input_field = false;
-    pending_state.reset();
+    state = State::inactive;
 }
 
 /// Electron appears to call show_input_panel() when commit_state() should be called.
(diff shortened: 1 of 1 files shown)


src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -427,16 +430,13 @@ void TextInputV1::commit_state(uint32_t client_serial)
...
         state_serials.push_back({client_serial, hub_serial});
         while (state_serials.size() > max_remembered_serials)
         {
             state_serials.pop_front();
         }
     }
-    else
-    {
-        ctx->text_input_hub->deactivate_handler(handler);
-    }
```
AlanGriffiths, line 439: This is the other substantive change: committing the state can no longer call `deactivate_handler()`

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -361,6 +362,7 @@ void TextInputV1::activate(wl_resource *seat_resource, wl_resource *surface)
     {
         on_new_input_field = true;
         pending_state.emplace();
+        pending_state_used = false;
```
AlanGriffiths, line 365: This is where we can break the hypothetical invariant, but I'm not sure what it achieves

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -427,16 +430,13 @@ void TextInputV1::commit_state(uint32_t client_serial)
     if (pending_state && current_surface)
     {
         auto const hub_serial = ctx->text_input_hub->set_handler_state(handler, on_new_input_field, *pending_state);
+        pending_state_used = true;
```
AlanGriffiths, line 433: It seems wrong not to reset `pending_state` value after using it in `commit_state()`. Vis:

    pending_state.emplace();

tarek-y-ismail:
Just as Alan suggested, going with `optional::emplace` instead of `optional::reset` somewhat fixes the issue a lot more elegantly. The keyboard pops up in text mode, and the user can manually switch to number mode to interact with numeric fields. I see this as a  good middle ground between not having a keyboard at all and having things "work" but risking future bugs with hacky workarounds. @Saviq What do you think?

Saviq:
> @Saviq What do you think?

So rather than popping up, and hiding, the numerical keyboard, we'd pop up the standard keyboard? Discarding the input field type?

Was Mir doing something wrong wrt. to the protocol? If not, I don't think we should be breaking that. If the OSK is reacting badly to us using the protocol correctly. that's something to fix in Squeekboard, IMO.

tarek-y-ismail:
I think it's more of the client (electron) misbehaving. We get a `reset` request between `activate` and `deactivate` for no apparent reason. On main, this causes us to null the optional which effectively disables the keyboard, Alan's solution guarantees at least a usable OSK.

Saviq:
> Alan's solution guarantees at least a usable OSK.

Ah, but a well behaved client will still get the full monty. WFM!

tarek-y-ismail:
> Ah, but a well behaved client will still get the full monty. WFM!

Any examples of such client? Just for testing purposes

Saviq:
> Any examples of such client? Just for testing purposes

I've only actually seen browsers choose the field format, so trying in different browsers (e.g. https://snapcraft.io/wpe-webkit-mir-kiosk), provided they actually support the input method protocol…

tarek-y-ismail:
Ok, so I just tested the demo URL in the original issue on both electron and wpe-webkit. With no changes to workaround the bug, wpe doesn't seem to exhibit this bug.

Saviq:
> wpe doesn't seem to exhibit this bug.

You mean with WPE and ubuntu-frame-osk, this PR is not needed and everything behaves fine?

So the issue is in Electron? Can you check Chromium out as well?

Saviq:
> because WPE webkit uses `TextInputV3`, which doesn't have a `reset` request.

And Electron / Chromium misuse the `reset` request in `TextInputV1`, sending it at the wrong time?

And how does the `.reset()` > `.emplace()` affect our protocol compliance? I suppose it clears the pending state, so what it says on the tin?

AlanGriffiths:
> And how does the `.reset()` > `.emplace()` affect our protocol compliance? I suppose it clears the pending state, so what it says on the tin?

It occurred to me that it still isn't quite right. If we receive a `reset()` _when not active_ (which is a weird scenario, and I've not checked if the protocol allows it), then we shouldn't be setting a value (by `emplace()`)

Saviq:
> If we receive a `reset()` _when not active_ (which is a weird scenario, and I've not checked if the protocol allows it)

> ```
> void zwp_text_input_v1_reset(
>     struct zwp_text_input_v1* zwp_text_input_v1
> )
> ```
> Should be called by an editor widget when the input state should be
> reset, for example after the text was changed outside of the normal
> input method flow.

Guess what. It's undefined.

> then we shouldn't be setting a value (by `emplace()`)

It's an empty pending state, so does it matter?

AlanGriffiths:
> It's an empty pending state, so does it matter?

The code is confusing: instead of having active/inactive states, or even an `is_active` boolean, `pending_state` is an optional that indicates "active" by having a value (even an empty one). See the rant in my review earlier

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -386,25 +394,30 @@ void TextInputV1::hide_input_panel()
 
 void TextInputV1::reset()
 {
-    pending_state.reset();
+    if(state == State::inactive)
+    {
+        mir::log_warning("Attempted to reset state while inactive");
+        return;
+    }
```
AlanGriffiths, line 401: Is this check worthwhile? It is likely to spam the log to no benefit.

If it is needed, we should probably be identifying the guilty client.

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -360,18 +368,18 @@ void TextInputV1::activate(wl_resource *seat_resource, wl_resource *surface)
     if (current_surface)
     {
         on_new_input_field = true;
-        pending_state.emplace();
+        state = State::active;
```
AlanGriffiths, line 371: It is still necessary to set `pending_state` to a default initialised state

src/server/frontend_wayland/text_input_v1.cpp
```diff
@@ -20,7 +20,9 @@
 #include "wl_surface.h"
 #include "mir/executor.h"
 #include "mir/scene/text_input_hub.h"
-#include "mir/wayland/client.h"
+
+#define MIR_LOG_COMPONENT "TextInputV1"
+#include "mir/log.h"
```
AlanGriffiths, line 25: I don't think we need the logging but, in general, I don't think we need `MIR_LOG_COMPONENT` at a file level. It is better to use `MIR_LOG_COMPONENT_FALLBACK` at a library level.

AlanGriffiths (changes requested):
Yes, that's easier to follow

src/server/frontend_wayland/text_input_v1.cpp
tarek-y-ismail: It will be helpful to identify misbehaving clients in bug reports in the future. And from what I saw when debugging, `reset` isn't called on every frame. Rather, it's called on specific events. So it shouldn't be spamming logs unless a client is real naughty.

I tried identifying the client through `Session::name` but it comes up as empty. Maybe there's another way that I didn't see?

AlanGriffiths: Is there a bug to report?

                Should be called by an editor widget when the input state should be
                reset, for example after the text was changed outside of the normal
                input method flow.

There's nothing there that says a client is misbehaving by requesting `reset()` when inactive

PR #3682, opened 2024-11-27
We broke the platform ABI since 2.18

AlanGriffiths:
As noted in https://github.com/canonical/mir/pull/3590#pullrequestreview-2299754528

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -77,7 +77,7 @@ Description: Display server for Ubuntu - server library
  .
  Contains the shared library needed by server applications for Mir.
 
-Package: libmirplatform29
+Package: libmirplatform30
 Section: libs
 Architecture: linux-any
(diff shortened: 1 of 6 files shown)


RAOF (changes requested):
Since we're bumping mirplatform ABI we should additionally be combining the stanzas in src/platform/symbols.map into a single 2.19 stanza, right?

AlanGriffiths:
> Since we're bumping mirplatform ABI we should additionally be combining the stanzas in src/platform/symbols.map into a single 2.19 stanza, right?

True

Saviq (approved):
> we should additionally be combining the stanzas in src/platform/symbols.map into a single 2.19 stanza, right?

Could our symbols check CI be extended to verify this?

PR #3693, opened 2024-12-04
Add focus stealing prevention to `MinimalWindowManager`

### LINKED ISSUE #2586: Implement focus stealing prevention

tarek-y-ismail:
Closes #2586.

Adds a new constructor to `FloatingWindowManagerPolicy` to control focus stealing prevention. This in a nutshell stops new windows from being focused and raised. When used with xdg-activation-v1, this improves security as external actors can't just steal focus by opening a new window (in addition to the niceties xdg-activation-v1 adds to usability).

TODO:
- [ ] ~Fix window offsetting. This is built off the position of the focused window, so all windows after the second one open in the same (x,y) position.~ 
Split off to #3695 
- [ ] ~Fix windows closing in the order of opening instead of front to back. see https://github.com/canonical/mir/pull/3693#discussion_r1869951543~ 
Moved to #3694 
- [ ] ~Make sure Xwayland applications work properly. My focus has been on Wayland applications so far.~
Xwayland doesn't seem to support xdg-activation at the moment?
- [x] Alt + tab predictably broken :/
Have to focus other applications before they work with alt + tab
- [x] Decorations are not pushed behind the focused window
- [ ] ~Need some way to focus applications launched via Mir (Ctrl-Alt+t/T for example)~
Will be in its own PR (#3703)


examples/example-server-lib/floating_window_manager.h
```diff
@@ -118,6 +126,8 @@ class FloatingWindowManagerPolicy : public miral::MinimalWindowManager
     void apply_workspace_hidden_to(miral::Window const& window);
 
     void keep_spinner_on_top();
+
+    bool focus_stealing_prevention{true};
```
AlanGriffiths, line 130: Should be `const`

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -63,6 +63,17 @@ FloatingWindowManagerPolicy::FloatingWindowManagerPolicy(
...
+FloatingWindowManagerPolicy::FloatingWindowManagerPolicy(
+    WindowManagerTools const& tools,
+    std::shared_ptr<SplashSession> const& spinner,
+    miral::InternalClientLauncher const& launcher,
+    std::function<void()>& shutdown_hook,
+    bool focus_stealing_prevention) :
+    FloatingWindowManagerPolicy(tools, spinner, launcher, shutdown_hook)
+{
+    this->focus_stealing_prevention = focus_stealing_prevention;
+}
```
AlanGriffiths, line 75: You could avoid updating `this->focus_stealing_prevention` and use initialisers (and `const`) by forwarding from the old constructor to the new.

AlanGriffiths (changes requested):
This is not what I expected. I expected a miral API to "opt in" to "focus stealing prevention" [FSP].

This demonstrates that there's already a way to "opt in" to FSP by writing a customise window management policy, but isn't what I expected (a new constructor to `MinimalWindowManager`)

AlanGriffiths (changes requested):
That's more like what I was expecting. One question and several nits.

On a API design note: `bool` parameters tend to make code hard to read:

```c++
    // unreadable
    foo(..., false, ...);
```

Consider using an `enum class`:

```c++
    // readable
    foo(..., FocusStealing::allow, ...);
```

That is written assuming something like
```c++
    enum class FocusStealing
    {
        allow,
        prevent
    };
```

src/miral/basic_window_manager.cpp
```diff
@@ -530,7 +530,20 @@ void miral::BasicWindowManager::focus_next_application()
...
+        while (focus_controller->focused_surface() != nullptr &&
+               !info_for(focus_controller->focused_surface()).can_be_active())
+        {
+            focus_controller->focus_next_session();
+
+            // If we wrapped around, then we found no sessions we can focus.
+            // We'll have to depend on the last resort below
+            if (focus_controller->focused_session() == first_session)
+                break;
+        }
```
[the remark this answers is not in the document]
tarek-y-ismail: I think this logic is likely bugged anyway. I was experimenting with fixing the window offsetting and figured out that windows still close in t he order of opening. While in my opinion the current focused window should close, then the other windows in the opposite order of opening.

For example, if we open windows 1,2,3,4. With 1 focused, and the order of windows behind 1 is 4, 3, 2.  
If we start closing windows, they will close in the order 1,2,3,4 (focused, far back, mid back, close back). I think it should be 1,4,3,2.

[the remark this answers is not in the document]
AlanGriffiths: You could also drop this from the "focus stealing" PR as it doesn't seem to be specifically related. (By my reading you can set up this scenario without it.)

There are quite a few scenarios with suboptimal window management behaviour: if not trivial to fix (without fear of breaking things) they can be logged as issues. We do have a window management "cleanup" epic.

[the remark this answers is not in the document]
tarek-y-ismail: Yeah I think I'll drop it for the time being

include/miral/miral/minimal_window_manager.h
```diff
@@ -29,6 +29,7 @@ class MinimalWindowManager : public WindowManagementPolicy
 {
 public:
     explicit MinimalWindowManager(WindowManagerTools const& tools);
+    explicit MinimalWindowManager(WindowManagerTools const& tools, bool focus_stealing_prevention);
```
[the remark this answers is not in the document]
tarek-y-ismail: A bit outside the scope of this PR, but maybe we should add a builder to avoid a combinatorial explosion with the next option we add?

src/miral/basic_window_manager.cpp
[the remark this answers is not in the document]
AlanGriffiths: > I'll drop it for the time being

Do file an issue with a reproducer

src/miral/minimal_window_manager.cpp
```diff
@@ -277,6 +298,13 @@ auto miral::MinimalWindowManager::confirm_inherited_move(WindowInfo const& windo
 void miral::MinimalWindowManager::advise_new_window(miral::WindowInfo const& window_info)
 {
     self->application_selector.advise_new_window(window_info);
+
+    // If focus stealing prevention is on, swap the old focused window (now in
+    // the back) with the new window in the front.
+    // If it's a legitimate window, it'll be focused and raised via
+    // xdg-activation.
+    if ((self->focus_stealing == FocusStealing::prevent) && tools.active_window())
+        tools.swap_tree_order(tools.active_window(), window_info.window());
```
AlanGriffiths, line 307: I'm not convinced that this is exactly right.

1. Consider there being a fullscreen window with focus when a notification window is created. I don't expect the notification to be activated (as it shouldn't get focus), but it should be above the fullscreen window. Trying to push it "below" seems wrong (even if it ultimately fails because of an implementation detail: the windows are not in the same layer);
2. Consider an application that creates a new window, e.g. a menu. This should go on top without needing activation. Here the call fails with `log_error("Unable to swap tree order due to overlap");` because BasicWindowManager checks for an overlap between the trees. Spamming the log with errors is a bad idea;
3. In addition to the `SurfaceStack`, `application_selector` is maintaining a stack of windows. I don't see this swap being notified (it will still have the new window on top). Having inconsistencies between the two stacks will surely lead to problems.

examples/miral-shell/shell_main.cpp
```diff
@@ -68,9 +71,24 @@ int main(int argc, char const* argv[])
...
+            return FocusStealing::allow;
+        }
+        else if (option == "prevent")
+        {
+            return FocusStealing::prevent;
+        }
+
+        BOOST_THROW_EXCEPTION(std::runtime_error("Invalid focus stealing choice: " + option));
+    };
+    auto focus_stealing_prevention = string_to_focus_stealing(default_focus_stealing);
```
AlanGriffiths, line 88: You get better diagnostics by reporting the error by throwing `AbnormalExit`.

But could avoid the custom parsing and error reporting using a boolean option (`--focus-stealing-prevention=off`).

AlanGriffiths (changes requested):
The API looks good. But I'm not sure the implementation is quite there

src/miral/minimal_window_manager.cpp
Saviq: On 1., what is "a notification window"? How would we distinguish it from any other surface?

My original bug report about stealing did mention notifications, because that's how it affected me. But _notifications_ should never be toplevels, and we can't be expected to treat them differently to any other toplevel.

AlanGriffiths: > we can't be expected to treat them differently to any other toplevel

This code is handling ALL windows, not just toplevels

tarek-y-ismail: Looking at window types:
```
    mir_window_type_normal,       /**< AKA "regular"                       */
    mir_window_type_utility,      /**< AKA "floating"                      */
    mir_window_type_dialog,
    mir_window_type_gloss,
    mir_window_type_freestyle,
    mir_window_type_menu,
    mir_window_type_inputmethod,  /**< AKA "OSK" or handwriting etc.       */
    mir_window_type_satellite,    /**< AKA "toolbox"/"toolbar"             */
    mir_window_type_tip,          /**< AKA "tooltip"                       */
    mir_window_type_decoration,
```

Maybe we can only swap `mir_window_type_{normal,utility}`?

tarek-y-ismail: For 3., the first thing that comes to mind is overloading `advise_new_window` to sync the two stacks together

AlanGriffiths: I proposes that the windows that should be swapped are `mir_window_type_normal`, in the "application" layer, and parentless; anything else can be left on top

tarek-y-ismail: Seems like some wayland applications use `mir_window_type_freestyle` as well.

AlanGriffiths: OK: in the "application" layer, and parentless

tarek-y-ismail: I relaxed the condition and tried around with different applications and things seem to work as expected.

tarek-y-ismail:
@Saviq @AlanGriffiths Window offsetting might be a bit more involved? 

The simplest thing I have in mind is making sure that when a window is placed, it shouldn't be completely occluding the windows beneath it in the surface stack. In the case of too many windows underneath, or just a really big window being opened, we can give up and occlude them. What do you think? Should we spin it off into its own thing?

Saviq:
> What do you think? Should we spin it off into its own thing?

Yes.

There's something that dawned on me… does e.g. Firefox activate its own new windows on e.g. Ctrl+N, or do we need an exception here for new windows of the currently active client? Which would also address any children opening?

examples/miral-shell/shell_main.cpp
```diff
@@ -68,9 +71,18 @@ int main(int argc, char const* argv[])
 
     SpinnerSplash spinner;
     InternalClientLauncher launcher;
+    auto constexpr default_focus_stealing_prevention = false;
+    auto const string_to_focus_stealing = [](bool focus_stealing_prevention)
```
AlanGriffiths, line 75: Not the best name now

src/miral/minimal_window_manager.cpp
```diff
@@ -126,7 +145,9 @@ auto miral::MinimalWindowManager::place_new_window(
 
 void miral::MinimalWindowManager::handle_window_ready(WindowInfo& window_info)
 {
-    if (window_info.can_be_active())
+    // If focus stealing prevention isn't enabled, activate on window ready (if
+    // possible). Otherwise, only activate the first opened window.
+    if (((self->focus_stealing == FocusStealing::allow) || !tools.active_window()) && window_info.can_be_active())
```
tarek-y-ismail, line 150: I think `!in_background` should be used here as well?

tarek-y-ismail:
> Firefox activate its own new windows on e.g. Ctrl+N

Nope. Only the first (mother?) instance activates. Ctrl+N opens a new window, but in the background.

AlanGriffiths (approved):
This looks reasonable. Not done much experimenting with it as I'm not entirely sure what the expected behaviour is. Will leave merging to @Saviq

Saviq:
> Nope. Only the first (mother?) instance activates. Ctrl+N opens a new window, but in the background.

That makes sense. And means the stealing prevention logic should only act when the new surface is from a client different than the active one.

AlanGriffiths:
> the stealing prevention logic should only act when the new surface is from a client different than the active one

+1

tarek-y-ismail:
Some bugs I found while trying to fix alt+tab/grave:
1. qterminal (or Mir) seems to have a bug where `Window::application` is not the same between two different instances? This isn't the case with other terminals/applications. So it's likely to be a qterminal bug.

2. There's a bug where if you spawn an app from a terminal, switch to the newly spawned app, and press alt+grave, instead of doing nothing because you only have one instance of the app, you'll switch to the terminal again. This happens with (gnome-terminal + xeyes, gnome-terminal+weston-terminal, qterminal+vlc)

    Debugging showed that `it->application()` and `(*originally_selected_it).application()` both have the same value, even with one instance of two different apps.
    
    https://github.com/canonical/mir/blob/56714dc3aee5bb20403db182ef1807f9f0c4ad25/src/miral/application_selector.cpp#L240

src/miral/application_selector.cpp
```diff
@@ -19,6 +19,7 @@
 #include <miral/application.h>
 #include <mir/log.h>
 #include <mir/scene/session.h>
+#include <unordered_set>
```
AlanGriffiths, line 22: Why?

src/miral/mru_window_list.h
```diff
@@ -29,6 +29,7 @@ class MRUWindowList
 public:
 
     void push(Window const& window);
+    void push_unfocused(Window const& window);
```
AlanGriffiths, line 32: This is more an "insert_below_top" than a "push"

AlanGriffiths (changes requested):
> qterminal (or Mir) seems to have a bug where Window::application is not the same between two different instances? This isn't the case with other terminals/applications. So it's likely to be a qterminal bug.

I don't think it is a bug: if you start two qterminal instances, you get two programs running each of which independently connects to the server, so the server sees two different applications.

If you start two `gnome-terminal` instances you get one `/usr/bin/gnome-terminal.real` program running which connect to the server once and opens two windows. So the server sees one application. (`cat $(which gnome-terminal)` for details)

src/miral/application_selector.cpp
tarek-y-ismail: Left over from when I tried to "optimize" the code. 2 minutes in and I realized it's written like that for a reason

src/miral/mru_window_list.h
tarek-y-ismail: I agree. I thought this naming was more concise but it's likely misleading and/or vague

src/miral/application_selector.cpp
```diff
@@ -204,6 +213,7 @@ auto ApplicationSelector::advance(bool reverse, bool within_app) -> Window
     auto it = find(selected);
 
     std::optional<Window> next_window = std::nullopt;
+    auto should_continue = true;
```
tarek-y-ismail, line 216: Pretty much the same logic, but I think this is a bit clearer. Not sure if I should revert or keep these changes..

src/miral/basic_window_manager.cpp
```diff
@@ -196,6 +196,16 @@ void miral::BasicWindowManager::surface_ready(std::shared_ptr<scene::Surface> co
...
+
+    // Need to push onto the MRU list for alt+tab to work correctly.
+    //
+    // Originally, applications were pushed only when selected, but due to
+    // focus stealing prevention, they were not activated and thus weren't
+    // selected nor added to mru_active_windows, breaking alt+tab
+    auto window = info_for(surface).window();
+    auto const top = mru_active_windows.top();
+    if(top && top != window)
+        mru_active_windows.push_unfocused(window);
```
tarek-y-ismail, line 208: Pretty sure this is a tad too hacky. Suggestions welcome :)

src/miral/minimal_window_manager.cpp
```diff
@@ -126,7 +147,9 @@ auto miral::MinimalWindowManager::place_new_window(
 
 void miral::MinimalWindowManager::handle_window_ready(WindowInfo& window_info)
 {
-    if (window_info.can_be_active())
+    // If focus stealing prevention isn't enabled, activate on window ready (if
+    // possible). Otherwise, only activate the first opened window.
```
tarek-y-ismail, line 151: Comment out of date

src/miral/minimal_window_manager.cpp
```diff
@@ -276,7 +299,20 @@ auto miral::MinimalWindowManager::confirm_inherited_move(WindowInfo const& windo
...
 {
-    self->application_selector.advise_new_window(window_info);
+    // If focus stealing prevention is on, swap the old focused window (now in
+    // the back) with the new window in the front.
+    //
+    // If it's a legitimate window, it'll be focused and raised via
+    // xdg-activation.
+    //
+    // This is limited to a couple of window types, as well as windows in the
+    // "application" layer that don't have a parent.
```
tarek-y-ismail, line 309: Comment out of date

examples/miral-shell/shell_main.cpp
```diff
@@ -139,6 +149,10 @@ int main(int argc, char const* argv[])
             WaylandExtensions{},
             X11Support{},
             ConfigureDecorations{},
+            pre_init(ConfigurationOption{[&](bool option)
+                    { focus_stealing_prevention = to_focus_stealing(option); },
+                    "focus-stealing-prevention", "allow or prevent focus stealing",
+                    default_focus_stealing_prevention}),
```
Saviq, line 155: Can this not require the `1`? Just be a command-line switch, rather than a valued option?

Saviq (approved):
> I'm not entirely sure what the expected behaviour is.

1. `miral-shell --focus-stealing-prevention 1`
2. run gnome-terminal
3. `gedit &`
4. `mate-terminal`
5. go back to the terminal
6. `gedit`

Because gedit supports activation, it should be activated each time (as described by #3639). But because `mate-terminal` doesn't, it doesn't get activated.

So +1 from me on the behaviour.

Just a nit on the command line option.

src/miral/minimal_window_manager.cpp
```diff
@@ -276,7 +297,17 @@ auto miral::MinimalWindowManager::confirm_inherited_move(WindowInfo const& windo
...
 {
-    self->application_selector.advise_new_window(window_info);
+    // If the window is prevented from stealing focus, swap the old focused
+    // window (now in the back) with the new window in the front.
+    //
+    // If it's a legitimate window, it'll be focused and raised via
+    // xdg-activation.
+    auto in_background = self->prevent_focus_stealing(window_info);
+
+    self->application_selector.advise_new_window(window_info, in_background);
```
AlanGriffiths, line 307: 1. It looks odd to call a function on `self` to get a value passed to a function on `self` (shouldn't all this logic be in `self`?); and,
2. `in_background` isn't the clearest name (as evidenced by the long comment explaining it) and could be `const`

src/miral/minimal_window_manager.cpp
```diff
@@ -606,3 +637,12 @@ void miral::MinimalWindowManager::Impl::apply_resize_by(Displacement movement)
         gesture = Gesture::none;
     }
 }
+
+bool miral::MinimalWindowManager::Impl::prevent_focus_stealing(miral::WindowInfo const& info)
+{
+    auto const normal_app_without_parent = (info.depth_layer() == mir_depth_layer_application && !info.parent());
+    auto const decoration_surface = info.type() == mir_window_type_decoration;
```
AlanGriffiths, line 644: Why are we checking for decorations? Won't that mean that decorations will be raised and focussed regardless of parents and layer? (Which seems wrong)

AlanGriffiths (changes requested):
The TODO list doesn't look "ready", does it need updating?

src/miral/minimal_window_manager.cpp
tarek-y-ismail: So far, I only see decorations attached to parent windows, so we don't need to check for that. But your comment about the layer is true.

Saviq:
@tarek-y-ismail don't add the `launch_app_env` here, make a separate PR based on this one, instead.

Saviq:
@tarek-y-ismail anything left to do here?

tarek-y-ismail:
> @tarek-y-ismail anything left to do here?

No, just waiting for @AlanGriffiths to give it the thumbs up. (Maybe I should've nagged him a bit...)

examples/miral-shell/shell_main.cpp
```diff
@@ -132,13 +134,23 @@ int main(int argc, char const* argv[])
...
 
     return runner.run_with(
         {
             CursorTheme{"default:DMZ-White"},
             WaylandExtensions{},
             X11Support{},
             ConfigureDecorations{},
+            pre_init(ConfigurationOption{[&](bool is_set)
+                    { focus_stealing_prevention = to_focus_stealing(is_set); },
+                    "focus-stealing-prevention", "allow or prevent focus stealing"}),
```
AlanGriffiths, line 153: The semantics of "allow or prevent focus stealing" true are unclear, as is the absence of a default.
```suggestion
            pre_init(ConfigurationOption{[&](bool is_set)
                    { focus_stealing_prevention = to_focus_stealing(is_set); },
                    "focus-stealing-prevention", "prevent focus stealing", true}),
```

src/miral/minimal_window_manager.cpp
AlanGriffiths: So, we shouldn't be checking for decorations?

AlanGriffiths (changes requested):
> just waiting for @AlanGriffiths to give it the thumbs up

I don't have to "thumbs up" every PR. And it is hard to know that a PR is ready when there are unchecked "TODO" items in the top comment.

I've not tested the behaviour, but there's a few comments to address

examples/miral-shell/shell_main.cpp
Saviq: That's on me, I thought just having a flag would be simpler, but that wouldn't allow for disabling it.

AlanGriffiths: Well, a flag could also work (it just changes from the default) I just don't like both the description and the default being unclear

src/miral/minimal_window_manager.cpp
tarek-y-ismail: Removing both booleans entirely keeps things working fine (according to [saviq's test case](https://github.com/canonical/mir/pull/3693#pullrequestreview-2511613682)[^1]). In retrospect, I don't really know why I added those extra checks.

[^1]: Had to use `weston-terminal` instead of `mate-terminal`. Both don't support activation.

tarek-y-ismail:
> I don't have to "thumbs up" every PR.

Just making sure the changes you wanted before are all satisfied

AlanGriffiths (changes requested):

AlanGriffiths (dismissed):
This is not working as I expect.

1. start miral-app
2. from an unrelated terminal run `WAYLAND_DISPLAY=wayland-1 gedit`
3. In the Mir-on-X window (which now shows gedit) press `ctrl-alt-x` to launch xterm

Expect: xterm on top, with focus
Actual: xterm behind gedit and the log shows:
```
[2025-01-17 17:14:10.687630] < - ERROR - > mirserver: XdgActivationV1::activate invalid token: ��.]
```

The same with Ctrl-Alt-T to launch a terminal (but that can conflict with the host desktop)

AlanGriffiths (changes requested):
The default "focus-stealing-prevention" should maintain existing behaviour (i.e. be `false`

examples/miral-shell/shell_main.cpp
```diff
@@ -132,13 +134,23 @@ int main(int argc, char const* argv[])
...
 
     return runner.run_with(
         {
             CursorTheme{"default:DMZ-White"},
             WaylandExtensions{},
             X11Support{},
             ConfigureDecorations{},
+            pre_init(ConfigurationOption{[&](bool is_set)
+                    { focus_stealing_prevention = to_focus_stealing(is_set); },
+                    "focus-stealing-prevention", "prevent focus stealing", true}),
```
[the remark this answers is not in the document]
tarek-y-ismail: I'm also at fault for applying the suggestion without a second thought. Should be fixed now.

AlanGriffiths (approved):
Done

PR #3703, opened 2024-12-18
Pass XDG_ACTIVATION_TOKEN to apps launched via `launch_app_env`

tarek-y-ismail:


src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -231,6 +146,21 @@ std::shared_ptr<mf::XdgActivationTokenData> const& mf::XdgActivationV1::create_t
     }
 }
 
+std::string mf::XdgActivationV1::create_token_string()
+{
+    auto generated = generate_token();
+    auto token = std::make_shared<XdgActivationTokenData>(generated, main_loop->create_alarm([this, generated]()
+    {
+        try_consume_token(generated);
+    }), nullptr);
```
tarek-y-ismail, line 155: As far as I can tell from testing, passing `nullptr` for the session doesn't cause issues. But I think it deserves a closer look tomorrow. In the meantime, @mattkae since your originally wrote this, what do you think?

mattkae: I mean that is a required field so it will probably break if this is nullptr :) I'm assuming that the code will see that the parent session has expired and refuse to raise anything.

src/server/symbols.map
```diff
@@ -1404,3 +1404,10 @@ global:
 local: *;
 };
 
+MIR_SERVER_INTERNAL_2.18 {
```
AlanGriffiths, line 1407: The next release will not be 2.18

src/server/server.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include "mir/server.h"
 
+#include "frontend_wayland/xdg_activation_v1.h"
```
AlanGriffiths, line 19: This looks like a wrong approach.

We shouldn't be exposing the whole (or, probably, any) of xdg-activation to the server code: 

1. it is pointless to "pretend" to be a client requesting a token, as the pretend client will not have focus when the token is invoked;
2. this code shouldn't care when xdg-activation-v2 or xdg-activation (stable) comes along

A better approach would be to have a token issue/validation API used by both the launch code and the activation protocol

src/miral/launch_app.h
```diff
@@ -35,6 +35,12 @@ auto launch_app_env(std::vector<std::string> const& app,
     mir::optional_value<std::string> const& wayland_display,
     mir::optional_value<std::string> const& x11_display,
     AppEnvironment const& app_env) -> pid_t;
+
+auto launch_app_env(std::vector<std::string> const& app,
+    mir::optional_value<std::string> const& wayland_display,
+    mir::optional_value<std::string> const& x11_display,
+    std::optional<std::string> const& xdg_activation_token,
+    AppEnvironment const& app_env) -> pid_t;
```
AlanGriffiths, line 43: I don't think we need a new function, just a change to the signature

src/server/server.cpp
```diff
@@ -480,6 +478,22 @@ auto mir::Server::x11_display() const -> mir::optional_value<std::string>
     BOOST_THROW_EXCEPTION(std::logic_error("Cannot open connection when not running"));
 }
 
+auto mir::Server::get_activation_token() const -> std::optional<std::string>
+{
+    if (auto const config = self->server_config)
+    {
+        auto const& connector = std::dynamic_pointer_cast<mir::frontend::WaylandConnector>(config->the_wayland_connector());
+        auto xdg_activation_impl = std::static_pointer_cast<mir::frontend::XdgActivationV1>(connector->get_extension("xdg_activation_v1"));
```
AlanGriffiths, line 486: We shouldn't need to break abstractions like this. (When you find yourself writing code like this, stop and reconsider the approach)

AlanGriffiths (changes requested):
Drilling into the frontend implementation from the Server abstraction is placing logic in the wrong place, fragile when that implementation changes, and probably won't work.

We first need to break out the token issue & validation logic from xdg-activation so that it can meet the new requirements. (With this scenario there is no "requester" to check still has focus)

src/miral/launch_app.h
tarek-y-ismail: Oh I didn't notice that this was internal

src/include/server/mir/shell/token_authority.h
tarek-y-ismail, line 1: Will be moved to the source file, just having the code  here for easier experimentation

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,137 @@
...
+        main_loop{main_loop}
+    {
+    }
+
+    auto issue_token(std::optional<Token::RevocationListener> revocation_listener) -> Token
+    {
+        std::scoped_lock lock{mutex};
+
+        auto const generated = generate_token();
+        auto alarm = main_loop->create_alarm(
```
tarek-y-ismail, line 93: This line intermittently returns a null alarm for some reason

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -329,13 +260,18 @@ void mf::XdgActivationV1::Instance::activate(std::string const& token, struct wl
...
     auto xdg_token = xdg_activation_v1->try_consume_token(token);
     if (!xdg_token)
     {
         mir::log_error("XdgActivationV1::activate invalid token: %s", token.c_str());
         return;
     }
 
+    // try_consume_token removes the token from our list of tokens, need to
+    // inform the token authority as well.
+    xdg_activation_v1->token_authority->revoke_token(token);
```
tarek-y-ismail, line 273: This call will call `try_consume_token` one more but it will do nothing since the token has already been removed, not sure how I can prevent this extra call

src/server/server.cpp
```diff
@@ -480,6 +479,16 @@ auto mir::Server::x11_display() const -> mir::optional_value<std::string>
     BOOST_THROW_EXCEPTION(std::logic_error("Cannot open connection when not running"));
 }
 
+auto mir::Server::get_activation_token() const -> std::optional<std::string>
+{
+    if (auto const config = self->server_config)
+    {
+        return config->the_token_authority()->issue_token({});
```
tarek-y-ismail, line 486: Not sure if  this is better or I should add an overload with no parameters

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,137 @@
...
+            std::unique_ptr<time::Alarm> alarm,
+            std::optional<RevocationListener> revocation_listener) :
+            token{token},
+            alarm{std::move(alarm)},
+            revocation_listener{revocation_listener}
+        {
+            alarm->reschedule_in(timeout_ms);
+        }
+
+        explicit Token(std::string token) :
```
tarek-y-ismail, line 55: A bit sloppy but useful for the set calls below. Should be private with `TokenAuthority` as a friend  thouth though

src/server/frontend_wayland/xdg_activation_v1.cpp
tarek-y-ismail, line 273: Maybe a better approach would be to consume the token on the token authority side and  return it, which would call the "revocation callback" and remove it from xdg activation as well

tarek-y-ismail: Can't work because we need the xdg-activation specific data. So we need to first get that, then request the token authority to remove the token (which removes the extra xdg-activation specific data as well)

src/include/server/mir/shell/token_authority.h
tarek-y-ismail: I  was moving out of an object and into a data member in a constructor. Both had the same name. Then I used the name to call a function. For some reason the compiler chose to use the passed in object instead of `this->object`

src/server/frontend_wayland/xdg_activation_v1.h
tarek-y-ismail, line 1: Changes made here should be reverted

src/server/symbols.map
```diff
@@ -100,6 +100,7 @@ global:
     mir::DefaultServerConfiguration::the_surface_input_dispatcher*;
     mir::DefaultServerConfiguration::the_surface_stack*;
     mir::DefaultServerConfiguration::the_text_input_hub*;
+    mir::DefaultServerConfiguration::the_token_authority*;
```
tarek-y-ismail, line 103: Don't think these belong here

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,86 @@
...
+            return s.hash();
+        }
+    };
+    std::unordered_set<Token, TokenHash> issued_tokens;
+
+    std::shared_ptr<MainLoop> main_loop;
+    std::mutex mutable mutex;
+
+    /// Time in milliseconds that the compositor will wait before invalidating a token
+    static auto constexpr timeout_ms = std::chrono::seconds(3000);
```
AlanGriffiths, line 80: The comment and naming agree, but the code doesn't!

src/server/symbols.map
```diff
@@ -1404,3 +1404,22 @@ global:
...
+    mir::shell::TokenAuthority::Token::Token*;
+    mir::shell::TokenAuthority::Token::hash*;
+    mir::shell::TokenAuthority::Token::operator*;
+    mir::shell::TokenAuthority::TokenAuthority*;
+    mir::shell::TokenAuthority::issue_token*;
+    mir::shell::TokenAuthority::revoke_token*;
+    non-virtual?thunk?to?mir::DefaultServerConfiguration::the_token_authority*;
+    typeinfo?for?mir::shell::TokenAuthority::Token;
+    typeinfo?for?mir::shell::TokenAuthority;
+    virtual?thunk?to?mir::DefaultServerConfiguration::the_token_authority*;
```
AlanGriffiths, line 1422: Adding virtual functions is an ABI break (at least in this case): so this should be an soname bump, not a new symbols stanza

AlanGriffiths (changes requested):
Looks like a better architecture. But I've not yet looked closely enough to decide how tokens supplied via `launch_app_env()` will be handled by xdg-activation

src/server/symbols.map
tarek-y-ismail: Done. I also renamed `debian/libmirserver62.install` to `debian/libmirserver63.install` and merged the two stanzas  into the 2.20 stanza (there was a comment next to `MIRSERVER_ABI` version that said to update the minor version, and since you said a new stanza was not needed, I understood that the old and new ones should be merged). 

Anyway, I checked both the DSO versioning guide and the guide linked at its bottom (just skimmed the C++ virtual function section, still need to read through it at some point...) and both don't mention this. So maybe the DSO versioning guide is due to have some updates?

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,86 @@
...
+        std::string token;
+        std::shared_ptr<mir::time::Alarm> const alarm;
+        std::optional<RevocationListener> revocation_listener;
+    };
+
+    TokenAuthority(std::shared_ptr<MainLoop> main_loop);
+
+    auto issue_token(std::optional<Token::RevocationListener> revocation_listener) -> Token;
+
+    void revoke_token(std::string to_remove);
```
RAOF, line 62: I think it's a bit strange that `issue_token` returns a `Token` but `revoke_token` takes  a `std::string`. This should probably be symmetric? (Or possibly even a `Token::revoke()` method?)

RAOF (changes requested):
Broadly, I think the `TokenAuthority` is kinda half way there. It's not actually an authority - code can't use it to check if a `Token` is valid, and that's tracked separately by `XdgActivationV1`. This is also why there's spooky action-at-a-distance in `XdgActivationV1::invalidate_*()` - you're maintaining a second list of valid tokens, mutating `XdgActivationV1::pending_tokens` by calling `token_authority->revoke_token(...)`.

I think this would be simpler if:
a) The `TokenAuthority` was the... authority for whether a token was valid or not, and
b) `Token` was treated more as an opaque... token, rather than immediately decaying into a `std::string` and all the interfaces taking that.

src/miral/launch_app.cpp
```diff
@@ -93,46 +93,21 @@ auto Environment::exec_env() const & -> std::vector<char const*>
...
     return result;
 }
-}  // namespace
 
-auto miral::launch_app_env(
-    std::vector<std::string> const& app,
-    mir::optional_value<std::string> const& wayland_display,
-    mir::optional_value<std::string> const& x11_display,
-    miral::AppEnvironment const& app_env) -> pid_t
+void assign_or_unset(Environment& env, std::string const& key, auto optional_value)
```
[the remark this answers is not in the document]
tarek-y-ismail: This won't quite work because `mir::optional_value` can't be converted to `std::optional`.

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,86 @@
...
+    {
+        std::size_t operator()(mir::shell::TokenAuthority::Token const& s) const noexcept
+        {
+            return s.hash();
+        }
+    };
+    std::unordered_set<Token, TokenHash> issued_tokens;
+
+    std::shared_ptr<MainLoop> main_loop;
+    std::mutex mutable mutex;
```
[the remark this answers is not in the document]
tarek-y-ismail: It protects `issued_tokens` assuming that the token authority might be accessed from multiple places. At least `ExternalClientLauncher::launch` and `XdgActivationV1` at the moment. I guess the naming was not the best.

Anyway, since the mutex protects only one thing, I can see `mir::Synchronised` being useful. Thanks for the suggestion!

src/include/server/mir/shell/token_authority.h
tarek-y-ismail: I've went through and tightened up the places where strings were used when I could've used tokens. Some places still have to use strings, but that's unavoidable.

src/miral/launch_app.cpp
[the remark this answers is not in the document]
RAOF: Ah, whoops. Missed the actual argument type (which is why we shouldn't be deducing it here :wink:)

This doesn't need to be a template function at all though, right? This would work, yes?
```suggestion
void assign_or_unset(Environment& env, std::string const& key, mir::optional_value<std::string> const& optional_value)
```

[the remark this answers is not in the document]
RAOF: Ah, no, because `miral::launch_app_env` takes *both* `mir::optional_value` and `std::optional` parameters; *that* is the thing that should be fixed!

src/miral/launch_app.cpp
```diff
@@ -155,12 +130,34 @@ auto miral::launch_app_env(
...
     return pid;
 }
+}  // namespace
+
+
+auto miral::launch_app_env(
+    std::vector<std::string> const& app,
+    mir::optional_value<std::string> const& wayland_display,
+    mir::optional_value<std::string> const& x11_display,
+    std::optional<std::string> const& xdg_activation_token,
```
RAOF, line 150: We should either migrate the `mir::optional_value`s to `std::optional`, or consistently use `mir::optional_value` here; I think mixing the two is worse than either option.

RAOF (commented):
Yeah, that's cleaner. On further thought, I'm not sure we sensibly *can* decouple `XdgActivation` and the `TokenAuthority` further.

src/miral/launch_app.cpp
tarek-y-ismail: Moved to `std::optional`, nothing user-facing broken

tarek-y-ismail:
TIL that github's conflict editor merges (there's probably a rebase option that I missed)

AlanGriffiths (changes requested):
```
/usr/bin/ld: unable to find version dependency `MIR_SERVER_INTERNAL_2.19'
/usr/bin/ld: duplicate version tag `MIR_SERVER_INTERNAL_2.20'
collect2: error: ld returned 1 exit status
```
src/server/symbols.map is ill formed

AlanGriffiths:
(With the symbols.map fixed)

From Gnome:

1. `miral-app --focus-stealing-prevention true`
2. `Ctrl-Alt-X` to launch `xterm`
3. `Ctrl-Alt-Shift-T` to launch `gnome-terminal`

Expect: `gnome-terminal` gets focus and raised
Actual: `xterm` remains on top with focus

tarek-y-ismail:
Hopefully both issues should be fixed now (facepalm)

I recommend testing with firefox and kgx. gnome-terminal doesn't seem to request activation.

AlanGriffiths (changes requested):
I think this is the usecase the PR is intended to address:

> From Gnome:
> 
>     1. `miral-app --focus-stealing-prevention true`
> 
>     2. `Ctrl-Alt-X` to launch `xterm`
> 
>     3. `Ctrl-Alt-Shift-T` to launch `gnome-terminal`
> 
> 
> Expect: `gnome-terminal` gets focus and raised
> Actual: `xterm` remains on top with focus

Same with:

    miral-app --focus-stealing-prevention true --shell-terminal-emulator=xfce4-terminal
    miral-app --focus-stealing-prevention true --shell-terminal-emulator=qterminal

So it isn't just gnome-terminal being weird

AlanGriffiths:
> * 
> 
> 
> Applications that DO NOT request activation:
> 
>     * gnome-terminal

gnome terminal *does* request activation:

```
[2025-01-22 11:46:17.010620] < - ERROR - > mirserver: XdgActivationV1::activate invalid token: gnome-shell/CLion/962086-6-Octopull-Framework16_TIME194122238
```

But that's not the token in `XDG_ACTIVATION_TOKEN`

tarek-y-ismail:
> gnome terminal does request activation

That's very weird, it doesn't seem to do so when launched via miral-terminal. But then again, gnome-terminal has been a massive headache so far

> But that's not the token in XDG_ACTIVATION_TOKEN

I have no idea why gnome-terminal is trying to doxx you, I'm not even sure how it put that as the activation token.

AlanGriffiths (changes requested):
Nitpicking

AlanGriffiths:
Also
```
  New Symbols 🟢🟢🟢
    mir::shell::TokenAuthority::get_bogus_token*;
```

src/include/server/mir/shell/token_authority.h
```diff
@@ -71,7 +72,7 @@ class TokenAuthority
     std::shared_ptr<MainLoop> main_loop;
 
     /// Time in milliseconds that the compositor will wait before invalidating a token
-    static auto constexpr timeout_ms = std::chrono::milliseconds(3000);
+    static auto constexpr timeout_ms = std::chrono::milliseconds(10000);
```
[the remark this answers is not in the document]
tarek-y-ismail: Used it for debugging purposes. Will revert to 3 seconds, although I feel that some applications might need more (IntelliJ comes to mind)

src/miral/x11_support.cpp
```diff
@@ -21,6 +21,7 @@
 #include <mir/options/configuration.h>
 #include <mir/main_loop.h>
 #include <mir/fd.h>
+#include <mir/fatal.h>
```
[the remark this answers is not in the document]
tarek-y-ismail: I don't think so. One of the includes I cleaned up (`mir/optional_value.h`) used to include this header, when it was removed, this header was not included and the code wouldn't compile. So I included it explicitly.

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -200,68 +188,97 @@ mf::XdgActivationV1::XdgActivationV1(
...
+}
+
+std::shared_ptr<mf::XdgActivationTokenData> const mf::XdgActivationV1::create_token(std::shared_ptr<ms::Session> const& session)
+{
+    if (!session_listener->is_focused_session(session))
     {
-        try_consume_token(generated);
-    }), session);
+        mir::log_warning(
+            "Client from a different session trying to create a token. Session name: %s, PID: %d. Returning bogus "
```
AlanGriffiths, line 234:
```suggestion
            "Client from a unfocussed session trying to create a token. Session name: %s, PID: %d. Returning bogus "
```

Also, I don't think frontend_wayland populates session names. So that may be uninformative.

From the discussion about client behaviour, this might spam the logs.

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -329,21 +364,32 @@ void mf::XdgActivationV1::Instance::activate(std::string const& token, struct wl
...
     //    of the surface with that token
-    auto xdg_token = xdg_activation_v1->try_consume_token(token);
-    if (!xdg_token)
+    //
+
+    auto token = xdg_activation_v1->token_authority->get_token_for_string(string_token);
+    if(!token)
     {
-        mir::log_error("XdgActivationV1::activate invalid token: %s", token.c_str());
+        mir::log_error("XdgActivationV1::activate invalid token: %s", string_token.c_str());
```
AlanGriffiths, line 372: I'm not sure this is an error? Possibly a warning?

But I fear spamming the logs

src/server/shell/token_authority.cpp
```diff
@@ -0,0 +1,161 @@
...
+    // Let it pass, we'll refuse to activate when we get the token later
+    if(token == get_bogus_token())
+        return true;
+
+    return issued_tokens.contains(token);
+}
+
+auto msh::TokenAuthority::get_bogus_token() const -> Token
+{
+    return Token{"I AM A TEAPOT", {}};
```
AlanGriffiths, line 106: I'm not convinced a bogus token should have such an easily identified value. Shouldn't it be initialised identically to the non-bogus ones, but not registered in `issued_tokens`?

AlanGriffiths (changes requested):
Behaves a bit better now. :)

src/server/shell/token_authority.cpp
```diff
@@ -0,0 +1,161 @@
...
+    return *iter;
+}
+
+auto msh::TokenAuthority::is_token_valid(Token const& token) const -> bool
+{
+    std::scoped_lock lock{mutex};
+
+    // Let it pass, we'll refuse to activate when we get the token later
+    if(token == get_bogus_token())
+        return true;
```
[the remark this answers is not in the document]
tarek-y-ismail: Leftover from when it was constant. Good catch

src/server/shell/token_authority.cpp
```diff
@@ -0,0 +1,161 @@
...
+        issued_tokens.erase(iter);
+    }
+
+    // Remove the alarm that triggered this revocation, if any
+    std::ranges::remove_if(
+        revocation_alarms,
+        [](auto const& alarm)
+        {
+            return alarm->state() == time::Alarm::triggered;
+        });
```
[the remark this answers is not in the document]
tarek-y-ismail: Yup. C++ shining as an excellent foot remover as always.

src/include/server/mir/shell/token_authority.h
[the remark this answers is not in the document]
tarek-y-ismail: I already realized  that using `Synchronized<T>` wouldn't be the best idea when I introduced `revocation_alarms`. Is there anything broken that I'm missing?

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -200,68 +189,90 @@ mf::XdgActivationV1::XdgActivationV1(
...
+std::shared_ptr<mf::XdgActivationTokenData> const mf::XdgActivationV1::create_token(std::shared_ptr<ms::Session> const& session)
 {
+    if (!session_listener->is_focused_session(session))
+        return std::make_shared<XdgActivationTokenData>(token_authority->get_bogus_token(), session);
+
+    auto generated = token_authority->issue_token(
+        [this](auto const& token)
+        {
+            this->remove_token(token);
+        });
```
AlanGriffiths, line 239: Pre-existing, but worth correcting whilst moving the logic around: acquiring a token should be happening in `commit()`. (It is likely the `get_activation_token()` request is in response to the same input event as `commit()`, but it isn't guaranteed.)

AlanGriffiths (changes requested):

src/server/frontend_wayland/xdg_activation_v1.cpp
mattkae: Agreed, that is what the spec says, isn't it :)

https://wayland.app/protocols/xdg-activation-v1#xdg_activation_token_v1:request:commit

> Requests an activation token based on the different parameters that have been offered through set_serial, set_surface and set_app_id.

mattkae (changes requested):
A few things, but overall this looks sensible to me

src/server/server.cpp
```diff
@@ -464,22 +463,32 @@ auto mir::Server::open_wayland_client_socket() -> Fd
...
 
+auto mir::Server::get_activation_token() const -> std::string
+{
+    if (auto const config = self->server_config)
+    {
+        return std::string(config->the_token_authority()->issue_token({}));
+    }
+
+    BOOST_THROW_EXCEPTION(std::logic_error("Cannot open connection when not running"));
+}
```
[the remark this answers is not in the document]
tarek-y-ismail: So far I've understood that exposing something used in the core/server directly isn't such a good idea.

src/server/shell/token_authority.cpp
```diff
@@ -0,0 +1,161 @@
...
+auto msh::TokenAuthority::get_bogus_token() const -> Token
+{
+    return Token{generate_token(), {}};
+}
+
+void msh::TokenAuthority::revoke_token(Token to_remove)
+{
+    std::scoped_lock lock{mutex};
+
+    auto const iter = issued_tokens.find(Token{to_remove});
```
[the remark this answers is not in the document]
tarek-y-ismail: Remained from when `to_remove` was a string. Thanks for pointing it out.

src/server/shell/token_authority.cpp
```diff
@@ -0,0 +1,161 @@
...
+            (*cb)(*iter);
+        issued_tokens.erase(iter);
+    }
+
+    // Remove the alarm that triggered this revocation, if any
+    //
+    // Note to self: remove_if actually moves the "removed" elements to the end
+    // of the container and gives you a range to remove them yourself. Thanks
+    // C++.
+    auto removed = std::ranges::remove_if(
```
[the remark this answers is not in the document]
tarek-y-ismail: I honestly don't remember. I just merged the two arrays into an array of pairs and it works flawlessly

src/server/server.cpp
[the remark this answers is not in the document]
AlanGriffiths: I didn't make this choice, but you're right. `the_token_authority()` and the like are not thread safe and should only be used during initialization

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,80 @@
...
+    class Token
+    {
+    public:
+        using RevocationListener = std::function<void(Token const& token)>;
+
+        bool operator==(Token const& token) const;
+        explicit operator std::string() const;
+
+    private:
+        friend TokenAuthority;
```
mattkae, line 43: Why does this need to be a `friend` here? The `TokenAuthority` should have full access to the `Token`'s data members because it is in its scope, right?

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,80 @@
...
+    auto get_token_for_string(std::string const& string_token) const -> std::optional<Token>;
+    auto get_bogus_token() const -> Token;
+
+    void revoke_token(Token to_remove);
+    void revoke_all_tokens();
+
+private:
+    static auto generate_token() -> std::string;
+
+    using TimedToken = std::pair<Token, std::unique_ptr<time::Alarm>>;
```
mattkae, line 63: Why do we need this `std::pair`. Can the `time::Alarm` not live on the `Token`?

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,80 @@
...
+#include <memory>
+#include <unordered_set>
+
+namespace mir
+{
+class MainLoop;
+namespace time { class Alarm; }
+namespace shell
+{
+class TokenAuthority
```
mattkae, line 31: We could have unit tests for this class too :)

src/include/server/mir/shell/token_authority.h
```diff
@@ -0,0 +1,80 @@
...
+    private:
+        friend TokenAuthority;
+
+        Token(std::string token, std::optional<RevocationListener> revocation_listener);
+
+        std::string token;
+        std::optional<RevocationListener> revocation_listener;
+    };
+
+    TokenAuthority(std::shared_ptr<MainLoop>&& main_loop);
```
mattkae, line 51: Can be `explicit`

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -85,11 +79,12 @@ class XdgActivationV1 : public wayland::XdgActivationV1::Global
         std::shared_ptr<ms::SessionCoordinator> const& session_coordinator,
         std::shared_ptr<MainLoop> const& main_loop,
         std::shared_ptr<ObserverRegistrar<input::KeyboardObserver>> const& keyboard_observer_registrar,
-        Executor& wayland_executor);
+        Executor& wayland_executor,
+        std::shared_ptr<msh::TokenAuthority> token_authority);
```
mattkae, line 83:
```suggestion
        std::shared_ptr<msh::TokenAuthority> const& token_authority);
```

mattkae (changes requested):
A few more things that I found :)

src/include/server/mir/shell/token_authority.h
tarek-y-ismail: I believe that's not the case. I have to explicitly declare `TokenAuthority` as a friend.

src/include/server/mir/shell/token_authority.h
tarek-y-ismail: Having the `unique_ptr` in `Token` would make it non-copyable. Besides, timers are only relevant to `TokenAuthority`. Going even further, maybe `revocation_listener` should also move to `TokenAuthority` to make tokens as light as possible. What do you think?

src/include/server/mir/shell/token_authority.h
tarek-y-ismail: Maybe as another PR? This has been going on for a _long_ time

AlanGriffiths: > This has been going on for a _long_ time

We will need to support it for far _longer_, we should make that easier

tarek-y-ismail: I didn't mean to put off the work of testing (In fact, I'm writing tests right now), just saying it should be in another PR

src/include/server/mir/shell/token_authority.h
mattkae: I don't necessarily think that `revocation_listener` is bad on the token, since a token can be removed for a number of reasons.

src/server/frontend_wayland/xdg_activation_v1.cpp
```diff
@@ -85,11 +79,12 @@ class XdgActivationV1 : public wayland::XdgActivationV1::Global
...
         std::shared_ptr<ObserverRegistrar<input::KeyboardObserver>> const& keyboard_observer_registrar,
-        Executor& wayland_executor);
+        Executor& wayland_executor,
+        std::shared_ptr<msh::TokenAuthority> const& token_authority);
     ~XdgActivationV1();
 
-    std::shared_ptr<XdgActivationTokenData> const& create_token(std::shared_ptr<ms::Session> const& session);
-    std::shared_ptr<XdgActivationTokenData> try_consume_token(std::string const& token);
+    std::shared_ptr<XdgActivationTokenData> get_token_data(Token const& token);
+    std::shared_ptr<XdgActivationTokenData> const create_token(struct XdgActivationTokenV1* owner, std::shared_ptr<ms::Session> session);
```
mattkae, line 87: Why is this returning a `const` here? Clang tidy does not like it:

```
Clang-Tidy: Return type 'const std::shared_ptr<mf::XdgActivationTokenData>' is 'const'-qualified at the top level, which may reduce code readability without improving const correctness 
```

mattkae (changes requested):
A question about return types, but it is working well for me AFAICT!

src/server/frontend_wayland/xdg_activation_v1.cpp
tarek-y-ismail: Used to be a const reference. Funny that clangd or clang didn't complain about this as well

mattkae (approved):
I am happy now, but I want to leave it up to @AlanGriffiths for any final thoughts and words!

AlanGriffiths (changes requested):
More nits. But I think this does what it claims

src/server/shell/CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ set(
   surface_specification.cpp
   surface_stack_wrapper.cpp ${CMAKE_SOURCE_DIR}/src/include/server/mir/shell/surface_stack_wrapper.h
   basic_idle_handler.cpp ${CMAKE_SOURCE_DIR}/src/include/server/mir/shell/idle_handler.h
+  token_authority.cpp
```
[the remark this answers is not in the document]
AlanGriffiths: It adds the header to the project. (Can help various automations)

src/server/shell/token_authority.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    return token;
+}
+
+auto msh::TokenAuthority::get_token_for_string(std::string const& string_token) const -> std::optional<Token>
+{
+    std::scoped_lock lock{mutex};
+
+    // Incoming string doesn't have null terminator, just like the one leaving Mir
```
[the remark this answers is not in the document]
AlanGriffiths: Sorry, but the comment causes me more confusion than it saves. Surely any comment belongs where we strip the `nul` from the token? Not where it isn't expected?

src/server/shell/token_authority.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return token;
+}
+
+bool msh::TokenAuthority::Token::operator==(Token const& token) const
+{
+    return this->token == token.token;
+}
+
+msh::TokenAuthority::TokenAuthority(std::shared_ptr<MainLoop>&& main_loop) :
+    main_loop{main_loop}
```
AlanGriffiths, line 55:
```suggestion
    main_loop{std::move(main_loop)}
```

tarek-y-ismail: Borks the tests because the main loop then points to the void. Not sure why miral-app worked fine though...

tarek-y-ismail:
Had a small accident where I merged the testing PR and followed that by a force push without pulling first. Should alright now

src/server/shell/token_authority.cpp
```diff
@@ -0,0 +1,144 @@
...
```
AlanGriffiths, line 55: We shouldn't have a constructor that accepts a transfer of ownership and tests that rely on it not transferring

AlanGriffiths (changes requested):
A new last nit

src/server/shell/token_authority.cpp
tarek-y-ismail: D'oh! Const correctness bites me again...

tests/unit-tests/shell/test_token_authority.cpp
```diff
@@ -34,7 +34,7 @@ struct TestTokenAuthority : testing::Test
     TestTokenAuthority() :
         clock{std::make_shared<mir::time::SteadyClock>()},
         main_loop{std::make_shared<mir::GLibMainLoop>(clock)},
-        token_authority{std::forward<std::shared_ptr<mir::MainLoop>>(main_loop)},
+        token_authority{std::shared_ptr<mir::MainLoop>(main_loop)},
```
AlanGriffiths, line 37:
```suggestion
        token_authority{main_loop},
```

tarek-y-ismail: Well, since `main_loop` is now a const pointer, it can't be converted to a universal ref since it it will lost its constness, hence needing to copy the pointer.

AlanGriffiths: Sorry, I don't like spelling out the type explicitly, but `std::remove_const_t<decltype(main_loop)>{main_loop}` isn't better.

AlanGriffiths (approved):

PR #3705, opened 2024-12-19
smoke: increase the test timeout

Saviq:
Slower devices (RPi3) did not keep up reliably.


Saviq:
I'll make it a distro-patch actually.

PR #3707, opened 2024-12-19
feature: adding an is_above method to the SurfaceStack so that compositor authors can write tests to assert the relative z-order of surfaces

mattkae:
## What's new?

- Bumped the project minor version to 20 (new symbols)
- Added `AbstractShell::is_above`, `FocusController::is_above`, and `SurfaceStack::is_above`
- Added methods to the `WindowManagementTestHarness` to focus a window and query `is_above`
- Added two `MinimalWindowManagerTest`s that assert points about `is_above`
- Updated the symbols.map file


src/server/scene/surface_stack.cpp
```diff
@@ -493,6 +493,24 @@ void ms::SurfaceStack::send_to_back(const mir::scene::SurfaceSet &ss)
...
+    auto const lock_surface = surface.lock();
+    auto const& layer = surface_layers[lock_surface->depth_layer()];
+    for (size_t surface_index = 0; surface_index < layer.size(); surface_index++)
+    {
+        auto const& other = layer[surface_index];
+        if (surface.lock() == other)
+            return surface_index;
+    }
+
+    return {};
```
robert-ancell, line 511: Nit: This was a little unexpected, especially since 0 is returned explicitly above.

robert-ancell (approved):
Code changes look fine, will need someone else who better knows Mir architecture to clarify if there should be this z_order feature.

src/server/scene/surface_stack.cpp
mattkae: Whoops! Leftover, good catch!

mattkae:
>
> I think we would do better with an interface for making assertions:
> 
> ```c++
> EXPECT_TRUE(shell->assert_that(window1, is_above, window2));
> ```

@AlanGriffiths Do you mean this API literally, or would the shell have a method like `Shell::is_above(Window const&, Window const&)`?

AlanGriffiths:
> @AlanGriffiths Do you mean this API literally, or would the shell have a method like `Shell::is_above(Window const&, Window const&)`?

The API I suggested is more fluent (reads more like English) and allows _some_ additional predicates to be added without breaking ABI (e.g. `shares_layer_with`). But I didn't really intend it literally. HTH

src/server/symbols.map
```diff
@@ -1404,3 +1404,17 @@ global:
...
+    mir::shell::AbstractShell::z_order*;
+    mir::shell::ShellWrapper::z_order*;
+    mir::shell::SurfaceStackWrapper::z_order*;
+    non-virtual?thunk?to?mir::shell::AbstractShell::z_order*;
+    non-virtual?thunk?to?mir::shell::ShellWrapper::z_order*;
+    non-virtual?thunk?to?mir::shell::SurfaceStackWrapper::z_order*;
+    virtual?thunk?to?mir::shell::AbstractShell::z_order*;
+    virtual?thunk?to?mir::shell::ShellWrapper::z_order*;
+  };
+} MIR_SERVER_INTERNAL_2.19;
```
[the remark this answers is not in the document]
mattkae: Wouldn't we still want to bump the minor version since we would be adding symbols for the new minor version?

src/server/scene/surface_stack.cpp
```diff
@@ -493,6 +493,40 @@ void ms::SurfaceStack::send_to_back(const mir::scene::SurfaceSet &ss)
...
+        else if (shared_b == layer[surface_index])
+            b_index = surface_index;
+    }
+
+    if (!a_index.has_value())
+        return false;
+    else if (!b_index.has_value())
+        return true;
+
+    return a_index.value() > b_index.value();
```
AlanGriffiths, line 527: I don't see that iterating over the whole layer every time is significantly better than calling `std::find()` twice.

In any case, isn't this faster?

```suggestion
    bool found_b = false;
    for (auto const& entry : layer)
    {
        if (entry == b) found_b = true;
        if (entry == a) return found_b;
    }
    
    return false;
```

AlanGriffiths (approved):
Just some nitpicking

PR #3712, opened 2025-01-07
decoration-strategy-rework

AlanGriffiths:
Improves the interface between the decoration strategies and the rest of the subsystem:

1. Encapsulates the StaticGeometry
2. Moves the Button enums into the struct (with a few renames)
3. Adds some documentation
4. Simplifies buffer creation API
5. Make InputState a struct
6. Simplify use of WindowState

Generally, making the decoration strategy simpler and easier to customise

### DIFF
--- a/src/server/shell/decoration/window.cpp
+++ b/src/server/shell/decoration/window.cpp
@@ -30,36 +30,51 @@ class msd::WindowState::Self
 {
 public:
     Self(
-    std::shared_ptr<const DecorationStrategy>&& decoration_strategy,
-    std::shared_ptr<scene::Surface> const& surface,
-    float scale);
+        std::shared_ptr<scene::Surface> const& surface,
+        geometry::Height fixed_titlebar_height,
+        geometry::Width fixed_side_border_width,
+        geometry::Height fixed_bottom_border_height,
+        float scale);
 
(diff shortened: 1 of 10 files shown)


src/server/shell/decoration/window.cpp
```diff
@@ -26,42 +26,83 @@ namespace ms = mir::scene;
...
+    geometry::Width const fixed_side_border_width;
+    geometry::Height const fixed_bottom_border_height;
+    geometry::Size const window_size;
+    BorderType const border_type;
+    MirWindowFocusState const focus_state;
+    std::string const window_name;
+    float const scale;
+};
+
+msd::WindowState::Self::Self(
```
mattkae, line 49: Nit: It is a bit weird that the declaration of this class is defined in `decoration_stategy,h` instead of `window.h`

mattkae (approved):
I think this is a sensible moving around of things. I just have the one nit

src/server/shell/decoration/window.cpp
AlanGriffiths: Because this is (roughly) what we want to expose to shell devs, I was aiming to make `decoration_strategy.h` self-contained. Doesn't seem much value in an dragging an extra `window_state.h` file along.

tarek-y-ismail (approved):
Things are consolidating nicely :)

PR #3713, opened 2025-01-08
Fix one of our memory leaks

AlanGriffiths:
This leak seems to correlate best with #3710

I'm still on the track of other leaks, but pushing as it is the only strong correlation with the number of client connect/disconnect iterations


AlanGriffiths:
No! This isn't the whole story for this leak

AlanGriffiths:
In practice, this logic doesn't ever execute as we leak the rendering platform: this destructor isn't called

PR #3714, opened 2025-01-09
Don't try to refocus on an app of the same client.

### LINKED ISSUE #3309: Closing a window brings one of the remaining windows of the same client to the front

tarek-y-ismail:
Fixes #3309 

Note that this bug only occurs with apps that can have multiple windows running under the same process (example: gnome-terminal, pluma)

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -288,9 +288,6 @@ void miral::BasicWindowManager::refocus(
     if (hint && select_active_window(hint))
         return;
 
-    if (can_activate_window_for_session_in_workspace(application, workspaces_containing_window))
-        return;
-
     // Try to activate to recently active window of any application in a shared workspace
     {
         miral::Window new_focus;


src/miral/basic_window_manager.cpp
```diff
@@ -288,9 +288,6 @@ void miral::BasicWindowManager::refocus(
     if (hint && select_active_window(hint))
         return;
 
-    if (can_activate_window_for_session_in_workspace(application, workspaces_containing_window))
-        return;
```
Saviq, line 292: @AlanGriffiths this was here since you introduced MirAL… 8 years ago…

Can you comment on whether you're OK with this change?

Saviq (approved):
WFM!

src/miral/basic_window_manager.cpp
AlanGriffiths: It seems like only yesterday!

With what I now know about window management, it just looks wrong.

AlanGriffiths (approved):
The other uses for this function are suspect too: the idea of a "focussed session" was probably a mistake. But changing that is likely to unravel too much at this time

PR #3715, opened 2025-01-09
spread: Fedora 39 is gone, Alpine 3.21 is in

Saviq:


AlanGriffiths (approved):
Sure

Although... do we still want to keep alpine-3.19?

Saviq:
> Although... do we still want to keep alpine-3.19?

It's still alive until November: https://alpinelinux.org/releases/

PR #3723, opened 2025-01-20
Remove unused `InputPanelV1::display`

tarek-y-ismail:
After an update, clang-20 started marking `InputPanelV1::display` as unused, which is true as there is no need to store it. This still happens with clang-20 (20250112081750+42da12063f49-1\~exp1\~20250112081925.659)

### DIFF
--- a/src/server/frontend_wayland/input_method_v1.cpp
+++ b/src/server/frontend_wayland/input_method_v1.cpp
@@ -576,7 +576,6 @@ mf::InputPanelV1::InputPanelV1(
     OutputManager* const output_manager,
     std::shared_ptr<scene::TextInputHub> const text_input_hub)
     : Global(display, Version<1>()),
-      display{display},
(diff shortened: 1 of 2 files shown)


src/server/frontend_wayland/input_method_v1.cpp
```diff
@@ -576,7 +576,6 @@ mf::InputPanelV1::InputPanelV1(
     OutputManager* const output_manager,
     std::shared_ptr<scene::TextInputHub> const text_input_hub)
     : Global(display, Version<1>()),
-      display{display},
```
Saviq, line 579: Shouldn't the `display` argument go away, too?

tarek-y-ismail: Still being used in the global?

Saviq (approved):
If it compiles ¯\\_(ツ)_/¯.

PR #3724, opened 2025-01-21
Fix window offsetting with focus stealing prevention

### LINKED ISSUE #3695: Window offsetting shouldn't assume new windows are automatically focused
Related: #2586, #3693 

Our current implementation of window offsetting assumes that the offset is applied relative to the position of the currently focused window. This works when automatically focusing newly opened windows as each window is offset slightly from the one before it. But with focus stealing prevention, that is not the case as multiple windows could be spawned without changing focus, stacking on top of each other. 

This might potentially cause clients that don't support xdg-activation to open  behind the currently focused window, being completely occluded; or in the case of similarly sized windows,  to completely occlude the windows behind them.

A reasonable implementation should take into account all visible windows and make an attempt  to position new windows such that they don't occlude existing windows. In case this is not possible (fullscreen application, too many windows open), it would be acceptable to occlude an existing  window.

tarek-y-ismail:
Closes #3695

Adds occlusion checking when placing down new windows. The checks continue with an increasingly larger radius until no positions around the window fit in the active output **or** no possible window rectangle overlaps with the active output. The checks are also capped to 16 "iterations" so they don't go on forever, just in case.

Here's how it looks like now:

https://github.com/user-attachments/assets/6e9b8889-08d9-4919-8592-cb52fd5c0894

How it works:
- We repeatedly subtract the rectangles of existing windows from our candidate window rectangle. If after iterating through all windows (or the current window, as an early exit) we still have any visible area of the rectangle, then that placement is used. 
- We further filter placements by checking if at least a 50x50 chunk of the app is visible to prevent thin lines that could be hard to notice

### DIFF
--- a/examples/example-server-lib/floating_window_manager.cpp
+++ b/examples/example-server-lib/floating_window_manager.cpp
@@ -734,36 +734,28 @@ void FloatingWindowManagerPolicy::try_place_new_window_and_account_for_occlusion
     for (auto multiplier = 1; multiplier < max_iterations; multiplier++)
     {
         auto const offset{base_offset * multiplier};
-        std::array const positions{
-            parameters.top_left().value(),
-            parameters.top_left().value() + Displacement(offset, offset),
-            parameters.top_left().value() + Displacement(-offset, offset),
-            parameters.top_left().value() + Displacement(-offset, -offset),
-            parameters.top_left().value() + Displacement(offset, -offset)};
+        auto const top_left = parameters.top_left().value();
+        auto const size = parameters.size().value_or({});
+        std::array const test_rects{
+            geom::Rectangle{top_left, size},
+            geom::Rectangle{top_left + Displacement(offset, offset), size},
+            geom::Rectangle{top_left + Displacement(-offset, offset), size},
+            geom::Rectangle{top_left + Displacement(-offset, -offset), size},
+            geom::Rectangle{top_left + Displacement(offset, -offset), size}};
 
-        // Make sure the active output contains any test point
-        auto const all_positions_invalid = std::ranges::all_of(
-            positions,
-            [this](auto const position)
+        auto valid_test_rects = std::ranges::filter_view(
+            test_rects,
+            [this](auto const rectangle)
             {
-                return !tools.active_output().contains(position);
+                return tools.active_output().overlaps(rectangle);
             });
 
-        if (all_positions_invalid)
-            return;
-
-        for (auto const& position : positions)
+        for (auto const& test_rect : valid_test_rects)
         {
-            auto const test_rect = geom::Rectangle{position, parameters.size().value_or({})};
-
-            // Make sure the test rectangle overlaps with the current output
-            if (!tools.active_output().overlaps(test_rect))
-                continue;
-
             auto const visible_rects = get_visible_rects(test_rect, window_rects);
             if (!visible_rects.empty() && visible_area_large_enough(visible_rects))
             {
-                parameters.top_left().value() = position;
+                parameters.top_left().value() = test_rect.top_left;
                 return;
             }
         }
--- a/examples/example-server-lib/floating_window_manager.cpp
+++ b/examples/example-server-lib/floating_window_manager.cpp
@@ -16,6 +16,7 @@
 
 #include "floating_window_manager.h"
 #include "decoration_provider.h"
+#include "mir_toolkit/common.h"
 #include "miral/minimal_window_manager.h"
 #include "miral/window_specification.h"
 
@@ -675,17 +676,17 @@ void FloatingWindowManagerPolicy::try_place_new_window_and_account_for_occlusion
             return std::vector(valid_rects.cbegin(), valid_rects.cend());
         };
 
-        auto const occluded_by_visible_windows =
-            [subtract_rectangles](geometry::Rectangle initial_test_rect, std::vector<geometry::Rectangle> window_rects)
-        {
-            using RectangleSet = std::unordered_set<
+        using RectangleSet = std::unordered_set<
                     geometry::Rectangle,
                     decltype([](geometry::Rectangle const& s) {
                         return std::hash<int>{}(s.size.height.as_int() * 31 + s.size.width.as_int());
                     })>;
 
+        auto const get_visible_rects =
+            [subtract_rectangles](geometry::Rectangle initial_test_rect, std::vector<geometry::Rectangle> window_rects)
+        {
             if (window_rects.empty())
-                return false;
+                return RectangleSet{initial_test_rect};
 
             // Starting with rectangle we're trying to place, repeatedly
             // subtract all visible rectangles from it, accumulating
@@ -707,10 +708,25 @@ void FloatingWindowManagerPolicy::try_place_new_window_and_account_for_occlusion
                 test_rects = std::move(iter_output_rects);
 
                 if (test_rects.empty())
-                    return true;
+                    return test_rects;
             }
 
-            return test_rects.empty();
+            return test_rects;
+        };
+
+        auto const visible_area_large_enough = [](RectangleSet const& visible_rects)
+        {
+            auto constexpr min_visible_width = 50;
+            auto constexpr min_visible_height = 50;
+
+            return std::ranges::any_of(
+                visible_rects,
+                [min_visible_width, min_visible_height](auto const& rect)
+                {
+                    return rect.size.width.as_int() >= min_visible_width &&
+                           rect.size.height.as_int() >= min_visible_height;
+                });
+
         };
 
         // Try place the window around the suggested position
@@ -744,7 +760,8 @@ void FloatingWindowManagerPolicy::try_place_new_window_and_account_for_occlusion
                 if (!tools.active_output().overlaps(test_rect))
                     continue;
 
-                if (!occluded_by_visible_windows(test_rect, window_rects))
+                auto const visible_rects = get_visible_rects(test_rect, window_rects);
+                if (!visible_rects.empty() && visible_area_large_enough(visible_rects))
                 {
                     parameters.top_left().value() = position;
                     return;
--- a/examples/example-server-lib/floating_window_manager.cpp
+++ b/examples/example-server-lib/floating_window_manager.cpp
@@ -16,9 +16,6 @@
 
 #include "floating_window_manager.h"
 #include "decoration_provider.h"
-#include "mir_toolkit/common.h"
-#include "miral/minimal_window_manager.h"
-#include "miral/window_specification.h"
 
 #include <miral/application_info.h>
 #include <miral/internal_client.h>
@@ -26,14 +23,9 @@
 #include <miral/window_info.h>
 #include <miral/window_manager_tools.h>
 #include <miral/zone.h>
-#include <mir/geometry/rectangle.h>
 
 #include <linux/input.h>
 #include <csignal>
-#include <algorithm>
-#include <ranges>
-#include <unordered_set>
-#include <functional>
 
 using namespace miral;
 using namespace miral::toolkit;
@@ -480,8 +472,6 @@ WindowSpecification FloatingWindowManagerPolicy::place_new_window(
 {
     auto parameters = MinimalWindowManager::place_new_window(app_info, request_parameters);
 
-    try_place_new_window_and_account_for_occlusion(parameters);
-
     if (app_info.application() == decoration_provider->session())
     {
         parameters.type() = mir_window_type_decoration;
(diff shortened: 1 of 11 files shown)


tarek-y-ismail:
There's a bug where a 1280x1024 fullscreen window is returned from `for_each_window_in_workspace`. It's not clear to me what this window is. My guess is either the background or something leftover from the spinner. Will investigate next week.

AlanGriffiths:
> There's a bug where a 1280x1024 fullscreen window is returned from `for_each_window_in_workspace`. It's not clear to me what this window is. My guess is either the background or something leftover from the spinner. Will investigate next week.

Check which layer it is in. You're likely seeing the background

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -472,6 +481,9 @@ WindowSpecification FloatingWindowManagerPolicy::place_new_window(
 {
     auto parameters = MinimalWindowManager::place_new_window(app_info, request_parameters);
 
+    if(focus_stealing == FocusStealing::prevent)
+        try_place_new_window_and_account_for_occlusion(parameters);
```
Saviq, line 485: Focus stealing prevention shouldn't be a factor when placing new windows, why wouldn't you just use the same in both cases?

Saviq (changes requested):
This looks like a piece of code that could use and would yield nicely to some unit testing?

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -605,3 +617,159 @@ void FloatingWindowManagerPolicy::handle_modify_window(WindowInfo& window_info,
...
+            for (auto const& window_rect : window_rects)
+            {
+                RectangleSet iter_output_rects;
+                for (auto const& test_rect : test_rects)
+                {
+                    auto const window_output_rects = subtract_rectangles(test_rect, window_rect);
+                    iter_output_rects.insert(window_output_rects.begin(), window_output_rects.end());
+                }
+
+                test_rects = std::move(iter_output_rects);
```
mattkae, line 708: Why do you need `iter_output_rects` at all if you're seemingly just setting `test_rects` to it every time? Are we actually "accumulating" here?

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -605,3 +617,159 @@ void FloatingWindowManagerPolicy::handle_modify_window(WindowInfo& window_info,
...
+                window_rects.emplace_back(window.top_left(), window.size());
+            }
+        );
+
+        // Worst case scenario: you try taking a hole out of the middle of
+        // a rect, results in 4 rectangles. Every other case is a special
+        // case of this one where the width or height of the resulting
+        // rects are < 0.
+        auto const subtract_rectangles = [](geometry::Rectangle rect,
+                                            geometry::Rectangle hole) -> std::vector<geometry::Rectangle>
```
mattkae, line 646:
```suggestion
        auto const subtract_rectangles = [](geometry::Rectangle first,
                                            geometry::Rectangle second) -> std::vector<geometry::Rectangle>
```

Is the name `hole` a poor one here?

mattkae (changes requested):
I have a few initial items. This whole thing seems quite complicated and potentially inefficient though, which scares me a bit :smile:

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -605,3 +617,159 @@ void FloatingWindowManagerPolicy::handle_modify_window(WindowInfo& window_info,
...
+            if (all_positions_invalid)
+                return;
+
+            for (auto const& position : positions)
+            {
+                auto const test_rect = geometry::Rectangle{position, parameters.size().value_or({})};
+
+                // Make sure the test rectangle overlaps with the current output
+                if (!tools.active_output().overlaps(test_rect))
+                    continue;
```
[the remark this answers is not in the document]
tarek-y-ismail: Hmm, I think just testing the rectangles alone could be enough. So `all_positions_invalid` can be skipped altogether

examples/example-server-lib/floating_window_manager.cpp
tarek-y-ismail: Once a set of test rectangles is iterated through, we don't need it, we only need the new  set of test rectangles resulting from subtracting all test windows from the current window.

As an example, if you have a window in the center of the screen and a test rectangle slightly offset to the bottom right, once you subtract the window from the test rectangle, you'll have two new test rectangles, one wide and one long. The old one can be discarded in favor of the new ones. If you have more windows, you can use the new set of test rectangles and subtract from those, only keeping the resulting test rectangles.

examples/example-server-lib/floating_window_manager.cpp
tarek-y-ismail: It made more sense at the time, you're subtracting a _hole_ from a _rect_. Of course, this naming only makes sense in the worst case scenario (the hole is wholly contained within the rect), but all other cases can be derived from this base case.

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -605,3 +617,159 @@ void FloatingWindowManagerPolicy::handle_modify_window(WindowInfo& window_info,
...
+            // Make sure the active output contains any test point
+            auto const all_positions_invalid = std::ranges::all_of(
+                positions,
+                [this](auto const position)
+                {
+                    return !tools.active_output().contains(position);
+                });
+
+            if (all_positions_invalid)
+                return;
```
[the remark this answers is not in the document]
tarek-y-ismail: Check my answer to the next comment

examples/example-server-lib/floating_window_manager.cpp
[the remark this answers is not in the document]
tarek-y-ismail: To elaborate further on this comment with some context from the previous one, the reason we're not checking against other outputs is that there is no way to access them. So we only check against the current output. Besides, I'm not sure it's a good idea to move windows between outputs even if there's no space.

AlanGriffiths (changes requested):
Leaving aside the logic introduced here: why is it being added to example code and not to `MinimalWindowManager`?

Assuming we can agree the placement logic, surely it is desirable for any "floating" shell based on MirAL?

(Not objection to iterating on the logic here, I'm just questioning whether this example is what we want to land)

examples/example-server-lib/floating_window_manager.cpp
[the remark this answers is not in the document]
tarek-y-ismail: One thing I should also mention is that `WindowManagerTools` doesn't have a direct way of iterating over outputs.

examples/example-server-lib/floating_window_manager.cpp
tarek-y-ismail: Something very weird is happening. The tests should be failing, but they aren't. I spent some time investigating and it seems that my way of collecting visible windows isn't working properly in tests. 

When the second window is being created, `for_each_workspace_containing(active_window)` doesn't actually return _any_ workspaces, although the first window should be in one by then.

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -16,16 +16,22 @@
 
 #include "floating_window_manager.h"
 #include "decoration_provider.h"
+#include "mir_toolkit/common.h"
+#include "miral/minimal_window_manager.h"
+#include "miral/window_specification.h"
```
AlanGriffiths, line 21: These are not local headers, so use `<...>` includes (and they belong in the block of Mir headers)

AlanGriffiths (changes requested):
It doesn't look like _any_ of the `floating_window_manager.{h,cpp}` changes are needed for this PR

examples/example-server-lib/floating_window_manager.cpp
AlanGriffiths: Have you put the window into any workspaces?

src/miral/minimal_window_manager.cpp
```diff
@@ -655,3 +663,160 @@ bool miral::MinimalWindowManager::Impl::advise_new_window(WindowInfo const& info
...
+                    geom::Rectangle,
+                    decltype([](geom::Rectangle const& s) {
+                        return std::hash<int>{}(s.size.height.as_int() * 31 + s.size.width.as_int());
+                    })>;
+
+    if (parameters.state() != mir_window_state_restored)
+        return;
+
+    // Assumption, we're opening the new window in the same workspace as the
+    // active window
```
AlanGriffiths, line 681: I don't think this is a reasonable assumption.

There are no guarantees about how, or if, the shell is using workspaces.

examples/example-server-lib/floating_window_manager.cpp
tarek-y-ismail: Turns out _my_ expectations were wrong. Since we now reject placements which are too small, the second window is placed in a position that lets the test I was expecting to fail to pass.

src/miral/minimal_window_manager.cpp
tarek-y-ismail: Just pushed some changes to how windows are collected. No assumptions made anymore.

examples/example-server-lib/floating_window_manager.h
```diff
@@ -120,7 +120,6 @@ class FloatingWindowManagerPolicy : public miral::MinimalWindowManager
     std::shared_ptr<miral::Workspace> active_workspace;
     std::map<int, std::shared_ptr<miral::Workspace>> key_to_workspace;
     std::map<std::shared_ptr<miral::Workspace>, miral::Window> workspace_to_active;
-
```
AlanGriffiths, line 123: Unnecessary whitespace change

src/miral/minimal_window_manager.cpp
```diff
@@ -14,13 +14,19 @@
...
 #include <miral/toolkit_event.h>
 #include <miral/application_info.h>
 #include "application_selector.h"
+#include "mir/geometry/forward.h"
+
 #include <linux/input.h>
 #include <gmpxx.h>
 
+#include <ranges>
+#include <unordered_set>
```
AlanGriffiths, line 28: We tend to local headers (with `"`), Mir headers (with `<`), system headers (with `<`) and c++ headers (with `<`).

Also, it seems unlikely to need `mir/geometry/forward.h` in an implementation file.

Vis:

```suggestion
#include "application_selector.h"

#include <miral/minimal_window_manager.h>
#include <miral/toolkit_event.h>
#include <miral/application_info.h>

#include <linux/input.h>
#include <gmpxx.h>

#include <algorithm>
#include <ranges>
#include <unordered_set>
```

AlanGriffiths (changes requested):
I still haven't got my head around the logic (too many distractions this week). But a couple of hygiene quibbles

mattkae (changes requested):
This makes better sense to me this morning :100:

src/miral/minimal_window_manager.cpp
```diff
@@ -655,3 +666,166 @@ bool miral::MinimalWindowManager::Impl::advise_new_window(WindowInfo const& info
...
+            [min_visible_width, min_visible_height](auto const& rect)
+            {
+                return rect.size.width.as_int() >= min_visible_width && rect.size.height.as_int() >= min_visible_height;
+            });
+    };
+
+    // First, check if the given rectangle (if any) is good enough
+    auto const initial_top_left = parameters.top_left().value_or({});
+    auto const size = parameters.size().value_or({});
+    auto const initial_rectangle = geom::Rectangle{initial_top_left, size};
```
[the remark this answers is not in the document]
tarek-y-ismail: Feels quite a bit harder to read, what do you say we keep it as is?

```diff
-    auto const initial_top_left = parameters.top_left().value_or({});
-    auto const size = parameters.size().value_or({});
-    auto const initial_rectangle = geom::Rectangle{initial_top_left, size};
+    auto const initial_rectangle = geom::Rectangle{parameters.top_left().value_or({}), parameters.size().value_or({})};
     if (auto initial_position_visible_rects = get_visible_rects(initial_rectangle, window_rects);
         !initial_position_visible_rects.empty() && visible_area_large_enough(initial_position_visible_rects))
         return;
@@ -803,10 +801,10 @@ void miral::MinimalWindowManager::Impl::try_place_new_window_and_account_for_occ
     {
         auto const offset{base_offset * multiplier};
         std::array const test_rects{
-            geom::Rectangle{initial_top_left + Displacement(offset, offset), size},
-            geom::Rectangle{initial_top_left + Displacement(-offset, offset), size},
-            geom::Rectangle{initial_top_left + Displacement(-offset, -offset), size},
-            geom::Rectangle{initial_top_left + Displacement(offset, -offset), size}};
+            geom::Rectangle{initial_rectangle.top_left + Displacement(offset, offset), initial_rectangle.size},
+            geom::Rectangle{initial_rectangle.top_left + Displacement(-offset, offset), initial_rectangle.size},
+            geom::Rectangle{initial_rectangle.top_left + Displacement(-offset, -offset), initial_rectangle.size},
+            geom::Rectangle{initial_rectangle.top_left + Displacement(offset, -offset), initial_rectangle.size}};

```

src/miral/minimal_window_manager.cpp
```diff
@@ -655,3 +666,166 @@ bool miral::MinimalWindowManager::Impl::advise_new_window(WindowInfo const& info
...
+    if (auto initial_position_visible_rects = get_visible_rects(initial_rectangle, window_rects);
+        !initial_position_visible_rects.empty() && visible_area_large_enough(initial_position_visible_rects))
+        return;
```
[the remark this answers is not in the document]
tarek-y-ismail: Best we can do at the moment is:
```diff
-    if (auto initial_position_visible_rects = get_visible_rects(initial_rectangle, window_rects);
-        !initial_position_visible_rects.empty() && visible_area_large_enough(initial_position_visible_rects))
-        return;
+
+    {
+        auto const initial_position_visible_rects = get_visible_rects(initial_rectangle, window_rects);
+        if (!initial_position_visible_rects.empty() && visible_area_large_enough(initial_position_visible_rects))
+            return;
+    }
```

src/miral/minimal_window_manager.cpp
```diff
@@ -655,3 +666,166 @@ bool miral::MinimalWindowManager::Impl::advise_new_window(WindowInfo const& info
...
+
+    // Try place the window around the suggested position
+    auto constexpr max_iterations{16};
+    auto constexpr base_offset{48};
```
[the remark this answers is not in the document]
tarek-y-ismail: Wouldn't say it really matters since you're depending on the placement of existing windows, which when accounting for user input (moving windows, resizing them) is effectively random

src/miral/minimal_window_manager.cpp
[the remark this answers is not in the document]
mattkae: I prefer either option over the current one. The reason being that the current implementation makes it annoying to set proper breakpoints in my debugger. I've found that doing too much in a conditional clause always leads to trouble and lack of readability down the road.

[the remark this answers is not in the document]
tarek-y-ismail: I think this has been somewhat remedied by moving some of the code to `is_occluded`.

include/miral/miral/minimal_window_manager.h
```diff
@@ -118,6 +118,11 @@ class MinimalWindowManager : public WindowManagementPolicy
     struct Impl;
     Impl* const self;
 };
+
+auto is_occluded(
+    mir::geometry::Rectangle test_rectangle,
+    std::vector<mir::geometry::Rectangle> const& occluding_rectangles,
+    mir::geometry::Size min_visible_size) -> bool;
```
mattkae, line 125: Why are we making this method public? Just for tests? If it is an implementation detail, I would be in favor of not exposing it in the API

mattkae (changes requested):
I am personally against exposing symbols for test only, as it unnecessarily adds to the API. Perhaps we can just test `try_place_new_window_and_account_for_occlusion` without exposing a new method?

include/miral/miral/minimal_window_manager.h
mattkae: Why would that chunk prevent you from testing what you want? Could you not write the test in: `tests/window_management_tests/test_minimal_window_manager.cpp`?

tarek-y-ismail: That's what I tried at first, since you mentioned it in a comment above. Whenever I used `create_window` without setting the window state (no case in the chunk matches what I want), the code would execute the default case, basically discarding whatever placement I defined. So it would be really difficult to write tests that way

AlanGriffiths: I see no reason to expose the function for testing. That's what `miral-test-internal` is for

src/miral/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ add_library(miral-internal STATIC
                                          join_client_threads.h
                                          window_info_defaults.h
     window_specification_internal.cpp    window_specification_internal.h
+    is_occluded.cpp			 is_occluded.h
```
mattkae, line 35: Nit: Can you align this with the rest of them :)

src/miral/is_occluded.h
```diff
@@ -0,0 +1,32 @@
...
+#define MIRAL_IS_OCCLUDED_H
+
+#include "mir/geometry/forward.h"
+
+#include <vector>
+
+namespace miral
+{
+auto is_occluded(
+    mir::geometry::Rectangle test_rectangle,
```
mattkae, line 27:
```suggestion
    mir::geometry::Rectangle const& test_rectangle,
```

src/miral/minimal_window_manager.cpp
```diff
@@ -147,7 +156,9 @@ auto miral::MinimalWindowManager::place_new_window(
     ApplicationInfo const& /*app_info*/, WindowSpecification const& requested_specification)
     -> WindowSpecification
 {
-    return requested_specification;
+    auto spec_copy = requested_specification;
+    self->try_place_new_window_and_account_for_occlusion(spec_copy);
+    return spec_copy;
```
mattkae, line 161: Nit: I would rather if `try_place_new_window_and_account_for_occlusion` would take a `WindowSpecification const&` as a parameter and return a `WindowSpecification` so that you can do:

```suggestion
    return  self>try_place_new_window_and_account_for_occlusion(requested_specification);
```

src/miral/minimal_window_manager.cpp
```diff
@@ -655,3 +666,77 @@ bool miral::MinimalWindowManager::Impl::advise_new_window(WindowInfo const& info
...
+void miral::MinimalWindowManager::Impl::try_place_new_window_and_account_for_occlusion(WindowSpecification& parameters)
+{
+    if (parameters.state() != mir_window_state_restored)
+        return;
+
+    std::vector<geom::Rectangle> window_rects;
+    tools.for_each_application(
+        [&window_rects, this](ApplicationInfo& app_info)
+        {
+            auto& app_windows = app_info.windows();
```
mattkae, line 679:
```suggestion
            auto const& app_windows = app_info.windows();
```

src/miral/is_occluded.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef MIRAL_IS_OCCLUDED_H
+#define MIRAL_IS_OCCLUDED_H
+
+#include "mir/geometry/forward.h"
+
+#include <vector>
+
+namespace miral
+{
+auto is_occluded(
```
mattkae, line 26:
```suggestion
/// Returns true if [test_rectangle] is hidden by the provided list
/// of [occluding_rectangles]. By "hidden", we mean that there does
/// not exist a visible area of [test_rectangle] that is bigger
/// than [min_visible_size] in both width and height.
auto is_occluded(
```

mattkae (changes requested):
I have some final nits, but I think I won't have any more problems if we land them :) I think the algorithm is maybe _too_ clever, but we will see how it performs under stress in a real environment.

src/miral/CMakeLists.txt
tarek-y-ismail: Huh? It looks properly indented in the web interface and my editor

AlanGriffiths: Don't use tabs?

tarek-y-ismail:
> I think the algorithm is maybe too clever

I agree, but it was the most "grug" solution to the occlusion issue that I had when I started. I'm open to experimentation though

AlanGriffiths:
> I agree, but it was the most "grug" solution to the occlusion issue that I had when I started. I'm open to experimentation though

The obvious way to experiment is to start with documenting some scenarios by writing them as tests. And then write the simplest code needed to pass them.

mattkae (changes requested):
Just the remaining poor placement problem

tarek-y-ismail:
> @tarek-y-ismail I have no more code problems, but I seem to be getting some weird placements now

Hmm, I get different behavior with kgx and gnome-terminal. They don't seem to request a specific size on the initial window placement request. I suppose I could use the old code, but I'm starting to see the cracks...

mattkae:
> > @tarek-y-ismail I have no more code problems, but I seem to be getting some weird placements now
> 
> Hmm, I get different behavior with kgx and gnome-terminal. They don't seem to request a specific size on the initial window placement request. I suppose I could use the old code, but I'm starting to see the cracks...
> 
> Edit: Another thing I've encountered is that `WindowManagementTools::window_at` returns the first window which contains the point given. Now in the context of the placement problem, this isn't enough since there might be two windows, one at the position we're testing, but its occluded behind another one that overlaps it but doesn't share the same top left. `window_at` will return the one on top, which causes us to possibly place the new window on top of the one in the back. I think another variant `window_at_exact` could be introduced that only returns a window if the top left matches the given point. This would allow me to somewhat handle the case with no provided size.
> 
> Now, again. I'm starting to think that my current approach might be too over-engineered. Maybe like what @mattkae said, we could use a simpler way that follows a predetermined pattern?
> 
> Edit 2: So, I investigated a bit further. We could go around the issue of apps not submitting their size at window creation time by checking in `modify_surface` if the current size is (0,0) and the modifications contain a new size. We can then use the new placement code to modify the window's placement. The only issue so far is that the window shows up in the original position for a frame before being properly positioned. I'm trying to figure out where exactly is the new size submitted and where the window is shown. If `modify_surface` is called between those two points, the window should start at the proper position.

Oof I didn't realize how much of a web this is. Given all that, my vote is just to iterate down the position of sizes until we find one that is not currently take up by a window. In my opinion, occlusion doesn't matter so much, as clients should be xdg activated. If they aren't, then that's their problem :smile:

tarek-y-ismail:
> Oof I didn't realize how much of a web this is. Given all that, my vote is just to iterate down the position of sizes until we find one that is not currently take up by a window. In my opinion, occlusion doesn't matter so much, as clients should be xdg activated. If they aren't, then that's their problem 😄

Yeah, I think I need to rework this in my head. I'll mark this as a draft until I iron out the remaining issues so you all can review the other PRs.

My main goal with this PR was to have a solution for clients that didn't implement activation yet (which from memory were the majority). I'm kinda torn between letting them stay broken and sunk cost fallacy-ing myself into merging this PR. 

The only visible issue at the moment is the one frame of flickering between the initial and "fixed" placement.

PR #3726, opened 2025-01-23
Ctrl+Alt+{t,T,x,X} now prevent other event filters from running.

tarek-y-ismail:
This fixes terminals failing to activate themselves due to the key listener immediately revoking their token in response to the terminal launch shortcut being pressed.


AlanGriffiths (approved):
Oops!

PR #3727, opened 2025-01-23
Disable token creation for a bit on keyboard input

tarek-y-ismail:
Fixes apps being activated even while typing.

To test:
1. Run miral-app
2. Open a terminal
3. Run `pluma &`
4. Run `sleep 2 && pluma`
5. Start typing on your keyboard

Without this fix, pluma manages to request an activation token and request activation with this token between keystrokes. This PR adds a small window (100 ms) after each keystroke where token creation is disabled.

Note: To properly test, the session check [here](
https://github.com/canonical/mir/pull/3727/files#diff-b445f58ec25aa464dc1b22558007b0ae65356f06b5e33c70288f3c68b0cc7000R243) should be removed. We're still discussing whether this check should be included or not as it doesn't make sense for unfocused applications to be able to request tokens and activate themselves, but we haven't seen any applications that behave as expected so far.

### DIFF
--- a/src/server/frontend_wayland/xdg_activation_v1.cpp
+++ b/src/server/frontend_wayland/xdg_activation_v1.cpp
@@ -25,6 +25,7 @@
 #include "mir/scene/session_listener.h"
 #include "mir/shell/shell.h"
 #include "mir/shell/token_authority.h"
+#include "mir/time/alarm.h"
 #include "mir/wayland/client.h"
 #include "mir/wayland/protocol_error.h"
 #include "mir/log.h"
@@ -37,6 +38,7 @@
 #include <ranges>
 #include <string>
 #include <vector>
+#include <atomic>
 
 namespace mf = mir::frontend;
 namespace mw = mir::wayland;
@@ -50,6 +52,9 @@ namespace mir
 {
 namespace frontend
 {
+
+auto constexpr token_creation_delay = std::chrono::milliseconds{100};
+
 struct XdgActivationTokenData
 {
     XdgActivationTokenData(
@@ -81,6 +86,7 @@ public:
 
     std::shared_ptr<XdgActivationTokenData> get_token_data(Token const& token);
     std::shared_ptr<XdgActivationTokenData> const create_token(std::shared_ptr<ms::Session> const& session);
+    std::atomic<bool> can_create_tokens{true};
     void invalidate_all();
     void invalidate_if_not_from_session(std::shared_ptr<ms::Session> const&);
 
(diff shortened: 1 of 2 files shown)


Saviq:
> 3. Run `pluma &`
> 4. Run `sleep 2 && pluma`
> 5. Start typing on your keyboard

All this within the terminal?

> Without this fix, pluma manages to request an activation token and request activation with this token between keystrokes. This PR adds a small window (100 ms) after each keystroke where token creation is disabled.

> Note: To properly test, the session check [here](https://github.com/canonical/mir/pull/3727/files#diff-b445f58ec25aa464dc1b22558007b0ae65356f06b5e33c70288f3c68b0cc7000R243) should be removed.

That's wrong. Pluma is not focused, it can't get a valid token. There's just no world where that makes sense :shrug:. It would mean apps in background could activate themselves completely on its own. Why tokens, then?

I don't think this PR makes sense, typing in the terminal post `sleep 2 && pluma` should invalidate any tokens the terminal passed on, there's no timeout needed.

tarek-y-ismail:
> All this within the terminal?

Yes

> That's wrong. Pluma is not focused, it can't get a valid token

Yes, but when you actually discovered the bug, the code to check if the application/session requesting the token was different from the focused one didn't exist on the branch you were building from.

> I don't think this PR makes sense, typing in the terminal post sleep 2 && pluma should invalidate any tokens the terminal passed on, there's no timeout needed.

The bug isn't related to _who_ is asking for the token, it's related to _when_. The whole roundtrip happens between key presses. So we don't have a chance to invalidate tokens.

AlanGriffiths:
I too don't understand what this is supposed to address:

1. An unfocussed application requesting an activation token is pointless, as it is unfocussed it should get a "bogus token" that cannot be used by anyone

2. An application requesting and using a token itself is only makes sense if one of its windows still has focus when the use happens. Otherwise the token will have been revoked

None of that has to with when the request happens

Saviq:
> The bug isn't related to who is asking for the token, it's related to when.

OK so what you're saying is that when the `sleep 2` runs out, the terminal gets a separate token for the `&& pluma` and pluma uses that to activate, all between keypresses?

TBH feels to me like the whole `sleep 2 && pluma` should have just one token - so I'd say that's a bug in the terminal implementation that we shouldn't be working around on the compositor side...?

AlanGriffiths:
Oh, so is it the terminal that requests the token?

I guess the meaningful key event for `sleep 2 && pluma` is the `enter`. The problem being that, during the sleep, we might see more keypresses and invalidate the token?

Saviq:
> The problem being that, during the sleep, we might see more keypresses and invalidate the token?

Not a bug, feature :)

PR #3728, opened 2025-01-27
Add a basic test suite for `TokenAuthority`

tarek-y-ismail:
Just the tests I could come up with as I wrote. Feel free to provide edge cases or other test cases.

### DIFF
--- a/tests/unit-tests/shell/test_token_authority.cpp
+++ b/tests/unit-tests/shell/test_token_authority.cpp
@@ -1,0 +1,178 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 or 3 as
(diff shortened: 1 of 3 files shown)


tests/unit-tests/shell/test_token_authority.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    ASSERT_FALSE(revoked1);
+    ASSERT_FALSE(revoked2);
+
+    token_authority.revoke_token(token1);
+
+    auto empty_token_opt = token_authority.get_token_for_string(std::string(token1));
+    ASSERT_FALSE(empty_token_opt.has_value());
+    ASSERT_TRUE(revoked1);
+    ASSERT_FALSE(revoked2);
```
mattkae, line 140: Nit: Should we assert that the `token2` is still present as well?

mattkae (approved):
One small nit pick, but I sure do like tests :)

PR #3730, opened 2025-01-28
Make key repeat and delay configurable

tarek-y-ismail:
How to test:
- Create a file at `~/.config/mir_demo_server.input` with the following content:
```
repeat_rate=25
repeat_delay=600
```
- Run mir demo server
- Run a wayland compatible text editor or terminal (`kgx`), or and X application (`xterm`)
- press and hold any key
- Change `repeat_rate` to a higher or lower value and watch as the repeat rate speeds or slows down
- Change `repeat_delay` to a higher or lower value and watch as the duration you need to hold the key down increases or decreases.

### DIFF
--- a/examples/mir_demo_server/server_example.cpp
+++ b/examples/mir_demo_server/server_example.cpp
@@ -121,15 +121,15 @@ try
     miral::MirRunner runner{argc, argv, "mir/mir_demo_server.config"};
 
     miral::InputConfiguration input_configuration;
+    auto mouse = input_configuration.mouse();
+    auto touchpad = input_configuration.touchpad();
+    auto keyboard = input_configuration.keyboard();
 
     miral::ConfigFile test{runner, "mir_demo_server.input", miral::ConfigFile::Mode::reload_on_change,
-        [&input_configuration](auto& in, auto path)
+        [&](auto& in, auto path)
     {
         std::cout << "** Reloading: " << path << std::endl;
 
-        auto mouse = input_configuration.mouse();
-        auto touchpad = input_configuration.touchpad();
-        auto keyboard = input_configuration.keyboard();
 
         for (std::string line; std::getline(in, line);)
         {
@@ -189,6 +189,14 @@ try
         input_configuration.touchpad(touchpad);
         input_configuration.keyboard(keyboard);
     }};
+
+    runner.add_start_callback(
+        [&]
+        {
+            input_configuration.mouse(mouse);
+            input_configuration.touchpad(touchpad);
+            input_configuration.keyboard(keyboard);
+        });
 
     runner.set_exception_handler(exception_handler);
 
--- a/examples/mir_demo_server/server_example.cpp
+++ b/examples/mir_demo_server/server_example.cpp
@@ -18,6 +18,7 @@
 #include "server_example_input_filter.h"
 #include "server_example_test_client.h"
 
+#include <functional>
 #include <miral/cursor_theme.h>
 #include <miral/display_configuration_option.h>
 #include <miral/input_configuration.h>
@@ -40,6 +41,7 @@
 #include <chrono>
 #include <cstdlib>
 #include <iostream>
+#include <mutex>
 
 namespace mir { class AbnormalExit; }
 
@@ -124,12 +126,21 @@ try
     auto mouse = input_configuration.mouse();
     auto touchpad = input_configuration.touchpad();
     auto keyboard = input_configuration.keyboard();
+    std::mutex config_mutex;
+
+    auto config_applier = [&] mutable
+    {
+        input_configuration.mouse(mouse);
+        input_configuration.touchpad(touchpad);
+        input_configuration.keyboard(keyboard);
+    };
 
     miral::ConfigFile test{runner, "mir_demo_server.input", miral::ConfigFile::Mode::reload_on_change,
         [&](auto& in, auto path)
     {
         std::cout << "** Reloading: " << path << std::endl;
 
+        std::lock_guard lock{config_mutex};
 
         for (std::string line; std::getline(in, line);)
         {
@@ -185,17 +196,14 @@ try
             }
         }
 
-        input_configuration.mouse(mouse);
-        input_configuration.touchpad(touchpad);
-        input_configuration.keyboard(keyboard);
+        config_applier();
     }};
 
     runner.add_start_callback(
         [&]
         {
-            input_configuration.mouse(mouse);
-            input_configuration.touchpad(touchpad);
-            input_configuration.keyboard(keyboard);
+            std::lock_guard lock{config_mutex};
+            config_applier();
         });
 
     runner.set_exception_handler(exception_handler);
--- a/examples/mir_demo_server/server_example.cpp
+++ b/examples/mir_demo_server/server_example.cpp
@@ -33,6 +33,7 @@
 #include "mir/options/option.h"
 #include "mir/report_exception.h"
 #include "mir/server.h"
+#include "mir/log.h"
 
 #include <boost/exception/diagnostic_information.hpp>
 
@@ -128,6 +129,7 @@ try
 
         auto mouse = input_configuration.mouse();
         auto touchpad = input_configuration.touchpad();
+        auto keyboard = input_configuration.keyboard();
 
         for (std::string line; std::getline(in, line);)
         {
@@ -146,9 +148,46 @@ try
                 touchpad.scroll_mode(mir_touchpad_scroll_mode_edge_scroll);
             if (line == "mir_touchpad_scroll_mode_button_down_scroll")
                 touchpad.scroll_mode(mir_touchpad_scroll_mode_button_down_scroll);
+
+            if (line.contains("="))
+            {
+                auto const eq = line.find_first_of("=");
+                auto const key = line.substr(0, eq);
+                auto const value = line.substr(eq+1);
+
+                auto const parse_and_validate = [](std::string const& key, std::string_view val) -> std::optional<int>
+                {
+                    auto const int_val = std::atoi(val.data());
+                    if (int_val < 0)
+                    {
+                        mir::log_warning(
+                            "Config value %s does not support negative values. Ignoring the supplied value (%d)...",
+                            key.c_str(), int_val);
+                        return std::nullopt;
+                    }
+
+                    return int_val;
+                };
+
+                if (key == "repeat_rate")
+                {
+                    auto const parsed = parse_and_validate(key, value);
+                    if (parsed)
+                        keyboard.set_repeat_rate(*parsed);
+                }
+
+                if (key == "repeat_delay")
+                {
+                    auto const parsed = parse_and_validate(key, value);
+                    if (parsed)
+                        keyboard.set_repeat_delay(*parsed);
+                }
+            }
         }
+
         input_configuration.mouse(mouse);
         input_configuration.touchpad(touchpad);
+        input_configuration.keyboard(keyboard);
     }};
 
     runner.set_exception_handler(exception_handler);
--- a/src/include/server/mir/shell/accessibility_manager.h
+++ b/src/include/server/mir/shell/accessibility_manager.h
@@ -18,6 +18,7 @@
 #define MIR_SHELL_ACCESSIBILITY_MANAGER_H
 
 #include <memory>
+#include <optional>
 #include <vector>
 
 namespace mir
@@ -30,11 +31,12 @@ class AccessibilityManager
 public:
     void register_keyboard_helper(std::shared_ptr<shell::KeyboardHelper>);
 
-    int repeat_rate() const;
+    std::optional<int> repeat_rate() const;
     int repeat_delay() const;
 
     void repeat_rate(int new_rate);
     void repeat_delay(int new_rate);
+    void override_key_repeat_settings(bool enable);
 
     void notify_helpers() const;
 
@@ -44,6 +46,7 @@ private:
     // 25 rate and 600 delay are the default in Weston and Sway
     int repeat_rate_{25};
     int repeat_delay_{600};
+    bool enable_key_repeat{true};
 };
 }
 }
--- a/src/include/server/mir/shell/accessibility_manager.h
+++ b/src/include/server/mir/shell/accessibility_manager.h
@@ -1,0 +1,51 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_SHELL_ACCESSIBILITY_MANAGER_H
+#define MIR_SHELL_ACCESSIBILITY_MANAGER_H
+
+#include <memory>
+#include <vector>
+
+namespace mir
+{
+namespace shell
+{
+class KeyboardHelper;
+class AccessibilityManager
+{
+public:
+    void register_keyboard_helper(std::shared_ptr<shell::KeyboardHelper>);
+
+    int repeat_rate() const;
+    int repeat_delay() const;
+
+    void repeat_rate(int new_rate);
+    void repeat_delay(int new_rate);
+
+    void notify_helpers() const;
+
+private:
+    std::vector<std::shared_ptr<shell::KeyboardHelper>> keyboard_helpers;
+
+    // 25 rate and 600 delay are the default in Weston and Sway
+    int repeat_rate_{25};
+    int repeat_delay_{600};
+};
+}
+}
+
+#endif
--- a/src/include/server/mir/shell/accessibility_manager.h
+++ b/src/include/server/mir/shell/accessibility_manager.h
@@ -29,7 +29,7 @@ class KeyboardHelper;
 class AccessibilityManager
 {
 public:
-    void register_keyboard_helper(std::shared_ptr<shell::KeyboardHelper>);
+    void register_keyboard_helper(std::shared_ptr<shell::KeyboardHelper> const&);
 
     std::optional<int> repeat_rate() const;
     int repeat_delay() const;
--- a/src/miral/input_configuration.cpp
+++ b/src/miral/input_configuration.cpp
@@ -17,7 +17,9 @@
 #include "miral/input_configuration.h"
 
 #include "input_device_config.h"
+#include "mir/shell/accessibility_manager.h"
 
+#include <memory>
 #include <mir/server.h>
 #include <mir/input/input_device_hub.h>
 
@@ -35,16 +37,25 @@ class miral::InputConfiguration::Touchpad::Self : public TouchpadInputConfigurat
     using TouchpadInputConfiguration::operator=;
 };
 
+class miral::InputConfiguration::Keyboard::Self : public KeyboardInputConfiguration
+{
+    friend class miral::InputConfiguration::Self;
+    using KeyboardInputConfiguration::operator=;
+};
+
 class miral::InputConfiguration::Self
 {
 public:
     std::weak_ptr<mir::input::InputDeviceHub> input_device_hub{};
     std::shared_ptr<InputDeviceConfig> input_device_config;
+    std::shared_ptr<mir::shell::AccessibilityManager> accessibility_manager;
 
     auto mouse() -> Mouse;
     void mouse(Mouse const& val);
     auto touchpad() -> Touchpad;
     void touchpad(Touchpad const& val);
+    auto keyboard() -> Keyboard;
+    void keyboard(Keyboard const& val);
 
     void apply(Mouse const& m)
     {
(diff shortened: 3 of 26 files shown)


src/server/report/default_server_configuration.cpp
```diff
@@ -113,3 +114,12 @@ auto mir::DefaultServerConfiguration::the_shell_report() -> std::shared_ptr<shel
...
 
+auto mir::DefaultServerConfiguration::the_accessibility_manager() -> std::shared_ptr<shell::AccessibilityManager>
+{
+    return accessibility_manager(
+        []
+        {
+            return std::make_shared<shell::AccessibilityManager>();
+        });
+}
+
```
tarek-y-ismail, line 125: Uh, I don't know how this got here? I think it fits better in `src/server/shell/default_configuration.cpp`

include/miral/miral/config_file.h
```diff
@@ -56,6 +57,8 @@ class ConfigFile
     ConfigFile(MirRunner& runner, std::filesystem::path file, Mode mode, Loader load_config);
     ~ConfigFile();
 
+    void operator()(mir::Server&);
+    void load_and_watch();
```
AlanGriffiths, line 61: Unfortunately, changing the way this class is used breaks the API. We should try to find an alternative.

I think the best way to achieve that would be to use `MirRunner::add_start_callback()` in the constructor to `self->load_and_watch()`.

In any case, there should be no need for `ConfigFile::load_and_watch()` - that's an implementation detail that users need not be exposed to.

src/include/server/mir/frontend/keyboard_helper.h
```diff
@@ -74,6 +73,15 @@ class KeyboardHelper
...
 private:
+    friend class mir::frontend::WlSeat;
+    friend class mir::shell::AccessibilityManager;
+    KeyboardHelper(
+        KeyboardCallbacks* keybaord_impl,
+        std::shared_ptr<mir::input::Keymap> const& initial_keymap,
+        std::shared_ptr<input::Seat> const& seat,
+        int default_repeat_rate,
+        int default_repeat_delay);
+
```
AlanGriffiths, line 84: Why are we making the constructor private? And adding `friends`?

Especially making `mir::shell::AccessibilityManager` a friend looks wrong: why would `shell` code know anything about the `frontend` implementation classes?

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -0,0 +1,57 @@
...
+{
+class KeyboardCallbacks;
+class KeyboardHelper;
+}
+namespace shell
+{
+class AccessibilityManager
+{
+public:
+    void register_keyboard_helper(std::shared_ptr<frontend::KeyboardHelper>);
```
AlanGriffiths, line 35: Looks wrong: the alternative is to declare a `shell::KeyboardHelper` interface (which `frontend::KeyboardHelper` could implement)

src/server/frontend_wayland/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ set(
   window_wl_surface_role.cpp    window_wl_surface_role.h
   wl_surface.cpp                wl_surface.h
   wl_seat.cpp                   wl_seat.h
-  keyboard_helper.cpp           keyboard_helper.h
+  keyboard_helper.cpp
```
AlanGriffiths, line 29: I think removing this header from the project is fallout from not having a `KeyboardHelper` interface in shell

src/server/report/default_server_configuration.cpp
AlanGriffiths: +1

AlanGriffiths (changes requested):
No tests for the new code?

A few things need shuffling around.

Also, it would be nice to split the ConfigFile changes into a separate, _precursor_, PR. (They should be uncontentious and quick to land)

src/server/shell/accessibility_manager.cpp
```diff
@@ -0,0 +1,54 @@
...
+{
+    keyboard_helpers.push_back(helper);
+}
+
+int mir::shell::AccessibilityManager::repeat_rate() const {
+    // https://wayland.app/protocols/wayland#wl_keyboard:event:repeat_info
+    // " A rate of zero will disable any repeating (regardless
+    // of the value of delay)."
+    if(!enable_key_repeat)
+        return 0;
```
mattkae, line 30: Nit: I would return `std::nullopt` here instead of 0 since the wayland bit is an implementation detail of the Wayland frontend.

mattkae (changes requested):
This looks pretty reasonable! I've got a few points for improvement

src/include/server/mir/shell/accessibility_manager.h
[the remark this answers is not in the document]
tarek-y-ismail: I think this might be good for the future if other accessibility features need it, but for the time being I think Alan's suggestion of `shell::KeyboardHelper` would be better since it fits what we're doing _now_

src/miral/input_configuration.cpp
```diff
@@ -321,3 +374,29 @@ void miral::InputConfiguration::Touchpad::tap_to_click(std::optional<bool> const
...
+miral::InputConfiguration::Keyboard::~Keyboard() = default;
+
+miral::InputConfiguration::Keyboard::Keyboard(Keyboard const& that) :
+    self{std::make_unique<Self>(*that.self)}
+{
+}
+
+auto miral::InputConfiguration::Keyboard::operator=(Keyboard that) -> Keyboard&
+{
+    std::swap(self, that.self);
```
mattkae, line 392: Why would `operator=` be a swap? Shouldn't there be an overrided `operator=` on `self` so that we copy all of its values?

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -0,0 +1,54 @@
...
+
+namespace mir
+{
+namespace shell
+{
+class KeyboardHelper;
+class AccessibilityManager
+{
+public:
+    void register_keyboard_helper(std::shared_ptr<shell::KeyboardHelper>);
```
mattkae, line 32:
```suggestion
    void register_keyboard_helper(std::shared_ptr<shell::KeyboardHelper> const&);
```

src/server/report/default_server_configuration.cpp
```diff
@@ -112,4 +112,3 @@ auto mir::DefaultServerConfiguration::the_shell_report() -> std::shared_ptr<shel
             return report_factory(options::shell_report_opt)->create_shell_report();
         });
 }
-
```
mattkae, line 115: Add new line back please :)

src/server/shell/accessibility_manager.cpp
```diff
@@ -0,0 +1,53 @@
...
+}
+
+void mir::shell::AccessibilityManager::override_key_repeat_settings(bool enable)
+{
+    enable_key_repeat = enable;
+}
+
+void mir::shell::AccessibilityManager::notify_helpers() const {
+    for(auto const& helper: keyboard_helpers)
+        helper->repeat_info_changed(repeat_rate_, repeat_delay_);
```
mattkae, line 52: If `enable_key_repeat` is false, should `repeat_rate_` be `std::nullopt`?

mattkae (changes requested):
A few more questions, but looking good!

tarek-y-ismail:
Note to self: Just came across this while working on something else, might be a good idea to check out
https://github.com/canonical/mir/blob/5e15fde724fa99e37e09fc9a2719170896586ea7/src/server/input/default_configuration.cpp#L144-L164

src/miral/input_configuration.cpp
tarek-y-ismail: Hmm, I copied and pasted the touchpad code to keep everything consistent, so they also do the same thing... 

I'm not sure why that is

src/server/shell/accessibility_manager.cpp
tarek-y-ismail: According to: https://wayland.app/protocols/wayland#wl_keyboard:event:repeat_info,  "rate of zero will disable any repeating (regardless of the value of delay).". My interpretation was that we can just deliver the rate as an optional and the delay can be a normal value. It's then up to the user to interpret it. Which reminds me that I should use `repeat_rate()` here as well

src/miral/input_configuration.cpp
AlanGriffiths: That's the canonical way to write an assignment operator in c++: copy into the parameter and swap with `*this`.

tarek-y-ismail:
> Note to self: Just came across this while working on something else, might be a good idea to check out

Very confused what this does since repeat on both wayland and X seem to work without any modification to it.

examples/mir_demo_server/server_example.cpp
```diff
@@ -120,14 +121,15 @@ try
     miral::MirRunner runner{argc, argv, "mir/mir_demo_server.config"};
 
     miral::InputConfiguration input_configuration;
+    auto mouse = input_configuration.mouse();
+    auto touchpad = input_configuration.touchpad();
+    auto keyboard = input_configuration.keyboard();
```
AlanGriffiths, line 126: I think we could grow a utility class that combines these and the repeated "apply" logic?

Also, we need to be careful: we're now accessing `mouse`, `touchpad` and `keyboard` on multiple threads.

AlanGriffiths: Having said that, I like the revised approach

tarek-y-ismail: Hmm, do you have anything particular in mind? The simplest solution I can think of is keeping the initialization logic as is and wrapping the application logic in a lambda with a mutex.

AlanGriffiths: It is always _possible_ to have data and related functions separately, but it is clearer to the reader when they are grouped as a class. In particular, it is easier to check invariants when it is clear which code can access the data. 

In this case, I think the logic around `config_mutex`, `mouse`, `touchpad` and `keyboard` is complex enough to make isolating this data desirable:

```c++
class InputConfigFile : miral::ConfigFile
{
public:
    InputConfigFile(miral::MirRunner& runner, std::filesystem::path file) :
        ConfigFile{runner, file, ConfigFile::Mode::reload_on_change, [this](auto args...) {loader(args...); }}
    {
        runner.add_start_callback(
            [&]
            {
                std::lock_guard lock{config_mutex};
                config_applier();
            });
    }

    void operator()(mir::Server& server)
    {
        input_configuration(server);
    }

private:
    miral::InputConfiguration input_configuration;
    auto mouse = input_configuration.mouse();
    auto touchpad = input_configuration.touchpad();
    auto keyboard = input_configuration.keyboard();
    std::mutex config_mutex;

    auto config_applier = [&] mutable
    {
        input_configuration.mouse(mouse);
        input_configuration.touchpad(touchpad);
        input_configuration.keyboard(keyboard);
    };

    
    void loader(std::istream& istream, std::filesystem::path const& path)
    {
        std::lock_guard lock{config_mutex};
        // ...
    }
};
```

It is quick to see that the data is only touched by this code, and that it is appropriately locked.

mattkae (dismissed):
I believe I found a bug, but I don't know where it could be in the code! Here's a reproducer for you:

1. Start `mir_demo_server`
2. Open up `gedit`
3. Have the following input config:
   ```
   repeat_rate=1
   repeat_delay=1
   ```
4. Hold down the 'a' key and start inserting 'a's. It should insert two 'a's almost immediately, and then 1 'a' every second or so
5. Clear the gedit file. Set `repeat_rate` to 2 and save the config. Then set it to 1 again immediately and save the config.
6. Go back to gedit and hold down the 'a' key. Two 'a's will be inserted almost immediately and then 1 'a' will appear every second or so. HOWEVER, note that at some point later, you will get the "double 'a' insertion" again, even though you didn't start inserting again!

Other than this bug, it works well for me!

examples/mir_demo_server/server_example.cpp
AlanGriffiths: Here's a version without the undefined behaviour and compilation errors:

```c++
class InputConfigFile
{
public:
    InputConfigFile(miral::MirRunner& runner, std::filesystem::path file) :
        config_file{
            runner,
            file,
            miral::ConfigFile::Mode::reload_on_change,
            [this](std::istream& istream, std::filesystem::path const& path) {loader(istream, path); }}
    {
        runner.add_start_callback([this]
            {
                std::lock_guard lock{config_mutex};
                apply_config();
            });
    }

    void operator()(mir::Server& server)
    {
        input_configuration(server);
    }

private:
    miral::InputConfiguration input_configuration;
    miral::InputConfiguration::Mouse mouse = input_configuration.mouse();
    miral::InputConfiguration::Touchpad touchpad = input_configuration.touchpad();
    miral::InputConfiguration::Keyboard keyboard = input_configuration.keyboard();
    miral::ConfigFile config_file;
    std::mutex config_mutex;

    void apply_config()
    {
        input_configuration.mouse(mouse);
        input_configuration.touchpad(touchpad);
        input_configuration.keyboard(keyboard);
    };


    void loader(std::istream& in, std::filesystem::path const& path)
    {
        std::cout << "** Reloading: " << path << std::endl;

        std::lock_guard lock{config_mutex};

        for (std::string line; std::getline(in, line);)
        {
            std::cout << line << std::endl;

            if (line == "mir_pointer_handedness_right")
                mouse.handedness(mir_pointer_handedness_right);
            if (line == "mir_pointer_handedness_left")
                mouse.handedness(mir_pointer_handedness_left);

            if (line == "mir_touchpad_scroll_mode_none")
                touchpad.scroll_mode(mir_touchpad_scroll_mode_none);
            if (line == "mir_touchpad_scroll_mode_two_finger_scroll")
                touchpad.scroll_mode(mir_touchpad_scroll_mode_two_finger_scroll);
            if (line == "mir_touchpad_scroll_mode_edge_scroll")
                touchpad.scroll_mode(mir_touchpad_scroll_mode_edge_scroll);
            if (line == "mir_touchpad_scroll_mode_button_down_scroll")
                touchpad.scroll_mode(mir_touchpad_scroll_mode_button_down_scroll);

            if (line.contains("="))
            {
                auto const eq = line.find_first_of("=");
                auto const key = line.substr(0, eq);
                auto const value = line.substr(eq+1);

                auto const parse_and_validate = [](std::string const& key, std::string_view val) -> std::optional<int>
                {
                    auto const int_val = std::atoi(val.data());
                    if (int_val < 0)
                    {
                        mir::log_warning(
                            "Config value %s does not support negative values. Ignoring the supplied value (%d)...",
                            key.c_str(), int_val);
                        return std::nullopt;
                    }

                    return int_val;
                };

                if (key == "repeat_rate")
                {
                    auto const parsed = parse_and_validate(key, value);
                    if (parsed)
                        keyboard.set_repeat_rate(*parsed);
                }

                if (key == "repeat_delay")
                {
                    auto const parsed = parse_and_validate(key, value);
                    if (parsed)
                        keyboard.set_repeat_delay(*parsed);
                }
            }
        }

        apply_config();
    }
};
```

One small annoyance is that this isn't Copyable, and that means the `run_with()` parameter needs passing with `std::ref()`. Hoisting the state into a shared Self would work, but seems overkill.

AlanGriffiths:
> I believe I found a bug, but I don't know where it could be in the code

From the description the bug could lie in GTK (implementing the repeat is the client's responsibility).

The thing to check is what repeat_info events are being sent to the client when the edit occurs.

mattkae:
> > I believe I found a bug, but I don't know where it could be in the code
> 
> From the description the bug could lie in GTK (implementing the repeat is the client's responsibility).
> 
> The thing to check is what repeat_info events are being sent to the client when the edit occurs.

Same issue with `foot`. Here was my WAYLAND_DEBUG trace when it happened:

```

[ 271826.715] wl_keyboard@22.key(71, 26758364, 30, 1)
[ 271828.200]  -> wl_surface@3.damage_buffer(96, 104, 12, 13)
[ 271828.233]  -> wl_surface@3.frame(new id wl_callback@42)
[ 271828.248]  -> wl_surface@3.set_buffer_scale(1)
[ 271828.258]  -> wp_viewport@27.set_destination(688, 470)
[ 271828.272]  -> wl_surface@3.attach(wl_buffer@38, 0, 0)
[ 271828.287]  -> wl_surface@3.commit()
[ 271828.300]  -> zwp_text_input_v3@12.set_cursor_rectangle(102, 104, 6, 13)
[ 271828.311]  -> zwp_text_input_v3@12.commit()
[ 271828.994] wl_buffer@35.release()
[ 271831.243] wl_display@1.delete_id(42)
[ 271831.378] wl_callback@42.done(26758369)
[ 271834.147]  -> wl_surface@3.damage_buffer(102, 104, 12, 13)
[ 271835.000]  -> wl_surface@3.frame(new id wl_callback@42)
[ 271835.040]  -> wl_surface@3.set_buffer_scale(1)
[ 271835.063]  -> wp_viewport@27.set_destination(688, 470)
[ 271835.093]  -> wl_surface@3.attach(wl_buffer@35, 0, 0)
[ 271835.115]  -> wl_surface@3.commit()
[ 271835.137]  -> zwp_text_input_v3@12.set_cursor_rectangle(108, 104, 6, 13)
[ 271835.160]  -> zwp_text_input_v3@12.commit()
[ 271836.445] wl_buffer@38.release()
[ 271837.392] wl_display@1.delete_id(42)
[ 271837.411] wl_callback@42.done(26758375)
[ 272829.082]  -> wl_surface@3.damage_buffer(108, 104, 12, 13)
[ 272829.125]  -> wl_surface@3.frame(new id wl_callback@42)
[ 272829.145]  -> wl_surface@3.set_buffer_scale(1)
[ 272829.163]  -> wp_viewport@27.set_destination(688, 470)
[ 272829.181]  -> wl_surface@3.attach(wl_buffer@38, 0, 0)
[ 272829.199]  -> wl_surface@3.commit()
[ 272829.215]  -> zwp_text_input_v3@12.set_cursor_rectangle(114, 104, 6, 13)
[ 272829.234]  -> zwp_text_input_v3@12.commit()
[ 272829.907] wl_buffer@35.release()
[ 272832.842] wl_display@1.delete_id(42)
[ 272832.884] wl_callback@42.done(26759371)
[ 273345.816] wl_keyboard@22.key(72, 26759883, 30, 0)
[ 273345.958] wl_keyboard@22.key(73, 26759883, 30, 1)
[ 273347.145]  -> wl_surface@3.damage_buffer(114, 104, 12, 13)
[ 273347.205]  -> wl_surface@3.frame(new id wl_callback@42)
[ 273347.222]  -> wl_surface@3.set_buffer_scale(1)
[ 273347.237]  -> wp_viewport@27.set_destination(688, 470)
[ 273347.252]  -> wl_surface@3.attach(wl_buffer@35, 0, 0)
[ 273347.268]  -> wl_surface@3.commit()
[ 273347.282]  -> zwp_text_input_v3@12.set_cursor_rectangle(120, 104, 6, 13)
[ 273347.299]  -> zwp_text_input_v3@12.commit()
[ 273347.935] wl_buffer@38.release()
[ 273351.941] wl_display@1.delete_id(42)
[ 273352.030] wl_callback@42.done(26759889)
[ 273352.483]  -> wl_surface@3.damage_buffer(120, 104, 12, 13)
[ 273352.554]  -> wl_surface@3.frame(new id wl_callback@42)
[ 273352.612]  -> wl_surface@3.set_buffer_scale(1)
[ 273352.655]  -> wp_viewport@27.set_destination(688, 470)
[ 273352.734]  -> wl_surface@3.attach(wl_buffer@38, 0, 0)
[ 273352.778]  -> wl_surface@3.commit()
[ 273352.819]  -> zwp_text_input_v3@12.set_cursor_rectangle(126, 104, 6, 13)
[ 273352.863]  -> zwp_text_input_v3@12.commit()
[ 273353.886] wl_buffer@35.release()
[ 273355.961] wl_display@1.delete_id(42)
[ 273356.020] wl_callback@42.done(26759894)
[ 273890.650] wl_keyboard@22.key(74, 26760428, 30, 0)

```

AlanGriffiths:
> our X11 platform - I know there's some messy logic handling the input there

Some "printf() debugging" supports this:

```
void mir::input::X::XInputDevice::key_press(EventTime, xkb_keysym_t, int)
void mir::input::X::XInputDevice::key_release(EventTime, xkb_keysym_t, int)
void mir::input::X::XInputDevice::key_press(EventTime, xkb_keysym_t, int)
void mir::input::X::XInputDevice::key_release(EventTime, xkb_keysym_t, int)
```

That is, X11 is telling Mir the key has been released and pressed and we pass that on. (I thought we had logic to suppress these release/press pairs, but I don't see it know)

AlanGriffiths:
@tarek-y-ismail I think the only thing to address is my InputConfigFile suggestion. (If nobody else thinks it worthwhile, then :shrug:)

AlanGriffiths:
> X11 is telling Mir the key has been released and pressed and we pass that on

Likely the same cause as #1471

mattkae:
> > X11 is telling Mir the key has been released and pressed and we pass that on
> 
> Likely the same cause as #1471

Great. I marked it as `triaged`. That's a good thing to fix, as it will negatively affect certain experience (e.g. gaming)

mattkae (approved):
My issue was not your issue, so I am happy :smile:

AlanGriffiths:
> Great. I marked it as `triaged`. That's a good thing to fix, as it will negatively affect certain experience (e.g. gaming)

Is anyone really going to game on the X11 platform?!

AlanGriffiths (approved):
I'd still like to improve the example code, but won't block on it

Saviq:
> Is anyone really going to game on the X11 platform?!

That's the only way for WINE/Proton-based games (i.e. the majority) still...

AlanGriffiths:
> > 
> 
> That's the only way for WINE/Proton-based games (i.e. the majority) still...

The `mir:X11` **graphics platform**

PR #3735, opened 2025-02-04
Cursor Scaling Support

tarek-y-ismail:
Implements scaling for software, gbm (and atomic-kms by extension), and wayland cursors.


src/server/CMakeLists.txt
```diff
@@ -14,6 +15,7 @@ include_directories(
   ${PROJECT_SOURCE_DIR}/src/include/client
   ${PROJECT_SOURCE_DIR}/src/include/server
   ${PROJECT_SOURCE_DIR}/src/include/wayland
+  ${PIXMAN_INCLUDE_DIRS}
```
Saviq, line 18: You shouldn't need this, target_link_libraries over an `IMPORTED_TARGET` adds include dirs automagically.
```suggestion
```

src/server/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 pkg_check_modules(FREETYPE freetype2 REQUIRED IMPORTED_TARGET) # For rendering decorated window titles
+pkg_check_modules(PIXMAN REQUIRED IMPORTED_TARGET pixman-1)
```
Saviq, line 2: You will need to add the pixman dev package as appropriate in:
- `debian/control`
- `spread/build/*/task.yaml`

Saviq (changes requested):

src/server/CMakeLists.txt
Saviq: Oh well, must be something dumb in the pixman `.pc` file, then.

src/server/graphics/software_cursor.cpp
```diff
@@ -145,17 +150,18 @@ void mg::SoftwareCursor::show(CursorImage const& cursor_image)
 {
     std::lock_guard lg{guard};
 
-    auto const to_remove = visible ? renderable : nullptr;
+    // Store the cursor image for later use with `set_scale`
+    this->current_cursor_image = &cursor_image;
```
AlanGriffiths, line 154: The guarantee with a reference parameter is that it is "alive" for the duration of the call. Storing a pointer to it for later use (i.e. after the call) isn't safe.

We should either change the interface to explicitly share ownership, or make a copy for later use

tarek-y-ismail: No worries, was just hacking something together to get the ball going. I fully intend on changing this.

AlanGriffiths: Yeah, I wasn't intending to look at this until I saw Saviq discussing pkg-config and...

src/include/server/mir/default_server_configuration.h
```diff
@@ -432,13 +432,13 @@ class DefaultServerConfiguration : public virtual ServerConfiguration
...
     CachedPtr<shell::Shell> shell;
     CachedPtr<shell::ShellReport> shell_report;
-    CachedPtr<shell::AccessibilityManager> accessibility_manager;
     CachedPtr<shell::decoration::Manager> decoration_manager;
     CachedPtr<scene::ApplicationNotRespondingDetector> application_not_responding_detector;
     CachedPtr<cookie::Authority> cookie_authority;
     CachedPtr<input::receiver::XKBMapperRegistrar> xkb_mapper_registrar;
     std::shared_ptr<ConsoleServices> console_services;
     std::shared_ptr<DecorationStrategy> decoration_strategy;
+    std::shared_ptr<shell::AccessibilityManager> accessibility_manager;
```
tarek-y-ismail, line 441: Not sure how we can fix the lifetime issues that come with cached ptrs, any suggestions?

tarek-y-ismail:
`SoftwareCursor` can be scaled by updating the width and height returned from `screen_position`
https://github.com/canonical/mir/blob/fbf87c7c12c9d872fa8ae562eb4f6f92a6a97ebe/src/server/graphics/software_cursor.cpp#L81-L85

include/platform/mir/graphics/pixman_image_scaling.h
```diff
@@ -0,0 +1,35 @@
...
+#include "mir/geometry/size.h"
+#include <memory>
+
+namespace mir::graphics
+{
+struct PixBuffer
+{
+    std::unique_ptr<uint32_t[]> data;
+    geometry::Size size;
+};
```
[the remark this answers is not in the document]
tarek-y-ismail: This could be useful:

https://github.com/canonical/mir/blob/08ccc410d47933d4e0847909c74d0147eca94d5f/include/platform/mir/renderer/sw/pixel_source.h#L40-L64

src/include/server/mir/default_server_configuration.h
tarek-y-ismail: Could be fixed by inverting the dependency between the cursor and the accessibility manager

tarek-y-ismail: - Before initialization, the cursor is initialized, which initializes the accessibility manager
- We set the accessibility manager's cursor pointer to point at the cursor we just initialized
- The accessibility manager gets destroyed because nothing else holds a reference to it
- Somewhere _after_ initialization, `InputConfiguration` gets a reference to the accessibility manager (a long lived reference)
- We read the input config
- Because the accessibility manager was destroyed, the previous reference to the cursor is lost
- We don't apply the cursor changes

Possible solution: Pass the cursor to the accessibility manager during construction (if that doesn't break other things)

include/platform/mir/graphics/pixman_image_scaling.h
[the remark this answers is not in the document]
tarek-y-ismail: Hmm, a lot of code added for not much benefit. I'll keep the patch around somewhere, but I don't think it's worth the amount of code added

tarek-y-ismail:
Bug: If you change the cursor scale from inside the running compositor, a copy of the new cursor is displayed where the old cursor was.

tarek-y-ismail:
Another bug: the cursor doesn't work properly with x apps

src/platform/symbols.map
```diff
@@ -84,6 +84,7 @@ MIR_PLATFORM_2.19 {
     mir::graphics::initialise_egl_logger*;
     mir::graphics::operator*;
     mir::graphics::red_channel_depth*;
+    mir::graphics::scale_cursor_image*;
```
AlanGriffiths, line 87: This PR breaks the platform ABI, everything should move to a 2.20 stanza

src/include/server/mir/default_server_configuration.h
```diff
@@ -432,13 +432,13 @@ class DefaultServerConfiguration : public virtual ServerConfiguration
...
     CachedPtr<shell::Shell> shell;
     CachedPtr<shell::ShellReport> shell_report;
-    CachedPtr<shell::AccessibilityManager> accessibility_manager;
     CachedPtr<shell::decoration::Manager> decoration_manager;
     CachedPtr<scene::ApplicationNotRespondingDetector> application_not_responding_detector;
     CachedPtr<cookie::Authority> cookie_authority;
     CachedPtr<input::receiver::XKBMapperRegistrar> xkb_mapper_registrar;
     std::shared_ptr<ConsoleServices> console_services;
     std::shared_ptr<DecorationStrategy> decoration_strategy;
+    CachedPtr<shell::AccessibilityManager> accessibility_manager;
```
AlanGriffiths, line 441: Looks like an unnecessary reordering change?

src/server/symbols.map
```diff
@@ -1420,6 +1420,8 @@ global:
     mir::Server::get_activation_token*;
     mir::Server::the_token_authority*;
     mir::shell::AbstractShell::is_above*;
+    mir::shell::AccessibilityManager::cursor_scale_changed*;
+    mir::shell::AccessibilityManager::set_cursor*;
```
AlanGriffiths, line 1424: Changing the layout of `AccessibilityManager` is an ABI break: we need to move to a 2.20 stanza

tests/unit-tests/graphics/test_software_cursor.cpp
```diff
@@ -424,12 +426,12 @@ TEST_F(SoftwareCursor, handles_argb_8888_cursor_surface)
     ON_CALL(mock_buffer_allocator, supported_pixel_formats())
         .WillByDefault(Return(std::vector<MirPixelFormat>{ mir_pixel_format_argb_8888 }));
 
-    StubCursorImage test_image{{8, 8}};
+    auto test_image = std::make_shared<StubCursorImage>(geom::Displacement{8, 8});
```
AlanGriffiths, line 429:
```suggestion
    auto const test_image = std::make_shared<StubCursorImage>(geom::Displacement{8, 8});
```

tests/unit-tests/graphics/test_software_cursor.cpp
```diff
@@ -478,12 +480,12 @@ TEST_F(SoftwareCursor, handles_argb_8888_buffer_with_stride)
...
```
AlanGriffiths, line 483:
```suggestion
    auto const test_image = std::make_shared<StubCursorImage>(geom::Displacement{8, 8});
```

tests/unit-tests/input/test_cursor_controller.cpp
```diff
@@ -73,9 +74,9 @@ struct ZeroSizedCursorImage : public mg::CursorImage
     geom::Displacement hotspot() const override { return geom::Displacement{0, 0}; }
 };
 
-bool cursor_is_named(mg::CursorImage const& i, std::string const& name)
+bool cursor_is_named(std::shared_ptr<mg::CursorImage> const& i, std::string const& name)
 {
-    auto image = dynamic_cast<NamedCursorImage const*>(&i);
+    auto image = std::dynamic_pointer_cast<NamedCursorImage const>(i);
```
AlanGriffiths, line 79:
```suggestion
    auto const image = std::dynamic_pointer_cast<NamedCursorImage const>(i);
```

tests/unit-tests/platforms/gbm-kms/kms/test_cursor.cpp
```diff
@@ -321,7 +322,7 @@ struct MesaCursorTest : ::testing::Test
     }
 
     testing::NiceMock<mtd::MockDRM> mock_drm;
-    StubCursorImage stub_image;
+    std::shared_ptr<StubCursorImage> stub_image = std::make_shared<StubCursorImage>();
```
AlanGriffiths, line 325:
```suggestion
    auto const stub_image = std::make_shared<StubCursorImage>();
```

AlanGriffiths (changes requested):
A bunch of "housekeeping"...

src/server/symbols.map
tarek-y-ismail: Everything, or just these two symbols?

tests/unit-tests/platforms/gbm-kms/kms/test_cursor.cpp
tarek-y-ismail: This one won't work since we're inside a class definition

tarek-y-ismail:
@Saviq @AlanGriffiths @RAOF  Took a moment to fix some minor blemishes, feel free to review whenever you can

debian/libmiral7.symbols
```diff
@@ -494,10 +494,14 @@ libmiral.so.7 libmiral7 #MINVER#
  MIRAL_5.2@MIRAL_5.2 5.2.0
  (c++)"miral::InputConfiguration::Keyboard::Keyboard()@MIRAL_5.2" 5.2.0
  (c++)"miral::InputConfiguration::Keyboard::Keyboard(miral::InputConfiguration::Keyboard const&)@MIRAL_5.2" 5.2.0
+ (c++)"miral::InputConfiguration::Keyboard::merge_settings_from(miral::InputConfiguration::Keyboard const&)@MIRAL_5.2" 5.2.0
```
AlanGriffiths, line 497: Miral-5.2 was released in Mir-2.19, we're now on 5.3

include/platform/mir/graphics/cursor.h
```diff
@@ -31,16 +31,20 @@ class CursorImage;
 class Cursor
 {
 public:
-    virtual void show(CursorImage const& cursor_image) = 0;
+    virtual void show(std::shared_ptr<CursorImage> const& cursor_image) = 0;
     virtual void hide() = 0;
 
     virtual void move_to(geometry::Point position) = 0;
 
+    virtual void set_scale(float new_scale) = 0;
```
AlanGriffiths, line 39: This breaks the platform ABI

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -23,12 +23,18 @@
...
+{
+class Cursor;
+}
 namespace shell
 {
 class KeyboardHelper;
 class AccessibilityManager
 {
 public:
+    AccessibilityManager(std::shared_ptr<mir::graphics::Cursor> const& cursor);
```
AlanGriffiths, line 36: Converting constructors should be `explicit`

AlanGriffiths (changes requested):

include/platform/mir/graphics/cursor.h
tarek-y-ismail: Move all symbols to a new stanza?

src/include/server/mir/shell/accessibility_manager.h
tarek-y-ismail: This isn't meant to convert a cursor to an accessibility manager, I just needed a reference to the cursor so I can notify it when the scale changes.

AlanGriffiths: Quite

tarek-y-ismail: I think you accidentally erased most of the message

AlanGriffiths: No I didn't. I agree this isn't meant to convert a cursor to an accessibility manager.

Therefore, it should be `explicit` and not an implicitly converting constructor.

include/platform/mir/graphics/cursor.h
tarek-y-ismail: Oh, I think I also forgot to bump the ABI to 31, anything else I'm missing?

RAOF (changes requested):
Looks good, apart from the misunderstanding around buffer stride; that should be fixed.

src/platform/graphics/pixman_image_scaling.cpp
```diff
@@ -0,0 +1,83 @@
...
+    {
+        return static_cast<int>(in_value / multiple) * multiple;
+    };
+
+    // Round everything to the closest multiple of four since pixman expects
+    // the stride to be a multiple of 4. Should also work with any number of
+    // bytes per pixel even.
+    // https://gitlab.freedesktop.org/pixman/pixman/-/blob/master/pixman/pixman-bits-image.c#L1339
+    auto const scaled_width = to_closest_multiple_of(width.as_value() * new_scale, 4);
+    auto const scaled_height = to_closest_multiple_of(height.as_value() * new_scale, 4);
```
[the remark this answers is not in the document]
tarek-y-ismail: Originally, the code would crash with scales that caused the stride to be not a multiple of 4. For example, if the cursor size was 24 and the scale was 1.9. `24 * 1.9 ≈ 45.6`, pixman would then crash at the line I linked to. 

Originally, I thought pixman expected the width and height to be multiples of 4 (sometimes I wonder how I come up with these ideas), hence why the "rounding" here. But now that you mentioned it again, I tried changing things to just  round the width and height up/down to the nearest integer and things continue to work fine.

One reason why this didn't blow up when you tested is because later I changed software cursors to use `screen_position` to scale themselves, so this code is never executed with software cursors. Reverting the commit applying these changes and trying with a few different scale values got things to blow up on my end.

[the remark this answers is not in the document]
RAOF: > Originally, the code would crash with scales that caused the stride to be not a multiple of 4

I don't see *how* the stride could not be a multiple of 4, because you're multiplying the width (in pixels) by `MIR_BYTES_PER_PIXEL(...)`, which is 4. Unless you were trying to have a fractional width, and that indeed is not going to work (since you can't write to .4 of a pixel)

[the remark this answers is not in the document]
tarek-y-ismail: Exactly, I misinterpreted the message before, but when you mentioned it again the original message it was all cleared up

RAOF (approved):
Yup, LGTM now.

include/miral/miral/input_configuration.h
```diff
@@ -78,6 +78,11 @@ class InputConfiguration::Mouse
     void acceleration_bias(std::optional<double>const& val);
     void vscroll_speed(std::optional<double>const& val);
     void hscroll_speed(std::optional<double>const& val);
+    void scale(std::optional<double>const& val);
+
+    /// Copies the corresponding value from `other` for all unset fields.
+    /// \remark Since MirAL 5.2
```
AlanGriffiths, line 84:
```suggestion
    /// \remark Since MirAL 5.3
```

include/miral/miral/input_configuration.h
```diff
@@ -118,6 +123,10 @@ class InputConfiguration::Touchpad
     void scroll_mode(std::optional<MirTouchpadScrollMode>const& val);
     void tap_to_click(std::optional<bool>const& val);
 
+    /// Copies the corresponding value from `other` for all unset fields.
+    /// \remark Since MirAL 5.2
```
AlanGriffiths, line 127:
```suggestion
    /// \remark Since MirAL 5.3
```

include/miral/miral/input_configuration.h
```diff
@@ -139,6 +148,10 @@ class InputConfiguration::Keyboard
     void set_repeat_rate(int new_rate);
     void set_repeat_delay(int new_delay);
 
+    /// Copies the corresponding value from `other` for all unset fields.
+    /// \remark Since MirAL 5.2
```
AlanGriffiths, line 152:
```suggestion
    /// \remark Since MirAL 5.3
```

AlanGriffiths (changes requested):
Nits.

Oh, and...

> Missing the actual configuration portion.

...is unclear and could be expanded on: there is an example "actual configuration portion"

include/miral/miral/input_configuration.h
AlanGriffiths: Actually, this is redundant: the whole of InputConfiguration::Keyboard is already "Since MirAL 5.3"

src/platform/graphics/pixman_image_scaling.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <cmath>
+#include <cstring>
+
+mir::graphics::PixBuffer mir::graphics::scale_cursor_image(
+    mir::graphics::CursorImage const& cursor_image, float new_scale)
+{
+    auto const [width, height] = cursor_image.size();
+
+    // Special case: no resizing. Just copy the data.
+    if(new_scale == 1)
```
AlanGriffiths, line 33: Possibly clearer without an implicit conversion?

```suggestion
    if(new_scale == 1.0f)
```

tarek-y-ismail:
> Oh, and...
> 
> > Missing the actual configuration portion.
> 
> ...is unclear and could be expanded on: there is an example "actual configuration portion"

D'oh, forgot to update the description. Thanks for catching that ;)

tarek-y-ismail:
@AlanGriffiths Ready for review (fr this time)! :saluting_face:

RAOF (changes requested):
I think the symbols files need merging with the new Mir release, but otherwise, :+1:

tarek-y-ismail:
> I think the symbols files need merging with the new Mir release, but otherwise, 👍

Just tried rebasing on main and running the symbol check targets. No new changes, everything seems to be A-OK!

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -48,18 +57,20 @@ class AccessibilityManager
...
+    void cursor_scale_changed(float new_scale);
+
 private:
     std::vector<std::shared_ptr<shell::KeyboardHelper>> keyboard_helpers;
 
-    // 25 rate and 600 delay are the default in Weston and Sway
-    int repeat_rate_{25};
-    int repeat_delay_{600};
+    int repeat_rate_;
+    int repeat_delay_;
```
AlanGriffiths, line 66: These no longer appear to be initialised (not here, not in the constructor)

AlanGriffiths (changes requested):
Some tweaking

src/platform/symbols.map
tarek-y-ismail: 2.21 since 2.20 was released!

include/platform/mir/options/configuration.h
```diff
@@ -38,11 +38,14 @@ extern char const* const input_report_opt;
...
 extern char const* const cursor_opt;
+extern char const* const cursor_scale_override_opt;
 extern char const* const fatal_except_opt;
 extern char const* const debug_opt;
 extern char const* const composite_delay_opt;
-extern char const* const enable_key_repeat_opt;
 extern char const* const enable_mouse_keys_opt;
+extern char const* const enable_key_repeat_opt;
+extern char const* const key_repeat_rate_opt;
+extern char const* const key_repeat_delay_opt;
```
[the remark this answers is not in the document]
tarek-y-ismail: They didn't. I just saw `enable_key_repeat_opt` and naively thought this was a good place to put related options. Thanks for catching this

src/include/server/mir/shell/accessibility_manager.h
tarek-y-ismail: I failed to account for the case where the shell author doesn't use `InputConfiguration`, reverted. Thanks for pointing this out.

src/platform/options/default_configuration.cpp
```diff
@@ -37,11 +37,12 @@ char const* const mo::shared_library_prober_report_opt = "shared-library-prober-
 char const* const mo::shell_report_opt            = "shell-report";
 char const* const mo::touchspots_opt              = "enable-touchspots";
 char const* const mo::cursor_opt                  = "cursor";
+char const* const mo::cursor_scale_override_opt   = "cursor-scale-override";
```
AlanGriffiths, line 40: I think "override" is redundant. This is just the cursor scale (with a default of 1)

```suggestion
char const* const mo::cursor_scale_opt   = "cursor-scale";
```

src/server/shell/accessibility_manager.cpp
```diff
@@ -114,12 +117,24 @@ struct MouseKeysTransformer: public mir::input::InputEventTransformer::Transform
...
     enable_key_repeat{options->get<bool>(options::enable_key_repeat_opt)},
     enable_mouse_keys{options->get<bool>(options::enable_mouse_keys_opt)},
     event_transformer{event_transformer},
-    transformer{std::make_shared<MouseKeysTransformer>()}
+    transformer{std::make_shared<MouseKeysTransformer>()},
+    cursor{cursor}
 {
+    auto const cursor_scale = static_cast<float>(options->get<double>(mir::options::cursor_scale_override_opt));
+    if (cursor_scale != 1.0)
+        cursor->set_scale(cursor_scale);
```
AlanGriffiths, line 130: It seems unnecessary to test `cursor_scale != 1.0 `

src/server/shell/accessibility_manager.cpp
```diff
@@ -114,12 +117,24 @@ struct MouseKeysTransformer: public mir::input::InputEventTransformer::Transform
...
+    std::shared_ptr<input::InputEventTransformer> const& event_transformer,
+    std::shared_ptr<mir::graphics::Cursor> const& cursor) :
     enable_key_repeat{options->get<bool>(options::enable_key_repeat_opt)},
     enable_mouse_keys{options->get<bool>(options::enable_mouse_keys_opt)},
     event_transformer{event_transformer},
-    transformer{std::make_shared<MouseKeysTransformer>()}
+    transformer{std::make_shared<MouseKeysTransformer>()},
+    cursor{cursor}
 {
+    auto const cursor_scale = static_cast<float>(options->get<double>(mir::options::cursor_scale_override_opt));
```
AlanGriffiths, line 128: I know it would need an insane scale to be problematic, but this is a narrowing conversion of external input. The value should be range checked or clamped. (I'd clamp to something much stricter: (0..100] say.)

AlanGriffiths (changes requested):
More tweaking

src/server/shell/accessibility_manager.cpp
tarek-y-ismail: Just saves a call to `set_scale`, but if you don't think it's worth it...

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -48,18 +57,24 @@ class AccessibilityManager
...
+    void cursor_scale_changed(float new_scale);
+
 private:
     std::vector<std::shared_ptr<shell::KeyboardHelper>> keyboard_helpers;
 
-    // 25 rate and 600 delay are the default in Weston and Sway
+    // Initialized via `repeat_{delay,rate}`
+    // Keep initial values in case nothing calls them.
     int repeat_rate_{25};
     int repeat_delay_{600};
```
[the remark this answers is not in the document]
AlanGriffiths: It is pre-existing, but are they independent? Or should they be set together? Vis:

```c++
   set_key_repeat(std::optional<int> repeat_rate,  std::optional<int> repeat_delay);
```

debian/libmiral7.symbols
```diff
@@ -503,3 +503,8 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::MinimalWindowManager::MinimalWindowManager(miral::WindowManagerTools const&, MirInputEventModifier, miral::FocusStealing)@MIRAL_5.0" 5.2.0
  (c++)"miral::MinimalWindowManager::MinimalWindowManager(miral::WindowManagerTools const&, miral::FocusStealing)@MIRAL_5.0" 5.2.0
  (c++)"miral::WindowManagementPolicy::~WindowManagementPolicy()@MIRAL_5.0" 5.2.0
+ MIRAL_5.3@MIRAL_5.3 5.3.0
+ (c++)"miral::InputConfiguration::Keyboard::merge_settings_from(miral::InputConfiguration::Keyboard const&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::InputConfiguration::Mouse::merge_settings_from(miral::InputConfiguration::Mouse const&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::InputConfiguration::Mouse::scale(std::optional<double> const&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::InputConfiguration::Touchpad::merge_settings_from(miral::InputConfiguration::Touchpad const&)@MIRAL_5.3" 5.3.0
```
AlanGriffiths, line 510: Do these Keyboard and Touchpad changes belong in a "Cursor Scaling Support" PR?

src/server/graphics/software_cursor.cpp
```diff
@@ -217,3 +213,37 @@ void mg::SoftwareCursor::move_to(geometry::Point position)
     // This doesn't need to be called in a specific order with other potential calls, so it doesn't go on the executor
     scene->emit_scene_changed();
 }
+
+void mir::graphics::SoftwareCursor::set_scale(float new_scale)
+{
+    std::lock_guard lg{guard};
+    set_scale_unlocked(new_scale);
```
AlanGriffiths, line 220: This is confusing to read:

1. we acquire a lock; then while holding the lock,
2. we call `*_unlocked`
3. the implementation of `set_scale_unlocked()` assumes the caller holds a lock

Maybe the naming is wrong? `set_scale_unlocked()` => `set_scale_locked()`

AlanGriffiths: Although, it would probably be better to have a lockable state member than a mix of `foo()` and `foo_locked()` member functions

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -48,18 +57,24 @@ class AccessibilityManager
...
-    bool enable_key_repeat;
 
+    // Always have a value supplied by platform options
+    bool enable_key_repeat;
     bool enable_mouse_keys;
 
     std::shared_ptr<mir::input::InputEventTransformer> const event_transformer;
     std::shared_ptr<mir::input::InputEventTransformer::Transformer> const transformer;
+
+    std::shared_ptr<graphics::Cursor> cursor;
```
AlanGriffiths, line 77:
```suggestion
    std::shared_ptr<graphics::Cursor> const cursor;
```

src/server/shell/accessibility_manager.cpp
```diff
@@ -114,12 +117,25 @@ struct MouseKeysTransformer: public mir::input::InputEventTransformer::Transform
...
+    cursor->set_scale(cursor_scale);
+
     if (enable_mouse_keys)
         event_transformer->append(transformer);
 }
+
+void mir::shell::AccessibilityManager::cursor_scale_changed(float new_scale)
+{
+    if(cursor)
+        cursor->set_scale(new_scale);
```
AlanGriffiths, line 140:
```suggestion
    cursor->set_scale(new_scale);
```

There is no need to test for null, the constructor would segfault if a null cursor is supplied. BTW that's in line with the project style of using _NullObject_ rather than repeated null checks.

src/miral/input_device_config.h
```diff
@@ -31,17 +31,20 @@ namespace miral
 struct MouseInputConfiguration
 {
     void apply_to(mir::input::Device& device) const;
+    void merge_settings_from(MouseInputConfiguration const&);
     std::optional<MirPointerHandedness> handedness;
     std::optional<MirPointerAcceleration> acceleration;
     std::optional<double> acceleration_bias;
     std::optional<double> vscroll_speed;
     std::optional<double> hscroll_speed;
+    std::optional<double> scale;
```
AlanGriffiths, line 40: The cursor scale is not part of the mouse input settings

include/miral/miral/input_configuration.h
```diff
@@ -78,6 +78,11 @@ class InputConfiguration::Mouse
     void acceleration_bias(std::optional<double>const& val);
     void vscroll_speed(std::optional<double>const& val);
     void hscroll_speed(std::optional<double>const& val);
+    void scale(std::optional<double>const& val);
```
AlanGriffiths, line 81: The cursor scale is not part of the mouse input settings

AlanGriffiths (changes requested):
I think two sets of changes have been conflated here:

1. Support for cursor scaling; and,
2. A bunch of input settings changes

Can we have these as separate PRs?

src/include/server/mir/shell/accessibility_manager.h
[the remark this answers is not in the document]
tarek-y-ismail: I think we should give shell authors the ability to modify each independently

src/server/graphics/software_cursor.cpp
tarek-y-ismail: My intention was that "set_scale_unlocked" was meant to be read as "set_scale, doesn't lock the mutex", but maybe your suggestion is better. Anyway, I'll try your second suggestion and see if that improves the code.

tarek-y-ismail: So the problem that `set_scaled_unlocked` was introduced to solve was that `set_scaled` needed to be called under two contexts. 
1. From external code  -> can lock immediately
2. When its called in `show`, where we first have to set the cursor image -> need to set the image first

Now, in a previous commit, I moved the code responsible for showing the cursor from `show` to `set_scale`. Reading it again now made me think that was not the best idea. So I think I'll move that code back and just call `show` under `set_scale`. I'll lock once to set the scale, unlock, then call `show`, which does its own locking. This should be pretty similar to a `Synchronized<T>`, but would require fewer code changes since we only need these changes in this narrow scope.

include/miral/miral/input_configuration.h
tarek-y-ismail: Moved scale setting and application to the only other place I know of that we can do that kind of stuff: `miral::CursorScale`.

examples/mir_demo_server/server_example.cpp
```diff
@@ -14,6 +14,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "miral/cursor_scale.h"
```
AlanGriffiths, line 17: Maybe put this with the other miral includes?

src/miral/input_device_config.cpp
```diff
@@ -273,7 +274,7 @@ miral::InputDeviceConfig::InputDeviceConfig(std::shared_ptr<mir::options::Option
         to_acceleration_profile(get_optional<std::string>(options, mouse_cursor_acceleration_opt)),
         clamp<-1.0, 1.0>(get_optional<double>(options, mouse_cursor_acceleration_bias_opt)),
         get_optional<double>(options, mouse_vscroll_speed_override_opt, mouse_scroll_speed_opt),
-        get_optional<double>(options, mouse_hscroll_speed_override_opt, mouse_scroll_speed_opt)
+        get_optional<double>(options, mouse_hscroll_speed_override_opt, mouse_scroll_speed_opt),
```
AlanGriffiths, line 277: Unnecessary

src/miroil/mir_server_hooks.cpp
```diff
@@ -14,6 +14,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <memory>
 #include <miroil/mir_server_hooks.h>
```
AlanGriffiths, line 18: The first header should be the one corresponding to the implementation file

AlanGriffiths (commented):
Nitpicking again

AlanGriffiths (changes requested):
We should provide easy to use abstractions, not leak implementation complexity

src/miral/cursor_scale.cpp
```diff
@@ -0,0 +1,95 @@
...
+    void set_scale(float new_scale)
+    {
+        scale = new_scale;
+    }
+
+    void apply_scale() const
+    {
+        if(accessibility_manager.expired())
+        {
+            mir::log_warning("AccessibilityManager not available. Will not set cursor scale");
```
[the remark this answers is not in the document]
tarek-y-ismail: My initial point of view was that cursor scaling wasn't that important to justify bringing the compositor down. But since your comment implies that end users are not likely to be able to fix this, then it might be a better idea to crash the compositor to force shell authors to handle this.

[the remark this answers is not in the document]
tarek-y-ismail: Taking the comment right above this into account, I can't both provide one method _and_ throw an exception if its called before the server is started. because we need to allow shell authors to set the cursor scale value before the server starts (which would throw), and set it after the server starts.

My preferred approach is to not throw and have one method. What do you think?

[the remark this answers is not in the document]
AlanGriffiths: Yes, an API that just "does the right thing" is preferable to one that requires the caller to get it right

AlanGriffiths (changes requested):
Looking good...

include/platform/mir/graphics/pixman_image_scaling.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef MIR_GRAPHICS_PIXMAN_IMAGE_SCALING_H_
+#define MIR_GRAPHICS_PIXMAN_IMAGE_SCALING_H_
+
+#include "mir/geometry/size.h"
+#include <memory>
+
+namespace mir::graphics
+{
+struct ARGB8888Buffer
```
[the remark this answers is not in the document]
tarek-y-ismail: `ARGB8Buffer`?

src/miral/cursor_scale.cpp
```diff
@@ -0,0 +1,77 @@
...
+        _scale = new_scale;
+
+        if(accessibility_manager.expired())
+            return;
+
+        accessibility_manager.lock()->cursor_scale_changed(*_scale);
+    }
+
+    std::optional<float> _scale;
+    std::weak_ptr<mir::shell::AccessibilityManager> accessibility_manager;
```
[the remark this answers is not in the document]
tarek-y-ismail: No can do, `accessibility_manager` is set after `Self` is created when the server is being initialized. But we also need `Self` to exist before then so any changes made before the server is initialized are stored.

src/miral/cursor_scale.cpp
```diff
@@ -0,0 +1,77 @@
...
+    {
+        _scale = new_scale;
+
+        if(accessibility_manager.expired())
+            return;
+
+        accessibility_manager.lock()->cursor_scale_changed(*_scale);
+    }
+
+    std::optional<float> _scale;
```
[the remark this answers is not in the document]
tarek-y-ismail: For 3, we could add a constructor argument `override` that if set, would make the code ignore the config option. Is this what you had in mind?

src/miral/cursor_scale.cpp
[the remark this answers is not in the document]
AlanGriffiths: Then we probably need a comment to explain why this is threadsafe

src/miral/cursor_scale.cpp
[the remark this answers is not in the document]
AlanGriffiths: No. I'm thinking that the default scale must always be provided to the `Self` constructor.

I don't think the config option should be ignored: however, the config default set in `CursorScale::operator()` should be whatever is passed to the `CursorScale` constructor (or a default from the default constructor)

AlanGriffiths (changes requested):

src/miral/cursor_scale.cpp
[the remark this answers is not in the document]
tarek-y-ismail: Oh, I missed your point about `_scale` being optional. I think this is necessary to have some idea whether a custom scale was provided or not.

My understanding is that values configured via `CursorScale` should take precedence over config options like `cursor_scale_opt`. If I ditch the `std::optional`, I have no way of knowing whether a non-default scale value was provided. On top of this, since I provided a default value in `add_configure_option`, `Options::is_set` always returns true. So I can't figure out whether the config option was really set or not.

If instead we keep the optional value, and introduce `float const default_value`. We can pass the default value from the constructor to `add_configure_option`, and have a way of knowing whether or not `scale` was called with a value, and can preserve semantics. 

I'm probably not doing a very good job of explaining things, so commit incoming!

src/miral/cursor_scale.cpp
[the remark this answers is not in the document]
tarek-y-ismail: I don't think it's particularly threadsafe. For example, in mir_demo_server, if the user changes a config file while the server is starting, then we have a race on both `accessibility_manager` and `scale`.

AlanGriffiths (changes requested):

src/miral/cursor_scale.cpp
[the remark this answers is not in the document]
AlanGriffiths: > Don't watches run on their own thread?

Yes. But so long as `self->accessibility_manager` only changed while the program is single threaded all threads, including the watcher thread, will see the same value

src/miral/cursor_scale.cpp
```diff
@@ -24,26 +24,34 @@
...
-    std::optional<float> _scale;
-    std::weak_ptr<mir::shell::AccessibilityManager> accessibility_manager;
+    float const default_scale;
+
+    struct State {
+        std::optional<float> scale;
+        std::weak_ptr<mir::shell::AccessibilityManager> accessibility_manager;
+    };
+
+    mir::Synchronised<State> state;
```
AlanGriffiths, line 50: As discussed, there's no need to guard `accessibility_manager`, once multiple threads have started it is effectively "const".

There is no need for a separate `default_scale` in addition to `scale`.

There is a need to guard `scale`, but `Synchronised<State>` is likely overkill.

```suggestion

    std::atomic<float> scale;
    
    // accessibility_manager is only updated during the single-threaded initialization phase of startup
    std::weak_ptr<mir::shell::AccessibilityManager> accessibility_manager;
```

src/miral/cursor_scale.cpp
```diff
@@ -63,14 +71,18 @@ void miral::CursorScale::operator()(mir::Server& server) const
 {
     auto const* const cursor_scale_opt = "cursor-scale";
     server.add_configuration_option(
-        cursor_scale_opt, "Scales the mouse cursor visually. Accepts any value in the range [0, 100]", 1.0);
+        cursor_scale_opt, "Scales the mouse cursor visually. Accepts any value in the range [0, 100]", self->default_scale);
```
AlanGriffiths, line 74:
```suggestion
        cursor_scale_opt, "Scales the mouse cursor visually. Accepts any value in the range [0, 100]", self->scale);
```

src/miral/cursor_scale.cpp
```diff
@@ -0,0 +1,89 @@
...
+        {
+            auto const& options = server.get_options();
+
+            auto const scale = [&]{
+                auto const locked_state = self->state.lock();
+                locked_state->accessibility_manager = server.the_accessibility_manager();
+                return locked_state->scale.value_or(options->get<double>(cursor_scale_opt));
+            }();
+
+            self->scale(scale);
```
AlanGriffiths, line 87: This provides the user a configuration option that mysteriously doesn't work if the shell author has set a scale. That is a bad user experience.

If we want a standard configuration option, then we should only allow the shell author to provide the default (which we do elsewhere)

```suggestion
            self->accessibility_manager = server.the_accessibility_manager();
            auto const scale = server.get_options()->get<double>(cursor_scale_opt);
            
            self->scale(scale);
```

AlanGriffiths (changes requested):
Having thought about this. We should probably not offer a configuration option for the cursor scale: It is easy for the shell author to provide this for themselves, ideally from gsettings or a "live" configuration file (not the command line or environment).

NB We would still need to apply `self::scale` when setting collecting the `accessibility_manager` in the init callback, but we don't need to worry where the value came from

tarek-y-ismail:
> NB We would still need to apply self::scale when setting collecting the accessibility_manager in the init callback, but we don't need to worry where the value came from

Can't quite follow. What are you trying to say here?

src/platforms/gbm-kms/server/kms/cursor.cpp
```diff
@@ -253,16 +255,20 @@ void mgg::Cursor::pad_and_write_image_data_locked(
     write_buffer_data_locked(lg, buffer, &padded[0], padded_size);
 }
 
-void mgg::Cursor::show(CursorImage const& cursor_image)
+void mgg::Cursor::show(std::shared_ptr<CursorImage> const& cursor_image)
 {
     std::lock_guard lg(guard);
 
-    size = cursor_image.size();
+    this->current_cursor_image = cursor_image;
```
AlanGriffiths, line 262: Do we need the `this->` wart?

src/server/graphics/software_cursor.h
```diff
@@ -45,12 +47,13 @@ class SoftwareCursor : public Cursor
         std::shared_ptr<input::Scene> const& scene);
     ~SoftwareCursor();
 
-    void show(CursorImage const& cursor_image) override;
+    void show(std::shared_ptr<CursorImage> const& cursor_image) override;
     void hide() override;
     void move_to(geometry::Point position) override;
+    void set_scale(float) override;
```
AlanGriffiths, line 53: We tend to avoid `set_/get_` warts as they add noise, but not readability

AlanGriffiths (approved):
Some minor naming nits: Feel free to land after updating

Issue #3739, opened 2025-02-06
`WaylandExecutor` should implement `time::AlarmFactory` (and probably the rest of `MainLoop`?)


RAOF:
The Wayland event loop pretty much implements our `MainLoop` interfaces - timers, signal watches, fd watches. When interacting with Wayland it's almost always necessary to do so *from* the Wayland event loop, so currently when we want to watch a file descriptor and then do something on the Wayland loop, or set a timer to do something on the Wayland loop we need to first set the timer, and then from the callback push *another* callback into the `WaylandExecutor` that can then be run on the Wayland loop.

Instead we could just have a Wayland-main-loop-backed implementation of `MainLoop` that does everything on the event loop, making everything cleaner.

PR #3740, opened 2025-02-06
Improve Firefox `surface.frame` event quirk

RAOF:
We currently queue a `send_frame_events()` 16ms after receiving
a `surface.commit()` with no buffer but that requests a frame
event.

When `send_frame_events()` is triggered it sends *all* the frame
events at the time of trigger; this results in us sending frame
events at a sub-optimal time and breaks the way that WLCS uses
the frame event to determine when a buffer has been used.

Instead, maintain a separate `heartbeat_quirk_frame_events` list
that gets sent by the 16ms timer.

### DIFF
--- a/src/server/frontend_wayland/wl_surface.cpp
+++ b/src/server/frontend_wayland/wl_surface.cpp
@@ -452,13 +452,17 @@ void mf::WlSurface::commit(WlSurfaceState const& state)
                 end(state.frame_callbacks));
 
             frame_callback_executor->spawn(
-                [weak_self = mw::make_weak(this)]
+                [executor = wayland_executor, weak_self = mw::make_weak(this)]
                 {
-                    if (weak_self)
-                    {
-                        auto& self = weak_self.value();
-                        self.send_frame_callbacks(self.heartbeat_quirk_frame_callbacks);
-                    }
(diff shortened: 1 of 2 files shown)


RAOF:
I think we might also want to fix issue #3739 while we're here, but I didn't want to block this on that. I've had a look, and it's not *much* work, but there's some subtlety around `wl_event_loop` lifetime.

AlanGriffiths (commented):
Looks sensible, it isn't immediately obvious why the wp-viewporter tests are failing. But as they sync using the frame callback there's likely an implicit assumption that no longer holds

RAOF:
Ok! The problem was that the new code was broken! (it was never triggered). The WpViewporter tests happen to use buffer-less frame events, so that's why those broke.

It seems that Firefox is more tolerant of not sending those events than it was (although, now that I've *actually* fixed it I *can* see that there are behavioural differences that just seemed like regular weirdness :woman_shrugging: )

RAOF:
There's now also a WLCS test that catches this: https://github.com/canonical/wlcs/pull/350

AlanGriffiths (approved):
Ah, that's what we missed!

Issue #3677, opened 2024-11-19
atomic-kms: hardware cursor disappears crossing screen boundaries


Saviq:
Steps:
- run Mir with `--display-config=sidebyside --platform-display-libs=mir:atomic-kms --cursor=hardware`
- move the mouse across the screen boundary

Current:
- cursor disappears

Expected
- cursor workable

Workaround:
- `--cursor=software`

PR #3742, opened 2025-02-07
platforms/atomic-kms: Fix cursors across multiple monitors.

RAOF:
`KMSOutput::has_cursor` was misleadingly named, and so we implemented it incorrectly. It is *used* to tell whether or not a cursor image needs to be set on that particular output before it is visible, *not* whether that output *supports* a cursor.

Rename it to something hopefully more easily understandable, and then fix the implementation.

Fixes: #3677


AlanGriffiths (approved):
gbm-kms has similar code (but not broken), perhaps the names there should be kept in step?

PR #3749, opened 2025-02-10
Extend the lifetime of the accessibility manager to keep enable-key-repeat

tarek-y-ismail:
closes #3743

PR #3754, opened 2025-02-11
Introduce `InputEventTransformer` to read and synthesize input events

tarek-y-ismail:
Some notes:
1. The latter half of this PR (the accessibility manager demo) is just for making sure its functionality works properly. ~I will move this commit to the mousekeys  PR after this one is approved.~ Well, since that PR is based on this, it doesn't matter if I leave it here or move it there, so...
2. Currently, there's a segfault when shutting down mir thrown somewhere while destructing `SeatInputDeviceTracker::touch_visualizer`. It's not entirely clear what's the cause, but I have a "workaround" for it. ~I'm still trying to find the root cause~ #3768

### DIFF
--- a/src/include/server/mir/input/input_event_transformer.h
+++ b/src/include/server/mir/input/input_event_transformer.h
@@ -36,17 +36,13 @@ class EventBuilder;
 class InputEventTransformer : public EventFilter
 {
 public:
-    struct EventDispatcher
+    class EventDispatcher
     {
+    public:
         EventDispatcher(
             std::shared_ptr<MainLoop> const main_loop,
             input::InputSink* const sink,
-            input::EventBuilder* const builder) :
-            main_loop{main_loop},
-            sink{sink},
-            builder{builder}
-        {
-        }
+            input::EventBuilder* const builder);
 
         void dispatch_key_event(
             std::optional<std::chrono::nanoseconds> timestamp,
@@ -72,8 +68,9 @@ public:
         input::EventBuilder* const builder;
     };
 
-    struct Transformer
+    class Transformer
     {
+    public:
         virtual ~Transformer() = default;
 
         // Returning true means that the event has been successfully processed and
@@ -89,7 +86,6 @@ public:
     bool handle(MirEvent const&) override;
 
     void append(std::weak_ptr<Transformer> const&);
-    void prepend(std::weak_ptr<Transformer> const&);
 
 private:
     void lazily_init_virtual_input_device();
--- a/src/include/server/mir/input/input_event_transformer.h
+++ b/src/include/server/mir/input/input_event_transformer.h
@@ -19,8 +19,10 @@
 
 #include "mir/input/event_filter.h"
 #include "mir/input/event_builder.h"
+#include "mir_toolkit/events/event.h"
 
 #include <chrono>
+#include <functional>
 #include <mutex>
 #include <vector>
 #include <memory>
@@ -37,47 +39,7 @@ class EventBuilder;
 class InputEventTransformer : public EventFilter
 {
 public:
-    class EventDispatcher
-    {
-    public:
-        EventDispatcher(
-            std::shared_ptr<MainLoop> const main_loop,
-            input::InputSink* const sink,
-            input::EventBuilder* const builder);
-
-        void dispatch_key_event(
-            std::optional<std::chrono::nanoseconds> timestamp,
-            MirKeyboardAction action,
-            xkb_keysym_t keysym,
-            int scan_code);
-
-        void dispatch_pointer_event(
-            std::optional<std::chrono::nanoseconds> timestamp,
-            MirPointerAction action,
-            MirPointerButtons buttons,
-            std::optional<mir::geometry::PointF> position,
-            mir::geometry::DisplacementF motion,
-            MirPointerAxisSource axis_source,
-            events::ScrollAxisH h_scroll,
-            events::ScrollAxisV v_scroll);
-
-        void maybe_update(InputSink* maybe_new_sink, EventBuilder* maybe_new_builder);
-
-    private:
-        void dispatch(mir::EventUPtr e);
-
-        std::shared_ptr<MainLoop> const main_loop;
-
-        // Since dispatchers are handed down to transformers which might keep a
-        // reference to them, then we should be careful to not change the
-        // builder as they dispatch an event.
-        std::mutex mutex;
-
-        // Sinks and/or builders are not constant for one virtual device.
-        input::InputSink* sink;
-        input::EventBuilder* builder;
-    };
-
+    using EventDispatcher = std::function<void(std::shared_ptr<MirEvent>)>;
     class Transformer
     {
     public:
@@ -87,7 +49,7 @@ public:
         // shouldn't be handled by later transformers, whether the transformer is
         // accumulating events for later dispatching or has immediately dispatched
         // an event is an implementation detail of the transformer
-        virtual bool transform_input_event(std::shared_ptr<EventDispatcher> const& dispatcher,  MirEvent const&) = 0;
+        virtual bool transform_input_event(EventDispatcher const& dispatcher, EventBuilder*,  MirEvent const&) = 0;
     };
 
     InputEventTransformer(std::shared_ptr<InputDeviceRegistry> const&, std::shared_ptr<MainLoop> const&);
@@ -98,12 +60,10 @@ public:
     void append(std::weak_ptr<Transformer> const&);
 
 private:
-
     std::mutex mutex;
     std::vector<std::weak_ptr<Transformer>> input_transformers;
-    std::shared_ptr<input::VirtualInputDevice> virtual_pointer;
-    std::shared_ptr<EventDispatcher> dispatcher;
 
+    std::shared_ptr<input::VirtualInputDevice> const virtual_pointer;
     std::shared_ptr<input::InputDeviceRegistry> const input_device_registry;
     std::shared_ptr<MainLoop> const main_loop;
 };
--- a/src/include/server/mir/input/input_event_transformer.h
+++ b/src/include/server/mir/input/input_event_transformer.h
@@ -1,0 +1,108 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_INPUT_INPUT_EVENT_TRANSFORMER_H_
+#define MIR_INPUT_INPUT_EVENT_TRANSFORMER_H_
+
+#include "mir/input/event_filter.h"
+#include "mir/input/event_builder.h"
+
+#include <chrono>
+#include <vector>
+#include <memory>
+
+namespace mir
+{
+class MainLoop;
+namespace input
+{
+class VirtualInputDevice;
+class InputDeviceRegistry;
+class InputSink;
+class EventBuilder;
+class InputEventTransformer : public EventFilter
+{
+public:
+    struct EventDispatcher
+    {
+        EventDispatcher(
+            std::shared_ptr<MainLoop> const main_loop,
+            input::InputSink* const sink,
+            input::EventBuilder* const builder) :
+            main_loop{main_loop},
+            sink{sink},
+            builder{builder}
+        {
+        }
+
+        void dispatch_key_event(
+            std::optional<std::chrono::nanoseconds> timestamp,
+            MirKeyboardAction action,
+            xkb_keysym_t keysym,
+            int scan_code);
+
+        void dispatch_pointer_event(
+            std::optional<std::chrono::nanoseconds> timestamp,
+            MirPointerAction action,
+            MirPointerButtons buttons,
+            std::optional<mir::geometry::PointF> position,
+            mir::geometry::DisplacementF motion,
+            MirPointerAxisSource axis_source,
+            events::ScrollAxisH h_scroll,
+            events::ScrollAxisV v_scroll);
+
+    private:
+        void dispatch(mir::EventUPtr e);
+
+        std::shared_ptr<MainLoop> const main_loop;
+        input::InputSink* const sink;
+        input::EventBuilder* const builder;
+    };
+
+    struct Transformer
+    {
+        virtual ~Transformer() = default;
+
+        // Returning true means that the event has been successfully processed and
+        // shouldn't be handled by later transformers, whether the transformer is
+        // accumulating events for later dispatching or has immediately dispatched
+        // an event is an implementation detail of the transformer
+        virtual bool transform_input_event(std::shared_ptr<EventDispatcher> const& dispatcher,  MirEvent const&) = 0;
+    };
+
+    InputEventTransformer(std::shared_ptr<InputDeviceRegistry> const&, std::shared_ptr<MainLoop> const&);
+    ~InputEventTransformer();
+
+    bool handle(MirEvent const&) override;
+
+    void append(std::weak_ptr<Transformer> const&);
+    void prepend(std::weak_ptr<Transformer> const&);
+
+private:
+    void lazily_init_virtual_input_device();
+
+    std::mutex mutex;
+    std::vector<std::weak_ptr<Transformer>> input_transformers;
+    std::shared_ptr<input::VirtualInputDevice> virtual_pointer;
+    std::shared_ptr<EventDispatcher> dispatcher;
+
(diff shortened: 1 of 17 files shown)


examples/mir_demo_server/server_example.cpp
```diff
@@ -142,8 +142,8 @@ class InputConfigFile
     miral::InputConfiguration::Mouse mouse = input_configuration.mouse();
     miral::InputConfiguration::Touchpad touchpad = input_configuration.touchpad();
     miral::InputConfiguration::Keyboard keyboard = input_configuration.keyboard();
-    miral::ConfigFile config_file;
     std::mutex config_mutex;
+    miral::ConfigFile config_file;
```
tarek-y-ismail, line 146: Without this line, the demo server crashes because we try locking the mutex before it's initialized. This change comes originally from #3735.

src/server/frontend_wayland/wayland_default_configuration.cpp
tarek-y-ismail, line 1: These changes are also necessary so that the accessibility manager isn't constructed/destructed redundantly. Originally come from #3749

examples/mir_demo_server/server_example.cpp
AlanGriffiths: This looks unrelated to the `InputEventTransformer` and could be landed in its own PR. (Looks trivial to review too)

AlanGriffiths (dismissed):

tarek-y-ismail:
@AlanGriffiths pretty sure I'm done with this at the moment. Feel free to review. I've kept the "demo" for testing, but I'll drop it once the PR is approved.

src/server/input/input_event_transformer.cpp
```diff
@@ -0,0 +1,131 @@
...
+    lazily_init_virtual_input_device();
+    input_transformers.insert(input_transformers.begin(),transformer);
+}
+
+void mi::InputEventTransformer::lazily_init_virtual_input_device()
+{
+    if (input_transformers.empty())
+    {
+        virtual_pointer =
+            std::make_shared<input::VirtualInputDevice>("mousekey-pointer", mir::input::DeviceCapability::pointer);
```
mattkae, line 99: Why are lazily initializing this?

src/server/input/input_event_transformer.cpp
```diff
@@ -0,0 +1,131 @@
...
+            for (auto it = input_transformers.begin(); it != input_transformers.end(); ++it)
+            {
+                auto const& t = it->lock();
+                if (!t)
+                {
+                    it = input_transformers.erase(it);
+                    if(it == input_transformers.end())
+                        return;
+                    continue;
+                }
```
mattkae, line 67:
```suggestion
                if (!t)
                {
                    it = input_transformers.erase(it);
                    continue;
                }
                else
                    it++;
```

src/server/input/input_event_transformer.cpp
```diff
@@ -0,0 +1,131 @@
...
+        return false;
+
+    auto handled = false;
+    virtual_pointer->if_started_then(
+        [this, &event, &handled](mir::input::InputSink* sink, mir::input::EventBuilder* builder)
+        {
+            if (!dispatcher)
+                dispatcher = std::make_shared<EventDispatcher>(main_loop, sink, builder);
+
+            for (auto it = input_transformers.begin(); it != input_transformers.end(); ++it)
```
mattkae, line 58:
```suggestion
            for (auto it = input_transformers.begin(); it != input_transformers.end();)
```

src/server/shell/accessibility_manager.cpp
```diff
@@ -42,12 +51,80 @@ void mir::shell::AccessibilityManager::repeat_delay(int new_delay) {
...
+
+            dispatcher->dispatch_pointer_event(
+                std::nullopt,
+                mir_pointer_action_motion,
+                0,
+                std::nullopt,
+                offset,
+                mir_pointer_axis_source_none,
+                mir::events::ScrollAxisH{},
+                mir::events::ScrollAxisV{});
```
mattkae, line 111: Question: can `transform_input_event` both modify and create new events? Or can it only create new events?

mattkae (changes requested):
Looks mostly reasonable! I've got a few things about naming and roles of objects that I would like to understand

src/server/shell/accessibility_manager.cpp
mattkae: @tarek-y-ismail this is my main question by the way :)

tarek-y-ismail: Since we get the input event by const ref, we can't modify it. We can either "consume" the event or let another transformer handle it. We also can dispatch a new one depending on the input. Consumption of the existing event and the synthesization of a new one are not mutually exclusive by the way.

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -0,0 +1,108 @@
...
+
+    private:
+        void dispatch(mir::EventUPtr e);
+
+        std::shared_ptr<MainLoop> const main_loop;
+        input::InputSink* const sink;
+        input::EventBuilder* const builder;
+    };
+
+    struct Transformer
```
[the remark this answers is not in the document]
tarek-y-ismail: Never said the names are final :) Maybe `InputEventTransformerManager` and `Transformer`? Though I think the former is a bit too long and encroaching on Java territory

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -0,0 +1,108 @@
...
+        virtual bool transform_input_event(std::shared_ptr<EventDispatcher> const& dispatcher,  MirEvent const&) = 0;
+    };
+
+    InputEventTransformer(std::shared_ptr<InputDeviceRegistry> const&, std::shared_ptr<MainLoop> const&);
+    ~InputEventTransformer();
+
+    bool handle(MirEvent const&) override;
+
+    void append(std::weak_ptr<Transformer> const&);
+    void prepend(std::weak_ptr<Transformer> const&);
```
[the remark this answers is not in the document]
tarek-y-ismail: If you mean `prepend`, then no, not at the moment. I just thought it might be useful to have for the future.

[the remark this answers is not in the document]
tarek-y-ismail: Hmm, we can add it when we need it anyway

src/server/input/input_event_transformer.cpp
tarek-y-ismail: Some accessibility settings might not require a virtual device (key repeat, cursor scale), so I thought it would be wasteful if we created the device on construction. Instead, we create it when a transformer is added so we know it will be used from that point onward.

src/include/server/mir/input/input_event_transformer.h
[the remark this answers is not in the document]
mattkae: :) 

Yeah I hate to nitpick, but when I read the class, I assumed that it was transforming input events from `A -> B`. Now that I've seen the implementation, it seems more like an `InputEventProcessor` or something like that, which takes events, does _some_ processing on the, and outputs new events.

src/server/input/input_event_transformer.cpp
mattkae: Shouldn't we just be initializing it when the `virtual_pointer` is `nullptr` rather than when the `input_transformers.empty()`?

mattkae: Also - I doubt it would be much of a performance hit to have this always initialized :)

tarek-y-ismail:
I'm testing this at the moment and it seems to crash at `EventBuilder::pointer_event` which doesn't make sense. It always crashes at the same line if and only if I open firefox and try to call the method. I'm investigating this at the moment.

src/include/server/mir/input/input_event_transformer.h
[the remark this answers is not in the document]
tarek-y-ismail: Processor might give off the wrong idea that the output of one transformer can be the input of another (which I guess is true, but not in the sense of a pipeline, which is what I had in mind when reading `InputEventProcessor`)

src/server/input/input_event_transformer.cpp
tarek-y-ismail: Sure, if you insist. We can change it later on if it proves to be an issue

tarek-y-ismail:
Aaand tests are failing because we're adding a new input device. @mattkae should I jump head-first into fixing the tests or should I revert adding the device on construction?

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -0,0 +1,113 @@
...
+    class Transformer
+    {
+    public:
+        virtual ~Transformer() = default;
+
+        // Returning true means that the event has been successfully processed and
+        // shouldn't be handled by later transformers, whether the transformer is
+        // accumulating events for later dispatching or has immediately dispatched
+        // an event is an implementation detail of the transformer
+        virtual bool transform_input_event(std::shared_ptr<EventDispatcher> const& dispatcher,  MirEvent const&) = 0;
```
mattkae, line 90: I am of the opinion that this would be easier without the need for the `EventDispatcher` class at all, especially since the event builder class will have to keep in sync with anything that could be built (e.g. touch events). 

```suggestion
        virtual bool transform_input_event(MirEvent const&, input::InputSink* sink, input::EventBuilder* builder, std::function<void(mir::EventUPtr e) const& dispatch) = 0;
```

I think this gets rid of some unnecessary complexity. What do you think?

mattkae (changes requested):
An architecture question

src/include/server/mir/input/input_event_transformer.h
tarek-y-ismail: I originally added `EventDispatcher` to hoist the issuing of the even onto another thread (the whole purpose of `main_loop` in the struct). You can see for yourself, just apply these changes and the first movement should occur, but then miral should deadlock.

```diff
diff --git a/src/include/server/mir/input/input_event_transformer.h b/src/include/server/mir/input/input_event_transformer.h
index 2153dc90d1..801821a425 100644
--- a/src/include/server/mir/input/input_event_transformer.h
+++ b/src/include/server/mir/input/input_event_transformer.h
@@ -63,7 +63,6 @@ public:
 
         void maybe_update(InputSink* maybe_new_sink, EventBuilder* maybe_new_builder);
 
-    private:
         void dispatch(mir::EventUPtr e);
 
         std::shared_ptr<MainLoop> const main_loop;
diff --git a/src/server/shell/accessibility_manager.cpp b/src/server/shell/accessibility_manager.cpp
index f116f21866..c15482ed0e 100644
--- a/src/server/shell/accessibility_manager.cpp
+++ b/src/server/shell/accessibility_manager.cpp
@@ -100,7 +100,7 @@ struct MouseKeysTransformer: public mir::input::InputEventTransformer::Transform
                 return false;
             }
 
-            dispatcher->dispatch_pointer_event(
+            dispatcher->sink->handle_input(dispatcher->builder->pointer_event(
                 std::nullopt,
                 mir_pointer_action_motion,
                 0,
@@ -108,7 +108,7 @@ struct MouseKeysTransformer: public mir::input::InputEventTransformer::Transform
                 offset,
                 mir_pointer_axis_source_none,
                 mir::events::ScrollAxisH{},
-                mir::events::ScrollAxisV{});
+                mir::events::ScrollAxisV{}));
 
             return true;
         }
```

mattkae: I think I misspoke when I said you should pass the `input_sink` :)

Still, you could just pass the `builder` and a `dispatch` function as parameters to this callback and avoid the `EventDispatcher` altogether. 

e.g.

```c++
// input_event_transformer.cpp
bool mi::InputEventTransformer::handle(MirEvent const& event)
{
    ...
    auto dispatch = [sink=sink, main_loop=main_loop](mir::EventUPtr e) {
        main_loop->spawn(
            [e = std::shared_ptr(std::move(e)), this]
            {
                sink->handle_input(e);
            });
    };

    if (t->transform_input_event(event, builder, dispatch)) { ... }
}
```

The builder part of the event does not have to be thread safe. Only the dispatch part does. Hence, we can remove a lot of complexity by capturing it in a lambda. Then, we can do:

```c++
// accessibility_manager.cpp

dispatch(builder->pointer_event(...))
```

tarek-y-ismail: In the context of #3757, this might break since we'll capture the sink and builder and keep them as long as the user holds one of the keys used to move the mouse

mattkae: I don't understand how it is any different from what you have however, since it just removes the class but keeps all of the same functionality. Perhaps I'm missing something?

tarek-y-ismail: `std::shared_ptr<EventDispatcher>` adds an extra level of indirection that should allow the pointers to be updated when they change. Now, I investigate a little more and the pointers should not change, so I'm very confused at the moment

tarek-y-ismail:
@mattkae Should be safe to review again

src/server/shell/accessibility_manager.cpp
```diff
@@ -42,12 +52,81 @@ void mir::shell::AccessibilityManager::repeat_delay(int new_delay) {
...
+    bool transform_input_event(
+        mir::input::InputEventTransformer::EventDispatcher const& dispatcher,
+        mir::input::EventBuilder* builder,
+        MirEvent const& event) override
+    {
+        using namespace mir; // For operator<<
+
+        std::stringstream ss;
+        ss << const_cast<MirEvent&>(event);
+        mir::log_debug("%s", ss.str().c_str());
```
mattkae, line 71: Does there not just exist a "toString()" method here to avoid creating the string stream? If not, can we add one? e.g.

```suggestion
        mir::log_debug("%s", event.toString());
```

src/server/shell/accessibility_manager.cpp
```diff
@@ -42,12 +52,81 @@ void mir::shell::AccessibilityManager::repeat_delay(int new_delay) {
...
+                offset = geometry::DisplacementF{0, 10};
+                break;
+            case XKB_KEY_KP_4:
+                offset = geometry::DisplacementF{-10, 0};
+                break;
+            case XKB_KEY_KP_6:
+                offset = geometry::DisplacementF{10, 0};
+                break;
+            case XKB_KEY_KP_8:
+                offset = geometry::DisplacementF{0, -10};
```
mattkae, line 98: Should this offset (i.e. `-10`) be configurable somehow?

mattkae (changes requested):
I like this a lot more! Should we add tests too :innocent: ?

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -0,0 +1,73 @@
...
+        // an event is an implementation detail of the transformer
+        virtual bool transform_input_event(EventDispatcher const& dispatcher, EventBuilder*,  MirEvent const&) = 0;
+    };
+
+    InputEventTransformer(std::shared_ptr<InputDeviceRegistry> const&, std::shared_ptr<MainLoop> const&);
+    ~InputEventTransformer();
+
+    bool handle(MirEvent const&) override;
+
+    void append(std::weak_ptr<Transformer> const&);
```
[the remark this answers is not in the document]
tarek-y-ismail: Once all references to the `shared_ptr` backing the `weak_ptr` are destroyed, the `weak_ptr` gets destroyed on the next `handle` invocation

src/server/shell/accessibility_manager.cpp
tarek-y-ismail: The only thing I could find was `operator<<` in `event_printer.h`. I guess since this is mostly debug code that it doesn't matter if its slightly messy. What do you think?

src/server/shell/accessibility_manager.cpp
tarek-y-ismail: That's in #3757. This code is just for basic testing

src/server/shell/accessibility_manager.cpp
mattkae: Eh I don't really care, but it is just another object to create every time an input event happens, which could get a bit costly.

tests/unit-tests/input/test_input_event_transformer.cpp
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    std::shared_ptr<mir::input::DefaultInputDeviceHub> const input_device_hub;
+    mir::input::InputEventTransformer input_event_transformer;
+};
+
+struct StubTransformer : public mir::input::InputEventTransformer::Transformer
+{
+    std::function<bool()> const on_event;
+    bool called{false};
```
mattkae, line 83: This implementation sounds like you're looking for a mock class instead

mattkae (changes requested):
Two minor things, but we are almost landing this! I'm testing it now too

src/server/shell/accessibility_manager.cpp
tarek-y-ismail: We'll it is debug code after all. Anyway, I removed these few lines in the latest batch of commits

src/include/server/mir/input/input_event_transformer.h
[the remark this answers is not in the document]
tarek-y-ismail: I'm not opposed to this. I think I should also swap the parameter to a `shared_ptr` so it's clear that it has to be explicitly removed. Would save us some work in `handle` as well ;)

mattkae (approved):
I am officially happy, nice work!

PR #3756, opened 2025-02-12
Make sure the config file mutex is initialized before being used

tarek-y-ismail:
Since the order of initialization is the same as the order of declaration, the config file was initialized before the mutex. But it would use the mutex during its construction since it loads on construction, causing an error to be thrown and the application to crash.

### DIFF
--- a/examples/mir_demo_server/server_example.cpp
+++ b/examples/mir_demo_server/server_example.cpp
@@ -138,12 +138,12 @@ public:
     }
 
 private:
+    std::mutex config_mutex;
     miral::InputConfiguration input_configuration;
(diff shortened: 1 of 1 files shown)


AlanGriffiths (commented):
Non-blocking comment: It might be clearer to place the mutex above the `mouse`, `touchpad` and `keyboard` variables as their use is actually what it guards. (Not that the code is particularly clear about this either way - ideas welcome)

tarek-y-ismail:
Seems like I sneaked in this change in another small PR and forgot. This PR is not needed anymore.

PR #3757, opened 2025-02-13
Implement mousekeys

tarek-y-ismail:
Implements mousekey and adds some basic configuration options:
- enable-mouse-keys
- mouse-keys-acceleration-constant-factor
- mouse-keys-acceleration-linear-factor
- mouse-keys-acceleration-quadratic-factor
- mouse-keys-max-speed-x
- mouse-keys-max-speed-y

To test:
- Add `miral::MouseKeysConfig` to the mir runner in your server to enable mousekeys if not already there
- Run mir with `--enable-mouse-keys=1` or toggle it at any time by using `MouseKeysConfig::enabled`
- Controls are similar to [KDE's mousekeys controls](https://docs.kde.org/stable5/en/plasma-desktop/kcontrol/kcmaccess/index.html). 2,4,6,8 for movement, 5 for clicks, /,*,- for left,middle, and right mouse button selection, + for double clicks, 0,. to start/stop dragging.

Notes:
- #3768 affects this PR. The changes introduced here trigger it, but are not the root cause for it.


tarek-y-ismail (commented):
Self review

src/miral/enable_mousekeys.cpp
```diff
@@ -0,0 +1,86 @@
...
+        auto const* input_event = mir_event_get_input_event(&event);
+        if(mir_input_event_get_type(input_event) != mir_input_event_type_key)
+            return false;
+
+
+        auto const* key_event = mir_input_event_get_keyboard_event(input_event);
+        auto const modifiers = mir_keyboard_event_modifiers(key_event);
+
+        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) &&
+            (modifiers & mir_input_event_modifier_num_lock))
```
mattkae, line 44: This modifier matching should be provided by the user of `EnableMouseKeys` instead of hard coded.

mattkae: Actually, should is this the right thing to do at all? Compositor authors may want to toggle the mouse keys accessibility feature through a number of ways (key press, mouse click, internal client, dbus interface, etc.).

Instead of exposing an interface that adds an event filter, we should expose an interface that allows users to toggle on and off mouse keys. Then, in their compositor (`shell_main.cpp` for example), they can turn on/off mouse keys based off of anything. In this case, it will be based off of this particualr key press event.

mattkae (changes requested):
I have an initial big-ish architectural suggestion

src/miral/enable_mousekeys.cpp
tarek-y-ismail: That already exists in the form of `AccessibilityManager::toggle_mousekeys`. You could say that `EnableMouseKeys` is just a demo/default. If shell authors want to implement something different, they've got a blueprint

include/miral/miral/enable_mousekeys.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef MIRAL_ENABLE_MOUSEKEYS_H
+#define MIRAL_ENABLE_MOUSEKEYS_H
+
+#include <memory>
+
+namespace mir { class Server; }
+
+namespace miral
+{
+class EnableMouseKeys
```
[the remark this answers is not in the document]
tarek-y-ismail: Added a brief description. Tell me if it's _too_ brief.

src/miral/enable_mousekeys.cpp
tarek-y-ismail: Ok, I'm a convinced. The easiest way forward that I can see is to expose `toggle_mousekeys` via `EnableMouseKeys`. That way, if a shell author/user wants to toggle on a key combination, a config file (like mir_demo_sever), or whatever else, they can. What do you think?

mattkae: That sounds reasonable enough to me!

src/server/shell/mouse_keys_transformer.h
```diff
@@ -0,0 +1,97 @@
...
+class Option;
+}
+namespace time
+{
+class Alarm;
+}
+namespace input
+{
+
+struct MouseKeysTransformer: public mir::input::InputEventTransformer::Transformer
```
mattkae, line 34: Why is this in its own file here? Could we not put this in an anonymous namespace in the `accessibility_manager.cpp`?

src/server/shell/accessibility_manager.cpp
```diff
@@ -54,72 +55,29 @@ void mir::shell::AccessibilityManager::notify_helpers() const {
...
-                mir::events::ScrollAxisV{}));
-
-            return true;
-        }
-
-        return false;
+        transformer = std::make_shared<mir::input::MouseKeysTransformer>(main_loop, options);
+        event_transformer->append(transformer);
+    } else {
+        transformer.reset();
```
mattkae, line 68: Should we remove the transformer from the event_transformer here?

src/server/shell/mouse_keys_transformer.h
mattkae: This should also be a class in my opinion :)

src/server/shell/mouse_keys_transformer.h
```diff
@@ -0,0 +1,97 @@
...
+{
+    using Dispatcher = mir::input::InputEventTransformer::EventDispatcher;
+
+    MouseKeysTransformer(
+        std::shared_ptr<mir::MainLoop> const& main_loop, std::shared_ptr<mir::options::Option> const& options);
+    bool transform_input_event(
+        mir::input::InputEventTransformer::EventDispatcher const& dispatcher,
+        mir::input::EventBuilder* builder,
+        MirEvent const& event) override;
+    bool handle_motion(
```
mattkae, line 44: Pretty much all of these methods should be private to this class.

I would love some unit tests here too!

mattkae (changes requested):
A few initial big items :)

src/server/shell/mouse_keys_transformer.h
tarek-y-ismail: Yeah that's possible. Just thought it outgrew `accessibility_manager.cpp`. Considering we're going to add a few more transformers, I think keeping each in its own file is alright?

+1 on the class suggestion, just a leftover from when I was prototyping (I really should slap a TODO comment on these thingss...)

src/server/shell/accessibility_manager.cpp
tarek-y-ismail: My original thinking was that since the pointer is freed, it'll be erased in the next `handle` call. But since you mentioned symmetry in an earlier comment...

src/server/shell/mouse_keys_transformer.h
tarek-y-ismail: Hmmm, thinking about the namespace it's in. Should it be moved to `src/server/input`, or should I change the namespace to `shell` instead of `input`? I'm leaning towards the latter but I'm open to suggestions.

mattkae: Yeah its own file is okay for readability if we will add many transformers. I think it is personally fine where it is, since it's only used by the AccessibilityManager after all

src/server/shell/mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,381 @@
...
+            motion_event_generator.reset();
+        }
+
+        return true;
+    case mir_keyboard_action_down:
+        {
+            if (!motion_event_generator)
+            {
+                auto const shared_weak_alarm = std::make_shared<std::weak_ptr<mir::time::Alarm>>();
+                motion_event_generator = main_loop->create_alarm(
```
mattkae, line 133: This whole bit seems weird. Could you comment on what you're trying to work around here?

src/server/shell/mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,381 @@
...
+    {
+        if (action == mir_keyboard_action_down)
+            buttons_down |= button;
+        else if (action == mir_keyboard_action_up)
+            buttons_down &= ~(button);
+    };
+
+    switch (mir_keyboard_event_keysym(kev))
+    {
+    case XKB_KEY_KP_2:
```
mattkae, line 103: I think that all of these keys should be configurable.

I also think that we need some documentation to describe how to use these accessibility features, since I would have no clue right now.

The former might be reasonable in this PR. The latter could come in a follow up PR

mattkae (changes requested):
Some more comments, but looking good :saluting_face:

src/server/shell/mouse_keys_transformer.h
```diff
@@ -0,0 +1,104 @@
...
+        AccelerationParameters params;
+
+        AccelerationCurve(AccelerationParameters const& params);
+
+        double evaluate(double t) const;
+    };
+
+    std::shared_ptr<mir::MainLoop> const main_loop;
+
+    std::shared_ptr<mir::time::Alarm> motion_event_generator; // shared_ptr so we can get a weak ptr to it
```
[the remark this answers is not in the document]
tarek-y-ismail: Why remove the comment? It helps explain why we're using a `shared_ptr` as opposed to a `unique_ptr`

src/server/shell/mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+                  if (max_speed.dy < geometry::DeltaY{0})
+                  {
+                      mir::log_warning(
+                          "%s set to a value less than zero, clamping to zero", mir::options::mouse_keys_max_speed_y);
+                      max_speed.dy = std::max(max_speed.dy, geometry::DeltaY{0.0});
+                  }
+
+                  return max_speed;
+              }()}
```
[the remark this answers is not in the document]
tarek-y-ismail: I've got a slightly different idea, commit incoming....

src/server/shell/mouse_keys_transformer.cpp
tarek-y-ismail: Well, I went down the rabbit hole of adding keymap customization. Take a look at it and tell me your opinion. The only thing I'm not quite comfortable with is having to expose a bunch of stuff in `include/server/mir/input/mousekeys_common.h`, but I think duplicating the classes in miral would be too much.

I intent to implement acceleration curve customization in a similar fashion

include/miral/miral/mousekeys_config.h
```diff
@@ -0,0 +1,58 @@
...
+    void operator()(mir::Server& server) const;
+
+    /// Toggles mousekeys on or off depending on the passed parameter.
+    /// \note Can only be called after the server is initialized.
+    void toggle_mousekeys(bool enabled) const;
+
+    /// Changes the keymap for various mousekeys actions
+    /// \note Can only be called after the server is initialized.
+    /// \note The keymap _must_ cover all actions found in [MouseKeysActions].
+    void update_keymap(mir::input::MouseKeysKeymap const& new_keymap) const;
```
mattkae, line 50: I see what you tried to do with this, and I think it is _almost_ the right idea. Typically in the Mir codebase, we would implement something like this like so:

```c++
struct MouseKeysSpecification
{
  std::optional<XkbKeyAction> move_left_key;
  std::optional<XkbKeyAction> move_right_key;
  // etc...
};
```

Then we only update the ones that the user has set.

mattkae: see: `miral::WindowSpecification` as inspiration of this pattern in action

mattkae: I would also keep this `MouseKeysKeymap` struct in this file instead of another one. It is tightly coupled with this idea.

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -48,6 +61,9 @@ class AccessibilityManager
 
     void notify_helpers() const;
 
+    void toggle_mousekeys(bool on);
+    void update_mousekeys_keymap(input::MouseKeysKeymap const& new_keymap) const;
```
mattkae, line 65: I think it would make more sense to have a method per key here. It allows us to more easily deprecate methods on particular keys by marking them as "deprecated. Also it makes adding new options really straightforward.

src/server/shell/mouse_keys_transformer.cpp
mattkae: Understood. My recommendation would be to include a comment and a link to where you got it from.

Although encapsulating the idea and providing it somehow is not a bad idea too :)

src/server/shell/mouse_keys_transformer.h
```diff
@@ -0,0 +1,104 @@
...
+    bool handle_change_pointer_button(
+        MirKeyboardEvent const* kev, Dispatcher const& dispatcher, mir::input::EventBuilder* const builder);
+    bool handle_double_click(
+        MirKeyboardEvent const* kev, Dispatcher const& dispatcher, mir::input::EventBuilder* const builder);
+    bool handle_drag_start(
+        MirKeyboardEvent const* kev, Dispatcher const& dispatcher, mir::input::EventBuilder* const builder);
+    bool handle_drag_end(
+        MirKeyboardEvent const* kev, Dispatcher const& dispatcher, mir::input::EventBuilder* const builder);
+
+    enum DirectionalButtons
```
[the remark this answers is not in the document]
mattkae: I would prefix all of them then, i.e.:

```
mouse_key_transformer_direction_button_none
mouse_key_transformer_direction_button_up
etc
```

src/server/shell/mouse_keys_transformer.h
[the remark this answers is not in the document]
mattkae: Can the comment be more descriptive then? Of course you can get a weak_ptr from a shared_ptr :smile: Why is this case special?

mattkae (changes requested):
I have another idea about how the customization might work :brain:

src/server/shell/mouse_keys_transformer.h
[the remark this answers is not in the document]
tarek-y-ismail: Small question, doesn't `MouseKeysTransformer` restrict the scope of the enum to itself? Just asking since the `mouse_key_transformer` prefix might not be needed.

AlanGriffiths:
> * Run mir with `--enable-mouse-keys=1` or toggle it at any time by using ctrl + shift + numlock

This is misleading: Only `miral-shell` provides this options

include/miral/miral/mousekeys_config.h
AlanGriffiths: The miral "abstraction layer" should be usable without breaking the abstraction and using mirserver.

It would be, in principle, possible to move `MouseKeysKeymap` to mircore.

Matt's suggestion would be problematic for a stable API as the structure would change layout or size with the addition or removal of members. (`WindowSpecification` is a better example)

Alternatively:

```
class MouseKeys
{
...
public:

    enum class Action { ... };
    void set_action<XkbSymkey key, std::optional<Action> action);
    std::optional<Action> get_action<XkbSymkey key);
};
```

include/platform/mir/options/configuration.h
```diff
@@ -43,6 +43,11 @@ extern char const* const debug_opt;
 extern char const* const composite_delay_opt;
 extern char const* const enable_key_repeat_opt;
 extern char const* const enable_mouse_keys_opt;
+extern char const* const mouse_keys_acceleration_constant_factor;
+extern char const* const mouse_keys_acceleration_linear_factor;
+extern char const* const mouse_keys_acceleration_quadratic_factor;
+extern char const* const mouse_keys_max_speed_x;
+extern char const* const mouse_keys_max_speed_y;
```
AlanGriffiths, line 50: I don't think there's any need for these in mirplatform. 

If you look at, for example, `miral::CursorTheme::operator()()` in `cursor_theme.cpp` the associated options are defined and used in one place.

These are only used in `accessibility_manager.cpp` so can they just be there?

src/include/server/mir/shell/accessibility_manager.h
AlanGriffiths: I disagree: changing a set of mappings should be "atomic" - which is easier with a single call

src/server/input/input_event_transformer.cpp
```diff
@@ -24,8 +24,6 @@
 #include "mir/log.h"
 
 #include <algorithm>
-#include <boost/throw_exception.hpp>
-#include <cassert>
```
AlanGriffiths, line 28: Probably a good change, but unrelated

AlanGriffiths (changes requested):

include/miral/miral/mousekeys_config.h
mattkae: Agreed^

include/platform/mir/options/configuration.h
tarek-y-ismail: I'm a bit lost on this. `accessibility_manager.cpp` doesn't have a way to add extra configuration options as far as I can tell. I could move all configuration to `MouseKeysConfig`, which seems like the more sensible option. That would require us to expose methods to set each of these variables. What do you think, is it worth the work?

AlanGriffiths: > I could move all configuration to `MouseKeysConfig`, which seems like the more sensible option

That seems sensible

> That would require us to expose methods to set each of these variables

By "expose" you mean as part of the `AccessibilityManager` interface (or a `MouseAccessibility` interface it extends)? These methods makes sense as we would want to be able to change these settings at runtime.

src/platform/options/default_configuration.cpp
```diff
@@ -181,8 +180,6 @@ mo::DefaultConfiguration::DefaultConfiguration(
             "Cursor (mouse pointer) to use [{auto,null,software}]")
         (enable_key_repeat_opt, po::value<bool>()->default_value(true),
              "Enable server generated key repeat")
-        (enable_mouse_keys_opt, po::value<bool>()->default_value(false),
-             "Enable mousekeys (controlling the mouse with the numpad)")
```
tarek-y-ismail, line 185: Not very sure about this and related changes. Do we want mousekeys to be enabled regardless of `MouseKeysConfig` being used, or do we want to tie the two together? It feels "cleaner" to have everything in one place.

tests/unit-tests/shell/test_mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,375 @@
...
+                    finished.raise();
+                    break;
+                }
+            });
+
+    input_event_transformer.handle(*up_event(XKB_KEY_KP_0));
+    input_event_transformer.handle(*up_event(XKB_KEY_KP_Decimal));
+
+    finished.wait_for(std::chrono::milliseconds(10));
+}
```
tarek-y-ismail, line 375: Should be updated to handle the new acceleration curve, max speed, and keymap configuration code.

tarek-y-ismail (commented):
Self review

include/core/mir/input/mousekeys_common.h
```diff
@@ -0,0 +1,64 @@
...
+#include <memory>
+#include <optional>
+#include <functional>
+
+
+namespace mir
+{
+namespace input
+{
+using XkbSymkey = unsigned int;
```
mattkae, line 30: This is not feedback, but I didn't know this bit of trivia: https://stackoverflow.com/questions/10747810/what-is-the-difference-between-typedef-and-using

This is basically equivalent to `typedef`, except it can't be used in a template. Cool stuff

examples/miral-shell/shell_main.cpp
```diff
@@ -142,6 +143,37 @@ int main(int argc, char const* argv[])
             return FocusStealing::allow;
     };
 
+    auto const initial_mousekeys_state = false;
+    miral::MouseKeysConfig mousekeys_config{initial_mousekeys_state};
+
+    auto toggle_mousekeys_filter = [mousekeys_on = initial_mousekeys_state, &mousekeys_config](MirEvent const* event) mutable {
+
```
mattkae, line 150:
```suggestion
```

examples/miral-shell/shell_main.cpp
```diff
@@ -142,6 +143,37 @@ int main(int argc, char const* argv[])
...
+        if(mir_event_get_type(event) != mir_event_type_input)
+            return false;
+
+        auto const* input_event = mir_event_get_input_event(event);
+        if(mir_input_event_get_type(input_event) != mir_input_event_type_key)
+            return false;
+
+
```
mattkae, line 158:
```suggestion
```

src/miral/mousekeys_config.cpp
```diff
@@ -0,0 +1,123 @@
...
+        mir::log_error("AccessibilityManager not initialized. Will not set acceleration factors.");
+
+    self->accessibility_manager.lock()->set_acceleration_factors(constant, linear, quadratic);
+}
+
+
+void miral::MouseKeysConfig::set_max_speed(double x_axis, double y_axis) const
+{
+    if (self->accessibility_manager.expired())
+        mir::log_error("AccessibilityManager not initialized. Will not set max speed.");
```
mattkae, line 78:
```suggestion
    if (self->accessibility_manager.expired())
    {
        mir::log_error("AccessibilityManager not initialized. Will not set max speed.");
        return;
    }
```

src/miral/mousekeys_config.cpp
```diff
@@ -0,0 +1,123 @@
...
+    if (self->accessibility_manager.expired())
+        mir::log_error("AccessibilityManager not initialized. Will not update keymap.");
+
+    self->accessibility_manager.lock()->set_mousekeys_keymap(new_keymap);
+}
+
+void miral::MouseKeysConfig::set_acceleration_factors(double constant, double linear, double quadratic) const
+{
+    if (self->accessibility_manager.expired())
+        mir::log_error("AccessibilityManager not initialized. Will not set acceleration factors.");
```
mattkae, line 69:
```suggestion
    if (self->accessibility_manager.expired())
    {
        mir::log_error("AccessibilityManager not initialized. Will not set acceleration factors.");
        return;
    }
```

src/miral/mousekeys_config.cpp
```diff
@@ -0,0 +1,123 @@
...
+    if(self->accessibility_manager.expired())
+        mir::log_error("AccessibilityManager not initialized. Will not toggle mousekeys.");
+
+    self->accessibility_manager.lock()->set_mousekeys_enabled(enabled);
+}
+
+void miral::MouseKeysConfig::set_keymap(mir::input::MouseKeysKeymap const& new_keymap) const
+{
+    if (self->accessibility_manager.expired())
+        mir::log_error("AccessibilityManager not initialized. Will not update keymap.");
```
mattkae, line 61:
```suggestion
    if (self->accessibility_manager.expired())
    {
        mir::log_error("AccessibilityManager not initialized. Will not update keymap.");
        return;
    }
```

mattkae (changes requested):
This is getting to a really great place! To land this, we're going to need more tests though, particularly for:

- MouseKeysKeymap
- MouseKeysConfig
- AccessibilityManager

Pretty much all new code should be tested nowadays, unless we've got a good reason not to.

The general architecture should no longer change now though :tada:

include/core/mir/input/mousekeys_common.h
tarek-y-ismail: Maybe I got confused while reading, you mean that typedefs can't be used in templates, not "symbols" defined with `using`, right?

mattkae: Yup sorry I wrote that wrong.

`typedef`s cannot be used in templates. `using`s can. https://stackoverflow.com/a/32959200

mattkae:
> > we're going to need more tests though
> 
> I'm down for this, but are you sure the PR is not going to be too big to comfortably review? It's already quite big right now

All of the bits are straightforward in my head now. Also, now that the code has settled, it's a good time to write tests so that we know it all works before it lands.

src/miral/mousekeys_config.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+miral::MouseKeysConfig::MouseKeysConfig(bool enabled_by_default) :
+    self{std::make_shared<miral::MouseKeysConfig::Self>(enabled_by_default)}
+{
+}
+
+void miral::MouseKeysConfig::set_mousekeys_enabled(bool enabled) const
+{
+    if(self->accessibility_manager.expired())
+        mir::log_error("AccessibilityManager not initialized. Will not toggle mousekeys.");
```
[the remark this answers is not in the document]
tarek-y-ismail: :facepalm:

src/server/symbols.map
```diff
@@ -1458,3 +1458,17 @@ global:
...
+    mir::Server::override_the_accessibility_manager*;
+    mir::shell::AccessibilityManager::?AccessibilityManager*;
+    mir::shell::AccessibilityManager::set_acceleration_factors*;
+    mir::shell::AccessibilityManager::set_max_speed*;
+    mir::shell::AccessibilityManager::set_mousekeys_enabled*;
+    mir::shell::AccessibilityManager::set_mousekeys_keymap*;
+    non-virtual?thunk?to?mir::shell::AccessibilityManager::?AccessibilityManager*;
+    vtable?for?mir::shell::AccessibilityManager;
+  };
+} MIR_SERVER_INTERNAL_2.20;
```
AlanGriffiths, line 1473: When breaking ABI everything ends up in the single new stanza

AlanGriffiths (changes requested):
We probably need more tests. E.g. for setting the same key several times

src/server/shell/mouse_keys_transformer.h
```diff
@@ -0,0 +1,140 @@
...
+
+    std::unique_ptr<mir::time::Alarm> click_event_generator;
+    std::unique_ptr<mir::time::Alarm> double_click_event_generator;
+
+    mir::geometry::DisplacementF motion_direction;
+    MirPointerButtons current_button{mir_pointer_button_primary};
+
+    uint32_t buttons_down{directional_buttons_none};
+
+    bool is_dragging{false};
```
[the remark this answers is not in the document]
tarek-y-ismail: What do you suggest we do?

AlanGriffiths:
```
dh_makeshlibs -Nlibmiral7 -V -Nmir-test-tools
dpkg-gensymbols: warning: failed to parse line in debian/libmircore2.symbols:  (c++)"mir::input::MouseKeysKeymap::MouseKeysKeymap()@MIR_CORE_2.10"
dpkg-gensymbols: warning: failed to parse line in debian/libmircore2.symbols:  (c++)"mir::input::MouseKeysKeymap::MouseKeysKeymap()@MIR_CORE_2.10"
dpkg-gensymbols: error: some new symbols appeared in the symbols file: see diff output below
dpkg-gensymbols: error: some symbols or patterns disappeared in the symbols file: see diff output below
dpkg-gensymbols: warning: debian/libmircore2/DEBIAN/symbols doesn't match completely debian/libmircore2.symbols
--- debian/libmircore2.symbols (libmircore2_2.20.0_arm64)
+++ dpkg-gensymbolsPgbqzv	2025-03-21 15:01:03.631248714 +0000
@@ -1,6 +1,13 @@
 libmircore.so.2 libmircore2 #MINVER#
- MIR_CORE_2.10@MIR_CORE_2.10 2.9.0
+#MISSING: 2.20.0# MIR_CORE_2.10@MIR_CORE_2.10 2.9.0
  MIR_CORE_2.9@MIR_CORE_2.9 2.8.0
+ _ZN3mir5input15MouseKeysKeymap10set_actionEjSt8optionalINS1_6ActionEE@MIR_CORE_2.9 2.20.0
+ _ZN3mir5input15MouseKeysKeymapC1ESt16initializer_listISt4pairIjNS1_6ActionEEE@MIR_CORE_2.9 2.20.0
+ _ZN3mir5input15MouseKeysKeymapC1Ev@MIR_CORE_2.9 2.20.0
+ _ZN3mir5input15MouseKeysKeymapC2ESt16initializer_listISt4pairIjNS1_6ActionEEE@MIR_CORE_2.9 2.20.0
+ _ZN3mir5input15MouseKeysKeymapC2Ev@MIR_CORE_2.9 2.20.0
+ _ZNK3mir5input15MouseKeysKeymap10get_actionEj@MIR_CORE_2.9 2.20.0
+ _ZNK3mir5input15MouseKeysKeymap24for_each_key_action_pairEOSt8functionIFvjNS1_6ActionEEE@MIR_CORE_2.9 2.20.0
  (c++|arch-bits=32)"mir::AnonymousShmFile::AnonymousShmFile(unsigned int)@MIR_CORE_2.9" 2.8.0
  (c++|arch-bits=64)"mir::AnonymousShmFile::AnonymousShmFile(unsigned long)@MIR_CORE_2.9" 2.8.0
  (c++)"mir::AnonymousShmFile::base_ptr() const@MIR_CORE_2.9" 2.8.0
@@ -28,11 +35,11 @@
  (c++)"mir::geometry::Rectangles::remove(mir::geometry::generic::Rectangle<int> const&)@MIR_CORE_2.9" 2.8.0
  (c++)"mir::geometry::Rectangles::size() const@MIR_CORE_2.9" 2.8.0
  (c++)"mir::geometry::operator<<(std::basic_ostream<char, std::char_traits<char> >&, mir::geometry::Rectangles const&)@MIR_CORE_2.9" 2.8.0
- (c++)"mir::input::MouseKeysKeymap::MouseKeysKeymap()@MIR_CORE_2.10" 2.9.0
- (c++)"mir::input::MouseKeysKeymap::MouseKeysKeymap(std::initializer_list<std::pair<unsigned int, mir::input::MouseKeysKeymap::Action> >)@MIR_CORE_2.10" 2.9.0
- (c++)"mir::input::MouseKeysKeymap::for_each_key_action_pair(std::function<void (unsigned int, mir::input::MouseKeysKeymap::Action)>&&) const@MIR_CORE_2.10" 2.9.0
- (c++)"mir::input::MouseKeysKeymap::get_action(unsigned int) const@MIR_CORE_2.10" 2.9.0
- (c++)"mir::input::MouseKeysKeymap::set_action(unsigned int, std::optional<mir::input::MouseKeysKeymap::Action>)@MIR_CORE_2.10" 2.9.0
+#MISSING: 2.20.0# (c++)"mir::input::MouseKeysKeymap::MouseKeysKeymap()@MIR_CORE_2.10" 2.9.0
+#MISSING: 2.20.0# (c++)"mir::input::MouseKeysKeymap::MouseKeysKeymap(std::initializer_list<std::pair<unsigned int, mir::input::MouseKeysKeymap::Action> >)@MIR_CORE_2.10" 2.9.0
+#MISSING: 2.20.0# (c++)"mir::input::MouseKeysKeymap::for_each_key_action_pair(std::function<void (unsigned int, mir::input::MouseKeysKeymap::Action)>&&) const@MIR_CORE_2.10" 2.9.0
+#MISSING: 2.20.0# (c++)"mir::input::MouseKeysKeymap::get_action(unsigned int) const@MIR_CORE_2.10" 2.9.0
+#MISSING: 2.20.0# (c++)"mir::input::MouseKeysKeymap::set_action(unsigned int, std::optional<mir::input::MouseKeysKeymap::Action>)@MIR_CORE_2.10" 2.9.0
  (c++)"mir::mir_depth_layer_get_index(MirDepthLayer)@MIR_CORE_2.9" 2.8.0
  (c++)"typeinfo for mir::AnonymousShmFile@MIR_CORE_2.9" 2.8.0
  (c++)"typeinfo for mir::ShmFile@MIR_CORE_2.9" 2.8.0
dh_makeshlibs: error: failing due to earlier errors
```

Looks like

1. src/core/symbols.map is messed up: the new symbols should be in a `MIR_CORE_2.21` stanza, not `MIR_CORE_2.9` 
2. debian/libmircore2.symbols is messed up, the new symbols should be in a `MIR_CORE_2.21@MIR_CORE_2.21 2.21.0` stanza, not `MIR_CORE_2.10@MIR_CORE_2.10 2.9.0`

src/server/shell/mouse_keys_transformer.h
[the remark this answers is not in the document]
tarek-y-ismail: The first thing that comes into mind is to wrap all state in a `Synchronised<T>`, is this what you had in mind?

src/server/shell/accessibility_manager.cpp
```diff
@@ -16,110 +16,4 @@
...
-    std::shared_ptr<input::InputEventTransformer> const& event_transformer) :
-    enable_key_repeat{options->get<bool>(options::enable_key_repeat_opt)},
-    enable_mouse_keys{options->get<bool>(options::enable_mouse_keys_opt)},
-    event_transformer{event_transformer},
-    transformer{std::make_shared<MouseKeysTransformer>()}
-{
-    if (enable_mouse_keys)
-        event_transformer->append(transformer);
-}
+mir::shell::AccessibilityManager::~AccessibilityManager() = default;
```
mattkae, line 19: Why do you need this file anymore? Can't this `=default` be put in the header file?

src/server/shell/basic_accessibility_manager.h
```diff
@@ -0,0 +1,106 @@
...
+        std::shared_ptr<input::InputEventTransformer> const& event_transformer,
+        std::shared_ptr<time::Clock> const& clock,
+        bool enable_key_repeat);
+
+    BasicAccessibilityManager(
+        std::shared_ptr<MainLoop> const& main_loop,
+        std::shared_ptr<input::InputEventTransformer> const& event_transformer,
+        std::shared_ptr<time::Clock> const& clock,
+        bool enable_key_repeat,
+        MouseKeysTransformerBuilder&& builder);
```
mattkae, line 65: I don't think that you need the former constructor here. That feels like a sneaky way to provide some default behavior when you intend for the `MouseKeysTransformer` to be a swappable implementation detail.

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -17,56 +17,36 @@
...
-    bool const enable_key_repeat;
-    bool const enable_mouse_keys;
+    virtual void notify_helpers() const = 0;
 
-    std::shared_ptr<mir::input::InputEventTransformer> const event_transformer;
-    std::shared_ptr<mir::input::InputEventTransformer::Transformer> const transformer;
+    virtual void set_mousekeys_enabled(bool on) = 0;
+    virtual void set_mousekeys_keymap(input::MouseKeysKeymap const& new_keymap) = 0;
+    virtual void set_acceleration_factors(double constant, double linear, double quadratic) = 0;
+    virtual void set_max_speed(double x_axis, double y_axis) = 0;
```
mattkae, line 49:
```suggestion
    virtual void mousekeys_enabled(bool on) = 0;
    virtual void mousekeys_keymap(input::MouseKeysKeymap const& new_keymap) = 0;
    virtual void acceleration_factors(double constant, double linear, double quadratic) = 0;
    virtual void max_speed(double x_axis, double y_axis) = 0;
```

mattkae: Just to make it match how you're now doing setters above without the verb

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -0,0 +1,149 @@
...
+{
+}
+
+void mir::shell::BasicAccessibilityManager::set_acceleration_factors(double constant, double linear, double quadratic)
+{
+    acceleration_constant = constant;
+    acceleration_linear = linear;
+    acceleration_quadratic = quadratic;
+    if(transformer)
+        transformer->set_acceleration_factors(constant, linear, quadratic);
```
mattkae, line 126: If we did not recreate the `transformer` every time and simply added/removed it as we needed it, wouldn't we not need to keep copies of its properties?

I think that would be simpler :) That way, you would pass the transformer object directly and you wouldn't have to store a builder function.

mattkae (changes requested):
I like this direction :+1: 

I have a few architectural things first

CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
 set(MIR_VERSION_MAJOR 2)
-set(MIR_VERSION_MINOR 20)
+set(MIR_VERSION_MINOR 22)
```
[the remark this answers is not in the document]
tarek-y-ismail: Mixed 2.20.1 with 2.21.0...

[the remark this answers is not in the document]
tarek-y-ismail: Or more likely, got confused by this comment:
https://github.com/canonical/mir/blob/362ad1d51c1c0abf0ba4c35854aee5717f4be914/src/server/CMakeLists.txt#L171-L172

I had already bumped mirserver to 2.21 before this, so I thought I'd need to bump again. @AlanGriffiths I only need to go up to 2.21, right?

src/server/shell/accessibility_manager.cpp
tarek-y-ismail: I remember having linker issues previously with `dtor = default;` in header files

src/server/shell/basic_accessibility_manager.h
tarek-y-ismail: I only needed a swappable `MouseKeysTranformer` implementation for testing, most users of this class will just use the first constructor with the default behavior.

src/include/server/mir/shell/accessibility_manager.h
tarek-y-ismail: Do we want to do the same in `MouseKeysTransformer`?

src/server/shell/basic_accessibility_manager.cpp
tarek-y-ismail: This was a bad optimization on my part :) Thanks for pointing that out

src/include/server/mir/shell/accessibility_manager.h
mattkae: I suppose it would be best for consistency's sake.

tests/unit-tests/shell/test_mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,496 @@
...
+            mir_keyboard_action_up,
+            button,
+            0,
+            mir_input_event_modifier_none);
+    }
+};
+
+TEST_F(TestMouseKeysTransformer, single_keyboard_event_to_single_pointer_motion_event)
+{
+    for (auto const button : {XKB_KEY_KP_2, XKB_KEY_KP_4, XKB_KEY_KP_6, XKB_KEY_KP_8})
```
mattkae, line 125: Nit: for this test and the similar ones, I personally prefer using a parametrized test. Having each one be an individual test ensures that previous state doesn't accidentally step on the toes of another test

tests/unit-tests/shell/test_mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,496 @@
...
+    {
+        auto const timeout = std::chrono::milliseconds(20);
+        auto const repeat_delay = std::chrono::milliseconds(2);
+        auto const num_invocations = timeout / repeat_delay;
+
+        EXPECT_CALL(mock_seat, dispatch_event(_))
+            .Times(AtLeast(num_invocations / 2)) // Account for various timing inconsistencies
```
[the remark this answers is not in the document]
tarek-y-ismail: The problem here is that our main loop doesn't seem as accurate as this needs to be. I just tried your suggestion, and I get around 4600-4800 invocations before timing out. I'm sure the number of invocations is likely to be lower on CI as well.

tests/unit-tests/shell/test_mouse_keys_transformer.cpp
tarek-y-ismail: I thought of doing this while I was writing the last one or two tests, but thought my usecase didn't really justify it. I'll try it and see how it goes.

tests/unit-tests/shell/test_mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,485 @@
...
+TEST_F(TestMouseKeysTransformer, clicks_dispatch_pointer_down_and_up_events)
+{
+    for (auto switching_button : {XKB_KEY_KP_Divide, XKB_KEY_KP_Multiply, XKB_KEY_KP_Subtract})
+    {
+        mt::Signal finished;
+        auto state = State::waiting_for_up;
+
+        EXPECT_CALL(mock_seat, dispatch_event(_))
+            .Times(3) // up (switching), down, up
+            .WillRepeatedly(
```
mattkae, line 259: Instead of having `state`, you should be able to `EXPECT_CALL(mock_seat, dispatch_event(_)).Times(1).WillOnce(...)` three times in a row using `InSequence`. At least I think that should work :smile: That would simplify what you believe should happen here

tests/unit-tests/shell/test_mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,485 @@
...
+        }
+    }
+}
+
+TEST_F(TestMouseKeysTransformer, double_click_dispatch_four_events)
+{
+    auto state = State::waiting_for_down;
+    mt::Signal finished;
+
+    EXPECT_CALL(mock_seat, dispatch_event(_))
```
mattkae, line 316: Also a good candidate for `InSequence`, as well as others that use `State` later on

mattkae (changes requested):
A few things! Looking good :saluting_face:

tests/unit-tests/shell/test_basic_accessibility_manager.cpp
```diff
@@ -0,0 +1,185 @@
...
+namespace mir::input
+{
+    bool operator==(MouseKeysKeymap const& left, MouseKeysKeymap const& right)
+    {
+        std::vector<std::pair<XkbSymkey, MouseKeysKeymap::Action>> leftPairs, rightPairs;
+        left.for_each_key_action_pair([&leftPairs](auto key, auto action){ leftPairs.emplace_back(key, action); });
+        right.for_each_key_action_pair([&rightPairs](auto key, auto action){ rightPairs.emplace_back(key, action); });
+
+        return leftPairs == rightPairs;
+    }
```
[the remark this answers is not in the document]
tarek-y-ismail: Will probably be nuked as it pollutes the `mir::input` namespace. See https://github.com/canonical/mir/pull/3830#discussion_r2006182961

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -0,0 +1,95 @@
...
+        helper->repeat_info_changed(repeat_rate(), repeat_delay());
+}
+
+void mir::shell::BasicAccessibilityManager::mousekeys_enabled(bool on)
+{
+    if (on)
+    {
+        if (transformer)
+            return;
+
```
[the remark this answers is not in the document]
tarek-y-ismail: It's possible that shell authors call `mousekeys_enabled(true)` twice in a row.

mattkae (changes requested):
Some more minor things, I will look into the MainLoop issue

src/server/shell/basic_accessibility_manager.cpp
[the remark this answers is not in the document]
tarek-y-ismail: Ooooooooh I get it. My brain still associates `transformer` being null/not null with it being added to the input event transformer. 

Since there's no way to query the input event transformer for whether a transformer is inserted in it, nor do users of it need that. Do you see it as a `BOOST_THROW_EXCPETION`  worthy error inside `append`?

tests/unit-tests/shell/test_mouse_keys_transformer.cpp
tarek-y-ismail: From manual testing locally, no. But maybe that's only my machine and why the tests sometimes fail for no reason. I err on the side of caution and add them to tests with multiple `Will*` clauses.

src/server/shell/mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,405 @@
...
+                        // all buttons (if for example you press two
+                        // opposite directions), reset the time for
+                        // acceleration.
+                        if (motion_direction.length_squared() == 0)
+                            motion_start_time = motion_step_time;
+
+                        dispatcher(builder->pointer_event(
+                            std::nullopt,
+                            mir_pointer_action_motion,
+                            is_dragging ? current_button : 0,
```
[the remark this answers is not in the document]
tarek-y-ismail: This would cause you to drag all the time though?

src/server/shell/mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+    press_current_cursor_button(dispatcher, builder);
+
+    click_event_generator = main_loop->create_alarm(
+        [dispatcher, this, builder]
+        {
+            release_current_cursor_button(dispatcher, builder);
+        });
+
+    click_event_generator->reschedule_in(std::chrono::milliseconds(50));
```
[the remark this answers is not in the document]
tarek-y-ismail: I haven't tried setting it to 0. This just looked like a good value at first and it worked

src/server/shell/mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,405 @@
...
+        break;
+    case button_secondary:
+        current_button = mir_pointer_button_secondary;
+        break;
+    default:
+        std::unreachable();
+    }
+
+    if (current_button)
+        release_current_cursor_button(dispatcher, builder);
```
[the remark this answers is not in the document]
tarek-y-ismail: Don't think users can press and release buttons that fast :)

src/server/shell/mouse_keys_transformer.cpp
```diff
@@ -0,0 +1,405 @@
...
+    release_current_cursor_button(dispatcher, builder);
+
+    double_click_event_generator = main_loop->create_alarm(
+        [dispatcher, this, builder]
+        {
+            press_current_cursor_button(dispatcher, builder);
+            release_current_cursor_button(dispatcher, builder);
+        });
+
+    double_click_event_generator->reschedule_in(std::chrono::milliseconds(100));
```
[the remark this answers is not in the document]
tarek-y-ismail: Nope. Chose it arbitrarily to space out the first and second clicks.

src/server/shell/basic_accessibility_manager.cpp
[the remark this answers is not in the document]
mattkae: Nope! I think it is guaranteed to be initialized by the constructor, so we will never have to check that case

src/server/shell/mouse_keys_transformer.cpp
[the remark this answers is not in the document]
mattkae: If the computer can let it happen, then we should at least code against it :) I could probably write a test that causes this to happen

src/server/shell/mouse_keys_transformer.cpp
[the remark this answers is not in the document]
mattkae: And why do we need to space them out?

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -0,0 +1,88 @@
...
+void mir::shell::BasicAccessibilityManager::notify_helpers() const {
+    for (auto const& helper: keyboard_helpers)
+        helper->repeat_info_changed(repeat_rate(), repeat_delay());
+}
+
+void mir::shell::BasicAccessibilityManager::mousekeys_enabled(bool on)
+{
+    if (on)
+        event_transformer->append(transformer);
```
tarek-y-ismail, line 60: Juuuust in case. I modified this to check duplicates similar to `remove`. I don't print an error message at the moment, since it's likely that this is a shell author error so users can't do much about it. But I don't think throwing an exception is required in this case.

mattkae: Agreed :+1:

mattkae (changes requested):
We appear to have a deadlock! Here is what is happening:

- `BasicMouseKeysTransformer::transform_input_event` takes the `state_mutex`. It then tries to reset `motion_event_generator`. This causes `TimerGSource::disable` to take its internal mutex. Unfortunately it (sometimes) gets stuck here.
- Why is it stuck, you ask? There is a timing issue if the `motion_event_generator`'s callback starts triggering while between the time that `BasicMouseKeysTransformer::transform_input_event` and the reset was requested. This is because `TimerGSource::dispatch` takes the mutex, and then it wants to take `state_mutex` in the callback.

Hence, `BasicMouseKeysTransformer::transform_input_event` is stuck waiting on `TimerGSource`'s internal mutex, while `motion_event_generator`'s callback is stuck waiting on `state_mutex`. A deadlock.

mattkae:
> Alright, so that might be the wrong place/time to reset the alarm.
> 
> What if I change the motion event generator to be created once at initialization, and instead of resetting the pointer, I just send some signal (`std::atomic`?) to prevent it from repeating. Whenever its needed again, it can be rescheduled on the input thread.

That would certainly get around it. The other thing is that the alarm itself is already gating its access behind a mutex, so you don't need to hold the `state_mutex` while resetting the alarm.

mattkae (changes requested):
Nevermind :)

mattkae:
I seem to be getting random-ish `mir_keyboard_action_up` events. I wonder why...

mattkae (changes requested):
Found some more potential bugs:

1. Open up gedit
2. Type some things
3. Click `0` to double click, but note that the last selected word isn't selected as you would expect from a double click
4. Similarly, all mouse buttons seem to trigger a right click from me as far as I can tell?

tarek-y-ismail:
> Click `0` to double click, but note that the last selected word isn't selected as you would expect from a double click

`0` is for dragging. `+` is what you press to double click. 

> Similarly, all mouse buttons seem to trigger a right click from me as far as I can tell?

Our implementation is stateful, if you press `-` to choose the right mouse button, then all the following button related actions will use the right mouse button.

src/server/symbols.map
```diff
@@ -545,6 +545,7 @@ global:
     mir::scene::SessionLockObserver::?SessionLockObserver*;
     mir::scene::SessionLockObserver::SessionLockObserver*;
     mir::scene::SessionLockObserver::operator*;
+    mir::scene::Surface::get_input_region*;
```
tarek-y-ismail, line 548: Git blame shows this was added in 2019, why wasn't this in the symbols map already?

src/server/shell/mouse_keys_transformer.h
tarek-y-ismail, line 1: Should both the header and source be moved to `src/server/input`? or the namespace changed to `mir::shell`?

tarek-y-ismail: Namespace should be changed

src/server/shell/mouse_keys_transformer.cpp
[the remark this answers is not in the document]
mattkae: This has been resolved, but it has been _resolved_ ? :thinking:

mattkae (changes requested):
A few final things!

mattkae (approved):
I am approving, but I've got two nits on the doc stuff :smile:

PR #3762, opened 2025-02-19
server: add missing key repeat symbol

Saviq:


tarek-y-ismail (approved):
Whoopsies, thanks for catching this!

PR #3763, opened 2025-02-19
std::wstring_convert is deprecated, but there's no obvious replacement. Yet

AlanGriffiths:
This is "Appendix D" and slated for removal in C++26, hopefully there will be an obvious replacement by then


AlanGriffiths:
This fixes the build errors I see building Mir on Neil's lxqt preview

Saviq (approved):
Sures.

PR #3764, opened 2025-02-19
Workaround for LXQt panel

AlanGriffiths:
Provide a way to hack Mir into ignoring protocol constraints on the anchor rectangle

See https://github.com/lxqt/lxqt-wayland-session/pull/40#issuecomment-2668999542 for more discussion


mattkae (approved):
Not in love with it, but if it fixes the issue, I can accept it :+1:

Issue #3765, opened 2025-02-20
Crash when pcmanfm-qt does DnD


AlanGriffiths:
This is in our data_device.start_drag() implementation: we access the size of the icon, but pcmanfm-qt has not submitted a buffer (so we have no size).

PR #3766, opened 2025-02-20
Do not crash when pcmanfm-qt does DnD 

AlanGriffiths:
I.e.  check we have a buffer size before using it

Fixes: #3765


Saviq (approved):
Yups.

PR #3767, opened 2025-02-21
MirOIL updates for post-new-Platform-API QtMir

RAOF:
The main active ingredient here is passing through the list of `GLRenderingProviders` when invoking the `SetCompositor` hook, necessary for QtMir to allocate `Framebuffer`s to submit to output and to get `gl::Texture`s out of `mg::Buffer`s.

Also some miscellaneous drive-through cleanup, and the addition of `gl::Texture::tex_id()` that significantly simplifies QtMir by exposing something Mir already¹ knows.

¹: With the exception of `ShmBuffer`, but this also prompts a bit of cleanup.


tarek-y-ismail (approved):
Nothing screams "broken" at me, though I suggest waiting for an extra review :)

src/platforms/common/server/shm_buffer.cpp
```diff
@@ -91,13 +91,42 @@ bool mgc::ShmBuffer::supports(MirPixelFormat mir_format)
     return mg::get_gl_pixel_format(mir_format, gl_format, gl_type);
 }
 
+namespace
+{
+auto get_tex_id_on_context(mgc::EGLContextExecutor& egl_executor) -> std::shared_future<GLuint>
+{
+    auto tex_promise = std::make_shared<std::promise<GLuint>>();
+    egl_executor.spawn(
```
mattkae, line 99: Why must this be spawned on the egl executor thread?

tests/platform_test_harness/graphics_platform_test_harness.cpp
```diff
@@ -240,22 +239,22 @@ auto test_display_has_at_least_one_enabled_output(mg::Display& display) -> bool
     return output_count > 0;
 }
 
-auto dump_egl_config(mg::Display& display) -> bool
-{
-    auto& context_source = dynamic_cast<mir::renderer::gl::ContextSource&>(display);
-    auto ctx = context_source.create_gl_context();
+// auto dump_egl_config(mg::Display& display) -> bool
```
mattkae, line 242: Should we uncomment or remove this bit?

tests/platform_test_harness/graphics_platform_test_harness.cpp
```diff
@@ -553,7 +552,7 @@ int main(int argc, char const** argv)
         {
             if (auto display = test_display_construction(*platform, config))
             {
-                success &= dump_egl_config(*display);
+//                success &= dump_egl_config(*display);
```
mattkae, line 555: Should this also be uncommented?

mattkae (changes requested):
A few minor things :+1:

src/platforms/common/server/shm_buffer.cpp
RAOF: Because we don't guarantee that the `ShmBuffer` constructor is called on a thread with an EGL context current, but we need one in order to allocate a GL texture ID. Documented.

mattkae (approved):
Looks good to me now :)

mattkae:
But many things appear to be failing @RAOF

Saviq:
```
 AssertionError: Command `dpkg-gensymbols -e/<<PKGBUILDDIR>>/build-amd64/lib/libmiroil.so.6 -plibmiroil6 -v2.20.0 -O/tmp/temp_miroil_debian_symbols.symbols -c4` failed with return code 1:
  OUT: --- debian/libmiroil6.symbols (libmiroil6_2.20.0_amd64)
       +++ dpkg-gensymbolsoAftu7	2025-02-26 08:08:49.165818540 +0000
       @@ -82,7 +82,7 @@
         (c++)"miroil::SetCompositor::operator()(mir::Server&)@MIROIL_6.0" 2.17.0
         (c++)"miroil::Surface::Surface(std::shared_ptr<mir::scene::Surface>)@MIROIL_6.0" 2.17.0
         (c++)"miroil::Surface::add_observer(std::shared_ptr<miroil::SurfaceObserver> const&)@MIROIL_6.0" 2.17.0
       - (c++)"miroil::Surface::buffers_ready_for_compositor(void const*) const@MIROIL_6.0" 2.17.0
       +#MISSING: 2.20.0# (c++)"miroil::Surface::buffers_ready_for_compositor(void const*) const@MIROIL_6.0" 2.17.0
         (c++)"miroil::Surface::configure(MirWindowAttrib, int)@MIROIL_6.0" 2.17.0
         (c++)"miroil::Surface::generate_renderables(void const*) const@MIROIL_6.0" 2.17.0
         (c++)"miroil::Surface::get_wrapped() const@MIROIL_6.0" 2.17.0
       
  ERR: dpkg-gensymbols: error: some symbols or patterns disappeared in the symbols file: see diff output below
       dpkg-gensymbols: warning: /tmp/temp_miroil_debian_symbols.symbols doesn't match completely debian/libmiroil6.symbols
```

RAOF:
Ah. This:
```
[ FAILED ] ShmBufferTest.texture_is_destroyed_on_thread_with_current_context (0 ms)
```
is a race. But is it a race in our test code, or the real code?

RAOF:
Ah! There's the race - we construct a ShmBuffer in the test constructor:
https://github.com/canonical/mir/blob/23bc32bafc16fa6dad24d5c0b3c939bf131244b5/tests/unit-tests/graphics/test_shm_buffer.cpp#L102
which enqueues a functor that's going to call `glGenTextures`. If the `glGenTextures` expectation: 
https://github.com/canonical/mir/blob/23bc32bafc16fa6dad24d5c0b3c939bf131244b5/tests/unit-tests/graphics/test_shm_buffer.cpp#L319 
is set up before the egl delegate thread hits the call, we'll end up calling `glGenTextures` twice - once for the test fixture's `ShmBuffer`, once for the `ShmBuffer` we're constructing in the test.

debian/control
```diff
@@ -608,7 +608,7 @@ Description: Developer files for the Mir ABI-stable abstraction layer
  Contains header files required for development using the MirAL abstraction
  layer.
 
-Package: libmiroil5
+Package: libmiroil7
```
AlanGriffiths, line 611: What happened to libmiroil6?

AlanGriffiths (changes requested):
`libmiroil7.symbols.armhf` is screwed up: at a minimum, all the versions strings are wrong

include/platform/mir/graphics/texture.h
```diff
@@ -70,6 +72,11 @@ class Texture
      */
     virtual void bind() = 0;
 
+    /**
+     * Get the GL ID of the texture
+     */
+    virtual auto tex_id() const -> GLuint = 0;
+
```
AlanGriffiths, line 79: ABI break

AlanGriffiths (approved):
Unfortunately, this follows the recent tradition of breaking the graphics platform ABI.

As that's also happening in #3735 I think we can ignore it here. But we need to pay attention to ensure both land in 2.20 (@Saviq ping!)

Issue #3771, opened 2025-02-21
Segfault in TokenAuthority


AlanGriffiths:
With Miriway/edge Miriway crashes when double clicking an image in Telegram, (it doesn't happen with /stable

```
#0  0x00007ffff6cadcc2 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::size (this=0x0) at /usr/include/c++/13/bits/basic_string.h:1072
#1  0x00007ffff6d6a877 in std::operator==<char, std::char_traits<char>, std::allocator<char> >
    (__lhs=<error reading variable: Cannot access memory at address 0x8>, __rhs="eb9cc114-9d0a-4fe5-a9c5-eb4274043541") at /usr/include/c++/13/bits/basic_string.h:3714
#2  0x00007ffff70d9da9 in mir::shell::TokenAuthority::Token::operator== (this=0x0, token=...) at /home/alan/CLionProjects/mir/src/server/shell/token_authority.cpp:51
#3  0x00007ffff703c3b1 in operator()<std::shared_ptr<mir::frontend::XdgActivationTokenData> >
    (__closure=0x7fffbfffd670, pending_token=std::shared_ptr<mir::frontend::XdgActivationTokenData> (empty) = {...})
    at /home/alan/CLionProjects/mir/src/server/frontend_wayland/xdg_activation_v1.cpp:263
#4  0x00007ffff703e898 in std::__invoke_impl<bool, mir::frontend::XdgActivationV1::remove_token(const Token&)::<lambda(const auto:65&)>&, std::shared_ptr<mir::frontend::XdgActivationTokenData>&>(std::__invoke_other, struct {...} &) (__f=...) at /usr/include/c++/13/bits/invoke.h:61
#5  0x00007ffff703e594 in std::__invoke<mir::frontend::XdgActivationV1::remove_token(const Token&)::<lambda(const auto:65&)>&, std::shared_ptr<mir::frontend::XdgActivationTokenData>&>(struct {...} &) (__fn=...) at /usr/include/c++/13/bits/invoke.h:96
[...]

PR #3772, opened 2025-02-21
Do not crash in XdgActivationV1

AlanGriffiths:
The segfault was because of null entries in `pending_tokens`

Fixes: #3771

The prior code left rubbish at the end of the vector, and never erased anything


mattkae (approved):
We do indeed have to erase things :+1:

Issue #3776, opened 2025-02-25
[SSDs] the cursor icons for top-left and bottom-left resize are wrong


AlanGriffiths:
Both the `InputManager::set_cursor()` code and the `xcursor_name_for_mir_cursor()` code seem to believe there are no "nw-resize" and "sw-resize" (a.k.a. "top_left_corner" & "bottom_left_corner") cursors.

PR #3777, opened 2025-02-25
Use the *correct* xcursor cursor names on decorations

AlanGriffiths:
We don't really have a use for the `mir_*_cursor_name` constants, just bypass them and use the right names correctly

Fixes: #3776


Saviq (approved):
Why do you remove the default cursor here?

Otherwise looks legit.

AlanGriffiths:
> Why do you remove the default cursor here?

I deduplicated the default cursor as the second copy isn't ever used

PR #3789, opened 2025-02-28
Exposing the_input_scene to users who want it

mattkae:
I want the input::Scene in miracle

### DIFF
--- a/src/include/server/mir/server.h
+++ b/src/include/server/mir/server.h
@@ -67,6 +67,7 @@ class SessionLock;
 }
 namespace input
 {
+class Scene;
 class SeatObserver;
 }
 namespace renderer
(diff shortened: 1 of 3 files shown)


AlanGriffiths (commented):
I'd like to understand the purpose of this.

The `input::Scene` is the interface to the scene required _by the input subsystem_, so what is the usecase for users wanting it?

mattkae:
> I'd like to understand the purpose of this.
> 
> The `input::Scene` is the interface to the scene required _by the input subsystem_, so what is the usecase for users wanting it?

It is the only interface that allows me to do an `emit_scene_changed` to trigger a rerender whenever I want. I made the minimal change here, but I would be okay with introducing a new interface for this if need be.

mattkae:
I originally wanted this in order to pump a custom renderable and and emit a scene change so that I could implement fade-out on surface for miracle-wm, but that ended up being unnecessary. Instead, I am inserting a custom surface into the SurfaceStack, which works much better!

Issue #3788, opened 2025-02-28
"prefer-ssd" isn't working


AlanGriffiths:
This is on main, 2.19 seems fine

```
MIRAL_SHELL_DECORATIONS=prefer-ssd miral-app
```

Start a Qt application (qterminal, kate, etc)

Expect: SSDs
Actual: Usually no decorations (but occasionally they appear)

AlanGriffiths:
Bisected to: 9962fa7087e6e92ce3a99f46cf32373f3dee2dbf

PR #3791, opened 2025-03-03
Do not allow max_{width,height} to overflow

AlanGriffiths:
Fixes: #3788


tarek-y-ismail:
Why are clients sending int max anyway?

AlanGriffiths:
> Why are clients sending int max anyway?

Because it isn't forbidden

RAOF (approved):
Yup, that works.

Pity we don't have C++26 support; [`add_sat`](https://en.cppreference.com/w/cpp/numeric/add_sat) would be exactly what we wanted here.

Issue #3792, opened 2025-03-03
[main & 2.20] Windows not rendering, black rectangles, ...


AlanGriffiths:
Running Miriway from both `edge` and `beta` gives a lot of rendering artefacts:

* yambar is black unless cursor is moving over it
* A lot of dialogs & popups are black until resized
* synapse appears with a flickering black rectangle around it

I've eliminate Miriway changes as a cause by building Miriway/main locally (which picks up Mir 2.19.3 from ppa:mir-team/release)

(Running dual external monitors via HDMI/usb-c)

AlanGriffiths:
And... reverting #3784 fixes it

AlanGriffiths:
Not sure where I went wrong last time around, nor why my "fix" seemed to fix it. But I've now tracked this down to the `Miroil" changes (#3767). That is a lot more plausible than the lifetime hacks...

AlanGriffiths:
More notes on this:

* I'm seeing this on my Framework (Ubuntu 24.04/amdgpu) laptop 
  * Reproduce with main and miral-app
  * also with just the internal display
* I don't see this on my X1C3 (Fedora 42/i915) laptop

AlanGriffiths:
Reverting the changes to `ShmBuffer::bind()` does fix the rendering problems...

```diff
diff --git a/src/platforms/common/server/shm_buffer.cpp b/src/platforms/common/server/shm_buffer.cpp
index 3f9caeb72e..4fdeeffa2c 100644
--- a/src/platforms/common/server/shm_buffer.cpp
+++ b/src/platforms/common/server/shm_buffer.cpp
@@ -222,7 +222,21 @@ mg::NativeBufferBase* mgc::ShmBuffer::native_buffer_base()
 
 void mgc::ShmBuffer::bind()
 {
-    glBindTexture(GL_TEXTURE_2D, tex_id());
+    std::lock_guard lock{tex_id_mutex};
+    bool const needs_initialisation = tex_id_ == 0;
+    if (needs_initialisation)
+    {
+        glGenTextures(1, &tex_id_);
+    }
+    glBindTexture(GL_TEXTURE_2D, tex_id_);
+    if (needs_initialisation)
+    {
+        // The ShmBuffer *should* be immutable, so we can just upload once.
+        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
[...]

PR #3793, opened 2025-03-03
Another way of fixing destructor order

AlanGriffiths:
More complex than https://github.com/canonical/mir/pull/3784, but doesn't have unwanted side-effect

Fixes:  #3792

### DIFF
--- a/src/server/compositor/multi_threaded_compositor.h
+++ b/src/server/compositor/multi_threaded_compositor.h
@@ -74,8 +74,8 @@ private:
     void destroy_compositing_threads();
 
     std::shared_ptr<graphics::Display> const display;
-    std::shared_ptr<Scene> const scene;
     std::shared_ptr<DisplayBufferCompositorFactory> const display_buffer_compositor_factory;
+    std::shared_ptr<Scene> const scene;
     std::shared_ptr<DisplayListener> const display_listener;
     std::shared_ptr<CompositorReport> const report;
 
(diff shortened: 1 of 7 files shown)


src/server/compositor/multi_threaded_compositor.h
```diff
@@ -74,8 +74,8 @@ class MultiThreadedCompositor : public Compositor
     void destroy_compositing_threads();
 
     std::shared_ptr<graphics::Display> const display;
-    std::shared_ptr<Scene> const scene;
     std::shared_ptr<DisplayBufferCompositorFactory> const display_buffer_compositor_factory;
+    std::shared_ptr<Scene> const scene;
```
RAOF, line 78: This is the only active ingredient, right? Moving `scene` to be destroyed before `display_buffer_compositor_factory` which incidentally has shared ownership of the platforms?

Parameter order in the constructor isn't relevant, so we don't need all the other changes?

RAOF (changes requested):

src/server/compositor/multi_threaded_compositor.h
AlanGriffiths: Agreed the parameter order isn't an essential part of the fix: only the data member order and the initialiser order are required. But an inconsistency between the parameter order and the initialiser order gives the reader something unnecessary to think about: it is simpler with the parameter order change.

Saviq: I agree, the _change_ is more complex, but the result is not.

PR #3796, opened 2025-03-04
Fix shutdown order

AlanGriffiths:
Keep references to server components by weak pointer so that we do not extend their life beyond the life of the server


AlanGriffiths:
Don't land yet - I need to check against #3792

tarek-y-ismail (approved):
Can confirm this seems to fix the crash. Feel free to merge if #3792 doesn't rear its head.

AlanGriffiths:
> I need to check against #3792

Weird. #3792 is back, but is also on main.

Anyway, this is the right way to fix the destruction order. I will investigate #3792 again (Maybe there's something intermittent that confused my bisection?)

PR #3800, opened 2025-03-06
mgc::ShmBuffer: Fix texture setup synchronisation

RAOF:
Because we're on an `EGLContextExecutor` we're probably on a different
thread to where the GL state is going to be used, *and* we don't
have implicit flushing with `eglMakeCurrent` happening (because
the context just stays current on the `EGLContextExecutor`.

If the GL implementation has per-thread execution queues (for example,
amdgpu by default), this might mean that the texture setup commands
aren't visible to command stream that's actually using the
texture.

Explicitly `glFlush()` after our texture setup, to ensure
these commands are visible to any `EGLContext` that might
need them.

Closes: #3792

### DIFF
--- a/include/test/mir/test/doubles/mock_gl.h
+++ b/include/test/mir/test/doubles/mock_gl.h
@@ -65,6 +65,7 @@ public:
     MOCK_METHOD(void, glEnable, (GLenum));
     MOCK_METHOD(void, glEnableVertexAttribArray, (GLuint));
     MOCK_METHOD(void, glFinish, ());
+    MOCK_METHOD(void, glFlush, ());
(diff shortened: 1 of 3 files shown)


AlanGriffiths:
Fixes the bug but...

```
250/255 Test #252: unit-tests-do-not-link-with-GL-or-EGL ......................................................................***Failed    0.02 sec
  NEEDED               libGLESv2.so.2
  NEEDED               libGLESv2.so.2
```

AlanGriffiths:
Looks like MockGL is missing glFlush

AlanGriffiths (approved):
Makes more sense than my hack

PR #3802, opened 2025-03-06
Streamline decorations buffer management

tarek-y-ismail:


tarek-y-ismail:
> Yeah, this is simpler.
> 
> The downside is that it introduces a dependency on libmirplatform-dev, which isn't as stable as we would like it to be. 🤷

I think keeping things simple while we get user customizable decorations implemented is worth it. What do you think? Is it merge-worthy?

AlanGriffiths:
> What do you think? Is it merge-worthy?

I approved it. But it's a draft, so didn't queue.

Want to give @RAOF or @Saviq a chance to review? (Opinions can reasonably differ on the libplatform dependency)

tarek-y-ismail:
> Want to give @RAOF or @Saviq a chance to review? (Opinions can reasonably differ on the libplatform dependency)

Yeah, my discussion with Chris' was why I noticed all of this to start, so it would be a good idea to get his opinions as well.

Saviq (approved):
> Opinions can reasonably differ on the libplatform dependency

I _think_ server side decorations are bound to the compositor enough that it's not wrong to have a tighter dependency like that.

AlanGriffiths:
@tarek-y-ismail is this ready for review yet?

tarek-y-ismail:
> @tarek-y-ismail is this ready for review yet?

Yup. Since it already got approval, should I merge or do you want to take a second look?

AlanGriffiths:
> Yup. Since it already got approval, should I merge or do you want to take a second look?

Want to give @RAOF a chance to review? (Opinions can reasonably differ on the libplatform dependency)

RAOF (approved):
Yeah, I'm in favour of this. Changes to this level of the platform should be easy enough to hide, or sufficiently infrequent to not be a problem.

PR #3811, opened 2025-03-08
Use drmModeGetResources as fallback for modeset probe

Azkali:
This allows the use of drmModeGetResources to check if KMS is supported on devices where drmCheckModesettingSupported is not supported or fails to detect.

Since we use drmModeGetResources later on, it's a safe check since if this fails, we will fail later anyway.

### DIFF
--- a/src/platforms/gbm-kms/server/display_helpers.cpp
+++ b/src/platforms/gbm-kms/server/display_helpers.cpp
@@ -130,7 +130,6 @@ mgmh::DRMHelper::open_all_devices(
         }
         else
         {
-            drmModeRes* drmDev;
             switch (auto err = -drmCheckModesettingSupported(busid.get()))
             {
             case 0: break;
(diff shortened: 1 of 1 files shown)


src/platforms/gbm-kms/server/display_helpers.cpp
```diff
@@ -144,9 +145,18 @@ mgmh::DRMHelper::open_all_devices(
 
                 [[fallthrough]];
             case EINVAL:
+                // Trying fallback
                 mir::log_warning(
-                    "Failed to detect whether device %s supports KMS, but continuing anyway",
+                    "Failed to detect whether device %s supports KMS, checking using fallback method",
                     device.devnode());
+
+                drmDev = drmModeGetResources(tmp_fd);
```
RAOF, line 153: This should probably be a `std::unique_ptr<drmModeRes>`, so we don't need to worry about remembering to clean it up.

RAOF (commented):
One stylistic nit, and one question:
Is this fixing something? Do you have a device for which `drmCheckModesettingSupported` returns `EINVAL`, and which then subsequently fails to work?

AlanGriffiths:
> @Azkali can you follow up on @RAOF's feedback?

@Saviq @RAOF I've pushed that refactor. (The only hard part was avoiding enormously long lines)

AlanGriffiths:
@Azkali: what is the motivation for this?

> Is this fixing something? Do you have a device for which `drmCheckModesettingSupported` returns `EINVAL`, and which then subsequently fails to work?

Azkali:
@AlanGriffiths the idea was to port other UT patches that where marked "to be upstreamed" I simply executed but I don't have the history of this patch

AlanGriffiths:
> @AlanGriffiths the idea was to port other UT patches that where marked "to be upstreamed" I simply executed but I don't have the history of this patch

We have nothing against patches being upstreamed, that's good for everyone. But we do need to understand the problem being addressed: It may be that there's a better or more general solution.

In this case, it would be useful to reference the original patch you are working from and any associated history

fredldotme:
Maybe need to tag the commit author @mariogrip who might know what it's for.

Saviq:
Please reopen if this is still needed.

Issue #3814, opened 2025-03-12
[miral] Failing to launch a client reports spurious "fatal error"


AlanGriffiths:
At least that's what I think happened:

```
[2025-03-11 16:34:17.382763] < -warning- > miral: Failed to execute client ("/usr/local/libexec/miriway-session-startup") error: No such file or directory
[2025-03-11 16:34:17.382797] < - debug - > miral: Configuring pointer: 'basic-window-manager'
[2025-03-11 16:34:17.382820] < - debug - > miral: Configuring pointer: 'x11-mouse-device'
!!! Fatal signal received. Attempting cleanup, but deadlock may occur
Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGSEGV
!!! Fatal signal received. Attempting cleanup, but deadlock may occur
Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGABRT
...
```

The server continued without other issues

PR #3815, opened 2025-03-12
Terminate immediately when exec() fails

AlanGriffiths:
When we fail to execute a client we don't want to try any cleanup, run global destructors or atexit entries

Fixes: #3814


mattkae (approved):
[The more you know](https://stackoverflow.com/questions/5422831/what-is-the-difference-between-using-exit-exit-in-a-conventional-linux-fo) :)

PR #3817, opened 2025-03-13
tests/wlcs: Load the expected failures from a file.

RAOF:
This makes it a little bit easier to extract the
list of known failures when outside the Mir buildsystem, such as in the DEP-8 test infrastructure.


AlanGriffiths (approved):
Nice idea

Issue #3816, opened 2025-03-12
Crash when disconnecting monitors


AlanGriffiths:
When unplugging USB-C cable with two HDMI/usb monitors...

```
(gdb) bt
#0  0x0000789bfca37620 in mir::graphics::atomic::AtomicKMSOutput::set_crtc (this=0x606646d4c420, fb=...)
    at /home/alan/CLionProjects/mir/src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp:314
#1  0x0000789bfca7a2a0 in mir::graphics::atomic::DisplaySink::set_crtc (this=0x789b4418c490, forced_frame=...)
    at /home/alan/CLionProjects/mir/src/platforms/atomic-kms/server/kms/display_buffer.cpp:147
#2  0x0000789bfca7a40a in mir::graphics::atomic::DisplaySink::post (this=0x789b4418c490) at /home/alan/CLionProjects/mir/src/platforms/atomic-kms/server/kms/display_buffer.cpp:180
#3  0x0000789bfe8aa1f1 in mir::compositor::CompositingFunctor::operator() (this=0x606647477330) at /home/alan/CLionProjects/mir/src/server/compositor/multi_threaded_compositor.cpp:159
#4  0x0000789bfe8b01f9 in std::__invoke_impl<void, mir::compositor::CompositingFunctor&> (__f=...) at /usr/include/c++/13/bits/invoke.h:61
#5  0x0000789bfe8b006b in std::__invoke<mir::compositor::CompositingFunctor&> (__fn=...) at /usr/include/c++/13/bits/invoke.h:96
#6  0x0000789bfe8af988 in std::reference_wrapper<mir::compositor::CompositingFunctor>::operator()<>() const (this=0x6066471435a0) at /usr/include/c++/13/bits/refwrap.h:359
#7  0x0000789bfe8aee52 in std::__invoke_impl<void, std::reference_wrapper<mir::compositor::CompositingFunctor>&> (__f=...) at /usr/include/c++/13/bits/invoke.h:61
#8  0x0000789bfe8ae050 in std::__invoke_r<void, std::reference_wrapper<mir::compositor::CompositingFunctor>&> (__fn=...) at /usr/include/c++/13/bits/invoke.h:111
#9  0x0000789bfe8ad0e8 in std::_Function_handler<void (), std::reference_wrapper<mir::compositor::CompositingFunctor> >::_M_invoke(std::_Any_data const&) (__functor=...)
[...]

PR #3820, opened 2025-03-13
AtomicKMSOutput::ensure_crtc() may see an output disconnect

AlanGriffiths:
This fixes a segfault in subsequent code that assumes the connector can be used

Fixes: #3816


AlanGriffiths:
@RAOF could you sanity check this approach? (I'll continue digging Monday)

AlanGriffiths:
There are definitely issues left: I just had a crash on reconnecting and switching to "docked", but this is a significant improvement.

Saviq (approved):
Looks legit.

Issue #3821, opened 2025-03-14
Detect shells duplicating Wayland protocol extensions


AlanGriffiths:
When testing Lomiri with Mir 2.20 a crash was reported due to both Mir and QtMir were implementing the same unstable protocol: xdg-decoration-unstable-v1. This is something we could detect in Mir and error on

stack trace: https://paste.ubuntu.com/p/nvnSz2PmMQ/

WAYLAND_DEBUG: https://paste.ubuntu.com/p/kgCsQxMHRW/

QtMir patch: https://paste.debian.net/1363018/

PR #3822, opened 2025-03-14
Quit on attempt to duplicate wayland extensions

AlanGriffiths:
Fixes: #3821

### DIFF
--- a/src/miral/wayland_extensions.cpp
+++ b/src/miral/wayland_extensions.cpp
@@ -217,6 +217,10 @@ struct miral::WaylandExtensions::Self
 
     void add_extension(Builder const& builder)
     {
+        if (supported_extensions.find(builder.name) != supported_extensions.end())
+        {
+            throw mir::AbnormalExit{"Attempt to duplicate wayland extensions: " + builder.name};
+        }
         wayland_extension_hooks.push_back(builder);
         supported_extensions.insert(builder.name);
         printf("%s = %s\n", __PRETTY_FUNCTION__, builder.name.c_str());
--- a/tests/miral/wayland_extensions.cpp
+++ b/tests/miral/wayland_extensions.cpp
@@ -29,6 +29,7 @@
 
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
+#include <mir/abnormal_exit.h>
 
 using namespace testing;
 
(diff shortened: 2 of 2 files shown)


RAOF (changes requested):
Hm. I think this means that every time we implement a new extension in Mir it's a potentially breaking change for shells - if they had implemented that extension, they can't run against new Mir.

I think we should go about this the other way - an extension added by a shell with `add_extension` that duplicates a Mir built-in should disable the built-in Mir implementation, (with a warning, suggesting people can migrate).

AlanGriffiths:
> I think this means that every time we implement a new extension in Mir it's a potentially breaking change for shells - if they had implemented that extension, they can't run against new Mir.

That's true now (as shown by the linked bug)

> an extension added by a shell with `add_extension` that duplicates a Mir built-in should disable the built-in Mir implementation

That can also cause problems (if we code assumptions based on the internal implementation)

RAOF:
> > I think this means that every time we implement a new extension in Mir it's a potentially breaking change for shells - if they had implemented that extension, they can't run against new Mir.
> 
> That's true now (as shown by the linked bug)

Yeah, but it's what we want to fix :wink: 

> 
> > an extension added by a shell with `add_extension` that duplicates a Mir built-in should disable the built-in Mir implementation
> 
> That can also cause problems (if we code assumptions based on the internal implementation)

Yeah, and likewise if the shell assumes their implementation is used. I think our options are:

1. Status quo: a non-obvious crash if both Mir and shell implement an extension. Bad, each extension we implement in Mir is a possibly-breaking change.
2. This PR: Mir cleanly refuses to work if both Mir and shell implement an extension. Makes implementing an extension in Mir a possibly-breaking change, but at least it's an obvious break.
3. My alternative: Mir uses the shell's implementation (with a warning) if duplicates exist. Implementing a new extension in Mir is no longer a breaking change, but requires defensive coding if extensions are entangled.
4. Versioned default extensions - shells explicitly choose a set of base Mir extensions. This seems like an excessively cautious approach, but *does* ensure no breaking changes.

All in all, since we don't expect this to happen much, (2) is fine for now and we can revisit this later if it turns into more of an issue.

RAOF (approved):

RAOF:
> > an extension added by a shell with `add_extension` that duplicates a Mir built-in should disable the built-in Mir implementation
> 
> That can also cause problems (if we code assumptions based on the internal implementation)

This is actually true now; users can disable arbitrary Wayland extensions, so if our code assumes the presence of a set of extensions it can be broken by user configuration.

AlanGriffiths:
> This is actually true now; users can disable arbitrary Wayland extensions, so if our code assumes the presence of a set of extensions it can be broken by user configuration.

True, but doesn’t need a rebuild to fix

Issue #3824, opened 2025-03-15
Looking up non-existing sessions and windows in MirAL leads to exception


fredldotme:
As noticed on Lomiri, the following procedure creates a crash, seemingly from a race:

- Share a file in Telegram
- Select a file in the GTK file dialog
- Observe crash stemming from removing an already gone surface

WAYLAND_DEBUG: https://paste.ubuntu.com/p/3rp74738Pc/
Stack trace: https://paste.ubuntu.com/p/wM4MSt7b44/

AlanGriffiths:
Reporting a post-condition failure by an exception (existing code) is established best practice.

The problem lies elsewhere: the situation should not happen, or the exception needs to be handled. Will investigate…

AlanGriffiths:
Looking at the stack trace, the client is destroying the XdgToplevel (which clearly must exist) but the surface is unknown to libmiral. That seems counter-intuitive: was the surface already "removed" somewhere other than the destructor?

AlanGriffiths:
@fredldotme is there any chance of capturing a log with MIR_SERVER_WINDOW_MANAGEMENT_TRACE set? And/or a stack trace at the point of "throw"?

It looks to me that the only `info_for()` on this path is at the end of:

```
void miral::BasicWindowManager::remove_surface(
    std::shared_ptr<scene::Session> const& session,
    std::weak_ptr<scene::Surface> const& surface)
{
    Locker lock{this};
    if (app_info.find(session) == app_info.end())
    {
        log_debug(
            "BasicWindowManager::remove_surface() called with unknown or already removed session %s (PID: %d)",
            session->name().c_str(),
            session->process_id());
        return;
    }

    if (!surface_known(surface, "remove"))
        return;

    remove_window(session, info_for(surface));
}
```

And I can't quite see how that gets past the `surface_know()` check.

fredldotme:
Stack trace catching the exception: https://paste.ubuntu.com/p/db7BK3qzjV/
Window management trace & WAYLAND_DEBUG since app start: https://paste.ubuntu.com/p/jSzbGgqhMH/

AlanGriffiths:
> Stack trace catching the exception

I was hoping for the throw trace

fredldotme:
Here: https://paste.ubuntu.com/p/57DPssGMP7/

AlanGriffiths:
Thanks! The problem is being detected here:

```c++
void ApplicationSelector::select(miral::Window const& window)
{
    if (selected)
[...]

PR #3825, opened 2025-03-15
miral: Avoid exception when looking up sessions & windows

fredldotme:
The way miral::BasicWindowManager::info_for() methods are used is without additional error checking whether those ApplicationInfos & WindowInfos actually exist.

Instead of triggering an exception in std::map::at() for out of bounds access, just check whether those are there and return empty informations otherwise.

Fixes a crash in Lomiri where destruction of an xdg_toplevel leads to looking up a non-existing session.

Reference: #3824

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -474,12 +474,24 @@ auto miral::BasicWindowManager::find_application(std::function<bool(ApplicationI
 auto miral::BasicWindowManager::info_for(std::weak_ptr<scene::Session> const& session) const
 -> ApplicationInfo&
 {
+    if (app_info.find(session) == app_info.end())
+    {
+        static ApplicationInfo const null_app_info;
+        return const_cast<ApplicationInfo&>(null_app_info);
+    }
+
     return const_cast<ApplicationInfo&>(app_info.at(session));
 }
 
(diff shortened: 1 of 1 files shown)


fredldotme:
Maybe it makes more sense to have the null's be statics..

AlanGriffiths:
First impression is this is fixing the wrong thing. Instead “going GIGO” and returning nonsense, the calling code should take responsibility for respecting preconditions. (Or handle the exception when they may be violated)

RAOF (changes requested):
This is a real problem, but I don't think this is an appropriate solution.

Firstly, this makes it easy for the consumer to invoke UB - we're handing back a non-const reference to a const object.

Secondly, as I now note @AlanGriffiths said, this'll return garbage rather than a detectable error.

(I also note that the existing code is not threadsafe with no easy way for client code to use correctly, but this isn't helping that)

AlanGriffiths:
I didn’t mean that there is no problem to resolve. However, at the local level, reporting a post-condition failure by an exception (existing code) is better than hiding the failure. The solution lies elsewhere

fredldotme:
Should the caller then catch the exception and log a warning instead?

AlanGriffiths:
Let's move the discussion of alternatives to the issue

PR #3830, opened 2025-03-18
Add key repeat rate/delay config options and properly merge config settings

tarek-y-ismail:
Split off from #3735


include/miral/miral/input_configuration.h
```diff
@@ -79,6 +79,10 @@ class InputConfiguration::Mouse
     void vscroll_speed(std::optional<double>const& val);
     void hscroll_speed(std::optional<double>const& val);
 
+    /// Copies the corresponding value from `other` for all unset fields.
+    /// \remark Since MirAL 5.3
+    void merge_settings_from(InputConfiguration::Mouse const& other);
```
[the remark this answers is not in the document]
tarek-y-ismail: I think `merge_from` is better since `merge_with` doesn't really indicate which object would be modified (if you don't look at the constness of the method and the parameter :))

include/miral/miral/input_configuration.h
```diff
@@ -65,6 +65,7 @@ class InputConfiguration::Mouse
 
     Mouse(Mouse const& that);
     auto operator=(Mouse that) -> Mouse&;
+    auto operator==(Mouse that) const -> bool;
```
mattkae, line 68: Should this be passed as a `const&`? I imagine we don't want to make a copy here, right?

src/miral/input_configuration.cpp
```diff
@@ -400,3 +410,41 @@ void miral::InputConfiguration::Keyboard::set_repeat_rate(int new_rate) {
...
+{
+    self->merge_from(*other.self);
+}
+
+void miral::InputConfiguration::Keyboard::merge_from(InputConfiguration::Keyboard const& other)
+{
+    self->merge_from(*other.self);
+}
+
+auto miral::InputConfiguration::Mouse::operator==(Mouse that) const -> bool
```
mattkae, line 429:
```suggestion
auto miral::InputConfiguration::Mouse::operator==(Mouse const& that) const -> bool
```

src/miral/input_configuration.cpp
```diff
@@ -400,3 +410,41 @@ void miral::InputConfiguration::Keyboard::set_repeat_rate(int new_rate) {
...
+{
+    return self->handedness == that.self->handedness && self->acceleration == that.self->acceleration &&
+           self->acceleration_bias == that.self->acceleration_bias && self->vscroll_speed == that.self->vscroll_speed &&
+           self->hscroll_speed == that.self->hscroll_speed;
+}
+
+auto miral::InputConfiguration::Touchpad::operator==(Touchpad that) const -> bool
```
mattkae, line 436:
```suggestion
auto miral::InputConfiguration::Touchpad::operator==(Touchpad const& that) const -> bool
```

src/miral/input_configuration.cpp
```diff
@@ -400,3 +410,41 @@ void miral::InputConfiguration::Keyboard::set_repeat_rate(int new_rate) {
...
+    return self->disable_while_typing == that.self->disable_while_typing &&
+           self->disable_with_external_mouse == that.self->disable_with_external_mouse &&
+           self->acceleration == that.self->acceleration && self->acceleration_bias == that.self->acceleration_bias &&
+           self->vscroll_speed == that.self->vscroll_speed && self->hscroll_speed == that.self->hscroll_speed &&
+           self->click_mode == that.self->click_mode && self->scroll_mode == that.self->scroll_mode &&
+           self->tap_to_click == that.self->tap_to_click &&
+           self->middle_button_emulation == that.self->middle_button_emulation;
+}
+
+auto miral::InputConfiguration::Keyboard::operator==(Keyboard that) const -> bool
```
mattkae, line 447:
```suggestion
auto miral::InputConfiguration::Keyboard::operator==(Keyboard const& that) const -> bool
```

mattkae (changes requested):
A few minor things, but looking reasonable

include/miral/miral/input_configuration.h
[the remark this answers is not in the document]
tarek-y-ismail: > There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors.

src/miral/input_configuration.cpp
```diff
@@ -400,3 +410,41 @@ void miral::InputConfiguration::Keyboard::set_repeat_rate(int new_rate) {
...
+           self->vscroll_speed == that.self->vscroll_speed && self->hscroll_speed == that.self->hscroll_speed &&
+           self->click_mode == that.self->click_mode && self->scroll_mode == that.self->scroll_mode &&
+           self->tap_to_click == that.self->tap_to_click &&
+           self->middle_button_emulation == that.self->middle_button_emulation;
+}
+
+auto miral::InputConfiguration::Keyboard::operator==(Keyboard const& that) const -> bool
+{
+    return self->repeat_rate == that.self->repeat_rate && self->repeat_delay == that.self->repeat_delay;
+}
```
AlanGriffiths, line 450: If these were needed it would be more maintainable and easier to review letting the compiler generate them:

```suggestion
auto miral::InputConfiguration::Mouse::operator==(Mouse const& that) const -> bool = default;

auto miral::InputConfiguration::Touchpad::operator==(Touchpad const& that) const -> bool = default;

auto miral::InputConfiguration::Keyboard::operator==(Keyboard const& that) const -> bool = default;
```

AlanGriffiths (changes requested):
1. The tests only cover the cases where the target is default initialized.
2. The tests do not cover the accessor methods
3. The comparison operators are not needed, they are only there to bypass the accessor methods for the purpose of testing

Providing non-member comparison operators _in the test code_ that test the accessors used by the production code would be better.

```c++
namespace
{
auto operator==(Touchpad const& lhs, Touchpad const& rhs) const -> bool
{
    return std::tie(
            lhs.disable_while_typing(), 
            lhs.disable_with_external_mouse()
            ...)
        == std::tie(
            rhs.disable_while_typing(), 
            rhs.disable_with_external_mouse()
            ...);
}
```

AlanGriffiths:
> > The tests only cover the cases where the target is default initialized.
> 
> You mean `Mouse` and friends? As far as I know that's the only way to initialize them?

I mean when target isn't in the initialised state:

```c++
TEST_F(TestInputConfiguration, mouse_merge_does_not_overwrite_values)
{
    std::initializer_list<std::function<void(Mouse&)>> const setters = {
        [](auto& mouse_config)
        {
            mouse_config.handedness(mir_pointer_handedness_right);
        },
        [](auto& mouse_config)
        {
            mouse_config.acceleration(mir_pointer_acceleration_adaptive);
        },
        [](auto& mouse_config)
        {
            mouse_config.acceleration_bias(-1);
        },
        [](auto& mouse_config)
        {
            mouse_config.vscroll_speed(2.0);
        },
        [](auto& mouse_config)
        {
            mouse_config.hscroll_speed(3.0);
        },
    };

    auto const target_settings = [](auto& mouse_config)
        {
            mouse_config.handedness(mir_pointer_handedness_left;
            mouse_config.acceleration(mir_pointer_acceleration_none);
            mouse_config.acceleration_bias(1);
            mouse_config.vscroll_speed(3.0);
            mouse_config.hscroll_speed(2.0);
        };


    for(auto const& setter: setters)
    {
        Mouse expected;
        setter(expected);
        target_settings(expected);

        Mouse modified;
        setter(modified);

        Mouse target;
        target_settings(target); // Set some values that should not be overridden
        target.merge(modified);
        ASSERT_EQ(target, expected);
    }
}
```

> > The tests do not cover the accessor methods
> 
> The only accessors that do anything special are `{Mouse,Touchpad}::acceleration_bias`, I can test those, but the rest seem unnecessary?

The behaviour we want to test is that the accessors return the expected values, but that can be done by calling a test == function

tarek-y-ismail:
@AlanGriffiths Had veer a bit off your suggestions, but hopefully that's okay. Care to give the new commits a quick review?

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,240 @@
...
+        ASSERT_EQ(touch.acceleration_bias(), value);
+    }
+    for (auto const& [value, expected] : clamped_test_values)
+    {
+        touch.acceleration_bias(value);
+        ASSERT_EQ(touch.acceleration_bias(), expected);
+    }
+}
+
+namespace miral
```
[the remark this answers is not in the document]
tarek-y-ismail: I don't think your earlier suggestion [here](https://github.com/canonical/mir/pull/3830#pullrequestreview-2699475789) will quite work. After a bit of googling, I've reached the conclusion that I either have to define the non-member equality operators in the `miral` namespace (my original conclusion), use `EXPECT_TRUE(thing == other_thing)`, or use `EXPECT_PRED2(predicate, thing, other_thing)`

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,538 @@
...
+        if (setting_to_test != hscroll_speed)
+            mouse_config.hscroll_speed(2.0);
+    };
+
+    for (auto const& [setter, setting_to_test] : std::ranges::zip_view(mouse_setters, settings_to_test))
+    {
+        // Initialize all members, overwrite the one we're testing
+        Mouse expected;
+        target_settings(expected, setting_to_test);
+        setter(expected);
```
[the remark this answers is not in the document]
tarek-y-ismail: It sets all properties _except_ the one we're testing, that will get a special value a bit later in the test.

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,538 @@
...
+
+        // Set the value of member we're testing
+        Mouse modified;
+        setter(modified);
+
+        // Initialize all members, _except_ the one we're testing
+        Mouse target;
+        target_settings(target, setting_to_test);
```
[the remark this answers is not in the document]
tarek-y-ismail: We set it, but since we're testing `merge`, we set it in the object we're merging from.

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,538 @@
...
+
+        // This merge should only modify the variable we're testing
+        target.merge(modified);
```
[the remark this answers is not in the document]
tarek-y-ismail: This test is meant to make sure that when merging from an object with some property set, no other properties get accidentally modified as well.

For example, let's say our struct has three properties: prop 1, prop 2, and prop 3. We have three setters, one for each. 

Let's say we're in the iteration to test prop 2. 
1. we set up our expected variable by setting prop 1 and 3 via `target_settings`. Then we set prop 2 via `setter`. Now we have all three props set.
2. Then we set up `modified` by setting prop 2 in it using `setter`.
3. Then we set up `target` through `target_settings` similar to the first half of step 1.
4. The final set up step is to merge `modified` into `target`, which should result in **only** prop 2 being set in `target`, the rest of the values shouldn't change. If for example there's a bug and prop 1 gets changed as well, this test will catch it.
5. Finally, we make sure that both our expected struct, created without merging, and our target struct, modified through merging, are equal.

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,538 @@
...
+    for (auto const& [setter, setting_to_test] : std::ranges::zip_view(mouse_setters, settings_to_test))
+    {
+        // Initialize all members, overwrite the one we're testing
+        Mouse expected;
+        target_settings(expected, setting_to_test);
+        setter(expected);
+
+        // Set the value of member we're testing
+        Mouse modified;
+        setter(modified);
```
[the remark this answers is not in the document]
tarek-y-ismail: It does. Please give the following two comment replies a look.

tests/miral/test_input_configuration.cpp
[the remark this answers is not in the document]
AlanGriffiths: Ah, got it now

AlanGriffiths (changes requested):
OK, I didn't understand the test - it was too clever for me to grasp how it works (and I still think the name suggests something different).

I think "too clever" is a problem: Too much is being done in the test body and in an unusual way.

Another problem is that it is very unclear what the problem is when the test fails. If I hack the test to force failures:

```
        target.hscroll_speed(6.0); // hack the test to force failures
        EXPECT_PRED2(mouse_equal, target, expected);
```
This is the unhelpful output:
```
$ cmake-build-debug/bin/miral-test-internal --gtest_filter=TestInputConfiguration.mouse_merge_does_not_overwrite_values
LD_LIBRARY_PATH=cmake-build-debug/bin/../lib/
MIR_SERVER_PLATFORM_PATH=cmake-build-debug/bin/../lib/server-modules/
exec=cmake-build-debug/bin/miral-test-internal.bin
Running main() from main.cpp
Note: Google Test filter = TestInputConfiguration.mouse_merge_does_not_overwrite_values
[==========] Running 1 test from 1 test suite.
[----------] 1 test from TestInputConfiguration
[ RUN      ] TestInputConfiguration.mouse_merge_does_not_overwrite_values
/home/alan/CLionProjects/mir/tests/miral/test_input_configuration.cpp:306: Failure
mouse_equal(target, expected) evaluates to false, where
target evaluates to 8-byte object <90-55 D3-8B 36-56 00-00>
expected evaluates to 8-byte object <B0-43 D3-8B 36-56 00-00>

/home/alan/CLionProjects/mir/tests/miral/test_input_configuration.cpp:306: Failure
mouse_equal(target, expected) evaluates to false, where
target evaluates to 8-byte object <90-55 D3-8B 36-56 00-00>
expected evaluates to 8-byte object <B0-43 D3-8B 36-56 00-00>

/home/alan/CLionProjects/mir/tests/miral/test_input_configuration.cpp:306: Failure
mouse_equal(target, expected) evaluates to false, where
target evaluates to 8-byte object <90-55 D3-8B 36-56 00-00>
expected evaluates to 8-byte object <B0-43 D3-8B 36-56 00-00>

/home/alan/CLionProjects/mir/tests/miral/test_input_configuration.cpp:306: Failure
mouse_equal(target, expected) evaluates to false, where
target evaluates to 8-byte object <90-55 D3-8B 36-56 00-00>
expected evaluates to 8-byte object <B0-43 D3-8B 36-56 00-00>

/home/alan/CLionProjects/mir/tests/miral/test_input_configuration.cpp:306: Failure
mouse_equal(target, expected) evaluates to false, where
target evaluates to 8-byte object <90-55 D3-8B 36-56 00-00>
expected evaluates to 8-byte object <B0-43 D3-8B 36-56 00-00>

[  FAILED  ] TestInputConfiguration.mouse_merge_does_not_overwrite_values (0 ms)
[----------] 1 test from TestInputConfiguration (0 ms total)

[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] TestInputConfiguration.mouse_merge_does_not_overwrite_values

 1 FAILED TEST
```

Contrast that with:

```
        EXPECT_THAT(target.hscroll_speed(), Eq(expected.hscroll_speed()));
```

Which gives helpful output:

```
...
/home/alan/CLionProjects/mir/tests/miral/test_input_configuration.cpp:309: Failure
Value of: target.hscroll_speed()
Expected: is equal to (3)
  Actual: (6)
...
```

tl;dr: there are several changes I think should be made:

1. The test(s) should be parameterised (`TEST_P`) and instantiated for each property.
2. The properties should be compared individually (that could be in a helper function)

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,653 @@
...
+        EXPECT_EQ(touch_config.tap_to_click(), value);
+    }
+
+    {
+        Touchpad touch_config;
+        auto const value = true;
+        touch_config.middle_mouse_button_emulation(value);
+        EXPECT_EQ(touch_config.middle_mouse_button_emulation(), value);
+    }
+}
```
AlanGriffiths, line 497: This looks like five tests in one
```suggestion
struct TestMouseInputConfiguration : TestInputConfiguration
{
    Mouse mouse_config;
};

TEST_F(TestMouseInputConfiguration, mouse_acceleration_return_expected_value)
{
    auto const value = mir_pointer_acceleration_adaptive;
    mouse_config.acceleration(value);
    EXPECT_THAT(mouse_config.acceleration(), Eq(value));
}

TEST_F(TestMouseInputConfiguration, mouse_acceleration_bias_return_expected_value)
{
    auto const value = -1;
    mouse_config.acceleration_bias(value);
    EXPECT_THAT(mouse_config.acceleration_bias(), Eq(value));
}

...
```

AlanGriffiths (changes requested):
It feels that there ought to be a way to parameterize the testing with the parameters being the initial, merging and expected properties.

If you search for `INSTANTIATE_TEST_SUITE_P(MinimalWindowManagerMaximizedSurfaceExclusionZoneTest` you'll see something similar

Then the test body becomes something like

```
TEST_P(Foo, bar)
{
    auto const param = GetParam();

    setup(merged, param.merging)
    setup((target, param.initial)

    target.merge(merged);

    check_result(target, param.expected);
}

**Comment by tarek-y-ismail** [comment]
@AlanGriffiths I implemented something as close as possible to what you mentioned while trying to minimize boilerplate in https://github.com/canonical/mir/pull/3830/commits/716542d773e2fe8c8b1d48c04c8376fa48cdb63b, does this work for you?

**Review by AlanGriffiths — CHANGES_REQUESTED** [review]
There's a bit of "and elsewhere" in the test suggestions, but we're definitely heading in a better direction :+1:

File: tests/miral/test_input_configuration.cpp [hunk 13]
```diff
@@ -0,0 +1,631 @@
...
+struct TestMouseInputConfiguration: public TestInputConfiguration
+{
+    Mouse mouse_config;
+};
+
+TEST_F(TestMouseInputConfiguration, mouse_acceleration_return_expected_value)
+{
+    auto const value = mir_pointer_acceleration_adaptive;
+    mouse_config.acceleration(value);
+    EXPECT_EQ(mouse_config.acceleration(), value);
```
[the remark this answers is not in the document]
tarek-y-ismail: Why not a parametric test?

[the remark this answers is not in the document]
AlanGriffiths: Also possible, but there's a trade off between boilerplate for that parametric tests and extra logic in a test. Sometimes the extra test logic is easier to follow. C.f. `mouse_acceleration_bias_is_set_and_clamped` where you've chose the same way

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,615 @@
...
+INSTANTIATE_TEST_SUITE_P(TestInputConfiguration, TestMouseMergeOneProperty, ::testing::ValuesIn(all_mouse_props));
+
+struct TestMouseMergeOnePropertyWithoutOverwrite :
+    public TestInputConfiguration,
+    testing::WithParamInterface<MouseProperty>
+{
+};
+
+// Make sure tht merging does not alter other data members
+TEST_P(TestMouseMergeOnePropertyWithoutOverwrite, mouse_merge_does_not_overwrite_values)
```
AlanGriffiths, line 420: Comment and code disagree! The test body does what the comment says.

Maybe we need another test that tests what test name says?

```c++
// Make sure tht merging does not alter other data members
TEST_P(TestMouseMergeOnePropertyWithoutOverwrite, mouse_merge_does_not_overwrite_values)
{
    auto const property = GetParam();

    auto const expected = get_mouse_config_with_properties(all_mouse_props);
    auto const modified = get_mouse_config_with_different_values_for_properties(all_mouse_props);
    auto target = get_mouse_config_with_properties(all_mouse_props);

    target.merge(modified);

    mouse_expect_equal(target, expected);
}
```

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,615 @@
...
+    repeat_delay
+};
+static auto const all_keyboard_props = {KeyboardProperty::repeat_rate, KeyboardProperty::repeat_delay};
+
+struct TestInputConfiguration: testing::Test
+{
+    inline static auto const unclamped_test_values = {-1.0, -0.5, 0.0, 0.5, 1.0};
+    inline static auto const clamped_test_values = {std::pair{-10.0, -1.0}, {-1.1, -1.0}, {1.1, 1.0}, {12.0, 1.0}};
+
+    inline static std::unordered_map<MouseProperty, PropertySetter<Mouse>> const mouse_setters = {
```
AlanGriffiths, line 87: Only used in one place, should this move there?

tests/miral/test_input_configuration.cpp
tarek-y-ismail: What I originally intended was that merging from a config with a property of type `double` wouldn't change other properties of type `double`. But your extra test is another case I didn't think of. Thanks!

AlanGriffiths: Don't credit me. I just reflected back to you what your test name says. (You need a better name for the test that does what the comment says)

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,690 @@
...
+    target.merge(modified);
+
+    mouse_expect_equal(target, expected);
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    TestInputConfiguration, TestMouseMergeOnePropertyWithoutOverwrite, ::testing::ValuesIn(all_mouse_props));
+
+// Make sure tht merging with properties already set results in no change.
+TEST_F(TestMouseMergeOnePropertyWithoutOverwrite, mouse_merge_with_properties_already_set_does_not_change_target)
```
AlanGriffiths, line 439: The test fixture name is repeated in the test name
```suggestion
TEST_F(TestMouseMergeOnePropertyWithoutOverwrite, does_not_change_target_properties_that_are_already_set)
```

tests/miral/test_input_configuration.cpp
```diff
@@ -0,0 +1,690 @@
...
+        Mouse different_set_of_values;
+
+        different_set_of_values.handedness(mir_pointer_handedness_left);
+        different_set_of_values.acceleration(mir_pointer_acceleration_none);
+        different_set_of_values.acceleration_bias(1.0);
+        different_set_of_values.vscroll_speed(37.0);
+        different_set_of_values.hscroll_speed(-19.0);
+
+        return different_set_of_values;
+    }();
```
AlanGriffiths, line 453: The immediately executed lambda is just so the variable can be const? I think we can do this directly

AlanGriffiths (changes requested):
There's a lot of repetition between the test fixtures and the test names:

```
TestTouchpadInputConfiguration, touchpad_disable_with_external_mouse_returns_expected_value
```

Doesn't really to say "touchpad" twice

tests/miral/test_input_configuration.cpp
tarek-y-ismail: In addition to the variable being `const`, it just also scopes the changes so accidental changes don't break things later on. Maybe I'm overthinking it though..

include/miral/miral/input_configuration.h
```diff
@@ -106,6 +110,8 @@ class InputConfiguration::Touchpad
     auto click_mode() const -> std::optional<MirTouchpadClickMode>;
     auto scroll_mode() const -> std::optional<MirTouchpadScrollMode>;
     auto tap_to_click() const -> std::optional<bool>;
+    /// \remark Since MirAL 5.3
+    auto middle_mouse_button_emulation() const -> std::optional<bool>;
```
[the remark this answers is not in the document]
AlanGriffiths: In general, separating things out into independent pieces of work makes it easier to review and land those pieces. For example, a separate 10 line PR could have landed while we're still discussing this.

AlanGriffiths (approved):
> In addition to the variable being `const`, it just also scopes the changes so accidental changes don't break things later on. Maybe I'm overthinking it though..

I think the simpler test body would be clear enough enough for accidents to be rare.  But :shrug:

mattkae (approved):
I reviewed a while ago, but approving so that you don't have to force it :)

AlanGriffiths:
For the avoidance of doubt. I have approved this, but not queued to give @tarek-y-ismail the opportunity to tweak first following discussion

PR #3832, opened 2025-03-18
Protect `AccessibilityManager` data member from racy mutations

tarek-y-ismail:
Addresses https://github.com/canonical/mir/pull/3735#discussion_r1999090513
Didn't really fit with either #3735 nor #3830, so I made it into its own PR.

### DIFF
--- a/src/include/server/mir/shell/accessibility_manager.h
+++ b/src/include/server/mir/shell/accessibility_manager.h
@@ -55,9 +55,9 @@ private:
     std::mutex mutable state_mutex;
     int repeat_rate_{25};
     int repeat_delay_{600};
-    bool enable_key_repeat;
 
-    bool enable_mouse_keys;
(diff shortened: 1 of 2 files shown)


RAOF (commented):
Generally speaking, I think we should be leaning on `mir::Synchronised<>` for these more; it makes it obvious what data is protected by the mutex.

(Also, `enable_key_repeat` and `enable_mouse_keys` should be `const`, to avoid the question "why isn't `enable_mouse_keys` under lock, too?)

I'm not marking this as "request changes" because it's correct and acceptable; I just think it can be improved. Feel free to land it as is, or take the suggestion.

tarek-y-ismail:
Thanks for the suggestions!

AlanGriffiths (approved):
OK now

But I still think that there should be an option for atomically setting both `repeat_rate` and `repeat_delay`

PR #3833, opened 2025-03-18
testing: adding attached surface tests for all compass directions

mattkae:
## What's new?
- Adding more tests for how different attached surfaces interact with one another given an exclusion zone that takes up the full dimensions of the surface

These tests are a bit redundant with those found in `window_placement_attached.cpp`, but this tests more of end-to-end scenario, so I think it is worthwhile

## To test
1. Build the project
2. Run `mir_window_management_tests`


tarek-y-ismail (changes requested):
Tiny stuff

tarek-y-ismail (approved):
https://github.com/canonical/mir/pull/3833#discussion_r2001392191 still remains, but that's harmless enough to let go.

PR #3836, opened 2025-03-19
tics: run on the self-hosted runner

Saviq:
---

This has a much lower complexity than https://github.com/canonical/mir/compare/main...tics-parallel and slightly better result, even if it is a little slower at first.

A run with this in place: https://github.com/canonical/mir/actions/runs/13925531981


AlanGriffiths (approved):
Sure

PR #3840, opened 2025-03-19
testing: add tests for closing the active window

mattkae:
## What's new?
- Add tests for closing windows and asserting facts about focus order and size
- Prioritizing using `tools()` over anything else to make changes to the underlying system. To accomplish this, I created a proxy object class called `NotifyingWindowManagerToolsImplementation` which mainly delegates changes to the wrapped object, but also has the property that it will notify the underlying system when a change has occurred. This allows us to queue up actions (which - in the normal system - would have happened asynchronously) and execute them before the request returns. This simplifies most events going into the system.
- Removed locks around the `surface_cache`, as they are guaranteed to be accessed synchronously now. The cache is only accessed (1) when we create a window or (2) when we remove a window. Both of these operations will happen from the thread that the tests are running on (there is no wayland or other thread that could modify these). This cache only exists to keep some shared_ptrs alive.

### DIFF
--- a/tests/include/mir_test_framework/window_management_test_harness.h
+++ b/tests/include/mir_test_framework/window_management_test_harness.h
@@ -41,6 +41,8 @@ class WindowManagementTestHarness : public HeadlessInProcessServer
 {
 public:
     WindowManagementTestHarness();
+    ~WindowManagementTestHarness() override;
+
     void SetUp() override;
     void TearDown() override;
 
@@ -54,7 +56,6 @@ public:
     void publish_event(MirEvent const& event) const;
     void request_resize(miral::Window const&, MirInputEvent const*, MirResizeEdge) const;
     void request_move(miral::Window const&, MirInputEvent const*) const;
-    void request_focus(miral::Window const&) const;
 
     auto focused(miral::Window const&) const -> bool;
     auto tools() const -> miral::WindowManagerTools&;
@@ -62,9 +63,9 @@ public:
 
     virtual auto get_builder() -> WindowManagementPolicyBuilder = 0;
     virtual auto get_output_rectangles() -> std::vector<mir::geometry::Rectangle> = 0;
-
+private:
     class Self;
-    std::shared_ptr<Self> self;
+    std::unique_ptr<Self> const self;
 };
 
 }
--- a/tests/mir_test_framework/window_management_test_harness.cpp
+++ b/tests/mir_test_framework/window_management_test_harness.cpp
@@ -21,15 +21,17 @@
 #include "mir_test_framework/window_management_test_harness.h"
 #include <mir/executor.h>
 #include <mir/scene/surface.h>
-#include <mir/scene/surface_observer.h>
+#include <mir/scene/null_surface_observer.h>
 #include <mir/shell/surface_specification.h>
 #include <mir/shell/shell.h>
 #include <mir/wayland/weak.h>
-#include <miral/window_manager_tools.h>
-#include <miral/set_window_management_policy.h>
 #include <mir/log.h>
 #include <mir/main_loop.h>
-#include <future>
+#include <miral/window_manager_tools.h>
+#include <miral/set_window_management_policy.h>
+#include <miral/zone.h>
+
+#include "src/miral/window_manager_tools_implementation.h"
 
 namespace ms = mir::scene;
 namespace msh = mir::shell;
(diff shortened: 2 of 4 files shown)


tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -130,12 +297,72 @@ class mir_test_framework::WindowManagementTestHarness::Self final : public ms::S
         pending_actions.clear();
     }
 
-    mir::Server& server;
+    class SurfaceObserver final : public ms::SurfaceObserver
+    {
+    public:
+        explicit SurfaceObserver(Self* self)
+            : self(self) {}
```
AlanGriffiths, line 304: It would separate concerns better to pass in a `std::function<void(ms::Surface const* surf)>` that invokes the logic on `Self`?

Then the construction site becomes something like:

```c++
    explicit Self(mir::Server& server)
        : server(server),
          surface_observer(std::make_shared<SurfaceObserver>([this])(ms::Surface const* surf) { client_surface_close_requested(surf); })
```

And the `client_surface_close_requested()` logic is implemented in `Self`

AlanGriffiths (changes requested):
You're removing locking around, for example, `surface_cache`. It isn't immediately apparent to me that this can only be accessed on one thread. Could you explain (maybe in comments)?

I guess the ideas is that this is a consequence of the way `NotifyingWindowManagerToolsImplementation` is implemented?

tests/include/mir_test_framework/window_management_test_harness.h
```diff
@@ -24,11 +24,14 @@
...
 
-namespace mir::scene
+namespace mir
+{
+namespace scene
 {
 class Surface;
 class Session;
 }
+}
```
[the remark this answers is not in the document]
mattkae: I don't think we do the `mir::scene` thing anywhere else, so I changed it to be more consistent with everything

mattkae:
> You're removing locking around, for example, `surface_cache`. It isn't immediately apparent to me that this can only be accessed on one thread. Could you explain (maybe in comments)?
> 
> I guess the ideas is that this is a consequence of the way `NotifyingWindowManagerToolsImplementation` is implemented?

Let me know if the comment that I added to the description clears that up!

mattkae:
> You're removing locking around, for example, `surface_cache`. It isn't immediately apparent to me that this can only be accessed on one thread. Could you explain (maybe in comments)?
> 
> I guess the ideas is that this is a consequence of the way `NotifyingWindowManagerToolsImplementation` is implemented?

I added a note about this. Let me know if it clears things up!

AlanGriffiths:
> I added a note about this. Let me know if it clears things up!

It would if it were with the code the next time someone reads it

mattkae:
> It would if it were with the code the next time someone reads it

I walked back my thinking on this @AlanGriffiths. We should reasonably support creating surfaces from external clients in testing, which would mean that those surfaces are added from another thread (e.g. the Wayland thread). Personally, I would find a lot of benefit from being able to test specific clients.

tests/include/mir_test_framework/window_management_test_harness.h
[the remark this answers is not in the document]
AlanGriffiths: The only reason it isn't done in most places is the syntax wasn't support when the code was written. We don't need to go backwards

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -582,14 +583,177 @@ TEST_F(MinimalWindowManagerTest,  when_a_window_is_focused_then_it_appears_above
...
+    auto const window2 = create_window(app2, spec);
+    auto const window3 = create_window(app3, spec);
+    EXPECT_TRUE(focused(window3));
+
+    // Act: Close the focused window
+    tools().ask_client_to_close(window3);
+
+    // Expect: window2 to have focus
+    EXPECT_TRUE(focused(window2));
+    EXPECT_TRUE(is_above(window2, window1));
```
AlanGriffiths, line 613: [non blocking] I wonder how hard it would be to support the "fluent" style of assertions:

```suggestion
    EXPECT_THAT(window2, IsAbove(window1));
```

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -582,14 +583,177 @@ TEST_F(MinimalWindowManagerTest,  when_a_window_is_focused_then_it_appears_above
...
+    spec.depth_layer() = mir_depth_layer_application;
+    spec.state() = mir_window_state_maximized;
+    auto const window = create_window(app, spec);
+
+    auto const top_spec = create_top_attached();
+    auto const attached = create_window(app, top_spec(output_rectangle.size));
+    EXPECT_EQ(window.size(), geom::Size(
+        output_rectangle.size.width.as_int(),
+        output_rectangle.size.height.as_int() - EXCLUSIVE_SURFACE_SIZE
+    ));
```
AlanGriffiths, line 745: We've been trying to shift to the "fluent" style of test assertions as they are often easier to read and give better diagnostics:

```suggestion
    EXPECT_THAT(window.size(), Eq(geom::Size(
        output_rectangle.size.width.as_int(),
        output_rectangle.size.height.as_int() - EXCLUSIVE_SURFACE_SIZE
    )));
```

AlanGriffiths (changes requested):

tests/window_management_tests/test_minimal_window_manager.cpp
mattkae: I can set out to do transfer the window management stuff over on a followup pr! This seems like a reasonable direction though.

tests/window_management_tests/test_minimal_window_manager.cpp
mattkae: I'll update the two instance of size/geometry equality that I added in this PR, but I will do the "fluent" update in a later iteration, just to keep the changes focused

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -57,90 +59,304 @@ class CachedSurfaceData
...
+        on_change();
+    }
+
+    auto active_application_zone() -> miral::Zone override
+    {
+        return tools.active_application_zone();
+    }
+private:
+    miral::WindowManagerTools tools;
+    std::function<void()> on_change;
```
AlanGriffiths, line 280: Looks like this can be `const`

tests/mir_test_framework/window_management_test_harness.cpp
```diff
@@ -57,90 +59,304 @@ class CachedSurfaceData
...
+              on_surface_closed(surf);
+          }))
+    {
+    }
+
+    void process_pending()
+    {
+        for (auto const& action : pending_actions)
+            action();
+        pending_actions.clear();
```
AlanGriffiths, line 300: I can see that we currently won't update `pending_actions` while iterating, but it seems plausible that future tests will lead `pending_actions` that lead recursively to more `pending_actions`.
```suggestion
        decltype(pending_actions) actions;
        actions.swap(pending_actions);
        for (auto const& action : actions)
            action();
```

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -582,14 +583,177 @@ TEST_F(MinimalWindowManagerTest,  when_a_window_is_focused_then_it_appears_above
...
+
+    auto const workspace2 = tools().create_workspace();
+    tools().add_tree_to_workspace(window3, workspace2);
+
+    EXPECT_TRUE(focused(window3));
+
+    // Act: Close the focused window
+    tools().ask_client_to_close(window3);
+
+    // Expect: window2 to have focus even though it's on another workspace
```
AlanGriffiths, line 689: IIRC `MinimalWindowManager` ignores workspaces. It is shell policies like FloatingWindowManager and downstream equivalents that specify that behaviour? Have I missed something?

AlanGriffiths (changes requested):

tests/mir_test_framework/window_management_test_harness.cpp
mattkae: Yup that makes sense

tests/window_management_tests/test_minimal_window_manager.cpp
mattkae: If you use `WindowManagerTools`, then you can make any existing policy have workspaces, which might be a bit odd. I can remove this test if you think it is unnecessary. It does, however, test a true fact about the `BasicWindowManager`.

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -582,14 +586,177 @@ TEST_F(MinimalWindowManagerTest,  when_a_window_is_focused_then_it_appears_above
...
+    auto const app1 = open_application("app1");
+    auto const app2 = open_application("app2");
+    auto const app3 = open_application("app3");
+    miral::WindowSpecification spec;
+    spec.size() = { geom::Width {100}, geom::Height{100} };
+    spec.depth_layer() = mir_depth_layer_application;
+    auto const window1 = create_window(app1, spec);
+    auto const window2 = create_window(app2, spec);
+    auto const window3 = create_window(app3, spec);
+    EXPECT_TRUE(focused(window3));
```
AlanGriffiths, line 609: While this is a valid test, it isn't what `if_active_window_is_removed_then_last_focused_window_is_focused` should test

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -582,14 +586,177 @@ TEST_F(MinimalWindowManagerTest,  when_a_window_is_focused_then_it_appears_above
...
+    spec.size() = { geom::Width {100}, geom::Height{100} };
+    spec.depth_layer() = mir_depth_layer_application;
+    spec.state() = mir_window_state_maximized;
+    auto const window = create_window(app, spec);
+
+    auto const top_spec = create_top_attached();
+    auto const attached = create_window(app, top_spec(output_rectangle.size));
+    EXPECT_THAT(window.size(), Eq(geom::Size(
+        output_rectangle.size.width.as_int(),
+        output_rectangle.size.height.as_int() - EXCLUSIVE_SURFACE_SIZE
```
AlanGriffiths, line 747: [pre-existing] I just spotted this: `UPPERCASE_IS_FOR_MACROS`, `snake_case_is_for_variables` (including constants)

AlanGriffiths (changes requested):
There seems to be a lot of repetition in the test setup. Maybe that's better done in a test fixture?

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -582,14 +586,177 @@ TEST_F(MinimalWindowManagerTest,  when_a_window_is_focused_then_it_appears_above
...
+    // Act: Close the focused window
+    tools().ask_client_to_close(window3);
+
+    // Expect: window1 to have focus since that is the parent
+    EXPECT_TRUE(focused(window1));
+    EXPECT_TRUE(is_above(window1, window2));
+}
+
+TEST_F(MinimalWindowManagerTest,
+    if_active_window_is_removed_and_we_cannot_focus_any_window_on_closing_window_workspaces_then_try_focus_within_app)
```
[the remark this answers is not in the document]
mattkae: I think that should be covered by `if_active_window_is_removed_then_last_focused_window_is_focused`, but let me rename it to be more obvious

tests/window_management_tests/test_minimal_window_manager.cpp
mattkae: They do affect the focus behavior when a window closes. We try to refocus after it closes, and that might be on another workspace according to the code.

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -582,14 +586,177 @@ TEST_F(MinimalWindowManagerTest,  when_a_window_is_focused_then_it_appears_above
...
+
+    // Expect: window1 to have focus since it is within our app but on another workspace
+    EXPECT_TRUE(focused(window1));
+    EXPECT_TRUE(is_above(window1, window2));
+}
+
+TEST_F(MinimalWindowManagerTest,
+    if_active_window_is_removed_and_we_cannot_focus_any_window_on_closing_window_workspaces_and_there_are_no_other_windows_in_this_app_then_focus_other_app)
+{
+    // Setup: Create three apps with three windows on three different workspaces
```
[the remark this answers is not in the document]
mattkae: The first window is added to an implicit workspace though, no? So it would be three?

tests/window_management_tests/test_minimal_window_manager.cpp
AlanGriffiths: Gah! The code keeps changing when I'm not looking. I wonder why it does that...

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -554,40 +558,152 @@ TEST_F(MinimalWindowManagerTest, can_resize_south_east_with_touch)
...
+    spec.size() = { geom::Width {100}, geom::Height{100} };
+    spec.parent() = windows[0];
+    auto const child_window = create_window(apps[0], spec);
+
+    tools().ask_client_to_close(child_window);
+    EXPECT_TRUE(focused(windows[0]));
+    EXPECT_TRUE(is_above(windows[0], windows[1]));
+}
+
+TEST_F(ThreeWindowsMinimalWindowManagerTest, if_active_window_is_removed_then_last_focused_window_on_same_workspace_is_focused)
```
AlanGriffiths, line 612: There are no workspaces in this test body, so why are the mentioned in the test name?
```suggestion
TEST_F(ThreeWindowsMinimalWindowManagerTest, if_active_window_is_removed_then_last_focused_window_is_focused)
```

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -554,40 +558,152 @@ TEST_F(MinimalWindowManagerTest, can_resize_south_east_with_touch)
...
+            windows[i] = create_window(apps[i], spec);
+        }
+    }
+
+    miral::Application apps[N];
+    miral::Window windows[N];
+};
+
+using TwoWindowsMinimalWindowManagerTest = MultipleWindowsMinimalWindowManagerTest<2>;
+using ThreeWindowsMinimalWindowManagerTest = MultipleWindowsMinimalWindowManagerTest<3>;
```
AlanGriffiths, line 584: I know it reads easier, but selecting tests with globbing is easier, and more consistent with prior art, with `--gtest_filter=MinimalWindowManager*`
```suggestion
using MinimalWindowManagerWithTwoWindows = MultipleWindowsMinimalWindowManagerTest<2>;
using MinimalWindowManagerWithThreeWindows = MultipleWindowsMinimalWindowManagerTest<3>;
```

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -554,40 +558,152 @@ TEST_F(MinimalWindowManagerTest, can_resize_south_east_with_touch)
...
+
+TEST_F(ThreeWindowsMinimalWindowManagerTest, new_windows_are_inserted_above_older_windows)
+{
+    EXPECT_TRUE(is_above(windows[2], windows[1]));
+    EXPECT_TRUE(is_above(windows[1], windows[0]));
+}
+
+TEST_F(ThreeWindowsMinimalWindowManagerTest, when_a_window_is_focused_then_it_appears_above_other_windows)
```
AlanGriffiths, line 592: This test depends on "focus stealing" being configured - maybe need tests for both options

tests/window_management_tests/test_minimal_window_manager.cpp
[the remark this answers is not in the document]
AlanGriffiths: I don't think it does cover it: I think such a test would need three windows, `window[0]` and `window[2]` in a workspace and closing the latter focuses the former (despite `window[1]` starting above `window[0]`)

tests/window_management_tests/test_minimal_window_manager.cpp
```diff
@@ -554,40 +558,152 @@ TEST_F(MinimalWindowManagerTest, can_resize_south_east_with_touch)
...
+    // Act: Close the focused window
+    tools().ask_client_to_close(window);
+
+    // Expect: no window to have focus
+    EXPECT_TRUE(focused({}));
 }
 
-TEST_F(MinimalWindowManagerTest,  when_a_window_is_focused_then_it_appears_above_other_windows)
+TEST_F(MinimalWindowManagerTest,
+    if_active_windows_session_closes_and_there_is_nothing_to_select_then_focused_is_emtpy)
```
AlanGriffiths, line 665:
```suggestion
    if_active_windows_session_closes_and_there_is_nothing_to_select_then_focused_is_empty)
```

tests/window_management_tests/CMakeLists.txt
```diff
@@ -1,6 +1,28 @@
-include(CMakeDependentOption)
+# We can't tell which version of gtest we're building against and INSTANTIATE_TEST_SUITE_P changed in
+# a way that relies on a gcc extension to support backward-compatible code, So...
+check_cxx_compiler_flag(-Wno-gnu-zero-variadic-macro-arguments MIRAL_COMPILE_WITH_W_NO_GNU_ZERO_VARIADIC_MACRO_ARGUMENTS)
+check_cxx_compiler_flag(-Wno-pedantic MIRAL_COMPILE_WITH_W_NO_PEDANTIC)
+if ("${CMAKE_CXX_COMPILER}" MATCHES ".*clang.*" AND MIRAL_COMPILE_WITH_W_NO_GNU_ZERO_VARIADIC_MACRO_ARGUMENTS)
+  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-gnu-zero-variadic-macro-arguments") # clang
+elseif(MIRAL_COMPILE_WITH_W_NO_PEDANTIC)
```
AlanGriffiths, line 7: `MIRAL_...`?

Actually, this is obviously copied from an old hack:  maybe we don't need to write backward-compatible code here?

AlanGriffiths (changes requested):
Mostly tidy-up

A couple of side issues about backward compatibility with some ancient gtest (which we likely don't care about) and why MinimalWindowManager is even aware of workspaces, but those are not blocking for this PR.

RAOF (approved):
The behaviour of workspaces seems odd to me - my *expectation* is that if I'm currently on a workspace then focus will never automatically leave it. My expectation is that closing the last window on a workspace will result in nothing being focused, regardless of the contents of other workspaces.

That said, this obviously tests the *current* behaviour; changes to that behaviour are out of scope!

The assertions will read better with the fluent `EXPECT_THAT(...` style follow-up.

This looks good to merge to me; feel free to commit the `const` nit and hit the button

AlanGriffiths:
> The behaviour of workspaces seems odd to me - my _expectation_ is that if I'm currently on a workspace then focus will never automatically leave it. My expectation is that closing the last window on a workspace will result in nothing being focused, regardless of the contents of other workspaces.

There is a problem in expectation here. For historical reasons Mir does not any semantics of "workspaces". It only tracks window membership.

miral-shell, Miriway and Miracle-WM all implement workspaces with similar, but distinct workspace models.

PR #3842, opened 2025-03-21
CMake: Don't error on some GCC warnings with high false-positive rates

RAOF:
`-Warray-bounds` and `-Wmaybe-uninitialized` have been known since GCC 11 to generate false positives. They come and go, as optimisations change, but we hit these sufficiently often to make it not useful to error on them.

Particularly: we get hit by [this specific array-bounds issue](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111415)

(Meta bugs on the GCC bugtracker for these are [array-bounds](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=56456) and [list for maybe-uninitialized](https://gcc.gnu.org/bugzilla/buglist.cgi?quicksearch=maybe%20uninitialized%20false%20positive))

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,10 +68,22 @@ set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--as-needed")
 set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,--as-needed")
 set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--as-needed")
 
+
 set(MIR_FATAL_COMPILE_WARNINGS ON CACHE BOOL "Should compiler warnings fail the build")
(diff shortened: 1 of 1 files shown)


AlanGriffiths (approved):
Do we want to localise this hack at all?

E.g. Only apply on specific compiler versions, or only on part of the code (IIRC this is most problematic in the test code)

RAOF:
> Do we want to localise this hack at all?

*Maaaaaybe*? It appears the problem here is that these diagnostics are very optimisation-dependent; the code needs to get optimised a bit, but not enough that the optimiser can prove the invalid situation doesn't occur. It's true that the GTest code seems to be fairly consistently just the right level of complexity to fall into this, but the underlying issue is sensitive to (apparrently) unrelated code motion and can be architecture-dependent.

I don't think that we've ever had a *true*-positive on either of these warnings, and where they pop up is going to change over time, so I lean towards not failing the build on these warnings globally.

If someone has a strong opinion I could expand the `MIR_COMPILER_QUIRKS` we've currently got to handle this; that'd mean we could just add an extra default quirk each time we hit something :woman_shrugging:

PR #3844, opened 2025-03-21
evdev/platform: cleanup device watcher properly

bhush9:
5205713574d5e7c9 fixed this somewhat but it is still behaving little bit weird with Ubuntu Touch on fairphone 5, in most cases, libinput_device_get_udev_device returns the null udev_device because by that time device is already removed from perspective of kernel too.

Which then results in the device being in device_watchers even when device is long gone, when attaching device next time, it sees that it is already in device_watchers and does not process them.

Fix this by cleaning up device_watcher entry when we get udev event, at that time we get valid devnum for device which is being removed.


bhush9:
Is this unstable test or something caused by my patch? It's bit weird.

AlanGriffiths:
> Is this unstable test or something caused by my patch? It's bit weird.

We see this fail intermittently. Probably not related to your patch

AlanGriffiths:
> So shall we merge this? :)

Sorry, I've not got to this yet. It is not forgotten

RAOF (approved):
Yes, this is sensible.

We don't need to do anything with `pending_devices`, as that should automatically get cleaned up anyway.

PR #3846, opened 2025-03-21
miroil drag'n'drop

### LINKED ISSUE #3845: [miroil] Dropping a DND operation keeps on sending wl_data_device.motion events
Drag and drop on Lomiri is not using any particular interfaces, as DND currently is handled underneath QtMir in Mir.
When stopping a DND operation the compositor keeps on sending wl_data_device.motion events.

WAYLAND_DEBUG: https://paste.ubuntu.com/p/nnh3yzJJ6C/

Way to reproduce it:
- Launch Nautilus and another window
- Start DND operation
- Either cancel or perform the DND operation (doesn't matter)
- Move mouse pointer over the other window
- See it crash
- Nautilus input gets stuck

AlanGriffiths:
Add ending gestures to miroil

Fixes #3845

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
 set(MIR_VERSION_MAJOR 2)
 set(MIR_VERSION_MINOR 20)
-set(MIR_VERSION_PATCH 1)
+set(MIR_VERSION_PATCH 2)
 
 add_compile_definitions(MIR_VERSION_MAJOR=${MIR_VERSION_MAJOR})
 add_compile_definitions(MIR_VERSION_MINOR=${MIR_VERSION_MINOR})
@@ -68,23 +68,13 @@ set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,--as-needed")
 set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -Wl,--as-needed")
 set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--as-needed")
 
-
 set(MIR_FATAL_COMPILE_WARNINGS ON CACHE BOOL "Should compiler warnings fail the build")
 if(MIR_FATAL_COMPILE_WARNINGS)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror")
-  # Disable maybe-uninitialized errors on GCC when optimisation is enabled
-  # They are notorious for false-positives
-  # Also, disable array-bounds errors, they are likewise significantly
-  # false-positivey
-  # (We particularly hit https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111415)
-  if ((CMAKE_C_COMPILER_ID MATCHES "GNU") AND NOT (CMAKE_BUILD_TYPE MATCHES "Debug"))
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-error=maybe-uninitialized -Wno-error=array-bounds")
-  endif()
-  if ((CMAKE_CXX_COMPILER_ID MATCHES "GNU") AND NOT (CMAKE_BUILD_TYPE MATCHES "Debug"))
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error=maybe-uninitialized -Wno-error=array-bounds")
-  endif()
 endif()
+
+set(MIR_COMPILER_QUIRKS "" CACHE STRING "Semicolon-separated list of compiler quirks to enable")
 
 include(CheckCXXCompilerFlag)
 check_cxx_compiler_flag(-Wmismatched-tags HAS_W_MISMATCHED_TAGS)
--- a/src/include/server/mir/default_server_configuration.h
+++ b/src/include/server/mir/default_server_configuration.h
@@ -385,6 +385,7 @@ protected:
     CachedPtr<input::CompositeEventFilter> composite_event_filter;
     CachedPtr<input::InputManager>    input_manager;
     CachedPtr<input::SurfaceInputDispatcher>    surface_input_dispatcher;
+    CachedPtr<frontend::PointerInputDispatcher> pointer_input_dispatcher;
     CachedPtr<input::DefaultInputDeviceHub>    default_input_device_hub;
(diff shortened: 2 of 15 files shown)


AlanGriffiths:
@fredldotme are you able to test this fixes #3845? I don't currently have a suitable environment set up

fredldotme:
Tested it and it doesn't seem to fix the problem. Motion events are still sent to unrelated clients after DND is "performed" or cancelled.

fredldotme:
You are right, this indeed *does* fix the issue. Seems my initial install pulled debs from the lomiri archive due to `--reinstall` in the apt command. I'd say go for it!

AlanGriffiths:
> I'd say go for it!

Actually, this iteration is inelegant: I'm just taking the feedback as proof I understand what is going wrong.

My main concern is that it relies on `miroil::dispatch_input_event()` being called for the button release. I want to look for a less fragile approach.

AlanGriffiths:
OK, so this does fix part of the problem: after the release "normal service" is resumed.

However, with my testing hack mentioned above, there's not enough integration of the input logic for "drop" to work. Is that working with Lomiri?

fredldotme:
> OK, so this does fix part of the problem: after the release "normal service" is resumed.
> 
> 
> 
> However, with my testing hack mentioned above, there's not enough integration of the input logic for "drop" to work. Is that working with Lomiri?

I just applied this PR as a .patch and dropping in the same window indeed seems to function as tested with Nautilus. Can't test multi-window drag and drop as something causes Lomiri to crash in this mixed PPA+custom deb build.

AlanGriffiths:
@fredldotme @sunweaver I've pushed one change (to preserve ABI) that needs retesting (I don't expect it to break things).

@Saviq If this (still) fixes things, then we can land and cherry-pick to a 2.20.2 release

sunweaver:
@AlanGriffiths I also tested this MR (before ABI compliance change):

  * it fixes dragging a tab from Firefox and dropping it as individual browser window
  * dragging one Firefox tab to another Firefox window and having it added as a tab to that window does not work
  * dragging tabs from one chromium window to another chromium window works
  * I failed to test d'n'd'ing files/folders between Nautilus windows, but @fredldotme provided some info on that

I can re-test the recent change on this branch tomorrow (maybe this evening).

fredldotme:
@sunweaver was that with chromium and ˋ--ozone-platform waylandˋ? That one also fixes 3 dot menu not showing.

fredldotme:
Can confirm it working on Wayland with Chromium, even supports dragging tab back into another window.

src/miroil/eventdispatch.cpp
```diff
@@ -24,7 +27,38 @@
...
+            for_pressed_buttons(pev, [&any_pressed](MirPointerButton){ any_pressed = true; });
+            return !any_pressed && mir_pointer_event_action(pev) == mir_pointer_action_button_up;
+        }
+    }
+
+    return false;
+}
+}
+
+void miroil::dispatch_input_event(const miral::Window& window, MirInputEvent const* event)
```
Saviq, line 61: Should move the `const` for Window, too?

src/include/server/mir/default_server_configuration.h
```diff
@@ -385,6 +385,8 @@ class DefaultServerConfiguration : public virtual ServerConfiguration
     CachedPtr<input::CompositeEventFilter> composite_event_filter;
     CachedPtr<input::InputManager>    input_manager;
     CachedPtr<input::SurfaceInputDispatcher>    surface_input_dispatcher;
+    static inline // Hack to preserve ABI
```
Saviq, line 388: Should we land into `main` without the hack, and only backport to 2.20.2, with?

src/miroil/gesture_ender.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include <memory>
+
+namespace mir::frontend { class PointerInputDispatcher; }
+
+namespace miroil
+{
+void end_gesture();
+
+auto the_pointer_input_dispatcher() -> std::shared_ptr<mir::frontend::PointerInputDispatcher>;
```
Saviq, line 28: Should we rename this to avoid confusion with `mir::DefaultServerConfiguration`?

I mean, "gesture_ender_input_dispatcher()" or such?

src/miroil/gesture_ender.cpp
```diff
@@ -0,0 +1,68 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "gesture_ender.h"
+
+#include <mir/frontend/pointer_input_dispatcher.h>
+
+namespace
+{
+struct PointerInputDispatcher : mir::frontend::PointerInputDispatcher
```
Saviq, line 23:
```suggestion
struct GestureEnder : mir::frontend::PointerInputDispatcher
```

Or something along those lines?

Saviq (changes requested):
You're sneaking in some unrelated changes, but OK :P

Inline nits.

src/include/server/mir/default_server_configuration.h
Saviq: Is our symbol checker not smart enough?

```
Internal Symbols Diff:
  New Symbols 🟢🟢🟢
    mir::DefaultServerConfiguration::pointer_input_dispatcher*;
    mir::Server::the_pointer_input_dispatcher*;
```

src/miroil/gesture_ender.h
AlanGriffiths: I think that would cause different confusion

CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
 set(MIR_VERSION_MAJOR 2)
 set(MIR_VERSION_MINOR 20)
-set(MIR_VERSION_PATCH 0)
+set(MIR_VERSION_PATCH 2)
```
Saviq, line 29: This should now be `2.21.0`?

src/server/symbols.map
```diff
@@ -1458,3 +1458,9 @@ global:
   };
 } MIR_SERVER_INTERNAL_2.19;
 
+MIR_SERVER_INTERNAL_2.20.2 {
+global:
+  extern "C++" {
+    mir::Server::override_the_pointer_input_dispatcher*;
+  };
+} MIR_SERVER_INTERNAL_2.20;
```
Saviq, line 1466: And this, need a 2.21 stanza.

Saviq (changes requested):
Need symbols fixed up now.

CMakeLists.txt
AlanGriffiths: That depends on whether we're rolling a 2.20.2?

src/server/symbols.map
AlanGriffiths: That depends on whether we're rolling a 2.20.2?

AlanGriffiths:
Switching to "Draft" #3875 reworks this for a bugfix release

PR #3847, opened 2025-03-21
Fix `Mouse::{v,h}scroll_speed` returning `acceleration_bias`

tarek-y-ismail:
Caught by tests in #3830


AlanGriffiths (approved):
I trust the tests will land separately

PR #3852, opened 2025-03-25
Deprecate "persistent" surface IDs

AlanGriffiths:
IMO the `PersistentSurfaceStore` and associated infrastructure are zombie code: Mark the corresponding APIs as deprecated before removal


mattkae (approved):
Makes sense, I always just glossed over that anyway

PR #3855, opened 2025-03-25
tests: drop `MIR_COMPILER_QUIRKS`

Saviq:
(https://github.com/canonical/mir/pull/3842) disabled `array-bounds` errors in optimized builds, that was the
only remaining quirk.


AlanGriffiths (approved):
Great!

PR #3856, opened 2025-03-25
debian: tighten LTO-disabling rules

Saviq:
Only see those issues in plucky,oracular/arm64 at the moment.


Saviq:
Test packages with this quirk removed completely uploaded to:

https://launchpad.net/~mir-team/+archive/ubuntu/test/+packages

Only arm64/plucky showed the FTBFS due to toolchain issues.

AlanGriffiths (approved):
OK, let's see if this is stable

PR #3859, opened 2025-03-26
WIP: gbm-kms: do not pass invalid mode index as current mode

bhush9:
If preferred mode exists then pass it as a current mode if current mode is not set. This fixes the coredump in lomiri when one tries to connect HDMI display, because of invalid current mode index.

### DIFF
--- a/src/platforms/gbm-kms/server/kms/real_kms_output.cpp
+++ b/src/platforms/gbm-kms/server/kms/real_kms_output.cpp
@@ -593,7 +593,10 @@ void mgg::RealKMSOutput::update_from_hardware_state(
     // There's no need to warn about failing to find a current display mode on a disconnected display.
     if (connected && (current_mode_index == invalid_mode_index)) {
         mir::log_warning(
-            "Unable to determine the current display mode.");
+            "Unable to determine the current display mode, setting to preferred mode");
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
> If preferred mode exists then pass it as a current mode if current mode is not set. This fixes the coredump in lomiri when one tries to connect HDMI display, because of invalid current mode index.

This sounds wrong on principle: it is lying about the state reported by the hardware

In practice, shouldn't the hardware be told which mode is being applied? (Which may not happen if the software thinks there's already a current mode)

bhush9:
> In practice, shouldn't the hardware be told which mode is being applied? (Which may not happen if the software thinks there's already a current mode)

Hm, you are right now that I see it, somehow `current_crtc` is not valid and we are basically then comparing modes returned from driver with "empty" mode. I will debug this bit more.

bhush9:
Okay, I did bit more reading of code, and I feel current code on side of the mir is okay, though not sure if code on lomiri needs to be adapted?

https://gitlab.com/ubports/development/core/qtmir/-/blob/personal/sunweaver/debian-upstream/src/platforms/mirserver/platformscreen.cpp#L316

Can display->config() include invalid mode id as a current mode id?

Issue #3854, opened 2025-03-25
`TestInputDeviceHub.*` unstable


Saviq:
E.g.

https://launchpadlibrarian.net/784124224/buildlog_ubuntu-plucky-ppc64el.mir_2.20.1+dev101-g362ad1d51c-0ubuntu25.04_BUILDING.txt.gz
https://launchpadlibrarian.net/784123706/buildlog_ubuntu-oracular-ppc64el.mir_2.20.1+dev101-g362ad1d51c-0ubuntu24.10_BUILDING.txt.gz

Suggests it's `ppc64el`-specific :'-(

```
[ RUN ] TestInputDeviceHub.notifies_input_device_observer_about_available_devices
# ...
unknown file: Failure

Unexpected mock function call - returning directly.
Function call: changes_complete()
Google Mock tried the following 2 expectations, but none matched:

./tests/acceptance-tests/test_input_device_hub.cpp:72: tried expectation #0: EXPECT_CALL(observer, changes_complete())...
Expected: all pre-requisites are satisfied
Actual: the following immediate pre-requisites are not satisfied:
./tests/acceptance-tests/test_input_device_hub.cpp:70: pre-requisite #0
(end of pre-requisites)
Expected: to be called once
Actual: never called - unsatisfied and active
./tests/acceptance-tests/test_input_device_hub.cpp:76: tried expectation #1: EXPECT_CALL(observer, changes_complete())...
Expected: all pre-requisites are satisfied
Actual: the following immediate pre-requisites are not satisfied:
./tests/acceptance-tests/test_input_device_hub.cpp:75: pre-requisite #0
(end of pre-requisites)
Expected: to be called once
Actual: never called - unsatisfied and active

[2025-03-25 06:40:42.772932] <information> input-hub: Device configuration: mousekey-pointer, capabilities={pointer}
[2025-03-25 06:40:42.776677] <information> input-hub: Device configuration: keyboard, capabilities={keyboard}
[2025-03-25 06:40:42.776735] <information> mirserver: Stopping
[ FAILED ] TestInputDeviceHub.notifies_input_device_observer_about_available_devices (24 ms)
```

Saviq:
Seems to be important to look at the order of things here:
```
[2025-03-26 12:57:37.774697] <information> input-hub: Device configuration: mousekey-pointer, capabilities={pointer}
unknown file: Failure

Unexpected mock function call - returning directly.
[...]

PR #3861, opened 2025-03-27
Lazily register input event transformer virtual pointer

tarek-y-ismail:
Works around issues in #3854

### DIFF
--- a/src/include/server/mir/input/input_event_transformer.h
+++ b/src/include/server/mir/input/input_event_transformer.h
@@ -52,10 +52,7 @@ public:
         virtual bool transform_input_event(EventDispatcher const&, EventBuilder*,  MirEvent const&) = 0;
     };
 
-    InputEventTransformer(
-        std::shared_ptr<InputDeviceRegistry> const&,
-        std::shared_ptr<MainLoop> const&,
-        std::shared_ptr<VirtualInputDevice> const&);
+    InputEventTransformer(std::shared_ptr<MainLoop> const&);
     ~InputEventTransformer();
 
     bool handle(MirEvent const&) override;
@@ -63,12 +60,12 @@ public:
     void append(std::weak_ptr<Transformer> const&);
     bool remove(std::shared_ptr<Transformer> const&);
 
+    void virtual_device(std::shared_ptr<input::VirtualInputDevice> const&);
 private:
     std::mutex mutex;
     std::vector<std::weak_ptr<Transformer>> input_transformers;
+    std::shared_ptr<input::VirtualInputDevice> virtual_pointer;
 
-    std::shared_ptr<input::VirtualInputDevice> const virtual_pointer;
-    std::shared_ptr<input::InputDeviceRegistry> const input_device_registry;
     std::shared_ptr<MainLoop> const main_loop;
 };
 }
--- a/src/include/server/mir/input/input_event_transformer.h
+++ b/src/include/server/mir/input/input_event_transformer.h
@@ -52,7 +52,10 @@ public:
         virtual bool transform_input_event(EventDispatcher const&, EventBuilder*,  MirEvent const&) = 0;
     };
 
-    InputEventTransformer(std::shared_ptr<InputDeviceRegistry> const&, std::shared_ptr<MainLoop> const&);
+    InputEventTransformer(
+        std::shared_ptr<InputDeviceRegistry> const&,
+        std::shared_ptr<MainLoop> const&,
+        std::shared_ptr<VirtualInputDevice> const&);
     ~InputEventTransformer();
 
     bool handle(MirEvent const&) override;
@@ -67,8 +70,6 @@ private:
     std::shared_ptr<input::VirtualInputDevice> const virtual_pointer;
     std::shared_ptr<input::InputDeviceRegistry> const input_device_registry;
     std::shared_ptr<MainLoop> const main_loop;
-
-    bool virtual_pointer_registered{false};
 };
 }
 }
--- a/src/include/server/mir/input/input_event_transformer.h
+++ b/src/include/server/mir/input/input_event_transformer.h
@@ -67,6 +67,8 @@ private:
     std::shared_ptr<input::VirtualInputDevice> const virtual_pointer;
     std::shared_ptr<input::InputDeviceRegistry> const input_device_registry;
     std::shared_ptr<MainLoop> const main_loop;
+
+    bool virtual_pointer_registered{false};
 };
 }
 }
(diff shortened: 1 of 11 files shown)


AlanGriffiths:
I guess I have a meta question: Why is `InputEventTransformer` unconditionally creating any `VirtualInputDevice`?

It has `append()` and `remove()` methods which suggest that the responsibility lies elsewhere.

tarek-y-ismail:
We discussed this in https://github.com/canonical/mir/pull/3754#discussion_r1972405011. It was my idea to register when the constructor of `InputEventTransformer` was called. Didn't cause much trouble in tests after a couple of quick fixes.

AlanGriffiths:
> We discussed this in [#3754 (comment)](https://github.com/canonical/mir/pull/3754#discussion_r1972405011)

That discussion is about _when_ `InputEventTransformer` creates the `VirtualInputDevice`. My question is _why_?

tarek-y-ismail:
> My question is why?

No particular reason. Didn't want to complicate the code, and `VirtualInputDevice`s are quite lightweight and basically do nothing on construction.

tarek-y-ismail:
> You are complicating the code (see this PR)

How so?

> Well, that's also wrong: there's no need to create the object when it isn't append()ed and there's no code to remove() it. (Having append()/remove() is a leak prone design)

Thanks for catching that. I'll make a PR to fix that and we can discuss alternatives to `append`/`remove` there.

AlanGriffiths (changes requested):
> > You are complicating the code (see this PR)
> 
> How so?

Without automatically creating , `InputEventTransformer` is a familiar multiplexer design. With it has the additional responsibility of owning, registering and removing virtual_pointer.

Compare this code with the code for registering `MouseKeysTransformer` where `AccessibilityManager` takes responsibility for the lifecycle and `InputEventTransformer` for the multiplexing

AlanGriffiths:
> we can discuss alternatives to `append`/`remove` there.

Here's one alternative, return a class responsible for the deregistration from `append()` and make `remove()` private:

```c++
class InputEventTransformer
...
    struct Registration
    {
        Registration() : unregister([](){}){}

        Registration(
            InputEventTransformer* event_transformer,
            std::shared_ptr<mir::input::InputEventTransformer::Transformer> transformer) :
            unregister([event_transformer, transformer]() { event_transformer->remove(transformer); })
        {
        }

        ~Registration()
        {
            unregister();
        }
        std::function<void()> const unregister;

        Registration& operator=(Registration const&) = delete;
        Registration(Registration const&) = delete;
    };

    auto append(std::weak_ptr<Transformer> const&) -> Registration;
```

tarek-y-ismail:
> Here's one alternative

Split it off into #3871 since it didn't seem immediately related to this PR.

AlanGriffiths (changes requested):
> [tarek-y-ismail](https://github.com/tarek-y-ismail) requested a review from [AlanGriffiths](https://github.com/AlanGriffiths) [April 9, 2025 09:23](https://github.com/canonical/mir/pull/3861#event-17184393715)

This is still the wrong object to be allocating `VirtualInputDevices`

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -52,20 +52,20 @@ class InputEventTransformer : public EventFilter
...
-    InputEventTransformer(std::shared_ptr<InputDeviceRegistry> const&, std::shared_ptr<MainLoop> const&);
+    InputEventTransformer(std::shared_ptr<MainLoop> const&);
     ~InputEventTransformer();
 
     bool handle(MirEvent const&) override;
 
     void append(std::weak_ptr<Transformer> const&);
     bool remove(std::shared_ptr<Transformer> const&);
 
+    void virtual_device(std::shared_ptr<input::VirtualInputDevice> const&);
```
AlanGriffiths, line 63: Why does the `InputEventTransformer` need to take ownership of a ` VirtualInputDevice`? Isn't it enough that the devices is appended as a Transformer?

tarek-y-ismail: What do you mean? The device is not appended as a transformer. Transformers depend on a virtual device to be able to dispatch events as necessary. 

As for ownership/lifetime, I'm not sure if a `{weak,shared}_ptr` make any difference. At the moment, the virtual device is allocated once when `BasicAccessibilityManager` is created, then its registered or removed as transformers are enabled or disabled. So either _can_ work, even a raw pointer will work since it always exists after the initial allocation.

AlanGriffiths (changes requested):
OK, some of this  is pre-existing but this seems to be working around `InputEventTransformer` performing multiple roles:

1. It is a register of input transformers
2. It intercepts events and passes them to `virtual_pointer`
3. It passes a functor to `virtual_pointer` that dispatches the events _and_ iterates over the transformers until one processes the event

That makes its function somewhat unclear and hard to use correctly.

A better division would seem to be:

`InputEventTransformer`

1. It is a register of input transformers
2. iterates over the transformers until one processes the event

`VirtualPointer`

1. It intercepts events and...
2. dispatches the events and...
3. passes them to `InputEventTransformer`

That reverses the _HasA_ relationship between `InputEventTransformer` and `VirtualPointer`, but seems clearer

AlanGriffiths:
> * It passes a functor to `virtual_pointer` that dispatches the events _and_ iterates over the transformers until one processes the event

Correcting myself: there's no "and" here, it is the transformed event that might be dispacthed

AlanGriffiths:
I think the missing piece is a `VirtualPointer` class that extends `VirtualInputDevice` and implements `EventFilter`:

```c++
class VirtualPointer : public VirtualInputDevice, public EventFilter 
{
public:
    VirtualPointer(shared_ptr<InputEventTransformer> iet);
    bool handle(MirEvent const& event) override
    {
        auto handled = false;
        if_started_then(
            [this, &event, &handled](auto* sink, auto* builder)
            {
                auto const dispatcher = [main_loop = this->main_loop, sink](auto event)
                {
                    main_loop->spawn([sink, event]
                        {
                             sink->handle_input(event);
                        });
                };
                handled = iet->transform(event, builder, dispatcher);
    }
};

...

void mir::shell::BasicAccessibilityManager::mousekeys_enabled(bool on)
{
    if (on)
    {
        virtual_input_device = std::make_shared<VirtualPointer>(event_transformer);
        input_device_registry->add_device(virtual_input_device);
    }
    else
    {
        event_transformer->remove(transformer);
        input_device_registry->remove_device(virtual_input_device);
    }
}

AlanGriffiths:
Closing as we're landing an alternative: https://github.com/canonical/mir/pull/3898

PR #3862, opened 2025-03-27
Only initialize `AccessibilityManager::transformer` if mousekeys are enabled

tarek-y-ismail:
Split out from https://github.com/canonical/mir/pull/3861#issuecomment-2758295300

### DIFF
--- a/src/include/server/mir/shell/accessibility_manager.h
+++ b/src/include/server/mir/shell/accessibility_manager.h
@@ -68,7 +68,7 @@ private:
     bool const enable_mouse_keys;
 
     std::shared_ptr<mir::input::InputEventTransformer> const event_transformer;
-    std::shared_ptr<mir::input::InputEventTransformer::Transformer> transformer;
+    std::shared_ptr<mir::input::InputEventTransformer::Transformer> const transformer;
 };
 }
 }
--- a/src/include/server/mir/shell/accessibility_manager.h
+++ b/src/include/server/mir/shell/accessibility_manager.h
@@ -65,7 +65,6 @@ private:
     Synchronised<MutableState> mutable_state;
 
     bool const enable_key_repeat;
-    bool const enable_mouse_keys;
 
     std::shared_ptr<mir::input::InputEventTransformer> const event_transformer;
     std::shared_ptr<mir::input::InputEventTransformer::Transformer> const transformer;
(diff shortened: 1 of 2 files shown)


src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -66,7 +68,7 @@ class AccessibilityManager
     bool const enable_mouse_keys;
 
     std::shared_ptr<mir::input::InputEventTransformer> const event_transformer;
-    std::shared_ptr<mir::input::InputEventTransformer::Transformer> const transformer;
+    std::shared_ptr<mir::input::InputEventTransformer::Transformer> transformer;
```
AlanGriffiths, line 71: There's no reason to remove `const`: just move the logic into the initializer

tarek-y-ismail: @RAOF You were right!

src/server/shell/accessibility_manager.cpp
```diff
@@ -118,8 +118,24 @@ mir::shell::AccessibilityManager::AccessibilityManager(
...
+                        event_transformer->append(transformer);
+                        return t;
+                    }
+
+                    return std::shared_ptr<MouseKeysTransformer>{};
+                }()
+
+    }
+{
+}
```
AlanGriffiths, line 135: I don't think `enable_mouse_keys` is needed anymore

```suggestion
    transformer{[&]
                {
                    return options->get<bool>(options::enable_mouse_keys_opt) ?
                        std::make_shared<MouseKeysTransformer>() :
                        std::shared_ptr<MouseKeysTransformer>{};
                }()
    }
{
    if(transformer)
        event_transformer->append(transformer);
}
```

src/server/shell/accessibility_manager.cpp
```diff
@@ -116,10 +116,22 @@ mir::shell::AccessibilityManager::AccessibilityManager(
...
+    }
 {
-    if (enable_mouse_keys)
+    if (transformer)
         event_transformer->append(transformer);
 }
+
+mir::shell::AccessibilityManager::~AccessibilityManager()
+{
+    if(transformer)
```
AlanGriffiths, line 135:
```suggestion
    if (transformer)
```

AlanGriffiths (approved):
Looks sensible. Thanks

PR #3863, opened 2025-03-27
tics: run on all `main` pushes and gate on pull requests

Saviq:
Gating up for evaluation - not a required check just yet.


mattkae (approved):
Looks good to me!

PR #3866, opened 2025-03-28
refactor: mouse keys transformer tests no longer rely on timings from other threads and alarms

mattkae:


tarek-y-ismail (approved):
Black magic. Thanks!

PR #3868, opened 2025-04-01
ci: rename symbols check job

Saviq:
Otherwise can't make it required.


Saviq:
@mattkae symbols check is now Required here, meaning it won't let things through with broken symbols.

mattkae (approved):

PR #3869, opened 2025-04-01
feature: add the ability to associate a layout with custom user data

mattkae:
## What's new?
- Add `DisplayConfiguration::layout_userdata` and `DisplayConfiguration::set_layout_userdata_builder` so that callers can get/set an arbitrary `shared_ptr<void>` user-defined data member on a per layout basis.
- Bumped the minor version of miral from 5.2 to 5.3
- Wrote a test for setting/getting the userdata for a layout
- Added a new stanza to the `symbols.map` file

### DIFF
--- a/include/miral/miral/display_configuration.h
+++ b/include/miral/miral/display_configuration.h
@@ -20,6 +20,9 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <functional>
+#include <optional>
+#include <any>
 
 namespace mir { class Server; }
 
@@ -41,6 +44,43 @@ class ConfigurationOption;
 class DisplayConfiguration
 {
 public:
+    /// A class providing access to an arbitrary piece of data from the [DisplayConfiguration].
+    /// This is specifically useful when a user wants to extend the based display configuration
+    /// with some sort of custom payload (e.g. a user may want to extend the layout configuration
+    /// with information describing the position and size of specific applications).
+    /// \remark Since MirAL 5.3
+    class Node
+    {
+    public:
+        enum class Type
+        {
+            integer,
+            string,
+            sequence,
+            map,
+            unknown
+        };
+
+        auto type() const -> Type;
+        auto as_string() const -> std::string;
+        auto as_int() const -> int;
+        void for_each(std::function<void(Node const&)> const& f) const;
+        auto has(std::string const& key) const -> bool;
+        auto at(std::string const& key) const -> std::optional<Node>;
+        Node(Node&&) noexcept = default;
+        Node& operator=(Node&&) noexcept = default;
+        Node(Node const&) = delete;
+        Node& operator=(Node const&) = delete;
+        ~Node();
+
+    private:
+        friend DisplayConfiguration;
+        struct Self;
+        explicit Node(std::unique_ptr<Self>&& self);
+
+        std::unique_ptr<Self> self;
+    };
+
     explicit DisplayConfiguration(MirRunner const& mir_runner);
 
     /// Provide the default 'display-layout' configuration option
@@ -55,6 +95,24 @@ public:
     /// Enable a custom output attribute in the .display YAML
     /// \remark Since MirAL 3.8
     void add_output_attribute(std::string const& key);
+
+    /// Retrieve the user data associated with the active layout
+    /// for this provided key. Callers should provide this user data
+    /// via [layout_userdata_builder].
+    /// \remark Since MirAL 5.3
+    auto layout_userdata(std::string const& key) const -> std::optional<std::any const>;
+
+    /// Enable a custom layout attribute in the .display YAML.
+    /// The caller must provides the key of this custom attribute in
+    /// addition to a function that will be used to build the custom
+    /// payload that is associated with the layer. This function is
+    /// provided with the raw details of the node at the provided key.
+    /// The function must return any piece of data that may later be
+    /// retrieved via [layout_userdata].
+    /// \remark Since MirAL 5.3
+    void layout_userdata_builder(
+        std::string const& key,
+        std::function<std::any(Node const& value)> const& builder) const;
 
     void operator()(mir::Server& server) const;
 
--- a/src/miral/display_configuration.cpp
+++ b/src/miral/display_configuration.cpp
@@ -21,12 +21,33 @@
 
 #include <mir/server.h>
 
+#include <yaml-cpp/yaml.h>
 #include <boost/throw_exception.hpp>
 
 #include <fstream>
 #include <sstream>
 
 using namespace std::string_literals;
+
+namespace
+{
+const char* node_type_to_string(miral::DisplayConfiguration::Node::Type type)
+{
+    switch (type)
+    {
+        case miral::DisplayConfiguration::Node::Type::integer:
+            return "integer";
+        case miral::DisplayConfiguration::Node::Type::string:
+            return "string";
+        case miral::DisplayConfiguration::Node::Type::sequence:
+            return "sequence";
+        case miral::DisplayConfiguration::Node::Type::map:
+            return "map";
+        default:
+            return "unknown";
+    }
+}
+}
 
 class miral::DisplayConfiguration::Self : public ReloadingYamlFileDisplayConfig
 {
@@ -118,6 +139,120 @@ auto miral::DisplayConfiguration::list_layouts() -> std::vector<std::string>
 void miral::DisplayConfiguration::add_output_attribute(std::string const& key)
 {
     self->add_output_attribute(key);
+}
+
+auto miral::DisplayConfiguration::layout_userdata(std::string const& key) const
+    -> std::optional<std::any const>
+{
+    return self->layout_userdata(key);
+}
+
+struct miral::DisplayConfiguration::Node::Self
+{
+    explicit Self(YAML::Node const& node)
+        : node(node)
+    {
+    }
+
+    ~Self() = default;
+
+    template <typename T>
+    auto as() const -> T
+    {
+        try
+        {
+            return node.as<T>();
+        }
+        catch (YAML::Exception const&)
+        {
+            return T();
+        }
+    }
+
+    YAML::Node const node;
+};
+
+miral::DisplayConfiguration::Node::Node(std::unique_ptr<Self>&& self)
+    : self(std::move(self))
+{
+}
+
+miral::DisplayConfiguration::Node::~Node() = default;
+
+auto miral::DisplayConfiguration::Node::type() const -> Type
+{
+    if (self->node.IsScalar())
+    {
+        try {
+            (void)self->node.as<int>();
+            return Type::integer;
+        } catch (const YAML::BadConversion& e) {
+            return Type::string;
+        }
+    }
+    else if (self->node.IsMap())
+        return Type::map;
+    else if (self->node.IsSequence())
+        return Type::sequence;
+    else
+    {
+        mir::log_warning("Parsing custom display configuration node: Checked type is none of the supported types");
+        return Type::unknown;
+    }
+}
+
+auto miral::DisplayConfiguration::Node::as_string() const -> std::string
+{
+    if (type() != Type::string)
+        mir::fatal_error("Attempting to access a Node of type %s as a string", node_type_to_string(type()));
+
+    return self->as<std::string>();
+}
+
+auto miral::DisplayConfiguration::Node::as_int() const -> int
+{
+    if (type() != Type::integer)
+        mir::fatal_error("Attempting to access a Node of type %s as an integer", node_type_to_string(type()));
+
+    return self->as<int>();
+}
+
+void miral::DisplayConfiguration::Node::for_each(std::function<void(Node const&)> const& f) const
+{
+    if (type() != Type::sequence)
+        mir::fatal_error("Attempting to access a Node of type %s as a sequence", node_type_to_string(type()));
+
+    for (auto const& item : self->node)
+        f(Node(std::make_unique<Self>(item)));
+}
+
+auto miral::DisplayConfiguration::Node::has(std::string const& key) const -> bool
+{
+    if (type() != Type::map)
+        mir::fatal_error("Attempting to access a Node of type %s as a map", node_type_to_string(type()));
+
+    if (self->node[key])
+        return true;
+
+    return false;
+}
+
+auto miral::DisplayConfiguration::Node::at(std::string const& key) const -> std::optional<Node>
+{
+    if (!has(key))
+        return std::nullopt;
+
+    return Node(std::make_unique<Self>(self->node[key]));
+}
+
+void miral::DisplayConfiguration::layout_userdata_builder(
+    std::string const& key,
+    std::function<std::any(Node const& value)> const& builder) const
+{
+    self->layout_userdata_builder(key, [builder=builder](YAML::Node const& data)
+    {
+        return builder(Node(std::make_unique<Node::Self>(data)));
+    });
 }
 
 miral::DisplayConfiguration::~DisplayConfiguration() = default;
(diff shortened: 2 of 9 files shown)


src/miral/symbols.map
```diff
@@ -499,10 +499,10 @@ global:
     typeinfo?for?miral::ConfigFile;
     typeinfo?for?miral::Decorations;
     typeinfo?for?miral::IdleListener;
+    typeinfo?for?miral::IdleListener;
     typeinfo?for?miral::InputConfiguration::Mouse;
     typeinfo?for?miral::InputConfiguration::Touchpad;
     typeinfo?for?miral::InputConfiguration;
-    typeinfo?for?miral::IdleListener;
```
robert-ancell, line 505: Nit: Is this move intentional?

include/miral/miral/display_configuration.h
```diff
@@ -52,10 +54,23 @@ class DisplayConfiguration
     /// List all layouts found in the config file
     auto list_layouts() -> std::vector<std::string>;
 
+    /// Retrieve the user data of the active layout.
+    /// \remark Since MirAL 5.3
+    auto layout_userdata() const -> std::shared_ptr<void>;
```
robert-ancell, line 59: Because the custom configuration is per layout and the feature this is being added for the configuration is per card this means this object needs to contain a duplicate mapping for that. Ideally we would be extending the card configuration, however it doesn't seem that is exposed in a class easily. Not sure if there is a solution, but noting it here.

include/miral/miral/display_configuration.h
```diff
@@ -20,7 +20,9 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <functional>
 
+namespace YAML { class Node; }
```
robert-ancell, line 25: We discussed elsewhere if it was a good idea to expose the YAML objects rather than an interface, however I think looking at the code this is probably reasonable. The class is clearly defined as coming from YAML and it doesn't seem to be designed to be sourced from other locations.

robert-ancell (approved):
The code changes make sense. It does feel a little awkward to use, but I think that's due to the structure of the classes and it doesn't seem like there's a good opportunity to make it more flexible. I think it will meet the purposes it is there for, and is sufficiently generic for any future changes necessary. It would be nice if we could add some helper code to make managing per-card config a little easier.

include/miral/miral/display_configuration.h
AlanGriffiths: The summary says "callers can get/set an arbitrary `shared_ptr<void>`", so why is there a need for YAML in the API?

include/miral/miral/display_configuration.h
AlanGriffiths: The summary says "callers can get/set an arbitrary `shared_ptr<void>`", so why is there a need for YAML in the API?

mattkae: The need for YAML is so that users can set arbitrary userdata _based off of_ the YAML that defines the layout. We need to be able to load a custom data format from the yaml here.

include/miral/miral/display_configuration.h
mattkae: Is the feature per card? I thought that the spec wants a new `applications` list per-layout in frame. The `snap-name` and `surface-title` per-card is already preexisting it seems

mattkae:
> The code changes make sense. It does feel a little awkward to use, but I think that's due to the structure of the classes and it doesn't seem like there's a good opportunity to make it more flexible. I think it will meet the purposes it is there for, and is sufficiently generic for any future changes necessary. It would be nice if we could add some helper code to make managing per-card config a little easier.

I think the trouble that I have with this design is that it necessitates consumers using the same version of yaml-cpp, which is awkward. I think that an interface class might be a better idea here.

src/miral/symbols.map
mattkae: We use a tool to automate this, and some times it just reorders things :shrug:

AlanGriffiths: FWIW I think it the locale used by sort differs between team members

tests/miral/static_display_config.cpp
```diff
@@ -539,3 +540,44 @@ TEST_F(StaticDisplayConfig, given_no_custom_attributes_when_they_are_added_exist
...
+    {
+        std::string value;
+    };
+
+    std::istringstream stream{
+        "layouts:\n"
+        "  another:\n"
+        "    cards:\n"
+        "    - VGA-1:\n"
+        "        state: disabled\n"
```
robert-ancell, line 556: Referring to a previous comment - could you put a custom field under VGA-1 to confirm that not just custom toplevel layout fields work?

Saviq: That is preexisting:

https://github.com/canonical/mir/blob/880f253b4dd9ca85ea26f45ca6c1e27ddad35c26/tests/miral/static_display_config.cpp#L455-L541

src/miral/symbols.map
RAOF: But also, this is a duplicate? Why is it adding a second `typeinfo?for?miral::IdleListener;`?

RAOF (commented):
A couple of nits, and a larger question:
Should `layout_userdata_builder` only see the tree under a `custom_attrib:` node (if it exists), rather than build out of the whole `layout` tree?

eg:
```yaml
layouts:
  one:
    cards:
    - VGA-1:
      state: disabled
      not_seen_by_custom_builder: "this thing"
    custom_attribs:
      - this_is_seen_by_custom_builder:
        "this thing"
```

The reason this might be a good idea is to ensure that further iterations to the display configuration in Mir don't accidentally run into custom YAML properties from shells.

mattkae:
> A couple of nits, and a larger question: Should `layout_userdata_builder` only see the tree under a `custom_attrib:` node (if it exists), rather than build out of the whole `layout` tree?
> 
> eg:
> 
> ```yaml
> layouts:
>   one:
>     cards:
>     - VGA-1:
>       state: disabled
>       not_seen_by_custom_builder: "this thing"
>     custom_attribs:
>       - this_is_seen_by_custom_builder:
>         "this thing"
> ```
> 
> The reason this might be a good idea is to ensure that further iterations to the display configuration in Mir don't accidentally run into custom YAML properties from shells.

Yeah that's not a bad idea, but it does run contrary to the spec that we've decided on. @Saviq and @robert-ancell, do you have any preference?

The argument against this is that users editing a display configuration don't care if the attribute is custom or not, so it would be a bit odd for them to write `custom_attribs` there. The argument in favor of this is to save ourselves from colliding keys.

I am of the mindset that risking colliding keys isn't too likely here.

mattkae:
> Shells should be able to register a per-layout key, and they only get that subtree, from the active layout, rather than "parsing" the whole doc themselves?
> 
> Similar to how they can register a custom per-output attribute now:
> 
> https://github.com/canonical/mir/blob/880f253b4dd9ca85ea26f45ca6c1e27ddad35c26/include/miral/miral/display_configuration.h#L55-L57

Yeah that makes sense. And the implementation will associate user data with each individual key per layout, rather than just once per layout.

src/miral/symbols.map
mattkae: Good question! I made a bug for this: https://github.com/canonical/mir/issues/3876

include/miral/miral/display_configuration.h
```diff
@@ -28,6 +31,21 @@ namespace miral
...
+/// \remark Since MirAL 5.3
+class DisplayConfigurationNode
+{
+public:
+    virtual ~DisplayConfigurationNode() = default;
+    virtual void for_each(std::function<void(std::unique_ptr<DisplayConfigurationNode>)> const& f) const = 0;
+    virtual auto as_string() const -> std::optional<std::string> = 0;
+    virtual auto as_int() const -> std::optional<int> = 0;
+    virtual auto at(std::string const& key) const -> std::optional<std::unique_ptr<DisplayConfigurationNode>> = 0;
+};
```
AlanGriffiths, line 47: I am unclear what this class represents. Assuming it represents a node in the parsed data, then:

1. Why is it an interface (not a pimpl)? 
2. Why is there no way to tell the type? (the accessors appear to correspond to sequence, string, integer and dictionary)
3. Passing ownership of the nodes (e.g. via `for_each`) seems unusual when, conceptually at least, the nodes visited are not "owned". Could we just pass a `DisplayConfigurationNode const&`?
4. Would it be better named as `DisplayConfiguration::Node`?

Or, have I guess wrong? Is the API user expected to implement this interface and use it Visitor style?

AlanGriffiths (changes requested):
There are no tests of the miral API using `DisplayConfigurationNode` & `DisplayConfiguration`

include/miral/miral/display_configuration.h
mattkae: That is all very fair :+1:

mattkae:
> # TICS Quality Gate
> ### ❌ Failed
> ## The following errors have occurred during analysis:
> ❌ [ERROR 5383] When using the TICS client to run a code coverage metric, the option '-resultdir' is mandatory.
> TICS / TICS / Run TICS analysis

@Saviq  :eyes:

include/miral/miral/display_configuration.h
```diff
@@ -41,6 +45,25 @@ class ConfigurationOption;
...
 public:
+    /// A class providing access to an arbitrary piece of data from the [DisplayConfiguration].
+    /// This is specifically useful when a user wants to extend the based display configuration
+    /// with some sort of custom payload (e.g. a user may want to extend the layout configuration
+    /// with information describing the position and size of specific applications).
+    /// \remark Since MirAL 5.3
+    class Node
+    {
+    public:
+        explicit Node(YAML::Node const& node);
```
AlanGriffiths, line 56: 1. This is the only use of `YAML::Node` in the API. That smells like an implementation detail being leaked.
2. We don't expect users of the API to create these objects

So can we make the constructor private and make the parameter something we own (e.g. `std::unique_ptr<Self>`)?

include/miral/miral/display_configuration.h
```diff
@@ -41,6 +45,25 @@ class ConfigurationOption;
...
+        auto as_string() const -> std::optional<std::string>;
+        auto as_int() const -> std::optional<int>;
+        void for_each(std::function<void(Node const&)> const& f) const;
+        auto at(std::string const& key) const -> std::optional<Node>;
```
AlanGriffiths, line 60: This still looks harder to use than having an explicit type

```suggestion
        auto type() const -> Type;
        auto as_string() const -> std::string  pre (type() == string);
        auto as_int() const -> int pre (type() == integer);
        void for_each(std::function<void(Node const&)> const& f) const pre (type() == sequence);
        auto at(std::string const& key) const -> Node pre (type() == dictionary);
```

AlanGriffiths (changes requested):
OK, we can't _actually_ use C++ `pre`conditions for a few years, but the "discriminated union" style seems less fragile than the "all attributes are optional" style.

include/miral/miral/display_configuration.h
```diff
@@ -41,6 +44,40 @@ class ConfigurationOption;
...
 public:
+    /// A class providing access to an arbitrary piece of data from the [DisplayConfiguration].
+    /// This is specifically useful when a user wants to extend the based display configuration
+    /// with some sort of custom payload (e.g. a user may want to extend the layout configuration
+    /// with information describing the position and size of specific applications).
+    /// \remark Since MirAL 5.3
+    class Node
+    {
+    public:
+        ~Node();
```
AlanGriffiths, line 55: Is there any value in this being public?

include/miral/miral/display_configuration.h
```diff
@@ -41,6 +44,40 @@ class ConfigurationOption;
...
+        void for_each(std::function<void(Node const&)> const& f) const;
+        auto has(std::string const& key) const -> bool;
+        auto at(std::string const& key) const -> Node;
+
+    private:
+        friend DisplayConfiguration;
+        struct Self;
+
+        explicit Node(std::unique_ptr<Self> self);
+        std::unique_ptr<Self> self;
```
AlanGriffiths, line 78:
```suggestion
        explicit Node(std::unique_ptr<Self>&& self);
        std::unique_ptr<Self> const self;
```

src/miral/display_configuration.cpp
```diff
@@ -120,6 +121,117 @@ void miral::DisplayConfiguration::add_output_attribute(std::string const& key)
...
+            return Type::string;
+        }
+    }
+    else if (self->node.IsMap())
+        return Type::map;
+    else if (self->node.IsSequence())
+        return Type::sequence;
+    else
+    {
+        mir::log_error("Checked type is none of the supported types");
```
AlanGriffiths, line 179: I don't think this is an error here (we're bubbling the failure up to the caller). And we should provide as much context as possible to help a user fix the problem. (The filename and line number would be ideal, but I don't think that's available)

```suggestion
        mir::log_warning("Parsing custom display configuration node: Checked type is none of the supported types");
```

src/miral/display_configuration.cpp
```diff
@@ -120,6 +121,117 @@ void miral::DisplayConfiguration::add_output_attribute(std::string const& key)
...
+        return Type::unknown;
+    }
+}
+
+auto miral::DisplayConfiguration::Node::as_string() const -> std::string
+{
+    if (type() != Type::string)
+        mir::fatal_error("Attempting to access a Node of type %d as a string", type());
```
AlanGriffiths, line 187: [non blocking] More context would be good, as would a type->text mapping

src/miral/static_display_config.cpp
```diff
@@ -529,11 +539,37 @@ auto miral::YamlFileDisplayConfig::list_layouts() const -> std::vector<std::stri
...
+{
+    std::lock_guard lock{mutex};
+    for (auto const& [first, second]: config)
+    {
+        if (first == layout)
+        {
+            if (second.userdata.contains(key))
+                return second.userdata.at(key);
+
+            mir::log_warning("No user data on layout=%s for key=%s", layout.c_str(), key.c_str());
```
AlanGriffiths, line 552:
```suggestion
            mir::log_info("Parsing display configuration: No user data on layout=%s for key=%s", layout.c_str(), key.c_str());
```

src/miral/static_display_config.cpp
```diff
@@ -529,11 +539,37 @@ auto miral::YamlFileDisplayConfig::list_layouts() const -> std::vector<std::stri
...
+            return std::nullopt;
+        }
+    }
+
+    mir::log_error("Cannot find layout: %s", layout.c_str());
```
AlanGriffiths, line 557:
```suggestion
    mir::log_warning("Parsing display configuration: Cannot find layout: %s", layout.c_str());
```
(And similar)

src/miral/display_configuration.cpp
```diff
@@ -25,6 +25,7 @@
 
 #include <fstream>
 #include <sstream>
+#include <yaml-cpp/yaml.h>
```
AlanGriffiths, line 28: We usually put std headers at the end

AlanGriffiths (changes requested):
Nitpicking

include/miral/miral/display_configuration.h
mattkae: It needs to be public for `std::optional` to work

include/miral/miral/display_configuration.h
mattkae: This can't be const, because we need `Node` to be moveable to use std::optional

AlanGriffiths (approved):
Is there really no case for being able to iterate over the keys of a dictionary node?

include/miral/miral/display_configuration.h
mattkae: Yeah I think non-const is fine for now :+1:

mattkae:
> Is there really no case for being able to iterate over the keys of a dictionary node?

There might be a reason in the future, but I am hesitant to add it to the API until that reason shows itself

PR #3871, opened 2025-04-03
Add `BasicAccessibilityManager::Registration`

tarek-y-ismail:
Just a simple wrapper over a boolean + a transformer to prevent double-registrations and erroneous `on_{enabled,disabled}` calls.

### DIFF
--- a/src/server/shell/basic_accessibility_manager.cpp
+++ b/src/server/shell/basic_accessibility_manager.cpp
@@ -32,7 +32,7 @@ inline Transformer* mir::shell::BasicAccessibilityManager::Registration<Transfor
 }
 
 template <typename Transformer>
-inline void mir::shell::BasicAccessibilityManager::Registration<Transformer>::remove_registration()
+inline void mir::shell::BasicAccessibilityManager::Registration<Transformer>::remove_registration() const
 {
     // atomic::compare_exchange_strong(expected, desired) checks the atomic
     // value against the "expected" value. If they're equal, it returns `true`
@@ -46,7 +46,7 @@ inline void mir::shell::BasicAccessibilityManager::Registration<Transformer>::re
 }
 
 template <typename Transformer>
-inline void mir::shell::BasicAccessibilityManager::Registration<Transformer>::add_registration()
+inline void mir::shell::BasicAccessibilityManager::Registration<Transformer>::add_registration() const
 {
     auto expected = false;
     if (registered.compare_exchange_strong(expected, true))
--- a/src/server/shell/basic_accessibility_manager.h
+++ b/src/server/shell/basic_accessibility_manager.h
@@ -85,8 +85,8 @@ private:
             std::shared_ptr<Transformer> const& transformer,
             std::shared_ptr<input::InputEventTransformer> const& event_transformer);
 
-        void add_registration();
-        void remove_registration();
+        void add_registration() const;
+        void remove_registration() const;
 
         Transformer* operator->() const noexcept;
 
@@ -96,14 +96,14 @@ private:
 
         std::shared_ptr<Transformer> const transformer;
         std::shared_ptr<input::InputEventTransformer> const event_transformer;
-        std::atomic<bool> registered{false};
+        std::atomic<bool> mutable registered{false};
     };
 
     Synchronised<MutableState> mutable_state;
 
     bool const enable_key_repeat;
     std::shared_ptr<graphics::Cursor> const cursor;
-    Registration<MouseKeysTransformer> transformer;
+    Registration<MouseKeysTransformer> const transformer;
 };
 }
 }


src/server/symbols.map
tarek-y-ismail, line 1: Since I'm breaking ABI (and API) by removing a symbol, I needed to bump and move all symbols to the new stanza.

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -52,15 +52,33 @@ class InputEventTransformer : public EventFilter
...
 
+    struct Registration
+    {
+        Registration(
+            InputEventTransformer* event_transformer,
+            std::shared_ptr<mir::input::InputEventTransformer::Transformer> transformer);
+
+        ~Registration();
+
+        Registration(Registration&& other) = default;
```
AlanGriffiths, line 63: I don't think the `std::function` move constructor does what we need here.

Specifically, we need the "source" `unregister` to hold a no-op. But I don't think that is what the library specification gives us. By my reading it gives a "valid but unspecified state".

AlanGriffiths (changes requested):
Looks sensible, but I think move-from `Registration` objects will cause premature calls to `remove()`

src/server/input/input_event_transformer.cpp
```diff
@@ -142,3 +142,28 @@ mir::input::InputEventTransformer::Registration::~Registration()
...
     unregister();
 }
+
+void mir::input::InputEventTransformer::Registration::swap(Registration& other) noexcept
+{
+    unregister.swap(other.unregister);
+}
+
+mir::input::InputEventTransformer::Registration::Registration(Registration&& other) noexcept :
+    Registration()
```
AlanGriffiths, line 152: Hmm, you are telling the compiler this doesn't throw, but the delegated constructor, `Registration()`, isn't `noexcept` (and I don't think it is).

tarek-y-ismail: Following the LSP into whatever constructor was called shows the following documentation:
```cpp
      /**
       *  @brief Builds a %function that targets a copy of the incoming
       *  function object.
       *  @param __f A %function object that is callable with parameters of
       *  type `ArgTypes...` and returns a value convertible to `Res`.
       *
       *  The newly-created %function object will target a copy of
       *  `__f`. If `__f` is `reference_wrapper<F>`, then this function
       *  object will contain a reference to the function object `__f.get()`.
       *  If `__f` is a null function pointer, null pointer-to-member, or
       *  empty `std::function`, the newly-created object will be empty.
       *
       *  If `__f` is a non-null function pointer or an object of type
       *  `reference_wrapper<F>`, this function will not throw.
       */
      // _GLIBCXX_RESOLVE_LIB_DEFECTS
      // 2774. std::function construction vs assignment
      template<typename _Functor,
	       typename _Constraints = _Requires<_Callable<_Functor>>>
	function(_Functor&& __f)
	noexcept(_Handler<_Functor>::template _S_nothrow_init<_Functor>())
```

So I think the constructor will not throw since I doubt `[]{}` is allocated at runtime (which could fail) , I just forgot to mark it as `noexcept`

src/server/input/input_event_transformer.cpp
```diff
@@ -123,3 +126,43 @@ bool mi::InputEventTransformer::remove(std::shared_ptr<mi::InputEventTransformer
...
+auto mir::input::InputEventTransformer::Registration::operator=(Registration&& other) noexcept -> Registration&
+{
+    other.swap(*this);
+    return *this;
+}
+
+mir::input::InputEventTransformer::Registration::Registration() noexcept :
+    unregister{[]
+               {
+               }}
```
AlanGriffiths, line 166: You are telling the compositor that `std::function<void()>::std::function(F&& f)` does not throw. That is not true

tarek-y-ismail: Ok, I'm willing to concede since I trust your expertise more than mine ;), but I've got a couple of related questions:
1. Are  you saying `std::function` might potentially throw because it might have to allocate, or for another reason?
2. what's the difference between constructors 4 and 5 on [cppreference](https://en.cppreference.com/w/cpp/utility/functional/function/function)?

AlanGriffiths: 1. because the standard says it might
2. Moving from a `std::function` of the same type is noexcept, constructing from a callable rvalue reference isn't

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -52,15 +52,36 @@ class InputEventTransformer : public EventFilter
...
+    };
+
     InputEventTransformer(std::shared_ptr<InputDeviceRegistry> const&, std::shared_ptr<MainLoop> const&);
     ~InputEventTransformer();
 
     bool handle(MirEvent const&) override;
 
-    void append(std::weak_ptr<Transformer> const&);
-    bool remove(std::shared_ptr<Transformer> const&);
+    [[nodiscard]] auto append(std::weak_ptr<Transformer> const&) -> std::optional<Registration>;
```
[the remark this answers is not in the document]
tarek-y-ismail: Depending on the pointer being deleted might make preserving the state of transformers between toggles a bit more difficult. So I think sticking to `append`/`remove` and removing all the `weak_ptr` related code might actually make more sense for our use case

tarek-y-ismail:
@AlanGriffiths still have one more commit that wraps transformers and their registration up to tidy things up a tad. Just keeping it locally until things settle down.

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -103,11 +103,10 @@ void mir::shell::BasicAccessibilityManager::mousekeys_enabled(bool on)
         main_loop->spawn([this]
         {
             auto state = this->mutable_state.lock();
-            if (!state->mousekey_pointer)
+            if (!transformer.is_registered())
```
AlanGriffiths, line 106: The point of `InputEventTransformer::Registration` was to avoid the need for BAM to track registrations explicitly. Adding a wrapper around that so that we can test `is_registered()` rather defeats the point

src/server/shell/basic_accessibility_manager.h
```diff
@@ -86,15 +88,63 @@ class BasicAccessibilityManager : public AccessibilityManager
...
     std::shared_ptr<MainLoop> const main_loop;
     std::shared_ptr<input::CompositeEventFilter> const the_composite_event_filter;
     bool const enable_key_repeat;
     std::shared_ptr<graphics::Cursor> const cursor;
-    std::shared_ptr<mir::input::InputEventTransformer> const event_transformer;
-    std::shared_ptr<mir::shell::MouseKeysTransformer> const transformer;
     std::shared_ptr<mir::input::InputDeviceRegistry> const input_device_registry;
+    std::shared_ptr<mir::input::InputEventTransformer> const event_transformer;
+
+    Registration<MouseKeysTransformer> transformer;
```
AlanGriffiths, line 147: This is mutable - shouldn't it be part of the state subobject?

src/server/shell/basic_accessibility_manager.h
```diff
@@ -86,15 +88,63 @@ class BasicAccessibilityManager : public AccessibilityManager
...
+        }
+
+        ~Registration()
+        {
+            remove_registration();
+        }
+
+        void add_registration()
+        {
+            assert(!registration.has_value());
```
AlanGriffiths, line 109: 1. Do not put implementation in header files without a good reason: it makes for fragile ABIs and increases compile times.
3. Do not use `assert()` to test preconditions: if they are worth testing, then they should (mostly) be tested regardless of `NDEBUG`.

src/server/shell/basic_accessibility_manager.h
```diff
@@ -17,10 +17,12 @@
 #ifndef MIR_SHELL_BASIC_ACCESSIBILITY_MANAGER_H
 #define MIR_SHELL_BASIC_ACCESSIBILITY_MANAGER_H
 
+#include "mir/input/input_event_transformer.h"
 #include "mir/shell/accessibility_manager.h"
 
 #include "mir/input/mousekeys_keymap.h"
 #include "mir/synchronised.h"
+#include <cassert>
```
AlanGriffiths, line 25: Do not include `cassert` in a header file: there should be nothing in a header that uses it: if there is, it is very easy to break the One Definition Rule by accident: two translation units can include the header with different values for `NDEBUG`.

AlanGriffiths (changes requested):
The original point of this PR was to make the `InputEventTransformer` API simpler to use. The current iteration makes it more complex

src/server/shell/basic_accessibility_manager.cpp
AlanGriffiths: > The wrapper isn't for `is_registered`, I could also do `std::optional::has_value` to the same effect. The problem `is_registered` solves here is covering for shell author errors such as disabling mousekeys when already disabled or double-enabling it. Later on once #3873 is implemented, we'll also need to track registrations to enable or disable the virtual device.

I agree that `std::optional::has_value` is equivalent. But the idea of `InputEventTransformer::Registration`
to avoid tracking registrations. _Something_ is wrong if are also adding code to track registrations.

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -53,12 +75,15 @@ class InputEventTransformer
         EventBuilder* builder,
         EventDispatcher const& dispatcher);
 
-    void append(std::weak_ptr<Transformer> const&);
-    bool remove(std::shared_ptr<Transformer> const&);
+    /// Appends the transformer to be used in `transform`
+    /// Precondition: Transformer isn't already registered.
```
AlanGriffiths, line 79: Do we need this precondition?

1. We don't check the precondition in the implementation;
2. There's no way for the user of the API to verify the precondition; and,
3. There's no obvious difficulty with registering a transformer twice (see below)

I realize that (pseudo code):

```c++
auto ra = append(a);
auto rb = append(b);
{
    auto rc = append(a);
}
```

Will leave the transformers as `{b, a}`, not the slightly more intuitive `{a, b}`, but how much of a problem is that?

AlanGriffiths:
@tarek-y-ismail what is the status of this? There are conflicts and unresolved discussions

tarek-y-ismail:
> @tarek-y-ismail what is the status of this? There are conflicts and unresolved discussions

Will give it a look soon™. Last I remember,  I was waiting for your feedback.

tarek-y-ismail:
So it seems that the original purpose of this PR was "lost". At the moment, the current set up with `InputEventTransformer` works well enough to automatically unregister transformers when their owners are freed. This was the main purpose of `InputEventTransformer::Registration`. 

The other thing that this PR addressed was the need for both the transformer, and an accompanying boolean to help with avoiding double (un)registerations (originally), and to prevent `on_{disabled, enabled}` from being erroneously called (recently). 

So here's what I'm going to propose: We remove `InputEventTransformer::Registration`, and keep `BasicAccessibilityManager::Registeration` as a helper.

src/server/shell/basic_accessibility_manager.h
```diff
@@ -68,23 +68,42 @@ class BasicAccessibilityManager : public AccessibilityManager
...
+        Registration(
+            std::shared_ptr<Transformer> const& transformer,
+            std::shared_ptr<input::InputEventTransformer> const& event_transformer);
+
+        void add_registration();
+        void remove_registration();
+
+        Transformer* operator->() const noexcept;
+
+        bool is_registered() const noexcept;
```
AlanGriffiths, line 91: Unused
```suggestion
```

src/server/shell/basic_accessibility_manager.h
```diff
@@ -68,23 +68,42 @@ class BasicAccessibilityManager : public AccessibilityManager
...
+
+    private:
+        Registration(Registration const&) = delete;
+        Registration& operator=(Registration const&) = delete;
+
+        std::shared_ptr<Transformer> const transformer;
+        std::shared_ptr<input::InputEventTransformer> const event_transformer;
+        bool registered{false};
```
AlanGriffiths, line 99: This mutable state is accessed without synchronization. This is probably a case for `std::atomic<bool>` and `compare_exchange_strong()`

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -18,16 +18,54 @@
...
+template <typename Transformer>
+inline Transformer* mir::shell::BasicAccessibilityManager::Registration<Transformer>::operator->() const noexcept
+{
+    return transformer.get();
+}
+
+template <typename Transformer>
+inline void mir::shell::BasicAccessibilityManager::Registration<Transformer>::remove_registration()
+{
+    if(registered)
```
AlanGriffiths, line 43: Whitespace

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -18,16 +18,54 @@
...
+    {
+        event_transformer->remove(transformer);
+        registered = false;
+    }
+}
+
+template <typename Transformer>
+inline void mir::shell::BasicAccessibilityManager::Registration<Transformer>::add_registration()
+{
+    if(!registered)
```
AlanGriffiths, line 53: Whitespace

AlanGriffiths (changes requested):
I'd be very tempted to make `transformer` `const` (as it is logically const) by making `registered` `mutable`

Also the PR title and description are obsolete

PR #3873, opened 2025-04-04
Implement Simulated Secondary Click

tarek-y-ismail:


src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,168 @@
...
+            // means the user let go of the _left_ mouse button, we should
+            // cancel the pending right click, expect the down and up
+            // events of that left click, and issue a left click.
+            if (action == mir_pointer_action_button_up &&
+                secondary_click_dispatcher->state() == time::Alarm::State::pending)
+            {
+                secondary_click_dispatcher->cancel();
+                state = State::waiting_for_synthesized_left_down;
+                click(dispatcher, builder, mir_pointer_button_primary);
+                return false;
```
tarek-y-ismail, line 116: For some reason, returning `true` from this causes simulated secondary clicks followed by a "normal" left click to not register the normal left click. I think there might be some sort of funkiness here since `dispatch` will ~yeet~ throw the event dispatching onto the main thread, which may or may not cause `transform_input_event` to be called again _before_ we get to the return statement.

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,168 @@
...
+                {
+                    secondary_click_dispatcher = main_loop->create_alarm(
+                        [dispatcher, builder]
+                        {
+                            click(dispatcher, builder, mir_pointer_button_secondary);
+                        });
+                }
+
+                secondary_click_dispatcher->cancel();
+                secondary_click_dispatcher->reschedule_in(std::chrono::seconds(1));
```
tarek-y-ismail, line 96: Should be customizable

src/server/shell/accessibility_manager.cpp
```diff
@@ -114,24 +115,30 @@ struct MouseKeysTransformer: public mir::input::InputEventTransformer::Transform
...
     transformer{[&]
                 {
                     return (options->get<bool>(options::enable_mouse_keys_opt)) ?
                                std::make_shared<MouseKeysTransformer>() :
                                std::shared_ptr<MouseKeysTransformer>{};
-                }()
-
+                }()},
+    simulated_secondary_click_transformer{
+        std::make_shared<SimulatedSecondaryClickTransformer>(main_loop),
```
tarek-y-ismail, line 129: Shouldn't be always enabled. Needs `miral::SimulatedSecondaryClickConfig` to add `--simulated-secondary-click` an `--simulated-secondary-click-delay` as a config options and as public methods for runtime changes.

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -21,7 +21,6 @@
 #include "mir/input/event_builder.h"
 #include "mir_toolkit/events/event.h"
 
-#include <chrono>
```
tarek-y-ismail, line 24: Irrelevant

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,168 @@
...
+    dispatcher(builder->pointer_event(
+        std::nullopt,
+        mir_pointer_action_button_up,
+        0,
+        std::nullopt,
+        {0, 0},
+        mir_pointer_axis_source_none,
+        mir::events::ScrollAxisH{},
+        mir::events::ScrollAxisV{}));
+}
```
tarek-y-ismail, line 51: This is likely to be common between other pointer related accessibility methods, should probably be pulled into a common place.

src/server/shell/simulated_secondary_click_transformer.h
```diff
@@ -0,0 +1,58 @@
...
+    SimulatedSecondaryClickTransformer(std::shared_ptr<mir::MainLoop> const& main_loop);
+
+    bool transform_input_event(
+        input::InputEventTransformer::EventDispatcher const& dispatcher,
+        input::EventBuilder*,
+        MirEvent const&) override;
+
+private:
+    std::shared_ptr<mir::MainLoop> const main_loop;
+    std::unique_ptr<time::Alarm> secondary_click_dispatcher;
```
tarek-y-ismail, line 40: Not `const` since we need to create it at a time where we have a reference to the dispatcher and event builder.

src/server/shell/simulated_secondary_click_transformer.h
```diff
@@ -0,0 +1,58 @@
...
+
+    enum class State
```
tarek-y-ismail, line 42: Would be a lot simpler if we could tag synthesized events somehow.

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -68,6 +69,7 @@ class AccessibilityManager
 
     std::shared_ptr<mir::input::InputEventTransformer> const event_transformer;
     std::shared_ptr<mir::input::InputEventTransformer::Transformer> const transformer;
+    std::shared_ptr<mir::input::InputEventTransformer::Transformer> const simulated_secondary_click_transformer;
```
tarek-y-ismail, line 72: TODO: `AccessibilityManager` tests for `SimulatedSecondaryClickTransformer`'s stuff once customization is added. Will need to pull out `SimulatedSecondaryClickTransformer` to be an interface so we can create a mock of it.

src/server/shell/simulated_secondary_click_transformer.h
```diff
@@ -0,0 +1,58 @@
...
+namespace mir
+{
+namespace shell
+{
+class SimulatedSecondaryClickTransformer : public mir::input::InputEventTransformer::Transformer
+{
+public:
+    SimulatedSecondaryClickTransformer(std::shared_ptr<mir::MainLoop> const& main_loop);
+
+    bool transform_input_event(
```
tarek-y-ismail, line 33: TODO: test this fella

tarek-y-ismail (commented):
Self review

src/server/shell/simulated_secondary_click_transformer.h
tarek-y-ismail: I think I've tested all I could?

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,206 @@
...
+                {
+                    mutable_state.lock()->on_hold_cancel();
+                    click(dispatcher, builder, mir_pointer_button_primary);
+                }
+
+                // Whether the user cancelled early or not, reset the state
+                // machine and the alarm, and consume the event.
+                state = State::waiting_for_real_left_down;
+                secondary_click_dispatcher->cancel();
+                return false;
```
tarek-y-ismail, line 170: My mind says that this should be `true` to consume the event, but ~my body~ manual testing tells me otherwise.

src/server/input/input_event_transformer.cpp
```diff
@@ -32,13 +34,31 @@
...
-    std::shared_ptr<InputDeviceRegistry> const& input_device_registry, std::shared_ptr<MainLoop> const& main_loop) :
+    std::shared_ptr<InputDeviceRegistry> const& input_device_registry,
+    std::shared_ptr<MainLoop> const& main_loop,
+    std::shared_ptr<InputDeviceHub> const& input_device_hub) :
     virtual_pointer{
         std::make_shared<mir::input::VirtualInputDevice>("mousekey-pointer", mir::input::DeviceCapability::pointer)},
     input_device_registry{input_device_registry},
-    main_loop{main_loop}
+    main_loop{main_loop},
+    virtual_device_id_{[this, &input_device_registry, &input_device_hub]
```
tarek-y-ismail, line 44: This looks like a hack because it is. If anyone knows a better way to get the virtual device ID so I can ignore events dispatched from it, I'm all ears ;)

src/miral/simulated_secondary_click_config.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+void miral::SimulatedSecondaryClickConfig::enabled(bool enabled) const
+{
+    if (self->accessibility_manager.expired())
+    {
+        mir::log_error("AccessibilityManager not initialized. Will not toggle simulated secondary click");
+        return;
+    }
+
+    self->accessibility_manager.lock()->simulated_secondary_click_enabled(enabled);
```
RAOF, line 48: This is a possible TOCTTOU problem?

It would be more obviously correct as something like:
```suggestion
    if (auto manager = self->accessibility_manager.lock())
    {
        manager->simulated_secondary_click_enabled(enabled);
    }
    else
    {
        mir::log_error("AccessibilityManager not initialized. Will not toggle simulated secondary click");
        return;
    }
```

src/miral/simulated_secondary_click_config.cpp
```diff
@@ -0,0 +1,117 @@
...
+    std::weak_ptr<mir::shell::AccessibilityManager> accessibility_manager;
+    bool const enabled_by_default;
+};
+
+miral::SimulatedSecondaryClickConfig::SimulatedSecondaryClickConfig(bool enabled_by_default) :
+    self{std::make_shared<Self>(enabled_by_default)}
+{
+}
+
+void miral::SimulatedSecondaryClickConfig::enabled(bool enabled) const
```
RAOF, line 40: I'd expect `enabled()` to be an accessor, rather than a setter? Should this be `enable()`?

src/server/input/input_event_transformer.cpp
tarek-y-ismail: `add_device` returns a `weak_ptr` to the device, where we can access `device->id()` directly.

src/miral/simulated_secondary_click_config.cpp
tarek-y-ismail: Just staying consistent with `MouseKeysConfig`. But willing to change both if you deem it necessary.

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,206 @@
...
+                state = State::waiting_for_drag_end_left_up;
+                return true;
+            }
+
+            // Can't check `pointer_event->buttons()` because that's 0 for button up events
+            if (action == mir_pointer_action_button_up)
+            {
+                // If we get an up event BEFORE the alarm is triggered, that means
+                // the user let go of the _left_ mouse button, we should cancel the
+                // pending right click and issue a left click.
```
tarek-y-ismail, line 159: I don't think this is entirely true. There must be a way to figure out which button was released.

src/server/symbols.map
```diff
@@ -401,6 +402,7 @@ global:
     mir::input::Validator::Validator*;
     mir::input::Validator::validate_and_dispatch*;
     mir::input::VirtualInputDevice::VirtualInputDevice*;
+    mir::input::VirtualInputDevice::get_device_info*;
```
tarek-y-ismail, line 405: Don't forget to remove this once the corresponding method is made to be private.

src/server/symbols.map
```diff
@@ -892,6 +894,7 @@ global:
     non-virtual?thunk?to?mir::input::TouchVisualizer::?TouchVisualizer*;
     non-virtual?thunk?to?mir::input::VTFilter::?VTFilter*;
     non-virtual?thunk?to?mir::input::VTFilter::handle*;
+    non-virtual?thunk?to?mir::input::VirtualInputDevice::get_device_info*;
```
tarek-y-ismail, line 897: This one too!

tests/unit-tests/shell/test_simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+struct TestDifferentReleaseTimings :
+    public TestSimulatedSecondaryClickTransformer,
+    public WithParamInterface<std::tuple<std::chrono::milliseconds, MirPointerButton>>
+{
+};
+
+TEST_P(
+    TestDifferentReleaseTimings,
+    releasing_primary_button_before_secondary_click_delay_dispatches_a_primary_click)
```
tarek-y-ismail, line 106: Feels like this is testing too many things at once, but I don't think splitting things is worth the amount of duplication

tarek-y-ismail (commented):
Another self review

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -67,11 +69,17 @@ mir::shell::BasicAccessibilityManager::BasicAccessibilityManager(
...
     event_transformer{event_transformer},
-    transformer{mousekeys_transformer}
+    transformer{mousekeys_transformer},
+    simulated_secondary_click_transformer{simulated_secondary_click_transformer}
+{
+}
+
+mir::shell::BasicAccessibilityManager::~BasicAccessibilityManager()
 {
 }
```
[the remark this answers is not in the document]
tarek-y-ismail: Pre-existing, so let's leave it as is (at least until all a11y work is done)

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,206 @@
...
+    virtual_device_id{virtual_device_id}
+{
+}
+
+bool mir::shell::BasicSimulatedSecondaryClickTransformer::transform_input_event(
+    input::InputEventTransformer::EventDispatcher const& dispatcher,
+    input::EventBuilder* builder,
+    MirEvent const& event)
+{
+    std::lock_guard guard{reentrance_mutex};
```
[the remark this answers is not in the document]
tarek-y-ismail: Two things:
1. `InputEventTransformer` prepends itself onto the start of the input chain.
2. `InputEventTransformer::handle` is protected by a mutex. 

Thus, this fellow is _really_ not needed.

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+                state = State::waiting_for_drag_end_left_up;
+                return true;
+            }
+
+            // `pointer_event->buttons()` contains the _current_ button state.
+            // Since we're in this state machine state, it's implicitly known
+            // that the left button was pressed, so we just need to make sure
+            // it's not pressed now.
+            if (action == mir_pointer_action_button_up && !(pointer_event->buttons() & mir_pointer_button_primary))
```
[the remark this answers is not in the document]
tarek-y-ismail: You fell for it as well :)

`buttons()` returns the current state of whatever buttons are down. On the up event, the bit corresponding to the primary should be cleared, not set.

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,210 @@
...
+                if (secondary_click_dispatcher->state() == time::Alarm::State::pending)
+                {
+                    mutable_state.lock()->on_hold_cancel();
+                    click(dispatcher, builder, mir_pointer_button_primary);
+                }
+
+                // Whether the user cancelled early or not, reset the state
+                // machine and the alarm, and consume the event.
+                state = State::waiting_for_real_left_down;
+                secondary_click_dispatcher->cancel();
```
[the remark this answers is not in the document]
tarek-y-ismail: Should be easily fixed by cancelling first

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+#include "mir/events/event.h"
+#include "mir/events/input_event.h"
+#include "mir/events/pointer_event.h"
+#include "mir/main_loop.h"
+#include "mir/events/event_builders.h"
+
+#include <cassert>
+#include <mir/log.h>
+#include <mir/event_printer.h>
```
tarek-y-ismail: Added for debugging, should be removed.

src/server/shell/simulated_secondary_click_transformer.cpp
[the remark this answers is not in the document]
tarek-y-ismail: You fell for it as well :)

`buttons()` contains the _current_ state of all mouse buttons, so when the left button is released, its bit is cleared.

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,210 @@
...
+                //
+                // It will be in the order: motion (you are here), "fake" down
+                // event, "fake" motion event.
+                dispatcher(
+                    mir::events::make_pointer_event(
+                        pointer_event->device_id(),
+                        pointer_event->event_time(),
+                        pointer_event->modifiers(),
+                        pointer_event->action(),
+                        pointer_event->buttons() | mir_pointer_button_primary,
```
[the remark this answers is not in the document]
tarek-y-ismail: Was just being paranoid, I logged the event and it does have the primary button down, so that's not necessary.

src/server/shell/simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,210 @@
...
+            if (action == mir_pointer_action_motion)
+            {
+                if (secondary_click_dispatcher->state() == time::Alarm::pending)
+                {
+                    secondary_click_dispatcher->cancel();
+                    mutable_state.lock()->on_hold_cancel();
+                }
+
+                // Down event instead of the one we consumed the last time around
+                dispatcher(
```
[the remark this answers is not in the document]
tarek-y-ismail: Nice catch. Luckily events have a `clone()` and `set_event_time()`, so I can clone the event and just update the time.

src/server/shell/simulated_secondary_click_transformer.cpp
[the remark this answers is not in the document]
tarek-y-ismail: Should be easily fixed by cancelling _before_ dispatching the click. Thank You!

include/miral/miral/simulated_secondary_click_config.h
```diff
@@ -0,0 +1,51 @@
...
+    SimulatedSecondaryClickConfig(bool enabled_by_default);
+
+    void operator()(mir::Server& server) const;
+
+    void enabled(bool enabled) const;
+    void hold_duration(std::chrono::milliseconds hold_duration) const;
+
+    void hold_start(std::function<void()>&& on_hold_start) const;
+    void hold_cancel(std::function<void()>&& on_hold_cancel) const;
+    void secondary_click(std::function<void()>&& on_secondary_click) const;
```
AlanGriffiths, line 43: What requirement are we satisfying here? I can imagine the shell wants to enable/disable the feature and specify a hold duration to trigger the simulation, but what is the purpose of the various callbacks?

src/include/server/mir/shell/accessibility_manager.h
```diff
@@ -49,6 +52,12 @@ class AccessibilityManager
     virtual void mousekeys_keymap(input::MouseKeysKeymap const& new_keymap) = 0;
     virtual void acceleration_factors(double constant, double linear, double quadratic) = 0;
     virtual void max_speed(double x_axis, double y_axis) = 0;
+
+    virtual void simulated_secondary_click_enabled(bool enabled) = 0;
+    virtual void simulated_secondary_click_hold_duration(std::chrono::milliseconds hold_duration) = 0;
+    virtual void simulated_secondary_click_hold_start(std::function<void()>&&) = 0;
+    virtual void simulated_secondary_click_hold_cancel(std::function<void()>&&) = 0;
+    virtual void simulated_secondary_click_secondary_click(std::function<void()>&&) = 0;
```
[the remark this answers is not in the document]
AlanGriffiths: This set of functions does look like a missing abstraction

src/miral/simulated_secondary_click_config.cpp
AlanGriffiths: It might also be clearer with `enable()`/`disable()`

src/miral/simulated_secondary_click_config.cpp
```diff
@@ -0,0 +1,130 @@
...
+{
+    if (auto const accessibility_manager = self->accessibility_manager.lock())
+    {
+        self->accessibility_manager.lock()->simulated_secondary_click_enabled(enabled);
+    }
+    else
+    {
+        mir::log_error("AccessibilityManager not initialized. Will not toggle simulated secondary click");
+        return;
+    }
```
AlanGriffiths, line 50: Wouldn't it make more sense to the shell if we recorded the setting for use when the init callback runs?

src/miral/simulated_secondary_click_config.cpp
```diff
@@ -0,0 +1,130 @@
...
+{
+    if (auto const accessibility_manager = self->accessibility_manager.lock())
+    {
+        accessibility_manager->simulated_secondary_click_hold_duration(delay);
+    }
+    else
+    {
+        mir::log_error("AccessibilityManager not initialized. Will not update simulated secondary click hold duration");
+        return;
+    }
```
AlanGriffiths, line 63: Wouldn't it make more sense to the shell if we recorded the setting for use when the init callback runs?

AlanGriffiths (changes requested):
Some initial thoughts: Will look closer tomorrow...

tests/unit-tests/shell/test_simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include <mir/event_printer.h>
+
+namespace mi = mir::input;
+namespace mt = mir::test;
+namespace mtd = mt::doubles;
+
+using namespace ::testing;
+using namespace std::chrono_literals;
+
+struct TestSimulatedSecondaryClickTransformer : Test
```
mattkae, line 42: Could you also add tests for clicking and releasing another button while holding down the left click?

tests/unit-tests/shell/test_simulated_secondary_click_transformer.cpp
mattkae: Personally, I would split them up, but I won't nit pick on this

mattkae (changes requested):
A few more things :)

tarek-y-ismail:
Just missing header and sphinx docs. Should be easy to whip up tomorrow.

tarek-y-ismail (commented):
Self review

src/server/shell/basic_simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,190 @@
...
+                state = State::waiting_for_motion_or_real_left_up;
+                mutable_state.lock()->on_hold_start();
+
+                return true; // Consume event
+            }
+        }
+        break;
+    case State::waiting_for_motion_or_real_left_up: // This state implies that the left button was pressed down
+        {
+            if (action == mir_pointer_action_motion)
```
tarek-y-ismail, line 106: Should probably add a threshold for the displacement/distance that would cancel a simulated secondary click.

src/server/shell/basic_simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,190 @@
...
+        return false;
+
+    auto const input_event = event.to_input();
+    if (input_event->input_type() != mir_input_event_type_pointer)
+        return false;
+
+    auto const pointer_event = input_event->to_pointer();
+    // Skip events dispatched from transformers using the virtual device
+    if(pointer_event->device_id() == virtual_device_id)
+        return false;
```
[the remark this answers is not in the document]
tarek-y-ismail: Something like `as_pointer_event` at least...

include/miral/miral/simulated_secondary_click_config.h
```diff
@@ -0,0 +1,52 @@
...
+
+    void operator()(mir::Server& server) const;
+
+    void enable() const;
+    void disable() const;
+    void hold_duration(std::chrono::milliseconds hold_duration) const;
+
+    void hold_start(std::function<void()>&& on_hold_start) const;
+    void hold_cancel(std::function<void()>&& on_hold_cancel) const;
+    void secondary_click(std::function<void()>&& on_secondary_click) const;
```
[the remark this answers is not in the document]
mattkae: I also think that this is a good opportunity for method chaining (i.e. returning `*this`) as users may want to set all of the callbacks at startup

src/server/shell/basic_simulated_secondary_click_transformer.cpp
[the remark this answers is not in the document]
mattkae: Eh, I'm okay with the boilerplate for now :)

mattkae (changes requested):
A few minor nits!

tarek-y-ismail:
Just realized I didn't test `SimulatedSecondaryClickConfig` yet :man_facepalming:

mattkae (changes requested):
I have no more problems with the code, but it appears there are some bugs :bug:

1. Simulated secondary click does not work in Gedit for me (but right click does bring up a context menu)
2. When I hold down left click and trigger a context menu, I am unable to interact with that context menu because it immediately disappears

tarek-y-ismail:
> I have no more problems with the code, but it appears there are some bugs 🐛

I think I figured it out. I was testing using XTerm and apps launched from it just because that was faster. But it turns out that Wayland apps don't just accept events coming from any input device. Masquerading as the "real" input device by hijacking its input device ID fixes the issues you mentioned.

Now, since I was using the device ID to filter "synthesized" events, I'll have to figure out another way to do that. The two options that immediately come to mind are: 
1. Add a "synthesized" property to input events. I'm not sure whether that would break ABI since users access events through opaque pointers. If that doesn't work, there's the second option/hack
2. Use some unused property or unlikely combination of properties. `window_id` seems unused at the moment.

src/server/shell/default_configuration.cpp
```diff
@@ -194,6 +195,8 @@ auto mir::DefaultServerConfiguration::the_accessibility_manager() -> std::shared
                 the_input_event_transformer(),
                 the_options()->get<bool>(mir::options::enable_key_repeat_opt),
                 the_cursor(),
-                std::make_shared<shell::BasicMouseKeysTransformer>(the_main_loop(), the_clock()));
+                std::make_shared<shell::BasicMouseKeysTransformer>(the_main_loop(), the_clock()),
+                std::make_shared<shell::BasicSimulatedSecondaryClickTransformer>(
+                    the_main_loop(), the_input_event_transformer()->virtual_device_id()));
```
AlanGriffiths, line 200: This is fragile: `the_input_event_transformer()` returns a temporary, and the only guarantee that the lifetime of the input event transformer that returns `virtual_device_id()` extends beyond the statement is that there's another `the_input_event_transformer()` temporary being passed to `BasicAccessibilityManager`

AlanGriffiths (changes requested):

debian/libmiral7.symbols
```diff
@@ -536,3 +536,14 @@ libmiral.so.7 libmiral7 #MINVER#
...
+ (c++)"miral::SimulatedSecondaryClickConfig::disable() const@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::disabled(std::function<void ()>&&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::displacement_threshold(float)@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::enable() const@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::enabled(std::function<void ()>&&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::hold_cancel(std::function<void ()>&&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::hold_duration(std::chrono::duration<long, std::ratio<1l, 1000l> >)@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::hold_start(std::function<void ()>&&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::operator()(mir::Server&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::SimulatedSecondaryClickConfig::secondary_click(std::function<void ()>&&)@MIRAL_5.3" 5.3.0
```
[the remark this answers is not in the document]
AlanGriffiths: Oops! I got that wrong

tarek-y-ismail:
@mattkae I think I now have some idea why GTK apps don't like our SSC implementation. The problem stems from how we handle mouse button state. 

Somewhere (maybe in `InputDeviceStateTracker` or the input platform) we set the button state whether or not the event was consumed (i.e. should be treated as if it never existed). By the time the synthesized right click down event is delivered to the surface, it also shows that the left button was down. When the right click up is delivered, it shows the left button also being released. So whatever app receiving the event sees both a left and right click simultaneously.

src/include/common/mir/events/input_event.h
tarek-y-ismail, line 1: Other than the unnecessary changes on lines 24 and 25, I think we can ditch these changes in favor of using the device ID once again. 

There isn't any easy way to obtain that from the virtual input device, but one way could be to just create a dummy event and read the input device id from that. Another way would be to add a method that allows the mousekeys pointer to register itself, where it would read the input device id from the device weak pointer.

include/miral/miral/simulated_secondary_click_config.h
```diff
@@ -0,0 +1,80 @@
...
+    /// [enabled_by_default]
+    SimulatedSecondaryClickConfig(bool enabled_by_default);
+
+    void operator()(mir::Server& server);
+
+    /// Enables simulated secondary click
+    void enable() const;
+
+    /// Disables simulated secondary click
+    void disable() const;
```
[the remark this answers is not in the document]
tarek-y-ismail: I originally imagined these as the last methods you'd call in the chain. But I think you're right, it's better to be consistent. There isn't a particular need to force those onto the end anyway.

src/server/input/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ set(
   virtual_input_device.cpp
   xkb_mapper_registrar.cpp
   input_event_transformer.cpp
+  mousekey_pointer.cpp mousekey_pointer.h
```
AlanGriffiths, line 32:
```suggestion
  mousekey_pointer.cpp ${PROJECT_SOURCE_DIR}/src/include/server/mir/input/mousekey_pointer.h
```

examples/miral-shell/shell_main.cpp
```diff
@@ -172,6 +176,11 @@ int main(int argc, char const* argv[])
         return false;
     };
 
+    auto const ssc_config = miral::SimulatedSecondaryClickConfig{false}
+                                .displacement_threshold(30)
+                                .hold_duration(std::chrono::milliseconds{2000})
+                                .secondary_click([] { mir::log_info("Simulated secondary click dispatched!"); });
+
```
tarek-y-ismail, line 183: Alternative way: reduce the SSC hold duration and remove the log statement.

include/miral/miral/simulated_secondary_click_config.h
```diff
@@ -0,0 +1,52 @@
...
+#include <memory>
+
+namespace mir
+{
+class Server;
+}
+
+namespace miral
+{
+class SimulatedSecondaryClickConfig
```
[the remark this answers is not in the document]
tarek-y-ismail: Now just needs sphinx docs.

src/include/common/mir/events/input_event.h
```diff
@@ -21,6 +21,8 @@
 
 struct MirInputEvent : MirEvent
 {
+    MirInputEvent(MirInputEvent&&) = delete;
+    MirInputEvent& operator=(MirInputEvent&&) = delete;
```
tarek-y-ismail, line 25: Leftover. Should be removed.

src/include/server/mir/default_server_configuration.h
```diff
@@ -350,7 +353,7 @@ class DefaultServerConfiguration : public virtual ServerConfiguration
 
     auto the_decoration_strategy() -> std::shared_ptr<DecorationStrategy> override;
     void set_the_decoration_strategy(std::shared_ptr<DecorationStrategy> strategy) override;
-    auto the_display_configuration_observer() -> std::shared_ptr<graphics::DisplayConfigurationObserver>;
+    std::shared_ptr<graphics::DisplayConfigurationObserver> the_display_configuration_observer();
```
tarek-y-ismail, line 356: Should be reverted.

src/include/server/mir/default_server_configuration.h
```diff
@@ -326,8 +327,10 @@ class DefaultServerConfiguration : public virtual ServerConfiguration
     virtual std::shared_ptr<input::CursorListener> the_cursor_listener();
     virtual std::shared_ptr<input::TouchVisualizer> the_touch_visualizer();
     virtual std::shared_ptr<input::Seat> the_seat();
+
     virtual std::shared_ptr<input::KeyMapper> the_key_mapper();
     virtual std::shared_ptr<input::LedObserverRegistrar> the_led_observer_registrar();
+    virtual std::shared_ptr<input::MousekeyPointer> the_mousekey_pointer();
```
tarek-y-ismail, line 333: I'm not sure whether this is the best way to share `MousekeyPointer` between `the_seat` where its used to filter/transform events, and `the_accessibility_manager` where it's enabled/disabled depending on what accessibility methods are enabled. Suggestions welcome.

src/include/server/mir/input/mousekey_pointer.h
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+
+#ifndef MIR_INPUT_MOUSEKEY_POINTER
+#define MIR_INPUT_MOUSEKEY_POINTER
+
+#include "mir/input/event_filter.h"
+#include "mir/input/input_event_transformer.h"
+#include "mir/input/virtual_input_device.h"
+#include "mir/main_loop.h"
```
tarek-y-ismail, line 24: Should be replaced by forwards decls.

src/miral/simulated_secondary_click_config.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    std::weak_ptr<mir::shell::AccessibilityManager> accessibility_manager;
+
+    struct State
+    {
+        bool enabled;
+        std::chrono::milliseconds hold_duration{1000};
+        float displacement_threshold{20.0f};
+        std::function<void()> on_enabled{[]{}}, on_disabled{[]{}}, on_hold_start{[]{}}, on_hold_cancel{[]{}}, on_secondary_click{[]{}};
```
tarek-y-ismail, line 39: I think this is a little too paranoid and unnecessarily adds noise.

src/miral/simulated_secondary_click_config.cpp
```diff
@@ -0,0 +1,190 @@
...
+            self->accessibility_manager = server.the_accessibility_manager();
+
+            // The default values for these options were set from `self->state`
+            // when the options are added a few lines ago.
+            hold_duration(
+                std::chrono::milliseconds(server.get_options()->get<int>(simulated_secondary_click_delay_opt)));
+            displacement_threshold(
+                server.get_options()->get<double>(simulated_secondary_click_displacement_threshold_opt));
+
+            if (auto on_enabled = self->state.lock()->on_enabled)
```
tarek-y-ismail, line 176: Probably should use `std::move` or use a reference?

src/server/input/basic_seat.h
```diff
@@ -41,7 +41,8 @@ class CursorListener;
 class InputDispatcher;
 class KeyMapper;
 class SeatObserver;
-
+class EventFilter;
```
tarek-y-ismail, line 44: Should be removed

src/server/input/default_configuration.cpp
```diff
@@ -214,7 +214,7 @@ mir::DefaultServerConfiguration::the_touch_visualizer()
             {
                 visualizer->enable();
             }
-            
+
```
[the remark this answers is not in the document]
tarek-y-ismail: Seriously not sure what the heck brings this back. AFAIK I have format-on-save off in my editor!

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -15,51 +15,21 @@
  */
 
 #include "basic_accessibility_manager.h"
+#include "mir/input/input_device_registry.h"
```
tarek-y-ismail, line 18: Probably should be moved somwhere below.

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -165,3 +129,46 @@ void mir::shell::BasicAccessibilityManager::max_speed(double x_axis, double y_ax
...
+
+auto mir::shell::BasicAccessibilityManager::simulated_secondary_click()
+    -> SimulatedSecondaryClickTransformer&
+{
+    return *simulated_secondary_click_transformer;
+}
+
+void mir::shell::BasicAccessibilityManager::ensure_mousekey_pointer_added()
+{
+    if (std::ranges::any_of(std::initializer_list<bool>{mousekeys_on, ssc_on}, std::identity{}))
```
tarek-y-ismail, line 165: Very ugly (thanks, CI compilers & C++), I have some ideas related to #3871 that might improve this.

src/server/shell/basic_accessibility_manager.h
```diff
@@ -19,6 +19,10 @@
 
 #include "mir/shell/accessibility_manager.h"
 
+#include "mir/input/event_filter.h"
+#include "mir/input/input_event_transformer.h"
+#include "mir/input/virtual_input_device.h"
```
tarek-y-ismail, line 24: Could probably also be replaced by forward decls.

src/server/shell/basic_accessibility_manager.h
```diff
@@ -74,27 +79,34 @@ class BasicAccessibilityManager : public AccessibilityManager
...
         std::vector<std::shared_ptr<shell::KeyboardHelper>> keyboard_helpers;
-        std::shared_ptr<MousekeyPointer> mousekey_pointer;
     };
 
+    void ensure_mousekey_pointer_added();
+    void ensure_mousekey_pointer_removed();
+
     Synchronised<MutableState> mutable_state;
 
     std::shared_ptr<MainLoop> const main_loop;
```
tarek-y-ismail, line 99: Redundant

tests/unit-tests/shell/test_basic_accessibility_manager.cpp
```diff
@@ -15,9 +15,11 @@
  */
 
 #include "include/server/mir/shell/keyboard_helper.h"
+#include "mir/input/mousekey_pointer.h"
```
tarek-y-ismail, line 18: Should be moved below

tarek-y-ismail (commented):
Self re-re-review

src/server/shell/basic_accessibility_manager.cpp
AlanGriffiths:
```suggestion
    if (mousekeys_on || ssc_on)
```

examples/miral-shell/shell_main.cpp
tarek-y-ismail: Still want to showcase the callbacks somehow though...

AlanGriffiths (changes requested):
I think there's an architectural problem which already landed with `MouseKeysConfig`.

1. We have a bunch of input devices managed by the `the_default_input_device_hub()`;
2. that feeds `the_seat()`; and,
3. `the_seat()` feeds `the_input_dispatcher()`

The current approach is to take events from `the_input_dispatcher()` and transform them into events from a virtual device that then feeds them to the `the_default_input_device_hub()`.

It would make a lot more sense for any event sequence transformation to take place between the `the_default_input_device_hub()` and `the_seat()`. An identity transform would preserve existing behaviour and there's no possible confusion from events being transformed multiple times

include/miral/miral/simulated_secondary_click_config.h
```diff
@@ -0,0 +1,80 @@
...
+namespace mir
+{
+class Server;
+}
+
+namespace miral
+{
+/// Enables configuring simulated secondary click at runtime.
+/// \remark Since MirAL 5.3
+class SimulatedSecondaryClickConfig
```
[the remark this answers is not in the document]
tarek-y-ismail: Already created #3900 to fix things up once this lands.

src/miral/mousekeys_config.cpp
```diff
@@ -137,6 +137,8 @@ void miral::MouseKeysConfig::operator()(mir::Server& server) const
                 options->get<double>(mouse_keys_acceleration_linear_factor),
                 options->get<double>(mouse_keys_acceleration_quadratic_factor));
             set_max_speed(options->get<double>(mouse_keys_max_speed_x), options->get<double>(mouse_keys_max_speed_y));
-            enabled(enable);
+
+            if(enable)
+                enabled(true);
```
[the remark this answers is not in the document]
tarek-y-ismail: I think this was a quick fix for crashes caused by double removal of the virtual device, which should be handled in `MousekeyPointer` now properly.

src/include/server/mir/input/mousekey_pointer.h
```diff
@@ -0,0 +1,55 @@
...
+namespace mir
+{
+class MainLoop;
+namespace input
+{
+class InputEventTransformer;
+class InputDeviceRegistry;
+class VirtualInputDevice;
+class Device;
+class MousekeyPointer : public EventFilter
```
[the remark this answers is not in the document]
tarek-y-ismail: Some suggestions I have in mind `AccessibilityEvent{Filter,Transformer}`. @mattkae Have anything in mind?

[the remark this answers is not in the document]
mattkae: I am curious if I understand the architecture her correctly:

- `MousekeyPointer` is a gate at the `Seat` level that will ask the `InputEventTransformer` if an event can be let through, or if it is "handled" by the transformer
- The `BasicAccessibilityManager` controls whether or not `MousekeyPointer` is enabled by adding/removing the virtual device from the registry
- When enabled, it calls into the `InputEventTransformer` which will then ask one of its `Transformer`s if it handles the event. In this case, it would ask the `SimulatedSecondaryClickTransformer`
- If the `Transformer` does handle the event, it is not passed along to the `InputStateTracker` but it _may_ build a new event and add that onto the `InputStateTracker` with the virtual device id.

Do I have that correct?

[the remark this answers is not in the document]
mattkae: Or maybe `AccessibilityEventFilter` just implements both interfaces?

src/server/input/mousekey_pointer.cpp
```diff
@@ -0,0 +1,67 @@
...
+mir::input::MousekeyPointer::MousekeyPointer(
+    std::shared_ptr<MainLoop> main_loop, std::shared_ptr<input::InputEventTransformer> iet) :
+    main_loop{std::move(main_loop)},
+    iet{std::move(iet)},
+    virtual_device{
+
+        std::make_shared<mir::input::VirtualInputDevice>("mousekey-pointer", mir::input::DeviceCapability::pointer)}
+
+{
+}
```
mattkae, line 34:
```suggestion
    virtual_device{
        std::make_shared<mir::input::VirtualInputDevice>("mousekey-pointer", mir::input::DeviceCapability::pointer)}
{
}
```

mattkae (changes requested):
I have some more architectural questions that might involve Alan :)

AlanGriffiths:
> I remember trying out something similar before. If you mean making all transformations synchronous by ditching the virtual input device and just switching the event that gets dispatched down the chain if it gets transformed, then this doesn't quite fit with what we have now. A bunch of our transformations are asynchronous, so they happen at some point that might not align with incoming input events.

I didn't say "make all transformations synchronous", but otherwise yes.

> For example, simulated secondary click is issued after some delay following the initial left click down event. If we force things to be synchronous, then that obviously will not work at all.

Agreed, that's why I didn't say "make all transformations synchronous"

> To prevent transformed (synthesized) events from feeding back into transformers, `SeatInputDeviceTracker` could be used. Going any higher in the call stack puts us in `RegisteredDevice::handle_input` and platform input code. Going any lower and some state changes will not occur.

That argument implies that a seat should be responsible for transformations?

> The best hackaround I can come up with right now which would fit the current code would be to create a seat wrapper that forwards most existing calls to `BasicSeat` (or whatever seat it receives in the constructor), and only intercepts `handle_event`. I believe this doesn't much though.

That is the architecture I was thinking of: that in place `the_seat()`, the input hub is constructed with `the_input_transformer()` (which, in turn, is constructed with `the_seat()`). Note that there's nothing the stopping the InputTransformer (or whatever name) creating events (or even devices) asynchronously from the input hub.

----

I don't have a strong opinion (yet, that may change when I see code) on whether the transformer should sit between the hub and the seat, or whether it is a role on the seat.

However, I don't think the transformer should be feeding events back into the hub. There are just two many cases that affect multiple components to consider when making changes for that to be a good design.

One hypothetical to consider is "what happens when we have multiple seats?" (This is a usecase we've been asked about a few times.) Having all the transformation implementation "downstream" of the hub makes that a lot easier to reason about.

src/server/input/input_event_transformer.cpp
```diff
@@ -75,9 +101,14 @@ void mi::InputEventTransformer::append(std::weak_ptr<mi::InputEventTransformer::
             });
 
     if (duplicate_iter != input_transformers.end())
-        return;
+        return false;
 
     input_transformers.push_back(transformer);
+
+    if(weak_device.expired())
+        weak_device = input_device_registry->add_device(virtual_device);
```
mattkae, line 109: I think you can avoid a lot of complication just by not adding and removing the `virtual_device` when the list of transformers is empty. What is the harm in keeping it around until we next need it?

src/server/shell/basic_accessibility_manager.h
```diff
@@ -49,17 +45,20 @@ class Clock;
...
-        std::shared_ptr<input::CompositeEventFilter> the_composite_event_filter,
         std::shared_ptr<input::InputEventTransformer> const& event_transformer,
         bool enable_key_repeat,
         std::shared_ptr<mir::graphics::Cursor> const& cursor,
         std::shared_ptr<shell::MouseKeysTransformer> const& mousekeys_transformer,
-        std::shared_ptr<input::InputDeviceRegistry> const& input_device_registry);
+        std::shared_ptr<SimulatedSecondaryClickTransformer> const& simulated_secondary_click_transformer,
+        std::shared_ptr<input::InputDeviceRegistry> const& input_device_hub);
+
+    ~BasicAccessibilityManager();
```
mattkae, line 61:
```suggestion
    ~BasicAccessibilityManager() override;
```

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -26,10 +28,15 @@
...
+class MainLoop;
 namespace input
 {
 class EventBuilder;
-class InputEventTransformer
+class VirtualInputDevice;
+class InputDeviceRegistry;
+class Device;
+class Seat;
+class InputEventTransformer : public EventFilter
```
mattkae, line 39: Maybe I said this before - but should this be renamed to `AccessibilityEventFilter`? It isn't just a generic input transformer as far as I can see?

src/server/input/default_configuration.cpp
```diff
@@ -319,18 +319,23 @@ std::shared_ptr<mi::DefaultInputDeviceHub> mir::DefaultServerConfiguration::the_
...
-               the_led_observer_registrar());
+               the_led_observer_registrar(),
+               event_transformer);
 
            // lp:1675357: KeyRepeatDispatcher must be informed about removed input devices, otherwise
            // pressed keys get repeated indefinitely
            if (key_repeater)
                key_repeater->set_input_device_hub(hub);
+
+           event_transformer->init(hub);
```
mattkae, line 338: I see why you're doing this, but typically this is a sign of a circular abstraction that isn't ideal.

If we stop adding/removing the virtual pointer constantly, we could encapsulate this in the `DefaultInputDeviceHub`'s constructor instead with `hub->add_device(event_transformer->get_device())` or something similar

mattkae (changes requested):
A few ideas :)

src/server/input/default_configuration.cpp
tarek-y-ismail: I'm all for simplicity, but this would return us to the problems we had when we were adding the virtual device during server initialization (see https://github.com/canonical/mir/issues/3854).

src/include/server/mir/input/input_event_transformer.h
tarek-y-ismail: Will do once I get Alan's review in just in case he has another idea

src/server/input/default_configuration.cpp
```diff
@@ -319,18 +319,23 @@ std::shared_ptr<mi::DefaultInputDeviceHub> mir::DefaultServerConfiguration::the_
        {
            auto input_dispatcher = the_input_dispatcher();
            auto key_repeater = std::dynamic_pointer_cast<mi::KeyRepeatDispatcher>(input_dispatcher);
+           auto seat = the_seat();
```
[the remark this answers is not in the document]
tarek-y-ismail: From memory, I was experimenting with changing up the seat before I arrived at what's implemented at the moment, and I was a bit worried about the seat being destroyed between invocations of `the_seat` so I kept a reference to extend its lifetime, but that's unnecessary now. Thanks for catching it!

src/server/input/default_configuration.cpp
AlanGriffiths: As [stated before](https://github.com/canonical/mir/pull/3873#issuecomment-2830116800), I don't think we should be adding a virtual device to the hub at all.

tarek-y-ismail: I'm okay with ditching the virtual device. The only issue with that is that I don't want to reuse the source event device ID to avoid issues and/or confusion later on. For example, in the case of mousekeys, the source device is a keyboard, but the transformed events are mouse events so they shouldn't use that ID.

Right now, we don't even feedback events to the start of the chain, so the only use of the virtual device is reserving a device ID and generating events (which we could do without a virtual device):
https://github.com/canonical/mir/pull/3873/files#diff-bf5b6c1899a7fb166dbbc1472a01718321167d3ed64fedd9036f99522e5c8baeR50

mattkae (changes requested):
Some threads that are still open. Re-requesting to point them out :)

src/server/input/input_event_transformer.cpp
tarek-y-ismail: Why keep it around when it's not needed? The code overhead is not much as far as I can see, just checking when adding or removing a transformer, which shouldn't occur much anyway

src/server/input/mousekey_pointer.cpp
```diff
@@ -0,0 +1,67 @@
...
+            handled = iet->transform(
+                event,
+                builder,
+                [this, sink](auto event) { main_loop->spawn([sink, event] { sink->handle_input(event); }); },
+                state.lock()->device_id);
+        });
+    return handled;
+}
+
+void mir::input::MousekeyPointer::add_to_registry(std::shared_ptr<InputDeviceRegistry> const& registry)
```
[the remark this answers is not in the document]
tarek-y-ismail: Forgot to delete this!

tarek-y-ismail:
@AlanGriffiths Out of curiosity I tried removing the virtual device and all the fluff that comes with it. Passing a random device id (-1) to `mir::events::make_pointer_event` causes a crash here:
https://github.com/canonical/mir/blob/561c307440b4928dc340f019e9a1fdc266fbd728/src/server/input/seat_input_device_tracker.cpp#L130-L135

Which means we can either use the device id from the source event (which works, but is kinda hacky), or register a virtual device and continue on as is.

src/server/shell/basic_simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+                                // Consume the upcoming _real_ up event
+                                state->state = State::waiting_for_ssc_end_left_up;
+                            }
+                            click(dispatcher, builder, mir_pointer_button_secondary);
+                        });
+                }
+
+                consumed_left_down = std::unique_ptr<MirPointerEvent>(pointer_event->clone());
+                secondary_click_dispatcher->reschedule_in(mutable_state.lock()->hold_duration);
```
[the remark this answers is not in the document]
AlanGriffiths: Yes, both involve setting alarms, but I have a feeling that the idle hub has a different intent and it would be confusing to conflate them

include/miral/miral/simulated_secondary_click.h
```diff
@@ -0,0 +1,80 @@
...
+    SimulatedSecondaryClick& hold_duration(std::chrono::milliseconds hold_duration);
+
+    /// Configures the maximum displacement the mouse pointer can move before
+    /// the simulated secondary click is cancelled.
+    SimulatedSecondaryClick& displacement_threshold(float displacement);
+
+    /// Configures the callbacks invoked when simulated secondary click is
+    /// enabled or disabled
+    SimulatedSecondaryClick& on_enabled(std::function<void()>&& on_enabled);
+    SimulatedSecondaryClick& on_disabled(std::function<void()>&& on_disabled);
```
AlanGriffiths, line 59: The generated docs won't understand that the comment applies to both methods.

Also, what is the purpose of providing this function in the API? The only use is internal:

```
            if (auto& on_enabled_ = self->state.lock()->on_enabled)
                on_enabled(std::move(on_enabled_));
```

AlanGriffiths (changes requested):
Not really finished reviewing, but there's a few details to pick at

include/miral/miral/simulated_secondary_click.h
```diff
@@ -0,0 +1,80 @@
...
+namespace miral
+{
+/// Enables configuring simulated secondary click at runtime.
+/// \remark Since MirAL 5.3
+class SimulatedSecondaryClick
+{
+public:
+    /// \note `--enable-simulated-secondary-click` has higher precedence than
+    /// [enabled_by_default]
+    SimulatedSecondaryClick(bool enabled_by_default);
```
mattkae, line 38:
```suggestion
    explicit SimulatedSecondaryClick(bool enabled_by_default);
```

src/miral/simulated_secondary_click.cpp
```diff
@@ -0,0 +1,195 @@
...
+    Self(bool enabled_by_default) :
+        state{State{enabled_by_default}}
+    {
+    }
+
+    std::weak_ptr<mir::shell::AccessibilityManager> accessibility_manager;
+
+    struct State
+    {
+        State(bool enabled) :
```
mattkae, line 36:
```suggestion
        explicit State(bool enabled) :
```

src/server/shell/basic_simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,229 @@
...
+            if (action == mir_pointer_action_button_down && (pointer_event->buttons() & mir_pointer_button_primary))
+            {
+                if (!secondary_click_dispatcher)
+                {
+                    secondary_click_dispatcher = main_loop->create_alarm(
+                        [dispatcher, builder, this]
+                        {
+                            {
+                                auto const state = mutable_state.lock();
+                                state->on_secondary_click();
```
mattkae, line 89: Nit: Should the callback happen after the click has been dispatched?

mattkae (commented):
Some minor things, but nothing too concerning on my end

src/server/shell/basic_simulated_secondary_click_transformer.cpp
```diff
@@ -0,0 +1,229 @@
...
+        {
+            if (action == mir_pointer_action_motion)
+            {
+                auto const motion_threshold_squared = std::pow(mutable_state.lock()->displacement_threshold, 2);
+                auto const displacement =
+                    (pointer_event->position().value_or(initial_position) - initial_position).length_squared();
+                if (displacement <= motion_threshold_squared)
+                    return true;
+
+                auto const state = mutable_state.lock();
```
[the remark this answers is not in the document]
tarek-y-ismail: I just tried again and it works fine.

Originally I had some deadlock issues, hence the frequent locking and unlocking.

src/server/shell/basic_simulated_secondary_click_transformer.cpp
tarek-y-ismail: Nice catch!

src/server/shell/basic_accessibility_manager.h
```diff
@@ -67,6 +69,9 @@ class BasicAccessibilityManager : public AccessibilityManager
     void acceleration_factors(double constant, double linear, double quadratic) override;
     void max_speed(double x_axis, double y_axis) override;
 
+    void simulated_secondary_click_enabled(bool enabled) override;
+    auto simulated_secondary_click() -> SimulatedSecondaryClickTransformer& override;
+
 private:
     class MousekeyPointer;
```
tarek-y-ismail, line 77: wart

tarek-y-ismail: For clarity, I mean the `class MousekeyPointer` line. I don't think it's a blocker though since I intend on doing a bunch of cleanup in a separate PR

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -32,7 +33,7 @@ inline Transformer* mir::shell::BasicAccessibilityManager::Registration<Transfor
 }
 
 template <typename Transformer>
-inline void mir::shell::BasicAccessibilityManager::Registration<Transformer>::remove_registration() const
+inline bool mir::shell::BasicAccessibilityManager::Registration<Transformer>::remove_registration() const
```
AlanGriffiths, line 36: Why change the signature? The return value isn't useful, nor used

AlanGriffiths: Correction: it is used in `BasicAccessibilityManager::simulated_secondary_click_enabled`

include/miral/miral/simulated_secondary_click.h
```diff
@@ -0,0 +1,83 @@
...
+    SimulatedSecondaryClick& on_hold_start(std::function<void()>&& on_hold_start);
+
+    /// Configures the callback to invoke when the user cancels a simulated
+    /// secondary click either by letting go of the left button before the
+    /// hold duration is up or moving the cursor
+    SimulatedSecondaryClick& on_hold_cancel(std::function<void()>&& on_hold_cancel);
+
+    /// Configures the callback to invoke after the user successfully
+    /// dispatches a simulated secondary click
+    SimulatedSecondaryClick& on_secondary_click(std::function<void()>&& on_secondary_click);
```
AlanGriffiths, line 75: These comments document how the API is implemented, not the the problem it solves: When would a shell author use each of these `on_XXX` callbacks?

tarek-y-ismail: You mean something like the following?
```cpp
    /// Configures the callback invoked when simulated secondary click is
    /// enabled. 
    /// Usage: To observe the state of the transformer for GUI settings applications and the like.
    SimulatedSecondaryClick& on_enabled(std::function<void()>&& on_enabled);
    
    ...
    
    /// Configures the callback to invoke when the user clicks _down_ the left
    /// mouse button
    /// Usage: To provide graphical feedback to the user.
    SimulatedSecondaryClick& on_hold_start(std::function<void()>&& on_hold_start)
```

AlanGriffiths (changes requested):
I think implementation details are leaking

include/miral/miral/simulated_secondary_click.h
tarek-y-ismail: Ok, I see what you meant now.

SSC could be enabled based on multiple things: The existing launch config option, a shortcut, a GUI application, a config file.

This is an issue currently as well with mousekeys, if the launch config option doesn't agree with the assumed default (`false` or disabled), then the first toggle doesn't work properly. The current interface doesn't provide a way to query whether mousekeys is active or not. I discovered this by accident a bit ago and it's one of the things I aim to fix when I go back and clean mousekeys.

Anyway, The same could apply to SSC. In the case of a shortcut and a GUI configuration application: When the shortcut toggles SSC, the GUI application needs to be informed somehow. `on_{enabled,disabled}` should allow shell authors to do whatever they need to propagate the changes.

I agree that not calling the passed callback to provide the initial state of the transformer was an oversight on my part. But I hope that the examples above justify our need for the callback.

For the transformer to be able to call the callbacks at the right time, I think we'd need to push `{enabled,disabled}()` to the base transformer interface. In addition, we also need to add `is_enabled()` to query the state of the transformer for registration purposes. This also makes `Registration<T>` useless.

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -58,7 +74,7 @@ class InputEventTransformer : public Seat
         EventBuilder* builder,
         EventDispatcher const& dispatcher);
 
-    void append(std::weak_ptr<Transformer> const&);
+    bool append(std::weak_ptr<Transformer> const&);
     bool remove(std::shared_ptr<Transformer> const&);
```
tarek-y-ismail, line 78: Both should return `void` now.

examples/miral-shell/shell_main.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "miral/minimal_window_manager.h"
+#include "miral/simulated_secondary_click.h"
```
tarek-y-ismail, line 18: Should be moved below and use angled brackets.

src/server/shell/mouse_keys_transformer.h
```diff
@@ -122,6 +122,9 @@ class BasicMouseKeysTransformer: public MouseKeysTransformer
         AccelerationCurve acceleration_curve{30, 100, 100};
         geometry::DisplacementF max_speed_{400, 400};
         mir::input::MouseKeysKeymap keymap_;
+
+        bool enabled;
+        std::function<void()> on_enabled{[]{}}, on_disabled{[]{}};
```
tarek-y-ismail, line 127: Redundant

AlanGriffiths (changes requested):
Continuing the API discussion (no point in discussing the implementation until that's settled)

> Ok, I see what you meant now.
> 
> SSC could be enabled based on multiple things: The existing launch config option, a shortcut, a GUI application, a config file.

Most of those would lead to a call, by the shell, to `enabled()`. The only exception being the config option - and that is not necessarily a good idea as this ought to be a "live" configuration option, not a fixed one.

Note that we had similar discussions around `miral::InputConfiguration`. I know it is a pain that we don't have a canonical "live configuration" implementation yet.

> This is an issue currently as well with mousekeys, if the launch config option doesn't agree with the assumed default (`false` or disabled), then the first toggle doesn't work properly. The current interface doesn't provide a way to query whether mousekeys is active or not. I discovered this by accident a bit ago and it's one of the things I aim to fix when I go back and clean mousekeys.

Yes, there is the same problem with MouseKeys. That doesn't mean we should [continue digging](https://en.wikipedia.org/wiki/Law_of_holes).

> Anyway, The same could apply to SSC. In the case of a shortcut and a GUI configuration application: When the shortcut toggles SSC, the GUI application needs to be informed somehow. `on_{enabled,disabled}` should allow shell authors to do whatever they need to propagate the changes.

Unless we have a shortcut implementation (for example) to toggle SSC _without the shell being informed_ we don't need this complexity. An alternative design would be the "shortcut" to request the shell make the change.

[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

> I agree that not calling the passed callback to provide the initial state of the transformer was an oversight on my part. But I hope that the examples above justify our need for the callback.

I hope the above discussion clarifies why I think this is not needed (yet)

src/miral/simulated_secondary_click.cpp
```diff
@@ -0,0 +1,170 @@
...
+    {
+        explicit State(bool enabled) :
+            enabled{enabled}
+        {
+        }
+
+        bool enabled;
+        std::chrono::milliseconds hold_duration{1000};
+        float displacement_threshold{20.0f};
+        std::function<void()> on_hold_start, on_hold_cancel, on_secondary_click;
```
AlanGriffiths, line 44: We tend not to define multiple variables in a statement.
Initializing these with an empty implementation would avoid the need to test for "callable".

```suggestion
        std::function<void()> on_hold_start{[]{}};
        std::function<void()> on_hold_cancel{[]{}};
        std::function<void()> on_secondary_click{[]{}};
```

src/miral/simulated_secondary_click.cpp
```diff
@@ -0,0 +1,170 @@
...
+    return *this;
+}
+
+miral::SimulatedSecondaryClick& miral::SimulatedSecondaryClick::on_hold_start(
+    std::function<void()>&& on_hold_start)
+{
+    auto const state = self->state.lock();
+    state->on_hold_start = std::move(on_hold_start);
+    if (auto const accessibility_manager = self->accessibility_manager.lock())
+        accessibility_manager->simulated_secondary_click().on_hold_start(std::move(state->on_hold_start));
```
AlanGriffiths, line 98:
```suggestion
    if (auto const accessibility_manager = self->accessibility_manager.lock())
    {
        accessibility_manager->simulated_secondary_click().on_hold_start(std::move(on_hold_start));
    }
    else
    {
        self->state.lock()->on_hold_start = std::move(on_hold_start);
    }
```

include/miral/miral/simulated_secondary_click.h
```diff
@@ -0,0 +1,79 @@
...
+    /// When already enabled, further calls have no effect.
+    SimulatedSecondaryClick& enable();
+
+    /// Disables simulated secondary click.
+    /// When already disabled, further calls have no effect.
+    SimulatedSecondaryClick& disable();
+
+    /// Configures the duration users have to keep the left mouse button down
+    /// to dispatch a secondary click.
+    SimulatedSecondaryClick& hold_duration(std::chrono::milliseconds hold_duration);
```
AlanGriffiths, line 52: [non-blocking thought] We could avoid the need for armhf symbols by taking a raw int. (That's consistent with, for example, `displacement_threshold()`)

tarek-y-ismail:
One more rebase....

tarek-y-ismail:
Think this is the most I can compress the commit history without squashing unrelated things together

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -102,39 +64,70 @@ void mir::shell::BasicAccessibilityManager::repeat_rate_and_delay(
...
 {
-    if (on)
-        transformer.add_registration();
-    else
-        transformer.remove_registration();
+    if (on && !mouse_keys_transformer->is_enabled())
+    {
+        event_transformer->append(mouse_keys_transformer);
+        mouse_keys_transformer->enable();
+    }
```
AlanGriffiths, line 71: This is racy: Two threads can call `is_enabled()` before either calls `enable()` resulting in multiple calls to `append()`

The previous `compare_exchange_strong()` implementation avoided this

src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -102,39 +64,70 @@ void mir::shell::BasicAccessibilityManager::repeat_rate_and_delay(
...
+    mouse_keys_transformer->max_speed(x_axis, y_axis);
+}
+
+void mir::shell::BasicAccessibilityManager::simulated_secondary_click_enabled(bool enabled)
+{
+    if (enabled && !simulated_secondary_click_transformer->is_enabled())
+    {
+        event_transformer->append(simulated_secondary_click_transformer);
+        simulated_secondary_click_transformer->enable();
+    }
```
AlanGriffiths, line 121: This is racy: Two threads can call `is_enabled()` before either calls `enable()` resulting in multiple calls to `append()`

AlanGriffiths (changes requested):
Thread safety needs care

src/include/server/mir/input/input_event_transformer.h
```diff
@@ -58,8 +58,8 @@ class InputEventTransformer : public Seat
         EventBuilder* builder,
         EventDispatcher const& dispatcher);
 
-    void append(std::weak_ptr<Transformer> const&);
-    bool remove(std::shared_ptr<Transformer> const&);
+    virtual void append(std::weak_ptr<Transformer> const&);
+    virtual void remove(std::shared_ptr<Transformer> const&);
```
tarek-y-ismail, line 62: Could either expose the interface, or just make these two function virtual so I can override them in tests. I chose the easier solution (this surely will not backfire)

tests/unit-tests/input/test_input_event_transformer.cpp
```diff
@@ -83,11 +83,14 @@ struct TestInputEventTransformer : testing::Test
 
 struct MockTransformer : public mir::input::InputEventTransformer::Transformer
 {
+    MockTransformer() = default;
+
```
AlanGriffiths, line 87:
```suggestion
```

tests/unit-tests/input/test_input_event_transformer.cpp
```diff
@@ -83,11 +83,14 @@ struct TestInputEventTransformer : testing::Test
...
     MOCK_METHOD(
         (bool),
         transform_input_event,
         (mir::input::InputEventTransformer::EventDispatcher const&, mir::input::EventBuilder*, MirEvent const&),
         (override));
+
```
AlanGriffiths, line 93:
```suggestion
```

tests/unit-tests/shell/test_basic_accessibility_manager.cpp
```diff
@@ -245,3 +246,93 @@ TEST_F(TestBasicAccessibilityManager, calling_set_max_speed_calls_set_max_speed_
...
+TEST_F(TestBasicAccessibilityManager, setting_on_hold_start_sets_it_on_transformer)
+{
+    auto const expected_on_hold_start = []
+    {
+    };
+
+    EXPECT_CALL(*mock_simulated_secondary_click_transformer, on_hold_start(_));
+
+    basic_accessibility_manager.simulated_secondary_click().on_hold_start(expected_on_hold_start);
+}
```
AlanGriffiths, line 268: [Non blocking] This looks somewhat like testing implementation details, not the desired behavior.

We don't need to care that `mock_simulated_secondary_click_transformer, on_hold_start()` is called. If the interaction between `BasicAccessibilityManager` and `Transformer` were changed _without changing functionality_ then we'd need to update this test.

What we care about is that `expected_on_hold_start()` is called when the hold starts.

AlanGriffiths (changes requested):
Some whitespace tweaks and a warning about writing fragile tests.

(The warning is non-blocking and doesn't need addressing as it is a "beware of the slippery slope")

AlanGriffiths (approved):
At last ;)

AlanGriffiths:
Re-requesting a review from @mattkae as it is a lot of churn since his approval

mattkae (approved):
Let's land it :tada:

PR #3874, opened 2025-04-07
Screencasting how-to

tarek-y-ismail:


doc/sphinx/how-to/how-to-enable-screencasting.md
```diff
@@ -0,0 +1,51 @@
...
+chooser_type=simple
+chooser_cmd=slurp -f %o -or
+```
+
+Where `chooser_type=simple` runs the given `chooser_cmd`. `chooser_cmd` in this case should return the name of the output that's going to be shared. In the example above, `slurp -f %o -or` uses a transparent white overlay to indicate that its running, if you click on a specific output, then it will return the name of that output.
+
+3. The following script should run automatically when your compositor starts up. For testing purposes, you can run the script after you start your compositor.
+```sh
+export XDG_CURRENT_DESKTOP=my-mir-compositor                            # Refer to step 1
+export XDG_DESKTOP_PORTAL_DIR=/usr/share/xdg-desktop-portal/portals     # Tells `xdg-desktop-portal` where custom configurations are stored
```
mattkae, line 37: You should not have to set this. We did it in the snap because things in the snap are a bit spooky, but this variable purposefully circumvents how portals are supposed to be found (i.e `xdg-desktop-portal` will search this directory anyway :smile: )

doc/sphinx/how-to/how-to-enable-screencasting.md
```diff
@@ -0,0 +1,51 @@
...
+
+# DBus has its own separate environment. For the sake of hygiene, we only copy over `XDG_CURRENT_DESKTOP` and `WAYLAND_DISPLAY`
+VARIABLES="XDG_CURRENT_DESKTOP WAYLAND_DISPLAY"
+dbus-update-activation-environment --systemd $VARIABLES
+
+# Run `xdg-desktop-portal`, using the configuration above, it will take care of starting `xdg-desktop-portal-wlr`
+systemd-run --user --setenv=XDG_DESKTOP_PORTAL_DIR=${XDG_DESKTOP_PORTAL_DIR} /usr/libexec/xdg-desktop-portal --replace
```
mattkae, line 44: You shouldn't need the `--setenv` option there

mattkae: I also believe that if you `systemctl --user enable xdg-desktop-portal-wlr`, then it will be automatically started. But I have to double check if you have to take the systemd graphical session for this to work.

mattkae (changes requested):

doc/sphinx/how-to/how-to-enable-screencasting.md
```diff
@@ -0,0 +1,50 @@
...
+3. The following script should run automatically when your compositor starts up. For testing purposes, you can run the script after you start your compositor.
+```sh
+export XDG_CURRENT_DESKTOP=my-mir-compositor                            # Refer to step 1
+
+# DBus has its own separate environment. For the sake of hygiene, we only copy over `XDG_CURRENT_DESKTOP` and `WAYLAND_DISPLAY`
+VARIABLES="XDG_CURRENT_DESKTOP WAYLAND_DISPLAY"
+dbus-update-activation-environment --systemd $VARIABLES
+
+# Run `xdg-desktop-portal`, using the configuration above, it will take care of starting `xdg-desktop-portal-wlr`
+systemd-run --user /usr/libexec/xdg-desktop-portal --replace
```
mattkae, line 43: So this method of doing this works for me, but it would be very useful if we had instructions on how to automatically start xdg-desktop-portal-wlr as well.

https://github.com/emersion/xdg-desktop-portal-wlr/wiki/FAQ#how-do-i-run-xdpw should provide some info, but we may need to rename some things to get it to work

mattkae (changes requested):

doc/sphinx/how-to/how-to-enable-screencasting.md
Saviq: > how to automatically start xdg-desktop-portal-wlr as well.

I _think_ that's out of scope for this part. Different use cases will warrant different approaches here. We might link to existing solutions (Miriway, the link above), but likely don't want to get too much into the detail.

mattkae: I am outnumbered! This is okay by me then :+1:

mattkae (approved):
This is a good initial doc in my opinion :)

PR #3881, opened 2025-04-10
Fix leak of `LinuxDmaBufUnstable` resources

RAOF:
Fix `mf::WaylandConnector` to always call `GraphicsBufferAllocator::unbind_display`
on the Wayland eventloop (or directly, if the eventloop isn't running), and then
use this to clean up `LinuxDmaBufUnstable` resources synchronously rather than
deferring them to the `WaylandExecutor` which may never execute that work.


AlanGriffiths (approved):
Looks sensible, TiCS errors look like Coverity not understanding the language

PR #3882, opened 2025-04-10
mg::SoftwareCursor: Synchronously clean up in destructor

RAOF:
During `hide()` we defer work to the `scene_executor` in order to avoid calling into the scene while holding a lock, but in the destructor we don't need to *hold* a lock, so just call in directly.

Resolves a leak of `cursor_renderable` during teardown, where the `scene_executor` may no longer be dispatching tasks, resulting in the cleanup never being performed.


RAOF:
Replaced by #3886.

PR #3883, opened 2025-04-10
Fix deadlock in `BasicAccessibilityManager::notify_helpers`

tarek-y-ismail:
The deadlock was caused by the for loop extending the lifetime of the state lock to cover the duration of the whole loop. In the loop, we also locked the state to read some data, which would cause the deadlock.

Why did this decide to come up now? I have no idea, it used to work before (and should've failed in CI), but now it deadlocks...

### DIFF
--- a/src/server/shell/basic_accessibility_manager.cpp
+++ b/src/server/shell/basic_accessibility_manager.cpp
@@ -51,7 +51,8 @@ void mir::shell::BasicAccessibilityManager::repeat_delay(int new_delay) {
 }
 
 void mir::shell::BasicAccessibilityManager::notify_helpers() const {
-    for (auto const& helper: mutable_state.lock()->keyboard_helpers)
+    auto const keyboard_helpers = mutable_state.lock()->keyboard_helpers;
+    for (auto const& helper: keyboard_helpers)
         helper->repeat_info_changed(repeat_rate(), repeat_delay());
 }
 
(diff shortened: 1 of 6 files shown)


src/server/shell/basic_accessibility_manager.cpp
```diff
@@ -51,7 +51,8 @@ void mir::shell::BasicAccessibilityManager::repeat_delay(int new_delay) {
 }
 
 void mir::shell::BasicAccessibilityManager::notify_helpers() const {
-    for (auto const& helper: mutable_state.lock()->keyboard_helpers)
+    auto const keyboard_helpers = mutable_state.lock()->keyboard_helpers;
+    for (auto const& helper: keyboard_helpers)
         helper->repeat_info_changed(repeat_rate(), repeat_delay());
 }
```
AlanGriffiths, line 57: Isn't this the wrong change? What should the lock be protecting? Should we be locking and unlocking multiple times in a method call like this?

During the iteration we already have a lock on `mutable_state` giving safe access to `repeat_rate` and `repeat_delay`. The only problem is that it is an unnamed temporary, so extracting it seems better:

```c++
void mir::shell::BasicAccessibilityManager::notify_helpers() const {
    auto const state = mutable_state.lock();
    for (auto const& helper: state->keyboard_helpers)
        helper->repeat_info_changed(state.repeat_rate, repeat_delay);
}
```

AlanGriffiths (changes requested):
Something like my suggestion is probably good for a local fix. (Or putting a `notify_helper()` method directly on `State`)

However, calling out to other objects whilst holding a lock is likely to cause an issue at some point. We've been there a few times, I even wrote it up once:

https://www.accu.org/journals/overload/22/124/griffiths_2040/

PR #3886, opened 2025-04-11
mg::SoftwareCursor: Don't `hide()` on destruction

RAOF:
We only ever destroy `the_cursor()` during teardown, and
*during* teardown work deferred to the `GLibMainLoop` can
get leaked; in this case resulting in the cursor `Renderable`
being leaked (along with the `Scene`).

Since we don't need to release any *hardware* resources,
just don't try to hide the cursor; we're going to stop
drawing it immediately anyway!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .idea
 .project
+.cache
 benchmarks/mir_perf_framework_setup.py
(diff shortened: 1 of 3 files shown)


AlanGriffiths (approved):
OK

PR #3887, opened 2025-04-11
platforms/generic-egl: Correct order of EGL resource destruction

RAOF:
We want `EGLTerminate` to be called *last*, but there are a bunch of member variables that will try cleaning up EGL resources in *their* destructors.

Move `EGLDisplay` handling into a trivial class with its own destructor so we can rely on member ordering to enforce correct destruction order.

### DIFF
--- a/src/platforms/renderer-generic-egl/rendering_platform.h
+++ b/src/platforms/renderer-generic-egl/rendering_platform.h
@@ -49,8 +49,20 @@ protected:
 private:
     explicit RenderingPlatform(std::tuple<EGLDisplay, bool> dpy);
 
-    EGLDisplay const dpy;
-    bool const owns_dpy;
+    class EGLDisplayHandle
+    {
+    public:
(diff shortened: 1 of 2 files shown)


src/platforms/renderer-generic-egl/rendering_platform.h
```diff
@@ -49,8 +49,20 @@ class RenderingPlatform : public graphics::RenderingPlatform
 private:
     explicit RenderingPlatform(std::tuple<EGLDisplay, bool> dpy);
 
-    EGLDisplay const dpy;
-    bool const owns_dpy;
+    class EGLDisplayHandle
```
mattkae, line 52: Nit: I would consider leaving a comment on this class explaining its existence so that future readers do not think it to be some unnecessary abstraction

mattkae (approved):
Reasonable enough to me!

PR #3891, opened 2025-04-14
Fix TOCTTOU bug in `MouseKeysConfig` setters

### LINKED ISSUE #3890: `MouseKeysConfig` possible TOCTTOU bug.
Similar to https://github.com/canonical/mir/pull/3873#discussion_r2038965655, `MouseKeysConfig` also suffers from the same TOCTTOU bug.

The following code block and friends should be updated.
https://github.com/canonical/mir/blob/536de0fb8c00cf00ce122f19a4989963516d0872/src/miral/mousekeys_config.cpp#L50-L59

tarek-y-ismail:
Closes #3890


mattkae (approved):
Aha! Nice catch :)

PR #3897, opened 2025-04-15
Locate pointer core implementation

tarek-y-ismail:
Still missing testing (will have to pull the changes required for that from #3873 into its own PR), and documentation (once the implementation settles).


tests/miral/test_mousekeys_config.cpp
```diff
@@ -27,21 +26,6 @@
...
-    MOCK_METHOD(int, repeat_delay, (), (const override));
-    MOCK_METHOD(void, repeat_rate_and_delay, (std::optional<int> new_rate, std::optional<int> new_delay), (override));
-    MOCK_METHOD(void, notify_helpers, (), (const override));
-    MOCK_METHOD(void, cursor_scale, (float), (override));
-    MOCK_METHOD(void, mousekeys_enabled, (bool on), (override));
-    MOCK_METHOD(void, mousekeys_keymap, (mir::input::MouseKeysKeymap const& new_keymap), (override));
-    MOCK_METHOD(void, acceleration_factors, (double constant, double linear, double quadratic), (override));
-    MOCK_METHOD(void, max_speed, (double x_axis, double y_axis), (override));
-};
-
```
tarek-y-ismail, line 44: Thought I needed the mock accessibility manager but turns out that I don't. So these changes should be reverted.

tests/include/mir/test/doubles/mock_accessibility_manager.h
tarek-y-ismail, line 1: Unneeded for this PR

tests/miral/test_locate_pointer.cpp
```diff
@@ -0,0 +1,249 @@
...
+    locate_pointer_invoked.wait_for(delay);
+    test_device->if_started_then(
+        [up_event_time](auto* sink, auto* builder)
+        {
+            auto up_event = builder->key_event(up_event_time, mir_keyboard_action_up, XKB_KEY_Control_L, 29);
+            sink->handle_input(std::move(up_event));
+        });
+
+    EXPECT_THAT(locate_pointer_invoked.raised(), Eq(on_pointer_locate_called));
+}
```
tarek-y-ismail, line 100: Unfortunately, this event can't be dispatched in the previous call to `if_started_then`

tests/miral/test_locate_pointer.cpp
```diff
@@ -0,0 +1,249 @@
...
+    EXPECT_THAT(locate_pointer_y, Optional(expected_position.y.as_value()));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    TestLocatePointer, TestPointerMovementTracking, ::Values(std::nullopt, mir::geometry::PointF{100.0f, 100.0f}));
+
+TEST_F(TestLocatePointer, enable_not_called_when_already_enabled)
+{
+    mir::test::Signal enable_called;
+    locate_pointer.on_enabled([&enable_called] { enable_called.raise(); });
```
tarek-y-ismail, line 154: Could reuse `locate_pointer_enabled`

tests/miral/test_locate_pointer.cpp
```diff
@@ -0,0 +1,249 @@
...
+            return true;
+        });
+
+    test_device->if_started_then(
+        [](auto* sink, auto* builder)
+        {
+            sink->handle_input(left_ctrl_down_event(builder));
+            sink->handle_input(builder->key_event(std::nullopt, mir_keyboard_action_down, XKB_KEY_Alt_L, 56));
+            sink->handle_input(builder->key_event(std::nullopt, mir_keyboard_action_down, XKB_KEY_Shift_L, 42));
+            sink->handle_input(builder->key_event(std::nullopt, mir_keyboard_action_down, XKB_KEY_p, 25));
```
tarek-y-ismail, line 244: Do we have an automated way to get the scan code corresponding to a keysym?

tests/miral/test_mousekeys_config.cpp
tarek-y-ismail, line 1: Unnecessary

tests/unit-tests/shell/test_basic_accessibility_manager.cpp
tarek-y-ismail, line 1: ???

tarek-y-ismail (commented):
Self review

src/miral/locate_pointer.cpp
```diff
@@ -0,0 +1,240 @@
...
+                break;
+            default:
+                break;
+            }
+            return false;
+        }
+
+        void handle_key_combination(MirInputEvent const* input_event)
+        {
+            if (!state->lock()->key_combination(input_event))
```
mattkae, line 80: This architecture is a bit constricting for the user. What if they want to trigger a pointer location in response to something that is _not_ a key press, or even in response to multiple key presses.

A more flexible solution would be to expose methods `LocatePointer::trigger_locate` and `LocatePointer::cancel_locate`

mattkae: Then, if the user wants to use a specific key combination, then they can just do it themselves

mattkae: In this case, `key_combination` can go away

mattkae (changes requested):
I have an architectural piece of feedback first

src/miral/locate_pointer.cpp
tarek-y-ismail: All done. Again, I'm not really in love with either approach. I'll leave evaluation of what's best to the team :)

In a related note, I considered adding `on_locate_pointer_cancelled` but decided against it since now users (shell authors) control when to schedule and cancel a callback.

src/miral/locate_pointer.cpp
```diff
@@ -0,0 +1,214 @@
...
+        state->locate_pointer_filter = std::make_shared<Self::LocatePointerFilter>(self->filter_state);
+        // Need to account that this can be called from an event filter, which
+        // when invoked has the composite filter's lock.
+        main_loop->spawn(
+            [composite_filter, self = this->self]
+            {
+                auto state = self->state.lock();
+                composite_filter->append(state->locate_pointer_filter);
+                self->filter_state->lock()->on_enabled();
+            });
```
mattkae, line 185: This is overly complicated :) I would:

1. Not recreate the `locate_pointer_filter` every time, and just make it once at construction.
2. Use the `enabled` flag (that we already have) in `LocateEventFilter::handle` to gate whether or not we need to update our state.

This way, we dont' have to keep recreating the shared ptr.

src/miral/locate_pointer.cpp
```diff
@@ -0,0 +1,214 @@
...
+        auto handle(MirEvent const& event) -> bool override
+        {
+            if (event.type() != mir_event_type_input)
+                return false;
+
+            auto const* input_event = event.to_input();
+            if (input_event->input_type() != mir_input_event_type_pointer)
+                return false;
+
+            record_pointer_position(input_event);
```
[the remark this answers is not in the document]
tarek-y-ismail: Yep, we have to keep track of the cursor's position somehow.

tests/miral/test_locate_pointer.cpp
```diff
@@ -0,0 +1,238 @@
...
+// Mostly to account for CI slowness.
+// Adds a test pointer and waits till it's added before proceeding
+std::shared_ptr<mir::input::VirtualInputDevice> add_test_pointer(
+    std::weak_ptr<mir::input::InputDeviceRegistry> const& input_device_registry,
+    std::weak_ptr<mir::input::InputDeviceHub> const& input_device_hub)
+{
+    std::shared_ptr<mir::input::VirtualInputDevice> test_device =
+        std::make_shared<mir::input::VirtualInputDevice>("test-device", mir::input::DeviceCapability::pointer);
+
+    struct SignalingObserver : public mir::input::InputDeviceObserver
```
mattkae, line 51: You should extend the `NullInputDeviceObserver` here and just override what you need

mattkae (changes requested):
A few more things, but its looking 95% there :)

src/server/shell/basic_accessibility_manager.h
```diff
@@ -49,6 +45,8 @@ class Clock;
 }
 namespace shell
 {
+class MouseKeysTransformer;
+class LocatePointerFilter;
```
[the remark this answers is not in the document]
tarek-y-ismail: Only the moving of `MouseKeysTransformer` was intentional. I initially planned to implement locate pointer as a transformer/filter here, but decided implementing it in miral would be better.

mattkae (changes requested):
One bug I found:

If I set the delay to something small (e.g. 1ms),  I get the notification twice

src/server/shell/basic_accessibility_manager.h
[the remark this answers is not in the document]
mattkae: Can you remove it?

tests/miral/test_locate_pointer.cpp
mattkae: Heh true

mattkae (changes requested):
I have one remaining thing! but it is good for me after that!

examples/miral-shell/shell_main.cpp
```diff
@@ -35,6 +36,8 @@
 #include <miral/wayland_extensions.h>
 #include <miral/mousekeys_config.h>
 
+#define MIR_LOG_COMPONENT "miral-shell"
+#include <mir/log.h>
```
AlanGriffiths, line 40: I don't think we should need to add logging

mattkae (approved):
I am happy, but I will give Alan the final word :)

include/miral/miral/locate_pointer.h
```diff
@@ -0,0 +1,65 @@
...
+    LocatePointer& on_disabled(std::function<void()>&&);
+
+    LocatePointer& enable();
+    LocatePointer& disable();
+
+    /// Schedules the callback provided to [on_locate_pointer] to be called
+    /// with the current mouse position at a point in the future specified by
+    /// [delay]. If [cancel_request] is called after this, then the callback
+    /// will not be called.
+    void schedule_request();
```
AlanGriffiths, line 52: Why do we need the compositor to schedule anything? That smells like something libmiral should hande

include/miral/miral/locate_pointer.h
```diff
@@ -0,0 +1,59 @@
...
+    LocatePointer& delay(std::chrono::milliseconds delay);
+    LocatePointer& on_locate_pointer_requested(std::function<void(float x, float y)>&&);
+    LocatePointer& on_enabled(std::function<void()>&&);
+    LocatePointer& on_disabled(std::function<void()>&&);
+
+    LocatePointer& enable();
+    LocatePointer& disable();
+
+    void schedule_request();
+    void cancel_request();
```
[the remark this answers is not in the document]
AlanGriffiths: Again, why?

examples/miral-shell/shell_main.cpp
```diff
@@ -172,6 +175,39 @@ int main(int argc, char const* argv[])
         return false;
     };
 
+    auto locate_pointer = miral::LocatePointer{false}
+                              .on_locate_pointer([](auto, auto) { mir::log_info("Locate pointer!"); })
```
AlanGriffiths, line 179: So the only thing this example does is spam the log with a fixed string? Is that a good example?

If writing to the console is the best we can do we should, at  least, be outputting some location information.

AlanGriffiths (changes requested):
The API looks overly complex.

What does the user need to know to combine `enable()`, `on_enable()`, `disable()`, `on_disable()`, `schedule_request()` and `cancel_request()` effectively?

I would expect `enable()` and `disable()` to be sufficient

PR #3903, opened 2025-04-18
static_display_config: add display properties to output

Saviq:
Adds to the default config:

```yaml
    # displays:
    # A list of display configurations matched by the displays' properties
    #
    # These take the same display options as above,
    # and take precedence over the port-based configuration.
    #
    # - vendor: BOE
    #   model: 
    #   product: 2399
    #   serial: 0
```


mattkae (approved):
Looks reasonable enough to me! Could we write some tests, or would that be hard with the edid stuff bits?

Saviq:
> Could we write some tests, or would that be hard with the edid stuff bits?

Added a couple tests for the added functionality.

Saviq:
> Added a couple tests for the added functionality.

And now actually passing…

PR #3907, opened 2025-04-22
tests/acceptance-tests/wayland/CMakeLists.txt: Reference expected_wlc…

sunweaver:
…s_failures.list with full-path relative to builddir, so that autopkgtest's cmake call succeeds in finding the file.

### DIFF
--- a/tests/acceptance-tests/wayland/CMakeLists.txt
+++ b/tests/acceptance-tests/wayland/CMakeLists.txt
@@ -42,7 +42,7 @@ endif()
 
 # Pull list of expected failures
 file(STRINGS
-  expected_wlcs_failures.list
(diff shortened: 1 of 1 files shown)


sunweaver:
Please cherry-pick this to a 2.20.x branch whenever this gets created... (which would be nice, so we can possibly directly cherry-pick fixes from such a branch to the Debian trixie version of Mir (shipping 2.20.2 at the moment).

Saviq:
@sunweaver since this is a packaging/build only change, we don't usually roll patch releases for that.

Doesn't this fit as a distro patch?

Saviq:
> The suggestion to branch off at Mir 2.20.2 was more like a general idea and the question if upstream can maintain a 2.20.x branch while heading towards Mir 2.22.x and beyond...

Ah. Yes, when we do roll a 2.20.2, we'll include this, sure.

PR #3911, opened 2025-04-24
Add an output filter for the purposes of accessibility. 

robert-ancell:
This is implemented in the OpenGL renderer by rendering to a framebuffer and then rendering this using a GLSL shader.

The default shells can enable this using the `--output-filter=invert` option.

Two filters are provided as a starting point (grayscale and invert). More useful filters will be added later (i.e. high contrast).


robert-ancell:
There's some pixelization occurring when the filter is enabled - I'm fixing that.

src/renderers/gl/renderer.cpp
```diff
@@ -642,6 +859,44 @@ void mrg::Renderer::set_output_transform(glm::mat2 const& t)
...
+        return;
+    case mir_output_filter_grayscale:
+        filter_src = grayscale_src;
+        break;
+    case mir_output_filter_invert:
+        filter_src = invert_src;
+        break;
+    }
+    output_filter_shader = std::make_unique<OutputFilterShader>(output_surface->size().width.as_value(), output_surface->size().height.as_value(), filter_src);
+}
```
Saviq, line 898: Should our API be that prescriptive? Shouldn't we, instead, accept custom shaders set by shells?

That, or have `DefaultBufferCompositorFactory` take the shader?

robert-ancell: That was what I was initially going for, but I had trouble finding the correct place to have the shader live. So the PR is a bit of a straw man in that I passed the enum all the way to the renderer.

There's also the challenge of how this fits into software rendering, for the common a11y features we want these to work across both GL and software without any special work, but it is also desirable for shells to define something custom. The question is, do these shells make use of this renderer class or do they override all this with their own class?

robert-ancell:
Yes, I do expect we need to change at run-time. I'll look into those existing features to work out how to do this here.

src/renderers/gl/renderer.cpp
```diff
@@ -252,6 +289,166 @@ class mrg::Renderer::ProgramFactory : public mir::graphics::gl::ProgramFactory
     std::mutex compilation_mutex;
 };
 
+class mrg::Renderer::OutputFilterShader
```
RAOF, line 292: Hm. I think this *could be* a [`gl::OutputSurface`](https://github.com/canonical/mir/blob/706fd3b0b8ded81274600d261b0544244d6ab816/include/platform/mir/renderer/gl/gl_surface.h#L31)?

You'd take the underlying `gl::OutputSurface` in the constructor, delegate `{make,release}_current` to it, `bind()` would make *this* framebuffer the draw framebuffer, and then `commit` would `bind()` the underlying `gl::OutputSurface`, do what `render()` does currently, and then call `commit` on the underlying surface?

That would keep all the rest of the renderer code the same, with the only difference being the content of `output_surface`?

src/renderers/gl/renderer.cpp
```diff
@@ -642,6 +859,43 @@ void mrg::Renderer::set_output_transform(glm::mat2 const& t)
...
+    else if (filter_name == "invert")
+    {
+        filter_src = invert_src;
+    }
+    else
+    {
+        output_filter_shader = nullptr;
+        return;
+    }
+    output_filter_shader = std::make_unique<OutputFilterShader>(output_surface->size().width.as_value(), output_surface->size().height.as_value(), filter_src);
```
RAOF, line 896: This is both a threadsafety and `EGLContext`-safety problem; this is *definitely* not going to be called on the compositor thread, so this access races, and we're not guaranteed to have an `EGLContext` current on whatever thread is calling `set_output_filter`.

(Both these problems go away if we move the configuration to `DisplayConfigurationOutput`, because it becomes initialise-only config)

src/server/graphics/default_output_filter.cpp
```diff
@@ -0,0 +1,41 @@
...
+std::string mg::DefaultOutputFilter::filter()
+{
+    return filter_;
+}
+
+void mg::DefaultOutputFilter::set_filter(std::string new_filter)
+{
+    filter_ = new_filter;
+    scene->emit_scene_changed();
+}
```
RAOF, line 40: I'm *pretty* sure this doesn't actually work? `Renderer::set_output_filter` is only ever called from `mc::DefaultDisplayBufferCompositorFactory::create_compositor_for`, but `scene->emit_scene_changed()` doesn't result in tearing down the old `Compositor` and creating a new one, it just wakes up all existing `Compositor`s

(Again, putting config in `DisplayConfigurationOutput` cleanly solves this :wink: )

RAOF (changes requested):
As discussed on our call, I think this makes more sense as an enum on [`DisplayConfigurationOutput`](https://github.com/canonical/mir/blob/706fd3b0b8ded81274600d261b0544244d6ab816/include/platform/mir/graphics/display_configuration.h#L97). That then avoids all the threading issues, we can set up the filter at `Renderer` construction time (and destroy the filter at `Renderer` *destruction* time, avoiding annoying GL context issues), and we can provide specific, optimised implementations for the filters in a software renderer.

src/renderers/gl/renderer.cpp
Saviq: ACK, I wasn't thinking about software rendering in particular.

> I can see the attraction of @Saviq's suggestion of allowing shells to define arbitrary shaders ("night mode" anyone?), but that opens a "can of worms" over what is supported.

Indeed.

src/renderers/gl/renderer.cpp
robert-ancell: Changed to nesting an OutputSurface.

src/renderers/gl/renderer.cpp
robert-ancell: I now compile this in bind(), so I think it is safer? Though the changing of the filter might still need some synchronization?

src/server/graphics/default_output_filter.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+mg::DefaultOutputFilter::DefaultOutputFilter(
+    std::shared_ptr<mi::Scene> const& scene)
+    : scene{scene}
+{
+}
+
+MirOutputFilter mg::DefaultOutputFilter::filter()
+{
+    return filter_;
```
robert-ancell, line 33: Make atomic - this is being written by the configuration thread and read by the rendering thread.

robert-ancell:
Looks like this was being confused by some invalid lines in the .symbols file - see https://github.com/canonical/mir/pull/3959

AlanGriffiths (changes requested):
Some tweaking

include/platform/mir/graphics/output_filter.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace mir
+{
+namespace graphics
+{
+class OutputFilter
+{
+public:
+    virtual MirOutputFilter filter() = 0;
+    virtual void set_filter(MirOutputFilter new_filter) = 0;
```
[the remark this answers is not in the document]
robert-ancell: Updated

src/miral/output_filter.cpp
```diff
@@ -0,0 +1,109 @@
...
+        break;
+    case mir_output_filter_grayscale:
+        default_filter_name = "grayscale";
+        break;
+    case mir_output_filter_invert:
+        default_filter_name = "invert";
+        break;
+    }
+    server.add_configuration_option(
+        output_filter_opt, "Output filter to use [{none,grayscale,invert}]", default_filter_name);
```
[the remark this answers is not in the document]
robert-ancell: This was copied from cursor_scale which has both - should that also do the same?

PR #3920, opened 2025-05-01
feature: implement xdg toplevel tiled states

mattkae:
## What's new?
- Added tiled edge hints per the [xdg shell spec](https://wayland.app/protocols/xdg-shell#xdg_toplevel:enum:state:entry:tiled_left)
- Piped it through `WindowInfo`, `WindowSpecification`, `SurfaceObserver`, `Surface`, `SurfaceSpecification`, and ultimately `XdgToplevelStable` to notify the client of the state

## How To Test
- You can do something like this in `FloatingWindowManagerPolicy::place_new_window`:

```c++
parameters.tiled_edges() = mir::Flags<MirTiledEdge>(mir_tiled_edge_north | mir_tiled_edge_south);
```

Then you should be able to use `WAYLAND_DEBUG=1` to see the flag going by. I don't really know of any clients that use this in a significant way, but it gets to them

### DIFF
--- a/src/server/frontend_wayland/wayland_surface_observer.cpp
+++ b/src/server/frontend_wayland/wayland_surface_observer.cpp
@@ -153,6 +153,15 @@ void mf::WaylandSurfaceObserver::rescale_output(ms::Surface const*, graphics::Di
         });
 }
 
+void mir::frontend::WaylandSurfaceObserver::tiled_edges(scene::Surface const*, Flags<MirTiledEdge> edges)
+{
+    run_on_wayland_thread_unless_window_destroyed(
+        [edges](Impl* impl, WindowWlSurfaceRole*)
+        {
+            impl->window.value().handle_tiled_edges(edges);
+        });
+}
+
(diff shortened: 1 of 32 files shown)


src/server/frontend_wayland/wayland_surface_observer.cpp
```diff
@@ -153,6 +153,16 @@ void mf::WaylandSurfaceObserver::rescale_output(ms::Surface const*, graphics::Di
...
+void mir::frontend::WaylandSurfaceObserver::tiled_edges(scene::Surface const*, Flags<MirTiledEdge> edges)
+{
+    run_on_wayland_thread_unless_window_destroyed(
+        [edges](Impl* impl, WindowWlSurfaceRole*)
+        {
+            impl->window.value().handle_tiled_edges(edges);
+        });
+}
+
+
```
robert-ancell, line 165: Nit: double newline

robert-ancell (approved):
Makes sense to me!

PR #3924, opened 2025-05-05
miral::TestDisplayServer: Allow using real `RenderingPlatform`s in tests

RAOF:
We want to be able to test platform-specific interfaces
(particularly: linux-drm-syncobj) that require a particular
`RenderingPlatform` to be loaded.

This implements `CPUAddressableDisplay*` in the stub
display platform, which then allows a non-stub
`RenderingPlatform` to be loaded and used.

With that, stop forcing the use of the stub rendering
platform in `miral::TestDisplayServer`.

### DIFF
--- a/tests/include/mir/test/doubles/stub_buffer.h
+++ b/tests/include/mir/test/doubles/stub_buffer.h
@@ -55,7 +55,7 @@ public:
                   size,
                   mir_pixel_format_abgr_8888,
                   graphics::BufferUsage::hardware},
-              geometry::Stride{}}
+              geometry::Stride{size.width.as_uint32_t() * MIR_BYTES_PER_PIXEL(mir_pixel_format_abgr_8888)}}
 
     {
     }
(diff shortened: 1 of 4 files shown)


mattkae (approved):
This makes sense to me. Is CI broken because of this though, or something else?

RAOF:
> This makes sense to me. Is CI broken because of this though, or something else?

CI is *super broken* by this. :grimacing:

AlanGriffiths:
Rebased on main to get rid of some unrelated failures

RAOF:
We're leaking DRI resources *some*where; I've cleaned up a few EGL leaks, but because EGL resources are reference-counted I think this is going to be a bit more of a slog until the last leak is found.

AlanGriffiths (approved):
Looking good!

RAOF:
Aha! The infrastructure has fixed itself?

PR #3934, opened 2025-05-07
Implement bounce keys 

tarek-y-ismail:
Adds bounce keys, an accessibility feature that rejects consecutive keypresses of the same key if the occur withing a specific duration. 

Exposes methods for:
- Enabling/Disabling bouncekeys
- Specifying the minimum delay needed for presses to register
- Hooking into rejection notifications.

Adds tests for:
- Basic rejection
- Alternating between different keys
- Pressing the same key with different delays


tarek-y-ismail:
Implementing bounce keys as a filter has a small issue. From what I observed
through manual testing and print debugging , repeat events need to be preceeded
by a down event. This works if you hold down whatever button you want to repeat
on the first press. But for subequent presses, the down event is consumed. The
repeat events that follow are then ignored by applications. To work around
this, we'd need to somehow "see into the future" to let a down event pass if
it's part of a hold and not a press.

Since our code isn't clairvoyant, we need to do this in retrospect. We'll need
to record the last rejected key press, and if it has the same keysym as the key
being repeated, we'll first need to dispatch the down event, followed by repeat
events. Which will require us to push the implementation down to be a 
transformer.

tarek-y-ismail:
Implementing this as a transformer introduces a new issue: The key repeat dispatcher doesn't even get the down events required to begin generating repeat events (because we debounce the events before they reach it)

tarek-y-ismail:
Seems like I was bamboozled by GNOME. Applications follow the "broken" behavior (two taps + hold = no repeat, only one tap registers and the other is debounced), but only the app launcher somehow registers the repeat events.

examples/miral-shell/shell_main.cpp
```diff
@@ -14,6 +14,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "miral/bounce_keys.h"
```
tarek-y-ismail: Should be moved down and changed to look line its fellows

examples/miral-shell/shell_main.cpp
```diff
@@ -38,6 +39,9 @@
 #include <miral/simulated_secondary_click.h>
 #include <miral/hover_click.h>
 
+#define MIR_LOG_COMPONENT "miral-shell"
+#include <mir/log.h>
+
```
tarek-y-ismail: Should be removed

examples/miral-shell/shell_main.cpp
```diff
@@ -210,6 +214,9 @@ int main(int argc, char const* argv[])
 
     miral::HoverClick hover_click_config{false};
 
+    auto bounce_keys =
+        miral::BounceKeys{true}.on_press_rejected([](auto keysym) { mir::log_info("Keysym rejected: %d", keysym); });
```
tarek-y-ismail: No need for logging

include/miral/miral/bounce_keys.h
```diff
@@ -0,0 +1,53 @@
...
+#include <chrono>
+#include <functional>
+#include <memory>
+
+namespace mir
+{
+namespace input
+{
+class XkbSymKey;
+}
```
tarek-y-ismail: Completely unnecessary

include/miral/miral/bounce_keys.h
```diff
@@ -0,0 +1,53 @@
...
+{
+public:
+    explicit BounceKeys(bool enable_by_default);
+    void operator()(mir::Server&);
+
+    BounceKeys& delay(std::chrono::milliseconds);
+    BounceKeys& on_press_rejected(std::function<void(unsigned int keysym)>&&);
+
+    BounceKeys& enable();
+    BounceKeys& disable();
```
tarek-y-ismail: All of these need docs

src/miral/bounce_keys.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+    struct Config
+    {
+        Config(bool enabled_by_default) :
+            enabled{enabled_by_default}
+        {
+        }
+
+        bool enabled;
+        std::chrono::milliseconds delay{500};
```
tarek-y-ismail: This may be too high as a default

src/miral/bounce_keys.cpp
```diff
@@ -0,0 +1,175 @@
...
+                return false;
+            }
+
+            config_->on_press_rejected(keysym);
+
+            return true;
+        }
+
+        mir::Synchronised<RuntimeState> runtime_state;
+        std::shared_ptr<mir::Synchronised<Config>> config;
```
tarek-y-ismail: Can be const?

src/miral/symbols.map
```diff
@@ -524,6 +524,14 @@ global:
 MIRAL_5.3 {
 global:
   extern "C++" {
+    miral::BounceKeys::BounceKeys*;
+    miral::BounceKeys::delay*;
+    miral::BounceKeys::disable*;
+    miral::BounceKeys::enable*;
+    miral::BounceKeys::on_disabled*;
+    miral::BounceKeys::on_enabled*;
```
tarek-y-ismail: Should be removed

tests/miral/test_bounce_keys.cpp
```diff
@@ -0,0 +1,115 @@
...
+                bounce_keys(server);
+            });
+
+        bounce_keys.delay(test_bounce_keys_delay);
+    }
+
+    void SetUp() override
+    {
+        miral::TestServer::SetUp();
+    }
```
tarek-y-ismail: Might not be needed anymore. Depends on if we have to add a wait to account for CI being **slow**

tests/miral/test_bounce_keys.cpp
```diff
@@ -0,0 +1,115 @@
...
+    std::weak_ptr<mir::input::CompositeEventFilter> composite_event_filter;
+    std::weak_ptr<mir::input::InputDeviceRegistry> input_device_registry;
+    std::weak_ptr<mir::input::InputDeviceHub> input_device_hub;
+};
+
+struct TestDifferentDelays: public TestBounceKeys, public WithParamInterface<std::chrono::milliseconds>
+{
+};
+
+TEST_P(TestDifferentDelays, subsequent_keys_rejected_if_in_within_delay)
```
tarek-y-ismail, line 79: Can add another two tests:
- Alternating between pressing two different keys. Should not reject any press.
- Pressing the same key but with different periods between presses. Something like 100ms, 200ms, 100ms, etc..

tests/unit-tests/input/test_input_event_transformer.cpp
tarek-y-ismail, line 1: Not sure if these changes warrant their own PR. Opinions welcome

include/miral/miral/bounce_keys.h
```diff
@@ -0,0 +1,59 @@
...
+
+    /// Enables bounce keys.
+    /// When already enabled, further calls have no effect.
+    BounceKeys& enable();
+
+    /// Disables bounce keys.
+    /// When already disabled, further calls have no effect.
+    BounceKeys& disable();
+
+
```
mattkae:
```suggestion
```

src/miral/bounce_keys.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "mir/time/alarm.h"
+
+#include <chrono>
+#include <optional>
+
+struct miral::BounceKeys::Self
+{
+    struct Config
+    {
+        Config(bool enabled_by_default) :
```
mattkae:
```suggestion
        explicit Config(bool enabled_by_default) :
```

src/miral/bounce_keys.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+miral::BounceKeys& miral::BounceKeys::disable()
+{
+    auto const config = self->config->lock();
+    if (!config->enabled)
+        return *this;
+    config->enabled = false;
+
+    self->bounce_keys_filter.reset();
```
mattkae, line 225: Nit: You have `self->enable()`. It would be nice to also have `self->disable()`

src/miral/bounce_keys.cpp
```diff
@@ -0,0 +1,177 @@
...
+            auto const* input_event = event.to_input();
+            if(input_event->input_type() != mir_input_event_type_key)
+                return false;
+
+            auto const* key_event = input_event->to_keyboard();
+
+            // Only consume "down" events
+            auto const action = key_event->action();
+            if (action == mir_keyboard_action_up || action == mir_keyboard_action_repeat)
+                return false;
```
mattkae:
```suggestion
            // Only consume "down" events
            auto const action = key_event->action();
            if (action != mir_keyboard_action_down)
                return false;
```

mattkae: Right?

src/miral/bounce_keys.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+            auto const keysym = mir_keyboard_event_keysym(key_event);
+
+            auto state = runtime_state.lock();
+            auto config_ = config->lock();
+
+            last_key_clearer->cancel();
+            last_key_clearer->reschedule_in(config_->delay);
+
+            if(!state->last_key || *state->last_key != keysym)
```
mattkae, line 88: @Saviq Is TICS wrong here? `last-key` is checked for null?

src/miral/bounce_keys.cpp
```diff
@@ -0,0 +1,177 @@
...
+    Self(bool enabled_by_default) :
+        config{std::make_shared<mir::Synchronised<Config>>(enabled_by_default)}
+    {
+    }
+
+    std::shared_ptr<mir::Synchronised<Config>> const config;
+    std::shared_ptr<BounceKeysFilter> bounce_keys_filter;
+
+    std::weak_ptr<mir::MainLoop> main_loop;
+    std::weak_ptr<mir::input::CompositeEventFilter> composite_event_filter;
```
mattkae: Prefer using the `namespace mi = mir::input` aliases here (and elsewhere!)

tests/miral/test_bounce_keys.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    miral::BounceKeys bounce_keys{true};
+
+    std::weak_ptr<mir::input::CompositeEventFilter> composite_event_filter;
+    std::weak_ptr<mir::input::InputDeviceRegistry> input_device_registry;
+    std::weak_ptr<mir::input::InputDeviceHub> input_device_hub;
+};
+
+struct TestDifferentDelays: public TestBounceKeys, public WithParamInterface<std::chrono::milliseconds>
```
mattkae, line 75:
```suggestion
struct TestBounceKeysDifferentDelays: public TestBounceKeys, public WithParamInterface<std::chrono::milliseconds>
```

tests/miral/test_bounce_keys.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include <gtest/gtest.h>
+
+using namespace ::testing;
+using namespace std::chrono_literals;
+
+namespace
+{
+static constexpr auto test_bounce_keys_delay = 20ms;
+
+void press(mir::input::InputSink* sink, mir::input::EventBuilder* builder, unsigned int keysym, unsigned int scancode)
```
mattkae:
```suggestion
void press(mir::input::InputSink* sink, mi::EventBuilder* builder, unsigned int keysym, unsigned int scancode)
```

tests/miral/test_bounce_keys.cpp
```diff
@@ -0,0 +1,160 @@
...
+{
+    sink->handle_input(builder->key_event(std::nullopt, mir_keyboard_action_down, keysym, scancode));
+    sink->handle_input(builder->key_event(std::nullopt, mir_keyboard_action_up, keysym, scancode));
+}
+}
+
+struct TestBounceKeys : miral::TestServer
```
mattkae, line 47: Could you add tests for the `enable/disable` methods, as well as the `on_press_rejected` callback?

mattkae (changes requested):
A few comments, but it looks straightforward to me :)

src/miral/bounce_keys.cpp
tarek-y-ismail: I thought of adding it, but that's basically `self->bounce_keys_filter.reset();` so it's really not helping at all

src/miral/bounce_keys.cpp
tarek-y-ismail: This is a known issue with TICS

tests/miral/test_bounce_keys.cpp
tarek-y-ismail: Don't think that's needed, here's a snipped of the output from `miral-test-internal`:
```
[ RUN      ] TestBounceKeys/TestDifferentDelays.subsequent_keys_rejected_if_in_within_delay/0
```

tests/miral/test_bounce_keys.cpp
tarek-y-ismail: `on_press_rejected` is already tested? Am I missing something? 
For `enable/disable`. Their logic is fairly simple, but how do you propose I test them? They don't go through `the_accessibility_manager`, so using `override_the_accessibility_manager` is not possible. I can add `override_the_composite_event_filter` so we can listen to the filter being added and removed, but do we want to add that just for this test?

debian/libmiral7.symbols
```diff
@@ -504,6 +504,13 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::MinimalWindowManager::MinimalWindowManager(miral::WindowManagerTools const&, miral::FocusStealing)@MIRAL_5.0" 5.2.0
  (c++)"miral::WindowManagementPolicy::~WindowManagementPolicy()@MIRAL_5.0" 5.2.0
  MIRAL_5.3@MIRAL_5.3 5.3.0
+ (c++)"miral::BounceKeys::BounceKeys(bool)@MIRAL_5.3" 5.3.0
+ (c++|arch-bits=64)"miral::BounceKeys::delay(std::chrono::duration<long, std::ratio<1l, 1000l> >)@MIRAL_5.3" 5.3.0
+ (c++|arch-bits=32)"miral::BounceKeys::delay(std::chrono::duration<long long, std::ratio<1ll, 1000ll> >)@MIRAL_5.3" 5.3.0
+ (c++)"miral::BounceKeys::disable()@MIRAL_5.3" 5.3.0
+ (c++)"miral::BounceKeys::enable()@MIRAL_5.3" 5.3.0
+ (c++)"miral::BounceKeys::on_press_rejected(std::function<void (unsigned int)>&&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::BounceKeys::operator()(mir::Server&)@MIRAL_5.3" 5.3.0
```
AlanGriffiths: MIRAL_5.3 sailed with 2.21

include/miral/miral/bounce_keys.h
```diff
@@ -0,0 +1,58 @@
...
+#include <chrono>
+#include <functional>
+#include <memory>
+
+namespace mir { class Server; }
+
+namespace miral
+{
+/// Enables configuring bounce keys at runtime.
+/// \remark Since MirAL 5.3
```
AlanGriffiths:
```suggestion
/// \remark Since MirAL 5.4
```

include/miral/miral/bounce_keys.h
```diff
@@ -0,0 +1,58 @@
...
+class BounceKeys
+{
+public:
+    explicit BounceKeys(bool enable_by_default);
```
AlanGriffiths: I'm not keen on using `bool` parameters like this. There's no context for readers of the code to interpret `true`/`false`. Example:

```c++
    BounceKeys bounce_keys{true};
```

Compare with alternative interface designs:

```c++
    auto bounce_keys = BounceKeys::enabled();
```

```c++
    BounceKeys bounce_keys{BounceKeys::enabled};
```

AlanGriffiths: I realise there's prior art in `HoverClick`, `MouseKeysConfig` and `SimulatedSecondaryClick` - sorry for not spotting this earlier

src/miral/bounce_keys.cpp
AlanGriffiths: > This is a known issue with TICS

Don't be so confident that clang-tidy is the problem. I think the warning arises from the two sub-expressions `state->last_key`: you expect them to evaluate to the same thing. It may be that "local" reasoning by clang cannot prove they do

Maybe a clearer condition would help?

```c++
            if (!state->last_key.value_or(keysim) != keysym)
```

BTW it is also possible to put logic into `RuntimeState` directly

src/miral/symbols.map
```diff
@@ -524,6 +524,12 @@ global:
 MIRAL_5.3 {
 global:
   extern "C++" {
+    miral::BounceKeys::BounceKeys*;
+    miral::BounceKeys::delay*;
+    miral::BounceKeys::disable*;
+    miral::BounceKeys::enable*;
+    miral::BounceKeys::on_press_rejected*;
+    miral::BounceKeys::operator*;
```
AlanGriffiths: Wrong stanza

AlanGriffiths (changes requested):

src/miral/bounce_keys.cpp
tarek-y-ismail: I don't think `last_key.value_or(keysym) != keysym` is equivalent to the original expression. In the case of `last_key == nullopt`, which is the initial state, it will evaluate to `keysym != keysym`, causing it to never be set. We'd have to handle this case explicitly by writing `if(!last_key || ...)`. 

I guess using `value_or` could appease TICS, but it really doesn't do much besides adding noise to the line.

AlanGriffiths: You're right, the suggested rewrite isn't equivalent.

examples/miral-shell/shell_main.cpp
```diff
@@ -210,6 +211,8 @@ int main(int argc, char const* argv[])
 
     miral::HoverClick hover_click_config{false};
 
+    auto bounce_keys = BounceKeys::disabled();
```
AlanGriffiths: Not saying you should, but you could inline this variable

include/miral/miral/bounce_keys.h
```diff
@@ -0,0 +1,64 @@
...
+/// Enables configuring bounce keys at runtime.
+/// \remark Since MirAL 5.4
+class BounceKeys
+{
+public:
+    /// Creates a `BounceKeys` instance that's enabled by default.
+    static auto enabled() -> BounceKeys;
+
+    /// Creates a `BounceKeys` instance that's disabled by default.
+    static auto disabled() -> BounceKeys;
```
AlanGriffiths, line 44: It would be good to provide integration with live configuration:

```suggestion
    /// Creates a `BounceKeys` instance that's enabled by default.
    static auto enabled() -> BounceKeys;

    /// Creates a `BounceKeys` instance that's disabled by default.
    static auto disabled() -> BounceKeys;
    
    /// Creates a `BounceKeys` instance that's configured by a live config store
    explicit BounceKeys(live_config::Store& config_store);

```

include/miral/miral/bounce_keys.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <chrono>
+#include <functional>
+#include <memory>
+
+namespace mir { class Server; }
+
+namespace miral
+{
+/// Enables configuring bounce keys at runtime.
```
AlanGriffiths, line 29: We could use a little more exposition on what the "bounce keys" feature is.

include/miral/miral/bounce_keys.h
AlanGriffiths: Here's an example of prior art: `miral::OutputFilter::OutputFilter(live_config::Store& config_store)`

AlanGriffiths (changes requested):
Implementation looks fine, but please add some documentation of this feature. Both in the dox comments and the PR description.

Integration with live config could be a follow-on

examples/miral-shell/shell_main.cpp
tarek-y-ismail: Honestly it feels pretty useless to have the variable with no way to enable/disable it at runtime. Since I just pushed a PR with live config in `server_example.cpp` and this was just added for testing, I think I'll remove this.

tarek-y-ismail: Or maybe its better to add a  `live_config::CliOptions` class and use that in this example? I'm not sure...

AlanGriffiths: `CLiOptions` doesn't sound very "live" to me

tarek-y-ismail: Fair enough. I'm just torn between the following:
- Removing bounce keys completely from miral shell
- Enabling it by default. However, there is no way for people exploring our examples to configure or disable it.
- Adding some extra code to handle CLI configuration

I think the first one is the easiest at the moment since we already showcase it in another example anyway

examples/miral-shell/shell_main.cpp
```diff
@@ -186,6 +187,8 @@ int main(int argc, char const* argv[])
...
```
AlanGriffiths: It seems strange to add `BounceKeys::disabled()` with no logic to enable the feature. What does this illustrate?

tarek-y-ismail: Was for testing. Then when I added the config options, it was there for example. Now with #4054, I don't think it's needed.

AlanGriffiths (approved):
Yup

examples/miral-shell/shell_main.cpp
```diff
@@ -212,6 +212,6 @@ int main(int argc, char const* argv[])
             output_filter,
             AppendKeyboardEventFilter{toggle_output_filter_filter},
             ssc_config,
-            hover_click_config
+            hover_click_config,
```
mattkae, line 215: Nit: spurious change

include/miral/miral/bounce_keys.h
```diff
@@ -0,0 +1,74 @@
...
+{
+public:
+    /// Creates a `BounceKeys` instance that's enabled by default.
+    static auto enabled() -> BounceKeys;
+
+    /// Creates a `BounceKeys` instance that's disabled by default.
+    static auto disabled() -> BounceKeys;
+
+    /// Construct a `BounceKeys` instance with access to a live config store.
+    explicit BounceKeys(live_config::Store& config_store);
```
mattkae, line 47: Did we completely remove the no-parameter constructor? Do we want that @AlanGriffiths ?

mattkae (approved):
I just had two small comments, but nothing blocking

PR #3939, opened 2025-05-08
platforms: Use `NativeBufferBase` in `as_texture`

RAOF:
Users may wish to wrap a `Buffer` in a class with some extra behaviour of their own (notably, tracking when the buffer is no longer used by the rendering pipeline).

In order to allow this wrapping while still having `as_texture` access the *real* buffer type we should be using `NativeBufferBase` so the wrapper can return the underlying buffer.


RAOF:
Note that this is a somewhat minimal change in order to unblock #3905; I think we would ideally reflect this change in the platform API, changing the signature of `GLRenderingProvider::as_texture` to take a `NativeBufferBase`.

PR #3941, opened 2025-05-08
Simplify MinimalWindowManagerTests

AlanGriffiths:
We have been testing that a behaviour fails when that behaviour happens intermittently (and is probably what we want).

We don't currently understand why the behaviour isn't consistent, but changing the tests to test for the correct behaviour and disabling them seems a good intermediate condition

### DIFF
--- a/tests/window_management_tests/test_minimal_window_manager.cpp
+++ b/tests/window_management_tests/test_minimal_window_manager.cpp
@@ -1390,13 +1390,13 @@ TEST_F(MinimalWindowManagerTest, maximized_window_respects_output_removal)
 TEST_F(MinimalWindowManagerTest, maximized_window_respects_output_unused)
 {
     tools().move_cursor_to(geom::PointF(0, 0));
-    auto const app = this->open_application("test");
+    auto const app = open_application("test");
     miral::WindowSpecification spec;
     spec.state() = mir_window_state_maximized;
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
OK, it doesn't cause the failure. But we can now see better what is wrong:

```
2025-05-08T12:53:57.5999696Z [ OK ] MinimalWindowManagerTest.maximized_window_respects_output_unused (18 ms)
2025-05-08T12:53:57.5999914Z [ RUN ] MinimalWindowManagerTest.maximized_window_respects_output_disconnected
2025-05-08T12:53:57.6000216Z Saving server logs to: /tmp/MinimalWindowManagerTest_maximized_window_respects_output_disconnected_server.log
2025-05-08T12:53:57.6000458Z /spread/mir/tests/window_management_tests/test_minimal_window_manager.cpp:1420: Failure
2025-05-08T12:53:57.6000538Z Value of: window_rect
2025-05-08T12:53:57.6000655Z Expected: isn't equal to (800, 0, (1000, 600))
2025-05-08T12:53:57.6000736Z Actual: (800, 0, (1000, 600))
2025-05-08T12:53:57.6000741Z 
2025-05-08T12:53:57.6000993Z [ FAILED ] MinimalWindowManagerTest.maximized_window_respects_output_disconnected (21 ms)
```

Issue #3466, opened 2024-07-08
Introduce fallback to ARGB if XRGB is not supported


Asalle:
### Environment
Architecture: arm64
Operation System: Ubuntu 22.04 (custom image for [Xilinx](https://ubuntu.com/download/amd))
ubuntu-frame: 144-mir2.17.0
HDMI monitor connected for output: Samsung U28R550UQP (4K)

### Steps to reproduce
~ sudo ubuntu-frame

### Expected result
ubuntu-frame started without issue, ubuntu-frame grey gradient visible on the HDMI monitor

### Actual result:
```
# in ubuntu-frame log
[2024-06-26 14:13:47.623448] < - ERROR - > gbm-kms: Output DisplayPort-1 has no associated CRTC to schedule page flips on

# in dmesg
[  +0.000032] [drm:drm_mode_setcrtc] Invalid pixel format XR24 little-endian (0x34325258), modifier 0x0
```

[Hardcoding ARGB](https://github.com/canonical/mir/compare/main...Asalle:mir:xlnx-pixel-format-fallback) solved the issue, but a long-term solution would be beneficial for other kernel drivers that do not support XRGB. [Similar fallback](https://gitlab.gnome.org/GNOME/mutter/-/issues/2354) has been introduced to mutter because of this very device.

robert-ancell:
I've done some work trying to fix this in https://github.com/robert-ancell/mir/tree/gbm-formats. I think the appropriate function to use is `gbm_device_is_format_supported`, though I'm unsure if any flags need to be set on that and we don't use GBM on the eglstream-kms backend (either use the current defaults or query via EGL configs?).

PR #3950, opened 2025-05-15
Check DRM for supported formats

robert-ancell:
Copied from the atomic-kms backend

Fixes https://github.com/canonical/mir/issues/3466

### DIFF
--- a/tests/include/mir/test/doubles/mock_drm.h
+++ b/tests/include/mir/test/doubles/mock_drm.h
@@ -63,7 +63,7 @@ public:
                        drmModeSubPixel subpixel_arrangement = DRM_MODE_SUBPIXEL_UNKNOWN,
                        std::vector<uint32_t> prop_ids = {},
                        std::vector<uint64_t> prop_values = {});
-    void add_plane(std::vector<uint32_t>& formats, uint32_t plane_id, uint32_t crtc_id,
+    void add_plane(std::vector<uint32_t> formats, uint32_t plane_id, uint32_t crtc_id,
                    uint32_t fb_id, uint32_t crtc_x, uint32_t crtc_y, uint32_t x,
                    uint32_t y, uint32_t possible_crtcs_mask, uint32_t gamma_size,
                    std::vector<uint32_t> prop_ids = {}, std::vector<uint64_t> prop_values = {});
@@ -120,7 +120,6 @@ private:
     std::vector<drmModeModeInfo> modes;
     std::vector<drmModeModeInfo> modes_empty;
     std::vector<uint32_t> connector_encoder_ids;
-    std::vector<uint32_t> plane_formats;
 };
 
 class MockDRM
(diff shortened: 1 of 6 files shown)


robert-ancell:
Upon discussion with @RAOF we noticed that the calculated formats is not actually used. So another PR will follow to pass these through the layers and make use of them.

tests/include/mir/test/doubles/mock_drm.h
```diff
@@ -66,20 +83,44 @@ class FakeDRMResources
...
+        uint32_t object_type;
+        std::vector<uint32_t> ids;
+        std::vector<uint64_t> values;
+    };
+
+    struct DRMProperty
+    {
+        uint32_t id;
+        char name[DRM_PROP_NAME_LEN];
+    };
```
AlanGriffiths, line 97: Identical types already exist in `test_drm_mode_resources.cpp` (in a different scope) is there scope for consolidation?

tests/include/mir/test/doubles/mock_drm.h
```diff
@@ -43,21 +44,37 @@ class FakeDRMResources
...
-    void add_encoder(uint32_t encoder_id, uint32_t crtc_id, uint32_t possible_crtcs_mask);
+    drmModePlaneRes* plane_resources_ptr();
+    drmModeObjectProperties* get_object_properties(uint32_t id, uint32_t type);
+    drmModePropertyRes* get_property(uint32_t prop_id);
+    drmModePropertyBlobRes* get_property_blob(uint32_t prop_id);
+
+    uint32_t add_property(const char *name);
+    void add_crtc(uint32_t id, drmModeModeInfo mode,
+                  std::vector<uint32_t> prop_ids = {},
+                  std::vector<uint64_t> prop_values = {});
```
AlanGriffiths, line 55: There's a lot of passing around `std::vector<uint32_t> prop_ids, std::vector<uint64_t> prop_values` in this code which makes me wonder if there's an abstraction here with implicit invariants (like `prop_ids.size() == prop_values.size()`).

AlanGriffiths (approved):
Looks OK, but (non-blocking) the test infrastructure propagates existing mess

tests/include/mir/test/doubles/mock_drm.h
robert-ancell: Yes, it was copied from there. I was going to do a follow up PR to update that test to the new API. I've done it now in this PR.

tests/include/mir/test/doubles/mock_drm.h
robert-ancell: Improved.

robert-ancell:
The Fedora build seems to be timing out, assuming this is unrelated?

RAOF:
>  Error: HTTP request failed with status 401. Please provide a valid TICS authentication token***

PR #3953, opened 2025-05-21
Implement a ScreenShooterFactory so that separate clients do not need to share the same ScreenShooter

mattkae:
## What's new?
- Introduced the `ScreenShooterFactory`
- Deprecated `the_screen_shooter()` in favor of `the_screen_shooter_factory()`
- Implemented the `BasicScreenShooter`
- Implemented the `NullScreenShooter`
- Updated `wlr_screencopy_v1` to use the new factory object
- Moved `MockRendererFactory` to its own file

## Why?
For the magnification accessibility feature, we will also want the screen shooter. If you are magnifiying a small region at the same time that you're recording your screen, you will get a ton of allocations, which is obviously very inefficient. This makes it so that each user can have their own object and avoid reallocations :tada: 

## To Test
Run something like `grimshot save output` and it should work as it did before :)

### DIFF
--- a/src/server/frontend_wayland/wlr_screencopy_v1.cpp
+++ b/src/server/frontend_wayland/wlr_screencopy_v1.cpp
@@ -16,6 +16,7 @@
 
 #include "wlr_screencopy_v1.h"
 
+#include "mir/log.h"
 #include "mir/compositor/screen_shooter.h"
 #include "mir/compositor/screen_shooter_factory.h"
 #include "mir/graphics/graphic_buffer_allocator.h"
@@ -423,15 +424,17 @@ void mf::WlrScreencopyFrameV1::capture(geom::Rectangle buffer_space_damage)
 {
     if (!target)
     {
-        fatal_error(
-            "WlrScreencopyFrameV1::capture() called without a target, copy %s been called",
+        log_error("WlrScreencopyFrameV1::capture() called without a target, copy %s been called",
             copy_has_been_called ? "has" : "has not");
+        report_result(std::nullopt, buffer_space_damage);
+        return;
     }
 
     if (!manager)
     {
-        fatal_error(
-            "WlrScreencopyFrameV1::capture() called without a manager");
+        log_error("WlrScreencopyFrameV1::capture() called without a manager");
+        report_result(std::nullopt, buffer_space_damage);
+        return;
     }
 
     manager.value().screen_shooter->capture(std::move(target), params.output_space_area,
--- a/tests/unit-tests/compositor/test_basic_screen_shooter_factory.cpp
+++ b/tests/unit-tests/compositor/test_basic_screen_shooter_factory.cpp
@@ -74,5 +74,5 @@ public:
 
 TEST_F(BasicScreenShooterFactoryTest, creates_basic_screen_shooter)
 {
-    EXPECT_THAT(factory->create(), Ne(nullptr));
+    EXPECT_THAT(factory->create(), NotNull());
 }


src/server/compositor/default_configuration.cpp
```diff
@@ -134,3 +136,45 @@ auto mir::DefaultServerConfiguration::the_screen_shooter() -> std::shared_ptr<co
...
+            }
+            catch (...)
+            {
+                mir::log(
+                    ::mir::logging::Severity::error,
+                    "",
+                    std::current_exception(),
+                    "failed to create BasicScreenShooter");
+                return std::make_shared<compositor::NullScreenShooterFactory>(thread_pool_executor);
+            }
```
AlanGriffiths, line 178: It looks as though the only exception for which this would make sense is the one thrown on L158. There is no value in using exceptions for that path. Am I missing something?

mattkae: You're very right, I was mindlessly copying the old code. I will fix

src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -424,7 +427,14 @@ void mf::WlrScreencopyFrameV1::capture(geom::Rectangle buffer_space_damage)
             "WlrScreencopyFrameV1::capture() called without a target, copy %s been called",
             copy_has_been_called ? "has" : "has not");
     }
-    ctx->screen_shooter->capture(std::move(target), params.output_space_area,
+
+    if (!manager)
+    {
+        fatal_error(
+            "WlrScreencopyFrameV1::capture() called without a manager");
+    }
```
AlanGriffiths, line 435: This feels more like the state machine assumed by the protocol is broken than the "we messed up" implied by the fatal error, but I've not looked deeper, but maybe an exception would be better as the client destroying things under us deserves a disconnect

tests/unit-tests/compositor/test_basic_screen_shooter_factory.cpp
```diff
@@ -0,0 +1,78 @@
...
+    std::shared_ptr<mtd::MockGlRenderingProvider> gl_provider{std::make_shared<NiceMock<mtd::MockGlRenderingProvider>>()};
+    std::vector<std::shared_ptr<mg::GLRenderingProvider>> gl_providers{gl_provider};
+    std::shared_ptr<mtd::MockRendererFactory> renderer_factory{std::make_shared<NiceMock<mtd::MockRendererFactory>>()};
+    std::shared_ptr<mtd::StubBufferAllocator> buffer_allocator;
+    std::unique_ptr<mc::BasicScreenShooterFactory> factory;
+};
+
+TEST_F(BasicScreenShooterFactoryTest, creates_basic_screen_shooter)
+{
+    EXPECT_THAT(factory->create(), Ne(nullptr));
```
AlanGriffiths, line 77:
```suggestion
    EXPECT_THAT(factory->create(), NotNull());
```

AlanGriffiths (approved):
OK

src/server/frontend_wayland/wlr_screencopy_v1.cpp
mattkae: According to the protocol, the "right" think to do is to send a failed event: https://wayland.app/protocols/wlr-screencopy-unstable-v1#zwlr_screencopy_frame_v1:request:copy

I will log the error and send that, as it is the least destructive thing to do that will also notify the client

PR #3955, opened 2025-05-22
Update exception text when fail to call drmModeGetProperty

robert-ancell:
Seems to be a copy-paste error as it duplicates the exception text for drmModeGetPlane


RAOF (approved):
Yep, I think you are correct.

PR #3958, opened 2025-05-22
Remove `Surface::primary_buffer_stream` and add `Surface::get_streams` for testing purposes in its place

mattkae:
## What's new?
- Removed `Surface::primary_buffer_stream`
- Added `Surface::get_streams` so that the WLCS tests (or anyone else) can still manually grab the streams if they need to
- Updated the WLCS test to use `get_streams`

## Why?
- With magnification, `BasicSurface`s should be able to be stream-less, so this needs to be done :)


src/server/symbols.map
```diff
@@ -1,4 +1,4 @@
-MIR_SERVER_INTERNAL_2.21 {
+MIR_SERVER_INTERNAL_2.22 {
```
AlanGriffiths, line 1: @Saviq what are we doing with release/2.21? Could this be part of it?

AlanGriffiths (approved):
Apart from wondering if the 2.21 ship has sailed, uncontentious

src/server/symbols.map
Saviq: I'll pull `main` to `release/2.21` again next week, so yes it can be.

PR #3968, opened 2025-05-28
(#3933) fixing improper texture allocation across cards

### LINKED ISSUE #3933: Texture allocation happens with an incorrect context for the renderer that needs it
Spun out of https://github.com/canonical/mir/issues/3923

## Problem
I have two monitors: one connected to an AMD GPU, the other to my motherboard. This commit https://github.com/canonical/mir/commit/8b48e1305ae3ee24cae15a4b7df46f3fc37d7f3c caused the monitor connected to my motherboard to no longer render anything. It just renders black squares for all textures that it is supposed to render.

Before this commit, we would generate the `tex_id` on the first `bind()`, which ensured that the consumer's `EGLContext` was current.

Now, we always generate the `tex_id` on the `BufferAllocator`'s `EGLContext`, which might not be the context current when the renderer `bind()`s to it.

## Solution
- Ideally, we defer texture making for SHM buffers to `GlRenderingProvider::as_texture()`, because prior to that, we do not know they'll ever _need_ to be accessed as a texture

mattkae:
fixes #3933 

## What's new?
- `GLRenderingProvider::as_texture` now checks if we have a `ShmBuffer` and returns a texture for that specific provider instead of a single texture for all providers
- Implemented `ShmBuffer::texture_for_provider` which will return a `gl::Texture` either from a map cache or elsewhere
- Added `ShmBuffer::create_texture` which has a default implementation, but will be overridden for each of the subclasses. These subclasses mostly need to override the `bind()` method
- `ShmBuffer` no longer implements `gl::Texture`
- `ShmBuffer` no longer takes an `EGLContextExecutor` unless it is creating a texture
- Updated the associated unit tests

### DIFF
--- a/src/platforms/common/server/shm_buffer.cpp
+++ b/src/platforms/common/server/shm_buffer.cpp
@@ -29,13 +29,27 @@
 
 #include <boost/throw_exception.hpp>
 
-#include <string.h>
-#include <endian.h>
-
 namespace mg=mir::graphics;
 namespace mgc = mir::graphics::common;
 namespace geom = mir::geometry;
 namespace mrs = mir::renderer::software;
+
+namespace
+{
+GLuint new_texture()
+{
+    GLuint tex;
+    glGenTextures(1, &tex);
+
+    glBindTexture(GL_TEXTURE_2D, tex);
+    // The ShmBuffer *should* be immutable, so we can just set up the properties once
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
+    glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
+    return tex;
+}
+}
 
 bool mg::get_gl_pixel_format(MirPixelFormat mir_format,
                          GLenum& gl_format, GLenum& gl_type)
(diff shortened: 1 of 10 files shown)


src/platforms/common/server/kms-utils/kms_connector.cpp
```diff
@@ -201,5 +203,11 @@ auto mgk::find_crtc_with_primary_plane(
         }
     }
 
+    if (plane_res.planes().begin() != plane_res.planes().end())
+    {
+        mir::log_error("Could not find primary plane for CRTC, using the first available one");
+        return std::make_pair(std::move(crtc), std::move(*plane_res.planes().begin()));
+    }
```
mattkae, line 210: @RAOF @tarek-y-ismail 

I need to add this bit to make things work here, but this is obviously unrelated to this PR. I can move it to another PR, but is this even the right fix for this? I can't even start Mir on my machine with out this...

RAOF: That is extremely weird? Also, that code hasn't changed in ages; why would you now need this? (This looks like a standup-tomorrow topic)

RAOF:
I think this might be missing a few `as_texture()` implementations, and maybe there can be some simplifications? I need to spend some more time reviewing tomorrow.

mattkae:
> I think this might be missing a few `as_texture()` implementations, and maybe there can be some simplifications? I need to spend some more time reviewing tomorrow.

I think I just need to fix `renderer-generic-egl`, but the eglstreams one doesn't need saving since it is disappearing anyway, right?

src/platforms/common/server/kms-utils/kms_connector.cpp
RAOF: (This is actually https://github.com/canonical/mir/pull/3978)

RAOF (commented):
Hm. I *think* this can actually be dramatically simplified? I think you only need a single `ShmTexture` implementation; the differences are only how to upload the texture (and, a little annoyingly, needing to update the texture sometime for the `MemoryBacked` one).

If the `create_texture` implementations created a ShmTexture and also uploaded to that texture at the same time, you could have a single implementation (calling `on_consumed()` at `texture_for_provider()` time rather than `bind()` time [should be] is fine).

That leaves `mark_dirty`; this could be either just dropping the `provider_to_texture_map` when it's dirty (and re-allocating textures) or adding a `refresh` callback called in `texture_for_provider`?

tests/unit-tests/graphics/test_shm_buffer.cpp
```diff
@@ -407,11 +410,12 @@ TEST_F(ShmBufferTest, tex_id_is_generated_on_thread_with_current_egl_context)
...
+    eglMakeCurrent(dummy_dpy, EGL_NO_SURFACE, EGL_NO_SURFACE, dummy_ctx);
 
-    PlatformlessShmBuffer buffer{size, pixel_format, egl_delegate};
+    PlatformlessShmBuffer buffer{size, pixel_format};
 
     // Ensure that the texture ID has been resolved, and to the value we expect
-    EXPECT_THAT(buffer.tex_id(), Eq(tex_id));
+    auto const texture = buffer.texture_for_provider(egl_delegate, rendering_provider.get());
+    EXPECT_THAT(texture->tex_id(), Eq(tex_id));
 }
```
RAOF: I think this test can just be removed entirely, since it no longer really makes sense now?

src/platforms/common/server/shm_buffer.h
```diff
@@ -53,24 +55,47 @@ class ShmBuffer :
...
+            geometry::Stride const& stride,
+            MirPixelFormat pixel_format);
+        void mark_dirty();
+
+    private:
+        std::shared_ptr<EGLContextExecutor> egl_delegate;
+        GLuint tex_id_;
+        std::mutex uploaded_mutex;
+        bool uploaded = false;
+    };
```
RAOF: This doesn't actually need to be in the header, right? This could be in an anonymous namespace in the implementation file?

src/platforms/common/server/shm_buffer.h
```diff
@@ -53,24 +55,47 @@ class ShmBuffer :
...
+
+    virtual void on_texture_accessed(std::shared_ptr<ShmBufferTexture> const&) = 0;
+
+    std::mutex tex_mutex;
+    std::map<RenderingProvider*, std::shared_ptr<ShmBufferTexture>> provider_to_texture_map;
```
RAOF:
```suggestion
    mir::Synchronised<std::map<RenderingProvider*, std::shared_ptr<ShmBufferTexture>>> provider_to_texture_map;
```
?

src/platforms/common/server/shm_buffer.h
```diff
@@ -53,24 +55,47 @@ class ShmBuffer :
...
+            geometry::Size const& size,
+            geometry::Stride const& stride,
+            MirPixelFormat pixel_format);
+        void mark_dirty();
+
+    private:
+        std::shared_ptr<EGLContextExecutor> egl_delegate;
+        GLuint tex_id_;
+        std::mutex uploaded_mutex;
+        bool uploaded = false;
```
RAOF: It might be cleaner to move these out into `MemoryBackedShmBuffer`, as they're only needed there. But that'd require changing `texture_for_provider` to upload the texture the first time, and call `on_texture_accessed` on subsequent accesses.

(Or, actually, `MemoryBackedShmBuffer` could just delete the `ShmBufferTexture` when marked dirty; it's already going to be re-allocating :woman_shrugging: )

src/platforms/common/server/shm_buffer.cpp
```diff
@@ -436,3 +431,9 @@ auto mgc::NotifyingMappableBackedShmBuffer::map_rw() -> std::unique_ptr<mrs::Map
     notify_consumed();
     return MappableBackedShmBuffer::map_rw();
 }
+
+void mgc::NotifyingMappableBackedShmBuffer::on_texture_accessed(std::shared_ptr<ShmBufferTexture> const& texture)
+{
+    MappableBackedShmBuffer::on_texture_accessed(texture);
+    notify_consumed();
```
RAOF, line 449: This call doesn't actually need to be here - `on_texture_accessed()` calls `map_readable`, which already calls `notify_consumed()`.

(Which, in turn, means that this whole overload doesn't have to exist)

RAOF (commented):
Looks better; I think this is pretty much ok.

Feel free to try to remove the `uploaded`/`mark_dirty()` from `ShmBufferTexture`, but if it doesn't go cleanly that's ok.

src/platforms/common/server/shm_buffer.cpp
mattkae: This actually is necessary. If I remove `notify_consumed` here, then the texture never gets updated

src/platforms/common/server/shm_buffer.h
mattkae: Personally I like the idea of not reallocating `ShmBufferTexture` and just getting a single texture id and updating it. IMO, it is easier to have an `uploaded` per texture, as we know better when that particular texture has been uploaded and then we can handle that in `on_texture_accessed` for that texture.

RAOF (approved):
LGTM now.

I think the TICS findings are correct, though; this will reduce test coverage.

We could probably test the generic-egl-renderer by running some wlcs tests with that renderer? I'd need to check what lines we're missing in shm_buffer.cpp to see if that's something easily testable.

mattkae:
@RAOF 

So it seems like two things:

1. `buffer_allocator.cpp` is claiming that we reduced coverage, but TICS isn't even showing the lines that I added :thinking: In `mge::GLRenderingProvider::as_texture`, my new lines aren't shown
2. `shm_buffer.cpp` is mad about things that were preexisting, but that I moved. All new functionality appears to be tested :)

Saviq:
> TICS isn't even showing the lines that I added

There's three different `buffer_allocator.cpp`s, did TICS get confused?

Saviq:
image

mattkae:
> image

That seems to just be the old version of `src/platforms/renderer-generic-egl/buffer_allocator.cpp` which now looks like:

```c++
auto mge::GLRenderingProvider::as_texture(std::shared_ptr<Buffer> buffer) -> std::shared_ptr<gl::Texture>
{
    std::shared_ptr<NativeBufferBase> native_buffer{buffer, buffer->native_buffer_base()};
    if (dmabuf_provider)
    {
        if (auto tex = dmabuf_provider->as_texture(native_buffer))
        {
            return tex;
        }
    }
    if (auto shm = std::dynamic_pointer_cast<mgc::ShmBuffer>(native_buffer))
    {
        return shm->texture_for_provider(egl_delegate, this);
    }

    // TODO: Should this be abstracted, like dmabuf_provider above?
    return std::dynamic_pointer_cast<gl::Texture>(native_buffer);
}
```

Saviq:
> That seems to just be the old version of `src/platforms/renderer-generic-egl/buffer_allocator.cpp` which now looks like:

So yeah, TICS seems to have got confused. I'll reopen the PR to see if a re-run is better.

PR #3970, opened 2025-05-28
Account for the OutputFilter in the BasicScreenShooter

mattkae:
## What's new?
- When doing magnification work, I noticed that the color of the scene was always inverted after pulling latest from `main`
- Looks like we needed to set the filter in the `BasicScreenShooter`

## To test
- Run miral-shell
- Run `grimshot save output`. Note that the image is correctly not inverted
- Invert the color using Ctrl + I
- Run `grimshot save output` again. Note that the image is correctly inverted


robert-ancell:
The default behaviour being invert was due to an uninitialized variable - fixed in https://github.com/canonical/mir/pull/3971

robert-ancell (approved):
Thanks!

PR #3974, opened 2025-05-29
Redo `mgk::find_crtc_with_primary_plane`

### LINKED ISSUE #3972: Clean up `mgk::find_crtc_with_primary_plane`
This section: 

https://github.com/canonical/mir/blob/23c585ab5ab7c3ad34a618c469e1a7eb08aa4511/src/platforms/common/server/kms-utils/kms_connector.cpp#L185-L202

Basically boils down to:
1. Find a CRTC not in use that's compatible with the encoders attached to the passed connector
2. Find a primary plane that's compatible with the CRTC found in step 1

But given that `drmModePlane::possible_crtcs` and `drmModeEncoder::possible_crtcs` bit masks both refer to the same set of CRTCs, we could instead:
1. Find primary planes
2. For all encoders attached to the passed connector:
  3. For every nonzero bit in  `plane->possible_crtcs & encoder->possible_crtcs`:
    4. Find the corresponding CRTC (CRTC index is the index of the bit)
    5. Check if the CRTC is not used
    6. Return the CRTC
7. If nothing is found, blow up like before.

tarek-y-ismail:
Might close #3972 

# Whats new
Redid `mgk::find_crtc_with_primary_plane` to try and use any CRTC and primary plane compatible with the passed connector.

# How to test
Just launch mir like usual

### DIFF
--- a/src/platforms/common/server/kms-utils/kms_connector.cpp
+++ b/src/platforms/common/server/kms-utils/kms_connector.cpp
@@ -126,6 +126,92 @@ std::tuple<mgk::DRMModeCrtcUPtr, int> find_crtc_and_index_for_connector(
 
     BOOST_THROW_EXCEPTION(std::runtime_error{"Failed to find CRTC"});
 }
+
+namespace mgk = mir::graphics::kms;
+auto try_find_crtc_and_plane_for_connector(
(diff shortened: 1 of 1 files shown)


tarek-y-ismail:
Superseded by #4626

PR #3975, opened 2025-05-29
RFC/live-config-prototype

AlanGriffiths:
*This is not ready to merge*

I am looking for feedback before fixing this approach as an API.

* Is this doing what folks expect?
* Do the names make sense?

Proposal:  we need something like the `AttributeHandler` interface introduced to miral. This will allow miral types that support "live configuration" to register the attributes they support without knowing the configuration engine that supports these attributes.

I've illustrated this with extensions:

* `OutputFilter : miral::OutputFilter`;
* `InputConfiguration : miral::InputConfiguration`;  and,
* `CursorScale : miral::CursorScale`

These show the logic that would need to be added to these classes (via an additional constructor?) 

As a POC I've reworked the `mir_demo_server` `DemoConfigFile` on this basis, and extended the types it was already supporting to provide the existing functionality.

I've not tried to illustrate use with a YAML, or a gsettings backend, but these should be tractable. For these the `AttributeHandler` supplied would know any additional context needed (such as the node on which these attributes exist)

### DIFF
--- a/examples/mir_demo_server/server_example.cpp
+++ b/examples/mir_demo_server/server_example.cpp
@@ -42,6 +42,7 @@
 #include <chrono>
 #include <cstdlib>
 #include <iostream>
+#include <list>
 #include <mutex>
 
 namespace mir { class AbnormalExit; }
@@ -131,10 +132,15 @@ public:
     using HandleInt = std::function<void(std::string_view key, std::optional<int> value)>;
     using HandleFloat = std::function<void(std::string_view key, std::optional<float> value)>;
     using HandleString = std::function<void(std::string_view key, std::optional<std::string_view> value)>;
+    using HandleDone = std::function<void()>;
 
     virtual void add_attribute(std::string_view key, HandleInt handler) = 0;
     virtual void add_attribute(std::string_view key, HandleFloat handler) = 0;
     virtual void add_attribute(std::string_view key, HandleString handler) = 0;
+
+    /// Called following a set of related updates (e.g. a file reload) to allow
+    /// multiple attributes to be updated transactionally
+    virtual void on_done(HandleDone handler) = 0;
 
     virtual ~AttributeHandler() = default;
 };
@@ -152,47 +158,32 @@ public:
         runner.add_start_callback([this]
             {
                 std::lock_guard lock{config_mutex};
-
-                // Merge the input options collected from the command line,
-                // environment, and default `.config` file with the options we
-                // read from the `.input` file.
-                //
-                // In this case, the `.input` file takes precedence. You can
-                // reverse the arguments to de-prioritize it.
-                mouse.merge(input_configuration.mouse());
-                keyboard.merge(input_configuration.keyboard());
-                touchpad.merge(input_configuration.touchpad());
-
                 apply_config();
             });
     }
 
     void operator()(mir::Server& server)
     {
-        input_configuration(server);
         cursor_scale(server);
     }
 
     void add_attribute(std::string_view key, HandleInt handler) override;
     void add_attribute(std::string_view key, HandleFloat handler) override;
     void add_attribute(std::string_view key, HandleString handler) override;
+    void on_done(HandleDone handler) override;
 
 private:
     std::map<std::string, HandleString> attribute_handlers;
+    std::list<HandleDone> done_handlers;
 
-    miral::InputConfiguration input_configuration;
-    miral::InputConfiguration::Mouse mouse = input_configuration.mouse();
-    miral::InputConfiguration::Touchpad touchpad = input_configuration.touchpad();
-    miral::InputConfiguration::Keyboard keyboard = input_configuration.keyboard();
     std::mutex config_mutex;
     miral::CursorScale cursor_scale;
     miral::ConfigFile config_file;
 
     void apply_config()
     {
-        input_configuration.mouse(mouse);
-        input_configuration.touchpad(touchpad);
-        input_configuration.keyboard(keyboard);
+        for (auto const& handler : done_handlers)
+            handler();
     };
 
     void loader(std::istream& in, std::filesystem::path const& path)
@@ -216,20 +207,6 @@ private:
                     handler->second(key, value);
                 }
 
-                auto const parse_and_validate_int = [](std::string const& key, std::string_view val) -> std::optional<int>
-                {
-                    auto const int_val = std::atoi(val.data());
-                    if (int_val < 0)
-                    {
-                        mir::log_warning(
-                            "Config value %s does not support negative values. Ignoring the supplied value (%d)...",
-                            key.c_str(), int_val);
-                        return std::nullopt;
-                    }
-
-                    return int_val;
-                };
-
                 auto const parse_and_validate_float = [](std::string const& key, std::string_view val) -> std::optional<float>
                 {
                     auto const float_val = std::atof(val.data());
@@ -245,44 +222,6 @@ private:
                     return float_val;
                 };
 
-                if (key == "pointer_handedness")
-                {
-                    if (value == "right")
-                        mouse.handedness(mir_pointer_handedness_right);
-                    else if (value == "left")
-                        mouse.handedness(mir_pointer_handedness_left);
-                    else
-                        mir::log_warning("Config key %s does not support value: '%s'", key.c_str(), value.c_str());
-                }
-
-                if (key == "touchpad_scroll_mode")
-                {
-                    if (value == "none")
-                        touchpad.scroll_mode(mir_touchpad_scroll_mode_none);
-                    else if (value == "two_finger_scroll")
-                        touchpad.scroll_mode(mir_touchpad_scroll_mode_two_finger_scroll);
-                    else if (value == "edge_scroll")
-                        touchpad.scroll_mode(mir_touchpad_scroll_mode_edge_scroll);
-                    else if (value == "button_down_scroll")
-                        touchpad.scroll_mode(mir_touchpad_scroll_mode_button_down_scroll);
-                    else
-                        mir::log_warning("Config key %s does not support value: '%s'", key.c_str(), value.c_str());
-                }
-
-                if (key == "repeat_rate")
-                {
-                    auto const parsed = parse_and_validate_int(key, value);
-                    if (parsed)
-                        keyboard.set_repeat_rate(*parsed);
-                }
-
-                if (key == "repeat_delay")
-                {
-                    auto const parsed = parse_and_validate_int(key, value);
-                    if (parsed)
-                        keyboard.set_repeat_delay(*parsed);
-                }
-
                 if (key == "cursor_scale")
                 {
                     auto const parsed = parse_and_validate_float(key, value);
@@ -295,6 +234,12 @@ private:
         apply_config();
     }
 };
+
+void DemoConfigFile::on_done(HandleDone handler)
+{
+    std::lock_guard lock{config_mutex};
+    done_handlers.emplace_back(std::move(handler));
+}
 
 void DemoConfigFile::add_attribute(std::string_view key, HandleInt handler)
 {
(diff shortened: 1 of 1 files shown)


examples/mir_demo_server/server_example.cpp
```diff
@@ -117,7 +118,186 @@ catch (...)
...
-class DemoConfigFile
+/// Interface for adding attributes to a configuration tool.
+///
+/// The handlers should be called when the configuration is updated. There is
+/// no requirement to check the previous value has changed.
+///
+/// The key is passed to the handler as it can be useful (e.g. for diagnostics)
+///
+/// This could be supported by various backends (an ini file, a YAML node, etc)
+class AttributeHandler
```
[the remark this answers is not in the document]
AlanGriffiths: More observations for discussion...

In comparison to our existing config we are lacking:

1. Boolean options
2. Valueless options (e.g. `enable-touchspots`)
3. Repeating `vector<string>` options

Compared to `YAML::Node` we are lacking:

1. `IsDefined()`
2. `IsSequence()`
3. `IsMap()`
4. The templated conversions to `int`, `unsigned`, ..., `long double`

I don't think we need equivalents of any of these, but this should be discussed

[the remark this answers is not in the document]
Saviq: > 2. Valueless options (e.g. `enable-touchspots`)

These are basically booleans, IMO.

> 3. Repeating `vector<string>` options

Arrays are first-class citizens in many configuration schemes, losing those would mean we have to marshal them around?

> 1. `IsDefined()`

The optional being empty would be that, no?

> 2. `IsSequence()`

Arrays...

> 3. `IsMap()`

Shouldn't go there, don't think. But then, can we express what we use today?

NB: we can keep display YAML special, I think - we don't need to replicate it all here.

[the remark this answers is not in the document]
AlanGriffiths: > > 2. Valueless options (e.g. `enable-touchspots`)
> 
> These are basically booleans, IMO.

You might think that, but they are not as general as booleans (there's no value to parse, and no `false` equivalent).

> > 3. Repeating `vector<string>` options
> 
> Arrays are first-class citizens in many configuration schemes, losing those would mean we have to marshal them around?

Should arrays of all types be handled? So far we have only needed arrays of string in our dead config.

Thinking about it, do we need _explicit_ arrays? A "foo" handler can be called multiple times before a done handler is called without changing the interface. 

> > 1. `IsDefined()`
> 
> The optional being empty would be that, no?

No, that's `IsNull()`

> > 2. `IsSequence()`
> 
> Arrays...
> 
> > 3. `IsMap()`
> 
> Shouldn't go there, don't think. But then, can we express what we use today?
> 
> NB: we can keep display YAML special, I think - we don't need to replicate it all here.

Agreed. This is about _the options we need_ (for input config, accessibility, etc) from different "live config" implementations. We can bundle generic implementations, but shells can implement and use more capabilities than the minimum we require here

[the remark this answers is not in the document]
Saviq: > > These are basically booleans, IMO.
> You might think that, but they are not as general as booleans (there's no value to parse, and no `false` equivalent).

I know, but we can model them as such in this API.

> Should arrays of all types be handled? So far we have only needed arrays of string in our dead config.
> 
> Thinking about it, do we need _explicit_ arrays? A "foo" handler can be called multiple times before a done handler is called without changing the interface.

Not a hot path, this, so can be convinced.

> No, that's `IsNull()`

Right, I think that is useful to differentiate "default" from "same as default" from "empty".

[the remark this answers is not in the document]
RAOF: > > > These are basically booleans, IMO.
> 
> > You might think that, but they are not as general as booleans (there's no value to parse, and no `false` equivalent).
> 
> I know, but we can model them as such in this API.

I don't think it's interesting to distinguish boolean arguments and valueless arguments?
Does it benefit us to distinguish `--foo=true` from `--foo`, and `--foo=false` from not passing `--foo`?
Additionally, I don't think all backends can support this (eg: JSON doesn't have valueless keys; best you could do is `key: null`)

I think we *do* want to support boolean options; we currently have a bunch!

> > > 3. Repeating `vector<string>` options
> > 
> > 
> > Arrays are first-class citizens in many configuration schemes, losing those would mean we have to marshal them around?
> 
> Should arrays of all types be handled? So far we have only needed arrays of string in our dead config.
> 
> Thinking about it, do we need _explicit_ arrays? A "foo" handler can be called multiple times before a done handler is called without changing the interface.

Hm. I was *going* to say that I don't think we need explicit arrays, but I think that's wrong - if we want to write out a sample config, we need to be able to distinguish arrays from one-value arguments, as the syntax tends to be different.

> 
> > > 1. `IsDefined()`
> > 
> > 
> > The optional being empty would be that, no?
> 
> No, that's `IsNull()` 

No, I don't think we need this. We can model going from “value set” to “value unset” by invoking the handler with a `nullopt`, and we can map `IsDefined()` and `IsNull()` onto boolean options without loss of generality. 

> > > 3. `IsMap()`
> > 
> > 
> > Shouldn't go there, don't think. But then, can we express what we use today?
> > NB: we can keep display YAML special, I think - we don't need to replicate it all here.
> 
> Agreed. This is about _the options we need_ (for input config, accessibility, etc) from different "live config" implementations. We can bundle generic implementations, but shells can implement and use more capabilities than the minimum we require here

I think we might want this? At least, it's the cleanest way to map to a hierarchical structure.

If we flipped this structure around, and had something like
```
template<typename T>
struct Attribute
{
   std::string key;
   std::function<void(std::string_view key, std::optional<T> value)> handler;

   // Also handling the description/default value, while we're at it
   std::string description;
   T default_value;
   // Maybe this would handle arrays?
   bool allow_multiple;
};

struct Attributes
{
    std::vector<Attribute<int>> int_opts;
    // the other ones here

    // and now we can have hierarchy; input - mouse - handedness
    std::vector<Attribute<Attributes>> suboptions;
    HandleDone when_done;
};
```
we could get all the stuff we need in?

(Hm. Does this want to interact with `mir::options::Option` at all? It's a very similar thing)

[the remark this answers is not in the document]
AlanGriffiths: > I think we might want this? At least, it's the cleanest way to map to a hierarchical structure.

The question is whether we should require a hierarchical structure for "live config" interaction. There is a cost to generality, both as a user and as an implementer.

I am aware that configuration schemes typically support them, but here the deepest hierarchy we're discussing of is, in the context of input: `touchpad/scroll_mode`. Or, in the context of accessibility: `mouse-keys/acceleration-constant-factor`. (And, yes "context" is a level, but does not need encoding in the attribute.)

The simplest thing to support is simple "key, value" pairs. I don't yet see a compelling argument for something more complex. (I'm not yet convinced that "key" needs to be a tuple, but Matt raised the question)

> (Hm. Does this want to interact with `mir::options::Option` at all? It's a very similar thing)

I don't think so, because it is so different. The similarity is that it deals with configuration settings, the difference is how:

* `mir::options::Option` is a way to look up configuration settings during initialization. The existing code assumes these settings are fixed prior to initialization and builds them into the runtime. There are a lot of implicit assumptions that would need fixing to "interact" here.

* Here we are specifying a generic mechanism to handle configuration options changing and enable multiple "live" configuration backends.

I think these are two different types of configuration. There's certainly stuff in the old config file that is better handled "live", but that's a migration exercise to follow.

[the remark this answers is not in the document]
mattkae: The key doesn't have to be a tuple I suppose, but I will certainly break it up into one in miracle to map it better into YAML. That being said, that is simply a quirk that I might have to live with, e.g.:

YAML -> Internal Representation -> Calling Function for key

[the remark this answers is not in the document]
AlanGriffiths: OK, looking through the mouse-keys options I found:

```c++
char const* const mouse_keys_max_speed_x = "mouse-keys-max-speed-x";
char const* const mouse_keys_max_speed_y = "mouse-keys-max-speed-y";
```

That looks like a 3-tuple to me:

```yaml
    ...
        mouse_keys:
            max_speed:
                x: 400.0
                y: 400.0
```

And the fact that there's no obvious split() algorithm has me "sold" on key being a tuple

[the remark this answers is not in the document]
mattkae: Yes I like that :)

AlanGriffiths:
New iteration:

1. Keys are now compound (but, in this spike, not constraining characters)
2. Added descriptions
3. Added bool attributes
4. Added sequence attributes (for all types)

I don't think the following are needed to advance the discussion, but they are required for the real thing

1. Default values
3. _Implementing_ sequence attributes in `DemoConfigFile`

I don't think we need to require configuration systems to:

1. support nested attributes
2. have "valueless" options
4. offer a way to query attribute values

examples/mir_demo_server/server_example.cpp
```diff
@@ -117,7 +118,272 @@ catch (...)
...
+    virtual void add_bool_attribute(HandleBool handler, ConfigurationKey const& key) = 0;
+    virtual void add_bools_attribute(HandleBools handler, ConfigurationKey const& key) = 0;
+    virtual void add_float_attribute(HandleFloat handler, ConfigurationKey const& key) = 0;
+    virtual void add_floats_attribute(HandleFloats handler, ConfigurationKey const& key) = 0;
+    virtual void add_string_attribute(HandleString handler, ConfigurationKey const& key) = 0;
+    virtual void add_strings_attribute(HandleStrings handler, ConfigurationKey const& key) = 0;
+
+    /// Called following a set of related updates (e.g. a file reload) to allow
+    /// multiple attributes to be updated transactionally
+    virtual void on_done(HandleDone handler) = 0;
```
mattkae, line 220: Nit: I wonder if the wording `commit` would be better than `done`

examples/mir_demo_server/server_example.cpp
```diff
@@ -117,7 +118,272 @@ catch (...)
...
+                                                    else if (value == "left")
+                                                        state->mouse.handedness(mir_pointer_handedness_left);
+                                                    else
+                                                        mir::log_warning(
+                                                            "Config key '%s' has invalid value: %s",
+                                                            key.to_string().c_str(),
+                                                            val->data());
+                                                }
+                                            },
+                                            {"pointer", "handedness"});
```
mattkae, line 298: Would these also be under `input` as a general category? Just asking as I would do this in miracle regardless

```suggestion
                                            {"input", "pointer", "handedness"});
```

examples/mir_demo_server/server_example.cpp
mattkae: Also: it is kind of funky that each implementation of `AttributeHandler` is forced to implement `on_done` and share a list of `HandleDone`s when it seems like everyone is going to implement it in the same way that you implemented `apply_config`. Do we need this method then? Or can `AttributeHandler` simply implement `apply_config` and all implementations can use it?

mattkae (commented):
Some more minor things, but the direction looks reasonable to me :)

examples/mir_demo_server/server_example.cpp
```diff
@@ -117,7 +118,272 @@ catch (...)
...
+    using HandleInts = std::function<void(ConfigurationKey const& key, std::optional<std::span<int>> value)>;
+    using HandleBool = std::function<void(ConfigurationKey const& key, std::optional<bool> value)>;
+    using HandleBools = std::function<void(ConfigurationKey const& key, std::optional<std::span<bool>> value)>;
+    using HandleFloat = std::function<void(ConfigurationKey const& key, std::optional<float> value)>;
+    using HandleFloats = std::function<void(ConfigurationKey const& key, std::optional<std::span<float>> value)>;
+    using HandleString = std::function<void(ConfigurationKey const& key, std::optional<std::string_view> value)>;
+    using HandleStrings = std::function<void(ConfigurationKey const& key, std::optional<std::span<std::string_view>> value)>;
+    using HandleDone = std::function<void()>;
+
+    virtual void add_int_attribute(HandleInt handler, ConfigurationKey const& key) = 0;
```
tarek-y-ismail, line 209: Minor nit: Why the switch from `(key, handler)` to `(handler,key)`? I believe the former is less frustrating since you have to pass the key upfront, then you can focus on writing the handler. In my experience I tend to forget the key when its the other way around.

tarek-y-ismail:
Right now we load the config file once on `DemoConfigFile` creation (which doesn't do anything since no handlers are added at that point)  and whenever it changes. But we should also be loading at least once after all attributes are added too.

examples/mir_demo_server/server_example.cpp
```diff
@@ -117,7 +118,272 @@ catch (...)
...
+
+        auto operator<=>(State const& that) const { return key <=> that.key; }
+
+        auto to_string() const -> std::string
+        {
+            std::string result;
+            for (auto const& element : path)
+            {
+                if (!result.empty())
+                    result += '_';
```
tarek-y-ismail, line 175: One last thing, do we want to force shell authors into snake_case, or should it be customizable as well? (kebab-case or any:other:case?)

examples/mir_demo_server/server_example.cpp
AlanGriffiths: With the "ini file" implementation there's just one way to do it. But with `AttributeHandler` adapter wrapping a gsettings or YAML node we likely won't keep the adapter around.

examples/mir_demo_server/server_example.cpp
AlanGriffiths: As discussed above I was thinking more along the lines of:

```c++
    InputConfiguration input_configuration{mattkae_configuration.node("input")};
```

But including the "namespace" in the key is certainly an alternative

examples/mir_demo_server/server_example.cpp
AlanGriffiths: IRL we'd want "(handler, key, description)" and "(handler, key, description, default)" in some order I just felt this is the more consistent order, but "(key, description, handler)" and "(key, description, handler, default)" is also possible. Votes anyone?

examples/mir_demo_server/server_example.cpp
AlanGriffiths: They are not _forced_ (in theory they can use `as_path()` and do their own `join()`. The question is what flexibility do _we_ want to support?

AlanGriffiths: Also, I'm not sure we even need `to_string()` - it is convenient for an INI style backend, but pointless for YAML, mysql, gsettings, etc

AlanGriffiths:
Re description: do we want to enter the translation rabbithole? It is one thing to have our keys in English, but descriptions too?

examples/mir_demo_server/server_example.cpp
mattkae: I like the general category too!

mattkae:
> Re description: do we want to enter the translation rabbithole? It is one thing to have our keys in English, but descriptions too?

Is it standard to have those translated in other projects? Perhaps this can be an adapter layer on top?

Saviq:
> > Re description: do we want to enter the translation rabbithole? It is one thing to have our keys in English, but descriptions too?
> 
> Perhaps this can be an adapter layer on top?

Where do we even expect those descriptions to surface? Different shells will have different config storage, and likely their own UI for most of the settings. I don't think this will be useful, really.

EDIT: Any case, translations would be applied on top, the config system should be oblivious.

AlanGriffiths:
> Where do we even expect those descriptions to surface? Different shells will have different config storage, and likely their own UI for most of the settings. I don't think this will be useful, really.

Well, this originates from @mattkae saying ["Each item should include a description ideally"](https://github.com/canonical/mir/pull/3975#discussion_r2114523106). I think the assumption was that a default configuration could be dumped with the description(s) in the comments. Like:

```
# For this to work you will also need host keys in /etc/ssh/ssh_known_hosts
#HostbasedAuthentication no
# Change to yes if you don't trust ~/.ssh/known_hosts for
# HostbasedAuthentication
#IgnoreUserKnownHosts no
# Don't read the user's ~/.rhosts and ~/.shosts files
#IgnoreRhosts yes
```

> EDIT: Any case, translations would be applied on top, the config system should be oblivious.

mattkae:
> > Where do we even expect those descriptions to surface? Different shells will have different config storage, and likely their own UI for most of the settings. I don't think this will be useful, really.
> 
> Well, this originates from @mattkae saying ["Each item should include a description ideally"](https://github.com/canonical/mir/pull/3975#discussion_r2114523106). I think the assumption was that a default configuration could be dumped with the description(s) in the comments. Like:
> 
> ```
> # For this to work you will also need host keys in /etc/ssh/ssh_known_hosts
> #HostbasedAuthentication no
> # Change to yes if you don't trust ~/.ssh/known_hosts for
> # HostbasedAuthentication
> #IgnoreUserKnownHosts no
> # Don't read the user's ~/.rhosts and ~/.shosts files
> #IgnoreRhosts yes
> ```
> 
> > EDIT: Any case, translations would be applied on top, the config system should be oblivious.

Perhaps we don't need descriptions for v1? It might be overkill and be attempting to solve a problem that we don't yet have

PR #3977, opened 2025-05-29
Add the software renderable cursor overlays to their own list

mattkae:
## Why?
As part of the magnification work, I'm going to need to make it so that the `BasicScreenShooter` can selectively not render the cursor (also per the [wlr screencopy protocol](https://wayland.app/protocols/wlr-screencopy-unstable-v1#zwlr_screencopy_manager_v1:request:capture_output)). To do this, we need to develop a better understanding of what is a cursor and what isn't a cursor. I will then open up a follow up PR where we tell the scene "hey, I want everything except the cursors". We will also want a way to temporarily add a software cursor to the render in the event that we have a hardware cursor. Fun stuff :smile: 

ALSO - cursors should not be obfuscated by other overlays (e.g. touchspots), so it makes sense to put it on its own layer. The user pretty much always wants to see the cursor.

## What's new?
- Add `mi::Scene::add_cursor` and `mi::Scene::remove_cursor`
- Added them to the scene at the appropriate time
- Updated the tests

### DIFF
--- a/src/include/server/mir/input/scene.h
+++ b/src/include/server/mir/input/scene.h
@@ -48,10 +48,16 @@ public:
     virtual void remove_observer(std::weak_ptr<scene::Observer> const& observer) = 0;
 
     // An interface which the input stack can use to add certain non-interactive input visualizations
-    // in to the scene (i.e. cursors, touchspots). Overlay renderables will be rendered above all surfaces.
+    // in to the scene (i.e. touchspots). Overlay renderables will be rendered above all surfaces.
     // Within the set of overlay renderables, rendering order is undefined.
     virtual void add_input_visualization(std::shared_ptr<graphics::Renderable> const& overlay) = 0;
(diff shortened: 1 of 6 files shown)


Saviq:
@RAOF had mentioned we should be making the cursor a renderable ~like any other in general, and opportunistically place cursor renderable(s) on the cursor plane(s). Wonder if that takes us a step in that direction, and / or whether we should plan that out to _make_ it such a step?

mattkae:
> @RAOF had mentioned we should be making the cursor a renderable ~like any other in general, and opportunistically place cursor renderable(s) on the cursor plane(s). Wonder if that takes us a step in that direction, and / or whether we should plan that out to _make_ it such a step?

This might take us in that direction, but after talking to Chris, we decided that this direction is wrong for the implementation :) we are going to redesign it such that the cursor is something added directly to the compositor. I suppose it could be pushed onto the proper plane from there...

PR #3978, opened 2025-05-30
platforms/gbm-kms: Enable DRM Universal Planes support.

RAOF:
This is required for pulling the available output colour formats out of KMS, as was enabled in #3950 (oops!).

`DRM_CLIENT_CAP_UNIVERSAL_PLANES` was unconditionally enabled in the 3.16 kernel; that's old enough for us to unconditionally enable it, too.


Saviq (approved):
Sures.

PR #3980, opened 2025-05-30
Expose the cursor's renerable via Cursor::renderable and access it in the compositor itself instead of the scene

mattkae:
## What's new?
- Exposed `Cursor::renderable()` that will return the renderable associated with the cursor, if there is any
- No longer adding the `Cursor::renderable()` to the `Scene`. Instead, adding directly in the compositor
- Updated the associated `SoftwareCursor` implementation and tests for this new interface
- Existing hardware implementations return `nullptr` for now (but I will address this in later work, as I state below!)

## Why?
The `Cursor` is not logically a part of the `Scene` as it is not a surface or anything. Hence, it is up to the compositor to decide whether or not it will display it.

## What's next?
This work isn't done, but this is the minimal reviewable chunk that I want to get in before tackling the rest of it. Coming next:

1. Add `Cursor::needs_compositing` so that the compositor can rely on that boolean flag to decide whether or not to add that `Cursor`'s renderable
2. Implement `Cursor::renderable()` for each hardware cursor (including the Multiplexing cursor!)
3. Render the cursor in the `BasicScreenShooter` based off whether or not an `include_cursor` flag is set

### DIFF
--- a/include/platform/mir/graphics/cursor.h
+++ b/include/platform/mir/graphics/cursor.h
@@ -20,6 +20,7 @@
 
 #include "mir/geometry/size.h"
 #include "mir/geometry/point.h"
+#include "mir/graphics/renderable.h"
 
 #include <memory>
 
@@ -28,6 +29,7 @@ namespace mir
 namespace graphics
 {
 class CursorImage;
+class Renderable;
 class Cursor
 {
 public:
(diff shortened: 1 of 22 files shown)


RAOF (approved):
That looks very pleasant! Let the subsequent PRs roll!

include/platform/mir/graphics/cursor.h
```diff
@@ -20,6 +20,7 @@
 
 #include "mir/geometry/size.h"
 #include "mir/geometry/point.h"
+#include "mir/graphics/renderable.h"
```
AlanGriffiths, line 23: I don't think we need this - we introduce a declaration on L32 that should be enough

PR #3982, opened 2025-06-02
Implement `Cursor::needs_compositing` so that the compositor can selectively grab a renderable for a cursor

mattkae:
## What's new?
- Implement `Cursor::needs_compositing` on all platforms
- Write unit tests for it


include/platform/mir/graphics/cursor.h
```diff
@@ -41,6 +41,8 @@ class Cursor
     virtual void scale(float new_scale) = 0;
 
     virtual auto renderable() -> std::shared_ptr<Renderable> = 0;
+
+    virtual auto needs_compositing() -> bool = 0;
```
AlanGriffiths: I'd expect this to be `const`

src/server/compositor/multi_threaded_compositor.cpp
```diff
@@ -182,8 +183,14 @@ class CompositingFunctor
...
                         auto& compositor = std::get<1>(tuple);
                         auto scene_elements = scene->scene_elements_for(compositor.get());
-                        if (auto const cursor_renderable = cursor->renderable())
-                            scene_elements.push_back(std::make_shared<CursorSceneElement>(cursor_renderable));
+                        if (cursor->needs_compositing())
+                        {
+                            if (auto const cursor_renderable = cursor->renderable())
+                                scene_elements.push_back(std::make_shared<CursorSceneElement>(cursor_renderable));
+                            else
+                                log_error("Cursor needs compositing by the renderable is unavailable");
```
AlanGriffiths: Is this a condition that will correct itself? Or is this likely to spam the logs if it occurs?

```suggestion
                                log_error("Cursor needs compositing, however no renderable is available");
```

Also, should `!needs_compositing() || renderable()` be an invariant maintained by the cursor implementation (as opposed to being checked by the compositor. I'm thinking of an interface more like: 

```c++
    cursor->if_needs_compositing([&](auto const* cursor_renderable) {
        scene_elements.push_back(std::make_shared<CursorSceneElement>(cursor_renderable));
    });
```

include/platform/mir/graphics/cursor.h
AlanGriffiths: Although, I'm not sure this is the best API design anyway. Any logic around needing compositing and supplying a renderable belongs inside the cursor class.

AlanGriffiths (changes requested):
"rather than asking an object for data and acting on that data, we should instead tell an object what to do"

https://martinfowler.com/bliki/TellDontAsk.html

src/server/compositor/multi_threaded_compositor.cpp
mattkae: The proposed API does not solve our core issue unfortunately. In the "normal" compositor, we want to composite a renderable only if it needs compositing. Hence, we leave the drawing of hardware cursors to the platforms.

On the other hand, during screen capture we want to get renderables for cursors that _don't_ need compositing. For example, the `gbm-kms` cursor does not typically need compositing, but we are going to read its buffer into a `Renderable` when the `ScreenShooter` needs to render it so that we're capturing the right thing.

This is why they are separated

mattkae:
@AlanGriffiths re-requesting before I change anything else because of the my response to your main architectural concern

src/server/compositor/multi_threaded_compositor.cpp
RAOF: > Is this a condition that will correct itself? Or is this likely to spam the logs if it occurs?

I would not expect this to resolve (and, indeed, once everything has landed I would not expect this to be *possible*; every `Cursor` should provide a non-null `renderable()`)

-----

Fundamentally, `needs_compositing()` is slightly badly named - it's actually the negation of `doesnt_need_compositing_on_my_hardware_DisplaySink()`.
You *could* have a `cursor->if_needs_compositing(DisplaySink& target, Functor)` interface, but that would be awkward.

We know in `MultiThreadedCompositor` that we're compositing onto a hardware display sink, but in the screen copy code we're compositing onto a virtual display sink.

In Magical Future Mir™, where we drop explicit hardware cursor support and just opportunistically use planes everywhere, this goes back the other way and we just unconditionally dump the `Cursor` renderable into the list of `Renderable`s.

AlanGriffiths: > In Magical Future Mir™, where we drop explicit hardware cursor support and just opportunistically use planes everywhere, this goes back the other way and we just unconditionally dump the `Cursor` renderable into the list of `Renderable`s.

Except, of course, when we screen capture "without cursor". ;)

AlanGriffiths: > The proposed API does not solve our core issue unfortunately. In the "normal" compositor, we want to composite a renderable only if it needs compositing. Hence, we leave the drawing of hardware cursors to the platforms.
> 
> On the other hand, during screen capture we want to get renderables for cursors that _don't_ need compositing. For example, the `gbm-kms` cursor does not typically need compositing, but we are going to read its buffer into a `Renderable` when the `ScreenShooter` needs to render it so that we're capturing the right thing.

There are several ways to address that. The simplest being `ScreenShooter` uses the existing unconditional API. Another is:

```c++
    cursor->unless_using_cursor_plane(on_XXX, [&](auto const* cursor_renderable) {
        scene_elements.push_back(std::make_shared<CursorSceneElement>(cursor_renderable));
    });
```

(Or two different functions.)

> Fundamentally, `needs_compositing()` is slightly badly named - it's actually the negation of `doesnt_need_compositing_on_my_hardware_DisplaySink()`.
> You _could_ have a `cursor->if_needs_compositing(DisplaySink& target, Functor)` interface, but that would be awkward.

Does that imply that `needs_compositing()` is not a property of the cursor alone? That is, with a hybrid system it could depend on which card is compositing? I don't see how the current proposal addresses that.

AlanGriffiths (changes requested):
While I have concerns over the API being offered, my blocking concern is about the potential for logging on every composition pass.

If the logging is just an intermediate state, and the condition will be a `fatal_error()` eventually then, maybe, a TODO explanation could be sufficient. Or, if we only need the logging for development, it could enabled by an environment variable?

mattkae:
I will test the Wayland bits later when I implement `renderable` there

src/server/compositor/multi_threaded_compositor.cpp
RAOF: > > Fundamentally, `needs_compositing()` is slightly badly named - it's actually the negation of `doesnt_need_compositing_on_my_hardware_DisplaySink()`.
> > You _could_ have a `cursor->if_needs_compositing(DisplaySink& target, Functor)` interface, but that would be awkward.
> 
> Does that imply that `needs_compositing()` is not a property of the cursor alone? That is, with a hybrid system it could depend on which card is compositing? I don't see how the current proposal addresses that.

Ah, sorry for the confusion here.

At the level of the individual `DisplayPlatform`s (from which the `DisplaySink`s are derived), yes, this could be a property of which card is compositing *but* when we create a `MultiplexingCursor` it only succeeds if *all* `DisplayPlatform`s can create a hardware cursor.

So in principle `needs_compositing()` is a property of the `DisplaySink`¹ the cursor is on but in practise we only attempt to use hardware cursors where the answer is “yes” for every `DisplaySink`.

¹: It's *possible*, although I've never seen a real example, that this could be a per-output rather than per-card; the KMS API can describe hardware for which “these outputs have a HW cursor, these other outputs do not” is true.

src/server/graphics/multiplexing_hw_cursor.cpp
```diff
@@ -75,6 +75,16 @@ void mir::graphics::MultiplexingCursor::scale(float new_scale)
 
 auto mg::MultiplexingCursor::renderable() -> std::shared_ptr<Renderable>
 {
-    return nullptr;
+    if (platform_cursors.empty())
+        return nullptr;
```
RAOF: Can we actually hit this case?
&lt;Checks&gt;
No, no we cannot.

We should remove these checks.

tests/unit-tests/graphics/test_multiplexing_cursor.cpp
```diff
@@ -0,0 +1,130 @@
...
+    mg::MultiplexingCursor multiplexing_cursor(displays);
+    multiplexing_cursor.scale(scale);
+}
+
+TEST_F(MultiplexingCursorTest, renderable_is_null_without_any_cursors)
+{
+    std::vector<mg::Display*> displays;
+    mg::MultiplexingCursor multiplexing_cursor(displays);
+    EXPECT_THAT(multiplexing_cursor.renderable(), IsNull());
+}
```
RAOF: This is actually a precondition failure

RAOF (changes requested):

tests/unit-tests/graphics/test_multiplexing_cursor.cpp
AlanGriffiths: Looks like a POSTcondition to me. (It checks state after the call, not before)

include/platform/mir/graphics/cursor.h
AlanGriffiths: > I'd expect this to be `const`

Any reason it might mutate the cursor?

src/server/compositor/multi_threaded_compositor.cpp
```diff
@@ -182,8 +183,12 @@ class CompositingFunctor
                     {
                         auto& compositor = std::get<1>(tuple);
                         auto scene_elements = scene->scene_elements_for(compositor.get());
-                        if (auto const cursor_renderable = cursor->renderable())
-                            scene_elements.push_back(std::make_shared<CursorSceneElement>(cursor_renderable));
+                        if (cursor->needs_compositing())
+                        {
+                            if (auto const cursor_renderable = cursor->renderable())
+                                scene_elements.push_back(std::make_shared<CursorSceneElement>(cursor_renderable));
```
AlanGriffiths, line 189: Following previous discussions: is `!cursor->renderable()` possible? I thought we decided that that violates a postcondition? (Is this proposing an intermediate form where we ignore the postcondition violation? Or is it not a post condition?)

AlanGriffiths: Looking at:

```c++
auto mg::SoftwareCursor::needs_compositing() -> bool
{
    std::lock_guard lg{guard};
    return visible;
}
```

and

```c++
auto mg::SoftwareCursor::renderable() -> std::shared_ptr<Renderable>
{
    std::lock_guard lg{guard};
    if (!visible)
        return nullptr;

    return renderable_;
}
```

Then `!cursor->renderable()` is definitely possible as `visible` may change between the calls.

mattkae: Yup that's what I realized later on. It can change if the `renderable` is destroyed or `visible` is no longer true

include/test/mir/test/doubles/mock_cursor.h
```diff
@@ -36,7 +36,7 @@ struct MockCursor : public mir::graphics::Cursor
     MOCK_METHOD1(move_to, void(mir::geometry::Point));
     MOCK_METHOD1(scale, void(float));
     MOCK_METHOD0(renderable, std::shared_ptr<mir::graphics::Renderable>());
-    MOCK_METHOD0(needs_compositing, bool());
+    MOCK_CONST_METHOD0(needs_compositing, bool());
```
AlanGriffiths: We're trying to move away from the ...[digit] version of this macro

AlanGriffiths (approved):
Okay

include/test/mir/test/doubles/mock_cursor.h
AlanGriffiths: Come on! You didn't need to change this, but as you chose to, you could have fixed the surrounding code

PR #3984, opened 2025-06-03
Merge libmiroil7.symbols.armhf into libmiroil7.symbols

AlanGriffiths:
Consolidate the libmiroil symbols too


AlanGriffiths:
> File "/<<PKGBUILDDIR>>/src/miroil/check-and-update-debian-symbols.py", line 149, in check_symbols
>     result.assert_success() # Something's not right. This will fail with a useful error

Oops! Fixed

PR #3985, opened 2025-06-04
Pipe the graphcis buffer allocator to Display::create_hardware_cursor

mattkae:


mattkae:
> Hm. I don't think you need to pass the `GraphicBufferAllocator` in to the hardware cursor implementations. Now that `ShmBuffer`s are entirely independent of the rendering platform you could just create a `MappableBackedShmBuffer` in the `Cursor` implementation.
> 
> This would want a. A helper to create `RWMappableBuffer` backed by a `uint8_t*` you've already got (in `argb8888.data()`), and b. _Might_ need `shm_buffer.h` moved somewhere else in the source tree?

Yup totally true! Even better, we can just have a single `Buffer` and reference its data instead of using that `uint8` array that we had previously. Much easier :tada:

src/platforms/gbm-kms/server/kms/cursor.cpp
```diff
@@ -184,15 +186,15 @@ void mgg::Cursor::pad_and_write_image_data_locked(
...
+    auto const buffer_stride = std::max(min_width*4, gbm_bo_get_stride(gbm_buffer));  // in bytes
+    auto const buffer_height = std::max(min_height, gbm_bo_get_height(gbm_buffer));
     size_t const padded_size = buffer_stride * buffer_height;
 
     auto padded = std::unique_ptr<uint8_t[]>(new uint8_t[padded_size]);
     size_t rhs_padding = buffer_stride - 4*image_width;
 
     auto const filler = 0; // 0x3f; is useful to make buffer visible for debugging
-    uint8_t const* src = argb8888.data();
+    uint8_t const* src = buffer->map_readable()->data();
```
RAOF: Hm. This (accidentally) works, but you should really be keeping the `Mapping` around for the duration of use.

(Particularly, this will immediately call `~Mapping`, which will call `mark_dirty`, but before you've actually written any data)

RAOF:
```suggestion
    auto mapping = buffer->map_readable();
    uint8_t const* src = mapping->data();
```

mattkae: Will we me marking it dirty with a `map_readable` though? I will still keep it around during the read however :)

mattkae: Whoops! We should use `map_writeable` huh!

src/platforms/gbm-kms/server/kms/cursor.cpp
```diff
@@ -184,15 +186,16 @@ void mgg::Cursor::pad_and_write_image_data_locked(
...
+    auto const buffer_stride = std::max(min_width*4, gbm_bo_get_stride(gbm_buffer));  // in bytes
+    auto const buffer_height = std::max(min_height, gbm_bo_get_height(gbm_buffer));
     size_t const padded_size = buffer_stride * buffer_height;
 
     auto padded = std::unique_ptr<uint8_t[]>(new uint8_t[padded_size]);
     size_t rhs_padding = buffer_stride - 4*image_width;
 
     auto const filler = 0; // 0x3f; is useful to make buffer visible for debugging
-    uint8_t const* src = argb8888.data();
+    auto const mapping = buffer->map_writeable();
```
RAOF: Uuuh...
```suggestion
    auto const mapping = buffer->map_readable();
```
:woman_facepalming:

src/platforms/gbm-kms/server/kms/cursor.cpp
RAOF: No, you're right, it *should* be `map_readable` (we're only reading from it; my comment about `mark_dirty` was misleading).

We still should be keeping the `mapping` alive as long as we're using the pointer; the interface only guarantees the pointer is valid as long as the mapping is alive.

PR #3986, opened 2025-06-05
Enable generation of compile_commands.json for use by clangd

jhenstridge:
This improves the experience for people using editors with LSP support, and file is small relative to a full Mir build tree so there is not much harm in generating it in cases where it won't be used.


RAOF (approved):
Yeah, I don't think there's any reason not to generate this file.

PR #3987, opened 2025-06-05
ci: only run TICS when we've secrets

Saviq:


.github/workflows/tics.yml
```diff
@@ -17,6 +17,9 @@ concurrency:
 
 jobs:
   TICS:
+    # Only run if we have access to secrets.
+    if: ${{ github.event_name != 'pull_request' || github.repository == github.event.pull_request.head.repo.full_name }}
```
jhenstridge: The outer `${{ ... }}` shouldn't be needed here.

I wonder if it could be simplified further to just check the secret directly? Something like:

    if: secrets.TICSAUTHTOKEN != ''

RAOF: That would mean that accidental failures to acquire the auth token would be silent, right? As in, TICS would just not run, and we'd need to notice the absence of a run where we'd expect it?

jhenstridge: Good point. I'd still recommend removing the `${{ ... }}` bit, since we don't need string interpolation here.

Saviq: Done.

jhenstridge (approved):
Looks good!

PR #3991, opened 2025-06-05
Implement Renderable::orientation

mattkae:
## What's new?
- Added `Renderable::orientation`
- Implemented in the various places that `Renderable` is implemented
- Added tests for each implementation


AlanGriffiths (approved):
Sure

PR #3992, opened 2025-06-06
HACKING.md: Add some intructions for MESA debugging

RAOF:


AlanGriffiths:
@robert-ancell, as we're spread across time zones, the approver is encouraged to queue merges rather than waiting for the world to turn. (Of course there are lots of cases where that's not appropriate, but this doesn't seem like one.)

PR #3993, opened 2025-06-06
server/frontend_xwayland: use the XResource extension to identify X clients

### LINKED ISSUE #2060: Use XResource extension to find client PID where available
The [X-Resource 1.2](https://www.x.org/releases/X11R7.7/doc/resourceproto/resproto.txt) extension has a request for getting the PID of a client. Unlike `_NET_WM_PID` which we currently use, this doesn't require the client to behave; it works for all clients (at least, those running on the same machine, for which the PID makes sense).

jhenstridge:
Previously the _NET_WM_PID property was used for this purpose, which is relies on clients to (a) set the property and (b) not lie. In contrast, the process ID provided by the [XResource extension](https://www.x.org/releases/X11R7.7/doc/resourceproto/resproto.txt) is based on peer credentials for the X11 connection.

I couldn't find any existing tests covering the existing behaviour, but it does appear to be determining the correct pids when I tested it in miral-app with some extra debug logging.

It also looks like the extension would allow us to group remote clients by their X client ID. I'm not sure whether that would be useful or not (especially since it's unlikely that XWayland will ever be listening on TCP).

Fixes #2060


Saviq:
> [src/server/frontend_xwayland/xwayland_surface.cpp](https://canonical.tiobe.com/tiobeweb/TICS/AnnotatedSource.html#axes=Suppressions(no),Date(1749181795),ClientData(ubuntu%3AykxjTfLAFEopXb2hYCtC),Project(mir),Window(changed,ge,1749181795),Level(Set(1,2,3)),DeltaDate(Run(-2)),DiffType(new),File(Path(HIE,mir,master,src,server,frontend_xwayland,xwayland_surface.cpp))&diff=true&metrics=G(Diff(CS,DiffType(new)),Level(Set(1,2,3))))  	0	+8
> [src/server/frontend_xwayland/xcb_connection.cpp](https://canonical.tiobe.com/tiobeweb/TICS/AnnotatedSource.html#axes=Suppressions(no),Date(1749181795),ClientData(ubuntu%3AykxjTfLAFEopXb2hYCtC),Project(mir),Window(changed,ge,1749181795),Level(Set(1,2,3)),DeltaDate(Run(-2)),DiffType(new),File(Path(HIE,mir,master,src,server,frontend_xwayland,xcb_connection.cpp))&diff=true&metrics=G(Diff(CS,DiffType(new)),Level(Set(1,2,3))))  	0	+7

These are preexisting, I've reported the issue to TIOBE.

RAOF (approved):
Does what it says on the tin.

I don't think we should block on this being tested (we *should* look at adding some automated testing for the this XWayland code, but I believe that's low-priority)

PR #3995, opened 2025-06-06
live configuration facilities

AlanGriffiths:
This is the first part of providing a pluggable live configuration mechanism

This PR provides 

1. the `miral::live_config::Store` API to be implemented over a configuration mechanism; and,
2. Constructors for  `OutputFilter`, `CursorScale` and `InputConfiguration` that register for configuration changes

There's also a POC .ini file style implementation in `mir_demo_server`.

Further PRs will provide implementations of `miral::live_config::Store` based on ".ini file" (replacing the POC with something complete and product ready), YAML and gsettings.

Shells will be able to choose between these implementations and rolling their own

### DIFF
--- a/examples/mir_demo_server/server_example.cpp
+++ b/examples/mir_demo_server/server_example.cpp
@@ -123,9 +123,9 @@ catch (...)
 // Struct for illustrative purposes, this would be rolled into miral::OutputFilter
 struct OutputFilter : miral::OutputFilter
 {
-    explicit OutputFilter(mlc::Store& config_handler)
+    explicit OutputFilter(mlc::Store& config_store)
     {
-        config_handler.add_string_attribute({"output_filter"}, "Output filter to use [{none,grayscale,invert}]",
+        config_store.add_string_attribute({"output_filter"}, "Output filter to use [{none,grayscale,invert}]",
                                             [this](mlc::Key const& key, std::optional<std::string_view> val)
                                             {
                                                 MirOutputFilter new_filter = mir_output_filter_none;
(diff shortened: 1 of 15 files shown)


src/miral/cursor_scale.cpp
```diff
@@ -51,6 +53,29 @@ miral::CursorScale::CursorScale()
...
 
+miral::CursorScale::CursorScale(live_config::Store& config_store) : miral::CursorScale{}
+{
+    config_store.add_float_attribute(
+        {"cursor", "scale"},
+        "Cursor scale",
+       [this](live_config::Key const& key, std::optional<float> val)
+        {
+            if (val)
+            {
```
mattkae, line 66: Question: Should we reset to the default when a value isn't provided? Is there any facility for resetting to the "default" whatever that might be (e.g. an "unset").

I ask, because I would expect removing a key from my YAML config entirely to "reset" it to default

src/miral/input_configuration.cpp
```diff
@@ -91,6 +93,36 @@ class miral::InputConfiguration::Self
...
+
+    struct Config
+    {
+        Config(miral::InputConfiguration::Mouse mouse,
+        miral::InputConfiguration::Touchpad touchpad,
+        miral::InputConfiguration::Keyboard keyboard) :
+        mouse{mouse},
+        touchpad{touchpad},
+        keyboard{keyboard}
+        {}
```
mattkae:
```suggestion
        Config(miral::InputConfiguration::Mouse mouse,
            miral::InputConfiguration::Touchpad touchpad,
            miral::InputConfiguration::Keyboard keyboard)
            : mouse{mouse},
              touchpad{touchpad},
              keyboard{keyboard}
        {}
```

src/miral/cursor_scale.cpp
AlanGriffiths: Aren't you asking this in the wrong place? Your YAML config should certainly DTRT.

Or do you mean that `CursorScale` ought to be treating `nullopt` as the default?

mattkae: Yes I mean the latter.

What might happen is:
1. I parse my YAML config
2. I see that `cursor: scale...` is no longer specified
3. I pass `std::nullopt` to the handler and expect it to return to "default" (i.e. `1`)

AlanGriffiths: > Yes I mean the latter.
> 
> What might happen is:
> 
>     1. I parse my YAML config
> 
>     2. I see that `cursor: scale...` is no longer specified
> 
>     3. I pass `std::nullopt` to the handler and expect it to return to "default" (i.e. `1`)

In this scenario, why would the handler pass `nullopt` instead of the "default"?

Look at it from the other side of the contract:

1. I tell the configuration implementation that the default is `1`
2. "Stuff" happens
3. The configuration implementation tells me a new value, I use that value

mattkae: I agree with that. Can we query for the default for a handler?

AlanGriffiths: > Can we query for the default for a handler?

```C++
    HandleInt handler_with_default = [preset, old_handler](Key const& key, std::optional<int> value)
        {
             old_handler(key, value.or_else(preset));
        }
```

AlanGriffiths: I think the takeaway from this discussion is that the semantics of `preset` in this interface are unclear.

I can think of two interpretations:

1. A value with which to populate a new key in a config
2. A value to supply when nothing else is available

It is also not clear whether handlers for keys that have "defaults" should take an `optional`.

I need to think a bit more

AlanGriffiths: > I need to think a bit more

I think this covers it: https://github.com/canonical/mir/pull/3995/commits/985c67a61c62bfc664b63ec8168f3a2ce181f825

examples/mir_demo_server/server_example.cpp
```diff
@@ -117,56 +120,74 @@ catch (...)
...
+
+    std::map<live_config::Key, HandleString> attribute_handlers;
+    std::list<HandleDone> done_handlers;
+
     std::mutex config_mutex;
-    miral::CursorScale cursor_scale;
-    miral::OutputFilter output_filter;
-    miral::ConfigFile config_file;
+    std::optional<miral::ConfigFile> config_file;
+    std::function<void()> const init_part2;
```
mattkae, line 185: Nit: Why is this called `init_part2`? Seems like a weird name :)

examples/mir_demo_server/server_example.cpp
```diff
@@ -117,56 +120,74 @@ catch (...)
 {
 }
 
-class DemoConfigFile
+// for illustrative purposes
```
mattkae, line 123: Is this a necessary comment?

examples/mir_demo_server/server_example.cpp
mattkae: Maybe call it `on_initial_config_ready`?

mattkae (approved):
I have two non-blocking pieces of feedback, but LGTM

AlanGriffiths:
> I have two non-blocking pieces of feedback, but LGTM

Queuing for merge as that feedback belongs to the PoC implementation in mir_demo_server. That should be superseded by an "INI file" configuration file implementation in libmiral. (In the next few days...)

Issue #3937, opened 2025-05-08
Should add `NullInputDeviceObserver`


tarek-y-ismail:
Multiple places in code and tests define on-off observers that inherit from `mir::input::InputDeviceObserver`, usually not caring about anything other than `device_added`. We should add `NullInputDeviceObserver` and update these structs to use it to reduce code noise.

panagiotisevaggelou:
Hi @tarek-y-ismail, I’ve opened [PR #3999](https://github.com/canonical/mir/pull/3999) to address this by adding NullInputDeviceObserver and updating related code. 😊

PR #3999, opened 2025-06-09
feature: add NullInputDeviceObserver to simplify default server setup

panagiotisevaggelou:
Fixes issue #3937: Introduces NullInputDeviceObserver as a no-op implementation of mir::input::InputDeviceObserver to simplify default and test setups.

Changes
- Added NullInputDeviceObserver with empty method overrides
- Replaced redundant anonymous observer implementations
- Cleaned up related usage for clarity and maintainability


panagiotisevaggelou:
I updated the symbols.map file accordingly, but the pipeline also failed for this job: [BuildAndTest (lxd:fedora-rawhide:spread/build/fedora:amd64)](https://github.com/canonical/mir/actions/runs/15535602915/job/43740378092?pr=3999#logs)

Saviq:
> the pipeline also failed for this job: [BuildAndTest (lxd:fedora-rawhide:spread/build/fedora:amd64)](https://github.com/canonical/mir/actions/runs/15535602915/job/43740378092?pr=3999#logs)

That's a known issue, not a problem in your code :+1:. Thanks for the contribution!

I'll let Tarek progress this tomorrow :bow:.

panagiotisevaggelou:
@tarek-y-ismail I have rebased my commit with the extra changes. Thanks for pointing that out!

src/server/input/config_changer.cpp
```diff
@@ -102,7 +103,7 @@ MirInputDevice get_device_config(mi::Device const& device)
     return cfg;
 }
 
-struct DeviceChangeTracker : mi::InputDeviceObserver
+struct DeviceChangeTracker : mi::NullInputDeviceObserver
```
AlanGriffiths: This is not a good idea in many cases.

By deriving directly from `InputDeviceObserver` we ensure that any changes (e.g. adding a member function) to the interface cause a compile error here and force someone to evaluate whether there are changes needed to `DeviceChangeTracker`.

With the proposed change, that function is implemented by `NullInputDeviceObserver` and there is a risk that nobody looks at `DeviceChangeTracker`

AlanGriffiths (changes requested):
Sorry, but I disagree with @tarek-y-ismail about the correct use of `NullInputDeviceObserver`.

I would suggest only using `NullInputDeviceObserver` as a base class in test code. (Maybe also move the header into `include/tests` to avoid misuse in future?)

src/server/input/config_changer.cpp
panagiotisevaggelou: Just to make sure everything is clear: I created a commit that reverts the usage of NullInputDeviceObserver in the main code, and keeps it only in the tests (also moved the header into include/test/mir/test). Is this acceptable?

src/server/symbols.map
```diff
@@ -404,6 +404,11 @@ global:
     mir::input::KeyboardObserver::?KeyboardObserver*;
     mir::input::KeyboardObserver::KeyboardObserver*;
     mir::input::KeyboardObserver::operator*;
+    mir::input::NullInputDeviceObserver::?NullInputDeviceObserver*;
+    mir::input::NullInputDeviceObserver::changes_complete*;
+    mir::input::NullInputDeviceObserver::device_added*;
+    mir::input::NullInputDeviceObserver::device_changed*;
+    mir::input::NullInputDeviceObserver::device_removed*;
```
AlanGriffiths: 1. these entry points do not exist in libmirserver
2. if they did, they would not belong in the 2.21 stanza (as they are not part of the 2.21 release)

panagiotisevaggelou: I used `cmake --build build --target generate-mirserver-symbols-map` to regenerate the symbols.map, and since the entries were still present I assumed they should stay. Thanks for the explanation, fixed it!

AlanGriffiths (approved):
I'm OK with this. But as it is @tarek-y-ismail's issue I'll leave final approval to him

tarek-y-ismail (approved):
Looks good. Alan's comment didn't really pop up in my mind before. Sorry for initially misleading you!

PR #4001, opened 2025-06-10
Update linux_dmabuf_v1 to version 5.

robert-ancell:


src/platform/graphics/linux_dmabuf.cpp
```diff
@@ -856,6 +858,78 @@ class LinuxDmaBufParams : public mir::wayland::LinuxBufferParamsV1
...
+        auto d = static_cast<dev_t*>(wl_array_add(&main_device, sizeof(dev_t)));
+        *d = this->provider->devnum();
+        send_main_device_event(&main_device);
+
+        // We only currently support one device, which accessess all formats.
+        wl_array device;
+        wl_array_init(&device);
+        d = static_cast<dev_t*>(wl_array_add(&device, sizeof(dev_t)));
+        *d = this->provider->devnum();
+        send_tranche_target_device_event(&device);
```
RAOF: `wl_array` is really bad, and I feel we could probably do with a helper wrapping this up.

That said, not blocking here.

src/platform/graphics/linux_dmabuf.cpp
```diff
@@ -856,6 +858,78 @@ class LinuxDmaBufParams : public mir::wayland::LinuxBufferParamsV1
...
+        {
+            auto format = formats[i];
+            for (auto j = 0u; j < format.modifiers.size(); ++j)
+            {
+                data[k].format = format.format;
+                data[k].padding = 0;
+                data[k].modifier = format.modifiers[j];
+                k++;
+            }
+        }
```
RAOF, line 911: Hm. I was *going* to suggest some fancier range-based C++ here, but that's not possible because `DmaBufFormatDescriptors` doesn't model a range.

And now is not the time to be implementing that.

RAOF (changes requested):
This is a `libmirplatform` ABI break, so we need to do that dance.

Otherwise looks good (and works, too). Can we get a WLCS test? :wink:

src/platform/graphics/linux_dmabuf.cpp
AlanGriffiths: Something like `CopyableWlArray`?

robert-ancell:
I've updated the ABI (please check if correct) and taken a task to look at WLCS next pulse.

RAOF:
Sorry for not getting to this today! I got waylaid while trying to test it by #4047

src/platform/symbols.map
```diff
@@ -55,9 +56,9 @@ MIR_PLATFORM_2.21 {
     mir::graphics::has_egl_extension*;
     mir::graphics::has_egl_client_extension*;
     mir::graphics::LinearGammaLUTs::LinearGammaLUTs*;
-    mir::graphics::LinuxDmaBufUnstable::?LinuxDmaBufUnstable*;
-    mir::graphics::LinuxDmaBufUnstable::LinuxDmaBufUnstable*;
-    mir::graphics::LinuxDmaBufUnstable::buffer_from_resource*;
+    mir::graphics::LinuxDmaBuf::?LinuxDmaBuf*;
+    mir::graphics::LinuxDmaBuf::LinuxDmaBuf*;
+    mir::graphics::LinuxDmaBuf::buffer_from_resource*;
```
AlanGriffiths, line 59: These changes do not belong in the 2.21 stanza (that ship has sailed)

src/platform/graphics/linux_dmabuf.cpp
AlanGriffiths, line 1: I think we are lacking changes to determine the protocol version the client binds to. As proposed we will calling `send_format_event()` regardless of the prohibition:

>         Starting version 4, the format event is deprecated and must not be
>         sent by compositors. Instead, use get_default_feedback or
>         get_surface_feedback.

AlanGriffiths (changes requested):
Are we safe dropping `linux_dmabuf_unstable_v1`? Won't we encounter clients that expect the old protocol extension?

robert-ancell:
> Are we safe dropping `linux_dmabuf_unstable_v1`? Won't we encounter clients that expect the old protocol extension?

The interface name hasn't changed - it is still "zwp_linux_dmabuf_v1". The protocol name dropped the unstable suffix, but this only affects the .xml file name. So all existing clients will continue to work.

src/platform/graphics/linux_dmabuf.cpp
robert-ancell: The LinuxDmabuf::Instance constructor was updated to do this:

```
        if (wl_resource_get_version(new_resource) < 4)
        {
            send_formats();
        }
```

src/platform/symbols.map
robert-ancell: Moved

robert-ancell:
This was blocked by TICS stating that the addition to `src/platform/graphics/egl_extensions.cpp` not having code coverage.

src/platform/symbols.map
AlanGriffiths, line 1: We're breaking ABI, so we can consolidate the symbols into a single 2.22 stanza.

(Sorry, didn't think of that last time)

AlanGriffiths (approved):

PR #4007, opened 2025-06-11
platforms/eglstream-kms: Update for ShmBuffer interface changes

RAOF:
This is a follow up for #3968, fixing an immediate crash on eglstream-kms.


RAOF:
The expectation was that we would be dropping eglstream-kms before this mattered, but we've (correctly) pulled #3968 into 2.21.

Saviq:
Lab confirmed :+1:

PR #4008, opened 2025-06-11
Don't raise the parent of windows sent to the back

AlanGriffiths:
Not sure why this code was here. Doesn't seem to belong


Saviq:
@mattkae this your doing, thoughts?

- https://github.com/canonical/mir/pull/2981

PR #4009, opened 2025-06-11
Applying the inverse orientation of the renderable at render time

mattkae:


include/platform/mir/graphics/transformation.h
```diff
@@ -35,6 +35,11 @@ inline glm::mat2 transformation(MirOrientation ori)
     return glm::mat2(cos, sin, -sin, cos);  // column-major order, GL-style
 }
 
+inline glm::mat2 inverse_transformation(MirOrientation ori)
+{
+    return glm::inverse(transformation(ori));
```
robert-ancell: Note this is more expensive than using a switch statement for the four cases, but since it's a 2x2 matrix I guess not an issue?

tests/unit-tests/renderers/gl/test_gl_renderer.cpp
```diff
@@ -230,6 +235,41 @@ TEST_F(GLRenderer, avoids_src_alpha_for_rgbx_blending)  // LP: #1423462
...
+                    mat[i][j] = ptr[i * 4 + j];
+                }
+            }
+            glm::mat4 const expected = trans
+                * glm::mat4(mir::graphics::inverse_transformation(mir_orientation_left));
+            for (size_t i = 0; i < 4; i++)
+            {
+                for (size_t j = 0; j < 4; j++)
+                {
+                    printf("%f, %f\n", mat[i][j], expected[i][j]);
```
robert-ancell: Debugging code left in?

robert-ancell:
This seems to be lacking the comment to describe why this is being done - was orientation broken for all cases before this?

tests/unit-tests/renderers/gl/test_gl_renderer.cpp
```diff
@@ -230,6 +235,41 @@ TEST_F(GLRenderer, avoids_src_alpha_for_rgbx_blending)  // LP: #1423462
...
+                }
+            }
+
+            return mat == expected;
```
RAOF: Would it be simpler to multiply the matrix by the `orientation` matrix and expect to get the identity matrix?

ie:
```
return (mat * mir::graphics::transformation(mir_orientation_left)) == glm::mat4{1.0f};
```
?

tests/unit-tests/renderers/gl/test_gl_renderer.cpp
```diff
@@ -230,6 +235,41 @@ TEST_F(GLRenderer, avoids_src_alpha_for_rgbx_blending)  // LP: #1423462
     renderer.render(renderable_list);
 }
 
+TEST_F(GLRenderer, applies_inverse_orientation_matrix)
```
RAOF, line 239: I'm not wild about the rigidity of this test - if we change any of the call sequence in `Renderer` this will spuriously fail.

That said, I don't have a great alternative, either.

RAOF:
I don't understand TICS' objection, either. There's an extra line of code, but that line of code *should* be hit by the tests? @Saviq?

Saviq:
> I don't understand TICS' objection, either.

I _think_ it's getting confused about what's old, what's new code, and what it compares against…

Thanks for flagging, another one for discussion.

tests/unit-tests/renderers/gl/test_gl_renderer.cpp
mattkae: Yeah I did not like this test either :) I figured we _should_ test it... but I didn't know of any better way

include/platform/mir/graphics/transformation.h
mattkae: Yeah to be honest, maybe it's worth just making it a straight look up :+1:

robert-ancell (approved):
Looks great!

robert-ancell:
I'll leave it to you @mattkae to merge, not sure if you're still chasing down CI issues.

RAOF (approved):

PR #4011, opened 2025-06-12
Implement Cursor::renderable for the atomic-kms platform

mattkae:
## What's new?
- Gave the `atomic-kms` cursor the same implementation as the `gbm-kms` cursor
- Added our first `atomic-kms` tests for the new methods

### DIFF
--- a/src/platforms/atomic-kms/server/kms/cursor.cpp
+++ b/src/platforms/atomic-kms/server/kms/cursor.cpp
@@ -19,6 +19,7 @@
 #include "kms_output.h"
 #include "kms_output_container.h"
 #include "kms_display_configuration.h"
+#include "shm_buffer.h"
 #include "mir/geometry/rectangle.h"
 #include "mir/graphics/cursor_image.h"
 #include "mir/graphics/pixman_image_scaling.h"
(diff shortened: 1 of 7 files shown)


robert-ancell (approved):
This looks good to me, but I suspect you'll want a second opinion from @RAOF

src/platforms/atomic-kms/server/kms/cursor.cpp
```diff
@@ -455,10 +460,88 @@ void mir::graphics::atomic::Cursor::scale(float new_scale)
...
+
+    return std::make_shared<CursorRenderable>(
+        buffer,
+        current_position);
 }
 
 auto mir::graphics::atomic::Cursor::needs_compositing() const -> bool
 {
-    return false;
+    return true;
```
RAOF: Wait. Doesn't this mean that the cursor *will* always be added by the compositor, and so will be double-composited?

RAOF (changes requested):
I think this will double-composite the cursor, which is not what we want?

src/platforms/atomic-kms/server/kms/cursor.cpp
mattkae: Whoopsie! That was some debug code ;)

PR #4012, opened 2025-06-12
MIRENG-633-live-ini-file

AlanGriffiths:
Implement IniFile configuration store

### DIFF
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -599,6 +599,8 @@ global:
     miral::live_config::Key::to_string*;
     miral::live_config::Key::operator*;
     miral::live_config::Store::?Store*;
+    miral::live_config::Store::Store*;
+    miral::live_config::Store::operator*;
     non-virtual?thunk?to?miral::live_config::IniFile::?IniFile*;
     non-virtual?thunk?to?miral::live_config::IniFile::add_bool_attribute*;
     non-virtual?thunk?to?miral::live_config::IniFile::add_float_attribute*;


src/miral/live_config_ini_file.cpp
```diff
@@ -0,0 +1,382 @@
...
+            auto const eq = line.find_first_of("=");
+            auto const key = Key{line.substr(0, eq)};
+            auto const value = line.substr(eq+1);
+
+            if (auto const details = attribute_handlers.find(key); details != attribute_handlers.end())
+            {
+                details->second.value = value;
+            }
+
+            if (auto const details = array_attribute_handlers.find(key); details != array_attribute_handlers.end())
```
mattkae: Should this be an `else if` or can you add an attribute that is both an attribute and an array attribute? I guess nothing prevents you from doing so, huh?

src/miral/live_config_ini_file.cpp
```diff
@@ -0,0 +1,382 @@
...
+                }
+                else
+                {
+                    details->second.value = std::vector<std::string>{value};
+                }
+            }
+        }
+        catch (const std::exception& e)
+        {
+            mir::log_warning("Error processing '%s': %s", path.c_str(), e.what());
```
mattkae, line 136: Perhaps for the future: it would be nice to get the errors out of the ini file loader with line number and whatnot. That way the caller could potentially display the error user

src/miral/live_config_ini_file.cpp
AlanGriffiths: Now something does prevent you doing so.

src/miral/live_config_ini_file.cpp
AlanGriffiths: Well, the error does report the file, key and problem:

```
[2025-06-13 10:12:06.167901] < -warning- > miral: Error processing 'a_bad_bool_value_is_ignored': Config key 'a_scope_an_int' has invalid value: 42x
```

examples/mir_demo_server/server_example.cpp
```diff
@@ -118,283 +119,25 @@ catch (std::exception const& error)
...
-    demo_configuration.handle_initial_config();
+    miral::CursorScale cursor_scale{config_store};
+    miral::OutputFilter output_filter{config_store};
+    miral::InputConfiguration input_configuration{config_store};
+
+    miral::ConfigFile config_file{
+        runner,
+        "mir_demo_server.live-config",
+        miral::ConfigFile::Mode::reload_on_change,
+        [&config_store](auto&... args){ config_store.load_file(args...); }};
```
AlanGriffiths, line 140: Is this too big an incantation for user code? Would a helper function make sense? Or another `ConfigFile` constructor?

src/miral/live_config_ini_file.cpp
mattkae: It could be nice to be able to pipe this into a notifier or something though (e.g. `notify-send` or otherwise)

examples/mir_demo_server/server_example.cpp
mattkae: What would the other constructor look like? This is pretty easy to follow in my opinion

include/miral/miral/live_config_ini_file.h
```diff
@@ -0,0 +1,61 @@
...
+namespace miral::live_config
+{
+/// "ini file" implementation of a live configuration store.
+///
+/// \remark Since MirAL 5.4
+class IniFile : public Store
+{
+public:
+    IniFile();
+    ~IniFile();
```
mattkae:
```suggestion
    ~IniFile() override;
```

Right?

mattkae (approved):
Still happy, still some minor suggestions

PR #4016, opened 2025-06-16
Implement locate pointer + graphical feedback

tarek-y-ismail:
# What's new
- Introduces `AnimationDriver` which through you can register to receive vsync events, along with the delta time since the last vsyc.
- Reimplements `LocatePointer`, and adds graphical feedback on top

# TODO
Should've been a spec, but that's alright
- [ ] Agree on how much graphical feedback customization we want to provide. Do we want users to be able to customize the color of the circle? The duration of the animation? The drawing procedure? 
- [ ] Implement `Animation` and `TweenedAnimation` from the spec. They are not strictly needed for graphical feedback to function, but will be nice long-term.
- [ ] As we've circumvented window management, clicks (after the splash screen) throw a `std::out_of_range` exception. We need to agree on how to fix this.


src/miral/locate_pointer.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+            auto const* input_event = event.to_input();
+            if (input_event->input_type() != mir_input_event_type_pointer)
+                return false;
+
+            auto const* pointer_event = input_event->to_pointer();
+            
+            if (pointer_event->action() != mir_pointer_action_motion)
+                return false;
+            if (auto position = pointer_event->position())
```
tarek-y-ismail, line 93: @mattkae just reported that `pointer_event->position()` doesn't get set on `gbm-kms` like it does on `mir-x11`. Will have to track position another way.

tarek-y-ismail:
Since #4342 got merged recently. Will close this for now, at least until the buffer rework is done.

PR #4018, opened 2025-06-16
MIRENG-1243-Improve-support-for-bespoke-Wayland-extensions

AlanGriffiths:
Add WaylandTools for building bespoke Wayland extensions


AlanGriffiths:
This iteration is enough to implement ext-workspace in Miriway. It also provides a template for further iterations

include/miral/miral/wayland_tools.h
```diff
@@ -0,0 +1,55 @@
...
+{
+public:
+    WaylandTools();
+    ~WaylandTools();
+    WaylandTools(WaylandTools const&);
+    auto operator=(WaylandTools const&) -> WaylandTools&;
+
+    void operator()(mir::Server& server) const;
+
+    void for_each_binding(mir::wayland::Client* client, Output const& output, std::function<void(wl_resource* output)> const& callback) const;
```
mattkae, line 47: Nit: I would still call this `for_each_output_binding`

src/miral/wayland_tools.cpp
```diff
@@ -0,0 +1,72 @@
...
+        this->server = server;
+    }
+
+    void for_each_output_binding(
+        mir::wayland::Client* client,
+        mir::graphics::DisplayConfigurationOutputId output,
+        std::function<void(struct wl_resource* output)> const& callback)
+    {
+        std::lock_guard lock(mutex);
+        if (server)
```
mattkae, line 39: Should we warn a user if they try to access this method before the server has started? That seems like it would only ever be a logic error

mattkae (approved):
Two questions, both non-blocking

src/miral/wayland_tools.cpp
AlanGriffiths: It is a (implicit) precondition that the `client` and `output` arguments are valid. I don't see how a shell could have either _before_ the server has started.

After the stop callback is called and the server is shutting down I think it safer to ignore any use of this API.

include/miral/miral/wayland_tools.h
AlanGriffiths: I thought that a bit tautological. Here's an example use:

```c++
    wltools->for_each_binding(client, output,
        [this](wl_resource* output){ the_workspace_group->send_output_enter_event(output); });
```

...compare with...

```c++
    wltools->for_each_output_binding(client, output,
        [this](wl_resource* output){ the_workspace_group->send_output_enter_event(output); });
```

(But I don't feel strongly - anyone for painting the bikeshed?)

mattkae: I don't feel strongly either, so I will let it be left as-is :)

PR #4019, opened 2025-06-16
Use MirMirrorMode on renderables in the renderer

mattkae:
## What's new?
- To properly support [all of the allowed transforms](https://wayland.app/protocols/wayland#wl_output:enum:transform), we need to support `mirror_mode()` on renderables in addition to orientation
- We apply this mirror mode to the renderable at render time to un-flip the content
- Wrote tests for everything

### DIFF
--- a/include/core/mir_toolkit/common.h
+++ b/include/core/mir_toolkit/common.h
@@ -184,7 +184,8 @@ typedef enum MirPixelFormat
                                 (f) == mir_pixel_format_rgba_4444 ? 2 : \
                                                                     4)
 
-/** Direction relative to the "natural" orientation of the display */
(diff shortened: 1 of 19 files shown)


robert-ancell (changes requested):
Changes look generally good. I think you need a test that has both a rotate and a mirror, as the order of this will produce different results. Based on your inverse transform Wayland does the rotate then the flip I assume?

mattkae:
> Changes look generally good. I think you need a test that has both a rotate and a mirror, as the order of this will produce different results. Based on your inverse transform Wayland does the rotate then the flip I assume?

The client flips and then rotates. We unrotate and then flip (again to unflip).

But I can add a test!

PR #4021, opened 2025-06-17
RFC: demonstrating how internal surfaces can be true windows

mattkae:
## What's new?
- Exposed `Shell::add_internal_surface`
- Demonstrated how this could be used for the magnifier surface


robert-ancell (commented):
This only seems useful for surfaces that do not perform user interaction, as all the surfaces share a common session with a NullEventSink. Does that mean for other surfaces we would not be able to use `add_internal_session` and have to make out own session and use that?

mattkae:
> This only seems useful for surfaces that do not perform user interaction, as all the surfaces share a common session with a NullEventSink. Does that mean for other surfaces we would not be able to use `add_internal_session` and have to make out own session and use that?

To be honest, I am a bit unsure as to what the event sink is for there. If you look in `wl_client.cpp`, they use `NullEventSink` too. So do xwayland clients. I think that the event sink is just an artifact there, and probably serves no purpose.

AlanGriffiths:
I feel I'm lacking context.

What exactly is an "internal surface"? I've asked Matt directly a couple of times and got directed here as an example.

Matt has mentioned ambitions to use implement decorations and docks using "internal surface"s. In the case of a "dock" we're clearly talking about something which is only different to a "dock" provided by a client in terms of the source of buffers and the consumption of input.

This example of "magnifier surface" is complicated by the necessities of rendering into the surface. But from my understanding:

1. a "magnifier surface" content is rendered in the server
2. a "magnifier surface" is rendered to the display buffer
3. a "magnifier surface" is not managed by window management
4. a "magnifier surface" should be transparent to input

The "not visible to window management" should probably imply "transparent to input", so the question really is how to manage the different types of "internal surface". Are they best modeled as different states of a generalised `BasicSurface`, or are they distinct types with individual implementations?

AlanGriffiths:
> To be honest, I am a bit unsure as to what the event sink is for there. If you look in `wl_client.cpp`, they use `NullEventSink` too. So do xwayland clients. I think that the event sink is just an artifact there, and probably serves no purpose.

I think that's a legacy of the mirclient API - and was used for an "is alive" message pulse.

PR #4022, opened 2025-06-18
Integrate some Rust code into the Mir build

jhenstridge:
Add scaffolding for integrating Rust code into Mir. This includes:

* A top level Cargo workspace declaring dependencies and pointing at member packages deeper in the tree.
* Some CMake helpers to build Rust staticlib packages with [rust-cxx](https://cxx.rs/index.html). This involves invoking Cargo to build the Rust part, and CMake directives to build the generated C++ stubs and make them easy to link against. 
* Adds a trivial Rust package to `srv/server/input`. At the moment, it provides a function that prints a a log message containing a string provided by the C++ side.

Some nice things about this approach:
* We can have multiple Rust packages in different parts of the tree that can share dependencies (and dependency versions) from the workspace.
* We should be able to produce tarballs that can build offline using `cargo vendor`.

Problems:
* I haven't looked at how to eliminate dead code from the final `libmirserver.so`, so it is huge.


cmake/RustLibrary.cmake
```diff
@@ -0,0 +1,51 @@
...
+  set(one_value_args CRATE CXX_BRIDGE_SOURCE_FILE)
+  set(multi_value_args DEPENDS)
+  cmake_parse_arguments(arg "" "${one_value_args}" "${multi_value_args}" ${ARGN})
+
+  if("${arg_CRATE}" STREQUAL "")
+    message(FATAL_ERROR "add_rust_cxx_library called without CRATE <value>")
+  endif()
+
+  if("${arg_CXX_BRIDGE_SOURCE_FILE}" STREQUAL "")
+    set(arg_CXX_BRIDGE_SORUCE_FILE "src/lib.rs")
```
Saviq:
```suggestion
    set(arg_CXX_BRIDGE_SOURCE_FILE "src/lib.rs")
```

cmake/RustLibrary.cmake
```diff
@@ -0,0 +1,57 @@
...
+  if("${arg_CRATE}" STREQUAL "")
+    message(FATAL_ERROR "add_rust_cxx_library called without CRATE <value>")
+  endif()
+
+  if("${arg_CXX_BRIDGE_SOURCE_FILE}" STREQUAL "")
+    set(arg_CXX_BRIDGE_SOURCE_FILE "src/lib.rs")
+  endif()
+
+  set(rust_target_dir "${CMAKE_BINARY_DIR}/target")
+  if("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
```
RAOF, line 17: We'll also need to handle Debian package builds, which to my knowledge do *not* set `CMAKE_BUILD_TYPE=Release`, and instead set `CFLAGS` etc directly.

Alternatively, maybe we should just require the package build to use the `Release` type!

cmake/RustLibrary.cmake
```diff
@@ -0,0 +1,57 @@
...
+    IMPORTED_LOCATION ${crate_staticlib})
+
+  add_library(${target}-cxxbridge
+    ${cxxbridge_source} ${cxxbridge_header})
+  # rust-cxx generates symbols named like "cxxbridge1$foo", which
+  # triggers a warning in Clang.
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    target_compile_options(${target}-cxxbridge
+      PRIVATE -Wno-error=dollar-in-identifier-extension)
+  endif()
```
RAOF: I think we have a slight preference for
```suggestion
  check_cxx_compiler_flag(-Wdollar-in-identifier-extension SUPPORTS_DOLLAR_IN_ID_WARNING)
  if (SUPPORTS_DOLLAR_IN_ID_WARNING)  
    target_compile_options(${target}-cxxbridge
      PRIVATE -Wno-error=dollar-in-identifier-extension)
  endif()
```

Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[workspace]
+members = [
+  "src/server/input",
+]
+resolver = "2"
+
+[workspace.package]
+name = "mir"
+version = "0.1.0"
+edition = "2021"
```
RAOF, line 10: It's a pity we can't use `edition = "2024"`; the archive rustc is too old (even questing!)

jhenstridge:
I've adjusted this branch to what I feel should make it more suitable to merge:

1. I moved the proof of concept Rust package from `src/server/input` to `src/platforms/evdev`, since that's where we will be experimenting with Rust.
2. The code is now built conditionally on a `-DMIR_ENABLE_RUST=ON` cmake option. The changes to the packaging and docs builds have been backed out, since they don't enable this.
3. There is a new spread task and Github actions job to build on Ubuntu 24.04 with this option turned on to ensure it still builds and passes the tests.

cmake/RustLibrary.cmake
jhenstridge: I'm unsure of what the best option here is. While we treat this as an experimental feature, it probably doesn't matter that we're always building with debug info.

Looking at librsvg for inspiration again, here is what they have done:

https://gitlab.gnome.org/GNOME/librsvg/-/blob/main/meson.build?ref_type=heads#L475-485

So they are defaulting to a release build with no options. Perhaps that is the best option here, and only do non-release builds on particular recognised build types (maybe Debug, Coverage, and *Sanitizer?). Not sure if there is a good story for preparing a release build with debug info though.

RAOF: This should be fine for now.

Issue #4023, opened 2025-06-18
Docking kgx left/right in miriway is misplaced


AlanGriffiths:
Miriway uses Mir's "attach" logic to position the window, so filing here.

Running Miriway/beta on 24.04 (so may affect #3930)

There's no problem if the kgx window starts from fullscreen - so I suspect it has to do with the buffer offset (used for drop shadows).

Image shows left and right docked windows.
 
![Image](https://github.com/user-attachments/assets/7fcbb10b-fa7d-46cb-8d4f-fdd5f01673e6)

AlanGriffiths:
Cannot reproduce on Miriway/stable

AlanGriffiths:
OK, tracked it down to f3dc0f7a77a23ac188268fd516806f4331b4e932

PR #4025, opened 2025-06-18
Mostly ignore `WlSurface::offset()`

AlanGriffiths:
Fixes: #4023

### DIFF
--- a/src/server/frontend_wayland/wl_surface.cpp
+++ b/src/server/frontend_wayland/wl_surface.cpp
@@ -670,7 +670,10 @@ void mf::WlSurface::set_buffer_scale(int32_t scale)
 
 void mir::frontend::WlSurface::offset(int32_t x, int32_t y)
 {
-    pending.offset = geom::Displacement{x, y};
+    if (x != 0 || y != 0)
+    {
+        mir::log_warning("Client requested unimplemented non-zero offset. Rendering will be incorrect.");
+    }
 }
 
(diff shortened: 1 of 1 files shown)


src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -670,7 +670,10 @@ void mf::WlSurface::set_buffer_scale(int32_t scale)
 
 void mir::frontend::WlSurface::offset(int32_t x, int32_t y)
 {
-    pending.offset = geom::Displacement{x, y};
```
Saviq, line 673: This was there for 4 years… why is it only a problem now?

What do we do to ensure this stays unbroke?

AlanGriffiths: It has not been four years since May 6th

Saviq: :man_facepalming:, found the wrong offset…

https://github.com/canonical/mir/blame/b0ec855d9e3bf405d6b36423f59e9243dd9537d2/src/server/frontend_wayland/wl_surface.cpp#L200

Saviq: Still, anything we can easily do to ensure this remains unbroken?

AlanGriffiths: This fixes the breakage. I'm looking into how we _should_ be handling `attach()/offset()`. But fixing that and adding tests is less urgent than having a cherry-pick for the release

Issue #3996, opened 2025-06-09
Real `RenderingPlatform` in tests causes wlcs failures on `riscv64`


Saviq:
- #3924

Ref.:

https://launchpad.net/~mir-team/+archive/ubuntu/dev/+packages?field.name_filter=mir&field.status_filter=&field.series_filter=noble

```
The following tests FAILED:
	 18 - wlcs.LayerSurfaceTest.* (Failed)
	 20 - wlcs.ForeignToplevelManagerTest.* (Failed)
	 21 - wlcs.ForeignToplevelHandleTest.* (Failed)
	 23 - wlcs.TextInputV3WithInputMethodV2Test.* (Failed)
	 24 - wlcs.VirtualPointerV1Test.* (Failed)
	 25 - wlcs.TextInputV2WithInputMethodV1Test.* (Failed)
	 27 - wlcs.WpViewporterTest.* (Failed)
	 29 - wlcs.PointerCrossingSurfaceCorner/SurfacePointerMotionTest.* (Failed)
	 30 - wlcs.PointerCrossingSurfaceEdge/SurfacePointerMotionTest.* (Failed)
	 31 - wlcs.AllSurfaceTypes/TouchTest.* (Failed)
	 32 - wlcs.MultiRectEdges/RegionSurfaceInputCombinations.* (Timeout)
	 33 - wlcs.DefaultEdges/RegionSurfaceInputCombinations.* (Timeout)
	 34 - wlcs.FullSurface/RegionSurfaceInputCombinations.* (Failed)
	 35 - wlcs.SmallerRegion/RegionSurfaceInputCombinations.* (Failed)
	 36 - wlcs.ClippedLargerRegion/RegionSurfaceInputCombinations.* (Failed)
	 37 - wlcs.MultiRectCorners/RegionSurfaceInputCombinations.* (Failed)
	 38 - wlcs.SurfaceInputRegions/SurfaceInputCombinations.* (Timeout)
	 60 - wlcs.Anchor/LayerSurfaceLayoutTest.* (Timeout)
```

Saviq:
@RAOF will you be able to look into this? CI is running in loops trying to get a consistent set of snaps built...

RAOF:
Ah, sure, I'll switch to this.

Saviq:
@RAOF this doesn't seem to have registered?

```
[...]

PR #4027, opened 2025-06-19
Bump WLCS test timeouts on RISCV

RAOF:
Each individual test takes ~12sec on RISCV, and some of the suites have >40
individual tests in them, so we hit the 600sec default per-suite CTest timeout.

Bump that to 3600sec on RISCV.


RAOF:
(Other half of) Fixes: https://github.com/canonical/mir/issues/3996

AlanGriffiths (approved):
Worth trying

PR #4031, opened 2025-06-20
Implement WlSurface::set_buffer_transform and pipe the result to the renderable

mattkae:


RAOF:
This can be tested with `weston-transformed` from the `weston` package; `left`/`right` rotate, `space` reflects, and `z` switches buffer scale. It should look the same in all orientations.

src/server/frontend_wayland/output_manager.cpp
```diff
@@ -325,6 +325,47 @@ void mf::OutputManager::remove_listener(OutputManagerListener* listener)
     std::erase_if(listeners, [&](auto candidate) { return candidate == listener; });
 }
 
+auto mf::OutputManager::from_output_transform(int32_t transform)
+    -> std::optional<std::tuple<MirOrientation, MirMirrorMode>>
```
RAOF: I *think* I'd prefer this to `throw` a `std::out_of_range` exception, with the call-site catching it and re-raising the correct `ProtocolError` exception?

src/server/frontend_wayland/output_manager.h
```diff
@@ -119,6 +120,9 @@ class OutputManager
     void add_listener(OutputManagerListener* listener);
     void remove_listener(OutputManagerListener* listener);
 
+    static auto from_output_transform(int32_t transform)
+        -> std::optional<std::tuple<MirOrientation, MirMirrorMode>>;
```
RAOF: Not entirely clear to me that this should be a static method on `OutputManager`? That said, I don't have anywhere better for it :woman_shrugging:

RAOF (commented):
A couple of not-entirely-clear nits; we might need a tiebreaker for these.

Otherwise, you need to fix the miroil symbols, it seems.

src/server/frontend_wayland/output_manager.h
mattkae: Same. I put it there because the transforms are on the outputs, but it is a long-ish shot

robert-ancell (approved):
LGTM

RAOF:
This appears to not work correctly? Here's the result of running `weston-transformed` on miral-shell:
Initial image
And after hitting `<right>`:
After hitting <right>

In contrast, on Weston (and GNOME Shell), you get the following rendering:
Initial image
And after hitting `<right>`:
After hitting right

mattkae:
> I haven't managed to work out _where_ it's broken, but it is.

Yup! Fixed :) It was broken in many ways!

RAOF (changes requested):
Still not quite fixed! It now alternates between two orientations:

(even orientation)
transform2
(odd orientation)
transform3

PR #4033, opened 2025-06-23
server: add `--vt-switching` option

Saviq:


RAOF (approved):
Yup, LGTM, apart from the test failure which is weird?

Saviq:
> the test failure which is weird?

Yeah shouldn't have taken that shortcut. Reworked with a mock now.

RAOF (approved):
Yeah, now that works :)

PR #4037, opened 2025-06-24
Convert cmake $<TARGET_OBJECTS:...> usage to regular target_link_libraries() dependencies

jhenstridge:
[Since CMake 3.12](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries), it has been possible to link to an object library with a regular `target_link_libraries` directive. This is preferable to the use of the `$<TARGET_OBJECTS:...>` generator, since it will also bring in dependencies of the object library instead of just the object files.

This makes it possible to specify dependencies with the object library, rather than replicating them everywhere the object library is linked. This is particularly useful for optional dependencies, since the build logic can be localised to where the optional dependency is used.

Since Mir's current minimum CMake version is 3.16, we should be able to rely on this feature. This PR converts all the `$<TARGET_OBJECTS:...>` use to equivalent `target_link_libraries` calls. There is probably more room for simplification, but this is a start.

### DIFF
--- a/src/platform/CMakeLists.txt
+++ b/src/platform/CMakeLists.txt
@@ -39,8 +39,12 @@ target_include_directories(mirplatform
 target_link_libraries(mirplatform
   PRIVATE
     mircommon
-  PUBLIC
     ${MIR_PLATFORM_REFERENCES}
+  PUBLIC
+    mirwayland
+    mirplatformgraphicscommon
+    miroptions
+    mirudev
     PkgConfig::EPOXY
 )
 
--- a/src/platforms/common/server/CMakeLists.txt
+++ b/src/platforms/common/server/CMakeLists.txt
@@ -35,5 +35,4 @@ target_link_libraries(server_platform_common
     ${KMS_UTILS_STATIC_LIBRARY}
     Boost::system
     PkgConfig::WAYLAND_SERVER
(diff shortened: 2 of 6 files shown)


jhenstridge:
I think I've resolved the test problems I had with this branch now.

The underlying problem was that if you had a target with dependencies "A B C D", and A depended on X and Y, then the link line would be end up as "A B C D X Y". This was a problem if D is the mock OpenGL static lib and Y is the real OpenGL libraries.

As X comes after the mock GL library, it won't bring in any new objects from the static lib. And we can't just make X depend on the mock GL implementation, because it is also used to build the installed binaries. We can work around that by specifying the target's dependencies as "A X B C D" (or some variant) to ensure the mock GL symbols are picked first.

This is potentially fragile, but the `unit-tests-do-not-link-with-GL-or-EGL` should pick up regressions (as it did with previous versions of this PR).

jhenstridge:
> I agree the linking of the mock GL library is fragile, but that's the pre-existing reason we have the tests

I filed #4046 as an alternative approach we could use in the future: the "real OpenGL" we link to is a dispatch library, so we could have it dispatch to our mock implementation in the tests. This would allow for a stronger dependency on the GL/EGL libraries where it makes sense in the build system.

PR #4038, opened 2025-06-24
Add missing debian symbols

tarek-y-ismail:


debian/libmiral7.symbols
```diff
@@ -510,6 +510,7 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::CursorScale::scale(float) const@MIRAL_5.3" 5.3.0
  (c++)"miral::CursorScale::~CursorScale()@MIRAL_5.3" 5.3.0
  (c++)"miral::DisplayConfiguration::Node::Node(std::unique_ptr<miral::DisplayConfiguration::Node::Self, std::default_delete<miral::DisplayConfiguration::Node::Self> >&&)@MIRAL_5.3" 5.3.0
+ (c++)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.3" 5.3.0
```
tarek-y-ismail, line 513: Is this the right place for this symbol or should it be added to the 5.4.0 section?

AlanGriffiths (changes requested):
I don't think any of these symbols need exist: they are all inline functions

tarek-y-ismail:
> I don't think any of these symbols need exist: they are all inline functions

As long as they don't cause issues, that's okay with me!

AlanGriffiths:
> As long as they don't cause issues, that's okay with me!

I wonder why you see them being generated. (I don't)

PR #4040, opened 2025-06-24
Fixing issue where alarm could be scheduled with a null click_dispatcher

mattkae:
Whoops! `state->click_dispatcher` is initialized lazily. We should make it more obvious that this is the case :)

### DIFF
--- a/src/include/server/mir/input/input_event_transformer.h
+++ b/src/include/server/mir/input/input_event_transformer.h
@@ -47,7 +47,7 @@ public:
         // shouldn't be handled by later transformers, whether the transformer is
         // accumulating events for later dispatching or has immediately dispatched
         // an event is an implementation detail of the transformer
-        virtual bool transform_input_event(EventDispatcher const&, EventBuilder*,  MirEvent const&) = 0;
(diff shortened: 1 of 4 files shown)


src/include/server/mir/input/input_event_transformer.h
tarek-y-ismail, line 1: Unrelated?

src/server/input/input_event_transformer.cpp
tarek-y-ismail, line 1: Unrelated?

src/server/shell/basic_hover_click_transformer.h
```diff
@@ -62,9 +62,10 @@ class BasicHoverClickTransformer : public HoverClickTransformer
     static constexpr auto grace_period_percentage = 0.1f;
     static constexpr auto hover_delay_percentage = 1.0f - grace_period_percentage;
 
+    class CursorObserver;
```
tarek-y-ismail, line 65: Is this needed?

mattkae: Just making it a touch easier to read

PR #4041, opened 2025-06-24
Add AppendKeyboardEventFilter helper

robert-ancell:
A lot of code adds event filters for keyboard events only; add a helper class to remove this boilerplate code.

### DIFF
--- a/debian/libmiral7.symbols
+++ b/debian/libmiral7.symbols
@@ -5,6 +5,8 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::AddInitCallback::~AddInitCallback()@MIRAL_5.0" 5.0.0
  (c++)"miral::AppendEventFilter::AppendEventFilter(std::function<bool (MirEvent const*)> const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::AppendEventFilter::operator()(mir::Server&)@MIRAL_5.0" 5.0.0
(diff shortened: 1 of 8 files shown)


robert-ancell:
Documentation build failure seems to be due to existing words - added in https://github.com/canonical/mir/pull/4042

tarek-y-ismail (approved):
This has been a long time coming. Thank you!

tarek-y-ismail:
Maybe this could be a utility function? That way it could be used in event filters and anything else that handles input events. 

Something like:
```cpp
auto maybe_keyboard_event(MirEvent const*) -> std::optional<MirKeyboardEvent*>;
```

robert-ancell:
> Maybe this could be a utility function? That way it could be used in event filters and anything else that handles input events.
> 
> Something like:
> 
> ```c++
> auto maybe_keyboard_event(MirEvent const*) -> std::optional<MirKeyboardEvent*>;
> ```

I picked the new class as this seems to be the most common case for filtering events. Using the `std::optional` means all the code that uses this has to check if the event is provided. Using the class means the event is always valid, and no checking code is required.

tarek-y-ismail:
> I picked the new class as this seems to be the most common case for filtering events. Using the std::optional means all the code that uses this has to check if the event is provided. Using the class means the event is always valid, and no checking code is required.

I still think a more general solution would be nice (for us and users). But your code should help until then.

tarek-y-ismail (approved):
Caught one final thing, hopefully no other surprises. Let's see what CI has to say.

PR #4051, opened 2025-06-27
Fix Alpine CI failure

robert-ancell:
error: ignoring returned value of type 'testing::AssertionResult', declared with attribute 'nodiscard' [-Werror=unused-result]

### DIFF
--- a/tests/unit-tests/input/test_default_input_device_hub.cpp
+++ b/tests/unit-tests/input/test_default_input_device_hub.cpp
@@ -100,7 +100,7 @@ struct InputDeviceHubTest : ::testing::Test
 
     void expect_and_execute_multiplexer()
     {
-        mt::fd_becomes_readable(multiplexer.watch_fd(), 2s);
+        EXPECT_TRUE(mt::fd_becomes_readable(multiplexer.watch_fd(), 2s));
         multiplexer.dispatch(mir::dispatch::FdEvent::readable);
     }
 };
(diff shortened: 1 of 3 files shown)


RAOF (approved):
image

Sure.

tests/unit-tests/input/test_default_input_device_hub.cpp
```diff
@@ -100,7 +100,7 @@ struct InputDeviceHubTest : ::testing::Test
 
     void expect_and_execute_multiplexer()
     {
-        mt::fd_becomes_readable(multiplexer.watch_fd(), 2s);
+        (void)mt::fd_becomes_readable(multiplexer.watch_fd(), 2s);
```
AlanGriffiths, line 103: I've not looked at the broader context but, given the function name, shouldn't this be more like:

```suggestion
        EXPECT_TRUE(mt::fd_becomes_readable(multiplexer.watch_fd(), 2s));
```

robert-ancell: I thought so too, but when I did that originally the tests failed. So it seems this line is not essential, which I guess is why it didn't have an expectation?

AlanGriffiths: I guess the underlying problem is that the `fd_becomes_readable()` return type is wrong: it is _never_ used as a test assertion. (With the possible exception of the indirect use via `fd_is_readable()`.) 

But that, and figuring out the "correct" semantics of `expect_and_execute_multiplexer()` is a can I don't want to open either

PR #4053, opened 2025-06-30
Initialize structs.

robert-ancell:
As TICS likes them.

### DIFF
--- a/examples/client/spinner.cpp
+++ b/examples/client/spinner.cpp
@@ -26,7 +26,7 @@ static void shutdown(int)
 
 int main()
 {
-    struct sigaction action = {};
+    struct sigaction action{};
     action.sa_handler = shutdown;
     sigemptyset(&action.sa_mask);
     action.sa_flags = 0;
--- a/examples/client/spinner.cpp
+++ b/examples/client/spinner.cpp
@@ -26,7 +26,7 @@ static void shutdown(int)
 
 int main()
 {
-    struct sigaction action;
+    struct sigaction action = {};
     action.sa_handler = shutdown;
     sigemptyset(&action.sa_mask);
     action.sa_flags = 0;
--- a/src/common/logging/logger.cpp
+++ b/src/common/logging/logger.cpp
@@ -81,7 +81,7 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
         "< - debug - > "
     };
 
-    struct timespec ts = {};
+    timespec ts{};
     clock_gettime(CLOCK_REALTIME, &ts);
     char now[32];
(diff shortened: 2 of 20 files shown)


examples/client/spinner.cpp
```diff
@@ -26,7 +26,7 @@ static void shutdown(int)
 
 int main()
 {
-    struct sigaction action;
+    struct sigaction action = {};
```
RAOF: I think I would mildly prefer
```suggestion
    struct sigaction action{};
```
but not blocking.

RAOF (approved):
You're getting a lot of this, it seems :)
image

I *think* all of these are false-positives, but there's no language mechanism TICS could use to determine that these are false-positives, so sure.

robert-ancell:
Yes, I'm sure none of these are bugs (unless the libraries we are calling into have bugs)

tests/unit-tests/test_udev_wrapper.cpp
```diff
@@ -402,7 +402,7 @@ TEST_F(UdevWrapperTest, UdevMonitorFdIsReadableWhenEventsAvailable)
 
     udev_environment.add_device("drm", "card0", NULL, {}, {});
 
-    struct pollfd fds;
+    struct pollfd fds = {};
     fds.fd = monitor.fd();
     fds.events = POLLIN;
```
AlanGriffiths:
```suggestion
    pollfd fds{ .fd = monitor.fd(), .events = POLLIN, .revents = {} };
```

examples/client/spinner.cpp
AlanGriffiths: I concur: we should use the universal initialization syntax

src/common/logging/logger.cpp
```diff
@@ -81,7 +81,7 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
         "< - debug - > "
     };
 
-    struct timespec ts;
+    struct timespec ts = {};
```
AlanGriffiths:
```suggestion
    timespec ts{};
```

src/server/console/linux_virtual_terminal.cpp
```diff
@@ -317,7 +317,7 @@ mir::LinuxVirtualTerminal::LinuxVirtualTerminal(
       prev_tcattr(),
       active{true}
 {
-    struct termios tcattr;
+    struct termios tcattr = {};
```
AlanGriffiths: Better as an initialization on first use:

```c++
L353    auto tcattr = prev_tcattr;
```

AlanGriffiths:
> I _think_ all of these are false-positives

I concur. The root problem is the C APIs that use a function call for initialization with no way to mark "out" parameters.

Maybe we should do better at wrapping these APIs for C++.

PR #4054, opened 2025-07-01
Add live configuration facilities to bounce keys

tarek-y-ismail:


examples/miral-shell/shell_main.cpp
tarek-y-ismail, line 1: Wrong PR!

tarek-y-ismail:
> Missing changes to symbols.map and libmiral.symbols?

Only the Debian symbols. `symbols.map` is already covered by `BounceKeys*` from the parent PR.

PR #4056, opened 2025-07-02
Drop some dead code

AlanGriffiths:
This PR drops some unused infrastructure. Since the migration to Wayland this is no longer used.

The mirclient API was able to send clients various events: keep-alive pulses, input configuration, lifecycle, error and prompt session events.


AlanGriffiths:
There's more threads left to this: in particular, prompt sessions can go, and with them the need for `ApplicationSession::event_sink` and the methods that use it

src/server/input/CMakeLists.txt
```diff
@@ -4,9 +4,7 @@ set(
 
   basic_seat.cpp
   builtin_cursor_images.cpp
-  config_changer.cpp
-  config_changer.h
-  cursor_controller.cpp
+        cursor_controller.cpp
```
tarek-y-ismail: Not sure if this is GitHub incorrectly rendering things, or if it's a real misalignment

src/server/shell/abstract_shell.cpp
```diff
@@ -34,6 +34,7 @@
 #include <algorithm>
 #include <iterator>
 #include <vector>
+#include <mir/frontend/event_sink.h>
```
tarek-y-ismail:
```suggestion
```

src/server/symbols.map
tarek-y-ismail, line 1: Since we're removing symbols, should we do the whole bump the API and soname thing?

tarek-y-ismail (commented):
Quick skim, things seem alright though

src/server/symbols.map
AlanGriffiths: We already did

tarek-y-ismail (approved):

PR #4058, opened 2025-07-02
Accounting for surface orientation in BasicSurface::input_area_contains

mattkae:
## What's new?
- Updated `BasicSurface::input_area_contains` to account for the orientation of the surface when intersecting its custom input rectangles so that oriented surfaces work
- Wrote a test

## How to test?
- Run `weston-transformed`
- When the surface is oriented either 90 or 270 degrees, note that your cursor treats the input area such that the width is always 500 and the height is always 250
- Your cursor will disappear when intersecting the surface (by design of weston-transformed!)

Note that this does NOT contain the visual side of things (that is #4031)

### DIFF
--- a/src/server/scene/basic_surface.cpp
+++ b/src/server/scene/basic_surface.cpp
@@ -360,16 +360,31 @@ bool ms::BasicSurface::input_area_contains(geom::Point const& point) const
 
     if (state->custom_input_rectangles.empty())
     {
-        // no custom input, restrict to bounding rectangle
+        // The surface has no custom input, so it is restricted to its bounding rectangle.
+        // Note that the "content_size" returned here is already oriented properly, so the
+        // width and height do NOT need to be swapped, ujnlike the custom input rectangles
+        // case.
         auto const input_rect = geom::Rectangle{content_top_left(*state), content_size(*state)};
         return input_rect.contains(point);
     }
     else
     {
-        auto local_point = as_point(point - content_top_left(*state));
+        auto const local_point = as_point(point - content_top_left(*state));
         for (auto const& rectangle : state->custom_input_rectangles)
         {
-            if (rectangle.contains(local_point))
+            // When a surface is rendered with an orientation, it is unrotated so that it appears
+            // upright. The input region that the surface provides is also rotated according to the
+            // same rules. Hence, we need to swap the width and height of the input region if the
+            // surface is rotated 90 or 270 degrees.
+            if (state->orientation == mir_orientation_right || state->orientation == mir_orientation_left)
+            {
+                auto rotated_rect = rectangle;
+                rotated_rect.size.width = geom::Width(rectangle.size.height.as_value());
+                rotated_rect.size.height = geom::Height(rectangle.size.width.as_value());
+                if (rotated_rect.contains(local_point))
+                    return true;
+            }
+            else if (rectangle.contains(local_point))
                 return true;
         }
     }
--- a/tests/unit-tests/scene/test_basic_surface.cpp
+++ b/tests/unit-tests/scene/test_basic_surface.cpp
@@ -575,6 +575,28 @@ TEST_F(BasicSurfaceTest, set_input_region)
     }
 }
 
+class BasicSurfaceOrientationTest : public BasicSurfaceTest, public testing::WithParamInterface<MirOrientation>
+{};
+
+TEST_P(BasicSurfaceOrientationTest, set_input_region_on_rotated_surface)
+{
+    std::vector<geom::Rectangle> const rectangles = {
+        {{geom::X{0}, geom::Y{0}}, {geom::Width{2}, geom::Height{5}}},
+    };
+
+    surface.set_orientation(GetParam());
+    surface.register_interest(observer, executor);
+    surface.set_input_region(rectangles);
+
+    auto const test_pt = rect.top_left + geom::Displacement{4, 1};
+    EXPECT_THAT(surface.input_area_contains(test_pt), testing::Eq(true));
+}
+
+INSTANTIATE_TEST_SUITE_P(BasicSurfaceOrientationTest, BasicSurfaceOrientationTest, ::testing::Values(
+    mir_orientation_right,
+    mir_orientation_left
+));
+
 TEST_F(BasicSurfaceTest, updates_default_input_region_when_surface_is_resized_to_larger_size)
 {
     geom::Rectangle const new_rect{rect.top_left,{20,20}};


src/server/scene/basic_surface.cpp
```diff
@@ -365,16 +365,31 @@ bool ms::BasicSurface::input_area_contains(geom::Point const& point) const
 
     if (state->custom_input_rectangles.empty())
     {
-        // no custom input, restrict to bounding rectangle
+        // The surface has no custom input, so it is restricted to its bounding rectangle.
+        // Note that the "content_size" returned here is already oriented properly, so the
+        // width and height do NOT need to be swapped, ujnlike the custom input rectangles
```
tarek-y-ismail:
```suggestion
        // width and height do NOT need to be swapped, unlike the custom input rectangles
```

src/server/scene/basic_surface.cpp
```diff
@@ -365,16 +365,31 @@ bool ms::BasicSurface::input_area_contains(geom::Point const& point) const
...
+            // upright. The input region that the surface provides is also rotated according to the
+            // same rules. Hence, we need to swap the width and height of the input region if the
+            // surface is rotated 90 or 270 degrees.
+            if (state->orientation == mir_orientation_right || state->orientation == mir_orientation_left)
+            {
+                auto rotated_rect = rectangle;
+                rotated_rect.size.width = geom::Width(rectangle.size.height.as_value());
+                rotated_rect.size.height = geom::Height(rectangle.size.width.as_value());
+                if (rotated_rect.contains(local_point))
+                    return true;
```
tarek-y-ismail, line 408: Maybe
```suggestion
                return rotated_rect.contains(local_point));
```

Would be better?

tests/unit-tests/scene/test_basic_surface.cpp
```diff
@@ -575,6 +575,28 @@ TEST_F(BasicSurfaceTest, set_input_region)
...
+{};
+
+TEST_P(BasicSurfaceOrientationTest, set_input_region_on_rotated_surface)
+{
+    std::vector<geom::Rectangle> const rectangles = {
+        {{geom::X{0}, geom::Y{0}}, {geom::Width{2}, geom::Height{5}}},
+    };
+
+    surface.set_orientation(GetParam());
+    surface.register_interest(observer, executor);
```
tarek-y-ismail: Is the observer of any use here?

src/server/scene/basic_surface.cpp
```diff
@@ -365,16 +365,31 @@ bool ms::BasicSurface::input_area_contains(geom::Point const& point) const
...
+            // When a surface is rendered with an orientation, it is unrotated so that it appears
+            // upright. The input region that the surface provides is also rotated according to the
+            // same rules. Hence, we need to swap the width and height of the input region if the
+            // surface is rotated 90 or 270 degrees.
+            if (state->orientation == mir_orientation_right || state->orientation == mir_orientation_left)
+            {
+                auto rotated_rect = rectangle;
+                rotated_rect.size.width = geom::Width(rectangle.size.height.as_value());
+                rotated_rect.size.height = geom::Height(rectangle.size.width.as_value());
+                if (rotated_rect.contains(local_point))
```
AlanGriffiths: Don't we also need to "rotate" the top-left of the rectangles?

tests/unit-tests/scene/test_basic_surface.cpp
```diff
@@ -575,6 +575,28 @@ TEST_F(BasicSurfaceTest, set_input_region)
     }
 }
 
+class BasicSurfaceOrientationTest : public BasicSurfaceTest, public testing::WithParamInterface<MirOrientation>
+{};
+
+TEST_P(BasicSurfaceOrientationTest, set_input_region_on_rotated_surface)
+{
+    std::vector<geom::Rectangle> const rectangles = {
+        {{geom::X{0}, geom::Y{0}}, {geom::Width{2}, geom::Height{5}}},
```
AlanGriffiths: We should test with top_left != `{0, 0}`

AlanGriffiths (changes requested):

src/server/scene/basic_surface.cpp
AlanGriffiths: Would be worse: `return` exits the loop

src/server/scene/basic_surface.cpp
```diff
@@ -365,16 +365,33 @@ bool ms::BasicSurface::input_area_contains(geom::Point const& point) const
...
+            // upright. The input region that the surface provides is also rotated according to the
+            // same rules. Hence, we need to swap the width and height of the input region if the
+            // surface is rotated 90 or 270 degrees.
+            if (state->orientation == mir_orientation_right || state->orientation == mir_orientation_left)
+            {
+                auto rotated_rect = rectangle;
+                rotated_rect.top_left.x = geom::X(rectangle.top_left.y.as_value());
+                rotated_rect.top_left.y = geom::Y(rectangle.top_left.x.as_value());
+                rotated_rect.size.width = geom::Width(rectangle.size.height.as_value());
+                rotated_rect.size.height = geom::Height(rectangle.size.width.as_value());
```
AlanGriffiths: I think this transformation is wrong: it looks like a reflection along `x==y`, not a rotation.

For example, rotating `{ 1, 2, 3, 4 }` by 90 degrees should give `{ x-6, y+1, 4, 3 }` (where `x, y` is the center of rotation - which you don't currently use)

src/server/scene/basic_surface.cpp
tarek-y-ismail: Oh, you're right, I completely forgot about the loop

src/server/scene/basic_surface.cpp
AlanGriffiths: Actually, that example is wrong too. 6 & 1 are only correct when x & y are 0 (which obviously isn't the case here)

mattkae: I believe that I'm right, but let me write it out.

This rectangle is given in local coordinates such that (0, 0) is always the top-left corner of the surface. Let's say I declare an inpout rectangle with xy=(1, 2) and a 3x4 area. Let's now say that the surface is rotated 90 degrees.

As far as the spec is concerned (and the `weston-transformed` client), we don't _really_ rotate _about_ the topleft point. Instead,  we rotate the input rectangle on its side and align the new top left with the top left of the actual surface. Hence, all we need to do is a simple swap to get it to work.

AlanGriffiths: That is true _if the input area has the same center as the window_.

However, that's not the general case: Consider an input area of 5px along the top edge. After a 90 degree rotation it is on the right edge. After 180 degrees it is on the bottom edge and after 270 degrees it is on the left edge.

src/server/scene/basic_surface.cpp
```diff
@@ -377,22 +377,42 @@ bool ms::BasicSurface::input_area_contains(geom::Point const& point) const
...
+                return rotated_rect.contains(local_point);
+            }
+            case mir_orientation_inverted:
+            {
+                auto rotated_rect = rectangle;
+                rotated_rect.top_left.y = geom::Y(surface_size.height.as_value() - rectangle.top_left.y.as_value() - rectangle.size.height.as_value());
+                return rotated_rect.contains(local_point);
+            }
```
AlanGriffiths: This looks wrong:

Suppose the input rectangle is a LHS border, then inverted is a 180 degree rotation and the transformed rectangle should be a RHS border.

This transform keeps the rectangle (x is unchanged) on the left and moves it below the window (y goes from 0 to `surface_size.height`)

AlanGriffiths (approved):
I've looked at this so many times I don't see problems anymore

src/server/scene/basic_surface.cpp
```diff
@@ -365,17 +365,63 @@ bool ms::BasicSurface::input_area_contains(geom::Point const& point) const
...
+            // When a surface is rendered with an orientation, it is unrotated so that it appears
+            // upright. The input region that the surface provides is given in surface-local coordinates
+            // meaning that it will need to be rotated to match up with the surface.
+            switch (state->orientation)
+            {
+            case mir_orientation_left:
+            {
+                auto rotated_rect = rectangle;
+                rotated_rect.top_left.x = geom::X(rectangle.top_left.y.as_value());
+                rotated_rect.top_left.y = geom::Y(rectangle.top_left.x.as_value());
```
AlanGriffiths: This looks wrong: it is a reflection, not a rotation.

```suggestion
                rotated_rect.top_left.x = geom::X(rectangle.top_left.y.as_value());
                rotated_rect.top_left.y = geom::Y(surface_size.height.as_value()-rectangle.top_left.x.as_value() - rectangle.size.width.as_value());
```

src/server/scene/basic_surface.cpp
```diff
@@ -365,17 +365,63 @@ bool ms::BasicSurface::input_area_contains(geom::Point const& point) const
...
+                rotated_rect.size.height = geom::Height(rectangle.size.width.as_value());
+                if (rotated_rect.contains(local_point))
+                    return true;
+                break;
+            }
+            case mir_orientation_right:
+            {
+                auto rotated_rect = rectangle;
+                rotated_rect.top_left.x = geom::X(surface_size.width.as_value() - rectangle.top_left.y.as_value() - rectangle.size.height.as_value());
+                rotated_rect.top_left.y = geom::Y(surface_size.height.as_value() - rectangle.top_left.x.as_value() - rectangle.size.width.as_value());
```
AlanGriffiths: This looks wrong: it is a reflection, not a rotation.
```suggestion
                rotated_rect.top_left.x = geom::X(surface_size.width.as_value() - rectangle.top_left.y.as_value() - rectangle.size.height.as_value());
                rotated_rect.top_left.y = geom::Y(rectangle.top_left.x.as_value());
```

AlanGriffiths (approved):
Looks good to me now

Issue #4003, opened 2025-06-10
`GLib-CRITICAL` in `TestMouseKeysConfig` on `fedora-rawhide`


Saviq:
> ( Rolled over from https://github.com/canonical/mir/issues/3931#issuecomment-2908900206)
> ## `TestMouseKeysConfig.mousekeys_config_set_keymap_calls_accessibility_manager_set_mousekeys_keymap`
> 
> 
> Fedora rawhide (the GLib-CRITICAL suggests this is a real bug).
> 
> ```
> Running main() from main.cpp
> Note: Google Test filter = TestMouseKeysConfig.*:-:--
> [==========] Running 4 tests from 1 test suite.
> [----------] Global test environment set-up.
> [----------] 4 tests from TestMouseKeysConfig
> [ RUN ] TestMouseKeysConfig.mousekeys_config_enabled_calls_accessibility_manager_set_mousekeys_enabled
> 
> (process:21630): GLib-CRITICAL **: 02:19:47.335: Failed to get RW lock 0x7fc62bceb9d0: Resource deadlock avoided
> [ OK ] TestMouseKeysConfig.mousekeys_config_enabled_calls_accessibility_manager_set_mousekeys_enabled (21 ms)
> [ RUN ] TestMouseKeysConfig.mousekeys_config_set_keymap_calls_accessibility_manager_set_mousekeys_keymap
> unknown file: Failure
> C++ exception with description "Failed to stop server" thrown in TearDown().
> ``` 

 _Originally posted by @AlanGriffiths in [#3994](https://github.com/canonical/mir/issues/3994#issuecomment-2948575992)_

RAOF:
It looks like this is a result of our workaround for https://bugs.launchpad.net/mir/+bug/1401488? But it's not entirely clear when that happened.

PR #4059, opened 2025-07-03
server/GLibMainLoop: Turn off workaround for now-fixed GLib bug

RAOF:
GLib 2.85 fixes a looooong-standing bug we were hitting. Unfortunately, the fix interacts badly with the workaround we use. Fortunately, we put the workaround behind a `#ifndef GLIB_IS_FIXED`, so we can now just add that `#define` on sufficiently-new GLibs.

C.f. https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2187

Fixes: #4003

### DIFF
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -83,6 +83,17 @@ set_property(
     SOURCE glib_main_loop.cpp glib_main_loop_sources.cpp default_server_configuration.cpp
     PROPERTY COMPILE_OPTIONS -Wno-variadic-macros)
 
+if(GLIB_VERSION VERSION_GREATER_EQUAL 2.85)
+    # This version fixes the (at least) 11 year old bug we encountered in LP #1401488
+    # Unfortunately, the fix interacts badly with our workaround to the bug
+    # Fortunately, we added the workaround we added a #define to turn it off, and
+    # use the correct code!
+    set_property(
+        SOURCE glib_main_loop_sources.cpp
+        APPEND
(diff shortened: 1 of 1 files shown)


src/server/CMakeLists.txt
```diff
@@ -83,6 +83,17 @@ set_property(
     SOURCE glib_main_loop.cpp glib_main_loop_sources.cpp default_server_configuration.cpp
     PROPERTY COMPILE_OPTIONS -Wno-variadic-macros)
 
+if(GLIB_VERSION VERSION_GREATER_EQUAL 2.85)
+    # This version fixes the (at least) 11 year old bug we encountered in LP #1401488
+    # Unfortunately, the fix interacts badly with our workaround to the bug
+    # Fortunately, we added the workaround we added a #define to turn it off, and
+    # use the correct code!
```
jhenstridge, line 90: I guess this is the change that fixed the problem:

https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2187

Would it be worth including a reference to it here? Also: do we care about the case of someone compiling with glib >= 2.85 and deploying to a system with an earlier version?

AlanGriffiths (approved):
Works for me!

> do we care about the case of someone compiling with glib >= 2.85 and deploying to a system with an earlier version?

I don't think that's an important scenario. Snaps are our usual way to deploy on "earlier versions" and they would come with a corresponding glib. Distros should just be consistent

src/server/CMakeLists.txt
RAOF: Correct, that's the fix.

PR #4062, opened 2025-07-03
Update wl_output::geometry to include the correct transformation

mattkae:
## What's new?
- Update `OutputInstance::output_config_changed` to send the right transform in the [geometry event](https://wayland.app/protocols/wayland#wl_output:event:geometry)
- Update `OutputInstance::output_config_changed` to send the untransformed width and height in the [mode event](https://wayland.app/protocols/wayland#wl_output:event:mode)
- Wrote a unit test for the new transformation code

### DIFF
--- a/src/server/frontend_wayland/output_manager.cpp
+++ b/src/server/frontend_wayland/output_manager.cpp
@@ -54,21 +54,6 @@ auto as_subpixel_arrangement(MirSubpixelArrangement arrangement) -> uint32_t
     }
 }
 
-auto transform_size(Size const& size, MirOrientation orientation) -> Size
-{
-    switch (orientation)
-    {
-    case mir_orientation_normal:
-    case mir_orientation_inverted:
-        return size;
-
-    case mir_orientation_left:
-    case mir_orientation_right:
-        return {size.height.as_uint32_t(), size.width.as_uint32_t()};
-    }
-
-    return size;
-}
 }
 
 mf::OutputInstance::OutputInstance(wl_resource* resource, OutputGlobal* global)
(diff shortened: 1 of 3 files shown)


src/server/frontend_wayland/output_manager.cpp
```diff
@@ -105,20 +88,19 @@ auto mf::OutputInstance::output_config_changed(mg::DisplayConfigurationOutput co
...
         "Fake model",
-        mw::Output::Transform::normal);
+        OutputManager::to_output_transform(config.orientation, mir_mirror_mode_none));
 
     for (size_t i = 0; i < config.modes.size(); ++i)
     {
         auto const& mode = config.modes[i];
 
         // As we are not sending the display orientation as a transform (see above),
         // we doctor the size of each mode to match the extents
```
AlanGriffiths: Is this comment still true?

src/server/frontend_wayland/output_manager.cpp
```diff
@@ -365,6 +347,55 @@ auto mf::OutputManager::from_output_transform(int32_t transform) -> std::tuple<M
...
+            {
+                case mir_mirror_mode_horizontal:
+                    return mw::Output::Transform::flipped_270;
+                default:
+                    return mw::Output::Transform::_270;
+            }
+        }
+    }
+
+    return mw::Output::Transform::normal;
```
AlanGriffiths: [Non blocking] This looks like it would be easier to read as a lookup table:

```c++
    return lookup[orientation][mirror_mode];
```

(Yeah, I know orientation isn't [0, 4), which slightly complicates the lookup table)

AlanGriffiths (approved):
LGTM

PR #4064, opened 2025-07-04
Only load drivers compatible to libmirplatform

AlanGriffiths:
Ensure we only load drivers compatible with the current libmirplatform by sourcing the stanza version from the symbols.map.
 
Note: We probably need to be more selective than this in loading drivers (and avoid loading multiple versions of drivers), but this is an improvement on the status quo


tarek-y-ismail (approved):
Cool!

Issue #4065, opened 2025-07-07
TestBounceKeys failing on Alpine


AlanGriffiths:
```
/spread/mir/tests/miral/test_bounce_keys.cpp:104: Failure
Value of: rejection_counter
Expected: is equal to 10
Actual: 9 (of type int)

[ FAILED ] TestBounceKeys/TestDifferentDelays.subsequent_keys_rejected_if_in_within_delay/0, where GetParam() = 35ms (417 ms)
```

```
[ RUN ] TestBounceKeys.different_keys_are_not_rejected
[2025-07-07 14:33:35.655771] < - debug - > mircommon: Cannot check "add_graphics_platform_options" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[2025-07-07 14:33:35.655790] < - debug - > mircommon: Cannot check "describe_graphics_module" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[2025-07-07 14:33:35.655818] < - debug - > mircommon: Cannot check "add_graphics_platform_options" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[2025-07-07 14:33:35.655826] < - debug - > mircommon: Cannot check "describe_graphics_module" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[2025-07-07 14:33:35.655835] < - debug - > mircommon: Cannot check "add_graphics_platform_options" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[2025-07-07 14:33:35.655841] < - debug - > mircommon: Cannot check "describe_graphics_module" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[2025-07-07 14:33:35.655870] < - debug - > mircommon: Cannot check "add_graphics_platform_options" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[2025-07-07 14:33:35.655878] < - debug - > mircommon: Cannot check "describe_graphics_module" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[2025-07-07 14:33:35.656250] < - debug - > mircommon: Cannot check "add_graphics_platform_options" symbol version is "MIR_GRAPHICS_PLATFORM_2.22": dlvsym() is unavailable
[...]

PR #4066, opened 2025-07-08
Don't call out to other objects under lock

AlanGriffiths:
The `BounceKeys::last_key_clearer` callback locks the `runtime_state` and can be called by `last_key_clearer->cancel()`. So the latter should not be called while `runtime_state` is locked.

Fixes: #4065


mattkae (approved):
Not failing for me :+1:

Issue #4047, opened 2025-06-26
NVIDIA driver now requires GBM_BO_USE_RENDERING flag to gbm_surface_create


RAOF:
As of (at least) NVIDIA 575.57.08, `eglCreatePlatformWindowSurface` will fail with `EGL_BAD_ALLOC` when attempting to create an `EGLSurface` on a `gbm_surface` allocated with `gbm_surface_create(..., flags = 0)`.

Unfortunately, on *earlier* versions of the NVIDIA driver, calling `gbm_surface_create(..., flags)` with `flags ≠ 0` fails.

Looks like we need to make our NVIDIA quirk version-dependent 🤦‍♀ 

Before that, `--driver-quirks=gbm-create-surface-flags:driver:nvidia:default` will allow NVIDIA to work.

tarek-y-ismail:
Discussed in yesterday's standup, we can just try to allocate with flags, and if that fails, we can try again _without_ flags. If this doesn't work, then we can also make the quirk's behavior dependent on the driver version.

PR #4070, opened 2025-07-09
Try allocating surfaces with flags first

tarek-y-ismail:
# What's new
Nvidia fixed  `gbm_surface_create`  with their 575 drivers. But that now causes `eglCreatePlatformWindowSurface` to error out with our workaround of passing no flags (see #4047 for the nitty gritty)

We still have to support 570 and 550. After a brief discussion, we agreed that we can try allocating normally which should work for the majority of cases, and if that fails, we can allocate with no flags to specifically support pre 575 nvidia drivers.

In addition, this nullifies the need for `CreateSurfaceFlagsQuirk`, so that has now been removed along with its quirk option.

# How to test
If you have an Nvidia card, run mir with `--platform-display-libs=mir:atomic-kms`. On 570 and earlier, you should get a warning "Failed to create surface with flags. Attempting to create one without flags", but things should still work. On 575, this warning should not be printed.

### DIFF
--- a/src/platforms/atomic-kms/server/gbm_display_allocator.cpp
+++ b/src/platforms/atomic-kms/server/gbm_display_allocator.cpp
@@ -18,6 +18,9 @@
 #include "kms_framebuffer.h"
 #include "kms/quirks.h"
 
+#define MIR_LOG_COMPONENT "atomic-kms:gbm-display-allocator"
+#include "mir/log.h"
+
 #include <drm_fourcc.h>
 #include <xf86drmMode.h>
 #include <gbm.h>
@@ -125,12 +128,9 @@ auto create_gbm_surface(
     gbm_device* gbm,
     geom::Size size,
     mg::DRMFormat format,
-    std::span<uint64_t> modifiers,
-    mga::GbmQuirks const& gbm_quirks) -> std::shared_ptr<gbm_surface>
+    std::span<uint64_t> modifiers) -> std::shared_ptr<gbm_surface>
 {
-    auto const flags = gbm_quirks.gbm_create_surface_flags();
-    auto const surface =
-        [&]()
+    auto const try_create_surface = [&](auto flags)
         {
             if (modifiers.empty())
             {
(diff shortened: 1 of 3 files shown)


src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -18,6 +18,9 @@
 #include "kms_framebuffer.h"
 #include "kms/quirks.h"
 
+#define MIR_LOG_COMPONENT "atomic-kms:gbm-display-allocator"
```
AlanGriffiths: I don't think it is useful to override `MIR_LOG_COMPONENT_FALLBACK` (from `src/platforms/atomic-kms/CMakeLists.txt` with such a long string.

```suggestion
```

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -151,9 +151,21 @@ auto create_gbm_surface(
                     modifiers.size(),
                     flags);
             }
-        }();
+        };
+
+    // Try allocating with flags first
+    // Covers Nvidia cards with 575 drivers and later
```
AlanGriffiths: It is misleading to call out Nvidia here. This works on Intel and AMD too
```suggestion
```

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -151,9 +151,21 @@ auto create_gbm_surface(
...
+    auto surface = try_create_surface(GBM_BO_USE_RENDERING | GBM_BO_USE_SCANOUT);
 
     if (!surface)
+    {
+        mir::log_warning("Failed to create surface with flags. Attempting to create one without flags");
```
AlanGriffiths: I don't think this is something someone running Mir needs care about, so a "warning" seems excessive.

```suggestion
        mir::log_info("Failed to create display buffer surface with correct flags. Attempting to create one without flags");
```

AlanGriffiths (changes requested):
Cosmetic changes

Does it work on your machine?

tarek-y-ismail:
> Does it work on your machine? 

The path for < 575, yes. I didn't try 575 since that's not available on Noble yet. When my proper internet comes back, I might try upgrading to test that if you want.

PR #4071, opened 2025-07-09
Bring MouseKeys up to our current standards

tarek-y-ismail:
# What's new
MouseKeys was the first accessibility method to be implemented. Through reviews of later accessibility methods, I learned about multiple things that could be modified and updated to improve it. This PR implements those things.

# How to test
You can follow the same procedure as #3757


include/miral/miral/mousekeys_config.h
```diff
@@ -27,20 +27,57 @@ class Server;
...
+    class Store;
+}
 class MouseKeysConfig
 {
 public:
+    [[deprecated(
+        "MouseKeysConfig(bool) is deprecated. Please use MouseKeysConfig::enabled or MouseKeysConfig::disabled")]]
     explicit MouseKeysConfig(bool enabled_by_default);
 
+    /// Creates a `MouseKeys` instance that's enabled by default.
```
tarek-y-ismail:
```suggestion
    /// Creates a `MouseKeysConfig` instance that's enabled by default.
```

src/include/server/mir/shell/mousekeys_transformer.h
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_SHELL_MOUSEKEYS_TRANSFORMER_H
+#define MIR_SHELL_MOUSEKEYS_TRANSFORMER_H
+
+#include "mir/input/input_event_transformer.h"
+#include "mir/input/mousekeys_keymap.h"
```
tarek-y-ismail: Can probably replace this include with a forward decl

src/include/server/mir/shell/mousekeys_transformer.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <xkbcommon/xkbcommon-keysyms.h>
```
tarek-y-ismail: Of what use is this?

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,354 @@
...
+auto constexpr acceleration_linear = 100.0;
+auto constexpr acceleration_quadratic = 30.0;
+auto constexpr max_speed_x = 400.0;
+auto constexpr max_speed_y = 400.0;
+}
+
+template <typename T>
+auto maybe_option(std::shared_ptr<mir::options::Option> const& options, char const* const option_name) -> std::optional<T>
+{
+    return options->is_set(option_name) ? options->get<bool>(option_name) : std::optional<T>{};
```
tarek-y-ismail:
```suggestion
    return options->is_set(option_name) ? options->get<T>(option_name) : std::optional<T>{};
```

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,354 @@
...
+            if (!value) return;
+
+            if (*value)
+                this->enable();
+            else
+                this->disable();
+             
+        });
+
+    auto const ensure_non_negative = [](auto key, auto val, auto do_the_thing)
```
tarek-y-ismail: Can probably ditch `auto key` and rename `do_the_thing` to be more descriptive (`do_something_with_non_negative_val`?), and since all uses of this are floats, might as well ditch the other `auto`s as well

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,354 @@
...
+        }
+        else
+        {
+            mir::log_warning(
+                "Config value %s does not support negative values. Ignoring the supplied value (%f)...",
+                key.to_string().c_str(),
+                *val);
+        }
+    };
+
```
tarek-y-ismail, line 165: Can probably compress this and its friends to a list of `(key, description, setter)` and a for loop instead of all this code

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,354 @@
...
+        server.add_configuration_option(
+            mouse_keys_max_speed_x,
+            "The maximum speed in pixels/second the mousekeys pointer can "
+            "reach on the x axis. Pass zero to disable the speed limit",
+            current_value_or_default(s->max_speed_x, defaults::max_speed_x));
+        server.add_configuration_option(
+            mouse_keys_max_speed_y,
+            "The maximum speed in pixels/second the mousekeys pointer can "
+            "reach on the y axis. Pass zero to disable the speed limit",
+            current_value_or_default(s->max_speed_x, defaults::max_speed_x));
```
tarek-y-ismail:
```suggestion
            current_value_or_default(s->max_speed_y, defaults::max_speed_y));
```

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,354 @@
...
-            if (options->is_set(enable_mouse_keys_opt))
-                enable = options->get<bool>(enable_mouse_keys_opt);
+            if(enable)
+                this->enable();
 
-            enabled(enable);
+            // If a config (CLI or environment variable) option is set, then it
+            // overrides both the state set options (until server init at
+            // least) and defaults. Otherwise, state set options override
+            // defaults. Defaults are used only if neither are provided.
```
tarek-y-ismail, line 335: I'm not quite sure about this. In my mind, if I set an option during launch, then that should override even configuration files, but I'm not sure if this is the convention or not.

src/miral/symbols.map
```diff
@@ -604,6 +604,9 @@ global:
     "miral::CursorScale::CursorScale(miral::live_config::Store&)";
     "miral::InputConfiguration::InputConfiguration(miral::live_config::Store&)";
     "miral::OutputFilter::OutputFilter(miral::live_config::Store&)";
+    miral::MouseKeysConfig::disable*;
+    miral::MouseKeysConfig::disabled*;
```
tarek-y-ismail: Note that `enabled*` is already covered by the existing symbol `enabled(bool)`

src/server/symbols.map
```diff
@@ -720,6 +720,7 @@ global:
     mir::shell::AbstractShell::surface_ready*;
     mir::shell::AbstractShell::swap_z_order*;
     mir::shell::AccessibilityManager::?AccessibilityManager*;
+    mir::shell::AccessibilityManager::mousekeys*;
```
tarek-y-ismail, line 726: I'm very confused why the symbol generator chose to export `mousekeys` but not `simulated_secondary_click` or other accessibility symbols.

src/miral/symbols.map
tarek-y-ismail: I'm having second thoughts about this actually, would it be safe to add a duplicate in the 5.4 section?

debian/libmiral7.symbols
```diff
@@ -578,6 +578,15 @@ libmiral.so.7 libmiral7 #MINVER#
...
  (c++)"miral::InputConfiguration::InputConfiguration(miral::live_config::Store&)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::MouseKeysConfig(miral::live_config::Store&)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::MouseKeysConfig(std::shared_ptr<miral::MouseKeysConfig::Self>)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::disable() const@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::disable()@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::disabled()@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::enable() const@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::enable()@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::enabled()@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::enabled()@MIRAL_5.4" 5.4.0
```
tarek-y-ismail: dupe

debian/libmiral7.symbols
```diff
@@ -578,6 +578,15 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::BounceKeys::operator()(mir::Server&)@MIRAL_5.4" 5.4.0
  (c++)"miral::CursorScale::CursorScale(miral::live_config::Store&)@MIRAL_5.4" 5.4.0
  (c++)"miral::InputConfiguration::InputConfiguration(miral::live_config::Store&)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::MouseKeysConfig(miral::live_config::Store&)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::MouseKeysConfig(std::shared_ptr<miral::MouseKeysConfig::Self>)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::disable() const@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::disable()@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::disabled()@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::enable() const@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::enable()@MIRAL_5.4" 5.4.0
```
tarek-y-ismail: Why is this here?

debian/libmiral7.symbols
```diff
@@ -578,6 +578,15 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::BounceKeys::operator()(mir::Server&)@MIRAL_5.4" 5.4.0
  (c++)"miral::CursorScale::CursorScale(miral::live_config::Store&)@MIRAL_5.4" 5.4.0
  (c++)"miral::InputConfiguration::InputConfiguration(miral::live_config::Store&)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::MouseKeysConfig(miral::live_config::Store&)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::MouseKeysConfig(std::shared_ptr<miral::MouseKeysConfig::Self>)@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::disable() const@MIRAL_5.4" 5.4.0
+ (c++)"miral::MouseKeysConfig::disable()@MIRAL_5.4" 5.4.0
```
tarek-y-ismail: This too!

tarek-y-ismail:
Trying to appease the Debian symbols check, but hopefully we won't need all the changes in place right now

tarek-y-ismail:
To properly test `MouseKeysConfig` with the new changes, we need to pass a some sort of mock mousekeys transformer, which is not possible at the moment because `InputEventTransformer::Transformer` is defined in `input_event_transformer.h`  and that depends on the seat, and that depends on `mir/events/xkb_modifiers.h` which miral cannot access. 

One possible solution is to split `InputEventTransformer` and `InputEventTransformer::Transformer` into different header files, in which case we should be able to include it and create a mock mousekeys transformer, and then return that from `MockAccessibilityManager::mousekeys()`.

debian/libmiral7.symbols
```diff
@@ -538,7 +538,6 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::InputConfiguration::Touchpad::middle_mouse_button_emulation() const@MIRAL_5.3" 5.3.0
  (c++)"miral::InputConfiguration::Touchpad::middle_mouse_button_emulation(std::optional<bool> const&)@MIRAL_5.3" 5.3.0
  (c++)"miral::MouseKeysConfig::MouseKeysConfig(bool)@MIRAL_5.3" 5.3.0
- (c++)"miral::MouseKeysConfig::enabled(bool) const@MIRAL_5.3" 5.3.0
```
AlanGriffiths: This symbols should not be removed

src/miral/symbols.map
AlanGriffiths: We have to disambiguate by spelling things out without wildcards: C.f. `"miral::CursorScale::CursorScale()";`

Note: We should probably switch away from wildcards everywhere, but when the utilities were originally written the linker didn't understand this quoted format.

tests/miral/test_mousekeys_config.cpp
```diff
@@ -79,14 +77,16 @@ TEST_F(TestMouseKeysConfig, mousekeys_config_enabled_calls_accessibility_manager
 
     add_server_init(config);
     start_server();
-    config.enabled(true);
-    config.enabled(false);
```
AlanGriffiths, line 85: We shouldn't drop tests for deprecated, but supported functions

src/miral/mousekeys_config.cpp
AlanGriffiths: With these "old" configuration settings the order is:

1. command line; or else
2. environment variable; or else
3. user config file; or else
4. system config file; or else
5. default (if any)

AlanGriffiths: We might also want to consider if we even want to support these "old" options when the "new" API is used. (Is more pain now, but allows us to drop them later.)

tests/miral/test_mousekeys_config.cpp
```diff
@@ -79,14 +77,16 @@ TEST_F(TestMouseKeysConfig, mousekeys_config_enabled_calls_accessibility_manager
...
+    config.enable();
+    config.disable();
 }
 
+
+// TODO: Maybe we should remove these tests?
```
AlanGriffiths: Well, the test names and the test bodies now disagree on what is being tested. So something should change

tests/miral/test_mousekeys_config.cpp
```diff
@@ -79,14 +77,16 @@ TEST_F(TestMouseKeysConfig, mousekeys_config_enabled_calls_accessibility_manager
...
 TEST_F(TestMouseKeysConfig, mousekeys_config_set_keymap_calls_accessibility_manager_set_mousekeys_keymap)
 {
     // We don't call `set_keymap` at startup
-    EXPECT_CALL(*accessibility_manager, mousekeys_keymap(_));
+    EXPECT_CALL(*accessibility_manager, mousekeys());
 
     add_server_init(config);
     start_server();
```
AlanGriffiths: `mousekeys_config_set_keymap_calls_accessibility_manager_set_mousekeys_keymap` is a pretty bad test name. We should be testing that functionality "works" not that a specific function is called.

The poor name is evident now as the implementation has changed, leading to a change in the test body and the name being wrong.

But the test should be named something like "configuring_keymap_updates_mousekeys" and that test name would have remained valid. But this test implementation fails to test what it should.

AlanGriffiths (dismissed):

tests/miral/test_mousekeys_config.cpp
tarek-y-ismail: The tests right now are band-aided so the code can compile, but I agree that they need to be updated and renamed. 

I don't want to test if the setter calls `AccessibilityManager::mousekeys` since that leaves the possibility of the wrong setter being called on the transformer. To really make sure that the right methods are called, I'd need to pass a mock transformer to the accessibility manager and expect calls on that. That's not straightforward at the moment (see [this comment](https://github.com/canonical/mir/pull/4071#issuecomment-3056650934)). 

The change to that possible itself shouldn't be too difficult, I'll just need to move `Transformer` out of `InputEventTransformer` into its own header, and patch up all previous references to it. But that will require changes to a handful of files. I'm thinking of doing that in a separate PR so things don't balloon up here. What do you think?

AlanGriffiths: > To really make sure that the right methods are called, I'd need to pass a mock transformer to the accessibility manager and expect calls on that. That's not straightforward at the moment (see [this comment](https://github.com/canonical/mir/pull/4071#issuecomment-3056650934)).
> 
> The change to that possible itself shouldn't be too difficult, I'll just need to move `Transformer` out of `InputEventTransformer` into its own header, and patch up all previous references to it. But that will require changes to a handful of files. I'm thinking of doing that in a separate PR so things don't balloon up here. What do you think?

That sounds like a sensible approach: fix the stuff that makes this change hard and iterate. (https://understandlegacycode.com/blog/a-process-to-do-safe-changes-in-a-complex-codebase/)

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,322 @@
...
+auto miral::MouseKeysConfig::disabled() -> MouseKeysConfig 
+{
+    return MouseKeysConfig{std::make_shared<Self>(false)};
+}
+
+miral::MouseKeysConfig::MouseKeysConfig(live_config::Store& config_store)
+{
+    config_store.add_bool_attribute(
+        {"mouse_keys", "enable"},
+        "Whether or not to start the compositor with mousekeys enabled",
```
tarek-y-ismail: Same mistake as here https://github.com/canonical/mir/pull/4074#discussion_r2197905067

tests/miral/test_mousekeys_config.cpp
tarek-y-ismail: The compiler complains about deprecated methods and refuses to compile

include/miral/miral/mousekeys_config.h
```diff
@@ -27,20 +27,57 @@ class Server;
...
+}
 class MouseKeysConfig
 {
 public:
+    [[deprecated(
+        "MouseKeysConfig(bool) is deprecated. Please use MouseKeysConfig::enabled or MouseKeysConfig::disabled")]]
     explicit MouseKeysConfig(bool enabled_by_default);
 
+    /// Creates a `MouseKeysConfig` instance that's enabled by default.
+    /// \remark Since MirAL 5.4
```
mattkae: Should this be `5.5`? I think 5.4 got burned (although it doesn't seem to be updated in the code!)

src/include/server/mir/shell/mousekeys_transformer.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace mir
+{
+namespace input
+{
+class MouseKeysKeymap;
+}
+namespace shell
+{
+class MouseKeysTransformer: public mir::input::Transformer::Transformer
```
mattkae: Why is this `Transformer::Transformer` :thinking: ?

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,322 @@
...
+namespace
+{
 char const* const enable_mouse_keys_opt = "enable-mouse-keys";
 char const* const mouse_keys_acceleration_constant_factor = "mouse-keys-acceleration-constant-factor";
 char const* const mouse_keys_acceleration_linear_factor = "mouse-keys-acceleration-linear-factor";
 char const* const mouse_keys_acceleration_quadratic_factor = "mouse-keys-acceleration-quadratic-factor";
 char const* const mouse_keys_max_speed_x = "mouse-keys-max-speed-x";
 char const* const mouse_keys_max_speed_y = "mouse-keys-max-speed-y";
 
+namespace defaults
```
mattkae, line 48: Nit: I would leave these in the anonymous namespace. This pattern doesn't really happen in mir

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,322 @@
...
+{
+    return MouseKeysConfig{std::make_shared<Self>(false)};
+}
+
+miral::MouseKeysConfig::MouseKeysConfig(live_config::Store& config_store)
+{
+    config_store.add_bool_attribute(
+        {"mouse_keys", "enable"},
+        "Whether or not mousekeys is enabled",
+        [this](live_config::Key const&, std::optional<bool> value)
```
mattkae, line 135: Should this have a default?

```suggestion
    config_store.add_bool_attribute(
        {"mouse_keys", "enable"},
        "Whether or not mousekeys is enabled",
        false,
        [this](live_config::Key const&, std::optional<bool> value)
```

src/miral/mousekeys_config.cpp
```diff
@@ -24,121 +28,322 @@
...
+        {
+            mir::log_warning(
+                "Config value %s does not support negative values. Ignoring the supplied value (%f)...",
+                key.to_string().c_str(),
+                *val);
+        }
+    };
+
+    std::initializer_list<std::tuple<live_config::Key, std::string_view, std::optional<double> Self::State::*>>
+        float_attribs = {
```
mattkae, line 167: Should each of these have defaults as well?

src/miral/mousekeys_config.cpp
mattkae: I would vote for removing them

src/include/server/mir/shell/mousekeys_transformer.h
tarek-y-ismail: Typo, but you can extend it to `mir::input::Transformer::Transformer::Transformer::Transformer::Transformer` and beyond and it compiles fine! Not sure what's going on

src/miral/mousekeys_config.cpp
tarek-y-ismail: Maybe we should start doing it? I think it's pretty harmless to leave it as is at the moment since this is all internal code anyway

src/miral/mousekeys_config.cpp
tarek-y-ismail: We already handle defaults with `option_or_current_value_or_default` on server startup

src/miral/mousekeys_config.cpp
tarek-y-ismail: I think Alan meant that we could check if both option types were used and maybe print out a warning so we give users time to adjust. Something like: "Old CLI/environment based options are being phased out in favor of live config options and will be removed in a future release"

AlanGriffiths: Alan meant that the shell explicitly chooses to use live_config. If it does, we don’t need to support the legacy options.

(if it doesn’t, we shouldn’t break existing behaviour - only deprecate it)

src/miral/symbols.map
```diff
@@ -629,6 +629,12 @@ global:
     miral::StickyKeys::on_modifier_clicked*;
     miral::StickyKeys::operator*;
     miral::StickyKeys::should_disable_if_two_keys_are_pressed_together*;
+    "miral::MouseKeysConfig::MouseKeysConfig(miral::live_config::Store&)";
+    "miral::MouseKeysConfig::MouseKeysConfig(std::shared_ptr<miral::MouseKeysConfig::Self>)";
+    "miral::MouseKeysConfig::disable() const";
+    "miral::MouseKeysConfig::disabled()";
+    "miral::MouseKeysConfig::enable() const";
+    "miral::MouseKeysConfig::enabled()";
```
mattkae: Did you manually add these yourself? If so - you will have a bad time :)

mattkae: I am very confused by the troubles that you're having with symbols :smile: We should chat about this tomorrow

mattkae (changes requested):

src/miral/symbols.map
tarek-y-ismail: Yes, I did. I think this has a similar format to the Debian symbols file, so I could theoretically push the changes and wait for CI to implode (because the Debian symbols check doesn't provide the same output on my machine as CI). I could then take the mangled symbols, demangle them, and put them here. 

Compared to that, writing symbols by hand is a bit faster. I think a good temporary solution until we get the symbol generator beefed up is to generate the symbols with wildcards, then manually go through them and replace the wildcards with concrete text.

mattkae: I believe that this is the wrong way to go :)

We can generate the symbols.map. I think the only issue here is the new constructors, which we'll probably have to add by hand.

mattkae: Oh or maybe the enable/disable bits are also breaking you :facepalm:

mattkae: At any rate, I would not add all of the symbols there. It's only going to make things worse :smile:

mattkae (approved):
LGTM

PR #4074, opened 2025-07-10
Simulated secondary click cleanup

tarek-y-ismail:
Similarly to #4071, this PR cleans up and updates simulated secondary click code, mainly:
- Deprecates the confusing boolean constructor, providing named alternatives. 
- Restores the old config options as live config options
- Actually explains what is simulated secondary click in the user facing header


src/miral/simulated_secondary_click.cpp
```diff
@@ -54,6 +55,78 @@ miral::SimulatedSecondaryClick::SimulatedSecondaryClick(bool enabled_by_default)
...
+                return;
+
+            if (*val < 0)
+            {
+                mir::log_warning(
+                    "Config value %s does not support negative values. Ignoring the supplied value (%d)...",
+                    key.to_string().c_str(),
+                    *val);
+                return;
+            }
```
tarek-y-ismail, line 109: This pattern keeps repeating in different live config setup code sections, maybe it'd be a wise idea to pull that out into a miral header/source file combo?

Something like:

```
auto validate_non_negative_integer(live_config::Key const& key, std::optional<int> val) -> bool
{
    // Do the thing
}
```

If our logging system didn't use printf specifiers, we could have one templated function, but alas...

src/miral/symbols.map
```diff
@@ -604,6 +604,8 @@ global:
     "miral::CursorScale::CursorScale(miral::live_config::Store&)";
     "miral::InputConfiguration::InputConfiguration(miral::live_config::Store&)";
     "miral::OutputFilter::OutputFilter(miral::live_config::Store&)";
+    miral::SimulatedSecondaryClick::disabled*;
+    miral::SimulatedSecondaryClick::enabled*;
```
tarek-y-ismail: As Alan mentioned in #4071, wildcards should be dropped

src/miral/simulated_secondary_click.cpp
```diff
@@ -54,6 +55,78 @@ miral::SimulatedSecondaryClick::SimulatedSecondaryClick(bool enabled_by_default)
 {
 }
 
+miral::SimulatedSecondaryClick::SimulatedSecondaryClick(live_config::Store& config_store)
+{
+    config_store.add_bool_attribute(
+        {"simulated_secondary_click", "enable"},
+        "Whether or not to start the compositor with simulated secondary click enabled",
```
mattkae: Technically the user can cause a config update at any time, not just when the compositor starts

```suggestion
        "Whether or not simulated secondary click is enabled",
```

src/miral/simulated_secondary_click.cpp
AlanGriffiths: Nobody says you can't fix the problems you encounter

PR #4075, opened 2025-07-10
Move `InputEventTransformer::Transformer` into its own header

tarek-y-ismail:
Prerequisite to test `MouseKeysConfig` properly in #4071. 

In the aforementioned PR, I wanted to add tests for `miral::MouseKeyConfig`, which required that I implement a `MockMouseKeysTransformer`, which requires the definition of `Transformer`. That is defined in `input_event_transformer.h`. Unfortunately, since `InputEventTransformer` inherits from `Seat`, it also includes `seat.h`, which transitively includes headers that can't be included in miral. The solution was to split the definition of `Transformer` so that these headers are not included.

### DIFF
--- a/include/common/mir/input/transformer.h
+++ b/include/common/mir/input/transformer.h
@@ -1,0 +1,57 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_INPUT_TRANSFORMER_H_
+#define MIR_INPUT_TRANSFORMER_H_
+
+#include <functional>
+#include <memory>
+
+class MirEvent;
+
+namespace mir
+{
+namespace input
+{
+class EventBuilder;
+/// Transformers are used early on in the input processing pipeline to process
+/// events for various accessibilty methods. They can consume or pass events,
+/// similar to filters. They additionally can emit events at a later time using
+/// the passed `EventBuilder` and `EventDispatcher`.
+///
+/// Do note that despite the event dispatcher being a reference and the event
+/// builder being a pointer, they're guaranteed to outlive the transformer.
+///
+/// Transformers need to be registered to an `InputEventTransformer` to receive
+/// events. At the moment, the central instance of `InputEventTransformer` is
+/// not accessible from miral. This class is only exposed for testing.
+class Transformer
+{
+public:
+    using EventDispatcher = std::function<void(std::shared_ptr<MirEvent>)>;
+
+    virtual ~Transformer() = default;
+
+    // Returning true means that the event has been successfully processed and
+    // shouldn't be handled by later transformers, whether the transformer is
+    // accumulating events for later dispatching or has immediately dispatched
+    // an event is an implementation detail of the transformer
+    virtual bool transform_input_event(EventDispatcher const&, EventBuilder*, MirEvent const&) = 0;
+};
+}
+}
+
+#endif // MIR_INPUT_INPUT_EVENT_TRANSFORMER_H_ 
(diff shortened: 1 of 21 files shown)


CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
 set(MIR_VERSION_MAJOR 2)
-set(MIR_VERSION_MINOR 22)
```
AlanGriffiths: We haven't branched 22 yet - this isn't needed

debian/control
```diff
@@ -66,7 +66,7 @@ Vcs-Git: https://github.com/canonical/mir
 
 #TODO: Packaging infrastructure for better dependency generation,
 #      ala pkg-xorg's xviddriver:Provides and ABI detection.
-Package: libmirserver65
```
AlanGriffiths: Mir 2.21 has 64, no need to bump this again

src/common/symbols.map
```diff
@@ -720,3 +720,14 @@ global:
...
   };
 } MIR_COMMON_2.19;
+
+MIR_COMMON_2.22 {
+global:
+  extern "C++" {
+    mir::input::Transformer::?Transformer*;
+    non-virtual?thunk?to?mir::input::Transformer::?Transformer*;
+    typeinfo?for?mir::input::Transformer;
+    vtable?for?mir::input::Transformer;
```
AlanGriffiths: I suspect none of these symbols exist in libmircommon

tarek-y-ismail: I had to move the new header under `src/include/mir/common/input`, was that the wrong place? Note that this is the output of the symbol generator,/checker, not handwritten

AlanGriffiths: No, but as there's no implementation file in mircommon I doubt that there are any symbols exported.

tarek-y-ismail: Oh, so I got bamboozled...

include/common/mir/input/transformer.h
```diff
@@ -0,0 +1,46 @@
...
+#include <memory>
+
+class MirEvent;
+
+namespace mir
+{
+namespace input
+{
+class EventBuilder;
+class Transformer
```
mattkae, line 41: Should this class include some information about _when_ these events are processed? It is somewhat important that the events that they process are very early in the events pipeline

mattkae: Perhaps this can have a better name too :) Like `EventPreprocessor` or something.

tools/symbols_map_generator/main.py
```diff
@@ -42,6 +42,10 @@
     "miral::SimulatedSecondaryClick::enable*;",
     "miral::SimulatedSecondaryClick::enabled*;",
     "miral::SimulatedSecondaryClick::operator*;",
+    "mir::input::Transformer::?Transformer*;",
+    "non-virtual?thunk?to?mir::input::Transformer::?Transformer*;",
+    "typeinfo?for?mir::input::Transformer;",
+    "vtable?for?mir::input::Transformer;",
```
mattkae, line 48: Why?

mattkae: Oh I see. I'm surprised that this happened though!

mattkae: I don't think you need this. I just removed it and then ran the symbols generator, and everything was ok :)

tarek-y-ismail: It seems like the symbol generator was in a joking mood :trollface: 
I removed both the `Transformer` and `SimulatedSecondaryClick` symbols from the hidden symbols list and now it doesn't generate them!

For the sake of this PR, I'll just remove the `Transformer` entries, but I made an issue (#4094) so we don't forget to review this list.

include/common/mir/input/transformer.h
tarek-y-ismail: > Should this class include some information about when these events are processed? It is somewhat important that the events that they process are very early in the events pipeline

At the moment, this class is exposed only for the purposes of testing. But a brief comment explaining things is always a good thing ;)

> Perhaps this can have a better name too :) Like EventPreprocessor or something.

I myself am not completely proud of the current name, but I think this is a problem for later since a proper rename will involve a lot of churn in code and filenames. Also, I'm not sure if `EventPreprocessor` captures the asynchronousity of transformers, I guess `Transformer` also doesn't do that...

tools/symbols_map_generator/main.py
tarek-y-ismail: Nevermind, I think we both made the same mistake of checking mir**server** symbols and not mir**common**

mattkae: Ah okay that is fair! I don't think Id did mircommon either

include/common/mir/input/transformer.h
mattkae: Yeah `Transformer` is very generic. We can revisit later though, I agree

src/common/symbols.map
```diff
@@ -720,3 +720,4 @@ global:
     typeinfo?for?mir::time::is_steady_specialisation;
   };
 } MIR_COMMON_2.19;
+
```
mattkae: Nit: Spurious newline :)

mattkae (approved):
Looks good by me!

PR #4076, opened 2025-07-10
feature: implement sticky keys accessibility feature

mattkae:


debian/libmiral7.symbols
```diff
@@ -594,12 +594,19 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::SlowKeys::disabled()@MIRAL_5.4" 5.4.0
  (c++)"miral::SlowKeys::enable()@MIRAL_5.4" 5.4.0
  (c++)"miral::SlowKeys::enabled()@MIRAL_5.4" 5.4.0
- (c++|arch-bits=64)"miral::SlowKeys::hold_delay(std::chrono::duration<long, std::ratio<1l, 1000l> >)@MIRAL_5.4" 5.4.0
```
tarek-y-ismail: We really need to make a PR to follow the symbol generator's ordering so that changes like this stop popping up in PRs...

examples/miral-shell/shell_main.cpp
```diff
@@ -270,6 +272,30 @@ int main(int argc, char const* argv[])
         return false;
     };
 
+    StickyKeys sticky_keys = StickyKeys::disabled();
+    bool sticky_keys_enabled = false;
+    auto const sticky_keys_filter = [sticky_keys=sticky_keys, &sticky_keys_enabled](MirKeyboardEvent const* key_event) mutable
```
tarek-y-ismail: Can probably do
```suggestion
    auto const sticky_keys_filter = [sticky_keys, sticky_keys_enabled=false](MirKeyboardEvent const* key_event) mutable
```

include/miral/miral/sticky_keys.h
```diff
@@ -0,0 +1,68 @@
...
+namespace mir
+{
+class Server;
+}
+
+namespace miral
+{
+namespace live_config { class Store; }
+
+class StickyKeys
```
tarek-y-ismail, line 34: Should have a comment explaining what sticky keys is (just copy the PR description), and a remark noting when it was introduced

include/miral/miral/sticky_keys.h
```diff
@@ -0,0 +1,68 @@
...
+
+    // Creates a 'StickyKeys' instance that is disabled by default.
+    static auto disabled() -> StickyKeys;
+
+    /// Construct a 'StickyKeys' instance with access to a live config store.
+    explicit StickyKeys(live_config::Store& config_store);
+
+    void operator()(mir::Server& server);
+
+    /// Enables sticky keys.
```
tarek-y-ismail: Maybe mention that double enables/disables are idempotent

include/miral/miral/sticky_keys.h
```diff
@@ -0,0 +1,68 @@
...
+    StickyKeys& enable();
+
+    /// Disables sticky keys
+    StickyKeys& disable();
+
+    /// When set to true, clicking two modifier keys are once will result
```
tarek-y-ismail: The usage of "clicking" with physical buttons is a bit awkward. Maybe "pressing down" is more fitting?

include/miral/miral/sticky_keys.h
```diff
@@ -0,0 +1,68 @@
...
+    /// in sticky keys being temporarily disabled until all keys are released.
+    /// Defaults to true.
+    StickyKeys& should_disable_if_two_keys_are_pressed_together(bool on);
+
+    /// Configure a callback to be invoked whenever a modifier key is clicked
+    /// and is now sticky pending an action that will release it.
+    StickyKeys& on_modifier_clicked(std::function<void(int32_t)>&&);
```
tarek-y-ismail, line 63: Should note that the passed integer is a bitmask of modifiers

src/include/server/mir/shell/sticky_keys_transformer.h
```diff
@@ -0,0 +1,35 @@
...
+class StickyKeysTransformer : public input::InputEventTransformer::Transformer
+{
+public:
+    virtual void should_disable_if_two_keys_are_pressed_together(bool) = 0;
+    virtual void on_modifier_clicked(std::function<void(int32_t)>&&) = 0;
+};
+
+}
+
+#endif //MIR_SHELL_STICKY_KEYS_TRANSFORMER_H
```
tarek-y-ismail: Nit:
```suggestion
#endif // MIR_SHELL_STICKY_KEYS_TRANSFORMER_H
```

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+class miral::StickyKeys::Self
+{
+public:
+    explicit Self(bool enabled)
+        : state(State{enabled})
+    {
+    }
+
+    void set_accessibility_manager(std::shared_ptr<msh::AccessibilityManager> const& manager)
```
tarek-y-ismail, line 36: Maybe rename to `on_server_init`?

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "mir/shell/sticky_keys_transformer.h"
+
+namespace mi = mir::input;
+namespace msh = mir::shell;
+
+class miral::StickyKeys::Self
+{
+public:
+    explicit Self(bool enabled)
+        : state(State{enabled})
```
tarek-y-ismail: Nit: We usually prefer curly braces over parens
```suggestion
        : state{State{enabled}}
```

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,190 @@
...
+            auto const locked_state = state.lock();
+            locked_state->on_modifier_clicked = std::move(callback);
+        }
+    }
+
+private:
+    struct State
+    {
+        explicit State(bool enabled)
+            : enabled(enabled)
```
tarek-y-ismail:
```suggestion
            : enabled{enabled}
```

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,190 @@
...
+        "When set to true, clicking two modifier keys are once will result "
+        "in sticky keys being temporarily disabled until all keys are released",
+        [this](live_config::Key const&, std::optional<bool> val)
+        {
+            if (val)
+            {
+                if (*val)
+                    this->should_disable_if_two_keys_are_pressed_together(true);
+                else
+                    this->should_disable_if_two_keys_are_pressed_together(false);
```
tarek-y-ismail:
```suggestion
                this->should_disable_if_two_keys_are_pressed_together(*val);
```

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,190 @@
...
+            }
+        });
+}
+
+miral::StickyKeys::StickyKeys(std::shared_ptr<Self> self)
+    : self(std::move(self))
```
tarek-y-ismail:
```suggestion
    : self{std::move(self)}
```

src/miral/symbols.map
```diff
@@ -567,18 +572,13 @@ global:
...
-    "miral::SimulatedSecondaryClick::displacement_threshold(float)";
-    "miral::SimulatedSecondaryClick::enable()";
     miral::SimulatedSecondaryClick::hold_duration*;
     miral::SimulatedSecondaryClick::on_hold_cancel*;
     miral::SimulatedSecondaryClick::on_hold_start*;
     miral::SimulatedSecondaryClick::on_secondary_click*;
-    "miral::SimulatedSecondaryClick::operator()(mir::Server&)";
+    miral::WindowInfo::tiled_edges*;
+    miral::WindowSpecification::tiled_edges*;
+    miral::WindowSpecification::to_surface_specification*;
```
tarek-y-ismail: Too many non-sticky keys changes. Do you mind reverting them?

src/server/input/input_event_transformer.cpp
```diff
@@ -116,28 +116,33 @@ mir::input::InputEventTransformer::~InputEventTransformer() = default;
...
 bool mi::InputEventTransformer::transform(MirEvent const& event)
 {
-    std::lock_guard lock{mutex};
-
-    for (auto it = input_transformers.begin(); it != input_transformers.end();)
+    // This cache needs to be introduced because a transformer may be removed in response
+    // to one of its own dispatched events (e.g. a transformer may dispatch some sort of
+    // click that removes the transformer itself). As a result, we cannot be holding the
+    // lock to the transformer while it is trying to remove itself.
+    std::vector<std::shared_ptr<Transformer>> transformers_cache;
```
tarek-y-ismail, line 123: _Very minor_ performance optimization: You can reserve `input_transformers.size()` so we allocate only once.

src/server/input/input_event_transformer.cpp
```diff
@@ -116,28 +116,33 @@ mir::input::InputEventTransformer::~InputEventTransformer() = default;
 
 bool mi::InputEventTransformer::transform(MirEvent const& event)
 {
-    std::lock_guard lock{mutex};
-
-    for (auto it = input_transformers.begin(); it != input_transformers.end();)
+    // This cache needs to be introduced because a transformer may be removed in response
+    // to one of its own dispatched events (e.g. a transformer may dispatch some sort of
+    // click that removes the transformer itself). As a result, we cannot be holding the
+    // lock to the transformer while it is trying to remove itself.
```
tarek-y-ismail, line 122: If I understand correctly, the issue stems from the lock being held while `transform_input_event` is being used. In the case of sticky keys, you can press ctrl and alt, then s to toggle it, which would then dispatch an event that would try to lock the mutex again, causing a deadlock before reaching the filter down the chain, right?

src/server/shell/basic_accessibility_manager.cpp
tarek-y-ismail, line 1: Maybe it would be wiser to split the namespace changes into their own PR?

src/server/shell/basic_sticky_keys_transformer.cpp
```diff
@@ -0,0 +1,184 @@
...
+        KEY_NUMLOCK
+    };
+
+    for (auto const modifier : modifiers)
+    {
+        if (modifier == scan_code)
+            return true;
+    }
+
+    return false;
```
tarek-y-ismail, line 54: You can probably use `std::find` or use a map (I know it won't help performance, just readability)

src/server/shell/basic_sticky_keys_transformer.cpp
```diff
@@ -0,0 +1,184 @@
...
+    bool is_temporarily_disabled = false;
+    std::vector<std::shared_ptr<MirKeyboardEvent>> pending;
+
+    bool process_event(
+        MirEvent const& event,
+        bool is_up_event,
+        mi::InputEventTransformer::EventDispatcher const& dispatcher)
+    {
+        if (!is_up_event)
+            return false;
```
tarek-y-ismail, line 77: Maybe invert the flow? Instead of checking inside the method, check before calling it at all

src/server/shell/basic_sticky_keys_transformer.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+        dispatcher(std::shared_ptr<MirEvent>(event.clone()));
+
+        for (auto const& item : pending)
+        {
+            dispatcher(item);
+        }
```
tarek-y-ismail, line 84: I think we should dispatch the pending events before the new one. Also, can't we just dispatch one event with modifiers set?

src/server/shell/basic_sticky_keys_transformer.cpp
```diff
@@ -0,0 +1,184 @@
...
+        pending.clear();
+        return true;
+    }
+};
+
+msh::BasicStickyKeysTransformer::BasicStickyKeysTransformer()
+    : self(std::make_shared<Self>())
```
tarek-y-ismail:
```suggestion
    : self{std::make_shared<Self>()}
```

src/server/shell/basic_sticky_keys_transformer.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    auto* const input_event = event.to_input();
+    auto const input_type = input_event->input_type();
+    if (input_type == mir_input_event_type_key)
+    {
+        auto* const key_event = input_event->to_keyboard();
+        auto const scan_code = key_event->scan_code();
+        if (!is_modifier(scan_code))
+        {
+            return self->process_event(
```
tarek-y-ismail: A better name could be `dispatch_all_pending`

src/server/shell/basic_sticky_keys_transformer.cpp
```diff
@@ -0,0 +1,184 @@
...
+    }
+    else if (input_type == mir_input_event_type_pointer)
+    {
+        auto const pointer_event = input_event->to_pointer();
+        return self->process_event(
+            event,
+            pointer_event->action() == mir_pointer_action_button_up,
+            dispatcher);
+    }
+    else if (input_type == mir_input_event_type_touch)
```
tarek-y-ismail, line 163: Are sticky keys usually used with touch devices?

tests/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,128 @@
...
+namespace mi = mir::input;
+using namespace testing;
+
+namespace
+{
+void click(mi::InputSink* sink, mi::EventBuilder* builder, unsigned int keysym, unsigned int scancode)
+{
+    sink->handle_input(builder->key_event(std::nullopt, mir_keyboard_action_down, keysym, scancode));
+    sink->handle_input(builder->key_event(std::nullopt, mir_keyboard_action_up, keysym, scancode));
+}
```
tarek-y-ismail, line 43: Seems familiar, maybe we should pull this and similarly common testing code into a shared header?

tests/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,128 @@
...
+    std::weak_ptr<mi::InputDeviceRegistry> input_device_registry;
+    std::weak_ptr<mi::InputDeviceHub> input_device_hub;
+};
+
+TEST_F(TestStickyKeys, when_enabled_clicking_modifier_results_in_callback)
+{
+    int count = 0;
+    sticky_keys.on_modifier_clicked([&](int32_t)
+    {
+        count++;
```
tarek-y-ismail, line 74: Should probably make sure that that the modifier passed is the same as the one pressed

tests/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,128 @@
...
+    auto const virtual_keyboard = miral::test::add_test_device(
+        input_device_registry.lock(), input_device_hub.lock(), mi::DeviceCapability::keyboard);
+    virtual_keyboard->if_started_then(
+        [&](mi::InputSink* sink, mi::EventBuilder* builder)
+        {
+            click(sink, builder, XKB_KEY_Shift_L, KEY_LEFTSHIFT);
+        });
+
+    EXPECT_THAT(count, Eq(0));
+}
```
tarek-y-ismail, line 105: I know we discussed this in another PR, but I don't really see a reason for this test and the one before it to be two distinct tests.

tests/unit-tests/shell/test_basic_accessibility_manager.cpp
```diff
@@ -338,7 +354,8 @@ enum class TransformerToTest
     MouseKeys,
     SSC,
     HoverClick,
-    SlowKeys
+    SlowKeys,
+    StickyKeys
```
tarek-y-ismail, line 358: Looks like I forgot to add slow keys to the parameter list below, and lead you down the same path :), maybe it's time we pull that list into some variable?

tests/unit-tests/shell/test_sticky_keys_transformer.cpp
```diff
@@ -0,0 +1,209 @@
...
+    EXPECT_THAT(transformer.transform_input_event(dispatch, &event_builder, *event.get()), Eq(false));
+}
+
+class TestStickyKeysTransformerWithModifiers : public TestStickyKeysTransformer, public WithParamInterface<int32_t>
+{};
+
+TEST_P(TestStickyKeysTransformerWithModifiers, modifier_keyboard_up_events_are_consumed)
+{
+    auto const modifier_key = GetParam();
+    auto const down_event = event_builder.key_event(std::nullopt, mir_keyboard_action_down, modifier_key, modifier_key);
```
tarek-y-ismail: Is the xkb keysym and the scancode the same?

include/miral/miral/sticky_keys.h
mattkae: "depressing" is the proper term IMO (although it is quite depressing!)

src/miral/sticky_keys.cpp
mattkae: I like `set_accessibility_manager` a bit manager because it describes exactly what that method is doing for the object.

src/miral/symbols.map
mattkae: The change would be the same. This is what happens when I run the symbol generator script, it some times shuffles things around

src/server/input/input_event_transformer.cpp
mattkae: Yup exactly! And then you'd try to remove it while you're still in its callstack

src/server/shell/basic_accessibility_manager.cpp
mattkae: It's a little spurious, but it doesn't change any functionality, so I'm tempted to not split it out. If it was functionality altering, I definitely would

src/server/shell/basic_sticky_keys_transformer.cpp
mattkae: I find the iterator just as readable personally

src/server/shell/basic_sticky_keys_transformer.cpp
mattkae: I like it because it allows the caller to easily pass their own meaning for an "up" event and achieve the same behavior

src/server/shell/basic_sticky_keys_transformer.cpp
mattkae: We have to dispatch the pending events _after_ the new one because this dispatch will _release_ the modifiers. We want the modifiers to still be held while we're dispatching the event.

mattkae: We need to dispatch many because these events do not yet have their modifiers sent - that happens later in the input chain. These are just raw key input events.

src/server/shell/basic_sticky_keys_transformer.cpp
mattkae: They could be, so we might as well support them. I can imagine someone plugging in a keyboard and then touching their screen

tests/miral/sticky_keys.cpp
mattkae: It is essentially two sets of functionality with two separate code paths. If one of them breaks in the future, it's easier to see that the "disabled callback path broke but not the enabled callback path" so that you know what the required fix might be

tests/unit-tests/shell/test_basic_accessibility_manager.cpp
mattkae: Let's make a ticket for the future: https://github.com/canonical/mir/issues/4083

Out of scope for this PR

tests/unit-tests/shell/test_sticky_keys_transformer.cpp
mattkae: Good point!

tests/miral/sticky_keys.cpp
mattkae: Eh if it happens again, I'll split it out :) It's easy to keep them separated for now

tarek-y-ismail (approved):
I still think the symbol changes are a bit too messy, but otherwise the rest is noice :metal: 

Will leave the honor of merging to you.

src/miral/symbols.map
```diff
@@ -571,8 +571,11 @@ MIRAL_5.5 {
 global:
   extern "C++" {
     "miral::CursorScale::CursorScale(miral::live_config::Store&)";
+    "miral::CursorScale::CursorScale(miral::live_config::Store&)";
+    "miral::InputConfiguration::InputConfiguration(miral::live_config::Store&)";
     "miral::InputConfiguration::InputConfiguration(miral::live_config::Store&)";
     "miral::OutputFilter::OutputFilter(miral::live_config::Store&)";
+    "miral::OutputFilter::OutputFilter(miral::live_config::Store&)";
```
AlanGriffiths: Why are we duplicating these entries? Or rather, we have duplicates that are now sorted together. But let's dedup as we're changing this

src/server/shell/basic_sticky_keys_transformer.cpp
AlanGriffiths: You could also use a `switch` letting the compiler know the intention and optimise accordingly

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,185 @@
...
+        {"sticky_keys", "enable"},
+        "Whether or not sticky keys are enabled",
+        [this](live_config::Key const&, std::optional<bool> val)
+        {
+            if (val)
+            {
+                if (*val)
+                    this->enable();
+                else
+                    this->disable();
```
AlanGriffiths: Do we need the `this->` wart?

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,185 @@
...
+{
+    return StickyKeys{std::make_shared<Self>(false)};
+}
+
+miral::StickyKeys::StickyKeys(live_config::Store& config_store)
+{
+    config_store.add_bool_attribute(
+        {"sticky_keys", "enable"},
+        "Whether or not sticky keys are enabled",
+        [this](live_config::Key const&, std::optional<bool> val)
```
AlanGriffiths: Would it be safer to capture `self`? `StickyKeys` is copyable and we may end up with a different instance surviving and a "dead" this pointer.

src/server/input/input_event_transformer.cpp
```diff
@@ -116,28 +116,34 @@ mir::input::InputEventTransformer::~InputEventTransformer() = default;
...
+    std::vector<std::shared_ptr<Transformer>> transformers_cache;
+    transformers_cache.reserve(input_transformers.size());
     {
-        if (it->expired())
-        {
-            it = input_transformers.erase(it);
-        }
-        else
+        std::lock_guard lock{mutex};
```
AlanGriffiths: Looks like we're accessing `input_transformers.size()` and then guarding access to `input_transformers`

AlanGriffiths (changes requested):
Not really understood the implementation yet, but some thoughts...

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    config_store.add_bool_attribute(
+        {"sticky_keys", "disable_if_two_keys_are_pressed_together"},
+        "When set to true, clicking two modifier keys are once will result "
+        "in sticky keys being temporarily disabled until all keys are released",
+        [self=self](live_config::Key const&, std::optional<bool> val)
+        {
+            if (val)
+                self->should_disable_if_two_keys_are_pressed_together(*val);
+        });
```
AlanGriffiths: Shouldn't we provide default values?
```suggestion
    config_store.add_bool_attribute(
        {"sticky_keys", "enable"},
        "Whether or not sticky keys are enabled",
        false,
        [self=self](live_config::Key const&, std::optional<bool> val)
        {
            if (val)
            {
                if (*val)
                    self->enable();
                else
                    self->disable();
            }
        });

    config_store.add_bool_attribute(
        {"sticky_keys", "disable_if_two_keys_are_pressed_together"},
        "When set to true, clicking two modifier keys are once will result "
        "in sticky keys being temporarily disabled until all keys are released",
        false,
        [self=self](live_config::Key const&, std::optional<bool> val)
        {
            if (val)
                self->should_disable_if_two_keys_are_pressed_together(*val);
        });
```

src/miral/sticky_keys.cpp
```diff
@@ -0,0 +1,185 @@
...
+    return StickyKeys{std::make_shared<Self>(true)};
+}
+
+auto miral::StickyKeys::disabled() -> StickyKeys
+{
+    return StickyKeys{std::make_shared<Self>(false)};
+}
+
+miral::StickyKeys::StickyKeys(live_config::Store& config_store)
+{
```
AlanGriffiths: I don't think you've run this code

```suggestion
miral::StickyKeys::StickyKeys(live_config::Store& config_store) :   StickyKeys{std::make_shared<Self>(false)}
{
```

AlanGriffiths (approved):
Looks good  (do you want @tarek-y-ismail to take another look?)

PR #4077, opened 2025-07-11
Add documentation about the input platform

jhenstridge:
This adds some overview documentation about how an input platform module integrates into the server.

### DIFF
--- a/doc/sphinx/contributing/explanation/input_platform.md
+++ b/doc/sphinx/contributing/explanation/input_platform.md
@@ -10,15 +10,19 @@ following entrypoints using C linkage (i.e. no symbol name mangling):
 mir::ModuleProperties const* describe_input_module();
 ```
 
-Returns a structure describing the module.
+Describes the module: the name, version, and shared object file that
+implements it.
+
 
 ```c++
 void add_input_platform_options(boost::program_options::options_description& config);
 ```
 
-Add command line options specific to this platform. This seems like it
-should be called from
-`src/platform/options/default_configuration.cpp`, but isn't.
+Adds command line options specific to this platform to the provided
+options description object.
+
+This is not currently called, as detailed in
+[mir #4109](https://github.com/canonical/mir/issues/4109).
 
 
 ```c++
@@ -27,9 +31,10 @@ mir::input::PlatformPriority probe_input_platform(
     mir::ConsoleServices& console);
 ```
 
-Makes a decision on whether this input platform module should be
-used. The return value can range from saying the module is unsupported
-all the way to being the best choice for the system.
+Report how suitable this input platform module for to the runtime
+environment.  The return value can range from saying the module is
+unsupported all the way to being the best choice for the system.
+
 
 ```c++
 mir::UniqueModulePtr<mir::input::Platform> create_input_platform(


AlanGriffiths:
Hurrah for the spellchecker! :wink:

doc/sphinx/reference/input_platform.md
```diff
@@ -0,0 +1,106 @@
...
+{class}`mir::input::InputDevice` subclasses. The subclass must implement the
+following methods:
+
+* `start` and `stop` to start/stop consuming events from the device.
+* `get_device_info` to retrieve an {class}`mir::input::InputDeviceInfo` struct describing the device.
+* `get_pointer_settings`, `get_touchpad_settings`, `get_touchscreen_settings`, and `apply_settings` to manage settings of the device.
+
+If reading events from the input device requires the use of the server
+event loop, this should be achieved through the platform's
+`dispatchable` method.
```
mattkae: Why would we want this specifically? So that there is a "polling" type of behavior?

doc/sphinx/reference/input_platform.md
```diff
@@ -0,0 +1,106 @@
...
+
+The `EventBuilder` class includes methods for constructing various
+types of input events (keyboard events, mouse pointer events,
+touchscreen events, etc). This gives the consumer a way to customise
+how the events are represented.
+
+In addition to sending events to the `InputSink`, the device is also
+expected to send a more raw version of the event data to the
+platform's {class}`InputReport <mir::input::InputReport>` object to
+enable tracing.
```
mattkae: Newline at end please :)

mattkae (approved):
I have two non-blocking comments, but this is more-or-less my understanding of the input platform as well

doc/sphinx/reference/input_platform.md
jhenstridge: I had assumed that some of the other code might rely on certain callbacks happening in the main event loop thread, but Chris says this is not the case.

I had also thought this was something that would be needed to hook up Rust async code, but on further investigation that's probably not the case: we could definitely write async code that could hook to the Mir event loop with this interface, but code using e.g. tokio is always going to run through the tokio runtime's event loop.

I'll reword this to indicate that it is available if needed.

doc/sphinx/reference/input_platform.md
```diff
@@ -0,0 +1,106 @@
...
+## Introduction
+
+The Mir input platform is provided by a loadable module implementing the
+following entrypoints using C linkage (i.e. no symbol name mangling):
+
+```c++
+mir::ModuleProperties const* describe_input_module();
+```
+
+Returns a structure describing the module.
```
keirthana: If I could make a small style suggestion:

Some statements (e.g., “Returns a structure describing the module”) are a bit terse. Would you mind rephrasing these into complete sentences (e.g., “This function returns a structure that describes the module”)? It improves clarity and consistency for readers. Thanks!

doc/sphinx/reference/input_platform.md
```diff
@@ -0,0 +1,106 @@
...
+
+```c++
+void add_input_platform_options(boost::program_options::options_description& config);
+```
+
+Add command line options specific to this platform. This seems like it
+should be called from
+`src/platform/options/default_configuration.cpp`, but isn't.
```
keirthana: "This seems like it should be called from `src/platform/options/default_configuration.cpp`, but isn't."

It may be just me but I don't understand what information are we conveying to the users with this observation. Are we asking them to ignore the discrepancy or stating that it is expected behaviour?

AlanGriffiths: I think the statement is plain wrong: Why would adding options specific to a platform (that might not be in the Mir codebase) be done in the default configuration?

It seems to me that it should be called it when loading the module.

doc/sphinx/reference/input_platform.md
AlanGriffiths: I don't think either the original or the suggested rewording helps the reader: they just repeat what the code says.
```suggestion
Describes the module: the name, version and shared object file that implements it.
```

doc/sphinx/reference/input_platform.md
AlanGriffiths: Ah! This is actually a bug report, not documentation: #4109

doc/sphinx/reference/input_platform.md
```diff
@@ -0,0 +1,106 @@
...
+
+```c++
+mir::input::PlatformPriority probe_input_platform(
+    mir::options::Option const& options,
+    mir::ConsoleServices& console);
+```
+
+Makes a decision on whether this input platform module should be
+used. The return value can range from saying the module is unsupported
+all the way to being the best choice for the system.
```
AlanGriffiths: I don't think this "makes a decision" - that is ultimately done by the caller

```suggestion
Report how suitable this input platform module for to the runtime environment.
The return value can range from saying the module is unsupported
all the way to being the best choice for the system.
```

AlanGriffiths (changes requested):

AlanGriffiths:

PR #4080, opened 2025-07-15
(#4078) package mir-platform-graphics-stub and mir-platform-input-stub

### LINKED ISSUE #4078: We do not install the `mir:stub-graphics` in any package on debian, so we cannot use our fancy new features for testing the whole policy
## Problem
- I want to use `WindowManagementTestHarness` in miracle
- However, when I run a test, I cannot find the `stub-graphics` platform
- It seems that this platform is not packages for Ubuntu

mattkae:
fixes #4078 

## What's new?
- Introduce `mir-platform-graphics-stub` and `mir-platform-input-stub`
- Remove library install from `mir-wlcs-integration`

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -411,6 +411,8 @@ Section: libs
 Architecture: linux-any
 Multi-Arch: same
 Pre-Depends: ${misc:Pre-Depends}
+Breaks: mir-wlcs-integration (<< 2.21.1+dev518~)
+Replaces: mir-wlcs-integration (<< 2.21.1+dev518~)
 Depends: ${misc:Depends},
          ${shlibs:Depends},
 Description: Display server for Ubuntu - stub driver metapackage
@@ -423,6 +425,8 @@ Section: libs
 Architecture: linux-any
 Multi-Arch: same
 Pre-Depends: ${misc:Pre-Depends}
+Breaks: mir-wlcs-integration (<< 2.21.1+dev518~)
+Replaces: mir-wlcs-integration (<< 2.21.1+dev518~)
 Depends: ${misc:Depends},
          ${shlibs:Depends},
 Description: Display server for Ubuntu - stub input platform library


AlanGriffiths (changes requested):
As with other graphics platforms, we should version the stub platform

I guess we "got away with it" because of the limited deployment

mattkae:
> Wonder if we should just put it in `mir-test-tools`?
> 
> Or at least put it in a single `mir-platforms-stub`?
> 
> We've too many packages already :P

Agreed, but I didn't want to break from the existing platforms packaging scheme that we have going :) Also, they could potentially have separate updates, so separate versioning

debian/control
```diff
@@ -432,6 +462,18 @@ Description: Display server for Ubuntu - input platform library
...
+Section: libs
+Architecture: linux-any
+Multi-Arch: same
+Pre-Depends: ${misc:Pre-Depends}
+Depends: mir-platform-input-stub10
+Description: Display server for Ubuntu - stub input platform library
+ Mir is a display server running on linux systems, with a focus on efficiency,
+ robust operation and a well-defined driver model.
+ .
+ This package depends on nothing.
```
AlanGriffiths: Do we need this level of indirection? Can't `mir-test-tools` and `mirtest-dev` depend directly on the binaries?

AlanGriffiths (approved):
OK by me but could, perhaps, be simplified a bit

debian/control
mattkae: What do you mean by this? Are you asking "why do we need this package at all?"

AlanGriffiths: Yes

mattkae: My feeling is that the WLCS integration relies on this, but not necessarily mir-test-tools and mirtest-dev, so it makes sense to keep this separated

AlanGriffiths: I'm confused now.

You've coded:

```
mirtest-dev -> mir-platform-graphics-stub -> mir-platform-graphics-stub23
mir-test-tools -> mir-platform-graphics-stub -> mir-platform-graphics-stub23

mirtest-dev -> mir-platform-input-stub -> mir-platform-input-stub10
mir-test-tools -> mir-platform-input-stub -> mir-platform-input-stub10
```

But you're saying "not necessarily mir-test-tools and mirtest-dev"

Meanwhile, `miral_wlcs_integration` _doesn't_ depend on `mir-platform-graphics-stub`, but you're saying "WLCS integration relies on this".

Either way, isn't this simpler and sufficient:

```
mirtest-dev -> mir-platform-graphics-stub23
mir-test-tools -> mir-platform-graphics-stub23

mirtest-dev ->  mir-platform-input-stub10
mir-test-tools -> mir-platform-input-stub10
```

And, if wlcs integration does need it:,

```
miral_wlcs_integration -> mir-platform-graphics-stub23

miral_wlcs_integration -> mir-platform-input-stub10
```

I don't see anyone outside the project using the proposed `mir-platform-graphics-stub` directly

debian/control
```diff
@@ -402,6 +406,32 @@ Description: Display server for Ubuntu - virtual display platform
...
+Architecture: linux-any
+Multi-Arch: same
+Pre-Depends: ${misc:Pre-Depends}
+Depends: ${misc:Depends},
+         ${shlibs:Depends},
+Description: Display server for Ubuntu - stub driver metapackage
+ Mir is a display server running on linux systems, with a focus on efficiency,
+ robust operation and a well-defined driver model.
+ .
+ This package depends on nothing.
```
RAOF: This is a very weird package description. Perhaps something better:

> Contains stub graphics platform libraries, useful for testing
> Mir or Mir-using compositors when real hardware interactions
> are undesirable.

debian/control
```diff
@@ -402,6 +406,32 @@ Description: Display server for Ubuntu - virtual display platform
...
+Architecture: linux-any
+Multi-Arch: same
+Pre-Depends: ${misc:Pre-Depends}
+Depends: ${misc:Depends},
+         ${shlibs:Depends},
+Description: Display server for Ubuntu - stub input platform library
+ Mir is a display server running on linux systems, with a focus on efficiency,
+ robust operation and a well-defined driver model.
+ .
+ This package depends on nothing.
```
RAOF: Likewise, “contains stub input platform libraries…”

debian/control
RAOF: Yeah, we probably don't need the metapackage here. At most, we'd want `mir-platforms-stub` depending on both the stub graphics and stub input platform, but I think we can just depend directly on the versioned packages from `mirtest-dev` / `mir-wlcs-integration`.

debian/control
```diff
@@ -402,6 +406,32 @@ Description: Display server for Ubuntu - virtual display platform
  Contains the shared libraries required for the Mir server to provide virtual
  output support.
 
+Package: mir-platform-graphics-stub23
+Section: libs
+Architecture: linux-any
+Multi-Arch: same
+Pre-Depends: ${misc:Pre-Depends}
+Depends: ${misc:Depends},
+         ${shlibs:Depends},
```
robert-ancell, line 417: You need to add `Breaks: mir-wlcs-integration (<< NEW_PACKAGE_VERSION)` and `Replaces: mir-wlcs-integration (<< NEW_PACKAGE_VERSION)`, see case 7 in https://wiki.debian.org/PackageTransition

This will require a new entry in `debian/changelog` to add the new version.

mattkae: @RAOF Will this be weird for me to change the version to 2.22.0 in the changelog _before_ we do a release?

Saviq: It will be worse than weird. We'd burn 2.22, because you'd end up with `2.22+dev…`, which is higher than 2.22, and with debs we can't go back.

robert-ancell: You could have 2.22~dev1, i.e. just before the 2.22 release.

Saviq: Yeah, it's just our automation does `+dev…` based on the most recent tag…

robert-ancell (approved):
LGTM

PR #4081, opened 2025-07-15
Fix miral versioning

AlanGriffiths:
The 2.21 release was delayed and restarted which led to the accidental inclusion of MirAL APIs labeled both "5.3" and "5.4". That led to confusion as it was unclear whether 2.21 was MirAL "5.3" or "5.4":

1. Some APIs labeled "5.4" remained on `main` without being released in 2.21
2. New APIs were being labeled "5.4"
3. Some "5.4" APIs were "deprecated" despite never being released

This fixes main to label all new MirAL APIs as "5.5"

It also drops APIs that have been deprecated without first being released


AlanGriffiths:
Oops! The symbols need fixing too. BBIAB...

tarek-y-ismail (approved):
Looks OK, but let's wait for another opinion in case I missed anything. Feel free to merge if you're confident enough

Saviq (approved):
Looks legit!

PR #4085, opened 2025-07-16
libmircore-dev needs libmircore2

AlanGriffiths:


debian/control
```diff
@@ -98,7 +98,7 @@ Section: libdevel
 Architecture: linux-any
 Multi-Arch: same
 Pre-Depends: ${misc:Pre-Depends}
-Depends: ${misc:Depends},
+Depends: libmircore2 (= ${libmircore2:Version}),
```
Saviq:
```suggestion
Depends: libmircore2 (= ${binary:Version}),
```

debian/control
```diff
@@ -98,7 +98,7 @@ Section: libdevel
 Architecture: linux-any
 Multi-Arch: same
 Pre-Depends: ${misc:Pre-Depends}
-Depends: ${misc:Depends},
```
Saviq, line 101: But don't we want to keep `misc`?

PR #4096, opened 2025-07-22
(#4086) Using the output transform in the BasicScreenshooter and no longer sending the Y inversion flag to wlr screen copy clients

### LINKED ISSUE #4086: `BasicScreenshooter` does not take the output transformation into account
## Repro
1. Run mir with a left or right oriented output 
2. Run `grim` and take a screenshot of the output
3. Note that the resulting image is _odd_. The buffer is the correct dimensions, but the screen contents are rotated inside of the center of that buffer.

## What's happening?
Let's say the screen as a real with of 500x1000. In this case, we will tell the screen capture client that they will have a buffer of 500x1000. It is up to the client to then rotate that buffer when they render it. _This part is correct!_

However, when we go to render to that buffer in the `BasicScreenshooter`, the renderer expects that there will be some sort of output transform, or else it will accidentally try to render content made for the 1000x500 buffer into the center of our buffer (because the renderer adds black bars when the sizing doesn't make sense). We _should_ be providing the same output transform as the display sink here, and then the screen shot would make sense!

## Open questions
- How would this work with a region that straddles two outputs: one that is on a rotated output and one that is on a non rotated output? My feeling is that we will have to apply a different screenshooter for each output and then stitch the final image together.

mattkae:
fixes #4086 

## What's new?
- Sending the output transform the `ScreenShooter::capture`
- Fixing erroneous `OutputManager::from_output_transform` code
- Removing `send_flags_event(Flags::y_invert);` because the renderer inverts the image for us
- Fixing broken tests


src/include/server/mir/compositor/screen_shooter.h
```diff
@@ -43,11 +44,17 @@ class ScreenShooter
     ScreenShooter() = default;
     virtual ~ScreenShooter() = default;
 
-    /// The callback may be called on a different thread. It will be given the timestamp of the capture if it succeeds
-    /// or nullopt if there was an error.
+    /// Capture a rectangle on the screen given by [area] into [buffer]. The resulting
+    /// render will have the [transform] applied to it. It is recommended that the
+    /// [area] is an area on a single output with that output's correspondign [transform],
```
robert-ancell: correspondign -> corresponding

src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -520,8 +520,8 @@ void mf::WlrScreencopyFrameV1::report_result(
 {
     if (captured_time)
     {
-        send_flags_event(Flags::y_invert);
-
+        // (mattkae): Note that we never send the inverted flag. The reason for this
```
robert-ancell: What's the purpose of the mattkae on the note? This seems like a permanent comment and the author can be found with 'git annotate' if necessary.

robert-ancell (approved):
LGTM

PR #4099, opened 2025-07-24
graphics: introduce EDID parsing in `DisplayConfigurationOutput`

Saviq:


src/platform/graphics/display_configuration.cpp
```diff
@@ -251,11 +251,79 @@ mir::geometry::Rectangle extents_of(
...
-    return custom_logical_size.is_set() ?
-           mir::geometry::Rectangle(top_left, custom_logical_size.value()) :
-           extents_of(modes, current_mode_index, orientation, top_left, scale);
+    return self().custom_logical_size.is_set() ?
+           mir::geometry::Rectangle(self().top_left, self().custom_logical_size.value()) :
+           extents_of(self().modes, self().current_mode_index, self().orientation, self().top_left, self().scale);
+}
+
+template<typename OutputType>
+bool mg::DisplayConfigurationOutputCommon<OutputType>::parse_edid()
```
mattkae: Minor suggestion: While this template subclass business makes sense to me, it might be too cute. Perhaps it would be more straightforward to:

1. Define the same methods on `DisplayConfigurationOutput` and `UserDisplayConfigurationOutput`
2. Have the implementations in both methods call into templated methods inside of an anonymous namespace, passing `*this`

This way, you'd avoid adding `DisplayConfigurationOutputCommon` altogether and you'll just be augmenting the existing classes with new methods. WDYT?

mattkae (commented):
A light suggestion, but I could be missing something :)

src/platform/graphics/display_configuration.cpp
Saviq: The thing that would annoy me is that we'd need to have the cached `di_info` on both… and then have the boilerplate forwarding methods…

I couldn't even move `edid` to the base class 'cause it's `const` in one, and mutable on the other…

The only annoyance I have with as it is now, really, is the need for explicit template instantiation… FWIW it's not like I came up with this [1] ;)

But I won't push, and maybe there's another pattern that's more in place here?…

[1] https://en.cppreference.com/w/cpp/language/crtp.html

mattkae: Hmmm yes that's a weird problem! There's probably another way, but I doubt that it's cleaner...

include/platform/mir/graphics/display_configuration.h
```diff
@@ -27,6 +27,12 @@
 
 #include <glm/glm.hpp>
 
+extern "C"
+{
+#include <libdisplay-info/edid.h>
+#include <libdisplay-info/info.h>
+}
+
```
AlanGriffiths: This looks like detail that should be confined to the implementation?

AFAICS the only use in the header is for the `di_info` type - which could simply be declared.

include/platform/mir/graphics/display_configuration.h
```diff
@@ -91,10 +97,33 @@ struct DisplayConfigurationMode
...
+private:
+    DisplayConfigurationOutputCommon() = default;
+    friend OutputType;
+
+    OutputType const self() const { return static_cast<OutputType const&>(*this); }
+    OutputType self() { return static_cast<OutputType&>(*this); }
+
+    std::shared_ptr<di_info const> display_info{nullptr};
+};
+
```
AlanGriffiths: It seems very odd to use a template class in a header with member function definitions available for only one translation unit.

Is there no design that keeps this all out of the header? E.g. a templated helper function in the implementation file?

include/platform/mir/graphics/display_configuration.h
```diff
@@ -164,13 +190,64 @@ struct DisplayConfigurationOutput
...
+    {
+    }
 };
 
 /**
  * Mirror of a DisplayConfigurationOutput, with some fields limited to
  * being read-only, preventing users from changing things they shouldn't.
  */
-struct UserDisplayConfigurationOutput
+struct UserDisplayConfigurationOutput : DisplayConfigurationOutputCommon<UserDisplayConfigurationOutput>
```
AlanGriffiths: Does `UserDisplayConfigurationOutput` need the parsing helper? Shouldn't it just forward any accessors to the underlying `DisplayConfigurationOutput`

include/platform/mir/graphics/display_configuration.h
```diff
@@ -91,10 +97,33 @@ struct DisplayConfigurationMode
...
+    OutputType self() { return static_cast<OutputType&>(*this); }
+
+    std::shared_ptr<di_info const> display_info{nullptr};
+};
+
 /**
  * Configuration information for a display output.
  */
-struct DisplayConfigurationOutput
+struct DisplayConfigurationOutput : public DisplayConfigurationOutputCommon<DisplayConfigurationOutput>
```
AlanGriffiths: Mixing inheritance with structs is going to confuse readers. Can you really claim that a `DisplayConfigurationOutput` really IsA `DisplayConfigurationOutputCommon<DisplayConfigurationOutput>`

include/platform/mir/graphics/display_configuration.h
```diff
@@ -91,10 +97,33 @@ struct DisplayConfigurationMode
...
+    auto model() const -> std::optional<std::string>;
+    auto serial() const -> std::optional<std::string>;
+    auto product_code() const -> std::optional<uint16_t>;
+
+private:
+    DisplayConfigurationOutputCommon() = default;
+    friend OutputType;
+
+    OutputType const self() const { return static_cast<OutputType const&>(*this); }
+    OutputType self() { return static_cast<OutputType&>(*this); }
```
AlanGriffiths: This is dangerous:

```c++
    DisplayConfigurationOutputCommon<DisplayConfigurationOutput> slice = DisplayConfigurationOutput{};
    slice.parse_edid();
```

I agree that it is unlikely that someone writes such a mouthful but why tempt fate?

(Yes, we could make that mostly illegal using private inheritance, but I'm unconvinced inheritance is the right solution)

src/platform/graphics/display_configuration.cpp
AlanGriffiths: I think the right solution is to introduce a non-template EDID info member on `DisplayConfigurationOutput` whose constructor takes a `DisplayConfigurationOutput const&`.

include/platform/mir/graphics/display_configuration.h
```diff
@@ -150,9 +179,6 @@ struct DisplayConfigurationOutput
     /// Custom attributes (typically set via the .display configuration file
     std::map<std::string, std::optional<std::string>> custom_attribute = {};
 
-    /** The logical rectangle occupied by the output, based on its position,
-        current mode and orientation (rotation) */
-    geometry::Rectangle extents() const;
```
AlanGriffiths: Why is this function migrating to `DisplayConfigurationOutputCommon`?

AlanGriffiths (changes requested):
I think this implementation introduces needless complexity: There is no need for a templated implementation as the EDID parsing should only be happening in one place (`UserDisplayConfigurationOutput` is an adaptor for `DisplayConfigurationOutput`).

Further, I think all the complexity can be moved to the implementation file.

I'm not clear whether the attributes (vendor,  model, serial, product_code) need to be queried dynamically (which would require a new object to supply them), or whether they could simply be extracted in the `DisplayConfigurationOutput` constructor and store as fields on the struct.

include/platform/mir/graphics/display_configuration.h
Saviq: Because it is common between `{,User}DisplayConfigurationOutput`, so I took the opportunity…

src/platform/graphics/display_configuration.cpp
AlanGriffiths: > I think the right solution is to introduce a non-template EDID info member on `DisplayConfigurationOutput` whose constructor takes a `DisplayConfigurationOutput const&`.

On further examination there's really no need for the `DisplayConfigurationOutput const&` passing: the only uses of `self()` are to access the edid field and for the unrelated "cleanup"  of `extents()`.

So, an object like this is all that is needed:

```
struct DisplayInfo
{
    explicit DisplayInfo(std::vector<uint8_t> const& edid);
    DisplayInfo& operator=(std::vector<uint8_t> const& edid);

    std::optional<std::string> vendor;
    std::optional<std::string> model;
    std::optional<std::string> serial;
    std::optional<uint16_t> product_code;

    // and maybe?
    std::vector<uint8_t> raw_edid;
};
```

This could replace the existing `edid` field

Saviq: > This could replace the existing `edid` field

This would break API, so I would leave it be?

AlanGriffiths: Well, we're breaking ABI anyway and anyone using the edid should probably look at whether we've replaced what they need

Saviq:
> Well, we're breaking ABI anyway and anyone using the edid should probably look at whether we've replaced what they need

Marking `edid` as deprecated proves to be more effort than is worth it… we'd need explicit `DisplayConfigurationOutput` constructors to `#pragma` them to ignore the deprecation… and then still _look_ at it, in case it's populated by whoever is using the deprecated `edid`… I'm leaning towards dropping it after all. Thoughts?

include/platform/mir/graphics/display_configuration.h
```diff
@@ -150,6 +166,9 @@ struct DisplayConfigurationOutput
     /// Custom attributes (typically set via the .display configuration file
     std::map<std::string, std::optional<std::string>> custom_attribute = {};
 
+    /// Display information for this output, if available.
+    DisplayInfo display_info = {};
```
AlanGriffiths:
```suggestion
    DisplayInfo display_info{};
```

src/platform/graphics/display_configuration.cpp
```diff
@@ -321,10 +359,11 @@ mg::UserDisplayConfigurationOutput::UserDisplayConfigurationOutput(
         subpixel_arrangement(main.subpixel_arrangement),
         gamma(main.gamma),
         gamma_supported(main.gamma_supported),
-        edid(main.edid),
+        edid(main.display_info.raw_edid),
```
AlanGriffiths: What is happening with the `edid` members? `DCO::edid` is no longer being set (nor referred to by `UDCO::edid` or the logging, ...)

AlanGriffiths (changes requested):
I think it would be better to remove ``DCO::edid` and `UDCO::edid`

Saviq:
> I think it would be better to remove ``DCO::edid`and`UDCO::edid`

That's it then. Buh-bye.

src/miroil/persist_display_config.cpp
```diff
@@ -246,7 +246,7 @@ void PersistDisplayConfigPolicy::save_config(mg::DisplayConfiguration const& con
         try {
             miroil::DisplayId display_id;
             // FIXME - output.edid should be std::vector<uint8_t>, not std::vector<uint8_t const>
-            display_id.edid.parse_data(reinterpret_cast<std::vector<uint8_t> const&>(output.edid));
+            display_id.edid.parse_data(reinterpret_cast<std::vector<uint8_t> const&>(output.display_info.raw_edid));
```
AlanGriffiths, line 249: And here

src/miral/static_display_config.cpp
```diff
@@ -374,9 +374,9 @@ void miral::YamlFileDisplayConfig::apply_to(mg::DisplayConfiguration& conf)
                 {
                     using mir::graphics::Edid;
                     std::optional<Edid const*> edid;
-                    if (conf_output.edid.size() >= Edid::minimum_size)
+                    if (conf_output.display_info.raw_edid.size() >= Edid::minimum_size)
                     {
-                        edid = reinterpret_cast<Edid const*>(conf_output.edid.data());
+                        edid = reinterpret_cast<Edid const*>(conf_output.display_info.raw_edid.data());
                     }
```
AlanGriffiths, line 380: here

AlanGriffiths (approved):
There is going to be a follow-up that uses `display_info.XXX()` instead of parsing the `raw_edid` again?

Saviq:
> There is going to be a follow-up that uses `display_info.XXX()` instead of parsing the `raw_edid` again?

Exactly, I have it lined up, but felt like splitting this one out was the right thing.

PR #4102, opened 2025-07-24
Hover click cleanup

tarek-y-ismail:
### What's new
1. Removes the old `HoverClick(bool)` constrcutor, replacing it with `enabled()` and `disabled()`
2. Adds a live config constructor, which adds the following options:
    - hover_click_enable
    - hover_click_hover_duration
    - hover_click_cancel_displacement
    - hover_click_reclick_displacement
3. Adds a comment that explains to users what hover click is
4. Fixed a bug where hover clicks would continuously be initiated regardless of whether the pointer moved `hover_click_reclick_displacement` pixels away or not

### How to test
You can add the following options `~/.config/mir_demo_server.live-config`:
```
hover_click_enable=true
hover_click_hover_duration=3000
hover_click_cancel_displacement=100
hover_click_reclick_displacement=100
```

Hover click does not start dispatching clicks until the cursor first moves. After that, if the cursor stays still  for the hover duration, a click is dispatched. The cursor can additionally move slightly once the click is initiated without cancelling the click. The amount of movement to cancel the click can be customized through `hover_click_cancel_displacement`. This displacement is computed as the distance between where the cursor was when the hover click was initiated, and the current position of the cursor. If a click is dispatched through hover click, no more clicks will be initiated unless the cursor moves from the last click position. The distance in pixels can be customized through `hover_click_reclick_displacement`. This displacement is computed in the same way as the cancel displacement.

### DIFF
--- a/include/miral/miral/hover_click.h
+++ b/include/miral/miral/hover_click.h
@@ -38,6 +38,11 @@ class HoverClick
 {
 public:
     explicit HoverClick(bool enabled_by_default);
+    /// Creates a `HoverClick` instance that's enabled by default.
+    auto static enabled() -> HoverClick;
+
+    /// Creates a `HoverClick` instance that's disabled by default.
+    auto static disabled() -> HoverClick;
 
     void operator()(mir::Server& server);
 
@@ -79,6 +84,7 @@ public:
 
 private:
     struct Self;
+    HoverClick(std::shared_ptr<Self> self);
     std::shared_ptr<Self> const self;
 };
 }
--- a/include/miral/miral/hover_click.h
+++ b/include/miral/miral/hover_click.h
@@ -24,7 +24,15 @@ namespace mir { class Server; }
 
 namespace miral
 {
-/// Enable configuring hover click at runtime.
+/// Enables configuring hover click at runtime.
+///
+/// Hover click is an accessibility feature that allows users to dispatch primary
+/// clicks by holding the pointer still for some configurable period of time.
+/// The pointer may move by a configurable distance to accommodate users with
+/// disabilities. After a hover click, no further clicks are dispatched until
+/// the cursor moves by a configurable distance from the position of the
+/// previous click.
+/// 
 /// \remark Since MirAL 5.5
 class HoverClick
 {
--- a/include/miral/miral/hover_click.h
+++ b/include/miral/miral/hover_click.h
@@ -24,6 +24,7 @@ namespace mir { class Server; }
 
 namespace miral
 {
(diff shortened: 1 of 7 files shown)


include/miral/miral/hover_click.h
```diff
@@ -24,12 +24,27 @@ namespace mir { class Server; }
...
 {
-/// Enable configuring hover click at runtime.
+namespace live_config { class Store; }
+/// Enables configuring hover click at runtime.
+///
+/// Hover click is an accessibility feature that allows users to dispatch left
+/// clicks by holding the pointer still for some amount of time. The pointer is
+/// allowed to move slightly to accomodate users with disabilities. After a
+/// hover click, no more clicks are dispatched until the cursor moves slightly
+/// from the position of the last click
```
mattkae:
```suggestion
/// Hover click is an accessibility feature that allows users to dispatch left
/// clicks by holding the pointer still for some configurable period of time. The pointer may
/// move by a configurable distance to accommodate users with disabilities. After a
/// hover click, no further clicks are dispatched until the cursor moves by a
/// configurable distance from the position of the previous click.
```

src/miral/hover_click.cpp
```diff
@@ -50,9 +53,75 @@ struct miral::HoverClick::Self
     std::weak_ptr<mir::shell::AccessibilityManager> accessibility_manager;
 };
 
-miral::HoverClick::HoverClick(bool enabled_by_default)
-    : self{std::make_shared<Self>(enabled_by_default)}
+miral::HoverClick::HoverClick(miral::live_config::Store& config_store)
+    : self{std::make_shared<Self>(false)}
+{
+    config_store.add_bool_attribute(
```
mattkae, line 59: Isn't it recommended to provide the `preset` attribute for these?

tarek-y-ismail: I don't think that's necessary. If the config doesn't have a key-value pair, then the callback will be called with `std::nullopt` and will return without doing anything, preserving whatever value shell authors might've set in code.

include/miral/miral/hover_click.h
```diff
@@ -24,12 +24,28 @@ namespace mir { class Server; }
 
 namespace miral
 {
-/// Enable configuring hover click at runtime.
+namespace live_config { class Store; }
+/// Enables configuring hover click at runtime.
+///
+/// Hover click is an accessibility feature that allows users to dispatch left
```
mattkae:
```suggestion
/// Hover click is an accessibility feature that allows users to dispatch primary
```

PR #4104, opened 2025-07-29
Add slow keys configuration tests

tarek-y-ismail:
Adds a bunch of tests to make sure that changes in miral are relayed to the transformer. The core transformer is already tested so no need to add miral side tests for that. Also pulls out some code in common with mousekeys tests.

### DIFF
--- a/tests/miral/test_slow_keys.cpp
+++ b/tests/miral/test_slow_keys.cpp
@@ -24,16 +24,6 @@ using namespace testing;
 class TestSlowKeys : public miral::TestAccessibilityManager
 {
 public:
-    TestSlowKeys()
-    {
-        ON_CALL(accessibility_manager->slow_keys_transformer, on_key_down(_))
-            .WillByDefault([](auto okd_cb) { okd_cb(0); });
-        ON_CALL(accessibility_manager->slow_keys_transformer, on_key_rejected(_))
-            .WillByDefault([](auto okr_cb) { okr_cb(0); });
-        ON_CALL(accessibility_manager->slow_keys_transformer, on_key_accepted(_))
-            .WillByDefault([](auto oka_cb) { oka_cb(0); });
-    }
-
     miral::SlowKeys slow_keys{miral::SlowKeys::enabled()};
 };
 
@@ -57,6 +47,8 @@ TEST_F(TestSlowKeys, enabling_from_miral_enables_it_in_the_accessibility_manager
 
 TEST_F(TestSlowKeys, setting_on_key_down_sets_transformer_on_key_down)
 {
+    ON_CALL(accessibility_manager->slow_keys_transformer, on_key_down(_)).WillByDefault([](auto okd_cb) { okd_cb(0); });
+
     auto calls = 0;
     auto okd = [&calls](auto){ calls += 1; };
     add_server_init(slow_keys);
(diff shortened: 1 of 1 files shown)


tests/miral/test_slow_keys.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+using namespace testing;
+
+class TestSlowKeys : public miral::TestAccessibilityManager
+{
+public:
+    miral::SlowKeys slow_keys{miral::SlowKeys::enabled()};
+};
+
+TEST_F(TestSlowKeys, enabling_slow_keys_from_miral_enables_it_in_the_accessibility_manager)
```
mattkae: I don't think you need the `slow_keys` word in the test name for this test or any of the others. It's already namespaces by `TestSlowKeys`

tests/miral/test_slow_keys.cpp
```diff
@@ -0,0 +1,134 @@
...
+    slow_keys.enable();
+    slow_keys.disable();
+}
+
+TEST_F(TestSlowKeys, slow_keys_setting_on_key_down_sets_transformer_on_key_down)
+{
+    auto calls = 0;
+    auto on_key_down = [&calls](auto) mutable
+    {
+        calls += 1;
```
mattkae: Can you explain to me why this is being called? I thought that we were just _setting_ the callback. Why is it also _calling_ the callback?

tarek-y-ismail: I wanted to make sure that the callback that we passed in was the one that was handed down to the transformer. Since the move the callback into the transformer, I didn't see a straightforward way of doing this. I added the following lines to `MockSlowKeysTransformer`:
```cpp
    MockSlowKeysTransformer()
    {
        using testing::_;
        ON_CALL(*this, on_key_down(_)).WillByDefault([](auto okd) { okd(0); });
        ON_CALL(*this, on_key_rejected(_)).WillByDefault([](auto okr) { okr(0); });
        ON_CALL(*this, on_key_accepted(_)).WillByDefault([](auto oka) { oka(0); });
    }
```

So, whenever the transformer gets handed a callback, we just call it. By observing the side effects, we can make sure that the callback we handed down was the one passed to the transformer. 

I know it's a bit convoluted, and I'm happy to change it if you have something simpler in mind.

mattkae: That is definitely unexpected. Why isn't it enough to just register that `on_key_down` gets call `X` many times?

tests/miral/test_slow_keys.cpp
```diff
@@ -0,0 +1,142 @@
...
+{
+public:
+    TestSlowKeys()
+    {
+        ON_CALL(accessibility_manager->slow_keys_transformer, on_key_down(_))
+            .WillByDefault([](auto okd_cb) { okd_cb(0); });
+        ON_CALL(accessibility_manager->slow_keys_transformer, on_key_rejected(_))
+            .WillByDefault([](auto okr_cb) { okr_cb(0); });
+        ON_CALL(accessibility_manager->slow_keys_transformer, on_key_accepted(_))
+            .WillByDefault([](auto oka_cb) { oka_cb(0); });
```
mattkae: These should be localized to their respective tests in my opinion

mattkae (approved):
LGTM :+1:

PR #4110, opened 2025-08-01
many: consolidate EDID parsing

Saviq:
This will let Lomiri build (possibly with a workaround for the `-Wpedantic`), but it will change the name of the cached display settings.

Alternative is to use `libdisplay-info` directly to try and replicate the previous behaviour.


src/miroil/CMakeLists.txt
```diff
@@ -12,7 +12,6 @@ add_definitions(-DMIR_LOG_COMPONENT_FALLBACK="miroil" -DMIROIL_ENABLE_DEPRECATIO
 
 add_library(miroil SHARED
     compositor.cpp ${miroil_include}/miroil/compositor.h    
-    edid.cpp ${miroil_include}/miroil/edid.h
```
AlanGriffiths, line 15: Isn't `${miroil_include}/miroil/edid.h` still there?

include/miroil/miroil/edid.h
```diff
@@ -34,31 +34,13 @@ struct Edid
...
-        std::string string_value() const;
+        std::string string_value() const { return {}; };
     };
-    Descriptor descriptors[4];
 
-    Edid& parse_data(std::vector<uint8_t> const&);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
+    Descriptor descriptors[0];
+#pragma GCC diagnostic pop
```
AlanGriffiths: This breaks the miroil ABI - we should bump the .soname

include/miroil/miroil/edid.h
```diff
@@ -34,31 +34,13 @@ struct Edid
...
-
-        Type type{Type::undefined};
-        Value value{{0}};
-
-        std::string string_value() const;
+        std::string string_value() const { return {}; };
     };
-    Descriptor descriptors[4];
 
-    Edid& parse_data(std::vector<uint8_t> const&);
```
AlanGriffiths, line 61: Unused downstream?

include/miroil/miroil/edid.h
```diff
@@ -34,31 +34,13 @@ struct Edid
...
-        };
-
-        union Value {
-            char monitor_name[13];
-            char unspecified_text[13];
-            char serial_number[13];
-        };
-
-        Type type{Type::undefined};
-        Value value{{0}};
```
AlanGriffiths, line 55: Unused downstream?

include/miroil/miroil/edid.h
```diff
@@ -34,31 +34,13 @@ struct Edid
...
-
-        std::string string_value() const;
+        std::string string_value() const { return {}; };
     };
-    Descriptor descriptors[4];
 
-    Edid& parse_data(std::vector<uint8_t> const&);
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpedantic"
+    Descriptor descriptors[0];
```
AlanGriffiths: I get that this allows Lomiri to build - but this could break things at runtime. Have we checked?

AlanGriffiths (changes requested):
Nits

include/miroil/miroil/edid.h
Saviq: Correct, only usage of `miroil::Edid` found here:

https://gitlab.com/ubports/development/core/lomiri/-/blob/9516822a91062dadfce075b4d678cf2ff7484086/src/DisplayConfigurationStorage.cpp#L30-40

include/miroil/miroil/edid.h
Saviq: Ditto.

include/miroil/miroil/edid.h
Saviq: We're already bumping in 2.22: 974b779b8af92ff985693d7bd045eaffb9cbe124

But I did clear the removed symbol from the map and `.symbols` file.

src/miroil/CMakeLists.txt
Saviq: Indeed.

include/miroil/miroil/edid.h
Saviq: You're not wrong…

```cpp
    for (int i = 0; i < 4; i++) {
        str += QString::fromStdString(edid.descriptors[i].string_value());
    }
```

Saviq: So with that, we have the option of keeping the 4 descriptors, empty, or even populated with the values we have in `DisplayInfo`… or modify Lomiri to just stop using them and drop them from `miroil::Edid`…

Saviq: Let's see how they react:

https://gitlab.com/ubports/development/core/lomiri/-/merge_requests/237

AlanGriffiths: :peanuts:

AlanGriffiths: Seems it got merged

Saviq:
> Can we simply lose `descriptors` now, instead of this dubious 0-length array?

Yes! Done.

Issue #3267, opened 2024-03-05
Add support for listening and reconfiguring the keyboard settings from locale1


Conan-Kudo:
### Proposal

Please add support for listening and dynamically configuring keyboard settings based on [`org.freedesktop.locale1`](https://www.freedesktop.org/software/systemd/man/latest/org.freedesktop.locale1.html).

### Use-cases

This is needed in order to handle cases like keyboard layout switching from the login manager (e.g. SDDM) or the installer (e.g. Calamares or Anaconda).

### References

Here are a couple of example implementations for other compositors:

* KWin: https://invent.kde.org/plasma/kwin/-/merge_requests/2921
* Sway: https://github.com/alebastr/sway-systemd/blob/main/src/locale1-xkb-config

Conan-Kudo:
This is going to be needed for resolving [rhbz#2278864](https://bugzilla.redhat.com/show_bug.cgi?id=2278864) for Fedora Miracle WM Spin releasing with Fedora 41.

cc: @mattkae

Conan-Kudo:
This is really needed with Fedora 42 now that LXQt is also leveraging Mir through Miriway.

cc: @AlanGriffiths

AlanGriffiths:
I know it doesn’t handle all cases, but launching miriway-shell via a login shell at least picks up the session settings.

This shouldn’t be hard, but there’s limited bandwidth. (PRs welcome)

AlanGriffiths:
I've done a little investigation:

The Mir support needed is already in place with `miral::Keymap::set_keymap()` - a shell (such as Miriway) would need to add the dbus listener

AlanGriffiths:
After a bit of puzzlement I have concluded (testing on 24.04) that locale changes are not actually notified over dbus.

I can read the settings before and after, but don't get a "PropertiesChanged" notification when I change them.

```
$ dbus-send --system --dest=org.freedesktop.DBus --type=method_call --print-reply /org/freedesktop/DBus org.freedesktop.DBus.ListNames | grep -i -e properties -e locale
$
```

AlanGriffiths:
Hmm. I don't have a native install of fedora right now, but in a LXD container I use for testing I have similar problems. Maybe it is me?

AlanGriffiths:
After a bit of a rabbithole I now have the dbus side understood.

But the question arises how to integrate with the existing configuration...

Currently, `miral::Keymap` defaults to keyboard settings pulled from udev:

```c++
[...]

PR #4111, opened 2025-08-04
Allow tracking keymap from locale1

AlanGriffiths:
This allows the Keymap to track the org.freedesktop.locale1 properties.

There's an example allowing this based on `MIRAL_SHELL_SYSTEM_LOCALE1 being set:

```
MIRAL_SHELL_SYSTEM_LOCALE1=1 miral-app
```

Fixes: #3267


Saviq (commented):
Anything appropriate we can do to make TICS happy?

Maybe a test?

---

So e.g. Miriway would construct with `::system_locale1()` if it doesn't find `keyboard-layout` in the standard options?

AlanGriffiths:
> Maybe a test?

It is difficult to provide meaningful tests of interaction with externals like dbus. They tend to involving more about emulating the external than the functionality under test.

> So e.g. Miriway would construct with `::system_locale1()` if it doesn't find `keyboard-layout` in the standard options?

As in the example here, Miriway will check an environment variable to select behaviour

AlanGriffiths:
An incantation from manual testing:

```bash
$ dbus-send --system --print-reply --dest=org.freedesktop.locale1 \
 /org/freedesktop/locale1 org.freedesktop.locale1.SetX11Keyboard \
 string:gb string:pc105 string: string: boolean:false boolean:false
```

src/miral/keymap.cpp
```diff
@@ -244,3 +245,166 @@ void miral::Keymap::set_keymap(std::string const& keymap)
...
+
+    GError* error = nullptr;
+
+    if (Variant const result{g_dbus_connection_call_sync(connection,
+                                                        bus_name,
+                                                        object_path,
+                                                        properties_interface,
+                                                        method_name,
+                                                        g_variant_new("(ss)", interface_name, entry),
+                                                        nullptr,
```
robert-ancell: Suggest using `G_VARIANT_TYPE("(v)")` here to validate the return type.

src/miral/keymap.cpp
```diff
@@ -244,3 +245,166 @@ void miral::Keymap::set_keymap(std::string const& keymap)
...
+                                                        &error)})
+    {
+        Variant const unwrap{g_variant_get_child_value(result, 0)};
+        Variant const unwrap2{g_variant_get_child_value(unwrap, 0)};
+
+        if (g_variant_is_of_type(unwrap2, G_VARIANT_TYPE_STRING))
+        {
+            gsize len = 0;
+            return g_variant_get_string(unwrap2, &len);
```
robert-ancell: Since len is unused, can set it to NULL.

src/miral/keymap.cpp
```diff
@@ -244,3 +245,166 @@ void miral::Keymap::set_keymap(std::string const& keymap)
...
+char const* const sender = interface_name;
+char const* const object_path = "/org/freedesktop/locale1";
+char const* const properties_interface = "org.freedesktop.DBus.Properties";
+char const* const signal_name = "PropertiesChanged";
+
+auto read_entry(Connection const& connection, char const* entry) -> std::optional<std::string>
+{
+    static char const* const method_name = "Get";
+
+    GError* error = nullptr;
```
robert-ancell: Suggest use `g_autoptr(GError) error = nullptr` to remove the need for the `g_error_free` later.

src/miral/keymap.cpp
```diff
@@ -244,3 +245,166 @@ void miral::Keymap::set_keymap(std::string const& keymap)
...
+        std::shared_ptr<MainLoop> const main_loop;
+
+        void callback(GVariant* parameters)
+        {
+            std::optional<std::string> layout;
+            std::optional<std::string> options;
+            std::optional<std::string> variant;
+
+            const char* _;
+            GVariant* changed_properties;
```
robert-ancell: Can use `g_autoptr(GVariant) changed_properties = nullptr;` to avoid the use of the g_variant_free later (and same for `invalidated_properties`).

src/miral/keymap.cpp
```diff
@@ -244,3 +245,166 @@ void miral::Keymap::set_keymap(std::string const& keymap)
...
+            GVariant* invalidated_properties;
+            g_variant_get(parameters, "(&s@a{sv}@as)",
+                         &_,
```
robert-ancell: NULL can be used instead of `_` to ignore the return value.

robert-ancell: And same for `invalidated_properties`.

src/miral/keymap.cpp
```diff
@@ -244,3 +245,166 @@ void miral::Keymap::set_keymap(std::string const& keymap)
...
+        void callback(GVariant* parameters)
+        {
+            std::optional<std::string> layout;
+            std::optional<std::string> options;
+            std::optional<std::string> variant;
+
+            const char* _;
+            GVariant* changed_properties;
+            GVariant* invalidated_properties;
+            g_variant_get(parameters, "(&s@a{sv}@as)",
```
robert-ancell: Instead of using `@a{sv}` use `a{sv}` and the return type is GVariantIter - you will be able to remove the creation of the iter later. i.e. (including other changes)

```c
g_autoptr(GVariantIter) changed_properties = NULL;
g_variant_get(parameters, "(sa{sv}as)", NULL, &changed_properties, NULL);
const gchar *key = NULL;
GVariant* value = NULL;
while (g_variant_iter_loop(&iter, "{&sv}", &key, &value))
{
   // ...
}
```

src/miral/keymap.cpp
```diff
@@ -244,3 +245,166 @@ void miral::Keymap::set_keymap(std::string const& keymap)
...
+                      error->message, bus_name, object_path, properties_interface, method_name, interface_name);
+        g_error_free(error);
+    }
+
+    return std::nullopt;
+}
+
+auto read_keymap(Connection const& connection) -> std::optional<std::string>
+{
+    if (auto const layout = read_entry(connection, "X11Layout"))
```
robert-ancell: This is three blocking round-trips to the D-Bus server - you could do this in one with GetAll. Perhaps add a comment so it could be improved in the future.

src/miral/keymap.cpp
```diff
@@ -244,3 +245,166 @@ void miral::Keymap::set_keymap(std::string const& keymap)
...
+            while (g_variant_iter_loop(&iter, "{&sv}", &key, &value))
+            {
+                if (g_variant_is_of_type(value, G_VARIANT_TYPE_STRING))
+                {
+                    using namespace std::string_literals;
+
+                    if (key == "X11Layout"s)
+                    {
+                        gsize len = 0;
+                        layout = g_variant_get_string(value, &len);
```
robert-ancell: Can use NULL for all these len cases.

robert-ancell (approved):
Some suggestions for making the DBus code a bit cleaner, but otherwise looks good when TICS passes.

src/miral/gdbus.h
```diff
@@ -0,0 +1,65 @@
...
+
+    static auto the_main_loop() -> std::shared_ptr<MainLoop>;
+
+    ~MainLoop();
+
+private:
+    MainLoop();
+    GMainLoop* const loop;
+    std::jthread const t;
+};
```
RAOF: Do we need our own mainloop? `mir::Server::the_main_loop()` should exist, and provide a default main loop, which is all we're using here?

AlanGriffiths: That's what I thought initially - but then subscribe() didn't result in callbacks. Will investigate further...

AlanGriffiths: ...and I can still reproduce that. Maybe there's some magic that can consolidate these "mainloop"s, but I don't see it (yet) - likely because I don't know why it doesn't "just work"

AlanGriffiths: OK. I think I understand why this doesn't "just work":

`Server::the_main_loop()` uses a bespoke loop calling `g_main_context_iteration(main_context, TRUE);
` on a non-default  `GMainContext`.

The code here subscribes on the default context.

So these two "main loops" are independent. I'm not sure what would be involved in merging them - there do not appear to be any hooks to `mir::MainLoop` that would allow them to be consolidated.

(Roadmap sessions have started, but maybe I'll have an insight before EOD)

AlanGriffiths: And looking at the logind I could see what is needed

AlanGriffiths:
The asan failures are https://github.com/canonical/wlcs/issues/364

The TICS failure is a lack of a test framework for the dbus interaction. (I don't think we need to fix this - the code is simple enough and should be stable)

AlanGriffiths:
@Saviq, the versioning here assumes it will be cherry-picked to 2.22. Ok with landing that?

Saviq:
> @Saviq, the versioning here assumes it will be cherry-picked to 2.22. Ok with landing that?

+1, self-contained enough.

PR #4118, opened 2025-08-08
ci: run cla-check on merge queue, too

Saviq:


Saviq:
This is not the check `merge_group` is looking for… (waves hand)

- https://github.com/canonical/cla-workflow-hosting/pull/3 is

Issue #4120, opened 2025-08-11
Improve accessibility APIs


Saviq:
[rant, unrelated to this PR...]
Using an integer keysym here isn't a developer friendly API. For one thing keycode might be better in some circumstances and for another strongly typed values (not an alias like xkb_keysym_t - which is still better than uint).

Really, I think `MirKeyboardEvent const*` would be better.

_Originally posted by @AlanGriffiths in https://github.com/canonical/mir/pull/4113#discussion_r2266856948_

PR #4121, opened 2025-08-11
Use strong types in APIs where possible

AlanGriffiths:
Fixes: #4120


tarek-y-ismail:
Should probably handle sticky keys and slow keys as well

AlanGriffiths:
> Should probably handle sticky keys and slow keys as well

Added `SlowKeys`, I think `StickyKeys` is good as it is - let the reviewer decide...

mattkae (approved):
Yup that looks better to me :)

PR #4128, opened 2025-08-12
`mir::scene::Observer::surface_exists` now presents surfaces in their focus order rather than by layer

mattkae:
## What's new?
- Added `SurfaceStack::fous_order` to keep track of the global focus order of surfaces. Before this, we were only keeping track of focus order per layer, which is technically incorrect as you could have an application on the "always above" layer who is less focused than application the regular layer
- Modified the protocol to claim that the initial surface list is in the focus order

## What's next?
1. Remove `SurfaceStack::swap_z_order`. I don't see a good reason for this method to exist.
2. Finish the implementation of the application selector app demo and document it
3. Remove the `ApplicationSelector` class entirely
4. Update `ForeignSceneObserver::surface_exists` to only send toplevels (right now it sends everything)


Saviq:
Don't think we should modify the protocol, it's our implementation choice?

wayland-protocols/wlr-foreign-toplevel-management-unstable-v1.xml
```diff
@@ -32,7 +32,9 @@
       letting them request certain actions on them, like maximizing, etc.
 
       After a client binds the zwlr_foreign_toplevel_manager_v1, each opened
-      toplevel window will be sent via the toplevel event
+      toplevel window will be sent via the toplevel event. Toplevels are sent
+      in their focus order, with the most recently focused toplevel sent first
+      and the least recently focused toplevel send last.
```
AlanGriffiths: We don't own this protocol: we can't change it here - the change would have to come from upstream

tests/unit-tests/scene/test_surface_stack.cpp
```diff
@@ -409,8 +409,8 @@ TEST_F(SurfaceStack, scene_observer_informed_of_existing_surfaces)
     MockSceneObserver observer;
 
     InSequence seq;
-    EXPECT_CALL(observer, surface_exists(stub_surface1)).Times(1);
     EXPECT_CALL(observer, surface_exists(stub_surface2)).Times(1);
+    EXPECT_CALL(observer, surface_exists(stub_surface1)).Times(1);
```
AlanGriffiths, line 413: I wonder why we previously _required_ the opposite order in this test. I see nothing in the current codebase that depends on the order.

src/server/scene/surface_stack.cpp
```diff
@@ -541,6 +546,9 @@ void ms::SurfaceStack::insert_surface_at_top_of_depth_layer(std::shared_ptr<Surf
     if (surface_layers.size() <= depth_index)
         surface_layers.resize(depth_index + 1);
     surface_layers[depth_index].push_back(surface);
+
+    std::erase_if(focus_order, [&](auto const& s) { return s.lock() == surface; });
+    focus_order.insert(focus_order.begin(), surface);
```
AlanGriffiths: Always inserting at the beginning of a vector suggests the wrong data structure is in use. (Or that we are using it wrong - consider putting the MRU element at the end)

AlanGriffiths (changes requested):
I have no issue with tracking focus order in `SurfaceStack`, but this proposed implementation will "leak": expired pointers are never purged from the list.

I am concerned by the plan outlined:

>    Modified the protocol to claim that the initial surface list is in the focus order

We cannot modify a protocol we don't own

> If we can guarantee that the initial focus order presented my wlr foreign toplevel management is true, then we don't actually need global shortcuts. 

How then do `Alt-Tab` etc get notified to the application selector?

>  Remove SurfaceStack::swap_z_order. I don't see a good reason for this method to exist.

This supports `miral::WindowManagerTools::swap_tree_order()` - how do you intend to replace it?

>    Remove the ApplicationSelector class entirely

Before this is removed, there needs to be an adequate replacement for the functionality it supports. (Making it possible to implement an external switcher isn't itself a replacement.) That seems a lot of steps from this PR.

mattkae:
> I have no issue with tracking focus order in `SurfaceStack`, but this proposed implementation will "leak": expired pointers are never purged from the list.
> 
> I am concerned by the plan outlined:
> 
> > Modified the protocol to claim that the initial surface list is in the focus order
> 
> We cannot modify a protocol we don't own
> 
> > If we can guarantee that the initial focus order presented my wlr foreign toplevel management is true, then we don't actually need global shortcuts.
> 
> How then do `Alt-Tab` etc get notified to the application selector?
> 
> > Remove SurfaceStack::swap_z_order. I don't see a good reason for this method to exist.
> 
> This supports `miral::WindowManagerTools::swap_tree_order()` - how do you intend to replace it?
> 
> > Remove the ApplicationSelector class entirely
> 
> Before this is removed, there needs to be an adequate replacement for the functionality it supports. (Making it possible to implement an external switcher isn't itself a replacement.) That seems a lot of steps from this PR.

mattkae:
> We cannot modify a protocol we don't own

True enough

> How then do `Alt-Tab` etc get notified to the application selector?

It requires a little bit of infrastructure on the shell's end, but here's how we can do it:

1. The first alt+tab will be intercepted and will open the application switcher
2. When alt+tab is sent again, we check if the pid of the application switcher is valid. If it is, then we don't intercept it and let it propagate to the client.
3. Since the alt+tab client is a fullscreen layer shell client with grabbing focus, then it will definitely receive the alt+tab input

This might be easier with global shortcuts, but it's not too hard without it.


> This supports `miral::WindowManagerTools::swap_tree_order()` - how do you intend to replace it?

This is an unnecessary step. I just wanted to deprecate it as it seems to be a weird leftover. I don't know why shell clients want to swap Z orders of windows. Is there a good reason?

> Before this is removed, there needs to be an adequate replacement for the functionality it supports. (Making it possible to implement an external switcher isn't itself a replacement.) That seems a lot of steps from this PR.

Yeah exactly. We will have to package our own switcher somehow and ship it with miral shell or something else. Or at least ship a headless switcher.

tests/unit-tests/scene/test_surface_stack.cpp
mattkae: This was the guaranteed order, but it didn't _have_ to be guaranteed, you are right there

AlanGriffiths (approved):
OK

PR #4132, opened 2025-08-15
get-wlcs-from-the-ppa

AlanGriffiths:


.github/workflows/symbols-check.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt-get update
+          sudo --preserve-env apt-add-repository --yes ppa:mir-team/dev
```
mattkae, line 27: Should you update after this? Or is that unnecessary?

AlanGriffiths: It used to be necessary, but not for some years

PR #4136, opened 2025-08-18
Consolidate repeating alarm code

tarek-y-ismail:
* Moves the repeating alarm code duplicated between `KeyRepeatDispatcher` and `BasicMouseKeysTransformer` to `time::AlarmFactory::create_repeating_alarm`.
* Adds a new module `mirservertime` at `src/server/time/` to house the implementation of `time::AlarmFactory::create_repeating_alarm`

### DIFF
--- a/src/include/server/mir/time/alarm_factory.h
+++ b/src/include/server/mir/time/alarm_factory.h
@@ -61,6 +61,17 @@ public:
      */
     virtual std::unique_ptr<Alarm> create_alarm(std::unique_ptr<LockableCallback> callback) = 0;
 
+    /**
+     * \brief Create an Alarm that will not fire until scheduled and repeats
+     * itself with a certain delay.
+     *
+     * \param callback Function to call when the Alarm signals
+     * \param repeat_delay The delay between self-invocations
+     * \return A handle to an Alarm that can later be scheduled
+     */
+    auto create_repeating_alarm(std::function<void()> const& callback, std::chrono::milliseconds repeat_delay)
+        -> std::shared_ptr<Alarm>;
+
 protected:
     AlarmFactory() = default;
     AlarmFactory(AlarmFactory const&) = delete;
--- a/src/server/symbols.map
+++ b/src/server/symbols.map
@@ -1625,3 +1625,10 @@ global:
 local: *;
 };
 
+MIR_SERVER_INTERNAL_2.23 {
+global:
+  extern "C++" {
+    mir::time::AlarmFactory::create_repeating_alarm*;
+  };
+local: *;
+} MIR_SERVER_INTERNAL_2.22;
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -29,6 +29,7 @@ add_subdirectory(frontend_wayland/)
 add_subdirectory(frontend_xwayland/)
 add_subdirectory(shell/)
 add_subdirectory(console/)
+add_subdirectory(time/)
 
 configure_file(
   ${CMAKE_CURRENT_SOURCE_DIR}/mirserver.pc.in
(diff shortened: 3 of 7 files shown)


AlanGriffiths:
> #d

What's that?

src/include/server/mir/repeating_alarm.h
```diff
@@ -0,0 +1,39 @@
...
+        std::chrono::milliseconds repeat_delay);
+
+    bool cancel() override;
+    State state() const override;
+    bool reschedule_in(std::chrono::milliseconds delay) override;
+    bool reschedule_for(time::Timestamp timeout) override;
+
+private:
+    std::shared_ptr<std::weak_ptr<time::Alarm>> shared_weak_alarm;
+    std::shared_ptr<time::Alarm> wrapped;
```
AlanGriffiths:
```suggestion
    std::shared_ptr<std::weak_ptr<time::Alarm>> const shared_weak_alarm;
    std::shared_ptr<time::Alarm> const wrapped;
```

tarek-y-ismail:
> What's that?

I either was going to type out `# Description`, went and wrote the description but forgot the header, or I mistyped it (new keyboard feels quite different from my laptop's keyboard)

tarek-y-ismail:
> Have you considered using a factory function instead? Unless I've missed something this class could be replaced...

Haven't considered it, but it seems a lot more concise than the class. Let's try it!

AlanGriffiths:
> Haven't considered it, but it seems a lot more concise than the class. Let's try it!

Another design iteration: make this a [template method](https://en.wikipedia.org/wiki/Template_method_pattern) on `AlarmFactory`. It doesn't really need a header file all of its own. WDYT? Seems even simpler to me

tarek-y-ismail:
> Another design iteration: make this a [template method](https://en.wikipedia.org/wiki/Template_method_pattern) on AlarmFactory. It doesn't really need a header file all of its own. WDYT? Seems even simpler to me

The usage of "template" confused me for a moment there, but it seems quite fitting and simple. Thank you for the suggestion :)

Once you approve, can you give me a chance to squash the commits to hide my shame?

src/include/server/mir/time/alarm_factory.h
```diff
@@ -61,6 +61,33 @@ class AlarmFactory
...
+                cb();
+
+                if (auto const& repeat_alarm = swk->lock())
+                    repeat_alarm->reschedule_in(repeat_delay);
+            });
+
+        *shared_weak_alarm = result;
+
+        return result;
+    }
```
AlanGriffiths: Let's hide the implementation in an implementation file. 😺

tarek-y-ismail: So I moved the implementation to `src/server/time/alarm_factory.cpp`, to follow existing patterns, I made the new `src/server/time/` directory a new library (`mirservertime` because `mirtime` was already taken) and added it to `src/CMakeLists.txt` as a subdirectory and to `MIR_SERVER_REFERENCES`. It feels a bit overkill, but that's what other modules are doing as well. Is there a simpler way that I didn't come across?

src/server/symbols.map
```diff
@@ -812,6 +812,7 @@ global:
     mir::time::AlarmFactory::?AlarmFactory*;
     mir::time::AlarmFactory::AlarmFactory*;
     mir::time::AlarmFactory::operator*;
+    mir::time::AlarmFactory::create_repeating_alarm*;
```
AlanGriffiths: 2.22 has branched: This needs a new MIR_SERVER_INTERNAL_2.23 stanza

AlanGriffiths (approved):
LGTM

Issue #4129, opened 2025-08-13
SDL apps misplaced


Saviq:
Didn't track how far back it broke, yet, but SDL apps are meh, again.

Not fitting in decorations, when windowed:
<img width="2804" height="2370" alt="Image" src="https://github.com/user-attachments/assets/01329558-b743-41d6-b4c6-c99d3c6253c4" />

In fullscreen, it's shifted and/or wrong size:
<img width="2804" height="2370" alt="Image" src="https://github.com/user-attachments/assets/27fc69f9-197b-4b4f-b119-e5ae139c71ee" />

AlanGriffiths:
What I'm seeing (on 24.04) is that any sizing done by the compositor (e.g. by dragging SSDs) is ignored. Sizing done through Options/Graphics/Configure does seem to work

AlanGriffiths:
> What I'm seeing (on 24.04) is that any sizing done by the compositor (e.g. by dragging SSDs) is ignored. Sizing done through Options/Graphics/Configure does seem to work

Trying with `MIRAL_SHELL_DECORATIONS=always-csd` the client-side decorations do not support resizing. That is consistent with SDL2 clients ignoring resizing events

AlanGriffiths:
I see the above effect on 25.10, the size of the offset buffer looks suspiciously like the size of the window with SSDs.

Things "work" with `MIRAL_SHELL_DECORATIONS=always-csd`

AlanGriffiths:
OK, so looking at the `WAYLAND_DEBUG` trace, SDL is setting the min & max size to, in my example, 2380x830 and submitting buffers of size 2380x830. Which is all sane (as the client area size is all it knows).

The window management trace also sees min & max size 2380x830 - without any allowance for decorations. Which is wrong

AlanGriffiths:
This is just wrong...

```
[2025-08-18 12:59:05.908040] <information> miral::Window Management: handle_modify_window window_info={name=Neverputt 1.6.0, type=freestyle, state=restored, top_left=528, 188, size=(1280, 800), children={}, min_width=1280, min_height=800, max_width=1280, max_height=800, exclusive_rect=0, preferred_orientation=0xf, confine_pointer=0}, modifications={size=(1292, 830)}
[...]

PR #4139, opened 2025-08-19
Fix window size constraints with ssd

AlanGriffiths:
The size constraints need to reflect the SSD margins

This was not possible as the Surface was responsible for the margins and the WindowInfo was responsible for the size constraints. So move the latter to the surface and fixup all around

Fixes: #4129


src/server/scene/basic_surface.cpp
```diff
@@ -1023,6 +1023,84 @@ void mir::scene::BasicSurface::set_tiled_edges(Flags<MirTiledEdge> edges)
     observers->tiled_edges(this, edges);
 }
 
+auto mir::scene::BasicSurface::min_width() const -> geometry::Width
```
mattkae, line 1026: These methods could be tested to make sure that the margin calculations are accurate

PR #4142, opened 2025-08-19
rfc: an example of how we can run an application switcher client in mir

mattkae:
This assumes that `mir-application-switcher` is a valid binary somewhere!

## How does it work?
1. The initial alt + tab runs `mir-application-switcher`
2. Because the switcher is a layer shell application, it will grab focus
3. Subsequent alt tabs will have no effect while the pid is still valid

### DIFF
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -43,6 +43,8 @@
 #include <xkbcommon/xkbcommon-keysyms.h>
 
 #include <cstring>
+#include <csignal>
+#include <sys/wait.h>
 
 namespace
 {
@@ -64,12 +66,32 @@ struct ConfigureDecorations
         decorations(s);
     }
 };
+
+bool is_process_running(pid_t pid) {
+    if (kill(pid, 0) == 0) {
+        return true;
+    }
+
+    if (errno == ESRCH)
+        return false;
+    else if (errno == EPERM)
+        return true;
+    else
+        return false;
+}
+
+void sigchld_handler(int) {
+    while (waitpid(-1, nullptr, WNOHANG) > 0) {
+    }
+}
 }
 
 int main(int argc, char const* argv[])
 {
     using namespace miral;
     using namespace miral::toolkit;
+
+    signal(SIGCHLD, sigchld_handler);
 
     std::function<void()> shutdown_hook{[]{}};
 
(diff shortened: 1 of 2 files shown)


tarek-y-ismail:
I'm a bit confused. What happens after the application switcher is launched? Does it get focus automatically and thus all subsequent alt+tabs go directly to it?

examples/miral-shell/shell_main.cpp
```diff
@@ -303,6 +325,35 @@ int main(int argc, char const* argv[])
...
+        if (modifiers & mir_input_event_modifier_alt)
+        {
+            switch (mir_keyboard_event_keysym(key_event))
+            {
+                case XKB_KEY_Tab:
+                {
+                    if (!app_switcher_pid || !is_process_running(app_switcher_pid.value()))
+                    {
+                        std::string app_selector = "mir-application-switcher";
+                        app_switcher_pid = external_client_launcher.launch({app_selector});
```
AlanGriffiths, line 344: I know this is a POC, but shouldn't we just be sending an activation token to an existing process? Not starting a new process.

examples/miral-shell/shell_main.cpp
```diff
@@ -64,13 +66,33 @@ struct ConfigureDecorations
...
+    else if (errno == EPERM)
+        return true;
+    else
+        return false;
+}
+
+void sigchld_handler(int) {
+    while (waitpid(-1, nullptr, WNOHANG) > 0) {
+    }
+}
```
AlanGriffiths, line 86: Ugly, and (assuming we migrated into miral) likely to interfere with any other process control a shell implements

examples/miral-shell/shell_main.cpp
```diff
@@ -333,6 +384,7 @@ int main(int argc, char const* argv[])
             sticky_keys,
             magnifier,
             AppendKeyboardEventFilter{magnifier_filter},
-            AppendKeyboardEventFilter{sticky_keys_filter}
+            AppendKeyboardEventFilter{sticky_keys_filter},
+            AppendKeyboardEventFilter{alt_tab_filter}
```
AlanGriffiths, line 388: Shouldn't this be `PrependKeyboardEventFilter`? I.e. take precedence over other filters?

examples/miral-shell/shell_main.cpp
mattkae: This way of doing things favors restarting the app every time instead of talking to some long-running process. This makes it so that global shortcuts are unnecessary. This is just a proof of concept, but it is the "simple" thing to do

examples/miral-shell/shell_main.cpp
mattkae: Yeah, this is very ugly :)

examples/miral-shell/shell_main.cpp
mattkae: True enough :)

PR #4144, opened 2025-08-21
Symbol checker updates

tarek-y-ismail:
### What's new
- Adds `requirements.txt`
- Converts an assertion to an if check to avoid the tool crashing


tools/symbols_map_generator/requirements.txt
```diff
@@ -0,0 +1 @@
+clang==20.1.5
```
Saviq: Is this correct, though? This needs to match the clang installed on the system, so it's not necessarily 20?
```
python3-clang/plucky 1:20.0-63ubuntu1 amd64
  Clang Python Bindings

python3-clang-14/plucky 1:14.0.6-20ubuntu1 amd64
  Clang Python Bindings

python3-clang-15/plucky 1:15.0.7-15ubuntu1 amd64
  Clang Python Bindings

python3-clang-17/plucky 1:17.0.6-21 amd64
  Clang Python Bindings

python3-clang-18/plucky 1:18.1.8-17 amd64
  Clang Python Bindings

python3-clang-19/plucky 1:19.1.7-3ubuntu1 amd64
  Clang Python Bindings

python3-clang-20/plucky 1:20.1.2-0ubuntu1 amd64
  Clang Python Bindings
```

tools/symbols_map_generator/main.py
```diff
@@ -249,11 +249,14 @@ def add_internal(s: str):
             # we need to generate a a virtual?thunk?to? for the method/constructor/deconstructor.
             if is_virtual:
                 def search_class_hierarchy_for_virtual_thunk(derived: clang.cindex.Cursor):
-                    assert (derived.kind == clang.cindex.CursorKind.CLASS_DECL
+                    
+                    if not (derived.kind == clang.cindex.CursorKind.CLASS_DECL
```
Saviq, line 253: Is it better not to "crash" in this case?

I mean, if we encounter an invalid (unexpected, really) derived kind, why continue?

tools/symbols_map_generator/requirements.txt
tarek-y-ismail: Hmm, then maybe its better to update the readme to guide users of the tools to install `python3-clang-<clang version>` instead of installing via `requirements.txt`?

Saviq: Yeah, we already guide them to install llvm-19+, so adding the matching python3-clang to it would make sense, yes.

tools/symbols_map_generator/main.py
tarek-y-ismail: Well, the code obviously compiles on my machine, so this is likely to be a clang error. If we crash, we don't get any output from the tool, but if we ignore invalid symbol kinds, then at least we can have some partial output.

tools/symbols_map_generator/requirements.txt
tarek-y-ismail: Completely forgot we had a how to in our sphinx docs. Should we just merge the readme into that?

Saviq: Sounds right, and refer there from the README here.

tools/symbols_map_generator/main.py
Saviq: What's the error? What's the `derived.kind` we're not expecting?

tarek-y-ismail: Added a bit more info to the print: 
```
Encountered invalid derived kind: CursorKind.INVALID_FILE in file:/home/tarek/code-projects/mir/src/include/server/mir/observer_multiplexer.h, location <SourceLocation file '/home/tarek/code-projects/mir/src/include/server/mir/observer_multiplexer.h', line 257, column 37>, display name register_interest(const std::weak_ptr<Observer> &)
Encountered invalid derived kind: CursorKind.INVALID_FILE in file:/home/tarek/code-projects/mir/src/include/server/mir/observer_multiplexer.h, location <SourceLocation file '/home/tarek/code-projects/mir/src/include/server/mir/observer_multiplexer.h', line 263, column 37>, display name register_interest(const std::weak_ptr<Observer> &, Executor &)
Encountered invalid derived kind: CursorKind.INVALID_FILE in file:/home/tarek/code-projects/mir/src/include/server/mir/observer_multiplexer.h, location <SourceLocation file '/home/tarek/code-projects/mir/src/include/server/mir/observer_multiplexer.h', line 273, column 37>, display name register_early_observer(const std::weak_ptr<Observer> &, Executor &)
Encountered invalid derived kind: CursorKind.INVALID_FILE in file:/home/tarek/code-projects/mir/src/include/server/mir/observer_multiplexer.h, location <SourceLocation file '/home/tarek/code-projects/mir/src/include/server/mir/observer_multiplexer.h', line 283, column 37>, display name unregister_interest(const Observer &)
```

doc/sphinx/how-to/how-to-update-symbols-map.md
```diff
@@ -81,8 +81,25 @@ export MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib
 It is recommended that you put those `export ...` commands in your `.bashrc`
 so that you don't have to think about it in the future.
 
-And that's it! Now we are ready to update our symbols automatically.
+One final set up step, you need to set up a venv environment and install the
```
Saviq:
```suggestion
One final set up step, you need to install the
```

doc/sphinx/how-to/how-to-update-symbols-map.md
```diff
@@ -81,8 +81,25 @@ export MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib
...
+```
+
+If you want to use a virtual environment, you can do so as follows, note that
+you might have to pass `--break-system-packages`:
+```
+cd tools/symbols_map_generator/
+python -m venv venv
+. ./venv/bin/activate
+pip install clang
+```
```
Saviq: I think we can drop this.
~~~suggestion
~~~

tools/symbols_map_generator/main.py
Saviq: @mattkae thoughts on `CursorKind.INVALID_FILE`? Should we just add it to the assertion?

Saviq (changes requested):

doc/sphinx/how-to/how-to-update-symbols-map.md
```diff
@@ -81,8 +81,16 @@ export MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib
...
-And that's it! Now we are ready to update our symbols automatically.
+One final set up step, you need to install the
+corresponding python clang library:
+```
+sudo apt install python3-clang-19
+```
 
+You'll have to run `. tools/symbols_map_generator/venv/bin/activate` before
+invoking the symbols map generator in a new terminal session.
+
```
tarek-y-ismail: Should be nuked as well

tools/symbols_map_generator/README.md
```diff
@@ -22,5 +19,5 @@ The following will write the introduced symbols to `src/miral/symbols.map` for v
 ```
 
 ## Requirements
-- `libclang-dev`
-- `pip install libclang`
+- `libclang-dev` from apt
+- The `clang` python package
```
tarek-y-ismail, line 23: small note here, libclang seems to be a slightly out of date mirror of LLVM's clang python bindings. Since the latter works fine, we can use that instead.

tools/symbols_map_generator/README.md
```diff
@@ -3,14 +3,11 @@ This tool parses the header files of a library in the Mir project to extract sym
...
-## Usage
-It is recommended that the script be run from a python venv:
-```sh
-python -m venv venv
-source venv/bin/activate
-pip install -r requirements.txt
-```
+## Set Up
+Please see [How to Update Symbols Map
+Files](https://canonical-mir.readthedocs-hosted.com/stable/how-to/how-to-update-symbols-map/)
```
mattkae, line 8: I am not a big fan of linking to another doc outside of the README for this? It feels like we're spreading the info around. I would still keep the install instructions here :)

tools/symbols_map_generator/main.py
mattkae: The reason that was an `assert` is that it's a precondition of getting into that statement rather than something that _might_ happen. The issue with getting partial data is that you may uniwittingly partially update the symbols, which would be unfortunate.

mattkae: It is very odd that we're getting `CursorKind.INVALID_FILE`. Apparently that means that the node is entirely invalid, maybe because the header file isn't in included in the build? I've never encountered that before.

At any rate, we probably need to fix something else if we have that error instead of just printing the error (e.g. we may need to include some other file for clang so that the generator can find it)

mattkae (changes requested):

tools/symbols_map_generator/README.md
tarek-y-ismail: The two install instructions would eventually go out of sync and confuse someone (like what happened to me earlier today)

tools/symbols_map_generator/main.py
tarek-y-ismail: Does this error occur on your end? It might be because that I'm using the clang package which looks like it's more recent libclang, so something might've changed

tools/symbols_map_generator/README.md
Saviq: I wouldn't leave anything more in the README than a reference to the doc. That, or moving that content here. DRY!

tools/symbols_map_generator/requirements.txt
Saviq: Didn't we agree this file should go away?

tools/symbols_map_generator/main.py
mattkae: I ran `cmake --build cmake-build-debug-regular --target generate-miral-symbols-map` and it worked fine. Were you running something else?

tarek-y-ismail: It happens only if you install bindings for clang 20 through apt (python3-clang-20) or the `clang` python package. Trying older versions of the clang python package doesn't show the error. And libclang doesn't show the error because it hasn't been updated past clang 18. 

I'm not sure if this is a bug with the bindings or clang, I took a look at clang's issue list and nothing explicitly mentions `CursorKind.INVALID_FILE`

Googling also doesn't help much, which might mean we're doing something wrong? If that's the case, We probably should band-aid this for the time being until we have time to revisit the symbol generator.

mattkae: Or maybe we just enforce clang 18 or 19 for now? I am wary of skipping definitions in the event that we start committing half-baked symbol files

tarek-y-ismail: I'm okay for that for the time being, I'll try submitting a bug report to the LLVM repo tomorrow morning and see what they have to say

PR #4149, opened 2025-08-21
`Renderer::render` should return a `Buffer` rather than a `Framebuffer`

tarek-y-ismail:
Note: tests are failing intentionally. Build with `-DMIR_ENABLE_TESTS=off` if you want to run locally. Will fix tests once the core idea is validated.

### DIFF
--- a/include/platform/mir/graphics/buffer.h
+++ b/include/platform/mir/graphics/buffer.h
@@ -26,7 +26,18 @@ namespace mir
 {
 namespace graphics
 {
-class Framebuffer;
+class Framebuffer
+{
+public:
+    Framebuffer() = default;
+    virtual ~Framebuffer() = default;
+
+    /**
+     * The size of this framebuffer, in pixels
+     */
+    virtual auto size() const -> geometry::Size = 0;
+};
+
 class NativeBufferBase
 {
 protected:
--- a/include/platform/mir/graphics/buffer.h
+++ b/include/platform/mir/graphics/buffer.h
@@ -45,7 +45,7 @@ public:
     virtual geometry::Size size() const = 0;
     virtual MirPixelFormat pixel_format() const = 0;
 
-    virtual NativeBufferBase* native_buffer_base() = 0;
+    virtual NativeBufferBase* native_buffer_base() { return nullptr; }
 
     virtual auto to_framebuffer() -> std::unique_ptr<Framebuffer> { return nullptr; }
 
--- a/include/platform/mir/graphics/buffer.h
+++ b/include/platform/mir/graphics/buffer.h
@@ -20,12 +20,13 @@
 #include "mir/graphics/buffer_id.h"
 #include "mir/geometry/size.h"
 #include "mir_toolkit/common.h"
+#include <memory>
 
 namespace mir
 {
 namespace graphics
 {
-
+class Framebuffer;
 class NativeBufferBase
 {
 protected:
@@ -45,6 +46,8 @@ public:
     virtual MirPixelFormat pixel_format() const = 0;
 
     virtual NativeBufferBase* native_buffer_base() = 0;
+
+    virtual auto to_framebuffer() -> std::unique_ptr<Framebuffer> { return nullptr; }
 
 protected:
     Buffer() = default;
--- a/include/platform/mir/graphics/platform.h
+++ b/include/platform/mir/graphics/platform.h
@@ -315,13 +315,13 @@ public:
     {
     };
 
-    class MappableFB :
-        public Framebuffer,
+    class MappableBuffer : 
+        public Buffer, 
         public mir::renderer::software::WriteMappableBuffer
     {
     public:
-        MappableFB() = default;
-        virtual ~MappableFB() override = default;
+        MappableBuffer() = default;
+        virtual ~MappableBuffer() override = default;
 
         using renderer::software::WriteMappableBuffer::size;
     };
@@ -329,8 +329,8 @@ public:
     virtual auto supported_formats() const
         -> std::vector<DRMFormat> = 0;
 
-    virtual auto alloc_fb(DRMFormat format)
-        -> std::unique_ptr<MappableFB> = 0;
+    virtual auto alloc_buffer(DRMFormat format) 
+        -> std::unique_ptr<MappableBuffer> = 0;
 
     virtual auto output_size() const -> geometry::Size = 0;
 };
@@ -397,7 +397,7 @@ public:
          * of buffers available and attempting to claim a framebuffer when no buffers
          * are free will result in an EBUSY std::system_error being raised.
          */
-        virtual auto claim_framebuffer() -> std::unique_ptr<Framebuffer> = 0;
+        virtual auto claim_buffer() -> std::unique_ptr<Buffer> = 0;
     };
 
     virtual auto make_surface(DRMFormat format, std::span<uint64_t> modifiers) -> std::unique_ptr<GBMSurface> = 0;
@@ -464,6 +464,7 @@ public:
         virtual void make_current() = 0;
         virtual void release_current() = 0;
         virtual auto clone_handle() -> std::unique_ptr<EGLFramebuffer> = 0;
+        virtual auto format() const -> MirPixelFormat = 0;
     };
 
     virtual auto alloc_framebuffer(GLConfig const& config, EGLContext share_context) -> std::unique_ptr<EGLFramebuffer> = 0;
--- a/include/platform/mir/graphics/platform.h
+++ b/include/platform/mir/graphics/platform.h
@@ -288,18 +288,6 @@ public:
     virtual ~DisplayAllocator() = default;
 };
 
-class Framebuffer
-{
-public:
-    Framebuffer() = default;
-    virtual ~Framebuffer() = default;
-
-    /**
-     * The size of this framebuffer, in pixels
-     */
-    virtual auto size() const -> geometry::Size = 0;
-};
-
 class CPUAddressableDisplayProvider : public DisplayProvider
 {
 public:
--- a/include/platform/mir/graphics/platform.h
+++ b/include/platform/mir/graphics/platform.h
@@ -458,16 +458,15 @@ public:
     {
     };
 
-    class EGLFramebuffer : public graphics::Framebuffer
+    class EGLBuffer : public graphics::Buffer
     {
     public:
         virtual void make_current() = 0;
         virtual void release_current() = 0;
-        virtual auto clone_handle() -> std::unique_ptr<EGLFramebuffer> = 0;
-        virtual auto format() const -> MirPixelFormat = 0;
+        virtual auto clone_handle() -> std::unique_ptr<EGLBuffer> = 0;
     };
 
-    virtual auto alloc_framebuffer(GLConfig const& config, EGLContext share_context) -> std::unique_ptr<EGLFramebuffer> = 0;
+    virtual auto alloc_buffer(GLConfig const& config, EGLContext share_context) -> std::unique_ptr<EGLBuffer> = 0;
 };
 
 class DisplayPlatform
--- a/src/platforms/atomic-kms/server/gbm_display_allocator.cpp
+++ b/src/platforms/atomic-kms/server/gbm_display_allocator.cpp
@@ -235,11 +235,6 @@ MirPixelFormat mir::graphics::atomic::GBMBuffer::pixel_format() const
     return mir_pixel_format_invalid;
 }
 
-mir::graphics::NativeBufferBase* mir::graphics::atomic::GBMBuffer::native_buffer_base()
-{
-    return this;
-}
-
 auto mga::GBMBuffer::to_framebuffer() -> std::unique_ptr<Framebuffer>
 {
     if (auto cached_fb = static_cast<std::shared_ptr<uint32_t const>*>(gbm_bo_get_user_data(front_buffer.get())))
--- a/src/platforms/atomic-kms/server/gbm_display_allocator.cpp
+++ b/src/platforms/atomic-kms/server/gbm_display_allocator.cpp
@@ -59,45 +59,6 @@ using LockedFrontBuffer = std::unique_ptr<gbm_bo, std::function<void(gbm_bo*)>>;
 class GBMBoFramebuffer : public mg::FBHandle
 {
 public:
-    static auto framebuffer_for_frontbuffer(mir::Fd const& drm_fd, LockedFrontBuffer bo) -> std::unique_ptr<GBMBoFramebuffer>
-    {
-        if (auto cached_fb = static_cast<std::shared_ptr<uint32_t const>*>(gbm_bo_get_user_data(bo.get())))
-        {
-            return std::unique_ptr<GBMBoFramebuffer>{new GBMBoFramebuffer{std::move(bo), *cached_fb}};
-        }
-
-        auto fb_id = std::shared_ptr<uint32_t>{
-            new uint32_t{0},
-            [drm_fd](uint32_t* fb_id)
-            {
-                if (*fb_id)
-                {
-                    drmModeRmFB(drm_fd, *fb_id);
-                }
-                delete fb_id;
-            }};
-        uint32_t handles[4] = {gbm_bo_get_handle(bo.get()).u32, 0, 0, 0};
-        uint32_t strides[4] = {gbm_bo_get_stride(bo.get()), 0, 0, 0};
-        uint32_t offsets[4] = {gbm_bo_get_offset(bo.get(), 0), 0, 0, 0};
-
-        auto format = gbm_bo_get_format(bo.get());
-
-        auto const width = gbm_bo_get_width(bo.get());
-        auto const height = gbm_bo_get_height(bo.get());
-
-        /* Create a KMS FB object with the gbm_bo attached to it. */
-        auto ret = drmModeAddFB2(drm_fd, width, height, format,
-                                 handles, strides, offsets, fb_id.get(), 0);
-        if (ret)
-            return nullptr;
-
-        // It is weird allocating a smart pointer on the heap, but we delete it
-        // via gbm_bo_set_user_data()'s destroy_user_data parameter.
-        gbm_bo_set_user_data(bo.get(), new std::shared_ptr<uint32_t>(fb_id), [](gbm_bo*, void* fb_ptr) { delete static_cast<std::shared_ptr<uint32_t const>*>(fb_ptr); });
-
-        return std::make_unique<GBMBoFramebuffer>(std::move(bo), std::move(fb_id));
-    }
-
     operator uint32_t() const override
     {
         return *fb_id;
--- a/src/platforms/atomic-kms/server/gbm_display_allocator.cpp
+++ b/src/platforms/atomic-kms/server/gbm_display_allocator.cpp
@@ -54,7 +54,6 @@ auto mga::GBMDisplayAllocator::modifiers_for_format(DRMFormat /*format*/) const 
 
 namespace
 {
-using LockedFrontBuffer = std::unique_ptr<gbm_bo, std::function<void(gbm_bo*)>>;
 
 class GBMBoFramebuffer : public mg::FBHandle
 {
@@ -72,13 +71,13 @@ public:
                 gbm_bo_get_height(bo.get())};
     }
 
-    GBMBoFramebuffer(LockedFrontBuffer bo, std::shared_ptr<uint32_t const> fb)
+    GBMBoFramebuffer(mga::LockedFrontBuffer bo, std::shared_ptr<uint32_t const> fb)
         : bo{std::move(bo)},
           fb_id{std::move(fb)}
     {
     }
 private:
-    LockedFrontBuffer const bo;
+    mga::LockedFrontBuffer const bo;
     std::shared_ptr<uint32_t const> const fb_id;
 };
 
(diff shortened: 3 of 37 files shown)


include/platform/mir/graphics/platform.h
```diff
@@ -315,22 +315,22 @@ class CPUAddressableDisplayAllocator : public DisplayAllocator
...
+        MappableBuffer() = default;
+        virtual ~MappableBuffer() override = default;
 
         using renderer::software::WriteMappableBuffer::size;
     };
 
-    virtual auto supported_formats() const
-        -> std::vector<DRMFormat> = 0;
+    virtual auto
+    supported_formats() const -> std::vector<DRMFormat> = 0;
```
tarek-y-ismail: Unnecessary change

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -222,12 +222,56 @@ class GBMSurfaceImpl : public mga::GBMDisplayAllocator::GBMSurface
...
+        public:
+            GBMBuffer(LockedFrontBuffer front_buffer)
+                : front_buffer{std::move(front_buffer)}
+            {
+            }
+
+            mg::BufferID id() const override
+            {
+                // https://stackoverflow.com/a/21574751
+                return mg::BufferID{static_cast<uint32_t>(reinterpret_cast<uintptr_t>(front_buffer.get()))};
```
tarek-y-ismail: Might be a bad idea

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -222,12 +222,56 @@ class GBMSurfaceImpl : public mga::GBMDisplayAllocator::GBMSurface
...
+                if(auto iter = mapping.find(gbm_format); iter != mapping.end())
+                    return iter->second;
+
+                return mir_pixel_format_invalid;
+            }
+
+            mg::NativeBufferBase* native_buffer_base() override
+            {
+                return this;
+            }
```
tarek-y-ismail: Also probably a bad idea

src/platforms/common/server/cpu_addressable_fb.h
```diff
@@ -24,34 +24,53 @@
 
 namespace mir::graphics
 {
-class CPUAddressableFB : public FBHandle, public CPUAddressableDisplayAllocator::MappableFB
+class CPUAddressableBuffer : 
+    public FBHandle,
```
tarek-y-ismail: Probably shouldn't inherit from `FBHandle` anymore

src/platforms/common/server/cpu_addressable_fb.h
```diff
@@ -24,34 +24,53 @@
...
+
+    MirPixelFormat pixel_format() const override
+    {
+        return format();
+    }
+
+    NativeBufferBase* native_buffer_base() override
+    {
+        return this;
+    }
```
tarek-y-ismail: Should be moved to the corresponding source file

src/platforms/common/server/kms_cpu_addressable_display_provider.cpp
```diff
@@ -38,6 +38,7 @@ auto drm_get_cap_checked(mir::Fd const& drm_fd, uint64_t cap) -> uint64_t
 
 namespace mg = mir::graphics;
 namespace geom = mir::geometry;
+namespace mrs = mir::renderer::software;
```
tarek-y-ismail: Not needed anymore

src/platforms/renderer-generic-egl/buffer_allocator.cpp
```diff
@@ -330,6 +330,39 @@ auto mge::GLRenderingProvider::as_texture(std::shared_ptr<Buffer> buffer) -> std
...
+
+public:
+    FramebufferAdapter(std::unique_ptr<EGLFramebuffer> framebuffer) :
+        framebuffer{std::move(framebuffer)}
+    {
+    }
+
+    mg::BufferID id() const override
+    {
+        return mg::BufferID{0u};
```
tarek-y-ismail: I gave up on the ID stuff at this point, luckily, it seems like it doesn't matter much for actual functionality

include/platform/mir/graphics/buffer.h
```diff
@@ -46,6 +47,8 @@ class Buffer
 
     virtual NativeBufferBase* native_buffer_base() = 0;
 
+    virtual auto to_framebuffer() -> std::unique_ptr<Framebuffer> { return nullptr; }
```
tarek-y-ismail, line 61: Hack, was added because I needed to extract a a framebuffer out of an `mga::GBMBuffer`, but since this was in the GBM module, I couldn't include the defintion of the buffer. The most straightforward solution in my opinion is to expose an interface that both modules (gbm, atomic) can implement that would have `to_framebuffer`

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -242,3 +197,84 @@ auto mga::GBMDisplayAllocator::make_surface(DRMFormat format, std::span<uint64_t
...
+    front_buffer{std::move(front_buffer)},
+    drm_fd{drm_fd}
+{
+}
+
+mir::graphics::BufferID mir::graphics::atomic::GBMBuffer::id() const
+{
+    // https://stackoverflow.com/a/21574751
+    return mir::graphics::BufferID{static_cast<uint32_t>(reinterpret_cast<uintptr_t>(front_buffer.get()))};
+}
```
tarek-y-ismail, line 210: Need to verify if this is okay or not, but it's low priority for now

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -242,3 +197,84 @@ auto mga::GBMDisplayAllocator::make_surface(DRMFormat format, std::span<uint64_t
...
+        {GBM_FORMAT_RGBA5551, mir_pixel_format_rgba_5551},
+        {GBM_FORMAT_RGBA4444, mir_pixel_format_rgba_4444},
+    };
+
+    auto const gbm_format = gbm_bo_get_format(front_buffer.get());
+    if (auto iter = mapping.find(gbm_format); iter != mapping.end())
+        return iter->second;
+
+    return mir_pixel_format_invalid;
+}
```
tarek-y-ismail, line 236: This is duplicated between the GBM and Atomic implementations, could be a good target for the base class

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -242,3 +197,84 @@ auto mga::GBMDisplayAllocator::make_surface(DRMFormat format, std::span<uint64_t
...
+
+mir::graphics::NativeBufferBase* mir::graphics::atomic::GBMBuffer::native_buffer_base()
+{
+    return this;
+}
```
tarek-y-ismail: :suspect:

src/platforms/eglstream-kms/server/buffer_allocator.cpp
```diff
@@ -811,10 +812,33 @@ auto mge::GLRenderingProvider::make_framebuffer_provider(DisplaySink& /*sink*/)
...
+
+                    operator uint32_t() const override
+                    {
+                        return wrapped->operator uint32_t();
+                    }
+
+                    std::shared_ptr<FBHandle> const wrapped;
+                };
+                return make_unique<Wrapper>(cpu_buffer);
+            }
```
tarek-y-ismail: Should probably have a `to_framebuffer` method that returns `CPUFramebuffer`, which would contain all the `FBHandle` bits of `CPUBuffer`

src/platforms/gbm-kms/server/buffer_allocator.cpp
```diff
@@ -591,10 +593,13 @@ auto mgg::GLRenderingProvider::make_framebuffer_provider(DisplaySink& sink)
...
     {
     public:
-        auto buffer_to_framebuffer(std::shared_ptr<Buffer>) -> std::unique_ptr<Framebuffer> override
+        auto buffer_to_framebuffer(std::shared_ptr<Buffer> buf) -> std::unique_ptr<Framebuffer> override
         {
             // It is safe to return nullptr; this will be treated as “this buffer cannot be used as
             // a framebuffer”.
+            if (auto mgg_gbm_buffer = std::dynamic_pointer_cast<mgg::GBMBuffer>(buf))
+                return mgg_gbm_buffer->to_framebuffer();
+
```
tarek-y-ismail: Is this still needed? `Null*` shouldn't really be doing anything

src/platforms/gbm-kms/server/gbm_display_allocator.cpp
```diff
@@ -214,3 +170,86 @@ auto mgg::GBMDisplayAllocator::make_surface(DRMFormat format, std::span<uint64_t
...
+    front_buffer{std::move(front_buffer)}, 
+    drm_fd{drm_fd}
+{
+}
+
+mir::graphics::BufferID mir::graphics::gbm::GBMBuffer::id() const
+{
+    // https://stackoverflow.com/a/21574751
+    return BufferID{static_cast<uint32_t>(reinterpret_cast<uintptr_t>(front_buffer.get()))};
+}
```
tarek-y-ismail, line 183: sus

src/platforms/wayland/wl_egl_display_provider.cpp
```diff
@@ -148,6 +164,11 @@ auto mgw::WlDisplayAllocator::Framebuffer::size() const -> geom::Size
     return size_;
 }
 
+auto mgw::WlDisplayAllocator::Framebuffer::format() const -> MirPixelFormat
+{
+    return state->format();
+}
+
```
tarek-y-ismail, line 176: More dupes

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -41,10 +41,12 @@ class mc::BasicScreenShooter::Self::OneShotBufferDisplayProvider : public mg::CP
 public:
     OneShotBufferDisplayProvider() = default;
 
-    class FB : public mg::CPUAddressableDisplayAllocator::MappableFB
+    class Buffer : 
+        public mg::CPUAddressableDisplayAllocator::MappableBuffer, 
+        public graphics::NativeBufferBase
```
tarek-y-ismail: I wonder if `NativeBufferBase` is even required..

src/platforms/renderer-generic-egl/buffer_allocator.cpp
```diff
@@ -330,6 +330,39 @@ auto mge::GLRenderingProvider::as_texture(std::shared_ptr<Buffer> buffer) -> std
 namespace
 {
 
+class FramebufferAdapter : public mg::Buffer, public mg::NativeBufferBase
+{
+    using EGLFramebuffer = mir::graphics::GenericEGLDisplayAllocator::EGLFramebuffer;
```
tarek-y-ismail: Probably needs to become `EGLBuffer` instead of adapting it into a buffer

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
tarek-y-ismail: Looks like `NativeBufferBase` is a way to sneak DMA, shared memory, and OpenGL textures via buffers. In the case of all the classes I've made inherit from it, they're probably not being checked, which probably means I can save some lines of code and return `nullptr`, but this makes me feel like `Buffer` shouldn't need to require `native_buffer_base` at all...

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
tarek-y-ismail: Seems like they're only used for tracing/logging, so I guess as long as the ID is unique, it should be alright?

src/platforms/gbm-kms/server/buffer_allocator.cpp
tarek-y-ismail: Since they're needed now, they should be renamed to `DefaultFramebufferProvider`

src/platforms/atomic-kms/server/kms/display_buffer.cpp
```diff
@@ -81,7 +81,7 @@ mga::DisplaySink::DisplaySink(
         auto mapping = initial_fb->map_writeable();
         ::memset(mapping->data(), 24, mapping->len());
 
-        visible_fb = std::move(initial_fb);
+        visible_fb = CPUAddressableBuffer::to_fb_handle(initial_fb);
```
tarek-y-ismail: This can be a method on `CPUAddressableBuffer`

src/platforms/atomic-kms/server/kms/display_buffer.cpp
```diff
@@ -72,7 +72,7 @@ mga::DisplaySink::DisplaySink(
     {
         mir::log_info("Clearing screen due to differing encountered and target modes");
         // TODO: Pull a supported format out of KMS rather than assuming XRGB8888
-        auto initial_fb = std::make_shared<graphics::CPUAddressableFB>(
+        auto initial_fb = std::make_shared<graphics::CPUAddressableBuffer>(
```
tarek-y-ismail: Can be a `unique_ptr`

src/platforms/common/server/cpu_addressable_fb.cpp
```diff
@@ -271,32 +271,82 @@ mg::CPUAddressableFB::CPUAddressableFB(
...
-    return fb_id;
+    return BufferID{fb_id};
+}
+
+MirPixelFormat mir::graphics::CPUAddressableBuffer::pixel_format() const
+{
+    return format();
+}
+
+auto mir::graphics::CPUAddressableBuffer::to_framebuffer(std::shared_ptr<CPUAddressableBuffer> buf)
```
tarek-y-ismail: If we implement this as a method, there's no straightforward way to get a shared reference to `this` other than `shared_from_this` which has its own footguns from my last exposure to it. And since the buffers passed to `buffer_to_framebuffer` are shared, it would be quite a bad idea to just use a raw pointer.

Maybe wrapping the state in a struct, having `CPUAddressableBuffer` store a shared pointer to that, and then pointing to that in the returned struct would be a less hacky solution. Still feels hacky though...

src/platforms/common/server/cpu_addressable_fb.cpp
```diff
@@ -271,32 +271,82 @@ mg::CPUAddressableFB::CPUAddressableFB(
...
+        {
+            return wrapped->size();
+        }
+
+        std::shared_ptr<CPUAddressableBuffer> const wrapped;
+    };
+    return make_unique<CPUAddressableFramebuffer>(std::move(buf));
+}
+
+auto mir::graphics::CPUAddressableBuffer::to_fb_handle(std::shared_ptr<CPUAddressableBuffer> buf)
```
tarek-y-ismail: This should become a method and move everything out of the buffer. It's only used in two places and both don't need to share ownership with anything else so it shouldn't cause issues.

src/platforms/common/server/cpu_addressable_fb.h
```diff
@@ -24,34 +24,40 @@
...
     auto stride() const -> geometry::Stride override;
-    auto size() const -> geometry::Size override; 
+    auto size() const -> geometry::Size override;
 
-    operator uint32_t() const override;
-    
-    CPUAddressableFB(CPUAddressableFB const&) = delete;
-    CPUAddressableFB& operator=(CPUAddressableFB const&) = delete;
+    BufferID id() const override;
+    MirPixelFormat pixel_format() const override;
```
tarek-y-ismail, line 44: Both `Buffer` and `BufferDescriptor` have `pixel_format` and `format` respectively. Maybe we could improve this

src/platforms/gbm-kms/server/buffer_allocator.cpp
```diff
@@ -584,21 +586,24 @@ auto mgg::GLRenderingProvider::make_framebuffer_provider(DisplaySink& sink)
...
+    class DefaultFramebufferProvider : public FramebufferProvider
     {
     public:
-        auto buffer_to_framebuffer(std::shared_ptr<Buffer>) -> std::unique_ptr<Framebuffer> override
+        auto buffer_to_framebuffer(std::shared_ptr<Buffer> buf) -> std::unique_ptr<Framebuffer> override
         {
             // It is safe to return nullptr; this will be treated as “this buffer cannot be used as
             // a framebuffer”.
+            if (auto mgg_gbm_buffer = std::dynamic_pointer_cast<mgg::GBMBuffer>(buf))
+                return mgg_gbm_buffer->to_framebuffer();
```
tarek-y-ismail: If we keep `Buffer::to_framebuffer` then we won't need to cast here.

src/platforms/gbm-kms/server/buffer_allocator.cpp
```diff
@@ -563,42 +564,43 @@ auto mgg::GLRenderingProvider::make_framebuffer_provider(DisplaySink& sink)
...
+
             return {};
```
tarek-y-ismail:
```suggestion
            buf->to_framebuffer();
```

src/platforms/gbm-kms/server/buffer_allocator.cpp
```diff
@@ -563,42 +564,43 @@ auto mgg::GLRenderingProvider::make_framebuffer_provider(DisplaySink& sink)
...
             DmaBufDisplayAllocator* allocator;
         };
 
-        return std::make_unique<FooFramebufferProvider>(allocator);
+        return std::make_unique<DmaBufFramebufferProvider>(allocator);
     }
 
     // TODO: Make this not a null implementation, so bypass/overlays can work again
-    class NullFramebufferProvider : public FramebufferProvider
+    class DefaultFramebufferProvider : public FramebufferProvider
```
tarek-y-ismail, line 590: Since most `FramebufferProvider` impls are doing the same thing, maybe it'd be better to make a `DefaultFramebufferProvider` and just return that?

include/platform/mir/graphics/buffer.h
tarek-y-ismail: Seems like the kind of hack worth keeping...

include/platform/mir/graphics/buffer.h
```diff
@@ -20,11 +20,23 @@
...
+{
+public:
+    Framebuffer() = default;
+    virtual ~Framebuffer() = default;
+
+    /**
+     * The size of this framebuffer, in pixels
+     */
+    virtual auto size() const -> geometry::Size = 0;
+};
```
tarek-y-ismail, line 39: Pretty sure this breaks something. But for some reason clang dealt with it like a champ while gcc immediately imploded..

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -242,3 +197,79 @@ auto mga::GBMDisplayAllocator::make_surface(DRMFormat format, std::span<uint64_t
...
+        {GBM_FORMAT_RGB565, mir_pixel_format_rgb_565},
+        {GBM_FORMAT_RGBA5551, mir_pixel_format_rgba_5551},
+        {GBM_FORMAT_RGBA4444, mir_pixel_format_rgba_4444},
+    };
+
+    auto const gbm_format = gbm_bo_get_format(front_buffer.get());
+    if (auto iter = mapping.find(gbm_format); iter != mapping.end())
+        return iter->second;
+
+    return mir_pixel_format_invalid;
```
RAOF, line 235: `GBM_FORMAT_foo` is guaranteed to be the same value as `DRM_FORMAT_foo`, and `mg::DRMFormat` has an `as_mir_format()` translator.

include/platform/mir/graphics/buffer.h
RAOF: We should not have this here on this interface.

The `Framebuffer` is specific to the `DisplayPlatform` that it needs to display on, which is not necessarily the one it's allocated on.

Instead, the `FrameBufferProvider::buffer_to_framebuffer` should `dynamic_cast` to the platform-specific concrete type (from `->native_buffer_base()`) in order to get it.

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
```diff
@@ -242,3 +197,79 @@ auto mga::GBMDisplayAllocator::make_surface(DRMFormat format, std::span<uint64_t
     return std::make_unique<GBMSurfaceImpl>(fd, gbm.get(), size, format, modifiers, gbm_quirks);
 }
 
+mir::graphics::atomic::GBMBuffer::GBMBuffer(mir::Fd drm_fd, LockedFrontBuffer front_buffer) :
+    front_buffer{std::move(front_buffer)},
+    drm_fd{drm_fd}
+{
+}
+
+mir::graphics::BufferID mir::graphics::atomic::GBMBuffer::id() const
```
RAOF, line 206: Derive from `mg::BufferBasic` :wink:

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
RAOF, line 235:
```suggestion
    return mg::DRMFormat{gbm_bo_get_format(front_buffer.get())}.as_mir_format().value_or(mir_pixel_format_invalid);
```

RAOF (changes requested):
Pair reviewed in 1:1; outcomes:
Maybe this would be simpler to make `mg::Framebuffer` implement `mg::Buffer`; drop all the renamings and keep `Renderer::render() -> unique_ptr<Framebuffer>` (but *now* that `Framebuffer` will *be* a `Buffer`, so code like the Magnifier will be able to use it directly as a buffer, which is the idea of this change.)

PR #4150, opened 2025-08-22
Fix crash in miral-kiosk and Frame

AlanGriffiths:
We might calculate the `restore_rect()` before there is a surface asociated with the WindowInfo


Saviq (approved):
Looks legit!

PR #4153, opened 2025-08-26
debian: drop extra symbol

Saviq:


tarek-y-ismail:
Should the corresponding symbol in `src/miroil/symbols.map` be deleted as well?

Saviq:
> Should the corresponding symbol in `src/miroil/symbols.map` be deleted as well?

Good catch. Done.

PR #25, opened 2017-11-10
Set up bors.tech merge bot

Saviq:
Tell travis to build staging and trying branches, and tell bors to wait
on travis before pushing to master.

PR #22, opened 2017-11-10
Add wlcs tests

RAOF:
This integrates `wlcs` with the Mir test-suite, gated on MIR_RUN_ACCEPTANCE_TESTS.

It then fixes the bugs exposed!

Issue #21, opened 2017-11-09
Using Milestones


pharaone:
I think you can set milestones here like you did in launchpad, thanks for maintaining mir ;)

pharaone:
You are doing it with 0.32  so I can close this issue, thanks!

PR #20, opened 2017-11-08
Find the gmock library on Fedora 28 (Fixes #12)

AlanGriffiths:

PR #4154, opened 2025-08-26
wip: implement the ApplicationSwitcher as an internal client instead of an internal class

mattkae:


Saviq:
FYI, getting a SIGSEGV here:

<details>
<summary>gdb</summary>

```gdb
#0  0x00007ffff7b86244 in wl_proxy_get_version () at /lib/x86_64-linux-gnu/libwayland-client.so.0
#1  0x00007ffff7eafa04 in zwlr_layer_shell_v1_get_layer_surface (zwlr_layer_shell_v1=0x0, surface=0x7fff7000c3b0, output=0x0, layer=3, namespace_=0x7ffff7ee5941 "miral")
    at /home/michal/dev/canonical/mir/build-Debug/miral-generated/wlr-layer-shell-unstable-v1.h:229
        id = 0x7fff70004e40
#2  0x00007ffff7eafcf6 in miral::tk::LayerShellWaylandSurface::LayerShellWaylandSurface (this=0x7fff7000c430, app=0x7fff70004e40, creation_params=...)
    at /home/michal/dev/canonical/mir/src/miral/layer_shell_wayland_surface.cpp:48
#3  0x00007ffff7d3c040 in std::_Construct<miral::tk::LayerShellWaylandSurface, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&>
    (__p=0x7fff7000c430) at /usr/include/c++/14/bits/stl_construct.h:119
#4  0x00007ffff7d3ae0f in std::allocator_traits<std::allocator<void> >::construct<miral::tk::LayerShellWaylandSurface, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (__p=0x7fff7000c430) at /usr/include/c++/14/bits/alloc_traits.h:706
#5  std::_Sp_counted_ptr_inplace<miral::tk::LayerShellWaylandSurface, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_Sp_counted_ptr_inplace<(anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (this=0x7fff7000c420, __a=...) at /usr/include/c++/14/bits/shared_ptr_base.h:607
#6  0x00007ffff7d39508 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count<miral::tk::LayerShellWaylandSurface, std::allocator<void>, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (this=0x7fff85ff9f88, __p=@0x7fff85ff9f80: 0x0, __a=...) at /usr/include/c++/14/bits/shared_ptr_base.h:969
        __guard = {_M_alloc = 0x7fff85ff9e0f, _M_ptr = 0x7fff7000c420}
        __mem = 0x7fff7000c420
        __a2 = {<std::__new_allocator<std::_Sp_counted_ptr_inplace<miral::tk::LayerShellWaylandSurface, std::allocator<void>, (__gnu_cxx::_Lock_policy)2> >> = {<No data fields>}, <No data fields>}
        __pi = 0x7fff85ff9f88
        __a2 = {<std::__new_allocator<std::_Sp_counted_ptr_inplace<miral::tk::LayerShellWaylandSurface, std::allocator<void>, (__gnu_cxx::_Lock_policy)2> >> = {<No data fields>}, <No data fields>}
        __guard = {_M_alloc = <optimized out>, _M_ptr = <optimized out>}
        __mem = <optimized out>
        __pi = <optimized out>
#7  0x00007ffff7d3739a in std::__shared_ptr<miral::tk::LayerShellWaylandSurface, (__gnu_cxx::_Lock_policy)2>::__shared_ptr<std::allocator<void>, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (this=0x7fff85ff9f80, __tag=...) at /usr/include/c++/14/bits/shared_ptr_base.h:1713
#8  0x00007ffff7d34eaf in std::shared_ptr<miral::tk::LayerShellWaylandSurface>::shared_ptr<std::allocator<void>, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (this=0x7fff85ff9f80, __tag=...) at /usr/include/c++/14/bits/shared_ptr.h:463
#9  0x00007ffff7d32116 in std::make_shared<miral::tk::LayerShellWaylandSurface, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> ()
    at /usr/include/c++/14/bits/shared_ptr.h:1008
        __a = {<No data fields>}
#10 0x00007ffff7d2da1a in (anonymous namespace)::WaylandApp::init (this=0x7fff70004e40, display=0x7fff70000b70)
    at /home/michal/dev/canonical/mir/src/miral/application_switcher.cpp:309
        surface_params = {size = {width = {value = 0}, height = {value = 0}}, layer = 3, anchor = 15}
#11 0x00007ffff7d2ee66 in miral::ApplicationSwitcher::Self::run_client (this=0x5555556e69f0, display=0x7fff70000b70)
    at /home/michal/dev/canonical/mir/src/miral/application_switcher.cpp:606
        fds = {{fd = -2046844896, events = 32767, revents = 0}, {fd = 1433314272, events = 21845, revents = 0}}
        lock = {_M_device = @0x5555556e6a00}
#12 0x00007ffff7d2aa60 in miral::ApplicationSwitcher::operator() (this=0x5555556ea3e0, display=0x7fff70000b70)
    at /home/michal/dev/canonical/mir/src/miral/application_switcher.cpp:698
#13 0x000055555557a904 in std::__invoke_impl<void, miral::ApplicationSwitcher&, wl_display*> (__f=...) at /usr/include/c++/14/bits/invoke.h:61
#14 0x0000555555579017 in std::__invoke_r<void, miral::ApplicationSwitcher&, wl_display*> (__fn=...) at /usr/include/c++/14/bits/invoke.h:111
#15 0x0000555555575ecb in std::_Function_handler<void(wl_display*), miral::ApplicationSwitcher>::_M_invoke (__functor=..., __args#0=@0x7fff85ffa0d0: 0x7fff70000b70)
    at /usr/include/c++/14/bits/std_function.h:290
#16 0x00007ffff7dd769f in std::function<void(wl_display*)>::operator() (this=0x5555556ea420, __args#0=0x7fff70000b70) at /usr/include/c++/14/bits/std_function.h:591
#17 0x00007ffff7dd701f in WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}::operator()() const (__closure=0x55555570a948)
    at /home/michal/dev/canonical/mir/src/miral/internal_client.cpp:128
        deleter = std::unique_ptr<wl_display> = {get() = 0x7fff70000b70}
        display = 0x7fff70000b70
#18 0x00007ffff7dd8e07 in std::__invoke_impl<void, WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}>(std::__invoke_other, WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}&&) (__f=...) at /usr/include/c++/14/bits/invoke.h:61
#19 0x00007ffff7dd8d63 in std::__invoke<WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}>(WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}&&) (__fn=...) at /usr/include/c++/14/bits/invoke.h:96
#20 0x00007ffff7dd8cd4 in std::thread::_Invoker<std::tuple<WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}> >::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0x55555570a948) at /usr/include/c++/14/bits/std_thread.h:301
#21 0x00007ffff7dd8c84 in std::thread::_Invoker<std::tuple<WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}> >::operator()()
    (this=0x55555570a948) at /usr/include/c++/14/bits/std_thread.h:308
#22 0x00007ffff7dd8c40 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}> > >::_M_run() (this=0x55555570a940) at /usr/include/c++/14/bits/std_thread.h:253
#23 0x00007ffff7862324 in ??? () at /lib/x86_64-linux-gnu/libstdc++.so.6
#24 0x00007ffff75c47f1 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:448
        ret = <optimized out>
        pd = <optimized out>
        out = <optimized out>
        unwind_buf = {cancel_jmp_buf = {{jmp_buf = {128, 1979338188565450446, -3360, 32, 11, 140737488339936, 1979338188544478926, 1979579287733997262}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x0, 0x0}, data = {prev = 0x0, cleanup = 0x0, canceltype = 0}}}
        not_first_call = <optimized out>
#25 0x00007ffff7655c9c in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```

</details>

mattkae:
> FYI, getting a SIGSEGV here:
> gdb
> 
> ```gdb
> #0  0x00007ffff7b86244 in wl_proxy_get_version () at /lib/x86_64-linux-gnu/libwayland-client.so.0
> #1  0x00007ffff7eafa04 in zwlr_layer_shell_v1_get_layer_surface (zwlr_layer_shell_v1=0x0, surface=0x7fff7000c3b0, output=0x0, layer=3, namespace_=0x7ffff7ee5941 "miral")
>     at /home/michal/dev/canonical/mir/build-Debug/miral-generated/wlr-layer-shell-unstable-v1.h:229
>         id = 0x7fff70004e40
> #2  0x00007ffff7eafcf6 in miral::tk::LayerShellWaylandSurface::LayerShellWaylandSurface (this=0x7fff7000c430, app=0x7fff70004e40, creation_params=...)
>     at /home/michal/dev/canonical/mir/src/miral/layer_shell_wayland_surface.cpp:48
> #3  0x00007ffff7d3c040 in std::_Construct<miral::tk::LayerShellWaylandSurface, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&>
>     (__p=0x7fff7000c430) at /usr/include/c++/14/bits/stl_construct.h:119
> #4  0x00007ffff7d3ae0f in std::allocator_traits<std::allocator<void> >::construct<miral::tk::LayerShellWaylandSurface, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (__p=0x7fff7000c430) at /usr/include/c++/14/bits/alloc_traits.h:706
> #5  std::_Sp_counted_ptr_inplace<miral::tk::LayerShellWaylandSurface, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_Sp_counted_ptr_inplace<(anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (this=0x7fff7000c420, __a=...) at /usr/include/c++/14/bits/shared_ptr_base.h:607
> #6  0x00007ffff7d39508 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count<miral::tk::LayerShellWaylandSurface, std::allocator<void>, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (this=0x7fff85ff9f88, __p=@0x7fff85ff9f80: 0x0, __a=...) at /usr/include/c++/14/bits/shared_ptr_base.h:969
>         __guard = {_M_alloc = 0x7fff85ff9e0f, _M_ptr = 0x7fff7000c420}
>         __mem = 0x7fff7000c420
>         __a2 = {<std::__new_allocator<std::_Sp_counted_ptr_inplace<miral::tk::LayerShellWaylandSurface, std::allocator<void>, (__gnu_cxx::_Lock_policy)2> >> = {<No data fields>}, <No data fields>}
>         __pi = 0x7fff85ff9f88
>         __a2 = {<std::__new_allocator<std::_Sp_counted_ptr_inplace<miral::tk::LayerShellWaylandSurface, std::allocator<void>, (__gnu_cxx::_Lock_policy)2> >> = {<No data fields>}, <No data fields>}
>         __guard = {_M_alloc = <optimized out>, _M_ptr = <optimized out>}
>         __mem = <optimized out>
>         __pi = <optimized out>
> #7  0x00007ffff7d3739a in std::__shared_ptr<miral::tk::LayerShellWaylandSurface, (__gnu_cxx::_Lock_policy)2>::__shared_ptr<std::allocator<void>, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (this=0x7fff85ff9f80, __tag=...) at /usr/include/c++/14/bits/shared_ptr_base.h:1713
> #8  0x00007ffff7d34eaf in std::shared_ptr<miral::tk::LayerShellWaylandSurface>::shared_ptr<std::allocator<void>, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> (this=0x7fff85ff9f80, __tag=...) at /usr/include/c++/14/bits/shared_ptr.h:463
> #9  0x00007ffff7d32116 in std::make_shared<miral::tk::LayerShellWaylandSurface, (anonymous namespace)::WaylandApp*, miral::tk::LayerShellWaylandSurface::CreationParams&> ()
>     at /usr/include/c++/14/bits/shared_ptr.h:1008
>         __a = {<No data fields>}
> #10 0x00007ffff7d2da1a in (anonymous namespace)::WaylandApp::init (this=0x7fff70004e40, display=0x7fff70000b70)
>     at /home/michal/dev/canonical/mir/src/miral/application_switcher.cpp:309
>         surface_params = {size = {width = {value = 0}, height = {value = 0}}, layer = 3, anchor = 15}
> #11 0x00007ffff7d2ee66 in miral::ApplicationSwitcher::Self::run_client (this=0x5555556e69f0, display=0x7fff70000b70)
>     at /home/michal/dev/canonical/mir/src/miral/application_switcher.cpp:606
>         fds = {{fd = -2046844896, events = 32767, revents = 0}, {fd = 1433314272, events = 21845, revents = 0}}
>         lock = {_M_device = @0x5555556e6a00}
> #12 0x00007ffff7d2aa60 in miral::ApplicationSwitcher::operator() (this=0x5555556ea3e0, display=0x7fff70000b70)
>     at /home/michal/dev/canonical/mir/src/miral/application_switcher.cpp:698
> #13 0x000055555557a904 in std::__invoke_impl<void, miral::ApplicationSwitcher&, wl_display*> (__f=...) at /usr/include/c++/14/bits/invoke.h:61
> #14 0x0000555555579017 in std::__invoke_r<void, miral::ApplicationSwitcher&, wl_display*> (__fn=...) at /usr/include/c++/14/bits/invoke.h:111
> #15 0x0000555555575ecb in std::_Function_handler<void(wl_display*), miral::ApplicationSwitcher>::_M_invoke (__functor=..., __args#0=@0x7fff85ffa0d0: 0x7fff70000b70)
>     at /usr/include/c++/14/bits/std_function.h:290
> #16 0x00007ffff7dd769f in std::function<void(wl_display*)>::operator() (this=0x5555556ea420, __args#0=0x7fff70000b70) at /usr/include/c++/14/bits/std_function.h:591
> #17 0x00007ffff7dd701f in WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}::operator()() const (__closure=0x55555570a948)
>     at /home/michal/dev/canonical/mir/src/miral/internal_client.cpp:128
>         deleter = std::unique_ptr<wl_display> = {get() = 0x7fff70000b70}
>         display = 0x7fff70000b70
> #18 0x00007ffff7dd8e07 in std::__invoke_impl<void, WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}>(std::__invoke_other, WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}&&) (__f=...) at /usr/include/c++/14/bits/invoke.h:61
> #19 0x00007ffff7dd8d63 in std::__invoke<WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}>(WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}&&) (__fn=...) at /usr/include/c++/14/bits/invoke.h:96
> #20 0x00007ffff7dd8cd4 in std::thread::_Invoker<std::tuple<WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}> >::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0x55555570a948) at /usr/include/c++/14/bits/std_thread.h:301
> #21 0x00007ffff7dd8c84 in std::thread::_Invoker<std::tuple<WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}> >::operator()()
>     (this=0x55555570a948) at /usr/include/c++/14/bits/std_thread.h:308
> #22 0x00007ffff7dd8c40 in std::thread::_State_impl<std::thread::_Invoker<std::tuple<WlInternalClientRunner<miral::StartupInternalClient::Self>::run(mir::Server&)::{lambda()#1}> > >::_M_run() (this=0x55555570a940) at /usr/include/c++/14/bits/std_thread.h:253
> #23 0x00007ffff7862324 in ??? () at /lib/x86_64-linux-gnu/libstdc++.so.6
> #24 0x00007ffff75c47f1 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:448
>         ret = <optimized out>
>         pd = <optimized out>
>         out = <optimized out>
>         unwind_buf = {cancel_jmp_buf = {{jmp_buf = {128, 1979338188565450446, -3360, 32, 11, 140737488339936, 1979338188544478926, 1979579287733997262}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x0, 0x0}, data = {prev = 0x0, cleanup = 0x0, canceltype = 0}}}
>         not_first_call = <optimized out>
> #25 0x00007ffff7655c9c in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
> ```

Fixed :)

examples/miral-shell/shell_main.cpp
```diff
@@ -319,7 +367,8 @@ int main(int argc, char const* argv[])
             config_keymap,
             AppendKeyboardEventFilter{quit_on_ctrl_alt_bksp},
             StartupInternalClient{spinner},
-            ConfigurationOption{run_startup_apps, "startup-apps", "Colon separated list of startup apps.", ""},
+            StartupInternalClient{application_switcher},
+            ConfigurationOption{run_startup_apps, "startup-apps", "Colon separated list of startup apps", ""},
```
tarek-y-ismail, line 374: Removed the period by mistake?

src/miral/wayland_shm.cpp
```diff
@@ -0,0 +1,195 @@
...
+                        return mir::Fd{};
+                    }
+                }
+                return fd;
+            };
+
+            return open_shm_file();
+        };
+
+    auto fd = open_shm_file();
```
tarek-y-ismail: Using the same name for both the inner and outer functions might be slightly confusing :pinching_hand:

examples/example-server-lib/decoration_provider.cpp
```diff
@@ -47,8 +47,12 @@ struct preferred_codecvt : std::codecvt_byname<wchar_t, char, std::mbstate_t>
 {
     preferred_codecvt() : std::codecvt_byname<wchar_t, char, std::mbstate_t>("C") {}
     ~preferred_codecvt() = default;
+
+    static std::locale::id id;
 };
 
+std::locale::id preferred_codecvt::id;
+
```
tarek-y-ismail, line 55: Huh?

src/miral/application_switcher.cpp
```diff
@@ -0,0 +1,734 @@
...
+
+    struct Color {
+        uint8_t r, g, b, a;
+    };
+
+    static constexpr Color white{255, 255, 255, 255};
+    static constexpr Color yellow{255, 255, 0, 255};
+
+    // Font search logic should be kept in sync with src/server/shell/decoration/renderer.cpp
+    static auto default_font() -> std::string
```
tarek-y-ismail: Other parts feel like they're also borrowed from other parts of Mir. Maybe it'd be wise to unify things properly? Though we first need to agree whether or not we want to go down this path first.

src/miral/application_switcher.cpp
```diff
@@ -0,0 +1,734 @@
...
+            for (auto i = 0; i != 3; ++i)
+                pattern[i] = (j * bottom_color[i] + (height - j) * top_color[i]) / height;
+            pattern[3] = 0xff;
+
+            auto const pixel = reinterpret_cast<uint32_t*>(row);
+            for (int i = 0; i < width; i++)
+                mempcpy(pixel + i, pattern, sizeof pixel[i]);
+
+            row += stride;
+        }
```
tarek-y-ismail, line 524: Maybe we can make it so only a small area around the text is opaque and the rest is transparent? I also remember you mentioning that layer shell would center the buffer anyway. If that's true, then we can probably draw to a small area instead too

tarek-y-ismail (commented):
Initial thoughts, will come back again after letting this simmer in my head for a bit

Saviq:
```
/usr/bin/ld: /tmp/ccgwf4hu.ltrans0.ltrans.o:(.data.rel+0x48): undefined reference to `xdg_popup_interface'
collect2: error: ld returned 1 exit status
```

tarek-y-ismail:
> ```
> /usr/bin/ld: /tmp/ccgwf4hu.ltrans0.ltrans.o:(.data.rel+0x48): undefined reference to `xdg_popup_interface'
> collect2: error: ld returned 1 exit status
> ```

This is really weird, what do any of the changes here have to do with this? It happens on my machine as well

Saviq:
> what do any of the changes here have to do with this?

Just a missing `target_link_libraries` somewhere, most likely…

mattkae:
> > what do any of the changes here have to do with this?
> 
> Just a missing `target_link_libraries` somewhere, most likely…

Probably because I'm not generating it :)

tarek-y-ismail:
~Ok I officially have no idea why CI fails at this point:~ ~Edit: Of course, I found something suspicious with my changes immediately after posting this comment. Let's see if that fixes CI.~ Nope, it's still broken. @mattkae maybe I'm missing something? Can you take a closer look?

```
dh_makeshlibs -plibmiral7 -- -v5.6.0
dpkg-gensymbols: error: some symbols or patterns disappeared in the symbols file: see diff output below
dpkg-gensymbols: warning: debian/libmiral7/DEBIAN/symbols doesn't match completely debian/libmiral7.symbols
--- debian/libmiral7.symbols (libmiral7_5.6.0_armhf)
+++ dpkg-gensymbolsfs90se	2025-09-18 08:28:20.153557679 +0000
@@ -39,7 +39,7 @@
  (c++)"miral::BasicSetApplicationAuthorizer::operator()(mir::Server&)@MIRAL_5.0" 5.0.0
  (c++)"miral::BasicSetApplicationAuthorizer::the_application_authorizer() const@MIRAL_5.0" 5.0.0
  (c++)"miral::BasicSetApplicationAuthorizer::~BasicSetApplicationAuthorizer()@MIRAL_5.0" 5.0.0
- (c++)"miral::BounceKeys::BounceKeys(bool)@MIRAL_5.5" 5.5.0
+#MISSING: 5.6.0# (c++)"miral::BounceKeys::BounceKeys(bool)@MIRAL_5.5" 5.5.0
  (c++)"miral::BounceKeys::BounceKeys(miral::live_config::Store&)@MIRAL_5.5" 5.5.0
  (c++)"miral::BounceKeys::BounceKeys(std::shared_ptr<miral::BounceKeys::Self>)@MIRAL_5.5" 5.5.0
  (c++|arch-bits=32)"miral::BounceKeys::delay(std::chrono::duration<long long, std::ratio<1ll, 1000ll> >)@MIRAL_5.5" 5.5.0
@@ -93,8 +93,8 @@
  (c++)"miral::Decorations::prefer_ssd()@MIRAL_5.1" 5.1.0
  (c++)"miral::DisplayConfiguration::DisplayConfiguration(miral::DisplayConfiguration const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::DisplayConfiguration::DisplayConfiguration(miral::MirRunner const&)@MIRAL_5.0" 5.0.0
- (c++)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.3" 5.5.0
- (c++)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.5" 5.5.0
+#MISSING: 5.6.0# (c++)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.3" 5.5.0
+#MISSING: 5.6.0# (c++)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.5" 5.5.0
  (c++)"miral::DisplayConfiguration::Node::Node(std::unique_ptr<miral::DisplayConfiguration::Node::Self, std::default_delete<miral::DisplayConfiguration::Node::Self> >&&)@MIRAL_5.3" 5.3.0
  (c++)"miral::DisplayConfiguration::Node::as_int() const@MIRAL_5.3" 5.3.0
  (c++)"miral::DisplayConfiguration::Node::as_string[abi:cxx11]() const@MIRAL_5.3" 5.3.0
@@ -618,14 +618,14 @@
  (c++)"miral::live_config::IniFile::load_file(std::basic_istream<char, std::char_traits<char> >&, std::filesystem::__cxx11::path const&)@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::IniFile::on_done(std::function<void ()>)@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::IniFile::~IniFile()@MIRAL_5.5" 5.5.0
- (c++)"miral::live_config::Key::Key(miral::live_config::Key const&)@MIRAL_5.5" 5.5.0
+#MISSING: 5.6.0# (c++)"miral::live_config::Key::Key(miral::live_config::Key const&)@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::Key::Key(std::initializer_list<std::basic_string_view<char, std::char_traits<char> > const>)@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::Key::operator<=>(miral::live_config::Key const&) const@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::Key::operator==(miral::live_config::Key const&) const@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::Key::to_string[abi:cxx11]() const@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::Key::with_key(std::function<void (std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)> const&) const@MIRAL_5.5" 5.5.0
- (c++)"miral::live_config::Store::Store()@MIRAL_5.5" 5.5.0
- (c++)"miral::live_config::Store::~Store()@MIRAL_5.5" 5.5.0
+#MISSING: 5.6.0# (c++)"miral::live_config::Store::Store()@MIRAL_5.5" 5.5.0
+#MISSING: 5.6.0# (c++)"miral::live_config::Store::~Store()@MIRAL_5.5" 5.5.0
  (c++)"miral::name_of[abi:cxx11](std::shared_ptr<mir::scene::Session> const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::operator<(miral::Window const&, miral::Window const&)@MIRAL_5.0" 5.0.0
  (c++)"miral::operator==(miral::Output::PhysicalSizeMM const&, miral::Output::PhysicalSizeMM const&)@MIRAL_5.0" 5.0.0
@@ -673,4 +673,4 @@
  (c++)"vtable for miral::MinimalWindowManager@MIRAL_5.0" 5.0.0
  (c++)"vtable for miral::WindowManagementPolicy@MIRAL_5.0" 5.0.0
  (c++)"vtable for miral::live_config::IniFile@MIRAL_5.5" 5.5.0
- (c++)"vtable for miral::live_config::Store@MIRAL_5.5" 5.5.0
+#MISSING: 5.6.0# (c++)"vtable for miral::live_config::Store@MIRAL_5.5" 5.5.0
dh_makeshlibs: error: failing due to earlier errors
make[1]: *** [debian/rules:102: override_dh_makeshlibs] Error 25
make[1]: Leaving directory '/<<PKGBUILDDIR>>'
make: *** [debian/rules:13: binary-arch] Error 2
dpkg-buildpackage: error: fakeroot debian/rules binary-arch subprocess returned exit status 2
```

examples/example-server-lib/decoration_provider.cpp
mattkae: We need this unfortunately. Without it, the instance will get mad that the two separate implementations of `preferred_codecvt` (one in miral and one in example-server-lib) have the same id, which is invalid at compile time.

src/miral/application_switcher.cpp
mattkae: To be honest, they don't need to be kept in sync. I just copy and pasted this without thinking :)

src/miral/wayland_shm.cpp
mattkae: This is purposefully a self-modifying method call that then calls itself :)

RAOF: This *is* a neat optimisation, but
1. It is desperately in need of being documented as such,
1. It makes the function thread-unsafe,
1. This is not a fast-path; at worst three extra failing `open` syscalls won't be too much of a problem, and if it *was*
1. This would be more clearly expressed with a top-level `static bool have_tmpfile` and `static char const* shm_dir`

src/miral/wayland_shm.cpp
```diff
@@ -0,0 +1,195 @@
...
+    if (fd < 0) {
+        BOOST_THROW_EXCEPTION((std::system_error{errno, std::system_category(), "Failed to open shm buffer"}));
+    }
+
+    if (auto const error = posix_fallocate(fd, 0, size))
+    {
+        BOOST_THROW_EXCEPTION((std::system_error{error, std::system_category(), "Failed to allocate shm buffer"}));
+    }
+
+    if ((*data = mmap(nullptr, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0)) == MAP_FAILED)
```
RAOF: Yeah, I actually agree with the TICS warning here; I don't think it's unreasonable to split the assignment and the check.

src/miral/wayland_shm.h
```diff
@@ -0,0 +1,121 @@
...
+    /// \param data the data in the buffer
+    /// \param size size of the buffer
+    /// \param stride stride of the buffer
+    /// \param buffer wayland buffer object
+    WaylandShmBuffer(
+        std::shared_ptr<WaylandShmPool> const& pool,
+        void* data,
+        mir::geometry::Size const& size,
+        mir::geometry::Stride const& stride,
+        wl_buffer* buffer);
```
RAOF: I don't know why this constructor is public, or why it takes a `std::shared_ptr<WaylandShmPool>` argument (a `wl_shm_pool` can be a throwaway object; it doesn't need to outlast the `wl_shm_buffer` allocated from it).

src/miral/wayland_shm.h
```diff
@@ -0,0 +1,121 @@
...
+    mir::geometry::Stride const stride_;
+    wl_buffer* const buffer;
+    std::shared_ptr<WaylandShmBuffer> self_ptr; ///< Is set on use and cleared on release
+};
+
+/// A buffer pool that provides access to Wayland shared memory buffers.
+///
+/// An instance of this class does not efficiently provision multiple buffers for multiple window sizes.
+/// Please use one instance of this class per window if windows may have distinct sizes.
+class WaylandShm
```
RAOF: Maybe this would be better named `WaylandShmPool` (given that it's described a pool of buffers? :grin: )

src/miral/wayland_shm.cpp
```diff
@@ -0,0 +1,195 @@
...
+        {
+            free_buffer = buffer;
+            break;
+        }
+    }
+
+    // If we don't find one, we create a new buffer with a single-use pool
+    if (!free_buffer)
+    {
+        void* data;
```
RAOF: You could avoid the TICS complaint by changing the signature of `make_shm_pool` to `std::pair<wl_shm_pool*, void*> make_shm_pool(wl_shm* shm, size_t size)` and then
```
auto const [pool, data] = make_shm_pool(shm, data_size)
```

src/miral/wayland_shm.h
```diff
@@ -0,0 +1,121 @@
...
+
+    /// \returns the stride
+    auto stride() const -> mir::geometry::Stride { return stride_; }
+
+    /// Returns if this buffer is currently being used by the compositor.
+    ///
+    /// In-use buffers keep themselves alive.
+    ///
+    /// \returns `true` if in use, otherwise `false`.
+    auto is_in_use() const -> bool { return self_ptr != nullptr; }
```
RAOF: Does this need to exist (or, rather, to be public)? The pool is only giving out not-in-use buffers.

src/miral/wayland_shm.h
```diff
@@ -0,0 +1,121 @@
...
+
+    /// Marks this buffer as in-use and assumes the resulting #wl_buffer is sent to the compositor.
+    ///
+    /// Keeps this buffer  alive until the compositor releases it. If the buffer is not sent to the compositor
+    /// or the compositor never releases it, this buffer is leaked. This method should only be called if
+    /// the buffer is not already in use.
+    ///
+    /// \returns the wayland buffer object
+    auto use() -> wl_buffer*;
```
RAOF, line 52: Would this be simpler as
```c++
void attach_to(wl_surface* surface);
```
doing the only valid thing possible with the buffer?

src/miral/wayland_shm.cpp
```diff
@@ -0,0 +1,195 @@
...
+    auto const self = static_cast<WaylandShmBuffer*>(data);
+    self->self_ptr.reset();
+}
+
+miral::tk::WaylandShm::WaylandShm(wl_shm* shm)
+    : shm{shm}
+{
+}
+
+auto miral::tk::WaylandShm::get_buffer(geom::Size const& size, geom::Stride const& stride) -> std::shared_ptr<WaylandShmBuffer>
```
RAOF: It seems a bit of a foot-gun to have a method that will allocate buffers if there are none free but with no way to free those buffers other than by asking for a different `size`/`stride` pair.

RAOF (commented):
(Most of) the missing `debian/libmiral7.symbols` look to be of the form
```
class Foo {
...
   Foo(Foo&&) = default;
...
}
```
and this is a case where the symbol may well legitimately not be emitted - for an inline function (as the implicit constructor is here) the compiler is not required to emit a symbol for it unless some code takes the address of it.

These should probably be marked as `(c++|optional)` in the `debian/libmiral7.symbols` file; it truly doesn't matter if they appear or not.

That's not *all* the symbols (why is `BounceKeys::BounceKeys(bool)` in there? That function doesn't exist in the code? :confused: ), but it's at least some of them.

This also sniped me into some review; comments added.

debian/libmiral7.symbols
```diff
@@ -544,7 +544,6 @@ libmiral.so.7 libmiral7 #MINVER#
  MIRAL_5.5@MIRAL_5.5 5.5.0
  (c++)"miral::AppendKeyboardEventFilter::AppendKeyboardEventFilter(std::function<bool (MirKeyboardEvent const*)> const&)@MIRAL_5.5" 5.5.0
  (c++)"miral::AppendKeyboardEventFilter::operator()(mir::Server&)@MIRAL_5.5" 5.5.0
- (c++)"miral::BounceKeys::BounceKeys(bool)@MIRAL_5.5" 5.5.0
```
AlanGriffiths: Why are you changing an existing, released, stanza?

AlanGriffiths: Huh?! There hasn't been such a symbol since 82e5457bd9f397d8e0b78d41eb776c54b5b3dee7 - how did this ever work?

include/miral/miral/application_switcher.h
```diff
@@ -0,0 +1,76 @@
...
+/// to trigger #ApplicationSwitcher::next_app.
+///
+/// Instances of this class should be provided to #miral::StartupInternalClient.
+///
+/// Users must call #ApplicationSwitcher::stop in the stop callback of their server
+/// to stop the switcher.
+///
+/// This internal client relies on the following protocols:
+/// 1. wlr foreign toplevel management v1
+/// 2. wlr layer shell v1
```
AlanGriffiths, line 40:
```suggestion
/// 2. wlr layer shell v1
/// \remark Since MirAL 5.6
```

debian/libmiral7.symbols
tarek-y-ismail: That was my doing ✋ . I got nerd sniped by the symbol problems as well since my review and was trying to poke at the symbols to see what might be causing the issue

src/miral/wayland_shm.h
mattkae: Yeah I agree, and the existing `WaylandShmPool` can just be `Self` :)

src/miral/wayland_shm.h
mattkae: This is a weird design that I copy-and-pasted from the code in the example directory. But we should update it, because it feels needlessly complicated

mattkae: That's a weird design originally.

We make a "pool" for each buffer. That is not a pool at all!

mattkae: Ok I take it back - we actually need this. The pool gets allocated for each buffer, so the buffer itself needs to own the pool. I can make this easier on the eyes thoguh

src/miral/wayland_shm.h
mattkae: I think it makes more sense for the surface to do the attaching instead of the buffer, just in terms of sequestering responsibilities and whatnot.

src/miral/wayland_shm.cpp
mattkae: Isn't that the purpose of the pool here? The pool essentially has ownership over the buffers, and they in turn get "leased" when needed.

The only way to free them is to free the pool, by design. Unless you think that the pool is unnecessary altogether?

src/miral/wayland_shm.cpp
mattkae: I just cleaned this up to be a straightforward function. No need to go wild ;)

debian/libmiral7.symbols
```diff
@@ -632,18 +631,18 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::live_config::IniFile::load_file(std::basic_istream<char, std::char_traits<char> >&, std::filesystem::__cxx11::path const&)@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::IniFile::on_done(std::function<void ()>)@MIRAL_5.5" 5.5.0
  (c++)"miral::live_config::IniFile::~IniFile()@MIRAL_5.5" 5.5.0
- (c++)"miral::live_config::Key::Key(miral::live_config::Key const&)@MIRAL_5.5" 5.5.0
+ (c++|optional)"miral::live_config::Key::Key(miral::live_config::Key const&)@MIRAL_5.5" 5.5.0
```
AlanGriffiths: It makes no sense for ` miral::live_config::Key::Key(miral::live_config::Key const&)` to be marked optional - there is no way the toolchain can inline that.

debian/libmiral7.symbols
```diff
@@ -554,8 +553,8 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::BounceKeys::on_press_rejected(std::function<void (MirKeyboardEvent const*)>&&)@MIRAL_5.5" 5.5.0
  (c++)"miral::BounceKeys::operator()(mir::Server&)@MIRAL_5.5" 5.5.0
  (c++)"miral::CursorScale::CursorScale(miral::live_config::Store&)@MIRAL_5.5" 5.5.0
- (c++)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.3" 5.5.0
- (c++)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.5" 5.5.0
+ (c++|optional)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.3" 5.5.0
+ (c++|optional)"miral::DisplayConfiguration::Node::Node(miral::DisplayConfiguration::Node&&)@MIRAL_5.5" 5.5.0
```
AlanGriffiths: Something weird is going on: we shouldn't be publishing nor requiring this symbol in two stanzas. Marking it optional in both doesn't fix the root cause.

mattkae:
> We should also bump `MIRAL_VERSION_MINOR`: we are adding to the API

Did I not do that? https://github.com/canonical/mir/pull/4154/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4

include/miral/miral/live_config.h
```diff
@@ -97,8 +97,8 @@ class Store
     /// multiple attributes to be updated transactionally
     virtual void on_done(HandleDone handler) = 0;
 
-    Store() = default;
-    virtual ~Store() = default;
+    Store();
+    virtual ~Store();
```
AlanGriffiths: This does change the ABI and API. But, AFAICS, clients built against the old can use the new

AlanGriffiths:
> > We should also bump `MIRAL_VERSION_MINOR`: we are adding to the API
> 
> Did I not do that? https://github.com/canonical/mir/pull/4154/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4

Sorry, missed that

Saviq:
You need to kill the old switcher ;)

I saw a noticeable delay between me hitting Tab and the highlight moving down the list, something to look at :)

include/miral/miral/live_config.h
mattkae: This is now addressed in https://github.com/canonical/mir/pull/4300 instead of this PR

examples/example-server-lib/decoration_provider.cpp
AlanGriffiths: This doesn't change the fact there are two separate implementations (of two different things - the definitions are in different namespaces). it just makes the implementations more alike.

Saviq:
> 1. Doesn't seem to work. (see below)

`--add-wayland-extensions zwlr_layer_shell_v1:zwlr_foreign_toplevel_manager_v1` will fix that, but we should try and make that not necessary… can we enable extensions for internal clients like we do for shell components in Miriway?

> 2. AIUI this is going to break every downstream that uses miral::MinimalWindowManager - we probably don't want to do that.

By "break" you mean they would lose Alt+Tab handling? The new switcher should provide the default, visual, way for that.

mattkae:
> > 2. AIUI this is going to break every downstream that uses miral::MinimalWindowManager - we probably don't want to do that.
> 
> By "break" you mean they would lose Alt+Tab handling? The new switcher should provide the default, visual, way for that.

I suppose that they can continue to use it, but they should be aware that it's unsupported behavior going forward. Unfortunately, they won't be aware of this because they won't receive any deprecation notice in the code.

AlanGriffiths:
> > 1. Doesn't seem to work. (see below)
> 
> `--add-wayland-extensions zwlr_layer_shell_v1:zwlr_foreign_toplevel_manager_v1` will fix that, but we should try and make that not necessary… can we enable extensions for internal clients like we do for shell components in Miriway?

Yes. There are ways for the compositors to do that, and the potential for miral to provide better defaults.

> By "break" you mean they would lose Alt+Tab handling? The new switcher should provide the default, visual, way for that.

Yes, I mean that if you run `miral-app -demo-server` (or many existing downstreams) it has no app switching. THIS IS BREAKING THE ABI. (This can happen to Miriway, for example, without even recompiling Miriway.)

What we _should_ be doing is deprecating the old API and introducing a new one. That way downstreams can migrate in their own time

AlanGriffiths:
The list isn't very intuitive: neither name corresponds to what I, as a user, see the apps as:

screenshot

mattkae:
@Saviq and @AlanGriffiths 

Here is the game plan:
1. I reverted the removal of the `ApplicationSelector` and made the example code use `Ctrl + Tab` to control the new `ApplicationSwitcher` internal client
2. Next, we will deprecate `MinimalWindowManager` in favor of `MinimalWindowManagerV2`. This can come at any point, but it is what is necessary if we wan to deprecate the `ApplicationSelector`.
3. Also, we can add some mechanism for internal clients to "automatically" gain access to the protocols that they require. We can do this later as well.

mattkae:
https://github.com/canonical/mir/issues/4314 boom

https://github.com/canonical/mir/issues/4315 and boom

mattkae:
> The list isn't very intuitive: neither name corresponds to what I, as a user, see the apps as:
> screenshot

V1 of this thing is intentionally very barebones. I could resolve the names, but it would take a bit extra to do so. Could that be a followup? We could even include the icons too

AlanGriffiths:
> V1 of this thing is intentionally very barebones

Which isn't a problem if it is hidden behind a feature flag - but replacing the existing implementation requires more

src/miral/application_switcher.cpp
```diff
@@ -0,0 +1,743 @@
...
+            }},
+            FontPath{"LiberationSans-Bold.ttf", {
+                "liberation-sans",      // Fedora
+                "liberation-sans-fonts",// Fedora >= 42
+                "liberation",           // Arch/Ubuntu
+            }},
+            FontPath{"OpenSans-Bold.ttf", {
+                "open-sans",            // Fedora/Ubuntu
+            }},
+        };
```
AlanGriffiths: This is at least the third copy of this logic. We should rationalize some of the functionality this supports. We have sufficient usecases to identify the commonality

mattkae:
> Can we indicate it's experimental status? Maybe hide it in some way? in `example-server-lib`? Or a compile flag?
> 
> ```c++
> #ifndef MIRAL_EXPERIMENTAL_FEATURES_ENABLE_APPLICATION_SWITCHER
> #error This feature is under active development use at own risk!
> #endif
> ```
> 
> Or wait with landing until some of the above are addressed?

I think that the current interface is fine as-is, but we should maybe mark it as "experimental", but not at compile time. Perhaps we just add a big warning to the docstring? Marking it as deprecated is probably the next best option.

I wouldn't want to put it behind an `#ifndef` out of fear that the symbols are going to be upset.

AlanGriffiths:
> I think that the current interface is fine as-is, but we should maybe mark it as "experimental", but not at compile time. Perhaps we just add a big warning to the docstring? Marking it as deprecated is probably the next best option.

I agree the current interface isn't an issue. A doxygen `\warning` that the implementation is WIP would at least manage the expectations of anyone stumbling across the class.

src/miral/application_switcher.cpp
mattkae: I will keep it copied in the example (since that is "technically" a downstream and can decide on whatever font it wants), but I will refactor it out for the other cases.

mattkae:
@AlanGriffiths Re-requested you because I moved `default_font` to a place

AlanGriffiths (approved):
I'll leave someone else to decide if we want to bump `MIR_VERSION_MINOR` before branching 2.23

Issue #4157, opened 2025-08-27
Changing cursor_scale causes weird cursors and crashes


AlanGriffiths:
Using miriway from edge/pr168 (i.e. with the Mir 2.22 "live config" integration) it is easy to edit `~/.config/miriway-shell.settings` and change `cursor_scale`.

Choosing 1.9 leads to a broken cursor image and/or a compositor crash.

PR #4161, opened 2025-08-28
Fix use of std::clamp

AlanGriffiths:
Fixes: #4157


AlanGriffiths:
> Could swear we fixed this before

That was: 148ed5a086ca5c9d3edc0efa89b74e9358b0fb7b

Issue #4160, opened 2025-08-28
on Ubuntu 25.10 wf_recorder y-inverts some frames


AlanGriffiths:
It is unclear where this problem lies:

On 24.04 wf-recorder 0.4.1 binds linux-dmabuf-v1 version 3 and works
On 25.10 wf-recorder 0.5.0 binds linux-dmabuf-v1 version 4 and fails

I see nothing suspicious in our implementation of this protocol

https://github.com/user-attachments/assets/1406112d-95e6-47d0-8c6c-c11787d86626

AlanGriffiths:
Using wayvnc works fine (but that also binds linux-dmabuf-v1 version 3)

AlanGriffiths:
Still on 25.10 wf-recorder 0.5.0 works with Sway

AlanGriffiths:
The only obvious difference with Sway is that we send `zwlr_screencopy_frame_v1.buffer()` where Sway sends `zwlr_screencopy_frame_v1.linux_dmabuf()`

AlanGriffiths:
OK, the difference is sway sends `zwlr_screencopy_frame_v1.flags()` and we don't. Fix incoming...

PR #4162, opened 2025-08-28
Fix wlr-screencopy

AlanGriffiths:
Send the missing flags() event

Fixes: #4160


Saviq (approved):
> If the capture is successful, the compositor will send a "flags" event followed by a "ready" event.

Not optional indeed.

PR #4163, opened 2025-08-29
tests: add missing test requirement checks

Saviq:


Saviq:
> didn't fail as expected.

`rm CMakeCache.txt` :shrug:

tarek-y-ismail:
Why remove the xwayland check?

Saviq:
I dropped the `Xwayland` req, as the tests quirk themselves off in CI:

https://github.com/canonical/mir/blob/70e1d0b7bcacf7cd820284af316b85bab3492826/tests/miral/external_client.cpp#L71-L77

Saviq:
Dunno, on the other hand maybe we should have it still…

tarek-y-ismail:
> Dunno, on the other hand maybe we should have it still…

Probably a good idea to check for it for users trying Mir from  X based display servers, though I'm not sure that _all_ developers will want the dependency. What do you think? In my opinion it should be okay for now until Wayland is the default on more things.

tarek-y-ismail:
Aaaand CI now fails

Edit: I think it's under `xorg-x11-server-Xwayland` on fedora

Saviq:
> check for it for users trying Mir from X based display servers

It's a dependency of the test, so if we want to run that test, it's a build dependency… We'd need another switch to quirk X11 off, but that's out of scope.

That the test happens to turn itself off if `/tmp` is the runtime dir is a quirk of the test, but it still assumes `Xwayland` is there…

> Aaaand CI now fails

Yeah I'll work through these.

tarek-y-ismail (approved):

PR #4179, opened 2025-09-03
Drop legacy design diagram

AlanGriffiths:
It isn't used anywhere, is in an obsolete format and overlaps with more recent content


mattkae (approved):
Drop it1

PR #4183, opened 2025-09-03
doc: bring linkcheck back

Saviq:


tarek-y-ismail (dismissed):

tarek-y-ismail:
We'll need to add a step to configure a Mir cmake project, similar to what we do with readthedocs

PR #4187, opened 2025-09-03
doc: include internal APIs in documentation

jhenstridge:
This adds the internal headers to the Doxygen API docs build.

I had looked at trying to build this as a separate build to the public headers, but while the breathe extension supports multiple projects, the exhale extension we're using does not. It also [doesn't support Doxygen's grouping feature](https://exhale.readthedocs.io/en/latest/faq.html#my-documentation-is-setup-using-groups-how-can-i-use-exhale), so that is also not an option.

I also noticed a few errors in the build process that I've corrected too:

* add a bunch of `-I` arguments to help Doxygen parse headers with clang. As is, many files failed to parse on unknown `#include` directives.
* Update a couple of headers that were missing includes.

### DIFF
--- a/doc/sphinx/Doxyfile.in
+++ b/doc/sphinx/Doxyfile.in
@@ -852,6 +852,9 @@ INPUT                  = @CMAKE_SOURCE_DIR@/README.md \
                          @CMAKE_SOURCE_DIR@/include/core \
                          @CMAKE_SOURCE_DIR@/include/miral \
                          @CMAKE_SOURCE_DIR@/include/miroil \
+                         @CMAKE_SOURCE_DIR@/include/server \
+                         @CMAKE_SOURCE_DIR@/include/platform \
+                         @CMAKE_SOURCE_DIR@/src/include \
                          @MIR_GENERATED_INCLUDE_DIRECTORIES_FLATTENED@
 
 # This tag can be used to specify the character encoding of the source files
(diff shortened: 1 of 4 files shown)


AlanGriffiths:
image

I worry that this is giving Doxygen too much work: it was already taking a long time and this pushes a lot more into every build

jhenstridge:
> I worry that this is giving Doxygen too much work: it was already taking a long time and this pushes a lot more into every build

The underlying issue here is that the sphinx-exhale extension buffers all of the output from Doxygen until it completes, which makes it look like the build has stalled. I had a look to see if it was possible to turn that off, but it seems to go one step further and detect when being run on RTD and completely discard that output:

https://github.com/svenevs/exhale/blob/c99592533143b5144a69bf685c85cdb63e534b0b/exhale/deploy.py#L96-L102

We could tell exhale not to run Doxygen itself (i.e. set `exhaleExecutesDoxygen` to False in custom_conf.py), and run it before calling sphinx.

If we decide not to include the internal APIs in the docs, it is probably still worth keeping the include directives changes from this PR for better parsing of the headers.

Saviq:
> We could tell exhale not to run Doxygen itself (i.e. set `exhaleExecutesDoxygen` to False in custom_conf.py), and run it before calling sphinx.

That's in #4189 indeed

PR #4189, opened 2025-09-03
Integrate docs build with CMake, disable building API docs by default

tarek-y-ismail:
# What's new  
- Forwards the following targets to the sphinx Makefile: `clean`, `clean-doc`, `html`, `linkcheck`, `lint-md`, `pa11y`, `run`, `serve`, `spelling`, `vale`, `woke`
- API docs are not built by default to save time on each build. To build the docs with them, configure cmake with `-DBUILD_DOXYGEN`, or set `MIR_BUILD_API_DOCS=1` in your environment before calling CMake (or make)
- Added a bit of error handling and error reporting to users when the relevant files are missing
- Most generated doc files are now saved to the CMake build directory. The main exception is the `doc/sphinx/api` directory that's output from exhale, as that directory must be a subdirectory of `doc/sphinx`.

# How to test
- Configure Mir: `cmake -B <build_dir>`, optionally `-DBUILD_DOXYGEN=ON` to build API docs (takes a long time)
- Build and serve docs at `0.0.0.0:8000` via `--target doc-run`, 
 
`doc-clean` calls the `clean` target of the sphinx starter pack Makefile, cleaning the venv and generated files. `doxygen-clean` only removes doxygen's and exhale's output. If the cmake project is configured to build with doxygen, `doc-clean` removes all of the aforementioned files.

# TODO
- [x] There's also a `run` Makefile target that's not yet exposed. I think it opens the page in your browser automatically, and watches for changes. But that got annoying during testing. Tell me if you want that as a cmake target as well.
- [x] The output only shows after the command completes, but serving keeps running indefinitely. How can we show the output?
- [x] Test if this works properly with CI (most likely not at the moment)
  It "works", but doesn't build API docs by default. We need to pass `MIR_BUILD_API_DOCS=1` to whatever is launching sphinx, but according to [this issue](https://github.com/readthedocs/readthedocs.org/issues/6311), that's only available at the moment through the project admin panel.
- [ ] ~Serving multiple times in a row seems to eat all of my RAM to the point of the OOM having to kill the process. Should check if this is a memory leak or something else. It's solved by cleaning, re-installing, and re-running~
  Can't reliably reproduce it at the moment. Building API docs does seem to trigger it, but I'm not sure if they just use so much memory, or if there's actually a bug.
- [x] ~If you generate API docs, and then don't set the environment variable and re-server, the API docs will still be there because the files are not removed. Should we try removing those files if we're serving/building without the API docs env var?~ Let's not delete them and see what happens.
- [x] ~Sometimes the index page uses the cppguide for some reason???????????????????~ cppguide is being converted to markdown


doc/CMakeLists.txt
```diff
@@ -1,3 +1,27 @@
+add_custom_target(doc-setup
+    COMMAND make -C ${CMAKE_SOURCE_DIR}/doc/sphinx install
+    BYPRODUCTS ${CMAKE_SOURCE_DIR}/doc/sphinx/.sphinx/venv
+)
+
+add_custom_target(doc ${ALL}
```
Saviq: `${ALL}` is always empty here, you want to move the `set()` from down below.

doc/sphinx/conf.py
```diff
@@ -296,11 +294,14 @@
 ############################################################
 ### Additional configuration
 ############################################################
-cmake_build_dir = Path('../../')  # Leave doc/sphinx
+cmake_build_dir = Path(os.getenv('MIR_CMAKE_BUILD_DIR') or '../../')  # Default for CI
```
Saviq:
```suggestion
cmake_build_dir = Path(os.getenv('MIR_CMAKE_BUILD_DIR', '../../')  # Default for CI
```

doc/CMakeLists.txt
```diff
@@ -10,47 +34,38 @@ endif()
...
   )
 
+  add_custom_target(doxygen-clean
+      COMMAND rm -r ${WORKING_DIR}/doxygen_output/xml
+  ) 
+  add_dependencies(doc-clean-all doxygen-clean)
+
   install(
-      DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build/
+      DIRECTORY ${CMAKE_SOURCE_DIR}/doc/sphinx/_build/
```
Saviq: We don't want to mess with what's in `CMAKE_SOURCE_DIR`.

doc/CMakeLists.txt
```diff
@@ -10,47 +34,38 @@ endif()
...
+
 if(DOXYGEN_FOUND)
-  cmake_path(RELATIVE_PATH CMAKE_SOURCE_DIR BASE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR} OUTPUT_VARIABLE MIR_SOURCE_RELATIVE_PATH)
-  if(NOT ${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
-    file(
-        COPY ${CMAKE_CURRENT_SOURCE_DIR}/sphinx/
-        DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/sphinx/
-        PATTERN *.in EXCLUDE
-    )
-  endif()
```
Saviq, line 21: I still think we want this, if the user pointed at a build dir, we shouldn't mess with what's in the source dir.

doc/sphinx/conf.py
```diff
@@ -317,42 +318,39 @@
...
-    "contentsDirectives": False,
-    "exhaleDoxygenStdin": doxyfile_contents
-}
-
-# Setup the breathe extension
-breathe_projects = {"Mir": "./xml/"}
-breathe_default_project = "Mir"
-breathe_default_members = ('members', 'undoc-members')
-breathe_order_parameters_first = True
+build_api_docs = os.getenv('MIR_BUILD_API_DOCS') == '1'
```
Saviq, line 305: This should be set by (some of?) the CMake targets.

doc/CMakeLists.txt
tarek-y-ismail: Doesn't this just copy the built documentation files to `<install prefix>/share/doc/mir-doc`? (which just made me realize this is in the wrong place in the file..)

doc/sphinx/conf.py
tarek-y-ismail: I didn't want to add even more targets for such a special case, but if you're okay with that...

doc/CMakeLists.txt
tarek-y-ismail: We don't mess with the source dir. Doxygen's output is saved inside the build dir. This is merely saving a copy that we really don't need anymore.

doc/CMakeLists.txt
Saviq: Yes, but the built documentation shouldn't be in `CMAKE_SOURCE_DIR`.

doc/CMakeLists.txt
Saviq: Doxygen's output may be saved to the build dir, but Sphinx's is not. All the above targets work on `CMAKE_SOURCE_DIR`, where they should on `CMAKE_CURRENT_BINARY_DIR`, having copied the tree there.

doc/sphinx/conf.py
Saviq: This way we reduce the magic we need to put into CI and readthedocs.

Saviq (changes requested):
This builds the documentation in the source tree. We don't want to mess with the source tree, if you configured within a build dir.

That's why we had the `COPY` - to copy the whole `doc/sphinx` tree into the build dir, and continue from there.

Having built the docs, I end up with a lot of stuff in the source dir:
```
$ git status --ignored
On branch MIRENG-1400/improve-docs-build
Your branch is up to date with 'origin/MIRENG-1400/improve-docs-build'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        doc/sphinx/api/
        doc/sphinx/reference/introducing_the_miral_api.md

Ignored files:
  (use "git add -f <file>..." to include in what will be committed)
        build-Debug/
        doc/sphinx/.sphinx/.doctrees/
        doc/sphinx/.sphinx/venv/
        doc/sphinx/.sphinx/warnings.txt
        doc/sphinx/_build/
        include/miral/miral/version.h
        include/server/mir/version.h
```

doc/CMakeLists.txt
```diff
@@ -1,3 +1,48 @@
...
+# The later only overrides variables in the makefile.
+add_custom_target(doc
+  COMMAND ${CMAKE_COMMAND} -E env "MIR_CMAKE_BUILD_DIR=${CMAKE_BINARY_DIR}" 
+          make -C ${CMAKE_SOURCE_DIR}/doc/sphinx html 
+          BUILDDIR=${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build
+  DEPENDS ${CMAKE_SOURCE_DIR}/doc/sphinx/.sphinx/venv
+  BYPRODUCTS ${CMAKE_SOURCE_DIR}/doc/sphinx/_build
+)
+
+add_custom_target(doc-with-api
```
Saviq: Shouldn't this depend on `doxygen`?

doc/sphinx/conf.py
```diff
@@ -296,11 +294,14 @@
 ############################################################
 ### Additional configuration
 ############################################################
-cmake_build_dir = Path('../../')  # Leave doc/sphinx
+cmake_build_dir = Path(os.getenv('MIR_CMAKE_BUILD_DIR', '../../'))  # Default for CI
```
Saviq: We could actually require `MIR_CMAKE_BUILD_DIR` to be set, no?

It's unclear whether environment set [at job level](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idenv) trickles down to the [reused workflow](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows), but it's easy to check :)

doc/CMakeLists.txt
tarek-y-ismail: Since CMake isn't the one handling the input dependencies of doxygen, it calls the command every time, even when nothing changes. I chose to forego the dependency on the doxygen target to save time for us. We can invoke the doxygen target manually in CI

doc/CMakeLists.txt
tarek-y-ismail: Should be handled now by us passing `BUILDDIR=<path>` when invoking the makefile targets

doc/sphinx/conf.py
tarek-y-ismail: One more environment variable to set in readthedoc's admin panel :)

doc/CMakeLists.txt
Saviq: Right, but if you `make doc-with-api`, and then `make doc`, that would still include the, potentially stale, API docs - but it won't take ages. So if you explicitly `make doc-with-api`, wouldn't you expect that the API docs would get updated?

doc/sphinx/conf.py
Saviq: Ah, b/c RTD calls Sphinx on its own. Fine!

Saviq: image

doc/CMakeLists.txt
```diff
@@ -1,3 +1,44 @@
...
+  COMMAND ${CMAKE_COMMAND} -E env "MIR_BUILD_API_DOCS=1" 
+          ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --target doc
+)
+
+add_custom_target(doc-serve
+  COMMAND ${CMAKE_COMMAND} -E env "MIR_CMAKE_BUILD_DIR=${CMAKE_BINARY_DIR}" 
+          make -C ${CMAKE_SOURCE_DIR}/doc/sphinx serve
```
Saviq: How about this?
```suggestion
          make -C ${CMAKE_SOURCE_DIR}/doc/sphinx run
```
I know this is lying, a bit, but `run` is much more useful than `serve`?

Saviq: Or well, this?
```suggestion
add_custom_target(doc-run
  COMMAND ${CMAKE_COMMAND} -E env "MIR_CMAKE_BUILD_DIR=${CMAKE_BINARY_DIR}" 
          make -C ${CMAKE_SOURCE_DIR}/doc/sphinx run
```

doc/CMakeLists.txt
Saviq: Whee, this way we could drop the `MirAL` `configure_file` bit :tada:.

doc/CMakeLists.txt
Saviq: As is, it's unpleasant to use, and unnecessary IMO?
```
$ make doc-with-api
# ...
    Exception: Doxygen outputs at /home/michal/dev/canonical/mir/build-Debug/doc/sphinx/doxygen_output/xml don't exist
    Hint: Make sure MIR_CMAKE_BUILD_DIR points at the the configured Mir cmake project correctly
    If you've already ran doxygen, a directory named `doxyen_output` should exist inside the build directory
```

doc/CMakeLists.txt
```diff
@@ -1,3 +1,44 @@
...
+          make -C ${CMAKE_SOURCE_DIR}/doc/sphinx html 
+          BUILDDIR=${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build
+  DEPENDS ${CMAKE_SOURCE_DIR}/doc/sphinx/.sphinx/venv
+  BYPRODUCTS ${CMAKE_SOURCE_DIR}/doc/sphinx/_build
+)
+
+add_custom_target(doc-with-api
+  COMMAND ${CMAKE_COMMAND} -E env "MIR_BUILD_API_DOCS=1" 
+          ${CMAKE_COMMAND} --build ${CMAKE_BINARY_DIR} --target doc
+)
```
Saviq, line 37: Doesn't this also have `_build` as byproduct, or will that anger CMake to have two targets with the same byproduct?

In which case, should this be `add_custom_command`, rather than a target?

doc/CMakeLists.txt
```diff
@@ -1,3 +1,44 @@
...
+
+add_custom_target(doc-serve
```
Saviq: This also feels like a `_command`, and not a `_target`?

doc/CMakeLists.txt
```diff
@@ -1,3 +1,44 @@
...
+  COMMAND ${CMAKE_COMMAND} -E env "MIR_CMAKE_BUILD_DIR=${CMAKE_BINARY_DIR}" 
+          make -C ${CMAKE_SOURCE_DIR}/doc/sphinx serve
+          BUILDDIR=${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build
+  USES_TERMINAL
+)
+
+add_custom_target(doc-clean
```
Saviq: Ditto?

doc/CMakeLists.txt
```diff
@@ -10,47 +51,22 @@ endif()
...
-                 ${CMAKE_CURRENT_BINARY_DIR}/sphinx/xml
-  )
-
-  install(
-      DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build/
-      DESTINATION ${CMAKE_INSTALL_PREFIX}/share/doc/mir-doc
-      OPTIONAL
-  )
+  add_custom_target(doxygen-clean
+    COMMAND rm -r ${WORKING_DIR}/doxygen_output/xml ${CMAKE_SOURCE_DIR}/doc/sphinx/api
```
Saviq: I think we want to kill `doxygen_output` itself, too?
```suggestion
    COMMAND rm -r ${WORKING_DIR}/doxygen_output ${CMAKE_SOURCE_DIR}/doc/sphinx/api
```

doc/sphinx/conf.py
```diff
@@ -317,42 +311,53 @@
...
+                        "`doc/sphinx/conf.py`, or an absolute path")
+
+    doxygen_xml_output_dir = sphinx_dir / "doxygen_output/xml"
+    doxygen_xml_exists = os.path.exists(doxygen_xml_output_dir)
+
+    if build_api_docs and not doxygen_xml_exists:
+        raise Exception(f"Doxygen outputs at {doxygen_xml_output_dir} don't exist\n"
+                        "Hint: Make sure MIR_CMAKE_BUILD_DIR points at the the configured Mir cmake project correctly\n"
+                        "If you've already ran doxygen, a directory named `doxyen_output` should exist inside the build"
+                        " directory")
```
Saviq: I _think_ this can be flattened to a single error, along the lines of:
```suggestion
  try:
    doxygen_xml_output_dir = Path(os.environ["MIR_CMAKE_BUILD_DIR"]) / "doc/sphinx/doxygen_output/xml"
    assert doxygen_xml_output_dir.is_dir(), "Doxygen XML output not found"

  except KeyError, AssertionError as ex:
    raise Exception("MIR_CMAKE_API_DOCS is set, but couldn't find Doxygen output. "
                    "Make sure `MIR_CMAKE_BUILD_DIR` points at the CMake build directory "
                    "and that `doxygen_output` is populated inside\n"
                    f"{ex}") from ex
```

doc/sphinx/conf.py
```diff
@@ -317,42 +311,53 @@
...
+
+    extensions.extend(['exhale', 'breathe'])
+
+    # Setup the exhale extension
+    exhale_args = {
+        # These arguments are required
+        "containmentFolder": './api',
```
tarek-y-ismail: This still outputs to the source `doc/sphinx/api` directory, let's see if we can have it output into the build directory..

Saviq: I _think_ this one we might need to live with. Unless we can somehow give Sphinx an extra doc root?

tarek-y-ismail: Facepalm...
```
      File "/home/tarek/code-projects/mir/doc/sphinx/.sphinx/venv/lib/python3.13/site-packages/exhale/configs.py", line 1442, in apply_sphinx_configurations
        raise ConfigError(
        ...<3 lines>...
        )
    sphinx.errors.ConfigError: The given `containmentFolder` [/home/tarek/code-projects/mir/build/doc/sphinx/api] must be a *SUBDIRECTORY* of [/home/tarek/code-projects/mir/doc/sphinx].

```

tarek-y-ismail:
I think API docs are broken on local builds, probably because we moved the output. It still works on readthedocs because the outputs are pretty much in the same place as before. I'll try a completely clean build in case the problem is cause by some leftover files

tarek-y-ismail:
Ok, API docs _do work_. `doc-run` just doesn't have `MIR_BUILD_API_DOCS=1`. Time to rethink our targets

Saviq:
@tarek-y-ismail we probably need a `doc/README.md` talk about the targets available, btw.

tarek-y-ismail:
> @tarek-y-ismail we probably need a `doc/README.md` talk about the targets available, btw.

Seems like it's more fitting for contributor docs (when that lands)

Saviq:
@tarek-y-ismail so here are the last two paths that if we could, would be good to kick out:
```plain
        doc/sphinx/.sphinx/.doctrees/
        doc/sphinx/.sphinx/venv/
```

We might need to live with those.

But `doc/sphinx/api` we need in a `.gitignore` somewhere

doc/CMakeLists.txt
```diff
@@ -10,47 +38,43 @@ endif()
...
+            make -C ${CMAKE_SOURCE_DIR}/doc/sphinx html
+            BUILDDIR=${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build
+    DEPENDS ${CMAKE_SOURCE_DIR}/doc/sphinx/.sphinx/venv
+            ${WORKING_DIR}/doxygen_output/xml
+    BYPRODUCTS ${WORKING_DIR}/_build ${CMAKE_SOURCE_DIR}/doc/sphinx/api
   )
+
+  add_custom_target(doc-clean-all)
+  add_dependencies(doc-clean-all doc-clean doxygen-clean)
 endif()
```
Saviq:
```suggestion
  add_dependencies(doc-clean doxygen-clean)
endif()
```
Or even this?
```suggestion
  add_dependencies(clean doxygen-clean)
endif()
add_dependencies(clean doc-clean)
```

doc/CMakeLists.txt
```diff
@@ -1,3 +1,31 @@
+add_custom_command(
+  COMMAND make -C ${CMAKE_SOURCE_DIR}/doc/sphinx install
+  OUTPUT ${CMAKE_SOURCE_DIR}/doc/sphinx/.sphinx/venv
+)
+
+# The difference between the `VAR=<value> make <target>` and `make <target>
+# VAR=<value>` is the the former passes the variable down to child processes of
+# make. In this case, sphinx. Which is then read back when `conf.py` is
+# executed to determine where the build directory is to read doxygen's output.
+# The later only overrides variables in the makefile.
```
Saviq:
```suggestion
# In `VAR1=<value> make <target> VAR2=<value>, only `VAR1` is inherited by
# `make`'s children, `VAR2` only applies within `make`.
```

doc/CMakeLists.txt
```diff
@@ -10,47 +38,43 @@ endif()
...
-      ${CMAKE_CURRENT_SOURCE_DIR}/sphinx/Doxyfile.in
-      ${CMAKE_CURRENT_BINARY_DIR}/sphinx/Doxyfile @ONLY IMMEDIATE
-  )
+if(DOXYGEN_FOUND)
+  set(WORKING_DIR ${CMAKE_CURRENT_BINARY_DIR}/sphinx)
 
   configure_file(
-      ${CMAKE_CURRENT_SOURCE_DIR}/sphinx/reference/introducing_the_miral_api.md.in
-      ${CMAKE_CURRENT_BINARY_DIR}/sphinx/reference/introducing_the_miral_api.md @ONLY IMMEDIATE
+    ${CMAKE_SOURCE_DIR}/doc/sphinx/Doxyfile.in
```
Saviq: Slightly more portable:
```suggestion
    ${CMAKE_CURRENT_SOURCE_DIR}/sphinx/Doxyfile.in
```

doc/CMakeLists.txt
```diff
@@ -1,3 +1,31 @@
+add_custom_command(
+  COMMAND make -C ${CMAKE_SOURCE_DIR}/doc/sphinx install
+  OUTPUT ${CMAKE_SOURCE_DIR}/doc/sphinx/.sphinx/venv
```
Saviq: The `venv` path should be a variable, used so many times.

doc/CMakeLists.txt
```diff
@@ -1,3 +1,31 @@
...
+
+# The difference between the `VAR=<value> make <target>` and `make <target>
+# VAR=<value>` is the the former passes the variable down to child processes of
+# make. In this case, sphinx. Which is then read back when `conf.py` is
+# executed to determine where the build directory is to read doxygen's output.
+# The later only overrides variables in the makefile.
+add_custom_target(doc-run
+  COMMAND ${CMAKE_COMMAND} -E env "MIR_CMAKE_BUILD_DIR=${CMAKE_BINARY_DIR}"
+          make -C ${CMAKE_SOURCE_DIR}/doc/sphinx run
+          BUILDDIR=${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build
```
Saviq: `_build` path, too.

doc/CMakeLists.txt
```diff
@@ -10,47 +38,43 @@ endif()
...
+    COMMAND doxygen ${WORKING_DIR}/Doxyfile
+    OUTPUT ${WORKING_DIR}/doxygen_output/xml
   )
 
-  if(BUILD_DOXYGEN)
-    set(ALL "ALL")
-  endif()
+  add_custom_target(doxygen-clean
+    COMMAND rm -r ${WORKING_DIR}/doxygen_output ${CMAKE_SOURCE_DIR}/doc/sphinx/api
+  )
```
Saviq: Since we don't have a `doxygen` target, weird to have a `doxygen-clean` one?

Should we make this a `_command` that runs with the `clean` target?

Saviq: Alternatively, maybe we do actually want the `doxygen` target? As is, you don't have a way to refresh the API docs other than cleaning them?

doc/CMakeLists.txt
```diff
@@ -10,47 +38,43 @@ endif()
...
-              guides
-      BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build
-                 ${CMAKE_CURRENT_BINARY_DIR}/sphinx/xml
+  add_custom_target(doc-serve-with-api
+    COMMAND ${CMAKE_COMMAND} -E env "MIR_CMAKE_BUILD_DIR=${CMAKE_BINARY_DIR}"
+            make -C ${CMAKE_SOURCE_DIR}/doc/sphinx serve
+            BUILDDIR=${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build
+    DEPENDS ${CMAKE_SOURCE_DIR}/doc/sphinx/.sphinx/venv
+            ${WORKING_DIR}/doxygen_output/xml
+    BYPRODUCTS ${WORKING_DIR}/_build ${CMAKE_SOURCE_DIR}/doc/sphinx/api
```
Saviq: IMO this is one target too many to maintain… but I get why you're keeping it, so won't block.

I would `s/serve/run/`, though, as with the separate doxygen run it shouldn't be too annyoing, and one less distinction for our devs to think about?

doc/CMakeLists.txt
tarek-y-ismail: At that point we might as well run doxygen manually

doc/CMakeLists.txt
tarek-y-ismail: The different name implies a different function, which relates to the target in the Makefile. `serve` just builds and serves, it doesn't watch for changes. `run` builds, serves, _and_ watches for changes. It's a small thing, and I'm sure anyone would notice the API docs re-building on their machine (because they will run out of ram), but I think this is better

doc/CMakeLists.txt
```diff
@@ -1,3 +1,52 @@
+set(VENVDIR ${CMAKE_CURRENT_BINARY_DIR}/sphinx/venv)
+set(BUILDDIR ${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build)
+set(MAKEOPTS
+  VENVDIR=${VENVDIR}
+  BUILDDIR=${BUILDDIR}
+)
```
Saviq, line 6: We want `DOCTREESDIR` here now, right?

tarek-y-ismail: It's built off `BUILDDIR` anyway, so it's not strictly necessary

doc/sphinx/Makefile
```diff
@@ -5,7 +5,8 @@
 # You can set these variables from the command line, and also
 # from the environment for the first two.
 SPHINXDIR       = .sphinx
-SPHINXOPTS      ?= -c . -d $(SPHINXDIR)/.doctrees -j auto
+DOCTREESDIR     = $(BUILDDIR)/../.doctrees
```
Saviq: I think we shouldn't change the default? And going up the tree is risky. And `BUILDDIR` isn't defined yet.
```suggestion
DOCTREESDIR     ?= $(SPHINXDIR)/.doctrees
```

doc/CMakeLists.txt
Saviq: Since we shouldn't change the default (that we'll propose upstream), and b/c that's in `SPHINXDIR`, I think we should set it explicitly.

doc/CMakeLists.txt
```diff
@@ -1,3 +1,53 @@
+set(VENVDIR ${CMAKE_CURRENT_BINARY_DIR}/sphinx/venv)
+set(BUILDDIR ${CMAKE_CURRENT_BINARY_DIR}/sphinx/_build)
+set(MAKEOPTS
+  VENVDIR=${VENVDIR}
+  BUILDDIR=${BUILDDIR}
+  DOCTREESDIR=${BUILDDIR}/../.doctrees
```
Saviq: My OCD asks for…
```suggestion
  DOCTREESDIR=${CMAKE_CURRENT_BINARY_DIR}/sphinx/.doctrees
```

doc/CMakeLists.txt
```diff
@@ -1,3 +1,53 @@
+set(VENVDIR ${CMAKE_CURRENT_BINARY_DIR}/sphinx/venv)
```
Saviq:
```suggestion
```

doc/CMakeLists.txt
```diff
@@ -1,3 +1,53 @@
...
+  woke
+)
+
+foreach(target IN LISTS SPHINX_TARGETS)
+  add_custom_target(doc-${target}
+    # In `VAR1=<value> make <target> VAR2=<value>, only `VAR1` is inherited by
+    # `make`'s children, `VAR2` only applies within `make`.
+    COMMAND ${CMAKE_COMMAND} -E env "MIR_CMAKE_BUILD_DIR=${CMAKE_BINARY_DIR}"
+            make -C ${SPHINXDIR} ${target} ${MAKEOPTS}
+    DEPENDS ${VENVDIR}
```
Saviq:
```suggestion
```

doc/CMakeLists.txt
```diff
@@ -1,3 +1,53 @@
...
+  spelling
+  vale
+  woke
+)
+
+foreach(target IN LISTS SPHINX_TARGETS)
+  add_custom_target(doc-${target}
+    # In `VAR1=<value> make <target> VAR2=<value>, only `VAR1` is inherited by
+    # `make`'s children, `VAR2` only applies within `make`.
+    COMMAND ${CMAKE_COMMAND} -E env "MIR_CMAKE_BUILD_DIR=${CMAKE_BINARY_DIR}"
```
Saviq: Didn't I rename this?…
```suggestion
    COMMAND ${CMAKE_COMMAND} -E env "MIR_BUILD_DIR=${CMAKE_BINARY_DIR}"
```

doc/CMakeLists.txt
```diff
@@ -10,46 +60,37 @@ endif()
 
 find_package(Doxygen 1.8.0 ${REQUIRED})
 
-if(DOXYGEN_FOUND)
```
Saviq, line 13: Can get rid of the `if() ... "REQUIRED"` bit above (can't comment on unchanged lines…) and just `find_package` inside `if(BUILD_DOXYGEN)`.

doc/sphinx/.gitignore
```diff
@@ -15,6 +15,7 @@
 
 # Build outputs
 _build
+api
```
Saviq, line 18: This now has our custom `library_root.rst`… do we want to be more specific here and call out the autogenerated names/prefixes?
```
api/class*.rst
api/namespace*.rst
…
# etc.
```

doc/CMakeLists.txt
```diff
@@ -10,46 +60,37 @@ endif()
...
+  set(WORKING_DIR ${CMAKE_CURRENT_BINARY_DIR}/sphinx)
+  set(DOXYGEN_OUTPUT_DIR ${WORKING_DIR}/doxygen_output)
 
-  configure_file(
-      ${CMAKE_CURRENT_SOURCE_DIR}/sphinx/Doxyfile.in
-      ${CMAKE_CURRENT_BINARY_DIR}/sphinx/Doxyfile @ONLY IMMEDIATE
+  set(MAKEOPTS
+    ${MAKEOPTS}
+    MIR_BUILD_API_DOCS=1
   )
```
Saviq: My bad, this doesn't actually work… We need `MIR_BUILD_API_DOCS` in the environment, not in `MAKEOPTS`.

Saviq (changes requested):
Almost there…

I think we don't need the `venv` target, do we?

doc/CMakeLists.txt
Saviq: Maybe we simply just `export` it at the beginning of the Makefile?

doc/CMakeLists.txt
tarek-y-ismail: I don't think this is the best idea. Removing this would break `MAKEOPTS` below, and would break the custom command that calls `install`, requiring callers to first call `install` on the makefile

doc/CMakeLists.txt
tarek-y-ismail: That works for me

doc/sphinx/.gitignore
tarek-y-ismail: It's probably easier to ignore all and list the exceptions

doc/CMakeLists.txt
Saviq: Of course, we still need the variable… :facepalm:

>  requiring callers to first call `install` on the makefile

I don't think that's true? The `Makefile` has its own dependencies on `install`, no?

Saviq (changes requested):
`doc-clean` errors out, but also deletes `api/library_root.rst`…

doc/CMakeLists.txt
tarek-y-ismail: > I don't think that's true? The Makefile has its own dependencies on install, no?

Yuuup :facepalm:

tarek-y-ismail:
> doc-clean errors out, but also deletes api/library_root.rst…

I think what's happening is that `doxygen-clean` is executed first. Which removes all of `api/`, which is now handled by deleting everything _but_ that file. Then `doc-clean` is executed and fails because it has a test to make sure that `VENVDIR` is not an absolute path. It's not clear to me why this exists, so I worked around it temporarily by disabling the suspicious part of the check, and creating an issue (https://github.com/canonical/sphinx-docs-starter-pack/issues/435) to see if this is a legitimate bug or something else.

PR #4190, opened 2025-09-03
Remove the 'developing a wayland compositor using mir' how-to because it is redundant with the main tutorial + added a 'compositors built with mir' reference

mattkae:
## What's new?
- Removed 'Developing a Wayland Compositor Using Mir' as it is redundant with the tutorial. The tutorial is better :)
- Broke out a list of compositors that use Mir into a reference (plus removed mirco and egmde)


doc/sphinx/reference/compositors-built-with-mir.md
```diff
@@ -0,0 +1,20 @@
+(compositors-built-with-mir)=
+
+# Compositors Built with Mir
+
+This page is intended as a showcase of active projects that are using Mir.
+
+Here is the full list, linking to the source code:
```
robert-ancell: What does it mean to be a full list? Is this is all the compositors that use Mir? Or only those we know about or are promoting. Suggest rewording more like "Here is a list of active projects that are using Mir:"

doc/sphinx/how-to/developing-a-wayland-compositor-using-mir.md
AlanGriffiths, line 1: We should have a redirect for the old slug

AlanGriffiths (changes requested):

doc/sphinx/conf.py
```diff
@@ -168,7 +168,8 @@
 #       the sphinx_reredirects extension will be disabled.
 redirects = {
     'how-to/how-to-enable-graphics-core22-on-a-device': '../how-to-enable-graphics-for-snaps-on-a-device',
-    'explanation/ok-so-what-is-this-wayland-thing-anyway': '../../how-to/developing-a-wayland-compositor-using-mir'
+    'explanation/ok-so-what-is-this-wayland-thing-anyway': '../../tutorial/write-your-first-wayland-compositor',
+    'how-to/developing-a-wayland-compositor-using-mir.md*': '../../tutorial/write-your-first-wayland-compositor'
```
AlanGriffiths: Do we really want `.md*` here?

Saviq: No, wildcards don't seem to work (and the `.md` would've broke it anyway.
```suggestion
    'how-to/developing-a-wayland-compositor-using-mir': '../../tutorial/write-your-first-wayland-compositor'
```

PR #4192, opened 2025-09-04
Fix inconsistent capitalization of program option descriptions

robert-ancell:
Also reword --help description context as it may be shown in the documentation.


RAOF (approved):
Yup; most existing options follow sentence capitialisation.

AlanGriffiths (approved):

PR #4198, opened 2025-09-04
pre-commit: add initial config and workflow

Saviq:
## What's new?
This introduces a https://pre-commit.com/ config, and an action

## How to test
```
pip install pre-commit
pre-commit install
# proceed as normal
```

## TODO (for PR authors)
- [x] Tests added and pass

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,3 +23,4 @@ repos:
     - mdformat-beautysh
     - mdformat-gfm
     - mdformat-myst
+    - setuptools # Necessary until beautysh fix their stuff, see: https://github.com/hukkin/mdformat/issues/442


.pre-commit-config.yaml
```diff
@@ -0,0 +1,25 @@
+# Recommended workflow:
+# - `git add --verbose --patch` to stage changes
+# - `git commit` to commit them
+# - rinse and repeat, if pre-commit wants / applied changes
+
+files: ^doc/sphinx
+exclude: ^doc/sphinx/(\..*|Makefile|conf\.py)
```
tarek-y-ismail, line 7: Wouldn't it be easier to do:
```suggestion
files: ^doc/sphinx/*.md
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,25 @@
...
+  - id: end-of-file-fixer
+  - id: check-added-large-files
+
+- repo: https://github.com/executablebooks/mdformat
+  rev: 0.7.22
+  hooks:
+  - id: mdformat
+    additional_dependencies:
+    - mdformat-black
+    - mdformat-beautysh
```
tarek-y-ismail, line 23: Might cause issues in the future as it depends on `setuptools`, which will be removed soon:
```
/home/tarek/.cache/pre-commit/repozcpt3kr2/py_env-python3.13/lib/python3.13/site-packages/beautysh/__init__.py:9: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.py
pa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources  # part of setuptools
```

And seeing that mdformat-beautysh and beautysh haven't been updated in a couple of years, this might be an issue..

tarek-y-ismail (changes requested):

.pre-commit-config.yaml
Saviq: No, because it's not just the `.md` files we want to treat (the first three hooks apply to any file).

This just excludes the starter pack files, 'cause we don't want to maintain a diff.

.pre-commit-config.yaml
Saviq: Yeah let's run with it for now and we'll deal with it when it breaks. The alternative (`shfmt`) requires go or Docker, which is even more of a chore.

Saviq:
@tarek-y-ismail I dropped the CI action for now, 'cause it'd be annoying until we commit what it wants.

.pre-commit-config.yaml
tarek-y-ismail: Ok, I thought it was just markdown files, though I think this might require tightening later. But we won't know that until we try.

tarek-y-ismail:
> @tarek-y-ismail I dropped the CI action for now, 'cause it'd be annoying until we commit what it wants.

Yup, sure. It's still going to nag whoever installs the hook anyway :)

Saviq:
> Yup, sure. It's still going to nag whoever installs the hook anyway :)

Only about the things they're committing, fortunately.

AlanGriffiths:
> If one of your developers doesn’t have node installed but modifies a JavaScript file, pre-commit automatically handles downloading and building node to run eslint without root

Sometimes automation can seem a bit too eager to "help"

Saviq:
> Sometimes automation can seem a bit too eager to "help"

Fortunately we don't have JS files ;)

*We* choose the hooks we're happy with.

PR #4201, opened 2025-09-04
iDemonstrating how we can better use doxygen for append filter classes in miral

mattkae:


include/miral/miral/append_event_filter.h
```diff
@@ -26,13 +26,26 @@ namespace mir { class Server; }
...
+///
+/// \sa AppendKeyboardEventFilter
 class AppendEventFilter
 {
 public:
-    /// Append an event filter (after any existing filters, including the window manager).
-    /// The supplied filter should return true if and only if it handles the event as filters
-    /// later in the list will not be called.
-    /// \remark the filter return type changed to bool in MirAL 3.6
+    /// Constructs a new filter wrapper using the provided `filter` function.
```
robert-ancell: "new filter" -> "new event filter" ?

include/miral/miral/append_keyboard_event_filter.h
```diff
@@ -26,13 +26,29 @@ namespace mir { class Server; }
...
+///
+/// An event is passed to this filter only if no earlier filter has already handled it.
+///
+/// See also:
+/// - WindowManagementPolicy::handle_keyboard_event - the keyboard event handler for the
+///   window management policy.
+/// - AppendEventFilter - add a filter for any event, not just keyboards
+///
+/// \sa miral::WindowManagementPolicy
+/// \sa AppendEventFilter -
```
robert-ancell: Trailing '-'?

robert-ancell (approved):
Couple of small things but otherwise looks good.

include/miral/miral/append_event_filter.h
```diff
@@ -26,13 +26,26 @@ namespace mir { class Server; }
 
 namespace miral
 {
+/// Appends an event filter to the Mir server event pipeline.
+///
+/// Filters are processed in order. This filter is appended **after** any existing filters,
+/// including those provided by [miral::WindowManagementPolicy].
```
Saviq: This actually renders the square brackets, can we avoid that?

include/miral/miral/append_event_filter.h
```diff
@@ -26,13 +26,26 @@ namespace mir { class Server; }
...
+///
+/// An event is passed to this filter only if no earlier filter has already handled it.
+///
+/// See also:
+/// - AppendKeyboardEventFilter - a specialized event filter which only filters keyboard events
+///
+/// \sa AppendKeyboardEventFilter
```
Saviq: There's double "See also" in the result, we should decide on one way?
image

include/miral/miral/append_keyboard_event_filter.h
Saviq: I _think_ this was the beginning of reconciling _which_ See also style to use:

image

include/miral/miral/append_event_filter.h
mattkae: I was doing this because most IDEs won't see the doxygen-style "See also", but I guess it's not a big deal

mattkae: Let's just stick with the doxygen way :)

Saviq: > most IDEs won't see the doxygen-style "See also"

That sounds like a broken IDE to me… if anything, I would expect the Doxygen stuff to be parsable.

It is a bit heavy when there's more than one… but that's breathe/exhale's fault - maybe something we can fix when multiple `/sa`s exist atop one other.

Saviq (changes requested):
Since you're trying to establish a style here, should all references be monospace, or none of them?

It almost feels, again, like exhale/breathe should take care of it?

I mean, in the comments we should stick to what's common in doxygen, and later configure/fix the tooling to result in what we want to see rendered.

mattkae:
> Since you're trying to establish a style here, should all references be monospace, or none of them?
> 
> It almost feels, again, like exhale/breathe should take care of it?
> 
> I mean, in the comments we should stick to what's common in doxygen, and later configure/fix the tooling to result in what we want to see rendered.

The backtick style is purely stylistic as far as I can tell. The "better" style actually is `#MyClass::x` which tells doxygen explicitly to find and create a link. The backticks can optionally surround that

include/miral/miral/append_event_filter.h
mattkae: I agree, I wonder if there is something that we can do additionally to group them together

mattkae:
> > The backtick style is purely stylistic as far as I can tell
> > The backticks can optionally surround that
> 
> Right, I'd avoid that in the Doxygen comments, and if we'd like to have monospace in the rendered output, that's something to fix on the processing side.
> 
> > The "better" style actually is `#MyClass::x` which tells doxygen explicitly
> 
> Shouldn't be there in `\sa`: image

Yeah I guess `\sa` doesn't require them. Just wanted to see if it did

Saviq (approved):

PR #4203, opened 2025-09-05
Convert the C++ style guide to markdown

jhenstridge:
Closes MIRENG-1413

# What's new?
This converts the C++ style guide from it's custom XML format to Markdown, and adds it to the regular Sphinx documentation build.

The approach taken for the conversion was:

1. Modify the XSLT style sheet to generate simple HTML: no JavaScript, regular paragraphs, etc.
2. Use pandoc to convert that HTML to Markdown using the `commonmark_x` output format.
3. Make a small number of manual fixes to the result.

The first two steps were iterated a few times to try and get something that would be reasonable to treat as the primary version for future edits.


jhenstridge:
Here's the what the updated version looks like:

https://canonical-mir--4203.com.readthedocs.build/4203/reference/cppguide/

The one bit of formatting that was lost but would be nice to keep was the different background colours for the "good" vs "bad" code samples. I've left a `<!-- badcode -->` comment for the cases that were marked as bad in the original XML.

RAOF:
It looks like this slightly mangles some of the docs - for example:

```
sizeof(``varname{.variable}) instead of sizeof(``type{.variable})
```

I'll finish exhaustively reviewing this later, just flagging this for now.

jhenstridge:
> It looks like this slightly mangles some of the docs - for example:

This comes from the original as:

```xml
  <SUMMARY>
    Use <code>sizeof(<var>varname</var>)</code> instead of
    <code>sizeof(<var>type</var>)</code> whenever possible.
  </SUMMARY>
```

Which got converted to:

```md
Use `sizeof(``varname`{.variable}`)` instead of
`sizeof(``type`{.variable}`)` whenever possible.
```

That is, pandoc split it into three code elements, with an `{.variable}` attribute attached to the middle one. It's probably going to be easiest to strip out the `<var>` tags and re-export it.

PR #4204, opened 2025-09-05
Remove --enable-input

robert-ancell:
This option was added in 2013 in f864944fc3de3b6a4953c90189a503349f0e60a5 and defaulted to false at the time. It was later changed to default to true.

Removing as it doesn't seem useful and adds confusion to the active options. If no input is required it is probably better to use an input platform that provides this behaviour.

### DIFF
--- a/include/platform/mir/options/configuration.h
+++ b/include/platform/mir/options/configuration.h
@@ -28,7 +28,6 @@ class SharedLibrary;
 namespace options
 {
 extern char const* const arw_server_socket_opt;
-extern char const* const enable_input_opt;
 extern char const* const shared_library_prober_report_opt;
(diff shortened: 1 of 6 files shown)


robert-ancell:
I think the modern equivalent is `--platform-input-lib mir:stub-input`

Saviq:
@robert-ancell should we close this? Log a reminder work item post 26.04?

robert-ancell:
> @robert-ancell should we close this? Log a reminder work item post 26.04?

Yeah, should be closed. I wasn't sure the best way to track this - I'll open an item.

robert-ancell:
https://warthogs.atlassian.net/browse/MIRENG-1441

PR #4207, opened 2025-09-05
Drop workarounds for link failure with g++6. (We can't compile with it anyway)

AlanGriffiths:
We don't need this now


AlanGriffiths:
Phew! I saw a lot of red and thought I broke it...

...but "Unable to connect to archive.ubuntu.com" doesn't seem like my doing

Saviq (approved):
If it blends! :man_shrugging:

tests/CMakeLists.txt
```diff
@@ -49,9 +49,6 @@ if (HAS_W_INFINITE_RECURSION)
   add_compile_options(-Wno-error=infinite-recursion)
 endif()
 
-set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-lto")
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-lto")
-
```
RAOF, line 54: Wasn't this here because we are shipping a static test archive for downstreams and didn't want to ship LTO bytecode in there (requiring them to have a compatible toolchain?)

AlanGriffiths: Fixed

PR #4209, opened 2025-09-05
Provide a custom library_root file for doxygen

mattkae:
# What's new?
* Removed a bunch of cruft that was added by the default library root file


Saviq:
Wonder if we actually just want the Class view hierarchy?

mattkae:
> Wonder if we actually just want the Class view hierarchy?

Yeah I wonder that too. Would we accidentally lose something there you think? Maybe some functions?

PR #4210, opened 2025-09-05
MIRENG-1229-High-level-architecture

AlanGriffiths:
# What's new?
* Introduce a consumer-facing architecture doc
  * "Explanation"/"User libraries"
* Refresh the diagrams and text for existing architecture and libraries docs
  * "Contributing to Mir/"Explanation"/"Architecture"
  * "Contributing to Mir/"Explanation"/"Libraries"

### DIFF
--- a/doc/sphinx/explanation/consumer-libraries.md
+++ b/doc/sphinx/explanation/consumer-libraries.md
@@ -2,7 +2,7 @@
 These are the principle libraries *Mir* provides compositor authors.
 
 These libraries will not break ABI or API between minor versions of Mir.
-(We do not provide the same guarantees for other, internally used, libraries.) 
+We do not provide the same guarantees for other, internally used, libraries. 
 
 ## APIs for compositor authors
 ```{mermaid} consumer-libraries.mmd


doc/sphinx/explanation/consumer-libraries.md
```diff
@@ -0,0 +1,24 @@
+# User libraries
+These are the principle libraries *Mir* provides compositor authors.
+
+These libraries will not break ABI or API between minor versions of Mir.
+(We do not provide the same guarantees for other, internally used, libraries.)
```
mattkae:
```suggestion
We do not provide the same guarantees for other, internally used, libraries.
```

doc/sphinx/explanation/consumer-libraries.md
```diff
@@ -0,0 +1,24 @@
...
+```{mermaid} consumer-libraries.mmd
+```
+
+*Mir* provides compositor authors with a set of libraries that they can use to build Wayland based shells. These libraries are:
+
+Library | Description
+--      | --
+*miral*|The most commonly used library is **miral** (the "*Mir* Abstraction Layer"). This provides an API that makes *Mir* easy for compositor authors to work with. It provides core window management concepts and an interface that is more ABI stable than Mir's internal APIs.
+*mirwayland*|Compositor authors may want to define their own wayland protocol extensions in addition to the ones that the core *Mir* implementation defines. The **mirwayland** library satisfies this requirement. This library may be used in conjunction with either `miral` or `miroil`.
+*miroil*|A custom library for the [Lomiri](https://lomiri.com/) folks. It is like **miral** in that it provides an abstraction over Mir's internal APIs. _Other compositor authors should not need to interact with **miroil**._
```
mattkae: We shouldn't mention `miroil` in this document in my opinion. While it's a consumer library, it is really only intended for a single consumer who understands that it's special for them. I think that it isn't worthwhile to put it right in regular people's faces

mattkae: If we drop it here, then we can also remove it from the diagram

mattkae (changes requested):
I have a nit about including `miroil` as a "consumer" library. I understand that it lives at that level in the code, but I think that it's not worthwhile to mention it for the average developer

PR #4217, opened 2025-09-08
Use the exhale class view over the unabriged api view

mattkae:
I think Saviq was right, this is nicer


Saviq (approved):
Yup, let's run with it and see what else do we want.

PR #4212, opened 2025-09-08
Update argument descriptions for consistency and clarity

robert-ancell:
Consistent capitalization and style.
Describe features in more detail.
Update formatting to more easily work in other domains, e.g. markdown.

PR #4219, opened 2025-09-09
Automatically generate command line documentation

robert-ancell:
Run `--help-markdown` on miral-shell and write it into `docs/`. Checked in CI - if a change has occured a diff is shown and the author needs to update the docs.


.github/workflows/command-line-docs.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo --preserve-env apt-add-repository --yes ppa:mir-team/dev
+          sudo apt-get install --yes libclang1-18 python3-clang-18
+          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH=/usr/lib/llvm-18/lib/libclang-18.so.1" >> $GITHUB_ENV
+          echo "MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-18/lib" >> $GITHUB_ENV
```
jhenstridge: Since these are constant, might it be simpler to use the [`env:` stanza at the job level](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idenv)?

src/platform/options/default_configuration.cpp
```diff
@@ -340,6 +341,60 @@ auto mo::DefaultConfiguration::options_for(SharedLibrary const& module) const ->
...
+            escaped_text += "&quot;";
+            break;
+        case '\'':
+            escaped_text += "&apos;";
+            break;
+        default:
+            escaped_text += c;
+            break;
+        }
+    }
```
jhenstridge, line 373: It's probably not necessary to escape quote characters here, and there are other characters that have special meaning in markdown that maybe should be escaped if the description is intended to be plain text.

It might also be simpler to do backslash escaping of the characters in this list:

https://www.markdownguide.org/basic-syntax/#escaping-characters

You'd then be using the same form of escape for all special characters.

src/platform/options/default_configuration.cpp
```diff
@@ -340,6 +341,60 @@ auto mo::DefaultConfiguration::options_for(SharedLibrary const& module) const ->
...
+
+    return escaped_text;
+}
+
+std::string options_to_markdown(const boost::program_options::options_description& desc)
+{
+    std::ostringstream text;
+    for (auto& o: desc.options())
+    {
+        text << "### `" << o->format_name();
```
jhenstridge, line 383: Might it be worth formatting the options as a definition list?

https://mystmd.org/guide/typography#definition-lists

AlanGriffiths (commented):
An additional `mir-command-line-doc` example is going to confuse anyone looking at our examples. Does it give us anything that the existing examples don't already provide?

robert-ancell:
> An additional `mir-command-line-doc` example is going to confuse anyone looking at our examples. Does it give us anything that the existing examples don't already provide?

Whoops, I had meant to remove that. I was looking into making a binary to generate the command line flags, but then decided that miral-shell was good enough at least as a starting point.

src/platform/options/default_configuration.cpp
robert-ancell: I want to keep some of the markdown from the text, for example in https://github.com/canonical/mir/pull/4212 I intentionally changes some lists to be in markdown format. Note I was looking for an existing method to do this escaping but ended up having to write one.

src/platform/options/default_configuration.cpp
robert-ancell: I was playing around with the formatting but decided in the end to just do it the simplest way - I think it could be improved but this is better done as an enhancement later.

src/platform/options/default_configuration.cpp
jhenstridge: Okay. The `[^SCALE]` bit in the `--x11-output` docs gets syntax highlighted Github's pretty printing, as it looks like a possible footnote reference.

src/platform/options/default_configuration.cpp
```diff
@@ -340,6 +341,60 @@ auto mo::DefaultConfiguration::options_for(SharedLibrary const& module) const ->
...
+        if (!o->format_parameter().empty())
+        {
+            text << " " << o->format_parameter();
+        }
+        text << "`" << std::endl;
+        text << std::endl;
+        text << markdown_escape(o->description()) << std::endl;
```
AlanGriffiths: It is needlessly inefficient to flush the output for every line:
```suggestion
        text << "`\n\n";
        text << markdown_escape(o->description()) << '\n';
```

src/platform/options/default_configuration.cpp
```diff
@@ -405,6 +462,25 @@ void mo::DefaultConfiguration::parse_arguments(
...
+            help_text << "# Mir command line options" << std::endl;
+            help_text << std::endl;
+            help_text << options_to_markdown(desc);
+            for (auto& platform : platform_libraries)
+            {
+                auto& module_desc = module_options_desc.at(platform->get_handle());
+                if (!module_desc.options().empty())
+                {
+                    help_text << "## " << option_header_for(*platform) << std::endl;
+                    help_text << std::endl;
```
AlanGriffiths: Too much `std::endl` here too

AlanGriffiths:
> I can see this being the basis of an _explanation_ of the configuration options

Or as part of a configuring _tutorial_...

keirthana:
> Stepping back a bit: I'm not sure this belongs as a consumer "reference" - the configuration options (which are **not** usually supplied on the command-line) are specific to the compositor they are working on.
> 
> I can see this being the basis of an _explanation_ of the configuration options using `miral-shell` as a worked example, but it needs some additional explanatory text to make that work.

I agree with the overall sentiment but this is to have a quick reference of commands available in the documentation. We have that in every product - Juju, LXD etc. Mir may be slightly different because of the multiple audience types but I believe this is still useful to have in there. The command reference can reside in the same user classification as the explanation (if we write one in the future)

AlanGriffiths:
> I agree with the overall sentiment but this is to have a quick reference of commands available in the documentation. We have that in every product - Juju, LXD etc. Mir may be slightly different because of the multiple audience types but I believe this is still useful to have in there. The command reference can reside in the same user classification as the explanation (if we write one in the future)

Mir is different here, not because of the different audience types but, because compositors built with it are not normally run as command-line applications.

So, while _technically_ these options are are available on the command line, that is not how they should be used or documented.

With Ubuntu Frame, these options _cannot be supplied on the command line_, but only via Snap configuration.

With Miriway, all the configuration mentioned in this PR (and more) is supplied via `~/.config/miriway-shell.config`, _not via the command line_.

<details>
<summary>cat ~/.config/miriway-shell.config </summary>

```
x11-window-title=Miriway
idle-timeout=600
touchpad-tap-to-click=true
app-env-amend=XDG_SESSION_TYPE=wayland:GTK_USE_PORTAL=0:XDG_CURRENT_DESKTOP=Miriway:GTK_A11Y=none:-GTK_IM_MODULE:SSH_AUTH_SOCK=/run/user/1623796318/keyring/ssh
shell-component=miriway-unsnap systemd-run --user --scope --slice=background.slice synapse --startup
shell-component=systemd-run --user --scope --slice=background.slice swaybg --mode fill --output '*' --image /usr/share/backgrounds/warty-final-ubuntu.png
shell-component=systemd-run --user --scope --slice=background.slice swaync
shell-component=systemd-run --user --scope --slice=background.slice yambar
shell-component=miriway-unsnap systemd-run --user --scope --slice=background.slice gnome-keyring-daemon --foreground

ctrl-alt=t:miriway-unsnap kgx
shell-meta=a:miriway-unsnap synapse
meta=Print:miriway-unsnap sh -c "grim ~/Pictures/screenshot-$(date --iso-8601=seconds).png"

shell-ctrl-alt=l:miriway-unsnap loginctl lock-session
lockscreen-app=miriway-unsnap swaylock -i /usr/share/backgrounds/warty-final-ubuntu.png

ctrl-alt=d:cp /home/alan.griffiths@canonical.com/.config/miriway-shell.display~docked /home/alan.griffiths@canonical.com/.config/miriway-shell.display
ctrl-alt=u:cp /home/alan.griffiths@canonical.com/.config/miriway-shell.display~undocked /home/alan.griffiths@canonical.com/.config/miriway-shell.display
ctrl-alt=s:miriway-swap
ctrl-alt=Up:@toggle-always-on-top
shell-ctrl-alt=y:systemd-run --user --scope --slice=background.slice yambar

meta=Left:@dock-left
meta=Right:@dock-right
meta=Space:@toggle-maximized
meta=Home:@workspace-begin
meta=End:@workspace-end
meta=Page_Up:@workspace-up
meta=Page_Down:@workspace-down
ctrl-alt=BackSpace:@exit

display-config=static=/home/alan.griffiths@canonical.com/.config/miriway-shell.display
composite-delay=60
platform-display-libs=mir:atomic-kms
#shell-component=miriway-unsnap lxqt-policykit-agent
shell-component=miriway-unsnap /usr/libexec/polkit-mate-authentication-agent-1
```
</details>

The idea of automatically documenting these options in markdown is attractive, but calling this "command line documentation" is misleading. Also, as this example illustrates, Miriway adds a lot of its own options that are not in `miral-shell` (nor any other of the Mir examples).

I'd love to use this functionality to auto-generate Miriway documentation downstream but here it is a bad fit and a bad name. At the very least it should be called "Configuration Options".

keirthana:
> > I agree with the overall sentiment but this is to have a quick reference of commands available in the documentation. We have that in every product - Juju, LXD etc. Mir may be slightly different because of the multiple audience types but I believe this is still useful to have in there. The command reference can reside in the same user classification as the explanation (if we write one in the future)
> 
> Mir is different here, not because of the different audience types but, because compositors built with it are not normally run as command-line applications.
> 
> So, while _technically_ these options are are available on the command line, that is not how they should be used or documented.
> 
> With Ubuntu Frame, these options _cannot be supplied on the command line_, but only via Snap configuration.
> 
> With Miriway, all the configuration mentioned in this PR (and more) is supplied via `~/.config/miriway-shell.config`, _not via the command line_.
> 
> cat ~/.config/miriway-shell.config
> The idea of automatically documenting these options in markdown is attractive, but calling this "command line documentation" is misleading. Also, as this example illustrates, Miriway adds a lot of its own options that are not in `miral-shell` (nor any other of the Mir examples).
> 
> I'd love to use this functionality to auto-generate Miriway documentation downstream but here it is a bad fit and a bad name. At the very least it should be called "Configuration Options".

Thank you for explaining, I understand the need for moving to a better title, Configuration Options. We can reflect what we offer correctly on the documentation. I think it is worth documenting these options if we require users to know which option to use in a particular scenario. Even if it means adding the options to a config file and loading it. If users are not required to use or update the granular details of the options, then we don't have to spend too much effort into this.

Saviq:
Since this just prints Markdown, maybe, rather than focusing on command line switches, we use this to describe the available options and explain how they can be provided?

In pseudo-doc:
```
Any of the options detailed below can be provided through the following means (in order of precedence):

# in a config file
option_key=value             # true/false for switches

# in the environment
MIR_SERVER_OPTION_KEY=value  # true/false for switches

# on the command line
--option_key                 # for switches
--option_key=value

## here goes the description of all the available options
```

AlanGriffiths:
> Since this just prints Markdown, maybe, rather than focusing on command line switches, we use this to describe the available options and explain how they can be provided?

Yes, that addresses my issue with the naming.

----

The other issue is that this cannot, in the Mir project, be a definitive list and _that needs to be clear to anyone using it_. Each downstream will have a different selection of our and bespoke options. There will be options here that won't work in Miriway and options in Miriway that are not listed here, (And the same with Frame and miracle-wm)

But the ability to auto-generate this documentation is great and I would want to use it in both Miriway and Frame

Saviq:
> There will be options here that won't work in Miriway

Can we build a minimal server just for the purpose of having a definitive list that _does_ work in any Mir-based compositor?

And then another list of "Downstream compositors may or may not also have…"?

> But the ability to auto-generate this documentation is great and I would want to use it in both Miriway and Frame

Exactly!

.github/workflows/command-line-docs.yml
robert-ancell: Removed

src/platform/options/default_configuration.cpp
robert-ancell: Fixed by putting inside ticks

AlanGriffiths:
> Can we build a minimal server just for the purpose of having a definitive list that _does_ work in any Mir-based compositor?

Our "hello world" example is such a server, and we now build it in the docs pass. (Not quite a "definitive list" as platform options _could_ vary with the available platforms.)

> And then another list of "Downstream compositors may or may not also have…"?

We'd need a compositor that is guaranteed to expose every option - we don't have that at present and maintaining that would be error prone.

As I said above, we're probably better providing an example of how to generate the docs for a downstream compositor.

Saviq:
> Trying to diagnose CI

In case it helps, we've sprinkled this on our jobs - lets you restart a failing job with debugging on and get a shell into it:
https://github.com/canonical/mir/blob/58e00e86d4fee20a1a5794da4dbb526fab7e698b/.github/workflows/spread.yml#L123-L127

robert-ancell:
https://github.com/canonical/mir/pull/4278 replaces this with JSON which can then be used by Frame to generate documentation.

PR #4220, opened 2025-09-09
Make the "Write your first Wayland compositor" example compositor project compilable

tarek-y-ismail:
# What's new?
- Modifies `write-your-first-wayland-compositor.md` to pull the `CMakeLists.txt` and `main.cpp`
- Adds a spread task (that hopefully works) to install dependencies (should be pulled out) and build the example compositor project.

### DIFF
--- a/doc/sphinx/tutorial/first-wayland-compositor/CMakeLists.txt
+++ b/doc/sphinx/tutorial/first-wayland-compositor/CMakeLists.txt
@@ -5,11 +5,8 @@ project(demo-mir-compositor)
 set(CMAKE_CXX_STANDARD 23)
 
 include(FindPkgConfig)
-pkg_check_modules(MIRAL miral REQUIRED)
-pkg_check_modules(XKBCOMMON xkbcommon REQUIRED)
+pkg_check_modules(MIRAL miral REQUIRED IMPORTED_TARGET)
 
 add_executable(demo-mir-compositor main.cpp)
 
-target_include_directories(demo-mir-compositor PUBLIC SYSTEM ${MIRAL_INCLUDE_DIRS})
-target_link_libraries(demo-mir-compositor ${MIRAL_LDFLAGS})
-target_link_libraries(demo-mir-compositor ${XKBCOMMON_LIBRARIES})
+target_link_libraries(demo-mir-compositor PkgConfig::MIRAL)


doc/sphinx/tutorial/first-wayland-compositor/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.5)
```
Conan-Kudo: This should be set to 3.12, since CMake is deprecating everything below that progressively over the next few versions.

doc/sphinx/tutorial/first-wayland-compositor/write-your-first-wayland-compositor.md
```diff
@@ -24,8 +24,8 @@ The example program requires
   different graphics drivers
 
 ### Install dependencies on Debian or derivatives:
-```sh
-sudo apt install libmiral-dev mir-graphics-drivers-desktop
+```{literalinclude} install-dependencies.sh
+:language: bash
```
Conan-Kudo: Could we have Fedora and Ubuntu instructions?

For Fedora, this is `sudo dnf install 'pkgconfig(miral)'` (or `sudo dnf install mir-devel` if you want to use the direct package name).

spread/build/foo/task.yaml
```diff
@@ -0,0 +1,16 @@
+systems: [ubuntu-*]
+variants:
+- -ubuntu*
+- -debian*
```
Conan-Kudo: It'd be great to also have Fedora builds too. 😄 

(Also, should it _really_ be called "foo" in spread? Probably a better name can be used here...)

doc/sphinx/tutorial/first-wayland-compositor/write-your-first-wayland-compositor.md
tarek-y-ismail: Debian instructions will work for Ubuntu as well (since it's a derivative). Fedora instructions are a few lines below

spread/build/foo/task.yaml
tarek-y-ismail: Will test everything that spread and our infra allows. The "foo" name placeholder while I poke at the code ;)

tarek-y-ismail (commented):
Hey @Conan-Kudo, thanks for the initial review! I'm still poking at the code so expect a lot of improvements

Saviq (changes requested):
We could keep the URL as is, just `literalinclude` from a subdirectory?

You'll need to add a line to `.github/workflows/spread.yml` for this to run in CI.

tarek-y-ismail:
I want to keep the test (almost) identical to what users will do. If that means we have to maintain a list of dependencies until we figure out a way to automate this on different distros, then I'm okay with that.

spread/first-compositor-example/fedora/task.yaml
```diff
@@ -0,0 +1,21 @@
+systems: [fedora-*]
+variants: [amd64]
+
+summary: Build Mir Example
+
+execute: |
+    cd $SPREAD_PATH
+    dnf install cmake clang pkgconf glib2-devel -y
```
Conan-Kudo: Why clang instead of gcc?

spread/first-compositor-example/fedora/task.yaml
```diff
@@ -0,0 +1,33 @@
...
+        dnf --assumeyes install clang
+    else
+        dnf --assumeyes install gcc
+    fi
+    dnf install cmake pkgconf glib2-devel -y
+
+    cd doc/sphinx/tutorial/first-wayland-compositor
+
+    # [doc:first-compositor:fedora-dependencies-install]
+    dnf install mir-devel libxkbcommon-devel -y
```
Conan-Kudo: Submitted a PR to fix an undeclared dependency in the pkgconfig file for mircommon: https://github.com/canonical/mir/pull/4240

Saviq:
> I want to keep the test (almost) identical to what users will do. If that means we have to maintain a list of dependencies until we figure out a way to automate this on different distros, then I'm okay with that.

What I'm thinking is adding something [like this](https://unix.stackexchange.com/questions/87130/how-to-quickly-create-a-local-apt-repository-for-random-packages-using-a-debian/595448#595448) to `sbuild/task.yml` (I'm testing what can be done now):
```shell
apt-ftparchive packages $built_debs_path > Packages
add-apt-repository "deb [trusted=yes] …"
```
And then we can follow as you had it, with packages from the current checkout. That would have to be adapted per platform, of course.

Conan-Kudo:
Yeah, the equivalent can be done for Fedora using `createrepo_c`:

```bash
createrepo $abs_built_rpms_path
dnf config-manager addrepo --repoid=mirpr --set=baseurl="file://$abs_built_rpms_path" --set=gpgcheck=0 --set=priority=90 --set=enabled=1 --save-filename=mirpr.repo
```

spread/build/sbuild/task.yaml
```diff
@@ -100,3 +100,41 @@ execute: |
...
+      apt-ftparchive packages . > Packages
+      echo "deb [trusted=yes] file:$PWD ./" >> /etc/apt/sources.list
+      echo "APT::Get::Assume-Yes true;" >> /etc/apt/apt.conf.d/99assume-yes
+      apt-get update
+
+      export \
+        MIR_SERVER_PLATFORM_DISPLAY_LIBS=mir:virtual \
+        MIR_SERVER_VIRTUAL_OUTPUT=800x600
+
+      # [doc:first-compositor:debian-dependencies-install]
```
tarek-y-ismail: Need to update path in the tutorial

tarek-y-ismail:
Sorry @Conan-Kudo, I was a bit too ambitious with including Fedora (and Alpine). We'll come back to them in the future :crossed_fingers:

doc/sphinx/tutorial/first-wayland-compositor/write-your-first-wayland-compositor.md
Saviq: Working on this in a separate PR :+1:

PR #4234, opened 2025-09-09
Add documentation for miral::DisplayConfiguration

mattkae:


include/miral/miral/display_configuration.h
```diff
@@ -61,12 +74,56 @@ class DisplayConfiguration
...
+        /// \param f function to all on each child node
         void for_each(std::function<void(Node const&)> const& f) const;
+
+        /// Check if the node has a value at the given key.
+        ///
+        /// If the node does not have a #type of #Type::map, this will
+        /// cause a fatal error.
+        ///
+        /// \param key the key to check for existence
+        /// \returns `true` if the key exists, otherwise `false`
```
robert-ancell, line 113: Nit: the 'otherwise false' seems redundant - if it's boolean then this is implied.

include/miral/miral/display_configuration.h
```diff
@@ -61,12 +74,56 @@ class DisplayConfiguration
...
         auto has(std::string const& key) const -> bool;
+
+        // Get the node at the given key.
```
robert-ancell: Missing '/'

include/miral/miral/display_configuration.h
```diff
@@ -81,35 +138,74 @@ class DisplayConfiguration
...
+    /// \returns the list of valid layouts
     auto list_layouts() -> std::vector<std::string>;
 
-    /// Enable a custom output attribute in the .display YAML
+    /// Enable a custom output attribute in the configuration file.
+    ///
+    /// The value of this attribute will be made available through
+    /// #miral::UserDisplayConfigurationOutput::custom_attribute at the
+    /// \p key. Note that the data is value for the key has to have a
+    /// type of `std::optional<std::string>`.
```
robert-ancell: "Note that the data is value for the key has to have a type of" - not sure what this means? "that is the value for the key"?

robert-ancell (approved):
Small stuff but otherwise looks good.

include/miral/miral/display_configuration.h
```diff
@@ -31,23 +31,36 @@ namespace miral
...
+/// `$XDG_CONFIG_HOME` or `$HOME/.config`. Please see this default file for more
+/// information about content expected from this project.
+///
+/// Mir will continue to monitor the configuration file as it runs.
+///
+/// Additionally, the selected layout of the file may be overridden using a corresponding
+/// file: miral::MirRunner::display_config_file + "-layout", which will also be monitored.
+///
 /// \remark Since MirAL 2.4
-/// \note From MirAL 3.8 will monitor the configuration file or, if none found,
```
AlanGriffiths: It seems inconsistent to retain the "Since MirAL 2.4" but drop the change in behavior since MirAL 3.8

include/miral/miral/display_configuration.h
```diff
@@ -31,23 +31,36 @@ namespace miral
...
+/// Once a file is found, the remaining directories are not searched.
+///
+/// If no file is found, a default display configuration will be written to either
+/// `$XDG_CONFIG_HOME` or `$HOME/.config`. Please see this default file for more
+/// information about content expected from this project.
+///
+/// Mir will continue to monitor the configuration file as it runs.
+///
+/// Additionally, the selected layout of the file may be overridden using a corresponding
+/// file: miral::MirRunner::display_config_file + "-layout", which will also be monitored.
```
AlanGriffiths:
```suggestion
/// file: miral::MirRunner::display_config_file() + "-layout", which will also be monitored.
```

include/miral/miral/display_configuration.h
AlanGriffiths: Unfortunately, it is ambiguous whether or not this is implied. Clarity can also be achieved with "if, and only if, ..." (which is more common in my experience).

[rant= on] When I did a mathematics degree "if, and only if," was abbreviated to "iff" - a shame that never got wider adoption:

```suggestion
        /// \returns `true` iff the key exists
```

mattkae: I will leave as-is for now. I'm familiar with `iff`, but I can see how people would assume it's a misspell :)

include/miral/miral/display_configuration.h
```diff
@@ -173,8 +171,8 @@ class DisplayConfiguration
     ///
     /// The value of this attribute will be made available through
     /// #miral::UserDisplayConfigurationOutput::custom_attribute at the
-    /// \p key. Note that the data is value for the key has to have a
-    /// type of `std::optional<std::string>`.
+    /// \p key. Note that the value at the \p key must have a type of
+    /// `std::optional<std::string>`.
```
AlanGriffiths: I can guess, but the language is entirely unclear what the "Note that the value at the `key` must have a type of `std::optional<std::string>`" requirement means to the calling code.

If the reference is to types in the `.yaml`, then use of C++ types is inappropriate

mattkae: Agreed, I updated.

PR #4238, opened 2025-09-10
RFC: Drop our "internal" apis from consumer docs

AlanGriffiths:
We're showing the "MIr C++ API" documentation as a consumer reference, so it should focus on the APIs presented by libmiral-dev, libmircommon-dev and libmircore-dev.  (And, maybe, libmirwayland-dev)

### DIFF
--- a/doc/sphinx/Doxyfile.in
+++ b/doc/sphinx/Doxyfile.in
@@ -856,11 +856,9 @@ INPUT                  = @CMAKE_SOURCE_DIR@/README.md \
                          @CMAKE_SOURCE_DIR@/HACKING.md \
                          @CMAKE_SOURCE_DIR@/doc \
                          @CMAKE_SOURCE_DIR@/include/core \
+                         @CMAKE_SOURCE_DIR@/include/common \
                          @CMAKE_SOURCE_DIR@/include/miral \
-                         @CMAKE_SOURCE_DIR@/include/miroil \
-                         @CMAKE_SOURCE_DIR@/include/server \
-                         @CMAKE_SOURCE_DIR@/include/platform \
-                         @CMAKE_SOURCE_DIR@/src/include \
+                         @CMAKE_SOURCE_DIR@/include/mirwayland \
                          @MIR_GENERATED_INCLUDE_DIRECTORIES_FLATTENED@
 
 # This tag can be used to specify the character encoding of the source files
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
I'm not against generating docs for more of our classes elsewhere, but we shouldn't be overloading downstream developers with irrelevant (to them) details.

Saviq:
> we shouldn't be overloading downstream developers with irrelevant (to them) details.

I agree with that, but I don't think just removing them is the way to go. The index page is quite useless as-is. I think we need dedicated, curated pages that serve as entry points into the API documentation for different consumers.

AlanGriffiths:
> I agree with that, but I don't think just removing them is the way to go. The index page is quite useless as-is. I think we need dedicated, curated pages that serve as entry points into the API documentation for different consumers.

Please don't characterize this as "just removing them", they have only been added recently and are not yet on "stable".

It isn't like reverting this would be hard when we can manage the API documentation better.

While I agree that the API page could be improved a lot, dumping in a lot of extra material (that often lacks doc comments) makes things much worse.

Unless "dedicated, curated pages that serve as entry points into the API documentation" are coming real soon, I think we should make the consumer focused documentation as good as possible.

mattkae:
> I agree with that, but I don't think just removing them is the way to go. The index page is quite useless as-is. I think we need dedicated, curated pages that serve as entry points into the API documentation for different consumers.

I think that we can (slowly) make the existing doxygen way of doing things be the go-to for all of our docs, in a similar way to how the Flutter project does it. This will obviously take some time though :)

In the meantime, I agree with Alan that these docs should not be grouped in with the consumer-facing doxygen info. Perhaps we add another build for them and organize their namespace in to the "contributing" section, as no average user will want to interact with them.

Saviq:
> Please don't characterize this as "just removing them"

The PR title says "Drop" ;)

> Perhaps we add another build for them and organize their namespace

That's what @jhenstridge was trying to do, but the tooling only allows for a single Doxygen project.

There is [grouping](https://www.doxygen.nl/manual/grouping.html), which might do something, but it's manual (e.g. you'll have to remember to put things in the appropriate group) - and it's spread throughout the project.

That's why I'm thinking we might need to take the generated reference page and split it apart, and maintain it ourselves, manually. We could add checks that at least notify us when the origin gets updated…

I just worry that if we now sweep it (back) under the rug, it's going to stay there.

mattkae:
> That's why I'm thinking we might need to take the generated reference page and split it apart, and maintain it ourselves, manually. We could add checks that at least notify us when the origin gets updated…
> 
> I just worry that if we now sweep it (back) under the rug, it's going to stay there.

Perhaps manually "publicizing" the classes that we care about (instead of leaning on exhale) is a good idea then.

AlanGriffiths:
> Perhaps manually "publicizing" the classes that we care about (instead of leaning on exhale) is a good idea then.

We don't currently have a perfect (or even good) solution to "publicizing" the classes that we care about.

@jhenstridge made some improvements but the current tooling is inflexible and we're left with issues. The thing I'm questioning is what is least bad right now:

1. having good quality, consumer facing content badly organized; or,
2. having lots of dubious quality, unfocused content badly organized?

> I just worry that if we now sweep it (back) under the rug, it's going to stay there.

We are far from an end state on publishing docs - we're not going to forget there are problems.

mattkae:
> 1. having good quality, consumer facing content badly organized; or,
>  2. having lots of dubious quality, unfocused content badly organized?

I vote for 1 :)

AlanGriffiths:
> > 1. having good quality, consumer facing content badly organized; or,
> > 2. having lots of dubious quality, unfocused content badly organized?
> 
> I vote for 1 :)

I vote for 1 too

tarek-y-ismail:
> > > 1. having good quality, consumer facing content badly organized; or,
> > > 2. having lots of dubious quality, unfocused content badly organized?
> > 
> > 
> > I vote for 1 :)
> 
> I vote for 1 too

Another one!

AlanGriffiths:
> Fine, but if you're not going to plan on bringing those back, I'll find you ;)

Don't be a sore loser: You could have waited for more votes! (And voted!)

PR #4240, opened 2025-09-10
mircommon: Add xkbcommon to pkgconfig dependency

Conan-Kudo:
This is needed in practice for building against mircommon, so declare it in the pkgconfig file so it is automatically set up as a dependency.


Conan-Kudo:
Can this be cherry-picked into 2.22.x?

AlanGriffiths:
> Can this be cherry-picked into 2.22.x?

Added to 2.22.1 milestone

PR #4249, opened 2025-09-11
doc: organize subtrees

Saviq:
## What's new?
Move content around into the target structure

## How to test
Verify the navigation and redirects work

## TODO
getting-started-with-mir needs a place

## Checklist
- [x] Tests added and pass


doc/sphinx/conf.py
```diff
@@ -171,6 +171,21 @@
...
+    'how-to/how-to-use-on-screen-keyboards': '../configuring/how-to/use-on-screen-keyboards',
+    'explanation/architecture': '../contributing/explanation/architecture',
+    'explanation/libraries': '../contributing/explanation/libraries',
+    'how-to/how-to-enable-graphics-for-snaps-on-a-device': '../contributing/how-to/enable-graphics-for-snaps-on-a-device',
+    'how-to/getting_involved_in_mir': '../contributing/how-to/getting-involved-in-mir',
+    'how-to/how-to-test-mir-for-a-release': '../contributing/how-to/test-mir-for-a-release',
+    'how-to/how-to-update-symbols-map': '../contributing/how-to/update-symbols-map',
+    'how-to/how-to-use-checkbox-mir': '../contributing/how-to/use-checkbox-mir',
+    'reference/continuous-integration': '../contributing/reference/continuous-integration',
+    'reference/cppguide': '../contributing/reference/cppguide',
```
jhenstridge: Some of these URLs have only been valid for the last week (i.e. never in stable documentation), so probably don't need redirects. From what I can see, that would be:

- how-to/how-to-enable-remote-desktop
- how-to/how-to-use-on-screen-keyboards
- how-to/how-to-enable-graphics-for-snaps-on-a-device
- reference/cppguide

doc/sphinx/configuring/index.md
```diff
@@ -35,7 +35,7 @@ Mir is a member of the Ubuntu family. It’s an open source project that warmly
 
 ```{toctree}
 :hidden:
+:glob:
 
-how-to/configuring-index
-explanation/configuring-index
+*/index
```
jhenstridge: This changes the order of the items in the tree ("Explanation" comes before "How-to guides"). I think we probably want to keep things consistent with other documentation sets.

doc/sphinx/contributing/index.md
```diff
@@ -43,8 +43,7 @@ Mir is a member of the Ubuntu family. It’s an open source project that warmly
 
 ```{toctree}
 :hidden:
+:glob:
 
-how-to/contributing-index
-explanation/contributing-index
-reference/contributing-index
+*/index
```
jhenstridge: Same here, re: section ordering.

Saviq:
> I don't think we want to use it for the top level sections

Yeah, I actually undid it for the top-level one, but forgot here, will fix!

doc/sphinx/conf.py
Saviq: Thanks, fixed those up - `cppguide` actually needed one from its original place in `_static`, and `graphics` I folded into the previous one.

doc/sphinx/configuring/how-to/index.md
```diff
@@ -0,0 +1,16 @@
+(configuring-how-to-index)=
+
+# How-to guides
+These how-to guides cover the key aspects of using Mir based compositors.
+
+- [Calibrating a touchscreen device](how-to-calibrate-a-touchscreen-device): creating and using a calibration matrix using [the libinput snap](https://snapcraft.io/libinput)
```
AlanGriffiths, line 6: Explicitly providing link text means there is a mismatch between the index and the links:

"How to calibrate a touchscreen device" vs "Calibrating a touchscreen device"

doc/sphinx/explanation/index.md
```diff
@@ -13,14 +13,7 @@ These pages provide additional detail about a number of aspects related to using
 
 ```{toctree}
 :hidden:
+:glob:
```
AlanGriffiths, line 16: Using glob means the order on the page doesn't match the order in the index

doc/sphinx/configuring/how-to/index.md
Saviq: Right, but that's pre-existing - I do think we should default to doc titles (and that's what goes into the toctree), but particularly for How-to's I dislike the "How-to…" in every line.

Let's sort that in a follow-up :+1:

doc/sphinx/explanation/index.md
Saviq: The landing pages will get a rehaul, let's sort out then :+1:

Saviq:
> But there's tidy-up needed on the sub-(sub-)indexes that list docs

ACK, let's do it in a follow up :+1:.

PR #4250, opened 2025-09-11
Add documentation for `mg::Buffer`

tarek-y-ismail:
## What's new?
- Adds Doxygen documentation for `mg::Buffer`

### DIFF
--- a/include/platform/mir/graphics/buffer.h
+++ b/include/platform/mir/graphics/buffer.h
@@ -57,9 +57,8 @@ public:
 
     /// The unique ID of this buffer.
     ///
-    /// This ID is mainly used to identify the buffer for tracing. Returning 0
-    /// or any arbitrary value does not affect normal operation of the
-    /// compositor.
+    /// This ID is only used to identify the buffer for tracing and is not used
+    /// for rendering.
     ///
     /// \returns The unique ID of this buffer.
     virtual BufferID id() const = 0;
--- a/include/platform/mir/graphics/buffer.h
+++ b/include/platform/mir/graphics/buffer.h
@@ -45,12 +45,11 @@ protected:
 /// #mir::graphics::gl::Texture for compositing.
 ///
 ///
-/// \sa GraphicBufferAllocator::buffer_from_resource
-/// \sa GraphicBufferAllocator::buffer_from_shm
-/// \sa DMABufEGLProvider::import_dma_buf
-/// \sa GraphicBufferAllocator::alloc_software_buffer
-/// \sa FramebufferProvider::buffer_to_framebuffer
-/// \sa GLRenderingProvider::as_texture
+/// \sa
+/// - GraphicBufferAllocator - Allocates different types of buffers
+/// - DMABufEGLProvider::import_dma_buf
+/// - FramebufferProvider::buffer_to_framebuffer
+/// - GLRenderingProvider::as_texture - Convert a buffer to an OpenGL texture
 class Buffer
 {
 public:
--- a/include/platform/mir/graphics/buffer.h
+++ b/include/platform/mir/graphics/buffer.h
@@ -58,10 +58,10 @@ public:
     /// \returns The unique ID of this buffer.
     virtual BufferID id() const = 0;
 
+    /// \returns The size of the buffer.
+    ///
     /// \note The size is the logical size of the buffer, where the width is
     /// the perceived width, and NOT the stride.
-    ///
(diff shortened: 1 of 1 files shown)


include/platform/mir/graphics/buffer.h
```diff
@@ -35,15 +35,51 @@ class NativeBufferBase
...
+/// Some buffer types can be converted to [mg::Framebuffer] for display, or
+/// [gl::Texture] for compositing.
+///
+///
+/// \sa GraphicBufferAllocator::buffer_from_resource
+/// \sa GraphicBufferAllocator::buffer_from_shm
+/// \sa DMABufEGLProvider::import_dma_buf
+/// \sa GraphicBufferAllocator::alloc_software_buffer
+/// \sa FramebufferProvider::buffer_to_framebuffer
+/// \sa GLRenderingProvider::as_texture
```
mattkae:
```suggestion
/// \sa GraphicBufferAllocator - allocated different types of buffers
/// \sa DMABufEGLProvider::import_dma_buf - import a dmabuf
/// \sa FramebufferProvider::buffer_to_framebuffer - converts a buffer to a framebuffer
/// \sa GLRenderingProvider::as_texture - converts a buffer to a texture
```

include/platform/mir/graphics/buffer.h
```diff
@@ -35,15 +35,51 @@ class NativeBufferBase
...
 };
 
+/// Generic graphics buffer interface.
+///
+/// Buffers can come from Wayland resources, shared memory, DMA buffers. CPU
+/// accessible buffers can also be created by the compositor for internal
+/// rendering.
+///
+/// Some buffer types can be converted to [mg::Framebuffer] for display, or
+/// [gl::Texture] for compositing.
```
mattkae:
```suggestion
/// Some buffer types can be converted to #mir::graphics::Framebuffer for display, or
/// #mir::graphics::gl::Texture for compositing.
```

mattkae: If we use the full namespace, we can get full links in doxygen. Also, the `#` is recommended :)

tarek-y-ismail:
Doesn't get rendered in RTD, but I think that's intentional from us

include/platform/mir/graphics/buffer.h
```diff
@@ -35,15 +35,50 @@ class NativeBufferBase
...
 class Buffer
 {
 public:
     virtual ~Buffer() {}
 
+    /// The unique ID of this buffer.
+    ///
+    /// This ID is mainly used to identify the buffer for tracing. Returning 0
+    /// or any arbitrary value does not affect normal operation of the
+    /// compositor.
```
AlanGriffiths: 1. This contradicts the "unique ID" description.
2. Who does this help? It's effect is that users of this interface cannot rely on the id being unique, and implementers have no constraints.

tarek-y-ismail: The uniqueness is only a requirement for tracing as far as I understand. But outside of that, the buffer ID doesn't actually matter. Maybe this shouldn't be a pure virtual method?

AlanGriffiths: The documentation should describe what promises the implementer of the interface is making to the consumer of the interface.

It isn't the place to discuss whether all or most consumers care about the promise.

include/platform/mir/graphics/buffer.h
```diff
@@ -57,9 +57,8 @@ class Buffer
 
     /// The unique ID of this buffer.
     ///
-    /// This ID is mainly used to identify the buffer for tracing. Returning 0
-    /// or any arbitrary value does not affect normal operation of the
-    /// compositor.
+    /// This ID is only used to identify the buffer for tracing and is not used
+    /// for rendering.
```
AlanGriffiths: Is this an observation of the way things are right now? Or a requirement that this must remain the case?

If the former, I don't think this is the way to document it.

tarek-y-ismail: Hmm, would a non-doxygen comment do it?

AlanGriffiths: I fear such a comment will just get out of sync with the code. It is unlikely anyone would think to change the comment when they start using the `id()` for something else.

Really, I don't think such observations belong in the interface at all

tarek-y-ismail: Ok, I'll remove these lines and create an issue to handle this in code instead of in docs

PR #4257, opened 2025-09-11
doc: add Fedora first-compositor build

Saviq:
## What's new?
Tutorial now tested on Fedora, too.

## How to test
Check the tutorial and Fedora CI

## Checklist
- [x] Tests added and pass

### DIFF
--- a/spread/build/fedora/task.yaml
+++ b/spread/build/fedora/task.yaml
@@ -27,3 +27,39 @@ execute: |
     popd
 
     ccache --show-stats --zero-stats > ${CCACHE_DIR}/ccache.stats
+
+    # build the local repository for end-to-end tests
+    dnf --assumeyes install createrepo
+    createrepo /root/rpmbuild
+    dnf config-manager addrepo \
+       --id=mirpr \
+       --set=baseurl="file:///root/rpmbuild" \
+       --set=gpgcheck=0 \
+       --set=priority=90 \
+       --set=enabled=1 \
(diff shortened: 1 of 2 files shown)


spread/build/fedora/task.yaml
```diff
@@ -27,3 +27,39 @@ execute: |
     popd
 
     ccache --show-stats --zero-stats > ${CCACHE_DIR}/ccache.stats
+
+    # build the local repository for end-to-end tests
```
Saviq, line 31: @z3ntu hey, would you be able to provide something equivalent for Alpine?

Ideally we'd add https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/testing/mir/APKBUILD in a manner that allows us to build packages and a local pkg cache?

z3ntu: To be fair, I don't quite understanding what this is doing, without looking further into it. This is about running some cmake commands to build that tutorial thing?

Saviq: @z3ntu this is providing a real-life scenario where you build a compositor _on top of_ Mir.

The CMake bit is there - as part of CI what we do here (for Ubuntu and Fedora) is build Mir packages (rather than just the binaries) and then use those packages (from a local repository) to build a downstream CMake project.

We'd like to do the same on Alpine:
- build packages
- create a local repository ("pkg cache"?)
- add the local repository to the system
- install packages from the repository

PR #4262, opened 2025-09-12
doc: run a client in the first-compositor tutorial

Saviq:
## What's new?
The tutorial test now also runs the client.

## How to test
CI

## Checklist
- [x] Tests added and pass


Saviq:
> I think it might be cleaner to install `bomber` separately

Good call, will do! Just as soon as I grok why `apt-get install bomber; bomber` results in `bomber` not being found XD

Saviq:
> and in the section where it is run.

Ah, but the problem there is that I need only Mir and the client to run within that section to not have package installation affect the timeout…

I've added a note why we're installing `bomber`, see if that's OK with you?

PR #4266, opened 2025-09-12
Add initial docs for `mg::software::BufferDescriptor`

tarek-y-ismail:
## What's new?
- Adds initial docs for `BufferDescriptor`

### DIFF
--- a/include/platform/mir/renderer/sw/pixel_source.h
+++ b/include/platform/mir/renderer/sw/pixel_source.h
@@ -47,7 +47,7 @@ public:
 
     /// \returns The pixel format of the buffer
     virtual MirPixelFormat format() const = 0;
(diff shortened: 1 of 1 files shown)


include/platform/mir/renderer/sw/pixel_source.h
```diff
@@ -37,13 +37,19 @@ namespace renderer
...
+///
+/// Used for reading, writing, and transferring data between CPU buffers.
 class BufferDescriptor
 {
 public:
     virtual ~BufferDescriptor() = default;
 
+    /// \returns The pixel format of the buffer
     virtual MirPixelFormat format() const = 0;
+    /// \returns The stride (width + any hardware imposed padding)
```
AlanGriffiths: hardware is not the only reason for padding. It could be software a created sub-buffer.

PR #4269, opened 2025-09-12
Add some documentation for `mg::NativeBufferBase`

tarek-y-ismail:
## What's new?
- Adds a few lines of documentation explaining the purpose of `NativeBufferBase`

### DIFF
--- a/include/platform/mir/graphics/buffer.h
+++ b/include/platform/mir/graphics/buffer.h
@@ -26,7 +26,7 @@ namespace mir
 namespace graphics
 {
 
-/// Used with buffers that wrap other buffers. This method provides direct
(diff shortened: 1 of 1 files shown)


include/platform/mir/graphics/buffer.h
```diff
@@ -26,6 +26,10 @@ namespace mir
 namespace graphics
 {
 
+/// Used with buffers that wrap other buffers. This method provides direct
+/// access to the innermost buffer.
```
AlanGriffiths:
```suggestion
/// Used with buffers that wrap other buffers. This interface supports direct
/// access to the innermost buffer.
```

PR #4272, opened 2025-09-16
Simplify `Renderer::render` 

tarek-y-ismail:
Related: #4149 

## What's new?
- `Framebuffer` now extends `Buffer`
- `Renderer::render` now returns `Framebuffer`
- `ScreenShooter::capture` now passes a buffer to consumers in its callback, instead of requiring consumers to hand it a buffer.
- `BufferPool` and `ClaimedBuffer` are now not needed.

## How to test
- Run `miral-app`
- Enable magnification
- Magnification should work like before


src/include/server/mir/compositor/screen_shooter.h
```diff
@@ -52,11 +53,12 @@ class ScreenShooter
     /// The [callback] may be called on a different thread. It will be given the timestamp
     /// of the capture if it succeeds or nullopt if there was an error.
     virtual void capture(
-        std::shared_ptr<renderer::software::WriteMappableBuffer> const& buffer,
         geometry::Rectangle const& area,
         glm::mat2 const& transform,
         bool overlay_cursor,
-        std::function<void(std::optional<time::Timestamp>)>&& callback) = 0;
+        std::function<
+            void(std::optional<time::Timestamp>, std::shared_ptr<mir::graphics::Buffer> buffer)>&&
```
tarek-y-ismail: Should this become:
```suggestion
            void(std::optional<std::pair<time::Timestamp, std::shared_ptr<mir::graphics::Buffer>>)>&&
```

src/platforms/atomic-kms/server/gbm_display_allocator.cpp
tarek-y-ismail, line 1: I could drop most of the changes here except the addition of `native_buffer_base` and `pixel_format`. Opinions?

src/platforms/common/server/cpu_addressable_fb.cpp
```diff
@@ -296,6 +296,16 @@ mg::CPUAddressableFB::operator uint32_t() const
...
 }
 
+auto mg::CPUAddressableFB::pixel_format() const -> MirPixelFormat
+{
+    return buffer->format();
+}
+
+auto mg::CPUAddressableFB::native_buffer_base() -> NativeBufferBase*
+{
+    return this;
```
tarek-y-ismail, line 306:
```suggestion
    return buffer.get();
```

src/platforms/common/server/cpu_addressable_fb.h
```diff
@@ -24,7 +24,7 @@
 
 namespace mir::graphics
 {
-class CPUAddressableFB : public FBHandle, public CPUAddressableDisplayAllocator::MappableFB
+class CPUAddressableFB : public FBHandle, public CPUAddressableDisplayAllocator::MappableFB, public NativeBufferBase
```
tarek-y-ismail, line 27:
```suggestion
class CPUAddressableFB : public FBHandle, public CPUAddressableDisplayAllocator::MappableFB
```

src/platforms/gbm-kms/server/gbm_display_allocator.cpp
tarek-y-ismail, line 1: Again, most of the `format` related changes could be dropped in favor of `gbm_bo_get_format`

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -198,38 +219,46 @@ auto mc::BasicScreenShooter::Self::render(
...
+        display_provider->allocate_buffer();
+        offscreen_sink = std::make_unique<OffscreenDisplaySink>(display_provider, buffer_size);
         auto gl_surface = render_provider->surface_for_sink(*offscreen_sink, *config);
         current_renderer = renderer_factory->create_renderer_for(std::move(gl_surface), render_provider);
         last_rendered_size = buffer_size;
     }
+    else
+    {
+        // TODO: Do we need to allocate a new buffer every time?
+        // I think we do
```
tarek-y-ismail: Should be nuked

src/server/compositor/basic_screen_shooter.h
```diff
@@ -58,11 +58,11 @@ class BasicScreenShooter: public ScreenShooter
         std::shared_ptr<graphics::Cursor> const& cursor);
 
     void capture(
-        std::shared_ptr<renderer::software::WriteMappableBuffer> const& buffer,
         geometry::Rectangle const& area,
         glm::mat2 const& transform,
         bool overlay_cursor,
-        std::function<void(std::optional<time::Timestamp>)>&& callback) override;
+        std::function<void(std::optional<time::Timestamp>, std::shared_ptr<mir::graphics::Buffer>)>&& callback) // TODO: This should be an optional pair
```
tarek-y-ismail:
```suggestion
        std::function<void(std::optional<time::Timestamp>, std::shared_ptr<mir::graphics::Buffer>)>&& callback)
```

src/server/compositor/default_display_buffer_compositor.cpp
```diff
@@ -51,6 +51,26 @@ mc::DefaultDisplayBufferCompositor::DefaultDisplayBufferCompositor(
...
+    {
+        return std::unique_ptr<To>{to_src};
+    }
+    delete unowned_src;
+    BOOST_THROW_EXCEPTION((
+        std::bad_cast()));
+}
+}
+}
+
```
tarek-y-ismail, line 73: This is duplicated at least two more times across the codebase, maybe it should be moved to `mircommon` or another common place?

src/server/compositor/null_screen_shooter.h
```diff
@@ -34,11 +34,10 @@ class NullScreenShooter: public ScreenShooter
     NullScreenShooter(Executor& executor);
 
     void capture(
-        std::shared_ptr<renderer::software::WriteMappableBuffer> const& buffer,
         geometry::Rectangle const& area,
         glm::mat2 const& transform,
         bool overlay_cursor,
-        std::function<void(std::optional<time::Timestamp>)>&& callback) override;
+        std::function<void(std::optional<time::Timestamp>, std::shared_ptr<mir::graphics::Buffer>)>&& callback) override;
```
tarek-y-ismail:
```suggestion
        std::function<void(std::optional<time::Timestamp>, std::shared_ptr<graphics::Buffer>)>&& callback) override;
```

src/server/compositor/basic_screen_shooter.h
tarek-y-ismail:
```suggestion
        std::function<void(std::optional<time::Timestamp>, std::shared_ptr<graphics::Buffer>)>&& callback) // TODO: This should be an optional pair
```

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -96,16 +112,21 @@ class mc::BasicScreenShooter::Self::OneShotBufferDisplayProvider : public mg::CP
...
+        buffer_size = new_size;
+    }
+
+    void allocate_buffer()
+    {
+        next_buffer = allocator->alloc_software_buffer(buffer_size, buffer_format);
     }
 private:
-    std::shared_ptr<mrs::WriteMappableBuffer> next_buffer;
+    std::shared_ptr<graphics::GraphicBufferAllocator> const allocator;
```
tarek-y-ismail:
```suggestion
    std::shared_ptr<mg::GraphicBufferAllocator> const allocator;
```

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -198,38 +219,46 @@ auto mc::BasicScreenShooter::Self::render(
...
 
     // Because we might be called on a different thread next time we need to
     // ensure the renderer doesn't keep the EGL context current
     renderer.suspend();
-    return captured_time;
+    return {captured_time, std::move(frame)};
 }
 
-auto mc::BasicScreenShooter::Self::renderer_for_buffer(std::shared_ptr<mrs::WriteMappableBuffer> buffer)
+auto mc::BasicScreenShooter::Self::renderer_for_buffer(geometry::Rectangle const& area)
```
tarek-y-ismail:
```suggestion
auto mc::BasicScreenShooter::Self::renderer_for_buffer(geom::Rectangle const& area)
```

src/include/server/mir/compositor/screen_shooter.h
tarek-y-ismail:
```suggestion
            void(std::optional<time::Timestamp>, std::shared_ptr<graphics::Buffer> buffer)>&&
```

src/platforms/common/server/cpu_addressable_fb.cpp
tarek-y-ismail: `std::uniqe_ptr<Buffer> buffer;` is actually a bespoke class, not an instance of `mg::Buffer`

tarek-y-ismail:
I don't know why Alpine 3.19 and 3.20 fail `RenderSceneIntoSurfaceTest`, seems the connection to them is spotty at the moment so I'll try debugging them again tomorrow.

src/miral/render_scene_into_surface.cpp
```diff
@@ -200,16 +95,24 @@ class SceneRenderingSurface : public ms::BasicSurface
...
+                if (buffer)
+                {
+                    get_streams().begin()->stream->submit_buffer(
+                        buffer, capture_rect.size, geom::RectangleD({0, 0}, capture_rect.size));
+                }
+                capture_complete->set_value();
+            });
+
+        future.wait();
         return BasicSurface::generate_renderables(id);
```
AlanGriffiths, line 107: I've not considered the context, but holding a lock whilst waiting on a future is not a good look.

Also wouldn't a `binary_semaphore` be simpler than a `promise` + a `future`?

src/include/server/mir/compositor/screen_shooter.h
```diff
@@ -52,11 +53,12 @@ class ScreenShooter
...
     /// of the capture if it succeeds or nullopt if there was an error.
     virtual void capture(
-        std::shared_ptr<renderer::software::WriteMappableBuffer> const& buffer,
         geometry::Rectangle const& area,
         glm::mat2 const& transform,
         bool overlay_cursor,
-        std::function<void(std::optional<time::Timestamp>)>&& callback) = 0;
+        std::function<
+            void(std::optional<time::Timestamp>, std::shared_ptr<graphics::Buffer> buffer)>&&
+            callback) = 0;
```
mattkae:
```suggestion
        std::function<void(std::optional<time::Timestamp>, std::shared_ptr<graphics::Buffer> buffer)>&& callback) = 0;
```

src/platforms/wayland/wl_egl_display_provider.h
```diff
@@ -80,4 +83,4 @@ class WlDisplayAllocator : public GenericEGLDisplayAllocator
 };
 }
 
-#endif // MIR_PLATFORM_WAYLAND_DISPLAY_PROVIDER_H_
\ No newline at end of file
+#endif // MIR_PLATFORM_WAYLAND_DISPLAY_PROVIDER_H_
```
mattkae, line 86: Readd empty line plz?

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -39,34 +40,49 @@ namespace geom = mir::geometry;
 class mc::BasicScreenShooter::Self::OneShotBufferDisplayProvider : public mg::CPUAddressableDisplayAllocator
 {
 public:
-    OneShotBufferDisplayProvider() = default;
+    OneShotBufferDisplayProvider(std::shared_ptr<mg::GraphicBufferAllocator> allocator, geom::Size initial_size) :
```
mattkae:
```suggestion
    OneShotBufferDisplayProvider(std::shared_ptr<mg::GraphicBufferAllocator> allocator, geom::Size const& initial_size) :
```

src/miral/render_scene_into_surface.cpp
tarek-y-ismail: This was mostly as an attempt to fix tests failing on Alpine. I had thought that maybe it exposes a race condition that other distros don't, but the issue still seems to persist. So it looks like the issue might be something else entirely

src/server/compositor/basic_screen_shooter.h
```diff
@@ -78,15 +78,13 @@ class BasicScreenShooter: public ScreenShooter
...
-        auto render(
-            std::shared_ptr<renderer::software::WriteMappableBuffer> const& buffer,
-            geometry::Rectangle const& area,
-            glm::mat2 const& transform,
-            bool overlay_cursor) -> time::Timestamp;
+        auto render(geometry::Rectangle const& area, glm::mat2 const& transform, bool overlay_cursor)
+            -> std::pair<time::Timestamp, std::shared_ptr<graphics::Buffer>>;
 
-        auto renderer_for_buffer(std::shared_ptr<renderer::software::WriteMappableBuffer> buffer)
+        auto renderer_for_buffer(geometry::Rectangle const& area)
```
mattkae:
```suggestion
        auto renderer_for_area(geometry::Rectangle const& area)
```

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -198,38 +219,44 @@ auto mc::BasicScreenShooter::Self::render(
...
+        display_provider->output_size(buffer_size);
+        display_provider->allocate_buffer();
+        offscreen_sink = std::make_unique<OffscreenDisplaySink>(display_provider, buffer_size);
         auto gl_surface = render_provider->surface_for_sink(*offscreen_sink, *config);
         current_renderer = renderer_factory->create_renderer_for(std::move(gl_surface), render_provider);
         last_rendered_size = buffer_size;
     }
+    else
+    {
+        display_provider->allocate_buffer();
```
mattkae: Why are we continuously allocating the buffer here if the size matches? This can be inefficient if I'm (for example) recording a monitor on my desktop

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -39,34 +40,49 @@ namespace geom = mir::geometry;
 class mc::BasicScreenShooter::Self::OneShotBufferDisplayProvider : public mg::CPUAddressableDisplayAllocator
 {
 public:
-    OneShotBufferDisplayProvider() = default;
+    OneShotBufferDisplayProvider(std::shared_ptr<mg::GraphicBufferAllocator> allocator, geom::Size initial_size) :
+        allocator{allocator},
+        buffer_size{initial_size}
+    {
+        allocate_buffer();
```
mattkae: Why bother having an initial size allocated if you're most likely going to throw it away anyway?

src/server/compositor/default_display_buffer_compositor.cpp
```diff
@@ -128,7 +148,9 @@ bool mc::DefaultDisplayBufferCompositor::composite(mc::SceneElementSequence&& sc
         renderer->set_viewport(view_area);
         renderer->set_output_filter(output_filter->filter());
 
-        display_sink.set_next_image(renderer->render(renderable_list));
+        auto buf = renderer->render(renderable_list);
+        if(auto framebuffer = unique_ptr_cast<mg::Framebuffer>(std::move(buf)))
+            display_sink.set_next_image(std::move(framebuffer));
```
mattkae, line 153: I wonder if @RAOF has any feelings about this, but this looks like an unfortunate bit that we have to do on every frame.

src/server/compositor/basic_screen_shooter.cpp
tarek-y-ismail: Just to avoid any nasty surprises :)

src/server/compositor/basic_screen_shooter.cpp
tarek-y-ismail: Should be handled now.

src/server/compositor/basic_screen_shooter.cpp
mattkae: Kk, I guess that's fine, but it always an incorrect guess ;)

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -39,34 +40,49 @@ namespace geom = mir::geometry;
...
+        next_buffer{allocator->alloc_software_buffer(buffer_size, buffer_format)}
+    {
+    }
 
     class FB : public mg::CPUAddressableDisplayAllocator::MappableFB
     {
     public:
-        FB(std::shared_ptr<mrs::WriteMappableBuffer> buffer)
-            : buffer{std::move(buffer)}
+        FB(std::shared_ptr<mg::Buffer> buffer) :
```
mattkae:
```suggestion
        explicit FB(std::shared_ptr<mg::Buffer> buffer) :
```

mattkae (changes requested):
I believe screenshotting is broken? Maybe I'm doing something wrong, but I get an empty image

src/server/compositor/default_display_buffer_compositor.cpp
AlanGriffiths: Not a great name: Probably better to use `boost::dynamic_pointer_cast<>()` which has a better name and avoids duplication

tarek-y-ismail: I don't think boost's alternative would help here. We need a unique pointer back because that's what `set_next_image` takes as a parameter, and converting a shared pointer back to a unique pointer is not possible (safely) as far as I know

tarek-y-ismail:
Will try debugging through tmate once more next week.

src/server/frontend_wayland/wlr_screencopy_v1.cpp
```diff
@@ -435,10 +436,19 @@ void mf::WlrScreencopyFrameV1::capture(geom::Rectangle buffer_space_damage)
...
+            (std::optional<time::Timestamp> captured_time, auto buffer)
         {
+            assert(buffer->pixel_format() == self.value().target->format());
+
+            auto source_as_readable = mrs::as_read_mappable_buffer(std::move(buffer));
+            auto source_rmap = source_as_readable->map_readable();
+            auto target_wmap = self.value().target->map_writeable();
+
+            assert(target_wmap->len() == source_rmap->len());
+            std::memcpy(target_wmap->data(), source_rmap->data(), target_wmap->len());
```
mattkae: 1. I don't exactly love the asserts here. Maybe we can throw a protocol error or something, but the assert seems a bit dramatic
2. This snippet of code is very unfortunate :) I imagine that a batter solution would be to just use the buffer that the client provided to us. The copy seems like extreme overkill when we already have a buffer that would work for us

AlanGriffiths: We shouldn't be using the `assert` macro here, this looks like an invariant check.

Problems with `assert`:

1. if a condition is worth checking it is (usually) worth checking in all builds (not just when `NDEBUG` happens to be unset).
2. If the condition has side-effects (I know, but too often it does) then we have different logic in different builds complicating testing.
3. we have `fatal_error()` when we want to terminate the process.

See our [Error handling strategy](https://github.com/canonical/mir/blob/main/HACKING.md#error-handling-strategy) for acceptable usage of `assert()` .

src/include/server/mir/compositor/screen_shooter.h
```diff
@@ -58,6 +59,21 @@ class ScreenShooter
...
 
+    /// Capture a rectangle on the screen given by [area] into a buffer
+    /// supplied by the screenshooter. The resulting render will have the
+    /// [transform] applied to it. It is recommended that the [area] is an area
+    /// on a single output with that output's corresponding [transform],
+    /// however this is up to the discretion of the caller.
+    ///
+    /// The [callback] may be called on a different thread. It will be given
+    /// the timestamp of the capture and the buffer containing the screenshot
+    /// if it succeeds or nullopt and a nullptr if there was an error.
```
mattkae, line 70:
```suggestion
    /// Capture a rectangle on the screen given by \p area into a buffer
    /// supplied by the screenshooter.
    
    /// The resulting render will have the \p transform applied to it.
    /// It is recommended that the \p area is an area
    /// on a single output with that output's corresponding \p transform,
    /// however this is up to the discretion of the caller.
    ///
    /// The \p callback may be called on a different thread. It will be given
    /// the timestamp of the capture and the buffer containing the screenshot
    /// if it succeeds or nullopt and a nullptr if there was an error.
```

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -75,36 +92,50 @@ class mc::BasicScreenShooter::Self::OneShotBufferDisplayProvider : public mg::CP
...
         }
-        return std::make_unique<FB>(std::exchange(next_buffer, nullptr));
+
+        return std::make_unique<FB>(std::exchange(
+            next_buffer, mrs::as_write_mappable_buffer(allocator->alloc_software_buffer(buffer_size, buffer_format))));
     }
 
     auto output_size() const -> geom::Size override
     {
+        // TODO nuke this
```
mattkae: What do we mean here? XD

```suggestion
        // TODO nuke this
```

src/include/server/mir/compositor/screen_shooter.h
```diff
@@ -58,6 +59,21 @@ class ScreenShooter
...
+    virtual void capture(
```
mattkae, line 71: I'm unsure how I feel about this API here. In one case, you bring your own buffer, and in the other case you get one supplied to you that is appropriate to the size. In my opinion, callers should always supply their own buffer to the screen shooter. In that way, the concern of getting the buffer is separated from the concern of filling it with stuff.

WDYT?

src/server/compositor/basic_screen_shooter.cpp
tarek-y-ismail: We always have a `next_buffer`, so we don't need to check if it's null

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
tarek-y-ismail, line 1: Might want to see if `buffer` is unused now

src/include/server/mir/compositor/screen_shooter.h
tarek-y-ismail: The only use case right now where we need to supply the buffer is for screencopy to avoid an extra copy. On the other hand, this PR adds a _lot_ of changes for not much simplification, as the only other use case that uses the other implementation is in `RenderSceneIntoSurface`, which already worked with the existing impl.

@RAOF What do you think? Am missing any other parts of the code which could be simplified further with the changes in this PR, or should we close it down?

src/miral/render_scene_into_surface.cpp
AlanGriffiths: So, this looks wrong and isn't useful?

AlanGriffiths:
Is this really a simplification?

> +513-250

src/miral/render_scene_into_surface.cpp
tarek-y-ismail: It has been removed

PR #4277, opened 2025-09-17
Use more markdown in command line options.

robert-ancell:
Also substitute key values in description for --display-config as with other options.


mattkae (approved):
Mark :arrow_down: :tada:

PR #4280, opened 2025-09-18
Fix program return value of 1 when run with --help/--version

robert-ancell:
Since the output of --help and --version are generated using exceptions, 1 is returned from main which looks like an error to the shell.

Instead, print this information to stdout and then exit before running the server.

### DIFF
--- a/src/miral/runner.cpp
+++ b/src/miral/runner.cpp
@@ -180,6 +180,12 @@ try
         fd_manager->set_main_loop(main_loop);
     });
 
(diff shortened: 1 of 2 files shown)


robert-ancell:
The server peeking at the options to determine if it should run is a bit magic. This would be cleaner if `mo::DefaultConfiguration::parse_arguments()` set a flag when running `--help` and `--version` and then that could be read in `mir::Server::run()` but I'm struggling to get access to `mo::DefaultConfiguration` at that point.

PR #4287, opened 2025-09-19
Add a --help-markdown flag to generate documentation

robert-ancell:
This is intended to be used by projects that are built using Mir to generate their own documentation.

For example, if a project has a shell called `my-shell` they would generate this using:
```
$ my-shell --help-markdown > options.md.include
```

And then include that into their own documentation with:
```
Options for my-shell, which can be set on the command line, e.g. `--version`.

{include} options.md.include
```

### DIFF
--- a/src/platform/options/default_configuration.cpp
+++ b/src/platform/options/default_configuration.cpp
@@ -344,41 +344,57 @@ namespace
 {
 std::string escape_markdown(std::string const& text)
 {
-    std::string escaped_text;
+    std::stringstream escaped_text;
 
     for (char const c : text)
     {
         switch(c)
         {
         case '&':
-            escaped_text += "&amp;";
+            escaped_text << "&amp;";
             break;
         case '<':
-            escaped_text += "&lt;";
+            escaped_text << "&lt;";
             break;
         case '>':
-            escaped_text += "&gt;";
+            escaped_text << "&gt;";
             break;
         case '\"':
-            escaped_text += "&quot;";
+            escaped_text << "&quot;";
             break;
         case '\'':
-            escaped_text += "&apos;";
+            escaped_text << "&apos;";
             break;
         default:
-            escaped_text += c;
+            escaped_text << c;
             break;
         }
     }
 
-    return escaped_text;
+    return escaped_text.str();
 }
 
-std::string options_to_markdown(const boost::program_options::options_description& desc)
+std::string make_markdown_label(std::string const& text)
+{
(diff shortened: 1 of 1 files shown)


robert-ancell:
Example output:
```markdown
## Options

(global-arw-file)=
### `arw-file`

Make server Wayland socket readable and writeable by all users. For debugging purposes only.

(global-platform-display-libs)=
### `platform-display-libs`

Comma separated list of libraries to use for platform output support, e.g. `mir:x11,mir:wayland`. If not provided the libraries are autodetected.

(global-platform-rendering-libs)=
### `platform-rendering-libs`

Comma separated list of libraries to use for platform rendering support, e.g. `mir:egl-generic`. If not provided the libraries are autodetected.

<!-- ... -->

## Options for mir:wayland platform

(mir-wayland-wayland-host)=
### `wayland-host`

Display name for host compositor, e.g. `wayland-0`.

(mir-wayland-wayland-surface-app-id)=
### `wayland-surface-app-id`

Application ID for the window containing the Mir output.

(mir-wayland-wayland-surface-title)=
### `wayland-surface-title`

Title of the window containing the Mir output.

## Options for mir:virtual platform

(mir-virtual-virtual-output)=
### `virtual-output`

Colon separated list of outputs to use. Dimensions are in the form `WIDTHxHEIGHT`, e.g. `1920x1080:3840x2160`.
```

Saviq:
#4288 consumes this for `miral-shell` as an example.

Saviq:
Should we sort the options? I can't see a valuable logic to the current ordering…

Saviq:
> Should we sort the options?

Yeah, we should ;)

https://github.com/canonical/mir/actions/runs/17921384668/job/50957038513?pr=4288#step:8:4286

robert-ancell:
> Should we sort the options? I can't see a valuable logic to the current ordering…

Good idea. Generally the options are in an order of some sort of importance that makes sense on the command line, but in a document alphabetical is probably more predictable.

robert-ancell:
> Can we document default values? I know they are a PITA to extract[*], but they are helpful to the reader.
> 
> * I think it can be done via `value_semantic::apply_default()`

I left it out because it was a PITA, added it now though, thanks for the tip!

AlanGriffiths (approved):
Cool. I look forward to using this in Miriway.

The alphabetical sorting isn't great for the `*-report` options, but we don't really have semantic context to do better.

PR #4288, opened 2025-09-19
doc: add options reference

Saviq:
## What's new?
Add an options reference under `configuring/reference`

## How to test
RTD

## Checklist
- [ ] ~Tests added and pass~
- [X] Adequate documentation added

### DIFF
--- a/doc/sphinx/configuring/reference/options.md
+++ b/doc/sphinx/configuring/reference/options.md
@@ -1,0 +1,11 @@
+(options-reference)=
+
+# Mir configuration options
+
+Mir compostiros can usually be configured (in the order of preference) on the command line, in the environment, or in a configuration file.
+
+Different compositors will expose different options, but the below is a common set.
+
+```{include} options.md.include
+
+```
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -6,6 +6,14 @@ set(MAKEOPTS
 )
 set(SPHINXDIR ${CMAKE_CURRENT_SOURCE_DIR}/sphinx)
 
+set(OPTIONS_REFERENCE ${SPHINXDIR}/configuring/reference/options.md.include)
+add_custom_target(
+  COMMAND
+    ${CMAKE_BINARY_DIR}/bin/miral-shell --help-markdown
+    > ${OPTIONS_REFERENCE}
+  BYPRODUCTS ${OPTIONS_REFERENCE}
+)
+
 option(
   BUILD_DOXYGEN
   "Build Doxygen documentation as part of the default build"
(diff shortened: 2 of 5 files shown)


doc/sphinx/configuring/reference/options.md
```diff
@@ -0,0 +1,11 @@
+(options-reference)=
+
+# Mir configuration options
+
+Mir compostiros can usually be configured (in the order of preference) on the command line, in the environment, or in a configuration file.
```
AlanGriffiths:
```suggestion
Mir compositors can usually be configured (in the order of preference) on the command line, in the environment, or in a configuration file.
```

AlanGriffiths (changes requested):
This looks like an empty set. I think we can do better 
image

Saviq:
> This looks like an empty set. I think we can do better

Ah. So that's why Robert wanted to cache this, we don't want to build Mir on ReadTheDocs :facepalm:

doc/sphinx/configuring/reference/options.md
```diff
@@ -0,0 +1,11 @@
...
+
+Different compositors will expose different options, but the below is a common set.
```
AlanGriffiths: Rather that "a common set" which is open to misinterpretation, perhaps say "from the miral-shell example"?

AlanGriffiths (changes requested):
There's still a miss-spelling.

Also there are no default values shown - but that's for the other PR

doc/sphinx/configuring/reference/options.md
Saviq:
```suggestion
Different compositors will expose different options, here's a list from the `miral-shell` example:
```

AlanGriffiths (approved):
Just wondering - didn't we just introduce a way to build the "hello world" example? That would give us an actual "common set of options supported by every compositor".

Saviq:
> build the "hello world" example

It's set up to build as a downstream, so not trivial to include in our usual CMake machinery to have it built and executed as a target…

Saviq:
> That would give us an actual "common set of options supported by every compositor".

Done! With a test to boot.

doc/sphinx/configuring/reference/options.md
```diff
@@ -0,0 +1,11 @@
+(options-reference)=
+
+# Mir configuration options
+
+Mir compositors can usually be configured (in the order of precedence) on the command line, in the environment, or in a configuration file.
+
+Different compositors will expose different options, here's a list from the `miral-shell` example:
```
AlanGriffiths: So, this isn't `miral-shell` anymore?

Saviq: Yup, fixed.

AlanGriffiths:
> Done! With a test to boot.

```
The following tests FAILED:
	  4 - verify-options-reference-unchanged (Failed)
```

"First, write a failing test" :tada:

Saviq:
> "First, write a failing test" 🎉

Hence: https://github.com/canonical/mir/pull/4287#issuecomment-3320131116

Saviq:
Ah, now we need to sort platforms :man_facepalming:

AlanGriffiths (approved):
OK

PR #4291, opened 2025-09-22
Add a documentation comment to `BufferBasic`

tarek-y-ismail:
## What's new?
- Explains what `BufferBasic` is.

### DIFF
--- a/include/platform/mir/graphics/buffer_basic.h
+++ b/include/platform/mir/graphics/buffer_basic.h
@@ -24,6 +24,9 @@ namespace mir
 namespace graphics
 {
 
+/// Automatically implements #Buffer::id.
+///
+/// IDs are not guaranteed to be stable across executions of Mir.
 class BufferBasic : public Buffer
 {
 public:


include/platform/mir/graphics/buffer_basic.h
```diff
@@ -24,6 +24,11 @@ namespace mir
 namespace graphics
 {
 
+/// A class that automatically generates runtime unique buffer IDs.
+///
+/// IDs are not guaranteed to be stable across executions of Mir. Use this
+/// class instead of #mir::graphics::Buffer if you don't want to implement
+/// #mir::graphics::Buffer::id yourself.
```
AlanGriffiths:
```suggestion
/// Shared implementation of the Buffer interface.
/// Specifically, automatically generating unique buffer IDs.
///
/// IDs are not guaranteed to be stable across executions of Mir.
```

include/platform/mir/graphics/buffer_basic.h
```diff
@@ -24,6 +24,10 @@ namespace mir
 namespace graphics
 {
 
+/// Shared implementation of the Buffer interface.
+/// Specifically, automatically generating unique buffer IDs.
+///
+/// IDs are not guaranteed to be stable across executions of Mir.
```
mattkae:
```suggestion
/// Shared base class for the `Buffer` interface.
/// 
/// This base class automatically generates unique buffer IDs
/// for derivations. These IDs are not guaranteed to be
/// stable across executions of Mir.
///
/// \sa mir::graphics::common::ShmBuffer - a derivation of this class which uses shared memory as a backing buffer
```

And you may also consider adding other `\sa` :)

mattkae (approved):
One request, but I won't block on it if you dislike my suggested wording

include/platform/mir/graphics/buffer_basic.h
tarek-y-ismail: I think both "Shared implementation..." and "Shared base ..." might be a bit confusing. Maybe:
```suggestion
/// Automatically implements #Buffer::id
///
/// IDs are not guaranteed to be stable across executions of Mir.
/// ...
```

mattkae: I don't like that either :)

How about:

```
/// Base class for buffers.
```

Then you can move the `id` docs to the definition in the class

tarek-y-ismail: Maybe? I just want to indicate to readers that this class does one thing, and just this one thing. Having the comment moved to `id` is a step away from that

mattkae: Cool by me then :)

tarek-y-ismail: One more thing, since generated docs already show classes that inherit from this, I don't think the "see also" section is necessary.

PR #4293, opened 2025-09-22
Rename `{Read,Write,RW}MappbleBuffer` to `*Mappable`

### LINKED ISSUE #4292: `BufferDescriptor` should be renamed (and documented)
`BufferDescriptor` can be easily mistaken for `Buffer`. In fact, until I went through the usages of `BufferDescriptor` is always thought the two are related. In reality, `BufferDescriptor` is used as a base for `ReadMappable` and `WriteMappable`. So it follows that it should be renamed to `Mappable` or something similar.

In addition, to reduce confusion in the future, it should be documented.

tarek-y-ismail:
Closes #4292 

The naming of `*MappableBuffer` implies that the class inherits from `mg::Buffer`, which is not the case. This caused me a bit of confusion when working on #4272. I think dropping the `Buffer` suffix should be enough to make it clear that *it does not inherit from `mg::Buffer`*

## What's new?
- Rename `{Read,Write,RW}MappbleBuffer` to `*Mappable`
- Rename `{Read,Write}TransferrableBuffer` to `*Transferrable`
- Rename `as_{read,write}_mappable_buffer` to `as_{read,write}_mappable`
- Bump mirplatform ABI to 33
- Bump mirplatform symbols to 2.23

### DIFF
--- a/include/platform/mir/renderer/sw/pixel_source.h
+++ b/include/platform/mir/renderer/sw/pixel_source.h
@@ -72,10 +72,10 @@ public:
 /**
  * A Buffer that can be mapped into CPU-accessible memory and directly written to.
  */
-class WriteMappableBuffer : public virtual BufferDescriptor
+class WriteMappable : public virtual BufferDescriptor
 {
 public:
-    virtual ~WriteMappableBuffer() = default;
+    virtual ~WriteMappable() = default;
 
     /**
      * Map the buffer into CPU-writeable memory.
@@ -94,10 +94,10 @@ public:
 /**
  * A Buffer that can be mapped into CPU-readable memory.
  */
-class ReadMappableBuffer : public virtual BufferDescriptor
+class ReadMappable : public virtual BufferDescriptor
 {
 public:
-    virtual ~ReadMappableBuffer() = default;
+    virtual ~ReadMappable() = default;
 
     /**
      * Map the buffer into CPU-readable memory.
@@ -112,12 +112,12 @@ public:
 /**
  * A buffer that can be mapped into CPU-accessible memory for both reading and writing.
  */
-class RWMappableBuffer :
-    public ReadMappableBuffer,
-    public WriteMappableBuffer
+class RWMappable :
+    public ReadMappable,
+    public WriteMappable
 {
 public:
-    ~RWMappableBuffer() override = default;
+    ~RWMappable() override = default;
 
     /**
      * Map the buffer into CPU-accessible memory for both reading and writing.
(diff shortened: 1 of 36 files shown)


AlanGriffiths:
This breaks the mirplatform ABI - you know the drill

tarek-y-ismail:
@RAOF I think we also should rename child classes.. `WriteMappableBuffer` -> `CpuWriteMapping` perhaps. Basically we should drop `Buffer` from all names so developers (us) don't confuse these classes with `mg::Buffer`. None are exposed as symbols except `as_{read,write}_mappable_buffer`.

tarek-y-ismail:
> This breaks the mirplatform ABI - you know the drill

Will do once we decide how much we want to change (maybe nothing at all)

Edit: turned out to be less involved than I remember

include/platform/mir/renderer/sw/pixel_source.h
```diff
@@ -40,10 +40,10 @@ namespace software
 /// Describes the layout of CPU accessible buffers.
 ///
 /// Used for reading, writing, and transferring data between CPU buffers.
-class BufferDescriptor
+class CpuMappable
```
tarek-y-ismail: We can probably drop the `Cpu` prefix, though I fear this would cause confusion between `Mappable` and `Mapping`

include/platform/mir/renderer/sw/pixel_source.h
```diff
@@ -54,7 +54,7 @@ class BufferDescriptor
 };
 
 template<typename T>
-class Mapping : public BufferDescriptor
+class Mapping : public CpuMappable
```
AlanGriffiths: This really doesn't sound like an _IsA_ relationship: "A `Mapping` _IsA_ `CpuMappable`"?

["Naming things is hard"](https://www.mediawiki.org/wiki/Naming_things)

include/platform/mir/renderer/sw/pixel_source.h
RAOF, line 67:
```suggestion
virtual auto descriptor() const -> BufferDescriptor& = 0;
```
????

RAOF (commented):
My opinion: I think the rename of the `BufferDescriptor` is not what we want to do; that is a better (but not great) name for what it is.

The other renaming looks good to me; `*MappableBuffer` -> `*Mappable` *should* be less confusing, and I think is a reasonable thing to do.

Additionally, and maybe not for this PR but for subsequent work, Alan is quite right that none of `Mapping` or `*Mappble` *IsA* `BufferDescriptor`; rather, they have `format()`s, `size()`s, `stride()`s. Whether this is done by adding those methods directly, or by `HasA`ing a `descriptor()` accessor or some secret third thing might be worth investigating.

include/platform/mir/renderer/sw/pixel_source.h
AlanGriffiths: I had a scan through the code. Nothing is _using_ the `BufferDescriptor` interface. (That is these functions can be pushed down the class hierarchy without breaking code.)

Apart from this accidental commonality, I see it is used as a virtual base class - but don't see it used for dynamic casting anywhere (a plausible reason for its existence).

In short, I don't think we need this class at all (so it doesn't need a name).

See #4316

PR #4294, opened 2025-09-22
Revert "Check DRM for supported formats (#3950)"

### LINKED ISSUE #4290: "Failed to find CRTC" on amd+intel hybrid
### Version

v2.22.1

### Environment

OS: jammy 
CPU: N/A
GPU 1: AMD Radeon RX 6500 (radeonsi, navi24, LLVM 19.1.1, DRM 3.49, 6.1.0-1006-oem)
GPU 2: Intel…
Monitor Setup:
```
Initial display configuration:
* Output 1: DisplayPort connected, used
. |_ EDID monitor name: 28E850
. |_ EDID manufacturer: AOC
. |_ EDID product code: 3277
. |_ EDID serial number: 0
. |_ Physical size 27.8" 620x340mm
. |_ Power is on
. |_ Current mode 1920x1080 60.00Hz
. |_ Preferred mode 1920x1080 60.00Hz
. |_ Orientation normal
. |_ Logical size 1920x1080
. |_ Logical position +0+0
. |_ Scaling factor: 1.00
* Output 2: DisplayPort disconnected
* Output 3: DisplayPort disconnected
* Output 4: HDMI-A disconnected
* Output 5: DisplayPort disconnected
* Output 6: HDMI-A disconnected
* Output 7: DisplayPort disconnected
* Output 8: HDMI-A disconnected
* Output 9: DisplayPort disconnected
* Output 10: HDMI-A disconnected
```

For things to work!

https://certification.canonical.com/hardware/202209-30616/submission/450558/

### Reproduction steps

_No response_

### Relevant log output

```shell
ERROR: ./src/platforms/common/server/kms-utils/kms_connector.cpp(127): Throw in function std::tuple<std::unique_ptr<_drmModeCrtc, std::function<void(_drmModeCrtc*)> >, int> {anonymous}::find_crtc_and_index_for_connector(const mir::graphics::kms::DRMModeResources&, const mir::graphics::kms::DRMModeConnectorUPtr&)
std::exception::what: Failed to find CRTC

sh: 1: /usr/share/apport/recoverable_problem: not found
```

Saviq:
Closes #4290

This reverts commit 63234359c2c1adbd7e730d5d41bc0c2521a9cbb0, reversing changes made to 0b0f87257f34d0afc31a04d891ead2affa32c132.

Related:
- #3466

## What's new?
This reverts:
- #3950

## How to test
Run on a device with AMD + Intel GPUs.

## Checklist
- [ ] ~~Adequate documentation added~~

### DIFF
--- a/src/platforms/gbm-kms/server/kms/real_kms_output.cpp
+++ b/src/platforms/gbm-kms/server/kms/real_kms_output.cpp
@@ -21,8 +21,6 @@
 #include "kms-utils/kms_connector.h"
 #include "mir/fatal.h"
 #include "mir/log.h"
-#include <drm_fourcc.h>
-#include <drm_mode.h>
 #include <string.h> // strcmp
 
 #include <boost/throw_exception.hpp>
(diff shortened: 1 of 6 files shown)


src/platforms/gbm-kms/server/kms/real_kms_output.cpp
```diff
@@ -632,17 +541,8 @@ void mgg::RealKMSOutput::update_from_hardware_state(
...
-
-    auto supported_formats = formats_for_output(drm_fd_, connector);
-    formats.reserve(supported_formats.size());
-    for (auto const& format : supported_formats)
-    {
-        if (auto mir_format = format.as_mir_format())
-        {
-            formats.push_back(*mir_format);
-        }
-    }
```
mattkae, line 645: What is this snippet even doing? This `formats` variable isn't used anywhere...

src/platforms/gbm-kms/server/kms/real_kms_output.cpp
```diff
@@ -583,42 +528,6 @@ std::vector<uint8_t> edid_for_connector(int drm_fd, uint32_t connector_id)
 
     return edid;
 }
-
-auto formats_for_output(int drm_fd, mgk::DRMModeConnectorUPtr const& connector) -> std::vector<mg::DRMFormat>
-{
-    auto [_, plane] = mgk::find_crtc_with_primary_plane(drm_fd, connector);
```
mattkae, line 589: @RAOF 

I am walking through this code path, and I _think_ that it can reasonably fail (i.e. it should not be an "exceptional" case), but I want to get your take.

1. There could be two connectors
2. Both connectors share the same encoder (e.g. the HDMI encoder)
3. Only one of them can use the encoder at a time
4. Bam! We hit our exception because one of the CRTCs could not get an encoder (as we saw is the case of zero encoders)

Is that case _really_ an exception? It seems more like something that can possibly happen

RAOF: Yeah, this reasoning makes sense to me.

This uses the infrastructure we have, but that infrastructure is for modesetting, and so has more constraints than we have here - we actually don't care whether or not the CRTC or encoder is in use, we just want to find one that *could* drive this connector so we can query its format properties.

If there is a case where there is legitimately *no* encoder that could work, then we need to make it clear there's something wrong and just ignore that connector. But assuming it is possible given *some* configuration we should just be able to pull out the formats without caring whether the *current* configuration can work.

src/platforms/gbm-kms/server/kms/real_kms_output.cpp
RAOF: Yeah, this should actually be assigning to `output.pixel_formats`

PR #4303, opened 2025-09-24
libmiral should have no public dependency on libmircommon

### LINKED ISSUE #4302: We are exposing mircommon in the libmiral interface
Our recommended and stable compositor API is supposed to be libmiral + libmircore.

But we've exposed use of several APIs from libmircommon in the libmiral headers.

The ones I've noted:

1. `#include <mir_toolkit/events/event.h>` (in bounce_keys.h and slow_keys.h) - that should have been `#include <miral/toolkit_event.h>`
2. `#include <mir/flags.h>` in window_info.h and window_specification.h

AlanGriffiths:
Closes #4302

## What's new?
Removes use of mircommon from the "public" API

## How to test
Build a downstream. (NB Miriway has unintentionally used `mir/log.h` from mircommon, so that breaks)


Saviq:
> NB Miriway has unintentionally used `mir/log.h` from mircommon, so that breaks

Looks like Frame did, too… should we fix there before landing this?

Saviq:
This should go away as well?

https://github.com/canonical/mir/blob/56d359d447441d643a8bceec740796731fd10b28/debian/control#L639

AlanGriffiths:
> Looks like Frame did, too… should we fix there before landing this?

Maybe we should move `mir/log.h` to mircore? (Open to discussion: that's a bit more of a thing than `mir/flags.h`)

Saviq (approved):
Yas!

PR #4304, opened 2025-09-24
pre-commit: run on all files

Saviq:
## What's new?
We only have hooks that we can safely run across all files.

## How to test
No significant changes pushed by pre-commit.

## Checklist
- [ ] ~~Adequate documentation added~~


COPYING.GPL3
jhenstridge, line 1: For the COPYING files, ideally I'd expect these to be identical to versions published by the FSF. This might mean excluding them from the list of files pre-commit checks.

For the whitespace differences in the GPL3 text specifically, it looks like the [current version from the FSF](https://www.gnu.org/licenses/gpl-3.0.txt) also fixes these whitespace issues anyway, along with changing the URLs to `https://`. Maybe it would be better to update it to that version?

Saviq: Done.

AlanGriffiths (approved):
I see red...

....whitespace

AlanGriffiths:
> ❌ [ERROR 5335] Multiple matches are possible between the source file(s) 'src/server/shell/surface_stack_wrapper.cpp' in TICS and the file(s) '/home/ubuntu/actions-runner/_work/mir/mir/src/server/shell/surface_stack_wrapper.cpp, /home/ubuntu/actions-runner/_work/mir/mir/src/server/shell/surface_stack_wrapper.cpp' in 'build/coverage.info, build/coverage.raw.info', and TICS cannot uniquely identify the correct match. As such, no analysis results are available for file 'src/server/shell/surface_stack_wrapper.cpp'.

1. Why do we have `src/server/shell/surface_stack_wrapper.cpp` under multiple roots?
2. Why is TICS looking at multiple roots?

Saviq:
> 1. Why do we have `src/server/shell/surface_stack_wrapper.cpp` under multiple roots?

We don't. It's the same path twice:
```
/home/ubuntu/actions-runner/_work/mir/mir/src/server/shell/surface_stack_wrapper.cpp
/home/ubuntu/actions-runner/_work/mir/mir/src/server/shell/surface_stack_wrapper.cpp
```

> 2. Why is TICS looking at multiple roots?

I reported this up to TIOBE, to me, looks like breakage on their side.

PR #4305, opened 2025-09-25
mrs::{Read,Write,RW}MappableBuffer: Standardise on std::byte for raw-memory type

RAOF:
This is both shorter than `unsigned char`, and more semantically clear than `unsigned char` (avoiding questions like: why isn't it `char`?)

## Checklist
- ~~[ ] Tests added and pass~~
- ~~[ ] Adequate documentation added~~
- ~~[ ] (optional) Added Screenshots or videos~~


robert-ancell (approved):
Seems clearer to me!

PR #4312, opened 2025-09-25
Tics fix coverage

Saviq:
## What's new?
Only generate LCOV output for TICS

## How to test
TICS happy.

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -109,7 +109,10 @@ jobs:
 
     - name: Measure coverage
       timeout-minutes: 20
-      run: cmake --build build --target coverage
+      run: |
+        cmake --build build --target coverage
+        # avoid duplicate coverage data
+        rm build/coverage*.info
 
     - name: Get changed files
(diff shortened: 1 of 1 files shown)


Saviq:
See https://github.com/canonical/mir/actions/runs/18014787691 for when TICS actually had something to concern itself with (I had #4305 cherry-picked here for that).

.github/workflows/tics.yml
```diff
@@ -109,7 +109,10 @@ jobs:
 
     - name: Measure coverage
       timeout-minutes: 20
-      run: cmake --build build --target coverage
+      run: |
+        cmake --build build --target coverage
+        # avoid duplicate coverage data
+        rm build/coverage*.info
```
mattkae, line 115: Should we force it in case it doesn't exist? Although I guess it would be good to know if the outputted files change

```suggestion
        rm -f build/coverage*.info
```

Saviq: Yeah we wanna know

PR #4316, opened 2025-09-26
Drop BufferDescriptor - it is accidental commonality

AlanGriffiths:


tarek-y-ismail (commented):
I think `BufferDescriptor` was intended as a way to reduce duplication. I don't think we get anything from expanding it into the classes that use it.

AlanGriffiths:
> I think `BufferDescriptor` was intended as a way to reduce duplication. I don't think we get anything from expanding it into the classes that use it.

In Smalltalk this might be a "mixin" type, but it doesn't meet C++'s more restricted type system.

As I said in #4293 _nothing_ is using the `BufferDescriptor` interface - everything is using a derived interface. As Chris says "none of `Mapping` or `Mappble` _IsA_ `BufferDescriptor`"

AlanGriffiths:
The documentation-check error (missing Coverage measurement) is fixed on `main`

RAOF (approved):
While this will make changing the return type of `format()` to `DRMFormat` marginally, harder, I don't think any code is going to actually act on an abstract “`BufferDescriptor`” concept and so it's reasonable to duplicate those methods into the actual types code will use.

PR #4317, opened 2025-09-26
Add `mrs::Mapping::descriptor`

tarek-y-ismail:
Related: #4293, #4316 

## What's new?
- Instead of `mrs::Mapping` extending `mrs::BufferDescriptor`. Which doesn't immediately make sense, it now has `descriptor()`, which allows access to the descriptor in a more intuitive way.

In my mind, this feels more correct than the existing relationship, but I'm not sure that the magnitude of changes justifies it.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -81,7 +81,7 @@ Description: Display server for Ubuntu - server library
  .
  Contains the shared library needed by server applications for Mir.
 
-Package: libmirplatform32
+Package: libmirplatform33
 Section: libs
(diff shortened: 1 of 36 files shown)


AlanGriffiths:
> In my mind, this feels more correct than the existing relationship, but I'm not sure that the magnitude of changes justifies it.

I don't think any of the types _HasA_ `BufferDescriptor` - they do _HasA_ `width`

PR #4318, opened 2025-09-26
Implement ext-data-control

### LINKED ISSUE #3660: Implement ext-data-control
Mir currently lacks an implementation of a protocol for clipboard services. A protocol for this [recently landed upstream](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/336). Mir should support this so that clipboard services can work when a desktop is running on Mir.

tarek-y-ismail:
Closes #3660

## What's new?

- Implements `ext-data-control-v1`, which enables clipboard services on Wayland.

## How to test

Note: the latest release of `wl-clipboard` (where `wl-copy` and `wl-paste ` come from) does not yet support `ext-data-control-v1`. For that, you will have to compile from source.

- Download any clipboard that supports the protocol. For example: [cliphist](https://github.com/sentriz/cliphist)
- Run miral-app with `--add-wayland-extension ext_data_control_manager_v1`
- Inside, run a terminal, and run `wl-paste -w cliphist store`.
- In another terminal, run `cliphist list`. The clipboard should be empty
  - Keep in mind that cliphist stores entries in a file, so entries from previous runs will appear. To wipe, run `cliphist wipe`
- run `wl-copy 'AAAAAAA'`
- Run `cliphist list`, you should see the ~screaming~ text stored in it.
- run `wl-copy 'BBBBBBBBBBBB'`
- Again, run `cliphist list`, you should see the :bee:s listed in it.

## Checklist

- [x] Tests added and pass https://github.com/canonical/wlcs/pull/371
- [x] Resolve TODOs inside the code
- [x] Self-review and minor cleanups

### DIFF
--- a/src/include/server/mir/scene/clipboard.h
+++ b/src/include/server/mir/scene/clipboard.h
@@ -56,7 +56,7 @@ public:
     /// Get the current copy-paste source.
     virtual auto paste_source() const -> std::shared_ptr<DataExchangeSource> = 0;
 
-    /// Run the given callback on the current paste source. 
+    /// Run the given callback on the current paste source.
     /// Will execute on the calling thread and the source will not be modified while
     /// the callback is being executed.
     virtual void do_with_paste_source(


src/server/frontend_wayland/wayland_connector.cpp
tarek-y-ismail, line 1: Shouldn't this have been caught with the mega precommit commit a while ago? Not sure if it should go in or be reverted...

src/server/frontend_wayland/wl_seat.h
tarek-y-ismail, line 1: Not necessary anymore with the use of clipboards

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,408 @@
...
+        }
+    };
+
+    struct State
+    {
+        State(std::shared_ptr<ms::Clipboard> const& clipboard, std::shared_ptr<ms::Clipboard> const& primary_clipboard);
+
+        std::shared_ptr<ms::Clipboard> const clipboard, primary_clipboard;
+        wayland::Weak<DataControlSourceV1> current_source, current_primary_source;
+    };
```
tarek-y-ismail: Should probably be moved to the manager instance or the top of the file.

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,408 @@
...
+            send_offer_event(mime);
+        }
+    }
+
+    void receive(std::string const& mime, mir::Fd fd) override
+    {
+        parent.value().receive_from_current_source(mime, fd, is_primary);
+    }
+
+    wayland::Weak<DataControlDeviceV1> parent;
```
tarek-y-ismail: Should be const

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,408 @@
...
+{
+    new Instance{resource, state};
+}
+
+mf::DataControlManagerV1::Instance::Instance(
+    struct wl_resource* resource,
+    std::shared_ptr<DataControlDeviceV1::State> const& state) :
+    wayland::DataControlManagerV1(resource, Version<1>{}),
+    state{state}
+
```
tarek-y-ismail: Useless whitespace

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,408 @@
...
+        if (is_primary && state->current_primary_source)
+        {
+            state->current_primary_source.value().send_cancelled_event();
+            state->current_primary_source = {};
+        }
+        else if (!is_primary && state->current_source)
+        {
+            state->current_source.value().send_cancelled_event();
+            state->current_source = {};
+        }
```
tarek-y-ismail: Probably should be moved after all the checks right before when we set the new source

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+        // If the notification comes from a different seat, ignore it
+        if(data_exchange_source->seat != this->seat)
+                return;
+
+        // Make sure the device is not destryoed
```
tarek-y-ismail:
```suggestion
        // Make sure the device is not destroyed
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,408 @@
...
+        if (!data_exchange_source->device)
+            return;
+
+        // If the listener is the same as the source, do nothing. We're trying
```
tarek-y-ismail: Should be more precise: "If the device being used to set the clipboard is the same as the one receiving the notification, do nothing"

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,408 @@
...
+    private:
+        void create_data_source(struct wl_resource* id) override;
+        void get_data_device(struct wl_resource* id, struct wl_resource* seat) override;
+
+        std::shared_ptr<mf::DataControlDeviceV1::State> const state;
+    };
+
+    void bind(wl_resource*) override;
+
+    std::shared_ptr<mf::DataControlDeviceV1::State> const state;
```
tarek-y-ismail: Should probably be `Synchronised` as well since multiple clients can access this concurrently

src/server/frontend_wayland/wayland_connector.cpp
Saviq: Yeah not sure why this is skipped over :shrug:

Keep it!

src/server/frontend_wayland/wayland_connector.cpp
```diff
@@ -444,7 +444,7 @@ mf::WaylandConnector::~WaylandConnector()
                 cleanup_promise->set_value();
             });
         cleanup_done.get();
-        
+
```
mattkae: Sporadic change, I'd just revert this file :)

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,397 @@
...
+struct DataControlStateV1
+{
+    DataControlStateV1(
+        std::shared_ptr<ms::Clipboard> const& clipboard, std::shared_ptr<ms::Clipboard> const& primary_clipboard) :
+        clipboard{clipboard},
+        primary_clipboard{primary_clipboard}
+    {
+    }
+
+    std::shared_ptr<ms::Clipboard> const clipboard, primary_clipboard;
```
mattkae: Does this need to be in the synchronized state necessarily? Just asking because it feels that these won't be reassigned and won't need to be accessed behind a mutex

mattkae: Also - we tend not to do the "comma" approach. I would split these two up. Same for below.

```suggestion
    std::shared_ptr<ms::Clipboard> const clipboard;
    std::shared_ptr<ms::Clipboard> const primary_clipboard;
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,397 @@
...
+    DataControlManagerV1(
+        struct wl_display* display,
+        std::shared_ptr<ms::Clipboard> const& clipboard,
+        std::shared_ptr<ms::Clipboard> const& primary_clipboard) :
+        wayland::DataControlManagerV1::Global(display, Version<1>{}),
+        state{std::make_shared<mir::Synchronised<DataControlStateV1>>(DataControlStateV1{clipboard, primary_clipboard})}
+    {
+    }
+
+    ~DataControlManagerV1() = default;
```
mattkae:
```suggestion
    ~DataControlManagerV1() override = default;
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+    std::shared_ptr<ms::Clipboard> const clipboard, primary_clipboard;
+    wayland::Weak<DataControlSourceV1> current_source, current_primary_source;
+};
+using SyncedState = mir::Synchronised<DataControlStateV1>;
+
+class DataControlDeviceV1 : public wayland::DataControlDeviceV1
+{
+public:
+    struct ClipboardObserver : public ms::ClipboardObserver
```
mattkae, line 174: Can this be `private`?

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,397 @@
...
+    void set_selection(std::optional<struct wl_resource*> const& data_control_source) override
+    {
+        auto s = state->lock();
+        if (data_control_source)
+        {
+            auto source = mf::DataControlSourceV1::from(*data_control_source);
+            if (!source)
+                return;
+
+            source->finalize();
```
mattkae: Nit: I think that the `DataControlDeviceV1` should be throwing this exception instead of the source. The exception being thrown feels like an implementation detail of the device.

mattkae: I would refactor to:

```
if (!source->try_finalize())
    throw ...
```

mattkae: Or maybe rename that to `claim()`?

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,397 @@
...
+        }
+        void end_of_dnd_gesture() override
+        {
+        }
+    };
+
+    // Impl seperated out because it depends on `DataControlOfferV1`, which itself depends on this class
+    DataControlDeviceV1(struct wl_resource* id, std::shared_ptr<SyncedState> const& state, WlSeat const* seat);
+
+    ~DataControlDeviceV1()
```
mattkae:
```suggestion
    ~DataControlDeviceV1() override
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+            source->finalize();
+
+            auto const data_exchange_source =
+                std::make_shared<DataExchangeSource>(wayland::Weak{source}, wayland::Weak{this}, seat);
+
+            if(s->current_primary_source)
+                s->current_primary_source.value().send_cancelled_event();
+            s->primary_clipboard->set_paste_source(data_exchange_source);
+            s->current_primary_source = data_exchange_source->source;
```
mattkae: Should this also invalidate `s->current_source`? And vice versa for above

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,397 @@
...
+    }
+}
+
+void mf::DataControlDeviceV1::on_clipboard_set(std::shared_ptr<ms::DataExchangeSource> const& source, bool is_primary)
+{
+    if (auto data_exchange_source = std::dynamic_pointer_cast<mf::DataExchangeSource>(source))
+    {
+        // If the notification comes from a different seat, ignore it
+        if(data_exchange_source->seat != this->seat)
+                return;
```
mattkae:
```suggestion
        if(data_exchange_source->seat != this->seat)
            return;
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+        // Make sure the device is not destroyed
+        if (!data_exchange_source->device)
+            return;
+
+        // If the device receiving the notification is the same as the device
+        // initiating the notification, do nothing. We're trying to notify
+        // _other_ devices listening for ext-data-control
+        if(data_exchange_source->device.is(*this))
```
mattkae:
```suggestion
        if (data_exchange_source->device.is(*this))
```

src/server/frontend_wayland/data_control_v1.cpp
tarek-y-ismail: I don't believe the primary clipboard should affect the "normal one", but I might be wrong

src/server/frontend_wayland/data_control_v1.cpp
tarek-y-ismail: > I think that the DataControlDeviceV1 should be throwing this exception instead of the source.

Yup, even the protocol has the error in the device.

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,404 @@
...
+    {
+        return mime_types_;
+    }
+
+    auto try_finalize() -> bool
+    {
+        auto const not_finalized = finalized_ != true;
+        finalized_ = true;
+
+        return not_finalized;
```
mattkae: Nit: This feels simpler to me?

```suggestion
        if (!finalized_)
        {
            finalized_ = true;
            return true;
        }
        
        return false;
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,404 @@
...
+        }
+        void end_of_dnd_gesture() override
+        {
+        }
+    };
+
+    auto data_exchange_source_from_source(std::optional<struct wl_resource*> const& data_control_source)
+        -> std::shared_ptr<ms::DataExchangeSource>
+    {
+        if(auto source = mf::DataControlSourceV1::from(*data_control_source))
```
mattkae:
```suggestion
        if (auto source = mf::DataControlSourceV1::from(*data_control_source))
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,404 @@
...
+        }
+        void drag_n_drop_source_cleared(std::shared_ptr<ms::DataExchangeSource> const&) override
+        {
+        }
+        void end_of_dnd_gesture() override
+        {
+        }
+    };
+
+    auto data_exchange_source_from_source(std::optional<struct wl_resource*> const& data_control_source)
```
mattkae: Can this be a `static` method?

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,404 @@
...
+        {
+            return wl_data_source->make_source();
+        }
+
+        if(auto primary_data_source = mf::PrimarySelectionSource::from(*data_control_source))
+        {
+            return primary_data_source->make_source();
+        }
+
+        std::unreachable();
```
mattkae: This seems dangerous :) Should we throw a protocol error here in this case? In its current state, a malformed client can bring down the whole server

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,404 @@
...
+        }
+
+        auto const paste_source = shared_state->clipboard->paste_source();
+        if (!is_primary && paste_source)
+        {
+            paste_source->initiate_send(mime, fd);
+            return;
+        }
+
+        std::unreachable();
```
mattkae: This also seems dangerous. We should probably not potentially bring down the whole server due to a logic error?

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,404 @@
...
+        send_finished_event();
+    }
+
+    void receive_from_current_source(std::string const& mime, mir::Fd fd, bool is_primary)
+    {
+        // FIXME time-of-check vs time-of-use bug
+        // If the clipboard is changed between checking and using it (cleared),
+        // it might cause a crash.
+        // Workaround: add `Clipboard::use_current_clipboard(Function&&)` which
+        // would lock around the execution of the function.
```
mattkae: Should we fix this before landing this PR? Feels like potentially crashing the server isn't a good idea :smile:

src/server/frontend_wayland/data_control_v1.cpp
tarek-y-ismail: That was the first idea I had. I'd have to pass the control device resource for the protocol error and the seat for when the data exchange source is being created, which I felt was too much for a small utility function

src/server/frontend_wayland/data_control_v1.cpp
AlanGriffiths: Can you expand on the issue here:
1. Where is the check? 
2. Where is the use? And,
3. how can the clipboard change between?

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,424 @@
...
+#include "data_control_v1.h"
+#include "primary_selection_v1.h"
+#include "wl_data_source.h"
+#include "wl_seat.h"
+
+#include "mir/scene/clipboard.h"
+#include "mir/scene/data_exchange.h"
+#include "mir/wayland/protocol_error.h"
+
+#define MIR_LOG_COMPONENT "ext-data-control-v1"
```
AlanGriffiths:
```suggestion
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,424 @@
...
+    {
+        shared_state->primary_clipboard->do_with_paste_source(
+            [is_primary, mime, fd](auto const& primary_paste_source)
+            {
+                if (is_primary && primary_paste_source)
+                {
+                    primary_paste_source->initiate_send(mime, fd);
+                    return;
+                }
+            });
```
AlanGriffiths: Is there any point to constructing the lambda and calling `do_with_paste_source()` when `is_primary` is false?

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+        shared_state->clipboard->do_with_paste_source(
+            [is_primary, mime, fd](auto const& paste_source)
+            {
+                if (!is_primary && paste_source)
+                {
+                    paste_source->initiate_send(mime, fd);
+                    return;
+                }
+            });
```
AlanGriffiths: s there any point to constructing the lambda and calling do_with_paste_source() when is_primary is true?

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+        mir::log_warning("Attempt to receive from invalid source");
```
AlanGriffiths: This will be logged on every invocation as there are no early returns

AlanGriffiths:
Also...

https://github.com/canonical/mir/blob/4acd1d94462a60519286d3545687348598ac2646/src/server/frontend_wayland/wayland_default_configuration.cpp#L78-L79

src/server/frontend_wayland/data_control_v1.cpp
tarek-y-ismail: The check is the if statement where we check if the paste source is set or not. The use is after when we call `initiate_send` on the value we just checked. If the set source changes in between, won't that be an issue? I might be overly cautious here but I'm not sure. 

Thinking about it, if the source changes, then our pointer won't be null but will be out of date. If it's cleared, it's the same case, out of date.

include/miral/miral/wayland_extensions.h
```diff
@@ -163,6 +163,9 @@ class WaylandExtensions
     /// Allows clients to act as a screen lock.
     static char const* const ext_session_lock_manager_v1;
 
+    /// Allows clients to manage the current selection and act as a clipboard manager.
+    static char const* const ext_data_control_manager_v1;
```
AlanGriffiths, line 168:
```suggestion
    /// Allows clients to manage the current selection and act as a clipboard manager.
    /// \remark Since MirAL 5.6
    static char const* const ext_data_control_manager_v1;
```

src/server/scene/basic_clipboard.cpp
```diff
@@ -26,6 +26,13 @@ auto ms::BasicClipboard::paste_source() const -> std::shared_ptr<DataExchangeSou
     return paste_source_;
 }
 
+void ms::BasicClipboard::do_with_paste_source(
+    std::move_only_function<void(std::shared_ptr<DataExchangeSource> const&)>&& fn)
```
AlanGriffiths: TICS has a point

src/include/server/mir/scene/clipboard.h
```diff
@@ -55,6 +56,11 @@ class Clipboard : public ObserverRegistrar<ClipboardObserver>
     /// Get the current copy-paste source.
     virtual auto paste_source() const -> std::shared_ptr<DataExchangeSource> = 0;
 
+    /// Do something with the current paste source. Must guarantee thread
+    /// safety.
```
AlanGriffiths: > Must guarantee thread safety

In general, code should always guarantee thread safety: What are the specific considerations being documented here? E.g. 

- Is the `fn` code executed on the current thread? 
- Or another thread?
- Is there a lock held or need acquiring for something?

AlanGriffiths:
> * Download any clipboard that supports the protocol. For examplel: [cliphist](https://github.com/sentriz/cliphist)
> 
> * Run miral-app with `--add-wayland-extension ext_data_control_manager_v1`
> 
> * Inside, run a terminal, and run `wl-paste -w cliphist store`.

With the version of `wl-paste` in 25.10 I get an "requires a compositor that supports the **wlroots** data-control protocol". Maybe there needs to be a bit more clarity in the instructions.

tarek-y-ismail:
> With the version of `wl-paste` in 25.10 I get an "requires a compositor that supports the **wlroots** data-control protocol". Maybe there needs to be a bit more clarity in the instructions.

My bad, the latest release does not have support for `ext-data-control-v1`, so I had to compile from source (which is quite simple). 

With the changes pushed in the last couple of days, our implementation should work with the core Wayland and primary selection protocols, so you should be able to copy from Wayland apps into cliphist.

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,436 @@
...
+    if(source)
+    {
+        auto new_offer = new DataControlOfferV1(wayland::Weak{this}, source->mime_types(), is_primary);
+        // No need to lock, we already lock at the start of `set_selection` and `set_primary_selection`
+        if (is_primary)
+            send_primary_selection_event({new_offer->resource});
+        else
+            send_selection_event({new_offer->resource});
+    } else
+    {
```
mattkae:
```suggestion
    }
    else
    {
```

src/include/server/mir/scene/clipboard.h
AlanGriffiths: > Must guarantee thread safety

In general, code should always guarantee thread safety: So, can we expand on the specific considerations being documented here?

AlanGriffiths (changes requested):

src/include/server/mir/scene/clipboard.h
```diff
@@ -55,6 +56,12 @@ class Clipboard : public ObserverRegistrar<ClipboardObserver>
     /// Get the current copy-paste source.
     virtual auto paste_source() const -> std::shared_ptr<DataExchangeSource> = 0;
 
+    /// Run the given callback on the current paste source. Must run on the
+    /// same thread as the caller and hold the paste source lock to prevent any
+    /// modifications from occuring while the callback is being executed.
```
AlanGriffiths:
```suggestion
    /// Run the given callback on the current paste source. 
    /// Will execute on the calling thread and the source will not be modified while
    /// the callback is being executed.
```

src/server/frontend_wayland/data_control_v1.cpp
```diff
@@ -0,0 +1,416 @@
...
+#include "wl_seat.h"
+
+#include "mir/scene/clipboard.h"
+#include "mir/scene/data_exchange.h"
+#include "mir/wayland/protocol_error.h"
+
+#include "mir/log.h"
+
+#include <cassert>
+
```
AlanGriffiths: Not used.

NB using the `assert` macro is generally a bad idea for our code - it gives different behaviour depending on `NDEBUG` and we rarely want that

tarek-y-ismail: Thanks for pointing it out, I think I was debugging something during implementation and just forgot it. But old habits die hard...

tarek-y-ismail:
@AlanGriffiths I'm just about to push a cleanup of the commits. I'll merge when it's done

PR #4325, opened 2025-10-01
Handle potentially transparent surfaces in `renderable_is_occluded`

### LINKED ISSUE #4185: Throttle down occluded surfaces
### Discussed in https://github.com/canonical/mir/discussions/2914

<div type='discussions-op-text'>

<sup>Originally posted by **Saviq** August 25, 2022</sup>
We don't stop occluded surfaces from drawing. Related to (but not the same as) #2007.

Can be seen with e.g. `WAYLAND_DEBUG=1 mate-terminal --disable-factory`, `while true; do sleep .01; echo .; done` and putting another window in front of it. You can still observe surfaces being submitted.</div>

tarek-y-ismail:
Closes #4185

## What's new?

- Handled `wl_surface.set_opaque_region`. Add getter in `Surface` and `WlSurface`
- Made `BasicSurface::wayland_surface` a const method
- Opaque regions are now used in `renderable_is_occluded` if the renderable is shaped.
- Modified `filter_occlusions_from` to take ownership of the element list and return the occluded and visible elements, instead taking a non-const reference to the element list, removing the occluded elements, and returning them.

## How to test

*borrowed from the original issue and modified*
- Run mir with `WAYLAND_DEBUG=client`
- Run two terminals
- In the first terminal, run `i=0; while true; do sleep 0.1; echo $i; i=$(($i+1)); done`
- Observe the Wayland log, it should print about 10 messages per second
- Completely cover the first terminal with the second by changing its position and size
- Observe the Wayland log again, it should shortly stop printing

## Checklist

- [x] Instead of throttling down to 10 FPS, kgx (and gnome-terminal) just stop submitting frames. Maybe we should completely stop posting frame callbacks in that case
- [x] ~The changes to `Surface` and `Renderable` feel to intrusive. Maybe there is another, better path?~
  - ~The only other path I've been able to find was through `RenderingTracker`, but the attribute it sets seems like legacy code and shouldn't be depended on.~
  - Occlusion checking was already implemented, it just didn't handle shaped/potentially transparent surfaces correctly 
- [x] Tests added and pass

https://github.com/user-attachments/assets/f75aca25-462e-4be2-9def-44d98e619968

The case in this video wouldn't previously work as we didn't handle shaped surfaces (opaque surface, but potentially transparent pixel format) with opaque regions supplied.

### DIFF
--- a/src/server/compositor/occlusion.cpp
+++ b/src/server/compositor/occlusion.cpp
@@ -80,9 +80,10 @@ bool renderable_is_occluded(
 std::pair<OccludedElementSequence, SceneElementSequence> mir::compositor::split_occluded_and_visible(
     SceneElementSequence&& elements, Rectangle const& area)
 {
-    OccludedElementSequence occluded;
-    SceneElementSequence visible{std::move(elements)};
+    std::pair<OccludedElementSequence, SceneElementSequence> result{{}, std::move(elements)};
     std::vector<Rectangle> coverage;
+
+    auto& [occluded, visible] = result;
 
     auto it = visible.rbegin();
     while (it != visible.rend())
@@ -99,5 +100,5 @@ std::pair<OccludedElementSequence, SceneElementSequence> mir::compositor::split_
         }
     }
 
-    return {std::move(occluded), std::move(visible)};
+    return result;
 }


Saviq:
> If a surface is occluded according to `renderable_is_occluded`, surface updates are throttled to about 10 FPS

[The protocol](https://wayland.app/protocols/wayland#wl_surface:request:frame) says:
> A server should avoid signaling the frame callbacks if the surface is not visible in any way, e.g. the surface is off-screen, or completely obscured by other opaque surfaces.

Can't we simply delay the callback until such a time the surface is _not_ occluded?

src/server/compositor/occlusion.cpp
```diff
@@ -54,16 +54,33 @@ bool renderable_is_occluded(
         }
     }
 
-    if (!occluded && renderable.alpha() == 1.0f && !renderable.shaped())
+    auto const opaque_region = renderable.opaque_region();
```
tarek-y-ismail: Can be moved to inside the if statement

src/server/compositor/occlusion.cpp
```diff
@@ -54,16 +54,33 @@ bool renderable_is_occluded(
         }
     }
 
-    if (!occluded && renderable.alpha() == 1.0f && !renderable.shaped())
```
tarek-y-ismail, line 57: should only add to the coverage if not occluded

src/server/graphics/software_cursor.cpp
```diff
@@ -134,6 +134,11 @@ class mg::detail::CursorRenderable : public mg::Renderable
         scale = new_scale;
     }
 
+    auto opaque_region() const -> std::optional<geom::Rectangles> override
+    {
+        return std::nullopt;
+    }
```
tarek-y-ismail, line 140: Should return `screen_position` like its siblings

src/server/scene/CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ ADD_LIBRARY(
   ${CMAKE_SOURCE_DIR}/src/include/server/mir/scene/surface_observer.h
 )
 
+# TODO: Having to do this suggests that `Surface::wayland_surface` might need some changes
+include_directories(../frontend_wayland)
+
```
tarek-y-ismail: 😞

tests/include/mir/test/doubles/fake_renderable.h
```diff
@@ -119,6 +119,11 @@ class FakeRenderable : public graphics::Renderable
         return std::nullopt;
     }
 
+    auto opaque_region() const -> std::optional<geometry::Rectangles> override
+    {
+        return std::nullopt;
+    }
```
tarek-y-ismail, line 137: Probably should not return `nullopt`. Instead, it should use the information passed to the renderable on construction to figure out whether it's opaque or not

tests/platform_test_harness/graphics_platform_test_harness.cpp
```diff
@@ -454,6 +454,12 @@ void basic_software_buffer_drawing(
         {
             return std::nullopt;
         }
+
+        auto opaque_region() const -> std::optional<mir::geometry::Rectangles> override
+        {
+            return std::nullopt;
```
tarek-y-ismail: Should check `shaped` first

tarek-y-ismail: Might be dead code

src/server/compositor/occlusion.cpp
```diff
@@ -54,16 +54,36 @@ bool renderable_is_occluded(
...
+        if (renderable.shaped())
+        {
+            if (auto const opaque_region = renderable.opaque_region())
+            {
+                // TODO: If a window is made out of multiple opaque regions, adding
+                // each region alone to the coverage list will not work.
+                // Separately, each region might not occlude a window, but
+                // together, they might.
+                //
+                // For now, I think their bounding rect will do.
```
AlanGriffiths: Oh no it won't ;)

tarek-y-ismail: What do you mean? I don't think having perfect occlusion checking is really worth going down the rabbit how (again)

AlanGriffiths: I mean that it will lead to bad results: Imagine a "picture frame" - where you can see through the middle. The "bounding rectangle" is the whole frame - we should not treat that as opaque.

AlanGriffiths: If we're not doing "perfect" then we should err on the side of "not occluded"

src/server/scene/basic_surface.cpp
```diff
@@ -863,6 +891,18 @@ mg::RenderableList ms::BasicSurface::generate_renderables(mc::CompositorID id) c
...
         if (info.stream->has_submitted_buffer())
         {
+            auto const opaque_region = [&state, &info, &id, this] -> std::optional<geom::Rectangles>
+            {
+                auto const shaped = info.stream->next_submission_for_compositor(id);
+                if (state->surface_alpha == 1.0f && !shaped)
+                    return std::nullopt;
+
+                if (auto const ws = this->wayland_surface())
+                    return ws.value().opaque_region();
```
AlanGriffiths: We can't use this function on the compositing thread. We should simply store any opacity information on the surface. (And not touch the frontend types)

src/server/scene/CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ ADD_LIBRARY(
   ${CMAKE_SOURCE_DIR}/src/include/server/mir/scene/surface_observer.h
 )
 
+# TODO: Having to do this suggests that `Surface::wayland_surface` might need some changes
+include_directories(../frontend_wayland)
```
AlanGriffiths: Agreed. DO NOT DO THIS

src/server/compositor/occlusion.cpp
```diff
@@ -54,16 +54,31 @@ bool renderable_is_occluded(
...
+    }
     return occluded;
 }
 }
 
-SceneElementSequence mir::compositor::filter_occlusions_from(
-    SceneElementSequence& elements,
-    Rectangle const& area)
+std::pair<OccludedElementSequence, SceneElementSequence> mir::compositor::filter_occlusions_from(
+    SceneElementSequence&& elements, Rectangle const& area)
```
AlanGriffiths, line 81: TICS has a point: We don't move from `elements`, so why are we passing it with a universal reference? (And supplying it with `std::move()`).

src/server/compositor/occlusion.cpp
```diff
@@ -83,5 +98,5 @@ SceneElementSequence mir::compositor::filter_occlusions_from(
         }
     }
 
-    return occluded;
+    return {occluded, elements};
```
AlanGriffiths: This is going to defeat any optimizations. We will copy both `occluded` and `elements` - which is potentially expensive.

I'm guessing you were intending something like this but lost the thread at some point?

```c++
std::pair<OccludedElementSequence, SceneElementSequence> mir::compositor::filter_occlusions_from(
    SceneElementSequence&& elements, Rectangle const& area)
{
    std::pair<OccludedElementSequence, SceneElementSequence> result{{}, std::move(elements)};
    std::vector<Rectangle> coverage;

    auto it = result.second.rbegin();
    while (it != result.second.rend())
    {
        auto const renderable = (*it)->renderable();
        if (renderable_is_occluded(*renderable, area, coverage))
        {
            result.first.insert(result.first.begin(), *it);
            it = SceneElementSequence::reverse_iterator(result.second.erase(std::prev(it.base())));
        }
        else
        {
            it++;
        }
    }

    return result;
}
```

But I'm not sure this gains anything over the pre-exisiting code that modified `elements` in place.

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -326,8 +326,18 @@ void mf::WlSurface::frame(wl_resource* new_callback)
...
+        auto opaque_region = geom::Rectangles{}; // Unfortunately no vector ctor :(
+        for (auto const& subregion : WlRegion::from(*r)->rectangle_vector())
+            opaque_region.add(subregion);
+
+        pending.opaque_region.emplace(opaque_region);
+    }
+    else
+    {
+        pending.opaque_region.reset();
+    }
```
AlanGriffiths: This seems overcomplicated: why introduce `r`? Why have two branches assigning to the same target?
```suggestion
    pending.opaque_region = region.transform([](auto*r)
        {
            geom::Rectangles opaque_region;
            for (auto const& subregion : WlRegion::from(r)->rectangle_vector())
                opaque_region.add(subregion);
            return opaque_region;
        });
```

src/server/scene/basic_surface.cpp
```diff
@@ -742,16 +747,35 @@ class SurfaceSnapshot : public mg::Renderable
...
+                {
+                    auto const translated_top_left = geom::Point{
+                        r.top_left.x + geom::as_delta(screen_position_.top_left.x),
+                        r.top_left.y + geom::as_delta(screen_position_.top_left.y),
+                    };
+                    region.add(geom::Rectangle{translated_top_left, r.size});
+                }
+
+                return region;
+            }()}
```
AlanGriffiths: 1. Have you forgotten about `optional.transform()`?
2. This lambda has enough logic to consider separating out as a utility function.

src/server/symbols.map
```diff
@@ -1645,6 +1645,8 @@ MIR_SERVER_INTERNAL_2.23 {
 global:
   extern "C++" {
     mir::time::AlarmFactory::create_repeating_alarm*;
+    mir::scene::BasicSurface::set_opaque_region*;
+    non-virtual?thunk?to?mir::scene::BasicSurface::set_opaque_region*;
```
AlanGriffiths: Sorry! 2.23 has been branched

src/server/compositor/occlusion.cpp
tarek-y-ismail: I've seen that, but got distracted by other fixes in the code. Thanks for the poke 👉

src/server/compositor/occlusion.cpp
tarek-y-ismail: The original code was slightly confusing for me as it wasn't clear at a glance if the vector we pass is modified or not. A better interface in my opinion is to move the input into the function, and return both the occluded and visible elements separately. I thought the compiler would auto-move the returned values (as clangd usually tells me a move is unnecessary). Though I'm not familiar with this.

src/server/symbols.map
```diff
@@ -1645,8 +1645,15 @@ MIR_SERVER_INTERNAL_2.23 {
...
 } MIR_SERVER_INTERNAL_2.22;
+
+
+MIR_SERVER_INTERNAL_2.24 {
+global:
+  extern "C++" {
+    mir::scene::BasicSurface::set_opaque_region*;
+    non-virtual?thunk?to?mir::scene::BasicSurface::set_opaque_region*;
+local: *;
+} MIR_SERVER_INTERNAL_2.23;
```
AlanGriffiths: Syntax error! BUT changing the `BasicSurface` vtab is an ABI break, so...

src/server/compositor/default_display_buffer_compositor.cpp
```diff
@@ -60,28 +60,28 @@ bool mc::DefaultDisplayBufferCompositor::composite(mc::SceneElementSequence&& sc
     report->began_frame(this);
 
     auto const& view_area = display_sink.view_area();
-    auto const& occlusions = mc::filter_occlusions_from(scene_elements, view_area);
+    auto [occluded_elements, visible_elements] = mc::filter_occlusions_from(std::move(scene_elements), view_area);
```
AlanGriffiths: You've changed the semantics and signature of `filter_occlusions_from()` but not the name. The old name doesn't fit anymore as it doesn't just remove occlusions from `scene_elements`. Maybe `split_occluded_and_visible()`?

CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
 set(MIR_VERSION_MAJOR 2)
-set(MIR_VERSION_MINOR 23)
+set(MIR_VERSION_MINOR 24)
```
AlanGriffiths: We usually leave bumping the minor version for the release branch

src/server/compositor/default_display_buffer_compositor.cpp
tarek-y-ismail: I though the change in the function's signature were good enough, but more info >>> less info :+1:

include/platform/mir/graphics/renderable.h
```diff
@@ -18,11 +18,11 @@
 #define MIR_GRAPHICS_RENDERABLE_H_
 
 #include <optional>
-#include <mir/geometry/rectangle.h>
 #include <glm/glm.hpp>
 #include <memory>
 #include <vector>
 
+#include "mir/geometry/rectangles.h"
```
AlanGriffiths:
```suggestion
#include <mir/geometry/rectangles.h>
```

src/server/compositor/occlusion.cpp
```diff
@@ -54,34 +54,50 @@ bool renderable_is_occluded(
...
+            it = SceneElementSequence::reverse_iterator(visible.erase(std::prev(it.base())));
         }
         else
         {
             it++;
         }
     }
 
-    return occluded;
+    return {std::move(occluded), std::move(visible)};
```
AlanGriffiths: [non-blocking] You can avoid a move by constructing a named return value at the top and using the members of that instead of creating and moving separate local variables. (The explicit moves here prevents the compiler optimizing those away.)

tarek-y-ismail: Oh, so that's why the code in [this comment](https://github.com/canonical/mir/pull/4325#discussion_r2421142487) looked weird to me. I copied the change over with one small difference. Do you think that might throw the compiler off?

AlanGriffiths: There are several things in play here:

1. the original code didn't copy or move the containers. (It did do a lot of moving within and between the containers.)
2. You introduced moving things into and out of the function and created temporary named containers.

The "named return value optimization" [NRVO] is guaranteed (since C++-23 IIRC) and means that a named return value can be constructed "in place" avoiding any copies or moves. The `std::move()`  commented forces either a move or copy to happen but can allow a move to replace a copy.

FWIW the original logic looks fine to me but I doubt your version is measurably less efficient.

CMakeLists.txt
```diff
@@ -25,8 +25,8 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
 set(MIR_VERSION_MAJOR 2)
-set(MIR_VERSION_MINOR 23)
-set(MIR_VERSION_PATCH 0)
+set(MIR_VERSION_MINOR 22)
+set(MIR_VERSION_PATCH 2)
```
AlanGriffiths: Please drop the commits that change the version

AlanGriffiths (approved):
Thanks!

Issue #3659, opened 2024-11-01
Replace wlr-screencopy with ext-image-copy-capture


Conan-Kudo:
Earlier this year, `wlr-screencopy` was replaced with a pair of protocols that substantially improve the experience:

* [`ext-image-copy-capture`](https://wayland.app/protocols/ext-image-copy-capture-v1)
* [`ext-image-capture-source`](https://wayland.app/protocols/ext-image-capture-source-v1)

The major improvement here is that now you can capture toplevels specifically, so screencasting just a single window is possible.

You currently have screencopy implemented for wayvnc, but the upcoming version will use these protocols (as [the developer is the one that made the new protocols](https://andri.yngvason.is/making-a-wayland-screen-capturing-protocol.html)).

This will also be necessary for fully enabling portals-based screen/window capture.

Conan-Kudo:
[WayVNC 0.9 was released today](https://github.com/any1/wayvnc/releases/tag/v0.9.0), so now there's a client that uses the new Wayland protocol.

AlanGriffiths:
We should add `ext-image-copy-capture` WITHOUT removing `wlr-screencopy`

Conan-Kudo:
At this point, we have a partial implementation of the new protocol, but toplevel/window capture support is not yet present. That's the main gap for closing this ticket.

Saviq:
> That's the main gap for closing this ticket.

We view it slightly differently - the _migration_ (AKA feature parity) will be done once cursors land. Toplevels are a new feature, which happens to require quite some thinking (and a pre-composition pass for subsurfaces at least, but maybe also children?).

PR #4326, opened 2025-10-02
Partial implementation of ext-image-capture-source and ext-image-copy-capture Wayland extensions

jhenstridge:
https://github.com/user-attachments/assets/1f657e54-d3dd-4dc5-8283-09a0798698de

Related: #3659, MIRENG-1319

## What's new?

This implements parts of the ext-image-capture-source and ext-image-copy-capture extensions:

1. Only the `ext_output_image_capture_source_manager_v1` global for creating sources: no toplevel capture yet.
2. The `create_pointer_cursor_session` method is not implemented. The cursor can be composited onto the captured image, but separate cursor tracking is still to be done.

## How to test

For simple screenshots, start Miral with the extensions enabled:

```sh
./bin/miral-app --add-wayland-extensions ext_output_image_capture_source_manager_v1:ext_image_copy_capture_manager_v1
```

Build [grim](https://gitlab.freedesktop.org/emersion/grim) >= 1.5 (the version in the archive is too old). A screenshot with the cursor overlaid can be taken with:

```sh
WAYLAND_DISPLAY=wayland-1 ./grim -c screenshot.png
```

For repeated capture, install wayvnc (archive version is fine), then run:

```sh
WAYLAND_DISPLAY=wayland-1 wayvnc --verbose --disable-input
```

Use a VNC client to connect to `localhost:5900`. If wayvnc is run under wayland-debug, we can see that the capture rate is metered by the server: waiting for changes to the screen.

It will also respond to output reconfiguration when resizing the miral-app window, although wayvnc tears unnecessarily tears down the capture session in this case.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/ext_image_capture_v1.cpp
+++ b/src/server/frontend_wayland/ext_image_capture_v1.cpp
@@ -159,7 +159,7 @@ private:
     void apply_damage(std::optional<geom::Rectangle> const& damage);
 
     wayland::Weak<OutputGlobal> const output;
-    uint32_t const options;
+    bool overlay_cursor;
     std::shared_ptr<ExtImageCaptureV1Ctx> const ctx;
 
     DamageAmount damage_amount = DamageAmount::full;
@@ -306,7 +306,7 @@ mf::ExtImageCopyCaptureSessionV1::ExtImageCopyCaptureSessionV1(
     : wayland::ImageCopyCaptureSessionV1{resource, Version<1>()},
       screen_shooter(ctx->screen_shooter_factory->create(thread_pool_executor)),
       output{source->output},
-      options{options},
+      overlay_cursor{(options & wayland::ImageCopyCaptureManagerV1::Options::paint_cursors) != 0},
       ctx{ctx}
 {
     if (output)
@@ -342,18 +342,25 @@ auto mf::ExtImageCopyCaptureSessionV1::output_config_changed(graphics::DisplayCo
 void mf::ExtImageCopyCaptureSessionV1::apply_damage(
     std::optional<geom::Rectangle> const& damage)
 {
+    if (!output) {
+        return;
+    }
+
     if (damage && damage_amount != DamageAmount::full)
     {
-        if (damage.value().size != geom::Size{})
+        auto const output_space_area = output.value().current_config().extents();
+
+        auto const intersection = intersection_of(damage.value(), output_space_area);
+        if (intersection.size != geom::Size{})
         {
             if (damage_amount == DamageAmount::partial)
             {
-                output_space_damage = geom::Rectangles{output_space_damage, damage.value()}.bounding_rectangle();
+                output_space_damage = geom::Rectangles{output_space_damage, intersection}.bounding_rectangle();
             }
             else // damage_amount == DamageAmount::none
             {
                 damage_amount = DamageAmount::partial;
-                output_space_damage = damage.value();
+                output_space_damage = intersection;
             }
         }
     }
@@ -407,7 +414,6 @@ void mf::ExtImageCopyCaptureSessionV1::maybe_capture_frame()
     auto const buffer_size = output_config.modes[output_config.current_mode_index].size;
     auto const output_space_area = output_config.extents();
     auto const transform = output_config.transformation();
-    bool const overlay_cursor = (options & wayland::ImageCopyCaptureManagerV1::Options::paint_cursors) != 0;
 
     switch (damage_amount)
     {
@@ -463,7 +469,7 @@ mf::ExtImageCopyCaptureFrameV1::ExtImageCopyCaptureFrameV1(
 }
 
 void mf::ExtImageCopyCaptureFrameV1::attach_buffer(
-    [[maybe_unused]] wl_resource* buffer)
+    wl_resource* buffer)
 {
     if (capture_has_been_called)
     {


src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,607 @@
...
+mf::ExtImageCopyCaptureSessionV1::~ExtImageCopyCaptureSessionV1()
+{
+    if (change_notifier)
+    {
+        ctx->surface_stack->remove_observer(change_notifier);
+    }
+}
+
+
+void mf::ExtImageCopyCaptureSessionV1::send_buffer_format()
```
RAOF: This should perhaps be `send_buffer_constraints()`, since it's sending *all* the constraints, not just the format?

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,607 @@
...
+    {
+        // If damage amount is already full, or if damage is nullopt
+        // (and thus full damage should be applied)
+        damage_amount = DamageAmount::full;
+    }
+
+    if (damage_amount != DamageAmount::none)
+    {
+        maybe_capture_frame();
+    }
```
RAOF, line 377: Hm. I think this is the wrong place to do this, and the check is checking the wrong thing. If a multiple-capture client reuses the same buffer each time, it will *correctly* not send any damage (as the buffer has exactly the content from the last successful `::capture`).

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,607 @@
...
+    geom::Size buffer_size,
+    geom::Rectangle buffer_space_damage)
+{
+    if (!session)
+    {
+        send_failed_event(FailureReason::stopped);
+        return;
+    }
+
+    auto shm_buffer = dynamic_cast<ShmBuffer*>(wayland::as_nullable_ptr(target));
```
RAOF, line 558:
```suggestion
    auto shm_buffer = ShmBuffer::from(wayland::as_nullable_ptr(target));
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,607 @@
...
+            // Send release event for underlying buffer if necessary
+            executor->spawn(
+                [shm_data, weak_buffer]()
+                {
+                    if (weak_buffer)
+                    {
+                        weak_buffer.value().send_release_event();
+                    }
+                });
+        }
```
RAOF: We actually don't need to (and shouldn't²) send a `wl_buffer.release` event here; that is only for `wl_surface.attach`¹. The client knows when the buffer is ready for reuse from the `ready` event we send.

¹: And it's actually regarded as somewhat of a mistake there, too 🤷‍♀️ 
²: The [protocol](https://wayland.app/protocols/ext-image-copy-capture-v1#ext_image_copy_capture_frame_v1:request:attach_buffer) here says “the `wl_buffer.release` request (sic) is unused”

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+    session.value().screen_shooter->capture(
+        std::move(buffer), output_space_area, transform, overlay_cursor,
+        [executor=ctx->wayland_executor, buffer_space_damage, self=wayland::make_weak(this)](std::optional<time::Timestamp> captured_time)
+        {
+            executor->spawn([self, captured_time, buffer_space_damage]()
+                {
+                    if (self)
+                    {
+                        self.value().report_result(captured_time, buffer_space_damage);
```
RAOF, line 583: This is going to send the wrong damage.

There are two *distinct* sources of damage in this protocol -
1. The [`ext_image_copy_capture_frame_v1::damage`](https://wayland.app/protocols/ext-image-copy-capture-v1#ext_image_copy_capture_frame_v1:event:damage) event that we, the compositor, send. This describes where the scene has changed since the last `ready` event, and is useful for clients to optimise, as they know everything outside the reported damage is the same as last frame, and
2. The [`ext_image_copy_capture_frame_v1::damage_buffer`](https://wayland.app/protocols/ext-image-copy-capture-v1#ext_image_copy_capture_frame_v1:request:damage_buffer) request that the client sends to us. *This* describes where the currently-submitted buffer differs from the *last captured buffer*, and is useful for the compositor to optimise rendering, as we don't need to render anything outside that region + whatever changed *this* frame.

Both of these are solely for optimisation purposes - for a first implementation it's fine for us to always send full-buffer damage in the `ext_image_copy_capture_frame_v1::damage` event, and to ignore what the client says in `ext_image_copy_capture_frame_v1::damage_buffer` and just render the whole output.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,607 @@
...
+void mf::ExtImageCopyCaptureFrameV1::capture_frame(
+    geom::Rectangle output_space_area,
+    glm::mat2 transform,
+    bool overlay_cursor,
+    geom::Size buffer_size,
+    geom::Rectangle buffer_space_damage)
+{
+    if (!session)
+    {
+        send_failed_event(FailureReason::stopped);
```
RAOF, line 554: We *might* need to revisit this later; I don't think the protocol actually prohibits destroying the session before the frame has been captured (although that would be odd).

RAOF (commented):
I think there's a misunderstanding of damage handling here, but apart from that it looks broadly correct.

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: I've been looking over the wlroots implementation, and it seems to keep track of two damage regions.

The add_damage method from the client updates `frame->buffer_damage`:

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/406aa5f7f5649a9774fd89880037be3a731e96fa/types/wlr_ext_image_copy_capture_v1.c#L235-L259

While the capture method checks `frame->session->damage` to check whether to delay the capture:

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/406aa5f7f5649a9774fd89880037be3a731e96fa/types/wlr_ext_image_copy_capture_v1.c#L284-L288

When it comes time to copy data into the buffer, the session damage is merged into the client-provided damage:

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/406aa5f7f5649a9774fd89880037be3a731e96fa/types/wlr_ext_image_copy_capture_v1.c#L428-L432

It also looks like it only reports damage from `frame->session->damage` back to the client, rather than repeating back the client-requested damage:

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/406aa5f7f5649a9774fd89880037be3a731e96fa/types/wlr_ext_image_copy_capture_v1.c#L93-L99

From the look of it, that should let wayvnc properly meter it's capture rate.

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: Here, as_nullable_ptr will return a `wayland::Buffer`, while `ShmBuffer::from` wants a `wl_resource`. I could do `wayland::as_nullable_ptr(target)->resource`, but that would need to guard against a null pointer.

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: It looks like the wlroots implementation actively fails the current frame when the session is destroyed:

https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/406aa5f7f5649a9774fd89880037be3a731e96fa/types/wlr_ext_image_copy_capture_v1.c#L383-L391

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: I've filed [wayland-protocols #282](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/282) about clarifying that `damage_buffer` damage should not be treated the same as damage originating from the server.

My original reading of the spec makes it impossible for the server to meter the capture rate when it is called.

Saviq:
@jhenstridge I think we need to prioritize implementing the cursor session (in a follow up), that's why `wayvnc` needs `--disable-input`:
```
wl_display#1: error 3: Mir internal error processing ImageCopyCaptureManagerV1::create_pointer_cursor_session() request
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,605 @@
...
+void mf::ExtImageCopyCaptureFrameV1::report_result(
+    std::optional<time::Timestamp> captured_time,
+    geom::Rectangle buffer_space_damage)
+{
+    if (!captured_time)
+    {
+        send_failed_event(FailureReason::unknown);
+        return;
+    }
+    // send_transform_event();
```
robert-ancell: Dead code or need comment on why this is currently disabled?

src/wayland/CMakeLists.txt
```diff
@@ -41,6 +41,11 @@ mir_generate_protocol_wrapper(mirwayland "wp_" viewporter.xml)
 mir_generate_protocol_wrapper(mirwayland "wp_" fractional-scale-v1.xml)
 mir_generate_protocol_wrapper(mirwayland "z" xdg-activation-v1.xml)
 mir_generate_protocol_wrapper(mirwayland "wp_" linux-drm-syncobj-v1.xml)
+# FIXME: conflicts with wlr-foreign-toplevel-management-unstable-v1 if
+# we remove the ext_ prefix.
+mir_generate_protocol_wrapper(mirwayland "" ext-foreign-toplevel-list-v1.xml)
```
AlanGriffiths, line 47: Looks like comment and code disagree: trust neither

AlanGriffiths:
Also be aware of 

```
/// Extensions that are not in the set returned by mf::get_standard_extensions() should generally be listed in
/// include/miral/miral/wayland_extensions.h for easy access by shells.
```

src/wayland/CMakeLists.txt
jhenstridge: What I was trying to convey is that if I set the second argument to "ext_" here (which causes that prefix to be removed from symbols), we get a conflict with the other extension.

We have `zwlr_foreign_toplevel_handle_v1` and `ext_foreign_toplevel_handle_v1` interfaces, so we can't remove the prefixes on both. I guess this should just be a regular comment rather than a fixme.

AlanGriffiths: I see what you meant. Yes, the phrasing is unfortunate.

Maybe
```
# Suppressing the "ext_" prefix would conflict with wlr-foreign-toplevel-management-unstable-v1
```

include/miral/miral/wayland_extensions.h
```diff
@@ -155,6 +155,19 @@ class WaylandExtensions
     /// trusted to view all displayed content, including windows of other apps.
     static char const* const zwlr_screencopy_manager_v1;
 
+    /// Allows clients to take screenshots or record the screen. Only
+    /// enable for clients that are trusted to view all displayed
+    /// content, including windows of other apps.
+    /// This extension relies on other extensions like
+    /// ext_output_image_capture_source_manager_v1 to create capture
+    /// sources.
+    static char const* const ext_image_copy_capture_manager_v1;
```
AlanGriffiths, line 165:
```suggestion
    /// sources.
    /// \remark Since MirAL 5.6
    static char const* const ext_image_copy_capture_manager_v1;
```

include/miral/miral/wayland_extensions.h
```diff
@@ -155,6 +155,19 @@ class WaylandExtensions
...
+
+    /// Allows clients to create an image capture source representing
+    /// an output. This extension is only useful in conjunction with
+    /// ext_image_copy_capture_manager_v1.
+    static char const* const ext_output_image_capture_source_manager_v1;
```
AlanGriffiths, line 171:
```suggestion
    /// ext_image_copy_capture_manager_v1.
    /// \remark Since MirAL 5.6
    static char const* const ext_output_image_capture_source_manager_v1;
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,604 @@
...
+    auto const source_instance = ExtImageCaptureSourceV1::from_or_throw(source);
+    new ExtImageCopyCaptureSessionV1{new_resource, source_instance, options, ctx};
+}
+
+void mf::ExtImageCopyCaptureManagerV1::create_pointer_cursor_session(
+    [[maybe_unused]] wl_resource* new_resource,
+    [[maybe_unused]] wl_resource* source,
+    [[maybe_unused]] wl_resource* pointer)
+{
+    BOOST_THROW_EXCEPTION(std::runtime_error("not implemented"));
```
AlanGriffiths, line 298: Not sure what the right thing to do here is. But `wayland::ProtocolError()` seems more likely to be a graceful fail. Or a `fatal_error()`

AlanGriffiths:
Also

1. https://github.com/canonical/mir/pull/4326#discussion_r2431655434
2. https://github.com/canonical/mir/pull/4326#discussion_r2397401769 (I don't follow the whole discussion, but seems unresolved)

src/server/frontend_wayland/ext_image_capture_v1.cpp
RAOF: This will be translated into an [`implementation`](https://wayland.app/protocols/wayland#wl_display:enum:error:entry:implementation) protocol error by our Wayland machinery; this is a reasonable approach.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,604 @@
...
+    : wayland::ImageCopyCaptureSessionV1{resource, Version<1>()},
+      screen_shooter(ctx->screen_shooter_factory->create(thread_pool_executor)),
+      output{source->output},
+      options{options},
+      ctx{ctx}
+{
+    if (output)
+    {
+	output.value().add_listener(this);
+	output_config_changed(output.value().current_config());
```
RAOF: Ooops! You need to fix your editor settings; this is using tabs (which is why it appears incorrectly indented)

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,604 @@
...
+    ExtImageCopyCaptureSessionV1* session,
+    std::shared_ptr<ExtImageCaptureV1Ctx> const& ctx)
+    : wayland::ImageCopyCaptureFrameV1{resource, Version<1>()},
+      session{session},
+      ctx{ctx}
+{
+}
+
+void mf::ExtImageCopyCaptureFrameV1::attach_buffer(
+    [[maybe_unused]] wl_resource* buffer)
```
RAOF:
```suggestion
    wl_resource* buffer)
```
We can drop the `maybe_unused` now that this is implemented :)

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,604 @@
...
+
+mf::ExtImageCopyCaptureSessionV1::ExtImageCopyCaptureSessionV1(
+    wl_resource *resource,
+    ExtImageCaptureSourceV1* source,
+    uint32_t options,
+    std::shared_ptr<ExtImageCaptureV1Ctx> const& ctx)
+    : wayland::ImageCopyCaptureSessionV1{resource, Version<1>()},
+      screen_shooter(ctx->screen_shooter_factory->create(thread_pool_executor)),
+      output{source->output},
+      options{options},
```
RAOF: We should parse `options` here into something meaningful (specifically, `paint_cursor`) at the interface between Wayland and our code.

src/server/frontend_wayland/ext_image_capture_v1.cpp
RAOF: This now does the right thing. I think the implicit state machine here could do with an explanatory comment, but it's waiting until both the client has called `.capture()` *and* the scene has updated since the last capture.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,604 @@
...
+        {
+            if (damage_amount == DamageAmount::partial)
+            {
+                output_space_damage = geom::Rectangles{output_space_damage, damage.value()}.bounding_rectangle();
+            }
+            else // damage_amount == DamageAmount::none
+            {
+                damage_amount = DamageAmount::partial;
+                output_space_damage = damage.value();
+            }
```
RAOF: Hm. We should clip the damage to the capture region - on multi monitor setups we don't care about anything changing on an output that we're *not* capturing, and we might be sending out-of-bounds damage to the client in that case?

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: None of the `ext_image_copy_capture_manager_v1::error` values really match up to a missing implementation. So I think this is probably okay as is for now. It'll be properly resolved in a followup PR.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,610 @@
...
+    rect.size.width = rect.size.width * x_scale;
+    rect.size.height = rect.size.height * y_scale;
+    auto displacement = rect.top_left - input_space.top_left;
+    rect.top_left.x = output_space.top_left.x + displacement.dx * x_scale;
+    rect.top_left.y = output_space.top_left.y + displacement.dy * y_scale;
+    return rect;
+}
+}
+
+namespace mir::frontend {
```
AlanGriffiths, line 58:
```suggestion
namespace mir::frontend
{
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,610 @@
...
+public:
+    ExtImageCaptureSourceV1(wl_resource* resource,
+                            OutputGlobal& output,
+                            ExtOutputImageCaptureSourceManagerV1* manager);
+
+    static ExtImageCaptureSourceV1* from_or_throw(wl_resource* resource);
+
+    // FIXME: currently this is only set up for output capture. It
+    // will need to change to also cover toplevel capture.
+    wayland::Weak<OutputGlobal> const output;
```
AlanGriffiths, line 94: I know this is only in the .cpp file, but mixing public and private data members in the same class is an odd design choice. Why?

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -0,0 +1,610 @@
...
+mf::ExtImageCaptureSourceV1* mf::ExtImageCaptureSourceV1::from_or_throw(wl_resource *resource)
+{
+    auto source = dynamic_cast<ExtImageCaptureSourceV1*>(wayland::ImageCaptureSourceV1::from(resource));
+
+    if (!source)
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "ext_image_capture_source_v1@" +
+            std::to_string(wl_resource_get_id(resource)) +
+            " is not a mir::frontend::ExtImageCaptureSourceV1"));
```
RAOF, line 247: Non-blocking nit, but `std::format()` probably does this better:
```suggestion
        BOOST_THROW_EXCEPTION(std::runtime_error(
            std::format(
                "ext_image_capture_source_v1@{} is not a mir::frontend::ExtImageCaptureSourceV1",
                wl_resource_get_id(resource)));
```

src/wayland/CMakeLists.txt
```diff
@@ -42,6 +42,11 @@ mir_generate_protocol_wrapper(mirwayland "wp_" fractional-scale-v1.xml)
 mir_generate_protocol_wrapper(mirwayland "z" xdg-activation-v1.xml)
 mir_generate_protocol_wrapper(mirwayland "wp_" linux-drm-syncobj-v1.xml)
 mir_generate_protocol_wrapper(mirwayland "ext_" ext-data-control-v1.xml)
+# We can't strip the "ext_" prefix here, because it creates a conflict
+# with wlr-foreign-toplevel-management-unstable-v1.
```
AlanGriffiths, line 46:
```suggestion
# We don't strip the "ext_" prefix here, because it would conflict
# with wlr-foreign-toplevel-management-unstable-v1.
```

PR #4328, opened 2025-10-03
MIRENG-1493-input-triggers

AlanGriffiths:


wayland-protocols/ext-input-trigger-activation-v1.xml
```diff
@@ -0,0 +1,92 @@
...
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+    THE SOFTWARE.
+  </copyright>
+
+  <description summary="Registration for global input actions">
+    This protocol allows for a privileged Wayland client to for notification of
+    input trigger actions.
```
Saviq: This?
```suggestion
    This protocol allows for a privileged Wayland client to subscribe to input
    trigger action notifications.
```

wayland-protocols/ext-input-trigger-activation-v1.xml
```diff
@@ -0,0 +1,92 @@
...
+  <interface name="ext_input_trigger_activation_v1" version="1">
+    <description summary="receives notifications of input trigger actions">
+      The compositor sends notification of the activation and deactivation of
+      the associated action.
+    </description>
+
+    <event name="activate">
+      <description summary="the input action has been activated">
+      </description>
+      <arg name="activation_token" type="string" summary="and xdg_activation token"/>
```
Saviq:
```suggestion
      <arg name="activation_token" type="string" summary="a xdg_activation token"/>
```

wayland-protocols/ext-input-trigger-activation-v1.xml
```diff
@@ -0,0 +1,92 @@
...
+    <event name="deactivate">
+      <description summary="the input action has been deactivated">
+      </description>
+    </event>
+
+    <event name="unavailable">
+      <description summary="the input action is (or has become) unavailable">
+        Sent if the token is or becomes unavailable.
+
+        This the object is of no further use and should be destroyed.
```
Saviq:
```suggestion
        This means the object is of no further use and should be destroyed.
```

wayland-protocols/ext-input-trigger-activation-v1.xml
```diff
@@ -0,0 +1,92 @@
...
+
+    <event name="activate">
+      <description summary="the input action has been activated">
+      </description>
+      <arg name="activation_token" type="string" summary="and xdg_activation token"/>
+    </event>
+
+    <event name="deactivate">
+      <description summary="the input action has been deactivated">
+      </description>
```
Saviq: What does it mean for an action to be deactivated?

I find it confusing that "activate" is not contrary to "deactivate".

wayland-protocols/ext-input-trigger-activation-v1.xml
```diff
@@ -0,0 +1,92 @@
...
+    </request>
+  </interface>
+
+  <interface name="ext_input_trigger_activation_v1" version="1">
+    <description summary="receives notifications of input trigger actions">
+      The compositor sends notification of the activation and deactivation of
+      the associated action.
+    </description>
+
+    <event name="activate">
```
Saviq: Should this be `trigger`?

wayland-protocols/ext-input-trigger-activation-v1.xml
```diff
@@ -0,0 +1,92 @@
...
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+    THE SOFTWARE.
+  </copyright>
+
+  <description summary="Registration for global input actions">
+    This protocol allows for a privileged Wayland client to for notification of
```
Saviq: Not a privileged client, though? Just one that holds the token?

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,156 @@
...
+        This informs the compositor that the input trigger registration manager
+        object will no longer be used. Existing objects created through this
+        interface remain valid.
+      </description>
+    </request>
+
+    <request name="register_input_trigger_event">
+      <description summary="register input trigger event">
+        This request creates an input trigger asks the compositor to reserve it
+        to the manager.
```
Saviq: This?
```suggestion
        This request creates an input trigger and asks the compositor to reserve it
        to the manager.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,156 @@
...
+      <description summary="register input trigger action event">
+        This request creates an input trigger action.
+
+        The name parameter names the action (it has no semantic content and
+        may be duplicated across actions).
+
+        The compositor will send the ext_input_trigger_action_v1.done event in
+        response to this request.
+      </description>
+      <arg name="name" type="string"/>
```
Saviq: This is where we need a description, no?

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,156 @@
...
+    <event name="done">
+      <description summary="input trigger successfully registered">
+        This client is now responsible for handling this input trigger.
+      </description>
+    </event>
+
+    <event name="failed">
+      <description summary="input trigger failed to register">
+        The failed event indicates that the trigger is not "owned" by the manager.
+        This the object is of no further use and should be destroyed.
```
Saviq:
```suggestion
        This means the object is of no further use and should be destroyed.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,156 @@
...
+      </description>
+    </event>
+  </interface>
+
+  <interface name="ext_input_trigger_action_v1" version="1">
+    <description summary="represents the registration of an input trigger action">
+      On binding this interface the compositor will immediately an
```
Saviq:
```suggestion
      On binding this interface the compositor will immediately send an
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,156 @@
...
+
+        This informs the compositor that the input trigger action will no
+        longer be used.
+      </description>
+    </request>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the input trigger action">
+        This informs the compositor that the input trigger action will no
+        longer be used.
```
Saviq, line 235: But other clients might still use it? Is that the difference between `cancel` and `destroy`?

Saviq (changes requested):
I see this is still a draft, but just leaving some notes on it.

wayland-protocols/ext-input-trigger-activation-v1.xml
AlanGriffiths: No. Why?

wayland-protocols/ext-input-trigger-activation-v1.xml
tarek-y-ismail: In the context of a keyboard action, `activate` would be the pressing down of all buttons specified, and `deactivate` would be the release of _any_ of them. I don't see this carrying 1:1 to touchpad gestures or other input methods though.

I guess oneshot events could be emulated by a single `activate` event. What about continuous events? Would it be `activate`, `update(x,y)`, `deactivate` for example? This would work for touchpad gestures, but more novel input hardware might not fit neatly into that.

wayland-protocols/ext-input-trigger-registration-v1.xml
tarek-y-ismail: +1 on this.

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,186 @@
...
+      </description>
+      <arg name="trigger" type="object" interface="ext_input_trigger_event_v1" summary="the trigger being added"/>
+    </request>
+
+    <request name="drop_input_trigger_event">
+      <description summary="drop an input trigger from the action">
+        This request drops an input trigger from those that activate the action.
+      </description>
+      <arg name="trigger" type="object" interface="ext_input_trigger_event_v1" summary="the trigger being dropped"/>
+    </request>
```
tarek-y-ismail, line 206: Just to confirm. To change the input trigger event on an action, we'd drop the existing one and add another one, right? 

How would we notify applications that the underlying input trigger event changed?

The main use case for this is for apps to display the shortcuts they have displayed. Maybe add an `action_input_trigger_event_changed` event which would return a list of strings (assuming all input trigger events can be represented by strings)?

tarek-y-ismail (commented):
f

wayland-protocols/ext-input-trigger-activation-v1.xml
Saviq: For events, should we use past tense? `activated`, `deactivated`, `triggered`, `released`?

I'm seeing this more as "trigger was …", not "… action".

wayland-protocols/ext-input-trigger-activation-v1.xml
Saviq: Discussion below.

wayland-protocols/ext-input-trigger-registration-v1.xml
Saviq: Shouldn't this also happen at bind time? When a client receives a token from the intermediary, then binds to an action, we would send a description of the configured triggers?

---

Or maybe we rather want to send a "link" to open the shortcut configuration app, so the client only displays a "Manage shortcuts" button?

> assuming all input trigger events can be represented by strings

Unless we build a whole language to describe them… using a string gives compositor freedom in what triggers they implement - the protocol does not need to be able to interpret them. Some will be broadly implemented (like `[<Modifier>+ ...]<Key>`), others less so (say `3FingerDrag`, `3FingerPinch`). If we go with the "manage here" approach, those will just be hints to the compositor, apps need not interpret them, either.

AlanGriffiths: The mapping between triggers and actions is controlled by the gatekeeper. The gatekeeper informs the compositor and app. There is no need for the compositor to also inform the app.

wayland-protocols/ext-input-trigger-activation-v1.xml
AlanGriffiths: Maybe `start()`/`stop()` or `begin()`/`end()` would be a better pairing?

tarek-y-ismail: After discussing the update event with @Saviq, I think just sending `update(float progress)` where `progress` is in the range `[0,1]` should do?

wayland-protocols/ext-input-trigger-registration-v1.xml
Saviq: In that case, we'll need to define the language the gatekeeper uses to define triggers after all, because the lease is per trigger, not per action.

AlanGriffiths: Let's keep _this_ discussion focused. This context is adding and removing triggers from an action. And whether the compositor needs to send notifications to the app when this happens.

Discussion about identifying triggers to the compositor belongs with the `register_input_trigger_keyboard_event` request and the RFC above it.

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,186 @@
...
+          deprecated (and were deprecated before it was written).
+
+           And "LOGO"? I think "Meta" is more widely known.
+        -->
+        The keycombo must follow the XDG Shortcut standard.
+
+        The compositor will send either the ext_input_trigger_event_v1.done or
+        ext_input_trigger_event_v1.failed event on the created object in
+        response to this request.
+      </description>
```
AlanGriffiths: I've realized this is simplistic. Take `Alt+Tab` as an example...

The action starts when `Alt` is held and `Tab` is pressed
The action, typically, ends when `Alt` is released

Essentially, we need to define actions in more detail than "ALT+Tab" - especially any end condition

tarek-y-ismail: I think we discussed this before in the context of an app switcher. Once the app switcher gets the `activated` event, it can focus itself and then receive later input events. Is there any new use case that might benefit from a more complex model?

AlanGriffiths: > Is there any new use case that might benefit from a more complex model?

Actually, I came to this from touchpad gestures - e.g. three finger drag for switching workspaces. That's a case where we really need to "capture" input in some way.

But with Alt+Tab task switching there is an unsettling gap between the compositor sending the `begin()` and responding to the `xdg_activation_v1.activate()`. But, maybe, I'm trying to be too precise here - human interaction can be a bit messy: my thinking leads into the weeds if the app doesn't react.

Another iteration coming up!

Saviq: > That's a case where we really need to "capture" input in some way.

And I believe this should be a simple update on gesture progress in the `[0.0, 1.0]` range with a timestamp. The compositor determines what's the `[0.0, 1.0]` distance for the gesture, but the client determines what constitutes a complete gesture in that range, given the time difference between `begin()` and `end()`.

AlanGriffiths: I don’t see how a scalar helps. A drag has direction, quantity and speed

Saviq: Direction should be part of the gesture itself, a north-bound drag should be separate from an east-bound one. And speed can be determined from the quantity and time.

AlanGriffiths: I think this discussion highlights that we don't (yet) have a domain language for input triggers beyond what the GlobalShortcuts covers.

I've commented out the begin/end events pending further discussion. (Which is probably best addressed elsewhere - will put some notes into the whitepaper and see if I can get to the BoB standup tomorrow)

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -0,0 +1,117 @@
...
+
+        The app may begin the activity it associates with the action.
+      </description>
+      <arg name="time" type="uint" summary="timestamp with millisecond granularity"/>
+      <arg name="activation_token" type="string" summary="an xdg_activation token"/>
+    </event>
+
+    <event name="update">
+      <description summary="progress an the input action">
+        The input trigger that led to the "begin" event has clanged.
```
tarek-y-ismail:
```suggestion
        The input trigger that led to the "begin" event has changed.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,214 @@
...
+  </interface>
+
+  <interface name="ext_input_trigger_v1" version="1">
+    <description summary="represents the (attempted) registration of an input trigger">
+      In response to the creation of this object the compositor must send
+      either the done or failed event.
+
+      The done event indicates that the trigger is "owned" by the manager.
+      This means the compositor will not allow this, or any other manager to
+      register the trigger event.
```
tarek-y-ismail, line 161: Maybe this is better?
```suggestion
      The done event indicates that the trigger is "owned" by the manager.
      This means the compositor will not allow any other manager to
      register the trigger event, nor will it allow duplicate registrations from 
      this manager
```

wayland-protocols/ext-input-trigger-registration-v1.xml
tarek-y-ismail, line 1: So here's a quick overview of how I understand the general flow:
2. event: `ext_input_trigger_registration_manager_v1.capabilities(...)` 
3. request: `action_control = ext_input_trigger_registration_manager_v1.get_action_control("Do special CTRL+ALT+c thing")`
4. request: `ctrl_alt_c = ext_input_trigger_registration_manager_v1.register_keyboard_trigger("ctrl+alt+c")`
5. event `ctrl_alt_c.done()`
6. request: `action_control.add_input_trigger_event(ctrl_alt_c)`
7. event: `token = action_control.done("foo_42")`

registration done

---

listening phase

8.  request: `trigger_action = ext_input_trigger_action_manager_v1.get_input_trigger_action("foo_42")`
9. User presses "ctrl + alt + c"
10. event: `trigger_action.begin(...)`
11. User lets go of "ctrl + alt + c"
12. event: `trigger_action.end(...)`

AlanGriffiths: Mostly

> event: ext_input_trigger_registration_manager_v1.bind()

A `wl_registry_bind` is not an event - it is an API call on the client side.

> request: ctrl_alt_c = ext_input_trigger_registration_manager_v1.register_keyboard_trigger("ctrl+alt+c")

That string doesn't strictly match the referenced freedesktop shortcuts-spec. "CTRL+ALT+c"

> request: action_control.add_input_trigger_event(ctrl_alt_c)

This is not ordered w.r.t. the `done()` event - so it may be clearer to put all the trigger related calls after all the event related ones.

> event: token = action_control.done("foo_42")

The parameter to the `done()` event is the token - there is no return type to the callback.

tarek-y-ismail: > That string doesn't strictly match the referenced freedesktop shortcuts-spec. "CTRL+ALT+c"

That's okay, I'm just making sure I understand the bigger picture.

> This is not ordered w.r.t. the done() event - so it may be clearer to put all the trigger related calls after all the event related ones.

Updated to do so.

> The parameter to the done() event is the token - there is no return type to the callback.

Yup, I'm just making obvious to myself what new data we get out of each event vs parameters that might be used but not held onto

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -0,0 +1,117 @@
...
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+    THE SOFTWARE.
+  </copyright>
+
+  <description summary="Registration for global input actions">
+    This protocol allows for a Wayland clients to subscribe to input trigger
+    actions for activate and deactivate notifications.
```
tarek-y-ismail:
```suggestion
    actions for "begin", "update,  and "end" notifications.
```

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -0,0 +1,117 @@
...
+      <description summary="destroy input trigger activation manager object">
+        This informs the compositor that the input trigger activation manager
+        object will no longer be used. Existing objects created through this
+        interface remain valid.
+      </description>
+    </request>
+
+    <request name="get_input_trigger_action">
+      <description summary="register interest in input trigger action">
+        This request creates an input trigger activation object.
```
tarek-y-ismail: action

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,214 @@
...
+      </description>
+      <entry name="keyboard" value="1" summary="key combinations can be used as triggers"/>
+      <entry name="pointer" value="2" summary="pointer gestures can be used as triggers"/>
+      <entry name="touchpad" value="4" summary="touchpad gestures can be used as triggers"/>
+    </enum>
+
+    <event name="capabilities">
+      <description summary="input trigger capabilities">
+        This is emitted when ext_input_trigger_registration_manager_v1 is bound
+        and whenever a system gains or loses the capabilities.  The argument is
```
tarek-y-ismail:
```suggestion
        and whenever a system gains or loses the capabilities. The argument is
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -113,6 +113,38 @@
...
+        ext_input_trigger_v1.failed event on the created object in
+        response to this request.
+
+        Input actions triggered by this request will result in an
+        ext_input_trigger_action_v1.begin event when the mouse is pushed to the
+        target edge(s) and ext_input_trigger_action_v1.end when mouse moves away.
+      </description>
+      <arg name="edges" type="uint" enum="screen_edges"/>
+      <arg name="id" type="new_id" interface="ext_input_trigger_v1"/>
+    </request>
```
tarek-y-ismail: Slightly out there, but one cool thing we could add if we decide to keep this is to have `begin()` be be called when we "get in range" of an edge, then `update()` as we get closer, then `end()` as we move away past "the range" of the gesture. The only "cool" use for this that I have right now is to reduce music volume / mute as the mouse moves towards an edge, or change the track being played if the mouse is flicked to the left or right edges (admittedly not very ergonomic on larger/wide screens)

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -113,6 +113,38 @@
...
+      <entry name="bottom" value="2" summary="bottom edge"/>
+      <entry name="left" value="4" summary="left edge"/>
+      <entry name="top_left" value="5" summary="top left corner"/>
+      <entry name="bottom_left" value="6" summary="bottom left corner"/>
+      <entry name="right" value="8" summary="right edge"/>
+      <entry name="top_right" value="9" summary="top right corner"/>
+      <entry name="bottom_right" value="10" summary="bottom right corner"/>
+    </enum>
+
+    <request name="register_mouse_trigger">
```
tarek-y-ismail: One thing I think some users might expect is using their extra mouse buttons for triggers. Does this cover it?

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -43,8 +43,9 @@
         set, then triggers can be set.
       </description>
       <entry name="keyboard" value="1" summary="key combinations can be used as triggers"/>
-      <entry name="pointer" value="2" summary="pointer gestures can be used as triggers"/>
-      <entry name="touchpad" value="4" summary="touchpad gestures can be used as triggers"/>
+      <entry name="modifiers" value="2" summary="keyboard modifier taps and/or holds can be used as triggers"/>
+      <entry name="pointer" value="4" summary="pointer gestures can be used as triggers"/>
+      <entry name="touchpad" value="8" summary="touchpad gestures can be used as triggers"/>
```
tarek-y-ismail: Should  wee add this when we're yet to add `register_touchpad_trigger`?

wayland-protocols/ext-input-trigger-registration-v1.xml
AlanGriffiths: It doesn't - I didn't think of that case. Is it a simple button press/release? Or button + movement/scroll?

I think that would be a different request - it doesn't need the edge argument and does need a button(s) argument.

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,352 @@
...
+        This request creates an input trigger and asks the compositor to reserve
+        it to the manager.
+
+        The compositor will send either the ext_input_trigger_v1.done or
+        ext_input_trigger_v1.failed event on the created object in
+        response to this request.
+
+        Input actions triggered by this request will result in an
+        ext_input_trigger_action_v1.begin event when the pointer is pushed to the
+        target edge(s) and ext_input_trigger_action_v1.end when pointer moves away.
```
AlanGriffiths: There are other design choices...

@tarek-y-ismail suggested we could have a "margin" to the edge, with `begin` on entering the margin and `update` indicating progress across the margin.

With such a trigger, it is more consistent if the compositor decides the margin for all triggers (specifying independently for each trigger won't give a consistent user experience).

wayland-protocols/ext-input-trigger-registration-v1.xml
tarek-y-ismail: I would assume just a button press, but I don't use this feature with my mouse so I think we might want to wait for other feedback on it. And I agree that it would need to be another request. Something like `register_pointer_button_trigger`

wayland-protocols/ext-input-trigger-registration-v1.xml
tarek-y-ismail: Yeah, I agree. Having each app specifying its own margin for the trigger would be messy..

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -0,0 +1,119 @@
...
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+    THE SOFTWARE.
+  </copyright>
+
+  <description summary="Registration for global input actions">
+    This protocol allows for a Wayland clients to subscribe to input trigger
+    actions for "begin", "update,  and "end" notifications.
```
tarek-y-ismail:
```suggestion
    actions for "begin", "update, and "end" notifications.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,352 @@
...
+        The various shift modifiers on the keyboard
+      </description>
+      <entry name="alt" value="0x01" summary="any Alt key"/>
+      <entry name="alt_left" value="0x02" summary="left Alt key"/>
+      <entry name="alt_right" value="0x04" summary="Right Alt key"/>
+      <entry name="shift" value="0x08" summary="any Shift key"/>
+      <entry name="shift_left" value="0x10" summary="left Shift key"/>
+      <entry name="shift_right" value="0x20" summary="right Shift key"/>
+      <entry name="sym" value="0x40" summary="Sym key"/>
+      <entry name="function" value="0x80" summary="Function key"/>
```
tarek-y-ismail, line 76: Not sure about those. What's the sym key? And can function keys be used as modifiers?

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,352 @@
...
+        modifier keys. As modifiers are used normally in combination with other
+        keys, the hold_delay and tap_count should not both be zero.
+
+        The compositor will send either the ext_input_trigger_v1.done or
+        ext_input_trigger_v1.failed event on the created object in
+        response to this request.
+
+        Input actions triggered by this request will result in an
+        ext_input_trigger_action_v1.begin event when triggered
+        target edge(s) and ext_input_trigger_action_v1.end when the trigger ends.
```
tarek-y-ismail:
```suggestion
        and ext_input_trigger_action_v1.end when the trigger ends.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,352 @@
...
+      </description>
+      <arg name="modifiers" type="uint" enum="modifiers"/>
+      <arg name="hold_delay" type="uint" summary="minimum milliseconds pressed before triggering"/>
+      <arg name="tap_count" type="uint" summary="number of consecutive taps before triggering"/>
```
tarek-y-ismail: How do these interact? For example, if a client registers a modifier action for `alt+shift` with a delay of 100ms and a tap count of 3. Does activation require holding both keys for 100ms, then three quick taps, or does it require a hold then 3 taps at least 100 ms apart? What about if the two modifiers aren't pressed at the exact same time? I suppose compositors could allow for small window of difference...

wayland-protocols/ext-input-trigger-registration-v1.xml
tarek-y-ismail: We should also mention that the size of the edge is defined by the compositor

wayland-protocols/ext-input-trigger-registration-v1.xml
AlanGriffiths: These are modifier keys that may exist. For example, I have a 'fn' (function) key between Ctrl and Meta of this keyboard that changes the meaning of the top row.

wayland-protocols/ext-input-trigger-registration-v1.xml
AlanGriffiths: Yes, the compositor has to allow some flexibility. It also needs to filter out sequences including unrelated keys.

Maybe this is overbroad: support for multi-modifier taps, or taps with a hold delay needs discussion. (I was thinking that it is obvious that the hold delay would be the duration of the "tap", and any of the modifiers could be tapped while the others are held)

wayland-protocols/ext-input-trigger-registration-v1.xml
AlanGriffiths: I don't think that needs saying. Similarly that the compositor decides what an edge is (a screen, the union of all active screens, the edge of an "applications" area...)

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -134,25 +134,42 @@
...
+
+        The compositor will send either the ext_input_trigger_v1.done or
+        ext_input_trigger_v1.failed event on the created object in
+        response to this request.
+
+        Input actions triggered by this request will result in an
+        ext_input_trigger_action_v1.begin event when triggered
+        and ext_input_trigger_action_v1.end when the trigger ends.
+      </description>
+      <arg name="modifier" type="uint" enum="modifier"/>
```
tarek-y-ismail: Wouldn't it be possible to replace these two requests by just having the keysym be an optional argument in `register_keyboard_sym_trigger` (or keycodes,  I don't think modifiers change with layouts but I might be wrong). It would then be the responsibility of the client to keep track of taps/holds, and begin/end events are enough for that.

AlanGriffiths: No.

1. The client cannot track taps or holds as it doesn't have the same context as the compositor (e.g knowing whether other input is happening).
2. An action begin event is semantically the beginning of some condition. Its meaning should not depend on the trigger: I may want an action triggered when either "tapping Shift five times" or "holding both Shift keys"
3. To be consistent across apps the compositor decides what is a "hold" or "3-taps", not the individual apps.

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -0,0 +1,119 @@
...
+        interface remain valid.
+      </description>
+    </request>
+
+    <request name="get_input_trigger_action">
+      <description summary="register interest in input trigger action">
+        This request creates an input trigger action object.
+
+        The token should be one issued by the compositor, if not then the
+        compositor can raise a protocol error.
```
RAOF: This should be more prescriptive: the compositor *should* raise a protocol error, but which protocol error? None appear to be defined.

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,369 @@
...
+        This describes a trigger based tapping one a modifier key. The tap_count
+        may not be zero.
+
+        The compositor will send either the ext_input_trigger_v1.done or
+        ext_input_trigger_v1.failed event on the created object in
+        response to this request.
+
+        Input actions triggered by this request will result in an
+        ext_input_trigger_action_v1.begin event when triggered
+        and ext_input_trigger_action_v1.end when the trigger ends.
```
RAOF: When does the trigger begin/end? This seems to be a oneshot event - the compositor doesn't know that the taps are a gesture until the gesture is complete.

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,369 @@
...
+      <arg name="id" type="new_id" interface="ext_input_trigger_v1"/>
+    </request>
+
+    <request name="register_modifier_tap_trigger">
+      <description summary="register input trigger">
+        This request creates an input trigger and asks the compositor to reserve
+        it to the manager.
+
+        This describes a trigger based tapping one a modifier key. The tap_count
+        may not be zero.
```
RAOF:
```suggestion
        This describes a trigger based on tapping a modifier key. The tap_count
        may not be zero.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
RAOF: Hm. Unless you have a combination tap/hold event? ie: register shift, tap_count 2 and then the user goes: tap shift, press and hold shift - would this result in the `begin` event being sent, with the `end` event sent when the user releases the shift key?

RAOF:
Oh, additionally - as discussed in person, I think we need to describe how this interacts with regular input dispatch. If a client has registered a `<ctrl>+f` trigger and it has input focus, is it *also* going to receive a `<ctrl>+f` keyboard event?

wayland-protocols/ext-input-trigger-registration-v1.xml
tarek-y-ismail: In this current iteration, tap and hold events are separate as far as I understand

wayland-protocols/ext-input-trigger-action-v1.xml
tarek-y-ismail: Below in the `unavailable` event:
> Sent if the token is or becomes unavailable. 

But in my opinion, it's valid here to raise a protocol error and terminate the client if they pass in an incorrect token. I don't think activation tokens for input triggers would be so short lived that this would cause a problem for well-behaving clients.

tarek-y-ismail:
>  when is a trigger dropped and so can be re-bound by a manager?

When
1. The user deletes the input trigger, which should call `remove_input_trigger_event`
2. When the client binding the trigger disconnects

> when does the exported action token become invalid?

A few seconds after the token is first issued. Which in the case of the action control, would probably call `cancel`

> I think we need to describe how this interacts with regular input dispatch

Huh, I thought we already settled on this? Anyway, I don't think the focused client, whether or not it has an input trigger registered, should get it as a normal input.

wayland-protocols/ext-input-trigger-action-v1.xml
RAOF: > But in my opinion, it's valid here to raise a protocol error and terminate the client if they pass in an incorrect token.

Yes, but *what* protocol error? You can't just send “error”, it needs to be a particular thing (but there are no errors defined in this protocol).

Saviq: @copilot can you add a protocol error fitting this use case

tarek-y-ismail:
So I used Copilot as a rubber duck, here's a quick "summary" of the points it brought up about the protocol and my opinions/responses. Apologies in advance for the wall of text.

--- 

> Concrete protocol bug (must fix)
> - enum name mismatch in `register_modifier_tap_trigger`
> - Fix: change `enum="modifier"` to `enum="modifiers"` and/or rename arg to `modifiers` if you intend a bitfield (and document if multiple modifiers are allowed).

It's worth discussing whether we want to allow just one modifier in this request, or multiple ones. The use case I recall from when we discussed this was tapping both the left and right shifts to toggle mute, which would require us to specify two modifiers.

---

### High priority semantic and security issues

> 1. Token/auth model is underspecified: token generation properties (entropy, unguessability), token format / length,

Tokens are mostly meant to be opaque strings similar to tokens used in ext_activation_v1.

> 2. No seat/device scoping (multi-seat/multi-device implications)

This is true. Do we want to take this into account for v1?

> 3. `register_pointer_trigger` takes `edges` but does not say whether edges are per-output, global screen-space, or per-monitor; it doesn’t accept an output id or bounding region.

I think we discussed this before. We settled on this being compositor-defined.

>  4. Lack of ownership/consumption semantics for captured input
>  - When a trigger is “done” and the manager “owns” it, the docs say “This client is now responsible for handling this input trigger.” That is vague:
>    - Does the compositor stop delivering the underlying raw input events to normal clients?
>    - Does the compositor synthesize events for the manager only (i.e., exclusive grab) or are events still delivered to applications?

We should clear this up. From memory, we decided that all inputs will be delivered to the focused client as usual. Once a registered trigger is invoked, that input will be consumed by the compositor and diverted to the listening client.

Implementation details: As @RAOF mentioned before, the modifier state will need to be reset by synthesizing events to clear the pressed modifiers so the focused client is not confused when focus returns to it.

Example, `Ctrl + Shift + X`. The focused client would get `Ctrl`, then `Shift`. If another app takes focus temporarily and the user lets go of the two modifiers before returning to the first app, then the app will still think the two modifiers are still down.

> 5. Concurrency and correlation of begin/update/end events
>    - `ext_input_trigger_action_v1` has `begin`, `update`, `end` events but no identifier to correlate multiple simultaneous activations for the same action token. The protocol doesn't forbid simultaneous activations; it doesn't state whether `begin`/`end` pairs are strictly nested or distinct, or whether multiple `begin`s can be outstanding.

Is this even possible? Unless you have multiple input devices (keyboards / mice), I don't think it is.

> 6. Time/timestamp semantics unclear
>    - Each event includes `time` (uint) "timestamp with millisecond granularity" but it doesn't state the epoch or clock reference (compositor monotonic clock? boot time?), wrap-around behavior, or whether the client should use it for ordering vs the Wayland event order.

I think we should specify this. For us, our clocks are usually relative to the UNIX epoch, right? I don't think we need to worry about wrap-around. And ordering of the events _should_ match the Wayland event order so that's also a non-issue.

> 7. `activation_token` semantics and lifecycle are ambiguous
>    - `activation_token` is a string described as “an xdg_activation token” but there’s no link to exact usage: should clients call `xdg_activation_v1.activate` with it? Is it optional/nullable? Who issues these tokens and when do they expire?

We should clear up the usage of the token. From the API, it's obviously not nullable/optional, so that's not a concern.  We should specify that tokens are issued by the compositor and expire after a certain (compositor-defined) period of time.

### Medium priority design omissions / missing error handling

> 8. Poor/few error codes and lack of failure reasons
>    - `ext_input_trigger_v1` has `done`/`failed` with no failure reason, and the action manager only has `invalid_token` as an error. There is no reason code taxonomy (e.g., conflict, unsupported capability, insufficient privileges, out of resources).

Maybe we want to explore this a bit. I believe we discussed the conflict case before and settled on leaving it up to the DE maintainers to handle conflicts between the various previliged clients externally. Though, this doesn't rule out the user modifying a trigger and accidentally using one that's already taken. Which in this case would just call `failed`. I'm not so sure about the other modes of failure.

> 9. Missing constraints, thresholds and configuration options for gestures
>    - Touch drag/tap triggers have `touches`, `direction`, `hold_delay`, but no tolerances:
>      - angle tolerance for "direction",
>      - minimum/maximum movement distance to classify as drag vs tap,
>      - inter-tap timeout for `tap_count`,
>      - maximum finger spread or bounding box, velocity thresholds.
>    - Result: different compositors may implement wildly different behavior; false positives/negatives are likely.

Again, I believe we discussed this before and agreed that it should be up to the compositor to define these to ensure a consistent experience.

> 10. Pointer trigger hysteresis and debouncing unspecified
>     - For screen edge triggers: what counts as “pushed to the target edge(s)”? If a pointer is at x=1 pixel from edge, does that count? Without hysteresis, you can get flapping between begin/end as the pointer hovers near the edge.

Compositor defined, again.

> 11. `register_keyboard_sym_trigger` vs `register_keyboard_code_trigger` implications & layout issues
>     - Using keysyms is layout-dependent; the semantics are currently “interpreted according to xkbcommon-keysyms.h” but there’s no guidance on how layout changes affect registered triggers, or whether triggers bind to physical key locations vs interpreted symbols (i.e., Ctrl+comma vs Ctrl+less).

I think Copilot got a bit confused, but this probably means we need to clarify things a bit. Code triggers should not change with the layout, but sym triggers should. Though the more I think about it, the more I see sym triggers as superfluous, what was their use case again?

> 12. Tap/hold semantics vagueness
>     - `register_modifier_hold_trigger` and `register_modifier_tap_trigger` are underspecified: what exactly is a 'tap' (press+release within time/without movement), maximum allowed inter-key times for multi-taps, etc.

Fair point. I think this should also be compositor-defined to ensure consistency.

> 13. Lack of clarity about lifecycle interactions between `cancel` vs `destroy`
>     - `ext_input_trigger_action_control_v1` has `cancel` (destructor) making actions unavailable and `destroy` (destructor) that leaves created action objects unaffected. This is potentially confusing: documentation should explain use-cases for each and guarantee ordering semantics when clients hold existing `ext_input_trigger_action_v1` objects.
>     - Also, `ext_input_trigger_action_manager_v1.get_input_trigger_action` is supposed to emit `unavailable` if the token is no longer available — the relationship between `cancel`, `destroy`, and token invalidation needs to be clarified.

I agree with this. I confuse all three quite a bit.

### Lower priority / niceties / improvements

> 14. Missing documentation about key repeat and update events
>     - For keyboard triggers, how should repeated key events (auto-repeat) be treated? Should they produce update events? The document states begin on depression and end on change, but it should clarify how repeats are treated.

I just had to handle this in the spike. I don't have a use case in mind where repeat would be that useful, so I believe we should specify that it's should not emit `update` events.

> 15. `progress` fixed type precision and range
>     - `progress` is a Wayland `fixed` normalized to [0.0, 1.0]. It’s good to specify whether values must be clamped, whether 1.0 must be sent before `end`, and how to interpret fractional precision (Wayland `fixed` is 16.16 usually — but calling out the expected precision is helpful).

Fair point, my opinion at the moment is that 1.0 must be sent.

> 16. No means to enumerate current triggers or query which triggers belong to an action
>     - Once you create an action and add triggers, there is no introspection API to list them later. This can make debugging and synchronization complex.

Again, I remember that we agreed that this should not be a part of these two protocols.

> 17. No way to attach metadata to tokens or actions
>     - `name` is a free-form description but there’s no capability for richer metadata / display name / icon that the compositor might show to users. Consider allowing optional metadata.

This might be a minor issue for applications that want to show button icons instead of `Ctrl + Shift + C`, but I don't believe it's a major issue at the moment.

> 18. Unclear how gloves/virtual keyboard input is handled
>     - Not critical, but worth mentioning: the protocol assumes physical devices; if applications synthesize input, should synthesized events trigger global actions?

I don't think virtual devices or OSKs should trigger global actions.

> 19. Documentation comments in XML mention “This protocol allows for a privileged Wayland client” but the actual interface offers no way for the compositor/ecosystem to differentiate privileged vs normal clients. You must define how compositors expose these globals (e.g., only to trusted clients) and how clients obtain the global.

We talked about this before as well. "Priviliged clients" can mean clients launched explicitly by the compositor. In which case we would advertise the globals. Otherwise, we don't.

tarek-y-ismail:
> Fair point, my opinion at the moment is that 1.0 must be sent.

Thinking about this again, maybe this isn't the best idea as it might introduce some unwanted jank. For example, if a user is moving to the right workspace via a gesture, the DE might have a threshold of 0.5 progress after which the move is done whether the gesture is cancelled or not. Requiring compositors to send 1.0 will just introduce complexity as they'll have to handle and ignore sudden jumps in progress.

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -0,0 +1,155 @@
...
+        undefined base. clients must not assume any relation to wall-clock
+        time; timestamps are only valid for comparison/deltas.
+
+        The activation token argument is a short-lived opaque token that may be
+        used with `xdg_activation_v1` to request to focus another surface.
+      </description>
+      <arg name="time" type="uint" summary="timestamp with millisecond granularity"/>
+      <arg name="activation_token" type="string" summary="an xdg_activation token"/>
+    </event>
+
```
Saviq: I think we decided against introducing that (yet)?

tarek-y-ismail: Yeah that was shorty after I updated the PR. Haven't gotten around to it again, sorry

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -0,0 +1,132 @@
...
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+    THE SOFTWARE.
+  </copyright>
+
+  <description summary="Registration for global input actions">
+    This protocol allows for a Wayland clients to subscribe to input trigger
+    actions for "begin", "update", and "end" notifications.
```
tarek-y-ismail:
```suggestion
    actions for "begin" and "end" notifications.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,241 @@
...
+      <description summary="destroy input trigger registration manager object">
+        This informs the compositor that the input trigger registration manager
+        object will no longer be used. Existing objects created through this
+        interface remain valid.
+      </description>
+    </request>
+
+    <enum name="modifiers" bitfield="true">
+      <description summary="modifier keys">
+        The various shift modifiers on the keyboard
```
tarek-y-ismail: "shift modifiers" might be confusing

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -0,0 +1,241 @@
...
+
+  <interface name="ext_input_trigger_action_control_v1" version="1">
+    <description summary="Allows configuration of an input trigger action">
+      In response to the creation of this object the compositor must send
+      the done event.
+    </description>
+
+    <request name="add_input_trigger_event">
+      <description summary="add an input trigger to the action">
+        This request adds an input trigger to those that activate the action.
```
tarek-y-ismail, line 196: Right now in the spike the compositor sends a `ext_input_trigger_action_control_v1.done` in response to each `add_input_trigger_event` request the clients sends. Is my understanding correct? Should this be written down in the protocol?

AlanGriffiths: In that case the spike is wrong. "done" is sent in response to `get_action_control`:

```xml
    <request name="get_action_control">
      <description summary="get input trigger action control object">
        This request creates an action control object.

        The name parameter describes the action (it has no semantic content and
        may be duplicated across actions).

        The compositor will send the ext_input_trigger_action_control_v1.done event in
        response to this request.
      </description>
```

(I don't think we need to document all the places it isn't sent.)

tarek-y-ismail: Aha, that makes more sense. I'll fix that in the spike then.

PR #4331, opened 2025-10-06
mg::Buffer: Add `map_readable`

RAOF:
This adds a `map_readable` method to the base `Buffer` class.

All¹ buffers can be read from the CPU in *some* way, and it would be useful in other code to have a guaranteed-available way to access buffer content.

¹: Well, *almost* all - see comment.

## How to test

Should not change normal usage of Mir. The branch https://github.com/canonical/mir/tree/hack-gbm-to-go-through-shm-path hacks into GBM buffer allocator to send all texture access through the new `map_readable` path, and can be used for extra manual testing.

## Checklist

- [ ] Tests added and pass


RAOF:
It's not clear that this can be usefully unit-tested; feel free to suggest some.

I think `PtrBackedMapping` could probably use some documentation, and I'll push that later; feel free to pipe up if I've missed anywhere else!

include/platform/mir/graphics/buffer.h
```diff
@@ -57,28 +72,44 @@ class NativeBufferBase
...
+    ///       such as [GLRenderingProvider::as_texture].
+    ///
+    /// \returns A [Mapping] describing a CPU-accessible view of the buffer. The lifetime of
+    ///          this [Mapping] is tied to the lifetime of this [Buffer].
+    ///
+    /// \throws An [UnmappableBuffer] if the buffer cannot be mapped in this way. This may
+    ///         occur if the buffer has a format not representable by [MirPixelFormat]
+    ///         (for example, a planar YUV format) or if the buffer is being protected by the
+    ///         GPU Digital Rights Management path.
+    /// \throws A std::system_error if mapping
```
Saviq: "if mapping…"?

src/platform/graphics/egl_wayland_allocator.cpp
```diff
@@ -241,11 +241,84 @@ class WaylandTexBuffer :
...
+
+        auto len() const -> size_t override
+        {
+            return size().height.as_uint32_t() * stride().as_uint32_t();
+        }
+
+        auto format() const -> MirPixelFormat override
+        {
+            // Because we're reading through GL the underlying format doesn't matter
+           return mir_pixel_format_argb_8888;
```
Saviq:
```suggestion
            return mir_pixel_format_argb_8888;
```

src/platform/graphics/linux_dmabuf.cpp
```diff
@@ -1158,24 +1188,168 @@ class DmabufTexBuffer :
...
+
+        auto len() const -> size_t override
+        {
+            return size().height.as_uint32_t() * stride().as_uint32_t();
+        }
+
+        auto format() const -> MirPixelFormat override
+        {
+            // Because we're reading through GL the underlying format doesn't matter
+           return mir_pixel_format_argb_8888;
```
Saviq:
```suggestion
            return mir_pixel_format_argb_8888;
```

Saviq (changes requested):
Formatting / doxygen nits

include/platform/mir/renderer/sw/pixel_source.h
```diff
@@ -167,9 +153,6 @@ class WriteTransferable
     virtual geometry::Size size() const = 0;
 };
 
-auto as_read_mappable(
```
robert-ancell, line 170: Does this need to be removed from `src/platform/symbols.map`?

tarek-y-ismail: Yup, and since it's and ABI breaking change, a mirserver version bump

RAOF:
https://github.com/canonical/wlcs/pull/387 is required to fix the (spurious) WLCS failure here.

include/platform/mir/renderer/sw/pixel_source.h
RAOF: ABI bump done!

include/platform/mir/graphics/buffer.h
RAOF: Ooops. The rest of the sentence is now there!

tarek-y-ismail:
@raof Poke :index_pointing_at_the_viewer:

include/core/mir/geometry/dimensions.h
```diff
@@ -50,6 +64,25 @@ struct Value
...
+    /// as<uint64_t> and as<float> do not exist,
+    /// as int32_t can contain negative values not representable
+    /// in uint64_t, and float does not have sufficient precision
+    /// to exactly represent any int32_t value.
+    template <typename Q>
+        requires conversion_is_lossless<T, Q>
+    constexpr Q as() const
+    {
+        return this->value;
+    }
```
AlanGriffiths, line 84: Cool! I was just thinking we need this (in a PR not far away)

include/core/mir/geometry/dimensions.h
```diff
@@ -18,17 +18,31 @@
...
 
 namespace mir
 {
 namespace geometry
 {
 namespace generic
 {
+// Concept guaranteeing conversion from U to T is lossless
+template<typename U, typename T>
+concept conversion_is_lossless = requires
```
AlanGriffiths: I don't think we've named concepts before, but I would expect `ConversionIsLossless` given our other conventions. (I.e. is closer to a type than to a variable or function)

RAOF:
Yes! Fixed TICS!

include/core/mir/geometry/dimensions.h
RAOF: Yeah, fair. Fixed!

Saviq (commented):
Is it worth introducing `MIR_USING_WLCS_DEV`? Shouldn't we just release WLCS and maintain the list of expected failing tests?

AlanGriffiths:
> Is it worth introducing `MIR_USING_WLCS_DEV`? Shouldn't we just release WLCS and maintain the list of expected failing tests?

Looking at the use:

```
 if ((NOT MIR_USING_WLCS_DEV) AND (WLCS_VERSION VERSION_LESS_EQUAL "1.8.1"))
```

That suggests that this is is going to require maintenance (or accumulate legacy) when wlcs is next released.

It seems that the dev WLCS versioning isn't telling us what we want - it really should be > 1.8.1. (Yes, I know we have a similar issue with Mir versioning)

RAOF:
> > Is it worth introducing `MIR_USING_WLCS_DEV`? Shouldn't we just release WLCS and maintain the list of expected failing tests?

The thing is that the test will be failing until we release wlcs *and* it's available in all the Fedora releases we test on. We probably don't want to disable tests until a new release; that's one of the reasons the Ubuntu tests pull wlcs from the dev PPA.

> Looking at the use:
> 
> ```
>  if ((NOT MIR_USING_WLCS_DEV) AND (WLCS_VERSION VERSION_LESS_EQUAL "1.8.1"))
> ```
> 
> That suggests that this is is going to require maintenance (or accumulate legacy) when wlcs is next released.

No? This'll work just fine when we release the next wlcs; it'll have the fixed test and have a version > 1.8.1, so this should never be wrong. When Fedora picks up the next version, it'll stop XFAILing the test.

This can then sit harmlessly there until we feel like removing it.

> It seems that the dev WLCS versioning isn't telling us what we want - it really should be > 1.8.1. (Yes, I know we have a similar issue with Mir versioning)

Yeah, having even/odd versioning would be a nice solution to this problem (we'd want >= 1.9 to match everything in the dev branch, and then release 1.10 that people are expected to package).

RAOF:
Thinking overnight, there's another option: marking the tests that we change the behaviour of with a detectable pkgconf variable; that's done in https://github.com/canonical/wlcs/pull/398

If that's what we want to go for, then the change here is to mark the test bad if the pkgconf variable `BadBuffer_test_truncated_shm_file_revision` is less than 1.

AlanGriffiths:
> This can then sit harmlessly there until we feel like removing it.

That's what I said: "require maintenance (or accumulate legacy)" :wink:

AlanGriffiths:
> Yeah, having even/odd versioning would be a nice solution to this problem (we'd want >= 1.9 to match everything in the dev branch, and then release 1.10 that people are expected to package).

Can we not just bump the version on `main` when branching a release (1.9, say) a release and use that (1.9) for the next release? Any changes on main will be in the next release. (Not true for all projects, but our development cycle just branches releases from main, we don't cherry-pick a bunch of features.)

AlanGriffiths:
I guess the  real issue is that CMake thinks `1.8.1+dev116-g775ad52-0ubuntu26.04 VERSION_EQUAL 1.8.1`. If it thought it were greater we'd not be discussing this. If it worked like `dpkg --compare-versions` we'd be fine already.

RAOF:
> > This can then sit harmlessly there until we feel like removing it.
> 
> That's what I said: "require maintenance (or accumulate legacy)" 😉

Then I'm not sure what your point is, here? A solution based on wlcs versioning will have exactly the same requirements (we'll put in a `WLCS_VERSION VERSION_GREATER 1.8.1`, that can sit harmlessly there until we feel like removing it)?

As far as I can tell, the only solution that *doesn't* “require maintenance or accumulate legacy” is embedding wlcs as a git submodule, and building it with Mir itself? (Although even that requires the maintenance of updating the version of wlcs we pull).

Saviq (approved):
Yups.

PR #4333, opened 2025-10-06
server: log security events according to OWASP format

Saviq:
## What's new?

OWASP-formatted security logging.

## How to test

```
env -u WAYLAND_DISPLAY ./build-Debug/bin/mir_demo_server 2>&1 | grep appid
env -u WAYLAND_DISPLAY ./build-Debug/bin/mir_demo_server --platform-display-libs mir:virtual 2>&1 | grep appid
```

## Checklist

### DIFF
--- a/src/common/symbols.map
+++ b/src/common/symbols.map
@@ -700,7 +700,6 @@ MIR_COMMON_INTERNAL_2.24 {
 global:
   extern "C++" {
     mir::default_font*;
-    mir::security_fmt*;
     mir::security_log*;
   };
 } MIR_COMMON_INTERNAL_2.22;


AlanGriffiths:
Where are the format requirements?

This seems somewhat redundant:
```
[2025-10-06 17:20:51.029067] < -warning- > security: {"datetime": "2025-10-06T17:20:51Z", "appid": "mir.canonical.com", "event": "sys_startup", "level": "WARN", "description": "Mir is starting up" }
```

And, when miral-shell crashed I got:
```
Mir fatal error: Unsupported attempt to continue after a fatal signal: SIGSEGV
Aborted (core dumped)
```

Shouldn't we also be logging a security event when we get a "Fatal signal received"?

Saviq:
> Where are the format requirements?

Per the doc: https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Logging_Vocabulary_Cheat_Sheet.md

> This seems somewhat redundant:

Didn't want it to stand out among other lines… this also puts it in the `security` component, which may be useful?

> Shouldn't we also be logging a security event when we get a "Fatal signal received"?

Will do.

Saviq:
> Will do.

Done.

src/server/run_mir.cpp
```diff
@@ -162,6 +163,9 @@ extern "C" [[noreturn]] void fatal_signal_cleanup(int sig, siginfo_t* info, void
         constexpr size_t len = std::char_traits<char>::length(warning);
         [[maybe_unused]]auto n = write(STDERR_FILENO, warning, len);
     }
+    auto security_log = mir::security_fmt(mir::logging::Severity::error, "sys_crash", std::string{"Mir fatal signal received: "} + signum_to_string(sig)) + "\n";
```
AlanGriffiths: This attempts to allocate memory in a signal context. That is undefined behaviour (and we can avoid it here).

Also the reference above says "sys_crash...Level: WARN" is that prescriptive?

AlanGriffiths (changes requested):
> Done.

So that gives the following:

```
$ cmake-build-debug/bin/miral-app 2>&1 | grep appid
[2025-10-07 10:27:07.921636] < -warning- > security: {"datetime": "2025-10-07T10:27:07Z", "appid": "mir.canonical.com", "event": "sys_startup", "level": "WARN", "description": "Mir is starting up" }
{"datetime": "2025-10-07T10:27:08Z", "appid": "mir.canonical.com", "event": "sys_crash", "level": "ERROR", "description": "Mir fatal signal received: SIGSEGV" }
{"datetime": "2025-10-07T10:27:08Z", "appid": "mir.canonical.com", "event": "sys_crash", "level": "ERROR", "description": "Mir fatal signal received: SIGABRT" }
```
Where only the first entry has the `[2025-10-07 10:27:07.921636] < -warning- > security: ` prefix from the default logging engine. It isn't clear to me that this variability is going to be easy to consume. Grepping for "stuff between braces somewhere on the line" seems prone to false positives.

I'm still unclear on what this should look like: you linked to a document that identifies fields required in the log, but I'm not sure that this format is the correct interpretation.

It may be that we should not be conflating application level logging with OWASP security events. Having a separate stream would simply consuming these events.

src/server/report_exception.cpp
```diff
@@ -48,6 +51,15 @@ void mir::report_exception(std::ostream& out_stream, std::ostream& err_stream)
...
         err_stream << "ERROR: " << boost::diagnostic_information(error) << std::endl;
+
+        // Log a security event
+        std::string function_str = "<function unknown>";
+        if (auto function = boost::get_error_info<boost::throw_function>(error))
+        {
+            function_str = *function;
+        }
+        auto message = std::format("Mir unhandled exception in: {}", function_str);
+        err_stream << mir::security_fmt(mir::logging::Severity::error, "sys_crash", message) << std::endl;
```
AlanGriffiths: There is no reason to report this exception is unhandled.

E.g. in `miral::MirRunner::Self::run_with()` this is called from catch blocks that handle the exception.

In `WindowManagementTrace` it is called as the exception is propagated to the caller.

src/server/run_mir.cpp
```diff
@@ -162,6 +163,12 @@ extern "C" [[noreturn]] void fatal_signal_cleanup(int sig, siginfo_t* info, void
         constexpr size_t len = std::char_traits<char>::length(warning);
         [[maybe_unused]]auto n = write(STDERR_FILENO, warning, len);
     }
+    char security_log[] = "{\"datetime\": \"YYYY-MM-DDThh:mm:ssZ\", \"appid\": \"mir.canonical.com\", \"event\": \"sys_crash\", \"level\": \"WARN\", \"description\": \"Mir fatal signal received\" }\n";
```
AlanGriffiths: Is this appid correct? I would expect it to reflect the current compositor (e.g. Ubuntu Frame or Miriway).

AlanGriffiths: You're not adding the signal to the buffer? (`strncpy()` is async signal safe)

src/common/log.cpp
```diff
@@ -84,4 +87,42 @@ void log(
...
+    auto now = std::chrono::system_clock::now();
+    auto time_t = std::chrono::system_clock::to_time_t(now);
+
+    std::stringstream ss;
+    ss << std::put_time(std::localtime(&time_t), "%FT%TZ");
+
+    return std::format(
+        "{{"
+            "\"datetime\": \"{}\", "
+            "\"appid\": \"mir.canonical.com\", "
```
AlanGriffiths: Should this be hardcoded? I don't expect the appid to come from a library the process uses.

src/server/run_mir.cpp
```diff
@@ -17,6 +17,7 @@
 #include "mir/run_mir.h"
 #include "mir/terminate_with_current_exception.h"
 #include "mir/display_server.h"
+#include "mir/log.h"
```
AlanGriffiths: Not used?

AlanGriffiths (changes requested):
Down to the nits now 😁

src/common/log.cpp
Saviq: We don't have API for the compositor to identify itself to us that I know of? `argv[0]`?

src/server/run_mir.cpp
Saviq: > You're not adding the signal to the buffer?

Didn't think it adds anything worth the added complexity if anyone needs those details, they will hunt in logs around the OWASP notice.

src/common/log.cpp
AlanGriffiths: > We don't have API for the compositor to identify itself to us that I know of? `argv[0]`?

This is stored on `mir::Server::Self` - but that's not globally accessible.

There is an answer on [stack overflow](https://stackoverflow.com/questions/9097201/how-to-get-current-process-name-in-linux) that may be usable.

src/server/run_mir.cpp
```diff
@@ -162,6 +168,22 @@ extern "C" [[noreturn]] void fatal_signal_cleanup(int sig, siginfo_t* info, void
...
+    char security_tail[] = "\", \"event\": \"crash\", \"description\": \"Fatal signal received\"}\n";
+    std::time_t time = std::time(nullptr);
+    // strftime isn't listed as async-signal-safe, but neither "%F" or "%T" use locale information
+    [[maybe_unused]]auto n = std::strftime(security_head + 14, 21, "%FT%TZ", std::gmtime(&time));
+    n = write(STDERR_FILENO, security_head, sizeof(security_head) - 1);
+    n = write(STDERR_FILENO, security_mid, sizeof(security_mid) - 1);
+    if (program_invocation_short_name)
+    {
+        n = write(STDERR_FILENO, program_invocation_short_name, std::strlen(program_invocation_short_name));
+    }
```
AlanGriffiths, line 179: This leads to an ill-formed log if `!program_invocation_short_name`: Either move writing the `"appid:"` into the condition or provide an `else` clause with a dummy value.

include/common/mir/log.h
```diff
@@ -40,6 +40,28 @@ void log(
...
 
+/// Format a security log message according to the OWASP specification
+///
+/// \param severity The severity of the event
+/// \param event A short string identifying the event type
+/// \param description A human-readable description of the event
+auto security_fmt(
+    logging::Severity severity,
+    std::string const& event,
+    std::string const& description) -> std::string;
```
AlanGriffiths: Do we need this as a separate function?

src/server/report_exception.cpp
```diff
@@ -16,9 +16,12 @@
 
 #include "mir/report_exception.h"
 #include "mir/abnormal_exit.h"
+#include "mir/log.h"
 
+#include <boost/exception/exception.hpp>
 #include <boost/exception/diagnostic_information.hpp>
 
+#include <format>
```
AlanGriffiths: These changes look spurious

src/server/run_mir.cpp
```diff
@@ -32,6 +32,12 @@
 #include <unistd.h>
 #include <boost/throw_exception.hpp>
 
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <errno.h>
```
AlanGriffiths: Not needed

src/common/log.cpp
```diff
@@ -17,10 +17,19 @@
...
+#include <iostream>
 
 #include <exception>
 #include <boost/exception/diagnostic_information.hpp>
 
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+#include <errno.h>
+extern char* program_invocation_short_name;  // TICS !cppcoreguidelines-avoid-non-const-global-variables
```
AlanGriffiths: Why duplicate the declaration from `errno.h`? At best it is unnecessary, at worst it will break things

src/common/symbols.map
```diff
@@ -700,5 +700,7 @@ MIR_COMMON_INTERNAL_2.24 {
 global:
   extern "C++" {
     mir::default_font*;
+    mir::security_fmt*;
```
AlanGriffiths:
```suggestion
```

PR #4336, opened 2025-10-06
Implementation of an evdev platform in rust 

mattkae:
## What's new?

- Added an evdev-rs platform that implements libinput in rust
- Wrote a bunch of bindings to facilitate it

## How to test?
1. Build the project
2. Modify `src/platforms/evdev-rs/platform_factory.cpp` to make the evdev-rs platform supported (hack for now until it is ready!)
3. Run:
    ```sh
    ./bin/miral-app --platform-path=/path/to/lib/server-modules --platform-input-lib=mir:evdev-input-rs
    ```
4. Use your keyboard, mouse, and scroll wheel and note that they work

## What's next?
- Implement touch gestures
- Fix some rough edges
- Event reporting

### DIFF
--- a/src/platforms/evdev-rs/src/lib.rs
+++ b/src/platforms/evdev-rs/src/lib.rs
@@ -23,1203 +23,20 @@
 //    instead.
 // 2. All enums are changed to i32 for ABI stability
 
-use cxx;
-use cxx::SharedPtr;
-use input;
-use input::event;
-use input::event::keyboard;
-use input::event::keyboard::KeyboardEventTrait;
-use input::event::pointer;
-use input::event::pointer::PointerEventTrait;
-use input::event::pointer::PointerScrollEvent;
-use input::event::EventTrait;
-use input::AsRaw;
-use libc;
-use nix::unistd;
-use std::os::fd::{AsRawFd, FromRawFd};
-use std::os::unix::ffi::OsStrExt;
-use std::os::unix::io;
-use std::sync::Arc;
-use std::sync::Mutex;
-use std::thread;
-
-use crate::ffi::PlatformBridge;
-use cxx::ExternType;
-
-/// Creates a platform on the Rust side of things.
-///
-/// The `bridge` and `device_registry` parameters are provided by the C++ side of things.
-/// The C++ platform simply forwards all calls to the Rust implementation.
-pub fn evdev_rs_create(
-    bridge: cxx::SharedPtr<ffi::PlatformBridge>,
-    device_registry: cxx::SharedPtr<ffi::InputDeviceRegistry>,
-) -> Box<PlatformRs> {
-    return Box::new(PlatformRs {
-        bridge,
-        device_registry: device_registry,
-        state: None,
-    });
-}
-
-pub struct PlatformRs {
-    /// The platform bridge provides access to functionality that must happen in C++.
-    bridge: cxx::SharedPtr<ffi::PlatformBridge>,
-
-    /// The input device registry is used for registering and unregistering input devices.
-    device_registry: cxx::SharedPtr<ffi::InputDeviceRegistry>,
-
-    state: Option<Arc<Mutex<LibinputLoopState>>>,
-}
-
-impl PlatformRs {
-    /// Start the input platform.
-    ///
-    /// This method will spawn a thread to handle input events from both
-    /// libinput and Mir. The thread will run until `stop()` is called.
-    pub fn start(&mut self) {
-        println!("Starting the evdev-rs platform");
-
-        let bridge = self.bridge.clone();
-        let mut libinput = input::Libinput::new_with_udev(LibinputInterfaceImpl {
-            bridge: bridge.clone(),
-            fds: Vec::new(),
-        });
-
-        // TODO: This does not handle multi-seat. If/when we do multi-seat, this will need to be refactored.
-        let started = match libinput.udev_assign_seat("seat0") {
-            Err(_) => false,
-            _ => true,
-        };
-
-        if !started {
-            eprintln!("Failed to call udev_assign_seat, not running the libinput loop");
-            return;
-        }
-
-        self.state = Some(Arc::new(Mutex::new(LibinputLoopState {
-            libinput: libinput,
-            known_devices: Vec::new(),
-            next_device_id: 0,
-            button_state: 0,
-            pointer_x: 0.0,
-            pointer_y: 0.0,
-            scroll_axis_x_accum: 0.0,
-            scroll_axis_y_accum: 0.0,
-            x_scroll_scale: 1.0,
-            y_scroll_scale: 1.0,
-        })));
-    }
-
-    pub unsafe fn libinput_fd(&mut self) -> i32 {
-        match self.state.as_mut() {
-            Some(state_arc) => match state_arc.lock() {
-                Ok(state) => state.libinput.as_raw_fd(),
-                Err(_) => -1,
-            },
-            None => -1,
-        }
-    }
-
-    pub fn continue_after_config(&self) {}
-    pub fn pause_for_config(&self) {}
-
-    /// Stop the input platform.
-    ///
-    /// This method will signal the input thread to stop and wait for it to finish.
-    pub fn stop(&mut self) {
-        // Note: When the main loop exits, we need to remove all devices from the device registry
-        // or else Mir will try to free the devices later but we won't have access to the platform
-        // code at that point. This results in a segfault.
-        let state_opt = self.state.as_mut();
-        if state_opt.is_none() {
-            // Platform not started or already stopped.
-            self.state = None;
-            return;
-        }
-
-        match state_opt.unwrap().lock() {
-            Ok(mut state_guard) => {
-                for device_info in state_guard.known_devices.iter_mut().rev() {
-                    println!("Removing device id {} from registry", device_info.id);
-                    device_info.input_sink = None;
-                    device_info.event_builder = None;
-                    // # Safety
-                    //
-                    // Because we need to use pin_mut_unchecked, this is unsafe.
-                    unsafe {
-                        self.device_registry
-                            .clone()
-                            .pin_mut_unchecked()
-                            .remove_device(&device_info.input_device);
-                    }
-                }
-            }
-            Err(_) => {
-                println!("Unable to gain access to device info; lock poisoned");
-            }
-        }
-
-        self.state = None;
-    }
-
-    pub fn create_device_observer(&self) -> Box<DeviceObserverRs> {
-        Box::new(DeviceObserverRs::new())
-    }
-
-    fn create_input_device(&mut self, device_id: i32) -> Box<InputDeviceRs> {
-        let state_arc = match self.state.as_mut() {
-            Some(s) => s.clone(),
-            None => {
-                println!(
-                    "PlatformRs::create_input_device: state not initialized; creating fallback state"
-                );
-                let mut libinput = input::Libinput::new_with_udev(LibinputInterfaceImpl {
-                    bridge: self.bridge.clone(),
-                    fds: Vec::new(),
-                });
-
-                let started = match libinput.udev_assign_seat("seat0") {
-                    Err(_) => false,
-                    _ => true,
-                };
-                if !started {
-                    println!(
-                        "PlatformRs::create_input_device: failed to assign seat for fallback state"
-                    );
-                }
-
-                Arc::new(Mutex::new(LibinputLoopState {
-                    libinput,
-                    known_devices: Vec::new(),
-                    next_device_id: 0,
-                    button_state: 0,
-                    pointer_x: 0.0,
-                    pointer_y: 0.0,
-                    scroll_axis_x_accum: 0.0,
-                    scroll_axis_y_accum: 0.0,
-                    x_scroll_scale: 1.0,
-                    y_scroll_scale: 1.0,
-                }))
-            }
-        };
-
-        Box::new(InputDeviceRs {
-            device_id,
-            state: state_arc,
-            bridge: self.bridge.clone(),
-        })
-    }
-
-    pub fn process(&mut self) {
-        if self.state.is_none() {
-            return;
-        }
-
-        match self.state.as_mut().unwrap().try_lock() {
-            Ok(mut state) => {
-                PlatformRs::process_libinput_events(
-                    &mut *state,
-                    self.device_registry.clone(),
-                    self.bridge.clone(),
-                );
-            }
-            _ => {}
-        }
-    }
-
-    fn process_libinput_events(
-        state: &mut LibinputLoopState,
-        device_registry: cxx::SharedPtr<ffi::InputDeviceRegistry>,
-        bridge: cxx::SharedPtr<ffi::PlatformBridge>,
-    ) {
-        fn handle_input(
-            input_sink: &mut Option<InputSinkPtr>,
-            event: cxx::SharedPtr<ffi::MirEvent>,
-        ) {
-            if let Some(input_sink) = input_sink {
-                input_sink.handle_input(&event);
-            }
-        }
-
-        fn get_scroll_axis(
-            value120: f64,
-            value: f64,
-            scale: f64,
-            accum: &mut f64,
-        ) -> (f64, f64, f64, bool) {
-            let precise = value * scale;
-            let stop = precise == 0.0;
-            *accum = *accum + value120;
-
-            let discrete = *accum / 120.0;
-            *accum = *accum % 120.0;
-            return (precise, discrete, value120, stop);
-        }
-
-        fn get_device_info_from_libinput_device<'a>(
-            known_devices: &'a mut Vec<DeviceInfo>,
-            libinput_device: &input::Device,
-        ) -> Option<&'a mut DeviceInfo> {
-            known_devices
-                .iter_mut()
-                .find(|d| d.device.as_raw() == libinput_device.as_raw())
-        }
-
-        if state.libinput.dispatch().is_err() {
-            println!("Error dispatching libinput events");
-            return;
-        }
-
-        for event in &mut state.libinput {
-            let libinput_device = event.device();
-
-            match event {
-                input::Event::Device(device_event) => match device_event {
-                    event::DeviceEvent::Added(_) => {
-                        state.known_devices.push(DeviceInfo {
-                            id: state.next_device_id,
-                            device: libinput_device,
-                            input_device: bridge.create_input_device(state.next_device_id),
-                            input_sink: None,
-                            event_builder: None,
-                        });
-
-                        // The device registry may call back into the input device, but the input device
-                        // may call back into this thread to retrieve information. To avoid deadlocks, we
-                        // have to queue this work onto a new thread to fire and forget it. It's not a big
-                        // deal, but it is a bit unfortunate.
-                        let input_device = state.known_devices.last().unwrap().input_device.clone();
-                        let device_registry = device_registry.clone();
-
-                        // # Safety
-                        //
-                        thread::spawn(move || unsafe {
-                            device_registry
-                                .clone()
-                                .pin_mut_unchecked()
-                                .add_device(&input_device);
-                        });
-
-                        state.next_device_id += 1;
-                    }
-                    event::DeviceEvent::Removed(removed_event) => {
-                        let dev = removed_event.device();
-                        let index = state
-                            .known_devices
-                            .iter()
-                            .position(|x| x.device.as_raw() == dev.as_raw());
-                        if let Some(index) = index {
-                            // # Safety
-                            //
-                            // Because we need to use pin_mut_unchecked, this is unsafe.
-                            unsafe {
-                                device_registry
-                                    .clone()
-                                    .pin_mut_unchecked()
-                                    .remove_device(&state.known_devices[index].input_device);
-                            }
-                            state.known_devices.remove(index);
-                        }
-                    }
-                    _ => println!("TODO: Unhandled device event type"),
-                },
-
-                input::Event::Pointer(pointer_event) => {
-                    if let Some(device_info) = get_device_info_from_libinput_device(
-                        &mut state.known_devices,
-                        &libinput_device,
-                    ) {
-                        match pointer_event {
-                            event::PointerEvent::Motion(motion_event) => {
-                                if let Some(event_builder) = &mut device_info.event_builder {
-                                    let pointer_event = PointerEventDataRs {
-                                        has_time: true,
-                                        time_microseconds: motion_event.time_usec(),
-                                        action: ffi::MirPointerAction::mir_pointer_action_motion
-                                            .repr,
-                                        buttons: state.button_state,
-                                        has_position: false,
-                                        position_x: 0 as f32,
-                                        position_y: 0 as f32,
-                                        displacement_x: motion_event.dx() as f32,
-                                        displacement_y: motion_event.dy() as f32,
-                                        axis_source:
-                                            ffi::MirPointerAxisSource::mir_pointer_axis_source_none
-                                                .repr,
-                                        precise_x: 0.0,
-                                        discrete_x: 0,
-                                        value120_x: 0,
-                                        scroll_stop_x: false,
-                                        precise_y: 0.0,
-                                        discrete_y: 0,
-                                        value120_y: 0,
-                                        scroll_stop_y: false,
-                                    };
-                                    handle_input(
-                                        &mut device_info.input_sink,
-                                        event_builder.pin_mut().pointer_event(&pointer_event),
-                                    );
-                                }
-                            }
-
-                            event::PointerEvent::MotionAbsolute(absolute_motion_event) => {
-                                if let Some(event_builder) = &mut device_info.event_builder {
-                                    if let Some(input_sink) = &mut device_info.input_sink {
-                                        // # Safety
-                                        //
-                                        // Calling as_ref on a NonNull is unsafe.
-                                        let sink_ref: &ffi::InputSink =
-                                            unsafe { input_sink.0.as_ref() };
-                                        let bounding_rect = bridge.bounding_rectangle(sink_ref);
-                                        let width = bounding_rect.width() as u32;
-                                        let height = bounding_rect.height() as u32;
-
-                                        let old_x = state.pointer_x;
-                                        let old_y = state.pointer_y;
-                                        state.pointer_x = absolute_motion_event
-                                            .absolute_x_transformed(width)
-                                            as f32;
-                                        state.pointer_y = absolute_motion_event
-                                            .absolute_y_transformed(height)
-                                            as f32;
-
-                                        let movement_x = state.pointer_x - old_x;
-                                        let movement_y = state.pointer_y - old_y;
-                                        let pointer_event = PointerEventDataRs{
-                                            has_time: true,
-                                            time_microseconds: absolute_motion_event.time_usec(),
-                                            action: ffi::MirPointerAction::mir_pointer_action_motion.repr,
-                                            buttons: state.button_state,
-                                            has_position: true,
-                                            position_x: state.pointer_x,
-                                            position_y: state.pointer_y,
-                                            displacement_x: movement_x,
-                                            displacement_y: movement_y,
-                                            axis_source: ffi::MirPointerAxisSource::mir_pointer_axis_source_none.repr,
-                                            precise_x: 0.0,
-                                            discrete_x: 0,
-                                            value120_x: 0,
-                                            scroll_stop_x: false,
-                                            precise_y: 0.0,
-                                            discrete_y: 0,
-                                            value120_y: 0,
-                                            scroll_stop_y: false,
-                                        };
-                                        handle_input(
-                                            &mut device_info.input_sink,
-                                            event_builder.pin_mut().pointer_event(&pointer_event),
-                                        );
-                                    }
-                                }
-                            }
-
-                            event::PointerEvent::ScrollWheel(scroll_wheel_event) => {
-                                if let Some(event_builder) = &mut device_info.event_builder {
-                                    let (precise_x, discrete_x, value120_x, stop_x) =
-                                        if scroll_wheel_event
-                                            .has_axis(input::event::pointer::Axis::Horizontal)
-                                        {
-                                            let scroll_value120_x = scroll_wheel_event
-                                                .scroll_value_v120(
-                                                    input::event::pointer::Axis::Horizontal,
-                                                );
-                                            let scroll_value_x = scroll_wheel_event.scroll_value(
-                                                input::event::pointer::Axis::Horizontal,
-                                            );
-                                            get_scroll_axis(
-                                                scroll_value120_x,
-                                                scroll_value_x,
-                                                state.x_scroll_scale as f64,
-                                                &mut state.scroll_axis_x_accum,
-                                            )
-                                        } else {
-                                            (0.0, 0.0, 0.0, true)
-                                        };
-
-                                    let (precise_y, discrete_y, value120_y, stop_y) =
-                                        if scroll_wheel_event
-                                            .has_axis(input::event::pointer::Axis::Vertical)
-                                        {
-                                            let scroll_value120_y = scroll_wheel_event
-                                                .scroll_value_v120(
-                                                    input::event::pointer::Axis::Vertical,
-                                                );
-                                            let scroll_value_y = scroll_wheel_event.scroll_value(
-                                                input::event::pointer::Axis::Vertical,
-                                            );
-                                            get_scroll_axis(
-                                                scroll_value120_y,
-                                                scroll_value_y,
-                                                state.y_scroll_scale as f64,
-                                                &mut state.scroll_axis_y_accum,
-                                            )
-                                        } else {
-                                            (0.0, 0.0, 0.0, true)
-                                        };
-
-                                    let pointer_event = PointerEventDataRs {
-                                        has_time: true,
-                                        time_microseconds: scroll_wheel_event.time_usec(),
-                                        action: ffi::MirPointerAction::mir_pointer_action_motion
-                                            .repr,
-                                        buttons: state.button_state,
-                                        has_position: false,
-                                        position_x: 0.0,
-                                        position_y: 0.0,
-                                        displacement_x: 0.0,
-                                        displacement_y: 0.0,
-                                        axis_source:
-                                            ffi::MirPointerAxisSource::mir_pointer_axis_source_wheel
-                                                .repr,
-                                        precise_x: precise_x as f32,
-                                        discrete_x: discrete_x as i32,
-                                        value120_x: value120_x as i32,
-                                        scroll_stop_x: stop_x,
-                                        precise_y: precise_y as f32,
-                                        discrete_y: discrete_y as i32,
-                                        value120_y: value120_y as i32,
-                                        scroll_stop_y: stop_y,
-                                    };
-                                    handle_input(
-                                        &mut device_info.input_sink,
-                                        event_builder.pin_mut().pointer_event(&pointer_event),
-                                    );
-                                }
-                            }
-
-                            event::PointerEvent::ScrollContinuous(scroll_continuous_event) => {
-                                if let Some(event_builder) = &mut device_info.event_builder {
-                                    let (precise_x, discrete_x, value120_x, stop_x) =
-                                        if scroll_continuous_event
-                                            .has_axis(input::event::pointer::Axis::Horizontal)
-                                        {
-                                            let scroll_value120_x = 0.0;
-                                            let scroll_value_x = scroll_continuous_event
-                                                .scroll_value(
-                                                    input::event::pointer::Axis::Horizontal,
-                                                );
-                                            get_scroll_axis(
-                                                scroll_value120_x,
-                                                scroll_value_x,
-                                                state.x_scroll_scale as f64,
-                                                &mut state.scroll_axis_x_accum,
-                                            )
-                                        } else {
-                                            (0.0, 0.0, 0.0, true)
-                                        };
-
-                                    let (precise_y, discrete_y, value120_y, stop_y) =
-                                        if scroll_continuous_event
-                                            .has_axis(input::event::pointer::Axis::Vertical)
-                                        {
-                                            let scroll_value120_y = 0.0;
-                                            let scroll_value_y = scroll_continuous_event
-                                                .scroll_value(
-                                                    input::event::pointer::Axis::Vertical,
-                                                );
-                                            get_scroll_axis(
-                                                scroll_value120_y,
-                                                scroll_value_y,
-                                                state.y_scroll_scale as f64,
-                                                &mut state.scroll_axis_y_accum,
-                                            )
-                                        } else {
-                                            (0.0, 0.0, 0.0, true)
-                                        };
-
-                                    let pointer_event = PointerEventDataRs {
-                                        has_time: true,
-                                        time_microseconds: scroll_continuous_event.time_usec(),
-                                        action: ffi::MirPointerAction::mir_pointer_action_motion
-                                            .repr,
-                                        buttons: state.button_state,
-                                        has_position: false,
-                                        position_x: 0.0,
-                                        position_y: 0.0,
-                                        displacement_x: 0.0,
-                                        displacement_y: 0.0,
-                                        axis_source:
-                                            ffi::MirPointerAxisSource::mir_pointer_axis_source_wheel
-                                                .repr,
-                                        precise_x: precise_x as f32,
-                                        discrete_x: discrete_x as i32,
-                                        value120_x: value120_x as i32,
-                                        scroll_stop_x: stop_x,
-                                        precise_y: precise_y as f32,
-                                        discrete_y: discrete_y as i32,
-                                        value120_y: value120_y as i32,
-                                        scroll_stop_y: stop_y,
-                                    };
-                                    handle_input(
-                                        &mut device_info.input_sink,
-                                        event_builder.pin_mut().pointer_event(&pointer_event),
-                                    );
-                                }
-                            }
-
-                            event::PointerEvent::ScrollFinger(scroll_finger_event) => {
-                                if let Some(event_builder) = &mut device_info.event_builder {
-                                    let (precise_x, discrete_x, value120_x, stop_x) =
-                                        if scroll_finger_event
-                                            .has_axis(input::event::pointer::Axis::Horizontal)
-                                        {
-                                            let scroll_value120_x = 0.0;
-                                            let scroll_value_x = scroll_finger_event.scroll_value(
-                                                input::event::pointer::Axis::Horizontal,
-                                            );
-                                            get_scroll_axis(
-                                                scroll_value120_x,
-                                                scroll_value_x,
-                                                state.x_scroll_scale as f64,
-                                                &mut state.scroll_axis_x_accum,
-                                            )
-                                        } else {
-                                            (0.0, 0.0, 0.0, true)
-                                        };
-
-                                    let (precise_y, discrete_y, value120_y, stop_y) =
-                                        if scroll_finger_event
-                                            .has_axis(input::event::pointer::Axis::Vertical)
-                                        {
-                                            let scroll_value120_y = 0.0;
-                                            let scroll_value_y = scroll_finger_event.scroll_value(
-                                                input::event::pointer::Axis::Vertical,
-                                            );
-                                            get_scroll_axis(
-                                                scroll_value120_y,
-                                                scroll_value_y,
-                                                state.y_scroll_scale as f64,
-                                                &mut state.scroll_axis_y_accum,
-                                            )
-                                        } else {
-                                            (0.0, 0.0, 0.0, true)
-                                        };
-
-                                    let pointer_event = PointerEventDataRs {
-                                        has_time: true,
-                                        time_microseconds: scroll_finger_event.time_usec(),
-                                        action: ffi::MirPointerAction::mir_pointer_action_motion
-                                            .repr,
-                                        buttons: state.button_state,
-                                        has_position: false,
-                                        position_x: 0.0,
-                                        position_y: 0.0,
-                                        displacement_x: 0.0,
-                                        displacement_y: 0.0,
-                                        axis_source:
-                                            ffi::MirPointerAxisSource::mir_pointer_axis_source_wheel
-                                                .repr,
-                                        precise_x: precise_x as f32,
-                                        discrete_x: discrete_x as i32,
-                                        value120_x: value120_x as i32,
-                                        scroll_stop_x: stop_x,
-                                        precise_y: precise_y as f32,
-                                        discrete_y: discrete_y as i32,
-                                        value120_y: value120_y as i32,
-                                        scroll_stop_y: stop_y,
-                                    };
-                                    handle_input(
-                                        &mut device_info.input_sink,
-                                        event_builder.pin_mut().pointer_event(&pointer_event),
-                                    );
-                                }
-                            }
-
-                            event::PointerEvent::Button(button_event) => {
-                                if let Some(event_builder) = &mut device_info.event_builder {
-                                    const BTN_LEFT: u32 = 0x110;
-                                    const BTN_RIGHT: u32 = 0x111;
-                                    const BTN_MIDDLE: u32 = 0x112;
-                                    const BTN_SIDE: u32 = 0x113;
-                                    const BTN_EXTRA: u32 = 0x114;
-                                    const BTN_FORWARD: u32 = 0x115;
-                                    const BTN_BACK: u32 = 0x116;
-                                    const BTN_TASK: u32 = 0x117;
-
-                                    let mir_button = match button_event.button() {
-                                        BTN_LEFT => {
-                                            if device_info.device.config_left_handed() {
-                                                ffi::MirPointerButton::mir_pointer_button_secondary
-                                            } else {
-                                                ffi::MirPointerButton::mir_pointer_button_primary
-                                            }
-                                        }
-                                        BTN_RIGHT => {
-                                            if device_info.device.config_left_handed() {
-                                                ffi::MirPointerButton::mir_pointer_button_primary
-                                            } else {
-                                                ffi::MirPointerButton::mir_pointer_button_secondary
-                                            }
-                                        }
-                                        BTN_MIDDLE => {
-                                            ffi::MirPointerButton::mir_pointer_button_tertiary
-                                        }
-                                        BTN_BACK => ffi::MirPointerButton::mir_pointer_button_back,
-                                        BTN_FORWARD => {
-                                            ffi::MirPointerButton::mir_pointer_button_forward
-                                        }
-                                        BTN_SIDE => ffi::MirPointerButton::mir_pointer_button_side,
-                                        BTN_EXTRA => {
-                                            ffi::MirPointerButton::mir_pointer_button_extra
-                                        }
-                                        BTN_TASK => ffi::MirPointerButton::mir_pointer_button_task,
-                                        _ => ffi::MirPointerButton::mir_pointer_button_primary,
-                                    };
-
-                                    let action: ffi::MirPointerAction = if button_event
-                                        .button_state()
-                                        == pointer::ButtonState::Pressed
-                                    {
-                                        state.button_state = state.button_state | mir_button.repr;
-                                        ffi::MirPointerAction::mir_pointer_action_button_down
-                                    } else {
-                                        state.button_state =
-                                            state.button_state & !(mir_button.repr);
-                                        ffi::MirPointerAction::mir_pointer_action_button_up
-                                    };
-
-                                    let pointer_event = PointerEventDataRs {
-                                        has_time: true,
-                                        time_microseconds: button_event.time_usec(),
-                                        action: action.repr,
-                                        buttons: state.button_state,
-                                        has_position: false,
-                                        position_x: 0.0,
-                                        position_y: 0.0,
-                                        displacement_x: 0.0,
-                                        displacement_y: 0.0,
-                                        axis_source:
-                                            ffi::MirPointerAxisSource::mir_pointer_axis_source_none
-                                                .repr,
-                                        precise_x: 0.0,
-                                        discrete_x: 0,
-                                        value120_x: 0,
-                                        scroll_stop_x: false,
-                                        precise_y: 0.0,
-                                        discrete_y: 0,
-                                        value120_y: 0,
-                                        scroll_stop_y: false,
-                                    };
-                                    handle_input(
-                                        &mut device_info.input_sink,
-                                        event_builder.pin_mut().pointer_event(&pointer_event),
-                                    );
-                                }
-                            }
-
-                            #[allow(deprecated)]
-                            event::PointerEvent::Axis(_) => {} // Deprecated, unhandled.
-                            _ => println!("TODO: Unhandled pointer event type"),
-                        }
-                    }
-                }
-
-                input::Event::Keyboard(keyboard_event) => {
-                    if let Some(device_info) = get_device_info_from_libinput_device(
-                        &mut state.known_devices,
-                        &libinput_device,
-                    ) {
-                        match keyboard_event {
-                            event::KeyboardEvent::Key(key_event) => {
-                                let keyboard_action = match key_event.key_state() {
-                                    keyboard::KeyState::Pressed => {
-                                        ffi::MirKeyboardAction::mir_keyboard_action_down
-                                    }
-                                    keyboard::KeyState::Released => {
-                                        ffi::MirKeyboardAction::mir_keyboard_action_up
-                                    }
-                                };
-
-                                if let Some(event_builder) = &mut device_info.event_builder {
-                                    let key_event_data = ffi::KeyEventData {
-                                        has_time: true,
-                                        time_microseconds: key_event.time_usec(),
-                                        action: keyboard_action.repr,
-                                        scancode: key_event.key(),
-                                    };
-                                    let created =
-                                        event_builder.pin_mut().key_event(&key_event_data);
-
-                                    if let Some(input_sink) = &mut device_info.input_sink {
-                                        input_sink.handle_input(&created);
-                                    }
-                                }
-                            }
-                            _ => {}
-                        }
-                    }
-                }
-
-                input::Event::Touch(_) => println!("TODO: Handle touch events"),
-
-                input::Event::Tablet(_) => println!("TODO: Handle tablet events"),
-
-                input::Event::TabletPad(_) => println!("TODO: Handle tablet pad events"),
-
-                input::Event::Gesture(_) => println!("TODO: Handle gesture events"),
-
-                input::Event::Switch(_) => println!("TODO: Handle switch events"),
-
-                _ => println!("TODO: Unhandled libinput event type"),
-            }
-        }
-    }
-}
-
-struct LibinputInterfaceImpl {
-    /// The bridge used to acquire the device.
-    bridge: cxx::SharedPtr<ffi::PlatformBridge>,
-
-    /// The list of opened devices.
-    ///
-    /// Mir expects the caller who acquired the device to keep it alive until the device
-    /// is closed. To do this, we keep it in the vector.
-    fds: Vec<cxx::UniquePtr<ffi::DeviceWrapper>>,
-}
-
-impl input::LibinputInterface for LibinputInterfaceImpl {
-    // This method is called whenever libinput needs to open a new device.
-    fn open_restricted(&mut self, path: &std::path::Path, _: i32) -> Result<io::OwnedFd, i32> {
-        let cpath = std::ffi::CString::new(path.as_os_str().as_bytes()).unwrap();
-
-        // # Safety
-        //
-        // This is an unsafe libc function.
-        let mut st: libc::stat = unsafe { std::mem::zeroed() };
-        let ret = unsafe { libc::stat(cpath.as_ptr(), &mut st) };
-        if ret != 0 {
-            return Err(ret);
-        }
-
-        let major_num = libc::major(st.st_rdev);
-        let minor_num = libc::minor(st.st_rdev);
-
-        // By keeping the device referenced in the fds vector, we ensure it stays alive
-        // until close_restricted is called.
-        let device = self
-            .bridge
-            .acquire_device(major_num as i32, minor_num as i32);
-        let fd = device.raw_fd();
-        self.fds.push(device);
-
-        // # Safety
-        //
-        // Calling from_raw_fd is unsafe.
-        let owned = unsafe { io::OwnedFd::from_raw_fd(fd) };
-        Ok(owned)
-    }
-
-    // This method is called when libinput is done with a device.
-    fn close_restricted(&mut self, fd: io::OwnedFd) {
-        let fd_raw = fd.as_raw_fd();
-        let _ = unistd::close(fd);
-        self.fds.retain(|d| d.raw_fd() != fd_raw);
-    }
-}
-
-struct LibinputLoopState {
-    libinput: input::Libinput,
-    known_devices: Vec<DeviceInfo>,
-    next_device_id: i32,
-    button_state: u32,
-    pointer_x: f32,
-    pointer_y: f32,
-    scroll_axis_x_accum: f64,
-    scroll_axis_y_accum: f64,
-    x_scroll_scale: f64,
-    y_scroll_scale: f64,
-}
-
-impl LibinputLoopState {
-    pub fn find_device_by_id<'a>(&mut self, id: i32) -> Option<&'_ mut DeviceInfo> {
-        self.known_devices.iter_mut().find(|d| d.id == id)
-    }
-}
-
-struct DeviceInfo {
-    id: i32,
-    device: input::Device,
-    input_device: cxx::SharedPtr<ffi::InputDevice>,
-    input_sink: Option<InputSinkPtr>,
-    event_builder: Option<cxx::UniquePtr<ffi::EventBuilderWrapper>>,
-}
-
-// # Safety
-//
-// The other side of the classes below are known by us to be thread-safe, so we can
-// safely send it to another thread. However, Rust has no way of knowing that, so
-// we have to assert it ourselves.
-unsafe impl Send for ffi::PlatformBridge {}
-unsafe impl Sync for ffi::PlatformBridge {}
-unsafe impl Send for ffi::InputDeviceRegistry {}
-unsafe impl Sync for ffi::InputDeviceRegistry {}
-unsafe impl Send for ffi::InputDevice {}
-unsafe impl Sync for ffi::InputDevice {}
-unsafe impl Send for ffi::InputSink {}
-unsafe impl Sync for ffi::InputSink {}
-unsafe impl Send for ffi::EventBuilder {}
-unsafe impl Sync for ffi::EventBuilder {}
-
-// Because *mut InputSink and *mut EventBuilder are raw pointers, Rust assumes
-// that they are neither Send nor Sync. However, we know that the other side of
-// the pointer is actually a C++ object that is thread-safe, so we can assert
-// that these pointers are Send and Sync. We cannot define Send and Sync on the
-// raw types to fix the issue unfortunately, so we have to wrap them in a new type
-// and define Send and Sync on that.
-pub struct InputSinkPtr(std::ptr::NonNull<ffi::InputSink>);
-impl InputSinkPtr {
-    fn handle_input(&mut self, event: &cxx::SharedPtr<ffi::MirEvent>) {
-        // # Safety
-        //
-        // Calling new_unchecked is unsafe.
-        let pinned = unsafe { std::pin::Pin::new_unchecked(self.0.as_mut()) };
-        pinned.handle_input(event);
-    }
-}
-
-// # Safety
-//
-// These needs to be unsafe because we are asserting that Send and Sync are valid on them.
-unsafe impl Send for InputSinkPtr {}
-unsafe impl Sync for InputSinkPtr {}
-
-pub struct InputDeviceInfoRs {
-    name: String,
-    unique_id: String,
-    capabilities: u32,
-    valid: bool,
-}
-
-impl InputDeviceInfoRs {
-    pub fn valid(&self) -> bool {
-        self.valid
-    }
-
-    pub fn name(&self) -> &str {
-        &self.name
-    }
-
-    pub fn unique_id(&self) -> &str {
-        &self.unique_id
-    }
-
-    pub fn capabilities(&self) -> u32 {
-        self.capabilities
-    }
-}
-
-pub struct DeviceObserverRs {
-    fd: Option<i32>,
-}
-
-impl DeviceObserverRs {
-    pub fn new() -> Self {
-        DeviceObserverRs { fd: None }
-    }
-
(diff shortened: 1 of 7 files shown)


src/platforms/evdev-rs/Cargo.toml
```diff
@@ -9,6 +9,9 @@ crate-type = ["staticlib"]
 
 [dependencies]
 cxx.workspace = true
+input = "0.9.1"
+libc = "0.2.176"
+nix = { version = "0.30.1", features = ["poll"] }
```
jhenstridge: I'd suggest putting these dependency versions in the toplevel `Cargo.toml`, then reference them here similar to `cxx`. This will ensure we share the same versions across multiple Rust projects within the source tree.

src/platforms/evdev-rs/src/enums.rs
jhenstridge, line 1: It'd be better to define these enums in the cxx ffi mod as described here:

https://cxx.rs/shared.html#extern-enums

That will generate static_asserts on the C++ side to verify that the Rust enum values match the C++ ones.

mattkae: You are probably right! Let me see if I can make them work, they did not like me earlier

dimaqq, line 1: This file being pretty large, would it make sense to publish this as a crate on crates.io?
Perhaps as is, or maybe sans Mir stuff?

dimaqq: Sorry wrong file.

src/platforms/evdev-rs/src/lib.rs
dimaqq, line 1: This file being pretty large, would it make sense to publish this as a crate on crates.io?
Perhaps as is, or maybe sans Mir stuff?

Saviq:
Exiting `miral-shell` with Ctrl+Alt+BkSp (after a client was run?) results in a crash:
```gdb
[Thread 0x7fffbffff6c0 (LWP 3504948) exited]
terminate called after throwing an instance of 'boost::wrapexcept<std::system_error>'
  what():  Failed to remove fd monitor: Bad file descriptor

Thread 50 "Mir/Input Reade" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffccfb26c0 (LWP 3504946)]
__GI_abort () at ./stdlib/abort.c:95
warning: 95     ./stdlib/abort.c: Nie ma takiego pliku ani katalogu
(gdb) bt full
#0  __GI_abort () at ./stdlib/abort.c:95
        act = {__sigaction_handler = {sa_handler = 0x0, sa_sigaction = 0x0}, sa_mask = {__val = {18446744073709551615, 0 <repeats 15 times>}}, sa_flags = 0, sa_restorer = 0x0}
#1  0x00007ffff783e05d in ??? () at /lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007ffff78541ea in ??? () at /lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff783da9c in std::terminate() () at /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff78544a1 in __cxa_throw () at /lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff686517e in boost::throw_exception<std::system_error> (e=..., loc=...) at /usr/include/boost/throw_exception.hpp:171
#6  0x00007ffff688f8cd in mir::dispatch::MultiplexingDispatchable::remove_watch (this=0x555555de3e60, fd=...)
    at /home/michal/dev/canonical/mir/src/common/dispatch/multiplexing_dispatchable.cpp:214
        lock = {_M_device = 0x7ffff68cc808, _M_owns = 8}
#7  0x00007ffff688f6cf in mir::dispatch::MultiplexingDispatchable::remove_watch
    (this=0x555555de3e60, dispatchee=std::shared_ptr<mir::dispatch::Dispatchable> (use count 1, weak count 0) = {...})
    at /home/michal/dev/canonical/mir/src/common/dispatch/multiplexing_dispatchable.cpp:198
#8  0x00007ffff6ddc692 in mir::input::DefaultInputManager::stop_platforms (this=0x555555e903f0)
    at /home/michal/dev/canonical/mir/src/server/input/default_input_manager.cpp:162
#9  0x00007ffff6ddbbc7 in operator() (__closure=0x555555ec15e8) at /home/michal/dev/canonical/mir/src/server/input/default_input_manager.cpp:92
#10 0x00007ffff6dde2f4 in std::__invoke_impl<void, mir::input::DefaultInputManager::start()::<lambda()>&>(std::__invoke_other, struct {...} &) (__f=...)
    at /usr/include/c++/15/bits/invoke.h:63
#11 0x00007ffff6dddf16 in std::__invoke_r<void, mir::input::DefaultInputManager::start()::<lambda()>&>(struct {...} &) (__fn=...) at /usr/include/c++/15/bits/invoke.h:113
#12 0x00007ffff6ddd76f in std::_Function_handler<void(), mir::input::DefaultInputManager::start()::<lambda()> >::_M_invoke(const std::_Any_data &) (__functor=...)
    at /usr/include/c++/15/bits/std_function.h:292
#13 0x00007ffff688dc7e in std::function<void()>::operator() (this=0x555555ec15e8) at /usr/include/c++/15/bits/std_function.h:593
#14 0x00007ffff68924f2 in (anonymous namespace)::dispatch_loop
    (name="Mir/Input Reader", thread_register=std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler> (use count 4, weak count 0) = {...}, dispatcher=std::shared_ptr<mir::dispatch::Dispatchable> (use count 2, weak count 0) = {...}, exception_handler=...)
    at /home/michal/dev/canonical/mir/src/common/dispatch/threaded_dispatcher.cpp:212
        running = true
        thread_registrar = {deleter = {__thread_register = std::shared_ptr<class mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler> (use count 4, weak count 0) = {get() = 0x5555557389d0}}, owner = true}
#15 0x00007ffff689b153 in std::__invoke_impl<void, void (*)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::Dispatchable>, std::function<void()> const&), std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::MultiplexingDispatchable>, std::function<void()> >
    (__f=@0x555555ec1648: 0x7ffff689224f <(anonymous namespace)::dispatch_loop(std::string const&, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::Dispatchable>, std::function<void()> const&)>) at /usr/include/c++/15/bits/invoke.h:63
#16 0x00007ffff689afc0 in std::__invoke<void (*)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::Dispatchable>, std::function<void()> const&), std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::MultiplexingDispatchable>, std::function<void()> >
    (__fn=@0x555555ec1648: 0x7ffff689224f <(anonymous namespace)::dispatch_loop(std::string const&, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::Dispatchable>, std::function<void()> const&)>) at /usr/include/c++/15/bits/invoke.h:98
#17 0x00007ffff689ae4f in std::thread::_Invoker<std::tuple<void (*)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::Dispatchable>, std::function<void()> const&), std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::MultiplexingDispatchable>, std::function<void()> > >::_M_invoke<0ul, 1ul, 2ul, 3ul, 4ul> (this=0x555555ec15e8) at /usr/include/c++/15/bits/std_thread.h:303
#18 0x00007ffff689ad74 in std::thread::_Invoker<std::tuple<void (*)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::Dispatchable>, std::function<void()> const&), std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::MultiplexingDispatchable>, std::function<void()> > >::operator() (this=0x555555ec15e8) at /usr/include/c++/15/bits/std_thread.h:310
#19 0x00007ffff689aa2c in std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (*)(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::Dispatchable>, std::function<void()> const&), std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::shared_ptr<mir::dispatch::ThreadedDispatcher::ThreadShutdownRequestHandler>, std::shared_ptr<mir::dispatch::MultiplexingDispatchable>, std::function<void()> > > >::_M_run (this=0x555555ec15e0) at /usr/include/c++/15/bits/std_thread.h:255
#20 0x00007ffff7885584 in ??? () at /lib/x86_64-linux-gnu/libstdc++.so.6
#21 0x00007ffff75c4d64 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:448
        ret = <optimized out>
        pd = <optimized out>
        out = <optimized out>
        unwind_buf = {cancel_jmp_buf = {{jmp_buf = {140736632399552, 8665876633890931886, 140736632399552, 140737488339296, 11, 140737488339559, 8665876633916097710, 8665796138938877102}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x0, 0x0}, data = {prev = 0x0, cleanup = 0x0, canceltype = 0}}}
        not_first_call = <optimized out>
#22 0x00007ffff76583bc in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```

mattkae:
I have such a funky error! It looks like the memory backing the MirEvent is gone, but it's somehow still getting freed

```
(gdb) bt
#0  0x000075bf388b3e9b in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x75bee807d250) at /usr/include/c++/15/bits/shared_ptr_base.h:345
#1  0x000075bf388b4f97 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x75beec0b10b8) at /usr/include/c++/15/bits/shared_ptr_base.h:1069
#2  0x000075bf38993110 in std::__shared_ptr<MirEvent const, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (this=0x75beec0b10b0) at /usr/include/c++/15/bits/shared_ptr_base.h:1531
#3  0x000075bf38993130 in std::shared_ptr<MirEvent const>::~shared_ptr (this=0x75beec0b10b0) at /usr/include/c++/15/bits/shared_ptr.h:175
#4  0x000075bf38b4e2da in std::pair<unsigned int, std::shared_ptr<MirEvent const> >::~pair (this=0x75beec0b10a8) at /usr/include/c++/15/bits/stl_pair.h:302
#5  0x000075bf38b4fac6 in std::destroy_at<std::pair<unsigned int, std::shared_ptr<MirEvent const> > > (__location=0x75beec0b10a8) at /usr/include/c++/15/bits/stl_construct.h:88
#6  0x000075bf38b50732 in std::_Destroy<std::pair<unsigned int, std::shared_ptr<MirEvent const> > > (__pointer=0x75beec0b10a8) at /usr/include/c++/15/bits/stl_construct.h:164
#7  0x000075bf38b50251 in std::_Destroy<std::pair<unsigned int, std::shared_ptr<MirEvent const> >*> (__first=0x75beec0b10a8, __last=0x75beec0b1258) at /usr/include/c++/15/bits/stl_construct.h:226
#8  0x000075bf38b4f78e in std::_Destroy<std::pair<unsigned int, std::shared_ptr<MirEvent const> >*, std::pair<unsigned int, std::shared_ptr<MirEvent const> > > (__first=0x75beec0b1060, __last=0x75beec0b1258)
    at /usr/include/c++/15/bits/alloc_traits.h:1045
#9  std::deque<std::pair<unsigned int, std::shared_ptr<MirEvent const> >, std::allocator<std::pair<unsigned int, std::shared_ptr<MirEvent const> > > >::_M_destroy_data_aux (this=0x75beec00f218, __first={...}, __last={...})
    at /usr/include/c++/15/bits/deque.tcc:1034
#10 0x000075bf38b4edc8 in std::deque<std::pair<unsigned int, std::shared_ptr<MirEvent const> >, std::allocator<std::pair<unsigned int, std::shared_ptr<MirEvent const> > > >::_M_destroy_data
    (this=0x75beec00f218, __first={...}, __last={...}) at /usr/include/c++/15/bits/stl_deque.h:2204
#11 0x000075bf38b4e5be in std::deque<std::pair<unsigned int, std::shared_ptr<MirEvent const> >, std::allocator<std::pair<unsigned int, std::shared_ptr<MirEvent const> > > >::~deque (this=0x75beec00f218)
    at /usr/include/c++/15/bits/stl_deque.h:1043
#12 0x000075bf38b4da14 in mir::frontend::WlClient::~WlClient (this=0x75beec00f1d0) at /home/matthew/Github/mir/src/server/frontend_wayland/wl_client.cpp:100
#13 0x000075bf38b4da60 in mir::frontend::WlClient::~WlClient (this=0x75beec00f1d0) at /home/matthew/Github/mir/src/server/frontend_wayland/wl_client.cpp:100
#14 0x000075bf38b50cb2 in std::_Sp_counted_ptr<mir::frontend::WlClient*, (__gnu_cxx::_Lock_policy)2>::_M_dispose (this=0x75beec00f1b0) at /usr/include/c++/15/bits/shared_ptr_base.h:427
#15 0x00006281531f20bd in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use (this=0x75beec00f1b0) at /usr/include/c++/15/bits/shared_ptr_base.h:174
#16 0x00006281531f05e6 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use_cold (this=0x75beec00f1b0) at /usr/include/c++/15/bits/shared_ptr_base.h:198
#17 0x00006281531efa2e in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x75beec00f1b0) at /usr/include/c++/15/bits/shared_ptr_base.h:352
#18 0x00006281531f0615 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x75beec0236b0) at /usr/include/c++/15/bits/shared_ptr_base.h:1069
#19 0x000075bf38220c70 in std::__shared_ptr<mir::wayland::Client, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (this=0x75beec0236a8) at /usr/include/c++/15/bits/shared_ptr_base.h:1531
#20 0x000075bf38220c90 in std::shared_ptr<mir::wayland::Client>::~shared_ptr (this=0x75beec0236a8) at /usr/include/c++/15/bits/shared_ptr.h:175
#21 0x000075bf382206ed in mir::wayland::Resource::~Resource (this=0x75beec0236a0, __vtt_parm=0x75bf3906f640 <VTT for mir::frontend::ShmPool+16>, __in_chrg=<optimized out>) at /home/matthew/Github/mir/src/wayland/resource.cpp:40
#22 0x000075bf38226225 in mir::wayland::ShmPool::~ShmPool (this=0x75beec0236a0, __vtt_parm=0x75bf3906f638 <VTT for mir::frontend::ShmPool+8>, __in_chrg=<optimized out>)
    at /home/matthew/Github/mir/cmake-build-debug-regular/src/wayland/wayland_wrapper.cpp:359
#23 0x000075bf38c44fea in mir::frontend::ShmPool::~ShmPool (this=0x75beec0236a0, __in_chrg=<optimized out>, __vtt_parm=<optimized out>) at /home/matthew/Github/mir/src/server/frontend_wayland/shm.h:72
#24 0x000075bf38c4502a in mir::frontend::ShmPool::~ShmPool (this=0x75beec0236a0, __in_chrg=<optimized out>, __vtt_parm=<optimized out>) at /home/matthew/Github/mir/src/server/frontend_wayland/shm.h:72
#25 0x000075bf3822cead in mir::wayland::ShmPool::Thunks::resource_destroyed_thunk (resource=0x75beec0507e0) at /home/matthew/Github/mir/cmake-build-debug-regular/src/wayland/wayland_wrapper.cpp:340
#26 0x000075bf39e0c150 in ??? () at /lib/x86_64-linux-gnu/libwayland-server.so.0
#27 0x000075bf39e0c382 in wl_client_destroy () at /lib/x86_64-linux-gnu/libwayland-server.so.0
#28 0x000075bf39e0d81d in wl_display_destroy_clients () at /lib/x86_64-linux-gnu/libwayland-server.so.0
#29 0x000075bf38b3b7f2 in mir::frontend::WaylandConnector::~WaylandConnector (this=0x628178717c40) at /home/matthew/Github/mir/src/server/frontend_wayland/wayland_connector.cpp:434
#30 0x000075bf38b38e95 in std::destroy_at<mir::frontend::WaylandConnector> (__location=0x628178717c40) at /usr/include/c++/15/bits/stl_construct.h:88
#31 0x000075bf38b38d33 in std::_Destroy<mir::frontend::WaylandConnector> (__pointer=0x628178717c40) at /usr/include/c++/15/bits/stl_construct.h:164
#32 0x000075bf38b38098 in std::allocator_traits<std::allocator<void> >::destroy<mir::frontend::WaylandConnector> (__p=0x628178717c40) at /usr/include/c++/15/bits/alloc_traits.h:819
#33 std::_Sp_counted_ptr_inplace<mir::frontend::WaylandConnector, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose (this=0x628178717c30) at /usr/include/c++/15/bits/shared_ptr_base.h:615
#34 0x000075bf388b5e99 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use (this=0x628178717c30) at /usr/include/c++/15/bits/shared_ptr_base.h:174
#35 0x000075bf388b4f02 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release_last_use_cold (this=0x628178717c30) at /usr/include/c++/15/bits/shared_ptr_base.h:198
#36 0x000075bf388b3f48 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x628178717c30) at /usr/include/c++/15/bits/shared_ptr_base.h:352
#37 0x000075bf388b4f97 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x628178758e00) at /usr/include/c++/15/bits/shared_ptr_base.h:1069
#38 0x000075bf388bd0a2 in std::__shared_ptr<mir::frontend::Connector, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (this=0x628178758df8) at /usr/include/c++/15/bits/shared_ptr_base.h:1531
#39 0x000075bf388bd0c2 in std::shared_ptr<mir::frontend::Connector>::~shared_ptr (this=0x628178758df8) at /usr/include/c++/15/bits/shared_ptr.h:175
#40 0x000075bf388be48e in mir::DisplayServer::Private::~Private (this=0x628178758da0) at /home/matthew/Github/mir/src/server/display_server.cpp:71
#41 0x000075bf388bca1c in mir::DisplayServer::~DisplayServer (this=0x7ffe68a33eb0) at /home/matthew/Github/mir/src/server/display_server.cpp:224
#42 0x000075bf388b2fa5 in mir::run_mir (config=..., init=..., terminator_=...) at /home/matthew/Github/mir/src/server/run_mir.cpp:306
#43 0x000075bf388effd8 in mir::Server::run (this=0x6281787112b0) at /home/matthew/Github/mir/src/server/server.cpp:423
#44 0x000075bf39aeebd4 in miral::MirRunner::Self::run_with (this=0x62817870f6f0, options=std::initializer_list of length 28 = {...}) at /home/matthew/Github/mir/src/miral/runner.cpp:184
#45 0x000075bf39aef4bc in miral::MirRunner::run_with (this=0x7ffe68a34200, options=std::initializer_list of length 28 = {...}) at /home/matthew/Github/mir/src/miral/runner.cpp:260
#46 0x00006281531eb46c in main (argc=3, argv=0x7ffe68a34c78) at /home/matthew/Github/mir/examples/miral-shell/shell_main.cpp:337
```

src/platforms/evdev-rs/.vscode/settings.json
```diff
@@ -0,0 +1,74 @@
+{
```
Saviq: Probably didn't want to check this in?

src/platforms/evdev-rs/src/lib.rs
Saviq: Doubt we want to publish it to crates, but maybe splitting this up a bit?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+use std::os::fd::{AsFd, AsRawFd, BorrowedFd, FromRawFd};
+use std::os::unix::ffi::OsStrExt;
+use std::os::unix::io;
+use std::sync::mpsc;
+use std::thread;
+
+/// Creates a platform on the rust side of things.
+///
+/// The `bridge` and `device_registry` parameters are provided by the C++ side of things.
+/// The C++ platform simply forwards al calls to the rust implementation.
```
Saviq:
```suggestion
/// Creates a platform on the Rust side of things.
///
/// The `bridge` and `device_registry` parameters are provided by the C++ side of things.
/// The C++ platform simply forwards all calls to the Rust implementation.
```

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// TODO: Report errors to Mir's logging facilities
```
Saviq: I think this should wait for the logging refactor we're planning.

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+            button_state: 0,
+            pointer_x: 0.0,
+            pointer_y: 0.0,
+            scroll_axis_x_accum: 0.0,
+            scroll_axis_y_accum: 0.0,
+            x_scroll_scale: 1.0,
+            y_scroll_scale: 1.0,
+        };
+
+        state.libinput.udev_assign_seat("seat0").unwrap();
```
Saviq: We don't do this anywhere else?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                        // have to queue this work onto a new thread to fire and forget it. It's not a big
+                        // deal, but it is a bit unfortunate.
+                        let input_device = state.known_devices.last().unwrap().input_device.clone();
+                        let device_registry = device_registry.clone();
+                        thread::spawn(move || unsafe {
+                            device_registry
+                                .clone()
+                                .pin_mut_unchecked()
+                                .add_device(&input_device);
+                        });
```
Saviq: Copilot here suggested to queue all this up on a single worker thread, wdyt?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                            unsafe {
+                                device_registry
+                                    .clone()
+                                    .pin_mut_unchecked()
+                                    .remove_device(&state.known_devices[index].input_device);
+                            }
+                            state.known_devices.remove(index);
+                        }
+                    }
+                    _ => todo!("Unhandled device event type"),
```
Saviq: No other event type, let's not overlook when any get added.
```suggestion
```

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                                        let old_y = state.pointer_y;
+                                        state.pointer_x = absolute_motion_event
+                                            .absolute_x_transformed(width)
+                                            as f32;
+                                        state.pointer_y = absolute_motion_event
+                                            .absolute_y_transformed(height)
+                                            as f32;
+
+                                        let movement_x = state.pointer_x - old_x;
+                                        let movement_y = state.pointer_y - old_y;
```
Saviq: Yuckie :)

We really need a dedicated `absolute_motion` action :)

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                                            0.0,
+                                            0,
+                                            0,
+                                            false,
+                                        ),
+                                    );
+                                }
+                            }
+
+                            _ => todo!("Unhandled pointer event type"),
```
Saviq: Wanna make this exhaustive instead as well?
```suggestion
```

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+
+                input::Event::Tablet(_) => todo!("Handle tablet events"),
+
+                input::Event::TabletPad(_) => todo!("Handle tablet pad events"),
+
+                input::Event::Gesture(_) => todo!("Handle gesture events"),
+
+                input::Event::Switch(_) => todo!("Handle switch events"),
+
+                _ => todo!("Unhandled libinput event type"),
```
Saviq:
```suggestion
```

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                    }
+                }
+
+                input::Event::Touch(_) => todo!("Handle touch events"),
+
+                input::Event::Tablet(_) => todo!("Handle tablet events"),
+
+                input::Event::TabletPad(_) => todo!("Handle tablet pad events"),
+
+                input::Event::Gesture(_) => todo!("Handle gesture events"),
```
Saviq: This one's actually pretty important, as two-finger scroll comes here :)

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+        let mut buf = [0u8];
+        nix::unistd::read(&rfd, &mut buf).unwrap();
+        // handle commands
+        while let Ok(cmd) = rx.try_recv() {
+            match cmd {
+                ThreadCommand::Stop => {
+                    *is_running = false;
+                }
+                ThreadCommand::GetDeviceInfo(id, tx) => {
+                    // For simplicity, just return info about the first device
```
Saviq: Legacy comment?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                        };
+                        let _ = tx.send(info);
+                    } else {
+                        let info = InputDeviceInfoRs {
+                            name: "".to_string(),
+                            unique_id: "".to_string(),
+                            capabilities: 0,
+                        };
+                        let _ = tx.send(info);
+                    }
```
Saviq: Isn't this an warning?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                                {
+                                    if accel_profile == input::AccelProfile::Adaptive {
+                                        ffi::MirPointerAcceleration::mir_pointer_acceleration_adaptive.repr
+                                    } else {
+                                        ffi::MirPointerAcceleration::mir_pointer_acceleration_none
+                                            .repr
+                                    }
+                                } else {
+                                    ffi::MirPointerAcceleration::mir_pointer_acceleration_none.repr
+                                };
```
Saviq: Is one of those branches an error?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                            }
+                            false => {
+                                let settings = ffi::PointerSettingsRs::empty();
+                                let _ = tx.send(settings);
+                            }
+                        }
+                    } else {
+                        let settings = ffi::PointerSettingsRs::empty();
+                        let _ = tx.send(settings);
+                    }
```
Saviq: Here, too - asking for pointer settings on a non-pointer device should be an error? I suppose asking about a non-existent device would be a warning instead?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+                                .config_accel_set_profile(accel_profile)
+                                .unwrap();
+                            device_info
+                                .device
+                                .config_accel_set_speed(settings.cursor_acceleration_bias as f64)
+                                .unwrap();
+                            state.x_scroll_scale = settings.horizontal_scroll_scale;
+                            state.y_scroll_scale = settings.vertical_scroll_scale;
+                        }
+                    }
```
Saviq: Else error?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+    id: i32,
+    device: input::Device,
+    input_device: cxx::SharedPtr<ffi::InputDevice>,
+    input_sink: Option<InputSinkPtr>,
+    event_builder: Option<cxx::UniquePtr<ffi::EventBuilderWrapper>>,
+}
+
+// The other side of the PlatformBridgeC is known by us to be thread-safe, so we can
+// safely send it to another thread. However, Rust has no way of knowing that, so
+// we have to assert it ourselves.
```
Saviq: Can probably use a single comment for all the below?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+unsafe impl Sync for ffi::InputSink {}
+
+// The other side of the EventBuilder is known by us to be thread-safe, so we can
+// safely send it to another thread. However, Rust has no way of knowing that, so
+// we have to assert it ourselves.
+unsafe impl Send for ffi::EventBuilder {}
+unsafe impl Sync for ffi::EventBuilder {}
+
+// This is another hack.
+//
```
Saviq: It is what it is… maybe `~rust` folk have another idea?
```suggestion
```

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+    }
+}
+
+#[cxx::bridge(namespace = "mir::input::evdev_rs")]
+mod ffi {
+    // Note(mattkae): The device capability enum MUST be in its expanded
+    // form instead of using a bit shift (e.g. 1 << 3) because cxx
+    // currently does not support bit shifts in enums. This will result
+    // in the error: "enums with non-integer literal discriminants are
+    // not supported yet".
```
Saviq: If anyone tries, the error will be clear enough, no? :)
```suggestion
```

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1404 @@
...
+
+    #[repr(i32)]
+    #[derive(Debug, Copy, Clone, PartialEq, Eq)]
+    enum MirKeyboardAction {
+        mir_keyboard_action_up,
+        mir_keyboard_action_down,
+        mir_keyboard_action_repeat,
+        mir_keyboard_action_modifiers,
+        mir_keyboard_actions,
+    }
```
Saviq: No way to bring these from `.h`?

How do we ensure these are equivalent to those…?

Saviq (commented):
Overall, this seems fine, and works :)

---

Please document each use of `unsafe` [explaining how it is, in fact, safe](https://std-dev-guide.rust-lang.org/policy/safety-comments.html).

---

It feels like the different scroll handling could be combined?

---

There's a lot of places where you silently fall back to a default value - at least some of those should be warnings / errors (missing device, device of wrong capabilities etc.).

src/platforms/evdev-rs/src/lib.rs
mattkae: I will mention in the TODO

src/platforms/evdev-rs/src/lib.rs
mattkae: Added a "TODO" so that we know to refactor this when we do multi-seat

src/platforms/evdev-rs/src/lib.rs
mattkae: I think that is more overhead than it's worth for now

src/platforms/evdev-rs/src/lib.rs
mattkae: Rust doesn't want us to remove this, even though the switch is exhaustive, which is unfortunate :/

src/platforms/evdev-rs/src/lib.rs
mattkae: Yeah it felt weird putting this code into the platform

src/platforms/evdev-rs/src/lib.rs
mattkae: Same as above :(

src/platforms/evdev-rs/src/lib.rs
mattkae: Same as above :(

src/platforms/evdev-rs/src/lib.rs
mattkae: Yeah, I want to do this separately though just to leave off the corners that could be broken right now. Do you mind if we tackle this stuff next? It will be a smaller PR then

src/platforms/evdev-rs/src/lib.rs
mattkae: I will print the error in rust, but it should not be fatal or anything

src/platforms/evdev-rs/src/lib.rs
mattkae: There is no way to bring them in!

By putting them in the `mod ffi` namespace, cxx rust does some magic to make sure that the values in rust match those in C++, so we at least get that safety. It won't build otherwise.

src/platforms/evdev-rs/src/lib.rs
RAOF: There's another TODO here: what to do if `udev_assign_seat()` returns `Err`; the current behaviour will end up calling `abort()`, which is probably not what we want to do here.

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1443 @@
...
+                events: libc::POLLIN,
+                revents: 0,
+            },
+        ];
+
+        let mut is_running = true;
+        while is_running {
+            // # Safety
+            //
+            // Calling libc::poll is unsafe.
```
RAOF: A better safety comment would be:
```suggestion
            //
            // `libc::poll` requires that the first parameter be a pointer to an array of `struct pollfd`, with the second parameter being at most the number of valid elements of that array.
            // `fds` is an array of valid `pollfd`s of length `fds.len()`, so this is safe by construction
```
This describes *why* the `unsafe` call is valid.

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1443 @@
...
+        fn removed(self: &mut DeviceObserverRs);
+
+        /// # Safety
+        ///
+        /// This is unsafe because it is receiving raw pointers from C++ as parameters.
+        unsafe fn start(
+            self: &mut InputDeviceRs,
+            input_sink: *mut InputSink,
+            event_builder: *mut EventBuilder,
+        );
```
RAOF, line 160: Why isn't this

```suggestion
        fn start(
            self: &mut InputDeviceRs,
            input_sink: &mut InputSink,
            event_builder: &mut EventBuilder,
        );
```
It's not clear to me why we're using raw pointers (and hence `unsafe`) here, rather than references?

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1443 @@
...
+        fn add_device(
+            self: Pin<&mut InputDeviceRegistry>,
+            device: &SharedPtr<InputDevice>,
+        ) -> WeakPtr<Device>;
+
+        #[namespace = "mir::input"]
+        fn remove_device(self: Pin<&mut InputDeviceRegistry>, device: &SharedPtr<InputDevice>);
+
+        #[namespace = "mir::input"]
+        fn handle_input(self: Pin<&mut InputSink>, event: &SharedPtr<MirEvent>);
```
RAOF: Why is this `Pin<&mut InputSink>`? Why can't it just be `self: &mut InputSink`?

mattkae:
```
mutable reference to opaque C++ type requires a pin -- use `self: Pin<&mut InputSink>`rust-analyzer[macro-error](https://rust-analyzer.github.io/book/diagnostics.html#macro-error)
```

src/platforms/evdev-rs/src/lib.rs
mattkae: This is a direct translate of the C++ code, which uses raw pointers.

RAOF: Hm. I think the C++ code should be reworked to use references? (There doesn't seem to be any reason why this would be hard)

mattkae: Another issue seems to be that we're storing these pointers in the `InputDevice` as well. Perhaps these should be `shared_ptr` and the device holds them as `weak_ptr`.

src/platforms/evdev-rs/src/lib.rs
RAOF: Ah, right. Yeah, fair enough, you can't let the Rust tode `mem::replace` the reference.

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1457 @@
...
+
+            if (fds[1].revents & libc::POLLIN) != 0 {
+                PlatformRs::process_thread_events(
+                    rfd.try_clone().unwrap(),
+                    &rx,
+                    &mut state,
+                    &bridge_locked,
+                    &mut is_running,
+                );
+            }
```
RAOF: This would be cleaner (and avoid `unsafe`) using the `nix::poll::Poll` interface that we have already available.

RAOF (approved):
I think we can probably make this substantially more friendly and idiomatic with some extra infrastructure (probably including implementing a `futures::Executor` on top of `Dispatchable`) and with some tweaks to the C++ side.

I think that's fine to do in follow-up work, though; I'm happy to merge this and then incrementally improve it. (Assuming the crash on shutdown has been resolved?)

RAOF:
Also need to fix the input platform selection tests:
```
[ RUN ] InputPlatformProbe.stub_platform_not_picked_up_by_default
[2025-10-27 18:22:57.632508] <information> mirserver: Selected input driver: mir:evdev-input-rs (version: 2.23.0)
../../../tests/unit-tests/input/test_input_platform_probing.cpp:148: Failure
Value of: platform
Expected: is a or derived from a N3mir5input5evdev8PlatformE
Actual: (ptr = 0x55a75f3f5100, value = 8-byte object <A8-BB BF-94 73-7F 00-00>)

[ FAILED ] InputPlatformProbe.stub_platform_not_picked_up_by_default (25 ms)
[ RUN ] InputPlatformProbe.x11_input_platform_not_used_when_vt_specified
[2025-10-27 18:22:57.653514] <information> mirserver: Selected input driver: mir:evdev-input-rs (version: 2.23.0)
../../../tests/unit-tests/input/test_input_platform_probing.cpp:167: Failure
Value of: platform
Expected: is a or derived from a N3mir5input5evdev8PlatformE
Actual: (ptr = 0x55a75f3ea9b0, value = 8-byte object <A8-BB BF-94 73-7F 00-00>)

[ FAILED ] InputPlatformProbe.x11_input_platform_not_used_when_vt_specified (20 ms)
```
(Looks like these tests should be checking the selected platform is *not* the platform we don't expect, rather than checking that the selected platform is the evdev platform).

src/platforms/evdev-rs/src/lib.rs
mattkae: I no longer need to do this anyway because the dispatchable polls for us

include/platform/mir/input/input_device.h
tarek-y-ismail, line 1: Unrelated

src/platforms/evdev-rs/platform.cpp
```diff
@@ -15,56 +15,207 @@
...
-class DispatchableStub : public md::Dispatchable
+template <typename Impl>
+class DeviceObserverWrapper : public miers::DeviceObserverWithFd
 {
 public:
-    mir::Fd watch_fd() const override
+    explicit DeviceObserverWrapper(rust::Box<Impl>&& box)
+        : box(std::move(box)) {}
+
+    void activated(mir::Fd&& device_fd) override
```
tarek-y-ismail, line 45: Is TICS correct here?

src/platforms/evdev-rs/platform_bridge.h
```diff
@@ -0,0 +1,142 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_INPUT_EVDEV_RS_PLATFORM_BRIDGE_H
+#define MIR_INPUT_EVDEV_RS_PLATFORM_BRIDGE_H
+
+#include <memory>
```
tarek-y-ismail, line 24: Should be at the very bottom of the include list

src/platforms/evdev-rs/platform_bridge.cpp
```diff
@@ -0,0 +1,141 @@
...
+        int32_t action,
+        uint32_t scancode
+    ) const
+{
+    return event_builder->key_event(
+        has_time
+            ? std::chrono::microseconds(time_microseconds)
+            : std::optional<EventBuilder::Timestamp>(std::nullopt),
+        static_cast<MirKeyboardAction>(action),
+        xkb_keysym_t{0},
```
tarek-y-ismail, line 112: No need to pass a keysym?

Saviq:
@mattkae trying to package on Fedora I'm getting some linking errors, care to have a look?

https://github.com/canonical/mir/actions/runs/19575622413/job/56060022149?pr=4453#step:8:3265

Should be able to reproduce locally with `spread-mir-ci.spread -reuse -debug lxd:fedora-43:spread/build/fedora:amd64`.

Saviq:
@jhenstridge we seem to be missing some dependency that I couldn't spot immediately… thougths?

`
cd /<<PKGBUILDDIR>>/build-amd64/src/platforms/evdev-rs && ccache /usr/bin/c++ -DEGL_NO_X11 -DLTTNG_UST_HAVE_SDT_INTEGRATION -DMIR_LOG_COMPONENT_FALLBACK=\"evdev-input-rs\" -DMIR_USE_APPARMOR -DMIR_VERSION_MAJOR=2 -DMIR_VERSION_MICRO=0 -DMIR_VERSION_MINOR=23 -D_FILE_OFFSET_BITS=64 -D_GNU_SOURCE -I/<<PKGBUILDDIR>>/include/platform -I/<<PKGBUILDDIR>>/include/core -I/<<PKGBUILDDIR>>/src/include/platform -I/<<PKGBUILDDIR>>/src/platforms/evdev-rs -I/<<PKGBUILDDIR>>/src/include/server -I/<<PKGBUILDDIR>>/include/common -I/<<PKGBUILDDIR>>/src/include/common -isystem /usr/include/libdrm -g -O2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffile-prefix-map=/<<PKGBUILDDIR>>=. -flto=auto -ffat-lto-objects -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -fdebug-prefix-map=/<<PKGBUILDDIR>>=/usr/src/mir-2.23.0 -Wdate-time -D_FORTIFY_SOURCE=3 -pthread -g -Wall -fno-strict-aliasing -pedantic -Wnon-virtual-dtor -Wextra -fPIC -Werror -Wno-error=maybe-uninitialized -Wno-error=array-bounds -Wno-mismatched-tags -Wno-psabi -std=c++23 -MD -MT src/platforms/evdev-rs/CMakeFiles/mirplatforminputevdevrs_rust-cxxbridge.dir/__/__/__/target/cxxbridge/mir_platforms_evdev_rs/src/lib.rs.cc.o -MF CMakeFiles/mirplatforminputevdevrs_rust-cxxbridge.dir/__/__/__/target/cxxbridge/mir_platforms_evdev_rs/src/lib.rs.cc.o.d -o CMakeFiles/mirplatforminputevdevrs_rust-cxxbridge.dir/__/__/__/target/cxxbridge/mir_platforms_evdev_rs/src/lib.rs.cc.o -c /<<PKGBUILDDIR>>/build-amd64/target/cxxbridge/mir_platforms_evdev_rs/src/lib.rs.cc
cc1plus: fatal error: /<<PKGBUILDDIR>>/build-amd64/target/cxxbridge/mir_platforms_evdev_rs/src/lib.rs.cc: No such file or directory
`

https://github.com/canonical/mir/actions/runs/19636890722/job/56229722137#step:8:8420

mattkae:
> @mattkae trying to package on Fedora I'm getting some linking errors, care to have a look?
> 
> https://github.com/canonical/mir/actions/runs/19575622413/job/56060022149?pr=4453#step:8:3265
> 
> Should be able to reproduce locally with `spread-mir-ci.spread -reuse -debug lxd:fedora-43:spread/build/fedora:amd64`.

I'll look at this today

tarek-y-ismail (commented):
Quick initial review on everything but `lib.rs`, will come back to it later.

src/platforms/evdev-rs/platform.cpp
AlanGriffiths: Obviously TICS is correct here

src/platforms/evdev-rs/platform_bridge.cpp
mattkae: Nope, this API is weird. It takes a keysym, but the keysym is resolved later on anyway. The regular evdev platform does the same thing

cmake/RustLibrary.cmake
```diff
@@ -27,18 +27,30 @@ function(add_rust_cxx_library target)
   set(cxxbridge_source "${cxxbridge_include_dir}/${arg_CRATE}/${arg_CXX_BRIDGE_SOURCE_FILE}.cc")
   set(crate_staticlib "${rust_binary_dir}/lib${arg_CRATE}.a")
 
+  # Build the Rust crate and copy the generated cxxbridge files to expected location
   add_custom_command(
     OUTPUT ${cxxbridge_header} ${cxxbridge_source} ${crate_staticlib}
     COMMAND ${CARGO_EXECUTABLE} build ${cargo_release_flag} --target-dir ${rust_target_dir} -p ${arg_CRATE}
+    COMMAND ${CMAKE_COMMAND} -E make_directory "${cxxbridge_include_dir}/${arg_CRATE}/src"
+    COMMAND sh -c "cp -n ${rust_binary_dir}/build/${arg_CRATE}-*/out/cxxbridge/include/${arg_CRATE}/${arg_CXX_BRIDGE_SOURCE_FILE}.h ${cxxbridge_header}"
+    COMMAND sh -c "cp -n ${rust_binary_dir}/build/${arg_CRATE}-*/out/cxxbridge/sources/${arg_CRATE}/${arg_CXX_BRIDGE_SOURCE_FILE}.cc ${cxxbridge_source}"
```
Saviq:
```
       -n, --no-clobber
              (deprecated) do not overwrite an existing file and do not fail

              (overrides a -u or

              previous -i option). See also --update; equivalent to --update=none.
```

Saviq: That said, this surfaces something this:
```
Running `CARGO=/usr/lib/rust-1.88/bin/cargo CARGO_CRATE_NAME=mir_platforms_evdev_rs CARGO_MANIFEST_DIR=/build/reproducible-path/mir-2.23.0/src/platforms/evdev-rs CARGO_MANIFEST_PATH=/build/reproducible-path/mir-2.23.0/src/platforms/evdev-rs/Cargo.toml CARGO_PKG_AUTHORS='' CARGO_PKG_DESCRIPTION='' CARGO_PKG_HOMEPAGE='' CARGO_PKG_LICENSE='' CARGO_PKG_LICENSE_FILE='' CARGO_PKG_NAME=mir_platforms_evdev_rs CARGO_PKG_README='' CARGO_PKG_REPOSITORY='' CARGO_PKG_RUST_VERSION=1.75.0 CARGO_PKG_VERSION=0.1.0 CARGO_PKG_VERSION_MAJOR=0 CARGO_PKG_VERSION_MINOR=1 CARGO_PKG_VERSION_PATCH=0 CARGO_PKG_VERSION_PRE='' CARGO_PRIMARY_PACKAGE=1 CARGO_SBOM_PATH='' LD_LIBRARY_PATH=/build/reproducible-path/mir-2.23.0/build-amd64/target/debug/deps OUT_DIR=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/build/mir_platforms_evdev_rs-e70387cbe9ad0b9b/out rustc --crate-name mir_platforms_evdev_rs --edition=2021 src/platforms/evdev-rs/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type staticlib --emit=dep-info,link -C embed-bitcode=no -C debuginfo=2 --check-cfg 'cfg(docsrs,test)' --check-cfg 'cfg(feature, values())' -C metadata=6920815b1a9f8517 -C extra-filename=-65f3d02f7ffa1bde --out-dir /build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/deps --target x86_64-unknown-linux-gnu -C incremental=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/incremental -L dependency=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/deps -L dependency=/build/reproducible-path/mir-2.23.0/build-amd64/target/debug/deps --extern cxx=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/deps/libcxx-bf70210464da97c3.rlib --extern input=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/deps/libinput-b3e2f9187400831c.rlib --extern libc=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/deps/liblibc-fde8e99a40b53c5c.rlib --extern nix=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/deps/libnix-6f49ce0f9ded23cb.rlib -C debuginfo=2 -C strip=none --cap-lints warn -C linker=x86_64-linux-gnu-gcc -C link-arg=-Wl,-Bsymbolic-functions -C link-arg=-flto=auto -C link-arg=-ffat-lto-objects -C link-arg=-Wl,-z,relro -C link-arg=-Wl,-z,now --remap-path-prefix /build/reproducible-path/mir-2.23.0=/usr/share/cargo/registry/mir-2.23.0 --remap-path-prefix /build/reproducible-path/mir-2.23.0/debian/cargo_registry=/usr/share/cargo/registry -Cforce-frame-pointers=yes -L native=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/build/cxx-f8c32018ca4f5302/out -L native=/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/build/link-cplusplus-728db9356ca2a576/out -L native=/usr/lib/x86_64-linux-gnu`
Finished `dev` profile [unoptimized + debuginfo] target(s) in 22.85s
cd /build/reproducible-path/mir-2.23.0 && /usr/bin/cmake -E make_directory /build/reproducible-path/mir-2.23.0/build-amd64/target/cxxbridge/mir_platforms_evdev_rs/src
cd /build/reproducible-path/mir-2.23.0 && sh -c cp\ -n\ /build/reproducible-path/mir-2.23.0/build-amd64/target/debug/build/mir_platforms_evdev_rs-*/out/cxxbridge/include/mir_platforms_evdev_rs/src/lib.rs.h\ /build/reproducible-path/mir-2.23.0/build-amd64/target/cxxbridge/mir_platforms_evdev_rs/src/lib.rs.h
cp: warning: behavior of -n is non-portable and may change in future; use --update=none instead
cp: cannot stat '/build/reproducible-path/mir-2.23.0/build-amd64/target/debug/build/mir_platforms_evdev_rs-*/out/cxxbridge/include/mir_platforms_evdev_rs/src/lib.rs.h': No such file or directory
```

`OUT_DIR` looks wrong, here's where the files end up:
```
tree /build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/build/mir_platforms_evdev_rs-e70387cbe9ad0b9b/out
/build/reproducible-path/mir-2.23.0/build-amd64/target/x86_64-unknown-linux-gnu/debug/build/mir_platforms_evdev_rs-e70387cbe9ad0b9b/out
└── cxxbridge
    ├── crate
    │   └── mir_platforms_evdev_rs -> ../../../../../../../../../src/platforms/evdev-rs
    ├── include
    │   ├── mir_platforms_evdev_rs
    │   │   └── src
    │   │       ├── lib.rs -> lib.rs.h
    │   │       └── lib.rs.h
    │   └── rust
    │       └── cxx.h -> /build/reproducible-path/mir-2.23.0/debian/cargo_registry/cxx-1.0.189/include/cxx.h
    └── sources
        └── mir_platforms_evdev_rs
            └── src
                └── lib.rs.cc

11 directories, 4 files
```

Ideally we'd _tell_ Cargo where to put it?

Saviq: > Ideally we'd tell Cargo where to put it?

Ah, it's dpkg tells it. But we can query instead. Will fix downstream.

Saviq:
```suggestion
```

What happens without this?

In my packaging endeavour I got (without the `-n`):
```
cp: '.../build-Debug/target/debug/build/mir_platforms_evdev_rs-f9d1192186e601a6/out/cxxbridge/include/mir_platforms_evdev_rs/src/lib.rs.h' and '...build-Debug/target/cxxbridge/mir_platforms_evdev_rs/src/lib.rs.h' are the same file
```

😕

mattkae: Yeah that's why I added the `-n`. It was _sometimes_ the same file

cmake/RustLibrary.cmake
```diff
@@ -27,18 +27,28 @@ function(add_rust_cxx_library target)
   set(cxxbridge_source "${cxxbridge_include_dir}/${arg_CRATE}/${arg_CXX_BRIDGE_SOURCE_FILE}.cc")
   set(crate_staticlib "${rust_binary_dir}/lib${arg_CRATE}.a")
 
+  # Build the Rust crate and copy the generated cxxbridge files to expected location
```
Saviq:
```suggestion
```

cmake/RustLibrary.cmake
```diff
@@ -27,18 +27,28 @@ function(add_rust_cxx_library target)
...
   add_custom_command(
     OUTPUT ${cxxbridge_header} ${cxxbridge_source} ${crate_staticlib}
     COMMAND ${CARGO_EXECUTABLE} build ${cargo_release_flag} --target-dir ${rust_target_dir} -p ${arg_CRATE}
+    COMMAND ${CMAKE_COMMAND} -E make_directory "${cxxbridge_include_dir}/${arg_CRATE}/src"
```
Saviq:
```suggestion
```

cmake/RustLibrary.cmake
```diff
@@ -27,18 +27,28 @@ function(add_rust_cxx_library target)
...
     DEPENDS ${arg_DEPENDS}
-    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+    COMMENT "Building Rust crate ${arg_CRATE} and copying cxxbridge files")
```
Saviq:
```suggestion
    COMMENT "Building Rust crate ${arg_CRATE}")
```

Saviq (changes requested):

src/platforms/evdev-rs/CMakeLists.txt
```diff
@@ -41,7 +44,15 @@ target_link_libraries(mirplatforminputevdevrs
...
   CRATE mir_platforms_evdev_rs
-  DEPENDS Cargo.toml src/lib.rs)
+  DEPENDS Cargo.toml src/lib.rs
+  LIBRARIES PkgConfig::UDEV PkgConfig::LIBINPUT
+  INCLUDES
+    ${PROJECT_SOURCE_DIR}/src/platforms/evdev-rs
+    ${PROJECT_SOURCE_DIR}/src/include/platform
+    ${PROJECT_SOURCE_DIR}/src/include/server
+    ${PROJECT_SOURCE_DIR}/include/common
+    ${PROJECT_SOURCE_DIR}/src/include/common
```
AlanGriffiths: This looks wrong in several ways:

1. a platform shouldn't be depending on `/src/include/*`
2. cmake is managing dependencies - so I would expect adding `LIBRARIES ... mirplatform` to suffice

(In practice, the latter may not apply to `add_rust_cxx_library()`, but it should)

src/platforms/evdev-rs/platform.h
```diff
@@ -14,23 +14,42 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "mir/console_services.h"
```
AlanGriffiths:
```suggestion
#include <mir/console_services.h>
```

src/platforms/evdev-rs/platform_bridge.h
```diff
@@ -0,0 +1,142 @@
...
+    virtual ~PlatformBridgeC() = default;
+    std::unique_ptr<DeviceBridgeC> acquire_device(int major, int minor) const;
+    std::shared_ptr<InputDevice> create_input_device(int device_id) const;
+    std::unique_ptr<EventBuilderWrapper> create_event_builder_wrapper(EventBuilder* event_builder) const;
+    std::unique_ptr<RectangleC> bounding_rectangle(InputSink const&) const;
+
+private:
+    Platform* platform;
+    std::shared_ptr<mir::ConsoleServices> console;
+};
```
AlanGriffiths: This class looks weird: it has CopyAssign enabled, yet it has a virtual destructor but no other virtual methods. Why?

src/platforms/evdev-rs/platform_bridge.h
```diff
@@ -0,0 +1,142 @@
...
+class DeviceBridgeC
+{
+public:
+    DeviceBridgeC(std::unique_ptr<Device> device, int fd);
+    int raw_fd() const;
+
+private:
+    std::unique_ptr<Device> device;
+    int fd;
+};
```
AlanGriffiths: Is this intentionally a move-only type?

src/platforms/evdev-rs/platform_bridge.h
```diff
@@ -0,0 +1,142 @@
...
+        uint64_t time_microseconds,
+        int32_t action,
+        uint32_t scancode
+    ) const;
+
+private:
+    EventBuilder* event_builder;
+};
+
+class RectangleC
```
AlanGriffiths: Is this "C" suffix a new naming convention? Or part of the translation layer mechanics?

AlanGriffiths (changes requested):
Not yet dived into the Rust code, but a first few thoughts...

src/platforms/evdev-rs/CMakeLists.txt
mattkae: 1. Yup, you're very right about the includes. I was being greedy there.
2. Let me see if we can be clever about resolving from the library (we probably can somehow)

src/platforms/evdev-rs/platform_bridge.h
mattkae: Yeah, it should not be virtual. I think that was just leftover

src/platforms/evdev-rs/platform_bridge.h
mattkae: The device is only intended to be stored as a unique_ptr by one unique owner, so yes

src/platforms/evdev-rs/platform_bridge.h
mattkae: It's my convention, but maybe a poor one. I was trying to describe where the type originates from (although "Cpp" is probably better in this case). Do you have any other suggestion? Or is this a non-problem to begin with?

src/platforms/evdev-rs/CMakeLists.txt
mattkae: Is this any better:

```cmake
add_rust_cxx_library(mirplatforminputevdevrs_rust
  CRATE mir_platforms_evdev_rs
  DEPENDS Cargo.toml src/lib.rs
  LIBRARIES PkgConfig::UDEV PkgConfig::LIBINPUT
  INCLUDES
    ${PROJECT_SOURCE_DIR}/src/platforms/evdev-rs
    $<TARGET_PROPERTY:mircommon,INTERFACE_INCLUDE_DIRECTORIES>
)
```

src/platforms/evdev-rs/platform_bridge.h
AlanGriffiths: I haven't yet seen the need for such a distinction, but I've not delved far enough into the code to be confident there isn't one

src/platforms/evdev-rs/CMakeLists.txt
Saviq: Wouldn't you pass `mircommon` to `LIBRARIES` above and extract the header dirs from that inside `RustLibrary.cmake`?

mattkae: I believe that I fixed it!

src/platforms/evdev-rs/platform_bridge.h
mattkae: Kk, let me know. We can always change that later as we see fit

cmake/RustLibrary.cmake
```diff
@@ -31,14 +31,30 @@ function(add_rust_cxx_library target)
...
+  if(arg_INCLUDES)
+    foreach(inc_dir IN LISTS arg_INCLUDES)
+      target_include_directories(${target}-cxxbridge PRIVATE "${inc_dir}")
+    endforeach()
+  endif()
+
+  if(arg_LIBRARIES)
+    foreach(library ${arg_LIBRARIES})
+      target_include_directories(${target}-cxxbridge PRIVATE
+              "$<TARGET_PROPERTY:${library},INTERFACE_INCLUDE_DIRECTORIES>}"
```
Saviq: Isn't there an extra `}` here?

mattkae: True!

src/platforms/evdev-rs/platform_bridge.cpp
```diff
@@ -0,0 +1,143 @@
...
+std::unique_ptr<miers::DeviceBridgeC> miers::PlatformBridgeC::acquire_device(int major, int minor) const
+{
+    mir::log_info("Acquiring device: %d.%d", major, minor);
+    auto observer = platform->create_device_observer();
+    DeviceObserverWithFd* raw_observer = observer.get();
+    auto future = console->acquire_device(major, minor, std::move(observer));
+    future.wait();
+    if (!raw_observer->raw_fd())
+        throw std::runtime_error("Failed to acquire device");
+    return std::make_unique<DeviceBridgeC>(future.get(), raw_observer->raw_fd().value());
```
AlanGriffiths: TICS is understandably unhappy as there's no way for the compiler to know that `raw_observer->raw_fd()` returns the same result.
```suggestion
    if (auto const raw_fd = raw_observer->raw_fd(); !raw_fd)
        throw std::runtime_error("Failed to acquire device");
    else
        return std::make_unique<DeviceBridgeC>(future.get(), raw_fd.value());
```

src/platforms/evdev-rs/CMakeLists.txt
```diff
@@ -1,7 +1,14 @@
+include_directories(
+  ${PROJECT_SOURCE_DIR}/include/platform
+  ${PROJECT_SOURCE_DIR}/include/core
+  ${PROJECT_SOURCE_DIR}/src/include/platform
+)
+
```
AlanGriffiths: I've been cleaning out this sort of legacy approach elsewhere. Is it needed here? Why?

src/platforms/evdev-rs/platform_bridge.cpp
```diff
@@ -0,0 +1,144 @@
...
+    int32_t axis_source,
+    float precise_x,
+    int discrete_x,
+    int value120_x,
+    bool scroll_stop_x,
+    float precise_y,
+    int discrete_y,
+    int value120_y,
+    bool scroll_stop_y) const
+{
```
AlanGriffiths: There's a parameter list from hell!

Are we constrained to do it this way? Or should we pass a struct with these members?

src/platforms/evdev-rs/platform_bridge.h
AlanGriffiths: If you're using this convention, maybe a comment explaining it?

src/platforms/evdev-rs/platform_bridge.h
```diff
@@ -0,0 +1,141 @@
...
+    int32_t height() const
+    {
+        return height_;
+    }
+
+private:
+    int32_t const x_;
+    int32_t const y_;
+    int32_t const width_;
+    int32_t const height_;
```
AlanGriffiths:
```suggestion
    int32_t const x;
    int32_t const y;
    int32_t const width;
    int32_t const height;
```

src/platforms/evdev-rs/platform_bridge.h
```diff
@@ -0,0 +1,141 @@
...
+///
+/// The rust code may use this bridge to request things from the system.
+class PlatformBridgeC
+{
+public:
+    PlatformBridgeC(Platform* platform, std::shared_ptr<mir::ConsoleServices> const& console);
+    std::unique_ptr<DeviceBridgeC> acquire_device(int major, int minor) const;
+    std::shared_ptr<InputDevice> create_input_device(int device_id) const;
+    std::unique_ptr<EventBuilderWrapper> create_event_builder_wrapper(EventBuilder* event_builder) const;
+    std::unique_ptr<RectangleC> bounding_rectangle(InputSink const&) const;
```
AlanGriffiths: Why don't `EventBuilderWrapper` and `InputDevice` have the "C" suffix?

src/server/input/default_input_manager.cpp
```diff
@@ -154,12 +154,14 @@ void mi::DefaultInputManager::stop()
 void mi::DefaultInputManager::start_platforms()
 {
     platform->start();
-    multiplexer->add_watch(platform->dispatchable());
+    if (auto const dispatchable = platform->dispatchable())
+        multiplexer->add_watch(dispatchable);
```
AlanGriffiths: Hmm. You've quietly changed the contract for `Platform::dispatchable()` (that it may now return an empty pointer). I agree that a NullObject would be hard for this, but probably needs documenting in `<mir/input/platform.h>`

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1300 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// TODO: Report errors to Mir's logging facilities. We should do this following Mir's logging refactor.
+// TODO: Need to set up reporting events when received from libinput (report->received_event_from_kernel)
+// TODO: Implement continue after_config and pause_for_config
```
AlanGriffiths, line 19: I know this isn't blocking  - but these seem pretty significant block to real use

src/platforms/evdev-rs/src/lib.rs
```diff
@@ -1,25 +1,1300 @@
...
+
+    pub unsafe fn libinput_fd(&mut self) -> i32 {
+        self.state
+            .as_mut()
+            .unwrap()
+            .lock()
+            .unwrap()
+            .libinput
+            .as_raw_fd()
+    }
```
AlanGriffiths: `unsafe` and `unwrap` :firecracker: :grinning:

src/server/input/default_input_manager.cpp
mattkae: We actually no longer need this, so I can revert.

src/platforms/evdev-rs/src/lib.rs
mattkae: To be honest, only the reporting is significant here. We never implement "continue" and "pause"

src/platforms/evdev-rs/platform_bridge.h
mattkae: You win, I remove the "C" suffix ;)

src/platforms/evdev-rs/platform_bridge.cpp
mattkae: Let's try and pass a struct!

src/platforms/evdev-rs/platform_bridge.h
mattkae: I removed it. The `Rs` suffix to denote a type coming form Rust is the useful one, not the `C` suffix

src/platforms/evdev-rs/src/lib.rs
AlanGriffiths: I agree, there seems a good case for splitting things out. E.g. the ffi related thunks are pretty much independent of the implementation logic.

AlanGriffiths: @copilot would this file be a good candidate for splitting out different concerns?

Saviq (approved):

PR #4341, opened 2025-10-07
Don't try to run the hosted Wayland platform unless there's a current desktop

AlanGriffiths:
If there's a Wayland session open on VT2 and this is run on VT4 the Wayland platform will hang waiting for the "host" to respond.

## What's new?

We detect we're not running on a host desktop

## How to test

Leave a Wayland session open on VT2 and on VT4 (as the same user) run `mir-smoke-test-runner`


Saviq (approved):
Sure.

PR #4342, opened 2025-10-09
Locate pointer graphical feedback with cursor scale

tarek-y-ismail:
## What's new?

- Adds `LocatePointer`, a class where shell authors can schedule a callback to look up the cursor position with optional delay. The callback can also be cancelled.
- Adds `CursorScale::temporarily_scale` to temporarily change the cursor's scale (with animation!)
- Hooks up `LocatePointer::on_locate_pointer` to temporarily scale the cursor when left or right click are held for 1 second.
- Adds MirAL tests for `LocatePointer`

## How to test
- Run `miral-app`
- Hold down left or right control
- Watch as the cursor scales up, then back down

## Checklist
- [x] SYMBOLS!
- [x] Tests added and pass
- [x] Rebase 

https://github.com/user-attachments/assets/524b6d48-129d-4bf3-aeb5-c06c345402d5

### DIFF
--- a/include/miral/miral/locate_pointer.h
+++ b/include/miral/miral/locate_pointer.h
@@ -94,5 +94,3 @@ private:
 }
 
 #endif // MIRAL_LOCATE_POINTER_H
-
-


src/miral/cursor_scale.cpp
```diff
@@ -15,39 +15,119 @@
...
+        s->animation_stop_alarm->reschedule_in(animation_duration);
     }
 
     float const default_scale;
-    std::atomic<float> scale_;
+
+    struct State
+    {
+        float scale;
+        float previous_scale;
```
tarek-y-ismail: I think we can drop `previous_scale`

debian/libmiral7.symbols
tarek-y-ismail, line 1: Outdated

tarek-y-ismail, line 1: Also missing `CursorScale::scale_temporarily`

include/miral/miral/locate_pointer.h
```diff
@@ -0,0 +1,71 @@
...
+    static auto disabled() -> LocatePointer;
+
+    void operator()(mir::Server& server);
+
+    // TODO live config
+
+    LocatePointer& delay(std::chrono::milliseconds delay);
+    LocatePointer& on_locate_pointer(std::function<void(float x, float y)>&&);
+    LocatePointer& on_enabled(std::function<void()>&&);
+    LocatePointer& on_disabled(std::function<void()>&&);
```
tarek-y-ismail: Need documentation

src/miral/symbols.map
tarek-y-ismail, line 1: Out of date

src/miral/symbols.map
```diff
@@ -555,6 +555,18 @@ global:
...
+    miral::LocatePointer::delay*;
+    miral::LocatePointer::disable*;
+    miral::LocatePointer::enable*;
+    miral::LocatePointer::on_disabled*;
+    miral::LocatePointer::on_enabled*;
+    miral::LocatePointer::on_locate_pointer*;
+    miral::LocatePointer::operator*;
+    miral::LocatePointer::schedule_request*;
+    miral::MouseKeysConfig::MouseKeysConfig*;
+    miral::MouseKeysConfig::enabled*;
```
tarek-y-ismail: Why are these here?

src/server/shell/basic_accessibility_manager.h
tarek-y-ismail, line 1: Unrelated

tests/miral/test_locate_pointer.cpp
```diff
@@ -0,0 +1,237 @@
...
+    mir::test::Signal locate_pointer_invoked;
+
+    std::weak_ptr<mir::input::CompositeEventFilter> composite_event_filter;
+    std::weak_ptr<mir::input::InputDeviceRegistry> input_device_registry;
+    std::weak_ptr<mir::input::InputDeviceHub> input_device_hub;
+};
+
+struct TestDifferentDelays : public TestLocatePointer, public WithParamInterface<std::pair<std::chrono::milliseconds, bool>>{};
+
+TEST_P(TestDifferentDelays, responding_to_ctrl_down_with_different_delays)
```
tarek-y-ismail: Incorrect name

tests/miral/test_locate_pointer.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    locate_pointer.disable();
+
+    mir::test::Signal enable_called;
+    locate_pointer.on_enabled([&enable_called] { enable_called.raise(); });
+    locate_pointer.enable();
+
+    enable_called.wait();
+    EXPECT_TRUE(enable_called.raised());
+}
```
tarek-y-ismail: `on_{enabled,disabled}` should be removed, and these tests along with them

tarek-y-ismail:
Something worth mentioning: If the cursor starts at scale 6x, and locate pointer is called on it, it actually shrinks down as locate pointer is coded to scale it to 5x and back to the original scale. Do we `CursorScale::scale_temporarily` to instead take a scale multiplier? Or can we trust shell authors to manage this on their own?

Saviq:
> Do we `CursorScale::scale_temporarily` to instead take a scale multiplier? Or can we trust shell authors to manage this on their own?

I think that'd be better - more so if the scale is actually changed within the duration?

examples/miral-shell/shell_main.cpp
```diff
@@ -39,7 +40,10 @@
 #include <miral/hover_click.h>
 #include <miral/sticky_keys.h>
 #include <miral/magnifier.h>
+#include <miral/cursor_scale.h>
 
+#define MIR_LOG_COMPONENT "miral-shell"
+#include <mir/log.h>
```
AlanGriffiths: 1. I don't see this being used
2. Consider defining `MIR_LOG_COMPONENT_FALLBACK` in CMakeLists.txt to apply to the binary

src/miral/locate_pointer.cpp
```diff
@@ -0,0 +1,232 @@
...
+                disable();
+        });
+}
+
+miral::LocatePointer::LocatePointer(miral::live_config::Store& config_store) :
+    self{std::make_shared<Self>(false)}
+{
+    config_store.add_bool_attribute(
+        {"locate_pointer", "enable"},
+        "Enable or disable slow keys",
```
AlanGriffiths, line 133: Please provide a `preset` default

src/miral/locate_pointer.cpp
```diff
@@ -0,0 +1,232 @@
...
+                    this->enable();
+                else
+                    this->disable();
+            }
+        });
+
+    config_store.add_int_attribute(
+        {"locate_pointer", "delay"},
+        "The delay between scheduling a request and the on_locate_pointer callback being called.",
+        [this](live_config::Key const& key, std::optional<int> val)
```
AlanGriffiths, line 150: Please provide a `preset` default

include/miral/miral/locate_pointer.h
```diff
@@ -0,0 +1,89 @@
...
+    void operator()(mir::Server& server);
+
+    /// Construct a `LocatePointer` instance with access to a live config
+    /// store.
+    ///
+    /// Available options:
+    ///     - {locate_pointer, enable}: Enable or disable locate pointer.
+    ///     - {locate_pointer, delay}: The delay between scheduling a request
+    ///     and the #on_locate_pointer callback being called.
+    explicit LocatePointer(miral::live_config::Store& config_store);
```
AlanGriffiths, line 53: What are the semantics of `enable` & `disable`? I'm guessing that calls to `schedule_request()` & `cancel_request()` only have an effect when enabled. But the docs should probably say this.

include/miral/miral/locate_pointer.h
```diff
@@ -0,0 +1,89 @@
...
+    /// The delay between scheduling a request and the #on_locate_pointer
+    /// callback being called.
+    /// \note The default delay is 500 milliseconds
+    LocatePointer& delay(std::chrono::milliseconds delay);
+
+    /// Configures the callback that's invoked after #schedule_request is
+    /// called by #delay milliseconds.
+    /// Useful for providing feedback to users.
+    LocatePointer& on_locate_pointer(std::function<void(float x, float y)>&&);
+
```
AlanGriffiths, line 64: What do the parameters to the callback represent? Should that be `PointF pointer_position`?

examples/miral-shell/shell_main.cpp
tarek-y-ismail: It was just for initial testing. Thanks for pointing it out.

tarek-y-ismail:
> So, based solely on the example code, the main functionality appears to be an alarm (that may be cancelled): maybe an Alarm is a better API to offer. I.e. the shell code could be...

I agree with your conclusion. Do recall that this is a half-step between implementing locate pointer without any graphical feedback and implementing it with the feedback shown in the spec. At the moment, the type definitions of alarms are not exposed in miral, nor is there an easy way to create alarms. I'm not sure how much work will be needed to do so.

Once surfaces can be created and drawn into from miral, the implementation of this class will change significantly. The x and y coordinates passed to the callback will be used to position the surface being drawn into, and the remainder of the class will manage the tracking of the cursor and the lifetime of the surface.

include/miral/miral/locate_pointer.h
```diff
@@ -17,6 +17,7 @@
 #ifndef MIRAL_LOCATE_POINTER_H
 #define MIRAL_LOCATE_POINTER_H
 
+#include "mir/geometry/forward.h"
```
AlanGriffiths: Don't give the compiler extra work

```suggestion
#include <mir/geometry/forward.h>
```

PR #4343, opened 2025-10-10
Add support for the ext_foreign_toplevel_list_v1 extension

jhenstridge:
Related: MIRENG-1499

## What's new?

This PR implements the [ext_foreign_toplevel_list_v1](https://wayland.app/protocols/ext-foreign-toplevel-list-v1) extension. The extension is effectively a subset of the existing [wlr_foreign_toplevel_manager_v1](https://wayland.app/protocols/wlr-foreign-toplevel-management-unstable-v1). The changes are:

1. The toplevel_list object is not implicitly destroyed after it sends the finished event.
2. There are no requests or events related to manipulating toplevels. The only information provided about the toplevels is a unique identifier (new), title, and app_id.
3. The toplevel identifier is required to be stable across different connections, and remain the same while the toplevel is mapped.

To manage identifiers, we maintain a map from `std::weak_ptr<Surface>` to string identifiers shared by all connections. It is possible the map may end up with some stale entries if there are no clients currently using the extension. So we scan through and remove expired keys whenever a client binds the global object.

## How to test

1. Download and build [lswt](https://git.sr.ht/~leon_plickat/lswt).
2. Run `./bin/miral-app --add-wayland-extensions ext_foreign_toplevel_list_v1`, and start some applications.
3. Run `WAYLAND_DISPLAY=wayland-1 lswt -w` to list the toplevels, and watch as toplevels are opened and closed.
4. Run a second instance, and note that the identifiers reported by both instances of lswt are the same.

## Checklist

- [x] Tests added and pass (https://github.com/canonical/wlcs/pull/377)

### DIFF
--- a/include/miral/miral/wayland_extensions.h
+++ b/include/miral/miral/wayland_extensions.h
@@ -131,6 +131,7 @@ public:
     /// Allows a client to list toplevels of all clients.
     /// Similar to zwlr_foreign_toplevel_manager_v1, but without the ability to control toplevels.
     /// Could allow a client to extract information about other programs the user is running.
+    /// \remark Since MirAL 5.6
     static char const* const ext_foreign_toplevel_list_v1;
 
     /// Allows clients to act as a virtual keyboard, useful for on-screen keyboards.
--- a/src/server/frontend_wayland/foreign_toplevel_list_v1.cpp
+++ b/src/server/frontend_wayland/foreign_toplevel_list_v1.cpp
@@ -138,7 +138,7 @@ private:
 }
 
 /// Informs a client about toplevels from itself and other clients
-/// The Wayland objects it creates for each toplevel can be used to aquire information and control that toplevel
+/// The Wayland objects it creates for each toplevel can be used to acquire information
 /// Useful for task bars and app switchers
 class ExtForeignToplevelListV1Global
     : public wayland::ExtForeignToplevelListV1::Global


src/server/frontend_wayland/foreign_toplevel_list_v1.cpp
```diff
@@ -0,0 +1,560 @@
...
+
+namespace
+{
+class ForeignSurfaceObserver;
+
+// A helper class to keep track of toplevel IDs
+class ForeignToplevelIdentifierMap
+{
+public:
+    std::string toplevel_id(std::shared_ptr<scene::Surface> const& surface);
```
RAOF:
```suggestion
    auto toplevel_id(std::shared_ptr<scene::Surface> const& surface) -> std::string;
```

src/server/frontend_wayland/foreign_toplevel_list_v1.cpp
```diff
@@ -0,0 +1,560 @@
...
+    {
+        BOOST_THROW_EXCEPTION(std::runtime_error(
+            "ext_foreign_toplevel_handle_v1@" +
+            std::to_string(wl_resource_get_id(resource)) +
+            " is not a mir::frontend::ExtForeignToplevelHandleV1"));
+    }
+    return *handle;
+}
+
+std::shared_ptr<ms::Surface> mf::ExtForeignToplevelHandleV1::surface() const
```
RAOF:
```suggestion
auto mf::ExtForeignToplevelHandleV1::surface() const -> std::shared_ptr<ms::Surface>
```

src/server/frontend_wayland/foreign_toplevel_list_v1.h
```diff
@@ -0,0 +1,70 @@
...
+    static auto from(wl_resource* resource) -> ExtForeignToplevelHandleV1*;
+    static auto from_or_throw(wl_resource* resource) -> ExtForeignToplevelHandleV1&;
+    std::shared_ptr<scene::Surface> surface() const;
+
+    /// Sends the .closed event and makes this surface inert
+    void should_close();
+
+private:
+    std::weak_ptr<scene::Surface> weak_surface;
+};
```
RAOF, line 57: It seems like this class could not exist? `from`, `from_or_throw`, and `surface()` are all dead code, `should_close()` only calls `wayland::ExtForeignToplevelHandleV1::send_close_event()`, and the constructor only calls `manager.send_toplevel_event()`.

What am I missing here?

src/server/frontend_wayland/foreign_toplevel_list_v1.cpp
```diff
@@ -0,0 +1,560 @@
...
+    auto const iter = surface_observers.find(surface);
+    if (iter == surface_observers.end())
+    {
+        log_error(
+            "Can not remove ForeignSurfaceObserver: surface %p not in observers map",
+            static_cast<void*>(surface.get()));
+    }
+    else
+    {
+        iter->second->cease_and_desist();
```
RAOF: Aha!

I was wondering why `ForeignSurfaceObserver` contained a mutex - all the observations take place on the `WaylandExecutor`, so there should not need to be any locking. But *this* call gets made on an arbitrary thread.

But that's a problem, because this call will eventually make Wayland calls, and they have to be done on the Wayland thread.

So I think this should be reworked so that `ForeignSurfaceObserver` all happens on the Wayland thread, and then it doesn't need a mutex at all.

src/server/frontend_wayland/foreign_toplevel_list_v1.h
jhenstridge: These methods are mostly here as scaffolding to implement [`ext_foreign_toplevel_image_capture_source_manager_v1:request:create_source`](https://wayland.app/protocols/ext-image-capture-source-v1#ext_foreign_toplevel_image_capture_source_manager_v1:request:create_source).

We need to know what surface the toplevel handle represents so the eventual capture session knows what it should be copying.

src/server/frontend_wayland/foreign_toplevel_list_v1.cpp
```diff
@@ -0,0 +1,560 @@
...
+    : mw::ExtForeignToplevelHandleV1{manager},
+      weak_surface{surface}
+{
+    manager.send_toplevel_event(resource);
+}
+
+auto mf::ExtForeignToplevelHandleV1::from(
+    wl_resource* resource) -> ExtForeignToplevelHandleV1*
+{
+    return dynamic_cast<ExtForeignToplevelHandleV1*>(wayland::ExtForeignToplevelListV1::from(resource));
```
RAOF: This is always going to fail, right? `ExtForeignToplevelHandleV1` is not a subclass of `ExtForeignToplevelListV1`.

src/server/frontend_wayland/foreign_toplevel_list_v1.h
RAOF: Ah! Maybe that should be in the follow-up PR, when it's needed?

src/server/frontend_wayland/foreign_toplevel_list_v1.cpp
jhenstridge: I've changed the SceneObserver methods to delegate their work to tasks in the wayland_executor thread. The tasks take a weak pointer to the observer, to handle the case where tasks might run after the observer has been destroyed.

I've removed the mutexes, since all the data access is now being serialised on that thread.

include/miral/miral/wayland_extensions.h
```diff
@@ -128,6 +128,11 @@ class WaylandExtensions
     /// Could allow a client to extract information about other programs the user is running
     static char const* const zwlr_foreign_toplevel_manager_v1;
 
+    /// Allows a client to list toplevels of all clients.
+    /// Similar to zwlr_foreign_toplevel_manager_v1, but without the ability to control toplevels.
+    /// Could allow a client to extract information about other programs the user is running.
+    static char const* const ext_foreign_toplevel_list_v1;
```
AlanGriffiths, line 135:
```suggestion
    /// Could allow a client to extract information about other programs the user is running.
    /// \remark Since MirAL 5.6
    static char const* const ext_foreign_toplevel_list_v1;
```

src/server/frontend_wayland/foreign_toplevel_list_v1.cpp
```diff
@@ -0,0 +1,567 @@
...
+    /// a toplevel window or if the client has destroyed the handle.
+    wayland::Weak<ExtForeignToplevelHandleV1> handle;
+
+    std::shared_ptr<DesktopFileManager> const desktop_file_manager;
+    std::shared_ptr<ForeignToplevelIdentifierMap> const id_map;
+};
+}
+
+/// Informs a client about toplevels from itself and other clients
+/// The Wayland objects it creates for each toplevel can be used to aquire information and control that toplevel
```
AlanGriffiths: The "and control` clause should have been dropped after copying this comment from ForeignToplevelManagerV1Global. :wink:

Issue #4147, opened 2025-08-21
Check for `WL_bind_wayland_display` when probing eglstream-kms support


Saviq:
Nvidia drivers 550+ dropped the `WL_bind_wayland_display` extension, resulting in a late:

```
ERROR: ./src/platform/graphics/egl_extensions.cpp(111): Throw in function mir::graphics::EGLExtensions::WaylandExtensions::WaylandExtensions(EGLDisplay)
Dynamic exception type: boost::wrapexcept<std::runtime_error>
std::exception::what: EGL display doesn't support EGL_WL_bind_wayland_display

ERROR: Dynamic exception type: std::_Nested_exception<std::runtime_error>
std::exception::what: Failed to bind Wayland EGL display
```

We should, instead, check for the extension when probing, so we never select the `eglstream-kms` platform when it's not going to work.

PR #4352, opened 2025-10-13
Add eglstream-kms to mir-smoke-test-runner script

### LINKED ISSUE #4351: Smoke test runner ignores `MIR_SERVER_PLATFORM_DISPLAY_LIBS`
> > ```
> > I: The following clients failed to execute successfully:
> > I:     eglinfo/virtual
> > I:     eglinfo/gbm-kms
> > I:     eglinfo/atomic-kms
> > I:     /usr/bin/mir_demo_client_wayland_egl_spinner/atomic-kms
> > I: Smoke testing complete with returncode -1, platforms tested: virtual gbm-kms atomic-kms
> > ```
> 
> Ah! There's been a change to `mir-smoke-test-runner` - it now tries to enumerate platforms and is ignoring `MIR_SERVER_PLATFORM_DISPLAY_LIBS` (see last line of output). It lacks logic for `mir:eglstream` - maybe you can submit a fix?
> 

_Originally posted by @AlanGriffiths in https://github.com/canonical/mir/issues/4321#issuecomment-3397154987_

tarek-y-ismail:
Closes #4351

## What's new?

- Runs tests for `eglstream-kms` in the smoke test runner.

## How to test

- Run `mir-smoke-test-runner`
- Observe as it tests eglstream-kms

### DIFF
--- a/tools/mir-smoke-test-runner.sh
+++ b/tools/mir-smoke-test-runner.sh
@@ -67,6 +67,12 @@ then
 
   echo Test atomic-kms platform
   run_tests atomic-kms
+
+  if readlink -f /sys/class/drm/*/device/driver | grep -q nvidia$
+  then
+    echo Test eglstream-kms platform
+    run_tests eglstream-kms
+  fi
 fi
 
 date --utc --iso-8601=seconds | xargs echo "[timestamp] End time :"


tools/mir-smoke-test-runner.sh
```diff
@@ -67,6 +67,9 @@ then
 
   echo Test atomic-kms platform
   run_tests atomic-kms
+
+  echo Test eglstrem-kms platform
+  run_tests eglstrem-kms
```
AlanGriffiths: "eglstrem"?

Also, can this be guarded so that it isn't attempted on systems where we don't expect `eglstream-kms` to work?

Saviq: I'm afraid at this point this would require [checking for an EGL extension](https://github.com/canonical/mir/issues/4147)… but we could, at least, check `readlink -f /sys/class/drm/*/device/driver | grep -q nvidia$`

PR #4363, opened 2025-10-16
Refactored the wayland extension management such that a policy can be set per-extension and processes originating from the server can always use the required extensions

mattkae:
Alternative to https://github.com/canonical/mir/pull/4361

## What's new?
- Refactored the extension management system such that each interface has its own policy, specified via `set_wayland_extension_policy`.
- Deprecated the methods that set the extension filter globally and override the enabled extensions
- In `miral::WaylandExtensions`, we now conditionally enable extensions on a per-extension basis
- The server enables all extensions by default now
- Made it so that sessions originating from the shell process will always get approved by the policy, even if they fail the interface check
- Added the application switcher to `miral-shell` as an example

## How to test
1. Run `miral-shell`
2. Do app switcher stuff and see that it works
3. Try and launch `swaybg` and note that it will not work because it is not conditionally enabled

## Checklist
- [x] Tests added and pass

### DIFF
--- a/src/include/server/mir/server.h
+++ b/src/include/server/mir/server.h
@@ -491,10 +491,6 @@ public:
             wl_display*,
             std::function<void(std::function<void()>&& work)> const&)> builder);
 
-    [[deprecated("Prefer using set_wayland_extension_policy() to set a policy on a per-interface bassis.")]]
-    void set_wayland_extension_filter(
-        std::function<bool(std::shared_ptr<scene::Session> const&, char const*)> const& extension_filter);
-
     /// Get the name of the Wayland endpoint (if any) usable as a $WAYLAND_DISPLAY value
     auto wayland_display() const -> std::optional<std::string>;
 
@@ -502,10 +498,6 @@ public:
     auto x11_display() const -> std::optional<std::string>;
 
     auto get_activation_token() const -> std::string;
-
-    /// Overrides the standard set of Wayland extensions (mir::frontend::get_standard_extensions()) with a new list
-    [[deprecated("Prefer using set_wayland_extension_policy() to set a policy on a per-interface bassis.")]]
-    void set_enabled_wayland_extensions(std::vector<std::string> const& extensions);
 
     /// Set a policy for the provided wayland extension.
     ///
--- a/src/include/server/mir/server_configuration.h
+++ b/src/include/server/mir/server_configuration.h
@@ -114,12 +114,6 @@ public:
 
     using WaylandProtocolExtensionFilter = std::function<bool(std::shared_ptr<scene::Session> const&, char const*)>;
 
-    [[deprecated("Prefer using set_wayland_extension_policy() to set a policy on a per-interface bassis.")]]
-    virtual void set_wayland_extension_filter(WaylandProtocolExtensionFilter const& extension_filter) = 0;
-
-    [[deprecated("Prefer using set_wayland_extension_policy() to set a policy on a per-interface bassis.")]]
-    virtual void set_enabled_wayland_extensions(std::vector<std::string> const& extensions) = 0;
-
     /// Set a policy for the provided wayland extension.
     ///
     /// Only clients whose \p policy returns `true` will be allowed to use the interface
--- a/src/server/frontend_wayland/wayland_default_configuration.cpp
+++ b/src/server/frontend_wayland/wayland_default_configuration.cpp
@@ -422,18 +422,6 @@ void mir::DefaultServerConfiguration::add_wayland_extension(
     wayland_extension_hooks.push_back({name, builder});
 }
 
-void mir::DefaultServerConfiguration::set_wayland_extension_filter(WaylandProtocolExtensionFilter const& extension_filter)
-{
-    wayland_extension_filter = extension_filter;
-}
-
-void mir::DefaultServerConfiguration::set_enabled_wayland_extensions(std::vector<std::string> const& extensions)
-{
-    wayland_extension_policy_map.clear();
-    for (auto const& extension : extensions)
-        wayland_extension_policy_map[extension] = [](auto const&, const char*) { return true; };
-}
-
 void mir::DefaultServerConfiguration::set_wayland_extension_policy(
     std::string const& interface_name,
     WaylandProtocolExtensionFilter const& policy)
--- a/src/include/server/mir/default_server_configuration.h
+++ b/src/include/server/mir/default_server_configuration.h
@@ -200,8 +200,6 @@ public:
             wl_display*,
             std::function<void(std::function<void()>&& work)> const&)> builder) override;
 
-    void set_wayland_extension_filter(WaylandProtocolExtensionFilter const& extension_filter) override;
-    void set_enabled_wayland_extensions(std::vector<std::string> const& extensions) override;
     void set_wayland_extension_policy(
         std::string const& interface_name,
         WaylandProtocolExtensionFilter const& policy) override;
--- a/src/server/server.cpp
+++ b/src/server/server.cpp
@@ -529,23 +529,6 @@ void mir::Server::add_wayland_extension(
     }
 }
 
-void mir::Server::set_wayland_extension_filter(
-    std::function<bool(std::shared_ptr<scene::Session> const&, char const*)> const& extension_filter)
-{
-    if (auto const config = self->server_config)
-    {
-        config->set_wayland_extension_filter(extension_filter);
-    }
-}
-
-void mir::Server::set_enabled_wayland_extensions(std::vector<std::string> const& extensions)
-{
-    if (auto const config = self->server_config)
-    {
-        config->set_enabled_wayland_extensions(extensions);
-    }
-}
-
 void mir::Server::set_wayland_extension_policy(
     std::string const& interface_name,
     std::function<bool(std::shared_ptr<scene::Session> const&, char const*)> const& policy)
--- a/src/server/symbols.map
+++ b/src/server/symbols.map
@@ -14,9 +14,7 @@ global:
     mir::DefaultServerConfiguration::DefaultServerConfiguration*;
     mir::DefaultServerConfiguration::add_wayland_extension*;
     mir::DefaultServerConfiguration::default_reports*;
-    mir::DefaultServerConfiguration::set_enabled_wayland_extensions*;
     mir::DefaultServerConfiguration::set_the_decoration_strategy*;
-    mir::DefaultServerConfiguration::set_wayland_extension_filter*;
     mir::DefaultServerConfiguration::set_wayland_extension_policy*;
     mir::DefaultServerConfiguration::the_accessibility_manager*;
     mir::DefaultServerConfiguration::the_buffer_allocator*;
@@ -191,11 +189,9 @@ global:
     mir::Server::set_command_line*;
     mir::Server::set_command_line_handler*;
     mir::Server::set_config_filename*;
-    mir::Server::set_enabled_wayland_extensions*;
     mir::Server::set_exception_handler*;
     mir::Server::set_terminator*;
     mir::Server::set_the_decoration_strategy*;
-    mir::Server::set_wayland_extension_filter*;
     mir::Server::set_wayland_extension_policy*;
     mir::Server::stop*;
     mir::Server::supported_pixel_formats*;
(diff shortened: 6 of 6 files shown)


examples/miral-shell/shell_main.cpp
```diff
@@ -334,10 +336,60 @@ int main(int argc, char const* argv[])
             return false;
     };
 
+    // The following filter triggers the application selector internal application
+    // on "Ctrl + Tab" and "Ctrl + Shift + Tab". Note that this does NOT use
+    // "Alt + Tab" as that is claimed by the MinimalWindowManager for the time being.
```
AlanGriffiths: I don't think we want to duplicate this into miral-shell as well as mir_demo_server - in its current state it is not an example we want anyone to adopt.

examples/miral-shell/shell_main.cpp
```diff
@@ -334,10 +336,60 @@ int main(int argc, char const* argv[])
...
+        return false;
+    };
+
     return runner.run_with(
         {
             CursorTheme{"default:DMZ-White"},
-            WaylandExtensions{},
+            WaylandExtensions{}
+                .conditionally_enable(WaylandExtensions::zwlr_layer_shell_v1, [](WaylandExtensions::EnableInfo const&) { return false; })
+                .conditionally_enable(WaylandExtensions::zwlr_foreign_toplevel_manager_v1, [](WaylandExtensions::EnableInfo const&) { return false; }),
```
AlanGriffiths: AIUI this is, in effect, test code? I don't think test code belongs in our examples.

src/server/frontend_wayland/wayland_connector.cpp
```diff
@@ -593,6 +596,9 @@ bool mf::WaylandConnector::wl_display_global_filter_func(wl_client const* client
 {
     auto const* const interface = wl_global_get_interface(global);
     auto const session = get_session(const_cast<wl_client*>(client));
+    if (session->process_id() == getpid())
+        return true;
```
AlanGriffiths: [non blocking] This is sensible but probably over broad - I think we should be allowing "privileged" protocols, but things like wl_shell and xdg_shell_v6 are questionable.

examples/miral-shell/shell_main.cpp
```diff
@@ -334,6 +336,54 @@ int main(int argc, char const* argv[])
             return false;
     };
 
+    // The following filter triggers the application selector internal application
+    // on "Ctrl + Tab" and "Ctrl + Shift + Tab". Note that this does NOT use
+    // "Alt + Tab" as that is claimed by the MinimalWindowManager for the time being.
```
AlanGriffiths: I still don't think we should be adding this in this form to the miral-shell example. Ctrl-Tab is a commonly used application shortcut (e.g. switching tabs) and showing this in an example is a Bad Idea.

src/include/server/mir/server_configuration.h
```diff
@@ -113,9 +113,25 @@ class ServerConfiguration
...
+    ///
+    /// Only clients whose \p policy returns `true` will be allowed to use the interface
+    /// given by \p interface_name.
+    ///
+    /// \param interface_name the interface to apply the policy on
+    /// \param policy the policy predicate with the session and interface name provided
+    ///               as arguments.
+    virtual void set_wayland_extension_policy(
+        std::string const& interface_name,
+        std::function<bool(std::shared_ptr<scene::Session> const&, const char*)> const& policy) = 0;
```
AlanGriffiths, line 127: This is an ABI break... we need to bump the so number and rearrange the symbols.map

src/server/default_server_configuration.cpp
```diff
@@ -51,14 +52,35 @@ namespace ms = mir::scene;
...
 }
 
 mir::DefaultServerConfiguration::DefaultServerConfiguration(std::shared_ptr<mo::Configuration> const& configuration_options) :
     configuration_options(configuration_options),
-    enabled_wayland_extensions(frontend::get_standard_extensions())
+    wayland_extension_policy_map(get_default_wayland_extension_policy_map()),
+    wayland_extension_filter([this](auto const& session, auto const& interface_name)
+    {
+        if (wayland_extension_policy_map.contains(interface_name))
+            return wayland_extension_policy_map[interface_name](session, interface_name)
```
AlanGriffiths: Searching the map for the same key twice seems inefficient.

src/include/server/mir/server_configuration.h
mattkae: Should I just remove the deprecated methods then too?

examples/miral-shell/shell_main.cpp
mattkae: Whoops! I can make it use Alt-Tab instead

CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 
 set(MIR_VERSION_MAJOR 2)
-set(MIR_VERSION_MINOR 23)
+set(MIR_VERSION_MINOR 24)
```
AlanGriffiths: Please drop (not revert) this change. Our current release process is to bump on the release branch. (Yes, that should probably change and Saviq has a proposal to change it but...)

mattkae: Kk XD

debian/control
```diff
@@ -67,7 +67,7 @@ Vcs-Git: https://github.com/canonical/mir
 
 #TODO: Packaging infrastructure for better dependency generation,
 #      ala pkg-xorg's xviddriver:Provides and ABI detection.
-Package: libmirserver66
+Package: libmirserver67
```
AlanGriffiths: We already bumped to 66 since the 2.23 release

src/include/server/mir/server_configuration.h
AlanGriffiths: Yes. (Sorry, I think I may have prompted you to double bump the soname)

src/include/server/mir/server.h
```diff
@@ -495,7 +496,20 @@ class Server
     auto get_activation_token() const -> std::string;
 
     /// Overrides the standard set of Wayland extensions (mir::frontend::get_standard_extensions()) with a new list
+    [[deprecated("Prefer using set_wayland_extension_policy() to set a policy on a per-interface bassis.")]]
     void set_enabled_wayland_extensions(std::vector<std::string> const& extensions);
+
+    /// Set a policy for the provided wayland extension.
+    ///
+    /// Only clients whose \p policy returns `true` will be allowed to use the interface
+    /// given by \p interface_name.
```
AlanGriffiths: This phrasing seems strange: clients don't have their own policies.

```suggestion
    /// Only clients for which the \p policy returns `true` will be offered the global \p interface_name.
```

src/server/frontend_wayland/wayland_default_configuration.cpp
```diff
@@ -363,9 +365,9 @@ std::shared_ptr<mf::Connector>
             auto options = the_options();
             bool const arw_socket = options->is_set(options::arw_server_socket_opt);
 
-            auto wayland_extensions = std::set<std::string>{
-                enabled_wayland_extensions.begin(),
-                enabled_wayland_extensions.end()};
+            std::set<std::string> wayland_extensions;
+            for (const auto& interface_name: wayland_extension_policy_map | std::views::keys)
+                wayland_extensions.insert(interface_name);
```
AlanGriffiths: Isn't there a range constructor for `std::set<>`?

AlanGriffiths (changes requested):

CMakeLists.txt
AlanGriffiths: You didn't drop the change, you reverted it

src/server/frontend_wayland/wayland_default_configuration.cpp
AlanGriffiths: :disappointed:

src/miral/wayland_extensions.cpp
```diff
@@ -371,20 +389,16 @@ struct miral::WaylandExtensions::Self
...
             {
                 std::optional<bool> user_pref;
                 if (manually_enabled_extensions.contains(protocol))
-                {
                     user_pref = true;
-                }
                 else if (manually_disabled_extensions.contains(protocol))
-                {
                     user_pref = false;
-                }
```
AlanGriffiths: Why are we dropping these braces?

src/server/frontend_wayland/wayland_default_configuration.cpp
```diff
@@ -365,9 +365,10 @@ std::shared_ptr<mf::Connector>
             auto options = the_options();
             bool const arw_socket = options->is_set(options::arw_server_socket_opt);
 
-            std::set<std::string> wayland_extensions;
-            for (const auto& interface_name: wayland_extension_policy_map | std::views::keys)
-                wayland_extensions.insert(interface_name);
+            std::set<std::string> wayland_extensions(
+                std::ranges::begin(wayland_extension_policy_map | std::views::keys),
+                std::ranges::end(wayland_extension_policy_map | std::views::keys)
+            );
```
AlanGriffiths: These iterators don't refer to the same range do they?

src/server/CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ install(DIRECTORY
     ${CMAKE_SOURCE_DIR}/src/include/server/mir DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/mirserver-internal"
 )
 
-set(MIRSERVER_ABI 67) # Be sure to increment MIR_VERSION_MINOR at the same time
+set(MIRSERVER_ABI 66) # Be sure to increment MIR_VERSION_MINOR at the same time
```
AlanGriffiths, line 183: Please drop the change, not revert it

tests/miral/wayland_extensions.cpp
```diff
@@ -138,12 +137,71 @@ struct ClientGlobalEnumerator
...
+    }
+
+    char line[256];
+    while (fgets(line, sizeof(line), stream))
+    {
+        size_t const len = strlen(line);
+        if (len && line[len - 1] == '\n')
+            line[len - 1] = '\0';
+
+        interfaces.emplace_back(line);
```
AlanGriffiths: Can't we simply pass two iterators to the string constructor?

```suggestion
        interfaces.emplace_back(line, line + len-1);
```

examples/miral-shell/shell_main.cpp
```diff
@@ -334,6 +336,65 @@ int main(int argc, char const* argv[])
...
+            {
+                application_switcher.confirm();
+                is_running = false;
+                return true;
+            }
+        }
+
+        return false;
+    };
+
```
AlanGriffiths, line 397: This is a lot of boilerplate for the downstream to vendor. Can't we provide a helper? (Maybe one that has options for the keys and modifiers)

mattkae: For a follow up PR? Seems out of scope for this one

CMakeLists.txt
mattkae: Rewrote the history

src/include/server/mir/server.h
```diff
@@ -491,6 +491,7 @@ class Server
             wl_display*,
             std::function<void(std::function<void()>&& work)> const&)> builder);
 
+    [[deprecated("Prefer using set_wayland_extension_policy() to set a policy on a per-interface bassis.")]]
     void set_wayland_extension_filter(
         std::function<bool(std::shared_ptr<scene::Session> const&, char const*)> const& extension_filter);
```
AlanGriffiths: Any reason for keeping the deprecated functions? We break the mirserver ABI almost every time anyway

Saviq: That's avoiding API breaks, though, so a no-change rebuild is _possible_?

AlanGriffiths: > That's avoiding API breaks, though, so a no-change rebuild is _possible_?

Do you really think anyone is using this API?

Saviq: Not a whole lot, no…

https://github.com/search?q=set_wayland_extension_filter&type=code

mattkae: Kk, I will remove the deprecated functions too

PR #4366, opened 2025-10-20
Rework bounce keys to not use an alarm

tarek-y-ismail:
Didn't really need an alarm in the first place..

Related: #3934 

## What's new?
- Reworks key rejection in bounce keys to not use alarms, instead, it just stores the timestamp of the last press and compares it to the incoming press.

## How to test
- Enable bounce keys in `miral-shell` by creating an enabled bounce keys object and passing it to the runner
- Run `miral-shell`
- Open a terminal
- Try pressing the same button quickly multiple times
- Observe that only the first press will be registered
- Slow your presses down
- Observe that as you slow your presses down, keys will begin registering

Alternating between different keys should work, and key repeat should not be affected.


src/miral/bounce_keys.cpp
```diff
@@ -51,13 +50,12 @@ struct miral::BounceKeys::Self
...
+            mi::XkbSymkey last_key;
+            mir::time::Timestamp last_key_time;
         };
 
-        BounceKeysFilter(
-            std::shared_ptr<mir::MainLoop> const& main_loop, std::shared_ptr<mir::Synchronised<Config>> config) :
-            config{std::move(config)},
-            last_key_clearer{main_loop->create_alarm([this] { this->runtime_state.lock()->last_key.reset(); })}
+        BounceKeysFilter(std::shared_ptr<mir::Synchronised<Config>> config) :
+            config{std::move(config)}
```
mattkae:
```suggestion
        explicit BounceKeysFilter(std::shared_ptr<mir::Synchronised<Config>> const& config) :
            config{config}
```

No need to `std::move` if you pass a const&

AlanGriffiths: > No need to std::move if you pass a const&

Well yes, that forces a copy from the parameter with or without `std::move()`. The alternatives:

The proposed code avoids a copy (by moving into the parameter and moving out).

A universal reference reduces that to a single move (if the argument is an rvalue reference)

tarek-y-ismail: Since in `enable` we have to pass an existing `config` (not an rvalue ref), then I think the best I can do is matt's suggestion

PR #4369, opened 2025-10-20
Deprecate MinimalWindowManager in favor of FloatingWindowManager

mattkae:
## What's new?
- Deprecate `MinimalWindowManager`
- Introduce `MinimalWindowManagerV2`
- `MinimalWindowManagerV2` removes alt tab/tick functionality, cleans up the constructor, and cleans up a bit of the code
- Added ``MinimalWindowManagerV2Test` which is a copy of ``MinimalWindowManagerTest` without the alt tab/tick tests
- Using `MinimalWindowManagerV2` in the example code instead of `MinimalWindowManager`

## How to test
Run miral-shell and see that alt tab is gone but everything else works :tada: 

## Checklist

- [x] Tests added and pass

### DIFF
--- a/debian/libmiral7.symbols
+++ b/debian/libmiral7.symbols
@@ -705,6 +705,8 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::LocatePointer::operator()(mir::Server&)@MIRAL_5.6" 5.6.0
  (c++)"miral::LocatePointer::schedule_request()@MIRAL_5.6" 5.6.0
  (c++)"miral::WaylandExtensions::ext_data_control_manager_v1@MIRAL_5.6" 5.6.0
+ (c++)"miral::WaylandExtensions::ext_image_copy_capture_manager_v1@MIRAL_5.6" 5.6.0
+ (c++)"miral::WaylandExtensions::ext_output_image_capture_source_manager_v1@MIRAL_5.6" 5.6.0
  (c++)"miral::live_config::Key::Key(miral::live_config::Key const&)@MIRAL_5.5" 5.6.0
  (c++)"typeinfo for miral::FloatingWindowManager@MIRAL_5.6" 5.6.0
  (c++)"vtable for miral::FloatingWindowManager@MIRAL_5.6" 5.6.0
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -21,6 +21,7 @@
 #include "wallpaper_config.h"
 #include "spinner/splash.h"
 
+#include <miral/application_switcher.h>
 #include <miral/display_configuration_option.h>
 #include <miral/external_client.h>
 #include <miral/runner.h>
@@ -86,8 +87,9 @@ int main(int argc, char const* argv[])
         };
 
     MirRunner runner{argc, argv};
+    ApplicationSwitcher application_switcher;
 
-    runner.add_stop_callback([&] { shutdown_hook(); });
+    runner.add_stop_callback([&] { shutdown_hook(); application_switcher.stop(); });
 
     ExternalClientLauncher external_client_launcher;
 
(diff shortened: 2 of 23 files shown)


tarek-y-ismail:
@mattkae Can you fix the conflict and the build error?

mattkae:
> @mattkae Can you fix the conflict and the build error?

Good idea :smile: Done!

examples/example-server-lib/floating_window_manager.cpp
```diff
@@ -77,7 +78,7 @@ FloatingWindowManagerPolicy::~FloatingWindowManagerPolicy() = default;
 
 bool FloatingWindowManagerPolicy::handle_pointer_event(MirPointerEvent const* event)
 {
-    if (MinimalWindowManager::handle_pointer_event(event))
+    if (MinimalWindowManagerV2::handle_pointer_event(event))
```
RAOF: I wonder if we sholudn't use the `using super = MinimalWindowManagerV2` trick in the definition of `FloatingWindowManagerPolicy` to make all of these `super::handle_pointer_event()` etc.

More a comment than a question, though :)

RAOF (approved):
Looks sensible to me.

include/miral/miral/floating_window_manager.h
```diff
@@ -0,0 +1,85 @@
...
+    void advise_new_window(WindowInfo const& app_info) override;
+
+protected:
+    WindowManagerTools tools;
+
+    bool begin_pointer_move(WindowInfo const& window_info, MirInputEvent const* input_event);
+    bool begin_pointer_resize(WindowInfo const& window_info, MirInputEvent const* input_event, MirResizeEdge const& edge);
+
+    bool begin_touch_move(WindowInfo const& window_info, MirInputEvent const* input_event);
+    bool begin_touch_resize(WindowInfo const& window_info, MirInputEvent const* input_event, MirResizeEdge const& edge);
```
AlanGriffiths, line 135: No doc comments?

include/miral/miral/minimal_window_manager_v2.h
```diff
@@ -0,0 +1,85 @@
...
+/// - Alt + F4 keybind to close windows
+/// - Focus stealing management
+///
+/// Compositor authors may elect to use this class as provided or subclass it in order
+/// to extend the base functionality.
+///
+/// This class is the successor to #miral::MinimalWindowManager.
+///
+/// \remark Since MirAL 5.6
+class MinimalWindowManagerV2 : public WindowManagementPolicy
```
AlanGriffiths: Maybe we should have a name with better appeal for the long term? `StackingWindowManager`? `FloatingWindowManager`?

include/miral/miral/minimal_window_manager_v2.h
```diff
@@ -0,0 +1,85 @@
...
+
+    bool begin_pointer_move(WindowInfo const& window_info, MirInputEvent const* input_event);
+    bool begin_pointer_resize(WindowInfo const& window_info, MirInputEvent const* input_event, MirResizeEdge const& edge);
+
+    bool begin_touch_move(WindowInfo const& window_info, MirInputEvent const* input_event);
+    bool begin_touch_resize(WindowInfo const& window_info, MirInputEvent const* input_event, MirResizeEdge const& edge);
+
+private:
+    struct Impl;
+    std::shared_ptr<Impl> const self;
```
AlanGriffiths: We should not be copyable
```suggestion
    std::unique_ptr<Impl> const self;
```
(I guess that would need the destructor implemented in the .cpp)

include/miral/miral/minimal_window_manager.h
```diff
@@ -17,19 +17,16 @@
 #ifndef MIRAL_MINIMAL_WINDOW_MANAGER_H
 #define MIRAL_MINIMAL_WINDOW_MANAGER_H
 
+#include "focus_stealing.h"
```
AlanGriffiths: It is probably better to be consistent with:

```suggestion
#include <miral/focus_stealing.h>
```

include/miral/miral/minimal_window_manager_v2.h
```diff
@@ -0,0 +1,85 @@
...
+///
+/// This class is the successor to #miral::MinimalWindowManager.
+///
+/// \remark Since MirAL 5.6
+class MinimalWindowManagerV2 : public WindowManagementPolicy
+{
+public:
+    explicit MinimalWindowManagerV2(WindowManagerTools const& tools);
+    MinimalWindowManagerV2& focus_stealing(FocusStealing focus_stealing);
+    MinimalWindowManagerV2& pointer_drag_modifier(MirInputEventModifier pointer_drag_modifier);
```
AlanGriffiths: `MinimalWindowManager` had constructors for these options that work with `set_window_management_policy()`. As downstream code doesn't directly create the window manager I don't see how these methods can be used effectively.

include/miral/miral/focus_stealing.h
AlanGriffiths, line 1: This header seems somewhat lightweight, I wonder if there's a more generic place for `FocusStealing` - `window_management_policy.h`?

AlanGriffiths (changes requested):

include/miral/miral/minimal_window_manager_v2.h
mattkae: I like `FloatingWindowManager` personally

include/miral/miral/minimal_window_manager_v2.h
mattkae: Yeah I was just trying to prevent us having every combination of constructors there. Would it be okay if I wrapped those options in a `struct` so that users only need to provide a single parameter?

AlanGriffiths: I'd be good with one constructor that takes all three arguments. The others were for backward compatibility

mattkae: Ok, ill do that instead

PR #4373, opened 2025-10-22
Route all cursor input through `the_cursor_controller()`

RAOF:
Rather than have various components - `BasicSeat` and `SeatInputDeviceTracker` - take both the `CursorController` and the `CursorObserverMultiplexer`, instead pass only the `CursorController` and have *it* dispatch to the `CursorObserverMultiplexer`.

This will then make it easy to add a `CursorObserver::image_set_to` observation, which in turn will be useful for the image-copy cursor session code.

### DIFF
--- a/tests/mir_test_framework/test_wlcs_display_server.cpp
+++ b/tests/mir_test_framework/test_wlcs_display_server.cpp
@@ -854,6 +854,7 @@ miral::TestWlcsDisplayServer::TestWlcsDisplayServer(int argc, char const** argv)
     WlcsDisplayServer::create_pointer = &wlcs_server_create_pointer;
     WlcsDisplayServer::create_touch = &wlcs_server_create_touch;
 
+    add_to_environment("MIR_SERVER_CURSOR", "null");
     add_to_environment("MIR_SERVER_ENABLE_KEY_REPEAT", "false");
     char buffer[32];
     snprintf(buffer, sizeof buffer, "wlcs-tests-%d", getpid());


Saviq:
WLCS doesn't like it, though.

AlanGriffiths (changes requested):
> WLCS doesn't like it, though.

jhenstridge:
It looks like this might send fewer `pointer_usable` / `pointer_unusable` events to observers on the multiplexer.

It's not clear that's necessarily wrong, but it could be triggering the test failures.

src/server/input/cursor_controller.cpp
```diff
@@ -290,6 +293,7 @@ void mir::input::CursorController::pointer_usable()
 
     if (became_usable)
     {
+        cursor_observer->pointer_usable();
```
jhenstridge, line 296: Here, you're calling the cursor observer multiplexer while holding the the `serialize_pointer_usable_unusable` lock, which I'm not sure is desirable. It probably doesn't matter because it is just queuing up a bunch of tasks in the executor.

The fact that we have `CursorController` doing work here immediately rather than deferring it like the multiplexer does might be the cause of the wlcs failure.

Saviq:
@raof poke!

AlanGriffiths:
> @RAOF poke!

@RAOF poke!

src/server/input/cursor_controller.cpp
AlanGriffiths: I think this is a theoretical issue (and in `pointer_unusable()`) because we don't locally know that the executor actually queues work (it could be an `ImmediateExecutor`, especially in tests).

AlanGriffiths (approved):
I don't think it likely we'll get deadlock from `serialize_pointer_usable_unusable` callbacks in practice.

PR #4380, opened 2025-10-24
Fix `LocatePointer` being invoked with extra unintended keys pressed

### LINKED ISSUE #4378: Any key being pressed with CTRL should cancel "Locate Pointer"
When using `miral-app`, pressing CTRL will schedule Locate Pointer to animate the pointer, but pressing another button should cancel the scheduled request. 

For example, pressing CTRL+ALT should _not_ trigger Locate Pointer.

tarek-y-ismail:
Closes #4378 

## What's new?

- Modifies the Locate Pointer keyboard filter in `shell_main.cpp` to cancel the scheduled callback if any modifier is pressed before or after `ctrl`, and if any key is pressed _after_ ctrl

## How to test

- Run `miral-app`
- Try pressing just CTRL (left or right). The cursor should scale temporarily
- Try pressing CTRL + any other key, or any other modifier + CTRL. The cursor should not scale.

## Checklist
- [ ] ~Tests added and pass~ Not easily testable at the moment. See  #4401

### DIFF
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -315,23 +315,45 @@ int main(int argc, char const* argv[])
 
     auto const locate_pointer_filter = [&locate_pointer](auto const* keyboard_event)
     {
-            auto const keysym = mir_keyboard_event_keysym(keyboard_event);
-            if (keysym != XKB_KEY_Control_R && keysym != XKB_KEY_Control_L)
-                return false;
+        auto const keysym = mir_keyboard_event_keysym(keyboard_event);
+        auto const action = mir_keyboard_event_action(keyboard_event);
+        auto const modifiers = mir_keyboard_event_modifiers(keyboard_event);
 
-            switch (mir_keyboard_event_action(keyboard_event)) {
-                case mir_keyboard_action_down:
-                    locate_pointer.schedule_request();
-                    break;
-                case mir_keyboard_action_up:
-                    locate_pointer.cancel_request();
-                    break;
+        auto only_ctrl_down = [modifiers]
+        {
+            auto constexpr ctrls_mask = mir_input_event_modifier_ctrl | mir_input_event_modifier_ctrl_left |
+                                    mir_input_event_modifier_ctrl_right;
+            auto const any_ctrl_down = modifiers & ctrls_mask;
+            auto const any_other_modifier_down = modifiers & ~ctrls_mask;
 
-                default:
-                    return false;
+            return any_ctrl_down && !any_other_modifier_down;
+        }();
+
+
+        auto other_key_down = false;
+        // If ctrl is the only modifier pressed, make sure the keysym is also ctrl
+        if (only_ctrl_down)
+        {
+            switch (action)
+            {
+            case mir_keyboard_action_down:
+                other_key_down = (keysym != XKB_KEY_Control_R && keysym != XKB_KEY_Control_L);
+                break;
+            case mir_keyboard_action_up:
+                only_ctrl_down = false;
+                break;
+
+            default:
+                break;
             }
+        }
 
-            return false;
+        if (only_ctrl_down && !other_key_down)
+            locate_pointer.schedule_request();
+        else
+            locate_pointer.cancel_request();
+
+        return false;
     };
 
     return runner.run_with(


examples/miral-shell/shell_main.cpp
```diff
@@ -313,25 +314,53 @@ int main(int argc, char const* argv[])
...
 
-                default:
-                    return false;
-            }
+        // More than one key, bail
+        if(pressed_keys.size() != 1)
+        {
+            locate_pointer.cancel_request();
+            return false;
+        }
```
mattkae: I feel like the unordered set is overkill here? Can't you just check if the incoming pressed key is not ctrl, and then cancel the request if it is? And then you can just have a `bool` for "is_non_ctrl_key_down" or something?

tarek-y-ismail: That would work only if CTRL was pressed earlier in the key sequence, but for things like ALT+CTRL or other sequences ending with CTRL, we need to keep track of which keys are down

mattkae: I think that the set is still overkill though, and potentially wrong. This would be a simpler algorithm (pseudocode):

```c++
int key_pressed = 0;
auto const locate_pointer_filter
{
    if (is_key_down_event(event))
    {
        if (key != Ctrl || pressed > 0)
            locate_pointer.cancel_request();
        else
            locate_pointer.schedule_request();

        pressed++;
    }
    else if (is_key_up_event(event))
    {
        locate_pointer.cancel_request();
        pressed--;
    }
}
```

So:
- If `ctrl` is pressed and held, then locate pointer will be scheduled.
- If `alt` is pressed and then `ctrl`, then locate pointer will always be cancelled.
- If `ctrl` is pressed and then `alt`, then locate pointer will always be cancelled.
- If an key is released, then locate pointer will always be cancelled.

I think that this is simpler and more robust. Am I missing any case?

tarek-y-ismail: Nice one! :+1:

examples/miral-shell/shell_main.cpp
```diff
@@ -43,6 +42,7 @@
 #include <miral/magnifier.h>
 #include <miral/cursor_scale.h>
 
+#include <unordered_set>
```
AlanGriffiths:
```suggestion
```

examples/miral-shell/shell_main.cpp
```diff
@@ -315,81 +314,29 @@ int main(int argc, char const* argv[])
                                                  { cursor_scale.scale_temporarily(5, std::chrono::seconds{1}); })
                               .delay(std::chrono::milliseconds{1000});
 
-    auto const locate_pointer_filter = [&locate_pointer](auto const* keyboard_event)
+    auto const locate_pointer_filter = [&locate_pointer, pressed_keys = 0](auto const* keyboard_event) mutable
```
AlanGriffiths: I worry that this `pressed_keys` counter approach is fragile. Other input filters could, in principle, consume key up/down events before they are seen by `locate_pointer_filter` which would leave the counter out of step with reality.

Another approach is to check the modifier state against `mir_keyboard_event_modifiers()` when we see the Ctrl up/down events.

tarek-y-ismail: If we only check modifiers, then combinations like CTRL+S will still schedule the callback

AlanGriffiths: Not if we cancel when we see up/down/repeat events for !Ctrl events

examples/miral-shell/shell_main.cpp
```diff
@@ -313,25 +313,29 @@ int main(int argc, char const* argv[])
...
-                    locate_pointer.cancel_request();
-                    break;
+        switch (action)
+        {
+        case mir_keyboard_action_down:
+            if ((keysym != XKB_KEY_Control_R && keysym != XKB_KEY_Control_L))
+                locate_pointer.cancel_request();
+            else
+                locate_pointer.schedule_request();
+            break;
```
mattkae: If we press Alt and then Ctrl, won't it be broken? Should we check that no `modifier` is set when seeing the `ctrl` click?

mattkae (changes requested):

examples/miral-shell/shell_main.cpp
tarek-y-ismail: Gah! You're right, this is also the case if any other key is pressed as well :facepalm:

tarek-y-ismail: I'm not sure if there's a workaround to this without keep track of button state, which as Alan mentioned, might go out of sync..

mattkae: I think we just need to make sure there are no `modifiers` down, right? If someone does `Alt + Ctrl` we will catch it, because the modifiers will be `alt`.

However, this does not catch `s + Ctrl`, but I think that is fine personally

tarek-y-ismail: The common case is `modifier + ctrl` or `ctrl + modifier`. Let's get that in and if `key + ctrl` proves to be too annoying in daily use, we can add back the code for that.

mattkae: I think the later is less of a problem. Agreed. Just check for modifiers for now :)

examples/miral-shell/shell_main.cpp
```diff
@@ -315,83 +313,44 @@ int main(int argc, char const* argv[])
                                                  { cursor_scale.scale_temporarily(5, std::chrono::seconds{1}); })
                               .delay(std::chrono::milliseconds{1000});
 
-    auto const locate_pointer_filter = [&locate_pointer](auto const* keyboard_event)
+    auto const locate_pointer_filter = [&locate_pointer](auto const* keyboard_event) mutable
```
AlanGriffiths: No need for this change

examples/miral-shell/shell_main.cpp
```diff
@@ -317,81 +315,44 @@ int main(int argc, char const* argv[])
...
-                case mir_keyboard_action_down:
-                    locate_pointer.schedule_request();
-                    break;
-                case mir_keyboard_action_up:
-                    locate_pointer.cancel_request();
-                    break;
+        auto only_ctrl_down = [modifiers]
+        {
+            auto const ctrls_mask = mir_input_event_modifier_ctrl | mir_input_event_modifier_ctrl_left |
+                                    mir_input_event_modifier_ctrl_right;
```
AlanGriffiths:
```suggestion
            auto constexpr ctrls_mask = mir_input_event_modifier_ctrl | mir_input_event_modifier_ctrl_left |
                                    mir_input_event_modifier_ctrl_right;
```

PR #4382, opened 2025-10-27
feature: implement the StandardApplicationSwitcher for ease of use

mattkae:
## What's new?
- Provide a generalized usage of the `ApplicationSwitcher` via the `StandardApplicationSwitcher`
- Use the `StandardApplicationSwitcher` in `miral-shell`
- Removed the app switcher from the demo server, as it's unneeded there
- Added a new constructor to `miral::StartupInternalClient` so that internal clients may provide a callback before they are joined

## How to test
1. Run `miral-app`
2. Click alt + tab, alt + shift tab, alt + grave, or alt + shift grave and note that they all work :tada: 

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -763,4 +763,3 @@ global:
     vtable?for?miral::FloatingWindowManager;
   };
 } MIRAL_5.5;
-


include/miral/miral/application_switcher.h
```diff
@@ -89,6 +92,83 @@ class ApplicationSwitcher
...
+
+    /// Sets the application scancode.
+    ///
+    /// This is the key that must be clicked in order to trigger either
+    /// #miral::ApplicationSwitcher::next_app or #miral::ApplicationSwitcher::prev_app,
+    /// depending on the modifiers that are currently being held.
+    ///
+    /// Defaults to KEY_TAB.
+    ///
+    /// \param key the key to trigger an application switch
```
robert-ancell: parameter name mismatch: key, scancode

include/miral/miral/application_switcher.h
```diff
@@ -89,6 +92,83 @@ class ApplicationSwitcher
...
+
+    /// Sets the window scancode.
+    ///
+    /// This is the key that must be clicked in order to trigger either
+    /// #miral::ApplicationSwitcher::next_window or #miral::ApplicationSwitcher::prev_window,
+    /// depending on the modifiers that are currently being held.
+    ///
+    /// Defaults to KEY_GRAVE.
+    ///
+    /// \param key the key to trigger a window switch
```
robert-ancell: Name mismatch like above.

debian/libmiral7.symbols
```diff
@@ -664,6 +664,7 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++|arch-bits=64)"miral::live_config::IniFile::add_strings_attribute(miral::live_config::Key const&, std::basic_string_view<char, std::char_traits<char> >, std::span<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, 18446744073709551615ul>, std::function<void (miral::live_config::Key const&, std::optional<std::span<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, 18446744073709551615ul> >)>)@MIRAL_5.5" 5.5.0
  MIRAL_5.6@MIRAL_5.6 5.6.0
  (c++)"miral::ApplicationSwitcher::ApplicationSwitcher()@MIRAL_5.6" 5.6.0
+ (c++)"miral::ApplicationSwitcher::ApplicationSwitcher(miral::ApplicationSwitcher const&)@MIRAL_5.6" 5.6.0
```
tarek-y-ismail, line 667: Unrelated? We really never need to copy the application switcher if I understand correctly

include/miral/miral/application_switcher.h
```diff
@@ -89,6 +92,83 @@ class ApplicationSwitcher
...
+    /// Sets the primary modifier for the keyboard filter.
+    ///
+    /// This is the modifier that must be held for any application switcher
+    /// action to begin.
+    ///
+    /// Defaults to #mir_input_event_modifier_alt.
+    ///
+    /// \param modifier the new modifier
+    /// \returns this
+    ApplicationSwitcherKeyboardFilter& primary_modifier(MirInputEventModifier modifier);
```
AlanGriffiths: Do we really need the complexity of chaining and allowing changes at runtime? This seems more like a case for a parameter struct (and designated initializers) passed to the constructor.

examples/miral-shell/shell_main.cpp
```diff
@@ -430,6 +371,6 @@ int main(int argc, char const* argv[])
             locate_pointer,
             AppendKeyboardEventFilter{locate_pointer_filter},
             StartupInternalClient{application_switcher},
-            AppendKeyboardEventFilter{application_switcher_filter}
+            AppendKeyboardEventFilter{ApplicationSwitcherKeyboardFilter(application_switcher)}
```
AlanGriffiths: This still looks complex to use:

We need to:

1. start the internal client
2. create the ApplicationSwitcherKeyboardFilter
3. pass events to the ApplicationSwitcherKeyboardFilter

This gets even more messy if we use function chaining to override the defaults.

We could have something like:

```C++
            EasyApplicationSwitcher({.primary_modifier = mir_input_event_modifier_ctrl}}
```

With `EasyApplicationSwitcher` doing the app startup and event filtering in `operator()()`

include/miral/miral/application_switcher.h
mattkae: I personally like being able to change these types of things at runtime, as the need inevitably shows up, but I would be okay leaving it out for the first version of this. I think it will come up though :)

examples/miral-shell/shell_main.cpp
mattkae: I can do that instead

include/miral/miral/application_switcher.h
AlanGriffiths: I'm seeing the requirement as a simple way to do the default thing. Those with complex requirements have the tools to "roll their own"

debian/libmiral7.symbols
mattkae: It gets copied when it is provided to the `runner.run_with`

AlanGriffiths: I agree it is copied, but the question here is whether this is a symbol provided by `libmiral` or whether the copy constructor is inlined (or both).

In any case, I don't see this PR changing the behavior.

include/miral/miral/application_switcher.h
```diff
@@ -89,6 +97,67 @@ class ApplicationSwitcher
...
+        /// #miral::ApplicationSwitcher::next_window or #miral::ApplicationSwitcher::prev_window,
+        /// depending on the modifiers that are currently being held.
+        ///
+        /// Defaults to KEY_GRAVE.
+        int window_key = KEY_GRAVE;
+    };
+
+    /// Create a new keyboard filter for the provided \p switcher with the provided
+    /// \p keybind_configuration.
+    ApplicationSwitcherKeyboardFilter(ApplicationSwitcher const& switcher, KeybindConfiguration const& keybind_configuration);
```
AlanGriffiths: Is there any value in forcing the user to create and pass the `switcher`?

Can't we encapsulate the remaining "boilerplate" the user needs? (Starting the internal client, registering the stop callback, etc)

mattkae:
@AlanGriffiths I refactored per your suggestion and created the `StandardApplicationSwitcher` instead. It is much cleaner :smile: 

As part of this work, I also added a new constructor to `StartupInternalClient`. This let's `StartupInternalClient` take a `join_callback`, which may be called before the join is triggered on the thread. Previously, users had to manually call `stop()` in the stop callback. This worked by accident because if `join_client_threads` was called before the `stop()` could be triggered, then the internal client would hang forever. Things only worked because they were accidentally ordered correctly in the list. This is a much better way of doing it in my opinion.

Let me know if you would prefer the changes to `StartupInternalClient` to be a separate pull request.

include/miral/miral/internal_client.h
```diff
@@ -55,6 +55,26 @@ class StartupInternalClient
...
+    ///
+    /// The \p connect_notification call will be called when a connection to the server has
+    /// been secured. This is called on a worker thread and must not block.
+    ///
+    /// The \p client_code callback will be called when the client has been initialized and is ready
+    /// to start interacting with the server. Note that this callback happens on another thread.
+    /// This callback must exit.
+    ///
+    /// The \p join_callback will be called before a join is requested on the client thread so that
+    /// the client may perform some cleanup (e.g. setting a flag to tell the Wayland event loop to stop).
```
AlanGriffiths: `join_callback` isn't in the language of the "internal client" code. From that perspective it is `handle_stop` (or similar).

include/miral/miral/internal_client.h
```diff
@@ -55,6 +55,26 @@ class StartupInternalClient
...
+    ///
+    /// \param client_code called when the Wayland client is initialized
+    /// \param connect_notification called when the session has connected
+    /// \param join_callback called before the client thread joins
```
AlanGriffiths: We typically note the MirAL version this is introduced in.

src/miral/internal_client.cpp
```diff
@@ -99,14 +100,17 @@ class WlInternalClientRunner : public Base, public virtual InternalClientRunner
     std::thread thread;
     std::function<void(struct ::wl_display* display)> const client_code;
     std::function<void(std::weak_ptr<mir::scene::Session> const session)> const connect_notification;
+    std::optional<std::function<void()>> join_callback;
```
AlanGriffiths: It would be simpler (no need for an `optional<>` or branching logic) to always have a callback, but to default it to a NOP

include/miral/miral/application_switcher.h
```diff
@@ -20,6 +20,14 @@
 #include <miral/window_manager_tools.h>
 #include <memory>
 #include <wayland-client.h>
+#include <functional>
+#include <miral/toolkit_event.h>
+#include <linux/input-event-codes.h>
```
AlanGriffiths, line 24: Not our usual header grouping

include/miral/miral/application_switcher.h
```diff
@@ -89,6 +97,68 @@ class ApplicationSwitcher
...
+/// This class will provide the following keybinds by default:
+/// - Call #miral::ApplicationSwitcher::next_app on `alt + tab`.
+/// - Call #miral::ApplicationSwitcher::prev_app on `alt + shift + tab`
+/// - Call #miral::ApplicationSwitcher::next_window on `alt + grave`
+/// - Call #miral::ApplicationSwitcher::prev_window` on `alt + shift + grave`.
+///
+/// This class should be provided to a #miral::MirRunner::run_with.
+///
+/// \remark Since MirAL 5.6
+class StandardApplicationSwitcher
```
AlanGriffiths: This name invites bike-shedding 🫤

AlanGriffiths (changes requested):
Yes, that looks better. One thing unresolved from previous review and a few suggestions

include/miral/miral/application_switcher.h
mattkae: Yeah... It is a particular weird name. I can't think of anything better though

AlanGriffiths: "Simple", "Basic", "Quick", "Canonical", ...

mattkae:
@AlanGriffiths The symbols are mad because I'm missing `_ZTIN5miral19ApplicationSwitcherE@MIRAL_5.6 5.6.0`. I think I need it then, right?

src/miral/application_switcher.cpp
```diff
@@ -837,3 +841,86 @@ void miral::ApplicationSwitcher::stop() const
 {
     self->stop();
 }
+
+class miral::StandardApplicationSwitcher::Self
+{
+public:
+    explicit Self(KeybindConfiguration const& keybind_configuration)
+        : startup_internal_client(switcher, switcher, [this] { switcher.stop(); }),
```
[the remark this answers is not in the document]
AlanGriffiths: The explanation is rubbish, but the issue is real: capturing `this` in the initializer can be problematic as it gives apparent access to the whole object which is not yet initialized. (`switcher`, however, has been initialized and is not a problem)

debian/libmiral7.symbols
```diff
@@ -704,10 +705,17 @@ libmiral.so.7 libmiral7 #MINVER#
...
+ (c++)"miral::StandardApplicationSwitcher::StandardApplicationSwitcher(miral::StandardApplicationSwitcher::KeybindConfiguration const&)@MIRAL_5.6" 5.6.0
+ (c++)"miral::StandardApplicationSwitcher::operator()(mir::Server&) const@MIRAL_5.6" 5.6.0
+ (c++)"miral::StandardApplicationSwitcher::~StandardApplicationSwitcher()@MIRAL_5.6" 5.6.0
+ (c++)"miral::StartupInternalClient::StartupInternalClient(std::function<void (wl_display*)>, std::function<void (std::weak_ptr<mir::scene::Session>)>, std::function<void ()>)@MIRAL_5.0" 5.6.0
+ (c++)"miral::StartupInternalClient::StartupInternalClient<miral::ApplicationSwitcher>(miral::ApplicationSwitcher const&)@MIRAL_5.0" 5.6.0
  (c++)"miral::WaylandExtensions::ext_data_control_manager_v1@MIRAL_5.6" 5.6.0
  (c++)"miral::WaylandExtensions::ext_image_copy_capture_manager_v1@MIRAL_5.6" 5.6.0
  (c++)"miral::WaylandExtensions::ext_output_image_capture_source_manager_v1@MIRAL_5.6" 5.6.0
  (c++)"miral::live_config::Key::Key(miral::live_config::Key const&)@MIRAL_5.5" 5.6.0
+ (c++)"typeinfo for miral::ApplicationSwitcher@MIRAL_5.6" 5.6.0
```
AlanGriffiths, line 718: Why does this PR introduces this symbol?

mattkae: No idea! I wonder why the symbols generator demands it...

mattkae: Ok I think I understand it better.

I am using a raw `ApplicationSwitcher` in `class miral::StandardApplicationSwitcher::Self`. It is a direct member (not a pointer), so the compiler needs complete type information.

I can avoid this by wrapping it in a `std::shared_ptr` probably, but that feels like overkill.

WDYT?

include/miral/miral/application_switcher.h
mattkae: Maybe `Basic` is best?

debian/libmiral7.symbols
AlanGriffiths: I don't think we need a `shared_ptr`, just to understand.

So what you are saying is that the inline, compiler generated, copy constructor and typeinfo require `.symbols` entries if the the copy constructor is used internally in libmiral. (But, obviously, was not needed previously when the copy constructor was used in the examples.)

It sounds mad, but I guess using the function causes a weak symbol to be emitted and that is seen by `dpkg-gensymbols`.

All a bit fragile...

PR #4384, opened 2025-10-29
output_filter=none

AlanGriffiths:
1. Stop logging `< -warning- > miral: Config key 'output_filter' has invalid integer value: none` when `output_filter=none`
2. Provides a preset value to the config

## How to test

Watch the log when `output_filter=none` is in the configuration

### DIFF
--- a/tests/miral/output_filter.cpp
+++ b/tests/miral/output_filter.cpp
@@ -1,0 +1,67 @@
+/*
+* Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


tests/miral/output_filter.cpp
tarek-y-ismail, line 1: I don't think these tests would catch the bug

AlanGriffiths: True, but TICS is happy now

AlanGriffiths: @tarek-y-ismail sorry, that was flippant. I would also like such tests - I just haven't had headspace for creating them yet.

tarek-y-ismail: It's alright, we can split it off into a separate issue for now

AlanGriffiths: Here's the test we were missing: https://github.com/canonical/mir/pull/4388

PR #4390, opened 2025-10-31
wayland: separate the output capture specific details of ext-image-copy-capture into a backend class

jhenstridge:
Closes #???

Related: #4326

## What's new?

This is mostly a reorganisation of the ext-image-copy-capture extension to better support multiple capture backends. The current code handles the case of capturing a whole output, but we also want to support capture of cursor images and eventually toplevel windows.

The approach I took was:

1. Move all the output specific code from the `ExtImageCopyCaptureSessionV1` and `ExtImageCopyCaptureFrameV1` classes into an `ExtImageCopyBackend` helper class.
2. The Session class creates the backend using a factory function provided by the `ExtImageCaptureSourceV1`.
3. The `ExtOutputImageCaptureSourceManagerV1` class will provide an appropriate factory function for the sources it creates.

I've split the `ExtImageCopyBackend` class into an abstract base and a concrete `ExtOutputImageCopyBackend` subclass. Some stuff like damage tracking is in the base class, although it will probably still need a few tweaks for the cursor version.

## How to test

Start miral with the extensions enabled:

```
./bin/miral-app --add-wayland-extensions ext_output_image_capture_source_manager_v1:ext_image_copy_capture_manager_v1
```

Start wayvnc pointing at miral:

```
WAYLAND_DISPLAY=wayland-1 wayvnc --disable-input
```

Use a VNC client to observe it tracks output changes on the server.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/ext_image_capture_v1.cpp
+++ b/src/server/frontend_wayland/ext_image_capture_v1.cpp
@@ -224,6 +224,7 @@ public:
     ExtImageCopyCaptureFrameV1(wl_resource* resource, ExtImageCopyCaptureSessionV1* session);
 
     bool is_ready() const;
+    // \pre backend.has_damage() == true
     void begin_capture(ExtImageCopyBackend& backend);
     void report_result(ExtImageCopyBackend::CaptureResult const& result);
 


src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,401 @@ auto translate_and_scale(
...
+    wl_display* display,
+    std::shared_ptr<Executor> const& wayland_executor,
+    std::shared_ptr<compositor::ScreenShooterFactory> const& screen_shooter_factory,
+    std::shared_ptr<SurfaceStack> const& surface_stack)
 -> std::shared_ptr<wayland::OutputImageCaptureSourceManagerV1::Global>
 {
-    return std::make_shared<ExtOutputImageCaptureSourceManagerV1Global>(display);;
+    auto ctx = std::make_shared<ExtImageCaptureV1Ctx>(
+        wayland_executor, screen_shooter_factory, surface_stack);
+    return std::make_shared<ExtOutputImageCaptureSourceManagerV1Global>(display, std::move(ctx));
```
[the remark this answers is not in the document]
AlanGriffiths: To supplement what Copilot says: the parameter is taken by `const&`, not by value. In the latter case `move()` would be appropriate here.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,401 @@ auto translate_and_scale(
...
+    auto const buffer_size = output_config.modes[output_config.current_mode_index].size;
+    auto const output_space_area = output_config.extents();
+    auto const transform = output_config.transformation();
+
+    geom::Rectangle buffer_space_damage;
+    switch (damage_amount)
+    {
+    case DamageAmount::none:
+        BOOST_THROW_EXCEPTION(std::runtime_error("ExtImageCopyBackend::begin_capture called when there is no damage available"));
+        break;
```
AlanGriffiths, line 403: I *think* we can't get here with the existing logic in `maybe_capture_frame()` which might mean this is a precondition failure (or a `logic_error`, not a `runtime_error`) but mostly this looks like a change from the previous code (which quietly `return`ed)

AlanGriffiths:
Sorry, didn't finish looking at this before EOD. Will pick up again tomorrow

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,401 @@ auto translate_and_scale(
...
-    ExtImageCopyCaptureFrameV1(wl_resource* resource, ExtImageCopyCaptureSessionV1* session, std::shared_ptr<ExtImageCaptureV1Ctx> const& ctx);
+    ExtImageCopyCaptureFrameV1(wl_resource* resource, ExtImageCopyCaptureSessionV1* session);
 
     bool is_ready() const;
+    void report_result(std::optional<time::Timestamp> captured_time,
+                       geom::Rectangle buffer_space_damage);
 
-    void capture_frame(geom::Rectangle output_space_area, glm::mat2 transform, bool overlay_cursor, geom::Size buffer_size, geom::Rectangle buffer_space_damage);
+    wayland::Weak<wayland::Buffer> target;
+    geom::Rectangle frame_damage;
```
AlanGriffiths: Public, especially non-`const` data on a class looks odd.

They are is only accessed outside the class by `ExtImageCopyCaptureSessionV1::maybe_capture_frame()` and the usage there looks like there could be clearer abstractions:

* `maybe_get_sbm_buffer() -> ShmBuffer*`
* `reset_frame_damage() -> Rectangle` member would be a clearer abstraction than `read...reset`.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -403,46 +548,44 @@ void mf::ExtImageCopyCaptureSessionV1::maybe_capture_frame()
...
-    auto const buffer_size = output_config.modes[output_config.current_mode_index].size;
-    auto const output_space_area = output_config.extents();
-    auto const transform = output_config.transformation();
-
-    switch (damage_amount)
+    using FailureReason = wayland::ImageCopyCaptureFrameV1::FailureReason;
+    if (stopped)
     {
-    case DamageAmount::none:
+        current_frame.value().send_failed_event(FailureReason::stopped);
```
AlanGriffiths: I guess it is "pre-existing", but repeatedly calling `current_frame.value()` in this method seems wrong. (It can also prevent the compiler optimizing as it introduces potential exception paths.)

Would it be sensible to move this logic into `ExtImageCopyCaptureFrameV1` where the lifetime checks would be unnecessary?

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,401 @@ auto translate_and_scale(
...
+{
+    std::shared_ptr<Executor> const wayland_executor;
+    std::shared_ptr<compositor::ScreenShooterFactory> const screen_shooter_factory;
+    std::shared_ptr<SurfaceStack> const surface_stack;
+};
+
+/* Image copy backends */
+class ExtImageCopyCaptureSessionV1;
+
+class ExtImageCopyBackend {
```
RAOF: Nit: Coding style

```suggestion
class ExtImageCopyBackend
{
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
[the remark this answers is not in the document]
RAOF: Hm. I think I disagree with what Copilot says? I don't think `std::move()`ing here reduces code clarity; it clarifies that we're done with it here.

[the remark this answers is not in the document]
AlanGriffiths: I disagree with Chris: it is just syntactic noise that leaves the reader wondering what the intention was...

1. If `ExtOutputImageCaptureSourceManagerV1Global` took `ctx` by value then the `std::move()` would avoid copying the `shared_ptr` (with fewer atomic operations and synchronization primitives) at the cost of a larger entry on the stack.

2. If `ExtOutputImageCaptureSourceManagerV1Global` took `ctx` by universal reference (`&&`) and `std::move()`d from it, then that would avoid any copies and minimize the size of the stack entry.

The combination of `std::move()` and `const&` leaves the reader wondering if the author intended something like these.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,401 @@ auto translate_and_scale(
...
+
+    wayland::DestroyListenerId destroy_listener_id;
+    std::shared_ptr<scene::SceneChangeNotification> change_notifier;
+    std::unique_ptr<compositor::ScreenShooter> const screen_shooter;
+
+    void create_change_notifier();
+    auto output_config_changed(graphics::DisplayConfigurationOutput const& config) -> bool override;
+};
+
+using ExtImageCopyBackendFactory = std::function<ExtImageCopyBackend*(ExtImageCopyCaptureSessionV1*,bool)>;
```
RAOF: I think this would be clearer if the memory management was explicit
```suggestion
using ExtImageCopyBackendFactory = std::function<std::unique_ptr<ExtImageCopyBackend>(ExtImageCopyCaptureSessionV1*,bool)>;
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,428 @@ auto translate_and_scale(
...
-
-    std::shared_ptr<ExtImageCaptureV1Ctx> const ctx;
 };
 
 class ExtImageCopyCaptureManagerV1
     : public wayland::ImageCopyCaptureManagerV1
 {
 public:
-    ExtImageCopyCaptureManagerV1(wl_resource* resource, std::shared_ptr<ExtImageCaptureV1Ctx> const& ctx);
+    ExtImageCopyCaptureManagerV1(wl_resource* resource);
```
AlanGriffiths: Single parameter constructors should usually be `explicit`

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,428 @@ auto translate_and_scale(
...
 
 class ExtImageCopyCaptureFrameV1;
 
 class ExtImageCopyCaptureSessionV1
-    : public wayland::ImageCopyCaptureSessionV1, OutputConfigListener
+    : public wayland::ImageCopyCaptureSessionV1
 {
 public:
-    ExtImageCopyCaptureSessionV1(wl_resource* resource, ExtImageCaptureSourceV1* source, uint32_t options, std::shared_ptr<ExtImageCaptureV1Ctx> const& ctx);
+    ExtImageCopyCaptureSessionV1(wl_resource* resource, bool overlay_cursor, ExtImageCopyBackendFactory const& backend_factory);
```
AlanGriffiths, line 204: I thought we had some guidance against `bool` parameters - because it isn't clear from the call site what they mean. But I can't find it now.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,428 @@ auto translate_and_scale(
...
+        callback(std::make_optional(FailureReason::buffer_constraints), {}, {});
+        return;
+    }
+
+    geom::Rectangle buffer_space_damage;
+    switch (damage_amount)
+    {
+    case DamageAmount::none:
+        BOOST_THROW_EXCEPTION(std::logic_error("ExtImageCopyBackend::begin_capture called when there is no damage available"));
+        break;
```
AlanGriffiths: Do we really want to treat this as an exception? The previous logic simply `return`ed, has our understanding changed?

And is this really a logic error? (I.e. an error in our calling code) Or is it a runtime error? (the client has done something forbidden by the protocol)

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,428 @@ auto translate_and_scale(
...
 {
 public:
-    ExtImageCopyCaptureFrameV1(wl_resource* resource, ExtImageCopyCaptureSessionV1* session, std::shared_ptr<ExtImageCaptureV1Ctx> const& ctx);
+    ExtImageCopyCaptureFrameV1(wl_resource* resource, ExtImageCopyCaptureSessionV1* session);
 
     bool is_ready() const;
+    void begin_capture(ExtImageCopyBackend& backend);
+    void report_result(std::optional<uint32_t> failure_reason,
+                       time::Timestamp captured_time,
+                       geom::Rectangle buffer_space_damage);
```
AlanGriffiths: The whole call stack around this uses `std::optional` to signal whether the other parameters are meaningful.  That's a confusing style: This looks like a case for `std::expected`!

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,428 @@ auto translate_and_scale(
...
+
+void mf::ExtOutputImageCopyBackend::begin_capture(
+    std::shared_ptr<renderer::software::RWMappable> const& shm_data,
+    [[maybe_unused]] geom::Rectangle const& frame_damage,
+    std::function<void(std::optional<uint32_t>, time::Timestamp, geom::Rectangle)> const& callback)
+{
+    using FailureReason = wayland::ImageCopyCaptureFrameV1::FailureReason;
+    if (!output)
+    {
+        callback(std::make_optional(FailureReason::stopped), {}, {});
```
AlanGriffiths: I've commented that `std::expected` would be better for these calls, so this is likely moot, but note that `std::make_optional()` is unnecessary noise - there's a converting constructor.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +53,428 @@ auto translate_and_scale(
...
+        partial,
+        full,
+    };
+
+    ExtImageCopyCaptureSessionV1* const session;
+    bool const overlay_cursor;
+
+    geom::Rectangle output_space_area;
+    geom::Rectangle output_space_damage;
+    DamageAmount damage_amount = DamageAmount::full;
```
AlanGriffiths, line 106: Public[*] mutable data makes it hard to reason about invariants. (I realise the scope is limited to the current file - but classes do sometimes get moved into headers.)

[*] For this sort of thing `protected` is just as public as `public`

src/server/frontend_wayland/ext_image_capture_v1.cpp
AlanGriffiths: Hmm, I just saw my previous comment about the previous iteration of this. Clearly there's a need for some explanation of where any problem lies.

jhenstridge: I agree that there are some warts in the damage tracking part of the code. I know that some of it is likely to be common to other backends (mostly toplevel capture: cursor capture is likely to always be full damage). But some aspects are not general (e.g. the output space vs. buffer space parts).

I'm not entirely sure what the code should eventually look like, so I see this more as a stepping stone to a better split, once those other backends are in place.

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: I had extracted this from the flags parameter based on this suggestion: https://github.com/canonical/mir/pull/4326#discussion_r2446947646

I can see the problem with passing flags as multiple boolean arguments, but in this case we've only got a single one.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +56,430 @@ auto translate_and_scale(
...
+
+bool mf::ExtImageCopyBackend::has_damage()
+{
+    return damage_amount != DamageAmount::none;
+}
+
+mf::ExtOutputImageCopyBackend::ExtOutputImageCopyBackend(
+    ExtImageCopyCaptureSessionV1 *session,
+    bool overlay_cursor,
+    OutputGlobal *output,
```
RAOF: Stupid nit:
```suggestion
    ExtImageCopyCaptureSessionV1* session,
    bool overlay_cursor,
    OutputGlobal* output,
```

RAOF (approved):
I think I have a *slight* preference for the output-has-expired case to be handled in the `backend_factory`, so that the `ExtOutputImageCopyBackend` constructor is always given valid. But then the factory would need to return nullptr (or a Null object), so maybe what's here is fine 🤷‍♀️ 

Otherwise, I think this is an improvement on the existing state, and we can clean up further once there's another implementation of `ExtImageCopyBackend` to inform us.

src/server/frontend_wayland/ext_image_capture_v1.cpp
AlanGriffiths: The main reason to throw exceptions is a post-condition failure. And this does not look like one.

Looking at the rest of this call stack, it seems that a "nop" (which is what the old code did) is a valid post-condition when `DamageAmount::none`.

It looks like a _possible_ precondition failure. In particular, `ExtImageCopyCaptureSessionV1::maybe_capture_frame()` checks for `has_damage()` before calling `begin_capture()`. So, logically, "we can't get here". But we could consider the check in `maybe_capture_frame()` as an optimization, not a requirement.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +56,430 @@ auto translate_and_scale(
...
+/* Image copy backends */
+class ExtImageCopyCaptureSessionV1;
+
+class ExtImageCopyBackend
+{
+public:
+    using CaptureResult = std::expected<std::tuple<time::Timestamp, geom::Rectangle>, uint32_t>;
+    using CaptureCallback = std::function<void(CaptureResult const&)>;
+
+    ExtImageCopyBackend(ExtImageCopyCaptureSessionV1 *session, bool overlay_cursor);
```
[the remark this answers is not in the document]
AlanGriffiths: I disagree with copilot
```suggestion
    ExtImageCopyBackend(ExtImageCopyCaptureSessionV1* session, bool overlay_cursor);
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +56,430 @@ auto translate_and_scale(
...
+
+protected:
+    enum class DamageAmount
+    {
+        none,
+        partial,
+        full,
+    };
+
+    ExtImageCopyCaptureSessionV1* const session;
```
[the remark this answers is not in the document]
AlanGriffiths: I disagree with copilot

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: This method is only called once we've committed to capturing the frame. Having `ExtImageCopyCaptureSessionV1::maybe_capture_frame` return on no damage is now handled by this check:

https://github.com/canonical/mir/blob/34a0cf24944c2ada44a5d1412ea3036034cd8eb9/src/server/frontend_wayland/ext_image_capture_v1.cpp#L582-L587

So having no damage in `ExtOutputImageCopyBackend::begin_capture` is a precondition failure. I could switch out the exception for an assert() call if that's clearer.

AlanGriffiths: @copilot please review this thread and make a recommendation

AlanGriffiths:
Let me summarize the above discussion in the hope the bikeshed finally gets some paint!

As Chris says, `maybe_capture_frame()` is trying to coordinate too much detail, but this is non-blocking.

One of the things that is introduced to `maybe_capture_frame()`  here is a check for current damage and, if there is none, to return immediately.

James (and copilot) think that the existence of this check establishes a pre-condition for calling `begin_capture()` which would previously have treated "no damage" as an (expensive) no-op. It can also be viewed as an optimization to avoid the expensive no-op.

To enforce this hypothetical precondition, James introduced an exception and Copilot suggests aborting. That suggests that it is logically wrong to call `begin_capture()` as opposed to being potentially wasteful.

I'm not convinced, I would prefer not to have an implicit precondition on `begin_capture()` and also prefer not to report preconditions by exceptions but this isn't a hill I will die on. What does the team think?

AlanGriffiths:
Whoops! clang on 24.04 doesn't seem to know `std::{,un}expected`

AlanGriffiths:
> Whoops! clang on 24.04 doesn't seem to know `std::{,un}expected`

And that's because of the check `__cpp_concepts >= 202002L` (clang 18.1 has `201907L`)

... workaround here: https://github.com/canonical/mir/pull/4525

Saviq:
> I'm not convinced,

Same, but the other way around… while I agree it's not, logically, an error, it gives us the opportunity of avoiding "holding it wrong" cases… Do we have a better way to assert the precondition that there _is_ damage?

AlanGriffiths:
> Do we have a better way to assert the precondition that there _is_ damage?

We will soon: https://en.cppreference.com/w/cpp/language/functions/function_contract_specifiers.html

Saviq:
> We will soon

> triggering a violation (e.g. termination) in debug builds … and can be ignored in release builds for performance.

Sounds like an `assert` until we have those, then? ¯\\\_(ツ)_/¯

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +56,430 @@ auto translate_and_scale(
...
 
     static ExtImageCaptureSourceV1* from_or_throw(wl_resource* resource);
 
-    // FIXME: currently this is only set up for output capture. It
-    // will need to change to also cover toplevel capture.
-    wayland::Weak<OutputGlobal> const output;
-
-private:
-    wayland::Weak<ExtOutputImageCaptureSourceManagerV1> const manager;
+    ExtImageCopyBackendFactory const backend_factory;
```
tarek-y-ismail, line 172: Can be private?

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +56,430 @@ auto translate_and_scale(
...
+    }
+
+    return ExtImageCopyBackend::has_damage();
+}
+
+void mf::ExtOutputImageCopyBackend::create_change_notifier()
+{
+    auto callback = [this, weak_self=wayland::make_weak(this)](std::optional<geom::Rectangle> const& damage)
+        {
+            ctx->wayland_executor->spawn([weak_self, damage]()
```
tarek-y-ismail, line 359: Can drop the parens
```suggestion
            ctx->wayland_executor->spawn([weak_self, damage]
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
tarek-y-ismail: I agree that in normal code, this path should never be taken. If it's taken, it's a programmer error. Whether or not we use an assert or an exception comes up to this: 
- Exceptions allow us to catch them and recover, they exist in both debug and non-debug builds. 
- Asserts don't allow recovery and crash the program, they exist in debug builds.

The previous code indeed didn't blow up when this happened, but then again it _should have_ since we were explicitly checking for damage before calling the method.

Based on all the previous, both throwing `std::logic_error` or `assert`ing _can_ work. Though since we don't really want to recover, then an exception is not much more useful.

If the big issue with asserts is them changing behavior between debug and non-debug builds, there are many other ways of blowing up in both cases.

jhenstridge:
The `fatal_error_abort` option seems like the right one to me. One thing that I've realised is that there is another constraint I wasn't enforcing: that each frame object is only captured into once. If we've got to the point of calling begin_capture() on the frame, the only two outcomes should be either (a) a capture is successful, or (b) a failure is reported and the client should destroy the frame.

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: At the moment, `ExtImageCopyCaptureManagerV1::create_session` accesses the member. I could add an accessor and make the member private, but the public const member seems okay given how closely tied the classes are.

tarek-y-ismail: I did not realize that. That makes more sense now.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +57,432 @@ auto translate_and_scale(
...
+        callback(std::unexpected(FailureReason::buffer_constraints));
+        return;
+    }
+
+    geom::Rectangle buffer_space_damage;
+    switch (damage_amount)
+    {
+    case DamageAmount::none:
+        // This should never happen as maybe_capture_frame() checks has_damage() before calling begin_capture()
+        fatal_error_abort("begin_capture() called with no damage - this is a precondition violation");
```
AlanGriffiths, line 402: Should we document this precondition in `ExtImageCopyBackend`? I don't think this will be obvious to potential callers without reading this implementation of the interface.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -55,183 +57,432 @@ auto translate_and_scale(
...
+            {{}, buffer_size});
+        break;
+
+    case DamageAmount::full:
+        buffer_space_damage = {{}, buffer_size};
+        break;
+    }
+
+    // TODO: capture only the region covered by
+    // union(buffer_space_damage, frame->frame_damage)
```
AlanGriffiths, line 418: Does this TODO need to be resolved? Or is this already planned for a followup?

PR #4404, opened 2025-11-06
Providing the user with useful errors in the symbols map generator

mattkae:
## What's new?
- The symbols map generator now tells you how to get it to work by installing the right packages
- Removed the README.md

## Checklist

- [x] Tests added and pass


tools/symbols_map_generator/main.py
```diff
@@ -15,15 +15,27 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
+_logger = logging.getLogger(__name__)
+
+library_not_found_error_msg: str = ("    Please install libclang-19-dev and python3-clang-19."
+    "\n    Also, ensure that the required environment variables are set correctly. (e.g.)"
+    "\n        export MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH=/usr/lib/llvm-19/lib/libclang.so.1"
+    "\n        export MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib")
```
jhenstridge: This might be more readable as:
```suggestion
library_not_found_error_msg: str = """\
    Please install libclang-19-dev and python3-clang-19.
    Also, ensure that the required environment variables are set correctly. (e.g.)
        export MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH=/usr/lib/llvm-19/lib/libclang.so.1
        export MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib"""
```

tools/symbols_map_generator/main.py
```diff
@@ -15,15 +15,27 @@
...
 from typing import TypedDict, Optional
 from collections import OrderedDict
 import bisect
-import clang.cindex
+try:
+    import clang.cindex
+except ModuleNotFoundError as e:
+    _logger.error("Could not import clang in python.")
+    _logger.error(library_not_found_error_msg)
+    raise e
```
jhenstridge:
```suggestion
except ModuleNotFoundError:
    _logger.error("Could not import clang in python.")
    _logger.error(library_not_found_error_msg)
    raise
```

Also: this logging is happening before the `logging.basicConfig()` call, so will likely display differently to later log messages.

tools/symbols_map_generator/main.py
```diff
@@ -597,4 +605,8 @@ def main():
         f.truncate()
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except Exception as e:
+        _logger.error(library_not_found_error_msg)
+        _logger.error(e)
```
jhenstridge:
```suggestion
    except Exception:
        _logger.exception(library_not_found_error_msg)
        sys.exit(1)
```

We still want a non-zero exit code if the program fails.

tools/symbols_map_generator/main.py
```diff
@@ -15,16 +15,31 @@
...
-import os
 
+logging.basicConfig(level=logging.INFO)
 _logger = logging.getLogger(__name__)
 
+library_not_found_error_msg: str = """\
+    Please install libclang-19-dev and python3-clang-19.
+    Also, ensure that the required environment variables are set correctly. (e.g.)
+        export MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH=/usr/lib/llvm-19/lib/libclang.so.1
+        export MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH=/usr/lib/llvm-19/lib"""
```
AlanGriffiths, line 33: `libclang-18-dev` is also a viable option and is in `noble/universe`, unlike `libclang-19-dev`.

And, in the future, we'll likely move on to newer versions of clang. So would it be better to specify `libclang-dev` (as the README did) and `python3-clang`?

mattkae: I would, but unfortunately the script falls over right now when we use llvm-20, which would be the default in questing. We should point people to use the version that works in the meantime.

tools/symbols_map_generator/main.py
```diff
@@ -597,4 +607,9 @@ def main():
         f.truncate()
 
 if __name__ == "__main__":
-    main()
+    try:
+        main()
+    except Exception as e:
+        _logger.exception(e)
```
jhenstridge: You can remove this line. The `exception()` method will log the provided message plus the current exception, so you can remove this line

AlanGriffiths (approved):
I guess this is an improvement.

But it still feels fragile: as we're giving instructions based on Ubuntu package names  we could default the `MIR_SYMBOLS_MAP_GENERATOR_CLANG_*` environment variables; and, `libclang-18-dev` is an option.

PR #4415, opened 2025-11-17
Replace use of tmpnam with mkstemp

robert-ancell:
Was generating the warning:
warning: the use of `tmpnam' is dangerous, better use `mkstemp' While the use was in tests and not dangerous replaced to remove compile warnings.

### DIFF
--- a/tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
+++ b/tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
@@ -43,11 +43,8 @@ struct GDesktopFileCache : Test
     static void SetUpTestSuite()
     {
         // Establish the temporary directory
-        char tmp_file_name[] = "/tmp/mir-XXXXXX";
-        int fd = mkstemp(tmp_file_name);
-        close(fd);
-        std::filesystem::path tmp_dir_path {std::filesystem::temp_directory_path() /= tmp_file_name};
-        std::filesystem::create_directories(tmp_dir_path);
+        char tmp_dir_path[] = "/tmp/mir-XXXXXX";
+        mkdtemp(tmp_dir_path);
         std::string desktop_file_directory_name = tmp_dir_path;
         setenv("XDG_DATA_DIRS", desktop_file_directory_name.c_str(), 1);
         applications_dir = desktop_file_directory_name + "/applications";


tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
```diff
@@ -43,8 +43,8 @@ struct GDesktopFileCache : Test
     static void SetUpTestSuite()
     {
         // Establish the temporary directory
-        std::filesystem::path tmp_dir_path {std::filesystem::temp_directory_path() /= std::tmpnam(nullptr)};
-        std::filesystem::create_directories(tmp_dir_path);
+        char tmp_dir_path[] = "/tmp/mir-XXXXXX";
+        mkdtemp(tmp_dir_path);
```
RAOF, line 47: Yup, this is straightforwardly good.

tests/unit-tests/frontend_wayland/test_desktop_file_manager.cpp
```diff
@@ -228,10 +230,10 @@ TEST_F(DesktopFileManager, app_id_will_not_resolve_from_flatpak_info_when_name_i
     std::stringstream flatpak_info_path;
     flatpak_info_path << "/proc/" << PID << "/root/.flatpak-info";
     auto const flatpak_info = flatpak_info_path.str();
-    auto tmp_file_name = std::tmpnam(NULL);
+    char tmp_file_name[] = "/tmp/mir-XXXXXX";
     {
-        std::ofstream tmp_file;
-        tmp_file.open(tmp_file_name);
+        boost::iostreams::file_descriptor fd{mkstemp(tmp_file_name), boost::iostreams::close_handle};
+        boost::iostreams::stream<boost::iostreams::file_descriptor> tmp_file(fd);
```
RAOF, line 236: Because we're closing `tmp_file` when it falls out of scope this isn't any better than `tmpnam()`.

If we kept `tmp_file` alive then this would be an improvement.

tests/miral/static_display_config.cpp
```diff
@@ -163,7 +163,10 @@ struct StaticDisplayConfig : Test
 
 TEST_F(StaticDisplayConfig, nonexistent_config_file_is_no_error)
 {
-    miral::StaticDisplayConfig{tmpnam(nullptr)};
+    char tmp_file_name[] = "/tmp/mir-XXXXXX";
+    int fd = mkstemp(tmp_file_name);
+    close(fd);
+    miral::StaticDisplayConfig{tmp_file_name};
```
RAOF, line 169: This breaks the test - we're no longer testing that a non-existent file is not an error, because we're passing it the name of a file that we've just caused to *exist*.

Saviq:
Ref.: https://github.com/canonical/mir/pull/3497 ;)

AlanGriffiths:
The warning is annoying, but this approach is demonstrably more error prone and doesn't fix the underlying issue

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
robert-ancell: Split out into https://github.com/canonical/mir/pull/4430

tests/miral/static_display_config.cpp
robert-ancell: Fixed in https://github.com/canonical/mir/pull/4431

robert-ancell:
Split out some smaller PRs for the simpler cases.

PR #4429, opened 2025-11-17
Unloading the input platform after all other destruction has happened

mattkae:
This pull request makes it so that the input platform is unloaded after all of the other certain code to avoid issues where destructors are going missing after the library is unloaded.

### DIFF
--- a/tests/unit-tests/platforms/gbm-kms/kms/test_cursor.cpp
+++ b/tests/unit-tests/platforms/gbm-kms/kms/test_cursor.cpp
@@ -302,8 +302,7 @@ struct MesaCursorTest : ::testing::Test
 
     size_t const cursor_side{64};
     MesaCursorTest()
-        : cursor{output_container,
-            mt::fake_shared(current_configuration)}
+        : cursor{output_container,current_configuration}
     {
         using namespace ::testing;
         ON_CALL(mock_drm, drmGetCap(_, DRM_CAP_CURSOR_WIDTH, _))
@@ -324,7 +323,7 @@ struct MesaCursorTest : ::testing::Test
     testing::NiceMock<mtd::MockDRM> mock_drm;
     std::shared_ptr<StubCursorImage> const stub_image = std::make_shared<StubCursorImage>();
     std::shared_ptr<StubKMSOutputContainer> output_container = std::make_shared<StubKMSOutputContainer>();
-    StubCurrentConfiguration current_configuration{*output_container};
+    std::shared_ptr<StubCurrentConfiguration> current_configuration = std::make_shared<StubCurrentConfiguration>(*output_container);
     mgg::Cursor cursor;
 };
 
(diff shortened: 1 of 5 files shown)


tarek-y-ismail (changes requested):
Locally, I get:
```
{"datetime": "2025-11-18T11:28:12Z", "appid": "miral-shell.bin", "event": "sys_shutdown", "level": "WARN", "description": "Mir is shutting down" }terminate called after throwing an instance of 'boost::wrapexcept<
std::system_error>'
  what():  Failed to make context current: EGL_BAD_DISPLAY (0x3008)
Aborted
```

On shutdown, which I assume is also what's causing tests to fail.

I tried fiddling around with the code and I think the cause of this is the change of `input_manager` from a `CachedPtr` to a `shared_ptr` as changing the declaration order didn't fix the crash.

mattkae:
> Locally, I get:
> 
> ```
> {"datetime": "2025-11-18T11:28:12Z", "appid": "miral-shell.bin", "event": "sys_shutdown", "level": "WARN", "description": "Mir is shutting down" }terminate called after throwing an instance of 'boost::wrapexcept<
> std::system_error>'
>   what():  Failed to make context current: EGL_BAD_DISPLAY (0x3008)
> Aborted
> ```
> 
> On shutdown, which I assume is also what's causing tests to fail.
> 
> I tried fiddling around with the code and I think the cause of this is the change of `input_manager` from a `CachedPtr` to a `shared_ptr` as changing the declaration order didn't fix the crash.

Yeah I see this too. The `mgg::Cursor` relies on the `KMSOutputContainer&`, which isn't there when we destruct. This is all very fun :smile: Let me see what I can do...

mattkae:
@tarek-y-ismail I think I fixed it, but let's see what CI says! Can you re-test on your end?

tests/unit-tests/platforms/gbm-kms/kms/test_cursor.cpp
```diff
@@ -323,8 +323,8 @@ struct MesaCursorTest : ::testing::Test
 
     testing::NiceMock<mtd::MockDRM> mock_drm;
     std::shared_ptr<StubCursorImage> const stub_image = std::make_shared<StubCursorImage>();
-    StubKMSOutputContainer output_container;
-    StubCurrentConfiguration current_configuration{output_container};
+    std::shared_ptr<StubKMSOutputContainer> output_container = std::make_shared<StubKMSOutputContainer>();
+    StubCurrentConfiguration current_configuration{*output_container};
```
tarek-y-ismail: Maybe you can make `current_configuration` a `std::shared_ptr` and just pass it below instead of constructing a shared pointer in multiple locations?

mattkae: True enough! And I should also fix atomic kms while I'm at it

PR #4430, opened 2025-11-17
Replace usage of tmpnam() for creating a directory with mkdtemp()

robert-ancell:

### DIFF
--- a/tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
+++ b/tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
@@ -34,7 +34,6 @@ using ::testing::NotNull;
 namespace
 {
 static std::string temp_dir;
-static std::string applications_dir;
 }
 
 struct GDesktopFileCache : Test
@@ -49,7 +48,7 @@ struct GDesktopFileCache : Test
             throw std::runtime_error("failed to create temporary directory");
         setenv("XDG_DATA_DIRS", tmp_dir_path, 1);
         temp_dir = tmp_dir_path;
-        applications_dir = temp_dir + "/applications";
+        auto applications_dir = temp_dir + "/applications";
         mkdir(applications_dir.c_str(), S_IRWXU);
     }
 
@@ -60,7 +59,7 @@ struct GDesktopFileCache : Test
 
     void write_desktop_file(const char* contents, const char* filename)
     {
-        auto filepath = applications_dir + "/" + filename;
+        auto filepath = temp_dir + "/applications/" + filename;
         std::fstream file(filepath, std::fstream::out);
         file << contents;
     }


tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
```diff
@@ -33,6 +33,7 @@ using ::testing::NotNull;
 
 namespace
 {
+static std::string temp_dir;
 static std::string applications_dir;
```
AlanGriffiths: Mutable data at namespace scope should be avoided. These could be private, static members of the test fixture.

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
```diff
@@ -43,11 +44,12 @@ struct GDesktopFileCache : Test
     static void SetUpTestSuite()
     {
         // Establish the temporary directory
-        std::filesystem::path tmp_dir_path {std::filesystem::temp_directory_path() /= std::tmpnam(nullptr)};
-        std::filesystem::create_directories(tmp_dir_path);
-        std::string desktop_file_directory_name = tmp_dir_path;
-        setenv("XDG_DATA_DIRS", desktop_file_directory_name.c_str(), 1);
-        applications_dir = desktop_file_directory_name + "/applications";
+        char tmp_dir_path[] = "/tmp/mir-XXXXXX";
```
AlanGriffiths, line 46: Why are we dropping `filesystem::temp_directory_path()` in favor of hard-coding `/tmp/`? (Do we actually need either of them - AFAICS we just need somewhere with write access?)

robert-ancell: I dropped it as it would make it harder to make the array to pass to `mkdtemp`. In many other tests we have hardcoded `/tmp` and this seems reasonable (extremely unlikely to need to support another directory for temporary files).

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
robert-ancell: If I move this I get weird linking errors - I guess this is why `applications_dir` is here in the first place? I changed to only storing `temp_dir`.

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
jhenstridge: Perhaps it might help to rely on `std::string` rather than C char arrays? Something like:

```c++
std::string tmp_dir = (std::filesystem::temp_directory_path() /= "mir-XXXXXX").string();
ASSERT_THAT(mkdtemp(tmp_dir.data()), NotNull());
```

The writable char array returned by `data()` is guaranteed to be null terminated, so is safe to pass to mkdtemp here. We can't do this with a `std::filesystem::path`, since it doesn't give access to the path's backing array.

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
```diff
@@ -43,24 +43,23 @@ struct GDesktopFileCache : Test
...
+        auto applications_dir = temp_dir + "/applications";
         mkdir(applications_dir.c_str(), S_IRWXU);
     }
 
     void TearDown() override
     {
-        // Remove all files from the temporary directory
-        for (const auto& entry : std::filesystem::directory_iterator(applications_dir))
-            std::filesystem::remove_all(entry.path());
+        std::filesystem::remove_all(temp_dir);
```
jhenstridge, line 57: What happens here if the mkdtemp call failed? I think it's probably going to throw an exception if temp_dir is empty, which is probably not too bad given the test would have already failed at that point.

Maybe it would be worth making this conditional on `!temp_dir.empty()` anyway though?

robert-ancell: From the documentation: "Returns the number of files and directories that were deleted (which may be zero if p did not exist to begin with)." So there should be no issue in that case.

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
robert-ancell: @AlanGriffiths do you have somewhere else we can assume is writable to? Can we assume the tests can write to the current directory, i.e. `char tmp_dir_path[] = "mir-XXXXXX"`?

AlanGriffiths: @robert-ancell, no good suggestions: I was just seeking clarification.

I don't think assuming the current directory is writable is a good plan, and [James's suggestion](https://github.com/canonical/mir/pull/4430#discussion_r2540435777) addresses portability concerns.

AlanGriffiths:
This is failing tests and has not been touched for three weeks. Closing. (Feel free to reopen if/when fixed)

PR #4431, opened 2025-11-17
Replace usage of tmpnam() for creating a non-existant file by creating an empty directory with mkdtemp()

robert-ancell:

### DIFF
--- a/tests/miral/static_display_config.cpp
+++ b/tests/miral/static_display_config.cpp
@@ -20,6 +20,7 @@
 
 #include <mir/test/doubles/mock_display_configuration.h>
 
+#include <filesystem>
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
 #include "yaml-cpp/yaml.h"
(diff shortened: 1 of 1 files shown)


tests/miral/static_display_config.cpp
```diff
@@ -163,7 +163,10 @@ struct StaticDisplayConfig : Test
 
 TEST_F(StaticDisplayConfig, nonexistent_config_file_is_no_error)
 {
-    miral::StaticDisplayConfig{tmpnam(nullptr)};
+    char tmp_dir_path[] = "/tmp/mir-XXXXXX";
+    ASSERT_THAT(mkdtemp(tmp_dir_path), NotNull());
+    miral::StaticDisplayConfig{std::string(tmp_dir_path) + "/display.conf"};
+    rmdir(tmp_dir_path);
```
jhenstridge: It's kind of academic here, but consider using `std::filesystem::remove` or `std::filesystem::remove_all` here. The rmdir shouldn't fail, but we may as well use something that will report a failure.

PR #4432, opened 2025-11-17
Replace usage of tmpnam() for creating a temporary output file by creating an empty directory with mkdtemp()

robert-ancell:

### DIFF
--- a/tests/miral/external_client.cpp
+++ b/tests/miral/external_client.cpp
@@ -20,6 +20,7 @@
 
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
+#include <filesystem>
 #include <fstream>
 
 using namespace testing;


tests/miral/external_client.cpp
```diff
@@ -30,20 +30,28 @@ struct ExternalClient : miral::TestServer
...
+        output = output_dir + "/output";
+
         start_server_in_setup = false;
         add_server_init(external_client);
     }
 
     ~ExternalClient()
     {
         unlink(output.c_str());
+        rmdir(output_dir.c_str());
```
jhenstridge: Perhaps the following would be simpler?

```suggestion
        std::filesystem::remove_all(output_dir);
```

This will report errors as exceptions, which should be fine in this context. That should also address Copilot's complaint (which skips error checking of rmdir).

tests/miral/external_client.cpp
```diff
@@ -30,20 +31,27 @@ struct ExternalClient : miral::TestServer
 {
     ExternalClient()
     {
+        char tmp_dir_path[] = "/tmp/mir-XXXXXX";
+        if (mkdtemp(tmp_dir_path) == nullptr)
+            throw std::runtime_error("failed to create temporary directory");
```
AlanGriffiths, line 36: Do we need a temporary directory? Can we use `tmpfile()` and direct output to the fd?

robert-ancell: I had a [go at this](https://github.com/canonical/mir/tree/external-client-redirect-stdout-stderr) - it would require `ExternalClientLauncher` to be able to redirect stdout/stderr. This is public API and I couldn't find an easy way to do this. Do you have any ideas on how to do this cleanly? It seems adding an `ExternalClientLauncher::redirect_output()` wouldn't be appropriate and it doesn't make sense to modify or duplicate all the launch methods.

AlanGriffiths: These tests are calling via bash - so I was thinking of '>&{fd}'

PR #4434, opened 2025-11-18
Improve Cursor Scale animation

tarek-y-ismail:
## What's new?

- Improves the code for animating the cursor in `scale_temporarily` to avoid jumps in scale.

## How to test

- Run `miral-app`
- Hold down `Ctrl` to invoke Locate Pointer, which uses Cursor Scale.
- Observe as the cursor scales up and down smoothly.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,291 @@
+# Mir Copilot Instructions
+
+This document provides essential knowledge for working with the Mir codebase, including architecture, coding standards, and workflows.
+
+## Overview
+
+Mir is a Wayland compositor library written in C++23 with some Rust components. It provides a stable ABI layer (mirAL) for building shells, handling graphics/input hardware abstraction, and window management.
+
+## Architecture
+
+### Core Layers
+
+- **include/**: Public API headers with minimal implementation details. Interfaces here should not expose platform-specific types.
+- **src/**: Implementation organized by component. Internal headers stay within their component directory.
+- **miral/** (Mir Abstraction Layer): ABI-stable layer in `include/miral/miral/` providing shell-building APIs like `MirRunner`, `WindowManagementPolicy`, window managers, and accessibility features.
+- **Platform backends**: `src/platforms/{gbm-kms,atomic-kms,x11,wayland,eglstream-kms,virtual}` - hardware abstraction layers for different graphics/input systems.
+- **Input platforms**: `evdev` (C++), `evdev-rs` (Rust with CXX bindings)
+
+### Key Components
+
+- **Server** (`src/server/`): Core compositor services (scene, compositor, graphics, frontend_xwayland)
+- **Renderers** (`src/renderers/gl/`): GL-based rendering implementations
+- **Tests** (`tests/`): Unit tests, integration tests, performance tests using gtest/gmock
+
+## Coding Standards
+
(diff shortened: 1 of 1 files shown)


src/include/server/mir/time/alarm_factory.h
```diff
@@ -72,6 +72,9 @@ class AlarmFactory
     auto create_repeating_alarm(std::function<void()> const& callback, std::chrono::milliseconds repeat_delay)
         -> std::shared_ptr<Alarm>;
 
+    auto create_repeating_alarm(std::function<bool()> const& callback, std::chrono::milliseconds repeat_delay)
+        -> std::shared_ptr<Alarm>;
```
AlanGriffiths: Is this overload actually needed? At first sight the alarm can be dropped entirely by resetting `animation_alarm` (rather than leaving it around but not scheduling it).

At second sight, we are already resetting `animation_alarm`, so I don't thin we need this at all.

src/miral/cursor_scale.cpp
```diff
@@ -65,57 +68,59 @@ struct miral::CursorScale::Self
         auto const start_scale = s->scale;
         auto const end_scale = s->scale * scale_multiplier;
         auto const ml = main_loop.lock();
+        auto const animation_duration = wind_time + duration + wind_time;
 
-        s->animation_alarm = ml->create_repeating_alarm(
+        std::function<bool()> animation_callback{
```
AlanGriffiths:
```suggestion
        std::function<void()> animation_callback{
```

tarek-y-ismail:
>  When I try this the cursor shrinks after each use

Do you mean with `LocatePointer`? That's the intended and usual behavior

AlanGriffiths:
> > When I try this the cursor shrinks after each use
> 
> Do you mean with `LocatePointer`? That's the intended and usual behavior

Why? I don't want the cursor to get harder to see each time I "locate" it! I want it returned to the original size.

tarek-y-ismail:
> Why? I don't want the cursor to get harder to see each time I "locate" it! I want it returned to the original size.

Ah, that might be the bug I just fixed. Can you try once more?

src/miral/cursor_scale.cpp
```diff
@@ -65,57 +68,57 @@ struct miral::CursorScale::Self
...
+                        s->animation_alarm->cancel();
+                        s->animation_alarm = nullptr;
+                    }
+
+                    // Reset scale just in case
+                    this->scale(start_scale);
+                    return;
+                }
+
                 auto const scale = [&] -> std::optional<float>
```
AlanGriffiths: I don't think there's a path that returns nullopt

src/miral/cursor_scale.cpp
```diff
@@ -65,57 +68,57 @@ struct miral::CursorScale::Self
...
                 {
                     if (time < wind_time)
                     {
                         auto const t = static_cast<float>(time.count()) / wind_time.count();
```
AlanGriffiths: Extracting the counts seems unnecessarily verbose:

```suggestion
                        auto const t = 1.0*time / wind_time;
```

src/miral/cursor_scale.cpp
```diff
@@ -65,57 +68,56 @@ struct miral::CursorScale::Self
...
+                    this->scale(start_scale);
+                    return;
+                }
+
+                auto const scale = [&]
                 {
                     if (time < wind_time)
                     {
-                        auto const t = static_cast<float>(time.count()) / wind_time.count();
+                        auto const t = (1.0f * time) / wind_time;
```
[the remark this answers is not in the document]
tarek-y-ismail: I don't think I will.

Issue #3778, opened 2025-02-25
[SSD] Incorrectly sizes some windows


AlanGriffiths:
When SSDs are applied we need to adjust the size and size constraints as they are relative to the content area, but window management assumes they apply to the full area.

We have code that does this for X11 windows in `AbstractShell::create_surface()` (c.f. #3567), but that depends on the window spec having width and height set. That may not be the case, so we need to revisit this adjustment when a buffer is submitted. Likely in `AbstractShell::surface_ready()`.

AlanGriffiths:
One of the issues identified in https://github.com/lxqt/lxqt-wayland-session/pull/40#issuecomment-2682727220

PR #4468, opened 2025-11-26
Add tests to check how different combinations of surface spec parameters affect SSD size adjustments

### LINKED ISSUE #3781: Improve unit coverage of `msh::AbstractShell::create_surface`
> No, but we should have

_Originally posted by @AlanGriffiths in https://github.com/canonical/mir/pull/3780#discussion_r1971334776_

tarek-y-ismail:
Closes #3781 (hopefully)

## What's new?
- Adds two new parametric tests which try different combinations of width, height, min width/height, and max width/height being set or unset with decorations enabled and disabled, in hopes of notifying us when bugs similar to #3778 pop up.

## How to test

Run `mir_unit_tests --gtest_filter=*SsdSizeConstraints*`

## Checklist

- [x] Tests added and pass

### DIFF
--- a/tests/unit-tests/scene/test_abstract_shell.cpp
+++ b/tests/unit-tests/scene/test_abstract_shell.cpp
@@ -128,6 +128,8 @@ struct MockSurfaceFactory : public ms::SurfaceFactory
 struct MockDecorationManager : public msh::decoration::NullManager
 {
     MOCK_METHOD(void, decorate, (std::shared_ptr<ms::Surface> const&));
+    MOCK_METHOD(
+        geom::Size, compute_size_with_decorations, (geom::Size content_size, MirWindowType type, MirWindowState state));
 };
 
 using NiceMockWindowManager = NiceMock<mtd::MockWindowManager>;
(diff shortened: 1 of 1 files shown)


tests/unit-tests/scene/test_abstract_shell.cpp
```diff
@@ -1090,3 +1092,201 @@ TEST_F(AbstractShell, as_focus_controller_emits_input_device_state_event_on_focu
...
+INSTANTIATE_TEST_SUITE_P(
+    SsdSizeConstraints,
+    SsdSizeConstraintsTest,
+    testing::Values(
+        SsdConstraintsParams{.set_width = false, .set_height = false, .set_min_size = false, .set_max_size = false},
+        SsdConstraintsParams{.set_width = true, .set_height = true, .set_min_size = true, .set_max_size = true},
+        SsdConstraintsParams{.set_width = true, .set_height = false, .set_min_size = false, .set_max_size = false},
+        SsdConstraintsParams{.set_width = false, .set_height = true, .set_min_size = false, .set_max_size = false},
+        SsdConstraintsParams{.set_width = false, .set_height = false, .set_min_size = true, .set_max_size = false},
+        SsdConstraintsParams{.set_width = false, .set_height = false, .set_min_size = false, .set_max_size = true}),
```
AlanGriffiths, line 1283: Slightly curious: why, when there are four boolean options, are there six test instantiations (and not sixteen)?

tarek-y-ismail: The original bug would've required both width and height to be set, which is tested by the first four combinations. Other than that, `adjust_size_constraints_for_ssd` tests each of the parameters individually, so we don't need to test all combinations.

Issue #4406, opened 2025-11-06
Deadlock during shutdown


AlanGriffiths:
### Version

v2.23.0

### Environment

OS: Ubuntu 25.10

### What happened? What did you expect to happen?

After testing XFCE/Miriway closing the server lead to a deadlock

### Relevant log output

In `WaylandExecutor::drain()` we call `workqueue.clear()` whilst holding the mutex.

This leads to destroying a `NotifyingMappableBackedShmBuffer` whose destructor calls `on_release()` - and that tries to `executor->spawn()` releasing the buffer.

`WaylandExecutor::spawn()` then blocks because the mutex is locked.

```
(gdb) bt
#0  futex_wait (futex_word=0x59a23f67f4c0, expected=2, private=0) at ../sysdeps/nptl/futex-internal.h:146
#1  __GI___lll_lock_wait (futex=futex@entry=0x59a23f67f4c0, private=0) at ./nptl/lowlevellock.c:49
#2  0x00007c92c92a75e1 in lll_mutex_lock_optimized (mutex=0x59a23f67f4c0) at ./nptl/pthread_mutex_lock.c:48
#3  ___pthread_mutex_lock (mutex=0x59a23f67f4c0) at ./nptl/pthread_mutex_lock.c:93
#4  0x00007c92c88bb2d4 in __gthread_mutex_lock (__mutex=0x59a23f67f4c0)
    at /usr/include/x86_64-linux-gnu/c++/15/bits/gthr-default.h:795
#5  std::mutex::lock (this=0x59a23f67f4c0) at /usr/include/c++/15/bits/std_mutex.h:115
#6  0x00007c92c88bb358 in std::lock_guard<std::mutex>::lock_guard (this=0x7ffed25cb0c8, __m=...)
    at /usr/include/c++/15/bits/std_mutex.h:252
#7  0x00007c92c8b4fcb3 in mir::frontend::WaylandExecutor::State::enqueue (this=0x59a23f67f4c0, work=...)
    at /home/alan/mir/src/server/frontend_wayland/wayland_executor.cpp:75
#8  0x00007c92c8b4f5ed in mir::frontend::WaylandExecutor::spawn (this=0x59a23f5cd740, work=...)
    at /home/alan/mir/src/server/frontend_wayland/wayland_executor.cpp:327
#9  0x00007c92c8b8d04e in operator() (__closure=0x7c9274000e90)
    at /home/alan/mir/src/server/frontend_wayland/wl_surface.cpp:414
[...]

PR #4469, opened 2025-11-26
WaylandExecutor::spawn() - do not try to lock the queue unless the execution state is Running

AlanGriffiths:
Closes: #4406

## What's new?

When new work is enqueued first check that the queue is running before locking the queue.

This avoids a deadlock when the queue is being drained

## How to test

Run an XFCE/Miriway session and close it.  There should be no deadlock

### DIFF
--- a/src/server/frontend_wayland/wayland_executor.cpp
+++ b/src/server/frontend_wayland/wayland_executor.cpp
@@ -136,6 +136,13 @@ public:
 private:
     static thread_local bool on_wayland_thread;
     std::mutex mutex;
+    /*
+     * `state` is atomic because it is used in a double-checked locking pattern in `enqueue()`.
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
Note for reviewers: I have been unable to reproduce the [linked error](https://github.com/canonical/mir/issues/4406) today. (It was easily reproducible previously, and I am not sure what has changed - presumably some XFCE update has changed the sequence or timing of events.)

However, I think the potential for deadlock still remains and is fixed by the proposed change.

I don't, however see a way to introduce a test without a significant refactoring of WaylandExecutor (so that `drain()` is accessible from the test) and that risks introducing new bugs.

mattkae (approved):
Makes sense. Consider adding some comments per copilot's suggestion.

PR #4475, opened 2025-11-27
Tidy up some of our "common" GL code

AlanGriffiths:
Tidy up some of our GL code and modernise associated build configuration.

Eliminates circular dependencies by moving `report_exception` to `mircore`; moving `mirgl` & `mirrenderergl` to `mirplatform`; and dropping the `libmirrenderer-dev` & `mir-renderer-gl-dev` packages (rolling them into `mirplatform-dev`).

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -149,8 +149,14 @@ Depends: libmirplatform33 (= ${binary:Version}),
          libboost-program-options-dev,
          ${misc:Depends},
          libgbm-dev,
-Breaks: libmirplatform (<< 0.6)
-Replaces: libmirplatform (<< 0.6)
+Breaks: libmirplatform (<< 0.6),
+        libmirrenderer-dev (<< 2.24),
+        mir-renderer-gl-dev (<< 2.24)
+Replaces: libmirplatform (<< 0.6),
+          libmirrenderer-dev (<< 2.24),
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
TICS is complaining about code that landed months ago. (There is a legitimate concern, but not a new one)

AlanGriffiths:
> Do we need to do anything special to "sunset" the old package?

@copilot WDYT?

AlanGriffiths:
> Do we need to do anything special to "sunset" the old package?

Good catch! I'm no expert on debian packaging (I've failed to internalize what `Breaks:` etc really mean). But copilot guessed something very similar to me so I pushed it

AlanGriffiths:
> it will break miracle, but I can fix

I'm curious: how does it break miracle? The pkg-config changes?

mattkae:
> > it will break miracle, but I can fix
> 
> I'm curious: how does it break miracle? The pkg-config changes?

I install libmirrenderer-dev, so I'll have to just install the other package. It is no big deal though

mattkae:
Heh, this is breaking me and I can't figure out why! https://github.com/miracle-wm-org/miracle-wm/actions/runs/20046857706/job/57494146623?pr=734

There might be a poor dependency somewhere...

PR #4497, opened 2025-12-04
Fix error message when clang import is not installed and look for clang libs if environment variables are unset

AlanGriffiths:
A couple of tweaks to simplify using check-mirserver-symbols-map et al

1. better error message when `pip3 install clang
2. Don't rely on env variables when files can be found

### DIFF
--- a/tools/symbols_map_generator/main.py
+++ b/tools/symbols_map_generator/main.py
@@ -479,7 +479,9 @@ def main():
             clang.cindex.Config.set_library_file(found_so)
             _logger.info(f"Using found libclang.so.1 at: {found_so}")
         else:
-            _logger.error("Expected MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH to be defined in the environment")
+            _logger.error("Could not find libclang.so.1 in the environment. "
+                          "You may want to set MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH "
+                          "if libclang.so.1 is manually installed in a non-standard directory.")
             exit(1)
 
     if 'MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH' in os.environ:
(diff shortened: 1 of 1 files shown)


tools/symbols_map_generator/main.py
```diff
@@ -456,16 +456,63 @@ def main():
...
+            if os.path.isfile(path):
+                found_so = path
+                _logger.debug(f"Found libclang.so.1 at: {path}")
+                break
+
+        if found_so is not None:
+            clang.cindex.Config.set_library_file(found_so)
+            _logger.info(f"Using found libclang.so.1 at: {found_so}")
+        else:
+            _logger.error("Expected MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH to be defined in the environment")
```
mattkae:
```suggestion
            _logger.error("Could not find libclang.so.1 in the environment. You may want to set MIR_SYMBOLS_MAP_GENERATOR_CLANG_SO_PATH if libclang.so.1 is manually installed in a non-standard directory.")
```

tools/symbols_map_generator/main.py
```diff
@@ -456,16 +456,63 @@ def main():
...
+                        found_lib = lib_dir
+                        _logger.debug(f"Found clang library directory at: {lib_dir}")
+                        break
+            if found_lib:
+                break
+        if found_lib is not None:
+            clang.cindex.Config.set_library_path(found_lib)
+            _logger.info(f"Using found clang library directory at: {found_lib}")
+        else:
+            _logger.error("Expected MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH to be defined in the environment")
```
mattkae:
```suggestion
            _logger.error("Could not find clang library directory. You may want to set MIR_SYMBOLS_MAP_GENERATOR_CLANG_LIBRARY_PATH if clang is not installed in the standard directory.")
```

Issue #4488, opened 2025-12-03
ASan detects heap use after free when miral-app's tiling window manager is used


tarek-y-ismail:
### Version

main

### Environment

Mir under ASan.

### What happened? What did you expect to happen?

When running `miral-app --window-manager=tiling` on an ASan build, Mir crashes immediately with a heap use after free error.

### Relevant log output

```shell
==482738==ERROR: AddressSanitizer: heap-use-after-free on address 0x7411479cddd0 at pc 0x77a14afa6b56 bp 0x73a11048c050 sp 0x73a11048c048
READ of size 8 at 0x7411479cddd0 thread T28
    #0 0x77a14afa6b55 in auto (anonymous namespace)::WaylandApp::remove(zwlr_foreign_toplevel_handle_v1*)::'lambda'(auto const&)::operator()<(anonymous namespace)::ToplevelInfo>(auto const&) 
const /home/tarek/code-projects/mir/src/miral/application_switcher.cpp:379:28
    #1 0x77a14afa6b0c in auto std::__invoke_impl<bool, (anonymous namespace)::WaylandApp::remove(zwlr_foreign_toplevel_handle_v1*)::'lambda'(auto const&)&, (anonymous namespace)::ToplevelInfo
&>(std::__invoke_other, (anonymous namespace)::WaylandApp::remove(zwlr_foreign_toplevel_handle_v1*)::'lambda'(auto const&)&, (anonymous namespace)::ToplevelInfo&) /usr/lib/gcc/x86_64-linux-gn
u/15/../../../../include/c++/15/bits/invoke.h:63:14
    #2 0x77a14afa6adc in std::__invoke_result<auto, (anonymous namespace)::ToplevelInfo&>::type std::__invoke<(anonymous namespace)::WaylandApp::remove(zwlr_foreign_toplevel_handle_v1*)::'lam
bda'(auto const&)&, (anonymous namespace)::ToplevelInfo&>(auto&&, (anonymous namespace)::ToplevelInfo&) /usr/lib/gcc/x86_64-linux-gnu/15/../../../../include/c++/15/bits/invoke.h:98:14
    #3 0x77a14afa69be in T std::ranges::__find_if_fn::operator()<__gnu_cxx::__normal_iterator<(anonymous namespace)::ToplevelInfo*, std::vector<(anonymous namespace)::ToplevelInfo, std::alloc
[...]

PR #4509, opened 2025-12-08
Stop the app switcher only when `Self` is destroyed.

tarek-y-ismail:
Fixes a bug introduced in 91844b9560965fc83a64ba1d91bcf631323939a2 where the app switcher was assumed to act like a `shared_ptr`. Instead, on destruction, it would call `self->stop()` which should have been called in `~Self` instead.

Related: #4488 (found this bug when investigating that)

## What's new?

- Makes the app switcher truly copyable by making `self->stop()` get called only when all instances of `self` are destroyed.

## How to test

- Try using the application switcher before and after applying this patch.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/miral/application_switcher.cpp
+++ b/src/miral/application_switcher.cpp
@@ -688,6 +688,11 @@ public:
             mir::log_error("Failed to create shutdown notifier");
     }
 
+    ~Self()
+    {
+        stop();
+    }
+
     void run_client(wl_display* display)
     {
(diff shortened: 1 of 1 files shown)


tarek-y-ismail:
Food for thought: Is there an easy way to test the app switcher working properly? This bug lurked in our code for a bit more than two weeks without anyone noticing!

PR #4525, opened 2025-12-10
Hack for clang 18.1 using <expected>

AlanGriffiths:
clang 18.1 sets __cpp_concepts=201907L
libstdc++ requires __cpp_concepts=202002L in \<expected>

### DIFF
--- a/src/server/frontend_wayland/CMakeLists.txt
+++ b/src/server/frontend_wayland/CMakeLists.txt
@@ -68,7 +68,7 @@ set(
   surface_registry.cpp          surface_registry.h
 )
 
-if("${CMAKE_CXX_COMPILER}" MATCHES "clang" AND $(CMAKE_CXX_COMPILER_VERSION) VERSION_LESS_EQUAL 18.1)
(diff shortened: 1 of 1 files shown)


src/server/frontend_wayland/CMakeLists.txt
```diff
@@ -68,6 +68,13 @@ set(
   surface_registry.cpp          surface_registry.h
 )
 
+if("${CMAKE_CXX_COMPILER}" MATCHES "clang" AND $(CMAKE_CXX_COMPILER_VERSION) VERSION_LESS_EQUAL 18.1)
```
Saviq: `_ID` is the canonical way to check the compiler, also curly brackets, no?
```suggestion
if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS_EQUAL 18.1)
```

AlanGriffiths: We use both `CMAKE_CXX_COMPILER` and `CMAKE_CXX_COMPILER_ID`, I just found the wrong one first and copied it

PR #4537, opened 2025-12-11
packaging: flip Rust build

Saviq:
## What's new?

Flipped the `MIR_ENABLE_RUST` to default `ON`, and package it except for Debian/Ubuntu that waits for a cxx update.

## How to test

CI

## Checklist

- [x] Tests added and pass

### DIFF
--- a/spread/build/alpine/task.yaml
+++ b/spread/build/alpine/task.yaml
@@ -53,7 +53,7 @@ execute: |
         -DCMAKE_C_COMPILER_LAUNCHER=ccache
         -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
     )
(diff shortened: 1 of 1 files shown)


spread/build/fedora/task.yaml
```diff
@@ -41,7 +41,7 @@ execute: |
       --plugin-option=ccache:hashdir=False \
       --plugin-option=ccache:show_stats=True \
       --with debug \
-      --with run_tests \
+      --with check \
```
Conan-Kudo: This is not necessary anymore since the default is to run them. You only need to add an option for the non-default case (ie. `--without check`).

Saviq:
```suggestion
```

Conan-Kudo (approved):

Saviq:
> LGTM 👍

Looking into why Alpine 3.20 is unhappy.

Saviq:
> Looking into why Alpine 3.20 is unhappy.

And fixed now :+1:

PR #4544, opened 2025-12-14
Use PROJECT_VERSION instead of the old MIR_VERSION variables

mattkae:
## What's new?
- Mir's version strings are broken, so I am unable to inspect the version in both CMake and code in Miracle
- Mir expects to replace @VARIABLE_NAME@ in spaces, so we should ensure that that can still happen

Caused by: https://github.com/canonical/mir/commit/e79ad7107880896700274e06405b334ef9bebd51

## Checklist

- [x] Tests added and pass


mattkae:
> I say we should use `PROJECT_VERSION*` in the templates instead.

Fixed!

Saviq (approved):
Of course 🤦, sorry.

2.25.1 incoming.

PR #4572, opened 2025-12-16
Replace atoi with std::from_chars (#4569)

### LINKED ISSUE #4569: Replace `atoi` with `std::from_chars` with proper error handling
`atoi` has a more idiomatic C++ alternative, `std::from_chars`. We should use that.

Example:
https://github.com/canonical/mir/blob/7a9370befcace079fcc6d2103d5a33218462f86e/src/server/frontend_xwayland/xwayland_cursors.cpp#L123-L126

DiegoOst:
- Use std::from_chars instead of atoi to use a more modern conversion
- Add proper validation for integer conversion
- Fixes issue #4569"

Closes #4569

Related: https://github.com/canonical/mir/issues/4569

## What's new?

Use std::from_chars instead of atoi to use modern C++ and add proper error handling for it.
In case of errors the default values are used.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/examples/client/mir_shell_demo.cpp
+++ b/examples/client/mir_shell_demo.cpp
@@ -74,6 +74,10 @@ void parse_options(int argc, char* argv[])
                 {
                     main_width = parsed_value;
                 }
+                else
+                {
+                    printf("Could not parse option %s with arg %s\n", long_options[option_index].name, optarg);
+                }
             }
             else if (option_main_height == long_options[option_index].name)
             {
@@ -82,6 +86,10 @@ void parse_options(int argc, char* argv[])
                 if (ec == std::errc())
                 {
                     main_height = parsed_value;
+                }
+                else
+                {
+                    printf("Could not parse option %s with arg %s\n", long_options[option_index].name, optarg);
                 }
             }
             else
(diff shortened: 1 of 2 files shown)


src/server/frontend_xwayland/xwayland_cursors.cpp
```diff
@@ -122,7 +124,14 @@ auto mf::XWaylandCursors::Loader::get_xcursor_size() -> int
     int result = 0;
     if (size_env_var_string)
     {
-        result = atoi(size_env_var_string);
+        int parsed_value = 0;
+        auto const [end, err] = std::from_chars(size_env_var_string,
+                                                size_env_var_string + strlen(size_env_var_string),
+                                                parsed_value);
+        if ((err == std::errc{}))
```
robert-ancell: Extra parentheses here.

examples/client/mir_shell_demo.cpp
```diff
@@ -67,11 +68,21 @@ void parse_options(int argc, char* argv[])
         case 0:
             if (option_main_width == long_options[option_index].name)
             {
-                main_width = atoi(optarg);
+                int32_t parsed_value;
```
robert-ancell: You haven't initialized the value here (and below) like in the third case. Even though this value is only accessed on success it might be safer to set it to zero.

DiegoOst:
@robert-ancell thanks, I have both pushed a commit for the small fixes and added another commit for the logs in case it is decided to keep them

robert-ancell (approved):
Looks good, thanks!

PR #4579, opened 2025-12-18
`input::CursorObserver`: Add `image_set_to` observation

RAOF:
This allows `CursorObserver`s to monitor when the cursor image changes.

Adds a new virtual method to `CursorObserver`, which breaks ABI, so this also bumps the mirserver ABI as that hasn't been done yet this release.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -67,7 +67,7 @@ Vcs-Git: https://github.com/canonical/mir
 
 #TODO: Packaging infrastructure for better dependency generation,
 #      ala pkg-xorg's xviddriver:Provides and ABI detection.
-Package: libmirserver66
+Package: libmirserver67
 Section: libs
 Architecture: linux-any
 Multi-Arch: same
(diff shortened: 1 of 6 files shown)


src/server/input/cursor_observer_multiplexer.cpp
```diff
@@ -39,3 +39,8 @@ void mi::CursorObserverMultiplexer::pointer_unusable()
 {
     for_each_observer(&CursorObserver::pointer_unusable);
 }
+
+void mi::CursorObserverMultiplexer::image_set_to(std::shared_ptr<graphics::CursorImage> cursor)
+{
+    for_each_observer(&CursorObserver::image_set_to, std::move(cursor));
```
AlanGriffiths, line 45: Is there any value to `std::move()` here?

Saviq: Approved or?

AlanGriffiths: I approved: I think this is just a distraction

PR #4581, opened 2025-12-18
Fix code and comment disagreement

AlanGriffiths:


Saviq (approved):
Not sure there's really disagreement between comment and code, but ACK on variable case name :)

AlanGriffiths:
> Not sure there's really disagreement between comment and code

OK, disagreement between comment and the variable name then

PR #4582, opened 2025-12-18
Changing %d to %u when logging the output of wl_resource_get_id

### LINKED ISSUE #4570: Use `%u` when logging the output of `wl_resource_get_id` instead of `%d`
`wl_resource_get_id` returns a `uint32_t`, but `%d` is the specifier for signed integers, while `%u` is the specifier for unsigned integers.

magedss:
Closes #4570
Updated the logging format for wl_resource_get_id() calls from %d to %u, ensuring that the correct format specifier is used for the uint32_t return type of wl_resource_get_id.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -3,7 +3,9 @@ name: PreCommit
 on:
   pull_request:
   push:
-    branches: [main]
+    branches:
+    - main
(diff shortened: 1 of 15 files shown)


magedss:
Thanks! Changes applied

AlanGriffiths:
The dictionary issues are unrelated. Landing

PR #4590, opened 2025-12-24
feature: replace strerror with thread safe version( strerror_r )

### LINKED ISSUE #4568: Replace `strerror` with `strerror_r`
We use `strerror` in multiple places that may run on multiple threads. And while having multiple threads error out at the same time is quite rare, it is not impossible. In that case, different concurrent invocations of `strerror` might override each other, leaving us without valuable data.

A better approach would be to use `strerror_r` and pass it a buffer, an even better approach would be to wrap all of that in a more idiomatic C++ function.

aliihsancengiz:
Closes #4568

Related: #4568  

## What's new?

Replace strerror with thread safe version of strerror_r

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/common/errno_utils.cpp
+++ b/src/common/errno_utils.cpp
@@ -13,35 +13,15 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
+#undef _GNU_SOURCE
+#define _POSIX_C_SOURCE 200112L
 
 #include <mir/errno_utils.h>
 
 #include <cstring>
 
-namespace
-{
-// Since we are normalizing between GNU/POSIX impl, one of these functions will be unused at compile time
-
-[[maybe_unused]]
-inline auto strerror_r_to_cstr(int ret, const char* buf) noexcept
-{
-    // POSIX
-    return ret ? "Unknown error" : buf;
-}
-
-[[maybe_unused]]
-inline auto strerror_r_to_cstr(char* ret, const char* /*buf*/) noexcept
-{
-    // GNU
-    return ret ? ret : "Unknown error";
-}
-
-}
-
 auto mir::errno_to_cstr(int err) noexcept -> const char*
 {
     thread_local char buf[256] = {};
-
-    auto ret = ::strerror_r(err, buf, sizeof(buf));
-    return strerror_r_to_cstr(ret, buf);
+    return ::strerror_r(err, buf, sizeof(buf)) ? "Unknown error" : buf;
 }


examples/mir_demo_server/server_example_test_client.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "server_example_test_client.h"
+#include "mir/errno_utils.h"
```
tarek-y-ismail:
```suggestion
#include <mir/errno_utils.h>
```

This occurs with every inclusion of this header. I won't point it out in every file since it would be too annoying.
Checking the style guide, we forgot to update it to point this out. Opened #4597 to fix this.

src/miral/launch_app.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "launch_app.h"
+#include "mir/errno_utils.h"
```
tarek-y-ismail:
```suggestion
#include <mir/errno_utils.h>
```

include/common/mir/errno_utils.h
```diff
@@ -0,0 +1,77 @@
...
+#ifndef MIR_ERRNO_UTILS_H_
+#define MIR_ERRNO_UTILS_H_
+
+#include <cerrno>
+#include <cstring>
+#include <string>
+
+namespace mir
+{
+    namespace detail {
```
tarek-y-ismail: I believe we don't follow the `detail` pattern in the codebase. Can you split the implementation details into their own source file?

src/platforms/atomic-kms/server/kms/atomic_kms_output.cpp
```diff
@@ -22,6 +22,7 @@
 #include <mir/graphics/gamma_curves.h>
 #include <mir_toolkit/common.h>
 #include "kms-utils/kms_connector.h"
+#include "mir/errno_utils.h"
```
tarek-y-ismail:
```suggestion
#include <mir/errno_utils.h>
```

tests/unit-tests/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ set(
   test_shared_library_prober.cpp
   test_lockable_callback.cpp
   test_module_deleter.cpp
-    test_posix_rw_mutex.cpp
+  test_posix_rw_mutex.cpp
```
tarek-y-ismail: While this is a correct fix. I'm not sure if we should sneak it in with this PR

include/common/mir/errno_utils.h
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    // Thread-safe errno to char* conversion.
+    inline const char* errno_to_cstr(int err) noexcept {
+        thread_local char buf[256] = {};
+
+        // Call strerror_r and normalize both GNU/POSIX return types.
+        auto ret = ::strerror_r(err, buf, sizeof(buf));
+        return detail::strerror_r_to_cstr(ret, buf);
+    }
```
tarek-y-ismail: Depending on whether or not the compiler chooses to inline this function, we may be shooting ourselves in the foot by accessing stack-allocated memory after the function returns.

In the string case above, that is alright because the string copies the buffer before it goes out of scope.

It's probably okay to always use `errno_to_string` and just call `c_str()` on the returned string. I don't think the extra allocation will be an issue since errors are not a hot path.

include/common/mir/errno_utils.h
```diff
@@ -0,0 +1,77 @@
...
+        inline const char* strerror_r_to_cstr(int /*ret*/, const char* buf) noexcept {
+            return buf; // POSIX
+        }
+
+        inline const char* strerror_r_to_cstr(char* ret, const char* /*buf*/) noexcept {
+            return ret ? ret : "Unknown error"; // GNU
+        }
+
+
+    } // namespace detail
```
AlanGriffiths: I think this detail would be better encapsulated in a .cpp file

include/common/mir/errno_utils.h
```diff
@@ -0,0 +1,77 @@
...
+
+
+    // Thread-safe errno to string conversion.
+    inline std::string errno_to_string(int err) {
+        thread_local char buf[256] = {};
+
+        // Call strerror_r and normalize both GNU/POSIX return types.
+        auto ret = ::strerror_r(err, buf, sizeof(buf));
+        return detail::strerror_r_to_string(ret, buf);
+    }
```
AlanGriffiths: Do we need this? I can see some uses, but `errno_to_cstr()` seems adequate in all these cases.

include/common/mir/errno_utils.h
AlanGriffiths: The implementation would be better encapsulated in a .cpp file

include/common/mir/errno_utils.h
```diff
@@ -0,0 +1,77 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_ERRNO_UTILS_H_
+#define MIR_ERRNO_UTILS_H_
+
+#include <cerrno>
```
AlanGriffiths: Why do we include `<cerrno>`?

include/common/mir/errno_utils.h
tarek-y-ismail: Looks like a typo

include/common/mir/errno_utils.h
tarek-y-ismail: That was my initial opinion, but I think this uses stack memory after returning. See [this comment](https://github.com/canonical/mir/pull/4590#discussion_r2660930268)

include/common/mir/errno_utils.h
AlanGriffiths: > Depending on whether or not the compiler chooses to inline this function, we may be shooting ourselves in the foot by accessing stack-allocated memory after the function returns.

1. whether the compiler inlines makes no difference
2. we are not accessing stack memory, we are accessing thread local storage (with static duration)

> It's probably okay to always use `errno_to_string` and just call `c_str()` on the returned string. I don't think the extra allocation will be an issue since errors are not a hot path.

While the extra allocation is unlikely to be an issue, there's no need for it.

tarek-y-ismail: > we are not accessing stack memory, we are accessing thread local storage (with static duration)

Aha, that flew right over my head :)

include/common/mir/errno_utils.h
tarek-y-ismail: Ok, after Alan clearing up `thread_local` to me. I agree with this, we can just use the `cstr` version and wrap in in `std::string{...}` in all the places that require it.

aliihsancengiz:
Hey @AlanGriffiths and @tarek-y-ismail , thanks for the review. 
I have rebased and addressed comments. Please find the first iteration.

- Got rid of `errno_to_string`
- Splitted implementation into `.cpp` file
- Some include tweaking
- exported `mir::errno_to_cstr*;` for symbol map, i did this manually, but not sure that's the right way.


Thanks.

src/server/console/linux_virtual_terminal.cpp
```diff
@@ -22,6 +22,10 @@
 #include <mir/fd.h>
 #include <mir/graphics/display_report.h>
 #include <mir/graphics/event_handler_register.h>
+#include <mir/fd.h>
+#include <mir/emergency_cleanup_registry.h>
+#include "ioctl_vt_switcher.h"
+#include <mir/errno_utils.h>
```
AlanGriffiths:
```suggestion
#include <mir/errno_utils.h>
```

src/common/symbols.map
```diff
@@ -373,6 +373,7 @@ global:
     mir::dispatch::ThreadedDispatcher::remove_thread*;
     mir::dispatch::epoll_to_fd_event*;
     mir::dispatch::fd_event_to_epoll*;
+    mir::errno_to_cstr*;
```
AlanGriffiths: This is in the wrong stanza (the symbol didn't exist in Mir-2.22).

You need to add it in a new `MIR_COMMON_2.26` stanza. Vis:

```
MIR_COMMON_2.26 {
global:
  extern "C++" {
    mir::errno_to_cstr*;
  };
} MIR_COMMON_INTERNAL_2.24;
```

(That may look confusing, but there's a pre-existing issue with abusing "`_INTERNAL_`". We can't address that without breaking ABI.)

tests/unit-tests/test_errno_utils.cpp
```diff
@@ -0,0 +1,87 @@
+#include <gtest/gtest.h>
+
+#include <atomic>
+#include <cstring>
+#include <string>
+#include <thread>
+
+#include <mir/errno_utils.h>
+
+TEST(ErrnoToCstr, ReturnsNonNullAndNonEmpty) {
```
AlanGriffiths: Project style is `snake_case` for test names and braces opened on the next line.

```suggestion
TEST(ErrnoToCstr, returns_non_null_and_non_empty) 
{
```

tests/unit-tests/test_errno_utils.cpp
```diff
@@ -0,0 +1,87 @@
+#include <gtest/gtest.h>
```
AlanGriffiths, line 17: Source files should start with a copyright notice

src/server/console/linux_virtual_terminal.cpp
```diff
@@ -19,9 +19,13 @@
 
 #include <mir/constexpr_utils.h>
 #include <mir/emergency_cleanup_registry.h>
+#include <mir/errno_utils.h>
 #include <mir/fd.h>
 #include <mir/graphics/display_report.h>
 #include <mir/graphics/event_handler_register.h>
+#include <mir/fd.h>
+#include <mir/emergency_cleanup_registry.h>
+#include "ioctl_vt_switcher.h"
```
AlanGriffiths: Why?

tests/unit-tests/test_errno_utils.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+#include <gtest/gtest.h>
+
+#include <atomic>
+#include <cstring>
+#include <string>
+#include <thread>
+
+#include <mir/errno_utils.h>
+TEST(ErrnoToCstr, returns_non_null_and_non_empty) {
```
AlanGriffiths: Project style is to have the opening brace on the next line

src/server/console/linux_virtual_terminal.cpp
aliihsancengiz: That was not intended; indeed, it reverted.

include/common/mir/errno_utils.h
```diff
@@ -0,0 +1,33 @@
...
+#define MIR_ERRNO_UTILS_H_
+
+#include <string>
+
+namespace mir
+{
+// This overload set normalizes both to char*
+// inline const char* strerror_r_to_cstr(int ret, const char* buf) noexcept;
+// inline const char* strerror_r_to_cstr(char* ret, const char* buf) noexcept;
+
```
AlanGriffiths: Confusing comment

aliihsancengiz: Indeed, a leftover from clean-up, Thanks.

include/common/mir/errno_utils.h
```diff
@@ -0,0 +1,29 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIR_ERRNO_UTILS_H_
+#define MIR_ERRNO_UTILS_H_
+
+#include <string>
```
AlanGriffiths: Not needed

src/common/errno_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <mir/errno_utils.h>
+
+#include <cstring>
+
+namespace  {
```
AlanGriffiths:
```suggestion
namespace 
{
```

src/common/errno_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+{
+const char* errno_to_cstr(int err) noexcept
+{
+    thread_local char buf[256] = {};
+
+    auto ret = ::strerror_r(err, buf, sizeof(buf));
+    return strerror_r_to_cstr(ret, buf);
+}
+
+} // namespace mir
```
AlanGriffiths: We use a different style that cannot declare a new name:

```suggestion
const char* mir::errno_to_cstr(int err) noexcept
{
    thread_local char buf[256] = {};

    auto ret = ::strerror_r(err, buf, sizeof(buf));
    return strerror_r_to_cstr(ret, buf);
}
```

AlanGriffiths (approved):
> A couple of style tweaks

Pushed these to avoid another review cycle

src/common/errno_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include <mir/errno_utils.h>
+
+#include <cstring>
+
+namespace  {
+// Since we are normalizing between GNU/POSIX impl, one of function will be unused at compile time
+
+[[maybe_unused]]
+inline const char* strerror_r_to_cstr(int /*ret*/, const char* buf) noexcept
```
jhenstridge: If `ret` is not zero, then `buf` could contain garbage. It should probably return "Unknown error" in this case, similar to what you've done in the GNU version.

src/common/errno_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+namespace mir
+{
+const char* errno_to_cstr(int err) noexcept
+{
+    thread_local char buf[256] = {};
+
+    auto ret = ::strerror_r(err, buf, sizeof(buf));
+    return strerror_r_to_cstr(ret, buf);
```
jhenstridge: Is it really worth adding support for two versions of `strerror_r`? It basically means we'll have a code path that is unused for most of our testing.

Given that this code is in it's own source file now, we know what environment it is being compiled in. It would probably be simpler to put `#undef _GNU_SOURCE` before the includes, and use an assert to ensure we have the POSIX version. Something like:

```
static_assert(std::is_same_v<decltype(ret), int>);
```

AlanGriffiths:
> The use of thread_local means the thread safety guarantees are similar to those for plain `strerror` in modern glibc

We don't know we're on glibc. For example, Alpine uses musl

src/common/errno_utils.cpp
aliihsancengiz: That would be a better aproach indeed. 
But I am not sure undefining `_GNU_SOURCE` will force the compiler to use a POSIX/XSI-compliant version. 

At least, when I apply it, my compiler is unable to find the POSIX version.

```
gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
g++ (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
ldd (Ubuntu GLIBC 2.39-0ubuntu8.6) 2.39

```
Maybe i am missing something?

jhenstridge: The man page has this to say:

```
       strerror_r():
           The XSI-compliant version is provided if:
               (_POSIX_C_SOURCE >= 200112L) && ! _GNU_SOURCE
           Otherwise, the GNU-specific version is provided.
```

and:

> If no feature test macros are explicitly defined, then (since glibc 2.4) _POSIX_C_SOURCE is defined by default with the value 200112L,  so  that  the  XSI-compliant version of strerror_r() is provided by default.

That doesn't fully line up with my observed behaviour under C++ mode at least: it seems `_GNU_SOURCE` gets forced on, and `_POSIX_C_SOURCE` doesn't get set. I did manage to get the POSIX version selected with the following test program though:

```c++
#undef _GNU_SOURCE
#define _POSIX_C_SOURCE 200112L
#include <cstring>
#include <cerrno>
#include <type_traits>

int main()
{
        char buf[256];
        auto ret = strerror_r(EPERM, buf, sizeof buf);
        static_assert(std::is_same_v<decltype(ret), int>);
}
```

AlanGriffiths:
@aliihsancengiz you appear to be in the habit of rebasing your PRs. This is a practice that makes it harder on reviewers:  it is hard to separate what has changed in the PR itself from unrelated changes. Could you avoid rebasing in future (unless there's a merge conflict that needs addressing)?

src/common/errno_utils.cpp
AlanGriffiths: > it seems `_GNU_SOURCE` gets forced on

I see Mir's top level `CMakeLists.txt` sets `_GNU_SOURCE` globally. That doesn't seem either necessary nor appropriate. I'll try narrowing down the scope of this separately

src/common/errno_utils.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include <mir/errno_utils.h>
+
+#include <cstring>
+
+namespace
+{
+// Since we are normalizing between GNU/POSIX impl, one of these functions will be unused at compile time
+
+[[maybe_unused]]
```
tarek-y-ismail: Would be nice if we could get rid of these, but I understand that they're required so far for the GNU/POSIX dance

aliihsancengiz:
> @aliihsancengiz you appear to be in the habit of rebasing your PRs. This is a practice that makes it harder on reviewers: it is hard to separate what has changed in the PR itself from unrelated changes. Could you avoid rebasing in future (unless there's a merge conflict that needs addressing)?

Sorry @AlanGriffiths , was just for keeping the branch up to date, now i have added new version with new commit, should be easy to review.

src/common/errno_utils.cpp
AlanGriffiths: > I'll try narrowing down the scope of this separately

And then this was pointed out to me: https://github.com/canonical/mir/commit/1d128ba911ae2e65e950dce587b13089c377af18 - apparently we define `_GNU_SOURCE` everywhere to avoid ODR violations

PR #4593, opened 2026-01-02
Fix/logging snprintf negative time 4556

### LINKED ISSUE #4556: Fix logging `snprintf` with negative time values
https://github.com/canonical/mir/blob/a3798aab082e6cc483538ba274cdf10d9c79384f/src/common/logging/input_timestamp.cpp#L30-L31

The reminder operator preserves the sign `age_ns` is negative (saw it a couple of times when debugging), so the output can looks something like `-1.-500000ms`

DigraJatin:
Closes #4556 

## What's new?

1. Updated input_timestamp() to correctly handle future timestamps.
2. Replaced the buggy modulo logic—which produced confusing "double negative" strings like -100.-500ms ago—with absolute duration calculations.
3. Added a contextual suffix (ago vs in future) to make the logs human-readable regardless of clock drift or event timing. (No negative sign would be printed for future timestamps).

Past values will be printed like - 
`500 ms ago`
Future values will be printed like - 
`100 ms in future`

## How to test
Added unit tests for the function, run `./bin/mir_unit_tests --gtest_filter=TimestampTest.*` and verify.

## Checklist

- [ ✅] Tests added and pass
- [ ✅] Adequate documentation added

### DIFF
--- a/tests/unit-tests/logging/test_input_timestamp.cpp
+++ b/tests/unit-tests/logging/test_input_timestamp.cpp
@@ -39,7 +39,6 @@ TEST(TimestampTest, past_time_is_correctly_formatted)
 
 TEST(TimestampTest, future_time_is_correctly_formatted)
 {
-    using namespace std::chrono_literals;
     auto const future_event = steady_clock::now().time_since_epoch() + 1000ms + 10us;
 
     std::string out = ml::input_timestamp(future_event);


tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,49 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <mir/logging/input_timestamp.h>
+
+#include <gtest/gtest.h>
+#include <chrono>
+#include <string>
+#include <cstdio>
```
tarek-y-ismail: Why are we including this?

tarek-y-ismail (changes requested):
Looks good. One small nit...

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+using namespace std::chrono;
+namespace ml = mir::logging;
+
+class TimestampTest : public ::testing::Test {
+};
```
AlanGriffiths: You don't need a test fixture - so don't declare one: Just use the `TEST` macro instead of the `TEST_F` macro.

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+TEST_F(TimestampTest, HandlesPastTime) {
```
AlanGriffiths: 1. We usually put braces on the next line
2. You don't need a test fixture
3. we use snake_case for test names
 
```suggestion
TEST(TimestampTest, past_time_is_correctly_formatted) 
{
```

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,49 @@
...
+    auto now = steady_clock::now().time_since_epoch();
+    nanoseconds past_event = duration_cast<nanoseconds>(now - milliseconds(500));
+
+    std::string out = ml::input_timestamp(past_event);
+
+    EXPECT_EQ(out.find(".-"), std::string::npos);
+    EXPECT_NE(out.find("500.000000ms ago"), std::string::npos);
```
AlanGriffiths: Non-blocking, but we prefer the "fluent" style of test expectations:

```suggestion
    EXPECT_THAT(out, Not(HasSubstr(".-"));
    EXPECT_THAT(out, HasSubstr("500.000000ms ago"));
```

Note this needs a bit more context from gmock before it compiles:

```c++
#include <gmock/gmock.h>
...
using namespace testing;
```

src/common/logging/input_timestamp.cpp
```diff
@@ -26,9 +26,15 @@ std::string mir::logging::input_timestamp(nanoseconds when)
...
+    long long const milliseconds = abs_age_ns / 1000000LL;
+    long long const sub_milliseconds = abs_age_ns % 1000000LL;
+    
+    char const* sign_suffix = (age_ns < 0) ? "in future" : "ago";
+
     char str[64];
-    snprintf(str, sizeof str, "%lld (%lld.%06lldms ago)",
-             when_ns, age_ns / 1000000LL, age_ns % 1000000LL);
+    snprintf(str, sizeof str, "%lld (%lld.%06lldms %s)",
+             when_ns, milliseconds, sub_milliseconds, sign_suffix);
```
AlanGriffiths: Non blocking (pre-existing) the `str` buffer can be replaced by `std::format`

src/common/logging/input_timestamp.cpp
```diff
@@ -26,9 +26,15 @@ std::string mir::logging::input_timestamp(nanoseconds when)
     long long const now_ns = duration_cast<nanoseconds>(now).count();
     long long const age_ns = now_ns - when_ns;
 
+    long long const abs_age_ns = std::llabs(age_ns);
```
AlanGriffiths, line 30: We should be including `cstdlib` for `llabs()`

AlanGriffiths (changes requested):
Some tweaks and some non-blocking comments you can chose to ignore

src/common/logging/input_timestamp.cpp
DigraJatin: Leaving this for now, i think better would be to have a enhacement/refactor ticket to remove such buffers throughout codebase.

tests/unit-tests/logging/test_input_timestamp.cpp
DigraJatin: removed.

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+#include <mir/logging/input_timestamp.h>
+
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <chrono>
+#include <string>
+
+using namespace std::chrono;
```
tarek-y-ismail, line 24: You can additionally use `std::chrono_literals` which will enable you to define `past_event` as 
```
    auto const past_event = steady_clock::now().time_since_epoch() - 500ms;
```
The same applies to `future_event`.

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,50 @@
...
+using namespace testing;
+namespace ml = mir::logging;
+
+TEST(TimestampTest, past_time_is_correctly_formatted)
+{
+    auto now = steady_clock::now().time_since_epoch();
+    nanoseconds past_event = duration_cast<nanoseconds>(now - milliseconds(500));
```
tarek-y-ismail: These two lines can be merged as mentioned above. The final variable can also be const.

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    std::string out = ml::input_timestamp(past_event);
+
+    EXPECT_THAT(out, Not(HasSubstr(".-")));
+    EXPECT_THAT(out, HasSubstr("500.000000ms ago"));
```
tarek-y-ismail: Seems like CI has some float formatting inconsistencies. I think you should loosen the test a bit. Just testing for "500.0" and "ago" should be enough.

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,50 @@
...
+}
+
+TEST(TimestampTest, future_time_is_correctly_formatted)
+{
+    auto now = steady_clock::now().time_since_epoch();
+    nanoseconds future_event = duration_cast<nanoseconds>(now + nanoseconds(1000000123LL));
```
tarek-y-ismail: Can be one line & const.

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+    std::string out = ml::input_timestamp(future_event);
+
+    EXPECT_THAT(out, Not(HasSubstr(".-")));
+    EXPECT_THAT(out, HasSubstr("in future"));
+    EXPECT_THAT(out, HasSubstr("1000.000123ms"));
```
tarek-y-ismail: Again, CI has float formatting inconsistencies. Might have to bump the timestamp by a millisecond and check for "1000.0" and "in the future"

src/common/logging/input_timestamp.cpp
```diff
@@ -26,9 +27,15 @@ std::string mir::logging::input_timestamp(nanoseconds when)
...
+    long long const milliseconds = abs_age_ns / 1000000LL;
+    long long const sub_milliseconds = abs_age_ns % 1000000LL;
+
+    char const* sign_suffix = (age_ns < 0) ? "in future" : "ago";
```
tarek-y-ismail:
```suggestion
    char const* sign_suffix = (age_ns < 0) ? "in the future" : "ago";
```

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <chrono>
+#include <string>
+
+using namespace std::chrono;
+using namespace testing;
+namespace ml = mir::logging;
+
+TEST(TimestampTest, past_time_is_correctly_formatted)
+{
+    using namespace std::chrono_literals;
```
tarek-y-ismail: Small nit: Since this isn't a header file, you can move this to the top of the file safely. Though it's not a big issue unless we add a lot more tests.

tarek-y-ismail (approved):
Looking good! Now let's see if CI is happy.

tests/unit-tests/logging/test_input_timestamp.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    EXPECT_THAT(out, Not(HasSubstr(".-")));
+    EXPECT_THAT(out, HasSubstr("500.0"));
+    EXPECT_THAT(out, HasSubstr("ms ago"));
+}
+
+TEST(TimestampTest, future_time_is_correctly_formatted)
+{
+    using namespace std::chrono_literals;
+    auto const future_event = steady_clock::now().time_since_epoch() + 1000ms + 123ns;
```
tarek-y-ismail: Looks like CI still isn't happy. I tried

```suggestion
    auto const future_event = steady_clock::now().time_since_epoch() + 1000ms + 10us;
```

and it seems to work consistently locally. Can you try that?

tarek-y-ismail (changes requested):

DigraJatin:
Not sure what's failing now 😅

Saviq:
> Not sure what's failing now 😅

Not your fault @DigraJatin:
- #4610

Thanks for this!

PR #4612, opened 2026-01-08
debian: refresh and sort build dependencies

Saviq:
## What's new?

Cleared some unneeded dependencies. Best reviewed commit at a time, since sorting is on top.

## How to test

CI

## Checklist

- [x] Tests added and pass


debian/control
```diff
@@ -27,6 +27,7 @@ Build-Depends: cmake,
                libudev-dev,
                libgtest-dev,
                libgmock-dev,
+               libgmp-dev,
```
AlanGriffiths: AFAICS this shouldn't be needed: #4613

PR #4617, opened 2026-01-09
Reduce scope of _GNU_SOURCE

AlanGriffiths:
There are a few bits of code that depend on features enabled by `_GNU_SOURCE`, but there is no reason to enable it unless it is needed.


Saviq:
I think @RAOF may have thoughts: https://github.com/canonical/mir/commit/1d128ba911ae2e65e950dce587b13089c377af18.

AlanGriffiths:
> I think @RAOF may have thoughts: [1d128ba](https://github.com/canonical/mir/commit/1d128ba911ae2e65e950dce587b13089c377af18).

That's bad

PR #4632, opened 2026-01-14
Partial implementation of ext_image_copy_capture_v1 cursor sessions

jhenstridge:
## What's new?

This implements cursor the cursor session parts of the ext_image_copy_capture_v1 protocol. At present, it sends a dummy cursor image, which will be replaced with the real cursor in a follow-up PR.

Cursor position tracking is done via the CursorObserverMultiplexer, with a source-dependent backend class responsible for transforming the output space cursor coordinates to those of the corresponding image capture, and indicating whether the cursor is within bounds.

This can be tested like so:

1. Run miral with the following command:
    ```
    ./bin/miral-app --add-wayland-extensions ext_output_image_capture_source_manager_v1:ext_image_copy_capture_manager_v1:zwlr_virtual_pointer_manager_v1:zwp_virtual_keyboard_manager_v1
    ```
2. Run wayvnc with `WAYLAND_DISPLAY=wayland-1 wayvnc`
3. Connect using a VNC client that supports client side cursors. Remina works here, but GNOME Connections does not.
4. See that the placeholder white triangle cursor is displayed.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/ext_image_capture_v1.cpp
+++ b/src/server/frontend_wayland/ext_image_capture_v1.cpp
@@ -535,7 +535,8 @@ void mf::ExtOutputImageCaptureSourceManagerV1::create_source(wl_resource* new_re
     ExtImageCopyBackendFactory backend_factory = [output=wayland::make_weak(&output_global), ctx=ctx](auto *session, bool overlay_cursor) {
         return std::make_unique<ExtOutputImageCopyBackend>(session, overlay_cursor, wayland::as_nullable_ptr(output), ctx);
     };
-    ExtImageCopyCursorMapPosition cursor_map_position = [output=wayland::make_weak(&output_global)](float abs_x, float abs_y) -> std::optional<geom::Point> {
+    ExtImageCopyCursorMapPosition cursor_map_position = [output=wayland::make_weak(&output_global)](float abs_x, float abs_y) -> std::optional<geom::Point>
+{
         if (!output)
         {
             return std::nullopt;


src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -478,14 +573,52 @@ void mf::ExtOutputImageCaptureSourceManagerV1::create_source(wl_resource* new_re
...
+    // Translate pointer coordinates to buffer space. We don't take
+    // rotations/flips into account because our image copy backend is
+    // producing untransformed frames.
+    auto const buffer_size = config.modes[config.current_mode_index].size;
+    auto const x_scale = static_cast<double>(buffer_size.width.as_value()) / output_space_area.size.width.as_value();
+    auto const y_scale = static_cast<double>(buffer_size.height.as_value()) / output_space_area.size.height.as_value();
+    return geom::Point{
+        (abs_x - output_space_area.left().as_value()) * x_scale,
+        (abs_y - output_space_area.top().as_value()) * y_scale
+    };
```
AlanGriffiths: We can avoid unwrapping the values and gain type safety and clarity:

```suggestion
    auto const raw_offset = abs_pos - output_space_area.top_left;
    decltype(raw_offset) const scaled_offset{x_scale * raw_offset.dx, y_scale * raw_offset.dy};

    return as_point(scaled_offset);
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -734,3 +884,192 @@ void mf::ExtImageCopyCaptureFrameV1::report_result(
...
+    wayland::Weak<ExtImageCopyCaptureCursorSessionV1> const session;
+};
+
+class mf::ExtImageCopyCaptureCursorSessionV1::ImageCopyBackend
+    : public ExtImageCopyBackend
+{
+public:
+    ImageCopyBackend(ExtImageCopyCaptureSessionV1 *session,
+                     std::shared_ptr<time::Clock> const& clock);
+    //~ImageCopyBackend();
```
AlanGriffiths:
```suggestion
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -734,3 +884,192 @@ void mf::ExtImageCopyCaptureFrameV1::report_result(
...
+void mf::ExtImageCopyCaptureCursorSessionV1::ImageCopyBackend::begin_capture(
+        std::shared_ptr<renderer::software::RWMappable> const& shm_data,
+        [[maybe_unused]] geom::Rectangle const& frame_damage,
+        CaptureCallback const& callback)
+{
+    auto mapping = shm_data->map_writeable();
+    auto data = mapping->data();
+    auto size = mapping->size();
+    auto stride = mapping->stride();
+    printf("Buffer length is %d (%dx%d, stride %d)\n", (int)mapping->len(), size.width.as_int(), size.height.as_int(), stride.as_int());
```
AlanGriffiths: Debugging?

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -734,3 +884,192 @@ void mf::ExtImageCopyCaptureFrameV1::report_result(
...
+            }
+            else
+            {
+                data[pos] = std::byte{0x00};
+                data[pos+1] = std::byte{0x00};
+                data[pos+2] = std::byte{0x00};
+                data[pos+3] = std::byte{0x00};
+            }
+        }
+    }
```
AlanGriffiths, line 924: I guess this temporary faking for image capture? I won't whine about the C-style loops

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -108,6 +113,26 @@ class ExtImageCopyBackend
...
 
+class ExtImageCopyCaptureCursorSessionV1;
+
+class ExtImageCopyCursorBackend
+{
+public:
+    using CaptureResult = std::expected<std::tuple<time::Timestamp, geom::Rectangle>, uint32_t>;
+    using CaptureCallback = std::function<void(CaptureResult const&)>;
+
+    ExtImageCopyCursorBackend(ExtImageCopyCaptureCursorSessionV1* session);
```
mattkae: Nit:

```suggestion
    explicit ExtImageCopyCursorBackend(ExtImageCopyCaptureCursorSessionV1* session);
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -108,6 +113,26 @@ class ExtImageCopyBackend
...
+    virtual ~ExtImageCopyCursorBackend() = default;
+
+    ExtImageCopyCursorBackend(ExtImageCopyCursorBackend const&) = delete;
+    ExtImageCopyCursorBackend& operator=(ExtImageCopyCursorBackend const&) = delete;
+
+    virtual auto map_position(float abs_x, float abs_y) -> std::optional<geom::Point> = 0;
```
mattkae: Can this just be a static method that takes the `ExtImageCopyCaptureCursorSessionV1 const* session` as a parameter? The class might be overkill for our use case

mattkae: Oh - or is this virtual so that we can override the behavior in test?

mattkae: And - in that case - should it even take the `ExtImageCopyCaptureCursorSessionV1* session` since that's an implementation detail?

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -734,3 +884,192 @@ void mf::ExtImageCopyCaptureFrameV1::report_result(
...
+        auto const& point = position.value();
+        send_position_event(point.x.as_int(), point.y.as_int());
+    }
+    else
+    {
+        if (was_visible)
+        {
+            send_leave_event();
+        }
+    }
```
mattkae:
```suggestion
    else if (was_visible)
    {
        send_leave_event();
    }
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: My thinking was that we'd need a way to tell the session that the source has gone away. It looks like there isn't a `shutdown` event directly on the cursor session object, but there is one on the related image copy session for the cursor image itself.

It's likely clients would destroy the cursor session objects when the related image copy session for the source itself shuts down, but it seems reasonable for these ones to be able to clean up too. That's not yet implemented though.

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -108,6 +113,26 @@ class ExtImageCopyBackend
     void apply_damage(std::optional<geom::Rectangle> const& damage);
 };
 
+class ExtImageCopyCaptureCursorSessionV1;
+
+class ExtImageCopyCursorBackend
```
AlanGriffiths: I don't understand the role of this class:

- It isn't an _interface_ as it has a data member and a constructor parameter that initializes it;
- The only implementation of `map_position()` makes no reference to `session`; and,
- The members `CaptureResult` and `CaptureCallback` are not used in the interface

It looks a simpler design could be making `map_position()` part of an _interface_ implemented by `ExtImageCopyCaptureCursorSessionV1`?

If it is intended to "manage" the lifetime of `session`, then perhaps that should be a `Weak` handle?

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -241,6 +286,39 @@ class ExtImageCopyCaptureFrameV1
...
+  ~ExtImageCopyCaptureCursorSessionV1();
+
+private:
+    class CursorObserver;
+    class ImageCopyBackend;
+
+    void get_capture_session(wl_resource* session) override;
+
+    // From CursorObserver
+    void cursor_moved_to(float abs_x, float abs_y);
```
AlanGriffiths: What does this comment mean? `ExtImageCopyCaptureCursorSessionV1` is not a `CursorObserver`

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: After discussing with Chris, we decided that clients will likely rely on the shutdown event on the main image copy session, so I've reworked the code to use a std::function instead.

src/server/frontend_wayland/ext_image_capture_v1.cpp
AlanGriffiths: @jhenstridge why do you keep marking this comment "resolved"? It still seems valid to me

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -734,3 +836,174 @@ void mf::ExtImageCopyCaptureFrameV1::report_result(
         timespec.tv_nsec);
     send_ready_event();
 }
+
+class mf::ExtImageCopyCaptureCursorSessionV1::CursorObserver
+    : public mi::CursorObserver
+{
+public:
+    CursorObserver(ExtImageCopyCaptureCursorSessionV1& session)
```
mattkae: Nit:

```suggestion
    explicit CursorObserver(ExtImageCopyCaptureCursorSessionV1& session)
```

src/server/frontend_wayland/ext_image_capture_v1.cpp
AlanGriffiths: Still a confusing comment

src/server/frontend_wayland/ext_image_capture_v1.cpp
AlanGriffiths: Yes, better to be consistent.

AlanGriffiths (changes requested):
> Cool by me! I'll let Alan's feedback take it over the line

Just the same feedback I left two weeks ago

src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -478,14 +535,41 @@ void mf::ExtOutputImageCaptureSourceManagerV1::create_source(wl_resource* new_re
     ExtImageCopyBackendFactory backend_factory = [output=wayland::make_weak(&output_global), ctx=ctx](auto *session, bool overlay_cursor) {
         return std::make_unique<ExtOutputImageCopyBackend>(session, overlay_cursor, wayland::as_nullable_ptr(output), ctx);
     };
-    new ExtImageCaptureSourceV1{new_resource, backend_factory};
+    ExtImageCopyCursorMapPosition cursor_map_position = [output=wayland::make_weak(&output_global)](float abs_x, float abs_y) -> std::optional<geom::Point> {
```
AlanGriffiths:
```suggestion
    ExtImageCopyCursorMapPosition cursor_map_position = [output=wayland::make_weak(&output_global)](float abs_x, float abs_y) -> std::optional<geom::Point>
{
```

AlanGriffiths (approved):
Just some braces to twiddle

PR #4640, opened 2026-01-16
Work around warping on nvidia with Atomic KMS + GBM KMS

### LINKED ISSUE #4035: Rendered applications are warped when rendered on one GPU and displayed on another (intel + nvidia)
Applying the workaround in #4125 to get the compositor to launch on atomic-kms + gbm-kms, some applications appear to warp when rendered. The warping is fixed it the texture width is a multiple of 8, and increases the further you get from it. 

Image

This only occurs if the application is rendered on the intel GPU and is displayed on the nvidia GPU. In this case, the texture data is passed on via DMABUF through a common format texture (RGBA8888), and then uploaded to OpenGL on nvidia display GPU.

Dumping the textures at various points (intel GPU EGLImage, common EGLImage, nvidia EGLImage pre-upload, and nvidia texture post-upload) shows no warping whatsoever. The texture is read via `glReadPixels`.

One known workaround is to force the width of the common and imported textures to be a multiple of 8.

If the app is rendered on the nvidia GPU, it displays correctly on both monitors.

tarek-y-ismail:
Closes #4035 

## What's new?
- Adds a fallback path in `gbm_bo_with_modifiers_or_linear`  to account for nvidia non-compliance 
- Adds a codepath in `DMABufEGLProvider` to copy buffers via the CPU
- Adds a quirk to detect to determine which codepath to use for buffer transfers, and passes that down to `DMABufEGLProvider`
- Any transfer involving nvidia goes through the CPU for now until we get to the bottom of the bug.

## How to test
On a machine with an intel iGPU and an nvidia dGPU, with one output connected to each:
1. Launch mir as follows:
```
miral-app --platform-display-libs=mir:atomic-kms --platform-rendering-libs=mir:gbm-kms --display-config=sidebyside --driver-quirks=gbm-buffer-transfer-strategy:nvidia:nvidia:dma
```
2. Launch one of `firefox`, `ghostty`, `gtk4-demo` as follows 
```
__EGL_VENDOR_LIBRARY_FILENAMES=/usr/share/glvnd/egl_vendor.d/50_mesa.json  DRI_PRIME=1 <command>
```
3. Drag the window onto the screen connected to the nvidia card
4. If the app still renders correctly, resize the *width* using the mouse 
5. Observe as the window warps in a sawtooth pattern as you change the width
6. Repeat steps 1-4, but remove the last flag or change it  to `--driver-quirks=gbm-buffer-transfer-strategy:nvidia:nvidia:cpu`
7. Observe as the window renders correctly.

## Checklist
- [ ] Adequate documentation added (How should we document this?)

### DIFF
--- a/src/platform/graphics/linux_dmabuf.cpp
+++ b/src/platform/graphics/linux_dmabuf.cpp
@@ -1852,7 +1852,7 @@ auto mg::DMABufEGLProvider::dma_transfer(std::shared_ptr<DmabufTexBuffer> const&
      * This is a logic bug, so go noisily.
      */
     BOOST_THROW_EXCEPTION(
-        (std::logic_error{"Failed to find import parameterns for buffer we explicitly allocated for import"}));
+        (std::logic_error{"Failed to find import parameters for buffer we explicitly allocated for import"}));
 }
 
 auto mg::DMABufEGLProvider::as_texture(std::shared_ptr<Buffer> buffer) -> std::shared_ptr<gl::Texture>
--- a/src/platform/graphics/linux_dmabuf.cpp
+++ b/src/platform/graphics/linux_dmabuf.cpp
@@ -1538,7 +1538,7 @@ void mg::DMABufEGLProvider::validate_import(DMABufBuffer const& dma_buf)
     }
 }
 
-auto mg::DMABufEGLProvider::as_texture(std::shared_ptr<NativeBufferBase> buffer)
+auto mg::DMABufEGLProvider::as_texture(std::shared_ptr<Buffer> buffer)
     -> std::shared_ptr<gl::Texture>
 {
     if (auto dmabuf_tex = std::dynamic_pointer_cast<DmabufTexBuffer>(buffer))
--- a/src/platform/graphics/linux_dmabuf.cpp
+++ b/src/platform/graphics/linux_dmabuf.cpp
@@ -980,8 +980,9 @@ auto format_has_known_alpha(mg::DRMFormat format) -> std::optional<bool>
     return format.info().transform([](auto const& info) { return info.has_alpha(); });
 }
 }
+}
 
-class DmabufTexBuffer :
+class mg::DMABufEGLProvider::DmabufTexBuffer :
     public mg::BufferBasic,
     public mg::DMABufBuffer
 {
@@ -1277,8 +1278,6 @@ private:
     mg::DRMFormat const format_;
 };
 
-}
-
 class mg::LinuxDmaBuf::Instance : public mir::wayland::LinuxDmabufV1
 {
 public:
@@ -1401,15 +1400,17 @@ mg::DMABufEGLProvider::DMABufEGLProvider(
     std::shared_ptr<EGLExtensions> egl_extensions,
     mg::EGLExtensions::EXTImageDmaBufImportModifiers const& dmabuf_ext,
     std::shared_ptr<mgc::EGLContextExecutor> egl_delegate,
-    EGLImageAllocator allocate_importable_image)
-    : dpy{dpy},
-      egl_extensions{std::move(egl_extensions)},
-      dmabuf_export_ext{mg::EGLExtensions::MESADmaBufExport::extension_if_supported(dpy)},
-      devnum_{get_devnum(dpy)},
-      formats{std::make_unique<DmaBufFormatDescriptors>(dpy, dmabuf_ext)},
-      egl_delegate{std::move(egl_delegate)},
-      allocate_importable_image{std::move(allocate_importable_image)},
-      blitter{std::make_unique<mg::EGLBufferCopier>(this->egl_delegate)}
+    EGLImageAllocator allocate_importable_image,
+    BufferTransferStrategy buffer_transfer_strategy):
+    dpy{dpy},
+    egl_extensions{std::move(egl_extensions)},
+    dmabuf_export_ext{mg::EGLExtensions::MESADmaBufExport::extension_if_supported(dpy)},
+    devnum_{get_devnum(dpy)},
+    formats{std::make_unique<DmaBufFormatDescriptors>(dpy, dmabuf_ext)},
+    egl_delegate{std::move(egl_delegate)},
+    allocate_importable_image{std::move(allocate_importable_image)},
+    blitter{std::make_unique<mg::EGLBufferCopier>(this->egl_delegate)},
+    buffer_transfer_strategy{std::move(buffer_transfer_strategy)}
 {
 }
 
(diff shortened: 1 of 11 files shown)


Saviq:
> How should we document this?

These seem appropriate?
- https://canonical-mir.readthedocs-hosted.com/latest/explanation/platform-support/
- https://canonical-mir.readthedocs-hosted.com/latest/explanation/driver-quirks/

src/platform/graphics/linux_dmabuf.cpp
```diff
@@ -1538,125 +1463,432 @@ void mg::DMABufEGLProvider::validate_import(DMABufBuffer const& dma_buf)
...
+        importable_buf->planes(),
+        importing_provider->dpy,
+        *importing_provider->egl_extensions);
+    auto sync = importing_provider->blitter->blit(src_image, importable_image, dmabuf_tex->size());
+    if (sync)
+    {
+        BOOST_THROW_EXCEPTION((std::logic_error{"EGL_ANDROID_native_fence_sync support not implemented yet"}));
+    }
+    auto importable_dmabuf = export_egl_image(
+        *importing_provider->dmabuf_export_ext, importing_provider->dpy, importable_image, dmabuf_tex->size());
```
RAOF, line 1911: We don't need this; `importable_buf` is *already* a dma-buf with an importable format/modifier setup.

(This means we can also delete the `export_egl_image` function entirely)

src/platforms/gbm-kms/server/kms/platform.cpp
```diff
@@ -275,48 +283,73 @@ class DMABufBuffer : public mg::DMABufBuffer
...
+        switch(errno)
         {
-            BOOST_THROW_EXCEPTION((
-                std::system_error{
-                    errno,
-                    std::system_category(),
-                    "Failed to allocate GBM bo without modifiers"}));
+        case ENOSYS:
+            // We get ENOSYS if the GBM implementation can't handle modifiers
+            if (std::ranges::contains(modifiers, DRM_FORMAT_MOD_LINEAR))
```
RAOF: This conditional appears reversed?

RAOF (commented):
The quirk is pragmatic, but it would ideally take into account *both* GPUs - as far as I can tell, this bug only manifests when importing from Intel to NVIDIA (it imports fine from AMD to NVIDIA on my system).

src/platforms/gbm-kms/server/kms/quirks.cpp
```diff
@@ -219,7 +230,26 @@ class mgg::Quirks::Impl
...
+        auto gbm_buffer_transfer_strategy_name = mgc::apply_quirk(
+            devnode,
+            driver,
+            gbm_buffer_transfer_strategy.devnodes,
+            gbm_buffer_transfer_strategy.drivers,
+            "gbm-buffer-transfer-strategy");
+
+        auto gbm_buffer_transfer_strategy = [&]() -> GbmQuirks::BufferTransferStrategy
+        {
+            if (gbm_buffer_transfer_strategy_options.contains(gbm_buffer_transfer_strategy_name))
```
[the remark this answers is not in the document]
tarek-y-ismail: Intended behavior.

src/platform/graphics/linux_dmabuf.cpp
```diff
@@ -1538,125 +1463,432 @@ void mg::DMABufEGLProvider::validate_import(DMABufBuffer const& dma_buf)
...
+}
+}
+
+auto mg::DMABufEGLProvider::cpu_transfer(std::shared_ptr<DmabufTexBuffer> const& dmabuf_tex) -> std::shared_ptr<mg::gl::Texture>
+{
+    auto const pixel_format = mir_pixel_format_argb_8888;
+    auto const size = dmabuf_tex->size();
+    return std::make_shared<ShmBufferTexture>(
+        egl_delegate,
+        mg::BufferID{0xfade}, // TODO: BufferID registry to keep track of and generate new buffer IDs?
```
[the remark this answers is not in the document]
tarek-y-ismail: Should not matter for now

tarek-y-ismail:
The more I work on this problem, the weirder it gets:
- Apparently the warping occurs when we import from nvidia to nvidia for some reason.
  Forcing that to use the CPU path via `--driver-quirks=gbm-buffer-transfer-strategy:nvidia-drm:nvidia-drm:cpu` fixes the warping. That points to the following code path: https://github.com/canonical/mir/blob/1a04fce5a3ce8775e6e154a3001c9b5faa2c6cc2/src/platform/graphics/linux_dmabuf.cpp#L1796-L1800
- Logs show the source and importing sides are flipped. Isn't the buffer provider the source?
  Rendering a firefox window on the iGPU, and dragging it to the iGPU's output causes transfers from the dGPU to the iGPU??
- Rendering a firefox window on the dGPU and dragging it from the dGPU output to the iGPU output causes the iGPU output to stop displaying frames altogether, even if firefox is shutdown and relaunched on the iGPU
  The splash screen seems to trigger this as well. This might be related to the next point too.
- Still have freezes during shutdown. Dunno from where exactly but last time I checked it was the iGPU rendering thread getting stuck in `eglSwap`. Won't snoop after it right now until the main bug is fixed

EDIT: Attaching a log. Opened qterminal, then firefox rendering on the iGPU and displaying on the dGPU. Then dragged the window to the iGPU's output and back. Closed firefox. Relaunched it without specifying GPUs (which usually renders on the dGPU). Repeated the roundtrip to the iGPU output and back. Shut down.

[20260119_141900-mir-atomic-kms-hybrid-log.txt](https://github.com/user-attachments/files/24715336/20260119_141900-mir-atomic-kms-hybrid-log.txt)

src/platform/graphics/linux_dmabuf.cpp
tarek-y-ismail: Removed it but reversed the change temporarily other issues are taken care of

tarek-y-ismail:
Looks like DMA transfers from nvidia to intel in the case where firefox is rendering on nvidia and _should_ display on intel cause the intel compositing thread to completely freeze somewhere in the mesa stack in `dri_flush`. Again, going through the CPU alleviates this.

Edit: Forgot to add, checking dmesg logs shows no errors being thrown, mesa debugging symbols don't help too. I think I might have to build from source to investigate this further. But I think I've spent enough time on this issue as is.

tarek-y-ismail:
Well, this is awkward. I just tried running kgx **rendering on the nvidia GPU** and it still suffers from warping. Its colors are broken compared to KDE as well. Pluma and gedit work fine though.

src/platform/graphics/linux_dmabuf.cpp
```diff
@@ -1538,125 +1575,385 @@ void mg::DMABufEGLProvider::validate_import(DMABufBuffer const& dma_buf)
...
+        dmabuf_tex->map_readable()->data(),
+        size,
+        geom::Stride{size.width.as_int() * MIR_BYTES_PER_PIXEL(pixel_format)},
+        pixel_format);
```
RAOF:
```suggestion
    auto mapping = dmabuf_tex->map_readable();
    return std::make_shared<ShmBufferTexture>(
        egl_delegate,
        mg::BufferID{0xfade}, // TODO: BufferID registry to keep track of and generate new buffer IDs?
        mapping->data(),
        size,
        mapping->stride(),
        pixel_format);
```

tarek-y-ismail: Tried applying this fix, and also using the size and pixel format from the mapping. Could not reproduce the warping, though kgx still shows incorrect colors.

I also tried reproducing the bug by forcing the DMA path (hardcoded in code, options route to atomic for some reason). I could not reproduce the warping with that either. The [logs](https://github.com/user-attachments/files/26021752/20260316_130716-mir-atomic-kms-hybrid-log.txt) show nothing is using the CPU path. They only broken thing was that the compositing thread on the intel output still froze, which is an already known issue.

It's worth noting that I've added a second external monitor to my set up since last time. And that I'm using the 590.48.01 drivers. It could be that one of these changes is why warping isn't occurring. The fact that the minimal reproducer is still producing warped output on my machine is making me suspect that I might not be reproducing correctly.

Just wanted to try this before I forgot. Will try looking at it again when I have more time.

RAOF:
Oooh! As https://forums.developer.nvidia.com/t/egl-import-via-egl-ext-image-dma-buf-import-modifiers-ignores-explicit-stride-causes-image-distortion-in-virtio-gpu-venus/364360 points out, the quirk we need here is minimum-stride-alignment-for-linear-dmabufs; setting that to 32 gets us a quirk that'll fix the broken cases and leave the currently working cases (like imports from amdgpu, or imports from correctly-sized intel surfaces) fast.

tarek-y-ismail:
Well, at least we know it's not a Mir problem. I'll poke at it when I get some time.

tarek-y-ismail:
Can confirm that the latest commit fixes the tinting of kgx on the device that originally exhibited that bug (1660Ti), curiously, it does _not_ occur on the second device (5070Ti), despite both using the exact same driver (590.48.01) and being connected to the same monitor.

Note: Opus found a couple more minor bugs in the code. But I have to drop this for today.

PR #4644, opened 2026-01-19
Move logging APIs (and defaults) to mircore

AlanGriffiths:
The logging APIs are stable and useful to downstream projects. Make them more easily available

### DIFF
--- a/debian/libmircore2.symbols
+++ b/debian/libmircore2.symbols
@@ -45,13 +45,16 @@ libmircore.so.2 libmircore2 #MINVER#
  (c++)"mir::log(mir::logging::Severity, char const*, std::__exception_ptr::exception_ptr const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)@MIR_CORE_2.26" 2.26.0~dev
  (c++)"mir::log(mir::logging::Severity, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)@MIR_CORE_2.26" 2.26.0~dev
  (c++)"mir::log(mir::logging::Severity, char const*, char const*, ...)@MIR_CORE_2.26" 2.26.0~dev
- (c++)"mir::logv(mir::logging::Severity, char const*, char const*, __va_list_tag*)@MIR_CORE_2.26" 2.26.0~dev
+ (c++|optional)"mir::logv(mir::logging::Severity, char const*, char const*, __va_list_tag*)@MIR_CORE_2.26" 2.26.0~dev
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -2,7 +2,7 @@ mir (2.26.0~dev) UNRELEASED; urgency=medium
 
   * Begin development for 2.26.0 release
 
- -- Michał Sawicz <michal.sawicz@canonical.com>  Thu, 11 Dec 2025 10:56:11 +0100
+ -- Michał Sawicz <michal.sawicz@canonical.com>  Fri, 19 Dec 2025 10:56:11 +0100
```
mattkae: Spurious change?

src/core/CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ add_library(mircore SHARED
     depth_layer.cpp
     geometry/rectangles.cpp
     input/mousekeys_keymap.cpp
+    log.cpp
+    logging/dumb_console_logger.cpp
```
mattkae, line 14: Before we land this in the public API, can we change this name to something a little more informative? Perhaps `StdOutConsoleLogger` or something similar?

debian/changelog
AlanGriffiths: No, an attempt to avoid

> ```
> [41](https://github.com/canonical/mir/actions/runs/21179890981/job/60919006440?pr=4644#step:8:13542)
> DEBUG: RPM build errors:
> DEBUG:     %changelog not in descending chronological order
> ```

src/core/CMakeLists.txt
AlanGriffiths: Isn't `StdOut` both wrong (doesn't mention `stderr`) and tautologous?

AlanGriffiths:
> It seems likely that we'll be changing (or perhaps deprecating parts of) this API as a part of the logging work coming up.
> 
> Should we wait until we've decide what logging API additions/deprecations we are going to make before landing something in mircore we might almost immediately deprecate?

I have been wondering the same thing. But, from what I've seen of the logging API discussion, we can implement this API in terms of the proposed solution (which is an easy migration path). The main change that seems likely is `std::print()` style logging in place of the `printf()` style we have here. I don't see carrying both as a significant overhead.

src/core/CMakeLists.txt
AlanGriffiths: FWIW I'll also defend the `Dumb`: that is a comparison to a `Smart` logger that enabled things like filtering based on level and component. (Which something we're talking about doing now.)

mattkae: Maybe `Basic` or `Simple` is better, but I'm fine either way :+1:

PR #4646, opened 2026-01-20
Implementation of a minimal Rust Wayland server with C++ bindings and an example C++ host application

mattkae:
## What's new?
- Introduced the `wayland_rs` crate that contains a minimal Wayland server and corresponding C++ bindings for that server
- Packaged the crate using cxx in `mirwayland_rs`
- Use `mirwayland_rs` in the `mirwayland_rs-demo` example program which spins up the Wayland server and connects a client to it

## How to test
1. Run `mirwayland_rs-demo`
2. Note that the program exits gracefully when `wayland-98` is available

### DIFF
--- a/debian/libmircore2.symbols
+++ b/debian/libmircore2.symbols
@@ -48,6 +48,7 @@ libmircore.so.2 libmircore2 #MINVER#
  (c++|optional)"mir::logv(mir::logging::Severity, char const*, char const*, __va_list_tag*)@MIR_CORE_2.26" 2.26.0~dev
  (c++|optional)"mir::logv(mir::logging::Severity, char const*, char const*, std::__va_list)@MIR_CORE_2.26" 2.26.0~dev
  (c++|optional)"mir::logv(mir::logging::Severity, char const*, char const*, char*)@MIR_CORE_2.26" 2.26.0~dev
+ (c++|optional)"mir::logv(mir::logging::Severity, char const*, char const*, void*)@MIR_CORE_2.26" 2.26.0~dev
  (c++)"mir::logging::set_logger(std::shared_ptr<mir::logging::Logger> const&)@MIR_CORE_2.26" 2.26.0~dev
  (c++)"mir::logging::format_message(std::basic_ostream<char, std::char_traits<char> >&, mir::logging::Severity, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)@MIR_CORE_2.26" 2.26.0~dev
  (c++)"mir::logging::DumbConsoleLogger::log(mir::logging::Severity, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)@MIR_CORE_2.26" 2.26.0~dev


src/wayland/wayland_rs/src/wayland_server.rs
```diff
@@ -0,0 +1,148 @@
...
+                },
+            )
+            .expect("Failed to insert listener into event loop");
+
+        // Next, get the raw file descriptor from the display's backend and add it to the event loop.
+        // We need to get the fd without holding a borrow of the display.
+        let display_fd = unsafe {
+            let raw_fd = self.display.backend().poll_fd().as_raw_fd();
+            std::os::fd::BorrowedFd::borrow_raw(raw_fd)
+        };
```
RAOF: Why? I *think* this could be replaced withQ

```suggestion
        let display_fd = self.display.as_fd();
```
I don't think we need to reset the lifetime here?

src/wayland/wayland_rs/src/wayland_server.rs
```diff
@@ -0,0 +1,148 @@
...
+                    if let Ok(stream) = listener.accept() {
+                        if let Some(stream) = stream {
+                            // Insert the client into the display.
+                            // This registers the client's socket with the Display's internal backend.
+                            if let Err(e) = server
+                                .display
+                                .handle()
+                                .insert_client(stream, std::sync::Arc::new(ClientState))
+                            {
+                                eprintln!("Failed to add client: {}", e);
```
RAOF: We should probably be using `log::error!` (or possibly `tracing::error!`) here, right?

Not a blocker for this time, but we'll soon want to standardise how we log from Rust.

src/wayland/wayland_rs/src/wayland_server.rs
```diff
@@ -0,0 +1,148 @@
...
+    stop_requested: Arc<AtomicBool>,
+}
+
+impl WaylandServer {
+    /// Create a new wayland server.
+    pub fn new() -> Self {
+        WaylandServer {
+            display: Display::new().expect("Failed to create Wayland display"),
+            state: ServerState {},
+            stop_requested: Arc::new(AtomicBool::new(false)),
```
RAOF: Since we've got an event loop, maybe this should be a [nix::EventFd](https://docs.rs/nix/latest/nix/sys/eventfd/struct.EventFd.html), and then the loop could block indefinitely rather than running at 60Hz.

src/wayland/wayland_rs/src/wayland_server.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    /// Run the wayland server.
+    ///
+    /// This function will block the current thread.
+    ///
+    /// # Arguments
+    /// * `socket` - The name of the socket to bind to (e.g. "wayland-0").
+    pub fn run(&mut self, socket: &str) {
+        let mut event_loop: EventLoop<'_, WaylandServer> =
+            EventLoop::try_new().expect("Failed to create event loop");
```
RAOF: It's not *that* much extra effort to return `Result<(), SomeSortOfError>`; we might as well start on the right foot 😉

mattkae:
> Huh. How does `WaylandServer::stop()` work? You need a mutable reference to `self` there, but `WaylandServer::run()` already _has_ a mutable reference to `self`, which should make calling `stop()` impossible?

I think that is is one of the "We are calling this from C++ so who Rust doesn't know what's happening" cases

src/wayland/wayland_rs/src/wayland_server.rs
mattkae: Oh whoops! Missed that method

mattkae: Although that is considered a borrow there... Perhaps I'm missing something?

src/wayland/wayland_rs/src/wayland_server.rs
mattkae: Sure :)

src/wayland/wayland_rs/src/wayland_server.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    /// Run the wayland server.
+    ///
+    /// This function will block the current thread.
+    ///
+    /// # Arguments
+    /// * `socket` - The name of the socket to bind to (e.g. "wayland-0").
+    pub fn run(&mut self, socket: &str) -> Result<(), &'static str> {
```
RAOF:
```suggestion
    pub fn run(&mut self, socket: &str) -> Result<(), Box<dyn error::Error>> {
```
(plus `use std::error` above).

We actually might want to return a proper error enum here (in that calling code could usefully behave differently for “that socket name is already in use” as opposed to “permission denied” or “some internal error happened”), but that can safely be punted.

mattkae: Let's punt it until we have more reason to do so, but I agree in principle. The string is just easy for now :)

src/wayland/wayland_rs/src/wayland_server.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Stops the wayland server if it is running.
+    pub fn stop(&self) {
+        if let Some(signal) = self
+            .stop_signal
+            .lock()
+            .expect("No way to recover from lock poisioning")
+            .as_ref()
+        {
+            signal.ping();
+        }
```
RAOF, line 136: I think we could clean this up (remove the `Mutex` and `Optional`) pretty easily by putting the thread inside the Rust code, rather than in the C++ code.

mattkae:
We should land https://github.com/canonical/mir/pull/4655 before this

PR #4649, opened 2026-01-23
wayland-rs: Simplify by holding state locally

RAOF:
A couple of things here:
* Simplify lifetime stuff by moving a bunch of state into the only place it is used
* Return `Box::<dyn Error>` (although, as I mentioned, we may want to move to a proper error type
* Use `calloop::Ping` for the shutdown signal because we might as well

This also upgrades the version of calloop to 0.14(.3). Was there a particular reason for the weird version restriction you had?

(There's also a bunch of stuff here that could be cleaned up further; particularly the comments! :woman_facepalming: )

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,5 +18,5 @@ input = { version = ">=0.8.3, <0.10", features = ["libinput_1_19"] }
 libc = "0.2.175"
 nix = { version = "0.30.1", features = ["poll"] }
 wayland-server = "0.31"
-calloop = { version = "0.14" }
+calloop = { version = ">=0.10, <0.15" }
 log = "0.4"


mattkae:
I went with 0.10 of calloop because it is packages in the archive, but I think that we're good so long as we support a range

Cargo.toml
```diff
@@ -18,5 +18,5 @@ input = { version = ">=0.8.3, <0.10", features = ["libinput_1_19"] }
 libc = "0.2.175"
 nix = { version = "0.30.1", features = ["poll"] }
 wayland-server = "0.31"
-calloop = { version = ">=0.10, <0.15" }
+calloop = { version = "0.14" }
```
mattkae: This had a range so that we can support the version from the archive on questing, but maybe it doesn't necessarily matter for now. Do we need 0.14 minimum for the new features that we're using?

src/wayland/wayland_rs/src/wayland_server.rs
```diff
@@ -14,43 +14,27 @@
...
-    /// Stop flag set by the event loop callback.
-    stop_requested: bool,
+    stop_signal: Mutex<Option<Ping>>,
 }
 
 impl WaylandServer {
     /// Create a new wayland server.
     pub fn new() -> Self {
         WaylandServer {
-            display: Display::new().expect("Failed to create Wayland display"),
```
mattkae, line 47: This is fine for now, but I will have to have a way to register a global from C++ in the future using `DisplayHandle.create_global`.

How do you propose accessing that? I suppose that I could queue the work onto the event loop?

PR #4650, opened 2026-01-23
Parsing the Wayland protocols in our project into a Rust-consumable struct

mattkae:
## What's new?
Using serde + quick to load the Wayland protocol XMLs into a Rust struct

## How to test
Run `cargo build` and see that it works

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -103,7 +103,7 @@ jobs:
 
         sudo --preserve-env apt-get build-dep --yes ./
 
-        echo PATH=${PATH}:/usr/lib/rust-1.82/bin >> $GITHUB_ENV
+        echo PATH=/usr/lib/rust-1.82/bin:${PATH} >> $GITHUB_ENV
 
     - name: Configure
       run: >
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ resolver = "2"
 name = "mir"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.75.0"
+rust-version = "1.82.0"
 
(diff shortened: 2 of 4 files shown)


jhenstridge:
Is there anything in particular you'd want to achive at build time here, rather than using [wayland-scanner's proc macros](https://docs.rs/wayland-scanner/0.31.8/wayland_scanner/) to generate the code?

And outside of the Mir-specific protocols, is there any benefit to generating code from the in-tree XML files rather than using the versions provided by wayland-rs?

mattkae:
> Is there anything in particular you'd want to achive at build time here, rather than using [wayland-scanner's proc macros](https://docs.rs/wayland-scanner/0.31.8/wayland_scanner/) to generate the code?

Yeah there is unfortunately a lot of stuff that we have to do with these protocols. For every existing protocol, I will have to:

1. Generate a Dispatch or GlobalDispatch implementation
2. Generate an abstract C++ class that our C++ classes will inherit
3. Make the arms of the dispatchers call out to the generated C++ class

We will still have to use the proc macro for code gen, but we're in the business of generating a bunch of things for the time being :)

> And outside of the Mir-specific protocols, is there any benefit to generating code from the in-tree XML files rather than using the versions provided by wayland-rs?

I am unsure about this one, but I imagine there isn't much benefit, other than unifying how their implemented for the time being.

robert-ancell (approved):
Hard to judge without consuming the protocols but the parsing makes sense to me.

mattkae:
> Hard to judge without consuming the protocols but the parsing makes sense to me.

Agreed! Let's merge this now, since it will only get more interesting in the next PR

PR #4652, opened 2026-01-26
Hide "above" layer surfaces when a surface is fullscreened.

### LINKED ISSUE #3274: Windows marked `mir_window_state_fullscreen` appear behind shell components

tarek-y-ismail:
Closes #3274 

## What's new?

- Adds tracking for fullscreen surfaces
- Filters surfaces in the "above" layer if so, except surfaces placed _after_ the current fullscreen surface was focused.

## How to test

Scenario 1: waybar on the above layer + fullscreen app
1. Run mir with layer shell:
 ```
miral-app --add-wayland-extensions=zwlr_layer_shell_v1
```
2. Set waybar to be on the "bottom" layer (`~/.config/waybar/config`):
```
{
    "layer": "bottom", // Waybar at top layer
   ...
}
```
3. Start an application, a terminal for example
4. Maximize it, it should take up all screen area until it hits waybar.
5. Fullscreen it, it should take up the whole output
6. Change the layer configuration waybar's config to be "top"
7. Repeat step 4. Behavior should be identical
8. Repeat step 5. The application should take up all the screen except for the area used by waybar.

Scenario 2: Multiple outputs
1. Start mir with multiple virtual outputs:
```
   miral-app --add-wayland-extensions=zwlr_layer_shell_v1  --x11-output=960x500:400x400
```
2. Start a terminal via CTRL + ALT + t
3. Make sure waybar is configured on the top layer and start it
4. Fullscreen the terminal via F11
5. Waybar should be hidden on the output where the terminal is fullscreened, and still visible on the other output
6. Un-fullscreen and move the terminal to the other output, then re-fullscreen
7. Behavior should be similar to step 5.

Scenario 3: Fullscreen app and OSK
1. Run mir with all extensions:
```
miral-app --add-wayland-extensions=all
```
4. Start a terminal via CTRL + ALT + t, run `ubuntu-frame-osk`
8. Start `kate`, open any text file. OSK should pop up.
9. Press F11 to fullscreen
10. OSK should be hidden.

## Checklist

- [ ] Tests added and pass
- [ ] Move Magnifier to the overlay layer in another PR
- [ ] Spawn off an issue to fix OSKs.

### DIFF
--- a/src/server/scene/surface_stack.cpp
+++ b/src/server/scene/surface_stack.cpp
@@ -758,13 +758,34 @@ auto ms::SurfaceStack::has_fullscreen_on_output(geom::Rectangle const& output_ar
 
     if (focused_fs && focused_fs->state() == mir_window_state_fullscreen)
     {
+        auto const renderables = focused_fs->generate_renderables(compositor::CompositorID{0});
+        auto const maybe_shaped = std::ranges::any_of(
+            renderables, [](auto const& renderable) {
+                return renderable->shaped();
+            });
+
+        if (maybe_shaped)
+        {
+            return std::ranges::any_of(
+                renderables,
+                [&output_area](auto const& renderable)
+                {
+                    auto const opaque_region = renderable->opaque_region();
+                    if (!opaque_region)
+                        return false; // Shaped but no opaque region = transparent
+
+                    return std::ranges::any_of(
+                        *opaque_region,
+                        [&output_area](auto const& opaque_rect) { return opaque_rect.contains(output_area); });
+                });
+        }
+
         // Check if the fullscreen surface overlaps with this output
         auto const fs_bounds = focused_fs->input_bounds();
         auto const overlap = intersection_of(fs_bounds, output_area);
-        if (overlap.size.width.as_int() > 0 && overlap.size.height.as_int() > 0)
-        {
-            return true;
-        }
+        auto const full_overlap = overlap.size.width.as_int() > 0 && overlap.size.height.as_int() > 0;
+
+        return full_overlap;
     }
     return false;
 }

### CODE REVIEW

File: src/server/scene/surface_stack.cpp [hunk 1]
```diff
@@ -147,15 +161,61 @@ ms::SurfaceStack::~SurfaceStack() noexcept(true)
...
+    mir::log_debug("Fullscreen surface bounds: x=%d,y=%d,w=%d,h=%d",
+        focused_fs ? focused_fs->input_bounds().top_left.x.as_int() : 0,
+        focused_fs ? focused_fs->input_bounds().top_left.y.as_int() : 0,
+        focused_fs ? focused_fs->input_bounds().size.width.as_int() : 0,
+        focused_fs ? focused_fs->input_bounds().size.height.as_int() : 0);
+    mir::log_debug("Output area: x=%d,y=%d,w=%d,h=%d",
+        output_area.top_left.x.as_int(),
+        output_area.top_left.y.as_int(),
+        output_area.size.width.as_int(),
+        output_area.size.height.as_int());
```
tarek-y-ismail: Debug logs, should be removed.

src/server/scene/surface_stack.cpp
```diff
@@ -147,15 +161,61 @@ ms::SurfaceStack::~SurfaceStack() noexcept(true)
...
+    if (focused_fs && focused_fs->state() == mir_window_state_fullscreen)
+    {
+        // Check if the fullscreen surface overlaps with this output
+        auto const fs_bounds = focused_fs->input_bounds();
+        auto const overlap = intersection_of(fs_bounds, output_area);
+        if (overlap.size.width.as_int() > 0 && overlap.size.height.as_int() > 0)
+        {
+            has_fullscreen_on_output = true;
+        }
+    }
```
tarek-y-ismail: Probably possible to move this to an IILE and avoid declaring `focused_fs` at all.

src/server/scene/surface_stack.cpp
```diff
@@ -147,15 +161,61 @@ ms::SurfaceStack::~SurfaceStack() noexcept(true)
...
+                auto const surface_bounds = surface->input_bounds();
+                auto const overlap = intersection_of(surface_bounds, output_area);
+                if (overlap.size.width.as_int() > 0 && overlap.size.height.as_int() > 0)
+                {
+                    // Surface is on same output as fullscreen and was pre-existing - skip it
+                    mir::log_debug("Skipping pre-existing 'above' layer surface '%s' on output with fullscreen", surface->name().c_str());
+                    continue;
+                }
+            }
+
```
tarek-y-ismail: Maybe this is worth moving to a separate function? `maybe_filter_above_layer_surfaces`

src/server/scene/surface_stack.cpp
```diff
@@ -176,7 +236,6 @@ mc::SceneElementSequence ms::SurfaceStack::scene_elements_for(mc::CompositorID i
     }
     return elements;
 }
-
```
tarek-y-ismail, line 179: Whitespace change. Should be removed

src/server/scene/surface_stack.cpp
```diff
@@ -296,10 +357,37 @@ auto ms::SurfaceStack::surface_at(geometry::Point cursor) const
...
+                // Check if this surface overlaps with the fullscreen surface's output
+                auto const surface_bounds = surface->input_bounds();
+                auto const overlap = intersection_of(surface_bounds, fullscreen_area);
+                if (overlap.size.width.as_int() > 0 && overlap.size.height.as_int() > 0)
+                {
+                    // Surface is on same output as fullscreen and was pre-existing - skip it for input
+                    continue;
+                }
+            }
+
```
tarek-y-ismail: Same critique as above. Should be moved to its own fn

src/server/scene/surface_stack.cpp
```diff
@@ -311,7 +399,6 @@ auto ms::SurfaceStack::surface_at(geometry::Point cursor) const
 
     return {};
 }
-
```
tarek-y-ismail, line 314: Whitespace change

src/server/scene/surface_stack.h
```diff
@@ -169,6 +177,10 @@ class SurfaceStack :
     Observers observers;
     /// If not expired the screen is locked (and only surfaces that appear on the lock screen should be shown)
     std::atomic<bool> is_locked = false;
+    std::weak_ptr<Surface> focused_fullscreen_surface;
+    std::weak_ptr<Surface> focus_surface;
+    /// Set of above-layer surfaces that existed when fullscreen was activated (to be filtered)
+    std::set<Surface*> above_surfaces_to_filter;
```
tarek-y-ismail: Raw pointers aren't the best idea.

src/server/shell/abstract_shell.cpp
```diff
@@ -495,24 +496,34 @@ void msh::AbstractShell::set_focus_to(
...
     }
 
-    update_focus_locked(lock, new_focus_session, new_focus_surface);
+    // Update fullscreen tracking for per-output layer filtering
+    // Must be called outside focus_mutex to avoid lock ordering issues
+    // Cast to scene::SurfaceStack since shell::SurfaceStack doesn't have this method
+    if (auto scene_surface_stack = std::dynamic_pointer_cast<mir::scene::SurfaceStack>(surface_stack))
+    {
+        scene_surface_stack->set_focused_surface(new_focus_surface);
+    }
```
tarek-y-ismail: Dynamic casting means that this probably fits better in the API

src/server/scene/default_configuration.cpp
```diff
@@ -81,6 +81,13 @@ mir::DefaultServerConfiguration::the_surface_stack()
                  auto const wrapped = scene_surface_stack([this]()
                      { return std::make_shared<ms::SurfaceStack>(the_scene_report()); });
 
+                 // Set the FocusController on the SurfaceStack to enable per-output fullscreen filtering
+                 // This is done after construction to break the circular dependency (shell depends on surface_stack)
+                 if (auto scene_stack = std::dynamic_pointer_cast<ms::SurfaceStack>(wrapped))
+                 {
+                     scene_stack->set_focus_controller(the_focus_controller());
+                 }
+
```
tarek-y-ismail: Not too happy about this. Unfortunately, the focus controller depends on the shell which depends on the surface stack. So we have a circular dependency. This might suggest that this solution is not the best. But I'm constrained by the fact that the highest layer I can apply the fix at is in `SurfaceStack` so `BasicScreenShooter` does not break.

src/server/scene/surface_stack.cpp
```diff
@@ -123,6 +125,17 @@ struct SurfaceDepthLayerObserver : ms::NullSurfaceObserver
...
     }
 
+    void attrib_changed(ms::Surface const* surface, MirWindowAttrib attrib, int value) override
+    {
+        (void)value;
+        (void)surface;
+        if (attrib == mir_window_attrib_state)
+        {
+            // Trigger fullscreen tracking update for the focused surface
+            stack->on_surface_focused(nullptr);
```
tarek-y-ismail: The fact that I have to pass `nullptr` to force an update might hint at this needing to be its own thing.

src/server/scene/surface_stack.h
```diff
@@ -86,8 +90,20 @@ class SurfaceStack :
...
     virtual ~SurfaceStack() noexcept(true);
 
+    /// Set the FocusController to query for focused surface
+    ///
+    /// This is called after construction to break the circular dependency:
+    /// - The shell (which implements FocusController) depends on SurfaceStack
+    /// - SurfaceStack needs to query the focused surface for per-output fullscreen filtering
+    ///
+    /// The FocusController is stored as a weak_ptr, so SurfaceStack gracefully handles
+    /// cases where it's not available (e.g., during initialization or in tests).
```
tarek-y-ismail: Leaks implementation details.

src/server/scene/surface_stack.h
```diff
@@ -113,6 +129,18 @@ class SurfaceStack :
...
+    /// Notify SurfaceStack that focus has changed
+    ///
+    /// This triggers an update of fullscreen filtering state by:
+    /// - Querying the current focused surface from FocusController
+    /// - Detecting if fullscreen state has changed (entering/exiting fullscreen)
+    /// - Rebuilding the filter list of above-layer surfaces to hide when fullscreen is active
+    /// - Triggering a scene update to refresh rendering
+    ///
+    /// The surface parameter is optional - if nullptr, the focused surface is queried from FocusController.
+    /// This method is called by the shell when focus changes and by observers when surface state changes.
```
tarek-y-ismail: More leaking of impl details

src/server/scene/default_configuration.cpp
tarek-y-ismail: Another idea I had was to call `scene_elements_for`, then filter the list afterwards. Aside from being less performant, this is blocked by `SceneElement` and `Renderable` not having access to the depth layer. I'm not sure if adding it there would be an improvement.

tarek-y-ismail:
Not happy with the current implementation, but kinda out of ideas. Suggestions welcome.

tarek-y-ismail:
@copilot Would you kindly review this PR? What are the upsides and downsides of this approach? What other approaches exist?

src/server/scene/surface_stack.cpp
```diff
@@ -717,3 +761,81 @@ void ms::SessionLockObserverMultiplexer::on_unlock()
...
+            return;
+
+        cached_fullscreen_surface = new_fullscreen;
+        changed = true;
+
+        // Clear the filter list when fullscreen state changes
+        above_surfaces_to_filter.clear();
+
+        if (!new_fullscreen)
+            return;
```
mattkae: Won't this exit early and you won't get your `emit_scene_changed` below?

src/server/scene/surface_stack.cpp
```diff
@@ -717,3 +761,81 @@ void ms::SessionLockObserverMultiplexer::on_unlock()
...
+
+        // If entering fullscreen, capture current above layer surfaces to filter
+        for (auto const& layer : surface_layers)
+        {
+            for (auto const& surf : layer)
+            {
+                if (surf->depth_layer() == mir_depth_layer_above)
+                    above_surfaces_to_filter.insert(surf);
```
mattkae: Why do you need to keep this list in memory at all? Can you not just filter it when you get the scene elements?

src/server/scene/surface_stack.cpp
```diff
@@ -717,3 +761,81 @@ void ms::SessionLockObserverMultiplexer::on_unlock()
...
+    }
+    return false;
+}
+
+auto ms::SurfaceStack::should_skip_surface(
+    std::shared_ptr<Surface> const& surface, geom::Rectangle const& output_area) const -> bool
+{
+    // Skip "above" layer surfaces if they're in the pre-existing filter list
+    if (has_fullscreen_on_output(output_area) && surface->depth_layer() == mir_depth_layer_above &&
+        above_surfaces_to_filter.count(surface) > 0)
```
mattkae: Is this check not the same as on line 792?

mattkae (changes requested):
I have an initial logic check for you

src/server/scene/surface_stack.cpp
tarek-y-ismail: This was originally the case, but someone came up with the case when you fullscreen an app, _then_ place a new surface in the scene. In that case, the new surface should not be hidden. For example, if you fullscreen firefox, then open a terminal to do something before returning to firefox.

To do that, we need to know which surfaces were there _before_  the app was fullscreened and only filter those. Which is why `above_surfaces_to_filter` exists.

src/server/scene/surface_stack.cpp
tarek-y-ismail: Just tried it, it's not needed.

src/server/scene/surface_stack.cpp
tarek-y-ismail: Yes, but one is to fill the set, and the other is to look up the surface. But this has brought to my attention the fact that we don't really need to check the layer since by necessity surfaces in the set will be in the "above" layer.

src/server/scene/surface_stack.cpp
AlanGriffiths: > This was originally the case, but someone came up with the case when you fullscreen an app, _then_ place a new surface in the scene. In that case, the new surface should not be hidden. For example, if you fullscreen firefox, then open a terminal to do something before returning to firefox.

Doesn't the new surface get focus? Which implies the fullscreen surface loses focus: we don't need any filtering?

mattkae (changes requested):
I'm not in love with this solution now that I'm digging a bit deeper into it. It seems like the `SurfaceStack` shouldn't be involved with the filtering here at all, and that this should be the job of the `Shell`. 

Why can't the following happen:
1. A window gets fullscreened
2. A window gets focused
3. The shell sees this and goes "Ok, time to modify all other surfaces to be in the 'hidden' state"
4. A different window gets focused
5. The shell sees this and goes "Nice, now I have to unhide those previously hidden surfaces!"

In this way, we won't have to think of the circular dependency between the stack and the shell. In addition, the shell remains the arbiter of truth for window management things, which makes things like input focus much easier to think about.

Are there any downsides to this idea?

Issue #4653, opened 2026-01-27
How to handle keyboard input


AlanGriffiths:
### Version

v2.25.0

### Environment

n/a

### What happened? What did you expect to happen?

The example code should be compiled in CI to ensure it is correct

### Relevant log output

```shell

```

Saviq:
> Should we move the snippets into https://github.com/canonical/mir/tree/main/doc/sphinx/tutorial/first-wayland-compositor then?
>
> _Originally posted by @Saviq in https://github.com/canonical/mir/issues/4654#issuecomment-3816623502_

> You mean "make first-wayland-compositor more complex than it need be"? I think not, but that discussion belongs on the linked issue
> 
> _Originally posted by @AlanGriffiths in https://github.com/canonical/mir/issues/4654#issuecomment-3816635533_

OK so, we can put that snippet there without exposing it to the user in the tutorial, since we have control over which parts end up there.

That, or well, since the input document are diffs, it would be even more appropriate to _apply_ them before building? `quilt` might be in order?

PR #4654, opened 2026-01-27
Fix "How to handle keyboard input"

AlanGriffiths:
Fixes the example.

But also: https://github.com/canonical/mir/issues/4653

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -103,7 +103,7 @@ jobs:
 
         sudo --preserve-env apt-get build-dep --yes ./
 
-        echo PATH=${PATH}:/usr/lib/rust-1.82/bin >> $GITHUB_ENV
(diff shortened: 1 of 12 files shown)


tarek-y-ismail:
> But also: https://github.com/canonical/mir/issues/4653

*Checks 4653*:
> The example code should be compiled in CI to ensure it is correct

Saviq:
Should we move the snippets into https://github.com/canonical/mir/tree/main/doc/sphinx/tutorial/first-wayland-compositor then?

AlanGriffiths:
> Should we move the snippets into https://github.com/canonical/mir/tree/main/doc/sphinx/tutorial/first-wayland-compositor then?

You mean "make first-wayland-compositor more complex than it need be"? I think not, but that discussion belongs on the linked issue

PR #4659, opened 2026-01-30
Remove strdup & free calls, use mutable string

### LINKED ISSUE #4565: We shouldn't use `strdup` and `free` when strings exist
In `make_shm_pool`:
https://github.com/canonical/mir/blob/7a9370befcace079fcc6d2103d5a33218462f86e/src/platforms/wayland/cursor.cpp#L45-L48

We can just use a mutable string and pass it's `data()` pointer to `mkostemp` and its `c_str()` to `unlink`.

DigraJatin:
Closes #4565

As mentioned in issue, used a mutable string and passed it's data() pointer to mkostemp and its c_str() to unlink instead of using strdup() and free() methods.


DigraJatin:
TICS Check is also failing, no idea why.

AlanGriffiths:
> TICS Check is also failing, no idea why.

It passed initially. I can't access the logs from the failure - I'm assuming that it was an infrastructure problem, not yours.

PR #4662, opened 2026-02-03
Make CursorObserverMultiplexer send the initial state to newly registered CursorObservers

jhenstridge:
## What's new?

This PR alters `CursorObserverMultiplexer` to send the initial state to observers when they are registered.

While some cursor state such as the position changes frequently, others such as the cursor image or pointer usability may change less frequently. By sending the initial state to newly registered observers, we don't need a second API to request the current state.

## How to test

A simple unit test is included.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/include/server/mir/input/cursor_observer_multiplexer.h
+++ b/src/include/server/mir/input/cursor_observer_multiplexer.h
@@ -35,7 +35,6 @@ public:
     void register_early_observer(std::weak_ptr<CursorObserver> const& observer, Executor& executor) override;
     // Make the other overloads from the base class vislble
     using ObserverMultiplexer<CursorObserver>::register_interest;
-    using ObserverMultiplexer<CursorObserver>::register_early_observer;
 
     virtual void cursor_moved_to(float abs_x, float abs_y) override;
     virtual void pointer_usable() override;
--- a/src/include/server/mir/observer_multiplexer.h
+++ b/src/include/server/mir/observer_multiplexer.h
@@ -50,7 +50,7 @@ template<class Observer>
 class ObserverMultiplexer : public ObserverRegistrar<Observer>, public Observer
 {
 public:
-    void register_interest(std::weak_ptr<Observer> const& observer) override;
+    void register_interest(std::weak_ptr<Observer> const& observer) override final;
     void register_interest(
         std::weak_ptr<Observer> const& observer,
         Executor& executor) override;


src/include/server/mir/input/cursor_observer_multiplexer.h
```diff
@@ -31,6 +31,12 @@ class CursorObserverMultiplexer : public ObserverMultiplexer<CursorObserver>
     explicit CursorObserverMultiplexer(Executor& default_executor);
     virtual ~CursorObserverMultiplexer();
 
+    void register_interest(std::weak_ptr<CursorObserver> const& observer, Executor& executor) override;
+    void register_early_observer(std::weak_ptr<CursorObserver> const& observer, Executor& executor) override;
+    // Make the other overloads from the base class vislble
+    using ObserverMultiplexer<CursorObserver>::register_interest;
+    using ObserverMultiplexer<CursorObserver>::register_early_observer;
```
AlanGriffiths: This seems a questionable design: Why would we not override all the overloads?

src/server/input/cursor_observer_multiplexer.cpp
```diff
@@ -25,22 +25,60 @@ mi::CursorObserverMultiplexer::CursorObserverMultiplexer(Executor& default_execu
 
 mi::CursorObserverMultiplexer::~CursorObserverMultiplexer() = default;
 
+void mi::CursorObserverMultiplexer::register_interest(std::weak_ptr<CursorObserver> const& observer, Executor& executor)
+{
+    std::lock_guard lock{mutex};
+    ObserverMultiplexer::register_interest(observer, executor);
```
AlanGriffiths, line 31: Holding locks whilst calling other objects is fragile: It depends on them not calling back indirectly. Do we need to hold the lock while calling the base function?

src/server/input/cursor_observer_multiplexer.cpp
```diff
@@ -25,22 +25,60 @@ mi::CursorObserverMultiplexer::CursorObserverMultiplexer(Executor& default_execu
...
+    last_abs_x = abs_x;
+    last_abs_y = abs_y;
     for_each_observer(&CursorObserver::cursor_moved_to, abs_x, abs_y);
 }
 
 void mi::CursorObserverMultiplexer::pointer_usable()
 {
+    std::lock_guard lock{mutex};
+    pointer_is_usable = true;
     for_each_observer(&CursorObserver::pointer_usable);
```
AlanGriffiths, line 69: Holding locks whilst calling other objects is fragile: It depends on them not calling back indirectly. Do we need to hold the lock while calling all the observers?

tests/integration-tests/input/test_cursor_observer.cpp
```diff
@@ -98,11 +96,14 @@ TEST_F(CursorObserverIntegrationTest, cursor_observer_receives_motion)
...
+    // initial state delivered to observer
+    EXPECT_CALL(*cursor_observer, cursor_moved_to(0.0f, 0.0f)).Times(1);
+
     static const float x = 100.f;
     static const float y = 100.f;
-
     EXPECT_CALL(*cursor_observer, cursor_moved_to(x, y)).Times(1).WillOnce(mt::WakeUp(signal));
 
+    cursor_observer_multiplexer->register_interest(cursor_observer);
     fake_mouse->emit_event(mis::a_pointer_event().with_movement(x, y));
```
AlanGriffiths: This conflates expectations related to the test setup with those actually part of the `cursor_observer_receives_motion` test:

In `SetUp()`
```c++
    // During setup allow initial state to be delivered to observer
    ASSERT_CALL(*cursor_observer, cursor_moved_to(_, _)).Times(AtMost(1));
    cursor_observer_multiplexer->register_interest(cursor_observer);
    VerifyAndClearExpectations(cursor_observer);
```

In `cursor_observer_receives_motion`
```c++
    static const float x = 100.f;
    static const float y = 100.f;

    EXPECT_CALL(*cursor_observer, cursor_moved_to(x, y)).Times(1).WillOnce(mt::WakeUp(signal));
    fake_mouse->emit_event(mis::a_pointer_event().with_movement(x, y));
```

src/include/server/mir/input/cursor_observer_multiplexer.h
RAOF: Yeah. I can see the reasoning behind doing this, though; perhaps we should mark these overloads `final` in `observer_multiplexer.h` and document that they defer to the two-parameter overloads so classes which need to customise registration only need to do so in one place?

Alternatively, have an explicit extension point?

jhenstridge: The overloads I haven't overridden call the ones I have. Also, the base class versions rely on the private `default_executor` member. So overriding those methods would either just call the base class versions, or I'd need to keep track of the default executor separately to call the other overload.

Making those other overloads `final` so there's only one pair methods to override seems like a good idea. We'd still need the `using` directive to make them visible though.

src/server/input/cursor_observer_multiplexer.cpp
jhenstridge: The lock is being used to guarantee that the initial condition events are sent before any others. If we did not hold the lock here, it's possible e.g. a `cursor_moved_to` event might be sent to the new observer before the initial state events are sent through.

I used the `mir::graphics::DisplayConfigurationObserverMultiplexer` class as a model, which behaves in a similar fashion.

src/server/input/cursor_observer_multiplexer.cpp
jhenstridge: The lock is necessary to guarantee the order that events are sent to the observers. I agree that this could be a problem for observers using `immediate_executor`, but in practice the execution is going to be deferred: either to another thread, or to be run by the main loop.

tests/integration-tests/input/test_cursor_observer.cpp
jhenstridge: I'll rework the change to be closer to this. The code in SetUp() will likely be a little more complicated, since the method call will be deferred to the main loop, similar to the later call.

PR #4664, opened 2026-02-04
Show/Hide attached surfaces when entering/exiting from fullscreen

### LINKED ISSUE #3274: Windows marked `mir_window_state_fullscreen` appear behind shell components

tarek-y-ismail:
Closes #3274 

## What's new?

- Adds code `BasicWindowManager` to hide/restore attached surfaces in the "above" layer when a surface is fullscreened, unfullscreened, and removed.

## How to test

Scenario 1: waybar on the above layer + fullscreen app
1. Run mir with layer shell:
 ```
miral-app --add-wayland-extensions=zwlr_layer_shell_v1
```
2. Set waybar to be on the "bottom" layer (`~/.config/waybar/config`):
```
{
    "layer": "bottom", // Waybar at top layer
   ...
}
```
3. Start an application, a terminal for example
4. Maximize it, it should take up all screen area until it hits waybar.
5. Fullscreen it, it should take up the whole output
6. Change the layer configuration waybar's config to be "top"
7. Repeat step 4. Behavior should be identical
8. Repeat step 5. Behavior should be identical

Scenario 2: Multiple outputs
1. Start mir with multiple virtual outputs:
```
   miral-app --add-wayland-extensions=zwlr_layer_shell_v1  --x11-output=960x500:400x400
```
2. Start a terminal via CTRL + ALT + t
3. Make sure waybar is configured on the top layer and start it
4. Fullscreen the terminal via F11
5. Waybar should be hidden on the output where the terminal is fullscreened, and still visible on the other output
6. Un-fullscreen and move the terminal to the other output, then re-fullscreen
7. Behavior should be similar to step 5.

Scenario 3: Fullscreen app and OSK
1. Run mir with all extensions:
```
miral-app --add-wayland-extensions=all
```
2. Start a terminal via CTRL + ALT + t, run `ubuntu-frame-osk`
3. Start `kate`, open any text file. OSK should pop up.
4. Press F11 to fullscreen
5. OSK should be hidden.

Preferably, we'd like the OSK to be visible when an app is fullscreened, but since it's placed in the "above" layer, and we have no way in the layer shell protocol to identify it, it will be broken for now.

Scenario 4: Waybar + Fullscreen app + non-fullscreen app
This follows scenario 1, with the following extra steps
1. Run a different fullscreen app, a terminal via CTRL + ALT + t for example
2. Focus should switch to it, way bar should be visible once more
3. If you use the alt-tab switcher, waybar should hide and be restored as you focus on the fullscreened and non-fullscreened apps respectively.
4. Changing focus via the pointer should behave similarly to step 3

Scenario 5: Waybar + Fullscreen app exits
Repeat scenario 1. Then close the app via a shortcut. If using a terminal, CTRL + D. Waybar should pop into view once more.

Scenario 5: Waybar + Fullscreen app with child surfaces
1. Repeat scenario 1, but with VLC.
2. With VLC fullscreened, open any entry from the menu bar
3. Waybar should remain hidden
4. Open a dialog via `Media > Open File` or Ctrl +  o,
5. Way bar should remain hidden.

Scenario 6:
Same as scenario 4, but instead of switching focus between the two apps, you exit the non-fullscreen app first.

Focus should switch to the fullscreen app, and waybar should remain hidden.

## Checklist
- [x] Tests added and pass (coming (if possible) once this approach is approved)
- [ ] Adequate documentation added (coming (if possible) once this approach is approved)

### CODE REVIEW

File: src/miral/basic_window_manager.cpp [hunk 1]
```diff
@@ -49,6 +49,54 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+        auto& info = bwm.info_for(window);
+        if (info.depth_layer() != mir_depth_layer_above)
+            continue;
+
+        auto const mir_surface = std::shared_ptr<scene::Surface>(window);
+
+        hidden_attached_window_states.push_back({window, info.state()});
+        bwm.policy->advise_state_change(info, mir_window_state_hidden);
+        info.state(mir_window_state_hidden);
+        mir_surface->hide();
```
tarek-y-ismail: These lines are borrowed from `set_state`. It unfortunately calls `update_attached_and_fullscreen_sets`, which itself may modify `attached_windows`, which we are iterating over.

tarek-y-ismail: Additionally, `ms::Surface::configure` was not borrowed for two reasons:
1. It is legacy code
2. It causes waybar to restore to the middle of the screen, not the top (or wherever it was configured)

src/miral/basic_window_manager.cpp
```diff
@@ -1441,16 +1497,27 @@ void miral::BasicWindowManager::update_attached_and_fullscreen_sets(WindowInfo c
...
     {
         fullscreen_surfaces.erase(window);
+
+        // Account for when a child window pops up (menubar dropdowns, etc..)
+        auto const any_parent_is_fullscreen = is_window_or_parent_fullscreen(window, *this);
+        if(!any_parent_is_fullscreen && window_info.depth_layer() == mir_depth_layer_application)
+            display_area->restore_all_attached(*this);
     }
 
+
```
AlanGriffiths:
```suggestion
```

src/miral/basic_window_manager.h
```diff
@@ -207,6 +207,10 @@ class BasicWindowManager : public virtual mir::shell::WindowManager,
         {
         }
 
+        void hide_all_attached(BasicWindowManager&);
+
+        void restore_all_attached(BasicWindowManager&);
+
```
AlanGriffiths: No parameter names?

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,54 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+        mir_surface->hide();
+    }
+}
+
+void miral::BasicWindowManager::DisplayArea::restore_all_attached(BasicWindowManager& bwm)
+{
+    for (auto& [window, prev_state] : hidden_attached_window_states)
+    {
+        auto& info = bwm.info_for(window);
+
```
AlanGriffiths: Probably worth confirming that the window is still valid before using it

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,54 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+        bwm.policy->advise_state_change(info, mir_window_state_hidden);
+        info.state(prev_state);
+        auto const mir_surface = std::shared_ptr<scene::Surface>(window);
+        mir_surface->show();
```
AlanGriffiths: If `prev_state` is `mir_window_state_hidden`, then we shouldn't be calling `show()`

src/miral/basic_window_manager.cpp
AlanGriffiths: I don't have an issue with the current approach, but to answer your points above...

> `set_state`. It unfortunately calls `update_attached_and_fullscreen_sets`, which itself may modify `attached_windows`, which we are iterating over.

We can, in principle, avoid that by iterating over a copy of `attached_windows`. Or by "extract method" refactoring the common code.

> * It causes waybar to restore to the middle of the screen, not the top (or wherever it was configured)

What is the mechanism of that? (We should understand to be sure the current approach hasn't a problem that will surface later.)

src/miral/basic_window_manager.cpp
```diff
@@ -280,6 +328,14 @@ void miral::BasicWindowManager::remove_window(Application const& application, mi
 
     application->destroy_surface(info.window());
 
+    // If a fullscreen window is closed, restore any hidden attached surfaces
+    bool const was_fullscreen{info.state() == mir_window_state_fullscreen};
+    if (was_fullscreen)
```
AlanGriffiths: No need for a variable (especially with scope beyond the if statement).

But... does this code belong here anyway? I mean, if the focus shifts to another fullscreen window we need to undo this work.

Instead, when a window (or no window) gets focus we should be checking the state of the attached windows and acting accordingly.

src/miral/basic_window_manager.cpp
```diff
@@ -1441,16 +1497,27 @@ void miral::BasicWindowManager::update_attached_and_fullscreen_sets(WindowInfo c
...
+            display_area->hide_all_attached(*this);
     }
     else
     {
         fullscreen_surfaces.erase(window);
+
+        // Account for when a child window pops up (menubar dropdowns, etc..)
+        auto const any_parent_is_fullscreen = is_window_or_parent_fullscreen(window, *this);
+        if(!any_parent_is_fullscreen && window_info.depth_layer() == mir_depth_layer_application)
+            display_area->restore_all_attached(*this);
```
AlanGriffiths: Do we know this surface has focus?

AlanGriffiths (changes requested):
No tests?

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,54 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+}
+
+void miral::BasicWindowManager::DisplayArea::hide_all_attached(BasicWindowManager& bwm)
+{
+    for (auto& window : attached_windows)
+    {
+        auto& info = bwm.info_for(window);
+        if (info.depth_layer() != mir_depth_layer_above)
+            continue;
+
```
mattkae:
```suggestion
```

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,54 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+        auto& info = bwm.info_for(current);
+        if (info.state() == mir_window_state_fullscreen)
+            return true;
+        current = info.parent();
+    }
+    return false;
+}
+}
+
+void miral::BasicWindowManager::DisplayArea::hide_all_attached(BasicWindowManager& bwm)
```
mattkae: Can we comment that it is a precondition for this method that `hidden_attached_window_states` is currently empty?

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,54 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+    for (auto& window : attached_windows)
+    {
+        auto& info = bwm.info_for(window);
+        if (info.depth_layer() != mir_depth_layer_above)
+            continue;
+
+
+        hidden_attached_window_states.push_back({window, info.state()});
+        bwm.policy->advise_state_change(info, mir_window_state_hidden);
+        info.state(mir_window_state_hidden);
```
mattkae: Why are you not just using `bwm.set_state` here? That will do all of the relevant hiding things and keep that information localized.

src/miral/basic_window_manager.cpp
mattkae: Same here, why not just use `bwm.set_state()`

mattkae (dismissed):
An initial ask!

src/miral/basic_window_manager.h
tarek-y-ismail: Are they necessary in this case?

src/miral/basic_window_manager.cpp
tarek-y-ismail: Figured it out. Mir was trying to restore the surface whenever waybar tried submitting a new buffer. 

Details: https://matrix.to/#/!ivAdrrLoWkBYEnSIae:matrix.org/$UE4Z4bLIvmGEuweQFRiQwM-FVY-h8eIUKUJc1YQyhqw?via=matrix.org&via=ubuntu.com&via=fedora.im

src/miral/basic_window_manager.cpp
tarek-y-ismail: I looked into it. This is the latest we can do this check. It was added specifically for the case where you have a fullscreen window, and you close it without exiting fullscreen first. 

In that case, a few lines later, the info for this window will be erased before we reach the focus checking code, so we won't be able to move this code there.

tarek-y-ismail:
Something I've just realized is that I haven't taken into account cases where focus changes (or a surface is closed) with the closed surface and the newly focused surface on different outputs/display areas. So far I've assumed both are on the same display area.

---

Another thing I should mention is that focus handling feels quite messy at the moment. For example, when we have two apps open. Let's call them A and B. If we close A, `remove_window` gets called which removes the window and updates the MRU window stack among other things. `active_window()` now returns B . 

Context: 
https://github.com/canonical/mir/blob/38d8e69d2c56b139fbc3623bfa76101b9f793bc2/src/miral/basic_window_manager.cpp#L327-L349

https://github.com/canonical/mir/blob/38d8e69d2c56b139fbc3623bfa76101b9f793bc2/src/miral/basic_window_manager.cpp#L578-L581

We also call `refocus`, for which itself traverses the MRU window stack from top to bottom, starting with B. 

https://github.com/canonical/mir/blob/38d8e69d2c56b139fbc3623bfa76101b9f793bc2/src/miral/basic_window_manager.cpp#L385-L402

Then that calls `set_active_window`, and gets the active window, which is B.
https://github.com/canonical/mir/blob/38d8e69d2c56b139fbc3623bfa76101b9f793bc2/src/miral/basic_window_manager.cpp#L1703-L1703

Then we execute the rest of `set_active_window`, which at that point we don't have information about the window we just removed!

---

Edit: I just realized that I've worked on this _alone_ for 120% of my time today. So there might be a solution to the mini-rant above that my neurons are just too fried to see right now.

tarek-y-ismail:
Bug: With two outputs, if we:
1.  start a terminal (A), 
2. start waybar, full screen the terminal, 
3. and open a new terminal (B), waybar is restored as expected. 
4. When terminal B is dragged to the second output, I believe the first output should hide waybar again. Will see if I can fix this, but leaving this here in case I can't immediately do so.


Edit: After thinking about it properly, no. waybar should not be automatically hidden. It will be hidden again if we focus terminal A, but it doesn't need to be hidden when focus is not on it.

src/miral/basic_window_manager.cpp
```diff
@@ -927,6 +992,78 @@ auto miral::BasicWindowManager::collect_windows(const miral::WindowInfo &info) -
...
+                {
+                    auto const& info = info_for(fs);
+                    return info.depth_layer() == mir_depth_layer_application &&
+                           display_area_for(info) == prev_display_area;
+                });
+
+            if (prev_was_fullscreen && !any_fullscreen_apps_on_prev)
+                prev_display_area->restore_all_attached(*this);
+            else if (!prev_was_fullscreen && any_fullscreen_apps_on_prev)
+                prev_display_area->hide_all_attached(*this);
```
tarek-y-ismail: Slightly wrong. The decision should be based on `current_active`.

src/miral/basic_window_manager.h
AlanGriffiths: What does "necessary" mean? Obviously "it compiles" as it is. But the project style is to provide names in function declarations.

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,58 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+    {
+        auto& info = bwm.info_for(window);
+
+        if (info.depth_layer() != mir_depth_layer_above)
+            continue;
+
+        hidden_attached_windows.push_back(window);
+        bwm.set_state(info, mir_window_state_hidden);
+    }
+}
```
AlanGriffiths, line 76: This mixing of logic across `BasicWindowManager` and `DisplayArea` seems inelegant.

Another way to divide the logic:

```c++
void miral::BasicWindowManager::DisplayArea::update_hidden_attached(std::function<bool(Window const& w) predicate)
{
    // Copy as `predicate` may modify `attached_windows` and invalidate the iterator.
    auto const attached_windows = this->attached_windows;
    decltype(hidden_attached_windows) hidden_attached_windows;

    for (auto& window : attached_windows)
    {
        if (predicate(window))
        {
            hidden_attached_windows.push_back(window);
        }
    }

    this->hidden_attached_windows = std:move(hidden_attached_windows);
}
```

```C++
            current_display_area->update_hidden_attached([this](Window const& w)
                {
                     auto& info = info_for(window);
                     if (info.depth_layer() != mir_depth_layer_above) return false;
                     set_state(info, mir_window_state_hidden);
                     return true;
                });
```

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,58 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+        auto& info = bwm.info_for(current);
+        if (info.state() == mir_window_state_fullscreen)
+            return true;
+        current = info.parent();
+    }
+    return false;
+}
+}
+
+void miral::BasicWindowManager::DisplayArea::hide_all_attached(BasicWindowManager& bwm)
```
AlanGriffiths: The name is wrong - only some attached are hidden.

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,58 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+        return false;
+
+    auto current = window;
+    while (current)
+    {
+        auto& info = bwm.info_for(current);
+        if (info.state() == mir_window_state_fullscreen)
+            return true;
+        current = info.parent();
+    }
```
AlanGriffiths: Isn't this a weirdly written `for` loop?

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,58 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
 {
     rect.top_left.y = zone.top_left.y + 0.5 * std::max(DeltaY{}, zone.size.height - rect.size.height);
 }
+
+auto is_window_or_parent_fullscreen(miral::Window const& window, miral::BasicWindowManager& bwm) -> bool
+{
+    if (!window)
+        return false;
```
AlanGriffiths: Redundant check?

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,58 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
 {
     rect.top_left.y = zone.top_left.y + 0.5 * std::max(DeltaY{}, zone.size.height - rect.size.height);
 }
+
+auto is_window_or_parent_fullscreen(miral::Window const& window, miral::BasicWindowManager& bwm) -> bool
```
AlanGriffiths: Simpler as a `BasicWindowManager` member function?

src/miral/basic_window_manager.cpp
```diff
@@ -49,6 +49,58 @@ void clamp_top(Rectangle const& zone, Rectangle& rect)
...
+    {
+        if(!window)
+            continue;
+
+        auto& info = bwm.info_for(window);
+        bwm.set_state(info, mir_window_state_attached);
+    }
+
+    hidden_attached_windows.clear();
+}
```
AlanGriffiths: Name wrong: 

1. it is not _all_ attached that are restored; and, 
2. "restored" is a window state distinct from "attached"

src/miral/basic_window_manager.cpp
```diff
@@ -221,6 +273,13 @@ void miral::BasicWindowManager::modify_surface(
     auto& info = info_for(surface);
     WindowSpecification mods{modifications};
     validate_modification_request(mods, info);
+
+    // If a layer shell surface (waybar) tries to submit a buffer while
+    // its hidden (in favor of another fullscreen surface), don't try
+    // to modify its placement.
+    if (display_area_for(info)->is_hidden_attached(info))
+        mods.state().consume();
```
AlanGriffiths, line 267: This seems overbroad: it bans any transitions. E.g. a panel that decides to transition from "attached" to "restored" (and, maybe, change layer too).

AlanGriffiths (changes requested):
This seems messy.

I don't have a clear solution to offer, but trying to model a window state of "attached and hidden" as "hidden and in a list" introduces a lot of corner cases that might be avoided with another approach. (A separate "hidden" property or a new "attached and hidden" state.)

src/miral/basic_window_manager.cpp
tarek-y-ismail: I thought `miral::Window` was more of a weak pointer than a shared pointer, but looking into the implementation code, it looks like a shared one. Thanks for catching this.

src/miral/basic_window_manager.cpp
tarek-y-ismail: The problem is that some bars (waybar, ironbar, etc...) update themselves even after being set to hidden. [This thread on matrix](https://matrix.to/#/!ivAdrrLoWkBYEnSIae:matrix.org/$UE4Z4bLIvmGEuweQFRiQwM-FVY-h8eIUKUJc1YQyhqw?via=matrix.org&via=ubuntu.com&via=fedora.im) has all the details.

AlanGriffiths: I don't see much detail:

> tarek-y-ismail
> I also may have discovered a Waybar bug...
> tarek-y-ismail
> Doesn't happen with Yambar, happens with ironbar
> tarek-y-ismail
> Also breaks with gBar

What are these components actually doing?

tarek-y-ismail: Oops, I may have supplied the wrong link. Please try [this](https://matrix.to/#/!ivAdrrLoWkBYEnSIae:matrix.org/$VxmfJvp6IspC_Az3zkPltQPVE4xCFYtw4xUk2r_DMm4?via=matrix.org&via=ubuntu.com&via=fedora.im)

AlanGriffiths: It may be the trouble I have following these links

tarek-y-ismail: > Ok I think I figured out what's happening.
> 
> When we enter fullscreen, we set the window's state to hidden, which works fine.
> 
> The bug occurs when the bar submits a frame. This piece of code here sees that the surface is not mapped, but it submitted a buffer, so it should be mapped. It then sets its state to restored: https://github.com/canonical/mir/blob/7e7e41f259a0cba81e870467d488d1dda37913a2/src/server/frontend_xwayland/xwayland_surface_role.cpp#L136-L141
> 
> Execution reaches this piece of code, where it sees the state as restored and sees that the restore rect does not overlap the application zone, because layer shell zones may have an exclusion zone, which shrinks the application zone: https://github.com/canonical/mir/blob/7e7e41f259a0cba81e870467d488d1dda37913a2/src/miral/basic_window_manager.cpp#L1445-L1452
> 
> Then clamp_top actually does not clamp, it centers the surface vertically: https://github.com/canonical/mir/blob/7e7e41f259a0cba81e870467d488d1dda37913a2/src/miral/basic_window_manager.cpp#L48-L51
> 
> Which leads to the bars (see next thread) appearing in the middle of the screen.
>

src/miral/basic_window_manager.cpp
```diff
@@ -673,11 +720,11 @@ auto miral::BasicWindowManager::display_area_for(WindowInfo const& info) const -
...
-        for (auto& area_window : area->attached_windows)
-        {
-            if (area_window == window)
-                return area;
-        }
+        if(std::ranges::contains(area->attached_windows, window))
+            return area;
+
+        if(std::ranges::contains(area->hidden_attached_windows, window))
+            return area;
```
AlanGriffiths:
```suggestion
        if (std::ranges::contains(area->attached_windows, window) ||
            std::ranges::contains(area->hidden_attached_windows, window))
            return area;
```

But... isn't `hidden_attached_windows` a subset of `attached_windows` anyway?

src/miral/basic_window_manager.cpp
```diff
@@ -927,6 +974,94 @@ auto miral::BasicWindowManager::collect_windows(const miral::WindowInfo &info) -
...
+    auto const prev_display_area = display_area_for(prev_info);
+
+    handle_attached_surfaces_for_window_removal(prev_was_fullscreen, prev_display_area);
+}
+
+void miral::BasicWindowManager::handle_attached_surfaces_for_window_removal(
+    bool prev_was_fullscreen, std::shared_ptr<DisplayArea> const& prev_display_area)
+{
+    auto const current = active_window();
+    auto const cleanup = [prev_was_fullscreen, prev_display_area, this]()
```
AlanGriffiths: No need to capture `prev_display_area` by value?

AlanGriffiths (changes requested):

src/miral/basic_window_manager.cpp
AlanGriffiths: OK so there's a mess: this logic is clearly simplistic...

```c++
        bool const is_mapped = surface.value()->visible();
        bool const should_be_mapped = static_cast<bool>(wl_surface->buffer_size());
        if (!is_mapped && should_be_mapped && surface.value()->state() == mir_window_state_hidden)
        {
            spec.state = mir_window_state_restored;
        }
        else if (is_mapped && !should_be_mapped)
        {
            spec.state = mir_window_state_hidden;
        }
```
...not only does it ignore the existence of attached windows, it ignores fullscreen, maximized, vertmaximized,, etc.

For this approach to work we should maintain the correct state to "unhide" to. (The example `FloatingWindowManagerPolicy` does something along these lines with `PolicyData::old_state`.)

But that isn't the end of the story.

In `BasicWindowManager::modify_surface()` we should recognise that the client is trying to show a window we want hidden and attached and prevent that. (I.e. we should not get to `BasicWindowManager::update_attached_and_fullscreen_sets()` in this scenario.)

AlanGriffiths (changes requested):
I think there's a cascade of problems caused by conflating "not visible" with `mir_window_state_hidden`.

In the general case, "hidden" is an aspect of window state orthogonal to "fullscreen", "maximized", "attached", etc. (Yes, it is implied by `mir_window_state_hidden` but can co-exist with the others.)

Maybe things would be simpler if we were only be setting `BasicSurface::State::hidden` without changing the state from `attached`.

AlanGriffiths:
> In the general case, "hidden" is an aspect of window state orthogonal to "fullscreen", "maximized", "attached", etc. (Yes, it is implied by `mir_window_state_hidden` but can co-exist with the others.)
> 
> Maybe things would be simpler if we were only be setting `BasicSurface::State::hidden` without changing the state from `attached`.

I just came across `SurfaceStateTracker` - which sounds like it is doing what we need for this.

AlanGriffiths:
> I just came across `SurfaceStateTracker` - which sounds like it is doing what we need for this.

Introduced here: https://github.com/canonical/mir/pull/2015

tarek-y-ismail:
@AlanGriffiths spawned a few PRs off of your feedback as it didn't seem directly applicable to this PR. The commit history needs some cleaning up here. Will do that once you give it the thumbs up.

src/miral/basic_window_manager.cpp
AlanGriffiths: Summarising conversation from #4694:

The code quoted above does not distinguish between the window manager hiding the window and the client hiding the window. The current behaviour is correct for the client hiding the window, but wrong for the window manager hiding the window in the case of shell surfaces hidden when an app fullscreens.

`SurfaceStateTracker` has logic covering the current case but cannot be applied for everything: we shouldn't be breaking the behaviour client expects. (In particular, losing the "attached" status as described above.)

tl;dr: this still needs fixing

src/miral/basic_window_manager.cpp
```diff
@@ -677,11 +724,8 @@ auto miral::BasicWindowManager::display_area_for(WindowInfo const& info) const -
 
     for (auto& area : display_areas)
     {
-        for (auto& area_window : area->attached_windows)
-        {
-            if (area_window == window)
-                return area;
-        }
+        if(std::ranges::contains(area->attached_windows, window))
```
AlanGriffiths:
```suggestion
        if (std::ranges::contains(area->attached_windows, window))
```

src/miral/basic_window_manager.cpp
```diff
@@ -1680,6 +1825,10 @@ auto miral::BasicWindowManager::select_active_window(Window const& hint) -> mira
             policy->advise_focus_lost(info_for(prev_window));
 
         policy->advise_focus_gained(info_for_hint);
+
+        if(prev_window != hint)
+            handle_attached_surfaces_for_focus_change(prev_window, hint);
```
AlanGriffiths, line 1830: There's an _apparent_ mismatch between the supplied "`prev_window, hint`" and the parameter names "`prev, current`". But I guess that's pre-existing.

AlanGriffiths (changes requested):

src/miral/basic_window_manager.cpp
tarek-y-ismail: Before I forget all about this. 

`set_state` is called in three locations: in `modify_window`, and when hiding/reattaching windows.

Right now what we do should work, as attached windows are returned to `attached` when visible once again. This covers `set_state` calls from hiding/reattaching code. If we really want to fix this properly for all other cases, we'll have to distinguish `modify_window` calls from the client and internally from the compositor. And once that's done, we should be able to apply a solution similar to the on in #4694 

In any case, this PR should move forward, and the proper fix should be in its own PR since this issue is pre-existing.

AlanGriffiths:
No changes for a couple of months: Marking draft until review comments can be addressed

PR #4668, opened 2026-02-05
Generating empty Dispatch and GlobalDispatch implementations for each interface in wayland-rs

mattkae:
## What's new?
- Implemented `write_dispatch_rs` which writes the `Dispatch` and `GlobalDispatch` implementations for each Wayland interface to `src/dispatch.rs`
- The generated code contains no-ops for each incoming request as well as a generic data binding function
- Updated the protocol parser to check whether or not an interface is global so that we can properly generate the `GlobalDispatch` implementation
- Simplified how metadata is handled for the `WaylandProtocol` objects by removing the tuple

## How to test
1. Run `cargo build`
2. See that `src/dispatch.rs` gets generated with a bunch of no-op implementations

image

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/wayland/wayland_rs/build_script/main.rs
+++ b/src/wayland/wayland_rs/build_script/main.rs
@@ -221,8 +221,8 @@ fn generate_dispatch_impl(
 ) -> TokenStream {
     let interface_name = dash_name_to_snake_case_ident(&interface.name);
 
-    if interface_name == "wl_display" {
-        // wl_display is handled specially in wayland_server crate via the 'Display' struct.
+    if interface_name == "wl_display" || interface_name == "wl_registry" {
+        // wl_display and wl_registry are handled specially in wayland_server crate via the 'Display' struct.
         return quote! {};
     }
 


src/wayland/wayland_rs/build_script/main.rs
```diff
@@ -135,6 +138,130 @@ fn write_protocols_rs(protocols: &Vec<(WaylandProtocolMetadata, WaylandProtocol)
     write_generated_rust_file(generated_protocol_rs, "protocols.rs");
 }
 
+fn write_dispatch_rs(protocols: &Vec<WaylandProtocol>) {
+    // Generate the GlobalDispatch and Dispatch implementations for each protocol.
+    let generated_dispatch_implementations = protocols.iter().map(|protocol| {
+        let namespace_name = if protocol.name == "wayland" {
+            quote! { wayland_server::protocol }
+        } else {
+            let protocol_module = format_ident!("{}", protocol.name.replace('-', "_"));
```
robert-ancell: This mapping is done a number of times - suggest moving to a shared function to make it more readable.

src/wayland/wayland_rs/build_script/main.rs
```diff
@@ -135,6 +138,130 @@ fn write_protocols_rs(protocols: &Vec<(WaylandProtocolMetadata, WaylandProtocol)
...
+                        #namespace_name::#interface_name::Request::#request_name { #( #arg_names ),* } => {
+                            // Handle the #request_name request here
+                        }
+                    })
+                } else {
+                    None
+                }
+            });
+
+            // If the interface comes form wayland, we need to generate an empty arm because the
```
robert-ancell: form -> from

src/wayland/wayland_rs/build_script/protocol_parser.rs
```diff
@@ -206,20 +208,58 @@ pub fn parse_protocols() -> Vec<(WaylandProtocolMetadata, WaylandProtocol)> {
...
                         }
                         _ => {}
                     }
                 }
             }
 
-            protocols.push((metadata, protocol));
+            // Check if the interface is global or not.
+            // An interface is considered global if it is not created by any other interface's request or event.
+            let global_interfaces: Vec<bool> = protocol
```
robert-ancell, line 226: nit: This is heavily nested - could we split it into some functions? The double break also implies this could be a bit cleaner with functions.

robert-ancell (approved):
Some small stuff suggested but otherwise looks good. Probably better to get a review from @RAOF to check the Rust style.

PR #4669, opened 2026-02-06
Basic window manager cleanups

tarek-y-ismail:
Just some cleanups that I found while working on #4664 and don't quite fit there.

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -40,12 +40,16 @@ using namespace mir::geometry;
 
 namespace
 {
-void clamp_left(Rectangle const& zone, Rectangle& rect)
+// Attempts to center the given rect horizontally within the given zone, but
+// clamps to the left edge of the zone if the rect is wider than the zone.
+void center_horizontally_with_clamp_left(Rectangle const& zone, Rectangle& rect)
 {
     rect.top_left.x = zone.top_left.x + 0.5 * std::max(DeltaX{}, zone.size.width - rect.size.width);
 }
 
-void clamp_top(Rectangle const& zone, Rectangle& rect)
+// Attempts to center the given rect vertically within the given zone, but
+// clamps to the top edge of the zone if the rect is taller than the zone.
+void center_vertically_with_clamp_top(Rectangle const& zone, Rectangle& rect)
 {
     rect.top_left.y = zone.top_left.y + 0.5 * std::max(DeltaY{}, zone.size.height - rect.size.height);
 }
@@ -1391,8 +1395,8 @@ void miral::BasicWindowManager::place_and_size_for_state(
         rect = restore_rect;
         if (!rect.overlaps(application_zone))
         {
-            clamp_left(application_zone, rect);
-            clamp_top(application_zone, rect);
+            center_horizontally_with_clamp_left(application_zone, rect);
+            center_vertically_with_clamp_top(application_zone, rect);
         }
         rect = policy->confirm_placement_on_display(window_info, new_state, rect);
         break;
@@ -1406,7 +1410,7 @@ void miral::BasicWindowManager::place_and_size_for_state(
         rect.size = {application_zone.size.width, restore_rect.size.height};
         if (!rect.overlaps(application_zone))
         {
-            clamp_top(application_zone, rect);
+            center_vertically_with_clamp_top(application_zone, rect);
         }
         rect = policy->confirm_placement_on_display(window_info, new_state, rect);
         break;
@@ -1416,7 +1420,7 @@ void miral::BasicWindowManager::place_and_size_for_state(
         rect.size = {restore_rect.size.width, application_zone.size.height};
         if (!rect.overlaps(application_zone))
         {
-            clamp_left(application_zone, rect);
+            center_horizontally_with_clamp_left(application_zone, rect);
         }
         rect = policy->confirm_placement_on_display(window_info, new_state, rect);
         break;


src/miral/basic_window_manager.cpp
```diff
@@ -40,12 +40,12 @@ using namespace mir::geometry;
 
 namespace
 {
-void clamp_left(Rectangle const& zone, Rectangle& rect)
+void center_horizontally_in_zone(Rectangle const& zone, Rectangle& rect)
```
AlanGriffiths: Isn't the new name also confusing? If `rect` is wider than `zone` the placement is not centered horizontally, but clamped to the left. (No, that isn't a defense of the old name.)

src/miral/basic_window_manager.cpp
```diff
@@ -40,12 +40,12 @@ using namespace mir::geometry;
...
 {
     rect.top_left.x = zone.top_left.x + 0.5 * std::max(DeltaX{}, zone.size.width - rect.size.width);
 }
 
-void clamp_top(Rectangle const& zone, Rectangle& rect)
+void center_vertically_in_zone(Rectangle const& zone, Rectangle& rect)
```
AlanGriffiths: Isn't the new name also confusing? If `rect` is taller than `zone` the placement is not centered vertically, but clamped to the top.

AlanGriffiths (changes requested):
Sorry, don't have good suggestions: `center_horizontally_with_clamp_left()` is just as confusing and even longer

tarek-y-ismail:
> center_horizontally_with_clamp_left

I believe this would be okay with a comment explaining when it centers and when it clamps. What do you think?

Alternative: `center_horizontally_or_align_left`

src/miral/basic_window_manager.h
```diff
@@ -332,6 +332,8 @@ class BasicWindowManager : public virtual mir::shell::WindowManager,
     void for_each_descendent_in(WindowInfo const& info, std::function<void(const Window&)> func);
     /// Gathers windows provided WindowInfo
     auto collect_windows(WindowInfo const& info) -> SurfaceSet;
+
+    void find_replacement_focus_target(WindowInfo& info);
```
mattkae: This method doesn't just find and return, it sets the focus replacement:

```suggestion
    void set_focus_replacement_for(WindowInfo& info);
```

AlanGriffiths:
> > center_horizontally_with_clamp_left
> 
> I believe this would be okay with a comment explaining when it centers and when it clamps. What do you think?

I don't have a better idea

src/miral/basic_window_manager.cpp
```diff
@@ -1470,6 +1474,63 @@ void miral::BasicWindowManager::update_attached_and_fullscreen_sets(WindowInfo c
...
 
+void miral::BasicWindowManager::set_focus_replacement_for(WindowInfo& window_info)
+{
+    // This is a bit hacky as it relies on non-local knowledge, but it works without adding complexity elsewhere.
+    //
+    // We set the state to mir_window_state_hidden (even if value is mir_window_state_minimized) before
+    // looking for another window in the same "tree" to give focus. This means we can call the widely used
+    // function select_active_window() which would otherwise restore the window we want to minimize.
+    //
+    // We set the correct state after working out the new active window.  alan_g
```
AlanGriffiths: This comment was true in the previous context, but is no longer true.

1. it assumes the window state is `mir_window_state_hidden` or `mir_window_state_minimized`; and,
2. the correct state is not being restored in the extracted code but in a line that was not extracted

tarek-y-ismail: 1. This point  is still true because this is only called in one place where the state is `mir_window_state_hidden` or `mir_window_state_minimized`. If you're worried that it might be misused in the future, passing the state and asserting inside should be enough feedback to other programmers that they misused this function
2. My bad. I assumed the whole block was related to the block below it and didn't read it thoroughly. Should be easy enough to remove the rest of the switch case inside.

AlanGriffiths: > 1. This point  is still true because this is only called in one place

It used to be that the comment was in a branch that handled these window states: anyone touching the code will see that context. Now it is in a function without that context that makes no reference to those states. Not as part of the name, nor as a precondition.

Talking of the name, does `set_focus_replacement_for()` really reflect what the function provides?

tarek-y-ismail: > Now it is in a function without that context that makes no reference to those states. Not as part of the name, nor as a precondition.
> Talking of the name, does set_focus_replacement_for() really reflect what the function provides?

Then an assert and a comment should be enough (until we get contracts). As for the name, I suggest `select_replacement_for_minimized_or_hidden`. It's a bit too long for my linking but I don't have alternatives in mind.

AlanGriffiths: The naming and precondition is only an issue because you want to move the code out of its current context into a function. Maybe the right discussion is:

Does moving the code solve a problem?
What is that problem?
Are there other ways to solve it?

tarek-y-ismail: I see. The problem I had with this part of the code was that it was too annoying when I was working on and debugging #4664. The most obvious solution was to move it to a separate function. Other ways? Nothing comes to mind.

AlanGriffiths: I agree it isn't a thing of beauty, but what was annoying you about the code? How did moving it resolve that annoyance?

tarek-y-ismail: For memory, I was tracking down a bug that was not related to the window being minimized or hidden, so this big block was not of interest. It sounds silly in hindsight...

AlanGriffiths: Not silly, long functions are a problem to maintain. (I still remember debugging one of around 57kloc(!) after several decades. Although the length wasn't the biggest problem with that code...)

But there are trade-offs - we've now seen that extracting this cleanly isn't trivial. There other ways to extract stuff - like turning the function context into an object and splitting the cases into individual methods. But is it worth the time?

tarek-y-ismail: Okay. You convinced me. I'll drop it

AlanGriffiths:
Seems to have hung waiting on "documentation-checks / Run documentation checks" -I'm just gonna land it

PR #4677, opened 2026-02-12
feature: generating virtual C++ classes for each protocol in build.rs

mattkae:
## What's new?
For each Wayland protocol, we now generate a C++ class for each interface that contains:

1. A virtual destructor
2. The relevant forward declarations
3. All of the enums for that interface
4. Request handlers for all `request`s on the interface

This will put us in a position where we can hook up the C++ interfaces to the Rust dispatchers. But that is for another pull request, as it is a bit involved. This is a nice lay-up :basketball: 

We do NOT yet generate:
- The corresponding rust glue code (coming soon)
- The event code (coming soon)

## How to test
1. Build the package
2. Inspect the `include` directory to see the new headers

image

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/wayland/wayland_rs/build_script/cpp_builder.rs
+++ b/src/wayland/wayland_rs/build_script/cpp_builder.rs
@@ -216,7 +216,7 @@ pub enum CppType {
     CppF32,
     String,
     Object(String),
-    NewId(String),
+    NewId(Option<String>),
     Array,
     Fd,
 }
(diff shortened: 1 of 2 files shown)


src/wayland/wayland_rs/build_script/cpp_builder.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+fn cpp_type_to_string(cpp_type: &CppType) -> String {
+    match cpp_type {
+        CppType::CppI32 => "int32_t".to_string(),
+        CppType::CppU32 => "uint32_t".to_string(),
+        CppType::CppF32 => "float".to_string(),
+        CppType::String => "rust::Str const&".to_string(),
+        CppType::Object(name) => format!("rust::SharedPtr<{}> const&", name),
+        CppType::NewId(interface) => "int32_t".to_string(), // TODO
```
robert-ancell: TODO the interface name? Either implement or add clearer comment please.

src/wayland/wayland_rs/build_script/cpp_builder.rs
```diff
@@ -0,0 +1,246 @@
+pub struct CppBuilder {
+    guards: String,
```
robert-ancell: This name implied a plural - perhaps guard_name instead?

robert-ancell: Lower down the local variable that sets this is called "guard".

robert-ancell (approved):
Some small stuff but otherwise looks good, thanks!

PR #4681, opened 2026-02-13
Fix messages about CMP0167 on resolute

AlanGriffiths:
Use the "new" way to find Boost

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,9 @@ cmake_minimum_required(VERSION 3.28)
 
 cmake_policy(SET CMP0015 NEW)
 cmake_policy(SET CMP0022 NEW)
-cmake_policy(SET CMP0167 NEW)
+if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.30")
+    cmake_policy(SET CMP0167 NEW)
+endif()
 
 set(CMAKE_GCOV gcov)
 


AlanGriffiths:
> On second thought - maybe the new Cmake version is breaking CI?

Yes. I was too optimistic:

```
$ cmake --help-policy CMP0167
CMP0167
-------

.. versionadded:: 3.30

The ``FindBoost`` module is removed.

CMake 3.29 and below provide a ``FindBoost`` module, but it needs constant
updates to keep up with upstream Boost releases.  Upstream Boost 1.70
and above provide a ``BoostConfig.cmake`` package configuration file.
``find_package(Boost CONFIG)`` finds the upstream package directly,
without the find module.

CMake 3.30 and above prefer to not provide the ``FindBoost`` module
so that ``find_package(Boost)`` calls, without the ``CONFIG`` or
``NO_MODULE`` options, find the upstream ``BoostConfig.cmake`` directly.
This policy provides compatibility for projects that have not been ported
to use the upstream Boost package.

The ``OLD`` behavior of this policy is for ``find_package(Boost)`` to
load CMake's ``FindBoost`` module.  The ``NEW`` behavior is for
``find_package(Boost)`` to search for the upstream ``BoostConfig.cmake``.

This policy was introduced in CMake version 3.30.
It may be set by ``cmake_policy()`` or ``cmake_minimum_required()``.
If it is not set, CMake warns, and uses ``OLD`` behavior.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
```

PR #4685, opened 2026-02-16
Send the cursor image via ext_image_copy_capture_v1

### LINKED ISSUE #3659: Replace wlr-screencopy with ext-image-copy-capture

jhenstridge:
Closes #3659

## What's new?

This adds support for sending the cursor image via the `ext_image_copy_capture_v1` protocol. This makes the `ImageCopyBackend` used by the cursor session into a `CursorObserver` so it can receive `image_set_to` notifications.

If the cursor is hidden, we send a 1x1 transparent cursor. Sending 0x0 caused problems with wayvnc.

## How to test

Run miral with appropriate extensions enabled:

```
./bin/miral-app --add-wayland-extensions ext_output_image_capture_source_manager_v1:ext_image_copy_capture_manager_v1:zwlr_virtual_pointer_manager_v1:zwp_virtual_keyboard_manager_v1
```

Run wayvnc connecting to the this compositor:

```
WAYLAND_DISPLAY=wayland-1 wayvnc
```

Connect to the VNC session with a client that supports client cursor display like Remmina (GNOME Connections [does not work](https://gitlab.gnome.org/GNOME/gnome-connections/-/issues/140#note_2679520)).

Verify that the cursor changes as expected when resizing windows, or changes to a text insertion point in text entry fields. Try playing a video in Totem and note that the cursor is hidden when not in motion.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/ext_image_capture_v1.cpp
+++ b/src/server/frontend_wayland/ext_image_capture_v1.cpp
@@ -936,7 +936,7 @@ void mf::ExtImageCopyCaptureCursorSessionV1::ImageCopyBackend::begin_capture(
     if (cursor_image)
     {
         auto const dest_stride = mapping->stride();
-        auto const cursor_data = reinterpret_cast<char const*>(cursor_image->as_argb_8888());
+        auto const cursor_data = static_cast<char const*>(cursor_image->as_argb_8888());
         if (dest_stride == cursor_stride)
         {
             memcpy(mapping->data(), cursor_data, mapping->len());
@@ -944,7 +944,7 @@ void mf::ExtImageCopyCaptureCursorSessionV1::ImageCopyBackend::begin_capture(
         else
         {
             // strides don't match: copy data in rows
-            for (auto y = 0u; y < cursor_size.height.as_uint32_t(); ++y)
+            for (size_t y = 0u; y < cursor_size.height.as_uint32_t(); ++y)
             {
                 memcpy(mapping->data() + (dest_stride.as_uint32_t() * y),
                        cursor_data + (cursor_stride.as_uint32_t() * y),


src/server/frontend_wayland/ext_image_capture_v1.cpp
```diff
@@ -887,60 +899,61 @@ mf::ExtImageCopyCaptureCursorSessionV1::ImageCopyBackend::ImageCopyBackend(
...
-            else
-            {
-                data[pos] = std::byte{0x00};
-                data[pos+1] = std::byte{0x00};
-                data[pos+2] = std::byte{0x00};
-                data[pos+3] = std::byte{0x00};
-            }
-        }
+    using FailureReason = wayland::ImageCopyCaptureFrameV1::FailureReason;
+    // TODO: handle mismatched strides?
```
robert-ancell: Perhaps elaborate why we would get mismatched strides and why we should handle that? Or remove comment if no longer applicable.

robert-ancell (approved):
Small nit with a TODO.

src/server/frontend_wayland/ext_image_capture_v1.cpp
jhenstridge: The destination is a buffer allocated by the client. With the old wlr screencopy extension, the server communicated what stride the client should use in the [`buffer` event](https://wayland.app/protocols/wlr-screencopy-unstable-v1#zwlr_screencopy_frame_v1:event:buffer). With the image copy capture extension, the [`buffer_size` event](https://wayland.app/protocols/ext-image-copy-capture-v1#ext_image_copy_capture_session_v1:event:buffer_size) only indicates the height and width.

So a conformant client could attach a buffer with any stride it wants. Wayvnc happens to use a stride that matches Mir's `CursorImage`, so a simple memcpy worked here.

jhenstridge:
One other thing I noticed where the behaviour is correct, but may be better to do more explicitly:

The protocol says that the `hotspot` event on the `ext_image_copy_capture_cursor_session_v1` should be sent before the `ready` event on the `ext_image_copy_capture_frame_v1` for the cursor image. I'm triggering these from `image_set_to` calls on two different `CursorObserver`s. As they are both using the WaylandExecutor, the ordering should be deterministic based on the order they were registered, and this gives the correct event ordering.

It might be better not to rely on this, and send both events from the same CursorObserver.

src/server/frontend_wayland/ext_image_capture_v1.cpp
RAOF: It's not terribly difficult to handle stride (you just do a row-by-row copy), so maybe we should just implement this properly rather than leaving a TODO?

RAOF (changes requested):
I think we should just do the TODO.

Saviq:
@jhenstridge you'll want to look at the TiCS violations, please.

Conan-Kudo:
This shouldn't close #3659 since it doesn't implement toplevel support.

PR #4694, opened 2026-02-18
Use `SurfaceStateTracker` to hide/restore windows to the proper state

tarek-y-ismail:
Spawned off https://github.com/canonical/mir/pull/4664#discussion_r2815946423

## What's new?

- Instead of forcing `mir_window_state_restored` or `mir_window_state_hidden` when mapping/unmapping windows, the code now uses `SurfaceStateTracker` which manages state precedence. 

## How to test

- A separate PR for WLCS tests will be opened shortly.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/window_wl_surface_role.cpp
+++ b/src/server/frontend_wayland/window_wl_surface_role.cpp
@@ -350,13 +350,14 @@ void mf::WindowWlSurfaceRole::commit(WlSurfaceState const& state)
     {
         bool const is_mapped = scene_surface->visible();
         bool const should_be_mapped = static_cast<bool>(surface.value().buffer_size());
-        if (!is_mapped && should_be_mapped && scene_surface->state() == mir_window_state_hidden)
+        auto const state_tracker = scene_surface->state_tracker();
+        if (!is_mapped && should_be_mapped && state_tracker.has(mir_window_state_hidden))
         {
-            spec().state = mir_window_state_restored;
+            spec().state = state_tracker.without(mir_window_state_hidden).active_state();
         }
         else if (is_mapped && !should_be_mapped)
         {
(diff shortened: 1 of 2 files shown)


tarek-y-ismail:
Yeah I was diagnosing if CI was still borked with @Saviq. Thought I just forced pushed to drop the commit... Maybe I didn't. Let me  try again.

tarek-y-ismail:
Are we actually implementing the right solution?

From [`xdg_toplevel`](https://wayland.app/protocols/xdg-shell#xdg_toplevel):
> Unmapping an xdg_toplevel means that the surface cannot be shown by the compositor until it is explicitly mapped again. All active operations (e.g., move, resize) are canceled and all attributes (e.g. title, state, stacking, ...) are discarded for an xdg_toplevel surface when it is unmapped. The xdg_toplevel returns to the state it had right after xdg_surface.get_toplevel. The client can re-map the toplevel by performing a commit without any buffer attached, waiting for a configure event and handling it as usual (see xdg_surface description).

Looking at `XdgSurfaceStable::get_toplevel`, I can see we create an instance of `XdgToplevelStable`, which delegates most state to `WindowWlSurfaceRole`. The window state of that is stored in `WindowWlSurfaceRole::scene_surface()`. Tracking that down to `BasicSurface`, the default state is `mir_window_state_restored`

tarek-y-ismail:
I'm not sure if the code modified in this PR can be invoked directly by the compositor/window manager from what I've seen trying to test this outside of WLCS. The client has to submit a buffer for us to hit this path.

AlanGriffiths:
> I'm not sure if the code modified in this PR can be invoked directly by the compositor/window manager from what I've seen trying to test this outside of WLCS. The client has to submit a buffer for us to hit this path.

Ack. Then this is just wrong

PR #4696, opened 2026-02-18
If a surface has an opaque region BUT the alpha is not 1.0, then we shouldn't use it as an occlusion

mattkae:
Related:
https://github.com/miracle-wm-org/miracle-wm/issues/770

## What's new?
This has been breaking me in a funky way. In miracle, if I have Firefox on workspace 1 and some terminals on workspace 2, running the fade animation between the two screens was the fade to not happen on the terminals in workspace 2 because they believed themselves to be occluded by Firefox, even though it was fading away. The solution is to only use the occlusion zone if alpha is not 1.0, or else you can see through the surface.

## How to test
1. Build miracle with this version of Mir and this version of miracle: https://github.com/miracle-wm-org/miracle-wm/pull/785
2. Have fade animations configured in Miracle
3. Open up Firefox on workspace 1
4. Open up 5 terminals on workspace 2
5. Switch between the workspaces and note that the fade and transform animations work again

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/server/compositor/occlusion.cpp
+++ b/src/server/compositor/occlusion.cpp
@@ -54,7 +54,7 @@ bool renderable_is_occluded(
         }
     }
 
-    if (!occluded)
+    if (!occluded && renderable.alpha() == 1.0f)
     {
         if (renderable.shaped())
         {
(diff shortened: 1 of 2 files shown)


RAOF (approved):
Yup.

tarek-y-ismail:
CI is borked, bypassing checks.

PR #4697, opened 2026-02-18
fix: preserve errno when close() follows a failed unlink() (#4557)

### LINKED ISSUE #4557: `errno` clobbering when using `unlink` followed by `close`
In a couple of places in our code, we use `unlink`, and if that fails, we call `close`. Each of these calls can modify `errno`. Which we later use to  print an error and might cause the wrong error to be printed. 

https://github.com/canonical/mir/blob/a3798aab082e6cc483538ba274cdf10d9c79384f/src/core/anonymous_shm_file.cpp#L43-L67

https://github.com/canonical/mir/blob/a3798aab082e6cc483538ba274cdf10d9c79384f/tests/unit-tests/test_shm_backing.cpp#L48-L73

(The code is duplicated and might be worth a refactoring as well, but that's a sidequest)

One solution to this would to save `errno` for each call separately. So adding `unlink_errno` after `unlink` and `close_errno` after `close` and checking each separately. Another solution could be to wrap both to automatically do error checking and logging, and replacing all relevant invocations.

pierre-l:
Closes #4557                                                                                                                                                                                                                                                                                   
   
## What's new?                                                                                                                                                                                                                                                                                 

When `unlink()` fails and `close()` is called for cleanup, `close()` can overwrite `errno` with an unrelated error. The subsequent error message then reports the wrong reason for the failure.

This fix saves `errno` immediately after the failing call, before `close()` can clobber it.

## How to test

I don't think unit tests are warranted here given the simplicity of the fix, a review should be enough, but let me know if I'm being naive.

## Checklist

- [x] Tests ~~added~~ and pass (unit tests not included in the PR to keep things simple, see [that separate branch](https://github.com/pierre-l/mir/commit/e19160554fdfedaa2c465bbd2b53de65ddfde034))
- [x] ~~(optional) Added Screenshots or videos~~

### DIFF
--- a/src/core/anonymous_shm_file.cpp
+++ b/src/core/anonymous_shm_file.cpp
@@ -54,8 +54,11 @@ mir::Fd create_anonymous_file(off_t size)
             {
                 if (unlink(template_filename) < 0)
                 {
+                    // Save errno before close() can overwrite it with an unrelated error
+                    int const original_errno = errno;
                     close(raw_fd);
                     raw_fd = -1;
+                    errno = original_errno;
                 }
             }
         }
--- a/tests/unit-tests/test_shm_backing.cpp
+++ b/tests/unit-tests/test_shm_backing.cpp
@@ -59,8 +59,11 @@ auto make_shm_fd(size_t size) -> mir::Fd
             {
                 if (unlink(template_filename) < 0)
                 {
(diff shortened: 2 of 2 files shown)


tarek-y-ismail:
CI is borked, bypassing checks.

PR #4699, opened 2026-02-18
Generating Rust bindings for the corresponding C++ classes

mattkae:
## What's new?
- Rust bindings are now generated for the C++ classes in `src/ffi_cpp.rs`
- Added `CppBuilder::to_rust_bindings` to convert the C++ constructs to rust bindings

image

## How to test
1. Run `cargo build` from `src/wayland/wayland_rs`
2. See that `src/ffi_cpp.rs` is outputted and everything builds

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/wayland/wayland_rs/build_script/cpp_builder.rs
+++ b/src/wayland/wayland_rs/build_script/cpp_builder.rs
@@ -308,12 +308,6 @@ fn sanitize_identifier(name: &str) -> String {
         return "_empty".to_string();
     }
 
-    // First check if it starts with a digit
-    let first_char = name.chars().next().unwrap();
-    if first_char.is_ascii_digit() {
-        return format!("_{}", name);
-    }
-
     // Try to parse as a regular identifier
     // If it fails (e.g., it's a keyword), use raw identifier
     match syn::parse_str::<Ident>(name) {


src/wayland/wayland_rs/build_script/cpp_builder.rs
```diff
@@ -113,6 +120,49 @@ impl CppBuilder {
...
+                    let methods = class.methods.iter().map(|method| {
+                        let method_name = format_ident!("{}", sanitize_identifier(&method.name));
+                        let args = method.args.iter().map(|arg| {
+                            let arg_name = format_ident!("{}", sanitize_identifier(&arg.name));
+                            let arg_type = cpp_type_to_rust_type(&arg.cpp_type);
+                            quote! { #arg_name: #arg_type }
+                        });
+
+                        quote! {
+                            pub fn #method_name(self: &#class_name, #(#args),*);
```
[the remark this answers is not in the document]
mattkae: No. We need to qualify the class name so that the generated code can distinguish between different methods with the same name.

src/wayland/wayland_rs/build_script/cpp_builder.rs
```diff
@@ -238,6 +284,44 @@ fn cpp_type_to_string(cpp_type: &CppType) -> String {
...
+    if first_char.is_ascii_digit() {
+        return format!("_{}", name);
+    }
+
+    // Try to parse as a regular identifier
+    // If it fails (e.g., it's a keyword), use raw identifier
+    match syn::parse_str::<Ident>(name) {
+        Ok(_) => name.to_string(),
+        Err(_) => format!("r#{}", name),
+    }
```
RAOF: Why are these two separate cases? Can't you put all doesn't-parse-as-an-`Ident` identifiers to the `r#{}, name` case?

RAOF (approved):
LGTM, with one optional question/suggestion.

src/wayland/wayland_rs/build_script/cpp_builder.rs
mattkae: True! And in practice, the digit-based name will never happen! I was too accepting of copilot here

PR #4707, opened 2026-02-20
ci: only cache from `main` builds

Saviq:
## What's new?

We only _store_ ccache on `main` builds, otherwise they often fall off the storage allowed and we get cache misses across the board.

## How to test

PR CI _doesn't_ store cache for spread and tics jobs.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -69,13 +69,19 @@ jobs:
         echo "merge-base=$( git merge-base origin/main ${{ github.sha }} )" >> $GITHUB_OUTPUT
 
     - name: CCache
+      if: ${{ github.event_name == 'push' }}
       uses: actions/cache@v5
-      with:
+      with: &ccache_with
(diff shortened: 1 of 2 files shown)


.github/workflows/tics.yml
```diff
@@ -69,14 +69,20 @@ jobs:
         echo "merge-base=$( git merge-base origin/main ${{ github.sha }} )" >> $GITHUB_OUTPUT
 
     - name: CCache
+      if: ${{ github.event_name == 'push' }}
```
mattkae, line 72: Does this need to only be on `main` too?

Saviq: We only run on `push` for main here:

https://github.com/canonical/mir/blob/391bb7be0c71a386f1cb32baafa591b6ad23783e/.github/workflows/tics.yml#L3-L7

PR #4716, opened 2026-02-23
Add support for WLCS keyboard

tarek-y-ismail:
## What's new?

- Adds support for WLCS's keyboard interface
- Bumps WLCS support from 2 to 4 (3 was for an optional feature we didn't need)

## How to test

- Build mir from this branch: `cmake -B build && cmake --build build`
- Check out the code for https://github.com/canonical/wlcs/pull/407, or main if it was merged, and build it
- Run WLCS's new keyboard tests, assuming it was built in `build`: `./build/wlcs <path to mir>/build/lib/miral_wlcs_integration.so --gtest_filter=KeyboardTest*`

## Checklist

- [x] Tests added and pass: in WLCS

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
+++ b/src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
@@ -451,7 +451,7 @@ fn cpp_return_type_to_cpp_source(cpp_type: &CppType) -> String {
         CppType::CppF64 => "double".to_string(),
         CppType::String => "std::string".to_string(),
         CppType::Object(name) => {
-            format!("std::unique_ptr<{}>", name)
+            format!("std::shared_ptr<{}>", name)
         }
         CppType::Array => "std::vector<uint8_t>".to_string(),
         CppType::Fd => "int32_t".to_string(),
@@ -472,8 +472,14 @@ fn cpp_arg_type_to_cpp_source(cpp_type: &CppType, originates_from_rust: bool) ->
         (CppType::CppU32, _) => "uint32_t".into(),
         (CppType::CppF64, _) => "double".into(),
         (CppType::Fd, _) => "int32_t".into(),
-        (CppType::Object(name), _) => format!("std::unique_ptr<{}> const&", name),
-        (CppType::Box(name), _) => format!("rust::Box<{}>", name),
+        (CppType::Object(name), _) => format!("std::shared_ptr<{}> const&", name),
+        (CppType::Box(name), _) => {
+            if originates_from_rust {
+                format!("rust::Box<{}>", name)
+            } else {
+                format!("rust::Box<{}> const&", name)
+            }
+        }
         (CppType::String, true) => "rust::String".into(),
         (CppType::String, false) => "std::string const&".into(),
         (CppType::Array, true) => "rust::Vec<uint8_t>".into(),
@@ -491,7 +497,7 @@ fn cpp_return_type_to_rust_source(cpp_type: &CppType) -> TokenStream {
         CppType::String => quote! { &CxxString },
         CppType::Object(name) => {
             let type_name = format_ident!("{}", name);
-            quote! { UniquePtr<#type_name> }
+            quote! { SharedPtr<#type_name> }
         }
         CppType::Array => quote! { &CxxVector<u8> },
         CppType::Fd => quote! { i32 },
@@ -521,7 +527,7 @@ fn cpp_arg_type_to_rust_source(cpp_type: &CppType, originates_from_rust: bool) -
         }
         CppType::Object(name) => {
             let type_name = format_ident!("{}", name);
-            quote! { &UniquePtr<#type_name> }
+            quote! { &SharedPtr<#type_name> }
         }
         CppType::Array => {
             if originates_from_rust {
@@ -533,7 +539,11 @@ fn cpp_arg_type_to_rust_source(cpp_type: &CppType, originates_from_rust: bool) -
         CppType::Fd => quote! { i32 },
         CppType::Box(name) => {
             let type_name = format_ident!("{}", name);
-            quote! { Box<#type_name> }
+            if originates_from_rust {
+                quote! { Box<#type_name> }
+            } else {
+                quote! { &Box<#type_name> }
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tarek-y-ismail:
Looks like the latest version of WLCS with keyboard support is not packaged outside of Ubuntu.

tests/mir_test_framework/test_wlcs_display_server.cpp
```diff
@@ -1039,3 +1094,50 @@ std::shared_ptr<mir::test::Signal> miral::TestWlcsDisplayServer::expect_event_wi
...
 }
+
+WlcsKeyboard* miral::TestWlcsDisplayServer::create_keyboard()
+{
+    auto constexpr uid = "keyboard-uid";
+
+    class DeviceObserver : public mir::input::NullInputDeviceObserver
+    {
+    public:
+        DeviceObserver(std::shared_ptr<mir::test::Signal> const& done)
```
mattkae:
```suggestion
        explicit DeviceObserver(std::shared_ptr<mir::test::Signal> const& done)
```

tests/mir_test_framework/test_wlcs_display_server.cpp
```diff
@@ -1039,3 +1094,50 @@ std::shared_ptr<mir::test::Signal> miral::TestWlcsDisplayServer::expect_event_wi
...
+    mir_server->the_input_device_hub()->add_observer(observer);
+
+    auto fake_keyboard_dev = mtf::add_fake_input_device(
+        mi::InputDeviceInfo{"keyboard", uid, mi::DeviceCapability::keyboard});
+
+    keyboard_added->wait_for(a_long_time);
+    executor->spawn([observer=std::move(observer), the_input_device_hub=mir_server->the_input_device_hub()]
+                        { the_input_device_hub->remove_observer(observer); });
+
+    auto fake_keyboard = new FakeKeyboard;
```
mattkae:
```suggestion
    auto fake_keyboard = new FakeKeyboard();
```

The interesting part of NOT using the parentheses is that it won't be zero-initialized. This rarely matters, but when it does, it's so annoying that it's worth always doing for classes :)

See: https://stackoverflow.com/questions/620137/do-the-parentheses-after-the-type-name-make-a-difference-with-new

tarek-y-ismail:
@Saviq Good case to bypass rules?

Saviq:
> @Saviq Good case to bypass rules?

No, this'd break CI for us. We need macro guards on a version of WLCS that will supply this.

tests/mir_test_framework/test_wlcs_display_server.cpp
```diff
@@ -844,13 +830,14 @@ miral::TestWlcsDisplayServer::TestWlcsDisplayServer(int argc, char const** argv)
...
 {
-    WlcsDisplayServer::version = 2;
+    WlcsDisplayServer::version = 4;
     WlcsDisplayServer::start = &wlcs_server_start;
     WlcsDisplayServer::stop = &wlcs_server_stop;
     WlcsDisplayServer::create_client_socket = &wlcs_server_create_client_socket;
     WlcsDisplayServer::position_window_absolute = &wlcs_server_position_window_absolute;
     WlcsDisplayServer::create_pointer = &wlcs_server_create_pointer;
     WlcsDisplayServer::create_touch = &wlcs_server_create_touch;
+    setup_keyboard(this);
```
AlanGriffiths: The combination of `WlcsDisplayServer::version = 4` and conditionally setting `WlcsDisplayServer::create_keyboard` (via a function call) is not going to ensure a runtime match:

You can compile with wlcs 3 and, following a wlcs update, run with wlcs 4. Which then has no way to know that `WlcsDisplayServer::create_keyboard` hasn't been set.

It would be better to set `WlcsDisplayServer::version` consistently with the compile time version of wlcs. (And to move the conditional setting of `create_keyboard` here to be consistent with the other attributes.)

tarek-y-ismail: Hopefully should be better now!

tests/mir_test_framework/test_wlcs_display_server.cpp
```diff
@@ -837,20 +824,29 @@ WlcsTouch* wlcs_server_create_touch(WlcsDisplayServer* server)
...
+}
 }
 
 miral::TestWlcsDisplayServer::TestWlcsDisplayServer(int argc, char const** argv) :
     TestDisplayServer{argc, argv},
     resource_mapper{std::make_shared<ResourceMapper>()},
     event_listener{std::make_shared<InputEventListener>(*this)}
 {
-    WlcsDisplayServer::version = 2;
+    WlcsDisplayServer::version = WLCS_DISPLAY_SERVER_VERSION;
```
AlanGriffiths: We don't support, say, WLCS version 35. So don't promise it.

```suggestion
    WlcsDisplayServer::version = std::min(WLCS_DISPLAY_SERVER_VERSION, 4);
```

tests/mir_test_framework/test_wlcs_display_server_internals.h
```diff
@@ -0,0 +1,56 @@
...
+    emitter->emit_event(event.with_event_time(event_time));
+
+    if (!event_sent->wait_for(a_long_time))
+        mir::fatal_error("fake event failed to go through");
+}
+
+} // namespace
+
+void setup_keyboard(WlcsDisplayServer* server);
+
```
AlanGriffiths: Unused and unimplemented

tests/mir_test_framework/test_wlcs_display_server_keyboard.cpp
AlanGriffiths, line 1: [non blocking] Is there a need to split this out into its own translation unit? It doesn't seem reusable or independent

tarek-y-ismail: Because that would require multiple `#if`s. Those would probably get lost in the noise or confuse someone (I know they did that to me). So having them in a separate file makes things clearer.

AlanGriffiths: You can avoid the `#ifs` entirely: just have two files and compile the right one. :wink:

PR #4717, opened 2026-02-23
feature: generating the GlobalDispatch and Dispatch method bodies for wayland-rs

mattkae:
## What's new?
- Generating the `GlobalFactory` class in C++ and Rust. This is responsible for creating C++ global interfaces.
- Generating methods for protocols that create new Wayland objects in C++ and Rust. These will now return a `shared_ptr` with the proper interface type.
- `generate_request_handler_arms` now transforms the Rust argument into a type that is compatible with the C++ interface.
  - e.g. Fds become ints, Wayland objects become SharedPtr, enums become u32
- Added `WaylandArgType` enum so that arguments are strongly typed
- No longer generating CppBuilders for `wl_registry` and `wl_display`, as those are handled internally by the Wayland Rust crate
- Implemented proper type generation for both C++ and Rust (e.g. fixed => f64, NewId => SharePtr, and so on)
- Moved forward declarations in `CppBuilder` to be per-namespace
- Added return types to methods in `CppBuilder`

## How to test
1. Run:
  ```
  cd src/wayland/wayland_rs
  cargo build
  ```
2. See:
  a. `src/wayland/wayland_rs/include/global_factory.h`: contains methods for creating globals
  b. `src/wayland/wayland_rs/src/dispatch.rs`: has full implementations both for creating globals and delegating requests back to the C++ layer

image

image

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/.github/workflows/spread.yml
+++ b/.github/workflows/spread.yml
@@ -75,6 +75,8 @@ jobs:
     steps:
     - name: Set up LXD
       uses: canonical/setup-lxd@main
+      with:
+        channel: stable
 
     - name: Set up Spread
       run: |
--- a/.github/workflows/tics.yml
+++ b/.github/workflows/tics.yml
@@ -165,13 +165,6 @@ jobs:
         additionalFlags: ${{ contains(fromJSON('["pull_request", "merge_group"]'), github.event_name) && '-resultdir build' || '' }}
         tmpdir: ${{ runner.debug && env.TICS_TMPDIR || null }}
 
-    - if: ${{ always() && runner.debug }}
-      uses: actions/upload-artifact@v6
-      with:
-        name: TICS tmpdir
-        path: ${{ env.TICS_TMPDIR }}
-        if-no-files-found: error
-
     - if: ${{ contains(fromJSON('["push", "workflow_dispatch"]'), github.event_name) && cancelled() }}
       name: Detach TICS QServer to prevent "project already connected" issue
       env:


src/wayland/wayland_rs/build_script/cpp_builder.rs
```diff
@@ -100,9 +91,16 @@ impl CppBuilder {
                         .collect();
                     let args_str = args.join(", ");
 
+                    let retstring = if let Some(retval) = &method.retval {
+                        cpp_type_to_string(retval)
+                    } else {
+                        "void".to_string()
+                    };
```
RAOF: I think chaining option transformers often leads to clearer code?
```suggestion
                    let retstring = method.retval.as_ref()
                        .map(cpp_type_to_string)
                        .value_or("void".to_string());
```

src/wayland/wayland_rs/build_script/cpp_builder.rs
```diff
@@ -143,11 +141,19 @@ impl CppBuilder {
...
                     let args = method.args.iter().map(|arg| {
                         let arg_name = format_ident!("{}", sanitize_identifier(&arg.name));
-                        let arg_type = cpp_type_to_rust_type(&arg.cpp_type);
+                        let arg_type = cpp_type_to_rust_type(&arg.cpp_type, false);
                         quote! { #arg_name: #arg_type }
                     });
-                    quote! {
-                        pub fn #method_name(self: &#class_name, #(#args),*);
+
+                    if let Some(retval) = &method.retval {
```
RAOF, line 148: Maybe would also be clearer as a transformer? Less clear here.

src/wayland/wayland_rs/build_script/main.rs
```diff
@@ -45,11 +47,11 @@ fn main() {
...
     write_protocols_rs(&protocols);
 
-    // Next, generate the dispatch and global dispatch methods.
-    write_dispatch_rs(&protocols);
-
     // Next, generate a C++ abstract class for each interface.
     write_cpp_protocol_headers(&protocols);
+
+    // Next, generate the dispatch and global dispatch methods.
+    write_dispatch_rs(&protocols);
```
RAOF, line 51: Why this change? I doesn't seem necessary?

src/wayland/wayland_rs/build_script/main.rs
```diff
@@ -169,31 +171,137 @@ fn generate_global_dispatch_impl(
...
+fn transform_argument_for_cpp(arg: &WaylandArg) -> Option<TokenStream> {
+    let arg_name = format_ident!("{}", arg.name);
+
+    // Enums are provided as WlEnum, so we need to cast them to u32 so that
+    // they can be sent across the barrier.
+    if arg.enum_.is_some() {
+        return match arg.type_ {
+            // The "int" appears to be a faff on the `WlSurface.set_buffer_transform` call,
+            // since most enums are u32s, so much so that the wayland Rust crate assumes
+            // that this is always the case.
```
RAOF: Hm, could we propagate the "always u32" into the C++ code instead? Unsigned arithmetic is better defined than signed arithmetic.

src/wayland/wayland_rs/build_script/main.rs
```diff
@@ -421,27 +614,46 @@ fn wayland_enum_to_cpp_enum(enum_: &WaylandEnum) -> CppEnum {
...
-            }), // Note: WlRegistry allows for a bind without a defined interface
-            "array" => CppType::Array,
-            "fd" => CppType::Fd,
-            _ => panic!("Unknown type: {}", arg.type_),
+                // Note: WlRegistry allows for a bind without a defined interface.
+                // WlRegistry should never generate cpp classes to begin with however,
+                // so this should NOT be an issue in practice. If it is an issue, we want
+                // to know about it immediately, so let's panic for now.
+                panic!("NewId lacks an interface. This should not happen except for WlRegistry.");
+            }),
```
RAOF: This whole match arm is unreachable, right? You're iterating over `.filter(|arg| arg.type_ != WaylandArgType::NewId)`, so you're guaranteed not to have `NewId`s here, yes?

mattkae:
> Some comments, and a general question: Returning `cxx:SharedPtr<>` into the Rust side is going to make everything very hairy when we want to mutate stuff (on the Rust side, at least). Would `cxx::UniquePrt<>` work?

The reason that I went with SharedPtr is because I figured some protocols might want a WeakPtr to a Wayland object so that they can do interesting things on it. What do you think?

src/wayland/wayland_rs/build_script/main.rs
mattkae: Eh not necessarily. Some protocols, i.e. https://wayland.app/protocols/xx-input-method-v2#xx_input_method_v1:request:set_preedit_string expect to be able to send -1

src/wayland/wayland_rs/build_script/main.rs
mattkae: Oh duh!

PR #4724, opened 2026-02-26
Make TestWlcsDisplayServer's CursorObserverMultiplexer dispatch events

jhenstridge:
## What's new?

`TestWlcsDisplayServer` registeres a custom `CursorObserverMultiplexer` that records the cursor position and then sends calls to the regular `CursorObserverMultiplexer`.

Unfortunately other observers are registered with this custom multiplexer, and it never dispatches events to its own observers (except for `image_set_to`, since it didn't override that). This prevents code relying on the multiplexer from working under test (e.g. ext-image-copy-capture's cursor movement tracking).

I also noticed some minor issues with the initial state events in the multiplexer while tracking those down, so I've included them here.

## How to test

The `ExtImageCopyCaptureTest::cursor_session_sends_pointer` test in https://github.com/canonical/wlcs/pull/408 depends on pointer events sourced from the multiplexer.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/tests/mir_test_framework/test_wlcs_display_server.cpp
+++ b/tests/mir_test_framework/test_wlcs_display_server.cpp
@@ -864,13 +864,11 @@ miral::TestWlcsDisplayServer::TestWlcsDisplayServer(int argc, char const** argv)
                     return resource_mapper;
                 });
 
-            server.wrap_cursor_observer_multiplexer(
-                [this](auto existing_cursor_multiplexer)
+            server.add_init_callback([this, &server]()
                 {
                     cursor_observer = std::make_shared<RunnerCursorObserver>(this);
-                    existing_cursor_multiplexer->register_early_observer(
+                    server.the_cursor_observer_multiplexer()->register_early_observer(
                         cursor_observer, mir::immediate_executor);
-                    return existing_cursor_multiplexer;
                 });
 
             mir_server = &server;


tests/mir_test_framework/test_wlcs_display_server.cpp
```diff
@@ -871,8 +867,10 @@ miral::TestWlcsDisplayServer::TestWlcsDisplayServer(int argc, char const** argv)
             server.wrap_cursor_observer_multiplexer(
                 [this](auto existing_cursor_multiplexer)
                 {
-                    cursor_observer = std::make_shared<RunnerCursorObserver>(this, existing_cursor_multiplexer);
-                    return cursor_observer;
+                    cursor_observer = std::make_shared<RunnerCursorObserver>(this);
+                    existing_cursor_multiplexer->register_early_observer(
+                        cursor_observer, mir::immediate_executor);
+                    return existing_cursor_multiplexer;
                 });
```
RAOF: I think this would be more idiomatic as
```
server.add_init_callback([this, &server]()
{
    cursor_observer = std:;make_shared<RunnerCursorObserver>(this);
    server.the_cursor_observer_multiplexer->register_observer(cursor_observer, mir::immediate_executor);
}
```
instead of (not) wrapping existing multiplexer.

jhenstridge: Thanks. I'd first tried getting rid of the callback entirely, and hit the `verify_accessing_allowed` check. This does look simpler than using `wrap_cursor_observer_multiplexer`.

PR #4726, opened 2026-02-26
Add and update ABI summary to changelog

AlanGriffiths:
We now maintain the ABI summary during development

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -2,6 +2,17 @@ mir (2.26.0~dev) UNRELEASED; urgency=medium
 
   * Begin development for 2.26.0 release
 
+    - ABI summary:
+      - miral ABI unchanged at 7
+      - mircommon ABI bumped to 12
+      - mircore ABI unchanged at 2
+      - miroil ABI unchanged at 8
+      - mirplatform ABI unchanged at 34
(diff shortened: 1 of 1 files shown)


debian/changelog
```diff
@@ -2,6 +2,17 @@ mir (2.26.0~dev) UNRELEASED; urgency=medium
...
+    - ABI summary:
+      - miral ABI unchanged at 7
+      - mircommon ABI bumped to 12
+      - mircore ABI unchanged at 2
+      - miroil ABI unchanged at 8
+      - mirplatform ABI unchanged at 34
+      - mirserver ABI bumped to 67
+      - mirwayland ABI unchanged at 5
+      - mirplatformgraphics ABI unchanged at 23
+      - mirinputplatform ABI unchanged at 10
```
Saviq, line 14: How about this until release?

```suggestion
      - mircommon ABI bumped to 12
      - mirserver ABI bumped to 67
```

AlanGriffiths: I thought about that, but I think it is harder to maintain: there's a risk of missing or duplicate entries when they are added piecemeal. (And even spurious merge conflicts if two version bumps are added.)

PR #4732, opened 2026-02-27
Expose the window transformation in the mirAL API

mattkae:
## What's new?
- Piped the glm::mat4 transform through the mirAL API so that compositor authors can now set it as they wish

## How to test
Modify `FloatingWindowManagerPolicy::place_new_window` to set the transformation on a window, something like:

```c++
#include <glm/ext.hpp>

...

parameters.transformation() = glm::scale(glm::mat4(1.f), glm::vec3(0.5f, 0.5f, 0.5f));
```

This will scale on windows by 0.5. Run the app, and see that it works. Woohoo!

## Checklist

- [ ] Tests added and pass
- [X] Adequate documentation added

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -254,6 +254,17 @@ pkg_check_modules(GLM glm)
 if(NOT GLM_FOUND)
   find_package(glm REQUIRED)  # There's no glm.pc on Fedora, but find_package(glm) fails on Ubuntu...
 endif()
+# Ensure a consistent glm::glm interface target exists.
+# pkg_check_modules does not create an imported target, but find_package(glm) does.
+if(GLM_FOUND AND NOT TARGET glm::glm)
+  add_library(glm::glm INTERFACE IMPORTED)
+  set_target_properties(glm::glm PROPERTIES
+    INTERFACE_INCLUDE_DIRECTORIES "${GLM_INCLUDE_DIRS}")
+  # glm.pc exists on this platform, so downstream pkg-config consumers can find it
+  set(MIRAL_GLM_PC_REQUIRES "glm")
+else()
+  set(MIRAL_GLM_PC_REQUIRES "")
+endif()
 pkg_check_modules(DRM REQUIRED IMPORTED_TARGET libdrm)
 pkg_check_modules(EGL REQUIRED IMPORTED_TARGET egl)
 pkg_check_modules(EPOXY REQUIRED IMPORTED_TARGET epoxy)
--- a/src/miral/CMakeLists.txt
+++ b/src/miral/CMakeLists.txt
@@ -197,6 +197,7 @@ target_include_directories(miral
 target_link_libraries(miral-external
     PUBLIC
         mircore
+        glm::glm
     PRIVATE
         miral-internal
         mirserver
--- a/src/miral/miral.pc.in
+++ b/src/miral/miral.pc.in
@@ -5,6 +5,6 @@ includedir=@PKGCONFIG_INCLUDEDIR@/miral
 Name: miral
 Description: Mir Abstraction Layer library
 Version: @MIRAL_VERSION@
-Requires: mircore, xkbcommon
+Requires: mircore, xkbcommon @MIRAL_GLM_PC_REQUIRES@
 Libs: -L${libdir} -lmiral
 Cflags: -I${includedir}


src/miral/basic_window_manager.cpp
```diff
@@ -1235,6 +1235,12 @@ void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpe
 
     if (modifications.confine_pointer().is_set())
         std::shared_ptr<scene::Surface>(window)->set_confine_pointer_state(modifications.confine_pointer().value());
+
+    if (modifications.alpha().is_set())
```
robert-ancell, line 1239: I would recommend this in a separate PR - it will be clearer in the history this is fixing a bug and this should be a trivial PR to land quickly, where the transformation might require more discussion.

tarek-y-ismail: +1 for small PRs (from experience :))

src/miral/window_specification.cpp
```diff
@@ -57,7 +57,9 @@ miral::WindowSpecification::Self::Self(mir::shell::SurfaceSpecification const& s
     server_side_decorated(spec.server_side_decorated),
     focus_mode(spec.focus_mode),
     visible_on_lock_screen(spec.visible_on_lock_screen),
-    tiled_edges(spec.tiled_edges)
+    tiled_edges(spec.tiled_edges),
+    opacity(spec.alpha),
```
tarek-y-ismail, line 61: This is related to the alpha bug fix, right?

src/server/shell/surface_specification.cpp
```diff
@@ -68,7 +68,10 @@ bool msh::SurfaceSpecification::is_empty() const
         !application_id.is_set() &&
         !server_side_decorated.is_set() &&
         !focus_mode.is_set() &&
-        !visible_on_lock_screen.is_set();
+        !visible_on_lock_screen.is_set() &&
+        !tiled_edges.is_set() &&
+        !alpha.is_set() &&
```
tarek-y-ismail, line 73: more unrelated changes?

src/miral/basic_window_manager.cpp
mattkae: You have won this round.

Please review: https://github.com/canonical/mir/pull/4748

src/miral/window_specification.cpp
mattkae: Please review: https://github.com/canonical/mir/pull/4748

src/server/shell/surface_specification.cpp
mattkae: Please review: https://github.com/canonical/mir/pull/4748

include/miral/miral/window_info.h
```diff
@@ -385,6 +387,14 @@ struct WindowInfo
     /// \remark Since MirAL 5.7
     auto alpha() const -> float;
 
+    /// The transformation matrix of the window.
+    ///
+    /// By default, the identity matrix is used.
+    ///
+    /// \returns the transformation matrix
+    /// \remark Since MirAL 5.7
```
RAOF, line 398: For this to be useful, we need more documentation about where this transformation is applied - does it include the window position, what is it relative to, etc?

mattkae:
> If we want to expose this, we need to specify where in the pipeline this transform is being applied - how does it interact with window position, buffer scale, etc?
> 
> I'm not sure we _do_ want to expose this, though. It by necessity stabilises details of our rendering pipeline in the public API and it won't be possible to sensibly support a software renderer with this.
> 
> Also, it's not _actually_ what Matt wants, it's just the thing that happens to already exist (although unused) internally.
> 
> That said, it _can_ be useful now, so 🤷‍♀️

I think it is generally useful now to apply this transformation as if it were the "model" transformation to a single window. I would happily use at least that in Miracle.

Of course I personally have another layer which would require a scene graph, but I don't want to wait for that :smile: 

I will update the comment to make the effect of the transform more apparent.

RAOF:
So, this doesn't interact with input *at all*, right? ie: if you do the scale-by-0.5 example, you'll have a ½-size window which receives input over where the window *would* be if it were not scaled, right?

Would it be acceptable for your use-case if setting a non-identity transform here *disabled* input to the surface? That would be something we could easily make forward-compatible if-and-when we do this *properly* (with integration with input dispatch).

mattkae:
> So, this doesn't interact with input _at all_, right? ie: if you do the scale-by-0.5 example, you'll have a ½-size window which receives input over where the window _would_ be if it were not scaled, right?
> 
> Would it be acceptable for your use-case if setting a non-identity transform here _disabled_ input to the surface? That would be something we could easily make forward-compatible if-and-when we do this _properly_ (with integration with input dispatch).

Oof I forgot that I worked around this.

I don't think it would be acceptable and would probably be fairly confusing to most people if we offered it in the API but didn't propagate it throughout.

I will make a ticket on me to handle the rotated input logic that is a precursor to this one :)

AlanGriffiths:
> I will make a ticket on me to handle the rotated input logic that is a precursor to this one :)

Should we more this to draft?

Saviq:
We also discussed another way for much of this to be possible - that is windows "as subsurfaces". So the (privileged) client can draw whatever it wants, and create a subsurface-like thing that it will attach a "foreign buffer" to. Add transformations, and it should allow for a lot of freedom to shell components (switchers, expose effect etc., including animations from whatever toolkit you choose to use).

AlanGriffiths:
> We also discussed another way for much of this to be possible

Much of what? This allows the compositor to attach a transform to the window. You talk about a client submitting buffers with associated transforms. When are these solving the same problem?

Saviq:
> Much of what?

The prospective bling that is inspiration for this PR. If we feel uncomfortable with these changes, let's not convince ourselves of them. But maybe that's not mutually exclusive - one serves the compositor, another - shell components.

mattkae:
I've decided to close this one for now. After speaking with Chris and Saviq, we have a deeper (maybe epic?) that we should work on in lieu of this.

PR #4740, opened 2026-03-03
Refactor the event processing logic in evdev-rs to increase clarity

mattkae:
## What's new?
Same functionality, but now more readable!

## How to test
1. Run mir with evdev-rs
2. It will work the same as always :tada: 

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/platforms/evdev-rs/src/event_processing.rs
+++ b/src/platforms/evdev-rs/src/event_processing.rs
@@ -55,7 +55,7 @@ impl Default for ScrollAxisData {
             precise: 0.0,
             discrete: 0,
             value120: 0,
-            stop: true,
+            stop: false,
         }
     }
 }
@@ -614,36 +614,41 @@ fn handle_touch_frame(
     device_info: &mut LibinputDeviceInfo,
     frame_event: input::event::touch::TouchFrameEvent,
 ) -> bool {
-    let mut contacts: Vec<crate::TouchContactData> = vec![];
     let mut empty_touches = 0;
 
-    for (slot, contact_data) in &mut device_info.touch_properties {
-        // Note: This was logic taken from the existing evdev implementation, and we are
-        // keeping it for backwards compatibility.
-        // Sanity check: Bogus panels are sending sometimes empty events that all point
-        // to (0, 0) coordinates. Detect those and drop the whole frame in this case.
-        // Count touches at (0, 0) but still include them in contacts for now.
-        if contact_data.x == 0.0 && contact_data.y == 0.0 {
-            empty_touches += 1;
-        }
+    let contacts: Vec<crate::TouchContactData> = device_info
+        .touch_properties
+        .iter_mut()
+        .map(|(slot, contact_data)| {
+            // Note: This was logic taken from the existing evdev implementation, and we are
+            // keeping it for backwards compatibility.
+            // Sanity check: Bogus panels are sending sometimes empty events that all point
+            // to (0, 0) coordinates. Detect those and drop the whole frame in this case.
+            // Count touches at (0, 0) but still include them in contacts for now.
+            if contact_data.x == 0.0 && contact_data.y == 0.0 {
+                empty_touches += 1;
+            }
 
-        contacts.push(crate::TouchContactData {
-            touch_id: *slot as i32,
-            action: contact_data.action.repr,
-            tooltype: contact_data.tooltype.repr,
-            position_x: contact_data.x,
-            position_y: contact_data.y,
-            pressure: contact_data.pressure,
-            touch_major: contact_data.major,
-            touch_minor: contact_data.minor,
-            orientation: contact_data.orientation,
-        });
+            let contact = crate::TouchContactData {
+                touch_id: *slot as i32,
+                action: contact_data.action.repr,
+                tooltype: contact_data.tooltype.repr,
+                position_x: contact_data.x,
+                position_y: contact_data.y,
+                pressure: contact_data.pressure,
+                touch_major: contact_data.major,
+                touch_minor: contact_data.minor,
+                orientation: contact_data.orientation,
+            };
 
-        if contact_data.action == MirTouchAction::mir_touch_action_down {
-            contact_data.action = MirTouchAction::mir_touch_action_change;
-            contact_data.down_notified = true;
-        }
-    }
+            if contact_data.action == MirTouchAction::mir_touch_action_down {
+                contact_data.action = MirTouchAction::mir_touch_action_change;
+                contact_data.down_notified = true;
+            }
+
+            contact
+        })
+        .collect();
 
     // Remove any property that is now "up" so that we do not keep sending
     // the up notification.


src/platforms/evdev-rs/src/event_processing.rs
```diff
@@ -31,25 +31,54 @@ use input::AsRaw;
...
 
+// Linux input button codes.
+const BTN_LEFT: u32 = 0x110;
+const BTN_RIGHT: u32 = 0x111;
+const BTN_MIDDLE: u32 = 0x112;
+const BTN_SIDE: u32 = 0x113;
+const BTN_EXTRA: u32 = 0x114;
+const BTN_FORWARD: u32 = 0x115;
+const BTN_BACK: u32 = 0x116;
+const BTN_TASK: u32 = 0x117;
```
tarek-y-ismail, line 42: Is it a good idea to have these duplicated from the headers? I know they're unlikely to change, but it would be nice to have a single source of truth.

src/platforms/evdev-rs/src/event_processing.rs
```diff
@@ -66,18 +95,586 @@ fn get_bounding_rectangle(
...
+        buttons: device_info.button_state,
+        axis_source: data.axis_source,
+        precise_x: x.precise,
+        discrete_x: x.discrete,
+        value120_x: x.value120,
+        scroll_stop_x: x.stop,
+        precise_y: y.precise,
+        discrete_y: y.discrete,
+        value120_y: y.value120,
+        scroll_stop_y: y.stop,
```
tarek-y-ismail, line 306: Maybe make these fields use `ScrollAxisData`? I just tried it locally. Had to move the struct and its impl to `ffi.rs`, derive `Copy, Clone`. You can then use `x` and `y` directly.

As a bonus, `PointerEventData`'s default impl shrinks as well: 
```diff
 impl Default for PointerEventData {
@@ -66,14 +82,8 @@ impl Default for PointerEventData {
             displacement_x: 0.0,
             displacement_y: 0.0,
             axis_source: crate::MirPointerAxisSource::mir_pointer_axis_source_none.repr,
-            precise_x: 0.0,
-            discrete_x: 0,
-            value120_x: 0,
-            scroll_stop_x: false,
-            precise_y: 0.0,
-            discrete_y: 0,
-            value120_y: 0,
-            scroll_stop_y: false,
+            scroll_data_x: ScrollAxisData::default(),
+            scroll_data_y: ScrollAxisData::default(),
         }
     }
 }
```

Though I just noticed that the old default for `scroll_stop_{x,y}` is `false`, while it's `true` in `ScrollAxisData`. Intentional or bug?

```rust
impl Default for ScrollAxisData {
    fn default() -> Self {
        Self {
            precise: 0.0,
            discrete: 0,
            value120: 0,
            stop: true,
        }
    }
}
```

src/platforms/evdev-rs/src/event_processing.rs
```diff
@@ -66,18 +95,586 @@ fn get_bounding_rectangle(
...
+    pointer_event: event::PointerEvent,
+) -> bool {
+    match pointer_event {
+        event::PointerEvent::Motion(motion_event) => {
+            handle_pointer_motion(device_info, motion_event);
+        }
+
+        event::PointerEvent::MotionAbsolute(absolute_motion_event) => {
+            if !handle_pointer_motion_absolute(device_info, bridge, absolute_motion_event) {
+                return false;
```
tarek-y-ismail, line 393:
```suggestion
            return !handle_pointer_motion_absolute(device_info, bridge, absolute_motion_event);
```

src/platforms/evdev-rs/src/event_processing.rs
```diff
@@ -66,18 +95,586 @@ fn get_bounding_rectangle(
...
+/// (used when the bounding rectangle is unavailable or a frame should be dropped).
+fn handle_touch_event(
+    device_info: &mut LibinputDeviceInfo,
+    bridge: &cxx::SharedPtr<crate::PlatformBridge>,
+    touch_event: event::TouchEvent,
+) -> bool {
+    match touch_event {
+        event::TouchEvent::Down(down_event) => {
+            let Some(slot) = down_event.slot() else {
+                return true;
```
tarek-y-ismail, line 541: One thing that just came to mind is that we could use an enum here instead of returning booleans. Would make it a lot easier to reason about without having to translate what `true` and `false` map to.

src/platforms/evdev-rs/src/event_processing.rs
```diff
@@ -66,18 +95,586 @@ fn get_bounding_rectangle(
...
+            touch_major: contact_data.major,
+            touch_minor: contact_data.minor,
+            orientation: contact_data.orientation,
+        });
+
+        if contact_data.action == MirTouchAction::mir_touch_action_down {
+            contact_data.action = MirTouchAction::mir_touch_action_change;
+            contact_data.down_notified = true;
+        }
+    }
```
tarek-y-ismail: Maybe this would be a bit more idiomatic?


```suggestion
    let mut empty_touches = 0;

    let contacts  = device_info.touch_properties.iter_mut()
        .map(|(slot, contact_data)| {
            // Note: This was logic taken from the existing evdev implementation, and we are
            // keeping it for backwards compatibility.
            // Sanity check: Bogus panels are sending sometimes empty events that all point
            // to (0, 0) coordinates. Detect those and drop the whole frame in this case.
            // Count touches at (0, 0) but still include them in contacts for now.
            if contact_data.x == 0.0 && contact_data.y == 0.0 {
                empty_touches += 1;
            }

            let data = crate::TouchContactData {
                touch_id: *slot as i32,
                action: contact_data.action.repr,
                tooltype: contact_data.tooltype.repr,
                position_x: contact_data.x,
                position_y: contact_data.y,
                pressure: contact_data.pressure,
                touch_major: contact_data.major,
                touch_minor: contact_data.minor,
                orientation: contact_data.orientation,
            };

            if contact_data.action == MirTouchAction::mir_touch_action_down {
                contact_data.action = MirTouchAction::mir_touch_action_change;
                contact_data.down_notified = true;
            }

            data
        }).collect::<Vec<crate::TouchContactData>>();

```

src/platforms/evdev-rs/src/event_processing.rs
mattkae: There's a crate for this, but I wanted to limit a dependency for now.

src/platforms/evdev-rs/src/event_processing.rs
mattkae: It doesn't matter either way, but we can leave it `false` :)

mattkae: I don't want to move `ScrollAxisData` to ffi.rs personally, as that is reserved for structs that are meant to cross the ffi barrier.

src/platforms/evdev-rs/src/event_processing.rs
mattkae: I'll leave it as-is for now, as it matches all of the other ones. That might be a later refactor :+1: good catch tho!

src/platforms/evdev-rs/src/event_processing.rs
mattkae: I think the docstring says it well enough, so I will leave as a bool unless we introduce a 3rd state :)

src/platforms/evdev-rs/src/event_processing.rs
mattkae: Very correct :100:

PR #4410, opened 2025-11-13
Input triggers

tarek-y-ismail:
## What's new?

Initial implementation of the input triggers protocols. Allows privileged clients to reserve a key combination (modifiers + symkey/keycode) and be notified whenever that combination is pressed/released, regardless of which client has focus.

## How to test

Direct communication with privileged client:
- Run mir with `--add-wayland-extensions=ext_input_trigger_action_manager_v1:ext_input_trigger_registration_manager_v1`
- Inside the running compositor, launch a terminal
- Run `./build/bin/trigger_client`
- At the moment, it's hardcoded to request `Ctrl + Shift + C` and `ALT + x` for key syms, and `Alt + z` for scan codes. If you press any of them. "Hey from <combination>" will be printed
- Letting go will result in "Bye from <combination>" being printed.

To test `Alt + z`, run mir with `--keymap fr`. This should use the azerty layout and make the `z` key print `w`, and the `q` key print `z`. Even in the azerty layout, `Alt + z` with the azerty z will not work, but the same combination with the qwerty z will. Note: Right alt is used for special letters, so use left alt :)

--- 

Gatekeeper:
- Run mir with `--add-wayland-extensions=ext_input_trigger_action_manager_v1:ext_input_trigger_registration_manager_v1`
- Inside the running compositor, launch a terminal
- Run `python3 examples/gatekeeper/gatekeeper.py --show-ui`
- Run `python3 examples/gatekeeper/test_client.py`
- The gatekeeper should request to register two shortcuts. You're free to reject, modify them, or accept them as is.
- Pressing the shortcuts with the test client not focused will print different messages when `begin` and `end` are received.

## Checklist

- [x] Tests added and pass
  - WLCS tests require the addition of keyboard access. Probably for another PR.
- [x] Final rebase: once the PR is approved

PR #4746, opened 2026-03-05
Add keyboard state tracker

tarek-y-ismail:
Related #4410 

## What's new?

- Adds `KeyboardStateTracker` to provide global keyboard state tracking. This is meant as a utility class to be used with input triggers to provide a consistent global view of keyboard states:

https://github.com/canonical/mir/blob/594cfcf01add05d4f46c1d9a1dffbf3fab4d0440/src/server/frontend_wayland/wl_seat.cpp#L163

https://github.com/canonical/mir/blob/84f77ea995ad6ce5760359b371b99db779dcde13/src/server/frontend_wayland/input_trigger_registration_v1.cpp#L192

https://github.com/canonical/mir/blob/84f77ea995ad6ce5760359b371b99db779dcde13/src/server/frontend_wayland/input_trigger_registration_v1.cpp#L217

## How to test

- Run mir unit tests with `--gtest_filter=KeyboardStateTracker*`

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/keyboard_state_tracker.cpp
+++ b/src/server/frontend_wayland/keyboard_state_tracker.cpp
@@ -39,7 +39,10 @@ uint32_t constexpr to_xkb_scan_code(uint32_t evdev_scan_code)
 void mf::KeyboardStateTracker::XkbKeyState::update_keymap(
     std::shared_ptr<mir::input::Keymap> const& new_keymap, xkb_context* context)
 {
-    if (!new_keymap || (current_keymap && current_keymap->matches(*new_keymap)))
+    if (!new_keymap)
+        mir::fatal_error("KeyboardStateTracker: received null keymap");
+
+    if (current_keymap && current_keymap->matches(*new_keymap))
         return;
 
     current_keymap = new_keymap;
@@ -49,15 +52,12 @@ void mf::KeyboardStateTracker::XkbKeyState::update_keymap(
 
 void mf::KeyboardStateTracker::XkbKeyState::update_key(uint32_t xkb_keycode, MirKeyboardAction action)
 {
-    if (state)
-        xkb_state_update_key(state.get(), xkb_keycode, action == mir_keyboard_action_down ? XKB_KEY_DOWN : XKB_KEY_UP);
+    xkb_state_update_key(state.get(), xkb_keycode, action == mir_keyboard_action_down ? XKB_KEY_DOWN : XKB_KEY_UP);
 }
 
 void mf::KeyboardStateTracker::XkbKeyState::rederive_keysyms_from_scancodes(
     std::unordered_map<uint32_t, xkb_keysym_t>& scancode_to_keysym) const
 {
-    if (!state)
-        return;
 
     for (auto& [sc, ks] : scancode_to_keysym)
     {
--- a/src/server/frontend_wayland/keyboard_state_tracker.h
+++ b/src/server/frontend_wayland/keyboard_state_tracker.h
@@ -66,7 +66,7 @@ private:
     struct XkbKeyState
     {
         /// Update the compiled keymap and XKB state when a new keymap arrives.
-        /// \param new_keymap  The keymap carried on the incoming event.
+        /// \param new_keymap  The keymap carried on the incoming event. Must not be null.
         /// \param context     The shared XKB context owned by the tracker.
         void update_keymap(std::shared_ptr<mir::input::Keymap> const& new_keymap, xkb_context* context);
 


include/platform/mir/input/event_builder.h
```diff
@@ -47,6 +47,13 @@ class EventBuilder
         xkb_keysym_t keysym,
         int scan_code) = 0;
 
+    virtual EventUPtr key_event(
+        std::optional<Timestamp> timestamp,
+        MirKeyboardAction action,
+        xkb_keysym_t keysym,
+        int scan_code,
+        MirInputEventModifiers modifiers) = 0;
```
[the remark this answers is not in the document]
tarek-y-ismail: @AlanGriffiths Not sure if copilot is hallucinating? Is this even exposed? I can't find any exported symbols for `EventBuilder`

src/server/frontend_wayland/keyboard_state_tracker.cpp
```diff
@@ -0,0 +1,87 @@
...
+    if (action == mir_keyboard_action_down)
+    {
+        pressed_keysyms.insert(keysym);
+        pressed_scancodes.insert(scancode);
+        processed = true;
+    }
+    else if (action == mir_keyboard_action_up)
+    {
+        pressed_keysyms.erase(keysym);
+        pressed_scancodes.erase(scancode);
```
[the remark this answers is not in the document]
tarek-y-ismail: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/canonical/mir/pull/4746#discussion_r2890212425)

include/platform/mir/input/event_builder.h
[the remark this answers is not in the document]
AlanGriffiths: Why is this even part of the PR? It wasn't part of #4410 from which this was split.

But `include/platform` is the API for libmirplatform. And, yes, inserting functions into the vtab is an ABI break. As `EventBuilder` is an interface _used_ by platforms there's no reason to expect symbols for it.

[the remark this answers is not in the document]
tarek-y-ismail: This was added because I needed a way to create keyboard events with modifiers attached since Mir's input stack passes shift as a modifier, not a keysym AFAIS. `EventBuilder` did not provide a way to do that.

I can probably use `mir::events::make_key_event` instead since this is just a test.

tests/unit-tests/frontend_wayland/test_keyboard_state_tracker.cpp
```diff
@@ -0,0 +1,241 @@
...
+}
+
+TEST_F(KeyboardStateTrackerTest, shift_demotion_does_not_affect_non_alpha_keysyms)
+{
+    tracker.process(*key_down(XKB_KEY_Shift_L, shift_l_scancode));
+    tracker.process(*key_down(XKB_KEY_Control_L, ctrl_l_scancode));
+    tracker.process(*key_up(XKB_KEY_Shift_L, shift_l_scancode));
+
+    EXPECT_TRUE(tracker.keysym_is_pressed(XKB_KEY_Control_L));
+}
```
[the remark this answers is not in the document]
tarek-y-ismail: A test related to this was added in #4747

src/server/frontend_wayland/keyboard_state_tracker.h
```diff
@@ -0,0 +1,70 @@
...
+/// down. This decouples state accumulation from the individual objects that
+/// need to inspect it, so a single tracker instance can be shared across many
+/// objects without each one having to maintain its own pressed-key set.
+///
+/// Shift-key transitions are handled specially: when a Shift key is pressed
+/// all lowercase keysyms already in the pressed set are promoted to their
+/// uppercase equivalents, and when all Shift keys are released they are
+/// demoted back. This keeps the stored keysyms consistent with what the
+/// keyboard layer reports as the logical key for subsequent events.
+class KeyboardStateTracker
```
mattkae, line 50: Should this be an abstract class that is available on the server configuration, like the `KeyMapper` and others? It seems like a good candidate!

src/server/frontend_wayland/keyboard_state_tracker.cpp
```diff
@@ -0,0 +1,95 @@
...
+        processed = true;
+    }
+
+    // Transitioned from no shift to at least one shift
+    if (prev_shift_state == 0 && shift_state != 0)
+    {
+        auto const uppercase =
+            std::ranges::views::transform(pressed_keysyms, [](auto key) { return xkb_keysym_to_upper(key); });
```
mattkae: I may be wrong, but I do not believe that `xkb_keysym_to_upper` will work in instances where you have held `Shift + 0` for example. You won't get `)`. I am not sure if this matters for your case, but it is the reality AFAIK

src/server/frontend_wayland/keyboard_state_tracker.h
tarek-y-ismail: Why? It's only used internally for input triggers at the moment. I don't understand what we gain (extensibility or otherwise) by converting it to an abstract class at the moment.

AlanGriffiths: > Why? It's only used internally for input triggers at the moment. I don't understand what we gain (extensibility or otherwise) by converting it to an abstract class at the moment.

This confusion is caused by the PR author providing no context for the functionality being added: either update the top comment to tell the reviewer why this is needed; or, add some code that uses it!

mattkae: Yeah I was misunderstanding where it was used personally :)

Where will this be used? That would be helpful to know if it is smart to make it abstract or not.

tarek-y-ismail: It'll be used in a few places in #4410

https://github.com/canonical/mir/blob/594cfcf01add05d4f46c1d9a1dffbf3fab4d0440/src/server/frontend_wayland/wl_seat.cpp#L163

https://github.com/canonical/mir/blob/84f77ea995ad6ce5760359b371b99db779dcde13/src/server/frontend_wayland/input_trigger_registration_v1.cpp#L192

https://github.com/canonical/mir/blob/84f77ea995ad6ce5760359b371b99db779dcde13/src/server/frontend_wayland/input_trigger_registration_v1.cpp#L217

As mentioned in the PR description, it's mostly a utility class. Keeps track of the global key state and allows querying. I split it off based on Alan's suggestion, which was a good idea since we caught a couple of bugs along the way!

mattkae: And my other thought is: who constructs it, and should it be constructed once globally in the configuration?

tarek-y-ismail: In #4410, it's constructed right before the Wayland connector context is created. It's passed down to the seat, and then to the context where its passed down to the registration manager.

At the moment, I want to keep it scoped as tight as possible, which is why it's limited to the Wayland connector's context.

src/server/frontend_wayland/keyboard_state_tracker.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <xkbcommon/xkbcommon.h>
+
+#include <algorithm>
+
+namespace mf = mir::frontend;
+
+namespace
+{
+// xkb scancodes are offset by 8 from evdev scancodes for compatibility with
+// the X protocol. This matches the same helper used in xkb_mapper.cpp.
```
mattkae, line 32: Nit: Should we refactor it out then?

src/server/frontend_wayland/keyboard_state_tracker.cpp
```diff
@@ -0,0 +1,144 @@
...
+uint32_t constexpr to_xkb_scan_code(uint32_t evdev_scan_code)
+{
+    return evdev_scan_code + 8;
+}
+} // namespace
+
+void mf::KeyboardStateTracker::XkbKeyState::update_keymap(
+    std::shared_ptr<mir::input::Keymap> const& new_keymap, xkb_context* context)
+{
+    if (!new_keymap || (current_keymap && current_keymap->matches(*new_keymap)))
```
mattkae: Nit: Should the fact that `new_keymap` is not null be a precondition? It seems odd that we would ever call it with that since its essentially a no-op

src/server/frontend_wayland/keyboard_state_tracker.cpp
```diff
@@ -0,0 +1,144 @@
...
+        return;
+
+    current_keymap = new_keymap;
+    compiled_keymap = new_keymap->make_unique_xkb_keymap(context);
+    state = {xkb_state_new(compiled_keymap.get()), xkb_state_unref};
+}
+
+void mf::KeyboardStateTracker::XkbKeyState::update_key(uint32_t xkb_keycode, MirKeyboardAction action)
+{
+    if (state)
```
mattkae: Should we warn if state is unset? This goes for every call. We'd probably want to know about it

AlanGriffiths:
> My main thought however is: This looks like it should live at a higher level (maybe near the input transformers, or just after where the modifiers get processed). This doesn't feel like a "WlSeat" utility: this feels like a general utility 😄

There is something with a similar scope in input: `XkbMappingState::state`. But it is not clear we should try to generalize (yet).

src/server/frontend_wayland/keyboard_state_tracker.cpp
tarek-y-ismail: Nah. As @AlanGriffiths mentioned on [Matrix](https://matrix.to/#/!ivAdrrLoWkBYEnSIae:matrix.org/$SsD28fgnxYnAf8_iGubooTrneOCdeEXv6RzEfl-6cmE?via=matrix.org&via=ubuntu.com&via=fedora.im), it's unlikely to change. 

I got some minor cleanups out of that though (#4761), and got https://github.com/canonical/mir/issues/4762#issue-4057287856 out of it.

src/server/frontend_wayland/keyboard_state_tracker.cpp
```diff
@@ -49,15 +50,24 @@ void mf::KeyboardStateTracker::XkbKeyState::update_keymap(
 
 void mf::KeyboardStateTracker::XkbKeyState::update_key(uint32_t xkb_keycode, MirKeyboardAction action)
 {
-    if (state)
-        xkb_state_update_key(state.get(), xkb_keycode, action == mir_keyboard_action_down ? XKB_KEY_DOWN : XKB_KEY_UP);
+    if (!state)
+    {
+        mir::log_error("XkbKeyState::update_key: cannot update key without valid XKB state");
```
AlanGriffiths: I don't think this rates `log_error` - the server isn't likely to fail as a result.

Log messages (especially warning and above) need to be worded so that anyone looking at the log will have a clue what to do about it. (think "help message")

Is it an error in the program logic that needs fixing for `KeyboardStateTracker` to work? That would be  a `fatal_error()` (and the message would be fine).

src/server/frontend_wayland/keyboard_state_tracker.cpp
```diff
@@ -49,15 +50,24 @@ void mf::KeyboardStateTracker::XkbKeyState::update_keymap(
...
+    xkb_state_update_key(state.get(), xkb_keycode, action == mir_keyboard_action_down ? XKB_KEY_DOWN : XKB_KEY_UP);
 }
 
 void mf::KeyboardStateTracker::XkbKeyState::rederive_keysyms_from_scancodes(
     std::unordered_map<uint32_t, xkb_keysym_t>& scancode_to_keysym) const
 {
     if (!state)
+    {
+        mir::log_error(
+            "XkbKeyState::rederive_keysyms_from_scancodes: cannot re-derive keysyms without a valid XKB state");
```
AlanGriffiths: Again "error" means the server is likely going down and we want to leave an explanation. This is not that.

PR #4736, opened 2026-03-02
task-4660 std::format instead of snprintf(buffer...) in 9 files

yotam-medini:
It may close #4660.

https://github.com/canonical/mir/issues/4660

## What's new?

This commit replaces all calls of snprintf(buffer...) with automatic buffer with std::format.

Use of vsnprintf(...) is not changed. Moving from vsnprinf to std::format requires design decisions.

## How to test

Under build directory, I ran:
ctest --exclude-regex wlcs --output-on-failure --stop-on-failure
And it ran successfully. Last message lines:

```
19/19 Test #19: mir_unit_tests_virtual ..................   Passed    0.38 sec
100% tests passed, 0 tests failed out of 19
Total Test time (real) = 234.74 sec
```

## Checklist

- [ ] Tests added and pass
- [ ] Adequate documentation added
- [ ] (optional) Added Screenshots or videos

Issue #4660, opened 2026-01-30
Use std::Format instead of char buffers


DigraJatin:
As suggested here,
https://github.com/canonical/mir/pull/4593/changes#r2661184043
Replace use of string buffers with std::format.
Can be done throughout codebase.

AlanGriffiths:
While this may be a good idea in some places, `std::format` is going to allocate memory which may affect performance in some cases or violate async-signal-safety in others.

[edit] it is also, at least theoretically, possible to introduce exceptions in `noexcept` contexts (such as destructors).

PR #4755, opened 2026-03-09
std::format in test_glmark2.cpp (see issue #4660)

yotam-medini:
Closes nothing.

Related: #4660

## What's new?

Related to PR #4736
std::format instead of snprintf in test_glmark2-es2.cpp

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/tests/performance-tests/test_glmark2-es2.cpp
+++ b/tests/performance-tests/test_glmark2-es2.cpp
@@ -64,9 +64,8 @@ struct AbstractGLMark2Test : testing::Test, mtf::AsyncServerRunner {
         auto const output_filename = std::format("/tmp/{}_{}.log",
                  test_info->test_case_name(), test_info->name());
 
-        printf("Saving GLMark2 detailed results to: %s\n",
-            output_filename.c_str());
-        // ^ Although I would vote to just print them to stdout instead
+        std::cout << std::format("Saving GLMark2 detailed results to: {}\n",
+            output_filename);
 
         std::string line;
         std::ofstream glmark2_output;
@@ -158,7 +157,7 @@ struct GLMark2Wayland : AbstractGLMark2Test
 struct HostedGLMark2Wayland : GLMark2Wayland
 {
     static char const* const host_socket;
-    const char *host_output_filename{nullptr};
+    std::string host_output_filename;
     pid_t pid = 0;
 
     HostedGLMark2Wayland()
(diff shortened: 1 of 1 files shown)


tests/performance-tests/test_glmark2-es2.cpp
```diff
@@ -60,12 +61,11 @@ struct AbstractGLMark2Test : testing::Test, mtf::AsyncServerRunner {
...
 
-        char output_filename[256];
-        snprintf(output_filename, sizeof(output_filename) - 1,
-                 "/tmp/%s_%s.log",
+        auto const output_filename = std::format("/tmp/{}_{}.log",
                  test_info->test_case_name(), test_info->name());
 
-        printf("Saving GLMark2 detailed results to: %s\n", output_filename);
+        printf("Saving GLMark2 detailed results to: %s\n",
+            output_filename.c_str());
```
AlanGriffiths: Wouldn't it be consistent to change from using `printf()` here?

tests/performance-tests/test_glmark2-es2.cpp
```diff
@@ -158,17 +158,17 @@ struct GLMark2Wayland : AbstractGLMark2Test
 struct HostedGLMark2Wayland : GLMark2Wayland
 {
     static char const* const host_socket;
-    char host_output_filename[256];
+    const char *host_output_filename{nullptr};
```
AlanGriffiths: There seems to be no good reason to use a `char const*` here (and not a `std::string`) and it is definitely more for the reader to understand than using a separate, `private`,  `std::string` to hold the value

tests/performance-tests/test_glmark2-es2.cpp
yotam-medini: done

tests/performance-tests/test_glmark2-es2.cpp
yotam-medini: Originally I wanted to minimize changes and avoid multi-use of ...c_str().
But I followed your suggestion.

AlanGriffiths: > Originally I wanted to minimize changes

For the sake of the next developer, aim for the simplest resulting code, not the simplest change.

C.f. [The Boy Scout Rule](https://97-things-every-x-should-know.gitbooks.io/97-things-every-programmer-should-know/content/en/thing_08/)

tests/performance-tests/test_glmark2-es2.cpp
```diff
@@ -27,6 +27,7 @@
 
 #include <cstdio>
```
AlanGriffiths, line 28: We can probably drop the direct include of `<cstdio>` as we're not using `{sn,}printf()` from it

PR #4757, opened 2026-03-09
Generating C++ to Rust binding so that the Wayland C++ objects can publish events through Rust to the clients, but not yet fully connecting them

mattkae:
## What's new?
- Generating `src/middleware.rs`, which contains Rust structs that ferry data from Rust to C++
- Generating `src/ffi.rs`, which contains:
   1. FFI definitions for the middleware structs that sanitize the data going across the language boundary (C++ -> Rust)
   2. FFI definitions for the C++ implementation classes (Rust -> C++)
- Adding an `associate` method to each Wayland interface's C++ class. This class will "associate" the `Box<InterfaceExt>` struct with the C++ class so that the C++ class can call methods on its underlying Rust implementation
- Introduced `include/ffi_fwd.h` to get around a circular dependency issue where the generated C++ headers want to include the Rust headers, but the Rust headers want to include the generated C++ headers. This came about because the `<Interface>Impl` class now hold an `<Interface>Ext` box.

## How to test
1. Run `cargo build` from `src/wayland/wayland_rs`
2. See `src/middleware.rs`
3. See `src/ffi.rs`
4. See `src/dispatch.rs`

image

image

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/wayland/wayland_rs/example/main.cpp
+++ b/src/wayland/wayland_rs/example/main.cpp
@@ -58,7 +58,9 @@ int main()
     std::cout << "    3) The server stops." << std::endl;
     std::cout << "    4) We exit with exit code 0." << std::endl << std::endl;
 
-    auto server = mir::wayland_rs::create_wayland_server();
+    // TODO: Add a real GlobalFactory here, but we will keep it nullptr for the time being
+    // as that task is involved.
+    auto server = mir::wayland_rs::create_wayland_server(nullptr);
     std::thread server_thread([&] { run_server(server); });
 
     run_client();


robert-ancell (commented):
I've read through this; there's a lot going on and I can only follow parts of it. I can't see any issues but it needs a second opinion / someone more knowledgeable in Rust to have a look too.

src/wayland/wayland_rs/build_script/helpers.rs
```diff
@@ -0,0 +1,60 @@
...
+    fs::write(dest_path, content).unwrap();
+}
+
+pub fn dash_to_snake(name: &str) -> String {
+    name.replace('-', "_")
+}
+
+pub fn dash_to_snake_ident(name: &str) -> Ident {
+    format_ident!("{}", dash_to_snake(name))
+}
```
RAOF, line 36: These are unused, right?

Incidentally, if there's much more case manipulation needed, [heck](https://libs.rs/heck) might be a good choice.

RAOF (commented):
The actual code generation code is fine.

I'm still digesting the code that it's generating; it seems at this point that we probably need more documentation about the high-level theory of what's happening here, because it's very much non-obvious.

mattkae:
> The actual code generation code is fine.
> 
> I'm still digesting the code that it's generating; it seems at this point that we probably need more documentation about the high-level theory of what's happening here, because it's very much non-obvious.

I was going to leave this as a task for later where I write a big diagram/markdown of what is getting generated and why. I am waiting for "later" because things feel prone to change now, and I don't want to keep rewriting it

src/wayland/wayland_rs/build_script/helpers.rs
mattkae: `dash_to_snake_ident` is still used

mattkae:
@RAOF I update to use pin_mut, UniquePtr, and with comments to make things easier to follow :)

PR #4767, opened 2026-03-11
std::format instead of snprintf in logger.cpp

yotam-medini:
Closes nothing

Related: #4660 

Related to #4736

## What's new?

std::format instead of snprintf in logger.cpp

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/core/logging/logger.cpp
+++ b/src/core/logging/logger.cpp
@@ -98,8 +98,7 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
         auto now = system_clock::now();
         auto now_us = time_point_cast<microseconds>(now);
         auto local = zoned_time{current_zone(), now_us};
-        std::format_to(std::ostreambuf_iterator<char>(out),
-            "[{:%F %T}] {}{}: {}\n",
+        std::format_to(std::ostreambuf_iterator{out}, "[{:%F %T}] {}{}: {}\n",
             local, lut[static_cast<int>(severity)], component, message);
         out.flush();
     }


src/core/logging/logger.cpp
```diff
@@ -81,27 +83,50 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
+        fmt_now = std::format("{:%F %T}", local);
+    }
+    catch (...)
+    {
+        auto now_time_t = std::chrono::system_clock::to_time_t(now);
+        auto micros = now_us.time_since_epoch().count() % 1000000;
+        char buf[32]{};
+        std::tm tm{};
+        size_t offset;
+        if (gmtime_r(&now_time_t, &tm) != nullptr &&
```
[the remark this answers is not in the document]
yotam-medini: done

src/core/logging/logger.cpp
```diff
@@ -81,27 +83,50 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
+            fmt_now = buf;
+        }
+        else
+        {
+            fmt_now = "unknown-time";
+        }
+    }
 
+    // From here, no dynamic allocations
+    char ts_lut[64]; // "[yyyy:mm:dd HH:MM:SS.uuuuuu] < CRITICAL! >: "
```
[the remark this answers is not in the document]
yotam-medini: done.

src/core/logging/logger.cpp
```diff
@@ -81,27 +83,50 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
+        char buf[32]{};
+        std::tm tm{};
+        size_t offset;
+        if (gmtime_r(&now_time_t, &tm) != nullptr &&
+            (offset = std::strftime(
+                buf, sizeof(buf), "%Y-%m-%d %H:%M:%S", &tm)))
+        {
+            auto eos = std::format_to_n(buf + offset, sizeof(buf) - offset,
+                ".{:06}", micros).out;
+            *eos = '\0';
```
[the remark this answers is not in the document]
yotam-medini: done

src/core/logging/logger.cpp
```diff
@@ -81,27 +83,50 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
+        {
+            fmt_now = "unknown-time";
+        }
+    }
 
+    // From here, no dynamic allocations
+    char ts_lut[64]; // "[yyyy:mm:dd HH:MM:SS.uuuuuu] < CRITICAL! >: "
+    auto ctx_eos = std::format_to_n(ts_lut, sizeof(ts_lut), "[{}] {}",
+        fmt_now, lut[static_cast<int>(severity)]).out;
+    *ctx_eos = '\0';
```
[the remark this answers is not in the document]
yotam-medini: done

src/core/logging/logger.cpp
```diff
@@ -81,27 +83,50 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
+            auto eos = std::format_to_n(buf + offset, sizeof(buf) - offset - 1,
+                ".{:06}", micros).out;
+            *eos = '\0';
+            fmt_now = buf;
+        }
+        else
+        {
+            fmt_now = "unknown-time";
+        }
+    }
```
robert-ancell, line 108: Since this is such a large block of code this will be more readable in a helper function e.g. `std::string generate_timestamp()`

yotam-medini: done.

src/core/logging/logger.cpp
```diff
@@ -70,6 +72,42 @@ void ml::set_logger(std::shared_ptr<Logger> const& new_logger)
     }
 }
 
+static std::string format_timestamp()
```
AlanGriffiths: We use the C++ idiom of anonymous namespaces to prevent symbols leaking, not `static`

src/core/logging/logger.cpp
```diff
@@ -81,27 +119,19 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
-        << "] "
-        << lut[static_cast<int>(severity)]
-        << component
-        << ": "
-        << message
-        << std::endl;
-
+    const auto fmt_now = format_timestamp();
+
+    // From here, no dynamic allocations
```
AlanGriffiths: This comment draws attention to the fact that the function call on the previous line does introduce dynamic allocation. That could have a performance penalty in widely used logging code or, theoretically, introduce exceptions in noexcept code.

src/core/logging/logger.cpp
yotam-medini: done

src/core/logging/logger.cpp
yotam-medini: In recent commit 
fe3781af yotam.m  rewrite of format_timestamp(...), no allocation
no dynamic allocations, and this comment removed.

src/core/logging/logger.cpp
```diff
@@ -70,6 +72,25 @@ void ml::set_logger(std::shared_ptr<Logger> const& new_logger)
...
+    char *eos;
+    try {
+        auto local =
+            std::chrono::zoned_time{std::chrono::current_zone(), now_us};
+        eos = std::format_to_n(buffer, bufsize - 1, "{:%F %T}", local).out;
+    } catch (...) {
+        eos = std::format_to_n(buffer, bufsize - 1, "{:%F %T} UTC", now_us).out;
+    }
+    *eos = '\0';
+}
```
AlanGriffiths: A significant part of the motivation to move away from `snprintf()` is the fragile nature of passing `char*, size_t` pairs.

It is possible do avoid this for C++ functions with something like:

```suggestion
using TimeStampBuffer = char[64];

void format_timestamp(TimeStampBuffer& buffer)
{
    auto now = std::chrono::system_clock::now();
    auto now_us = std::chrono::time_point_cast<std::chrono::microseconds>(now);
    char *eos;
    try {
        auto local =
            std::chrono::zoned_time{std::chrono::current_zone(), now_us};
        eos = std::format_to_n(buffer, sizeof buffer -1, "{:%F %T}", local).out;
    } catch (...) {
        eos = std::format_to_n(buffer, sizeof buffer -1, "{:%F %T} UTC", now_us).out;
    }
    *eos = '\0';
}
```

But we still end up doing the same thing for `std::format_to_n()`.

yotam-medini: done via using TimeStampBuffer = std::array<char, 64>

yotam-medini:
>>  _Is there anything it makes better?_

In my recent commit I tried to avoid explicit passing the size-limit
in "our" mir functions (via TimeStampBuffer).
As I understood from previous review comments, in ml::format_message
we want to avoid dynamic allocation, so exceptions will not be thrown.
Thus we cannot use std::format(...) and use automatic buffer on the stack.

If we want to move to modern C++ API as reasonably possible,
Can you suggest an alternative to C++ format-API?

Or, would you suggest to give up on eliminating snprintf 
from  ml::format_message(...) in logger.cpp?

AlanGriffiths:
> > _Is there anything it makes better?_
> 
> In my recent commit I tried to avoid explicit passing the size-limit in "our" mir functions (via TimeStampBuffer). As I understood from previous review comments, in ml::format_message we want to avoid dynamic allocation, so exceptions will not be thrown. Thus we cannot use std::format(...) and use automatic buffer on the stack.

Yes, we've explored alternatives: None of these appear to be a clear improvement to the existing code.

> If we want to move to modern C++ API as reasonably possible, Can you suggest an alternative to C++ format-API?

We seem to have different objectives: I want to improve the code, you want to "move to modern C++ API as reasonably possible". While new APIs and features are, in general, good options they are not _always_ an improvement.

> Or, would you suggest to give up on eliminating snprintf from ml::format_message(...) in logger.cpp?

I'm open to being convinced some variant of this is an improvement. But:

1. one call to `snprintf()` seems simpler than two calls to `format_to_n()` (and some logic);
2. `snprintf()` and  `format_to_n()` have the same memory safety issue; and,
3. `format(...)` leads to potential allocations with possible performance or exceptions being raised.

yotam-medini:
Before coding. 
How about avoiding char buffers and use something like:
```
   // no (char*,size) pairs, direct write to the stream
   auto iter = std::ostreambuf_iterator<char>(out);
   iter = std::format_to(iter, "{:%F %T}", ...);
   [iter =] std::format_to(iter, " {}  ...", severity, ...);
   ...
   out << std::endl; // for flush

```
While at it. When I tested the try/catch section in "my" format_timestamp(...)
I wanted to fail the current_zone() and have the exception caught.
By playing with env-vars ("TZ", "TZINFO"} or [re]moving system files
/etc/localtime, or under /usr/share/zoneinfo, 
but current_zone() overcame all such "landmines".
Eventually, I just added temporary throw for testing.

How about breaking backward compatibility, and just use:
  ... , "{:%F %T} UTC", now_us)

AlanGriffiths:
> Before coding. How about avoiding char buffers and use something like:
...
>    iter = std::format_to(iter, "{:%F %T}", ...);

That would avoid both allocating a `string` and passing `..., buf, len, ...`; and also looks like it will actually compile with g++-13. So that could actually be an improvement.

> While at it. When I tested the try/catch section in "my" format_timestamp(...) I wanted to fail the current_zone() and have the exception caught.

Maybe that points to a design issue with `format_timestamp(...)`? The try-catch logic could, in principle, be separated out. But this logic is too elaborate: if the runtime cannot find any `tzdb`, the environment is likely toxic anyway. And throwing and catching an exception for every log message is expensive. Maybe:

```c++
void format_timestamp(...)
try
{
    ...
}
catch (std::runtime_error const& e)
{
    mir::fatal_error_abort("Cannot format timestamp: %s", e.what());
}
```

src/core/logging/logger.cpp
```diff
@@ -14,14 +14,19 @@
...
 #include <mir/logging/dumb_console_logger.h>
 #include <mir/logging/logger.h>
 
 #include <iostream>
+#include <chrono>
+#include <format>
 #include <mutex>
 #include <cstdarg>
 #include <cstdio>
 #include <ctime>
```
AlanGriffiths: I don't think we need cstdio or ctime anymore

src/core/logging/logger.cpp
```diff
@@ -70,6 +75,25 @@ void ml::set_logger(std::shared_ptr<Logger> const& new_logger)
...
+    try {
+        auto local =
+            std::chrono::zoned_time{std::chrono::current_zone(), now_us};
+        iter = std::format_to(iter, "[{:%F %T}]", local);
+    } catch (std::runtime_error const& e) {
+        mir::fatal_error_abort("Cannot format timestamp: %s", e.what());
+    }
+    return iter;
+}
+}
```
AlanGriffiths: I think this can all be inlined - it is simple enough to be part of `ml::format_message()`

src/core/logging/logger.cpp
```diff
@@ -81,27 +105,17 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
-        << lut[static_cast<int>(severity)]
-        << component
-        << ": "
-        << message
-        << std::endl;
-
+    OutIter iter{out};
+    iter = format_timestamp(iter);
+    format_to(iter, "{}{}: {}\n",
+        lut[static_cast<int>(severity)], component, message);
```
AlanGriffiths: After inlining `format_timestamp()`, the `format_to()` statements can be combined into a single call.

src/core/logging/logger.cpp
```diff
@@ -70,6 +75,25 @@ void ml::set_logger(std::shared_ptr<Logger> const& new_logger)
...
+{
+    auto now = std::chrono::system_clock::now();
+    auto now_us = std::chrono::time_point_cast<std::chrono::microseconds>(now);
+    try {
+        auto local =
+            std::chrono::zoned_time{std::chrono::current_zone(), now_us};
+        iter = std::format_to(iter, "[{:%F %T}]", local);
+    } catch (std::runtime_error const& e) {
+        mir::fatal_error_abort("Cannot format timestamp: %s", e.what());
+    }
```
AlanGriffiths: According to the style guide the braces should be on their own lines.

src/core/logging/logger.cpp
yotam-medini: We still need <cstdio>, since logger.cpp still uses vsnprintf(...)
that is not addressed in this task.

src/core/logging/logger.cpp
yotam-medini: In recent commit:
  c9461d2b yotam.m inline format_timestamp, single std::format_to call(...)
I tried to follow your suggection.
Two remarks.
1. My earlier separation of format_timestamp(...) was following
   Robert's suggesion from March 11.
2. In general, don't we want to localize try/catch block
   to the smallest possible code, for which exception may be thrown?
   In this case: std::chrono::current_zone().

AlanGriffiths: 1. It is simpler now than when Robert suggested separating it
2. That depends on how it affects the complexity of the code. We _could_ write:

```c++
auto const local = []
        {
            try
            {
                return std::chrono::zoned_time{std::chrono::current_zone(), now_us};
            }
            catch (std::runtime_error const& e)
            {
                mir::fatal_error_abort("Cannot format log timestamp: %s", e.what());
            }
        }();
```

But that's an absurd extreme.

yotam-medini:
BTW, how can I run a single failed CI test. For example: the current single failed:
Spread / BuildAndTest (lxd:...:spread/ubuntu:asan) (pull_request)Failing after 4m 
?

src/core/logging/logger.cpp
yotam-medini: inlined

src/core/logging/logger.cpp
yotam-medini: single call now.

src/core/logging/logger.cpp
```diff
@@ -81,27 +85,28 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
+    {
+        std::ostreambuf_iterator<char> iter{out};
+        auto now = std::chrono::system_clock::now();
+        auto now_us =
+            std::chrono::time_point_cast<std::chrono::microseconds>(now);
+        auto local =
+            std::chrono::zoned_time{std::chrono::current_zone(), now_us};
+        std::format_to(iter, "[{:%F %T}] {}{}: {}\n",
+            local, lut[static_cast<int>(severity)], component, message);
+        out.flush();
```
AlanGriffiths: We don't really need a name for `iter`:

```suggestion
        auto now = std::chrono::system_clock::now();
        auto now_us =
            std::chrono::time_point_cast<std::chrono::microseconds>(now);
        auto local =
            std::chrono::zoned_time{std::chrono::current_zone(), now_us};
        std::format_to(std::ostreambuf_iterator{out}, "[{:%F %T}] {}{}: {}\n",
            local, lut[static_cast<int>(severity)], component, message);
        out.flush();
```

src/core/logging/logger.cpp
```diff
@@ -81,27 +85,28 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
+    }
+    catch (std::runtime_error const& e)
+    {
+        mir::fatal_error_abort("Cannot format message: %s", e.what());
```
AlanGriffiths: This message is a little too generic: we ought to mention logging so that the reader has some idea where it comes from:

```suggestion
        mir::fatal_error_abort("Cannot format log message: %s", e.what());
```

src/core/logging/logger.cpp
```diff
@@ -14,14 +14,18 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <mir/fatal.h>
```
AlanGriffiths: Project standard is to begin a .cpp file with the corresponding .h (to ensure the .h has no hidden dependencies).

AlanGriffiths:
> BTW, how can I run a single failed CI test. For example: the current single failed: Spread / BuildAndTest (lxd:...:spread/ubuntu:asan) (pull_request)Failing after 4m ?

I doubt you have the permissions to re-run CI tasks. In this case it looks like a network timeout, nothing to do with your changes.

Team members (like me) could re-run it, but as I've requested changes there's no point.

src/core/logging/logger.cpp
```diff
@@ -81,27 +85,27 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
-    timespec ts{};
-    clock_gettime(CLOCK_REALTIME, &ts);
-    char now[32];
-    auto offset = strftime(now, sizeof(now), "%F %T", localtime(&ts.tv_sec));
-    snprintf(now+offset, sizeof(now)-offset, ".%06ld", ts.tv_nsec / 1000);
-
     if (!out || !out.good())
     {
         std::cerr << "Failed to write to log file: " << errno << std::endl;
         return;
```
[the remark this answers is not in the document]
yotam-medini: include <cerrno> added.

src/core/logging/logger.cpp
```diff
@@ -81,27 +86,27 @@ void ml::format_message(std::ostream& out, Severity severity, std::string const&
...
-        << message
-        << std::endl;
-
+    try
+    {
+        using namespace std::chrono;
+        auto now = system_clock::now();
+        auto now_us = time_point_cast<microseconds>(now);
+        auto local = zoned_time{current_zone(), now_us};
+        std::format_to(std::ostreambuf_iterator<char>(out),
```
AlanGriffiths: It's only six characters, but the compiler can deduce the template parameter.
```suggestion
        std::format_to(std::ostreambuf_iterator(out),
```

yotam-medini:
documentation-checks ... failed with:
matplotlib is not installed, social cards will not be generated
<rst_epilog>:2: CRITICAL: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'reuse/links.txt'.
<rst_epilog>:3: CRITICAL: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'reuse/substitutions.txt'.

Do I need to do anything to have this PR merged ?

AlanGriffiths:
> Do I need to do anything to have this PR merged

No, I'm dealing with it

PR #4771, opened 2026-03-13
doc: add sitemap configuration

Saviq:
## What's new?

Correct sitemap chain.

## How to test

Check that `/robots.txt`, `/sitemapindex.xml` and `/sitemap.xml` are reachable and their content is valid (the whole chain will only start working when `stable` and `v2.17.2` get this, and updated Starter Pack.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/doc/sphinx/robots.txt
+++ b/doc/sphinx/robots.txt
@@ -1,5 +1,6 @@
 User-agent: *
 
-Disallow: # Allow everything
+# Allow everything
+Disallow:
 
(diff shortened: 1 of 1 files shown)


doc/sphinx/robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: *
+
+Disallow: # Allow everything
```
tarek-y-ismail: Seems like we dropped the fix that Copilot mentioned.

Saviq: Good catch! Had some entanglement between this and the Frame one…

Issue #4775, opened 2026-03-17
Flaky `wlcs.ExtImageCopyCaptureTest.cursor_session_captures_pointer_image` test


Saviq:
### Version

v2.25.2

### Environment

CI

### What happened? What did you expect to happen?

The new ImageCopy test is flaky, see https://github.com/canonical/mir/actions/workflows/spread.yml

NB: this may be an issue with [the test](https://github.com/canonical/wlcs/blob/b814ac24609110f16f36019828bb3eed79c3af28/tests/ext_image_copy_capture_v1.cpp#L517-L581), not necessarily with Mir.

### Reproduction steps

```
/usr/libexec/wlcs/wlcs ./build-Debug/lib/miral_wlcs_integration.so --gtest_filter='ExtImageCopyCaptureTest.cursor_session_captures_pointer_image' --gtest_repeat=1000
```

### Relevant log output

```shell
[2026-03-17 09:55:26.710673] <information> input-hub: Device configuration: mouse, capabilities={pointer}, pointer_config={handedness=right,cursor-acceleration=none,cursor-acceleration-bias=0.00,scroll-hspeed=1.00,scroll-vspeed=1.00}
./tests/ext_image_copy_capture_v1.cpp:552: Failure
Value of: capture_session.buffer_size()
Expected: value is equal to (32, 32)
  Actual: ((1, 1)), whose value (1, 1) doesn't match

[2026-03-17 09:55:26.713010] <information> mirserver: Stopping
[2026-03-17 09:55:26.732049] < -warning- > security: {"datetime": "2026-03-17T09:55:26Z", "appid": "wlcs", "event": "sys_shutdown", "level": "WARN", "description": "Mir is shutting down" }
[  FAILED  ] ExtImageCopyCaptureTest.cursor_session_captures_pointer_image (126 ms)
```

PR #4776, opened 2026-03-17
Perform DRM client cap check with proper ioctl call

### LINKED ISSUE #4774: DRM_CLIENT_CAP_ATOMIC check is performed with wrong ioctl call
### Version

v2.22.0

### Environment

Board: NXP IMX8MM
System: Ubuntu (custom integration)

Hello. I'm using Mir on NXP's IMX8MM platform and I believe that the Mir component doesn't recognize correctly the support of Atomic KMS in my driver (imx-drm) which, as confirmed by other tools and examinations, supports mentioned capability. I have a suspicion there is a tiny bug in the check procedure and would like to ask for confirmation...

Mir queries atomic capability check via `mgk::get_cap_checked` function here:

https://github.com/canonical/mir/blob/3a7a35e9950e707458da76ccdf442309299737c5/src/platforms/atomic-kms/server/kms/platform_symbols.cpp#L208

On my platform, this results with prints:

```
[2026-03-13 09:58:55.494745] < - debug - > atomic-kms: Quirks(skip/allow): checking device with devnode: /dev/dri/card1, driver imx-drm
[2026-03-13 09:58:55.495164] <information> atomic-kms: KMS device /dev/dri/card1 does not support Atomic KMS
```

However, one thing bothers me. When we dig into the implementation, we can see following chain of calls:

`mgk::get_cap_checked -> drmGetCap -> drmIoctl(fd, DRM_IOCTL_GET_CAP, &cap);`

The problematic part is `DRM_IOCTL_GET_CAP`. Looking at kernel sources

https://github.com/torvalds/linux/blob/master/include/uapi/drm/drm.h

one can see:

```
#define DRM_IOCTL_GET_CAP           DRM_IOWR(0x0c, struct drm_get_cap)
#define DRM_IOCTL_SET_CLIENT_CAP    DRM_IOW( 0x0d, struct drm_set_client_cap)
```

`drm_get_cap` looks like it should not be probed for client capabilities. For client caps, one should use `drm_set_client_cap` (`DRM_IOCTL_SET_CLIENT_CAP`) and check if it succeeded. So instead of:

`if (!mgk::get_cap_checked(tmp_fd, DRM_CLIENT_CAP_ATOMIC)) // which then calls drmGetCap(...)`

Mir should call something like:

`if (drmSetClientCap(tmp_fd, DRM_CLIENT_CAP_ATOMIC, 1) != 0)`

This is visible when you search "globally" for `DRM_CLIENT_CAP_ATOMIC` - from what I checked, in vast majority (all?) of other projects which sources are available on github, the verification is performed by calling `drmSetClientCap` and then checking the result of this call. Please also see this file:

https://github.com/torvalds/linux/blob/3abab27c322e0f2acf981595aa8040c9164dc9fb/drivers/gpu/drm/drm_ioctl.c

`DRM_CLIENT_CAP_ATOMIC` is handled in switch-case under `drm_setclientcap`, not under `drm_getcap`.

Moreover, the kernel docs (https://docs.kernel.org/gpu/drm-uapi.html#c.DRM_CLIENT_CAP_ATOMIC) state:

> DRM_CLIENT_CAP_ATOMIC
> [...]
> If set to 1, the DRM core will expose atomic properties to userspace. This implicitly enables DRM_CLIENT_CAP_UNIVERSAL_PLANES and DRM_CLIENT_CAP_ASPECT_RATIO.
> **If the driver doesn’t support atomic mode-setting, enabling this capability will fail with -EOPNOTSUPP.**
> [...]

Could someone confirm / deny that the procedure for probing Atomic KMS capability is faulty? If my understanding is correct, then I can prepare a one-line patch that will change the cap check. When I applied this small patch on my platform, the check was successful and implementation allowed me to use Atomic KMS with no visible issues.

### Reproduction steps

No reproduction steps as this is an "experimental" integration, although the question is generic and applicable for the whole component.

### Relevant log output

```shell

```

krzysztof-buczak-red:
Closes #4774

## What's new?

The DRM_CLIENT_CAP_ATOMIC is now checked via drmSetClientCap, not mgk::get_cap_checked (which evaluates to drmGetCap). Previous call was not reliable as drmGetCap is not able to validate client caps.

## How to test

- Code compiles.
- Platform with Atomic KMS support reports it successfully.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/platforms/atomic-kms/server/kms/platform_symbols.cpp
+++ b/src/platforms/atomic-kms/server/kms/platform_symbols.cpp
@@ -205,7 +205,7 @@ auto probe_display_platform(
                         std::string{"Failed to set DRM interface version on device "} + device.devnode()}));
                 }
 
-                if (!mgk::get_cap_checked(tmp_fd, DRM_CLIENT_CAP_ATOMIC))
+                if (drmSetClientCap(tmp_fd, DRM_CLIENT_CAP_ATOMIC, 1) != 0)
                 {
                     mir::log_info("KMS device %s does not support Atomic KMS", device.devnode());
                     continue;


AlanGriffiths:
Build failure is #4775

PR #4781, opened 2026-03-18
Live config override directory

tarek-y-ismail:
## What's new?

- `ConfigFile` now discovers and monitors `<config-file>.d/` directories, automatically loading `.conf` files in lexicographic order.

## How to test

```
 ./build/bin/miral-test --gtest_filter="TestOverrideConfigFile*:TestMultirootConfigFile*"
 ./build/bin/miral-test-internal --gtest_filter="TestOverrideWatcher*:TestMultiRootOverrideWatcher*:TestThreeRootOverrideWatcher*"
```

## Checklist

- [x] Tests added and pass


tarek-y-ismail:
I did not realize I wrote 1600+ lines of code. Might be wise to split this into a PR for the `IniFile` changes and another for the `ConfigFile` changes

tarek-y-ismail:
Ok, splitting `IniFile` changes out improved things a bit. But `ConfigFile` tests are almost 1000 lines, so the PR is still big...

tarek-y-ismail:
After a quick discussion with @Saviq, we're going to implement this without support for symlinks initially. That can come later.

src/miral/config_file.cpp
```diff
@@ -14,6 +14,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <filesystem>
```
AlanGriffiths: Not as first include!

Saviq:
`std::ranges::to` soon, but not just yet :pray:

```
:: 2026-05-04 13:03:12.301 ::   173 |         | std::views::transform([](auto const& f) { return f.path(); }) | std::ranges::to<std::vector>();
```

tarek-y-ismail:
Did a quick initial review and cleaned up some stuff. I think it's ready for review now. I'll probably give it another pass tomorrow as well.

AlanGriffiths:
> * `ConfigFile` now discovers and monitors `<config-file>.d/` directories, automatically loading `.conf` files in lexicographic order.

Why `.conf`? We have configuration in `.config`, `.display`, `.layout` and `.settings` files. Surely this should be configurable (derived from the filename?).

tests/miral/config_file.cpp
```diff
@@ -618,3 +628,665 @@ TEST_F(TestConfigFile, with_no_reloading_a_config_in_xdg_conf_dir2_is_loaded)
 
     wait_for_load();
 }
+
+
+namespace
+{
+using namespace std::string_literals;
+// The override config file name must match the pattern: <name>.d/*.conf
+// For a config file named "test_override.config", the override directory is "test_override.config.d/"
```
AlanGriffiths: This seems an odd requirement and unlike existing practice. E.g.

```
$ ls /etc/apt/sources.list{,.d}
/etc/apt/sources.list

/etc/apt/sources.list.d:
mir-team-ubuntu-dev-noble.sources                          ubuntu-esm-apps.sources.save
mir-team-ubuntu-dev-noble.sources.save                     ubuntu-esm-infra.sources
mir-team-ubuntu-rc-noble.sources                           ubuntu-esm-infra.sources.save
ubuntu-enterprise-desktop-ubuntu-authd-noble.sources       ubuntu.sources
ubuntu-enterprise-desktop-ubuntu-authd-noble.sources.save  ubuntu.sources.curtin.orig
ubuntu-esm-apps.sources                                    ubuntu.sources.save
```

(The `.sources` files are used)

src/miral/config_file.cpp
```diff
@@ -85,33 +100,50 @@ class Watcher : public std::enable_shared_from_this<Watcher>
...
-    Self(MirRunner& runner, path file, Mode mode, Loader load_config);
+    using OverrideLoader = miral::ConfigFile::OverrideLoader;
+    OverrideWatcher(path base_config, OverrideLoader load_config);
 
+    void register_handler(miral::MirRunner& runner);
 private:
-    std::shared_ptr<Watcher> watcher;
-};
+    void handler(int);
+    // Someone smell Coffee?
```
AlanGriffiths: Tea, please!

AlanGriffiths (changes requested):
Initial, high level comments.

I need to look closer, but this looks quite disruptive of the existing ConfigFile implementation. I would have expected adding this functionality to be largely "add", not mostly "amend". (E.g. by deriving the override implementation from the existing Self). That makes reviewing harder as both the existing and new implementation need to be considered.

Saviq:
> Why `.conf`?

That's on me, oversimplified in the design discussion.

And sorry for the close/open dance, fat-fingered.

tarek-y-ismail:
> Why .conf? We have configuration in .config, .display, .layout and .settings files. Surely this should be configurable (derived from the filename?).

I'm not particularly attached to a specific suffix. If we were to derive it from the filename, then my suggestion is to use the same suffix as the base config. Are you on board?

tarek-y-ismail:
> I need to look closer, but this looks quite disruptive of the existing ConfigFile implementation.

The diff is indeed messy, but it's really a straightforward extension of what we have. A new constructor is added to allow access to `FileChanges` via `OverrideLoader`. And instead of using a `Watcher`, an `OverrideWatcher` is used which handles all the work needed to collect/watch files and directories.

The only blemish I can clearly see is that we store both `Watcher` and `OverrideWatcher`. It will be fairly trivial to derive a common interface to fix that. Otherwise, I don't see much disruption. Though I will try making the diff a bit easier to read.

AlanGriffiths:
> > 
> 
> I'm not particularly attached to a specific suffix. If we were to derive it from the filename, then my suggestion is to use the same suffix as the base config. Are you on board?

That also doesn't match the example I reference above. Maybe we should let the user decide?

src/miral/config_file.cpp
```diff
@@ -165,68 +376,271 @@ miral::ConfigFile::ConfigFile(MirRunner& runner, path file, Mode mode, Loader lo
...
+            auto const base_config_changed =
+                write_or_move && is_base_config_directory_event && event.name == base_config_filename;
+            auto const override_directory_created =
+                create && is_base_config_directory_event && event.name == override_directory;
+            auto const override_directory_removed =
+                remove && is_base_config_directory_event && event.name == override_directory;
+
+            auto const is_conf_file = std::string_view{event.name}.ends_with(extension);
+            auto const override_file_changed = (write_or_move || remove) && is_override_directory_event && is_conf_file;
+
```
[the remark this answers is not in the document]
tarek-y-ismail: @copilot We don't watch on those events in `watch_descriptor`, why should I guard against them?

src/miral/config_file.cpp
```diff
@@ -136,25 +191,181 @@ miral::ConfigFile::Self::Self(MirRunner& runner, path file, Mode mode, Loader lo
...
+    mir::Fd const& inotify_fd, std::optional<path> const& base_config_directory, std::string const& override_directory)
+    -> std::optional<int>
+{
+    if (!base_config_directory)
+        return std::nullopt;
+
+    auto const override_directory_path = base_config_directory.value() / override_directory;
+    if (!exists(override_directory_path))
+    {
+        mir::log_warning("Override directory '%s' does not exist, ignoring", override_directory_path.c_str());
```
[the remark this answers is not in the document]
tarek-y-ismail: This is fine since it's called only on launch and when the directory is created

tests/miral/config_file.cpp
```diff
@@ -618,3 +628,656 @@ TEST_F(TestConfigFile, with_no_reloading_a_config_in_xdg_conf_dir2_is_loaded)
...
+}
+}
+
+TEST_F(TestOverrideConfigFile, override_loader_with_no_file_is_not_called)
+{
+    invoke_runner([this](miral::MirRunner& runner)
+    {
+        config = ConfigFile{
+            runner,
+            std::move("no_such_file.config"),
```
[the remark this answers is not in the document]
tarek-y-ismail: My bad, fam

tests/miral/config_file.cpp
```diff
@@ -40,13 +41,22 @@ class PendingLoad
...
-            std::cerr << "wait_for_load() timed out" << std::endl;
+            switch (fail_on_timeout)
+            {
+            case FailOnTimeout::yes:
+                FAIL() << "wait_for_load() timed out" << std::endl;
+                break;
+            case FailOnTimeout::no:
+                std::cerr << "wait_for_load() timed out" << std::endl;
+                break;
+            }
```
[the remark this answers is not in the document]
tarek-y-ismail: Fine!

src/miral/config_file.cpp
```diff
@@ -165,68 +368,280 @@ miral::ConfigFile::ConfigFile(MirRunner& runner, path file, Mode mode, Loader lo
...
+        }
+        else if (summary.override_dir_created)
+        {
+            reload_needed = on_override_directory_created(builder, base_config, make_stream);
+        }
+        else if (summary.override_dir_removed)
+        {
+            on_override_directory_removed(builder, base_config, make_stream);
+        }
+        else // override file changes
```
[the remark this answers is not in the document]
tarek-y-ismail: Tried poking this. Not really possible to reproduce it over here so I'm going to not think about it.

src/miral/config_file.cpp
```diff
@@ -136,25 +180,184 @@ miral::ConfigFile::Self::Self(MirRunner& runner, path file, Mode mode, Loader lo
...
+auto collect_override_files(path const& override_directory, std::string_view extension) -> std::vector<path>
+{
+    if(!exists(override_directory) || !is_directory(override_directory))
+        return {};
+
+    auto unsorted =
+        directory_iterator{override_directory}
+        | std::views::filter([extension](auto const& f)
+                             { return std::filesystem::is_regular_file(f) && f.path().extension() == extension; })
+        | std::views::transform([](auto const& f) { return f.path(); });
```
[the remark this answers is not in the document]
tarek-y-ismail: Should be handled now, but I think the whole file needs another cleanup pass to simplify everything and add more error checks. For example, inotify Fds look like they might benefit from their own RAII class. Another example is checking if a path exists/is a directory, which we _might_ not need to check at _every_ step like we do right now.

include/miral/miral/config_file.h
```diff
@@ -54,6 +56,21 @@ class ConfigFile
...
 
     ConfigFile(MirRunner& runner, std::filesystem::path file, Mode mode, Loader load_config);
+
+    /// Loads the base configuration file together with any override files from
+    /// `<config-file>.d/` in the same directory, applied in lexicographic
+    /// order. If the base file is not found, \p load_config is not invoked. If
+    /// the drop-in directory is empty, it is ignored; if it is absent or is not
+    /// a directory, that may be reported by the implementation.
+    /// \remark since MirAL 5.8
+    /// \param extension file extension (including the leading dot) used to filter override files.
```
AlanGriffiths: Why are we only documenting one parameter out of five?

src/miral/config_file.cpp
```diff
@@ -165,68 +368,280 @@ miral::ConfigFile::ConfigFile(MirRunner& runner, path file, Mode mode, Loader lo
...
 
-    alignas(inotify_event) char buffer[sizeof(inotify_event) + NAME_MAX + 1];
+OverrideWatcher::OverrideWatcher(path base_config, OverrideLoader load_config, std::string_view ext) :
+    override_loader{std::move(load_config)},
+    extension{std::move(ext)},
+    base_config_filename{base_config.filename().string()},
+    override_directory{base_config_filename + ".d"},
+    base_config_directory{config_directory(base_config)},
+    base_config_directory_watch_descriptor{watch_descriptor(inotify_fd, base_config_directory)},
+    override_directory_watch_descriptor{watch_override_directory(inotify_fd, base_config_directory, override_directory)}
```
AlanGriffiths: This copies logic from the single file implementation that is likely wrong for this requirement.

Specifically, it sets up a watch on `~/.config/base.file` and `~/.config/base.file.d` and not anywhere further along the list of places specified by the xdg base directory specification. If there are overrides in system config directories they are not handled by the watch.

src/miral/config_file.cpp
```diff
@@ -165,68 +368,280 @@ miral::ConfigFile::ConfigFile(MirRunner& runner, path file, Mode mode, Loader lo
...
+{
+    if (base_config_directory && override_directory_watch_descriptor)
+    {
+        // Watch is set up only if the override directory exists and is a directory.
+        auto const override_dir_path = base_config_directory.value() / override_directory;
+        known_override_files = collect_override_files(override_dir_path, extension);
```
AlanGriffiths: We should be watching for the override directory being created and set a watch if that happens.

src/miral/config_file.cpp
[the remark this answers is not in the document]
AlanGriffiths: It might be necessary to rework the code to test this case reliably, but it is logically possible and should be handled correctly

src/miral/symbols.map
```diff
@@ -460,10 +460,11 @@ local: *;
 
 MIRAL_5.1 {
 global:
+  _ZN5miral10ConfigFileC1ERNS_9MirRunnerENSt10filesystem7__cxx114pathENS0_4ModeESt8functionIFvRSiRKS5_EE;
```
AlanGriffiths, line 463: Worth a comment to explain why this is needed

AlanGriffiths (commented):
Slightly orthogonal to the PR itself (therefore non blocking):

We likely have multiple watches set on the same directories. We should aggregate these and register multiple handlers with each watch instead.

(I know that's an optimization rework for a pre-existing problem, but this usage will add to the problem.)

AlanGriffiths (changes requested):
Sorry, finger trouble: that last review should have been "Request changes"

include/miral/miral/config_file.h
tarek-y-ismail: To mention that the leading dot should be included and that this is for override files only. The other parameters are pretty self-explanatory, or follow an existing pattern.

src/miral/config_file.cpp
tarek-y-ismail: Already done a few dozen lines after. The comment was outdated and is now removed.

src/miral/symbols.map
tarek-y-ismail: For the same reason we use the mangled symbol at the start of the file (ld vs lld format differences). Didn't see a reason it warranted a duplicated comment.

tarek-y-ismail:
Still have a bunch of changes lined up (more refactoring of impl + refactoring of tests), but ran out of time for today. Thought this was a good time to push.

tarek-y-ismail:
@robert-ancell A new pair of eyes never hurts

include/miral/miral/config_file.h
```diff
@@ -54,6 +56,22 @@ class ConfigFile
     };
 
     ConfigFile(MirRunner& runner, std::filesystem::path file, Mode mode, Loader load_config);
+
+    /// Loads the base configuration file together with any override (drop-in)
+    /// files from `<config-file>.d/` directories across all XDG config roots.
+    /// Files are sorted by basename and deduplicated following systemd
+    /// conventions: when multiple roots provide a file with the same basename,
```
AlanGriffiths: Does mentioning `systemd` add anything? As we can't assume readers are familiar with these conventions, they need to be explained.

src/miral/config_file.cpp
```diff
@@ -14,138 +14,136 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include <miral/config_file.h>
+#include "miral/config_file.h"
```
AlanGriffiths: We should be using `<>` includes for files in the include directories

src/miral/config_file.cpp
```diff
@@ -14,138 +14,136 @@
...
+#include "live_config_overrides_list_builder.h"
+#include "live_config_watcher.h"
+#include "override_watcher.h"
 
-#include <miral/runner.h>
+#include "miral/live_config_overrides_list.h"
```
AlanGriffiths: We should be using `<>` includes for files in the include directories

src/miral/config_file.cpp
```diff
@@ -14,138 +14,136 @@
...
+    auto const it = std::ranges::find_if(
+        config_roots,
+        [&filename](path const& config_root)
+        {
+            std::error_code ec; // Don't care about the specific error. Likely "file not found"
+            return fs::exists(config_root / filename, ec);
+        });
+
+    if (it != config_roots.end())
+        return *it / filename;
```
AlanGriffiths, line 54:
```suggestion
    {
        return *it / filename;
    }
```

src/miral/config_file.cpp
```diff
@@ -14,138 +14,136 @@
...
     else
-    {
         return std::nullopt;
```
AlanGriffiths, line 56:
```suggestion
    {
        return std::nullopt;
    }
```

src/miral/live_config_overrides_list_builder.cpp
```diff
@@ -20,33 +20,21 @@
...
-                { callback(path, *stream_factory()); }}
+auto mlc::OverridesListBuilder::push_unchanged(std::filesystem::path filepath, Opener opener) -> OverridesListBuilder&
 {
-}
-
-auto mlc::OverridesListBuilder::push_unchanged(
-    std::filesystem::path filepath, std::function<std::unique_ptr<std::istream>()> stream) -> OverridesListBuilder&
-{
-    ctx->events.emplace_back(Context::Kind::unchanged, filepath, stream);
+    ctx->events.push_back({Context::Kind::unchanged, std::move(filepath), std::move(opener)});
```
AlanGriffiths: What motivates the change to `push_back()`? Doesn't that create a temporary (that may be elided) instead of constructing in place?

src/miral/live_config_overrides_list.cpp
```diff
@@ -26,55 +26,40 @@ mlc::OverridesList::OverridesList(std::unique_ptr<Context> ctx) :
...
             dropped(event.path);
-            return;
+            continue;
         }
-    };
 
-    for (auto const& event : ctx->events)
-    {
-        switch (event.kind)
+        // Openers must not throw; they signal failure by returning nullptr.
```
AlanGriffiths: Who is the audience for this comment? It seems unlikely that the implementer of an `Opener` will be guaranteed to read, remember and enforce this comment. So it isn't a requirement on `Opener` and it isn't repeating a requirement already on `Opener` (I looked).

Is it just saying "this code is broken if `opener()` can ever throw"?

src/miral/live_config_watcher.cpp
```diff
@@ -0,0 +1,244 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "live_config_watcher.h"
+#include "miral/live_config_overrides_list.h"
```
AlanGriffiths:
```suggestion
#include <miral/live_config_overrides_list.h>
```

AlanGriffiths (changes requested):
Initial review: there's a lot being changed and I don't have an overview yet. But this is a bunch of code hygiene issues that can be addressed

src/miral/config_file.cpp
tarek-y-ismail: Why? I think this is simple enough to not require braces. 

The general pattern I've seen in the codebase is that we usually add braces if the contained statement become unwieldy, or more than one statement is contained under the if statement. 

Sidenote: we should probably update https://github.com/canonical/mir/blob/61439601c785a0b1588704ca566274a870e606d2/doc/sphinx/contributing/reference/cppguide.md?plain=1#L2464-L2485 to mention this.

src/miral/live_config_overrides_list.cpp
tarek-y-ismail: Pretty much, yes. We control the innards of the list, the builder, and all openers created. But there is no way to check if an opener can throw. 

I presume we can `catch(...) { throw std::logic_error("Openers must not throw"); }` to signal that this is a programmer error. What do you think?

src/miral/live_config_overrides_list_builder.cpp
tarek-y-ismail: It's been a while now, but I remember having some compilation errors in this file due to use `std::move_only_function` instead of a plain function. So this is likely a leftover from trying to get the code to compile. Nonetheless I've restored the code and it should compile now. Thanks!

tarek-y-ismail:
> Initial review: there's a lot being changed

Yes. I've contemplated and played a lot with the Git history to try and make it easier to review. In the end, it's a lot of related code that will be difficult to split up, but I think the current git history should help guide your review.

src/miral/config_file.cpp
AlanGriffiths: A change to the guideline is best debated on a PR to change the guideline. (Personally, I have been following the guideline and don't see a need for change.)

src/miral/symbols.map
```diff
@@ -778,6 +779,8 @@ global:
 
 MIRAL_5.8 {
 global:
+  _ZN5miral10ConfigFileC1ERNS_9MirRunnerENSt10filesystem7__cxx114pathENS0_4ModeESt18move_only_functionIJFvRKNS_11live_config13OverridesListEEEESt17basic_string_viewIcSt11char_traitsIcEE;
+
```
AlanGriffiths: Wrong stanza

src/miral/config_file.cpp
```diff
@@ -16,136 +16,134 @@
...
+    config_streams.push_new(resolved_base_config, mlc::open_file);
+
+    // Collect override files from all roots, deduplicating by basename.
+    // Higher-priority roots (earlier in config_roots, later in reverse iteration) shadow lower ones.
+    std::map<std::string, path> by_basename;
+    for (auto const& root : config_roots | std::views::reverse)
+    {
+        for (auto const& override_file : mlc::collect_override_files(root / override_directory_name, extension))
+            by_basename.insert_or_assign(override_file.filename().string(), override_file);
+    }
```
AlanGriffiths: Is there any advantage of converting the `filename()` to a `std::string`?
```suggestion
    std::map<path, path> by_filename;
    for (auto const& root : config_roots | std::views::reverse)
    {
        for (auto const& override_file : mlc::collect_override_files(root / override_directory_name, extension))
            by_filename.insert_or_assign(override_file.filename(), override_file);
    }
```

tarek-y-ismail:
Merge went wrong. Cleaned up the commit history. No new changes.

src/miral/live_config_watcher.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    auto const readsize = read(inotify_fd, buffer.data(), buffer.size());
+    if (readsize < static_cast<ssize_t>(sizeof(inotify_event)))
+        return;
+
+    auto remaining = std::span{buffer}.first(static_cast<std::size_t>(readsize));
+    while (remaining.size() >= sizeof(inotify_event))
+    {
+        // LEGACY(Clang) Replace with std::start_lifetime_as<inotify_event const> once Clang supports it (C++23, P2590).
+        auto const& event = reinterpret_cast<inotify_event const&>(
```
AlanGriffiths: We should document why this is safe. In particular, `std::start_lifetime_as`/`reinterpret_cast` have UB when the alignment is unsuitable.

tests/miral/config_file.cpp
```diff
@@ -60,8 +72,9 @@ class PendingLoad
         cv.notify_one();
     }
 
-private:
     std::mutex mutex;
```
AlanGriffiths: Making a mutex public is weird: it suggests the code using is poorly encapsulated. The use (in `OverrideConfigTestBase::record_load()`) is also weird: some stuff is done under lock, and then the lock is released and reseized to do more work. It looks as though that work is not independent and should all be under a single lock.

debian/libmiral7.symbols
```diff
@@ -730,10 +730,13 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::WindowSpecification::alpha() const@MIRAL_5.7" 5.7.0
  (c++)"miral::WindowSpecification::alpha()@MIRAL_5.7" 5.7.0
  MIRAL_5.8@MIRAL_5.8 5.8.0
+ (c++)"miral::ConfigFile::ConfigFile(miral::MirRunner&, std::filesystem::__cxx11::path, miral::ConfigFile::Mode, std::move_only_function<void (miral::live_config::OverridesList const&)>, std::basic_string_view<char, std::char_traits<char> >)@MIRAL_5.8" 5.8.0
```
AlanGriffiths: Wrong stanza. ⛵

include/miral/miral/config_file.h
```diff
@@ -20,44 +20,92 @@
 #include <mir/fd.h>
 
 #include <filesystem>
-#include <istream>
 #include <functional>
+#include <iosfwd>
```
AlanGriffiths: We don't use `<iosfwd>` in this file

src/miral/override_watcher.cpp
```diff
@@ -0,0 +1,424 @@
...
+      known_override_files{std::move(known)}
+{
+}
+
+auto mlc::OverrideWatcher::WatchedRoot::has_base_watch() const -> bool
+{
+    return static_cast<bool>(base_dir_wd);
+}
+
+auto mlc::OverrideWatcher::WatchedRoot::classify_event(  // NOLINT(bugprone-easily-swappable-parameters)
```
AlanGriffiths: Do not sprinkle `NOLINT` liberally to make stuff pass: there should be an explanation why we are not applying the rule. (And rules we disagree with generally should be challenged.)

src/miral/override_watcher.h
```diff
@@ -0,0 +1,148 @@
...
+
+private:
+    auto find_base_config() const -> std::optional<std::filesystem::path>;
+    auto collect_inotify_events() const -> BatchSummary;
+
+    OverrideLoader override_loader;
+
+    std::string const extension;
+    std::string const base_config_filename;
+    std::string const override_directory_name;
```
AlanGriffiths: This code seems to be stringly typed. We have `std::filesystem` for better type safety, why not use it: it doesn't look like much churn

src/miral/live_config_overrides_list.cpp
```diff
@@ -26,55 +31,50 @@ mlc::OverridesList::OverridesList(std::unique_ptr<Context> ctx) :
...
-        catch (std::exception const& e)
+        catch (...)
         {
-            mir::log_warning("Error opening file %s: %s. Treating as a drop.", event.path.c_str(), e.what());
-            dropped(event.path);
-            return;
+            BOOST_THROW_EXCEPTION(
+                std::logic_error{
+                    std::format("Openers must not throw exceptions. Failed to open '{}'", event.path.string())});
         }
```
AlanGriffiths, line 58: Is throwing a new exception a sensible way to treat encountering an exception that must not be?

Also, are we really expecting to deal with exceptions that are not derived from `std::exception`? The previous code logging a warning (with some the exception content) and continue normal flow seems better. (Actually logging the full exception would be even better.)

And if it really "should not be" isn't a fatal error a better approach?

AlanGriffiths (changes requested):
Still getting into this, but a first batch of thoughts

src/miral/config_file.cpp
tarek-y-ismail: For a map mapping filenames to their path, a `string->path` map seemed more sensible than a `path->path` map. When I see `path`, it could contain the filename, a relative path, an absolute path, etc...

src/miral/live_config_overrides_list.cpp
tarek-y-ismail: Since we control both the creation and consumption of openers, an opener throwing is a programmer error. We shouldn't catch and log because this should blow up as loud as possible so it doesn't sneak past us.

Throwing a `logic_error` or raising a fatal error are both valid ways in my knowledge to signal a programmer error. Both exist in the code as well. 

I suppose in this case we don't want to ever recover, so a fatal error is better.

include/miral/miral/config_file.h
tarek-y-ismail: I thought we'd need it for `istream` but after compilation, apparently I was wrong.

src/miral/live_config_watcher.cpp
tarek-y-ismail: `buffer`, the source of the data in `remaining`  is already aligned explicitly a few lines above. Do we need to do more?

src/miral/override_watcher.cpp
tarek-y-ismail: I spent some time thinking about how this could be fixed before deciding to disable the lint. 

We could wrap the string views in structs so they're different types that are harder to mix up when passing to the method. But given that this method is only called in one place, I thought against that. We could wrap them in a struct, but that just moves the problem elsewhere. This is the first time I've came across this lint, so we don't have enough data to determine whether it should be disabled or not.

src/miral/override_watcher.h
tarek-y-ismail: Like my other comment, I believe `fs::path` could be misleading/confusing at least initially for such simple use cases.

tests/miral/config_file.cpp
tarek-y-ismail: I made it public because I needed to access it from the class's child `TestConfigFile`. It would work as a protected or public member, but we usually shy away from protected members. 

We can't inline `PendingLoad` since it's used in two classes. Adding an accessor would not be much better, and passing the mutex down would be incredibly verbose and annoying.

Consulting Opus, it found a suitable refactor that addresses your two points in one change. `notify_load` now takes a lambda to execute before marking a load as pending under the mutex. This allows us to make the mutex private, _and_ it merges the two locks/unlocks into just one.

tests/miral/live_config_test_helpers.h
```diff
@@ -0,0 +1,87 @@
...
+        else
+            unsetenv("XDG_CONFIG_DIRS");
+    }
+
+    XdgEnvGuard(XdgEnvGuard const&) = delete;
+    XdgEnvGuard& operator=(XdgEnvGuard const&) = delete;
+
+    std::optional<std::string> original_home;
+    std::optional<std::string> original_dirs;
+};
```
AlanGriffiths, line 65: See prior art: `TemporaryEnvironmentValue` & `TestRuntimeEnvironment`

src/miral/override_watcher.cpp
AlanGriffiths: > there should be an explanation why we are not applying the rule

src/miral/live_config_watcher.cpp
AlanGriffiths: `buffer` is aligned. I'm talking about the data accessed via `remaining`.

Your argument only applies to the first iteration. There the data has the same alignment as `buffer`'s first element. Yes the `std::array` is aligned, however, not everyone will see immediately that this implies the data is also aligned.

After the first iteration `remaining` has been reassigned with an offset of `event_size`. Your argument no longer applies.

I don't think you can assume every reader is so familiar with the layout of `std::array` and the constraints on `event_size` that this is obvious.

src/miral/config_file.cpp
AlanGriffiths: When I see a `string` I know less about its content than when I see a `path`

tests/miral/override_watcher.cpp
```diff
@@ -0,0 +1,925 @@
...
+        tmp_dir = make_tmp("test-override-watcher-home");
+        dummy_system_dir = make_tmp("test-override-watcher-sys");
+
+        base_config = tmp_dir / "test.conf";
+        override_dir = tmp_dir / "test.conf.d";
+        std::ofstream{base_config} << "base";
+
+        setenv("XDG_CONFIG_HOME", tmp_dir.c_str(), 1);
+        setenv("XDG_CONFIG_DIRS", dummy_system_dir.c_str(), 1);
+    }
```
AlanGriffiths:
```suggestion
struct TestOverrideWatcher : testing::Test
{
    auto const tmp_dir = make_tmp("test-override-watcher-home");
    auto const dummy_system_dir = make_tmp("test-override-watcher-sys");
    auto const base_config = tmp_dir / "test.conf";
    auto const override_dir = tmp_dir / "test.conf.d";

    TemporaryEnvironmentValue config_home{"XDG_CONFIG_HOME", tmp_dir.c_str()};
    TemporaryEnvironmentValue config_dirs{"XDG_CONFIG_DIRS", dummy_system_dir.c_str()};

    void SetUp() override
    {
        std::ofstream{base_config} << "base";
    }
```

tests/miral/live_config_test_helpers.h
```diff
@@ -0,0 +1,62 @@
...
+        [&](std::filesystem::path const& p, auto&&) { all_paths.push_back(p); },
+        [&](std::filesystem::path const& p) { all_paths.push_back(p); });
+    return all_paths;
+}
+
+struct XdgEnvGuard
+{
+    mir_test_framework::TemporaryEnvironmentValue home{"XDG_CONFIG_HOME", getenv("XDG_CONFIG_HOME")};
+    mir_test_framework::TemporaryEnvironmentValue dirs{"XDG_CONFIG_DIRS", getenv("XDG_CONFIG_DIRS")};
+};
```
AlanGriffiths, line 65: I don't think we need yet another way to do this

src/miral/override_watcher.cpp
```diff
@@ -0,0 +1,428 @@
...
+auto mlc::OverrideWatcher::WatchedRoot::has_base_watch() const -> bool
+{
+    return static_cast<bool>(base_dir_wd);
+}
+
+//  Wrapping the three parameters into separate structs would make the code
+//  more verbose for no benefit since this is called in only one place.
+//  Wrapping them into one struct just moves the problem to where the struct is
+//  constructed. This is fine for the time being.
+auto mlc::OverrideWatcher::WatchedRoot::classify_event(  // NOLINT(bugprone-easily-swappable-parameters)
```
AlanGriffiths: Having the `NOLINT` and corresponding explanation in two separate comments isn't easy on the reader (who is likely expecting some background on the function implementation from the first comment)

src/miral/live_config_watcher.cpp
AlanGriffiths: I guess you don't think it worth expanding on why the data has the same alignment as `std::array`?

src/miral/config_file.cpp
```diff
@@ -16,137 +16,134 @@
...
-    if (!path.has_value())
-        return std::nullopt;
+    auto const override_directory_name = resolved_base_config.filename().string() + ".d";
+
+    mlc::OverridesListBuilder config_streams;
+    config_streams.push_new(resolved_base_config, mlc::open_file);
+
+    // Collect override files from all roots, deduplicating by filename.
+    // Higher-priority roots (earlier in config_roots, later in reverse iteration) shadow lower ones.
+    std::map<path, path> by_filename;
```
AlanGriffiths: Is the iteration order is going to be dependent on current locale?

AlanGriffiths (changes requested):
Still waiting for CI results - not clear what is going on there.

Meanwhile, there's a merge conflict to resolve.

There are some unresolved comments from earlier, and some follow-up discussion here. Would be good to clarify those too.

tests/miral/override_watcher.cpp
tarek-y-ismail: Had to tweak it a bit, but it's better 👍  Thanks!

src/miral/config_file.cpp
tarek-y-ismail: Why? Unless the files we're collecting change their names depending on the locale, which I don't think happens, then we are safe.

src/miral/live_config_watcher.cpp
tarek-y-ismail: It's explained a few lines below: "the name field is padded..."

AlanGriffiths: No, you misunderstand. _If the data is aligned_, "the name field is padded..." explains why the result is also aligned.

What has not been explained is why _if `std::array` is aligned then the data it owns is aligned_.

Take another example: if you `alignas(inotify_event) std::unique_ptr<...> p`, then you wouldn't expect `*p` to be aligned would you? But with `alignas(inotify_event) std::array a` you are expecting `a.front()` to be aligned.

tarek-y-ismail: `std::array` has no indirection like `std::unique_ptr`. So aligning the array to `inotify_event` guarantees the alignment of the first element.

As mentioned before, elements are padded to the closest `inotify_event` alignment too, and this is also recorded in `inotify_event.len`. So advancing by `sizeof(inotify_event) + event.len` is perfectly safe and reads aligned data.

AlanGriffiths: What does "has no indirection" mean? That the data is held in the body of `std::array`? That isn't enough.

C++ types often have "hidden" elements in the layout (pointer to vtab for example), these can affect alignment.

And, in general, the standard doesn't dictate the layout of standard library types. That lack of specification allows, for example, a debugging implementation to allocates some guard bytes around the data, you could have:

```
{  <--------------------------------------------------- // alignment on 8 byte boundary
    [[maybe_unused]] uchar8_t canary_before[4];
    uchar8_t data[128];   <---------------------------- // NOT alignment on 8 byte boundary
    [[maybe_unused]] uchar8_t canary_after[4];
}
```

I know that the standard does, in fact, guarantee that the address of `std::array` is the address of the first element. But that isn't obvious, or even in one place in the standard.

My problem isn't that it isn't true, but reader may not know it and certainly won't know the author thought about it.

tarek-y-ismail: Hmm, I did not know that some debug implementations could put padding _before_ the first member. Anyway, I've went through the standard (draft at https://eel.is/c++draft) and tracked down the references needed to guarantee our assumptions. Might've overdone it a bit with the comment ;)

src/miral/config_file.cpp
AlanGriffiths: I guess I phrased it badly. The character set used by the filesystem is going to be mapped to some series of bytes and those bytes are what gets compared. But the mapping is potentially unintuitive (and not necessarily a SBCS)

tarek-y-ismail: That was an, uuuh, interesting rabbit hole.

Here's the gist: systemd doesn't care about locale as in `LC_LOCALE`. nor does it give much thought to what language the filenames are written in. It mostly uses the number prefixes for priority, and if those match, then it uses byte comparison for the remainder of the string. Other projects also follow this convention. 

For our code, we were using byte comparison for everything, and not comparing number prefixes correctly. Nothing else in the code base seems to have hit this problem before, and nothing in the standard library handles this. GNU libc has `strverscmp`, but I don't think we want to use that since it's an extension. I went on and had Copilot implement a draft, then I went and modernized it. We implemented some tests along the way as well. Hopefully this addresses your concerns :)

src/miral/live_config_watcher.cpp
```diff
@@ -0,0 +1,260 @@
...
+    // address (no padding at the beginning). data() returns addressof(front()),
+    // which is that same address. Therefore, alignas(inotify_event) guarantees
+    // buffer.data() is aligned to alignof(inotify_event).
+    // References from the C++ standard (https://eel.is/c++draft):
+    // - [array.overview]/2: std::array is an aggregate (and thus standard-layout)
+    // - [class.mem.general]/31: a standard-layout object's address equals its
+    //   first non-static data member's address (note 11: no padding at beginning)
+    // - [array.members]/2: data() == addressof(front()) for non-empty arrays
+    // - [sequence.reqmts]/73: front() returns *begin() (the first element)
+    alignas(inotify_event) std::array<std::byte, sizeof(inotify_event) + NAME_MAX + 1> buffer;
```
AlanGriffiths: I guess we need to disable cppcoreguidelines-pro-type-member-init

src/miral/live_config_watcher.cpp
AlanGriffiths: Definitely overdone. But OK. (It isn't just "padding before the first member" - in most cases the standard says nothing about what members there are or their order.

src/miral/live_config_watcher.cpp
tarek-y-ismail: Worked around by zero-initializing the buffer. Shouldn't be too much overhead for a few hundred bytes

src/miral/live_config_watcher.cpp
```diff
@@ -127,7 +127,7 @@ void mlc::Watcher::for_each_inotify_event(std::function<void(inotify_event const
     //   first non-static data member's address (note 11: no padding at beginning)
     // - [array.members]/2: data() == addressof(front()) for non-empty arrays
     // - [sequence.reqmts]/73: front() returns *begin() (the first element)
-    alignas(inotify_event) std::array<std::byte, sizeof(inotify_event) + NAME_MAX + 1> buffer;
+    alignas(inotify_event) std::array<std::byte, sizeof(inotify_event) + NAME_MAX + 1> buffer = {};
```
AlanGriffiths: Do we need copy initialization? Or any initialization? We're going to read into the buffer immediately.

If we do need initialization, why not list initialization? (It is usually the safe choice!)

tarek-y-ismail: I was torn between the two. One one had, initialization makes things explicit, on the other, we read right after and can just slap a `NOLINT` on it (though this case is one of  more justified ones). But since I've got an opinion nudging me the other way...

AlanGriffiths (approved):
Let's land this

PR #4786, opened 2026-03-20
sbuild: move wlcs run to autopkgtest

Saviq:
## What's new?

Rather than running WLCS during the build, run it in autopkgtest, testing the stack, rather than the built artifacts.

## How to test

CI

## Checklist

- [x] Tests added and pass


Saviq:
> Neat! Why do we need wlcs >= 1.8? Should we comment that?

Just aligned with `debian/control`, is all.

PR #4787, opened 2026-03-21
Introduce safe mode when there is a problem with configuration

### LINKED ISSUE #3637: Introduce a Safe Mode when there was a problem with configuration
> We could also use the diagnostic feature to (for a time?) signify a problem to whoever's at the display.

That I like (for Frame). That could be generalised to any startup/initialization error reporting: we would just need to reexec with a "safe" config and error text.

_Originally posted by @AlanGriffiths in https://github.com/canonical/mir/issues/3628#issuecomment-2416145905_

deepakshirkem:
Closes #3637 

Related #3628 

## What's new?
When Mir encounters a configuration error (mir::AbnormalExit) during startup, instead of exiting with failure, it now re-execs itself with minimal configuration (safe mode)

## How to test

Run mir_demo_server with an invalid option and observe it re-execs into safe mode instead of dying:

`./bin/mir_demo_server --invalid-option=garbage`

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/examples/mir_demo_server/server_example.cpp
+++ b/examples/mir_demo_server/server_example.cpp
@@ -131,6 +131,7 @@ try
 
     miral::live_config::IniFile config_store;
     runner.set_exception_handler(exception_handler);
+    runner.enable_safe_mode();
 
     miral::CursorScale cursor_scale{config_store};
     miral::OutputFilter output_filter{config_store};
--- a/include/miral/miral/runner.h
+++ b/include/miral/miral/runner.h
@@ -161,6 +161,15 @@ public:
     /// \sa miral::X11Support - provides X11 support to a Mir server
     auto x11_display() const -> mir::optional_value<std::string>;
 
+    /// Enable safe mode for this compositor.
+    ///
+    /// This is intended for use in embedded compositors (e.g. Ubuntu Frame)
+    /// where there is no terminal to display errors and the compositor must
+    /// start in order to show the error to the user.
+    ///
+    /// Safe mode is disabled by default.
+    void enable_safe_mode();
+
 private:
     MirRunner(MirRunner const&) = delete;
     MirRunner& operator=(MirRunner const&) = delete;
--- a/src/miral/runner.cpp
+++ b/src/miral/runner.cpp
@@ -26,6 +26,7 @@
 #include <mir/main_loop.h>
 #include <mir/report_exception.h>
 #include <mir/options/option.h>
+#include <boost/program_options/errors.hpp>
 
 #include <algorithm>
 #include <mutex>
@@ -60,6 +61,7 @@ struct miral::MirRunner::Self
     std::string const config_file;
     std::string const display_config_file;
     bool const safe_mode{getenv("MIR_SAFE_MODE") != nullptr};
+    bool safe_mode_enabled{false};
 
     std::mutex mutex;
     std::function<void()> start_callback{[]{}};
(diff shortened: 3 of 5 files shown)


deepakshirkem:
Hi @AlanGriffiths,
Thank you for your feedback. I think your approach makes more sense. Please let me know if I am heading in the right direction. For the API, would you prefer `MirRunner::enable_safe_mode()` or a callback-based approach?
Also, for the .config file errors, should I handle those in the same apply_settings() catch block, or is there existing code that I should check?

Thank You::)

AlanGriffiths:
> would you prefer `MirRunner::enable_safe_mode()` or a callback-based approach?

I've not thought about how this should be implemented. But I'd actually explore a different approach first: Trying to do this from the downstream side of the API...

* `MirRunner::MirRunner()` has an overload that allows the config file to be specified. So, if we are `reexec`ing, we can use a "safe" configuration. (Obviously, environment and command-line need trimming too.)
* `MirRunner::set_exception_handler()` allows a compositor (e.g. Ubuntu Frame) to handle exceptions and, therefore, to pick out `AbnormalExit` as you do in the current proposal, but in the downstream project.

In the case of Ubuntu Frame, the content of the `AbnormalExit` should be available for the "diagnostic" screen. E.g. by writing to the underlying file before re-exec. Also, Frame should prevent any external client attaching (to ensure the diagnostic is visible).

None of that needs obvious changes to Mir - and doing in Frame allows Frame's specific needs to be addressed.

Saviq:
FWIW I think this may have a place outside of Frame - a desktop that doesn't die, but falls back to a safe mode instead is a worthy goal, too, IMO.

deepakshirkem:
Hi @AlanGriffiths and @Saviq, Thank you for the feedback. Would a solution like this satisfy both concerns?
I am planning to add `MirRunner::enable_safe_mode() `as an opt-in API in Mir. When enabled, it would catch AbnormalExit from all config sources (CLI, .config file, and environment variables), then re-execute using a safe config file via the existing MirRunner(argc, argv, config_file) overload. It would also expose the error message so downstream compositors like Frame can display it.

Does this approach make sense?

AlanGriffiths:
> FWIW I think this may have a place outside of Frame - a desktop that doesn't die

Perhaps: I'm not sure that it is the compositor that should be orchestrating that for a desktop environment. In any case, I think Mir already provides all the features needed.

deepakshirkem:
Hi @AlanGriffiths and @Saviq, Could you please check if this solutions is going in the right direction? I have added the `enable_safe_mode()` opt-in API and I am now heading towards implementing the visual overlay so user can see the error on screen.
I have a questions regarding the overlay do i need to create new screen for this , or there is something already available in the codebase that i can reuse? 

Thank You:)

AlanGriffiths:
> Hi @AlanGriffiths and @Saviq, Could you please check if this solutions is going in the right direction? I have added the `enable_safe_mode()` opt-in API 

Hi @deepakshirkem, I don't see anything in recent changes that deals with the issues mentioned above: [This does not do what it sets out to do](https://github.com/canonical/mir/pull/4787#pullrequestreview-3991576311).

And, as I said above, [I don't think there's any need for a new API to support this feature](https://github.com/canonical/mir/pull/4787#issuecomment-4111167595). But if I've overlooked something and a new API is needed, then it needs to support the downstream compositor supplying an appropriate safe configuration in the reexec.

> and I am now heading towards implementing the visual overlay so user can see the error on screen. I have a questions regarding the overlay do i need to create new screen for this , or there is something already available in the codebase that i can reuse?

The mechanism for reporting "safe mode" likely depends on the downstream compositor. In the case of Ubuntu Frame, there's already a "diagnostic message" feature that could be used.

deepakshirkem:
Hi @AlanGriffiths, My apologies for the confusion. After investigating the Ubuntu Frame codebase, I found that this feature can be implemented entirely in Ubuntu Frame using existing Mir APIs — no changes to Mir core are needed.
Ubuntu Frame now shows the error on screen instead of dying silently.
Before: Frame dies with a black screen and the user sees nothing useful.
After:
Screenshot from 2026-03-24 20-28-16

I will open a PR against Ubuntu Frame for this change. Should I close this Mir PR?

AlanGriffiths:
> I will open a PR against Ubuntu Frame for this change. Should I close this Mir PR?

:+1:

deepakshirkem:
Hi @AlanGriffiths ,
I  have opened a PR against Ubuntu Frame instead: https://github.com/canonical/ubuntu-frame/pull/295

Closing this PR.

Thank You ((::

PR #4793, opened 2026-03-23
use std::format* instead of snprintf in xwayland_spawner.cpp

### LINKED ISSUE #4660: Use std::Format instead of char buffers

yotam-medini:
May close #4660

Related: #4736

## What's new?
use std::format or std::format_to_n instead of snprintf in xwayland_spawner.cpp

Save generated paths as XWaylandSpawner data members.
Introduce private XWaylandSpawner::XDisplayPaths struct.
Delegate public XWaylandSpawner ctor to private ctor via XDisplayPaths.
create_lockfile(), choose_display() are now static class methods.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_xwayland/xwayland_spawner.cpp
+++ b/src/server/frontend_xwayland/xwayland_spawner.cpp
@@ -28,6 +28,7 @@
 
 #include <array>
 #include <format>
+#include <iterator>
 #include <string>
 #include <fcntl.h>
 #include <sys/socket.h>
@@ -101,7 +102,8 @@ auto create_sockets(const std::string &x11_socket) -> std::vector<mir::Fd>
     std::vector<mir::Fd> result;
     struct sockaddr_un addr{ .sun_family = AF_UNIX, .sun_path = { 0 } };
 
-    auto name_create_socket = [&](char *bos, size_t limit)
+    auto name_create_socket = [&](
+        char *bos, std::iter_difference_t<char *> limit)
     {
         auto eos = std::format_to_n(bos, limit, "{}", x11_socket).out;
         *eos = '\0';


src/server/frontend_xwayland/xwayland_spawner.cpp
```diff
@@ -36,17 +39,14 @@ namespace md = mir::dispatch;
 
 namespace
 {
-auto const x11_lock_fmt = "/tmp/.X%d-lock";
-auto const x11_socket_fmt = "/tmp/.X11-unix/X%d";
+constexpr std::string_view x11_lock_fmt = "/tmp/.X{}-lock";
+constexpr std::string_view x11_socket_fmt = "/tmp/.X11-unix/X{}";
```
robert-ancell: This seems more complicated than the original code - why change it?

src/server/frontend_xwayland/xwayland_spawner.cpp
```diff
@@ -139,13 +138,15 @@ auto create_sockets(int xdisplay) -> std::vector<mir::Fd>
...
     std::vector<mir::Fd> result;
     struct sockaddr_un addr{ .sun_family = AF_UNIX, .sun_path = { 0 } };
-    size_t path_size;
-
-    path_size = snprintf(addr.sun_path + 1, sizeof(addr.sun_path) - 1, x11_socket_fmt, xdisplay);
-    create_socket(result, &addr, path_size);
 
-    path_size = snprintf(addr.sun_path, sizeof(addr.sun_path), x11_socket_fmt, xdisplay);
-    create_socket(result, &addr, path_size);
+    for (size_t skip: {1, 0})
```
robert-ancell: The loop is making this more confusing - I'd keep this with two lines. Also I think we can simplify the first case with a new format string, i.e.:
```c
auto const x11_abstract_socket_fmt = "\0/tmp/.X11-unix/X{}";
auto end = std::format_to_n(addr.sun_path, sizeof(addr.sun_path), x11_abstract_socket_fmt, xdisplay).out;
create_socket(result, &addr, end - addr.sun_path);

auto const x11_socket_fmt = "/tmp/.X11-unix/X{}";
auto end = std::format_to_n(addr.sun_path, sizeof(addr.sun_path), x11_socket_fmt, xdisplay).out;
create_socket(result, &addr, end - addr.sun_path);
```

src/server/frontend_xwayland/xwayland_spawner.cpp
```diff
@@ -194,11 +195,15 @@ mf::XWaylandSpawner::XWaylandSpawner(std::function<void()> spawn)
...
-    unlink(path);
+    std::array<char, 256> buffer; // avoid allocation in noexcept code
+    auto eos = std::format_to_n(buffer.data(), buffer.size() - 1,
+        x11_lock_fmt, xdisplay).out;
+    *eos = '\0';
+    unlink(buffer.data());
+    eos = std::format_to_n(buffer.data(), buffer.size() - 1,
+        x11_socket_fmt, xdisplay).out;
+    *eos = '\0';
+    unlink(buffer.data());
```
robert-ancell: Might this be simpler just to store the paths in the class to avoid having to regenerate them? If not, then this code for generating them should be shared, i.e. `mf::XWaylandSpawner::create_x11_lock_path(xdisplay)`.

yotam-medini: In the ~XWaylandSpawner destructor, we need to avoid dynamic allocations,
thus we should not use std::format(...) there.

The paths depend runtime-wise only on (int) xdisplay.
Storing them as data members of XWaylandSpawner (along with xdisplay)
seems unnecessary, but may simplify code.

Currently the first generation of the paths occur outside the class context,
in choose_display() → create_lockfile().

We can introduce 
```
struct DisplayPaths
{
    int xdisplay{-1};
    std::string x11_lock_path;
    std::string x11_socket_path;
};
```
which will:
1. be a data member of XWaylandSpawner replacing the simple 'int xdisplay',
2. have choose_display() return this struct, to be used in XWaylandSpawner initialization.

But isn't this an overkill ?

robert-ancell: If you pass those strings into `create_lockfile()` then you wont need the struct. Or maybe these should become methods instead of functions. It's always a bad sign when you're doing something twice and a risk that it breaks in the future due to only being changed in one place.

yotam-medini: Passing to method will require moving most (or all) of the initialization list into the constructor body.

robert-ancell:
@Saviq the CI seems to be stalled - any ideas why?

Saviq:
> @Saviq the CI seems to be stalled - any ideas why?

That'd be https://github.com/canonical/mir/commit/136cefd7193aade370d7410a52f8e5aa76eb51ca…

But why is TiCS fine being skipped, and the others, not?

src/server/frontend_xwayland/xwayland_spawner.cpp
```diff
@@ -135,17 +96,19 @@ auto create_socket(std::vector<mir::Fd>& fds, struct sockaddr_un *addr, size_t p
...
-    size_t path_size;
-
-    path_size = snprintf(addr.sun_path + 1, sizeof(addr.sun_path) - 1, x11_socket_fmt, xdisplay);
-    create_socket(result, &addr, path_size);
 
-    path_size = snprintf(addr.sun_path, sizeof(addr.sun_path), x11_socket_fmt, xdisplay);
-    create_socket(result, &addr, path_size);
+    auto name_create_socket = [&](char *bos, size_t limit)
+    {
+        auto eos = std::format_to_n(bos, limit, "{}", x11_socket).out;
```
Saviq, line 108: TiCS has this to say:

> narrowing conversion from 'size_t' (aka 'unsigned long') to signed type 'iter_difference_t<char *>' (aka 'long') is implementation-defined.

yotam-medini: [How] can I run this TiCS locally ?

Saviq: Sorry, you can't, but that actually comes from `clang-tidy`. You should be able to reproduce this with the following CMake options:

```shell
-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_CLANG_TIDY=clang-tidy
```

yotam-medini: After some struggling with cmake, looking at output from: 'ps laxww' I got to the attached:
[tidy.sh](https://github.com/user-attachments/files/26386101/tidy.sh)
Running it re-produced the error:
```
$ /tmp/tidy.sh
24 warnings generated.
/home/yotam/pub/mir/src/server/frontend_xwayland/xwayland_spawner.cpp:106:42: warning: narrowing conversion from 'size_t' (aka 'unsigned long') to signed type 'iter_difference_t<char *>' (aka 'long') is implementation-defined [cppcoreguidelines-narrowing-conversions]
  106 |         auto eos = std::format_to_n(bos, limit, "{}", x11_socket).out;
      |                                          ^
```

The commit:  
91ca9480 yotam.m  iter_difference_t instead of size_t for format_to_n
fixes it.

PR #4798, opened 2026-03-25
Add tests for skipping bad values in int and float arrays

tarek-y-ismail:
## How to test

- Adds two small tests to make sure that bad values being inserted into arrays are ignored.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/tests/miral/live_config_ini_file.cpp
+++ b/tests/miral/live_config_ini_file.cpp
@@ -294,6 +294,34 @@ TEST_F(LiveConfigIniFile, a_bad_bool_value_is_ignored)
     ini_file.load_file(bad_istream, fake_filename());
 }
 
+TEST_F(LiveConfigIniFile, a_bad_integer_value_in_an_array_is_skipped)
+{
+    std::istringstream bad_istream{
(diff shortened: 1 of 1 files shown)


AlanGriffiths:
>     * Adds two small tests to make sure that bad values being inserted into arrays are ignored.

I know it is the implemented behavior. But... is this actually the desired behavior? Should we be rejecting the whole key?

tarek-y-ismail:
> Should we be rejecting the whole key?

What do you mean by that? Do you mean marking the whole thing as invalid or maybe throwing an error?

AlanGriffiths:
> > Should we be rejecting the whole key?
> 
> What do you mean by that? Do you mean marking the whole thing as invalid or maybe throwing an error?

Throwing an error seems wrong for parsing external input that we can reasonably expect to have issues. Unless that is caught in the file reload handler and simply rejects the whole file. (But if we're doing that, then it is probably better to continue parsing the file to identify further errors, but not to apply ANY new settings - which doesn't need an exception.)

Rejecting the whole key would mean the system can continue with the current setting, not get a new one with missing entries. (Without real world examples it is hard to know which is the best approach.)

PR #4805, opened 2026-03-30
Adding a method to set the cursor theme at runtime

mattkae:
Closes #???

Related: #???, https://..., ...

## Checklist

- [ ] Tests added and pass


mattkae:
Now that I have full understanding, cursors are SO complex that they should not be set dynamically, as nothing else will update accordingly. A logout and login flow is expected, unfortunately.

PR #4812, opened 2026-03-31
fix(mousekeys): improve Ctrl+Shift+Numlock toggle shortcut

### LINKED ISSUE #3902: Improve mousekeys toggle shortcut
Mousekeys can be toggled by pressing `ctrl-shift-numlock` twice (or just `ctrl-shift-numlock-numlock). This is confusing for a number of reasons:
- Mousekeys only works when numlock is on, so if you start with the numlock turned off and toggle, you'll have to turn on numlock afterwards to get mousekeys working.
- Having to press numlock twice can be a tad confusing, especially since it has its own state.

So for the sake of UX, we should use a more intuitive shortcut. Some suggestions:
- `ctrl-shift-numlock on`: Clearer in my opinion, ensures that numlock is on so that users can use mousekeys right away. `ctrl-shift-numlock off` could also be used to turn off mousekeys.
- Use a key other than numlock, keep the double tap: Unlikely to be used in applications. More intuitive than numlock.

deepakshirkem:
Closes #3902 

## What's new?

Changed mousekeys toggle shortcut from Ctrl+Shift+Numlock to Ctrl+Shift+F12 to avoid Numlock state conflicts
Simplified toggle logic using a clean boolean flag Use initial_mousekeys_state to initialize mousekeys_on

## How to test

1. Run miral-shell
2. Press Ctrl+Shift+F12 to enable mousekeys (on laptops: Ctrl+Shift+Fn+F12)
3. Press numpad 2 4 6 8 — pointer should move
4. Press Ctrl+Shift+F12 again to disable mousekeys
5. Press numpad keys — pointer should no longer move

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -155,7 +155,7 @@ int main(int argc, char const* argv[])
         auto const keysym = mir_keyboard_event_keysym(key_event);
         auto const action = mir_keyboard_event_action(key_event);
 
-        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) && (keysym == XKB_KEY_F12)) 
+        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) && (keysym == XKB_KEY_F12))
         {
             if (action != mir_keyboard_action_down)
                 return true;


examples/miral-shell/shell_main.cpp
```diff
@@ -13,7 +13,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-
+#include <mir/log.h>
```
tarek-y-ismail, line 16: We are not logging anywhere here, so this is not needed. We prefer to minimize includes to reduce compilation times and minimize diffs in the future.

examples/miral-shell/shell_main.cpp
```diff
@@ -150,25 +150,34 @@ int main(int argc, char const* argv[])
         return initial_mousekeys_state ? MouseKeysConfig::enabled() : MouseKeysConfig::disabled();
     }();
 
-    auto toggle_mousekeys_filter = [mousekeys_on = initial_mousekeys_state, &mousekeys_config](MirKeyboardEvent const* key_event) mutable {
+    auto toggle_mousekeys_filter = [mousekeys_on = false, &mousekeys_config](MirKeyboardEvent const* key_event) mutable {
```
tarek-y-ismail, line 153: Having two separate states might cause subtle bugs when someone tries modifying the code.
```suggestion
    auto toggle_mousekeys_filter = [mousekeys_on = initial_mousekeys_state, &mousekeys_config](MirKeyboardEvent const* key_event) mutable {
```

examples/miral-shell/shell_main.cpp
```diff
@@ -150,25 +150,34 @@ int main(int argc, char const* argv[])
...
+        using namespace miral::toolkit;
         auto const modifiers = mir_keyboard_event_modifiers(key_event);
+        auto const keysym = mir_keyboard_event_keysym(key_event);
+        auto const action = mir_keyboard_event_action(key_event);
+
+        if (mousekeys_on && keysym == XKB_KEY_Num_Lock)
+            return true;
```
tarek-y-ismail, line 160: Wouldn't this interfere with disabling mousekeys once it's enabled? Besides, we already consume the event by returning `true` in the if statement below, so this isn't necessarily needed.

examples/miral-shell/shell_main.cpp
```diff
@@ -150,25 +150,34 @@ int main(int argc, char const* argv[])
...
+            if (mousekeys_on)
+            {
                 mousekeys_config.enable();
+            }
             else
+            {
                 mousekeys_config.disable();
-
+            }
+
```
tarek-y-ismail, line 178: For single/simple statements, we usually don't wrap them in braces under if statements. Generally, we try sticking to the existing code style if the PR is not specifically for fixing the code style.

examples/miral-shell/shell_main.cpp
```diff
@@ -150,25 +150,34 @@ int main(int argc, char const* argv[])
...
+            }
             else
+            {
                 mousekeys_config.disable();
-
+            }
+            
             return true;
         }
-
```
tarek-y-ismail, line 171: Unrelated edit.

examples/miral-shell/shell_main.cpp
```diff
@@ -150,25 +150,34 @@ int main(int argc, char const* argv[])
...
         auto const modifiers = mir_keyboard_event_modifiers(key_event);
+        auto const keysym = mir_keyboard_event_keysym(key_event);
+        auto const action = mir_keyboard_event_action(key_event);
+
+        if (mousekeys_on && keysym == XKB_KEY_Num_Lock)
+            return true;
 
-        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) &&
-            (modifiers & mir_input_event_modifier_num_lock))
+        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) && (keysym == XKB_KEY_Num_Lock))
```
tarek-y-ismail, line 162: In hindsight, my original choice of `Ctrl + Shift + Numlock` was probably not the best. Mousekeys already depends on numlock being on for numpad keysyms to be emitted. And since the numlock key event is handled before it reaches the keyboard filter, our consumption of the event only hides it from whatever client is focused, it doesn't prevent numlock from being toggled.

This results in numlock toggling in tandem with mousekeys. For example, if numlock is initially enabled and mousekeys is not, and you press `Ctrl + Shift + Numlock` to enable mousekeys, mouse keys will be enabled, but numlock will be disabled. This will then require you to press numlock again to be able to use mousekeys. Generally, I see this as poor UX, and think we should probably change this shortcut to something else. 

I suggest `Ctrl + Shift + F12`, what do you think?

examples/miral-shell/shell_main.cpp
tarek-y-ismail: I just tried it and it indeed prevents mousekeys from being disabled.

tarek-y-ismail:
One more thing, adding testing instructions would be nice to make sure we're all testing the same thing. For me, I:

- With numlock disabled Press `Ctrl + Shift + Numlock`
- Press numpad 2, 4, 6, 8 and observe the pointer moving
- Press `Ctrl + Shift + Numlock` again, and observe mousekeys not disabling (double-checked by adding some logging statements)

deepakshirkem:
Hi @tarek-y-ismail, Comments has been addressed in the latest commit. I have switched the shortcut to `Ctrl+Shift+F12` as you suggested, which avoids the Numlock state issue entirely. The toggle now works correctly press once to enable, press again to disable.
I have also added testing instructions to the PR description.

Thank You ::))

examples/miral-shell/shell_main.cpp
```diff
@@ -13,7 +13,6 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-
```
tarek-y-ismail, line 16: Unrelated edit

examples/miral-shell/shell_main.cpp
```diff
@@ -152,23 +151,21 @@ int main(int argc, char const* argv[])
...
-            (modifiers & mir_input_event_modifier_num_lock))
-        {
-            if (mir_keyboard_event_action(key_event) == mir_keyboard_action_down ||
-                mir_keyboard_event_action(key_event) == mir_keyboard_action_repeat)
+        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) &&(keysym == XKB_KEY_F12)) {
+            if (action != mir_keyboard_action_down)
                 return true;
 
             mousekeys_on = !mousekeys_on;
-            if(mousekeys_on)
```
tarek-y-ismail, line 164: Unrelated edit

examples/miral-shell/shell_main.cpp
```diff
@@ -152,23 +151,21 @@ int main(int argc, char const* argv[])
...
         auto const modifiers = mir_keyboard_event_modifiers(key_event);
+        auto const keysym = mir_keyboard_event_keysym(key_event);
+        auto const action = mir_keyboard_event_action(key_event);
 
-        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) &&
-            (modifiers & mir_input_event_modifier_num_lock))
-        {
-            if (mir_keyboard_event_action(key_event) == mir_keyboard_action_down ||
-                mir_keyboard_event_action(key_event) == mir_keyboard_action_repeat)
+        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) &&(keysym == XKB_KEY_F12)) {
```
tarek-y-ismail:
```suggestion
        if ((modifiers & mir_input_event_modifier_ctrl) && (modifiers & mir_input_event_modifier_shift) && (keysym == XKB_KEY_F12)) {
```

deepakshirkem:
Hi @tarek-y-ismail, I have addressed all the review comments. Please review when you get a chance.

Thank you ::))

PR #4813, opened 2026-04-01
Eliminate tmpnam

yotam-medini:
Closes nothing

Related: #???, https://..., ...

## What's new?

Use mkstemp and mkdtemp instead of tmpnam.
Using tmpnam results with warning such as:
/bin/ld: CMakeFiles/mir_unit_tests.dir/frontend_wayland/test_desktop_file_manager.cpp.o: in function `DesktopFileManager_can_resolve_from_valid_flatpak_info_Test::TestBody()':
/home/yotam/pub/mir/tests/unit-tests/frontend_wayland/test_desktop_file_manager.cpp:197:(.text+0x1f96): warning: the use of `tmpnam' is dangerous, better use `mkstemp'

While at it, remove temporary files and better clean temporary directory at end of test.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
+++ b/tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
@@ -34,6 +34,7 @@ using ::testing::NotNull;
 
 namespace
 {
+static fs::path desktop_file_directory_path;
 static std::string applications_dir;
 }
 
@@ -48,9 +49,15 @@ struct GDesktopFileCache : Test
         auto desktop_file_directory_name = fs_path.string();
         auto path_ptr = ::mkdtemp(desktop_file_directory_name.data());
         ASSERT_THAT(path_ptr, NotNull());
+        desktop_file_directory_path = fs::path(path_ptr);
         setenv("XDG_DATA_DIRS", desktop_file_directory_name.c_str(), 1);
         applications_dir = desktop_file_directory_name + "/applications";
         mkdir(applications_dir.c_str(), S_IRWXU);
+    }
+
+    static void TearDownTestSuite()
+    {
+        fs::remove_all(desktop_file_directory_path);
     }
 
     void TearDown() override


tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
```diff
@@ -42,10 +43,11 @@ struct GDesktopFileCache : Test
...
     static void SetUpTestSuite()
     {
-        // Establish the temporary directory
-        std::filesystem::path tmp_dir_path {std::filesystem::temp_directory_path() /= std::tmpnam(nullptr)};
-        std::filesystem::create_directories(tmp_dir_path);
-        std::string desktop_file_directory_name = tmp_dir_path;
+        std::filesystem::path fs_path {
+            fs::temp_directory_path() /= "mir_test_gdtp_fcache_XXXXXX"};
+        auto desktop_file_directory_name = fs_path.string();
+        auto path_ptr = ::mkdtemp(desktop_file_directory_name.data());
```
AlanGriffiths, line 50: This "XXXXXX" is never substituted. That means concurrent test runs will share a directory and could step on each other.

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
```diff
@@ -54,8 +56,8 @@ struct GDesktopFileCache : Test
     void TearDown() override
     {
         // Remove all files from the temporary directory
-        for (const auto& entry : std::filesystem::directory_iterator(applications_dir))
-            std::filesystem::remove_all(entry.path());
+        for (const auto& entry : fs::directory_iterator(applications_dir))
+            fs::remove_all(entry.path());
```
AlanGriffiths, line 67: `fs::remove_all(fs_path);`?

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
yotam-medini: The following test, shows that "XXXXXX" is well and uniquely substituted via mkdtemp.
[make_tempdir.cpp](https://github.com/user-attachments/files/26413015/make_tempdir.cpp)
$ cat make_tempdir.cpp; g++ -Wall -g --std=c++20 -o /tmp/make_tempdir make_tempdir.cpp; /tmp/make_tempdir 3
```
#include <filesystem>
#include <format>
#include <string>

namespace fs = std::filesystem;

std::string make_tempdir() {
    std::filesystem::path fs_path {
        fs::temp_directory_path() /= "mir_test_gdtp_fcache_XXXXXX"};
    auto desktop_file_directory_name = fs_path.string();
    auto path_ptr = ::mkdtemp(desktop_file_directory_name.data());
    return path_ptr;
}

#include <iostream>
#include <vector>
#include <cstdlib>
int main(int argc, char **argv) {
    size_t n_calls = argc > 1 ? std::stod(argv[1]) : 1;
    std::vector<std::string> dir_names;
    while (dir_names.size() < n_calls) {
        auto const dn = make_tempdir();
        std::cout << std::format("{}: dn={}\n", dir_names.size(), dn);
        dir_names.push_back(dn);
    }
    for (auto const &dn: dir_names) {
        fs::remove_all(dn);
    }
    return 0;
}
```
0: dn=/tmp/mir_test_gdtp_fcache_RvKQzD
1: dn=/tmp/mir_test_gdtp_fcache_Vs3Cvo
2: dn=/tmp/mir_test_gdtp_fcache_4ZBI1e

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
yotam-medini: This function works fine. This PR does not change the code it produces here. It just uses the namespace fs:: abbreviation.
In TearDown we do not want to remove the new top tmp directory. But clean just the contents of the applications_dir, so next tests of the suite can still use it.

But your comment encouraged me to add TearDownTestSuite() that cleans the directory created by mkdtemp.

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
AlanGriffiths: Yes, sorry

tests/unit-tests/frontend_wayland/test_g_desktop_file_cache.cpp
```diff
@@ -33,6 +34,7 @@ using ::testing::NotNull;
 
 namespace
 {
+static fs::path desktop_file_directory_path;
 static std::string applications_dir;
 }
```
AlanGriffiths, line 39: Pre-existing, so not blocking:

1. There's no need for both the anonymous namespace and `static`
2. These variables would be better at class scope than at namespace scope

PR #4820, opened 2026-04-01
Fix TICS narrowing conversion int to float violation

### LINKED ISSUE #4713: Resolve TICS violations in `seat_input_device_tracker.cpp`
Ref https://github.com/canonical/mir/pull/4410#issuecomment-3936774672

* narrowing conversion from 'typename std::enable_if<std::is_integral<int>::value, int>::type' (aka 'int') to 'float
https://github.com/canonical/mir/blob/c0c8a0a0f148c50655b6a9a436338e4bb2ffe284/src/server/input/seat_input_device_tracker.cpp#L243-L244

DigraJatin:
Closes #4713

Instead of .as_int() use round_to<float>() that uses static_cast<int>() internally.


AlanGriffiths (approved):
Yup

PR #4799, opened 2026-03-25
Live config ini file changes

tarek-y-ismail:
Related: #4781 

## What's new?

- Extends `IniFile` to read from multiple files.

## How to test

```
miral-test --gtest_filter=LiveConfigIniFile*
```

## Checklist

- [x] Tests added and pass

PR #4830, opened 2026-04-06
Add `ConfigAggregator` for merging config state across multiple files

tarek-y-ismail:
Related: #4799  

`ConfigAggregator` is a new `Store` implementation that merges configuration from multiple files into a single consistent view. It takes a `ParsingStore` (e.g. an `IniFile`) and a list of streams, parses each in order, accumulates all values, and dispatches handlers exactly once with the final merged result.

This is the intended way to support layered configuration — a base file plus a drop-in directory of overrides — without callers having to manage accumulation themselves.

## What's new

- New `ConfigAggregator` class implementing `Store`, which takes a `ParsingStore` and a list of config streams via `load_all()`. Later files take precedence for scalar keys; array keys accumulate across files and can be cleared with an empty assignment.
- New `ParsingStore` abstract interface extending `Store` with a `load_file()` method, as a common base for format-specific parsers. `IniFile` now implements `ParsingStore`.
- `BasicStore` extracted from `IniFile` to provide the reusable attribute registration, typed storage, and dispatch logic that `ConfigAggregator` builds on.

## How to test

- Existing `IniFile` tests continue to pass unchanged.
- New `ConfigAggregator` tests cover scalar and array attribute handling, multi-file merging and override precedence, preset values, empty-value array clearing, done handlers, and error cases.

## Checklist

- [x] Tests added and pass


tests/miral/config_aggregator.cpp
```diff
@@ -610,3 +610,115 @@ TEST_F(ConfigAggregatorTest, override_wins_for_shared_key_after_base_update)
...
+              {std::ref(stream2), "base.conf.d/10-override.conf"}});
+}
+
+TEST_F(ConfigAggregatorTest, invalid_scalar_value_in_one_of_multiple_files_yields_nullopt_key_assigned_twice)
+{
+    aggregator.add_int_attribute(a_key, "a scoped int", [this](auto... args) { int_handler(args...); });
+
+    // stream1 assigns the key twice with valid values; stream2 (higher priority) assigns the key twice, both unparseable
+    std::istringstream stream1{a_key.to_string() + "=10\n" + a_key.to_string() + "=20\n"};
+    std::istringstream stream2{a_key.to_string() + "=30\n" + a_key.to_string() + "=also_not_a_number\n"};
```
Saviq: Comment disagrees with fixture - only one value is unparseable?

tests/miral/config_aggregator.cpp
```diff
@@ -610,3 +610,115 @@ TEST_F(ConfigAggregatorTest, override_wins_for_shared_key_after_base_update)
...
+{
+    aggregator.add_int_attribute(a_key, "a scoped int", [this](auto... args) { int_handler(args...); });
+
+    // stream1 assigns the key twice with valid values; stream2 (higher priority) assigns the key three times:
+    // one valid, one invalid, then another valid value
+    std::istringstream stream1{a_key.to_string() + "=10\n" + a_key.to_string() + "=20\n"};
+    std::istringstream stream2{a_key.to_string() + "=30\n" + a_key.to_string() + "=also_not_a_number\n" + a_key.to_string() + "=40\n"};
+
+    // The last valid assignment in the highest-priority file wins for scalars
+    EXPECT_CALL(*this, int_handler(a_key, Optional(40)));
```
Saviq, line 657: This, and below, is what we need to make a policy call on, but IMO this is right: we accept valid values, ignore invalid ones _in the drop-in_.

@AlanGriffiths?

Saviq (changes requested):
There's a lot of comments saying "wins… holds invalid value… nullopt", but test expects that not to be true?

Do we have a test for the main file providing invalid values - and that baiing, as it does today?

I would also like a test with:
```
# base.conf
array=foo
# base.conf.d/10-override.conf
array=bar
{ "invalid": "stuff" }
array=baz
```

This should keep `[foo]`.

tarek-y-ismail:
> Do we have a test for the main file providing invalid values - and that baiing, as it does today?

Isn't the base config being valid a core assumption? With overrides, users don't have do touch that, so it shouldn't get corrupted anyway. 

> 
> I would also like a test with:
> ```
> # base.conf
> array=foo
> # base.conf.d/10-override.conf
> array=bar
> { "invalid": "stuff" }
> array=baz
> ```
> This should keep [foo].

We do our parsing line by line, so in this case it will go like so:
```
# base.conf
# `aggregator.array = []`
array=foo  # base.array = [foo]
# End of base config, aggregator notified of `array += [foo]`. `aggregator.array = [foo]`

# base.conf.d/10-override.conf
array=bar # `10-override.array=[bar]`
{ "invalid": "stuff" } # Invalid format, ignored
array=baz # `10-override.array=[bar,baz]`
# end of override file, aggregator notified of `array += [bar,baz]`, `aggregator.array = [foo,bar,baz]`
```

We already have a couple of test cases that test simpler cases, but I'll add this one for completeness sake.

tests/miral/config_aggregator.cpp
```diff
@@ -0,0 +1,744 @@
...
+    aggregator.add_int_attribute(a_key, "a scoped int", [this](auto... args) { int_handler(args...); });
+
+    std::istringstream stream1{a_key.to_string() + "=10\n" + a_key.to_string() + "=20\n"};
+    std::istringstream stream2{a_key.to_string() + "=30\n" + a_key.to_string() + "=not_a_number\n"};
+
+    EXPECT_CALL(*this, int_handler(a_key, Optional(20)));
+
+    load_all({{std::ref(stream1), "base.conf"},
+              {std::ref(stream2), "base.conf.d/10-override.conf"}});
+}
```
tarek-y-ismail, line 644: This shouldn't be the case. Instead, the final value should be 30 (we ignore just the invalid line), or the following test has the final value of 20 (we ignore the key for the whole file)

tests/miral/config_aggregator.cpp
AlanGriffiths: Yes.

I think our only difference is how we consider "overrides" in the drop-in directory.

Model A: Each file in turn is processed using the running result of previous files as defaults. "Bad" files are skipped.
Model S: All files are concatenated and processed. "Bad" files are skipped.

Both give the same result here.

Saviq:
> Isn't the base config being valid a core assumption? With overrides, users don't have do touch that, so it shouldn't get corrupted anyway.

I'm just asking if we have a test for that. Not everyone will be using overrides.

> > This should keep [foo].
> ```
> # end of override file, aggregator notified of `array += [bar,baz]`, `aggregator.array = [foo,bar,baz]`
> ```

That's wrong, then - the whole file needs to be rejected on _parsing_ errors (parsing comes before validation).

tests/miral/config_aggregator.cpp
Saviq: Agree, this should yield 30.

tests/miral/config_aggregator.cpp
Saviq: By my definition "Bad" files are only those that fail to parse.

Within those that _parse_ fine (are valid INI), validation is what may reject individual values.

But that's not a hill I'll die on.

src/miral/symbols.map
tarek-y-ismail, line 1: Probably already out of date since we just made a release

tarek-y-ismail:
Array clearing was dropped somewhere along the way. Re-implemented it again, though For the sake of keeping this PR focused on changes needed for `ConfigAggregator`, I'll probably split `BasicStore` and `IniFile` changes into a separate PR, as this one is already quite big.

src/miral/symbols.map
```diff
@@ -775,3 +775,34 @@ global:
     miral::WindowSpecification::alpha*;
   };
 } MIRAL_5.6;
+
+MIRAL_5.8 {
+global:
+  extern "C++" {
+    miral::live_config::ConfigAggregator::?ConfigAggregator*;
+    miral::live_config::ConfigAggregator::ConfigAggregator*;
```
tarek-y-ismail, line 783: This symbol is missing in the debian symbols list.

include/miral/miral/live_config.h
```diff
@@ -94,6 +94,18 @@ class Store
...
+        std::string_view description,
+        HandleStrings handler,
+        std::span<std::string const> initial_values) = 0;
+    virtual void add_strings_attribute(
+        Key const& key,
+        std::string_view description,
+        std::span<std::string const> preset,
+        HandleStrings handler,
+        std::span<std::string const> initial_values) = 0;
+
```
AlanGriffiths, line 108: These functions do not belong in the `Store` (or `IniFile`) APIs.

1. They are not part of the `Store` abstraction used to subscribe to configuration
2. They are an implementation detail of `ConfigAggregator`
3. Adding to the vtab breaks ABI

include/miral/miral/config_aggregator.h
```diff
@@ -0,0 +1,94 @@
...
+{
+class ConfigAggregator : public live_config::Store
+{
+public:
+    struct Source
+    {
+        std::shared_ptr<Store> store;
+        std::move_only_function<void()> load;
+        std::filesystem::path file_path;
+    };
```
AlanGriffiths, line 35: Having structs like this in the API is a recipe for unstable APIs.

include/miral/miral/config_aggregator.h
```diff
@@ -0,0 +1,94 @@
...
+    ~ConfigAggregator();
+
+    ConfigAggregator(ConfigAggregator const&) = delete;
+    ConfigAggregator(ConfigAggregator&&) = delete;
+    ConfigAggregator& operator=(ConfigAggregator const&) = delete;
+    ConfigAggregator& operator=(ConfigAggregator&&) = delete;
+
+    void add_source(Source&& source);
+    void update_source(Source&& source);
+    void remove_source(std::filesystem::path const& path);
```
AlanGriffiths, line 47: This _might_ make sense in an internal API, but is not what we want as a public API for this.

The end user shouldn't need to deal with sources, they should simply specify the type the sources and the files to monitor/aggregate. Any adding or deleting of sources should be done internally in response to filesystem changes.

src/miral/basic_store.h
```diff
@@ -56,18 +56,29 @@ class BasicStore : public Store
...
+    using ForeachAttributeCallback =
+        std::function<void(Key const&, std::string_view, std::optional<std::string> const& preset)>;
+    using ForeachArrayCallback = std::function<void(
+        Key const&,
+        std::string_view,
+        std::span<std::string const> current_value,
+        std::optional<std::span<std::string const>> preset)>;
+
+    void foreach_attribute(ForeachAttributeCallback const& callback) const;
+    void foreach_array_attribute(ForeachArrayCallback const& callback) const;
```
AlanGriffiths, line 78: These functions appear to poking holes in the `BasicStore` abstraction - which suggests the abstraction is wrong. Or that we need a different abstraction.

Specifically, there appears to be a missing `AttributesCollection` that both `BasicStore` and `ConfigAggregator` can use to maintain a list of attributes. (Instead of poking a hole in `BasicStore` so that `ConfigAggregator` can get at the attributes.)

src/miral/basic_store.h
```diff
@@ -56,18 +56,29 @@ class BasicStore : public Store
...
+
     void update_key(Key const& key, std::string_view value, std::filesystem::path const& modification_path);
+    void set_array(Key const& key, std::optional<std::span<std::string const>> value, std::filesystem::path const& modification_path);
```
AlanGriffiths: These functions appear to poking holes in the BasicStore abstraction - which suggests the abstraction is wrong. Or that we need a different abstraction.

Specifically, `BasicStore` is _not_ basic. It has `IniFile` specific logic for accumulating arrays. This `set_array()` function is an alternative implementation, not one that belongs in the same class.

Maybe we can "push down" the `IniFile` specifics into `BasicStore::Self` and leave something truly basic?

src/miral/config_aggregator.cpp
```diff
@@ -0,0 +1,209 @@
...
+                        std::span<std::string const> current_value,
+                        std::optional<std::span<std::string const>> preset)
+                    {
+                        if (preset)
+                            source.store->add_strings_attribute(
+                                key, description, *preset, self->array_accumulator(), current_value);
+                        else
+                            source.store->add_strings_attribute(
+                                key, description, self->array_accumulator(), current_value);
+                    });
```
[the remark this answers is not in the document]
tarek-y-ismail: For scalar attributes, we can register once and re-register on updates. But, for arrays, it's required that we re-register before reloading so that the loaded value from the previous file is set as the initial value for the current file. I decided it would be less confusing to do the same to non-array keys for the sake of symmetry.

src/miral/basic_store.cpp
```diff
@@ -187,6 +191,32 @@ void mlc::BasicStore::Self::update_key(Key const& key, std::string_view value, s
...
+            details.parsed_values = std::vector<std::string>{value->begin(), value->end()};
+            details.clear_requested = false;
+        }
+        else
+        {
+            details.parsed_values.clear();
+            details.clear_requested = true;
+        }
+
+        details.modification_paths.push_back(modification_path);
```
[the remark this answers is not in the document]
tarek-y-ismail: I could swear I implemented this like three times already!

AlanGriffiths (changes requested):
I think this is trying to do too many things and loses sight of the end goal. Let's talk Monday

AlanGriffiths:
Closing in favour of #4926

PR #4840, opened 2026-04-09
Allowing override of the associate method and passing boxes from C++ -> Rust as references

mattkae:
## What's new?

- The `associate` method should be overridable
- Also, the `Box` parameters (coupled with associate) should be `const&` when going from C++ to Rust so as not to transfer ownership

## Checklist

- [ ] Tests added and pass


robert-ancell (approved):
My Rust is not strong enough to check the validity of the Copilot comment but otherwise looks good.

PR #4842, opened 2026-04-13
fix(console): restore previously active VT on LinuxVirtualTerminal teardown

AristoChen:
## What's new?

When Mir activates a specific VT via open_vt() (i.e. when vt_number > 0), it calls VT_ACTIVATE to switch the display to that VT. However, restore() only restored the keyboard mode and KD graphics state of that VT. It never called VT_ACTIVATE to switch the display back to the VT that was active before Mir took over. This left the display stuck on Mir's VT (showing a blank text console) after shutdown.

Fix this by recording the active VT number just before activating Mir's VT in open_vt(), then issuing VT_ACTIVATE on that saved number in both restore() and the emergency cleanup handler.

## How to test

On a system where ubuntu-frame is installed

1. Note the active VT before starting Mir (e.g. VT1 with a login prompt): `cat /sys/class/tty/tty0/active`
2. Start ubuntu-frame (e.g. `sudo snap start ubuntu-frame.daemon`)
3. Stop ubuntu-frame (e.g. `sudo snap stop ubuntu-frame.daemon`)
4. Confirm the active VT has returned to its original value and the display shows the expected content (e.g. the login prompt): `cat /sys/class/tty/tty0/active`

Before this fix step 4 would show Mir's VT still active, leaving a blank display. After this fix the active VT is restored to its pre-Mir value.
## Checklist

- [ ] Tests added and pass


AlanGriffiths (changes requested):
This doesn't work as I expect: the previous session is not correctly restored.

On a desktop I ran this as root:

```
mkdir /run/user/0
XDG_RUNTIME_DIR=/run/user/0 miral-shell --console-provider=vt --vt=4
```

It correctly switches to VT 4 and partially switches back (`fgconsole` returns 2, or 1 depending what the starting VT was) but the previous "owner" doesn't show (In my case GDM on 1 and Miriway on 2). Subsequent switching with `chvt` works as expected.

AristoChen:
Hi @AlanGriffiths 

Thanks for the feedback! The issue should be fixed now, could you please take a look when you have time?

AlanGriffiths:
> Thanks for the feedback! The issue should be fixed now, could you please take a look when you have time?

I see no difference in behaviour

AristoChen:
Hi @AlanGriffiths 

thanks for the feedback, just to clarify if my testing steps are correct or not, here is what I did on a arm64 device

```sh
# In serial console
ubuntu@ubuntu:~$ sudo mkdir -p /run/user/0
ubuntu@ubuntu:~$ sudo fgconsole
3
ubuntu@ubuntu:~$ cat /sys/class/tty/tty0/active
tty3
ubuntu@ubuntu:~$ sudo XDG_RUNTIME_DIR=/run/user/0 /home/ubuntu/mir/build/bin/miral-shell --console-provider=vt --vt=4

# In a new SSH console
ubuntu@ubuntu:~$ sudo fgconsole
4
ubuntu@ubuntu:~$ cat /sys/class/tty/tty0/active
tty4

# afterwards, either use CTRL+C in the serial console or execute `sudo pkill -SIGINT miral-shell.bin` in the SSH console to stop the miral-shell

ubuntu@ubuntu:~$ sudo fgconsole
3
ubuntu@ubuntu:~$ cat /sys/class/tty/tty0/active
tty3
```

AFAICT, the switch works as I expected, could you please tell me if my testing steps are correct or I am actually doing something incorrectly? Thanks!

AlanGriffiths:
> AFAICT, the switch works as I expected, could you please tell me if my testing steps are correct or I am actually doing something incorrectly? Thanks!

The switch does work according to, for example, `fgconsole`. We are seeing the same there.

However, the process that previously owned previous VT doesn't repaint. E.g. if previous is "GDM displayed on VT1" then then post is "blank VT1". Using, `chvt` to switch VTs does display the original content. Similarly if previous is "Miriway on VT2"

AristoChen:
Hi @AlanGriffiths 

> the process that previously owned previous VT doesn't repaint

I was able to see the previous VT repainted, I will try to recored a video if that helps

May I know what is your environment setup? I can try to reproduce it as well

AlanGriffiths:
> May I know what is your environment setup? I can try to reproduce it as well

Ubuntu 24.04

AristoChen:
> Ubuntu 24.04

I am also using Ubuntu 24.04

May I know if you are using any GPU hardware accerlation on your device or it's a pure software rendering?

I am asking because if I understand correctly, on my NXP i.MX platform, the NXP GPU libraries somehow hijacked the signal handler, so whenever the miral-shell is terminated, it is not able to catch the signal correctly, and the original VT never get repainted.

However, if using software rendering without the NXP GPU libraries on my device, the original VT can be repainted without issue.

AlanGriffiths:
> May I know if you are using any GPU hardware accerlation on your device

My "device" is a standard laptop with amdgpu. (I'm running the tests from a ssh session)

AristoChen:
Tested on a HP ELITEBOOK with AMD GPU and fresh new installed Ubuntu Noble Desktop, I am still not able to reproduce the issue that you have mentioned, do you have any logs or ideas what could be the root cause?

AlanGriffiths:
> I am still not able to reproduce the issue that you have mentioned, do you have any logs or ideas what could be the root cause?

Sorry, not been able to get to this today. But it is not forgotten

tarek-y-ismail:
Can confirm it works when testing locally. 

Steps I followed:
1. `sudo bash` to get root
2. `mkidr -p /run/user/1` (0 was already created)
3. `XDG_RUNTIME_DIR=/run/user/1 <path/to/build>/bin/miral-shell --console-provider=vt --vt 4`
4. Once the shell launched, I opened a terminal and ran `fgconsole`, the output was "4".
5. Trying again with `--vt 5`, and re-running steps, `fgconsole` prints "5".

AlanGriffiths:
> 4\. Once the shell launched, I opened a terminal and ran `fgconsole`, the output was "4".

That isn't what this PR is about. It is about _returning to the previous VT session_ when the shell exits.

The issue I am seeing is that:

1. ssh in from another computer.
2. Note the current VT session (GDM on VT1, say)
3. Run `miral-shell --console-provider=vt --vt 4` as root (as described above)
4. Exit the shell

Expect: previous VT session resumes
Actual: a blank screen, `fgconsole` reports the expected 1 (AristoChen sees GDM resume, so we're wondering what's different)

tarek-y-ismail:
Works on my machine :tm: following the steps you mentioned.

AlanGriffiths:
> Works on my machine ™️ following the steps you mentioned.

And... works on a different machine (running 26.04 FWIW)

tarek-y-ismail:
> running 26.04 FWIW

Mine is still on 25.10. So we know it's not a release version thing.

AlanGriffiths:
As this works in 3/4 cases I've dismissed my "Request changes" review. I've not had time to investigate further and have not re-read the current code, so not approving. Yet.

AlanGriffiths (approved):
OK, the problem was at my end. (A pull failed without me noticing)

PR #4844, opened 2026-04-13
Converting shared_ptr references to wayland_rs::Weak for a more idiomatic API

mattkae:
## What's new?
- Add `wayland_rs::Weak` which is similar to the existing `wayland::Weak`
- Generating a middleware function for all methods in C++ that accept a `shared_ptr` to a wayland object. Instead, we generate _another_ method that needs to be implemented which takes a `wayland_rs::Weak` as a parameter in place of a shared_ptr
- The `wayland_rs::Weak` construct entirely disallows protocol implementers from gaining access to a `shared_ptr` for a protocol object

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
+++ b/src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
@@ -19,6 +19,7 @@
 
 #include <memory>
 #include <stdexcept>
+#include <boost/throw_exception.hpp>
 
 namespace mir
 {
@@ -33,11 +34,13 @@ public:
     {
     }
 
+    Weak(Weak const& weak) = default;
+
     auto value() const -> T&
     {
         if (auto locked = ptr_.lock())
             return *locked;
-        throw std::logic_error{"Accessing expired wayland_rs::Weak"};
+        BOOST_THROW_EXCEPTION(std::logic_error{"Accessing expired wayland_rs::Weak"});
     }
 
     operator bool() const
--- a/src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
+++ b/src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
@@ -19,6 +19,7 @@
 
 #include <memory>
 #include <stdexcept>
+#include <utility>
 #include <boost/throw_exception.hpp>
 
 namespace mir
@@ -36,10 +37,11 @@ public:
 
     Weak(Weak const& weak) = default;
 
-    auto value() const -> T&
+    template<typename F>
+    decltype(auto) with_value(F&& f) const
     {
         if (auto locked = ptr_.lock())
-            return *locked;
+            return std::forward<F>(f)(*locked);
         BOOST_THROW_EXCEPTION(std::logic_error{"Accessing expired wayland_rs::Weak"});
     }
 
--- a/src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
+++ b/src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
@@ -1,0 +1,72 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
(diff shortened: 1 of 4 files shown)


src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
```diff
@@ -0,0 +1,72 @@
...
+    explicit Weak(std::shared_ptr<T> const& ptr)
+        : ptr_{ptr}
+    {
+    }
+
+    auto value() const -> T&
+    {
+        if (auto locked = ptr_.lock())
+            return *locked;
+        throw std::logic_error{"Accessing expired wayland_rs::Weak"};
```
RAOF:
```suggestion
        BOOST_THROW_EXCEPTION((std::logic_error{"Accessing expired wayland_rs::Weak"}));
```

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
```diff
@@ -0,0 +1,72 @@
...
+        return !ptr_.expired();
+    }
+
+    auto is(T const& other) const -> bool
+    {
+        auto locked = ptr_.lock();
+        return locked && locked.get() == &other;
+    }
+
+    auto operator==(Weak<T> const& other) const -> bool
```
RAOF, line 59: This could probably usefully be an `auto operator<=>(Weak<T> const& other) const -> std::partial_order`?

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
```diff
@@ -0,0 +1,72 @@
...
+        return !ptr_.owner_before(other.ptr_) && !other.ptr_.owner_before(ptr_);
+    }
+
+    auto operator!=(Weak<T> const& other) const -> bool
+    {
+        return !(*this == other);
+    }
+
+    auto operator=(Weak<T> const&) -> Weak<T>& = default;
+
```
RAOF, line 70: Should we also explicitly default the copy constructor here?

RAOF (commented):
Some minor tweaks, but broadly OK.

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
mattkae: My compiler is upset because it doesn't recognize the spaceship operator. And I guess on top of that less than or greater than isn't very important here, right?

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
```diff
@@ -0,0 +1,77 @@
...
+    Weak(Weak const& weak) = default;
+
+    template<typename F>
+    decltype(auto) with_value(F&& f) const
+    {
+        if (auto locked = ptr_.lock())
+            return std::forward<F>(f)(*locked);
+        BOOST_THROW_EXCEPTION(std::logic_error{"Accessing expired wayland_rs::Weak"});
+    }
```
RAOF, line 46: Since we're not going to guarantee that everything's on a single Wayland thread it makes some sense to do this dance rather than the simple `auto value() -> T&` accessor.

RAOF (approved):
Yeah, I'm ok with this.

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
RAOF: How can your compiler not recognize the spaceship operator? We use it elsewhere in the code!

Yeah, the `<` and `>` are not very interesting; the interesting thing is having `==` and `!=` automatically derived from a single implementation.

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/weak.h
AlanGriffiths: That conflicts with making the step from libwayland as small as possible: the existing `Weak` class has syntax and semantic is `auto value()`

PR #4848, opened 2026-04-14
Implement a C++ friendly Wayland client and WaylandServerNotificationHandler for receiving notifications about new clients (and beyond)

mattkae:
This is intended to fill the gaps of the existing `WlCient`.

## What's new?
- Created `WaylandServerNotificationHandler` to receive notifications form the `WaylandServer` in Rust to the C++ business logic
- Created `WaylandClient` and `WaylandClientId` and added notifications to them so that users know when the clients become available
- Triggering the notifications at the proper time

## How to test
Run the mirwayland_rs-demo demo

## Checklist

- [ ] Tests added and pass


src/server/frontend_wayland/wayland_rs/src/wayland_client.rs
```diff
@@ -0,0 +1,55 @@
...
+        self.client
+            .get_credentials(&self.handle)
+            .map_or(0, |c| c.pid)
+    }
+
+    /// Retrieve the uid of the client.
+    pub fn uid(&self) -> u32 {
+        self.client
+            .get_credentials(&self.handle)
+            .map_or(0, |c| c.uid)
```
RAOF: Hm. `0` is not a great fallback as it's a valid `uid` — root! Unfortunately, there' *isn't* a value specified to be an invalid uid 🤦‍♀️

RAOF (changes requested):
Looks broadly fine, but we need some way of distinguishing between “unable to get credentials of the client” and “the client is running as root” 😬

src/server/frontend_wayland/wayland_rs/src/wayland_client.rs
mattkae: We should be able to return a `Result` type to C++ and allow the caller to decide what to do here (e.g. an exception, a nullopt, etc.)

RAOF (approved):
Yup, that's better.

PR #4847, opened 2026-04-14
Add tests for `IniFile` failure modes

tarek-y-ismail:
Related: #4863, #4829, #4846

## What's new

- Adds a bunch of tests for `IniFile`'s unhappy cases: invalid syntax, whitespace, etc...

## How to test

- `<path/to/build>/miral-test --gtest_filter=LiveConfigIniFile.*`

## Checklist

- [x] Tests added and pass

PR #4850, opened 2026-04-15
Split parsing out from `BasicStore`

tarek-y-ismail:
Related: #4847 

## What's new?

- Splits out the type related parsing code (including presets) from `BasicStore` into `TypedStoreAdapter` to make it reusable. Now, `BasicStore` becomes a simple string/handler storage class, and conversion to floats/ints/other is handled externally. This will enable `ConfigAggregator` to share the parsing code used in `IniFile`.
- Moves `BasicStore` tests to use `TypedStoreAdapter`. We can probably re-introduce tests for the basic use cases of `BasicStore`, but that's out of scope for this PR.

## How to test

- `LiveConfigIniFile` tests must still pass: `<path/to/build>/miral-test --gtest_filter=LiveConfigIniFile.*`
- `TypedStoreAdapter` tests must still pass (follows the same test filter pattern)


src/miral/typed_store_adapter.cpp
```diff
@@ -0,0 +1,451 @@
...
+                return std::string{"never set"};
+
+            return join_comma(
+                array.modification_paths | std::views::transform([](auto const& p) { return p.string(); }));
+        }();
+
+        if (!preset.empty())
+        {
+            auto const foo = std::views::transform(preset, [](auto const& e) { return to_string(e); });
+            auto const preset_str = join_comma(foo);
```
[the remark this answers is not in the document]
tarek-y-ismail: Oops, was debugging some compiler error and forgot to remove this

src/miral/typed_store_adapter.cpp
```diff
@@ -0,0 +1,451 @@
...
+
+#include "typed_store_adapter.h"
+
+#include <mir/log.h>
+
+#include <boost/throw_exception.hpp>
+
+#include <charconv>
+#include <format>
+#include <ranges>
```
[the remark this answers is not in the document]
tarek-y-ismail: Should not be a big problem. Wontfix.

src/miral/live_config_ini_file.cpp
```diff
@@ -27,6 +28,7 @@ class mlc::IniFile::Self: public mlc::BasicStore
 {
 public:
     void load_file(std::istream& istream, std::filesystem::path const& path);
+    TypedStoreAdapter adapter{*this};
```
AlanGriffiths: Wouldn't `class Self : BasicStore, public TypedStoreAdapter`  _IsA_, not _HasA_ be a cleaner design?

src/miral/basic_store.h
```diff
@@ -20,39 +20,33 @@
...
+    using HandleAttribute = std::function<void(Key const& key, std::optional<ScalarValue> scalar)>;
+
+    struct ArrayValue
+    {
+        std::vector<std::string> values;
+        std::set<std::filesystem::path> modification_paths;
+    };
+    using HandleArrayAttribute = std::function<void(Key const& key, ArrayValue array)>;
+
+    void add_scalar_attribute(Key const& key, std::string_view description, HandleAttribute handler);
```
AlanGriffiths, line 48: Odd naming: why are we calling this "scalar"?

[scalar](https://dictionary.cambridge.org/dictionary/english/scalar): "something that has size but no direction, such as a quantity, distance, speed, or temperature"

src/miral/typed_store_adapter.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include <span>
+#include <string_view>
+
+namespace miral::live_config
+{
+class TypedStoreAdapter
+{
+public:
+    explicit TypedStoreAdapter(BasicStore& store);
```
AlanGriffiths, line 32: Doesn't this leave `TypedStoreAdapter` too coupled to `BasicStore`? I thought we'd want to use the parsing it does later. Or is this implementation specific to IniFile?

AlanGriffiths (changes requested):
Apart from the naming, I'm just seeking clarification on the way forward

src/miral/live_config_ini_file.cpp
tarek-y-ismail: Marginally, yes. It would save this line and would replace `self->adapter.` with `self->`. I always thought of this class as something on top of a `BasicStore` to handle type conversions, but it can also work as a `TypedStore`

src/miral/typed_store_adapter.h
tarek-y-ismail: I don't see an issue with that. It doesn't affect its usage in `ConfigAggregator` (check #4830. Warning: It's a bit messy)

src/miral/basic_store.h
tarek-y-ismail: That's the first thing that came to mind for "singular element".

tarek-y-ismail:
Something I noticed reading through the code is that `BasicStore` no longer implements `Store`. The name might cause some confusion.

tarek-y-ismail:
Converting to draft for now while we get array clearing merged.

tarek-y-ismail:
After dabbling a bit with #4917, I don't think this PR is needed.

Issue #1612, opened 2020-07-17
Desktop session loses the touchpad after suspend/resume


AlanGriffiths:
1. On a laptop run mir-shell or egmde from the greeter.
2. Close the lid to suspend
3. Open the lid to resume

The greeter starts up, and the keyboard (and external mouse) work fine. But the touchpad doesn't.

wmww:
I can reproduce running master on a TTY. FWIW TTY switching away and back fixes it.

AlanGriffiths:
Possibly shares a cause with #617

AlanGriffiths:
I don't know what's behind it, but after resume touchpad input leads to:
```plain
[2020-08-05 16:35:01.130492] < -warning- > evdev-input: Failed to find requested fd for path /dev/input/event6
Failed to set flags: Bad file descriptor (9)
[2020-08-05 16:35:01.130570] < -warning- > evdev-input: Failed to find requested fd for path /dev/input/event6
```
(And, from other debugging, `/dev/input/event6` is "UNIW0001:00 093A:1336 Touchpad-event6-2362-4918")

PR #4853, opened 2026-04-15
evdev: Fix FDs being removed by libinput

### LINKED ISSUE #4818: Mir's evdev plugin cannot handle libinput device suspension/resumption
### Version

v2.25.2

### Environment

OS: Debian testing
CPU: Intel Core i7-1360P
GPU 1: Intel Xe Graphics
Monitor Setup: Internal laptop monitor

libinput's philosophy is to offload the process of keeping track of FDs of event nodes to the program making use of libinput, by providing an instance of libinput_interface. libinput will automatically open and close event devices by using this interface. Mir implements this with its FdStore class. But the issue is that logind takes exclusive control of these event devices, and Mir fills its internal map of FDs by querying logind initially. If libinput asks for a FD, Mir will pass the same FD allocated by logind. Whenever the laptop is switched into tablet mode, evdev will suspend the trackpad, keyboard and trackpoint, which will cause libinput to call FdStore for closing the FD. I assume Mir developers figured out that you cannot close an FD allocated by logind, so all Mir does is remove the reference to the FD from its internal map. However, when you switch back to notebook mode, libinput will try to resume the devices it suspended, which makes it call FdStore to obtain an FD.

Now in normal circumstances you would just open the file again and pass an FD, but since logind has exclusive control over the event device, you can't do that. Moreover logind's concept of "paused" devices is vastly different from libinput's concept, which causes this issue to happen in the first place. If logind was correctly pausing and resuming devices, Mir would automatically refill its internal map of FDs with the resumed devices, as the logind_console class holds a list of acquired devices that it uses to fill the FdStore. Now in that video I made Mir not remove the FD from its map, however for e.g. the vt_console provider, the FDs are owned by Mir itself, so we'll have to close them. Now I don't really know how to go around this, so I figured I'd drop this here

### Reproduction steps

It can be reproduced by sending a lid event or tablet-switch event to libinput, which causes it to suspend keyboard and trackpad.

### Relevant log output

```shell

```

muhammad23012009:
Closes https://github.com/canonical/mir/issues/4818

Related: https://github.com/canonical/mir/issues/1612

## What's new?

This PR makes the evdev plugin ignore fd_close events from libinput, as we don't want any external lib meddling with Mir's internal structure of FDs and their ownership.

## How to test

* Test with a device that is capable of suspending input devices. E.g. a laptop with lid closed and connected to a monitor, or a 2-in-1 flipped back.
* If patch works properly, the input devices should still work after reverting to the original state.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/platforms/evdev/fd_store.cpp
+++ b/src/platforms/evdev/fd_store.cpp
@@ -38,11 +38,6 @@ mir::Fd mie::FdStore::take_fd(char const* path)
     }
     catch (std::out_of_range const&)
     {
-        if (removed.first == path)
-        {
-            mir::log_warning("Requested fd for path %s was removed", path);
-            return fds.insert(std::move(removed)).first->second;
-        }
         mir::log_warning("Failed to find requested fd for path %s", path);
     }
     return mir::Fd{};
@@ -64,8 +59,6 @@ void mie::FdStore::remove_fd(int fd)
     }
     else
     {
-        // We keep the last removed element in case libinput asks for it
-        removed = *element;
         fds.erase(element);
     }
 }
--- a/src/platforms/evdev/fd_store.h
+++ b/src/platforms/evdev/fd_store.h
@@ -42,13 +42,6 @@ private:
     FdStore& operator=(FdStore const&) = delete;
 
     std::unordered_map<std::string, mir::Fd> fds;
-
-    // LibInputPtr calls remove_fd() for touchpads on suspend but
-    // continues to use take_fd() to access the fd after resume!
-    // As a workaround, we remember the last removed fd and reinstate
-    // it if asked for.
-    //                  https://github.com/canonical/mir/issues/1612
-    std::pair<std::string, mir::Fd> removed;
 };
 
 }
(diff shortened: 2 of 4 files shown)


RAOF (approved):
Yup, this looks good now.

RAOF:
Urgh, sorry. The final thing to do here is to get the [Canonical CLA check](https://github.com/canonical/mir/pull/4853/checks?check_run_id=71520845554) to pass. :woman_facepalming: 

That should just require going to https://canonical.com/legal/contributors and clicking the relevant buttons (if you are happy to click those buttons, of course).

muhammad23012009:
> Urgh, sorry. The final thing to do here is to get the [Canonical CLA check](https://github.com/canonical/mir/pull/4853/checks?check_run_id=71520845554) to pass. 🤦‍♀️
> 
> That should just require going to https://canonical.com/legal/contributors and clicking the relevant buttons (if you are happy to click those buttons, of course).

Done, signed and submitted it.

muhammad23012009:
Hmm, is there a reason why this PR failed to merge? I don't see any failed checks...

Saviq:
> I don't see any failed checks...

There were infrastructure issues, sorry for the delay, and thank you!

PR #4860, opened 2026-04-16
Providing each Wayland interface instance with access to their client

mattkae:
## What's new?
- Interface implementations will require their client, so all constructors must provide it in a protected constructor

## Checklist

- [ ] Tests added and pass


src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
```diff
@@ -365,6 +411,20 @@ impl CppClass {
             .expect("enums cannot be empty after push")
     }
 
+    pub fn add_protected_constructor_arg(&mut self, arg: CppArg) -> &mut CppArg {
+        self.protected_constructor_args.push(arg);
+        self.protected_constructor_args
+            .last_mut()
+            .expect("protected_constructor_args cannot be empty after push")
```
tarek-y-ismail, line 418: Kinda redundant to check right after pushing. What is this trying to assert?

src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
```diff
@@ -149,6 +149,48 @@ impl CppBuilder {
...
+
+                    if !class.protected_constructor_args.is_empty() {
+                        let ctor_args_str: Vec<String> = class
+                            .protected_constructor_args
+                            .iter()
+                            .map(|arg| {
+                                format!(
+                                    "{} {}",
+                                    cpp_arg_type_to_cpp_source(&arg.cpp_type, true),
+                                    arg.name
```
tarek-y-ismail, line 165: I think I've seen this pattern before while reviewing related code. Might be worth refactoring out at some point (not urgent)

tarek-y-ismail (approved):
Pokes here and there, but fine otherwise.

src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
RAOF, line 433: General review: none of these need to exist? Any code that uses them could instead just mutate the member of `CppClass` directly (and when we can use Rust 1.95 the `-> &mut _` return value is just [`vec.push_mut`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push_mut)).

Maybe we should wait until we can use 1.95 to do this cleanup, though.

RAOF (approved):
Yup, that's reasonable.

src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
mattkae: It's useful as a sanity check, we'll never hit it in practice, but it would be preferable over a panic

src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
mattkae: Yeah this might be better addressed later

PR #4863, opened 2026-04-17
Better handle unhappy paths in `BasicStore`

tarek-y-ismail:
## What's new?

- When a value is un-parsable, either calls the handler with the preset if provided, or with `nullopt` otherwise.
- Adds tests for `BasicStore` 
- Modifies `IniFile` tests to handle the new behavior.

## How to test

- `<path/to/build>/bin/miral-test-internal --gtest_filter=BasicStoreTest*`

## Checklist

- [x] Tests added and pass


src/miral/basic_store.cpp
```diff
@@ -138,18 +162,42 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
...
+                key.to_string().c_str(),
+                path.c_str(),
+                pe.what());
+
+            if (details.preset)
+            {
+                mir::log_warning(
+                    "Using preset value '%s' for key '%s' instead.", details.preset->c_str(), key.to_string().c_str());
 
+                details.handler(key, details.preset);
```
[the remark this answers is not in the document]
tarek-y-ismail: Preset values are guaranteed to be parsable, so it should not be possible to throw unless the user provided a handler that always throws or invalid preset values.

src/miral/basic_store.cpp
```diff
@@ -176,32 +224,40 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
...
         {
-            auto const path = [&]
-            {
-                // TODO Not all CI compilers support `std::ranges::to`
-                auto comma_separated =
-                    std::ranges::views::transform(details.modification_paths, [](auto const& p) { return p.string(); })
-                    | std::views::join_with(',');
+            auto const path = comma_join(
+                std::ranges::views::transform(details.modification_paths, [](auto const& p) { return p.string(); }));
+            auto const value_str = maybe_value.transform([](auto const& v) { return comma_join(v); }).value_or("unset");
```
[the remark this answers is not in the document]
tarek-y-ismail: Lost in translation. Nice catch :)

tarek-y-ismail:
Not sure if it's worth splitting up further..

src/miral/basic_store.cpp
```diff
@@ -9,9 +9,23 @@
...
+namespace
+{
+
+class ParsingError : public std::runtime_error
+{
+public:
+    ParsingError(std::string const& message) :
+        std::runtime_error(message)
+    {
+    }
```
AlanGriffiths: Like most single argument constructors, this should probably be explicit. But there's a better syntax for what you want anyway:

```suggestion
    using std::runtime_error::runtime_error;
```

src/miral/basic_store.cpp
```diff
@@ -9,9 +9,23 @@
...
 #include <ranges>
 #include <set>
+#include <stdexcept>
 
 namespace mlc = miral::live_config;
 
+namespace
+{
+
+class ParsingError : public std::runtime_error
```
AlanGriffiths, line 19: We *might* want to consider making this available to handler implementations. (Probably as a more generic `BadValue` type.) But that can be a later PR.

src/miral/basic_store.cpp
```diff
@@ -138,18 +152,42 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
                 return std::nullopt;
         }();
 
+
```
AlanGriffiths:
```suggestion
```

src/miral/basic_store.cpp
AlanGriffiths: But, seeing the usage, wouldn't a better constructor be:

```suggestion
    explicit ParsingError(Key const& key, std::string_view const& value) :
        std::runtime_error{std::format("Config key '{}' has invalid value: {}", key.to_string(), value)}
    {
    }
```

And, shouldn't we add a formatter for `Key` instead of `key.to_string()` when we call std::format? C.f. #4865

src/miral/basic_store.cpp
```diff
@@ -138,18 +152,42 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
...
-                    .value_or("never set");
+            auto const path = details.value.transform([&](auto const& v) { return v.modification_path.string(); })
+                                  .value_or("never set");
 
+            mir::log_warning(
+                "Failed to parse value '%s' for key '%s' in file %s: %s.",
+                std::string{maybe_value.value_or("unset")}.c_str(),
+                key.to_string().c_str(),
+                path.c_str(),
+                pe.what());
```
AlanGriffiths: `pe.what()` already contains the key and value, so this message will be repetitive

src/miral/basic_store.cpp
```diff
@@ -138,18 +152,42 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
...
+                mir::log_warning(
+                    "Using preset value '%s' for key '%s' instead.", details.preset->c_str(), key.to_string().c_str());
+
+                details.handler(key, details.preset);
+            }
+            else
+            {
+                mir::log_warning("No preset value, passing nullopt for key '%s' instead.", key.to_string().c_str());
+                details.handler(key, std::nullopt);
+            }
```
AlanGriffiths: I think two log messages for each `ParseError` is excessive. Combine them into one

tests/miral/basic_store.cpp
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    void run_empty()
+    {
+        store.do_transaction([]{});
+    }
+};
+}
+
+TEST_F(BasicStoreTest, valid_int_value_is_forwarded)
```
AlanGriffiths: "forwarded" introduces an unnecessary ambiguous term. The code has "handler", so why not use "handled"

AlanGriffiths (changes requested):
I'm wondering whether all these tests belong on `BasicStore`. We talked about making this class a {Key, string} collection and separating out parsing the values into a helper class. Is that still on the path ahead?

tarek-y-ismail:
> I'm wondering whether all these tests belong on BasicStore. We talked about making this class a {Key, string} collection and separating out parsing the values into a helper class. Is that still on the path ahead?

That it indeed coming in #4850.

src/miral/basic_store.cpp
```diff
@@ -239,7 +275,7 @@ void process_as(std::function<void(mlc::Key const&, std::optional<Type>)> const&
         }
         else
         {
-            throw std::runtime_error(std::format("Config key '{}' has invalid value: {}", key, *val));
+            throw ParsingError(key, *val);
```
RAOF: Pre-existing, but we normally `BOOST_THROW_EXCEPTION()`; is there any reason not to use that here?

tests/miral/basic_store.cpp
```diff
@@ -0,0 +1,377 @@
...
+        store.update_key(a_strings_key, "bar", a_path);
+    });
+}
+
+TEST_F(BasicStoreTest, ints_preset_is_used_when_no_values_are_set)
+{
+    std::vector<int> const preset_vals{10, 20};
+    store.add_ints_attribute(an_ints_key, "some ints", preset_vals, [this](auto k, auto v){ ints_handler(k, v); });
+
+    EXPECT_CALL(*this, ints_handler(an_ints_key, Optional(ElementsAre(10, 20))));
```
RAOF: This would be cleaner as:
```suggestion
    EXPECT_CALL(*this, ints_handler(an_ints_key, Optional(ElementsAreArray(preset_vals))));
```

RAOF (commented):
Couple of nits, but more broadly, there seems to be a bit of inconsistent behaviour (or I'm missing design context) when there are invalid arguments:
* When there's a preset, you never witness an empty `optional` (maybe these overloads should take a callback with a different signature?), otherwise…
* When there's a list, you get the parseable arguments, the invalid ones silently disappear, unless…
* That list is *entirely* absent (not merely contains only unparsable values), in which case you get `nullopt`.
* If it's a single-value arg, then unset and set-to-unparsable-value are indistinguishable, both `nullopt`

Why is list-contains-only-invalid-values distinguishable from list-unset, whereas single-value-set-to-invalid-value is indistinguishable from single-value-unset?

tarek-y-ismail:
> When there's a preset, you never witness an empty optional (maybe these overloads should take a callback with a different signature?), otherwise…

Having a separate callback signature in that case is probably clearer. But the API ship has already sailed. We can add overloads for this case, but existing interface methods must remain untouched. The store interface is already quite wide though...

> Why is list-contains-only-invalid-values distinguishable from list-unset, whereas single-value-set-to-invalid-value is indistinguishable from single-value-unset?

That was an oversight on my part. I've added an extra bit of code to check if the store was given some values but none of them were parsed. Which is now handled by sending `nullopt` or the preset if possible.

src/miral/basic_store.cpp
tarek-y-ismail: No particular reason, just built atop what was there.

src/miral/basic_store.cpp
```diff
@@ -17,11 +16,13 @@ namespace mlc = miral::live_config;
...
 namespace
 {
-template <typename Range> auto comma_join(Range const& strings) -> std::string
+template <typename Range> auto join_comma(Range const& strings) -> std::string
 {
     auto const sep = std::string_view{", "};
     auto comma_separated = strings | std::views::join_with(sep);
 
+    // LEGACY(24.04) - `std::ranges::to` is not yet widely available, so we need to copy the result into a string
+    // manually if it's not available. Remove this workaround once we upgrade to 26.04
```
AlanGriffiths: I know what you mean, but it isn't "upgrading to 26.04" - we already support that. It is dropping support for 24.04 which comes somewhat later.

AlanGriffiths (approved):
OK. (Not sure we need `BOOST_THROW_EXCEPTION` for exceptions that should never escape the local context. But... 🤷)

In fact the more I look at it, the more I doubt whether exceptions are the right solution. But "pre-existing".

tarek-y-ismail:
> OK. (Not sure we need BOOST_THROW_EXCEPTION for exceptions that should never escape the local context. But... 🤷)

Better to stay consistent I would say. 

> In fact the more I look at it, the more I doubt whether exceptions are the right solution. But "pre-existing".

I too got this feeling while modifying the code earlier today. But as this code is still in flux, I say we delay refining it a bit more.

PR #4864, opened 2026-04-17
Minor `BasicStore` improvements

tarek-y-ismail:
## What's new?

- Clears `modification_paths` on transaction
- Adds `comma_join` to join string ranges with a comma. 

## How to test

- `IniFile` tests must still pass.

### DIFF
--- a/src/miral/basic_store.cpp
+++ b/src/miral/basic_store.cpp
@@ -2,7 +2,6 @@
 
 #include <mir/log.h>
 
-#include <algorithm>
 #include <format>
 #include <list>
 #include <map>
--- a/src/miral/basic_store.cpp
+++ b/src/miral/basic_store.cpp
@@ -122,7 +122,10 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
         details.value = std::nullopt;
 
     for (auto& [key, details] : array_attribute_handlers)
+    {
         details.parsed_values.resize(0);
+        details.modification_paths.clear();
+    }
 
     transaction_body();
 
--- a/src/miral/basic_store.cpp
+++ b/src/miral/basic_store.cpp
@@ -12,6 +12,26 @@
 
 namespace mlc = miral::live_config;
 
+namespace
+{
+template <typename Range> auto join_comma(Range const& strings) -> std::string
+{
+    auto const sep = std::string_view{", "};
+    auto comma_separated = strings | std::views::join_with(sep);
+
+    // LEGACY(24.04) - `std::ranges::to` is not yet widely available, so we need to copy the result into a string
+    // manually if it's not available. Remove this workaround once we upgrade to 26.04
+#ifdef __cpp_lib_ranges_to_container
+    return comma_separated | std::ranges::to<std::string>();
+#else
+    std::string temp;
+    std::ranges::copy(comma_separated, std::back_inserter(temp));
+
+    return temp;
+#endif
+}
+}
+
 class mlc::BasicStore::Self
 {
 public:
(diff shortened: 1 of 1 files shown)


src/miral/basic_store.cpp
```diff
@@ -12,6 +12,20 @@
 
 namespace mlc = miral::live_config;
 
+namespace
+{
+template <typename Range> auto comma_join(Range const& strings) -> std::string
+{
+    auto comma_separated = strings | std::views::join_with(std::string_view{", "});
+
+    // TODO Not all CI compilers support `std::ranges::to`
```
mattkae, line 21: Why is this a TODO? Will we act on it?

src/miral/basic_store.cpp
```diff
@@ -180,28 +197,14 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
         {
             auto const path = [&]
             {
-                // TODO Not all CI compilers support `std::ranges::to`
-                auto comma_separated =
-                    std::ranges::views::transform(details.modification_paths, [](auto const& p) { return p.string(); })
-                    | std::views::join_with(',');
+                if (details.modification_paths.empty())
+                    return std::string{"never set"};
```
mattkae, line 201:
```suggestion
                    return std::string{"unset"};
```

mattkae (commented):
Minor nits!

src/miral/basic_store.cpp
tarek-y-ismail: This code could be written as 
```cpp
return strings | std::views::join_with(std::string_vew{", "}) | std::ranges::to<std::string>();
```

But right now, the GCC version in our CI doesn't support this. I believe it might be possible to use `std::ranges::to` once 26.04 is released. Though I'm not sure. I left the TODO in the code as a reminder.

src/miral/basic_store.cpp
tarek-y-ismail: Why?

src/miral/basic_store.cpp
RAOF: It would be helpful if the reminder contained the code this can be replaced with ;)

src/miral/basic_store.cpp
RAOF: It's more familiar jargon for config handling, I think, but I personally don't mind either way.

RAOF (approved):
LGTM, but I *would* like the `TODO` comment to contain the actual line of code we want to use, but can't.

src/miral/basic_store.cpp
AlanGriffiths: Can I suggest not using a comment, especially a TODO: C++ has `__cpp_lib_ranges_to_container` to detect the availability

src/miral/basic_store.cpp
```diff
@@ -12,6 +12,20 @@
...
+{
+template <typename Range> auto comma_join(Range const& strings) -> std::string
+{
+    auto comma_separated = strings | std::views::join_with(std::string_view{", "});
+
+    // TODO Not all CI compilers support `std::ranges::to`
+    std::string temp;
+    std::ranges::copy(comma_separated, std::back_inserter(temp));
+
+    return temp;
```
AlanGriffiths:
```suggestion
    // TODO remove conditional once we drop support for Ubuntu 24.04
#ifdef __cpp_lib_ranges_to_container
    return strings | std::views::join_with(std::string_vew{", "}) | std::ranges::to<std::string>();
#else    
    std::string temp;
    std::ranges::copy(comma_separated, std::back_inserter(temp));

    return temp;
#endif
```

AlanGriffiths (changes requested):
We might want to agree a comment tag for code we want to change post 24.04: "TODO" is a bit too generic

src/miral/basic_store.cpp
tarek-y-ismail: I did not know that one existed. Thank you.

tarek-y-ismail:
> We might want to agree a comment tag for code we want to change post 24.04: "TODO" is a bit too generic

Initial suggestion: "UPDATEME"

AlanGriffiths:
> > We might want to agree a comment tag for code we want to change post 24.04: "TODO" is a bit too generic
> 
> Initial suggestion: "UPDATEME"

I was think more along the lines of "PRE_26.04" or "24.04 LEGACY"

AlanGriffiths:
> I was think more along the lines of "PRE_26.04" or "24.04 LEGACY"

In six months we'll be able to search for this stuff and update it. Having an identifiable tag will make things easier.

src/miral/basic_store.cpp
```diff
@@ -12,6 +12,24 @@
 
 namespace mlc = miral::live_config;
 
+namespace
+{
+template <typename Range> auto comma_join(Range const& strings) -> std::string
```
AlanGriffiths: [non-blocking] The name isn't ideal: the main concept being modelled is "join", not "comma". (It is "join with comma", not "using comma, join")

src/miral/basic_store.cpp
```diff
@@ -12,6 +12,24 @@
...
+{
+    auto const sep = std::string_view{", "};
+    auto comma_separated = strings | std::views::join_with(sep);
+
+#ifdef __cpp_lib_ranges_to_container
```
AlanGriffiths, line 23: Do we want to start tagging code written to support 24.04/core24 for cleanup?

AlanGriffiths (approved):
Nothing blocking, but thoughts

src/miral/basic_store.cpp
tarek-y-ismail: Good idea. Tagged it with `LEGACY(24.04)`

src/miral/basic_store.cpp
mattkae: Yeah +1 for the jargon, but it doesn't matter too much, just a nit

mattkae (approved):

PR #4870, opened 2026-04-21
Return protocol error if trying to put duplicate roles on a surface

robert-ancell:


AlanGriffiths (approved):
Makes sense. I guess it is _possible_ some misbehaving clients could be affected, but we'll find that out sooner by landing.

PR #4872, opened 2026-04-21
Validate WlSurface buffer scale

robert-ancell:


RAOF (approved):
We should work out whether we're going to be *always* using `BOOST_THROW_EXCEPTION`, or where we should be using a raw `throw`, but this is mirroring the existing code, so 👍

RAOF:
Likewise, any chance of a WLCS test? :stuck_out_tongue:

PR #4880, opened 2026-04-21
Adding an explicit post error call to Wayland impl objects

mattkae:
## What's new?
- We may need to post an error on an object from another thread, so we handle that by exposing an explicit method (replacement for https://manpages.debian.org/unstable/libwayland-doc/wl_resource_post_error.3.en.html)

## Checklist

- [ ] Tests added and pass


src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -127,6 +127,18 @@ fn generate_extension_for_interface(interface: &WaylandInterface) -> Option<Toke
...
         impl #interface_name_ext {
+            pub fn post_error(&mut self, code: u32, message: &CxxString) {
+                use wayland_server::Resource;
+                if let Some(handle) = self.wrapped.handle().upgrade() {
+                    handle.post_error(
+                        self.wrapped.id(),
+                        code,
+                        std::ffi::CString::new(message.to_string())
+                            .expect("Error message contained null byte"),
+                    );
```
RAOF: Maybe I'm missing context, but why isn't this
```suggestion
                if self.wrapped.is_alive() {
                    self.wrapped.post_error(
                        code,
                        message.to_string()
                    );
```
It doesn't seem like we need to go through the resource -`handle()`> Weak -`upgrade()`> Handle dance?

mattkae: True!

src/server/frontend_wayland/wayland_rs/build_script/main.rs
```diff
@@ -925,6 +925,17 @@ fn wayland_interface_to_cpp_class(interface: &WaylandInterface) -> CppClass {
...
     class.add_method(get_box_method);
 
+    // Add a post_error method that posts a protocol error to the client for this resource.
+    // This follows the same pattern as event-sending methods, routing through the Rust
+    // middleware Ext type to call wayland-server's DisplayHandle::post_error().
+    let mut post_error_method = CppMethod::new("post_error", None, false, false, false, true);
+    post_error_method.add_arg(CppArg::new(CppType::CppU32, "code", false));
+    post_error_method.add_arg(CppArg::new(CppType::String, "message", false));
+    post_error_method.set_body(
+        "if (!instance_.has_value()) {\n    throw \"post_error() used before associate()\";\n}\ninstance_.value()->post_error(code, message);",
```
RAOF: Please do not throw a string :)
Since this is a very easily avoidable precondition error, we can just `assert()` it.
```suggestion
        "assert(instance_.has_value());\ninstance_.value()->post_error(code, message);",
```

RAOF (changes requested):
Other than throwing a `char const*`, this looks good to me.

RAOF (approved):
Actually, I'll mark it as approved, so you can land it immediately after fixing the `throw`.

src/server/frontend_wayland/wayland_rs/build_script/main.rs
mattkae: True! This will disappear soon anyway, as I have a ticket to remove this 2-step initialization

PR #4882, opened 2026-04-21
Validate anchor enum provided to positioner

robert-ancell:
This is currently done in Mutter, and specified for the gravity anchor. See https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/507 for this to be formalized in the specification.


RAOF (approved):
Sensible.

PR #4883, opened 2026-04-21
Check session lock succeeds to avoid a null pointer dereference

robert-ancell:
Matches existing code pattern.


RAOF (approved):
Yup, surfaces can outlast their session.

PR #4885, opened 2026-04-22
Validate all inputs to wl_shm_pool::create_buffer

robert-ancell:
Replace existing check that could overflow an int32_t.
Check pools only increase size.


robert-ancell:
Specifically `if (stride < (width * 4))` would overflow if the width was too large.

src/server/frontend_wayland/shm.cpp
```diff
@@ -251,25 +251,43 @@ void mf::ShmPool::create_buffer(
...
+    if (format != wayland::Shm::Format::argb8888 && format != wayland::Shm::Format::xrgb8888)
     {
-        backing_range = backing_store->get_rw_range(offset, size);
+        throw wayland::ProtocolError{
+            resource,
+            wayland::Shm::Error::invalid_format,
+            "Invalid SHM format requested"};
     }
-    catch (std::logic_error const&)
+    auto const pixel_size = 4; // All supported formats are 4 bytes per pixel
```
RAOF: Could you please fix the TODO while you're at it?

That would be adding an `auto bits_per_pixel() const -> std::optional<uint32_t>` accessor to `DRMFormat`, and updating the `FormatInfo` list in drm_formats.cpp; then you can use it here for the checks.

src/server/frontend_wayland/shm.cpp
```diff
@@ -251,25 +251,43 @@ void mf::ShmPool::create_buffer(
...
+    }
+
+    auto const max_size = std::numeric_limits<int32_t>::max() / pixel_size;
+    auto const max_height = max_size / stride;
+    if (width > max_size || height > max_height)
+    {
+        throw wayland::ProtocolError{
+            resource,
+            wayland::Shm::Error::invalid_stride,
+            "Requested SHM buffer size %dx%d (stride %d) is too large", width, height, stride};
```
RAOF: Rather than limiting our buffer size to those representable in `int`, why don't we do our size calculations in `size_t`?

Although I guess we still need to check when `sizeof(size_t)` == `sizeof(int)` 🤷‍♀️

src/server/shm_backing.cpp
```diff
@@ -418,6 +420,12 @@ void ShmBacking::resize(size_t new_size)
         backing_size_is_guaranteed_at_least(this->backing_store, new_size));
 }
 
+size_t ShmBacking::size()
+{
+   auto mapping = *current_mapping.lock();
```
RAOF: Oooooh! We've dropped 20.04, so `std::atomic<std::shared_ptr<>>` should now work! We should drop this synchronisation! (But not now).

RAOF (changes requested):
While you're here, can I get you to do just one small bit of extra work 🥺

src/server/frontend_wayland/shm.cpp
robert-ancell: If you use size_t it become less clear when you would have overflowed an int32_t - instead starting with the maximum and reducing it by dividing ensures your upper limits are always representable,

src/server/frontend_wayland/shm.cpp
robert-ancell: In a follow up PR :)

src/server/frontend_wayland/shm.cpp
RAOF: But our maximum size *isn't* limited by int32_t; `ShmBacking` uses `size_t` everywhere, so we're perfectly capable of representing `int32_t::max() * int32_t::max()`. (On size_t-is-64-bit platforms, to be fair).

Now, the maximum size that the Wayland frontend is going to resize the backing store to *is* `int32_t::max()`, because that size comes over as an int, but that means the `ShmBacking` is going to catch it (and, if in the incredibly unlikely future Wayland adds a the ability to allocate a >4GB shm pool, this code will still work :stuck_out_tongue: )

RAOF: That said, part of what was confusing here is the `width > max_size` leg of the if; that's subsumed in `height > max_height` and the later `stride`-is-not-impossible check.

Perhaps remove the `width` bit?

robert-ancell: Looking at `include/core/mir/geometry/forward.h` both Size and Stride use int, not size_t:
```c++
using Stride = generic::Value<int, StrideTag>;
using Size = generic::Size<int>;
```

src/server/frontend_wayland/shm.cpp
```diff
@@ -15,6 +15,7 @@
  */
 
 #include "shm.h"
+#include <limits>
```
AlanGriffiths: We traditionally put standard library headers at the end of the include list.

There are pre-existing breaks from tradition in this file too: e.g. "local" includes ("../shm_backing.h") should follow the own header

src/server/frontend_wayland/shm.cpp
robert-ancell: Now using the size type from the template.

src/server/frontend_wayland/shm.cpp
```diff
@@ -299,6 +334,20 @@ void mf::ShmPool::create_buffer(
...
 void mf::ShmPool::resize(int32_t new_size)
 {
+    if (new_size < 0)
+    {
+        throw wayland::ProtocolError{
+            resource,
+            wayland::Shm::Error::invalid_stride,
+            "Invalid new size %d", new_size};
+    }
+    if (static_cast<size_t>(new_size) < backing_store->size())
```
AlanGriffiths: We are not holding a lock here, so this truth value is ephemeral. Would it be better for `backing_store->resize()` to return a success indicator?

src/server/shm_backing.h
```diff
@@ -136,6 +136,7 @@ class ReadOnlyPool
 
     virtual auto get_ro_range(size_t start, size_t len) -> std::unique_ptr<ReadMappableRange> = 0;
     virtual void resize(size_t new_size) = 0;
+    virtual size_t size() const = 0;
```
AlanGriffiths:
```suggestion
    virtual auto resize(size_t new_size) -> std::expected<void, ResizeError> = 0;
```

AlanGriffiths (changes requested):
I don't like the QueryAttributeCheckItThenChangeIt pattern.

We don't need the multiple `virtual size_t size() const = 0` functions as the right place to check `new_size >= size` is in the `resize()` implementation. (Which needs to report failure)

src/server/frontend_wayland/shm.cpp
robert-ancell: Agreed. I've updated the code to do this. I was concerned that there might be cases where we do want to shrink the backing store but I think it's only used and will be used via the Wayland APIs.

src/server/frontend_wayland/shm.cpp
```diff
@@ -299,7 +334,27 @@ void mf::ShmPool::create_buffer(
...
+    if (new_size < 0)
+    {
+        throw wayland::ProtocolError{
+            resource,
+            wayland::Shm::Error::invalid_stride,
+            "Invalid new size %d", new_size};
+    }
+
+    auto result = backing_store->resize(new_size);
+    if (!result)
```
AlanGriffiths:
```suggestion
    if (auto result = backing_store->resize(new_size); !result)
```

src/server/shm_backing.h
```diff
@@ -104,6 +105,11 @@ class Mapping
     }
 };
 
+enum class ResizeError
+{
+    shrunk,
```
AlanGriffiths: Maybe not the best name for the error? "shrunk" sounds like something that happened, not a problem. Maybe "cannot_shrink"?

AlanGriffiths (approved):
I would prefer a better name for the error, but not blocking on that

src/server/frontend_wayland/shm.cpp
```diff
@@ -29,6 +29,8 @@
 #include <boost/throw_exception.hpp>
 #include <wayland-server-protocol.h>
 
+#include <limits>
```
robert-ancell, line 33:
```suggestion
#include <expected>
#include <limits>
```

robert-ancell:
Build is failing due to not having std::expected, which is a surprise since this is already used in `src/server/frontend_wayland/ext_image_capture_v1.cpp`

RAOF (approved):
Yeah, fair enough.

PR #4886, opened 2026-04-22
Validate virtual pointer motion

robert-ancell:
Handle the divide by zero case.


AlanGriffiths (approved):
Looks sensible given the spec

PR #4888, opened 2026-04-22
Accept EGL_TEXTURE_EXTERNAL_WL

Azkali:
On devices running libhybris based OS the EGL driver selects gralloc based formats that are exposed to Mir as EGL_TEXTURE_EXTERNAL_WL, accept the external format so apps can render in that case


AlanGriffiths (approved):
"Mostly harmless" I guess.

Is this really the only change needed to work on the libhybris stack? I guess anything else needed is in the graphics platform?

AlanGriffiths:
CI failure will be fixed by #4897

Azkali:
Thank you for the merge, nothing else needed elsewhere !

PR #4891, opened 2026-04-22
Replace server exception with Wayland protocol error in layer shell

robert-ancell:


src/server/frontend_wayland/layer_shell_v1.cpp
```diff
@@ -605,8 +605,9 @@ void mf::LayerSurfaceV1::ack_configure(uint32_t serial)
 
     if (acked_event == std::end(inflight_configures))
     {
-        BOOST_THROW_EXCEPTION(std::runtime_error(
-            "Could not find acked configure with serial " + std::to_string(serial)));
+        BOOST_THROW_EXCEPTION(mw::ProtocolError(
+            resource, mw::generic_error_code,
```
mattkae, line 613: I wonder if we should throw one of these codes, but it might not make sense:

https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:enum:error

The protocol fails to mention this edge case :)

mattkae (approved):

src/server/frontend_wayland/layer_shell_v1.cpp
robert-ancell: Yes, it's unclear if this is `invalid_surface_state`. I can't find any documentation or usage of this. Perhaps safer to not set it until that is made clear.

robert-ancell: Added a comment to note this for future fixing.

PR #4894, opened 2026-04-22
Ensure no loops in toplevel parents

robert-ancell:


src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -466,7 +466,36 @@ void mf::XdgToplevelStable::set_parent(std::optional<struct wl_resource*> const&
...
+                if (parent_surface.value() == this_surface.value())
+                {
+                    BOOST_THROW_EXCEPTION(mw::ProtocolError(
+                        resource,
+                        Error::invalid_parent,
+                        "Parent toplevel must not be a descendant of the child toplevel"));
+                }
+                auto const grandparent = parent_surface.value()->parent();
+                parent_surface = grandparent ? std::optional{grandparent} : std::nullopt;
+            }
```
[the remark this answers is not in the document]
tarek-y-ismail: Unfortunately, Copilot may be correct in this case. The parent set by `WindowWlSurfaceRole::set_parent` is `const`. I see no immediate way the parent we set and the parents we check connect.

tarek-y-ismail (changes requested):

src/server/frontend_wayland/xdg_shell_stable.cpp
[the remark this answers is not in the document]
robert-ancell: Split out https://github.com/canonical/mir/pull/4964

src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -466,7 +466,36 @@ void mf::XdgToplevelStable::set_parent(std::optional<struct wl_resource*> const&
...
+        {
+            throw mw::ProtocolError{
+                resource,
+                Error::invalid_parent,
+                "A toplevel cannot be its own parent"};
+        }
+
+        // Check that the parent is not a descendant of this toplevel
+        auto const this_surface = scene_surface();
+        if (this_surface)
```
AlanGriffiths: The scope of `this_surface` can be reduced

```suggestion
        if (auto const this_surface = scene_surface())
```

src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -466,7 +466,36 @@ void mf::XdgToplevelStable::set_parent(std::optional<struct wl_resource*> const&
...
+        {
+            auto parent_surface = parent_toplevel->scene_surface();
+            while (parent_surface)
```
AlanGriffiths: Again the scope can be reduced:
```suggestion
            while (auto parent_surface = parent_toplevel->scene_surface())
```

AlanGriffiths: But maybe just better as a `for ()` loop:

```suggestion
            for (auto ancestor = parent_toplevel->scene_surface().value();
                ancestor; 
                ancestor = ancestor->parent())
```

src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -466,7 +466,36 @@ void mf::XdgToplevelStable::set_parent(std::optional<struct wl_resource*> const&
...
+            {
+                if (parent_surface.value() == this_surface.value())
+                {
+                    throw mw::ProtocolError{
+                        resource,
+                        Error::invalid_parent,
+                        "Parent toplevel must not be a descendant of the child toplevel"};
+                }
+                auto const grandparent = parent_surface.value()->parent();
```
AlanGriffiths: We don't check `parent_surface` is not `nullopt` the first time through the loop

AlanGriffiths (changes requested):
Seems logical, but the code could be a bit simpler and safer.

```c++
        // Check that the parent is not a descendant of this toplevel
        static std::shared_ptr<scene::Surface> const null{};
        
        if (auto const this_surface = scene_surface().value_or(null))
        {
            for (auto ancestor = parent_toplevel->scene_surface().value_or(null); ancestor; ancestor = ancestor->parent())
            {
                if (ancestor == this_surface)
                {
                    throw mw::ProtocolError{
                        resource,
                        Error::invalid_parent,
                        "Parent toplevel must not be a descendant of the child toplevel"};
                }
            }
        }
```

src/server/frontend_wayland/xdg_shell_stable.cpp
robert-ancell: Confused by this - the original code had `while (parent_surface)`, therefore it can never be `nullopt`?

robert-ancell: This this is fixed by changing to your suggested use of `value_or()`

AlanGriffiths (approved):
👍

tarek-y-ismail (approved):

PR #4895, opened 2026-04-23
Initial implementation of the new logging API

RAOF:
## What's new?

This adds the logging side of the new logging API, in parallel to the existing API.

## How to test

The startup logs for a Mir compositor will now say `core: Starting` rather than `mirserver: Starting`.

## Checklist

- [x] Tests added and pass


Saviq (dismissed):
Some symbols shenanigans:

```
External Symbols Diff:

  New Symbols 🟢🟢🟢
    std::formatter::format*;
    typeinfo?for?std::formatter;
```

```
[ RUN ] SharedLibrary.load_valid_versioned_function_works
unknown file: Failure
C++ exception with description "/<<PKGBUILDDIR>>/build-amd64/bin/../lib/libmircore.so.3: undefined symbol: _ZN3mir17fatal_error_abortEPKcz, version MIR_CORE_2.9" thrown in the test body.

[ FAILED ] SharedLibrary.load_valid_versioned_function_works (0 ms)
```

RAOF:
Looks like we might want to make a few fixes to `tools/update_package_abis.sh` :)

RAOF:
Heh. Turns out our `SharedLibrary` tests need to be updated every time we break `mircore` ABI. And the symbols problem is the compilation environment changing in a way that makes GCC emit a symbol for the `std::formatter` in miral; we'll just need to make that always emitted and add it to the symbols.

.github/filters.yaml
tarek-y-ismail, line 26: Shouldn't `**` already recurse into subdirectories?

include/core/mir/log.h
```diff
@@ -56,10 +68,44 @@ void security_log(
...
+void log_debug(logging::Tags tags, std::format_string<Args...> fmt, Args&&... args)
+{
+    log(logging::Severity::debug, tags, fmt, std::forward<Args>(args)...);
+}
+
+template <typename... Args>
+void log_info(logging::Tags tags, std::format_string<Args...> fmt, Args&&... args)
+{
+    log(logging::Severity::informational, tags, fmt, std::forward<Args>(args)...);
+}
```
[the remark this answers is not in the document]
tarek-y-ismail: Valid point, but not essential. We can ignore this for now and add them when needed.

src/core/logging/logger.cpp
```diff
@@ -43,6 +96,20 @@ void ml::Logger::log(char const* component, Severity severity, char const* forma
...
+{
+    // Default implementation uses :-delimited tags as the component
+    // TODO: Remove the log(Severity, std::string, std::string) interface and replace
+    // with this one.
+
+
+    std::string component;
+    std::ranges::copy(
+        tags | std::views::transform([](Tag const& tag) { return tag.name(); }) | std::views::join_with(':'),
+        std::back_inserter(component));
```
tarek-y-ismail, line 106: Could use a `LEGACY(24.04)` comment. Once we drop 24.04 support and can use `ranges::to` this can be rewritten as:
```cpp
auto const component = tags 
    | std::views::transform([](Tag const& tag) { return tag.name(); })
    | std::views::join_with(':')
    | std::ranges::to<std::string>();
```

include/core/mir/logging/logger.h
```diff
@@ -54,6 +81,14 @@ class Logger
...
          __attribute__ ((format (printf, 4, 5)));
 
+    virtual void log(Severity severity, Tags tags, std::string_view message);
+
+    template<typename... Args>
+    void log(Severity severity, Tags tags, std::format_string<Args...> fmt, Args&&... args)
+    {
+        log(severity, tags, std::format(fmt, std::forward<Args>(args)...));
+    }
+
```
AlanGriffiths: We're intentionally dropping the performance objective: "don't do work (fetching data or formatting) if logging is suppressed"?

tests/unit-tests/logging/test_logging.cpp
```diff
@@ -0,0 +1,145 @@
...
+        // allowing it to be destroyed and verified after the test finishes.
+        mir::logging::set_logger(std::make_shared<MockLogger>());
+        ASSERT_TRUE(logger.unique()) << "Test needs to destroy the MockLogger to verify assertions";
+        logger.reset();
+    }
+
+    std::shared_ptr<testing::NiceMock<MockLogger>> logger;
+};
+
+TEST_F(TestLog, log_calls_logger)
```
tarek-y-ismail, line 57: This and friends could probably be turned to one parametric test

AlanGriffiths (changes requested):
Copilot is right about global const: AFAICS there is no reason the global objects are not `const`. But it might be better to provide wrapper functions (that return a reference) - that gives us a point where we could insert additional logic in future.

Note: We _could_ preserve ABI by introducing the new member functions in a new derived class. That would need some extra logic around `set_logger()` (so not clear it is worth the hassle)

tarek-y-ismail (commented):
Mostly fine, but i have some comments/questions.

include/core/mir/logging/logger.h
tarek-y-ismail: I think that's meant for a followup

tarek-y-ismail:
> Copilot is right about global const

Huh?

.github/filters.yaml
Saviq: [Not in this case](https://github.com/micromatch/picomatch#basic-globbing):

> `**` will only match path separators (`/`, and `\\` with the windows option) when they are the only characters in a path segment.

include/core/mir/logging/logger.h
RAOF: That's an additional override, yeah. (Along with the (`Severity`, `Tags`, `string_view`) overrides)

That said, I can totally add it here now.

tests/unit-tests/logging/test_logging.cpp
RAOF: I was thinking that, but I wasn't sure the up-front complexity would be worth the flexibility.

tarek-y-ismail: Fair.

tarek-y-ismail (approved):
Still have to fix symbols. But the core of the PR is solid 👍

Saviq:
> Still have to fix symbols

This will help with the debian and armhf symbols (the diff will become sane):
- https://github.com/canonical/mir/pull/4905

RAOF:
> Copilot is right about global const: AFAICS there is no reason the global objects are not const.

A follow-up will make them non-const (they'll store their level).

> But it might be better to provide wrapper functions (that return a reference) - that gives us a point where we could insert additional logic in future.

I was going to start to do this, but then:

We already have a point where we can introduce additional logic; the Tag constructor. What switching to a wrapper function *could* do is make initialisation lazy, on first access, but the spec says we explicitly want eager, on-DSO-load behaviour. We could make these globals references to hidden objects, but I don't think that indirection gets us anything.

RAOF:
On further further thought, no, the visible tags can and should be `const`; the internals need to be mutable, but that's what `Impl` is for!

RAOF:
>  New Symbols 🟢🟢🟢
>    std::formatter::format*;
>    typeinfo?for?std::formatter;

Unfortunately, this is a limitation of our symbols checker; these symbols do not exist. Instead, `std::formatter<live_config::Key, char>::format` exists, but the symbols checker doesn't handle templated classes.

AlanGriffiths:
> > Copilot is right about global const: AFAICS there is no reason the global objects are not const.
> 
> A follow-up will make them non-const (they'll store their level).

Is that in addition to the (already const) `Impl`? That will be another ABI break.

> > But it might be better to provide wrapper functions (that return a reference) - that gives us a point where we could insert additional logic in future.
> 
> I was going to start to do this, but then:
> 
> We already have a point where we can introduce additional logic; the Tag constructor. What switching to a wrapper function _could_ do is make initialisation lazy, on first access, but the spec says we explicitly want eager, on-DSO-load behaviour. We could make these globals references to hidden objects, but I don't think that indirection gets us anything.

Do we actually have the guarantees we need? Especially that the order of related tags is correct? 

See also: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#ri-global

[edit]

I think this is the relevant passage:

> it happens before the first [ODR-use](https://en.cppreference.com/cpp/language/definition#ODR-use) of any variable with static/thread storage duration defined in the same translation unit as the variable to be initialized. If no variable or function is ODR-used from a given translation unit, the non-local variables defined in that translation unit may never be initialized (this models the behavior of an on-demand dynamic library)

https://en.cppreference.com/cpp/language/initialization#Non-local_variables

RAOF:
I'm *pretty* sure that we do have the guarantees we need. But you know what's better than being pretty sure? It being absolutely obvious.

Updated to make it obvious.

src/miral/symbols.map
```diff
@@ -775,3 +775,10 @@ global:
     miral::WindowSpecification::alpha*;
   };
 } MIRAL_5.6;
+
+MIRAL_5.8 {
```
AlanGriffiths, line 779: Should we be bumping the libmiral ABI? We have dependencies on libmircore, so I can imagine a compositor based on mircore2 linking with a miral5 based on mircore3 and UB

include/core/mir/logging/logger.h
```diff
@@ -54,6 +83,19 @@ class Logger
     virtual void log(char const* component, Severity severity, char const* format, ...)
          __attribute__ ((format (printf, 4, 5)));
 
+    /*
+     * Cruft removal: when we're sure we've got the right API/ABI, we should
+     * *remove* log(severity, string, string) above and make this the
+     * required-to-implement interface.
+     */
+    void log(Severity severity, Tags tags, std::string_view message);
```
AlanGriffiths, line 91: Will this be the required interface? Or will that be one that allows the severity to be validated before constructing `message`?

AlanGriffiths (approved):
Sure

include/core/mir/logging/logger.h
RAOF: This will be the required interface; we will implement the "don't construct `message` before severity is checked" interface on top of this (there's no need or benefit in pushing that responsibility down to implementors).

RAOF:
Moar rebasing!

PR #4898, opened 2026-04-23
Re-implement array clearing in BasicStore and IniFile

tarek-y-ismail:
## What's new?

- What it says on the tin

## How to test

- `<path/to/build>/bin/miral-test --gtest_filter=LiveConfigIniFile.*`
- `<path/to/build>/bin/miral-test-internal --gtest_filter=BasicStoreTest.*`

## Checklist

- [x] Tests added and pass


src/miral/basic_store.cpp
```diff
@@ -228,6 +244,31 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
...
+                std::ranges::views::transform(details.modification_paths, [](auto const& p) { return p.string(); }));
+        }();
+
+        if (details.clear_requested)
+        {
+            try
+            {
+                details.handler(key, std::span<std::string const>{});
+                if (details.parsed_values.empty())
+                    continue;
```
tarek-y-ismail: Note to self & reviewers: This check ensures that we don't accidentally clear, then call the handler with the preset or nullopt a few lines below. This is a decision I made on the fly, so I'm not particularly attached to it.

src/miral/live_config_ini_file.cpp
```diff
@@ -61,7 +61,10 @@ void mlc::IniFile::Self::load_file(std::istream& istream, std::filesystem::path
                         auto const key = Key{trim_start_and_end(line_view.substr(0, eq))};
                         auto const value = trim_start_and_end(line_view.substr(eq + 1));
 
-                        update_key(key, value, path);
+                        if (!value.empty())
+                            update_key(key, value, path);
+                        else if (!clear_array(key))
+                            update_key(key, value, path);
```
tarek-y-ismail: I do agree the two branches _can_ be collapsed. But I don't think they _should_

tarek-y-ismail:
TICS complaints are:
1. Don't use a variadic function (`mir::log_warning`)
2. Don't duplicate an if statement's branch (see https://github.com/canonical/mir/pull/4898#discussion_r3135938374)

src/miral/live_config_ini_file.cpp
AlanGriffiths: I think it is confusing because `clear_array()` has surprising semantics: it includes a check to see whether `key` refers to an array value and returns the result of that check.

Wouldn't that all be clearer if the test and clear logic was all in `update_key()` and we dropped the `clear_array()` method entirely? That already has separate logic for single and array values.

src/miral/basic_store.cpp
AlanGriffiths: I'm not sure why there's a check. Surely if we've called a handler for this key  we should `continue` to the next key regardless.

src/miral/basic_store.cpp
```diff
@@ -228,6 +244,31 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
...
+            }
+            catch (std::exception const& e)
+            {
+
+                mir::log_warning(
+                    "Error clearing key '%s' in file '%s': %s", key.to_string().c_str(), modification_paths_str.c_str(), e.what());
```
AlanGriffiths: This message will be misleading: the 'file' is a comma-separated list of files.

src/miral/basic_store.cpp
```diff
@@ -91,6 +92,7 @@ class mlc::BasicStore::Self
         std::optional<std::vector<std::string>> const preset;
         std::vector<std::string> parsed_values;
         std::set<std::filesystem::path> modification_paths;
```
AlanGriffiths: Pre-existing, but isn't `set` the wrong type? Wouldn't it be better to maintain the list in processing order so that this is reflected in any error messages.

AlanGriffiths (changes requested):

src/miral/live_config_ini_file.cpp
tarek-y-ismail: Good idea. My original thinking was to give the caller the power to decide when to clear. But for now, this will work.

src/miral/basic_store.cpp
tarek-y-ismail: I foolishly thought that `std::set` would preserve order due to the existence of `std::unordered_set` :/

src/miral/basic_store.cpp
tarek-y-ismail: There are two cases: 1. We cleared, and didn't append any values afterwards, 2. We cleared, and appended values afterwards. The handler needs to be called with an empty span in both cases, but only in the second case is it strictly necessary to call the handler with the appended values.

If we call the handler with an empty span, then continue to the next key, the user will not be informed of the values appended after clearing.

This is mainly needed so the `ConfigAggregator` is informed of clears and can update its internal state accordingly.

AlanGriffiths: > There are two cases: 1. We cleared, and didn't append any values afterwards, 2. We cleared, and appended values afterwards. The handler needs to be called with an empty span in both cases, but only in the second case is it strictly necessary to call the handler with the appended values.

That sounds like making `BasicStore` a leaky abstraction: why would we call the handler with an empty span when we have values? Why don't we just call it with the values?

My expectation is that the aggregator will initialize each store with the working value (if any) as the preset and the store will call back once with the resulting value.

tarek-y-ismail: Yes. This is a carry-over from the work done to make `ConfigAggregator` work. Since `BasicStore` is internal, we can modify its API to better convey the cases. 

Though for `ConfigAggregator`, we still hit the same issue of `std::optional<std::span>>` not being able to cleanly express our cases on one side, and us being confined to the `Store` interface on the other side , but that's not part of this PR.

AlanGriffiths: OK, so let's make `BasicStore` in this PR work consistently: one callback per key in a "transaction" regardless of whether there was a clear or not.

AlanGriffiths (changes requested):

src/miral/basic_store.cpp
tarek-y-ismail: Done. Note that in `IniFile` handlers, we still have to call the handler twice to notify users that a clear happened in the file being parsed.

AlanGriffiths:
Dropped in favour of #4914

PR #4914, opened 2026-04-28
Re-implement array clearing in BasicStore and IniFile

AlanGriffiths:
## What's new?

- What it says on the tin

## How to test

- `<path/to/build>/bin/miral-test --gtest_filter=LiveConfigIniFile.*`
- `<path/to/build>/bin/miral-test-internal --gtest_filter=BasicStoreTest.*`

## Checklist

- [x] Tests added and pass

Note: what I think #4898 should be

### DIFF
--- a/src/miral/basic_store.cpp
+++ b/src/miral/basic_store.cpp
@@ -258,29 +258,11 @@ void mlc::BasicStore::Self::do_transaction(std::function<void()> transaction_bod
 
     for (auto const& [key, details] : array_attribute_handlers)
     {
-        if (details.clear_requested)
-        {
-            try
-            {
-                details.handler(key, std::span<std::string const>{});
-                if (details.parsed_values.empty())
-                    continue;
-            }
-            catch (std::exception const& e)
-            {
-                mir::log_warning(
-                    "Error clearing key '%s' in file '%s': %s",
-                    key.to_string().c_str(),
-                    details.last_clear_file.c_str(),
-                    e.what());
-            }
-        }
-
         auto const maybe_value = [&]() -> std::optional<std::vector<std::string>>
         {
             if (!details.parsed_values.empty())
                 return details.parsed_values;
-            else if (details.preset)
+            else if (details.preset && !details.clear_requested)
                 return details.preset;
             else
                 return std::nullopt;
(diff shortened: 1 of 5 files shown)


AlanGriffiths:
@copilot apply changes based on the comments in [this thread](https://github.com/canonical/mir/pull/4914#pullrequestreview-4195189124)

tarek-y-ismail:
I don't see how this would allow us to handle clear-and-append when it's time to integrate it with `ConfigAggregator`. 

```cpp
auto const maybe_value = [&]() -> std::optional<std::vector<std::string>>
{
    if (!details.parsed_values.empty())
        return details.parsed_values;
    else if (details.preset && !details.clear_requested)
        return details.preset;
    else
        return std::nullopt;
}();
```

If a file clears and doesn't assign, then we pass `nullopt` to the handler. That's fine. If a file clears, _then_ appends onto the array, we go through the first branch. On the handler end, it will receive a span with the appended elements, but no signal that a clear occurred. So

```
# File 1
my_array=1
my_array=2

# File 2
my_array=
my_array=3
```

Will result in `[1, 2, 3]` instead of `[3]`

AlanGriffiths:
> I don't see how this would allow us to handle clear-and-append when it's time to integrate it with `ConfigAggregator`.

1. That is out of scope
2. We have had a long conversation about this on Mattermost

The way it will work is that we need to pass in the current value:

```
process(file1, nullopt) -> {1, 2}
process(file2, {1 ,2}) -> {3}
```

But we don't yet have an API to pass in the current value

AlanGriffiths:
> If a file clears and doesn't assign, then we pass `nullopt` to the handler. That's fine. If a file clears, _then_ appends onto the array, we go through the first branch. On the handler end, it will receive a span with the appended elements, but no signal that a clear occurred. So

You're implicitly assuming that `ConfigAggregator` is responsible for appending arrays and `BasicStore`(?) is responsible for returning the content of file. I'm assuming that `ConfigAggregator` is responsible for co-ordinating `BasicStores`(?) and `BasicStore`(?) is responsible for returning the new value

[edit]

I've pushed a sketch of this: https://github.com/canonical/mir/pull/4917/changes

src/miral/basic_store.cpp
```diff
@@ -147,8 +148,18 @@ void mlc::BasicStore::Self::update_key(Key const& key, std::string_view value, s
...
+        {
+            details.parsed_values.clear();
+            details.clear_requested = true;
+        }
+        else
+        {
+            details.parsed_values.push_back(std::string{value});
+        }
+
+        details.modification_paths.push_back(modification_path);
```
tarek-y-ismail: If we're moving from a set to a vector in this PR, then we should check for duplicates.

tarek-y-ismail (approved):
One nit. Good to go otherwise.

PR #4915, opened 2026-04-29
Fix output from symbols checker

robert-ancell:
If external symbols were found, then the internal symbol check was skipped due to the second part of the boolean operation not being run.

Don't show headers if no symbols added/removed and show a clear message if no symbols have changed.


RAOF (approved):
Yup, that seems helpful.

AlanGriffiths (approved):

PR #4917, opened 2026-04-29
Use BasicStore to aggregate strings

AlanGriffiths:
We will need a way to update store content across multiple IniFile fragments

PR #4918, opened 2026-04-29
Simpler array clearing

tarek-y-ismail:
## What's new?

- Re-implements array clearing in a simpler way by allowing external users to force the initial value of `parsed_values`. Which are then used as is, cleared, or appended onto.

## How to test

`<path/to/build>/bin/miral-test-internal --gtest_filter=BasicStoreTest.*`

## Checklist

- [x] Tests added and pass


AlanGriffiths (changes requested):
I see the motivation (similar to #4917) but I don't think this is the right API to expose. The problem is that the interaction between the "transaction" and "set_ints_value" needs to be tightly controlled and this approach leaves that as a responsibility for the user

tarek-y-ismail:
Closed in favor of #4914 and #4917

PR #4920, opened 2026-04-30
ci: further reduce unnecessary push runs

Saviq:
## What's new?

SymbolsCheck is required on merge queue, no need to run on `main`.
Metadata files don't impact packaging or tests.

## How to test

Can't…

## Checklist

- [ ] Tests added and pass


AlanGriffiths (approved):
> SymbolsCheck is required on merge queue, no need to run on `main`.

Is that assuming we _never_ push directly to main? I guess that is rare enough and not usually worth the check

Saviq:
> Is that assuming we _never_ push directly to main? I guess that is rare enough and not usually worth the check

Correct, that's an exception

PR #4923, opened 2026-05-04
Tidy up parse_output_sizes

robert-ancell:
Remove cast (always use size_t) and simplify length check.


RAOF (approved):
Yeah, fair enough.

PR #4924, opened 2026-05-04
Remove C casts

robert-ancell:
As per the style guide.


robert-ancell:
Hmm, the TICS failure is due to replacing the C cast with a reinterpret_cast. I think this PR is an improvement, and not sure if we can avoid the reinterpret_cast without changing the code.

Saviq:
> I think this PR is an improvement, and not sure if we can avoid the reinterpret_cast without changing the code.

It's probably a good candidate for [suppression](https://ticsdocumentation.tiobe.com/latest/docs/index.html#doc=user/clientusecases.html%23ViolationSuppression), then.

robert-ancell:
Replaced the reinterpret_cast - should be good now!

src/server/shell/decoration/decoration_strategy.cpp
```diff
@@ -82,6 +82,23 @@ static constexpr auto color(unsigned char r, unsigned char g, unsigned char b, u
 #endif
 }
 
+static constexpr auto unpack_color(uint32_t c, unsigned char &r, unsigned char &g, unsigned char &b, unsigned char &a) -> void
```
AlanGriffiths: We typically use the anonymous namespace, not `static` to control visibility. Also, as we're also handling alpha would `unpack_pixel` be a better name?

src/server/shell/decoration/decoration_strategy.cpp
```diff
@@ -618,14 +635,14 @@ void RendererStrategy::Text::Impl::render_glyph(
         for (geom::X buffer_x = buffer_left; buffer_x < buffer_right; buffer_x += geom::DeltaX{1})
         {
             geom::X const glyph_x = buffer_x - glyph_offset.dx;
-            unsigned char const glyph_alpha = ((int)glyph_row[glyph_x.as_int()] * color_alpha) / 255;
-            unsigned char* const buffer_pixels = (unsigned char *)(buffer_row + buffer_x.as_int());
+            unsigned char const glyph_alpha = (static_cast<int>(glyph_row[glyph_x.as_int()]) * color_alpha) / 255;
+            unsigned char* const buffer_pixels = reinterpret_cast<unsigned char *>(buffer_row + buffer_x.as_int());
```
AlanGriffiths: Using `reinterpret_cast<>` to alias memory is problematic. What would be better is a way to repack the pixel. (Analogous to `unpack_color()`

src/server/shell/decoration/decoration_strategy.cpp
```diff
@@ -618,14 +635,14 @@ void RendererStrategy::Text::Impl::render_glyph(
...
             for (int i = 0; i < 3; i++)
             {
                 // Blend color with the previous buffer color based on the glyph's alpha
                 buffer_pixels[i] =
-                    ((int)buffer_pixels[i] * (255 - glyph_alpha)) / 255 +
-                    ((int)color_pixels[i] * glyph_alpha) / 255;
+                    (static_cast<int>(buffer_pixels[i]) * (255 - glyph_alpha)) / 255 +
+                    (static_cast<int>(color_pixels[i]) * glyph_alpha) / 255;
```
AlanGriffiths: We unpacked in an endian aware manner, we need to repack in a consistent way

AlanGriffiths (changes requested):

AlanGriffiths (approved):

PR #4926, opened 2026-05-04
Simpler config override

tarek-y-ismail:
Related #4781, #4858, #4830 
## What's new?

- Merges `ConfigAggregator` and `IniFile`, producing a simpler implementation.

## How to test

- `<path/to/build>/bin/miral-test/miral-test --gtest_filter=IniFileWithOverridesTest*`

## Checklist

- [x] Tests added and pass


include/miral/miral/ini_file_with_overrides.h
```diff
@@ -0,0 +1,91 @@
...
+        HandleStrings handler) override;
+
+    void on_done(HandleDone handler) override;
+
+    /// Parse an ordered list of ini-file streams and dispatch all handlers once.
+    ///
+    /// The first element is treated as the base config; subsequent elements are
+    /// overrides in ascending priority order. A single transaction wraps the
+    /// entire load, so \c on_done fires exactly once per call.
+    void load(std::span<std::pair<std::unique_ptr<std::istream>, std::filesystem::path>> files);
```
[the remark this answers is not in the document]
tarek-y-ismail: We're likely to tweak this slightly to when updating `ConfigFile`, so I'm not gonna block on this.

src/miral/live_config_ini_file_with_overrides.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    auto const first_non_space = string.find_first_not_of(whitespace);
+    if (first_non_space == std::string::npos)
+    {
+        return string.substr(0, 0);
+    }
+
+    auto const last_non_space = string.find_last_not_of(whitespace);
+    return string.substr(first_non_space, last_non_space - first_non_space + 1);
+}
```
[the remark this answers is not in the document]
tarek-y-ismail: True, can be moved to a shared header/source. Will do that if this comes to pass.

include/miral/miral/ini_file_with_overrides.h
AlanGriffiths: This API has downsides:

1. It requires all the streams to be opened simultaneously: a potential resource bottleneck
2. There is no indication of what changed - which prevents future optimizations
3. It entrusts iteration over the files to to the load method.

How about something like:


```suggestion
    /// Parse an ordered list of ini-file streams and dispatch all handlers once.
    ///
    /// The first element is treated as the base config; subsequent elements are
    /// overrides in ascending priority order. A single transaction wraps the
    /// entire load, so \c on_done fires exactly once per call.
    void load(OverrideIterator iter);
```

Where `OverrideIterator` is something like:

```C++
class OverrideElement
{
public:
     enum class Status { unchanged, added, removed, updated };

     virtual auto status() const -> Status = 0;
     virtual auto filepath() const -> std::filesystem::path = 0;
   
     // Only valid when `status != removed` and while this is the current element 
     virtual auto stream() const -> std::istream& = 0;
};

class OverrideIterator
{
public:
    using difference_type = std::ptrdiff_t;
    using value_type = OverrideElement;

    value_type const& operator*() const;

    OverrideIterator& operator++();
    void operator++(int) { ++*this; }
};
```

src/miral/ini_file_with_overrides.cpp
```diff
@@ -0,0 +1,179 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "basic_store.h"
+
+#include <miral/ini_file_with_overrides.h>
+
```
AlanGriffiths, line 21: Wrong order

include/miral/miral/ini_file_with_overrides.h
AlanGriffiths: `OverrideIterator` also needs a default created sentinel value - or can we do this with ranges?

AlanGriffiths (changes requested):
Not sure what the right `load` signature is, but shared an idea

include/miral/miral/ini_file_with_overrides.h
tarek-y-ismail: I think a span and a class that opens streams lazily is a good middle ground. We still have to trust the load method to load in the proper order, but I don't see that as an issue.

tarek-y-ismail: Thinking more about it: Do we actually need anything other than access to the path and lazy access to the stream? Does `status` provide any extra info we need?

include/miral/miral/ini_file_with_overrides.h
```diff
@@ -0,0 +1,86 @@
...
+        std::string_view description,
+        std::span<std::string const> preset,
+        HandleStrings handler) override;
+
+    void on_done(HandleDone handler) override;
+
+    /// Parse an ordered list of file-change events and dispatch all handlers once.
+    ///
+    /// \c on_done fires exactly once per call.
+    void load(std::span<std::unique_ptr<FileChange const>> changes);
```
AlanGriffiths: What is `unique_ptr<>` used for? There should be no ownership transfer.

include/miral/miral/live_config.h
```diff
@@ -28,6 +30,22 @@
...
+public:
+    virtual auto filepath() const -> std::filesystem::path = 0;
+    virtual auto stream() const -> std::unique_ptr<std::istream> = 0;
+
+    virtual ~FileChange() = default;
+
+    FileChange() = default;
+    FileChange(FileChange const&) = delete;
+    FileChange& operator=(FileChange const&) = delete;
+};
```
AlanGriffiths: The way this is being used, it would probably be better using pimpl idiom, not interface class

include/miral/miral/live_config.h
```diff
@@ -28,6 +30,22 @@
...
 namespace miral::live_config
 {
+/// Describes a single file change event delivered to a config store.
+///
+/// \remark Since MirAL 5.9
+class FileChange
+{
+public:
+    virtual auto filepath() const -> std::filesystem::path = 0;
+    virtual auto stream() const -> std::unique_ptr<std::istream> = 0;
```
AlanGriffiths: How are you representing the change annotation? (Is the file unchanged? New? Modified? Deleted?)

We discussed the need to support the potential to optimize reloading in the interface, even if it isn't (yet) implemented

AlanGriffiths: How are you representing the change annotation? (Is the file unchanged? New? Modified? Deleted?)

We discussed the need to support the potential to optimize reloading in the interface, even if it isn't (yet) implemented

AlanGriffiths (changes requested):

include/miral/miral/live_config.h
tarek-y-ismail: I don't see a need for it right now. That can be added in the future and it's an implementation detail which would not break users.

include/miral/miral/live_config.h
tarek-y-ismail: How would pimpl help if we don't store any members?

AlanGriffiths: It means we don't have a vtable in the ABI

include/miral/miral/live_config.h
AlanGriffiths: How will you add it without changing the vtable and breaking ABI and users?

include/miral/miral/live_config.h
AlanGriffiths: There are members in the implementation in both designs, so why are you claiming there are not?

RAOF: Adding a new virtual function breaks ABI, whereas adding a new concrete function does not.

We're not using virtual dispatch for polymorphism¹ here — we're using it for implementation hiding — so having a concrete, but hidden, implementation provides us with more flexiblity.

¹: With the exception of testing, which can be accomplished by supplying a different pimpl.

RAOF (commented):
This is also reasonable, once Alan's comments are addressed.

AlanGriffiths:
Passing a `span<>` affords a lot of usage patterns we don't want or need to support. A input/forward iterator would be a clearer expression of the semantics.

RAOF:
> Passing a `span<>` affords a lot of usage patterns we don't want or need to support. A input/forward iterator would be a clearer expression of the semantics.

For what it's worth, I agree with this, but don't have as strong an opinion as Alan; I'm fine with `span<>`, though I do think the iterator would be better.

AlanGriffiths:
Let me offer another approach that avoids abusing `unique_ptr<>` to make `span` compile and to pretend that `istream` is copyable:

```c++

using FileLoader = std::movable_function<void(path, istream&)>;
using FileDropped= std::movable_function<void(path)>;

class FileChanges
{
public:

     virtual void for_each(FileLoader unchanged, FileLoader moved_or_new, FileLoader modified, FileDropped dropped) const = 0;
};

void load(FileChanges const& changes);
```

RAOF:
I continue to prefer https://github.com/canonical/mir/pull/4926#discussion_r3181417990 (although possibly with `OverrideElement` as a concrete pimpl rather than interface), but I don't think *any* of the options are fundamentally critically flawed — we need to be careful about vtables vs ABI and ownership, but otherwise I don't block any of them.

tarek-y-ismail:
I've explored the two suggestions and some others. Both should work, but I really couldn't choose between either so I just picked one randomly and massaged it into the code. Tell me what you think.

include/miral/miral/live_config.h
```diff
@@ -28,6 +30,73 @@
...
+    auto push_modified(
+        std::filesystem::path filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChanges&;
+
+    auto push_dropped(std::filesystem::path filepath) -> FileChanges&;
+
+    auto push_renamed(
+        std::filesystem::path old_filepath,
+        std::filesystem::path new_filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChanges&;
```
RAOF: Why are these in the public API?

include/miral/miral/live_config.h
```diff
@@ -28,6 +30,73 @@
...
+        FileRenamed renamed) const;
+
+    ~FileChanges() = default;
+    FileChanges(FileChanges const&) = default;
+    FileChanges& operator=(FileChanges const&) = default;
+
+private:
+    struct Context;
+    std::shared_ptr<Context> self;
+};
```
AlanGriffiths: These don't belong in this file

include/miral/miral/live_config.h
```diff
@@ -28,6 +30,73 @@
...
+        std::filesystem::path filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChanges&;
+
+    auto push_dropped(std::filesystem::path filepath) -> FileChanges&;
+
+    auto push_renamed(
+        std::filesystem::path old_filepath,
+        std::filesystem::path new_filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChanges&;
+
```
AlanGriffiths: None of these functions belong on this interface: they are for the producer, not the consumer

include/miral/miral/live_config.h
```diff
@@ -28,6 +30,73 @@
...
+/// Callback type for file-removal events.
+///
+/// \remark Since MirAL 5.8
+using FileDropped = std::move_only_function<void(std::filesystem::path const&)>;
+
+/// Callback type for file-rename events.
+///
+/// \remark Since MirAL 5.8
+using FileRenamed = std::move_only_function<void(std::filesystem::path const&, std::filesystem::path const&, std::istream&)>;
```
AlanGriffiths: These types do not belong at namespace scope they are part of the FileChanges abstraction

include/miral/miral/live_config.h
AlanGriffiths: `Context` vs `self`?

AlanGriffiths (changes requested):

src/miral/CMakeLists.txt
```diff
@@ -36,6 +36,8 @@ add_library(miral-internal STATIC
                                          window_info_defaults.h
     window_specification_internal.cpp    window_specification_internal.h
     basic_store.cpp                      basic_store.h
+    file_changes_builder.cpp             file_changes_builder.h
+    live_config.cpp                      ${miral_include}/miral/live_config.h
```
tarek-y-ismail: I'm not entirely sure about linking `live_config.cpp` internally. I had to do this because I needed to use `FileChanges` (external) in `FileChangesBuilder` (internal), which itself feels wrong. 

Another approach I had in mind was for `FileChanges` to be just an interface which is implemented in `FileChangesBuilder`, but that would make the ABI brittle.

src/miral/file_changes_builder.h
```diff
@@ -0,0 +1,59 @@
...
+
+    auto push_dropped(std::filesystem::path filepath) -> FileChangesBuilder&;
+
+    auto push_renamed(
+        std::filesystem::path old_filepath,
+        std::filesystem::path new_filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChangesBuilder&;
+
+    /// Implicit conversion so a builder can be passed directly to load(FileChanges const&).
+    operator FileChanges const&() const;
```
tarek-y-ismail: I'm not sure this is the best idea, you'd think that Claude would know that explicit >>> implicit by now...

include/miral/miral/live_config_file_changes.h
```diff
@@ -0,0 +1,71 @@
...
+{
+class FileChangesBuilder;
+
+/// A pimpl'd, ordered batch of file-system change events.
+///
+/// Callers build the batch with push_unchanged(), push_moved_or_new(),
+/// push_modified(), push_dropped(), and push_renamed(), then pass it to
+/// IniFileWithOverrides::load(). The store consumes the batch by calling
+/// for_each() with one callback per event kind.
+///
```
RAOF: This describes methods that don't exist on this interface? We should be documenting the public API, which is *only* `for_each`; how the implementation(s) populate `for_each` is irrelevant.

include/miral/miral/live_config_file_changes.h
```diff
@@ -0,0 +1,71 @@
...
+/// Events are dispatched in push order, so relative file priority is
+/// preserved across kinds.
+///
+/// \remark Since MirAL 5.8
+class FileChanges
+{
+    struct Context;
+    friend FileChangesBuilder;
```
RAOF: This looks like a leftover from when `FileChanges(std::shared_ptr<Context>)` wasn't public?

src/miral/live_config_file_changes_builder.h
```diff
@@ -0,0 +1,68 @@
...
+{
+public:
+    auto push_unchanged(
+        std::filesystem::path filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChangesBuilder&;
+
+    auto push_moved_or_new(
+        std::filesystem::path filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChangesBuilder&;
+
```
RAOF: How is “moved” different from renamed (below)?

RAOF (commented):
Yeah, looks fine; some nits.

include/miral/miral/live_config_file_changes.h
AlanGriffiths: No: Context is not public

tarek-y-ismail: This is needed because Context needs to be declared at some point before the public constructor. Additionally, we want to restrict it to only the builder. So it's declared before the ctor, and made private, with the friend declaration right after. The friend declaration could be moved to the explicit private section, but that would separate it from why it's actually used.

src/miral/live_config_file_changes_builder.h
tarek-y-ismail: As I see it, if a file was moved in, we don't really care about where it was before or what its previous name was. But if a file was renamed, users of this class might be interested in knowing its previous name for bookkeeping purposes.

include/miral/miral/live_config_file_changes.h
RAOF: Ah, yes. So why is the `FileChanges(std::shared_ptr<Context>)` constructor public?

src/miral/live_config_file_changes_builder.h
RAOF: Should this be `push_new` then (and, likewise, `new_or_moved` -> `new` in the `for_each` signature)?

RAOF (approved):

src/miral/live_config_file_changes_builder.h
```diff
@@ -0,0 +1,68 @@
...
+    auto push_modified(
+        std::filesystem::path filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChangesBuilder&;
+
+    auto push_dropped(std::filesystem::path filepath) -> FileChangesBuilder&;
+
+    auto push_renamed(
+        std::filesystem::path old_filepath,
+        std::filesystem::path new_filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChangesBuilder&;
```
AlanGriffiths: For bookkeeping we need to notify: `file001.settings` events before `file999.settings` regardless of which is the new or old name.

AlanGriffiths (changes requested):

src/miral/live_config_file_changes_builder.h
tarek-y-ismail: That's doesn't conflict with what I'm suggesting. Wouldn't it be necessary for cases where internal file state/representation is stored to know that `file001` was renamed to `file1001` for example?

src/miral/live_config_file_changes_builder.h
tarek-y-ismail: Not sure. Might be slightly confusing as it would hide the fact that this event could happen on moves. What do you think?

src/miral/live_config_file_changes.cpp
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+#include <miral/live_config_file_changes.h>
+#include "live_config_file_changes_builder.h"
+
+miral::live_config::FileChanges::FileChanges(std::shared_ptr<Context> ctx) :
+    ctx{std::move(ctx)}
+{
+}
+
```
tarek-y-ismail: We have to define this symbol here in `miral-internal`, or define it in `miral-external` and expose it in the symbol map. Since it's only used here so far, I decided to keep it in `miral-internal`,

src/miral/live_config_file_changes_builder.h
AlanGriffiths: `FileChangesBuilder` collects a list of files, not a list of events. I'm with Chris: we should be providing the filenames annotated with what, if anything, changed.

1. Whether a file was created, moved to, renamed to, or modified it is essentially "new" (something that needs parsing).
2. When a file was deleted, moved from, or renamed from, it is essentially "gone" (any cached parse results can be discarded).
3. When a file is unchanged, cached results can (in theory) be used instead of parsing the file.

src/miral/live_config_file_changes_builder.h
AlanGriffiths: I guess this is now moot. But no: we need two entries in the list:

0. `file000` (unchanged)
1. `file001` moved (to `file1001`)
2. `file007` (new)
3. `file1001` moved (from `file001`)

src/miral/live_config_file_changes.cpp
AlanGriffiths: We should expose it in the symbol map

src/miral/live_config_file_changes.cpp
```diff
@@ -0,0 +1,58 @@
...
+void miral::live_config::FileChanges::for_each(
+    Loader  unchanged,
+    Loader  moved_or_new,
```
RAOF: *Ideally* this name would match the prototype (`new_`)

src/miral/live_config_file_changes.cpp
```diff
@@ -0,0 +1,58 @@
...
+    {
+        switch (event.kind)
+        {
+        case Context::Kind::unchanged:
+        {
+            auto stream = event.stream_factory();
+            unchanged(event.path, *stream);
+            break;
+        }
+        case Context::Kind::moved_or_new:
```
RAOF: And, again, for consistency.

src/miral/live_config_file_changes.cpp
```diff
@@ -0,0 +1,58 @@
...
+    Loader  modified,
+    Dropped dropped) const
+{
+    for (auto const& event : ctx->events)
+    {
+        switch (event.kind)
+        {
+        case Context::Kind::unchanged:
+        {
+            auto stream = event.stream_factory();
```
RAOF: `stream_factory()` is going to be able to throw an exception, right? If we can't open the file (for whatever reason), we should catch that exception and turn this into a `removed()`, I think?

src/miral/live_config_file_changes.cpp
```diff
@@ -0,0 +1,58 @@
...
+            unchanged(event.path, *stream);
+            break;
+        }
+        case Context::Kind::moved_or_new:
+        {
+            auto stream = event.stream_factory();
```
RAOF: ...and here we should probably just *skip* if we can't open the file...

src/miral/live_config_file_changes.cpp
```diff
@@ -0,0 +1,58 @@
...
+            moved_or_new(event.path, *stream);
+            break;
+        }
+        case Context::Kind::modified:
+        {
+            auto stream = event.stream_factory();
```
RAOF: ...and here we should `dropped` again.

RAOF (commented):
The interface looks fine to me; a couple more nits in the implementation though.

src/miral/live_config_file_changes_builder.h
```diff
@@ -0,0 +1,62 @@
...
+{
+struct FileChanges::Context
+{
+    enum class Kind { unchanged, moved_or_new, modified, dropped };
+
+    struct Event
+    {
+        Kind kind;
+        std::filesystem::path path;
+        std::function<std::unique_ptr<std::istream>()> stream_factory;
```
AlanGriffiths: Passing stream ownership around like this is inelegant, but at least it isn't in the public API.

I think I'd prefer:

```suggestion
        std::function<void(std::function<void(std::istream&, std::filesystem::path const& path)> callback> file_opener;
```

AlanGriffiths (changes requested):
Just a little polish needed

src/miral/live_config_file_changes.cpp
tarek-y-ismail: Why? Isn't it meant for internal use only and shouldn't be part of the public ABI?

src/miral/live_config_file_changes.cpp
AlanGriffiths: That would be easier if (as suggested elsewhere) we pass a callback to the file_opener

src/miral/live_config_file_changes_builder.h
tarek-y-ismail: I think the existing signature is good. How does this improve on it?

src/miral/live_config_file_changes_builder.cpp
```diff
@@ -0,0 +1,59 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "live_config_file_changes_builder.h"
+
+miral::live_config::FileChanges::FileChanges(std::shared_ptr<Context> ctx) :
+    ctx{std::move(ctx)}
+{
+}
```
AlanGriffiths: This is in the wrong file

src/miral/live_config_file_changes_builder.h
AlanGriffiths: It:

1. Avoids passing ownership of the `istream`;
2. Avoids the need to store path in two places; and,
3. Allows problems opening the file to be handled (by skipping the callback)

src/miral/live_config_file_changes_builder.h
```diff
@@ -0,0 +1,62 @@
...
+    auto push_modified(
+        std::filesystem::path filepath,
+        std::function<std::unique_ptr<std::istream>()> stream) -> FileChangesBuilder&;
+
+    auto push_dropped(std::filesystem::path filepath) -> FileChangesBuilder&;
+
+    auto build() -> FileChanges;
+private:
+    using Context = FileChanges::Context;
+    std::shared_ptr<Context> ctx = std::make_shared<Context>();
```
AlanGriffiths:
```suggestion
    std::unique_ptr<Context> ctx = std::make_shared<Context>();
```

include/miral/miral/live_config_file_changes.h
```diff
@@ -0,0 +1,57 @@
...
+    void for_each(Loader unchanged, Loader new_, Loader modified, Dropped dropped) const;
+
+    ~FileChanges() = default;
+    FileChanges(FileChanges const&) = default;
+    FileChanges& operator=(FileChanges const&) = default;
+
+private:
+    struct Context;
+    friend FileChangesBuilder;
+    explicit FileChanges(std::shared_ptr<Context> ctx);
```
AlanGriffiths: Why is this private? We have already enough access control by making `Context` private.

```suggestion
public:
    explicit FileChanges(std::unique_ptr<Context> ctx);
```

include/miral/miral/live_config_file_changes.h
```diff
@@ -0,0 +1,57 @@
...
+
+    /// Callback type for file-removal events.
+    using Dropped = std::move_only_function<void(std::filesystem::path const&)>;
+
+    /// Iterate all events in push order, invoking the matching callback for each.
+    void for_each(Loader unchanged, Loader new_, Loader modified, Dropped dropped) const;
+
+    ~FileChanges() = default;
+    FileChanges(FileChanges const&) = default;
+    FileChanges& operator=(FileChanges const&) = default;
```
AlanGriffiths: These should probably be defined as non-inline to avoid random symbols

include/miral/miral/live_config_file_changes.h
AlanGriffiths: We could add a precondition that `ctx != nullptr`

include/miral/miral/live_config_ini_file_with_overrides.h
```diff
@@ -0,0 +1,90 @@
...
+
+namespace miral::live_config
+{
+class FileChanges;
+
+/// An ini-file-based live configuration store that aggregates values from
+/// multiple files (a base config plus zero or more override files).
+///
+/// Scalar keys follow last-writer-wins semantics: if the same key appears in
+/// a later (higher-priority) file, it replaces the earlier value.
```
AlanGriffiths: "Scalar" is poor terminology. It implies something that is represented by a single value. (speed as opposed to velocity, or a string)

AlanGriffiths (changes requested):
The blocking problem is publishing the `FileChanges` constructor. It is public, the symbol should be available.

At some point in the future we have the option of providing a way to supply a `std::unique_ptr<Context>` through a public builder. There's no need to cause confusion by having a public ctor with no corresponding symbol

include/miral/miral/live_config_ini_file_with_overrides.h
```diff
@@ -0,0 +1,90 @@
...
+#define MIRAL_INI_FILE_WITH_OVERRIDES_H
+
+#include <miral/live_config.h>
+
+#include <memory>
+#include <span>
+
+namespace miral::live_config
+{
+class FileChanges;
```
AlanGriffiths: Maybe this isn't the best name. It is more a list of files in priority order than a list of changes.

`AnnotatedFileList` seems a bit wordy though

AlanGriffiths: Over breakfast Chris and I came up with `OverridesList`

Saviq: Will just throw "drop-in" as a term that can maybe be used here?

tarek-y-ismail: I'm not really for or against any name in particular

include/miral/miral/live_config_overrides_list.h
```diff
@@ -0,0 +1,59 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIRAL_LIVE_CONFIG_OVERRIDES_LIST_H
+#define MIRAL_LIVE_CONFIG_OVERRIDES_LIST_H
+
+#include <filesystem>
+#include <functional>
+#include <istream>
```
AlanGriffiths:
```suggestion
#include <iosfwd>
```

include/miral/miral/live_config_overrides_list.h
```diff
@@ -0,0 +1,59 @@
...
+#include <memory>
+
+namespace miral::live_config
+{
+class OverridesListBuilder;
+
+/// An ordered list of configuration file overrides to be applied.
+/// Each event represents a file that should be loaded into the configuration
```
AlanGriffiths: Not convinced by "event". Maybe "callback"?

include/miral/miral/live_config_overrides_list.h
```diff
@@ -0,0 +1,59 @@
...
+    explicit OverridesList(std::unique_ptr<Context> ctx);
+
+    /// Callback type for file-content events (unchanged, moved/new, or modified).
+    using Loader = std::move_only_function<void(std::filesystem::path const&, std::istream&)>;
+
+    /// Callback type for file-removal events.
+    using Dropped = std::move_only_function<void(std::filesystem::path const&)>;
+
+    /// Iterate all events in push order, invoking the matching callback for each.
+    void for_each(Loader unchanged, Loader new_, Loader modified, Dropped dropped) const;
```
AlanGriffiths: Not convinced by `new_`> maybe `fresh`?

src/miral/live_config_ini_file_with_overrides.cpp
```diff
@@ -0,0 +1,180 @@
...
+                            auto const key = Key{trim_start_and_end(line_view.substr(0, eq))};
+                            auto const value = trim_start_and_end(line_view.substr(eq + 1));
+
+                            update_key(key, value, path);
+                        }
+                        catch (std::exception const& e)
+                        {
+                            mir::log_warning("Error processing '%s': %s", path.c_str(), e.what());
+                        }
+                }
```
AlanGriffiths: Some genuine commonality with `IniFile::load_file()` - would be good to factor out to avoid them getting out of step

AlanGriffiths (approved):
This looks fine. (A few optional nits)

BUT we should sort out the git history before landing

AlanGriffiths (approved):

PR #4927, opened 2026-05-04
miral: Make std::formatter<live_config::Key> a strong symbol

RAOF:
This ensures the symbol reliably exists, so the symbols check reliably passes.

### DIFF
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -808,6 +808,7 @@ global:
     typeinfo?for?miral::live_config::OverridesList;
     typeinfo?for?miral::live_config::IniFileWithOverrides;
     vtable?for?miral::live_config::IniFileWithOverrides;
-    "std::formatter<miral::live_config::Key, char>::format(miral::live_config::Key const&, std::basic_format_context<std::__format::_Sink_iter<char>, char>&) const";
+    "std::formatter<miral::live_config::Key,char>::format*";
+    "typeinfo?for?std::formatter<miral::live_config::Key,char>";
   };
 } MIRAL_5.7;
--- a/src/miral/symbols.map
+++ b/src/miral/symbols.map
@@ -808,7 +808,7 @@ global:
     typeinfo?for?miral::live_config::OverridesList;
     typeinfo?for?miral::live_config::IniFileWithOverrides;
     vtable?for?miral::live_config::IniFileWithOverrides;
-    "std::formatter<miral::live_config::Key,char>::format*";
-    "typeinfo?for?std::formatter<miral::live_config::Key,char>";
+    std::formatter?miral::live_config::Key*char?::format*;
+    typeinfo?for?std::formatter?miral::live_config::Key*char?;
   };
 } MIRAL_5.7;


include/miral/miral/live_config.h
```diff
@@ -109,10 +110,8 @@ class Store
 template<>
 struct std::formatter<miral::live_config::Key> : std::formatter<std::string>
 {
-    auto format(miral::live_config::Key const& key, std::format_context& ctx) const
-    {
-        return std::formatter<std::string>::format(key.to_string(), ctx);
-    }
+        auto format(miral::live_config::Key const& key, std::format_context& ctx) const
+            -> decltype(std::declval<std::formatter<std::string>>().format(std::declval<std::string>(), std::declval<std::format_context&>()));
```
Saviq: Indent should be this, no?
```suggestion
    auto format(miral::live_config::Key const& key, std::format_context& ctx) const
        -> decltype(std::declval<std::formatter<std::string>>().format(std::declval<std::string>(), std::declval<std::format_context&>()));
```

RAOF: :woman_facepalming:

RAOF:
Bah. Of course, this also relies on the symbols checker updates to correctly handle templated symbols (@robert-ancell ?).

tarek-y-ismail:
@RAOF CI still doesn't pass :/ Is that related to Robert's fix?

tarek-y-ismail:
Ok, tried vibing a solution. Maybe it doesn't fit in this PR, but it will be useful to know whether or not CI passes.

tarek-y-ismail:
As usual, ld proves to be problematic :man_facepalming:. Let's see if the other way works in CI as it worked locally.

Saviq:
@raof this blends, what say you?

tarek-y-ismail:
@Saviq @RAOF Shall we merge?

RAOF:
#4930 is also an attempt at fix for this; I'll see which one is better.

RAOF:
I think #4930 generates better symbols; I've merged that and will rebase this.

tarek-y-ismail:
Sorry Chris, hijacking this PR (again).

tarek-y-ismail:
@RAOF @robert-ancell I'm not sure #4930 fixes our problem. We can't use quoted strings _and_ wildcards at the same time. Unquoting the strings still fails on the first `<`. So the new output from the symbol generator still fails to link.

Edit: Using `?` instead of angled brackets and `*` instead of commas (probably due to whitespace issues)` seems to work.

tarek-y-ismail:
The latest commit should hopefully work. I'm not sure if I was testing incorrectly. If not, maybe we can update the symbol checker? I still very much dislike using wildcards but I'm not sure if we can support all of ld, lld, and mold without it.

tarek-y-ismail:
I've had this idea for a while: instead of suffering with inadequate formats that are broken in different ways for different linkers, what if we have a sane central format that we generate the rest from? Obligatory [xkcd 927](https://xkcd.com/927/) reference.

RAOF:
> I've had this idea for a while: instead of suffering with inadequate formats that are broken in different ways for different linkers, what if we have a sane central format that we generate the rest from? Obligatory [xkcd 927](https://xkcd.com/927/) reference.

Now that the symbols file is a generated artefact rather than something we write by hand we *could* just generate the mangled symbols (rather than sticking them in an `extern "C++"` block). Those are *guaranteed* to work, as otherwise the linkers would generate ABI-incompatible outputs. There are tricks here — we will inevitably run into architecture-dependant symbols that are currently swept under a wildcard rug in the `symbols.map` file, but we already have to deal with them in the `debian/*.symbols` files so it shouldn't be a huge issue.

AlanGriffiths (approved):
Mostly harmless

PR #4934, opened 2026-05-07
Symbols map generator performance improvements

tarek-y-ismail:
## What's new

Implements two performance optimizations for the symbols map generator:
- Instructs libclang to skip function bodies (unnecessary for symbols)
  - This exposed a bug where we accidentally emitted symbols for inline functions after the optimization was applied. The changes to `is_function_inline` correctly check for inline-ness.
- Parses all headers in parallel

Depending on your hardware, expect performance gains of ~10x+. Below are the results on my machine:
 | Variant                        | Median  | vs baseline | Speedup |
 |-------------------------------|---------|-------------|---------|
 | Baseline (serial, full parse)  | 33.94s  | —           | 1×      |
 | Skip-bodies only (serial)      | 23.67s  | −10.27s     | 1.4×    |
 | Parallel only (full parse)     |  3.50s  | −30.44s     | 9.7×    |
 | Combined                       |  2.29s  | −31.64s     | 14.8×   |
 
 A few more optimizations were explored along the way but were dropped due to negligible effect (< 5% improvement). Correctness was verified by quickly vibing a test suite, which did discover a couple of new (now solved) and pre-existing bugs. That will follow in a future PR.
 
 ## How to test:
 Check for new/removed symbols (diff mode — read-only):
```
 cmake --build <path to build> --target check-miral-symbols-map
 cmake --build <path to build> --target check-miroil-symbols-map
````
Update symbols.map in-place with any new symbols:
```
 cmake --build <path to build> --target generate-miral-symbols-map
 cmake --build <path to build> --target generate-miroil-symbols-map
 ```

### DIFF
--- a/tools/symbols_map_generator/main.py
+++ b/tools/symbols_map_generator/main.py
@@ -187,10 +187,20 @@ def has_virtual_base_class(node: clang.cindex.Cursor):
 
 def is_function_inline(node: clang.cindex.CursorKind):
     # This method assumes that the node is a FUNCTION_DECL.
-    # There is no explicit way to check if a function is inlined
-    # but seeing that if it is a FUNCTION_DECL and it has
-    # a definition is good enough.
-    return node.is_definition()
+    # Use libclang's dedicated query rather than node.is_definition(): the
+    # latter is only a proxy for "inline" and, crucially, reports False once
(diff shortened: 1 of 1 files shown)


mattkae (approved):
Neat! Nice optimizations

PR #4935, opened 2026-05-08
Track what type of data offer is being provided and validate requests.

robert-ancell:


robert-ancell:
Came out of review of https://github.com/canonical/mir/pull/4893

src/server/frontend_wayland/wl_data_device.cpp
```diff
@@ -93,11 +98,25 @@ class mf::WlDataDevice::Offer : public wayland::DataOffer
 
     void finish() override
     {
+        if (type != OfferType::dnd) {
+            throw mw::ProtocolError(
+                resource,
+                Error::invalid_offer,
+                "finish is only valid for drag and drop offers");
```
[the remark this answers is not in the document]
robert-ancell: Newer code was decided to not use BOOST_THROW_EXCEPTION in this case.

AlanGriffiths (changes requested):
Nit: Not the project brace style

src/server/frontend_wayland/wl_data_device.cpp
```diff
@@ -132,21 +150,23 @@ class mf::WlDataDevice::Offer : public wayland::DataOffer
...
     device{device},
     source{source}
 {
     device->send_data_offer_event(resource);
-    for (auto const& type : source->mime_types())
+    for (auto const& mime_type : source->mime_types())
     {
-        send_offer_event(type);
+        send_offer_event(mime_type);
     }
```
tarek-y-ismail, line 173: Not completely necessary imo

tarek-y-ismail (changes requested):
I agree with Alan that the brace style needs fixing. I also think the diff can be made a bit smaller. Other than that, this matches the protocol.

src/server/frontend_wayland/wl_data_device.cpp
robert-ancell: Changed because it would otherwise shadow the new OfferType parameter.

AlanGriffiths (approved):

tarek-y-ismail (approved):

Issue #4710, opened 2026-02-23
Add toplevel capture support through the image capture protocol


Saviq:
> At this point, we have a partial implementation of the new protocol, but toplevel/window capture support is not yet present. That's the main gap for closing this ticket. 

 _Originally posted by @Conan-Kudo in [#3659](https://github.com/canonical/mir/issues/3659#issuecomment-3939225281)_

> Toplevels are a new feature, which happens to require quite some thinking (and a pre-composition pass for subsurfaces at least, but maybe also children?). 

 _Originally posted by @Saviq in [#3659](https://github.com/canonical/mir/issues/3659#issuecomment-3943602388)_

PR #4936, opened 2026-05-08
Add support for screen-shooting a single surface

jhenstridge:
Related: #4710

## What's new?

This adds some infrastructure necessary to implement toplevel capture via `ext_image_copy_capture_v1`. In particular:

1. Add a `mir::compositor::Scene` implementation that only includes the renderables for a single `mir::scene::Surface`.
2. Extend `mir::compositor::ScreenShooterFactory` to allow creating screen shooters for an alternative scene.

Together these will let us implement an `ExtImageCopyBackend` for toplevel capture:

1. Create a screen shooter for the toplevel's Surface.
2. Watch for size changes on the surface to set buffer constraints
3. Watch for damage on the surface rather than on the full surface stack.
4. Screenshot the rectangle where the surface is located in the scene.

One open question is what resolution to capture at. The simplest would be to just capture at 1x. If there are no subsurfaces, a better choice would be to capture at that surface's native resolution. If there are subsurfaces, there's no guarantee that they will have the same scale factor.

## How to test

This doesn't add the Wayland protocol pieces to make use of these changes. I hacked it into the output capture protocol to use a screen shooter for the surface at (100, 100) when the session is started:

https://github.com/user-attachments/assets/4bb1f8e2-4478-4c0f-91d2-813f761ebc91

Here you can see a single window is captured by wayvnc, and continues to render when occluded. For X11 apps, the decorations were a separate surface to the application content. That might need some additional work to support if we care.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/scene/single_surface_scene.cpp
+++ b/src/server/scene/single_surface_scene.cpp
@@ -1,0 +1,168 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <mir/scene/surface_scene.h>
+#include <mir/compositor/scene.h>
+#include <mir/compositor/scene_element.h>
+#include <mir/scene/observer.h>
+#include <mir/scene/surface.h>
+#include "rendering_tracker.h"
+#include "surface_stack.h"
+
+#include <boost/throw_exception.hpp>
+
+namespace ms = mir::scene;
+namespace mc = mir::compositor;
+namespace mg = mir::graphics;
+
+namespace mir::scene
+{
+class SingleSurfaceScene :
+    public compositor::Scene
+{
+public:
+    SingleSurfaceScene(std::weak_ptr<Surface> const& surface);
+
+    // From Scene
+    auto scene_elements_for(mc::CompositorID id) -> mc::SceneElementSequence override;
+    void register_compositor(mc::CompositorID id) override;
+    void unregister_compositor(mc::CompositorID id) override;
+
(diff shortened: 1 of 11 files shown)


src/server/scene/single_surface_scene.cpp
```diff
@@ -0,0 +1,137 @@
...
+    {
+        for (auto& renderable : locked->generate_renderables(id))
+        {
+            elements.emplace_back(
+                std::make_shared<SurfaceSceneElement>(
+                    locked->name(),
+                    renderable,
+                    rendering_tracker,
+                    id));
+        }
```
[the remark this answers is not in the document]
jhenstridge: We want to render toplevels that have been hidden, so this is fine. A test is included now.

src/server/scene/surface_stack.cpp
```diff
@@ -717,3 +718,73 @@ void ms::SessionLockObserverMultiplexer::on_unlock()
...
+
+    o->end_observation();
+
+    observers.remove(o);
+}
+
+auto ms::create_surface_scene(std::shared_ptr<Surface> const& surface) -> std::shared_ptr<mc::Scene>
+{
+    return std::make_shared<SingleSurfaceScene>(surface);
+}
```
AlanGriffiths: We're dumping too much stuff in `surface_stack.cpp`, it would make sense for `SingleSurfaceScene`/`create_surface_scene()` to have its own file.

src/server/scene/surface_stack.h
```diff
@@ -173,6 +173,32 @@ class SurfaceStack :
...
+    SingleSurfaceScene& operator=(const SingleSurfaceScene&) = delete;
+
+    RecursiveReadWriteMutex guard;
+    std::weak_ptr<Surface> surface;
+    std::shared_ptr<RenderingTracker> rendering_tracker;
+    std::set<compositor::CompositorID> registered_compositors;
+
+    Observers observers;
+};
+
```
AlanGriffiths: This isn't a logical part of surface_stack is it?

AlanGriffiths (changes requested):
Code looks reasonable, just deserves its own files

src/server/scene/surface_stack.cpp
jhenstridge: I had organised things that way first, but there was more shared code than not (SurfaceSceneElement and the Observers helper classes).

I originally had it split out into it's own file with a copy of the small SurfaceSceneElement class, which the copilot review complained about the duplicated code and `surface_stack.h` include, so I organised things this way. I can switch back to the original organisation if that would be better.

AlanGriffiths: I'd split out the reused code into its own files

AlanGriffiths (approved):
Approving in case splitting the code is too painful

PR #4937, opened 2026-05-08
miral: Add `register_log_filtering_config` convenience helper.

RAOF:


src/miral/logging_config.cpp
```diff
@@ -0,0 +1,65 @@
...
+    auto equals_pos = value.find('=');
+    if (equals_pos == value.length() || equals_pos == value.npos)
+    {
+      BOOST_THROW_EXCEPTION((std::runtime_error{
+          std::format("Failed to parse log-level: {}", value)}));
+    }
+    return std::make_pair(
+        value.substr(0, equals_pos),
+        value.substr(equals_pos + 1));
+}
```
tarek-y-ismail: We probably split strings on delimiters in other places in the code base. `IniFile` off the top of my head. Out of scope, but it may be worth it to have a small `mir/strings` module for stuff like this.

tarek-y-ismail (approved):
I only reviewed the latest commit as it's the only one that seems relevant.

Saviq:

src/miral/logging_config.cpp
RAOF: Yeah, we might have enough for a mir-utils internal-only static library to collect code that isn't relevant to API but that we want to use in multiple places.

src/miral/logging_config.cpp
```diff
@@ -0,0 +1,72 @@
...
+                mir::logging::tag::set_severity("core", mir::logging::Severity::warning);
+
+                for (auto const& filter : *value)
+                {
+                    auto const [tag_name, severity_name] = split_assignment(filter);
+                    auto const severity = mir::logging::parse_severity(severity_name);
+                    mir::logging::tag::set_severity(tag_name, severity);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
RAOF: Urgh, that's another time I've missed updating “core” → “base”. Possibly we should add `ml::tag::set_severity(Tag const&, Severity)`.

Fair call with unset-means-set-to-defaults; fixed.

I'll leave it up to others if “apply as much of a broken config as possible” is the right behaviour.

src/miral/logging_config.cpp
```diff
@@ -0,0 +1,73 @@
...
+            std::format("Failed to parse log-level: {}", config_value)}));
+    }
+    return std::make_pair(key, value);
+}
+}
+
+void miral::register_log_filtering_config(live_config::Store &config_store)
+{
+    config_store.add_strings_attribute(
+        {"log-level"},
```
AlanGriffiths: This is not a valid key:

```c++
    /// within the key is non-empty, starts with [a..z], and contains only [a..z,0..9,_]
```

We've used underscore separators for live config (to avoid complicating a hypothetical YAML implementation of Store)

src/miral/logging_config.cpp
```diff
@@ -0,0 +1,73 @@
...
+namespace
+{
+auto split_assignment(std::string_view config_value)
+    -> std::pair<std::string_view, std::string_view>
+{
+    auto equals_pos = config_value.find('=');
+    if (equals_pos == config_value.npos)
+    {
+      BOOST_THROW_EXCEPTION((std::runtime_error{
+          std::format("Failed to parse log-level: {}", config_value)}));
```
AlanGriffiths: I'd be inclined to use `key` rather than hard coding "log-level" in multiple places.

AlanGriffiths (changes requested):
> I'll leave it up to others if “apply as much of a broken config as possible” is the right behaviour.

Existing practice with live changes is “apply as much of a broken config as possible”. For example: https://github.com/canonical/mir/blob/ded2dd8bb138b309ca125cec8924a9a4b6266faa/src/miral/input_configuration.cpp#L118-L121

include/miral/miral/logging_config.h
```diff
@@ -0,0 +1,34 @@
...
+
+namespace miral
+{
+namespace live_config { class Store; }
+
+/// Hook up log filtering configuration to a `live_config::Store`
+///
+/// This is a convenience function that registers a log filtering
+/// option on the live config store and monitors it for changes.
+/// \remark Since MirAL 5.8
```
AlanGriffiths:
```suggestion
/// \remark Since MirAL 5.9
```

AlanGriffiths: And... doesn't this need to be a symbol?

RAOF: It does indeed!

AlanGriffiths (approved):

AlanGriffiths:
Obviously unrelated failure on rawhide (but why?):

```
DEBUG: RPM build errors:
DEBUG:     %changelog not in descending chronological order
```

Saviq:
> (but why?):

Was hiding behind #5017

PR #4941, opened 2026-05-08
DecorationProvider example: Don't constrain character size

AlanGriffiths:
With HIDPI screens limiting the maximum number of physical pixels a character can take is annoying: the text is small.


tarek-y-ismail (approved):
I wonder how we came up with the rest of the numbers. That's unrelated though

AlanGriffiths:
> I wonder how we came up with the rest of the numbers. That's unrelated though

IIRC they are properties of the displayed text (or were when they were written)

PR #4948, opened 2026-05-17
events: add support for switch events

muhammad23012009:
* currently these events are not handled by Mir itself. QtMir is the only user, with an event filter to steal these switch events.

## What's new?

Adds support for processing switch events from libinput

## Checklist

- [ ] Tests added and pass


include/common/mir/events/input_event.h
```diff
@@ -19,6 +19,8 @@
 
 #include <mir/events/event.h>
 
+typedef struct MirSwitchEvent MirSwitchEvent;
+
```
AlanGriffiths: I know the input headers are a recursive mess, but other types are declared in `include/common/mir_toolkit/events/event.h`

include/core/mir_toolkit/events/enums.h
```diff
@@ -47,6 +47,7 @@ typedef enum {
     mir_input_event_type_touch = 1,
     mir_input_event_type_pointer = 2,
     mir_input_event_type_keyboard_resync = 3,
+    mir_input_event_type_switch = 4,
```
AlanGriffiths: FLAGGING FOR ATTENTION: Extending this switch is an ABI break: downstream code needs to be recompiled. (That's not blocking this PR, but we need to do deal with it before a release.)

include/platform/mir/input/event_builder.h
```diff
@@ -113,6 +113,10 @@ class EventBuilder
         std::optional<Timestamp> timestamp,
         std::vector<mir::events::TouchContact> const& contacts) = 0;
 
+    virtual EventUPtr switch_event(
```
AlanGriffiths: FLAGGING FOR ATTENTION: Adding a virtual function to this interface is an ABI break: downstream code needs to be recompiled and implement this function. (That's not blocking this PR, but we need to do deal with it before a release.)

Note also: as we're breaking ABI, there's a bunch of `[[deprecated]]` functions we can get rid of.

AlanGriffiths (changes requested):
The approach looks sensible.

Copilot has identified a bunch of housekeeping (missing license headers, builder parameters etc). These need addressing. (Technically, copyright doesn't need to be assigned to Canonical as long as the CLA is in place.)

This does involve ABI breakages, so we also need to bump .sonames etc. (but the Mir team can deal with that separately.)

There are, however, parallel APIs in libmiral (`include/miral/miral/toolkit_event.h`) and, potentially, libmiroil (`include/miroil/miroil/input_device.h`) that need to reflect these changes.

include/common/mir/events/input_event.h
muhammad23012009: I'll move it out to that header

muhammad23012009:
@AlanGriffiths should I bother with the timestamps for switch events? They aren't used anywhere else right now, and definitely won't be passed to Wayland apps, so we don't need the timestamp for it

AlanGriffiths:
@muhammad23012009 I don't see a pressing need for timestamps. On the other hand, if we were ever to add them that would be an ABI change. So, perhaps better to have them.

The ABI change to mircore is, however, a possible concern. As there is another such change coming soon, can we hold off landing this until we can do both in the same release? (That's looking like 2.28)

include/common/mir/events/event_builders.h
```diff
@@ -187,6 +187,11 @@ EventUPtr make_touch_event(
     MirInputEventModifiers modifiers,
     std::vector<TouchContact> const& contacts);
 
+EventUPtr make_switch_event(
+    MirInputDeviceId device_id,
+    MirSwitchAction action,
+    MirSwitchState state);
```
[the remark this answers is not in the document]
AlanGriffiths: You can ignore this

src/platforms/evdev/libinput_device.cpp
```diff
@@ -366,6 +369,22 @@ mir::EventUPtr mie::LibInputDevice::convert_touch_frame(libinput_event_touch* to
...
+
+    switch (libinput_switch_type)
+    {
+        case LIBINPUT_SWITCH_TABLET_MODE:
+            return builder->switch_event(
+                mir_switch_action_tablet_mode,
+                libinput_switch_state == LIBINPUT_SWITCH_STATE_ON ? mir_switch_state_on : mir_switch_state_off);
+        default:
+            return {nullptr, [](auto){}};
+    }
```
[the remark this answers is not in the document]
AlanGriffiths: Yeah, I guess input events are expected to have timestamps

AlanGriffiths (changes requested):
Most of the copilot review comments are valid. If you can address those, we can now land this

AlanGriffiths:
@muhammad23012009 as there has been no activity for a while, I'm marking this as "draft" until you have an opportunity to address the issues raised in review. Feel free to mark as "ready" when it is ready

PR #4957, opened 2026-05-21
fix-wrong-foreign-toplevel-logic

### LINKED ISSUE #4944: Wrong ForeignToplevelHandleV1 creation logic
We should be creating handles for all application toplevel windows and no other windows but:

* ~We are sending handles with blank app ids and window names (doesn't make sense)~
* We are sending handles for overlay dialogs (which are not apps)
* We are not sending handles for apps that are docked

AlanGriffiths:
Closes #4944

## What's new?

Docked application windows are not skipped, system dialogs in the overlay layer are skipped.

## How to test

The app switcher in miral-shell uses these protocols (as does the Frame launcher)

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/desktop_file_manager.cpp
+++ b/src/server/frontend_wayland/desktop_file_manager.cpp
@@ -69,6 +69,10 @@ std::string mf::DesktopFileManager::resolve_app_id(scene::Surface const& surface
     // For more info on the checks happening here, see:
     // https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/src/shell-window-tracker.c?ref_type=heads#L387
 
+    // When packaged as a snap or flatpack the resulting desktop entry may not follow
+    // the naming style expected by the app. So we find the name based on the packaging
+    // and ignore the name provided by the application.
+    // https://github.com/canonical/mir/issues/4954#issuecomment-4510527223
     if (auto session = surface.session().lock())
     {
         auto pid = session->process_id();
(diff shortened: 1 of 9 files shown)


AlanGriffiths:
I've based https://github.com/Miriway/Miriway/pull/201 off this. So it can be tested with:

    snap refresh miriway --channel=edge/pr201

src/server/frontend_wayland/desktop_file_manager.cpp
```diff
@@ -60,21 +60,41 @@ mf::DesktopFileManager::DesktopFileManager(std::shared_ptr<DesktopFileCache> cac
...
     // For more info on the checks happening here, see:
     // https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/src/shell-window-tracker.c?ref_type=heads#L387
-    auto app_id = surface->application_id();
+
+    if (auto session = surface.session().lock())
+    {
+        auto pid = session->process_id();
+        auto socket_fd = session->socket_fd();
+
+        // First, check if the window belongs to snap
```
mattkae, line 81: I noticed that some reordering happened here! Can you leave a comment explaining why this reordering is necessary, if at all.

src/server/frontend_wayland/foreign_toplevel_manager_v1.cpp
```diff
@@ -351,39 +353,12 @@ void mf::ForeignSurfaceObserver::with_toplevel_handle(
 
 void mf::ForeignSurfaceObserver::create_or_close_toplevel_handle_as_needed(std::lock_guard<std::mutex>& lock)
 {
-    bool should_have_handle = true;
+    bool should_have_handle = false;
 
     auto const surface = weak_surface.lock();
```
mattkae:
```suggestion
    auto const surface = weak_surface.lock();
    auto const should_have_handle = surface ? should_create_foreign_toplevel_handle(*surface) : false;
```

mattkae (approved):
Two nits, but looks reasonable. I would call out in a comment why the order is important though in the event that it is.

PR #4960, opened 2026-05-26
Refactoring the evdev-rs platform to remove the blocking future await when grabbing a file descriptor for a device

mattkae:
## Problem
Previously, we were waiting for input device acquisition because the Rust libinput API demanded that we did so. This would lead us to await a `std::future` before proceeding. While this worked, it unnecessarily introduced a lot of complication, especially around the initial processing of input events.

## What's new?
- Modified `PlatformRs` to init via `new_from_path` instead of `new_with_udev`
- Introduced `udev_monitor.rs` which manually listens for udev events (mirror of the existing C++ class)
- Improved the event deferring logic
- Improved the lifetime management of udev devices

Here is the new device lifetime flow:
- Udev monitor is started and a device is "added" (`UdevEventType.Added`)
- `PlatformBridge::acquire_device` is triggered
- An `InputDeviceObserver` is created, which gets notified when the device has been acquired
- On acquisition, `PlatformRs.on_device_activated` is called, which calls `Libinput.path_add_device`
- `LibinputInterface.open_restricted` gets called, which reaches out to the bridge to claim the file descriptor that was set pending earlier
- As libinput events happen, `process_libinput_events` is getting called.
  - If the device has pending events (i.e. events that happened _before_ the device was acquired), we call them immediately
  - Otherwise, we process the incoming events
  - If the device is NOT yet set up to receive events, we defer the events for later

## How to test
```sh
# build
WAYLAND_DISPLAY=wayland-98 miral-shell --platform-path=/path/to/mir/build/lib/server-modules --platform-input-lib=mir:evdev-input-rs
```

Afterward, try some key events immediately. Move the mouse a bunch. VT switch. Suspend your session and open it again. Ensure that everything works :tada: 

## Checklist

- [x] Tests added and pass

## Related
- https://github.com/canonical/mir/pull/4780
- https://github.com/canonical/mir/pull/4791


src/platforms/evdev-rs/src/event_processing.rs
RAOF, line 155: It is unfortunate that we still need to do this, but we do.

src/platforms/evdev-rs/src/libinput_interface.rs
```diff
@@ -17,58 +17,47 @@
...
-    // This method is called when libinput is done with a device.
-    fn close_restricted(&mut self, fd: io::OwnedFd) {
-        let fd_raw = fd.as_raw_fd();
-        let _ = unistd::close(fd);
-        self.fds.retain(|d| d.raw_fd() != fd_raw);
-    }
+    /// Called by libinput when it is done with a device fd.
+    ///
+    /// The `OwnedFd` closes the file descriptor automatically when dropped.
+    fn close_restricted(&mut self, _fd: io::OwnedFd) {}
```
RAOF: This is susceptible to the problem we fixed in #4853 — because we're monitoring the devices, *we* want to be in charge of when the FDs are closed.

I think you could actually do this via something like:

```suggestion
    /// Called by libinput when it is done with a device fd.
    ///
    /// The `OwnedFd` closes the file descriptor automatically when dropped.
    fn close_restricted(&mut self, fd: io::OwnedFd)
    {
        // We handle FD externally, from UDev/the ConsoleProvider
        mem::forget(fd);
    }
```

RAOF (changes requested):
Dropping this partial review; I still need to go through the rest.

I wonder whether we shouldn't move the dispatch thread into the Rust side…

src/platforms/evdev-rs/platform.cpp
```diff
@@ -31,43 +31,111 @@
...
+        if (duped < 0)
+        {
+            mir::log_error("evdev-rs: dup() failed in activated() for %s", devnode.c_str());
+            return;
+        }
+
+        mir::log_info("evdev-rs: enqueuing on_device_activated for %s (duped fd=%d)", devnode.c_str(), duped);
+        device_queue->enqueue(
+            [devnode = devnode, devnum = devnum, fd = duped, &platform_impl = platform_impl]()
+            {
```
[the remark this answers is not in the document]
RAOF: Yeah, Copilot is correct here; you should move `fd` into the lambda and `dup` in there, so that we don't leak fds in case the lambda is dropped before being run.

src/platforms/evdev-rs/platform.cpp
```diff
@@ -31,43 +31,111 @@
...
+            });
     }
 
 private:
-    rust::Box<Impl> box;
-    std::optional<mir::Fd> fd;
+    std::string devnode;
+    uint64_t devnum;
+    std::shared_ptr<md::ActionQueue> device_queue;
+    rust::Box<miers::PlatformRs>& platform_impl;
```
RAOF: This is extremely weird (like storing a `std::unique_ptr<Foo>&`). As far as I can tell this *could* just be
```suggestion
    miers::PlatformRs& platform_impl;
```
right? `rust::Box<T>` has a `T& operator*()` you can get a reference out of.

src/platforms/evdev-rs/platform_bridge.cpp
```diff
@@ -33,43 +37,125 @@ miers::PlatformBridge::PlatformBridge(
...
+        if (it != backup_fds.end())
+        {
+            ::close(it->second);
+            it->second = backup;
+        }
+        else
+        {
+            backup_fds[key] = backup;
+        }
+    }
```
RAOF: This should all disappear, as we should only close FDs once the device notifies us they're gone.

src/platforms/evdev-rs/platform_bridge.cpp
```diff
@@ -33,43 +37,125 @@ miers::PlatformBridge::PlatformBridge(
...
 {
-    return platform->create_input_device(device_id);
+    std::lock_guard lock{pending_fds_mutex};
+    auto it = pending_fds.find(std::string(devnode));
+    if (it == pending_fds.end())
+        return -1;
+    int fd = it->second;
+    pending_fds.erase(it);
+    return fd;
 }
```
RAOF: We might as well store a `mir::Fd` and keep this as the owner of the file descriptor rather than erasing it.

RAOF (commented):
I think we're bouncing between Rust and C++ more than is necessary - particularly, I think the `DeviceObserver` could be implemented in Rust and we could keep all those FDs on the Rust side, but broadly this looks like a step in the right direction.

mattkae:
> I think we're bouncing between Rust and C++ more than is necessary - particularly, I think the `DeviceObserver` could be implemented in Rust and we could keep all those FDs on the Rust side, but broadly this looks like a step in the right direction.

Yeah I thought this too, but it was more of a convenience thing since the bridge was already being piped down everywhere. I think you're right and that we should make it so that we jump around less.

src/platforms/evdev-rs/platform.cpp
mattkae: Agreed, but both are sort of odd from a safety perspective. The "smart" thing to do would be to wrap this in some `std::shared_ptr` object that we can lock when we need it. That way, the owner of `PlatformRs` (i.e. `miers::Platform`) can feel free to release it per its own lifetime management without having to worry about a dangling reference.

That is what I will do :)

mattkae: Ah we can just use `Self`!

src/platforms/evdev-rs/platform_bridge.cpp
mattkae: This code is so unfortunate, but maybe necessary :thinking: 

The reason that I did this was because closing the lid led to a "switch" event internally. This does not suspend or remove the device. When the device comes back online, `open_restricted` triggers. However, we do not call `on_device_activated`, which means that `open_restricted` will not have an fd to capture. Hence, we keep this backup so that _something_ can get resolved there.

I do not have a better solution for this. Perhaps we can talk about it on Monday.

src/platforms/evdev-rs/platform_bridge.cpp
mattkae: Or we move it to RUst :)

src/platforms/evdev-rs/platform_bridge.cpp
RAOF: #4853 was the fix for this in evdev(non-rs); I'm suggesting we do the same thing here, and then we don't need do this stash-a-fd-when-libinput-closes-it dance (because we never let libinput close the FD anyway).

src/platforms/evdev-rs/src/fd_store.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        // Keep a dup'd backup so libinput can re-open the device later
+        // (e.g. after handling a lid switch event internally).
+        let backup = unsafe { libc::dup(fd) };
+        if backup >= 0 {
+            // Close any previous backup for this device.
+            if let Some(old_backup) = self.backup_fds.insert(devnode.to_string(), backup) {
+                unsafe { libc::close(old_backup) };
+            }
+        }
```
RAOF: We can remove this; basically what you're doing here is storing the FD twice, and then sometimes removing one of the copies.

Instead, just store the FD once and don't remove one copy. That is…

src/platforms/evdev-rs/src/fd_store.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /// Claim and remove the pre-acquired fd for a device path.
+    /// Returns the raw fd, or -1 if no pending fd exists.
+    pub fn claim_pending(&mut self, devnode: &str) -> RawFd {
+        self.pending_fds.remove(devnode).unwrap_or(-1)
```
RAOF: …just *return* the `pending_fd` rather than removing it from the map.

src/platforms/evdev-rs/src/fd_store.rs
```diff
@@ -0,0 +1,93 @@
...
+            None => -1,
+        }
+    }
+
+    /// Remove the backup fd for a device (called on device removal).
+    pub fn remove_backup(&mut self, devnode: &str) {
+        if let Some(fd) = self.backup_fds.remove(devnode) {
+            unsafe { libc::close(fd) };
+        }
+    }
```
RAOF: And then this doesn't need to exist (although you need `remove_backup` to become `remove_device` in order to remove the fd from `pending_fds`)

src/platforms/evdev-rs/src/libinput_interface.rs
```diff
@@ -14,61 +14,57 @@
...
-    /// The bridge used to acquire the device.
-    pub bridge: cxx::SharedPtr<crate::PlatformBridge>,
-
-    /// The list of opened devices.
-    ///
-    /// Mir expects the caller who acquired the device to keep it alive until the device
-    /// is closed. To do this, we keep it in the vector.
-    pub fds: Vec<cxx::UniquePtr<crate::DeviceWrapper>>,
+    /// The fd store used to claim pre-acquired file descriptors.
+    pub fd_store: Arc<Mutex<FdStore>>,
```
RAOF, line 25: This doesn't have to be an `Arc`; the `LibinputInterfaceImpl` can just own the FdStore (or a `Mutex<FdStore>`).

src/platforms/evdev-rs/src/fd_store.rs
```diff
@@ -0,0 +1,93 @@
...
+use std::os::fd::RawFd;
+
+/// Manages pre-acquired and backup file descriptors for device nodes.
+///
+/// When a device is activated via ConsoleServices, its fd is stored here
+/// so that libinput's `open_restricted` callback can claim it without
+/// blocking. A dup'd backup is kept so that libinput can re-open devices
+/// it has internally closed (e.g. after a lid switch event).
+pub struct FdStore {
+    pending_fds: HashMap<String, RawFd>,
```
RAOF: If we store an `OwnedFd` here I think things become simpler?

```suggestion
    fds: HashMap<String, OwnedFd>,
```

src/platforms/evdev-rs/src/libinput_interface.rs
```diff
@@ -14,61 +14,57 @@
...
+        let mut store = self.fd_store.lock().map_err(|_| -libc::ENOLCK)?;
 
-        let major_num = libc::major(st.st_rdev);
-        let minor_num = libc::minor(st.st_rdev);
+        let raw_fd = store.claim_pending(path_str);
+        if raw_fd >= 0 {
+            // Safety: claim_pending transferred ownership of this fd to us.
+            let owned = unsafe { io::OwnedFd::from_raw_fd(raw_fd) };
+            return Ok(owned);
+        }
```
RAOF: ...then if we're storing `OwnedFd`s, we can just `.try_clone()` it here and return the clone...

src/platforms/evdev-rs/src/libinput_interface.rs
```diff
@@ -14,61 +14,57 @@
...
-    // This method is called when libinput is done with a device.
+    /// Called by libinput when it is done with a device fd.
+    ///
+    /// The `OwnedFd` closes the file descriptor automatically when dropped.
     fn close_restricted(&mut self, fd: io::OwnedFd) {
-        let fd_raw = fd.as_raw_fd();
-        let _ = unistd::close(fd);
-        self.fds.retain(|d| d.raw_fd() != fd_raw);
+        // We handle FD externally, from UDev/the ConsoleProvider
+        std::mem::forget(fd);
```
RAOF: ...and if we returned a `.try_clone()` from `open_restricted` then we *would* want to drop the FD here.

src/platforms/evdev-rs/src/platform.rs
```diff
@@ -35,6 +36,18 @@ pub struct PlatformRs {
     report: cxx::UniquePtr<crate::InputReport>,
 
     state: Option<Arc<Mutex<LibinputDeviceState>>>,
```
RAOF, line 38: We actually don't need the `Arc` here - it is never `.clone()`d, `PlatformRs` has sole ownership.

```suggestion
    state: Option<Mutex<LibinputDeviceState>>,
```

src/platforms/evdev-rs/src/platform.rs
```diff
@@ -35,6 +36,18 @@ pub struct PlatformRs {
...
+    running: Arc<AtomicBool>,
+
+    /// Set of devnums we know about (pending or active), to deduplicate udev events.
+    known_devnums: HashSet<u64>,
+
+    /// Rust-owned udev monitor for input device hotplug events.
+    udev_monitor: Option<UdevMonitor>,
+
+    /// Manages pre-acquired and backup file descriptors for device nodes.
+    fd_store: Arc<Mutex<FdStore>>,
```
RAOF, line 50: This can be `&Mutex<FdStore>`, because the `LibinputInterfaceImpl` owns the `FdStore` and is guaranteed to outlive this here.

src/platforms/evdev-rs/src/platform.rs
RAOF: Ah, no, I missed a `.clone()`.

src/platforms/evdev-rs/platform.cpp
```diff
@@ -156,28 +154,133 @@ class InputDevice : public mi::InputDevice
...
+        return;
+    }
+
+    mir::log_info("evdev-rs: enqueuing on_device_activated for %s", devnode.c_str());
+    device_queue->enqueue(
+        [devnode = devnode, devnum = devnum, device_fd = device_fd, &platform_impl = platform_impl]()
+        {
+            // dup() so that mir::Fd's destructor closes the original without
+            // invalidating the fd we hand to libinput via open_restricted().
+            int const duped = ::dup(static_cast<int>(device_fd));
```
RAOF, line 209: You know, we could add a `std::shared_ptr<int>` accessor to `mir::Fd` and then pass *that* to Rust, which would mean we wouldn't need to `dup` quite so much.

src/platforms/evdev-rs/src/fd_store.rs
mattkae: Oh duh! I went tunnel vision this one :sweat_smile:

src/platforms/evdev-rs/src/libinput_interface.rs
mattkae: `OwnedFd` implements `Drop`, so it should be closed automatically:

```rust
#[stable(feature = "io_safety", since = "1.63.0")]
impl Drop for OwnedFd {
    #[inline]
    fn drop(&mut self) {
        unsafe {
            // Note that errors are ignored when closing a file descriptor. According to POSIX 2024,
            // we can and indeed should retry `close` on `EINTR`
            // (https://pubs.opengroup.org/onlinepubs/9799919799.2024edition/functions/close.html),
            // but it is not clear yet how well widely-used implementations are conforming with this
            // mandate since older versions of POSIX left the state of the FD after an `EINTR`
            // unspecified. Ignoring errors is "fine" because some of the major Unices (in
            // particular, Linux) do make sure to always close the FD, even when `close()` is
            // interrupted, and the scenario is rare to begin with. If we retried on a
            // not-POSIX-compliant implementation, the consequences could be really bad since we may
            // close the wrong FD. Helpful link to an epic discussion by POSIX workgroup that led to
            // the latest POSIX wording: http://austingroupbugs.net/view.php?id=529
            #[cfg(not(target_os = "hermit"))]
            {
                #[cfg(unix)]
                crate::sys::fs::debug_assert_fd_is_open(self.fd.as_inner());

                let _ = libc::close(self.fd.as_inner());
            }
            #[cfg(target_os = "hermit")]
            let _ = hermit_abi::close(self.fd.as_inner());
        }
    }
}
```

src/platforms/evdev-rs/src/platform.rs
mattkae: I don't think this is possible?  `Libinput::new_from_path` requires `I: 'static + LibinputInterface`. Once `LibinputInterfaceImpl` is moved into it, `PlatformRs` could not read the `fd_store` back out IIUC.

I understand that we technically have a guarantee, but I think the ownership semantics of the `input` crate actually prevent us from doing this.

src/platforms/evdev-rs/platform.cpp
mattkae: This is just a single `::dup` now, so I am tempted to leave it if that's okay with you

src/platforms/evdev-rs/src/platform.rs
RAOF: :woman_facepalming: yes, of course. I was missing the you-can't-borrow-from-it-then-move bit.

RAOF (approved):
Yup, that's good enough to move on to the next stage of cleanup :)

PR #4961, opened 2026-05-26
Don't use BOOST_THROW_EXCEPTION for Wayland protocol errors

robert-ancell:
These exceptions are always handled and passed back to the remote client. The additional functionality from boost is not required.


jhenstridge:
The only reason I can think of to keep the `BOOST_THROW_EXCEPTION` would be to make debugging easier if we happened to throw the error somewhere it wouldn't get caught. But it doesn't look like we use it consistently even within `server/frontend_wayland`.

RAOF (approved):
I agree, this is reasonable.

PR #4964, opened 2026-05-27
BasicSurface: allow parent to be changed after construction

robert-ancell:


src/include/server/mir/scene/surface.h
```diff
@@ -108,6 +108,9 @@ class Surface :
 
     virtual void request_client_surface_close() = 0;
     virtual std::shared_ptr<Surface> parent() const = 0;
+    /// Set the parent surface. Pass an empty weak_ptr to unset the parent.
+    /// Throws std::runtime_error if the new parent would create a cycle.
```
mattkae:
```suggestion
    /// \throws std::runtime_error if the new parent would create a cycle.
```

tests/unit-tests/scene/test_basic_surface.cpp
```diff
@@ -784,6 +784,70 @@ TEST_F(BasicSurfaceTest, stores_parent)
...
+        streams,
+        std::shared_ptr<mg::CursorImage>(),
+        report,
+        display_config_registrar};
+
+    EXPECT_EQ(child.parent(), nullptr);
+    child.set_parent(parent);
+    EXPECT_EQ(child.parent(), parent);
+    child.set_parent({});
+    EXPECT_EQ(child.parent(), nullptr);
```
mattkae, line 805: Nit: I would stylistically break this into 3 separate tests, but I'm ok either way

mattkae (approved):
Yup!

PR #4967, opened 2026-05-27
Remove double math header include

robert-ancell:


src/server/compositor/stream.cpp
```diff
@@ -19,7 +19,6 @@
 #include <mir/geometry/rectangle.h>
 #include <mir/graphics/buffer.h>
 #include <boost/throw_exception.hpp>
-#include <math.h>
 
 #include <cmath>
```
AlanGriffiths: Why is there any math header used in this file?

```suggestion
```

AlanGriffiths (approved):

PR #4969, opened 2026-05-28
Replace C system includes with the C++ equivalents

robert-ancell:


AlanGriffiths:
IIRC there is no guarantee that the C++ headers add the names to the global namespace. E.g. `<cstring>` defines `std::strcpy()`. I suspect we typically use the global names provided by the C headers.

mattkae (approved):
Yup

AlanGriffiths (changes requested):
Taking a sample of 1 (`examples/client/mir_shell_demo.cpp`), we do not use the `<cstring>` names: `std::strlen()` and `std::memset()` but instead the `string.h` names `::strlen()` and `::memset()`.

If we are going to change these headers, we ought to use the corresponding names. (I know it works in practice... until it doesn't.)

robert-ancell:
Updated with namespacing.

AlanGriffiths (changes requested):
Taking a sample of 1 (`examples/client/mir_shell_demo.cpp`), we do not use the `<cstring>` names: `std::strlen()` and `std::memset()` but instead the `string.h` names `::strlen()` and `::memset()`.

robert-ancell:
With a lot more hand-holding I think we got them all.

src/common/dispatch/multiplexing_dispatchable.cpp
```diff
@@ -22,7 +22,7 @@
 #include <shared_mutex>
 
 #include <sys/epoll.h>
-#include <string.h>
+#include <cstring>
```
AlanGriffiths: Is this even used?

src/common/dispatch/signal_blocker.h
```diff
@@ -17,7 +17,7 @@
 #ifndef MIR_SIGNAL_BLOCKER_H_
 #define MIR_SIGNAL_BLOCKER_H_
 
-#include <signal.h>
+#include <csignal>
```
AlanGriffiths: Is this actually the right header? We use `sigset_t` (without a `std::`) and that is not in [csignal](https://en.cppreference.com/cpp/header/csignal)

I think we are using posix extensions to `signal.h`, these are not mentioned in the C++ standard.

src/common/input/input_event.cpp
```diff
@@ -25,7 +25,7 @@
 
 #include "../handle_event_exception.h"
 
-#include <string.h>
+#include <cstring>
```
AlanGriffiths: Unused?

src/core/log.cpp
```diff
@@ -22,7 +22,7 @@
 #include <format>
 
 #include <boost/exception/diagnostic_information.hpp>
-#include <errno.h>
+#include <cerrno>
```
AlanGriffiths: Unused?

src/include/server/mir/shell/focus_controller.h
```diff
@@ -19,7 +19,7 @@
 
 #include <mir/geometry/forward.h>
 
-#include <stddef.h>
+#include <cstddef>
```
AlanGriffiths: Unused?

src/include/server/mir/shell/input_targeter.h
```diff
@@ -17,7 +17,7 @@
 #ifndef MIR_SHELL_INPUT_TARGETER_H_
 #define MIR_SHELL_INPUT_TARGETER_H_
 
-#include <stddef.h>
```
AlanGriffiths, line 20: Unused?

src/miral/static_display_config.cpp
```diff
@@ -26,7 +26,7 @@
 #include <miral/runner.h>
 
 #include "yaml-cpp/yaml.h"
-#include <string.h>
+#include <cstring>
```
AlanGriffiths: Unused?

src/server/frontend_xwayland/xwayland_clipboard_provider.cpp
```diff
@@ -23,7 +23,7 @@
 #include <mir/dispatch/multiplexing_dispatchable.h>
 
 #include <xcb/xfixes.h>
-#include <string.h>
+#include <cstring>
 #include <fcntl.h>
```
AlanGriffiths, line 26: Unused?

src/server/frontend_xwayland/xwayland_clipboard_source.cpp
```diff
@@ -22,7 +22,7 @@
 #include <mir/dispatch/multiplexing_dispatchable.h>
 
 #include <xcb/xfixes.h>
-#include <string.h>
+#include <cstring>
```
AlanGriffiths: Unused?

AlanGriffiths (changes requested):
This highlights a couple of things as there's a bunch of files where there are no other changes...

In most cases we actually make no use of the header. (Especially `string.h` seems to be sprinkled liberally.)

With `signal.h`, I'm not sure what status posix extensions to have with `csignal`, so maybe we shouldn't go there?

AlanGriffiths (approved):
Thanks!

PR #4972, opened 2026-05-28
Fix - #4552, gmtime is not thread safe

### LINKED ISSUE #4552: `gmtime` and `localtime` are not thread safe and are outdated
We use these primarily in our logger, which can be called from multiple threads at the same time, causing data corruption. In addition, they can be replaced with functions from `std::chrono` and `std::format`.

DigraJatin:
Closes #4552

Replace usage of gmtime with std::chrono and use std::format for print

### DIFF
--- a/src/core/log.cpp
+++ b/src/core/log.cpp
@@ -17,10 +17,10 @@
 #include <mir/log.h>
 #include <mir/logging/logger.h>
 #include <chrono>
+#include <ctime>
 #include <cstdio>
 #include <exception>
 #include <format>
-#include <sstream>
 
 #include <boost/exception/diagnostic_information.hpp>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


src/core/log.cpp
```diff
@@ -101,11 +101,8 @@ void security_log(
     std::string const& event,
     std::string const& description)
 {
-    auto now = std::chrono::system_clock::now();
-    auto time_t = std::chrono::system_clock::to_time_t(now);
-
-    std::stringstream ss;
-    ss << std::put_time(std::gmtime(&time_t), "%FT%TZ");
+    auto const now = std::chrono::floor<std::chrono::seconds>(std::chrono::system_clock::now());
```
AlanGriffiths: Why add the `floor<std::chrono::seconds>()` call?

src/core/log.cpp
```diff
@@ -17,10 +17,10 @@
 #include <mir/log.h>
 #include <mir/logging/logger.h>
 #include <chrono>
+#include <ctime>
```
AlanGriffiths: Is `<ctime>` used? I don't see anything in the changes

AlanGriffiths (changes requested):
A couple of "nits"

src/core/log.cpp
DigraJatin: Truncated sub seconds result to seconds (to keep the log output same as it was before).
Removing floor call also works i believe std::format will take care, but i had kept it for making the intent clear.
Removed the call now.

AlanGriffiths (changes requested):
> logv() still calls vsnprintf() later in this file

Yes, `#include <cstdio>` is still needed

AlanGriffiths (approved):
Thanks!

PR #4879, opened 2026-04-21
Adding a LifetimeTracker for each Wayland object + exposing more methods on the WaylandClient

mattkae:
## What's new?
- Added a `LifetimeTracker` per Wayland object instance
- C++ classes can now have super classes in the generated code
- C++ Impl classes now provide public access to their WaylandClient
- Provide a way for `WaylandClient` to be cloned in C++ between Impls

PR #4976, opened 2026-05-29
Adding a LifetimeTracker for each Wayland object + exposing more methods on the WaylandClient

RAOF:
## What's new?
- Added a `LifetimeTracker` per Wayland object instance
- C++ classes can now have super classes in the generated code
- C++ Impl classes now provide public access to their WaylandClient
- Provide a way for `WaylandClient` to be cloned in C++ between Impls


RAOF (commented):
See my approving review on PR #4879 (that needed to be closed and re-opened in order to unblock CI)

@mattkae if you would kindly do the approval :)

Saviq:
No need for a new PR, reopening the original :)

PR #4978, opened 2026-05-29
Apply live configuration in `start_callback()`

AlanGriffiths:
A last piece to migrate input configuration to "live".

## What's new?

The initial "live" configuration was not being applied: Devices have already been added when `start_callback()` is called. An active `apply()` is needed, not just a passive update to the settings.

A new test `keyboard_repeat_settings_are_applied_to_accessibility_manager_on_start` has been added to verify that keyboard repeat settings from the live config are applied to the accessibility manager on server startup — ensuring this active `apply()` path does not regress to a cache-only update.

## How to test

Miriway uses "live config" for it's .settings file. And the `settings-update` branch (https://github.com/Miriway/Miriway/pull/207) tries to use it exclusively.

Using that Miriway branch it is apparent that options that are only in .settings (and not in .config) are not applied until the .settings file is touched. I noticed it with `touchpad_tap_to_click=true`).

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/miral/input_configuration.cpp
+++ b/src/miral/input_configuration.cpp
@@ -354,9 +354,9 @@ public:
         config.keyboard.merge(keyboard());
         config.touchpad.merge(touchpad());
 
-        mouse(config.mouse);
-        touchpad(config.touchpad);
-        keyboard(config.keyboard);
+        apply(config.mouse);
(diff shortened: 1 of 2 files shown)


mattkae (approved):
👍

PR #4979, opened 2026-06-01
miral: fix constrain resizes of windows without a size

muhammad23012009:
## What's new?
If a client asks Mir to place a window on a position with a size of 0x0, Mir will incorrectly constrain the resize and apply a default size, which further breaks the window positioning formula, as the window doesn't have a previous size. Windows can have a size assigned to them after they have been positioned. Therefore, don't constrain resize of a window until the window has an actual size.

## How to test
Open a window in Lomiri. It will now spawn correctly on-screen instead of out of bounds

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -1164,16 +1164,20 @@ void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpe
         }
     }
 
-    if (modifications.size().is_set() || modifications.top_left().is_set() ||
-        modifications.min_width().is_set() || modifications.min_height().is_set() ||
-        modifications.max_width().is_set() || modifications.max_height().is_set() ||
-        modifications.width_inc().is_set() || modifications.height_inc().is_set())
+    bool const constrain_resize = modifications.size().is_set() ||
+                                  modifications.min_width().is_set() || modifications.min_height().is_set() ||
+                                  modifications.max_width().is_set() || modifications.max_height().is_set() ||
+                                  modifications.width_inc().is_set() || modifications.height_inc().is_set();
+
+    if (modifications.top_left().is_set() || constrain_resize)
     {
         Point new_pos = modifications.top_left().is_set() ? modifications.top_left().value() : window.top_left();
         Size new_size = modifications.size().is_set() ? modifications.size().value() : window.size();
 
         // The new size constraints have already been applied to window_info and constrain_resize() will use them
-        window_info.constrain_resize(new_pos, new_size);
+        if (constrain_resize)
+            window_info.constrain_resize(new_pos, new_size);
+
         place_and_size(window_info, new_pos, new_size);
     }
 


src/miral/window_info.cpp
```diff
@@ -239,11 +239,14 @@ void miral::WindowInfo::constrain_resize(Point& requested_pos, Size& requested_s
...
+    if (requested_size.width.as_int() > 0 && requested_size.height.as_int() > 0)
+    {
+        if (left_resize)
+            new_pos.x -= new_size.width - requested_size.width;
 
-    if (top_resize)
-        new_pos.y -= new_size.height - requested_size.height;
+        if (top_resize)
+            new_pos.y -= new_size.height - requested_size.height;
+    }
```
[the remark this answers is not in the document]
muhammad23012009: A window can never have a negative size though? The smallest size a window could have is 0x0, which would be invisible

AlanGriffiths:
I've not seen problems with the Mir example shells, with Frame, Miriway or Miracle-WM. So I'm trying to understand how this can manifest and whether it is the correct root cause of the problem.

The only place I see `WindowInfo::constrain_resize()` being called in in `BasicWindowManager::modify_window()` where I expect the window to have an existing size. Is it being used differently by Lomiri? Could you point at the relevant version of the code? (I get lost with the multiple branches and repos UBports maintain.)

muhammad23012009:
> I've not seen problems with the Mir example shells, with Frame, Miriway or Miracle-WM. So I'm trying to understand how this can manifest and whether it is the correct root cause of the problem.
> 
> The only place I see `WindowInfo::constrain_resize()` being called in in `BasicWindowManager::modify_window()` where I expect the window to have an existing size. Is it being used differently by Lomiri? Could you point at the relevant version of the code? (I get lost with the multiple branches and repos UBports maintain.)

Whenever Lomiri receives a new surface from Mir, it sets the [requested position](https://gitlab.com/ubports/development/core/lomiri/-/blob/main/plugins/WindowManager/Window.cpp?ref_type=heads#L198) for the surface. This hooks into QtMir's [MirSurface::setRequestedPosition](https://gitlab.com/ubports/development/core/qtmir/-/blob/main/src/modules/QtMir/Application/mirsurface.cpp?ref_type=heads#L1238), and following the call flow it ends up calling [WindowManagementTools::modify_window](https://gitlab.com/ubports/development/core/qtmir/-/blob/main/src/platforms/mirserver/wrappedwindowmanagementpolicy.cpp?ref_type=heads#L698). From there the flow continues down to the problematic part. For some reason when Lomiri/QtMir receives the surface, it has no size bound to it at that point, which is probably why this issue happens. Not sure if the surface is supposed to have a default size when it's allocated.

AlanGriffiths:
> Whenever Lomiri receives a new surface from Mir, it sets the [requested position](https://gitlab.com/ubports/development/core/lomiri/-/blob/main/plugins/WindowManager/Window.cpp?ref_type=heads#L198) for the surface...

Ah, this logic sounds like something that's legacy of mirclient - there a client had to request a window size when creating one. With Wayland a size isn't required until a buffer is submitted.

Looking at that though: the only `modifications` being set is `top_left()` - so _I_ wouldn't expect that to result in a call to `constrain_resize()`. That's not what `modify_window()` actually does though (at least since 11ac2894e38a58725e8921a105ac2d30cce57d19).

I think the right fix is to split the `modifications.top_left().is_set()` condition out there. Does this:

```patch
--- a/src/miral/basic_window_manager.cpp
+++ b/src/miral/basic_window_manager.cpp
@@ -1164,7 +1164,7 @@ void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpe
         }
     }
 
-    if (modifications.size().is_set() || modifications.top_left().is_set() ||
+    if (modifications.size().is_set() ||
         modifications.min_width().is_set() || modifications.min_height().is_set() ||
         modifications.max_width().is_set() || modifications.max_height().is_set() ||
         modifications.width_inc().is_set() || modifications.height_inc().is_set())
@@ -1176,6 +1176,10 @@ void miral::BasicWindowManager::modify_window(WindowInfo& window_info, WindowSpe
         window_info.constrain_resize(new_pos, new_size);
         place_and_size(window_info, new_pos, new_size);
     }
+    else if (modifications.top_left().is_set())
+    {
+        place_and_size(window_info, modifications.top_left().value(), window.size());
+    }
 
     if (modifications.placement_hints().is_set())
     {
```

also solve your problem?

muhammad23012009:
> I think the right fix is to split the `modifications.top_left().is_set()` condition out there.

This would cause the window to spawn with a size of 0x0, currently we're relying (incorrectly?) on constrain_resize to give us a proper window size

AlanGriffiths (changes requested):
I think `BasicWindowManager::modify_window()` is the right place to fix this

muhammad23012009:
Hm, so what should be depend on to set the initial size of a `miral::Window`? When we receive a window it doesn't have any size attached to it, and setting the position in this state adds a size but breaks position without that patch. Does Mir eventually send a resize event to the window that we have to wait for?

AlanGriffiths:
> currently we're relying (incorrectly?) on constrain_resize to give us a proper window size

Does it? Won't it just return `requested_size`? (Which, in `modify_window()` is just `window.size()`)

muhammad23012009:
> > currently we're relying (incorrectly?) on constrain_resize to give us a proper window size
> 
> Does it? Won't it just return `requested_size`? (Which, in `modify_window()` is just `window.size()`)

Yes, it returns `min_width()` and `min_height()`, as `window.size()` is 0x0 by then

muhammad23012009:
Actually, hmmm. I seem to recall QtMir/Mir also setting a window size after setting the top_left, but it was breaking due to this issue (constrain_resize). I will check again and report back

AlanGriffiths:
> Yes, it returns `min_width()` and `min_height()`, as `window.size()` is 0x0 by then

They too will be 0 a lot of the time

muhammad23012009:
Indeed QtMir does follow up with a size modification later, so we just need to place a window when a top_left event arrives. But it also makes me wonder if it's possible for a top_left and size modifications to be sent together? If that's the case then I should probably add more handling for it.

AlanGriffiths:
> But it also makes me wonder if it's possible for a top_left and size modifications to be sent together?

That's exactly what dragging the NW corner does to resize. So, yes, that can happen. But that's not the problem here.

muhammad23012009:
@AlanGriffiths I've went ahead and made constrain_resize only be applied if any of the size properties of a window are modified, which fixes the issue pretty cleanly I'd say

src/miral/window_info.cpp
AlanGriffiths, line 1: With the changes to `modify_window()` I don't think we need these changes?

muhammad23012009: Hm, they shouldn't be here, I removed them with a force-push. I'll clean them out again

muhammad23012009: Okay, should be good now

AlanGriffiths (approved):

PR #4983, opened 2026-06-01
tools: encode template symbols as glob wildcards in version scripts

tarek-y-ismail:
## What's new?

- Instead of the symbol generator trying to match quoted strings from the symbols map exactly, it will first replace angled brackets and commas. This should allow the current symbols map to pass the symbols check.

## How to test

- The symbols check should pass successfully.


tarek-y-ismail:
Just realized I had a typo in the symbols map generator path :facepalm: 

Hope I can sneak it in this PR

Edit: In hindsight, that's a pretty bad idea. This is a good chance to make sure the filter actually works!

tarek-y-ismail:
@Saviq Symbols check passes :) Just need an approval now.

Saviq (approved):
Sure!

PR #4989, opened 2026-06-02
Fix RC version parsing in ppa-upload script

tarek-y-ismail:
## What's new?

(?:...)` is a PCRE non-capturing group, which is not supported by bash's `[[ =~ ]]` operator (POSIX ERE). Replaced it with a standard capturing group `(...)` and adjusted the `BASH_REMATCH` indices accordingly.

## How to test

- You can run parts of `ppa_upload.sh` locally to verify that it doesn't error out.


Saviq (approved):
Worky.

PR #4991, opened 2026-06-03
Custom server-side decorations via MirAL

### LINKED ISSUE #3372: Shell-customized server-side decorations
Shells ought to be able to customise the default SSD Mir provides (currently only for X11 clients)

tsimonq2:
This PR adds limited, in-process custom server-side decoration rendering through MirAL. Compositor authors subclass `miral::DecorationStrategy`, paint software buffers via `miral::DecorationBuffers::create_mapping()`, and return opaque `miral::DecorationSurface` handles from the `render_*` hooks. The strategy is installed with `miral::CustomDecorations` in `MirRunner::run_with`.                                                                                                   
                                                                                                                                                                                                                                                
Closes #3372.                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                
## Background: three different "decoration strategy" types                                                                                                                                                                                      
                                                                                                                                                                                                                                                
Mir already had most of the shell-side plumbing; what was missing was a stable MirAL surface and a configuration hook to inject a custom renderer at startup. The naming is unfortunately overloaded:                                           

| Type | Role |
|------|------|
| `mir::DecorationStrategy` / `miral::Decorations` | SSD vs CSD **negotiation** with clients (`the_decoration_strategy`) |
| `mir::shell::decoration::DecorationStrategy` | Shell-internal strategy that owns geometry + spawns a `RendererStrategy` per frame |
| `miral::DecorationStrategy` | **New** public API your compositor implements to paint SSD buffers |

You typically want **both** `miral::Decorations::prefer_ssd()` (or `always_ssd()`) **and** `miral::CustomDecorations`: the former ensures clients negotiate server-side decorations; the latter replaces how those decoration buffers are drawn once SSD is active. They hook different server configuration slots and are intentionally independent.

## How it plugs into the existing shell path

Nothing in `BasicDecoration`'s render/commit loop changes shape. The shell already factored painting behind `mir::shell::decoration::RendererStrategy` (titlebar + three borders, optional buffers, `update_state()` per frame). The built-in `DecorationStrategy::default_decoration_strategy()` returns a strategy whose `render_strategy()` produces the stock FreeType/theme renderer.

This PR adds a parallel injection point:

1. `miral::CustomDecorations` registers a pre-init callback on `mir::Server`.
2. At pre-init, MirAL wraps the user strategy in `DecorationStrategyAdapter` (implements shell `DecorationStrategy`) and calls `server.set_the_decoration_renderer_strategy(adapter)`.
3. `DefaultServerConfiguration::the_decoration_manager()` reads that hook; if set, it passes the custom strategy to `BasicManager` instead of `default_decoration_strategy()`.
4. Each decorated window still gets a `BasicDecoration` that calls `render_strategy()->update_state()` then `render_titlebar()` / `render_*_border()` exactly as before.

The adapter's `RendererAdapter` (implements shell `RendererStrategy`) is the only new code on the hot path: it translates shell `WindowState` / `InputState` into MirAL value types, invokes the user's `render_*` methods, and unwraps the returned `DecorationSurface` back into `std::shared_ptr<mir::graphics::Buffer>` for the shell.

On server stop, `CustomDecorations` clears `the_decoration_renderer_strategy` to `nullptr` so the shell does not hold a renderer adapter across EGL teardown.

## Server configuration changes

New read/write hook on `ServerConfiguration` and `mir::Server`:

- `the_decoration_renderer_strategy()` / `set_the_decoration_renderer_strategy()`

Stored in `DefaultServerConfiguration` alongside the existing decoration negotiation strategy. Exported under `MIR_SERVER_INTERNAL_2.29` in `src/server/symbols.map`. Stub server configuration in tests exposes the setter for unit coverage.

`the_decoration_manager()` factory now does:

```cpp
auto const custom = the_decoration_renderer_strategy();
return std::make_shared<msd::BasicManager>(
    custom ? custom : msd::DecorationStrategy::default_decoration_strategy(the_buffer_allocator()),
    ...);
```

If no custom renderer is configured, behaviour is identical to main.

## Shell / decoration internal changes

**Header relocation:** `decoration_strategy.h` moves from `src/server/shell/decoration/` into `src/include/server/mir/shell/decoration/` so MirAL's adapter can include the shell strategy interface without pulling server-private paths.

**Title text extraction:** FreeType title rendering is pulled out of `decoration_strategy.cpp` into `render_title_text.cpp` / `render_title_text.h` in the shell decoration OBJECT library. The built-in default renderer and MirAL's default `DecorationStrategy::render_title_text()` both call this helper. It is exported from `libmirserver` as an unstable internal symbol (not a supported public API); MirAL reaches it through the existing `libmirserver` link — no new MirAL link dependencies.

**`window_id()` token:** Shell `WindowState::window_id()` previously exposed the `scene::Surface` pointer as an integer token. That is replaced with a monotonic `std::atomic<uint64_t>` assigned at `WindowState` construction. The MirAL public `WindowState::window_id()` documents this as an opaque per-window token stable for the decorated window's lifetime (not a `miral::Window` handle, not preserved across surface destroy/recreate). This keeps scene internals out of the public API while still allowing compositors to key per-window render caches.

## MirAL public API (MIRAL 6.0)

New headers under `include/miral/miral/`:

**`miral::DecorationStrategy`** — stateless render interface. Each `render_*` call receives immutable `WindowState` and `InputState` snapshots plus a `DecorationBuffers&` reference. Implementations must provide:

- `compute_size_with_decorations()` — outer size from content size + window type/state
- `button_placement(n, ws)` — hit-target rectangles for close / maximize / minimize
- `render_titlebar()` / `render_left_border()` / `render_right_border()` / `render_bottom_border()` — return `std::optional<DecorationSurface>`
- `buffer_format()`, `resize_corner_input_size()`
- Optional geometry overrides: `titlebar_height()`, `side_border_width()`, `bottom_border_height()` (defaults match the built-in theme)
- Default `render_title_text()` delegating to the shell FreeType helper

Nested types `WindowState`, `InputState`, and `Button` mirror the shell semantics but deliberately omit shell-only data (e.g. `input_shape` is not exposed in v1; only the three standard buttons are passed through `InputState`).

**`miral::DecorationBuffers`** — software buffer helper constructed by MirAL, not by compositor code. `create_mapping(size, format)` allocates via the server's `GraphicBufferAllocator`, maps CPU-writable memory, and returns `Pixels{surface, mapping}`. The pimpl hides allocator ownership. Compositor authors paint through `Mapping::pixels32()` (or `data()` / `stride()`), then `return std::move(pixels.surface)`.

**`miral::DecorationSurface`** — opaque, move-only return type from `render_*`. Public headers never mention `mir::graphics::Buffer`; the shell adapter unwraps the internal buffer handle. Copy is deleted with a C++26 `= delete("…")` rationale string.

**`miral::CustomDecorations`** — explicit `CustomDecorations(std::shared_ptr<DecorationStrategy>)` only (null strategy is a fatal error). Strategy is fixed at `MirRunner` construction; there is no runtime swap. Documented in `decorations.h` as distinct from `miral::Decorations`.

**`miral::pack_color(a, r, g, b)`** — constexpr endian-safe ARGB packing for `pixels32()` writes.

## Adapter implementation (MirAL private)

`DecorationStrategyAdapter` and `RendererAdapter` live in `src/miral/decoration_strategy_adapter.{h,cpp}` and are not exported.

`DecorationStrategyAdapter::new_window_state()` passes the user's geometry overrides (`titlebar_height()`, etc.) into the shell `WindowState` constructor so custom metrics flow through the same path as the built-in strategy.

`RendererAdapter` holds one `DecorationBuffers` member (constructed with the server allocator) and reuses it across `render_*` calls within a single decoration instance. `update_state()` caches translated MirAL `WindowState` / `InputState`. `render_from_user()` dispatches to the appropriate user `render_*` member function and, on success, extracts the `shared_ptr<graphics::Buffer>` from the returned `DecorationSurface`.

Type translation functions (`to_miral()` for border type, button function/state, window state, input state) keep the public API decoupled from shell enum naming (`BorderType::Full` vs `full`, etc.).

## v1 limitations (intentional)

- **Software buffers only** — no GL/EGL/GPU decoration path in this API.
- **Standard buttons only** — close, maximize, minimize; no custom button types or shadows.
- **InputState** exposes button rectangles only; internal non-button hit regions (`input_shape`) are not available to custom renderers.
- **`window_id()`** is an opaque `uint64_t`, not a `miral::Window`.
- **In-process only** — no plugin loader, no out-of-process renderer, no dynamic strategy replacement after server start.
- A richer Wayland-client decoration model may come later; this is the pragmatic in-process v1 discussed on the June review threads.

## Try it

`examples/miral-custom-decorations/miral-custom-decorations.cpp` — green chrome, buttons on the left, scale-aware. Visibly different from the built-in theme. Installed via `mir-demos`.

```cpp
runner.run_with({
    miral::set_window_management_policy<miral::MinimalWindowManager>(),
    miral::Decorations::prefer_ssd(),
    miral::CustomDecorations{std::make_shared<MyDecoration>()},
    // ...
});
```

Typical render body:

```cpp
auto pixels = buffers.create_mapping(scaled_size, buffer_format());
// paint via pixels.mapping.pixels32()
return std::move(pixels.surface);
```

## Tests and documentation

- `tests/unit-tests/shell/test_decoration_strategy_public_include.cpp` — public header smoke test, `pack_color` endian check, null-strategy death test for `CustomDecorations`, `RendererAdapter` bridge exercise, geometry override propagation.
- `doc/sphinx/how-to/custom-server-side-decorations.md` (new) — naming table, quick start, API overview, limitations.
- Updates to `introducing_the_miral_api.md` and cross-link from `specifying-csd-ssd-preference.md`.

## ABI / symbols

**MirAL 6.0** (`debian/libmiral8.symbols`): `CustomDecorations`, `DecorationBuffers::create_mapping` and `Mapping` accessors, `DecorationSurface` move ctor/dtor, `WindowState::build`, `InputState::build`, `DecorationStrategy::render_title_text`.

**Server internal 2.29**: `the_decoration_renderer_strategy` getter/setter on `Server` and `DefaultServerConfiguration`, plus `mir::shell::decoration::render_title_text`.

Please verify clean compositor exit on your machine after exercising custom decorations; the stop callback clears the renderer hook specifically to reduce EGL teardown ordering risk.

Prepared with assistance from Grok Build (Composer 2.5).


tsimonq2:
When mir-demos is installed, with a simple change to Miriway's Runner call:

```
--- a/miriway.cpp
+++ b/miriway.cpp
@@ -598,6 +598,7 @@ int main(int argc, char const* argv[])
             set_window_management_policy<WindowManagerPolicy>(commands),
             lockscreen,
             getenv_decorations(),
+            miral::Decorations::load_renderer_plugin("miral-minimal-decoration"),
             CursorTheme{"default"},
             wltools,
             cursor_scale,
```

Windows show as follows:

image

src/include/server/mir/default_server_configuration.h
tarek-y-ismail, line 1: This, and other changes in this commit ("Make core DecorationStrategy thread-safe for XWayland") are mostly about exposing the decoration strategy, not updating threading.

Another better name for this PR could be "expose and plumb `{set_,}the_decoration_renderer_strategy`"

tarek-y-ismail:
Adding myself and Alan to give this a review since I've already had a couple of attempts at implementing this and it modifies the public API. I'm halfway through a quick initial review though it will take a bit more time :)

HACKING.md
```diff
@@ -124,3 +124,27 @@ corresponding source requires manually specifying a directory to find
...
+
+## Decoration plugins
+
+Custom server-side decoration rendering is supplied via plugins loaded with the
+established `SharedLibrary` + `libraries_for_path` mechanism (identical to
+graphics and input platforms).
+
+The public (mirserver) entry points and authoring headers live under
+`include/server/mir/shell/decoration/`. The recommended registration uses `MIR_REGISTER_DECORATION_PLUGIN` (no `extern "C"` required).
+The legacy `describe_decoration_plugin` / `create_decoration_strategy_factory` symbols
```
AlanGriffiths: There are no legacy `describe_decoration_plugin` / `create_decoration_strategy_factory` symbols

HACKING.md
```diff
@@ -124,3 +124,27 @@ corresponding source requires manually specifying a directory to find
...
+- Add new public types under that tree; they are automatically installed into
+  `mirserver` (plugin authors depend only on the mirserver package for authoring
+  types; full Server hooks are advanced, see how-to).
+- Run `tools/update_package_abis.sh` for any packaging or ABI updates (never
+  edit `src/server/symbols.map` manually). (This rule applies to decoration plugin ABI updates as well.)
+- Reference in-process decoration plugin example is in
+  `examples/decoration-plugins/minimal/`.
+- The loader and default wiring live in `src/server/shell/decoration/`.
+
+See the Sphinx how-to "Writing a Decoration Plugin" for the full authoring guide,
```
AlanGriffiths: Sphinx is a tool, not part of the user facing documentation

include/miral/miral/decorations.h
```diff
@@ -49,6 +50,11 @@ class Decorations
     /// mode. Otherwise use the mode specified by the client.
     static auto prefer_csd() -> Decorations;
 
+    /// Load the named decoration renderer plugin (if present in the
+    /// plugin search path) and use it for server-side decorations.
+    /// Falls back to the built-in renderer if the plugin cannot be loaded.
+    static auto load_renderer_plugin(std::string const& name) -> Decorations;
```
AlanGriffiths: A file should be `std::filesystem::path`, not a `std::string`. But... should we be specifying a raw filename? Or use the same model as we do for graphics and input platforms? (I.e. specifying a name.)

include/server/mir/shell/decoration/plugin/settings.h
```diff
@@ -0,0 +1,71 @@
...
+ */
+struct Settings
+{
+    Pixel focused_background   = 0x323232FF;
+    Pixel unfocused_background = 0x808080FF;
+    Pixel focused_text         = 0xFFFFFFFF;
+    Pixel unfocused_text       = 0xA0A0A0FF;
+    Pixel button_color         = 0x606060FF;
+    Pixel close_button_color   = 0xA02020FF;
+    Pixel button_icon_color    = 0xFFFFFFFF;
```
AlanGriffiths: These default makes assumptions about endianness and color format

include/server/mir/shell/decoration/plugin/frame.h
```diff
@@ -0,0 +1,214 @@
...
+        mir::geometry::Size content_size, MirWindowType type, MirWindowState state) const -> mir::geometry::Size override;
+    auto new_window_state(
+        std::shared_ptr<mir::scene::Surface> const& window_surface, float scale) const -> std::unique_ptr<WindowState> override;
+    auto buffer_format() const -> MirPixelFormat override;
+    auto resize_corner_input_size() const -> mir::geometry::Size override;
+
+    /// Pass theme assets path (from loader 2/3-param or direct) through to
+    /// frames/settings. Mir never interprets the path. See class docs + how-to
+    /// for complete author examples (both factory paths).
+    void set_theme_path(std::string const& path) override;
```
AlanGriffiths: A path should actually be a `std::filesystem::path`

include/server/mir/shell/decoration/plugin/client.h
```diff
@@ -0,0 +1,57 @@
...
+ * a decoration renderer needs about the underlying surface without
+ * exposing Mir internals directly.
+ *
+ * Implementations are provided by adapters when using higher-level
+ * plugin renderers.
+ *
+ * Theme assets (if any) are accessed via the paired Frame::settings().theme_path
+ * (see settings.h); Client gives per-window state only.
+ */
+class Client
```
AlanGriffiths: As a client may have several windows, this naming seems confusing.

Also, it is common to decorate different window types differently: e.g. thinner borders for dialogs, but there's no proxying of the window type.

We may also want to use the pimpl idiom here to make for a more stable API

include/server/mir/shell/decoration/plugin/frame.h
```diff
@@ -0,0 +1,214 @@
...
+ * Button in decoration_strategy.h for geometry).
+ *
+ * The adapter or plugin Frame populates these from InputState + internal state.
+ * Theme plugins (Openbox etc) can style based on these + theme_path assets.
+ * (Basic impl + input regions now provided so painted buttons are hit targets.)
+ */
+class Button
+{
+public:
+    enum class Function { Close, Maximize, Minimize, Help, Menu, Shade, Stick /* extensible */ };
```
AlanGriffiths: Elements do not match the coding guideline (`lower_case`)

What does the "/* extensible */" comment mean? What guarantees has the consumer/provider of these have?

debian/libmiral7.symbols
```diff
@@ -730,6 +730,7 @@ libmiral.so.7 libmiral7 #MINVER#
  (c++)"miral::WindowSpecification::alpha() const@MIRAL_5.7" 5.7.0
  (c++)"miral::WindowSpecification::alpha()@MIRAL_5.7" 5.7.0
  MIRAL_5.8@MIRAL_5.8 5.8.0
+ (c++)"miral::Decorations::load_renderer_plugin(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)@MIRAL_5.8" 5.8.0
```
AlanGriffiths: 5.8 has ⛵ with Mir 2.2[78]

AlanGriffiths (commented):
Thanks for this, it is a feature we need and the approach is interesting.

But there's a lot of detail here that needs polish and, particularly the "out of process" support may be problematic. Having an out of process renderer definitely offers a lot of new failure modes. We have some existing thoughts about making in-process rendering less painful over the current cycle.

Can you join us at the [Mir Office Hours](https://discourse.ubuntu.com/t/mir-office-hours-2026-06-03-15-15utc/83611) later today?

mattkae:
My initial thought is that the plugin system is overkill here. If a shell maintainer wants to write a decoration implementation that dynamically loads `.so` files, they should be able to with what we provide them via the `miral` API, but it should not be a feature of `miral` itself. The `miral` level API should be generic enough that the decorations could be loaded from _any_ source (maybe that's a .so, but maybe that's also from an embedded interpreter or elsewhere).

My two cents on how we should approach this are:

1. We should provide a way to implement SSD with _actual_ Wayland clients that have 4 surfaces (1 for each side of the frame)
2. The client should be "internal" to start, but can be external some day
3. This client is special in that Mir registers it as a "decoration client" for a particular window
4. The window management of the client is such that it will follow the size of the window as a frame rather than going through the policy implemented by the shell author

This makes it so that we avoid building a small toolkit just for decorations in Mir. At the same time, allowing clients to use an actual toolkit will make drawing (especially text) so much easier in the long term. This also enables particularly fancy designs and text in the decorations (e.g. displaying random things like the size of the window in the decorations).

AlanGriffiths:
The proposed approach is for the `libmiral` API to identify a plugin, and for that plugin to support an interface defined in `libmirserver`. That is problematic:

1. `libmirserver` is our most unstable ABI, so this risks breaking on every Mir release
2. the interface is unversioned, which means nothing prevents an incompatible implementation being loaded

As Matt suggests, there seems little value in, and substantial cost to, Mir providing a way to dynamically load a decoration plugin. The simplest customization point is for a downstream compositor to incorporate the code directly.

Ideally, that would be via a stable, supported API in `libmiral`.

src/server/frontend_xwayland/xwayland_wm_shell.h
```diff
@@ -20,6 +20,7 @@
 #define MIR_FRONTEND_XWAYLAND_WM_SHELL_H
 
 #include <memory>
+#include <mir/decoration_strategy.h>
```
tarek-y-ismail: Per the [cppguide](https://github.com/canonical/mir/blob/50d8720926bfb6a9bce6bf41b3c85fedaa2174af/doc/sphinx/contributing/reference/cppguide.md?plain=1#L233-L238), C++ stdlib headers are to be included _after_ public/private project headers/

src/server/symbols.map
tarek-y-ismail, line 1: Since we're adding to the server symbols, all the new symbols should be in a new `MIR_SERVER_INTERNAL_2.27` stanza.

tests/include/mir_test_framework/stubbed_server_configuration.h
```diff
@@ -46,6 +46,9 @@ class StubbedServerConfiguration : public DefaultServerConfiguration
     std::shared_ptr<graphics::Cursor> the_cursor() override;
     std::shared_ptr<logging::Logger> the_logger() override;
 
+    using DefaultServerConfiguration::the_decoration_renderer_strategy;
+    using DefaultServerConfiguration::set_the_decoration_renderer_strategy;
+
```
tarek-y-ismail, line 51: What's the purpose of these two lines?

tests/include/mir_test_framework/testing_server_configuration.h
```diff
@@ -50,6 +50,8 @@ class TestingServerConfiguration : public StubbedServerConfiguration
     std::shared_ptr<mir::ServerStatusListener> the_server_status_listener() override;
 
     using DefaultServerConfiguration::the_options;
+    using DefaultServerConfiguration::the_decoration_renderer_strategy;
+    using DefaultServerConfiguration::set_the_decoration_renderer_strategy;
```
tarek-y-ismail: What's the purpose of these two lines?

tests/unit-tests/shell/test_decoration_basic_manager.cpp
```diff
@@ -24,7 +24,7 @@
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
-#include <src/server/shell/decoration/decoration_strategy.h>
+#include <mir/shell/decoration/decoration_strategy.h>
```
tarek-y-ismail: Header ordering. Refer to my first comment on this commit.

include/miral/miral/decorations.h
tarek-y-ismail, line 1: Unless we collide with another file, we probably can discard changes made here. The removal of the `Server` forward decl can be its own cleanup PR (along with other similar instances)

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,177 @@
...
+};
+
+struct Button
+{
+    enum class Function
+    {
+        close,
+        maximize,
+        minimize,
+    };
```
tarek-y-ismail: What happens when users (shell authors) want to implement something that doesn't adhere to this? For example a shell with an "always on top button", or "do a barrel roll" button?

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,177 @@
...
+    abgr_8888 = 1,
+    xbgr_8888 = 2,
+    argb_8888 = 3,
+    xrgb_8888 = 4,
+    bgr_888 = 5,
+    rgb_888 = 6,
+    rgb_565 = 7,
+    rgba_5551 = 8,
+    rgba_4444 = 9,
+};
```
tarek-y-ismail: We probably shouldn't duplicate the existing enum. Do we want to expose it instead?

tests/unit-tests/shell/test_decoration_strategy_public_include.cpp
```diff
@@ -40,5 +49,13 @@ TEST(MirserverInternalCompile, decoration_strategy_header_and_server_hook)
     (void)bt;
     (void)fn;
 
+    // Trivial runtime check exercising the MirAL public API entry points
+    // (CustomDecorations via the runner arg pattern, and Decorations).
+    // Cheap: no full server startup required.
+    [[maybe_unused]] auto const custom = miral::CustomDecorations::strategy(nullptr);
```
tarek-y-ismail: This will probably fail to link as `CustomDecorations` was introduced in the previous commit with only a declaration.

debian/libmiral7.symbols
```diff
@@ -762,3 +762,6 @@ libmiral.so.7 libmiral7 #MINVER#
...
  (c++|arch-bits=64)"miral::live_config::IniFileWithOverrides::add_strings_attribute(miral::live_config::Key const&, std::basic_string_view<char, std::char_traits<char> >, std::span<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, 18446744073709551615ul>, std::function<void (miral::live_config::Key const&, std::optional<std::span<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, 18446744073709551615ul> >)>)@MIRAL_5.8" 5.8.0
  (c++)"std::formatter<miral::live_config::Key, char>::format(miral::live_config::Key const&, std::basic_format_context<std::__format::_Sink_iter<char>, char>&) const@MIRAL_5.8" 5.8.0
+ (c++)"miral::CustomDecorations::strategy(std::shared_ptr<miral::DecorationStrategy>)@MIRAL_5.8" 5.8.0
+ (c++)"miral::CustomDecorations::CustomDecorations(std::shared_ptr<miral::CustomDecorations::Self>)@MIRAL_5.8" 5.8.0
+ (c++)"miral::CustomDecorations::operator()(mir::Server&) const@MIRAL_5.8" 5.8.0
```
tarek-y-ismail: 5.8 has been released. We're onto 5.9 now

src/miral/decorations.cpp
```diff
@@ -84,3 +83,30 @@ auto miral::Decorations::prefer_csd() -> Decorations
...
+    if (self->strategy) {
+        server.add_pre_init_callback(
+            [this, &server]()
+            {
+                auto const allocator = server.the_buffer_allocator();
+                auto adapter = std::make_shared<miral::DecorationStrategyAdapter>(self->strategy, allocator);
+                server.set_the_decoration_renderer_strategy(adapter);
+            });
+    }
+}
```
tarek-y-ismail, line 116: Should probably be moved to the previous commit, or the one before it.

src/miral/symbols.map
tarek-y-ismail, line 1: Again, 5.8 has been released. Please move them to a new 5.9 stanza.

doc/sphinx/reference/introducing_the_miral_api.md
```diff
@@ -87,6 +87,50 @@ Most of the remaining MirAL components are designed to be passed into
...
+      miral::set_window_management_policy<MyWindowManager>(),
+      miral::CustomDecorations::strategy(decorations),
+  });
+  ```
+
+The {struct}`miral::CustomDecorations` runner argument wires your strategy as the decoration renderer.
+
+A self-contained demonstration is provided in `examples/miral-custom-decorations/`. `custom_decoration.cpp` shows a basic subclass (with example button placement logic, state caching, and stub renders); `main.cpp` shows integration. You can copy and extend this for production use (e.g. adding QPainter/gradients or Openbox-style texture parsing).
+
+This approach keeps custom SSD as a stable, in-process MirAL feature (no plugins or ABI extension points required). It can be used alongside or instead of the built-in basic decorations and the {class}`miral::Decorations` preference controls.
```
tarek-y-ismail: No need to mention the old plugin/extension architecture.

tarek-y-ismail (commented):
Better and leaner than the previous iteration. I've attached some nits you're free to ignore until we decide whether we want to move forward with this or not. My main gripe is that we're exposing a lot in the API that might not be flexible enough for long-term support.  Remember, this has to support all kinds of decorations we can think of _and_ can't yet think of. 

For example:
- Decorations with extra buttons that we don't know/implement
- Decorations with drop shadows
- Decorations rendered on the GPU
- Stretch: Wacky decorations (special effects, animations, etc...)

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,180 @@
...
+    mir::geometry::Rectangle rect;
+};
+
+struct InputState
+{
+    std::vector<Button> buttons;
+};
+
+// Local enum classes matching the legacy toolkit values (for self-contained public API
+// and simple casts in the adapter).
```
AlanGriffiths: Just use `mir_toolkit/common.h` the same as other miral APIs: miral already publicly depends on mircore.

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,180 @@
...
+    virtual auto render_titlebar() -> std::optional<std::shared_ptr<mir::graphics::Buffer>> = 0;
+    virtual auto render_left_border() -> std::optional<std::shared_ptr<mir::graphics::Buffer>> = 0;
+    virtual auto render_right_border() -> std::optional<std::shared_ptr<mir::graphics::Buffer>> = 0;
+    virtual auto render_bottom_border() -> std::optional<std::shared_ptr<mir::graphics::Buffer>> = 0;
+
+    virtual auto buffer_format() const -> MirPixelFormat = 0;
+    virtual auto resize_corner_input_size() const -> mir::geometry::Size = 0;
+
+    virtual void set_buffer_allocator(std::shared_ptr<mir::graphics::GraphicBufferAllocator> const& allocator) = 0;
+    virtual void set_theme_path(std::filesystem::path const&) {}
```
AlanGriffiths: Detritus?

src/miral/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ add_library(miral-internal STATIC
     window_specification_internal.cpp    window_specification_internal.h
     basic_store.cpp                      basic_store.h
     live_config_ini_file_common.cpp      live_config_ini_file_common.h
+    decoration_strategy_adapter.cpp
```
AlanGriffiths:
```suggestion
    decoration_strategy_adapter.cpp      decoration_strategy_adapter.h
```

src/miral/decoration_strategy_adapter.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include <mir/graphics/graphic_buffer_allocator.h>
+#include <mir/scene/surface.h>
+
+#include <filesystem>
+#include <memory>
+
+namespace miral
+{
+
+WindowState to_miral(mir::shell::decoration::WindowState const& ws)
```
AlanGriffiths: Project style is only using namespace blocks for declarations. Also to use the newer function style:

```suggestion
auto miral::to_miral(mir::shell::decoration::WindowState const& ws) -> WindowState
```

Also, these functions are only used in this translation unit: for hygiene they should be in the anonymous namespace

src/miral/decoration_strategy_adapter.h
```diff
@@ -0,0 +1,62 @@
...
+    auto render_strategy() const -> std::unique_ptr<mir::shell::decoration::RendererStrategy> override;
+    auto button_placement(unsigned n, mir::shell::decoration::WindowState const& ws) const -> mir::geometry::Rectangle override;
+    auto compute_size_with_decorations(mir::geometry::Size content_size, ::MirWindowType type, ::MirWindowState state) const -> mir::geometry::Size override;
+    auto new_window_state(std::shared_ptr<mir::scene::Surface> const& surface, float scale) const -> std::unique_ptr<mir::shell::decoration::WindowState> override;
+    auto buffer_format() const -> ::MirPixelFormat override;
+    auto resize_corner_input_size() const -> mir::geometry::Size override;
+    void set_theme_path(std::filesystem::path const& path);
+private:
+    std::shared_ptr<miral::DecorationStrategy> user_;
+    std::shared_ptr<mir::graphics::GraphicBufferAllocator> allocator_;
```
AlanGriffiths: There is no need for the `_` warts

src/miral/decoration_strategy_adapter.cpp
```diff
@@ -0,0 +1,159 @@
...
+    }
+    return m;
+}
+
+DecorationStrategyAdapter::DecorationStrategyAdapter(
+    std::shared_ptr<miral::DecorationStrategy> user,
+    std::shared_ptr<mir::graphics::GraphicBufferAllocator> alloc)
+    : user_(std::move(user)), allocator_(std::move(alloc))
+{
+    if (user_) user_->set_buffer_allocator(allocator_);
```
AlanGriffiths: Is there any need for `user` to be nullable? Shouldn't `!user` be a precondition violation?

src/server/symbols.map
AlanGriffiths: > `MIR_SERVER_INTERNAL_2.27` stanza.

No, 2.27 has sailed. We're now working toward 2.29

tests/unit-tests/shell/test_decoration_strategy_public_include.cpp
```diff
@@ -0,0 +1,61 @@
...
+    [[maybe_unused]] auto const set_ptr = &Server::set_the_decoration_renderer_strategy;
+
+    // Touch a type from the rich header
+    [[maybe_unused]] auto bt = shell::decoration::BorderType::Full;
+    [[maybe_unused]] auto fn = shell::decoration::Button::Function::Close;
+
+    (void)get_ptr;
+    (void)set_ptr;
+    (void)bt;
+    (void)fn;
```
AlanGriffiths: You don't need both `[[maybe_unused]]foo` and `(void)foo`

doc/sphinx/reference/introducing_the_miral_api.md
```diff
@@ -87,6 +87,50 @@ Most of the remaining MirAL components are designed to be passed into
 - Controlling the available Wayland extension protocols: {class}`miral::WaylandExtensions`; and,
 - Selecting server-side or client-side decorations: {class}`miral::Decorations`.
 
+### Custom server-side decorations
+
+Beyond the high-level preferences provided by {class}`miral::Decorations` (see {ref}`specifying-csd-ssd-preference`), MirAL supports implementing fully custom server-side decorations (SSD) using the {class}`miral::DecorationStrategy` interface. This is ideal for custom theming, advanced rendering (gradients, xbm masks, text, shadows), or specific layout needs.
```
AlanGriffiths: The API being exposed falls short of "fully custom" - it allows some limited customization, but is limited in scope.

doc/sphinx/reference/introducing_the_miral_api.md
```diff
@@ -87,6 +87,50 @@ Most of the remaining MirAL components are designed to be passed into
...
+
+To implement:
+
+- Derive from {class}`miral::DecorationStrategy`.
+
+- Provide implementations for the pure virtual methods (the API is self-contained; it defines its own {enum}`miral::MirWindowType`, {enum}`miral::MirWindowState`, {enum}`miral::MirWindowFocusState`, and {enum}`miral::MirPixelFormat` to avoid external dependencies):
```
AlanGriffiths: Why?

examples/miral-custom-decorations/custom_decoration.cpp
```diff
@@ -0,0 +1,129 @@
...
+    {
+        current_ws = ws;
+        current_input = input;
+    }
+
+    auto render_titlebar()
+        -> std::optional<std::shared_ptr<mir::graphics::Buffer>> override
+    {
+        (void)current_ws;
+        (void)current_input;
```
AlanGriffiths: Huh?

examples/miral-custom-decorations/main.cpp
```diff
@@ -0,0 +1,48 @@
...
+    MirRunner runner{argc, argv};
+
+    auto const custom_decoration = make_my_custom_decoration();
+
+    return runner.run_with({
+        set_window_management_policy<MinimalWindowManager>(),
+        Decorations::prefer_ssd(),
+        CustomDecorations::strategy(custom_decoration),
+        Keymap{},
+        ConfigurationOption{[](bool){}, "demo", "Demo option.", false},
```
AlanGriffiths: Why add this option? But, if you're adding an option, how about choosing a decoration theme?

Also, there doesn't seem to be much point in having two files for this example.

AlanGriffiths (changes requested):
Running the example there's nothing very different about the decorations. Is this working?

How about demonstrating something visibly different? Like https://github.com/canonical/mir/pull/3708#issuecomment-2557463363

AlanGriffiths:
Also, I see a crash on exit:

```
terminate called after throwing an instance of 'boost::wrapexcept<std::system_error>'
  what():  Failed to make context current: EGL_BAD_DISPLAY (0x3008)

Thread 1 "miral-custom-de" received signal SIGABRT, Aborted.
Download failed: Invalid argument.  Continuing without source file ./nptl/./nptl/pthread_kill.c.
__pthread_kill_implementation (threadid=<optimised out>, signo=6, no_tid=0) at ./nptl/pthread_kill.c:44
⚠️ warning: 44  ./nptl/pthread_kill.c: No such file or directory
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimised out>, signo=6, no_tid=0) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (threadid=<optimised out>, signo=6) at ./nptl/pthread_kill.c:89
#2  __GI___pthread_kill (threadid=<optimised out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:100
#3  0x00007ffff7445b7e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff74288ec in __GI_abort () at ./stdlib/abort.c:77
#5  0x00007ffff78ac275 in ??? () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff78c364a in ??? () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007ffff78abaf4 in __cxa_call_terminate () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007ffff78c2b2c in __gxx_personality_v0 () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#9  0x00007ffff7f91592 in ??? () at /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
#10 0x00007ffff7f920db in _Unwind_Resume () at /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
#11 0x00007ffff45ac4f0 in std::allocator<char>::~allocator (this=<optimised out>) at /usr/include/c++/15/bits/allocator.h:189
#12 (anonymous namespace)::SurfacelessEGLContext::make_current (this=0x555555af3bf0) at /home/alan.griffiths@canonical.com/canonical/mir/src/platforms/renderer-generic-egl/buffer_allocator.cpp:132
#13 0x00007ffff45ac352 in (anonymous namespace)::SurfacelessEGLContext::~SurfacelessEGLContext (this=0x555555af3bf0)
    at /home/alan.griffiths@canonical.com/canonical/mir/src/platforms/renderer-generic-egl/buffer_allocator.cpp:123
#14 0x00007ffff45ac3a4 in (anonymous namespace)::SurfacelessEGLContext::~SurfacelessEGLContext (this=0x555555af3bf0)
    at /home/alan.griffiths@canonical.com/canonical/mir/src/platforms/renderer-generic-egl/buffer_allocator.cpp:126
#15 0x00007ffff67ea4b6 in std::default_delete<mir::renderer::gl::Context>::operator() (this=0x5555556d8368, __ptr=0x555555af3bf0) at /usr/include/c++/15/bits/unique_ptr.h:92
#16 0x00007ffff67ea013 in std::unique_ptr<mir::renderer::gl::Context, std::default_delete<mir::renderer::gl::Context> >::~unique_ptr (this=0x5555556d8368) at /usr/include/c++/15/bits/unique_ptr.h:398
#17 0x00007ffff67e9827 in mir::graphics::common::EGLContextExecutor::~EGLContextExecutor (this=0x5555556d8360) at /home/alan.griffiths@canonical.com/canonical/mir/src/platform/graphics/egl_context_executor.cpp:37
#18 0x00007ffff45b9e92 in std::destroy_at<mir::graphics::common::EGLContextExecutor> (__location=0x5555556d8360) at /usr/include/c++/15/bits/stl_construct.h:88
#19 0x00007ffff45b9de9 in std::_Destroy<mir::graphics::common::EGLContextExecutor> (__pointer=0x5555556d8360) at /usr/include/c++/15/bits/stl_construct.h:164
#20 0x00007ffff45b911a in std::allocator_traits<std::allocator<void> >::destroy<mir::graphics::common::EGLContextExecutor> (__p=0x5555556d8360) at /usr/include/c++/15/bits/alloc_traits.h:819
#21 std::_Sp_counted_ptr_inplace<mir::graphics::common::EGLContextExecutor, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose (this=0x5555556d8350) at /usr/include/c++/15/bits/shared_ptr_base.h:615
#22 0x00007ffff45af569 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release (this=0x5555556d8350) at /usr/include/c++/15/bits/shared_ptr_base.h:345
#23 0x00007ffff45b0dc3 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count (this=0x55555563a518) at /usr/include/c++/15/bits/shared_ptr_base.h:1069
#24 0x00007ffff45afb50 in std::__shared_ptr<mir::graphics::common::EGLContextExecutor, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr (this=0x55555563a510) at /usr/include/c++/15/bits/shared_ptr_base.h:1531
#25 0x00007ffff45afb70 in std::shared_ptr<mir::graphics::common::EGLContextExecutor>::~shared_ptr (this=0x55555563a510) at /usr/include/c++/15/bits/shared_ptr.h:175
#26 0x00007ffff45ac966 in mir::graphics::egl::generic::BufferAllocator::~BufferAllocator (this=0x55555563a500) at /home/alan.griffiths@canonical.com/canonical/mir/src/platforms/renderer-generic-egl/buffer_allocator.cpp:171
```

Looks like something is holding on to graphics resources during shutdown and to a point they cannot be released (because the platform that allocated them is no longer available)

examples/miral-custom-decorations/main.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <miral/decorations.h>
+#include <miral/keymap.h>
+#include <miral/minimal_window_manager.h>
+#include <miral/runner.h>
+#include <miral/set_window_management_policy.h>
+#include <miral/wayland_extensions.h>
+
+#include <memory>
+
+extern std::shared_ptr<miral::DecorationStrategy> make_my_custom_decoration();
```
mattkae: I'd prefer this as a `static` on the class for readability's sake.

examples/miral-custom-decorations/custom_decoration.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+#include <miral/decoration_strategy.h>
+
+#include <filesystem>
+#include <memory>
+#include <optional>
+
+std::shared_ptr<miral::DecorationStrategy> make_my_custom_decoration();
+
+class MyCustomDecoration : public miral::DecorationStrategy
```
mattkae: I am still skeptical of the `miral::DecorationStrategy` as an overall direction. This requires all of the drawing of decorations to be done by something that can fill a `mir::graphics::Buffer`, which means that we can never use a true toolkit to fill these buffers, which in turn means that we can never render something _particularly_ interesting in decorations (e.g. an animating background that displays text in multiple locales, has tooltips, images, etc.).

My idea is as follows, but I'm sure others will think differently:
1. Drawing should be done via a _true_ Wayland client (maybe just internal to start, but external could be possible)
2. That client will use foreign top level (and maybe shell-internal capabilities) to close, minimize, maximize, and transform the window that it surrounds.
3. That client will react to regular Wayland clicks, keyboard inputs, and more to render itself
4. The window manager will _somehow_ know how to place that client's surfaces around its parent surface (perhaps this is a good opportunity to design a Wayland protocol that does just that).

@RAOF @AlanGriffiths I am not sure if you agree with this direction, but this would unlock the best possible solution of decorations (in my opinion).

AlanGriffiths: > @RAOF @AlanGriffiths I am not sure if you agree with this direction, but this would unlock the best possible solution of decorations (in my opinion).

I think this direction could allow customization of the current decorations within the limitations of the current framework. Further, even with these limitations I believe this would be of use downstream. I was thinking along those lines with #3708: I even gave an example (which introduced transparency and shifted the buttons to the other side).

Because the approach is inherently limited we have aimed to do something something better. In particular, we've been making it easier to do rendering within the process. OTOH we don't want to start implementing an entire toolkit within Mir and that's a slippery slope to avoid.

For perfectly sound reasons, we've not actually delivered the hypothetical "better solution" for more than a year. Much of that time it has been "real soon now :tm:".

I don't mind a contributor doing something basic now and getting an immediate benefit, even if it is later then deprecated it in favor of something better (when we have it).

src/server/symbols.map
tsimonq2: Doesn't this specific symbols file have all of it in one block?

tests/include/mir_test_framework/stubbed_server_configuration.h
tsimonq2: These two `using` declarations follow the established pattern used throughout `StubbedServerConfiguration` (and `TestingServerConfiguration`).

They bring `the_decoration_renderer_strategy()` and `set_the_decoration_renderer_strategy()` into scope so that test code and fixtures can access/override the custom decoration rendering strategy the same way they already do for `the_decoration_strategy`, `the_buffer_allocator`, etc.

include/miral/miral/decorations.h
```diff
@@ -56,6 +55,21 @@ class Decorations
...
+    static auto strategy(std::shared_ptr<DecorationStrategy> custom) -> CustomDecorations;
+
+    void operator()(mir::Server& server) const;
+
+private:
+    struct Self;
+    std::shared_ptr<Self> self;
+
+    explicit CustomDecorations(std::shared_ptr<Self> self);
+};
```
AlanGriffiths: The range of constructors and "named constructors" is confusing and unnecessary

```suggestion
struct CustomDecorations
{
    explict CustomDecorations(std::shared_ptr<DecorationStrategy> custom);

    void operator()(mir::Server& server) const;

private:
    struct Self;
    std::shared_ptr<Self> self;
};
```

src/include/server/mir/shell/decoration/decoration_strategy.h
```diff
@@ -0,0 +1,178 @@
...
+    virtual auto render_titlebar() -> std::optional<std::shared_ptr<graphics::Buffer>> = 0;
+    virtual auto render_left_border() -> std::optional<std::shared_ptr<graphics::Buffer>> = 0;
+    virtual auto render_right_border() -> std::optional<std::shared_ptr<graphics::Buffer>> = 0;
+    virtual auto render_bottom_border() -> std::optional<std::shared_ptr<graphics::Buffer>> = 0;
+};
+
+/// Internal free function used to render title text using Mir's FreeType-backed renderer.
+/// This is the bridging mechanism for `miral::DecorationStrategy::render_title_text` default impl
+/// and internal decoration title rendering. It is **not** a general public extension point
+/// and is subject to change.
```
AlanGriffiths: As this function is only used internally and `DecorationStrategy::render_title_text()` is available in the stable miral interface I don't see why is is part of the proposed API.

include/miral/miral/decorations.h
```diff
@@ -56,6 +55,21 @@ class Decorations
 
     std::shared_ptr<Self> self;
 };
+
+struct CustomDecorations
+{
```
AlanGriffiths: We use the `class` keyword for types with behaviour

src/server/shell/decoration/decoration_strategy.h
```diff
@@ -133,19 +137,32 @@ class RendererStrategy
     virtual auto render_bottom_border() -> std::optional<std::shared_ptr<graphics::Buffer>> = 0;
 };
 
+/// Internal free function used to render title text using Mir's FreeType-backed renderer.
+/// This is the bridging mechanism for `miral::DecorationStrategy::render_title_text` default impl
+/// and internal decoration title rendering. It is **not** a general public extension point
+/// and is subject to change.
```
AlanGriffiths: This is an internal header, so most of this comment is redundant. What remains is a record of how it is used now, not the documentation of the function contract.

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,178 @@
...
+    auto data() -> uint8_t*;
+    auto stride() const -> mir::geometry::Stride;
+    auto size() const -> mir::geometry::Size;
+    auto format() const -> MirPixelFormat;
+    auto pixels32() -> uint32_t*;
+
+private:
+    struct Self;
+    std::unique_ptr<Self> self;
+};
```
AlanGriffiths: Dumping these types directly into the miral namespace seems untidy. They are very specific to decorations rendering. Without that context a name such as `Button` might raise a very different expectation. Maybe nest them inside `DecorationStrategy`.

AlanGriffiths (commented):
Surface stuff:

> Earlier approaches involving a plugin loader were set aside to maintain a simple and stable ABI.

This didn't happen, no need to record it in git history

> No new plugin loading or registration mechanisms are introduced.

This didn't happen, no need to record it in git history


>   New Symbols 🟢🟢🟢
>     miral::CustomDecorations::CustomDecorations*;
>     miral::CustomDecorations::operator*;
>     miral::CustomDecorations::strategy*;
>     miral::DecorationStrategy::?DecorationStrategy*;
>     miral::DecorationStrategy::bottom_border_height*;
>     miral::DecorationStrategy::create_software_buffer*;
>     miral::DecorationStrategy::map_software_buffer*;
>     miral::DecorationStrategy::render_title_text*;
>     miral::DecorationStrategy::set_buffer_allocator*;
>     miral::DecorationStrategy::side_border_width*;
>     miral::DecorationStrategy::titlebar_height*;
>     miral::SoftwareBufferMapping::?SoftwareBufferMapping*;
>     miral::SoftwareBufferMapping::SoftwareBufferMapping*;
>     miral::SoftwareBufferMapping::data*;
>     miral::SoftwareBufferMapping::format*;
>     miral::SoftwareBufferMapping::pixels32*;
>     miral::SoftwareBufferMapping::size*;
>     miral::SoftwareBufferMapping::stride*;
>     miral::WindowState::titlebar_height*;
>     miral::WindowState::titlebar_width*;
>     non-virtual?thunk?to?miral::DecorationStrategy::?DecorationStrategy*;
>     non-virtual?thunk?to?miral::DecorationStrategy::bottom_border_height*;
>     non-virtual?thunk?to?miral::DecorationStrategy::render_title_text*;
>     non-virtual?thunk?to?miral::DecorationStrategy::set_buffer_allocator*;
>     non-virtual?thunk?to?miral::DecorationStrategy::side_border_width*;
>     non-virtual?thunk?to?miral::DecorationStrategy::titlebar_height*;
>     typeinfo?for?miral::Button;
>     typeinfo?for?miral::CustomDecorations;
>     typeinfo?for?miral::DecorationStrategy;
>     typeinfo?for?miral::InputState;
>     typeinfo?for?miral::SoftwareBufferMapping;
>     typeinfo?for?miral::WindowState;
>     vtable?for?miral::DecorationStrategy;

https://canonical.com/mir/docs/2.19/how-to/how-to-update-symbols-map/#how-to-update-miral-symbols

It is confusing to have two identical files:

    diff src/server/shell/decoration/decoration_strategy.h src/include/server/mir/shell/decoration/decoration_strategy.h

include/miral/miral/decorations.h
```diff
@@ -56,6 +55,21 @@ class Decorations
...
+    CustomDecorations();
```
mattkae: Why would you want to create an empty Custom Decoration?

Also - should you be able to change the custom decorations at runtime via setter?

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,178 @@
...
+        std::string const& text,
+        mir::geometry::Point top_left,
+        mir::geometry::Height height_pixels,
+        uint32_t color);
+    virtual void set_buffer_allocator(std::shared_ptr<mir::graphics::GraphicBufferAllocator> const& allocator);
+
+protected:
+    auto create_software_buffer(mir::geometry::Size size) -> std::shared_ptr<mir::graphics::Buffer>;
+
+    auto map_software_buffer(std::shared_ptr<mir::graphics::Buffer> const& buffer) -> SoftwareBufferMapping;
```
mattkae: All of these buffer creator methods are very odd to have in the public API.

We probably need some other API to manage internally buffers. Users shouldn't have to interact deeply with `mir::graphics` to fill buffers.

examples/miral-custom-decorations/miral-custom-decorations.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+        return {{x, y}, {button_width, button_height}};
+    }
+
+    // Caches WindowState and InputState for render methods.
+    void update_state(WindowState const& ws, InputState const& input) override
+    {
+        current_ws = ws;
+        current_input = input;
+    }
```
mattkae: This is a weird API :)

Why don't we just pass the input and window states each render method instead of caching them?

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,178 @@
...
+    mir::geometry::Size window_size;
+    BorderType border_type;
+    mir::geometry::Height titlebar_height_; // supports titlebar_height() accessor
+    mir::geometry::Width side_border_width;
+    mir::geometry::Height bottom_border_height;
+    float scale;
+
+    mir::geometry::Width titlebar_width() const { return window_size.width; }
+    mir::geometry::Height titlebar_height() const { return titlebar_height_; }
+};
```
mattkae: Can we somehow associate this with a `miral::Window` as well? Should all of the accessors here be `const` since users should just consume it

mattkae (changes requested):
I am okay with the overall decoration, but we should reconsider parts of this API. It would be a good idea to separate out the buffer allocation bits from the decorations in my opinion. They are almost two different problems that need solving separately.

AlanGriffiths:
@tsimonq2 is this still active? Can we expect an update? When?

Saviq:
I'll move it to draft :+1:

doc/sphinx/how-to/custom-server-side-decorations.md
```diff
@@ -0,0 +1,165 @@
...
+
+| Type                                                    | Role                                                                    |
+| ------------------------------------------------------- | ----------------------------------------------------------------------- |
+| {class}`miral::Decorations` / `mir::DecorationStrategy` | SSD vs CSD **negotiation** with clients                                 |
+| {class}`miral::CustomDecorations`                       | Installs a custom SSD **renderer** (`the_decoration_renderer_strategy`) |
+| {class}`miral::DecorationStrategy`                      | Public API your compositor implements to paint SSD buffers              |
+
+You typically need both `miral::Decorations::prefer_ssd()` (or `always_ssd()`) **and**
+{class}`miral::CustomDecorations`: the former ensures clients use server-side decorations;
+the latter replaces how those decorations are drawn.
```
AlanGriffiths: AFAICS `prefer_csd` and `CustomDecorations` is also a meaningful combination

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,180 @@
...
+    private:
+        uint64_t window_id_{};
+        std::string window_name_;
+        MirWindowFocusState focused_state_{};
+        mir::geometry::Size window_size_{};
+        BorderType border_type_{BorderType::none};
+        mir::geometry::Height titlebar_height_{};
+        mir::geometry::Width side_border_width_{};
+        mir::geometry::Height bottom_border_height_{};
+        float scale_{1.0f};
```
AlanGriffiths: We avoid publishing internal structures as part of the API - changes should not necessitate an ABI break.

include/miral/miral/decorations.h
```diff
@@ -56,6 +59,24 @@ class Decorations
...
+public:
+    explicit CustomDecorations(std::shared_ptr<DecorationStrategy> strategy);
+
+    void operator()(mir::Server& server) const;
+
+private:
+    struct Self;
+    std::shared_ptr<Self> self;
+};
+
```
AlanGriffiths: There seems to be no reason to add this to `decorations.h`.

Clearly it is more closely related to `DecorationStrategy` (see constructor). Maybe both `CustomDecorations` and `DecorationStrategy` belong in the same `custom_decorations.h` header?

And maybe also `DecorationBuffer` and `DecorationSurface` belong there too - I don't see any of them being used independently.

include/miral/miral/decoration_surface.h
```diff
@@ -0,0 +1,57 @@
...
+    struct Self;
+    std::unique_ptr<Self> self;
+
+    explicit DecorationSurface(std::shared_ptr<mir::graphics::Buffer> buffer);
+
+    auto take_buffer() -> std::shared_ptr<mir::graphics::Buffer>;
+
+    friend class DecorationBuffers;
+    friend class RendererAdapter;
+};
```
AlanGriffiths: This class name and interface seem very odd. The description as an "Opaque software buffer" seems at odds with the name and the public behaviours are "create" and "move".

src/include/server/mir/shell/decoration/decoration_strategy.h
```diff
@@ -109,6 +109,8 @@ class WindowState
 
     auto scale() const -> float;
 
+    auto window_id() const -> uint64_t;
+
```
AlanGriffiths, line 114: Is this any use?

include/miral/miral/decoration_strategy.h
```diff
@@ -0,0 +1,180 @@
...
+        auto buttons() const -> std::vector<Button> const& { return buttons_; }
+
+    private:
+        std::vector<Button> buttons_;
+    };
+
+    /// Per-window immutable state for a single render call.
+    /// window_id() is an opaque token stable for the lifetime of the decorated
+    /// window; it is not a miral::Window handle and is not preserved if the
+    /// underlying surface is destroyed and recreated.
```
AlanGriffiths: I don't see `window_id()` used anywhere. What is it for?

AlanGriffiths (changes requested):
This looks a bit cleaner. But could be tidied up further.

There are also some build errors to fix. E.g. https://github.com/canonical/mir/actions/runs/28751653969/job/85251733001?pr=4991

henrikschnor:
We're in the process of switching to Mir2 with Lomiri and I recently noticed that we will need something similar for Lomiri's custom decorations. Without having looked into all the details, I think either approach that was suggested here would work for us. I.e. either an API for creating a `miral::DecorationStrategy` that is then used by Mir to draw the decorations, or simply the possibility to disable Mir's own decorations and draw them by other means (which is what Lomiri currently does, though, not as a wayland client).

@tsimonq2 are you intending to continue working on this?

PR #4994, opened 2026-06-03
Replace typedef with using

robert-ancell:
This is not an exhaustive replace, just fixing the simplest cases.


include/test/mir/test/doubles/mock_egl.h
```diff
@@ -111,8 +111,8 @@ class MockEGL
     // eglSwapBuffers to invoke EGLNativeSurface::swap_buffers_sync
     void provide_stub_platform_buffer_swapping();
 
-    typedef void (*generic_function_pointer_t)(void);
-    typedef void* AnyNativeType;
+    using generic_function_pointer_t = void (*)(void);
```
tarek-y-ismail, line 114: I did not know you could do this

tarek-y-ismail (approved):

PR #4995, opened 2026-06-04
core: Rename “core” tag to “base”

RAOF:


tarek-y-ismail (approved):
Simple and straightforward

PR #4998, opened 2026-06-04
Pushing the 2-step initialization required by Dispatch implementation s from C++ to Rust so that C++ implementers need not think about it

mattkae:
## What's new?
**Problem**: The `C++` implementations in `wayland-rs` had a two step initialization process previously: first we constructed them, and then we called `associate` to associate the Rust middleware with the C++ implementation. The reason for this is that C++ needs access to the protocol object returned by `data_init.init` so that it can perform actions on it, but the `data_init.init` needs to associate with the data, which was the C++ class.

This led to a whole host of problems where work done inside of the C++ constructor was invalid because we didn't yet have a middleware `instance` yet.

**Solution**: implement a wrapper around each type in Rust. That wrapper will be the data associated with each `Dispatch` protocol. Afterward, we can initialize and use the C++ implementation directly without having to consider if it is fully initialized.

**Downsides**: we are pushing the 2-step initialization from C++ to Rust. We cannot make it go away fully, but this lessens the cognitive overhead when implementing a protocol in C++ considerably.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/build_script/cpp_protocol_generation.rs
+++ b/src/server/frontend_wayland/wayland_rs/build_script/cpp_protocol_generation.rs
@@ -78,7 +78,9 @@ fn create_global_factory(protocols: &Vec<WaylandProtocol>) -> CppBuilder {
             .filter(|interface| interface.name != "wl_display" && interface.name != "wl_registry")
             .for_each(|global_interface| {
                 let class_name = format_wayland_interface_to_cpp_class(&global_interface.name);
+                let ext_name = format_wayland_interface_to_rust_extension_struct(&global_interface.name);
                 namespace.add_forward_declaration_class(&class_name);
+                namespace.add_forward_declaration_struct(&snake_to_pascal(&ext_name));
 
                 let mut method = CppMethod::new(
                     format!("create_{}", global_interface.name),
@@ -93,6 +95,12 @@ fn create_global_factory(protocols: &Vec<WaylandProtocol>) -> CppBuilder {
                     "client",
                     false,
                 ));
+                method.add_arg(CppArg::new(
+                    CppType::Box(snake_to_pascal(&ext_name)),
+                    "instance",
+                    false,
+                ));
+                method.add_arg(CppArg::new(CppType::CppU32, "object_id", false));
                 class.add_method(method);
             })
     });
@@ -212,7 +220,6 @@ fn create_cpp_builder(protocol: &WaylandProtocol) -> CppBuilder {
     builder.add_header_include("\"ffi_fwd.h\"");
     builder.add_header_include("\"weak.h\"");
     builder.add_header_include("<memory>");
-    builder.add_header_include("<optional>");
     builder.add_header_include("<cstdint>");
     builder.add_header_include("<rust/cxx.h>");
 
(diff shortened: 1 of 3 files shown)


src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -173,26 +167,39 @@ fn transform_argument_for_cpp(arg: &WaylandArg) -> Option<TokenStream> {
...
-                    let #arg_name: &cxx::SharedPtr<ffi::#arg_type> = match #arg_name.as_ref() {
-                        Some(o) => o.data().unwrap(),
-                        None => &cxx::SharedPtr::<ffi::#arg_type>::null()
+                    let #arg_name: cxx::SharedPtr<ffi::#arg_type> = match #arg_name.as_ref() {
+                        Some(o) => {
+                            let guard = o.data::<Arc<Mutex<#wrapper_type>>>().unwrap().lock().unwrap();
+                            guard.inner.clone().unwrap_or_else(|| cxx::SharedPtr::null())
+                        }
+                        None => cxx::SharedPtr::<ffi::#arg_type>::null()
                     };
```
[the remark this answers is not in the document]
mattkae: For later

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -173,26 +167,39 @@ fn transform_argument_for_cpp(arg: &WaylandArg) -> Option<TokenStream> {
...
+                    let #arg_name = &#arg_name;
                 })
             } else {
                 Some(quote! {
-                    let #arg_name: &cxx::SharedPtr<ffi::#arg_type> = #arg_name.data().unwrap();
+                    let #arg_name: cxx::SharedPtr<ffi::#arg_type> = {
+                        let guard = #arg_name.data::<Arc<Mutex<#wrapper_type>>>().unwrap().lock().unwrap();
+                        guard.inner.clone().unwrap_or_else(|| cxx::SharedPtr::null())
+                    };
+                    let #arg_name = &#arg_name;
```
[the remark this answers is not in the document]
mattkae: There was an issue with this, and this is safe for now. We have a task to clean this up too.

src/server/frontend_wayland/wayland_rs/build_script/cpp_protocol_generation.rs
```diff
@@ -457,11 +488,16 @@ fn wayland_request_to_cpp_method(method: &WaylandRequest) -> Vec<CppMethod> {
     if !wayland_weak_transformers.is_empty() {
         // Build the body: wrap shared_ptrs in Weak, then delegate to the virtual overload
         let return_prefix = if has_retval { "return " } else { "" };
+        let mut all_delegation_args = delegation_args;
```
tarek-y-ismail: `delegation_args` looks like it's only used here. Maybe we can inline it? You'll probably have to clone it one more time in the for loop at line 474.

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -173,26 +167,39 @@ fn transform_argument_for_cpp(arg: &WaylandArg) -> Option<TokenStream> {
...
         WaylandArgType::Object => {
             let arg_type = format_ident!(
                 "{}",
                 format_wayland_interface_to_cpp_class(
                     arg.interface.as_ref().expect("Object is missing interface"),
                 )
             );
+            let wrapper_type = format_ident!(
+                "{}Wrapper",
+                snake_to_pascal(arg.interface.as_ref().expect("Object is missing interface"))
```
tarek-y-ismail, line 181: Probably something for later, but maybe things that are passed often to functions like `snake_to_pascal` can be stored in a struct that implements `to_pascal` and such?

Generally, it feels like lots of patterns are repeating. It may pay off to see what can be refactored out. Maybe Copilot/Claude can help with that?

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -258,30 +274,27 @@ fn generate_request_body(request: &WaylandRequest) -> TokenStream {
...
 
         quote! {
+            // Step 1: Create a wrapper with no inner value and register it via data_init.
+            let child_wrapper = Arc::new(Mutex::new(#child_wrapper_struct_name { inner: None }));
+            let instance = data_init.init(#new_id_name, child_wrapper.clone());
+            register_resource(&instance);
+            let protocol_id = Resource::id(&instance).protocol_id();
+
+            // Step 2: Create the middleware object wrapping the child resource.
+            let boxed = Box::new(crate::middleware::#ext_interface_struct_name{ wrapped: instance });
```
tarek-y-ismail, line 284: This looks duplicated from line 124. Maybe we can factor this out?

tarek-y-ismail (approved):
A  bit tough to wrap one's head around. But the fact that C++ objects are going to be initialized correctly is worth it.

I've got some non-blocking nits 👇

src/server/frontend_wayland/wayland_rs/build_script/cpp_protocol_generation.rs
mattkae: Sure, we can give `args` the `.clone()`

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
mattkae: Yeah maybe for later :)

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
mattkae: I'll keep it explicit for now since its generator code and I want the pathing to be so obvious

PR #5000, opened 2026-06-05
graphics: add bytes-per-pixel to DRMFormat and use it for SHM stride validation

robert-ancell:
DRMFormat now exposes the storage size of a pixel via
DRMFormat::Info::bytes_per_pixel(), backed by an explicit field in the
internal format table. This is stored rather than derived because opaque
"X" formats (e.g. XRGB8888) are padded and so cannot be computed from their
component bit counts alone.

shm.cpp's create_buffer() previously hard-coded 4 bytes-per-pixel when
checking that a client's stride was large enough. It now validates the
requested format first, looks up its bytes-per-pixel, and performs the
stride check with overflow-safe 64-bit arithmetic. A supported format with
missing metadata is rejected rather than silently bypassing the check.

Adds unit tests for bytes_per_pixel(), including the padded-format case.

### DIFF
--- a/include/platform/mir/graphics/drm_formats.h
+++ b/include/platform/mir/graphics/drm_formats.h
@@ -50,7 +50,7 @@ public:
          * \note This includes any padding (for example, XRGB8888 reports 4
          *       even though only 24 of its bits carry colour information).
          */
-        auto bytes_per_pixel() const -> uint32_t;
+        auto bytes_per_pixel() const -> size_t;
 
         auto components() const -> std::optional<RGBComponentInfo> const&;
     private:
--- a/include/platform/mir/graphics/drm_formats.h
+++ b/include/platform/mir/graphics/drm_formats.h
@@ -44,6 +44,14 @@ public:
 
         bool has_alpha() const;
 
+        /**
+         * The number of bytes occupied by a single pixel of this format.
+         *
+         * \note This includes any padding (for example, XRGB8888 reports 4
+         *       even though only 24 of its bits carry colour information).
+         */
+        auto bytes_per_pixel() const -> uint32_t;
+
         auto components() const -> std::optional<RGBComponentInfo> const&;
     private:
         friend class DRMFormat;
--- a/src/server/frontend_wayland/shm.cpp
+++ b/src/server/frontend_wayland/shm.cpp
@@ -308,7 +308,7 @@ void mf::ShmPool::create_buffer(
     auto const format_supported = std::any_of(
         supported_formats->begin(),
         supported_formats->end(),
-        [&](auto supported) { return static_cast<uint32_t>(supported) == static_cast<uint32_t>(drm_format); });
+        [&drm_format](auto supported) { return supported == drm_format; });
     if (!format_supported)
     {
         throw wayland::ProtocolError{
--- a/src/server/frontend_wayland/shm.cpp
+++ b/src/server/frontend_wayland/shm.cpp
@@ -316,6 +316,7 @@ void mf::ShmPool::create_buffer(
             wayland::Shm::Error::invalid_format,
             "Missing pixel-size information for SHM format requested"};
     }
+    // Casting to stop the below calculations being unsigned and then comparing against signed width/height.
     auto const bytes_per_pixel = static_cast<geometry::Size::ValueType>(format_info->bytes_per_pixel());
 
     auto const max_size = std::numeric_limits<geometry::Size::ValueType>::max();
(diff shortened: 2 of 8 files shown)


robert-ancell:
Builds on https://github.com/canonical/mir/pull/4999 - review that first

src/server/frontend_wayland/shm.cpp
```diff
@@ -268,23 +288,36 @@ void mf::ShmPool::create_buffer(
...
+    auto const bytes_per_pixel = format_info->bytes_per_pixel();
+
+    if (static_cast<int64_t>(stride) < static_cast<int64_t>(width) * bytes_per_pixel)
+    {
+        throw wayland::ProtocolError{
+            resource,
+            wayland::Shm::Error::invalid_stride,
+            "Invalid stride %d (too small for width %d. Did you specify stride in pixels?)",
+            stride, width};
     }
```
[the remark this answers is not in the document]
robert-ancell: This is fixed in https://github.com/canonical/mir/pull/4885

src/server/frontend_wayland/shm.cpp
```diff
@@ -275,18 +294,32 @@ void mf::ShmPool::create_buffer(
...
+    }
+
+    auto const format_info = drm_format.info();
+    if (!format_info)
     {
         throw wayland::ProtocolError{
             resource,
             wayland::Shm::Error::invalid_format,
-            "Invalid SHM format %u", format};
+            "Missing pixel-size information for SHM format requested"};
```
AlanGriffiths, line 317: We are reporting less information (`format`) than the existing code

src/server/frontend_wayland/shm.cpp
```diff
@@ -275,18 +294,32 @@ void mf::ShmPool::create_buffer(
...
             "Invalid SHM buffer stride %d", stride};
     }
-    // TODO: Pull supported formats out of RenderingPlatform to support more than the required formats
-    if (format != wayland::Shm::Format::argb8888 && format != wayland::Shm::Format::xrgb8888)
+
+    auto const drm_format = wl_shm_format_to_drm_format(format);
+    auto const format_supported = std::any_of(
+        supported_formats->begin(),
+        supported_formats->end(),
+        [&](auto supported) { return static_cast<uint32_t>(supported) == static_cast<uint32_t>(drm_format); });
```
AlanGriffiths: Why do we need the cast? And why to `uint32_t`? Are both sides not `DRMFormat`?

src/server/frontend_wayland/shm.cpp
```diff
@@ -275,18 +294,32 @@ void mf::ShmPool::create_buffer(
...
+    if (!format_info)
     {
         throw wayland::ProtocolError{
             resource,
             wayland::Shm::Error::invalid_format,
-            "Invalid SHM format %u", format};
+            "Missing pixel-size information for SHM format requested"};
     }
-    auto const pixel_size = 4; // All supported formats are 4 bytes per pixel
+    auto const bytes_per_pixel = static_cast<geometry::Size::ValueType>(format_info->bytes_per_pixel());
```
AlanGriffiths, line 320: On the surface, this seems a very strange choice of type: the size of a pixel is a number of bytes, which has nothing directly to do with "the underlying type of a `geometry::Size`".

Why are we casting at all? I guess because `bytes_per_pixel()` returns an unsigned type of the same size as `geometry::Size::ValueType` and therefore the following calculations "promote" to unsigned because of the arbitrary rules we inherited from C?

include/platform/mir/graphics/drm_formats.h
```diff
@@ -44,6 +44,14 @@ class DRMFormat
 
         bool has_alpha() const;
 
+        /**
+         * The number of bytes occupied by a single pixel of this format.
+         *
+         * \note This includes any padding (for example, XRGB8888 reports 4
+         *       even though only 24 of its bits carry colour information).
+         */
+        auto bytes_per_pixel() const -> uint32_t;
```
AlanGriffiths: Returning `uint32_t` causes problems in the one place this is used (which actually needs a signed type). Wouldn't `int` be more useful?

AlanGriffiths (approved):
I am not sure whether copying around `shared_ptr<>`s is better than simply copying the referenced `vector`, but 🤷 .

There are a few nits to, optionally, address. But otherwise good.

include/platform/mir/graphics/drm_formats.h
robert-ancell: Changed to size_t

src/server/frontend_wayland/shm.cpp
robert-ancell: Yes, it makes max_width and min_stride unsigned, which then hits a compile error checking against width which is signed (though checked to be positive earlier in the function). I tried moving the cast around or replacing the auto types with explicit types but it makes it more confusing - I'll add a comment.

PR #5003, opened 2026-06-07
Implement wl_region::subtract and tidy wl_region::add

robert-ancell:
WlRegion::subtract was a no-op that merely logged a warning, so surface input/opaque regions and pointer-constraint regions built with wl_region.subtract were incorrect (the subtracted area was retained).

Implement region subtraction by decomposing each stored rectangle minus the hole into up to four non-overlapping slices, which is correct for the union semantics used by all consumers of rectangle_vector(). The core wl_region interface defines no protocol errors, so degenerate (non-positive) holes are treated as subtracting nothing rather than raised as errors.

WlRegion::add now ignores degenerate (non-positive) rectangles and drops any existing rectangle wholly contained by the new one, so redundant rectangles no longer accumulate in the region.

This behaviour is exercised by the WLCS test suite.


src/server/frontend_wayland/wl_region.cpp
```diff
@@ -42,14 +42,58 @@ mf::WlRegion* mf::WlRegion::from(wl_resource* resource)
...
+            result.push_back(rect);
+            continue;
+        }
+
+        auto const rx1 = rect.left().as_int();
+        auto const rx2 = rect.right().as_int();
+        auto const ry1 = rect.top().as_int();
+        auto const ry2 = rect.bottom().as_int();
+
+        // The portion of the hole that actually intersects this rectangle.
```
tarek-y-ismail:
```suggestion
        // The portion of the hole that actually intersects this rectangle:
        //  rx1      hx1     hx2     rx2
        //   +---------+-------+-------+  ry1
        //   |         |       |       |
        //   |         +-------+       |  hy1
        //   |         | (hole)|       |
        //   |         +-------+       |  hy2
        //   |         |       |       |
        //   +---------+-------+-------+  ry2
```

An illustration would be nice.

src/server/frontend_wayland/wl_region.cpp
```diff
@@ -42,14 +42,58 @@ mf::WlRegion* mf::WlRegion::from(wl_resource* resource)
...
+
+        // Decompose rect minus hole into up to four non-overlapping slices.
+        if (ry1 < hy1) // above the hole
+            result.push_back(geom::Rectangle{{rx1, ry1}, {rx2 - rx1, hy1 - ry1}});
+        if (hy2 < ry2) // below the hole
+            result.push_back(geom::Rectangle{{rx1, hy2}, {rx2 - rx1, ry2 - hy2}});
+        if (rx1 < hx1) // left of the hole
+            result.push_back(geom::Rectangle{{rx1, hy1}, {hx1 - rx1, hy2 - hy1}});
+        if (hx2 < rx2) // right of the hole
+            result.push_back(geom::Rectangle{{hx2, hy1}, {rx2 - hx2, hy2 - hy1}});
```
tarek-y-ismail, line 103: Small, non-blocking note on this: I believe we can add a "consolidation" stage after this to group up the smaller rectangles into fewer larger ones. But I don't believe our use cases will be extreme enough to need this (yet?)

tarek-y-ismail (changes requested):
> This behaviour is exercised by the WLCS test suite.

I believe you've been deceived again, or my grepping skills are not good enough.

tarek-y-ismail:
@robert-ancell poke

src/server/frontend_wayland/wl_region.cpp
robert-ancell: Yeah, I doubt this will be worth it - they would have to be very exactly aligned for any consolidation to work.

robert-ancell:
Tests added in https://github.com/canonical/wlcs/pull/453

tarek-y-ismail (approved):
👍

PR #5004, opened 2026-06-07
Raise invalid_offset for non-zero wl_surface.attach offset (v5+)

robert-ancell:
Since wl_surface version 5, the x/y offset arguments to wl_surface.attach must be zero; the offset is instead supplied via wl_surface.offset. Passing a non-zero offset is a protocol error (invalid_offset).

Mir advertises wl_surface version 6 but previously only logged a warning and silently dropped the offset for every version. Raise the invalid_offset protocol error for surfaces bound at version 5 or later, while keeping the existing (legacy) warning for older clients where the offset is permitted.

This behaviour is exercised by the WLCS test suite.


tarek-y-ismail (approved):
LGTM, just can't find the tests

AlanGriffiths (approved):

PR #5005, opened 2026-06-07
Raise the correct protocol error on a wl_surface role conflict

robert-ancell:
Assigning a second role to a wl_surface previously threw a generic std::runtime_error from WlSurface::set_role, which surfaces to the client as an internal error/disconnect rather than the protocol-specific error the Wayland specs require.

Add a WlSurface::has_role() accessor and check it at each request that assigns a role, raising the error mandated by the relevant protocol before the role object is constructed:

  - wl_subcompositor.get_subsurface  -> wl_subcompositor.bad_surface
  - xdg_wm_base.get_xdg_surface      -> xdg_wm_base.role
  - zwlr_layer_shell_v1.get_layer_surface -> zwlr_layer_shell_v1.role
  - wl_pointer.set_cursor            -> wl_pointer.role
  - wl_data_device.start_drag (icon) -> wl_data_device.role

The xdg check is placed at get_xdg_surface to match the spec wording ("illegal to create an xdg_surface for a wl_surface which already has an assigned role"). Re-using the current cursor surface to update its hotspot remains allowed, as required by the spec. The generic throw in set_role is retained as a defensive backstop.

This behaviour is covered by WLCS.


src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -477,6 +478,16 @@ void mf::WlPointer::set_cursor(
...
+            // The surface is becoming a cursor for the first time (for this pointer).
+            // Re-using the current cursor surface is allowed and handled above, but a
+            // surface with any other role must raise a protocol error.
+            if (wl_surface->has_role())
+            {
+                throw mw::ProtocolError{
+                    resource,
+                    Error::role,
+                    "Surface already has a role"};
+            }
```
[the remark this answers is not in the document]
robert-ancell: This is already noted in the comment - the if branch above is already handling the case where it is already a cursor.

robert-ancell:
Failing in WLCS test - will need checking.

robert-ancell:
Oh, now I understand. The failure was:
```
30/55 Test #30: wlcs_XdgSurfaceStableTest.creating_xdg_surface_from_wl_surface_with_existing_role_is_an_error_fails ..............................***Failed    0.39 sec
Note: Google Test filter = XdgSurfaceStableTest.creating_xdg_surface_from_wl_surface_with_existing_role_is_an_error
[==========] Running 1 test from 1 test suite.
[----------] 1 test from XdgSurfaceStableTest
[ RUN      ] XdgSurfaceStableTest.creating_xdg_surface_from_wl_surface_with_existing_role_is_an_error
error in client communication (pid 108260)
xdg_wm_base@12: error 0: Surface already has a role
[       OK ] XdgSurfaceStableTest.creating_xdg_surface_from_wl_surface_with_existing_role_is_an_error (200 ms)
[----------] 1 test from XdgSurfaceStableTest (200 ms total)

[==========] 1 tests from 1 test cases run. (200ms total elapsed)
[  PASSED  ] 1 test
[XPASS]: Test unexpectedly passes
```

Which is because the test now works. The solution is to remove this from `tests/acceptance-tests/wayland/expected_wlcs_failures.list` as it's now fixed.

src/server/frontend_wayland/wl_pointer.cpp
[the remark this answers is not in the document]
RAOF: The branch above isn't actually handling the case where the newly-applied surface already has a cursor role; it's handling the case where the newly-applied surface is the *same* surface as the currently applied one.

Clients could create multiple `WlPointer` instances and set the same surface as a pointer on each; this is meant to work, but would fail.

(Also, we incorretly *unset* the pointer role when a surface is no longer being used as a cursor, meaning a client could set a surface as a cursor, unset that surface, and then (invalidly!) map that surface as an `xdg_toplevel`)

src/server/frontend_wayland/wl_surface.h
```diff
@@ -135,6 +135,10 @@ class WlSurface : public wayland::Surface
     geometry::Displacement total_offset() const { return offset_ + role->total_offset(); }
     std::optional<geometry::Size> buffer_size() const { return buffer_size_; }
     bool synchronized() const;
+    /// Whether the surface currently has a role assigned (e.g. subsurface, cursor,
+    /// drag-and-drop icon, or a window role such as xdg_toplevel). Used to raise the
+    /// appropriate protocol error when a client tries to assign a second role.
+    bool has_role() const { return role != &null_role; }
```
RAOF, line 141: Because roles are sticky, this might want to return the current role so that we can check if they're assigning a different role, or re-assigning back to the same roll.

RAOF (changes requested):
This is not *quite* the right set of checks — from the [`wl_surface` protocol:](https://wayland.app/protocols/wayland#wl_surface)

> A surface can have only one role at a time. Initially a wl_surface does not have a role. Once a wl_surface is given a role, it is set permanently for the whole lifetime of the wl_surface object. Giving the current role again is allowed, unless explicitly forbidden by the relevant interface specification.
> ...
> ...For instance, if a wl_subsurface object is destroyed, the wl_surface it was created for will be unmapped and forget its position and z-order. It is allowed to create a wl_subsurface for the same wl_surface again, but it is not allowed to use the wl_surface as a cursor (cursor is a different role than sub-surface, and role switching is not allowed).

robert-ancell:
Role checking more sophisticated and tests proposed in https://github.com/canonical/wlcs/pull/452 (fail on current Mir, pass with this branch).

AlanGriffiths (approved):
Looks good

src/server/frontend_wayland/wl_subcompositor.cpp
```diff
@@ -66,16 +66,23 @@ void mf::WlSubcompositorInstance::get_subsurface(
...
     }
     WlSurface* child_surface = WlSurface::from(surface);
     WlSurface* parent_surface = WlSurface::from(parent);
+    if (child_surface->has_role())
+    {
+        throw wayland::ProtocolError{
+            resource,
+            mw::Subcompositor::Error::bad_surface,
+            "Surface already has a role"};
+    }
```
AlanGriffiths, line 81: Should this, like `CursorSurfaceRole` also allow the same role? The example @RAOF quotes seems to suggest so.

src/server/frontend_wayland/wl_surface.h
```diff
@@ -135,6 +135,14 @@ class WlSurface : public wayland::Surface
...
+    /// The surface's current role, or nullptr if it has none. Kept role-agnostic: callers
+    /// that care about a specific role (e.g. wl_pointer.set_cursor) discriminate the
+    /// concrete type themselves rather than the surface exposing per-role predicates.
+    auto current_role() const -> WlSurfaceRole* { return has_role() ? role : nullptr; }
```
AlanGriffiths, line 145: Would this be better as:
```suggestion
    template<typename Role>
    bool has_role_other_than() const
    {
        return role != null_role && !dynamic_cast<Role*>(role));
    }
```

AlanGriffiths (commented):
On further consideration...

src/server/frontend_wayland/wl_subcompositor.cpp
andogq: I don't think so given [the spec](https://wayland.app/protocols/wayland#wl_subcompositor:request:get_subsurface):
> The to-be sub-surface must not already have another role [...]. Otherwise the bad_surface protocol error is raised.

RAOF: But in this case it does not already have *another* role; it has a `wl_subsurface` role, so it should be allowed.

src/server/frontend_wayland/wl_subcompositor.cpp
```diff
@@ -66,16 +66,23 @@ void mf::WlSubcompositorInstance::get_subsurface(
...
     {
         throw wayland::ProtocolError{
-            new_subsurface,
+            resource,
             mw::Subcompositor::Error::bad_parent,
             "Surface cannot be its own parent"};
     }
     WlSurface* child_surface = WlSurface::from(surface);
     WlSurface* parent_surface = WlSurface::from(parent);
+    if (child_surface->has_role() || !child_surface->can_set_role<WlSubsurface>())
```
andogq: I'm working off the assumption that unlike `wl_pointer` (which handles clearing an existing role), the child surface must not have any role.

RAOF: I think this actually *is* like the `wl_pointer` case; I think it *is* acceptable to `surface` -> `subsurface` -> `surface` -> `subsurface`.

> The to-be sub-surface must not already have another role, and it must not have an existing wl_subsurface object.

This reads to me as “cannot have a non-subsurface role and cannot have an existing `wl_subsurface` object associated”.

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -478,6 +479,21 @@ void mf::WlPointer::set_cursor(
...
+            {
+                throw mw::ProtocolError{
+                    resource,
+                    Error::role,
+                    "Surface already has a non-cursor role"};
+            }
+            // The surface may already hold a cursor role from a previous (or another
+            // pointer's) cursor; clear it so the new cursor can take the role.
+            if (wl_surface->has_role())
+                wl_surface->clear_role();
```
RAOF: I think we should, instead, do a `can_set_role` in `wl_surface::set_role()`; it seems that most Wayland protocols are happy with re-assigning the same role.

src/server/frontend_wayland/wl_data_device.cpp
```diff
@@ -248,6 +248,11 @@ void mf::WlDataDevice::start_drag(
     {
         auto const icon_surface = WlSurface::from(icon.value());
 
+        if (icon_surface->has_role())
```
RAOF: I think this should be `can_set_role`.

src/server/frontend_wayland/layer_shell_v1.cpp
```diff
@@ -275,9 +275,18 @@ void mf::LayerShellV1::Instance::get_layer_surface(
             "Invalid layer %u", layer};
     }
 
+    auto* const wl_surface = WlSurface::from(surface);
+    if (wl_surface->has_role())
```
RAOF: I think this should be `can_set_role()`.

src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -153,7 +153,15 @@ void mf::XdgShellStable::Instance::create_positioner(wl_resource* new_positioner
 
 void mf::XdgShellStable::Instance::get_xdg_surface(wl_resource* new_shell_surface, wl_resource* surface)
 {
-    new XdgSurfaceStable{new_shell_surface, WlSurface::from(surface), *shell};
+    auto* const wl_surface = WlSurface::from(surface);
+    if (wl_surface->has_role())
```
RAOF, line 159: Whereas I think *this* is correct - the `xdg_shell` spec says:

> It is illegal to create an xdg_surface for a wl_surface which already has anassigned role and this will result in a role error.

Rather than talking about *another* role

src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -196,13 +197,19 @@ void mf::WlSurface::set_role(WlSurfaceRole* role_)
     if (role != &null_role)
```
RAOF: I think this check should be
```suggestion
    if (!can_set_role_type(&typeid(role_)))
```
because, as I read it, most Wayland protocols are happy to re-assign the same role.

RAOF (changes requested):

src/server/frontend_wayland/wl_subcompositor.cpp
andogq: > cannot have an existing `wl_subsurface` object associated

If `child_surface->has_role()`, would you expect the existing role to be cleared (like `wl_pointer`), or a protocol error (what's already here)?

RAOF: Ooooh, right.

This is actually right as-is, isn't it? If the surface didn't have a `wl_subsurface`-role then `!child_surface->can_set_role<WlSubsurface>()` will fire, and correctly send the error. If the surface *currently* has a role then that means that it currently has a `wl_subsurface` object associated with it, which *also* means that a `bad_surface` error should be raised.

Ooof. Maybe this could be in a comment? :)

RAOF (approved):
Yes! This all seems good now 🎉

PR #5006, opened 2026-06-08
Raise invalid_size from xdg_surface.set_window_geometry

robert-ancell:
Per the xdg-shell spec, the width and height of the window geometry must be greater than zero and "setting an invalid size will raise an invalid_size error". The handler raised the generic error code instead of the xdg_surface invalid_size error, so the client received the wrong code.


RAOF (approved):
Yup.

PR #5007, opened 2026-06-08
Validate the serial in xdg_surface.ack_configure

robert-ancell:
ack_configure was a no-op, so a client acking a serial that was never sent (or already acked) went undetected. Per the xdg-shell spec this must raise an invalid_serial error.

Track the serials of configure events sent on each xdg_surface that have not yet been consumed. Acking a serial consumes that serial and every serial sent before it (as the spec requires), so re-acking the same serial or acking one older than the last ack is rejected with invalid_serial, as is acking a serial that was never sent.

This behaviour is covered by WLCS.

### DIFF
--- a/src/server/frontend_wayland/xdg_shell_stable.cpp
+++ b/src/server/frontend_wayland/xdg_shell_stable.cpp
@@ -29,6 +29,9 @@
 
 #include <boost/throw_exception.hpp>
 
+#include <algorithm>
+#include <vector>
+
 namespace mf = mir::frontend;
 namespace ms = mir::scene;
 namespace msh = mir::shell;
@@ -66,6 +69,10 @@ public:
 private:
     mw::Weak<WindowWlSurfaceRole> window_role_;
     mw::Weak<WlSurface> const surface;
+    /// Serials of configure events sent on this xdg_surface that have not yet been
+    /// consumed by an ack_configure, in the order they were sent. Acking a serial
+    /// consumes it and every serial sent before it.
+    std::vector<uint32_t> unacked_configure_serials;
 
 public:
     XdgShellStable const& xdg_shell;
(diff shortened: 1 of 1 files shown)


src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -66,6 +70,10 @@ class XdgSurfaceStable : public mw::XdgSurface
 private:
     mw::Weak<WindowWlSurfaceRole> window_role_;
     mw::Weak<WlSurface> const surface;
+    /// Serials of configure events sent on this xdg_surface that have not yet been
+    /// consumed by an ack_configure, in the order they were sent. Acking a serial
+    /// consumes it and every serial sent before it.
+    std::vector<uint32_t> unacked_configure_serials;
```
tarek-y-ismail, line 76: Non-blocking: I'm not sure what's the scale of events in real use, but a `std::deque` might be a better container if it turns out we add and remove serials rapidly

src/server/frontend_wayland/xdg_shell_stable.cpp
```diff
@@ -251,13 +259,22 @@ void mf::XdgSurfaceStable::set_window_geometry(int32_t x, int32_t y, int32_t wid
 
 void mf::XdgSurfaceStable::ack_configure(uint32_t serial)
 {
-    (void)serial;
-    // TODO
+    auto const it = std::find(unacked_configure_serials.begin(), unacked_configure_serials.end(), serial);
+    if (it == unacked_configure_serials.end())
```
tarek-y-ismail, line 263: > It is an error to ..., or to issue an ack_configure request referencing a serial from a configure event issued before the event identified by the last ack_configure request for the same xdg_surface. Doing so will raise an invalid_serial error.

This feels like we should also be checking `serial >= last_acked_serial` (which would require one more member to be stored). We can't quite depend on `unacked_configure_serials` because the last acked serial is going to be removed from the vector.

tarek-y-ismail (changes requested):
The PR description mentions:

> This behaviour is covered by WLCS.

But I can't quite find any existing tests or PRs to cover this. I'm I looking in the wrong place?

src/server/frontend_wayland/xdg_shell_stable.cpp
AlanGriffiths: We shouldn't need a `serial >= last_acked_serial` check - an `ack()` should consume any prior serials as well as the acked one

robert-ancell:
> The PR description mentions:
> 
> > This behaviour is covered by WLCS.
> 
> But I can't quite find any existing tests or PRs to cover this. I'm I looking in the wrong place?

This is a bit of a lie - this was added by Claude when I told it not to make local tests as this functionality is covered by WLCS. I'm currently adding tests to WLCS to check things like this.

src/server/frontend_wayland/xdg_shell_stable.cpp
tarek-y-ismail: As I understand the snippet above, this is covering for cases like `ack_configure(3)` followed by `ack_configure(2)`. I suppose the linked code should cover that.

robert-ancell:
Tests added in https://github.com/canonical/wlcs/pull/437

tarek-y-ismail (approved):

PR #5008, opened 2026-06-08
frontend_wayland: reject xdg_toplevel max size smaller than min size

robert-ancell:
The xdg-shell protocol specifies that set_max_size and set_min_size must raise an invalid_size error if the resulting maximum size is smaller than the minimum size in either dimension. These constraints are double-buffered (applied on commit), so the check is performed at commit time using the effective (pending this cycle, otherwise committed) min and max sizes.

A zero max dimension means "no limit" (stored internally as the maximum possible size) and a zero min means "no minimum", so the comparison only triggers for genuinely conflicting client-set constraints. An equal min and max is permitted (fixed size window).


RAOF (approved):
Since this only applies to `XdgToplevelStable` this seems like something that would be better handled completely at the `XdgToplevelStable` level rather than at the generic `WindowWlSurfaceRole`, but that's a largely pre-existing problem.

This is ok.

PR #5010, opened 2026-06-09
frontend_wayland: raise invalid_fd when SHM pool fd cannot be mapped

robert-ancell:
wl_shm.create_pool mmap()s the client-provided file descriptor to back the pool. When that mapping fails the wl_shm spec requires the invalid_fd error (value 2, "mmapping the file descriptor failed") to be raised on the wl_shm object.

Previously the mmap failure surfaced as a std::system_error which the generated request dispatcher caught in its catch-all branch and reported as a generic wl_display implementation error rather than the specific wl_shm.invalid_fd error.

Introduce a dedicated mir::shm::MmapError exception thrown at the mmap failure site and catch it in create_pool to re-raise it as the correct protocol error. Using a specific exception type (rather than the broad std::system_error) ensures only genuine mapping failures are translated to invalid_fd, while other failures continue to be reported as internal errors.

### DIFF
--- a/src/server/shm_backing.cpp
+++ b/src/server/shm_backing.cpp
@@ -414,7 +414,7 @@ auto ShmBacking::resize(size_t new_size) -> std::expected<void, mir::shm::Resize
     void* mapped_address = mmap(nullptr, new_size, prot, MAP_SHARED, backing_store, 0);
     if (mapped_address == MAP_FAILED)
     {
-        BOOST_THROW_EXCEPTION((std::system_error{
+        BOOST_THROW_EXCEPTION((mir::shm::MmapError{
             errno,
             std::system_category(),
             "Failed to map client-provided SHM pool"}));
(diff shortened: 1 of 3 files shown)


src/server/shm_backing.cpp
```diff
@@ -414,7 +414,7 @@ auto ShmBacking::resize(size_t new_size) -> std::expected<void, mir::shm::Resize
     void* mapped_address = mmap(nullptr, new_size, prot, MAP_SHARED, backing_store, 0);
     if (mapped_address == MAP_FAILED)
     {
-        BOOST_THROW_EXCEPTION((std::system_error{
+        BOOST_THROW_EXCEPTION((mir::shm::MmapError{
```
robert-ancell, line 417: I was also considering if this should throw `wayland::ProtocolError` here, but the code feels too far from the frontend to know about this exception.

robert-ancell:
main is currently failing due to this test (ShmTest.create_pool_with_invalid_fd_is_an_error) being added to WLCS and Mir not being updated.

RAOF (approved):
I agree with the separation of `shm::MmapError` and the `ProtocolError`; this looks good to me.

PR #5013, opened 2026-06-09
Implement destroy_and_delete on C++ interfaces generated by wayland_rs

mattkae:
## What's new?
Implemented `destroy_and_delete` on the generated C++ interface. This method calls back into Rust where the handle is manually destroyed

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
+++ b/src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
@@ -144,10 +144,12 @@ fn generate_extension_for_interface(interface: &WaylandInterface) -> Option<Toke
 
             pub fn destroy_and_delete(&self) {
                 use wayland_server::Resource;
-                if let Some(handle) = self.wrapped.handle().upgrade() {
-                    let _ = handle.destroy_object::<crate::wayland_server_core::ServerState>(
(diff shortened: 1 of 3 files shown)


src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -142,6 +142,15 @@ fn generate_extension_for_interface(interface: &WaylandInterface) -> Option<Toke
                 self.wrapped.version()
             }
 
+            pub fn destroy_and_delete(&self) {
```
RAOF, line 145: *Hmmmmm*.

I don't suppose this could take `self: Self` (ie: by value?). That would make it also self-destroying.

RAOF (approved):
If `destroy_and_delete` can't easily take `self` by value, land as is (I didn't check the rest of the middleware for compatibility). 

Otherwise, can we make `destroy_and_delete` take `self` by value?

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
mattkae: This gets called through the FFI barrier, so it can't be a non reference type unfortunately.

PR #5014, opened 2026-06-10
frontend_wayland: add wl_fixes support

robert-ancell:
Implement the wl_fixes core protocol global and its destroy_registry request, which allows clients to cleanly destroy a wl_registry object.

The wl_fixes interface is not present in libwayland's auto-created globals, so the compositor must advertise and implement it. Add the interface definition to wayland.xml so the wrapper generator emits the mir::wayland::Fixes wrapper, implement WlFixes/FixesInstance in frontend_wayland, and wire the global into WaylandConnector.

destroy_registry is the first request to take a wl_registry argument. The generator skips wrappers for wl_display/wl_registry but still references their <name>_interface_data symbols. Teach the generator to emit definitions for these libwayland core interfaces (aliasing the wl_interface objects libwayland provides) in the wrapper of whichever protocol file contains them, so the symbols are defined exactly once.


wayland-protocols/wayland.xml
AlanGriffiths, line 1: Looks like there's more changes to `wayland.xml` than this: `diff /usr/share/wayland/wayland.xml wayland-protocols/wayland.xml` we should track upstream better

robert-ancell: I'll do a follow up PR to bring these changes in.

src/wayland/generator/wrapper_generator.cpp
```diff
@@ -130,7 +140,25 @@ Emitter source_file(std::string input_file_path, std::vector<Interface> const& i
...
+        // objects are defined (aliased from libwayland) below rather than forward-declared.
+        if (core_interfaces_in_file.find(interface_name) == core_interfaces_in_file.end())
+            fwd_declare_interface_emitters.push_back({"extern struct wl_interface const ", interface_name, "_interface_data;"});
+    }
+
+    // Define interface-data objects for the libwayland core interfaces (wl_display, wl_registry)
+    // present in this protocol. The generator does not wrap these interfaces, so it aliases the
+    // wl_interface objects libwayland already provides. A value copy is sufficient because
+    // libwayland's argument type checking compares interfaces by name. The extern declaration
+    // gives the definition external linkage (a namespace-scope const otherwise has internal
```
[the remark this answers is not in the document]
robert-ancell: Refactored

robert-ancell:
Now includes fixes for the Rust generation.

robert-ancell:
TICS failure is WLCS test failing since we haven't landed https://github.com/canonical/mir/pull/5010 yet.

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -345,6 +380,17 @@ fn generate_request_handler_arms(
...
+            // Requests referencing a core interface (wl_display/wl_registry) object are
+            // handled entirely on the Rust side rather than forwarded to C++.
+            if request_references_core_object(request) {
+                let body = generate_core_object_request_body(request);
+                return Some(quote! {
+                    #namespace_name::#interface_name::Request::#request_name { #( #arg_names ),* } => {
+                        #body
+                    }
+                });
+            }
```
mattkae: Is this true across the board? I feel like this could lead to some weird unexpected behavior.

And why can't C++ handle it? We have the ability for C++ to destroy the item. I'd rather keep all of the implementation in C++ if we can :)

mattkae: Also - is there somewhere in the Wayland protocol that says that this is how it should be done?

mattkae (changes requested):
I am skeptical of the wayland-rs changes, cause I think it might lead to some confusion later on down the road potentially. I would rather let C++ reason about it so that the implementation is at least explicit.

Let me know if you disagree though, I am only 50% convinced of my stance :)

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
robert-ancell: Struggling with what is correct Rust behaviour but I have taken a second attempt at this - please review.

src/server/frontend_wayland/wayland_rs/build_script/helpers.rs
```diff
@@ -1,9 +1,74 @@
...
+                        if let Some(name) = arg.interface.as_deref() {
+                            if !names.iter().any(|existing| existing == name) {
+                                names.push(name.to_string());
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
```
mattkae, line 68: This may be better written with `.map`s and `.filter`s and what not, no? Rust iters are our friend

src/server/frontend_wayland/wayland_rs/build_script/helpers.rs
```diff
@@ -1,9 +1,74 @@
...
+    let mut names: Vec<String> = Vec::new();
+    for protocol in protocols {
+        for interface in &protocol.interfaces {
+            for item in &interface.items {
+                if let InterfaceItem::Request(request) = item {
+                    for arg in request_core_object_args(request) {
+                        if let Some(name) = arg.interface.as_deref() {
+                            if !names.iter().any(|existing| existing == name) {
+                                names.push(name.to_string());
+                            }
```
mattkae, line 62: A `HashSet` might be a better data structure to use here

src/server/frontend_wayland/wayland_rs/build_script/helpers.rs
```diff
@@ -1,9 +1,74 @@
...
+}
+
+/// Collect the names of the libwayland core interfaces that are referenced as `object`
+/// arguments by some request across all protocols (e.g. `wl_registry`).
+///
+/// These core interfaces have no generated C++ class, but the requests that reference them
+/// (such as `wl_fixes.destroy_registry`) are still forwarded to C++. To make that possible a
+/// minimal Rust middleware is generated for each of them, exposing just enough to let the C++
+/// implementation act on the object (currently `destroy_and_delete`).
+pub fn referenced_core_interfaces(protocols: &[WaylandProtocol]) -> Vec<String> {
```
mattkae, line 52: Let me know if I understand this correctly, cause it is complicated :)

We iterate all protocols, interfaces, and their requests to find arguments where the object interfaces being referenced is a `wl_display` or `wl_registry`. Afterward we add the argument name to the list... which means that the list always contains `["wl_display", "wl_registry"]`. Do I misunderstand that?

src/server/frontend_wayland/wayland_rs/build_script/ffi_generation.rs
```diff
@@ -27,6 +29,22 @@ pub fn generate_ffi(protocols: &Vec<WaylandProtocol>, builders: &Vec<CppBuilder>
...
+        .iter()
+        .map(|name| {
+            let ext = format_ident!(
+                "{}",
+                format_wayland_interface_to_rust_extension_struct(name)
+            );
+            quote! {
+                type #ext;
+                fn destroy_and_delete(self: &#ext);
+            }
```
mattkae, line 44: So my question is: when will C++ get access to the Ext class? Right now, they get access through their corresponding interface implementation in C++ (the C++ side wraps the Ext essentially). What is going on here?

mattkae (changes requested):
I have some questions, maybe some misunderstandings

AlanGriffiths:
@robert-ancell there are unaddressed comments from months ago. And merge conflicts.

I'm closing this. Feel free to reopen if the above are fixed.

PR #5015, opened 2026-06-10
all: rewrite equality operators, bump ABI and add `package` workflow action

Saviq:
## What's new?

- rewrite equality operators with `default`, `std::tie`, `std::make_tuple`, as appropriate
- bump all ABI as a result
- add a `package` workshop action

## How to test

CI

## Checklist

- [ ] Tests added and pass


src/include/server/mir/input/touch_visualizer.h
```diff
@@ -39,6 +39,9 @@ class TouchVisualizer
 
         // If pressure is zero, the touch-point can be interpreted as a hover.
         float pressure;
+
+        bool operator==(Spot const&) const = default;
+        bool operator!=(Spot const&) const = default;
```
AlanGriffiths: Elsewhere we've elided the secondary `!=` operator.

AlanGriffiths (approved):
There are a few cases where the equality implementation looks odd (but matches the existing code). Treating those as "pre-existing at worst".

E.g. `MirInputDevice` has a `unique_id` which is ignored by the comparison (with no explanation)

We also use two ways of comparing `std::weak_ptr<>` for equality: using `owner_before()` and using `lock()`. The former avoid unnecessary synchronization, but they are not quite equivalent (for expired pointers to different objects). Somehow I doubt the choice was informed.

src/include/server/mir/input/touch_visualizer.h
Saviq: Indeed, thanks for spotting!
```suggestion
```

src/miral/symbols.map
```diff
@@ -449,366 +753,20 @@ global:
...
+    typeinfo?for?miral::live_config::Store;
+    typeinfo?for?std::formatter?miral::live_config::Key*char?;
     vtable?for?miral::ApplicationAuthorizer;
     vtable?for?miral::CanonicalWindowManagerPolicy;
     vtable?for?miral::FdHandle;
+    vtable?for?miral::FloatingWindowManager;
     vtable?for?miral::MinimalWindowManager;
     vtable?for?miral::WindowManagementPolicy;
-  };
-local: *;
```
jhenstridge, line 458: You probably still want this line somewhere in the symbol map to hide unmatched symbols. This is probably what is responsible for the unversioned symbols.

PR #5016, opened 2026-06-10
Rework comparison operators

AlanGriffiths:
Updates comparison operators to C++20


Saviq:
This, and more, is happening in #5015 instead.

PR #5020, opened 2026-06-13
wayland platform: two fixes for running Mir nested under another compositor

Azkali:
Two small, independent fixes to the Wayland display platform needed to run
a Mir server nested as a Wayland client of another compositor.

**1. Fall back to the output mode size on a zero-size `toplevel_configure`**
xdg-shell allows the host to send `xdg_toplevel.configure` with width/height
`0` ("client picks its own size"). `DisplayClient::Output::toplevel_configure`
only set `pending_toplevel_size` for a non-zero size, so on a zero-size
configure `surface_configure`'s guard was skipped, `has_initialized` stayed
`false`, `on_display_config_changed()` never fired, and the `DisplayClient`
constructor's roundtrip loop waited forever. Falls back to the `wl_output`
current mode size when the host offers no preferred size.

**2. Block signals on the display event loop**
`Display::run()` treated any `poll()` returning `-1` as fatal. `poll()` also
returns `-1`/`EINTR` on signal delivery (e.g. `SIGCHLD` when miral spawns
Xwayland), so a routine signal aborted the platform.

Block signals on the display event loop thread as we do on other eventloop threads
to avoid this.


src/platforms/wayland/displayclient.cpp
```diff
@@ -290,6 +290,10 @@ void mgw::DisplayClient::Output::toplevel_configure(int32_t width, int32_t heigh
     {
         pending_toplevel_size = geometry::Size{host_scale*width, host_scale*height};
     }
+    else if (!dcout.modes.empty())
+    {
+        pending_toplevel_size = dcout.modes[dcout.current_mode_index].size;
+    }
```
RAOF, line 296: This is good.

src/platforms/wayland/display.cpp
```diff
@@ -239,7 +239,11 @@ try
             fds[display_fd].events &= ~POLLOUT;
         }
 
-        if (poll(fds, indices, -1) == -1)
+        int poll_rc;
+        while ((poll_rc = poll(fds, indices, -1)) == -1 && errno == EINTR)
+        {
+        }
+        if (poll_rc == -1)
         {
```
RAOF, line 246: Here I think we should instead use a `mir::SignalBlocker` at the top of the function to prevent signals being sent on this thread. That's what we do elsewhere to avoid interruptions from signals.

RAOF (changes requested):

src/platforms/wayland/display.cpp
Azkali: Noted !

Azkali: Done

RAOF (approved):
Huh. Did not know we didn't already export that in the headers!

I was considering whether `signal_blocker.h` should be in `mir/dispatch`, but I don't think that's necessary; it's *used* by our dispatch, but it's not intrinsically dispatch.

LGTM!

Edit: Oh. I'll update the merge commit message and then hit the button.

PR #5021, opened 2026-06-15
deprecated APIs die

AlanGriffiths:
Continue the process of dropping deprecated APIs


tarek-y-ismail (approved):
Less code FTW

mattkae (approved):
Why not if we can remove it!

Issue #5022, opened 2026-06-15
Someone please kill msh::SystemCompositorWindowManager ;)


AlanGriffiths:
`BasicWindowManager::add_display()` has been deprecated since (at least) 2017.

The only part of the class hierarchy with a meaningful implementation is `msh::SystemCompositorWindowManager::add_display()` which is only used by Mir's "default" `DefaultServerConfiguration::the_window_manager_builder()` - which only used by some tests.

If it were not for the test dependencies, we could clean this up

PR #5023, opened 2026-06-15
Drop obsolete `WindowManager::{add,remove}_display()` methods

AlanGriffiths:
These methods don't do anything useful.

The only implementation is in `msh::SystemCompositorWindowManager` which is only used as a default by test fixtures, and the tests don't use these functions. (Also see https://github.com/canonical/mir/issues/5022)


RAOF (approved):
Yup!

PR #5026, opened 2026-06-16
ShmBacking: Switch from `Synchronised` to `std::atomic` for `current_mapping`

RAOF:
We don't actually need locking around `current_mapping`, we only need to be able to view a consistent state, and since the target of the `shared_ptr` is `const` we only need to ensure that the `shared_ptr` itself is consistent. Now that 20.04 is in the past, we can do that with `std::atomic<shared_ptr<>>`!


mattkae (approved):
Sure!

PR #5027, opened 2026-06-16
Update filter for renamed WLCS test

robert-ancell:
See https://github.com/canonical/wlcs/pull/442


RAOF (approved):
Yup, LGTM.

Saviq:
We'll need a WLCS release before CI likes this. Should we keep both names until then?

AlanGriffiths:
> We'll need a WLCS release before CI likes this. Should we keep both names until then?

Yes

AlanGriffiths (changes requested):
We need to work also with the released wlcs

robert-ancell:
Updated to support old name

AlanGriffiths (approved):

PR #5030, opened 2026-06-16
Implement an Executor for wayland-rs that queues arbitrary C++ callbacks on the Rust thread

mattkae:
## What's new?
Because our Wayland implementation is single-threaded, we need to be able to queue arbitrary work on the thread. This aligns us with the current `frontend_wayland` implementation thus easing the refactoring requirements for later on.

Because we cannot send functions over the boundary, we generate an ID for a function in C++, then we tell Rust to tell us when to execute the function that is associated with that ID.

## How to test
1. Run the `mirwayland_rs-demo`
2. See the new logs get printed out

## Checklist

- [x] Tests added and pass


src/server/frontend_wayland/wayland_rs/build_script/cpp_protocol_generation.rs
```diff
@@ -157,7 +160,28 @@ fn create_wayland_server_notification_handler() -> CppBuilder {
...
+/// `execute(work_id)` on the event-loop thread to run it.
+fn create_work_callback() -> CppBuilder {
+    let mut builder: CppBuilder = CppBuilder::new("MIR_WAYLANDRS_WORK_CALLBACK", "work_callback");
+    builder.add_header_include("<cstdint>");
+
+    builder.add_cpp_include("\"wayland_rs/src/ffi.rs.h\"");
+    let mut namespace = CppNamespace::new(vec!["mir", "wayland_rs"]);
+    let mut class = CppClass::new("WorkCallback");
+
+    let mut execute_method = CppMethod::new("execute", None, true, false, true, true);
```
jhenstridge: Not relevant to the changes in this PR, but four consecutive boolean arguments seem a bit error prone. This might benefit from a refactor later on.

RAOF: `("execute", None, true, false, true, true)` is getting a bit arcane. Maybe we should refactor this into either enum flags (eg: `type::Virtual`, `type::NonVirtual`) or maybe even using the builder pattern
```
CppMethod::new("execute")
    .is_virtual(true)
    .throws(false)…
```

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/src/wayland_executor.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+}
+
+void mrs::WaylandExecutor::spawn(std::function<void()>&& work)
+{
+    std::uint32_t work_id;
+    {
+        std::lock_guard<std::mutex> lock{mutex};
+        work_id = next_id++;
+        pending_work.emplace(work_id, std::move(work));
```
jhenstridge: I know it is incredibly likely, but it might be worth checking if work_id is already in pending_work. Looking at the spec, `work` won't be added to the map if there is a collision.

It would be worth checking the result to ensure it was added.

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/src/wayland_executor.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    server.schedule_work(work_id);
+}
+
+auto mrs::WaylandExecutor::execute(std::uint32_t work_id) -> void
+{
+    std::function<void()> work;
+    {
+        std::lock_guard<std::mutex> lock{mutex};
+        auto const it = pending_work.find(work_id);
```
jhenstridge: Might `pending_work.extract(work_id)` be more concise here?

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/src/wayland_executor.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "wayland_rs/src/ffi.rs.h"
+
+namespace mrs = mir::wayland_rs;
+
+mrs::WaylandExecutor::WaylandExecutor(mrs::WaylandServer& server)
+    : server{server}
+{
+}
+
+void mrs::WaylandExecutor::spawn(std::function<void()>&& work)
```
jhenstridge, line 28: While talking with Chris, maybe we could get rid of the the work_id abstraction. We just need to tell the Rust side that there are C++ tasks to be executed, and when the Rust side schedules us we could just execute all the pending work.

So this could just create a vector/list of work items, and call the rust schedule_work() if the list was previously empty.

`execute` could then drain the list, executing each task.

src/server/frontend_wayland/wayland_rs/src/wayland_server_core.rs
```diff
@@ -131,14 +142,32 @@ impl WaylandServer {
...
+            .insert_source(task_channel, |event, _, state: &mut ServerState| {
+                if let ChannelEvent::Msg(task) = event {
+                    task(state);
+                }
+            })
+            .map_err(|_| "Failed to insert task queue into event loop")?;
+
         WaylandServer::register_globals(&state, Arc::new(Mutex::new(factory)));
 
         // TODO: Don't spin continuously
```
RAOF: Existing, but this TODO is long dead, right? `event_loop.dispatch()` does a sensible wait for events?

tests/wayland_rs/test_wayland_executor.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    char template_path[] = "/tmp/mir-wayland-rs-test-XXXXXX";
+    auto const dir = ::mkdtemp(template_path);
+    if (!dir)
+        throw std::runtime_error{"Failed to create temporary XDG_RUNTIME_DIR"};
+
+    ::chmod(dir, 0700);
+    ::setenv("XDG_RUNTIME_DIR", dir, 1);
+    return dir;
+}
```
RAOF: Does our test environment not have a working `XDG_RUNTIME_DIR` environment? Do we need to create (and then not clean up) a temporary directory here?

jhenstridge: Might it be worth using a temporary XDG_RUNTIME_DIR for each test and clean it up on TearDown? This looks like it will sometimes leave garbage behind.

jhenstridge (changes requested):
This looks pretty good. Chris had an idea about dropping the work_id concept which seems worth pursuing.

tests/wayland_rs/test_wayland_executor.cpp
```diff
@@ -0,0 +1,207 @@
...
+            (*server)->stop();
+        if (server_thread.joinable())
+            server_thread.join();
+    }
+
+    /// Spawn probe work repeatedly until one is actually executed, indicating
+    /// the server's event loop is running and its work queue is wired up.
+    ///
+    /// Work spawned before the queue exists is silently dropped, so this retries
+    /// rather than relying on a single probe.
```
RAOF: It's actually not necessary to have this restriction if you let the C++ side handle dispatch ordering; you can just queue on the C++ side and drain the queue on server start.

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/src/wayland_executor.cpp
mattkae: True! That simplifies everything so much

src/server/frontend_wayland/wayland_rs/build_script/cpp_protocol_generation.rs
mattkae: Let's do this in a follow-up, since this is pretty large right now :)

tests/wayland_rs/test_wayland_executor.cpp
mattkae: We should always point it at a tmp one and then clean that up. This is better for local runs IMO

mattkae: Yeah thats what I did

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/include/wayland_executor.h
```diff
@@ -0,0 +1,68 @@
...
+{
+public:
+    explicit WaylandExecutor(WaylandServer& server);
+
+    /// Schedule work to be run on the Wayland event loop thread.
+    void spawn(std::function<void()>&& work) override;
+
+    /// Run all currently queued work. Called from Rust on the event-loop
+    /// thread.
+    auto execute() -> void override;
```
jhenstridge:
```suggestion
    void execute() override;
```

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/src/wayland_executor.cpp
```diff
@@ -0,0 +1,68 @@
...
+    bool need_signal;
+    {
+        std::lock_guard<std::mutex> lock{mutex};
+        pending_work.push_back(std::move(work));
+
+        // Only raise a signal if one is not already outstanding; any work
+        // appended while a signal is pending (or a drain is in progress) is
+        // picked up by that drain, so one signal per batch suffices.
+        need_signal = !signal_pending;
+        if (need_signal)
```
jhenstridge: You can drop the conditional here.

src/server/frontend_wayland/wayland_rs/src/wayland_server_core.rs
jhenstridge: Small nit: as you're creating two pingers in the function now, it might be worth naming these locals to make it clear they related to the stop signal.

src/server/frontend_wayland/wayland_rs/wayland_rs_cpp/src/wayland_executor.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        std::lock_guard<std::mutex> lock{mutex};
+        work.swap(pending_work);
+        signal_pending = false;
+    }
+
+    // Run the work outside the lock so that it may itself schedule more work.
+    for (auto& task : work)
+    {
+        if (task)
```
jhenstridge: In what situations would we have unset functions in the work queue? If this is something we need to support, wouldn't it be better to catch them in `spawn` and never add them to the queue?

jhenstridge (approved):
Looks good. I've left a few small suggestions you could take or leave, but overall it looks like a good approach.

tests/wayland_rs/test_wayland_executor.cpp
```diff
@@ -0,0 +1,198 @@
...
+public:
+    void SetUp() override
+    {
+        // Always run against a private temporary XDG_RUNTIME_DIR so the test
+        // never pollutes (or collides with) a real runtime directory. It is
+        // removed again in TearDown.
+        runtime_dir = make_temp_runtime_dir();
+        if (auto const existing = ::getenv("XDG_RUNTIME_DIR"))
+            previous_xdg_runtime_dir = existing;
+        ::setenv("XDG_RUNTIME_DIR", runtime_dir.c_str(), 1);
```
jhenstridge: Would `mir_test_framework::TemporaryEnvironmentValue` simplify this code a bit?

PR #5028, opened 2026-06-16
all: migrate to C++26 along with cmake and some legacy code

Saviq:
## What's new?

C++26 is!

(NB: this is on top of #4738)

## How to test

CI

## Checklist

- [ ] Tests added and pass
- [x] Adequate documentation added
- [ ] (optional) Added Screenshots or videos

PR #5031, opened 2026-06-17
Help the compiler figure out const-ness

RAOF:


RAOF:
Just to check if this fixes the CI build for #5028

Saviq:
> Just to check if this fixes the CI build

It doesn't…

PR #5036, opened 2026-06-18
docs: update documentation for configuring screencasting via ext-image-copy-capture

jhenstridge:
## What's new?

Update  the screencasting documentation to instead cover the new ext-image-copy-capture protocol. In particular:

* recommends the ext protocols, with a base set to support output capture, and additional protocols to enable toplevel window capture.
* Bumps the minimum required xdg-desktop-portal-wlr version requirement to 0.8.0, with a note that 0.8.2 is needed for toplevel capture to (mostly) work.
* Document the new way of configuring which portals to use via a `<compositor-name>-portals.conf` file. The suggestion given is to use wlr for screenshots/screencasts and default to gtk for everything else.
* Suggest using wmenu instead of slurp as xdg-desktop-portal-wlr's source chooser, since slurp can't be used to select toplevels.
* Add a note about the bug in xdg-desktop-portal-wlr when resizing a toplevel that is being captured.
* Note that integration with systemd user sessions will be needed to function correctly with xdg-desktop-portal >= 1.21 due to the `After=graphical-session.target` line in its service file.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/doc/sphinx/.custom_wordlist.txt
+++ b/doc/sphinx/.custom_wordlist.txt
@@ -233,6 +233,7 @@ subprocess
 swaybg
 swaylock
 sys
+systemd
 testflinger
 Testflinger
 TilingWindowManagerPolicy
@@ -278,6 +279,7 @@ WL
 wlcs
 WLCS
 wlr
+wmenu
 wofi
 workspaces
 X11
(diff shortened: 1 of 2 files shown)


AlanGriffiths (approved):
Looks sensible.

What I'd like to do validate this is to update the miriway snap to use xdg-desktop-portal-wlr:0.8.2 (and maybe patch it too). (It currently pulls v0.5.0 and uses appropriate config.)

Not got time before EOD, so approving but not landing

mattkae (approved):

PR #5040, opened 2026-06-22
Add a method to listen for fd read ready on the Wayland thread in wayland-rs

mattkae:
## What's new?
- Add a new method in Rust called `WaylandServer::register_fd_ready_listener` which takes a file descriptor and and object representing a callback. This method will trigger the callback when the fd becomes readable.
- C++ implementations implement the `wayland_rs::FdReadyCallback` interface to hear about when an fd is ready
- Added usage to the example server
- Added unit tests

## How to test
1. Run the example server and see the logs
2. Run the unit tests and see that they pass

## Checklist

- [x] Tests added and pass


robert-ancell (approved):
Sure

src/server/frontend_wayland/wayland_rs/src/wayland_server_core.rs
```diff
@@ -14,25 +14,34 @@
...
+/// A registration of interest in a file descriptor becoming readable, awaiting
+/// insertion into the running event loop.
+type PendingFdListener = (RawFd, Box<dyn FdReadyListener>);
+
 /// The wayland server.
 pub struct WaylandServer {
     stop_signal: Mutex<Option<Ping>>,
     work_signal: Mutex<Option<Ping>>,
+    fd_listener_signal: Mutex<Option<Ping>>,
+    pending_fd_listeners: Arc<Mutex<Vec<PendingFdListener>>>,
```
RAOF: If you stored a reference to the mutex in `ServerState` you could drop the `Arc` here, I think.

tests/wayland_rs/test_fd_ready_listener.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    int pipe_fds[2]{-1, -1};
+};
+}
+
+TEST_F(FdReadyListenerTest, notifies_when_fd_becomes_readable)
+{
+    std::mutex mutex;
+    std::condition_variable cv;
+    int count{0};
```
RAOF: This is a `mir::test::Signal`, right?

Alternatively, this could also be `std::atomic<int>` with `.wait()` and `.notify_all()`.

RAOF (approved):
A couple of minor improvements available; feel free to apply and land.

PR #5042, opened 2026-06-22
Generated C++ implementations in wayland-rs take a std::optional instead of a value + has_value field in the implementor-facing API

mattkae:
## What's new?
- The `Middleware` object for each C++ method now handles the conversion from `(value, has_value)` to `std::optional<>(value)`.
- Going from C++ -> Rust (because of a Wayland event) now provides objects as pointers when they can be null. Afterward, they are manually unwrapped.

## Checklist

- [ ] Tests added and pass


AlanGriffiths (approved):
Looks plausible, but would be nice to see some code that uses it

mattkae:
> Looks plausible, but would be nice to see some code that uses it

That will come very soon!

PR #5043, opened 2026-06-23
Introduce mir-rs-sys and mir-rs and hook up unittests to CI

mattkae:
## What's new?
- Introduces the `mir-rs` and `mir-rs-sys` subprojects with skeleton builds
- Hooks up their unittests to CI with the help of a new CMake helper

## Checklist

- [x] Tests added and pass


Saviq (commented):
Should we nest these under a `crates`? Or `rs`?

mattkae:
> Should we nest these under a `crates`? Or `rs`?

There is prior art for a `crates` directory. I wouldn't be opposed if you aren't. I think `rs` would be more confusing

crates/mir-rs-sys/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+repository = "https://github.com/canonical/mir"
+readme = "README.md"
+keywords = ["mir", "wayland", "compositor", "ffi", "bindings"]
+categories = ["external-ffi-bindings", "gui"]
+links = "miral"
+edition.workspace = true
+rust-version.workspace = true
+
+[lib]
+crate-type = ["staticlib", "rlib"]
```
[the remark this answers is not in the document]
RAOF: Yeah, why are you building a `staticlib` here?

crates/mir-rs-sys/build.rs
```diff
@@ -0,0 +1,46 @@
...
+            \n",
+        );
+
+    // Link against miral and its dependencies
+    for lib in &miral.libs {
+        println!("cargo:rustc-link-lib={}", lib);
+    }
+    for path in &miral.link_paths {
+        println!("cargo:rustc-link-search=native={}", path.display());
+    }
```
RAOF: It might be better to use [`system-deps`](https://crates.io/crates/system-deps) and replace this with a declarative section in the `Cargo.toml` plus `system_deps::Config::new().probe().expect(…);` here.

crates/mir-rs-sys/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "mir-sys"
+version = "0.1.0"
+description = "Low-level FFI bindings to the Mir compositor's miral C++ library"
+license = "GPL-2.0-or-later"
```
RAOF: This should actually be

```suggestion
license = "GPL-2.0-only OR GPL-3.0-only"
```

crates/mir-rs/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "mir"
+description = "Idiomatic Rust API for building Wayland compositors with the Mir display server"
+license = "GPL-2.0-or-later"
```
RAOF:
```suggestion
license = "GPL-2.0-only OR GPL-3.0-only"
```

cmake/RustTest.cmake
```diff
@@ -0,0 +1,72 @@
...
+  # Where the freshly built shared libraries (libmiral, libmircore, ...) live.
+  set(lib_dir "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+
+  mir_add_test(NAME ${ARG_NAME}
+    COMMAND ${CMAKE_COMMAND} -E env
+      "PKG_CONFIG_PATH=${pkg_config_path}"
+      "RUSTFLAGS=-L native=${lib_dir}"
+      "LD_LIBRARY_PATH=${lib_dir}:$ENV{LD_LIBRARY_PATH}"
+      ${CARGO_EXECUTABLE} test --target-dir ${CMAKE_BINARY_DIR}/target ${cargo_release_flag} ${package_args}
+    WORKING_DIRECTORY ${PROJECT_SOURCE_DIR})
```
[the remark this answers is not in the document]
RAOF: This is true (but we don't care about MacOS). We could probably simplify this by also generating some `$FOO-uninstalled.pc` files, with the necessary `-L`/`-I` lines for an in-tree build.

Not necessarily today, though.

RAOF (changes requested):

RAOF (approved):
Yeah, LGTM.

PR #5046, opened 2026-06-24
Update WaylandClient API in wayland-rs with methods required for backward compatibility

mattkae:
## What's new?
- Updated the `WaylandClient` interface with the methods that we'll need later on

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/src/wayland_client.rs
+++ b/src/server/frontend_wayland/wayland_rs/src/wayland_client.rs
@@ -1,3 +1,6 @@
+use crate::wayland_server_core::ClientState;
+use cxx::CxxString;
+use wayland_backend::protocol::ProtocolError;
 use wayland_server::{backend::ClientId, Client, DisplayHandle};
 
 /// A C++ friendly wrapper around a wayland [Client] object.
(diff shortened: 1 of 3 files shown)


tarek-y-ismail (commented):
Does this have a C++ equivalent? I can't quite find it. The code seems sensible, just wanna make sure we're not forgetting anything.

src/server/frontend_wayland/wayland_rs/src/wayland_client.rs
```diff
@@ -44,15 +47,41 @@ impl WaylandClient {
...
     }
+
+    /// Kill this client with a protocol error.
+    ///
+    /// This will disconnect the client, sending the protocol error as the reason.
+    pub fn kill(&self, object_id: u32, code: u32, message: &CxxString) {
+        let error = ProtocolError {
+            code,
+            object_id,
+            object_interface: String::new(),
```
tarek-y-ismail, line 84: Is this a temporary placeholder, or do we really not need the interface name in this context?

tarek-y-ismail (commented):
Ok. Turns out the C++ equivalent is roughly `wayland::Client` (duh). I assume this is set up for `WaylandServerNotificationHandler::client_added`?

mattkae:
> Ok. Turns out the C++ equivalent is roughly `wayland::Client` (duh). I assume this is set up for `WaylandServerNotificationHandler::client_added`?

Yes this makes it a 1-to-1 replacement, which will make the final migration _much_ easier to grep through :) This is mostly to replace `wl_resource*` casting

src/server/frontend_wayland/wayland_rs/src/wayland_client.rs
mattkae: I'll add a comment, but we really don't need it :+1:

tarek-y-ismail (approved):
Ok. Let's get it in.

PR #5048, opened 2026-06-24
Make `wayland_rs::Weak`'s semantics match that of the existing `wayland::Weak` more closely + generate a `from<Dervied>` on generated wayland-rs Impls for backwards compatibilty

mattkae:
The goal of this work was to make storing pointers of a Wayland interfaces between each other translate better from the old C++ implementation to the new Rust-based implementation. My previous attempt at this used `std::shared_from_this`, but that was ergonomically poor and did not translate well to the existing interface. This implementation should make things much more backwards-compatible and easy to work with.

## What's new?
- `wayland_rs::Weak` now wraps a raw pointer and checks `is_destroyed` instead of wrapping a `shared_ptr`. This brings it inline with the ergonomics of the existing `wayland::Weak`, making backwards compatibility much better.
- Generating `::from<T>` methods on each Wayland class that will return a raw pointer `T*` from the `Weak` implementation. This is useful for get a `Derived` from a `wayland_rs::Weak<Base>`
- Wrote unit tests

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
+++ b/src/server/frontend_wayland/wayland_rs/build_script/cpp_builder.rs
@@ -129,6 +129,15 @@ impl CppBuilder {
                 // Virtual destructor
                 result.push_str(&format!("    virtual ~{}() = default;\n\n", class.name));
 
+                // Verbatim header-only public declarations (e.g. static templated helpers).
+                for decl in &class.raw_public_decls {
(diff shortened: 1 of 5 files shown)


robert-ancell (approved):
I don't know the full context of how this is used but the code looked correct to me.

RAOF (approved):
Yup, looks good.

PR #5057, opened 2026-06-26
drop-miral-legacy

AlanGriffiths:
Drops a load of unused and/or deprecated stuff from libmiral


tarek-y-ismail (approved):

tarek-y-ismail:
@AlanGriffiths Heads up: I think you dropped the symbols changes accidentally. I also don't recall seeing Debian symbols changes.

PR #5070, opened 2026-07-07
Fix documentation errors in input trigger protocols

robert-ancell:
Correct wording mistakes in the ext_input_trigger_action_v1 and ext_input_trigger_registration_v1 protocol descriptions:

  - ext_input_trigger_action_manager_v1.destroy and ext_input_trigger_action_v1.destroy referred to the object as an "activation" / "activation manager"; the interfaces are the input trigger "action" and "action manager". (The activation_token references elsewhere correctly refer to xdg_activation.)

  - ext_input_trigger_action_control_v1.cancel is a client request but was described as an "event" that "may be sent"; it is now described as a request that "may be made".

### DIFF
--- a/wayland-protocols/ext-input-trigger-action-v1.xml
+++ b/wayland-protocols/ext-input-trigger-action-v1.xml
@@ -43,8 +43,8 @@
     </enum>
 
     <request name="destroy" type="destructor">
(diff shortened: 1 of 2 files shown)


jhenstridge (approved):
Looks good!

PR #3954, opened 2025-05-21
feature: implement magnification accessibility feature

mattkae:

PR #5071, opened 2026-07-07
Update magnifier for touch.

tarek-y-ismail:
Related: #3954

This PR is only for the visual part of the magnifier. It will be followed up by another PR to handle input remapping through the magnification surface.

## What's new?

- Updates the magnifier to add a new decoupled mode that does not follow the pointer. And adds touch and pointer compatible buttons to move, resize, and control zoom.
- The area being zoomed is interpolated to land on edges and corners as the magnification surface does. (Thanks @Saviq!)

## How to test

- Enable the magnifier via `Ctrl + =`.
- It should start in coupled mode by default
- Press `Ctrl + Alt + M` to decouple it
- Observe as the surface stops following the mouse, and becomes accompanied by a few buttons to control it
- Touch emulation can be tested by pulling #5091

### DIFF
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -223,9 +223,9 @@ int main(int argc, char const* argv[])
         {
             if (modifiers & mir_input_event_modifier_shift)
             {
-                auto const magnifier_min_magnification = 1.5f;
+                auto const magnifier_min_magnification = 1.25f;
                 auto const magnifier_max_magnification = 8.0f;
-                auto const magnifier_magnification_step = 0.5f;
+                auto const magnifier_magnification_step = 0.25f;
 
                 // Zoom the magnifier in/out on ctrl shift +/-
                 switch (mir_keyboard_event_keysym(key_event))
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -47,6 +47,7 @@
 
 #include <xkbcommon/xkbcommon-keysyms.h>
 
+#include <algorithm>
 #include <cstdlib>
 #include <cstring>
 #include <linux/input-event-codes.h>
@@ -208,7 +209,11 @@ int main(int argc, char const* argv[])
         .magnification(magnification)
         .capture_size(capture_size)
         .enable(false);
-    auto magnifier_filter = [magnifier=magnifier, &magnification, &capture_size](MirKeyboardEvent const* key_event) mutable {
+    auto magnifier_filter = [magnifier = magnifier,
+                             &magnification,
+                             &capture_size,
+                             follows_cursor = true](MirKeyboardEvent const* key_event) mutable
+    {
         auto const modifiers = mir_keyboard_event_modifiers(key_event);
 
         if (mir_keyboard_event_action(key_event) != mir_keyboard_action_down)
@@ -218,31 +223,29 @@ int main(int argc, char const* argv[])
         {
             if (modifiers & mir_input_event_modifier_shift)
             {
+                auto const magnifier_min_magnification = 1.25f;
+                auto const magnifier_max_magnification = 8.0f;
+                auto const magnifier_magnification_step = 0.25f;
+
                 // Zoom the magnifier in/out on ctrl shift +/-
                 switch (mir_keyboard_event_keysym(key_event))
                 {
-                    case XKB_KEY_plus:
-                        magnification += 0.5f;
-                        if (magnification >= 5)
-                        {
-                            magnification = 5;
-                            return true;
-                        }
-
-                        magnifier.magnification(magnification);
-                        return true;
-                    case XKB_KEY_underscore:
-                        magnification -= 0.5f;
-                        if (magnification <= 1)
-                        {
-                            magnification = 1;
-                            return true;
-                        }
-
-                        magnifier.magnification(magnification);
-                        return true;
-                    default:
-                        break;
+                case XKB_KEY_plus:
+                    magnification = std::clamp(
+                        magnification + magnifier_magnification_step,
+                        magnifier_min_magnification,
+                        magnifier_max_magnification);
+                    magnifier.magnification(magnification);
+                    return true;
+                case XKB_KEY_underscore:
+                    magnification = std::clamp(
+                        magnification - magnifier_magnification_step,
+                        magnifier_min_magnification,
+                        magnifier_max_magnification);
+                    magnifier.magnification(magnification);
+                    return true;
+                default:
+                    break;
                 }
             }
             else if (modifiers & mir_input_event_modifier_alt)
@@ -250,20 +253,27 @@ int main(int argc, char const* argv[])
                 // Grow area on ctrl alt +/-
                 switch (mir_keyboard_event_keysym(key_event))
                 {
-                    case XKB_KEY_equal:
-                    {
-                        capture_size.width = Width(std::min(1000, capture_size.width.as_int() + 100));
-                        capture_size.height = Height(std::min(1000, capture_size.height.as_int() + 100));
-                        magnifier.capture_size(capture_size);
-                        return true;
-                    }
-                    case XKB_KEY_minus:
-                        capture_size.width = Width(std::max(100, capture_size.width.as_int() - 100));
-                        capture_size.height = Height(std::max(100, capture_size.height.as_int() - 100));
-                        magnifier.capture_size(capture_size);
-                        return true;
-                    default:
-                        break;
+                case XKB_KEY_equal:
+                {
+                    capture_size.width = Width(std::min(1000, capture_size.width.as_int() + 100));
+                    capture_size.height = Height(std::min(1000, capture_size.height.as_int() + 100));
+                    magnifier.capture_size(capture_size);
+                    return true;
+                }
+                case XKB_KEY_minus:
+                    capture_size.width = Width(std::max(100, capture_size.width.as_int() - 100));
+                    capture_size.height = Height(std::max(100, capture_size.height.as_int() - 100));
+                    magnifier.capture_size(capture_size);
+                    return true;
+                case XKB_KEY_m:
+                    follows_cursor = !follows_cursor;
+                    if (follows_cursor)
+                        magnifier.set_behavior(miral::Magnifier::Behavior::follow_cursor);
+                    else
+                        magnifier.set_behavior(miral::Magnifier::Behavior::freely_positioned);
+                    return true;
+                default:
+                    break;
                 }
             }
             else
@@ -271,14 +281,14 @@ int main(int argc, char const* argv[])
                 // Turn on/off the magnifier on ctrl +/-
                 switch (mir_keyboard_event_keysym(key_event))
                 {
-                    case XKB_KEY_equal:
-                        magnifier.enable(true);
-                        return true;
-                    case XKB_KEY_minus:
-                        magnifier.enable(false);
-                        return true;
-                    default:
-                        break;
+                case XKB_KEY_equal:
+                    magnifier.enable(true);
+                    return true;
+                case XKB_KEY_minus:
+                    magnifier.enable(false);
+                    return true;
+                default:
+                    break;
                 }
             }
         }
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -209,7 +209,11 @@ int main(int argc, char const* argv[])
         .magnification(magnification)
         .capture_size(capture_size)
         .enable(false);
-    auto magnifier_filter = [magnifier=magnifier, &magnification, &capture_size](MirKeyboardEvent const* key_event) mutable {
+    auto magnifier_filter = [magnifier = magnifier,
+                             &magnification,
+                             &capture_size,
+                             decoupled = false](MirKeyboardEvent const* key_event) mutable
+    {
         auto const modifiers = mir_keyboard_event_modifiers(key_event);
 
         if (mir_keyboard_event_action(key_event) != mir_keyboard_action_down)
@@ -226,22 +230,22 @@ int main(int argc, char const* argv[])
                 // Zoom the magnifier in/out on ctrl shift +/-
                 switch (mir_keyboard_event_keysym(key_event))
                 {
-                    case XKB_KEY_plus:
-                        magnification = std::clamp(
-                            magnification + magnifier_magnification_step,
-                            magnifier_min_magnification,
-                            magnifier_max_magnification);
-                        magnifier.magnification(magnification);
-                        return true;
-                    case XKB_KEY_underscore:
-                        magnification = std::clamp(
-                            magnification - magnifier_magnification_step,
-                            magnifier_min_magnification,
-                            magnifier_max_magnification);
-                        magnifier.magnification(magnification);
-                        return true;
-                    default:
-                        break;
+                case XKB_KEY_plus:
+                    magnification = std::clamp(
+                        magnification + magnifier_magnification_step,
+                        magnifier_min_magnification,
+                        magnifier_max_magnification);
+                    magnifier.magnification(magnification);
+                    return true;
+                case XKB_KEY_underscore:
+                    magnification = std::clamp(
+                        magnification - magnifier_magnification_step,
+                        magnifier_min_magnification,
+                        magnifier_max_magnification);
+                    magnifier.magnification(magnification);
+                    return true;
+                default:
+                    break;
                 }
             }
             else if (modifiers & mir_input_event_modifier_alt)
@@ -249,20 +253,24 @@ int main(int argc, char const* argv[])
                 // Grow area on ctrl alt +/-
                 switch (mir_keyboard_event_keysym(key_event))
                 {
-                    case XKB_KEY_equal:
-                    {
-                        capture_size.width = Width(std::min(1000, capture_size.width.as_int() + 100));
-                        capture_size.height = Height(std::min(1000, capture_size.height.as_int() + 100));
-                        magnifier.capture_size(capture_size);
-                        return true;
-                    }
-                    case XKB_KEY_minus:
-                        capture_size.width = Width(std::max(100, capture_size.width.as_int() - 100));
-                        capture_size.height = Height(std::max(100, capture_size.height.as_int() - 100));
-                        magnifier.capture_size(capture_size);
-                        return true;
-                    default:
-                        break;
+                case XKB_KEY_equal:
+                {
+                    capture_size.width = Width(std::min(1000, capture_size.width.as_int() + 100));
+                    capture_size.height = Height(std::min(1000, capture_size.height.as_int() + 100));
+                    magnifier.capture_size(capture_size);
+                    return true;
+                }
+                case XKB_KEY_minus:
+                    capture_size.width = Width(std::max(100, capture_size.width.as_int() - 100));
+                    capture_size.height = Height(std::max(100, capture_size.height.as_int() - 100));
+                    magnifier.capture_size(capture_size);
+                    return true;
+                case XKB_KEY_m:
+                    decoupled = !decoupled;
+                    magnifier.decouple(decoupled);
+                    return true;
+                default:
+                    break;
                 }
             }
             else
@@ -270,34 +278,16 @@ int main(int argc, char const* argv[])
                 // Turn on/off the magnifier on ctrl +/-
                 switch (mir_keyboard_event_keysym(key_event))
                 {
-                    case XKB_KEY_equal:
-                        magnifier.enable(true);
-                        return true;
-                    case XKB_KEY_minus:
-                        magnifier.enable(false);
-                        return true;
-                    default:
-                        break;
+                case XKB_KEY_equal:
+                    magnifier.enable(true);
+                    return true;
+                case XKB_KEY_minus:
+                    magnifier.enable(false);
+                    return true;
+                default:
+                    break;
                 }
             }
-        }
-
-        return false;
-    };
-
-    // Ctrl+Alt+M: toggle between cursor-follow mode and decoupled (draggable) mode
-    auto decouple_magnifier_filter = [magnifier=magnifier, decoupled=false](MirKeyboardEvent const* key_event) mutable {
-        if (mir_keyboard_event_action(key_event) != mir_keyboard_action_down)
-            return false;
-
-        auto const modifiers = mir_keyboard_event_modifiers(key_event);
-        if ((modifiers & mir_input_event_modifier_ctrl) &&
-            (modifiers & mir_input_event_modifier_alt) &&
-            mir_keyboard_event_keysym(key_event) == XKB_KEY_m)
-        {
-            decoupled = !decoupled;
-            magnifier.decouple(decoupled);
-            return true;
         }
 
         return false;
@@ -429,7 +419,6 @@ int main(int argc, char const* argv[])
             sticky_keys,
             magnifier,
             AppendKeyboardEventFilter{magnifier_filter},
-            AppendKeyboardEventFilter{decouple_magnifier_filter},
             AppendKeyboardEventFilter{sticky_keys_filter},
             cursor_scale,
             locate_pointer,
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -268,9 +268,9 @@ int main(int argc, char const* argv[])
                 case XKB_KEY_m:
                     follows_cursor = !follows_cursor;
                     if (follows_cursor)
-                        magnifier.follow_cursor();
+                        magnifier.set_behavior(miral::Magnifier::Behavior::follow_cursor);
                     else
-                        magnifier.stop_following_cursor();
+                        magnifier.set_behavior(miral::Magnifier::Behavior::freely_positioned);
                     return true;
                 default:
                     break;
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -47,6 +47,7 @@
 
 #include <xkbcommon/xkbcommon-keysyms.h>
 
+#include <algorithm>
 #include <cstdlib>
 #include <cstring>
 #include <linux/input-event-codes.h>
@@ -218,27 +219,25 @@ int main(int argc, char const* argv[])
         {
             if (modifiers & mir_input_event_modifier_shift)
             {
+                auto const magnifier_min_magnification = 1.5f;
+                auto const magnifier_max_magnification = 8.0f;
+                auto const magnifier_magnification_step = 0.5f;
+
                 // Zoom the magnifier in/out on ctrl shift +/-
                 switch (mir_keyboard_event_keysym(key_event))
                 {
                     case XKB_KEY_plus:
-                        magnification += 0.5f;
-                        if (magnification >= 5)
-                        {
-                            magnification = 5;
-                            return true;
-                        }
-
+                        magnification = std::clamp(
+                            magnification + magnifier_magnification_step,
+                            magnifier_min_magnification,
+                            magnifier_max_magnification);
                         magnifier.magnification(magnification);
                         return true;
                     case XKB_KEY_underscore:
-                        magnification -= 0.5f;
-                        if (magnification <= 1)
-                        {
-                            magnification = 1;
-                            return true;
-                        }
-
+                        magnification = std::clamp(
+                            magnification - magnifier_magnification_step,
+                            magnifier_min_magnification,
+                            magnifier_max_magnification);
                         magnifier.magnification(magnification);
                         return true;
                     default:
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -212,7 +212,7 @@ int main(int argc, char const* argv[])
     auto magnifier_filter = [magnifier = magnifier,
                              &magnification,
                              &capture_size,
-                             decoupled = false](MirKeyboardEvent const* key_event) mutable
+                             follows_cursor = true](MirKeyboardEvent const* key_event) mutable
     {
         auto const modifiers = mir_keyboard_event_modifiers(key_event);
 
@@ -266,8 +266,11 @@ int main(int argc, char const* argv[])
                     magnifier.capture_size(capture_size);
                     return true;
                 case XKB_KEY_m:
-                    decoupled = !decoupled;
-                    magnifier.decouple(decoupled);
+                    follows_cursor = !follows_cursor;
+                    if (follows_cursor)
+                        magnifier.follow_cursor();
+                    else
+                        magnifier.stop_following_cursor();
                     return true;
                 default:
                     break;
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -286,6 +286,24 @@ int main(int argc, char const* argv[])
         return false;
     };
 
+    // Ctrl+Alt+M: toggle between cursor-follow mode and decoupled (draggable) mode
+    auto decouple_magnifier_filter = [magnifier=magnifier, decoupled=false](MirKeyboardEvent const* key_event) mutable {
+        if (mir_keyboard_event_action(key_event) != mir_keyboard_action_down)
+            return false;
+
+        auto const modifiers = mir_keyboard_event_modifiers(key_event);
+        if ((modifiers & mir_input_event_modifier_ctrl) &&
+            (modifiers & mir_input_event_modifier_alt) &&
+            mir_keyboard_event_keysym(key_event) == XKB_KEY_m)
+        {
+            decoupled = !decoupled;
+            magnifier.decouple(decoupled);
+            return true;
+        }
+
+        return false;
+    };
+
     StickyKeys sticky_keys = StickyKeys::disabled();
     auto const sticky_keys_filter = [sticky_keys=sticky_keys, sticky_keys_enabled=false](MirKeyboardEvent const* key_event) mutable
     {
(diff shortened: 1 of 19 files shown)


src/miral/touch_emulator.cpp
```diff
@@ -0,0 +1,169 @@
...
+                 !(buttons & mir_pointer_button_primary))
+        {
+            dragging = false;
+            dispatch_touch(mir_touch_action_up, x, y, ts);
+            return true;
+        }
+
+        // Consume all other pointer events so they don't reach surfaces.
+        return true;
+    }
```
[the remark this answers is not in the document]
tarek-y-ismail: Intentional behavior.

examples/miral-shell/shell_main.cpp
```diff
@@ -388,10 +406,12 @@ int main(int argc, char const* argv[])
             sticky_keys,
             magnifier,
             AppendKeyboardEventFilter{magnifier_filter},
+            AppendKeyboardEventFilter{decouple_magnifier_filter},
             AppendKeyboardEventFilter{sticky_keys_filter},
             cursor_scale,
             locate_pointer,
             AppendKeyboardEventFilter{locate_pointer_filter},
+            TouchEmulator{},
             BasicApplicationSwitcher{}
```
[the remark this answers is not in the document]
tarek-y-ismail: Intended behavior.

src/miral/magnifier.cpp
```diff
@@ -144,30 +566,776 @@ class miral::Magnifier::Self
...
+            if (action == mir_touch_action_down)
+                apply_zoom_locked();
+        }
+
+        /// Clamps and applies the zoom step. Callers must hold self->mutex.
+        void apply_zoom_locked()
+        {
+            self->set_magnification_locked(
+                std::clamp(self->magnification + delta, min_magnification + zoom_step, max_magnification));
+        }
```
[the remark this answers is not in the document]
tarek-y-ismail: Hmmm, 1.0 is quite useless for the magnifier, but I didn't want to change existing behavior through the API, hence this behavior.

examples/miral-shell/shell_main.cpp
```diff
@@ -216,27 +218,25 @@ int main(int argc, char const* argv[])
...
-                        if (magnification <= 1)
-                        {
-                            magnification = 1;
-                            return true;
-                        }
-
+                        magnification = std::clamp(
+                            magnification - magnifier_magnification_step,
+                            magnifier_min_magnification,
+                            magnifier_max_magnification);
```
tarek-y-ismail: This is a behavior change, but I believe the minimum shouldn't be 1.0.

Saviq:
> 4 - verify-options-reference-unchanged (Failed)

Not clear why this only fails on Fedora…¿?

src/miral/magnifier.cpp
```diff
@@ -14,31 +14,360 @@
...
+
+static std::string name_for_kind(HandleKind kind)
+{
+    switch (kind)
+    {
+    case HandleKind::drag:     return "magnifier-drag-handle";
+    case HandleKind::resize:   return "magnifier-resize-handle";
+    case HandleKind::zoom_in:  return "magnifier-zoom-in-handle";
+    case HandleKind::zoom_out: return "magnifier-zoom-out-handle";
+    }
```
mattkae: Some UX feedback: add more padding between the plus and minus buttons so that it is less easy to accidentally click the incorrect button with your finger. I got big fingers

src/miral/magnifier.cpp
```diff
@@ -51,123 +380,1069 @@ class miral::Magnifier::Self
...
+    {
+        auto indicator = std::make_shared<HandleIndicator>(
+            rect,
+            kind,
+            server.the_buffer_allocator(),
+            server.the_scene_report(),
+            server.the_display_configuration_observer_registrar());
+
+        server.the_surface_stack()->add_surface(indicator, mi::InputReceptionMode::normal);
+        indicator->set_cursor_image(server.the_default_cursor_image());
```
mattkae: Why must we do this and what happens when we don't?

src/miral/render_scene_into_surface.h
```diff
@@ -19,6 +19,7 @@
 
 #include <functional>
 #include <memory>
+#include <mir/compositor/compositor_id.h>
```
mattkae, line 22: Spurious include?

src/miral/software_buffer_pool.h
mattkae, line 1: Can we rename this file to `software_buffer_pool.h`?

src/miral/software_buffer_stream.h
```diff
@@ -0,0 +1,98 @@
...
+
+private:
+    std::shared_ptr<mir::graphics::Buffer> const buffer;
+    std::function<void()> const release_callback;
+};
+
+/// Minimal double-buffered pool.  Callers claim one slot at a time via
+/// claim(); the slot is returned to the pool when the ClaimedBuffer is
+/// destroyed.  If all slots are in use the pool grows automatically.
+class BufferPool
```
mattkae: Rename to `SoftwareBufferPool`

src/miral/software_buffer_stream.h
```diff
@@ -0,0 +1,98 @@
...
+/// BasicSurface::generate_renderables() includes the surface in every frame.
+class AlwaysHasSubmittedBufferStream : public mir::compositor::Stream
+{
+public:
+    bool has_submitted_buffer() const override { return true; }
+};
+
+/// Wraps a buffer and calls a release callback when destroyed, returning it
+/// to its owning BufferPool.
+class ClaimedBuffer : public mir::graphics::Buffer
```
mattkae: Move this to the implementation file please.

src/miral/software_buffer_stream.h
```diff
@@ -0,0 +1,98 @@
...
+#include <vector>
+
+namespace mir::graphics { class GraphicBufferAllocator; }
+
+namespace miral
+{
+
+/// A mc::Stream that always reports having a submitted buffer so that
+/// BasicSurface::generate_renderables() includes the surface in every frame.
+class AlwaysHasSubmittedBufferStream : public mir::compositor::Stream
```
mattkae: Move this to implementation file please.

src/miral/render_scene_into_surface.cpp
```diff
@@ -319,7 +198,7 @@ class miral::RenderSceneIntoSurface::Self
     }
 
 private:
-    std::mutex mutex;
+    mutable std::mutex mutex;
```
mattkae, line 212: Why is this `mutable` now? It isn't called from a `const` context, right?

src/miral/touch_emulator.cpp
mattkae, line 1: This should have the same name as its header.

src/include/server/mir/scene/prompt_session_creation_parameters.h
mattkae, line 1: This feels like a standalone feature that we may want in another pull request :)

src/miral/touch_emulator.cpp
```diff
@@ -0,0 +1,200 @@
...
+        filter(server);
+    }
+
+    // Called during server init, only when the option is set.
+    void activate(mir::Server& server)
+    {
+        server.the_input_device_registry()->add_device(touch_device);
+        device_registered = true;
+        main_loop = server.the_main_loop();
+        active = true;
```
mattkae: Why do you have both `active` and `device_registered`?

Also -- they are never reset in `deactivate`, so they are always true.

mattkae (changes requested):
Some initial feedback!

I would really like to see the touch emulator be its own pull request, as it would (1) reduce the diff and (2) make it easy to revert one or the other if we have reason to do so. They are separate concerns.

Other than that, I love how it looks! Very nice to use 💯

Saviq:
UX feedback:
- put the buttons inside the magnifier surface
- on +/-, change only the zoom level, not the magnifier size
- I _think_ I would put move button top-right, with +/- right below that, and the resize bottom-right
- we might need a border? might be follow-up, though

src/miral/magnifier.cpp
tarek-y-ismail: The pointer becomes invisible over those surfaces :)

src/miral/render_scene_into_surface.cpp
tarek-y-ismail: Remnant from when I needed to grab the compositor ID. The method (now removed) was const, hence the need to make the mutex mutable.

tarek-y-ismail:
Known issue: The terminal output when clicking handles or resizing is a _little_ noisy.

examples/miral-shell/shell_main.cpp
AlanGriffiths: Explain your reasoning: What good is a magnifier that doesn't shrinks?

tarek-y-ismail: Any smaller and magnification would be at 1.0, which is no magnification

include/miral/miral/magnifier.h
```diff
@@ -27,8 +27,8 @@ namespace miral
 namespace live_config { class Store; }
 
 /// Renders a magnified region of the scene at the cursor position.
-/// By default, the magnifier will magnify a 400x300 region below
-/// the cursor by a 2x magnitude.
+/// By default, the magnifier will magnify a 150x150 region centred on
+/// the cursor by a 1.5x magnitude.
```
AlanGriffiths, line 31: Why change these defaults? 150x150 seems a very small area, especially in the days of hdpi screens

examples/miral-shell/shell_main.cpp
AlanGriffiths: Oh, I misunderstood. You mean the minimum should be more than 1, i.e. 1.5. Why not 1.25?

examples/miral-shell/shell_main.cpp
```diff
@@ -284,6 +283,24 @@ int main(int argc, char const* argv[])
...
+            (modifiers & mir_input_event_modifier_alt) &&
+            mir_keyboard_event_keysym(key_event) == XKB_KEY_m)
+        {
+            decoupled = !decoupled;
+            magnifier.decouple(decoupled);
+            return true;
+        }
+
+        return false;
+    };
```
AlanGriffiths: Why is this an additional filter? Can't the magnifier plumbing all be in one place?

examples/miral-shell/shell_main.cpp
```diff
@@ -216,27 +217,25 @@ int main(int argc, char const* argv[])
         {
             if (modifiers & mir_input_event_modifier_shift)
             {
+                auto const magnifier_min_magnification = 1.5f;
+                auto const magnifier_max_magnification = 8.0f;
+                auto const magnifier_magnification_step = 0.5f;
```
AlanGriffiths: This seems a very big step. I would go for something in the range [0.1,0.25]

AlanGriffiths (changes requested):
Not tried it on a touch screen yet (none handy). But there are some odd behaviors.

I ran miral-app, opened a terminal and maximized it.

Then after activating the magnifier, and switching to "decoupled" mode I pushed it to the top left over the menus. (Had to increase the area to a decent size.)

Now I expect the menus to behave according to where the cursor appears, but the cursor position isn't being remapped.

Also, the "File" menu is inaccessibe as it is under the top left control.

tarek-y-ismail:
>  Now I expect the menus to behave according to where the cursor appears, but the cursor position isn't being remapped.

Until we figure out a way to correctly map clicks and drags (blocked at them moment by global pointer state shenanigans), input shouldn't get through. This PR is just for the visual updates. I just forgot to add a filter to consume all pointer events in decoupled mode.

> Also, the "File" menu is inaccessibe as it is under the top left control.

Yeah, I can't come up with many ideas to work around this other than moving the resize handle. That was in the very first version, but it introduced significant complexity.

examples/miral-shell/shell_main.cpp
tarek-y-ismail: I juggled both values, just thought that 1.5 is a good start but I'm open to 1.25.

include/miral/miral/magnifier.h
tarek-y-ismail: It has always been 150x150, this is just a stale comment.

tarek-y-ismail:
:facepalm:

include/miral/miral/magnifier.h
```diff
@@ -43,12 +43,21 @@ class Magnifier
...
     Magnifier& magnification(float magnification);
     Magnifier& capture_size(mir::geometry::Size const& size);
 
+    /// When \p decoupled is true, the magnifier surface is no longer tied to
+    /// the cursor position and can instead be dragged freely with the pointer
+    /// or a single touch contact.  When false (the default), the surface
+    /// follows the cursor as before.
+    /// \remark Since MirAL 6.0
+    Magnifier& decouple(bool decoupled);
+
```
AlanGriffiths: The name `decouple()` doesn't immediately convey its function. Even the doc comment uses "tied" rather than "coupled". We also tend to use paired methods (see, for example, `enable()`/`disable()`) rather than `bool` parameters to indicate state transition requests.

> When false (the default),

I think this exposition is misplaced. "false"/"untied"/"decoupled" is the default for the initial state, not the default for this function.

src/miral/basic_window_manager.cpp
```diff
@@ -881,7 +892,16 @@ auto miral::BasicWindowManager::window_at(geometry::Point cursor) const
...
     auto surface_at = focus_controller->surface_at(cursor);
-    return surface_at ? info_for(surface_at).window() : Window{};
+    if (!surface_at)
+        return Window{};
+
+    // surface_at() returns any hit-testable scene surface, including internal
+    // compositor surfaces that are not managed windows.  surface_known() guards
+    // against the window_info lookup throwing for unregistered surfaces.
+    if (!surface_known(surface_at, "window_at"))
+        return Window{};
```
AlanGriffiths: This explanation suggests that `focus_controller->surface_at(cursor)` is returning the wrong result. Why would window management be told about surfaces that are not managed windows?

Note that this is a novel use of `surface_known()`: existing uses are checking that `surface` is still valid and the text supplied to the `action` parameter is a window management action, not a function called by the implementation.

src/miral/basic_window_manager.h
```diff
@@ -272,7 +280,7 @@ class BasicWindowManager : public virtual mir::shell::WindowManager,
     void update_event_timestamp(MirTouchEvent const* tev);
     void update_event_timestamp(MirInputEvent const* iev);
 
-    auto surface_known(std::weak_ptr<mir::scene::Surface> const& surface, std::string const& action) -> bool;
+    auto surface_known(std::weak_ptr<mir::scene::Surface> const& surface, std::string const& action) const -> bool;
```
AlanGriffiths: [non-blocking] I'm not sure the `surface_known()` check makes sense for `const` contexts.

src/miral/magnifier.cpp
```diff
@@ -14,35 +14,574 @@
...
+/// needs a full diameter of clearance both above and below the centred stack.
+auto const min_visual_dimension = 2 * drag_handle_diameter + zoom_stack_height.as_int();
+
+/// Magnification step applied by each zoom button press.
+auto const zoom_step = 0.25f;
+
+/// Painter colour/alpha constants for handle indicator graphics.
+uint8_t constexpr disc_grey  = 40;
+uint8_t constexpr icon_grey  = 210;
+uint8_t constexpr background_alpha = 150;
```
AlanGriffiths: Most of these constants look more like default values for configuration than fixed values that need to be used in multiple places. Do they really belong at namespace scope? Or can they be localized to the place(s) they are used? (That avoids accidentally using the default value when the configured value should be taken.)

src/miral/magnifier.cpp
```diff
@@ -14,35 +14,574 @@
...
+            handle_rect,
+            kind,
+            capture_compositor_id,
+            server.the_buffer_allocator(),
+            server.the_scene_report(),
+            server.the_display_configuration_observer_registrar());
+        handle_surface_stack = server.the_surface_stack();
+
+        server.the_surface_stack()->add_surface(indicator, mi::InputReceptionMode::normal);
+        indicator->set_cursor_image(server.the_default_cursor_image());
```
AlanGriffiths, line 83: We shouldn't be creating surfaces during the "init" phase of bringing the compositor up. The purpose of the single threaded "init" phase is for parts of the system to "wire up" connections to the things they need to communicate with. The only established exception to this is queuing work on `the_main_loop()` which is appropriate here.

src/miral/magnifier.cpp
```diff
@@ -14,35 +14,574 @@
...
+{
+public:
+    HandleIndicator(
+        geom::Rectangle const& initial_rect,
+        HandleKind kind,
+        mc::CompositorID capture_compositor_id,
+        std::shared_ptr<mg::GraphicBufferAllocator> const& allocator,
+        std::shared_ptr<ms::SceneReport> const& scene_report,
+        std::shared_ptr<mir::ObserverRegistrar<mg::DisplayConfigurationObserver>> const& display_config_registrar) :
+        BasicSurface{
```
AlanGriffiths: We probably don't want this to be `mir_window_type_normal` (I don't have a good suggestion here). But that essentially says it is controlled by window management, which we know is wrong.

AlanGriffiths (changes requested):
Incomplete review as I hit EOD. Nothing seriously problematic, just a lot of tidy-up

src/miral/basic_window_manager.cpp
tarek-y-ismail: Without some sort of check, the window manager would try accessing a window that doesn't exist in `window_info` and crash. Hence why these changes were made initially. 

I later discovered that `surface_known` also prints unknown surfaces, causing a log statement for every click of a handle or the magnification surface.

I've chopped down the changes to just checking `window_info`.

AlanGriffiths: > Without some sort of check, the window manager would try accessing a window that doesn't exist in

Only if a window that doesn't exist for window management is returned by `surface_at(cursor)`

src/miral/magnifier.cpp
tarek-y-ismail: Maybe `mir_window_type_utility`? But anyway, we don't expose a way to set the window type other than through the legacy `configure`. I tried exposing `set_type` just to experiment, and `utility` acts the same as `normal`.

AlanGriffiths: `mir_window_type_utility` is wrong too - that's a thing like floating audio controls that IS managed by window management.

What we actually need is a surface (which _shouldn't_ be the same as a window [rant off]) that is visible to compositing, but not window management. (At least not to the `WindowManagementStrategy` extension point.)

tarek-y-ismail: I agree that it's probably time to poke at this. Should I spin this discussion off into an issue?

tarek-y-ismail:
Bruh :facepalm:

src/include/server/mir/recent_items.h
```diff
@@ -0,0 +1,58 @@
...
+#include <optional>
+
+namespace mir
+{
+/// A fixed-capacity circular buffer of recently added items.
+template<typename T, std::size_t Capacity>
+class RecentItems
+{
+public:
+    static_assert(Capacity > 0);
```
AlanGriffiths: Clearer as a requires clause?

src/platform/renderers/gl/renderer.cpp
```diff
@@ -123,6 +125,122 @@ class GLMultiHandle
...
+
+        values->add(value);
+        return true;
+    }
+
+private:
+    mir::Synchronised<mir::RecentItems<T, Capacity>> items;
+};
+
+BoundedSet<RendererDiagnostics, 32> reported_diagnostics;
```
AlanGriffiths: This is a variable at namespace scope. That is generally a bad idea for synchronization and reproducibility. (You can likely find lots of discussion why Monostate is an anti-pattern.)

It is likely better to supply an explicit dependency that incorporates the desired behavior.

src/platform/renderers/gl/renderer.cpp
```diff
@@ -123,6 +125,122 @@ class GLMultiHandle
...
+public:
+    auto insert(T const& value) -> bool
+    {
+        auto values = items.lock();
+
+        if (values->contains(value))
+            return false;
+
+        values->add(value);
+        return true;
```
AlanGriffiths: Is this the right algorithm? One alternative would be to drop any existing matching entry and always add the new one.

src/platform/renderers/gl/renderer.cpp
```diff
@@ -123,6 +125,122 @@ class GLMultiHandle
...
+        {GL_VERSION,                   "GL version"},
+        {GL_SHADING_LANGUAGE_VERSION, "GLSL version"},
+        {GL_EXTENSIONS,               "GL extensions"},
+    };
+
+    for (auto const& [index, glstring] : glstrings | std::views::enumerate)
+    {
+        // TICS !cppcoreguidelines-pro-type-reinterpret-cast: glGetString returns an ASCII string,
+        // guaranteed not to have the high-bit set, so it is representationally-identical to signed char.
+        diagnostics.gl_strings[index] = as_string(reinterpret_cast<char const*>(glGetString(glstring.id)));
```
AlanGriffiths: [non-blocking] It seems unfortunate that the simple logic here is hidden by the TICS comment and `reinterpret_cast<>`.

But a `GLubyte const*` overload of `as_string()` is probably too weak (as `GLubyte` is a type alias, not actually a type).

src/platform/renderers/gl/renderer.cpp
```diff
@@ -123,6 +125,122 @@ class GLMultiHandle
 using TextureHandle = GLMultiHandle<&glDeleteTextures>;
 using FramebufferHandle = GLMultiHandle<&glDeleteFramebuffers>;
 
+struct RendererDiagnostics
+{
+    EGLDisplay display;
+    std::array<std::string, 4> egl_strings;
+    std::array<std::string, 5> gl_strings;
```
AlanGriffiths: Why are we creating and storing strings? The results from `eglQueryString()` and `glGetString()` are pointers to string constants: they can be stored directly.

AlanGriffiths (changes requested):
Really only looked at the "Deduplicate GL renderer diagnostics" commit so far. Does this deserve its own PR?

Essentially, I would prefer to see `Renderer::Renderer(...)` take a Diagnostics interface with a `log()` method.

src/platform/renderers/gl/renderer.cpp
tarek-y-ismail: Tried it. Moving to `std::string_view` breaks de-duplication. Moving back to `std::string` has it working again. I'm not sure what the root cause is, but I assume that our assumptions about the lifetime of the driver provided strings is incorrect.

src/platform/renderers/gl/renderer.cpp
tarek-y-ismail: Is there a difference? When we already have an equal entry in the list, we just don't add the new one. This is less work than removing then re-adding it.

src/platform/renderers/gl/renderer.cpp
tarek-y-ismail: My main concern was that not all users will get this fix, but since we provide access through the renderer factory, I think it should be alright.

Though this breaks the renderer factory ABI since we store a pointer to the new logger class inside of it. I'm not sure what ways we have around this. The highest we can go is one layer above `Renderer::Renderer`, but the highest we can go is the renderer factory.

I know we're already breaking ABI so my solution may be fine. Take a look at it in f613ea3

include/miral/miral/magnifier.h
```diff
@@ -27,8 +27,8 @@ namespace miral
 namespace live_config { class Store; }
 
 /// Renders a magnified region of the scene at the cursor position.
-/// By default, the magnifier will magnify a 400x300 region below
-/// the cursor by a 2x magnitude.
+/// By default, the magnifier will magnify a 150x150 region centred on
```
mattkae, line 30:
```suggestion
/// By default, the magnifier will magnify a 150x150 region centered on
```

include/miral/miral/magnifier.h
```diff
@@ -43,12 +43,23 @@ class Magnifier
...
     Magnifier& capture_size(mir::geometry::Size const& size);
 
+    /// Makes the magnifier follow the cursor.
+    /// \remark Since MirAL 6.0
+    Magnifier& follow_cursor();
+
+    /// Stops the magnifier following the cursor, allowing it to be dragged
+    /// freely with the pointer or a single touch contact.
+    /// \remark Since MirAL 6.0
+    Magnifier& stop_following_cursor();
```
mattkae: I would personally prefer something like this, as it isn't entirely unlikely that we'd have a future behavior, and this can prevent API churn:

```suggestion
    enum class MagnifierBehavior {
        center_on_cursor,
        freestyle
    }
    
    Magnifier& set_behavior(MagnifierBehavior behavior);
```

src/platform/renderers/gl/renderer_logger.h
```diff
@@ -0,0 +1,104 @@
...
+    std::array<std::string, 5> gl_strings;
+    GLint max_texture_size;
+    std::array<GLint, 6> framebuffer_bits;
+
+    auto operator==(Capabilities const&) const -> bool = default;
+};
+
+auto collect_capabilities() -> Capabilities;
+
+class Logger
```
mattkae: This work feels like another, unrelated pull request :)

src/miral/magnifier.cpp
mattkae: Yeah that might be an issue. This is like an "attached" window or something, but not really

mattkae (changes requested):
An initial suggestion while I read this drawing code

include/miral/miral/magnifier.h
tarek-y-ismail: Both are valid AFAIK?

tarek-y-ismail:
Huh? Why was this closed?

Saviq:
> Huh? Why was this closed?

Looks like a stray "Fixes #…", stoopid of GH to close _PRs_ from other PRs :)

tarek-y-ismail:
History seems somewhat saner now. Will re-review and mark as ready ASAP (tomorrow)

include/miral/miral/magnifier.h
```diff
@@ -43,12 +51,21 @@ class Magnifier
     ///     region that will be magnified.
     ///     - {magnifier, capture_size, height}: The height of the rectangular
     ///     region that will be magnified.
+    ///     - {magnifier, follow_cursor}: Whether the magnifier follows the
+    ///     cursor.
```
tarek-y-ismail, line 55: Should be updated to be a string to allow later expansion along with `Behavior`.

src/miral/magnifier_handle_indicator.cpp
```diff
@@ -0,0 +1,324 @@
...
+#include <mir/geometry/dimensions.h>
+#include <mir/graphics/graphic_buffer_allocator.h>
+
+namespace geomtery = mir::geometry;
+namespace mg = mir::graphics;
+namespace mrs = mir::renderer::software;
+
+namespace
+{
+class BufferPainter
```
tarek-y-ismail, line 28: May be useful to push this down so it's available to `BasicDecoration` as well.

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+    if (!contains(visual_bounds(placement)))
+        return false;
+
+    auto const handles = handle_positions(placement);
+    for (auto const& handle : {handles.drag, handles.resize, handles.zoom_in, handles.zoom_out})
+    {
+        if (contains({handle, {handle_diameter, handle_diameter}}))
+            return false;
+    }
```
tarek-y-ismail, line 202: Since all handles are placed _inside_ the magnifier surface, we don't need to check the handles as well.

src/miral/magnifier.cpp
```diff
@@ -14,169 +14,883 @@
...
+        auto s = state.lock();
+        if (!s->follow_cursor)
+            return;
+
+        s->follow_cursor = false;
+
+        if (auto const surf = s->surface.lock())
+        {
+            attach_observers(*s);
+            if (s->default_enabled)
```
tarek-y-ismail, line 447: Usage makes it read more like `enabled` than `default_enabled`

src/miral/magnifier.cpp
```diff
@@ -14,169 +14,883 @@
...
+                .value_or(false);
+    }
+
+    if (pointer_gesture_consumed)
+    {
+        auto const consumed = *pointer_gesture_consumed;
+        if (action == mir_pointer_action_button_up && event.buttons() == 0)
+            pointer_gesture_consumed.reset();
+        return consumed;
+    }
```
tarek-y-ismail, line 832: Why do we need all of this?

src/miral/magnifier.cpp
```diff
@@ -14,169 +14,883 @@
...
+    s->screen_bounds = rects;
+    if (auto const surf = s->surface.lock())
+    {
+        auto const top_left = surf->top_left();
+        auto const pre_clamp_visual_size = s->visual_size;
+        auto const layout = s->layout();
+        auto const placement =
+            s->follow_cursor ?
+                layout.centered_on(s->cursor_pos) :
+                layout.freely_positioned(top_left);
```
tarek-y-ismail, line 883: This seems to be a recurring pattern.

src/miral/magnifier.cpp
```diff
@@ -14,169 +14,883 @@
...
+    geom::Rectangles rects;
+    config->for_each_output(
+        [&rects](mg::DisplayConfigurationOutput const& output)
+        {
+            if (output.used && output.connected)
+                rects.add(output.extents());
+        });
+
+    auto s = self.state.lock();
+    s->screen_bounds = rects;
```
tarek-y-ismail, line 874: How do the screen bounds affect the code after?

tests/miral/magnifier.cpp
```diff
@@ -78,26 +122,506 @@ TEST_F(MagnifierTest, magnification_results_in_scaled_transform)
...
+{
+    magnifier.enable(true).set_behavior(Magnifier::Behavior::freely_positioned);
+    start_server();
+    wait_for_magnifier_initialization();
+
+    auto const mux = server().the_cursor_observer_multiplexer();
+    auto sentinel = std::make_shared<SentinelCursorObserver>();
+    mux->register_interest(sentinel);
+    ASSERT_TRUE(sentinel->signal.wait_for(2s)) << "timed out waiting for initial cursor state";
+    sentinel->signal.reset();
```
tarek-y-ismail, line 230: This is a repeating pattern.

tests/miral/magnifier.cpp
```diff
@@ -78,26 +122,506 @@ TEST_F(MagnifierTest, magnification_results_in_scaled_transform)
...
+TEST_F(MagnifierHandleTest, double_clicking_drag_handle_does_not_move_magnifier)
+{
+    auto const before = magnifier_top_left();
+    auto const handle_center = element_center(drag_handle_index);
+
+    click(handle_center);
+    click(handle_center);
+
+    EXPECT_THAT(magnifier_top_left(), Eq(before));
+}
```
tarek-y-ismail, line 470: That feature was removed. This can be discarded.

tests/miral/magnifier.cpp
```diff
@@ -78,26 +122,506 @@ TEST_F(MagnifierTest, magnification_results_in_scaled_transform)
...
+            1.0f,
+            1.0f,
+            1.0f,
+            0.0f}});
+
+    EXPECT_FALSE(composite_event_filter.lock()->handle(*pointer_event));
+    EXPECT_FALSE(composite_event_filter.lock()->handle(*touch_event));
+}
+
+TEST_F(MagnifierHandleTest, does_not_consume_events_when_magnifier_is_inactive)
```
tarek-y-ismail, line 539: misleading name

tests/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,106 @@
...
+TEST(MagnifierLayout, chooses_output_using_visual_bounds)
+{
+    auto bounds = outputs();
+    bounds.add({{800, 0}, {400, 300}});
+    miral::MagnifierLayout const layout{bounds, {500, 500}, 2.0f};
+
+    auto const placement = layout.centered_on({790, 150});
+
+    EXPECT_THAT(placement.visual_size, Eq(geom::Size{500, 480}));
+}
```
tarek-y-ismail, line 64: This is incorrect. If we center the magnifier on (790, 150), with a size of (500, 500), this means the magnifier would be extending outside the second smaller screen on both the x and y axes. This is not possible via dragging but may be possible when the cursor is moved and then the magnifier is set to not follow it.

In this case, the magnifier should probably snap so that
1. The drag handle is visible
2. The magnifier fits whatever output it's mostly in

I believe that in the above case, the magnifier should snap so that its bottom right sits at the bottom right of the original (800, 600) output.

tarek-y-ismail:
Superseded by a stack of smaller PRs, split as discussed:

1. #5132 — `miral: guard window_at() against unregistered scene surfaces`
2. #5133 — `miral: extract software buffer pooling into SoftwareBufferPool`
3. #5134 — `miral: add magnifier layout geometry helpers`
4. #5135 — `miral: add freely-positioned magnifier mode with on-screen controls`
5. #5136 — `miral: consume input over the freely-positioned magnifier`

Each is stacked on the previous one, so reviewing the last commit of each (or diffing against its base branch) shows only that PR's change. The tip of the stack is content-identical to this branch apart from moving `HandleKind` into `magnifier_layout.h` so the layout layer stands alone.

tarek-y-ismail:
Closed in favor of a more reviewable stack.

PR #5074, opened 2026-07-08
Clarify input trigger protocol descriptions

robert-ancell:
Improve the documentation of the ext_input_trigger_registration_v1 and ext_input_trigger_action_v1 protocols to follow Wayland conventions and describe object roles and lifecycles more clearly:

  - Rewrite both top-level protocol descriptions to explain, at a high level, what the protocols are for and how the provider and subscriber sides fit together, each with a pseudocode example.
  - Note the manager interfaces are globals advertised by the registry.
  - Give the ext_input_trigger_v1, ext_input_trigger_action_control_v1 and ext_input_trigger_action_v1 interfaces concise noun-phrase summaries and descriptions of what each object represents, how it is created, and its destruction semantics.
  - Explain that cancel_and_destroy purges an action (notifying derived ext_input_trigger_action_v1 objects via the unavailable event) whereas destroy leaves the action intact.
  - De-circularise the unavailable event description and name its causes.
  - Fix wording errors (action vs "activation"; a request described as an "event").

Rename the ext_input_trigger_action_control_v1 request cancel to cancel_and_destroy in the generated interface, updating the InputTriggerActionControlV1 override to match.


robert-ancell:
Updated documentation in preparation for proposal to wayland-protocols. Note I renamed `ext_input_trigger_action_control_v1.cancel` to `ext_input_trigger_action_control_v1.cancel_and_destroy` which is a protocol change but I think an important clarification and matches existing Wayland protocols. The other changes are just informational.

robert-ancell:
I will propose these to wayland-protocols from https://gitlab.freedesktop.org/rancell/wayland-protocols/-/commits/ext-input-trigger

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -206,35 +244,51 @@
...
-    <request name="cancel" type="destructor">
-      <description summary="purge the input trigger action">
-        This informs the compositor that the input trigger action will no
-        longer be used. This request may be made when the user requests to
-        delete an input trigger.
+    <request name="cancel_and_destroy" type="destructor">
+      <description summary="purge the input trigger action and destroy this object">
+        This purges the input trigger action and destroys this object. It would
+        typically be used when the user deletes the input trigger the action
+        represents.
```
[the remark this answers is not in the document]
robert-ancell: As discussed out of band, we don't think anyone is using these protocols yet so this is acceptable.

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -22,19 +22,39 @@
...
 
-  <description summary="Registration for global input actions">
-    This protocol allows for a Wayland clients to subscribe to input trigger
-    actions for "begin" and "end" notifications.
-
-    ext_input_trigger_action_v1 and the associated ext_input_trigger_registration_v1
-    are based on the following discussion document:
-    https://docs.google.com/document/d/1Hj3BXkmW1quIEWqmN7xeJY29nSPfXP2bmoZvpOsEeJQ/
+  <description summary="Subscription for global input actions">
+    This protocol lets a client react to a global input action. An action is an
```
tarek-y-ismail:
```suggestion
    This protocol lets a client react to a global action. An action is an
```

It doesn't have to be an input action per-se. Yes, we only expose keyboard shortcuts as a start, but the protocol was designed to support other things: programmatic action ("tell me when xyz happens"), gestures, etc...

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -22,19 +22,39 @@
...
+    A client subscribes by passing a token to
+    `ext_input_trigger_action_manager_v1.get_input_trigger_action`. The token
+    is issued by the compositor, delivered to the registering client, and then
+    passed to interested clients out of band (for example through a desktop
+    settings service).
+
+    For example, a media player reacting to the system "play/pause" action:
+
+        # obtain the action's token out of band (e.g. from a settings
+        # service), then subscribe to the action
```
tarek-y-ismail:
```suggestion
        # service), then subscribe to the action
        token = ...
```

I know this is pseudocode, but this is slightly less confusing.

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -22,19 +22,39 @@
...
+    This protocol lets a client react to a global input action. An action is an
+    activity (for example "play/pause" or "toggle-launcher") that a privileged
+    client has bound to one or more input triggers, such as a key combination,
+    using the `ext_input_trigger_registration_v1` protocol.
+
+    A client subscribes by passing a token to
+    `ext_input_trigger_action_manager_v1.get_input_trigger_action`. The token
+    is issued by the compositor, delivered to the registering client, and then
+    passed to interested clients out of band (for example through a desktop
+    settings service).
```
tarek-y-ismail: I think we should mention that the token is delivered to the registering client, which can itself listen to the actions and notify its clients indriectly, _or_ share the tokens for other clients to listen directly.

Something like:
```suggestion
    is issued by the compositor, delivered to the registering client, it can then
    be passed to interested clients out of band (for example through a desktop
    settings service) to listen directly to `begin` and `end` events, or the registering 
    client can itself listen and notify interested clients through any other method.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -23,18 +23,42 @@
...
+    (for example "play/pause" or "toggle-launcher"); a trigger is an input, such
+    as "Meta+p", that fires it. When a trigger fires, the client subscribed to
+    the action is notified through the `ext_input_trigger_action_v1` protocol.
+
+    The registering client reserves a trigger (obtaining an
+    `ext_input_trigger_v1`), creates an action (obtaining an
+    `ext_input_trigger_action_control_v1` and, with it, a token), and associates
+    triggers with the action. It then delivers the token to interested clients
+    out of band (for example through a desktop settings service), which subscribe
+    using `ext_input_trigger_action_manager_v1.get_input_trigger_action`.
```
tarek-y-ismail: Should be updated to follow https://github.com/canonical/mir/pull/5074/changes#r3550994054

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -23,18 +23,42 @@
...
+        # bind the manager global and reserve a key combination
+        manager = registry.bind(ext_input_trigger_registration_manager_v1)
+        trigger = manager.register_keyboard_sym_trigger(Meta, XKB_KEY_p)
+
+        # create the action and collect its token
+        control = manager.get_action_control("play/pause")
+        on control.done(token):
+            control.add_input_trigger_event(trigger)
+            # hand the token to interested clients out of band
+            settings_service.publish("play/pause", token)
```
tarek-y-ismail, line 53: I believe this is wrong. 

You are supposed to use the token returned in `done` to call `get_input_trigger_action`, then listen to the events on that. _Or_ pass the tokens to other clients to listen.

It would look something like:
```suggestion
    control.add_input_trigger_event(trigger)

    on control.done(token):
       action = action_manager.get_input_trigger_action(token)
       
       on action.begin(time, token):
         # Do something like notifying all subscribers
         
       on action.end(time, token):
         # Do something else.
         
       settings_service.publish("play/pause")
       
       # Or, if clients would like to listen directly
       settings_service.publish("play/pause", token)
```

AlanGriffiths (approved):
LGTM

tarek-y-ismail:
@AlanGriffiths Sorry, I'm still looking at it.

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -152,16 +177,19 @@
...
-      register the trigger event.
-
-      The failed event indicates that the trigger is not "owned" by the manager.
-      This means the object is of no further use and should be destroyed.
+    <description summary="a reserved input trigger">
+      This object represents a client's request to reserve an input trigger.
+      In response to its creation the compositor must send either the `done`
+      or the `failed` event.
+
+      The done event indicates the trigger is now owned by the manager: the
```
tarek-y-ismail:
```suggestion
      The `done` event indicates the trigger is now owned by the manager: the
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -152,16 +177,19 @@
...
+      compositor will not allow this or any other manager to register the same
+      trigger. The reservation lasts until the object is destroyed, which
+      releases the trigger.
+
+      The failed event indicates the trigger could not be owned by the manager,
```
tarek-y-ismail:
```suggestion
      The `failed` event indicates the trigger could not be owned by the manager,
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -186,9 +214,19 @@
...
-      In response to the creation of this object the compositor must send
-      the done event.
+    <description summary="an input trigger action">
+      This object represents an input trigger action owned by this client. The
+      action associates one or more input triggers with a token that other
+      clients can use, via the `ext_input_trigger_action_v1` protocol, to be
+      notified when the action fires.
+
+      In response to the creation of this object the compositor must send the
+      done event, which delivers the action's token. Triggers are associated
```
tarek-y-ismail:
```suggestion
      `done` event, which delivers the action's token. Triggers are associated
```

tarek-y-ismail (changes requested):
Some thoughts

wayland-protocols/ext-input-trigger-registration-v1.xml
tarek-y-ismail: I may have broken the indentation...

tarek-y-ismail: One more note: The order difference between adding the trigger or receiving `done` on the action control object is not important. The action is associated/hooked up when it's obtained through `get_input_trigger_action` via the token.

wayland-protocols/ext-input-trigger-action-v1.xml
robert-ancell: Does this also apply to the summary? (which was described originally as "global input actions")

wayland-protocols/ext-input-trigger-registration-v1.xml
robert-ancell: This should be improved with the changes from the previous comments.

wayland-protocols/ext-input-trigger-action-v1.xml
robert-ancell: Assuming yes.

wayland-protocols/ext-input-trigger-action-v1.xml
```diff
@@ -22,19 +22,48 @@
...
+    `ext_input_trigger_registration_v1` protocol, in which case the compositor
+    issues the token to the registering client, or out of band from another
+    client that registered it (for example through a desktop settings service).
+
+    For example, a media player reacting to the system "play/pause" action:
+
+        # acquire the token. Either receive it from a remote service...
+        token = settings_service.lookup("play/pause")
+
+        # ...or register the action yourself using
```
tarek-y-ismail: "register" here is misleading: the code below is just setting up listening. It's missing the actual registration of the keyboard trigger that would alert listeners. 

This may be better

```suggestion
        # ...or listen to an already registered action yourself using
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -22,19 +22,50 @@
...
-    https://docs.google.com/document/d/1Hj3BXkmW1quIEWqmN7xeJY29nSPfXP2bmoZvpOsEeJQ/
+  <description summary="Registration of input triggers and associated actions">
+    This protocol lets a privileged client define actions and bind them to
+    input triggers, such as key combinations. An action is an activity
+    (for example "play/pause" or "toggle-launcher"); a trigger is an input, such
+    as "Meta+p", that fires it. When a trigger fires, the client subscribed to
+    the action is notified through the `ext_input_trigger_action_v1` protocol.
+
+    The registering client reserves a trigger (obtaining an
+    `ext_input_trigger_v1`), creates an action (obtaining an
```
tarek-y-ismail:
```suggestion
    `ext_input_trigger_v1`), creates an action control (obtaining an
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -22,19 +22,50 @@
...
+    `ext_input_trigger_action_control_v1` and, with it, a token), and associates
+    triggers with the action. The token may then be used with
```
tarek-y-ismail:
```suggestion
    triggers with the action control object. The token may then be used with
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -47,9 +78,10 @@
 
     <event name="capabilities">
       <description summary="input trigger capabilities">
-        This is emitted when ext_input_trigger_registration_manager_v1 is bound
-        and whenever a system gains or loses the capabilities. The argument is
-        a capability enum containing the complete set of capabilities.
+        This is emitted when `ext_input_trigger_registration_manager_v1` is
+        bound and whenever a system gains or loses the capabilities. The
```
tarek-y-ismail, line 82: Note to self: We probably don't do the whole "whenever a system gains or loses the capabilities" part in Mir.

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -179,16 +214,26 @@
...
     <event name="failed">
       <description summary="input trigger failed to register">
-        The failed event indicates that the trigger is not "owned" by the manager.
+        The `failed` event indicates that the trigger is not "owned" by the manager.
         This means the object is of no further use and should be destroyed.
       </description>
     </event>
   </interface>
 
   <interface name="ext_input_trigger_action_control_v1" version="1">
```
tarek-y-ismail, line 223: I'm having second thoughts about naming it 'action control'. Maybe this should be named 'action' and the current action renamed 'action listener'.

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -206,35 +251,51 @@
...
-    <request name="cancel" type="destructor">
-      <description summary="purge the input trigger action">
-        This informs the compositor that the input trigger action will no
-        longer be used. This request may be made when the user requests to
-        delete an input trigger.
+    <request name="cancel_and_destroy" type="destructor">
+      <description summary="purge the input trigger action and destroy this object">
+        This purges the input trigger action and destroys this object. It would
+        typically be used when the user deletes the input trigger the action
+        represents.
```
tarek-y-ismail: Maybe a bit too rambly, but here's a suggestion:
```suggestion
        This purges the input trigger action control and destroys this object. It is up to the owning client when to call this. For example: a shortcut control center can expose a trigger deletion option, which would invoke this request when the last input trigger associated with the action group is deleted. Or it could expose an action group deletion option, which would call this directly.
```

wayland-protocols/ext-input-trigger-registration-v1.xml
```diff
@@ -206,35 +251,51 @@
...
 
-        Any ext_input_trigger_action_v1 created from the associated token
-        by this or other clients will become unavailable.
+        To destroy this object while leaving the action (and any
+        `ext_input_trigger_action_v1` objects derived from it) intact, use
+        `destroy` instead.
       </description>
     </request>
 
     <request name="destroy" type="destructor">
```
tarek-y-ismail, line 291: Is it logical to destroy the action control and leave actions intact? Since the action control is basically what associates triggers with actions, destroying the action control would break this association. Later `begin` and `end` events wouldn't make sense.

Thinking it from another point of view, the action control is meant to be owned by the privileged client that set up the triggers. If that client (a shortcut center for example) destroys this action control, or is itself destroyed, should triggers still send `begin` and `end` events?

tarek-y-ismail (changes requested):
Some more feedback. 

There's some work going on in parallel that relates to global shortcuts: https://www.youtube.com/watch?v=VEXugiJsNbM. We should take that into account and see if we should merge the two proposals or keep them separate as two different approaches.

wayland-protocols/ext-input-trigger-registration-v1.xml
AlanGriffiths: Yes. Suppose the privileged client intends (or is forced) to exit. Should existing actions stop working?

AlanGriffiths:
@robert-ancell has this been forgotten?

wayland-protocols/ext-input-trigger-action-v1.xml
robert-ancell: Yes, this is referring to an already registered action, I'll reword. The suggested change didn't quite make sense in the second line.

wayland-protocols/ext-input-trigger-registration-v1.xml
robert-ancell: I think changing the names is out of scope for this PR - make a follow up PR if necessary.

wayland-protocols/ext-input-trigger-registration-v1.xml
robert-ancell: What is the intended current behaviour? This PR is not intended to change the behaviour but better document what it is today.

AlanGriffiths: The intention is that

* `ext_input_trigger_action_control_v1.cancel()` destroys the control _and_ cancels any corresponding actions; and,
* `ext_input_trigger_action_control_v1.destroy()`  destroys the control instance only.

tarek-y-ismail:
@robert-ancell Did you forget to push your commits?

tarek-y-ismail (approved):
I'll let earlier comments slide as nits. Here's an approval so you can push your changes and merge.

robert-ancell:
> @robert-ancell Did you forget to push your commits?

Oops, this is truly the forgotten PR.

AlanGriffiths (approved):

PR #5075, opened 2026-07-08
Only warn on unsupported input trigger types

robert-ancell:
InputTriggerActionControlV1::add_input_trigger_event and drop_input_trigger_event logged an "Unsupported trigger type" warning unconditionally, even when a valid keyboard trigger was supplied, because the success branch fell through to the warning. Return after handling a recognised trigger so the warning is only emitted for genuinely unsupported types.


AlanGriffiths:
Looks familiar: #5082

PR #5078, opened 2026-07-08
miral: use std::span in interface where possible

AlanGriffiths:
Updates a bunch of APIs to use std::span


mattkae (approved):
Nice, I never use `std::span`, but I should :)

PR #5085, opened 2026-07-10
Drop `ServerConfiguration::the_screen_shooter()`

AlanGriffiths:
it is both unused and deprecated


mattkae (approved):
Yup, I remember this one!

PR #5086, opened 2026-07-10
Use `mir_add_test` instead of `add_test`

tarek-y-ismail:
Closes https://github.com/canonical/mir/pull/5071#issuecomment-4933416601

## What's new?

Modifies the options reference verification test to use `mir_add_test` instead of `add_test`. Fedora runs `ctest` directly, but other platforms go through `discover_all_tests.sh` first, and `add_test` didn't add the options reference verification test to that.


Saviq (approved):
Aha!

tarek-y-ismail:
@Saviq Looks like the demo compositor we derive the options from fails to run under alpine edge. Is this known?

AlanGriffiths:
> @Saviq Looks like the demo compositor we derive the options from fails to run under alpine edge. Is this known?

I have an alpine VM handy, can reproduce. Investigating...

tarek-y-ismail:
I'm also poking at it with Copilot.

AlanGriffiths:
OK, so the problem is that the output from `.../demo-mir-compositor --help-markdown` contains logging and other cruft.

```diff
diff --git a/doc/CMakeLists.txt b/doc/CMakeLists.txt
index fbb1613a37..00bcd40c5d 100644
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -30,6 +30,7 @@ add_test(
     ${OPTIONS_REFERENCE} \
     <(
       ${CMAKE_BINARY_DIR}/bin/demo-mir-compositor --help-markdown \
+      | grep -v -e mircommon -e LD_LIBRARY_PATH \
       | head -c -1 \
       | sed 's@${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}@\\<LIBDIR\\>@g' \
      )"


```

PR #5091, opened 2026-07-13
Add a touch emulator to map mouse clicks and drags to touch events

tarek-y-ismail:
Closes https://github.com/canonical/mir/pull/5071#discussion_r3559802152

## What's new?

- Adds a `miral::TouchEmulator`. When enabled, it converts mouse clicks and drags to touch events.
  Can be controlled via a keyboard shortcut for example:
  
  ```diff
  +#include <miral/touch_emulator.h>
  +#include <mir/log.h>
   
   #include <xkbcommon/xkbcommon-keysyms.h>
   
  @@ -358,6 +360,34 @@ int main(int argc, char const* argv[])
           return false;
       };
   
  +    auto touch_emulator = TouchEmulator::disabled();
  +    auto toggle_touch_emulator_filter = [&touch_emulator, touch_emulator_enabled = false](MirKeyboardEvent const* key_event) mutable
  +    {
  +        // ...
  +
  +        touch_emulator_enabled = !touch_emulator_enabled;
  +        if (touch_emulator_enabled)
  +        {
  +            touch_emulator.enable();
  +            mir::log_info({mir::logging::input()}, "Touch emulator enabled");
  +        }
  +        else
  +        {
  +            touch_emulator.disable();
  +            mir::log_info({mir::logging::input()}, "Touch emulator disabled");
  +        }
  +
  +        return true;
  +    };
  +
  +
       return runner.run_with(
           {
               CursorTheme{"default:DMZ-White"},
  @@ -392,6 +422,8 @@ int main(int argc, char const* argv[])
               cursor_scale,
               locate_pointer,
               AppendKeyboardEventFilter{locate_pointer_filter},
  -            BasicApplicationSwitcher{}
  +            BasicApplicationSwitcher{},
  +            touch_emulator,
  +            AppendKeyboardEventFilter{toggle_touch_emulator_filter},
           });
   }
  ``` 
  or via live config:
  
  ```diff
   #include <miral/slow_keys.h>
   #include <miral/hover_click.h>
   #include <miral/append_keyboard_event_filter.h>
  +#include <miral/touch_emulator.h>
   
   #include <mir/abnormal_exit.h>
   #include <mir/main_loop.h>
  @@ -140,6 +141,7 @@ try
       miral::StickyKeys sticky_keys{config_store};
       miral::Keymap keymap{config_store};
       miral::HoverClick hover_click{config_store};
  +    miral::TouchEmulator touch_emulator{config_store};
   
       miral::ConfigFile config_file{
           runner,
  @@ -175,6 +177,7 @@ try
           slow_keys,
           sticky_keys,
           hover_click,
  +        touch_emulator,
           keymap,
       });
  ```
  ```ini
  touch_emulator_enable=true
  ``` 

## How to test

- Install [`wev`](https://git.sr.ht/~sircmpwn/wev)
- Run `miral-shell` or `mir_demo_server` (with `--test-client <...> --test-timeout 9999`)
- Press `Ctrl+Alt+Shift+E` on `miral-shell` to toggle the touch emulator on, or modify `~/.config/mir_demo_server.live-config` with the content above.
- Launch a `wev` window
- Observe as clicks and drags are converted to touch events.
- Toggle touch emulation again and observe as motion and click events resume.

### DIFF
--- a/examples/miral-shell/shell_main.cpp
+++ b/examples/miral-shell/shell_main.cpp
@@ -42,6 +42,8 @@
 #include <miral/sticky_keys.h>
 #include <miral/magnifier.h>
 #include <miral/cursor_scale.h>
+#include <miral/touch_emulator.h>
+#include <mir/log.h>
 
 #include <xkbcommon/xkbcommon-keysyms.h>
 
@@ -358,6 +360,28 @@ int main(int argc, char const* argv[])
         return false;
     };
 
+    auto touch_emulator = TouchEmulator::disabled();
+    auto toggle_touch_emulator_filter = [&touch_emulator, touch_emulator_enabled = false](MirKeyboardEvent const* key_event) mutable
+    {
+        auto const modifiers = mir_keyboard_event_modifiers(key_event);
+
+        if (mir_keyboard_event_action(key_event) != mir_keyboard_action_down ||
+            !(modifiers & mir_input_event_modifier_ctrl) ||
+            !(modifiers & mir_input_event_modifier_alt) ||
+            !(modifiers & mir_input_event_modifier_shift) ||
+            mir_keyboard_event_keysym(key_event) != XKB_KEY_E)
+            return false;
+
+        touch_emulator_enabled = !touch_emulator_enabled;
+        if (touch_emulator_enabled)
+            touch_emulator.enable();
+        else
+            touch_emulator.disable();
+
+        return true;
+    };
+
+
     return runner.run_with(
         {
             CursorTheme{"default:DMZ-White"},
@@ -392,6 +416,8 @@ int main(int argc, char const* argv[])
             cursor_scale,
             locate_pointer,
             AppendKeyboardEventFilter{locate_pointer_filter},
-            BasicApplicationSwitcher{}
+            BasicApplicationSwitcher{},
+            touch_emulator,
+            AppendKeyboardEventFilter{toggle_touch_emulator_filter},
         });
 }
--- a/src/miral/touch_emulator.cpp
+++ b/src/miral/touch_emulator.cpp
@@ -1,0 +1,272 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "miral/touch_emulator.h"
+
+#include <miral/live_config.h>
+#include <miral/prepend_event_filter.h>
+
+#include <mir/events/touch_contact.h>
+#include <mir/geometry/point.h>
+#include <mir/input/device_capability.h>
+#include <mir/input/event_builder.h>
+#include <mir/input/input_device_registry.h>
+#include <mir/input/input_sink.h>
+#include <mir/input/virtual_input_device.h>
+#include <mir/main_loop.h>
+#include <mir/options/option.h>
+#include <mir/server.h>
+#include <mir/synchronised.h>
+
+#include <mir_toolkit/events/input/pointer_event.h>
+#include <mir_toolkit/events/input/input_event.h>
+#include <mir_toolkit/events/event.h>
+#include <mir_toolkit/events/enums.h>
+
+#include <chrono>
+
+namespace mi   = mir::input;
+namespace mev  = mir::events;
+namespace geom = mir::geometry;
+
+
+class miral::TouchEmulator::Self
+{
+public:
+    explicit Self(bool enabled) :
+        touch_device{std::make_shared<mi::VirtualInputDevice>("touch-emulator", mi::DeviceCapability::touchscreen)},
+        filter{[this](MirEvent const* event) { return handle(event); }},
+        state{State{enabled}}
+    {
+    }
+
+    // Called during server configuration to register the (initially inactive) event filter.
+    void register_with(mir::Server& server)
+    {
+        filter(server);
+    }
+
+    void init(mir::Server& server)
+    {
+        auto const state_ = state.lock();
+        main_loop = server.the_main_loop();
+        input_device_registry = server.the_input_device_registry();
+
+        if (state_->active)
+        {
+            input_device_registry.lock()->add_device(touch_device);
+            state_->device_registered = true;
+        }
+    }
+
+    void deinit()
+    {
+        auto const state_ = state.lock();
+        state_->active = false;
+        if (state_->device_registered)
+        {
+            input_device_registry.lock()->remove_device(touch_device);
+            state_->device_registered = false;
+        }
+    }
+
+    void activate()
+    {
+        auto const state_ = state.lock();
+        state_->active = true;
+
+        if (!state_->device_registered)
+        {
+            // If the server has not started, init() registers the device.
+            if (auto const registry = input_device_registry.lock())
+            {
+                registry->add_device(touch_device);
+                state_->device_registered = true;
+            }
+        }
+    }
+
+    void deactivate()
+    {
+        state.lock()->active = false;
+    }
+
+    bool handle(MirEvent const* event)
+    {
+        if (!state.lock()->active)
+            return false;
+
+        if (mir_event_get_type(event) != mir_event_type_input)
+            return false;
+
+        auto const* input_ev = mir_event_get_input_event(event);
+        if (mir_input_event_get_type(input_ev) != mir_input_event_type_pointer)
+            return false;
+
+        auto const* pev = mir_input_event_get_pointer_event(input_ev);
+
+        // Ctrl+drag passes through unchanged for normal pointer interaction.
+        if (mir_pointer_event_modifiers(pev) & mir_input_event_modifier_ctrl)
+            return false;
+
+        auto const action  = mir_pointer_event_action(pev);
+        auto const buttons = mir_pointer_event_buttons(pev);
+        float const x = mir_pointer_event_axis_value(pev, mir_pointer_axis_x);
+        float const y = mir_pointer_event_axis_value(pev, mir_pointer_axis_y);
+        auto const ts = std::chrono::nanoseconds{mir_input_event_get_event_time(input_ev)};
+
+        if (!dragging &&
+            action == mir_pointer_action_button_down &&
+            (buttons & mir_pointer_button_primary))
+        {
+            dragging = true;
+            dispatch_touch(mir_touch_action_down, x, y, ts);
+            return true;
+        }
+        else if (action == mir_pointer_action_motion && dragging)
+        {
+            dispatch_touch( mir_touch_action_change, x, y, ts);
+            return true;
+        }
+        else if (action == mir_pointer_action_button_up && dragging &&
+                 !(buttons & mir_pointer_button_primary))
+        {
+            dragging = false;
+            dispatch_touch(mir_touch_action_up, x, y, ts);
+            return true;
+        }
+
+        // Consume all other pointer events so they don't reach surfaces.
+        return true;
+    }
+
+private:
+    struct State
+    {
+        explicit State(bool enabled) :
+            active{enabled}
+        {
+        }
+
+        bool active;
+        bool device_registered{false};
+    };
+
+    void dispatch_touch(
+        MirTouchAction action,
+        float x,
+        float y,
+        std::chrono::nanoseconds timestamp)
+    {
+        auto const ml = main_loop.lock();
+        if (!ml)
+            return;
+
+        // Enqueue the dispatch so it runs after the current filter invocation
+        // returns.  Calling sink->handle_input() from inside the filter would
+        // re-enter the composite event filter while it is already executing,
+        // causing a deadlock.
+        ml->enqueue(
+            this,
+            [touch_device=touch_device, action, x, y, timestamp]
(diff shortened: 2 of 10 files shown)


src/miral/runner.cpp
```diff
@@ -151,6 +152,7 @@ try
         });
 
         add_input_device_configuration_options_to(*server);
+        add_touch_emulator_option_to(*server);
```
AlanGriffiths: Is this really an option every compositor needs?

AlanGriffiths (changes requested):
The description says "development and testing aid", which seems more like an option that may be added than part of the base level functionality. But this option is being added to all compositors.

It also seems like something that might usefully be toggled during execution (e.g. when testing an app under a Mir based shell). That would imply supporting "live config".

src/miral/runner.cpp
mattkae: Agreed! Maybe this is exposed in miral instead and provided to `runner.run_with`?

And perhaps it also has a class + state so that you can do `touch_emulator.toggle(bool)` + live config?

mattkae (commented):
I largely agree with Copilot + Alan's feedback, but I like that this change is nice and small and easy to reason about!

Making this optional for compositors would be a good first step, especially since some compositors may not care to work with touch.

tarek-y-ismail:
Not sure why the symbols check complains that the symbols I added are missing

tests/miral/test_touch_emulator.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+using namespace testing;
+using namespace std::chrono_literals;
+
+namespace mi = mir::input;
+namespace geom = mir::geometry;
+
+namespace
+{
+static constexpr auto touch_event_timeout = std::chrono::seconds{5};
```
AlanGriffiths: This doesn't need to be both in the anonymous namespace and static. But why is it at namespace scope anyway? It is only used once from inside the `TestTouchEmulator` fixture (it could be a `static` data member).

tests/miral/test_touch_emulator.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+auto capture_touch(MirEvent const* event) -> CapturedTouch
+{
+    auto const* input_ev = mir_event_get_input_event(event);
+    auto const* tev = mir_input_event_get_touch_event(input_ev);
+    return {
+        mir_touch_event_action(tev, 0),
+        mir_touch_event_axis_value(tev, 0, mir_touch_axis_x),
+        mir_touch_event_axis_value(tev, 0, mir_touch_axis_y)};
+}
```
AlanGriffiths: Could also be scoped to the fixture

tests/miral/test_touch_emulator.cpp
```diff
@@ -0,0 +1,380 @@
...
+
+                touch_emulator(server);
+                observer(server);
+            });
+    }
+
+    miral::TouchEmulator touch_emulator{miral::TouchEmulator::disabled()};
+
+    std::mutex events_mutex;
+    std::vector<CapturedTouch> touch_events;
```
AlanGriffiths: Nit (as everything is scoped to the file): `touch_events` is public, hence there's no guarantee the mutex is used when accessing it.

tests/miral/test_touch_emulator.cpp
```diff
@@ -0,0 +1,380 @@
...
+    pointer3->if_started_then(
+        [this](mi::InputSink* sink, mi::EventBuilder* builder)
+        {
+            send_pointer_event(
+                sink, builder,
+                mir_pointer_action_button_down,
+                MirPointerButtons{mir_pointer_button_primary},
+                50.0f, 60.0f);
+        });
+    ASSERT_TRUE(wait_for_touch_count(1));
```
AlanGriffiths:
```
2026-07-13T18:01:48.9206040Z ../../../tests/miral/test_touch_emulator.cpp:373: Failure
2026-07-13T18:01:48.9206374Z Value of: wait_for_touch_count(1)
2026-07-13T18:01:48.9206637Z Actual: false
2026-07-13T18:01:48.9206837Z Expected: true
2026-07-13T18:01:48.9206955Z 
2026-07-13T18:01:48.9207591Z [2026-07-13 17:59:49.012929] < -warning- > security: {"datetime": "2026-07-13T17:59:49.012827582Z", "appid": "miral-test-internal.bin", "event": "sys_shutdown", "level": "WARN", "description": "Mir is shutting down" }
2026-07-13T18:01:48.9208536Z [ FAILED ] TestTouchEmulator.disable_and_re_enable_toggles_behavior (6357 ms)
```

AlanGriffiths (changes requested):
Oh! I found these pending comments. They still seem relevant

src/miral/touch_emulator.cpp
```diff
@@ -0,0 +1,238 @@
...
+    {
+        main_loop = server.the_main_loop();
+        input_device_registry = server.the_input_device_registry();
+    }
+
+    void activate()
+    {
+        if(!active())
+            input_device_registry.lock()->add_device(touch_device);
+        active_.store(true);
```
AlanGriffiths: This is racy. Several threads could call `add_device()` before any of them set `active_`:
```suggestion
        auto activated = false;

        if (state.compare_exchange_strong(activated, true))
            input_device_registry.lock()->add_device(touch_device);
```

src/miral/touch_emulator.cpp
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    void deactivate()
+    {
+        if (active())
+            input_device_registry.lock()->remove_device(touch_device);
+        active_.store(false);
```
AlanGriffiths:
```suggestion
        auto activated = true;

        if (state.compare_exchange_strong(activated, false))
            input_device_registry.lock()->remove_device(touch_device);
```

AlanGriffiths (changes requested):
The description should include instructions for using a compositor that supports this feature. It could be added (disabled by default) to `miral-shell` or `miral-kiosk`?

src/miral/touch_emulator.cpp
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+
+    std::shared_ptr<mi::VirtualInputDevice> const touch_device;
+    miral::PrependEventFilter filter;
+    std::weak_ptr<mir::MainLoop> main_loop;
+    std::weak_ptr< mir::input::InputDeviceRegistry > input_device_registry;
+
+    bool dragging{false};
+    mir::Synchronised<bool> active_{false};
```
AlanGriffiths: It looks weird to have a bunch of mutable state, but only a single `bool` is protected by `mir::Synchronised`. Not only is `mir::Synchronised` overkill for a `bool` but is it correct that nothing else needs guarding? Should there be an explanation why?

tarek-y-ismail: - `filter` is only non-`const` because its parens operator is non-`const` (which may be something we might want to change`.
- `main_loop` and `input_device_registry` are both set at server start during the single-threaded init and never modified after.
- `dragging` is only set in the input thread, no need for synchronization
- `active_` could be accessed from the live config and input threads simultaneously for example, so that needs protection.

As for the use of `mir::Synchronised`, we can't use an atomic as we also should account for other threads calling `activate/deactivate`, which will synchronize `active_` correctly, but not the addition and removal of input devices (see 8aaa1cd's message for an example) as well. A mutex is the same as `mir::Synchronised` for this usage at least. But I prefer `mir::Synchronized` whenever possible.

src/miral/touch_emulator.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+namespace mi   = mir::input;
+namespace mev  = mir::events;
+namespace geom = mir::geometry;
+
+
+class miral::TouchEmulator::Self
+{
+public:
+    Self(bool enabled) :
```
mattkae:
```suggestion
    explicit Self(bool enabled) :
```

src/miral/touch_emulator.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+
+    std::shared_ptr<mi::VirtualInputDevice> const touch_device;
+    miral::PrependEventFilter filter;
+    std::weak_ptr<mir::MainLoop> main_loop;
+    std::weak_ptr<mir::input::InputDeviceRegistry> input_device_registry;
+
+    bool dragging{false};
+    mir::Synchronised<bool> active_{false};
```
mattkae: Is `std::atomic` a better candidate here? Synchronised is moreso used for instances where we want to synchronize a whole set of data.

```suggestion
    std::atomic<bool> active_{false};
```

tools/symbols_map_generator/main.py
```diff
@@ -61,6 +61,11 @@
     "miral::MouseKeysConfig::disabled*;",
     "miral::MouseKeysConfig::enable*;",
     "miral::MouseKeysConfig::enabled*;",
+    "miral::TouchEmulator::disable*;",
+    "miral::TouchEmulator::disabled*;",
+    "miral::TouchEmulator::enable*;",
+    "miral::TouchEmulator::enabled*;",
+    "miral::TouchEmulator::operator*;",
 }
```
mattkae, line 69: Why do we need to ignore these?

mattkae: I would leave a comment for the group with an explanation personally, or make it so we don't have to

mattkae (changes requested):
Some minor things, but this is much nice 🎉

src/miral/touch_emulator.cpp
tarek-y-ismail: See https://github.com/canonical/mir/pull/5091#discussion_r3613532891

tools/symbols_map_generator/main.py
tarek-y-ismail: Exporting `enable* or `disable* would also match `enabled` or `disabled` and export them. `operator*` would match any operators we'd add, though it's quite unlikely we'd add anything other than the function call operator to a public class anyway. I wanted to be a bit more explicit about the symbols being exported to avoid issues we've had in the past. Unfortunately, the symbols map generator can't match against quoted literals directly at the moment. So we have to ignore them. I'm honestly not 100% sure which way to go.

src/miral/touch_emulator.cpp
AlanGriffiths: > As for the use of `mir::Synchronised`, we can't use an atomic as we also should account for other threads calling `activate/deactivate`, which will synchronize `active_` correctly, but not the addition and removal of input devices (see [8aaa1cd](https://github.com/canonical/mir/commit/8aaa1cd2c8bcccb98e767afee8c76056949f9fff)'s message for an example) as well. A mutex is the same as `mir::Synchronised` for this usage at least. But I prefer `mir::Synchronized` whenever possible.

`std::atomic` works if used correctly. C.f. https://github.com/canonical/mir/pull/5091#discussion_r3601922440

tarek-y-ismail: How would that protect against this scenario?
    - Thread A calls `activate()`, the atomic exchange occurs but the
      thread is preempted before it could add the device by thread B.
    - Thread B calls `deactivate()`, the atomic exchange passes.
      It goes onto remove the device, but since the device wasn't added
      on thread A, `remove_device` throws `std::logic_error`.

AlanGriffiths: Oh, that's what I missed: `remove_device()` isn't idempotent. Sorry for the noise.

Thinking about that raises another point: do we actually want to add/remove the device? (As opposed to adding it once and {dis,en}abling it?)

tarek-y-ismail: Come to think of it, no. We can just leave it registered, I believe. Let me see how this simplifies the code.

tarek-y-ismail: Ugh, atomics are still not enough. They solve the case where one thread is enabling and the other is disabling, but not the case where the two threads are enabling or disabling. So we really need locks.

tools/symbols_map_generator/main.py
mattkae: Ugh right right. I always forget that :)

src/miral/touch_emulator.cpp
mattkae: Ah I see why, and I think your usage of `Synchronised` is misleading. 

`Synchronised` makes me think that _just_ the data underneath the hood needs to be synchronised, and nothing else.

A `std::mutex` would be more expressive here because you _want_ to hold the guard lock for the duration of the call, instead of just guarding a piece of data. I would change this to a global `std::mutex activation_mut;`

tests/miral/test_touch_emulator.cpp
```diff
@@ -0,0 +1,440 @@
...
+struct TestTouchEmulator : miral::TestServer
+{
+    struct CapturedTouch
+    {
+        auto static to_touch(MirEvent const* event) -> MirTouchEvent const*
+        {
+            return mir_input_event_get_touch_event(mir_event_get_input_event(event));
+        }
+
+        CapturedTouch(MirEvent const* event) :
```
mattkae:
```suggestion
        explicit CapturedTouch(MirEvent const* event) :
```

mattkae (approved):
Nice, I like it 🎉

examples/miral-shell/shell_main.cpp
```diff
@@ -358,6 +360,34 @@ int main(int argc, char const* argv[])
...
+            !(modifiers & mir_input_event_modifier_alt) ||
+            !(modifiers & mir_input_event_modifier_shift) ||
+            mir_keyboard_event_keysym(key_event) != XKB_KEY_E)
+            return false;
+
+        touch_emulator_enabled = !touch_emulator_enabled;
+        if (touch_emulator_enabled)
+        {
+            touch_emulator.enable();
+            mir::log_info({mir::logging::input()}, "Touch emulator enabled");
```
AlanGriffiths: This looks more like debug code than example (or "info") code: do need it?

AlanGriffiths (approved):
Optional nit to clean up

examples/miral-shell/shell_main.cpp
tarek-y-ismail: Busted!

PR #5098, opened 2026-07-15
Better logging::Logger ABI

RAOF:
Consolidating the log event data into a single object will allow us to evolve the set of data we provide.

Notably, we may want to introduce a `Span` concept; this will let us do so in an API- & ABI-preserving way. This also should make integrating a Rust-side `tracing::Subscriber` easier.


tests/acceptance-tests/test_command_line_handling.cpp
```diff
@@ -205,26 +204,27 @@ TEST_F(CommandLineHandling, subsequent_log_level_overrides_previous)
 
 
     std::string message = "hello";
-    EXPECT_CALL(*logger, log(ml::Severity::debug, message, "test_tag")).Times(1);
-    EXPECT_CALL(*logger, log(_, _, "base")).Times(0);
+    EXPECT_CALL(*logger, log(LogMatching(Eq(ml::Severity::debug), Eq(message), ElementsAre(IsTag(tag))))).Times(1);
+    EXPECT_CALL(*logger, log(LogMatching(_, _, ElementsAre(IsTag(ml::base()))))).Times(0);
 
     mir::log_debug({tag}, message);
     mir::log_error({ml::base()}, message);
```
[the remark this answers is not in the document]
RAOF: This is incorrect; C++ is perfectly happy to materialise a std::span from an initialiser-list.

src/core/logging/event.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+class ml::Event::Impl::ImplWithAllocatedMessage : public Impl
+{
+public:
+    ImplWithAllocatedMessage(
+        Severity sev,
+        std::unique_ptr<char const[]> message,
+        std::source_location loc)
+        : Impl(sev, {uncategorised_tag}, std::string_view{message.get()}, loc),
+          message_storage{std::move(message)}
```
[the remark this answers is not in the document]
RAOF: Yup, fixed.

tests/unit-tests/logging/test_logging.cpp
```diff
@@ -28,70 +28,73 @@
 #include <string_view>
 #include <cstring>
```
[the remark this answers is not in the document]
RAOF: Nope, `{ml::base()}` is a perfectly cromulent initialiser-list.

RAOF:
There will be a couple of follow-ups plumbing through `std::source_location` our various `mir::log_foo`s and looking at optimising `Event` to be zero-allocation, but that's an extension of what's here.

include/core/mir/logging/event.h
```diff
@@ -0,0 +1,81 @@
...
+        Severity sev,
+        Tags tags,
+        std::string_view message,
+        std::source_location location = std::source_location::current());
+
+    /**
+     * Transitional Event constructor for call-sites that have not yet migrated to Tags
+     *
+     * These Events are all categorised in a catch-all "uncategorised" tag;
+     * the `category` included here is added to the message
```
tarek-y-ismail, line 53: What category?

src/common/logging/file_logger.cpp
```diff
@@ -25,13 +27,18 @@ ml::FileLogger::FileLogger(std::ofstream stream)
...
     {
-        format_message(file, severity, message, component);
+        format_message(
+            file,
+            log_event.severity(),
+            log_event.message(),
+            log_event.tags() |
+                std::views::transform([](Tag const& tag) { return tag::name(tag); }) |
+                std::views::join_with(':') |
+                std::ranges::to<std::string>());
```
tarek-y-ismail, line 41: Maybe we should supply a utility fn?

src/common/logging/shared_library_prober_report.cpp
tarek-y-ismail, line 1: Good opportunity to convert the string concats into `std::format` calls. Though that's somewhat out of scope. A TODO comment would be nice.

src/core/symbols.map
```diff
@@ -9,16 +9,12 @@ MIR_CORE_2.29 {
...
-    /* mir::log(mir::logging::Severity, std::initializer_list<std::reference_wrapper<mir::logging::Tag const> const>, std::basic_string_view<char, std::char_traits<char> >); */
-    _ZN3mir3logENS_7logging8SeverityESt16initializer_listIKSt17reference_wrapperIKNS0_3TagEEESt17basic_string_viewIcSt11char_traitsIcEE;
-    /* mir::logging::log(mir::logging::Severity, std::initializer_list<std::reference_wrapper<mir::logging::Tag const> const>, std::basic_string_view<char, std::char_traits<char> >); */
-    _ZN3mir7logging3logENS0_8SeverityESt16initializer_listIKSt17reference_wrapperIKNS0_3TagEEESt17basic_string_viewIcSt11char_traitsIcEE;
-    /* mir::logging::Logger::log(mir::logging::Severity, std::initializer_list<std::reference_wrapper<mir::logging::Tag const> const>, std::basic_string_view<char, std::char_traits<char> >); */
-    _ZN3mir7logging6Logger3logENS0_8SeverityESt16initializer_listIKSt17reference_wrapperIKNS0_3TagEEESt17basic_string_viewIcSt11char_traitsIcEE;
+    /* mir::log(mir::logging::Severity, std::span<std::reference_wrapper<mir::logging::Tag const> const>, std::basic_string_view<char, std::char_traits<char> >); */
```
tarek-y-ismail, line 12: Might be a little easier to use "normal" names.

src/server/report/logging/display_report.cpp
```diff
@@ -73,34 +73,34 @@ void mrl::DisplayReport::report_drm_master_failure(int error)
     if (error == EPERM || error == EACCES)
         ss << " Try running Mir with root privileges.";
```
tarek-y-ismail, line 74: A comment to come back and clean this in the future would be nice.

src/server/report/logging/display_report.cpp
```diff
@@ -150,13 +150,13 @@ void mrl::DisplayReport::report_egl_configuration(EGLDisplay disp, EGLConfig con
...
-    logger->log(ml::Severity::informational, "Display EGL Configuration:", component());
+    logger->log(ml::Event{ml::Severity::informational, component(), "Display EGL Configuration:"});
     for( auto &i : egl_string_mapping)
     {
         EGLint value;
         if (eglGetConfigAttrib(disp, config, i.val, &value))
-            logger->log(ml::Severity::informational,
-                "    [" + i.name + "] : " + std::to_string(value), component());
+            logger->log(ml::Event{ml::Severity::informational,
+                component(), "    [" + i.name + "] : " + std::to_string(value)});
```
tarek-y-ismail, line 159: Another instance of string concatenation!

src/server/report/logging/scene_report.cpp
```diff
@@ -41,7 +41,7 @@ void mrl::SceneReport::surface_created(BasicSurfaceId id, std::string const& nam
     std::stringstream ss;
     ss << "surface_created(" << id << " [\"" << name << "\"])";
```
tarek-y-ismail, line 42: TODO!

src/server/report/logging/seat_report.cpp
tarek-y-ismail, line 1: Another TODO!

src/server/report/logging/shell_report.cpp
tarek-y-ismail, line 1: More `stringstream` and string concatenation

tests/miral/static_display_config.cpp
```diff
@@ -48,9 +50,15 @@ static const std::vector<uint8_t> basic_edid{
     0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x93,
 };
 
+MATCHER_P2(LogMatching, severity_matcher, message_matcher, "")
+{
+    return ExplainMatchResult(severity_matcher, arg.severity(), result_listener) &&
+           ExplainMatchResult(message_matcher, std::string{arg.message()}, result_listener);
+}
+
```
tarek-y-ismail, line 58: I recognize this from before? Can't we share impls?

tests/miral/static_display_config.cpp
```diff
@@ -99,6 +107,8 @@ struct StaticDisplayConfig : Test
             InvokeArgument<0>(mg::UserDisplayConfigurationOutput{vga1}),
             InvokeArgument<0>(mg::UserDisplayConfigurationOutput{hdmi1})));
 
+        // Ensure that our logger actually receives messages
+        mir::logging::tag::set_severity("base", ml::Severity::debug);
```
tarek-y-ismail, line 111: Since we're moving away from strings, should this be updated to accept a tag reference?

tests/unit-tests/logging/test_display_report.cpp
tarek-y-ismail, line 1: Another file we can replace concats in.

src/core/logging/event.cpp
```diff
@@ -0,0 +1,143 @@
...
+{
+auto create_message_string(std::string_view component, std::string_view message) -> std::unique_ptr<char const[]>
+{
+    auto const combined_message = std::views::concat(component, std::string_view{": "}, message);
+    auto storage = std::make_unique<char[]>(combined_message.size() + 1);
+
+    std::ranges::copy(combined_message, storage.get());
+
+    return storage;
+}
```
tarek-y-ismail: Maybe we could replace this and `message_storage`  with a 
```
message_storage{std::format("{}: {}", component, message)}
```
?

tarek-y-ismail (commented):
No major objections so far. A couple of questions and a suggestion for `ImplWithAllocatedMessage`. Most other comments are just files I think we should mark for a separate cleanup PR.

include/core/mir/logging/event.h
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * \warning The `tags` and `message` parameters are non-owning views; an
+     * Event should almost always be directly constructed during a function call
+     * to Logger::log().
+     */
+    Event(
+        Severity sev,
+        Tags tags,
+        std::string_view message,
+        std::source_location location = std::source_location::current());
```
jhenstridge, line 47: It'd be nice to build `std::format` functionality into the Event class such that we only try to format the message if a backend tries to log it: collect the args with `std::make_format_args`, and then only call `std::vformat` if the logger actually wants to record the message.

The only real snag for this is the `std::source_location` default argument preventing the use of an argument pack here. One possibility would be to move the argument to the `Logger::log(Event const&)` method. Might this be worth while?

src/core/logging/event.cpp
```diff
@@ -0,0 +1,143 @@
...
+{
+    return impl->location();
+}
+
+auto ml::Event::should_log() const -> bool
+{
+    auto const sev = impl->severity();
+    return
+        std::ranges::any_of(impl->tags(), [sev](Tag const& tag) { return ml::logging_enabled_for(tag, sev); });
+}
```
jhenstridge, line 183: I know that this fits our current logging model, but it kind of feels like the decision to log is the responsibility of the logger rather than the event. If a MultiLogger is in use, I could imagine each sub-logger wanting to make a different decision.

src/common/logging/shared_library_prober_report.cpp
jhenstridge, line 1: It probably doesn't make a huge amount of difference here, but it might make sense to have helpers like these take a `std:source_location` default argument so we can attribute the message to the caller rather than the helper.

include/core/mir/logging/logger.h
```diff
@@ -47,29 +49,29 @@ class Logger
...
-     * Cruft removal: when we're sure we've got the right API/ABI, we should
-     * *remove* log(severity, string, string) above and make this the
-     * required-to-implement interface.
-     */
-    void log(Severity severity, Tags tags, std::string_view message);
-
     template<typename... Args>
     void log(Severity severity, Tags tags, std::format_string<Args...> fmt, Args&&... args)
-    { log(severity, tags, std::format(fmt, std::forward<Args>(args)...)); }
+    { log(Event{severity, tags, std::format(fmt, std::forward<Args>(args)...)}); }
```
jhenstridge, line 54: Log messages output through this overload will all be attributed to this source line, since that's where the `Event` is created. I'm not sure there is an easy way around this while keeping this function signature.

If we did have format() capabilities in Event itself, perhaps we could do without this overload entirely? That would end up touching a lot more logging call sites though.

include/core/mir/logging/event.h
AlanGriffiths: Yes, this was one of the original motivations for reworking the logging APIs.

Note that we typically want to update call sites that use a format string, so we are not trying to maintain the exact syntax there. We could do stuff like:

    mir::log(tags/*, std::source_location::current()*/).debug(formatter, ...);

Where the default parameter and parameter pack are used in different parameter lists.

include/test/mir/test/doubles/null_logger.h
```diff
@@ -17,6 +17,7 @@
 #ifndef MIR_TEST_DOUBLES_NULL_LOGGER_H_
 #define MIR_TEST_DOUBLES_NULL_LOGGER_H_
 
+#include "mir/logging/event.h"
```
AlanGriffiths, line 20:
```suggestion
#include <mir/logging/event.h>
```

include/core/mir/logging/event.h
```diff
@@ -0,0 +1,81 @@
...
+
+    Event(Event const&) = delete;
+    Event(Event&&) = delete;
+    auto operator=(Event const&) -> Event& = delete;
+    auto operator=(Event&&) -> Event& = delete;
+
+private:
+    class Impl;
+    // Impl ptr has explicit deleter to allow us to do memory shenanagins for optimisation.
+    std::unique_ptr<Impl, void (*)(Impl*)> const impl;
```
AlanGriffiths: Is there any value in `unique_ptr` here? If we're doing memory shenanagins, then just store an `Impl* const` and put any shenanagins in the ctor/dtor

include/core/mir/logging/event.h
AlanGriffiths: Or

```c++
    mir::log_debug(tags/*, std::source_location::current()*/)(formatter, ...);
```

AlanGriffiths (dismissed):
I have the same doubts about the API as @jhenstridge.

I think the question that needs answering first is: how do the call sites change to use formatters? And can we make collecting the source location and tags painless?

My initial thought is that we'd want to support something like:

```c++
    mir::log_info(tags, "Just some text"); // Simple text, no formatting, default source location

    mir::log_error(tags)("Expected 'foo', got '{}'", "bar"); // default source location, formatter and pack
```

The latter is implementable by creating a functor from `mir::log_error(Tags, source_location)` et al that stores the level, tags and location. (Clearly not the only option, but seems fairly clean)

src/core/logging/event.cpp
RAOF: Yeah, I agree. I decided to follow the existing design here for the sake of simplicity, but if we want to revisit that design now would be the time.

include/core/mir/logging/event.h
RAOF: > It'd be nice to build `std::format` functionality into the Event class such that we only try to format the message if a backend tries to log it: collect the args with `std::make_format_args`, and then only call `std::vformat` if the logger actually wants to record the message.
> 
> The only real snag for this is the `std::source_location` default argument preventing the use of an argument pack here. One possibility would be to move the argument to the `Logger::log(Event const&)` method. Might this be worth while?

Hm, that might be nice. The other thing we might miss if we type-erase here is compile-time format string validation, but it might be possible to retain that?

Note that this is *not* the user-visible API; all of our logging actually goes through the `mir::log` family of APIs, which we *could* (but the [current PR doesn't](https://github.com/canonical/mir/pull/5104)) pre-filter

RAOF:
> I have the same doubts about the API as @jhenstridge.
> 
> I think the question that needs answering first is: how do the call sites change to use formatters? And can we make collecting the source location and tags painless?
> 

#5104 is an answer to that (namely: the call sites *don't* need to change at all, other than to use format strings; the existing API is sufficient). We *might* additionally want a `MIR_LOG_COMPONENT` equivalent, to have a default set of tags for a file.

include/core/mir/logging/event.h
jhenstridge: As I understand it, the intended use of the API is that you'd have a function with a signature like:

```
template<class... Args> void log(std::format_string<Args...> fmt, Args&&... args);
```

You then use `fmt.get()` to convert the format string to a `std::string_view`, and `std::make_format_args(args)` to convert the arguments to `std::format_args`. Those two type erased values can then be passed to `std::vformat`. You still get all the compile time checking of arguments at the `log` call site.

AlanGriffiths: Agreed. We want the arguments checked at the call site, but the formatting deferred to use. These (intermediate) APIs do not support this.

RAOF: Huh! I wasn't aware that special casing of `std::format_string` arguments was the way the compile-time check was implemented. Yeah, that makes sense now.

AlanGriffiths:
@RAOF are you planning to placate the compiler?

RAOF:
I am indeed; I'll switch back to this tomorrow.

RAOF:
This is superceded by #5260

PR #5099, opened 2026-07-15
Workaround for PreCommit linting error

AlanGriffiths:
Closes #5097

PR #5100, opened 2026-07-15
fatal_error

AlanGriffiths:
Simplify the "fatal_error()" design

A follow-up will switch to C++ formatters (and grab std::source_location - that is awkward right now since varargs and defaulted std::source_location don’t play nice)

### DIFF
--- a/src/core/symbols.map
+++ b/src/core/symbols.map
@@ -27,8 +27,6 @@ MIR_CORE_2.29 {
     mir::Fd::invalid*;
     mir::Fd::operator*;
     mir::fatal_error*;
-    mir::fatal_error_abort*;
-    mir::fatal_error_except*;
     mir::geometry::Rectangles::Rectangles*;
     mir::geometry::Rectangles::add*;
     mir::geometry::Rectangles::begin*;
--- a/src/miral/live_config_overrides_list.cpp
+++ b/src/miral/live_config_overrides_list.cpp
@@ -52,7 +52,7 @@ void mlc::OverridesList::for_each(Loader unchanged, Loader fresh, Loader modifie
         catch (std::exception const& e)
         {
             mir::fatal_error(
-                "Openers must not throw exceptions. Failed to open '{}' with exception '{}'",
+                "Openers must not throw exceptions. Failed to open '%s' with exception '%s'",
                 event.path.string().c_str(),
                 e.what());
         }
--- a/src/server/shell/basic_idle_handler.cpp
+++ b/src/server/shell/basic_idle_handler.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "basic_idle_handler.h"
+
 #include <mir/fatal.h>
 #include <mir/scene/idle_hub.h>
 #include <mir/scene/session_lock.h>
@@ -23,6 +24,7 @@
 #include <mir/input/scene.h>
 #include <mir/shell/display_configuration_controller.h>
 
+#include <cinttypes>
 #include <mutex>
 
 namespace ms = mir::scene;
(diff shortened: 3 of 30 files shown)


.github/workflows/symbols-check.yml
```diff
@@ -32,7 +32,7 @@ jobs:
   CheckSymbols:
     needs: [check-paths]
     if: needs.check-paths.outputs.should-run == 'true'
-    runs-on: ubuntu-26.04
+    runs-on: ubuntu-latest
```
jhenstridge: This seems to be a partial revert of 15ed1ee7747 (merged back in June).

Looking at the workflow run, `ubuntu-latest` currently results in the job running on an `ubuntu-24.04` image, which doesn't seem desirable here. In fact, it seems to be causing the workflow to fail because 24.04's compiler doesn't support C++26.

The commit message just says "Fix actions", so I'm not sure what exactly you're trying to fix here.

tarek-y-ismail: See #5099

src/core/symbols.map
tarek-y-ismail, line 1: Why are we adding new symbols?

src/miral/live_config_overrides_list.cpp
tarek-y-ismail, line 1: Maybe it's worth transitioning to brace style formatting, but that's for another PR.

src/server/shell/basic_idle_handler.cpp
```diff
@@ -286,7 +288,7 @@ void msh::BasicIdleHandler::register_observers(ProofOfMutexLock const&)
     {
         if (*off_timeout <= time::Duration{0})
         {
-            fatal_error("BasicIdleHandler given invalid timeout %d, should be >0", off_timeout->count());
+            fatal_error("BasicIdleHandler given invalid timeout %" PRId64 ", should be >0", static_cast<int64_t>(off_timeout->count()));
```
tarek-y-ismail, line 291: Can use `ld` (this is what the macro expands to), but this is fine. Do we need the portability here though?

tests/unit-tests/test_fatal.cpp
```diff
@@ -22,38 +22,9 @@
...
 }
-
-TEST(FatalErrorDeathTest, abort_raises_sigabrt)
-{
-    mir::FatalErrorStrategy on_error{mir::fatal_error_abort};
-
-    MIR_EXPECT_EXIT(mir::fatal_error("Hello world"),
-                    KilledBySignal(SIGABRT),
-                    "Hello world");
-}
```
tarek-y-ismail, line 41: I think this should stay.

tarek-y-ismail (changes requested):

src/core/symbols.map
AlanGriffiths: That looks like a rebase error that slipped past me

src/server/shell/basic_idle_handler.cpp
AlanGriffiths: > this is what the macro expands to

Only on some platforms

AlanGriffiths: FWIW this goes in the follow-up anyway

tests/unit-tests/test_fatal.cpp
AlanGriffiths: But the function it tests has gone

src/server/shell/basic_idle_handler.cpp
tarek-y-ismail: Yeah it's not a blocker, just pointing it out.

tests/unit-tests/test_fatal.cpp
tarek-y-ismail: It was renamed, not completely removed. If we were to follow this logic, the test right above it would also be removed.

AlanGriffiths: OK, I was thinking it was already covered by `fatal_error_formats_message_to_stderr` but it is a little different...

tarek-y-ismail (approved):

PR #5102, opened 2026-07-17
ci: add ubuntu-26.04 as a custom runner label so actionlint stops complaining

### LINKED ISSUE #5097: PreCommit hook failing since 2026-07-14
```
[19](https://github.com/canonical/mir/actions/runs/29344991401/job/87126227206#step:4:124)

.github/workflows/spread.yml:78:14: label "ubuntu-26.04" is unknown. available labels are "windows-latest", "windows-latest-8-cores", "windows-2025", "windows-2025-vs2026", "windows-2022", "windows-11-arm", "ubuntu-slim", "ubuntu-latest", "ubuntu-latest-4-cores", "ubuntu-latest-8-cores", "ubuntu-latest-16-cores", "ubuntu-24.04", "ubuntu-24.04-arm", "ubuntu-22.04", "ubuntu-22.04-arm", "macos-latest", "macos-latest-xlarge", "macos-latest-large", "macos-26-intel", "macos-26-xlarge", "macos-26-large", "macos-26", "macos-15-intel", "macos-15-xlarge", "macos-15-large", "macos-15", "macos-14-xlarge", "macos-14-large", "macos-14", "self-hosted", "x64", "arm", "arm64", "linux", "macos", "windows", "self-hosted-linux-amd64-resolute-large-tiobe". if it is a custom label for self-hosted runner, set list of labels in actionlint.yaml config file [runner-label]
```

It looks like "ubuntu-26.04" came with #5028 - which landed last month, so: what changed?

jhenstridge:
This is a slightly cleaner version of #5099. The underlying problem is that [actionlint v1.7.12](https://github.com/rhysd/actionlint) includes a hard coded list of runner images it believes that Github provides, and ubuntu-26.04 is not on that list. Ideally we could move to a newer version in our pre-commit configuration, but no newer versions have been released (there is https://github.com/rhysd/actionlint/pull/683 open to add it though).

We can augment the list of runner images actionlint will treat as valid through it's configuration file though, so that's what this PR does. This has the benefit that we'll be linting the `spread.yml` and `symbols-check.yml` workflows again, and no change will be necessary if we start using ubuntu-26.04 in other workflows.

Closes #5097

## How to test

Check if the PreCommit actions workflow succeeds.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,3 +1,6 @@
 self-hosted-runner:
   labels:
     - self-hosted-linux-amd64-resolute-large-tiobe
+    # actionlint v1.7.12 doesn't know about ubuntu-26.04:
+    #    https://github.com/rhysd/actionlint/pull/683
+    - ubuntu-26.04
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -26,9 +26,6 @@ jobs:
       with:
         python-version: 3.x
     - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
-      if: always()
-      with:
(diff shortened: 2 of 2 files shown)


tarek-y-ismail (approved):
Seems to blend

PR #5104, opened 2026-07-17
Plumb std::source_location through the `mir::log{,_debug,_info,_warning,_error,_critical}` paths.

RAOF:
This means that all calls to `mir::log`, `mir::log_debug`, `mir::log_info`, etc will result in an `Event` with the correct `std::source_location` being passed to `Logger::log()`.

### DIFF
--- a/debian/libmircore3.symbols
+++ b/debian/libmircore3.symbols
@@ -26,11 +26,9 @@ libmircore.so.3 libmircore3 #MINVER#
  (c++)"mir::input::MouseKeysKeymap::for_each_key_action_pair(std::function<void (unsigned int, mir::input::MouseKeysKeymap::Action)>&&) const@MIR_CORE_2.29" 2.29.0
  (c++)"mir::input::MouseKeysKeymap::get_action(unsigned int) const@MIR_CORE_2.29" 2.29.0
  (c++)"mir::input::MouseKeysKeymap::set_action(unsigned int, std::optional<mir::input::MouseKeysKeymap::Action>)@MIR_CORE_2.29" 2.29.0
- (c++)"mir::log(mir::logging::Severity, char const*, char const*, ...)@MIR_CORE_2.29" 2.29.0
- (c++)"mir::log(mir::logging::Severity, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)@MIR_CORE_2.29" 2.29.0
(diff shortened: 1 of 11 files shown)


jhenstridge (commented):
The argument deduction guide approach looks like a good approach. Following on from discussions in #5098,

1. Do we want to base the API on `std::format` instead of printf and va_args? This is a compatibility break, but the compile time format string checks would catch incompatible call sites. In this case, the `logv` signature might be:
    ```
    void logv(
        logging::Severity sev,
        logging::Tags tags,
        std::string_view fmt,
        std::format_args args,
        std::source_location loc);
    ```
    It isn't worth providing a default value for `loc` if we expect every caller to be a wrapper.
2. Would it make sense to make these `log`, `log_debug`, etc structs header-only, and forwarding to the `logv` function? That would simplify the ABI, and give us more leeway to fix these templates if we've missed an edge case somewhere.

AlanGriffiths:
```
2026-08-13T08:00:42.3393772Z cd /build/reproducible-path/mir-2.29.0/build-armhf/src/platform && /usr/bin/cmake -E cmake_link_script CMakeFiles/mirplatform.dir/link.txt --verbose=1
2026-08-13T08:00:42.3395409Z /usr/bin/arm-linux-gnueabihf-ld.bfd: options/CMakeFiles/miroptions.dir/default_configuration.cpp.o: in function `mir::log<mir::logging::Severity, std::span<std::reference_wrapper<mir::logging::Tag const> const, 4294967295u>, std::basic_format_string<char, char const*>, char const*>::log(mir::logging::Severity, std::span<std::reference_wrapper<mir::logging::Tag const> const, 4294967295u>, std::basic_format_string<char, char const*>, char const*, std::source_location)':
2026-08-13T08:00:42.3397295Z /usr/src/mir-2.29.0/build-armhf/src/platform/options/../../../../include/core/mir/log.h:166:(.text+0x3d44): undefined reference to `mir::log<mir::logging::Severity, std::span<std::reference_wrapper<mir::logging::Tag const> const, 4294967295u>, std::basic_string_view<char, std::char_traits<char> >, std::basic_format_args<std::basic_format_context<std::__format::_Sink_iter<char>, char> > >::log(mir::logging::Severity, std::span<std::reference_wrapper<mir::logging::Tag const> const, 4294967295u>, std::basic_string_view<char, std::char_traits<char> >, std::basic_format_args<std::basic_format_context<std::__format::_Sink_iter<char>, char> >, std::source_location)'
2026-08-13T08:00:42.3397418Z collect2: error: ld returned 1 exit status
```

RAOF:
I think that I'll mark this as draft; I think this could be simpler by doing a bit more work and entirely removing the `printf`-style and non-tagged API; LLMs should be reasonably good at that migration.

PR #5106, opened 2026-07-20
modernize fatal_error()

AlanGriffiths:
Replaces a printf-style fatal_error with a std::formatter style fatal_error

### DIFF
--- a/include/core/mir/fatal.h
+++ b/include/core/mir/fatal.h
@@ -47,9 +47,22 @@ void fatal_error(char const* reason, ...) __attribute__((format(printf, 1, 2)));
 [[noreturn]]
 void fatal_error(std::source_location const loc, std::string_view message);
 
+/**
+ * fatal_error() is strictly for "this should never happen" situations that you
+ * cannot recover from. Kills the program and dump core as cleanly as possible.
+ *   \param [in] fmt  A std::format-style format string.
+ *   \param [in] args  Type-erased format arguments.
+ *   \param [in] loc  The source location where the fatal error occurred.
+ */
+[[noreturn]]
+void fatal_error(std::string_view fmt, std::format_args args, std::source_location const loc);
+
 template<typename... Args>
-[[noreturn]] inline void fatal_error(std::source_location const loc, std::format_string<Args...> reason, Args&&... args)
-{ fatal_error(loc, std::format(reason, std::forward<Args>(args)...)); }
+[[noreturn]] inline void fatal_error(
+    std::source_location const loc,
+    std::format_string<Args...> reason,
+    Args const&... args)
+{ fatal_error(reason.get(), std::make_format_args(args...), loc); }
 } // namespace mir
 
 #define MIR_FATAL_ERROR(...) ::mir::fatal_error(::std::source_location::current(), __VA_ARGS__)
--- a/include/core/mir/fatal.h
+++ b/include/core/mir/fatal.h
@@ -36,7 +36,7 @@ namespace mir
  *   \param [in] reason  A printf-style format string.
  */
 [[noreturn, deprecated("Use MIR_FATAL_ERROR() instead")]]
-void fatal_error(char const* reason, ...) __attribute__((format(printf, 1, 2)));
+void fatal_error(char const* reason, ...) noexcept __attribute__((format(printf, 1, 2)));
 
 /**
  * fatal_error() is strictly for "this should never happen" situations that you
@@ -45,7 +45,7 @@ void fatal_error(char const* reason, ...) __attribute__((format(printf, 1, 2)));
  *   \param [in] message  A message describing the fatal error.
  */
 [[noreturn]]
-void fatal_error(std::source_location const loc, std::string_view message);
+void fatal_error(std::source_location const loc, std::string_view message) noexcept;
 
 /**
  * fatal_error() is strictly for "this should never happen" situations that you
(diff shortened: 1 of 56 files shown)


include/core/mir/fatal.h
```diff
@@ -23,15 +23,35 @@
...
+ * cannot recover from. Kills the program and dump core as cleanly as possible.
+ *   \param [in] loc  The source location where the fatal error occurred.
+ *   \param [in] message  A message describing the fatal error.
+ */
+[[noreturn]]
+void fatal_error(std::source_location const loc, std::string_view message);
+
+template<typename... Args>
+[[noreturn]] inline void fatal_error(std::source_location const loc, std::format_string<Args...> reason, Args&&... args)
+{ fatal_error(loc, std::format(reason, std::forward<Args>(args)...)); }
```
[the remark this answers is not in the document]
AlanGriffiths: @copilot please improve on this suggestion by moving the try-catch logic into a new overload so that we type erase the format string and args (using `std::make_format_args()`) and pass them to std::vformat in the implementation.

include/core/mir/fatal.h
```diff
@@ -23,15 +23,48 @@
...
  * fatal_error() is strictly for "this should never happen" situations that you
  * cannot recover from. Kills the program and dump core as cleanly as possible.
  *   \param [in] reason  A printf-style format string.
  */
+[[noreturn, deprecated("Use MIR_FATAL_ERROR() instead")]]
+void fatal_error(char const* reason, ...) noexcept __attribute__((format(printf, 1, 2)));
+
+/**
+ * fatal_error() is strictly for "this should never happen" situations that you
+ * cannot recover from. Kills the program and dump core as cleanly as possible.
```
tarek-y-ismail: Pre-existing:
```suggestion
 * cannot recover from. Kills the program and dumps core as cleanly as possible.
```

src/common/input/input_event.cpp
```diff
@@ -307,9 +299,9 @@ float mir_pointer_event_axis_value(MirPointerEvent const* pev, MirPointerAxis ax
        return pev->v_scroll().value120.as_value();
    case mir_pointer_axis_hscroll_value120:
        return pev->h_scroll().value120.as_value();
-   default:
-       mir::fatal_error("Invalid axis enumeration %d", axis);
-   }
+    default:
+        MIR_FATAL_ERROR("Invalid axis enumeration {}", static_cast<int>(axis));
+    }
```
tarek-y-ismail: Nit:
```suggestion
   default:
       MIR_FATAL_ERROR("Invalid axis enumeration {}", static_cast<int>(axis));
   }
```

src/common/symbols.map
```diff
@@ -63,6 +63,8 @@ global:
   mir_window_event_get_attribute_value;
 
   extern "C++" {
+    "std::formatter<MirEventType, char>::format(MirEventType, std::basic_format_context<std::__format::_Sink_iter<char>, char>&) const";
+    "std::formatter<MirInputEventType, char>::format(MirInputEventType, std::basic_format_context<std::__format::_Sink_iter<char>, char>&) const";
```
tarek-y-ismail, line 67: Will likely anger clang (or GCC, don't remember which wants spaces)

src/core/fatal.cpp
```diff
@@ -36,3 +36,29 @@ void mir::fatal_error(char const* reason, ...)
...
+void mir::fatal_error(std::string_view fmt, std::format_args args, std::source_location const loc) noexcept
+{
+    try
+    {
+        fatal_error(loc, std::vformat(fmt, args));
+    }
+    catch (...)
+    {
+        // Formatting failed with an exception! Just report what we can
+        fatal_error(loc, fmt);
```
tarek-y-ismail, line 62: Not sure about this, would this just print the format string unformatted?

src/miral/live_config_overrides_list.cpp
```diff
@@ -51,9 +51,9 @@ void mlc::OverridesList::for_each(Loader unchanged, Loader fresh, Loader modifie
         }
         catch (std::exception const& e)
         {
-            mir::fatal_error(
-                "Openers must not throw exceptions. Failed to open '%s' with exception '%s'",
-                event.path.string().c_str(),
+            MIR_FATAL_ERROR(
+                "Openers must not throw exceptions. Failed to open '{}' with exception '{}'",
+                event.path.string(),
```
tarek-y-ismail, line 56: I think we can ditch the `.string()` since we've embraced C++26?

tests/mir_test_framework/executable_path.cpp
```diff
@@ -46,9 +46,10 @@ std::string mir_test_framework::library_path()
 
     if (libpath.empty())
     {
-        // Try to find the location of libmircommon.so
+        // Try to find the location of libmircore.so
+        void (&fn_in_mircore)(std::source_location const loc, std::string_view) = mir::fatal_error;
         Dl_info library_info{nullptr, nullptr, nullptr, nullptr};
-        dladdr(reinterpret_cast<void*>(&mir::fatal_error), &library_info);
+        dladdr(reinterpret_cast<void*>(&fn_in_mircore), &library_info);
```
tarek-y-ismail, line 52: Probably a non-blocker: Need to look into this a bit more later. `dladdr` always confuses me.

tarek-y-ismail (approved):
Mostly good, got some pointers here and there.

src/common/symbols.map
tarek-y-ismail: Looks like it didn't. Maybe that was with `>>` ...

src/core/fatal.cpp
AlanGriffiths: Yes. But seems the least bad thing we can do as formatting the output threw an exception.

tests/mir_test_framework/executable_path.cpp
AlanGriffiths: You don't need to thinks about `dladdr()` to understand the change: This is just disambiguating the `fatal_error()` overloads.

src/miral/live_config_overrides_list.cpp
AlanGriffiths: "error: static assertion failed: std::formatter must be specialized for each type being formatted"

AlanGriffiths: I think this would be overkill:
```suggestion
#ifdef __cpp_lib_format_path
                event.path,
#else
                event.path.string(),
#endif
```

PR #5107, opened 2026-07-20
Modify a bunch of `miral::*::operator()(mir::Server&)`s to be const

tarek-y-ismail:
Related: https://github.com/canonical/mir/pull/5091#discussion_r3613532891

## What's new?

- What it says on the tin: If the operator doesn't modify the object, it should be `const`.

## How to test

- The code should stay compiling.


AlanGriffiths (changes requested):
> * If the operator doesn't modify the object, it should be `const`.

That's "implementation defined" `const` - which is a tempting but wrong idea:

1. When the implementation changes the API changes.
2. If the "logical state" is changed (e.g. on a pimpl) then `const` is a lie.

In modern (post C++11) C++ `const` means "immutable and absence of race conditions", which is generally not the case for `miral::*::operator()(mir::Server&)`

AlanGriffiths (approved):

PR #5108, opened 2026-07-21
Replace bespoke `enumerate_with_idx_type` with `std::views::enumerate`

tarek-y-ismail:
Quick cleanup opportunity I found while dabbling there.

### DIFF
--- a/src/platform/renderers/gl/renderer.cpp
+++ b/src/platform/renderers/gl/renderer.cpp
@@ -158,12 +158,6 @@ const GLchar* const vertex_shader_src =
     "   v_texcoord = texcoord;\n"
     "}\n"
 };
-
-template<typename Index, typename Range>
-    requires std::integral<Index>
-auto enumerate_with_idx_type(Range&& range) {
-    return std::views::zip(std::views::iota(Index{0}), std::forward<Range>(range));
-}
 }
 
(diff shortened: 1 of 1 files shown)


src/platform/renderers/gl/renderer.cpp
AlanGriffiths, line 747: This changes the type of `index` from `GLint` (alias `int`) to `std::array::size_type` (which is `std::size_t`).
That could lead to a narrowing conversion on L745

AlanGriffiths (approved):
I can see a "narrowing conversion" from the range [0,8) is safe. Can the tooling?

AlanGriffiths:
> I can see a "narrowing conversion" from the range [0,8) is safe. Can the tooling?

No:
> narrowing conversion from '__tuple_element_t<0UL, tuple<long, const int &>>' (aka 'long') to signed type 'GLint' (aka 'int') is implementation-defined.
> Blocking
> Rule:	bugprone-narrowing-conversions
> Level:	3
> Category:	BugProne rules
> Synopsis:	Checks for silent narrowing conversions.
> Ruleset:	C++ Coding Standard
> Tool:	ClangTidy 19.1.0

mattkae (approved):

PR #5111, opened 2026-07-24
Deduplicate gl renderer logs

tarek-y-ismail:
Addresses: https://github.com/canonical/mir/pull/5071#discussion_r3640766671

## What's new?

- De-duplicates `gl::Renderer` logs by capabilities such that when renderers are repeatedly re-created, our logging sinks aren't flooded with identical logs.

## How to test

```
<path/to/build>/bin/mir_unit_tests --gtest_filter=GLRenderer.logs_renderer_capabilities_only_once_for_successive_renderer_creation
```

### DIFF
--- a/src/platform/renderers/gl/renderer_factory.cpp
+++ b/src/platform/renderers/gl/renderer_factory.cpp
@@ -15,15 +15,31 @@
  */
 
 #include <mir/renderers/gl/renderer_factory.h>
+
+#include "renderer_logger.h"
+
+
 #include <mir/renderers/gl/renderer.h>
 #include <mir/graphics/platform.h>
 #include <mir/renderer/gl/gl_surface.h>
 
 namespace mrg = mir::renderer::gl;
 
+struct mrg::RendererFactory::Self
+{
+    Self() : logger{std::make_shared<Logger>()} {}
+
+    std::shared_ptr<Logger> const logger;
+};
+
+mrg::RendererFactory::RendererFactory()
+    : self{std::make_shared<mrg::RendererFactory::Self>()}
+{
+}
+
 auto mrg::RendererFactory::create_renderer_for(
     std::unique_ptr<graphics::gl::OutputSurface> output_surface,
     std::shared_ptr<graphics::GLRenderingProvider> gl_provider) const -> std::unique_ptr<mir::renderer::Renderer>
 {
-    return std::make_unique<Renderer>(std::move(gl_provider), std::move(output_surface));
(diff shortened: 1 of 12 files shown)


tarek-y-ismail:
I wonder if just updating the new logging system to add deduplication would completely sidestep this :thinking: ...

src/platform/renderers/gl/renderer_factory.cpp
```diff
@@ -15,15 +15,31 @@
...
+{
+    Self() : logger{std::make_shared<Logger>()} {}
+
+    std::shared_ptr<Logger> const logger;
+};
+
+mrg::RendererFactory::RendererFactory()
+    : self{std::make_shared<mrg::RendererFactory::Self>()}
+{
+}
```
mattkae, line 38: Why is the logger at the factory level? This does not make sense to me :)

Instead, I think we should do the following:

1. Move the logging mechanism to each renderer
2. Remove the `BoundedSet` in favor of just comparing that particular renderer's current capabilities to its previous capabilities and logging that
3. Hold the previous capabilities per-renderer on the renderer itself
4. Rewrite the `RendererLogger` so that it is just a straight function, and not a class with state (since the state will be held in `renderer.cpp`)

Does this sound right? Right now, it feels very complicated for no apparent gain.

mattkae (changes requested):
I see where this is going, but maybe it can be simpler?

src/platform/renderers/gl/renderer_factory.cpp
tarek-y-ismail: All of this work is to prevent log spam when renderers are re-created, specifically while resizing the magnifier. See `BasicScreenShooter::Self::renderer_for_buffer`, specifically `renderer_factory->create_renderer_for`.

If we hold the capabilities in a renderer, we can't access that when the next one is created. The solution is to hoist the data up one layer into the factory. It can't be hoisted up further as `create_renderer_for` is an API boundary I would like not to break. So the renderer factory is the only place fit to store capability data across renderer creations. 

Storing just the previous capability can work, but since resizing creates dozens of renderers, I'm not sure if that will be okay. Maybe down to 2 or 4 can work.

mattkae: I am still working through this (reading with the agent did), but would it be fixed if we _didn't_ recreate the renderer each time: https://github.com/canonical/mir/pull/5119

This would alleviate the issue (and maybe invalidate this entire PR?)

mattkae (changes requested):
Proposing an alternative

tarek-y-ismail:
Superseded by #5119

Issue #5112, opened 2026-08-09
gnome-console (kgx) does not start on Mir/main


AlanGriffiths:
### Version

main

### Environment

OS: Ubuntu 24.04
CPU: Intel
GPU 1: amdgpu
GPU 2:
Monitor Setup: n/a

### What happened? What did you expect to happen?

`kgx` opens a terminal window

### Reproduction steps

1. Use Miriway/edge with `alan_g` configuration
2. Press `Ctrl-Alt-T` 

### Relevant log output

```shell
$ WAYLAND_DEBUG=client kgx
[1251416.979]  -> wl_display@1.get_registry(new id wl_registry@2)
[1251416.992]  -> wl_display@1.sync(new id wl_callback@3)
[1251417.280] wl_display@1.delete_id(3)
[1251417.300] wl_registry@2.global(1, "zwp_linux_dmabuf_v1", 5)
[1251417.308]  -> wl_registry@2.bind(1, "zwp_linux_dmabuf_v1", 5, new id [unknown]@4)
[1251417.321]  -> zwp_linux_dmabuf_v1@4.get_default_feedback(new id zwp_linux_dmabuf_feedback_v1@5)
[1251417.329]  -> wl_display@1.sync(new id wl_callback@6)
[1251417.338] wl_registry@2.global(3, "wl_compositor", 6)
[1251417.345]  -> wl_registry@2.bind(3, "wl_compositor", 5, new id [unknown]@7)
[1251417.352] wl_registry@2.global(4, "wl_subcompositor", 1)
[1251417.359]  -> wl_registry@2.bind(4, "wl_subcompositor", 1, new id [unknown]@8)
[1251417.366] wl_registry@2.global(5, "wl_seat", 9)
[1251417.375] wl_registry@2.global(6, "wl_data_device_manager", 3)
[1251417.381]  -> wl_registry@2.bind(6, "wl_data_device_manager", 3, new id [unknown]@9)
[1251417.387] wl_registry@2.global(7, "wl_shell", 1)
[1251417.393] wl_registry@2.global(8, "xdg_wm_base", 5)
[1251417.398] wl_registry@2.global(10, "zxdg_output_manager_v1", 3)
[1251417.404]  -> wl_registry@2.bind(10, "zxdg_output_manager_v1", 3, new id [unknown]@10)
[1251417.411]  -> wl_display@1.sync(new id wl_callback@11)
[...]

PR #5113, opened 2026-08-09
Disable (by default) xdg_wm_dialog_v1. (Works around #5112)

AlanGriffiths:
We should investigate further and fix our wm_dialog implementation

PR #5116, opened 2026-08-10
Only type a toplevel as a dialog while it is actually modal

### LINKED ISSUE #5115: kgx not shown in application switcher
### Version

main

### Environment

OS: Ubuntu 26.04
Compositor: miriway, v26.08

Alt-tabbing with `kgx`  doesn't show any of its open windows. Alt-tabbing or Alt-graving with `kgx` focused switches to whatever app is selected in the app switcher. Other `kgx` windows opened later do not show up.

tarek-y-ismail:
Closes: #5115, #5112

## What's new?

- Creating an `xdg_dialog_v1` no longer retypes the toplevel to `mir_window_type_dialog`.
  GTK4 creates one for ordinary main windows and immediately calls
  `unset_modal()`, so every GTK4 main window (`kgx`, `gnome-calculator`, ...)
  was being marked a dialog. Dialogs are excluded from the foreign-toplevel
  protocols, so the compositor sent `closed` on a handle whose window was still
  mapped, and the window disappeared from the app switcher while staying on
  screen.
- `set_modal()`/`unset_modal()` are no longer no-ops. `unset_modal()` restores whatever the
  type was before, rather than assuming `normal`: a toplevel can carry a mir-shell archetype
  such as `satellite`, which hardcoding `normal` would silently discard.
- Adds `WindowWlSurfaceRole::pending_type()`, since these requests arrive before the first
  commit, when the scene surface doesn't exist yet and the requested type is still pending.

## How to test

```
bin/miral-test --gtest_filter='XdgDialog.*'
bin/mir_unit_tests --gtest_filter='ForeignToplevelHandleCreation.*'
```

Or on a live session:

- Run a shell with an app switcher (`miriway`, or `miral-shell` and Alt+Tab)
- Launch `kgx` (or `gnome-calculator`, or any GTK4 app)
- Observe the window is listed in the switcher. On `main` it's missing, as is every later instance.

### DIFF
--- a/tests/miral/xdg_dialog.cpp
+++ b/tests/miral/xdg_dialog.cpp
@@ -1,0 +1,333 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <miral/test_server.h>
+
+#include <miral/internal_client.h>
+#include <miral/minimal_window_manager.h>
+#include <miral/window_info.h>
+#include <miral/window_manager_tools.h>
+
+#include <mir/test/signal.h>
+
+#include <wayland-client.h>
+
+#include <sys/mman.h>
+#include <unistd.h>
+
+#include <condition_variable>
+#include <chrono>
+#include <optional>
+#include <cstring>
+#include <mutex>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+using namespace testing;
+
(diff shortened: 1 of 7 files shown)


AlanGriffiths:
The TICS violation needs fixing but looks sensible. Will test once the snap is built

tarek-y-ismail:
@AlanGriffiths Snaps are built, maybe check if that negates the issue on 24.04 as well?

AlanGriffiths:
> @AlanGriffiths Snaps are built, maybe check if that negates the issue on 24.04 as well?

The snap crashes on 24.04 - I doubt that has anything to do with these changes. It looks like some of the Mesa repackaging is broken on the `26/mir-libs-build` branch

AlanGriffiths (approved):
OK, having fixed Miriway `26/mir-libs-build` this does fix #5112 

(I still think we should ship 2.29 with xdg-dialog disabled though: c.f. #5113)

tests/miral/xdg_dialog.cpp
```diff
@@ -0,0 +1,333 @@
...
+            dialog_client(display);
+
+            if (auto const window = record->wait_for_window())
+                invoke_tools([&](miral::WindowManagerTools& tools) { type = tools.info_for(*window).type(); });
+
+            done.raise();
+        };
+
+        launcher.launch(client);
+        EXPECT_TRUE(done.wait_for(std::chrono::milliseconds{10}));
```
AlanGriffiths: This timeout appears be too short on some CI builds

```suggestion
        EXPECT_TRUE(done.wait_for(std::chrono::milliseconds{100}));
```

PR #5079, opened 2026-07-08
Migrate from wayland-server.h to wayland-rs

mattkae:
Closes #???

Related: #???, https://..., ...

## Checklist

- [ ] Tests added and pass
- [ ] Adequate documentation added
- [ ] (optional) Added Screenshots or videos

PR #5117, opened 2026-08-10
feat(wayland-rs): advertise wl_output as a dynamic per-monitor global

mattkae:
This was discovered during the grand refactor from wayland_server.h to wayland-rs.

Every global in the generated wayland-rs frontend is registered once at startup and bound through the single, shared `GlobalFactory`. That model does not fit `wl_output`: a compositor advertises one `wl_output` global per connected monitor, creates them as monitors are plugged in, and withdraws them as monitors go away. A single statically-registered global has no way to say *which* monitor a client is binding, and no way to disappear.

Generate a parallel, per-monitor path for `wl_output`:

* A new `OutputGlobalBinder` C++ interface is generated alongside `GlobalFactory`. Where `GlobalFactory` owns the binding logic for every static global, an `OutputGlobalBinder` captures the state of exactly one monitor: `bind()` builds that monitor's `wl_output` object for a client, and `can_view()` decides whether a client may see the global at all.

* A dedicated `GlobalDispatch<WlOutput, OutputGlobalData>` implementation is generated, using a distinct global-data type that carries the per-monitor binder rather than the shared factory. The binder lives behind a `Mutex` because `GlobalDispatch` hands out only a shared reference to the global data, while the cxx-generated binder methods need `Pin<&mut _>`.

* `wl_output` is consequently excluded from the static global registration, since it is now advertised exclusively through this path.

* `WaylandServer::create_output_global()` is exposed over the bridge so C++ can create one global per monitor. It returns a `Box<OutputGlobal>` whose `Drop` removes the global from the display, giving C++ RAII control over the global's lifetime -- which is what makes monitor hotplug/unplug expressible.

To support creating and destroying globals while the server is running, `WaylandServer` now publishes a clone of the display handle for the duration of `run()`. The handle is internally synchronised by the backend, so globals may be created from any thread; it is cleared again before `run()` returns so that no one can hand out globals against an inert display.

This lands ahead of the wider wayland-rs frontend migration, which needs per-monitor outputs before Mir's `OutputManager` can be ported onto the new frontend.

## How to test
Check that the code is generated properly.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
+++ b/src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
@@ -729,6 +729,10 @@ fn generate_dispatch_implementations(protocol: &WaylandProtocol) -> TokenStream 
         .interfaces
         .iter()
         .filter(|interface| interface.is_global)
+        // `wl_output` gets its `GlobalDispatch` from `generate_output_dynamic_global`
+        // (per-monitor, `OutputGlobalData`), so it must not also get a static
+        // `SharedFactory` implementation here.
+        .filter(|interface| interface.name != "wl_output")
         .collect();
 
     let global_dispatch_impls = global_interfaces.iter().map(|interface| {
--- a/src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
+++ b/src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
@@ -34,6 +34,8 @@ pub fn generate_dispatch_rs(protocols: &Vec<WaylandProtocol>) -> TokenStream {
 
     let server_side_factories = generate_server_side_factories(protocols);
 
+    let output_dynamic_global = generate_output_dynamic_global(protocols);
+
     quote! {
         #[allow(dead_code, unused_imports)]
         mod dispatch {
@@ -84,7 +86,63 @@ pub fn generate_dispatch_rs(protocols: &Vec<WaylandProtocol>) -> TokenStream {
             #(#generated_dispatch_implementations)*
 
             #server_side_factories
+
+            #output_dynamic_global
         }
+    }
+}
+
+/// Generate the machinery for dynamically-created `wl_output` globals.
+///
+/// Unlike every other global (which is registered once at startup and bound
+/// through the shared `GlobalFactory`), `wl_output` is advertised per-monitor
+/// and can come and go at runtime. To support this we use a distinct
+/// global-data type ([OutputGlobalData]) carrying a per-monitor C++
+/// [ffi::OutputGlobalBinder], and reuse [generate_global_dispatch_impl] to emit
+/// the `GlobalDispatch` implementation so that binding a `wl_output` follows
+/// exactly the same steps as binding any other global.
+///
+/// The binder is stored behind a `Mutex` because `GlobalDispatch::bind` and
+/// `can_view` only receive a shared reference to the global data, yet the C++
+/// binder methods (like the factory's) require a `Pin<&mut _>` to be invoked.
+fn generate_output_dynamic_global(protocols: &Vec<WaylandProtocol>) -> TokenStream {
+    let (protocol, interface) = protocols
+        .iter()
+        .find_map(|protocol| {
+            protocol
+                .interfaces
+                .iter()
+                .find(|interface| interface.name == "wl_output")
+                .map(|interface| (protocol, interface))
+        })
+        .expect("The wayland protocol must define wl_output");
+
+    let global_dispatch_impl = generate_global_dispatch_impl(
+        interface,
+        &generate_namespace(protocol),
+        GlobalBindSource::PerGlobalBinder,
+    );
+
+    quote! {
+        // A cxx opaque C++ type is not automatically `Send`/`Sync`. The binder
+        // is only ever accessed under the `Mutex` in `OutputGlobalData` (and,
+        // from C++, only on the event-loop thread), so this is sound.
+        unsafe impl Send for ffi::OutputGlobalBinder {}
+        unsafe impl Sync for ffi::OutputGlobalBinder {}
+
+        /// Per-global data for a dynamically-created `wl_output`, carrying the
+        /// C++ binder that builds the `wl_output` object for each client bind.
+        pub(crate) struct OutputGlobalData {
+            binder: Mutex<cxx::UniquePtr<ffi::OutputGlobalBinder>>,
+        }
+
+        impl OutputGlobalData {
+            pub(crate) fn new(binder: cxx::UniquePtr<ffi::OutputGlobalBinder>) -> Self {
+                OutputGlobalData { binder: Mutex::new(binder) }
+            }
+        }
+
+        #global_dispatch_impl
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/server/frontend_wayland/wayland_rs/src/wayland_server_core.rs
```diff
@@ -42,6 +42,11 @@ pub struct WaylandServer {
     work_signal: Mutex<Option<Ping>>,
     fd_listener_signal: Mutex<Option<Ping>>,
     pending_fd_listeners: Mutex<Vec<PendingFdListener>>,
+    /// A clone of the running display's handle, populated for the duration of
+    /// [WaylandServer::run]. It lets C++ create and destroy dynamic globals
+    /// (e.g. one `wl_output` per monitor) from any thread; the handle is
+    /// internally synchronised by the backend.
+    display_handle: Mutex<Option<DisplayHandle>>,
```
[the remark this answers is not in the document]
mattkae: It can just take `&self`, right?

src/server/frontend_wayland/wayland_rs/build_script/ffi_generation.rs
```diff
@@ -55,6 +55,9 @@ pub fn generate_ffi(protocols: &Vec<WaylandProtocol>, builders: &Vec<CppBuilder>
                 fn stop(self: &WaylandServer);
                 fn drain_queue(self: &mut WaylandServer) -> bool;
                 fn register_fd_ready_listener(self: &WaylandServer, fd: i32, callback: UniquePtr<FdReadyCallback>);
+                fn create_output_global(self: &WaylandServer, binder: UniquePtr<OutputGlobalBinder>) -> Box<OutputGlobal>;
```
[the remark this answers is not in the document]
mattkae: Should return a `Result` instead, u right

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -255,21 +376,21 @@ fn generate_global_dispatch_impl(
...
 
-            fn can_view(client: Client, global_data: &Arc<Mutex<cxx::UniquePtr<ffi::GlobalFactory>>>) -> bool {
-                let interface_name = #interface_name_str;
+            fn can_view(client: Client, global_data: &#global_data_type) -> bool {
+                #can_view_bindings
                 let client_id = Box::new(WaylandClientId::new(client.id()));
-                let mut guard = global_data.lock().unwrap();
-                (&mut *guard).pin_mut().can_view(interface_name, client_id)
+                let mut guard = #lock_target.lock().unwrap();
+                (&mut *guard).pin_mut().can_view(#can_view_args)
```
tarek-y-ismail: Are we always generating this? Is it necessary for non-dynamic globals?

src/server/frontend_wayland/wayland_rs/build_script/ffi_generation.rs
```diff
@@ -51,10 +51,13 @@ pub fn generate_ffi(protocols: &Vec<WaylandProtocol>, builders: &Vec<CppBuilder>
...
                 fn create_wayland_server() -> Box<WaylandServer>;
-                fn run(self: &mut WaylandServer, socket: &str, factory: UniquePtr<GlobalFactory>, notification_handler: UniquePtr<WaylandServerNotificationHandler>, work_callback: UniquePtr<WorkCallback>) -> Result<()>;
+                fn run(self: &WaylandServer, socket: &str, factory: UniquePtr<GlobalFactory>, notification_handler: UniquePtr<WaylandServerNotificationHandler>, work_callback: UniquePtr<WorkCallback>) -> Result<()>;
                 fn stop(self: &WaylandServer);
-                fn drain_queue(self: &mut WaylandServer) -> bool;
+                fn drain_queue(self: &WaylandServer) -> bool;
                 fn register_fd_ready_listener(self: &WaylandServer, fd: i32, callback: UniquePtr<FdReadyCallback>);
+                fn create_output_global(self: &WaylandServer, binder: UniquePtr<OutputGlobalBinder>) -> Result<Box<OutputGlobal>>;
+
+                type OutputGlobal;
```
tarek-y-ismail, line 59: Nit: Wouldn't it better to move this before its first use?

src/server/frontend_wayland/wayland_rs/build_script/ffi_generation.rs
```diff
@@ -51,10 +51,13 @@ pub fn generate_ffi(protocols: &Vec<WaylandProtocol>, builders: &Vec<CppBuilder>
             extern "Rust" {
                 type WaylandServer;
                 fn create_wayland_server() -> Box<WaylandServer>;
-                fn run(self: &mut WaylandServer, socket: &str, factory: UniquePtr<GlobalFactory>, notification_handler: UniquePtr<WaylandServerNotificationHandler>, work_callback: UniquePtr<WorkCallback>) -> Result<()>;
+                fn run(self: &WaylandServer, socket: &str, factory: UniquePtr<GlobalFactory>, notification_handler: UniquePtr<WaylandServerNotificationHandler>, work_callback: UniquePtr<WorkCallback>) -> Result<()>;
```
tarek-y-ismail, line 54: Why are we changing this to be non-mutable?

src/server/frontend_wayland/wayland_rs/src/wayland_server_core.rs
```diff
@@ -42,6 +42,11 @@ pub struct WaylandServer {
     work_signal: Mutex<Option<Ping>>,
     fd_listener_signal: Mutex<Option<Ping>>,
     pending_fd_listeners: Mutex<Vec<PendingFdListener>>,
+    /// A clone of the running display's handle, populated for the duration of
+    /// [WaylandServer::run]. It lets C++ create and destroy dynamic globals
+    /// (e.g. one `wl_output` per monitor) from any thread; the handle is
+    /// internally synchronised by the backend.
```
tarek-y-ismail: Half of the description says it's a handle, the other says its a factory

src/server/frontend_wayland/wayland_rs/src/wayland_server_core.rs
```diff
@@ -80,6 +127,27 @@ impl WaylandServer {
...
+        impl Drop for ClearDisplayHandleOnDrop<'_> {
+            fn drop(&mut self) {
+                *self
+                    .0
+                    .display_handle
+                    .lock()
+                    .expect("No recovery from lock poisoning") = None;
+            }
+        }
+        let _clear_display_handle = ClearDisplayHandleOnDrop(self);
```
tarek-y-ismail: This method accepts a `&mut self`, but `self` is non-mut. How does this work?

tarek-y-ismail (changes requested):
I'm quite rusty when it comes to wayland-rs, but I believe the old non-dynamic code paths still generate for `wl_output`? The rest of my comments are nits.

tarek-y-ismail:
For others that may review this, there are some pieces that integrate with this in #5079

src/server/frontend_wayland/wayland_rs/build_script/ffi_generation.rs
mattkae: Great question! `run` is holding `self` in a thread, but I also need to be able to call `create_output_global` at the same time. `run` can not hold a mutable reference therefore if we want to be able to safely call `create_output_global`. Luckily for us, it does not require a mutable ref!

src/server/frontend_wayland/wayland_rs/src/wayland_server_core.rs
mattkae: I can leave out the long-ish description, and just mention the duration for which it is valid.

src/server/frontend_wayland/wayland_rs/build_script/ffi_generation.rs
tarek-y-ismail: Aha, so you were just using a mutable reference where you didn't need too.

tarek-y-ismail (approved):
Looks sensible. Leaving it to you to merge or request an extra review 👍

PR #5119, opened 2026-08-10
Reuse one resizable renderer in BasicScreenShooter

mattkae:
This is a response to https://github.com/canonical/mir/pull/5111

## The Problem
`BasicScreenShooter` rebuilt its entire rendering stack whenever the capture buffer size changed. Because `WlrScreencopyManagerV1` owns a single `ScreenShooter` per client binding ([`wlr_screencopy_v1.cpp:365`](https://github.com/canonical/mir/blob/main/src/server/frontend_wayland/wlr_screencopy_v1.cpp#L365)), a client capturing two differently-sized outputs alternates sizes every frame, so `last_rendered_size` never stabilises and *every* capture paid for a fresh EGL context, renderbuffer, FBO, `OutputFilter` texture, `ProgramFactory` (all shaders recompiled and relinked) and ~10 `log_info()` lines dumping multi-kB EGL/GL extension strings.

Micro-benchmark on Mesa/radeonsi (AMD RX 7800 XT), alternating 1920x1080 ↔ 1280x720, 200 iterations:

| Path | per capture |
| --- | --- |
| Recreate context + storage + FBOs + programs (before) | **6.367 ms** |
| Reuse context, re-specify storage only (after) | **0.095 ms** |
| **Saving** | **6.27 ms (~67x)** |

Component breakdown on a warm context at 1920x1080: `eglCreateContext` + make current + destroy 1.238 ms; compile + link 2 programs 0.217 ms; `glRenderbufferStorage` 0.057 ms; `glTexImage2D` 0.040 ms. The parts only sum to ~1.5 ms — the remaining ~4.9 ms is the cost of allocating FBOs/renderbuffers/textures on a *cold* context, where the driver cannot reuse its buffer-object pools.

6.4 ms is ~38% of a 60Hz frame budget, burned per capture before any actual rendering, on a discrete GPU. On integrated/software GL (sanitiser and VM CI targets, low-end devices) it will be materially worse.

## What's new?
- `CPUCopyOutputSurface` now respecifies its storage size in `bind` to ensure that it has a properly sized texture
- `Renderer` now calls `ensure_texture_storage_for` on `bind` to ensure that it has a properly sized texture
- `BasicScreenShooter` builds its sink, surface and renderer once and resizes to match the new size after (thus avoiding the cost of rebuilding, unless the format changes)

## How to test
1. Run `miral-app`
2. Enable the magnifier
3. Resize it how you please and see that it still functions correctly.

## Notes for reviewers
- `BasicScreenShooter`'s sink only offers `CPUAddressableDisplayAllocator`, so `surface_for_sink` always lands on `CPUCopyOutputSurface` for both gbm-kms and renderer-generic-egl.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/platform/graphics/cpu_copy_output_surface.cpp
+++ b/src/platform/graphics/cpu_copy_output_surface.cpp
@@ -157,6 +157,12 @@ public:
     auto layout() const -> Layout;
 
 private:
+    /* The size of the target buffer can change between commits (notably, the
+     * screen shooter captures to whatever buffer the client provides), so
+     * (re)specify our storage whenever it no longer matches.
+     */
+    void ensure_storage_for(geom::Size size);
+
     mg::CPUAddressableDisplayAllocator& allocator;
     EGLDisplay const dpy;
     EGLContext ctx;
@@ -164,6 +170,7 @@ private:
     RenderbufferHandle colour_buffer;
     std::shared_ptr<RenderbufferHandle> depth_stencil_buffer;
     FramebufferHandle fbo;
+    geom::Size allocated_size;
 };
 
 mgc::CPUCopyOutputSurface::CPUCopyOutputSurface(
@@ -215,16 +222,29 @@ mgc::CPUCopyOutputSurface::Impl::Impl(
     : allocator{allocator},
       dpy{dpy},
       ctx{create_current_context(dpy, share_ctx)},
-      format{select_format_from(allocator)}
+      format{select_format_from(allocator)},
+      depth_stencil_buffer{
+          (config.depth_buffer_bits() || config.stencil_buffer_bits())
+              ? std::make_shared<RenderbufferHandle>()
+              : nullptr}
 {
-    glBindRenderbuffer(GL_RENDERBUFFER, colour_buffer);
-    glRenderbufferStorage(GL_RENDERBUFFER, GL_RGBA8_OES, size().width.as_int(), size().height.as_int());
+    ensure_storage_for(size());
+}
 
-    if (config.depth_buffer_bits() || config.stencil_buffer_bits())
+void mgc::CPUCopyOutputSurface::Impl::ensure_storage_for(geom::Size size)
+{
+    if (size == allocated_size)
     {
-        depth_stencil_buffer = std::make_shared<RenderbufferHandle>();
+        return;
+    }
+
+    glBindRenderbuffer(GL_RENDERBUFFER, colour_buffer);
+    glRenderbufferStorage(GL_RENDERBUFFER, GL_RGBA8_OES, size.width.as_int(), size.height.as_int());
+
+    if (depth_stencil_buffer)
+    {
         glBindRenderbuffer(GL_RENDERBUFFER, depth_stencil_buffer->operator GLuint());
-        glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH24_STENCIL8_OES, size().width.as_int(), size().height.as_int());
+        glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH24_STENCIL8_OES, size.width.as_int(), size.height.as_int());
     }
 
     glBindFramebuffer(GL_FRAMEBUFFER, fbo);
@@ -267,6 +287,8 @@ mgc::CPUCopyOutputSurface::Impl::Impl(
                         std::string{"Unknown GL framebuffer error code: "} + std::to_string(status)}));
         }
     }
+
+    allocated_size = size;
 }
 
 mgc::CPUCopyOutputSurface::Impl::~Impl()
@@ -291,6 +313,7 @@ mgc::CPUCopyOutputSurface::Impl::~Impl()
     // We're the current EGL context; destroy our GL resources...
     fbo.reset();
     colour_buffer.reset();
+    depth_stencil_buffer.reset();
 
     // Now release our context, and delete it.
     release_current();
@@ -299,6 +322,7 @@ mgc::CPUCopyOutputSurface::Impl::~Impl()
 
 void mgc::CPUCopyOutputSurface::Impl::bind()
 {
+    ensure_storage_for(size());
     glBindFramebuffer(GL_FRAMEBUFFER, fbo);
 }
 
--- a/src/platform/graphics/cpu_copy_output_surface.cpp
+++ b/src/platform/graphics/cpu_copy_output_surface.cpp
@@ -161,7 +161,7 @@ private:
      * screen shooter captures to whatever buffer the client provides), so
      * (re)specify our storage whenever it no longer matches.
      */
-    void ensure_storage_for(geom::Size size);
+    void ensure_storage_for_size();
 
     mg::CPUAddressableDisplayAllocator& allocator;
     EGLDisplay const dpy;
@@ -228,23 +228,24 @@ mgc::CPUCopyOutputSurface::Impl::Impl(
               ? std::make_shared<RenderbufferHandle>()
               : nullptr}
 {
-    ensure_storage_for(size());
+    ensure_storage_for_size();
 }
 
-void mgc::CPUCopyOutputSurface::Impl::ensure_storage_for(geom::Size size)
+void mgc::CPUCopyOutputSurface::Impl::ensure_storage_for_size()
 {
-    if (size == allocated_size)
+    auto const next_size = size();
+    if (next_size == allocated_size)
     {
         return;
     }
 
     glBindRenderbuffer(GL_RENDERBUFFER, colour_buffer);
-    glRenderbufferStorage(GL_RENDERBUFFER, GL_RGBA8_OES, size.width.as_int(), size.height.as_int());
+    glRenderbufferStorage(GL_RENDERBUFFER, GL_RGBA8_OES, next_size.width.as_int(), next_size.height.as_int());
 
     if (depth_stencil_buffer)
     {
         glBindRenderbuffer(GL_RENDERBUFFER, depth_stencil_buffer->operator GLuint());
-        glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH24_STENCIL8_OES, size.width.as_int(), size.height.as_int());
+        glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH24_STENCIL8_OES, next_size.width.as_int(), next_size.height.as_int());
     }
 
     glBindFramebuffer(GL_FRAMEBUFFER, fbo);
@@ -288,7 +289,7 @@ void mgc::CPUCopyOutputSurface::Impl::ensure_storage_for(geom::Size size)
         }
     }
 
-    allocated_size = size;
+    allocated_size = next_size;
 }
 
 mgc::CPUCopyOutputSurface::Impl::~Impl()
@@ -322,7 +323,7 @@ mgc::CPUCopyOutputSurface::Impl::~Impl()
 
 void mgc::CPUCopyOutputSurface::Impl::bind()
 {
-    ensure_storage_for(size());
+    ensure_storage_for_size();
     glBindFramebuffer(GL_FRAMEBUFFER, fbo);
 }
 
--- a/src/platform/renderers/gl/renderer.cpp
+++ b/src/platform/renderers/gl/renderer.cpp
@@ -317,7 +317,7 @@ public:
     // NOTE: This must be called with a current GL context
     OutputFilter(std::unique_ptr<mg::gl::OutputSurface> output)
      : output{std::move(output)},
-        texture{make_texture(this->output->size())},
+        texture{make_texture()},
         framebuffer{make_framebuffer(texture)},
         filter{mir_output_filter_none},
         program{nullptr},
@@ -356,6 +356,8 @@ public:
             output->bind();
             return;
         }
+
+        ensure_texture_storage_for(output->size());
 
         glBindFramebuffer(GL_DRAW_FRAMEBUFFER, framebuffer);
 
@@ -490,12 +492,28 @@ private:
         return program;
     }
 
-   static GLuint make_texture(mir::geometry::Size size)
+   static GLuint make_texture()
     {
         GLuint tex{0};
         glGenTextures(1, &tex);
         glActiveTexture(GL_TEXTURE0);
         glBindTexture(GL_TEXTURE_2D, tex);
+        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST);
+        return tex;
+    }
+
+    /* The output surface can be resized under us, so (re)specify the
+     * intermediate texture whenever it no longer matches. This is only ever
+     * reached with a filter active, so an unfiltered output never pays for it.
+     */
+    void ensure_texture_storage_for(mir::geometry::Size size)
+    {
+        if (size == texture_size)
+            return;
+
+        glActiveTexture(GL_TEXTURE0);
+        glBindTexture(GL_TEXTURE_2D, texture);
         glTexImage2D(GL_TEXTURE_2D, 0,
                      GL_RGBA,
                      size.width.as_value(),
@@ -504,9 +522,7 @@ private:
                      GL_RGBA,
                      GL_UNSIGNED_BYTE,
                      nullptr);
-        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
-        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST);
-        return tex;
+        texture_size = size;
     }
 
     static GLuint make_framebuffer(GLuint tex)
@@ -522,6 +538,7 @@ private:
     std::unique_ptr<mg::gl::OutputSurface> output;
     TextureHandle const texture;
     FramebufferHandle const framebuffer;
+    mir::geometry::Size texture_size;
     MirOutputFilter filter;
     std::unique_ptr<ProgramHandle> program;
     GLint position_attrib;
@@ -657,6 +674,16 @@ auto mrg::Renderer::render(mg::RenderableList const& renderables) const -> std::
 {
     output_surface->make_current();
     output_surface->bind();
+
+    /* The output surface can change size between renders (the screen shooter
+     * captures to whatever buffer the client hands us), and set_viewport() is a
+     * no-op when the logical area is unchanged, so the GL viewport has to be
+     * re-derived here.
+     */
+    if (output_surface->size() != last_output_size)
+    {
+        update_gl_viewport();
+    }
 
     glClearColor(clear_color[0], clear_color[1], clear_color[2], clear_color[3]);
     glColorMask(GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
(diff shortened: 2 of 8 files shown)


src/platform/graphics/cpu_copy_output_surface.cpp
```diff
@@ -215,16 +222,29 @@ mgc::CPUCopyOutputSurface::Impl::Impl(
...
+void mgc::CPUCopyOutputSurface::Impl::ensure_storage_for(geom::Size size)
+{
+    if (size == allocated_size)
+    {
+        return;
+    }
+
     glBindRenderbuffer(GL_RENDERBUFFER, colour_buffer);
-    glRenderbufferStorage(GL_RENDERBUFFER, GL_RGBA8_OES, size().width.as_int(), size().height.as_int());
+    glRenderbufferStorage(GL_RENDERBUFFER, GL_RGBA8_OES, size.width.as_int(), size.height.as_int());
```
tarek-y-ismail: Unrelated change?

src/platform/graphics/cpu_copy_output_surface.cpp
```diff
@@ -215,16 +222,29 @@ mgc::CPUCopyOutputSurface::Impl::Impl(
...
 
-    if (config.depth_buffer_bits() || config.stencil_buffer_bits())
+    if (depth_stencil_buffer)
     {
-        depth_stencil_buffer = std::make_shared<RenderbufferHandle>();
         glBindRenderbuffer(GL_RENDERBUFFER, depth_stencil_buffer->operator GLuint());
-        glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH24_STENCIL8_OES, size().width.as_int(), size().height.as_int());
+        glRenderbufferStorage(GL_RENDERBUFFER, GL_DEPTH24_STENCIL8_OES, size.width.as_int(), size.height.as_int());
     }
```
tarek-y-ismail, line 249: Two points:
1. This is unrelated to the purpose of this PR
2. Since we moved the depth stencil buffer creation to the member initializer list, and this code doesn't depend on anything before it, maybe the creation _and_ initialization can be wrapped in a utility function and all be done in the member initializer list.

src/platform/graphics/cpu_copy_output_surface.cpp
```diff
@@ -157,13 +157,20 @@ class mgc::CPUCopyOutputSurface::Impl
...
+    void ensure_storage_for(geom::Size size);
+
     mg::CPUAddressableDisplayAllocator& allocator;
     EGLDisplay const dpy;
     EGLContext ctx;
     DRMFormat const format;
     RenderbufferHandle colour_buffer;
     std::shared_ptr<RenderbufferHandle> depth_stencil_buffer;
     FramebufferHandle fbo;
+    geom::Size allocated_size;
```
tarek-y-ismail, line 173: It may become confusing to have two sizes:
-  `Impl::size()` -> `allocator->output_size()` -> `basic_screen_shooter::current_size`
- `allocated_size` -> `Impl::size()`!?

Which means we can probably skip adding `allocated_size`.

And since `ensure_storage_for` is always called with `size()`, we can probably drop the argument and rename it to `ensure_storage_for_size`

src/platform/graphics/cpu_copy_output_surface.cpp
```diff
@@ -291,6 +313,7 @@ mgc::CPUCopyOutputSurface::Impl::~Impl()
     // We're the current EGL context; destroy our GL resources...
     fbo.reset();
     colour_buffer.reset();
+    depth_stencil_buffer.reset();
```
tarek-y-ismail, line 317: Good, but unrelated

src/platform/renderers/gl/renderer.cpp
```diff
@@ -931,7 +958,7 @@ void mrg::Renderer::set_viewport(geometry::Rectangle const& rect)
     update_gl_viewport();
 }
 
-void mrg::Renderer::update_gl_viewport()
+void mrg::Renderer::update_gl_viewport() const
```
tarek-y-ismail: Why is this method `const` while mutating `last_output_size`? Since we're mutating data, we should drop both the `mutable` from `last_output_size` and the `const` from here.

src/platform/renderers/gl/renderer.cpp
```diff
@@ -490,12 +492,28 @@ class mrg::Renderer::OutputFilter : public mg::gl::OutputSurface
...
+        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST);
+        return tex;
+    }
+
+    /* The output surface can be resized under us, so (re)specify the
+     * intermediate texture whenever it no longer matches. This is only ever
+     * reached with a filter active, so an unfiltered output never pays for it.
+     */
+    void ensure_texture_storage_for(mir::geometry::Size size)
```
tarek-y-ismail, line 511: Only called with `output->size()`. Can drop the argument and be called `ensure_texture_storage_for_output`

tarek-y-ismail: It's only called in one place too, so it can be inlined.

src/platforms/gbm-kms/server/buffer_allocator.cpp
tarek-y-ismail, line 1: Good optimization, but it's not related to the "don't re-create the renderer" goal.

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -198,10 +223,29 @@ auto mc::BasicScreenShooter::Self::render(
 
     scene_elements.clear();
 
-    auto& renderer = renderer_for_buffer(buffer);
+    auto const buffer_size = buffer->size();
+    if (buffer_size.height == geom::Height{0} || buffer_size.width == geom::Width{0})
+    {
+        BOOST_THROW_EXCEPTION((std::runtime_error{"Attempt to capture to a zero-sized buffer"}));
+    }
```
tarek-y-ismail: Needless code churn, not sure why it was moved from `renderer_from_buffer` (now `renderer_for`)

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -198,10 +223,29 @@ auto mc::BasicScreenShooter::Self::render(
...
+     * sure it can't outlive this capture: a buffer left pending would fail every
+     * subsequent capture. On success it has already been claimed by alloc_fb().
+     */
+    output->set_next_buffer(buffer);
+    auto const drop_unconsumed_buffer = mir::raii::paired_calls(
+        [](){},
+        [this]()
+        {
+            output->clear_next_buffer();
+        });
```
tarek-y-ismail, line 241: We can just call `clear_next_buffer` after `render` or `suspend` I assume?

tarek-y-ismail: @copilot Check if this is a repeated pattern in the codebase. If so, create a new utility class called `Defer` which would take only the destruction callback. Replace all usages.

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -104,11 +112,23 @@ class mc::BasicScreenShooter::Self::OneShotBufferDisplayProvider : public mg::CP
...
+    /* The pending buffer is consumed by alloc_fb() during commit(), but the
+     * surface we render through outlives any individual capture and will still
+     * query us, so remember the geometry of the most recent buffer.
+     */
+    geom::Size current_size;
+    MirPixelFormat current_format{mir_pixel_format_invalid};
 };
 
-class OffscreenDisplaySink : public mg::DisplaySink
+class mc::BasicScreenShooter::Self::OffscreenDisplaySink : public mg::DisplaySink
```
tarek-y-ismail, line 131: Is this rename really required?

tarek-y-ismail: To answer my own question: Yes, it is required because using a generic `DisplaySink` prevents us from accessing `set_size` (added below)

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
```diff
@@ -316,3 +358,116 @@ TEST_F(BasicScreenShooter, compositor_id_is_not_null)
...
+{
+    supply_unlimited_renderers();
+
+    EXPECT_CALL(*gl_provider, surface_for_sink(_, _)).Times(1);
+
+    for (auto const& size : {geom::Size{800, 600}, geom::Size{1024, 768}, geom::Size{800, 600}})
+    {
+        EXPECT_CALL(callback, Call(std::make_optional(clock->now())));
+        capture_and_run(std::make_shared<mtd::StubBuffer>(size));
+        Mock::VerifyAndClearExpectations(&callback);
```
tarek-y-ismail: The lines removed were already tested in the previous test.

```suggestion
        capture_and_run(std::make_shared<mtd::StubBuffer>(size));
```

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
```diff
@@ -316,3 +358,116 @@ TEST_F(BasicScreenShooter, compositor_id_is_not_null)
...
+
+TEST_F(BasicScreenShooter, output_geometry_tracks_the_current_buffer)
+{
+    supply_unlimited_renderers();
+
+    for (auto const& size : {geom::Size{800, 600}, geom::Size{1920, 1080}, geom::Size{640, 480}})
+    {
+        EXPECT_CALL(callback, Call(std::make_optional(clock->now())));
+        capture_and_run(std::make_shared<mtd::StubBuffer>(size));
+        Mock::VerifyAndClearExpectations(&callback);
```
tarek-y-ismail:
```suggestion
        capture_and_run(std::make_shared<mtd::StubBuffer>(size));
```

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
```diff
@@ -316,3 +358,116 @@ TEST_F(BasicScreenShooter, compositor_id_is_not_null)
...
+    {
+        EXPECT_CALL(callback, Call(std::make_optional(clock->now())));
+        capture_and_run(std::make_shared<mtd::StubBuffer>(geom::Size{800, 600}, format));
+        Mock::VerifyAndClearExpectations(&callback);
+    }
+
+    EXPECT_THAT(renderers_created, Eq(3));
+}
+
+TEST_F(BasicScreenShooter, recovers_from_a_failure_to_build_a_renderer)
```
tarek-y-ismail: I assume this is testing the new `renderer_for` and how it only assigns after all pieces have been created? Assigning as each variable is created, like the old code, still passes this test. So it's either assuming something incorrect, or flat out wrong.

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
```diff
@@ -316,3 +358,116 @@ TEST_F(BasicScreenShooter, compositor_id_is_not_null)
...
+                ON_CALL(*surface, commit())
+                    .WillByDefault([cpu_provider, format]() { return cpu_provider->alloc_fb(format); });
+                return surface;
+            });
+
+    EXPECT_CALL(callback, Call(std::make_optional(clock->now())));
+    capture_and_run(buffer);
+}
+
+TEST_F(BasicScreenShooter, recovers_from_a_failure_in_render)
```
tarek-y-ismail, line 369: As I understand it, this is testing `clear_next_buffer` inside `BasicScreenShooter::Self::render`. Removing the call still results in the test passing.

tests/unit-tests/renderers/gl/test_gl_renderer.cpp
```diff
@@ -420,6 +420,50 @@ TEST_F(GLRenderer, unchanged_viewport_avoids_gl_calls)
...
+    auto output_surface = make_output_surface();
+    auto* const output_surface_ptr = output_surface.get();
+
+    ON_CALL(*output_surface, size())
+        .WillByDefault(Return(mir::geometry::Size{1920, 1080}));
+    ON_CALL(*output_surface, commit())
+        .WillByDefault([]() { return std::unique_ptr<mg::Framebuffer>(); });
+
+    mrg::Renderer renderer(gl_platform, std::move(output_surface));
+    renderer.set_viewport(view_area);
```
tarek-y-ismail, line 440:
```suggestion
    EXPECT_CALL(mock_gl, glViewport(0, 0, 1920, 1080));
    renderer.set_viewport(view_area);
```

tarek-y-ismail (changes requested):
That took a while to review 😵‍💫 

Maybe I should've done this before sinking this much time into the review, but it doesn't appear to actually prevent the log spamming.

Here's how I tested:
1. On this branch, I launched miral-app
2. Enabled the magnifier
3. Presset Ctrl+Alt++ to increase the size

As you increase or decrease the size, the code will log as before.

To double check I also pulled the new decoupled magnifier code and tried it. Had to make a couple modifications (disable manual log deduplication, set graphics log level to debug). But that also exhibits the same issue.

tarek-y-ismail:
Hmm, that's weird. I'm used to Copilot making a different PR when asked to make changes. Sorry, Matt. We can pull that out manually.

tarek-y-ismail:
@copilot Copy over the "defer" commits to a new PR.

src/platform/graphics/cpu_copy_output_surface.cpp
mattkae: Nope, because `size` is a parameter here now.

src/platform/graphics/cpu_copy_output_surface.cpp
mattkae: This is related because we don't want to recreate buffers whenever the surface is created. In this case, we just resize the stencil buffer, which saves us overhead of recreating it (although that is minor).

The more important reason is that `ensure_storage_for` does not always have access to the `GLConfig` (during `bind` for example) because it cannot gain ownership over it in its CTOR. Hence, we can only know if we want a depth stencil buffer at CTOR time. For that reason, it is simplest to just construct it once there and update its size when necessary.

src/platform/graphics/cpu_copy_output_surface.cpp
mattkae: We need to somehow compare what `CPUCopyOutputSurface` allocated vs what the current size of the output is though. So we do need somehow store that information to know when to update :)

I agree on the latter part though.

src/platform/renderers/gl/renderer.cpp
mattkae: True stuff!

src/platform/renderers/gl/renderer.cpp
mattkae: It is called int two places - CTOR and `bind`

src/server/compositor/basic_screen_shooter.cpp
mattkae: Let's land that type of change after this. It will touch other things for sure

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
mattkae: This is testing that we can throw in the `renderer_for` (e.g. when we have a zero width/height), but still recover from it in the next attempt at a capture if it is valid later.

I believe this is a correct test, no?

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
mattkae: Removing the RAII call you mean?

src/platform/renderers/gl/renderer.cpp
```diff
@@ -917,8 +907,16 @@ void mrg::Renderer::draw(mg::Renderable const& renderable) const
...
+    /* The output surface can change size under us (the screen shooter captures
+     * to whatever buffer the client hands us), which changes the GL viewport
+     * even when the logical area is unchanged.
+     */
     if (rect == viewport)
+    {
+        if (output_surface->size() != last_output_size)
+            update_gl_viewport();
         return;
+    }
```
tarek-y-ismail, line 920: Looked wrong at first. But this handles the case where the output surface size changes _or_ the viewport changes.

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
```diff
@@ -381,7 +381,9 @@ TEST_F(BasicScreenShooter, only_acquires_one_output_surface_across_differently_s
 
     for (auto const& size : {geom::Size{800, 600}, geom::Size{1024, 768}, geom::Size{800, 600}})
     {
+        EXPECT_CALL(callback, Call(std::make_optional(clock->now())));
```
tarek-y-ismail: You can also just use a `NiceMock` (no warnings), or a `NaggyMock` (prints warnings)

src/server/compositor/basic_screen_shooter.cpp
```diff
@@ -198,10 +223,23 @@ auto mc::BasicScreenShooter::Self::render(
...
     renderer.set_output_transform(transform);
     renderer.set_viewport(area);
     renderer.set_output_filter(output_filter->filter());
+
+    /* Hand the buffer over only once we have a renderer to consume it, and make
+     * sure it can't outlive this capture: a buffer left pending would fail every
+     * subsequent capture. On success it has already been claimed by alloc_fb().
+     */
+    output->set_next_buffer(buffer);
+    auto const drop_unconsumed_buffer = mir::raii::paired_calls(
```
tarek-y-ismail, line 236: `raii::defer` from #5130 may be useful here.

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
tarek-y-ismail: It's throwing in `surface_for_sink` though? That means we're past the width/height check. That's why I think it's related to the split between creating the new renderer and assignment. When I reverted the changes (assigning directly to the members), the test still passed.

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
tarek-y-ismail: Yes. If `render` fails, then that should run to do cleanup. But if the cleanup not happening doesn't affect the outcome of the test, then what does it actually test?

Edit: just tried it again. Tests fail as expected.

src/server/compositor/basic_screen_shooter.cpp
mattkae: Agreed, we can add it then.

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
mattkae: It's a `StrictMock` member, which I don't think is a problem personally. I think a good way to do this is to hoist it up and just do a `.Times(3)` on it to reduce noise.

tests/unit-tests/compositor/test_basic_screen_shooter.cpp
mattkae: I added `retains_the_previous_renderer_when_a_rebuild_fails`, which exercises the rebuild path (format change) instead. Reverting to direct assignment segfaults on it.

here's the commit: https://github.com/canonical/mir/pull/5119/changes/52e06494a7f50ad4ba89fc74985aac6392075685

Is that better in your opinion?

tarek-y-ismail: Yup. I think `recovers_from_a_failure_to_build_a_renderer` can be removed safely now.

tarek-y-ismail (approved):
One last thing to chop off, then you're free to merge!

PR #5126, opened 2026-08-12
gbm-kms: DisplaySink: stop using OverlappingOutputGroupping

muhammad23012009:
Fundamentally every DisplaySink should refer to an actual sink where a framebuffer can be rendered. With the previous approach, all displays with similar resolutions were getting grouped into one DisplaySink, which only works for clone mode and breaks otherwise. Therefore, revert back to the former approach to get multi-display back up.

## What's new?
Fixes broken multi-display without clone mode

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/platforms/gbm-kms/server/kms/display_buffer.cpp
+++ b/src/platforms/gbm-kms/server/kms/display_buffer.cpp
@@ -53,12 +53,12 @@ mgg::DisplaySink::DisplaySink(
     std::shared_ptr<struct gbm_device> gbm,
     mgg::BypassOption,
     std::shared_ptr<DisplayReport> const& listener,
-    std::vector<std::shared_ptr<KMSOutput>> const& outputs,
+    std::shared_ptr<KMSOutput> const& output,
     geom::Rectangle const& area,
     glm::mat2 const& transformation)
     : gbm{std::move(gbm)},
       listener(listener),
-      outputs(outputs),
+      output(output),
       area(area),
       transform{transformation},
       needs_set_crtc{false},
@@ -68,13 +68,7 @@ mgg::DisplaySink::DisplaySink(
 
     // If any of the outputs have a CRTC mismatch, we will want to set all of them
     // so that they're all showing the same buffer.
-    bool has_crtc_mismatch = false;
-    for (auto& output : outputs)
-    {
-        has_crtc_mismatch = output->has_crtc_mismatch();
-        if (has_crtc_mismatch)
-            break;
-    }
+    bool has_crtc_mismatch = output->has_crtc_mismatch();
 
     if (has_crtc_mismatch)
     {
@@ -90,9 +84,7 @@ mgg::DisplaySink::DisplaySink(
         ::memset(mapping->data(), 24, mapping->len());
 
         visible_fb = std::move(initial_fb);
-        for (auto &output: outputs) {
-            output->set_crtc(*visible_fb);
-        }
+        output->set_crtc(*visible_fb);
         listener->report_successful_drm_mode_set_crtc_on_construction();
     }
     listener->report_successful_display_construction();
@@ -151,7 +143,7 @@ void mgg::DisplaySink::for_each_display_sink(std::function<void(graphics::Displa
     // to all sorts of problems (e.g. an error during eglSwapBuffers
     // because the back buffer is empty). To avoid this, we don't iterate
     // over display sinks that refer to disconnected outputs.
-    if (!outputs.front()->connected())
+    if (!output->connected())
         return;
 
     f(*this);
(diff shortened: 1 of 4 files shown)


src/platforms/gbm-kms/server/kms/display_buffer.cpp
```diff
@@ -223,50 +222,10 @@ void mgg::DisplaySink::post()
...
-    if (outputs.size() == 1)
-    {
-        auto const& output = outputs.front();
-        auto const min_frame_interval = 1000ms / output->max_refresh_rate();
-        if (predicted_render_time < min_frame_interval)
-            recommend_sleep = min_frame_interval - predicted_render_time;
-    }
+    auto const min_frame_interval = 1000ms / output->max_refresh_rate();
+    if (predicted_render_time < min_frame_interval)
+        recommend_sleep = min_frame_interval - predicted_render_time;
```
RAOF, line 216: Now that we're always single-output, we should move the `wait_for_page_flip()` from the top of the function down to here.

```suggestion
    // Wait for the frame to be displayed
    // This makes us double-buffered; assuming we can reliably render within the frame deadline
    // that minimises latency.
    wait_for_page_flip();
    
    auto const min_frame_interval = 1000ms / output->max_refresh_rate();
    if (predicted_render_time < min_frame_interval)
        recommend_sleep = min_frame_interval - predicted_render_time;
```

src/platforms/gbm-kms/server/kms/display_buffer.cpp
```diff
@@ -69,11 +69,9 @@ mgg::DisplaySink::DisplaySink(
...
     bool has_crtc_mismatch = false;
-    for (auto& output : outputs)
+    if (output)
     {
         has_crtc_mismatch = output->has_crtc_mismatch();
-        if (has_crtc_mismatch)
-            break;
     }
 
     if (has_crtc_mismatch)
```
RAOF: This can be collapsed; `output` is guaranteed to be non-null, so we can just directly check `if (output->has_crtc_mismatch())` here.

src/platforms/gbm-kms/server/kms/display_buffer.cpp
```diff
@@ -90,7 +88,8 @@ mgg::DisplaySink::DisplaySink(
         ::memset(mapping->data(), 24, mapping->len());
 
         visible_fb = std::move(initial_fb);
-        for (auto &output: outputs) {
+        if (output)
+        {
             output->set_crtc(*visible_fb);
         }
```
RAOF: Again, `output` is¹ non-null; we don't need to check it.

¹: Or, at least, *should* be. If you're finding a null output we should be fixing that!

src/platforms/gbm-kms/server/kms/display_buffer.cpp
```diff
@@ -151,15 +150,15 @@ void mgg::DisplaySink::for_each_display_sink(std::function<void(graphics::Displa
     // to all sorts of problems (e.g. an error during eglSwapBuffers
     // because the back buffer is empty). To avoid this, we don't iterate
     // over display sinks that refer to disconnected outputs.
-    if (!outputs.front()->connected())
+    if (!output || !output->connected())
```
RAOF: Again, `output` should be non-null.

```suggestion
    if (!output->connected())
```

src/platforms/gbm-kms/server/kms/display_sink.h
```diff
@@ -86,11 +86,10 @@ class DisplaySink : public graphics::DisplaySink,
     void set_crtc(FBHandle const&);
 
     std::shared_ptr<struct gbm_device> const gbm;
-    bool holding_client_buffers{false};
     std::shared_ptr<FBHandle const> bypass_bufobj{nullptr};
     std::shared_ptr<DisplayReport> const listener;
 
-    std::vector<std::shared_ptr<KMSOutput>> outputs;
+    std::shared_ptr<KMSOutput> output;
```
RAOF: Hm. Is there any reason why this isn't `const`?
```suggestion
    std::shared_ptr<KMSOutput> const output;
```

RAOF (changes requested):
This looks reasonable, yeah. Just two things: when removing the iteration over `outputs` you don't need to check whether the new `output` is `null`; we should only ever be sending in a non-null `std::shared_ptr<KMSOutput>` to the constructor (and if you've found a time where we *aren't* doing that, we should fix that!).

The other is you've dropped the double-buffering optimisation; this can now be unconditional (rather than conditional on being not in clone mode, because we're now *never* in clone mode).

src/platforms/gbm-kms/server/kms/display_buffer.cpp
muhammad23012009: Hmm, so there's a problem with double buffering here, the unit tests expect page flips to only be executed after another page flip is scheduled, but with this approach it immediately waits for a page flip. should I change the unit tests to align with this as well?

RAOF: Hm, which unit test? They must have be testing cases with >1 output, because in the 1-output case we would previously have double-buffered anyway.

Regardless, yeah, please fix up the unit tests to align with always-wait-for-flip.

muhammad23012009: Tests are now aligned and pass locally.

RAOF (approved):
Yup, looks good now! Thanks!

PR #5130, opened 2026-08-12
Use `std::experimental::scope_exit` instead of a dtor only `paired_calls`

tarek-y-ismail:
Does what it says on the tin.

### DIFF
--- a/src/server/compositor/multi_threaded_compositor.cpp
+++ b/src/server/compositor/multi_threaded_compositor.cpp
@@ -37,6 +37,7 @@
 #include <thread>
 #include <chrono>
 #include <future>
+#include <experimental/scope>
 #include <boost/throw_exception.hpp>
 
 using namespace std::literals::chrono_literals;
@@ -106,8 +107,7 @@ public:
     try
     {
         mir::set_thread_name("Mir/Comp");
-        auto const signal_when_stopped = mir::raii::paired_calls(
-            [](){},
+        auto const signal_when_stopped = std::experimental::scope_exit(
             [this]()
             {
                 stopped.set_value();
--- a/src/server/console/linux_virtual_terminal.cpp
+++ b/src/server/console/linux_virtual_terminal.cpp
@@ -41,6 +41,7 @@
 #include <sstream>
 #include <stdexcept>
 #include <csignal>
+#include <experimental/scope>
 
 #include <linux/vt.h>
 #include <linux/kd.h>
@@ -456,8 +457,7 @@ void mir::LinuxVirtualTerminal::register_switch_handlers(
                      * Otherwise we leave the VT subsystem in a glorious state of
                      * waiting forever for us to ACK the switch.
                      */
-                    auto ack_when_done = mir::raii::paired_calls(
-                        [](){},
+                    auto ack_when_done = std::experimental::scope_exit(
                         [this]()
                         {
                             fops->ioctl(vt_fd.fd(), VT_RELDISP, VT_ACKACQ);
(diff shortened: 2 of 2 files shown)


include/common/mir/raii.h
```diff
@@ -73,6 +75,16 @@ inline auto paired_calls(Creator&& creator, Deleter&& deleter) ->
...
+/**
+ * Creates an RAII object that calls \p cleanup on destruction.
+ * Use this when only a cleanup action is needed (no paired setup call).
+ *
+ * \param cleanup called on destruction of the returned object
+ */
+template<typename Cleanup>
+inline auto defer(Cleanup&& cleanup)
+{ return std::experimental::scope_exit{std::forward<Cleanup>(cleanup)}; }
 }
```
AlanGriffiths, line 76: I think this would be clearer as:

```
template<typename T> struct defer : std::experimental::scope_exit<T>
{
    using std::experimental::scope_exit<T>::scope_exit;
};
```

But... do we need the indirection through `raii::defer<>` at all?

src/server/compositor/multi_threaded_compositor.cpp
```diff
@@ -106,8 +106,7 @@ class CompositingFunctor
     try
     {
         mir::set_thread_name("Mir/Comp");
-        auto const signal_when_stopped = mir::raii::paired_calls(
-            [](){},
+        auto const signal_when_stopped = mir::raii::defer(
```
AlanGriffiths:
```suggestion
        auto const signal_when_stopped = std::experimental::scope_exit(
```

src/server/console/linux_virtual_terminal.cpp
```diff
@@ -456,8 +456,7 @@ void mir::LinuxVirtualTerminal::register_switch_handlers(
                      * Otherwise we leave the VT subsystem in a glorious state of
                      * waiting forever for us to ACK the switch.
                      */
-                    auto ack_when_done = mir::raii::paired_calls(
-                        [](){},
+                    auto ack_when_done = mir::raii::defer(
```
AlanGriffiths:
```suggestion
                    auto ack_when_done = std::experimental::scope_exit(```

File: src/server/console/linux_virtual_terminal.cpp [hunk 4]
```diff
@@ -491,8 +490,7 @@ void mir::LinuxVirtualTerminal::register_switch_handlers(
...
```
AlanGriffiths:
```suggestion
                    auto ack_when_done = std::experimental::scope_exit(
```

AlanGriffiths (changes requested):
I am not sure of the utility of `raii::defer` - it doesn't really belong in the `raii` namespace, and is semantically identical to the library's `std::experimental::scope_exit`.

include/common/mir/raii.h
tarek-y-ismail: Maybe I was prematurely optimizing, but my thinking was that it'd save us work when `scope_guard` left the experimental namespace. Thought that could be a singular sed command... 

I don't think there'd be any overhead as it should be clear to the compiler that this is a simple wrapper.

AlanGriffiths: If/when `scope_exit` leaves the experimental namespace woudn't you also want to `s/mir::raii::defer/std::scope_exit`?

tarek-y-ismail: I concede.

AlanGriffiths (changes requested):
The code is fine. But the title and description are both wrong

AlanGriffiths (approved):

PR #5131, opened 2026-08-12
Fix late toplevel identity updates in app switcher

tarek-y-ismail:
## What's new?

- Keep foreign-toplevel handles without an initial title or app ID pending instead of permanently discarding them after `done`.
- Pending handles remain hidden and cannot be selected until their identity arrives. This lets Zed appear in the application switcher when Mir subsequently resolves `dev.zed.Zed`.

## How to test
On a live session:

- Run a shell with an app switcher, such as `miriway`.
- Launch Zed and press Alt+Tab.
- Observe Zed appears in the switcher. On `main`, it is absent despite Mir resolving its app ID after the initial foreign-toplevel `done`.

## Checklist

- ~[ ] Tests added and pass~ difficult to test

### DIFF
--- a/src/miral/application_switcher.cpp
+++ b/src/miral/application_switcher.cpp
@@ -28,7 +28,6 @@
 #include <mir/geometry/displacement.h>
 #include <mutex>
 #include <queue>
-#include <unordered_set>
 #include <sys/eventfd.h>
 #include <poll.h>
 #include <string>
@@ -52,8 +51,7 @@ struct ToplevelInfo
     std::string app_id;
     std::string window_title;
     /// True until the first app_id or title event is received from the compositor.
-    /// Ghost entries (done fired before any identity) are tracked but excluded from
-    /// display and navigation — they are a Mir bug: https://github.com/canonical/mir/issues/4944
+    /// Entries without an identity are excluded from display and navigation.
     bool ghost = true;
 };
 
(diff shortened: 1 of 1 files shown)


src/miral/application_switcher.cpp
```diff
@@ -463,8 +458,6 @@ class WaylandApp : public miral::tk::WaylandApp
     void add(zwlr_foreign_toplevel_handle_v1* toplevel)
     {
         toplevels_in_focus_order.push_back(ToplevelInfo{toplevel, "<unset>", "<unset>"});
-        if (!tentative_focus_index)
-            tentative_focus_index = 0;
```
mattkae, line 467: Curious as to why this is changed!

mattkae (approved):
One non-blocking question, but this looks reasonable to me.

src/miral/application_switcher.cpp
tarek-y-ismail: This is to prevent an infinite loop.

Consider the following case with a fresh session:
- Zed is started, but it hasn't set its app ID yet, so it's considered a ghost
  - Previously, this would set `tentative_focus_index = 0`
- If Alt+Tab is activated, the following loop would never terminate:
  ```
  do
  {
      (*tentative_focus_index)++;
      if (*tentative_focus_index == toplevels_in_focus_order.size())
          *tentative_focus_index = 0;
  } while (!is_first_toplevel_with_app_id(
      toplevels_in_focus_order[*tentative_focus_index]));
  ```
  
  as `is_first_toplevel_with_app_id` would always return `false` for a list of one ghost element.
  
Now, with `tentative_focus_index` unset, `next_app` would return early before hitting this loop, avoiding the infinite loop.

PR #5132, opened 2026-08-12
Handle unmanaged scene surfaces consistently

tarek-y-ismail:
Related: #5071

## What's new?

- Prevent crashes when scene surfaces are present but are not managed by the window manager.
- Preserve established behavior for managed and unmanaged surfaces, with regression coverage for lifecycle and lookup edge cases.

## How to test

- Build and run the MirAL internal tests:
  ```sh
  ./build/bin/miral-test-internal
  ```
- Run the server unit tests and the registered `miral-test` suite.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/include/server/mir/shell/system_compositor_window_manager.h
+++ b/src/include/server/mir/shell/system_compositor_window_manager.h
@@ -20,9 +20,6 @@
 #include <mir/shell/window_manager.h>
 #include <mir/graphics/display_configuration.h>
 
-#include <map>
-#include <mutex>
-
 namespace mir
 {
 namespace scene { class SessionCoordinator; }
@@ -104,10 +101,10 @@ private:
         MirInputEvent const* event,
         MirResizeEdge edge) override;
 
-    using OutputMap = std::map<std::weak_ptr<scene::Surface>, graphics::DisplayConfigurationOutputId, std::owner_less<std::weak_ptr<scene::Surface>>>;
+    auto managed_window_membership() const -> std::shared_ptr<ManagedWindowMembership const> override;
 
-    std::mutex mutable mutex;
-    OutputMap output_map;
+    std::shared_ptr<ManagedWindowRegistry<graphics::DisplayConfigurationOutputId>> const output_map{
+        std::make_shared<ManagedWindowRegistry<graphics::DisplayConfigurationOutputId>>()};
 };
 
 /**
--- a/src/include/server/mir/shell/window_manager.h
+++ b/src/include/server/mir/shell/window_manager.h
@@ -18,6 +18,7 @@
 #define MIR_SHELL_WINDOW_MANAGER_H_
 
 #include <mir/frontend/surface_id.h>
+#include <mir/shell/managed_window_membership.h>
 #include <mir_toolkit/common.h>
 #include <mir_toolkit/event.h>
 #include <mir/geometry/forward.h>
@@ -78,6 +79,8 @@ public:
         std::shared_ptr<scene::Surface> const& surface,
         MirInputEvent const* event,
         MirResizeEdge edge) = 0;
+
+    virtual auto managed_window_membership() const -> std::shared_ptr<ManagedWindowMembership const> = 0;
 
     virtual ~WindowManager() = default;
     WindowManager() = default;
--- a/src/miral/basic_window_manager.h
+++ b/src/miral/basic_window_manager.h
@@ -117,6 +117,8 @@ public:
         MirInputEvent const* event,
         ::MirResizeEdge edge) override;
 
+    auto managed_window_membership() const -> std::shared_ptr<mir::shell::ManagedWindowMembership const> override;
+
     auto create_workspace() -> std::shared_ptr<Workspace> override;
 
     void add_tree_to_workspace(Window const& window, std::shared_ptr<Workspace> const& workspace) override;
@@ -221,7 +223,6 @@ private:
         std::set<Window> attached_windows; ///< Maximized/anchored/etc windows attached to this area
     };
 
-    using SurfaceInfoMap = std::map<std::weak_ptr<mir::scene::Surface>, WindowInfo, std::owner_less<std::weak_ptr<mir::scene::Surface>>>;
     using SessionInfoMap = std::map<std::weak_ptr<mir::scene::Session>, ApplicationInfo, std::owner_less<std::weak_ptr<mir::scene::Session>>>;
 
     mir::shell::FocusController* const focus_controller;
@@ -237,11 +238,13 @@ private:
 
     std::shared_ptr<DeadWorkspaces> const dead_workspaces{std::make_shared<DeadWorkspaces>()};
 
+    std::shared_ptr<mir::shell::ManagedWindowRegistry<WindowInfo>> const window_info{
+        std::make_shared<mir::shell::ManagedWindowRegistry<WindowInfo>>()};
+
     std::unique_ptr<WindowManagementPolicy> const policy;
 
     std::mutex mutex;
     SessionInfoMap app_info;
-    SurfaceInfoMap window_info;
     mir::geometry::Rectangles outputs;
     mir::geometry::Point cursor;
     uint64_t last_input_event_timestamp{0};
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ set(CMAKE_GCOV gcov)
 
 project(
   Mir
-  VERSION 2.29.0
+  VERSION 2.30.0
   DESCRIPTION "Mir compositor library"
   LANGUAGES C CXX
 )
--- a/debian/control
+++ b/debian/control
@@ -56,7 +56,7 @@ Vcs-Git: https://github.com/canonical/mir
 
 #TODO: Packaging infrastructure for better dependency generation,
 #      ala pkg-xorg's xviddriver:Provides and ABI detection.
-Package: libmirserver68
+Package: libmirserver69
 Section: libs
 Architecture: linux-any
 Multi-Arch: same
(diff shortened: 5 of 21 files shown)


mattkae (commented):
This feels like a band-aid, so I must ask: why can't they also be registered?

I feel as though this check could potentially hide a future problem, so I'd be hesitant to put it in

tarek-y-ismail:
Should the window manager care about the magnifier? I don't think it matters at all. The magnifier is the same for floating, tiling, rolling, etc... window managers.

Come to think of it, the problem here is that the focus controller is returning a surface that shouldn't be focused:
```
                surf->set_focus_mode(mir_focus_mode_disabled);
```

I went ahead and investigated, and found another, maybe better solution:
```diff
@@ -689,7 +689,8 @@ bool msh::AbstractShell::handle(MirEvent const& event)
 auto msh::AbstractShell::surface_at(geometry::Point cursor) const
 -> std::shared_ptr<scene::Surface>
 {
-    return surface_stack->surface_at(cursor);
+    auto const surface = surface_stack->surface_at(cursor);
+    return surface->focus_mode() != mir_focus_mode_disabled ? surface : nullptr;
 }
```

I'm doing it in the abstract shell because that's the focus controller used in the basic window manager (everywhere, actually). Applying the same fix to the surface stack directly breaks input for the magnifier, as that is used in input handling.

AlanGriffiths:
> This feels like a band-aid, so I must ask: why can't they also be registered?

It is a band aid. The underlying problem is that `FocusController::surface_at()` is supporting multiple abstractions:

For input, it returns the surface that can accept events at a point and _until now_ that has worked for window management (as all surface that accept input were windows that can be managed). We've now introduced a type of surface that isn't managed by window management.

A more elegant approach would be to add a `FocusController::window_at()` that returns only surfaces for managed windows.

The harder question though, is should `FocusController::window_at()` map the input through the magnifier transform and return the window beneath?

tarek-y-ismail:
> The harder question though, is should FocusController::window_at() map the input through the magnifier transform and return the window beneath?

No. The focus controller should be separate from the magnifier. The magnifier should consume clicks and taps that land on its surface and emit new events with the remapped position. I had already prototyped this and found issues with the global input state, but those are workable.

tarek-y-ismail:
> A more elegant approach would be to add a FocusController::window_at() that returns only surfaces for managed windows.

How would the focus controller know which windows are managed and which are not? That either requires the window manager to tell the focus controller the windows it knows, or for the window manager to ignore windows it doesn't know about.

AlanGriffiths:
> the focus controller is returning a surface that shouldn't be focused

I wonder if there's a problem in legacy terminology... I can imagine surfaces that can accept input events, but that can't have focus.

tarek-y-ismail:
Here are my two cents:
- > The underlying problem is that FocusController::surface_at() is supporting multiple abstractions

  I disagree, the root problem may lie with the usage of `SurfaceStack::surface_at`. It returns _any_ surface that matches the criteria `surface_can_be_shown(surface) && surface->input_area_contains(cursor)`. 
  
  `surface_can_be_shown` is `return !is_locked || surface->visible_on_lock_screen();` so not entirely relevant to our discussion at the moment.
  
  This is fine, what's not fine is the usage at higher levels, where the surface is returned as-is. As I mentioned before, the focus controller should filter this so that only focus-able surfaces are returned. 
- On the input side `SurfaceStack::surface_at` is used as is with no filtration for `input_surface_at`
  This explains why filtering there breaks handles. As input routing calls this method.

So maybe `SurfaceStack::surface_at` should filtered more at usage sites, a la https://github.com/canonical/mir/pull/5132#issuecomment-5291222100

RAOF:
> > the focus controller is returning a surface that shouldn't be focused
> 
> I wonder if there's a problem in legacy terminology... I can imagine surfaces that can accept input events, but that can't have focus.

Indeed, layer shell surfaces by default cannot have keyboard focus, but do receive pointer and touch events.

We actually have at least *3* different “focus”ish concepts:
1. You have keyboard focus
2. The pointer is over you (this is what `window_at` is caring about)
3. You are the “active” window (`xdg_toplevel` state)

The focus controller should be caring about (1) (and possibly (3)); the focus controller maybe shouldn't be involved in (2) at all.

AlanGriffiths:
> * > The underlying problem is that FocusController::surface_at() is supporting multiple abstractions
>       
>       I disagree, the root problem may lie with the usage of `SurfaceStack::surface_at`. It returns _any_ surface that matches the criteria `surface_can_be_shown(surface) && surface->input_area_contains(cursor)`.

You're right, I didn't look at the actual hierarchy and named the wrong function.

Abstraction 1: `input::Scene::input_surface_at()` needs the behavior that exists now.

Abstraction 2: `shell::AbstractShell::surface_at()` (which implements `FocusController`) is doing the same thing, but for a different abstraction. We now have surfaces that can receive input but are not managed by the shell subsystem. These should be filtered out.

Both abstractions are supported by `SurfaceStack::surface_at()`, but they have different needs.

tarek-y-ismail:
The main issue is `BasicWindowManager::window_at` assuming all windows are managed, which was previously broken by the original magnifier implementation and fixed by disabling input for that surface. Now, for the freely-positioned implementation, we need input handling. 

Only the BWM knows what windows it manages. So we either push down that information (hard), or filter at the boundary (current solution in this PR).

The other issue we've stumbled on is `FocusController::surface_at` maybe mistakenly assuming all surfaces returned from `SurfaceStack::surface_at` are relevant. But I think this is a different thing that maybe requires its own dedicated issue. 

So my conclusion is this: we merge the current solution, and punt `FocusController::surface_at` to be looked at in more detail later on.

AlanGriffiths:
> The other issue we've stumbled on is `FocusController::surface_at` maybe mistakenly assuming all surfaces returned from `SurfaceStack::surface_at` are relevant. But I think this is a different thing that maybe requires its own dedicated issue.

I disagree. The `FocusController` interface is part of the `shell` subsystem which provides window management and should only return results that make sense in that context.

This is the implementation of `surface_at()`:

```c++
auto msh::AbstractShell::surface_at(geometry::Point cursor) const
-> std::shared_ptr<scene::Surface>
{
    return surface_stack->surface_at(cursor);
}
```

This is making the same call to the stack as is used by input, _but it is asking a different question_.

Input wants: which surface takes input at `cursor`?
WM wants: which managed window is at `cursor`?

Just because you can return the wrong thing here and patch it at the call site, doesn't make it the right approach

tarek-y-ismail:
Ok. Then the next step is to figure out if the abstract shell has access to that information, or pass it down.

`AbstractShell::{create,destroy}_surface` exist. We can use these to track and filter  surfaces that are unmanaged. What do you think? I just tried it locally, and it works. The only thing I'm worried about is the duplication of information between the BWM and the abstract shell now.

AlanGriffiths:
> duplication of information between the BWM and the abstract shell

Could the information be consolidated? Either as a shared object, or as part of the `Shell` implementation?

tarek-y-ismail:
> Could the information be consolidated? Either as a shared object, or as part of the Shell implementation?

Not straightforwardly. The abstract shell keeps a set just to know what surfaces to filter out. The BWM attaches extra info to the surface:
```
    auto const surface = build(session, make_surface_spec(spec));
    Window const window{session, surface};
    auto& window_info = this->window_info.emplace(window, WindowInfo{window, spec}).first->second;
```

tarek-y-ismail:
@AlanGriffiths @RAOF The new solution is a lot more involved but I believe is the correct one. Mind taking a look?

src/include/server/mir/shell/system_compositor_window_manager.h
```diff
@@ -104,10 +101,10 @@ class SystemCompositorWindowManager : public WindowManager
         MirInputEvent const* event,
         MirResizeEdge edge) override;
 
-    using OutputMap = std::map<std::weak_ptr<scene::Surface>, graphics::DisplayConfigurationOutputId, std::owner_less<std::weak_ptr<scene::Surface>>>;
+    auto managed_window_membership() const -> std::shared_ptr<ManagedWindowMembership const> override;
 
-    std::mutex mutable mutex;
-    OutputMap output_map;
+    std::shared_ptr<ManagedWindowRegistry<graphics::DisplayConfigurationOutputId>> const output_map{
+        std::make_shared<ManagedWindowRegistry<graphics::DisplayConfigurationOutputId>>()};
```
AlanGriffiths, line 107: This appears to be an unrelated usecase with a similar implementation. Why are we coupling these different pieces of code?

src/miral/basic_window_manager.h
```diff
@@ -237,11 +238,13 @@ class BasicWindowManager : public virtual mir::shell::WindowManager,
 
     std::shared_ptr<DeadWorkspaces> const dead_workspaces{std::make_shared<DeadWorkspaces>()};
 
+    std::shared_ptr<mir::shell::ManagedWindowRegistry<WindowInfo>> const window_info{
+        std::make_shared<mir::shell::ManagedWindowRegistry<WindowInfo>>()};
```
AlanGriffiths, line 242: This appears to be an unrelated usecase with a similar implementation. Why are we coupling these different pieces of code?

src/include/server/mir/shell/window_manager.h
```diff
@@ -79,6 +80,8 @@ class WindowManager
         MirInputEvent const* event,
         MirResizeEdge edge) = 0;
 
+    virtual auto managed_window_membership() const -> std::shared_ptr<ManagedWindowMembership const> = 0;
+
```
AlanGriffiths, line 84: Why are we not just adding `virtual auto contains(std::shared_ptr<scene::Surface> const& surface) const -> bool = 0` directly to `WindowManager`? With that we would not need to move the implementation(s) into another class.

AlanGriffiths (changes requested):
Maybe I am missing something, but this seems to do a lot more work than I expected

tarek-y-ismail:
As mentioned in the standup, the `contains` solution should deadlock, but after re-applying it again, it doesn't seem to do so. I'll look deeper into this.

Edit: Copilot added and used a different mutex...

tarek-y-ismail:
Closed in favor of #5187

PR #5133, opened 2026-08-12
miral: extract `SoftwareBufferPool`

tarek-y-ismail:

## What's new?

- Pull the software buffer allocation, recycling and pixel-writing logic out of `render_scene_into_surface.cpp` into a reusable `miral::SoftwareBufferPool`. No behaviour change; `RenderSceneIntoSurface` now delegates to the pool.

## How to test

```sh
./build/bin/miral-test-internal --gtest_filter='RenderSceneIntoSurface*'
```

## Checklist

- [x] Tests added and pass

Existing `RenderSceneIntoSurface` tests cover the refactored path; this is a pure extraction.

### DIFF
--- a/src/miral/software_buffer_pool.cpp
+++ b/src/miral/software_buffer_pool.cpp
@@ -1,0 +1,133 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "software_buffer_pool.h"
+#include "mir/fatal.h"
+
+#include <mir/compositor/stream.h>
+#include <mir/geometry/displacement.h>
+#include <mir/graphics/buffer.h>
+#include <mir/renderer/sw/pixel_source.h>
+#include <mir/report_exception.h>
+
+namespace mg = mir::graphics;
+
+namespace
+{
+/// Wraps a buffer and calls a release callback when destroyed, returning it
+/// to its owning SoftwareBufferPool.
+class ClaimedBuffer : public mir::graphics::Buffer
+{
+public:
+    ClaimedBuffer(std::shared_ptr<mir::graphics::Buffer> const& buffer, std::function<void()>&& release_callback) :
+        buffer(buffer),
+        release_callback(std::move(release_callback))
+    {}
+
+    ~ClaimedBuffer() override
+    {
+        try
+        {
+            release_callback();
+        }
+        catch (std::exception const& e)
+        {
+            MIR_FATAL_ERROR("Exception thrown while releasing a buffer back to its pool: {}", e.what());
+        }
+    }
+
+    ClaimedBuffer(ClaimedBuffer const&) = delete;
+    ClaimedBuffer& operator=(ClaimedBuffer const&) = delete;
+    ClaimedBuffer(ClaimedBuffer&&) = delete;
+    ClaimedBuffer& operator=(ClaimedBuffer&&) = delete;
+
+    std::unique_ptr<mir::renderer::software::Mapping<std::byte const>> map_readable() const override
+    { return buffer->map_readable(); }
+
+    mir::graphics::BufferID id() const override { return buffer->id(); }
+    mir::geometry::Size size() const override { return buffer->size(); }
+    MirPixelFormat pixel_format() const override { return buffer->pixel_format(); }
+    mir::graphics::NativeBufferBase* native_buffer_base() override { return buffer->native_buffer_base(); }
+
+private:
+    std::shared_ptr<mir::graphics::Buffer> const buffer;
+    std::function<void()> const release_callback;
+};
+}
+
+struct miral::SoftwareBufferPool::Self
+{
+    Self(BufferBuilder&& builder_func, size_t initial_size);
+    std::mutex mutex;
+    BufferBuilder builder;
+    std::vector<std::pair<bool, std::shared_ptr<mir::graphics::Buffer>>> buffers;
+};
+
+miral::SoftwareBufferPool::Self::Self(BufferBuilder&& builder_func, size_t initial_size) : builder(std::move(builder_func))
+{
+    for (size_t i = 0; i < initial_size; ++i)
+        buffers.emplace_back(false, builder());
+}
+
+miral::SoftwareBufferPool::SoftwareBufferPool(BufferBuilder&& builder_func, size_t initial_size) :
+    data(std::make_shared<Self>(std::move(builder_func), initial_size))
+{}
+
+std::shared_ptr<mg::Buffer> miral::SoftwareBufferPool::claim()
+{
+    std::lock_guard lock{data->mutex};
+    for (size_t i = 0; i < data->buffers.size(); ++i)
+    {
+        if (auto& [used, buffer] = data->buffers[i]; !used)
+        {
+            used = true;
+            std::weak_ptr weak_data = data;
+            return std::make_shared<ClaimedBuffer>(
+                buffer,
+                [weak_data, i]
+                {
+                    if (auto const locked = weak_data.lock())
+                    {
+                        std::lock_guard l{locked->mutex};
+                        locked->buffers[i].first = false;
+                    }
+                });
+        }
+    }
+
+    // All slots in use — grow the pool.
+    data->buffers.emplace_back(true, data->builder());
+    size_t const i = data->buffers.size() - 1;
+    std::weak_ptr weak_data = data;
+    return std::make_shared<ClaimedBuffer>(
+        data->buffers[i].second,
+        [weak_data, i]
+        {
+            if (auto const locked = weak_data.lock())
+            {
+                std::lock_guard l{locked->mutex};
+                locked->buffers[i].first = false;
+            }
+        });
+}
+
+void miral::SoftwareBufferPool::set_builder(BufferBuilder&& builder)
+{
+    std::lock_guard lock{data->mutex};
+    data->builder = std::move(builder);
+    for (auto& b : data->buffers)
+        b = {b.first, data->builder()};
+}
--- a/src/miral/software_buffer_pool.h
+++ b/src/miral/software_buffer_pool.h
@@ -1,0 +1,53 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIRAL_SOFTWARE_BUFFER_POOL_H
+#define MIRAL_SOFTWARE_BUFFER_POOL_H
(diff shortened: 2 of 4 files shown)


mattkae (approved):
Cool 👍

src/miral/software_buffer_pool.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+    auto map_readable() const -> std::unique_ptr<mir::renderer::software::Mapping<std::byte const>> override;
+    mir::graphics::BufferID id() const override;
+    mir::geometry::Size size() const override;
+    MirPixelFormat pixel_format() const override;
+    mir::graphics::NativeBufferBase* native_buffer_base() override;
+
+private:
+    std::shared_ptr<mir::graphics::Buffer> buffer;
+    std::function<void()> release_callback;
```
AlanGriffiths: Why did these lose `const` in the move?

src/miral/software_buffer_pool.cpp
```diff
@@ -0,0 +1,156 @@
...
+{
+    // A destructor must not let an exception escape, and the callback locks a
+    // mutex, which can throw.
+    try
+    {
+        release_callback();
+    }
+    catch (...)
+    {
+        mir::report_exception();
```
AlanGriffiths: This is probably not a good response: dump some debug information to cerr and carry on...

1. Something is seriously wrong - "this shouldn't happen";
2. Unless the program fails shortly afterwards nobody will notice the information;
3. If they do notice, the content is "debug" not "explanation/help" and likely to be ignored;
4. We're bypassing the logging infrastructure.

Consider following (or replacing) with a `MIR_FATAL_ERROR` that provides some context.

src/miral/software_buffer_pool.h
```diff
@@ -0,0 +1,67 @@
...
+    void set_builder(BufferBuilder&& builder);
+
+private:
+    struct Self
+    {
+        Self(BufferBuilder&& builder_func, size_t initial_size);
+        std::mutex mutex;
+        BufferBuilder builder;
+        std::vector<std::pair<bool, std::shared_ptr<mir::graphics::Buffer>>> buffers;
+    };
```
AlanGriffiths: No need to define this in the header

src/miral/software_buffer_pool.h
```diff
@@ -0,0 +1,67 @@
...
+
+    std::shared_ptr<Self> data;
```
AlanGriffiths:
```suggestion
    std::shared_ptr<Self> const data;
```

src/miral/software_buffer_pool.h
```diff
@@ -0,0 +1,67 @@
...
+};
+
+/// Creates a StreamInfo list containing a single AlwaysHasSubmittedBufferStream.
+std::list<mir::scene::StreamInfo> create_stream_info();
```
AlanGriffiths, line 56: There's no obvious reason for this to be in this header. What is the connection between "create_stream_info()" and "software_buffer_pool"?

src/miral/software_buffer_pool.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+namespace
+{
+/// A mc::Stream that always reports having a submitted buffer so that
+/// BasicSurface::generate_renderables() includes the surface in every frame.
+class AlwaysHasSubmittedBufferStream : public mir::compositor::Stream
+{
+public:
+    bool has_submitted_buffer() const override { return true; }
+};
```
AlanGriffiths: This, and `create_stream_info()` which uses it, do not seem logically connected with `SoftwareBufferPool` (They are only used by `miral::RenderSceneIntoSurface::Self::init()` which is in a different file entirely - putting them there would make more sense.)

src/miral/software_buffer_pool.cpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+auto ClaimedBuffer::map_readable() const -> std::unique_ptr<mrs::Mapping<std::byte const>>
+{ return buffer->map_readable(); }
+
+mg::BufferID ClaimedBuffer::id() const { return buffer->id(); }
+geom::Size ClaimedBuffer::size() const { return buffer->size(); }
+MirPixelFormat ClaimedBuffer::pixel_format() const { return buffer->pixel_format(); }
+mg::NativeBufferBase* ClaimedBuffer::native_buffer_base() { return buffer->native_buffer_base(); }
```
AlanGriffiths: [non blocking] I don't think it improves readability to move these trivial function definitions out of the class definition.  (I know this doesn't change the binary: the compiler will inline them anyway.)

src/miral/software_buffer_pool.cpp
```diff
@@ -0,0 +1,156 @@
...
+    ClaimedBuffer(ClaimedBuffer const&) = delete;
+    ClaimedBuffer& operator=(ClaimedBuffer const&) = delete;
+    ClaimedBuffer(ClaimedBuffer&&) = delete;
+    ClaimedBuffer& operator=(ClaimedBuffer&&) = delete;
+
+    auto map_readable() const -> std::unique_ptr<mir::renderer::software::Mapping<std::byte const>> override;
+    mir::graphics::BufferID id() const override;
+    mir::geometry::Size size() const override;
+    MirPixelFormat pixel_format() const override;
+    mir::graphics::NativeBufferBase* native_buffer_base() override;
```
AlanGriffiths: Mixture of new, preferred `auto ...` style and old `Type ...` function declaration. Be consistent

AlanGriffiths (changes requested):

src/miral/software_buffer_pool.cpp
tarek-y-ismail: This is required work for drawing done in #5135. `create_stream_info()` is required to construct `BasicSurface`s [usage](https://github.com/canonical/mir/blob/11cd0ed243b89558e2495db34cfea561c2ab5118/src/miral/magnifier_handle_indicator.cpp#L289-L296). And the pool is required to later submit buffers. They may not have a direct relationship, but they are used together.

Though `AlwaysHasSubmittedBufferStream` can be moved inside `create_stream_info`

src/miral/software_buffer_pool.cpp
tarek-y-ismail: I did not notice that. Thank you for pointing it out.

src/miral/software_buffer_pool.h
tarek-y-ismail: Pre-existing, but will fix.

src/miral/software_buffer_pool.cpp
tarek-y-ismail: I think this is some classic LLM "defensive programming". "Oh, the callback could be anything, better wrap it in a try-catch so the program doesn't abort". I think in this case it'd be better to crash and burn, Yes.

src/miral/software_buffer_pool.cpp
tarek-y-ismail: I think what happened is:
1. I moved all code to a header
2. It was too much, so I started moving implementations out
3. Figured out that `ClaimedBuffer` and `AlwaysHasSubmittedBufferStream` are internal, so I moved them in
4. Forgot to re-inline the methods

src/miral/software_buffer_pool.h
tarek-y-ismail: See https://github.com/canonical/mir/pull/5133/changes#r3914764673

tarek-y-ismail:
Moved `RenderSceneIntoSurface::capture_compositor_id` into #5135 where it's first used.

src/miral/software_buffer_pool.cpp
```diff
@@ -0,0 +1,152 @@
...
+    ~ClaimedBuffer() override
+    {
+        try
+        {
+            release_callback();
+        }
+        catch (std::exception const& e)
+        {
+            MIR_FATAL_ERROR("Exception thrown while releasing a buffer back to its pool: {}", e.what());
+        }
```
AlanGriffiths, line 49: [non blocking] We should probably create a formatter for `std::exception` that tries to extract the boost annotations

src/miral/software_buffer_pool.h
```diff
@@ -0,0 +1,60 @@
...
+    /// Replaces the buffer builder and rebuilds all existing buffers.
+    void set_builder(BufferBuilder&& builder);
+
+private:
+    struct Self;
+    std::shared_ptr<Self> const data;
+};
+
+/// Creates a StreamInfo list containing a single AlwaysHasSubmittedBufferStream.
+std::list<mir::scene::StreamInfo> create_stream_info();
```
AlanGriffiths: There's no obvious reason for this to be in this header. What is the connection between "create_stream_info()" and "software_buffer_pool"?

src/miral/software_buffer_pool.cpp
```diff
@@ -0,0 +1,152 @@
...
+    /// BasicSurface::generate_renderables() includes the surface in every frame.
+    class AlwaysHasSubmittedBufferStream : public mir::compositor::Stream
+    {
+    public:
+        bool has_submitted_buffer() const override { return true; }
+    };
+
+    result.push_back({std::make_shared<AlwaysHasSubmittedBufferStream>(), geom::Displacement{}});
+    return result;
+}
```
AlanGriffiths: There's no obvious reason for this to be in this file. There is no connection between "create_stream_info()" and `SoftwareBufferPool`. Why is it moved from `render_scene_into_surface.cpp‎` (The only place it is actually used)?

mattkae (approved):
Re-LGTM

AlanGriffiths (changes requested):

src/miral/software_buffer_pool.h
tarek-y-ismail: Have you seen my previous comment related to this? https://github.com/canonical/mir/pull/5133#discussion_r3914764673

AlanGriffiths: I'm being thick: I don't see how "`create_stream_info()` needs to be available" answers "why is it in this file?"

src/miral/software_buffer_pool.cpp
tarek-y-ismail: Spawned off a new issue.

src/miral/software_buffer_pool.h
tarek-y-ismail: Not opposed to moving it to a separate header just didn't see an immediate need for that.

AlanGriffiths: I don't see the need to move it at all. It started in the only file it is used in and you've moved it to another file

AlanGriffiths (changes requested):
This PR has no need to move `create_stream_info()` from the file in which it is used to an unrelated header & implementation.

tarek-y-ismail:
I think there's confusion over the purpose of this PR, which is my fault for not updating the various bits and pieces.

In #5135, the magnifier uses software rendering to render the drag, resize, and zoom handles. 

These are basic surfaces, which require a list of `StreamInfo` to be created. With an empty list, the surface would not generate renderables and wouldn't be visible. The existing implementation with `AlwaysHasSubmittedBufferStream` ensures that Mir's rendering code always generates a renderable from submitted buffers when rendering. 

On the other hand `SoftwareBufferPool` allows us to obtain buffers which we can submit for rendering.

The two pieces work together, rendering wouldn't work without either, hence why I'm including them in the same PR. The stream info implementation would be identical in most cases, hence why I pulled it to a shared header.

src/miral/basic_surface_stream.h
```diff
@@ -0,0 +1,44 @@
...
+inline auto create_stream_info() -> std::list<mir::scene::StreamInfo>
+{
+    class AlwaysHasSubmittedBufferStream : public mir::compositor::Stream
+    {
+    public:
+        bool has_submitted_buffer() const override { return true; }
+    };
+
+    return {{std::make_shared<AlwaysHasSubmittedBufferStream>(), mir::geometry::Displacement{}}};
+}
```
AlanGriffiths: Moving this function from an implementation file to wider visibility has implications. It is no longer an implementation detail of a specific context, it purports to be generally useful and have a clear role.

Because of this the doc comment doesn't seem right:

1. "Creates stream information" doesn't match the signature: it returns a `list`.
2.  "keeps a BasicSurface in every frame" it isn't about what the function promises (nor what it requires).

The name `create_stream_info()` seems too generic to be meaningful here. `create_always_submitted_streams()`? `create_default_streams()`?

The functionality seems weird:

1. We're creating a `BufferStream` that lies when it doesn't have a submitted buffer. I would expect that to lead to cases where a buffer is expected but none has been submitted.

2. This hard-codes the displacement: {0, 0} makes sense as a default, but for always?

Is there any advantage to having the implementation in a header file?

AlanGriffiths (changes requested):

AlanGriffiths (approved):

PR #5134, opened 2026-08-12
miral: add magnifier layout geometry helpers

tarek-y-ismail:
Related: #5071

## What's new?

- Add `miral::MagnifierLayout`, which owns all of the magnifier's geometry: given a requested visual size, a magnification and the set of outputs, it derives the captured region and the surface position, clamps both to the outputs, and places the control handles around the resulting bounds.
- Supported placements: centred on a point (cursor-following), freely positioned from a desired top-left, freely positioned centred on a point, and resize from a pinned corner.
- `HandleKind` is defined here because the enumerator value doubles as the index into `MagnifierLayout::HandlePositions`.
- Nothing consumes this yet; `Magnifier` is ported to it separately. Isolating the arithmetic makes it testable without standing up a server.

## How to test

```sh
./build/bin/miral-test-internal --gtest_filter='MagnifierLayout*'
```

7 tests, all passing.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/include/core/mir/geometry/rectangle.h
+++ b/include/core/mir/geometry/rectangle.h
@@ -82,6 +82,7 @@ struct Rectangle
     X<T> right() const { return bottom_right().x; }
     Y<T> top() const { return top_left.y; }
     Y<T> bottom() const { return bottom_right().y; }
+    Point<T> centre() const { return top_left + as_displacement(size / static_cast<T>(2)); }
 
     friend bool operator==(Rectangle const& lhs, Rectangle const& rhs) = default;
 
--- a/src/miral/magnifier_geometry.h
+++ b/src/miral/magnifier_geometry.h
@@ -1,0 +1,51 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIRAL_MAGNIFIER_GEOMETRY_H
+#define MIRAL_MAGNIFIER_GEOMETRY_H
+
+#include <mir/geometry/point.h>
+#include <mir/geometry/rectangle.h>
+#include <mir/geometry/size.h>
+
+namespace miral::magnifier_geometry
+{
+inline auto visual_bounds(
+    mir::geometry::Point const& surface_top_left,
+    mir::geometry::Size const& capture_size,
+    double const magnification) -> mir::geometry::RectangleD
+{
+    auto const center = mir::geometry::PointD{
+        surface_top_left.x.as_value() + capture_size.width.as_value() / 2.0,
+        surface_top_left.y.as_value() + capture_size.height.as_value() / 2.0,
+    };
+
+    auto const visual_size = mir::geometry::SizeD{
+        capture_size.width.as_value() * magnification,
+        capture_size.height.as_value() * magnification,
+    };
+
+    return {
+        {
+            center.x.as_value() - visual_size.width.as_value() / 2.0,
+            center.y.as_value() - visual_size.height.as_value() / 2.0,
+        },
+        visual_size,
+    };
+}
+}
+
+#endif
--- a/src/miral/magnifier_layout.cpp
+++ b/src/miral/magnifier_layout.cpp
@@ -1,0 +1,273 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "magnifier_layout.h"
+
+#include "magnifier_controls.h"
+#include "magnifier_geometry.h"
+
+#include <algorithm>
+#include <cmath>
+
+namespace geom = mir::geometry;
+namespace controls = miral::magnifier_controls;
+namespace layout = miral::magnifier_layout;
+using namespace miral::magnifier_geometry;
+
+namespace
+{
+template<typename RectangleType>
+inline auto as_rectangle_d(RectangleType rectangle) -> mir::geometry::RectangleD
+{
+    return {
+        mir::geometry::PointD{rectangle.top_left},
+        mir::geometry::SizeD{rectangle.size},
+    };
+}
+
+auto squared_distance_between(geom::PointD point, geom::Rectangle rectangle) -> double
+{
+    auto const rectangle_d = as_rectangle_d(rectangle);
+    auto const dx = std::max({
+        0.0,
+        (rectangle_d.left() - point.x).as_value(),
+        (point.x - rectangle_d.right()).as_value(),
+    });
+    auto const dy = std::max({
+        0.0,
+        (rectangle_d.top() - point.y).as_value(),
+        (point.y - rectangle_d.bottom()).as_value(),
+    });
+    return dx * dx + dy * dy;
+}
+
+auto selected_output(geom::Rectangles const& outputs, geom::PointD desired_center)
+    -> geom::Rectangle const&
+{
+    auto const first_containing = std::ranges::find_if(
+        outputs, [desired_center](auto const& output) { return as_rectangle_d(output).contains(desired_center); });
+
+    if (first_containing != outputs.end())
+        return *first_containing;
+
+    auto const nearest = std::ranges::min_element(
+        outputs,
+        std::less{},
+        [desired_center](auto const& output) { return squared_distance_between(desired_center, output); });
+
+    return *nearest;
+}
+
+auto clamp(geom::SizeD value, geom::SizeD min_value, geom::SizeD max_value) -> geom::SizeD
+{
+    return {
+        std::clamp(value.width, min_value.width, max_value.width),
+        std::clamp(value.height, min_value.height, max_value.height),
+    };
+}
+
+auto floor(geom::SizeD value) -> geom::Size
+{
+    return {
+        geom::Width{std::floor(value.width.as_value())},
+        geom::Height{std::floor(value.height.as_value())}};
+}
+
+auto capture_size_for(
+    geom::SizeD requested_visual_size,
+    geom::Rectangle output,
+    double magnification) -> geom::Size
+{
+    auto const maximum_visual = 0.8 * geom::SizeD{output.size};
+
+    auto const clamped_visual = clamp(requested_visual_size, geom::SizeD{controls::minimum_visual_size}, geom::SizeD{maximum_visual});
+    auto const requested_capture = geom::Size{clamped_visual / magnification};
+    auto const maximum_capture = floor(maximum_visual / magnification);
+    return {
+        std::min(requested_capture.width, maximum_capture.width),
+        std::min(requested_capture.height, maximum_capture.height)};
+}
+
+auto surface_top_left_centered_on(geom::PointD center, geom::Size capture_size) -> geom::Point
+{
+    return {
+        geom::X{center.x.as_value() - capture_size.width.as_value() / 2.0},
+        geom::Y{center.y.as_value() - capture_size.height.as_value() / 2.0}};
+}
+
+auto contains_point(geom::Rectangles const& outputs, double x, double y) -> bool
+{
+    return std::ranges::any_of(
+        outputs, [point = geom::PointD{x, y}](auto const& output) { return as_rectangle_d(output).contains(point); });
+}
+
+auto contains_corners(geom::Rectangles const& outputs, geom::RectangleD const& bounds) -> bool
+{
+    auto const left = bounds.left().as_value();
+    auto const right = bounds.right().as_value();
+    auto const top = bounds.top().as_value();
+    auto const bottom = bounds.bottom().as_value();
+    return contains_point(outputs, left, top) &&
+           contains_point(outputs, right, top) &&
+           contains_point(outputs, left, bottom) &&
+           contains_point(outputs, right, bottom);
+}
+
+auto confinement_for(geom::RectangleD const& bounds, geom::Rectangle output) -> geom::Displacement
+{
+    auto dx = 0;
+    auto const output_d = as_rectangle_d(output);
+    if (bounds.left() < output_d.left())
+        dx = static_cast<int>(std::ceil((output_d.left() - bounds.left()).as_value()));
+    else if (bounds.right() > output_d.right())
+        dx = static_cast<int>(std::floor((output_d.right() - bounds.right()).as_value()));
+
+    auto dy = 0;
+    if (bounds.top() < output_d.top())
+        dy = static_cast<int>(std::ceil((output_d.top() - bounds.top()).as_value()));
+    else if (bounds.bottom() > output_d.bottom())
+        dy = static_cast<int>(std::floor((output_d.bottom() - bounds.bottom()).as_value()));
+
+    return {dx, dy};
+}
+
+auto map_axis(
+    double clearance,
+    double visual_center,
+    double visual_extent,
+    int capture_extent,
+    int output_start,
+    int output_extent)
+{
+    auto const output_center = output_start + output_extent / 2.0;
+    auto const progress =
+        std::clamp((visual_center - output_center) / ((output_extent - visual_extent) / 2.0), -1.0, 1.0);
+    auto const capture_center = output_center + progress * ((output_extent - capture_extent) / 2.0 + clearance);
+    return static_cast<int>(capture_center - capture_extent / 2.0);
+}
+
+auto capture_top_left_for(
+    geom::RectangleD const& visual_bounds,
+    geom::Size capture_size,
+    geom::Rectangle output,
+    double magnification) -> geom::Point
+{
+    auto const clearance = std::ceil(controls::visual_edge_clearance / magnification);
+    auto const visual_center = visual_bounds.centre();
+    return {
+        map_axis(
+            clearance,
+            visual_center.x.as_value(),
+            visual_bounds.size.width.as_value(),
+            capture_size.width.as_value(),
+            output.left().as_value(),
+            output.size.width.as_value()),
+        map_axis(
+            clearance,
+            visual_center.y.as_value(),
+            visual_bounds.size.height.as_value(),
+            capture_size.height.as_value(),
+            output.top().as_value(),
+            output.size.height.as_value())};
+}
+
+// Places the magnifier capture area and confines its visual area when necessary.
+//
+// First, the visual area is calculated from the given surface top left and
+// capture size. If the visual area's corners are not all covered by the
+// outputs, the surface top left is adjusted to confine it to the selected output.
+// Then, the capture area is placed to center on the visual area, while
+// ensuring that it does not extend beyond the output more than a handle's diameter.
+auto free_placement(
+    geom::Rectangle output,
+    geom::Size capture_size,
+    geom::Point surface_top_left,
+    geom::Rectangles const& outputs,
+    double magnification) -> layout::FreePlacement
+{
+    auto bounds = miral::magnifier_geometry::visual_bounds(
+        surface_top_left, capture_size, magnification);
+    auto confinement = geom::Displacement{0, 0};
+    if (!contains_corners(outputs, bounds))
+    {
+        confinement = confinement_for(bounds, output);
+        surface_top_left += confinement;
+        bounds.top_left += geom::DisplacementD{confinement};
+    }
+
+    return {
+        layout::Placement{{capture_top_left_for(bounds, capture_size, output, magnification), capture_size}},
+        surface_top_left};
+}
+}
+
+auto layout::place_following_cursor(
+    geom::PointD cursor,
+    geom::SizeD requested_visual_size,
+    geom::Rectangles const& outputs,
+    double magnification) -> Placement
+{
+    auto const output = selected_output(outputs, cursor);
+    auto const capture_size = capture_size_for(requested_visual_size, output, magnification);
+    auto const surface_top_left = surface_top_left_centered_on(cursor, capture_size);
+
+    return {
+        .capture_area = {surface_top_left, capture_size},
+    };
+}
+
+auto layout::place_freely(
+    geom::PointD desired_center,
+    geom::SizeD requested_visual_size,
+    geom::Rectangles const& outputs,
+    double magnification) -> FreePlacement
+{
+    auto const output = selected_output(outputs, desired_center);
+    auto const capture_size = capture_size_for(requested_visual_size, output, magnification);
+    return free_placement(
+        output,
+        capture_size,
+        surface_top_left_centered_on(desired_center, capture_size),
+        outputs,
+        magnification);
+}
+
+auto layout::resize_freely(
+    geom::PointD dragged_visual_top_left,
+    geom::PointD pinned_visual_bottom_right,
+    geom::Rectangles const& outputs,
+    double magnification) -> FreePlacement
+{
+    auto const requested_visual_size = geom::SizeD{
+        (pinned_visual_bottom_right.x - dragged_visual_top_left.x).as_value(),
+        (pinned_visual_bottom_right.y - dragged_visual_top_left.y).as_value()};
+
+    auto const temp_x = dragged_visual_top_left.x.as_value() + pinned_visual_bottom_right.x.as_value();
+    auto const temp_y = dragged_visual_top_left.y.as_value() + pinned_visual_bottom_right.y.as_value();
+    auto const intended_center = geom::PointD{(temp_x) / 2.0, (temp_y) / 2.0};
+
+    auto const output = selected_output(outputs, intended_center);
+    auto const capture_size = capture_size_for(requested_visual_size, output, magnification);
+    auto const surface_top_left = geom::Point{
+        geom::X{
+            pinned_visual_bottom_right.x.as_value() -
+            (1.0 + magnification) * capture_size.width.as_value() / 2.0},
+        geom::Y{
+            pinned_visual_bottom_right.y.as_value() -
+            (1.0 + magnification) * capture_size.height.as_value() / 2.0}};
+
+    return free_placement(output, capture_size, surface_top_left, outputs, magnification);
+}
--- a/src/miral/magnifier_layout.h
+++ b/src/miral/magnifier_layout.h
@@ -1,0 +1,62 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef MIRAL_MAGNIFIER_LAYOUT_H
+#define MIRAL_MAGNIFIER_LAYOUT_H
+
+#include <mir/geometry/point.h>
+#include <mir/geometry/rectangle.h>
+#include <mir/geometry/rectangles.h>
+#include <mir/geometry/size.h>
+
+namespace miral::magnifier_layout
+{
+struct Placement
+{
+    mir::geometry::Rectangle capture_area;
+
+    bool operator==(Placement const&) const = default;
+};
+
+struct FreePlacement : Placement
+{
+    /// Top-left of the untransformed surface. Its size is always
+    /// `capture_area.size`.
+    mir::geometry::Point surface_top_left;
+
+    bool operator==(FreePlacement const&) const = default;
+};
+
+auto place_following_cursor(
+    mir::geometry::PointD cursor,
+    mir::geometry::SizeD requested_visual_size,
+    mir::geometry::Rectangles const& outputs,
+    double magnification) -> Placement;
+
+auto place_freely(
+    mir::geometry::PointD desired_center,
+    mir::geometry::SizeD requested_visual_size,
+    mir::geometry::Rectangles const& outputs,
+    double magnification) -> FreePlacement;
+
+auto resize_freely(
+    mir::geometry::PointD dragged_visual_top_left,
+    mir::geometry::PointD pinned_visual_bottom_right,
+    mir::geometry::Rectangles const& outputs,
+    double magnification) -> FreePlacement;
+}
+
+#endif
--- a/tests/miral/magnifier_layout.cpp
+++ b/tests/miral/magnifier_layout.cpp
@@ -1,0 +1,567 @@
+/*
+ * Copyright © Canonical Ltd.
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 or 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "magnifier_controls.h"
+#include "magnifier_geometry.h"
+#include "magnifier_layout.h"
+
+#include <mir/flags.h>
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include <array>
+#include <cmath>
+#include <tuple>
+
+namespace geom = mir::geometry;
+namespace controls = miral::magnifier_controls;
+namespace layout = miral::magnifier_layout;
+using namespace testing;
+
+namespace
+{
+enum class ExpectedEdge
+{
+    left = 1 << 0,
+    right = 1 << 1,
+    top = 1 << 2,
+    bottom = 1 << 3,
+};
+
+constexpr auto mir_enable_enum_bit_operators(ExpectedEdge edge) -> ExpectedEdge
+{
+    return edge;
+}
+
+using ExpectedEdges = mir::Flags<ExpectedEdge>;
+
+constexpr std::array test_magnifications{1.25, 1.5, 2.0, 3.5, 8.0};
+
+auto single_output(geom::Rectangle const& output = {{0, 0}, {800, 600}}) -> geom::Rectangles
+{
+    return {output};
+}
+
+auto surface_top_left(layout::Placement const& placement) -> geom::Point
+{
+    return placement.capture_area.top_left;
+}
+
+auto surface_top_left(layout::FreePlacement const& placement) -> geom::Point
+{
+    return placement.surface_top_left;
+}
+
+auto visual_bounds(auto const& placement, double const magnification) -> geom::RectangleD
+{
+    return miral::magnifier_geometry::visual_bounds(
+        surface_top_left(placement),
+        placement.capture_area.size,
+        magnification);
+}
+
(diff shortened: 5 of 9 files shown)


src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,347 @@
...
+    geom::PointF const& point,
+    Placement const& placement) const -> bool
+{
+    if (!contains(visual_bounds(placement), point))
+        return false;
+
+    return std::ranges::none_of(
+        handle_positions(placement),
+        [&point](auto const& handle)
+        { return contains({handle, {handle_diameter, handle_diameter}}, point); });
```
tarek-y-ismail: Since all handles are placed _inside_ the magnifier surface, we don't need to check the handles as well.

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3766315052_

tests/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+TEST(MagnifierLayout, chooses_output_using_visual_bounds)
+{
+    auto bounds = outputs();
+    bounds.add({{800, 0}, {400, 300}});
+    miral::MagnifierLayout const layout{bounds, {500, 500}, 2.0f};
+
+    auto const placement = layout.centered_on({790, 150});
+
+    EXPECT_THAT(placement.requested_visual_size, Eq(geom::Size{500, 480}));
```
tarek-y-ismail: This is incorrect. If we center the magnifier on (790, 150), with a size of (500, 500), this means the magnifier would be extending outside the second smaller screen on both the x and y axes. This is not possible via dragging but may be possible when the cursor is moved and then the magnifier is set to not follow it.

In this case, the magnifier should probably snap so that
1. The drag handle is visible
2. The magnifier fits whatever output it's mostly in

I believe that in the above case, the magnifier should snap so that its bottom right sits at the bottom right of the original (800, 600) output.

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3768778053_

src/miral/magnifier_layout.cpp
tarek-y-ismail: Correction to my earlier reply: this was wrong, and the check is staying.

I claimed the handle exclusion was dead code because the handles are scene surfaces (`class HandleIndicator : public mir::scene::BasicSurface`) stacked above the magnifier, so they would take input first. That fact is true but irrelevant. In #5071 `contains_content()` is reached from `point_is_on_magnifier_surface()`, which is called by an `mi::EventFilter` installed with `the_composite_event_filter()->prepend(input_filter)`. Event filters run *before* `SurfaceInputDispatcher` offers the event to any surface, so scene stacking never gets a say: reporting content under a handle makes the magnifier swallow that handle's clicks.

The exclusion is therefore restored, with a comment recording why stacking does not save us, and `content_hit_testing_excludes_controls` now probes the centre of all four handles via `HandlePositions::for_kind()`. The version I had briefly replaced it with only asserted the visual bounds, so it passed with either implementation and could not have caught this.

tests/miral/magnifier_layout.cpp
tarek-y-ismail: Fixed, though not quite as I first described - correcting that here.

`confined_surface_position()` now skips confinement only when every corner of the visual bounds lies on some output (`is_fully_on_outputs()`), instead of whenever the bounds overlap more than one output. A magnifier straddling the shared edge of two outputs is still left alone; one that hangs over empty space is pulled back onto the output it mostly occupies.

My earlier reply said follow-cursor placement was confined the same way. That has been reverted: `place_following_cursor_at()` is deliberately left unconfined, because tracking the cursor into a screen corner matters more than keeping the magnifier wholly on screen, and `follow_cursor_is_not_confined_to_an_output` now pins that. The snapping you asked for happens on the free-placement path, which is the case your comment described - cursor tracked to a position where the magnifier fits no output, following then switched off - covered by `snaps_back_onto_an_output_when_it_fits_on_none` via `place_freely_centered_on({790, 150})`.

For your example the result is `{{300, 0}, {500, 480}}`: the height is 480 rather than 500 because of the existing 80%-of-output cap, so the bottom right lands at (800, 480) rather than (800, 600), but the right edge is flush with the output and the drag handle is on screen.

Separately, `chooses_output_using_visual_bounds` moved from cursor (790, 150) to (810, 150). At the old position the capture rect and the visual bounds both select the large output, so the test could not fail for the reason its name claims; at (810, 150) they disagree. Verified by mutation: making `clamped_visual_size_at()` select on the capture rect fails that test and no other.

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,466 @@
...
+/// stack to clear the drag handle in the bottom right.
+geom::Width const minimum_visual_width{2 * miral::MagnifierLayout::handle_diameter};
+geom::Height const minimum_visual_height{
+    zoom_stack_height + geom::DeltaY{miral::MagnifierLayout::handle_diameter}};
+auto constexpr maximum_capture_dimension{1000};
+
+/// The visible rectangle of a surface of `capture_size` placed at
+/// `surface_top_left`, given that the compositor scales it about its centre.
+/// The single home of the inset arithmetic described in the header.
+auto visual_rect_for(geom::Point surface_top_left, geom::Size capture_size, float magnification)
```
RAOF: Would this be better called `scale_around_centre`? That's what it's doing, right?

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,466 @@
...
+        surface_top_left -
+            geom::Displacement{
+                geom::as_delta(inset * capture_size.width),
+                geom::as_delta(inset * capture_size.height)},
+        capture_size * magnification};
+}
+
+/// Inverse of visual_rect_for(), solving for the surface origin that puts the
+/// visible rectangle's top left where the caller wants it.
+auto surface_top_left_for_visual_top_left(
```
RAOF: This is actually `scale_around_centre` again, just with `1/magnification`.

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,466 @@
...
+    geom::Point visual_top_left, geom::Size capture_size, float magnification) -> geom::Point
+{
+    auto const inset = (magnification - 1.0f) / 2.0f;
+    return visual_top_left +
+           geom::Displacement{
+               geom::as_delta(inset * capture_size.width),
+               geom::as_delta(inset * capture_size.height)};
+}
+
+auto center(geom::Point top_left, geom::Size size) -> geom::Point
```
RAOF: This seems like it should be an accessor on `Rectangle<T>`?

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,466 @@
...
+/// so the cheap test is the useful one.
+auto is_fully_on_outputs(geom::Rectangles const& outputs, geom::Rectangle const& bounds) -> bool
+{
+    auto const covered = [&outputs](geom::Point point)
+    {
+        return std::ranges::any_of(outputs, [point](auto const& output) { return output.contains(point); });
+    };
+
+    auto const right = bounds.right() - geom::DeltaX{1};
+    auto const bottom = bounds.bottom() - geom::DeltaY{1};
```
RAOF: Why do we take a 1px bite out of these?

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,466 @@
...
+    //                     output_start          output_start + output_extent
+    //                     v                                                v
+    // output              ==================================================
+    // magnifier           {visual_extent}                  {visual_extent}
+    //                     |<------- visual_travel -------->|
+    // capture       {capture_extent}                                {capture_extent}
+    //               |<---------- capture_travel ------------------->|
+    //               ^
+    //               capture_start = output_start - out_of_bounds
+    //
```
RAOF: This diagram is weird, because it seems to imply the visual extent is smaller than the capture extent, which would be the opposite of magnification?

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,466 @@
...
+}
+
+auto miral::MagnifierLayout::capture_size_for(geom::Size visual_size) const -> geom::Size
+{
+    return {
+        geom::Width{std::max(
+            1.0f, std::round(static_cast<float>(visual_size.width.as_value()) / magnification))},
+        geom::Height{std::max(
+            1.0f, std::round(static_cast<float>(visual_size.height.as_value()) / magnification))}};
+}
```
RAOF: This is just `visual_size / magnification`, right?

It'd probably be simpler to handle the pathological cases at the interface, rather than defensively clamping to a minimum of 1×1 sample box (which would be useless!)

src/miral/magnifier_layout.h
```diff
@@ -0,0 +1,172 @@
...
+/// be handed to `Surface::move_to()`.
+///
+/// Three sizes are likewise distinct, and conflating them is a bug:
+///
+///   * preferred visual size - persistent user intent. Only ever constrained
+///     by permanent limits (the minimum that fits the controls). Never
+///     narrowed by whichever output the magnifier happens to sit on, so that
+///     moving to a small output and back again is lossless.
+///   * clamped visual size   - the preferred size cut down to fit the current
+///     output. Transient; recomputed for every placement.
```
RAOF: Do we actually need this?

RAOF (commented):
So, I *think* this is rather more complex than it needs to be, but I can't prove it to myself.

src/miral/magnifier_layout.cpp
tarek-y-ismail: Yup, does seem a little clearer.

src/miral/magnifier_layout.cpp
tarek-y-ismail: Yes, though this returns only the top left, not a rect. Let's try `shrink_point_around_centre`.

src/miral/magnifier_layout.cpp
tarek-y-ismail: Yup. And maybe other parts of the code can reuse it.

src/miral/magnifier_layout.cpp
tarek-y-ismail: According to `rectangle.h`
```cpp
    /**
     * The bottom right boundary point of the rectangle.
     *
     * Note that the returned point is *not* included in the rectangle
     * area, that is, the rectangle is represented as [top_left,bottom_right).
     */
    Point<T> bottom_right() const { return top_left + as_displacement(size); }
```

So when we're checking the corners against the passed rectangles, we should inset the bottom and right so we're checking the correct points _in the rectangle_, not the edges.

src/miral/magnifier_layout.cpp
tarek-y-ismail: Yeah, LLMs have been quite defensive in this file in particular.

src/miral/magnifier_layout.cpp
tarek-y-ismail: The visual extent _is_ smaller than the capture extent. This is because the visual size is _bigger_ than the capture size. 

Look at it this way. Both have the same initial extent, then the size of the corresponding rectangle is subtracted from each. Since the visual rectangle is larger, it takes out a bigger chunk. Finally, the capture extent is offset slightly to make moving the capture near the edge of the screen capture a bit past the edge so that handles don't occlude content near the edges.

src/miral/magnifier_layout.h
tarek-y-ismail: Yes, this is to account for the case where we move between outputs.

Think about this. The user has two outputs, 800x600 and 400x300. The user starts the magnifier and scales it to the maximum 80% on the first output, yielding a surface of 640x480. In this case, the preferred visual size is about equal to the clamped visual size. 

They then move the magnifier to the 400x300 output where it doesn't fit. So we clamp 640x480 to 80% of 400x300. The preferred visual size remains 640x480. But the clamped visual size becomes 320x240.

Moving back to the first output, the magnifier will snap back to 640x480.

src/miral/magnifier_layout.cpp
RAOF: But there's no reason it *couldn't* return a rect, and the code using it just pull out `top_left()`?

src/miral/magnifier_layout.h
RAOF: Yeah, but is this a thing that we actually need to support? We certainly need to handle the “clamp things to a maximum/minimum size”, but do we need the extra “bounce back” bit (indeed, could we get away with just limiting to the maximum size on the *smallest* output?)

tarek-y-ismail: We _could_ clamp to the maximum size on the smallest output, yes. But I don't think it would be any simpler than what we're doing at the moment.

src/miral/magnifier_layout.cpp
tarek-y-ismail: It can, but why do more work and then discard it? Unless you're thinking for just merging the two functions and making the scaling factor a parameter.

RAOF: Yeah, I'm thinking of merging the two functions; current `surface_top_left_for_visual_top_left(point, size, mag)` calls would become `scale_around_center(point, size, 1/mag).top_left()`

src/miral/magnifier_layout.cpp
RAOF: Ah, right. A rectangle exactly the same size and location as the output should be counted as on the output.

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,454 @@
...
+
+/// True when every corner of `bounds` lies on some output, i.e. nothing of the
+/// magnifier hangs over empty space. A magnifier straddling the shared edge of
+/// two outputs satisfies this and is left alone; one hanging off the desktop
+/// does not, and is pulled back.
+///
+/// Corners only: on an L-shaped desktop a rectangle can have all four corners
+/// covered while a bite out of one edge sits over the notch. Confining such a
+/// placement would be no better - there is no single output it belongs to -
+/// so the cheap test is the useful one.
```
RAOF: This comment is a little misleading (you need a U-shaped desktop before a grid-aligned rectangle can have a portion of a rectangle with all four corners on an output but some of the rectangle not be within the desktop).

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,454 @@
...
+        if (overlap_area > maximum_overlap || (maximum_overlap == 0 && distance < minimum_distance))
+        {
+            maximum_overlap = overlap_area;
+            minimum_distance = distance;
+            current = output;
+        }
+    }
+
+    return current;
+}
```
RAOF: I think this is unnecessarily complex; the magnifier is most on the output that its centre is on. (This is cheap, and is almost always going to find an output).

Then, if the centre is not on any output, find the minimum distance between the centre and any output corner; the magnifier is on that output.

src/miral/magnifier_layout.cpp
RAOF: Ah, I think the names are confusing me.

I *think* there's a clearer algorithm here, but I need to write it down to be sure; I'll drop the current comments now and get back to this tomorrow.

src/miral/magnifier_layout.cpp
tarek-y-ismail: I've spent some time trying poking at this. 

The root of the issue is the integer capture size. 

First, formulas used below:
```
capture_width =
    truncate(requested_visual_width / magnification)

actual_visual_width =
    truncate(capture_width × magnification)

capture_inset =
    truncate(((magnification - 1) / 2) × capture_width)

surface_left = visual_left + capture_inset
visual_left  = surface_left - capture_inset

reciprocal_inset =
    truncate(((1 - 1 / magnification) / 2) × requested_visual_width)
```

With one merged function using reciprocal scaling, one division pass through the function would truncate, losing some information. The next multiplication pass would then multiply that truncated number, which wouldn't restore the original. For example:

1. Scale  215px  by  1/8 :
    capture width = 215 / 8 → 26
    reverse inset = 94
    surface left  = 100 + 94 = 194
2. Scale that  26px  capture forward by  8 :
    visual width  = 26 × 8 = 208
    forward inset = 91
    visual left   = 194 - 91 = 103

The 94 and -91 insets don't cancel out, and the left edge moves by 3 pixels. Also note that the final size is 208, not 215.

Having two separate functions instead use the same capture size keeps calculations more consistent (although still lossy)

0. capture width = 215 / 8 → 26

1. Find the surface position:
    inset            = 91
    surface left = 100 + 91 = 191
2. Calculate its visual bounds:
    visual left = 191 - 91 = 100
    visual width = 26 × 8 = 208
    
The 91 and -91 insets cancel out. The size still ends up as 208.

The proper solution would be to make `BasicScreenShooter` accept a float source rectangle, but that may cascade to other places, so the current solution seems like the best we can do at the moment.

include/core/mir/geometry/rectangle.h
```diff
@@ -82,6 +82,7 @@ struct Rectangle
     X<T> right() const { return bottom_right().x; }
     Y<T> top() const { return top_left.y; }
     Y<T> bottom() const { return bottom_right().y; }
+    Point<T> centre() const { return top_left + as_displacement(size / static_cast<T>(2)); }
```
mattkae, line 85: Nit, since Canonical uses American English 🙃 :

```suggestion
    Point<T> center() const { return top_left + as_displacement(size / static_cast<T>(2)); }
```

src/miral/magnifier_layout.h
```diff
@@ -0,0 +1,172 @@
...
+#include <mir/geometry/rectangle.h>
+#include <mir/geometry/rectangles.h>
+#include <mir/geometry/size.h>
+
+/// Geometry of the magnifier.
+///
+/// Three rectangles are in play, all in the same global screen coordinates.
+/// Telling them apart is the key to reading this file:
+///
+///   1. capture area  - the region read out of the scene. Small.
```
mattkae:
```suggestion
///   1. capture area  - the region read out of the scene.
```

src/miral/magnifier_layout.h
```diff
@@ -0,0 +1,172 @@
...
+///   2. surface rect  - where the surface is placed before the compositor
+///                      scales it. Same size as the capture area.
+///   3. visual bounds - what the user actually sees: the surface scaled by
```
mattkae: Right?

```suggestion
///   3. visual bounds - what the user actually sees: the `surface_rect` scaled by
```

src/miral/magnifier_layout.h
```diff
@@ -0,0 +1,172 @@
...
+
+#ifndef MIRAL_MAGNIFIER_LAYOUT_H
+#define MIRAL_MAGNIFIER_LAYOUT_H
+
+#include <mir/geometry/point.h>
+#include <mir/geometry/rectangle.h>
+#include <mir/geometry/rectangles.h>
+#include <mir/geometry/size.h>
+
+/// Geometry of the magnifier.
```
mattkae: Should this comment move to the top of the file or maybe on top of the `class` that it's annotating instead of the namespace?

src/miral/magnifier_layout.h
```diff
@@ -0,0 +1,172 @@
...
+/// Hence, for a capture area of `size`:
+///
+///     visual size      = size * magnification
+///     visual top left  = surface top left - (magnification - 1) / 2 * size
+///     surface top left = visual bottom right - (magnification + 1) / 2 * size
+///
+/// Everything the user points at - handles, hit testing, dragging and resizing
+/// - is expressed in **visual** coordinates.
+/// `Placement::untransformed_surface_top_left` is an output only: it exists to
+/// be handed to `Surface::move_to()`.
```
mattkae: I don't understand what this phrase means in this context. Can we clear this up? I would expect the doc comment to be an intro into how things are working

src/miral/magnifier_layout.h
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Follow-cursor placement: captures exactly the region it covers, and is
+    /// deliberately not confined to an output so that the magnifier can track
+    /// the cursor into a screen corner.
+    auto place_following_cursor_at(mir::geometry::Point cursor) const -> Placement;
+
+    /// Freely positioned placement: confined to an output, with the captured
+    /// region mapped proportionally across that output.
+    auto place_freely_at(mir::geometry::Point visual_top_left) const -> Placement;
+    auto place_freely_centered_on(mir::geometry::Point center) const -> Placement;
```
mattkae: Why have both of these methods? Should they be separately documented?

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,445 @@
...
+#include <optional>
+#include <ranges>
+#include <utility>
+
+namespace geom = mir::geometry;
+
+// Definitions are ordered so that helpers precede their callers: free
+// functions, then Placement, then MagnifierLayout's private helpers, then its
+// public entry points.
+
```
mattkae:
```suggestion
```

src/miral/magnifier_layout.h
```diff
@@ -0,0 +1,172 @@
...
+    /// derived fields stay consistent with this layout.
+    auto placement_of(
+        mir::geometry::Rectangle capture_area,
+        mir::geometry::Point untransformed_surface_top_left) const -> Placement;
+
+private:
+    auto place_freely_at_visual(
+        mir::geometry::Point visual_top_left,
+        mir::geometry::Size preferred) const -> Placement;
+    auto capture_size_for(mir::geometry::Size visual_size) const -> mir::geometry::Size;
```
mattkae: This feels more like a helper method that you pass the magnification to in the implementation file.

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,445 @@
...
+    return {
+        .capture_area = capture_area,
+        .untransformed_surface_top_left = untransformed_surface_top_left,
+        .preferred_visual_size = preferred_visual_size,
+        .magnification = magnification};
+}
+
+auto miral::MagnifierLayout::place_following_cursor_at(geom::Point cursor) const -> Placement
+{
+    auto const provisional = top_left_centered_on(cursor, capture_size_for(preferred_visual_size));
```
mattkae: What would you say the word `provisional` means here?

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,445 @@
...
+/// Squared distance between a point and rectangle.
+auto squared_distance_between(geom::Point const& p, geom::Rectangle const& r) -> long long
+{
+    auto const dx = std::max({0, r.left().as_value() - p.x.as_value(), p.x.as_value() - r.right().as_value()});
+    auto const dy = std::max({0, r.top().as_value() - p.y.as_value(), p.y.as_value() - r.bottom().as_value()});
+    return static_cast<long long>(dx) * dx + static_cast<long long>(dy) * dy;
+}
+
+/// The output the magnifier is considered to be on is either the one
+/// containing its center, or the closest one to its center. Callers only use
```
mattkae: Is there a situation where we're putting the magnifier on a non-output? Would that just be a precondition gone awry?

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,445 @@
...
+
+auto clamp_to_output(geom::Size size, geom::Size output_size) -> geom::Size
+{
+    auto const maximum_size = output_size * 0.8;
+    return {std::min(size.width, maximum_size.width), std::min(size.height, maximum_size.height)};
+}
+
+auto is_usable(geom::Rectangle const& output) -> bool
+{
+    return output.size.width > geom::Width{0} && output.size.height > geom::Height{0};
```
mattkae: In practice, this does not happen, right? I also feel like this can be a precondition of this class, rather than having the complication of a filter.

mattkae (changes requested):
I think a `class` is simply the wrong encapsulating type for the `MagnifierLayout`, so it is making reasoning about this kind of difficult. Holding that state is kind of funky in my opinion.

For example, how someone would _use_ an API as defined by this class is pretty unclear, since it's essentially a wrapper around a bunch of algorithms.

I think that this could be cleaned up by making the following changes:

1. Have `MagnifierLayoutInfo` be a simple struct that holds the desired magnification, size, and list of outputs. Don't define any methods on the struct.
2. Pass it whenever you need to do a placement of some sort
3. Maybe define the methods to resize and recalculate a `Placement` on the `Placement` itself

That's just a recommendation, but I think its worth thinking about :)

src/miral/magnifier_layout.cpp
tarek-y-ismail: One example is an output being disconnected while the magnifier is on it.

src/miral/magnifier_layout.cpp
tarek-y-ismail: "Provisional" as in a "attempted" placement. We place the magnifier, clamp its size, then modify the top left to take this potentially modified size into account by shifting the top left.

src/miral/magnifier_geometry.h
```diff
@@ -0,0 +1,68 @@
...
+    };
+}
+
+template<typename RectangleType>
+inline auto contains(RectangleType const& rectangle, mir::geometry::PointD const& point) -> bool
+{
+    auto rectangle_d = as_rectangle_d(rectangle);
+    return point.x >= rectangle_d.left() && point.x < rectangle_d.right() && point.y >= rectangle_d.top() &&
+           point.y < rectangle_d.bottom();
+}
```
mattkae: `mir::geometry::Rectangle` already has `contains`.

src/miral/magnifier_geometry.h
```diff
@@ -0,0 +1,68 @@
...
+namespace miral::magnifier_geometry
+{
+template<typename RectangleType>
+inline auto as_rectangle_d(RectangleType const& rectangle) -> mir::geometry::RectangleD
+{
+    return {
+        mir::geometry::PointD{rectangle.top_left},
+        mir::geometry::SizeD{rectangle.size},
+    };
+}
```
mattkae: Maybe this would live better in `rectangle.h`?

src/miral/magnifier_geometry.h
```diff
@@ -0,0 +1,68 @@
...
+inline auto contains(RectangleType const& rectangle, mir::geometry::PointD const& point) -> bool
+{
+    auto rectangle_d = as_rectangle_d(rectangle);
+    return point.x >= rectangle_d.left() && point.x < rectangle_d.right() && point.y >= rectangle_d.top() &&
+           point.y < rectangle_d.bottom();
+}
+
+inline auto visual_bounds(
+    mir::geometry::Point const& surface_top_left,
+    mir::geometry::Size const& capture_size,
```
mattkae, line 28: Just move this (and whatever other helper functions that we want) to magnifier_layout.cpp. No reason to expose them generally

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+    if (first_containing != outputs.end())
+        return *first_containing;
+
+    auto const nearest = std::ranges::min_element(
+        outputs,
+        std::less{},
+        [&desired_center](auto const& output) { return squared_distance_between(desired_center, output); });
+
+    return *nearest;
```
mattkae, line 71: Nit: I still think that this is dead code in practice, but I suppose it's fine as a backup.

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,269 @@
...
+        std::min(requested_capture.width, maximum_capture.width),
+        std::min(requested_capture.height, maximum_capture.height)};
+}
+
+auto capture_size_for(
+    geom::SizeD const& requested_visual_size,
+    geom::Rectangle const& output,
+    double const magnification) -> geom::Size
+{
+    return clamp_size(
```
mattkae: `clamp_size` is only used here. I would rather see single-use functions be inlined so that they're easier to digest.

src/miral/magnifier_layout.cpp
mattkae, line 1: Nit: `const&` could probably be used a bit more here to reduce copies between functions

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+auto capture_top_left_for(
+    geom::RectangleD const& visual_bounds,
+    geom::Size const& capture_size,
+    geom::Rectangle const& output,
+    double const magnification) -> geom::Point
+{
+    auto const clearance = std::ceil(controls::visual_edge_clearance / magnification);
+
+    auto const map_axis = [clearance](
```
mattkae: Nit: I would just pass clearance, and maybe make this a free function like all of the rest of them.

src/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,269 @@
...
+{
+    auto bounds = miral::magnifier_geometry::visual_bounds(
+        surface_top_left, capture_size, magnification);
+    auto confinement = geom::Displacement{0, 0};
+    if (!is_covered(outputs, bounds))
+    {
+        confinement = confinement_for(bounds, output);
+        surface_top_left += confinement;
+        bounds.top_left += geom::DisplacementD{confinement};
+    }
```
mattkae, line 209: It would be nice to have a comment explaining the algorithm here.

Also - the `is_covered` method is kind of funky in my opinion. You're basically asking "is the entire visual bounds of the rectangle inside of the visual bounds of the collective outputs". `is_covered` is maybe a bad name for that.

mattkae (commented):
Overall - much more readable, thanks!

I have some feedback!

src/miral/magnifier_geometry.h
tarek-y-ismail: Maybe, but we should leave it here until it's needed elsewhere.

src/miral/magnifier_geometry.h
tarek-y-ismail: Good observation. Thank you. Though it doesn't apply to `visual_bounds` as it's used in tests as well.

src/miral/magnifier_layout.cpp
tarek-y-ismail: It's the opposite actually. Many of the structs here are small-ish, around 16 bytes at most, so they should be passable fine within registers. Only `Rectangles` really needs the reference.

src/miral/magnifier_layout.cpp
tarek-y-ismail: The magnifier being placed on an output, then that output being disconnected is a very real use case.

For example, you may have your laptop connected to an external monitor, then you disconnect it as you take the laptop somewhere else. If we don't move the magnifier surface somewhere else, it will be permanently stuck off-screen.

src/miral/magnifier_layout.cpp
tarek-y-ismail: Good observation. Thank you.

src/miral/magnifier_layout.cpp
tarek-y-ismail: `clearance` is the same for both invocations, hence why it's captured. This also fits as an inner function since it's only used here.

src/miral/magnifier_layout.cpp
tarek-y-ismail: Comment done. Renamed `is_covered` to `contains_corners`. I also looked into `Rectangles::confine` but that doesn't fit our use case.

tarek-y-ismail:
History was a bit too tangled with previous versions of #5133. Should be cleaner now. Though I want to clean up the history a bit more before merging. Sorry @mattkae!

include/core/mir/geometry/rectangle.h
tarek-y-ismail: Eh... Too nitty :skull:

src/miral/magnifier_layout.cpp
mattkae: Wouldn't the cursor origin automatically update to some point on screen after that happens though? It would be odd if we left the cursor offscreen then.

src/miral/magnifier_layout.cpp
mattkae: That "only used once = inner function" rule is not strictly followed in this file. For example, `contains_corners` is only called once from what I can see, but it is a free function.

My feeling is that you can make this code more similar to the outer style by making this a free function that takes `clearance` as a parameter. This is the only spot in the file where you opt for a lambda currently.

src/miral/magnifier_layout.cpp
mattkae: At any rate, this is non-blocking :)

src/miral/magnifier_geometry.h
mattkae: Any reason why it needs its own file though? Can we just put it in magnifier_controls.h? Less files is better :)

tests/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,420 @@
...
+}
+
+void expect_center_near(geom::PointD const& actual, geom::PointD const& expected, double const tolerance = 1.0)
+{
+    EXPECT_THAT(actual.x.as_value(), DoubleNear(expected.x.as_value(), tolerance));
+    EXPECT_THAT(actual.y.as_value(), DoubleNear(expected.y.as_value(), tolerance));
+}
+}
+
+TEST(MagnifierLayout, following_cursor_couples_capture_and_surface_with_bounded_center_error)
```
mattkae: This should be a parameterized test (i.e. `TEST_P`). That makes it easier to find which combination is failing in CI when it does.

Unless it is supposed to show a cursor changing/moving over time? I can't tell from the test's name if so.

src/miral/magnifier_layout.cpp
tarek-y-ismail: Ah, I see. I think you're not taking the free mode into account. That has its own position that's separate from the cursor.

tests/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,420 @@
...
+    expect_center_near(surface_center(placement), {400.0, 300.0}, 0.0);
+    expect_center_near(capture_center(placement), surface_center(placement), 0.0);
+}
+
+TEST(MagnifierLayout, edge_and_corner_mapping_preserves_handle_width_of_source_clearance)
+{
+    auto const outputs = single_output();
+    auto const requested_size = geom::SizeD{240.0, 200.0};
+
+    for (auto const& center : std::array{
```
mattkae: I tend to prefer unrolled `for` loop testing, the reason being that they are easier to follow. If the expectation on L144 fails, I will have to do a bunch of math to see what number caused the failure. If it is expanded, they I can clearly identify the case that caused it.

A looping test is rarely a good one in my opinion. It either needs to be parameterized or stepped through.

tests/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,420 @@
...
+            EXPECT_GE(-placement.capture_area.top().as_value() * 2.0, controls::handle_diameter);
+        }
+        if (bounds.bottom().as_value() == 600.0)
+        {
+            EXPECT_GE((placement.capture_area.bottom().as_value() - 600) * 2.0, controls::handle_diameter);
+        }
+    }
+}
+
+TEST(MagnifierLayout, source_clearance_is_not_capped_to_half_the_capture)
```
mattkae:
```suggestion
TEST(MagnifierLayout, source_clearance_is_capped_to_less_than_half_the_capture)
```

tests/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+TEST(MagnifierLayout, capture_dimensions_are_not_limited_to_one_thousand_pixels)
+{
+    auto const outputs = single_output({{0, 0}, {5000, 5000}});
+
+    auto const placement =
+        layout::place_following_cursor({2500.0, 2500.0}, {3500.0, 3500.0}, outputs, 1.0);
+
+    EXPECT_THAT(placement.capture_area.size, Eq(geom::Size{3500, 3500}));
+}
```
mattkae: That's a very random test. Why?

tests/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,420 @@
...
+
+TEST(MagnifierLayout, integral_translation_has_bounded_quantization_error)
+{
+    for (auto const magnification : {1.25, 1.5, 2.0, 3.5, 8.0})
```
mattkae: Parameterized?

tests/miral/magnifier_layout.cpp
```diff
@@ -0,0 +1,420 @@
...
+        EXPECT_THAT(translated.capture_area.size, Eq(original.capture_area.size));
+    }
+}
+
+TEST(MagnifierLayout, pinned_resize_keeps_bottom_right_handle_position)
+{
+    auto const outputs = single_output();
+    auto const pinned = geom::PointD{700.0, 550.0};
+
+    for (auto const magnification : {1.25, 1.5, 2.0, 3.5, 8.0})
```
mattkae: Parameterized?

mattkae (changes requested):
Mostly testing feedback, we should parameterize way more!

src/miral/magnifier_geometry.h
tarek-y-ismail: It feels wrong to move it to `magnifier_controls.h` to be honest.

tests/miral/magnifier_layout.cpp
tarek-y-ismail: Nope, your intuition is correct :) Been a while since I've done parameterized tests. Thanks for the reminder.

tests/miral/magnifier_layout.cpp
tarek-y-ismail: One of my earlier constraints was that the capture area would be capped to one thousand pixels in width and/or height, but that's gone now, and so should the test be.

src/miral/magnifier_geometry.h
mattkae: I disagree because it's only associated with the particularly functionality of that file. It can be namespace however we like in that file, but it definitely does not need to be its own file simply because it differs slightly from the controls.

Non blocking though 👍

mattkae (approved):
Alright, I took some time and this looks good to me! If you want to wait for Chris, go ahead, but this is good by me 👍

src/miral/magnifier_geometry.h
tarek-y-ismail: We can always move it after the fact.

PR #5135, opened 2026-08-12
miral: add freely-positioned magnifier mode with on-screen controls

tarek-y-ismail:
Closes #3954

Related: #5071

Stacked on #5206. Split from the original large change so freely positioned behavior can be reviewed independently from cursor-following layout updates and live configuration.

## What's new?

- Add `Magnifier::Behavior` and `Magnifier::set_behavior()` with a `freely_positioned` mode for environments such as touch devices where there is no cursor to follow.
- Add four on-screen controls for dragging, resizing, zooming in, and zooming out. The handle surfaces are excluded from the magnifier capture compositor.
- Preserve and constrain freely positioned geometry as outputs and magnification change.
- Add `Ctrl+Alt+M` in `miral-shell` to toggle behavior and provide a direct manual test path.

Input landing on the magnifier body still falls through to windows underneath; that is handled by a later PR. Live-config selection is split into the next PR in this stack.

## How to test

- Enable the magnifier with `Ctrl+=`.
- Press `Ctrl+Alt+M` to switch to freely positioned mode.
- Drag and resize it with its handles and use the zoom handles.

```sh
cmake --build build --target miral-test
LD_LIBRARY_PATH=build/lib ./build/bin/miral-test --gtest_filter='Magnifier*'
```

## Checklist

- [x] Tests added and pass
- [x] Documentation updated
- [ ] (optional) Added screenshots or videos


src/miral/magnifier.cpp
```diff
@@ -14,169 +14,780 @@
...
+        auto s = state.lock();
+        if (!s->follow_cursor)
+            return;
+
+        s->follow_cursor = false;
+        attach_observers(*s);
+
+        if (auto const surf = s->surface.lock())
+        {
+            if (s->default_enabled)
```
tarek-y-ismail: Usage makes it read more like `enabled` than `default_enabled`

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3766401138_

src/miral/magnifier.cpp
```diff
@@ -14,169 +14,780 @@
...
+    geom::Rectangles rects;
+    config->for_each_output(
+        [&rects](mg::DisplayConfigurationOutput const& output)
+        {
+            if (output.used && output.connected)
+                rects.add(output.extents());
+        });
+
+    auto s = self.state.lock();
+    s->screen_bounds = rects;
```
tarek-y-ismail, line 881: How do the screen bounds affect the code after?

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3766878670_

src/miral/magnifier.cpp
```diff
@@ -14,169 +14,780 @@
...
+    if (auto const surf = s->surface.lock())
+    {
+        auto const top_left = surf->top_left();
+        auto const pre_clamp_visual_size = s->visual_size;
+        auto const layout = s->layout();
+        auto const placement =
+            s->follow_cursor ?
+                layout.centered_on(s->cursor_pos) :
+                layout.freely_positioned(top_left);
```
tarek-y-ismail: This seems to be a recurring pattern.

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3766872941_

tests/miral/magnifier.cpp
```diff
@@ -78,26 +130,454 @@ TEST_F(MagnifierTest, magnification_results_in_scaled_transform)
...
+{
+    magnifier.enable(true).set_behavior(Magnifier::Behavior::freely_positioned);
+    start_server();
+    wait_for_magnifier_initialization();
+
+    auto const mux = server().the_cursor_observer_multiplexer();
+    auto sentinel = std::make_shared<SentinelCursorObserver>();
+    mux->register_interest(sentinel);
+    ASSERT_TRUE(sentinel->signal.wait_for(2s)) << "timed out waiting for initial cursor state";
+    sentinel->signal.reset();
```
tarek-y-ismail: This is a repeating pattern.

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3768556518_

tests/miral/magnifier.cpp
```diff
@@ -78,26 +130,454 @@ TEST_F(MagnifierTest, magnification_results_in_scaled_transform)
...
+TEST_F(MagnifierHandleTest, double_clicking_drag_handle_does_not_move_magnifier)
+{
+    auto const before = magnifier_top_left();
+    auto const handle_center = element_center(drag_handle_index);
+
+    click(handle_center);
+    click(handle_center);
+
+    EXPECT_THAT(magnifier_top_left(), Eq(before));
+}
```
tarek-y-ismail: That feature was removed. This can be discarded.

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3768709920_

tarek-y-ismail:
Look like I messed up the history. Will fix soon :tm:

src/miral/magnifier.cpp
```diff
@@ -70,38 +100,55 @@ class miral::Magnifier::Self
...
-                surf->set_depth_layer(mir_depth_layer_always_on_top);
-                surf->set_focus_mode(mir_focus_mode_disabled);
 
                 if (s->default_enabled)
                     surf->show();
                 else
                     surf->hide();
-
-                s->surface = surf;
             });
```
tarek-y-ismail: What use is this?

src/miral/magnifier.cpp
```diff
@@ -70,38 +100,55 @@ class miral::Magnifier::Self
...
             {
-                if (auto const locked = cursor_multiplexer.lock())
-                    locked->unregister_interest(*cursor_observer);
+                auto const cursor_mux = cursor_multiplexer.lock();
+                if (cursor_mux && cursor_observer)
+                    cursor_mux->unregister_interest(*cursor_observer);
+
+                if (display_config_observer)
+                    server.the_display_configuration_observer_registrar()->unregister_interest(
+                        *display_config_observer);
```
tarek-y-ismail: Do we actually need to hold onto the cursor multiplexer?

src/miral/magnifier.cpp
```diff
@@ -117,22 +164,81 @@ class miral::Magnifier::Self
...
-        s->magnification = new_magnification;
-        if (auto const surf = s->surface.lock())
-            surf->set_transformation(glm::scale(glm::mat4(1.0), glm::vec3(s->magnification, s->magnification, 1)));
+        set_magnification(*state.lock(), new_magnification);
+    }
+
+    void set_magnification(State& s, float new_magnification)
+    {
+        s.magnification = new_magnification;
+        if (!s.surface.lock() || !s.has_outputs())
```
tarek-y-ismail: We already check `has_outputs` in `place_at_cursor`

src/miral/magnifier.cpp
```diff
@@ -170,25 +203,63 @@ class miral::Magnifier::Self
...
+            return;
+
+        s->follow_cursor = false;
+
+        if (auto const surf = s->surface.lock(); surf && s->enabled)
+        {
+            s->freely_positioned_center = geom::PointD{
+                surf->top_left() +
+                geom::generic::as_displacement(s->render_scene_into_surface.capture_area().size / 2.0)};
+            place_freely(*s);
```
tarek-y-ismail: This is the third time this block repeats.

src/miral/magnifier.cpp
```diff
@@ -170,25 +203,63 @@ class miral::Magnifier::Self
...
+    {
+        auto s = state.lock();
+        if (s->follow_cursor)
+            return;
+
+        s->follow_cursor = true;
+        place_at_cursor(*s);
+    }
+
+    void stop_following_cursor()
```
tarek-y-ismail, line 452: Maybe should be names `place_freely` to match other parts of the code.

src/miral/magnifier.cpp
```diff
@@ -69,16 +71,28 @@ struct State
...
     std::weak_ptr<ms::Surface> surface;
     geom::Point cursor_pos;
+    geom::PointD freely_positioned_center;
     geom::Rectangles screen_bounds;
     float magnification{default_magnification};
     geom::SizeD requested_visual_size{
         default_capture_width * static_cast<double>(default_magnification),
         default_capture_height * static_cast<double>(default_magnification)};
-    bool default_enabled{false};
+    std::optional<mml::FreePlacement> applied_placement;
```
tarek-y-ismail, line 247: What does this do? It's only set and reset in this commit. Never checked,

tests/miral/magnifier.cpp
```diff
@@ -169,7 +189,7 @@ TEST_F(MagnifierTest, capture_size_is_limited_to_80_percent_of_the_output)
 
 TEST_F(MagnifierTest, reapplies_layout_after_outputs_are_restored)
 {
-    magnifier.capture_size(Size(1000, 1000)).enable(true);
+    magnifier.capture_size(Size(1000, 1000)).enable(true).set_behavior(Magnifier::Behavior::freely_positioned);
```
tarek-y-ismail, line 246: Why the change to freely positioned? Does it matter?

tests/miral/CMakeLists.txt
```diff
@@ -70,6 +71,7 @@ target_link_libraries(miral-test-internal
     mir-test-doubles-static
     miral-internal
     miral-external
+    mirnullreport
```
tarek-y-ismail, line 74: What's this?

src/miral/magnifier.cpp
```diff
@@ -52,6 +60,78 @@ auto const min_magnification = 1.25f;
...
+    void init(mir::Server& server, mmc::HandleKind kind, mc::CompositorID capture_compositor_id)
+    {
+        indicator = std::make_shared<miral::HandleIndicator>(
+            handle_rect,
+            kind,
+            capture_compositor_id,
+            server.the_buffer_allocator(),
+            server.the_scene_report(),
+            server.the_display_configuration_observer_registrar());
+        handle_surface_stack = server.the_surface_stack();
```
tarek-y-ismail, line 100: Can be moved to the member initializer list

src/miral/magnifier.cpp
```diff
@@ -82,11 +100,32 @@ class Handle
...
+
+    void detach_observer()
+    {
+        release_observer().unregister();
+    }
+
+    auto release_observer() -> ObserverRegistration
+    {
+        return {indicator, std::exchange(observer, {})};
+    }
```
tarek-y-ismail, line 132: What's the difference between those two?

src/miral/magnifier.cpp
```diff
@@ -402,21 +407,38 @@ class miral::Magnifier::Self
...
+                s->handles.zoom_out.release_observer()};
+
+            s->hide_all_handles();
+            place_at_cursor(*s);
+
+            return ret;
+        }();
+
+        for (auto obs : observers)
+            obs.unregister();
```
tarek-y-ismail, line 449: If I recall correctly, this contraption was made because un-registering observers with the state lock held would deadlock. Is this the case? Is there a cleaner way?

src/miral/magnifier.cpp
```diff
@@ -638,6 +640,80 @@ class miral::Magnifier::Self
...
+                mml::resize_freely(
+                    dragged_visual_top_left,
+                    pinned_visual_corner,
+                    s.screen_bounds,
+                    s.magnification));
+        }
+
+        geom::PointD pinned_visual_corner{};
+        geom::PointD resize_start_visual_top_left{};
+        bool has_pinned_visual_corner{false};
```
tarek-y-ismail, line 727: Can use `std::optional` instead.

src/miral/magnifier.cpp
```diff
@@ -638,6 +640,80 @@ class miral::Magnifier::Self
...
+        }
+
+        void on_drag_move(State& s, geom::Point point) override
+        {
+            auto const surf = s.surface.lock();
+            if (!surf || !has_pinned_visual_corner)
+                return;
+
+            if (!s.has_outputs())
+                return;
```
tarek-y-ismail, line 709: `apply_geometry` already checks `has_outputs`.

src/miral/magnifier.cpp
```diff
@@ -215,6 +215,27 @@ struct State
...
+            screen_bounds,
+            [](geom::Rectangle const& output)
+            { return output.size.width > geom::Width{0} && output.size.height > geom::Height{0}; });
+
+        if (primary == std::ranges::end(screen_bounds))
+            return {};
+
+        return primary->top_left +
+               geom::Displacement{
+                   geom::as_delta(primary->size.width / 2), geom::as_delta(primary->size.height / 2)};
```
tarek-y-ismail, line 235: I believe `Rectangle::centre` can be used in such cases?

src/miral/magnifier.cpp
```diff
@@ -377,7 +397,11 @@ class miral::Magnifier::Self
     {
         s.magnification = new_magnification;
         s.applied_placement.reset();
-        if (!s.surface.lock() || !s.has_outputs())
+        auto const surf = s.surface.lock();
+        if (!surf)
+            return;
+
+        if (!s.has_outputs())
```
tarek-y-ismail: Why?

src/miral/magnifier.cpp
```diff
@@ -768,12 +789,22 @@ void miral::Magnifier::Self::DisplayConfigObserver::update_bounds(
...
+    if (auto const surf = s->surface.lock())
     {
         if (s->follow_cursor)
-            self.place_at_cursor(*s);
-        else
-            self.place_freely(*s);
+        {
+            s->apply_geometry(
+                mml::place_following_cursor(
+                    geom::PointD{s->cursor_pos}, s->requested_visual_size, s->screen_bounds, s->magnification));
```
tarek-y-ismail, line 848: Can just use `place_at_cursor`?

src/miral/magnifier.cpp
```diff
@@ -768,12 +789,22 @@ void miral::Magnifier::Self::DisplayConfigObserver::update_bounds(
...
+        }
+        else if (
+            auto const placement = mml::place_freely(
+                s->free_placement_center(), s->requested_visual_size, s->screen_bounds, s->magnification);
+            placement != s->applied_placement)
+        {
+            s->applied_placement = placement;
+            s->apply_geometry(placement);
```
tarek-y-ismail, line 856: Maybe `place_freely` can check `applied_placement` instead?

tests/miral/magnifier.cpp
```diff
@@ -488,6 +529,21 @@ TEST_F(MagnifierHandleTest, returns_to_last_free_position_after_behavior_toggle)
...
+    auto const drag_from = element_center(drag_handle_index);
+    drag(drag_from, geom::PointF{drag_from.x.as_value() + 20, drag_from.y.as_value()});
+
+    EXPECT_THAT(magnifier_top_left().x.as_int(), Eq(before.x.as_int() + 20));
+
+    auto const before_zoom = magnifier_renderable()->transformation();
+    click(element_center(zoom_in_handle_index));
+
+    EXPECT_THAT(magnifier_renderable()->transformation(), Ne(before_zoom));
+}
```
tarek-y-ismail, line 591: Can be chopped down to just the zoom index test. If handles didn't work, other tests would fail as well.

tests/miral/CMakeLists.txt
tarek-y-ismail: Needed for handle indicator tests.

src/miral/magnifier.cpp
tarek-y-ismail: Bad idea. Need server which is not available at construction time.

src/miral/magnifier.cpp
tarek-y-ismail: Can just inline `detach_observer` into the only usage.

src/miral/magnifier.cpp
tarek-y-ismail: At a glance, no. We need the lock to release the observers, and we do a couple of things while we have it. Then we return the data we need to unregister to be used outside the lock.

src/miral/magnifier.cpp
tarek-y-ismail: Not anymore.

examples/miral-shell/shell_main.cpp
```diff
@@ -208,7 +209,11 @@ int main(int argc, char const* argv[])
         .magnification(magnification)
         .capture_size(capture_size)
         .enable(false);
-    auto magnifier_filter = [magnifier=magnifier, &magnification, &capture_size](MirKeyboardEvent const* key_event) mutable {
+    auto magnifier_filter = [magnifier = magnifier,
+                             &magnification,
+                             &capture_size,
+                             follows_cursor=true](MirKeyboardEvent const* key_event) mutable
```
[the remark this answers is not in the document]
tarek-y-ismail: Opinion: PR is big as it is, should be a followup. Nonetheless, we should agree on how to fix this (probably getters)

src/miral/magnifier_handle_indicator.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace mir
+{
+namespace graphics { class GraphicBufferAllocator; }
+namespace scene { class SceneReport; }
+}
+
+namespace miral
+{
+class HandleIndicator : public mir::scene::BasicSurface
```
mattkae:
```suggestion
class MagnifierHandleIndicator : public mir::scene::BasicSurface
```

src/miral/magnifier_handle_indicator.cpp
```diff
@@ -0,0 +1,346 @@
...
+        geom::Height thickness;
+    };
+
+    struct VerticalLine
+    {
+        geom::X x;
+        geom::Y start;
+        geom::Y end;
+        geom::Width thickness;
+    };
```
mattkae, line 51: Nit: I feel like a singular "Line" would be just as simple to write.

As a side note, this code makes me sad. I wish we could just bring in some simple toolkit to draw circles and lines. Ugh 😭 Maybe for a future date!

src/miral/magnifier_handle_indicator.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+    auto buffer_size() const -> geom::Size { return size; }
+
+private:
+    void set_pixel(geom::Point const& point, uint8_t grey)
+    {
+        auto const x = point.x.as_value();
+        auto const y = point.y.as_value();
+        if (x < 0 || x >= size.width.as_value() || y < 0 || y >= size.height.as_value())
+            return;
```
mattkae, line 112: Should we `warn` in this event or something? I feel like if you're doing this, you're mistaken, right?

src/miral/magnifier_handle_indicator.cpp
```diff
@@ -0,0 +1,346 @@
...
+        auto const true_end = std::max<geom::Y>(line.start, line.end);
+        for (int t = 0; t < line.thickness.as_value(); ++t)
+            for (auto y = true_start; y <= true_end; y = y + geom::DeltaY{1})
+                set_pixel(geom::Point{line.x + geom::DeltaX{t}, y}, icon_grey);
+    }
+
+    auto buffer_size() const -> geom::Size { return size; }
+
+private:
+    void set_pixel(geom::Point const& point, uint8_t grey)
```
mattkae, line 107:
```suggestion
    void set_pixel(geom::Point const& point, uint8_t color)
```

src/miral/magnifier_handle_indicator.h
```diff
@@ -0,0 +1,55 @@
...
+{
+public:
+    HandleIndicator(
```
mattkae, line 38: I would personally rather get rid of `HandleKind` in favor of just having a single static method that returns everything, e.g.:

```c++
static create_handles(...) -> Handles;
```

where `Handles` is your `struct` above.

This is simpler for the caller, and we get everything immediately (instead of a bunch of separate calls).

src/miral/magnifier.cpp
```diff
@@ -51,16 +69,185 @@ auto const min_magnification = 1.25f;
...
+
+    void show_all_handles() { handles.for_each([](Handle& handle, auto) { handle.show(); }); }
+
+    void hide_all_handles() { handles.for_each([](Handle& handle, auto) { handle.hide(); }); }
+
+    auto has_outputs() const -> bool { return screen_bounds.size() != 0; }
+
+    auto free_placement_center() const -> geom::PointD
+    {
+        return user_positioned ? freely_positioned_center : geom::PointD{primary_output_center()};
```
mattkae, line 219: `user_positioned` is only ever set to true once a move happens, right? If that's the case, can't we just get rid of it and return whatever `geom::PointD` we currently have as our center?

src/miral/magnifier.cpp
```diff
@@ -51,16 +69,185 @@ auto const min_magnification = 1.25f;
...
+        {
+            geom::Width{mmc::handle_diameter},
+            geom::Height{mmc::handle_diameter}}};
+
+    std::shared_ptr<miral::HandleIndicator> indicator;
+    std::weak_ptr<msh::SurfaceStack> handle_surface_stack;
+    std::shared_ptr<ms::NullSurfaceObserver> observer;
+};
+
+struct Handles
```
mattkae, line 160: In my opinion, you'd be better off with:

```c++
typedef std::array<Handle, 4> Handles;
```

Then you can just iterate them, and maybe have some helpers to get the specifics (e.g. `drag`) if you need it, although it doesn't really seem like you do

src/miral/magnifier.cpp
```diff
@@ -51,16 +69,185 @@ auto const min_magnification = 1.25f;
...
+    {
+        release_observer().unregister();
+
+        if (auto const surface_stack = handle_surface_stack.lock())
+            surface_stack->remove_surface(indicator);
+        indicator.reset();
+    }
+
+    template<typename ObserverType, typename... Args>
+    void attach_observer(Args&&... args)
```
mattkae: I am confused by this 😄 Why is this a template, wouldn't it always be a `SurfaceObserver`? Why does it not need to know the type? Can't we just pass a shared_ptr that derives from the base class?

src/miral/magnifier.cpp
```diff
@@ -106,61 +293,165 @@ class miral::Magnifier::Self
...
+        {
+            auto s = state.lock();
+            s->follow_cursor = true;
+
+            std::array<Handle::ObserverRegistration, 4> const ret{
+                s->handles.drag.release_observer(),
+                s->handles.resize.release_observer(),
+                s->handles.zoom_in.release_observer(),
+                s->handles.zoom_out.release_observer()};
+
```
mattkae: Why is this implemented this way? A comment would be useful. do we need this complexity?

tests/miral/magnifier.cpp
```diff
@@ -95,13 +104,36 @@ class MagnifierTest : public TestServer
             mir::test::Signal signal;
         };
 
+    /// Registering interest replays the current cursor state, so waiting for
+    /// the sentinel's first signal is how a test knows the initial placement
+    /// has settled.
+    void wait_for_initial_cursor_state()
+    {
+
```
mattkae:
```suggestion
```

tests/miral/magnifier.cpp
```diff
@@ -239,3 +303,669 @@ TEST_F(MagnifierTest, reapplies_layout_after_outputs_are_restored)
...
+    start_server();
+    wait_for_magnifier_initialization();
+    wait_for_initial_cursor_state();
+
+    magnifier.set_behavior(Magnifier::Behavior::freely_positioned);
+
+    EXPECT_THAT(magnifier_renderable()->screen_position().centre(), Eq(geom::Point{400, 300}));
+}
+
+TEST_F(MagnifierTest, cursor_not_tracked_in_decoupled_mode)
```
mattkae, line 385: Why do you keep calling this `decoupled_mode` rather than `freely_positioned_mode`?

tests/miral/magnifier.cpp
```diff
@@ -239,3 +303,669 @@ TEST_F(MagnifierTest, reapplies_layout_after_outputs_are_restored)
...
+    std::shared_ptr<mi::VirtualInputDevice> pointer_device;
+    std::shared_ptr<mi::VirtualInputDevice> touch_device;
+    std::shared_ptr<mi::VirtualInputDevice> second_touch_device;
+
+private:
+    /// Surface input observers use BasicSurface::Multiplexer which dispatches
+    /// via linearising_executor (a non-blocking deferred executor backed by a
+    /// thread pool). Spawn a sentinel task AFTER the input events so that when
+    /// it fires, all earlier callbacks — including the handle observer state
+    /// updates (move_to, resize, set_transformation) — have already run.
```
mattkae:
```suggestion
    /// Surface input observers use BasicSurface::Multiplexer which dispatches
    /// via linearising_executor. We queue an action on that same dispatcher
    /// and wait for it to complete before proceeding.
```

mattkae (changes requested):
I'll do another pass at the tests later, but this is my initial feedback about the main implementation :)

src/miral/magnifier.cpp
tarek-y-ismail: Just the first solution that came to mind when I was refactoring some code. But your solution is valid as well. Thank you.

src/miral/magnifier.cpp
tarek-y-ismail: Tried it, unfortunately it's not much better than what we already have. I

src/miral/magnifier.cpp
tarek-y-ismail: Good find. Took the chance to move `freely_positioned_center` updates to `update_bounds` so now every display configuration would recenter the magnifier. This should cover the case where a new "primary output" (i.e. biggest area) is plugged in at runtime.

src/miral/magnifier_handle_indicator.cpp
tarek-y-ismail: This contortion is mostly to get around TICS's "no two arguments of the same type" error. 

Other solutions include:
- `draw_horizontal_line(x, y, x, ...)` 
- Implementing a generic line drawing algorithm (see https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm)
- Using `NOLINTNEXTLINE`

But yes, we should probably refactor all simple shape and text drawing into an internal library. That much is true.

src/miral/magnifier_handle_indicator.cpp
tarek-y-ismail: It's really just how much gray the pixel is. Check a few lines after, you'll see `pixel[0] = pixel[1] = pixel[2] = grey`

src/miral/magnifier_handle_indicator.cpp
tarek-y-ismail: Yup, should throw a fatal error 👍

src/miral/magnifier_handle_indicator.h
tarek-y-ismail: I don't understand what this is getting at? `HandleIndicator` owns the drawing and buffer management, `Handle` marries that with an observer to respond to input events, and `Handles` groups the four into one.

tests/miral/magnifier.cpp
tarek-y-ismail: Old naming. Let me fix all references in one go.

src/miral/magnifier.cpp
```diff
@@ -51,16 +69,162 @@ auto const min_magnification = 1.25f;
...
+        if (auto const surface_stack = handle_surface_stack.lock())
+            surface_stack->remove_surface(indicator);
+        indicator.reset();
+    }
+
+    void attach_observer(std::shared_ptr<ms::SurfaceObserver> const& observer)
+    {
+        if (!indicator || this->observer)
+            return;
```
mattkae: I would at least `warn`. This seems like a programming error that the author should consider.

```suggestion
        if (!indicator)
        {
            mir::log_warning("The indicator is currently null on this handle, not adding the observer");
            return;
        }
            
        if (this->observer)
        {
            mir::log_warning("Observer already attached to handle, not adding another");
            return;
        }
```

src/miral/magnifier.cpp
mattkae: \`freely_positioned` would work in my opinion.

src/miral/magnifier.cpp
```diff
@@ -178,37 +439,263 @@ class miral::Magnifier::Self
...
+        state.handles.zoom_in.attach_observer(std::make_shared<ZoomButtonObserver>(this, zoom_step));
+        state.handles.zoom_out.attach_observer(std::make_shared<ZoomButtonObserver>(this, -zoom_step));
+    }
 
-        if (auto const surf = state.surface.lock())
-        {
-            surf->move_to(new_placement.capture_area.top_left);
-            surf->set_transformation(glm::scale(glm::mat4(1.0), glm::vec3(state.magnification, state.magnification, 1)));
-        }
+    void place(State& s)
```
mattkae: I would rename this to `display_bounds_changed`, since it is only gets called in that instance.

src/miral/magnifier.cpp
mattkae: Honestly, I would just entirely delegate the body of this funciton up to `Self` instead of having the logic here. It feels weird having this class dig into the internals of `Self` like that

src/miral/magnifier.cpp
```diff
@@ -178,37 +439,263 @@ class miral::Magnifier::Self
         Self& self;
     };
 
-    void apply_geometry(
-        mml::Placement const& new_placement,
-        miral::RenderSceneIntoSurface& render_scene_into_surface,
-        State& state)
+    /// Creates and registers concrete observers on each handle. Guards against
+    /// missing indicators.
+    void attach_observers(State& state)
```
mattkae: Why do we repeatedly call this in `set_enable` and `stop_following_cursor` if they are already attached on init, and subsequent attaches seem to have no effect?

src/miral/magnifier.cpp
```diff
@@ -119,48 +300,128 @@ class miral::Magnifier::Self
...
+                s->handles.resize.release_observer(),
+                s->handles.zoom_in.release_observer(),
+                s->handles.zoom_out.release_observer()};
+
+            s->hide_all_handles();
+            place_at_cursor(*s);
+        }
+
+        for (auto& obs : observers)
+            obs.unregister();
```
mattkae: Should this move into a `detatch_observers` private function, to match the `attach_observers`? Should this also be called when we `disable`, since we seem to call it when we reenable? Why do we call it when we reenable if we do not detach the observers on `disable`?

tests/miral/magnifier.cpp
```diff
@@ -239,3 +302,700 @@ TEST_F(MagnifierTest, reapplies_layout_after_outputs_are_restored)
...
+    magnifier.enable(true).set_behavior(Magnifier::Behavior::freely_positioned);
+    add_start_callback([&]
+    {
+        magnifier.enable(false);
+        EXPECT_THAT(scene_element_count(), Eq(0));
+    });
+    start_server();
+}
+
+TEST_F(MagnifierTest, toggling_to_coupled_hides_handles)
```
mattkae:
```suggestion
TEST_F(MagnifierTest, toggling_to_follow_cursor_hides_handles)
```

tests/miral/magnifier.cpp
```diff
@@ -239,3 +302,700 @@ TEST_F(MagnifierTest, reapplies_layout_after_outputs_are_restored)
...
+    add_start_callback([&]
+    {
+        EXPECT_THAT(scene_element_count(), Eq(5));
+        magnifier.set_behavior(Magnifier::Behavior::follow_cursor);
+        EXPECT_THAT(scene_element_count(), Eq(1));
+    });
+    start_server();
+}
+
+TEST_F(MagnifierTest, decoupling_after_start_shows_handles)
```
mattkae:
```suggestion
TEST_F(MagnifierTest, freely_positioning_after_start_shows_handles)
```

tests/miral/magnifier.cpp
```diff
@@ -239,3 +302,700 @@ TEST_F(MagnifierTest, reapplies_layout_after_outputs_are_restored)
...
+    magnifier.capture_size(Size(200, 200));
+    // Claim the stale frame so the next scene read receives the resized one.
+    magnifier_renderable()->buffer();
+    auto const after = magnifier_renderable()->screen_position().centre();
+
+    EXPECT_THAT(after, Eq(before));
+    EXPECT_THAT(after, Ne(geom::Point{700, 500}));
+}
+
+TEST_F(MagnifierTest, cursor_tracked_in_coupled_mode)
```
mattkae:
```suggestion
TEST_F(MagnifierTest, cursor_tracked_in_follow_cursor_mode)
```

mattkae (commented):
Some minor food for thought, but I think I am largely happy!

PR #5136, opened 2026-08-12
miral: consume input over the freely-positioned magnifier

tarek-y-ismail:
Related: #5071

## What's new?

- Pointer and touch events landing on the freely-positioned magnifier fell through to whatever happens to be underneath it, so interacting with the magnifier also poked at the windows it covers.
- Add an event filter that swallows pointer and touch events over the magnifier's content while it is freely positioned and enabled.
- Gestures are tracked from press to release, so a drag that starts on the magnifier stays consumed even after it leaves the surface, and a gesture that starts elsewhere is never stolen mid-way.
- Events over the control handles are deliberately left alone so the handles still receive them.

## How to test

- Enable the magnifier (`Ctrl + =`) and decouple it (`Ctrl + Alt + M`).
- Click and drag on the magnified content over a window: the window underneath no longer reacts.
- The handles still work, and events outside the magnifier are unaffected.

```sh
cmake --build build --target miral-test
LD_LIBRARY_PATH=build/lib ./build/bin/miral-test --gtest_filter='Magnifier*'
```

23 tests, all passing, including 4 new `consumes_*` cases.

## Checklist

- [x] Tests added and pass


src/miral/magnifier.cpp
```diff
@@ -749,11 +783,89 @@ class miral::Magnifier::Self
...
+                .value_or(false);
+    }
+
+    if (pointer_gesture_consumed)
+    {
+        auto const consumed = *pointer_gesture_consumed;
+        if (action == mir_pointer_action_button_up && event.buttons() == 0)
+            pointer_gesture_consumed.reset();
+        return consumed;
+    }
```
tarek-y-ismail, line 956: Why do we need all of this?

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3766474501_

tests/miral/magnifier.cpp
```diff
@@ -531,6 +536,112 @@ TEST_F(MagnifierHandleTest, double_clicking_drag_handle_does_not_move_magnifier)
...
+            1.0f,
+            1.0f,
+            1.0f,
+            0.0f}});
+
+    EXPECT_FALSE(composite_event_filter.lock()->handle(*pointer_event));
+    EXPECT_FALSE(composite_event_filter.lock()->handle(*touch_event));
+}
+
+TEST_F(MagnifierHandleTest, does_not_consume_events_when_magnifier_is_inactive)
```
tarek-y-ismail: misleading name

_Carried over from https://github.com/canonical/mir/pull/5071#discussion_r3768717304_

PR #5142, opened 2026-08-13
Fix libmircore3.symbols

AlanGriffiths:

PR #5149, opened 2026-08-14
Fix-the-build

AlanGriffiths:
Alternative to #5142


AlanGriffiths:
OK, that guess was wrong

Issue #4981, opened 2026-06-01
update_package_abis.sh does not update mir.spec


AlanGriffiths:
The utility `tools/update_package_abis.sh` has not been updated since we added Fedora packaging. In particular, it does not check/update the `rpm/mir.spec`

PR #366, opened 2018-05-10
test-wlcs-fix-22

AlanGriffiths:
Testing wlcs fix across distros

PR #5152, opened 2026-08-15
tools: sync rpm/mir.spec sover macros in update_package_abis.sh

xhon-pelushi:
## What this does

`tools/update_package_abis.sh` only ever wrote `debian/control` and `debian/*.install`. Since Fedora/RPM packaging (`rpm/mir.spec`) was added, the script was never extended to keep its `%global <lib>_sover N` macros in sync with the same CMake `*_ABI` variables, so `rpm/mir.spec` can silently drift after an ABI bump.

This adds RPM spec support alongside the existing debian handling:
- Maps each `*_ABI` variable to its corresponding `rpm/mir.spec` sover macro.
- Updates those macros in the default/update code path.
- Reports mismatches under `--check` (deduped, since several packages share one ABI variable).
- Corrects `.github/instructions/abi-symbols.instructions.md`, which explicitly (and now incorrectly) said the script "does not modify RPM packaging."

## Testing

- `./tools/update_package_abis.sh --check --verbose` on a clean tree: exit 0, all 9 ABI vars detected, no mismatches.
- Manually corrupted two sover macros in `rpm/mir.spec`, re-ran `--check`: correctly reported both mismatches (deduped where multiple packages share one ABI variable).
- Ran update mode: both macros restored; diff against the original `rpm/mir.spec` is identical, confirming a clean round-trip.
- `git status` confirms only `tools/update_package_abis.sh` and the one doc file are touched — `rpm/mir.spec`, `debian/control`, `debian/*.install` unchanged in the final diff.

## Related

#4981. A previous attempt (#5092) was self-closed by its author with no maintainer feedback.

### DIFF
--- a/tools/update_package_abis.sh
+++ b/tools/update_package_abis.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-# Script to check and update debian packaging to match current ABIs
+# Script to check and update debian and RPM packaging to match current ABIs
 
 set -e
 
@@ -37,12 +37,28 @@ package_abi_var()
     echo "${1##*:}"
 }
 
+# Maps each ABI variable to its corresponding soversion macro in rpm/mir.spec.
+rpm_sover_macro()
+{
+    case "$1" in
+        MIROIL_ABI) echo "miroil_sover" ;;
+        MIRAL_ABI) echo "miral_sover" ;;
+        MIRCORE_ABI) echo "mircore_sover" ;;
+        MIRCOMMON_ABI) echo "mircommon_sover" ;;
+        MIRPLATFORM_ABI) echo "mirplatform_sover" ;;
+        MIRSERVER_ABI) echo "mirserver_sover" ;;
+        MIR_SERVER_GRAPHICS_PLATFORM_ABI) echo "mirplatformgraphics_sover" ;;
+        MIR_SERVER_INPUT_PLATFORM_ABI) echo "mirplatforminput_sover" ;;
+        MIRWAYLAND_ABI) echo "mirwayland_sover" ;;
+    esac
+}
+
 print_help_and_exit()
 {
     local prog=$(basename $0)
 
     echo "Usage: $prog [OPTION]..."
-    echo "Update debian packaging to match current ABIs"
+    echo "Update debian and RPM packaging to match current ABIs"
     echo ""
     echo "      --no-vcs   do not register file changes (e.g., moves) with the detected VCS"
     echo "      --check    check for ABI mismatches without updating debian packaging"
@@ -157,6 +173,26 @@ update_install_files()
     done
 }
 
+update_spec_file()
+{
+    if [ ! -f rpm/mir.spec ];
+    then
+        return
+    fi
+
+    for p in $packages;
+    do
+        local abi_var=$(package_abi_var $p)
+        local sover_macro=$(rpm_sover_macro $abi_var)
+        if [ -z "$sover_macro" ];
+        then
+            continue
+        fi
+        local abi=$(eval "echo \$${abi_var}")
+        sed -i "s/^%global ${sover_macro} [[:digit:]]\+/%global ${sover_macro} ${abi}/" rpm/mir.spec
+    done
+}
+
 report_abi_mismatch()
 {
     log "ABI mismatch: $1"
@@ -198,6 +234,36 @@ check_install_files()
             then
                 report_abi_mismatch "$current_file contains $suffix, but $pkg ABI is $abi"
             fi
+        fi
+    done
+}
+
+check_spec_file()
+{
+    if [ ! -f rpm/mir.spec ];
+    then
+        return
+    fi
+
+    local seen_macros=""
+    for p in $packages;
+    do
+        local abi_var=$(package_abi_var $p)
+        local sover_macro=$(rpm_sover_macro $abi_var)
+        if [ -z "$sover_macro" ];
+        then
+            continue
+        fi
+        case " $seen_macros " in
+            *" $sover_macro "*) continue ;;
+        esac
+        seen_macros="$seen_macros $sover_macro"
+
+        local abi=$(eval "echo \$${abi_var}")
+        local current=$(grep -o "^%global ${sover_macro} [[:digit:]]\+" rpm/mir.spec | grep -o '[[:digit:]]\+$')
+        if [ -n "$current" ] && [ "$current" != "$abi" ];
+        then
+            report_abi_mismatch "rpm/mir.spec defines ${sover_macro} as ${current}, but ${abi_var} is ${abi}"
         fi
     done
 }
@@ -250,6 +316,7 @@ then
     check_for_unknown_packages
     check_control_file
     check_install_files
+    check_spec_file
     if [ "$has_check_error" = "yes" ];
     then
         log "Consider running tools/update_package_abis.sh to fix the mismatches."
@@ -259,4 +326,5 @@ else
     check_for_unknown_packages
     update_control_file
     update_install_files
+    update_spec_file
 fi
--- a/tools/update_package_abis.sh
+++ b/tools/update_package_abis.sh
@@ -177,15 +177,22 @@ update_spec_file()
 {
     if [ ! -f rpm/mir.spec ];
     then
+        echo "No rpm/mir.spec in this source tree: skipping RPM packaging" >&2
         return
     fi
 
+    local unmapped_vars=""
     for p in $packages;
     do
         local abi_var=$(package_abi_var $p)
         local sover_macro=$(rpm_sover_macro $abi_var)
         if [ -z "$sover_macro" ];
         then
+            case " $unmapped_vars " in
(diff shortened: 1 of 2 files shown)


tools/update_package_abis.sh
```diff
@@ -157,6 +173,26 @@ update_install_files()
     done
 }
 
+update_spec_file()
+{
+    if [ ! -f rpm/mir.spec ];
+    then
+        return
+    fi
```
RAOF, line 182: I don't *think* we need this; the script has verified that we're running in the root (or, at least, somewhere with a `debian/` directory).

If you'd like to keep it, then I think it should log (with an unconditional `echo ... >&2` to send it to stderr)  on error.

tools/update_package_abis.sh
```diff
@@ -157,6 +173,26 @@ update_install_files()
...
+
+    for p in $packages;
+    do
+        local abi_var=$(package_abi_var $p)
+        local sover_macro=$(rpm_sover_macro $abi_var)
+        if [ -z "$sover_macro" ];
+        then
+            continue
+        fi
```
RAOF, line 197: I think this, too, should at least log on error.

RAOF (changes requested):
Thanks for this! It looks sensible; I'd just like the error checking to be loud when an error is detected.

AlanGriffiths:
> Thanks for this! It looks sensible; I'd just like the error checking to be loud when an error is detected.

Me too

RAOF:
@xhon-pelushi Ping! Do you want to respond to the review comments? Otherwise, I'm happy to fix them up myself and then merge this.

If you don't respond in a week, I'll make the changes and merge this; I'd like to get this work in :grin:

tools/update_package_abis.sh
xhon-pelushi: I checked this against the branches the script actually ships on, and I think the guard is load-bearing: the top-of-script check proves `debian/` exists, but `debian/` existing doesn't imply `rpm/` existing. `tools/update_package_abis.sh` is on every release branch; `rpm/mir.spec` only appears from `release/2.23` onwards:

```
release/1.8    debian/=YES script=YES rpm/mir.spec=no
...
release/2.22   debian/=YES script=YES rpm/mir.spec=no
release/2.23   debian/=YES script=YES rpm/mir.spec=YES
release/2.25   debian/=YES script=YES rpm/mir.spec=YES
release/2.26   debian/=YES script=YES rpm/mir.spec=YES
release/2.28   debian/=YES script=YES rpm/mir.spec=YES
release/2.29   debian/=YES script=YES rpm/mir.spec=YES
```

So on `release/1.8` through `release/2.22` the root check passes and `rpm/mir.spec` is still absent.

I removed both guards locally and ran the script in a tree with `debian/` but no `rpm/` to see what the failure mode actually is (`/bin/sh` is dash here):

- `--check` **exits 0** — reports success — after printing nine `grep: rpm/mir.spec: No such file or directory` lines to stderr. `local current=$(grep ...)` doesn't trip `set -e` in dash, so the missing file never reaches the exit status. That's the one I'd most want to avoid: a check that passes without having checked anything.
- Update mode fails with `sed: can't read rpm/mir.spec: No such file or directory` and exit 2 — but only *after* `update_control_file` and `update_install_files` have already run. With `MIRAL_ABI` bumped 8 -> 9 and `rpm/` absent, the guard-less run left:

  ```
   M debian/control
   D debian/libmiral8.install
  ?? debian/libmiral9.install
  ```

  i.e. `debian/` rewritten, spec not written, command reports failure.

So I'd rather keep it — but you're right that it shouldn't be silent. Changed to your suggested form, in both `update_spec_file()` and `check_spec_file()`:

```diff
     if [ ! -f rpm/mir.spec ];
     then
+        echo "No rpm/mir.spec in this source tree: skipping RPM packaging" >&2
         return
     fi
```

Verified: with `rpm/` absent, both `--check` and update mode now print that one line to stderr and exit 0, and `debian/` is still updated correctly.

tools/update_package_abis.sh
xhon-pelushi: Agreed, and this one is more than cosmetic — it's the same drift the PR exists to prevent. If a package with a new `*_ABI` variable is added to `$packages` and `rpm_sover_macro()` isn't extended, that spec macro silently stops being synced, in both `--check` and update mode. `debian/control` already gets a loud version of this via `report_unknown_package()`, so I matched that style:

```diff
+report_unmapped_sover()
+{
+    echo "rpm/mir.spec has no known soversion macro for $1: rpm_sover_macro() in this script needs to be updated" >&2
+}
```

...and reported once per ABI variable rather than once per package, so an unmapped `MIR_SERVER_GRAPHICS_PLATFORM_ABI` (shared by six entries in `$packages`) doesn't print six identical lines:

```diff
         if [ -z "$sover_macro" ];
         then
+            case " $unmapped_vars " in
+                *" $abi_var "*) continue ;;
+            esac
+            unmapped_vars="$unmapped_vars $abi_var"
+            report_unmapped_sover $abi_var
             continue
         fi
```

Verified by deleting the `MIRAL_ABI` and `MIR_SERVER_GRAPHICS_PLATFORM_ABI` cases from `rpm_sover_macro()`: the seven affected `$packages` entries produce exactly two stderr lines, in both modes, and the remaining macros still sync. Re-ran the existing mismatch test too (`miral_sover` and `mirplatformgraphics_sover` set to 999) — still exactly two deduped `ABI mismatch:` lines and exit 1 under `--check`, and update mode round-trips `rpm/mir.spec` back to byte-identical.

One thing I'd like your call on: I've left this as a warning that doesn't change the exit status. `check_for_unknown_packages()` treats the equivalent `debian/control` case as fatal (`exit 1`). Making an unmapped soversion macro fatal too would be more consistent, but it's a behaviour change you didn't ask for, so I didn't take it unilaterally — happy to do it if you'd prefer.

xhon-pelushi:
Thanks for the ping @RAOF, and sorry for the wait — I've replied to both comments inline.

Short version: I'd like to keep the `rpm/mir.spec` existence check, because it's reachable rather than dead code (`release/1.8` through `release/2.22` all carry `debian/` and this script but no `rpm/`), and the guard-less behaviour is worse than a no-op — `--check` silently exits 0 after nine `grep: ... No such file or directory` lines, and update mode rewrites `debian/` and *then* fails on `sed`, leaving a half-updated tree. But you're right that it shouldn't be silent, so both spots now log unconditionally to stderr in the form you suggested, and the unmapped-soversion-macro case reports through the same style as the existing `report_unknown_package()`, deduped per ABI variable.

Both changes are written and tested locally. The one thing I'd like your call on before I push is whether an unmapped soversion macro should stay a warning or become fatal like `check_for_unknown_packages()` does for `debian/control` — details in the second inline reply.

Unrelated to this PR: the single red check, `BuildAndTest (lxd:alpine-edge:...:amd64)`, is a `-Werror=deprecated-declarations` failure on GoogleTest's `Invoke`/`InvokeWithoutArgs` in `mir::test::doubles::MockDRM` and `MockLibInput`, i.e. entirely in C++ test code. This branch changes one shell script and one markdown file, so it can't be the cause.

xhon-pelushi:
Pushed as `50286285c7`.

I said I'd wait for your call on the unmapped-soversion case before pushing — I've gone with the warning rather than sitting on the whole change for that one question, since it's the smaller-behaviour option and flipping it to fatal is a two-line change if you'd prefer consistency with `check_for_unknown_packages()`. Happy to do that in this PR; just say which.

Everything else is as described above: both missing-`rpm/mir.spec` paths now log unconditionally to stderr, and the unmapped-macro case reports through `report_unmapped_sover()`, deduped per ABI variable.

AlanGriffiths:
@RAOF OK now?

xhon-pelushi:
On the one red job, in case it's holding this up — I don't think it can be this PR.

The test that actually exercises this change is `package-abis`, which `CMakeLists.txt` defines as

```cmake
mir_add_test(NAME package-abis
  COMMAND /bin/sh -c "cd ${PROJECT_SOURCE_DIR} && tools/update_package_abis.sh --check --verbose")
```

In the failing alpine-edge run that test **passed** — `364/366 Test #366: package-abis ... Passed 0.19 sec`.

The single failure is elsewhere:

```
The following tests FAILED:
	 27 - miral-test.XdgDialog.* (Failed)

/spread/mir/tests/miral/xdg_dialog.cpp:276: Failure
Value of: done.wait_for(std::chrono::milliseconds{100})
  Actual: false
Expected: true
```

That's a 100 ms wait timing out in the xdg-dialog-v1 tests, on the one musl target. The other eleven jobs are green, including asan, ubsan, clang and mold. This PR touches only `tools/update_package_abis.sh` and `.github/instructions/abi-symbols.instructions.md`, neither of which is compiled into or linked by `miral-test`.

I haven't been able to show that same test failing on other branches, so I'd put it as timing-sensitive on that runner rather than assert it's a known flake — but the diff can't be reaching it either way. A re-run of just that job should settle it.

RAOF (approved):
Yup, I'm good with this now.

PR #5159, opened 2026-08-18
Fix compile error in PointerMovementEvent matcher: pass result_listener pointer not reference

AlanGriffiths:
This matcher is currently unused, which is why the typo hasn't been spotted


RAOF (approved):
Yup, that's straightforwardly wrong as is.

PR #5168, opened 2026-08-21
Set the cursor role on hidden cursor surfaces

### LINKED ISSUE #5166: Once GNOME showtime hides the cursor it doesn't reappear
### Version

v2.29.0

### Environment

OS: 26.04
CPU:
GPU 1:
GPU 2:
Monitor Setup:

1. Ran a video with showtime
2. Moved the cursor over the surface
3. Waited for the control overlay and cursor to hide
4. Moved the cursor over the surface again

Expect: the control overlay and cursor reappear
Actual: the control overlay reappears, but not the cursor

If the cursor is moved off of the video surface it reappears and remains visible when moved back over the surface

### Reproduction steps

1. Run a video with showtime: `showtime <video>`
2. Move the cursor over the surface
3. Wait for the control overlay and cursor to hide
4. Move the cursor over the surface again

Expect: the control overlay and cursor reappear
Actual: the control overlay reappears, but not the cursor

### Relevant log output

```shell

```

AlanGriffiths:
Fixes: #5166

### DIFF
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -503,6 +503,7 @@ void mf::WlPointer::on_commit(WlSurface* surface)
     {
         // No buffer: We should be unmapping the cursor
 
+        cursor.reset(); // clean up old cursor before creating new one
         cursor = std::make_unique<WlHiddenCursor>(surface, commit_handler);
         if (surface_under_cursor)
             cursor->apply_to(&surface_under_cursor.value());
--- a/src/server/frontend_wayland/wl_pointer.cpp
+++ b/src/server/frontend_wayland/wl_pointer.cpp
@@ -604,6 +604,8 @@ void WlSurfaceCursor::apply_latest_buffer()
 WlHiddenCursor::WlHiddenCursor(mf::WlSurface* surface, mf::CommitHandler* commit_handler) :
     surface_role{surface, std::move(commit_handler)}
 {
+    if (surface)
+        surface->set_role(&surface_role);
 }
 
 void WlHiddenCursor::apply_to(mf::WlSurface* surface)


src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -604,6 +604,8 @@ void WlSurfaceCursor::apply_latest_buffer()
 WlHiddenCursor::WlHiddenCursor(mf::WlSurface* surface, mf::CommitHandler* commit_handler) :
     surface_role{surface, std::move(commit_handler)}
 {
+    if (surface)
+        surface->set_role(&surface_role);
```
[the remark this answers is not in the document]
AlanGriffiths: Good catch! thanks

src/server/frontend_wayland/wl_pointer.cpp
```diff
@@ -503,6 +503,7 @@ void mf::WlPointer::on_commit(WlSurface* surface)
     {
         // No buffer: We should be unmapping the cursor
 
+        cursor.reset(); // clean up old cursor before creating new one
```
mattkae: Why is it not simply cleaned up by being dropped? Should it be?

mattkae: Although I guess this now does what the other path does, so it's fine 👍

mattkae, line 506: That's a fun subtlety to deal with 😄

mattkae (approved):
Good by me!

src/server/frontend_wayland/wl_pointer.cpp
AlanGriffiths: Yes. A symptom of a bad design (C++ does have tools to enforce the correct protocol). But fixing that is out of scope. (For now)

PR #5174, opened 2026-08-25
(wayland-rs) encode C++ exceptions over the FFI boundary such that they always log something sensible

mattkae:
## What's new?
If the C++ side threw some unknown exception, then it ended up being unparseable garbage in the Rust side of things. A better solution is to throw an implementation error on the display object.

- Updated the `mir::wayland_rs::ProtocolError` class to be more verbose and more closely match the existing `mir::wayland::ProtocolError` class
- Added `handle_dispatch_error` which will ensure that errors not matching the sentinel are still reported and handle appropriately
- Added unit tests for the `mir::wayland_rs::ProtocolError` class, which were missing before.

## How to test
New tests pass!

## Checklist

- [x] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
+++ b/src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
@@ -41,6 +41,7 @@ pub fn generate_dispatch_rs(protocols: &Vec<WaylandProtocol>) -> TokenStream {
         mod dispatch {
             use wayland_server::{Client, DataInit, Dispatch, GlobalDispatch, New, DisplayHandle, Resource};
             use wayland_server::backend::{ClientId, ObjectId};
+            use wayland_server::backend::protocol::Interface;
             use crate::protocols;
             use crate::wayland_server_core::ServerState;
             use crate::ffi;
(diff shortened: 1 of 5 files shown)


src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -83,6 +83,31 @@ pub fn generate_dispatch_rs(protocols: &Vec<WaylandProtocol>) -> TokenStream {
...
+            /// error and must not be dressed up as a client protocol violation.
+            fn handle_dispatch_error(resource: &impl Resource, what: &str) {
+                if let Some(encoded) = what.strip_prefix("MIR_PROTOCOL_ERROR:") {
+                    let (object_id, code, message) = parse_post_error(encoded);
+                    post_protocol_error(resource, object_id, code, message);
+                } else {
+                    log::error!("Internal error dispatching Wayland request on {}: {}", resource.id(), what);
+                    // No protocol-level error code fits an internal failure; what matters
+                    // is that the client sees an honest message rather than a garbled
+                    // pseudo-protocol-error, and the real cause is in the server log.
```
RAOF: This is not actually true! [`wl_display::error::implementation`](https://wayland.app/protocols/wayland#wl_display:enum:error) is the correct protocol error to raise.

Whether or not it's sufficiently easy to fish out the display to post that error on is a different question. Is it?

mattkae: You're right! It is hard to get the display though, but I tried my best. Check out the latest implementation :)

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -83,6 +84,66 @@ pub fn generate_dispatch_rs(protocols: &Vec<WaylandProtocol>) -> TokenStream {
                 }
```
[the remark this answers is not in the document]
mattkae: Overly defensive imo, this will be an exception message coming from C++. If that happens, we should probably just crash.

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
RAOF: Huh! I'm surprised it's not easier (this might be a reasonable PR to wayland-server!)

RAOF (approved):
This now looks good to me. Apparently there's a conflict to resolve; feel free to resolve that and land!

PR #5175, opened 2026-08-25
(wayland-rs) drop shared_ptr C++ reference from Rust explicitly on resource destruction

mattkae:
## What's new?
This fixes a bug where Rust keeps the C++ shared ptr (and hence the client object underneath it) alive longer than it should.

The problem is:

- Wrapper owns the C++ shared ptr
- C++ objects owns the middleware which owns the resource handle, which keeps the arc alive.

This bidirectionality is important so that events and requests can be sent from both sides, but it is unfortunate as it causes us to have this explicit release of the shared ptr. This is the most pragmatic solution for this AFAICT.

## Checklist

- [ ] Tests added and pass


RAOF (approved):
Ooof. Wayland-rs code is really difficult to review, but I think I've got a decent set up now.

This seems like a reasonable place to break the loop, and shouldn't make trouble elsewhere.

PR #5176, opened 2026-08-25
(wayland-rs) fix sending an event with a null object param leading to panic

mattkae:
## What's new?
- libwayland was chill with us sending a "zombie" object in a parameter, as it just silently ignored it. We take advantage of this fact in the `wl_keyboard.leave` event, which gets sent after a surface has disappeared. This is not incorrect technically, and I wanted to ensure that it worked the same in Rust.
- If rust encounters an object that is no longer alive in its list, it simply returns quietly, matching the existing C++ implementation

This could be construed as "wrong", but it makes us backwards compatible with the existing implementation, which was correct. I added a "TODO" to update it later if we feel so inclined.

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
+++ b/src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
@@ -185,10 +185,45 @@ fn generate_extension_method_for_event(event: &WaylandEvent, interface_name: &st
         })
         .collect();
 
+    // Sending an event whose target or object argument has been destroyed panics
+    // wayland-backend ("Attempting to send an event with objects from wrong client"),
+    // e.g. wl_keyboard.leave sent from the focused surface's own destruction path.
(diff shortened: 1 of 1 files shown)


RAOF (changes requested):
As mentioned in the standup, we think the correct response to the existing code doing something unreasonable (like sending a leave event to a dead surface) that the new code makes obvious is to fix the existing code so that it doesn't do the unreasonable thing.

mattkae:
> As mentioned in the standup, we think the correct response to the existing code doing something unreasonable (like sending a leave event to a dead surface) that the new code makes obvious is to fix the existing code so that it doesn't do the unreasonable thing.

So we also don't want the compositor crash in the meantime, especially since we don't know all of the possible paths where this could happen. So, we were thinking:

1. Recognize and log the error (loudly) in Rust, but still prevent the `panic!` entirely
2. Fix any sites (e.g. keyboard leave) where this is obviously a current issue

RAOF (approved):
Yup, That'll do.

PR #5180, opened 2026-08-26
Do not send leave event on surface that is being destroyed

mattkae:
## What's new?
- If a surface is in the middle of being destroyed, leave events need not be sent on it. For libwayland, this is okay because it silently ignores these events, but `wayland-rs` will panic if it references a zombie object. Hence, it is safest to ignore events on destroyed surfaces.

## Checklist

- [x] Tests added and pass

pairs with https://github.com/canonical/wlcs/pull/464


src/server/frontend_wayland/wl_surface.cpp
```diff
@@ -131,6 +131,8 @@ mf::WlSurface::WlSurface(
 
 mf::WlSurface::~WlSurface()
 {
+    being_destroyed = true;
+
```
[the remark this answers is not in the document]
AlanGriffiths: It seems tautological to have both `WlSurface` and `LifetimeTracker` tracking the lifetime.

src/server/frontend_wayland/text_input_v2.cpp
```diff
@@ -346,8 +346,13 @@ void mf::TextInputV2::focus_on(WlSurface* surface)
...
     {
-        auto const serial = client->next_serial(nullptr);
-        send_leave_event(serial, current_surface.value().resource);
+        // Don't send a leave event referencing a surface that is being destroyed: the client no
+        // longer cares about it, and doing so would reference a surface that is mid-teardown.
+        if (!current_surface.value().is_being_destroyed())
+        {
+            auto const serial = client->next_serial(nullptr);
+            send_leave_event(serial, current_surface.value().resource);
+        }
```
AlanGriffiths, line 355: This idiom violates the "tell, don't ask" design principle. (I know that can be hard to follow with C APIs).

I'd rather see something like:

```c++
current_surface.value().with_resource([&](auto r) { send_leave_event(serial, r); }
```

(The fact that `resource` is public doesn't encourage this - nor even the use of `raw_resource()`.)

AlanGriffiths (changes requested):
Obviously, the lifetime of the new destroyed flag is an issue. That's a blocker. (Maybe we can push `is_being_destroyed()` up to `LifetimeTracker`?)

Less obviously (and non blocking) the design around here is problematic: we ought to forcing something more like:

```c++
if (auto or = current_surface.and_then([](auto s) { s->resource_if_live(); })
{
    send_leave_event(client->next_serial(nullptr), or.value());
}
```

src/server/frontend_wayland/text_input_v2.cpp
mattkae: We can leave this for another time, but this makes sense :+1:

include/wayland/mir/wayland/lifetime_tracker.h
```diff
@@ -50,12 +50,19 @@ class LifetimeTracker
...
 
 protected:
     /// Subclasses are not required to call this, but may do so during the destruction process if the object needs to
     /// get marked as destroyed and fire its destroy listeners before some other part of the destructor runs.
     void mark_destroyed() const;
 
+    /// Set to true at the start of the destructor (of this class and optionally earlier by subclasses whose own
+    /// destructor runs first) so that destroy listeners can observe that the object is being torn down.
+    bool being_destroyed{false};
+
```
AlanGriffiths: We prefer private data members - could we not set this in `mark_destroyed()`? Or via a similar function?

[aside] why is `mark_destroyed()` **const**?! "Mark" doesn't sound very "const"

AlanGriffiths (approved):
Non blocking style suggestion

src/wayland/lifetime_tracker.cpp
```diff
@@ -69,10 +68,22 @@ void mw::LifetimeTracker::remove_destroy_listener(DestroyListenerId id) const
     }
 }
 
+bool mw::LifetimeTracker::is_being_destroyed() const
+{
+    return impl->being_destroyed;
+}
```
AlanGriffiths, line 74: Elsewhere we check `impl` for null. We should probably do that here (or ensure `impl` is never null)

AlanGriffiths (changes requested):

AlanGriffiths (approved):
Sure. (Assuming the TICS failure is transient)

mattkae:
Abandoned this in favor of putting the fix in wayland-rs, as it appears that Gtk (at least) expects this event: https://matrix.to/#/!ivAdrrLoWkBYEnSIae:matrix.org/$br4dYeucp_lM4fe0JlN6Sh_uvuo4gptw9TJj2RKK3OY?via=matrix.org&via=ubuntu.com&via=fedora.im

PR #5183, opened 2026-08-27
wayland_rs: use a valid, lazily-evaluated enum fallback

mattkae:
## Problem
The generated event middleware converted integer arguments into wayland-rs enums with:

```rust
    Enum::try_from(value).unwrap_or(Enum::try_from(0).unwrap())
```

This is wrong in two ways:

1. `unwrap_or` is eager so the fallback `Enum::try_from(0).unwrap()` is always evaluated even when the primary conversion succeeds
2. For any enum without a 0-valued entry `try_from(0)` returns `Err`, so `.unwrap()` panics unconditionally on every event carrying such an argument.

## Solution
Precompute a per-enum fallback that is guaranteed to convert (0 for bitfields or enums with a 0 entry, otherwise the enum's first declared entry) and emit a lazy `unwrap_or_else` so the fallback only runs on genuinely invalid input.

In reality, only xdg decoration has this problem (see [here](https://wayland.app/protocols/xdg-decoration-unstable-v1#zxdg_toplevel_decoration_v1:enum:mode)), but this will future proof it :tada: 

## Checklist

- [ ] Tests added and pass


src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -240,16 +287,25 @@ fn wayland_arg_to_rust_param_str(arg: &WaylandArg, interface_name: &str) -> Stri
...
-        let rust_enum_path = if let Some(dot_pos) = e.find('.') {
-            let interface = &e[..dot_pos];
-            let enum_name = &e[dot_pos + 1..];
-            format!("{}::{}", interface, snake_to_pascal(enum_name))
+        let (enum_interface, enum_name) = if let Some(dot_pos) = e.find('.') {
+            (&e[..dot_pos], &e[dot_pos + 1..])
         } else {
-            format!("{}::{}", interface_name, snake_to_pascal(e))
+            (interface_name, e.as_str())
         };
```
andogq:
```suggestion
        let (enum_interface, enum_name) = e.split_once('.').unwrap_or((interface_name, e.as_str()));
```

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -18,6 +18,34 @@ use crate::protocol_parser::{
...
+                    return None;
+                };
+                let zero_is_valid =
+                    e.bitfield.unwrap_or(false) || e.entries.iter().any(|entry| entry.value == 0);
+                let fallback = if zero_is_valid {
+                    0
+                } else {
+                    e.entries.first().map_or(0, |entry| entry.value as u32)
+                };
+                Some(((interface.name.clone(), e.name.clone()), fallback))
```
andogq:
```suggestion
                Some(((interface.name.as_str(), e.name.as_str()), fallback))
```

See: https://github.com/canonical/mir/pull/5183/changes#r3891234011

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -240,16 +287,25 @@ fn wayland_arg_to_rust_param_str(arg: &WaylandArg, interface_name: &str) -> Stri
...
+            (&e[..dot_pos], &e[dot_pos + 1..])
         } else {
-            format!("{}::{}", interface_name, snake_to_pascal(e))
+            (interface_name, e.as_str())
         };
+        let rust_enum_path = format!("{}::{}", enum_interface, snake_to_pascal(enum_name));
+        // The fallback must be lazy (`unwrap_or_else`) and use a value that is
+        // actually valid for this enum: not every enum has a 0-valued entry.
+        let fallback = enum_fallbacks
+            .get(&(enum_interface.to_string(), enum_name.to_string()))
```
andogq:
```suggestion
            .get(&(enum_interface, enum_name))
```

See: https://github.com/canonical/mir/pull/5183/changes#r3891234011

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -18,6 +18,34 @@ use crate::protocol_parser::{
...
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
+use std::collections::HashMap;
+
+/// Maps (interface_name, enum_name) to a value that is guaranteed to convert
+/// into the wayland-rs enum: 0 when valid (bitfields accept empty flags, and
+/// many enums have a 0-valued entry), otherwise the enum's first entry.
+type EnumFallbacks = HashMap<(String, String), u32>;
+
+fn build_enum_fallback_map(protocols: &[WaylandProtocol]) -> EnumFallbacks {
```
andogq:
```suggestion
type EnumFallbacks<'proto> = HashMap<(&'proto str, &'proto str), u32>;

fn build_enum_fallback_map<'proto>(protocols: &'proto [WaylandProtocol]) -> EnumFallbacks<'proto> {
```

Given the fallbacks only exist within the lifetime of `protocols`, we can use string slices as keys instead of needing allocations for every lookup.

- https://github.com/canonical/mir/pull/5183/changes#r3891229989
- https://github.com/canonical/mir/pull/5183/changes#r3891232220

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -240,16 +287,25 @@ fn wayland_arg_to_rust_param_str(arg: &WaylandArg, interface_name: &str) -> Stri
...
+        // The fallback must be lazy (`unwrap_or_else`) and use a value that is
+        // actually valid for this enum: not every enum has a 0-valued entry.
+        let fallback = enum_fallbacks
+            .get(&(enum_interface.to_string(), enum_name.to_string()))
+            .unwrap_or_else(|| {
+                panic!(
+                    "Unknown enum '{}' referenced by argument '{}' of interface '{}'",
+                    e, arg.name, interface_name
+                )
+            });
```
andogq: A `let..else` guard might convey intention slightly clearer (we're always going to panic, never provide a fallback-fallback).

```suggestion
        let Some(fallback) = enum_fallbacks.get(&(enum_interface, enum_name)) else {
            panic!(
                "Unknown enum '{}' referenced by argument '{}' of interface '{}'",
                e, arg.name, interface_name
            );
        };
```

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -18,6 +18,34 @@ use crate::protocol_parser::{
...
+            interface.items.iter().filter_map(move |item| {
+                let InterfaceItem::Enum(e) = item else {
+                    return None;
+                };
+                let zero_is_valid =
+                    e.bitfield.unwrap_or(false) || e.entries.iter().any(|entry| entry.value == 0);
+                let fallback = if zero_is_valid {
+                    0
+                } else {
+                    e.entries.first().map_or(0, |entry| entry.value as u32)
```
andogq: This `map_or` feels like it may inadvertently lead to a similar bug, if the enum doesn't contain any variants (granted, I'm assuming unlikely). Regardless, it wouldn't hurt to include an assertion verifying the enum does has zeroth field.

andogq (commented):
Mainly some style-adjacent suggestions.

---

An alternative would be to implement a method on `WaylandEnum` which does the same lookup as in `build_enum_fallback_map` for the current enum (ie. check `self.bitfield`, iterate through `self.items` for `0`, otherwise return `self.entries[0]`. Granted, it wouldn't be cached, but you'd save all the allocations. Also could prevent passing around that extra parameter everywhere, and the `WaylandEnum` could be directly looked-up from `WaylandProtocol`.

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
mattkae: Nice 😄

mattkae:
> An alternative would be to implement a method on `WaylandEnum` which does the same lookup as in `build_enum_fallback_map` for the current enum (ie. check `self.bitfield`, iterate through `self.items` for `0`, otherwise return `self.entries[0]`. Granted, it wouldn't be cached, but you'd save all the allocations. Also could prevent passing around that extra parameter everywhere, and the `WaylandEnum` could be directly looked-up from `WaylandProtocol`.

That's a very good point! I tend to opt for build-time resolution in wayland-rs over runtime resolution when I can, but I have no real reason to do this other than "it will be marginally faster for the user at runtime".

However in this case, I think it is so marginal, that it is worth implementing your suggestion. I will make this change!

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -19,6 +19,59 @@ use crate::protocol_parser::{
...
+                        // Prefer 0 when the enum accepts it (bitfields always do; some
+                        // enums define it), otherwise the first defined value.
+                        Self::try_from(0u32).unwrap_or_else(|_| Self::try_from(#first).unwrap())
+                    }
+                }
+            })
+        });
+
+    quote! {
+        trait EnumFallback: Sized {
```
andogq, line 102: If we can implement `Default` for these enums instead, `unwrap_or_default()` could be used instead.

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -19,6 +19,59 @@ use crate::protocol_parser::{
...
+            let enum_path: syn::Path =
+                syn::parse_str(&format!("{}::{}", interface_name, snake_to_pascal(&e.name)))
+                    .expect("Failed to parse enum path");
+            let first = e.entries.first().unwrap().value as u32;
+            Some(quote! {
+                impl EnumFallback for #enum_path {
+                    fn fallback() -> Self {
+                        // Prefer 0 when the enum accepts it (bitfields always do; some
+                        // enums define it), otherwise the first defined value.
+                        Self::try_from(0u32).unwrap_or_else(|_| Self::try_from(#first).unwrap())
```
andogq: It'd be nice to directly generate the variant, instead of relying on runtime `try_from` invocation (would completely remove the `panic` path), but I'd imagine it mightn't quite work that nicely for bitsets.
If it's not simple to do, or you don't think it's appropriate, don't worry.

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -19,6 +19,59 @@ use crate::protocol_parser::{
...
+                Some((interface.name.as_str(), e))
+            })
+        })
+        .filter_map(|(interface_name, e)| {
+            if !seen.insert((interface_name, e.name.as_str())) {
+                return None;
+            }
+            let enum_path: syn::Path =
+                syn::parse_str(&format!("{}::{}", interface_name, snake_to_pascal(&e.name)))
+                    .expect("Failed to parse enum path");
```
andogq: `parse_quote` gives us a little shortcut.
```suggestion
            let enum_name = snake_to_pascal(&e.name);
            let enum_path: syn::Path = syn::parse_quote!(#interface_name::#enum_name);
```

andogq (commented):
Apologies, I wasn't very clear in my original suggestion. I completely agree with preferring build-time over runtime!

Two small ideas for this change, feel free to ignore anything that doesn't seem right to you.

<details>
<summary>Here's a better explanation of my original thought, happy for you to take it or leave it, as your alternate idea works fine too!</summary>

What I was intending to say was implement a method on `WaylandEnum` (defined in `protocol_parser.rs`) which performs the resolution of the default value. Then in `wayland_arg_to_rust_param_str`, lookup the instance of `WaylandEnum` from `protocols`, and perform the resolution at that point. Essentially, replacing the original `HashMap` with information derived from the `WaylandProtocol` itself.

```rs
// protocol_parser.rs
pub struct WaylandEnum {
    // ...
}
impl WaylandEnum {
    pub fn resolve_default_value(&self) -> u32 {
        if self.bitfield.unwrap_or(false) {
            return 0;
        }

        self.entries
            .iter()
            .find(|entry| entry.value == 0)
            .or(self.entries.first())
            .expect("non-nonsensical enum")
            .value
    }
}

// protocol_middleware_generation.rs
fn wayland_arg_to_rust_param_str(
    protocol: &WaylandProtocol, // <-- new!
    arg: &WaylandArg,
    interface_name: &str,
) -> String {
    // ...
    if let Some(e) = &arg.enum_ {
        //...

        // Pull the enum definition out of the protocol, and resolve
        // the default value at that point.
        let default_value = protocol
            .interfaces
            .iter()
            .find(|interface| interface.name == enum_interface)
            .into_iter()
            .flat_map(|interface| &interface.items)
            .filter_map(|item| {
                if let InterfaceItem::Enum(enum_item) = item {
                    Some(enum_item)
                } else {
                    None
                }
            })
            .find(|enum_item| enum_item.name == enum_name)
            .expect("valid wayland enum")
            .resolve_default_value();

        param = format!(
            "{}::try_from({}).unwrap_or(|_| {}::try_from({}).unwrap())",
            rust_enum_path, param, rust_enum_path, default_value
        );

    }

}
```
</details>

mattkae:
> ```rust
> WaylandEnum
> ```

Ah that makes sense! Sorry, I had taken `WaylandEnum` as the generated Wayland enum :) I think this makes sense

mattkae:
> ```rust
> ...
> // protocol_middleware_generation.rs
> fn wayland_arg_to_rust_param_str(
>     protocol: &WaylandProtocol, // <-- new!
>     arg: &WaylandArg,
>     interface_name: &str,
> ) -> String {
>     // ...
>     if let Some(e) = &arg.enum_ {
>         //...
> 
>         // Pull the enum definition out of the protocol, and resolve
>         // the default value at that point.
>         let default_value = protocol
>             .interfaces
>             .iter()
>             .find(|interface| interface.name == enum_interface)
>             .into_iter()
>             .flat_map(|interface| &interface.items)
>             .filter_map(|item| {
>                 if let InterfaceItem::Enum(enum_item) = item {
>                     Some(enum_item)
>                 } else {
>                     None
>                 }
>             })
>             .find(|enum_item| enum_item.name == enum_name)
>             .expect("valid wayland enum")
>             .resolve_default_value();
> 
>         param = format!(
>             "{}::try_from({}).unwrap_or(|_| {}::try_from({}).unwrap())",
>             rust_enum_path, param, rust_enum_path, default_value
>         );
> 
>     }
> 
> }
> ```

One important thing to call out here is that the enum that we're referencing does not necessarily originate from the protocol, so we would potentially have to iterate all protocols to resolve the enum (e.g. [`ext-image-copy-capture-v1.xml` ](https://wayland.app/protocols/ext-image-copy-capture-v1#ext_image_copy_capture_session_v1:event:shm_format)). 

Doing that large search feels more complicated than resolving it at runtime, maybe with an `unwrap_or_default`. Let me see how it feels

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
mattkae: We can't unfortunately.

For the core Wayland protocol, the enum is generated in the `wayland-server` crate, and hence we can't define `Default` on a foreign type because it violates Rust's "orphan rule".

We'd have to then special case those enums, which would lead to even more confusion. However, we can define `EnumFallback` locally, since a local trait can be implemented on a foreign type. So that was our escape hatch :)

mattkae:
@andogq I am going to leave the runtime fallback for now for two reasons:

1. Iterating all of the protocols to find an enum each time is going to be a bit hefty for the build, and creating the map to start and then passing it around also muddies the code a whole lot
2. Using a fallback method at runtime is actually pretty clean and - in practice - will only happen very rarely. It also leads to the least complicated generation code, which is a huge plus for maintainability in my opinion.

Let me know if you disagree! Thanks for the great feedback :smile:

src/server/frontend_wayland/wayland_rs/build_script/protocol_middleware_generation.rs
```diff
@@ -19,6 +19,94 @@ use crate::protocol_parser::{
...
+            } else {
+                // Prefer the `0`-valued entry when present, otherwise the first
+                // defined entry, and reference the generated variant directly.
+                let entry = e
+                    .entries
+                    .iter()
+                    .find(|entry| entry.value == 0)
+                    .unwrap_or_else(|| e.entries.first().unwrap());
+                let variant = format_ident!("{}", wayland_enum_variant_ident(&entry.name));
+                quote! { Self::#variant }
```
andogq, line 90: Nice!

andogq (approved):
> the enum is generated in the wayland-server crate

> the enum that we're referencing does not necessarily originate from the protocol

Good things to know, thanks!

> I am going to leave the runtime fallback

Can’t argue with that :)

We got there in the end, I'm happy with it now! :shipit:

Issue #5125, opened 2026-08-11
Build instructions are incomplete


AlanGriffiths:
### Version

main

### Environment

OS: debian/sid

### What happened? What did you expect to happen?

Following the build instructions, there seem to be missing dependencies.

```
CMake Error at cmake/RustLibrary.cmake:1 (find_program):
  Could not find CARGO_EXECUTABLE using the following names: cargo
Call Stack (most recent call first):
  src/platforms/evdev-rs/CMakeLists.txt:1 (include)
```

Obviously `cmake -DMIR_ENABLE_RUST=false -B build` works. As does running `sudo apt install cargo` first.

Do we want to mark `cargo` as an explicit build dependency? Or mention `MIR_ENABLE_RUST=false` in the instructions, Or something else.

### Relevant log output

```shell

```

tarek-y-ismail:
I believe that with #5185, this is irrelevant now?

PR #5185, opened 2026-08-27
Drop MIR_ENABLE_RUST from the build, since Rust will be required in 26.04

mattkae:
## What's new?
To land the `wayland-rs` work, Rust is absolutely required. Now that we target 26.04 strictly, we have a version of cxx that works with our desired version. Hence, it should be safe to remove the feature flag and build with the Rust code always enabled.

If CI passes, we did it :tada: 

## Checklist

- [ ] Tests added and pass


saviq-work:
This should fix this one:
- #5125

debian/control
```diff
@@ -416,6 +428,21 @@ Description: Display server for Ubuntu - input platform library
  Contains the shared libraries required for the Mir server to interact with
  the input hardware using the evdev interface.
 
+Package: mir-platform-input-evdev-rs11
```
saviq-work, line 425: Wonder if we should just stick it in the existing package?

debian/control
```diff
@@ -24,6 +30,10 @@ Build-Depends: cmake,
                libinput-dev (>= 0.21),
                liblttng-ust-dev,
                libpixman-1-dev,
+# The target-architecture Rust standard library. rustc and cargo run natively
+# (see the :native markers), but cross-compiling the Rust components still needs
+# std for the host architecture, which nothing else in this list pulls in.
```
saviq-work: Don't think this comment is needed, we already have a bunch of `:native` (and not) deps.
```suggestion
```

debian/control
```diff
@@ -3,7 +3,13 @@ Section: x11
 Priority: optional
 Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
 XSBC-Original-Maintainer: Thomas Voß <thomas.voss@canonical.com>
-Build-Depends: cmake,
+Build-Depends: cargo:native,
+# Cargo fetches the Rust components' crate dependencies from crates.io over
+# HTTPS, so it needs the system CA bundle. Minimal build chroots don't pull
+# this in otherwise (apt uses a plain-http mirror), leaving cargo to fail with
+# "Problem with the SSL CA cert".
```
saviq-work:
```suggestion
# Needed for Cargo to fetch crates
```

cmake/RustLibrary.cmake
```diff
@@ -27,9 +27,18 @@ function(add_rust_cxx_library target)
   set(cxxbridge_source "${cxxbridge_include_dir}/${arg_CRATE}/${arg_CXX_BRIDGE_SOURCE_FILE}.cc")
   set(crate_staticlib "${rust_binary_dir}/lib${arg_CRATE}.a")
 
+  # The output paths above are nested under the target triple, which Cargo only
+  # emits when built with `--target`. Pass it when DEB_HOST_RUST_TYPE is set (deb
+  # packaging); otherwise Cargo writes to target/{release,debug} and the empty
+  # segment collapses to the same path (local builds, snap).
```
saviq-work:
```suggestion
  # DEB_HOST_RUST_TYPE is set in deb builds, which expect the build under
  # the architecture triplet.
```

debian/rules
```diff
@@ -1,10 +1,38 @@
...
+include /usr/share/rustc/architecture.mk
 include debian/opts.mk
 
 export DPKG_GENSYMBOLS_CHECK_LEVEL = 4
 
+# The Rust components (evdev-rs, wayland-rs) are built by CMake via the real
+# Cargo (/usr/bin/cargo from the `cargo` package), which fetches and compiles
+# their crate dependencies from crates.io, pinned by the committed Cargo.lock.
+# DEB_HOST_RUST_TYPE is the target triple that cmake/RustLibrary.cmake passes to
+# `cargo build --target`, so the crate artifacts land where CMake expects them.
```
saviq-work:
```suggestion
```

debian/rules
```diff
@@ -1,10 +1,38 @@
...
+#    debhelper's cmake buildsystem sets this for its own configure/test steps
+#    (which is how the C/C++ libraries locate libudev), but NOT for the build
+#    step that actually runs Cargo, so the crate would otherwise fall back to
+#    the native pkg-config and fail to find the cross libudev. Export it here so
+#    it applies to every step. For native builds this is the ordinary wrapper.
+#  * PKG_CONFIG_ALLOW_CROSS opts past the crate's guard that otherwise aborts
+#    with "pkg-config has not been configured to support cross-compilation".
+#    It is a no-op for native builds (host == target).
+export PKG_CONFIG = /usr/bin/$(DEB_HOST_GNU_TYPE)-pkg-config
+export PKG_CONFIG_ALLOW_CROSS = 1
```
saviq-work: I'm surprised this is only needed for Rust…?

How does CMake know to do this outside of Rust?

```suggestion
# Needed for cross-compiling Rust
export PKG_CONFIG = /usr/bin/$(DEB_HOST_GNU_TYPE)-pkg-config
export PKG_CONFIG_ALLOW_CROSS = 1
```

debian/rules
```diff
@@ -27,11 +55,20 @@ else
...
 
+# Disable LTO: the Rust components (evdev-rs, wayland-rs) are compiled by rustc
+# and their static archives are not GCC LTO objects. Linking them into
+# libmirserver.so under GCC + ld.bfd's LTO plugin produces spurious
+# "undefined reference" errors for COMDAT template instantiations pulled in
+# from the fat-LTO C++ archives (e.g. boost::throw_exception<>, the
+# std::string move constructor, vtable for boost::exception). Other linkers
+# (mold, lld) and LTO-disabled builds are unaffected, so drop LTO for the
+# whole package now that Rust is always built.
+DEB_BUILD_MAINT_OPTIONS += optimize=-lto
```
saviq-work: If we just don't support LTO any more, this feels like the wrong place to do this.

debian/rules
```diff
@@ -100,6 +137,11 @@ override_dh_makeshlibs:
 override_dh_strip:
 	dh_strip -Xmir-test-data
 
+# Remove the Cargo home set up for the Rust components
```
saviq-work: Really? :P
```suggestion
```

src/server/frontend_wayland/wayland_connector.cpp
```diff
@@ -268,12 +257,6 @@ mf::WaylandConnector::WaylandConnector(
     std::shared_ptr<input::CursorObserverMultiplexer> const& cursor_observer_multiplexer)
     : extension_filter{extension_filter},
       display{wl_display_create(), &cleanup_display},
-      // TODO(mattkae): Run the server and hook it up to the rest of the system
-      #ifdef MIR_ENABLE_RUST
-      server_wrapper{std::make_unique<ServerWrapper>(wayland_rs::create_wayland_server())},
-      #else
-      server_wrapper{std::make_unique<ServerWrapper>()},
-      #endif
```
saviq-work, line 276: We don't need `ServerWrapper` no more?

saviq-work (changes requested):
Save some bytes (and tokens!), man :)

src/server/frontend_wayland/wayland_connector.cpp
mattkae: It changes so much with the next PR that it is not even worth keeping around for now, and it is was behind the flag anyway, so who cares. Might as well remove it with the flag!

debian/control
mattkae: I did that at first too, but I don't know how to feel about it. This felt safest for now. I am ok either way

debian/rules
mattkae: I removed the code below and kept this. WDYT? Or is there another place we should do this?

debian/control
saviq-work: Had a chat about this in standup, should be fine in existing package - that's actually safer b/c we can then do it step by step, with no changes to packaging:
1. include `evdev-rs`
2. default to `evdev-rs`
3. remove `evdev`

debian/rules
saviq-work: This only works for Debian (assuming it will use our packaging, which the official packages, don't).

If we can't build with LTO, we should make sure we add `-fno-lto` to our flags, ideally localized to the target where that's necessary.

debian/control
mattkae: Done!

debian/rules
mattkae: Done!

debian/rules
saviq-work: @mattkae these were not dealt with:
> I'm surprised this is only needed for Rust…?

> How does CMake know to do this outside of Rust?

src/server/CMakeLists.txt
```diff
@@ -102,6 +102,10 @@ add_library(mirserver-static STATIC) # Internal-only use: For linking into test
 target_link_libraries(mirserver PUBLIC ${MIR_SERVER_REFERENCES})
 target_link_libraries(mirserver-static PUBLIC ${MIR_SERVER_REFERENCES})
 
+# The Rust static archives (via mirwayland_rs) are not GCC LTO objects.
+target_link_options(mirserver PRIVATE $<$<CXX_COMPILER_ID:GNU>:-fno-lto>)
+target_link_options(mirserver-static PRIVATE $<$<CXX_COMPILER_ID:GNU>:-fno-lto>)
```
saviq-work, line 107: Should propagate this through the `.pc` file, then?

saviq-work (changes requested):
Needs rebasing, and a couple things still to work out.

debian/rules
mattkae: I added a comment explaining why it's necessary :+1:

src/server/CMakeLists.txt
mattkae: I do not think so. `mirserver` is shared, so the non-LTO Rust archives are baked into `libmirserver.so`, meaning that consumer never re-link to the Rust objects, right?

And `mirserver-static` is private, so no one uses that anyway.

debian/rules
saviq-work: @mattkae couldn't we instead "inherit" the debhelper handling? We invoke `cargo` from CMake after all, and if someone else tries cross-compilation outside of packaging, they'll hit t he problem this deals with?

mattkae: u rite

saviq-work (approved):

PR #5187, opened 2026-08-28
Handle unmanaged surfaces in window_at safely

tarek-y-ismail:
## What's new?

- `window_at` now returns an empty Window when the focused surface is not present in window_info
-  Add a regression test covering that case.

## How to test

- Run the test with and without the patch applied.

## Checklist

- [x] Tests added and pass

### DIFF
--- a/tests/miral/test_window_manager_tools.cpp
+++ b/tests/miral/test_window_manager_tools.cpp
@@ -22,7 +22,6 @@
 #include <miral/output.h>
 
 #include <mir/shell/display_layout.h>
-#include <mir/shell/focus_controller.h>
 #include <mir/shell/persistent_surface_store.h>
 #include <mir/graphics/display_configuration_observer.h>
 #include <mir/wayland/weak.h>
@@ -40,36 +39,6 @@
 
 namespace
 {
-
-struct StubFocusController : mir::shell::FocusController
-{
-    void focus_next_session() override {}
-    void focus_prev_session() override {}
-
-    auto focused_session() const -> std::shared_ptr<mir::scene::Session> override { return {}; }
-
-    void set_popup_grab_tree(std::shared_ptr<mir::scene::Surface> const& /*surface*/) override {}
-
-    void set_focus_to(
-        std::shared_ptr<mir::scene::Session> const& /*focus_session*/,
-        std::shared_ptr<mir::scene::Surface> const& /*focus_surface*/) override {}
-
-    auto focused_surface() const -> std::shared_ptr<mir::scene::Surface> override { return {}; }
-
-    void raise(mir::shell::SurfaceSet const& /*windows*/) override {}
-
-    virtual auto surface_at(mir::geometry::Point /*cursor*/) const -> std::shared_ptr<mir::scene::Surface> override
-        { return {}; }
-
-    void swap_z_order(mir::shell::SurfaceSet const& /*first*/, mir::shell::SurfaceSet const& /*second*/) override {}
-
-    void send_to_back(mir::shell::SurfaceSet const& /*windows*/) override {}
-
-    auto is_above(std::weak_ptr<mir::scene::Surface> const& /*a*/, std::weak_ptr<mir::scene::Surface> const& /*b*/) const -> bool override
-    {
-        return false;
-    }
-};
 
 struct StubDisplayLayout : mir::shell::DisplayLayout
 {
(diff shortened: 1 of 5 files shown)


mattkae (approved):
Looks straightforward to me 👍

tests/miral/test_window_manager_tools.cpp
```diff
@@ -59,7 +59,9 @@ struct StubFocusController : mir::shell::FocusController
     void raise(mir::shell::SurfaceSet const& /*windows*/) override {}
 
     virtual auto surface_at(mir::geometry::Point /*cursor*/) const -> std::shared_ptr<mir::scene::Surface> override
-        { return {}; }
+        { return surface; }
+
+    std::shared_ptr<mir::scene::Surface> surface;
```
AlanGriffiths: This looks like `StubFocusController` is becoming a less of a stub. The idiomatic way to do that here is:

```c++
MOCK_METHOD(std::shared_ptr<mir::scene::Surface>, surface_at, (mir::geometry::Point), (const override));
```

And in the test

```c++
        ON_CALL(focus_controller, surface_at(_))
            .WillByDefault([&](auto) { return create_surface(session, {}); });
```

(Yes, I know the `TestWindowManagerTools` doesn't currently expose focus_controller, but if we're going to mock the calls it receives then we should.)

AlanGriffiths (changes requested):

tests/miral/test_window_manager_tools.cpp
tarek-y-ismail: Why do we use pimpl here anyway? I had to improvise a bit to get to `focus_controller` through `Self`.

tests/miral/test_window_manager_tools.cpp
```diff
@@ -375,6 +374,22 @@ auto mt::TestWindowManagerTools::create_and_select_window_for_session(
     return result;
 }
 
+void mt::TestWindowManagerTools::create_new_surface_in_surface_at()
+{
+    ON_CALL(self->focus_controller, surface_at(testing::_))
+        .WillByDefault(testing::Return(create_surface(session, {})));
+}
```
AlanGriffiths: Please inline this function so the test setup remains in the test body

tests/miral/test_window_manager_tools.cpp
```diff
@@ -375,6 +374,22 @@ auto mt::TestWindowManagerTools::create_and_select_window_for_session(
...
+
+struct WindowAt : mt::TestWindowManagerTools
+{
+};
+
+TEST_F(WindowAt, returns_empty_for_unmanaged_surface)
```
AlanGriffiths: I'm not convinced by the test name.
```suggestion
TEST_F(BasicWindowManager, window_at_returns_empty_for_unmanaged_surface)
```

AlanGriffiths (changes requested):
I'm not convinced that a test of `BasicWindowManager::window_at()` belongs in `test_window_manager_tools.cpp`

tests/miral/test_window_manager_tools.h
```diff
@@ -102,6 +133,7 @@ class TestWindowManagerTools : public testing::Test
     /// Creates a new session, adds the provided surface to the session, and then sets the resulting window as
     /// the active window.
     auto create_and_select_window_for_session(mir::shell::SurfaceSpecification&, std::shared_ptr<scene::Session>) -> miral::Window;
+
```
AlanGriffiths: Spurious whitespace change?

tests/miral/test_window_manager_tools.h
```diff
@@ -38,6 +39,35 @@ class DisplayConfiguration;
...
+    void swap_z_order(mir::shell::SurfaceSet const& /*first*/, mir::shell::SurfaceSet const& /*second*/) override {}
+
+    void send_to_back(mir::shell::SurfaceSet const& /*windows*/) override {}
+
+    auto is_above(std::weak_ptr<mir::scene::Surface> const& /*a*/, std::weak_ptr<mir::scene::Surface> const& /*b*/) const -> bool override
+    {
+        return false;
+    }
+};
```
AlanGriffiths, line 69: This could be in `mir::test::doubles` and is still more "stub" than "mock". But OK for now.

AlanGriffiths (approved):
Nits

PR #5189, opened 2026-08-28
Draft changelog

AlanGriffiths:


AlanGriffiths:
I think this is now the right place for a changelog entry - but put on a branch in case I have it wrong

PR #5191, opened 2026-09-01
all: drop toolchain workarounds

saviq-work:
## What's new?

Greatly reduced `-Wno-` suppressions.

## How to test

CI, and ppa:mir-team/test

## Checklist

- [ ] Tests added and pass

### DIFF
--- a/include/test/mir/test/event_matchers.h
+++ b/include/test/mir/test/event_matchers.h
@@ -83,7 +83,7 @@ auto inline touch_ids_match(
 template<typename T>
 auto inline modifiers_match(T const& expected, int actual, testing::MatchResultListener* result_listener) -> bool
 {
-    if (expected != actual)
+    if (expected != static_cast<T>(actual))
     {
         *result_listener << "Expected modifiers (" << expected << ") do not match actual (" << actual << ")";
         return false;
--- a/include/test/mir/test/event_matchers.h
+++ b/include/test/mir/test/event_matchers.h
@@ -78,12 +78,12 @@ auto inline touch_ids_match(
     return true;
 }
 
-/// Takes in a Mir*EventModifiers and an int bitmask and determines if they are the same.
+/// Takes in two Mir*EventModifiers and determines if they are the same.
 /// If not, the discrepancy is logged to the MatchResultListener.
 template<typename T>
-auto inline modifiers_match(T const& expected, int actual, testing::MatchResultListener* result_listener) -> bool
+auto inline modifiers_match(T const& expected, T const& actual, testing::MatchResultListener* result_listener) -> bool
 {
-    if (expected != static_cast<T>(actual))
+    if (expected != actual)
     {
         *result_listener << "Expected modifiers (" << expected << ") do not match actual (" << actual << ")";
         return false;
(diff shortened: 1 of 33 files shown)


saviq-work:
@copilot fix the issues found by review

src/server/run_mir.cpp
```diff
@@ -198,21 +198,18 @@ extern "C" [[noreturn]] void fatal_signal_cleanup(int sig, siginfo_t* info, void
...
          */
-        mir::fatal_error("Unsupported attempt to continue after a fatal signal: %s", signum_to_string(sig).c_str());
+        MIR_FATAL_ERROR("Unsupported attempt to continue after a fatal signal: {}", signum_to_string(sig));
     }
     // The handler doesn't care about fancy context, we can just call it via raise()
     std::raise(sig);
     // We definitely can't continue, though, even if their handler tries.
-    mir::fatal_error("Unsupported attempt to continue after a fatal signal: %s", signum_to_string(sig).c_str());
-#pragma GCC diagnostic pop
+    MIR_FATAL_ERROR("Unsupported attempt to continue after a fatal signal: {}", signum_to_string(sig));
```
AlanGriffiths: This change is wrong: We intend to use the deprecated `mir::fatal_error(()` here. (It isn't async-signal-safe in theory but, in practice, is far more likely to work than `MIR_FATAL_ERROR`)

include/test/mir/test/event_matchers.h
```diff
@@ -83,7 +83,7 @@ auto inline touch_ids_match(
 template<typename T>
 auto inline modifiers_match(T const& expected, int actual, testing::MatchResultListener* result_listener) -> bool
 {
-    if (expected != actual)
+    if (expected != static_cast<T>(actual))
```
AlanGriffiths: Is this actually the right fix? Why is `actual` an `int` in the first place (and not a `T`)

src/server/shell/decoration/decoration_strategy.cpp
```diff
@@ -667,27 +667,43 @@ auto RendererStrategy::Text::Impl::font_path() -> std::string
...
+    std::unique_ptr<gunichar, decltype(&g_free)> const converted{
+        g_utf8_to_ucs4_fast(text.data(), static_cast<glong>(validated_length), &length),
+        &g_free};
+    if (!converted)
+    {
+        mir::log_warning("Failed to convert window title %s from UTF-8", text.c_str());
+        return ascii_fallback();
+    }
+
+    return {converted.get(), converted.get() + length};
```
AlanGriffiths: This (and the associated changes) seem like unnecessary churn: While `std::wstring_convert<>` is technically deprecated in the C++ standard library, it must be supported by implementations and there is no movement towards replacing or removing it.

I would stick with what we have rather think about new code and introduce a new dependency.

AlanGriffiths (changes requested):
Mostly good. A couple of suggestions

include/test/mir/test/event_matchers.h
saviq-work: Fixed.

src/server/shell/decoration/decoration_strategy.cpp
saviq-work: Undone.

tarek-y-ismail:
@AlanGriffiths LGTY?

src/server/shell/decoration/CMakeLists.txt
```diff
@@ -32,4 +32,5 @@ target_link_libraries(mirshelldecoration
     mircore
   PRIVATE
     PkgConfig::FREETYPE
+    PkgConfig::GLIB
```
AlanGriffiths: Spurious?

AlanGriffiths (changes requested):
One nit

src/server/shell/decoration/CMakeLists.txt
saviq-work:
```suggestion
```

AlanGriffiths (approved):
Thanks for the cleanup!

PR #5193, opened 2026-09-02
frontend_wayland: add wl_fixes support

andogq:
See #5014 for C++ changes

## What's new?

- `wl_fixes` implemented in `wayland_rs` (not exposed via FFI)
- `implemented_protocols` helper to generate a build-time list of `wayland_rs` protocol implementations
- `protocol_requires_ffi_codegen` helper to conditionally skip some parts of codegen (instead of `interface_name != "wl_display" && interface_name != "wl_registry" && /* all the implemented protocols */`)

## Questions

- is `protocol_requires_ffi_codegen` too coarse, would it be better to stick with only a helper for protocol implementations?
- any strong opinions about registering the non-FFI protocols (currently [here](https://github.com/canonical/mir/pull/5193/changes#diff-91de74eea46f62f1afc036221072655d40344a58ef042f529ee06a0582444854R5))
- given these non-FFI protocols don't require access to `GlobalFactory`, is it fine to just use `()` as the data?

## Checklist

- [ ] Tests added and pass


src/server/frontend_wayland/wayland_rs/build_script/wayland_server_generation.rs
```diff
@@ -44,9 +44,17 @@ fn generate_register_globals_impl(protocol: &WaylandProtocol) -> Vec<TokenStream
...
+        .iter()
+        .cloned()
+        .chain(
+            // wl_display is handled specially in wayland_server crate via the 'Display' struct.
+            std::iter::once("wl_display"),
+        )
+        .any(|protocol| interface.name == protocol)
+    {
+        return None;
+    }
```
RAOF: I think this is incorrectly indented? Strange, I would have thought our precommit would have fixed this up. (Are we actually running `rustfmt`?)

src/server/frontend_wayland/wayland_rs/src/protocol_impls/mod.rs
```diff
@@ -0,0 +1,9 @@
+use crate::wayland_server_core::ServerState;
+
+mod wl_fixes;
+
+pub fn register_globals(state: &ServerState) {
+    state
+        .handle
+        .create_global::<ServerState, wayland_server::protocol::wl_fixes::WlFixes, ()>(1u32, ());
+}
```
RAOF, line 9: Hm. This works for now, but that's because `wl_fixes` is super-simple; I expect that other protocols will require state more complicated than `()` 😁 

I'm not sure what that would look like, though, and I don't think we need to decide that now.

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
```diff
@@ -397,8 +397,15 @@ fn generate_global_dispatch_impl(
 ) -> TokenStream {
     let interface_name = dash_to_snake_ident(&interface.name);
 
-    if interface_name == "wl_display" {
-        // wl_display is handled specially in wayland_server crate via the 'Display' struct.
+    if implemented_protocols()
```
RAOF: This might be clearer as `natively_implemented_protocols()`? It was a little confusing the first time round why we were replacing some simple string checks with a much more complex iteration.

RAOF (changes requested):
Only thing I think we should definitely do is fix the incorrect indentation; the rest are optional (or just comments).

src/server/frontend_wayland/wayland_rs/build_script/wayland_server_generation.rs
andogq: Oversight by me, whoops! `rustfmt` wasn't properly formatting due to the `quote` macro at the end of the closure, as it sometimes can't handle long chains and macros. Manually re-formatting the `quote` usage below restores `rustfmt`.

src/server/frontend_wayland/wayland_rs/build_script/dispatch_generator.rs
andogq: Done!

RAOF (approved):
Yup! Looks good to me now.

